PR #840, opened 2015-02-10
forward declare GLX typedefs to limit conflicts from glx.h

mikemorris:


jfirebaugh:
IMHO it would be better to use the [pimpl idiom](http://herbsutter.com/gotw/_100/) here.

mikemorris:
@jfirebaugh that really feels like unnecessary complexity in this case to me, and the compilation benefits would be negligible for such a narrowly-used header file.

jfirebaugh:
The point would be to not have to do all that forward declaration of GLX stuff, which could change between versions or platforms.

jfirebaugh:
I see the forward declarations are more minimal now, so I don't have strong objections though.

Issue #820, opened 2015-02-06
Fail to launch app on android


thomas-huet:
I have built the app from 91760aa to test it but it crashes immediately:
https://gist.github.com/thomas-huet/63de73eec2a99afc4959
I have used a nexus 7 with android 5.0.2

ljbade:
Hmm missing symbol `_ZN4mbgl8platform9assetRootEv`

ljbade:
How did you build it? Just `make android`?

ljbade:
Ah I am getting it to.

@kkaefer Related to your storage refactor?

thomas-huet:
Yes I have just built the app with `make android`.

ljbade:
Ah I know what it is!

@kkaefer you for an assetRoot for Android.

ljbade:
@thomas-huet https://github.com/mapbox/mapbox-gl-native/tree/android-asset-root will fix the missing symbol, but there are a number of crashes from a recent refactor that we need to fix before the app will actually work.

If you want to test the app when it was stable, try https://github.com/mapbox/mapbox-gl-native/commit/3bfea8bf30c978173f1ec2fab6f89d6b33afea86

thomas-huet:
OK thanks.

kkaefer:
We're now one step further, but the style still doesn't load, reporting a "SSL connect error: Unknown SSL protocol error in connection to www.mapbox.com:443"

PR #844, opened 2015-02-10
Android asset root

ljbade:
Fixes https://github.com/mapbox/mapbox-gl-native/issues/820


ljbade:
Moving to fix-android-travis.

Issue #830, opened 2015-02-10
iOS integration testing


incanus:
Carryover from https://github.com/mapbox/mapbox-gl-native/pull/824. Will need to pull in goodies from https://github.com/mapbox/mapbox-gl-cocoa/blob/f36ee137496c8cf11ab604cacbb63ca645e6741c/.travis.yml.

incanus:
Old `.travis.yml` at https://github.com/mapbox/mapbox-gl-cocoa/blob/f36ee137496c8cf11ab604cacbb63ca645e6741c/.travis.yml.

incanus:
Related: https://github.com/mapbox/mapbox-gl-native/issues/880

incanus:
Woo, the return of iOS tests! 

![anigif-1427848638](https://cloud.githubusercontent.com/assets/17722/6932327/980da9f8-d7cc-11e4-8d94-a3bf4862b46e.gif)

PR #853, opened 2015-02-11
fixes #476: pixel / projected meter / lat/long conversions

### LINKED ISSUE #476: conversion routines
* [x] pixels `<->` projected meters `<->` coordinates
 * [x] meters/pixel at a given zoom & latitude

Precursors to point plotting #60 and interactivity #352. 

Related: #97 

Happening in [`conversions`](https://github.com/mapbox/mapbox-gl-native/tree/conversions).

incanus:
Remaining: 

* [x] Pick up `conversions` work again and merge in `master` (over 1,000 commits).

* [x] Integrate old `gl-cocoa#conversions` changes into this repo.

* [x] Reduce overlap between `Transform` and `TransformState` as a result of this branch + https://github.com/mapbox/mapbox-gl-native/commit/4603ca037cab2d0f732285ce9e006744ad3118d6 (background image support /cc @jfirebaugh).

* [ ] Write specific tests for exercising conversion routines (will hit iOS specifically in #830). 

* [x] Merge this in (my excuse to finally try rebasing). 

* [x] Delete old [`gl-cocoa#conversions`](https://github.com/mapbox/mapbox-gl-cocoa/tree/conversions) to clean house.


incanus:
Ok, I think we're set here, including migrating API across OS X, iOS, Linux, and Android. 

Working on tests now.

ljbade:
:-D

include/mbgl/map/map.hpp
```diff
@@ -106,8 +108,8 @@ class Map : private util::noncopyable {
     double getScale() const;
     void setZoom(double zoom, std::chrono::steady_clock::duration duration = std::chrono::steady_clock::duration::zero());
     double getZoom() const;
-    void setLonLatZoom(double lon, double lat, double zoom, std::chrono::steady_clock::duration duration = std::chrono::steady_clock::duration::zero());
-    void getLonLatZoom(double &lon, double &lat, double &zoom) const;
+    void setLatLngZoom(LatLng latLng, double zoom, std::chrono::steady_clock::duration duration = std::chrono::steady_clock::duration::zero());
+    void getLatLngZoom(LatLng &latLng, double &zoom) const;
```
jfirebaugh: Can we remove this method and use separate calls to `getLatLng()` and `getZoom()`?

include/mbgl/map/map.hpp
```diff
@@ -127,6 +129,15 @@ class Map : private util::noncopyable {
     void setAccessToken(const std::string &token);
     const std::string &getAccessToken() const;
 
+    // Projection
+    inline void getWorldBoundsMeters(ProjectedMeters &sw, ProjectedMeters &ne) const { Projection::getWorldBoundsMeters(sw, ne); }
```
jfirebaugh: Instead of passing out args via reference, there should be a `template <class T> class Bounds` and the return type of this function should be `Bounds<ProjectedMeters>`.

include/mbgl/map/map.hpp
incanus: Yeah, I think so. This was just ported for historical reasons. Good reason to scrap it.

include/mbgl/map/transform.hpp
```diff
@@ -52,6 +53,10 @@ class Transform : private util::noncopyable {
     void startRotating();
     void stopRotating();
 
+    // Projection
+    void pixelForLatLng(const LatLng latLng, double &x, double &y) const;
```
jfirebaugh: Should return `vec2<double>`.

src/mbgl/renderer/painter.hpp
```diff
@@ -178,7 +177,7 @@ class Painter : private util::noncopyable {
     }();
 
 private:
-    TransformState state;
+    Transform& transform;
```
jfirebaugh: Can we revert the changes to painter.hpp/cpp and painter_*.cpp? We're trying to reduce shared state through references, and using `Transform&` reverses that.

incanus: Yeah, that's my one worry about this PR, the way we "poke through" the transform to grab the state. The painter needs to know the conversion (as put in originally in your background stuff in 4603ca0), but perhaps this can be pushed down to the state, then have the public map APIs poke through map to transform to transform state to get them. Will look at this.

jfirebaugh: If it helps, I think it's fine to have `Painter::render` take a `Transform` by copy rather than a `TransformState`, so that this member becomes `Transform transform`. In that case I would still keep a `TransformState& state` member around that references `transform.currentState()` to avoid needing to change all the paint implementation methods.

Long term I think `Transform` and `TransformState` will merge.

incanus: `Transform` is non-copyable?

include/mbgl/map/map.hpp
incanus: 429f7cf

include/mbgl/map/transform.hpp
incanus: Reason I did this was because calling code (e.g. iOS) would have to know about `vec2` type in this case.

incanus: Also isn't there an implicit `x, y` ordering that would be lost?

incanus: f13ea06803667d12bf501b8a98865603001bde95

src/mbgl/renderer/painter.hpp
incanus: I'm remembering why I did this now — this would require moving `transform.getLatLng()` down to the state as well, along with the spherical mercator math cache values. Seemed intrusive, but perhaps less so than the painter stuff. Trying this now.

incanus: c2a9adb13319b18e54f4c2637a7bbdd0f8283198

incanus:
This is now working, so I'm inclined to merge it once it passes Travis and write tests in `master`.

PR #860, opened 2015-02-11
Update android paths in make clean

ljbade:


ljbade:
@bleege sneaked this in already.

PR #879, opened 2015-02-13
Rework map threading

kkaefer:
This changes the way Map objects are being rendered. Instead of `start()` and `stop()` launching a new thread, we're now starting a new thread as soon as the Map object is constructed, which runs until the Map object is destructed. This means that we don't have to start/stop thread for static rendering and overall simplifies the way we access OpenGL from multiple threads (we don't anymore).

- [x] OpenGL initialization in Map thread
- [x] Handle HeadlessView resizes 
- [x] Make stop() wait until rendering is finished
- [x] Guard OpenGL extension function pointer acquisition against race conditions (=> uv_once)
- [x] Terminate must terminate all pending requests from this map. This likely requires that the FileSource knows about Map object and keeps track of which Request objects belong to what Maps so we can cancel those automatically
- [ ] Fix android builds
- [ ] Get iOS backgrounding/foregrounding actions right
- [ ] ~~Disable fading for static map renders~~ (see #942)

### DIFF
--- a/include/mbgl/storage/default_file_source.hpp
+++ b/include/mbgl/storage/default_file_source.hpp
@@ -20,11 +20,14 @@ class DefaultFileSource : public FileSource {
 public:
     DefaultFileSource(FileCache *cache, const std::string &root = "");
     DefaultFileSource(FileCache *cache, uv_loop_t *loop, const std::string &root = "");
-    ~DefaultFileSource();
+    ~DefaultFileSource() override;
 
-    Request *request(const Resource &resource, uv_loop_t *loop, Callback callback);
-    void cancel(Request *request);
-    void request(const Resource &resource, Callback callback);
+    Request *request(const Resource &resource, uv_loop_t *loop, const Environment &env,
+                     Callback callback) override;
+    void cancel(Request *request) override;
+    void request(const Resource &resource, const Environment &env, Callback callback) override;
+
+    void abort(const Environment &env) override;
 
     enum class CacheHint : uint8_t { Full, Refresh, No };
     void notify(SharedRequestBase *sharedRequest, const std::set<Request *> &observers,
@@ -39,14 +42,16 @@ private:
     struct RemoveRequestAction;
     struct ResultAction;
     struct StopAction;
-    using Action =
-        mapbox::util::variant<AddRequestAction, RemoveRequestAction, ResultAction, StopAction>;
+    struct AbortAction;
+    using Action = mapbox::util::variant<AddRequestAction, RemoveRequestAction, ResultAction,
+                                         StopAction, AbortAction>;
     using Queue = util::AsyncQueue<Action>;
 
     void process(AddRequestAction &action);
     void process(RemoveRequestAction &action);
     void process(ResultAction &action);
     void process(StopAction &action);
+    void process(AbortAction &action);
 
     SharedRequestBase *find(const Resource &resource);
 
--- a/.clang-format
+++ b/.clang-format
@@ -2,7 +2,7 @@ Standard: Cpp11
 IndentWidth: 4
 AccessModifierOffset: -4
 UseTab: Never
-BinPackParameters: false
+BinPackParameters: true
 AllowShortIfStatementsOnASingleLine: false
 AllowShortLoopsOnASingleLine: false
 AllowShortBlocksOnASingleLine: false
--- a/Makefile
+++ b/Makefile
@@ -181,6 +181,16 @@ android-deploy: $(ANDROID_ABIS)
 	cd android/java/MapboxGLAndroidSDK && chmod ugo+x deploy.sh && ./deploy.sh
 
 
+##### Render builds ############################################################
+
+.PRECIOUS: Makefile/render
+Makefile/render: bin/render.gyp config/$(HOST).gypi
+	deps/run_gyp bin/render.gyp $(CONFIG_$(HOST)) $(LIBS_$(HOST)) --generator-output=./build/$(HOST) -f make
+
+render: Makefile/render
+	$(MAKE) -C build/$(HOST) BUILDTYPE=$(BUILDTYPE) mbgl-render
+
+
 ##### Maintenace operations ####################################################
 
 .PHONY: clear_xcode_cache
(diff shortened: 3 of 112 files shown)


mb12:
I am unable to reason how the following scenario would  work correctly in the new threading model. 

1.) SourceInfo objects are created upon parsing style.json.
2.) TileData holds a reference to SourceInfo.
3.) TileParser holds a reference to TileData.

When style.json is changed these SourceInfo objects would get deleted. ( They reside under Map.style.layers.layers.bucket.style_source.info ).
If any of the TileParser workers has not finished after Map.style has been changed, this would cause an invalid access.

mikemorris:
Working with @ljbade to debug a deadlock on the [`std::recursive_mutex`](https://github.com/mapbox/mapbox-gl-native/blob/static-render/include/mbgl/map/transform.hpp#L81) in `mbgl::Transform`, seems to get blocked in [`Transform::resize`](https://github.com/mapbox/mapbox-gl-native/blob/static-render/src/mbgl/map/transform.cpp#L25). Digging through some of the suggestions in https://stackoverflow.com/questions/13755355/android-ndk-mutex-locking currently...

ljbade:
OK I set up some logging to print the mutex value before and after each lock in Transform.

Here is what happens during deadlock:
```
02-25 17:33:43.194    7895-7895/com.mapbox.mapboxgl.app D/Android﹕ mtx value before = 4000
02-25 17:33:43.194    7895-7895/com.mapbox.mapboxgl.app D/Android﹕ mtx value after = 1ed74001
...
02-25 17:33:43.194    7895-7895/com.mapbox.mapboxgl.app D/Android﹕ Transform::setzoom
02-25 17:33:43.194    7895-7895/com.mapbox.mapboxgl.app D/Android﹕ mtx value before = 4000
02-25 17:33:43.194    7895-7895/com.mapbox.mapboxgl.app D/Android﹕ mtx value after = 1ed74001
...
02-25 17:33:43.194    7895-7895/com.mapbox.mapboxgl.app D/Android﹕ Transform::setangle
02-25 17:33:43.194    7895-7895/com.mapbox.mapboxgl.app D/Android﹕ mtx value before = 4000
02-25 17:33:43.194    7895-7895/com.mapbox.mapboxgl.app D/Android﹕ mtx value after = 1ed74001
...
02-25 17:33:43.334    7895-7895/com.mapbox.mapboxgl.app I/MapView﹕ resize 4 1080 1677
02-25 17:33:43.344    7895-7895/com.mapbox.mapboxgl.app D/JNI﹕ nativeResize
02-25 17:33:43.344    7895-7895/com.mapbox.mapboxgl.app D/Android﹕ Map::resize
02-25 17:33:43.344    7895-7895/com.mapbox.mapboxgl.app D/Android﹕ Transform::resize
02-25 17:33:43.344    7895-7895/com.mapbox.mapboxgl.app D/Android﹕ mtx value before = 7989d5e8
```

Something is corrupting the mutex value which results in a deadlock because it thinks some random thread is holding the lock.

ljbade:
I also logged the pointer value (that points to the mutex value). You can see the pointer is being changed, when it shouldn't so the value is just random memory.

```
02-25 17:40:53.313  12000-12000/com.mapbox.mapboxgl.app D/Android﹕ mtx pointer before = 7288b274, mtx value before = 4000
02-25 17:40:53.313  12000-12000/com.mapbox.mapboxgl.app D/Android﹕ mtx pointer before = 7288b274, mtx value after = 2ee04001
02-25 17:40:53.313  12000-12000/com.mapbox.mapboxgl.app D/Android﹕ mtx pointer before = 7288b274, mtx value before = 4000
02-25 17:40:53.313  12000-12000/com.mapbox.mapboxgl.app D/Android﹕ mtx pointer before = 7288b274, mtx value after = 2ee04001
02-25 17:40:53.313  12000-12000/com.mapbox.mapboxgl.app D/Android﹕ mtx pointer before = 7288b274, mtx value before = 4000
02-25 17:40:53.313  12000-12000/com.mapbox.mapboxgl.app D/Android﹕ mtx pointer before = 7288b274, mtx value after = 2ee04001
02-25 17:40:53.473  12000-12000/com.mapbox.mapboxgl.app D/Android﹕ mtx pointer before = 7288b1f4, mtx value before = 7989e5e8
```

kkaefer:
@mikemorris please note that the API of `FileSource` changed: https://github.com/mapbox/mapbox-gl-native/commit/335cf4cb5138cc951e2d36db385dcda9e14f8d4f. There's now an opaque `Environment` ref  that is used to abort all requests originating from a Map. This is likely not very relevant for node-mapbox-gl-native since there's currently no way to abort a static map image render, but it needs to support the API nonetheless.

mikemorris:
Pushed up some minimal changes in the [`environment`](https://github.com/mapbox/node-mapbox-gl-native/compare/environment) branch to get everything compiling again. Is it acceptable to leave the `NodeFileSource::abort` method as a no-op and hold off on adding an `Abort` action, or should I work on building out this pieces too?

include/mbgl/storage/default_file_source.hpp
```diff
@@ -20,11 +20,14 @@ class DefaultFileSource : public FileSource {
...
-    void cancel(Request *request);
-    void request(const Resource &resource, Callback callback);
+    Request *request(const Resource &resource, uv_loop_t *loop, const Environment &env,
+                     Callback callback) override;
+    void cancel(Request *request) override;
+    void request(const Resource &resource, const Environment &env, Callback callback) override;
+
+    void abort(const Environment &env) override;
 
     enum class CacheHint : uint8_t { Full, Refresh, No };
```
mikemorris: It looks like this moved to `FileCache::Hint` and is no longer used?

kkaefer: Yup, removed in 10d84f6.

kkaefer:
Removing this from the milestone as all relevant patches have been integrated into master separately. Will circle back to this after iOS Beta release

mikemorris:
Did something change with the `Environment` work to start emitting `DEBUG` severity level events even with `BUILDTYPE=Release`?

tmpsantos:
I hit this one today with the current HEAD (1ce51a17e7de5d6a02346efce1539ff7f36e0a6d). I think the easiest way to trigger is to change the map orientation, because it seems that on the testapp a new Map object is created when you rotate the app. Sad that the backtrace is not telling much.

kkaefer:
This has been superseded by #1272

PR #902, opened 2015-02-18
Use Cask to install Android Studio Development Environment and dependencies on OS X

1ec5:
With this change, the preferred way to install Android Studio and its dependencies on OS X involves [Cask](http://caskroom.io/) instead of manually downloading binaries from the browser. These instructions work for me on OS X 10.10.2 with Homebrew 0.9.5 and Cask 0.53.3 installed, but of course I’d appreciate confirmation that they actually work.

### DIFF
--- a/README.md
+++ b/README.md
@@ -14,7 +14,7 @@ implemented in C++11, currently targeting iOS, OS X, Android, and Ubuntu Linux.
  - [`libcurl`](http://curl.haxx.se/libcurl/) (depends on OpenSSL; Linux only)
  - Apple Command Line Tools (for build on OS X; available at [Apple Developer](https://developer.apple.com/downloads))
  - `pkg-config` (for build only)
- - [Homebrew](http://brew.sh) (for build on OS X)
+ - [Homebrew](http://brew.sh) and [Cask](http://caskroom.io/) (for building on OS X)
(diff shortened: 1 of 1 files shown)


1ec5:
Anything missing here, @incanus @bleege?

bleege:
Looks pretty good @1ec5.  Only changes would be to:

```sh
JAVA_HOME=<wherever cask installs it>
STUDIO_JDK=$JAVA_HOME
```

incanus:
Will give this a try.

1ec5:
Hang on, I have a typo in there.

kkaefer:
Cool, will try it out once I set up my actual computer next week.

bleege:
Just updated the title of this.  "Build Dependencies" has a different meaning in the Java / Android world and don't want it to be confusing for people.

ljbade:
If only it were this easy on Linux.

incanus:
:+1: To me, gonna merge since it's `README`-only and kill the Travis build. Thanks @1ec5!

kkaefer:
Great write-up! You can also add `[skip ci]` to the commit message to prevent Travis builds

1ec5:
@kkaefer wow thanks!

PR #899, opened 2015-02-19
organize platform files better

jfirebaugh:
Android files seem to be spread out in four different places:

* `android`
* `include/mbgl/android`
* `include/mbgl/platform/android`
* `platform/android`

Similarly for iOS:

* `ios`
* `include/mbgl/ios`
* `platform/ios`

Proposed reorganization consolidates all platform files into subdirectories of `platform`:

* `platform`
 * `android`
   * `app`
    * `include`
    * `src`
 * `ios`
   * `app`
    * `include`
    * `src`
 * `darwin`
 * `default`
 * `linux`
 * `macosx`

Like the top-level `src` directory, `platform/{android,ios}/src/` can include private headers as well as implementation files. Only public headers live in `platform/{android,ios}/include/` (in an `mbgl` subdirectory).


jfirebaugh:
Converted to PR. I have very little idea what I'm doing here though!

Makefile
```diff
@@ -208,13 +208,8 @@ endif
 clean: clear_sqlite_cache clear_xcode_cache
 	-find ./deps/gyp -name "*.pyc" -exec rm {} \;
 	-rm -rf ./build/
-	-rm -rf ./macosx/build
-	-rm -rf ./linux/build
-	-rm -rf ./ios/build
 	-rm -rf ./test/build
 	-rm -rf ./config/*.gypi
-	-rm -rf ./android/java/build ./android/java/MapboxGLAndroidSDKTestApp/build ./android/java/MapboxGLAndroidSDKTestApp/build
```
ljbade: This line should be kept, but point to the new `platform/android/java/.../build` folders since the Gradle/Java keeps it's build files here instead of in `/build` which is only C++ build files.

Makefile
```diff
@@ -208,13 +208,8 @@ endif
...
-	-rm -rf ./android/java/MapboxGLAndroidSDK/src/main/jniLibs ./android/java/MapboxGLAndroidSDK/src/main/assets
```
ljbade: Same with this.

ljbade:
I think https://github.com/mapbox/mapbox-gl-native/blob/platform-reorg/scripts/android_env.sh should be moved to the same place as the other android scripts.

incanus:
I like where this is headed @jfirebaugh. When I picked up @kkaefer's `gl-cocoa` consolidation work, I didn't like how separated the iOS headers and implementations were. I think having them in peer older (along with `app`) is a nice way to organize.

jfirebaugh:
I made the changes @ljbade suggested.

mikemorris:
From #989, there are other implementation files floating around all over the place too, like the following:

- [include/mbgl/storage/default_file_source.hpp](https://github.com/mapbox/mapbox-gl-native/blob/master/include/mbgl/storage/default_file_source.hpp)
- [src/mbgl/storage/default_file_source.cpp](https://github.com/mapbox/mapbox-gl-native/blob/master/src/mbgl/storage/default_file_source.cpp)
- [include/mbgl/storage/file_cache.hpp](https://github.com/mapbox/mapbox-gl-native/blob/master/include/mbgl/storage/file_cache.hpp)
- [include/mbgl/storage/default/sqlite_cache.hpp](https://github.com/mapbox/mapbox-gl-native/blob/master/include/mbgl/storage/default/sqlite_cache.hpp)

jfirebaugh:
Closing here; I'll basically need to redo this from scratch because so much has changed.

PR #916, opened 2015-02-25
Removed special cased zoom level for HiDPI

1ec5:
No need to load higher-resolution vector tiles for HiDPI displays. This PR does change the zoom levels at which features appear on HiDPI screens, but it should be consistent with LoDPI:

before after


incanus:
Cool, this looks good to me. See some tests at https://www.dropbox.com/sh/ohcb9j56r1pw0ao/AACPXgPlNB2Ps1mMzQ5LFgyEa?dl=0. 

The change is most apparent between `@2x` raster views, where things were too tiny before and loaded `4x` as many tiles as needed, as a result of the original dev happening on the `v3` API but now having the `v4` one with `512px` tiles.

PR #292, opened 2014-05-31
Boost regex

springmeyer:
Use `boost::regex` on linux since `std::regex` is unimplemented in `libstdc++` at g++ 4.8. Once we start targeting a system with g++-4.9 we can revisit dropping `boost::regex` and moving back to `std::regex` for all platforms.

PR #924, opened 2015-02-27
Load 2× raster tiles on HiDPI screens

### LINKED ISSUE #919: v4 raster layers load with low-resolution tiles on high-resolution screens
#916 removed a special case for HiDPI that bumped up the zoom level, the idea being that vector tiles have a sufficiently high resolution that the HiDPI experience wouldn’t suffer. But this same code runs for raster layers. Now the Satellite layer is half as sharp:

before after

1ec5:
Based on `mapbox.normalizeTileURL()` in mapbox/mapbox-gl-js, but insert a `{ratio}` placeholder instead of `@2x` in order to keep the `pixelRatio` logic in `TileData::request()`.

Fixes #919, so Satellite should look like it did before feae53fb3f81b4a09372324c7c839a03d31d9b66:

fixed

### DIFF
--- a/src/mbgl/util/mapbox.cpp
+++ b/src/mbgl/util/mapbox.cpp
@@ -1,7 +1,6 @@
 #include <mbgl/util/mapbox.hpp>
 
 #include <stdexcept>
-#include <regex>
 
 namespace mbgl {
 namespace util {
@@ -44,8 +43,23 @@ std::string normalizeTileURL(const std::string& url, const std::string& sourceUR
     if (sourceURL.empty() || sourceURL.compare(0, mapbox.length(), mapbox) != 0)
         return url;
     
-    static std::regex extension_re("\\.((?:png|jpg)\\d*)(?=$|\\?)");
-    return std::regex_replace(url, extension_re, std::string("{ratio}.$1"));
+    std::string::size_type queryIdx = url.rfind("?");
+    // Trim off the right end but never touch anything before the extension dot.
+    std::string urlSansParams((queryIdx == std::string::npos) ? url : url.substr(0, queryIdx));
+    
+    while (!urlSansParams.empty() && isdigit(urlSansParams.back()))
+        urlSansParams.pop_back();
+    
+    std::string::size_type extensionIdx = urlSansParams.length() - 4;
+    if (extensionIdx <= 0)
+        return url;
+    std::string extension = urlSansParams.substr(extensionIdx);
+    if (extension.compare(".png") != 0 && extension.compare(".jpg") != 0)
+        return url;
+    
+    std::string normalizedURL(url);
+    normalizedURL.insert(extensionIdx, "{ratio}");
+    return normalizedURL;
 }
 
 }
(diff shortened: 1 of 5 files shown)


src/mbgl/util/mapbox.cpp
```diff
@@ -39,6 +39,20 @@ std::string normalizeGlyphsURL(const std::string& url, const std::string& access
     return normalizeURL(url, accessToken);
 }
 
+std::string normalizeTileURL(const std::string& url) {
+    if (url.compare(url.length() - 4, 4, ".png") == 0)
```
incanus: It's not just `.png` that's relevant here. 

https://www.mapbox.com/developers/api/maps/#format

jfirebaugh: It's also important that this only insert `{ratio}` on Mapbox sources, since we can't guarantee that other sources support `@2x`. That's why [gl-js check the source URL](https://github.com/mapbox/mapbox-gl-js/blob/8ee8e879d6863821278dbdd99ab427458b0f04ee/js/util/mapbox.js#L62) against `mapbox://`.

jfirebaugh: Fixed gl-js support for other image formats in https://github.com/mapbox/mapbox-gl-js/commit/241a07d0607447cbc393f2ccfab3f03e765ced13.

1ec5:
@incanus @jfirebaugh are we good to go here?

jfirebaugh:
Previous use of `std::regex` was discussed in #292 and #521. Did we switch to targeting gcc 4.9?

1ec5:
Oh, according to the README, we’re still targeting g++ 4.8 for Linux. I’ll get rid of the dependency.

jfirebaugh:
I wonder how the linux+gcc travis build using std::regex still passed though...

mb12:
Instead of hardcoding for file extensions (.jpg, .png etc),  you can use source.type == SourceInfo::Raster in TileData::Request.

1ec5:
Looks like it’s still failing on Linux with `REG_BADPAT`. Is it possible that this compiler is defaulting to some syntax other than `ECMAScript`?

1ec5:
It’s `error_escape` apparently.

1ec5:
I think we’re still looking at a [stubbed-out `std::regex` implementation](http://stackoverflow.com/questions/12530406/is-gcc4-7-buggy-about-regular-expressions/12665408#12665408) in GCC 4.8. It seems to only support basic regex even if you specify ECMAScript syntax (which should be the default).

jfirebaugh:
**sigh**

Ok, let's revert to the string based implementation and open an issue about upgrading to gcc 4.9.

mb12:
There is no need to do any string manipulation for this issue. Just pass an additional boolean flag to normalize method that '@ratio' needs to be appended. In caller just use SourceInfo.type == SourceType::Raster for this boolean flag.

1ec5:
“{ratio}” needs to be added before the file extension and any parameters, not appended, hence the regex / string manipulation. But `SourceInfo.type` is a good idea.

PR #930, opened 2015-02-28
Skip full-width CJK block

friedbunny:


friedbunny:
Should probably ignore this hacky patch, seems like it introduced glyph corruption and didn't address the underlying issues.

Issue #893, opened 2015-02-17
merge GeoJSON vector tiles work


incanus:
Picking up from https://github.com/mapbox/mapbox-gl-native/issues/507#issuecomment-69076871, we'll need to merge the `geojsonvt` branch, which will require it interfacing better with the library. 

* [x] match vector tile object structure with main library
* [x] allow for live reindexing when features are added/removed
* [x] handle non-GeoJSON (in-memory features)
* [x] nailing some hopefully-basic performance kinks
* [ ] `MultiGeometry` support
* [x] other API merging (e.g. `LatLng` struct)
* [x] figuring out how independent of or reliant on GL we want the library to be

/cc @mourner @1ec5

incanus:
The library consists of [`geojsonvt.hpp`](https://github.com/mapbox/mapbox-gl-native/blob/a9cf3ffb6a88aca56ecdc838a1513557af069058/include/mbgl/util/geojsonvt.hpp) and [`geojsonvt.cpp`](https://github.com/mapbox/mapbox-gl-native/blob/a9cf3ffb6a88aca56ecdc838a1513557af069058/src/mbgl/util/geojsonvt.cpp).

incanus:
@kkaefer @mourner Do you guys think this should be a part of GL itself, or should we try to keep it a separate library for general use of geo data `->` vector tiles?

mourner:
@incanus it would be nice to have it as a separate library — easier to maintain this way, with separate tests etc. It could also grow into a useful standalone tool.

incanus:
Related to https://github.com/mapbox/mapbox-gl-native/issues/904, we can't subclass `Source` to make e.g. a `LiveSource` (my first approach), so instead I am thinking: 

* Use `Source`, possibly with manually-created or passed `SourceInfo` related to the data (GeoJSON or other) setup. 
* Adapt `Source::addTile()` to handle `SourceType:GeoJSON` (currently unimplemented) as a first measure. 
* Above relies on a new `GeoJSONTileData` class which works similar to `VectorTileData`, but which pulls tiles from the `GeoJSONVT` structure. 
* A `Source` initialized in this way (perhaps with passed GeojSON) would async run the `GeoJSONVT` clipping & splitting algorithm upon creation so that it would be ready for these tile queries. This shouldn't take long, but maybe it will block `addTile()` calls until it's done?

incanus:
[...]

Issue #399, opened 2014-08-02
variable naming style


jfirebaugh:
We're currently using a mix of [camel-case](https://github.com/mapbox/mapbox-gl-native/blob/5f208c268b28d93027d9178ac4642a7b269db380/include/mbgl/map/map.hpp#L171-L175) and [underscore](https://github.com/mapbox/mapbox-gl-native/blob/5f208c268b28d93027d9178ac4642a7b269db380/include/mbgl/map/map.hpp#L193-L195) variable names. My vote is for camel-case simply to be consistent with JS.

jfirebaugh:
We're getting to be in good shape here. Let's continue to make things camel-case when we encounter variables that aren't.

tmpsantos:
I would like to suggest suffixing class attributes with a `_` as in `myAttribute_`.

jfirebaugh:
@tmpsantos If we were starting from scratch, I'd agree. But none of the existing code uses that convention, and in fact much of it uses a `_` suffix on parameters to avoid shadowing a member (the opposite convention). I think it's too much churn to make a change here.

PR #937, opened 2015-03-03
clean up members, arguments, and formatting

incanus:
Hit some weirdness over in #928 with the naming here and posting some cleanups. 

 * Since we're dealing with both `VectorTileData` (base: `TileData`) and `VectorTile` classes, we should refer to raw data retrieved from a vector tile download and used to instantiate a `VectorTile` as `raw_data`, not just `data`. 

* The `VectorTile` instantiated is now called `vector_tile`, not `vector_data`. 

* The `VectorTileData& tile` is not ideal, since elsewhere in the project we *also* have a `Tile` class, but for the purposes of convenience throughout the implementation, it stays. 

* Overall formatting changes like `ClassName& argument_name` — reference `&` next to type, then a space, then the variable name with `_` and not `camelCase`. 

kyleve_2015-jan-07

### DIFF
--- a/src/mbgl/map/tile_parser.cpp
+++ b/src/mbgl/map/tile_parser.cpp
@@ -29,14 +29,14 @@ namespace mbgl {
 // its header file.
 TileParser::~TileParser() = default;
 
-TileParser::TileParser(const std::string& raw_data_,
+TileParser::TileParser(const std::string& rawData_,
                        VectorTileData& tile_,
                        const util::ptr<const Style>& style_,
                        GlyphAtlas& glyphAtlas_,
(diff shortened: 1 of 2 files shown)


incanus:
BTW explicit mention of vector tiles will go away soon, once #928 lands and I'm back on #893 ;-)

jfirebaugh:
Looks good but we are moving toward camelCase variable names, not underscore: #399.

incanus:
Cool, will fix while I'm in there.

incanus:
Moved to camelCase for members and externally-available arguments.

Issue #936, opened 2015-03-03
iOS line-dasharray small values don't render properly


friedbunny:
When using outdoor-v7's "hatching" (narrow, perpendicular crossing lines) to decorate lines, iOS fattens them up.

**Aerial way**

![screen-shot-2015-03-02-at-4 07 58-pm](https://cloud.githubusercontent.com/assets/1198851/6453918/595d94a6-c0f8-11e4-9038-c0743bd07d2a.png)

```json
{
  "id": "bridge_aerialway_hatching",
  "ref": "bridge_aerialway_casing",
  "paint": {
    "line-color": "#876",
    "line-dasharray": {
      "stops": [[12, [0.1, 6]], [15, [0.12, 12]], [18, [0.15, 18]]]
    },
    "line-width": {
      "base": 1.2,
      "stops": [[12, 5], [20, 15]]
    }
  }
}
```

**Railroad**

![screen-shot-2015-03-02-at-4 10 04-pm](https://cloud.githubusercontent.com/assets/1198851/6453925/64ff8a3a-c0f8-11e4-9e36-322c443ae95a.png)

```json
{
  "id": "road_major_rail",
  "type": "line",
  "source": "mapbox",
  "source-layer": "road",
  "filter": ["==", "class", "major_rail"],
  "paint": {
    "line-color": "#c8c4c0",
    "line-width": 0.8
  }
}, {
  "id": "road_major_rail_hatching",
  "ref": "road_major_rail",
  "paint": {
    "line-color": "#c8c4c0",
    "line-dasharray": [0.05, 10],
    "line-width": {
      "stops": [[12, 5], [20, 10]]
    }
  }
}
```

friedbunny:
This turned out to be an issue with outdoors-v6 that has been fixed in outdoors-v7. Currently the OSX and iOS apps are on v7 and v6, respectively... which was confusing. :no_good:

PR #939, opened 2015-03-03
Bump iOS app styles versions to 7

friedbunny:
Fixes goose hunts like #936. :hatched_chick:

### DIFF
--- a/ios/app/MBXViewController.mm
+++ b/ios/app/MBXViewController.mm
@@ -9,10 +9,10 @@
 static UIColor *const kTintColor = [UIColor colorWithRed:0.120 green:0.550 blue:0.670 alpha:1.000];
 
 static NSDictionary *const kStyles = @{
-    @"bright-v6":    @"Bright",
-    @"basic-v6":     @"Basic",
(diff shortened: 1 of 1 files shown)


incanus:
:+1: 

Longer-term, we'll probably auto-update these constants somehow, but this is just for purposes of the demo/test app.

incanus:
Weirdly, this change surfaced a bug in the way that we list bundled styles in the test app, namely that we have an `NSDictionary` mapping style display names to JSON file names, but get at `[kStyles allValues]` when picking the style to show at start — which is not guaranteed order. In my setup, this was causing the name to not match the style shown. Fixed in 28228bfae0020f9d379dbcbb1834e5d898956750. 

Also, we needed to update in `MGLMapView` the default style to use upon startup, independent of it existing in the test app. Fixed in 6171cbb748facf9a783661316fc045d70ff9e7dc. 

When this [passes Travis](https://travis-ci.org/mapbox/mapbox-gl-native/builds/52944737), I will merge. Thanks @friedbunny!

friedbunny:
Ahah, much more robust than just changing a couple digits — good deal!

Issue #719, opened 2014-12-12
Android: Unable to load styles.json


hallahan:
I'm attempting to build the android-mason branch on my Samsung Galaxy S4 via the instructions on the README.md. So far so good. I got it all to compile and build on the phone.

It builds and the app opens when I run it, but the map fragment is blank, no tiles are loaded. It can't find the style file.

```
12-12 07:55:12.978    9780-9794/com.mapbox.mapboxgl.app E/Setup﹕ loading style failed: -1 (URL using bad/illegal format or missing URL)
12-12 07:55:13.018    9780-9799/com.mapbox.mapboxgl.app E/Setup﹕ loading style failed: 404 (No such file)
```

I see that we try to get the style by:

```java
mMapFragment.getMap().setStyleUrl("asset://styles/styles/bright-v6.json");
```

However, there is nothing in the assets/styles directory. Where do I get these styles? I suppose I should just drop them in?

I'm looking at #585 and see that a similar problem was encountered and was resolved.

![screenshot 2014-12-12 09 59 32](https://cloud.githubusercontent.com/assets/556367/5414557/a138a024-81e5-11e4-8a36-8e3a0fe36c66.png)

I'm excited to see this work!

ljbade:
@hallahan Can try this debug build: https://www.dropbox.com/s/rg13hvsecbolwoz/app-release.apk?dl=0

hallahan:
Your new APK Build is also failing. When I pinch to zoom in, the app crashes.

```
I/DEBUG   ( 2553): *** *** *** *** *** *** *** *** *** *** *** *** *** *** *** ***
I/DEBUG   ( 2553): Build fingerprint: 'samsung/ja3gub/ja3g:4.4.2/KOT49H/I9500UBUFNB3:user/release-keys'
I/DEBUG   ( 2553): Revision: '10'
I/DEBUG   ( 2553): pid: 7784, tid: 7801, name: ox.mapboxgl.app  >>> com.mapbox.mapboxgl.app <<<
I/DEBUG   ( 2553): signal 6 (SIGABRT), code -6 (SI_TKILL), fault addr --------
I/DEBUG   ( 2553):     r0 00000000  r1 00001e79  r2 00000006  r3 00000000
I/DEBUG   ( 2553):     r4 00000006  r5 00000001  r6 00001e79  r7 0000010c
I/DEBUG   ( 2553):     r8 778fa5e4  r9 789ed000  sl bef20318  fp 78aead98
[...]

Issue #585, opened 2014-11-11
Map styles do not load from local file


ljbade:
@incanus encountered this error while testing the app:
`11-10 13:33:42.691    4742-4776/com.mapbox.mapboxgl.app E/Setup﹕ loading style failed: 404 (no such file or directory)`

He got around it by changing setStyleURL to a https source.

However I have not seen this error myself before.

@incanus Could you please add more detail about what you did to get this error?
e.g. was it Android 4 or 5, was it the emulator or your Nexus 7, how did you install/run the app etc.
Any other relevant log messages? (perhaps copy whole log into a gist)

I am going to see if I can recreate it by deleting the app's data folder.

Issue #579, opened 2014-11-09
Load map resources from local assets


ljbade:
Currently takes the app a second or two to load.

This is due to extracting the asset folder from APK.

Need to make this faster, get rid of it (by reading directly from APK), or only do it once at install time (can hook this), or only if the files are not there.

TODO: more detail on this so you guys understand what I am talking about

ljbade:
@kkaefer So where did this get to? I think Android was done right?

kkaefer:
This is implemented in other systems as well. OS X  and iPhone bundles load data from the resources folder, Linux builds well load assets relative to the binary's location

Issue #829, opened 2015-02-10
re-enable CocoaPods support


incanus:
Carryover from https://github.com/mapbox/mapbox-gl-native/pull/824. We should install `MapboxGL.framework` when CocoaPods 0.36.0 lands. Depends on https://github.com/mapbox/mapbox-gl-native/issues/828.

1ec5:
Framework support has been moved to #828. We’re just looking at distributing a static library via CocoaPods for this ticket.

incanus:
Yep, we'll eventually distribute both, I think? Not sure offhand how CocoaPods 0.36.0 (which *just* went stable) handles this.

1ec5:
I’d prefer to eventually move towards only distributing a framework, because that makes asset distribution easier, and because niceties like inspectables and designables are only possible with frameworks.

incanus:
But then don't we drop iOS 7?

1ec5:
Yeah, I meant eventually as in someday, when we’re comfortable with dropping iOS 7.

incanus:
Cool, agree :+1:

1ec5:
This pod depends on a prebuilt binary. See “[Releasing Mapbox GL for iOS](https://github.com/mapbox/mapbox-gl-native/wiki/Releasing-Mapbox-GL-for-iOS)” for release instructions.

PR #852, opened 2015-03-04
iOS error: "Setup: loading style failed: no such file or directory"

kkaefer:
I cloned the latest version of the repository (including submodules), ran `package_ios.sh` without errors, and added `Headers/`, `MapboxGL.bundle` and `libMapboxGL.a` to the appropriate build phases.

The build compiles, but complains on startup that "loading style failed".

screenshot-mapbox

This issue has been discussed before for Android (#585, #719). Looking through the source, it seems that the library is looking for styles in "asset://styles/styles/". From [the previous discussion](https://github.com/mapbox/mapbox-gl-native/issues/579), I'm not sure where `asset://` is on iOS (maybe MapboxGL.bundle? I confirm MapboxGL.bundle contains the `styles/` directory).

How can this be fixed?

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -21,10 +21,13 @@ config/%.gypi: CMD = ./configure config/$*.gypi
 config/%.gypi: configure
 	@$(ENV_$*) ./scripts/flock.py build/Configure.lock ./configure config/$*.gypi
 
+styles/styles:
+	git submodule update --init styles
+
 #### Library builds ############################################################
 
 .PRECIOUS: Makefile/mbgl
-Makefile/mbgl: config/$(HOST).gypi
+Makefile/mbgl: config/$(HOST).gypi styles/styles
 	deps/run_gyp mbgl.gyp $(CONFIG_$(HOST)) $(LIBS_$(HOST)) --generator-output=./build/$(HOST) -f make
 
 mbgl: Makefile/mbgl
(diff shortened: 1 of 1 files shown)


robipresotto:
still the same problem after set the bundledStyle to 
[self.mapView useBundledStyleNamed:@"outdoors-v7"];

by robitto 2015-03-05 at 17 56 16 2

incanus:
Please make sure your Git submodule are initialized @robipresotto.

robipresotto:
I'm running: make clean > make distclean | init submodule init > init submodule sync > init submodule update and ./scripts/package_ios.sh -> after build the package I replace my current package and nothing changes (i see the sytles on bundle but the system not recognise it)
by robitto 2015-03-05 at 18 20 26

robipresotto:
MacBook-Pro:mapbox-gl-native robipresotto$ git submodule status
 03869fd11e4bc619d3ba7e4f8a1f00d4612c90ea .mason (03869fd)
 25b1b7dff37a18151e3286144bc8013b432a8860 styles (25b1b7d)
 ab5a46ff7e970de5578df48a1e1f013bd5e1bd4e test/ios/KIF (v3.0.8-1-gab5a46f)
 6ac15153826ab26f778d61034119ffa8f97566ac test/suite (6ac1515)

DavidChouinard:
This issue still exists.

It seems the commit only dealt with making sure the Git styles submodule was included. That wasn't the original issue: the `styles/` directory is included, but the system crashes with `loading style failed: no such file or directory`.

1ec5:
I also ran into this issue while looking into #829. The bundled demo app gets away with copying styles/ itself directly into the main bundle directory, rather than under an intermediate MapboxGL.bundle.

Over to #1011 since this is a merged PR.

DavidChouinard:
Thank you. I'm not sure how related this is, but I know get a `SIGABRT` crash.

screen shot 2015-03-16 at 7 47 09 pm

1ec5:
I filed #1014 for that crash. It isn’t related to the bundle path, although with #1011 fixed, the default style can load without running into the crash.

DavidChouinard:
Thank you for looking into this.

(I get a black screen when not specifying a style)

img_4222

1ec5:
@DavidChouinard Even if you incorporate #1012? I’m seeing `bright-v7` load by default.

Let’s continue the discussion in #1011, since this PR was for a very different issue.

1ec5:
@DavidChouinard even with #1012, it’s going to be black unless you pass either a JSON stylesheet or a Mapbox access token. You need a token to get the default Bright style.

DavidChouinard:
@1ec5 Sorry for the delayed response. I rebuilt the latest version and things now work as expected.

Thank you.

PR #949, opened 2015-03-06
Feature: Dynamically localize compass text

friedbunny:
Draw the compass text dynamically based on the user's device language. This adds `CompassBlank.png` at the three necessary scales.

I don't know what the localization plans are (and being pre-beta it's surely not a high priority), so this hardcodes the different localized strings rather than messing with anything more involved. Once localization gets on the roadmap this will have to be revisited.

**Japanese/Chinese**

[Mapbox GL in Japanese](https://cloud.githubusercontent.com/assets/1198851/6517153/121bba26-c34d-11e4-88f0-b2be4bbe4d0f.png)

**English/Default**

[Mapbox GL in English](https://cloud.githubusercontent.com/assets/1198851/6517151/0d67af44-c34d-11e4-97ca-8ecb0c9859a5.png)

**Apple Maps in Japanese**

[Apple Maps in Japanese](https://cloud.githubusercontent.com/assets/1198851/6517155/1be9813c-c34d-11e4-8452-0956062b6d0d.png)

... not sure why I've been so fixated on compasses, but hopefully this is useful. I also tossed in `Compass@3x.png` for the hell of it.


friedbunny:
Oh, and this doesn't work in the simulator — I believe language is (mistakenly) always taken from the system, not from the iOS simulator.

incanus:
Oh, interesting... So this hasn't come up in past with our iOS SDK. The image is actually cribbed from MapKit by use of UIKit-Artwork-Extractor. I'm going to close here for now, but the PR will stick around and we can revisit. Do you have a good sense @friedbunny what Maps.app does here?

friedbunny:
iOS 8 Maps follows this strategy (AFAIK) — I took the blanks from MapKit.framework. :sweat_smile:

incanus:
Cool, much appreciated and :heart: the attention to detail. We don't yet have localization on the roadmap, but it will become increasingly important, including at the GL labels level. We'll always have this here, but I think closing for now rather than leaving open is a good move. 

I WON'T FORGET U LOCALIZED COMPASS

friedbunny:
Sure thing, this is definitely not a priority. Just one of those little touches that puts a twinkle in my eye and doesn't trod anybody's toes too terribly.

And since I'm at it, perhaps I'll add the missing `@3x` resources in another PR — if those aren't going to be revamped.

incanus:
> And since I'm at it, perhaps I'll add the missing `@3x` resources in another PR — if those aren't going to be revamped. If you have them from Apple, yeah. Per https://github.com/mapbox/mapbox-gl-native/issues/843 we are going to do our style sprites differently but if we're copying Apple resources, I'm all for grabbing their `@3x` versions as well.

1ec5:
#4783 made the compass text localizable using a similar strategy as in this PR (but slightly more robust). However, no actual localizations have been added yet.

PR #956, opened 2015-03-06
Move configure to scripts/ and update the docs

tmpsantos:
A small refactoring on the buildsystem.


tmpsantos:
@kkaefer please review

tmpsantos:
> Hrm, the configuration files don't conceptually belong into the mason_packages folder, since it's managed by mason itself. For a rebuild, you should run make clean, which also deletes the gypi files, which will then trigger mason to redownload the files. To make this easier, we could instead also introduce a make clean-config or make confclean to just remove all config files.

Make clean will not cause a re-download, only `make distclean` (but that would remove everything). To redownload you need a target that removes both config/*gyp and the mason_packages/ directory. That is why I was thinking that they share a link.

Issue #935, opened 2015-03-02
Integrate hybrid style


1ec5:
Per discussion with @incanus, this issue tracks adding the “hybrid” style to the example apps. See mapbox/mapbox-gl-styles#35, mapbox/mapbox-gl-js#1041.

incanus:
Idea here is parity with what devs expect of <code>-[MKMapView.mapType]</code>.

1ec5:
Depends on #939.

1ec5:
To your left, contours+labels. To your right, just labels:
![contours labels](https://cloud.githubusercontent.com/assets/1231218/6517809/9d44a882-c353-11e4-9a23-cfd015d4cdfd.png) ![labels](https://cloud.githubusercontent.com/assets/1231218/6517810/9d49ba52-c353-11e4-9a8d-5d57d7c31d3b.png)

1ec5:
Here’s what we’re getting out of the style .json: `http://a.tiles.mapbox.com/v4/mapbox.mapbox-terrain-v2/{z}/{x}/{y}.vector.pbf?access_token=…`. The vector glyph zoom level is always one lower than the raster tile zoom level.

1ec5:
At the initial viewport, `map.getState()` is 14.5. The `TileData`s are being created with z14 for the contours and labels and z15 for the satellite tiles, based on the return value of `Source::getZoom()`.

incanus:
@jfirebaugh I believe you said you were going to take a look at this? Want to self-assign, or just take a peek?

gundersen:
@incanus can we use Mapbox Terrain rather than Mapbox Outdoors in when we launch. Terrain gets us the wins of terrain w/ the wins of Mapbox Streets. 

@ajashton what are your thoughts here on the 1 style we should launch in GL w/ the iOS launch?

jfirebaugh:
The rendering errors (black labels, missing contours) are reproducible on any platform.

9897ae74ecd7a81f250ee791be41fdf20727785c fixes contour rendering. fddd27f6c4b46aa8b164d2a5a97eb35ae3ef4cdc fixes label rendering but I'm not sure if it's correct. @kkaefer?

incanus:
Splitting vector terrain out to https://github.com/mapbox/mapbox-gl-native/issues/964 to keep this ticket purely raster satellite + vector labels (i.e. "hybrid").

incanus:
This will come in in b2.

PR #966, opened 2015-03-09
refs #935: remove dispatch_once, favor property use, and return immutable styles

incanus:
Carryover from #935. 

Just a couple code style tweaks, unless I'm mistaken @1ec5. 

* Use a property instead of an instance variable. We tended to the latter a lot in the SDK because that's how the upstream project was, but tend to prefer interface/class extension properties (if applicable) now. 

* No need for `dispatch_once()`? 

* Return an `NSArray` for the styles, not the `NSMutableArray` that we track internally. 

If this checks out in CI and looks good @1ec5, feel free to merge.


1ec5:
Fair enough. I’ll merge once it’s green.

PR #928, opened 2015-03-10
refactor VectorTile to vend geometry

kkaefer:
See background in https://github.com/mapbox/mapbox-gl-native/issues/893#issuecomment-76472315. Basically the following classes are tightly coupled to PBF-specific vector tile geometry routines: 

* `TileParser`
* `LineBucket`
* `FillBucket`
* `SymbolBucket`

All of these step through individual geometries in vector tiles ([example](https://github.com/mapbox/mapbox-gl-native/blob/e1ff75a47aa3893219e0a9842829e5566f26449a/src/mbgl/renderer/symbol_bucket.cpp#L106-L112)) which ties the use of their structures to PBF binary-format tiles. We want to be able to also have "live tiles"  that back annotations, which are structured similarly to vector tiles (multi-layer, pre-simplified, able to be filtered for styling purposes) yet come from sources such as live-adding or GeoJSON parsing. 

The parser and buckets should ask for geometries in an intermediate format (probably [`Tile`](https://github.com/mapbox/mapbox-gl-native/blob/d36caee2a97e0e8a902eee637b22f3ea551744f6/src/mbgl/util/geojsonvt.hpp#L159-L176), which also has 4096 extents) and then do things with them.


src/mbgl/geometry/pbf_geometry.hpp
```diff
@@ -8,10 +8,10 @@
 
 namespace mbgl {
 
-class Geometry : private util::noncopyable {
+class PBFGeometry : private util::noncopyable {
 
 public:
-    inline explicit Geometry(pbf& data);
+    inline explicit PBFGeometry(pbf data);
```
kkaefer: I assume this was an intentional to avoid mutating the passed in parameter?

src/mbgl/map/geometry_tile.hpp
```diff
@@ -0,0 +1,62 @@
...
+public:
+    virtual FeatureType getType() const = 0;
+    virtual mapbox::util::optional<Value> getValue(const std::string& key) const = 0;
+    virtual GeometryCollection getGeometries() const = 0;
+};
+
+class GeometryTileLayer : public mbgl::util::noncopyable {
+public:
+    virtual std::size_t featureCount() const = 0;
+    virtual const util::ptr<const GeometryTileFeature> feature(std::size_t i) const = 0;
```
kkaefer: Returning a `const` type that is not a ref doesn't make sense, since it'll be copied anyway.

Can we use a `const GeometryTileFeature &` here instead of shared pointers?

src/mbgl/map/geometry_tile.hpp
```diff
@@ -0,0 +1,62 @@
...
+};
+
+class GeometryTile : public mbgl::util::noncopyable {
+public:
+    virtual const util::ptr<const GeometryTileLayer> getLayer(const std::string&) const = 0;
```
kkaefer: Same here.

src/mbgl/map/geometry_tile.cpp
```diff
@@ -0,0 +1,16 @@
...
+
+mapbox::util::optional<Value> GeometryTileFeatureExtractor::getValue(const std::string& key) const {
+    if (key == "$type") {
+        return Value(uint64_t(feature.getType()));
+    }
+
+    return feature.getValue(key);
+}
+
+template bool evaluate(const FilterExpression&, const GeometryTileFeatureExtractor&);
```
kkaefer: Can we make this a non-free function or rename it some way so that we don't have a "evaluate" function in the global mbgl namespace?

src/mbgl/map/vector_tile.hpp
```diff
@@ -1,117 +1,55 @@
...
 public:
-    VectorTile();
-    VectorTile(pbf data);
-    VectorTile& operator=(VectorTile&& other);
-
-    std::map<std::string, const VectorTileLayer> layers;
-};
+    VectorTile(pbf);
 
+    virtual const util::ptr<const GeometryTileLayer> getLayer(const std::string&) const;
```
kkaefer: This is not a virtual function, but it overrides a virtual function. Instead of specifying it as `virtual`, please use [`override`](http://en.cppreference.com/w/cpp/language/override).

src/mbgl/map/vector_tile.hpp
```diff
@@ -1,117 +1,55 @@
...
-
-std::ostream& operator<<(std::ostream&, const FeatureType& type);
-
-class VectorTileFeature {
+class VectorTileFeature : public GeometryTileFeature {
 public:
-    VectorTileFeature(pbf feature, const VectorTileLayer& layer);
+    VectorTileFeature(pbf, const VectorTileLayer&);
 
+    virtual FeatureType getType() const { return type; }
```
kkaefer: These are all non-virtual. Mark as `override`.

src/mbgl/geometry/pbf_geometry.hpp
incanus: Accident, I think — was going to take a review back through and improve passing by reference where applicable.

src/mbgl/map/geometry_tile.hpp
incanus: Reason for shared pointers was because `GeometryTileFeature` is an abstract interface, and returning `const` was because that's the `const`ness that was needed by the callers.

src/mbgl/map/geometry_tile.cpp
incanus: Could be improved — this was adapted from `master`: 

https://github.com/mapbox/mapbox-gl-native/blob/f82de98899a826bd3d5dc935c5a4e8a33c2d9b6b/src/mbgl/map/vector_tile.cpp#L176

src/mbgl/map/vector_tile.hpp
incanus: :+1: All clear on this now.

src/mbgl/geometry/pbf_geometry.hpp
jfirebaugh: I intentionally changed from a reference to a value, as `pbf` is intended to be passed by value and at the call site was being manually copied to a temporary. In any case, I'm about to push a commit that inlines `PBFGeometry` entirely.

incanus: Ah, cool.

src/mbgl/map/geometry_tile.hpp
jfirebaugh: I'm assuming we don't want to force implementations of `GeometryTileLayer` to pre-allocate or cache feature objects. `VectorTileLayer::feature` for instance creates them on the fly from the raw PBF data.

Agree about removing `const` though. I'll do that.

src/mbgl/map/geometry_tile.cpp
jfirebaugh: Should we change `FilterExpression` to work with `GeometryTileFeature`s rather than using template polymorphism?

incanus:
Much :bow: to @jfirebaugh; merging now.

PR #972, opened 2015-03-11
Refactoring of the log system

tmpsantos:
Refactoring of the logging system making it static. Also removed duplicated logic across the platforms we support.

From the usage point of view nothing really changed, except that now when writing tests, instead of setting a custom backend for the tests when we need to verify the log messages, we now add a Log::Observer.


tmpsantos:
@kkaefer: Addressed the review comment. Thanks for reviewing.

PR #975, opened 2015-03-11
Build with current iOS SDK

### LINKED ISSUE #974: package_ios.sh is requiring obsolete version of iOS SDK
The script at scripts/package_ios.sh is set for iOS SDK version 8.1 which is not available in the current version of Xcode. This should probably be updated to 8.2, or better yet, set to use the latest installed version.

1ec5:
`xcodebuild` does not require a specific SDK version, but Mason puts packages in a version-specific directory. So this change uses [the logic for `$MASON_PLATFORM_VERSION`](https://github.com/mapbox/mason/blob/67a20e92a93c555808b25e660d86243a3cf0b4df/mason.sh#L61) in mason.sh in mapbox/mason.

Fixes #974.

### DIFF
--- a/scripts/package_ios.sh
+++ b/scripts/package_ios.sh
@@ -6,7 +6,7 @@ set -u
 
 NAME=MapboxGL
 OUTPUT=build/ios/pkg
-IOS_SDK_VERSION=8.1
+IOS_SDK_VERSION=`xcrun --sdk iphoneos --show-sdk-version`
 LIBUV_VERSION=0.10.28
 
 function step { >&2 echo -e "\033[1m\033[36m* $@\033[0m"; }


incanus:
Ah, forgot about this part — I put that there. I think we should keep this hardcoded, otherwise it will change without our knowing as Travis changes their config.

incanus:
Also: http://docs.travis-ci.com/user/languages/objective-c/#Supported-OS-X-iOS-SDK-versions

1ec5:
OK, let’s remain in control of the SDK version and track Travis’ latest supported version manually. We’ll continue to hardcode the version, and we’ll bump it to 8.2 when Travis has upgraded.

kkaefer:
Why do we hardcode this?

incanus:
`package_os.sh` is used for packaging, so we should be in explicit control of the SDK that we package against after having tested it manually as well as had Travis testing against it. This shouldn't change automatically for the user based on what SDK they happen to have installed at the moment.

incanus:
Per chat we are going to go back to using the latest available local SDK to get some better test coverage as Travis lags.

PR #976, opened 2015-03-11
Relax sqlite version check to major version only

### LINKED ISSUE #968: sqliteVersionCheck breaks apps
I had an app in Testflight that ran just fine on iOS 8.1 but crashes consistantly on launch after updating the phone (iPhone 6) to iOS 8.2 (released today). Rebuild the app (using XCode 6.2) fixed the crash but didn't offer any insight into why the app crashed (nothing in the logs). When I debugged the app in iOS 8.1 Simulator, I ran into `sqliteVersionCheck` exception. After commenting out the version check, app ran just fine.

This shouldn't happen if my app was using the custom built SQLite library so I think what happened was that the app ended up using system provided version somehow (still trying to figure it out) and iOS 8.2 has a different version of SQLite so this version check failed under two situations:

1. app built with iOS 8.1 SDK running on iOS 8.2 device/simulator.
2. app built with iOS 8.2 SDK running on iOS 8.1 device/simulator.

I don't know what the right thing to do is but it would help non-critical version checks resulted in mere warnings instead of exceptions.

jfirebaugh:
iOS 8.2 updates the included SQLite version to 3.8.4, while iOS 7.0 up
to iOS 8.1 provides version 3.7.13. When compiling with the 8.2 SDK and
running on 8.1 or below, or vice versa, a strict check fails.

(It'll probably never fail in this configuration, but :shrug:)

Fixes #968.

### DIFF
--- a/platform/default/sqlite3.cpp
+++ b/platform/default/sqlite3.cpp
@@ -6,18 +6,11 @@
 
 // Check sqlite3 library version.
 const static bool sqliteVersionCheck = []() {
-    if (sqlite3_libversion_number() != SQLITE_VERSION_NUMBER) {
+    if (sqlite3_libversion_number() / 1000000 != SQLITE_VERSION_NUMBER / 1000000) {
         char message[96];
         snprintf(message, 96,
                  "sqlite3 libversion mismatch: headers report %d, but library reports %d",
                  SQLITE_VERSION_NUMBER, sqlite3_libversion_number());
-        throw std::runtime_error(message);
-    }
-    if (strcmp(sqlite3_sourceid(), SQLITE_SOURCE_ID) != 0) {
-        char message[256];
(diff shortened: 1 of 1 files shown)


jfirebaugh:
@kkaefer What do you think about this approach?

kkaefer:
I think it's the best bet we've got. I don't see a solution given that we don't actually know which SQLite version we're linking to.

dominik-weber:
Possibly a better way to solve this is to specify a hard coded minimum SQLite version instead of comparing to the header version. For example the sqlite3_open_v2 method was introduced in 3.5.0 (I think), and if your app *would* use this method you would end up crashing with sqlite versions < 3.5.0.

There sould not be any need to restrict newer versions, as the SQLite team seems to be working hard to achieve backward compatibility.

PR #977, opened 2015-03-11
Use iOS 8.2 SDK to package for iOS

martinnormark:
Maybe this is too early, but it successfully produced package against iOS 8.2 SDK so there was no need to install iOS 8.1 simulator.

### DIFF
--- a/MapboxGL.podspec
+++ b/MapboxGL.podspec
@@ -1,0 +1,33 @@
+Pod::Spec.new do |m|
+
+  m.name    = 'MapboxGL'
+  m.version = '0.1.0'
+
+  m.summary          = 'Open source vector map solution for iOS with full styling capabilities.'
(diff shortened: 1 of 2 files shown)


incanus:
Per https://github.com/mapbox/mapbox-gl-native/issues/974 we are awaiting Travis support for 8.2 before we move to this project-wide. Until then, just apply the change in 052cd8d locally for now.

kkaefer:
This is fixed in #975

incanus:
Upstream is https://github.com/travis-ci/travis-ci/issues/3373.

1ec5:
Given the fact that Travis is now two versions behind – 8.3 has GM’d – we’re going to go with the original approach of using the latest SDK (#975).

martinnormark:
:+1: Gutted to see Travis being two versions behind, though.

incanus:
Seriously, they've had nearly an hour already ;-)

PR #838, opened 2015-03-11
black flicker on iOS device rotation

kkaefer:
Rotate 90 degrees to repro.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,12 +4,15 @@
 *.actual.png
 *.diff.png
 *.pyc
+/android/debug
+/android/sdk
 /mason_packages
 /config/*.gypi
 /build
 /macosx/build
 /linux/build
 /ios/build
+/ios/app/build
 /test/build
 /test/node_modules
 /include/mbgl/shader/shaders.hpp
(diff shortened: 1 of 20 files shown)


bleege:
This looks awesome @kkaefer!

Total nitpicking here, but the only question is that on rotation it looks like the GL Canvas is white and so for a split second before the GL map renders on rotation completes its visible against the black background of iOS.  This is most noticeable when rotating from landscape back to portrait.  I think the easy solution would be to [paint it black](https://www.youtube.com/watch?v=u6d8eKvegLI).

incanus:
Yeah, we discussed this in chat yesterday @bleege. I may take a look at `UIView`/`UIViewController` resources to work around that, as MapKit doesn't have the problem. Also keep in mind not every view controller will default to a black view.

bleege:
Thanks for the insight @incanus!  BTW, I just looked at Google Maps on iOS and it has the same momentary white flicker on rotation too.

kkaefer:
I spent quite some time on triggering the view resize *during*, not *before* the animation, but couldn't find a good solution. One that comes to mind is adding a second subview behind the GLKView that is a static image of the map pre-transition, and keeping it centered with `UIViewContentModeCenter`, and setting the GLKView's background color to transparent.

PR #984, opened 2015-03-12
Get rid of printf and cout by using Log::*

### LINKED ISSUE #613: Convert all stdout/stderr prints to use Log functions
There is still some code that writes to stdout/stderr instead of the Log functions. For example `tile_parser.cpp`

This means important error messages never make it to the log on Android and makes debugging harder.

So all code using some sort of print function needs to be found and changed to Log.

Also anywhere that throws a C++ exception needs to also print a Log::Error before it throws in case we don't catch the exception and we crash instead.

tmpsantos:
Use our logging system that will route the message accordingly on
the target platform.

Fixes #613

### DIFF
--- a/linux/main.cpp
+++ b/linux/main.cpp
@@ -17,7 +17,7 @@ GLFWView *view = nullptr;
 
 void quit_handler(int) {
     if (view) {
-        fprintf(stderr, "waiting for quit...\n");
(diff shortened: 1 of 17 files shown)


tmpsantos:
@ljbade PTAL

ljbade:
@jfirebaugh looks good to me, merge?

PR #993, opened 2015-03-12
v8 Omnibus 2

tmcw:
Downstream issue of https://github.com/mapbox/mapbox-gl-style-spec/pull/264

* [x] background-image -> background-pattern
* [x] line-image -> line-pattern
* [x] symbol-min-distance -> symbol-spacing
* [x] text-font -> array
  * [ ] update test-suite
* [ ] types
* [x] Color operations https://github.com/mapbox/mapbox-gl-style-spec/issues/38


tmcw:
/cc @jfirebaugh @kkaefer would love some assistance on this patch so that it doesn't get merge-pwned by changes in the master branch. This is needed to close https://github.com/mapbox/mapbox-gl-online/issues/1300 in gl-online and has completed twin issues in gl-style-spec and gl-js: I'm trying to implement style v8 compatibility and have a goal of:

* for the `text-font` property, it should now be an array of strings instead of a comma-separated string
* the least invasive way to implement this would be to comma-join the array as soon as it enters gl-native
* but the [template approach](https://github.com/mapbox/mapbox-gl-native/pull/993/files#diff-ba1fd90e858f344f93b48f01045f63efR884) means that we'll have to do some sort of templating magic to make this flow different than other arrays

kkaefer:
@tmcw looks like you figured it out? The only thing missing is changing the style fixtures since the style format changed.

tmcw:
Yep, forgot to post back here - sorry. The text-font issue is now fixed, but we then came to the agreement that we should go whole-hog and implement types too, which I'll do first in style-spec.

tmcw:
In the interest of shipping v8 sooner rather than later, does anyone @mapbox/gl want to adopt this side of the (spec, js, api, native) taskcombo?

lbud:
https://github.com/mapbox/mapbox-gl-native/pull/1390 …maybe

lbud:
Closing in favor of #1390

Issue #990, opened 2015-03-12
get rid of white side bars on device rotation


incanus:
Piggybacking off of https://github.com/mapbox/mapbox-gl-native/pull/838's fix in https://github.com/mapbox/mapbox-gl-native/commit/8cc1c00a96d5afe531b33cf3ed62c4ec910ecd75, we should figure out if we need to fake a view snapshot or similar in order to prevent unsightly bars during orientation changes. 

Per @bleege this might not be essential. 

> BTW, I just looked at Google Maps on iOS and it has the same momentary white flicker on rotation too.

incanus:
@bleege Want to take a crack at this?

bleege:
Sure!

bleege:
Should be good to go... waiting on Travis build to Green light it.

incanus:
Oh wow, it was as simple as this?

bleege:
Woo hoo Travis just green barred!  Time to merge!

incanus:
Fixed in #994.

PR #994, opened 2015-03-12
paint it black

bleege:
For #990.


incanus:
This is a beautiful thing and works a treat.

Issue #957, opened 2015-03-06
GlyphAtlas needs a better way to keep track of tiles


kkaefer:
The GlyphAtlas currently tracks glyphs by tile ID and removes the glyph if no tile exists that uses the glyph anymore. However, there can be multiple tiles with the same ID, but from different sources. Instead of tracking per ID, we should e.g. use the TileData pointer as a key.

ansis:
A part of the problem in #832 might be that the glyph atlas isn't properly cleaned up when sources are destroyed. In -js, GlyphAtlas is now part of the Style object, so whenever the style changes the atlas is blown away and we start from scratch.

incanus:
If I'm understanding correctly from https://github.com/mapbox/mapbox-gl-native/issues/832#issuecomment-77624088, the GlyphAtlas could be garbling text because of this. This seems to be in effect anytime there is more than one vector source, which is impacting https://github.com/mapbox/mapbox-gl-native/issues/893 annotations as well. 

Adding to iOS beta milestone.

incanus:
Keeping this on beta 1 milestone. @jfirebaugh I know you're gonna take a look — this is top of my list for blocking annotations right now.

jfirebaugh:
⇢ #995

PR #995, opened 2015-03-12
Working towards GlyphAtlas safety

jfirebaugh:
Refs #957

First part: move atlas ownership to Style. This follows gl-js and just makes sense -- whenever the style changes the atlases should be blown away.

@kkaefer, is it safe yet to pass Style by reference into `Source::update` ⇢ `VectorTileData` ⇢ `TileParser` as I've done here? I'm still not clear on what parts are running in a different thread that's not guaranteed to be shut down before the Style is deallocated. If this isn't safe I can fall back to a `shared_from_this` solution.

Second part: Index glyph use by tile pointer value, not tile ID. There can be multiple tiles with the same ID, but from different sources.

### DIFF
--- a/include/mbgl/map/map.hpp
+++ b/include/mbgl/map/map.hpp
@@ -23,7 +23,6 @@
 namespace mbgl {
 
 class Painter;
-class GlyphStore;
 class LayerDescription;
 class Sprite;
 class Style;
@@ -33,9 +32,6 @@ class StyleSource;
 class TexturePool;
 class FileSource;
 class View;
-class GlyphAtlas;
-class SpriteAtlas;
-class LineAtlas;
 class Environment;
 
 class Map : private util::noncopyable {
@@ -153,7 +149,6 @@ private:
     void resize(uint16_t width, uint16_t height, float ratio = 1);
     void resize(uint16_t width, uint16_t height, float ratio, uint16_t fbWidth, uint16_t fbHeight);
 
-    util::ptr<Sprite> getSprite();
     uv::worker& getWorker();
 
     // Checks if render thread needs to pause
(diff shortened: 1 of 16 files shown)


incanus:
Awesome, this totally solves my use case (which you can try via https://gist.github.com/incanus/43d0faa730347a3a5d01 BTW; start in DC at `z10` or higher).

kkaefer:
>  is it safe yet to pass Style by reference into `Source::update` ⇢ `VectorTileData` ⇢ `TileParser` as I've done here?

I don't think it is right now. If the user changes the Stylesheet, we'll destruct the current stylesheet and swap it with another one. This is the reason we've held the Style object as a shared_ptr so far. However, this isn't a good solution either. Instead, we should abort all tile parsing (not sure if there's a good way to do this quickly) since we won't need the old parsed tiles anyway when switching the style.

mb12:
The following approach should work in the current start/stop scheme.

We can store a weak Style pointer. Before using it any time convert it to a shared pointer  using weak_ptr.lock().  If the Style object is destroyed in Map::stop, this conversion would fail and return a nullptr. However this also means that any data member in Style object should not be cached in other classes.  Any time a data member of Style needs to be accessed, it would have to be done via .lock() on this weak_ptr.

jfirebaugh:
I'd prefer to figure out how to abort all asynchronous operations before deallocating the style, rather than a weak-pointer based solution.

For now I just reverted to shared pointers and `shared_from_this()`.

jfirebaugh:
Build failure is likely unrelated to this change. Ticketed: #1001, going to merge.

jfirebaugh:
I spoke too soon, the glyph atlas ownership changes looked like they were destabilizing. I backed them out in 0c985c024b4db6c8063648d493830c67ebff50bc and will open a refactoring ticket for post-beta.

incanus:
Just to clarify, because I wanted to be sure, backing out that change in https://github.com/mapbox/mapbox-gl-native/commit/0c985c024b4db6c8063648d493830c67ebff50bc doesn't affect the bug fix for my setup, which I believe is addressed by https://github.com/mapbox/mapbox-gl-native/commit/6325f1e25a47f1d8afa35aee0db1327c21466b67 and remains in `master`.

PR #997, opened 2015-03-13
Refs #986: Check TRAVIS_SECURE_ENV_VARS before iOS code signing

### LINKED ISSUE #986: Disable use of secure environment variables in non-organization-member pushes
Testing pull requests on Travis from non-organization members fails: https://travis-ci.org/mapbox/mapbox-gl-native/builds/54087275

This is because our build script depends on the existence of the secure variables, but they are disabled for pushes originating from other users. We should detect the presence of these variables and skip all parts of the build that requires these.

friedbunny:
I'm not aware of a way I can test this, so please give it an extra thorough look. It does pass linting, for what that's worth: http://lint.travis-ci.org/

Moved the OpenSSL certificate stuff to `scripts/travis_ios/add-key.sh`.

Potentially fixes #986.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -72,18 +72,13 @@ install:
 before_script:
  # Set the core file limit to unlimited so a core file is generated upon crash
  - ulimit -c unlimited -S
- # begin iOS code signing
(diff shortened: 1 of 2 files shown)


kkaefer:
Thanks for the code. Looks like we can't fix this as long as we're running Android tests on EC2.

friedbunny:
In that case, I'll close this and it can be revisited if Android starts behaving.

ljbade:
Hmm yeah I guess this is another motivation to ditch EC2. I will add this to the list of work items for Android beta to see if we can find a good solution.

Issue #621, opened 2014-11-20
test on iPhone 6 Plus (non-integer scale factor)


incanus:
> Apps that use OpenGL or Metal for rendering can opt out of the scaling stage and render directly into a renderbuffer that has the native screen dimensions. All you have to do is to set the `contentScaleFactor` of your GLKView to the value of `UIScreen.nativeScale` (the OpenGL ES/Metal project templates in Xcode do this by default). This is an important performance optimization as more pixels generally means lower framerates.

http://oleb.net/blog/2014/11/iphone-6-plus-screen/

Does anyone have one of these? @mapbox/gl

incanus:
Interesting — we also fall down on sprites, presumably because we are defaulting to `UIScreen.mainScreen().scale`, which is `3.0` and doesn't map to our `1.0` or `2.0` expectations. 

![screen shot 2015-02-10 at 2 00 06 pm](https://cloud.githubusercontent.com/assets/17722/6137393/5281f756-b12d-11e4-856f-b3cd217cf9b8.png)

incanus:
Sprites issue moved to new ticket; this ticket is about increasing GL performance per the original API tweak, which is done above & pending Travis.

PR #843, opened 2015-03-16
sprite images  missing on > 2.0 scale factors

kkaefer:
Carryover from https://github.com/mapbox/mapbox-gl-native/issues/621. Sprites are falling down on iPhone 6 Plus and not sure why. We are doing things right AFAICT with `1x/2x` sprite sheets etc. by just looking for `scale > 1` and using `@2x` in that case. But perhaps something in the point shader is making the wrong assumption in some math. 

This can easily be tested in iPhone 6 Plus simulator. 

/cc @ljbade in case this is also an issue on Android hidpi. 

screen shot 2015-02-10 at 2 00 06 pm


mb12:
I have an older git repo. Therein SpriteAtlas::bind uses  GL_CLAMP_TO_EDGE.   The latest repo uses GL_REPEAT.

MBGL_CHECK_ERROR(glTexParameteri(GL_TEXTURE_2D, GL_TEXTURE_WRAP_S, GL_CLAMP_TO_EDGE));
        MBGL_CHECK_ERROR(glTexParameteri(GL_TEXTURE_2D, GL_TEXTURE_WRAP_T, GL_CLAMP_TO_EDGE));

ljbade:
Hmm I will take a look into this. Android can have some very unusual pixel ratios. Your analysis seems to line up with the behaviour I had been seeing.

ljbade:
Currently I don't get any POI icons at all on my HTC One which is 3.0x.

PR #1012, opened 2015-03-16
Look inside MapboxGL.bundle when present

### LINKED ISSUE #1011: MGLMapView looks for bundled styles inside styles/ instead of MapboxGL.bundle
If you follow the setup instructions in the README, you wind up adding a MapboxGL.bundle to your project; a styles/ directory in that bundle contains all the bundled style JSONs. But `MGLMapView` looks for styles/ directly beneath the main bundle, rather than in MapboxGL.bundle.

Previously reported in #852, #969, etc.

1ec5:
Fixes the use of bundled styles when incorporating the library into a project according to the README. Fixes #1011.

### DIFF
--- a/platform/darwin/asset_root.mm
+++ b/platform/darwin/asset_root.mm
@@ -2,13 +2,22 @@
 
 #include <mbgl/platform/platform.hpp>
 
+@interface MGLApplicationRootBundleCanary : NSObject
+@end
+
+@implementation MGLApplicationRootBundleCanary
+@end
+
 namespace mbgl {
 namespace platform {
 
 // Returns the path to the root folder of the application.
 const std::string &assetRoot() {
     static const std::string root = []() -> std::string {
-        NSString *path = [[[NSBundle mainBundle] resourceURL] path];
+        NSString *path = [[NSBundle bundleForClass:[MGLApplicationRootBundleCanary class]] pathForResource:@"MapboxGL" ofType:@"bundle"];
+        if (!path) {
+            path = [[[NSBundle mainBundle] resourceURL] path];
+        }
         return {[path cStringUsingEncoding : NSUTF8StringEncoding],
                 [path lengthOfBytesUsingEncoding:NSUTF8StringEncoding]};
     }();


platform/darwin/asset_root.mm
```diff
@@ -2,13 +2,22 @@
...
 namespace platform {
 
 // Returns the path to the root folder of the application.
 const std::string &assetRoot() {
     static const std::string root = []() -> std::string {
-        NSString *path = [[[NSBundle mainBundle] resourceURL] path];
+        NSString *path = [[NSBundle bundleForClass:[MGLApplicationRootBundleCanary class]] pathForResource:@"MapboxGL" ofType:@"bundle"];
+        if (!path) {
+            path = [[[NSBundle mainBundle] resourceURL] path];
+        }
```
incanus: We did used to have something like this before, a fallback for our own test app. We should probably do something really drastic here to warn the user, since frequently we will see support around people not having included the bundle properly. But we need to strike a balance between this and not hampering our test app, either. 

Basically, "when present" should be 100% of the time when used as a third-party library, and we should scream about it a little more when this is not the case.

1ec5:
There's nearly identical logic in `MGLMapView`, which we should probably fold into here. How should we distinguish between the demo app and other apps? A cleaner solution would be to make the demo app put styles/ in MapboxGL.bundle too.

incanus:
> A cleaner solution would be to make the demo app put styles/ in MapboxGL.bundle too.

This has been my thought in the past, but the demo app has other exceptions such as linking against component libraries of GL and not `libMapboxGL.a`. 

@kkaefer Do you have any thoughts on this? I say don't let the demo app be an exception to the rule because of this need for logic branching to handle resources differently.

ljbade:
Ideally the demo app should follow the same logic as end user apps if possible.

kkaefer:
> but the demo app has other exceptions such as linking against component libraries of GL and not libMapboxGL.a.

We're doing this to get faster builds so we don't have to create the massive libMapboxGL.a library first and then statically link it into the app again.

As far as using MapboxGL.bundle goes, I'm all for it.

1ec5:
I started to look at making the iOS and OS X demo apps use MapboxGL.bundle as well but ran into problems with gyp’s project generator. Per offline discussion w/ @incanus, we’re going to punt on making the demo apps use the bundle, which means this PR is good to go as soon as the build finishes.

incanus:
> We're doing this to get faster builds so we don't have to create the massive libMapboxGL.a library first and then statically link it into the app again.

:+1: I knew there was something. 

>  this PR is good to go as soon as the build finishes

Cool, will review.

incanus:
Just got tests to pass; merging.

PR #1015, opened 2015-03-17
Scope threads to an Environment

tmpsantos:
By creating an Environment::Scope at the beginning of the thread, it will automatically register it to the Environment during its lifespan. The Main thread is registered automatically at startup.

Threads running within an Environment::Scope can at any moment query for the Environment and assert what is the current thread, to make sure the code is being executed on the right thread.

And finally, log messages will now show the Environment ID and the current thread name.

### DIFF
--- a/.clang-format
+++ b/.clang-format
@@ -2,7 +2,7 @@ Standard: Cpp11
 IndentWidth: 4
 AccessModifierOffset: -4
 UseTab: Never
-BinPackParameters: true
+BinPackParameters: false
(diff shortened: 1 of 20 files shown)


tmpsantos:
Argh, ignore it for now, I forgot to push the change that make it return references.

tmpsantos:
@kkaefer PTAL, I'm using reference as much as I can, but not on the ThreadInfo struct because I need to be able to store env = nullptr for the Main thread (that doesn't run on a Environment::Scope).

Issue #992, opened 2015-03-12
flesh out annotation add/remove client API


incanus:
This is the API where platforms (i.e. iOS `MGLMapView` internals) talk to core for manipulating point and shape annotations. 

## Points

* Set default point annotation symbol name (so system falls back to e.g. a red filled Mapbox marker)
* Add single point annotation as `LatLng` with optional symbol name (from style sprite sheet) upfront
* Add multiples of same (more performant than individual adds in a loop)
* Remove single point annotation
* Remove multiples (more performant than individual removes in a loop)
* Tell single point annotation to move to top in z-order (e.g. selection state)

## Shapes

* Add single shape annotation as `std::vector<LatLng>` and draw properties (stroke width, stroke color, fill color) upfront. Need a way to specify interior polygon holes. 
* Add multiples of same
* Remove single shape annotation
* Remove multiples

## Both

* Obtaining annotations in a given `LatLng` bounding box (enables e.g. `-annotationsInMapRect:`)
* Obtaining bounding box fully encompassing all currently-added annotations (enables e.g. `annotationVisibleRect`)
* Neither of these take symbology into account, just `LatLng`, meaning zooming to exactly a bounding box might have a marker partway off the edge of the screen. 

## TBD for iOS b1

* Allow explicit relative ordering of shape annotations within shape annotation layer (e.g. `-insertOverlay:aboveOverlay:` or `-insertOverlay:atIndex:`)
* Query with callback for updated symbol name each time point annotation becomes visible again, to give opportunity to update? 
* Query with callback for updated stroke/fill properties in same way? 

## iOS b2 or beyond

* Interlacing of shape overlays with roads & labels base map layers (or other arbitrary layers). This is like MapKit's `-addOverlay:level:` with values like `MKOverlayLevelAboveRoads` or `MKOverlayLevelAboveLabels`. In our case, this would more likely be a style layer name passed as a string. Also enables e.g. `-overlaysInLevel:`. 

/cc @1ec5

1ec5:
The plan is to have `MGLMapView` track `MGLAnnotationView`s in a dictionary, similar to what I had before 9c46c6c0e2726704f929f9c139c4b3de2b4dfcd9, but using string UUIDs as keys instead of `MGLAnnotation`s. That’ll allow us to support multiple annotations at the same location. Methods such as `-addAnnotation:` and `-addAnnotations:` are well positioned for the C++ API @incanus is creating.

incanus:
[...]

PR #1018, opened 2015-03-17
refs #893 #992: point annotations API

incanus:
This is Phase One of the point annotations API — adding, removing, querying, and setting sprite-based symbology for point annotations. 

@jfirebaugh If you have some time would you mind sharing some :eyes: — especially since this relates to the `GeometryTile` refactor. 

Two things I want to do to this next, but input is welcome: 

* [x] Get rid of the pointer casts per https://github.com/mapbox/mapbox-gl-native/commit/fbd55115b2383d0c5a6d057fd3d539bdeb99e12e#commitcomment-10224463. 

* [x] See about getting rid of `LiveTileParser` entirely and defaulting to `TileParser`. This was created in order to get a `createBucket()` and address troubles with the `getLayer()` API, but I think that's obsolete now.


src/mbgl/map/live_tile.hpp
```diff
@@ -0,0 +1,53 @@
+#ifndef MBGL_MAP_LIVE_TILE
+#define MBGL_MAP_LIVE_TILE
+
+#include <map>
+
+#include <mbgl/map/geometry_tile.hpp>
+
+namespace mbgl {
+
+class LiveTileFeature : public GeometryTileFeature, private util::noncopyable {
```
jfirebaugh: These classes are derived from `util::noncopyable` already so you don't need to re-inherit it here.

src/mbgl/map/live_tile.hpp
```diff
@@ -0,0 +1,53 @@
...
+public:
+    LiveTileFeature(FeatureType, GeometryCollection, std::map<std::string, std::string> = {{}});
```
jfirebaugh: Please name the final parameter here; it's not obvious from the type what it is.

src/mbgl/map/live_tile.hpp
incanus: cb47993

src/mbgl/map/live_tile.hpp
incanus: 543b657

src/mbgl/map/annotation.cpp
```diff
@@ -0,0 +1,191 @@
...
+
+    for (auto& annotationID : ids) {
+        auto annotation_it = annotations.find(annotationID);
+        if (annotation_it != annotations.end()) {
+            auto& annotation = annotation_it->second;
+            for (auto& tile_it : annotationTiles) {
+                auto features_it = annotation->tileFeatures.find(tile_it.first);
+                if (features_it != annotation->tileFeatures.end()) {
+                    auto layer = tile_it.second.second->getLayer(util::ANNOTATIONS_POINTS_LAYER_ID);
+                    auto liveLayer = std::static_pointer_cast<LiveTileLayer>(layer);
```
incanus: Getting this one now.

incanus: a878247

src/mbgl/map/tile_parser.hpp
```diff
@@ -31,12 +31,10 @@ class StyleLayoutSymbol;
...
 class VectorTileData;
 class Collision;
-class TexturePool;
 
-class TileParser : private util::noncopyable
-{
+class TileParser : private util::noncopyable {
 public:
-    TileParser(const std::string& rawData,
+    TileParser(const GeometryTile* geometryTile,
```
jfirebaugh: `const GeometryTile&`

src/mbgl/map/tile_parser.hpp
```diff
@@ -48,11 +46,11 @@ class TileParser : private util::noncopyable
 public:
     void parse();
 
-private:
+protected:
```
jfirebaugh: Why did this change?

incanus: This can go back; was related to the now-gone `LiveTileParser`.

incanus: 1f761ac

src/mbgl/map/tile_parser.hpp
incanus: d909e81

src/mbgl/map/tile_parser.hpp
jfirebaugh: You got one of them, but this one's still present. Also the added `virtual` below can be reverted.

incanus: d34caf6

include/mbgl/map/map.hpp
```diff
@@ -37,6 +37,9 @@ class GlyphAtlas;
 class SpriteAtlas;
 class LineAtlas;
 class Environment;
+class AnnotationManager;
+
+typedef std::vector<LatLng> AnnotationSegment;
```
jfirebaugh: This doesn't appear to be used by this header; should move to `annotation.hpp`.

src/mbgl/map/annotation.cpp
```diff
@@ -0,0 +1,190 @@
...
+        if (id.z == z) {
+            if (id.x >= nwTile.x && id.x <= seTile.x && id.y >= nwTile.y && id.y <= seTile.y) {
+                if (id.x > nwTile.x && id.x < seTile.x && id.y > nwTile.y && id.y < seTile.y) {
+                    // trivial accept; grab all of the tile's annotations
+                    std::copy(tile.second.first.begin(), tile.second.first.end(), std::back_inserter(matchingAnnotations));
+                } else {
+                    // check tile's annotations' bounding boxes
+                    std::copy_if(tile.second.first.begin(), tile.second.first.end(),
+                                 std::back_inserter(matchingAnnotations), [&](const uint32_t annotationID) -> bool {
+                        printf("checking annotation %i\n", annotationID);
```
jfirebaugh: Remove this debug output.

include/mbgl/map/map.hpp
incanus: The reasoning here is because the shapes API will have it; you pass vectors of `AnnotationSegment` objects to represent lines in a shape.

include/mbgl/util/geo.hpp
```diff
@@ -19,6 +19,14 @@ struct ProjectedMeters {
         : northing(n), easting(e) {}
 };
 
+struct BoundingBox {
```
jfirebaugh: `LatLngBounds`, to match gl-js.

src/mbgl/map/annotation.cpp
incanus: 98be4b1

include/mbgl/map/annotation.hpp
```diff
@@ -0,0 +1,71 @@
+#ifndef MBGL_MAP_ANNOTATIONS
+#define MBGL_MAP_ANNOTATIONS
+
+#include <mbgl/map/map.hpp>
```
jfirebaugh: Can this be a forward declaration instead of an include?

include/mbgl/map/annotation.hpp
```diff
@@ -0,0 +1,71 @@
...
+};
+
+class Annotation : private util::noncopyable {
+public:
+    Annotation(AnnotationType, std::vector<AnnotationSegment>);
+
+    LatLng getPoint() const { return geometry[0][0]; }
+    BoundingBox getBoundingBox() const { return bbox; }
+
+public:
```
jfirebaugh: Do these need to be public?

include/mbgl/map/map.hpp
```diff
@@ -147,6 +159,7 @@ class Map : private util::noncopyable {
 
     inline const TransformState &getState() const { return state; }
     inline std::chrono::steady_clock::time_point getTime() const { return animationTime; }
+    inline util::ptr<AnnotationManager> getAnnotationManager() const { return annotationManager; }
```
jfirebaugh: It doesn't look like shared ownership should be necessary here; `AnnotationManager&` is around for the lifetime of the Map, so return type can be `AnnotationManager&`.

src/mbgl/map/annotation.cpp
```diff
@@ -0,0 +1,190 @@
...
+Annotation::Annotation(AnnotationType type_, std::vector<AnnotationSegment> geometry_)
+    : type(type_),
+      geometry(geometry_) {
+    if (type == AnnotationType::Point) {
+        bbox = BoundingBox(getPoint(), getPoint());
+    } else {
+        LatLng sw, ne;
+        for (auto segment : geometry) {
+            for (auto point : segment) {
+                if (point.latitude < sw.latitude) sw.latitude = point.latitude;
```
jfirebaugh: Extract this to `BoundingBox::extend(const LatLng&)`.

src/mbgl/map/annotation.cpp
```diff
@@ -0,0 +1,190 @@
...
+    return matchingAnnotations;
+}
+
+BoundingBox AnnotationManager::getBoundingBoxForAnnotations(std::vector<uint32_t> ids) const {
+    LatLng sw, ne;
+    for (auto id : ids) {
+        auto annotation_it = annotations.find(id);
+        if (annotation_it != annotations.end()) {
+            auto bbox = annotation_it->second->getBoundingBox();
+            if (bbox.sw.latitude < sw.latitude) sw.latitude = bbox.sw.latitude;
```
jfirebaugh: Reuse `BoundingBox::extend(const LatLng&)` here.

include/mbgl/util/geo.hpp
incanus: 78cbab9

include/mbgl/map/annotation.hpp
incanus: This leads to not seeing `AnnotationSegment`, and that can't be forward declared because it is used in `getPoint()`. That method can't be moved inside of `annotation.cpp` because then you get a redefinition error between the forward declaration of `AnnotationSegment` and the one brought into `annotation.cpp` with `#include <mbgl/map/map.hpp>`. This went really deep and doing it like this seemed the only way to keep things happy.

jfirebaugh: The underlying problem is the cyclic dependency between `AnnotationManager` and `Map`. It looks like you can break this by removing `Map& map` from `AnnotationManager` and passing the max zoom to `AnnotationManager::getAnnotationsInBounds` and `AnnotationManager::addPointAnnotations`.

include/mbgl/map/map.hpp
incanus: That requires the member variable to move from `util::ptr<AnnotationManager>` to `AnnotationManager`, and that can't be forward declared, which then gets to the same cycle with `map.hpp` and `annotation.hpp` as in https://github.com/mapbox/mapbox-gl-native/pull/1018/files#r26625693.

src/mbgl/map/annotation.cpp
incanus: :+1:

include/mbgl/map/annotation.hpp
incanus: This is weird from an API point of view — clients like Cocoa would have to query, then pass the max zoom, when it doesn't appear to need to be an argument of querying bounds for annotations (they are independent of zoom and an implementation detail).

include/mbgl/map/map.hpp
jfirebaugh: You don't have to change the member variable:

```
inline AnnotationManager& getAnnotationManager() const { return *annotationManager; }
```

include/mbgl/map/annotation.hpp
jfirebaugh: I thought clients would use the API provided by `Map`?

include/mbgl/map/annotation.hpp
incanus: 0462b63

include/mbgl/map/map.hpp
incanus: Ohhhh, perfect. Right.

incanus: 5551f0a

src/mbgl/map/annotation.cpp
incanus: This will need updated to take a shape bounds later, but good for now.

incanus: 6eea1ac

include/mbgl/map/annotation.hpp
incanus: Ohhhh, got it.

src/mbgl/map/annotation.cpp
incanus: efdeb80

include/mbgl/map/annotation.hpp
incanus: 7d593a2

include/mbgl/map/map.hpp
incanus: 7d593a2

incanus:
→ https://github.com/mapbox/mapbox-gl-native/pull/1021

PR #381, opened 2015-03-17
fill-opacity not affecting outline

jfirebaugh:
Should `fill-opacity` be applied to the fill outline? It looks like it is on js but not native.

JS:

image

Native:

image

This is the `fill-opacity/literal` test in the test suite.

### DIFF
--- a/src/mbgl/renderer/painter.cpp
+++ b/src/mbgl/renderer/painter.cpp
@@ -87,6 +87,9 @@ void Painter::setup() {
     // Stencil test
     MBGL_CHECK_ERROR(glEnable(GL_STENCIL_TEST));
     MBGL_CHECK_ERROR(glStencilOp(GL_KEEP, GL_KEEP, GL_REPLACE));
+
+    // Depth test
+    glDepthFunc(GL_LEQUAL);
 }
 
 void Painter::setupShaders() {
@@ -193,7 +196,6 @@ void Painter::setOpaque() {
     if (pass != RenderPass::Opaque) {
         pass = RenderPass::Opaque;
         MBGL_CHECK_ERROR(glDisable(GL_BLEND));
-        depthMask(true);
     }
 }
 
@@ -201,7 +203,6 @@ void Painter::setTranslucent() {
     if (pass != RenderPass::Translucent) {
         pass = RenderPass::Translucent;
         MBGL_CHECK_ERROR(glEnable(GL_BLEND));
-        depthMask(false);
     }
 }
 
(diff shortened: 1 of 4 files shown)


ansis:
Yep, fringeline is for antialiasing. Your hunch sounds right to me. It looks like `fringline` is always false, but even when that is fixed there is a darker edge. It's probably a clipping issue as you suggested.

jfirebaugh:
Yeah, 1f1e07d5e3c636be376a9d121636022fe79a9e3b fixes fringeline always false.

Is there a way we can use the depth buffer to avoid double drawing?

ansis:
@jfirebaugh yes, that should be possible. I don't know offhand, but I can look into it if you want.

jfirebaugh:
That would be great, thank you.

ansis:
Changing [this line](https://github.com/mapbox/mapbox-gl-native/blob/ae7bd9ac6ac12b37da0272ea7d1836f18a78ec10/src/mbgl/renderer/painter.cpp#L204) to `depthMask(true)` fixes this. It enables writing depth values for transparent fragments to the depth buffer. Transparent fills get written to the depth buffer and can be used to drop overlapping antialiasing fragments later.

@kkaefer are there performance concerns about enabling depth buffer writing for transparent layers?
would it be better to enable it for fill layers but disable it for lines and symbols?

jfirebaugh:
In general we want transparent layers to overdraw, right? Seems like this should be specific to fill-fill + fill-fringe.

ansis:
> In general we want transparent layers to overdraw, right?

Yes, but if the depth values are right for other layers they would be allowed to overdraw anyways.

kkaefer:
Changing the value to false means that translucent layers will write to the depth buffer, blocking further attempts to write to that same position. While this works for opaque pixels, (potentially) translucent pixels are different because this disables overlapping translucent objects.

jfirebaugh:
@ansis I tried out that change locally and it breaks all kinds of stuff. Examples from the test suite:

image

image

ansis:
This is fixable by converting the `glDepthFunc` to `GL_LEQUAL` <= instead of `GL_LESS` < and fixing the depth values for symbol/halos.

It might be better to only enable it for fill layers. I'll do that today.

ansis:
https://github.com/mapbox/mapbox-gl-native/commit/92811a721eb585d432abc026438bf42a88e87569 turns on depth writes for transparent fills so that fill antialiasing can be clipped to areas outside the polygon.

PR #1021, opened 2015-03-18
refs #893 #992: point annotations API

incanus:
Take two of https://github.com/mapbox/mapbox-gl-native/pull/1018. Lots of review by @jfirebaugh, commits in to fix, and re-squashed.


incanus:
I had missed https://github.com/mapbox/mapbox-gl-native/commit/76f8a3d8468c623956c291ac7837e322b317a8f3 with the move away from `util::ptr<AnnotationManager>`, but it compiled earlier. 

I also had forgotten to clean up the map of tiles' annotations that is used for point-in-bounds routines when removing annotations.

incanus:
Tests pass; merging.

PR #1016, opened 2015-03-18
Crash when deleting buffers

kkaefer:
I'm seeing a fairly frequent crash bug that happens randomly during buffer deletion. The buffer IDs are valid and do exist.

What could be happening is that we have an OpenGL context loss that we're ignoring, which invalidates all of the buffers.

buffer hpp 2015-03-17 18-10-54

### DIFF
--- a/src/mbgl/map/map.cpp
+++ b/src/mbgl/map/map.cpp
@@ -61,6 +61,7 @@ using namespace mbgl;
 
 Map::Map(View& view_, FileSource& fileSource_)
     : env(util::make_unique<Environment>(fileSource_)),
+      scope(util::make_unique<EnvironmentScope>(*env, ThreadType::Main, "Main")),
       view(view_),
       transform(view_),
       fileSource(fileSource_),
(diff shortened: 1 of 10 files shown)


kkaefer:
~LiveTileData is run in the map thread

incanus:
Cross-ref to https://github.com/mapbox/mapbox-gl-native/issues/1022 — I will double-check that things are kosher here.

picciano:
+1 Awesome, no more crashing. Thanks!

src/mbgl/map/map.cpp
```diff
@@ -113,7 +127,6 @@ void Map::start(bool startPaused) {
 
         // Remove all of these to make sure they are destructed in the correct thread.
         style.reset();
-        workers.reset();
```
kkaefer: This crashes in Debug mode since the Worker threads are now joined from the wrong thread.

kkaefer: Fixed in 9d30d4042c3311d02568adbb38dd13fb82faf75d

PR #1024, opened 2015-03-18
Fix a few memory leaks

tmpsantos:
This branch also adds a reference Valgrind suppression file making it easier for further debugging.

### DIFF
--- a/platform/default/glfw_view.cpp
+++ b/platform/default/glfw_view.cpp
@@ -11,6 +11,7 @@ GLFWView::GLFWView(bool fullscreen_) : fullscreen(fullscreen_) {
 }
 
 GLFWView::~GLFWView() {
+    glfwDestroyWindow(window);
     glfwTerminate();
 }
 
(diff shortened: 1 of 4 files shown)


tmpsantos:
cc @kkaefer

platform/default/glfw_view.cpp
```diff
@@ -11,6 +11,7 @@ GLFWView::GLFWView(bool fullscreen_) : fullscreen(fullscreen_) {
 }
 
 GLFWView::~GLFWView() {
+    glfwDestroyWindow(window);
```
kkaefer: good call

platform/default/http_request_curl.cpp
```diff
@@ -75,7 +75,7 @@ class HTTPCURLContext : public HTTPContext<HTTPCURLContext> {
 
 public:
     // Used as the CURL timer function to periodically check for socket updates.
-    uv_timer_t *timeout = nullptr;
+    uv_timer_t timeout;
```
kkaefer: We should keep this as a pointer since we need to call `uv::close` to properly remove it from the loop handle later. When the destructor runs, we still have to wait for the loop to call our close callback until we can actually delete this timer object.

tmpsantos: :+1:

PR #1027, opened 2015-03-19
add iOS icon burning for Release builds

incanus:
This adds in icon burning using [iconoblast](https://github.com/mapbox/iconoblast) as a GYP build phase for the iOS testing app. It only applies to `Release` builds, but we need this in here so that we don't keep losing/forgetting it for internal builds of the app for testing. 

If this is problematic for anyone's workflow, post with some info and we can try to find other environment variables or something similar that can work around it.

### DIFF
--- a/ios/app/icon_burn.sh
+++ b/ios/app/icon_burn.sh
@@ -1,0 +1,36 @@
+export PATH=$HOME/bin:/usr/local/bin:/usr/bin
+
+#
+# don't burn during Debug builds
+#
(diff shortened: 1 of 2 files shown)


incanus:
This doesn't actually work (modifying the icons in the built product only) because GYP wants to separate it out as a separate, earlier build dependency and the product doesn't exist yet. So it has the effect of dirtying your working copy's icons, which is no good. 

Leaving this open and low priority for now, but I would like to get back to it to reduce overall work & risk of leaving this out as keeping rolling builds.

PR #1030, opened 2015-03-19
CocoaPod podspec for beta 1

### LINKED ISSUE #829: re-enable CocoaPods support

1ec5:
The podspec in this project was originally lifted from mapbox/mapbox-ios-sdk, which was structured and built quite differently. We could have the podspec build this repo as is, using scripts/package_ios.sh, but that would require developers to install the same version of Xcode as Travis is using (see #977). So this podspec instead pulls a prebuilt library. Much simpler.

Also package a LICENSE with the archive to satisfy a CocoaPods warning.

Fixes #829.

### DIFF
--- a/ios/MapboxGL.podspec
+++ b/ios/MapboxGL.podspec
@@ -8,25 +8,25 @@ Pod::Spec.new do |m|
   m.homepage         = 'https://www.mapbox.com/blog/mapbox-gl/'
   m.license          = 'BSD'
   m.author           = { 'Mapbox' => 'mobile@mapbox.com' }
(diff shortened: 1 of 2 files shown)


incanus:
Trying this out locally but hitting unrelated CocoaPods problems right now. When I'm past that, I think we are good here — `package_ios.sh` works well for me, and I'll try manually linking it in in the meantime.

incanus:
Looking :+1: with manual integration.

PR #1022, opened 2015-03-20
[not ready] Make Annotations API threadsafe

kkaefer:
This currently reaches into the innards of the Map objects from the main thread, which is not safe.

/cc @incanus


kkaefer:
Note that this change does not yet fix the actual issue of accessing the tile data and source objects from both threads.

kkaefer:
@incanus Can we remove the shared pointer use in `live_tile.*`?

incanus:
We have solid improvements across the board here, but still not a fix for the core threading problem.

incanus:
@kkaefer What's your ETA look like for the actual threading fixes here? I think we should get these improvements at they stand now over into `master` with all the improvements to passing by reference, `const`ness, and the like on the sooner side.

incanus:
Per voice with @kkaefer I'm going to merge these improvements and spin out a new ticket specific to the threading issues, which we don't have a good answer on yet. We are both meditating on this :bow:

incanus:
→ https://github.com/mapbox/mapbox-gl-native/issues/1069 for the thread work-to-be.

Issue #894, opened 2015-02-17
popups


incanus:
Popups, callouts, bubbles, whatever you want to call them, we need selection state and info bubbles on markers. 

Per @jfirebaugh (and I agree) we should do these for each platform because they are distinct and need to feel natural to the platform, such as being complex, even animated, views and being able to contain other platform-specific views. 

For this reason, I think we should only worry about this on Android and iOS. 

I've been very happy with [`SMCalloutView`](https://github.com/nfarina/calloutview), which we integrated into our iOS SDK, and I think we should continue to use that here.

incanus:
Also of note is that the performance implications of not doing this in GL core itself will be small. At least on iOS, only one callout is present at a time (single selection) and just simply tracks on the map.

ljbade:
On Android the OS uses GL anyway to render/draw overlaid views so unlikely to be slower then rendering in our GL context.

incanus:
@1ec5 has been starting to look into this, tracking using `CADisplayLink` but going to try @bleege's work on a "region is actively changing" delegate method in iOS.

incanus:
I'm grabbing this now to start on post-point annotations merge (#1021) since gestures, point-in-bounds, and platform-side callouts are all integrated.

incanus:
Annotation selection should work something like this. Note that unlike MapKit, the z-index of the selected marker doesn't change (that's in #991). 

![anigif-1426651756](https://cloud.githubusercontent.com/assets/17722/6702518/f5c40346-cce9-11e4-9636-df4d67015f14.gif)

This is done with something like d53fb94c86773ea45fb6df39c96cfbe161f4aa3e, which just does a dumb `44px` box around the touch for point-in-bounds querying, then iterates selection state through the close-by annotations. It doesn't take marker icon size into account; likely we'll want to make this box a bit larger and emphasize in the top direction. 

This is kind of like JS's `featuresAt()`, since point annotations' tiles work like an R-Tree.

incanus:
Summarizing state of popups. The tough nut to crack here is syncing pan and zoom — *including decelerated coasting* — between native iOS views and the GL view. We've tried this in three branches, with varying combos of `MGLAnnotation` protocol data objects or other placeholder markers. 

* [`877-display-link`](https://github.com/mapbox/mapbox-gl-native/compare/877-display-link) — using `CADisplayLink` callback to sync simple `UIImageView` markers with GL
[...]

PR #1037, opened 2015-03-20
[not ready] Cocoa annotations API

incanus:
Goal here is to take the best parts of these intermingled branches: 

* [x] [`877-display-link`](https://github.com/mapbox/mapbox-gl-native/tree/877-display-link)
* [x] [`877-callout`](https://github.com/mapbox/mapbox-gl-native/tree/877-callout)
* [x] [`877-point-callout`](https://github.com/mapbox/mapbox-gl-native/tree/877-point-callout)

Then to _get rid of them_ and shore up the definitive Cocoa model API here. 

/cc @1ec5


incanus:
The display link work is pretty straightforward. Correct me if wrong @bleege but there are two parts to the approach: 

* Be sure to add the views you're keeping in sync as subviews of the `GLKView`.
* As is usually the case, be sure to add the display link to the main run loop for `NSRunLoopCommonModes`. 

I've captured this branch's diff here since it's pretty light: 

https://gist.github.com/incanus/286b534c636c601a73f6

Now deleting `877-display-link`.

incanus:
`877-point-callout`'s whole, um, point, was to merge in an early sneak peek of the C++ annotations API into `877-callout`. Now that that is in `master`, these branches are identical, so removing `877-point-callout`.

incanus:
Ok, I have the important bits of `877-callout` ported over, with some discussion with @1ec5 on retains of annotations models from the client, fast lookup upon deletion, and then replacing the idea of annotations views as both a) a link between callouts and annotation models and b) a way to avoid copying annotation models as collection storage keys. We're using an `NSMapTable` with `NSMapTableStrongMemory` for both keys and values so that we retain the annotation models (keys) like MapKit does, as well as whatever we create (`NSNumber` value wrappers for C++ annotation IDs, for now) but don't copy them. 

Capturing diff of `877-callout` here for some annotation selection/deselection logic that I might use, but that I've also prototyped out a bit anyway in [`annotation-selection`](https://github.com/mapbox/mapbox-gl-native/tree/annotation-selection). 

https://gist.github.com/incanus/efe93044e9a9aa44e487

Deleting `877-callout`, then refining what we have here.

incanus:
Backlog: 

* [x] basic addition & removal
* [x] consult delegate for symbology
* [ ] selection/deselection gestures (no popups; that's in #894)
* [ ] header docs

incanus:
Comparing selection work from `annotation-selection` here, then deleting that branch: 

https://gist.github.com/incanus/1421fcb93700596cba6d

incanus:
Based on some [rough tests with MapKit](https://github.com/incanus/MapKitSelection), it would appear that despite taking and returning an array of selected annotations, `MKMapView` only ever really selects and/or pops up a callout on one at a time. Docs point somewhat to this, re: `-[MKMapView selectedAnnotations]`: 

> Assigning a new array to this property selects only the first annotation in the array.

Then, only one annotation is ever returned in this array, so I've mirrored that behavior. 

Next up is pulling over the gesture work to 1) properly select annotation based on GL-side marker taps and 2) properly iterate through nearby markers when the group is tapped upon.

incanus:
Working through the intricacies of gestures & cycling selection state now, which is tricky since the markers aren't actual touch-enabled views at all, just pinpoint coordinates, plus we somehow have to communicate the sprite image size back to the gesture interpretation, as well as whether it has transparent areas. 

anigif-1427000240

incanus:
Ok, I think I've gotten annotation selection/deselection gestures tuned well enough for a typical marker size and offset, including with map rotation (which doesn't change marker orientation, but does change the translation of screen-to-geography bounding boxes). Per https://github.com/mapbox/mapbox-gl-native/pull/1037#issuecomment-84167167, that just leaves header documentation for the new public API methods on `MGLMapView` so it appears in the SDK docs. 

For posterity, in case anyone ever needs the flashing red rectangle shown above for debug purposes, it's easily done near the top of `-[MGLMapView handleSingleTapGesture:]` with: 

```objc
UIView *debugView = [[UIView alloc] initWithFrame:tapRect];
debugView.backgroundColor = [[UIColor redColor] colorWithAlphaComponent:0.25];
debugView.userInteractionEnabled = NO;
[self addSubview:debugView];
[UIView animateWithDuration:1
                 animations:^
                 {
                     debugView.alpha = 0;
                 }
                 completion:^(BOOL finished)
                 {
                     (void)finished;
                     [debugView removeFromSuperview];
                 }];
```

incanus:
> The proof of this of course will be a natural-feeling popup in response to map taps, so I'm jumping over to #894 next.

Eh, scratch this. We should finish the docs here first and get this merged to `master` before starting on popups since they rely on selection state.

incanus:
Header docs are done in https://github.com/mapbox/mapbox-gl-native/commit/0a5e02d88ab9c04398b438ed2cce0c114a62fb94; making a new squashed PR now.

incanus:
→ https://github.com/mapbox/mapbox-gl-native/pull/1055

PR #1043, opened 2015-03-21
Show Emerald by default

### LINKED ISSUE #964: terrain style
Let's track progress on vector-only terrain style here. 

Carrying over @jfirebaugh's last comment in #935: 

> The rendering errors (black labels, missing contours) are reproducible on any platform.
> 
> 9897ae7 fixes contour rendering. fddd27f fixes label rendering but I'm not sure if it's correct. @kkaefer?

1ec5:
a1a7ab62e176afb354109ba23360c668f84771e8 added an Emerald option to the test app, but `MGLMapView` itself has its own default style.

Fixes #964.

### DIFF
--- a/platform/ios/MGLMapView.mm
+++ b/platform/ios/MGLMapView.mm
@@ -175,7 +175,7 @@ mbgl::DefaultFileSource *mbglFileSource = nullptr;
 {
     if ( ! styleJSON)
     {
-        [self useBundledStyleNamed:@"bright-v7"];
+        [self useBundledStyleNamed:@"emerald-v7"];
     }
     else
     {


friedbunny:
screen shot 2015-03-20 at 11 43 56 pm

1ec5:
Oh fun, I hadn’t seen #1044.

1ec5:
By the way, from the perspective of the demo app, Emerald was only showing up when you cycled through all the styles and back to Emerald. This PR fixed that, so Emerald shows up at launch when the navigation bar says “Emerald”.

friedbunny:
Accidentally posted that instruments trace screenshot here, sorry if that buzzed you — bed time for me. :zzz:

PR #938, opened 2015-03-03
Add compass/bearing snapping to iOS map view

friedbunny:
À la Apple Maps' compass snapping action. If map rotation ends oriented within ±7º of north, assume it was accidental or the user was trying to get back to true and reset the map view to 0º for them.

[Video of this in action.](https://www.dropbox.com/s/whvr60fqgm7jycf/snapCompassIfNeeded.mp4?dl=0)

I don't know what Apple's snap-back threshold is, but 7º seemed like a decent balance between actual rotation and friendly UX.

PR #1059, opened 2015-03-22
bring in tweaked compass snapping fixes from @friedbunny

incanus:
Squash / tweak of https://github.com/mapbox/mapbox-gl-native/pull/938. Feels great & MapKit-like in testing.


friedbunny:
Hrm, I'm seeing this reset the compass but not the map orientation when it's combined with zooming.

Issue #1031, opened 2015-03-19
remove bundledStyleNames and useBundledStyleNamed from iOS API


jfirebaugh:
For iOS beta, we are supporting just one style. For eventual release, we'll want to have the most common usage be to load from the Mapbox Style API via a URL. Bundling a style should be an application-level concern.

incanus:
:+1: Love the narrow scope.

incanus:
We did this.

PR #1062, opened 2015-03-22
Reset applied style classes when not Hybrid

friedbunny:
This fixes the issue of Hybrid's contours and labels still being applied to Satellite, which can happen if one keeps flipping through the styles in the demo app. The applied style classes never got reset, but only Satellite possesses those classes and would draw them.

@incanus, @1ec5: Is there/will there be a better way to do this? My patch would seem to lack any finesse.

### DIFF
--- a/platform/ios/MGLMapView.mm
+++ b/platform/ios/MGLMapView.mm
@@ -1145,6 +1145,8 @@ CLLocationCoordinate2D latLngToCoordinate(mbgl::LatLng latLng)
     [self setStyleURL:[NSString stringWithFormat:@"styles/%@.json", styleName]];
     if (isHybrid) {
         [self setStyleClasses:@[@"contours", @"labels"]];
+    } else {
+        [self setStyleClasses:nil];
     }
 }
 


friedbunny:
It feels like the applied style classes should be reset whenever the basic style changes (e.g., Hybrid → Emerald).

1ec5:
@friedbunny Resetting the style classes on every base style change sounds good to me. Emerald and Satellite might both have a style class “foo” but it might not mean the same thing to both styles.

friedbunny:
It looks like the best place to fix this is in `Map::start` `asyncTerminate` by simply adding `classes.clear()`. This works to reset the applied classes on base style changes and I don't see any immediate side-effects.

jfirebaugh:
In gl-js, we went the opposite direction, moving from resetting classes when the style changes to preserving them (https://github.com/mapbox/mapbox-gl-js/issues/911). I think native and SDKs should be the same.

We're planning on removing "bundled" styles, including Hybrid, anyway (https://github.com/mapbox/mapbox-gl-native/issues/1031), so this will be an app-level concern.

friedbunny:
Roger — to that end, I'll make sure to examine the eventual API docs for clarity on retained style classes and how to remove them.

PR #1063, opened 2015-03-23
Reset applied style classes on Map::stop

friedbunny:
Deeper, non-platform specific alternative to #1062.

> This fixes the issue of Hybrid's contours and labels still being applied to Satellite, which can happen if one keeps flipping through the styles in the demo app. The applied style classes never got reset, but only Satellite possesses those classes and would draw them.

I haven't noticed any side effects, but I could very well be blind to the larger picture. cc/@kkaefer

### DIFF
--- a/src/mbgl/map/map.cpp
+++ b/src/mbgl/map/map.cpp
@@ -128,6 +128,9 @@ void Map::start(bool startPaused) {
         // Remove all of these to make sure they are destructed in the correct thread.
         style.reset();
         activeSources.clear();
+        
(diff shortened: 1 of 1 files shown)


kkaefer:
I think this is the responsibility of the calling application, not of the library. So whenever the application changes the style, it should also change the classes if desired.

jfirebaugh:
Agree with @kkaefer and see also https://github.com/mapbox/mapbox-gl-native/pull/1062#issuecomment-85137816.

PR #1073, opened 2015-03-23
Fix crash on debug build

tmpsantos:
styleInfo is being passed to request() by reference but it is stack
allocated and goes out of scope. Pass the bits we are interested by
value instead.


tmpsantos:
cc @1ec5

tmpsantos:
cc @kkaefer

kkaefer:
Good find!

PR #1075, opened 2015-03-23
iOS demo app: Prompt users to open Settings.app if location access previously denied

friedbunny:
With iOS 8 it became possible to redirect users to an app's panel in Settings.app, so let's do that if the user hits the locate button but has previously denied access. Clearly this isn't terribly important, but I see this as an example of a best practice that we can pass along to developers.

This also happens to use `UIAlertController` — another iOS 8 addition — largely because I didn't want to bother with a `UIAlertView` delegate.

ios-simulator-screen-shot-mar-23 -2015 -4 39 08-pm

iOS 7 and lower will continue to show the old alert. (Tested in simulator.)

The one downside to this is that there's no direct way to return from Settings.app to the demo app. Maybe iOS 9...

### DIFF
--- a/ios/app/MBXViewController.mm
+++ b/ios/app/MBXViewController.mm
@@ -190,10 +190,37 @@ mbgl::Settings_NSUserDefaults *settings = nullptr;
 
     if ([CLLocationManager authorizationStatus] == kCLAuthorizationStatusDenied)
     {
-        [[[UIAlertView alloc] initWithTitle:@"Authorization Denied"
-                                    message:@"Please enable location services for this app in Privacy settings."
-                                   delegate:nil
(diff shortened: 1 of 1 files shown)


incanus:
Yeah, low priority, but thanks. My outlook is to try to keep the demo app doing as little as possible to flex the map view as a dev and testing framework. But passing along a best practice could be nice.

friedbunny:
Something to consider for future, more kitchen-sinky demo apps. :bulb:

friedbunny:
This particular implementation will be obsoleted by e9993b71d63e7844942070ef2c38f3f7619ce1e5 and the move of the basic location authorization ask to `[MGLMapView setShowsUserLocation:]`.

Issue #788, opened 2015-01-28
Changing style repeatedly consistently causes crash.


mb12:
Here is how to reproduce it consistently.

1. Launch sample iosapp.
2. Keep tapping on the center of the navigation bar to change styles.
3. It consistently crashes after 10 or so style switches.  

I have copy pasted the lldb stack traces for the crashes below. I've included two unique stack traces.

Crash number 1

```
eneral: painter setup: 0.057500000000000002ms
libc++abi.dylib: terminating with uncaught exception of type std::bad_alloc
(lldb) bt
* thread #35: tid = 0x40a830, 0x3a90b1f0 libsystem_kernel.dylib`__pthread_kill + 8, name = 'Map', stop reason = signal SIGABRT
    frame #0: 0x3a90b1f0 libsystem_kernel.dylib`__pthread_kill + 8
    frame #1: 0x3a975796 libsystem_pthread.dylib`pthread_kill + 58
    frame #2: 0x3a8bbfdc libsystem_c.dylib`abort + 76
    frame #3: 0x39beacd6 libc++abi.dylib`abort_message + 74
    frame #4: 0x39c036e4 libc++abi.dylib`default_terminate_handler() + 252
    frame #5: 0x3a348f88 libobjc.A.dylib`_objc_terminate() + 228
    frame #6: 0x39c011c6 libc++abi.dylib`std::__terminate(void (*)()) + 78
    frame #7: 0x39c00a1c libc++abi.dylib`__cxa_throw + 116
    frame #8: 0x006ee7c0 libglInterpose.dylib`operator new(unsigned long) + 160
  * frame #9: 0x00086188 Mapbox GL`std::__1::__split_buffer<std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char> >, std::__1::allocator<std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char> > >&>::__split_buffer(unsigned long, unsigned long, std::__1::allocator<std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char> > >&) [inlined] std::__1::allocator<std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char> > >::allocate(this=0x1a9445b8, __n=357913941, (null)=0x00000000) + 180 at memory:1630
[...]

Issue #1072, opened 2015-03-23
Crash switching styles in iOS demo app


1ec5:
#1022 introduced a serious regression. The iOS demo app now crashes every time you switch styles:

```
#4    0x000000010f3a2303 in mbgl::Map::start(bool) at /Users/mxn/hub/mapbox-gl-native/src/mbgl/map/map.cpp:117
#5    0x000000010f617fe6 in -[MGLMapView setStyleURL:] at /Users/mxn/hub/mapbox-gl-native/platform/ios/MGLMapView.mm:208
#6    0x000000010f626e80 in -[MGLMapView useBundledStyleNamed:] at /Users/mxn/hub/mapbox-gl-native/platform/ios/MGLMapView.mm:1148
#7    0x000000010f2fc006 in -[MBXViewController cycleStyles] at /Users/mxn/hub/mapbox-gl-native/ios/app/MBXViewController.mm:204
```

This assertion fails because `mode` is `Continuous`:

```cpp
assert(mode == Mode::None);
```

tmpsantos:
I have a fix for this one I believe.

tmpsantos:
Not sure if this one is related, but it is on the load style code path:
https://github.com/mapbox/mapbox-gl-native/tree/debug_crash

1ec5:
#1073 doesn’t fix this crash, unfortunately.

1ec5:
I keep forgetting we’re planning on omitting style switching from beta 1 (#1031). So this goes on beta 2, I guess, but it’s quite severe nonetheless.

incanus:
Pretty sure I got this in https://github.com/mapbox/mapbox-gl-native/pull/1077.

kkaefer:
Issue here is that `MGLMapView setStyleURL:` calls `start`, which it shouldn't.

incanus:
Yep, this was outdated. Fixed in https://github.com/mapbox/mapbox-gl-native/commit/fb41d0cd4c5c526696e33390f2e65ea581809100.

Issue #1014, opened 2015-03-17
[iOS] Crash using bundled style immediately after instantiation


1ec5:
@DavidChouinard reports in https://github.com/mapbox/mapbox-gl-native/pull/852#issuecomment-81997446 calling `-[MGLMapView useBundledStyleNamed:]` results in a crash in a project that links against libMapboxGL. Here’s the stack trace:

```
Thread 1Queue : com.apple.main-thread (serial)
#0	0x00000001093b0286 in __pthread_kill ()
#1	0x00000001093e342f in pthread_kill ()
#2	0x000000010913f19a in abort ()
#3	0x0000000108efb481 in abort_message ()
#4	0x0000000108f232f8 in default_terminate_handler() ()
#5	0x0000000108f20b01 in std::__terminate(void (*)()) ()
#6	0x0000000108f20b8e in std::terminate() ()
#7	0x00000001062fb2ba in std::__1::thread::operator=(std::__1::thread&&) [inlined] at /Applications/Xcode-beta.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/../include/c++/v1/thread:393
#8	0x00000001062fb2b5 in mbgl::Map::start(bool) at /Users/mxn/hub/mapbox-gl-native-2/src/mbgl/map/map.cpp:154
#9	0x0000000106381f52 in -[MGLMapView setStyleURL:] at /Users/mxn/hub/mapbox-gl-native-2/platform/ios/MGLMapView.mm:179
#10	0x00000001062d869f in -[ViewController viewDidLoad] at /Users/mxn/Documents/Playroom/PeaPod/PeaPod/ViewController.m:47
#11	0x0000000107d664a0 in -[UIViewController loadViewIfRequired] ()
#12	0x0000000107d6669e in -[UIViewController view] ()
#13	0x0000000107c81bb9 in -[UIWindow addRootViewControllerViewIfPossible] ()
#14	0x0000000107c81f7f in -[UIWindow _setHidden:forced:] ()
#15	0x0000000107c8eacd in -[UIWindow makeKeyAndVisible] ()
#16	0x0000000107c31ad7 in -[UIApplication _callInitializationDelegatesForMainScene:transitionContext:] ()
#17	0x0000000107c347ee in -[UIApplication _runWithMainScene:transitionContext:completion:] ()
#18	0x0000000107c336ec in -[UIApplication workspaceDidEndTransaction:] ()
[...]

PR #1077, opened 2015-03-24
don't stop/start map in related to style changes anymore since safe

incanus:
This behaves like OS X, where the map is started once at init, and then only ever again after being stopped from app backgrounding. Nothing to do with styling, which can happen before or after start, from the main thread, without otherwise touching the map since it triggers its own updates.

### DIFF
--- a/platform/ios/MGLMapView.mm
+++ b/platform/ios/MGLMapView.mm
@@ -192,17 +192,13 @@ mbgl::DefaultFileSource *mbglFileSource = nullptr;
     }
     else
     {
-        if ([@(mbglMap->getStyleJSON().c_str()) length]) mbglMap->stop();
         mbglMap->setStyleJSON((std::string)[styleJSON cStringUsingEncoding:[NSString defaultCStringEncoding]]);
-        mbglMap->start();
     }
 }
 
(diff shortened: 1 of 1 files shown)


1ec5:
Sweet! This change seems to fix #788, #1014, and #1072.

mb12:
How does this change prevent invalid access to SourceInfo objects from worker threads which have not yet finished?

1.) SourceInfo objects are created upon parsing style.json.
2.) TileData holds a reference to SourceInfo. <== This reference will become invalid when new style is loaded into the map.
3.) TileParser holds a reference to TileData.

When style.json is changed these SourceInfo objects would get deleted. ( They reside under Map.style.layers.layers.bucket.style_source.info ).
If any of the TileParser workers has not finished after Map.style has been changed, this would cause an invalid access.

incanus:
> How does this change prevent invalid access to SourceInfo objects from worker threads which have not yet finished?

That doesn't really fall to this PR specifically, but more to @kkaefer's new approach which enabled it: 

https://github.com/mapbox/mapbox-gl-native/commit/ebd2985f785e19f3d4976f2d803ac017b755d0d9

I think the fact that it's titled "make style updates a little more threadsafe" is indication that this is a moving target still :-)

PR #1079, opened 2015-03-24
Update Mapbox 🐶

1ec5:
Kept the [DOG](http://en.wikipedia.org/wiki/Digital_on-screen_graphic) in the same location, so that the bottom edge of the baseline’s outline remains aligned with the bottom of the :information_source: graphic.


friedbunny:
I like the sizing improvements but I kind of miss the crispness of the old one.

Tangentially related question: The :information_source: graphic-button seems superfluous and tapping the bug could serve the same purpose. I know Mapbox's ToS require some specific branding and wording, but can they be consolidated? Or is that button a placeholder for an iOS-SDK-like modal?

friedbunny:
ios-simulator-screen-shot-mar-24 -2015 -10 54 22-am

incanus:
The :information_source: button will eventually bring up attribution info similarly to how the iOS SDK does — right now it's just a web page jump to keep us covered. 

We'd prefer to have a logo on the map, though like the iOS SDK, it will also get a knob to remove it if people so choose.

1ec5:
@friedbunny This is the wordmark that we use across the board. Personally, I think it helps to view it as a watermark. :wink:

Issue #756, opened 2015-01-15
GPS marker on map


ljbade:
I want to take whatever @bleege cooks up in https://github.com/mapbox/mapbox-android-sdk/issues/593 and https://github.com/mapbox/mapbox-android-sdk/issues/592 to implement a GPS map marker in the demo app.

ljbade:
I changed the GPS button to toggle location updates on or off. The updates are once a second and keep the map centre updated.

I used the https://github.com/mapzen/LOST that @bleege found.

Now I need to figure out how to add a marker to the map with compass direction and accuracy circle.

@incanus I guess this would be a good use of the client-side data stuff you are working on?

incanus:
Yes, this is exactly what that is for. You can maybe fake it at screen center for tracking for now (i.e. put the marker *above* the map) but that won't help when you want to pan the map while showing the user.

ljbade:
Yeah I was thinking about faking it @incanus. To get around the scrolling issue I may calculate screen coord from the zoom level and lat/lon

incanus:
Yeah, for a single marker, I'll bet that will work fine for now. I'll be watching as that could be useful on iOS as a stopgap too.

ljbade:
Will need to use these conversions:
https://github.com/mapbox/mapbox-gl-native/issues/476

ljbade:
Related iOS ticket: https://github.com/mapbox/mapbox-gl-native/issues/147

incanus:
Related to annotations, I'm curious @ljbade @bleege @1ec5 @kkaefer to hear your thoughts on user tracking settings and how we might push some of that down to the GL core. 

For example, on iOS, with our SDK we took the model offered by MapKit, which is to have two separate API: 

### `showsUserLocation`

A boolean that either shows the user dot or not. 

### `trackingMode`

A three-way option, the latter two options of which require or enable `showsUserLocation`: 

* `.None` — Just show (or don't) the user dot.
* `.Follow` — Center the map view on the user dot (including during movement) until the mode is changed or the map is panned manually. 
* `.FollowWithHeading` — Same as above but also keep the map rotated to the direction the user is facing (compass). 

Basically, does it make sense to do *all* of this per-platform (as we have now on Android, for example), or to have the common modes and toggles at the core level and call out to e.g. Core Location on iOS for the location functionality?

ljbade:
[...]

PR #1082, opened 2015-03-24
[not ready] GPS marker (user dot)

1ec5:
This PR depends on #1074. It is a loose port of most of the GPS marker functionality in mapbox/mapbox-ios-sdk that respects the model/controller distinction that we make for annotations in this library.

Ref #756.


incanus:
Looking into this now.

incanus:
Hmm, I'm not seeing a dot on either sim or device.

incanus:
Oh, heh. `showsUserLocation`.

include/mbgl/ios/MGLMapView.h
```diff
@@ -1,6 +1,10 @@
 #import <UIKit/UIKit.h>
 #import <CoreLocation/CoreLocation.h>
 
+#import "MGLTypes.h"
```
incanus: Per https://github.com/mapbox/mapbox-gl-native/issues/1090 we can probably move the location-related stuff here instead of including this.

1ec5: No, I’ve got a patch locally that integrates the user tracking preferences into settings_nsuserdefaults.mm. We need `MGLUserTrackingMode` somewhere generic so that OS X builds don’t have to include this iOS-specific header, which pulls in UIKit.

incanus: Ah, nice. Carry on.

incanus:
Noticed a bug: show user location, then add points to the map with the debug menu. The user location is never hidden despite being 3,000 miles away.

1ec5:
@incanus Methods like `-setZoomLevel:` and `-setCenterCoordinate:zoomLevel:` trigger `MapChangeRegionDidChange` and `MapChangeRegionDidChangeAnimated`, which don’t lead to a call to `-updateUserLocationAnnotationView`. They should.

incanus:
We should right the map north-up, with animation, when panning during tracking mode with heading. The SDK does this pretty well.

incanus:
Ok @bleege @1ec5 in the interest of wrapping and iterating from there, let's try things as of the above. The one outstanding issue I'm seeing is tracking the dot to the map properly during animated zooms, like from a double-tap gesture. Current status is to make it disappear through the animation, then come back. I'm not seeing a good way to remedy without also causing the dot to appear on screen when it shouldn't (example: locate user, add points to map, which zooms, user dot is hanging out inappropriately). 

Thoughts?

bleege:
> in the interest of wrapping and iterating from there, let's try things as of the above.

:+1: Let's not hold the whole SDK dogfooding process up over one issue.

incanus:
Yep, I am seeing one more issue — user dot showing up on first call before tracking is enabled. If I can nail that and test a bit more, I'll merge pending tests.

incanus:
Ok, think I got it.

incanus:
Squash in https://github.com/mapbox/mapbox-gl-native/pull/1110.

PR #1084, opened 2015-03-24
[Do not merge] Exploring message passing between threads

tmpsantos:
Prototyping the postTask() idiom for thread communication.

### DIFF
--- a/include/mbgl/map/map.hpp
+++ b/include/mbgl/map/map.hpp
@@ -20,6 +20,8 @@
 #include <functional>
 #include <chrono>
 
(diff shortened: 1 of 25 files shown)


incanus:
Ooh, neato.

tmpsantos:
@kkaefer I added a ::postTaskSync based on what you suggested that worked just fine and was extremely trivial to implement. The last commit is a simple test case that will do a sync call when we toggle the debug mode to test the performance.

jfirebaugh:
Doing something similar in #1065.

Issue #1081, opened 2015-03-24
Dashed lines aren't consistent with GL-JS


friedbunny:
![screen-shot-2015-03-24-at-11 39 37-am](https://cloud.githubusercontent.com/assets/1198851/6810119/09e45f7e-d21c-11e4-97f5-45a341123efd.png)
![ios-simulator-screen-shot-mar-24 -2015 -11 40 00-am](https://cloud.githubusercontent.com/assets/1198851/6810118/09e4507e-d21c-11e4-9acb-aee099ce39ff.png)

Those are both z9 [outdoors-v7](https://github.com/mapbox/mapbox-gl-styles/blob/3e72c66af32e6223e0285aecee27517f02845122/styles/outdoors-v7.json#L1533-L1553):

```cartocss
{
    "id": "admin_l3",
    "type": "line",
    "source": "mapbox",
    "source-layer": "admin",
    "filter": ["in", "admin_level", 3, 4, 5],
    "layout": {
      "line-join": "round"
    },
    "paint": {
      "line-color": "#88a",
      "line-dasharray": [3, 1],
      "line-opacity": {
        "stops": [[3, 0], [4, 1]]
      },
      "line-width": "@admin_l3_width"
    },
    "paint.night": {
      "line-color": "@admin_night"
    }
}
```
```cartocss
"@admin_l3_width": {
    "stops": [[5, 0.6], [20, 6]]
},
```

ansis:
This is caused by -native and -js having different default values for the `base` property in functions. 

https://github.com/mapbox/mapbox-gl-native/pull/1085 would fix this.

PR #1085, opened 2015-03-24
make 1.0 the default base for all types

ansis:
At some point -js switched to having one default base value for all types instead of 1.0 for color and 1.75 for other types like width. This switches -native to match -js.

Or we could change -js back to having different default base values for different types.

@jfirebaugh was the change in -js intentional?

The mismatch is causing issues like https://github.com/mapbox/mapbox-gl-native/issues/1081#event-263932060


jfirebaugh:
@ansis You tell me: https://github.com/mapbox/mapbox-gl-js/pull/640 :smiley:

ansis:
ha, I completely forgot about that

incanus:
lol

PR #1107, opened 2015-03-25
Assume UTF-8–encoded JSON stylesheets

### LINKED ISSUE #1068: Standardize on an NSString encoding
Several `MGLMapView` methods make use of `+[NSString defaultCStringEncoding]` when converting between `NSString` and `std::string`. The [`+defaultCStringEncoding` documentation](https://developer.apple.com/library/mac/documentation/Cocoa/Reference/Foundation/Classes/NSString_Class/#//apple_ref/occ/clm/NSString/defaultCStringEncoding) makes clear that we should consider it a last resort. Relying on it could lead to weird bugs that depend on the system language (especially with right-to-left languages). We should standardize on `NSUTF8StringEncoding`.

1ec5:
Fixes #1068.

### DIFF
--- a/platform/ios/MGLMapView.mm
+++ b/platform/ios/MGLMapView.mm
@@ -178,7 +178,7 @@ mbgl::DefaultFileSource *mbglFileSource = nullptr;
 {
     if (accessToken)
     {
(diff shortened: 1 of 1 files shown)


1ec5:
Also c7b8cb1477139da834a5ceef051d9df375bfeb03.

PR #1111, opened 2015-03-26
fix repeating pattern seams

ansis:
This fixes the 1px padding around repeating pattern images.

It also adds wrapping support to the bilinear scaling to avoid seems on non-round-pixel-ratio devices like iphone 6.

It fixes the issues seen here: https://github.com/mapbox/mapbox-gl-styles/issues/50#issuecomment-86159351


ansis:
@kkaefer can you review this? The old test fails, but when comparing the reference and actual images (with my eyes) they look identical. Should I dig deeper or just update?

kkaefer:
It looks like the difference here is due to floating point precision errors which is due to the order we perform the computations in. The differences are very minor and only seem to happen at 1.5x:

out

I'm going to update the test case.

ansis:
great, thanks!

Issue #1109, opened 2015-03-26
release build exposes metrics lag


incanus:
I am seeing a pretty considerable lag with the three main gestures now due to metrics. 

![screen shot 2015-03-25 at 5 36 01 pm](https://cloud.githubusercontent.com/assets/17722/6838386/db2c51aa-d315-11e4-995e-4f1cbc93da82.png)

This is showing the gesture metrics for pan, pinch, and rotate taking 10-20x longer than the actual move, zoom, and bearing commands. It's impacting the framerate. 

* Pan: `479ms` to track, `22ms` to make the change, `13ms` to push the track
* Pinch: `373ms` to track, `21ms` to make the change
* Rotate: `306ms` to track, `33ms` to make the change

All on main thread. 

/cc @bleege

incanus:
I am seeing this both in `master` and the about-to-be-merged https://github.com/mapbox/mapbox-gl-native/pull/1110. It's unusable a lot of the time :-(

Is it just my setup? Seeing it on two devices.

bleege:
Just wrapped the gesture events in an async concurrent background thread so that they're as far away from the main UI thread as possible.  Also wrapped the pushEvent and flush methods in serial background queues so things stay in order, but if there's a lag it won't impact the UI.

incanus:
Hmm, I am still seeing all activity on `DISPATCH_QUEUE_PRIORITY_BACKGROUND`, which isn't serial.

bleege:
Thanks for the GCD education @incanus!  Now they should be serial queues.

incanus:
I should clarify in my original post that those were CPU time totals for the actions. 

Going to check this out shortly.

incanus:
Fixed in https://github.com/mapbox/mapbox-gl-native/pull/1112.

PR #1112, opened 2015-03-26
ui blocking

bleege:
Draino for #1109


incanus:
Looking awesome at first glance. Will try in a bit.

incanus:
Perf looks great here and I think the approach is sound. I'm not sure how much you've probed multiple map instances running per app, but at the very least it's good practice to unique ID any created queues. I put the app bundle ID into the name for debugging purposes, as well as a globally unique ID just in case. 

When tests pass I am :+1: to merge.

bleege:
Fantastic!  Glad it looks and runs great.  Thanks for the crash course on "GCD For Fun And Profit" last night.

Issue #1103, opened 2015-03-25
Multiple Maps EXEC_BAD_ACCESS TransformState::getLatLng() 


jonathanduty:
setup:
I modified the left and right navbar buttons to either push a new MBXViewController or pop it using the UINavigationController.  I wanted to test navigating from one map to another using push/pop UINavigation controller.

Steps:
Open app.  Let first map fully load.    Hit right bar button to navigate to new map.  Hit left button to navigate back to first map.  Touch the screen, you get a EXEC_BAD_ACCESS

```
Thread 1Queue : com.apple.main-thread (serial)
#0	0x0018efb2 in mbgl::TransformState::getLatLng() const at /Users/jonathanduty/Development/mapbox/mapbox-gl-native/src/mbgl/map/transform_state.cpp:98
#1	0x0018f953 in mbgl::TransformState::latLngForPixel(mbgl::vec2<double>) const at /Users/jonathanduty/Development/mapbox/mapbox-gl-native/src/mbgl/map/transform_state.cpp:205
#2	0x00379bc0 in mbgl::Map::latLngForPixel(mbgl::vec2<double>) const at /Users/jonathanduty/Development/mapbox/mapbox-gl-native/include/mbgl/map/map.hpp:152
#3	0x00367971 in -[MGLMapView convertPoint:toCoordinateFromView:] at /Users/jonathanduty/Development/mapbox/mapbox-gl-native/platform/ios/MGLMapView.mm:1154
#4	0x00365d4b in -[MGLMapView trackGestureEvent:forRecognizer:] at /Users/jonathanduty/Development/mapbox/mapbox-gl-native/platform/ios/MGLMapView.mm:992
#5	0x0035f652 in -[MGLMapView handlePanGesture:] at /Users/jonathanduty/Development/mapbox/mapbox-gl-native/platform/ios/MGLMapView.mm:626
#6	0x01923287 in _UIGestureRecognizerSendActions ()
```

incanus:
> `[MGLMapView trackGestureEvent:forRecognizer:]`

@bleege want to investigate?

jonathanduty:
At first glance following variables are declared in global scope in MGLMapView:

mbgl::Map *mbglMap = nullptr;
MBGLView *mbglView = nullptr;
mbgl::SQLiteCache *mbglFileCache = nullptr;
mbgl::DefaultFileSource *mbglFileSource = nullptr;

[...]

PR #1115, opened 2015-03-26
Gestures

bleege:
To address some ~~#1103~~ https://github.com/mapbox/mapbox-gl-native/issues/1116 issues.

### DIFF
--- a/platform/ios/MGLMapView.mm
+++ b/platform/ios/MGLMapView.mm
@@ -1032,10 +1032,11 @@ mbgl::DefaultFileSource *mbglFileSource = nullptr;
 
 - (void)trackGestureEvent:(NSString *)gesture forRecognizer:(UIGestureRecognizer *)recognizer
 {
+    CGPoint ptInView = CGPointMake([recognizer locationInView:recognizer.view].x, [recognizer locationInView:recognizer.view].y);
+    const CLLocationCoordinate2D coord = [self convertPoint:ptInView toCoordinateFromView:recognizer.view];
+
     dispatch_async(dispatch_get_global_queue(DISPATCH_QUEUE_PRIORITY_BACKGROUND, 0), ^{
         // Send Map Zoom Event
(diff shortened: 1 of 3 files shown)


bleege:
Waiting for Travis to :+1: and then will merge.

platform/ios/MGLMapView.mm
```diff
@@ -1032,10 +1032,11 @@ - (BOOL)gestureRecognizer:(UIGestureRecognizer *)gestureRecognizer shouldRecogni
 
 - (void)trackGestureEvent:(NSString *)gesture forRecognizer:(UIGestureRecognizer *)recognizer
 {
+    CGPoint ptInView = CGPointMake([recognizer locationInView:recognizer.view].x, [recognizer locationInView:recognizer.view].y);
+    const CLLocationCoordinate2D coord = [self convertPoint:ptInView toCoordinateFromView:recognizer.view];
```
incanus: This is still going to have a performance impact, doing projection math at the Cocoa level every hit of a gesture. We may want to see if we can do this in a `const` method in a non-main thread.

bleege:
Actually will address issues in #1116

bleege:
Travis approved the commits, but will do some more research into #1116 before merging.

platform/ios/MGLMapView.mm
bleege: I'm not sure that it can be put into a non-main UI thread, as the `convertPoint` method relies on `[UIView convertPoint:fromView]` itself.

incanus: > We may want to see if we can do this in a `const` method in a non-main thread.

Yeah, I meant more changing the C++ to make this safer somehow for non-main threads — like snapshotting the `TransformState` or something. Looking into it.

incanus:
Obsoleted by https://github.com/mapbox/mapbox-gl-native/pull/1132.

PR #1118, opened 2015-03-26
Add simple non-semantic version support to build script

friedbunny:
With version string ` ios-beta-preview-sirius-1.0v2` I was getting the following Xcode build error:

```
Traceback (most recent call last):
  File "/Users/jason/Documents/Mapbox/mapbox-gl-native/scripts/build-version.py", line 56, in <module>
    patch = tag[2],
IndexError: list index out of range
Command /bin/sh failed with exit code 1
```

My fixes would result in version `1.0.2` in this case, which is wrongish but builds.

### DIFF
--- a/scripts/build-version.py
+++ b/scripts/build-version.py
@@ -13,6 +13,8 @@ else:
     if sys.argv[2] == sys.argv[3]:
         tag = [0, 0, 0]
     else:
+        if "v" in sys.argv[2][-6:]:
+            sys.argv[2] = sys.argv[2][-6:].replace('v', '.')
         ver = re.sub("[^0-9.]", "", sys.argv[2])
         tag = map(int, ver.split('.'))
         rev = sys.argv[3][0:8]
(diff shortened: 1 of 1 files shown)


1ec5:
I think the use of a “v” here was an oversight, and the x.y.z format is probably the only one we need for the time being. But maybe this change will be useful if we ever decide to have things like 1.2.3b7.

incanus:
Yeah, I was just trying to stay similar to an internal project tag. Let's leave this open and take a peek when some dust settles. Thanks @friedbunny!

friedbunny:
Hah, the tag renaming had me confused — I didn't see a `ios-beta-preview-sirius-1.0v2` tag in git but I couldn't get my copy to build. Eventually did a tactical-nuking of my local repo. :boom:

jfirebaugh:
We're going to stick with x.y.z version numbers, so this shouldn't be necessary.

Issue #1128, opened 2015-03-26
Convert Boolean To Strings


bleege:
Convert all metrics to use `true` or `false` strings instead of boolean values.

/cc @camilleanne @edenh @mick

bleege:
Going with #1135 as it's setup for Post Metrics Refactoring.  Restarted build in Travis to get signoff and then will merge.

/cc @incanus

PR #1129, opened 2015-03-27
Bool To String

bleege:
For #1128


incanus:
I'm thinking the fact we use `true` and `false` is an implementation detail and shouldn't involve the client at all. Will resolve this internally + adapt for recent metrics refactoring.

incanus:
https://github.com/mapbox/mapbox-gl-native/pull/1135 does this post-refactor and keeps it internal to the metrics manager.

PR #1130, opened 2015-03-27
[not ready] metrics stability

incanus:
Refactoring of the metrics engine for threadsafety and performance. 

One quick question before the summary for you @bleege: are these two methods on `MGLMetricsLocationManager` needed? I don't see them called anywhere. 

```objc
- (BOOL) isAuthorizedStatusDetermined;
- (void) requestAlwaysAuthorization;
```

-

The overarching design goal here is performance for rapid-fire calls from the main UI thread (like during pan events) while maintaining internal integrity. 

Here are some design principles, cribbed from [the internals of `MGLMapboxEvents`](https://github.com/mapbox/mapbox-gl-native/blob/adec8d049ef4d77325324bc7cd341ad86832c97b/platform/ios/MGLMapboxEvents.m#L27-L41): 

* All variables accessed from more than one thread are designated `atomic` and accessed through dot syntax. The main thread uses underscore syntax during the initialization of the variable.

* All variables accessed outside of initialization and from within a single thread use underscore syntax.

* All captures of `self` from within asynchronous dispatches will use a `weakSelf` to avoid cyclical strong references.

Also:

* All posts of metrics and flushing of the events queue can be called from any thread. They always return immediately. 

* All access to the events queue for management and dispatching of network uploads happens on a dedicated serial queue. If events in this queue need access to the main thread, they block waiting for it so that other uses of the main thread have higher priority. 

* Any access to UIKit methods needed for metrics calls from the main thread (such as the determining if email or push is configured for `map.load` inside of `-[MGLMapView commonInit]`) should not block the main thread. Instead, events are posted and dedicated methods for gathering these details safely are used (e.g. `+[MGLMapboxEvents checkEmailEnabled]`). 

* All methods are documented in the header as to threadsafety. 

* Callers of this API shouldn't have to worry about making or using any queues. 

* I've moved all methods or variables to private where possible. 

* I cleaned up some types like avoiding use of `int`, using `NSTimeInterval` and other specialized types where possible, and using constant `NSString`'s for event keys instead of magic strings. 

* I used boxed literals like `@(scale)` instead of verbose `NSNumber`'s, especially in calling code. 

* I've moved from instance methods like `-[[MGLMapboxEvents sharedInstance] flush]` to abbreviated class methods like `+[MGLMapboxEvents flush]` to keep code in e.g. the map view lighter.


incanus:
To be clear: I'm still testing a bit, just wanted to push up what I had so far.

bleege:
This all looks good.  Thanks for working out the remaining kinks.  Speaking of which, both `isAuthorizedStatusDetermined` and `requestAlwaysAuthorization ` can be removed.  They are both legacy code that hadn't been pulled yet.

incanus:
:+1:

incanus:
Another question @bleege: is `flushAfter` really supposed to be 10,000 seconds, i.e. ~2.75 hours?

bleege:
That was a default value for `flushAfter` ported from the original JS version that has lived on longer that it should have.  I'd go with 15 to 30 seconds.

incanus:
Another question: if the device is offline, posts error out and don't retry and are lost? Are we not worrying about that for now? Seems an ok tradeoff given that we aren't supporting offline maps functionality yet.

incanus:
I went with one minute on the flush for now, since we also flush when the app goes into the background anyway.

bleege:
> Are we not worrying about that for now? Seems an ok tradeoff given that we aren't supporting offline maps functionality yet.

Exactly.

incanus:
I haven't been able to crash this yet and I've been getting get throughput on the events posting.

incanus:
One more note: I didn't about complicating the in-gesture calls to `-[MGLMapView convertPoint:toCoordinateFromView:]` by trying to get them off the main thread. They are `const` C++ methods operating on a rapidly changed `TransformState` anyway and are just floating point math. This should be light (and indeed profiles as such).

incanus:
→ https://github.com/mapbox/mapbox-gl-native/pull/1132

PR #1135, opened 2015-03-27
fixes #1128: convert metrics BOOLs to strings internally

### LINKED ISSUE #1128: Convert Boolean To Strings

incanus:
This fixes #1128 and keeps the logic around converting a `BOOL` like `YES` to `true` internal to the metrics engine. 

@bleege Want to take a look?

### DIFF
--- a/include/mbgl/ios/MGLMapboxEvents.h
+++ b/include/mbgl/ios/MGLMapboxEvents.h
@@ -33,7 +33,7 @@ extern NSString *const MGLEventMapLocation;
 
 // You can call these methods from any thread.
 //
-+ (NSString *) checkEmailEnabled;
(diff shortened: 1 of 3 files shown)


bleege:
Yep, I like this solution.  Keep the SDK internals BOOL and then just convert to String when sending over the fence to the metrics engine.  Will use #1135 instead of #1129 as it's "Metrics Refactored Enhanced"&trade;.

bleege:
Restarted build for Travis greenlighting.  https://travis-ci.org/mapbox/mapbox-gl-native/builds/56063078

PR #1143, opened 2015-03-27
Move iOS MGLUserLocation headers to "include" directory

friedbunny:


1ec5:
The intent was to keep these two headers private. We do plan to make MGLUserLocationAnnotationView.h public eventually, but only once we implement a view-based annotation API.

friedbunny:
Blargh.

Issue #1134, opened 2015-03-27
mail metrics erroneously reports false


incanus:
If `MFMailComposeViewController` is not found, mail setup status can't be checked. It might not be found because its framework isn't linked in, but email could still be setup. The `canSendMail` method is only meant to check that mail is setup before attempting to send. 

/cc @bleege

bleege:
Discussed with @camilleanne @edenh @incanus this afternoon and decided to pull the `enabled.mail` metric for now as the current setup "only tells us if the app is using email sending at all, not if email is setup on the device".

bleege:
Removed and merged!

PR #1150, opened 2015-03-28
Removing enabled.mail

bleege:
For #1134 .


bleege:
Build failed 3 of 11 sub builds, but they were all related to Android and not the iOS changes.

kkaefer:
> Build failed 3 of 11 sub builds, but they were all related to Android and not the iOS changes.

In this case, they were all build system/network failures, but please make sure Android passes to; otherwise we can just as well ignore build failures altogether.

Issue #1125, opened 2015-03-26
improve synced view tracking


incanus:
General ticket for improvement of "sticking" native, Cocoa-side views to the panning, zooming GL render view. 

This relates to the user dot now, later to callouts and annotation views. Let's keep a ticket around until we are 100% satisfied with the behavior here, but we can move it along milestones as we make staged improvements.

incanus:
I'm feeling pretty good about the current state of tracking, especially after the above commit which avoids invalid `CLLocationCoordinate2D`-based updates during `TransformState` "will change" events (and just catches the "did change"). 

The one issue I'm seeing which I'd like to clean up is the dot disappearing during e.g. a double-tap animated zoom. It doesn't show up again until the next `CLLocationManager` update. It gets hidden because the `CGPoint` received out of `-[MGLMapView convertCoordinate:toPointToView:]` is well out of range of given screen bounds. I think this is just some sort of math error in `TransformState` during animations.

incanus:
Another lead that confirms the above: pan the map until the dot is >150px off screen, then hit the user locate button. The map pans back into view (meaning the map centering to the valid user location is of course working) but the dot doesn't show up until the next CL update post-animation. During the animation it's still residing at the offscreen value, so it's hidden.

incanus:
Tracking is pretty good right now. Only outstanding issue I'd like to hit for the beta is the dot disappearing briefly during animated zooms (but coming back at least, now). I'm nearing some completion on this but don't want to hold up a build tonight so I'll work on that a bit over the weekend. 

Going to squash & merge a few commits now with improvements thus far.

incanus:
First batch of improvements in https://github.com/mapbox/mapbox-gl-native/pull/1151.

incanus:
Main blocker here now captured in the more general https://github.com/mapbox/mapbox-gl-native/issues/1160 bug with projection conversion.

kkaefer:
FWIW, I think we should get rid of overlaying iOS views onto the Map and instead render everything ourselves. Apple recommends against doing this as well: https://developer.apple.com/library/ios/documentation/3DDrawing/Conceptual/OpenGLES_ProgrammingGuide/Performance/Performance.html#//apple_ref/doc/uid/TP40008793-CH105-SW8

incanus:
[...]

PR #1151, opened 2015-03-28
refs #1125: user dot improvements

incanus:
* more quickly unhide dot when nearing viewport
* improvements to maintaining user tracking mode during tap zooms
* removed a redundant layer hide
* better at ignoring bad location data
* be sure to use existing coordinate utilities


incanus:
Travis is having some issues with this. I'm going for it.

PR #1161, opened 2015-03-30
fixes #892: support mapbox:// style URLs

### LINKED ISSUE #892: support mapbox:// style URLs
Port https://github.com/mapbox/mapbox-gl-js/commit/699826a4d5fec8c04fec35d0454f70a6e14db382 to native.

incanus:
Addresses #892 and ports over https://github.com/mapbox/mapbox-gl-js/commit/699826a4d5fec8c04fec35d0454f70a6e14db382. 

This lets the user continue to pass a bundled style like `styles/emerald-v7.json` from Cocoa, which will get converted into an `asset://` URL as before. 

But if a style URL like `mapbox://justin.80b0e36d` is passed instead, it gets converted to the style API format and loaded via the regular remote style URL facility. 

One thing I'm not clear on: 

* I'm not quite sure what `base` should be in `StyleInfo`. Does this matter for URL-based styles @kkaefer @jfirebaugh? Going to keep digging into this but can't quite figure it out yet.

### DIFF
--- a/platform/ios/MGLMapView.mm
+++ b/platform/ios/MGLMapView.mm
@@ -172,9 +172,18 @@ mbgl::DefaultFileSource *mbglFileSource = nullptr;
     }
 }
 
-- (void)setStyleURL:(NSString *)filePathURL
+- (void)setStyleURL:(NSString *)styleURLString
 {
-    mbglMap->setStyleURL(std::string("asset://") + [filePathURL UTF8String]);
+    std::string styleURL([styleURLString UTF8String]);
+
+    if ( ! [[NSURL URLWithString:styleURLString] scheme])
+    {
+        mbglMap->setStyleURL(std::string("asset://") + styleURL);
+    }
+    else
+    {
+        mbglMap->setStyleURL(styleURL);
+    }
 }
 
 - (BOOL)commonInit
--- a/src/mbgl/map/map.cpp
+++ b/src/mbgl/map/map.cpp
@@ -400,13 +400,15 @@ void Map::setup() {
 void Map::setStyleURL(const std::string &url) {
     assert(Environment::currentlyOn(ThreadType::Main));
 
-    const size_t pos = url.rfind('/');
+    const std::string styleURL = mbgl::util::mapbox::normalizeStyleURL(url, getAccessToken());
+
+    const size_t pos = styleURL.rfind('/');
     std::string base = "";
     if (pos != std::string::npos) {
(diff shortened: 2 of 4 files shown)


incanus:
As far as I can gather, we don't need to set a `base` for online hosted styles, as their sprite/glyph/etc. URLs would be specified as absolute and not relative. /cc @samanpwbb

platform/ios/MGLMapView.mm
```diff
@@ -172,9 +172,18 @@ - (void)setStyleJSON:(NSString *)styleJSON
     }
 }
 
-- (void)setStyleURL:(NSString *)filePathURL
+- (void)setStyleURL:(NSString *)styleURLString
 {
-    mbglMap->setStyleURL(std::string("asset://") + [filePathURL UTF8String]);
+    std::string styleURL([styleURLString UTF8String]);
+
+    if ( ! [[NSURL URLWithString:styleURLString] scheme])
```
jfirebaugh: Can we drop this case and require the app developer to use `"asset://..."` if that's what they want?

incanus: `setStyleURL:` is private right now. At such time that we support in-framework setting of a style URL (versus download of JSON externally and passing *that*) I think we could, but also, it's weird to use `asset://` — this has no meaning outside of GL and typically a Cocoa dev would use `+[NSURL fileURLWithPath]`.

jfirebaugh:
> As far as I can gather, we don't need to set a base for online hosted styles, as their sprite/glyph/etc. URLs would be specified as absolute and not relative.

Yes, they should be using `mapbox://` URLs in turn. Though for sprites that's pending a sprites API.

incanus:
Ok, I'm going to merge this then once CI passes.

jfirebaugh:
:+1:

platform/ios/MGLMapView.mm
kkaefer: It may have no meaning outside of our codebase, but we use to to make our styles compatible across renderers.

kkaefer:
Can you please add tests for this?

platform/ios/MGLMapView.mm
incanus: Right, no qualms with it, I'd just rather not expose the scheme to end users if we can avoid it.

incanus:
Will do.

incanus:
Tests done in https://github.com/mapbox/mapbox-gl-native/pull/1167.

PR #1163, opened 2015-03-30
[not ready] refs #892: Cocoa support for map ID style URLs

incanus:
Blocked by https://github.com/mapbox/mapbox-gl-native/pull/1161. 

This adds a Cocoa initializer for map ID style URLs (e.g. `mapbox://<user.style>`).


incanus:
Hmm, we should probably actually squash this down to map ID directly instead of `mapbox://`.

Issue #880, opened 2015-02-14
split OS X & two-way iOS builds to three-point matrix on Travis


incanus:
https://github.com/mapbox/mapbox-gl-native/blob/6da0ef3d6df9fa559252d4922f22f0c160c50fa2/scripts/travis_script.sh#L37-L51

jfirebaugh:
Let's wait to try this. Given how long OSX builds tend to be queued on Travis it could end up taking longer overall.

incanus:
Testing here has gone well in e.g. https://travis-ci.org/mapbox/mapbox-gl-native/builds/57421692. This will get rolled in as part of #1219.

Issue #1064, opened 2015-03-23
Compass/bearing snapping fails when also zooming


friedbunny:
If one rotates and then pinch zooms, the pinch gesture [cancels the map transition](https://github.com/mapbox/mapbox-gl-native/blob/9dd9bd6e9ddf40ae3ae1746b191ce7d0a8ea0842/platform/ios/MGLMapView.mm#L644) called by the fancy new snapping (#1059) but `bearing` still gets reset and becomes out of sync. :disappointed:

```
2015-03-22 21:19:55.291 Mapbox GL[5984:764115] Rotate: -2.208035º
2015-03-22 21:19:55.292 Mapbox GL[5984:764115] Should we snap?: 357.791965º
2015-03-22 21:19:55.292 Mapbox GL[5984:764115] SNAP!
2015-03-22 21:19:55.294 Mapbox GL[5984:764115] Pinch: -0.000000º
...
2015-03-22 21:26:44.939 Mapbox GL[5984:764115] Rotate: -3.104398º
2015-03-22 21:26:44.939 Mapbox GL[5984:764115] Should we snap?: 356.895602º
2015-03-22 21:26:44.940 Mapbox GL[5984:764115] SNAP!
2015-03-22 21:26:44.940 Mapbox GL[5984:764115] Pinch: -0.000000º
```
[![screen-shot-2015-03-22-at-9 20 17-pm](https://cloud.githubusercontent.com/assets/1198851/6774503/c3165644-d0de-11e4-92a1-8b563742219a.png)](https://cloud.githubusercontent.com/assets/1198851/6774501/b763a54a-d0de-11e4-8e23-8504be4ffce7.png)

incanus:
Ok, I'll probably end up reverting the feature in `master` then if it's got complications. Can hit it after the first beta...

friedbunny:
Yeah, unfortunately it's not quite there yet. I'll look into putting a `isSnapping` check or something in there.

friedbunny:
I've got a fix ready to go for this (e94c7a305de7e0df6e02ee58089d1ac0e9f50a29) that essentially duplicates `animatingGesture` stuff elsewhere. Not sure that `userInteractionEnabled` needs to be disabled (didn't notice any difference), but I've left that there. Wait on the PR?

PR #1168, opened 2015-04-01
fixes #830, #880: bring back iOS tests; split OS X & iOS tests

### LINKED ISSUE #830: iOS integration testing

incanus:
Bring it back! 

* `make itest` at the command line to run `xcodebuild test` in console
* builds static library using `package_os.sh` to run inside test harness app
  - This is basically impossible to integrate with GYP, so instead we test our framework as a user would, by using `libMapboxGL.a` in a ready-to-go app. This app has a KIF-based test bundle which you can run with `Command-U` in Xcode. 
* Updates the old `gl-cocoa`-based project for the latest setup & API. 
* Adds a few more tests like testing rotation now that KIF supports the gestures needed.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -12,7 +12,7 @@
 
 [submodule "test/ios/KIF"]
 	path = test/ios/KIF
-	url = https://github.com/mapbox/KIF.git
+	url = https://github.com/kif-framework/KIF.git
 
 [submodule "platform/ios/vendor/SMCalloutView"]
 	path = platform/ios/vendor/SMCalloutView
--- a/.travis.yml
+++ b/.travis.yml
@@ -57,6 +57,7 @@ env:
    - IOS_PROFILE_NAME="ios-in-house"
    - secure: "nQqSM8rd7OHtV4MqmNqVnkrVHqxKqQsaWRYk4/nPdhbeVWtTtkk0df711LrF1TUtbEPEewHxYUvTZ/UXmwJNeoKdzTHavI8hnatRkgjyxGERPn1il1Otelht9I+LQQHf+plrpRjVWBrNIW0Zox1B3cqn6d3NglpbXrEQ2EjYGNA="
    # end iOS code signing
+   - KIF_SCREENSHOTS="${TRAVIS_BUILD_DIR}/screenshots"
 
 before_install:
 - if [[ ${TRAVIS_OS_NAME} == "linux" ]]; then sudo service mysql stop; fi
@@ -68,6 +69,7 @@ before_install:
 - ./scripts/travis_before_install.sh
 - if [[ ${TRAVIS_OS_NAME} == "linux" && ${MASON_PLATFORM} != "android" ]]; then export LD_LIBRARY_PATH=`mason prefix mesa 10.4.3`/lib; fi
 - if [[ ${TRAVIS_OS_NAME} == "linux" && ${MASON_PLATFORM} != "android" ]]; then glxinfo; fi
+- if [[ ${TRAVIS_OS_NAME} == "osx" && ${TEST_OS} == "ios" ]]; then mkdir -p ${KIF_SCREENSHOTS}; fi
 
 install:
 - ulimit -c
@@ -87,6 +89,9 @@ script:
 
 after_script:
 - ./scripts/ios_travis/remove-key.sh
+
+after_failure:
+- if [[ ${TRAVIS_OS_NAME} == "osx" && ${TEST_OS} == "ios" ]]; then ./scripts/ios_travis/upload_screenshots.sh; fi
 
 notifications:
   slack:
--- a/Makefile
+++ b/Makefile
@@ -29,6 +29,9 @@ styles/styles:
 SMCalloutView:
 	git submodule update --init platform/ios/vendor/SMCalloutView
 
+KIF:
+	git submodule update --init test/ios/KIF
+
 #### Library builds ############################################################
 
 .PRECIOUS: Makefile/mbgl
@@ -128,6 +131,9 @@ isim: Xcode/ios
 
 ipackage: clean Xcode/ios
 	./scripts/package_ios.sh
+
+itest: ipackage KIF
+	./scripts/test_ios.sh
 
 # Legacy name
 iproj: ios-proj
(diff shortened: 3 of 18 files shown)


incanus:
Thinking about punting on this for a while. I am getting various tests failing on the same build — e.g. simple gesture tests that pass on iOS 8.1 but fail on iOS 7, varying between builds. Always works locally. I think it's a Travis problem. I don't know if it's KIF or command-line simulation or what.

incanus:
Holy crap, `a620eee` [actually passed](https://travis-ci.org/mapbox/mapbox-gl-native/builds/56812159). 

Re-squashing and force-pushing here.

incanus:
I'm aiming to support automated testing on the following: 

* iPhone 5s iOS 7.1
* iPhone 5s iOS 8.1
* iPad 2 iOS 7.1
* iPad 2 iOS 8.1

That gives us a fair coverage of iPhone/iPad, iOS 7/8, and retina/non-retina. 

This is what both Travis and `make itest` use. 

Manual testing (`Command-U` on `test/ios/ios-tests.xcodeproj` in Xcode) will use the currently-selected device configuration.

incanus:
iOS tests all [seem to be passing](https://travis-ci.org/mapbox/mapbox-gl-native/jobs/57111036) reliably now; we've also got the latest `master` merged in. I'm only seeing one device run, though, not the four we'd like that should be handled by `xcodebuild test -destination` with multiple flags. Will investigate.

friedbunny:
`xcpretty` doesn't appear to be available on Homebrew, [their suggested install](https://github.com/supermarin/xcpretty#installation) is `gem install xcpretty`.

incanus:
Oh, whoops. Yeah, we do actually already do it here in `master`: 

https://github.com/mapbox/mapbox-gl-native/blob/36cee94e61289a81f0a2931218295ca12101b116/scripts/travis_before_install.sh#L51

But I'll patch up the docs.

friedbunny:
I got this working locally after swapping to iOS 8.2 in `test_ios.sh` and `package_ios.sh`.

Despite having 7.1, 8.1, and 8.2 installed and usable as simulators in Xcode, `xcodebuild` doesn't see them.

```bash
jason@keaton ~/mapbox-gl-native $ xcodebuild -showsdks
OS X SDKs:
	OS X 10.9                     	-sdk macosx10.9
	OS X 10.10                    	-sdk macosx10.10

iOS SDKs:
	iOS 8.2                       	-sdk iphoneos8.2

iOS Simulator SDKs:
	Simulator - iOS 8.2           	-sdk iphonesimulator8.2
```

incanus:
Yep, these are the SDKs, not necessarily the runtime simulations. They are separate. A given version of Xcode (6.2 in the case) is going to build against the latest SDK at time of its release. You'll see only 8.3 for 6.3b, or 8.1 for 6.1.1. 

I too am having problems with simulator versions, but mostly on Travis. They [say](http://docs.travis-ci.com/user/osx-ci-environment/) they have 7.0, 7.1, and 8.1 currently. My problems are mostly with tests failing for unknown reasons despite **never** failing locally, or simulators timing out. 

Peeled off some useful work into https://github.com/mapbox/mapbox-gl-native/pull/1219, which will merge soon, then I'll rebase this and keep picking at it.

incanus:
Just split off the packaging phase of this as part of #1219, so this needs some quick rebasing and can be picked up again. 

I spent some time today trying to get screenshot uploads to S3 working without luck in the hopes of seeing what state things were in when tests mysteriously fail. 

I may consider just leaving this as `make itest` and not yet worrying about Travis automatic run of those tests for now.

incanus:
In some other tests, we've started to have good results with Bitrise for CI. I'm thinking the next steps here are to get these tests merged in, but run manually in Xcode / `Makefile` for the time being, then we have good luck with CI, automate on a per-commit basis. I will pick this up early next week.

friedbunny:
For those following along at home, with the move to Xcode 6.3 I had to also do:

```
make clean && make distclean && make itest
```

Otherwise I'd get variations on `libuv` linking errors:

```
  "_alphasort$INODE64", referenced from:
      _uv__fs_readdir in libMapboxGL.a(fs.o)
  "_scandir$INODE64", referenced from:
      _uv__fs_readdir in libMapboxGL.a(fs.o)
```

friedbunny:
@incanus Derailing your thread here a bit, but when and how would you like tests contributed? I've got a mock `CLLocationManager` working and couple tests around user tracking modes written.

There is a potential issue with the Travis simulators not liking the location authorization override, but there are several ways to attack that if my preferred way doesn't cut it.

incanus:
I've now got tests working locally across a matrix for both iPhone & iPad as well as iOS 7.1 and iOS 8.x. https://travis-ci.org/mapbox/mapbox-gl-native/jobs/59800251 indicates a failure with the compass going away when the map has north reset, which we've seen before (is https://github.com/mapbox/mapbox-gl-native/issues/1064 related?). Working on that bug now.

incanus:
Finally, success!! 

I'm inclined to stick to Travis right now since load is down and the rest of our build matrix is there. If in the future we decide to look elsewhere, we can consider maybe putting the KIF tests on Bitrise of Greenhouse CI, both of which we've had success with elsewhere. 

Sound good @bleege @1ec5? 

Going to squash this now and prepare for merge.

incanus:
To clarify one point about this whole setup: 

- Tests run from a separate `test/ios/ios-tests.xcodeproj` instead of the dev iOS app for two reasons: 
  - Making a unit testing bundle by way of GYP is a nightmare/maybe impossible. 
  - We should test integration by way of static library as opposed to the "inline" method that the dev app uses.

bleege:
> I'm inclined to stick to Travis right now since load is down and the rest of our build matrix is there.

Am I catching things out of order now (likely) as I thought we were on to Bitrise now?  Either way we should settle on one soon (today if possible) so that we can gain team confidence in the setup over a period of time.

incanus:
Sorry for the confusion @bleege. Downstream, user-facing apps are on Bitrise, but the GL framework itself is on Travis with the rest of GL. Reasoning above in https://github.com/mapbox/mapbox-gl-native/pull/1168#issuecomment-95756935. 

The separation is: 

- GL tests things like Do the map view constraints work? Does the compass work properly? 
- App tests things like Do my uses of annotations work? Do my other services integrate well?

PR #1164, opened 2015-04-01
Fix tile parsing performance regression

kkaefer:
50b5c1565366e223bf657d68ddac70b25ad13bcf introduced a performance regression for tile parsing, which slowed down parsing by ~50%. More information in https://github.com/mapbox/mapbox-gl-native/commit/50b5c1565366e223bf657d68ddac70b25ad13bcf#commitcomment-10491046


incanus:
Nice! Well done. ::goes to learn from it::

PR #1172, opened 2015-04-01
Use hash instead of string key when merging lines 

kkaefer:
In the vein of #1170, but for `merge_lines.cpp`.


kkaefer:
Before:

instruments1 2015-04-01 18-00-18

After:

instruments 2015-04-01 18-03-08

PR #1175, opened 2015-04-02
Make std::chrono use less painful

jfirebaugh:
Core of this is

```diff
+namespace mbgl {
+
+using clock = std::chrono::steady_clock;
+
+using time_point = clock::time_point;
+using duration   = clock::duration;
+
+}
```


kkaefer:
I changed the lower case names to `Clock`, `TimePoint`, and `Duration` to avoid shadowing issues altogether, and to be in line with our naming policy.

jfirebaugh:
:+1:

PR #1176, opened 2015-04-02
Rework easing transition code

### LINKED ISSUE #79: clean up linear interpolation determination
Over in #68 we brought in the new POI rendering, but the `map->startRotating()` etc. feels heavy-handed since: 

 * We immediately call `map->rotate()` etc. for events like these. 
 * There are timeouts, animations, and successive cancellations of each involved. 

It feels like the `Transform` should *know* when it's rotating/panning/scaling and adjust sprite interpolation accordingly, without having to enter and exit these modes at the platform level. 

For example, couldn't it check if active animations involving the angle/offset/scale are occurring presently and flip the switch then? 

Thoughts @kkaefer?

### LINKED ISSUE #126: Refactor transitions/transforms
We should be more explicit about finding situations where we need a rerender vs. situations where the animation values didn't change. This is especially true for the timeout transition, which only changes a value after a certain time, but does not change the value inbetween. In those cases, we should be smart about not needing to iterate through the event loop and only wake up the event loop once the timeout is up.

### LINKED ISSUE #1155: Only run Style::updateProperties when required
We're currently running `Style::updateProperties` too often; whenever anything about the transform changes. However, we only need to run it on zoom, or when adding/removing classes.

jfirebaugh:
Note: PR targets the `chrono` branch so that the diff is better, but this will merge to master.

This brings the easing transition code a bit closer to how easings work in gl-js. Instead of having an array of individual transitions for scale, rotate, and pan, there is a single transition function that does all the required calculations. This permits us to:

* Eliminate the "timeout" transition. (Fixes #126)
* Remove the start/stopPanning() and similar functions from the API. The transform is capable of determining when a transition that affects the pan position is in effect. (Fixes #79)
* Run style recalculations only when an ease transition that affects the zoom is in progress. (Fixes #1155)

@kkaefer can you review please?

### DIFF
--- a/include/mbgl/map/map.hpp
+++ b/include/mbgl/map/map.hpp
@@ -3,6 +3,7 @@
 
 #include <mbgl/map/transform.hpp>
 #include <mbgl/util/chrono.hpp>
+#include <mbgl/map/update.hpp>
 #include <mbgl/util/geo.hpp>
 #include <mbgl/util/projection.hpp>
 #include <mbgl/util/noncopyable.hpp>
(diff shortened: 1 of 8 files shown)


incanus:
Nice, this sounds great!

kkaefer:
Pull request against master is now in #1155.

PR #1155, opened 2015-04-02
Only run Style::updateProperties when required

kkaefer:
We're currently running `Style::updateProperties` too often; whenever anything about the transform changes. However, we only need to run it on zoom, or when adding/removing classes.

### DIFF
--- a/android/cpp/jni.cpp
+++ b/android/cpp/jni.cpp
@@ -461,20 +461,6 @@ jobject JNICALL nativeGetLatLng(JNIEnv *env, jobject obj, jlong nativeMapViewPtr
     return ret;
 }
 
-void JNICALL nativeStartPanning(JNIEnv *env, jobject obj, jlong nativeMapViewPtr) {
-    mbgl::Log::Debug(mbgl::Event::JNI, "nativeStartPanning");
-    assert(nativeMapViewPtr != 0);
-    NativeMapView *nativeMapView = reinterpret_cast<NativeMapView *>(nativeMapViewPtr);
-    nativeMapView->getMap().startPanning();
-}
-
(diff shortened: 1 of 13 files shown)


kkaefer:
From #1176:

> This brings the easing transition code a bit closer to how easings work in gl-js. Instead of having an array of individual transitions for scale, rotate, and pan, there is a single transition function that does all the required calculations. This permits us to:
> 
> * Eliminate the "timeout" transition. (Fixes #126)
> * Remove the start/stopPanning() and similar functions from the API. The transform is capable of determining when a transition that affects the pan position is in effect. (Fixes #79)
> * Run style recalculations only when an ease transition that affects the zoom is in progress. (Fixes #1155)

kkaefer:
This branch also removes the `startScaling/Panning/Rotating` and the corresponding stop functions. These are used to enable/disable texture interpolation:



Previous code:



Note how the icon doesn't jump between pixels but rather moves gradually at the expense of sharpness.

kkaefer:
(Android is still calling `startPanning` et al, so the test fails.)

jfirebaugh:
@kkaefer Do you want to keep `startPanning` et al? I think we should at least reduce them to one `start/stopInteraction` method pair and require calls to always be paired (no timeout mechanism).

kkaefer:
I think we should keep it. Combining them to one call is fine, but we'd need to still track them individually in the view code.

jfirebaugh:
Replaced `startPanning` et al with `setGestureInProgress(bool)` and updated GLFW and iOS views to use it. Android has never used `startPanning` et al so I simply removed the bindings. I'll ticket a separate issue to add bindings and use `setGestureInProgress(bool)` on Android.

incanus:
We should also grab `handleQuickZoomGesture()` since this is a zooming gesture like the pinch.

PR #1177, opened 2015-04-02
Some minor cleanups

kkaefer:


src/mbgl/map/vector_tile.hpp
```diff
@@ -39,7 +39,7 @@ class VectorTileLayer : public GeometryTileLayer {
 
     std::string name;
     uint32_t extent = 4096;
-    std::unordered_map<std::string, uint32_t> keys;
+    std::map<std::string, uint32_t> keys;
```
jfirebaugh: What's the reason to change to `std::map` here?

kkaefer: Sorry, should have made this a separate commit. The reason is that it's faster on iOS.

Issue #896, opened 2015-02-17
Use cached tiles while doing a if-none-modified request


kkaefer:
We should show cached, but expired tiles immediately, do a `If-None-Modified` request meanwhile and update tile if changed. While this technically violates the HTTP spec, it will also ensure that the user sees a map as soon as possible. In most cases, the map is either fresh, or the visible differences are negligible.

2beaucoup:
It's either `If-None-Match` (ETag based) or `If-Modified-Since` (time based). Just sayin.

kkaefer:
@2beaucoup you're right, and the code uses the correct header names:https://github.com/mapbox/mapbox-gl-native/blob/a3ea4acb7cb4d57ca2f5808b9b600d7c1b7b3cc7/platform/default/http_request_curl.cpp#L341-L348

kkaefer:
This is implemented as part of #2617.

PR #1180, opened 2015-04-02
Make responses and response data shared

kkaefer:
In preparation for #896

### DIFF
--- a/gyp/platform-osx.gypi
+++ b/gyp/platform-osx.gypi
@@ -26,6 +26,7 @@
           '<@(uv_static_libs)',
         ],
         'ldflags': [
+          '-framework Foundation',
           '-framework ImageIO',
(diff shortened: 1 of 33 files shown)


jfirebaugh:
@kkaefer Is this still relevant / necessary? Any `unique` ⇢ `shared` change makes me nervous.

kkaefer:
The idea is that we need the Response data we loaded from the database in two places: For returning stale responses immediately, and for doing the `if-none-modified` request. I think this change is not critical since the Response object is a data-only object that has no side-effects on destruction and doesn't hold references to other objects.

jfirebaugh:
Stale.

kkaefer:
Implemented in 5173bf1bb8d21054b0dd6251d23eb37323d6c525

PR #1184, opened 2015-04-02
Style API cleanup, designables, inspectables

### LINKED ISSUE #929: Make MGLMapView IB_DESIGNABLE (for iOS XIBs)
Developers who use XIBs or storyboards will find it more awkward to use `MGLMapView` than `MKMapView` because custom views such as `MGLMapView` are rendered as plain white squares by default:

ib
_Left: `MGLMapView`. Right: `MKMapView`._

`MGLMapView` should be marked [`IB_DESIGNABLE`](https://developer.apple.com/library/ios/recipes/xcode_help-IB_objects_media/chapters/CreatingaLiveViewofaCustomObject.html) and draw something recognizable inside Interface Builder. It would be impractical to load actual map content synchronously within the short timeout, but even something mimicking the `MKMapView`’s translucent “blue box” would be easier to work with.

### LINKED ISSUE #1070: Support XIBs and storyboards on iOS
It’s relatively difficult to add an `MGLMapView` to a storyboard-based project. As expected, you can drag a `UIView` into a scene and set its custom class to “MGLMapView”. However, that just gets you a blank map. So you need to call `[mapView setAccessToken:@"…"]` at launch, and `[mapView setStyleJSON:nil]` sometime after that, but not too soon or you’ll hit #1014.

Basic support for XIBs and storyboards means:

- [ ] Make `accessToken` KVO-compliant, so it can be set via a user-defined runtime attribute in Interface Builder.
- [ ] `-setAccessToken:` should apply the default style if it’s called for the first time and no style has been applied. It’ll be tricky to fix this and avoid triggering #1014 for the usual `-initWithFrame:accessToken:` case, so #1014 is probably a blocker.

After we get basic support, we can look at enhancing the IB workflow via #929.

### LINKED ISSUE #1147: Crash on launch without access token — mbgl::Style::loadJSON()
Every time I launch the iOS demo app in various older iOS simulators (e.g., iPhone 4s or iPad 2 with iOS 8.2), I get the following crash:

```
Map (13)
#0	0x038fe695 in __cxa_throw ()
#1	0x0027b784 in mbgl::Style::loadJSON(unsigned char const*) at /Users/mxn/hub/mapbox-gl-native-2/src/mbgl/style/style.cpp:75
#2	0x0012c9ab in mbgl::Map::loadStyleJSON(std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char> > const&, std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char> > const&) at /Users/mxn/hub/mapbox-gl-native-2/src/mbgl/map/map.cpp:807
#3	0x0012c581 in mbgl::Map::reloadStyle() at /Users/mxn/hub/mapbox-gl-native-2/src/mbgl/map/map.cpp:797
#4	0x00125be4 in mbgl::Map::prepare() at /Users/mxn/hub/mapbox-gl-native-2/src/mbgl/map/map.cpp:822
#5	0x00125535 in mbgl::Map::run() at /Users/mxn/hub/mapbox-gl-native-2/src/mbgl/map/map.cpp:279
#6	0x0013b366 in mbgl::Map::start(bool)::$_5::operator()() const at /Users/mxn/hub/mapbox-gl-native-2/src/mbgl/map/map.cpp:183
…
```

`Map::data` is completely empty at this point: no `json` and no `base`. Those values wouldn’t be filled in until `-[MGLMapView initWithFrame::accessToken:styleJSON:]` calls `-setStyleJSON:` later on.

1ec5:
This PR addresses the following:

- Fixes a race condition on launch (fixes #1147)
- Cleans up various initializers and style APIs, removing raw JSON access and fixes KVO compliance (fixes #1070)
- Frictionless, codeless setup story for developers via designables and inspectables (fixes #929)

So, what’s the easiest way to get started with Mapbox GL?

1. Add MapboxGL to your Podfile and run `pod install`.
1. Drag a UIView out from the Object library into a `UIViewController` in a storyboard.
1. In the Identity inspector, set the custom class to “MGLMapView”.
1. In the Attributes inspector, set your access token. Optionally set an access token, initial coordinates, and initial zoom level.
1. Build and run.

designable

Will squash before merging.

### DIFF
--- a/gyp/platform-ios.gypi
+++ b/gyp/platform-ios.gypi
@@ -20,6 +20,7 @@
         '../include/mbgl/ios/MGLMapboxEvents.h',
         '../platform/ios/MGLMapboxEvents.m',
         '../include/mbgl/ios/MGLMapView.h',
+        '../include/mbgl/ios/MGLMapView+IBAdditions.h',
         '../platform/ios/MGLMapView.mm',
         '../include/mbgl/ios/MGLAnnotation.h',
         '../include/mbgl/ios/MGLUserLocation.h',
--- a/gyp/platform-ios.gypi
+++ b/gyp/platform-ios.gypi
@@ -30,6 +30,8 @@
         '../include/mbgl/ios/MGLMetricsLocationManager.h',
         '../platform/ios/MGLMetricsLocationManager.m',
         '../include/mbgl/ios/MGLTypes.h',
+        '../platform/ios/NSProcessInfo+MGLAdditions.h',
+        '../platform/ios/NSProcessInfo+MGLAdditions.m',
         '../platform/ios/NSString+MGLAdditions.h',
         '../platform/ios/NSString+MGLAdditions.m',
         '../platform/ios/vendor/SMCalloutView/SMCalloutView.h',
--- a/gyp/platform-ios.gypi
+++ b/gyp/platform-ios.gypi
@@ -30,6 +30,8 @@
         '../include/mbgl/ios/MGLMetricsLocationManager.h',
         '../platform/ios/MGLMetricsLocationManager.m',
         '../include/mbgl/ios/MGLTypes.h',
+        '../platform/ios/NSString+MGLAdditions.h',
+        '../platform/ios/NSString+MGLAdditions.m',
         '../platform/ios/vendor/SMCalloutView/SMCalloutView.h',
         '../platform/ios/vendor/SMCalloutView/SMCalloutView.m',
       ],
@@ -50,6 +52,7 @@
           '-framework MobileCoreServices',
           '-framework QuartzCore',
           '-framework SystemConfiguration',
+          '-ObjC',
         ],
       },
 
--- a/include/mbgl/ios/MGLMapView.h
+++ b/include/mbgl/ios/MGLMapView.h
@@ -39,9 +39,8 @@ IB_DESIGNABLE
 
 /** @name Authorizing Access */
 
-/** Sets a Mapbox API access token for the map view. 
-*   @param accessToken A Mapbox API token. */
-@property (nonatomic) IBInspectable NSString *accessToken;
+/** Mapbox API access token for the map view. */
+@property (nonatomic) NSString *accessToken;
 
 #pragma mark - Managing Constraints
 
@@ -166,10 +165,11 @@ IB_DESIGNABLE
 /** @name Styling the Map */
 
 /** Mapbox map ID of the style currently displayed in the receiver, or `nil` if the style does not have a map ID.
- 
-    The style may lack a map ID if it is located at an HTTP, HTTPS, or asset: URL. Use -styleURL to get the URL in these cases.
- */
-@property (nonatomic) IBInspectable NSString *mapID;
+*
+*   The style may lack a map ID if it is located at an HTTP, HTTPS, or local file URL. Use `styleURL` to get the URL in these cases. 
+*
+*   To display the default style, set this property to `nil`. */
+@property (nonatomic) NSString *mapID;
 
 /** Returns the URLs to the styles bundled with the library. */
 - (NSArray *)bundledStyleURLs;
--- a/include/mbgl/ios/MGLMapView.h
+++ b/include/mbgl/ios/MGLMapView.h
@@ -13,20 +13,21 @@
 *   Use of MGLMapView requires a Mapbox API access token. Obtain an access token on the [Mapbox account page](https://www.mapbox.com/account/apps/). If you instantiate an MGLMapView from Interface Builder, rendering of the map won't begin until the accessToken property has been set.
 *
 *   @warning Please note that you are responsible for getting permission to use the map data, and for ensuring your use adheres to the relevant terms of use. */
+IB_DESIGNABLE
 @interface MGLMapView : UIView
 
 #pragma mark - Initializing a Map View
 
 /** @name Initializing a Map View */
 
-/** Initialize a map view with a given frame, style URL, and access token.
+/** Initialize a map view with a given frame, access token, and style URL.
 *   @param frame The frame with which to initialize the map view.
 *   @param accessToken A Mapbox API access token.
 *   @param styleURL The map style URL to use. Can be either an HTTP/HTTPS URL or a Mapbox map ID style URL (`mapbox://<user.style>`).
 *   @return An initialized map view, or `nil` if the map view was unable to be initialized. */
 - (instancetype)initWithFrame:(CGRect)frame accessToken:(NSString *)accessToken styleURL:(NSURL *)styleURL;
 
-/** Initialize a map view with a given frame, the default style, and an access token.
+/** Initialize a map view with the default style given a frame and access token.
 *   @param frame The frame with which to initialize the map view.
 *   @param accessToken A Mapbox API access token.
 *   @return An initialized map view, or `nil` if the map view was unable to be initialized. */
@@ -49,7 +50,7 @@
 /** A view controller whose top and bottom layout guides to use for proper setup of constraints in the map view internals.
 *
 *   Certain components of the map view, such as the heading compass and the data attribution button, need to be aware of the view controller layout in order to avoid positioning content under a top navigation bar or a bottom toolbar. */
-@property (nonatomic, weak) UIViewController *viewControllerForLayoutGuides;
+@property (nonatomic, weak) IBOutlet UIViewController *viewControllerForLayoutGuides;
 
 #pragma mark - Accessing Map Properties
 
@@ -81,7 +82,7 @@
 /** @name Accessing the Delegate */
 
 // TODO
-@property(nonatomic, weak) id<MGLMapViewDelegate> delegate;
+@property(nonatomic, weak) IBOutlet id<MGLMapViewDelegate> delegate;
 
 #pragma mark - Manipulating the Visible Portion of the Map
 
@@ -134,7 +135,7 @@
 - (void)setDirection:(CLLocationDirection)direction animated:(BOOL)animated;
 
 /** Resets the map rotation to a northern heading. */
-- (void)resetNorth;
+- (IBAction)resetNorth;
 
 #pragma mark - Converting Map Coordinates
 
--- a/include/mbgl/ios/MGLMapView.h
+++ b/include/mbgl/ios/MGLMapView.h
@@ -176,7 +176,9 @@ IB_DESIGNABLE
 
 /** URL of the style currently displayed in the receiver.
 *
-*   The URL may be a full HTTP or HTTPS URL or a Mapbox URL indicating the style’s map ID (`mapbox://<user.style>`). To display the default style, set this property to `nil`. */
+*   The URL may be a full HTTP or HTTPS URL or a Mapbox URL indicating the style’s map ID (`mapbox://<user.style>`).
+*
+*   To display the default style, set this property to `nil`. */
 @property (nonatomic) NSURL *styleURL;
 
 #pragma mark - Annotating the Map
--- a/include/mbgl/ios/MGLMapView.h
+++ b/include/mbgl/ios/MGLMapView.h
@@ -19,20 +19,6 @@
 
 /** @name Initializing a Map View */
 
-/** Initialize a map view with a given frame, style, and access token.
-*   @param frame The frame with which to initialize the map view.
-*   @param accessToken A Mapbox API access token.
-*   @param styleJSON The map stylesheet as JSON text.
-*   @return An initialized map view, or `nil` if the map view was unable to be initialized. */
-- (instancetype)initWithFrame:(CGRect)frame accessToken:(NSString *)accessToken styleJSON:(NSString *)styleJSON;
-
-/** Initialize a map view with a given frame, bundled style name, and access token.
-*   @param frame The frame with which to initialize the map view.
-*   @param accessToken A Mapbox API access token.
-*   @param styleName The map style name to use.
-*   @return An initialized map view, or `nil` if the map view was unable to be initialized. */
-- (instancetype)initWithFrame:(CGRect)frame accessToken:(NSString *)accessToken bundledStyleNamed:(NSString *)styleName;
-
 /** Initialize a map view with a given frame, style URL, and access token.
 *   @param frame The frame with which to initialize the map view.
 *   @param accessToken A Mapbox API access token.
@@ -178,26 +164,19 @@
 
 /** @name Styling the Map */
 
-/** Sets the map style.
-*   @param styleJSON The map stylesheet as JSON text. */
-- (void)setStyleJSON:(NSString *)styleJSON;
+/** Mapbox map ID of the style currently displayed in the receiver, or `nil` if the style does not have a map ID.
+ 
+    The style may lack a map ID if it is located at an HTTP, HTTPS, or asset: URL. Use -styleURL to get the URL in these cases.
+ */
+@property (nonatomic) IBInspectable NSString *mapID;
 
-/** Returns the raw JSON style as a native dictionary object. */
-- (NSDictionary *)getRawStyle;
+/** Returns the URLs to the styles bundled with the library. */
+- (NSArray *)bundledStyleURLs;
 
-/** Sets the raw JSON style as a native dictionary object with a transition animation.
-*   @param style The style JSON as a dictionary object. */
-- (void)setRawStyle:(NSDictionary *)style;
-
-/** Returns the names of the styles bundled with the library. */
-- (NSArray *)bundledStyleNames;
-
-/** Sets the map style to a named, bundled style.
-*   @param styleName The map style name to use. */
-@property (nonatomic) IBInspectable NSString *styleName;
-
-/** Sets the map style URL to use.
-*   @param styleURL The map style URL to use. Can be either an HTTP/HTTPS URL or a Mapbox map ID style URL (`mapbox://<user.style>`). */
+/** URL of the style currently displayed in the receiver.
+    
+    The URL may be a full HTTP or HTTPS URL or a Mapbox URL indicating the style’s map ID (`mapbox://<user.style>`). To display the default style, set this property to `nil`.
+ */
 @property (nonatomic) NSURL *styleURL;
 
 #pragma mark - Annotating the Map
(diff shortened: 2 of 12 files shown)


1ec5:
Some details from individual commits:

- This change removes most of the ways you used to be able to apply a style to the map, so existing clients may need to be updated. `styleURL` (HTTP(S), mapbox:, asset:) is now the canonical way to apply a style, and `mapID` is a convenient shorthand for Mapbox-hosted styles. Internally, we now construct an asset: URL for the default style in lieu of a “bundled style name”.

- When an access token is set in the Attributes inspector (or as a user-defined runtime attribute), we draw some lovely Mapbox branding so you don’t have to manipulate an invisible rectangle. I’m open to suggestions on the design (cc @samanpwbb). In the absence of an access token, the view displays a helpful message with directions for obtaining and setting the access token.

- We have to turn off some functionality while rendering the designable:
  - We completely cut `MGLMapboxEvents` out of the picture. No one cares what beach the developer happens to be at while they’re building with Mapbox GL. :wink: In fact, if you touch Core Location anywhere inside a designable, it throws this exception (typo and all):
     ```
Application Specific Information:
*** Terminating app due to uncaught exception 'NSInternalInconsistencyException', reason: 'Non-UI clients cannont be autopaused'
terminating with uncaught exception of type NSException
abort() called

Application Specific Backtrace 1:
0   CoreFoundation                      0x0000000109ccda75 __exceptionPreprocess + 165
1   libobjc.A.dylib                     0x00000001089d8bb7 objc_exception_throw + 45
2   CoreFoundation                      0x0000000109ccd8da +[NSException raise:format:arguments:] + 106
3   Foundation                          0x00000001085f1b6f -[NSAssertionHandler handleFailureInMethod:object:file:lineNumber:description:] + 195
4   CoreLocation                        0x00000001150df265 CLClientGetCapabilities + 7718
```
   - While running inside IBDesignablesAgentCocoaTouch, we also avoid starting or updating `mbgl::Map` to avoid a potential minefield of problems, such as this exception:
     ```
     terminating with uncaught exception of type mbgl::util::ShaderException: Program plain failed to link: Validation Failed: Current draw framebuffer is invalid.
     ```
     It would be a mistake to rely on asynchronous network requests inside a designable anyways.
   - Xcode supports a `TARGET_INTERFACE_BUILDER` target conditional for just this situation, except that we’re distributing Mapbox GL as a prebuilt static library, so we can’t rely on conditional compilation, even if the library eventually makes its way into the CocoaPods-generated framework. Instead, we check whether we’re running inside the IBDesignablesAgentCocoaTouch process.

- `-[MGLMapView initWithFrame:]` is overridden to call `-commonInit`. We’ve marked this initializer unavailable in the header, but IB still calls it regardless.

- IB-specific inspectables are located a separate category that developers are not expected to import. (So it doesn’t appear in the umbrella header.) Currently these inspectables are order-dependent due to #1181. In the screencap above, I purposely set Zoom before Longitude or Latitude. The bearing also appears to be somewhat order-dependent, so I left it out – it probably isn’t common to need a specific bearing at launch.

File: platform/ios/MGLMapView_IBAdditions.m [hunk 1]
```diff
@@ -0,0 +1,57 @@
...
+{
+    return self.centerCoordinate.latitude;
+}
+
+- (void)setLatitude:(double)latitude
+{
+    NSNumber *pendingLongitude;
+    if ((pendingLongitude = objc_getAssociatedObject(self, @selector(longitude))))
+    {
+        objc_setAssociatedObject(self, @selector(latitude), nil, OBJC_ASSOCIATION_RETAIN_NONATOMIC);
```
1ec5: I know this associated object stuff is crazy, but I wanted two separate numeric inspectables instead of one comma-separated string inspectable (since it wouldn’t be clear whether latitude or longitude comes first). Setting only longitude or latitude has no effect: the coordinate remains (NaN, NaN).

incanus:
Will take a look shortly.

bleege:
Just seeing this now and will look into it.  First Impression though is that customizing Xcode for Access Tokens is awesome!  :+1:

platform/ios/MGLMapView_IBAdditions.m
bleege: Distinct latitude and longitude fields seems like the right way to go.

platform/ios/MGLMapboxEvents.m
```diff
@@ -174,17 +175,20 @@ + (instancetype)sharedManager {
...
     static MGLMapboxEvents *_sharedManager;
     dispatch_once(&onceToken, ^{
-        void (^setupBlock)() = ^{
-            _sharedManager = [[self alloc] init];
-            // setup dedicated location manager on first use
-            [MGLMetricsLocationManager sharedManager];
-        };
-        if ( ! [[NSThread currentThread] isMainThread]) {
-            dispatch_sync(dispatch_get_main_queue(), ^{
+        if ( ! NSProcessInfo.processInfo.mgl_isInterfaceBuilderDesignablesAgent) {
```
bleege: I'm not sure what wrapping this in a "not IBDesignableAgent" means.  I assume this is some extra process that is only run when Xcode is working with the code via IB?

bleege:
@1ec5 Just the one question ^^.  Other than that just please make sure [`MGLEventTypeLocation`]( https://github.com/mapbox/mapbox-gl-native/blob/1070-kvo/platform/ios/MGLMapboxEvents.m#L19) doesn't slip back to "Location" for a value when the merge happens.  Overall :+1:

platform/ios/MGLMapboxEvents.m
1ec5: That’s correct. Whenever you modify an instance of `MGLMapView` in a storyboard (for instance by resizing it), Interface Builder spawns an IBDesignablesAgentCocoaTouch process in the background that instantiates an `MGLMapView` and calls methods like `-drawRect:` on it. There are two hooks for IB-only code that should never run in front of a user: `TARGET_INTERFACE_BUILDER` (which we can’t use) and `-prepareForInterfaceBuilder`. Here’s more information about designables, in two flavors: [corporate](https://developer.apple.com/library/ios/recipes/xcode_help-IB_objects_media/Chapters/CreatingaLiveViewofaCustomObject.html), [hipster](http://nshipster.com/ibinspectable-ibdesignable/).

In NSProcessInfo+MGLAdditions, I added a category method that cheaply determines whether the `MGLMapView` is being run inside of IBDesignablesAgentCocoaTouch. I’m using it here to prevent the shared `MGLMapboxEvents` instance from ever being created. The assumption is that you’d never release an app named IBDesignablesAgentCocoaTouch. :smile:

bleege: That's really cool!  I didn't realize IB could do that.  Thanks for info @1ec5!

gyp/platform-ios.gypi
```diff
@@ -30,6 +30,12 @@
         '../include/mbgl/ios/MGLMetricsLocationManager.h',
         '../platform/ios/MGLMetricsLocationManager.m',
         '../include/mbgl/ios/MGLTypes.h',
+        '../platform/ios/NSProcessInfo+MGLAdditions.h',
+        '../platform/ios/NSProcessInfo+MGLAdditions.m',
+        '../platform/ios/NSString+MGLAdditions.h',
+        '../platform/ios/NSString+MGLAdditions.m',
+        '../include/mbgl/ios/MGLMapView_IBAdditions.h',
+        '../platform/ios/MGLMapView_IBAdditions.m',
```
incanus: Why not name these with the same convention? (`MGLMapView+IBAdditions.h` etc.)

include/mbgl/ios/MGLMapView_IBAdditions.h
```diff
@@ -0,0 +1,25 @@
+/*
+    This category header is named with an underscore instead of a plus sign
+    because Interface Builder sorts inspectables in ASCII alphabetical order by
+    the names of the headers that declare them, and we want the core MGLMapView
```
incanus: Oh, heh.

include/mbgl/ios/MGLMapView.h
```diff
@@ -178,27 +165,20 @@
...
-*   @param style The style JSON as a dictionary object. */
-- (void)setRawStyle:(NSDictionary *)style;
+/** Mapbox map ID of the style currently displayed in the receiver, or `nil` if the style does not have a map ID.
+ 
+    The style may lack a map ID if it is located at an HTTP, HTTPS, or asset: URL. Use -styleURL to get the URL in these cases.
+ */
+@property (nonatomic) IBInspectable NSString *mapID;
 
 /** Returns the names of the styles bundled with the library. */
 - (NSArray *)bundledStyleNames;
```
incanus: Should we modify this to include full paths or style URLs to be passed, since we can't make direct use of the names returned here anymore?

gyp/platform-ios.gypi
incanus: Answered in https://github.com/mapbox/mapbox-gl-native/pull/1184/files#r27702538.

include/mbgl/ios/MGLMapView_IBAdditions.h
1ec5: Yeah. :smile: Alternatively, I could redeclare the core properties here with the `IBInspectable` keyword and rename this file to match the + convention.

platform/ios/MGLMapView_IBAdditions.m
incanus: > This is why these properties are strictly IB-only. Maybe I should assert that the coordinate is still invalid before proceeding?

I think this might be a good idea, so that you never get someone trying to set these via KVC or something.

incanus:
Just one question @1ec5 while I'm still testing things out. Can you, like the `styleURL`, set the `mapID` to `nil` in order to use the default style?

incanus:
Cool, this feels great. Testing some of the in-code methods out a bit now. 

screen shot 2015-04-02 at 4 04 24 pm

incanus:
> Does AppleDoc not support doc comments on categories?

It does, but do we need them when they are intended for IB and not code? They are purely visual.

incanus:
> Can you, like the styleURL, set the mapID to nil in order to use the default style?

I see that the case is no, but it's an easy fix that I'm putting in now.

incanus:
I think functionality-wise we are good here now. I think we should shoot for the following API cleanups though: 

* [ ] Get rid of the associated objects stuff by making `latitude` and `longitude` private properties on `MGLMapView`. As an example of how this could work, setting `latitude` could: 

    ```objc
    - (void)setLatitude:(CLLocationDegrees)latitude
    {
        self.centerCoordinate = CLLocationCoordinate2DMake(latitude, self.centerCoordinate.longitude);
    }
    ```

    That would take care of the "staging" necessary. Then, these methods can be exposed only in the private `MGLMapView_IBAdditions.h` so that the general public doesn't use them. 

* [ ] We should still address `-bundledStyleNames` because its output is no longer of much use. We could instead make a method that returns values that could be passed to `-styleURL` such as `asset://styles/emerald-v7.json`.

1ec5:
`setLatitude:` can't just call `CLLocationCoordinate2DMake()` and set it as the center coordinate because the resulting center coordinate will be (NaN, NaN) – a coordinate is either valid or invalid, never half valid. That said, I'll replace the pending associated objects with pending ivars and declare them in the IB-specific header.

1ec5:
In other words, your proposed code above passes a NaN longitude into `CLLocationCoordinate2DMake()`.

incanus:
> In other words, your proposed code above passes a NaN longitude into `CLLocationCoordinate2DMake()`.

Ah, right. Sorry, I know you've been through the trenches on the order of operations here. Just ideas. 

>  I'll replace the pending associated objects with pending ivars and declare them in the IB-specific header.

This does sound more standard.

incanus:
> We could instead make a method that returns values that could be passed to `-styleURL` such as `asset://styles/emerald-v7.json`.

BTW on this front, I'm hesitant to expose `asset://` publicly, but I think it's kind of required here if we are going to bundle styles and we are going to require a URL set to use them.

1ec5:
> We could instead make a method that returns values that could be passed to `-styleURL` such as `asset://styles/emerald-v7.json`.

Rather than enumerating the styles at runtime, the most Cocoa-like way of doing this would be to define a series of `NSURL` constants, one for each bundled style. Or we’d have a class method `+URLForBundledStyleNamed:` à la `-[UIImage imageNamed:]`. Either approach would make sense for the real world case of applying a specific desired bundled style, but it would frustrate demo apps that cycle through all the possible bundled styles.

For now I’ve done as you suggest above but we should definitely revisit this method for beta 2.

incanus:
Cool, I think https://github.com/mapbox/mapbox-gl-native/commit/b376fa6add67e08c0d053df3ec3215c39b8b96e5 is a good first step solution here. 

>  it would frustrate demo apps that cycle through all the possible bundled styles.

Not sure I follow. Just because of the two-part setup? 

I do think something like `+URLForBundledStyleNamed:` would be good down the road, because we'll likely package and advertise the bundled styles as something a little dressier than filenames, e.g. `Emerald` or `Bright`. Then, that utility method could do the legwork of handling the `asset://` URL without the dev having to care much.

incanus:
I think this is good if you're happy with testing @1ec5. Want to squash and push a new branch or force-push here?

1ec5:
>>  it would frustrate demo apps that cycle through all the possible bundled styles.

> Not sure I follow. Just because of the two-part setup? 

No, because the demo apps all list the styles by name, so you have to do some parsing to get a human-readable name, meaning even tighter coupling to the nascent asset: URL standard.

> I do think something like `+URLForBundledStyleNamed:` would be good down the road

We have that already, as a private `MGLURLForBundledStyleNamed()` function in MGLMapView.mm. I didn’t expose it publicly because I’m not sure what to do with the versions: treat them as part of the name or as a separate piece of information?

platform/ios/MGLMapView.mm
```diff
@@ -138,69 +137,73 @@ - (instancetype)initWithFrame:(CGRect)frame accessToken:(NSString *)accessToken
...
     std::string styleURLString([[styleURL absoluteString] UTF8String]);
 
     if ( ! [styleURL scheme])
     {
-        mbglMap->setStyleURL(std::string("asset://") + styleURLString);
-    }
-    else
-    {
-        mbglMap->setStyleURL(styleURLString);
+        styleURLString = std::string("asset://") + styleURLString;
```
1ec5: I think we should get rid of this case. We can easily be consistent internally about using the asset: scheme. If the developer then passes in a relative URL, we can treat it as being relative to the main bundle rather than MapboxGL.bundle.

jfirebaugh: :+1:

DavidChouinard:
screen shot 2015-04-03 at 4 48 30 pm

1ec5:
@DavidChouinard How are you integrating Mapbox GL into your project? We’re planning to distribute a prebuilt static library via CocoaPods but haven’t uploaded the binary yet; in the meantime, you can follow these steps to use Mapbox GL as it’ll be used once it’s officially released:

- [Create a prebuilt mapbox-gl-ios-0.1.0.zip](https://github.com/mapbox/mapbox-gl-native/wiki/Releasing-Mapbox-GL-for-iOS#cocoapods) – everything but the last two lines.
- [Set up a private spec repo](http://guides.cocoapods.org/making/private-cocoapods.html) that contains a single Specs/0.1.0/MapboxGL.podspec. The podspec should be identical to [the version in master](https://github.com/mapbox/mapbox-gl-native/blob/master/ios/MapboxGL.podspec) save for the following line:  
```rb
m.source = { :http => "file:///absolute/path/to/mapbox-gl-native/build/ios/pkg/mapbox-gl-ios-#{m.version.to_s}.zip" }
```
- Point the Podfile to your private repo (from whence you pushed):
```rb
source 'file:///absolute/path/to/pods/'
pod 'MapboxGL', '~> 0.1.0'
use_frameworks!
```
- `pod install && open MyApp.xcworkspace`

----

If all that’s too much, there’s also the option of building from source (via `make iproj`) and turning the demo app into your app.

incanus:
To clarify, GL has to be included as a framework (even if built statically) right now in order for the IB stuff to work.

1ec5:
@DavidChouinard Before you go through the rigmarole above, though, I wonder if your project has `-ObjC` set in `ldflags` in your target’s build settings?

DavidChouinard:
@1ec5 Sorry for taking so long to try this.

After a bit of fiddling with Cocoapods, it works. I get the blue Mapbox view in Interface Builder. However, when running the application, I get the following console messages and the view is completely white:

```
2015-04-20 12:17:37.944 Demo1[1671:572449] Unknown class MGLMapView in Interface Builder file.
2015-04-20 12:17:37.960 Demo1[1671:572449] Failed to set (accessToken) user defined inspected property on (UIView): [<UIView 0x15457dc00> setValue:forUndefinedKey:]: this class is not key value coding-compliant for the key accessToken.
2015-04-20 12:17:37.960 Demo1[1671:572449] Failed to set (latitude) user defined inspected property on (UIView): [<UIView 0x15457dc00> setValue:forUndefinedKey:]: this class is not key value coding-compliant for the key latitude.
```
2015-04-20 12:17:37.961 Demo1[1671:572449] Failed to set (longitude) user defined inspected property on (UIView): [<UIView 0x15457dc00> setValue:forUndefinedKey:]: this class is not key value coding-compliant for the key longitude.
2015-04-20 12:17:37.961 Demo1[1671:572449] Failed to set (zoomLevel) user defined inspected property on (UIView): [<UIView 0x15457dc00> setValue:forUndefinedKey:]: this class is not key value coding-compliant for the key zoomLevel.
2015-04-20 12:17:37.961 Demo1[1671:572449] Failed to set (allowsRotating) user defined inspected property on (UIView): [<UIView 0x15457dc00> setValue:forUndefinedKey:]: this class is not key value coding-compliant for the key allowsRotating.

1ec5:
@DavidChouinard It looks like MapboxGL.framework or MapboxGL.a inside it still aren’t getting linked into your app for some reason. If you [instantiate the `MGLMapView` programmatically](https://github.com/mapbox/mapbox-gl-native/wiki/Installing-Mapbox-GL-for-iOS#in-code), does the app build successfully?

PR #1194, opened 2015-04-03
fixes #1181: fix lat/lng & zoom ordering problems

### LINKED ISSUE #1181: Cannot set zoom immediately after setting center coordinate
The following code takes you to the lovely town of Emerald, Victoria, Australia:

```objc
self.mapView.zoomLevel = 15;
[self.mapView setCenterCoordinate:CLLocationCoordinate2DMake(-23.526029999999999, 148.16224)];
```

vic

But reverse the two calls and you wind up off the coast of Papua New Guinea:

```obc
[self.mapView setCenterCoordinate:CLLocationCoordinate2DMake(-23.526029999999999, 148.16224)];
self.mapView.zoomLevel = 15;
```

png

`-[MGLMapView setCenterCoordinate:zoomLevel:animated:]` works fine of course, but I need something less atomic to make #929 work well. Otherwise you have to ensure that `zoomLevel` appears before `latitude` or `longitude` in the storyboard XML. (Interface Builder doesn’t let you rearrange user-defined runtime attributes using the inspector.)

incanus:
There were two things going on here: 

* On iOS, we limit the minimum zoom to `3` like MapKit so that we don't ever see world edges or a wrapping, repeating map. However, in various center set routines, we weren't checking that the level was at least `3` before trying to move the map, which could have limited it artificially and failed at setting the center desired. This abstracts out an `MGLMinimumZoom` constant since we use it in various places anyway, plus ensures that we are at the minimum zoom on iOS before changing viewport properties. 

* The second problem was more subtle. In C++, all of the setters and getters for center and zoom queried `Map`'s `transform` _except for `getLatLng()`_, which queried `state` instead. This could cause things to be out of sync when setting properties one after the other in the same run loop pass. 

/cc @kkaefer just to be sure.

### DIFF
--- a/platform/ios/MGLMapView.mm
+++ b/platform/ios/MGLMapView.mm
@@ -48,7 +48,7 @@ NSString *const MGLDefaultStyleMarkerSymbolName = @"default_marker";
 
 const NSTimeInterval MGLAnimationDuration = 0.3;
 const CGSize MGLAnnotationUpdateViewportOutset = {150, 150};
-const CGFloat MGLMinZoom = 3;
+const CGFloat MGLMinimumZoom = 3;
 
 NSString *const MGLAnnotationIDKey = @"MGLAnnotationIDKey";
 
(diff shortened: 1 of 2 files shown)


1ec5:
Do these methods need to account for `Map::getMinZoom()` and `Map::getMaxZoom()` as well?

incanus:
Good point @1ec5 — will adjust.

incanus:
Adjusted for `Map::getMinZoom()` at each query, even though this is unlikely to change. I don't think we need to worry about max because we're not (yet?) talking about letting the (Cocoa) user cap this and it is meant to be in the 18-to-maybe-someday-20s range.

PR #1201, opened 2015-04-03
iOS projects require -ObjC in OTHER_LDFLAGS

1ec5:
#1184 added `-[NSProcessInfo(MGLAdditions) mgl_isInterfaceBuilderDesignablesAgent]`, which only gets pulled in if you add `-ObjC` to `OTHER_LDFLAGS`.


1ec5:
Currently waiting on 844d8b9ceb308c5a3305ba4587f821f11452e317 to pass on Travis, then I’ll merge regardless of the build status on this PR.

PR #1214, opened 2015-04-06
Expose style class API publicly

1ec5:
Also added some convenience methods to match the underlying C++ API.

One question I have: does order matter for style classes? If so, we should add a method to insert at a particular index. If not, we should use an `NSSet` and probably a `std::set` beneath it.


1ec5:
OK, apparently style class order is undefined at the moment: mapbox/mapbox-gl-js#969. I’ll keep this PR as-is and we can revisit that issue later.

PR #1222, opened 2015-04-07
Refactor user location annotation; add accuracy ring

friedbunny:
In digging around the user location annotation to fix its shadow (#1204), I saw some things that were a little rough and ended up reworking the entire thing. Hopefully this can be considered for beta 2, if everything hasn't moved entirely GL by that time.

I cribbed from Sam Vermette's [SVPulsingAnnotationView](https://github.com/TransitApp/SVPulsingAnnotationView) and converted entirely to `CALayer`, which has a couple advantages for this application: smoother edges on the animated circles and the ability to use a single, more comprehensible API. 

Even though pre-rendered `UIImage` is the almost always the most performant and best way to go, I was surprised that I couldn't find any measurable difference or slowdown when using rasterized `CALayer` here. Please test and let me know how you measured!

[
screen shot 2015-04-06 at 10 49 59 pm
](https://www.dropbox.com/s/0bdljq2gsi1c7fz/1222-location-annotation.mp4?dl=1)

Things to look for, mostly inspired by MapKit:

- Hooked up the accuracy ring (MapKit-like filled circle, not an actual ring)
- Accuracy ring animates changes (but faster than MapKit)
- Linked the fading sonar-wave animation to accuracy ring size (when visible)
- Tinted pulsing inner dot scales more smoothly, fades opacity
- Inner dot and outer wave pulses are synced

Things needing attention:

- [x] Heading indication arrow/flashlight graphic
- [ ] Add a pop-in animation when it first appears
- [x] ~~Accuracy ring changes are out of sync (should probably be converted to a GL circle annotation)~~
- [ ] Correct heading-indicator direction on device rotation

Oh, I also tested this on my sluggish iPad Mini (A5) and even it seemed fine.


1ec5:
Thanks for completing and cleaning up this feature, @friedbunny! It looks good to me, although we’re going to save this PR for beta 2. We’d like to reduce code churn and focus on stability in beta 1. Feel free to run with the tail work you listed above in the meantime.

friedbunny:
Sure thing, I definitely agree that these enhancements are a better fit for beta 2. I'm planning on having a go at the heading indicator today/soon, we'll see what else comes up during that.

For now, I've got every `CALayer` set as `drawsAsynchronously = YES` for the potential performance gains, but I believe #1125 is the underlying issue causing accuracy ring resizing to lag behind input.

picciano:
If we could set the color (tint color?) of the location annotation from the client, that would be awesome. Just my $0.02.

friedbunny:
@picciano This (and the current implementation) both sync their color with the map view's tint color, which should probably be the same as the global app tint color. If you do something like `mapView.tintColor = [UIColor whateverColor]` that change will propagate to the location annotation and the :information_source: button in the lower-right corner.

I think it might be interesting to add setters for the different layers' colors, but I'm going to leave any developer API choices to the professionals. ;)

picciano:
@friedbunny Ah, didn't even realize that, perfect.

friedbunny:
Here's how the heading indicator is looking:

screen-shot-2015-04-08-at-9 47 53-pm

Initially went with the MapKit heading mask image resources to see how they looked (fine), but decided to do it the hard way and draw it myself. Doesn't support changing the heading accuracy or tint yet.

friedbunny:
Now supports heading indicator tint and accuracy adjustments.

[This](http://blog.pixelingene.com/2012/02/animating-pie-slices-using-a-custom-calayer/) looks useful if there's ever a want to animate the heading accuracy changes.

kristfal:
@friedbunny Tested this branch, and it is very neat – awesome! 

Found one issue with heading tracking though. If you enable viewport rotation in the test app and rotate the device to landscape, the heading indicator is 90deg off. Resetting heading tacking after rotation fixes this.

friedbunny:
Thanks @kristfal, I'll fix that when I come back to this (which should be next week or so).

kristfal:
@friedbunny Cheers, no need to rush it though ;)

friedbunny:
Rebased onto `master` 1d38b636402a6581c491185090f5c81495a07d13 (`git pull --rebase origin master`) and force-pushed. Still working out my personal git strategy. :neutral_face:

jfirebaugh:
@1ec5 Can you do a code review here and merge when ready?

platform/ios/MGLUserLocationAnnotationView.m
```diff
@@ -17,9 +18,15 @@ @interface MGLUserLocationAnnotationView ()
 
 @implementation MGLUserLocationAnnotationView
 {
+    CALayer *_headingIndicatorLayer;
+    CAShapeLayer *_headingIndicatorMaskLayer;
     CALayer *_accuracyRingLayer;
     CALayer *_dotBorderLayer;
     CALayer *_dotLayer;
+    
+    double oldHeadingAccuracy;
```
1ec5: Ivars should start with `_`.

platform/ios/MGLUserLocationAnnotationView.m
```diff
@@ -37,164 +44,283 @@ - (instancetype)initInMapView:(MGLMapView *)mapView
...
+            oldHeadingAccuracy = self.annotation.heading.headingAccuracy;
+        }
+        
+        // update accuracy ring
+        //
+        if (_accuracyRingLayer)
+        {
+            // FIX: This stops EVERYTHING... and that isn't necessarily necessary, now is it?
+            if (oldZoom == self.mapView.zoomLevel && oldHorizontalAccuracy == self.annotation.location.horizontalAccuracy)
+                return;
```
1ec5: Nit: use curly braces, since the `return` statement is on a separate line.

incanus:
While I love the looks of this, b2 is focused on features leading to increased adoption right now. This is a lot of code, makes things diverge from our iOS SDK codebase, and isn't directly related to adoption.

friedbunny:
@1ec5 Thanks, I'll push those changes shortly.

@incanus :cry: These changes are fairly significant, but self-contained. Even with the rotation fix I need to add, merging should be painless, whenever it does go in.

I've been using this in Treble and I'll keep putting it through its paces there, in the meantime. I don't *think* #1550 is related, but that will need investigating.

The only side effects I foresee at this point are potential performance issues with the switch to `CALayer`, but that hasn't been borne out in my use so far.

incanus:
> These changes are fairly significant, but self-contained. 

Yeah, agreed. Nothing personal :-) We need to get our b2 game rolling and ship quicker than b1, but once we get into that flow, we'll keep gobbling the submissions. 

But even now, b2 talk is cart before horse...

friedbunny:
Rebased this on `master` @ af89714b3935ea7b1ca312f9a8ce028aec04d812. Fixed heading indicator not rotating.

friedbunny:
Closed in favor of squash in #2010.

Issue #1093, opened 2015-03-25
navigation bar user location tracking mode button


incanus:
Either port `RMUserTrackingBarButtonItem` from the iOS SDK or otherwise meet the capabilities of `MKUserTrackingBarButtonItem` such as pairing with a map view, automatic state update using KVO off of the map view, and in-progress animations.

friedbunny:
Answered a question about this [on Stack Overflow](http://stackoverflow.com/a/35727833/2094275) just now. We have [code available in ios-app](https://github.com/mapbox/mapbox-gl-native/blob/5ac6fc523dea6374da7ba15a7c0433b115987ed2/platform/ios/app/MBXViewController.mm#L628-L655) that gets us most of the way to a `MGLUserTrackingBarButtonItem`, we just haven’t put it together in the SDK itself yet.

jaltin:
Any idea of if/when this might get implemented?

PR #1239, opened 2015-04-09
Add simple tracking mode button to demo app

friedbunny:
This doesn't address #1093 by actually implementing `RMUserTrackingBarButtonItem` or similiar, it only updates the demo app's locate button to sync with the current tracking mode.

In the course of working with tracking mode code it's been a bit frustrating to mash the button and not have any feedback, so I put together this extremely basic switch.


friedbunny:
There's a size discrepancy for `TrackingHeadingMask` that makes it look a little janky, but it's functional!

1ec5:
Thanks!

PR #1244, opened 2015-04-09
Replace uv_messenger with AsyncQueue; rewrite remaining uv_* in C++

jfirebaugh:


jfirebaugh:
@kkaefer @tmpsantos Can you review please?

src/mbgl/util/worker.cpp
```diff
@@ -0,0 +1,58 @@
...
+        thread.join();
+    }
+
+    queue->stop();
+}
+
+void Worker::send(std::function<void ()> work,
+                  std::function<void ()> after) {
+    MBGL_VERIFY_THREAD(tid);
+    channel.send({work, after});
```
kkaefer: `send(nullptr, nullptr);` will:

1. terminate the worker
2. crash the application because `after()` on line 8 is invalid

kkaefer: We should either `assert` both `work` and `after`, use shim functions when the functions objects are empty, or discard the call altogether.

jfirebaugh: There's a guard condition for `after` in the thread function. I'll add an assert for `work`.

jfirebaugh:
Tests were hanging because `Worker` didn't do an `unref`/`ref` dance on the `AsyncQueue`. The queue needs to be referenced iff there is a pending work item.

PR #1250, opened 2015-04-10
Separate objects from Thread management

kkaefer:
We should separate objects like the SQLiteCache, DefaultFileSource and the Map object from thread management considerations.


jfirebaugh:
This is cool, I have an alternate factoring of these additions to suggest that I'll push up in a separate branch.

jfirebaugh:
@kkaefer - GCC 4.9 doesn't look like it's going to pan out. It [produces](https://travis-ci.org/mapbox/mapbox-gl-native/jobs/58022326#L1193):

```
In file included from ../../linux/main.cpp:10:0:
../../include/mbgl/util/thread.hpp: In constructor ‘mbgl::util::Thread<T>::Thread(Args&& ...) [with Args = {const char (&)[19]}; Object = mbgl::SQLiteCache]’:
../../include/mbgl/util/thread.hpp:40:1: internal compiler error: in gimple_expand_cfg, at cfgexpand.c:5658
 Thread<Object>::Thread(Args&&... args)
 ^
```

Also, installing the gcc-4.9 package [breaks compilation with clang](https://travis-ci.org/mapbox/mapbox-gl-native/jobs/58022324#L1094), due to a [bug in clang](https://bugs.launchpad.net/ubuntu/+source/llvm-toolchain-snapshot/+bug/1401202) (issue is for clang 3.5, but also exists in 3.4 apparently, which is what we're using).

jfirebaugh:
The reason I prefer to merge `Thread` and `RunLoop` is it makes the interface safer -- you can enforce that the only access to the wrapped object is through `invoke`, and `invoke` can be [defined in a way](https://github.com/mapbox/mapbox-gl-native/blob/8c3cfb84f1c3dfe41375a9689b54bcf5ee99eb8f/src/mbgl/util/threaded_run_loop.hpp#L25-L27) that permits only calling a member function on the thread, rather than arbitrary code with potential cross-thread lambda bindings. The member-function-based definition of `invoke` could also potentially support member functions with return values, automatically turning them into futures or callbacks.

jfirebaugh:
@kkaefer - I looked at this deeper this afternoon and I

* Agree that `Thread<>` and `RunLoop` should stay separate.
* See the need for access to the RunLoop from within Impl methods as closely related to the issue of transferring the result of a threaded function call back to the thread that the function call was invoked from. Closing the message-passing loop, essentially.
* Still think that `Thread<>` should expose a "pointer-to-member-call only" `invoke` interface for maximum safety.

I think we can accomplish all of these simultaneously. I've pushed up a [new implementation](https://github.com/mapbox/mapbox-gl-native/commit/3bfaea37bdbdb487ed8db03662f0b6a09d103e2c) to `more-separate` that's close (hopefully) to complete. There are a few tricky templating issues that I haven't cracked, but I don't think they are insoluble.

Summary of changes vs. your branch:

* `RunLoop` interface is streamlined and minimized. I removed the `friend` stuff and `protected`. `RunLoop` is now completely wrapped by `Thread<>`, so I don't think that's necessary.
* Added:
 * A thread local `RunLoop* current` variable which holds the run loop for the current thread. (This is currently erroring; need to figure out how to enable the `thread_local` storage specifier or manually implement thread locality.)
 * A templated `RunLoop::invoke` overload which accepts a `callback` and _invokes the callback on the invoking thread's `RunLoop`_. **This closes the loop of two evented threads doing message passing and is the key change.**
* Adjust `Thread<>` and users of it to this new mechanism.
 * Implement a "pointer-to-member-call only" `Thread<>::invoke` in terms of `RunLoop::invoke`.
 * Impl's return results directly rather than accepting a callback. See e.g. `SQLiteCache::Impl::processGet`.
 * Public interfaces use the callbacked version of `Thread<>::invoke` to obtain the result.

Then the remaining need for access to the underlying loop is only for doing other libuv things, and can be accomplished by adding a `uv_loop_t*` argument to the Impl methods that require it.

Let me know what you think, and if you can figure out how to fix the remaining compile errors.

kkaefer:
> This closes the loop of two evented threads doing message passing and is the key change.

Does that mean we can only invoke function on thread loops iff the current thread also has a runloop? That'd mean that we couldn't use the `Thread<RunLoop>` object for the Map thread, since the main thread which starts the Map thread doesn't have a RunLoop.

> Impl's return results directly rather than accepting a callback. See e.g. SQLiteCache::Impl::processGet.

I like this change, much easier interface and we avoid all the errors with accidentally not calling the callback function. However, this also means that we can't pass on the callback if the current function can't satisfy the request; it essentially forces all invoke requests to be synchronous.

kkaefer:
> if you can figure out how to fix the remaining compile errors.

The underlying issue here is that it's really hard to forward & move the arguments correctly; `std::bind` produces something that is movable-only and that we can't convert to a `std::function<>`, since `std::function<>` requires copy-constructability.

jfirebaugh:
> Does that mean we can only invoke function on thread loops iff the current thread also has a runloop? 

It means we can only invoke functions with a callback iff the current thread has a runloop. Fire-and-forget invocations are okay.

If we needed callbacks for the mean thread, we'd have to have some (platform-specific) means of dispatching on the native event loop of the main thread.

> However, this also means that we can't pass on the callback if the current function can't satisfy the request; it essentially forces all invoke requests to be synchronous.

Not being able to pass on the callback directly is good -- the callback needs to be invoked on the invoking thread, and passing it on would hide this requirement.

Maybe we could define another overload for `RunLoop::invoke`/`Thread<>::invoke` that pass a wrapped callback:

```
// Invoke fn, passing a wrapper that invokes callback on the current run-loop
void RunLoop::invoke(std::function<void(Callback)>&& fn, Callback callback) {
        RunLoop& outer = *current.get();
        invoke(std::bind(fn, [&] () { outer.invoke(callback); }));
}

template <class Object>
template <class... Args>
void Thread<Object>::invoke(void (Object::*fn)(Args..., Callback), Args&&... args, Callback) {
    loop->invoke(std::bind(fn, object, ::std::forward<Args>(args)...), Callback);
}

void A::method(Callback callback) {
  thread->invoke(&A::method, callback);
}

void A::Impl::method(Callback callback) {
  asyncFunction(callback);
}
```

Not sure if that actually works though.

> The underlying issue here is that it's really hard to forward & move the arguments correctly

Someone has to have solved the problem of generic cross-thread move semantics. I'll research this more today.

jfirebaugh:
Best reference so far: [this article by Herb Sutter](http://www.drdobbs.com/parallel/prefer-using-active-objects-instead-of-n/225700095) ([this comment](http://herbsutter.com/2010/07/12/effective-concurrency-prefer-using-active-objects-instead-of-naked-threads/#comment-2162) starts to touch on the problem we're facing).

jfirebaugh:
Asked a [Stack Overflow question](http://stackoverflow.com/questions/29639519/supporting-unique-ptr-in-an-implementation-of-active-object).

tmpsantos:
Chromium folks had the same problem:
https://code.google.com/p/chromium/codesearch#chromium/src/base/bind_helpers.h&l=346

jfirebaugh:
So close! https://github.com/mapbox/mapbox-gl-native/commit/ffd18363b26bcb15f9b81ee2e86e2859bbc55630

I _should_ be able to replace [these C++14 move-capturing lambdas](https://github.com/mapbox/mapbox-gl-native/blob/ffd18363b26bcb15f9b81ee2e86e2859bbc55630/src/mbgl/util/run_loop.hpp#L31-L32) with [this trick](http://stackoverflow.com/a/12744730/52207) but for some reason it doesn't work (runtime crash).

springmeyer:
I think @artemp and @kkaefer ruled this out, but its still worth mentioning that `Fn&& fn` looks like a "universal reference" to me. Which means it's not garunteed to be an rvalue: https://isocpp.org/blog/2012/11/universal-references-in-c11-scott-meyers

jfirebaugh:
This is how I would expect to be able to define `invokeWithResult` using pure C++11 and two instances of the [move-capturing trick](http://stackoverflow.com/questions/8236521/how-to-capture-a-unique-ptr-into-a-lambda-expression/12744730#12744730):

```
    // Invoke fn() in the runloop thread, then invoke callback(result) in the current thread.
    template <class Fn, class R>
    void invokeWithResult(Fn&& fn, std::function<void (R)> callback) {
        RunLoop* outer = current.get();
        invoke(std::bind([callback, outer] (Fn& fn_) {
            outer->invoke(std::bind([callback] (R& result) {
                callback(std::move(result));
            }, std::move(fn_())));
        }, std::move(fn)));
    }
```

But this fails to compile. The outer std::bind apparently winds up not having a eligible `operator()` template.

```
/Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/../include/c++/v1/functional:2084:9: note: candidate template ignored: substitution failure [with _Args = <>]: implicit instantiation of undefined template 'std::__1::__bind_return<(lambda at src/mbgl/util/run_loop.hpp:31:26), std::__1::tuple<std::__1::__bind<std::__1::unique_ptr<mbgl::Response, std::__1::default_delete<mbgl::Response> > (mbgl::SQLiteCache::Impl::*&)(const mbgl::Resource &), mbgl::SQLiteCache::Impl *&, const mbgl::Resource &> >, std::__1::tuple<>, false>'
        operator()(_Args&& ...__args)
        ^
```

jfirebaugh:
Looks like we don't need to move-capture there anyway. A copying capture works fine.

82c6bdb04d7ee426dd27f6303a3e3da6309ea214

jfirebaugh:
Storage tests are now failing because of an attempt to dispatch a result back to the main test thread, which doesn't have a run loop. Not sure how this used to work...

jfirebaugh:
@kkaefer You like?

I assume we'll convert the Map thread to this after merging #1065?

Issue #1256, opened 2015-04-11
iOS: Zooming out beyond ~z10 with tracking mode enabled will force zoom in, reset tracking


friedbunny:
When any `MGLUserTrackingMode` is enabled and the user zooms out beyond [`round(self.zoomLevel) >= 10`](https://github.com/mapbox/mapbox-gl-native/blob/e6e340afd17a29ee4d183540e716376cdb1ef88e/platform/ios/MGLMapView.mm#L1820-L1856), we force the map to zoom to "near accuracy confidence" — usually about z18.

The eventual call to `[MGLMapView setCenterCoordinate:zoomLevel:animated:]` also [mistakenly resets the tracking mode](https://github.com/mapbox/mapbox-gl-native/blob/e6e340afd17a29ee4d183540e716376cdb1ef88e/platform/ios/MGLMapView.mm#L1184).

friedbunny:
[Lies!](https://github.com/mapbox/mapbox-gl-native/blob/e6e340afd17a29ee4d183540e716376cdb1ef88e/platform/ios/MGLMapView.mm#L1222) :broken_heart:

```objc
- (void)zoomToSouthWestCoordinate:(CLLocationCoordinate2D)southWestCoordinate northEastCoordinate:(CLLocationCoordinate2D)northEastCoordinate animated:(BOOL)animated
{
    // NOTE: does not disrupt tracking mode
...
```

PR #1257, opened 2015-04-11
Refs #1256 - Fix some erronenous tracking mode resets

friedbunny:
- Adds private `preservingTracking:` version of `setCenterCoordinate:zoomLevel:animated:`
- Changes `zoomToSouthWestCoordinate:northEastCoordinate:animated:` to above `preservingTracking:`, fixes an untended tracking mode reset
- Removes tracking mode reset from `resetNorthAnimated:`
  - [`handleCompassTapGesture:`](https://github.com/mapbox/mapbox-gl-native/blob/e6e340afd17a29ee4d183540e716376cdb1ef88e/platform/ios/MGLMapView.mm#L612-L617) was intending to simply bump down to `MGLUserTrackingModeFollow` if it was following with heading, but it was getting clobbered by `resetNorthAnimated:`.

Only fixes half of #1256.

### DIFF
--- a/platform/ios/MGLMapView.mm
+++ b/platform/ios/MGLMapView.mm
@@ -1106,8 +1106,6 @@ mbgl::DefaultFileSource *mbglFileSource = nullptr;
 
 - (void)resetNorthAnimated:(BOOL)animated
 {
-    self.userTrackingMode = MGLUserTrackingModeNone;
-
     CGFloat duration = (animated ? MGLAnimationDuration : 0);
 
     mbglMap->setBearing(0, secondsAsDuration(duration));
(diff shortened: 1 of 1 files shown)


friedbunny:
May be better to move the tracking mode check from `handleCompassTapGesture` to `resetNorthAnimated`.

jfirebaugh:
@1ec5 / @friedbunny Can you coordinate what remains to be done here and get this merged?

1ec5:
@friedbunny, it looks like #1783 took care of the `-zoomToSouthWestCoordinate:northEastCoordinate:animated:` case. Do you think we still need a way to prevent resetting the user tracking mode generally? If so, that version of the method should be exposed publicly. Otherwise, this PR can be closed.

friedbunny:
The `resetNorthAnimated:` tracking mode reset still needs to be removed, but I've broken that out into a new pull request (#1809). Closing here.

PR #1260, opened 2015-04-13
define MACOSX_DEPLOYMENT_TARGET in common.gypi

mikemorris:
Needed to prevent node-gyp from overriding this everywhere with '10.5'
from https://github.com/joyent/node/blob/master/common.gypi

Does this seem reasonable @springmeyer @incanus?

### DIFF
--- a/gyp/common.gypi
+++ b/gyp/common.gypi
@@ -23,6 +23,7 @@
           ],
           'GCC_WARN_PEDANTIC': 'YES',
           'GCC_WARN_UNINITIALIZED_AUTOS': 'YES_AGGRESSIVE',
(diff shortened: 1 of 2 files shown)


springmeyer:
@mikemorris - that is great `target_defaults` works. Makes me wonder if putting the override inside https://github.com/mapbox/mapbox-gl-native/blob/master/gyp/common.gypi#L9-L23 would also work?

mikemorris:
Oh, didn't see there was already a `target_defaults` in common.gypi, should work there too!

mikemorris:
Rebased to simply and move to common.gypi, working in node-gyp now still!

PR #1065, opened 2015-04-15
Better separation between UI thread and Map thread

kkaefer:
TODO list:

* [x] `data.callback` is not threadsafe
* [x] refactor `View` and subclasses to interact with a single thread only (either main or map)


jfirebaugh:
How do you decide what goes in `MapData` and what goes in `MapContext`?

kkaefer:
All data that is shared between the main thread and the map thread goes into MapData, the rest goes into MapContext. Eventually this should be 0, so MapData goes away and data lives either on the main thread (in the `Map` object) or on the map thread (in the `MapContext` object).

jfirebaugh:
:+1:, sounds good.

kkaefer:
This is more of a work in progress and not intended to be merged as is.

jfirebaugh:
What tasks are remaining, in your mind?

kkaefer:
Mostly removal of accessing `MapContext` from `Map`

jfirebaugh:
I started looking at what needs to happen to privatize the remaining members of MapContext. The asyncs were the first target. They are currently public because they are used for control flow in the static rendering case. I think we should replace this with explicit checks of whether all resources are ready (i.e. port [`map.loaded()`](https://github.com/mapbox/mapbox-gl-js/blob/255e28e4cf97a24124973dcf9f54839388f3fed6/js/ui/map.js#L669-L675) from gl-js and call the `renderStill` callback at the equivalent point to where gl-js [emits the load event](https://github.com/mapbox/mapbox-gl-js/blob/255e28e4cf97a24124973dcf9f54839388f3fed6/js/ui/map.js#L722-L725).

jfirebaugh:
`View` and subclasses need to be refactored as a part of this split too. Right now they are a mix of code executed on the map thread and rendering thread (via abstract functions called by `MapContext`).

jfirebaugh:
```
    // Run the event loop once more to make sure our async delete handlers are called.
    uv_run(env.loop, UV_RUN_ONCE);
```

Is this for sure necessary?

mikemorris:
@jfirebaugh As you're digging into this, I would love if you could keep an eye out for threading issues that could possibly be causing https://github.com/mapbox/api-gl/issues/16 and https://github.com/mapbox/node-mapbox-gl-native/issues/114.

kkaefer:
>     // Run the event loop once more to make sure our async delete handlers are called.
>     uv_run(env.loop, UV_RUN_ONCE);
> Is this for sure necessary?

Not 100% sure, but there could be situations where we end an event loop, and destroy it, but don't run the callbacks from `uv_close(...)`.

tmpsantos:
Why not call MapContext Map::Impl like we do for the other objects?

jfirebaugh:
@mikemorris Will do.

> Why not call MapContext Map::Impl like we do for the other objects?

Yeah, I was considering renaming it later on, when everything is working again.

jfirebaugh:
Got it to compile for the first time using `Thread<MapContext>`. First crash is related to view functions calling back into map functions from the rendering thread:

```
#0	0x0000000100245e4f in void mbgl::util::Thread<mbgl::MapContext>::invoke<void (mbgl::MapContext::*)(mbgl::Update), mbgl::Update>(void (mbgl::MapContext::*)(mbgl::Update), mbgl::Update&&) at /Users/john/Development/mapbox-gl-native/src/mbgl/util/thread.hpp:47
#1	0x00000001002424bf in mbgl::Map::resize(unsigned short, unsigned short, float, unsigned short, unsigned short) at /Users/john/Development/mapbox-gl-native/src/mbgl/map/map.cpp:172
#2	0x00000001002fb160 in mbgl::View::resize(unsigned short, unsigned short, float, unsigned short, unsigned short) at /Users/john/Development/mapbox-gl-native/src/mbgl/map/view.cpp:24
#3	0x00000001004ffc73 in mbgl::HeadlessView::discard() at /Users/john/Development/mapbox-gl-native/platform/default/headless_view.cpp:212
#4	0x000000010050188a in mbgl::HeadlessView::activate() at /Users/john/Development/mapbox-gl-native/platform/default/headless_view.cpp:308
#5	0x0000000100267d2b in mbgl::MapContext::MapContext(uv_loop_s*, mbgl::Environment&, mbgl::View&, mbgl::MapData&, bool) at /Users/john/Development/mapbox-gl-native/src/mbgl/map/map_context.cpp:56
#6	0x0000000100269184 in mbgl::MapContext::MapContext(uv_loop_s*, mbgl::Environment&, mbgl::View&, mbgl::MapData&, bool) at /Users/john/Development/mapbox-gl-native/src/mbgl/map/map_context.cpp:62
#7	0x00000001002453de in void mbgl::util::Thread<mbgl::MapContext>::run<std::__1::tuple<mbgl::Environment&, mbgl::View&, mbgl::MapData&, bool&>, 0ul, 1ul, 2ul, 3ul>(std::__1::tuple<mbgl::Environment&, mbgl::View&, mbgl::MapData&, bool&>&&, (anonymous namespace)::index_sequence<0ul, 1ul, 2ul, 3ul>) at /Users/john/Development/mapbox-gl-native/src/mbgl/util/thread.hpp:120
#8	0x0000000100266213 in mbgl::util::Thread<mbgl::MapContext>::Thread<mbgl::Environment&, mbgl::View&, mbgl::MapData&, bool&>(std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char> > const&, mbgl::Environment&&&, mbgl::View&&&, mbgl::MapData&&&, bool&&&)::'lambda'()::operator()() const at /Users/john/Development/mapbox-gl-native/src/mbgl/util/thread.hpp:107
#9	0x0000000100265e72 in std::__1::__invoke<mbgl::util::Thread<mbgl::MapContext>::Thread<mbgl::Environment&, mbgl::View&, mbgl::MapData&, bool&>(std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char> > const&, mbgl::Environment&&&, mbgl::View&&&, mbgl::MapData&&&, bool&&&)::'lambda'()>(decltype(std::__1::forward<mbgl::util::Thread<mbgl::MapContext>::Thread<mbgl::Environment&, mbgl::View&, mbgl::MapData&, bool&>(std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char> > const&, mbgl::Environment&&&, mbgl::View&&&, mbgl::MapData&&&, bool&&&)::'lambda'()>(fp)(std::__1::forward<>(fp0))), mbgl::util::Thread<mbgl::MapContext>::Thread<mbgl::Environment&, mbgl::View&, mbgl::MapData&, bool&>(std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char> > const&, mbgl::Environment&&&, mbgl::View&&&, mbgl::MapData&&&, bool&&&)::'lambda'()&&) [inlined] at /Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/../include/c++/v1/__functional_base:413
#10	0x0000000100265e66 in _ZNSt3__116__thread_executeIZN4mbgl4util6ThreadINS1_10MapContextEEC1IJRNS1_11EnvironmentERNS1_4ViewERNS1_7MapDataERbEEERKNS_12basic_stringIcNS_11char_traitsIcEENS_9allocatorIcEEEEDpOT_EUlvE_JEJEEEvRNS_5tupleIJT_DpT0_EEENS_15__tuple_indicesIJXspT1_EEEE [inlined] at /Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/../include/c++/v1/thread:332
#11	0x0000000100265e56 in std::__1::__thread_proxy<std::__1::tuple<mbgl::util::Thread<mbgl::MapContext>::Thread<mbgl::Environment&, mbgl::View&, mbgl::MapData&, bool&>(std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char> > const&, mbgl::Environment&&&, mbgl::View&&&, mbgl::MapData&&&, bool&&&)::'lambda'()> >(void*, void*) at /Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/../include/c++/v1/thread:342
#12	0x00007fff8c8e0268 in _pthread_body ()
#13	0x00007fff8c8e01e5 in _pthread_start ()
#14	0x00007fff8c8de41d in thread_start ()
```

Going to start looking at how @kkaefer is restructuring this in `transform-on-main`.

jfirebaugh:
What is the purpose of `Environment::abandon*` as opposed to immediately disposing the resources?

jfirebaugh:
@kkaefer Can you give an outline of your strategy with `transform-on-main`?

jfirebaugh:
@mikemorris @kkaefer Is it essential that `Map::renderStill` be asynchronous? It's not reentrant, and I imagine that node-mbgl is running the maps on a thread pool anyway, right?

mb12:
Environment::abandonVAOs was added to make sure that buffers for DebugBucket get also deleted in the mapThread.  Prior to this change,  it used to result in "Opengl thread conflict" error.

https://github.com/mapbox/mapbox-gl-native/pull/1016

jfirebaugh:
Current status: GLFW-based app is working. Headless rendering hangs. iOS / Android are probably failing to build.

mikemorris:
> Is it essential that Map::renderStill be asynchronous? It's not reentrant, and I imagine that node-mbgl is running the maps on a thread pool anyway, right?

So I know there were problems with simply using `NanAsyncWorker` with Node's default threadpool, so the current architecture is to have `renderStill` actually perform work on the `Map` thread to not block Node.js.

kkaefer:
I'd like to keep `renderStill` asynchronous. Even if we'd offload the then blocking call to `renderStill()` to a libuv worker thread, it's still blocking which isn't something that you can do with libuv. It'd mean that the libuv threadpool is blocked by wait calls, which means we couldn't do any IO either anymore (i.e. we can't satisfy the dependencies the running render call needs), resulting in a deadlock.

jfirebaugh:
Current status:

* API test fails with the exception "OpenGL context was already current". This comes from trying to resize the HeadlessView while it's activated by the map thread. @kkaefer, I would like to know what you are thinking here with regards to `transform-on-main` and the fact that view objects right now have unclear thread affinity. This is my main blocker on futher progress.
* Headless tests are sometimes hanging.
* Headless tests sometimes render prematurely, before all tiles are finished loading.

jfirebaugh:
The hangs appear to be from a uv handle leak somewhere. The map run loop sometimes fails to exit. I set a breakpoint [here](https://github.com/mapbox/mapbox-gl-native/blob/ab36908391c832444d25744ecd519e1e55ebbf12/src/mbgl/util/run_loop.cpp#L35), and when the `async.unref()` happens there are still one or two other active handles.

jfirebaugh:
The handle leak is related to this sequence of events:

| Main Thread | Map Thread |
| --------------- | ----------------- |
| `Map::start()` | |
| `Map::setStyleJSON()` |
| | 1st `MapContext::update()` |
| | 2nd `MapContext::update()` |

In this sequence the 1st `MapContext::update()` (triggered by MapContext creation) creates the `Style` object, and then the 2nd  `MapContext::update()` (triggered by `setStyleJSON`) destroys it and creates a new one. This is not ideal.

What needs to happen is that `setStyleJSON` `invoke`s a `MapContext` method with the style JSON, instead of setting it on `data`, where it becomes readable by the 1st `MapContext::update()`. However, this imposes the constraint that the map thread is active when `Map::setStyleJSON()` is called; you wouldn't be able to call `setStyleJSON` and then `start`.

**Proposal: I think we should get rid of `start`/`terminate` and use `pause`/`resume` semantics instead.** The map thread should always be running, and be able to receive messages. When paused, it should just refuse to render.

kkaefer:
The idea for the transform change is to change how we drive the rendering: Instead of the map thread rendering whenever it feels like it, rendering is always triggered and controlled by (but not run on!) the main thread. The main thread also exclusively controls and has access to the transform object (so we can remove the locking). Whenever the main thread requires a frame to be rendered, we're snapshotting the current transform state and copy/move it over to the map thread. In some situations, the map thread needs to trigger a rerender (e.g. because new data arrived). In those cases, we call `view.invalidate()` and leave it up to the View implementation to trigger a render from the main thread.

incanus:
> The map thread should always be running, and be able to receive messages. When paused, it should just refuse to render.

This seems simple and less error-prone or drastic.

1ec5:
> Proposal: I think we should get rid of `start`/`terminate` and always use `pause`/`resume` semantics instead.

Would pausing and resuming avoid destroying the tile cache? That would address #1320.

jfirebaugh:
Yes, it would fix #1320 and simplify a bunch of things.

jfirebaugh:
I've reached the main end goal I wanted here, which was to ensure that Style outlives its worker tasks. There are a lot of cleanup tasks remaining, and a lot more threadsafety fixes to make, but this should take care of two major crashers we've been seeing.

I am seeing some instability in headless tests on linux still. Backtrace of the crash looks like:

```
Thread 1 (Thread 0x2b2bbe125300 (LWP 13514)):
#0  0x00002b2bb904e8a8 in ?? () from /lib/x86_64-linux-gnu/libc.so.6
#1  0x00002b2bb904f5a9 in ?? () from /lib/x86_64-linux-gnu/libc.so.6
#2  0x00002b2bb7a012da in _mesa_align_free () from /home/travis/build/mapbox/mapbox-gl-native/mason_packages/linux-x86_64/mesa/10.4.3/lib/libGL.so.1
#3  0x00002b2bb7b23c67 in _tnl_free_vertices () from /home/travis/build/mapbox/mapbox-gl-native/mason_packages/linux-x86_64/mesa/10.4.3/lib/libGL.so.1
#4  0x00002b2bb7b8940a in _swsetup_DestroyContext () from /home/travis/build/mapbox/mapbox-gl-native/mason_packages/linux-x86_64/mesa/10.4.3/lib/libGL.so.1
#5  0x00002b2bb78b4498 in XMesaDestroyContext () from /home/travis/build/mapbox/mapbox-gl-native/mason_packages/linux-x86_64/mesa/10.4.3/lib/libGL.so.1
#6  0x00002b2bb78ad71a in Fake_glXDestroyContext () from /home/travis/build/mapbox/mapbox-gl-native/mason_packages/linux-x86_64/mesa/10.4.3/lib/libGL.so.1
#7  0x00002b2bb78afbde in glXDestroyContext () from /home/travis/build/mapbox/mapbox-gl-native/mason_packages/linux-x86_64/mesa/10.4.3/lib/libGL.so.1
#8  0x000000000061c1c3 in mbgl::HeadlessView::~HeadlessView (this=0x7fff5670baa0) at ../../platform/default/headless_view.cpp:265
#9  0x000000000042c7b8 in HeadlessTest_render_Test::TestBody (this=<optimized out>) at ../../test/headless/headless.cpp:165
#10 0x0000000000650943 in void testing::internal::HandleSehExceptionsInMethodIfSupported<testing::Test, void>(testing::Test*, void (testing::Test::*)(), char const*) ()
#11 0x0000000000640421 in void testing::internal::HandleExceptionsInMethodIfSupported<testing::Test, void>(testing::Test*, void (testing::Test::*)(), char const*) ()
#12 0x00000000006262b7 in testing::Test::Run() ()
#13 0x000000000062710c in testing::TestInfo::Run() ()
#14 0x0000000000627935 in testing::TestCase::Run() ()
#15 0x000000000062f9bb in testing::internal::UnitTestImpl::RunAllTests() ()
#16 0x000000000064cea3 in bool testing::internal::HandleSehExceptionsInMethodIfSupported<testing::internal::UnitTestImpl, bool>(testing::internal::UnitTestImpl*, bool (testing::internal::UnitTestImpl::*)(), char const*) ()
#17 0x0000000000642b11 in bool testing::internal::HandleExceptionsInMethodIfSupported<testing::internal::UnitTestImpl, bool>(testing::internal::UnitTestImpl*, bool (testing::internal::UnitTestImpl::*)(), char const*) ()
#18 0x000000000062f667 in testing::UnitTest::Run() ()
#19 0x0000000000422b5c in RUN_ALL_TESTS () at ../../deps/gtest/gtest/gtest.h:20058
#20 main (argc=1, argv=0x21) at ../../test/fixtures/main.cpp:5
```

I will probably need to reproduce this locally and run under valgrind or something.

What this needs from **you** (@incanus, @kkaefer, @bleege, @mb12, and anyone else interested): heavy testing. This is likely to be a destabilizing change in the short term and I'd like to get a day or two of testing on the branch before merging it. I'll be out tomorrow (Friday) anyway, so I'll plan to merge on Monday with as much testing over the weekend as y'all can muster.

incanus:
> heavy testing

iOS integration testing just landed in `master`, so I'll start by merging that into here and seeing how it fares. But I'll certainly do lots of manual testing as well. /cc @1ec5 @friedbunny

gundersen:
@jfirebaugh this is awesome!

incanus:
Tests pass locally and the divergent commits from `master` were just iOS GUI-related + testing, so I merged and pushed above.

bleege:
Great stuff @jfirebaugh!  Can't wait to play with this in Sirius.

mikemorris:
Testing this out in [`separate-threads`](https://github.com/mapbox/node-mapbox-gl-native/tree/separate-threads) branch of node-mbgl, every render call throws an uncaught exception here:
```
frame #8: 0x00000001028bd2c5 mapbox-gl-native.node`mbgl::HeadlessView::resize(this=<unavailable>, width=<unavailable>, height=<unavailable>, pixelRatio=<u
navailable>) + 1125 at headless_view.cpp:200
    frame #9: 0x000000010280746d mapbox-gl-native.node`node_mbgl::NodeMap::startRender(this=0x0000000102a24dd0, options=unique_ptr<node_mbgl::NodeMap::RenderO
ptions, std::__1::default_delete<node_mbgl::NodeMap::RenderOptions> > at 0x00007fff5fbf4530) + 61 at node_map.cpp:202
    frame #10: 0x0000000102806c44 mapbox-gl-native.node`node_mbgl::NodeMap::Render(args=<unavailable>) + 524 at node_map.cpp:193
```

jfirebaugh:
@mikemorris Replace [`view.resize(...)`](https://github.com/mapbox/node-mapbox-gl-native/blob/5b84a62f650cf8e1fab78524c205520f755b03f3/src/node_map.cpp#L202) with `map.resize(...)`. This goes the other way now: map tells view to resize, rather than the other way around, so that the GL calls always happen in the map thread.

friedbunny:
Going to play with this now.

incanus:
Anecdotally, @friedbunny and I were discussing and it seems like the user dot is laggier in this branch. It does feel like overall perf is comparable, though. I will dig deeper and see if I can put my finger on why.

incanus:
The good news is I haven't crashed things!

friedbunny:
I crashed twice, but can't judge their relevance to these changes:

- https://github.com/mapbox/mapbox-gl-native/issues/1309#issuecomment-96108868 Assertion failure after backgrounding app during animation — Environment::abandonVAO()
- #1344 mbgl::HTTPRequestImpl::handleResponse crash at http_request_nsurl.mm:203

mikemorris:
> This goes the other way now: map tells view to resize, rather than the other way around, so that the GL calls always happen in the map thread.

:+1: Thanks @jfirebaugh, this sounds good - I'll test it out today.

mikemorris:
Alright, switched up the `view.resize` to `map.resize` and now I'm hitting an issue where a callback passed to `map.render` will never be called if the render emits an error (captured and emitted by `NodeLogObserver`). Reproducing this in [test/js/gzip.test.js](https://github.com/mapbox/node-mapbox-gl-native/blob/master/test/js/gzip.test.js#L109) in the [`separate-threads`](https://github.com/mapbox/node-mapbox-gl-native/tree/separate-threads) branch of node-mbgl.

jfirebaugh:
@mikemorris It looks like that test case sets up a scenario where the resources for the render will never be complete, because one of the tiles is invalid. Is that correct? What do you think the behavior of `Map::renderStill` should be in that case?

mikemorris:
@jfirebaugh Yea, that's correct. I was going to say I would expect `Map::renderStill` to return the callback early with an error, but then noticed the callback currently only returns a pointer to a `mbgl::StillImage` result.

jfirebaugh:
I'm still chasing a crash in the headless tests on Travis (can't reproduce locally), but it doesn't seem to manifest anywhere else and we need to get this into wider testing. I'm going to disable the failing tests temporarily, merge the branch, and then resume banging my head against the wall.

bleege:
> it doesn't seem to manifest anywhere else and we need to get this into wider testing. I'm going to disable the failing tests temporarily, merge the branch, and then resume banging my head against the wall.

:+1: We can roll a new Sirius later today to get this into wider circulation ASAP.

jfirebaugh:
Merged! Closing a bunch of issues that are known to be fixed or likely to be fixed by these changes. Please open new issues for any crashers observed post-merge.

incanus:
Let’s get this into a build today along with whatever else we can scare up, code or design. /cc @samanpwbb

PR #1285, opened 2015-04-16
fixes #1276: maintain snapshot in hierarchy & hide when appropriate

### LINKED ISSUE #1276: Extreme letterboxing on iOS after changing orientation while app is in background
If you change the device’s orientation while the app is in the background, the constraints get all wacky.

1. While the device is oriented in portrait mode, press the home button to leave the iOS demo app.
2. Double-press the home button to reveal the demo app in the app switcher.
3. Orient the device to landscape mode, then select the demo app to switch back to it.

letterboxed

incanus:
Moves from the `UIView` snapshot stuff back to `UIImageView`, which seems to be more reliable and works better with constraints, hitting #1276. We also now maintain the snapshot image view in the hierarchy after first use and just hide/show it.

### DIFF
--- a/platform/ios/MGLMapView.mm
+++ b/platform/ios/MGLMapView.mm
@@ -67,8 +67,7 @@ static NSURL *MGLURLForBundledStyleNamed(NSString *styleName)
 
 @property (nonatomic) EAGLContext *context;
 @property (nonatomic) GLKView *glView;
-/** A static view to simulate and obscure `glView` when the app is in the background. */
-@property (nonatomic) UIView *glSnapshotView;
+@property (nonatomic) UIImageView *glSnapshotView;
 @property (nonatomic) NSOperationQueue *regionChangeDelegateQueue;
 @property (nonatomic) UIImageView *compass;
 @property (nonatomic) UIImageView *logoBug;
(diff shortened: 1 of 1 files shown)


platform/ios/MGLMapView.mm
```diff
@@ -563,16 +562,7 @@ - (void)updateConstraints
...
-    {
-        NSUInteger snapshotIdx = [self.subviews indexOfObject:self.glSnapshotView];
-        NSUInteger glIdx = [self.subviews indexOfObject:self.glView];
-        NSAssert(snapshotIdx > glIdx,
-                 @"The snapshot view is the %luth subview, behind the GL view, which is the %luth. "
-                 @"The snapshot view should obscure the GL view to prevent crashing while the app is in the background.",
-                 (unsigned long)snapshotIdx, (unsigned long)glIdx);
-    }
-    else
+    if ( ! self.glSnapshotView || self.glSnapshotView.isHidden)
```
1ec5: Nit: Either `[self.glSnapshotView isHidden]` or `self.glSnapshotView.hidden`.

incanus:
```objc
@property(nonatomic,getter=isHidden) BOOL hidden; // default is NO. doesn't check superviews
```

:open_mouth:

1ec5: @incanus When you use dot syntax, you shouldn’t use the getter explicitly. (Try code completion in Xcode, or try using `.isHidden` in Swift.)

incanus: Oh, interesting! Cool, thanks for the tip. I'll make things consistent.

incanus: b6e2052

PR #1291, opened 2015-04-17
emplace into threadSet

mikemorris:
Map thread and Main thread should be separated in Mode::Static now

/cc @tmpsantos

### DIFF
--- a/src/mbgl/map/environment.cpp
+++ b/src/mbgl/map/environment.cpp
@@ -36,10 +36,7 @@ public:
     void registerThread(Environment* env, ThreadType type, const std::string& name) {
         std::lock_guard<std::mutex> lock(mtx);
 
-        // FIXME: We should never need to overwrite a thread here and we only allow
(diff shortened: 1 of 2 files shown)


jfirebaugh:
I rebased this on master and it seems to be triggering an assertion failure:

```
test: ../../src/mbgl/map/map_context.cpp:44: mbgl::MapContext::MapContext(uv_loop_t *, mbgl::View &, mbgl::FileSource &, mbgl::MapData &): Assertion `Environment::currentlyOn(ThreadType::Map)' failed.
```

mikemorris:
For context, this change is related to https://github.com/mapbox/node-mapbox-gl-native/issues/114, it's quite possible that additional work is needed to resolve this issue, https://github.com/mapbox/tilelive-gl/issues/25 and https://github.com/mapbox/api-gl/issues/16.

mikemorris:
For context, this change is related to https://github.com/mapbox/node-mapbox-gl-native/issues/114

PR #1295, opened 2015-04-19
Ensure correct spin direction

### LINKED ISSUE #1199: Heading-tracking mode takes the long way around
When using `MGLUserTrackingModeFollowWithHeading`, rotating the device several degrees around due south causes the map to go the wrong way to get there.

rotation-bug

1ec5:
Ported mapbox/mapbox-gl-js#821 as well as `util.wrap()` from mapbox-gl-js.

Fixes #1199.

/cc @mourner

### DIFF
--- a/include/mbgl/util/math.hpp
+++ b/include/mbgl/util/math.hpp
@@ -99,6 +99,12 @@ T clamp(T value, T min, T max) {
 }
 
 template <typename T>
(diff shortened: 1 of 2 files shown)


friedbunny:
Awesome, thanks for grabbing this one. :dizzy:

PR #1296, opened 2015-04-19
Cancel animations on touch down

### LINKED ISSUE #1258: Panning animation does not cancel upon tap. (iOS)
Steps to reproduce

1. Flick to pan the map fast
2. Tap to stop the panning

Panning animation continues instead of stopping upon tap.

1ec5:
Cancel any map animations as soon as a finger is set upon the screen.

Fixes #1258.

/cc @incanus


1ec5:
Note that after touch down, there’s a second-or-so window during which you can’t begin a pan gesture. `UIScrollView` doesn’t have this limitation, but `MKMapView` and Google Maps both do, so it’s a bug for another day.

PR #1297, opened 2015-04-19
weakSelf needs strongSelf

1ec5:
The `weakSelf` dance requires the use of a `strongSelf` in non-trivial¹ blocks. In these cases, `weakSelf` could disappear right out from under the block. A crash could potentially occur due to setting a `nil` value in an `NSMutableDictionary`.

Probably fixes any crash involving `__44-[MGLMapboxEvents pushEvent:withAttributes:]_block_invoke`, including:

https://rink.hockeyapp.net/manage/apps/163490/app_versions/11/crash_reasons/32765840

¹ Where trivial means the block does nothing but call some method on the (potentially `nil`) `weakSelf`.

/cc @incanus @bleege


1ec5:
💪

Issue #738, opened 2014-12-31
Error inflating class com mapbox.mapboxgl.lib.MapView  Android


Aminebr5:
I'm trying to compile Mapbox-gl-native android-mason on Android Studio, i've this error popping out every time i run it, is there any solution to this,

java.lang.RuntimeException: Unable to start activity ComponentInfo{com.mapbox.mapboxgl.app/com.mapbox.mapboxgl.app.MainActivity}: android.view.InflateException: Binary XML file line #2: Error inflating class com.mapbox.mapboxgl.lib.MapView
            at android.app.ActivityThread.performLaunchActivity(ActivityThread.java:2365)
            at android.app.ActivityThread.handleLaunchActivity(ActivityThread.java:2417)
            at android.app.ActivityThread.access$600(ActivityThread.java:169)
            at android.app.ActivityThread$H.handleMessage(ActivityThread.java:1377)
            at android.os.Handler.dispatchMessage(Handler.java:107)
            at android.os.Looper.loop(Looper.java:194)
            at android.app.ActivityThread.main(ActivityThread.java:5449)
            at java.lang.reflect.Method.invokeNative(Native Method)
            at java.lang.reflect.Method.invoke(Method.java:525)
            at com.android.internal.os.ZygoteInit$MethodAndArgsCaller.run(ZygoteInit.java:833)
            at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:600)
            at dalvik.system.NativeStart.main(Native Method)
     Caused by: android.view.InflateException: Binary XML file line #2: Error inflating class com.mapbox.mapboxgl.lib.MapView
            at android.view.LayoutInflater.createView(LayoutInflater.java:613)
            at android.view.LayoutInflater.createViewFromTag(LayoutInflater.java:687)
            at android.view.LayoutInflater.inflate(LayoutInflater.java:466)
            at android.view.LayoutInflater.inflate(LayoutInflater.java:396)
            at com.mapbox.mapboxgl.app.MapFragment.onCreateView(MapFragment.java:43)
            at android.support.v4.app.Fragment.performCreateView(Fragment.java:1786)
            at android.support.v4.app.FragmentManagerImpl.moveToState(FragmentManager.java:920)
            at android.support.v4.app.FragmentManagerImpl.moveToState(FragmentManager.java:1104)
            at android.support.v4.app.FragmentManagerImpl.addFragment(FragmentManager.java:1206)
[...]

PR #1245, opened 2015-04-20
Android TestApp crashes on startup

kkaefer:
I have tried building the Android project from the master branch. My environment is set up with homebrew and cask following the instructions in the readme. <code>make android</code> appears to complete successfully and the project under <code>android/java</code> imports into Android Studio and builds without error.

When I try deploying the TestApp to my Nexus 5 running Android 5.1 the application flashes on screen and crashes. Here's the output from logcat: https://gist.github.com/tnightingale/49b605b793bb1ced68be

This looks similar to the initial errors mentioned in #738.

### DIFF
--- a/android/cpp/jni.cpp
+++ b/android/cpp/jni.cpp
@@ -927,9 +927,7 @@ extern "C" JNIEXPORT jint JNICALL JNI_OnLoad(JavaVM *vm, void *reserved) {
         return JNI_ERR;
     }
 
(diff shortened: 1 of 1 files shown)


thanawatSuj:
Thank you for commit it OK, But When open map dont appeared. it white area.

It was from what?

Thank you

thanawatSuj:
screenshot_2015-04-21-14-10-24
screenshot_2015-04-21-13-51-16

Result When i compile and run,
I think about Touch gesture (Scale,rotate,zoom) is not correct.

Thank You.

PR #1289, opened 2015-04-21
Integrate node build into CI

kkaefer:
We should run the node module build and tests against the current commit to make sure we don't accidentally break it.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -9,6 +9,7 @@
 /mason_packages
 /config/*.gypi
 /build
+/node
 /macosx/build
 /linux/build
 /ios/build
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,6 +10,8 @@ matrix:
       env: FLAVOR=linux CXX=clang++ BUILDTYPE=Debug
     - os: linux
       env: FLAVOR=linux CXX=clang++ BUILDTYPE=Release
+    - os: linux
+      env: FLAVOR=node CXX=clang++ BUILDTYPE=Release
     - os: osx
       env: FLAVOR=osx BUILDTYPE=Debug
     - os: osx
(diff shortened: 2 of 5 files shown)


jfirebaugh:
I think we should merge node-mapbox-gl-native into mapbox-gl-native, the same way we merged mapbox-gl-cocoa. The node bindings have essentially the same role as the Android and iOS SDKs, and the same arguments for keeping them in the same repository apply. I bet @mikemorris would welcome such a change profusely.

mikemorris:
@jfirebaugh I'd actually prefer to keep node-mbgl as a separate project because [it integrates with mbgl on a much different level than Android or iOS](https://github.com/mapbox/mapbox-gl-cocoa/issues/28#issuecomment-72967012). The Node.js "bindings" are really just a library consuming the mbgl public API, and don't need the tight integration required by Android and iOS.

I had a long back-and-forth with @incanus about this in https://github.com/mapbox/mapbox-gl-cocoa/issues/28

jfirebaugh:
That discussion was before we realized it would be _so much easier_ to have mapbox-gl-cocoa not be a separate repository. The same is true of node-mapbox-gl-native. Right now you're doing a complex dance with the build scripts, and still struggling to find the right incantations, in a way that's made accidentally complex because of the separate repositories.

> The Node.js "bindings" are really just a library consuming the mbgl public API

Replace 'Node.js "bindings"' with "iOS bindings" or "Android bindings" and you have an accurate description of those two things. Node is really just another runtime/language binding.

incanus:
> The Node.js "bindings" are really just a library consuming the mbgl public API

Yeah, so are the mobile bindings. We just use `map.hpp` etc.

mikemorris:
-> https://github.com/mapbox/mapbox-gl-native/pull/2179

PR #1324, opened 2015-04-22
fixes #1308: normalize tile memory cache keys & allow dupe inserts

### LINKED ISSUE #1308: Assertion failed: (tiles.find(key) == tiles.end())
I get this assertion failure pretty frequently when panning and zooming in the desktop app:

```
Assertion failed: (tiles.find(key) == tiles.end()), function add, file /Users/john/Development/mapbox-gl-native/src/mbgl/map/tile_cache.cpp, line 23.
```

incanus:
See write-up in https://github.com/mapbox/mapbox-gl-native/issues/1308#issuecomment-95047478. 

- We need to normalize tile IDs for cache operations so that low zooms that show e.g. `1/-1/0` and `1/1/0` at the same time benefit from a shared cache object. 

- As a result, we need to allow for duplicate cache inserts / touching recently-used cache entries since both of those source tiles could go obsolete within a time period that keeps their shared object in the cache. 

/cc @jfirebaugh @ansis @1ec5

### DIFF
--- a/src/mbgl/map/tile_cache.cpp
+++ b/src/mbgl/map/tile_cache.cpp
@@ -20,11 +20,16 @@ void TileCache::setSize(size_t size_) {
 
 void TileCache::add(uint64_t key, std::shared_ptr<TileData> data) {
 
-    assert(tiles.find(key) == tiles.end());
+    // insert new or query existing data
+    if (tiles.emplace(key, data).second) {
+        // remove existing data key
+        orderedKeys.remove(key);
+    }
 
-    tiles.emplace(key, data);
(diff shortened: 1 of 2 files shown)


src/mbgl/map/tile_cache.cpp
```diff
@@ -20,11 +20,16 @@ void TileCache::setSize(size_t size_) {
 
 void TileCache::add(uint64_t key, std::shared_ptr<TileData> data) {
 
-    assert(tiles.find(key) == tiles.end());
```
kkaefer: We should keep this assert since normalized keys will never be duplicate?

Issue #1265, opened 2015-04-14
tweak logo placement on iOS


incanus:
This looks sloppy. 

![img_3268 copy](https://cloud.githubusercontent.com/assets/17722/7140952/4ed63752-e283-11e4-9f26-b39b8193aae2.png)

1ec5:
It’s aligned to the baseline of “Ma_box”, not the bottom of the “p”.

incanus:
Right, but it looks weird IMHO and the side margins are different, anyway.

incanus:
The placement to the _(i)_ by the way is Apple's default.

1ec5:
Ref #1079

1ec5:
I also considered aligning the baseline of “Ma_box” (not counting the outline) to the “i” in :information_source:.

bsudekum:
Have you considered remove the :information_source: and just going with the `Mapbox` logo? It's a lot going on.

incanus:
The :information_source: is how we reveal attribution info, and eventually contribution info. Currently on GL it just opens Safari, but on the iOS SDK, we've had a lot of luck with popping both the OSM and the open source code info in there so that folks don't forget it (it's opt-out, and [worded in a way](https://www.mapbox.com/mapbox-ios-sdk/api/Classes/RMMapView.html#task_Attributing%20Map%20Data) that makes people question that they should be hiding it). 

![anigif-1429029353](https://cloud.githubusercontent.com/assets/17722/7141937/ac1d8892-e289-11e4-925c-e98060704e5b.gif)

1ec5:
And if we had the logo subsume the function of the :information_source: then we’d probably launch a thousand [talk@](https://lists.openstreetmap.org/listinfo/talk/) threads.

incanus:
Things look ok now.

PR #1328, opened 2015-04-22
Refactored constraints within MGLMapView

### LINKED ISSUE #1300: User location dot appears outside MGLMapView
The current user location dot is outside of the mapbox gl class

### LINKED ISSUE #1327: iOS: 🐶s hug screen layout guides even when view is smaller than screen
The [:dog:s](https://en.wikipedia.org/wiki/Digital_on-screen_graphic) – logo, :information_source:, and compass – are all positioned relative to the screen layout guides, even when the `MGLMapView` only occupies a portion of the view controller’s root view. This scenario may be more common on the iPad than on the iPhone.

inset

1ec5:
This PR cleans up `MGLMapView`’s internal constraints:

 - Replaces Visual Format Language usage and various ternaries with raw `NSLayoutConstraint`-building (clearer logic, shorter lines)
 - Adds additional constraint relations so that the various subviews respect the `MGLView`’s frame when it’s smaller than the view controller’s root view (fixes #1327)
 - Clips `MGLMapView` to its bounds (fixes #1300)

That last change has the side-effect that callouts no longer draw outside the map’s bounds. That was either a bug or a feature. If we want to keep that capability, we’d need to clip subviews individually (so that the user dot could half-appear rather than hang halfway off the map) and hide callouts’ anchor triangles when they’d point somewhere outside the map.

/cc @incanus


1ec5:
9b8b072d28c61a6cca2e4aea3631f8b5660bedad ignores whitespace in the logo during layout. The only visual difference is that the logo’s left margin and the :information_source:’s right margin are now equal:

baselined

@incanus, I very much wanted to fix #1265 as part of this PR, but I think your proposed solution would be worse than what we have now, due to an optical illusion. I tried vertically aligning the centers of the logo and :information_source:, which happens to be the same as aligning the p’s descender with the bottom of the :information_source:. But the logo now appears to be cocked upward, as if the tops of both views were aligned (they aren’t):

centered

incanus:
You knew this was coming, but ... care to merge/test against `master` now that tests have landed? Also we might want to capture the old failure case here in tests to prevent a regression. I'll also take a look at this.

1ec5:
I’ve added tests for #1327 but not #1300. (For the latter, we’d have to mock the user location, ensure that the user dot wanders out of the shrunken `MGLMapView`, snapshot the root view, and verify that the snapshot image doesn’t include any blueish pixels in the neighborhood of the user dot.)

Issue #1225, opened 2015-04-07
Create Mapbox GL Start Infrastructure


bleege:
The current Mapbox GL startup system is all tied to `MGLMapView`.  This means that in order for an app developer to use any Mapbox service a MapView has to be built.  It also ties Metrics to MapView as well.

**Future State Goals**

1. Start Mapbox GL in AppDelegate's `didFinishLaunchingWithOptions`
  *  ```[MapboxGL startWithAccessToken:@"YOURACCESSTOKENGOESHERE"];```
1. `MapboxGL startWithAccessToken`'s responsibilities
  * Serve as singleton holder of Access Token for all SDK uses
  * Start Metrics services

@mapbox/mobile

1ec5:
As you work on this, please keep in mind the XIB/storyboard scenario we enabled in #1070: it should still be possible to set an access token on a view-by-view basis, without overriding anything in the app delegate. Codeless setup is a compelling story (MapKit parity, makes for a good beginner’s guide, shorter [demo GIFs](https://github.com/mapbox/mapbox-gl-native/pull/1184#issue-65976641)), even if we expect most developers to go the programmatic route eventually.

jfirebaugh:
I'm not sure this is a good idea. A required singleton init method typically makes libraries harder to use.

> in order for an app developer to use any Mapbox service a MapView has to be built

What Mapbox services does Mapbox GL provide besides MapViews?

bleege:
> A required singleton init method typically makes libraries harder to use.

Generally I agree, but in this particular case it's a very common iOS design pattern is to to initialize 3rd party code at app startup inside the app's `AppDelegate`.  Specifically in the following callback method:

```Objective-C
- (BOOL)application:(UIApplication *)application didFinishLaunchingWithOptions:(NSDictionary *)launchOptions
```

> What Mapbox services does Mapbox GL provide besides MapViews?

Currently there are no other services besides MapViews, but if others were to come along in the future (ex: Directions API wrapper) the existing infrastructure would require developers to first create a MapView in order to get access tokens to use.  This way would provide a service agnostic one line of code common setup.

1ec5:
> A required singleton init method typically makes libraries harder to use.

This is why any singleton setup step needs to be optional: individual top-level view classes like `MGLMapView` should continue to accept access tokens, so that you can instantiate a working map from within a XIB or stroryboard, or from within a custom React Native component, as @bsudekum is trying to do.

[...]

PR #1329, opened 2015-04-22
MapboxGL Startup Infrastructure

bleege:
For #1225


ios/app/MBXAppDelegate.m
```diff
@@ -1,11 +1,27 @@
...
 @implementation MBXAppDelegate
 
 - (BOOL)application:(UIApplication *)application didFinishLaunchingWithOptions:(NSDictionary *)launchOptions
 {
+    // Set Access Token
+    NSString *accessToken = [[NSProcessInfo processInfo] environment][@"MAPBOX_ACCESS_TOKEN"];
+    if (accessToken) {
+        // Store to preferences so that we can launch the app later on without having to specify
+        // token.
+        [[NSUserDefaults standardUserDefaults] setObject:accessToken forKey:@"access_token"];
```
1ec5: Nit: A common strategy for user defaults keys is to make them match the names of the constants that define them. So you’d define a constant called `MGLMapboxAccessTokenDefaultsKey` with the value `MGLMapboxAccessToken` in MGLTypes.h and use that constant here. Obviously it doesn’t matter as long as you’re consistent, but it’s one less thing to worry about naming.

ios/app/MBXAppDelegate.m
```diff
@@ -1,11 +1,27 @@
...
+    } else {
+        // Try to retrieve from preferences, maybe we've stored them there previously and can reuse
+        // the token.
+        accessToken = [[NSUserDefaults standardUserDefaults] objectForKey:@"access_token"];
+    }
+    if ( ! accessToken) NSLog(@"No access token set. Mapbox vector tiles won't work.");
+
+    [MapboxGL sharedInstanceWithAccessToken:accessToken];
```
1ec5: It looks like you’re creating a `MapboxGL` and throwing it away. Someone might come along later and delete the line, thinking it does nothing, when in fact you’re setting the access token globally by side effect. Can we call this class method `-setAccessToken:` or `-registerAccessToken:` instead?

platform/ios/MapboxGL.m
```diff
@@ -0,0 +1,48 @@
...
+            }
+            else {
+                setupBlock();
+            }
+        }
+    });
+    return _sharedManager;
+}
+
++ (NSString *) getAccessToken {
```
1ec5: Just `-accessToken`. “Get” would imply some kind of non-trivial activity like a network request or disk read.

include/mbgl/ios/MapboxGL.h
```diff
@@ -2,3 +2,10 @@
 #import "MGLMapView.h"
 #import "MGLTypes.h"
 #import "MGLUserLocation.h"
+
+@interface MapboxGL : NSObject
```
1ec5: This header was meant to be an [umbrella header](https://developer.apple.com/library/mac/documentation/MacOSX/Conceptual/BPFrameworks/Tasks/IncludingFrameworks.html), which does nothing but import all the other public-facing headers. (A developer should only ever have to import the one umbrella header unless they’re doing something out of the ordinary.)

I’d move this class to another header and rename it `MGLAccountManager` or somesuch, sticking to the framework’s established prefix.

platform/ios/MapboxGL.m
1ec5: On second thought, if the developer can just get the access token from a class method (`+accessToken`), why make them create a shared instance in the first place?

Let’s strive for parallelism in one of two ways:

 - A class method `+sharedManager` that returns an [`MGLAccountManager`](https://github.com/mapbox/mapbox-gl-native/pull/1329/files#r29024059) with nothing set, and two instance methods: `-setAccessToken:` and `-accessToken`. This would be similar to how `NSProcessInfo` works.
 - Two standalone C functions: `MGLRegisterAccessToken()` and `MGLGetRegisteredAccessToken()`. (On Apple platforms, Objective-C getters don’t say “get”, but C getters do, [to distinguish from “copy” functions](https://developer.apple.com/library/ios/documentation/CoreFoundation/Conceptual/CFMemoryMgmt/Concepts/Ownership.html#//apple_ref/doc/uid/20001148-SW3).) Instead of keeping a singleton object around, keep just a static access token string. This approach is simpler but less extensible.

bleege:
@1ec5's feedback has been incorporated.  The next step is to split MapboxGL.h back into a regular Umbrella Header file.

platform/ios/MapboxGL.m
bleege: I ended up going with just renaming to `+sharedAccessToken` to avoid any `get` convention.  The instance methods route wouldn't work due to [MGLMapView needing to retrieve it](https://github.com/mapbox/mapbox-gl-native/pull/1329/files#diff-f315cd6ead3270a0818a2dcfb84bd99bR126).  By keeping it in a `+sharedManager` this class stays more extensible.

1ec5: `MGLMapView` could’ve always getting the access token through `[[MGLAccountManager sharedInstance] accessToken]`.

platform/ios/MGLAccountManager.m
```diff
@@ -0,0 +1,50 @@
...
+
+
+@implementation MGLAccountManager
+
+static MGLAccountManager *_sharedManager;
+
+// Can be called from any thread. Called implicitly from any
+// public class convenience methods.
+//
++ (instancetype) sharedInstanceWithAccessToken:(NSString *)token {
```
1ec5: This method’s name is misleading. It implies that there can be more than one shared instance, that we’re mapping access tokens to shared instances somehow. What happens when this class begins to manage more than just an access token? `+sharedInstanceWithUserName:accessToken:expiryDate:cachingPragma:`?

I still think the two options in https://github.com/mapbox/mapbox-gl-native/pull/1329#discussion_r29024497 are the best ways forward.

bleege: If I understand correctly using the instance methods would then have the initial setup in the AppDelegate be something like:

```
[[MGLAcountManager sharedManager] setAccessToken:@"FOO"]

[MGLAcountManager sharedInstanceWithAccessToken:@"FOO"]

```

How about renaming it to `managerWithInititalAccessTokenValue` or a derivative?

FWIW, I'm not terribly fond of the C style option so I'd like to get the first option working.

incanus: > `[[MGLAcountManager sharedManager] setAccessToken:@"FOO"]`

The way we do this in MBXMapKit is more brief: 

> `[MGLAccountManager setAccessToken:]`

Which is basically just a shortcut to the above, but doesn't require the user to worry about shared instances.

incanus: Also, `+setAccessToken:` could return an `instancetype` for easy access.

ios/app/MBXAppDelegate.m
incanus: We'd be better off putting things in a constant to avoid future potential problems with the `@"access_token"` [magic string](http://www.cimgf.com/2013/01/29/down-with-magic-strings/), too.

platform/ios/MGLAccountManager.m
bleege: I ended up implementing `+setAccessToken` that returned `void` instead of `instancetype` as the latter was causing crashes.

bleege:
Ended up pulling the initial `MGLMapboxEvents` from the `MGLAccountManager` init block as the only way to create a `MGLAccountManager` now is via `+setAccessToken`.

platform/ios/MGLMapView.mm
```diff
@@ -28,6 +28,7 @@
 #import "SMCalloutView.h"
 
 #import "MGLMapboxEvents.h"
+#import "MapboxGL.h"
```
1ec5: This import statement likle

1ec5: This import statement likely makes some of the other ones redundant.

platform/ios/MGLMapView.mm
```diff
@@ -109,6 +110,7 @@ - (instancetype)initWithFrame:(CGRect)frame
     if (self && [self commonInit])
     {
         self.styleURL = nil;
+        self.accessToken = [MGLAccountManager accessToken];
```
1ec5: What about the other direction? `-[MGLMapView setAccessToken:]` should update the singleton.

bleege: Just ticked it -> https://github.com/mapbox/mapbox-gl-native/issues/1430

Issue #434, opened 2014-08-19
Coalesce VAO and Buffer creation and deletion


kkaefer:
Currently, we have a lot of calls to subsequent `glDeleteVertexArrays` or `glDeleteBuffers` that only specify one argument. We should collect all buffers we need to delete and release them in one go instead of making individual calls for every buffer.

kkaefer:
VAO deletion is now coalesced as of https://github.com/mapbox/mapbox-gl-native/commit/61fa43619334ac68c66d45d38aaf681f2a21fa96

To coalesce creation, we'll have to uncouple data buffer creation and OpenGL object creation.

kkaefer:
This is implemented.

brunoabinader:
Buffers, textures and VAOs deletion are coalesced in 9c26f063.

> To coalesce creation, we'll have to uncouple data buffer creation and OpenGL object creation.

This is partially done - OpenGL objects are created in a separate timeline, but we still call the GL creation functions individually for buffers, textures and VAOs (with exception to our texture pool).

@kkaefer do you think it is feasible to group the creation of these in a single call? I'm afraid this might create too much coupling between the GLHolder objects and its containers.

PR #1325, opened 2015-04-23
Simplify render path

kkaefer:
We're currently have a pretty complicated way of rendering a tile, going through many layers of objects and some double dispatching along the way. This makes the code hard to follow and we need to traverse the hierarchy several times, for rendering opaque and translucent objects. We also want to add an additional step before that, giving renderable objects (=buckets) an opportunity to upload data to the GPU before we kick off rendering a frame.

### DIFF
--- a/gyp/platform-android.gypi
+++ b/gyp/platform-android.gypi
@@ -12,6 +12,7 @@
       'sources': [
         '../platform/android/log_android.cpp',
         '../platform/android/asset_root.cpp',
+        '../platform/default/thread.cpp',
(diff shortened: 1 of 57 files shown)


kkaefer:
@mapbox/gl Please review.

ansis:
Is there a measurable performance improvement from wrapping all calls that modify gl state?
What performance impact did the switch to uploading at the beginning of a frame have?
Any plans to port this to -js?

kkaefer:
This fixes #434

PR #1346, opened 2015-04-25
fix SSL pinning for apps including the framework

incanus:
Made it clearer when we are referring to the **app** bundle and when we need to get the **framework** bundle for resources such as pinned certificates.

Uses of `+[NSBundle mainBundle]` refer to the host app's bundle, which in the case of the GL dev app could include framework-level resources since GYP makes it tricky to build non-app target resource bundles and that's how we roll the app. That's why the current `master` worked in testing. 

Uses `+[NSBundle bundleForClass:]` for cases when we specifically mean the framework bundle, such as we do in [`+[MGLMapView resourceBundlePath]`](https://github.com/mapbox/mapbox-gl-native/blob/8aa2e2fef76eb4dfdb6e5b28a815babd27647322/platform/ios/MGLMapView.mm#L2338). We use `MGLMapboxEvents ` in this fix since the `MGLMapboxEvents` class should be unaware of `MGLMapView` specifically and should be entirely self-contained. 

/cc @bleege @camilleanne @1ec5 @kkaefer

### DIFF
--- a/platform/ios/MGLMapboxEvents.m
+++ b/platform/ios/MGLMapboxEvents.m
@@ -111,13 +111,13 @@ NSString *const MGLEventGestureRotateStart = @"Rotation";
     if (self) {
         
         // Put Settings bundle into memory
-        NSString *settingsBundle = [[NSBundle mainBundle] pathForResource:@"Settings" ofType:@"bundle"];
-        if(!settingsBundle) {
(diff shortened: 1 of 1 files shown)


1ec5:
:+1: Good catch!

1ec5:
For more background, see #1011.

PR #1339, opened 2015-04-27
Move Travis CI to Docker machines

kkaefer:


mikemorris:
Can we break out the setup from the build step in `scripts/linux/run.sh`? I need to reconfigure the Travis build in https://github.com/mapbox/node-mapbox-gl-native/pull/118 now, and it would be nice to source a setup file instead of duplicating all the code.

mikemorris:
Ugh and the relative source path for `travis_helper.sh` isn't working either.

```
./vendor/mbgl/scripts/linux/install.sh: line 6: ./scripts/travis_helper.sh: No such file or directory
```

PR #1357, opened 2015-04-27
Automatically find view controller for layout guides

1ec5:
We inherited an annoying, easy-to-overlook setup step from the iOS SDK: [`MGLMapView`’s `viewControllerForLayoutGuides` outlet must be hooked up to the managing view controller](https://github.com/mapbox/mapbox-gl-native/wiki/Installing-Mapbox-GL-for-iOS/_compare/a551a579d1e03dd83340868416603e882e890ffe...6e8c9508f285741401ef62501d352308223bc99c) in order for the ornaments (logo, :information_source:, compass) to be laid out properly.

Based on the [`-[UIResponder nextResponder]` documentation](https://developer.apple.com/library/ios/documentation/UIKit/Reference/UIResponder_Class/index.html#//apple_ref/occ/instm/UIResponder/nextResponder) and the fixes for ornament layout in #1328, this PR causes `MGLMapView` to compute the managing view controller on its own by traversing the responder chain. (If there is a legitimate need to lay out the ornaments based on some view controller not in the responder chain, that would be a good opportunity to subclass `MGLMapView`; in that case, we’d need to re-expose `-viewControllerForLayoutGuides` publicly, but there would still be no need for a stored property.)

Once this PR is merged and contained in a tag, I’ll remove the step from the installation guide.

/cc @incanus @friedbunny

### DIFF
--- a/include/mbgl/ios/MGLMapView.h
+++ b/include/mbgl/ios/MGLMapView.h
@@ -46,11 +46,6 @@ IB_DESIGNABLE
 
 /** @name Managing Constraints */
 
-/** A view controller whose top and bottom layout guides to use for proper setup of constraints in the map view internals.
-*
-*   Certain components of the map view, such as the heading compass and the data attribution button, need to be aware of the view controller layout in order to avoid positioning content under a top navigation bar or a bottom toolbar. */
-@property (nonatomic, weak) IBOutlet UIViewController *viewControllerForLayoutGuides;
-
 #pragma mark - Accessing Map Properties
 
 /** @name Accessing Map Properties */
--- a/ios/app/MBXViewController.mm
+++ b/ios/app/MBXViewController.mm
@@ -67,7 +67,6 @@ mbgl::Settings_NSUserDefaults *settings = nullptr;
 
     self.mapView = [[MGLMapView alloc] initWithFrame:self.view.bounds accessToken:accessToken];
     self.mapView.autoresizingMask = UIViewAutoresizingFlexibleWidth | UIViewAutoresizingFlexibleHeight;
-    self.mapView.viewControllerForLayoutGuides = self;
     self.mapView.showsUserLocation = YES;
     self.mapView.delegate = self;
     [self.view addSubview:self.mapView];
--- a/platform/ios/MGLMapView.mm
+++ b/platform/ios/MGLMapView.mm
@@ -323,8 +323,6 @@ mbgl::DefaultFileSource *mbglFileSource = nullptr;
     [container addGestureRecognizer:[[UITapGestureRecognizer alloc] initWithTarget:self action:@selector(handleCompassTapGesture:)]];
     [self addSubview:container];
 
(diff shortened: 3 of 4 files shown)


incanus:
I'm torn here, because we autodetect in the iOS SDK but it felt like voodoo. I thought it better to leave this to the developer in GL. Unfortunately there is not a good president for doing this one way or the other that I can find — MapKit has access to private API which makes this solid from whatever autodetection they are doing internally. Can we search for another third-party similarity and compare before changing current behavior?

incanus:
See also the behavior in `RMMapView` in the SDK for `viewControllerPresentingAttribution`, which isn't necessarily the map view's parent view controller.

1ec5:
Traversing the responder chain is a [pretty common approach](http://stackoverflow.com/questions/1372977/given-a-view-how-do-i-get-its-viewcontroller), and it’s based entirely on officially documented behavior. (On OS X, the responder chain is how you get UI niceties as you input text in an `NSTextField` for instance.)

Either way, we’re [violating MVC](http://stackoverflow.com/questions/1340434/get-to-uiviewcontroller-from-uiview-on-iphone) by having the view know about its controller. The (pedantically) correct way to do it would be to support a delegate method `-viewControllerForAligningOrnamentsInMapView:` and requiring the developer to hook up a delegate that returns the view controller. I’m opposed to that approach because implementing a delegate should not be required just to set up the most basic map view.

As it turns out, `-[RMMapView viewControllerPresentingAttribution]` is [initially set to `-viewController`](https://github.com/mapbox/mapbox-ios-sdk/blob/84b1c0d5f117674dcc546d15260f41fcffe09450/MapView/Map/RMMapView.m#L652), which is [implemented by traversing the responder chain](https://github.com/mapbox/mapbox-ios-sdk/blob/84b1c0d5f117674dcc546d15260f41fcffe09450/MapView/Map/RMMapView.m#L512).

1ec5:
For those following along at home, the private API @incanus is referring to is `+[UIViewController viewControllerForView:]`. `-[MKMapView _updateInsets]` uses it via `-[MKMapView _findLayoutGuideVC]` to update edge insets that it then uses to manually position the `MKAttributionLabel`. But given the given the fact that we’re only using knowledge of the view controller to position views internal to `MGLMapView`, I think traversing the responder chain is perfectly non-hacky.

1ec5:
As for prior art: [`MASTOrmmaSharedDelegate`](https://code.google.com/p/mocean-sdk-ios/source/browse/Sources/Classes/Ad/Views/SubViews/AdWebView/ORMMA/MASTOrmmaSharedDelegate.m?r=141a356e5ede1d755f434cd1c09a68c4683ad339#66) and I’m sure plenty of others.

incanus:
Ok, cool. :shipit: :smile:

friedbunny:
This seems like a very sensible abstraction for end-developers, I don't mind if there's a small voodoo aspect. :goat:

PR #1362, opened 2015-04-28
Wait for initial region change notification from reset before testing (Fixes #1360)

### LINKED ISSUE #1360: testDelegateRegionDidChange is too fragile
I'm seeing relatively frequent test failures, but they're not reproducible. It looks like they heavily depend on timing.

See https://travis-ci.org/mapbox/mapbox-gl-native/jobs/60350680#L1051

@incanus @1ec5

friedbunny:
The `beforeEach` reset of `tester.mapView.centerCoordinate` will emit a `regionDidChangeAnimated` notification that `testDelegateRegionDidChange` sees but should ignore before proceeding.

This PR inserts a `waitForNotificationName:@"regionDidChangeAnimated"` that should hear from `beforeEach` before continuing on to the actual tests.

This test was passing locally, but failing on Travis (presumably because of performance issues on their end).

Fixes #1360.

/cc @incanus @kkaefer

### DIFF
--- a/test/ios/MapViewTests.m
+++ b/test/ios/MapViewTests.m
@@ -395,16 +395,18 @@
                                                       }
                                                   }];
 
-    [tester waitForTimeInterval:1];
+    // wait for initial reset before proceeding to actual tests
+    NSNotification *notification = [system waitForNotificationName:@"regionDidChangeAnimated"
+                                                            object:tester.mapView];
 
     unanimatedCount = 0;
     animatedCount = 0;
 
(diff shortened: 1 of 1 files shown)


friedbunny:
This could still fail if the timing is thrown off and we start listening for notifications after the `beforeEach` one comes in. Not sure if this is just swapping one timing issue for another...

friedbunny:
Appears to be unnecessary, but could be useful in the future if Travis starts balking on this again.

PR #1370, opened 2015-04-29
Antialias compass.png N text

friedbunny:
Brings the `@1x` compass :dog: asset in line with the other resolutions.


friedbunny:
Hold up, going to adjust the vertical spacing.

PR #1374, opened 2015-04-29
Types [v8] 

lbud:
@tmcw — this is a pretty MVP implementation of types. The question is: how deeply do we eventually want to enforce it? If more deeply we can add more strict style parsing rules, but it seems like for now this should serve our current needs…
(The other question is: was it a huge mistake to branch off `master` rather than off `v8-two`?)


lbud:
firetruck this is a big diff for 
image

lbud:
Closing in favor of #1390

PR #1380, opened 2015-04-30
Add benchmarking application for iOS

kkaefer:
Add a benchmarking application for iOS that renders certain predefined scenes with predefined tiles. It first sets the location, waits until all assets are loaded, then does a couple of frames of warmup before it does the measurement. Results are reproducible within ~1 FPS.


include/mbgl/ios/MGLMapView.h
```diff
@@ -32,6 +32,7 @@ IB_DESIGNABLE
  *   @param styleURL The map style URL to use. Can be either an HTTP/HTTPS URL or a Mapbox map ID style URL (`mapbox://<user.style>`).
  *   @return An initialized map view, or `nil` if the map view was unable to be initialized. */
 - (instancetype)initWithFrame:(CGRect)frame accessToken:(NSString *)accessToken styleURL:(NSURL *)styleURL;
+- (instancetype)initWithFrame:(CGRect)frame accessToken:(NSString *)accessToken styleURL:(NSURL *)styleURL cache:(BOOL)enabled;
```
incanus: Need to add docs.

include/mbgl/ios/MGLMapView.h
```diff
@@ -280,6 +281,12 @@ IB_DESIGNABLE
 /** Resets the map to the minimum zoom level, a center coordinate of (0, 0), and a northern heading. */
 - (void)resetPosition;
 
+/** Triggers another render pass even when it is not necessary. */
+- (void)invalidate;
+
+/** Returns whether the map view is currently loading or processing any assets required to render the map */
+- (BOOL)isFullyLoaded;
+
```
incanus: These shouldn't be public.

include/mbgl/ios/MGLMapView.h
```diff
@@ -364,6 +371,9 @@ IB_DESIGNABLE
 // TODO
 - (void)mapViewDidFinishRenderingMap:(MGLMapView *)mapView fullyRendered:(BOOL)fullyRendered;
 
+// TODO
+- (void)mapViewDidFinishLoadingAllAssets:(MGLMapView *)mapView;
+
```
incanus: This is weird to have alongside the MapKit ones we will eventually have.

kkaefer: Why is it weird? We already have all of the other notifications?

incanus: The other ones are mirrored on MapKit, and this one is close to these which exist (or will): 

```objc
- (void)mapViewWillStartLoadingMap:(MGLMapView *)mapView;
- (void)mapViewDidFinishLoadingMap:(MGLMapView *)mapView;
- (void)mapViewDidFailLoadingMap:(MGLMapView *)mapView withError:(NSError *)error;
- (void)mapViewWillStartRenderingMap:(MGLMapView *)mapView;
- (void)mapViewDidFinishRenderingMap:(MGLMapView *)mapView fullyRendered:(BOOL)fullyRendered;
```

"Assets" doesn't really have a context for regular map view of the map. If this is specific to benchmarking, we should make a `MGLBenchMarkDelegate` protocol which extends `MGLMapViewDelegate` (or, doesn't have to) for these purposes.

include/mbgl/ios/MGLMapView.h
incanus: Also, is there an end user common need for this initializer? They should be pared down to just the minimums to avoid confusion on different ways to create the map. Don't I want a cache? That sounds beneficial? Why wouldn't I cache? Does that mean file, memory, or both? These will show up in support.

include/mbgl/ios/MGLMapView.h
kkaefer: Removed this.

include/mbgl/ios/MGLMapView.h
kkaefer: How can I make these available for the Benchmark application, but not for other applications?

incanus: Best way would be a header called e.g. `MGLMapView_Private.h`.

include/mbgl/ios/MGLMapView.h
kkaefer: Removed this again; it was not needed for the benchmark app.

include/mbgl/ios/MGLMapView.h
kkaefer: Moved this to a private header.

Issue #1381, opened 2015-04-30
Set Metrics Location Accuracy To 3KM


bleege:
The [CLLocationManager that is responsible for tracking location metrics](https://github.com/mapbox/mapbox-gl-native/blob/bbcc7c26d9862f4fb8685174330a73132d8c8a27/platform/ios/MGLMetricsLocationManager.m#L17-L19) currently relies on the [default `desiredAccuracy` setting](https://developer.apple.com/library/ios/documentation/CoreLocation/Reference/CLLocationManager_Class/index.html#//apple_ref/occ/instp/CLLocationManager/desiredAccuracy) which is `kCLLocationAccuracyBest`.  This means that it will try to make use of GPS and therefore use more battery.  Change this to `kCLLocationAccuracyThreeKilometers` so that it will try to rely more on Wifi / Cell Towers location strategies which will likely be less battery intense.  This will also enable developers to have more control over what level of accuracy they want / need in their apps.

/cc @mapbox/mobile @sbma44 @camilleanne @mick @edenh @gundersen @aaronlidman @morganherlocker @TheMarex

sbma44:
:+1: -- once this is rolled let's plan to re-test Sirius to confirm its power usage looks like Swarm's. I'm guessing that tomorrow will be the right day for that given today's schedule/HOT party.

bleege:
Tested in Mapbox GL on iPod Touch on Garage Wifi.  The user dot is using the MapView's CLLocationManager which uses the default `desiredAccuracy` setting of `kCLLocationAccuracyBest` (which makes sense).  Either way this is a good initial example of what WiFi only mode looks like IRL.

![20150430-mapbox-gl-ipod-touch](https://cloud.githubusercontent.com/assets/989928/7417618/4255a53a-ef35-11e4-90d8-e4a28965741c.png)

bleege:
Waiting on Mr Travis.

bleege:
Travis approved and merged.

When doing Trace tests let's pay attention to the `horizontalAccuracy` field to see how much this changes in the new world.

PR #1383, opened 2015-04-30
Throttling The Requested Accuracy Level Down

bleege:
For #1381


picciano:
Will this only affect metrics? (I hope.) Or does it also affect the userLocation for the map view?
If so, how do we change it back to `kCLLocationAccuracyBest` from the client app?

bleege:
@picciano Yep, this is the Metrics Location Manager only.  It's independent of the map's user dot (which is `kCLLocationAccuracyBest` AND any Location Manager that developers would have in their apps.  Just trying to lighten the request load.

PR #1390, opened 2015-04-30
V8 merge branch

lbud:
~~@tmcw — after a long unsuccessful rebase attempt I decided to experiment with a different strategy — rebased master/v8types (which I had branched off of it yesterday), squashed some of the previous v8 commits and cherry-picked them here, which means a really clean diff…*however* this is failing :[ :[ :[ : looks like it's…leaking memory somewhere maybe? TBD, going to continue digging tomorrow.~~
Edit: think something was just caching in xcode; fixed.

* [ ] Fix test suite
* [ ] Rebase to be mergable


mb12:
Can you please also make changes to issue warning in XCode debugger spew if style.json uses older property names(symbol-min-distance, line-image, background-image)? It will save debugging time for those using styles modified from v7.

PR #1392, opened 2015-05-01
Fix some bugs found via valgrind

jfirebaugh:


tmpsantos:
@jfirebaugh are you using the suppresion file? https://github.com/mapbox/mapbox-gl-native/blob/master/scripts/valgrind.sup

I'm wondering if it is useful on mac too...

jfirebaugh:
Nice, I wasn't aware of that. I'll try it out next time.

PR #1400, opened 2015-05-01
Avoid creating MGLMapboxEvents when Metrics is disabled

1ec5:
Mapbox Metrics already has a bottleneck; might as well use it to avoid fetching data in the first place. If Metrics is disabled, we avoid creating the shared `MGLMapboxEvents` instance; any public methods that would otherwise message that shared instance will instead message `nil`, which is a no-op in Objective-C.

/cc @bleege


1ec5:
I’d love to add a test for this, but I don’t see a way to revoke Location Services authorization. If that turns out to be possible, we can always add a test later.

PR #1406, opened 2015-05-04
Fix integer division in transform.cpp

springmeyer:
Ran mbgl-native through coverity static analysis tonight. Coverity oncovered that our transform code has been doing integer division. This may be leading to loss of precision / truncated values - a classic C/C++ bug. If anyone is not familiar with this behavior have a look at: http://coliru.stacked-crooked.com/a/62dd1db484efeac8

The report looked like:

screen shot 2015-05-03 at 11 30 59 pm

Fuller report at https://scan4.coverity.com/reports.htm#v13662/p11099 (you should be able to login via github account).

Overall: I hit two coverity bugs that prevented most files from being able to be processed that I've emailed coverity support about. It looks like coverity does not support `atomic` and there are some libc++ coverity issues. I'll post more bug reports when I can process the entire libmgbl-core.

### DIFF
--- a/src/mbgl/map/transform.cpp
+++ b/src/mbgl/map/transform.cpp
@@ -199,14 +199,14 @@ void Transform::_setScale(double new_scale, double cx, double cy, const Duration
 
     // Zoom in on the center if we don't have click or gesture anchor coordinates.
     if (cx < 0 || cy < 0) {
-        cx = current.width / 2;
-        cy = current.height / 2;
+        cx = static_cast<double>(current.width) / 2.0;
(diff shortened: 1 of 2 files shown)


incanus:
Nice @springmeyer, this is enlightening.

springmeyer:
btw, you would think that `-Wsign-conversion` or `-Wconversion` would warn on this, but it does not in all cases (http://stackoverflow.com/questions/6076878/locating-numerical-errors-due-to-integer-division). So, we need to be smart about always using explicitly using `static_cast` when changing types. This will help prevent problems at the time you write the code and especially when others copy and paste to refactor.

Issue #1309, opened 2015-04-20
Assertion failure after backgrounding app during animation — Environment::abandonVAO()


1ec5:
I hit this assertion failure somewhat consistently on an iPhone 6 running iOS 8.3 after backgrounding the iOS demo app during a drift pan animation that caused more tiles to be fetched and bringing the app back to the foreground:

```
Assertion failed: (currentlyOn(ThreadType::Map)), function abandonVAO, file /Users/mxn/hub/mapbox-gl-native-2/src/mbgl/map/environment.cpp, line 144.
```

```
Worker (21)#0	0x0000000195a77270 in __pthread_kill ()
#1	0x0000000195b15170 in pthread_kill ()
#2	0x00000001959eeb18 in abort ()
#3	0x00000001959c3a3c in __assert_rtn ()
#4	0x00000001000efabc in mbgl::Environment::abandonVAO(unsigned int) at /Users/mxn/hub/mapbox-gl-native-2/src/mbgl/map/environment.cpp:144
#5	0x00000001000d0560 in mbgl::VertexArrayObject::~VertexArrayObject() at /Users/mxn/hub/mapbox-gl-native-2/src/mbgl/geometry/vao.cpp:15
#6	0x00000001000d05e0 in mbgl::VertexArrayObject::~VertexArrayObject() at /Users/mxn/hub/mapbox-gl-native-2/src/mbgl/geometry/vao.cpp:11
#7	0x00000001001c3834 in std::__1::array<mbgl::VertexArrayObject, 1ul>::~array() at /Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/../include/c++/v1/array:121
#8	0x00000001001c37b4 in std::__1::array<mbgl::VertexArrayObject, 1ul>::~array() at /Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/../include/c++/v1/array:121
#9	0x00000001001c374c in mbgl::ElementGroup<1>::~ElementGroup() at /Users/mxn/hub/mapbox-gl-native-2/src/mbgl/geometry/elements_buffer.hpp:14
#10	0x00000001001c36e4 in mbgl::ElementGroup<1>::~ElementGroup() at /Users/mxn/hub/mapbox-gl-native-2/src/mbgl/geometry/elements_buffer.hpp:14
[...]

Issue #1403, opened 2015-05-02
Use C++14 features / require gcc 4.9


jfirebaugh:
We've encountered several situations where we want to use C++14's [generalized lambda capture](https://isocpp.org/files/papers/N3648.html).

This is supported in [gcc 4.9](https://gcc.gnu.org/projects/cxx1y.html) and [clang 3.4](http://clang.llvm.org/cxx_status.html) (which we already require).

@springmeyer Do you forsee any issues requiring gcc 4.9 rather than 4.8?

jfirebaugh:
Notes from a previous effort to switch to gcc 4.9: https://github.com/mapbox/mapbox-gl-native/pull/1250#issuecomment-91736051

jfirebaugh:
Looking good [here](https://travis-ci.org/mapbox/mapbox-gl-native/builds/60976055).

Blocker seems to be getting a more recent compiler for the android builds:

https://travis-ci.org/mapbox/mapbox-gl-native/builds/60976494

springmeyer:
> @springmeyer Do you forsee any issues requiring gcc 4.9 rather than 4.8?

No. Although it would be interesting to test on ubuntu trusty with the stock 4.9 that comes there - would be nice if that worked and users of at Ubuntu Trusty would not need to upgrade their compiler to get mbgl to build.

kkaefer:
Android builds should use clang 3.5 IIRC.

tmpsantos:
I would be very cautious about this. Apparently we want to support windows, no?

Check this why C++11 is not fully supported by Chromium (C++14 is not even considered):
https://chromium-cpp.appspot.com/

/cc @BergWerkGIS

jfirebaugh:
Summarizing from scrum:

* We don't currently build on Windows and don't forsee doing so in the near future.
* In addition to potential C++11/14 issues, we'd need to port GL code or use [ANGLE](https://code.google.com/p/angleproject/).
* Support for the C++14 feature we need is in VS 2015 Preview, so may be generally available by the time we potentially want to support Windows.

Seems like we have consensus to start adopting C++14 features that have widespread compiler support as needed. I'm going to start with the feature required for #1309. Post-Beta we can look at using `std::make_unique` and any other things we have polyfilled.

jfirebaugh:
Done as part of #1412.

PR #1412, opened 2015-05-04
Ensure worker tasks are complete before destroying TileData

jfirebaugh:
This is the sure-fire fix for #1309. Should have gone this route earlier.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,23 +7,23 @@ matrix:
     - os: linux
   include:
     - os: linux
-      env: FLAVOR=linux CXX=g++-4.8 BUILDTYPE=Release
+      env: FLAVOR=linux CXX=g++-4.9 BUILDTYPE=Release
       addons:
         apt:
           sources: [ 'ubuntu-toolchain-r-test' ]
-          packages: [ 'gdb', 'g++-4.8', 'gcc-4.8', 'libllvm3.4', 'xutils-dev', 'libxxf86vm-dev', 'x11proto-xf86vidmode-dev', 'mesa-utils' ]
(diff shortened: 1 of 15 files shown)


incanus:
`README` is going to need updated for C++14 req.

springmeyer:
You mentioned at https://github.com/mapbox/mapbox-gl-native/issues/1403#issue-72661577 that only clang-3.4 is required. I see you bumped to clang-3.6 - is that needed specifically or did you just bump to the latest clang available for good practice? The reason I ask is that I've been noticing crashes which look like some kind of ABI incompatiblity when c++ binaries built with clang-3.5 are linked and run against other c++ binaries built with clang-3.6 https://github.com/mapnik/mapnik/issues/2785. I've not had a chance to investigate much but until I do I've been sticking with clang-3.5.

jfirebaugh:
@incanus README update is included.

@springmeyer [Previously](https://github.com/mapbox/mapbox-gl-native/pull/1250#issuecomment-91736051) I encountered a conflict between gcc 4.9 and clang 3.4/3.5. But the Travis infrastructure has changed in the meantime and this may no longer be an issue. I'll try reverting back to 3.4.

jfirebaugh:
clang-3.4 does not seem to support `-std=c++14`. I'll try 3.5.

incanus:
@jfirebaugh Would this be accurate to change as well in the `README`? 

> An OpenGL renderer for [Mapbox Vector Tiles](https://www.mapbox.com/blog/vector-tiles),
implemented in C++14, currently targeting iOS, OS X, Android, and Ubuntu Linux.

jfirebaugh:
Yes, I'll do that.

PR #1425, opened 2015-05-05
Fixes for static analyzer warnings

1ec5:
This PR fixes various warnings flagged by Xcode’s static analyzer. Mostly these changes fix leaks, but there is one user-visible change, which is that double-tapping while tracking user location no longer bounces around. See individual commit messages for details.

/cc @kkaefer @bleege


1ec5:
Hat tip goes to @kkaefer for remembering to run the static analyzer. I only followed its instructions. :smile:

1ec5:
Upstream bugs filed upstream: memononen/libtess2#19 and [angusj/polyclipping#134](https://sourceforge.net/p/polyclipping/bugs/134/).

PR #1429, opened 2015-05-05
break out setup.sh from run.sh

mikemorris:
Needed to avoid code duplication in node-mapbox-gl-native travis scripts, refs https://github.com/mapbox/mapbox-gl-native/pull/1339, https://github.com/mapbox/node-mapbox-gl-native/pull/119.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -82,14 +82,20 @@ env:
     - secure: "nQqSM8rd7OHtV4MqmNqVnkrVHqxKqQsaWRYk4/nPdhbeVWtTtkk0df711LrF1TUtbEPEewHxYUvTZ/UXmwJNeoKdzTHavI8hnatRkgjyxGERPn1il1Otelht9I+LQQHf+plrpRjVWBrNIW0Zox1B3cqn6d3NglpbXrEQ2EjYGNA="
     - KIF_SCREENSHOTS="${TRAVIS_BUILD_DIR}/screenshots"
 
+before_install:
+- source ./scripts/travis_helper.sh
+
 install:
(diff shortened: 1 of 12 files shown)


jfirebaugh:
This regressed headless tests. They are no longer using the version of mesa installed via mason, because loading that version depends on [setting `LD_LIBRARY_PATH`](https://github.com/mapbox/mapbox-gl-native/blob/8e15732954dfc5635490d42fff7142e793a5208c/scripts/linux/setup.sh#L24) and this moved the export from `run.sh` to a separate `setup.sh` shell file whose exports aren't available from `run.sh`.

jfirebaugh:
Oops, sorry, I misdiagnosed the cause. It's still this change, but it's because the `mason prefix` call within the `export LD_LIBRARY_PATH` command happens before `mason install` is run.

jfirebaugh:
I've reverted this for the time being so we have passing builds going into the final hours of the iOS b1 release.

mikemorris:
That seems odd, `setup.sh` is still sourced before `run.sh` is called, and Mesa should be installed in `install.sh`, call before `setup.sh`.

- https://github.com/mapbox/mapbox-gl-native/blob/8e15732954dfc5635490d42fff7142e793a5208c/.travis.yml#L88-L92
- https://github.com/mapbox/mapbox-gl-native/blob/8e15732954dfc5635490d42fff7142e793a5208c/scripts/linux/install.sh#L10-L11

mikemorris:
-> https://github.com/mapbox/mapbox-gl-native/pull/1527

PR #1441, opened 2015-05-06
Update to lastest Android SDK

ljbade:


Makefile
```diff
@@ -178,11 +178,11 @@ android-lib: android-lib-$(ANDROID_ABI)
 
 # Builds the selected/default Android library
 android: android-lib-$(ANDROID_ABI)
-	cd android/java && ./gradlew --parallel-threads=$(JOBS) assemble$(BUILDTYPE)
```
kkaefer: Why did you remove this?

ljbade: Good catch.

ljbade: I was experimenting with seeing if could update to Gradle 2.4 which deprecated --parallel-thread in favour of a new flag - https://gradle.org/docs/2.4/release-notes#setting-number-of-build-execution-threads-with---parallel-threads

But for some reason even after changing the Gradle bootstrap version number it kept using 2.2.1 so I forgot to put the old flag back in. Will fix.

PR #1443, opened 2015-05-06
Add style change functionality to the Linux test app

tmpsantos:
Makes the life of people using Linux as development environment way
easier.

Just press 's' to cycle through 3 different styles.

### DIFF
--- a/include/mbgl/platform/default/glfw_view.hpp
+++ b/include/mbgl/platform/default/glfw_view.hpp
@@ -31,6 +31,8 @@ public:
 
     void setShouldClose();
 
+    void setWindowTitle(const std::string&);
+
     void run();
     void fps();
 
--- a/include/mbgl/platform/default/glfw_view.hpp
+++ b/include/mbgl/platform/default/glfw_view.hpp
@@ -29,22 +29,23 @@ public:
     // The expected action is to set a new style, different to the current one.
     void setChangeStyleCallback(std::function<void()> callback);
 
+    void setShouldClose();
+
     void run();
     void fps();
 
-    GLFWwindow *window = nullptr;
-
 private:
     bool fullscreen = false;
+    bool tracking = false;
+    bool rotating = false;
 
     double lastX = 0, lastY = 0;
-    bool tracking = false;
-
-    bool rotating = false;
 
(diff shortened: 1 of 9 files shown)


tmpsantos:
@kkaefer please review

include/mbgl/platform/default/glfw_view.hpp
```diff
@@ -25,10 +25,16 @@ class GLFWView : public mbgl::View {
...
 
+    // Callback called when the user presses the key mapped to style change.
+    // The expected action is to set a new style, different to the current one.
+    void setChangeStyleCallback(std::function<void()> callback);
+
     void run();
     void fps();
 
-public:
+    GLFWwindow *window = nullptr;
```
kkaefer: Why is this public?

tmpsantos: > Why is this public?

Everything was public before. This one was kept because of:
`linux/main.cpp:31:        glfwSetWindowShouldClose(view->window, true);`

tmpsantos: I will make another PR fixing this.

incanus:
We should probably get it on the default shipping styles — Streets, Emerald, Light, and Dark.

ljbade:
Finally! Will make the Linux test app much faster for debugging different styles. I use this a lot to compare rendering to what I get on Android.

Good work @tmpsantos.

platform/default/default_styles.cpp
```diff
@@ -0,0 +1,18 @@
...
+    { "asset://styles/light-v7.json", "Light" },
+    { "asset://styles/dark-v7.json", "Dark" },
+    { "asset://styles/bright-v7.json", "Bright" },
+    { "asset://styles/basic-v7.json", "Basic" },
+    { "asset://styles/outdoors-v7.json", "Outdoors" },
+    { "asset://styles/satellite-v7.json", "Satellite" },
+};
+
+} // end namespace util
+} // end namespace mbgl
```
incanus: We should ticket out use of this `vector` by the iOS and Android apps, too, since this logic is now duplicated.

Issue #1396, opened 2015-05-01
Enable Metrics To TEST Server


bleege:
The [current implementation hardcodes the URL](https://github.com/mapbox/mapbox-gl-native/blob/51c069502880e38a6dabf520c032fd6f293ee744/platform/ios/MGLMapboxEvents.m#L447) endpoint to be [the PROD data server](https://github.com/mapbox/mapbox-gl-native/blob/51c069502880e38a6dabf520c032fd6f293ee744/platform/ios/MGLMapboxEvents.m#L16).  Need to enable access to push to TEST server to help minimize "data pollution" during ongoing development.

/cc @mapbox/mobile @camilleanne @mick @edenh @willwhite

1ec5:
@bleege and I were brainstorming about ways to enable support for a test server without either exposing that server to the public or enabling developers on free plans to override the server with their own. In the end, it sounds like we may just have to do the most straightforward thing codewise and rely on an honor system, just as we do for attribution. But if there is a stricter way to enforce the use of Mapbox’s Metrics server, we should, because there will be a much greater desire to avoid Mapbox Metrics than Mapbox branding.

1ec5:
The most straightforward thing to do would be to have `MGLMapboxEvents` implement a `-hostname` method. We shouldn’t hardcode the test server in here because (presumably) it wouldn’t be accessible to developers outside of Mapbox. So you’ll probably have to figure out how internal apps for testing the Metrics server should know about that endpoint.

bleege:
Moved to iOS Beta 2 due to internal discussion today.

bleege:
Look into using [`Info.plist` proposal](https://github.com/mapbox/mapbox-gl-native/issues/1535) to set Metrics server url.  This will hopefully allow for dynamic file substitution at library build time so that the default url will be set to TEST, and release builds can have the PROD url substituted.

bleege:
> We shouldn’t hardcode the test server in here because (presumably) it wouldn’t be accessible to developers outside of Mapbox.

I'm thinking that the way forward on this is to actually hardcode in the TEST URL for the project.  This will ensure that internal and external GL developers are always only sending data to TEST infrastructure during GL development.  The only time that Metrics data should be sent to the PROD infrastructure is for apps implementing the released SDK which is published on Cocoapods.  Of course, that begs the question of how likely do we think people will roll their own iOS GL SDK from source?

bleege:
[...]

PR #1445, opened 2015-05-06
Enabling TEST Metrics Server

bleege:
For #1396


include/mbgl/ios/MGLMapboxEvents.h
```diff
@@ -33,6 +33,7 @@ extern NSString *const MGLEventGestureRotateStart;
 // You must call these methods from the main thread.
 //
 + (void) setToken:(NSString *)token;
++ (void) setAlternateDataServerURL:(NSString *)alternateDataServerURL;
```
1ec5: There should be a getter to balance the setter.

platform/ios/MGLMapboxEvents.m
```diff
@@ -82,6 +82,7 @@ @interface MGLMapboxEvents()
 @property (atomic) NSURLSession *session;
 @property (atomic) NSData *digicertCert;
 @property (atomic) NSData *geoTrustCert;
+@property (atomic) NSString *alternateDataServerURL;
```
1ec5: This should be an `NSURL`, given the name.

PR #1449, opened 2015-05-06
Set accessToken in initWithCoder

friedbunny:
Fix `[MGLMapView initWithCoder:]` by adding support for `[MGLAccountManager setAccessToken:]`, which is useful when using storyboards.

Also cleans up unnecessary header imports in the demo app.

/cc @bleege @1ec5 @incanus

### DIFF
--- a/platform/ios/MGLMapView.mm
+++ b/platform/ios/MGLMapView.mm
@@ -142,6 +142,7 @@ std::chrono::steady_clock::duration secondsAsDuration(float duration)
     if (self && [self commonInit])
     {
         self.styleURL = nil;
+        self.accessToken = [MGLAccountManager accessToken];
         return self;
     }
 
--- a/ios/app/MBXAppDelegate.m
+++ b/ios/app/MBXAppDelegate.m
@@ -1,8 +1,6 @@
 #import "MBXAppDelegate.h"
 #import "MBXViewController.h"
 #import <mbgl/ios/MapboxGL.h>
-#import <mbgl/ios/MGLAccountManager.h>
-#import <mbgl/ios/MGLMapboxEvents.h>
 
 @implementation MBXAppDelegate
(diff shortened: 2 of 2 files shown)


incanus:
Cool, thanks. We're keeping this on b2 though — code :snowflake: right now.

platform/ios/MGLMapView.mm
```diff
@@ -142,6 +142,7 @@ - (instancetype)initWithCoder:(NSCoder *)decoder
     if (self && [self commonInit])
     {
         self.styleURL = nil;
+        self.accessToken = [MGLAccountManager accessToken];
```
1ec5: Note that this calls through to `-setAccessToken:`, which calls through to `Map::setAccessToken()` unconditionally. I think it should be fine, though.

friedbunny: I mimicked [`[MGLMapView initWithFrame:]`](https://github.com/friedbunny/mapbox-gl-native/blob/initWithCoder-access-token/platform/ios/MGLMapView.mm#L113) — is there any difference here that we should be cautious of?

1ec5: No difference, but I guess we should get `-[MGLMapView setAccessToken:]` to bail on `nil`.

1ec5:
Come to think of it, wouldn’t it be better to look for a `+[MGLAccountManager setAccessToken:]` in `-commonInit` (if one hasn’t been set explicitly via `-initWithFrame:accessToken:`)? That way the `MGLMapView` can “inherit” an access token from `MGLAccountManager` even if `-initWithFrame:` is called.

friedbunny:
No longer necessary with @1ec5's global token rejiggering in 256b748458571201fe2b64bdf84dae6ec9b5932e.

PR #1453, opened 2015-05-06
Prevent full-circle spin during animated rotation

1ec5:
#1295 made #1199 a lot less common, but it was still possible for the map to do a full 180° when drift-rotating past due-south. With this PR, the map is atomically rotated to a bearing in the same cycle as the final bearing.

/cc @kkaefer @friedbunny


friedbunny:
Looking good, I couldn't break this.

PR #1457, opened 2015-05-07
Added convenience target to run with Valgrind on OSX and Linux

tmpsantos:
The target (run-valgrind-[osx|linux]) will launch the test app
with valgrind using the leak checker and our custom suppresion file.


tmpsantos:
@kkaefer please review

I've been using this a lot. :)

PR #1461, opened 2015-05-07
Do not process partial tiles if no new resources have arrived

### LINKED ISSUE #1458: Partial tiles constantly being reparsed
When a tile is partially parsed (i.e. it is waiting for more data from the network like glyphs and sprites), we are reparsing it everytime we call `update()`. We should only reparse partial tiles when new glyphs or sprites arrive.

That would avoid the CPU being 100% all the time until everything is downloaded and partial tiles are on the viewport.

tmpsantos:
Partial tiles are now only processed again when there is a new resource available. If we don't have new resources - like new glyphs - the end result of the reparsing is the same, because no new SymbolBuckets would are created.

In my tests debugging with PowerTOP, this patch drops the CPU utilization during the sampling period from 100ms/s to 40ms/s. The improvements on slow connections (i.e. more time waiting for resources) are even more noticeable.

Fixes #1458


tmpsantos:
@kkaefer @jfirebaugh please review

PR #1464, opened 2015-05-07
Parse feature types in set filters

1ec5:
A filter key of `$type` should work in an `in` filter, just like in an `==` filter. This fixes an issue exposed by mapbox/mapbox-gl-styles@4a89b5d and mapbox/mapbox-gl-styles@eb477f6, wherein minor streets and their casings completely dropped out of the map.

/cc @jfirebaugh @peterqliu


jfirebaugh:
Fix looks good. Can you port [this JS test](https://github.com/mapbox/feature-filter/blob/d3198bc00513de013e10dcdb7332056379f45776/test.js#L228-L234) to [the other native tests](https://github.com/mapbox/mapbox-gl-native/blob/master/test/miscellaneous/comparisons.cpp)?

1ec5:
Ah, so that’s where the JS equivalent lives!

PR #1472, opened 2015-05-08
Render tool is broken

tmpsantos:
Fix the broken build. We should probably build it on the bots, but we need to wait for https://github.com/travis-ci/travis-ci/issues/3759 first.

### DIFF
--- a/bin/render.cpp
+++ b/bin/render.cpp
@@ -79,11 +79,8 @@ int main(int argc, char *argv[]) {
         }
     }
 
-
     HeadlessView view;
-    Map map(view, fileSource);
-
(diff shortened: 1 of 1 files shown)


scripts/linux/run.sh
```diff
@@ -42,6 +42,7 @@ git submodule update --init styles
 
 mapbox_time "compile_program" \
 make linux -j${JOBS} BUILDTYPE=${BUILDTYPE}
+make render -j${JOBS} BUILDTYPE=${BUILDTYPE}
```
kkaefer: :+1:

Issue #1294, opened 2015-04-18
Map center changes randomly/inconsistently when zooming in while panning animation is on (iOS)


mb12:
Map center changes inconsistently/randomly if one zooms in/out (using animated:NO) while panning animation is on. For e.g. if you are panning in land area with animation, zooming in/out changes map center so randomly that one ends  up in sea/water area.

This bug is consistently reproducible using the following steps:

1.) Add a stepper control that increments/decrements zoom without animation. (This is to be able to consistently zoom in/out without animation while panning animation is running).
2.) Pan with animation.
3.) Use the stepper controls to zoom in/out.

1ec5:
Hm, I think this may be what's complicating the zoom animation I'm working on for #1292.

incanus:
This probably also relates to the transform problems over in #1125 & https://github.com/mapbox/mapbox-gl-native/issues/1160.

incanus:
Is this still relevant @mb12? I don't think we see this anymore.

mb12:
You are right. Its not relevant any more. I haven't noticed random center changes on iOS for quite some time.

PR #1446, opened 2015-05-08
Prevent panning to NaN territory

kkaefer:
Rapid panning and rotation sometimes lets the map view end up in a state where both x and y coordinates are NaN. Typically, the only thing rendering will be the background color. Once you're in this state, you can set a breakpoint at `Painter::render()` and inspect `state`'s `x` and `y` variables.

### DIFF
--- a/platform/ios/MGLMapView.mm
+++ b/platform/ios/MGLMapView.mm
@@ -843,6 +843,11 @@ std::chrono::steady_clock::duration secondsAsDuration(float duration)
     else if (pinch.state == UIGestureRecognizerStateEnded || pinch.state == UIGestureRecognizerStateCancelled)
     {
         CGFloat velocity = pinch.velocity;
+        if (isnan(velocity))
+        {
+            // UIPinchGestureRecognizer sometimes returns NaN for the velocity
+            velocity = 0;
+        }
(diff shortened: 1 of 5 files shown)


jfirebaugh:
Any idea where the NaNs come from (other than `UIPinchGestureRecognizer`)? I'm happy taking the `Transform` guard clauses for b1 but finding and eliminating the upstream causes would be better long term.

mb12:
@jfirebaugh isnan check also needs to be done in to be done handlePanGesture: selector as well.

kkaefer:
I believe they're only coming from the gesture recognizer. I have never observed NaN issues in anything but the iOS app.

mb12:
handlePanGesture also calls [UIPanGestureRecognize velocity].

I see this consistently with the following steps on iOS:
1. Trigger pan animation using fast flick.
2. Zoom in using a stepper control.

1ec5:
@mb12, are you seeing `-[UIPanGestureRecognizer velocity]` itself return NaN for a standard flick? #1294 could’ve been caused by the lower-level transform issues that @kkaefer fixed in this PR.

mb12:
Sorry for the confusion. Everything looks good.

PR #1479, opened 2015-05-08
clang-3.7

springmeyer:
Adds support for clang-3.7 on OS X by silencing a  -Wbraced-scalar-init coming from `<atomic>`.

```
../../src/mbgl/map/tile_data.cpp:15:15: error: braces around scalar initializer [-Werror,-Wbraced-scalar-init]
      parsing(ATOMIC_FLAG_INIT),
              ^~~~~~~~~~~~~~~~
/Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/include/c++/v1/atomic:1762:26: note: expanded from macro 'ATOMIC_FLAG_INIT'
#define ATOMIC_FLAG_INIT {false}
                         ^~~~~~~
1 error generated.
make[1]: *** [Debug/obj.target/core/src/mbgl/map/tile_data.o] Error 1
make: *** [linux] Error 2
```

refs https://github.com/llvm-mirror/libcxx/blob/master/include/atomic#L1776

For reference I'm running clang source compiled from git:

```
$ /opt/llvm/bin/clang++ -v
clang version 3.7.0 (http://llvm.org/git/clang.git de723f5818e668f45fae30510c403eb7e4f09568) (http://llvm.org/git/llvm.git 89d0e99f863f906d64727f168686565248063a9a)
Target: x86_64-apple-darwin14.3.0
Thread model: posix
```

Built like: https://github.com/springmeyer/build-llvm

### DIFF
--- a/src/mbgl/map/tile_data.cpp
+++ b/src/mbgl/map/tile_data.cpp
@@ -12,7 +12,10 @@ TileData::TileData(const TileID& id_, const SourceInfo& source_)
     : id(id_),
       name(id),
       state(State::initial),
+#pragma GCC diagnostic push
+#pragma GCC diagnostic ignored "-Wbraced-scalar-init"
       parsing(ATOMIC_FLAG_INIT),
+#pragma GCC diagnostic pop
       source(source_),
       env(Environment::Get()),
       debugBucket(debugFontBuffer) {
--- a/src/mbgl/map/tile_data.cpp
+++ b/src/mbgl/map/tile_data.cpp
@@ -13,6 +13,7 @@ TileData::TileData(const TileID& id_, const SourceInfo& source_)
       name(id),
       state(State::initial),
 #pragma GCC diagnostic push
+#pragma GCC diagnostic ignored "-Wunknown-pragmas"
 #pragma GCC diagnostic ignored "-Wbraced-scalar-init"
       parsing(ATOMIC_FLAG_INIT),
 #pragma GCC diagnostic pop


jfirebaugh:
Can we initialize this in the class body instead, like:

```
    std::atomic_flag parsing = ATOMIC_FLAG_INIT;
```

That's the [only form of initialization](http://en.cppreference.com/w/cpp/atomic/ATOMIC_FLAG_INIT) that's specified by the standard.

tmpsantos:
I tested this locally here and the compiler was not complaining, even with my `ATOMIC_FLAG_INIT` defined to `{ 0 }`. I moved the initialization to the class definition here #1499 to comply with the spec. Thanks for pointing this out.

```
$ clang++-3.7 -v
Ubuntu clang version 3.7.0-svn227076-1 (trunk) (based on LLVM 3.7.0)
```

springmeyer:
#1499 compiles fine with clang-3.7. Thanks @tmpsantos! Closing this one as superseded by #1499

tmpsantos:
@springmeyer oh, I tested here and the `make test` target is broken on clang 3.7

```
  CXX(target) Debug/obj.target/test/test/miscellaneous/rotation_range.o
../../test/miscellaneous/rotation_range.cpp:34:35: error: no viable conversion from 'CollisionRange'
      (aka 'array<CollisionAngle, 2>') to 'float'
                  CollisionList({{CollisionRange({{1.0 / 2.0 * M_PI, M_PI}})}}),
```

springmeyer:
@tmpsantos - interesting, not seeing that on my system.

PR #1480, opened 2015-05-08
Rename settings keys

1ec5:
The new keys fit the standard naming convention for Info.plist keys and more clearly communicate how they’re to be used. [The documentation](https://github.com/mapbox/mapbox-gl-native/wiki/Installing-Mapbox-GL-for-iOS#configuring-mapbox-metrics) needs to be updated in tandem with merging this PR.

/cc @bleege


bleege:
:+1:

PR #1484, opened 2015-05-08
Fix potential integer overflow in line_atlas

springmeyer:
Spotted thanks to `clang++ -fsanitize=undefined`:

```
line_atlas.cpp:115: runtime error: value 249 is outside the range of representable values of type 'char'
line_atlas.cpp:115: runtime error: value 249 is outside the range of representable values of type 'char'
line_atlas.cpp:115: runtime error: value 250 is outside the range of representable values of type 'char'
line_atlas.cpp:115: runtime error: value 251 is outside the range of representable values of type 'char'
line_atlas.cpp:115: runtime error: value 252 is outside the range of representable values of type 'char'
line_atlas.cpp:115: runtime error: value 253 is outside the range of representable values of type 'char'
line_atlas.cpp:115: runtime error: value 254 is outside the range of representable values of type 'char'
line_atlas.cpp:115: runtime error: value 255 is outside the range of representable values of type 'char'
line_atlas.cpp:115: runtime error: value 255 is outside the range of representable values of type 'char'
line_atlas.cpp:115: runtime error: value 255 is outside the range of representable values of type 'char'
line_atlas.cpp:115: runtime error: value 255 is outside the range of representable values of type 'char'
line_atlas.cpp:115: runtime error: value 255 is outside the range of representable values of type 'char'
```

### DIFF
--- a/src/mbgl/geometry/glyph_atlas.cpp
+++ b/src/mbgl/geometry/glyph_atlas.cpp
@@ -4,6 +4,8 @@
 #include <mbgl/platform/log.hpp>
 #include <mbgl/platform/platform.hpp>
 
+#include <mbgl/util/std.hpp>
+
 #include <cassert>
 #include <algorithm>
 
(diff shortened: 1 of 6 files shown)


kkaefer:
We should replace this with `std::unique_ptr` managed code.

kkaefer:
Please review.

incanus:
Similar to https://github.com/mapbox/mapbox-gl-native/pull/1485, does this have any practical implications for b1 or is it just housekeeping?

PR #1485, opened 2015-05-08
fix integer overflow in line_buffer.cpp

springmeyer:
Spotted thanks to `clang++ -fsanitize=undefined`:

```
line_buffer.cpp:18: runtime error: value -146 is outside the range of representable values of type 'signed char'
line_buffer.cpp:18: runtime error: value -761 is outside the range of representable values of type 'signed char'
line_buffer.cpp:18: runtime error: value 761 is outside the range of representable values of type 'signed char'
```


kkaefer:
The underlying reason is that the line tessellation algorithm generates an extrude length that is way too long. We should find the reason that generates the overly long extrude vector and fix that instead.

incanus:
What's the practical implication of this? Anything? Can we push to b2?

jfirebaugh:
Thanks @springmeyer; closing here since the branch was stale and we need to address the underlying issue instead.

PR #1491, opened 2015-05-09
Refactor MGLMapboxEvents

1ec5:
This is a significant refactoring of `MGLMapboxEvents` that ensures it only hangs onto user-specific data when permitted. It consolidates logic around enabling and disabling services and removes some indirection that made it difficult to understand ownership and lifespan of those services. The salient behavioral changes are:

* `MGLMapboxEvents` shuts off as soon as the user opts out. (If the user opt-out is implemented in Settings, it shuts off as soon as the user returns to the host app.)
* After the user turns off Location Services and turns it back on, `MGLMapboxEvents` restarts its location manager as soon as the user returns to the host app.
* `MGLMapboxEvents` defaults to sending events even if the developer has failed to implement a user opt-out.
* If a Settings.bundle is present but omits the Mapbox Metrics switch, the default defaults to being on.
* If the developer has indicated that the Mapbox Metrics switch is implemented in-app but in reality it isn’t implemented, the default defaults to being on.

Architectural improvements include:

* The autonomous `MGLMetricsLocationManager` singleton no longer exists. `MGLMapboxEvents` is the clearinghouse for all Metrics-related logic.
* User-specific data is stored in a self-contained object that is completely discarded whenever `MGLMapboxEvents` is paused.
* `MGLMapboxMetrics` no longer stores the access token separately. Instead, it gets it from `MGLAccountManager`. `MGLMapView` forwards access tokens to `MGLAccountManager` as well.

Some things to test (please chime in if you can think of anything else):

* Various `MGLMapboxAcccountType`s
* User opt-out while Location Services enabled/disabled
* User opt-in while Location Services enabled/disabled
* Disable/enable Location Services while user is opted in/out
  * Disable Location Services, opt out, enable Location Services, opt in
  * Opt out, disable Location Services, opt in, enable Location Services
* Setting access token via `+[MGLAccountManager setAccessToken:]` and/or `-[MGLMapView setAccessToken:]` – ensure `MGLAccountManager`, `MGLMapView`, and `MGLMapboxEvents` are all aware of access token changes

/cc @bleege @jfirebaugh @incanus


1ec5:
I’m placing this PR on beta 2 for now, given its size, but I’m open to pulling in into beta 1 if we get enough testing.

mb12:
When metric collections is paused/disabled,  is there any plan to to avoid calls to convert point to lat long, create dictionaries etc? This computation is not really necessary if collection is paused or disabled by the user .

bleege:
Good stuff @1ec5!

> I’m placing this PR on beta 2 for now, given its size, but I’m open to pulling it into beta 1 if we get enough testing.

**Totally agree** with this decision.  Let's not rush things at this stage of the game, especially with something this big.  :+1:

1ec5:
> When metric collections is paused/disabled, is there any plan to to avoid calls to convert point to lat long, create dictionaries etc?

With these changes, none of that work should be happening. Most of it is [guarded by `paused`](https://github.com/mapbox/mapbox-gl-native/blob/1ec5-metrics-opt-way-out/platform/ios/MGLMapboxEvents.m#L423), and additionally we [clear out `eventQueue`](https://github.com/mapbox/mapbox-gl-native/blob/1ec5-metrics-opt-way-out/platform/ios/MGLMapboxEvents.m#L346), so it should never reach the point where event data would need to be constructed.

Incidentally, the rather aggressive approach to jettisoning data on pause means we do need to [rebuild all this data](https://github.com/mapbox/mapbox-gl-native/blob/1ec5-metrics-opt-way-out/platform/ios/MGLMapboxEvents.m#L73-97) on resume. If you’re interested, this is a good opportunity to profile `MGLMapboxMetrics` for performance issues. Hopefully it keeps a very light touch overall.

mb12:
I am specifically referrring calls to selectors like the two following in MGLMapView.mm. Will calls to  these also be protected if collection is paused? 

- (void)trackGestureEvent:(NSString *)gestureID forRecognizer:(UIGestureRecognizer *)recognizer
- (CLLocationCoordinate2D)convertPoint:(CGPoint)point toCoordinateFromView:(UIView *)view

1ec5:
In my rudimentary testing/profiling with quite a bit of finger-mashing, `-[MGLMapView trackGestureEvent:forRecognizer:]` takes up a minuscule fraction of the time spent in the gesture recognizer callbacks: roughly 20% of the time spent in `-handlePanGesture:`. `-convertPoint:toCoordinateFromView:` doesn’t even show up in my time profiles.

mb12:
Thank you very much @1ec5 for kindly clarifying all my queries. I really appreciate it.

1ec5:
I just rebased and force-pushed this branch due to some reverts on master.

1ec5:
Note when merging: [this assertion](https://github.com/mapbox/mapbox-gl-native/blob/87551c6d0e189cd6d98cc75e93ba4f0f176a3464/platform/ios/MGLMapboxEvents.m#L126) needs to stay in `-init`; otherwise, if it moves along with the rest of the method body to `+initialize`, developers will have no way of supressing it, no matter how soon they call the setter.

1ec5:
We decided to try to squeeze these changes into beta 1 pending some internal testing. Here goes…

platform/ios/MGLAccountManager.m
```diff
@@ -38,15 +38,15 @@ + (instancetype) sharedInstance {
 }
 
 + (void) setAccessToken:(NSString *) accessToken {
-    [[MGLAccountManager sharedInstance] setAccessToken:accessToken];
+    [[MGLAccountManager sharedManager] setAccessToken:accessToken];
 
     // Update MGLMapboxEvents
     // NOTE: This is (likely) the initial setup of MGLMapboxEvents
-    [MGLMapboxEvents setToken:accessToken];
+    [MGLMapboxEvents sharedManager];
```
incanus: Is there potential here for this call to be optimized away by ARC since it's not referred to strongly anywhere? Dumb idea, but curious about it.

include/mbgl/ios/MGLMapboxEvents.h
```diff
@@ -58,4 +54,7 @@ extern NSString *const MGLEventGestureRotateStart;
 //
 + (void) flush;
 
+// Main thread only
++ (void)validate;
+
 @end
```
incanus: Why is this header public at all? The user never needs to call it and it's not brought into `MapboxGL.h`.

platform/ios/MGLAccountManager.m
1ec5: No, the actual object is held by [a static global](https://github.com/mapbox/mapbox-gl-native/blob/1ec5-metrics-opt-way-out/platform/ios/MGLAccountManager.m#L16), so it will never go away for the lifetime of the process.

platform/ios/MGLMapboxEvents.m
```diff
@@ -136,7 +195,26 @@ - (instancetype) init {
             
             [[NSUserDefaults standardUserDefaults] registerDefaults:defaultsToRegister];
         }
+    }
+}
+
++ (BOOL)isEnabled {
+    return ([[NSUserDefaults standardUserDefaults] boolForKey:@"MGLMapboxMetricsEnabled"] &&
+            [[NSUserDefaults standardUserDefaults] integerForKey:@"MGLMapboxAccountType"] == 0);
```
incanus: Can we `assert()` here to ensure that the values are set and not just check their values? Because we are dealing with `BOOL` and `NSInteger`, I could see future potential traps around these being `nil` in some circumstances. If we can ensure that they are non-`nil` we can be proactive about those.

include/mbgl/ios/MGLMapboxEvents.h
1ec5: Originally `+pauseMetricsCollection` and `+resumeMetricsCollection` were designed to be called by the developer if necessary. I take it [that’s no longer the case](https://github.com/mapbox/mapbox-gl-native/wiki/Installing-Mapbox-GL-for-iOS/_compare/dd6d6bfd664b9cc3ee7fb32c2ea56854c6a21e2e...7a2a036bc0f66242d294f6972c5fbd932af31207#diff-81e1e4fc5b62518da42b4d3ceb9909faL126)?

platform/ios/MGLMapboxEvents.m
1ec5: `-[NSUserDefaults registerDefaults:]` guarantees that the defaults will always fall back to the passed-in dictionary values if unset. `-registerDefaults:` is called inside `+initialize`, which is guaranteed to be called when MapboxGL.a is loaded. I can certainly put in an assert for `MGLMapboxMetricsEnabled`, but it would add overhead.

include/mbgl/ios/MGLMapboxEvents.h
incanus: Let's leave this public for now.

platform/ios/MGLAccountManager.m
incanus: :+1:

platform/ios/MGLMapboxEvents.m
incanus: Nah, I think we're ok then. We have a tenuous expression of that dependency here, but I think it's fair.

platform/ios/MGLMapboxEvents.m
```diff
@@ -483,22 +565,13 @@ - (void) startTimer {
...
-    }
-    return self.userAgent;
-}
-
-// Can be called from any thread.
-//
-- (NSString *) formatDate:(NSDate *)date {
-    return [self.rfc3339DateFormatter stringFromDate:date];
+- (NSString *) userAgent {
+    return [NSString stringWithFormat:@"%@/%@/%@ %@", self.appName, self.appVersion, self.appBuildNumber, MGLMapboxEventsUserAgent];
```
incanus: Do we lose some sort of fallback ability here for when the values were `nil` in the old code?

incanus:
This feels really solid @1ec5 in code review. Easy to wrap the head around and good improvements made. Doing some testing now.

platform/ios/MGLMapboxEvents.m
1ec5: These properties are now [set unconditionally in `-init`](https://github.com/mapbox/mapbox-gl-native/blob/1ec5-metrics-opt-way-out/platform/ios/MGLMapboxEvents.m#L214-216) based on Info.plist fields that are required for the app to launch. Even if they’re `nil` for some reason, you’d wind up with a user agent string that starts with “(null)/(null)/(null)”. The old code would simply return `nil` and crash when [inserted](https://github.com/mapbox/mapbox-gl-native/blob/b88b917cd46f58b7acf6f97a26a9a7ec6e1a257f/platform/ios/MGLMapboxEvents.m#L445) into the HTTP request.

incanus: :+1:

1ec5:
I folded in #1536, which fixes #1518, because it didn’t make any sense to have the hard assert inside `+[MGLMapboxEvents initialize]`.

incanus:
Plan of attack here per chat with @bleege is to merge this to `master`, cherry pick to `ios-b1`, and follow up with test adaptation to help in ongoing maintenance. We've had enough :eyes: on this today to feel good about it.

Issue #1489, opened 2015-05-09
Rectangular park draws to the south of any annotation on Emerald


1ec5:
In Emerald, every annotation has a rectangular park just to the south that does not go away after panning or even changing to a different style.

![pinpark](https://cloud.githubusercontent.com/assets/1231218/7552221/86d07490-f65d-11e4-867c-5f6a0df2dc80.PNG)

/cc @tatsvc

incanus:
The sprite sizing is wrong: 

![](https://raw.githubusercontent.com/mapbox/mapbox-gl-styles/mb-pages/sprites/emerald%402x.png)

The pin icons should have double-space below them because of how the image center is placed at the point. 

Note the empty space below the pins in Streets: 

![](https://raw.githubusercontent.com/mapbox/mapbox-gl-styles/mb-pages/sprites/mapbox-streets%402x.png)

/cc @peterqliu

tatsvc:
extra context:

![image](https://cloud.githubusercontent.com/assets/2894119/7548001/6599ffbc-f5c7-11e4-88f0-b63ddfe6f29b.png)

1ec5:
@peterqliu fixed this issue upstream in a3a36603d67109cc7494af4601f8cd2fe3825891, but the submodule will need to be updated for that change to take effect.

1ec5:
The saga continues in #1492.

Issue #1488, opened 2015-05-09
Switching styles turns annotations invisible


1ec5:
Changing the style causes any annotations in view to become invisible. Any popover remains visible, and invisible annotations continue to respond to taps.

![pindrop](https://cloud.githubusercontent.com/assets/1231218/7552213/4831fb0a-f65d-11e4-9312-d618cb1bc83f.PNG)

1ec5:
Zooming in and out causes the annotations to appear and disappear – with the old style's sprites.

@bsudekum, this is likely the issue you were seeing with your React Native project.

incanus:
This will go away when we lift annotation imagery of the sprite sheet. This might be an acceptable part of the b1 stopgap.

incanus:
I'm ok with this being a stopgap again in b2, meaning you need to set style and leave it before adding/managing annotations for now. b3's focus thus far is better annotations.

ejdslayer:
Hi, so is there a beta version available that will address this issue for now? I don't mind re-adding the annotations etc, I just need this ready for a project due to launch in 3 weeks. Thanks

kristfal:
Ref #2326, is there any known workarounds for this? 

This issue persists across app launches in ios-v2.1.2 (even closing the app from the task manager), making annotations unusable in any context where a style may be swapped. Let me know if you want any help debugging or testing.

robipresotto:
@incanus map caches the icon annotation anyway if we are setting it before setting the map style. Any suggestion?

maciekish:
cc @zakay

maciekish:
FWIW 3ec610c did not fix the issue.

Zakay:
+1 I'm really hoping this gets fixed asap, we are so close to launch our app with a reach of 1millions users, and this is a major blocking issue.

ljbade:
@1ec5 would this be a case of saving the markers at `setStyleURL` then loading them back after style loaded?

1ec5:
@robipresotto, I think that might be #1628 complicating matters.

1ec5:
@ljbade, not really; you might be able to fashion a workaround at the MGLMapView level by removing the annotations before setting the style and readding all the annotations after it’s done loading, but I think the issue is deeper than MGLMapView.

jfirebaugh:
[...]

PR #1492, opened 2015-05-10
update emerald marker swatch, misc updates to emerald and streets

peterqliu:
fixes https://github.com/mapbox/mapbox-gl-native/issues/1489

also rolls in some minor style updates that we had put off til post-launch, but they seem :ok_hand: while testing them as custom styles. If need be, we can update only the sprite assets, to avoid potential breakage.

cc/ @1ec5 @incanus


1ec5:
So far so good with Emerald. What should I look out for when testing mapbox/mapbox-gl-styles@27e8ee84e58bc1a50342102ede000612ad39e49b?

peterqliu:
@nickidlugash ^

incanus:
> If need be, we can update only the sprite assets, to avoid potential breakage.

We should probably play it conservative and go with that. As for approach, one idea is that we could maintain a branch in `mapbox-gl-styles` that we roll into the shipping GL.

incanus:
We've got offset problems now: 

ios simulator screen shot may 11 2015 9 04 33 am

peterqliu:
@incanus those don't look like the markers in the swatch-- did we switch out?

image

1ec5:
@incanus, maybe you’re running into #1488? That can happen if you use the gear menu to add annotations, then change from Mapbox Streets to Emerald, then cause a redraw by zooming or panning. It wouldn’t be a regression if that’s the case.

1ec5:
@jfirebaugh and I are unable to reproduce the annotation artifacting in either the beta 1 branch or in the same version of the internal test app that I originally spotted this issue in. So it might be dependent on a more specific set of steps.

nickidlugash:
> What should I look out for when testing mapbox/mapbox-gl-styles@27e8ee8?

@1ec5 Sounds like you guys decided not to add this change into the release, but this change just splits up the water_label layer into two layers, filtered by `area`. One layer is styled the same as the original, the other just has a different `min-zoom` and `text-size`, so I imagine it shouldn't break anything.

jfirebaugh:
Moving to https://github.com/mapbox/mapbox-gl-native/pull/1506 instead.

PR #1495, opened 2015-05-10
Mark unused parameters with __unused

1ec5:
This PR replaces haphazard usage of pragmas and `(void)` expressions with the [`__unused` keyword](http://nshipster.com/__attribute__/#unused).

/cc @incanus

### DIFF
--- a/ios/app/MBXViewController.mm
+++ b/ios/app/MBXViewController.mm
@@ -83,10 +83,7 @@ mbgl::Settings_NSUserDefaults *settings = nullptr;
     [self restoreState:nil];
 }
 
-#pragma clang diagnostic push
-#pragma clang diagnostic ignored "-Wunused-parameter"
-
-- (void)saveState:(NSNotification *)notification
+- (void)saveState:(__unused NSNotification *)notification
 {
     if (self.mapView && settings)
     {
@@ -101,7 +98,7 @@ mbgl::Settings_NSUserDefaults *settings = nullptr;
     }
 }
 
-- (void)restoreState:(NSNotification *)notification
+- (void)restoreState:(__unused NSNotification *)notification
 {
     if (self.mapView && settings) {
         settings->load();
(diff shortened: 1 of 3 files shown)


incanus:
Oooh, neat. However, [Clang's docs](http://clang.llvm.org/docs/AttributeReference.html) don't seem to reference it, and it dates from GCC days. Apple's [original docs](https://developer.apple.com/library/mac/releasenotes/Cocoa/RN-ObjectiveC/index.html) back in the 10.5 days prefer: 

```objc
- (void)method:(id) __attribute__((unused)) param;
```

I've also [seen mention](http://stackoverflow.com/a/12047347/977220) of `#pragma unused(param)`, as well there as what we're using now.

1ec5:
There’s a tendency for Clang to support GCC extensions – it works in Clang 3.6 at least. Apple does use `__unused` in [some sample code](https://developer.apple.com/library/mac/samplecode/SandboxedFetch/Listings/fetch_service_fetch_service_c.html#//apple_ref/doc/uid/DTS40011117-fetch_service_fetch_service_c-DontLinkElementID_6) and [plenty of](https://opensource.apple.com/source/objc4/objc4-646/runtime/NSObject.mm) [shipping](https://opensource.apple.com/source/objc4/objc4-646/runtime/objc-auto.mm) [code](https://opensource.apple.com/source/libclosure/libclosure-65/objectTests/assigngc.m), and Xcode’s source code editor recognizes it.

`__attribute__((unused))` is just too verbose to use with already-verbose delegate method signatures. I’d also be fine with either `#pragma unused(param)` ([seen here in a Clang test](https://github.com/llvm-mirror/clang/blob/3e1cca7ad0cc730b54c1a2057f9ce36a85eab75a/test/SemaCXX/pragma-unused.cpp#L7)) or the `void`-casting trick from the good ol’ days (the most portable option, not like it matters in Objective-C). I mainly just dislike the push/pop thing because the syntax is hard to remember and it’s hard to spot the beginning and end of each region fenced off this way.

Issue #352, opened 2014-07-02
Feature eventing (native equivalent of queryRenderedFeatures and querySourceFeatures)


hallahan:
I've been playing with MapboxGL, and I'm trying to figure out how to handle click events to select features. Is this functionality built in?

Looking at the JavaScript in the several blog posts that use mapbox-gl-js, I am able to got the Point object from a click event. Then, I see a function `map.featuresAt`, but I am not able to get it to return any features...

Any clues? Regarding the native libraries, would this be in the Cocoa library?

hallahan:
What is the status for a native `featuresAt` implementation? Is the ability to select a feature on the map going to be a feasible use case for native?

incanus:
It will be eventually, but the focus for now has just been on point annotations, which exist in a runtime style layer and query based on `~50px` square padding for touch gestures. We'll eventually port over `featuresAt` and allow querying of all of the style layers such as OSM base data.

tylrtrmbl:
I am so pumped for `featuresAt`. Thanks all!

hallahan:
Being able to select OSM base data would be an extremely valuable feature. This is something I'd enjoy working on. Could you point me in the right direction?

incanus:
Yes, I would check out the `featuresAt` functionality in `gl-js`. The basic idea is an R-tree that bins the features at parse time so that efficient spatial queries can be made at runtime in response to the "at" and a radius. The structures that exist now aren't indexing parsed features this way. 

Point annotations do index based on tiles made at runtime as annotations are added and removed, which is how gestural touch of points on the map that are rendered in GL was possible.

hallahan:
It looks like you have boost's rtree in there. Are you using this in the same way as rbush in the JS implementation? Are you saying only points are added to a spatial index?

hallahan:
I'm poking around, and the only actual rtree usage I am seeing is in `collision.hpp`. It looks like it's being used to make sure labels don't collide? 

https://github.com/mapbox/mapbox-gl-native/blob/master/src/mbgl/text/collision.hpp#L32

Would we want to use the same container as is being used here for map geometries? I know PostGIS uses `geos`. I've used JTS Topology Suite in Java, the lib geos was ported from. 

http://stackoverflow.com/questions/9891374/is-boost-geometry-mature-enough

What do you guys think? Clearly this is not a trivial feature.

incanus:
[...]

PR #1496, opened 2015-05-10
VoiceOver accessibility

### LINKED ISSUE #1493: Compass, user dot, annotations are inaccessible on iOS
With VoiceOver enabled, the attribution button and annotation callouts can receive focus, but the compass, user dot, and annotation pins cannot. The pins are rendered directly in GL, but there should still be a way to give them accessibility regions. The other views all have accessibility labels.

### LINKED ISSUE #1494: Crash interacting with annotation callouts with VoiceOver enabled — -[NSObject(UIStorage) accessibilityPerformAction:withValue:]
Easily reproducible in iOS 8.3 on an iPhone 6:

1. Enable VoiceOver.
2. Tap an annotation’s callout view to focus on it. (It isn’t possible to interact with the annotation pin, due to #1493.)
3. Tap the callout view four times in short succession.

Usually, the map behind the callout view zooms in, but the focus ring remains in place. Double-tapping at this point triggers a crash:

https://rink.hockeyapp.net/manage/apps/163490/app_versions/22/crash_reasons/34511875

```
Exception Type:  SIGSEGV
Exception Codes: SEGV_ACCERR at 0x10
Crashed Thread:  0

Application Specific Information:
objc_msgSend() selector name: performSelector:withObject:withObject:

Thread 0 Crashed:
0   libobjc.A.dylib                      0x0000000195d3fbd0 objc_msgSend + 16
1   UIKit                                0x0000000188f7d404 -[UIApplication sendAction:to:from:forEvent:] + 92
2   UIKit                                0x0000000188f664e0 -[UIControl _sendActionsForEvents:withEvent:] + 608
3   UIKit                                0x000000018290cad8 0x1828dc000 + 199384
4   UIAccessibility                      0x0000000192290904 -[NSObject(UIStorage) accessibilityPerformAction:withValue:] + 1724
5   UIAccessibility                      0x000000019227237c _performActionCallback + 76
6   AXRuntime                            0x000000018a0a8368 _AXXMIGPerformAction + 96
7   AXRuntime                            0x000000018a0a3668 _XPerformAction + 208
8   AXRuntime                            0x000000018a0af0c4 mshMIGPerform + 268
…
```

The stack has nothing specific to Mapbox GL in it, but maybe the bug is that Mapbox GL isn’t somehow refreshing the callout view’s accessibility info when zooming in causes that view to move.

1ec5:
The changes in this PR are intended to reach ~~parity with MapKit~~ basic user expectations in terms of VoiceOver accessibility.

- [x] Make the compass accessible
- [x] Make the user dot accessible
- [x] Give the :information_source: sensible accessibility text
- [x] Make annotations accessible (fixes #1493)
- [x] Make annotation callouts and their accessory buttons accessible
- [x] Update accessibility when map viewport changes (fixes #1494)
- [x] Order annotation accessibility elements by distance from center / user location
- [x] Bonus: Add accessibility-only controls for navigating the map

By comparison, MapKit has pretty good accessibility, while Google Maps is largely inaccessible on iOS.

There are also architectural improvements in this PR:

* ~~`MGLMapView` now maps annotation IDs to annotation contexts, which contain the annotations themselves (#1504). Previously, annotations were mapped to annotation IDs. The new approach reduces the number of places we need to iterate over the entire mapping and will make it a bit easier to implement another kinds of annotations (#893).~~ This was split out into #3261.


1ec5:
Annotations are now accessible, but it requires more work to find a visible annotation by its unique ID than I’d like. Given how many times `annotationIDsByAnnotation`’s key enumerator is being turned into an array in various parts of `MGLMapView`, I think `annotationIDsByAnnotation` should be inverted to map annotation IDs to dictionaries that contain annotations and their accessibility elements.

incanus:
This is gonna be great.

Once we hit https://github.com/mapbox/mapbox-gl-native/issues/352 at get a `featuresAt()` equivalent, we can start to think about iterating navigation elements on the map as well like Apple Maps does — streets, POIs, etc.

1ec5:
> Once we hit #352 at get a featuresAt() equivalent, we can start to think about iterating navigation elements on the map as well like Apple Maps does — streets, POIs, etc.

To my knowledge, Apple Maps' VoiceOver support when simply viewing the map is limited to POIs. But it'd be awesome if we could use pick out road labels to deliver more comprehensible location readouts at high zoom levels ("near the corner of Broad and High Streets, facing Central Park"), instead of prosaic coordinates and compass directions, without requiring separate calls to a geocoder.

incanus:
> To my knowledge, Apple Maps' VoiceOver support when simply viewing the map is limited to POIs. 

Nah, they do all places visible + roads by default. Here's an example: 

https://dl.dropboxusercontent.com/u/575564/voiceover.mp4

Sorry about the annoying speed I went through it — wasn't hearing the audio when recording so wasn't sure it was working.

1ec5:
> Nah, they do all places visible + roads by default.

Oh cool, although it must be really annoying to have to sift through all the road labels just to get to a POI. Our developers are mostly presenting maps where the annotations matter much more than the base map, so once feature picking gets us granular accessibility elements, we’ll need to make the level of detail configurable.

1ec5:
2e71630574a357be6f9fbd62977158cac366293e most likely fixes #1628.

1ec5:
Mapping annotation IDs to annotation contexts has been split out into #3261. This PR will become a lot smaller once that’s merged and this is rebased.

1ec5:
Due to the large number of interrelated changes in this branch, I rebased it onto master (spanning almost a year of commits) instead of redoing the changes. There were tricky conflicts all along the way, so the current head is probably not as functional as the old head, 9e9ccb72ae6a1d55aac5800e68e2ff8f85747029.

1ec5:
I split out and cleaned up the display string formatting in #4802.

1ec5:
@boundsj, these changes shouldn’t conflict too badly with what you’re working on in #4801, but you’ll want to update some of this logic for annotation views. For example, a view-based annotation’s accessibility frame should be the view’s frame instead of the return value of `-frameOfImage:centeredAtCoordinate:`. A further optimization would be to have the annotation view be the annotation’s accessibility element, but I don’t think that’s strictly necessary.

1ec5:
Filed #4820, #4821, #4822, #4823 to track further accessibility work.

PR #1497, opened 2015-05-10
Attribute Mapbox and OpenStreetMap in-app

1ec5:
With this change, :information_source: pops up an action sheet with [the required attribution](https://www.mapbox.com/help/attribution/) and a direct link to the Map Feedback tool. Eventually we’ll tackle something more along the lines of https://github.com/mapbox/mapbox-gl-native/issues/1265#issuecomment-92964800, but this should be sufficient for beta 1.

/cc @lxbarth @incanus

### DIFF
--- a/platform/ios/MGLMapView.mm
+++ b/platform/ios/MGLMapView.mm
@@ -54,7 +54,7 @@ static NSURL *MGLURLForBundledStyleNamed(NSString *styleName)
 
 #pragma mark - Private -
 
-@interface MGLMapView () <UIGestureRecognizerDelegate, GLKViewDelegate, CLLocationManagerDelegate>
+@interface MGLMapView () <UIGestureRecognizerDelegate, GLKViewDelegate, CLLocationManagerDelegate, UIActionSheetDelegate>
 
 @property (nonatomic) EAGLContext *context;
 @property (nonatomic) GLKView *glView;
@@ -303,7 +303,7 @@ std::chrono::steady_clock::duration secondsAsDuration(float duration)
     //
     _attributionButton = [UIButton buttonWithType:UIButtonTypeInfoLight];
     _attributionButton.accessibilityLabel = @"Attribution info";
-    [_attributionButton addTarget:self action:@selector(showAttribution:) forControlEvents:UIControlEventTouchUpInside];
+    [_attributionButton addTarget:self action:@selector(showAttribution) forControlEvents:UIControlEventTouchUpInside];
     _attributionButton.translatesAutoresizingMaskIntoConstraints = NO;
     [self addSubview:_attributionButton];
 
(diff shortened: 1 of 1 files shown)


friedbunny:
Much improved over the Safari hyperlink. Seeing an action sheet popup from an :information_source: button feels jarring, though, maybe because only one of the three items is immediately obvious as being actionable. But it feels good enough for now, methinks.

Eventually doing something cool with a modal view and real text will be a good opportunity to subtly show off. :smirk:

incanus:
> Seeing an action sheet popup from an :information_source: button feels jarring

Apple does it in Maps, though it's in an app toolbar, not on the map. I think it's fine. 

anigif-1431312253

iPad popover seems to work well, too: 

screen shot 2015-05-10 at 7 58 50 pm

Except for post-rotation: 

anigif-1431313288

Maybe this is ameliorated by iOS 8's [`UIAlertController`](https://developer.apple.com/library/prerelease/ios/documentation/UIKit/Reference/UIAlertController_class/index.html)? Is it iOS 7-safe?

friedbunny:
`UIAlertController` is iOS 8+, not sure if it would fix this.

That Apple Maps alert view is crazy, hah. Segmented buttons, and the data attribution button is different than the actions. I think my perceived weirdness is more to do with the non-actionable action items — it's not immediately clear that "© Open Street Map" is a thing you can do.

But yeah, a little temporary weirdness (that I can't think of a good way to fix/reword, anyway) isn't the end of the world.

1ec5:
Originally, with 965f90c449bc607caa7a34ebf0f762fdfb01417a, I stuck the copyright notices in the alert title and called the first button “[About This Map](https://www.mapbox.com/about/maps/)”. But @lxbarth and I were concerned that this style wouldn’t technically meet Mapbox’s attribution requirements, which requires a hyperlink. Plus, a direct link to OpenStreetMap will be welcome on the OSM talk list, I’m sure. :smiley:

1ec5:
> Except for post-rotation

Indeed, a constraint to the popover has to be broken when the device is rotated:

```
Unable to simultaneously satisfy constraints.
	Probably at least one of the constraints in the following list is one you don't want. Try this: (1) look at each constraint and try to figure out which you don't expect; (2) find the code that added the unwanted constraint or constraints and fix it. (Note: If you're seeing NSAutoresizingMaskLayoutConstraints that you don't understand, refer to the documentation for the UIView property translatesAutoresizingMaskIntoConstraints) 
(
    "<NSLayoutConstraint:0x7aee1510 H:[UIView:0x7ea28ee0(304)]>",
    "<NSLayoutConstraint:0x7ef1ec70 UIView:0x7ef1e970.width == _UIAlertControllerView:0x7ea15640'Mapbox GL for iOS'.width>",
    "<NSAutoresizingMaskLayoutConstraint:0x7ec72640 h=--& v=--& H:[UIView:0x7ef1e970(0)]>",
    "<NSLayoutConstraint:0x7ea2ca00 _UIAlertControllerView:0x7ea15640'Mapbox GL for iOS'.width >= UIView:0x7ea28ee0.width>"
)

Will attempt to recover by breaking constraint 
<NSLayoutConstraint:0x7aee1510 H:[UIView:0x7ea28ee0(304)]>
```

It looks like a UIKit bug that probably wouldn’t affect `UIAlertController`. Even in the original orientation in the iPad Simulator, the popover’s pointer appears slightly off:

pointer

I think it’s because the center of the :information_source: is a bit closer to the right edge than the center of a bottom bar button would be.

friedbunny:
[Here's a quick go at `UIAlertAction`.](https://gist.github.com/friedbunny/02d4a4feccf6ff1d53ca) Using the same `sourceRect` approach still doesn't work with rotation, but I'm sure that can be fixed one way or another.

`UIAlertAction` does do this for iPad popovers, though:

ios simulator screen shot may 10 2015 10 52 40 pm

@incanus I'm not sure what you mean by "safe" — it definitely won't run in iOS 7, but it can be checked against. I think I'm being dense here. :cry:

friedbunny:
Hah, jinx @1ec5.

1ec5:
With 7da7f65579fae412fea0fecd9afc3b56e4577ca6, the popover is dismissed when the device is rotated (and presumably whenever the map resizes), before the constraints can go invalid.

1ec5:
> [Here's a quick go at `UIAlertController`.](https://gist.github.com/friedbunny/02d4a4feccf6ff1d53ca)

This code landed with a few modifications in #8373.

PR #1498, opened 2015-05-11
Add long/lat/zoom to feedback URL

friedbunny:
Carries over the user's map position to the feedback website.

Part of #1497, cc @1ec5.


1ec5:
Good catch – didn’t realize you could leave the style ID blank in the URL hash.

PR #1513, opened 2015-05-12
Set correct strata for background render

### LINKED ISSUE #1478: Roads and rivers are vanishing when panning
Just observed this:

out

jfirebaugh:
Fixes #1478

cc @kkaefer


1ec5:
Looks good. I haven’t seen any disappearing features so far, and Mapbox Streets is a whole lot greener at low zoom levels now.

Issue #1516, opened 2015-05-12
stronger enforcement of dev opt-out setup


incanus:
We should take a stronger stance on making sure devs: 

* Provide users with a way to opt-out, and know that they should. 
* Collect metrics by default. 
* Know about the `Settings.bundle` option. 

In working on metrics testing for #1491, I couldn't figure out why my metrics weren't behaving properly. It turns out my test app (`test/ios/ios-tests.xcodeproj`) didn't implement a `Settings.bundle` and was not collecting any metrics. 

We should probably take a stronger approach, such as: 

* Asserting that a `Settings.bundle` exists at first run instead of just logging about it. 
* Allowing for a class method (e.g. `[MGLMapboxEvents setShowsOptOutInApp:YES]`) which a dev has to implement in their code to work around the above. 

@mapbox/mobiledata @1ec5 @gundersen

friedbunny:
The :information_source: button will be a good place to always offer the opt-out, once that becomes more a custom view.

incanus:
That only gets ahead of metrics collection once a map view has been shown, though.

gundersen:
> stronger stance on making sure devs: 
> provide users with a way to opt-out

Great point @incanus. We should make an awesome page on this, maybe even put a section on the new /mobile. Other mapping apps, especially Apple, have it so buried on how to opt out. We really should push opting out as a feature; something that is great for anyone that wants supper low battery usage or just folks that don't like the idea of anyone collecting their data.

1ec5:
> It turns out my test app (`test/ios/ios-tests.xcodeproj`) didn't implement a `Settings.bundle` and was not collecting any metrics.

I fixed this in e839c2aa21592d9e1c09736a55a4044def9357e3 for #1491. The issue was that `-[NSUserDefaults registerDefaults:]` was being called in `-init` instead of `+initialize`.

bleege:
Added explicit checks for End User Opt Out ability to be implemented by the developer in 48e21ae6e915aecf24022e7ba568d436d876160b.  This can done via 2 ways:

1. Set `[MGLAccountManager setShowsOptOutInApp:YES]` in AppDelegate
1. Check for existence of `Settings.bundle` file via NSAssert that fails program if not satisfied

This code is based off the latest `ios-b1` branch and is available in the `1516-opt-out-checking` branch.

bleege:
[...]

PR #1517, opened 2015-05-12
[not ready] refs #1491: metrics testing

incanus:
Before landing #1491, we should be thinking about better metrics testing. This is a start towards testing various functionality in `MGLMapboxEvents` to: 

1. Get me familiar with it a bit more. 
1. Be in better shape to adapt the tests that break / need new coverage to land #1491. 

So far it tests a couple things we've been leaving uncovered: 

* That setting a token on the map view sets it for metrics as well. 
* That flush (public API) empties the events queue. 
* That pushing events adds to the events queue. 

I am using the popular [OCMock](http://ocmock.org/) to test this functionality. 

Things still to test: 

* That flushing posts certain network requests. 
* Opting out of metrics does what we think it does. 
* The at-run-time changes to opt-out that #1491 hopes to address. 
* etc. 

We can likely use [OHHTTPStubs](https://github.com/AliSoftware/OHHTTPStubs) for this network testing. 

Lastly, I turned up https://github.com/mapbox/mapbox-gl-native/issues/1516 along the way and my last test (`testPushEventAddsToEventQueue`) is currently broken as a result. 

/cc @bleege @1ec5


test/ios/MetricsTests.m
```diff
@@ -0,0 +1,228 @@
...
+
+@property (nonatomic) NSTimeInterval defaultFlushAfter;
+
+@end
+
+@implementation MetricsTests
+
+- (void)beforeAll {
+    [tester acknowledgeSystemAlert];
+    self.defaultFlushAfter = [[MGLMapboxEvents sharedManager] flushAfter];
```
1ec5: #1491 turns `-[MGLMapboxEvents flushAfter]` into a constant, [`MGLFlushInterval`](https://github.com/mapbox/mapbox-gl-native/blob/1ec5-metrics-opt-way-out/platform/ios/MGLMapboxEvents.m#L48) because it wasn’t being modified anywhere. I’m fine with exposing the constant to these tests, but given that modifying the flush interval will no longer be supported, it won’t be necessary to test that case.

incanus: Yeah, I did this just to cut down on testing time. Not a big deal.

1ec5:
Looks good from here.

incanus:
Going to squash & merge these into `master` and then `ios-b1` as they: 

1. Pass locally, except for the last one, which properly tests runtime opt-out. 
1. Only affect `test/ios`; no library code.

PR #1520, opened 2015-05-12
Explicit Opt Out Checking

bleege:
For #1516


jfirebaugh:
@bleege What's up with the commits in this PR? Did you branch off `ios-b1` and then make a pull request against `master`?

bleege:
@jfirebaugh good catch.  this is supposed to be off of `ios-beta1`.  i'll fix this.

incanus:
We should be dev'ing off of `master`, committing critical b1-related changes, and then cherry-picking to `ios-b1`.

bleege:
Reopening as @incanus is correct that we should be going `dev_branch` --> `master` --> `ios-1` with code.

bleege:
@1ec5 suggestions have been implemented, tested locally, and now Travis approved.  Time to merge to `master`.

jfirebaugh:
@bleege You branched off of `ios-b1`, but then you merged to `master`. This had the effect of merging `ios-b1` to master. We were probably going to do that anyway, but I would have rather have done it as an explicit decision rather than as a byproduct of an unrelated change.

bleege:
@jfirebaugh Yep, looks like my troubles branching last night carried over to this morning.  Sorry about that.

bleege:
@1ec5 The getter is now added!  Created #1528 to bring it into master.  I'll let you merge this time!  :-)

incanus:
Love it.

PR #1521, opened 2015-05-12
Explicit Opt Out Checking For iOS B1 Branch

bleege:
For #1516


bleege:
Wrong direction for code flow.  Should be: `dev_branch` --> `master` --> `ios-b1`

PR #1523, opened 2015-05-12
Still image rendering doesn't wait until all tiles are parsed

kkaefer:
In headless tests, I'm observing render failures:

Actual:

image

Expected:

image

### DIFF
--- a/src/mbgl/map/map_context.cpp
+++ b/src/mbgl/map/map_context.cpp
@@ -216,6 +216,11 @@ void MapContext::render() {
     // Cleanup OpenGL objects that we abandoned since the last render call.
     env.performCleanup();
 
+    if (data.mode == MapMode::Still && (!callback || !data.getFullyLoaded())) {
+        // We are either not waiting for a map to be rendered, or we don't have all resources yet.
+        return;
+    }
+
     assert(style);
 
     if (!painter) {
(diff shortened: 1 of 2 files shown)


src/mbgl/map/map_context.cpp
```diff
@@ -216,6 +216,12 @@ void MapContext::render() {
     // Cleanup OpenGL objects that we abandoned since the last render call.
     env.performCleanup();
 
+    if (data.mode == MapMode::Still &&
+        (!callback || !style->isLoaded() || !resourceLoader->getSprite()->isLoaded())) {
```
tmpsantos: I think `resourceLoader->getSprite()->isLoaded()` is probably not needed. `style->isLoaded()` will only be true if all the tiles are in the `parsed` state. If the sprite is not loaded, some tile there will be `partial`.

jfirebaugh:
What is it about these changes that fixes the missing tiles? I see that the short-circuit happens earlier, but the conditional appears to be the same.

kkaefer:
It's not obvious why this fixes it, but here's my reasoning. Before this change, something like this could happen:

- We are rendering a frame, without all required tiles loaded. This takes a while.
- Meanwhile, the worker thread finishes parsing a tile and marks it as parsed
- Then, the map rendering thread finished rendering, and checks whether all tiles are loaded. This is now the case, so it thinks that it can fire the callback even though the render was just partially done, since at the beginning of the rendering step, some tiles were still missing.

Moving the check to the beginning of the render step makes sure that everything is there at the beginning of the final render.

jfirebaugh:
Great explanation, thank you.

PR #1527, opened 2015-05-12
break out setup.sh from run.sh, take two

mikemorris:
`source ./scripts/${FLAVOR}/install.sh` this time to properly export `mason` on `PATH` and fix regression in headless tests.

/cc @jfirebaugh


jfirebaugh:
I see, so the actual problem was that `mason` wasn't on the path at the point the `export LD_LIBRARY_PATH="`mason prefix mesa 10.4.3`/lib:${LD_LIBRARY_PATH:-}"` command was run.

Sourcing all the things sounds like a good approach for now, but I still think an integration between node-mbgl and gl-native where the former depends on a specific configuration of setup scripts in the latter is going to be fragile compared to integrating node-mbgl in the same repo.

mikemorris:
It really only depends on the setup scripts as a result of needing a similar environment (Mesa 10.4.3) to run headless tests. Could easily duplicate some code to add the Mesa setup back to node-mbgl.

PR #1529, opened 2015-05-12
Add action sheet shortcut to Settings.app metrics opt-out

### LINKED ISSUE #1519: Add shortcut to Mapbox Metrics opt-out to ℹ️
Per #1518, we already have the technology to [detect that a user switch for Mapbox Metrics has been added](https://github.com/mapbox/mapbox-gl-native/blob/08f1d4b511ee8193735629214faeabf634617731/platform/ios/MGLMapboxEvents.m#L125-133) to the host app’s Settings bundle. When we see that, we should add an “Adjust Privacy Settings” button to the :information_source: menu that takes the user to the host app’s settings in the Settings app. According to #1075, this should do the trick:

```objc
[[UIApplication sharedApplication] openURL:[NSURL URLWithString:UIApplicationOpenSettingsURLString]];
```

friedbunny:
Dynamically checks if `Settings.bundle` has `MGLMapboxMetricsEnabled` key and only adds button to the :information_source: action sheet if that key exists.

Only applicable to iOS 8+, tested to be safe with iOS 7.

Fixes #1519.

/cc @1ec5

### DIFF
--- a/platform/ios/MGLMapView.mm
+++ b/platform/ios/MGLMapView.mm
@@ -82,6 +82,7 @@ static NSURL *MGLURLForBundledStyleNamed(NSString *styleName)
 @property (nonatomic, getter=isDormant) BOOL dormant;
 @property (nonatomic, getter=isAnimatingGesture) BOOL animatingGesture;
 @property (nonatomic, readonly, getter=isRotationAllowed) BOOL rotationAllowed;
+@property (nonatomic, readonly, getter=hasSettingsBundleMetricsOptOut) BOOL settingsBundleMetricsOptOut;
 
 @end
 
@@ -320,6 +321,12 @@ std::chrono::steady_clock::duration secondsAsDuration(float duration)
                          @"© OpenStreetMap",
                          @"Improve This Map",
                          nil];
+
+    // iOS 8+: add action that opens app's Settings.app panel, if applicable
+    if (&UIApplicationOpenSettingsURLString != NULL && self.hasSettingsBundleMetricsOptOut)
+    {
+        [_attributionSheet addButtonWithTitle:@"Adjust Privacy Settings"];
+    }
 
     // setup compass
     //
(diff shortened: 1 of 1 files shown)


platform/ios/MGLMapView.mm
```diff
@@ -82,6 +82,7 @@ @interface MGLMapView () <UIGestureRecognizerDelegate, GLKViewDelegate, CLLocati
 @property (nonatomic, getter=isDormant) BOOL dormant;
 @property (nonatomic, getter=isAnimatingGesture) BOOL animatingGesture;
 @property (nonatomic, readonly, getter=isRotationAllowed) BOOL rotationAllowed;
+@property (nonatomic, readonly, getter=hasSettingsBundleMetricsOptOut) BOOL settingsBundleMetricsOptOut;
```
1ec5: The property itself should be called `hasSettingsBundleMetricsOptOut`, because it doesn’t hold the opt-out control but only indicates its existence.

friedbunny: :ok_hand:

friedbunny:
ios simulator screen shot may 12 2015 10 50 54 am

1ec5:
Hi @friedbunny, thanks for your patience on this PR. I’ve been busy reworking how Metrics-related preferences are read, particularly in #1553. At this point, I think it may make more sense to just show and hide the menu item based on `+[MGLAccountManager mapboxMetricsEnabledSettingShownInApp]`. (Right at startup, `-[MGLMapboxEvents init]` trawls the Settings bundle plist, just as you’re doing in `-[MGLMapView hasSettingsBundleMetricsOptOut]`, and asserts when the developer hasn’t bothered to implement the switch either in Settings or in the host app.) That’ll simplify this change considerably. Other than that, this change looks great.

friedbunny:
@1ec5 That sounds good, I'll make that change and push to this branch.

You've really been shaking up the metrics prefs lately, I'm going to have to reread the code to see where things have settled. :dash:

friedbunny:
Squashed at #1611. YOLO.

Issue #1139, opened 2015-03-27
port new labeling to -native


ansis:
https://github.com/mapbox/mapbox-gl-js/pull/1079

mikemorris:
Oooh, will be great to get this into native!

ansis:
Large parts of this are ported: https://github.com/mapbox/mapbox-gl-native/tree/new-labelling

It's working (without rotation).

- [ ] redo placement when map rotates
- [ ] optimize getGlyphQuads in -js and then port that
- [ ] reparse overscaled tiles to get better density at high zooms https://github.com/mapbox/mapbox-gl-native/issues/912
- [ ] port tests

Redoing placement when map rotates looks like it'll be the hard part because source stuff and tile parsing is fairly different in -native.

ansis:
now at https://github.com/mapbox/mapbox-gl-native/pull/1666

PR #1532, opened 2015-05-12
Calculate text collision correctly

### LINKED ISSUE #1465: Text collision not calculated properly
After #1326, we render SymbolBuckets as we get glyphs and sprites from the network, but the parser is assuming we have all the buckets already in place when calculating the text collisions (discarding the Collision object after each parsing run).

This is hard to notice on the parts of the map using latin script because you rarely have to wait for glyphs, but can be easily spotted at metropolitan areas of China and Japan (you might see some labels overlapping).

What needs to be done:

* We must keep a reference to the Collision object until the tile is at the state `parsed`.
* The buckets must be added to the Collision object at the same order they appear in the style.
* We need to halt the parser from processing SymbolBuckets at the first empty SymbolBucket we find (so we keep the ordering).

tmpsantos:
Fixes #1465.

### DIFF
--- a/src/mbgl/map/live_tile_data.cpp
+++ b/src/mbgl/map/live_tile_data.cpp
@@ -21,7 +21,7 @@ LiveTileData::LiveTileData(const TileID& id_,
                                      spriteAtlas_, sprite_, source_),
       annotationManager(annotationManager_) {
     // live features are always ready
-    state = State::loaded;
+    setState(State::loaded);
 }
 
 LiveTileData::~LiveTileData() {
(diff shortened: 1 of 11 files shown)


tmpsantos:
Please review.

friedbunny:
@tmpsantos you might also want to have a look at @ansis's [`new-labelling`](https://github.com/mapbox/mapbox-gl-native/compare/new-labelling) branch (#1139), it entirely refactors the collision code.

kkaefer:
Next step is in #1539

PR #1534, opened 2015-05-12
Enable Metrics by default default default

### LINKED ISSUE #1518: MGLMapboxMetrics registers defaults for non-Mapbox settings
At initialization, `MGLMapboxEvents` [automatically registers defaults](https://github.com/mapbox/mapbox-gl-native/blob/08f1d4b511ee8193735629214faeabf634617731/platform/ios/MGLMapboxEvents.m#L125-133) for every setting it finds in the host app’s Settings bundle. That’s too aggressive: Mapbox Metrics is not the only thing that a developer might want to provide settings for. The developer should be responsible for registering any non-Mapbox settings they come up with. Instead of a `for` loop, we should register a default for `MGLMapboxMetricsEnabled` and nothing more.

/cc @bleege @incanus

1ec5:
Enable Mapbox Metrics by default when Settings.bundle is absent, but allow the `DefaultValue` in Settings.bundle to override that behavior. Don’t register defaults for non-Mapbox preferences. And provide more detail in the log message that is printed to the console when Settings.bundle is absent (copied from #1530).

This is a very light-touch approach that avoids some of the complications of adding the hard assertion in #1521.

Fixes #1518.

/cc @incanus @bleege @jfirebaugh

### DIFF
--- a/platform/ios/MGLMapboxEvents.m
+++ b/platform/ios/MGLMapboxEvents.m
@@ -119,23 +119,22 @@ NSString *const MGLEventGestureRotateStart = @"Rotation";
         
         // Put Settings bundle into memory
         NSString *appSettingsBundle = [[NSBundle mainBundle] pathForResource:@"Settings" ofType:@"bundle"];
+        NSMutableDictionary *defaultsToRegister = [NSMutableDictionary dictionaryWithObject:@YES forKey:@"MGLMapboxMetricsEnabled"];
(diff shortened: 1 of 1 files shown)


1ec5:
Based on chat with @incanus and @bleege, we’re OK with shipping the assertion in beta 1, so I’ve repackaged the other fixes in this PR as #1536.

PR #1536, opened 2015-05-12
Enable Metrics by default default default

### LINKED ISSUE #1518: MGLMapboxMetrics registers defaults for non-Mapbox settings

1ec5:
This PR enables Mapbox Metrics by default when Settings.bundle is absent but allows Settings.bundle to override the default. It avoids registering defaults for non-Mapbox preferences. The two-phase registration ensures that the defaults don’t default to unexpected values, yet gives the developer a chance to call `+[MGLAccountManager mapboxMetricsEnabledSettingShownInApp]` before the hard assertion.

This PR fixes #1518 and supersedes #1534. It also supersedes 4837359471bb7c55d685dcd626eceb41cec1de67 from #1491, which failed to take into consideration the following scenarios:

* The host app lacks a Settings.bundle.
* The host app already has a Settings.bundle but it doesn’t contain a Mapbox Metrics switch.

/cc @bleege @incanus

### DIFF
--- a/platform/ios/MGLMapboxEvents.m
+++ b/platform/ios/MGLMapboxEvents.m
@@ -110,6 +110,15 @@ NSString *const MGLEventGestureRotateStart = @"Rotation";
 
 @implementation MGLMapboxEvents
 
++ (void)initialize {
+    if (self == [MGLMapboxEvents class]) {
+        [[NSUserDefaults standardUserDefaults] registerDefaults:@{
+             @"MGLMapboxAccountType": @0,
+             @"MGLMapboxMetricsEnabled": @YES,
+         }];
+    }
+}
+
 // Must be called from the main thread. Only called internally.
 //
 - (instancetype) init {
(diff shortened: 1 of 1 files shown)


1ec5:
These changes have been pulled in #1491. We can reopen this PR if that one fails to make it into beta 1 for some reason.

PR #1537, opened 2015-05-12
compile OS X with --std=c++1y

mikemorris:
because Travis uses Apple LLVM 6.0 and full c++14 support landed in 6.1

https://developer.apple.com/library/ios/documentation/DeveloperTools/Conceptual/WhatsNewXcode/WhatsNewXcode.pdf

/cc @incanus @1ec5

### DIFF
--- a/gyp/common.gypi
+++ b/gyp/common.gypi
@@ -8,11 +8,11 @@
       ['OS=="mac"', {
         'xcode_settings': {
           'CLANG_CXX_LIBRARY': 'libc++',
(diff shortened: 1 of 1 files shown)


incanus:
Tagging @springmeyer @tmpsantos @kkaefer on this as well to have a :eyes:.

mikemorris:
node-mbgl is green again building against this branch https://github.com/mapbox/node-mapbox-gl-native/pull/119

mikemorris:
For reference, where I found the suggestion to use `--std=c++1y` http://stackoverflow.com/a/23018783

PR #1515, opened 2015-05-13
build should fail when headless tests that are expected to pass don't

kkaefer:
Several recent rendering regressions (#1478, #1514, #1523) should have been caught by our headless rendering tests but weren't because we don't fail the build when previously-passing tests start to fail.

We do have some expected failures, but rather than ignoring all test results, we should track the expected failures and ignore them one-by-one.

### DIFF
--- a/scripts/compare_images.sh
+++ b/scripts/compare_images.sh
@@ -3,4 +3,4 @@
 set -e
 set -o pipefail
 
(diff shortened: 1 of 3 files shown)


kkaefer:
This doesn't upload images because of https://github.com/mapbox/mapbox-gl-native/issues/1545

PR #1547, opened 2015-05-13
Return nil for IB designables agent

### LINKED ISSUE #1538: IB Designables Timed Out/Interface Builder Cocoa Touch Tool crashed
I'm getting this non-fatal error when building Treble, [which uses storyboards](https://github.com/friedbunny/treble/blob/619aa80178f98a94de6ff621467137c8b69c75dd/treble/Base.lproj/Main.storyboard):

> `treble/treble/Base.lproj/Main.storyboard`: error: IB Designables: Failed to render instance of `MGLMapView`: Rendering the view took longer than 200 ms. Your drawing code may suffer from slow performance.

Also this:

> `treble/treble/Base.lproj/Main.storyboard`: error: IB Designables: Failed to update auto layout status: Interface Builder Cocoa Touch Tool crashed

The build continues and works as expected on device, but Interface Builder doesn't allow any customization of the `MGLMapView` properties and the preview doesn't display (because it has timed-out/crashed).

This is on the latest `master`, ~`v0.2.24`.

screen shot 2015-05-13 at 12 35 30 am

/cc @1ec5

1ec5:
This PR fixes #1538, a regression from 190e91120927ea2e48a72c8337ba923dabe31f85 for #1491. In Objective-C, the return value for a non-void method is the last receiver by default. Hello NeXTSTEP.

/cc @friedbunny @incanus

### DIFF
--- a/platform/ios/MGLAccountManager.m
+++ b/platform/ios/MGLAccountManager.m
@@ -18,7 +18,7 @@
 //
 + (instancetype) sharedManager {
     if (NSProcessInfo.processInfo.mgl_isInterfaceBuilderDesignablesAgent) {
-        return;
+        return nil;
     }
     static dispatch_once_t onceToken;
     static MGLAccountManager *_sharedManager;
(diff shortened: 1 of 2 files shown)


incanus:
:dizzy_face: :gun: 

:+1:

PR #1539, opened 2015-05-13
Figure out Glyph PBF dependencies in the first parsing step

tmpsantos:
Follow up to https://github.com/mapbox/mapbox-gl-native/issues/1465: https://github.com/mapbox/mapbox-gl-native/commit/3b92725de146eb02ed0ae115b95486207ba63b95 aborts tile parsing for SymbolBuckets, but it also means that we won't be able to figure out that we need a Glyph PBF at the first parsing iteration. This means that it'll take longer until all layers are parsed because we start requesting the required Glyph PBFs later in the process.


tmpsantos:
Before:
```
$ rm cache.sqlite && time ./mbgl-render -s styles/emerald-v7.json -z 16 -y 25.01220 -x 121.46841 -w 1000 -h 1000
real    0m10.036s
user    0m32.160s
sys     0m0.496s
```

After:
```
$ rm cache.sqlite && time ./mbgl-render -s styles/emerald-v7.json -z 16 -y 25.01220 -x 121.46841 -w 1000 -h 1000
real    0m4.757s
user    0m12.448s
sys     0m0.404s
```

PR #1552, opened 2015-05-13
Don't do things that set -e in Travis's shell

### LINKED ISSUE #1545: Don't source shell files at .travis.yml root
Ticketing https://github.com/mapbox/mapbox-gl-native/commit/2b7d2c231908def5bd5a4bad4bf8c4e5e4569a5e#commitcomment-11175610:

We should never `source` files at the Travis root level because this allows `set -*` flags bleed into Travis' wrapper script, which messes with their workflow, e.g. preventing `after_failure` scripts to be triggered.

/cc @mikemorris

jfirebaugh:
Sourcing files that set flags bleed into Travis' wrapper script, which
messes with their workflow, e.g. preventing after_failure scripts from
triggering.

These changes fix that:

* Move exported variables from install.sh to setup.sh, so install.sh
  doesn't have to be sourced.
* Source setup.sh from run.sh, rather than directly from .travis.yml.

In addition, as a good practice, remove `set -e` etc. from setup.sh,
and add a comment not to add them back. It's the sourcing script's
responsibility to set these flags.

Fixes #1545

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -86,10 +86,7 @@ before_install:
 - source ./scripts/travis_helper.sh
 
 install:
-- source ./scripts/${FLAVOR}/install.sh
-
-before_script:
-- if [ -f ./scripts/${FLAVOR}/setup.sh ]; then source ./scripts/${FLAVOR}/setup.sh; fi
+- ./scripts/${FLAVOR}/install.sh
 
 script:
 - ./scripts/${FLAVOR}/run.sh
--- a/scripts/ios/install.sh
+++ b/scripts/ios/install.sh
@@ -5,7 +5,6 @@ set -o pipefail
 
 mapbox_time "checkout_mason" \
 git submodule update --init .mason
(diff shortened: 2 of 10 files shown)


jfirebaugh:
@mikemorris Can you please test that this works with node-mbgl?

mikemorris:
This looks good, I'll test tomorrow but it should work. I only need setup.sh for Linux and OS X builds, so you can drop the iOS setup.sh script and just export Mason in run.sh I think. Android scripts might need to be updated too, I think I changed something there, maybe dropping travis_helper.sh?

jfirebaugh:
> you can drop the iOS setup.sh script and just export Mason in run.sh

Good to know. I think we might as well have all the platforms be consistent though. I should have done Android as well but it already does some stuff differently and I got lazy.

mikemorris:
Green! :+1: https://travis-ci.org/mapbox/node-mapbox-gl-native/builds/62562534

Issue #1404, opened 2015-05-02
Storyboards should recognize when access token is set elsewhere


friedbunny:
When using storyboards to instantiate a map view, it isn't always desirable to set the access token there. We should do the legwork to check if the access token is set elsewhere (perhaps as a global) and present the pretty blue "everything is OK!" screen if it is actually working.

Per chat with @1ec5:

> Not sure if designable code can look at environment variables, but it could be made to read a resource file that wouldn't be checked in. Lots of hoops to go through then.
>
>But this isn't really specific to storyboards. Code works the same way; best to use a dedicated public token I suppose.

friedbunny:
There is also the (edge) case where developers may not be using Mapbox tiles and thus not require a Mapbox access token.

1ec5:
See also #1553: from within a build script, it’s easier to add the access token to the Info.plist (using `plutil -insert`) than to code or a storyboard (where you’d have to do some amount of parsing). But I think Treble demonstrates that the ignored access token file approach makes sense too.

1ec5:
https://github.com/mapbox/mapbox-gl-native/pull/1553#issuecomment-101932877 provides a sample Run Script build phase. I think I’d be comfortable putting that script in a help topic on mapbox.com and linking to it from the installation guide, for developers hesitant to make even their public access tokens public.

friedbunny:
Action on this happening in #1613.

PR #1553, opened 2015-05-13
Grab app-wide data from Info.plist

### LINKED ISSUE #1535: MGLAccountManager should get its data from Info.plist entries, not class setter method calls
`MGLAccountManager` currently forces the developer to send it `-setMapboxMetricsEnabledSettingShownInApp:` (optionally) and `-setAccessToken:` sometime at startup, in that order, before the first `MGLMapView` is initialized. The exact place where these setters must be called depends on how and where they use `MGLMapView` – whether programmatically or in a storyboard, whether in the initial view controller or some later view controller, etc.

iOS has an affordance for settings that need to be set right at launch: the Info.plist that comes with every project. `MGLAccountManager` should look there for the access token and in-app opt-out flag. For beta 2, we should deprecate the class setters and add code to `+[MGLAccountManager initialize]` (which is called as soon as MapboxGL.a loads) that looks inside the Info.plist using `-[NSBundle objectForInfoDictionaryKey:]`.

/cc @incanus @bleege

1ec5:
That way there’s no ambiguity about when you should call things like `+[MGLAccountManager setMapboxMetricsEnabledSettingShownInApp:]`. In fact, that method is now deprecated because it’s so easy to call in the wrong place.

Fixes #1535.

/cc @incanus @bleege

### DIFF
--- a/gyp/platform-ios.gypi
+++ b/gyp/platform-ios.gypi
@@ -27,6 +27,7 @@
         '../platform/ios/MGLFileCache.h',
         '../platform/ios/MGLFileCache.mm',
         '../include/mbgl/ios/MGLAccountManager.h',
+        '../platform/ios/MGLAccountManager_Private.h',
         '../platform/ios/MGLAccountManager.m',
         '../include/mbgl/ios/MGLAnnotation.h',
         '../include/mbgl/ios/MGLUserLocation.h',
--- a/include/mbgl/ios/MGLAccountManager.h
+++ b/include/mbgl/ios/MGLAccountManager.h
@@ -21,7 +21,7 @@
 
 /** Certain Mapbox plans require the collection of user metrics. If you aren't using a preference switch in an existing or new `Settings.bundle` in your application, set this value to `YES` to indicate that you are providing a metrics opt-out for users within your app's interface directly.
 *   @param showsOptOut Whether your application's interface provides a user opt-out preference. The default value is `NO`, meaning a `Settings.bundle` is expected for providing a user opt-out preference. */
-+ (void)setMapboxMetricsEnabledSettingShownInApp:(BOOL)showsOptOut;
++ (void)setMapboxMetricsEnabledSettingShownInApp:(BOOL)showsOptOut DEPRECATED_MSG_ATTRIBUTE("Set MGLMapboxMetricsEnabledSettingShownInApp in Info.plist");
 
 /** Whether in-app user metrics opt-out is configured. If set to the default value of `NO`, a user opt-out preference is expected in a `Settings.bundle` that shows in the application's section within the system Settings app. */
 + (BOOL)mapboxMetricsEnabledSettingShownInApp;
--- a/include/mbgl/ios/MGLAccountManager.h
+++ b/include/mbgl/ios/MGLAccountManager.h
@@ -21,7 +21,7 @@
 
 /** Certain Mapbox plans require the collection of user metrics. If you aren't using a preference switch in an existing or new `Settings.bundle` in your application, set this value to `YES` to indicate that you are providing a metrics opt-out for users within your app's interface directly.
 *   @param showsOptOut Whether your application's interface provides a user opt-out preference. The default value is `NO`, meaning a `Settings.bundle` is expected for providing a user opt-out preference. */
-+ (void)setMapboxMetricsEnabledSettingShownInApp:(BOOL)showsOptOut DEPRECATED_MSG_ATTRIBUTE("Set MGLMapboxMetricsEnabledSettingShownInApp in Info.plist");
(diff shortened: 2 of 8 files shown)


incanus:
Do we need to deprecate, vs. just pulling? We're still in beta.

1ec5:
I want the compiler warning to say what the alternative is rather than "I don't know what you're talking about", but come to think of it, an UNAVAILABLE macro makes more sense, as it's unignorable.

1ec5:
Does it make sense to keep `-[MGLMapView setAccessToken:]`? It’s a mostly redundant code path, since setting `MGLMapboxAccessToken` in Info.plist – either in the property list editor or in the project editor – is by far the preferred way to set the access token. The only use case I can think of for keeping this extra property is different access tokens for different `MGLMapView`s in the same app. But Metrics would be quite confused in that case, sending events tied to different maps based on which one happens to have been initialized last.

If we decide to remove `-[MGLMapView setAccessToken:]`, I’d probably mark it unavailable with a helpful message, as I’ve done for `+[MGLAccountManager setMapboxMetricsEnabledSettingShownInApp:]`, and I’d make the implementation assert with a helpful message for apps built with storyboards. (The assertion message would appear in the Issue navigator as an error while designing in Interface Builder.)

1ec5:
With these changes, if you wanted to upload your project to GitHub, leave out your access token (#1404), and avoid accidentally checking in your access token after you develop locally, you could employ the following Run Script build phase in Xcode (add it anywhere after the Copy Bundle Resources build phase):

```bash
token="$(cat ~/.mapbox)"
if [ "$token" ]; then
  plutil -replace MGLMapboxAccessToken -string $token $TARGET_BUILD_DIR/$INFOPLIST_PATH
fi
```

There’s no need to interact with `MGLAccountManager` at all: the plist entry is read the very moment the framework is loaded.

friedbunny:
I still see using `-[MGLMapView setAccessToken:]` in the app delegate as the primary way developers will want to set access tokens.

If token security is a priority, the plist will be the go-to option, but if it's a lone developer or a private codebase, setting the token in code feels like the easiest path to me. Using the app delegate's `application:didFinishLaunchingWithOptions:` method is also a [super-common and expected pattern in other SDKs](https://www.google.com/search?q=set+api+key+in+didFinishLaunchingWithOptions).

1ec5:
@friedbunny, that would be an argument for keeping `+[MGLAccountManager setAccessToken:]` around. `-[MGLMapView setAccessToken:]` is kinda weird because it affects the app globally via `MGLAccountManager` but looks like a view-local property.

@bsudekum, does React Native let you modify the Info.plist that customarily comes with a project? I’m thinking about obsoleting `-[MGLMapView setAccessToken:]` now that you can add the token to the Info.plist, but if that isn’t an option in your environment, that might be a good reason to keep the method around.

friedbunny:
Ah, bleh on my late night reading comprehension.

bleege:
`-[MGLMapView setAccessToken:]` was originally put in as a way for `+[MGLAccountManager setAccessToken:]` to update it.  It's not likely to be a volatile data object, but I like the idea of `+[MGLAccountManager setAccessToken:]` being the "ultimate source of access token truth at runtime".

1ec5:
`-[MGLMapView setAccessToken:]` has been around forever, but that was before we had any singletons like `MGLAccountManager`. In further defense of the Info.plist approach, note that the installation guide we have planned for beta 1 already has developers fiddle with settings in the project editor, so the UI won’t be foreign to them.

bsudekum:
From a react-native point of view it would not be a deal breaker, but it would be :-1: to have to set the access token in the plist. One of the goals/advantages of react-native is to keep out of xcode.

1ec5:
OK, I’m definitely convinced that `+[MGLAccountManager setAccessToken:]` should stay (and be used by react-native-mapbox-gl, but I’m souring on `-[MGLMapView setAccessToken:]` with each passing day.

To clarify, the intent of this PR is not about keeping the access token private – it just happens to make that easier. Rather, the intent is to replace a tricky setup dance with something that looks more like plain-vanilla, all-UIKit development.

Currently, if you want to implement the Metrics switch inside your app, rather than in Settings, you need to call `+[MGLAccountManager setMapboxMetricsEnabledSettingShownInApp:]` before calling any other `MGLAccountManager` method, initializing any `MGLMapView` with an access token (whether programmatically or in a storyboard), or calling `-[MGLMapView setAccessToken:]`. The number of settings to set this way may even go up over time. It’s fine to expect developers to modify `-[UIApplicationDelegate application:didFinishLaunchingWithOptions:]`, but not to require an order that’s only enforced at runtime, with assertions that can’t really tell you what you did wrong.

1ec5:
Before merging this PR, I’m also going to create an `-[MGLMapView initWithFrame:styleURL:]` and deprecate or remove the initializers that take access tokens. That’ll simply things greatly in `MGLMapView`.

PR #1559, opened 2015-05-14
Less bland default title for user dot annotation

1ec5:
What’s a map without a You Are Here sticker?

/cc @incanus

### DIFF
--- a/platform/ios/MGLUserLocation.m
+++ b/platform/ios/MGLUserLocation.m
@@ -63,7 +63,7 @@
 
 - (NSString *)title
 {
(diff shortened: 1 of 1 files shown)


incanus:
This was designed to match Apple Maps and MapKit-using apps, since it can't easily be changed by devs: 

img_3459
img_3460

1ec5:
Oh hmm, I hadn't really considered that. Couldn't you set `mapView.userAnnotation.title`? Anyhow, I'm fine with reverting if it's a problem.

incanus:
Ah, title is read/write. I think we're fine for now.

PR #1561, opened 2015-05-14
mapID → styleID

### LINKED ISSUE #1500: Rename mapID to styleID
Per discussion with @willwhite, `-[MGLMapView mapID]` should be renamed `-styleID` for consistency with `-styleURL` and because the mapbox: URL scheme expands to a style API URL rather than one from the TileJSON API.

We'll need to deprecate `-mapID`, update documentation, and issue a warning from within designables if possible.

1ec5:
Renamed `-mapID` and `-setMapID:` to `-styleID` and `-setStyleID:`, respectively, to accurately reflect the APIs the ID is used for. Marked the property unavailable but reimplemented its getter and setter to assert with helpful messages. Even if no one has reason to use `mapID` in beta 1, a developer could have tried to put a raster map ID in the storyboard inspectable; if we remove the old property entirely, there’s no longer any UI to recover from that situation. The assertion I’ve implemented is the best way to communicate the rename in that case.

Fixes #1500.

/cc @willwhite @incanus

### DIFF
--- a/include/mbgl/ios/MGLMapView+IBAdditions.h
+++ b/include/mbgl/ios/MGLMapView+IBAdditions.h
@@ -9,7 +9,7 @@
 // MGLMapView+IBAdditions.h, due to ASCII sort order.
 
 @property (nonatomic) IBInspectable NSString *accessToken;
-@property (nonatomic) IBInspectable NSString *mapID;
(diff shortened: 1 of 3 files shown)


incanus:
@1ec5  We need to ticket out updates to `/guides` and `/mapbox-gl-ios` to accompany this.

PR #1562, opened 2015-05-14
Run static analyzer on iOS targets on Travis

### LINKED ISSUE #1549: Run Xcode static analyzer on CircleCI for iOS builds
#1538 would’ve been caught like crazy by Xcode’s static analyzer. We should run the static analyzer on Travis since we apparently aren’t great about running it manually ourselves (#1425). From what I’ve seen, the analyzer doesn’t take all that long on our entire project. Note that we’ll need to whitelist some warnings in upstream libraries, which are tracked in memononen/libtess2#19 and [angusj/polyclipping#134](https://sourceforge.net/p/polyclipping/bugs/134/).

/cc @incanus

1ec5:
This PR adds a make target for performing static analysis everything that goes into an iOS build, as well as the demo app and benchmark app. It fails on any warning but also excludes two files from third-party libraries with known warnings.

Fixes #1549.

/cc @jfirebaugh @incanus @springmeyer

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -143,7 +143,7 @@ ipackage-sim: clean Xcode/ios
 itest: ipackage-sim KIF
 	./scripts/ios/test.sh
 
-ianalyze: clean Xcode/ios
+ianalyze: Xcode/ios
 	./scripts/ios/analyze.sh
 
 # Legacy name
--- a/Makefile
+++ b/Makefile
@@ -143,6 +143,9 @@ ipackage-sim: clean Xcode/ios
 itest: ipackage-sim KIF
 	./scripts/ios/test.sh
 
+ianalyze: Xcode/ios
+	./scripts/ios/analyze.sh
+
 # Legacy name
 iproj: ios-proj
 
--- a/Makefile
+++ b/Makefile
@@ -143,7 +143,7 @@ ipackage-sim: clean Xcode/ios
 itest: ipackage-sim KIF
 	./scripts/ios/test.sh
 
-ianalyze: Xcode/ios
+ianalyze:
 	./scripts/ios/analyze.sh
 
 # Legacy name
--- a/Makefile
+++ b/Makefile
@@ -143,7 +143,7 @@ ipackage-sim: clean Xcode/ios
 itest: ipackage-sim KIF
 	./scripts/ios/test.sh
 
-ianalyze: Xcode/ios
+ianalyze: clean Xcode/ios
 	./scripts/ios/analyze.sh
 
 # Legacy name
--- a/scripts/ios/analyze.sh
+++ b/scripts/ios/analyze.sh
@@ -15,4 +15,6 @@ xcodebuild \
     -project ./build/ios/gyp/ios.xcodeproj \
     -configuration Debug \
     -target All \
-    analyze | xcpretty -c | grep -qF 'The following commands produced analyzer issues'
+    analyze | xcpretty | tee ./build/ios/analyze.txt
+
+! grep -q 'The following commands produced analyzer issues' ./build/ios/analyze.txt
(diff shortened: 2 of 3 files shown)


1ec5:
What do you think about how I’ve opted these files out of the analyzer, versus parsing the output for individual warnings? Unfortunately Clang offers [no way to turn off specific checks](http://clang-analyzer.llvm.org/faq.html#suppress_issue), like the dead store check.

jfirebaugh:
Is it possible to run this on the osx line in the build matrix and get equivalent coverage? The ios build takes roughly twice as long as the osx build and it would be good to try to balance them out a little.

jfirebaugh:
> What do you think about how I’ve opted these files out of the analyzer, versus parsing the output for individual warnings?

Seems fine to me. These exceptions will drop out when we replace clipper and libtess with earcut.

1ec5:
> Is it possible to run this on the osx line in the build matrix and get equivalent coverage?

Coverage of the cross-platform code would be nearly identical, but we’d lose analysis of iosapp and ios-bench. It was a bug in iosapp that spurred #1549 in the first place.

Makefile
```diff
@@ -143,6 +143,9 @@ ipackage-sim: clean Xcode/ios
 itest: ipackage-sim KIF
 	./scripts/ios/test.sh
 
+ianalyze: clean Xcode/ios
```
jfirebaugh: Remove `clean` from the dependencies, this is adding several minutes to the build time.

I think we should remove it from `ipackage` and `ipackage-sim` too.

1ec5: > I think we should remove it from `ipackage` and `ipackage-sim` too.

Would removing the clean step from these targets have any impact on the release process we have now?

1ec5: #1584

1ec5: 43bc922a754e91fde86911289417f294e851ad9e removes `clean` from `ianalyze`.

jfirebaugh:
Grrr, last build took even longer (11.5 minutes). Not sure if it's Travis variability or what.

jfirebaugh:
Are the [recompiles](https://travis-ci.org/mapbox/mapbox-gl-native/jobs/63055262#L520) in the `make ianalyze` step necessary? These files are compiled already in `make ipackage-sim`.

1ec5:
I think it has to do with regenerating the project. I’ll try removing `Xcode/ios` from there too.

incanus:
https://github.com/mapbox/mapbox-gl-native/commit/52dd826e2b8a6b3c46c8c1bf248bef825b357303 and thus `make ianalyze` should continue to work as long as they are run after something like `make ipackage-sim`, because otherwise there won't be a project file. But I don't think `make ianalyze` won't run by itself in a clean environment.

jfirebaugh:
Still at 225s for `make analyze` on Travis, i.e. doubling the run time. That's a pretty high price. :fearful:

incanus:
Crazy idea — what about a random chance of running, like 1 in 5 or 1 in 10 commits?

1ec5:
I'm :-1: on Russian roulette on Travis. Makes it hard to tell what's at fault for a failure. Xcode does have single file analysis in the GUI; not sure if we can analyze more selectively. @springmeyer, what's your take on this, since you were hoping for even better coverage?

springmeyer:
> @springmeyer, what's your take on this, since you were hoping for even better coverage?

If the penalty in per-commit time is too great then what about only running it if a certain commit message is present like `git commit -m "[static-anal]"`. We could document workflows where that should be run at a minimum before pull requests are merged and before tagging releases.

1ec5:
This PR is no longer applicable, given that we moved iOS builds from Travis to Bitrise in #2898 and even moved OS X builds to Bitrise as well in #3441.

PR #1578, opened 2015-05-17
Audit headers for nullability

### LINKED ISSUE #1054: Audit headers for nullability (Swift support)
Once we distribute a framework (#828), developers will be able to interface with `MGLMapView` and other classes in Swift. We’ll want to audit the headers for nullability so that developers don’t have to deal with a litany of implicitly unwrapped optionals like `NSString!`.

More information:
* https://developer.apple.com/swift/blog/?id=25
* http://nshipster.com/swift-1.2/#nullability-annotations

1ec5:
All the headers have been audited for nullability. This is a breaking change in the sense that Swift code written against Mapbox GL will likely fail to compile. However, Xcode will offer fix-its for every error that comes up, and the resulting code will be much less stressful (less `shouting!`) and more idiomatic.

Fixes #1054.

/cc @incanus

### DIFF
--- a/include/mbgl/ios/MGLAccountManager.h
+++ b/include/mbgl/ios/MGLAccountManager.h
@@ -1,4 +1,6 @@
-#import <Foundation/Foundation.h>
+#import "MGLTypes.h"
+
+NS_ASSUME_NONNULL_BEGIN
 
 /** The MGLAccountManager object provides a global way to set a Mapbox API access token, as well as other settings used framework-wide. */
 @interface MGLAccountManager : NSObject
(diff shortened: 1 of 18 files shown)


1ec5:
@incanus @bleege @friedbunny, just a heads-up that this PR [brackets `@interface`s in `NS_ASSUME_NONNULL_BEGIN`/`NS_ASSUME_NONNULL_END`](https://github.com/mapbox/mapbox-gl-native/pull/1578/files#diff-7d5de345f22b888dcd00f7b22431cc65R3). So once this PR is merged, if you add any public methods that receive or return a pointer type, remember to add the `nullable` annotation if the value may be `nil` for any reason.

1ec5:
This needs to be on beta 2. Apple finally audited MapKit for nullability with Xcode 7 and the iOS 9 SDK, and we need to be compatible.

1ec5:
I get several nullability-related warnings in Xcode 7 with the iOS 9 SDK, due to newly-audited APIs in UIKit.

1ec5:
Looks like I’ll need to also bring over much of 54e2bb80d9a11cbe7d70b302c9b8c8483cd9f1e1 to satisfy a warning about invoking a non-designated initializer in `MGLUserLocationAnnotationView`.

incanus:
> Apple finally audited MapKit for nullability with Xcode 7 and the iOS 9 SDK, and we need to be compatible.

Is this mostly from a "so it still looks like Apple" POV? Or something more drastic as to why we should hit this on the sooner side?

1ec5:
Mostly from a “so it still looks like Apple” POV. But also so we can say we also audited Mapbox GL for nullability, just as Apple announced this week, and because this work has no stability implications.

1ec5:
This PR also blocks mapbox/blog#518.

jfirebaugh:
@1ec5 If you are happy with this please rebase and merge it. :ship:

Issue #1431, opened 2015-05-05
fix start/finish rendering map delegate callbacks


incanus:
Per https://github.com/mapbox/mapbox-gl-native/commit/fab65e75d0b13a5d03eaea151bb63f2fc6b76bd0#commitcomment-11046360, the `MapChangeWillStartRenderingMap` and `MapChangeDidFinishRenderingMap` notification types are [meant to drive delegate callbacks](https://github.com/mapbox/mapbox-gl-native/blob/fd099593dc0a8fe8cb156fa22bf3739a7c325036/platform/ios/MGLMapView.mm#L2311-L2334) which have parity with Apple's MapKit. These callbacks are meant to relate to the visual appearance of render, not anything frame-based, so that common iOS post-view-load techniques can be employed. 

* Capturing a snapshot of a fully-rendered complex view for use with Core Image/blur filters for view transition effects. 
* Not presenting the map view until it's completely tiled in (we've had requests for this before from partners for raster use cases). 
* Determining when an incomplete map view might be shown, but no more attempts to load anything will be happening (`fullyRendered: NO`). 

Some [MapKit background](https://developer.apple.com/library/ios/documentation/MapKit/Reference/MKMapViewDelegate_Protocol/index.html#//apple_ref/occ/intfm/MKMapViewDelegate/mapViewDidFinishRenderingMap:fullyRendered): 

> The map view calls this method when one or more tiles are revealed and require rendering.

[...]

> This method lets you know when the map view finishes rendering all of the currently visible tiles to the best of its ability. This method is called regardless of whether all tiles were rendered successfully. If there were errors loading one or more tiles that prevented map view from rendering them, the `fullyRendered` parameter is set to `NO`.

Calling these sorts of methods per-frame is not useful to users, either, because any work done in there will impact the render frame rate. 

If we need additional callbacks such as something like `MapChangeDidRenderFrame`, we should make and use them instead, optionally also exporting them to Cocoa. 

/cc @kkaefer

incanus:
Per chat we are moving this to b2 and not updating anything else around delegates — known bug that they no longer work like MapKit.

incanus:
This is a regression currently from b1 to b2.

friedbunny:
Seeing `isChanging` still firing after `didChange`:

```objc
2015-06-23 17:02:45.333 Mapbox GL[46110:4728517] will
2015-06-23 17:02:45.335 Mapbox GL[46110:4728517] is
[...]

PR #1582, opened 2015-05-18
Animate the user dot

### LINKED ISSUE #1041: User location tracking is rough
User location tracking for is functional #756 but rough compared to MapKit’s implementation. In MapKit, the user dot crawls continuously and the map pans just as smoothly. Our implementation jerks forward every time the GPS location changes. Since the panning motion is animated, the map always trails the user dot. Instead, perhaps it could lead the user dot, by overshooting it based on its _n_ previous coordinates.

1ec5:
Previously, the user dot would blit across the screen, and `MGLMapView` would try to keep up but would tragically remain one step behind. With this PR, the user and `MGLMapView` waltz happily together.

- [x] Animate the user dot using UIView block-based animation
- [x] Make the user annotation callout view track the user dot
- [x] ~~Make the viewport track the user dot more smoothly~~ (split out into #3589)
- [x] ~~Keep the user dot from drifting when panning the map view with gestures~~ (more or less addressed by #3683)

Depends on #3589. Fixes #1041. Ref #1125.

### DIFF
--- a/platform/ios/CHANGELOG.md
+++ b/platform/ios/CHANGELOG.md
@@ -6,6 +6,7 @@ Mapbox welcomes participation and contributions from everyone.  If you’d like 
 
 - Applications linking against the SDK static framework no longer need to add `-ObjC` to the Other Linker Flags (`OTHER_LDFLAGS`) build setting. If you previously added this flag solely for this SDK, removing the flag may potentially reduce the overall size of your application. ([#4641](https://github.com/mapbox/mapbox-gl-native/pull/4641))
 - Removed the `armv7s` slice from the SDK to reduce its size. iPhone 5 and iPhone 5c automatically use the `armv7` slice instead. ([#4641](https://github.com/mapbox/mapbox-gl-native/pull/4641))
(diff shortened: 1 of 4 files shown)


1ec5:
I’d like to make user tracking smoother by sticking the user dot in the center of the screen and moving the viewport instead of the user dot. Unfortunately, `-notifyMapChange:` is called way too often, even from the C++ level, preempting the animation by atomically moving the user dot to the final position.

1ec5:
Another problem is that `Transform::setLatLngZoom()` hard-codes an ease-in/ease-out animation, whereas a linear animation is called for when animating user tracking.

incanus:
> Unfortunately, `-notifyMapChange:` is called way too often

https://github.com/mapbox/mapbox-gl-native/issues/1431

> Calling these sorts of methods per-frame is not useful to users

/cc @kkaefer

1ec5:
#2204 complicates things here too.

1ec5:
I split the user tracking case out into #3589.

1ec5:
I rebased this branch atop #3589. It looks pretty good, other than the callout’s minor misplacement when selecting a user dot that’s in motion.

PR #1587, opened 2015-05-18
Add README instructions for running on Android emulator and devices

### LINKED ISSUE #1570: complete instructions for Android on OSX
The current README is good as far as it goes for getting up and running with Android on OS X, but it leaves you hanging after the "launch Android Studio" step -- I just spent an hour trying to get the test app running in the simulator to take a look #1468 and wasn't successful.

It looks like we need to document:

* Setting up a virtual device that's compatible with what we build -- seems like the default x86 AVD does not work.
* How to launch the AVD and test application.

Ideally, this would be done by someone who knows the ins and outs of Android Studio starting from a clean install and writing down every step along the way.

ljbade:
Fixes #1570 

@jfirebaugh Can you read to see if these instructions are good? It should be platform agnostic

### DIFF
--- a/README.md
+++ b/README.md
@@ -49,7 +49,7 @@ Target OS X: 10.9+
 
 #### CocoaPods
 
-[CocoaPods](https://cocoapods.org) is preferred. Put the following in your `Podfile`: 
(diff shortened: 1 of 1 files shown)


jfirebaugh:
This will be a big help, thanks @ljbade.

PR #1588, opened 2015-05-18
Notify renderStill() of request failures

### LINKED ISSUE #1565: MapMode::Still will never complete if requests fail
We are currently not handling the case when tiles, glyphs or sprites fail to load.

* In case of a glyph or sprite failure, the tile will be in a `partial` state forever and specially when rendering `MapMode::Still`, it will hang there indefinitely because the exit condition is all the tiles being on the `parsed` state.
* In case of a tile fails to load, it will go back to the `invalid` state which will cause a re-request of the tile. If the tile is corrupt on the server or if the server is down, etc it will cause an endless cycle of re-requests.

Probably the cause of: https://github.com/mapbox/node-mapbox-gl-native/issues/114

/cc @mikemorris

tmpsantos:
Propagates the failure all the way from the object making the requests to the failureCallback that has to be passed to renderStill() in addition to the usual successCallback.

The branch also adds unit tests simulating request failures on all resources we request.

Fixes #1565.


jfirebaugh:
Definitely agree with the approach here. I started on something similar in 667dc1f1c6c87fc0a0d064da06a53ac953f1b9a6, except I was going to change the signature of the existing callback to accept a `std::exception_ptr` as the first argument. In non-error cases it would be null. This would be a node-like callback convention, and I think I prefer it -- one less argument to thread through the code, and passing an error object of some sort will be more informative than a boolean "succeeded / failed" status.

mikemorris:
Alright, had a chance to test this out and while it does target the hanging tests seen in https://github.com/mapbox/tilelive-gl/issues/25, I'm not sure if the solution is ideal.

- If a render call requests multiple vector tiles and a single request fails, no image is returned. The render call would previously still return an image if some tile requests failed. In the case of the `1-0-1` test case, the three spurious tile requests are for subpixel border tiles that are not visible on the resulting image. I'm not sure what the desired behavior here should be

```
# 1-0-1
ok 19 null
{ class: 'HttpRequest',
  severity: 'ERROR',
  text: '[./fixtures/tiles/1-1-1.vector.pbf] tile loading failed: Error: ENOENT, open \'/Users/mikemorris/gyp/tilelive-gl/test/fixtures/tiles/1-1-1.vector.pbf\'',
  environment: 7,
  threadName: 'Map' }
{ class: 'HttpRequest',
  severity: 'ERROR',
  text: '[./fixtures/tiles/1-1-0.vector.pbf] tile loading failed: Error: ENOENT, open \'/Users/mikemorris/gyp/tilelive-gl/test/fixtures/tiles/1-1-0.vector.pbf\'',
  environment: 7,
  threadName: 'Map' }
not ok 20 Error: Didn't get an image
```

- It would be more helpful to pass an error argument (or `null`) to the single callback as @jfirebaugh suggests, rather than adding a separate [`failureCallback`](https://github.com/mapbox/node-mapbox-gl-native/compare/failed-requests).

- Without returning or throwing an exception, node-mbgl can only [infer that that render failed by the lack of an image](https://github.com/mapbox/node-mapbox-gl-native/blob/master/src/node_map.cpp#L265-L270), leading to an opaque error.

- This does not address skipped style loading tests in https://github.com/mapbox/node-mapbox-gl-native/issues/114 or corrupt vector tile data in the skipped test at https://github.com/mapbox/node-mapbox-gl-native/blob/master/test/js/gzip.test.js#L102-L150 (the callback is never called).

tmpsantos:
@jfirebaugh @mikemorris thanks for the feedback. I rewrote the patch using `std::exception_ptr` and it indeed is looking much better. It also made the unit tests better as a side effect.

mikemorris:
:+1: to merge this

src/mbgl/map/tile_data.cpp
```diff
@@ -34,15 +36,20 @@ void TileData::setState(const State& state_) {
     state = state_;
 }
 
-void TileData::request(Worker& worker, float pixelRatio, std::function<void()> callback) {
+void TileData::request(Worker& worker,
+                       float pixelRatio,
+                       const std::function<void()>& successCallback,
+                       const std::function<void(const std::string& message)>& failureCallback) {
```
jfirebaugh: Should we use the single callback convention here too?

tmpsantos: > Should we use the single callback convention here too?

Yes, I will refactor this one in another branch. I've to find a nice way of replacing this one because the Worker expects a std::function<void()> as callback and I want to report errors during the tile parsing.

jfirebaugh:
:+1:

mikemorris:
Testing changes for  https://github.com/mapbox/node-mapbox-gl-native/issues/122, I noticed that when a `std::exception` is returned with a `map.renderStill` call, the `cancel` method on the `FileSource` implementation may not get called. I suspect this could be preventing a [`std::unique_ptr::reset`](http://www.cplusplus.com/reference/memory/unique_ptr/reset/) call from completing, as the underlying `mbgl::Map` object may still be in use.

Replicated this with Node.js v0.12.0 on commit https://github.com/mapbox/node-mapbox-gl-native/commit/8d35b678ede00e51b12dd3fed95d92eed7b08ba6 by running `node test/js/map.test.js`

mikemorris:
^ Scratch that, rebuilt and not replicating this any more...

tmpsantos:
> Testing changes for mapbox/node-mapbox-gl-native#122, I noticed that when a std::exception is returned with a map.renderStill call, the cancel method on the FileSource implementation may not get called. I suspect this could be preventing a std::unique_ptr::reset call from completing, as the underlying mbgl::Map object may still be in use (this is pure conjecture and may not actually be how C++ works but something it preventing code from continuing to run at this point).

That is correct. I managed to isolate the problem in a unit test and I'm cooking a fix for this. I don't think we should call `cancel()` on the FileSource on failure, but we should cancel all pending requests of a Map object at its dtor. The Map object in theory could be destroyed at any moment and despite the fact that we join all the threads and etc before cleaning the object, it should be done quickly (and it is not the case now, it seems).

I will elaborate the problem better in a new issue and will cc you.

@mikemorris thanks for reporting!

PR #1540, opened 2015-05-19
Headless label placement tests aren't deterministic

kkaefer:
Here are few different renders of `text-color/default`:

test

/cc @ansis any idea how to debug this?


kkaefer:
What happened was that the Glyph PBF was marked as parsed before all of the glyphs were inserted into the FontStack. That means some labels (always at the beginning) were missing some letters because they couldn't be found when calculating the shaping.

PR #1590, opened 2015-05-19
port sprite atlas pollution fix

### LINKED ISSUE #1398: port sprite leaking fix
port https://github.com/mapbox/mapbox-gl-js/pull/1195

Fixes this test issue:

out

Diff:

image

ansis:
ports https://github.com/mapbox/mapbox-gl-js/pull/1195 to fix #1398

### DIFF
--- a/src/mbgl/geometry/glyph_atlas.cpp
+++ b/src/mbgl/geometry/glyph_atlas.cpp
@@ -69,8 +69,9 @@ Rect<uint16_t> GlyphAtlas::addGlyph(uintptr_t tileUID,
     uint16_t buffered_height = glyph.metrics.height + buffer * 2;
 
     // Add a 1px border around every image.
-    uint16_t pack_width = buffered_width;
-    uint16_t pack_height = buffered_height;
+    const uint16_t padding = 1;
+    uint16_t pack_width = buffered_width + 2 * padding;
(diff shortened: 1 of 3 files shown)


ansis:
@kkaefer want to take a quick look? this should be ready to merge

kkaefer:
Will review and adjust and re-enable test cases.

jfirebaugh:
This improves some test cases but degrades others. In particular, some of the icons appear to be shifted one full pixel to the right compared to the reference rendering. See `icon-size/default` for example:

http://mapbox.s3.amazonaws.com/mapbox-gl-native/tests/4313.1/index.html

ansis:
> some of the icons appear to be shifted one full pixel to the right compared to the reference rendering

I can't reproduce this locally. Can you?

The changes in this pr should affect all instances of an icon the same way so it's surprising that only some instances of the same icon are shifted. The only thing I can think of is that since the extrudes changed some precision issues caused the icons to snap (nearest neighbour filtering) in a different direction. This would be surprising though because it would be an unrelated issue that we haven't seen before.

kkaefer:
I couldn't reproduce the regressed tests either.  @jfirebaugh, do you have an image?

jfirebaugh:
I made some test suite adjustments so that things pass again. As always, "it's probably multiple issues".

Issue #1596, opened 2015-05-19
Add "mapbox.com" Checking For "events=true"


bleege:
Currently all Tile source providers get checked for ["events=true" query string appending](https://github.com/mapbox/mapbox-gl-native/blob/58c9e2687de5befc69cb50321d303d7fa704cbfa/platform/darwin/http_request_nsurl.mm#L157-L164).  This query string variable is only used / need by internal Mapbox systems and therefore is extraneous to other tile providers.  Let's a check to iOS GL's `http_request_nsurl.mm` to only do this for `mapbox.com` requests.  This will get this support out faster until we can work this into the Core GL stack in a future ticket.

@mapbox/mobile @mapbox/mobiledata @jfirebaugh

bleege:
Tested, Travised, and Merged.

PR #1599, opened 2015-05-19
Adding mapbox.com checking to http requests

bleege:
For #1596

### DIFF
--- a/platform/darwin/http_request_nsurl.mm
+++ b/platform/darwin/http_request_nsurl.mm
@@ -154,13 +154,15 @@ void HTTPRequest::start() {
     @autoreleasepool {
         
         NSMutableString *url = [NSMutableString stringWithString:@(resource.url.c_str())];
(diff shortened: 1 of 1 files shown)


platform/darwin/http_request_nsurl.mm
```diff
@@ -154,13 +154,15 @@
     @autoreleasepool {
         
         NSMutableString *url = [NSMutableString stringWithString:@(resource.url.c_str())];
-        if ([[NSUserDefaults standardUserDefaults] integerForKey:@"MGLMapboxAccountType"] == 0) {
-            if ([url rangeOfString:@"?"].location == NSNotFound) {
-                [url appendString:@"?"];
-            } else {
-                [url appendString:@"&"];
+        if ([url rangeOfString:@"mapbox.com"].location != NSNotFound) {
```
1ec5: What if “mapbox.com” appears elsewhere in the URL besides the hostname? `-[NSURL host]` is much better suited for this kind of comparison. If that turns out to be unperformant for some reason, there’s also `CFURLCopyHostName()`.

PR #1601, opened 2015-05-19
Compare hostname for adding events=true

1ec5:
If you happen to have “mapbox.com” in the URL but not in the hostname, or if your non-Mapbox server is called “mapbox.commonexamplecompany.com”, we shouldn’t append “events=true” either.

ref #1596, #1599  
cc @bleege

### DIFF
--- a/platform/darwin/http_request_nsurl.mm
+++ b/platform/darwin/http_request_nsurl.mm
@@ -91,6 +91,7 @@ public:
 
     NSURLSession *session = nil;
     NSString *userAgent = nil;
+    NSInteger accountType = 0;
 };
 
 HTTPNSURLContext::HTTPNSURLContext(uv_loop_t *loop_) : HTTPContext(loop_) {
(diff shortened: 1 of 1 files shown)


bleege:
:+1:

1ec5:
As of 90b70180c53c2945c10b9520f9e6aa5f4f697ac0:

```
Running Time	Self (ms)		Symbol Name
33.0ms    0.1%	0,0	 	                         mbgl::HTTPRequest::start()
29.0ms    0.1%	0,0	 	                          _CFPreferencesGetAppIntegerValueWithContainer
3.0ms    0.0%	0,0	 	                          -[__NSURLSessionLocal dataTaskForRequest:completion:]
1.0ms    0.0%	1,0	 	                          -[NSString rangeOfString:]
```

I’m going to cache the user default to avoid incurring `NSUserDefaults` performance hits on every request.

Issue #1595, opened 2015-05-19
Record IDFV For Internal Billing


bleege:
In order to Mapbox to monitor usage for billing purposes, it needs to be able to count the number of Unique Users per customer.  This will be done using the [iOS's IDFV](https://developer.apple.com/library/ios/documentation/UIKit/Reference/UIDevice_Class/#//apple_ref/occ/instp/UIDevice/identifierForVendor).

One `MapboxEndUser` event will be sent to the Mapbox Events API from the `MGLAccountManager` constructor.  This is possible because all implementing apps will need to / eventually will call directly or indirectly (ex: `-[MGLMapView setAccessToken]`) at some point:

* `+[MGLAccountManager setAccessToken:]`
* `+[MGLAccountManager setMapboxMetricsEnabledSettingShownInApp:]`

@mapbox/mobile @mapbox/mobiledata @pdgoodman

1ec5:
> One `MapboxEndUser` event will be sent to the Mapbox Events API from the `MGLAccountManager` constructor.

Will it still go through `MGLMapboxEvents`, or will `MGLAccountManager` now be responsible for sending HTTP requests too?

> all implementing apps will need to / eventually will call directly or indirectly (ex: `-[MGLMapView setAccessToken]`) at some point:

With #1553, `MGLAccountManager` and `MGLMapboxEvents` will have code that runs immediately at startup, as long as the account token is set in the Info.plist (which is the route I’d expect most developers to take), but in `+load` and `+initialize`, which are expected to do noninvasive tasks only.

bleege:
> Will it still go through MGLMapboxEvents, or will MGLAccountManager now be responsible for sending HTTP requests too?

I'm torn, but I'm leaning towards `MGLAccountManager` as it keeps this completely separate from Metrics and the benefits of not instantiating it if not needed.  It's unclear if we'll have to do URL checking on this re #1396.

As for implementation, I'm thinking that an Async HTTP GET could be fired off from [`+MGLAcountManager sharedManager`](https://github.com/mapbox/mapbox-gl-native/blob/f74fe69cb85d544e84d6a7ecab1d742ffdf87828/platform/ios/MGLAccountManager.m#L26-L29) with just the event name and the IDFV value.

@1ec5 What do you think ^ ?

1ec5:
> I'm torn, but I'm leaning towards `MGLAccountManager` as it keeps this completely separate from Metrics and the benefits of not instantiating it if not needed.

[...]

PR #1603, opened 2015-05-19
Counting App Users For Billing

bleege:
For #1595

### DIFF
--- a/platform/ios/MGLAccountManager.m
+++ b/platform/ios/MGLAccountManager.m
@@ -51,11 +51,9 @@
     [[MGLAccountManager sharedManager] setAccessToken:accessToken];
 
     // Update MGLMapboxEvents
-    // NOTE: This is (likely) the initial setup of MGLMapboxEvents
-    [MGLMapboxEvents sharedManager];
-
-    // Count App Users
-    [MGLMapboxEvents pushEvent:MGLEventTypeAppUserCount withAttributes:nil];
+    // NOTE: This is (likely) the initial setup of MGLMapboxEvents (via implicit call to +[MGLMapboxEvents sharedManager])
+    // Turn the Mapbox Turnstile to Count App Users
+    [MGLMapboxEvents pushEvent:MGLEventTypeAppUserTurnstile withAttributes:nil];
 }
 
 + (NSString *) accessToken {
--- a/platform/ios/MGLAccountManager.m
+++ b/platform/ios/MGLAccountManager.m
@@ -51,9 +51,8 @@
     [[MGLAccountManager sharedManager] setAccessToken:accessToken];
 
     // Update MGLMapboxEvents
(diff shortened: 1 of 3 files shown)


platform/ios/MGLAccountManager.m
```diff
@@ -51,8 +51,9 @@ + (void) setAccessToken:(NSString *) accessToken {
     [[MGLAccountManager sharedManager] setAccessToken:accessToken];
 
     // Update MGLMapboxEvents
-    // NOTE: This is (likely) the initial setup of MGLMapboxEvents
-    [MGLMapboxEvents sharedManager];
+    // NOTE: This is (likely) the initial setup of MGLMapboxEvents (via implicit call to +[MGLMapboxEvents sharedManager])
+    // Turn the Mapbox Turnstile to Count App Users
+    [MGLMapboxEvents pushEvent:MGLEventTypeAppUserTurnstile withAttributes:nil];
```
1ec5: `-setAccessToken:` can be called more than once. In particular, with #1553, if the developer decides to set the access token programmatically, `-setAccessToken:` will be called once with `nil` from within `+load` and again when the developer gets around to calling `+setAccessToken:`. You don’t want two turnstile events in that case.

A cleaner approach would be to leave the call to `+[MGLMapboxEvents sharedManager]` as is and just push a turnstile event at the end of `-[MGLMapboxEvents init]`.

bleege: GTK about the evolving different ways into the SDK.  I just moved it to `-[MGLMapboxEvents init]` and it works great.  Thanks!

platform/ios/MGLMapboxEvents.m
```diff
@@ -430,9 +435,25 @@ - (void) pushEvent:(NSString *)event withAttributes:(NSDictionary *)attributeDic
     __weak MGLMapboxEvents *weakSelf = self;
 
     dispatch_async(_serialQueue, ^{
+
         MGLMapboxEvents *strongSelf = weakSelf;
+
         if ( ! strongSelf) return;
-        
+
+        if ([MGLEventTypeAppUserTurnstile isEqualToString:event]) {
```
1ec5: Thanks for taking my earlier feedback into account. Now that I look at this, I’m thinking it would be cleaner to have a separate, specialized method called `-pushTurnstileEvent` just for this chunk of code (and all the weak/strong/dispatch scaffolding above). That way there are no special cases and no possibility of somehow accidentally turning the turnstile more than once.

Not critical though: I think this PR looks good overall.

bleege: I actually like that idea quite a bit.  It further separates the Turnstile tick from the stock Events API conceptually as well as gets rid of the "special cases" which tbh wasn't really sitting well with me.  :+1:

Issue #1334, opened 2015-04-23
-[MGLMapView mapID] always returns nil


1ec5:
`-[MGLMapView mapID]` always returns `nil` because `-styleURL` returns a URL like `https://api.tiles.mapbox.com/styles/v1/texample/texample.deadbeef?access_token=…` even when you set the `styleURL` to `mapbox://texample.deadbeef`.

jfirebaugh:
I'd rather we change `Map::getStyleURL` so that it returns the URL passed to `Map::setStyleURL`.

1ec5:
In that case, we’d need to more the normalization code from `Map::setStyleURL()` to `MapData::getStyleInfo()`.

jfirebaugh:
@1ec5 Can you verify this was fixed by #1607?

1ec5:
Yes, #1607 fixed this issue good.

PR #1607, opened 2015-05-20
DefaultFileSource has responsibility for handling mapbox:// URLs

jfirebaugh:
This moves `Map::setAccessToken` to `DefaultFileSource::setAccessToken` and passes `mapbox://` URLs through the `FileSource` interface. No change to the iOS or Android SDK public API. The node bindings will need the following changes (cc @mikemorris):

* [New values for `kind` enum](https://github.com/mapbox/mapbox-gl-native/blob/f9753043decc82733a0eb4b25a246b536b0f09ba/include/mbgl/storage/resource.hpp#L10-18).
* `FileSource` implementations need to handle `mapbox://` URLs.
* Remove `setAccessToken` from `Map` interface. node-mapbox-gl-native itself no longer needs to concern itself with access tokens. This is a task for `FileSource` implementations now.

Before merging this I'm planning to make another change to the interface, removing the `kind` enum in favor of specific `FileSource` methods like `requestStyle`, `requestTile`, etc.

### DIFF
--- a/platform/ios/MGLMapView.mm
+++ b/platform/ios/MGLMapView.mm
@@ -1303,7 +1303,7 @@ std::chrono::steady_clock::duration secondsAsDuration(float duration)
     if ([keyPath isEqualToString:@"accessToken"] && object == [MGLAccountManager sharedManager])
     {
         NSString *accessToken = change[NSKeyValueChangeNewKey];
-        _mbglMap->setAccessToken(accessToken ? (std::string)[accessToken UTF8String] : "");
+        _mbglFileSource->setAccessToken(accessToken ? (std::string)[accessToken UTF8String] : "");
     }
 }
 
--- a/android/cpp/jni.cpp
+++ b/android/cpp/jni.cpp
@@ -373,14 +373,14 @@ nativeSetAccessToken(JNIEnv *env, jobject obj, jlong nativeMapViewPtr, jstring a
     mbgl::Log::Debug(mbgl::Event::JNI, "nativeSetAccessToken");
     assert(nativeMapViewPtr != 0);
     NativeMapView *nativeMapView = reinterpret_cast<NativeMapView *>(nativeMapViewPtr);
-    nativeMapView->getMap().setAccessToken(std_string_from_jstring(env, accessToken));
+    nativeMapView->getFileSource().setAccessToken(std_string_from_jstring(env, accessToken));
 }
 
 jstring JNICALL nativeGetAccessToken(JNIEnv *env, jobject obj, jlong nativeMapViewPtr) {
     mbgl::Log::Debug(mbgl::Event::JNI, "nativeGetAccessToken");
     assert(nativeMapViewPtr != 0);
     NativeMapView *nativeMapView = reinterpret_cast<NativeMapView *>(nativeMapViewPtr);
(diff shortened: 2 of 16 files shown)


tmpsantos:
Nice!

1ec5:
At a glance, it looks like this will fix #1334 as well. :+1:

mikemorris:
Liking the direction here :+1:

1ec5:
I also like that the access token is no longer part of the style URL, so if the access token changes at some point during the lifetime of the file source, we don’t have to go updating other fields. This turns out to jive nicely with the last bit of refactoring I’m doing for #1553.

platform/ios/MGLMapView.mm
```diff
@@ -148,12 +148,12 @@ - (instancetype)initWithCoder:(NSCoder *)decoder
 
 - (NSString *)accessToken
 {
-    return @(_mbglMap->getAccessToken().c_str()).mgl_stringOrNilIfEmpty;
+    return @(_mbglFileSource->getAccessToken().c_str()).mgl_stringOrNilIfEmpty;
```
1ec5: With #1553, `MGLMapView` no longer calls `getAccessToken()` anywhere.

platform/ios/MGLMapView.mm
```diff
@@ -148,12 +148,12 @@ - (instancetype)initWithCoder:(NSCoder *)decoder
...
 }
 
 - (void)setAccessToken:(NSString *)accessToken
 {
-    _mbglMap->setAccessToken(accessToken ? (std::string)[accessToken UTF8String] : "");
+    _mbglFileSource->setAccessToken(accessToken ? (std::string)[accessToken UTF8String] : "");
```
1ec5: With #1553, the call to `setAccessToken()` has been moved to [`-observeValueForKeyPath:ofObject:change:context:`](https://github.com/mapbox/mapbox-gl-native/blob/master/platform/ios/MGLMapView.mm#L1285), which is automatically called on every instance of `MGLMapView` any time `MGLAccountManager.accessToken` changes.

jfirebaugh:
@1ec5 Thanks -- I'll rebase this on master (post #1553) and redo the iOS portions.

ljbade:
Looks good on Android side

kkaefer:
> Before merging this I'm planning to make another change to the interface, removing the kind enum in favor of specific FileSource methods like requestStyle, requestTile, etc.

We can revisit this in another pull request. Any particular reason for doing so? We don't actually use these enums anywhere (except as a hint for compressing responses), so we could do away with them as well?

jfirebaugh:
> We can revisit this in another pull request. Any particular reason for doing so?

I would like to avoid interface churn for node-mbgl or whatever the downstream consumer that implements the FileSource interface in node is nowadays. Maybe we can avoid updating it until the FileSource interface is finalized?

mikemorris:
Working on implementing this, and I think we'll need to add some capacity to pass an access token through a `Map::RenderStill` call through to the `mbgl::Request` objects created by the `FileSource` implementation. Explicitly setting an a global access token is going to cause race conditions when multiple maps use the same `FileSource`.

mikemorris:
> removing the kind enum in favor of specific FileSource methods

-> https://github.com/mapbox/mapbox-gl-native/issues/3374

PR #1608, opened 2015-05-20
Remove redundant didChangeAuthorizationStatus: in favor of didFailWithError:

friedbunny:
This PR removes `locationManager:didChangeAuthorizationStatus:` in favor of `locationManager:didFailWithError:` for turning off the map's location features and for notifying the developer of location permission changes.

[Setting `showsUserLocation = NO` in `locationManager:didChangeAuthorizationStatus:`](https://github.com/mapbox/mapbox-gl-native/blob/51618c35d79cfb237e62e70993182e87903c3ccd/platform/ios/MGLMapView.mm#L2121) would turn off the location manager, which meant that the subsequent call to `locationManager:didFailWithError:` would never happen, nor would the call to our  `mapView:didFailToLocateUserWithError:` delegate method.

Supercedes #1598.

/cc @incanus @1ec5 @bleege

### DIFF
--- a/platform/ios/MGLMapView.mm
+++ b/platform/ios/MGLMapView.mm
@@ -2113,15 +2113,6 @@ CLLocationCoordinate2D latLngToCoordinate(mbgl::LatLng latLng)
     }
 }
 
-- (void)locationManager:(__unused CLLocationManager *)manager didChangeAuthorizationStatus:(CLAuthorizationStatus)status
-{
-    if (status == kCLAuthorizationStatusDenied || status == kCLAuthorizationStatusRestricted)
-    {
-        self.userTrackingMode  = MGLUserTrackingModeNone;
-        self.showsUserLocation = NO;
-    }
-}
-
 - (void)locationManager:(__unused CLLocationManager *)manager didFailWithError:(NSError *)error
 {
     if ([error code] == kCLErrorDenied)


friedbunny:
I tested this with access to location services denied (`kCLAuthorizationStatusDenied`) and device-level "restricted" (`kCLAuthorizationStatusRestricted`). 

Both worked with `locationManager:didFailWithError:`, though gave the same `kCLErrorDenied` error.

This is also how MapKit behaves.

jfirebaugh:
Committed as f5c1e2630e553a727a91f0de623bf3ec1bdcc559.

MadeByDouglas:
As a user of mapBox following this discussion, I'm curious what your best practice recommendation are for handling various user authorization scenarios. For example, I have `mapView.showsUserLocation = true` in viewDidLoad, this seems to work as expected if the user accepts or denies the iOS message. 

However, on subsequent openings of the map, if the status is denied I would like to send a message suggesting the user to change their settings in iOS, what is the best way to do this with just using MapBox and not needing to import CoreLocation and calling `didChangeAuthorizationStatus` ? 

It seems from this discussion  it would be `didFailToLocateUserWithError` however how do I properly parse that NSError? I just get `NSError	domain: "kCLErrorDomain" - code: 1` What do I do with that? How do I know thats an error because authorization and not something else. It seems it would be more intuitive if it still passed the CLAuthorizationStatus object.

friedbunny:
`NSError domain: "kCLErrorDomain" - code: 1` is `kCLErrorDenied`. You can see the rest of the `kCLErrorDomain` enum definitions [in Apple’s docs](https://developer.apple.com/library/ios/documentation/CoreLocation/Reference/CoreLocationConstantsRef/index.html#//apple_ref/c/tdef/CLError).

You should compare the incoming error object against `kCLErrorDenied`/`CLError.Denied`:

```swift
func mapView(mapView: MGLMapView, didFailToLocateUserWithError error: NSError) {
    if error.domain == kCLErrorDomain && error.code == CLError.Denied.rawValue {
        ...
    }
}
```

MadeByDouglas:
ahh, ok that makes sense, thank you, i appreciate the apple docs link too

Issue #1363, opened 2015-04-28
Metrics connectivity tests (SSL pinning)


incanus:
Latest builds still do not seem to be sending live metrics events. Possibly a bug still in SSL pinning. 

1. Fix this. 
1. Write a test. 
    - Start app. 
    - Do stuff with the map. 
    - Background the app to flush the queue (can we do this with KIF?)
    - Observe SSL network connection, which we'll assume is for metrics since the map isn't being moved. 

Maybe a little brittle, but shouldn't false-positive and about as a good a way as any to test a pinned SSL. 

/cc @bleege

incanus:
Possibly useful: https://github.com/dstnbrkr/VCRURLConnection

bleege:
On brach [1306-metrics-debug](https://github.com/mapbox/mapbox-gl-native/compare/1363-metrics-debug).

incanus:
Actually, let's leave this open and think about testing strategies for SSL pinning still.

incanus:
@bleege Any sense of what we might be able to increase test coverage on here for b2, if at all?

bleege:
Picking this backup...

bleege:
Testing should be quicker now that the `turnstile` event is created and then flushed immediately at launch.  It uses the same events infrastructure so it should be sufficient for testing.

The definition of success for sending events is receiving an `HTTP 204` message back from the event server.

bleege:
Current Metrics Tests --> https://github.com/mapbox/mapbox-gl-native/blob/88b3cfe37de7918723fb9ebb6b886aaec4808bcd/test/ios/MetricsTests.m

bleege:
The actual sending of events data to the server via HTTP is done via `-[MGLMapboxEvents postEvents]` which is an internally accessible only method triggered by `+[MGLMapboxEvents flush]`.

The `-[MGLMapboxEvents pushTurnstileEvent]` method will push an event and then force a flush of the event queue, so it could be an ideal way to test this without too much extra configuration work.  This is executed once via `-[MGLMapboxEvents init]`.

bleege:
Current error message when running in via Mapbox GL dev app.

```
[...]

PR #1614, opened 2015-05-21
Update iOS tests for access token changes

friedbunny:
Updates the iOS testing apparatus to work with the access token changes introduced in #1553.

Moves the access token setting to `App-Info.plist`.

Also sets `MGLMapboxMetricsEnabledSettingShownInApp` in `App-Info.plist`.

/cc @incanus @1ec5 @bleege

### DIFF
--- a/test/ios/App-Info.plist
+++ b/test/ios/App-Info.plist
@@ -2,6 +2,8 @@
 <!DOCTYPE plist PUBLIC "-//Apple//DTD PLIST 1.0//EN" "http://www.apple.com/DTDs/PropertyList-1.0.dtd">
 <plist version="1.0">
 <dict>
+	<key>MGLMapboxAccessToken</key>
+	<string>pk.eyJ1IjoianVzdGluIiwiYSI6Ik9RX3RRQzAifQ.dmOg_BAp1ywuDZMM7YsXRg</string>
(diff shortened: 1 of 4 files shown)


1ec5:
Thanks, @friedbunny! Can't wait for the tests to run automatically again.

test/ios/App-Info.plist
```diff
@@ -34,6 +36,8 @@
 		<string>UIInterfaceOrientationLandscapeLeft</string>
 		<string>UIInterfaceOrientationLandscapeRight</string>
 	</array>
+	<key>MGLMapboxMetricsEnabledSettingShownInApp</key>
+	<true/>
```
incanus: We should probably actually do this, as well as write tests against it as part of https://github.com/mapbox/mapbox-gl-native/issues/1363#issuecomment-97195050.

friedbunny: Do you mean adding an actual button and KIF-pushing it? It seems like [`[MetricsTests testOptOutAtRuntimeDisablesMetrics]`](https://github.com/mapbox/mapbox-gl-native/blob/dfd6ff00a239ef7c4762346edf798aa5dfb7bf3f/test/ios/MetricsTests.m#L262-L279) covers that case, sans UI.

PR #1615, opened 2015-05-21
Fixed account type setting; reworded IB designable message

### LINKED ISSUE #1613: IB designable fails to recognize access token set in Info.plist
#1553 introduced a new place to set the access token, in the Info.plist, but `MGLMapView`’s IB designable code fails to recognize it: the designable always says “No Access Token”. The main bundle is most likely that of IBDesignablesAgentCocoaTouch, and it’d make no sense to make the developer to touch the framework’s own Info.plist.

1ec5:
This PR fixes some regressions from #1553, as well as some bugs that’ve been lingering for longer:

* We need to look in the host app’s Info.plist, not MapboxGL.framework’s.
* Currently, if you set `MGLMapboxMetricsEnabledSettingShownInApp = YES` in Info.plist, you still get the assertion failure in `-[MGLMapboxEvents init]` because the access token is read in first, causing the shared `MGLMapboxEvents` to be created. This change rearranges the order that the settings are read in, since one is dependent on the other.
* Interface Builder was still showing an “Access Token” inspectable, which would’ve led to an assertion failure if set.
* The IB designable message is only relevant to developers displaying Mapbox-hosted maps, so the message should reflect that scope.
* The IB designable message can’t detect whether the access token is set in Info.plist, and it can’t detect whether a Settings bundle has been added. So I nixed the “A-OK” Mapbox blue designable screen – we just show steps to get started no matter what and point to the “[First steps with Mapbox GL for iOS](https://www.mapbox.com/guides/first-steps-gl-ios/)” guide.

Fixes #1613.

### DIFF
--- a/include/mbgl/ios/MGLMapView+IBAdditions.h
+++ b/include/mbgl/ios/MGLMapView+IBAdditions.h
@@ -8,7 +8,6 @@
 // inspectables declared in MGLMapView.h are always sorted before those in
 // MGLMapView+IBAdditions.h, due to ASCII sort order.
 
-@property (nonatomic) IBInspectable NSString *accessToken;
 @property (nonatomic) IBInspectable NSString *mapID;
 
 // Convenience properties related to the initial viewport. These properties
--- a/platform/ios/MGLAccountManager.m
+++ b/platform/ios/MGLAccountManager.m
@@ -15,13 +15,15 @@
 @implementation MGLAccountManager
 
 + (void)load {
-    // Read initial configuration from Info.plist.
-    NSBundle *bundle = [NSBundle bundleForClass:self];
-    self.accessToken = [bundle objectForInfoDictionaryKey:@"MGLMapboxAccessToken"];
+    // Read the initial configuration from Info.plist. The shown-in-app setting
+    // preempts the Settings bundle check in -[MGLMapboxEvents init] triggered
+    // by setting the access token.
+    NSBundle *bundle = [NSBundle mainBundle];
     NSNumber *shownInAppNumber = [bundle objectForInfoDictionaryKey:@"MGLMapboxMetricsEnabledSettingShownInApp"];
     if (shownInAppNumber) {
         [MGLAccountManager sharedManager].mapboxMetricsEnabledSettingShownInApp = [shownInAppNumber boolValue];
     }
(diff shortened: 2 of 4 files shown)


friedbunny:
> Currently, if you set MGLMapboxMetricsEnabledSettingShownInApp = YES in Info.plist, you still get the assertion failure in -[MGLMapboxEvents init] because the access token is read in first, causing the shared MGLMapboxEvents to be created. This change rearranges the order that the settings are read in, since one is dependent on the other.

Mmm, was looking at this now because it was crashing the tests. Have a couple questions:

1. Why use `[MGLAccountManager sharedManager].mapboxMetricsEnabledSettingShownInApp` but then `self.accessToken` in `+[MGLAccountManager load]`?

1. Why start `[MGLMapboxEvents sharedManager]` in `+[MGLAccountManager setAccessToken:]` and not at the end of `+[MGLAccountManager load]`?

friedbunny:
`MapViewTests` pass with these changes (after applying 0c783141f68bf689e3fef294dcb5d84371dcad19), thanks to the reordering of the first `[MGLMapboxEvents sharedManager]` call.

1ec5:
> Why use `[MGLAccountManager sharedManager].mapboxMetricsEnabledSettingShownInApp` but then `self.accessToken` in `+[MGLAccountManager load]`?

#1553 made `+[MGLAccountManager setMapboxMetricsEnabledSettingShownInApp:]` unavailable as a precursor to removing it entirely. (Although I see it’s [still implemented](https://github.com/mapbox/mapbox-gl-native/blob/master/platform/ios/MGLAccountManager.m#L51-53).) As far as I can tell, there’s no way to make a method un-unavailable privately without subclassing. If you can think of a way to override the `unavailable()` attribute, I’m all ears!

> Why start `[MGLMapboxEvents sharedManager]` in `+[MGLAccountManager setAccessToken:]` and not at the end of `+[MGLAccountManager load]`?

Nothing that `MGLMapboxEvents` does makes sense without an access token. The events API (the server API, that is) requires an access token. Moreover, Mapbox Metrics shouldn’t be active if the app is using a non-Mapbox-hosted style and tile server.

1ec5:
Here’s what the designable looks like now in various form factors:

compactcompact
anyany
small
smallish

As you can see, I also added some additional constraints so that the designable looks more presentable when it’s too small to show everything.

PR #1620, opened 2015-05-21
Cleaned up MGLMapViewDelegate

### LINKED ISSUE #1385: refactor MGLMapViewDelegate into its own header
We should move this out of `MGLMapView.h` to cut down on size, much like MapKit and our raster SDK.

1ec5:
It’s now in a separate header, MGLMapViewDelegate.h, and there are no longer any “TODO” comments where documentation comments belong.

Fixes #1385.


incanus:
For the formerly-undocumented delegate methods, they don't all work. Past convention has been to only document methods that work properly so that developers feel free to use them. We should ticket out the methods and categories of functionality that don't work as well as mark the methods as buggy/unreliable in the header if we're going to do this. 

Related: https://github.com/mapbox/mapbox-gl-native/issues/1431

1ec5:
Ah, I read "TODO" as "to document" rather than "to implement". We shouldn't even declare methods we don't support yet. There are plenty of other APIs we'd love to have but it doesn't make sense to put them in developers' autocomplete prematurely.

1ec5:
Take two in #1976.

friedbunny:
I still like moving these to `include/mbgl/ios/MGLMapViewDelegate.h`, we should do that.

Issue #1026, opened 2015-03-19
remove delay argument from notifyMapChange


jfirebaugh:
It shouldn't be the responsibility of the view implementation to set up a timer to resend the notification after a delay -- notifications should just be dispatched at the right time.

incanus:
This is not a delay; it relates to the animation timeouts used. For example, if Cocoa tells the map to move on a `300ms` animation, the `Map`/`Transform` are authoritative on whether that animation completed or not, and should send the completion notification at the end of the delay.

incanus:
For context on the need for these, see callbacks like `-[MKMapViewDelegate mapView:regionDidChangeAnimated:]`. 

https://developer.apple.com/library/ios/documentation/MapKit/Reference/MKMapViewDelegate_Protocol/index.html#//apple_ref/occ/intfm/MKMapViewDelegate/mapView:regionDidChangeAnimated:

jfirebaugh:
> the `Map` / `Transform` are authoritative on whether that animation completed or not, and should send the completion notification at the end of the delay

They should be -- that's what I'm suggesting in this issue. But it doesn't look to me like that's the current behavior.

See, e.g.

https://github.com/mapbox/mapbox-gl-native/blob/b4dbba93f868b4d44d0f858a9c54924b910ef62d/src/mbgl/map/transform.cpp#L286-L289

and 

https://github.com/mapbox/mapbox-gl-native/blob/b4dbba93f868b4d44d0f858a9c54924b910ef62d/platform/ios/MGLMapView.mm#L1629-L1637

jfirebaugh:
Also `MGLMapView` shouldn't need to dispatch `notifyMapChange` manually:

https://github.com/mapbox/mapbox-gl-native/blob/b4dbba93f868b4d44d0f858a9c54924b910ef62d/platform/ios/MGLMapView.mm#L586-L598

incanus:
> Also `MGLMapView` shouldn't need to dispatch `notifyMapChange` manually:

It does this because each platform is responsible for the velocity-based deceleration that feels right on that platform, which means differing durations.

jfirebaugh:
Sure. But the point of this bug is: given an easing operation is started on `Map` with a particular duration, then `Map` should have sole responsibility for dispatching notifications of the progress of that easing operation.

Make sense?

incanus:
[...]

PR #1624, opened 2015-05-22
Always drive rendering from the main thread

kkaefer:
In most cases, the main/UI thread triggers rendering, except for our GLFW sample app.

### DIFF
--- a/platform/ios/MGLMapView.mm
+++ b/platform/ios/MGLMapView.mm
@@ -2755,11 +2755,16 @@ class MBGLView : public mbgl::View
         [EAGLContext setCurrentContext:nil];
     }
 
-    void invalidate(std::function<void()>) override
+    void invalidate() override
     {
         [nativeView performSelectorOnMainThread:@selector(invalidate)
                                      withObject:nil
                                   waitUntilDone:NO];
+    }
+
+    void swap() override
+    {
+        // no-op
     }
 
     private:
--- a/platform/ios/MGLMapView.mm
+++ b/platform/ios/MGLMapView.mm
@@ -2597,10 +2597,7 @@ CLLocationCoordinate2D MGLLocationCoordinate2DFromLatLng(mbgl::LatLng latLng)
 - (void)invalidate
 {
     MGLAssertIsMainThread();
-
     [self.glView setNeedsDisplay];
-
-    [self notifyMapChange:@(mbgl::MapChangeRegionIsChanging)];
 }
 
 - (BOOL)isFullyLoaded
--- a/platform/ios/MGLMapView.mm
+++ b/platform/ios/MGLMapView.mm
@@ -2726,23 +2726,10 @@ class MBGLView : public mbgl::View
         // no-op
     }
 
-    void notifyMapChange(mbgl::MapChange change, std::chrono::steady_clock::duration delay = std::chrono::steady_clock::duration::zero()) override
+    void notifyMapChange(mbgl::MapChange change) override
     {
-        if (delay != std::chrono::steady_clock::duration::zero())
-        {
-            dispatch_after(dispatch_time(DISPATCH_TIME_NOW, std::chrono::duration_cast<std::chrono::nanoseconds>(delay).count()), dispatch_get_main_queue(), ^
-            {
-                [nativeView performSelector:@selector(notifyMapChange:)
-                                 withObject:@(change)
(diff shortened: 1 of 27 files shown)


kkaefer:
@ljbade from the map thread, how can we signal to the main UI/Java thread that it must call `map->renderSync()`?

ljbade:
@Kkafer first issue is that you need to attach java runtime to the map thread then you can use JNI to invoke Java. An example is the event handling for the map class in the nativemapview.cpp file.

Next issue is that Android does not allow non UI thread to touch UI objects so likely you need to call an async handler that runs on the UI thread.

It may be simpler to implement async call in C++ to UI thread

kkaefer:
@ljbade so the flow would be:

- In the Map thread, attach the Java runtime (can we attach this in addition to the main thread or can only be one thread attached at one point?)
- call a Java function from native code on the Map thread
- the Java function dispatches to the Main thread in Java
- the Main thread in Java triggers a rerender

We can't really do an async call in C++ land since we don't have an event loop running in the Main thread. Essentially, what I'm looking for is a way to trigger a frame update from the Map thread. Do you see any other options?

ljbade:
@kkaefer you can have many threads attached, you just have to remember to detach before deleting the thread

Isn't most the Map functions async? Perhaps this can all be done in C++ side by just calling Map.

jfirebaugh:
We should pick this up again, I think it's a dependency for #1026.

kkaefer:
@jfirebaugh updated and fixed Android support. This is mostly a stopgap for Android to keep the app functional; we need a larger Android refactoring eventually.

src/mbgl/map/map_context.cpp
```diff
@@ -302,9 +306,16 @@ void MapContext::renderStill(StillImageCallback fn) {
     }
 
     callback = fn;
+    mayRender = true;
     triggerUpdate(Update::RenderStill);
 }
 
+void MapContext::renderSync() {
+    mayRender = true;
```
jfirebaugh: Since `render` doesn't check `mayRender`, it seems like it's equivalent to "is a still render pending?". Can we use `callback` in `update()` instead, like:

```
if (callback) {
    render();
} else {
    view.invalidate();
}
```

Or is there something I'm missing?

jfirebaugh:
I made `Transform` live on the main thread only. As part of that, `Transform::updateTransitions()` is now called on the main thread, which allows it to dispatch `MapChangeRegionIsChanging` and `MapChangeRegionDidChangeAnimated` correctly.

@incanus, can you review these changes (particularly 9dc0263 and 7549aaf) and test them with the view-syncing stuff you're working on?

The remaining fix for `MapChangeRegion*` events that I'm seeing is that `Transform` should suppress events if `state.gestureInProgress` is `true`. In this situation, `MGLMapView` will be dispatching its own set of `MapChangeRegion*` events as the gesture progresses. @incanus: agree?

incanus:
I don't see this affecting what I'm working on, as I'm no longer driving annotation view updates from the notification indirectly, but rather in the same method that the core GL view is told to render a frame synchronously. I will test, though. 

As for approach, it looks good and tests pass, but I'll take a closer look.

kkaefer:
@incanus @jfirebaugh any objections to merging this?

incanus:
Still got a bit more digging to do — will hit it this am.

platform/ios/MGLMapView.mm
```diff
@@ -2710,23 +2707,10 @@ void notify() override
...
-                                 afterDelay:0];
-            });
-        }
-        else
-        {
-            assert([[NSThread currentThread] isMainThread]);
-
-            [nativeView notifyMapChange:@(change)];
-        }
+        assert([[NSThread currentThread] isMainThread]);
```
1ec5: Please use `MGLAssertIsMainThread()`, which provides more details in the assertion message.

1ec5: On second thought, that won’t work because `_cmd` is only available inside Objective-C classes. `NSCAssert()` would still be preferable to `assert()` in Objective-C code, though.

incanus:
Ok I am :+1: if you are @jfirebaugh — this doesn't interfere with anything I'm working on and it certainly looks cleaner from the will/am/did update notification POV.

ljbade:
Great work. Thanks for fixing this.

PR #1629, opened 2015-05-23
Thread-safety for http_request_nsurl.mm

jfirebaugh:
Don't access the response pointer from two threads, perhaps simultaneously.

`HTTPRequest::completionHandler`, a static method (so that it has no access to member variables), takes the place of `HTTPRequest::handleResult`. The result is transferred to the FileSource thread via a promise.

This is a speculative fix for the occasional `HTTPRequest::handleResult` crashes we've seen.

CC @kkaefer @1ec5 for review.

### DIFF
--- a/include/mbgl/storage/response.hpp
+++ b/include/mbgl/storage/response.hpp
@@ -9,6 +9,12 @@ class Response {
 public:
     enum Status : bool { Error, Successful };
 
+    Response() = default;
+    Response(Status status_, std::string message_)
+        : status(status_),
+          message(message_) {
+    }
+
     Status status = Error;
     std::string message;
     int64_t modified = 0;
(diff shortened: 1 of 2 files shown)


kkaefer:
Do you have a scenario in which this could happen? `handleResponse` should only ever be called *after* the completion handler has finished running, since it is triggering the `handleResponse` through the async. I'm a little worried that we're having another issue here that we're covering rather than resolving with this patch.

jfirebaugh:
> Do you have a scenario in which this could happen?

I don't. My best guess is the race is between `HTTPRequest::handleResult` and `HTTPRequest::retry`.

kkaefer:
Hm, the async handle shouldn't be invoked when `retry` is; there's no interaction between the result thread and the request thread during retries.

1ec5:
I don’t see any problems here regarding ARC’s management of the Objective-C objects, assuming that the transient `-[NSString UTF8String]` and `-[NSData bytes]` values were not at fault.

PR #1632, opened 2015-05-23
Corrupt data not handled properly

tmpsantos:
We currently report correctly if a network request fails but we don't when the network request succeed but the data returned is corrupted. This is specially bad when using renderStill() because in some cases it never returns or the result of the rendering is wrong.

* For most of the resources we just silently ignore the error.
* In some cases like when loading the Source JSON, we do detect parsing errors but we report as network error.
* In case of vector tiles, the Map thread deadlocks.

The idea is to handle errors similarly how we do today with connection errors.

### DIFF
--- a/src/mbgl/map/source.cpp
+++ b/src/mbgl/map/source.cpp
@@ -287,26 +287,25 @@ TileData::State Source::addTile(MapData& data,
         new_tile.data = cache.get(normalized_id.to_uint64());
     }
 
-    auto successCallback = std::bind(&Source::emitTileLoaded, this, true);
-    auto failureCallback = std::bind(&Source::emitTileLoadingFailed, this, std::placeholders::_1);
-
     if (!new_tile.data) {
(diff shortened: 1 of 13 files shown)


tmpsantos:
Added the tests that I'm using to verify this issue.

kkaefer:
> when the network request succeed but the data returned is corrupted.

We have to way in the network layer to detect corruption beyond what the network stack already does for us (packet checksum etc.). Any failures can only be reported by the actual consuming code.

mikemorris:
FWIW, the inital spot I discovered this was handling errors properly when attempting to load a non-gzip-compressed vector tile https://github.com/mapbox/mapbox-gl-native/pull/1588#issuecomment-103970379

tmpsantos:
@mikemorris can you give this branch a try and see if it solves the issue of the non-gzip-compressed tile? No changes are needed on your code if you are already catching the exception_ptr from the request failures.

mikemorris:
Tested, works perfectly :+1:

PR #759, opened 2015-05-26
Add prefix MGL to Reachability

kkaefer:
Since Reachability is such a common name and a lot of existing projects/other libraries use this name, it can cause duplicate symbol issue. 
It would be very helpful if the prefix MGL could be prepended to this objective C class.


incanus:
Something to note here: Reachability is originally based on Apple sample code and has been forked and developed elsewhere, most notably in the somewhat stale [`tonymillion/Reachability`](https://github.com/tonymillion/Reachability). We should be watching for upstream changes in the APIs used here and treating this like we do clipper and other external libs.

Issue #1641, opened 2015-05-26
Add instance And version To Turnstile Data Model


bleege:
@camilleanne requested that we add [`instance`](https://github.com/mapbox/mapbox-gl-native/blob/f4707f72e0ce7f1c40337c72ae5c0c0c8ef0efed/platform/ios/MGLMapboxEvents.m#L485) and [`version`](https://github.com/mapbox/mapbox-gl-native/blob/f4707f72e0ce7f1c40337c72ae5c0c0c8ef0efed/platform/ios/MGLMapboxEvents.m#L483) to the [Turnstile event data model](https://github.com/mapbox/mapbox-gl-native/blob/f4707f72e0ce7f1c40337c72ae5c0c0c8ef0efed/platform/ios/MGLMapboxEvents.m#L439-L442).

@mapbox/mobile @mapbox/mobiledata

bleege:
Tested, Travised, and FF Merged.

PR #1642, opened 2015-05-26
Adding version and instance to Turnstile event

bleege:
For #1641

### DIFF
--- a/platform/ios/MGLMapboxEvents.m
+++ b/platform/ios/MGLMapboxEvents.m
@@ -53,7 +53,6 @@ const NSTimeInterval MGLFlushInterval = 60;
 // All of the following properties are written to only from
 // the main thread, but can be read on any thread.
 //
-@property (atomic) NSString *instanceID;
 @property (atomic) NSString *advertiserId;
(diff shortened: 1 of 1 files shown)


platform/ios/MGLMapboxEvents.m
```diff
@@ -144,6 +142,8 @@ @interface MGLMapboxEvents () <CLLocationManagerDelegate>
 @property (atomic) NSString *appName;
 @property (atomic) NSString *appVersion;
 @property (atomic) NSString *appBuildNumber;
+@property (atomic) NSNumber *version;
```
1ec5: This should just be a global `const NSUInteger`, since there’s no reason for subclassers to override the value or for the value to change at runtime.

bleege: It needs to be an Object and not just an NSUInteger otherwise Xcode complains.

screen shot 2015-05-26 at 2 01 59 pm

1ec5: You can box it, e.g. `@(MGLMapboxMetricsVersion)` every time you stick it in a dictionary.

PR #1651, opened 2015-05-27
Linux build instruction updates

ajashton:
Some changes to the README... I just ran through a build on Ubuntu 14.04 (LTS) and this is what I ran into.

- GCC 4.9 is only available starting with Ubuntu 14.10, needs to be installed from the backport PPA
- The default version of Boost in 14.04 is 1.54, but 1.55 is available from the main repos if specified
- I got a build error about some missing LLVM-3.4 libraries. Installing `libllvm3.4` fixed it.

Do these changes look ok?

### DIFF
--- a/README.md
+++ b/README.md
@@ -28,7 +28,7 @@ Be sure to pull down all submodules first:
 
 ## OS X
 
-Prerequisites include the Boost headers (for routines provided therein) and ImageMagick (for tests). To install all prerequisites, use [Homebrew](http://brew.sh/) and type `brew install pkg-config boost imagemagick`.
+ImageMagick is required for tests. It's recommended to install it with [Homebrew](http://brew.sh/) via `brew install imagemagick`.
 
 To create projects, you can run:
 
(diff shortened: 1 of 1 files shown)


jfirebaugh:
Thanks @ajashton. Can you try it without `libboost1.55-dev` and `libboost-program-options1.55-dev` apt installed? Those should both be handled with mason now.

ajashton:
@jfirebaugh yep, that worked. So the whole boost section can be removed?

jfirebaugh:
Yep. :tada:

mikemorris:
I think the `llvm-3.4` requirement is for Mesa, not sure what it would be required here in addition to mason binary packaging though (dynamic linking or something perhaps?)

ajashton:
@mikemorris do you think I should try to figure out why I needed it or would it be ok to leave the libllvm requirement in there for now?

mikemorris:
@ajashton Think it's fine to ignore for now, just curious why https://github.com/mapbox/mason/blob/mesa-10.4.3/.travis.yml#L18 isn't sufficient.

Issue #1092, opened 2015-03-25
allow region-based viewport setting


incanus:
https://github.com/mapbox/mapbox-gl-native/pull/1082 and user location display on the map introduces `-zoomToSouthWestCoordinate:northEastCoordinate:animated:` brought over from the iOS SDK. We should settle on whether we will use this approach or a more singular structure like MapKit's `MKRegion`. Then, we should expose these methods publicly for general use as well for things like zooming to an array of annotations (the reason the new annotations don't have this yet). 

Applies to all platforms, not just iOS.

incanus:
#1655 is bringing in a `LatLngBounds` struct that we can use here. I think we should stick to coordinates like this rather than use another layer of abstraction like MapKit does (`MKMapPoint` & friends). I'll hit this when shapes land.

incanus:
Downside to not using an abstraction layer like `MKMapPoint` is that what we are specifying in "bounds" are actually spherical trapeziums, like how we do in the raster SDK, i.e. this is a shape that doesn't not have straight lines on the sides and the top and bottom edges are not the same length. 

I don't think this is a problem in practice, but it does get into interesting territory when it comes to determining intersections and overlaps.

incanus:
Coming along with the addition of this API would be something like MapKit's `-showAnnotations:animated:`, which allows passing annotation objects for viewport setting, which is super handy.

incanus:
This isn't trivial, though, since you need to determine the region with regard to the view's current aspect ratio as well as possibly inset/offset padding.

friedbunny:
Per #1433, `-zoomToSouthWestCoordinate:northEastCoordinate:animated:` is bork right now and doesn't behave as it should.

1ec5:
This method should also be refactored to use MGLCoordinateBounds, which was introduced in #1655.

1ec5:
> I think we should stick to coordinates like this rather than use another layer of abstraction like MapKit does (`MKMapPoint` & friends).

MapKit’s geometry data types are named in parallel with Core Graphics geometry data types (MKMapPoint is to CGPoint as MKMapSize is to CGSize as MKMapRect is to CGRect). It’s elegant, but I agree that the additional level of abstraction isn’t necessary at this point.

PR #1655, opened 2015-05-27
shape annotations

incanus:
Runtime shape annotations. 

* [x] C++ API for adding single shape
* [x] C++ API for adding multiples at once
* [x] Runtime styling at annotation create time
* [x] Exposes `StyleProperties` et al. to `Map` users
* [ ] C++ API for reordering
* [x] C++ API for removing
* [x] Cocoa API (including selecting only points)
* [ ] Remove perf penalty when adding points and shapes at same time, since two invalidations occur
* [ ] Track shapes across style changes (https://github.com/mapbox/mapbox-gl-native/issues/1488)
* [x] Bring in GeoJSONVT poly simplification - we could get away with punting for a bit here in exchange for performance were it not for [problems](https://dl.dropboxusercontent.com/u/575564/anigif-1432771398.gif) rendering shapes that span tiles. I do feel good about the state of where I left the GeoJSONVT work and could probably pretty quickly get it into shape to farm out the tiling of shape features instead of the current stopgap [naive tiling based on it](https://github.com/mapbox/mapbox-gl-native/blob/94f7e2fe1d61c28a407a541cb4af80b32412183d/src/mbgl/map/annotation.cpp#L184-L264). 
* [ ] Polygons with cutout holes
* [ ] Shapes around antemeridian? 
* [x] Cocoa API docs (currently blocking tests)

Problems/checkpoints: 

- [x] Adding shape annotations before style is finished loading misses rendering them. This is because `MapContext::updateAnnotationTiles()` [returns early](https://github.com/mapbox/mapbox-gl-native/blob/9722c92db67dd490aea7972b5d2466877cea4f57/src/mbgl/map/map_context.cpp#L153). This doesn't affect points as their styling is all wrapped up in the style parsing step (since the variable parts are covered in the runtime `{sprite}` property), the conclusion of which performs a render. Worked around right now with [a hack to add after a delay upon startup](https://github.com/mapbox/mapbox-gl-native/commit/9722c92db67dd490aea7972b5d2466877cea4f57#diff-f11c34ebd1af988a0ff3d8643d1dede8R157). 

- [x] Threading sanity check. 

- [x] General sanity check. This has been messy work. 

-

```c++
map.addPointAnnotation(mbgl::LatLng(45, -120), "default_marker");

mbgl::FillProperties fillProperties;
fillProperties.fill_color = mbgl::Color({{ 1, 0, 0, 1 }});
fillProperties.stroke_color = mbgl::Color({{ 0, 0, 0, 1 }});
fillProperties.opacity = 0.25;

mbgl::StyleProperties shapeProperties;
shapeProperties.set<mbgl::FillProperties>(fillProperties);

map.addShapeAnnotation({
    mbgl::LatLng(44, -122),
    mbgl::LatLng(46, -122),
    mbgl::LatLng(46, -121),
    mbgl::LatLng(44, -122)
}, shapeProperties);
```

screen shot 2015-05-27 at 3 23 04 pm


macosx/main.mm
```diff
@@ -150,6 +150,27 @@ int main() {
     map.setStyleURL(newStyle.first);
     view.setWindowTitle(newStyle.second);
 
+    map.addPointAnnotation(mbgl::LatLng(45, -120), "default_marker");
```
1ec5: Remember to pull these changes out before merging.

incanus: :+1:

incanus:
Things feel good for some real-world data, too: 

screen shot 2015-05-28 at 2 49 47 pm

incanus:
>  Bring in GeoJSONVT poly simplification - we could get away with punting for a bit here in exchange for performance were it not for problems rendering shapes that span tiles. I do feel good about the state of where I left the GeoJSONVT work and could probably pretty quickly get it into shape to farm out the tiling of shape features instead of the current stopgap naive tiling based on it.

This should actually be pretty easily fixed. In shape adds, I'm just overflowing the tile extent and assuming all shapes belong in only one tile, as a naive extension of the point add. That's why this happens: 

anigif-1432856096

Fixing that up now as a staged approach, so that we don't necessarily need the GeoJSONVT simplification stuff in the first cut. This is a ~2,500 point polyline and it performs well now.

incanus:
Taking a break from feature clipping, which is a bit of going in circles. I may just punt and bring in the GeoJSONVT work which takes care of this now instead of later, bringing along simplification for the ride. My reluctance to do so there is a bottleneck around the `GeoJSONVT` object vending tiles to to the annotation manager — maybe not a valid concern. 

In the break I am going to rough out the Cocoa adding/removing/styling API since that part won't change and we are successfully using the direct C++ one right now.

incanus:
BTW current impact of bad feature clipping is: 

* Lines: small gaps at tile borders. 

    screen shot 2015-05-29 at 4 03 10 pm

* Polys: disappear. 

    screen shot 2015-05-29 at 4 03 20 pm

    screen shot 2015-05-29 at 4 03 23 pm

incanus:
Things are pretty feature-complete for a b2 candidate here and seem to be stable API-wise, both C++ and Cocoa. 

Working next on bringing in GeoJSONVT from #893 to solve multi-tile feature clipping problems. Based on my last work with it, it should be pretty straightforward to just outsource to a global `GeoJSONVT` object for tile creation of given `mbgl::LatLng` arrays, then add those tiles to the annotation manager the way we are doing so currently.

src/mbgl/map/map_context.cpp
```diff
@@ -115,15 +125,110 @@ void MapContext::updateTiles() {
...
+    triggerUpdate(Update::Classes);
+
+    data.annotationManager.resetStaleTiles();
+}
+
+void MapContext::cascadeClasses() {
+    style->cascade(data.getClasses());
+}
+
+void MapContext::recalculateClasses(TimePoint now) {
```
jfirebaugh: This should be called `recalculateStyle`. It's not recalculating classes.

src/mbgl/map/map_context.cpp
```diff
@@ -115,15 +125,110 @@ void MapContext::updateTiles() {
...
+
+    // invalidate annotations layer tiles
     for (const auto &source : style->sources) {
         if (source->info.type == SourceType::Annotations) {
             source->invalidateTiles(ids);
         }
     }
-    triggerUpdate();
+
+    cascadeClasses();
```
jfirebaugh: Why does this both call `cascadeClasses` and `triggerUpdate(Update::Classes)`? The latter should be sufficient.

src/mbgl/map/map_context.cpp
```diff
@@ -115,15 +125,110 @@ void MapContext::updateTiles() {
...
     assert(Environment::currentlyOn(ThreadType::Map));
+
+    data.annotationManager.markStaleTiles(ids);
+
     if (!style) return;
+
+    // grab existing, single shape annotations source
+    const auto& shapeID = AnnotationManager::ShapeLayerID;
+
+    const auto source_it = std::find_if(style->sources.begin(), style->sources.end(),
```
jfirebaugh: Camel-case variable names.

src/mbgl/map/source.hpp
```diff
@@ -46,6 +47,7 @@ class SourceInfo : private util::noncopyable {
     std::string attribution;
     std::array<float, 3> center = {{0, 0, 0}};
     std::array<float, 4> bounds = {{-180, -90, 180, 90}};
+    std::string source_id = "";
```
jfirebaugh: Camel-case.

src/mbgl/map/annotation.cpp
```diff
@@ -234,8 +388,18 @@ std::vector<TileID> AnnotationManager::removeAnnotations(const AnnotationIDs& id
     return affectedTiles;
 }
 
-std::vector<uint32_t> AnnotationManager::getAnnotationsInBounds(const LatLngBounds& queryBounds,
-                                                                const MapData& data) const {
+const std::unique_ptr<Annotation>& AnnotationManager::getAnnotationWithID(uint32_t annotationID) const {
```
jfirebaugh: Returning a reference is unlikely to be thread safe. What happens if the annotation is removed while another thread is using a reference to it?

include/mbgl/ios/MGLShape.h
```diff
@@ -0,0 +1,14 @@
+#import <Foundation/Foundation.h>
+
+#import "MGLAnnotation.h"
+
+/** The `MGLShape` class is an abstract class that defines the basic properties for all shape-based annotation objects. This class must be subclassed and cannot be used as is. Subclasses are responsible for defining the geometry of the shape and providing an appropriate value for the coordinate property inherited from the `MGLAnnotation` protocol. */
+@interface MGLShape : NSObject <MGLAnnotation>
+
+/** The title of the shape annotation. The default value of this property is `nil`. */
+@property (nonatomic, copy) NSString *title;
```
1ec5: This property should be `nonnull` when we ultimately merge in #1578. It looks like the `MKAnnotation` protocol has been audited for nullability but the concrete implementations have not. So [`-[MKAnnotation title]`](https://developer.apple.com/library/ios/documentation/MapKit/Reference/MKAnnotation_Protocol/#//apple_ref/occ/intfp/MKAnnotation/title) comes with this note, which implies `nonnull`:

> Although this property is optional, if you support the selection of annotations in your map view, you are expected to provide this property. This string is displayed in the callout for the associated annotation view.

while corresponding properties in `MKShape` and other subclasses lack this note.

include/mbgl/ios/MGLTypes.h
```diff
@@ -12,3 +12,8 @@ typedef NS_ENUM(NSUInteger, MGLUserTrackingMode)
     /** The map follows the user location and rotates when the heading changes. */
     MGLUserTrackingModeFollowWithHeading
 };
+
+typedef struct {
+    CLLocationCoordinate2D sw;
+    CLLocationCoordinate2D ne;
+} MGLMapBounds;
```
1ec5: “Bounds” has a specific meaning in Cocoa that we should avoid overloading in the public interface. `MGLMapView.bounds` can’t be set to an `MGLMapBounds` for instance.

1ec5: Why not `MGLRegion`? It doesn’t have to be identical to `MKRegion`.

platform/ios/MGLMultiPoint.mm
```diff
@@ -0,0 +1,103 @@
...
+    free(_coords);
+}
+
+- (CLLocationCoordinate2D)coordinate
+{
+    if ([self isMemberOfClass:[MGLMultiPoint class]])
+    {
+        [[NSException exceptionWithName:@"MGLAbstractClassException"
+                                 reason:@"MGLMultiPoint is an abstract class"
+                               userInfo:nil] raise];
```
1ec5: This should be a specialized macro that asserts, so you can call `MGLAssertAbstractClass()` and not have to repeat the same three lines everywhere.

platform/ios/MGLMultiPoint.mm
```diff
@@ -0,0 +1,103 @@
+#import "MGLMultiPoint.h"
+#import "MGLTypes.h"
+
+#import <mbgl/util/geo.hpp>
+
+@implementation MGLMultiPoint
+{
+    CLLocationCoordinate2D *_coords;
```
1ec5: Why not a `std::vector`, as you’ve been using elsewhere?

incanus:
Still chipping away at bringing in GeoJSONVT. I have a solid model for how, but I'm seeing problems in my test runs around on-the-fly tile generation that I'm still working through. 

One downside here will be losing bounds querying of shape annotations, but I think that's ok. We only need points for annotation selection on iOS, and post-b2 we may move to an indexing scheme and _always_ do on-the-fly tile generation, meaning we will query bounds from the index rather than the in-memory tiles.

platform/ios/MGLMultiPoint.mm
1ec5: Never mind, the annotation doesn’t have mutable length, and this is consistent with `-points` in MapKit.

incanus:
Current status: 

* Brought in the GeoJSONVT stuff, which solves tile clipping and spanning problems, though some lingering bugs exist. 
* iOS test app has a shape testing function in the gear menu — adds three shapes to the map. Shapes can exist with markers. 

Next up: 

- [ ] Fix ordering of markers over shapes (trivial)
- [ ] Investigate polyline/null island connection issues - likely GeoJSONVT bug
- [ ] Investigate high-zoom polygon fill issues - likely GeoJSONVT bug
- [ ] Clean up / finalize annotation tile invalidation - currently invalidate all because adding shapes creates tiles on-demand instead of upfront
- [ ] Clean up GeoJSONVT objects on shape removal

incanus:
Feeling fairly good here, and the remaining issues in https://github.com/mapbox/mapbox-gl-native/pull/1655#issuecomment-108965503 are definitely GeoJSONVT port bugs since the same shapes render great in the Swift version.

incanus:
>  Investigate polyline/Null Island connection issues - likely GeoJSONVT bug

Fixed in e9fcb6e — painful _Cocoa_ GeoJSON parsing bug :angry:

incanus:
Polylines look good (~2,800 point hike GPX `->` GeoJSON `->` parsed in `MGLPolyline` in Cocoa): 

anigif-1433490977

The high-zoom polygon fill issue looks like this: 

anigif-1433491001

incanus:
Note when the polyline gets zoomed in around z14-15 a portion disappears, but not everything in that tile. Maybe the same issue as the fill one.

incanus:
What I'm seeing is pretty subtle, but I think I have a lead on it. Something in our rendering either a) doesn't have big enough integers to handle tile geometries generated by GeoJSONVT, and/or we can't draw large negative tile point values. Unless I'm mistaken, GeoJSONVT needs this. 

Take, for example, this problematic bit of a triangle poly in tile `11/330/744`: 

anigif-1433542061

It renders just fine in my [Swift port](https://github.com/mapbox/geojsonvt.swift), where the relevant integers are type `Int`, or 32-bit signed. 

```
(geojsonvt.TileRing) $R1 = 0x00007fa1c2f020b0 {
  geojsonvt.TileGeometry = {}
  points = 4 values {
    [0] = 0x00007fa1c2f03090 {
      geojsonvt.TileGeometry = {}
      x = -182
      y = 2839
    }
    [1] = 0x00007fa1c2f030b0 {
      geojsonvt.TileGeometry = {}
      x = -182
      y = -34816
    }
    [2] = 0x00007fa1c2f030d0 {
      geojsonvt.TileGeometry = {}
      x = 13129
      y = -34816
    }
    [3] = 0x00007fa1c2f03130 {
      geojsonvt.TileGeometry = {}
      x = -182
      y = 2839
    }
  }
```

In tile extents of `4096`, these are wildly overreaching, but come together to draw the tile well. 

ios simulator screen shot jun 5 2015 3 09 48 pm

_Ignore the green grid — this just represents the four child tiles. This is `11/330/744`._

I managed to adapt the C++ GeoJSONVT a bit to work with `int32_t` instead of `uint16_t` as before, and I can get the same tile geometry out to render: 

```
(mbgl::GeometryCollection) $0 = size=1 {
  [0] = size=4 {
    [0] = (x = -182, y = 2839)
    [1] = (x = -182, y = -34816)
    [2] = (x = 13129, y = -34816)
    [3] = (x = -182, y = 2839)
  }
}
```

Whereas before we were getting wildly off values due to integer overflows. This involved defining a `TileCoordinate` `typedef` to `vec2<int32_t>` rather than use our current `Coordinate` of `vec2<int16_t>`, but still no dice in the rendering. 

I'm next going to check out where these points go further down the pipeline and see what I can dig up. 

/cc @mourner @kkaefer @ansis @jfirebaugh

incanus:
What GeoJSONVT basically seems to do is not really clip at all, but rather refer to points in the current tile's frame of reference, hence `-34816` in a `y` value means roughly 8-9 tiles north of us. This is necessary to preserve intra-tile line slopes and the like, but I'm betting our C++ render either discards or clamps them, leading to the weird mixed/twisted fills seen in https://github.com/mapbox/mapbox-gl-native/pull/1655#issuecomment-109196081.

incanus:
Per chat with @ansis, JS doesn't exhibit this high ranges and clamps things pretty normally to `(-buffer, extent + buffer)`. 

screen shot 2015-06-05 at 3 46 41 pm

screen shot 2015-06-05 at 3 48 31 pm

So, use of `int16_t` should't matter and I need to figure out what in both Swift and C++ is causing this in the GeoJSONVT ports.

incanus:
Here's a fun wrinkle. 

anigif-1433563467

incanus:
Fixed the outstanding bug, so this is ready to review! 

snoop

I will start addressing the commit comments by @jfirebaugh and @1ec5 above next before squashing. 

anigif-1433718694

include/mbgl/ios/MGLShape.h
incanus: :+1: We can do another once-over to land #1578 in b3?

include/mbgl/ios/MGLTypes.h
incanus: I thought it'd be more confusing to relate it to `region`, especially if we do decide to tackle the translation layer per https://github.com/mapbox/mapbox-gl-native/issues/1092#issuecomment-108005806. But I do see how `bounds` could be equally confusing. `MGLCoordinateBounds`?

macosx/main.mm
incanus: 7fae111

platform/ios/MGLMultiPoint.mm
incanus: Yep.

src/mbgl/map/map_context.cpp
incanus: For iterators that only fall in local scope, we've been doing `foo_it`.

src/mbgl/map/source.hpp
incanus: Was keeping in line with the other properties here like `min_zoom` and `tile_size`.

src/mbgl/map/map_context.cpp
incanus: c638825

src/mbgl/map/map_context.cpp
incanus: Calling `triggerUpdate(Update::Classes)` runs `cascadeClasses()` and `recalculateStyle()` in the same pass. But here in `MapContext::updateAnnotationTiles()` we've just possibly added some layers to the style on-the-fly in response to a refresh after adding new shape annotations whose styles can't be known at style parse time like points can (i.e. branch on `{sprite}`. As a result, we need to run `applyClassProperties()` for `ClassID::Default` first to enable all the layers for the default class before actually triggering a tile update with the second call.

include/mbgl/ios/MGLTypes.h
incanus: 38927c7

src/mbgl/map/annotation.cpp
incanus: Good call. Investigating.

platform/ios/MGLMultiPoint.mm
incanus: Yeah, this would be good, but it's slightly more than trivial since it would be called from several class implementations. @1ec5 Could you take a look at this? It's not a blocker, so it could be b2 or b3.

src/mbgl/map/annotation.cpp
incanus: 2c9a886

mourner:
@incanus nice job tracking this down!

src/mbgl/util/geojsonvt.cpp
```diff
@@ -0,0 +1,858 @@
...
+    int16_t y = extent * (p.y * z2 - ty);
+
+    return TilePoint(x, y);
+}
+
+#pragma mark - GeoJSONVT
+
+std::vector<ProjectedFeature> GeoJSONVT::convertFeatures(const std::string &data, uint8_t baseZoom, double tolerance, bool debug) {
+
+    if (debug) {
```
kkaefer: Can we `#ifdef DEBUG` that?

src/mbgl/style/style_parser.cpp
```diff
@@ -36,20 +36,30 @@ void StyleParser::parse(JSVal document) {
     if (document.HasMember("layers")) {
         parseLayers(document["layers"]);
 
+        // create shape annotations source
```
kkaefer: Instead of adding these fake layers to the Style when parsing, we should add a facility to add these layers to the Map object itself so that we can still switch styles without losing all of our annotations.

src/mbgl/map/source.cpp
```diff
@@ -521,11 +521,16 @@ bool Source::update(MapData& data,
     return allTilesUpdated;
 }
 
-void Source::invalidateTiles(const std::vector<TileID>& ids) {
+void Source::invalidateTiles(const std::unordered_set<TileID, TileID::Hash>& ids) {
```
kkaefer: Can you explain why this is an unordered set? TileIDs should be sortable fairly easily without having to resort to hashing them.

src/mbgl/map/map_context.cpp
kkaefer: Do we have that documented anywhere?

src/mbgl/map/annotation.hpp
```diff
@@ -18,35 +23,87 @@ namespace mbgl {
...
+    std::unordered_map<TileID, std::weak_ptr<const LiveTileFeature>, TileID::Hash> tilePointFeatures;
+    const LatLngBounds bounds;
+};
 
 class AnnotationManager : private util::noncopyable {
 public:
     AnnotationManager();
     ~AnnotationManager();
 
+    void markStaleTiles(std::unordered_set<TileID, TileID::Hash>);
```
kkaefer: does this need to be public?

src/mbgl/map/annotation.cpp
```diff
@@ -300,13 +533,89 @@ LatLngBounds AnnotationManager::getBoundsForAnnotations(const AnnotationIDs& ids
...
+                }
+            }
+        }
+    }
+
+    return renderTile;
 }
 
-const std::string AnnotationManager::layerID = "com.mapbox.annotations.points";
+const std::string AnnotationManager::PointLayerID = "com.mapbox.annotations.points";
```
kkaefer: Any reason for using strings for these?

src/mbgl/map/annotation.cpp
```diff
@@ -77,118 +67,314 @@ uint32_t AnnotationManager::nextID() {
...
 
-        const uint8_t maxZoom = data.transform.getMaxZoom();
+    if (type == AnnotationType::Shape) {
 
-        // side length of map at this zoom
+        orderedShapeAnnotations.push_back(annotationID);
+
+        using namespace mapbox::util::geojsonvt;
+
+        const uint32_t z2 = 1 << 14;
```
kkaefer: why 14?

src/mbgl/map/annotation.cpp
```diff
@@ -77,118 +67,314 @@ uint32_t AnnotationManager::nextID() {
...
+        const double baseTolerance = 3;
```
kkaefer: magic number?

src/mbgl/util/geojsonvt.hpp
```diff
@@ -0,0 +1,273 @@
...
+
+#include <rapidjson/document.h>
+
+#include <mbgl/util/variant.hpp>
+
+namespace mapbox { namespace util { namespace geojsonvt {
+
+#pragma mark -
+
+class Time {
```
kkaefer: there's already `util::stopwatch`. Any reason to use this instead?

src/mbgl/map/source.hpp
kkaefer: What is it?

src/mbgl/util/geojsonvt.cpp
```diff
@@ -0,0 +1,858 @@
...
+
+        ProjectedFeatureType projectedType = (type == "Polygon" ?
+            ProjectedFeatureType::Polygon :
+            ProjectedFeatureType::LineString);
+
+        ProjectedGeometryContainer *geometry = &rings;
+
+        features.push_back(create(tags, projectedType, *geometry));
+
+        printf("features now has %lu items\n", (unsigned long)features.size());
```
kkaefer: prints to the console in Release mode?

src/mbgl/map/annotation.cpp
mourner: it's a good default for simplification in my tests, although it should be configurable (it is in JS)

kkaefer: What is the unit of measurement here?

mourner: Projected pixels.

incanus:
Split GeoJSONVT out to https://github.com/mapbox/geojson-vt-cpp. Currently just dupe-copied here still; no formal build or dep process.

src/mbgl/map/source.cpp
incanus: My understanding was `unordered_set` was better when sorting wasn't needed. Is that not the case? Similarly, `unordered_map` vs `map`?

src/mbgl/map/annotation.hpp
incanus: Yes, because `MapContext` needs to call it?

src/mbgl/map/annotation.cpp
incanus: To make it a bit clearer when debugging. Else you'd always have some implicit arrangement like "points are `0`, shapes start at `1`. Making these strings didn't seem to have any overhead based on the way that they are used.

incanus: And also, the idea was to namespace them since they are actual style layers, so as to not conflict with any named layers the user may have put in the style.

src/mbgl/map/annotation.cpp
incanus: 74db485

src/mbgl/util/geojsonvt.hpp
incanus: Because GeoJSONVT will be standalone.

src/mbgl/util/geojsonvt.cpp
incanus: a6da538 & https://github.com/mapbox/geojson-vt-cpp/commit/18ef1689b8bbedce279bf4c0dcdd3fadcc1fb13a

src/mbgl/style/style_parser.cpp
incanus: Ticketed concretely at https://github.com/mapbox/mapbox-gl-native/issues/1712.

src/mbgl/util/geojsonvt.cpp
incanus: https://github.com/mapbox/geojson-vt-cpp/issues/5

platform/ios/MGLPolygon.m
```diff
@@ -0,0 +1,20 @@
+#import "MGLPolygon.h"
+
+@interface MGLMultiPoint (MGLPolygon)
+
+- (instancetype)initWithCoordinates:(CLLocationCoordinate2D *)coords count:(NSUInteger)count;
+- (BOOL)intersectsOverlayBounds:(MGLCoordinateBounds)overlayBounds;
```
1ec5: This is kind of a sneaky way of getting access to these methods. The declarations should go in a private header in platform/ios/.

incanus: e45bacf4f9ad8fe743a5cc8d28b7adc351c14236

src/mbgl/map/source.hpp
incanus: It's a way of finding the single (multi-layer) shape annotations source later in the tile update step. It's not the only vector source, it doesn't have a `url`, it doesn't have a `tiles` template, and it doesn't have `attribution`, so there is no other way to get a pointer to it later.

incanus:
* Squash commit just happened in `master` with https://github.com/mapbox/mapbox-gl-native/commit/bd0bf294bf1e350b184e54a824430c911dd66b9b. 
* C++ API for reordering: https://github.com/mapbox/mapbox-gl-native/issues/1728
* Polygons with cutout holes / interior rings: https://github.com/mapbox/mapbox-gl-native/issues/1729
* Perf, annotations disappearing, and antemeridian issues from original ticket in https://github.com/mapbox/mapbox-gl-native/pull/1655#issue-81651020 are all tracked in existing tickets. 

336x199px-ll-ffcde97d_edge

Issue #1477, opened 2015-05-08
Deadlock closing app after failed tile request


springmeyer:
To replicate:

  - launch the app, hit `s` until you get to the mapbox-streets style.
  - edit `./style/style/mapbox-streets-v7.json` and give a phony url for one of the layers:

```diff
diff --git a/styles/mapbox-streets-v7.json b/styles/mapbox-streets-v7.json
index 565acd2..d83e0fc 100644
--- a/styles/mapbox-streets-v7.json
+++ b/styles/mapbox-streets-v7.json
@@ -168,8 +168,8 @@
   },
   "sources": {
     "mapbox://mapbox.mapbox-streets-v6": {
-      "url": "mapbox://mapbox.mapbox-streets-v6",
-      "type": "vector"
+      "type": "vector",
+      "url":"http://something.thatdoesnotexist.net"
     },
     "mapbox://mapbox.mapbox-terrain-v2": {
       "url": "mapbox://mapbox.mapbox-terrain-v2",
```

 - launch the app and just the terrain will load
 - then try to close the app and it will not close since it will be hung in a deadlock

Seen with `mac osx` and https://github.com/mapbox/mapbox-gl-native/commit/21c419bd91fcb6fc518a6fcf0f8d99a44b673fe6. The Activity Monitor callstack looks like: https://gist.github.com/springmeyer/b39debea42daa9f3e472

kkaefer:
Looks like the request never gets canceled. This is a regression of https://github.com/mapbox/mapbox-gl-native/commit/9781785ab73e8394e8b92625cc4741952f47955d which was removed again in https://github.com/mapbox/mapbox-gl-native/issues/1336

kkaefer:
This regressed as part of #1065; we are no longer sending an explicit termination message to the Map thread, so we never destruct ~MapContext, never destruct ~Source, never cancel the request.

jfirebaugh:
`~MapContext` is called when the [Thread object](https://github.com/mapbox/mapbox-gl-native/blob/5c7867f435f9d94109ce11fcbbd153b016e02fe0/src/mbgl/util/thread.hpp#L138-138) goes out of scope. Something else must be going on, I'll investigate.

jfirebaugh:
Ah, I think I see what you mean about an explicit termination message @kkaefer. I think that should be avoidable with the current design though, where destructors get called through the course of shutting down the thread.

[...]

PR #1657, opened 2015-05-28
Map object destruction is hanging

tmpsantos:
Under some scenarios, destroying the Map object is hanging. This is specially easy to reproduce if we trigger rendering and when a request fails with a few other pending, we delete the Map object.

This is probably happening because there are pending requests holding the Map Thread's main loop alive and deadlocking somewhere. In the past we had a method for canceling all the requests that belong to an Environment, but now we need to make sure that every object doing requests is canceling pending requests at the dtor.

### DIFF
--- a/src/mbgl/map/map_context.cpp
+++ b/src/mbgl/map/map_context.cpp
@@ -36,6 +36,11 @@ MapContext::MapContext(uv_loop_t* loop, View& view_, FileSource& fileSource, Map
 }
 
 MapContext::~MapContext() {
+    // Make sure we call cleanup() before deleting this object.
+    assert(!style);
+}
+
+void MapContext::cleanup() {
     view.notify();
 
     // Explicit resets currently necessary because these abandon resources that need to be
--- a/src/mbgl/map/map_context.cpp
+++ b/src/mbgl/map/map_context.cpp
@@ -176,6 +176,11 @@ void MapContext::renderStill(StillImageCallback fn) {
         throw util::Exception("Map is currently rendering an image");
     }
 
+    if (style->getLastError()) {
+        fn(style->getLastError(), nullptr);
+        return;
+    }
+
     callback = fn;
     triggerUpdate(Update::RenderStill);
 }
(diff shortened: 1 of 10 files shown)


tmpsantos:
Progress update: the resource loading is starting at the moment we set the style instead of when we call `renderStill()`. If something goes wrong during this time window we never get notified because `renderStill()` sets the callback for reporting errors.

tmpsantos:
@jfirebaugh @kkaefer please take a look at this on too

mikemorris:
@tmpsantos Still seeing hangs as a race condition when testing node-mbgl against this. When three requests emit errors to the log and then the error callback is returned and the destructor is called, the process hangs. When all four requests emit errors before the callback returns, the process exits cleanly.

I'm testing this by enabling the [`returns an error`](https://github.com/mapbox/node-mapbox-gl-native/blob/master/test/js/map.test.js#L223-L243) test in node-mbgl.

src/mbgl/map/map_context.cpp
```diff
@@ -47,6 +47,14 @@ MapContext::MapContext(uv_loop_t* loop, View& view_, FileSource& fileSource, Map
 }
 
 MapContext::~MapContext() {
+    if (resourceLoader) {
+        // This is likely to be called only on unit tests when
+        // the MapContext is not bound to a Map object.
```
kkaefer: Can we avoid code that is specific to unit tests? This sort of defeats the purpose of having unit tests when we special case them in our regular code.

kkaefer:
In what cases do we not have a `ResourceLoader` object?

src/mbgl/map/map_context.cpp
tmpsantos: > Can we avoid code that is specific to unit tests? This sort of defeats the purpose of having unit tests when we special case them in our regular code.

I agree. Gonna change this.

jfirebaugh:
:+1: 

Did this fix #1477?

tmpsantos:
> Did this fix #1477?

Yes! :smile:

PR #1002, opened 2015-05-28
Move atlas ownership to Style

jfirebaugh:
This follows gl-js and just makes sense -- whenever the style changes the atlases should be blown away.

See also https://github.com/mapbox/mapbox-gl-native/issues/957#issuecomment-77633687.

This consists of reapplying bffee0715458530c6c86f440f757a4de667278a2, which was reverted in 0c985c024b4db6c8063648d493830c67ebff50bc because it destabilized the build, and figuring out what's causing the crashes.


tmpsantos:
> @tmpsantos I am looking at this now and I think ResourceLoader should be merged into Style. We have a lot of assertions / assumptions that these have the same lifetime and it feels like they should just be the same object. WDYT?

Hmm, I think it makes sense. The whole idea of the ResourceLoader was to have something self contained so we can test downloading the stuff from the internet without loading the the rendering pipeline.

You might need to update the ResourceLoader tests (I can do that for you if you want, just ping me when your branch is ready and I can rebase the tests for you).

jfirebaugh:
@tmpsantos I'm not sure what tests you're referring to. Turns out I didn't delete any tests, and CI was passing. I went ahead and merged, feel free to hit up the tests you have in mind.

test/test.gypi
```diff
@@ -62,7 +62,6 @@
         'resources/mock_file_source.cpp',
         'resources/mock_file_source.hpp',
         'resources/mock_view.hpp',
-        'resources/resource_loader.cpp',
```
tmpsantos: This test I had in mind. I can take care of it tomorrow.

jfirebaugh: Oh, I see. For some reason I was thinking that was the `ResourceLoader` source .cpp rather than a test file.

tmpsantos: I'm rebasing as part of #1674.

BTW, awesome job, merging and moving the stuff to Style made the utest a lot cleaner.

Issue #1659, opened 2015-05-28
Metrics To TEST Server Need SSLPinning Support


bleege:
Metrics are not being sent to the TEST / Staging server url due to lack of support for the SSL Certificate on https://cloudfront-staging.tilestream.net/.  Add this cert and check it only if Staging server is endpoint to get #1396 working.

/cc @mapbox/mobiledata

incanus:
Again, SSL pinning works! :wink:

bleege:
Added cert from test server, and only check it if the 2 primary PROD certs aren't found AND it's confirmed that the app is setup to send metrics to Staging.

Travising now and then will merge.

bleege:
Travis tested, merged.  Metrics events should now be showing in Staging for dev builds.

bleege:
Implementing @1ec5's [code review suggestions](https://github.com/mapbox/mapbox-gl-native/pull/1661/files#r31294516) and then will close.

bleege:
Implemented suggestions, re-Travised, merged.  Done.

PR #1661, opened 2015-05-29
SSL Pinning For Test Server

bleege:
For #1659


platform/ios/MGLMapboxEvents.m
```diff
@@ -16,6 +16,7 @@
 static const NSUInteger version = 1;
 static NSString *const MGLMapboxEventsUserAgent = @"MapboxEventsiOS/1.0";
 static NSString *MGLMapboxEventsAPIBase = @"https://api.tiles.mapbox.com";
+static BOOL usingTestServer = NO;
```
1ec5: Globals should be named according to the style seen above: MGLMapboxEventsUsesTestServer.

1ec5: Actually, since it isn't constant, let's make it a property or ivar on MGLMapboxEvents: usesTestServer.

platform/ios/MGLMapboxEvents.m
```diff
@@ -878,6 +885,22 @@ - (void)URLSession:(NSURLSession *)session didReceiveChallenge:(NSURLAuthenticat
...
+                    // See if this is test server
+                    for (int lc = 0; lc < numKeys; lc++) {
+                        SecCertificateRef certificate = SecTrustGetCertificateAtIndex(serverTrust, lc);
+                        NSData *remoteCertificateData = CFBridgingRelease(SecCertificateCopyData(certificate));
+
+                        // Compare Remote Key With Local Version
+                        if ([remoteCertificateData isEqualToData:_testServerCert]) {
+                            // Found the certificate; continue connecting
+                            completionHandler(NSURLSessionAuthChallengeUseCredential, [NSURLCredential credentialForTrust:challenge.protectionSpace.serverTrust]);
+                            found = true;
```
1ec5: Nit: true is YES in Objective-C.

PR #1666, opened 2015-05-29
port labeling, symbol sorting, and reparsing overscaled tiles

ansis:
this ports a bunch of stuff from -js:
- sort overlapping symbols in y-direction https://github.com/mapbox/mapbox-gl-js/pull/1184
- better and faster labeling https://github.com/mapbox/mapbox-gl-js/pull/1079
- reparse overscaled tiles https://github.com/mapbox/mapbox-gl-js/pull/1005

This is finally close to merging. Sorry about the amount of changes in one pull request.

TODO:
- [x] fix resource loader test failure https://travis-ci.org/mapbox/mapbox-gl-native/jobs/64636907#L1660
- [x] testing on an actual ios device. anyone have time to do this?

I haven't noticed any stability problems but a lot has changed so this will probably introduce new bugs.


jfirebaugh:
@ansis This doesn't work very well on device at the moment. It seems to stop loading tiles after a few gestures.

ansis:
@jfirebaugh I just pushed 1f5fb1d which I think should fix that

jfirebaugh:
Yep, feels good now!

bleege:
Same here.  :+1: @ansis

ansis:
A couple things should be reviewed:

- [ ] redoPlacement uses workers safely
- [Source::update calls VectorTileData::redoPlacement](https://github.com/mapbox/mapbox-gl-native/blob/1f5fb1db4afae8a8bae02c42870f6f9e5e5f7905/src/mbgl/map/source.cpp#L516-L518) in the map thread.
- [VectorTileData::redoPlacement](https://github.com/mapbox/mapbox-gl-native/blob/1f5fb1db4afae8a8bae02c42870f6f9e5e5f7905/src/mbgl/map/vector_tile_data.cpp#L121) checks if placement needs to be redone and that there is no worker working on the tile currently. If there isn't, it calls [VectorTileData::workerRedoPlacement](https://github.com/mapbox/mapbox-gl-native/blob/1f5fb1db4afae8a8bae02c42870f6f9e5e5f7905/src/mbgl/map/vector_tile_data.cpp#L138) in a worker which calls [VectorTileData::endRedoPlacement](https://github.com/mapbox/mapbox-gl-native/blob/1f5fb1db4afae8a8bae02c42870f6f9e5e5f7905/src/mbgl/map/vector_tile_data.cpp#L149) in the main thread when it is done.
- [Source::tileLoadingCompleteCallback calls VectorTileData::redoPlacement] (https://github.com/mapbox/mapbox-gl-native/blob/1f5fb1db4afae8a8bae02c42870f6f9e5e5f7905/src/mbgl/map/source.cpp#L570) in the map thread because the angle may have changed while the tile was loading.
- To avoid unsafely overwriting existing buffers while they are being used, SymbolBucket stores the new buffers separately in [`renderDataInProgress`](https://github.com/mapbox/mapbox-gl-native/blob/1f5fb1db4afae8a8bae02c42870f6f9e5e5f7905/src/mbgl/renderer/symbol_bucket.cpp#L343). `renderDataInProgress` gets moved to `renderData` later [from the main thread](https://github.com/mapbox/mapbox-gl-native/blob/1f5fb1db4afae8a8bae02c42870f6f9e5e5f7905/src/mbgl/map/vector_tile_data.cpp#L153).

- [ ] https://github.com/mapbox/mapbox-gl-native/commit/1e044a175cbb714182709b9929fe4a74bfa42322 which implements reparsing tiles for high zoom levels changes `TileID(z, x, y)` to `TileID(z, x, y, sourceZ)`. The first z is used to differentiate overparsed tiles that share the same actual vectortile. The second is used to download the data. For example if tile 15/200/300 needs to be overscaled for z17, the TileID would be (17, 200, 300, 15). Does this seem ok?

- [ ] @ansis should double check that out-of-bounds accesses aren't possible in a couple spots where vertices are accessed by index

mb12:
1.) Is the following scenario handled?

a. worker starts doing replacement.
b. Tile becomes obsolete before worker thread doing replacement finishes.
c. worker should not call endRedoPlacement or somehow VectorTileData pointer should not be invalid when endRedoPlacement is called.

This will work in most cases because TileCache will hold a shared_ptr to VectorTileData. However this needs to work even when TileCache size is zero.

2.) Also does it make sense to do some testing with TileCache size as zero? Tile Cache may hide some bugs because it holds a shared_ptr to TileData.

ansis:
1) Yes, the TileData destructor calls TileData::cancel, which [resets](https://github.com/mapbox/mapbox-gl-native/blob/8c2dbcec5aa1b7499ddcbb01da82b00a2db2876b/src/mbgl/map/tile_data.cpp#L73) workRequest, which [calls](https://github.com/mapbox/mapbox-gl-native/blob/8c2dbcec5aa1b7499ddcbb01da82b00a2db2876b/src/mbgl/util/work_request.cpp#L13) WorkTask::cancel. WorkTask::cancel [clears](https://github.com/mapbox/mapbox-gl-native/blob/8c2dbcec5aa1b7499ddcbb01da82b00a2db2876b/src/mbgl/util/work_task.cpp#L34) the callback (endRedoPlacement) so that it doesn't get called when WorkTask::runAfter gets called later (it's definitely later, because it gets called on the same thread).

2) Good idea. I tested it for a bit with cache size 0 and will keep that in mind for the future

thanks for the suggestions @mb12

1ec5:
So far so good with visual inspection.

jfirebaugh:
> redoPlacement uses workers safely

I don't see any glaring issues, but it's impossible to say much about thread safety conclusively, given the current structure. The TileData class hierarchy and worker code needs an overhaul to cleanly separate data on the worker thread from data on the map thread.

src/mbgl/map/live_tile_data.hpp
```diff
@@ -11,13 +11,15 @@ class LiveTileData : public VectorTileData {
...
                  AnnotationManager&,
-                 float mapMaxZoom,
                  Style&,
                  GlyphAtlas&,
                  GlyphStore&,
                  SpriteAtlas&,
                  util::ptr<Sprite>,
-                 const SourceInfo&);
+                 const SourceInfo&,
+                 float,
```
kkaefer: Can we name parameters when it's not clear from their type?

src/mbgl/util/clip_lines.cpp
```diff
@@ -0,0 +1,66 @@
+#include "clip_lines.hpp"
+
+namespace mbgl {
+namespace util {
+
+std::vector<std::vector<Coordinate>> clipLines(const std::vector<std::vector<Coordinate>> &lines,
+        const int16_t x1, const int16_t y1, const int16_t x2, const int16_t y2) {
```
kkaefer: can we use a box type for this instead of specifying separate ints?

ansis: The `box` (util/box.hpp) doesn't fit well. Is `Rect` (util/rect.hpp) what you were thinking of? or were you thinking of making a new type?

kkaefer: yeah, Rect could work for this?

mb12:
@ansis Does redoPlacement need to be called at higher zoom levels? Specifically  when (map's current zoom > SourceInfo.maxzoom + 1).  Or is this handled in some other way to ensure good symbol density at higher zoom levels?  (CollisionTileData.maxScale is const 2.0)

ansis:
@mb12 this is handled in some other way. mostly by https://github.com/mapbox/mapbox-gl-native/commit/1e044a175cbb714182709b9929fe4a74bfa42322. For zoom levels > maxzoom, tiles from the maximum zoom level are reparsed completely. Layout properties can be functions and in order for these functions to work it needs to reparse those layers at every zoom level. It reparses entire tiles instead of only layers that need it because that was easier to implement.

PR #1674, opened 2015-06-02
Rebase ResourceLoader tests after the it was merged with Style

tmpsantos:
Resource loading test cases were disabled after ResourceLoader was merged with the Style class.


jfirebaugh:
Thanks @tmpsantos!

PR #1686, opened 2015-06-04
Document TileData::State

tmpsantos:
Add documentation for the TileData life cycle and all its possible states.


tmpsantos:
states

* Currently, only raster tiles go to `invalid` state.

Issue #926, opened 2015-02-27
Don't create GL objects in the tile worker thread


kkaefer:
We should check that we're not creating any GL objects or calling any GL functions in general in anything but the Map thread.

ljbade:
Hmm, easiest way is it look at the attach/detach functions since any GL calls in a thread without a GL context bound will throw and exception.

kkaefer:
FWIW, I don't think we're currently doing this, but we should do a proper review to avoid any undefined effects later on.

ljbade:
Another trick would be to add a thread ID check to the GL function error check wrappers.

jfirebaugh:
Creation of GL objects now requires a `GLObjectStore`, which the worker thread does not have access to, so we can be fairly certain this does not happen.

Issue #667, opened 2014-12-01
Remove concurrent access to objects from worker threads


kkaefer:
Currently, we have a number of objects that are shared across threads. This means we're locking these objects when accessing them. To remove lock contention, and to align the architecture with JS, we should move away from shared memory objects.

The way JS (web workers) operates is by having a read-only state of the object in question (e.g. the sprite/glyph atlas). When a write access is required, the worker thread is suspended and the operation is carried out in the main thread (e.g. adding glyphs to the glyph atlas). Then, the updated state is sent to the worker thread again.

ChrisLoer:
Fixed with #8341.

Issue #1471, opened 2015-05-08
Worker pool should share queue instead of round-robin distribution


kkaefer:
From #1455:

> 4ccf246 intensified this problem by changing our worker pool from a shared queue to an individual queue for every thread with round-robin distribution of WorkRequests. This means there can be situations where a thread's work queue gets clogged up with work items while other workers would be available to process them.

jfirebaugh:
I am not sure we want to do this, actually. Two reasons:

* JS uses round-robin, not a shared queue. This would be different.
* I want to introduce an affinity between a map thread tile and a particular worker thread, like js has, so that there can be persistent data associated with the tile on the worker thread.

kkaefer:
> This would be different.

It may be different, but it'll improve the speed of the app quite a bit by using the cores more effectively.

> I want to introduce an affinity between a map thread tile and a particular worker thread, like js has, so that there can be persistent data associated with the tile on the worker thread.

This may make sense in a JS environment where you have to copy over data to the workers, but in C++ we can move this data between threads for free, so an affinity is not required. I also think it's not desired since it may cause situations where tile distribution across threads is out of balance, requiring more work.

jfirebaugh:
In addition to assigning work to threads more efficiently, the implementation should ensure that:

* An individual worker's work is performed in the order that it was queued
* An individual worker is not active on multiple threads simultaneously

In other words, each worker should have its own serial, FIFO execution queue, but any individual item can be processed on whatever thread in the pool is available.

jfirebaugh:
The other

PR #1691, opened 2015-06-05
TileData + Worker refactor

jfirebaugh:
Cleanly separate two different kinds of data:

| Map Tile Data | Worker Tile Data |
| ------ | ------ | 
| Affiliated with the Map thread | Affiliated with one of the Worker threads |
| May include GL objects | "Plain old data" only; no GL objects (#926) |
| Owned by a `Tile` object | Owned by a `TileWorker` object |
| | May be _moved_ to map thread at key moments (e.g. after a tile parse completes) |

Data will move between map thread and worker threads via the `Thread::invoke` message passing system. No tile data will be shared between the two. (For now we'll continue to share references to `Style` and atlases, but this is another step toward removing that sharing as well.)

Targeted bugs:

* https://github.com/mapbox/mapbox-gl-native/issues/926
* https://github.com/mapbox/mapbox-gl-native/issues/667 (partially)

Plan of attack:

* Collapse `TileData` inheritance hierarchy into a fully-abstract base class and `RasterTileData`, `VectorTileData`, `LiveTileData` child classes. Note that `LiveTileData` will not inherit from `VectorTileData`. (This will introduce code duplication that will later be factored away.)
* Introduce `TileWorker` class, move worker code from `TileData` to it. This may need to be a parallel inheritance hierarchy with `RasterTileWorker`, `VectorTileWorker`, and `LiveTileWorker`. At first objects of this class will be owned by `TileData` and descendants, but eventually they will live persistently on the worker thread.
* Move the making of tile requests to `TileWorker`.
* Add smarter dispatching to `Worker`. Assign a worker thread to a tile once, and then dispatch future work requests to the same worker thread, like in gl-js.
* Migrate `TileWorker` to worker threads.

There are many known unknowns here -- best way to learn them is to start working on this and see what we run into.


tmpsantos:
> These are all the types of things that I expect to migrate into TileWorker, so I'm going to try introducing that now.

Hmm, so the `TileWorker` is not exactly a `Worker`. I got a bit confused at the beginning but after reading the code is quite clear to me, but maybe we should call it something like `TileWorkerJob`.

tmpsantos:
@jfirebaugh :+1:

jfirebaugh:
Right, TileWorker is an interim class. Once it's fully separable from the TileData hierarchy it should merge into Worker::Impl. Then we should probably rename Worker to TileWorker.

src/mbgl/map/live_tile_data.cpp
```diff
@@ -1,58 +1,57 @@
...
+    : VectorTileData(id_, style_, source_, angle_, collisionDebug_),
       annotationManager(annotationManager_) {
-    // live features are always ready
-    setState(State::loaded);
+    // live features are always loaded
+    state = State::loaded;
 }
 
 LiveTileData::~LiveTileData() {
-    // Cancel in most derived class destructor so that worker tasks are joined before
```
kkaefer: These comments are very helpful. Any reason to remove them?

jfirebaugh: My eventual plan was to have a flat hierarchy, with `LiveTileData` _not_ inheriting from `VectorTileData`, so that (among other things) this would become moot. If we end up doing that I'll add the comments back.

src/mbgl/map/live_tile_data.cpp
```diff
@@ -1,58 +1,57 @@
...
-    // any member data goes away.
     cancel();
 }
 
-void LiveTileData::parse() {
-    if (getState() != State::loaded) {
-        return;
+bool LiveTileData::reparse(Worker&, std::function<void()> callback) {
+    if (parsing.test_and_set(std::memory_order_acquire)) {
+        return false;
```
kkaefer: Can we store this as a `dirty` flag so we know that we weren't able to start parsing and have to try again some time later?

src/mbgl/map/tile_worker.cpp
```diff
@@ -1,73 +1,71 @@
...
+    const auto it = buckets.find(layer.bucket->name);
+    if (it == buckets.end()) {
+        return nullptr;
+    }
+
+    assert(it->second);
+    return it->second.get();
+}
+
+size_t TileWorker::countBuckets() const {
```
kkaefer: Can we get rid of the bucket count? TileData objects don't necessarily have a notion of buckets, e.g. the Raster source only ever has one "bucket".

jfirebaugh:
I'm currently working on moving the after callback / canceling semantics offered by `WorkTask` into the core `RunLoop` implementation -- into [`Invoker`](https://github.com/mapbox/mapbox-gl-native/blob/f4707f72e0ce7f1c40337c72ae5c0c0c8ef0efed/src/mbgl/util/run_loop.hpp#L70) specifically.

This involves adding an additional template argument to `Invoker` which is instantiated with the callback type, either `std::function<void (R)>` for some type R if the invoked method returns a result, or `std::function<void ()>` if it does not. I then want to overload the `invoke` member function with implementations for these two cases:

```
        template <std::size_t... I>
        void invoke(std::index_sequence<I...>) {
            func(std::forward<Args>(std::get<I>(params))...);
            callback();
        }

        template <std::size_t... I>
        void invoke(std::index_sequence<I...>) {
            callback(func(std::forward<Args>(std::get<I>(params))...));
        }
```

I need to use some kind of `enable_if` or [SFINAE](http://en.cppreference.com/w/cpp/language/sfinae) to resolve the overloading but I'm stuck on that. Basically I want the first definition to be used for `std::function<void ()>` and the second one for `std::function<void (R)>`.

kkaefer:
> Assign a worker thread to a tile once, and then dispatch future work requests to the same worker thread, like in gl-js.

Why is this necessary?

jfirebaugh:
As you pointed out in https://github.com/mapbox/mapbox-gl-native/issues/1471#issuecomment-109190596, it's probably not, and I agree.

src/mbgl/map/live_tile_data.cpp
jfirebaugh: I don't fully understand partial parsing yet but I assume this is handled by the return value here and the code in `Style::handlePartialTile`.

src/mbgl/map/tile_worker.cpp
jfirebaugh: This is a question for @tmpsantos I think. `countBuckets` use used in the logic for emitting a "tile loaded" event.

tmpsantos: `countBuckets` is used to find out if, after reparsing a partial tile, anything has changed and we need to fire a "tile data changed" event. We can get rid of it after we map pending resources to tile data, because every reparse will always be caused by a resource that the tile data will be waiting for.

jfirebaugh:
This branch is currently hanging on headless rendering tests. It has something to do with reparsing partial tiles, but the data flow there is really hard to follow. @tmpsantos do you think you could take a look at it?

tmpsantos:
> This branch is currently hanging on headless rendering tests. It has something to do with reparsing partial tiles, but the data flow there is really hard to follow. @tmpsantos do you think you could take a look at it?

On it.

kkaefer:
@tmpsantos I would love the partial parsing to be refactored. I'm thinking we should create a list of all layers that need parsing when we start the parse step. In the parse thread, we iterate through that list, parse, then remove from that list when we've parsed a layer. This means that at some point, this list will be empty and we know that everything's been parsed. This is likely less error prone than using the total bucket count as a measure for whether we need to parse again. This also means that we can start removing the `state` property.

kkaefer:
The long term goal is to split up parsing per layer so that we could ideally parse one vector tile with multiple threads. This requires making the task of parsing a layer from a tile independent of order and ideally has no concurrent access. This means we'd have to give every layer their own buffers and not do label placement until we actually have the label layers parsed.

PR #1694, opened 2015-06-09
Convert point annotations to RTree instead of pre-creating tiles for all zoom levels

kkaefer:
depends on #1695

### DIFF
--- a/src/mbgl/annotation/point_annotation_layer.cpp
+++ b/src/mbgl/annotation/point_annotation_layer.cpp
@@ -1,0 +1,82 @@
+#include <mbgl/annotation/point_annotation_layer.hpp>
+
+#pragma GCC diagnostic push
+#pragma GCC diagnostic ignored "-Wunused-function"
+#pragma GCC diagnostic ignored "-Wunused-parameter"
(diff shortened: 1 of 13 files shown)


kkaefer:
@incanus, is there any reason we're using integer IDs rather than wrapping a shared_ptr to track annotations for later removal?

incanus:
> is there any reason we're using integer IDs rather than wrapping a shared_ptr to track annotations for later removal?

If annotation adds/removes are synchronous, why bother with the complexity? basically

See also how we map them Cocoa-side, which I suppose could be adapted to live pointers: 

https://github.com/mapbox/mapbox-gl-native/blob/cfee5aa8ef73b5bd2050bd68d584846192ebf0a2/platform/ios/MGLMapView.mm#L79

kkaefer:
>  why bother with the complexity? basically

We need to maintain a separate mapping of ID to Annotation so we can remove the correct elements from the RTree.

kkaefer:
@incanus can you review please?

incanus:
We're not hitting this as part of b3.

jfirebaugh:
⇢ #2673

Issue #1462, opened 2015-05-07
Use the defaultStyles constant on Android apps


tmpsantos:
The MacOS and Linux app are now using the defaultStyles as the list of styles we can select from. The iOS and Android apps should use the same list.

defaultStyles was added at https://github.com/mapbox/mapbox-gl-native/pull/1443

/cc @incanus

jfirebaugh:
`default_styles.hpp` should be a header in `include/mbgl/platform` so it can be included via `#include <mbgl/platform/default_styles.hpp>`. There shouldn't be relative includes [like this](https://github.com/mapbox/mapbox-gl-native/blob/606e2c738321682730833d2fe329f981b6e22b6e/linux/main.cpp#L2-2) in the test/sample apps.

kkaefer:
> There shouldn't be relative includes like this in the test/sample apps.

Why not? My reason for not moving it into the `include/mbgl/platform` is that it's technically not part of the library.

jfirebaugh:
I think it should be considered part of the library. If a third-party wanted to build an app with style-switching functionality, they should `#include <mbgl/platform/default_styles.hpp>` and then use it.

kkaefer:
The bundled styles aren't technically part of the library either; they're just there because we package them with the apps.

incanus:
They _are_ part of the library on iOS — we package a `MapboxGL.bundle` with the static library and headers which contains resources like the compass image and Mapbox logo as well as the included styles. It's a first-class part of the library. Currently we recommend users [query for the included styles with an API call](https://www.mapbox.com/mapbox-gl-ios/#map_style_changes), which we hook up by [iterating the filesystem directly](https://github.com/mapbox/mapbox-gl-native/blob/4798c502aa2a8c5e5c18193e1a221a8a371f36ab/platform/ios/MGLMapView.mm#L1528-L1544). This should be a first-class API internal to `mbgl` that bindings like iOS can query.

ljbade:
Hmm good idea @incanus.

incanus:
Punting off of b2 since the bundled styles haven't changed since b1 and nothing under the hood has moved here that makes this any more urgent.

ljbade:
@1ec5 is looking to implement this in https://github.com/mapbox/mapbox-gl-native/pull/2746#issuecomment-150703017 for iOS.

On Android I decided to just have a class that provides static constants in the `Styles` class.

tobrun:
On Android we have:

```java
[...]

PR #1709, opened 2015-06-11
Fix hardcoded Styles in iOS sample-app

frenos:
The current sample for iOS uses a hardcoded array for styles and because of that offers styles not supported anymore.
I fixed that by using bundledStyleURLs which always returns a current list of available styles.
To keep the UI intact and reasonable pretty I extract a name from the filename.


jfirebaugh:
Thanks for the PR. However, the direction we want to go here is to use [`defaultStyles`](https://github.com/mapbox/mapbox-gl-native/blob/master/platform/default/default_styles.hpp) (#1462) rather than `bundledStyleURLs`. `defaultStyles` provides both the style URL and human-readable name.

PR #1710, opened 2015-06-11
Use array of structs rather than parallel arrays for annotations

kkaefer:
Our current point annotations API uses two parallel vectors for points and icon names which leaves room for errors due to mismatched annotations. Instead, we should use an array of structs so the association between point and icon is fixed and not through the index.


1ec5:
:+1:

1ec5:
@kkaefer, per chat with @incanus, I reverted this change in 714b68f7b325c30d0646c8b237d65b013fb12f04, because it complicated landing #1655. I like the idea of making annotations more object-oriented at the mbgl level, but that’s a lower priority than landing shape annotations.

PR #1713, opened 2015-06-11
Use Mason modules for Variant, RapidJSON and Google Test

kkaefer:
We currently have all of this included into the main repository, but they should be downloaded from Mason instead.


jfirebaugh:
Stale. Fresh commits / PRs:

* variant: b46c102eb0ec3ca2ff0b422831cb98fa2bdebcba
* RapidJSON: #2476
* gtest: #2477
* GeoJSON-VT: #2352

PR #1718, opened 2015-06-12
Namespace Reachability notification constant

### LINKED ISSUE #1717: Rename kReachabilityChangedNotification
This is a popular class to include in client apps. By including it in the libMapboxGL.a , but not including the header in the Headers directory prevents client app from using it. Also, if a client tries to include its own version of reachability, duplicate symbols build errors are the result.

1ec5:
This PR renames the Reachability library’s notification constant so that clients can use their own copies of Reachability.

Fixes #1717.


picciano:
Thanks!

PR #1738, opened 2015-06-16
Defer to MGLMultiPoint for overlayBounds implementation

### LINKED ISSUE #1730: -[MGLOverlay overlayBounds] is always (0, 0), (0, 0)
With #1655 (bd0bf294bf1e350b184e54a824430c911dd66b9b), each class that conforms to MGLOverlay `@synthesize`s the required `overlayBounds` property but never sets it. Each of these classes should instead compute the `overlayBounds`, either at initialization or lazily, based on the overlay’s coordinates.

/cc @incanus

1ec5:
The synthesized `overlayBounds` shadows the MGLMultiPoint implementation with one that reads an `_overlayBounds` ivar. This PR causes the subclasses to look at MGLMultiPoint’s implementation.

Fixes #1730.

/cc @incanus

### DIFF
--- a/platform/ios/MGLPolygon.m
+++ b/platform/ios/MGLPolygon.m
@@ -4,7 +4,7 @@
 
 @implementation MGLPolygon
 
-@synthesize overlayBounds;
+@dynamic overlayBounds;
 
 + (instancetype)polygonWithCoordinates:(CLLocationCoordinate2D *)coords
                                  count:(NSUInteger)count
--- a/platform/ios/MGLPolyline.m
+++ b/platform/ios/MGLPolyline.m
@@ -4,7 +4,7 @@
 
 @implementation MGLPolyline
 
-@synthesize overlayBounds;
+@dynamic overlayBounds;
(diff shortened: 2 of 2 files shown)


incanus:
I think this was my original intention and I just got a wire crossed. Let's sit for a bit as we gauge stability but I think this will be fair game to merge.

1ec5:
The documentation comment for `overlayBounds` currently says:

> Implementers of this protocol must set this area when implementing their overlay class, and after setting it, you must not change it.

Is that accurate? Is our intent even to match MapKit here?

incanus:
Yep, we do this here: 

https://github.com/mapbox/mapbox-gl-native/blob/416928f824b97f33a7acc88d556ddaace78234ea/platform/ios/MGLMultiPoint.mm#L26

1ec5:
Oh, I read that comment as meaning that `overlayBounds` is a property, privately writeable, that implementers would set in the initializer.

Issue #1650, opened 2015-05-27
create iOS download without symbols


incanus:
Not everyone is used to a large static library (300MB+ currently) which includes symbols, since most frameworks aren't open source. 

https://github.com/mapbox/mapbox-gl-native/issues/1234, https://github.com/mapbox/mapbox-gl-native/issues/1633

We should push dual downloads of each release, one as we do currently and one with debug symbols stripped. Maybe something like: 

* `mapbox-gl-ios-X.Y.Z.zip`
* `mapbox-gl-ios-symbols-X.Y.Z.zip`

It's still the `Release` build configuration both ways.

incanus:
Needs some testing, but followed the above scheme — two different release products: 

* `mapbox-gl-ios-X.Y.Z.zip`
* `mapbox-gl-ios-symbols-X.Y.Z.zip`

incanus:
Latest sizes: 

* Unstripped, device & sim: 397.1MB
* Unstripped, sim only: 161.7MB
* Stripped, device & sim: 20MB

Big improvement there. I _think_ that means that we increase the size of apps by 20MB, though, in the best case, but need to confirm.

incanus:
* Basic Objective-C app from Xcode template: 28KB `.ipa`
* Same app with unstripped GL integrated showing map view: 1.1MB `.ipa`
* Same app with stripped GL integrated and showing map view: 1.1MB `.ipa`

So, significant increase but totally acceptable in both cases considering OTA apps are limited to 100MB. The end app strips during its archive process. 

Really this comes down to the question of whether we make the download/install process easier for devs or not, since the end app won't bloat if we leave symbols in. 

Either way, we should document this fact so people know what to expect in their app sizes.

incanus:
Publish process releases two packages per https://github.com/mapbox/mapbox-gl-native/issues/1650#issuecomment-112231647 and we'll add both to GitHub releases, which gets linked from the [binary install process](https://www.mapbox.com/mapbox-gl-ios/#binary) directly.

incanus:
Note that we'll have to do stuff like 7bef7b310777c41889aeb05ac0d9a2301dfc67da in order to use symbols-laden builds for internal testing apps.

PR #1746, opened 2015-06-17
Throw NSExceptions on client errors

1ec5:
Errors caused by client code should be communicated by raising `NSException`, not by calling `NSAssert()`. Assertions are intended for internal usage. Because they include helpful information like the absolute path to the file containing the assertion, stripping the build (#1650) also strips assertion messages.

Also replaced hardcoded references to methods in these messages with `__PRETTY_FUNCTION__`, which avoids future copy-pasta and handles cases in which the MGL class has been subclassed.

@incanus @friedbunny, please review wording and double-check that I’ve correctly inverted all the conditionals, since we’re no longer asserting.

### DIFF
--- a/platform/ios/MGLMapView.mm
+++ b/platform/ios/MGLMapView.mm
@@ -145,13 +145,20 @@ std::chrono::steady_clock::duration secondsAsDuration(float duration)
 
 - (nullable NSString *)accessToken
 {
-    NSAssert(NO, @"-[MGLMapView accessToken] has been removed. Use +[MGLAccountManager accessToken] or get MGLMapboxAccessToken from the Info.plist.");
+    [NSException raise:@"Method unavailable" format:
+     @"%s has been removed. "
+     @"Use +[MGLAccountManager accessToken] or get MGLMapboxAccessToken from the Info.plist.",
+     __PRETTY_FUNCTION__];
     return nil;
 }
 
 - (void)setAccessToken:(nullable NSString *)accessToken
 {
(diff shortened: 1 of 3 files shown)


incanus:
I will do some testing of this against `make ipackage` in https://github.com/mapbox/mapbox-gl-native/commit/73737445ec0333b701b64613d8b92e18f45ea314, too.

incanus:
This doesn't seem to have any effect over the use of `NSAssert`. 

screen shot 2015-06-17 at 3 36 37 pm

The way to test here is: 

1. `make package-strip`
1. Use the resulting library, headers, and bundle in a test app. [Here's mine](https://dl.dropboxusercontent.com/u/575564/sizetester.zip). 
1. Launch the app without setting a token.

1ec5:
> This doesn't seem to have any effect over the use of `NSAssert`.

This is an unrelated issue covered by #1748.

incanus:
So... do we still need to do these changes at all? Assuming we fix #1748.

1ec5:
It'd be lower priority but I think it's still more correct than what we have now. NSException is intended for telling others they made mistakes, while NSAssert is for slapping your own forehead. That's why I left the main thread assertions as assertions.

1ec5:
Does a stripped build still have the assertion messages, then?

incanus:
Correct @1ec5, I think before I may have been breaking on exceptions and that was throwing me. Whatever it is, the `0.4.0` release branch as it stands right now, stripped, produces useful messages: 

screen shot 2015-06-18 at 10 48 36 am

I think it's safe to punt this to b3 then for now? i.e. merge to `master` but not release branch.

1ec5:
The missing access token was never covered by either an NSAssert or an NSException. To verify that stripping retains the error message, you need to do something else, like forget to implement the Metrics opt-out.

1ec5:
I’ve verified that the stripped library still contains the full assertion messages. The exception name is “NSInternalInconsistencyException” instead of something more descriptive, but the exception description is there. So let’s put this on beta 3.

incanus:
That looks good also (re-added token, removed `.plist` line): 

screen shot 2015-06-18 at 10 57 50 am

PR #1388, opened 2015-06-18
don't use SIGSTOP/SIGCONT for flow control in tests

tmpsantos:
SIGSTOP and SIGCONT are being used to ensure that the server spawned by the headless and storage tests is ready. But they have several weird side effects:

* Debuggers break on SIGSTOP by default, so you have to continue several times if you are debugging the tests.
* SIGSTOP is a shell job control mechanism, so the test process winds up running as a background process in the shell.

### DIFF
--- a/test/fixtures/util.cpp
+++ b/test/fixtures/util.cpp
@@ -4,27 +4,51 @@
 
 #include <csignal>
 
+#include <unistd.h>
+
 namespace mbgl {
 namespace test {
 
 pid_t startServer(const char *executable) {
     const std::string parent_pid = std::to_string(getpid());
+    int pipefd[2];
+
+    if (pipe(pipefd)) {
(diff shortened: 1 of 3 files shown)


tmpsantos:
> Agreed. Do you have a suggestion for an alternative mechanism?

Pipes!

jfirebaugh:
:clap:

jfirebaugh:
@tmpsantos Is it possible to extend the use of pipes to have the child exit automatically when the parent exits? (In order to fix 3a86321aeecbf4acb2c0b6040d1f332b616a823a properly?)

tmpsantos:
> @tmpsantos Is it possible to extend the use of pipes to have the child exit automatically when the parent exits? (In order to fix 3a86321 properly?)

Yes, it is. @kkaefer is doing it on #3432

PR #1768, opened 2015-06-22
Style/ResourceLoading.RequestWithCorruptedData/3 test is flaky

tmpsantos:
See: https://travis-ci.org/mapbox/mapbox-gl-native/jobs/67502670#L1743

### DIFF
--- a/test/style/resource_loading.cpp
+++ b/test/style/resource_loading.cpp
@@ -13,6 +13,8 @@
 #include <mbgl/util/texture_pool.hpp>
 #include <mbgl/util/thread.hpp>
 
+#include <regex>
+
 using namespace mbgl;
 
 namespace {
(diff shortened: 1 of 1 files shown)


tmpsantos:
Merged.

kkaefer:
I'm not convinced that this is the right solution. Are you making the test case more lax so that it doesn't fail? If you think this is the right solution, can you explain why we get different values we need to account for?

tmpsantos:
> I'm not convinced that this is the right solution. Are you making the test case more lax so that it doesn't fail? If you think this is the right solution, can you explain why we get different values we need to account for?

Previously the test case was checking specifically for an error message with parsing error on the tile `[15/16384/16384]`. This was causing flakiness because we request 4 tiles and the `MockFileSource` sets corrupted data on all of them.

Since parsing is done in a worker thread, sometimes the tile say `[15/16384/16383]` was being parsed first, reporting the parsing error and the others getting canceled at the request stage. So we wouldn't have the error message we were looking for although what was being tested worked.

PR #1778, opened 2015-06-23
Style parser readability

brunoabinader:
I took a moment to understand what the `bool` being passed on the `std::tuple` means. Replacing `bool` values with `enum` is a good way to improve code readability without damaging performance. Found a minor issue along the way causing `parseProperty` to return failure when `T = std::array<float, 2>`.


jfirebaugh:
:+1:

Issue #1781, opened 2015-06-23
Allow watermark and info button to be moved (iOS)


parrots:
Given visual effect views in iOS 8+, sometimes maps need to extend below a UIView to affect the view's background color dynamically. Unfortunately this hides the required watermark in some instances. Even though the intent is non-malicious, it causes a violation of the TOS.

Users need to somehow configure where the logo is shown, even if it's just something as simple as an offset from the bottom.

Example:
![Sample](http://f.cl.ly/items/113N17131Z3Z1s1a3G0j/Simulator%20Screen%20Shot%20Jun%2023,%202015,%204.31.38%20PM.png)

1ec5:
MGLMapView currently knows to offset the ornaments (including the compass, obscured above by the 3D button) to make room for standard top and bottom bars. It would be reasonable to add an additional offset for custom views like this.

incanus:
I wonder if if we can be more general about it and expose connecting to the top/bottom layout guide-related stuff that we have now (e.g. [here](https://github.com/mapbox/mapbox-gl-native/blob/43d69b6a65a3ee05f41ed13ac5bf2ca943075e65/platform/ios/MGLMapView.mm#L504-L511) is the compass' constraint against the top layout guide)).

1ec5:
@incanus, you mean like being able to make the logo stick to the top of the view instead of the bottom, or like calling an optional delegate method to do the dirty work of setting up the constraints?

incanus:
No, more like exposing the current constraints to allow for things like Curtis’ custom view to push the bottom views up, for example.

picciano:
Once maps start being integrated into the design of production apps, what is needed is complete control over the placement of the overlays (logos, compass, etc.) Pretty much every layout we have in our app does not work with the given overlay locations. I think it's more than pushing an element up or down, but rather being able to fully customize the layout. At this point, we removed them all and will add back the required attribution in a different way.

![ios simulator screen shot jun 23 2015 3 37 53 pm](https://cloud.githubusercontent.com/assets/417933/8318021/b4d72274-19be-11e5-8765-d9cac6455cec.png)
![ios simulator screen shot jun 23 2015 3 41 21 pm](https://cloud.githubusercontent.com/assets/417933/8318022/b4d86800-19be-11e5-9b40-6f327d6c50e5.png)
[...]

PR #1783, opened 2015-06-23
Fix fit to bounds

### LINKED ISSUE #1092: allow region-based viewport setting

### LINKED ISSUE #1256: iOS: Zooming out beyond ~z10 with tracking mode enabled will force zoom in, reset tracking

### LINKED ISSUE #1433: [MGLMapView zoomToSouthWestCoordinate:northEastCoordinate:] not coordinates deterministic
Using [`[MGMapView zoomToSouthWestCoordinate:northEastCoordinate:animated:]`](https://github.com/mapbox/mapbox-gl-native/blob/acd27fc386ba413fbec4cac75d4407eb77712883/platform/ios/MGLMapView.mm#L1379-L1393) produces a different result based on the zoom/scale of the map.

I expected the method to always fit the given coordinates on the screen — instead, it zooms in on their center point at a zoom level calculated based on `_mbglMap->getScale()`. This often results in `z18` and never actually fits the viewport to the exact coordinate bounds.

Calling the method repeatedly with the same coordinates but at different zooms yields this:

```
Mapbox GL: input-zoom: 13.45   scale: 11161.71     output-zoom: 18.00
Mapbox GL: input-zoom: 15.78   scale: 56414.93     output-zoom: 18.00
Mapbox GL: input-zoom: 3.69    scale: 12.91        output-zoom: 9.18
Mapbox GL: input-zoom: 9.18    scale: 578.48       output-zoom: 14.66
Mapbox GL: input-zoom: 14.66   scale: 25918.91     output-zoom: 18.00
```

`[MGMapView zoomToSouthWestCoordinate:northEastCoordinate:animated:]` is only used right now for centering the map on the user location annotation and is somewhat broken for that purpose, too, besides not being a public method. (#1092 tracks what to do about this API decision.)

I would like to use this method for bounds-based syncing in [the demo app I'm working on](https://github.com/friedbunny/treble), so I've hacked together an interim solution, but the maths are absolutely wrong:

```objc
- (void)fitBoundsToSouthWestCoordinate:(CLLocationCoordinate2D)southWestCoordinate northEastCoordinate:(CLLocationCoordinate2D)northEastCoordinate animated:(BOOL)animated
{
    CLLocationCoordinate2D center = CLLocationCoordinate2DMake((northEastCoordinate.latitude + southWestCoordinate.latitude) / 2, (northEastCoordinate.longitude + southWestCoordinate.longitude) / 2);
    
    CGFloat scaleX = _mbglMap->getWidth() / (northEastCoordinate.longitude - southWestCoordinate.longitude);
    CGFloat scaleY = _mbglMap->getHeight() / (northEastCoordinate.latitude - southWestCoordinate.latitude);
    CGFloat minZoom = _mbglMap->getMinZoom();
    CGFloat maxZoom = _mbglMap->getMaxZoom();
    CGFloat hack = ([[UIApplication sharedApplication] statusBarOrientation] == UIInterfaceOrientationPortrait) ? 0.5f : 1.f;
    CGFloat zoomLevel = log(MIN(scaleX, scaleY)) / log(2) - hack;
    zoomLevel = MAX(MIN(zoomLevel, maxZoom), minZoom);
    
    [self setCenterCoordinate:center zoomLevel:zoomLevel animated:animated];
}
```

I put together a demonstration of this bug and my hack [in a branch here](https://github.com/mapbox/mapbox-gl-native/compare/master...friedbunny:1433-fit-bounds). Use the action sheet's "Zoom to Oregon" option to see new default behavior.

/cc @incanus @1ec5

### LINKED ISSUE #1775: Expose MGLMapView's zoomToSouthWestCoordinate:northEastCoordinate:animated: method in public header
Is there a reason for not exposing `MGLMapView`'s `zoomToSouthWestCoordinate:northEastCoordinate:animated:` method publicly? It's very useful to have.

(As an aside, being able to also pass a `UIEdgeInsets` to pad the zoomed area would be amazingly useful!)

1ec5:
This PR reimplements “fit to bounds” functionality to take advantage of MGLCoordinateBounds and additionally exposes the method publicly. I started by re-porting [`Camera.prototype.fitBounds()`](https://github.com/mapbox/mapbox-gl-js/blob/37145b8b9bb0de022811ad46f3e89a2714896b18/js/ui/camera.js#L345) in mapbox/mapbox-gl-js to ensure correct baseline behavior. Then I wrote some logic for edge insets and rotation. Along for the ride, I added a bunch of public utility functions related to MGLCoordinateBounds.

* [x] Refactor the method to use MGLCoordinateBounds
* [x] Derive center and zoom level from projected pixels (to fix #1433)
* [x] Expose the method publicly (to fix #1092)
* [x] Push the logic down into mbgl
* [x] Stop resetting the user tracking mode (to fix #1256)
* [x] Account for rotation (ref mapbox/mapbox-gl-js#1338)
* [x] ~~Adjust the viewport to account for top and bottom layout guides~~
* [x] Take an `edgePadding` parameter for parity with MapKit and mapbox-gl-js (to fix #1775)
* [x] Test “round-tripping” the visible bounds

@friedbunny @incanus, can you try out this implementation and see if there are any other edge cases that MapKit and Google Maps handle differently?

### DIFF
--- a/gyp/platform-ios.gypi
+++ b/gyp/platform-ios.gypi
@@ -36,6 +36,8 @@
         '../platform/ios/MGLUserLocationAnnotationView.m',
         '../include/mbgl/ios/MGLTypes.h',
         '../platform/ios/MGLTypes.m',
+        '../include/mbgl/ios/MGLGeometry.h',
+        '../platform/ios/MGLGeometry.m',
         '../include/mbgl/ios/MGLMultiPoint.h',
         '../platform/ios/MGLMultiPoint_Private.h',
         '../platform/ios/MGLMultiPoint.mm',
--- a/include/mbgl/ios/MGLGeometry.h
+++ b/include/mbgl/ios/MGLGeometry.h
@@ -1,0 +1,85 @@
+#pragma once
+
+#import "MGLTypes.h"
+
+#import <CoreLocation/CoreLocation.h>
+
+NS_ASSUME_NONNULL_BEGIN
+
+/** Defines the area spanned by an MGLCoordinateBounds. */
+typedef struct {
+    CLLocationDegrees latitudeDelta;
+    CLLocationDegrees longitudeDelta;
+} MGLCoordinateSpan;
+
+/** Creates a new MGLCoordinateSpan from the given latitudinal and longitudinal deltas. */
+NS_INLINE MGLCoordinateSpan MGLCoordinateSpanMake(CLLocationDegrees latitudeDelta, CLLocationDegrees longitudeDelta) {
+    MGLCoordinateSpan span;
+    span.latitudeDelta = latitudeDelta;
+    span.longitudeDelta = longitudeDelta;
+    return span;
+}
+
+/** Returns `YES` if the two coordinate spans represent the same latitudinal change and the same longitudinal change. */
+NS_INLINE BOOL MGLCoordinateSpanEqualToCoordinateSpan(MGLCoordinateSpan span1, MGLCoordinateSpan span2) {
+    return (span1.latitudeDelta == span2.latitudeDelta &&
+            span1.longitudeDelta == span2.longitudeDelta);
+}
+
+/** An area of zero width and zero height. */
+extern const MGLCoordinateSpan MGLCoordinateSpanZero;
+
+/** A rectangular area as measured on a two-dimensional map projection. */
+typedef struct {
+    CLLocationCoordinate2D sw;
+    CLLocationCoordinate2D ne;
+} MGLCoordinateBounds;
+
+/** Creates a new MGLCoordinateBounds structure from the given southwest and northeast coordinates. */
+NS_INLINE MGLCoordinateBounds MGLCoordinateBoundsMake(CLLocationCoordinate2D sw, CLLocationCoordinate2D ne) {
+    MGLCoordinateBounds bounds;
+    bounds.sw = sw;
+    bounds.ne = ne;
+    return bounds;
+}
+
+/** Returns `YES` if the two coordinate bounds are equal to each other. */
+NS_INLINE BOOL MGLCoordinateBoundsEqualToCoordinateBounds(MGLCoordinateBounds bounds1, MGLCoordinateBounds bounds2) {
+    return (bounds1.sw.latitude == bounds2.sw.latitude &&
+            bounds1.sw.longitude == bounds2.sw.longitude &&
+            bounds1.ne.latitude == bounds2.ne.latitude &&
+            bounds1.ne.longitude == bounds2.ne.longitude);
+}
+
+/** Returns the area spanned by the coordinate bounds. */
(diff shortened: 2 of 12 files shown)


1ec5:
> Adjust the viewport to account for top and bottom layout guides

In light of #1781, maybe it doesn’t make sense for MGLMapView to handle this itself. What does MKMapView do with top and bottom bars?

jfirebaugh:
I'd like to move the core native API towards gl-js's `Camera`. How hard would it be to push the calculations down into GL core via `Map::fitBounds()`?

1ec5:
> How hard would it be to push the calculations down into GL core via `Map::fitBounds()`?

Done in ed999048b631d34020bdbc28f8b8828d798b9158. That was pretty easy, actually, because I was relying on mbgl::Map instead of the view for all the calculations. That change also neatly avoids an erroneous call to `-setCenterCoordinate:animated:` instead of `-setCenterCoordinate:animated:preservingTracking:`, so that fixes #1256 (and obviates part of #1257).

friedbunny:
This works well so far. I'll test again once the layout guides adjustments are in, that's the biggest difference between MapKit/Google and us right now.

1ec5:
Note that fitting to bounds forcefully resets the direction back to north, without animation.

Also, in 31d291898416571a1aae921b96e3e730b98c20cd, I adapted the fit to bounds method to account for top and bottom bars, but it appears that MapKit does not. So I’m probably going to roll back that change. It’s easy enough to handle manually in client code.

1ec5:
If you try to fit an MKMapView to any empty rect, MKMapView zooms way into the top-left corner of the map (in the Arctic Sea, above the Bering Strait). With this PR, however, MGLMapView centers on the empty coordinate bounds (which may originate anywhere) – I think that’s better behavior.

1ec5:
The fit to bounds method now respects the current rotation. That turned out to require a lot less trig than I feared. :smiley:

friedbunny:
4435752_1431360735 7466_updates

PR #1786, opened 2015-06-24
Remove mbgl::Environment

tmpsantos:
After the threading refactoring this class is no longer need and we can scope the thread globals to a ThreadContext class living in the thread local storage.


tmpsantos:
@jfirebaugh please review

jfirebaugh:
LGTM. Last time I tried to use `thread_local` there was some platform I struggled with and we switched to `uv::tls` instead. If you can get `thread_local` working cross-platform we should remove `uv::tls`.

springmeyer:
thread_local is unsupported on OS X (https://github.com/mapnik/python-mapnik/issues/22)

mikemorris:
Whoa, neat!

src/mbgl/map/environment.cpp
```diff
@@ -1,174 +0,0 @@
...
-    abandonedVAOs.emplace_back(vao);
-}
-
-void Environment::abandonBuffer(uint32_t buffer) {
-    assert(currentlyOn(ThreadType::Map));
-    abandonedBuffers.emplace_back(buffer);
-}
-
-void Environment::abandonTexture(uint32_t texture) {
-    assert(currentlyOn(ThreadType::Map));
```
kkaefer: The new code for this may exist per thread, whereas the previous code only existed on one thread and `assert`ed that it wasn't called from another thread. Can we reintroduce that behavior please?

PR #1789, opened 2015-06-25
Reverse direction to be clockwise

### LINKED ISSUE #1780: Rotation based on heading going opposite of Google / Apple (iOS)
When setting the direction in MapBox GL to a specific heading it seems to rotate in a direction opposite Google Maps and Apple. Meaning at 0 and 180 the maps line up but at 90 MapBox is at an Apple equivalent to 270.

Probably just need to reverse the direction you start to rotate based on a heading.

Video here: http://f.cl.ly/items/0w3K2a3C0T1R2m1K1D0m/Rotation.mov

1ec5:
This PR reverses the direction of `direction` to be clockwise instead of counterclockwise, for consistency with MapKit, the Google Maps SDK, mapbox/mapbox-gl-js, and these methods’ own documentation.

Fixes #1780.

### DIFF
--- a/platform/ios/MGLMapView.mm
+++ b/platform/ios/MGLMapView.mm
@@ -1463,7 +1463,7 @@ std::chrono::steady_clock::duration secondsAsDuration(float duration)
 
 - (CLLocationDirection)direction
 {
-    double direction = _mbglMap->getBearing() * -1;
+    double direction = _mbglMap->getBearing();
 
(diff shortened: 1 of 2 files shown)


1ec5:
I’m going crazy – compasses are hard! :wink: Rebasing on d74b6dc12e80deff048318efdeb95834864d261b, which passed.

PR #1793, opened 2015-06-26
Avoid removing content size constraints

### LINKED ISSUE #1779: Storyboard view hierarchy can lead to squashed subviews
Setting your storyboard up with the map view at the base level of the hierarchy like this:

screen shot 2015-06-23 at 11 09 24 am

Leads to the attribution button and logo bug being squashed after the initial load — i.e., load the MBGL view controller, switch to a different view controller, come back to MBGL and this happens:

ios simulator screen shot jun 23 2015 11 10 27 am

Work around is to place the map view within another view, remembering to reset to suggested constraints:

screen shot 2015-06-23 at 11 19 34 am

/cc @1ec5

1ec5:
This PR avoids removing content size constraints that are needed to maintain the intrinsic size of the logo and :information_source:. Also, on iOS 8.0+, we now use `+[NSLayoutConstraint activateConstraints:]` and `+deactivateConstraints:` to efficiently prevent redundant constraints from lingering.

Fixes #1779.

/cc @friedbunny @incanus

### DIFF
--- a/platform/ios/MGLMapView.mm
+++ b/platform/ios/MGLMapView.mm
@@ -69,7 +69,9 @@ CLLocationDegrees MGLDegreesFromRadians(CGFloat radians)
 @property (nonatomic) NSOperationQueue *regionChangeDelegateQueue;
 @property (nonatomic) UIImageView *compass;
 @property (nonatomic) UIImageView *logoBug;
+@property (nonatomic) NS_MUTABLE_ARRAY_OF(NSLayoutConstraint *) *logoBugConstraints;
 @property (nonatomic) UIButton *attributionButton;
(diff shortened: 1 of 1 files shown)


robipresotto:
EXC_BAD_ACCESS

[NSLayoutConstraint deactivateConstraints:self.logoViewConstraints];

all the time if you are using singleton for the map view

1ec5:
@robipresotto, please open a new issue regarding this crash. If possible, please provide more information about the environment, for example the verisons of iOS and Mapbox GL, device model, and whether you’re creating the map view programmatically or with a storyboard. Thanks!

1ec5:
Also, please set an Objective-C exception breakpoint and provide the stack trace at the point of the exception being thrown.

PR #1787, opened 2015-06-29
reapply #1710 - array of structs rather than parallel arrays for annotations

jfirebaugh:
#1710 got reverted in 714b68f7b325c30d0646c8b237d65b013fb12f04 due to merge difficulties at deadline. We should now reapply it.


jfirebaugh:
CC @kkaefer @incanus for review.

Issue #1680, opened 2015-06-03
gpus_ReturnNotPermittedKillClient


robipresotto:
````ruby
Thread : Crashed: Map
0  libGPUSupportMercury.dylib     0x2e21e8a6 gpus_ReturnNotPermittedKillClient + 9
1  libGPUSupportMercury.dylib     0x2e21f363 gpusSubmitDataBuffers + 110
2  IMGSGX543GLDriver              0x25eea7c1 glrSGXRenderVertexArray + 1028
3  GLEngine                       0x29e7571d glDrawArrays_ES2Exec + 204
4  Embark                         0x00393a39 mbgl::Source::drawClippingMasks(mbgl::Painter&) (source.cpp:189)
5  Embark                         0x003ae355 mbgl::Painter::drawClippingMasks(std::__1::set<mbgl::Source*, std::__1::less<mbgl::Source*>, std::__1::allocator<mbgl::Source*> > const&) (painter_clipping.cpp:26)
6  Embark                         0x003ac825 mbgl::Painter::render(mbgl::Style const&, mbgl::TransformState, std::__1::chrono::time_point<std::__1::chrono::steady_clock, std::__1::chrono::duration<long long, std::__1::ratio<1ll, 1000000000ll> > >) (forward_list:540)
7  Embark                         0x00390173 mbgl::MapContext::render() (map_context.cpp:228)
8  Embark                         0x0038df55 std::__1::packaged_task<void ()>::operator()() (future:1999)
9  Embark                         0x003f166f mbgl::util::RunLoop::process() (deque:2553)
10 Embark                         0x004253cb uv__async_event (async.c:80)
11 Embark                         0x00425639 uv__async_io (async.c:156)
12 Embark                         0x0042c2c5 uv__io_poll (kqueue.c:233)
13 Embark                         0x00425a03 uv_run (core.c:317)
14 Embark                         0x0038e88f void mbgl::util::Thread<mbgl::MapContext>::run<std::__1::tuple<mbgl::View&, mbgl::FileSource&, mbgl::MapData&>, 0ul, 1ul, 2ul>(std::__1::tuple<mbgl::View&, mbgl::FileSource&, mbgl::MapData&>&&, std::__1::integer_sequence<unsigned long, 0ul, 1ul, 2ul>) (uv_detail.hpp:55)
[...]

Issue #1509, opened 2015-05-11
gpus_ReturnGuiltyForHardwareRestart crash (consistently reproducible).


mb12:
I had filed a bug for this issue earlier (Issue #1198 https://github.com/mapbox/mapbox-gl-native/issues/1198). At that time I did not have consistent steps to reproduce the issue. I've finally isolated the problem to a very trivial style file and a vector source that consistently reproduces the problem. 
This crash is consistently reproducible with the following configuration:
1. iPhone 5 running iOS 8.1.3 (It reproduces only on iPhone 5 and this OS version). 

It does not reproduce on other device/OS configurations. I've tested on iPhone 4s(7.x) , iPhone 6(8.3), iPad(8.1.3) and iPad Air (8.3)

Steps to reproduce:
1. In you git repo with sample test app copy over the style file at this link to an existing existing style (say emerald-v7.json)

http://54.227.241.87/static_html/test_grid_pbfs/style.json
2. Launch the sample test app. Switch to the new style.
3. Go to zoom level 4. 
4. Pan / pan with flick.

Testapp crashes within a few seconds.

Details on what the style/source does.
1. This style draw a grid of 8x8 squares  at zoom level 4.
2. The tile pbf is very trivial. It has 64 areas for each square. It also has lines marking the boundary of each square.

1ec5:
ref #1460

mb12:
#1460 is a different issue. #1460 is related to open gl calls not being disabled when app moves to background.

This issue consistently causes a crash when rendering a trivial grid of squares on iPhone 5.

mb12:
Even on devices where it does not crash, the fill-color is inaccurate. The expected behavior is that all grid squares would be rendered with yellow translucent color. But only half of the grid squares are rendered with this color. 
If I change the opacity to 1, then all the grid squares are rendered with yellow color. It looks like there is an issue with rendering fill areas with translucency (opacity < 1).

mb12:
I was able to get hold of a different iPhone 5 with a slightly different OS version (iOS 8.1.2) and it crashes on that device as well. Is it possible for someone to do some analysis on this? I have hit a dead end on this one.

friedbunny:
I'm inclined to blame this on a bug in the GPU driver for the PowerVR SGX543 series (with `RC2` seeming to mean "Release Candidate 2"?). The actual name of the iPhone 5 GPU is SGX543MP3.

[...]

Issue #1460, opened 2015-05-07
Crash in background — glFinish_Exec, gpus_ReturnNotPermittedKillClient, libGPUSupportMercury


picciano:
I started seeing these that past two days. This was from Crashlytics, so not exactly sure what user was doing at the time.

```
Thread : Crashed: com.apple.main-thread
0  libGPUSupportMercury.dylib     0x000000018c389f90 gpus_ReturnNotPermittedKillClient + 12
1  libGPUSupportMercury.dylib     0x000000018c38af40 gpusSubmitDataBuffers + 172
2  GLEngine                       0x0000000187078664 glFinish_Exec + 132
3  GLKit                          0x000000018446004c -[GLKView _display:] + 200
4  QuartzCore                     0x0000000187777a34 CA::Layer::display() + 148
5  QuartzCore                     0x0000000187758a7c CA::Layer::display_if_needed(CA::Transaction*) + 244
6  QuartzCore                     0x0000000187758734 CA::Layer::layout_and_display_if_needed(CA::Transaction*) + 44
7  QuartzCore                     0x0000000187757ebc CA::Context::commit_transaction(CA::Transaction*) + 276
8  QuartzCore                     0x0000000187757c3c CA::Transaction::commit() + 528
9  UIKit                          0x0000000187e0c56c _afterCACommitHandler + 152
10 CoreFoundation                 0x00000001833902a4 __CFRUNLOOP_IS_CALLING_OUT_TO_AN_OBSERVER_CALLBACK_FUNCTION__ + 32
11 CoreFoundation                 0x000000018338d230 __CFRunLoopDoObservers + 360
12 CoreFoundation                 0x000000018338d610 __CFRunLoopRun + 836
13 CoreFoundation                 0x00000001832b92d4 CFRunLoopRunSpecific + 396
14 GraphicsServices               0x000000018cacf6fc GSEventRunModal + 168
15 UIKit                          0x0000000187e7efac UIApplicationMain + 1488
16 TallyGo                        0x00000001001284f0 main (main.m:14)
17 libdyld.dylib                  0x000000019527aa08 start + 4
```

friedbunny:
This looks suspiciously similar to #1198, but not quite. @picciano, do you see any hardware/OS patterns for these crashes?

picciano:
I've been seeing it on iPhone 6, iOS versions 8.2 and 8.3.

[...]

PR #1805, opened 2015-06-30
Go dormant before termination

1ec5:
This is a speculative fix for #1460, ~~#1509~~, and #1680 per https://github.com/mapbox/mapbox-gl-native/issues/1460#issuecomment-99959434, building off the work in #1451. Once we get `UIApplicationWillTerminateNotification`, it’s unlikely that there will be any need to service further requests for rendering, programmatic viewport manipulation, or gesture recognition.

I’ll keep an eye on our HockeyApp stats to see if #1460 goes away at least.

/cc @incanus @friedbunny

### DIFF
--- a/platform/ios/MGLMapView.mm
+++ b/platform/ios/MGLMapView.mm
@@ -367,6 +367,7 @@ std::chrono::steady_clock::duration secondsAsDuration(float duration)
 
     // observe app activity
     //
+    [[NSNotificationCenter defaultCenter] addObserver:self selector:@selector(willTerminate) name:UIApplicationWillTerminateNotification object:nil];
     [[NSNotificationCenter defaultCenter] addObserver:self selector:@selector(sleepGL:) name:UIApplicationWillResignActiveNotification object:nil];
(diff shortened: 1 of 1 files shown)


1ec5:
@picciano @mb12 @robipresotto, feel free to take these changes for a spin. Hopefully we can get these mystery crashes to stop mucking up everyone’s analytics! :facepunch:→:deciduous_tree:

mb12:
#1680 is consistently reproducible. I have added steps to reproduce it in the issue itself.

mb12:
#1509 consistently reproducible. I have added steps to reproduce it in the issue itself.

1ec5:
@mb12, ah, that’s the alpha transparency one. Got a little ahead of myself.

1ec5:
Another attempt in #1821.

incanus:
Should we roll this back just to simplify a bit?

1ec5:
@incanus, I think handling will-terminate is correct but possibly unnecessary. If you think it’ll simplify the code (less redundancy), we can roll this bug. But let’s wait until #1821 shakes out before adding another variable.

PR #1807, opened 2015-06-30
makes sure STATE_IS_GPS_ON is correctly saved as state

zugaldia:
Because `onSaveInstanceState()` was called after the value of `mIsGpsOn` was already changed by `toggleGps()` its state wasn't correctly saved. I've added an extra `mIsGpsOnState` to track the actual state value.

I've also added a check in `onResume()` to make sure its value is respected when the app is resumed (for example, after a device rotation).

### DIFF
--- a/android/java/MapboxGLAndroidSDKTestApp/src/main/java/com/mapbox/mapboxgl/testapp/MainActivity.java
+++ b/android/java/MapboxGLAndroidSDKTestApp/src/main/java/com/mapbox/mapboxgl/testapp/MainActivity.java
@@ -57,6 +57,7 @@ public class MainActivity extends ActionBarActivity {
 
     // Used for GPS
     private boolean mIsGpsOn = false;
+    private boolean mIsGpsOnState = true;
     private LostApiClient mLocationClient;
     private GpsListener mGpsListener;
     private LocationRequest mLocationRequest;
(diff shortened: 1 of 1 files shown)


kkaefer:
Why are there two variables needed to track the state of the GPS setting?

zugaldia:
Basically, `mIsGpsOn` is tracking the GPS state within the activity instance (for example, when the user taps the menu icon to disable it), while `mIsGpsOnState` tracks the GPS state between instances (which happens, for example, when the user rotates the screen).

kkaefer:
I still don't understand the difference. In what circumstances would they ever be different?

zugaldia:
Let's suppose the user has the GPS enabled (`mIsGpsOn = true` and `mIsGpsOnState = true`) and then rotates the device:

1. Under a device rotation `onPause()` is called, and therefore `toggleGps()`, because it's a good practice to disconnect the GPS when the activity is paused. However, this method sets the value of `mIsGpsOn` to `false`.
2. Then, `onSaveInstanceState()` is called and sees that `mIsGpsOn` is `false`. However, the user had the GPS enabled (we need to store `mIsGpsOnState` which is still `true`).

A similar situation happens when the activity is created, which is currently set up to enable GPS by default (`mIsGpsOn = true`) but the user might have selected previously to disable it (`mIsGpsOnState = false`)

This is just a consequence of how `toggleGps()` is currently implemented that tackles both an UI update, and a client update, changing the value of `mIsGpsOn` every time. When we call `onPause()` we don't really need to update that value (or the UI for that matter), we only need to disconnect the client. Alternatively, if we separated `toggleGps()` into two different methods we could basically [combine both variables](https://github.com/googlesamples/android-play-location/blob/master/LocationUpdates/app/src/main/java/com/google/android/gms/location/sample/locationupdates/MainActivity.java).

ljbade:
Eventually this will all be moved to MapView anyway which should improve this.

/cc @bleege

tobrun:
closing, comment of @ljbade is in place

bleege:
@ljbade Agreed that this is going to end up in MapView at some point.  Is there a ticket for this?

ljbade:
@bleege Already been implemented. the isMyLocationEnabled is stored in the MapView's saved instance state, and TestApp and Sirius use this to update the FAB highlight on load.

PR #1808, opened 2015-07-01
Add collision box debugging to the global debug toggle

friedbunny:
@ansis was awesome enough to give us a collision box debug mode in https://github.com/mapbox/mapbox-gl-native/commit/8962442b1b838445efaf8dc12003225facac2a06, so let's enable it with `-[MGLMapView toggleDebug]`.

We *could* make it a separate public method, but for simplicity's sake this PR assumes everyone wants all the debug modes all the time.

/cc @incanus @1ec5


1ec5:
Fine by me. We should also nix the red tint on backgrounding eventually. I'm not sure it's even helpful today.

PR #1809, opened 2015-07-01
Remove tracking mode reset from resetNorthAnimated

friedbunny:
Removes tracking mode reset from `resetNorthAnimated:`.

[`handleCompassTapGesture:`](https://github.com/mapbox/mapbox-gl-native/blob/e6e340afd17a29ee4d183540e716376cdb1ef88e/platform/ios/MGLMapView.mm#L612-L617) intends to simply bump down to `MGLUserTrackingModeFollow` if it is following with heading, but the user tracking mode was getting clobbered by `resetNorthAnimated:`.

Supercedes #1257.

/cc @1ec5 @incanus

### DIFF
--- a/platform/ios/MGLMapView.mm
+++ b/platform/ios/MGLMapView.mm
@@ -1384,8 +1384,6 @@ std::chrono::steady_clock::duration secondsAsDuration(float duration)
 
 - (void)resetNorthAnimated:(BOOL)animated
 {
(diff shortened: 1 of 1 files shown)


1ec5:
:+1:

friedbunny:
Committed as 81d88755df16be60c5b14223989f39ff99c56943.

PR #1810, opened 2015-07-01
Do not pass uv_loop_t around

tmpsantos:
Use util::RunLoop::getLoop() instead when it is needed. This will
simplify a bit our interfaces.

### DIFF
--- a/platform/default/sqlite_cache.cpp
+++ b/platform/default/sqlite_cache.cpp
@@ -67,7 +67,7 @@ SQLiteCache::SQLiteCache(const std::string& path_)
 
 SQLiteCache::~SQLiteCache() = default;
 
-SQLiteCache::Impl::Impl(uv_loop_t*, const std::string& path_)
+SQLiteCache::Impl::Impl(const std::string& path_)
     : path(path_) {
 }
 
--- a/platform/default/sqlite_cache_impl.hpp
+++ b/platform/default/sqlite_cache_impl.hpp
@@ -3,8 +3,6 @@
 
 #include <mbgl/storage/sqlite_cache.hpp>
 
-typedef struct uv_loop_s uv_loop_t;
-
 namespace mapbox {
 namespace sqlite {
 class Database;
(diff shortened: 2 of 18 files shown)


tmpsantos:
@jfirebaugh review?

jfirebaugh:
I think I'm :-1: on this change -- it introduces more reliance on what is effectively a global variable, which hides these classes dependency on an active run loop, and can make things more difficult to test (I recall some tests passing `uv_default_loop()` rather than creating a `RunLoop`).

tmpsantos:
> I think I'm  on this change -- it introduces more reliance on what is effectively a global variable, which hides these classes dependency on an active run loop, and can make things more difficult to test (I recall some tests passing uv_default_loop() rather than creating a RunLoop).

That is true indeed, but I think it used to be the case before mbgl::Thread. Now I think it is very safe to assume that you always have a main loop per thread (exception being the Main thread that we don't assume any main loop at all).

What I tried to solve here is instead of receiving a loop reference that you don't know if it belongs to your current thread, you now say: reply to my thread, so use my current main loop.

And regarding the tests, we have to run them on a mbgl::Thread anyway because we have assertions like `currentlyOn(Map)` everywhere. That was the case for all the style tests I wrote.

jfirebaugh:
How about a compromise: remove the argument from the places where a `RunLoop`/`uv_loop_t*` dependency is "expected" (the `::Impl`s and `MapContext`), but leave it in `Style` and `GlyphStore` -- which are domain objects that shouldn't really have this dependency -- as a reminder that it's awkward and we'd like to properly remove it.

tmpsantos:
> How about a compromise: remove the argument from the places where a RunLoop/uv_loop_t* dependency is "expected" (the ::Impls and MapContext), but leave it in Style and GlyphStore -- which are domain objects that shouldn't really have this dependency -- as a reminder that it's awkward and we'd like to properly remove it.

Sounds great! :+1:

PR #1767, opened 2015-07-01
Setting styleURL immediately after initialization sometimes fails

tmpsantos:
Sometimes when I start my app (iOS) with Mapbox in it, it won't load the styles into the map. It generates the default MapKit styled map, without any markers. It's kind of annoying, because the problem is somewhat occurring on a random base. 

I load my custom stylesheet as a JSON file and link it to the map via the style url:

```mapView?.styleURL = NSURL(string: "asset://\(jsonFile)")```

Is it a known issue?

### DIFF
--- a/src/mbgl/map/map_context.cpp
+++ b/src/mbgl/map/map_context.cpp
@@ -50,6 +50,12 @@ MapContext::~MapContext() {
 void MapContext::cleanup() {
     view.notify();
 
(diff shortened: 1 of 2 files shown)


jfirebaugh:
Ah, a request/response race condition -- this is a common error in client-side JS, funny to see it in C++ code. :+1:

Issue #1273, opened 2015-04-15
two static renders in a row from same map w/ diff ratio breaks


samanpwbb:
First render, with ratio of 2:
![500x200 2x](https://cloud.githubusercontent.com/assets/108094/7159151/82e80bf8-e380-11e4-9711-6d4a45837f3f.png)

Second render, ratio is 1:
![500x200](https://cloud.githubusercontent.com/assets/108094/7159158/8ea54a50-e380-11e4-97a3-091120ff0389.png)

Third render, ratio still 1:
![500x200-1](https://cloud.githubusercontent.com/assets/108094/7159161/9445a34c-e380-11e4-9273-563211ea4000.png)

Looks like the ratio gets set after render is complete.

mikemorris:
This happens when running api-gl caching map object caching, requesting a @2x static image, then a @1x static image, going from @1x to @2x has similar problems. Refreshing to request the changed dimensions a second time (@1x, @2x, @2x) shows the correct image, because the view has been resized to the same dimensions as the prospective render by then.

```
current: 2
prospective: 1
painter->render
DEBUG [Android] Map::resize
DEBUG [Android] Transform::resize
DEBUG [Android] Transform::resize post lock
DEBUG [Android] Transform ::resize exit true
DEBUG [Android] Map::resize exit
view.swap
current: 1
prospective: 1
```

Backtrace:
- [Map::resize#L406-L413](https://github.com/mapbox/mapbox-gl-native/blob/static-render/src/mbgl/map/map.cpp#L406-L413)
- [View::resize#L17](https://github.com/mapbox/mapbox-gl-native/blob/static-render/src/mbgl/map/view.cpp#L17)
- [HeadlessView::discard#L208](https://github.com/mapbox/mapbox-gl-native/blob/static-render/platform/default/headless_view.cpp#L208)
- [Map::render#L746](https://github.com/mapbox/mapbox-gl-native/blob/static-render/src/mbgl/map/map.cpp#L746)

kkaefer:
Is this still happening?

mikemorris:
This should be resolved by https://github.com/mapbox/mapbox-gl-native/pull/1799

PR #1799, opened 2015-07-01
Never change pixel ratio after creating a Map object

kkaefer:
We're currently allowing users to change the pixel ratio after a Map object is created. There are very few use cases for this: moving a window from one screen to another on OS X, and connecting a TV to an iOS device. We should remove the ability to change the pixel ratio. If the frame buffer changes at some point, we can use `glViewport` to resize the rendering to the new framebuffer dimensions. It won't look perfect, but it's an edge case that causes more headcache than it's worth.

### DIFF
--- a/src/mbgl/map/map_context.hpp
+++ b/src/mbgl/map/map_context.hpp
@@ -26,6 +26,11 @@ class StillImage;
 struct LatLng;
 struct LatLngBounds;
 
+
+struct FrameData {
+    std::array<uint16_t, 2> framebufferSize;
+};
+
 class MapContext : public Style::Observer {
 public:
     MapContext(View&, FileSource&, MapData&);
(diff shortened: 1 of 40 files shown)


src/mbgl/map/map_context.hpp
```diff
@@ -40,13 +45,11 @@ class MapContext : public Style::Observer {
...
     void pause();
 
-    void resize(uint16_t width, uint16_t height, float ratio);
-
     using StillImageCallback = std::function<void(std::exception_ptr, std::unique_ptr<const StillImage>)>;
 
     void triggerUpdate(const TransformState&, Update = Update::Nothing);
-    void renderStill(const TransformState&, StillImageCallback callback);
-    RenderResult renderSync(const TransformState&);
+    void renderStill(const TransformState&, const FrameData&, StillImageCallback callback);
```
jfirebaugh: What's the rationale for introducing a separate argument rather than making `framebufferSize` part of `TransformState`?

kkaefer: The TransformState is a status that is set for the map in general. It contains the extent (width/height) and position and is used to determine the tiles we need to load. The Framebuffer size is independent of the logical dimensions of the map and is always associated with a single frame, whereas TransformState represents the overall state of that map and isn't tied to a particular frame.

I plan to move additional data into `FrameData`, such as the `animationTime`.

jfirebaugh:
CC @mikemorris -- mutable pixel ratio may be an important use-case for the node bindings.

kkaefer:
I think it shouldn't; instead of changing the dimensions of the map/view object for every request, we should cache those per size/ratio to save on memory allocation time.

mikemorris:
@jfirebaugh mutable pixel ratios was causing issues like https://github.com/mapbox/mapbox-gl-native/issues/1273, I'd be happy to see this gone and segment cache on ratio instead.

jfirebaugh:
Cool, sounds like this simplifies things all around then.

PR #1813, opened 2015-07-01
split renderSync and transition nudging to allow client view syncing

incanus:
Picking up from stuff learned in https://github.com/mapbox/mapbox-gl-native/issues/1125#issuecomment-117256712, this breaks the current `renderSync` into two parts — the actual render call and a `nudgeTransitions` which moves a step in the `Transform`, possibly calling `Map::update` if needed. 

The reason for this is needing to sync client-side view (i.e. `UIView` objects) positions with the GL map, and that step needing to come in between these two to remain in sync. 

Todo: 

* [ ] Review terminology used (@jfirebaugh @kkaefer?)
* [ ] Review for other possible side effects
* [ ] Update other platforms besides iOS if approach is :ok_hand:

### DIFF
--- a/include/mbgl/map/map.hpp
+++ b/include/mbgl/map/map.hpp
@@ -63,7 +63,10 @@ public:
     void renderStill(StillImageCallback callback);
 
     // Triggers a synchronous or asynchronous render.
-    void renderSync();
+    bool renderSync();
+
+    // Nudges transitions one step, possibly notifying of the need for a rerender.
+    void nudgeTransitions(bool forceRerender);
 
     // Notifies the Map thread that the state has changed and an update might be necessary.
     void update(Update update = Update::Nothing);
--- a/platform/ios/MGLMapView.mm
+++ b/platform/ios/MGLMapView.mm
@@ -682,7 +682,8 @@ std::chrono::steady_clock::duration secondsAsDuration(float duration)
 
         _mbglMap->setSourceTileCacheSize(cacheSize);
 
-        _mbglMap->renderSync();
+        bool needsRerender = _mbglMap->renderSync();
+        _mbglMap->nudgeTransitions(needsRerender);
     }
 }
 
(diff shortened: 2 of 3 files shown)


incanus:
Per chat, @kkaefer says we should probably snapshot time and use that to be consistent through the render of a frame and update of transitions, instead of having varying notions of time throughout. This PR is more of a workaround of that.

jfirebaugh:
@incanus Instead of splitting this up, can you implement this by updating to latest master and doing the syncing when the view receives `notifyMapChange` with `MapChangeDidFinishRenderingFrame` or `MapChangeDidFinishRenderingFrameFullyRendered`? This notification is sent at the relevant point in time for UIView syncing.

incanus:
Duh, yeah, great idea. We'll just need some tests to make sure that the transitions stay after the closing notification, though.

incanus:
Just talked this out with @kkaefer — @jfirebaugh, this won't work, because we do everything synchronously. So the order of operations is: 

1. Post "will render" notification. 
1. Render frame. 
1. Post "did render" notification.
1. **Don't** call out to update view syncing, because we're synchronous here. 
1. Update the transform
1. ...
1. At some point, update view syncing, which is out of sync because the transform has been modified. 

@kkaefer is going to look today & tomorrow into https://github.com/mapbox/mapbox-gl-native/pull/1813#issuecomment-117708076 to solve this.

jfirebaugh:
By "synchronous" you mean the notification is dispatched from the rendering thread, while the view syncing needs to happen on the main thread?

incanus:
Yeah, sorry, the post-render-post-update cycle all happens inline in the map thread, and the view update needs to happen on the main thread in between steps 3 and 4.

incanus:
It might make sense to go with this spilt for b3 for an easy win and circle back on the refactor work.

incanus:
Going to apply this to the `release-v0.5.0` branch for b3; we can revisit globally after hitting https://github.com/mapbox/mapbox-gl-native/pull/1813#issuecomment-117708076.

Issue #1814, opened 2015-07-01
line tessellation bug


ansis:
seen here: https://github.com/mapbox/mapbox-gl-native/issues/1771#issuecomment-117713118

ansis:
![screen shot 2015-07-02 at 1 28 27 pm](https://cloud.githubusercontent.com/assets/1421652/8476144/a20a2220-20be-11e5-8307-f18ebc5c44e2.png)

Issue #837, opened 2015-02-10
Runtime styling API


incanus:
[Updated 20 April 2016 by @jfirebaugh]

The [Mapbox GL Style Specification](https://www.mapbox.com/mapbox-gl-style-spec/) forms the [domain model](https://en.wikipedia.org/wiki/Domain_model) for Mapbox maps. It defines objects such as:

* styles
* sources
* layers
* paint and layout properties

These domain objects are central to rendering and manipulating the map, and they should be exposed by both the core mapbox-gl-native API and by the Android and iOS bindings.

This functionality exists in mapbox-gl-js already, via APIs such as [`addSource`](https://www.mapbox.com/mapbox-gl-js/api/#Map.addSource), [`addLayer`](https://www.mapbox.com/mapbox-gl-js/api/#Map.addLayer), and [`setPaintProperty`](https://www.mapbox.com/mapbox-gl-js/api/#Map.setPaintProperty). Together, we call this set of APIs the **runtime styling API** (not to be confused with the [Mapbox Styles API](https://www.mapbox.com/api-documentation/#styles), a web API).

To get a sense of what is possible with read and write access to the style and its contents, take a look the mapbox-gl-js examples: the large majority of them use one or more runtime styling APIs.

* [Show and hide layers](https://www.mapbox.com/mapbox-gl-js/example/toggle-layers/) (uses `setLayoutProperty`)
* [Create a timeline animation](https://www.mapbox.com/mapbox-gl-js/example/timeline-animation/) (uses `setFilter`)
* [Add live realtime data](https://www.mapbox.com/mapbox-gl-js/example/live-geojson/) (uses `addSource` and `setData`)

Furthermore, the rich functionality provided by mapbox-gl-js plugins like [mapbox-gl-directions](https://github.com/mapbox/mapbox-gl-directions), [mapbox-gl-geocoder](https://github.com/mapbox/mapbox-gl-geocoder), and [mapbox-gl-draw](https://github.com/mapbox/mapbox-gl-draw) are enabled by the runtime styling API.

#### What we have now

mapbox-gl-native and the mobile SDKs currently have an "Annotations API" which provides some forms of runtime styling, namely adding and removing markers, lines, and polygons, with a limited degree of control over their visual appearance.

[...]

PR #1820, opened 2015-07-02
expose annotation styling

### LINKED ISSUE #1734: expose more style properties for annotations
Shape annotations need to support more features in the exposed API. 

Definite: 

- line join
- line cap
- line dash

Maybe: 

- line/fill image
- translate (for putting lines next to roads? e.g. multiple transit line colors)
- antialias off? 

@friedbunny

incanus:
Fixes #1734 and allows a workaround for https://github.com/mapbox/mapbox-gl-native/issues/1814 by exposing the full set of line & fill layout and paint properties to map clients. 

Todo: 

- [ ] Safely query what's set in the passed style (i.e. no `.at()` directly). 
- [ ] Render more values. 
- [ ] Set sane defaults in core for the properties (not necessarily the same as the defaults in `LineProperties`/`FillProperties`; for example, shapes shouldn't be black or default to miter line join). 
- [ ] Test dead code removal from https://github.com/mapbox/mapbox-gl-native/commit/174fb6be7c07d9a177dda05fe8d6738526f86239. 
- [ ] Find a better way to determine fill/line distinction now that the use of the `StyleProperties` variant is gone (i.e. not [this](https://github.com/mapbox/mapbox-gl-native/blob/1734-expose-annotation-styling/src/mbgl/map/map_context.cpp#L169)).


jfirebaugh:
Is there a dependency chain pulling in `class_properties.hpp` and `property_transition.hpp`? Those looks like they might be unnecessary. I'm ok with the other includes being "semi-public", in that they are used by iOS but not considered part of the public core GL. These interfaces will probably change a lot with mutable style API and data driven styles.

jfirebaugh:
@incanus What's the status of this PR?

jfirebaugh:
Going in the direction of #837 instead.

incanus:
:+1:

PR #1821, opened 2015-07-02
[WIP] Defer initialization until app enters foreground

1ec5:
Another speculative fix for #1460, since #1805 didn’t quite do the trick. If MGLMapView is initialized while the app is in the background, defer the nitty-gritty details of initialization (~~particularly~~ setting up the GL context and GL subview, ~~but also things like events and background snapshotting~~) until the app enters the foreground for the first time.

~~This fix assumes that the developer won’t be doing things like switching styles or setting the global access token while the app is still in the background. Correctly handling such cases would require checking for the presence of a map at all times and queueing up changes to map view properties until the app enters the foreground.~~

Areas to test include backgrounding and foregrounding the app, launching the app via a URL handler, and switching between view controllers that do or do not contain an MGLMapView.

/cc @incanus @friedbunny

### DIFF
--- a/platform/ios/MGLMapView.mm
+++ b/platform/ios/MGLMapView.mm
@@ -206,58 +206,28 @@ std::chrono::steady_clock::duration secondsAsDuration(float duration)
 - (void)commonInit
 {
     _isTargetingInterfaceBuilder = NSProcessInfo.processInfo.mgl_isInterfaceBuilderDesignablesAgent;
+    
+    BOOL background = [UIApplication sharedApplication].applicationState == UIApplicationStateBackground;
+    if (!background)
+    {
+        [self createGLView];
+    }
 
-    // create context
-    //
-    _context = [[EAGLContext alloc] initWithAPI:kEAGLRenderingAPIOpenGLES2];
-    NSAssert(_context, @"Failed to create OpenGL ES context.");
 
-    // setup accessibility
-    //
     self.accessibilityLabel = @"Map";
-
-    // create GL view
-    //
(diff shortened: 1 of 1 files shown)


friedbunny:
Looking good. :fingerscrossed:

1ec5:
To be honest, I’m not entirely sold on this approach yet, because the developer still needs to take care to avoid touching the MGLMapView in any way before the app enters the foreground. It’s important to avoid accessing the mbgl::Map, which hasn’t been created yet because we can’t create the GL context.

1ec5:
c1d6239665d8fdb62c7b437f808298436e7a6c38 addresses the misgivings I had in https://github.com/mapbox/mapbox-gl-native/pull/1821#issuecomment-118094194. After some testing, I think this change will be ready to go in.

1ec5:
If for whatever reason the app goes to the foreground after launching in the background, the MGLMapView will continue to be functional. If I force that code path, the only difference I see is this message in the console, which seems benign:

```
2015-07-06 10:28:43.934 Mapbox GL[5010:736058] [WARNING] {Map}[OpenGL]: Not using Vertex Array Objects
```

1ec5:
https://rink.hockeyapp.net/manage/apps/163490/app_versions/34/crash_reasons/39240783

1ec5:
The issue must be that the app becomes “active” in the background, triggering a `UIApplicationWillEnterForegroundNotification` that this branch currently responds to by setting up the OpenGL ES context. caf874f9e6b1763fdfae6b441af71b1bf1361c9c would address that issue.

Issue #877, opened 2015-02-13
annotations plan


incanus:
"Annotations" is iOS-speak for points and polygons on the basemap with a high-level client-side API. 

* **Data** — This encapsulates such things as client-side data (#507) or formats like GeoJSON, and also includes programmatic addition of points and polygons that don't exist in files or formats at all (e.g. show user location, add these stores to the map, trace this route from parsed directions API). In fact, format-less features are the MVP for this. Spatial indexing falls under this category as well — the system needs to know what's on screen right now as well as be able to translate between pixel-space and geo-space (https://github.com/mapbox/mapbox-gl-native/pull/853). 

* **Design** — Polylines & polygons can get styling properties like width, fill color, opacity, and stroke color. This can change at runtime, so it's not part of the GL style. Markers can get a) static marker imagery, b) iOS `UIView` objects, or even c) animations of views. These can also all change / be provided at runtime. Markers always remain upright when the map rotates and order themselves in the z-axis from current top-to-bottom of screen, updating continuously as the map rotates. Lastly, marker annotations have callouts, which should be implemented specific to iOS (likely using [`SMCalloutView`](https://github.com/nfarina/calloutview) as our SDK does). 

* **Interaction** — Users can tap on poly & marker visuals and get callbacks with the associated data object referenced. 

* **API** — All APIs follow [Apple's MapKit](https://developer.apple.com/library/ios/documentation/MapKit/Reference/MapKit_Framework_Reference/index.html#//apple_ref/doc/uid/TP40008210), so relevant classes & protocols are `MKMapView`, `MKAnnotation`, `MKAnnotationView`, and `MKMapViewDelegate`. 

  - An `MKAnnotation`-derived data object is added (`map.addAnnotation(foo)`) .
  - An `MKMapViewDelegate` is set on a map (`map.delegate = bar`). 
  - When the visual for an annotation is relevant/on screen, a view is requested from the delegate, which creates it on the fly (`func mapView(map, viewForAnnotation: foo)`). 
  - The map maintains and responds to annotation selection state (`func mapView(map, didSelectAnnotation:foo)`)

Will kick off deeper tickets for each of these buildouts as we progress. 

![](https://github.com/mapbox/weekend-picks-template-ios/raw/master/screenshot.png)

incanus:
[...]

PR #941, opened 2015-07-02
allow for user-provided symbol imagery (aka Sprites)

kkaefer:
As part of #877, we need to take user-provided marker imagery and render it in the GL layer. We also support standard Maki icons, and beyond the iOS beta, we will support native `UIView` / Android views tracking atop the GL view. 

We need an API for a user to provide a bitmap into the system and have it become part of our sprite sheet, independent of style currently in play. 

/cc @kkaefer @1ec5 @bleege


kkaefer:
- [ ]  add an example/shortcut in the desktop + iOS test app
- [ ] add test that checks rendering images with custom sprite images
- [ ] Add tests for SpriteAtlas
	- [ ] Add a few images to the Sprite atlas, check that correct texture results
	- [ ] Add images that don't exist in the SpriteStore
	- [ ] Update SpriteStore, make sure that images in the atlas get updated
	- [ ] same texture, but different wrap values return different positions

kkaefer:
@incanus All of the C++ APIs should now be here. Do you want to sync up later on how the Cocoa APIs should look like?

kkaefer:
Open tasks that can become separate issues:
- [ ] Remove icons from sprite atlas once we're done with them (like GlyphAtlas does already)
- [ ] Merge GlyphAtlas and SpriteAtlas (and LineAtlas?) implementation
- [ ] When removing an image from the sprite atlas, we should not remove the image immediately, but move it to a free list. When adding a new image and there is no space in the atlas, we delete the image from the free list and reuse the space in the texture. When the image that is being added was the same that was previously removed, we don't need to do anything (this may save some texture uploads).

incanus:
Per voice with @kkaefer, I am going to grab the Cocoa wrapper for this around passing user `UIImage` objects as pixel arrays to C++ as sprites to add to core.

incanus:
https://github.com/mapbox/mapbox-gl-native/commit/b2d2690 has a first cut at a Cocoa API for this. 

- [ ] **HACK:** Currently [sending back](https://github.com/mapbox/mapbox-gl-native/blob/b2d2690255ce75cfc3bfd1afebc02808d8310c9d/platform/ios/MGLMapView.mm#L1902-L1929) a random-colored pixel array much like the OS X `P` key demo since I'm having trouble turning a runtime `UIImage` into a pixel array — I think this is what you were seeing @kkaefer with non-premultiplied alpha? 

Other notes: 

- Added `-[MGLMapViewDelegate mapView:imageForAnnotation:]`, which can return `nil` to use the default marker for a style (e.g. `default_marker`) or an `MGLAnnotationImage` object. 

- An `MGLAnnotationImage` object has an associated `UIImage` and a `reuseIdentifier` string much like `MKAnnotationView`. You can either create a new `MGLAnnotationImage` with a runtime image or you can obtain an existing one to reuse efficiently with `-[MGLMapView dequeueReusableAnnotationImageWithIdentifier:]`. 

- Custom imagery still properly queries the backend for the sprite offset so that the popup placement looks right. 

- Deprecated `-[MGLMapViewDelegate mapView:symbolForAnnotation:]`. Users no longer have to care about symbol names! 

- [ ] This currently logs a bit of info for debugging purposes so that you can see that sprites are properly being reused when indicated to have the same `reuseIdentifier`. 

- [ ] Still needs some cleanup around removal of the last instance of a custom sprite during annotation removal. 

- [ ] Currently doing [some hackery](https://github.com/mapbox/mapbox-gl-native/commit/b2d2690#diff-33b34b5f099a16158c8462652a0c3688) around using different sprites in the iOS demo app.

incanus:
@kkaefer https://github.com/mapbox/mapbox-gl-native/commit/8520e71f17ce3d1356883917fcb0ab97d1364051 stubs out annotation symbol cleanup inline during `removeAnnotations()` rather than some sort of periodic garbage collection. However, since neither `setSprite()` nor `removeAnnotations()` are called synchronously, this can crash as we still need to render annotations for at least a frame afterwards before removal. 

I will dig at this a bit, but it's crashing in `SpriteAtlas::copy()` when `holder.texture` is `NULL`. 

Any recommendations?

incanus:
Pretend `red_square` is our custom-drawn `UIImage`, when actually it's a random little circle, but still drawn on the fly. 

screen shot 2015-07-06 at 6 47 59 pm

/cc @1ec5 @friedbunny on https://github.com/mapbox/mapbox-gl-native/commit/b2d2690 for thoughts on the nascent Cocoa API.

incanus:
Note also that `MGLAnnotationImage` will be deprecated in favor of the eventual `MGLAnnotationView.image` once https://github.com/mapbox/mapbox-gl-native/issues/1784 is done.

incanus:
>  Still needs some cleanup around removal of the last instance of a custom sprite during annotation removal.

Maybe we can punt on this in light of https://github.com/mapbox/mapbox-gl-native/pull/941#issuecomment-118877956 and revert https://github.com/mapbox/mapbox-gl-native/commit/8520e71f17ce3d1356883917fcb0ab97d1364051 since the image isn't removed from the texture anyway.

incanus:
>  **HACK:** Currently sending back a random-colored pixel array much like the OS X `P` key demo since I'm having trouble turning a runtime `UIImage` into a pixel array — I think this is what you were seeing @kkaefer with non-premultiplied alpha?

I've got to be missing something basic here. I've tried: 

- Manually drawing the image in a `CGContextRef` to get at the pixel array. 

- Using `mbgl::util::Image` to do so. 

- Just using: 

  ```c++
  NSData *data = (NSData *)CFBridgingRelease(CGDataProviderCopyData(
      CGImageGetDataProvider(annotationImage.image.CGImage)));
  const char *bytes = (const char *)data.bytes;
  std::string byteString(bytes, data.length);
  ```

  Which gets me a good-looking pixel array, but I have to manually inspect the alpha channel info and byte order / endianness, as this varies between even code-drawn and disk-read imagery. For example, when creating a `UIImage` on the fly, the alpha is premultiplied last and the byte order is `BGRA`.

mb12:
Are you trying to convert UIImage to data that can be uploaded to GPU via glTexImage2D?

incanus:
No, at this point I'm just trying to take a runtime `UIImage` and create a `SpriteImage` out of it, which requires a raw pixel array as a `std::string`.

include/mbgl/ios/MGLMapView.h
```diff
@@ -258,6 +259,18 @@ IB_DESIGNABLE
 *   @param annotations The array of annotations to remove. Objects in the array must conform to the MGLAnnotation protocol. */
 - (void)removeAnnotations:(NS_ARRAY_OF(id <MGLAnnotation>) *)annotations;
 
+/** Returns a reusable annotation image object located by its identifier.
+*
+*   For performance reasons, you should generally reuse MGLAnnotationImage objects for annotations in your map views. Dequeueing saves time and memory during performance-critical operations such as scrolling.
+*
+*   @param identifier A string identifying the annotation image to be reused. This string is the same one you specify when initially returning the annotation image object using the mapView:imageForAnnotation: method.
+*   @return An annotation image object with the specified identifier, or `nil` if no such object exists in the reuse queue. */
+- (MGLAnnotationImage *)dequeueReusableAnnotationImageWithIdentifier:(NSString *)identifier;
```
1ec5: The return type should be `nullable`.

include/mbgl/ios/MGLAnnotationImage.h
```diff
@@ -0,0 +1,10 @@
+#import <UIKit/UIKit.h>
```
1ec5: This file should be audited for nullability. Since neither property should ever be `nil`, surround the header with `
NS_ASSUME_NONNULL_BEGIN` and `NS_ASSUME_NONNULL_END` and import MGLTypes.h (to get the Xcode 6.1 compatibility shim).

include/mbgl/ios/MGLMapView.h
```diff
@@ -336,7 +349,13 @@ IB_DESIGNABLE
 *   @param mapView The map view that requested the annotation symbol name.
 *   @param annotation The object representing the annotation that is about to be displayed. 
 *   @return The marker symbol to display for the specified annotation or `nil` if you want to display the default symbol. */
-- (nullable NSString *)mapView:(MGLMapView *)mapView symbolNameForAnnotation:(id <MGLAnnotation>)annotation;
+- (nullable NSString *)mapView:(MGLMapView *)mapView symbolNameForAnnotation:(id <MGLAnnotation>)annotation __attribute__((unavailable("Use -mapView:imageForAnnotation:.")));
```
1ec5: Would it be worth leaving the existing symbol-based annotation API around? Seems like it’d continue to be the most convenient way to use Maki icons as annotation images (for instance).

mb12:
Can you try something like this?

```
UIImage* image = [UIImage cocoa init method];
uint8_t* data;
int out_len;
int out_width;
int out_height;
// We are trying to convert UIImage* image to SpriteImage
// This method is defined below. 
UIImageToData(image,  data, out_len, out_width, out_height) ;
std::string data_in_string(data, out_len);
SpriteImage my_spriteimage(
  out_width, out_height,  1, data_in_string, false);
delete[] data;

// allocates data using new, make sure that caller calles delet[]
static void UIImageToData(UIImage* image, uint8_t* & data, int& out_len, int& out_width, int& out_height) 
{
        CGImageRef imageRef = [image CGImage];
        int width = (int)(CGImageGetWidth(imageRef));out_width = width;
        int height = (int)(CGImageGetHeight(imageRef)); out_height = height;

        const int bitsPerElement = 8;
        int sizeInBytes = height*width*4; out_len = sizeInBytes;
        int texturePitch = width*4;
        data = new uint8_t[sizeInBytes];
        memset(data, 0x00, sizeInBytes);
    
        CGColorSpaceRef colorSpace = CGColorSpaceCreateDeviceRGB();
        CGContextRef context = CGBitmapContextCreate(data, width, height, bitsPerElement, texturePitch, colorSpace, kCGImageAlphaPremultipliedLast);
        CGColorSpaceRelease(colorSpace);
        bool flip = true;
        if(flip) {
                CGAffineTransform flipVertical = CGAffineTransformMake(1, 0, 0, -1, 0,height);
                CGContextConcatCTM(context, flipVertical);
        }

        CGContextDrawImage(context, CGRectMake(0, 0, width, height), imageRef);
        CGContextRelease(context);
}

```

include/mbgl/ios/MGLAnnotationImage.h
incanus: Grabbed this last night already: https://github.com/mapbox/mapbox-gl-native/commit/56fbebf

include/mbgl/ios/MGLMapView.h
incanus: bcc3e3a

include/mbgl/ios/MGLAnnotationImage.h
incanus: > import `MGLTypes.h`

Is this strictly required if it's pulled in with `MapboxGL.h`?

include/mbgl/ios/MGLMapView.h
incanus: Hmm, hadn't thought about Maki. You'd have to know which were in the style, though, as it's not all of them by default. I don't think there's much benefit, until such time that we maybe roll out an `RMMarker`-like constructor that lets you pass _any_ Maki symbol name (which it would then probably fetch on demand and upload via the new runtime imagery API).

incanus:
@kkaefer helped me work through the issues around size and scale between the context and the image passed to `SpriteImage`. As of https://github.com/mapbox/mapbox-gl-native/commit/ebbd1d69d2954678668ba689d02291052a79eb6e we are now good with both on-disk and at-runtime `UIImage`!

include/mbgl/ios/MGLAnnotationImage.h
1ec5: No, but then this header should pull in MGLTypes.h. Headers shouldn't have implicit dependencies.

incanus: https://github.com/mapbox/mapbox-gl-native/commit/f8342da

incanus:
https://github.com/mapbox/mapbox-gl-native/commit/cb7d1e9759dbf45be7ee33e322c79d14b047de0e brings in a better example, which uses the last two characters of the point title to draw a marker image and reuses identical numbers across markers. 

ios simulator screen shot jul 7 2015 9 08 42 am

incanus:
This is feeling really good to me and is ready to go Cocoa-wise @kkaefer other than the fact that I'm going to look into changing an annotation's image after initial set. It's possible on the GL side; the trick will be telling Cocoa to refresh a particular point without having it move offscreen first.

incanus:
You know what — let's punt on changing imagery for b3. We're good here.

incanus:
This is in @kkaefer's hands to split into two PRs — one around the sprite work and the other for the user-provided symbols. Looking to ideally merge both into `master` tomorrow.

PR #1827, opened 2015-07-02
Fix round-tripping between pixelForLatLng() / latLngForPixel()

### LINKED ISSUE #1817: setVisibleCoordinateBounds:edgePadding:animated: not working for iPhone 6
Okay, here's a weird one. I am using `setVisibleCoordinateBoundsedgePaddinganimated:` to zoom the map into a route. On iPhone 5S, this work perfectly. However, on the iPhone 6 the map is centered way off into the mountains. I've double checked that I am passing in the same coordinates. The only difference (from what I can tell) is that the map is a bit wider on the iPhone 6 screen.

Same code base, running in Simulator. Any ideas?

iPhone 5S Results
ios simulator screen shot jul 1 2015 12 54 24 pm

iPhone 6 Results
ios simulator screen shot jul 1 2015 12 51 19 pm

iPhone 5S Values
screen shot 2015-07-01 at 12 53 43 pm

iPhone 6 Values
screen shot 2015-07-01 at 12 51 08 pm

1ec5:
Followup to cfa2eaf5c21858483cd12ee0768b43315ef29e95 for #1406. iPhone 6 has an odd width and an odd height. Dividing both figures by 2 but truncating the result leads to egregious errors when fitting to bounds at low zoom levels.

Fixes #1817.

### DIFF
--- a/src/mbgl/map/transform_state.cpp
+++ b/src/mbgl/map/transform_state.cpp
@@ -218,8 +218,8 @@ const LatLng TransformState::latLngForPixel(const vec2<double> pixel) const {
     LatLng ll = getLatLng();
     double zoom = getZoom();
 
-    const double centerX = width  / 2;
-    const double centerY = height / 2;
+    const double centerX = static_cast<double>(width) / 2.0;
+    const double centerY = static_cast<double>(height) / 2.0;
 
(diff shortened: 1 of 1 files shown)


jfirebaugh:
:+1:

PR #1687, opened 2015-07-02
Raster tiles are not reporting parsing errors

tmpsantos:
We should report the same as we do for vector tiles. Also, add tests for verification.


jfirebaugh:
:+1:

PR #1829, opened 2015-07-02
Fix rotation edge cases and compass overspinning

1ec5:
This PR addresses some edge cases around bearing, obsoleting #1558:

* `mbgl::util::wrap()` is now min-inclusive, max-exclusive. It is the responsibility of the caller to handle the case in which `min` is returned. Previously, `wrap(360, 0, 360)` would return 360 but `wrap(720, 0, 360)` would return 0. This is how mapbox-gl-js behaves, but I believe the max-exclusive behavior is more predictable.
* `-[MGLMapView direction]` now returns a value in the range [0, 360) instead of [0, 360].
* When pressing the compass to reset the direction to due north, the compass’ animation now begins at the current rotation. Previously, it began at due north and spun 360° every time, due to an overzealous response to change notifications. (ref #1431, #1624, 6532438fe47bca1af690779ffa7292d6f1080d0c)
* For consistency, during drift rotation, changes to the compass’ rotation are driven by mbgl animation rather than view-based animation.

/cc @mourner @friedbunny @kkaefer

### DIFF
--- a/include/mbgl/util/math.hpp
+++ b/include/mbgl/util/math.hpp
@@ -112,10 +112,12 @@ T clamp(T value, T min, T max) {
     return value < min ? min : (value > max ? max : value);
 }
 
+// Constrains n to the given range (including min, excluding max) via modular
+// arithmetic.
 template <typename T>
 T wrap(T value, T min, T max) {
     T d = max - min;
-    return value == max ? value : std::fmod((std::fmod((value - min), d) + d), d) + min;
+    return std::fmod((std::fmod((value - min), d) + d), d) + min;
 }
 
 template <typename T>
--- a/platform/ios/MGLMapView.mm
+++ b/platform/ios/MGLMapView.mm
@@ -15,6 +15,7 @@
 #include <mbgl/storage/default_file_source.hpp>
 #include <mbgl/storage/network_status.hpp>
 #include <mbgl/util/geo.hpp>
+#include <mbgl/util/math.hpp>
 #include <mbgl/util/constants.hpp>
 
 #import "MapboxGL.h"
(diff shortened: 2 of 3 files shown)


jfirebaugh:
:+1: 

Want to port the wrap change to gl-js too?

friedbunny:
It's still possible to trigger #1558 by doing a sharp-rotate across 180º, but only going right-to-left.

friedbunny:
Also, when tapping to reset to north, the compass does a strange little jaunt in the opposite direction before correcting to north.

1ec5:
> Want to port the wrap change to gl-js too?

Sure, once I’m reasonably confident about this change. `wrap()` has more callers in gl-js, but I think I understand the intent of each one.

> It's still possible to trigger #1558 by doing a sharp-rotate across 180º, but only going right-to-left.

Does 2ccb4e94f4936b6be1ffefb6c6ef0de48a6a1664 do the trick?

> Also, when tapping to reset to north, the compass does a strange little jaunt in the opposite direction before correcting to north.

Happens in master too. Turn on Slow Animations (⌘T) in the Simulator to see it.

1ec5:
4936f8637e12967a3046a7716aa3c598e966b968 fixes the compass overspinning.

1ec5:
As I pointed out in https://github.com/mapbox/mapbox-gl-native/pull/1582#issuecomment-102951670, we need to be careful about updating Cocoa-side state in response to model/mbgl changes. Something like 4936f8637e12967a3046a7716aa3c598e966b968 will be required for #1582 as well.

friedbunny:
> Does 2ccb4e9 do the trick?

Nope, though the behavior has changed. Sharply rotating over 180º from either direction will get the compass stuck on the screen upon tapping, but doing it again the reverse direction will unstick it. :grin:

I'll fiddle around with this over the long weekend.

1ec5:
Look good to you, @friedbunny?

friedbunny:
giphy

1ec5:
> Want to port the wrap change to gl-js too?

mapbox/mapbox-gl-js#1363

PR #1830, opened 2015-07-02
Turn the compass into a button

1ec5:
Touching down on the compass but dragging outside it should cause the map to pan, not reset to north. (This is how Maps behaves, though I’m open to changing the event back to `UIControlEventTouchDown`.) This PR also renames the property to reflect its new type.

This change obsoletes part of 90af18e708b617591766950413e3691449cad2ce for #1496 but omits the accessibility-related changes contained therein.

/cc @incanus @friedbunny

### DIFF
--- a/include/mbgl/ios/MGLMapView.h
+++ b/include/mbgl/ios/MGLMapView.h
@@ -73,8 +73,8 @@ IB_DESIGNABLE
 *   The default value of this property is `YES`. */
 @property(nonatomic, getter=isRotateEnabled) BOOL rotateEnabled;
 
-/** The compass image view shown in the upper-right when the map is rotated. */
-@property (nonatomic, readonly) UIImageView *compassView;
+/** The compass button shown in the upper-right when the map is rotated. */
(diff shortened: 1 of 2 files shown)


friedbunny:
This change introduces a delay between tapping the compass and return-to-north activation. Tapping highlights the compass button (i.e., adds a dark overlay) for ~0.5s before activating.

1ec5:
Yes, that’s the price to pay for accepting pans from the top-right corner of the screen. MapKit does the same thing, though it doesn’t highlight the button. I can easily nix the highlight by setting the same compass image for the highlighted state if that makes the compass feel more natural.

1ec5:
Switching to `UIControlEventTouchDown` would eliminate the delay but also eliminate the ability to detect pans from the corner.

1ec5:
Ah, I’m mistaken: we already do support panning from the corner because the compass and pan gestures aren’t allowed to run simultaneously. The MapKit compass also recognizes taps immediately without delay. Serves me right for testing only in the Simulator.

PR #1832, opened 2015-07-03
Missing parameters while issuing gdb via `make test-foo*`

brunoabinader:
Two things were preventing me from running the tests on OSX:

- We cannot assume that `gdb` is installed in all environments;
- We were not passing all arguments to make test (eg. using `$1` instead of `$@` in `scripts/run_tests.sh`)


tmpsantos:
:+1:

jfirebaugh:
Whoops, sorry about that. Fix looks good.

PR #1794, opened 2015-06-26
Add YCM config file for VIM users

tmpsantos:
Tailor-made for Mapbox GL Native.

For instructions of how to use, see:
https://github.com/Valloric/YouCompleteMe

PR #1836, opened 2015-07-05
Update YCM to support OSX

brunoabinader:
Besides a couple file path modifiers, we need to manually add the OSX SDK system root path.

Follow-up for #1794.

@tmpsantos review?


tmpsantos:
:shipit: :+1:

PR #1840, opened 2015-07-06
Updates to labels along lines

nickidlugash:
Port of https://github.com/mapbox/mapbox-gl-js/pull/1283.

- Removes anchors that are too close to another anchor with the same text
- Takes into account icon size for calculating label length
- Recalculates spacing for long labels
- Adjusts offsets to first anchors if line is continued from outside the tile boundary

### DIFF
--- a/src/mbgl/text/get_anchors.cpp
+++ b/src/mbgl/text/get_anchors.cpp
@@ -8,10 +8,10 @@
 namespace mbgl {
 
 Anchors resample(const std::vector<Coordinate> &line, const float offset, const float spacing,
-        const float angleWindowSize, const float maxAngle, const float labelLength, const bool placeAtMiddle) {
+        const float angleWindowSize, const float maxAngle, const float labelLength, const bool continuedLine, const bool placeAtMiddle) {
 
     float distance = 0;
-    float markedDistance = offset != 0.0f ? offset - spacing : 0;
+    float markedDistance = offset - spacing;
 
     Anchors anchors;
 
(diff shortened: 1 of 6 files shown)


mb12:
Is there a plan to add automated tests as well for these changes?

nickidlugash:
@mb12 not sure yet – not up to that point right now (@ansis – should I update /test in addition to mapbox-gl-test-suite?)

@ansis right now I’m seeing a fair amount of differences between js and native…in a lot of areas, the overall density and evenness of labels seems pretty similar (even if the specific labels themselves are not identical), but in dense urban areas, js seems noticeably better. I'm going to keep digging a bit on this. Sync up tomorrow?

JS:
screen shot 2015-07-06 at 6 03 08 pm

Native:
screen shot 2015-07-06 at 6 03 58 pm

src/mbgl/text/get_anchors.cpp
```diff
@@ -43,37 +46,79 @@ Anchors resample(const std::vector<Coordinate> &line, const float offset, const
...
-    // Add a bit of extra offset to avoid collisions at T intersections.
-    const float labelLength = right - left ? right - left : spacing;
-    const float extraOffset = glyphSize * 2;
-    const float offset = std::fmod((labelLength / 2 + extraOffset) * boxScale * overscaling, spacing);
-
-    return resample(line, offset, spacing, angleWindowSize, maxAngle, labelLength * boxScale, false);
+    
+    const float labelLength = fmax(
+                               (textLeft - textRight) != 0.0f ? textRight - textLeft : 0,
+                                       (iconLeft - iconRight) != 0.0f ? iconRight - iconLeft : 0);
```
ansis: this could be written as `fmax(textRight - textLeft, iconRight - iconLeft)`

nickidlugash:
@ansis thanks. I updated `labelLength`.

Issue #1724, opened 2015-06-14
More descriptive error when trying to load custom styles


friedbunny:
Trying to load a legacy custom style:

```objc
mapView.styleUrl = [NSURL URLWithString:@"mapbox://username.stylehash"];
```

... returns a 404 instead of informing developers that their style isn't compatible with GL.

jfirebaugh:
There's now a `Response::NotFound` enum value that can be used to implement this cleanly.

PR #1841, opened 2015-07-07
Improve style 404 error

friedbunny:
When trying to load a remote `mapbox://userid.mapid` style that exists but isn't compatible with GL, we throw: `[ERROR] {Map}[Setup]: loading style failed: HTTP status code 404`, which is misleading and has led to numerous support enquires.

This PR checks if the style is Mapbox-hosted and 404'ing, then throws: `[ERROR] {Map}[Setup]: style mapbox://userid.mapid could not be found or is not compatible`

Ideally the server would respond differently, but this is a prudent stopgap for now.

/cc @mapbox/mobile @kkaefer @tmpsantos @brunoabinader

### DIFF
--- a/src/mbgl/map/map_context.cpp
+++ b/src/mbgl/map/map_context.cpp
@@ -112,7 +112,12 @@ void MapContext::setStyleURL(const std::string& url) {
         if (res.status == Response::Successful) {
             loadStyleJSON(res.data, base);
         } else {
-            Log::Error(Event::Setup, "loading style failed: %s", res.message.c_str());
+            if (res.message.substr(res.message.length() - 3) == "404" &&
+                (styleURL.find("mapbox://") != std::string::npos)) {
+                Log::Error(Event::Setup, "style %s could not be found or is not compatible", styleURL.c_str());
+            } else {
+                Log::Error(Event::Setup, "loading style failed: %s", res.message.c_str());
+            }
         }
     });
 }


friedbunny:
Fixes #1724.

src/mbgl/map/map_context.cpp
```diff
@@ -112,7 +112,12 @@ void MapContext::setStyleURL(const std::string& url) {
         if (res.status == Response::Successful) {
             loadStyleJSON(res.data, base);
         } else {
-            Log::Error(Event::Setup, "loading style failed: %s", res.message.c_str());
+            if (res.message.substr(res.message.length() - 3) == "404" &&
```
tmpsantos: Maybe we could do this here:
https://github.com/mapbox/mapbox-gl-native/blob/master/platform/darwin/http_request_nsurl.mm#L254

So we don't need to add knowledge of the FileSource backend to the MapContext.

tmpsantos: Hmmm, apparently we could be done by adding an `else if (responseCode == 404) {` there.

friedbunny: I'll definitely defer to you on specific location (and this code is super-ugly right here), but it would be nice to add this check only once — especially since it's not platform dependent.

tmpsantos: The way I see to share this code would be create something like `mbgl::util::HTTPResponseStatusCodeToString(uint32_t)` and call it from the `http_request_*`.

But IMO it is also OK do it by adding an `else if` on both `http_request_nsurl.mm` and `http_request_curl.cpp` and duplicate the string with the error message, which is how we do it today.

jfirebaugh:
We need to work out an alternative to looking for "404" in a particular location in the string. That's much too fragile.

ljbade:
Why not use the HTTP integer status code from the Response object? Not sure how hard that is to get at here.

jfirebaugh:
`Response` isn't exactly HTTP specific -- it can also be a result from an `asset://` URL, which isn't HTTP-based. Faking an HTTP status code for that case is one possible solution, but I'm not sure it's the best one.

kkaefer:
Easiest solution is to add a more detailed status code to `Response`. We switched to a simple boolean flag because there are essentially only two responses; we got the data, or we didn't. In the latter case, passing an error code to the calling application is purely informational, but we can add that code back, designating it as such.

jfirebaugh:
Closing here -- good discussion, per which we need a more robust solution for this.

Issue #779, opened 2015-01-27
redraw continuously long enough for dash transitions to finish


ansis:
Dashed lines sometimes get stuck in the middle of a transition. It should repaint long enough for the transition to finish. I think this might also be a problem with labels fading in.

from @ljbade:
![screen shot 2015-01-26 at 6 28 26 pm](https://cloud.githubusercontent.com/assets/1421652/5912458/2acd7136-a589-11e4-9924-b251bc985c82.png)

ljbade:
Adding to the list that faded labels also need to redraw long enough.

brunoabinader:
#1548 should fix this.

PR #1548, opened 2015-07-07
Transitions fail to complete when zooming

brunoabinader:
When zooming with a scroll wheel or with pinch gestures, we sometimes fail to finish the fade transition (both fading in and fading out):

image

### DIFF
--- a/src/mbgl/map/map_context.cpp
+++ b/src/mbgl/map/map_context.cpp
@@ -343,7 +343,7 @@ MapContext::RenderResult MapContext::renderSync(const TransformState& state, con
 
     return RenderResult {
         style->isLoaded(),
-        style->hasTransitions()
+        style->hasTransitions() || painter->needsAnimation()
     };
 }
 


brunoabinader:
Review @kkaefer @jfirebaugh?

kkaefer:
While this fixes the bug, it also means that the map is continuously rendering at 60fps, using system resources and battery. Instead, we need to figure out in what circumstances we need to rerender (= when fading in/out), and not render otherwise.

brunoabinader:
Closing this PR in favour of pushing these into a `v8` branch.

brunoabinader:
I've cracked my head a bit these past days trying to figure out why this was not working. It turns out that we've missing checking if `Painter::needsAnimation()` along with `Style::hasTransitions()` in `MapContext::renderSync()`. I've learned a lot from the code while resolving this bug, and thus got the chance to do some other fixes and cleanups along the way:

Relevant changes:
- Unify default transition values in `MapData`:
   - Added `defaultFadeDuration` and `defaultTransitionDelay` to `MapData`;
   - `Painter` & `StyleParser` now receives a reference to `MapData`;
   - As previously seen on the code:
      - 300ms is the default fade duration;
      - 0ms is the default transition duration;
   - We no longer pass the current time point to `Style`, since it now uses `MapData.animationTime`, which gets updated in `MapContext::update()`.
   - Updated `StyleParser` check to use a mock `MapData`;
- Pass {`Duration`,`TimePoint`} by const ref whenever possible;
- Unify time point for Mapcontext::update:  
    - Let style classes cascade use the same time point as the one used to
      recalculate style.
    - Cleaned up `MapContext::update()` to return early whenever possible.
    - Cleaned up `MapContext::loadStyleJSON()` to avoid redundant calls to style
      classes cascade.

/cc @jfirebaugh @kkaefer @tmpsantos @tmcw

brunoabinader:
> What about moving these commits not exactly related to the fix to a separated PR?

:+1: will do, thanks!

brunoabinader:
There are changes in #1888 (specifically, we no longer pass raw fade transition duration values to `FrameHistory::needsAnimation()` in `Painter::needsAnimation()`) that prevents this fix from compiling. Will wait for #1888 to land first before updating this.

brunoabinader:
Got this in https://travis-ci.org/mapbox/mapbox-gl-native/jobs/70906410

brunoabinader:
Build failure cause is not related to these changes. See #1887 for details.

brunoabinader:
This should fix #779 as well.

PR #1842, opened 2015-07-07
Replace size() with empty() whenever possible

brunoabinader:
This is good because `empty()` is O(1) while `size()` _might be - depends on platform and/or compiler version -_ O(n).

### DIFF
--- a/src/mbgl/geometry/glyph_atlas.cpp
+++ b/src/mbgl/geometry/glyph_atlas.cpp
@@ -61,7 +61,7 @@ Rect<uint16_t> GlyphAtlas::addGlyph(uintptr_t tileUID,
     }
 
     // The glyph bitmap has zero width.
-    if (!glyph.bitmap.size()) {
+    if (glyph.bitmap.empty()) {
         return Rect<uint16_t>{ 0, 0, 0, 0 };
     }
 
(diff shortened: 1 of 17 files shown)


src/clipper/clipper.cpp
```diff
@@ -2013,7 +2013,7 @@ void Clipper::InsertLocalMinimaIntoAEL(const cInt botY)
 
     //if any output polygons share an edge, they'll need joining later ...
     if (Op1 && IsHorizontal(*rb) && 
-      m_GhostJoins.size() > 0 && (rb->WindDelta != 0))
+      !m_GhostJoins.empty() && (rb->WindDelta != 0))
```
1ec5: This file comes from the [polyclipping](https://sourceforge.net/p/polyclipping/) library. In the past, we’ve [attempted](https://github.com/mapbox/mapbox-gl-native/pull/1425#commitcomment-11049117) to keep this file in sync with the upstream project.

brunoabinader: Ah, cool - thanks for noticing :-) I'll send a patch upstream.

ljbade:
@kkaefer Should we merge this?

kkaefer:
Yes, this should be merged.

brunoabinader:
> Yes, this should be merged.

I am rebasing & updating the patch, then will proceed with merging it, thanks! :+1:

PR #1843, opened 2015-07-07
Enhance!

### LINKED ISSUE #963: raster tiles blurry
Possibly related to #924, but raster tiles are blurry on device. 

img_3099

The tile borders are 2x larger than they should be in debug mode. 

img_3100

In addition, sometimes it looks like maybe two different zoom levels are being drawn together? 

img_3101

/cc @1ec5 @jfirebaugh

1ec5:
For raster and video sources, fetch tiles from one zoom level higher when the map’s logical zoom level is halfway or greater. So z14.49 loads z14 tiles like before, but z14.5 loads z15 tiles. This behavior avoids blurriness in styles such as Satellite, given that the zoom level is rarely integral, especially after the user manually pinches to zoom. Leaflet apparently [does the same thing](https://github.com/Leaflet/Leaflet/blob/4101f017784f3c40ba744b4fd33bdc0401522983/src/layer/tile/GridLayer.js#L387).

Fixes #963.

/cc @incanus @peterqliu

### DIFF
--- a/src/mbgl/map/source.cpp
+++ b/src/mbgl/map/source.cpp
@@ -316,7 +316,13 @@ double Source::getZoom(const TransformState& state) const {
 }
 
 int32_t Source::coveringZoomLevel(const TransformState& state) const {
-    return std::floor(getZoom(state));
+    double zoom = getZoom(state);
+    if (info.type == SourceType::Raster || info.type == SourceType::Video) {
+        zoom = std::round(zoom);
+    } else {
+        zoom = std::floor(zoom);
+    }
+    return zoom;
 }
 
 std::forward_list<TileID> Source::coveringTiles(const TransformState& state) const {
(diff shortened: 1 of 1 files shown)


peterqliu:
:clap:

1ec5:
z  | Before | After
---|--------|-----
14.4 | before_14 4_thumb | after_14 4_thumb
14.6 | before_14 6_thumb | after_14 6_thumb

incanus:
Oh man this looks soooooo good. I have just been testing with latest `gl-styles@mb-pages` _Satellite_ style, which doesn't have any vectors. It'd be good to test with that, too, to see the interplay.

1ec5:
Yep, I'm working with mapbox/mapbox-gl-styles#161 locally.

incanus:
Raster styles look great too (color is off in this screenshot, but looks great on device). 

img_0652

mb12:
@1ec5 this could also have been fixed by just modifying the style.json file for raster and video source. In your style.json just add "tileSize": 256 for the raster source. This would have the equivalent effect because in Source::getZoom we also compute an offset like this. https://github.com/mapbox/mapbox-gl-native/blob/58b93553c10b77eedd573104b56d63916c635cbb/src/mbgl/map/source.cpp#L307

1ec5:
@mb12, the Satellite style in question did already have [`"tileSize": 256`](https://github.com/mapbox/mapbox-gl-styles/blob/mb-pages/styles/satellite-v7.json#L54), but the issue here was that we needed to be more tolerant of downloading and displaying higher-resolution imagery than the truncated zoom level would call for. This comment illustrates the problem well: https://github.com/mapbox/mapbox-gl-js/issues/1030#issuecomment-76043779.

mb12:
Thanks a lot for the clarification @1ec5. .

PR #1844, opened 2015-07-07
legacy raster mapID API

incanus:
For your perusal and consideration. When #963 lands, this will be trivially doable. 

```objc
[self.mapView setRasterMapID:@"examples.map-pgygbwdm"];
```

Only concerns are around interference with `styleID`, such as in IB. I don't want to overthink this, and would punt at the earliest sign of complication, but if this can stand like this, it's an easy win. 

/cc @1ec5 @friedbunny

### DIFF
--- a/include/mbgl/ios/MGLMapView.h
+++ b/include/mbgl/ios/MGLMapView.h
@@ -211,6 +211,10 @@ IB_DESIGNABLE
 *   To display the default style, set this property to `nil`. */
 @property (nonatomic, null_resettable) NSURL *styleURL;
 
+/** Compatibility API to use legacy raster-only tile sources. Expected format is e.g. `examples.map-pgygbwdm`.
+*   @param mapID A legacy Mapbox raster-only map ID. */
+- (void)setRasterMapID:(NSString *)mapID;
+
 /** Currently active style classes, represented as an array of string identifiers. */
 @property (nonatomic) NS_ARRAY_OF(NSString *) *styleClasses;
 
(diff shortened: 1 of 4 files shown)


incanus:
img_0653

1ec5:
Great idea. This will address plenty of confusion, some of which we're already seeing today. What if we treat any style ID starting with "raster/" as a raster style that we can then synthesize on the fly? We could do that at the mbgl level and catch both the style ID and style URL interfaces in one go.

incanus:
Hmm, putting magic tokens in feels complex to me. I'd rather keep it in a separate API that we can point people to, but that is 1) clearly labeled in a sort of deprecating, maybe-don't-really-use-this way, and 2) Can be removed in future with a more clean break.

1ec5:
In that case, maybe we just show how to create one of these compatibility styles in help documents?

incanus:
> In that case, maybe we just show how to create one of these compatibility styles in help documents?

We could do that as well, but the point in this PR is to not have to create a style, just pass a `mapID` that you were using in our raster tools.

1ec5:
I’m only in favor of this approach if we consider it a permanent part of the MGLMapView feature set. However, if the idea is to provide a temporary migration path for developers on the existing raster tools, I don’t think the added complexity is worth it.

The complexity would come in the form of a third way to populate an MGLMapView with content. (The PR doesn’t add an initializer, but presumably that would be necessary.) All three ways are mutually exclusive yet available at the same time. Simplifying initialization in #1184 made it possible to ensure KVO compliance, coherent setup instructions (as opposed to a choose-your-adventure), and minimal opportunity for race conditions, and I think we should be wary of pulling back that simplification.

friedbunny:
We show devs how to make their own raster JSON shim here: https://www.mapbox.com/ios-sdk/examples/raster-styles/

jfirebaugh:
I like that example. I think we should go with a GL style-based approach, like that does, and not bake legacy support into the SDK like this.

incanus:
:+1:

PR #1852, opened 2015-07-08
refs #1125, 1813: update user dot tracking in one place

incanus:
Now that we have cleanup around the map viewport change notifications, this only needs to happen in one place: the GL frame draw loop.


1ec5:
This is fine for now, but we’ll have to undo it for #1582 / #1041.

incanus:
:+1:

PR #1854, opened 2015-07-08
Assert if -mapView:symbolNameForAnnotation: is implemented

### LINKED ISSUE #1851: Assert if -mapView:symbolNameForAnnotation: is implemented
#941 introduced `-[MGLMapViewDelegate mapView:imageForAnnotation:]` and [marked `-mapView:symbolNameForAnnotation:` unavailable](https://github.com/mapbox/mapbox-gl-native/blob/445d3afa4f3aa1c603fab936756602b563cb4f7a/include/mbgl/ios/MGLMapView.h#L352). But marking a delegate method unavailable just means you can’t call it directly. If you implemented it for a previous release of Mapbox GL, your app will continue to run, but your markers will revert to the default. We should assert if the old method is implemented, with a note to move to the new API.

/ref mapbox/mapbox-gl-styles#162
/cc @incanus

1ec5:
Fixes #1851.

/cc @incanus

### DIFF
--- a/include/mbgl/ios/MGLMapView.h
+++ b/include/mbgl/ios/MGLMapView.h
@@ -345,10 +345,6 @@ IB_DESIGNABLE
 
 /** @name Managing the Display of Annotations */
 
-/** Returns the style's symbol name to use for the marker for the specified point annotation object.
-*   @param mapView The map view that requested the annotation symbol name.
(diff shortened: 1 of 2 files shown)


incanus:
This looks good, but my only concern is every `addAnnotations:` from here to eternity doing this runtime check now. Maybe not a big deal?

Issue #1740, opened 2015-06-16
Port circle render type (v8 styles)


kkaefer:
We need to port the [point symbol type](https://github.com/mapbox/mapbox-gl-style-spec/issues/276) to C++. This type will be able to render plain circles with arbitrary color, which is very useful for visualizing many small features that don't require a particular icon.

There is a [JS implementation](https://github.com/mapbox/mapbox-gl-js/pull/1241) already which contains the shaders and a general outline of what the feature will look like. Porting it includes:

- Parse the stylesheet into the correct structures (see style spec for how the JSON looks like)
- When parsing vector tiles into buckets, generate the OpenGL buffers required for rendering
- When rendering, add code that renders the buffers with the correct shaders
- Add tests for correct bucket creation (may not exist in JS yet)
- Add visual rendering tests (may not exist yet in the test suite)

jfirebaugh:
⇢ #1766

PR #1863, opened 2015-07-09
[WIP] circle layer

brunoabinader:
Implements breaking changes of mapbox/mapbox-gl-style-spec#264:

- [X] Circle type

Part of mapbox/mapbox-gl-native#1740. Follows mapbox/mapbox-gl-js#1192.

### DIFF
--- a/include/mbgl/style/style_properties.hpp
+++ b/include/mbgl/style/style_properties.hpp
@@ -45,6 +45,17 @@ struct LineProperties {
     }
 };
 
+struct CircleProperties {
+    inline CircleProperties() {}
+    float radius = 1.0f;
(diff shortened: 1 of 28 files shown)


brunoabinader:
As per discussion with @tmcw, we believe it would be a good idea to merge this PR with #1390. However, #1390 seems a bit old so rebasing it against master is quite challenging.

jfirebaugh:
That's correct, this needs to go onto the `v8types-merge` branch from #1390 (and that branch should be renamed to `v8` and the PR rerolled). I'm hopefully going to be working on v8 full time shortly and can look into doing that.

jfirebaugh:
I'm rebasing this onto an updated v8 branch.

PR #1869, opened 2015-07-09
Create GLKView before binding it and resuming

### LINKED ISSUE #1460: Crash in background — glFinish_Exec, gpus_ReturnNotPermittedKillClient, libGPUSupportMercury

### LINKED ISSUE #1855: Blank map after background location update
#1821 fixes #1460 in that background location updates no longer crash the app in the background. However, it introduces a serious regression in its stead: manually opening the app past that point causes it to launch with a blank, totally black map that you can't interact with until you switch out and back into it with the app switcher.

Opening the app apparently causes it to be revived from the suspended state, and for some reason MGLView is either failing to receive the entered-foreground notification or is bailing because the application state is still "background", not even "inactive".

This behavior is worse than before, because system would automatically relaunch the process if it crashed in the background. Now you just get a defective view.

/cc @incanus @friedbunny

1ec5:
Or: don’t resume the map until the application has entered the foreground.

UIApplicationWillEnterForegroundNotification is posted before UIApplicationDidBecomeActiveNotification, while the application is still in the background. That caused -createGLView to bail but did not cause the rest of -wakeGL: to bail. We would create the GLKView only after attempting to bind it (a no-op), resulting in a black view.

Fixes #1855. Fixes #1460 real good.

/cc @incanus @friedbunny

### DIFF
--- a/platform/ios/MGLMapView.mm
+++ b/platform/ios/MGLMapView.mm
@@ -213,7 +213,6 @@ std::chrono::steady_clock::duration secondsAsDuration(float duration)
         [self createGLView];
     }
 
-
     self.accessibilityLabel = @"Map";
     self.backgroundColor = [UIColor clearColor];
     self.clipsToBounds = YES;
@@ -251,7 +250,7 @@ std::chrono::steady_clock::duration secondsAsDuration(float duration)
     //
     _annotationIDsByAnnotation = [NSMapTable mapTableWithKeyOptions:NSMapTableStrongMemory valueOptions:NSMapTableStrongMemory];
 
-    _annotationImages = [NSMutableDictionary new];
+    _annotationImages = [NSMutableDictionary dictionary];
 
     std::string defaultSymbolName([MGLDefaultStyleMarkerSymbolName UTF8String]);
     _mbglMap->setDefaultPointAnnotationSymbol(defaultSymbolName);
@@ -376,9 +375,7 @@ std::chrono::steady_clock::duration secondsAsDuration(float duration)
 
 - (void)createGLView
 {
-    if (_context || [UIApplication sharedApplication].applicationState == UIApplicationStateBackground) {
-        return;
-    }
+    if (_context) return;
     
     // create context
     //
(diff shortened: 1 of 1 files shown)


friedbunny:
Awesome work on tracking this sneaky bastard down.

1ec5:
Now I just need to figure out why the tests fail on Travis. Something’s up…

incanus:
A+ work here @1ec5. 

max0060embttm00290x249

PR #1873, opened 2015-07-10
add instructions to install Android Support Repository

ansis:
add instructions to install Android Support Repository


bleege:
Will want the Repository and the Library.

screen shot 2015-09-16 at 3 23 10 pm

PR #1884, opened 2015-07-13
Do not return a value on a "[]() -> void"

tmpsantos:
This is an undefined behavior, and it is a mystery why the
compiler was not complaining about it.


kkaefer:
Good catch

PR #1886, opened 2015-07-13
Include iOS README in binary pod

1ec5:
This PR adds the iOS-specific README to the build folder that gets zipped up and distributed as a CocoaPods binary pod. Without this README, CocoaPods [dings](https://cocoapods.org/pods/MapboxGL/quality) the MapboxGL pod big time in its quality metrics for lacking a README. We’ve been compiling this README to serve as the index page of the API documentation that gets generated, but a README also needs to set right alongside the library and resource bundle.

/cc @incanus @friedbunny

### DIFF
--- a/scripts/ios/package.sh
+++ b/scripts/ios/package.sh
@@ -113,7 +113,7 @@ git fetch --tags
 DOCS_VERSION=$( git tag --sort -v:refname | grep -v '\-rc.' | sed -n '1p' | sed 's/^v//' )
 rm -rf /tmp/mbgl
 mkdir -p /tmp/mbgl/
-README=/tmp/mbgl/GL-README.md
+README=/tmp/mbgl/README.md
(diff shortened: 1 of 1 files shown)


friedbunny:
:+1:

https://guides.cocoapods.org/making/quality-indexes.html

>README Scoring

>The README score is based on an algorithm that looks at the variety of the bundled README. You can run the algorithm against any URL here on scoreme.report. A README is the front-page of your library, it can provide an overview of API or show what the library can do.

>Strange as it sounds, if you are providing a binary CocoaPod, it is worth embedding your README.md inside the zip. This means CocoaPods can use it to generate your Pod page.

incanus:
Great work — just noticed this ranking business recently myself. Good discussion on the mailing list at https://groups.google.com/forum/#!topic/cocoapods/Pvjsd57zKAk as well.

PR #1888, opened 2015-07-14
Clean up transition durations & time points

brunoabinader:
Relevant changes:
- Unify default transition values in `MapData`:
   - Added `defaultFadeDuration` and `defaultTransitionDelay` to `MapData`;
   - `Painter` & `StyleParser` now receives a reference to `MapData`;
   - As previously seen on the code:
      - 300ms is the default fade duration;
      - 0ms is the default transition duration;
   - We no longer pass the current time point to `Style`, since it now uses `MapData.animationTime`, which gets updated in `MapContext::update()`.
   - Updated `StyleParser` check to use a mock `MapData`;
- Pass {`Duration`,`TimePoint`} by const ref whenever possible;
- Unify time point for Mapcontext::update:
    - Let style classes cascade use the same time point as the one used to
      recalculate style.
    - Cleaned up `MapContext::update()` to return early whenever possible.
    - Cleaned up `MapContext::loadStyleJSON()` to avoid redundant calls to style
      classes cascade.

Part of #1548.

/cc @jfirebaugh @kkaefer @tmpsantos @tmcw

### DIFF
--- a/include/mbgl/map/map.hpp
+++ b/include/mbgl/map/map.hpp
@@ -79,8 +79,8 @@ public:
     void setClasses(const std::vector<std::string>&);
     std::vector<std::string> getClasses() const;
 
-    void setDefaultTransitionDuration(Duration = Duration::zero());
-    Duration getDefaultTransitionDuration();
+    void setDefaultTransitionDuration(const Duration& = Duration::zero());
+    Duration getDefaultTransitionDuration() const;
     void setStyleURL(const std::string& url);
     void setStyleJSON(const std::string& json, const std::string& base = "");
     std::string getStyleURL() const;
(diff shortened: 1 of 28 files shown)


include/mbgl/map/map.hpp
```diff
@@ -79,8 +79,8 @@ class Map : private util::noncopyable {
     void setClasses(const std::vector<std::string>&);
     std::vector<std::string> getClasses() const;
 
-    void setDefaultTransitionDuration(Duration = Duration::zero());
-    Duration getDefaultTransitionDuration();
+    void setDefaultTransitionDuration(const Duration& = Duration::zero());
+    const Duration getDefaultTransitionDuration() const;
```
kkaefer: returning a `const` integral type isn't meaningful.

brunoabinader: Good point, I guess that went through on my regular expression :ghost: will remove that.

tmpsantos:
:+1:

brunoabinader:
> `const TimePoint& now`?

This goes away in 28c6d44 :-)

include/mbgl/map/map.hpp
brunoabinader: Good point, I guess that went through on my regular expression :ghost: will remove that.

PR #1889, opened 2015-07-14
Map view gets black screen after changing style if default transition duration != 0

brunoabinader:
I have found this while debugging for #1548. If the client manually sets a default transition duration [(example)](https://github.com/mapbox/mapbox-gl-native/blob/master/platform/default/glfw_view.cpp#L109), after changing a style the screen goes black (with only symbols appearing) until you trigger another view update (like zooming in and out of screen).

Steps to reproduce:
1. Open Mapbox GL native test application
2. Press `R` key (which triggers a client-side update on the default transition duration)
3. Press `S` key
screen shot 2015-07-14 at 17 46 01
/cc @jfirebaugh @tmcw @kkaefer @tmpsantos

### DIFF
--- a/src/mbgl/map/map_context.cpp
+++ b/src/mbgl/map/map_context.cpp
@@ -133,6 +133,9 @@ void MapContext::loadStyleJSON(const std::string& json, const std::string& base)
     style->setJSON(json, base);
     style->setObserver(this);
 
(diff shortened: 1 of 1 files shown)


tmpsantos:
:+1:

PR #302, opened 2015-07-14
consider supporting transition delays without duration

brunoabinader:
If the default transition duration is specified in the caller's code, it might sometimes be useful to support only `delay` specified in the class. Currently we [only transition properties which explicit durations](https://github.com/mapbox/mapbox-gl-native/blob/d15fb9ca97e737997f93c69cbebfa54fd9305d8f/src/style/style.cpp#L37).

### DIFF
--- a/include/mbgl/map/map.hpp
+++ b/include/mbgl/map/map.hpp
@@ -84,6 +84,10 @@ public:
 
     void setDefaultTransitionDuration(const Duration& = Duration::zero());
     Duration getDefaultTransitionDuration() const;
+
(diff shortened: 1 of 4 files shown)


brunoabinader:
As per discussion with @tmcw, I've created an issue - mapbox/mapbox-gl-js#1379 - to provide the same feature in mapbox-gl-js.

incanus:
Looks great :+1:

jfirebaugh:
Hrm... I actually want to eliminate the `set/getDefaultTransitionDuration` APIs, not expand them. See #1025.

PR #1885, opened 2015-07-16
Only print Sprites not found messages after the sprites are loaded and parsed

tmpsantos:
Sprites are loaded asynchronously and if we get the tiles before the sprites, we are going to see a lot of "Sprites not found" messages, which can be misleading. We should only print these messages after the sprite define on the Style is loaded.


tmpsantos:
The `dot.sdf` sprite is indeed missing. The `satellite-v7.json` style is using https://www.mapbox.com/mapbox-gl-styles/sprites/bright.json that doesn't have it.

Issue #1716, opened 2015-06-12
Annotations API In Android


bleege:
Implement marker annotations in Android similar to how it was implemented in iOS in ecb8e2800d0fbc4e4f843b0d13efc28038ede3ef.

hallahan:
I forked the repo and am working on a branch that attempts to add a point annotation to the map via a call to the native [Map::addPointAnnotation](https://github.com/hallahan/mapbox-gl-native/blob/master/src/mbgl/map/map.cpp#L240).

https://github.com/hallahan/mapbox-gl-native/commit/ef31fc5aed4e07b3d9e98b25180b8f809d51a5cf

The JNI part of things is straightforward, and I have verified that `addPointAnnotation` in the C++ Map is indeed being called. However, my single point annotation is not successfully rendering on the map.

I'm studying how annotations work in the core library in this [MapboxGL Native Annotations Walkthrough](https://gist.github.com/hallahan/eb2de11bdd4619a11eec). @incanus, one conceptual question I have deals with that final conversion to the vector space of a tile for a given point:

```cpp
const Coordinate coordinate(extent * (pp.x * z2 - x), extent * (pp.y * z2 - y));
```

__We just computed `x`, which is a 32 bit unsigned int. We then do `pp.x * z2 - x` which just subtracts the truncated int from the complete double. How does that then relate to determine the correct coordinate in the vector space of the given tile?__

I was able to follow the code through setting up the annotation vector tiles as well as telling the map what to update. 

__Where do I look to learn about how that actual sprite is rendered on the map?__ My guess is that the actual problem here is that somehow the sprite itself is not being rendered in Android. I don't think the problem at hand is involved with the cutting of the annotation vector tiles.

hallahan:
Note, stepping through all of the function calls mentioned in the [point annotation walkthrough](https://gist.github.com/hallahan/eb2de11bdd4619a11eec) seem to execute fine when poking around my android device in `gdb`. My thought is that the annotation is actually added to the map just fine. Maybe the marker sprite is just not being rendered?

hallahan:
[...]

PR #1899, opened 2015-07-17
Android Annotations Basic Functionality

hallahan:
We've got basic functionality for adding annotations to AndroidGL. For the proof of concept that we can add markers, polylines, and polygons in the same way that you can on the Google Maps API, it might be a good time to pull into master. The overall Android Annotations API work is being tracked on issue #1716

[Google Maps API Breakdown](https://gist.github.com/hallahan/7b4a3aebe4d7a0b41a5a)

You can see the basic calls to add markers, polylines, and polygons in [MainActivity.java of the sample app](https://github.com/hallahan/mapbox-gl-native/blob/0f56c2b0d5766996d51c73f5e131ec8df21d5fca/android/java/MapboxGLAndroidSDKTestApp/src/main/java/com/mapbox/mapboxgl/testapp/MainActivity.java#L292). 

@ljbade - After spending some time working in [jni.cpp](https://github.com/hallahan/mapbox-gl-native/blob/0f56c2b0d5766996d51c73f5e131ec8df21d5fca/android/cpp/jni.cpp), we've got a whopping 1567ln file. What are your thoughts on the best way to break up this massive file (at least giving the annotation work it's own file)?

I also think that the [annotations module](https://github.com/hallahan/mapbox-gl-native/tree/0f56c2b0d5766996d51c73f5e131ec8df21d5fca/android/java/MapboxGLAndroidSDK/src/main/java/com/mapbox/mapboxgl/annotations) is the perfect candidate for some unit tests. I'd like to do this in further sprints.

There are several functionalities we need that arefound on Google's API that are currently not implemented in the C++ renderer, namely:

* Circles
* Polygon Holes
* Polygon Stroke Width

Other functionalities that need to be implemented include:

* Alpha
* Marker Rotation
* Marker Anchor
* Line / Polygon anti-aliasing
* Popup with Title & Snippet

One extra I wired up is the ability to [specify the name of the sprite](https://github.com/hallahan/mapbox-gl-native/blob/0f56c2b0d5766996d51c73f5e131ec8df21d5fca/android/java/MapboxGLAndroidSDKTestApp/src/main/java/com/mapbox/mapboxgl/testapp/MainActivity.java#L303) you would like as your marker. We also need to allow custom images like what is being done in iOS.


hallahan:
Screenshots of Annotations via Overpass GeoJSON extracts:

Campsites North of SF:


All Gas Stations & Campsites between SF & Sacramento:


All Restaurants & Gas Stations in SF:


CA-128 Highlighted ArcMap Style:


Buildings in Neighborhood of Mapbox:


Buildings Closeup:

nulvinge:
Great work, this almost worked out of the box for me.

Though, I had a problem with it taking over a minute to load about 800 polygons.
I improved the loading time by making it possible to add multiple polygons in one function call:
https://github.com/TrueFlow/mapbox-gl-native/commits/android-annotations
That also required some more memory management.

I would suggest restructuring most of the code to always handle multiple map elements, since that is most efficient, and add a wrapper in MapView that enables single elements.

hallahan:
@nulvinge - very cool! I was thinking about doing that as well. I love how quickly someone looked at this and made this better!

ljbade:
Awesome work @hallahan.

You can likely split up the JNI functions into different files but you will still need the long function list in `JNI_OnLoad`. It might be possible to use a series of static variables at the top of each file that contains parts of this list and then is combined later if you wan't to keep functions and definitions in the same file.

@nulvinge I imagine there is some work to be done on reducing JNI overhead. Being able to call a single function with an array or a list of polygons that is passed through to the C++ layer would be a good first step.

vespakoen:
Great work!

I mainly am replying to get updates on this.

I am especially interested in the ability to restyle a certain marker / line dynamically, is that possible with this approach? (using annotations).

At the moment I have turned my geojson into vector tiles with mapbox studio / tippecanoe which works fantastic but unfortunately I am not able to style individual features using that method yet.

bleege:
This is really great work @hallahan!  Thank you for making this happen!

PR #1904, opened 2015-07-20
fix iOS view sync regression

brunoabinader:
https://github.com/mapbox/mapbox-gl-native/pull/1903 introduced a regression in the improvements from #1125. The addition of `nudgeTransitions` was not related to https://github.com/mapbox/mapbox-gl-native/pull/1548, but rather happened in https://github.com/mapbox/mapbox-gl-native/commit/b8388168dd130c67c77254565cdb576df7905915 to allow iOS to insert position updates to native views _between_ the GL frame draw and any changes to the `Transform`, so that the views were in sync with the GL frame. 

@brunoabinader

### DIFF
--- a/include/mbgl/map/map.hpp
+++ b/include/mbgl/map/map.hpp
@@ -67,7 +67,10 @@ public:
     void renderStill(StillImageCallback callback);
 
     // Triggers a synchronous or asynchronous render.
-    void renderSync();
+    bool renderSync();
+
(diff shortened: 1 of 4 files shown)


brunoabinader:
> But perhaps that's more obfuscated.

:+1: on that. The code in there explicitly creates the `dirty` variable which could also be removed, but most probably left there for better readability.

incanus:
:+1:

PR #1916, opened 2015-07-22
Android annotations improvements

nulvinge:
More work on #1716. This pull request contains improvements on #1899.

This lowers my loading times from more than 1 minute to about 2 seconds in adding and removing polygons.
There was also a bug with colors not being remapped to a correct range.
And exposed setSprite.


bleege:
Thanks @nulvinge!  I'll take a look at this now.

bleege:
:+1: Thanks again for contributing!

PR #1919, opened 2015-07-23
Allow for Mapbox logo to be placed on either side of the map.

bwhtmn:
Created a property to choose whether the Mapbox logo should be shown on the right or left bottom corner of the maps. Defaults to bottom left, the same as current behavior. Having a bottom right placement option is convenient for developers who already have other UI components in the bottom left corner.


incanus:
While I appreciate the code here, we've got a discussion already in https://github.com/mapbox/mapbox-gl-native/issues/1781 about generic moving of the logo as well as the attribution button and the compass, why this gets complex, and why we will probably just stick with hiding and showing for now.

PR #1921, opened 2015-07-23
[android] Avoid nudging transitions if a gesture is in progress

brunoabinader:
We're now using `nudgeTransitions()` to tell the Map view that we want to update it. However, if we're on a gesture movement, `update()` gets called too fast and causes general slowdown. This check ensures we only call for `nudgeTransitions()` after all gesture events are finished. Fixes the issue pointed out by #1548 on Android.

/cc @incanus @kkaefer @tmpsantos


incanus:
Looks good; does this issue not affect iOS?

brunoabinader:
> Looks good; does this issue not affect iOS?

I haven't tested that, I'll borrow my wife's iPad today to check this out :+1:

PR #1925, opened 2015-07-23
Bail if setting user dot to invalid location

### LINKED ISSUE #1912: MGLUserAnnotation not displaying
Looking at Stackoverflow post [here](http://stackoverflow.com/questions/31546489/how-to-display-user-location-marker-in-mapbox-gl) it seems like `MGLUserLocation::setLocation` tests the distance between new and old locations before actually setting the new value. Because the initial latitude and longitude as set to `MAXFLOAT`, one of the conditionals checking distance always returns zero, preventing the new location value from being set:

```
 - (void)setLocation:(CLLocation *)newLocation {
    if ([newLocation distanceFromLocation:_location] /* always zero? */ 
     && newLocation.coordinate. latitude != 0 
     && newLocation.coordinate.longitude != 0)`...
```

I think this may be preventing the display of the MGLUserLocation annotation.

To test out my theory, I made the small change below (`MGLUserLocation.m`) and it seems to have worked - the MGLUserAnnotation now renders on my map as a pulsating circle.

```
 - (void)setLocation:(CLLocation *)newLocation
 {
    // new code to test my theory out
     if(_location.coordinate.latitude == MAXFLOAT || _location.coordinate.longitude == MAXFLOAT) {
          [self willChangeValueForKey:@"location"];
          _location = newLocation;
          [self didChangeValueForKey:@"location"];
    }`
    else if ([newLocation distanceFromLocation:_location] && newLocation.coordinate. latitude != 0 &&
             newLocation.coordinate.longitude != 0)
     {
         [self willChangeValueForKey:@"location"];
         _location = newLocation;
         [self didChangeValueForKey:@"location"];
     }
}
```

1ec5:
Just like `-[MGLMapView locationManager:didUpdateToLocation:fromLocation:]` does.

Fixes #1912.

/cc @incanus @bsudekum

### DIFF
--- a/platform/ios/MGLMapView.mm
+++ b/platform/ios/MGLMapView.mm
@@ -2309,7 +2309,7 @@ CLLocationCoordinate2D MGLLocationCoordinate2DFromLatLng(mbgl::LatLng latLng)
 {
     if ( ! _showsUserLocation || ! newLocation || ! CLLocationCoordinate2DIsValid(newLocation.coordinate)) return;
 
-    if ([newLocation distanceFromLocation:oldLocation] || ! oldLocation)
+    if (! oldLocation || ! CLLocationCoordinate2DIsValid(oldLocation.coordinate) || [newLocation distanceFromLocation:oldLocation])
     {
         self.userLocation.location = newLocation;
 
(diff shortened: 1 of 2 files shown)


incanus:
:+1:

PR #1926, opened 2015-07-24
This fixes latLngForPixel on Android

vespakoen:
It used GetDoubleField where it should have used GetFloatField.
This fixes it for me.

Thanks to the hint from @kkaefer on the meetup in Berlin!


incanus:
Looks great — thanks @vespakoen!

PR #1928, opened 2015-07-24
Avoid nudging transitions if a gesture is in progress #2

brunoabinader:
Following the same steps from #1548 and #1912 for GLFW and iOS ports.
Also removed a redundant check if transform state is changing since
we're already dealing with that.

/cc @incanus @kkaefer @tmpsantos

### DIFF
--- a/platform/default/glfw_view.cpp
+++ b/platform/default/glfw_view.cpp
@@ -336,7 +336,10 @@ void GLFWView::run() {
         const bool dirty = !clean.test_and_set();
         if (dirty) {
             const bool needsRerender = map->renderSync();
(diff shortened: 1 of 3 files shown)


incanus:
This seems to break iOS non-animated transitions. 

To reproduce: 

1. Start with map. 
1. Two-finger tap to (animated) zoom out. 
1. Map doesn't move.

brunoabinader:
> This seems to break iOS animated transitions.
Got it :+1: it seems self.isAnimatingGesture is not exactly what we want to check there.

incanus:
I know you merged already, but this is looking great in my testing @brunoabinader :+1:.

PR #1938, opened 2015-07-27
Update the user dot when the user moves

### LINKED ISSUE #1933: User dot does not move unless something else causes a redraw
On iOS, the user dot no longer moves on its own when user location tracking is off. If you cause the map to redraw, the user dot then updates once.

This is a recent regression, seen with and without #1925. Maybe a regression from #1852 or #1829?

/cc @incanus @friedbunny

1ec5:
Rolled back part of ed04e6d9f712437bc29502e134250cd979c74b64 for #1852. The user dot should move on its own even if the map has no other reason to rerender.

Fixes #1933.

/cc @incanus


incanus:
Aha! That nailed it — over-aggressive refactoring. Good find @1ec5.

PR #1964, opened 2015-07-29
initial implementation of Map::setView

mikemorris:
I'm not sure if something needs to change in `MapContext` or if I'm just implementing wrong, but calling `context->invokeSync(&MapContext::setView, *view);` results in the error `field type 'mbgl::View' is an abstract class`. Would love a quick review here from @kkaefer or @tmpsantos to check if I'm going in the right direction.

### DIFF
--- a/src/mbgl/map/map.cpp
+++ b/src/mbgl/map/map.cpp
@@ -15,7 +15,7 @@ namespace mbgl {
 Map::Map(View& view_, FileSource& fileSource, MapMode mode)
     : data(std::make_unique<MapData>(mode, view_.getPixelRatio())),
     context(std::make_unique<util::Thread<MapContext>>(util::ThreadContext{"Map", util::ThreadType::Map, util::ThreadPriority::Regular}, fileSource, *data)) {
-    setView(view_);
+    setView(&view_);
 }
 
 Map::Map(float pixelRatio, FileSource& fileSource, MapMode mode)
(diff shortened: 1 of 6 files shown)


mikemorris:
Got past the `invoke` trouble thanks to @tmpsantos' suggestion to make everything `View*`

src/mbgl/map/map_context.cpp
```diff
@@ -50,8 +47,15 @@ MapContext::~MapContext() {
     assert(!style);
 }
 
+void MapContext::setView(View* view_) {
+    assert(util::ThreadContext::currentlyOn(util::ThreadType::Map));
+
+    view = view_;
```
tmpsantos: You must call here `view->deactivate()` before replacing.

src/mbgl/map/map.cpp
```diff
@@ -43,22 +43,31 @@ void Map::resume() {
     paused = false;
 }
 
+void Map::setView(View* view_) {
+    view = view_;
+    transform = std::make_unique<Transform>(*view);
+    context->invoke(&MapContext::setView, view);
```
tmpsantos: You might need to make this a `context->invokeSync` to make sure it is safe to reuse the old view (i.e. `deactivate()` was called) in another map object right after calling this method.

jfirebaugh:
In https://github.com/mapbox/node-mapbox-gl-native/issues/142 I'm advocating we abandon this and take a different approach.

mikemorris:
-> https://github.com/mapbox/mapbox-gl-native/pull/2310

PR #1970, opened 2015-07-30
I made a naive selection and popup example

erf:
I made a naive selection and popup example for Marker annotations. Discussed in #894

Need some mappings (JNI) to native map to improve on this.


incanus:
Nice, this is a good start. I am working on basic pin-drop adding in our internal test app and will likely branch some work off of here as we iterate on this.

tobrun:
Closing, see #894

PR #1971, opened 2015-07-30
compensate for Android screen ratio in conversion routines

incanus:
Fixes bugs where these conversions are inaccurate because screen density isn't taken into account during viewport calculations.


ljbade:
Nice catch!

You should also fix up the other conversion functions and remove the compensation from the GPS marker code.

ljbade:
Oh silly me you did do both!

Issue #1804, opened 2015-06-30
support mapViewDidFinishRenderingMap:fullyRendered:NO


jfirebaugh:
Followup to #1431.

What remains unimplemented is support for fullyRendered:NO -- we currently do not surface tile errors in a way that made this easy to implement, so I'm punting for now.

mkrussel77:
We typically create VectorSources for data that changes and may or may not have anything visible.

Currently if there is a failure to download the tiles, it just looks to the user that there is nothing happening.  

Being able to report that there has been an error would let them know that there might be something there but we are unable to display it to them.

PR #1976, opened 2015-07-31
Document iOS region change delegate methods

friedbunny:
With #1026 the region delegate methods work reliably and as expected.

This commit adds `mapView:regionWillChangeAnimated:`, `mapViewRegionIsChanging:`, and `mapView:regionDidChangeAnimated:` to the official documentation.

/cc @1ec5 @incanus @jfirebaugh


friedbunny:
I didn't add docs/comments for the rendering methods because, while @jfirebaugh mostly finished them in #1431, there's still work to be done in #1804 before we can tell devs to use them.

Issue #1918, opened 2015-07-23
Change glyphs URL from /v4/fontstack to /fonts/v1


jakepruitt:
In order to unblock members of the team, update gl-js to use the `/fonts/v1` endpoint rather than the `/v4/fontstack`. This would allow users to upload their own fonts.

The current style-spec specifies that fonts be addressed with the following pattern:

```
"glyphs": "mapbox://fontstack/{fontstack}/{range}.pbf"
```

which does not have a user id in it. The glyph property of the style will have to eventually map to the /fonts/v1 endpoint with the format:

```
GET /fonts/v1/:owner/:face[,:face,:face...]/:start-:end.pbf
```

After talking with @yhahn, it seems like you must choose between changing the style spec or parsing the user from the `access_token`. I'm going to make a deal with the devil and parse the user from the access token and use that user in the request. Other opinions are very welcome.

cc/ @kkaefer @mikemorris @jfirebaugh @yhahn @tmcw

mikemorris:
Refs https://github.com/mapbox/mapbox-gl-style-spec/issues/309

jfirebaugh:
Looks like #1924 was merged and next actions are tracked in https://github.com/mapbox/mapbox-gl-style-spec/issues/309. Can this be closed?

PR #1991, opened 2015-08-04
Adjusting /fonts/ urls to normalize with /fonts/v1

jakepruitt:
This change will probably need to be a part of the v8 style spec, as per discussion at https://github.com/mapbox/gl-internal/issues/253#issuecomment-127437196. 

@jfirebaugh and @kkaefer what are the procedures for making this a part of V8?

refs #1918


jfirebaugh:
@lucaswoj is going to re-roll this with the changes discussed in https://github.com/mapbox/mapbox-gl-style-spec/issues/309.

PR #1992, opened 2015-08-04
fix crashes after network connection was restored

bwhtmn:
Mapbox GL works well without a network connection, but I found and fixed two crashes that happened after the network connection was restored. To repro the issue on iOS:
1. Launch mapbox sample app
2. Turn on airplane mode
3. Scroll around or zoom out to areas of the map that have not been previously downloaded
4. After ~10 seconds, turn airplane mode off


kkaefer:
Sorry, this slipped through my radar. The ultimate fix for this is described in https://github.com/mapbox/mapbox-gl-native/pull/2123#issuecomment-151451888.

PR #2003, opened 2015-08-06
Set thread priority on Linux

kkaefer:
We currently only have an implementation for Darwin, but not for the Linux platform.


brunoabinader:
:+1:

Issue #2007, opened 2015-08-06
Mobile connection optimization: Keep-Alive


tmcw:
@1ec5 @incanus @brunoabinader 

Given a high-latency but mildly-fast-once-the-download-has-started 3G network, how can we reduce round-trip time?

I'm a noob here but my two guesses:

1. Remove tile URL sharding: https://github.com/mapbox/mapbox-gl-native/blob/master/src/mbgl/map/source.cpp#L101 - this is a web optimization that is a disadvantage rather than an advantage on mobile
2. Keep-Alive: are we using it? With Cloudfront we can't use HTTP2 or SPDY but this gets us some of the way, potentially cutting down on negotiation and re-negotiation of tile load requests.

Thoughts?

friedbunny:
Here are iOS demo app headers captured using [Charles Proxy](http://www.charlesproxy.com). Looks like we're trying to use `keep-alive` on both ends.

**Request**

```http
GET /v4/mapbox.mapbox-terrain-v2,mapbox.mapbox-streets-v6/2/0/2.vector.pbf?access_token=tokenWasHere&events=true HTTP/1.1
Host: a.tiles.mapbox.com
If-None-Match: "33b78a6ffc95d2bf588374d3343d5fcb"
Proxy-Connection: keep-alive
Accept: */*
User-Agent: MapboxGL
Accept-Language: en-us
Accept-Encoding: gzip, deflate
Connection: keep-alive
```

**Response**

```http
HTTP/1.1 200 OK
Content-Type: application/x-protobuf
Content-Length: 13188
Connection: keep-alive
Access-Control-Allow-Methods: GET
Access-Control-Allow-Origin: *
Cache-Control: max-age=43200,s-maxage=300
Content-Encoding: gzip
Date: Thu, 06 Aug 2015 17:28:33 GMT
ETag: "33b78a6ffc95d2bf588374d3343d5fcb"
Last-Modified: Tue, 26 May 2015 21:18:51 GMT
Server: nginx
Timing-Allow-Origin: *
X-Powered-By: Express
X-Cache: Miss from cloudfront
Via: 1.1 45578d14a69df96accaab0d1aba82a5a.cloudfront.net (CloudFront)
X-Amz-Cf-Id: AaxSZk9YKyduZDmN7bag3oMioow79nVfb2qCFPnHccdckjaBZBUtNA==
```

kkaefer:
[...]

PR #2008, opened 2015-08-06
Ignore tile URL sharding

1ec5:
The source returns sharded tile URLs, but sharding doesn’t buy us anything on native platforms, so always grab the first.

ref mapbox/mapbox-gl-native#2007

/cc @jfirebaugh @tmcw


jfirebaugh:
:+1:

mb12:
Is there a way to measure the round trip time for each network call in gl-native client? It would be nice to see the improvement in round trip numbers due to this change.

friedbunny:
Using Charles Proxy I don't see a significant improvement here per tile, but initial DNS lookups are slow and minimized with this change.

PR #2024, opened 2015-08-10
Render solid background with glClear, rather than drawing quads

kkaefer:
This patch moves to setting the `glClear` color to the map's background color, if the background isn't textured. This allows us to skip drawing full screen quads for the background layer.


ljbade:
Ah yeah I wondered why we were doing this a long time ago.

1ec5:
:+1:

PR #2026, opened 2015-08-10
Don't constantly repaint

kkaefer:
Right now, we're constantly repainting, after you move the map once. This causes severe CPU usage and battery drain.


picciano:
This is awesome...I guess I can stop trying to optimize my UIScrollView issue now...
Seriously, nice catch.

PR #1292, opened 2015-04-17
Slip and slide

1ec5:
On the perceived performance front, we’ve made the map feel far too “heavy” when it comes to gestures. Our gesture recognition impacts the map’s responsiveness pretty heavily. So here’s what I’d like to do to make the map have the same slippery feel as `UIScrollView`:

 - [x] Drift-pan more aggressively after flicking (fixes #1266)
 - [x] Drift-zoom after pinching
 - [x] Drift-rotate after rotating

After a flick of the finger, `MGLMapView` now drifts much more naturally than before. It’s just _ever-so-slightly_ more aggressive than `MKMapView` but still a lot less aggressive than Google Maps.

/cc @mourner @incanus

PR #2027, opened 2015-08-10
Don't disable user tracking on `setDirection:` or quick zoom gesture

### LINKED ISSUE #1644: iOS setDirection:animated: resets the userTrackingMode
Calls to `setDirection:animated:` sets the userTrackingMode to MGLUserTrackingModeNone. There should be a way to set the direction while keeping the userTrackingMode set to MGLUserTrackingModeFollow.

friedbunny:
The user location tracking mode would be reset on `setDirection:`/`setDirection:animated:` — this is undesirable and has been changed to retain tracking.

The tracking mode would also be reset unnecessarily during the double-tap-hold-move-up-and-down quick-zoom gesture. This gesture now keeps tracking the user, though not while they're actively performing the gesture (it catches up when the zoom is over, even if the gesture is still being held).

Fixes #1644.

/cc @1ec5 @incanus

### DIFF
--- a/platform/ios/MGLMapView.mm
+++ b/platform/ios/MGLMapView.mm
@@ -1371,8 +1371,6 @@ std::chrono::steady_clock::duration secondsAsDuration(float duration)
         self.scale = _mbglMap->getScale();
 
         self.quickZoomStart = [quickZoom locationInView:quickZoom.view].y;
-
-        self.userTrackingMode = MGLUserTrackingModeNone;
     }
     else if (quickZoom.state == UIGestureRecognizerStateChanged)
     {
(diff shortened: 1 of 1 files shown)


1ec5:
> though not while they're actively performing the gesture (it catches up when the zoom is over, even if the gesture is still being held).

Yes, this is generally true about our gesture recognizers now. In #1292, for simplicity’s sake, I rewrote the recognizers to always manipulate the viewport based on the initial viewport.

friedbunny:
Build failures (3) are Android related.

friedbunny:
Rebasing and re-testing this, with an eye to merge today.

incanus:
Doesn't quick zooming zoom where the gesture occurred? In which case, the map center & user location would change?

friedbunny:
The quick zoom gesture [only modifies the scale](https://github.com/mapbox/mapbox-gl-native/blob/b406c942df3293f12415d24d03ef17eb44847c88/platform/ios/MGLMapView.mm#L1385) of the map and always zooms in/out of the current center of the viewport.

If the center changes during a quick zoom (e.g., the location updated during tracking) the map will try to move to that new center, but there's a possibility the ongoing gesture animation could clobber the center-change animation. Once the gesture pauses or finishes, the center is free to update.

This seems acceptable to me, if not ideal.

incanus:
Sounds pretty good.

Issue #2029, opened 2015-08-10
Log functions assume valid ThreadContext


ljbade:
Currently the log functions assume there is a valid ThreadContext at https://github.com/mapbox/mapbox-gl-native/blob/master/src/mbgl/platform/log.cpp#L52

There should be a check to see if the ThreadContext is valid first so the log functions can be safely used from any thread.

Found while working on https://github.com/mapbox/mapbox-gl-native/issues/1662

/cc @kkaefer

PR #2030, opened 2015-08-10
Fix ThreadContext to check for valid current pointer

ljbade:
`ThreadContext` now checks if `current` is a valid pointer.

Fixes https://github.com/mapbox/mapbox-gl-native/issues/2029

@kkaefer Can you review this.

### DIFF
--- a/src/mbgl/util/thread_context.hpp
+++ b/src/mbgl/util/thread_context.hpp
@@ -36,27 +36,51 @@ public:
     }
 
     static std::string getName() {
-        return current.get()->name;
+        if (current.get() != nullptr) {
+            return current.get()->name;
+        } else {
+            return "Unknown";
+        }
     }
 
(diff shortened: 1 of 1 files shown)


tmcw:
@kkaefer if you're leaving for vacation soon, could you review this so @ljbade can keep moving?

mikemorris:
Could [`HeadlessView::isActive`](https://github.com/mapbox/mapbox-gl-native/blob/b312820be89b147a91a7b596d6704ae69c0ddad2/platform/default/headless_view.cpp#L113-L115) be switched to using these (smarter?) checks instead?

ljbade:
Hit this crash again while trying to add logging to trace https://github.com/mapbox/mapbox-gl-native/issues/2058

ljbade:
Assuming this PR is OK.

PR #2033, opened 2015-08-11
Use OkHTTP to implement HTTPContext on Android

### LINKED ISSUE #823: Implement FileSource in Java
I think we should look at what it takes to implement the FileSource object in Java. This would allow us to remove the dependencies on CURL, OpenSSL and libzip, as well as SQLite.

With the FileSource refactor, we can now more easily swap out the storage backends: Either, we replace the FileSource object completely, which leaves us with implementing the FileSource API, which is essentially just a `request/cancel` method. All the underlying logic, like caching, HTTP requesting and loading from the APK file has to be implemented manually.

We could also replace only part of the API, e.g. only reimplement the HTTP logic in Java. I think we should start with the latter, but potentially move our way up the chain until we have we can get rid of all of these dependencies, especially given that they have Java APIs that use the system's built-in libs.

/cc @ljbade

ljbade:
Closes #823


ljbade:
This is a pretty big merge with lots of new JNI code.

@kkaefer @jfirebaugh Do you mind giving this a review?

jfirebaugh:
I'm not really familiar with any of this code. @incanus or @bleege are probably better reviewers.

mb12:
1.) Can you also share improvements in APK file size and any changes in performance w.r.t runtime and memory?
2.) Also do tile requests for Android do  a lookup in sqlite cache before going over the network?  Will this also be moved to managed code?

ljbade:
@mb12 In this branch everything still goes via sqlite cache. Can't see why we can't use Java side caching.

Does OkHTTP have support for that?

I haven't done checks of the APK size or performance comparisons yet.

ljbade:
Going to collect here before/after `.aar` file sizes:

For just ARM 32bit lib:
- Before: 3.0MB
- After: 2.0MB 

All libs:
- Before:
- After:

ljbade:
@kkaefer Do you know a good way to quickly benchmark the tile fetch performance with these changes?

Issue #2036, opened 2015-08-11
All animations broken


friedbunny:
Tapping the compass no longer reliably resets the bearing — upon tapping, the compass will sometimes stutter towards north a few degrees and halt. Other times the compass will haltingly reset to north and disappear.

Not seeing any particular pattern right now.

/cc @mapbox/mobile

ljbade:
Happens on Android too, noticed in https://github.com/mapbox/mapbox-gl-native/issues/1856#issuecomment-129666707

1ec5:
#1829 changed the iOS compass to rely on mbgl for frame updates, whereas it used to use view-based animation. So this is definitely a bug in mbgl’s update notifications.

1ec5:
This is causing `-[MapViewTests testCompassTap]` to fail reliably.

1ec5:
Regression from fa4636ffe4d2c00446292c229cd549cbaee9e20c.

incanus:
iOS tests should have caught this, yes? Why didn't they? 

https://travis-ci.org/mapbox/mapbox-gl-native/builds/74223611

1ec5:
The tests have been catching the issue about 80% of the time. I had thought it was #1887, but those failures were just burying this one.

1ec5:
Probably related: when tracking user location, the user dot often shuffles out of view.

1ec5:
All animation is broken, and I’m having a difficult time untangling fa4636ffe4d2c00446292c229cd549cbaee9e20c. @brunoabinader, can you look into this?

1ec5:
I may have confused the bisect by starting with release-0.5.0 as good, but when I start with two revisions on master, I get this commit instead: 0a527f623ed88ef46c104e97a86239b012141074. #2026 is interesting because perhaps this constant redrawing was only masking the underlying problem.

/cc @kkaefer

brunoabinader:
I am AFK this Thursday and Friday, will have a look on my way to US this weekend.
/cc @tmpsantos

brunoabinader:
Got lucky with wifi onboard, so I managed to find the root cause: `mbgl::Transform::{_moveBy,_setAngle}` were returning `Update::Nothing`, causing the render call to not repaint. Changing thost to `Update::Repaint` fixed all the aforementioned issues. Fixed in #2060.

brunoabinader:
[...]

PR #2040, opened 2015-08-11
fixes #1951: check for support of title before querying

### LINKED ISSUE #1951: Crash when selecting annotation without title
1. Implement a class conforming to MGLAnnotation but without -title, -subtitle
2. Setup map and setup concrete instance of your new class.
3. Provide image for this annotation via MGLMapViewDelegate
4. Click onto the annotation image in the map.

Actual result:
It crashes because -title is called without check if object responds to this selector
Expected result:
There should be runtime checks for all @optional methods/properties from protocols the object conforms to

incanus:


incanus:
Tests pass locally save the two compass/directions-related iOS UI ones per https://github.com/mapbox/mapbox-gl-native/issues/2036.

Issue #2042, opened 2015-08-12
getMapAsync for Android like in Google maps


erf:
Would it be an idea to load the mapview async as in google maps:
<pre>
   mapFragment = new SupportMapFragment();
        mapFragment.getMapAsync(this);

   FragmentManager fragmentManager = getSupportFragmentManager();
   FragmentTransaction fragmentTransaction = fragmentManager.beginTransaction();
   fragmentTransaction.replace(R.id.container, mapFragment);
   fragmentTransaction.commit();
</pre>
In this way you could make the mapFragment static so you don't have to reinstanciate it each time you create the activity containing the map.

I have made an implementation of this.

erf:
I created a pull request for this:
https://github.com/mapbox/mapbox-gl-native/pull/2043

erf:
The problem i described in #2018 is back after i load the mapview this way

ljbade:
I don't really understand what this does.

It seems you are only changing the way the test app works?

/cc @bleege

erf:
You probably would like to have MapView in a Fragment to reuse map functionality (like google do) and to add it programmatically you need to retrieve the MapView async from the MapFragment when it's added to the view or else it's null if you try to retrieve it just after you've created the MapFragment. 

But.. This causes another problem, i describe in #2018 that when i tried to switch from the MapActivity to some other Activity (after moved the maped) and back the app crashes..

tobrun:
@ljbade 

I would like to elaborate why Google exposes this async get method:

![image](https://cloud.githubusercontent.com/assets/2151639/10336371/9fdad0e0-6cf8-11e5-985b-7016b99fb8a1.png)

This would not work because the MapFragment is not created yet, for this reason that method exists.

cc @bleege

tobrun:
Also here is an example how getMap can work, I used to implement GoogleMaps in this fashion before the async method was introduced and getMap was deprecated:

![image](https://cloud.githubusercontent.com/assets/2151639/10336459/3da85acc-6cf9-11e5-9bcc-441ec2f143fc.png)

ljbade:
@tobrun Interesting! Perhaps we should put this on 2.2.0 milestone as part of the refactor.

tobrun:
This will be implemented with #3145

tobrun:
This has landed with #3145. Closing

Issue #2018, opened 2015-08-09
Freezes when going back and forth between activites


erf:
If you add a button to the main view in the test app, and open an arbitrary activity on a click event and then closes it, moves the map and click the button again, the app freezes on my nexus 7 (2012).
I made a simple test case here: 
https://github.com/erf/mapbox-gl-native/tree/freezes

erf:
BTW I did not get any log error events.. but this is a MAJOR issue, if creating an app that uses more than one Activity makes the app freeze up, than the library is pretty much useless. Does anyone else experience this?

erf:
This also seem to happen if i moved the map first and then click on a button which opens the new activity. I don't see any information in the Log regarding this. I suspected this may be a memory issue when opening the map several times, so i even tried to make the mapview static, but with same result. I suspect now that some state is getting set when you move the map, but i have no idea what causes the crash.

erf:
I just made a simple example now on latest 4025da6 and this seem to work now! I close this for now.

erf:
This issue is back after i added loading mapView as a fragment described in #2042

tobrunvannuland:
Was able to reproduce this occasionally with following setup:
Genymotion - Google Nexus 5.0.0 image - 0.0.1-SNAPSHOT

Actions:
Open activity with MapView, scroll-zoom actions like a maniac, open other Activity containing a MapView. Application becomes unresponsive and screens turns black. Only option is to kill-app by swiping it from recents.

Used Activities:
StyledMapActivity -  SearchMapActivity

[Full stacktrace](http://pastebin.com/vBK8Cx5S) - [code](https://github.com/tobrunvannuland/mapbox-gl-native/tree/playground/android/java/MapboxGLAndroidPlaygroundApp)

erf:
I think the app crashes when trying to open a new activity whilst the MapView is still loading tiles

ljbade:
I also wonder if it is related to state saving.

ljbade:
@kkaefer Is there anything in the C++ layer to be aware of when there are multiple instances of a map?

kkaefer:
@ljbade it should be possible and I'm not aware of an issues, but of course there could be bugs when using multiple instances.

incanus:
We are handling multiple maps in iOS ok right now.

ljbade:
Hmm going to have to create a simple test app to see if I can reproduce it.

[...]

PR #2043, opened 2015-08-12
now get the map async when you add it as a MapFragment

erf:
https://github.com/mapbox/mapbox-gl-native/issues/2042


erf:
I would wait to merge this pull request until it works with issue #2018

jfirebaugh:
#2042 was fixed by #3145.

PR #2050, opened 2015-08-12
fixes #2044: add -showAnnotations:animated:

### LINKED ISSUE #2044: Add -[MGLMapView showAnnotations:animated:]
MapKit has a method since iOS 7 which sets the visible region so that the map displays all specified annotations. It would be nice if `MGLMapView` has a method like this:
```
- (void)showAnnotations:(NSArray<id<MGLAnnotation>> * _Nonnull)annotations animated:(BOOL)animated
```

incanus:


incanus:
Want to give it a spin @jncn?

incanus:
We decided there may be some more subtlety in MapKit's implementation than a flat `100px` all around (something to do with map aspect ratio and/or the poles), but this is a good first cut.

incanus:
Also we may want to do something with view controller layout guides using `-[UILayoutSupport length]` and the current meters/pixel, but again, we can iterate later.

jncn:
Great! It does the job, thanks. Might indeed need some refinements. MapKit's counterpart isn't perfect either, btw.

PR #2060, opened 2015-08-13
Fix update flags in mbgl::Transform::{_moveBy,_setAngle}

### LINKED ISSUE #2036: All animations broken

brunoabinader:
This is needed to properly tell `Map::nudgeTransitions()` to trigger a repaint.

Fixes #2036.

/cc @friedbunny @ljbade @1ec5 @incanus @kkaefer @tmpsantos

### DIFF
--- a/include/mbgl/map/update.hpp
+++ b/include/mbgl/map/update.hpp
@@ -5,9 +5,7 @@
 
 namespace mbgl {
 
-using UpdateType = uint32_t;
-
-enum class Update : UpdateType {
+enum class Update : uint32_t {
     Nothing                   = 0,
     Dimensions                = 1 << 1,
     DefaultTransition         = 1 << 2,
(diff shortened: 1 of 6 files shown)


brunoabinader:
I debugged this while at the airport and found the issue: `transitionFrameFn` needs to call `MapData::setNeedsRepaint(true)` so `Map::nudgeTransitions` should better know what to do. This can be avoided if `transitionFrameFn` actually returns a valid `Update::UpdateType` instead of a frame value.

1ec5:
Thanks @brunoabinader!

Issue #1942, opened 2015-07-28
Add Android event listeners


ljbade:
We should add support for listening to the various Map events such as map position changes etc.

I see Google Maps has a variety of different listeners listed [here](https://developers.google.com/android/reference/com/google/android/gms/maps/GoogleMap)

@bleege What sort of listeners does the raster SDK have?

What should we implement?

incanus:
@ljbade Are https://www.mapbox.com/mapbox-android-sdk/api/0.7.4/com/mapbox/mapboxsdk/events/MapListener.html and https://www.mapbox.com/mapbox-android-sdk/api/0.7.4/com/mapbox/mapboxsdk/views/MapViewGestureDetectorListener.html helpful?

ljbade:
@incanus Yeah. I think there are also some other listeners we could add from `notifyMapChange`

Depends what is useful to apps vs what is just usefully internally.

ljbade:
List of touch events from @erf:
- click
- long click
- drag

erf:
I belive adding the following to MapView will solve the first two:
<pre>
        @Override
        public boolean onSingleTapUp(MotionEvent e) {
            // Cancel any animation
            mNativeMapView.cancelTransitions();

            performClick();

            return true;
        }

        // Called for single taps after a delay
        @Override
        public boolean onSingleTapConfirmed(MotionEvent e) {
            return false;
        }

        // Called for a long press
        @Override
        public void onLongPress(MotionEvent e) {
            // TODO
            performLongClick();
        }
</pre>

erf:
Now onMapChangedListener, get's a lot of events. Google maps only receives these events after map has stopped moving (after a drag).

erf:
Instead of drag - we could perhaps add an "OnFlingListener", since it's already an gesture for that?

erf:
Just add somthing like
<pre>
    public interface OnFlingListener {
        void onFling();
    }

    public interface OnScrollListener {
        void onScroll();
    }

    private OnFlingListener onFlingListener;
    private OnScrollListener onScrollListener;

    public void setOnScrollListener(OnScrollListener onScrollListener) {
        this.onScrollListener = onScrollListener;
    }

    public void setOnFlingListener(OnFlingListener onFlingListener) {
        this.onFlingListener = onFlingListener;
    }
</pre>
And then call those inside
<pre>
        public boolean onScroll(MotionEvent e1, MotionEvent e2, float distanceX, float distanceY) {
[...]

PR #2063, opened 2015-08-13
added click, longclick, scroll, fling events

erf:
as discussed in #1942

the event listeners should be converted to lists and the onfling and onscroll listeneers should probably have some parameters


tmcw:
@bleege this looks solid to me - I'm not sure about the Java-ism but is it better to check whether a listener is null and then call it, or set it initially to a no-op function and always call it?

bleege:
@erf Thanks for the PR.  I've been sidetracked the last few days, but hope to get to this soon!

erf:
No problem. Btw. This worked for me, but i don't have complete overview of what is needed here..

bleege:
Looking into this now.

bleege:
This looks good.  Thanks @erf!

altafrm3d:
Is there something like onZoom() Listener other than the one for buttons, i have not been able to find it anywhere.

ljbade:
@altafrm3d Not yet. Open a feature request ticket.

Issue #1676, opened 2015-06-02
While panning, location indicator wiggles significantly more on android compared to iOS (Android).


mb12:
Filing this issue for tracking purposes.
While panning, location indicator wiggles a lot more on Android compared to iOS.  Maybe whatever changes were done for the following needs to be migrated for Android as well.

https://github.com/mapbox/mapbox-gl-native/issues/1439

1ec5:
#1439 was duped to #1125, which is still open. #1582 will get us to parity with MapKit by using Core Animation. Does Android have anything similar?

ljbade:
@1ec5 there is currently no filtering of the compass data. It is updating the screen as frequently as we get sensor data. We need to implement some sort of rate limit or other data filtering.

1ec5:
Would it be necessary to throttle the user location updates if the user dot animated to each location instead of blitting there?

ljbade:
The compass and GPS locations are separate data sources. Android already limits location to once a second so we need to do something similar with the compass.

ljbade:
This is still a problem even after @bleege recent changes.

@incanus How did you speed up the map/location marker sync?

erf:
@ljbade i created a pull request now  #2065 regarding compass filtering. but after i had added the filtering like in a previous pull request, i saw that the FAB does not toggle into Follow mode. I guess @bleege is in the middle of some refactoring here, but would be nice if the filtering could be included if you test that it is ok. Thanks.

incanus:
We need to insert the routine which updates the user dot's location on screen between these two lines: 

https://github.com/mapbox/mapbox-gl-native/blob/fa18f1afd0f2c736e09ffc8989548fdc0bafdf3c/android/cpp/native_map_view.cpp#L710-L711

Per https://github.com/mapbox/mapbox-gl-native/issues/1125, `renderSync` draws the current GL frame, then `nudgeTransitions` updates the `Transform`, which affects any subsequent geo/screen conversions. If the user dot placement happens **not** between those two, it is using a `Transform` for the placement conversion that is not in sync with the GL draw.

ljbade:
@incanus Aha! Thanks for that pointer! I'm going to figure this one out

bleege:
Bumping to `0.2.0`.

@ljbade Any progress on this?

ljbade:
@bleege not yet, will pick this up after `0.1.0`

incanus:
[...]

PR #2065, opened 2015-08-13
filter compass

erf:
added compass filtering as discussed in #1676


tmcw:
This looks solid to me, can you take a look @ljbade ?

tobrun:
@erf 
I wanted to thank you for this contribution, I have been playing around with your suggestion. Your code has helped me a lot on setting this up but I must mention I have removed it again. Instead I'm rate limiting the sensor input and I'm seeing great improvements on lowering memory churn.

PR #2066, opened 2015-08-13
Restore mason version to d8597d9445

tmpsantos:
It was downgraded on 34d1c023de.


jfirebaugh:
We should add a check in CI: fail the build if a submodule is updated unless "[submodule]" is in the commit message.

bleege:
> We should add a check in CI: fail the build if a submodule is updated unless "[submodule]" is in the commit message.

:+1: It's way to easy to stomp on submodules.

PR #2067, opened 2015-08-13
Restore geojsonvt version to 995ffc72c5

tmpsantos:
It was downgraded on 34d1c023de.


bleege:
Thanks @tmpsantos!  I accidentally stomped on it yesterday.

PR #2080, opened 2015-08-14
Remove mutexes from LineAtlas

kkaefer:
`LineAtlas` has a bunch of recursive mutexes and show up during profilling, but from a cursory look, it seems like it's only ever used from the Map thread. Therefore, we can reduce those mutexes altogether.

### DIFF
--- a/src/mbgl/geometry/line_atlas.cpp
+++ b/src/mbgl/geometry/line_atlas.cpp
@@ -20,7 +20,7 @@ LineAtlas::LineAtlas(uint16_t w, uint16_t h)
 }
 
 LineAtlas::~LineAtlas() {
-    std::lock_guard<std::recursive_mutex> lock(mtx);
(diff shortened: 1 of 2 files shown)


mb12:
LineSdfShade/LineAtlas  is only used to draw dashed lines. Why don't we draw dashed lines by using the  a_linesofar vertex attribute to make gaps fully transparent in the fragment shader?  This should get rid of the Line Atlas  altogether.

kkaefer:
@mb12 please open a new ticket for issues not related to this pull request.

ansis:
> Why don't we draw dashed lines by using the a_linesofar vertex attribute to make gaps fully transparent in the fragment shader? This should get rid of the Line Atlas altogether.

We used to do this. We switched so that we could support:
- dash patterns of any length (for example `[10, 2, 5, 2]`)
- round caps

These could have been implemented a different way, but using textures seemed like a good idea since we were already doing that for patterned lines.

PR #2038, opened 2015-08-14
Intelligently enable/disable vertex attribute indices when VAOs are unavailable

kkaefer:


kkaefer:
This seems to have adverse effects on frame render time.

jfirebaugh:
> This seems to have adverse effects on frame render time.

If this is a deoptimization should we close this PR?

kkaefer:
Yes.

PR #2092, opened 2015-08-17
Move validateAccessToken to onCreate

ljbade:
Ensures Java exception thrown before hard to understand native crash in `util::mapbox::normalizeSourceURL`


bleege:
Will need to redo this as the validateAccessToken() can not be in `onCreate()` in order to support Access Tokens in `layout.xml`.  See https://github.com/mapbox/mapbox-gl-native/issues/1929

Issue #1140, opened 2015-03-27
port perspective view to -native


ansis:
https://github.com/mapbox/mapbox-gl-js/pull/1049
and
https://github.com/mapbox/mapbox-gl-js/pull/1114

PR #2116, opened 2015-08-18
perspective view

ansis:
Ports perspective view from -js.
fixes https://github.com/mapbox/mapbox-gl-native/issues/1140

screen shot 2015-08-18 at 8 59 24 am
screen shot 2015-08-18 at 8 59 54 am
screen shot 2015-08-18 at 9 02 28 am

TODO:
- [x] test on device. Can someone help with this?
- [x] fix render tests that fail on travis but not locally


tmpsantos:
@ansis awesome!

mb12:
@ansis I was able to build and run these changes on iPhone 5 as well as Samsung Note 3.  Here are two screenshots with pitch hardcoded to 45 degrees. It felt  a little slower on Note 3 (but this could very well be a general Android issue. I think  this repo doesn't have changes to call HTTPAndroidRequest::cancel on right thread for Android. This is causing crash on Android when one zooms in.). 

Also is it possible to share more details at a logical level regarding this change with everyone?
 
img_2041

screenshot_2015-08-18-10-46-22

friedbunny:
Basic iOS gesture support is over in [`2116-perspective-iOS`](https://github.com/mapbox/mapbox-gl-native/compare/2116-perspective-iOS).

/cc @incanus @1ec5

ansis:
@mb12 thanks for testing!

The implementation is pretty much the same as the one in -js. https://github.com/mapbox/mapbox-gl-js/pull/1049 and https://github.com/mapbox/mapbox-gl-js/pull/1114


The orthographic projection matrix was replaced with a perspective matrix. Two new transform properties were added to control the projection. *pitch* controls how tilted the map is and *altitude* controls how quickly lines approach the vanishing point. Only *pitch* is exposed externally.

`TransformState#pointCoordinate` was added to correctly convert point coordinates to tile coordinates. This is used to figure out which tiles are needed to cover the current view. All tiles are loaded from the same zoom level. There is no level of detail support.

A basic api for getting and setting the pitch was added.

Many of the changes are related to adapting antialiasing. Lines and text need different levels of blur depending on how far away they are. Both of these now have approximate solutions that work well enough but aren't 100% mathematically correct. This is probably the hardest and weirdest part of the perspective implementation.

Layers now have slightly different depthRanges. Both ends of the depthRange are shifted for each layer so that depth values between layers are different enough at both ends of the depth range. We could have made each layer have a single depth value instead of a range, but we'll need ranges for 3d buildings later.

Line labels are projected onto the surface because it was easier to render them that way.

1ec5:
Should mbgl be responsible for clamping the pitch to a reasonable value, or should that be left up to the platform bindings? Setting the pitch above π rad currently triggers this assertion failure:

```
Assertion failed: ((detail::is_valid(m_members.translator()(value)))&&("Indexable is invalid")), function raw_insert, file /path/to/mapbox-gl-native/mason_packages/headers/boost/1.57.0/include/boost/geometry/index/rtree.hpp, line 1242.
```

while setting it below 0 rad causes the map to go blank. But should we clamp to just the values that’ll cause this kind of behavior, or should we be stricter and disallow showing the horizon?

incanus:
> Should mbgl be responsible for clamping the pitch to a reasonable value, or should that be left up to the platform bindings? 

We should probably leave this to the platforms, just like right now how we don't allow rotation at low zoom levels on iOS and animate it back into place.

friedbunny:
> just like right now how we don't allow rotation at low zoom levels on iOS and animate it back into place

To that end, we'll have to do something about tilting into the edge of the world:

simulator screen shot aug 19 2015 6 32 29 pm

MapKit solves this by progressively allowing less pitch as you zoom out.

incanus:
pirates-of-the-caribbean-at-world039s-end-poster-artwork-johnny-depp-orlando-bloom-keira-knightley

incanus:
Blocker to merge: https://github.com/mapbox/mapbox-gl-native/issues/2139

Otherwise things like the user dot don't trace the right geo point on the map, and e.g. long-press to drop a pin will put things in the wrong place. Plus, these are needed in general for mobile-based app coordinate transformation.

ansis:
This pull request is already getting pretty big. Should we review and merge this into master?

There are things that still need to be improved but it would probably be easier to review and merge those separately. This branch shouldn't break or regress anything in non-perspective views.

incanus:
> This branch shouldn't break or regress anything in non-perspective views.

It does in https://github.com/mapbox/mapbox-gl-native/issues/2139.

1ec5:
Another regression: #2174.

1ec5:
I'm seeing some artifacting on some streets when tilted:

image
image

Issue #2012, opened 2015-08-07
{Map}[Style]: Failed to load [mapbox://mapbox.mapbox-streets-v6]: 


desduvauchelle:
Just installed MapboxGL through cocoapods. I initialize it in my ViewController like the example given:
```
 @IBOutlet weak var mapViewWrapper: UIView!

    var mapView: MGLMapView!

    override func viewDidLoad() {
        super.viewDidLoad()

        // Map
        let styleURL = NSURL(string: "asset://styles/dark-v7.json")
        mapView = MGLMapView(frame: mapViewWrapper.bounds, styleURL: styleURL)
        mapView.autoresizingMask = .FlexibleWidth | .FlexibleHeight

        // set the map's center coordinate
        mapView.setCenterCoordinate(CLLocationCoordinate2D(latitude: 38.894368,
            longitude: -77.036487),
            zoomLevel: 15, animated: false)
        mapViewWrapper.addSubview(mapView)
    }
```

When I launch the app, I see a black square and a Mapbox logo but no streets, see image below enter image description here: 
![Image of map](http://i.stack.imgur.com/GhjvX.png)

In the logs I'm getting this:
```
[ERROR] {Map}[Style]: Failed to load [mapbox://mapbox.mapbox-streets-v6]: 
[ERROR] {Map}[Style]: Failed to load [mapbox://mapbox.mapbox-terrain-v2]: 
```
Any ideas what I did wrong?

picciano:
I have several users (including myself) reporting the same thing in our app. Something has gone terribly wrong. We are building from source from master.

picciano:
This issue has been plaguing us all weekend.
The map often goes black when switching back and forth between apps as well as after a clean install. I have recreated this problem with the demo app. See attached, the map goes black at the end (last 10 seconds), but the compass still turns when performing the rotate gesture. There are no error messages or warnings in the console.

Screen capture from iPhone 5s: http://www.filedropper.com/appgoesblack

picciano:
Rolling back this commit seems to get rid of the "black screen of death", but does leave a flicker when coming back from background. Seem to be related to `view.invalidate()` not always getting called.

https://github.com/mapbox/mapbox-gl-native/commit/9fd7e883a7515f97ec4edd097a0db5cbebce3ca1

friedbunny:
@desduvauchelle Are you using `MapboxGL` in the podspec or are you pointing at our git prereleases?

[...]

Issue #1702, opened 2015-06-10
Area fill colors get corrupted upon scroll (Android).


mb12:
This is consistently reproducible on Samsung Nexus S. 
Just go to any area with fill colors and scroll.  I've attached two screenshots from Samsung Nexus S (one that shows expected fill colors and the other that shows corrupted fill colors).

![corrupted_fillcolor](https://cloud.githubusercontent.com/assets/8890269/8074764/258a51fa-0eeb-11e5-8f5e-11a0fbd78ec1.png)
![expected_fillcolor](https://cloud.githubusercontent.com/assets/8890269/8074763/258a1d84-0eeb-11e5-8fb2-cadf647a84f7.png)

mb12:
Here is another screenshot illustrating corruption.

![another_corruption](https://cloud.githubusercontent.com/assets/8890269/8074834/f40c3426-0eeb-11e5-8756-220ae285daac.png)

ljbade:
Hmm this is interesting.

Have you seen this before on iOS (or Android) @kkaefer @1ec5 ? I haven't

ljbade:
@mb12 Does this only happen on Nexus S? Have you been able to reproduce this on other phones?

1ec5:
I haven't seen this issue on iOS. The closest thing I can think of is that fills are all screwed up after loading Pencil style.

kkaefer:
Never seen this issue on my Moto G 2 test device.

mb12:
@ljbade Consistently reproducible on Nexus S. Does not happen on Note 3.
@1ec5  What is the pencil style issue you are referring to? Is it consistently reproducible?

mb12:
The "Light" style shows red lines and smudges even without scrolling.

![light_style](https://cloud.githubusercontent.com/assets/8890269/8088197/71d87f5a-0f54-11e5-861e-366eaa7aa61c.png)

1ec5:
> What is the pencil style issue you are referring to? Is it consistently reproducible?

See #636 and https://github.com/mapbox/mapbox-gl-styles/blob/mb-pages/styles/pencil-v7.json.

mb12:
Sometimes lines disappear on scrolling as well. I've attached two screenshots (one with lines and one without lines).

![lines_as_expected](https://cloud.githubusercontent.com/assets/8890269/8088741/8171df84-0f58-11e5-938f-a6864ff97ef8.png)
[...]

PR #2071, opened 2015-08-18
Switching applications causes black screen, wildly colored snapshot

brunoabinader:
Activating the application switcher when the user dot is visible causes the entire map view (other than ornaments and the user dot) to go black. The following screenshots are from the iOS 9 Simulator running as an iPhone 6:

black-everything

While the application is in the background, the snapshot we throw up on-screen has an artistic flair to it that suggests texture or [z-fighting](https://github.com/mapbox/mapbox-gl-native/issues/1702#issuecomment-127375809) issues. As soon as the entering-foreground animation completes, we take down the snapshot and the black screen returns.

switching

switched

If you enter the application switcher while the user dot is hidden, the snapshot is correct, although the black screen does come back as soon as you switch back to the application:

switching-good

switched-good

Notice how some features do appear in the corrupted snapshot, including a school grounds, part of a highway, and some terrain tinting.

The same thing happens on the iOS 8.4 Simulator running as an iPhone 6:

switching-8 4

If you merely activate the application switcher but don’t switch away from iosapp, you’re left with a live, shrunken view of the application. In this state, you can sometimes see the colored splotches shifting around as the map redraws.

Unlike #2012, nothing is logged to the console.

/cc @kkaefer @incanus @friedbunny

### DIFF
--- a/src/mbgl/map/map_context.cpp
+++ b/src/mbgl/map/map_context.cpp
@@ -74,6 +74,8 @@ void MapContext::cleanup() {
 void MapContext::pause() {
     MBGL_CHECK_ERROR(glFinish());
 
+    viewInvalidated = false;
+
     view.deactivate();
 
     std::unique_lock<std::mutex> lockPause(data.mutexPause);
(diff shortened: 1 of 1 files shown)


brunoabinader:
@1ec5 could you please verify the patch I just merged? It fixed the issue on my iPad Mini as well as iOS simulator (iPhone 6).

robipresotto:
@brunoabinader it looks great

1ec5:
Looks good, thanks @brunoabinader!

Issue #1870, opened 2015-07-10
MapboxGL on XCode Version 7.0 beta


danielgomezrico:
I´m one of those horrible compile error with the simple example on https://www.mapbox.com/mapbox-gl-ios/examples/. :(

My pod file:
```
source 'https://github.com/CocoaPods/Specs.git'
use_frameworks!
pod 'MapboxGL'
```

My view controller:

```
import UIKit
import MapboxGL

class MapViewController : UIViewController, MGLMapViewDelegate {

    override func viewDidLoad() {
        super.viewDidLoad()

        let mapView = MGLMapView(frame: view.bounds)
        mapView.autoresizingMask = [UIViewAutoresizing.FlexibleWidth, UIViewAutoresizing.FlexibleHeight]

        // set the map's center coordinate
        mapView.setCenterCoordinate(CLLocationCoordinate2D(latitude: 40.7326808,
            longitude: -73.9843407),
            zoomLevel: 12, animated: false)

        view.addSubview(mapView)
    }

}
```

Error:
```
Undefined symbols for architecture x86_64:
  "_OBJC_CLASS_$_MGLMapView", referenced from:
      type metadata accessor for ObjectiveC.MGLMapView in MapViewController.o
ld: symbol(s) not found for architecture x86_64
clang: error: linker command failed with exit code 1 (use -v to see invocation)
```
Is it a bug on Xcode or is mine?

1ec5:
Hi @danielgomezrico, this is a duplicate of #1849, which is caused by the new version of CocoaPods. Please add `-ObjC` to the `OTHER_LDFLAGS` (Other Linker Flags) build setting of your application target.

danielgomezrico:
But I have it...

![screen shot 2015-07-09 at 9 27 31 pm](https://cloud.githubusercontent.com/assets/76348/8611091/5b02894e-2681-11e5-96f6-c98a37172180.png)

friedbunny:
In that case, @danielgomezrico, you should try the voodoo dance of ⌘⇧K (CMD+SHIFT+K) and ⌘⌥⇧K (CMD+OPTION+SHIFT+K) to clean everything out.

1ec5:
You may also need to mark Pods.framework as optional instead of required in the General tab.

1ec5:
[...]

PR #2132, opened 2015-08-19
Replace deprecated podspec flag

1ec5:
Per https://github.com/mapbox/mapbox-gl-native/issues/1870#issuecomment-132417161.

ref CocoaPods/CocoaPods#3573


friedbunny:
It's been a couple months, time to pull the trigger here?

PR #2138, opened 2015-08-20
Speed up time to first render call

kkaefer:
I measured the time from the main thread calling `renderSync()` to us actually issuing our first render call (`glClear`). On an iPad 2, this is ~1.7ms, which is >10% of the total frame time. Most of this time is spent in `Painter::determineRenderOrder`, in particular the lookup of buckets by name in a tile object is very slow (and takes ~80% of the time).

Instead of building a list of all buckets we have to render on every frame, this list should be updated whenever we add/remove buckets/tiles, so we don't first have to build a list of things on every frame. Most frames don't actually see any changes to this list, in particular when panning.

### DIFF
--- a/src/mbgl/map/map_context.cpp
+++ b/src/mbgl/map/map_context.cpp
@@ -334,6 +334,7 @@ bool MapContext::renderSync(const TransformState& state, const FrameData& frame)
     if (!painter) {
         painter = std::make_unique<Painter>(data);
         painter->setup();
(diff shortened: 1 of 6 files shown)


kkaefer:
Current patch goes down the lazy route and updates the render order array whenever there are changes to tiles. We *could* change it so that we only add/remove the list with the new/abandoned buckets, but that'd also require sorting it on every update, so it's probably not significantly slower to just blow away the whole list.

PR #2128, opened 2015-08-20
Move FrameHistory::getFadeProperties to animation time

kkaefer:
`FrameHistory::getFadeProperties` currently calls `chrono::steady_clock::now()` on every invocation. Instead, we should use the global animation time to prevent issues with different parts of the app using different clocks. Additionally, not querying the time will improve the speed.


brunoabinader:
This looks good @kkaefer! :+1:

PR #2131, opened 2015-08-19
Increase timeouts for failing metrics tests

1ec5:
Appears to fix #1887. I tried rerunning the iOS build several times with no failures. (The only failure, ironically, is from rendering tests in Android builds.)

/cc @kkaefer @incanus

PR #2141, opened 2015-08-20
Support v8 sprite, glyph, and style mapbox:// urls

### LINKED ISSUE #2134: Support v8 mapbox:// urls
- [x] `mapbox://styles/[user]/[styleid]`
 - [x] `mapbox://sprites/[user]/[styleid]`
 - [x] `mapbox://fonts/[user]/{fontstack}/{range}.pbf`

cc @jfirebaugh 

 > native needs to be updated to support the final mapbox:// style and sprite URL formats, and maybe doesn't handle mapbox:// font URLs correctly either judging by what ansis is reporting

lucaswoj:
cc @jfirebaugh 

resolves #2134

### DIFF
--- a/src/mbgl/map/sprite.cpp
+++ b/src/mbgl/map/sprite.cpp
@@ -9,6 +9,7 @@
 #include <mbgl/util/raster.hpp>
 #include <mbgl/util/thread.hpp>
 #include <mbgl/util/uv_detail.hpp>
+#include <mbgl/util/mapbox.hpp>
 
 #include <rapidjson/document.h>
 
@@ -43,8 +44,8 @@ Sprite::Sprite(const std::string& baseUrl, float pixelRatio_)
         return;
     }
 
-    std::string spriteURL(baseUrl + (pixelRatio_ > 1 ? "@2x" : "") + ".png");
-    std::string jsonURL(baseUrl + (pixelRatio_ > 1 ? "@2x" : "") + ".json");
+    std::string spriteURL = mbgl::util::mapbox::normalizeSpriteURL(baseUrl, (pixelRatio_ > 1 ? "@2x" : ""), ".png", "NOCOMMIT");
+    std::string jsonURL = mbgl::util::mapbox::normalizeSpriteURL(baseUrl, (pixelRatio_ > 1 ? "@2x" : ""), ".json", "NOCOMMIT");
 
     loader = std::make_unique<Loader>();
 
--- a/src/mbgl/map/sprite.cpp
+++ b/src/mbgl/map/sprite.cpp
@@ -44,13 +44,13 @@ Sprite::Sprite(const std::string& baseUrl, float pixelRatio_)
         return;
     }
 
-    std::string spriteURL = mbgl::util::mapbox::normalizeSpriteURL(baseUrl, (pixelRatio_ > 1 ? "@2x" : ""), ".png", "NOCOMMIT");
-    std::string jsonURL = mbgl::util::mapbox::normalizeSpriteURL(baseUrl, (pixelRatio_ > 1 ? "@2x" : ""), ".json", "NOCOMMIT");
+    std::string spriteURL(baseUrl + (pixelRatio_ > 1 ? "@2x" : "") + ".png");
+    std::string jsonURL(baseUrl + (pixelRatio_ > 1 ? "@2x" : "") + ".json");
 
     loader = std::make_unique<Loader>();
 
     FileSource* fs = util::ThreadContext::getFileSource();
-    loader->jsonRequest = fs->request({ Resource::Kind::JSON, jsonURL }, util::RunLoop::getLoop(),
+    loader->jsonRequest = fs->request({ Resource::Kind::SpriteJSON, jsonURL }, util::RunLoop::getLoop(),
                                       [this, jsonURL](const Response& res) {
         loader->jsonRequest = nullptr;
         if (res.status == Response::Successful) {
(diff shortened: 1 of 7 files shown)


src/mbgl/util/mapbox.cpp
```diff
@@ -6,53 +6,80 @@ namespace mbgl {
...
+    std::string user;
+    std::string id;
+    bool draft;
+};
+
+bool isMapboxURL(const std::string& url) {
+    return url.compare(0, protocol.length(), protocol) == 0;
+}
+
+void parseMapboxURL(const std::string& url, MapboxURLParts &parts) {
```
jfirebaugh: This function should return its `MapboxURLParts` result. C++ will optimize away the copy.

src/mbgl/util/mapbox.cpp
```diff
@@ -6,53 +6,80 @@ namespace mbgl {
...
         return url;
-
-    std::string result = normalizeURL(url + ".json", "/v4/", accessToken);
-
-    // TileJSON requests need a secure flag appended to their URLs so
-    // that the server knows to send SSL-ified resource references.
-    result += "&secure";
-
-    return result;
+    } else {
```
jfirebaugh: Don't nest `else`s after a guard clause and early return. The reason for guard clauses is to get the uninteresting case taken care of and off your mental stack when reading a method -- using an `else` and extra level of nesting defeats that.

src/mbgl/util/mapbox.cpp
```diff
@@ -6,53 +6,80 @@ namespace mbgl {
...
+}
 
-    const std::string user = url.substr(mapbox.length(), url.find('.') - mapbox.length());
-
-    return normalizeURL(url, "/styles/v1/" + user + "/", accessToken);
+std::string normalizeSpriteURL(const std::string& url, const std::string& format, const std::string& extension, const std::string& accessToken) {
+    if (!isMapboxURL(url)) {
+        return url + format + extension;
+    } else {
+        static MapboxURLParts parts;
```
jfirebaugh: Function-local static means all uses of this variable use the same storage. In most cases, that introduces a data race when the function is called simultaneously from multiple threads. Avoid it unless you have extraordinary circumstances and proper synchronization.

src/mbgl/util/mapbox.cpp
```diff
@@ -6,53 +6,80 @@ namespace mbgl {
...
-        return normalizeURL("mapbox://" + user + "/{fontstack}/{range}.pbf", "/fonts/v1/", accessToken);
+        static MapboxURLParts parts;
+        parseMapboxURL(url, parts);
+        return baseURL + "fonts/v1/" + parts.user + "/{fontstack}/{range}.pbf?access_token=" + accessToken;
     }
 }
 
 std::string normalizeTileURL(const std::string& url, const std::string& sourceURL, SourceType sourceType) {
-    if (sourceURL.empty() || sourceURL.compare(0, mapbox.length(), mapbox) != 0 ||
+    if (sourceURL.empty() || sourceURL.compare(0, protocol.length(), protocol) != 0 ||
```
jfirebaugh: Can use `!isMapboxURL(sourceURL)` here.

src/mbgl/util/geojsonvt
```diff
@@ -1 +1 @@
-Subproject commit ed99a6290fa42107a982e7f3675aae49d29026b1
+Subproject commit 995ffc72c556da4b4880a6036fbcf7159fc5ecce
```
lucaswoj: Whoops. This shouldn't be here.

src/mbgl/map/sprite.cpp
```diff
@@ -43,8 +44,8 @@ Sprite::Sprite(const std::string& baseUrl, float pixelRatio_)
         return;
     }
 
-    std::string spriteURL(baseUrl + (pixelRatio_ > 1 ? "@2x" : "") + ".png");
-    std::string jsonURL(baseUrl + (pixelRatio_ > 1 ? "@2x" : "") + ".json");
+    std::string spriteURL = mbgl::util::mapbox::normalizeSpriteURL(baseUrl, (pixelRatio_ > 1 ? "@2x" : ""), ".png", "NOCOMMIT");
+    std::string jsonURL = mbgl::util::mapbox::normalizeSpriteURL(baseUrl, (pixelRatio_ > 1 ? "@2x" : ""), ".json", "NOCOMMIT");
```
jfirebaugh: We've made Mapbox URL normalization the responsibility of `FileSource` implementations, to permit the implementation in api-gl to load resources directly from the DB and not use an access token at all. We might need some changes to the `FileSource` interface to accommodate `mapbox://` sprite URLs. Mainly I think the implementation needs to be able distinguish sprite JSON requests from other JSON requests (if there are any).

lucaswoj:
@jfirebaugh I think the tests are :white_check_mark: but a little flaky in the absence of https://github.com/mapbox/mapbox-gl-native/pull/2131. Would love your feedback on the content of https://github.com/mapbox/mapbox-gl-native/commit/be83d6f09c99037783ff11f54414a5a8f38fe9f0 while I figure out the tests.

jfirebaugh:
Yeah, this works. Ideally the extension would be tacked on to the URL by the FileSource implementation based on Resource::Kind, and pixel ratio would be passed to the FileSource as a numeric value rather than embedded in the URL, but that will require changing the FileSource interface to use dedicated methods for each kind of resource.

lucaswoj:
> the extension would be tacked on to the URL by the FileSource implementation based on Resource::Kind

This would be pretty easy to do. Should I add it to this PR?

 > pixel ratio would be passed to the FileSource as a numeric value rather than embedded in the URL

This would require some refactoring. Should I add it to this PR?

jfirebaugh:
No, let's land it as is. One last thing: I think `Resource::Kind::JSON` and `Resource::Kind::Image` are now unused. Can you double check and remove them if so?

lucaswoj:
ref https://github.com/mapbox/mapbox-gl-style-spec/issues/309

lucaswoj:
Merged into master via `--ff-only`

PR #2147, opened 2015-08-21
Don't validate/delete shaders

kkaefer:
Some GPU drivers, such as the Tegra 3 (used in the Nexus 7) seem to have issues with our call to `glValidateProgram`. They are technically right about not validating since the time when we validate our shaders, we don't have the correct OpenGL state (e.g. textures) set yet.

Similarly, we're calling `glDeleteShader` in the shader object constructor, after the shader has been attached to the program. While this is technically correct according to the spec, this particular driver seems to crash when doing deleting the shader object. Instead, we can just hold off on deleting the shader object until the shader destructor.

### DIFF
--- a/src/mbgl/shader/shader.cpp
+++ b/src/mbgl/shader/shader.cpp
@@ -67,31 +67,6 @@ Shader::Shader(const char *name_, const GLchar *vertSource, const GLchar *fragSo
         }
     }
 
-    {
-        // Validate program
-        GLint status;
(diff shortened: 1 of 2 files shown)


mb12:
Is it  ensured that  Shader Destructor is called on map Thread?

kkaefer:
@mb12 yeah; the shader objects are owned by the Painter object, which is owned by the MapContext object, which gets destructed on the Map thread.

ljbade:
Interesting, is not not possible to tweak when we do validate program?

Issue #2032, opened 2015-08-10
Move Android C++ files


ljbade:
Currently Android doesn't follow the other platforms as far as location of the C++ files.

Proposal:
- move `android/cpp` to `platform/android`
- move `android/java` to `android`

/cc @bleege @jfirebaugh

kkaefer:
Yeah, let's do this!

ljbade:
Lets get this tidied up before our first release.

Shouldn't take much effort.

PR #2146, opened 2015-08-21
Restructure files

kkaefer:
We should move around some files to have a cleaner structure. In particular, we currently have our debug applications in root level folders, and the platform directory is a mess, making it unclear what headers have to ship with a particular build.

Follow up to #899


ljbade:
Did you also look at https://github.com/mapbox/mapbox-gl-native/issues/2032

jfirebaugh:
Stale.

PR #2153, opened 2015-08-22
Reverse and quicken quickzoom gesture

friedbunny:
This enhances the existing quickzoom gesture:

- Double-tap and hold now immediately engages quickzoom, presuming normal double-tap fails
- During the gesture, moving your finger up now zooms out, moving down zooms in (i.e., reversed)
- Zooming happens slightly more quickly/is more sensitive

/cc @incanus @1ec5 @bleege


friedbunny:
Tests pass locally.

1ec5:
:+1:

PR #2154, opened 2015-08-22
Introduce MGLConfigurationManager

boundsj:
Note: This is a continuation of https://github.com/mapbox/mapbox-gl-native/pull/2149

The previous PR failed to pass CI checks since a test was failing due to the acount/configuration manager name change. That problem is correct here.


boundsj:
Forgot to mention, it does not seem like Appledocs handles deprecation (https://github.com/tomaz/appledoc/issues/80). How does mapbox indicate deprecated methods in docs?

1ec5:
We've been removing the documentation commenting any deprecated method, causing appledoc to ignore that method. That means we have to find a roundabout way to refer to the method in the changelog. See `-[MGLMapView setMapID:]` for example.

include/mbgl/ios/MGLConfigurationManager.h
```diff
@@ -0,0 +1,43 @@
...
++ (void)setAccessToken:(nullable NSString *)accessToken;
+
+/** Retreive the Mapbox API access token for the framework.
+*
+*   You can set an access token on MGLConfigurationManager or on an individual map view. The same token is used throughout the framework.
+*   @return accessToken The Mapbox API access token. */
++ (nullable NSString *)accessToken;
+
+/** @name Providing User Metrics Opt-Out */
++ (void)setMapboxMetricsEnabledSettingShownInApp:(BOOL)showsOptOut __attribute__((unavailable("Set MGLMapboxMetricsEnabledSettingShownInApp in Info.plist.")));
```
1ec5: Since MGLConfigurationManager is replacing MGLAccountManager, this is a good time to remove obsoleted methods. (We marked them `unavailable` instead of `deprecated` in light of being pre-1.0, when backwards compatibility is less of a concern.)

Issue #1834, opened 2015-07-05
Implement camera API


incanus:
We need a camera-like API that allows not only timing curves for viewport changes, but also setting multiple properties simultaneously. Tickets forthcoming, but right now, you can't change both the center/zoom and the heading at the same time, but animated. 

Apple MapKit's `MKMapCamera` API is nice model here, which can later be extended for pitch as well.

1ec5:
cc @jfirebaugh, who was looking at ways to implement this at the mbgl level.

jfirebaugh:
We recently rationalized this on the gl-js side so that camera methods like `jumpTo`, `easeTo`, and `flyTo` all take the same options object of camera parameters (center, zoom, bearing, and pitch). Any omitted parameter defaults to the current value. We should port these method names and parameter conventions to native, although instead of a generic object we'll need to have something like:

```
struct CameraOptions {
    optional<LatLng> center;
    optional<float> zoom;
    optional<float> bearing;
    optional<float> pitch;
};
```

Related: #256.

1ec5:
I’ve started to unify some of the `Transform` methods to funnel into a single method that handles animation. Work is taking place on the [1ec5-camera-1834](https://github.com/mapbox/mapbox-gl-native/tree/1ec5-camera-1834) branch. I’ve left out pitch for now, since that is unimplemented.

The eventual Cocoa Touch API will probably use different terminology than what we’ve used up to this point: per MKMapCamera, we’ll probably have fields called `centerCoordinate`, `heading` (instead of `rotation`), `pitch`, and `altitude` (instead of `zoom`). Who knows, someday we may also want a [`bank`](https://en.wikipedia.org/wiki/Euler_angles#Alternative_names) property too.

PR #2159, opened 2015-08-24
iOS Camera API

bleege:
For #1834


1ec5:
#2116 has been merged, so we need a new PR for this work: #2193.

Issue #2039, opened 2015-08-11
addOverlay & friends


incanus:
Since we added shape annotations, and even though we don't yet support reordering (https://github.com/mapbox/mapbox-gl-native/issues/1728), we should flesh out these methods a bit more. 

From MapKit's list: 

> ## Adding and Inserting Overlays
> - _-addOverlay:level:_
> - _-addOverlays:level:_
> - **-addOverlay:**
> - **-addOverlays:**
> - _-insertOverlay:atIndex:level:_
> - _-insertOverlay:atIndex:_
> - _-insertOverlay:aboveOverlay:_
> - _-insertOverlay:belowOverlay:_
> - _-exchangeOverlay:withOverlay:_
> - _-exchangeOverlayAtIndex:withOverlayAtIndex:_
>
> ## Removing Overlays
> - **-removeOverlay:**
> - **-removeOverlays:**

friedbunny:
There's currently [an error in our docs](https://github.com/mapbox/mapbox-gl-native/blob/bcbb56c1886ef1f369c50ea675a97c48718e78ce/include/mbgl/ios/MGLOverlay.h#L13) that suggests `addOverlay:` is available now, when it is not.

incanus:
Essentially the breakdown here is that `addOverlay[s]` and `removeOverlay[s]` can take only objects conforming to `MGLOverlay`; otherwise it works (for now) just like `add/removeAnnotation[s]`.

incanus:
Closing in favor of the PR in #2252; when we do share reordering in #1728, we can add the associated Cocoa API for management in later milestones.

PR #2169, opened 2015-08-24
Correct docs to remove mention of nonexistent addOverlay: method

friedbunny:
Ref #2039

/cc @incanus @1ec5


incanus:
We should probably hit https://github.com/mapbox/mapbox-gl-native/issues/2039 instead, which we can do in partial to meet the state of docs currently.

friedbunny:
> We should probably hit #2039 instead, which we can do in partial to meet the state of docs currently.

And so we have: #2252

PR #2179, opened 2015-08-25
Node bindings

mikemorris:
Consolidates https://github.com/mapbox/node-mapbox-gl-native/tree/headless-resize (and https://github.com/mapbox/mapbox-gl-native/tree/before-render) into the mapbox-gl-native project under `platform/node` while preserving git history.

`cd platform/node && npm install --build-from-source` will kick off a compile through `make node` calling `node-pre-gyp` (hacked into `scripts/main.mk` currently).

TODO:
- [x] Run node tests along with the regular test suite.
- [x] Fix hanging render tests.
- [x] Clean up hacked-in implementation of `make node` in `scripts/main.mk`
  - ~~Abstract `deps/run_gyp` into an variable perhaps?~~
  - ~~Consolidate `configure` and `build` into a single task?~~
  - [ ] Re-add Node bindings Xcode builds - **punting on this for now**
- [x] Update README.md
- [x] Verify that `publish_binary` commits work
- [ ] ~~Figure out a solution to missing submodules in `npm publish`~~ -> https://github.com/mapbox/mapbox-gl-native/issues/2237
- [ ] ~~Clean up git merge history~~

POST MEREGE:
- [ ] Update CHANGELOG.md
- [ ] Tag and publish `node-mapbox-gl-native@v2.0.0`

/cc @jfirebaugh


jfirebaugh:
- [ ] Move `travis/travis-resources.template` to `cloudformation/travis.template`, to match standard naming conventions. Then merge `platform/node/cloudformation/travis.template` in and deploy. After the merge is complete, delete the `travis-node-mapbox-gl-native` stack.
- [ ] Merge `platform/node/scripts/*`? (Not sure exactly what is duplicate here.)
- [ ] Merge `.travis.yml` (part of "Run node tests along with the regular test suite")

## Test Matrix

Where in the test matrix should node tests go?

node-mapbox-gl-native currently has the following matrix:

* linux, clang, iojs 2.0.1
* linux, clang, node 0.12.0
* linux, clang, node 0.10.36
* linux, gcc, iojs 2.0.1
* linux, gcc, node 0.12.0
* linux, gcc, node 0.10.36
* osx, clang, iojs 2.0.1
* osx, clang, node 0.12.0
* osx, clang, node 0.10.36

Currently this PR adds:

* linux, clang, iojs 2.0.1
* linux, clang, node 0.12.0
* linux, clang, node 0.10.36
* osx, clang, iojs 2.0.1
* osx, clang, node 0.12.0
* osx, clang, node 0.10.36

Providing these builds parallelize well with the others and don't increase the overall time for a build, this seems fine. If they do, then we may want to investigate merging these with existing linux and osx build rows.

I think dropping gcc from the node build matrix is fine.

jfirebaugh:
@mikemorris You mentioned that [these lines](https://github.com/mapbox/mapbox-gl-native/blob/233e78d9f38346943fc42491401fd617674e1c0d/scripts/main.mk#L98-L107) in particular are hacky. What about them should be better?

mikemorris:
@jfirebaugh Was mostly thinking if there would be any way to just preserve the `Makefile/%` tasks and set a `GYP` var to use `node-pre-gyp` instead of `deps/run_gyp` or `MAKE`? It's also currently dumping a whole bunch of `*.mk` and a `mbgl.Makefile` into `/platform` for some reason, hence https://github.com/mapbox/mapbox-gl-native/blob/233e78d9f38346943fc42491401fd617674e1c0d/.gitignore#L31-L32 - would be nice to clean that up.

mikemorris:
I think it might be good to switch back to `nvm` node installs instead of `mason` to be able to keep up-to-date more easily, any objections @jfirebaugh?

jfirebaugh:
Would that be a new dev dependency? I'd rather not add another requirement.

mikemorris:
Would only be a Travis dependency @jfirebaugh. Alternative would be creating new `mason` branches regularly to keep recent versions available, which is a bit of a hassle if there's not an obvious benefit.

jfirebaugh:
Ah, yes, using nvm on Travis is :+1:.

bsudekum:
@mikemorris what are your thoughts on making logging optional?

jfirebaugh:
* [ ] Consolidate headless test scripts:
  * Do all headless testing using the node bindings
  * Remove headless tests from non-node rows of the test matrix
  * Remove `test/headless/{headless.cpp,server.js}
  * Remove `test/suite/bin/{compare_images,server}.py`

bsudekum:
How will this project be required in other node projects? I ask because the `node` product is in the subdirectory. Does this [`package.json`](https://github.com/mapbox/mapbox-gl-native/blob/node/platform/node/package.json) need to be moved to the root directory?

mikemorris:
> @mikemorris what are your thoughts on making logging optional?

The current implementation uses `NodeLogObserver` to monitor the `mbgl::Log`, and exposes an `EventEmitter` interface for listening to output in JavaScript. All logging should be done through this interface, which can either be ignored downstream, configured to write to `console.log`, or anything else.

> How will this project be required in other node projects? I ask because the node product is in the subdirectory. Does this package.json need to be moved to the root directory?

For releases this will still be published to npm and included like `mapbox-gl-native: 2.0.0`. For staging feature branches, I haven't yet looked into whether `mapbox-gl-native: https://git-repo#feature-branch` could work. I think we'll have to use `npm publish <tarball>` instead of `npm publish <folder>` for publishing source unless we move `package.json` to the root though.

mikemorris:
> It's also currently dumping a whole bunch of *.mk and a mbgl.Makefile into /platform for some reason

This is fixed with moving `package.json` and `binding.gyp` to root.

kkaefer:
\o/

Issue #2182, opened 2015-08-26
Add Density Specific Compass Images


bleege:
Android currently uses 1 image size for Compass which is then scaled per density.  iOS provides 3 different assets for each of their primary densities.  Let's repurpose those and fill in missing densities using `@3x` as the source and scale it down.

https://github.com/mapbox/mapbox-gl-native/blob/master/platform/ios/resources/Compass.png

https://github.com/mapbox/mapbox-gl-native/blob/master/platform/ios/resources/Compass@2x.png

https://github.com/mapbox/mapbox-gl-native/blob/master/platform/ios/resources/Compass@3x.png

bleege:
Sizes:

1x = 36x36
2x = 72x72
3x = 104x104

bleege:
Going to use the following scaling rules as it was the one suggested by Google.  ;-)

Accordingly, will start with this `1x == mdpi` and `2x == xhdpi`.

http://stackoverflow.com/questions/11581649/about-android-image-and-asset-sizes

![screen shot 2015-08-26 at 11 08 11 am](https://cloud.githubusercontent.com/assets/989928/9501875/cd15e5f4-4be2-11e5-8de7-52c44ca8c8ec.png)

bleege:
So, change of direction here.  Instead of adjusting an iOS styled compass, I've replaced it with a more Android styled compass image.  This will likely (and should) get replaced in time, but for an initial release this should work.  Let's see how it plays in test though.

![device-2015-08-26-172427](https://cloud.githubusercontent.com/assets/989928/9509729/d758828a-4c17-11e5-8bbf-bb4393a4945c.png)
_New Compass Image On Samsung S5_

ljbade:
Nice @bleege, looks way cooler.

ljbade:
Hmm I think I'm going to disable the elevation on the FAB again. Looks nice on my phones (well except for Samsung), then I see a screenshot like this...

bleege:
Merged.

tobrunvannuland:
@bleege this looks way better, but IMHO it is a bit pixelated.
Could a designer produce an svg for us? 
Android Studio has a Vector Asset wizard that generate pngs.

ljbade:
I also noticed that the white section of the compass is transparent. So it lets through the map in that triagngle.

bleege:
[...]

PR #2187, opened 2015-08-27
Android Style Compass Image

bleege:
For #2182


bleege:
The "failing build" is https://travis-ci.org/mapbox/mapbox-gl-native/jobs/77425154 which is actually iOS flavored and relates to a threading issue.  Neither of these have anything to do with the Android image change so I'm feeling ok to merge.  #travis

Issue #1581, opened 2015-05-18
Animated MGLMapView methods should accept completion handlers


1ec5:
Methods like `-[MGLMapView setCenterCoordinate:animated:]` should have corresponding methods that take completion handlers (callback blocks) as parameters, like `-[MGLMapView setCenterCoordinate:animated:completionHandler:]`. Completion handlers feel much more natural in modern Objective-C and especially Swift, and they’d make it much easier to chain together animated changes to the map.

jhoughjr:
Here here!

kdgwill:
Was their any further discussion on this, or does their at least exist a work around to determine when an mapbox animation has been completed?

robipresotto:
+1

1ec5:
GL JS can do this; so should the iOS SDK. The most straightforward way to implement it would be to have `CameraOptions` hold a C++ lambda that we can stuff the completion handler block into. Then `easeTo()` can plumb that through to the animation code’s completion code. I’m not sure what effect cancellation will have on this approach, though. For one thing, #2313 is still outstanding.

1ec5:
> does their at least exist a work around to determine when an mapbox animation has been completed? 

@kdgwill, `-[MGLMapViewDelegate mapView:regionDidChangeAnimated:]` is already available; it gets called for both programmatic and user-initiated region changes. It is possible to chain animations using that callback, but of course the completion handlers proposed here would be more convenient for that use case.

PR #2193, opened 2015-08-27
Camera API

### LINKED ISSUE #256: flyTo equivalent
It would be nice to port ballistic zoom-pan from llmr, as it is usually very useful in demos. I might look into it.

### LINKED ISSUE #1643: iOS setDirection:animated: cancels setCenterCoordinate:animated:
When the following calls are made:

```
[self.mapView setCenterCoordinate:newLocation.coordinate animated:YES];
[self.mapView setDirection:newLocation.course * -1 + 360 animated:YES];
```

The first call to `setCenterCoordinate:` is ignored. If the order of the calls is reversed, then the call to `setDirection:` is ignored.

Ideally, there would be a single method to set both centerCoordinate and direction with both items animated.

1ec5:
The pre-existing methods on `mbgl::Map` only allow you to modify 2–3 degrees of freedom at once. This PR adds the concept of the viewer’s “camera”, allowing you to transition center coordinate, zoom level, rotation, and pitch all in one animation, and to configure that animation. The camera is represented by a `CameraOptions` struct and manipulated by `jumpTo()` and `easeTo()` methods, much like in Mapbox GL JS. On iOS, the camera is represented by an MGLMapCamera object that imitates MKMapCamera.

* [x] CameraOptions
* [x] `jumpTo()`
* [x] `easeTo()` (fixes #1643, #1834)
* [ ] `flyTo()` (will fix #256)
* [x] MGLMapCamera
* [x] Round-trip camera altitude
* [x] …with rotation
* [x] …with pitch in portrait orientation
* [ ] Get altitude values closer to MapKit’s (#2266)
* [x] Clean up existing methods on `mbgl::Map`
* [x] Fix Linux builds
* [x] Custom animation durations and timing functions on iOS (ref #1581)
* [x] Move iosapp persistence over to MGLMapCamera
* [x] Remove standalone `MGLMapView.pitch`

(This PR was originally opened by @bleege as #2159, but we’ll need to merge into master instead of the perspective branch now that #2116 has landed.)

/cc @incanus @friedbunny @ansis @bleege


incanus:
:movie_camera: :stuck_out_tongue: :+1:

1ec5:
One small change here is that, where currently possible, I’ve changed some animation timing curves from linear to the system default, which eases in a little. So that’ll make the user dot move a little less jarringly, at least until we manage to make it move continuously (and animate linearly) for #1582.

src/mbgl/map/map.cpp
```diff
@@ -271,8 +289,8 @@ void Map::resetNorth() {
 
 #pragma mark - Pitch
 
-void Map::setPitch(double pitch) {
-    transform->setPitch(std::min(pitch, 60.0) * M_PI / 180);
+void Map::setPitch(double pitch, const Duration& duration) {
+    transform->setPitch(util::clamp(pitch, 0., 90.) * M_PI / 180, duration);
```
ansis: The 60 here was intentional. We don't support level of detail loading so if the pitch is set it to 90 it will try to load all the tiles in the world

1ec5: Ah, right, this is an artifact from rebasing.

1ec5:
Momentarily punting on #256 and #2266, but that’s what I’ll look at next. Does this look ready to merge @incanus?

incanus:
Cursory glance looks good. Let me pull it down & double-check.

incanus:
No red flags to me, but we'll roll a build and make sure things feel good. I think at this point we are :+1:.

include/mbgl/ios/MGLMapCamera.h
```diff
@@ -0,0 +1,36 @@
+#import "Mapbox.h"
+
+#pragma once
```
incanus: Curious @1ec5 why this is required, for an iOS header which should be brought in with the once-safe `#import` vs. `#include` anyway.

incanus:
Per chat last week, this is ready to merge; additional milestones will be hit in #256 and #2266.

Issue #2194, opened 2015-08-27
Remove Height / Width On Test App FAB


bleege:
[FAB shadow issue](https://github.com/mapbox/mapbox-gl-native/issues/2104#issuecomment-132104443) with Design Support Library.  Let's pull it for now in the TestApp.

bleege:
Current rendering on Samsung S5.

![device-2015-08-27-124436](https://cloud.githubusercontent.com/assets/989928/9531187/66d53042-4cb9-11e5-81ca-d5702ccca698.png)

bleege:
Now.... shadowless!

![device-2015-08-27-133025](https://cloud.githubusercontent.com/assets/989928/9532368/1ac878f6-4cc0-11e5-97ec-41d35ef012a0.png)

bleege:
Merged.

ljbade:
If anything it matches the flat compass.

PR #2195, opened 2015-08-27
Removing FAB Shadow In TestApp

bleege:
For #2194


bleege:
Travis Failed on iOS Metrics Test, which is a known issue.

Merging.

PR #2201, opened 2015-08-28
fixes #1969: support for marker tapping in Android

### LINKED ISSUE #1969: Add getAnnotationsInBounds in NativeMapView.java
I'm looking at implementing annotation selection in Android but i need the getAnnotationsInBounds method.

incanus:
Just logs the marker ID to console right now. 

@bleege @ljbade give it a whirl and let me know how it could be better, in particular [this routine](https://github.com/mapbox/mapbox-gl-native/blob/0c58172020fb13e0596667f3fd2125df6fc33e1f/android/java/MapboxGLAndroidSDK/src/main/java/com/mapbox/mapboxgl/views/MapView.java#L432-L449) feels a little wasteful, but I wasn't sure of the best way to make use of `ArrayList.contains()` otherwise given a `long[]`. 

We can tune this up possibly, then merge and pick up work in https://github.com/mapbox/mapbox-gl-native/issues/894 for popups connecting to this. 

Another weirdness is a `60x80` top two-thirds-weighted hit region felt best on Android as opposed to the `40x60` we use on iOS. I am making sure to take `MapView.mScreenDensity` into account.

### DIFF
--- a/android/java/MapboxGLAndroidSDK/src/main/java/com/mapbox/mapboxgl/views/MapView.java
+++ b/android/java/MapboxGLAndroidSDK/src/main/java/com/mapbox/mapboxgl/views/MapView.java
@@ -50,6 +50,7 @@ import com.mapbox.mapboxgl.annotations.Polygon;
 import com.mapbox.mapboxgl.annotations.PolygonOptions;
 import com.mapbox.mapboxgl.annotations.Polyline;
 import com.mapbox.mapboxgl.annotations.PolylineOptions;
+import com.mapbox.mapboxgl.geometry.BoundingBox;
 import com.mapbox.mapboxgl.geometry.LatLng;
 import com.mapbox.mapboxgl.geometry.LatLngZoom;
 import com.mapzen.android.lost.api.LocationListener;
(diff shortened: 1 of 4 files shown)


incanus:
You know what? Marker picking isn't performance-constrained, so if we can find improvements to the existing annotation iteration in future, feel free to change them. But for now I am going to squash + force push + merge this so we can build on it.

ljbade:
@incanus There is a function `Arrays.toList()` http://docs.oracle.com/javase/7/docs/api/java/util/Arrays.html#asList(T...)

ljbade:
Ah can't use it anyway because of `long` vs `Long`. Isn't Java neat?

android/java/MapboxGLAndroidSDK/src/main/java/com/mapbox/mapboxgl/views/MapView.java
```diff
@@ -428,6 +429,25 @@ public void removeAnnotations() {
...
+        List<Long> idsList = new ArrayList<>();
+        for(int i = 0; i < ids.length; i++) {
+            idsList.add(new Long(ids[i]));
+        }
+        for(int i = 0; i < mAnnotations.size(); i++) {
+            Annotation annotation = mAnnotations.get(i);
+            if (annotation instanceof Marker && idsList.contains(annotation.getId())) {
+                annotations.add(annotation);
+            }
+        }
```
bleege: Since `mAnnotations` is defined using Generics `List<Annotation>` it can be iterated easier and safer via:

```java
for (Annotation annotation : mAnnotations) {
...
}
```

bleege:
> Ah can't use it anyway because of long vs Long. Isn't Java neat?

Sure it can.  Autoboxing.  :smiley:

ljbade:
@bleege Ah yeah but apparently autoboxing does not happen with arrays.

PR #2209, opened 2015-08-31
Tidy up missing JNI deletes

ljbade:


incanus:
This is good; thanks @ljbade. I was thinking the deletes referred to cleanup only of used classes, but rather it's a cleanup of all prior allocations. 

This seems ripe for consolidation into common setup/teardown routines. It feels like so much of `jni.cpp` is boilerplate code that is prone to this sort of error.

ljbade:
@incanus Yeah the JNI stuff is not easily scalable.

Issue #1887, opened 2015-07-13
MGLMapboxEvents tests are failing


1ec5:
`-[MetricsTests testPushEventAddsToEventQueue]` and `-[MetricsTests testTimerCreatedAfterEventPushed]` sometimes [time out on Travis](https://travis-ci.org/mapbox/mapbox-gl-native/jobs/70809933#L1016), causing test failures that I can’t reproduce locally.

/cc @bleege @incanus

brunoabinader:
Got this in https://travis-ci.org/mapbox/mapbox-gl-native/jobs/70906410.

1ec5:
These tests have started failing locally for me as well. They never fail individually, only when running the whole suite.

jfirebaugh:
@1ec5: Are the failures due to code changes (bisectable?) or something external?

1ec5:
@jfirebaugh, I believe these failures may be influenced by code changes. I had a go at bisecting today, but it was a bit confusing because different failures came in and out at different times – pretty consistently for each commit. So there may be multiple issues.

1ec5:
Some of the failures I’m seeing are legitimate failures, better described by #2036.

PR #2213, opened 2015-08-31
Disable iOS metrics tests

1ec5:
They’re simply unreliable on Travis. The spurious failures are getting in the way of development, and we know from running them locally that they’re fine.

ref #1887

/cc @jfirebaugh @incanus @friedbunny


friedbunny:
This gets the hugest thumbs up from me.

PR #2227, opened 2015-09-01
Optimize bundled images' file sizes

friedbunny:
Saves 8KB, because we care. The `.bundle` creation process does not appear to crush these files, so let's do it manually. Tested to work!

/cc @1ec5 @incanus


incanus:
Is there a good way we could instead add this to the build process, so that in future we could just drop new images in? 

And/or could we add a `README` to the bundle or resources folder indicating the need to do this on future images?

friedbunny:
I'll add a `README` in `platform/ios/resources/` explaining the process.

I prefer [ImageOptim](https://imageoptim.com) for my image optimization, which uses a combination of:

> PNGOUT, Zopfli, Pngcrush, AdvPNG, extended OptiPNG, JpegOptim, jpegrescan, jpegtran, and Gifsicle

... which would be cumbersome to replicate in the build process. Unless we anticipate adding many new image resources, I don't think it's worth the effort to automate.

PR #2229, opened 2015-09-01
Tidy up iOS resources

friedbunny:
Follows on from the monumentally insignificant image size reductions in #2227.

- Adds `README` to `platform/ios/resources/` reminding future-us to use ImageOptim
- Includes the `platform/ios/resources/` folder in the iOS Xcode project
- Excludes `README` from the final `.bundle` compilation

/cc @incanus @1ec5

### DIFF
--- a/gyp/platform-ios.gypi
+++ b/gyp/platform-ios.gypi
@@ -61,6 +61,7 @@
         '../platform/ios/NSString+MGLAdditions.m',
         '../platform/ios/vendor/SMCalloutView/SMCalloutView.h',
         '../platform/ios/vendor/SMCalloutView/SMCalloutView.m',
+        '../platform/ios/resources/',
       ],
 
       'variables': {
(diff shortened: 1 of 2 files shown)


gyp/platform-ios.gypi
```diff
@@ -61,6 +61,7 @@
         '../platform/ios/NSString+MGLAdditions.m',
         '../platform/ios/vendor/SMCalloutView/SMCalloutView.h',
         '../platform/ios/vendor/SMCalloutView/SMCalloutView.m',
+        '../platform/ios/resources/',
```
incanus: Why?

friedbunny: So that it's visible in the project — it's not needed for compilation. Had a situation last week where I wanted to show Brad the compass images but needed to jump out to Github to find them.

friedbunny: Although, I didn't notice this before — where does it come from? It doesn't include `README` so is it the unbundled `.bundle`?

screen shot 2015-09-02 at 1 53 04 pm

PR #2233, opened 2015-09-02
User location puck for course tracking mode

### LINKED ISSUE #2188: Add puck for course tracking mode
The puck is the on-screen element that's shown during course-based user location tracking, usually in conjunction with navigation.

puck

/cc @incanus @1ec5

friedbunny:
Adds a "puck" that replaces the user location dot when `MGLUserLocationTrackingModeCourse` is active.

- Moves the current dot drawing code to `-drawDot`
- Adds `-drawPuck`
- Adds `_puckModeActivated` boolean
 - Used to reset the layers when switching between dot/puck
- Draws a bezier arrow graphic for the puck in `-puckArrow`

There are still a variety of things to do:

- [x] Pitch the puck with `CATransform3D`
- [x] Code cleanup
- [x] Tweak visual style (shadow?)
- [x] Test with rotated/pitched camera
- [ ] Add option to use the puck without `MGLUserLocationTrackingModeCourse`?
- [ ] Update `MGLMapView` where necessary
- [ ] Add transitions when switching between dot and puck

simulator screen shot sep 1 2015 11 06 05 pm

Fixes #2188.

/cc @incanus @1ec5 @mapbox/mobile

### DIFF
--- a/platform/ios/MGLUserLocationAnnotationView.m
+++ b/platform/ios/MGLUserLocationAnnotationView.m
@@ -144,16 +144,6 @@ const CGFloat MGLUserLocationAnnotationArrowSize = MGLUserLocationAnnotationPuck
     }
 }
 
-CGFloat MGLRadiansFromDegrees(CLLocationDegrees degrees)
-{
-    return degrees * M_PI / 180;
-}
-
-CLLocationDegrees MGLDegreesFromRadians(CGFloat radians)
-{
-    return radians * 180 / M_PI;
-}
-
 - (UIBezierPath *)puckArrow
 {
     CGFloat max = MGLUserLocationAnnotationArrowSize;
--- a/platform/ios/MGLUserLocationAnnotationView.m
+++ b/platform/ios/MGLUserLocationAnnotationView.m
@@ -31,6 +31,7 @@ const CGFloat MGLUserLocationAnnotationArrowSize = MGLUserLocationAnnotationPuck
     CALayer *_accuracyRingLayer;
     CALayer *_dotBorderLayer;
     CALayer *_dotLayer;
+    CALayer *_dotDomeLayer;
 
     double _oldHeadingAccuracy;
     CLLocationAccuracy _oldHorizontalAccuracy;
@@ -79,6 +80,7 @@ const CGFloat MGLUserLocationAnnotationArrowSize = MGLUserLocationAnnotationPuck
 
         _haloLayer.backgroundColor = [tintColor CGColor];
         _dotLayer.backgroundColor = [tintColor CGColor];
+        _dotDomeLayer.backgroundColor = [tintColor CGColor];
 
         _headingIndicatorLayer.contents = (__bridge id)[[self headingIndicatorTintedGradientImage] CGImage];
     }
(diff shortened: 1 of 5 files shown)


platform/ios/MGLUserLocationAnnotationView.m
```diff
@@ -55,189 +65,291 @@ - (instancetype)initWithCoder:(NSCoder *)decoder
...
+    }
+
+    if (self.mapView.pitch)
+    {
+        CATransform3D t = CATransform3DRotate(CATransform3DIdentity, MGLRadiansFromDegrees(self.mapView.pitch), 1.0, 0, 0);
+        self.layer.sublayerTransform = t;
+    }
+}
+
+CGFloat MGLRadiansFromDegrees(CLLocationDegrees degrees)
```
1ec5: These methods are also defined in MGLMapView.mm. Can you factor these implementations out into MGLGeometry.h?

friedbunny: Yup, I'll do that.

friedbunny:
Tilting the map now also tilts the puck:

pitchedpuck

Also: rebased on `master` and toned down the shadow (a bit too much, as it turns out).

We explicitly don't do this right now, but tilting the user dot actually looks pretty good, too:

dot

friedbunny:
### I added pitch to the user dot.

Why?

- The accuracy ring looks terrible unpitched on a tilted map
- It's cool and looks better, even if it diverges from Apple Maps

The pitched accuracy ring still has sizing issues: vertical screen movement isn't accounted for and all elements of the annotation (and puck) keep a constant scale, even though they shouldn't.

1ec5:
I agree with tilting the accuracy ring, but tilting the dot itself looks odd unless we give it some thickness or elevation.

friedbunny:
screen shot 2015-09-04 at 1 46 08 pm

incanus:
http://a.loudgif.com/1ilWRZ0

1ec5:
As of 17003db8675eb905b36af3b2615c22a51d78df15:

bubble

Back to @friedbunny for review.

friedbunny:
My current thinking is that:

1. Setting the `transform` property on many individual `CALayer`s is too slow and introduces noticeable lag when you're using the tilt gesture.
2. Go back to setting the parent `CALayer.subLayerTransform` and then retransform the dome layer to flat.
3. ... but this adds flickering in the dot layer and the dome layer during the tilting gesture.
4. So only allow the (animated) dome transform to happen after the tilting is over. WIP.

friedbunny:
Still fighting flickering during the tilt gesture, tracked it down to the animations. Now animations are paused and unpaused when the gesture starts/finishes.

Sadly, I'm considering ditching the flying-saucer look and seeing if not pitching the dot (only the accuracy ring) is a viable option.

friedbunny:
The masking appears to be an issue here, too. If we were to only transform the inner dot (and not split it in two, then mask one side) that might improve the flickering.

1ec5:
Instead of masking, maybe we could draw both halves of the dot as actual semicircles.

friedbunny:
I think we're ready to go here. Next step is to rebase, squash, and get some eyes on this.

faker-depth

Ended up removing [the dome](https://github.com/mapbox/mapbox-gl-native/pull/2233#issuecomment-137851118) and going with another faux 3D-ism by [setting the `zPosition` for the inner dot](https://github.com/mapbox/mapbox-gl-native/pull/2233/files#diff-e09e3dda7b5ee25776e6b32cb58f1eb4R127) to raise as the map is tilted. This skips the flickering and/or lag that I failed to eliminate from the dome implementation.

1ec5:
If we ever allow pitching beyond 60°, we’ll have to revisit this approach and probably figure out how to make the dome work, but this looks great as-is.

friedbunny:
screen shot 2015-09-10 at 2 06 07 pm

friedbunny:
Squashed over in #2300. Thanks for the memories, #2233.

Issue #1957, opened 2015-07-28
Adopt Bitcode


1ec5:
iosapp fails to build for iOS 9 devices in Xcode 7 with the following linker error:

```
ld: '/Users/mxn/hub/mapbox-gl-native/mason_packages/ios-9.0/libuv/0.10.28/lib/libuv.a(async.o)' does not contain bitcode. You must rebuild it with bitcode enabled (Xcode setting ENABLE_BITCODE), obtain an updated library from the vendor, or disable bitcode for this target. for architecture arm64
```

We’ll need to patch upstream and update the mason submodule pin.

/cc @kkaefer @incanus

PR #2238, opened 2015-09-02
Upgrade Travis to Xcode 7 image

### LINKED ISSUE #2205: Bitcode error on iOS 9.0 (Xcode 7 Beta 6)
When compiling my application for iOS 9.0 (Xcode 7 Beta 6), I see the following build error:

```
 ld: '/Users/tom/projects/CityCyclist/app/Pods/Mapbox-iOS-SDK/libMapbox.a(SMCalloutView.o)' does not contain bitcode. You must rebuild it with bitcode enabled (Xcode setting ENABLE_BITCODE), obtain an updated library from the vendor, or disable bitcode for this target. for architecture armv7
```

I'm using MapboxGL through Cocoapods, setting the podspec to `https://raw.githubusercontent.com/mapbox/mapbox-gl-native/ios-v2.1.0-pre.2/ios/Mapbox-iOS-SDK.podspec`.

Let me know if I can provide any more info.

1ec5:
This PR upgrades our Travis builds to use an [Xcode 7 image](http://blog.travis-ci.com/2015-08-27-xcode-70-beta-for-oss/). The same image includes new versions of other packages too, so we should be on the lookout for any fallout from that:

> * OS X: 10.10.4 (was: 10.10.3)
> * RubyMotion: 3.14 (was: 3.13)
> * xctool: 0.25
> * Git 2.5.0 (was: 2.4.4)
> * OpenSSL: 1.0.2d_1 (was: 1.0.2c)

Fixes #2205 via #1957 and mapbox/mason#106.

/cc @incanus @kkaefer @friedbunny @bsudekum

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -28,7 +28,7 @@ matrix:
       osx_image: xcode6.4
       env: FLAVOR=osx BUILDTYPE=Debug
     - os: osx
-      osx_image: xcode6.4
+      osx_image: xcode7
       env: FLAVOR=ios BUILDTYPE=Release
     - os: linux
       env: FLAVOR=android ANDROID_ABI=arm-v7 BUILDTYPE=Release
--- a/scripts/ios/test.sh
+++ b/scripts/ios/test.sh
@@ -4,8 +4,11 @@ set -e
 set -o pipefail
 set -u
 
-xctool \
+xcodebuild \
     -project ./test/ios/ios-tests.xcodeproj \
     -scheme 'Mapbox GL Tests' \
-    -sdk iphonesimulator8.4 \
-    test
+    -sdk iphonesimulator \
+    -destination 'platform=iOS Simulator,name=iPhone 6,OS=latest' \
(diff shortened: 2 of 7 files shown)


1ec5:
Supposedly this image has OS X 10.10.4, but Mason is trying to build against the OS X 10.11 SDK for some reason. That won’t work because of the .tbd files ([text-based stub libraries](https://forums.developer.apple.com/thread/4572)).

1ec5:
Travis’s iOS builds run xctool 0.2.5, which doesn’t quite support the latest Xcode 7 seed yet: facebook/xctool#528 + rdar://problem/22542632 . In particular, I’m running into an “XCTestConfiguration isn't available” assertion failure in `-[OCUnitIOSAppTestRunner testEnvironmentWithSpecifiedTestConfiguration]`, and it sounds like there will be downstream errors as well.

incanus:
@1ec5 says this is good to go. Let's lead `ios-v2.1.1` with Bitcode fixes — this one and https://github.com/mapbox/mapbox-gl-native/issues/2205.

Issue #934, opened 2015-03-02
Remove TargetAPI Annotations


bleege:
The `@TargetAPI` annotations just hide things from Lint and create spaghetti code.  Either something works for the supported API range or it doesn't.  For more [background see this issue from the Raster SDK](https://github.com/mapbox/mapbox-android-sdk/issues/292#issuecomment-39466159).

ljbade:
Well we can remove them and make API level errors warning instead, but then I won't be at zero errors/warnings again.

Perhaps we just drop support for Android 2.x completely?

bleege:
> then I won't be at zero errors/warnings again.

That's a good thing!  It means that the stuff to fix is visible and not hiding waiting for the most inopportune time to cause trouble.  :-)

> Perhaps we just drop support for Android 2.x completely?

API 8 would probably be okay to drop, but there are still a healthy sliver of API 10 devices around.

bleege:
Here's an updated API Market Share report from the [Android Dashboard](https://developer.android.com/about/dashboards/index.html).  Based on this I think we should definitely drop Froyo and possibly drop Gingerbread.  [Ice Cream Sandwich](https://developer.android.com/about/versions/android-4.0.html) was a major change UI wise as well as some updates to core UI classes (like [View](http://developer.android.com/reference/android/view/View.html)) and Input handling.

If we kept Gingerbread and up that would put the minimum at API 10 / 11.  If we kept Ice Cream Sandwich and up that would put the minimum at API 14.  Thoughts @ljbade @incanus?

![20150813-android-api-market-share](https://cloud.githubusercontent.com/assets/989928/9255589/00c9286c-41b0-11e5-90f8-949dbaca8d2e.png)

incanus:
Weird that Gingerbread has more than ICS. Is there a good rationale for that? 

Also, maybe dates here would be useful as we think about how old particular releases matter (though I realize market share is more important).

mb12:
Maybe that's because all carriers push at least one OS upgrade. Majority Froyo device got upgraded to Gingerbread and ICS to Jellybean.

bleege:
According to [Wikipedia's Android Version History](https://en.wikipedia.org/wiki/Android_version_history) here's how things break out:

| Codename | Version | Release Date |
| --------- | ---------- | ------------------- |
| Gingerbread| 2.3.7 | 21-Sept-2011 |
| Ice Cream Sandwich | 4.0.3 | 16-Dec-2011 |
[...]

PR #2240, opened 2015-09-02
Refactor constructors MapView

tobrun:
Constructor targeted at Lollipop and higher, 
Using defStyle and defStyleRes to obtain attributes, 
Android annotations for constructor arguments

### DIFF
--- a/android/java/MapboxGLAndroidSDK/src/main/java/com/mapbox/mapboxgl/views/MapView.java
+++ b/android/java/MapboxGLAndroidSDK/src/main/java/com/mapbox/mapboxgl/views/MapView.java
@@ -21,7 +21,10 @@ import android.net.ConnectivityManager;
 import android.net.NetworkInfo;
 import android.os.Build;
 import android.os.Bundle;
+import android.support.annotation.AttrRes;
 import android.support.annotation.NonNull;
+import android.support.annotation.Nullable;
+import android.support.annotation.StyleRes;
 import android.support.v4.content.ContextCompat;
 import android.support.v4.view.GestureDetectorCompat;
 import android.support.v4.view.ScaleGestureDetectorCompat;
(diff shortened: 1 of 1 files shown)


ljbade:
Hmm we decided to avoid `@TargetApi` in https://github.com/mapbox/mapbox-gl-native/issues/934

@tobrunvannuland Is it neccessary to support the new Lollipop constructor? What happens if you don't use `defStyle` and `defStyleRes`?

/cc @bleege

bleege:
> Hmm we decided to avoid @TargetApi in #934

Agreed.  Unless there's a backport available that supports the entire supported API range (currently 15+) we want to avoid `@TargetAPI` as it inevitably leads to hidden API range incompatibilities.

tobrun:
About the @TargetApi:
Sorry, I wasn't aware of that discussion and I completely agree with the decision made!

About Lollipop constructor:
The main gist behind it is that the users would have an extra way of customising the map.

From the Android docs:
> defStyleAttr An attribute in the current theme that contains a reference to a style resource that supplies defaults values for the StyledAttributes. Can be 0 to not look for defaults.

> defStyleRes A resource identifier of a style resource that supplies default values for the StyledAttributes, used only if defStyleAttr is 0 or can not be found in the theme. Can be 0 to not look for defaults.
 
I saw that another issue was created for the Android Annotations, so closing the PR

bleege:
> Sorry, I wasn't aware of that discussion and I completely agree with the decision made!

@tobrun Not a problem at all!  This is a HUGE project with tons of tickets.  No one is (nor can be) aware of everything.  :-)

ljbade:
@tobrun I find it useful to use GitHub's issue filter dropdowns to select only issues tagged with Android. Makes it a lot easier.

Issue #123, opened 2014-03-27
Overzoom cached/offline tiles


incanus:
Something that would boost perceived performance, at least on raster, is to consider a `TileData::state::cached` somewhere between `invalid` and `loaded`. It's odd because it's logic branching, not serial, but with this we could consider loading and parsing/decoding for use in child/parent tile scenarios. 

To put it another way, this wouldn't require the user to have ever zoomed to a parent or child tile zoom level in the current run of the app in order to make use of those tiles for drawing. If I launch the app at `z8` but have one or more `z7` tiles cached, they could get drawn until I get all the `z8` tiles I need for optimum resolution. 

There's a tradeoff between spending time pulling these from cache vs. spending resources in getting the `z8` tiles, but it might be worthwhile.

jfirebaugh:
Implemented in #5143.

PR #2248, opened 2015-09-03
When network is unreachable, always read from cache (even if expired)…

bwhtmn:
When network is unreachable, always read from cache (even if expired) rather than attempting to fetch remote data. Improves reliability of map view when network is unavailable.

### DIFF
--- a/include/mbgl/storage/network_status.hpp
+++ b/include/mbgl/storage/network_status.hpp
@@ -11,6 +11,8 @@ namespace mbgl {
 class NetworkStatus {
 public:
     static void Reachable();
+    static void Unreachable();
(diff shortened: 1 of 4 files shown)


ljbade:
@kkaefer Should we do this?

kkaefer:
yes, also see https://github.com/mapbox/mapbox-gl-native/issues/123

kkaefer:
I agree with the fact that we should show expired tiles, but I don't agree with this implementation. Instead, we should load always show expired tiles (see #123) and at the same time trigger a request to ensure freshness and then update the data if it has changed since (and we have network). I think this request is sufficiently captured by #123.

Issue #1956, opened 2015-07-28
Add arm64 As Supported ABI


bleege:
Nexus 9 devices are arm64 based so Android GL needs to provide a arm64 based `libmapboxgl.so` binding.  Update build process to support this ABI.

https://developer.android.com/ndk/guides/abis.html

/cc @incanus @ljbade

bleege:
Added `make android-lib-arm64-v8a` to `ANDROID_ABIS` to get the process started.  When runs it produces:

```sh
Brads-MacBook-Pro:mapbox-gl-android brad$ make android-lib-arm64-v8a
git submodule update --init src/mbgl/util/geojsonvt
deps/run_gyp android/mapboxgl-app.gyp -Dhost=android -Iconfig/android-arm64-v8a.gypi -Dheadless_lib=none -Dplatform_lib=android -Dasset_lib=zip -Dhttp_lib=curl -Dcache_lib=sqlite --depth=. -Goutput_dir=. --generator-output=./build/android-arm64-v8a -f make-android
gyp: Undefined variable openssl_ldflags in android/mapboxgl-app.gyp
make: *** [Makefile/android-arm64-v8a] Error 1
```

bleege:
Discovered that the `android-arm64-v8a.gypi` is missing `openssl_ldflags` variable compared to what it looks like in `android-arm-v7.gypi`

**android-arm64-v8a.gypi**
```
# Do not edit. Generated by the configure script.
{
  'target_defaults': {
    'cflags%': [],
    'default_configuration': 'Release',
    'defines': [],
    'include_dirs': [],
    'libraries': []
  },
  'variables': {
    'boost_cflags%': ['-I/Users/brad/development/mapbox-gl-android/mason_packages/headers/boost/1.57.0/include'],
    'boost_program_options_static_libs%': ['/Users/brad/development/mapbox-gl-android/mason_packages/osx-10.10/boost_libprogram_options/1.57.0/lib/libboost_program_options.a'],
    'curl_static_libs%': [],
    'curl_cflags%': ['-isysroot', '/Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX10.10.sdk', '-arch', 'x86_64', '-mmacosx-version-min=10.8', '-I/Users/brad/development/mapbox-gl-android/mason_packages/osx-10.10/libcurl/system/include'],
[...]

Issue #1683, opened 2015-06-03
Add ARM v5, x86-64 and MIPS 64 support to Travis


ljbade:
Just noticed for whatever reason we don't build `arm-v5`, `x86-64` and `mips-64` ABIs.

ljbade:
@jfirebaugh We probably don't want to be expanding the build matrix? If anything we could drop MIPS 32bit too.

However if we decide to implement pushing build snapshots from Travis to automate what @bleege is currently doing by hand we will need to somehow build all the ABIs on Travis. Basically https://github.com/mapbox/mapbox-gl-native/issues/1672

jfirebaugh:
Agreed, let's avoid expanding the matrix if possible.

If we need to build all ABIs for snapshots we can try to figure out a way to do that only for certain builds.

ljbade:
@jfirebaugh: yes I think that would be the easiest way. We can have something in the commit like [snapshot] that triggers a script that runs `make android-all` before uploading the results.

PR #2249, opened 2015-09-03
tidy up Android packaging & remove 64-bit ABI's

incanus:
Refs #1956. We should ship `0.1.0` with just 32-bit ABI's until we get a better handle on: 

1. Benefits of the 64-bit hassle on Android, particularly with NDK. 
1. How to solve #1956. 
1. How to best test all shipping ABI's (refs #1683). 

/cc @bleege @ljbade

### DIFF
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -315,6 +315,10 @@ When you plug your device in and then run or debug the Android project you will 
 
 If your device does not show up,  you have not set it up properly. Double check the [Google documentation](http://developer.android.com/tools/device.html).
 
+### Distribution
+
+Use the `Makefile` target `make apackage` in order to build JNI libraries for all supported ABI's for eventual distribution of the whole package. 
+
 ### Target
 
 Devices running Android 4.0+ (API level 14+).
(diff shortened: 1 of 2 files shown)


incanus:
BTW I don't think the current Travis matrix of Android ABI's is a bad idea; if tests pass now, let's keep them as a baseline and iterate from there, expanding their coverage (and ideally integration tests to ensure things stand up in 64-bit).

incanus:
Lemme know if this looks good @bleege and/or @ljbade and I'll get this merged.

bleege:
Kicking the tires on it now...

ljbade:
Looks perfect.

bleege:
Is there a reason we don't have ARM V6?  It doesn't look we ever did, but just wondering since there is a V5 and V7.

Other than that this looks good to go.  :+1:

incanus:
Rebased & force-pushed after tests passed originally.

ljbade:
Basically Android skipped ARM v6. The first few generations of Android devices all had ARM v5 based CPUs (basically no FPU), since then all recent Android devices now use ARM v7 (FPU, NEON), and this years devices use ARM v8 aka ARM64.

bleege:
@ljbade Good to know!  Thanks for the info.

incanus:
:+1:

PR #2251, opened 2015-09-03
Updating Compass Imagery

bleege:
For #2182


bleege:
All checks passed, except the Android builds in Travis because it can't find the Design Support library.  I'm not concerned in this particular case because this commit is about swapping image files.

screen shot 2015-09-03 at 2 43 41 pm

Issue #1728, opened 2015-06-16
add shape annotation reordering API


incanus:
Punted from #1655. Should be as simple as updating `AnnotationManager::orderedShapeAnnotations`, which affects the order of shape layers returned from `getTile()`.

akbsteam:
is there a timeline for this feature?

1ec5:
Note that this issue tracks an API for arbitrarily reordering non-view-backed point annotations. We don’t currently have a timeline for this feature, but pull requests are more than welcome.

View-backed annotations automatically come to the front as of mapbox/mapbox-gl-native#5264 (iOS) and mapbox/mapbox-gl-native#5294 (Android). It’s also possible to arbitrarily reorder view-backed annotations on iOS: see https://github.com/mapbox/mapbox-gl-native/pull/5287#issue-159247473 for an example. mapbox/mapbox-gl-native#5238 tracks an API for arbitrarily reordering view-backed annotations on Android.

Meanwhile, in the coming Android SDK v4.2.0 (currently in beta) and iOS SDK v3.4.0 (in alpha), the runtime styling API offers an alternative to annotations. You can arbitrarily order a style layer above or below any other style layer, even beneath parts of the base map like labels.

Hopefully the improvements in view-backed annotations and runtime styling meet your needs. If not, we’d love to hear about your specific use cases, to inform the improvements we make going forward.

incanus:
No, this issue was originally for polygon annotation reordering on the fly.

Xgamefactory:
these are simple and base features that every map sdks have like google,apple mapkit,heremaps event mapbox ios sdk.

PR #2252, opened 2015-09-03
refs #2039: add passthroughs for adding/removing overlays

incanus:
Per #2039, this just adds passthroughs on the methods people expect to pass `id <MGLOverlay>` objects to, passing them to the core annotation(s) add/remove routines. When we flesh out shape ordering in #1728, we can later add the insert/reorder Cocoa routines that are more than mere passthroughs. 

Keep in mind that I am still `assert()` type checking here for environments which don't support Objective-C generics (given that our macro there silently adds backwards compatibility). 

/cc @1ec5 @friedbunny

### DIFF
--- a/platform/ios/MGLMapView.mm
+++ b/platform/ios/MGLMapView.mm
@@ -1910,7 +1910,7 @@ CLLocationCoordinate2D MGLLocationCoordinate2DFromLatLng(mbgl::LatLng latLng)
 
     for (id <MGLAnnotation> annotation in annotations)
     {
-        assert([annotation conformsToProtocol:@protocol(MGLAnnotation)]);
+        NSAssert([annotation conformsToProtocol:@protocol(MGLAnnotation)], @"annotation should conform to MGLAnnotation");
 
         if ([annotation isKindOfClass:[MGLMultiPoint class]])
         {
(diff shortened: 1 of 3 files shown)


platform/ios/MGLMapView.mm
```diff
@@ -2098,6 +2098,36 @@ - (void)removeAnnotations:(NS_ARRAY_OF(id <MGLAnnotation>) *)annotations
...
+- (void)removeOverlay:(id <MGLOverlay>)overlay
+{
+    [self removeOverlays:@[ overlay ]];
+}
+
+- (void)removeOverlays:(NS_ARRAY_OF(id <MGLOverlay>) *)overlays
+{
+    for (id <MGLOverlay> overlay in overlays)
+    {
+        assert([overlay conformsToProtocol:@protocol(MGLOverlay)]);
```
1ec5: This should be an `NSAssert()` to work with Objective-C exception breakpoints (or breakpoints on `-[NSException raise]`).

incanus: Ah, copied from the passed-through-to methods. Will adjust all.

johndpope:
as per the comments in this file https://github.com/andreacremaschi/GEOSwift/blob/d4366af7c23e93dc8f585c5a19931b70da7d5da8/GEOSwift/MapboxGL/MapboxGLAnnotations.swift

the actual delegate method never gets called.

 func intersectsOverlayBounds(overlayBounds: MGLCoordinateBounds) -> Bool {
        print("intersectsOverlayBounds")
        return true
    }

friedbunny:
@johndpope Please go ahead and open a new issue.

PR #2255, opened 2015-09-04
plumb map notifications through JNI to Android

incanus:
Passes through `mbgl::MapChange` notification types into native Java. 

Two review issues: 

1. I _think_ that I am using Java `enum` right here — backing with an `int value`, but since unable to instantiate in the style of classes, using a `switch` later on to deduce the raw `int` passed and map to the `enum` case. This seems verbose and redundant (i.e. specifying the raw values `-1` to `13` two times). 

1. Since we already made use of pure Java `updateMap()` in places, I added a Java-only case `MapChangeNullChange(-1)` for these cases. However I'm confused why/if we need to notify at all in these cases since we're only modifying map properties such as tracking state, compass visibility, and the like and not actually modifying fetch/parse/render properties. 

Unblocks https://github.com/mapbox/mapbox-gl-native/issues/894#issuecomment-137590751. 

/cc @bleege @ljbade

### DIFF
--- a/android/java/MapboxGLAndroidSDK/src/main/java/com/mapbox/mapboxgl/views/MapView.java
+++ b/android/java/MapboxGLAndroidSDK/src/main/java/com/mapbox/mapboxgl/views/MapView.java
@@ -160,6 +160,67 @@ public class MapView extends FrameLayout implements LocationListener {
     // Used to manage Event Listeners
     private ArrayList<OnMapChangedListener> mOnMapChangedListener;
 
+    private enum MapChange {
+        MapChangeNullChange(-1),
+        MapChangeRegionWillChange(0),
+        MapChangeRegionWillChangeAnimated(1),
+        MapChangeRegionIsChanging(2),
+        MapChangeRegionDidChange(3),
+        MapChangeRegionDidChangeAnimated(4),
+        MapChangeWillStartLoadingMap(5),
+        MapChangeDidFinishLoadingMap(6),
+        MapChangeDidFailLoadingMap(7),
+        MapChangeWillStartRenderingFrame(8),
+        MapChangeDidFinishRenderingFrame(9),
+        MapChangeDidFinishRenderingFrameFullyRendered(10),
+        MapChangeWillStartRenderingMap(11),
+        MapChangeDidFinishRenderingMap(12),
+        MapChangeDidFinishRenderingMapFullyRendered(13);
+
+        private int value;
+
+        private MapChange(int value) {
+            this.value = value;
+        }
+
+        public static MapChange fromInteger(int value) {
+            switch (value) {
+                case -1:
+                    return MapChange.MapChangeNullChange;
+                case 0:
+                    return MapChange.MapChangeRegionWillChange;
+                case 1:
+                    return MapChange.MapChangeRegionWillChangeAnimated;
+                case 2:
+                    return MapChange.MapChangeRegionIsChanging;
+                case 3:
+                    return MapChange.MapChangeRegionDidChange;
+                case 4:
+                    return MapChange.MapChangeRegionDidChangeAnimated;
+                case 5:
+                    return MapChange.MapChangeWillStartLoadingMap;
+                case 6:
+                    return MapChange.MapChangeDidFinishLoadingMap;
+                case 7:
+                    return MapChange.MapChangeDidFailLoadingMap;
+                case 8:
+                    return MapChange.MapChangeWillStartRenderingFrame;
+                case 9:
+                    return MapChange.MapChangeDidFinishRenderingFrame;
+                case 10:
+                    return MapChange.MapChangeDidFinishRenderingFrameFullyRendered;
+                case 11:
+                    return MapChange.MapChangeWillStartRenderingMap;
+                case 12:
+                    return MapChange.MapChangeDidFinishRenderingMap;
+                case 13:
+                    return MapChange.MapChangeDidFinishRenderingMapFullyRendered;
+                default:
+                    return null;
+            }
+        }
+    }
+
     public interface OnFlingListener {
         void onFling();
     }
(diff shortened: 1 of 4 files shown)


incanus:
You can easily test this with a patch such as the following: 

```diff
diff --git a/android/java/MapboxGLAndroidSDK/src/main/java/com/mapbox/mapboxgl/views/MapView.java b/android/java/MapboxGLAndroidSDK/src/main/java/com/mapbox/mapboxgl/views/MapView.java
index b75aaba..9382bee 100644
--- a/android/java/MapboxGLAndroidSDK/src/main/java/com/mapbox/mapboxgl/views/MapView.java
+++ b/android/java/MapboxGLAndroidSDK/src/main/java/com/mapbox/mapboxgl/views/MapView.java
@@ -1829,6 +1829,9 @@ public class MapView extends FrameLayout implements LocationListener {
 
     // Updates the UI to match the current map's position
     private void updateMap(MapChange change) {
+
+        Log.d(TAG, "updateMap: " + Integer.toString(change.value));
+
         // Using direct access to mIsCompassEnabled instead of isCompassEnabled() for
         // small performance boost as this method is called rapidly.
         if (mIsCompassEnabled) {
```

Then see the following in the logs as you interact with the map: 

```
09-03 17:17:51.202  31972-31972/com.mapbox.mapboxgl.testapp D/MapView﹕ updateMap: 8
09-03 17:17:51.219  31972-31972/com.mapbox.mapboxgl.testapp D/MapView﹕ updateMap: 9
09-03 17:17:51.219  31972-31972/com.mapbox.mapboxgl.testapp D/MapView﹕ updateMap: 8
09-03 17:17:51.219  31972-31972/com.mapbox.mapboxgl.testapp D/MapView﹕ updateMap: 10
09-03 17:17:51.219  31972-31972/com.mapbox.mapboxgl.testapp D/MapView﹕ updateMap: 13
09-03 17:17:51.239  31972-31972/com.mapbox.mapboxgl.testapp D/MapView﹕ updateMap: 8
09-03 17:17:51.254  31972-31972/com.mapbox.mapboxgl.testapp D/MapView﹕ updateMap: 10
09-03 17:17:51.254  31972-31972/com.mapbox.mapboxgl.testapp D/MapView﹕ updateMap: 8
09-03 17:17:51.269  31972-31972/com.mapbox.mapboxgl.testapp D/MapView﹕ updateMap: 10
```

ljbade:
The switch is pretty much the accepted way to `int -> enum`: http://stackoverflow.com/questions/5878952/cast-int-to-enum-in-java

I think the idea with `updateMap()` on Java side was related to triggering the UI to fetch updates GPS coordinates and update the location of the GPS marker dot (it made sense at the time to use same code for map pan/zoom updates).

ljbade:
Also looking at Google Maps: https://developers.google.com/android/reference/com/google/android/gms/maps/GoogleMap

They split each type of map update into separate listeners? Should we do the same?

incanus:
Ah, that's a nice reference link; better than the stuff I was finding. I'll tune this up to move the switch into the class itself.

android/java/MapboxGLAndroidSDK/src/main/java/com/mapbox/mapboxgl/views/MapView.java
```diff
@@ -160,6 +160,30 @@
...
+        MapChangeDidFinishRenderingMap(12),
+        MapChangeDidFinishRenderingMapFullyRendered(13);
+
+        private int value;
+
+        private MapChange(int value) {
+            this.value = value;
+        }
+    }
+
```
bleege: This will definitely work, but if we don't need to maintain value for any particular reason it can definitely be safely removed.

```java
private enum MapChange {
MapChangeNullChange,
MapChangeRegionWillChange,
MapChangeRegionWillChangeAnimated,
...
}

// Comparison works this way then
MapChange mcEvent;
switch (mcEvent) {
   case MapChange.MapChangeNullChange:
      break;
   case MapChange.MapChangeRegionWillChange:
     break;
   case MapChange.MapChangeRegionWillChangeAnimated
     break;

....
   default:
     // Shouldn't get here
}

```

It can lead to a bit more verbose switch statements, but also safer ones as it doesn't require values to be copied and stored multiple places.

android/java/MapboxGLAndroidSDK/src/main/java/com/mapbox/mapboxgl/views/MapView.java
```diff
@@ -1754,7 +1828,7 @@ private void rotateImageView(ImageView imageView, float angle) {
     }
 
     // Updates the UI to match the current map's position
-    private void updateMap() {
+    private void updateMap(MapChange change) {
```
bleege: :+1:

android/java/MapboxGLAndroidSDK/src/main/java/com/mapbox/mapboxgl/views/MapView.java
incanus: >  if we don't need to maintain value for any particular reason it can definitely be safely removed.

I want to maintain equivalence to [this `enum`](https://github.com/mapbox/mapbox-gl-native/blob/1aee3c6c521b502b3c37b87ed28a3140f68e87b2/include/mbgl/map/view.hpp#L14-L29) on the native side.

bleege: Ah..  that makes sense.  :+1:

incanus: https://github.com/mapbox/mapbox-gl-native/commit/472846ef58ec702cf283c39eb0e803f3d618526c

PR #2257, opened 2015-09-04
Update LICENSE.MD

### LINKED ISSUE #846: Improve LICENSE.md
Currently the android project depends on several 3rd party libs, including one that I just keep a copy of the source code in the repo.

Also native (and ios?) uses many 3rd party libs under various licences as well as some libs that we keep as source files.

We need to check what we need to do w.r.t to the different licenses.

Also we should make sure we are not using any GPL libs.

/cc @jfirebaugh @michaelsteffen

ljbade:
Closes #846


incanus:
Reviewing.

PR #2258, opened 2015-09-04
map.render return only pixels

bsudekum:
Currently we're returning an object from the `map.render` function:

```json
{
  "height": 60,
  "width": 60,
  "pixels": "<SlowBuffer 34 43 45 ...>"
}
```

This is overkill since the height and width are already know _because they are necessary to render an image in the first place_. Although this is a small change, the system needs to allocate memory to do these processes which can add up. This change would no longer return an object but rather just the raw image buffer. 

/cc @mikemorris @springmeyer @jfirebaugh

### DIFF
--- a/platform/node/src/node_map.cpp
+++ b/platform/node/src/node_map.cpp
@@ -258,9 +258,6 @@ void NodeMap::renderFinished() {
 
         cb->Call(1, argv);
     } else if (img) {
-        auto result = NanNew<v8::Object>();
-        result->Set(NanNew("width"), NanNew(img->width));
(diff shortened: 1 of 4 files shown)


jfirebaugh:
:+1: -- I was thinking maybe we should do this, agree with your reasoning.

mikemorris:
:+1: to this change, but we should use the actual `width` and `height` variables passed into the render call rather than hardcoding `512` in the tests.

bsudekum:
Going to re-roll this PR.

PR #2262, opened 2015-09-04
Return raw image buffer instead of an object

bsudekum:
_continuation of https://github.com/mapbox/mapbox-gl-native/pull/2258_ 

Currently we're returning an object from the `map.render` function:

```json
{
  "height": 60,
  "width": 60,
  "pixels": "<SlowBuffer 34 43 45 ...>"
}
```

This is overkill since the height and width are already know _because they are necessary to render an image in the first place_. Although this is a small change, the system needs to allocate memory to do these processes which can add up. This change would no longer return an object but rather just the raw image buffer. 

/cc @mikemorris @springmeyer @jfirebaugh

### DIFF
--- a/platform/node/test/js/map.test.js
+++ b/platform/node/test/js/map.test.js
@@ -231,14 +231,12 @@ test('Map', function(t) {
         t.test('returns an image', function(t) {
             var map = new mbgl.Map(options);
             map.load(style);
-            map.render({}, function(err, data) {
+            map.render({}, function(err, pixels) {
                 t.error(err);
 
                 map.release();
 
-                t.ok(data.pixels);
-                t.equal(data.width, 512);
-                t.equal(data.height, 512);
+                t.ok(pixels);
                 t.end();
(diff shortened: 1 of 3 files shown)


platform/node/test/js/map.test.js
```diff
@@ -231,14 +231,12 @@ test('Map', function(t) {
...
-            map.render({}, function(err, data) {
+            map.render({}, function(err, pixels) {
                 t.error(err);
 
                 map.release();
 
-                t.ok(data.pixels);
-                t.equal(data.width, 512);
-                t.equal(data.height, 512);
+                t.ok(pixels);
```
jfirebaugh: Can you add a `t.ok(pixels instanceof Buffer)` assertion here? I've been meaning to add that.

jfirebaugh: Also `t.equal(pixels.length, 512 * 512 * 4)`.

PR #2265, opened 2015-09-04
Fix iosapp location ask

### LINKED ISSUE #1930: Location Services prompt disappears too soon
A fresh install of iosapp on iOS 9 displays a Location Services prompt for “Always” permissions, but this prompt disappears after a second without any user intervention. Then if you hit the user location tracking button, a “When in Use” prompt appears, so iosapp has no access to background location updates. If at some point you revoke iosapp’s “When in Use” permission, pressing the user location tracking button does nothing.

Unclear how this affects #1909.

/cc @incanus @friedbunny

friedbunny:
The iosapp demo app was asking for location permission via `showsUserLocation = YES` in `viewDidLoad`, then immediately having `restoreState:` set `showsUserLocation = NO` because that's what the default is.

This PR lets `restoreState:` do the setting of `showsUserLocation` and then only asks for permissions if the value is `NO` — after a delay to allow the map to load.

Fixes #1930.

/cc @1ec5 @incanus

### DIFF
--- a/ios/app/MBXViewController.mm
+++ b/ios/app/MBXViewController.mm
@@ -49,7 +49,6 @@ mbgl::Settings_NSUserDefaults *settings = nullptr;
 
     self.mapView = [[MGLMapView alloc] initWithFrame:self.view.bounds];
     self.mapView.autoresizingMask = UIViewAutoresizingFlexibleWidth | UIViewAutoresizingFlexibleHeight;
-    self.mapView.showsUserLocation = YES;
     self.mapView.delegate = self;
     [self.view addSubview:self.mapView];
 
@@ -76,6 +75,13 @@ mbgl::Settings_NSUserDefaults *settings = nullptr;
 
     settings = new mbgl::Settings_NSUserDefaults();
     [self restoreState:nil];
+
+    if ( ! settings->showsUserLocation)
+    {
+        dispatch_after(dispatch_time(DISPATCH_TIME_NOW, 2 * NSEC_PER_SEC), dispatch_get_main_queue(), ^{
+            self.mapView.showsUserLocation = YES;
+        });
+    }
 }
 
 - (void)saveState:(__unused NSNotification *)notification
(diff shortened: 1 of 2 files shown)


friedbunny:
RE: https://github.com/mapbox/mapbox-gl-native/issues/1930#issuecomment-137858609

> I'm not generally into timers, but in the interest of simplicity, maybe just set `showsUserLocation = YES` a second after `viewDidAppear:`? 

Went with two seconds at the end of `viewDidLoad:`, only on the first launch (or any launch after that doesn't have permission, I suppose). Whatcha' think?

incanus:
I think this is fine. However if we save & restore `userTrackingMode`, this will implicitly show the user dot, right?

1ec5:
> However if we save & restore `userTrackingMode`, this will implicitly show the user dot, right? 

The user might want to see the user dot without following it.

incanus:
My point is that restoring tracking mode at launch will have the same affect as restoring showing the user dot — still prompt the user without delay when restoring state.

friedbunny:
If the saved user tracking mode was `none` and we didn't explicitly set `mapView.showsUserLocation`, the dot would not appear initially. I kept `settings->showsUserLocation` because it's useful in determining if location is authorized.

The delayed `mapView.showsUserLocation` should only be triggered on the first launch, because we turn it on when location permissions are given and it is never turned off at any point after. (Disabling user tracking when the dot is shown doesn't disable `.showsUserLocation`.)

maciekish:
This commit breaks master for me. Xcode says "Use of undeclared identifier 'settings' here https://github.com/mapbox/mapbox-gl-native/pull/2265/files#diff-33b34b5f099a16158c8462652a0c3688R79

friedbunny:
This was broken subsequently in https://github.com/mapbox/mapbox-gl-native/commit/0a172a21fdc2a87473560fd7d45f4d495d95de91#diff-33b34b5f099a16158c8462652a0c3688 when we moved directly to using `NSUserDefaults` instead of an intermediary. I'll get a fix in.

maciekish:
I think you may want to prefix your defaults or use `initWithSuiteName:` or something to avid clashes with defaults names.

friedbunny:
>I think you may want to prefix your defaults or use initWithSuiteName: or something to avid clashes with defaults names.

Sandboxing means that we only have to contend with certain Apple `NSUserDefaults` (which tend to be namespaced themselves). A real app may need to worry about this, but our demo app likely does not.

PR #2277, opened 2015-09-08
Use GL types for GL objects

brunoabinader:
Based on [a public discussion](http://stackoverflow.com/questions/8932912/whats-the-advantage-of-using-gluint-instead-of-unsigned-int) on this subject, I believe we have reasons to prefer using GL types for GL-related objects instead of stdlib types.

/cc @mapbox/gl

### DIFF
--- a/src/mbgl/geometry/vao.cpp
+++ b/src/mbgl/geometry/vao.cpp
@@ -69,7 +69,7 @@ void VertexArrayObject::bindVertexArrayObject() {
 }
 
 void VertexArrayObject::verifyBinding(Shader &shader, GLuint vertexBuffer, GLuint elementsBuffer,
-                                      char *offset) {
+                                      GLbyte *offset) {
     if (bound_shader != shader.getID()) {
         throw std::runtime_error(std::string("trying to rebind VAO to another shader from " +
                                              util::toString(bound_shader) + "(" + bound_shader_name + ") to " +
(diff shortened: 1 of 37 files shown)


incanus:
This sounds good in theory to me, but let's wait for @kkaefer to ring in on this when he's back.

jfirebaugh:
:+1: 

Aren't there are a lot more than this though, e.g. all of the `*Shader` classes?

brunoabinader:
> Aren't there are a lot more than this though, e.g. all of the *Shader classes?

Indeed, thanks for the remind! Updating the patch now.

brunoabinader:
I am particularly happy with https://github.com/mapbox/mapbox-gl-native/commit/00bc524f174def177b4be281025eddb06224bcf4 and https://github.com/mapbox/mapbox-gl-native/commit/71526dfa113166888034a6557e22ad2828cf31d0 as the former removes a lot of duplicated code while the latter enforces the next Shader types to come with the same API.

brunoabinader:
This is a bit more complicated than I've imagined - I am sticking this PR for the first two patches (the one that makes `bind()` a pure virtual + moving `a_pos` to `Shader` base class) and letting the rest for further discussion (i.e. when @kkaefer comes back from vacations).

kkaefer:
Yeah, let's convert all types to the corresponding `Gl*` types.

PR #2280, opened 2015-09-08
Notify MGLMapView delegate about single and long taps on the map

maciekish:
Adds two delegate methods for notifying the delegate about single- and long taps on the map.

    - (void)mapView:(MGLMapView *)mapView singleTapOnMapAtCoordinate:(CLLocationCoordinate2D)coordinate;

    - (void)mapView:(MGLMapView *)mapView longPressOnMapAtCoordinate:(CLLocationCoordinate2D)coordinate;

### DIFF
--- a/include/mbgl/ios/MGLMapView.h
+++ b/include/mbgl/ios/MGLMapView.h
@@ -601,6 +601,26 @@ IB_DESIGNABLE
 *  @param annotation The annotation that was deselected. */
 - (void)mapView:(MGLMapView *)mapView didDeselectAnnotation:(id <MGLAnnotation>)annotation;
 
+#pragma mark - Map Interaction
+
+/** @name Map Interaction */
+
+/* Tells the delegate that the map was tapped outside of any annotations.
+ *
+ *  You can use this method to cancel overlays, remove dropped pins etc.
+ *
+ *  @param mapView The map view that was tapped.
+ *  @param coordinate The coordinate that was tapped. */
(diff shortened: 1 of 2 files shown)


Zakay:
Will these features find them selfes in Android version too? Or do they already exists? Would be great to see a feature/support matrix between android/ios. Sure mapbox gl's core is on both but the interaction features might differ between platforms.

friedbunny:
Thanks for the pull request, @maciekish. We're focusing on releasing [`v2.1.0`](https://github.com/mapbox/mapbox-gl-native/milestones/ios-v2.1.0) at the moment, so it may take a bit for us to get to this — thanks, too, for your patience.

ljbade:
@Zakay There is a ticket for adding more touch event hooks on Android, but I can't find it right now.

bsudekum:
@incanus @1ec5 @friedbunny could we re-take a look at this PR? Seems like a very good addition.

1ec5:
Thanks for the PR, @maciekish, but I’m leaning towards :-1: on this change. I think it would be more appropriate for the view controller to manage these gesture recognizers itself, [as MBXViewController does](https://github.com/mapbox/mapbox-gl-native/blob/6824a24acd2f3450a402f1b765d01f841bcd5d6c/ios/app/MBXViewController.mm#L301). Folding these recognizers into MGLMapView hides properties like `minimumPressDuration` that would should be configurable, and it would open the door to MGLMapView providing coordinate-savvy versions of all the gesture recognizers, including ones that an application may never need. (The view controller might want full access to the UIEvent in any case.)

The situation on Android may be very different if it’s more difficult to stand up a gesture recognizer that’s consistent with the MapView’s built-in ones.

PR #2281, opened 2015-09-08
Fix build, only ask for location permission in iosapp on user's instigation

friedbunny:
Previously in #2265 we would ask for location permission at app startup if `!settings->showsUserLocation`. #2193 0a172a21fdc2a87473560fd7d45f4d495d95de91 changed the way we use `NSUserDefaults` and mostly removed the `settings` object, which broke #2265.

Rather than fix our pestering location permissions ask at iosapp startup, this commit now only asks for location permissions when the user hits the locate-me button.

Once a user grants permission, the user dot appears because a `userTrackingMode` is set, `showsUserLocation` is permanently set to `YES` in `NSUserDefaults` and is restored at launch.

/cc @incanus @1ec5

### DIFF
--- a/ios/app/MBXViewController.mm
+++ b/ios/app/MBXViewController.mm
@@ -83,13 +83,6 @@ static NSUInteger const kStyleVersion = 8;
                                                                              action:@selector(locateUser)];
 
     [self restoreState:nil];
-
-    if ( ! settings->showsUserLocation)
-    {
-        dispatch_after(dispatch_time(DISPATCH_TIME_NOW, 2 * NSEC_PER_SEC), dispatch_get_main_queue(), ^{
-            self.mapView.showsUserLocation = YES;
-        });
-    }
 }
 
 - (void)saveState:(__unused NSNotification *)notification


1ec5:
MBXViewController [still sets `showsUserLocation`](https://github.com/mapbox/mapbox-gl-native/blob/9a55e9702915493ac34f58469674eaf9e77c019b/ios/app/MBXViewController.mm#L119). Does that need to happen on a delay still?

friedbunny:
> MBXViewController [still sets `showsUserLocation`](https://github.com/mapbox/mapbox-gl-native/blob/9a55e9702915493ac34f58469674eaf9e77c019b/ios/app/MBXViewController.mm#L119). Does that need to happen on a delay still?

Nope, because `restoreState:` will either:

1. Set `showsUserLocation` to `NO`, [as is the `NSUserDefault`](https://github.com/mapbox/mapbox-gl-native/blob/9a55e9702915493ac34f58469674eaf9e77c019b/ios/app/MBXViewController.mm#L35), in which case we don't ask for or need permissions. This is the first launch behavior.

2. Set `showsUserLocation` to `YES`, in which case we already have location permission because the user explicitly hit the locate-me button and said, "Yeah, track my location (and put the dot on the map)!"

friedbunny:
Incidentally, once we have location permissions, we assume that we always have location permissions. The user could disable them, in which case we fail silently and don't ask the user to reenable.

We could ask using `mapView:didFailToLocateUserWithError:`, [like I do in Treble...](https://github.com/friedbunny/treble/blob/456f1a16e6273d4e228e602a382bdadf45a0687c/treble/TRBLMapboxMapView.m#L157-L195)

incanus:
> Incidentally, once we have location permissions, we assume that we always have location permissions. The user could disable them, in which case we fail silently and don't ask the user to reenable.

Considering this isn't even an internally user-facing app, I think it's overkill. The only folks using this are those developing on GL.

PR #2286, opened 2015-09-08
Prefix user defaults in iosapp

1ec5:
Per https://github.com/mapbox/mapbox-gl-native/pull/2265#issuecomment-138685246, we technically don’t need to prefix these user defaults keys. However, the demo app should practice good hygiene since we expect developers to copy from it.

/cc @friedbunny


incanus:
Keep in mind we don't have a migration here (which is fine and just, just noting).

friedbunny:
pigpen

Issue #1688, opened 2015-06-04
Prevent flickering when adding/removing annotations


kkaefer:
When adding a new annotation, tiles that contain this annotation are flickering because we are removing the tile, and parsing the new tile may still take a frame or two. Instead of removing the tile, we should reparse the tile and update the bucket as required.

l2comdev:
I'm still noticing this for both adding & removing annotations. (iOS)

incanus:
@jfirebaugh Would you have time to look into this sometime, at least to scope the work required?

incanus:
This is pretty satisfactory right now. See https://github.com/mapbox/mapbox-gl-native/issues/1798#issuecomment-137858050.

parrots:
Re comment on #1798 addressing flickering, made a PR to Justin's sample project to show the flickering is still there. See here: https://github.com/incanus/UISyncedMap/pull/1 which just zooms out the map to show the entire user route.

Ideal outcomes, as far as I'd see, besides just "make it faster" which could always have a visible flicker on old hardware or with enough complex data:
- Find a way to not remove the old drawn layer until the new version can be swapped in so there is 0.1 seconds of stale data vs 0.1 seconds of empty space.
- Only redraw the annotations that were modified.

mourner:
> Find a way to not remove the old drawn layer until the new version can be swapped in so there is 0.1 seconds of stale data vs 0.1 seconds of empty space.

This is what GL JS does currently, which is good enough for use cases like mouse-driven polylines etc.

pkmital:
We're still experiencing issues with flickering for our use case.  We are drawing a polyline of about 100 points, and are constantly removing it and redrawing it with each update, though require updating the polyline so often as we are changing only the last few points of the line.

maciekish:
We're also experiencing this issue in 294b307cb3f470b1cf5fb94633c9082dc0bfd719. When the user pans our map we reload search results and the experience is horrible. The idea suggested by @mourner sounds like a very good one. cc @zakay

incanus:
Going to have to move this from the current iOS milestone as it's non-blocking.

futureshape:
Having the same problem here, my map markers show a different image based on live data (availability of bikes and spaces on a bicycle hire scheme) and the only way to change the image is to add/remove, which causes flicker when more than a couple of markers are visible.

kkaefer:
Depends on #2103 and #2481

kkaefer:
[...]

PR #2289, opened 2015-09-09
fixes #1675: fix initial annotation markers

### LINKED ISSUE #1675: No annotations at initial zoom level
If you follow the “[First steps with Mapbox GL for iOS](https://www.mapbox.com/guides/first-steps-gl-ios/#adding-a-marker)” guide, you wind up initializing an `MGLMapView`, centering it on the U.S. President’s front yard, and placing an annotation there, all within `-[ViewController viewDidLoad]`. But if you do all that in `-viewDidLoad`, the annotations fail to display at the initial zoom level. They appear when you zoom in or out to a different zoom level and disappear when you reenter the initial zoom level.

At launch, the map view happens to show z15 tiles:

initial

Then zoom in, right to the cusp of z16:

in

And zoom in just a little farther to z17:

inner

Or zoom out, right to the cusp of z14:

out

And a little further out to 14:

outer

The upshot is that our guide does not work as designed. We didn’t catch this issue in development because the iOS demo app only adds annotations in response to an action sheet button, not within `-viewDidLoad`.

Reproduces in ios-v0.3.1 as well as master, regardless of style.

/cc @incanus

incanus:
This does two things: 

1. It abstracts the pattern of checking if there are any stale annotation tiles (moving them to the caller in the process), then passing those back to the update routine to invalidate said annotation tiles. 

1. It uses that routine additionally in response to the sprite store loading, which necessarily loads after the sprite, which is after the style, which is the only way that either runtime-provided marker imagery or the default style sprite marker imagery will work. It extends the `Observer` pattern already used in the style and the map context for decoupled communications.

### DIFF
--- a/src/mbgl/map/annotation.hpp
+++ b/src/mbgl/map/annotation.hpp
@@ -52,6 +52,7 @@ public:
     ~AnnotationManager();
 
     void markStaleTiles(std::unordered_set<TileID, TileID::Hash>);
+    size_t getStaleTileCount() const { return staleTiles.size(); }
     std::unordered_set<TileID, TileID::Hash> resetStaleTiles();
 
     void setDefaultPointAnnotationSymbol(const std::string& symbol);
--- a/src/mbgl/map/map_context.cpp
+++ b/src/mbgl/map/map_context.cpp
@@ -142,9 +142,11 @@ void MapContext::loadStyleJSON(const std::string& json, const std::string& base)
 
     updateFlags |= Update::DefaultTransition | Update::Classes | Update::Zoom;
     asyncUpdate->send();
+}
 
-    auto staleTiles = data.getAnnotationManager()->resetStaleTiles();
-    if (!staleTiles.empty()) {
+void MapContext::updateAnnotationTilesIfNeeded() {
+    if (data.getAnnotationManager()->getStaleTileCount()) {
+        auto staleTiles = data.getAnnotationManager()->resetStaleTiles();
         updateAnnotationTiles(staleTiles);
     }
 }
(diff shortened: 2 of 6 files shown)


incanus:
Here's a simple diff to check the behavior: 

```diff
diff --git a/ios/app/MBXViewController.mm b/ios/app/MBXViewController.mm
index d94e41b..e7548c6 100644
--- a/ios/app/MBXViewController.mm
+++ b/ios/app/MBXViewController.mm
@@ -82,7 +82,14 @@ static NSUInteger const kStyleVersion = 8;
                                                                              target:self
                                                                              action:@selector(locateUser)];
 
-    [self restoreState:nil];
+//    [self restoreState:nil];
+
+    self.mapView.debugActive = YES;
+
+    MGLPointAnnotation *point = [MGLPointAnnotation new];
+    point.coordinate = CLLocationCoordinate2DMake(45, -123);
+    [self.mapView addAnnotation:point];
+    [self.mapView setCenterCoordinate:point.coordinate zoomLevel:15 animated:NO];
 }
 
 - (void)saveState:(__unused NSNotification *)notification
@@ -354,6 +361,8 @@ static NSUInteger const kStyleVersion = 8;
 
 - (MGLAnnotationImage *)mapView:(MGLMapView * __nonnull)mapView imageForAnnotation:(id <MGLAnnotation> __nonnull)annotation
 {
+    if (annotation.title == nil) return [MGLAnnotationImage annotationImageWithImage:[UIImage imageNamed:@"Icon.png"] reuseIdentifier:@"foo"];
+
     NSString *title = [(MGLPointAnnotation *)annotation title];
     NSString *lastTwoCharacters = [title substringFromIndex:title.length - 2];
 ```

incanus:
I should add that this `updateAnnotationTiles()` wholesale business is ripe for replacement as part of #1688, which should move to incremental parsing and rendering for annotations.

incanus:
Mind giving this a pass @friedbunny @1ec5 in light of the First Steps guide? Working good for me.

friedbunny:
Lookin' at this...

friedbunny:
Works well, polylines and markers both show up immediately when drawn in `viewDidLoad:`. :+1:

PR #2291, opened 2015-09-09
Fix MapView.addOnMapChangedListener and Makefile clean.

mayardb:
Makefile clean task when cleaning MapboxGLAndroidSDKTestApp twice instead of cleaning MapboxGLAndroidSDKTestApp and MapboxGLAndroidSDK

MapView.MapChange was a private enum while OnMapChangedListener which use it in onMapChanged callback was public. So change the enum visibility to public too to be able to use the listener from outside of the MapView.


jfirebaugh:
Thank you! Committed as 765bae4c5cb0c55517bf0b5edd79066b41316bb0 and 8d322c65728ffe0d4f69fcd3b73b007a53eb505a.

Issue #2244, opened 2015-09-03
Use TextureView


ljbade:
Now that we target API level 15 and above we should switch to [`TextureView`](http://developer.android.com/reference/android/view/TextureView.html)

The advantage is that we can treat a `MapView` the same as any other `View`. This will allow advanced layouts and may avoid bugs such as https://github.com/mapbox/mapbox-gl-native/issues/2018 and https://github.com/mapbox/mapbox-gl-native/issues/1894

@bleege Do we want to do this in 0.1.0 or 0.2.0?

ljbade:
Started work on this.

tobrunvannuland:
Nice, looking forward seeing this working!

Hopefully this will not be an issue:
> The main strike against TextureView is the performance of the composition step. With SurfaceView,  the content is written to a separate layer that SurfaceFlinger composites, ideally with an overlay. With TextureView, the View composition is always performed with GLES, and updates to its contents may cause other View elements to redraw as well (e.g. if they're positioned on top of the TextureView)

bleege:
> Do we want to do this in 0.1.0 or 0.2.0?

`0.2.0`.  @ljbade This sounds like a good thing, but it's too late for `0.1.0`.

ljbade:
@bleege Sure we might have to move those bugs to `0.2.0` then.

bleege:
Moving this to `0.1.0` as it solves Translucent Status bar issue in #1894 .  See https://github.com/mapbox/mapbox-gl-native/issues/1894#issuecomment-138603599 for more details.

Is there anything more to be done @ljbade or can we merge this one in as is?

mb12:
@ tobrun Were you able to reproduce the original issue without these changes?  My understanding is that this is a speculative fix for an issue which is not reproducible with latest master.

tobrun:
@mb12 will do a small test before it gets merged

incanus:
Per chat with @bleege, we are going to move this back to `0.2.0` since it's a bit of a refactor and is triggered only in an edge case. We can quickly wrap it into the next release cycle and go from there.

tobrun:
@mb12
Problem still exists on master so TextureView is a valid fix.

Nice job @ljbade :+1:

![device-2015-09-09-150700](https://cloud.githubusercontent.com/assets/2151639/9762543/96558b96-5704-11e5-91c0-1805955265c4.png)

mb12:
Great fix. This was a tough one.

[...]

PR #2293, opened 2015-09-09
Switch to TextureView

bleege:
For #2244

/cc @incanus @ljbade @tobrun


incanus:
Per voice with @bleege, this is is simple and atomic enough that we're going to roll it into `master` and `release-android-v0.1.0` and test fixes for other issues atop it.

bleege:
Merged.  Will roll a new SNAPSHOT next.

ljbade:
Another side effect of TextureView, FAB now has correct drop shadow on Samsung:
image

PR #2299, opened 2015-09-10
[node] change request semantics

### LINKED ISSUE #2287: change request semantics for node
Instead of calling a `respond` method on the request:

```
    request: function(req) {
        fs.readFile(path.join('base/path', req.url), function(err, data) {
            req.respond(err, { data: data });
        });
    },
```

It should be a normal callback:

```
    request: function(req, callback) {
        fs.readFile(path.join('base/path', req.url), function(err, data) {
            callback(err, { data: data });
        });
    },
```

cc @mikemorris

mikemorris:
Passes a second, callback argument to the `request` implementation instead
of needing to call `req.respond`

Fixes #2287

/cc @jfirebaugh

### DIFF
--- a/package.json
+++ b/package.json
@@ -18,7 +18,7 @@
     }
   ],
   "dependencies": {
-    "nan": "nodejs/nan#a71301e0f62877ca017c4ebec7fbac9f4359038e",
+    "nan": "^2.0.9",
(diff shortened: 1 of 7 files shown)


jfirebaugh:
:+1:

Issue #2186, opened 2015-08-26
Annotations don't maintain size during tilting


friedbunny:
When the map is tilted, marker annotations are scaled with the rest of the map — markers closer to the horizon are small and difficult to interact with.

We should provide an option (default: on) for annotations to maintain their normal scale during tilting.

![simulator screen shot aug 26 2015 4 47 13 pm](https://cloud.githubusercontent.com/assets/1198851/9509256/36f57c4e-4c12-11e5-86be-d6084a6f85b3.png)

/cc @ansis @incanus @1ec5

incanus:
This is especially jarring considering the wide range of sizes that can happen. 

![img_4133](https://cloud.githubusercontent.com/assets/17722/9513641/154f4da6-4c47-11e5-9e8b-31ebaa41a1e4.PNG)

ansis:
Yeah, it's not just annotations. Road and poi labels become illegible near the top. Nothing works well when the map is tilted this steep. The roads become so narrow they disappear. Steep tilt is a completely different game. Many things (rendering, data loading, styling, labeling) need to be different. I think this will be a lot of work so I'm not sure how soon we'll have it.

We can also tweak the [`altitude`](https://github.com/mapbox/mapbox-gl-native/blob/0e456940b98f53624618f4e2a42bf025ce392a9e/src/mbgl/map/transform_state.hpp#L103) parameter to make the maps less perspectivey. If altitude is higher than the map can be squished more in the y direction without making close things appear bigger.

What's the use case here? Why is the map in perspective mode and why do annotations need to be equally prominent in this case?

Does it matter that it's harder to interact? Can the user just pan the map forward to make things at the back bigger? (this will be easier with https://github.com/mapbox/mapbox-gl-native/issues/2139#issuecomment-135178162)

Does the hit detection need to be adjusted for perspective views?

mb12:
In Nokia maps, Google Maps and Apple maps,  the user can drop a pin (annotation) anywhere on the map. The application then does an object query and shows whats at that point (address etc.).  This user pin retains its original size/shape when the map is panned with perspective enabled.

incanus:
What @mb12 said — all mobile toolkits maintain size for markers. This affects hit detection as well as popup/callout placement. 

The use case is general tilt for interactive maps just like other frameworks. Once you've got tilt, users expect it to work the same way.

friedbunny:
[...]

Issue #2045, opened 2015-08-12
update iOS accuracy circle during zooms


incanus:
@friedbunny great user annotation work in https://github.com/mapbox/mapbox-gl-native/pull/2010 brought us an accuracy circle (among other things). But its size should change smoothly with map zoom instead of animated immediately after the zoom finishes. 

![anigif-1439404266](https://cloud.githubusercontent.com/assets/17722/9233036/b4df7f52-40e5-11e5-9198-7f252920a5bd.gif)

Should be able to hit this pretty easily inside of `-[MGLMapView updateUserLocationAnnotationView]`.

incanus:
Possibly helpful: `MGLCircle` by way of https://github.com/mapbox/mapbox-gl-native/issues/2167, which would mean moving at least the non-pulsing (accuracy) circle) to GL-side. Maybe we don't want to do that, but it's an option.

1ec5:
> change smoothly with map zoom instead of animated immediately after the zoom finishes

Actually, `-[MGLMapView updateUserLocationAnnotationView]` does cause the accuracy ring to resize in tandem with the map animation. The change lags behind the map animation because Core Animation applies an implicit animation to any property changes on the user dot’s layers.

1ec5:
See #3683 for an example of disabling the implicit animation.

friedbunny:
We’re good here after #6215.

PR #2300, opened 2015-09-10
User location puck for course tracking mode [squash]

friedbunny:
Squash of #2233.

- User location puck for course tracking mode
- Tilts the user location dot and puck with the map

puckanddot

The puck and dot continue to be drawn as a single `UIView` (`MGLUserLocationAnnotationView`), entirely using several stacked `CALayer` and vectors — no images allowed here. 

The pseudo-3D effects for the dot and puck while tilting are accomplished by moving the shadow offset and by vertically shifting the blue dot as pitch changes.

/cc @1ec5 @incanus @mapbox/mobile

### DIFF
--- a/include/mbgl/ios/MGLGeometry.h
+++ b/include/mbgl/ios/MGLGeometry.h
@@ -3,6 +3,7 @@
 #import "MGLTypes.h"
 
 #import <CoreLocation/CoreLocation.h>
+#import <CoreGraphics/CGBase.h>
 
 NS_ASSUME_NONNULL_BEGIN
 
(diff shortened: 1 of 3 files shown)


friedbunny:
Problems that will have to be addressed later:

- **Accuracy ring does not scale properly when tilted.** It's correct at the center of the map, but gets wildly inaccurate as you move the dot towards the top or bottom of the screen.

- **Fake 3D isn't going to work as the pitch approaches 90º.** We cap at 60º right now, though.

friedbunny:
Found a bug where a tilted off-screen user dot doesn't get reset when the pitch changes. Looking into a fix.

friedbunny:
> Found a bug where a tilted off-screen user dot doesn't get reset when the pitch changes. Looking into a fix.

Fixed in f502e90.

incanus:
Checking this out.

incanus:
> Accuracy ring does not scale properly when tilted.

Refs https://github.com/mapbox/mapbox-gl-native/issues/2186, https://github.com/mapbox/mapbox-gl-native/issues/2045

incanus:
This looks great and I think it's ready. Bummer about the accuracy circle, but that's bigger issues over in https://github.com/mapbox/mapbox-gl-native/issues/2186 / having a varying meters/pixel at different points on the screen.

PR #2301, opened 2015-09-10
Reattach user dot to map at low zoom levels

1ec5:
Fixes #2230 by increasing precision in a few key methods and folding in 214f997fbe881ab30a19756eedf34c7cda60aaad from @ansis’ `perspective-improved-gestures` branch. We may in all likelihood turn up additional places where `double` is needed, but this is safe for the release branch.

/cc @friedbunny @incanus

PR #2307, opened 2015-09-11
[wip] experiments with Android user dot accuracy & sync

incanus:
Building on our learning in https://github.com/mapbox/mapbox-gl-native/pull/2301, I attempted a few things (with no luck yet) to try to fix up #1676 Android user dot wiggle. 

- Using `math.geom2d.Point2D` for `double` accuracy in pixel/coordinate conversions instead of `PointF` with `float`. 
- Only re-laying out the user dot on frame render changes, not all viewport changes. 
- Specifically re-laying out the dot after `invalidate()` (not present here). 

These might help and are directions to explore, but I think the main problem is that we can only `requestLayout()` or `forceLayout()` on the user dot, but it's not guaranteed to be synchronous. 

So far, this helps pan activity but not zoom. 

/cc @ljbade @bleege @adam-mapbox

### DIFF
--- a/android/cpp/jni.cpp
+++ b/android/cpp/jni.cpp
@@ -94,10 +94,10 @@ jmethodID projectedMetersConstructorId = nullptr;
 jfieldID projectedMetersNorthingId = nullptr;
 jfieldID projectedMetersEastingId = nullptr;
 
-jclass pointFClass = nullptr;
-jmethodID pointFConstructorId = nullptr;
-jfieldID pointFXId = nullptr;
-jfieldID pointFYId = nullptr;
+jclass point2DClass = nullptr;
+jmethodID point2DConstructorId = nullptr;
+jfieldID point2DXId = nullptr;
+jfieldID point2DYId = nullptr;
 
 jclass httpContextClass = nullptr;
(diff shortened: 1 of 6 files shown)


ljbade:
Will this also tackle the lag?

ljbade:
Also if we decide to use `Point2D` for double, we should implement our own version instead of pulling in a whole new Java dependency for just that.

bleege:
The lag / jumpiness could also be helped via the `MapChange` events updates from e78eacefdb12e270ff990f31e22ee118b467d9b4.

incanus:
> Will this also tackle the lag?

Yes, exactly. 

> if we decide to use `Point2D` for double, we should implement our own version

Agree — I was just testing quickly without all the overhead of making a new class. This should be a totally simple class — constructor with two `double` arguments, `x` and `y` getters as `double`, and maybe a convenience constructor for `float` arguments. 

> The lag / jumpiness could also be helped via the `MapChange` events updates from e78eace.

I'm taking that into account here already by things like: 

> - Only re-laying out the user dot on frame render changes, not all viewport changes.

ljbade:
We fixed this in https://github.com/mapbox/mapbox-gl-native/pull/2712

Issue #1243, opened 2015-04-09
iOS: Background crash — "Failed to bind EAGLDrawable"


friedbunny:
Five or ten minutes after locking my device (iOS 8.3, iPhone 5) but leaving Xcode attached I noticed the demo app had crashed while everything was sitting idle. On `master` 9b0376e58ee9c7b3a94b816158b41eb8dcaf2090.

```objc
2015-04-09 12:40:18.718 Mapbox GL[239:6672] [DEBUG] {0}{Map}[General]: painter setup: 245ms
2015-04-09 12:56:26.258 Mapbox GL[239:6483] Failed to bind EAGLDrawable: <CAEAGLLayer: 0x17699380> to GL_RENDERBUFFER 1
2015-04-09 12:56:26.259 Mapbox GL[239:6483] Failed to make complete framebuffer object 8cd6
Assertion failed: (asyncRender), function triggerRender, file /mapbox-gl-native/src/mbgl/map/map.cpp, line 319.
```

1ec5:
Fixed in #1254.

Sh1mor1nka:
Still get this in SDK 3.1.2, in console log

```
2016-02-25 22:10:27.453 App[19087:4463327] Failed to bind EAGLDrawable: <CAEAGLLayer: 0x81db16f0> to GL_RENDERBUFFER 1
2016-02-25 22:10:27.453 App[19087:4463327] Failed to make complete framebuffer object 8cd6
```

I doesn't get crash, but i get blank map.
Any ideas how can i fix it? Or why i can get such behaviour?
![screen shot 2016-02-25 at 22 13 17](https://cloud.githubusercontent.com/assets/2759167/13332961/0fd24558-dc0d-11e5-84dc-e832e6034cf9.png)

1ec5:
@Sh1mor1nka, does the map view have a width or height of 0 points at the time you first add it to the view hierarchy? If so, you’re running into #1572.

Sh1mor1nka:
@1ec5 Correct, firstly i add it without frame.

Sh1mor1nka:
A'm not sure but may be it's connected...
I add pin on the map, and setCenterCoordinate right on this pin, and here is result:
![screen shot 2016-02-25 at 22 45 55](https://cloud.githubusercontent.com/assets/2759167/13333884/e6065336-dc11-11e5-9e1d-028c56841760.png)
Blue pin on top of map, not in centre.

[...]

Issue #1203, opened 2015-04-03
open gl error when moving back from UIViewController while scrolling is on.


mb12:
The following steps reproduce this error reasonably consistently.

1. Flick to start an animated scroll.
2. Switch/Push a  new UIViewController.
3. Tap on back button before scroll finishes.

It doesn't happen if one lets the scroll finish before pressing back button. Is the error happening because:

- Map thread is still animating (doing gl Calls)
- UI Thread calling renderbufferStorageFromDrawable on a context which is in use in a different thread (map Thread). 

If this is correct, then its a more fundamental problem that can cause random crashes since MGLMapView has no control on when the setNeedsDisplay will be called in the UIView hierarchy where  MGLMapView is embedded. Everytime the code path that triggers [GLKView _display:] -> ...[GLKView _createFramebuffer] is triggered when mapThread is animating will cause a crash.

```
(lldb) bt
* thread #1: tid = 0x9f60e, 0x00000001012d3f3c libglInterpose.dylib`__GPUTOOLS_HAS_DETECTED_AN_OPENGL_ERROR__, queue = 'com.apple.main-thread', stop reason = breakpoint 2.1 11.1 37.1
    frame #0: 0x00000001012d3f3c libglInterpose.dylib`__GPUTOOLS_HAS_DETECTED_AN_OPENGL_ERROR__
    frame #1: 0x00000001012d3fe4 libglInterpose.dylib`handle_opengl_error + 60
    frame #2: 0x0000000101482140 libglInterpose.dylib`EAGLContext_renderbufferStorageFromDrawable(EAGLContext*, objc_selector*, unsigned long, id<EAGLDrawable>) + 1608
    frame #3: 0x000000018771aa68 GLKit`-[GLKView _createFramebuffer] + 328
    frame #4: 0x000000018771b040 GLKit`-[GLKView _setFramebuffer:] + 144
    frame #5: 0x000000018771b294 GLKit`-[GLKView _display:] + 92
    frame #6: 0x000000018a7e6f74 QuartzCore`CA::Layer::display() + 148
    frame #7: 0x000000018a7c875c QuartzCore`CA::Layer::display_if_needed(CA::Transaction*) + 244
    frame #8: 0x000000018a7c8414 QuartzCore`CA::Layer::layout_and_display_if_needed(CA::Transaction*) + 44
    frame #9: 0x000000018a7c7c08 QuartzCore`CA::Context::commit_transaction(CA::Transaction*) + 276
[...]

Issue #1186, opened 2015-04-02
iOS Crash MGLMapView.mm line 554


picciano:
Version.txt is reporting version 9dae9cf06e
Unfortunately, I cannot detect a specific action causing this.

![screen shot 2015-04-02 at 1 57 26 pm](https://cloud.githubusercontent.com/assets/417933/6972352/731d1410-d940-11e4-891d-d783c6214d6f.png)

robipresotto:
i got it everyday too

```ruby
Thread : Crashed: com.apple.main-thread
0  Embark                         0x00000001001c4830 mbgl::Map::renderSync() (memory:2603)
1  Embark                         0x000000010023b140 -[MGLMapView glkView:drawInRect:] (MGLMapView.mm:508)
2  Embark                         0x000000010023b140 -[MGLMapView glkView:drawInRect:] (MGLMapView.mm:508)
3  GLKit                          0x000000018328b3a4 -[GLKView _display:] + 364
4  QuartzCore                     0x000000018638ad88 CA::Layer::display() + 148
5  QuartzCore                     0x000000018636c570 CA::Layer::display_if_needed(CA::Transaction*) + 244
6  QuartzCore                     0x000000018636c228 CA::Layer::layout_and_display_if_needed(CA::Transaction*) + 44
7  QuartzCore                     0x000000018636ba14 CA::Context::commit_transaction(CA::Transaction*) + 276
8  QuartzCore                     0x000000018636b798 CA::Transaction::commit() + 436
9  UIKit                          0x0000000186a14478 _afterCACommitHandler + 156
10 CoreFoundation                 0x00000001821cad98 __CFRUNLOOP_IS_CALLING_OUT_TO_AN_OBSERVER_CALLBACK_FUNCTION__ + 32
11 CoreFoundation                 0x00000001821c7d24 __CFRunLoopDoObservers + 360
12 CoreFoundation                 0x00000001821c8104 __CFRunLoopRun + 836
13 CoreFoundation                 0x00000001820f51f4 CFRunLoopRunSpecific + 396
14 GraphicsServices               0x000000018b43f6fc GSEventRunModal + 168
15 UIKit                          0x0000000186a8610c UIApplicationMain + 1488
16 Embark                         0x0000000100046474 main (main.m:14)
17 libdyld.dylib                  0x0000000193a72a08 start + 4
````

1ec5:
Fixed in #1254.

PR #2309, opened 2015-09-11
Improve transitions in iOS 9 multitasking

bwhtmn:
Leave GLview rendering on when app is inactive, to get nice smooth transitions between sizes  when using multitasking on iPad (iOS 9).

### DIFF
--- a/platform/ios/MGLMapView.mm
+++ b/platform/ios/MGLMapView.mm
@@ -367,6 +367,7 @@ std::chrono::steady_clock::duration secondsAsDuration(float duration)
     [[NSNotificationCenter defaultCenter] addObserver:self selector:@selector(willTerminate) name:UIApplicationWillTerminateNotification object:nil];
     [[NSNotificationCenter defaultCenter] addObserver:self selector:@selector(sleepGL:) name:UIApplicationDidEnterBackgroundNotification object:nil];
     [[NSNotificationCenter defaultCenter] addObserver:self selector:@selector(wakeGL:) name:UIApplicationWillEnterForegroundNotification object:nil];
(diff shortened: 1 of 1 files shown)


1ec5:
At the very least, the GL view needs to be inactive in the background on iOS 8.x and below, to avoid regressing #1050, #1186, #1203, #1243, and #1460. (It’s been quite a journey…) Can you verify that those crashes don’t occur in iOS 9?

1ec5:
Ah, I misread the change: you’re only removing the case where the application becomes inactive, not the case where it goes entirely into the background. That’s probably considerably safer.

bwhtmn:
Yup, this should apply to situations where the app is temporarily interrupted (e.g. multitasking or a system alert).

According to [Apple's docs](https://developer.apple.com/library/prerelease/ios/documentation/iPhone/Conceptual/iPhoneOSProgrammingGuide/StrategiesforHandlingAppStateTransitions/StrategiesforHandlingAppStateTransitions.html#//apple_ref/doc/uid/TP40007072-CH8-SW10), you'd ideally want to throttle down the OpenGL frame rate when the app is inactive but I didn't see an easy way to do that.

incanus:
>  you'd ideally want to throttle down the OpenGL frame rate when the app is inactive but I didn't see an easy way to do that.

Yeah, this is most commonly done in `GLKView` with a corresponding `GLKViewController` (`preferredFramesPerSecond`), which we don't use because we want to be a standalone view like `MKMapView`. Not seeing a shortcut way to do this.

1ec5:
Thanks for this PR. An updated version is now being tracked as #2988.

PR #2310, opened 2015-09-11
[DO NOT MERGE] expose NodeView to JavaScript

mikemorris:
Refs https://github.com/mapbox/node-mapbox-gl-native/issues/142, https://github.com/mapbox/node-mapbox-gl-native/pull/149

May end up not merging this, but wanted to reimplement it here so we can test different caching strategies.

### DIFF
--- a/platform/node/src/node_view.hpp
+++ b/platform/node/src/node_view.hpp
@@ -1,0 +1,45 @@
+#pragma once
+
+#include <mbgl/platform/default/headless_view.hpp>
+
+#pragma GCC diagnostic push
+#pragma GCC diagnostic ignored "-Wunused-parameter"
+#pragma GCC diagnostic ignored "-Wshadow"
+#include <node.h>
+#include <nan.h>
+#pragma GCC diagnostic pop
+
+namespace node_mbgl {
+
+class NodeView : public Nan::ObjectWrap {
+
+struct Options;
+
+    ////////////////////////////////////////////////////////////////////////////////////////////////
+    // Static Node Methods
+public:
+    static NAN_MODULE_INIT(Init);
+
+    static NAN_METHOD(New);
+
+    static Nan::Persistent<v8::FunctionTemplate> constructorTemplate;
+
+    ////////////////////////////////////////////////////////////////////////////////////////////////
+    // Instance
+private:
+    NodeView(Options*);
+    ~NodeView();
+
+    std::unique_ptr<mbgl::HeadlessView> view;
+
+public:
+    inline std::array<uint16_t, 2> getSize() { return view->getSize(); }
+    inline float getPixelRatio() { return view->getPixelRatio(); }
+    
+    inline mbgl::HeadlessView* get() {
+        return view.get();
+    }
+
+};
+
+}
--- a/platform/node/test/js/map.test.js
+++ b/platform/node/test/js/map.test.js
@@ -27,7 +27,7 @@ test('Map', function(t) {
         t.end();
     });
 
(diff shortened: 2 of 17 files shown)


mikemorris:
Tagging @jfirebaugh for a review here.

platform/node/src/node_view.hpp
```diff
@@ -0,0 +1,42 @@
...
+#pragma GCC diagnostic push
+#pragma GCC diagnostic ignored "-Wunused-parameter"
+#pragma GCC diagnostic ignored "-Wshadow"
+#include <node.h>
+#include <nan.h>
+#pragma GCC diagnostic pop
+
+namespace node_mbgl {
+
+class NodeView : public node::ObjectWrap {
```
jfirebaugh: I think this should be `public Nan::ObjectWrap` instead.

mikemorris:
```
terminate called after throwing an instance of 'std::runtime_error'
  what():  Couldn't create framebuffer: incomplete attachment
```
Seeing the above error on Travis Linux `API.RepeatedRender` test.

platform/node/src/node_view.hpp
mikemorris: HAH, yea bet that's it.

platform/node/test/js/map.test.js
```diff
@@ -48,12 +48,12 @@ test('Map', function(t) {
         options.cancel = function() {};
         t.throws(function() {
             new mbgl.Map(options);
-        }, /Options object must have a numerical 'ratio' property/);
+        }, /Options object must have a 'view' or  numerical 'ratio' property/);
```
mikemorris: Reminder to fix the double space after `or`

src/mbgl/map/map_context.cpp
```diff
@@ -27,14 +27,13 @@
 #include <mbgl/util/string.hpp>
 
 #include <algorithm>
+#include <iostream>
```
mikemorris: Drop this

mikemorris:
The [`returns an error`](https://github.com/mapbox/mapbox-gl-native/blob/node-view/platform/node/test/js/map.test.js#L199-L218) test in `map.test.js` is now segfaulting here:
```
Process 65928 stopped
* thread #1: tid = 0x696542, 0x00000001031039c1 mapbox-gl-native.node`mbgl::Map::update(this=0x0000000100d0cc80, flags=<unavailable>) + 33 at map.cpp:97, queue = 'com.apple.main-thread', stop reason = EXC_BAD_ACCESS (code=1, address=0x0)
    frame #0: 0x00000001031039c1 mapbox-gl-native.node`mbgl::Map::update(this=0x0000000100d0cc80, flags=<unavailable>) + 33 at map.cpp:97
   94   void Map::update(Update flags) {
   95       assert(view);
   96       if (flags & Update::Dimensions) {
-> 97           transform->resize(view->getSize());
   98       }
   99       context->invoke(&MapContext::triggerUpdate, transform->getState(), flags);
   100  }
```

mikemorris:
^ Fixed the segfault above (`view` was still `nullptr`) and added better error handling (will now throw a JS error when attempting to call `map.render` before the `view` has been set or if the `view` dimensions don't match the requested render dimensions)

mikemorris:
@lucaswoj @bsudekum This should be ready for some perf testing if you want to try building a `NodeView` cache/pool and swapping in matching views to the `NodeMap` at render time.

jfirebaugh:
Currently it's looking like we don't need this to get decent performance, though it's possible we'll revisit in a future optimization pass.

Issue #2313, opened 2015-09-11
MGLMapView viewport methods should cancel animations


1ec5:
Per #2296, methods like `-[MGLMapView setCenterCoordinate:zoomLevel:]` should call `mbgl::Map::cancelTransitions()`, just like the gesture recognizers do.

/cc @incanus

incanus:
This would probably help https://github.com/mapbox/mapbox-gl-native/issues/1975#issuecomment-138971829.

incanus:
To be clear: the goal here is to avoid any sort of stale viewport changes to occur after applying _another_ viewport change.

tomtaylor:
Just to confirm, while testing #2204, that this is an issue with `setCamera` too.

tomtaylor:
@incanus do you have an idea of how much work this is? The last good version for me was pre2, which seems to also contain some crashing bugs under iOS 9. I'm blocked on releasing an iOS 9 compatible version of my app at the moment.

1ec5:
@adam-mapbox, #2321 has been merged, but it doesn’t seem to be testing this scenario: `-testSetCenterCancelsTransitions` seems to:

1. Go to DC instantaneously.
2. Go to the west with animation.
3. Halfway through the animation:
 1. Go back to DC instantaneously.
 2. Immediately assert that step (i) happened.

However, to adequately test whether setting a coordinate cancels other animations, we need to perform (ii) after the animation in (3) runs to completion. This test doesn’t account for the possibility that the animation continues moving to the west after (3) runs to completion.

Issue #2296, opened 2015-09-10
Setting map position should cancel scroll


erf:
Setting map position should cancel scrolling. There should perhaps be another way of canceling scrolling as well.

ljbade:
I see related to https://github.com/mapbox/mapbox-gl-native/issues/1699

ljbade:
I think this might also affect iOS as there is no `cancelTransitions` for setting coordinates?

ljbade:
I made a PR which also calls `cancelTransitions` in the other `set` functions for position, direction and zoom level.

In https://github.com/mapbox/mapbox-gl-native/pull/2302

ljbade:
Merged PR to `master`.

Awaiting @incanus @bleege eyeballs for release branch cherry pick.

incanus:
Cherry-picked as d894bb7c107018b37a43921fa34c28efe4135d93.

PR #2321, opened 2015-09-14
Add test that setting the center coordinate cancels animations

adam-mapbox:


incanus:
Refs https://github.com/mapbox/mapbox-gl-native/issues/2313. Per @adam-mapbox we don't seem to need to do this on iOS as we did on Android in https://github.com/mapbox/mapbox-gl-native/issues/2296. These tests help keep it that way.

test/ios/MapViewTests.m
```diff
@@ -220,6 +220,26 @@ - (void)testPan {
                       @"panning map right should decrease center longitude");
 }
 
+- (void)testSetCenterCancelsTransitions {
+    CLLocationCoordinate2D dc = CLLocationCoordinate2DMake(38.894368, -77.036487);
+    CLLocationCoordinate2D dc_west = CLLocationCoordinate2DMake(38.894368, -77.076487);
+    [tester.mapView setCenterCoordinate:dc animated:NO];
+    [tester.mapView setCenterCoordinate:dc_west animated:YES];
+    dispatch_after(dispatch_time(DISPATCH_TIME_NOW, 0.15 * NSEC_PER_SEC),
```
incanus: I think this is actually problematic because KIF/XCTest won't necessarily wait for this async dispatch before assuming the test passed. We should perhaps `dispatch_after()`, then follow up with a `[tester waitForAnimationsToFinish];`, then test that the conditions inside the dispatch were met. For example, we should set a variable there, then after waiting, test that the variable is as expected. Otherwise there is no way to guarantee that the block is run. 

More: http://www.pumpmybicep.com/2014/08/20/asynchronous-unit-testing-with-xctest/

adam-mapbox: I amended this to use the asynchronous testing protocols.

Issue #2106, opened 2015-08-17
default marker image asset


incanus:
We should come up with a basic marker (blank Maki one? something else?) that looks good on all/most maps, bundle it alongside other assets like compass imagery, Mapbox logo, etc., and have the platforms fall back to using it if client code doesn't provide a raster image to use. 

We should do this before updating any style pin to https://github.com/mapbox/mapbox-gl-styles/commit/dbabdfee63fa625ed8af32efa3826c6197d8395b or later.

ljbade:
Good idea!

nickidlugash:
Per chat, I can help out with this. Does something like [this](https://github.com/mapbox/maki/blob/mb-pages/src/marker-18.svg) work? But probably a different color.

- What size/file format does this need to be in? 
- What should it be named?
Where should it live?

/cc @1ec5

1ec5:
> What size/file format does this need to be in?

Here’s what we’re currently using on iOS:

https://github.com/mapbox/mapbox-gl-styles/blob/mb-pages/sprites/streets-src/renders/extras/default_marker.png
https://github.com/mapbox/mapbox-gl-styles/blob/mb-pages/sprites/streets-src/renders/extras/default_marker%402x.png

PNG would be best (at both resolutions). SVG isn’t a native image format in UIKit.

> What should it be named? Where should it live?

For iOS, the images should live in https://github.com/mapbox/mapbox-gl-native/tree/master/platform/ios/resources and be called something like “default_marker.png” and “default_marker@2x.png”. Just PR the images and I or @incanus can take care of hooking them up to the SDK.

Per @bleege, Android has the images at various resolutions in https://github.com/mapbox/mapbox-gl-native/tree/master/android/java/MapboxGLAndroidSDKTestApp/src/main/res.

DiogoAndre:
Maybe also good to provide @3x to support the 6plus?  It is not mandatory, but would be nice to have the marker in all supported resolutions. 

Here's a handy guide on iPhone screen resolutions: http://iosdesign.ivomynttinen.com/

Looking forward to having default markers, and then #2250 :smile:

friedbunny:
Ideally we'd have `@3x`, too, but I'm not sure that the GL backend can handle `@3x` sprites right now — @incanus? Might as well produce and include the asset, in any case.

[...]

PR #2323, opened 2015-09-14
moved styles to v1.1.0

nickidlugash:
@1ec5 please double-check and merge if it looks good.


1ec5:
Aside from adding missing shields like `highway-truck-3-small`, what other areas should we be testing?

nickidlugash:
It looks like the past pin was before the sprite refactor, so I guess *all* the sprites in all the styles should be checked.

Also, I rewrote and reordered all the oneway arrow layers (for the roads).

jfirebaugh:
I'm reverting this -- we need to address #2106 first.

PR #2324, opened 2015-09-14
Basic support for Espresso testing added

zugaldia:
Includes docs, base classes, and Gradle dependencies. Works both locally and on AWS Device Farm.


styles
```diff
@@ -1 +1 @@
-Subproject commit 19dde3cb1c36fe5f42dafc78c806fd7aca98126c
+Subproject commit 014e36833420c81b29bec816b479ea61e94821d4
```
jfirebaugh: You probably didn't mean to change any of these submodule pins. Committing them accidentally is a common mistake.

bleege: Good catch @jfirebaugh!  We've gone through and cleaned up a few others as well.  Thanks!

bleege:
Yay!  Thanks so much @zugaldia for contributing!

incanus:
We should have squashed these commits, rebased, force-pushed, and fast-forwarded `master` the single commit, probably.

incanus:
We should move the changes in `android/java/MapboxGLAndroidSDKTestApp/README.md` over to https://github.com/mapbox/mapbox-gl-native/blob/master/INSTALL.md#android to match the other platforms, so build/dev/testing info is all in one place.

incanus:
Is there a way to run these tests locally with a `make` command like on iOS, Mac, Linux, etc? We should hook that up to something like `make atest` for easier testability.

PR #2330, opened 2015-09-15
[node] throw if a render is already pending, fixes #2247

### LINKED ISSUE #2247: `map.render` should throw if a render is already pending
I'm about to add a (skipped) test for this:

```
        t.skip('throws if called in parallel', function(t) {
            var completed = 0;
            var remaining = 10;
            var start = +new Date;

            var map = new mbgl.Map(options);
            map.load(style);

            t.throws(function() {
                map.render({}, function() {});
                map.render({}, function() {});
            });

            map.release();
            t.end();
        });
```

Since parallel renders on the same map are not supported, the node bindings (or core) should actively prevent it.

mikemorris:
/cc @jfirebaugh

### DIFF
--- a/platform/node/src/node_map.cpp
+++ b/platform/node/src/node_map.cpp
@@ -186,6 +186,10 @@ NAN_METHOD(NodeMap::Render) {
         return Nan::ThrowTypeError("Style is not loaded");
     }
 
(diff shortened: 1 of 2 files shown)


jfirebaugh:
:+1:

Issue #2329, opened 2015-09-15
Constraining the map to a predefined geographic region


maciekish:
Hi,
Our search service only covers Sweden, and we want to prevent the map from being panned outside of Sweden. This was possible in the old `MapBox`. Are there any plans to add this to `-gl`?

maciekish:
I don't like waiting so i went ahead and did i myself #2341

friedbunny:
Fixed in #5584.

PR #2341, opened 2015-09-16
 #2329 Constrain map to user specified bounds

maciekish:
This PR enables the user to constrain movement of the map to a specific area.

*Example*
Set the bounds to lock the map to Sweden:

    [self.mapView setUserConstraintsSouthWest:CLLocationCoordinate2DMake(55.336944, 10.9575) andNorthEast:CLLocationCoordinate2DMake(69.06, 24.155833)];

Implemented in `transform_state.cpp` by looking at how the existing `TransformState::constrain(double& scale_, double& y_)` function works to prevent panning the map outside 90 degrees latitude.

The new function `TransformState::userConstrain(double& scale_, double& x_, double& y_)` will constrain scale, x and y. By implementing it at this level both panning, flicking, pinching, double tapping and tapping the map with two fingers will be correctly constrained to the given coordinates. It does however not take angle or pitch into account.

Requested by me in #2329, but since i don't like waiting i went ahead and did it myself. This is my first c++ code **ever** so please be gentle :) Any feedback is appreciated.


maciekish:
Here it is in action http://s.swic.name/d7z3

maciekish:
Updated with the latest changes from master

1ec5:
Thanks for this PR. We’re thinking about a different approach to constraining the viewport that relies on a delegate method, similar to the way UIScrollView works: #2457. The nice thing about that approach is that it’ll handle a “rubber band” effect without much effort on the part of the developer. It’ll also be possible to constrain to a non-rectangular region, which may be important if rotation or tilting is allowed.

1ec5:
Closing for now; will revisit later if #2457 turns out to be inadequate.

PR #2342, opened 2015-09-16
Fix for style properties' transition updates

brunoabinader:
There was no way of knowing whether a style property transition was finished or not. This was causing weird rendering artefacts when e.g. adding/removing the `night` class from the GLFW port. This issue was is also not easily perceived because we're currently setting the default animation/transition times to zero (see `src/mbgl/map/map_data.hpp`).

For each style recalculation, we're now storing the information whether there is still some update needed to complete the style properties transitions. This information is then used by `Style::hasTransitions()` so we can properly call for an asynchronous update in `MapContext::renderSync()` if there are pending style property transitions.

/cc @jfirebaugh @kkaefer @incanus @ljbade @tmpsantos

### DIFF
--- a/src/mbgl/map/map_context.cpp
+++ b/src/mbgl/map/map_context.cpp
@@ -243,7 +243,12 @@ bool MapContext::renderSync(const TransformState& state, const FrameData& frame)
 
     view.afterRender();
 
-    if (style->hasTransitions() || painter->needsAnimation()) {
+    if (style->hasTransitions()) {
+        updateFlags |= Update::Classes;
+        asyncUpdate->send();
+    }
+
+    if (painter->needsAnimation()) {
         updateFlags |= Update::Repaint;
         asyncUpdate->send();
     }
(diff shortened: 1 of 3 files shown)


brunoabinader:
This enables `outdoors` style to properly switch between `day` and `night` now - I'm adding the v8 version here: https://github.com/mapbox/mapbox-gl-styles/pull/210

jfirebaugh:
@brunoabinader Can we hold on this? The future of style classes is in doubt and this adds some statefulness that makes me uncomfortable.

brunoabinader:
> @brunoabinader Can we hold on this? The future of style classes is in doubt and this adds some statefulness that makes me uncomfortable.

:+1: In that case I'd prefer for us to remove both `addClass` and `removeClass` from `Map` interface, as well as removing the shortcut from GLFW demo, so we can avoid an issue that was causing black screen maps after adding the `night` class and then quickly switching between styles. Sounds OK?

brunoabinader:
@jfirebaugh I've updated the patch so the code changes should be much simpler and easier to maintain.

jfirebaugh:
:+1:

PR #2288, opened 2015-09-09
Enable npm install from source

lucaswoj:
cc @jfirebaugh @mikemorris @incanus @1ec5 

resolves #2237

Issue #2295, opened 2015-09-10
Crash in HTTP request when activity destroyed


ljbade:
Following on from #1941 there is a crash in the HTTP request handling, related to HTTP requests that complete after an `Activity` and the corresponding `MapView` are destroyed.

Here is the crash log:
```
09-10 10:22:53.491    2964-2964/? I/DEBUG﹕ *** *** *** *** *** *** *** *** *** *** *** *** *** *** *** ***
09-10 10:22:53.491    2964-2964/? I/DEBUG﹕ Build fingerprint: 'samsung/zerofltedv/zeroflte:5.1.1/LMY47X/G920IDVU2COGA:user/release-keys'
09-10 10:22:53.491    2964-2964/? I/DEBUG﹕ Revision: '11'
09-10 10:22:53.491    2964-2964/? I/DEBUG﹕ ABI: 'arm'
09-10 10:22:53.491    2964-2964/? I/DEBUG﹕ pid: 14785, tid: 14911, name: Tkj2y_5pUcsWvCQ  >>> com.mapbox.mapboxgl.testapp <<<
09-10 10:22:53.491    2964-2964/? I/DEBUG﹕ signal 11 (SIGSEGV), code 1 (SEGV_MAPERR), fault addr 0x3c
09-10 10:22:53.501    2964-2964/? I/DEBUG﹕ r0 00000000  r1 00000000  r2 00000026  r3 0001b7a6
09-10 10:22:53.501    2964-2964/? I/DEBUG﹕ r4 55f1761d  r5 db7ff8a4  r6 db7ff898  r7 db7ff8d4
09-10 10:22:53.501    2964-2964/? I/DEBUG﹕ r8 000000c8  r9 fffff508  sl db7ff8c8  fp ee9cc7e0
09-10 10:22:53.501    2964-2964/? I/DEBUG﹕ ip e14055e8  sp db7ff7a8  lr e118502c  pc e12e8068  cpsr 600e0010
09-10 10:22:53.501    2964-2964/? I/DEBUG﹕ backtrace:
09-10 10:22:53.501    2964-2964/? I/DEBUG﹕ #00 pc 00337068  /data/app/com.mapbox.mapboxgl.testapp-1/lib/arm/libmapbox-gl.so (uv_async_send+8)
[...]

PR #2344, opened 2015-09-16
add OS X / Android armv7 combo to Travis builds

incanus:
The `geojsonvt`-related changes in https://github.com/mapbox/mapbox-gl-native/pull/2288 that landed yesterday turned up some problems in the Android build toolchain on an OS X host. Travis currently only tests Android on Linux hosts, so this adds _yet another_ matrix item to do a simple Android armv7 build on OS X (Xcode 6.4 currently; let's not get _too_ crazy). 

/cc @bleege @mikemorris @jfirebaugh @lucaswoj @ljbade

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -30,6 +30,14 @@ matrix:
     - os: osx
       osx_image: xcode7
       env: FLAVOR=ios BUILDTYPE=Release
+    - os: osx
+      osx_image: xcode6.4
+      env: FLAVOR=android ANDROID_ABI=arm-v7 BUILDTYPE=Release
+      android:
+        components: [ 'build-tools-23.0.1', 'android-23', 'extra-android-m2repository', 'extra-android-support', 'extra-google-m2repository' ]
+      addons:
+        apt:
+          packages: [ 'lib32stdc++6' ]
     - os: linux
       env: FLAVOR=android ANDROID_ABI=arm-v7 BUILDTYPE=Release
       android:
--- a/.travis.yml
+++ b/.travis.yml
@@ -31,7 +31,7 @@ matrix:
       osx_image: xcode7
       env: FLAVOR=ios BUILDTYPE=Release
     - os: osx
-      osx_image: xcode6.4
+      osx_image: xcode7
       env: FLAVOR=android ANDROID_ABI=arm-v7 BUILDTYPE=Release
       android:
         components: [ 'build-tools-23.0.1', 'android-23', 'extra-android-m2repository', 'extra-android-support', 'extra-google-m2repository' ]


ljbade:
Xcode version shouldn't matter

ljbade:
Only travis job failing is new OSX Android matrix which is expected until geojson stuff is fixed.

jfirebaugh:
The build failure here is not due to the geojsonvt stuff:

```
Installing Android dependencies
NaNs$ android-update-sdk --components=build-tools-23.0.1
/Users/travis/build.sh: line 41: android-update-sdk: command not found
```

incanus:
Blocked now by https://github.com/mapbox/mapbox-gl-native/issues/2351.

incanus:
Unblocked.

bleege:
Just to be clear here, in order for Android GL to be a "true" and "useful" test on Travis it needs to run the actual build script.  I'm admittedly a bit unclear about what it currently runs and how it's configured, but from my perspective an Android build on Travis needs to run the following:

* `make android`
  * Compiles the Core GL C++ layer and makes an arm7 ABI `.so` object for Android to use
  * Could also use `make apackage` instead as it compiles ALL shippable ABI
    * Would take longer but would be a "more true" and thorough test
  * The last part of `make android` and `make aproject` runs the [gradle build script](https://github.com/mapbox/mapbox-gl-native/blob/2141d95196cbdf63fd4bd8b34824660bf774bb08/Makefile#L81) which
    * Compiles the Java portion of the Android SDK
    * Produces shipable binaries (aka artifacts, aka Android Library `.aar')
    * Will be configured to run the automated tests on the Android Library

Again, it's not totally clear to me how or what the existing Android build jobs on Travis are configured to do (and I would :heart: to know more ).  It is clear to me that **they currently are not doing** what I described above ^^ as the original geojsonvt issue from #2288 would not have returned the false positives that it did.

incanus:
This PR attempts to run the build process. The problem is that currently we aren't building Android on OS X at all, just Linux, and that's what missed the issue.

bleege:
@incanus Right.  Can you help me understand what running on OS X will do to solve this that Linux currently is failing at?  TBH, I'd prefer that Android be built on OS X as that mirrors what the majority of developers use (at least internally).

Also, can you help me understand exactly what the "build process" means to you / what is being setup?  What I'm worried about is that Travis is being setup to run a "build process" that isn't accurate.

incanus:
OS X vs. Linux: When https://github.com/mapbox/mapbox-gl-native/pull/2288 landed, it wasn't manually tested for running `make android` on a Mac, only Travis-tested, which only tests Linux-based Android builds. Hence this PR to catch this class of problems. 

Build process: Once this is set, it's going to run like the Linux-based Android or [even the OS X client build tests](https://github.com/mapbox/mapbox-gl-native/blob/2141d95196cbdf63fd4bd8b34824660bf774bb08/scripts/osx/run.sh) run: using the stuff in `scripts/android` to essentially run `make android`.

bleege:
Thanks for the explaination @incanus!

Stupid Question: Can we remove the Android on Linux builds and just build Android on OS X?

incanus:
We could, but one thing we _do_ do on the Linux side is test a bunch of ABIs: 

https://github.com/mapbox/mapbox-gl-native/blob/d86d11fb3e22da3ecc5ad692a37fb8aa5efdfc27/.travis.yml#L33-L60

bleege:
True, but can those tests also be done on OS X?  I'm just trying to minimize the number of moving pieces that are floating around.

mikemorris:
@bleege We should make sure we continue to test:
1. Android builds on OS X and Linux.
2. All the Android platforms.
Maybe this means consolidating the matrix into `make apackage` on OS X and Linux (if that doesn't take forever), but we shouldn't just drop the Linux builds.

bleege:
@mikemorris Can you help me understand what running the Android builds on Linux achieves?  Java / Android compilation is platform independent.  As long as the C++ compiler / NDK work the same on both OS X and Linux I don't understand why part of a build is done on OS X and another part is done on Linux.

incanus:
> Can you help me understand what running the Android builds on Linux achieves?

It's the flip side of what we just saw with OS X + Android: inadvertently breaking another team's toolchain with the assumption that tests passing = everything's ok.

jfirebaugh:
What @incanus said. If we're really squeezed for build capacity or Android CI is the critical path (I don't think either is true at the moment) we might consider dropping just the Linux `FLAVOR=android ANDROID_ABI=arm-v7 BUILDTYPE=Release` row from the matrix and leaving the other three Linux Android rows. But testing at least one Android build on both Linux and OS X is valuable.

@incanus What's the current status here? Looks like CI is still failing on `android-update-sdk: command not found`?

bleege:
Ah... so this is more about extending test coverage to other binding SDK products in the specific binding SDK product's natural environment.  ie, npm's home is Linux but when committing it needs to know that it's not breaking iOS and / or Android in their natural environments.  Gotcha.

Thanks for the insight and humoring me @incanus @jfirebaugh @mikemorris .  :+1:

incanus:
> What's the current status here? 

Yes, failing on that command after first cut. I will get back to this after #2295, but I think I need to more comprehensively go through Travis' Android setup guide first.

incanus:
It's not clear from the [Android environment docs](http://docs.travis-ci.com/user/languages/android) what the OS is, but it's probably Linux. Trying this on OS X [fails](https://travis-ci.org/mapbox/mapbox-gl-native/jobs/82052294) with the first command related to the Android SDK. 

I attempted to use Travis' help chat, but no one was currently online, so I left a message per their system. Submitting it then had a web application failure and ate my message, so I'll be trying again later. 

paris_tuileries_garden_facepalm_statue

incanus:
Heard back: 

> At the moment we do not provide any support for Android on OSX. 
> 
> You could manually set it up using `before_install:` steps in your `.travis.yml`, but we can't guarantee it will work.

Will hold on this for now, but this is an approach we can try — downloading the Android SDK stuff and updating the reqs ourselves. Especially since we don't need to test in emulator right now, just build.

ljbade:
@incanus Ah like we used to do when we used Linux environment way back. Checkout some really old commits to see if you can revive the install scripts we had for installing Android on Travis.

incanus:
Nice, thanks @ljbade.

ljbade:
We are switching to bitrise, however they only provide Linux build support.

We will have to keep monitoring OSX build issues manually.

Issue #2358, opened 2015-09-17
Fix GL object management in HeadlessView


jfirebaugh:
In `HeadlessView`, `fboColor` is allocated with `glGenRenderbuffersEXT` [here](https://github.com/mapbox/mapbox-gl-native/blob/d86d11fb3e22da3ecc5ad692a37fb8aa5efdfc27/platform/default/headless_view.cpp#L133-133), but deallocated with `glDeleteTextures` [here](https://github.com/mapbox/mapbox-gl-native/blob/master/platform/default/headless_view.cpp#L205-205). I believe that the latter is incorrect, and will leak GPU resources. Instead, it should be `glDeleteRenderbuffersEXT`.

cc @lucaswoj @adam-mapbox

lucaswoj:
This looks like a legitimate issue that should be fixed. That being said, @adam-mapbox and I have tested this fix locally and think we still see a leak :disappointed:

adam-mapbox:
It is surprising - nay, even shocking - to me that this didn't cause all sorts of problems.  But - it didn't?  Anyway it's fixed now.

PR #2360, opened 2015-09-17
Renderbuffer create should be paired with renderbuffer delete.

adam-mapbox:


lucaswoj:
Resolves #2358 

cc @ljbade @kkaefer @jfirebaugh

kkaefer:
doh, good catch!

PR #2366, opened 2015-09-18
[iOS] disable perfectly good tests so they stop flapping on Travis

### LINKED ISSUE #2347: iOS test failure: testCompassTap, ((tester.mapView.direction) equal to (0)) failed: ("217.430747715") is not equal to ("0") - tapping compass should reset map direction
https://travis-ci.org/mapbox/mapbox-gl-native/jobs/80715437

If we don't have any leads on fixing this, let's disable the test.

cc @incanus @1ec5 @friedbunny

friedbunny:
Travis is inconsistently failing the following tests, so let's temporarily disable them.

- `testDirectionSet`
- `testCompassTap`
- `testDirectionReset`
- `testCenterSet`
- `testDelegateRegionDidChange`

It bears repeating that these tests _only_ fail on Travis, running them locally has always passed.

Fixes #2347

/cc @jfirebaugh @1ec5 @mapbox/gl

### DIFF
--- a/test/ios/MapViewTests.m
+++ b/test/ios/MapViewTests.m
@@ -66,7 +66,7 @@
 
     [tester.compass tap];
 
-    [tester waitForTimeInterval:1];
+    [tester waitForTimeInterval:2];
 
     XCTAssertEqual(tester.mapView.direction,
                    0,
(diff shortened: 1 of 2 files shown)


jfirebaugh:
The test failure is in the node build:

```
* failed background-opacity image
```

However the resulting image looks within tolerances. Might be a failure of shelling out to `compare`; I'll hit https://github.com/mapbox/mapbox-gl-test-suite/issues/8 to try to make this more reliable.

Issue #2076, opened 2015-08-14
Implement convertPoint, convertCoordinate and metersPerPixelAtLatitude on Android


ljbade:
These iOS methods are missing equivalents in Android:
- `convertPoint`
- `convertCoordinate`
- `metersPerPixelAtLatitude`

1ec5:
In Objective-C, parameters are spliced in between parts of a method name, so you're actually looking at `-convertPoint:toCoordinateFromView:` and `-convertCoordinate:toPointToView:`.

ljbade:
Oh wow. that's an interesting syntax.

incanus:
We have some of this already in the form of [these methods](https://github.com/mapbox/mapbox-gl-native/blob/0f80fb9103a3e014f82c7d4990b401706a0430fd/android/java/MapboxGLAndroidSDK/src/main/java/com/mapbox/mapboxgl/views/MapView.java#L816-L823).

ljbade:
@incanus Ah so its only `metersPerPixelAtLatitude` that's missing?

ljbade:
This was done in #2076

PR #2372, opened 2015-09-21
#2076 expose getMetersPerPixel on Android

tobrun:
In #2076 @ljbade indicated that only metersPerPixelAtLatitude was missing.

The [NativeMapview method](https://github.com/mapbox/mapbox-gl-native/blob/1d565a515953432ef9dc79fadaf62e3217c42920/android/java/MapboxGLAndroidSDK/src/main/java/com/mapbox/mapboxgl/views/NativeMapView.java#L404) was already in place, I only needed to expose the method in MapView

To be validated: 
To be more Java compliant, I removed atLatitude part of the method signature because this is already indicated by the parameter. Do we want expose exact the same signature as iOS or go for more standard java compliant signatures?

### DIFF
--- a/android/java/MapboxGLAndroidSDK/src/main/java/com/mapbox/mapboxgl/views/MapView.java
+++ b/android/java/MapboxGLAndroidSDK/src/main/java/com/mapbox/mapboxgl/views/MapView.java
@@ -813,6 +813,18 @@ public class MapView extends FrameLayout implements LocationListener {
         return new PointF(point.x * mScreenDensity, point.y * mScreenDensity);
     }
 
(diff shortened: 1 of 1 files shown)


ljbade:
@tobrun generally I try to follow Java signature as it is more familiar to Android developers.

ljbade:
@tobrun Can you move this to mapbox repo so it can be merged?

Also I think we should call Java function `getMetersPerPixelAtLatitude` to remain consistent with some of the other native functions we have exposed.

ljbade:
Otherwise this is good to merge.

Issue #2259, opened 2015-09-04
cameraForLatLngs fits to untilted bounds while in perspective view


ansis:
cc @1ec5

1ec5:
To recap: `fitBounds` does not account for the trapezoidal shape of the bounds when the map is tilted. As a result, the map’s camera fails to round-trip correctly (tending to zoom out) when the map starts out tilted.

1ec5:
This issue tracks the work to make `Map::cameraForLatLngs()` visually fit to the passed-in bounds when the map is tilted. To illustrate the problem, I rigged iosapp to:

1. Start out unrotated and untilted, and note the map view’s `visibleCoordinateBounds`
1. Add a polygon annotation that covers those bounds
1. Tilt the map by 60° (by taking the current `camera`, adjusting its `pitch`, and setting `camera` to that modified camera)
1. Pass the bounds from (1) into `-[MGLMapView setVisibleCoordinateBounds:animated:]`

The result is that the map is zoomed much farther out than it should be:

![trapezoid](https://cloud.githubusercontent.com/assets/1231218/10617325/bcc41f7e-771e-11e5-85e2-d182005f10ab.png)

ansis:
There are a couple possible behaviours in some cases. Imagine a box that's a horizontal line. It has a large width and small height. You could:

- show the map at the highest possible zoom by putting the box at the top edge of the map
- or you could vertically center the box and use a lower zoom level

Which is better? I prefer the second.

1ec5:
Absent tilt, you can think of the current behavior as a) centering the shape, then b) zooming to fit (but not panning or rotating). So I’d prefer (2) as well. In the future, we could choose an optimal rotation and pitch for a given object, similar to what the [Bing Maps API](https://blogs.bing.com/maps/2015/09/01/drive-app-engagement-with-amazing-new-capabilities-of-windows-10-maps/) does, but that would have to be a separate method.

1ec5:
On iOS, when the map is tilted and you enter targeted course tracking mode (in which both the user location and a target coordinate are fitted to opposite ends of the map view), `cameraForLatLngs()` underzooms, so the map bounces back and forth as it iteratively settles on the correct bounds.

cammace:
Confirmed an issue on Android as well. When the map is tilted and you try fitting the camera within a bounds, it doesn't take into account the tilt. 

```java
LatLngBounds latLngBounds = new LatLngBounds.Builder()
[...]

PR #2373, opened 2015-09-21
Fixed incorrect notification

maciekish:
This is incorrect, no?

`UIGestureRecognizerStateChanged` should yield `MapChangeRegionIsChanging` and not until `UIGestureRecognizerStateEnded` should you get `MapChangeRegionDidChangeAnimated`.

### DIFF
--- a/platform/ios/MGLMapView.mm
+++ b/platform/ios/MGLMapView.mm
@@ -884,7 +884,7 @@ std::chrono::steady_clock::duration secondsAsDuration(float duration)
 
         self.centerPoint = CGPointMake(self.centerPoint.x + delta.x, self.centerPoint.y + delta.y);
 
-        [self notifyMapChange:mbgl::MapChangeRegionDidChangeAnimated];
+        [self notifyMapChange:mbgl::MapChangeRegionIsChanging];
     }
     else if (pan.state == UIGestureRecognizerStateEnded || pan.state == UIGestureRecognizerStateCancelled)
     {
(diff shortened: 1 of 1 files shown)


1ec5:
Thanks @maciekish! MGLMapView has a few more gesture recognizers (`-handleQuickZoomGesture:`, `-handleRotateGesture:`, `-handleTwoFingerDragGesture:`) that would benefit from the same notification flow.

maciekish:
@1ec5 I added it to the quick zoom, but should rotate and pitch really send this notification? From what i understand they don't really change the actual region? I held off with this until further notice because `handleTwoFingerDragGesture` `state == UIGestureRecognizerStateEnded` has a commented out `//[self notifyMapChange:(mbgl::MapChangeRegionDidChange)];` which leads me to believe that someone thought the same..

friedbunny:
> I held off with this until further notice because handleTwoFingerDragGesture state == UIGestureRecognizerStateEnded has a commented out //[self notifyMapChange:(mbgl::MapChangeRegionDidChange)]; which leads me to believe that someone thought the same..

That would be me. :smirk:

Changing the pitch of the map does change the visible region and it should give a change notification, but I never got around to testing the practical effects of actually sending this notification.

maciekish:
It changes the visible region - but does it change `mapView.visibleCoordinateBounds`?

tyoshii-uie:
I have a same problem. Will this be merged shortly?

1ec5:
> I added it to the quick zoom, but should rotate and pitch really send this notification? From what i understand they don't really change the actual region?

Rotating by 45° does expand the visible coordinate bounds a bit. Tilting the map should alter the visible coordinate bounds but currently does not, due to #2259.

1ec5:
Thanks for the PR and your patience – tracking as #2700 now.

Issue #2374, opened 2015-09-21
BoundingBox using wrong latitude/longitudes


tobrun:
While implementing #1948, I noticed strange results in my testcases.
Looking into this, I'm thinking that BoundingBox is using incorrect latitudes/longitudes.

For example take following BoundingBox:

<img width="407" alt="screen shot 2015-09-21 at 11 51 08" src="https://cloud.githubusercontent.com/assets/2151639/9989231/144ab336-6057-11e5-8eb4-eda939187706.png">

Using current implementation the 4 arg constructor would be:
>  BoundingBox(10,5,0,5) 

and using the 2 arg constructor:
> BoundingBox( {20,10} , {0,0} )  
-> this will call the 4 arg as BoundingBox(20,10,0,0) 

Which gives us a different BoundingBox

##### Reason
We are currently using:
 > mLatNorth,  mLonEast, mLatSouth, mLonWest

But I think this should be:
 > mLonNorth, mLatEast,  mLonSouth, mLatWest

ljbade:
Good catch!

tobrun:
Will add supporting testcases and change the implementation

tobrun:
While working on this, I'm noticing that calling intersect always returns a [new BoundingBox](https://github.com/mapbox/mapbox-gl-native/blob/1d565a515953432ef9dc79fadaf62e3217c42920/android/java/MapboxGLAndroidSDK/src/main/java/com/mapbox/mapboxgl/geometry/BoundingBox.java#L216). 
In case two rectangles don't intersect we should return null or throw an exception.

ljbade:
@tobrun I'd say we should return null.

Issue #1948, opened 2015-07-28
Implement missing geography functions in Android from iOS


ljbade:
In `MGLMapView.mm` there is a bunch of functions under "Geography" that have not been implemented in Android.

1ec5:
Not all of them are publicly exposed; see MGLMapView.h for the public ones. There is a bunch more geometry in MGLGeometry.h as well.

ljbade:
OK specific public ones are at https://github.com/mapbox/mapbox-gl-native/issues/2076

As for `MGLGeometry` seems Android missing:
- `MGLCoordinateBoundsGetCoordinateSpan`
- `MGLCoordinateBoundsOffset`
- `MGLStringFromCoordinateBounds`
- `MGLCoordinateBoundsIsEmpty`

1ec5:
Note that the standalone functions in MGLGeometry should probably be implemented on Android as methods of a separate class of some sort. We’ll want a `CoordinateBounds` class, for example.

* `MGLCoordinateBoundsGetCoordinateSpan`: `BoundingBox.getLatitudeSpan()` and `BoundingBox.getLongitudeSpan()` are already implemented, but it would make sense to combine them into a `BoundingBox.getSpan()` that returns a `CoordinateSpan`.
* `MGLStringFromCoordinateBounds`: Already implemented as `BoundingBox.toString()`.

ljbade:
@1ec5 good spotting!

ljbade:
Need to implement `BoundingBox.getSpan()`, `MGLCoordinateBoundsOffset` and `MGLCoordinateBoundsIsEmpty`

tobrun:
The requested methods have been made obsolete in favor of `LatLngSpan` and `LatLngBounds`. 
Closing.

PR #2376, opened 2015-09-21
#2374 BoundingBox using correct latitude and longitude values

tobrun:
PR for #2374 and #1948
- Fixed BoundingBox using correct latitude and longitude values
- Corrected equals method to be java compliant (BoundingBox and CoordinateSpan)
- Added testcases for BoundingBox
- Replaced intersect to handle non-intersecting boundingboxes
- update jni.cpp to handle new naming of BoundingBox members
- added missing methods for #1948

### DIFF
--- a/android/java/MapboxGLAndroidSDK/src/main/java/com/mapbox/mapboxgl/geometry/BoundingBox.java
+++ b/android/java/MapboxGLAndroidSDK/src/main/java/com/mapbox/mapboxgl/geometry/BoundingBox.java
@@ -22,22 +22,21 @@ public final class BoundingBox implements Parcelable, Serializable {
      * Construct a new bounding box based on its corners, given in NESW
      * order.
      *
-     * @param north Northern Coordinate
-     * @param east Eastern Coordinate
-     * @param south Southern Coordinate
-     * @param west Western Coordinate
+     * @param northLatitude Northern Latitude
+     * @param eastLongitude   Eastern Longitude
+     * @param southLatitude Southern Latitude
+     * @param westLongitude   Western Longitude
      */
-    public BoundingBox(final double north, final double east, final double south, final double west) {
-        this.mLatNorth = north;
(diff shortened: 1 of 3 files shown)


ljbade:
@tobrun I'm confused about the naming of the `BoundingBox` members.

Isn't North/South latitude not longitude? Same with East/West as longitude not latitude

tobrun:
The reason why North/South longitude is used, is that these are constant over the whole top/bottom side of the rectangle. If we would take the variable latitude we will have a mismatch using our constructors: (see picture image in #2374)

These 2 constructors should describe the same BoundingBox but are different 
 - BoundingBox( {20,10} , {0,0} ) -> this will call BoundingBox(20,10,0,0)
 - BoundingBox(10,5,0,5)  

Using this PR you will get:
 - BoundingBox( {20,10} , {0,0} ) -> this will call BoundingBox(20,10,0,0)
 - BoundingBox(20,10,0,0)  

Personally I find the longitude/latitude names confusing. I think something like this would make more sense:

- BoundingBox( LatLng north, LatLng east, LatLng south, LatLng west)
- BoundingBox( double longitudeTopSide, double latitudeRightSide, double longitudeBottomSide, double latitudeLeftSide)

ljbade:
Still confused.

@bleege @incanus Can you double check this?

tobrun:
@ljbade you are right, I made the mistake of thinking that the order of (x,y) in geometry is the same order as LatLng in coordinates. Reverted those changes, this PR boils down to the following:

- added missing methods for #1948
- Intersect will not return new BoundingBox is 2 BoundingBoxed don't intersect 
- Testsuite for BoundingBox
- Java styles equals method for BoundingBox and CoordinateSpan (needed for testcases)

ljbade:
@tobrun Can you move this to mapbox repo so it can be merged?

ljbade:
Otherwise this is good to merge.

Issue #2378, opened 2015-09-21
[iOS] Metrics wifi check causes console spew when not connected to network


friedbunny:
```
Sep 21 18:08:18 Hobbes wifid[75] <Error>: WiFi:[464566098.097497]: Network NULL for client configd !
Sep 21 18:08:18 Hobbes wifid[75] <Error>: WiFi:[464566098.971163]: Network NULL for client configd !
Sep 21 18:08:19 Hobbes wifid[75] <Error>: WiFi:[464566099.687127]: Network NULL for client configd !
Sep 21 18:08:20 Hobbes wifid[75] <Error>: WiFi:[464566100.536828]: Network NULL for client configd !
Sep 21 18:08:20 Hobbes wifid[75] <Error>: WiFi:[464566100.867812]: Network NULL for client configd !
Sep 21 18:08:21 Hobbes wifid[75] <Error>: WiFi:[464566101.329745]: Network NULL for client configd !
```

friedbunny:
Oh, and by console I mean the device-wide console, not Xcode's.

friedbunny:
@1ec5 suggested that we use Reachability here, which I'll investigate. Reachability doesn't offer the SSID, but may allow us to avoid the console spam if we wrap [`-wifiNetworkName`](https://github.com/mapbox/mapbox-gl-native/blob/feae692dbaefa51b893528a31139fb8282b3e889/platform/ios/MGLMapboxEvents.m#L740-L753) in a check.

friedbunny:
Incidentally, [`CaptiveNetwork`](https://developer.apple.com/library/ios/documentation/SystemConfiguration/Reference/CaptiveNetworkRef/index.html) is deprecated in iOS 9 in favor of [`NEHotspotHelper`](https://developer.apple.com/library/ios/documentation/NetworkExtension/Reference/NEHotspotHelperClassRef/index.html), which requires special dispensation (an entitlement) from Apple in order to use. We would not qualify for this entitlement, as it's actually reserved for captive network setup.

friedbunny:
Fixed by #2472.

PR #2383, opened 2015-09-22
[iOS] Fix metrics wifi device console spew

friedbunny:
- Fixes this device console warning (#2378):

```
9/22/15, 6:28:22 AM wifid[75]: WiFi:[464610502.447121]: Network NULL for client configd !
```

- Adds reachability check to `-wifiNetworkName`
- Defaults `ssid` to `NULL`, because that's what the metrics backend expects in the case of no network

/cc @1ec5 @incanus


friedbunny:
This is obviated by #2472.

Issue #1837, opened 2015-07-06
Crash in HTTPRequest::handleResponse()


kkaefer:
I'm ocassionally seeing crashes in there on OS X with the NSURL backend, typically during the `Storage.HTTPLoad` test. I've also seen them on Travis, so they're probably not NSURL related. [This crash on Travis](https://travis-ci.org/mapbox/mapbox-gl-native/jobs/69289300#L1555) might also be related.

```
...

Thread 27 (Thread 0x1057 of process 31381):
#0  0x00007fff88fcf286 in __pthread_kill () from /usr/lib/system/libsystem_kernel.dylib
#1  0x00007fff83d1242f in pthread_kill () from /usr/lib/system/libsystem_pthread.dylib
#2  0x00007fff8db02b53 in abort () from /usr/lib/system/libsystem_c.dylib
#3  0x000000010021756d in uv__async_send (wa=0x1003d018e <sqlite3_os_init.aVfs+558>) at src/unix/async.c:190
#4  0x000000010021744d in uv_async_send (handle=0x100a21560) at src/unix/async.c:58
#5  0x0000000100185f82 in send (this=0x0) at ../../src/mbgl/util/uv_detail.hpp:116
#6  mbgl::Request::notify (this=0x6, response_=...) at ../../src/mbgl/storage/request.cpp:55
#7  0x00000001001832ac in mbgl::DefaultFileSource::Impl::notify (this=0x105f81000, request=0x104405b00, response=..., hint=mbgl::FileCache::Hint::Full) at ../../src/mbgl/storage/default_file_source.cpp:174
#8  0x0000000100183488 in operator() (this=<optimized out>, res=..., hint=<optimized out>) at ../../src/mbgl/storage/default_file_source.cpp:131
#9  __invoke<(lambda at ../../src/mbgl/storage/default_file_source.cpp:130:21) &, std::__1::shared_ptr<const mbgl::Response>, mbgl::FileCache::Hint> (__args=<optimized out>, __args=<optimized out>, __f=...) at /Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/../include/c++/v1/__functional_base:413
[...]

PR #2388, opened 2015-09-22
[node] Do not throw if raster image does not load

bsudekum:
Currently, if a vector tile does not load, we still render the map. This is not true for raster tiles; if a raster tile does not load, we throw an error.

For example, imagine a map that has imagery for a given city. When the user is zoomed in on the city and the tiles are in view, all is well. The moment they pan out side the imagery limits, tiles begin to `404` (as intended) but we should still render the given map.

This change allows raster tiles which `404` to pass by.

/cc @jfirebaugh @mikemorris

### DIFF
--- a/src/mbgl/util/worker.cpp
+++ b/src/mbgl/util/worker.cpp
@@ -18,12 +18,8 @@ public:
 
     void parseRasterTile(RasterBucket* bucket, std::string data, std::function<void (TileParseResult)> callback) {
         std::unique_ptr<util::Image> image(new util::Image(data));
-        if (!(*image)) {
-            callback(TileParseResult("error parsing raster image"));
-        }
-
-        if (!bucket->setImage(std::move(image))) {
-            callback(TileParseResult("error setting raster image to bucket"));
+        if (*image) {
+            bucket->setImage(std::move(image));
         }
 
         callback(TileParseResult(TileData::State::parsed));
--- a/src/mbgl/util/worker.cpp
+++ b/src/mbgl/util/worker.cpp
@@ -17,9 +17,18 @@ public:
     Impl() = default;
 
     void parseRasterTile(RasterBucket* bucket, std::string data, std::function<void (TileParseResult)> callback) {
+        if (data.empty()) {
+            return callback(TileParseResult(TileData::State::parsed));
+        }
+
         std::unique_ptr<util::Image> image(new util::Image(data));
-        if (*image) {
-            bucket->setImage(std::move(image));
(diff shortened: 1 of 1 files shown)


jfirebaugh:
This doesn't look the correct fix. We need to distinguish "no data" from "bad data" (== `*image` is false). Also this change removes error checking for `bucket->setImage()`.

mikemorris:
Possibly refs https://github.com/mapbox/mapbox-gl-native/issues/1837#issuecomment-121212673?

/cc @tmpsantos

jfirebaugh:
> Currently, if a vector tile does not load, we still render the map.

@bsudekum Are you sure this is true? It looks to me like a vector tile 404 will produce the same behavior as a raster tile 404.

bsudekum:
> It looks to me like a vector tile 404 will produce the same behavior as a raster tile 404.

@jfirebaugh No, if a raster tile 404's the the process fails with `error parsing raster image` (https://github.com/mapbox/mapbox-gl-native/blob/bc92c2de3988a39d6def34617ade4afb5b82e214/src/mbgl/util/worker.cpp#L22). This is not the case with a vector tile.

Looking at how this is handled in vector land:

```c++
void parseVectorTile(TileWorker* worker, std::string data, std::function<void (TileParseResult)> callback) {
    try {
        pbf tilePBF(reinterpret_cast<const unsigned char *>(data.data()), data.size());
        callback(worker->parse(VectorTile(tilePBF)));
    } catch (const std::exception& ex) {
        callback(TileParseResult(ex.what()));
    }
}
```

Changing the `pbf` function call to have a `data.size()` of 0, the map still renders but there is no vector layer:
`pbf tilePBF(reinterpret_cast<const unsigned char *>(data.data()), 0);`

This makes me think that the `pbf` can accept a empty data input. This is currently not the case with raster however. An image is created from a raster tile that 404's, causing `!image` to be true. 

```c++
std::unique_ptr<util::Image> image(new util::Image(data));
if (!(*image)) {
    callback(TileParseResult("error parsing raster image"));
}
```

jfirebaugh:
Yeah, you are right. I think this happens mostly by accident -- I don't think we were counting on `pbf` accepting an empty input (and we probably shouldn't).

I want to take a deeper look at this today and make 404 handling more consistent between vector and raster, node and other platforms, and unblock a clean fix for #1724.

bsudekum:
Closing, working on this https://github.com/mapbox/mapbox-gl-native/pull/2458

PR #2389, opened 2015-09-22
Remove access token from ios test app

bsudekum:
Although this token will load maps loaded in the `Mapbox.bundle`, custom styles made in mapbox studo will not load glyphs and sprites. This is because these assets are access controlled and cannot be accessed by another user. This will required devs using the test app to bring their own access token.

/cc @1ec5 @jfirebaugh

### DIFF
--- a/test/ios/App-Info.plist
+++ b/test/ios/App-Info.plist
@@ -25,7 +25,7 @@
 	<key>LSRequiresIPhoneOS</key>
 	<true/>
 	<key>MGLMapboxAccessToken</key>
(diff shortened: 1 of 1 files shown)


1ec5:
It looks like we’ve always hard-coded an access token in the test app, perhaps so that automated tests would work without requiring developer intervention to set an access token.

:-1: unless you find a way to keep automated tests happy, perhaps via an environment variable that’s always set as part of the Makefile’s `itest` target.

jfirebaugh:
I think this is confusion between `test/ios` and `ios/app`. The former is for CI and I'm OK with it hard-coding a token, the latter is an example for downstream devs and is BYO token.

Issue #1828, opened 2015-07-02
remove build-time dependency on llvm.org


jfirebaugh:
This host has poor reliability, and when it goes down our builds fail.

1ec5:
Can we at least make use of Travis’ `cache:` option?

jfirebaugh:
@springmeyer Do you know where in the build process the following is coming from:

```
$ curl -sSL "http://llvm.org/apt/llvm-snapshot.gpg.key" | sudo -E apt-key add -
curl: (56) Recv failure: Connection reset by peer
gpg: no valid OpenPGP data found.
```

springmeyer:
@jfirebaugh - I think it is coming from the whitelisted apt package we use for upgrading clang++: https://github.com/mapbox/mapbox-gl-native/blob/108792420c1907f4ecf1d7b03f28e4a5ac33b27f/.travis.yml#L19

> Can we at least make use of Travis’ cache: option?
@1ec5 - good idea. But I went to read about this and see: `Note that APT caching is not available for the container-based infrastructure.` at http://docs.travis-ci.com/user/caching/

ljbade:
Just hit this too.

Why o why Travis

mikemorris:
+1 here, got tripped up on this trying to get https://github.com/mapbox/mapbox-gl-native/pull/2391 merged yesterday

ljbade:
@1ec5 You sure it doesn't work on containers? @daniel-j-h post seems to say it does?

1ec5:
Based on the documentation, apt caching is unavailable to us, and it's an apt package that depends on llvm.org availability. Directory-based caching may work, but it doesn't look like it'll solve this particular problem.

ljbade:
@1ec5 I see now directory caching != apt caching

daniel-j-h:
You could set up a local mirror or make a snapshot of the packages (e.g. check out aptly), and cache its directory. Not sure if worth the effort. You also have to think about cache invalidation.

jfirebaugh:
There doesn't seem to be any reasonable way to do this.

PR #2391, opened 2015-09-22
disable -Wunused-local-typedef on OS X

### LINKED ISSUE #2390: Disable unused local typedef warnings to fix OS X builds
Due to [ticket #11240 in Boost](https://svn.boost.org/trac/boost/ticket/11240), we may need to disable `-Wunused-local-typedef` to fix `make osx` builds in Xcode 7.

/cc @mikemorris @jfirebaugh

mikemorris:
Fixes #2390

/cc @1ec5

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -25,7 +25,7 @@ matrix:
           sources: [ 'ubuntu-toolchain-r-test', 'llvm-toolchain-precise-3.5' ]
           packages: [ 'gdb', 'clang-3.5', 'libstdc++-4.9-dev', 'libstdc++6', 'libllvm3.4', 'xutils-dev', 'libxxf86vm-dev', 'x11proto-xf86vidmode-dev', 'mesa-utils' ]
     - os: osx
-      osx_image: xcode6.4
+      osx_image: xcode7
       env: FLAVOR=osx BUILDTYPE=Debug
     - os: osx
       osx_image: xcode7
@@ -77,15 +77,15 @@ matrix:
           sources: [ 'ubuntu-toolchain-r-test', 'llvm-toolchain-precise-3.5' ]
           packages: [ 'gdb', 'clang-3.5', 'libstdc++-4.9-dev', 'libstdc++6', 'libllvm3.4', 'xutils-dev', 'libxxf86vm-dev', 'x11proto-xf86vidmode-dev', 'mesa-utils' ]
     - os: osx
-      osx_image: xcode6.4
+      osx_image: xcode7
       compiler: clang
       env: FLAVOR=node NODE_VERSION=iojs-v3
     - os: osx
-      osx_image: xcode6.4
+      osx_image: xcode7
       compiler: clang
       env: FLAVOR=node NODE_VERSION=0.12
     - os: osx
-      osx_image: xcode6.4
+      osx_image: xcode7
       compiler: clang
       env: FLAVOR=node NODE_VERSION=0.10
 
--- a/bin/render.cpp
+++ b/bin/render.cpp
@@ -9,13 +9,17 @@
 #include <mbgl/storage/default_file_source.hpp>
 #include <mbgl/storage/sqlite_cache.hpp>
 
+#pragma clang diagnostic push
 #pragma GCC diagnostic push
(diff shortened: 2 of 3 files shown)


mikemorris:
Hrm, Travis OS X builds keep failing with `unknown warning option '-Wno-unused-local-typedef' [-Werror,-Wunknown-warning-option]` (still unknown with plural `typedefs`), but everything builds fine for me locally - what am I doing wrong here?

1ec5:
The iOS builds are happy but the OS X builds are not. I suspect the difference is that the OS X builds are still building on the Xcode 6.4 toolchain, whereas the iOS builds are building on the Xcode 7 toolchain.

mikemorris:
https://github.com/mapbox/mason/pull/114 should be merged and the submodule pin updated before merging this.

incanus:
Per https://github.com/mapbox/mason/pull/114#issuecomment-142677221, will test this in a moment, but is there a way we can limit this (on by default for a reason) compile flag just to Boost @mikemorris?

mikemorris:
@incanus Does clang offer functionality similar to GCC for [using `#pragma` to disable compiler warnings](https://github.com/mapbox/mapbox-gl-native/blob/d89aa51e0f59b2465dbbe5e56dd7c59ed88b2e86/macosx/main.mm#L17)?

1ec5:
[It does](https://github.com/mapbox/mapbox-gl-native/blob/20ad0d51f6e6d82587e2beac34f390356409e152/include/mbgl/ios/MGLTypes.h#L32), so an alternative to this PR would be to surround any import of the problematic Boost header with the relevant `#pragma`.

incanus:
Yes. 

https://github.com/mapbox/mapbox-gl-native/blob/20ad0d51f6e6d82587e2beac34f390356409e152/include/mbgl/ios/MGLTypes.h#L32

mikemorris:
Updated the mason submodule pin now that https://github.com/mapbox/mason/pull/114 has landed. Switched to just ignore `-Wunused-local-typedef` around the problematic `boost/program_options.hpp` include instead of setting `-Wno-unused-local-typedef` in the gypfile.

mikemorris:
Um, for some odd reason Travis can't find `clang` on Linux now...
```
make[2]: clang++-3.5: Command not found
```

mikemorris:
Emailed support@travis-ci.com about the Missing Clang Mystery.

incanus:
bilocchcuaasy_z jpg-large

_Travis_

mikemorris:
It looks like I might've gotten tripped up by an llvm apt repository update - `llvm-toolchain-precise-3.5` is no longer listed as of very recently.
```
Precise (12.04) - Last update : Wed, 23 Sep 2015 20:28:19 UTC / Revision: 248380

deb http://llvm.org/apt/precise/ llvm-toolchain-precise main
deb-src http://llvm.org/apt/precise/ llvm-toolchain-precise main
# 3.6 
deb http://llvm.org/apt/precise/ llvm-toolchain-precise-3.6 main
deb-src http://llvm.org/apt/precise/ llvm-toolchain-precise-3.6 main
# 3.7 
deb http://llvm.org/apt/precise/ llvm-toolchain-precise-3.7 main
deb-src http://llvm.org/apt/precise/ llvm-toolchain-precise-3.7 main
```

jfirebaugh:
If llvm removed the 3.5 packages, you'd presumably have to move to 3.6 as well as update the repository source.

jfirebaugh:
We should remove `--force-yes` from the `apt-get install` command; I bet it's hiding a "package clang-3.5 not found in any sources; ignoring" type error.

mikemorris:
> We should remove --force-yes from the apt-get install

Don't think we can do that, pretty sure that's happening on Travis' end?

jfirebaugh:
Yeah, you're right, must be part of the apt addon.

mikemorris:
> Could not connect to llvm.org:80 (192.17.58.186), connection timed out

Refs https://github.com/mapbox/mapbox-gl-native/issues/1828

mikemorris:
Missing clang was llvm.org's fault -> #1828

PR #2393, opened 2015-09-23
Ensure that LiveTileData can be reparsed

jfirebaugh:
Annotation tiles may become partially parsed just like regular tiles, for example if a point annotation is added to the map before the style's sprite has been loaded. In such cases, they need to be reparsed or the annotation will not be rendered. Previously, the code path for reparsing would be short-circuited by a `dynamic_cast<VectorTileData*>` followed by a null check. This commit removes that case and adds (back) a virtual `reparse` method to the TileData interface.

This is the first of several fixes for annotation rendering. I found it in the course of debugging why the existing `PointAnnotation` test was not actually rendering a marker. In addition to these changes, we need to make that test actually fail if the rendering does not match the expected rendering. This again requires some sort of forgiving image similarity test, or exposing annotations in the node bindings so that we can test in JS and use our existing comparison method.

:eyes: @tmpsantos this touches partial parsing, can you give it a look?

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -25,6 +25,7 @@
 /test/fixtures/api/1.png
 /test/fixtures/api/2.png
 /test/fixtures/database/*.db
+/test/output
 /include/mbgl/shader/shaders.hpp
 /src/shader/shaders_gl.cpp
 /src/shader/shaders_gles2.cpp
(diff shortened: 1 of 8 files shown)


jfirebaugh:
Subsumed by #2393. @tmpsantos :eyes: still welcome there.

PR #2412, opened 2015-09-24
Reposition map to fit callouts

### LINKED ISSUE #1089: adjust iOS callout offsets for screen edges
Background in https://github.com/mapbox/mapbox-gl-native/pull/1074#issuecomment-85766309. This needs to account both for would-be offscreen callouts with `SMCalloutView`'s repositioning delay, as well as would-be obscured callouts with navigation and/or toolbars on a view controller via `constrainedInsets`.

maciekish:
Fixes #1089

Example usage:

    - (void)viewDidAppear:(BOOL)animated
    {
        [super viewDidAppear:animated];
    
        self.mapView.calloutInsets = UIEdgeInsetsMake(self.topLayoutGuide.length,
                                                      0,
                                                      self.bottomLayoutGuide.length,
                                                      0);
    }

### DIFF
--- a/include/mbgl/ios/MGLMapView.h
+++ b/include/mbgl/ios/MGLMapView.h
@@ -41,6 +41,14 @@ IB_DESIGNABLE
 *   @return An initialized map view. */
 - (instancetype)initWithFrame:(CGRect)frame styleURL:(nullable NSURL *)styleURL;
 
+/** Edge insets for positioning callouts.
+ *  
+ *  If your `MGLMapView` is obstructed by transparent navbars or toolbars you may want to set the top and bottom insets to the
+ *  top and bottom layout guides from your view controller.
+ *
+ *  The default value of this property is `UIEdgeInsetsZero`. */
+@property(nonatomic) UIEdgeInsets calloutInsets;
+
 #pragma mark - Accessing Map Properties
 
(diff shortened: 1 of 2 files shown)


tmcw:
@incanus can you review this branch and post back?

incanus:
Thanks for this @maciekish. We need to do some internal testing on it, and there are a couple questions about the code here in particular: 

- Why is the `.mason` submodule altered? 
- We haven't yet used `delayForRepositionWithSize` in this SDK due to some complications around not dismissing the callout, even for such legitimate moves. Is it possible for you to write some tests around this behavior to ensure that we are capturing all scenarios?

maciekish:
@incanus Sorry about the `.mason` submodule. Looks like a merge issue, i removed the change.

Could you try to explain the issue around the callout please? What kind of test would you expect?

Zakay:
+1

1ec5:
Thanks for this pull request! Unfortunately, we later generalized the callout API to allow for custom callout views that don’t necessarily have the same methods or feature set as SMCalloutView (#3456). Additionally, we’re planning to revamp the callout API in #4392, which will remove the dependency on SMCalloutView. We’ll definitely want to address issues like #3154 and #3249, so we’ll revisit the #1089 at that time. Thanks again!

PR #2416, opened 2015-09-24
[linux] switch to llvm-toolchain-precise-3.6 and clang-3.6

mikemorris:
Not sure if this is necessary or worth merging, just discovered that it works while trying to work around the missing `clang-3.5` in https://github.com/mapbox/mapbox-gl-native/pull/2391#issuecomment-142706082


mikemorris:
Sigh, looks like llvm.org is down again -> https://github.com/mapbox/mapbox-gl-native/issues/1828

jfirebaugh:
Closing here since it appears we can continue using 3.5 (modulo [llvm.org stability issues](https://github.com/mapbox/mapbox-gl-native/issues/1828)).

PR #2433, opened 2015-09-28
[iOS] Fix links in generated docs

friedbunny:


friedbunny:
Bad description — more like, "distributed docs".

PR #2437, opened 2015-09-28
Adam track texture allocations

adam-mapbox:
I'm going to write up a doc about this, but this is the start of some code to track our GL usage.  If you want to turn it on, go to gl.hpp and uncomment GL_TRACK.

### DIFF
--- a/platform/default/headless_view.cpp
+++ b/platform/default/headless_view.cpp
@@ -11,7 +11,6 @@
 #include <string>
 #include <cstring>
 #include <cassert>
-#include <iostream>
 
 #ifdef MBGL_USE_CGL
 #include <CoreFoundation/CoreFoundation.h>
@@ -23,7 +22,6 @@ namespace mbgl {
 
 HeadlessView::HeadlessView(float pixelRatio_, uint16_t width, uint16_t height)
     : display(std::make_shared<HeadlessDisplay>()), pixelRatio(pixelRatio_) {
-    std::cout << "HeadlessView::HeadlessView" << std::endl;
     resize(width, height);
 }
 
@@ -32,12 +30,10 @@ HeadlessView::HeadlessView(std::shared_ptr<HeadlessDisplay> display_,
                            uint16_t width,
                            uint16_t height)
     : display(display_), pixelRatio(pixelRatio_) {
-    std::cout << "HeadlessView::HeadlessView" << std::endl;
     resize(width, height);
 }
 
 void HeadlessView::loadExtensions() {
-    std::cout << "HeadlessView::loadExtensions" << std::endl;
     if (extensionsLoaded) {
         return;
     }
@@ -67,7 +63,6 @@ void HeadlessView::loadExtensions() {
 }
 
 void HeadlessView::createContext() {
-    std::cout << "HeadlessView::createContext" << std::endl;
     if (!display) {
         throw std::runtime_error("Display is not set");
     }
@@ -120,7 +115,6 @@ bool HeadlessView::isActive() {
 }
 
 void HeadlessView::resizeFramebuffer() {
-    std::cout << "HeadlessView::resizeFramebuffer: " << this << std::endl;
     assert(isActive());
 
     if (!needsResize) return;
@@ -141,7 +135,6 @@ void HeadlessView::resizeFramebuffer() {
     MBGL_CHECK_ERROR(glRenderbufferStorageEXT(GL_RENDERBUFFER_EXT, GL_RGBA8, w, h));
     MBGL_CHECK_ERROR(glBindRenderbufferEXT(GL_RENDERBUFFER_EXT, 0));
 
-    std::cout << "I am creating a new framebuffer right here: " << this << " " << fbo << std::endl;
     MBGL_CHECK_ERROR(glGenFramebuffersEXT(1, &fbo));
     MBGL_CHECK_ERROR(glBindFramebufferEXT(GL_FRAMEBUFFER_EXT, fbo));
 
@@ -178,7 +171,6 @@ void HeadlessView::resize(const uint16_t width, const uint16_t height) {
 }
 
 std::unique_ptr<StillImage> HeadlessView::readStillImage() {
-    std::cout << "HeadlessView::readStillImage: " << this  << "fbo = " << fbo<< std::endl;
     assert(isActive());
 
     const unsigned int w = dimensions[0] * pixelRatio;
@@ -204,16 +196,13 @@ std::unique_ptr<StillImage> HeadlessView::readStillImage() {
 }
 
 void HeadlessView::clearBuffers() {
-    std::cout << "HeadlessView::clearBuffers: " << this << "fbo = " << fbo << std::endl;
     assert(isActive());
 
     MBGL_CHECK_ERROR(glBindFramebufferEXT(GL_FRAMEBUFFER_EXT, 0));
 
     if (fbo) {
         MBGL_CHECK_ERROR(glDeleteFramebuffersEXT(1, &fbo));
-        std::cout << "HeadlessView::clearBuffers2: " << this << "fbo = " << fbo << std::endl;
         fbo = 0;
-        std::cout << "HeadlessView::clearBuffers3: " << this << "fbo = " << fbo << std::endl;
     }
 
     if (fboColor) {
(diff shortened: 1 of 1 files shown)


incanus:
Nice. @lucaswoj is this the incarnation that you used? You have a better sense of the usefulness of this PR. 

Also, I'm curious why things like `GEOJSONVT_VERSION` and `src/mbgl/map/annotation.hpp` are updated here — I'm thinking this is out of sync with current `master`. @adam-mapbox was the merge at https://github.com/mapbox/mapbox-gl-native/commit/9b99e26830d8aab50266304e596779e3f390302f maybe from a stale `master`? Try merging `origin/master` after a `git fetch` to ensure you've got the latest upstream. 

After we're happy with the look here, we should squash it down a bit to cut some of the commit noise. 

/cc @kkaefer for some :eyes: as well.

lucaswoj:
Yes, this is based on a stale master or a fork of master and should be rebased.

@adam-mapbox and I used this in conjunction with a few other tools like `Instruments.app` and the shadowy [`Graphics Tools.app`](https://developer.apple.com/library/mac/technotes/tn2178/_index.html) to track down a couple GPU memory leaks. 

I'll let @adam-mapbox make the case for whenever or not this should be merged into `master` -- I'm not an expert in that area.

adam-mapbox:
I do think it can/should be merged in.  As to the stale fork, I think I made a mistake and I merged instead of rebasing.  Anyone want to help me extricate myself from the current situation?

lucaswoj:
Rebased and squashed for you @adam-mapbox. Previous SHA was 69765c5, if you need to back out my changes.

incanus:
@lucaswoj Help him walk through it next time :wink: 

@adam-mapbox Looking better, but there are a couple things I think we should address. I agree that this should live in `master` when it's ready, though. 

- [ ] We should back out the accidental app token commit in `ios/app/app-info.plist`. Reason is we want developers to have an account and get in the flow of using their own account's token. 

- [ ] Unless I'm missing something, instead of `std::cout`, we should be using our own `mbgl::Log` facilities. First, for consistency, but second, we should be able to control this logging, which otherwise would occur in all uses of `Headless`, right? Even without `GL_TRACK`?

adam-mapbox:
@incanus The reason I used std::cout instead of Log was that, when we were running it for @lucaswoj , through node, the Log messages weren't coming out.  We couldn't figure that out, so we went back to using cout.

incanus:
Ah! Good call. Through node, that does seem likely. @jfirebaugh @mikemorris does that seem ok to you?

jfirebaugh:
The `cout`s look to me like temporary debugging code that should be removed before this is merged.

incanus:
@adam-mapbox Looking like some of the out-of-scope changes are back as of the merge in https://github.com/mapbox/mapbox-gl-native/commit/5bf16d9aca96338e63778f0a6acc0282b1148421. I can tell this by looking at [the files tab for this PR](https://github.com/mapbox/mapbox-gl-native/pull/2437/files) and seeing the overall changes that are proposed. This updates with each commit. 

You can also double-check the overall impact with something like `git diff origin/master..adam_track_texture_allocations`, which will show the differences introduced in this branch.

incanus:
I can't seem to resolve the state of this branch anymore. @lucaswoj can you re-push your rebase from before? Then @adam-mapbox you will need to `git fetch; git reset --hard origin/adam_track_texture_allocations`, then reapply your commit of https://github.com/mapbox/mapbox-gl-native/commit/41edeafb10b80974a2e61f06f9c12f1b759fd276 before pushing again. You should not push any merge commits such as https://github.com/mapbox/mapbox-gl-native/commit/5bf16d9aca96338e63778f0a6acc0282b1148421.

lucaswoj:
Pushing now -- overwriting 41edeafb10b80974a2e61f06f9c12f1b759fd276 with 0797402a174acfdf0fb61a3475ef4aa4ba5123ed.

lucaswoj:
I'm not 100% on this, but I seem to remember us thinking `mbgl::Log` wasn't working when really our logging code wasn't being triggered at all.

mikemorris:
@lucaswoj in the Node bindings, mbgl::Log streams to an EventEmitter on either the module or map (forget if it changed with Environments), you have to listen on that for events and then log output somewhere.

incanus:
That explains the logging not working for @adam-mapbox and @lucaswoj, but it sounds like per @jfirebaugh that we should remove the logging entirely now.

jfirebaugh:
We should remove the debug logging in `HeadlessView`. The logging in `gl.cpp` is ok.

ios/app/app-info.plist
```diff
@@ -2,6 +2,8 @@
 <!DOCTYPE plist PUBLIC "-//Apple//DTD PLIST 1.0//EN" "http://www.apple.com/DTDs/PropertyList-1.0.dtd">
 <plist version="1.0">
 <dict>
+	<key>MGLMapboxAccessToken</key>
```
jfirebaugh: Revert this change, per https://github.com/mapbox/mapbox-gl-native/pull/2437#issuecomment-143886759.

src/mbgl/map/map.cpp
```diff
@@ -11,6 +11,8 @@
 #include <mbgl/util/thread.hpp>
 #include <mbgl/util/math.hpp>
 
+#include <iostream>
```
jfirebaugh: Not used.

src/mbgl/platform/gl.cpp
```diff
@@ -55,6 +59,143 @@ void checkError(const char *cmd, const char *file, int line) {
...
+#undef glBindTexture
+#undef glDeleteTextures
+#undef glTexImage2D
+#undef glClear
+#undef glShaderSource
+#undef glBufferData
+#undef glBindBuffer
+#undef glDeleteBuffers
+#undef glBufferData
+static unsigned int current_used_bytes = 0;
```
jfirebaugh: Please use camel case variables here and elsewhere in the change (`currentUsedBytes`).

src/mbgl/platform/gl.cpp
```diff
@@ -55,6 +59,143 @@ void checkError(const char *cmd, const char *file, int line) {
...
+static std::map<GLint, GLsizeiptr> buffer_binding_to_size_map;
+static unsigned int current_used_buffer_bytes = 0;
+static unsigned int largest_amount_used_so_far = 0;
+
+static std::mutex g_debug_mutex;
 
+void mbx_glBindBuffer(GLenum target,
+                      GLuint buffer) {
+    g_debug_mutex.lock();
+    if(target == GL_ARRAY_BUFFER) {
```
jfirebaugh: Space between `if` and `(` (and `for` and `(`, etc.)

src/mbgl/platform/gl.cpp
```diff
@@ -55,6 +59,143 @@ void checkError(const char *cmd, const char *file, int line) {
...
+static GLuint current_element_array_buffer = 0;
+static std::map<GLint, GLsizeiptr> buffer_binding_to_size_map;
+static unsigned int current_used_buffer_bytes = 0;
+static unsigned int largest_amount_used_so_far = 0;
+
+static std::mutex g_debug_mutex;
 
+void mbx_glBindBuffer(GLenum target,
+                      GLuint buffer) {
+    g_debug_mutex.lock();
```
jfirebaugh: Use `std::lock_guard` rather than manual `lock` / `unlock`.

adam-mapbox:
The couts are *not* temporary debugging code.  They are sort of the whole point of this.  Please don't remove them.

jfirebaugh:
@adam-mapbox I'm talking about the logging in HeadlessView like:

```
std::cout << "HeadlessView::HeadlessView" << std::endl;
```

We can't have this sort of debug logging in a production build: it would clutter the output in places like our headless rendering tests and server side GL rendering.

I agree that the logs in gl.cpp should remain.

adam-mapbox:
Yes, right, of course.  That was an error.

incanus:
test comment

incanus:
Force-pushed to c958ee45abe978274c1c86b8517b2c86bc303c8f — forget to force push before fast-forward marging into `master`.

incanus:
(Worked through rebasing etc. with @adam-mapbox on this branch locally).

Issue #2198, opened 2015-08-28
MGLAnnotationImage with scale greater than 1.0 causes crash.


petrmanek:
Hello, I'm using MapboxGL 0.5 in my iOS map. When trying to annotate map with images of markers on retina displays, I got this error:
```
libc++abi.dylib: terminating with uncaught exception of type mbgl::util::SpriteImageException: Sprite image pixel count mismatch
```

Here's my setup:
```swift
class MyViewController: UIViewController, MGLMapViewDelegate {
    var mapView: MGLMapView!

    override func viewDidLoad() {
        super.viewDidLoad()
        
        // Create the map view.
        mapView = MGLMapView(frame: CGRect(x: 0, y: 0, width: 300, height: 300))
        mapView.delegate = self
        view.addSubview(mapView)

        // Add a single marker.
        let point = MGLPointAnnotation()
        point.coordinate = ...
        mapView.addAnnotation(point)
        mapView.setCenterCoordinate(...)
    }

    func mapView(mapView: MGLMapView, imageForAnnotation annotation: MGLAnnotation) -> MGLAnnotationImage? {
        // Memory Efficiency: If a marker is available from the reusable queue, make use of it.
        if let pin = mapView.dequeueReusableAnnotationImageWithIdentifier("customPin") {
            return pin
        }

        // Scale of the image below is automatically selected based on the display's pixel density.
        // Both @1x and @2x version is available. However, only @2x version causes the crash.
        let image = UIImage(named: "myCustomImage")!
        return MGLAnnotationImage(image: image, reuseIdentifier: "customPin")
    }
}
```

As I said in the comments, the crash occurs only when `image.scale > 1.0`. I've attempted to use the @1x version of the image only but that looks blurry and ugly on the retina display.

To confirm that the error isn't caused by a malformed image file, I've also tried several other @2x images. All of them crashed the app.

I have even managed to prove that this bug is caused explicitly by the scale of the image. The following code works because I've manually tuned down the scale of the @2x image from 2.0 to 1.0. The annotation displays on the map view correctly but has double the size of the original image. You can try and substitute any other scale in the `let scaledUpImage...` line to see my point. If you use scale greater than 1.0, you get a crash.

```swift
    func mapView(mapView: MGLMapView, imageForAnnotation annotation: MGLAnnotation) -> MGLAnnotationImage? {
        if let pin = mapView.dequeueReusableAnnotationImageWithIdentifier("customPin") {
            return pin
        }

[...]

PR #2440, opened 2015-09-28
adds default marker files

nickidlugash:
Ref https://github.com/mapbox/mapbox-gl-native/issues/2106

@1ec5 @bleege @incanus @friedbunny I've added marker files to iOS. Looking at Android now. 

While I'm doing that, here's some visuals – let me know if you have any feedback.

#### Streets

screen shot 2015-09-28 at 5 54 10 pm
screen shot 2015-09-28 at 5 53 46 pm

#### Emerald

screen shot 2015-09-28 at 5 50 10 pm
screen shot 2015-09-28 at 5 50 39 pm

#### Satellite w/ labels

screen shot 2015-09-28 at 5 49 19 pm
screen shot 2015-09-28 at 5 49 29 pm

#### Dark

screen shot 2015-09-28 at 5 48 06 pm
screen shot 2015-09-28 at 5 48 40 pm

#### Light

screen shot 2015-09-28 at 5 46 25 pm
screen shot 2015-09-28 at 5 46 31 pm


incanus:
Two thoughts here: 

1. We're switching from blue as the current default for the vector styles to red now. 

1. Red markers of that style are pretty close to Google's look. Apple also uses red (pins), so using blue alongside our default iOS blue tint user dot and lower-right info button looks pretty darn good, IMHO. 

One minor thought: it always seems to me that the white dot on the markers should either be a) gone or b) alpha-transparent instead of white. Anyone else?

bleege:
I could do without the white dot.

nickidlugash:
> Red markers of that style are pretty close to Google's look. Apple also uses red (pins), so using blue alongside our default iOS blue tint user dot and lower-right info button looks pretty darn good, IMHO.

I chose red because it is higher contrast from more basemap features (e.g. that blue is pretty darn similar to interstate highway shields in both Streets and Emerald). I also think it's more readable for the markers to be a different color than the user dot (plus the two shades of blue clash). It *is* more similar to Google's and Apple's look, but I think there's a reason they both use red. I would still lean more towards red, unless there are any more considerations for keeping it blue.

> One minor thought: it always seems to me that the white dot on the markers should either be a) gone or b) alpha-transparent instead of white. Anyone else?

I was going to remove it because I found it extraneous, but @samanpwbb pointed out that it does help a lot to articulate clustered markers, and I agree. I can try making it alpha-transparent.

@bleege for the default markers for Android can you please confirm that:

1. I should put 1 file into each of the first 5 folders here: https://github.com/mapbox/mapbox-gl-native/tree/master/android/java/MapboxGLAndroidSDKTestApp/src/main/res
2. `mdpi` = "standard" size in iOS, `hdpi` = `@1.5x`, `xhdpi` = `@2x`, `xxhdpi` = `@3x`, and `xxxhdpi` = `@4x`
3. All the files should be named the same: `default_marker.png`

nickidlugash:
> I can try making it alpha-transparent.

Er, nevermind, this is not going to look good with clustered markers, which was the point of keeping the dot.

incanus:
I'm sold on red + white dot as is. Carry on :smiley: 

Let's square away Android assets as well, then @mapbox/mobile can jump in and adjust code as necessary to shore this up.

nickidlugash:
@bleege @incanus @1ec5 Android assets are added now. @bleege please double-check the Android assets when you get a chance – you weren't online when I added them but I wanted to get something up before I left tonight. If anything looks incorrect I can fix tomorrow.

friedbunny:
This looks great @nickidlugash, really like the change to red.

ljbade:
Looks good to me.

bleege:
@nickidlugash Yep, Android sizes [are documented here](https://github.com/mapbox/mapbox-gl-native/issues/2182#issuecomment-135127185) so it looks like [your ratios above](https://github.com/mapbox/mapbox-gl-native/pull/2440#issuecomment-143895706) are :ok_hand: .  The [assets generated from them](https://github.com/mapbox/mapbox-gl-native/commit/c021b0336f66e7e7e6e5601bae322a08c4595bd9) also look great!  Thanks!

1ec5:
The iOS side is done. Over to @bleege for the Android side.

ljbade:
@nickidlugash I'd say lets merge, we can do the connecting in a seperate PR

incanus:
@nickidlugash Do you have an SVG or similar vector source that we can roll into the project somehow for future reference? e.g. There may someday be higher-res requirements that can't simply scale up the largest we have here now, and we don't want to fish around trying to remember where the source is and/or recreate it.

nickidlugash:
@incanus sure – where do you want me to put the source file?

incanus:
@nickidlugash Let's make a `platform/ios/resources_source` folder maybe? It shouldn't need added to anything but Git, as it'll be fetched from the filesystem only (not Xcode, etc.)

incanus:
Actually, let's go with `platform/default/resources` @nickidlugash so we can source for all platforms.

ljbade:
I'm going to wire up the Android side.

ljbade:
I've hit a snag on Android. The current `setSprite` API is too iOS centric and blows up on Android where you can have up to 4x DPI (so far).

It also blows up when you need a bitmap scale <1x, which can happen on low DPI devices. Another case might be where a developer only includes hdpi bitmaps assets, then a user runs the app on a ldpi or mdpi phone. In this case we need to scale down.

Going to push a commit that just uses a scale of 1x for the moment which on my Samsung S6 results in a comically large marker. This is so we can get this merged for iOS. We will need to come back to it for Android and fix this scaling problem before we delete the "default_marker" from the styles submodule.

1ec5:
The strict resolution-dependent nature of this API has inconvenienced iOS developers too: #2198.

ljbade:
Going to rebase and merge next.

ljbade:
@jfirebaugh Looks like LLVM is down again:
```
 curl -sSL "http://llvm.org/apt/llvm-snapshot.gpg.key" | sudo -E apt-key add -
curl: (56) Recv failure: Connection reset by peer
gpg: no valid OpenPGP data found.
```

ljbade:
OK everything passed but Linux and node due to LLVM server issues. Since this change only affects iOS and Android, I'm going to merge.

Issue #2451, opened 2015-09-29
Add InfoWindow adapter


tobrun:
Google Maps allows it InfoWindows to be managed by an [InfoWindowAdapter](https://developers.google.com/android/reference/com/google/android/gms/maps/GoogleMap#setInfoWindowAdapter(com.google.android.gms.maps.GoogleMap.InfoWindowAdapter))

tobrun:
Currently InfoWindow requires this method to start showing the content:
```java
     String title = overlayItem.getTitle();
        ((TextView) mView.findViewById(mTitleId /*R.id.title*/)).setText(title);
        String snippet = overlayItem.getSnippet();
        ((TextView) mView.findViewById(mDescriptionId /*R.id.description*/)).setText(snippet);
 ```
This should only be executed if we are in default state and not while using the adapter (the end developer will be supplying the View content of InfoWindow).

PR #2453, opened 2015-09-29
Add InfoWindowAdapter

tobrun:
had to move some parts in InfoWindow + Marker too make it work
See issue #2451


tobrun:
Will add extra menu item to sample app first

tobrun:
Added example to sample app, Currently reusing secondMapActivity, will be refactored in other issue to update sample app with fragments

PR #2458, opened 2015-09-29
[node] Handle tiles which 404 better

bsudekum:
Continuation of: https://github.com/mapbox/mapbox-gl-native/pull/2388

Currently, if a vector tile does not load, we still render the map. This is not true for raster tiles; if a raster tile does not load, we throw an error.

For example, imagine a map that has imagery for a given city. When the user is zoomed in on the city and the tiles are in view, all is well. The moment they pan out side the imagery limits, tiles begin to 404 (as intended) but we should still render the given map.

This change short circuits all tiles which 404 and does not attempt to parse/render these tiles. With this change, a (node) filesource which retrieves a tile that 404's should return the `callback()` with no arguments. It might look like:

```js
if (res.statusCode === 404 && req.kind === 3) { // tile 404's and is of type tile
  callback(); // just return an empty callback, no data
} 
```

/cc @mikemorris @jfirebaugh

### DIFF
--- a/include/mbgl/storage/response.hpp
+++ b/include/mbgl/storage/response.hpp
@@ -7,7 +7,7 @@ namespace mbgl {
 
 class Response {
 public:
-    enum Status : bool { Error, Successful };
+    enum Status { Error, Successful, NotFound };
 
     Status status = Error;
     std::string message;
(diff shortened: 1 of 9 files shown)


incanus:
This has applications to mobile as well — the old raster kits used to allow for "patchwork" partial-coverage tile sources like this.

PR #2464, opened 2015-09-30
[core] Improve error message when style is not found or is incompatible

### LINKED ISSUE #1724: More descriptive error when trying to load custom styles

friedbunny:
Attempting to use a Mapbox-hosted incompatible raster map/style with GL would throw a simple 404, without any indication of the real issue.

Instead, show this message:

`Mapbox GL [ERROR] {Map}[Setup]: style mapbox://user.mapid could not be found or is an incompatible legacy map or style`

Fixes #1724; ugly first fix attempt was #1841.

/cc @jfirebaugh

### DIFF
--- a/src/mbgl/map/map_context.cpp
+++ b/src/mbgl/map/map_context.cpp
@@ -118,7 +118,7 @@ void MapContext::setStyleURL(const std::string& url) {
 
         if (res.status == Response::Successful) {
             loadStyleJSON(res.data, base);
-        } else if (res.status == Response::NotFound && (styleURL.find("mapbox://") != std::string::npos)) {
+        } else if (res.status == Response::NotFound && styleURL.find("mapbox://") == 0) {
             Log::Error(Event::Setup, "style %s could not be found or is an incompatible legacy map or style", styleURL.c_str());
         } else {
             Log::Error(Event::Setup, "loading style failed: %s", res.message.c_str());


src/mbgl/map/map_context.cpp
```diff
@@ -113,6 +113,8 @@ void MapContext::setStyleURL(const std::string& url) {
 
         if (res.status == Response::Successful) {
             loadStyleJSON(res.data, base);
+        } else if (res.status == Response::NotFound && (styleURL.find("mapbox://") != std::string::npos)) {
```
jfirebaugh: `styleURL.find("mapbox://") != 0`? Should match only at the beginning. Also, extra parentheses are not needed here.

friedbunny: `std::string::npos` is `-1` and matches against the entire string, yes. Check out b385ef269808a2be132de5a9ad41b5c8aa34d099, which only proceeds on matches that start at position `0` — i.e., the beginning of the string.

friedbunny:
Forgot to force-push this branch after rebasing. Also forgot to squash. :eggplant:

Merged as ae5a345cc4685bdfd0de6fa73b126b08ec895ae5  & e5aab73b604cb723742ead8a7762e7c8430853bf.

Issue #2467, opened 2015-09-30
Expose setVisibility for Logo - AttributionIcon


tobrun:
Indicated by @incanus that it must be possible to hide Logo and AttributionIcon,
Need to expose these methods in MapView.

tobrun:
![device-2015-09-30-180213](https://cloud.githubusercontent.com/assets/2151639/10198708/747937e4-679d-11e5-929f-28016d2ae79c.png)

tobrun:
PR added

PR #2468, opened 2015-09-30
Exposing attribution + logo visibility

tobrun:
See #2467


incanus:
@tobrun Looks like this has conflicts since the introduction of the view configurability. Also, I don't think we need to abstract the widget visibility for such small use, especially considering we use the direct view visibility API elsewhere. Want to patch this up and force push?

tobrun:
@incanus thank you for taking time to review, will patch it up and push. 
The conflicts are because of the Javadoc @ljbade pushed today

PR #2470, opened 2015-09-30
Update CONTRIBUTING doc to ask for [platform] prefixed commits

friedbunny:
We've discussed this internally, but let's indoctrinate the entire world. With the `mapbox-gl-native` repository now being home to actively developed node, iOS, and Android projects, it's important to be able to efficiently distinguish which commits belong to which project.

Here is what's been proposed by @tmpsantos, which many of us have been using:

>[node] Fixed this and that
>[iOS] Added feature x
>[Android] Fixed crash when blah
>[core] Improved tile rendering
>[tests] Added more tests

/cc @mapbox/gl @mapbox/mobile

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -9,3 +9,6 @@ If you want to contribute code:
 1. Ensure that the [existing issues](https://github.com/mapbox/mapbox-gl-native/issues?utf8=✓&q=) don't already cover your question or contribution. 
 
 1. Pull requests gladly accepted. 
+
+1. Prefix your commit messages with the platform(s) your changes affect: `[core]`, `[ios]`, `[android]`, `[node]`, and so on.
+


jfirebaugh:
Big :+1:.

mikemorris:
I've been tagging lowercase `[osx]` and `[ios]`, are we standardizing on capitalized versions?

friedbunny:
@mikemorris I like lowercase, we can standardize on that if there's agreement.

bleege:
I'd go along with all lowercase.  BTW, how should we handle issue tagging?  Something like:

```
#24 - [android] Made awesome new feature
```

or

```
[android] #24 - Made awesome new feature
```

1ec5:
@bleege, I’d prefer that we continue using GitHub’s labels feature for issues and PRs. They’re much easier to filter by.

bleege:
> I’d prefer that we continue using GitHub’s labels feature for issues and PRs.

Agreed @1ec5.  I don't think the intent of this is to stop using those.  I think it was to add the `[platform]` to the commit message to make it easier to :eyes: in commit lists in addition to the Github labels.  I was just asking about the format for how to include issue numbers too as I like to do that in my commit messages to keep things aligned.

1ec5:
Ah, thanks for the clarification. As far as I know, we aren't building any tools to parse out issues and platforms from commit messages yet, so I don't think we need to be concerned about machine readability for now.

1ec5:
Committed in ab6090fccb09ff3749c381137ffaa8eca4b7a82a.

PR #2469, opened 2015-09-30
[node] bump version to 2.0.0-pre.8

bsudekum:
* Fixes an issue which prevented tile rendering on tile that `404`'ed. https://github.com/mapbox/mapbox-gl-native/pull/2458

PR #2471, opened 2015-09-30
Ask contributors to update changelogs incrementally

1ec5:
Along the lines of #2470, we should ask contributors to update the iOS changelog as part of any pull request that significantly affects the iOS SDK. This way, we no longer have to root around for changes when it comes time to release, running the risk of missing a backwards-incompatible change.

For the past few iOS releases, we’ve been maintaining a running changelog in [the relevant milestone’s description](https://github.com/mapbox/mapbox-gl-native/milestones), but it’s easy to forget it’s there.

/cc @mapbox/gl

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,12 @@
 # Changelog
 
+## iOS master
+
+- The SDK now builds with Bitcode enabled. ([#2332](https://github.com/mapbox/mapbox-gl-native/issues/2332))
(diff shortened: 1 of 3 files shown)


friedbunny:
Agreed, this is easier than editing the release milestone (and actually possible for non-admins).

mikemorris:
Let's broaden this to all platforms? Refs https://github.com/mapbox/mapbox-gl-native/pull/2469

1ec5:
@mikemorris: Fixed in 7821b371be2058213c19e96eda6c03a8d0dca69e.

friedbunny:
:heart:

Will start filling that changelog right up.

PR #2477, opened 2015-10-01
Install gtest from mason

jfirebaugh:


kkaefer:
Yup, we should do this.

jfirebaugh:
I imagine if I merge this the warning will return though?

kkaefer:
Probably not, since we're using Google Test's own build system in mason, which sets the correct flags already.

PR #2478, opened 2015-10-01
Refactor MapView code to be more maintainable. Implement JavaDocs up to rotation.

ljbade:
This is the first of several MapView JavaDoc commits for #2002

### DIFF
--- a/android/java/MapboxGLAndroidSDK/build.gradle
+++ b/android/java/MapboxGLAndroidSDK/build.gradle
@@ -75,9 +75,14 @@ android.libraryVariants.all { variant ->
         destinationDir = new File(destinationDir, variant.baseName)
         source = files(variant.javaCompile.source)
         classpath = files(variant.javaCompile.classpath.files) + files(android.bootClasspath)
+        options.windowTitle("Mapbox Android SDK")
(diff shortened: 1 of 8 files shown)


ljbade:
I'm going to keep the branch alive

bleege:
@ljbade Can you clarify why moving the Annotation classes to the `views` package was necessary?

ljbade:
@bleege This was so some public functions in MapView that are only mean for internal use by InfoWindow could be made package private.

bleege:
@ljbade Ah... thanks for the background info.  That said, it still seems to make more sense to have all the Annotations stuff in one package.  I could see a developer using the SDK being confused as to why things were split up.  I also know that MapView and Annotations have a tight working relationship but keeping them separate has the additional benefit of helping us build stronger APIs and codebase.

ljbade:
@bleege In https://github.com/mapbox/mapbox-gl-native/issues/2538 I am proposing merging the annotations and views packages, perhaps just putting them under the base package. Google Maps API seems to happily keep everything in one package.

ljbade:
Actually I figure out how to fix this properly, top offset pixels should be a member of Marker that we pass in addMarker.

PR #2484, opened 2015-10-01
Compile gtest with an OS X deployment target of 10.9

kkaefer:
This will fix the

```
ld: warning: object file (./Release/libgtest.a(gtest-all.o)) was built for newer OSX version (10.11) than being linked (10.9)
```

error


jfirebaugh:
@kkaefer Did you see #2477?

PR #2103, opened 2015-10-01
Split up buffers

kkaefer:
We're currently reusing buffers until they fill up to 65535 vertices. Instead, every logical buffer should have its own buffer rather than sharing them.


kkaefer:
* [ ] Benchmark

kkaefer:
This looks good performance wise; couldn't measure any difference on iPad.

jfirebaugh:
Great, much simpler.

Issue #2490, opened 2015-10-01
allow for iOS annotation.enabled


incanus:
As with MapKit, this prevents selecting certain annotations or showing a callout.

incanus:
I've spent a fair amount of time trying to hook an `enabled` property on the `MGLAnnotationImage` instead of the annotation directly for points, because this is analogous to the way `MKAnnotationView` does it in MapKit, and we are using `MGLAnnotationImage` in its form until we get real, native annotations views. 

But we don't have a proper mapping for annotation IDs (needed under the GL hood) and annotation images (needed for this property), tied together by the annotation object in between. That would require a breaking change to the way `MGLAnnotationImages` objects are instantiated. 

```objc
+ (instancetype)annotationImageWithAnnotation:(id <MGLAnnotation>)annotation
                                        image:(UIImage *)image
                              reuseIdentifier:(NSString *)reuseIdentifier;
```

So instead, I'm going to going to hang the `enabled` property on the annotation directly (per the original ticket title), which is also what we used to do in the raster SDK. In fact, I'm not sure why MapKit does it the other way. You'd have a view instance possibly shared by 2+ annotations and displayed multiple times, but `enabled` only able to be toggled for all of them at once.

incanus:
Actually, now that I think about this more, I like the precedent set by `-[MGLMapViewDelegate mapView:annotationCanShowCallout:]`. In MapKit, such property would be on the annotation view (`-[MKAnnotationView canShowCallout]`), but we put it in the delegate instead for easier deprecation in future. The difference with `enabled` is that many more calls for it will be made, since callouts are only shown singularly and enabled status is needed doing the filtering of multiple selection candidates down to one. But I think this will still work.

incanus:
Ok, one more rethink here. Firing so many delegate callbacks (possibly thousands) every time a tap happens isn't gonna work. 

I'm going back to the original idea, which is building up `MGLAnnotationImage.enabled` in the style of MapKit's annotation views. However, this will only be queried the first time on the delegate, meaning the `enabled` flag can only be decided once and then not adjusted at runtime. 

Eventually, when/if we have annotation views, we'd add an `-[MGLMapView viewForAnnotation:]` (not a delegate method) which just lets the user get at this for future manipulation. But we'll need more re-architecting under the hood before that, and this meets the base need here, which is _a way_ to prevent annotation selection.

incanus:
[...]

PR #2498, opened 2015-10-01
[ios] refactor annotation map table name

incanus:
At some point, the `NSMapTable` we use internal to `MGLMapView` went from being a simple annotation-to-ID mapping and instead is keyed on annotation and maps to a dictionary (currently with keys `MGLAnnotationIDKey` for the ID integer and, optionally, `MGLAnnotationSymbolKey` for the sprite symbol string). 

This just renames this map table from `annotationIDsByAnnotation` to `annotationMetadataByAnnotation`. 

Refs https://github.com/mapbox/mapbox-gl-native/issues/2490


1ec5:
:+1:, although Info would be shorter than Metadata.

Issue #774, opened 2015-01-26
background-image broken


ansis:
adding `"background-image": "wave"` to the bright style makes it look like this:

![screen shot 2015-01-26 at 1 45 24 pm](https://cloud.githubusercontent.com/assets/1421652/5909029/a80cce76-a561-11e4-8131-0a5458dd2ad6.png)

PR #2527, opened 2015-10-05
GL extension use safety mechanism

incanus:
@adam-mapbox I am seeing this work fine on iOS and solve the problem described in https://github.com/mapbox/mapbox-gl-style-spec/issues/365 and seemingly captured in https://github.com/mapbox/mapbox-gl-native/issues/774 a long time ago when `background-pattern` was called `background-image`. 

However, the addition of `glBindVertexArrayOES()` doesn't work on OS X, and may cause problems on other platforms. 

```
../gl-native/src/mbgl/renderer/painter.cpp:422:9: Use of undeclared identifier 'glBindVertexArrayOES'; did you mean 'glBindVertexArrayAPPLE'?
```

Take note of [`vao.cpp`](https://github.com/mapbox/mapbox-gl-native/blob/b406c942df3293f12415d24d03ef17eb44847c88/src/mbgl/geometry/vao.cpp) which references this function. I think we need a way to make this work generically across the platforms. 

/cc @kkaefer

### DIFF
--- a/include/mbgl/platform/gl.hpp
+++ b/include/mbgl/platform/gl.hpp
@@ -7,6 +7,10 @@
 #include <stdexcept>
 #include <vector>
 
+#ifdef GL_TRACK
+#include <iostream>
+#endif
+
 #if __APPLE__
     #include "TargetConditionals.h"
     #if TARGET_OS_IPHONE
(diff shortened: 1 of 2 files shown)


incanus:
Per voice, assuming the Android combine fix in https://github.com/mapbox/mapbox-gl-native/commit/68149f08a7f49a9ae184f7748427b62162703a29 works, let's split this into two atomic changes: 

1. The one-line fix in https://github.com/mapbox/mapbox-gl-native/commit/ec87df4a22de6ff04e25b1d05532e1459f010e84#diff-463080b8a13421555827d7d0edd42421R422 which fixes the root `background-pattern` problem. 

1. The larger change which introduces a system for trapping our string-based calls to GL extensions and warns about a) general classes of problems and b) the specific VAO binding problem captured in the source issue.

incanus:
> The one-line fix

:point_right: https://github.com/mapbox/mapbox-gl-native/pull/2531

Waiting for CI / interactively rebasing this PR here now to address the second item.

incanus:
This looks good now. @adam-mapbox is going to rebase this into a single commit and force-push for @kkaefer review. Can you have a look? Seems sound in my mind, but a little over my head OpenGL-wise.

ljbade:
@incanus @kkaefer i thought we had developed a robust extension system? In the `gl::` namespace that will load the correct OES or APPLE extension for VAOs?

There is even boolean variable you can use with if statements.

ljbade:
Oh I see now this PR lets us pick up typos for GL extensions.

A long time ago I noted we don't carefully check for unsupported extensions but that got cleaned up.

I real gotcha though is extensions that don't add new GL functions but just expand the valid values for existing functions.

fezzik21:
This is a bit more than that.  The point of this is really to catch logic errors that involve the use of extension functions.  In this case, a particular negative pattern of buffer set up, but more generally to track issues involving patterns of calls.

kkaefer:
@adam-mapbox overall changeset looks good. Can you please rebase on master and squash commits into functional units? I.e. every commit must work on its own. That could also mean to squash all commits into one, if there is no intermediate step that keeps the build functional.

incanus:
I think one commit works fine here.

PR #2530, opened 2015-10-05
Look the other way when placing icon at start of line

### LINKED ISSUE #2474: Unrotated one-way arrows along tile boundaries
There’s subtle tearing of streets at what look like tile boundaries, causing many one-way arrows to point in a cardinal direction instead of along the street’s direction of travel.

south

img_0564

img_0565

The [actual OpenStreetMap data](https://www.openstreetmap.org/#map=19/37.35293/-121.90385) is very clean, with no kinks in the roads. The misdirected one-way arrows can be found in both gl-native and gl-js. Yet the gl-native screenshots above show that sometimes the issue occurs away from any tile boundary, depending on zoom level.

This issue reproduces on iOS in v0.5.2 but not v0.5.1. It also occurs in GL JS: mapbox/mapbox-gl-js#1461.

/cc @tmpsantos @brunoabinader @jfirebaugh

1ec5:
This PR fixes a bug exposed by #1840. When an anchor is at the beginning of a line, it is coincident with its segment, resulting in an angle of 0°. In that case, consider the next segment instead.

This change also fixes a rarer bug: for some reason, `clipLines()` inserts a duplicate point at the beginning of the line for [this way](https://www.openstreetmap.org/way/157751098), so we must check whether the anchor is coincident with its segment; we can’t simply check whether `segment` is 0.

before & after

This before and after shows that all the one-way arrows are pointing the right way again. However, collision boxes aren’t being generated for the icons being placed along tile boundaries, so they tend to run right up against icons in adjacent tiles.

Fixes #2474.

/cc @nickidlugash @ansis

[Wrong way (CC BY-SA 3.0 by Wikipedia user Coolcaesar)](https://commons.wikimedia.org/wiki/File:Californiaofframpwrongwaysignage.jpg)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@
 - The SDK now builds with Bitcode enabled. ([#2332](https://github.com/mapbox/mapbox-gl-native/issues/2332))
 - The double-tap-drag gesture for zooming in and out is now consistent with the Google Maps SDK. ([#2153](https://github.com/mapbox/mapbox-gl-native/pull/2153))
 - A new `MGLAnnotationImage.enabled` property allows you to disable touch events on individual annotations. ([#2501](https://github.com/mapbox/mapbox-gl-native/pull/2501))
+- Fixed a rendering issue that caused one-way arrows along tile boundaries to point due east instead of in the direction of travel. ([#2530](https://github.com/mapbox/mapbox-gl-native/pull/2530))
 - Fixed a rendering issue with styles that use the `background-pattern` property. ([#2531](https://github.com/mapbox/mapbox-gl-native/pull/2531))
 
 ## iOS 2.1.2
--- a/src/mbgl/text/quads.cpp
+++ b/src/mbgl/text/quads.cpp
@@ -28,7 +28,13 @@ SymbolQuads getIconQuads(Anchor &anchor, const PositionedIcon &shapedIcon,
     if (alongLine) {
         assert(static_cast<unsigned int>(anchor.segment) < line.size());
         const Coordinate &prev= line[anchor.segment];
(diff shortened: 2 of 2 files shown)


friedbunny:
> However, collision boxes aren’t being generated for the icons being placed along tile boundaries, so they tend to run right up against icons in adjacent tiles.

What's up with that?

1ec5:
The same problem – an anchor coincident with its segment – is tripping up `CollisionFeature::bboxifyLabel()`.

1ec5:
With d9aaefd3e20065df92e0caa20249137c3a662215, there are bounding boxes, but they don’t keep the anchors from running into anchors in other tiles:

bumper boxes

I don’t see a clear way to give each `CollisionFeature` enough context about surrounding tiles to keep them from colliding.

1ec5:
Also, [this line](https://github.com/mapbox/mapbox-gl-native/blob/1179096dd8aff05903edeb7cfe74ef323bfc5709/src/mbgl/text/get_anchors.cpp#L91) doesn’t achieve the intent of the [comment](https://github.com/mapbox/mapbox-gl-native/blob/1179096dd8aff05903edeb7cfe74ef323bfc5709/src/mbgl/text/get_anchors.cpp#L84) above it when the tile is overscaled: “Offset the first anchor by… half the spacing if the line is continued.”

1ec5:
Extraneous one-way arrows are now tracked in #2547.

1ec5:
I reverted the change that added collision boxes because it introduces a latent bug: `index` should never be less than 0 because it’s used to index into an array. The extra collision boxes did nothing anyways, because cross-tile collision isn’t supported yet.

PR #2534, opened 2015-10-06
Correctly handle spurious wake-ups in Map::pause()

### LINKED ISSUE #2522: Spurious wakeups
We're using `std::condition_variable` to ensure `MapContext` pauses. However, [`wait()`](http://en.cppreference.com/w/cpp/thread/condition_variable/wait) sometimes spuriously wakes up without having received an actual notification. In other words, there is no 1:1 mapping between a `notify` call and a `wait` being canceled. We should add a boolean state variable as shown in the example to make sure that we don't unpause too early.

/cc @ljbade @bleege

ljbade:
Fixes #2522

### DIFF
--- a/src/mbgl/map/map.cpp
+++ b/src/mbgl/map/map.cpp
@@ -32,21 +32,20 @@ Map::~Map() {
 void Map::pause() {
     assert(data->mode == MapMode::Continuous);
 
-    if (!paused) {
+    if (!data->paused) {
         std::unique_lock<std::mutex> lockPause(data->mutexPause);
         context->invoke(&MapContext::pause);
-        data->condPaused.wait(lockPause);
-        paused = true;
+        data->condPause.wait(lockPause, [&]{ return data->paused; });
     }
 }
 
 bool Map::isPaused() {
-    return paused;
+    return data->paused;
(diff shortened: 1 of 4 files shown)


ljbade:
This works on my phone.

src/mbgl/map/map.cpp
```diff
@@ -32,21 +32,20 @@ Map::~Map() {
 void Map::pause() {
     assert(data->mode == MapMode::Continuous);
 
-    if (!paused) {
+    if (!data->paused) {
```
kkaefer: access to `paused` needs to be within the lock.

src/mbgl/map/map.cpp
```diff
@@ -32,21 +32,20 @@ Map::~Map() {
...
 
 bool Map::isPaused() {
-    return paused;
+    return data->paused;
 }
 
 void Map::resume() {
-    data->condResume.notify_all();
-    paused = false;
+    data->paused = false;
```
kkaefer: This needs to be protected by a lock.

ljbade:
@kkaefer I will fix the missing locks.

PR #2541, opened 2015-10-06
Fix layer minZoom being ignored if not less than source maxZoom

mourner:
This condition led to layers with high minzoom like oneway street arrows (z15) get parsed and placed regardless of minzoom (e.g. z11+), wasting a lot of worker CPU time.

It became unnecessary after we added reparsing tiles when overscaling, and even before that it should have probably been `<=` rather than `<`.

The PR also removes `maxZoom` TileWorker field which is not necessary anymore.

cc @jfirebaugh @kkaefer


jfirebaugh:
:+1:

PR #2542, opened 2015-10-06
Disable non-null warning

friedbunny:
Silence non-null warning in deprecated `CLLocationManager` method. Cluttered up Travis, [as in here](https://travis-ci.org/mapbox/mapbox-gl-native/jobs/83786613#L939-L941).

```objc
LocationMocker/CLLocationManager+MockLocation.m:35:32: null passed to a callee that requires a non-null argument [-Wnonnull]
  [self.delegate locationManager:self didUpdateToLocation:self.location fromLocation:nil];
```

/cc @1ec5 @mikemorris


1ec5:
:+1:

PR #2561, opened 2015-10-08
[android] Refactor annotations functions

### LINKED ISSUE #2539: Organise annotations function definitions
While working on #2002 I noticed the `MapView` annotation functions need a bit of a tidy up as they do not follow a logical pattern.

E.g. `addMarker` is the only function that takes a `Marker` object directly not a `...Options` object. Important for this API which is supposed to take a mutable object and return a immutable object (#2538 will need this).

I made a pass at fixing this and will push to a branch.

ljbade:
Fixes #2539


ljbade:
@tobrun @bleege Can you check if you are happy with these changes. Particularly the change from selected annotation -> selected marker. I think this makes it clearer that we only consider markers for selection.

ljbade:
No objections. Going to merge after rebase and Travis green light.

PR #2340, opened 2015-09-16
Initial tvOS support

friedbunny:
Requirements to build/run:

- Xcode 7.1 (`sudo xcode-select --switch /Applications/Xcode-beta.app`)

In order to get a map on the screen:

- Disables user location heading (removed from CoreLocation)
- Disables phone carrier metrics key (CoreTelephony unavailable)
- Disables phone connectivity metrics key (ditto)
- Disables push notification check

Thanks to Arne Tempelhof (@webventil) for digging into tvOS immediately!

screen shot 2015-09-15 at 10 42 37 pm

/cc @mapbox/mobile

PR #2581, opened 2015-10-09
Remove CoreTelephony dependency

friedbunny:
Removes the CoreTelephony framework dependency by conditionally using it if already linked.

If a developer chooses to link CoreTelephony in their app, telemetry data on their users' cellular carrier and connectivity will be collected.

Should also help the tvOS situation in #2340, which does not include CoreTelephony at the OS level.

/cc @incanus @1ec5 @twbell @camilleanne @mapbox/mobiledata

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@
 - Fixed a rendering issue with styles that use the `background-pattern` property. ([#2531](https://github.com/mapbox/mapbox-gl-native/pull/2531))
 - Labels can now line wrap on hyphens and other punctuation. ([#2598](https://github.com/mapbox/mapbox-gl-native/pull/2598))
 - A new delegate callback was added for observing taps to annotation callout views. ([#2596](https://github.com/mapbox/mapbox-gl-native/pull/2596))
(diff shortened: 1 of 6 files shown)


1ec5:
> If a developer chooses to link CoreTelephony in their app, telemetry data on their users' cellular carrier and connectivity will be collected.

Should we still call that out as an option in the installation instructions?

friedbunny:
> Should we still call that out as an option in the installation instructions?

Figured we could do this mainly in the `ios-sdk` documentation, but I can add a note in `README.md`, too.

friedbunny:
Ehhhh, actually we don't mention metrics/telemetry anywhere in `INSTALL.md`. I'm inclined to go whole hog or leave it out entirely. Thoughts?

bleege:
FYI...  to make life easier in the future feel free to use the @mapbox/mobiledata team to keep all metrics / telemetry people in the loop.

friedbunny:
This is ready to go, pending me rebasing and Travis giving the green light. I'll hit documentation here with a brief note, as well as separately in `ios-sdk`.

Issue #2588, opened 2015-10-10
Move InfoWindow Classes To annotations Package


bleege:
As discussed https://github.com/mapbox/mapbox-gl-native/pull/2478#issuecomment-145920849 it makes for a more robust API and ease of developer use to have the InfoWindow classes in the `annotations` package instead of the `views` package.

I had gotten started on this in #2448 before deciding that the ticket didn't need addressing now.  I'll update the [original commit message on the branch](https://github.com/mapbox/mapbox-gl-native/commit/b990d31ae3184da903fc5b426ea980f3d85c4cc4) and associate it with this ticket for better tracking.

/cc @ljbade @tobrun

bleege:
Hmm... ran into some local issues [amending the commit message](https://help.github.com/articles/changing-a-commit-message/).  At this stage it's just easier to forget about the [`2448-infowindowontouchlistener`](https://github.com/mapbox/mapbox-gl-native/compare/2448-infowindowontouchlistener) and redo the changes in a new branch.

bleege:
Ok, I just got it redone on `2588-move-infowindow-classes` branch and successfully tested locally.  Will let Travis CI do it's thing and when it's ready to go will merge to `master`.

ljbade:
@bleege I'm going to make another PR that makes `getTopOffsetPixelsForAnnotationSymbol` private by passing it to `Marker` in `addMarker`. This makes more sense in the OOP idea that this should be a property of the `Marker` not the `MapView`.

ljbade:
PR made: https://github.com/mapbox/mapbox-gl-native/pull/2592

ljbade:
Merged.

Issue #2551, opened 2015-10-07
Encapsulate Marker position and sprite


bleege:
As #2535 showed us bad things can happen when field level access to Java objects is allowed.  Let's double back and tighten this up.  In order to do this we're going to have to first [refactor the JNI bindings as described](https://github.com/mapbox/mapbox-gl-native/issues/2535#issuecomment-146243307).

Going forward let's all @mapbox/mobile try to use the stock Java get/set encapsulation construct when possible.  It'll help with automated testing when that rolls around as well as make life a little less dangerous.

ljbade:
@bleege I'll hit moving the other `...Options` classes to use get/set in the PR for #2546. Will keep an eye out for other JNI gremlins too.

ljbade:
Landed in #2592

bleege:
Updated to `android-v2.1.0` as this was done early.

PR #2592, opened 2015-10-12
[android] #2588 - Make getTopOffsetPixelsForAnnotationSymbol private.

ljbade:
Make getTopOffsetPixelsForAnnotationSymbol private

Also:
- add `captures/` to .`gitignore`
- fix `setPosition`
- add missing `getSprite()`
- https://github.com/mapbox/mapbox-gl-native/issues/2551 make `sprite`/`position` `private` now that above bugs are fixed (JNI still can directly access the field)

### DIFF
--- a/android/java/MapboxGLAndroidSDK/src/main/java/com/mapbox/mapboxsdk/annotations/MarkerOptions.java
+++ b/android/java/MapboxGLAndroidSDK/src/main/java/com/mapbox/mapboxsdk/annotations/MarkerOptions.java
@@ -46,7 +46,8 @@ public class MarkerOptions extends AnnotationOptions {
         return ((Marker)annotation).getInfoWindowAnchorV();
     }
 
-    public Marker getMarker() {
+    public Marker getMarker(int topOffsetPixels) {
+        ((Marker)annotation).setTopOffsetPixels(topOffsetPixels);
         return (Marker)annotation;
     }
 
@@ -64,6 +65,10 @@ public class MarkerOptions extends AnnotationOptions {
 
     public String getTitle() {
         return ((Marker)annotation).getTitle();
+    }
+
+    public String getSprite() {
+        return ((Marker)annotation).getSprite();
     }
 
     public MarkerOptions infoWindowAnchor(float u, float v) {
@@ -84,7 +89,7 @@ public class MarkerOptions extends AnnotationOptions {
     }
 
     public MarkerOptions position(LatLng position) {
-        ((Marker)annotation).position = position;
+        ((Marker)annotation).setPosition(position);
         return this;
     }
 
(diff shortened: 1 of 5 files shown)


ljbade:
@bleege @tobrun Can you review.

android/java/MapboxGLAndroidSDK/src/main/java/com/mapbox/mapboxsdk/annotations/MarkerOptions.java
```diff
@@ -84,7 +89,7 @@ public boolean isVisible() {
     }
 
     public MarkerOptions position(LatLng position) {
-        ((Marker)annotation).position = position;
+        ((Marker)annotation).setPosition(position);
```
bleege: This will cause breakages with JNI as described in https://github.com/mapbox/mapbox-gl-native/issues/2551.

android/java/MapboxGLAndroidSDK/src/main/java/com/mapbox/mapboxsdk/annotations/MarkerOptions.java
```diff
@@ -100,7 +105,7 @@ public MarkerOptions snippet(String snippet) {
 
     public MarkerOptions sprite(@Nullable String sprite) {
         if (!TextUtils.isEmpty(sprite)) {
-            ((Marker)annotation).sprite = sprite;
+            ((Marker)annotation).setSprite(sprite);
```
bleege: This will cause breakages with JNI as described in https://github.com/mapbox/mapbox-gl-native/issues/2551.

ljbade:
@bleege That's what I thought, but it didn't seem to break JNI? Also it's not that change that would break, it would be changing the inner fields of Marker, JNI doesn't use the functions.

I tested it both with the `position` and `sprite` fields as package private (current definition) vs private and it made no difference. I think JNI does not obey Java visibility rules and can grab whatever fields it wants to.

ljbade:
Also noting that this fixed a bug in `Marker.setPosition`.

Going to do another test with full `make clean` and rebuild on my phone to double check JNI, then will merge.

ljbade:
Did more testing, JNI does not break. I can add and remove markers fine along with InfoWindows.

ljbade:
Node is failing for some reason. As this is Android only change, merging.

bleege:
@ljbade Was the Node issue related to #2597?

ljbade:
@bleege Yeah looked like it, @mikemorris opened that after I came across it. I did try restarting the job several times keep hitting same node error.

mikemorris:
[Node.js v4.2.1](https://github.com/nodejs/node/releases/tag/v4.2.1) landed today, should be all fixed now @ljbade

ljbade:
@mikemorris Thanks!

PR #2444, opened 2015-10-13
Switch to earcut

jfirebaugh:
We're currently using [libtess2](https://github.com/memononen/libtess2), but the caveat is that it crashes hard on geometries that aren't completely valid. Therefore, we also use [ClipperLib](http://www.angusj.com/delphi/clipper.php) to cleanup geometries before passing them on. This adds quite a bit of code, as well as a huge performance hit for processing tiles.

Instead, we should switch to [earcut.hpp](https://github.com/mapbox/earcut.hpp). Earcut also requires valid geometries, but doesn't crash hard when they aren't. It's also a 2⨉ to 15⨉ speed improvement in tessellation alone, in particular for polygons with low vertex counts, such as buildings. It also allows us to drop ClipperLib.

- [x] [Update earcut.hpp with recent fixes from JS version](https://github.com/mapbox/earcut.hpp/issues/6)
- [x] Make sure that all data sources return valid geometries
- [x] Integrate earcut into GL native

Alternatively, we could drop libtess2 in a first step, but still clean up geometries with ClipperLib.


jfirebaugh:
Instruments Time Profiler results from two scenes, rough average of 3 loads for each, all times are sums over all tiles loaded for the scene. I had to look at slightly different functions due to oddities in how the profiler nested functions (likely due to inlining).

**Mapbox SF @ z15, 6 tiles**

* With libtess, `FillBucket::tesselate` takes 70-80 ms, of which 35-45 ms is libtess and 25 ms is clipper
* With earcut, `FillBucket::addGeometry` takes 12-20 ms, of which 6-7 ms is earcut, 1 ms is ring classification, and the rest varies between runs

**SF @ z7, 9 tiles**

* With libtess, `FillBucket::tesselate` takes 310 ms, of which 190 ms is libtess and 80 ms is clipper
* With earcut, `FillBucket::addGeometry` takes 135 ms, of which 115 is earcut

So as a rough guideline, earcut is at least 2x faster and in the best case (lots of simple polygons like building footprints), 6-7x faster.

lucaswoj:
>  improved polygon validity in our datasources

Is this ticketed somewhere?

mourner:
@jfirebaugh @yhahn a middle option to make the transition easier and get at least half of the gain while we try to tackle the challenge of VT compatibility is to run Clipper + Earcut. In theory this should work for all legacy VT while still shaving off some good ms.

jfirebaugh:
That's an option, although from the numbers above, it would reduce the gain on fill tessellation to about 1.5-2x. Let's see how the source validity work pans out first.

mourner:
@jfirebaugh yeah, I just thought it would be easy enough to break down into multiple steps: 

1. libtess -> earcut
2. VT validity
3. remove clipper

Another advantage of trying clipper+earcut is that for shape annotations to be 100% stable, we'll probably have to run clipper on it since GeoJSON-VT simplification routine can make valid polygons invalid. Although artifacts would be extremely rare for user annotations (unlike crazy water polys in Streets), they're still a possibility we need to take care of.

mikemorris:
Seeing one failing test when rebasing this against `master` 
screen shot 2016-02-11 at 12 44 28 pm

mikemorris:
`API.Offline` test is failing similarly:

**Actual**
actual

**Expected**
expected

jfirebaugh:
The checked-in vector tiles for those tests probably need to be updated to be VTv2 compliant.

jfirebaugh:
Noting here that because we are shipping offline before VTv2 tiles are widespread, we will have to maintain backward compatibility for v1 tiles in the renderer, because offline databases will contain them. This PR will need to be updated to check and branch on the tile version, and retain the clipper and libtess dependencies.

mourner:
@jfirebaugh note that we may get away with just clipper, cleaning up polygons with it for earcut.

src/mbgl/renderer/fill_bucket.cpp
```diff
@@ -9,181 +7,139 @@
...
+    for (std::size_t i = 0; i < len; i++) {
+        double area = signedArea(rings[i]);
+
+        if (area == 0)
+            continue;
 
-    if (lineGroups.empty() || (lineGroups.back()->vertex_length + total_vertex_count > 65535)) {
-        // Move to a new group because the old one can't hold the geometry.
-        lineGroups.emplace_back(std::make_unique<LineGroup>());
+        if (!ccw)
```
mourner: culprit https://github.com/mapbox/mapbox-gl-js/commit/9369d51828d89d2ab0c7c30a792a7d91ce0afaa1

mikemorris:
With the patch for `!ccw` mentioned above, I'm seeing a variation on the failure at https://github.com/mapbox/mapbox-gl-native/pull/2444#issuecomment-182974429

**Actual**
screen shot 2016-02-26 at 3 15 11 pm
**Expected**
screen shot 2016-02-26 at 3 17 46 pm

mourner:
@mikemorris are you sure this test isn't using an outdated VT fixture?

mikemorris:
Pushed up a rebased branch with the patch for `!ccw` to https://github.com/mapbox/mapbox-gl-native/tree/earcut-master, still need to verify the (currently ignored) test failures and possibly update the vector tile fixtures.

jfirebaugh:
✅  Rebased on master, updated to earcut.hpp 0.11, and testing with latest test-suite, which includes V2 tile fixtures.

Remaining work is to add backward compatibility for V1 tiles, however we decide to to it:
* Full existing clipper and libtess pathway which we know is working.
* Clean up with clipper, then pass to earcut. This may require pulling in a custom modified version of clipper? cc @flippmoke @springmeyer

jfirebaugh:
The `Map.Offline` test is failing expectedly, because it tests against V1 tile fixtures. That's great -- when we get this test passing again, we'll know it works with offline database that contain V1 tiles.

springmeyer:
> Clean up with clipper, then pass to earcut. This may require pulling in a custom modified version of clipper? cc @flippmoke @springmeyer

👍  I think we should use earcut everywhere and for v1 tiles they should be detected and "upgraded" by running them through clipper. The latest clipper used in mapnik-vt has fixes which will be important for getting this right and @flippmoke can advise on this.

src/mbgl/tile/geometry_tile.cpp
```diff
@@ -15,6 +17,75 @@ static double signedArea(const GeometryCoordinates& ring) {
...
+        clipper.AddPath(toClipperPath(ring), ClipperLib::ptSubject, true);
+    }
+
+    clipper.AddPath({{
+        {    -util::EXTENT,    -util::EXTENT },
+        { 2 * util::EXTENT,    -util::EXTENT },
+        { 2 * util::EXTENT, 2 * util::EXTENT },
+        {    -util::EXTENT, 2 * util::EXTENT },
+        {    -util::EXTENT,    -util::EXTENT },
+    }}, ClipperLib::ptClip, true);
```
mourner: Do we need to add a clip box here? We're just fixing the polygons that are already clipped right?

mourner:
@jfirebaugh other than the minor comment above, this looks fantastic! 🚢

jfirebaugh:
I'm seeing some performance hot spots on the same complex European landuse polygons that cause issues with libtess2 (https://github.com/mapbox/mapbox-gl-native/issues/4777):

image

mourner:
@jfirebaugh try to isolate a particularly nasty polygon, and I'll see whether there's something in particular that makes the hole bridging routine get stuck for so long.

jfirebaugh:
The problematic polygons are almost certainly the same ones that are causing the problem for libtess in #4777.

I added some debug output for tiles taking longer than 1 second to process with earcut (in debug mode, so worst case):

```
2272ms layer=wood sourceLayer=landuse tileID=6/32/22
3399ms layer=wood sourceLayer=landuse tileID=6/32/23
8310ms layer=wood sourceLayer=landuse tileID=6/34/22
10681ms layer=wood sourceLayer=landuse tileID=6/33/22
5038ms layer=wood sourceLayer=landuse tileID=6/33/21
12985ms layer=wood sourceLayer=landuse tileID=6/34/21
```

https://gist.github.com/jfirebaugh/14c6d1b1aceb1c6f880baf1610e06817 contains the data for the fourth polygon above. It looks like this when tesselated with earcut:

image

jfirebaugh:
gl-js branch that logs slow earcut times: https://github.com/mapbox/mapbox-gl-js/compare/earcut-timing

src/mbgl/tile/geometry_tile.cpp
flippmoke: Instead of adding a clip box we could do a union operations instead?

jfirebaugh: Removed clip box and switched to union.

jfirebaugh:
@mourner 👀  on the latest change?

I think this is good to go -- with the combined Mapbox Streets VT update and ring limiting code, earcut time is down to a reasonable fraction of overall ticks in profiles. (In Europe at z6-8, it's still the dominant processing cost on worker threads, so any further optimizations are welcome.)

src/mbgl/tile/geometry_tile.cpp
```diff
@@ -114,6 +114,18 @@ std::vector<GeometryCollection> classifyRings(const GeometryCollection& rings) {
...
 
+void limitRings(GeometryCollection& polygon, uint32_t maxRings) {
+    if (polygon.size() > 1 + maxRings) {
+        std::nth_element(polygon.begin() + 1,
+                         polygon.begin() + 1 + maxRings,
+                         polygon.end(),
+                         [] (const auto& a, const auto& b) {
+                             return signedArea(a) > signedArea(b);
+                         });
+        polygon.resize(1 + maxRings);
```
mourner: Looks good! The only minor thing is that `maxRings` name implies a limit to a total number of rings, including the outer ring, so you would need to remove a few `+1`s. Or rename to `maxHoles`.

PR #942, opened 2015-10-13
Disable fading when rendering still images

brunoabinader:
When using the static image rendering API, we have to disable fading.

image


mikemorris:
This looks :100: now, thanks for hitting this @brunoabinader!

bsudekum:
To clarify this change:

Labels which which fade while zooming should not fade while a static image is created. This is because if the label is in transition, it could potential never end it's transition and show up opaque as seen above. 

This PR removes fading completely to make sure no label is opaque due to fading at render time.

========

Tested out this PR, tests looked great. @jfirebaugh good to merge this?

jfirebaugh:
:+1:

Issue #2610, opened 2015-10-14
fall back to bundled default marker for MGLAnnotationImage


incanus:
We should make use of the iOS bundled images from https://github.com/mapbox/mapbox-gl-native/pull/2440 and fall back to them when the delegate isn't implemented. Providing `nil` in an existing delegate method works fine (and is presumably what we tested already). 

This should have tests written for it since the test app doesn't hit this use case.

jfirebaugh:
I was under the impression that this was already implemented, but if not: :+1:.

incanus:
This works if you implement the delegate method and return `nil`, but not if you don't implement it.

incanus:
I'm currently working through a local setup caching issue, I think. The GL code looks correct as far as falling back if the delegate isn't implemented or returns `nil` for a point.

incanus:
Yep; all good here, sorry about the false alarm.

PR #2612, opened 2015-10-14
fixes #2611: iOS test app long press drops debug marker

### LINKED ISSUE #2611: add long press to exercise default marker on iOS
Refs https://github.com/mapbox/mapbox-gl-native/issues/2610

incanus:
Note that this uses our already-present delegate method to just return `nil` for these markers. We still need to hit https://github.com/mapbox/mapbox-gl-native/issues/2610 (with tests) for the case of not implementing the delegate method.

### DIFF
--- a/ios/app/MBXViewController.mm
+++ b/ios/app/MBXViewController.mm
@@ -82,6 +82,8 @@ static NSUInteger const kStyleVersion = 8;
                                                                              target:self
                                                                              action:@selector(locateUser)];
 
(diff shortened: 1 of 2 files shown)


incanus:
Looks pretty good. Writing some tests, then will merge this. 

simulator screen shot oct 13 2015 5 52 13 pm

ljbade:
So jealous of the lovely red marker...

incanus:
Tests pass locally; Travis testing unreliability problems. Merging.

PR #2614, opened 2015-10-14
[Android] fix proguard rules

mayardb:
fix proguard rules according to class packaging refactor: from com.mapbox.mapboxgl to com.mapbox.mapboxsdk


tobrun:
Looks good, will merge it in! 
Thank you for the contribution

PR #2609, opened 2015-10-13
Update libuv to 1.7.5

mikemorris:
Update from ancient v0.10.28 across all platforms, but retain support for older versions.

/cc @kkaefer

PR #2622, opened 2015-10-14
[meta] update INSTALL dependencies

friedbunny:
Updates `INSTALL.md` with libuv version recommendation and new project URL. This would be a good opportunity to check that the rest of our listed dependencies are up-to-date @mapbox/gl.

Refs #2609

### DIFF
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -31,14 +31,10 @@ targets.
  - Modern C++ compiler that supports `-std=c++14`
    - clang++ 3.5 or later _or_
    - g++-4.9 or later
- - [Boost headers](http://boost.org/)
- - [`zlib`](http://www.zlib.net/)
- - [`libpng`](http://www.libpng.org/pub/png/libpng.html)
- - [`libuv 0.10+`](https://github.com/joyent/libuv)
- - [`glfw 3.1+`](http://www.glfw.org/)
- - `pkg-config` (for build only)
  - Python 2.x (for build only)
- - [Imagemagick](http://www.imagemagick.org/)
+ - [`pkg-config`](https://wiki.freedesktop.org/www/Software/pkg-config/) (for build only)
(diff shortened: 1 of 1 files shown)


jfirebaugh:
I think we should actually remove the packages that will be installed by mason from this list entirely (boost, libuv, zlib, libpng, glfw). The relevant question for people coming to this page is "What do I need to manually install to be able to run the mapbox-gl-native build?", not "What package versions will the build itself install?"

friedbunny:
That'd be fine, too. Variant and rapidjson are two dependencies that mason installs that aren't on this list, while libtool and automake are both occasionally required if mason doesn't have a prebuilt binary.

ljbade:
Yeah it should just list stuff that our build system doesn't install/download for you.

Did we remove the references to manually installing the Android NDK (not the SDK)

mikemorris:
Gonna revisit this after https://github.com/mapbox/mapbox-gl-native/pull/2800 lands.

mikemorris:
Revised this to drop mason-installed packages Boost, zlib, libpng, libuv and glfw. Added libtool and automake, which may be needed on platforms where mason binaries are not available.

:eyes: @friedbunny

friedbunny:
Thanks @mikemorris, changes looks good. 

@mourner @jfirebaugh Do we need ImageMagick? It's [still installed](https://github.com/mapbox/mapbox-gl-native/blob/aaf090b87dbc96f7a09e783833c10d0b88a16d39/docker/linux/Dockerfile#L5) on Travis, anyway.

jfirebaugh:
ImageMagick is no longer required. Hooray!

friedbunny:
I clobbered @mikemorris' 8cc78b58909dda58b7d0a4b9b310d64aa50f0b7c removal of ImageMagick with a rebased version of the same. Apologies — luckily the result this time was effectively the same as rebasing/squashing. I think we're ready to merge here, @mikemorris?

PR #2626, opened 2015-10-15
MapContext setStyle{URL,JSON} should ignore redundant calls

brunoabinader:
Our engine should be robust enough to ignore redundant calls and avoid triggering a whole style update if the style URL and/or JSON are the same.

/cc @mapbox/mobile @mapbox/gl


kkaefer:
Looks good!

tmpsantos:
:+1:

jfirebaugh:
Heh, I think this will break a particular workflow that @springmeyer was using, but I suppose [XKCD 1172](https://xkcd.com/1172/) applies.

brunoabinader:
Oops :see_no_evil:

Issue #2599, opened 2015-10-13
After switching styles InfoWindow positions not updated


ljbade:
After switching styles the InfoWindow positions not updated if the marker size changes.

How to reproduce:
- add some markers
- touch marker to open InfoWindow
- change from Streets to Emerald

![device-2015-10-13-164015](https://cloud.githubusercontent.com/assets/1215089/10446726/9fdab9b8-71c9-11e5-9315-3d6968b8dcf2.png)
![device-2015-10-13-164043](https://cloud.githubusercontent.com/assets/1215089/10446729/a453aeaa-71c9-11e5-8993-b96691ff4b2a.png)

ljbade:
The fix for this depends on #2263.

After a style finishes loading, we need to loop through all active Markers, update their top offset pixels values, then show/hide the InfoWindow to calculate the new position.

ljbade:
Should /cc @bleege @tobrun

Issue #630, opened 2014-11-23
'load' or 'ready' event


ljbade:
It would be useful to have a way to know when the map has finished loading.

My first use would be as part of the automated Testmunk testing.

Inspired by https://github.com/mapbox/mapbox-gl-js/issues/826

ljbade:
@jfirebaugh I suppose something similar to you JS event would work here?

ljbade:
Ah I see I just need to hook this up.

bleege:
@ljbade Still feeling this is necessary?

ljbade:
Yeah I think we will need this when it comes time to run automated testing.

incanus:
You want https://github.com/mapbox/mapbox-gl-native/blob/2f30dfe97d851f2a2bb4141dc234013b01f8afd3/include/mbgl/map/view.hpp#L26-L28.

bleege:
GTK.  Let's still leave it for `0.2.0` as we need to :ship: `0.1.0` sooner rather than later.

jfirebaugh:
To implement this, keep a flag of whether or not a load event has been previously fired. If, when dispatching the other render events, the flag is not set and `MapContext::isLoaded()` / `Map::isFullyLoaded()` is true, also dispatch the "loaded" event.

Use the [JS implementation](https://github.com/mapbox/mapbox-gl-js/blob/a7092c2d52a9d04fd845cd671b7bcd8192aa5d77/js/ui/map.js#L749-L752) as a reference.

ljbade:
@jfirebaugh Ah excellent, will try that out.

ljbade:
Got it working, thanks @jfirebaugh

bleege:
I'm moving this back to `android-v2.2.0` due to timing.  While this is a useful feature there won't be sufficient enough time to finish building it nor testing it before we need to ship.

ljbade:
Fair enough

ljbade:
After discussion with @jfirebaugh the PR https://github.com/mapbox/mapbox-gl-native/issues/2599 will only implement the start and successful load events.

Failure event needs some work due to interaction with how `setStyleUrl` clears the set JSON but doesn't trigger an `update()`.

So going to leave this issue open until then.

@1ec5 or @incanus: Is there any work needed on iOS side now that the start and success events are hooked up? Documentation?

1ec5:
The methods do need to be documented still so that they show up in the API documentation.

[...]

PR #2627, opened 2015-10-15
#2599 Fix InfoWindow position after style switching

ljbade:
Fixes InfoWindow position after style switching (for #2599)

Also incorporates the #2588 commit that @tobrun reverted with the crash fixed.

As part of this #630 was implemented.

### DIFF
--- a/src/mbgl/map/map_context.cpp
+++ b/src/mbgl/map/map_context.cpp
@@ -92,10 +92,6 @@ void MapContext::triggerUpdate(const TransformState& state, const Update flags) 
 }
 
 void MapContext::setStyleURL(const std::string& url) {
-    if (styleURL == url) {
-        return;
-    }
-
     FileSource* fs = util::ThreadContext::getFileSource();
 
     if (styleRequest) {
@@ -120,15 +116,12 @@ void MapContext::setStyleURL(const std::string& url) {
             loadStyleJSON(res.data, base);
         } else {
             Log::Error(Event::Setup, "loading style failed: %s", res.message.c_str());
+            data.loading = false;
         }
     });
 }
 
 void MapContext::setStyleJSON(const std::string& json, const std::string& base) {
-    if (styleJSON == json) {
-        return;
-    }
-
     styleURL.clear();
     styleJSON = json;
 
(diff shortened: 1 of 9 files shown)


ljbade:
@jfirebaugh Can you review the changes made to core for implementing #630 

@tobrun Can you verify the behaviour of InfoWindow and InfoWindowAdapter. Try changing the style to Emerald and back to see how it now readjusts the InfoWindow position.

ljbade:
@tobrun I think there is a bug with InfoWindow and InfoWindowAdapter activities. It does not seem to load the offset.

Going to need to workout why the map finish loading event is not updating the markers on these activities.

tobrun:
@ljbade 
When testing I noticed that the MainActivity did not give any problems for adding markers/infowindow on a long press. I'm thinking it could be related to adding markers/infowindow while creating the MapView?

ljbade:
@tobrun Yeah it has something to do with adding Marker during onCreate.

ljbade:
@jfirebaugh How do I move that notification back to the main thread, since `setStyleUrl` is async? Is there a function I can use to post an event to the main thread?

ljbade:
@jfirebaugh or should I check for failure in `renderSync`?

jfirebaugh:
I don't think there's a good way to emit `MapChangeDidFailLoadingMap` right now.

ljbade:
@jfirebaugh An idea I had was to add `loadFailed` flag to `MapData` then set this on fail. `renderSync` will then check and clear this flag before it checks the `loading` flag.

ljbade:
I had to move the set for `data.loading` to `loadStyleJSON()` to avoid a false map loaded event right after the map is created (and finishes rendering a blank map) but before it starts loading the new style.

ljbade:
Only Travis failure is a known iOS hang. Other jobs passed.

src/mbgl/map/map_context.cpp
```diff
@@ -92,10 +92,6 @@ void MapContext::triggerUpdate(const TransformState& state, const Update flags)
 }
 
 void MapContext::setStyleURL(const std::string& url) {
-    if (styleURL == url) {
```
jfirebaugh: You're reverting 074d61a55e32e83173ec12b2eb34b78ddaaf0ef1 from earlier today. Accidental?

ljbade:
@jfirebaugh I missed you comment before merging. I used `git cherry-pick 074d61a55e32e83173ec12b2eb34b78ddaaf0ef1` to replay @brunoabinader's commit on master to restore it.

PR #2617, opened 2015-10-15
Request again when obtained data expires

kkaefer:
We're currently treating a request as a one-time thing. We should change this to have "Request" mean "monitor this resource and send me updates about it". This means that the file source needs to keep track of when the resource expires, and issue a new request once it expires. This solves the following cases:

- Tiles with a very short TTL will display stale content, especially in situations where a map view is visible longer (e.g. an iPad in a kiosk mode)
- We obtain a file from cache just before it expires
- We can use the same infrastructure to implement #896

### DIFF
--- a/src/mbgl/map/tile_worker.cpp
+++ b/src/mbgl/map/tile_worker.cpp
@@ -3,6 +3,7 @@
 #include <mbgl/style/style.hpp>
 #include <mbgl/style/style_layer.hpp>
 #include <mbgl/style/property_evaluator.hpp>
+#include <mbgl/geometry/sprite_atlas.hpp>
 #include <mbgl/geometry/glyph_atlas.hpp>
 #include <mbgl/renderer/fill_bucket.hpp>
 #include <mbgl/renderer/line_bucket.hpp>
@@ -10,6 +11,7 @@
 #include <mbgl/renderer/symbol_bucket.hpp>
 #include <mbgl/platform/log.hpp>
 #include <mbgl/util/constants.hpp>
+#include <mbgl/util/exception.hpp>
 
 using namespace mbgl;
 
@@ -33,34 +35,66 @@ TileWorker::~TileWorker() {
     style.glyphAtlas->removeGlyphs(reinterpret_cast<uintptr_t>(this));
 }
 
-Bucket* TileWorker::getBucket(const StyleLayer& layer) const {
-    std::lock_guard<std::mutex> lock(bucketsMutex);
+TileParseResult TileWorker::parseAllLayers(const GeometryTile& geometryTile) {
+    // We're doing a fresh parse of the tile, because the underlying data has changed.
+    pending.clear();
 
-    const auto it = buckets.find(layer.bucket->name);
-    if (it == buckets.end()) {
-        return nullptr;
-    }
-
-    assert(it->second);
-    return it->second.get();
-}
-
-TileParseResult TileWorker::parse(const GeometryTile& geometryTile) {
-    partialParse = false;
+    // We're storing a list of buckets we've parsed to avoid parsing a bucket twice that is
+    // referenced from more than one layer
+    std::set<StyleBucket*> parsed;
 
     for (auto i = layers.rbegin(); i != layers.rend(); i++) {
-        parseLayer(**i, geometryTile);
+        const StyleLayer& layer = **i;
+        if (layer.bucket && parsed.find(layer.bucket.get()) == parsed.end()) {
+            parsed.emplace(layer.bucket.get());
+            parseLayer(layer, geometryTile);
+        }
     }
 
-    return partialParse ? TileData::State::partial : TileData::State::parsed;
+    result.state = pending.empty() ? TileData::State::parsed : TileData::State::partial;
+    return std::move(result);
 }
 
-void TileWorker::redoPlacement(float angle, float pitch, bool collisionDebug) {
+TileParseResult TileWorker::parsePendingLayers() {
+    // Try parsing the remaining layers that we couldn't parse in the first step due to missing
+    // dependencies.
+    for (auto it = pending.begin(); it != pending.end();) {
+        auto& styleBucket = it->first;
+        auto& bucket = it->second;
+        assert(bucket);
+
+        if (styleBucket.type == StyleLayerType::Symbol) {
+            auto symbolBucket = dynamic_cast<SymbolBucket*>(bucket.get());
+            if (!symbolBucket->needsDependencies(*style.glyphStore, *style.sprite)) {
+                symbolBucket->addFeatures(reinterpret_cast<uintptr_t>(this), *style.spriteAtlas,
+                                          *style.glyphAtlas, *style.glyphStore, *collisionTile);
+                insertBucket(styleBucket.name, std::move(bucket));
+                pending.erase(it++);
+                continue;
+            }
+        }
+
+        // Advance the iterator here; we're skipping this when erasing an element from this list.
+        ++it;
+    }
+
+    result.state = pending.empty() ? TileData::State::parsed : TileData::State::partial;
+    return std::move(result);
+}
+
+void TileWorker::redoPlacement(
+    const std::unordered_map<std::string, std::unique_ptr<Bucket>>* buckets,
+    float angle,
+    float pitch,
+    bool collisionDebug) {
+
+    // Reset the collision tile so we have a clean slate; we're placing all features anyway.
     collisionTile = std::make_unique<CollisionTile>(angle, pitch, collisionDebug);
+
     for (auto i = layers.rbegin(); i != layers.rend(); i++) {
-        auto bucket = getBucket(**i);
-        if (bucket) {
-            bucket->placeFeatures(*collisionTile);
+        const auto it = buckets->find((*i)->id);
+        if (it != buckets->end()) {
+            it->second->placeFeatures(*collisionTile);
         }
     }
 }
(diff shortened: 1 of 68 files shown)


kkaefer:
A few notes on the current code: Requests are now monitoring until they're explicitly cancelled. We're also returning expired content (a big part of what #896 is about!). However, none of the stale content, or updating requests are used in code yet, since we immediately `return` for stale responses, and `cancel` requests immediately after we got the first response.

- [ ] Use updated data for re-parsing vector tiles. It's currently possible to just disable the cancellation, so they *will* get updates, however, there's still a slight flicker. Likely this is related to #1688.
- [ ] Be smarter about not notifying requesters when the data didn't change
- [ ] When marking a Response as stale, we should share its content to avoid copying (=> `std::shared_ptr<const std::string>`). Same goes for creating new Response objects for 304 HTTP responses.
- [ ] Figure out what to do when servers respond with a `max-age=0`, or with an `Expires` time that is (consistently) in the past. Ideally we would treat this as `expires = 0`, and not schedule reloads until we get a new explicit request.

kkaefer:
Here's a script to try this out:

Start the server with:
```bash
git clone git@gist.github.com:d74ed2b193bbd1a9507f.git tileserver
cd tileserver
npm install
node index.js
```

Apply this patch to `mapbox-gl-native`:
```diff
diff --git a/macosx/Info.plist b/macosx/Info.plist
index 514db11..f0e5c7f 100644
--- a/macosx/Info.plist
+++ b/macosx/Info.plist
@@ -32,6 +32,11 @@
     <string>0.0.1</string>
     <key>NSHighResolutionCapable</key>
     <true/>
+    <key>NSAppTransportSecurity</key>
+    <dict>
+        <key>NSAllowsArbitraryLoads</key>
+        <true/>
+    </dict>
     <key>CFBundleURLTypes</key>
     <array>
       <dict>
diff --git a/platform/default/default_styles.cpp b/platform/default/default_styles.cpp
index 9ee5414..2382e5c 100644
--- a/platform/default/default_styles.cpp
+++ b/platform/default/default_styles.cpp
@@ -4,6 +4,7 @@ namespace mbgl {
 namespace util {
 
 const std::vector<std::pair<std::string, std::string>> defaultStyles = {
+    { "http://localhost:3000/style.json", "Test" },
     { "asset://styles/streets-v8.json", "Mapbox Streets" },
     { "asset://styles/emerald-v8.json", "Emerald" },
     { "asset://styles/light-v8.json", "Light" },

```

Then, start the `osxapp` application and you will see this:

out

The server sends out tiles with a `max-age=1`, which contains a line from (0,0) to to the right bottom, with the length depending on the time.

src/mbgl/storage/request_holder.hpp
```diff
@@ -0,0 +1,26 @@
+#ifndef MBGL_STORAGE_REQUEST_HOLDER
+#define MBGL_STORAGE_REQUEST_HOLDER
+
+#include <memory>
+
+namespace mbgl {
+
+class Request;
+
+class RequestHolder {
```
jfirebaugh: What's the reason for a separate `RequestHolder` class, rather than the usual pattern of returning a `std::unique_ptr<Request>` and having `~Request` auto-cancel?

kkaefer: I implemented this at first, but backtracked because the pull request grew too big due to too many changes necessary for the unit tests. I'll try again.

jfirebaugh: Feel free to ticket and queue up for a later PR if you want.

ljbade:
@kkaefer Did a quick test on Android. No problems found. And MapView is correctly GC'd. 

:+1:

bleege:
Looks good @kkaefer

src/mbgl/map/tile_worker.cpp
```diff
@@ -87,21 +113,15 @@ void TileWorker::parseLayer(const StyleLayer& layer, const GeometryTile& geometr
         return;
     }
 
-    // This is a singular layer. Check if this bucket already exists.
-    if (getBucket(layer))
```
jfirebaugh: What's happening here? I believe this check is what makes the layer `ref` mechanism efficient, by avoiding creating duplicate buckets for the referent layers.

src/mbgl/map/tile_worker.cpp
```diff
@@ -32,34 +34,58 @@ TileWorker::~TileWorker() {
...
     }
 
-    return partialParse ? TileData::State::partial : TileData::State::parsed;
+    result.state = pending.empty() ? TileData::State::parsed : TileData::State::partial;
+    return std::move(result);
 }
 
-void TileWorker::redoPlacement(float angle, float pitch, bool collisionDebug) {
+void TileWorker::redoPlacement(
+    const std::unordered_map<std::string, std::unique_ptr<Bucket>>* buckets,
```
jfirebaugh: This map is being used by two threads (map context and worker). What's the synchronization strategy?

src/mbgl/map/tile_worker.cpp
kkaefer: That used to be the case, but the parsing is now split up into two distinct step: The initial bucket creation, and the actual parsing. We're no longer creating a new bucket just to throw away that object if we realize we don't have the resources yet. Instead, we're creating all bucket objects just  once, then try to parse. If that doesn't work just yet, we're saving the bucket for later. All later "parse" steps will call `parsePendingLayers`, which just processes the buckets we saved for later.

src/mbgl/map/tile_worker.cpp
kkaefer: You're right. In this case, we're only reading from the map in the worker thread (and the map thread), which doesn't need synchronization. The only time we're writing to the map is in the map thread in the work task *callback*. This means there can't be any reading going on in the worker thread (since there's no work task in progress).

src/mbgl/map/tile_worker.cpp
jfirebaugh: Where does `ref` deduping happen, exactly? I'm still not seeing that.

src/mbgl/map/vector_tile_data.cpp
```diff
@@ -95,39 +139,45 @@ bool VectorTileData::reparse(std::function<void()> callback) {
...
 }
 
 void VectorTileData::redoPlacement(float angle, float pitch, bool collisionDebug) {
-    if (angle == currentAngle &&
-        pitch == currentPitch &&
-        collisionDebug == currentCollisionDebug)
+    if (angle == currentAngle && pitch == currentPitch && collisionDebug == currentCollisionDebug)
+        return;
+
+    if (workRequest) {
```
jfirebaugh: Shouldn't this check should go _after_ the assignments to `lastAngle`, etc, so that when the callback calls `redoPlacement` again it uses the most recent values?

src/mbgl/map/tile_worker.cpp
jfirebaugh: That makes sense, and I couldn't think of any data races given how the worker request cancellation mechanism works. One thing that would make the reasoning a little bit clearer would be to adjust `VectorTileData::parse` so that it first explicitly does `workRequest.reset()`, and then does `workRequest = worker.parseVectorTile(...)`. (Right now, it's possible for another work task from the tile to continue to run briefly while the `parseVectorTile` request kicks off, but before the assignment to `workRequest` results in the previous task being cancelled.)

src/mbgl/map/tile_worker.cpp
kkaefer: Oops you're right. Will add that back.

src/mbgl/map/tile_worker.cpp
kkaefer: Good point about resetting first.

src/mbgl/map/vector_tile_data.cpp
kkaefer: Yup, cleaned this up.

src/mbgl/map/tile_worker.cpp
kkaefer: We should probably refactor the Worker objects to enforce that they're cleared before we create a new task?

src/mbgl/map/tile_worker.cpp
kkaefer: This is now done before even calling `parseLayer()`.

src/mbgl/map/tile_worker.cpp
```diff
@@ -161,101 +179,114 @@ void TileWorker::addBucketGeometries(Bucket& bucket, const GeometryTileLayer& la
...
+                            *style.glyphAtlas, *style.glyphStore, *collisionTile);
+        insertBucket(styleBucket.name, std::move(bucket));
     }
+}
 
-    // We do not proceed if the parser is in a "partial" state because
-    // the layer ordering needs to be respected when calculating text
-    // collisions. Although, at this point, we requested all the resources
-    // needed by this tile.
-    if (partialParse) {
```
jfirebaugh: Does this change how partial parsing works? Previously, because `partialParse` was a member variable, if _any_ symbol layer needed dependencies, all subsequent symbol buckets would be treated as partial. Now it looks like only a bucket that `needsDependencies` is treated as partial. From my reading of the above comment, the previous behavior was intentional.

kkaefer: We can create new buckets in an arbitrary order; the only place where order matters is when calling `SymbolBucket::placeFeatures`. In the current scenario it's indeed possible that we're placing buckets in the wrong order. Will create a pull request that fixes this.

kkaefer: See https://github.com/mapbox/mapbox-gl-native/issues/2820

Issue #1947, opened 2015-07-28
Implement styleId, improve styleUrl


ljbade:
Noticed Android does not have `styleId`.

`styleUrl` also needs some changes:
- allow `null` to choose default map style
- additional validation to ensure any `asset://` URLs are only in the list of `bundledStyleUrls`

Add `bundledStyleUrls`

1ec5:
It's styleID and styleURL now on the iOS side.

bleege:
Bumping to `0.2.0`

ljbade:
I'm not going to implement styleID on Android. Instead I might add a static function that formats a mapbox URL or just document it well in #2002.

Also going to look at adding static strings for the bundled styles.

bleege:
@ljbade Can you explain why you decided not to implement styleID?

ljbade:
@bleege It is redundant. `setStyleUrl` is a superset of `setStyleId`. I have carefully documented in JavaDoc that `mapbox://styles/user/style` is the correct way to load a custom Studio style. It is also the URL Studio gives you to copy paste into your app/webpage. Finally this provides parity with JS API (if I understand @jfirebaugh).

Also @1ec5 mentioned that `setStyleId` is a workaround for Xcode which can't handle NSURL functions in Interface Builder and that he was contemplating moving it to the IB only header (ie deprecating it for code use).

bleege:
@ljbade Cool!  Thanks for explaining.

PR #2632, opened 2015-10-15
Replace styleID property with styleURL__ inspectable

### LINKED ISSUE #2628: Remove styleID property; replace with style URL inspectable
I'm trying to use a custom map in the iOS SDK. I have tried typing out the styleID in interface builder like this:
interfacebuilder image

But that gives me the error message: `[ERROR] {Map}[Setup]: loading style failed: HTTP status code 404`

I have tried setting it in code too with the same error and tried numerous different styles.

1ec5:
This PR replaces the `styleID` convenience property with an Interface Builder inspectable (normally inaccessible from code) that imitates a URL field. `styleID` is inconsistent with other GL-based APIs, and we never implemented a convenience initializer that took a `styleID` anyways, so it had limited usefulness as a separate property.

The new inspectable’s name is a horrible hack and is marked as such. I’ve filed <[rdar://problem/23134321](http://openradar.appspot.com/radar?id=6123837150199808)> against Interface Builder to request support for NSURL-typed inspectables.

I considered merely deprecating `styleID` but wound up cutting over in a manner similar to #1561. `styleID` is primarily used as an inspectable. Deprecating an inspectable leads to a horrible user experience, because the inspectable’s value continues to work even though it silently disappears from the Attributes inspector. (You’d have to know to go to the User Defined Runtime Attributes section of the Identity inspector.) If you ever tried to re-set the style, this time via the new “Style URL” inspectable, you might run into weirdness like #2626. **This backwards-incompatible change requires bumping the version up to 3.0.0.**

This PR also removes some APIs that have been marked unavailable for a few releases, since back before the iOS SDK left the 0.x series.

Fixes #2628.

/cc @jfirebaugh @incanus @bsudekum @friedbunny

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## iOS master
 
+- The `styleID` property has been removed from MGLMapView. Instead, set the `styleURL` property to an NSURL in the form `mapbox://styles/STYLE_ID`. If you previously set the style ID in Interface Builder’s Attributes inspector, delete the `styleID` entry from the User Defined Runtime Attributes section of the Identity inspector, then set the new “Style URL” inspectable to a value in the form `mapbox://styles/STYLE_ID`. ([#2632](https://github.com/mapbox/mapbox-gl-native/pull/2632)
 - The SDK now builds with Bitcode enabled. ([#2332](https://github.com/mapbox/mapbox-gl-native/issues/2332))
 - The double-tap-drag gesture for zooming in and out is now consistent with the Google Maps SDK. ([#2153](https://github.com/mapbox/mapbox-gl-native/pull/2153))
 - A new `MGLAnnotationImage.enabled` property allows you to disable touch events on individual annotations. ([#2501](https://github.com/mapbox/mapbox-gl-native/pull/2501))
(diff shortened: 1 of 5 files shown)


1ec5:
Before merging, I need to address invalid URLs by asserting.

friedbunny:
I'm on the record as being not un-in-favor of huge version numbers. Will give this a spin tomorrow!

friedbunny:
Looks good and works in practice. :+1:

incanus:
Question: we are considering bumping to `3.x.x` for this, correct? Is there _any_ way we can _not_ do that (just checking)?

1ec5:
As explained above, we could have a non-inspectable property with the old name, as a compatibility shim. However, that’s no way to deprecate a symbol: developers will never see any indication that `styleID` has gone away. If they even have a reason to look at MGLMapView’s inspectables, all they’ll see is that there’s a new “Style URL” field that’s blank. If they set that inspectable (since it was never their intention to use the default style), the result would be potential problems like #2626 (if the URL matches the hidden ID) or undefined behavior (if it doesn’t).

This is a unique case, and I think the set of inspectables we’ll have after landing this change will be pretty stable for awhile, modulo any Interface Builder enhancements to their inspectable support.

tmcw:
Is this change going to be reflected in the Android SDK?

1ec5:
The Android SDK has never had an independent `styleID` getter or setter: https://github.com/mapbox/mapbox-gl-native/issues/1947#issuecomment-144886626.

PR #2635, opened 2015-10-16
[android] Remove R and BuildType from JavaDoc. Remove iOS image.

ljbade:
For #2002


ljbade:
iOS failing, but there is currently a known unstability in iOS tests. As this is Android only change, merging.

PR #2637, opened 2015-10-16
[android] Refactor annotations API

### LINKED ISSUE #2448: Add setOnInfoWindowClickListener()
Needed for parity with Google Maps, they expose a method setOnInfoWindowClickListener on MapView, Our current implementation uses a method directly on Marker:

```java
  marker.setInfoWindowOnTouchListener(new View.OnTouchListener() {
            @Override
            public boolean onTouch(View view, MotionEvent motionEvent) {
                Toast.makeText(getApplicationContext(), "Custom Info Touch Listener!!", Toast.LENGTH_SHORT).show();
                marker.hideInfoWindow();
                return true;
            }
        });
```

### LINKED ISSUE #2546: Make Annotation objects immutable
In the various annotation objects (`Marker` etc) we have a number of `set...` functions. These need to be removed so that they are immutable for the moment.

Main reason is the expected behaviour from developers is that for example `Marker.setColor()` will change the existing marker's colour. Instead nothing happens.

Full support for mutable annotations is not in the core layer at this stage. Even on iOS you can't update existing markers.

@bleege I think this is good candidate for 2.1.0

### LINKED ISSUE #2631: AnnotationOptions.alpha Needs To Return SuperType
Currently `AnnotationOptions.alpha()` [returns an `AnnotationsOptions` object](https://github.com/mapbox/mapbox-gl-native/blob/074d61a55e32e83173ec12b2eb34b78ddaaf0ef1/android/java/MapboxGLAndroidSDK/src/main/java/com/mapbox/mapboxsdk/annotations/AnnotationOptions.java#L9).  While this is functional, it requires casting to be applied in common Builder patterns which isn't good.  For example:

```java
MapView map = ...;

// works
map.addPolygon(new PolygonOptions().add(latLngs));

// currently doesn't work due to typing, but is ideal
map.addPolygon(new PolygonOptions().add(latLngs).alpha(0.5f));

// casting needed to make work, but not something that should be used 
map.addPolygon((PolygonOptions) new PolygonOptions().add(latLngs).alpha(0.5f));
```

/cc @tobrun @ljbade

ljbade:
Remove unimplemented properties.
Make annotations immutable.
Correct defintions of equals() and hasCode().
Change anchor U/V to int.
Move .alpha to MultiPoint and anchor() to PolylineOptions and PolygonOptions.
Make InfoWindow classes package private.
Add setOnInfoWindowClickListener and remove old method from Marker.
JavaDoc internal methods with "Do not use."
Refactor showInfoWindow() to remove need for exposing internal method.
Make select/deselectMarker public. Add getSelectedMarker.
Fix bug where you couldn't reselect a closed info window.
Fixes #2546
Fixes #2448


ljbade:
There is a lot going on in this commit, but the result is an annotations API that only exposes what we have implemented. It also improves parts of the code, and fixes a bug that had not been ticketed before.

I will aim to get this merged tonight, ready for Sirius build tomorrow.

I have carefully tested everything with the TestApp and there should be no new bugs introduced.

/cc @bleege @tobrun

ljbade:
The only Travis failure was the known hang in the iOS tests. Restarting that job, if it still fails we can merge as this is an Android only change.

tobrun:
@ljbade 
When doing a couple of test runs I was able to produce a crash.
I'm not able to reproduce it anymore, I believe I was too fast to press a marker before the infowindow was ready to handle the click event. This could well be a crash related to my code an not this PR. If you haven't picked this up tonight I will later this day.

> java.lang.NullPointerException: Attempt to invoke virtual method 'void com.mapbox.mapboxsdk.annotations.Marker.showInfoWindow()' on a null object reference
    at com.mapbox.mapboxsdk.views.MapView.adjustTopOffsetPixels(MapView.java:1978)
    at com.mapbox.mapboxsdk.views.MapView.access$500(MapView.java:98)
    at com.mapbox.mapboxsdk.views.MapView$1.onMapChanged(MapView.java:828)
    at com.mapbox.mapboxsdk.views.MapView$3.run(MapView.java:2823)
    at android.os.Handler.handleCallback(Handler.java:739)
    at android.os.Handler.dispatchMessage(Handler.java:95)
    at android.os.Looper.loop(Looper.java:148)
    at android.app.ActivityThread.main(ActivityThread.java:5417)
    at java.lang.reflect.Method.invoke(Native Method)
    at com.android.internal.os.ZygoteInit$MethodAndArgsCaller.run(ZygoteInit.java:726)
    at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:616)

ljbade:
@tobrun That is a strange crash, I can't see how it gets a `null` since it checks for that specifically:
```
if (mSelectedMarker != null) {
            if (mSelectedMarker.isInfoWindowShown()) {
                mSelectedMarker.hideInfoWindow();
                mSelectedMarker.showInfoWindow();
            }
        }
```

ljbade:
Ah might be race condition between the UI thread and the main thread. Map change event comes from main thread, other changes such as touch input come from UI thread.

ljbade:
I think this crash is from changes in https://github.com/mapbox/mapbox-gl-native/pull/2627

ljbade:
No doesn't look like map change problem since we already forward them to the UI thread.

Something must be touching MapView from a non-UI thread when it shouldn't.

@tobrun If you can't reproduce again, I think we can chalk it up to your device playing weird. And merge.

tobrun:
> @tobrun If you can't reproduce again, I think we can chalk it up to your device playing weird. And merge.
:+1:

bleege:
@ljbade This is a _LARGE_ change so close to launch.  I'm not comfortable merging this into `master` right now.  Let's bump this to `android-v2.2.0` to give us more time.

ljbade:
Sure, should we quickly add some do not use or not implemented javadoc to the annotation functions that don't do anything for 2.1.0 pending there removal?

bleege:
> should we quickly add some do not use or not implemented javadoc to the annotation functions that don't do anything for 2.1.0 pending there removal?

I was under the impression that these were going to be made `private` access?  If they haven't already then let's do that.

ljbade:
@bleege OK I'll make a version of this PR that includes just those changes (make non-functional methods private). I can't make all of the internal only methods private though (that was a big part of this refactor - to make it possible to make some of them private).

I'll include the `.alpha` fix too.

ljbade:
@bleege The other thing I've been puzzling about is how Google keep the annotation classes in the `map.model` package which is separate from the `GoogleMap` object in `map`. They manage to do this without exposing any internal APIs.

Do they perhaps strip internal functions from the SDK before the publish? This is how they do it in the Android framework.

ljbade:
Finally got this branch up to date with v2.1.0 merge.

Waiting for Travis.

ljbade:
Only failing on iOS due to hang in location tests. Merging.

PR #2643, opened 2015-10-16
[ios] fix user annotation view refresh when switched from course to n…

muZZkat:
Fixes a issue where the User Annotation was not updated to correctly reflect the non-tracking mode when switching from "course". The view would remain looking like the course puck until the next location update. There was also a issue in that the heading layer was incorrectly added.


friedbunny:
Thanks for the contribution, @muZZkat — this looks solid.

Because of our fast-forward merge strategy (no green button pushing!), I've rebased this and pushed up to a local branch: [`user-tracking-mode-refresh-2643`](https://github.com/mapbox/mapbox-gl-native/tree/user-tracking-mode-refresh-2643). Once the tests pass, I'll merge.

friedbunny:
Committed as f8c75c6e229c3e3cc8fa28a63656b599be60b158 and f6e3ee4b25eb4472d8d7b9408326c62532925cc1.

PR #2651, opened 2015-10-17
[android] Tidy up annotations API

ljbade:
Make unimplemented properties package private.
Make annotations object properties immutable.
Make InfoWindow and Circle classes package private.
Mininal version of #2546 suitable for v2.1.0 release.

### DIFF
--- a/android/java/MapboxGLAndroidSDK/src/main/java/com/mapbox/mapboxsdk/annotations/Annotation.java
+++ b/android/java/MapboxGLAndroidSDK/src/main/java/com/mapbox/mapboxsdk/annotations/Annotation.java
@@ -9,23 +9,26 @@ public abstract class Annotation implements Comparable<Annotation> {
      *
      * Internal C++ id is stored as unsigned int.
      */
-    protected long id = -1; // -1 unless added to a MapView
-    protected MapView mapView;
+    private long id = -1; // -1 unless added to a MapView
+    private MapView mapView;
 
-    float alpha = 1.0f;
-    boolean visible = true;
+    private float alpha = 1.0f;
+    private boolean visible = true;
 
-    public Annotation() {}
+    protected Annotation() {}
 
     public float getAlpha() {
         return alpha;
     }
 
+    /**
+     * Do not use this method. Used internally by the SDK.
+     */
     public long getId() {
         return id;
     }
 
-    public boolean isVisible() {
+    boolean isVisible() {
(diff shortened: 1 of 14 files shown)


android/java/MapboxGLAndroidSDK/src/main/java/com/mapbox/mapboxsdk/annotations/Annotation.java
```diff
@@ -12,20 +12,23 @@
...
 
+    /**
+     * Do not use this method. Used internally by the SDK.
+     */
     public long getId() {
         return id;
     }
 
-    public boolean isVisible() {
+    boolean isVisible() {
```
bleege: Why remove `public`?

android/java/MapboxGLAndroidSDK/src/main/java/com/mapbox/mapboxsdk/annotations/Annotation.java
```diff
@@ -34,19 +37,25 @@ public void remove() {
         mapView.removeAnnotation(this);
     }
 
-    public void setAlpha(float alpha) {
+    void setAlpha(float alpha) {
```
bleege: Why remove `public`?

android/java/MapboxGLAndroidSDK/src/main/java/com/mapbox/mapboxsdk/annotations/Annotation.java
```diff
@@ -34,19 +37,25 @@ public void remove() {
...
 
+    /**
+     * Do not use this method. Used internally by the SDK.
+     */
     public void setMapView(MapView mapView) {
         this.mapView = mapView;
     }
 
-    public void setVisible(boolean visible) {
+    void setVisible(boolean visible) {
```
bleege: Why remove `public`?

ljbade:
@bleege Core GL does not currently implement the `visible` or `isVisible` properties.
`setAlpha` on an existing `Polyline`/`Polygon` object will not update the marker already on the map and will confuse developers. It is only needed for the builder pattern in `PolylineOptions`/`PolygonOptions`.

Also I just remembered that `Marker` does not support `alpha` at all in Core GL. Will remove from `MarkerOptions`.

ljbade:
Also made methods `private` instead of package private where I could.

PR #2316, opened 2015-10-17
Mapbox GL and the GL state

brunoabinader:
We should be able to integrate Mapbox GL with other applications drawing using GL in the same GL context. That said, in order to play nice with a shared GL state we should:

- Not make any assumptions of the current GL state and reset it to our expectations before painting.
- Save the current state (scoped to things we might change) before painting and restore after painting is done so the client application won't be affected by us.
- Our CPU side cache is sometimes incorrect because the cached state might be changed by the client application after we paint. So in the next paint call, we will use incorrect GL capabilities.
- We don't always clear the background, so if the client application for some reason clears, we might see funny results. Obviously we should always observe our fill rate.

/cc @kkaefer

### DIFF
--- a/include/mbgl/map/mode.hpp
+++ b/include/mbgl/map/mode.hpp
@@ -10,6 +10,11 @@ enum class MapMode : uint8_t {
     Still, // a once-off still image
 };
 
-}
+enum class GLContextMode : uint8_t {
+    Unique,
+    Shared,
+};
 
-#endif
+} // namespace mbgl
+
+#endif // MBGL_MAP_MODE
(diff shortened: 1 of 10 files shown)


include/mbgl/map/mode.hpp
```diff
@@ -10,6 +10,11 @@ enum class MapMode : uint8_t {
     Still, // a once-off still image
 };
 
-}
+enum class GLContextMode : uint8_t {
```
jfirebaugh: Can you add a comment here about what the difference in behavior is between the two modes.

Issue #2049, opened 2015-08-12
Implement Responsive User Location Tracking Mode


bleege:
The User Location tracking mode (None, Follow, Bearing) was disabled in #1856 as the User Location dot was being updated on the main UI thread.  The core logic is there, but it needs to be implemented in a more efficient way so that the UI remains responsive.

/cc @erf

erf:
Could it be added as a annotation?

ljbade:
@bleege Is the issue that there is too many events?

erf:
i made a lowpass filter in  #1968 which could be used

erf:
in addition to some minimal time / angle diff

erf:
I think you should move out the setDirection call from updateMap to the onCompassChange event.

bleege:
> @bleege Is the issue that there is too many events?

Yes.

ljbade:
@bleege How does iOS handle that?

1ec5:
@ljbade, on iOS, MGLMapView doesn’t do anything special. But it seems that Core Location is doing a better job of coalescing location updates than what we’re using on Android: #1676.

bleege:
> But it seems that Core Location is doing a better job of coalescing location updates than what we’re using on Android

Exactly.  We need to put a regulator / distance filter on updates to location or specifically that a certain amount of distance / time needs to elapse before updating the dot on the map.

ljbade:
@bleege I recall the LOST framework has this ability to filter (certainly Google Play services do).

We still need to work in @erf's compass filter for that data.

ljbade:
I was getting confused between location/GPS events and map position change events.

ljbade:
In #2409 I'm cleaning up the `updateMap()` code so this should be unblocked by performance issues now.

mikemorris:
Have you tried sampling input/direction (at a constant rate of 15 times per second or so) instead of trying to process or filter all events? I've done this in the past for handling user input from a gyroscope and accelerometer in JavaScript, where it was previously choking on large bursts of minor change events.

Sample code at https://github.com/mikemorris/gyrocopter/blob/master/gyrocopter.js

ljbade:
Hmm I might have accidentally removed the user tracking stuff in #2002, but the code is still there in git history so perhaps we can have a go at making it really good this time.

Lets start with just location/GPS (and heading?) tracking and work on compass after.

Should we also hit the UI side of getting our GPS dot looking better?

[...]

PR #2670, opened 2015-10-19
[android] Render at 60fps

### LINKED ISSUE #2411: Make map more fluid
Even after #2409 there is still some work to be done.

Google Maps still feels much more fluid in response to gestures and animations.

#2410 should improve animations, but gestures are still a bit janky.

I don't think this is a GL performance issue as it seems we can easily hit 140 FPS on my Samsung S6.

This is more about avoiding any frame skipping and improving the lag between touch input and pixels on the screen.

There is also still a very small 1-2 frame lag between the map and the GPS marker position,

### LINKED ISSUE #2668: Update GPS marker after renderSync
While investigating why the GPS marker does not sync with the map in https://github.com/mapbox/mapbox-gl-native/issues/2049 I notice the order we update the GPS marker and call `renderSync` is reversed from iOS.

ljbade:
Fixes #2668
Fixes #2411

### DIFF
--- a/android/cpp/jni.cpp
+++ b/android/cpp/jni.cpp
@@ -481,7 +481,7 @@ void JNICALL nativeRenderSync(JNIEnv *env, jobject obj, jlong nativeMapViewPtr) 
     mbgl::Log::Debug(mbgl::Event::JNI, "nativeRenderSync");
     assert(nativeMapViewPtr != 0);
     NativeMapView *nativeMapView = reinterpret_cast<NativeMapView *>(nativeMapViewPtr);
-    nativeMapView->renderSync();
+    nativeMapView->getMap().renderSync();
 }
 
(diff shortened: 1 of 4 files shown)


ljbade:
Result of some investigation into marker sync issues #2049 

Improves current frame rate from 30fps to 60dps. Makes a noticeable difference to responsiveness and gestures.

Still have not yet reached perfect GPS marker sync, but this PR makes it a lot less jumpy.

ljbade:
@tobrun @bleege Do you mind giving this PR a spin to see if it makes the map smoother on your devices?

ljbade:
@tobrun I commented out the animation code for marker the moment as I was not sure how it worked, however if you think we need to hook it back up feel free to push a commit.

ljbade:
Hmm finding the occasional hang. Might have to hold this PR to 2.2.0 as I think there is still a lot of work needed to get marker sync working.

ljbade:
Closing in favour of `2668-render-sync` branch.

Issue #2082, opened 2015-08-14
Support selection of shape and polyline annotations


friedbunny:
We explicitly don't support selecting non-point annotations in iOS (820d2da203b184fbe369f3e5beb3d68a51c8a1f1), so polylines and polygons aren't particularly interactive yet.

Related: #1504

/cc @incanus @1ec5

friedbunny:
I'm unsure where Android stands on annotation selection — @ljbade @bleege?

incanus:
General Android annotation selection has been on my plate over in https://github.com/mapbox/mapbox-gl-native/issues/1969 but is incomplete. 

As for shapes, I don't think MapKit even supports this, or if it does, I know for a fact it doesn't support callouts on them. We hit this with the raster SDK and it was due to difficult callout placement algorithms. But we could try here?

1ec5:
This would be nice for gl-js parity.

lucaswoj:
The way this is done in JS is with 

 - runtime styling API
 - `featuresAt` API

ljbade:
You don't necessarily need to support callouts.

Perhaps just a way to select/hit test polygons.

maciekish:
We definitely need callout support as well as selection.

Zakay:
+1 for callout support as well.

incanus:
Would it be acceptable to require the developer to indicate where on a polyline/polygon the callout should attach? That would take the guesswork off of the framework. We could support some simple options like top/bottom/left/right/centroid and leave it at that. 

The problem is when you get into polylines like this — where do you attach the callout? 

![file sep 14 8 36 35 am](https://cloud.githubusercontent.com/assets/17722/9854077/c74d92ce-5abb-11e5-9ccd-b6398cb60f6f.jpeg)

Prior art: https://github.com/mapbox/mapbox-ios-sdk-legacy/issues/285

maciekish:
Definiteyly, that's a lot better than not being able to do it at all!

mb12:
@incanus  The callout for line can be attached to a 'point on the line'  that passed the hit-test for tap.

friedbunny:
> The callout for line can be attached to a 'point on the line' that passed the hit-test for tap.

Aye, tapped-point would be the most intuitive solution for polyline callouts. Polygons, seems like centroid would be best?

maciekish:
@friedbunny Please leave an option to set any point or tapped point on the polygon as well. Centroid won't always work in our use case.

[...]

Issue #1708, opened 2015-06-10
proposal: shape layer grouping API


incanus:
If we proceed with this, it will also happen on Android. 

Per voice with @kkaefer, our shape annotation implementation in #1655 could be improved by creating a layer grouping API identically-styled shapes. 

Currently, each shape annotation gets its own style layer for reasons of: 

1. Eventual z-ordering by the client, much like MapKit's [`insertOverlay:aboveOverlay:`](https://developer.apple.com/library/ios/documentation/MapKit/Reference/MKMapView_Class/index.html#//apple_ref/doc/uid/TP40008205-CH3-SW71). 
1. Styling, as shapes in the same layer get the same styling. 

However, performance is a concern here. Adding 10,000 similarly-styled shapes means 10,000 more style layers, which means 10,000 more GL draw calls. :rotating_light: **NOT TENABLE** :rotating_light: 

If we introduced the concept of a _shape group_ for identically-styled shapes, a consumer could add 10,000 blue polylines to the map and incur 1 style layer and 1 draw call. 

Instead of: 

```swift
mapView.addAnnotation(polyline1)
mapView.addAnnotation(polyline2)
// ...
mapView.addAnnotation(polyline10000)
```

It would look something like: 

```swift
let polylineGroup = MGLMultiPointGroup
polylineGroup.addObject(polyline1)
polylineGroup.addObject(polyline2)
// ...
polylineGroup.addObject(polyline10000)
mapView.addAnnotationGroup(polylineGroup)
```

This would also allow functionality like: 

- Easy restyling of shape groups
- Easy toggling of hide/show for shape groups

We would still support `mapView.addAnnotation(polyline)` like MapKit, with the understanding that it's more efficient to add many shapes with the groups API instead. 

Thoughts @1ec5 @friedbunny?

incanus:
If we are liking the looks of this, we should mock out what this Cocoa API looks like exactly, including styling configuration, how it interacts (if at all) with eventual layer reordering, and removal/other management.

1ec5:
Sure, why not?

incanus:
Only because it's unlike Apple's API and may raise questions as to when or when not to use it. The answer to that is: 

* When adding many shapes with identical style properties.

friedbunny:
Sounds like the performance gains would be worth the divergence/complexity. Having a built-in construct for managing groups of annotations is also a friendly addition in general.

[...]

PR #2673, opened 2015-10-19
Further annotation refactoring

jfirebaugh:
* Use an rtree for points instead of pre-creating tiles for all zoom levels (replaces #1694)
* Lazily create AnnotationTile objects and transfer ownership to worker (fixes thread safety issues)
* Reduce number of datastructures used
* Remove shape layers from style

:eyes: @kkaefer


platform/ios/MGLMapView.mm
```diff
@@ -285,9 +285,6 @@ - (void)commonInit
 
     _annotationImages = [NSMutableDictionary dictionary];
 
-    std::string defaultSymbolName([MGLDefaultStyleMarkerSymbolName UTF8String]);
```
1ec5: If `MGLDefaultStyleMarkerSymbolName` is no longer needed, it doesn’t need to be declared up top anymore.

incanus:
Cursory review looks solid. Questions: 

- In moving to `getPointAnnotationsInBounds()`, we give up the possibility of querying for shapes in future. In https://github.com/mapbox/mapbox-gl-native/issues/2082 we want to eventually do this. 

- Should we be using a combined R-tree for both annotations and server tile features, to lay groundwork for https://github.com/mapbox/mapbox-gl-native/issues/352? 

- I think per past discussion we agreed to discuss any introductions of new Boost utilities (`boost::geometry::index::rtree` being the addition here). Just a point of order. So let's discuss :smile:

jfirebaugh:
> - In moving to `getPointAnnotationsInBounds()`, we give up the possibility of querying for shapes in future. In https://github.com/mapbox/mapbox-gl-native/issues/2082 we want to eventually do this. 

I view `getPointAnnotationsInBounds()` as a stopgap, assuming that #2082 will be implemented via #352.

> - Should we be using a combined R-tree for both annotations and server tile features, to lay groundwork for https://github.com/mapbox/mapbox-gl-native/issues/352? 

If it follows the gl-js implementation, #352 will indeed use an R-tree. But I think that's a separable change.

I wish the existing annotations code could use geojson-vt-cpp for both points and lines, and not use an R-tree at all. But because the API encourages adding points one by one, rather than as a single group/layer, it's difficult/impossible to make that efficient. (Another example of how the public API is limiting the implementation options.)

> - I think per past discussion we agreed to discuss any introductions of new Boost utilities (`boost::geometry::index::rtree` being the addition here). Just a point of order. So let's discuss :smile: 

We are already using `boost::geometry::index::rtree` in [`CollisionTile`](https://github.com/mapbox/mapbox-gl-native/blob/master/src/mbgl/text/collision_tile.hpp).

incanus:
> I wish the existing annotations code could use geojson-vt-cpp for both points and lines, and not use an R-tree at all. But because the API encourages adding points one by one, rather than as a single group/layer, it's difficult/impossible to make that efficient. (Another example of how the public API is limiting the implementation options.)

We'll always need the ability to add points one-by-one on mobile, but perhaps we could think about the styling / management on the backend in the style of what's proposed in https://github.com/mapbox/mapbox-gl-native/issues/1708 as we start to consider ways to make our API do things other frameworks don't provide for due to our styling advantages. 

:+1: on everything else.

PR #2679, opened 2015-10-20
2668 sync marker

ljbade:


ljbade:
Ah silly GitHub double clicked.

PR #2680, opened 2015-10-20
[android] Sync GPS marker

### LINKED ISSUE #2411: Make map more fluid

### LINKED ISSUE #2668: Update GPS marker after renderSync

### LINKED ISSUE #2675: Check for View leaks
Was catching up on the Android performance videos.

This one is on leaking Views: https://www.youtube.com/watch?v=BkbHeFHn8JY

I think we hold referenced to MapView in NativeMapView, and possibly inside the JNI.

We should change these to weak references.

### LINKED ISSUE #2678: onMapChanged churns memory
Noticed that onMapChange creates a lot of new `Runnable` object.

If we ensure we only call `renderSync` from the UI thread, we can ditch the `post()`.

ljbade:
Lots of other performance and memory improvements.
Fixes #2668
Fixes #2411
Fixes #2678
Fixes #2675


ljbade:
Wrong branch for 2.2.0

tobrun:
NICE!  :tada: 

I have added some comments when retrieving the value from the weak reference holder. 
This could give nullpointers in case when a memory leak would occur.

PR #2686, opened 2015-10-20
Test metrics on staging endpoint

1ec5:
iOS metrics test metrics belong on the staging endpoint, not the production endpoint.

/cc @incanus @camilleanne


incanus:
:+1:

Issue #2597, opened 2015-10-12
Assertion failure at (object)->IsUint8Array() on Node.js v4.2.0


mikemorris:
I'm seeing an assertion failure at https://github.com/nodejs/node/blob/master/src/node_buffer.cc#L151 causing Travis Node.js v4.2.0 tests to fail (previously passing on v4.1.2). I think this is related to [changes to native buffer checks](https://github.com/nodejs/node/pull/3080) in Node.js v4.2.0.

```
Assertion failed: ((object)->IsUint8Array()), function WeakCallback, file ../src/node_buffer.cc, line 151.
Process 54799 stopped
* thread #1: tid = 0xdff217, 0x00007fff90e97286 libsystem_kernel.dylib`__pthread_kill + 10, queue = 'com.apple.main-thread', stop reason = signal SIGABRT
    frame #0: 0x00007fff90e97286 libsystem_kernel.dylib`__pthread_kill + 10
libsystem_kernel.dylib`__pthread_kill:
->  0x7fff90e97286 <+10>: jae    0x7fff90e97290            ; <+20>
    0x7fff90e97288 <+12>: movq   %rax, %rdi
    0x7fff90e9728b <+15>: jmp    0x7fff90e92c53            ; cerror_nocancel
    0x7fff90e97290 <+20>: retq
(lldb) bt
* thread #1: tid = 0xdff217, 0x00007fff90e97286 libsystem_kernel.dylib`__pthread_kill + 10, queue = 'com.apple.main-thread', stop reason = signal SIGABRT
  * frame #0: 0x00007fff90e97286 libsystem_kernel.dylib`__pthread_kill + 10
    frame #1: 0x00007fff9836b9f9 libsystem_pthread.dylib`pthread_kill + 90
    frame #2: 0x00007fff981059b3 libsystem_c.dylib`abort + 129
    frame #3: 0x00007fff980cda99 libsystem_c.dylib`__assert_rtn + 321
    frame #4: 0x000000010068ac7e node`node::Buffer::CallbackInfo::WeakCallback(v8::Isolate*, v8::Local<v8::Object>) + 238
    frame #5: 0x0000000100315b41 node`v8::internal::GlobalHandles::Node::PostGarbageCollectionProcessing(v8::internal::Isolate*) + 225
    frame #6: 0x0000000100314864 node`v8::internal::GlobalHandles::PostGarbageCollectionProcessing(v8::internal::GarbageCollector) + 148
[...]

PR #2688, opened 2015-10-20
Add make task for building node bindings via Xcode

mikemorris:
Think this should do it @kkaefer.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -84,8 +84,12 @@ apackage: android-lib-mips
 .PHONY: node
 node: ; $(RUN) HTTP=none ASSET=none CACHE=none Makefile/node
 
+
+.PHONY: Xcode/node
+Xcode/node: ; $(RUN) HTTP=none ASSET=none CACHE=none node/xproj
+
 .PHONY: xnode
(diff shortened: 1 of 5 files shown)


mikemorris:
Looks like this is grabbing the wrong libuv/node headers, I hit this before testing a pre-release Node version for https://github.com/mapbox/mapbox-gl-native/issues/2597, workaround was adding the node headers path to [`include_dirs`](https://github.com/mapbox/mapbox-gl-native/blob/00a1c2343abcdcfc222d26c9a579fc081fffff6a/binding.gyp#L13) in `binding.gyp`.

kkaefer:
on node v0.10, it's failing with `AttributeError: 'PBXAggregateTarget' object has no attribute ‘FrameworksPhase'`. Looks like `node-gyp` includes `-undefined dynamic_lookup` in the `libraries, which isn't something Xcode likes. Instead, that should be in `LDFLAGS`.

tmcw:
@mikemorris for the sake of documentation, why would someone want to run this instead of `make node`?

kkaefer:
@tmcw You want to run `make xnode` so that you have an Xcode project that you can open in Xcode. This helps when debugging code, since you'll get a GUI debugger in the case of a crash.

PR #2697, opened 2015-10-21
[core] Added more transform state getters

brunoabinader:
Some platforms requires specific information that was previously not available in public API, but stored internally in `TransformState` object.

### DIFF
--- a/src/mbgl/map/transform_state.cpp
+++ b/src/mbgl/map/transform_state.cpp
@@ -105,6 +105,15 @@ const LatLng TransformState::getLatLng() const {
     return ll;
 }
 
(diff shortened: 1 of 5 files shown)


tmpsantos:
:+1:

src/mbgl/map/transform_state.cpp
```diff
@@ -105,6 +105,15 @@ const LatLng TransformState::getLatLng() const {
     return ll;
 }
 
+double TransformState::pixel_x() const {
```
kkaefer: we settled on camelCase, so this should be `pixelX`

PR #2698, opened 2015-10-21
[core] Fix transition to default style when no classes are applied

brunoabinader:
Some property transitions were being set to their fallback values when no classes are being used. This issue is particularly perceived when the default transition duration is not zero (animated transitions).


tmpsantos:
:+1:

PR #2699, opened 2015-10-21
Cleanup Painter usage in MapContext

brunoabinader:
Both debug and animation time data can be obtained directly through `MapData` (which `Painter` also has access to), so no need to provide that to `Painter`. Also, `Painter::setup()` can be called within `Painter` ctor.


tmpsantos:
:+1:

PR #2700, opened 2015-10-21
Send isChanging to MGLMapViewDelegate during gestures

1ec5:
This PR pulls in #2373 and additionally causes `-mapViewRegionIsChanging:` to be called during rotation and pitch gestures.

/cc @friedbunny @incanus


1ec5:
Updated the changelog in 16122d4ec3a7dba1894c62ebf53c2f4f3e53d7d8.

PR #2702, opened 2015-10-21
[node] Add pitch

bsudekum:
`pitch` is a native option we're not yet using on node. This PR adds it as an option.

/cc @mikemorris @yhahn @jfirebaugh

### DIFF
--- a/package.json
+++ b/package.json
@@ -26,7 +26,7 @@
   ],
   "devDependencies": {
     "aws-sdk": "^2.2.9",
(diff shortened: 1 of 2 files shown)


mikemorris:
Add tests for the API and check if we can add or enable existing render tests?

jfirebaugh:
:+1:

bsudekum:
committed as https://github.com/mapbox/mapbox-gl-native/commit/93c5bc08b358f8959f82f5f364b3de58ebbe67c9

PR #2704, opened 2015-10-21
Minor telemetry improvements

friedbunny:
Made some minor changes to what we send as telemetry, mostly small improvements around how much actual data is sent.

- Bump telemetry user agent to 1.1
 - To help identify these and other recent changes
- Round zoom and horizontal/vertical accuracy to integer precision
 - These were floats with unnecessary padding added
- Only send cellular carrier/network type keys when there's something to send

/cc @mapbox/mobiledata


1ec5:
:+1:

friedbunny:
Squashed and rebased, going to merge when CI goes green(ish).

PR #2705, opened 2015-10-22
[core] Rationalize style property classes

jfirebaugh:
This is the first in a series of PRs that builds towards a full-fledged [programmatic style API](https://github.com/mapbox/mapbox-gl-native/issues/837). This is a big chunk of work, but I'm trying to break it up step by step and not create a long-lived branch with a lot of changes.

Overall, my first major goal is to eliminate `ClassProperties` -- this is essentially an untyped map of every possible style property. Instead, we should work with classes like:

```cpp
class FillPaintProperties {
public:
    bool antialias = true;
    float opacity = 1.0f;
    Color fill_color = {{ 0, 0, 0, 1 }};
    Color stroke_color = {{ 0, 0, 0, -1 }};
    std::array<float, 2> translate = {{ 0, 0 }};
    TranslateAnchorType translateAnchor = TranslateAnchorType::Map;
    Faded<std::string> image;

    inline bool isVisible() const {
        return opacity > 0 && (fill_color[3] > 0 || stroke_color[3] > 0);
    }
};
```

These classes already exist, they are just inconsistently named. This PR makes the naming both internally consistent, and consistent with terminology from the style specification: for each layer type, there is a `____PaintProperties` class and a `____LayoutProperties` class. For the time being, all of these classes live in `style_properties.hpp`.

In code that I touch, I'm also fixing style issues like `&` placement and variable naming conventions. This will be the case with the whole series of PRs.


include/mbgl/style/style_properties.hpp
```diff
@@ -105,26 +160,31 @@ struct RasterProperties {
...
-    RasterProperties,
-    BackgroundProperties,
+    FillPaintProperties,
+    LinePaintProperties,
+    CirclePaintProperties,
+    SymbolPaintProperties,
+    RasterPaintProperties,
+    BackgroundPaintProperties,
     std::false_type
 > StyleProperties;
```
kkaefer: Can we rename this to `StylePaintProperties`?

kkaefer:
Looks good!

brunoabinader:
It is great that now layout properties are now exported to the [public API](https://github.com/mapbox/mapbox-gl-native/pull/2705/files#diff-22f9c5df9ed2491e62718e2d7b68652b). :+1: !

include/mbgl/style/style_properties.hpp
jfirebaugh: It's going away in the next commit, so I held off.

PR #2710, opened 2015-10-22
[android] Fix bug in setSprite when calculating pixel data size

### LINKED ISSUE #2506: setSprite does not handle scales <1.0x or >2.0x
As I discovered while working on #2440 the current core `setSprite` function is not flexible enough for the needs of Android. @friedbunny also noted that 3x needed for new iOS devices is also not supported.

The `scale` parameter crashes if you pass it <1.0 or >2.0. Android needs to support at least up to 4.0 (currently - future devices can increase this) and below 1.0.

A below 1.0 case might be where a developer only includes hdpi bitmaps assets, then a user runs the app on a ldpi or mdpi phone. In this case we need to scale down.

There are two ways I can see to fix this:
- scale the Bitmap to the correct 1x or 2x sprite texture size using the CPU - will result in blurry markers on higher DPI devices
- make sprites scale independent by storing them in their native resolution in the texture, then using the vertex shader to scale them on the CPU

/cc @1ec5 @incanus @kkaefer @nickidlugash @bleege @jfirebaugh

ljbade:
Change Marker sprite() to icon() and use Sprite class
Load a Maki dog icon in test app
Fixes #2506


ljbade:
Both a Bitrise and Travis passed. Merging now to save time.

PR #2711, opened 2015-10-22
Add style string resources

### LINKED ISSUE #2703: Add public string resouces with style URLs
In the SDK you can set the style either via the `setStyle` function, or in the layout XML with the `style_url` property.

For code ease of use (as well as automatic migration to correct style spec when apps update the SDK) we provide `Styles` which has a set on String constants for the bundled styles.

We do not provide anything for XML use.

Thus we should add some XML string values and mark them as public so users can access them.

/cc @tobrun

ljbade:
Use `_` in public resources ids
Make visibility attributes `enum`s
Make `attr`s public
Remove info window margin public value
Reformat XML code
Test app uses XML map properties for InfoWindow activities
Fixes #2703


ljbade:
iOS tests on Bitrise and Travis failed, but no iOS files changed in this PR. Must be unstable currently.

PR #2717, opened 2015-10-22
[android] Add MapActivity and AppCompatMapAcitivity to SDK. Improve MapFragment

### LINKED ISSUE #2616: MapActivity in SDK
Acknowledged by both @bleege and @ljbade that this would be a nice addition to the SDK.
Current setup requires overriding multiple lifecycle methods of Activity. 
In the past I forgot to add one or I had overridden the wrong one.. imo this setup is error prone. 
Besides that I would also simplify the code in the test application, which makes it easier to understand, 

Proposition:
 - in similar fashion as [ListActivitity] (https://android.googlesource.com/platform/frameworks/base/+/refs/heads/master/core/java/android/app/ListActivity.java) in the Android SDK
  - uses a default layout file, allows setting a different one but enforces a certain id (android.R.id.list)

ljbade:
Fixes #2616

### DIFF
--- a/android/MapboxGLAndroidSDK/build.gradle
+++ b/android/MapboxGLAndroidSDK/build.gradle
@@ -20,6 +20,7 @@ repositories {
 dependencies {
     compile 'com.android.support:support-annotations:23.1.0'
     compile 'com.android.support:support-v4:23.1.0'
+    compile 'com.android.support:appcompat-v7:23.1.0'
     compile 'com.android.support:design:23.1.0'
     compile 'com.squareup.okhttp:okhttp:2.5.0'
     compile 'com.mapzen.android:lost:1.0.1'
--- a/android/MapboxGLAndroidSDK/src/main/java/com/mapbox/mapboxsdk/AppCompatMapActivity.java
+++ b/android/MapboxGLAndroidSDK/src/main/java/com/mapbox/mapboxsdk/AppCompatMapActivity.java
@@ -1,0 +1,119 @@
+package com.mapbox.mapboxsdk;
+
+import android.os.Bundle;
+import android.support.annotation.NonNull;
+import android.support.annotation.Nullable;
+import android.support.v7.app.AppCompatActivity;
+
+import com.mapbox.mapboxsdk.utils.ApiAccess;
+import com.mapbox.mapboxsdk.views.MapView;
+
+public class AppCompatMapActivity extends AppCompatActivity {
+
+    //
+    // Static members
+    //
+
+    // Tag used for logging
+    private static final String TAG = "MapActivity";
+
+    //
+    // Instance members
+    //
+
+    // The access token
+    private String mAccessToken;
+
+    // The map style
+    private String mStyleUrl;
+
+    // The map
+    private MapView mMap;
+
+    //
+    // Constructors
+    //
+
+    public AppCompatMapActivity(@Nullable String accessToken) {
+        if (accessToken == null) {
+            accessToken = ApiAccess.getToken(this);
+        }
+
+        mAccessToken = accessToken;
+    }
+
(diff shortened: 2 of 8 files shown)


ljbade:
@tobrun I think `MapActivity` and `AppCompatMapActivity` are ready. And `PressForMarkerActivity` now uses it.

However I am still lost with `MapFragment`. Are you able to fix it up to use arguments and `getInstance()`. I'm not exactly sure of the purpose of `getInstance()` but this is how Google API's `MapFragment` works.

Once this is done I can copy this and make a `SupportFragment` based class for apps that still use the support lib fragments.

jush:
The `MapActivity` and `AppCompatMapActivity` override the empty constructor so it won't be possible to start those activities directly ([Starting an Activity](https://developer.android.com/guide/components/activities.html#StartingAnActivity)).
Does it mean that us (users of the library) will need to subclass those classes and provide the empty constructor?

tobrun:
@ljbade: @jush comment is valid. The only way this would work for Activity is subclassing the empty constructor and calling a super constructor with an API key. By default you should not use constructors for Activity, Fragment and even most other SDK components. A lot of these have lifecycle events and are managed by the system. 'Managed by the system" means that these can be recreated by the system which could result to state loss (because the system will call the no arg constructor and not your custom one). 

So when I created the `MapFragment` in the SDK, I had to create the `ApiAccess` class and use it directly to support all types of possible interactions. For example using `newInstance` method would be better in most cases but this will not allow it to be used as a static fragment. You can't pass the access token as an attribute as with the `MapView` class. As noted by @jush, Activities have similar limitations. 

As far as I can see there are no workarounds possible and we just need to enforce usage of the metadata-tag in AndroidManifest for developers using MapFragment or MapActivity. This is a widely used setup principle and is the conform the setup as GoogleMaps. Another way would be to enforce the usage of a certain `StringRes` as with Google Analytics.

jush:
Actually, do we (users of the library) need a `MapActivity` from mapbox at all? Google Maps doesn't provide one as far as I can see ([Google Maps Android API classes](https://developers.google.com/android/reference/com/google/android/gms/maps/package-summary)).

tobrun:
@jush That is/was our standpoint as wel. We have noticed when implementing multiple activities in the Test app we are making silly mistakes (eg. forgetting to override onSaveInstanceState etc). We want to make it as easy as possible to integrate and limit the amount of friction introduced by these silly copy/paste errors. Even if we don't add it to the SDK itself we will still need this for test app.

ljbade:
Ah interesting, I had envisioned people subclassing MapActivity. @tobrun Perhaps it is best for you to tackle this since you are more familiar with the ins and outs here.

karthikb351:
+1 for having a `MapActivity` and `AppCompatMapActivity` as part of the sdk. It would make it really simple to integrate.

I agree with @tobrun, the token as a Manifest metadata is probably the best way to go. It is a common pattern that a lot of libraries follow and developers should be familiar with how it works. Overriding constructors in Activities can cause some unexpected/unpredictable results.

The only issue with this method is that you can't set the token at runtime. This might be overkill, but if you want to set it at runtime while still using `MapActivity` and `AppCompatMapActivity`, one common workaround it to have a setter function in `ApiAccess`, which saves to SharedPrefs or a global static. Now you just need to ensure you set an token before you load up your activity/fragment. In the case of a fragment, you can set the token in the parent Activity before you instantiate it, and in the case of Activity, you can set it at the Application's onCreate.

This is the design pattern used by libraries like [Helpshift](https://developers.helpshift.com/android/getting-started/#initializing) and [Crashlytics](https://docs.fabric.io/android/fabric/integration.html#fabric-with-kits) as well.

This way, you get the best of everything. You can add to the manifest or programmatically set the api token and subclass either of the activities.

tobrun:
Related issue was closed down, will close this PR

PR #2725, opened 2015-10-22
[android] #2552 - add 10000 markers

### LINKED ISSUE #2552: Support 10000 Annotations In Android
iOS has the ability to parse and render 10000 Annotations at a time.  We need to [bring this over](https://github.com/mapbox/mapbox-gl-native/blob/3a1f62e6b9e1e204dde3dd3cf3736cebc1ac0f45/ios/app/MBXViewController.mm#L254-L297) to Android.

NOTE:
10000 is not an exact number nor should be the outcome.  It's just a large number that we've used internally to test / refer to this feature.

tobrun:
Closes #2552 

add markers in bulk, 
updated geojson parser to handle multiple types of geojson, 
added feature activity, 
updated icons in nav drawer, 
partially fixed local ref overflow in jni


tobrun:
Travis approves, Only Android code, merging

PR #2730, opened 2015-10-22
[core] Introduce StyleLayer subclasses

jfirebaugh:
Second in a series of PRs that builds towards a full-fledged [programmatic style API](https://github.com/mapbox/mapbox-gl-native/issues/837). This is a big chunk of work, but I'm trying to break it up step by step and not create a long-lived branch with a lot of changes.

This change introduces subclasses of `StyleLayer`, one for each layer type. Each subclass holds an instance of the appropriate `___PaintProperties` that we introduced in the [previous commit](https://github.com/mapbox/mapbox-gl-native/pull/2705). That means we can eliminate the `StyleProperties` variant. (Well, not completely: `ShapeAnnotation` still uses a version reduced to varying between `LinePaintProperties` and `FillPaintProperties`.)

We can also move the explicit instantiations of `StyleLayer::applyStyleProperties` (one for each layer type) into pure virtual `StyleLayer::applyStyleProperties` and subclass implementations. Future commits in this series will make similar changes, replacing `switch` or `template` polymorphism with good old type polymorphism.

### DIFF
--- a/include/mbgl/style/style_properties.hpp
+++ b/include/mbgl/style/style_properties.hpp
@@ -1,14 +1,10 @@
 #ifndef MBGL_STYLE_STYLE_PROPERTIES
 #define MBGL_STYLE_STYLE_PROPERTIES
 
-#include <mapbox/variant.hpp>
-
 #include <mbgl/style/types.hpp>
 
 #include <array>
 #include <string>
-#include <type_traits>
-#include <memory>
 #include <vector>
 
 namespace mbgl {
@@ -169,21 +165,15 @@ public:
     float opacity = 1.0f;
     Color color = {{ 0, 0, 0, 1 }};
     Faded<std::string> image;
+
+    inline bool isVisible() const {
+        return opacity > 0;
+    }
 };
 
 class BackgroundLayoutProperties {
 public:
 };
-
-typedef mapbox::util::variant<
-    FillPaintProperties,
-    LinePaintProperties,
-    CirclePaintProperties,
-    SymbolPaintProperties,
-    RasterPaintProperties,
-    BackgroundPaintProperties,
-    std::false_type
-> StyleProperties;
 
 }
 
--- a/src/mbgl/annotation/shape_annotation_impl.hpp
+++ b/src/mbgl/annotation/shape_annotation_impl.hpp
@@ -6,6 +6,7 @@
 #include <mbgl/util/geo.hpp>
 #include <mbgl/util/geojsonvt/geojsonvt.hpp>
 
+#include <memory>
 #include <string>
 #include <map>
 
--- a/src/mbgl/map/tile_worker.cpp
+++ b/src/mbgl/map/tile_worker.cpp
@@ -2,6 +2,7 @@
 #include <mbgl/map/tile_worker.hpp>
 #include <mbgl/style/style.hpp>
 #include <mbgl/style/style_layer.hpp>
+#include <mbgl/style/property_evaluator.hpp>
 #include <mbgl/geometry/glyph_atlas.hpp>
 #include <mbgl/renderer/fill_bucket.hpp>
 #include <mbgl/renderer/line_bucket.hpp>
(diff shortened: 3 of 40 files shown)


src/mbgl/map/tile_worker.cpp
```diff
@@ -64,29 +65,6 @@ void TileWorker::redoPlacement(float angle, float pitch, bool collisionDebug) {
 }
 
 template <typename T>
-struct PropertyEvaluator {
```
jfirebaugh: `PropertyEvaluator` was duplicated in two places and another part of this commit is pulling it out to `mbgl/style/property_evaluator.hpp`.

src/mbgl/renderer/circle_bucket.cpp
```diff
@@ -19,10 +20,10 @@ void CircleBucket::upload() {
 }
 
 void CircleBucket::render(Painter& painter,
-                        const StyleLayer& layer_desc,
+                        const StyleLayer& layer,
                         const TileID& id,
                         const mat4& matrix) {
-    painter.renderCircle(*this, layer_desc, id, matrix);
+    painter.renderCircle(*this, dynamic_cast<const CircleLayer&>(layer), id, matrix);
```
jfirebaugh: A future commit will move the `Painter::render*` methods to the layer subclasses.

src/mbgl/style/style_layer.hpp
```diff
@@ -47,37 +45,74 @@ class StyleLayer : public util::noncopyable {
...
     // for this layer (feature property filters, tessellation instructions, ...).
-    util::ptr<const StyleBucket> bucket;
+    util::ptr<StyleBucket> bucket;
 
     // Contains all style classes that can be applied to this layer.
-    const std::map<ClassID, ClassProperties> styles;
+    std::map<ClassID, ClassProperties> styles;
+
+protected:
+    // TODO: extract
```
jfirebaugh: This happens in the next commit.

src/mbgl/style/style_parser.cpp
```diff
@@ -662,53 +664,151 @@ void StyleParser::parseLayers(JSVal value) {
...
-        } else {
-            layer->type = StyleLayerTypeClass(std::string { type.GetString(), type.GetStringLength() });
+        if (it->second.second) {
+            layers.emplace_back(it->second.second);
         }
     }
+}
 
-    if (layer->bucket) {
+void StyleParser::parseLayer(const std::string& id, JSVal value, util::ptr<StyleLayer>& layer) {
```
jfirebaugh: The `StyleParser` changes are necessitated by now needing to know the type of layer being instantiated, so that `StyleLayer::create` can create the appropriate subclass. So we have to delay that until later in the parsing process.

include/mbgl/style/style_properties.hpp
```diff
@@ -169,22 +165,16 @@ class BackgroundPaintProperties {
     float opacity = 1.0f;
     Color color = {{ 0, 0, 0, 1 }};
     Faded<std::string> image;
+
+    inline bool isVisible() const {
+        return opacity > 0;
+    }
 };
```
brunoabinader: What about adding a `PaintProperties` that every `____PaintProperties` inherits containing `isVisible` as a pure virtual?

src/mbgl/annotation/shape_annotation_impl.hpp
```diff
@@ -6,6 +6,7 @@
 #include <mbgl/util/geo.hpp>
 #include <mbgl/util/geojsonvt/geojsonvt.hpp>
 
+#include <memory>
```
brunoabinader: Can we move this to `src/mbgl/annotation/shape_annotation_impl.cpp`? It doesn't seem to be use elsewhere in this file.

brunoabinader:
This is great stuff! :+1: particularly excited about the way how `StyleLayer` code is now much more readable.

include/mbgl/style/style_properties.hpp
jfirebaugh: We'd want to do that only if we actually needed runtime polymorphism from `PaintProperties`. So far I don't see the need -- everywhere `isVisible()` is currently used, it's with a concrete type.

src/mbgl/annotation/shape_annotation_impl.hpp
jfirebaugh: It's necessary for:

```
using Map = std::map<AnnotationID, std::unique_ptr<ShapeAnnotationImpl>>
```

This was compiling before due to transitive includes.

Issue #2687, opened 2015-10-20
Crash on startup in Simulator — -[MGLMapboxEventsData init]


1ec5:
#2581 introduced a crash on startup in the iPhone 6s Simulator. `carrierVendor` is nil but is being [dereferenced](https://github.com/mapbox/mapbox-gl-native/blob/28be82c1c19505abb47ca1c03fd069cc4a015ec6/platform/ios/MGLMapboxEvents.m#L88):

```
#0	0x0000000000000000 in 0x00000000 ()
#1	0x000000010de9c0cd in -[MGLMapboxEventsData init] at /Users/mxn/hub/mapbox-gl-native/platform/ios/MGLMapboxEvents.m:88
#2	0x000000010de9ef11 in -[MGLMapboxEvents resumeMetricsCollection] at /Users/mxn/hub/mapbox-gl-native/platform/ios/MGLMapboxEvents.m:410
#3	0x000000010de9d4d6 in -[MGLMapboxEvents init] at /Users/mxn/hub/mapbox-gl-native/platform/ios/MGLMapboxEvents.m:249
#4	0x000000010de9e172 in __32+[MGLMapboxEvents sharedManager]_block_invoke_2 at /Users/mxn/hub/mapbox-gl-native/platform/ios/MGLMapboxEvents.m:313
#5	0x00000001118a349b in _dispatch_client_callout ()
#6	0x000000011188ee28 in dispatch_once_f ()
#7	0x000000010de9e11e in _dispatch_once [inlined] at /Applications/Xcode+1.app/Contents/Developer/Platforms/iPhoneSimulator.platform/Developer/SDKs/iPhoneSimulator9.1.sdk/usr/include/dispatch/once.h:68
#8	0x000000010de9e100 in __32+[MGLMapboxEvents sharedManager]_block_invoke at /Users/mxn/hub/mapbox-gl-native/platform/ios/MGLMapboxEvents.m:312
#9	0x000000010de9e054 in +[MGLMapboxEvents sharedManager] at /Users/mxn/hub/mapbox-gl-native/platform/ios/MGLMapboxEvents.m:322
#10	0x000000010dee9b39 in +[MGLAccountManager setAccessToken:] at /Users/mxn/hub/mapbox-gl-native/platform/ios/MGLAccountManager.m:65
[...]

PR #2733, opened 2015-10-22
Fix for a crash on iPod Touch related to carrier vendor.

adam-mapbox:


1ec5:
This is a better fix for #2687 than #2691. Can you revert #2691 as part of this PR? Thanks.

friedbunny:
@1ec5 is correct, this does address the simulator case, too. I'm anxious to jump in here and atone for my hastiness. :mask:

PR #2735, opened 2015-10-22
[core] Style-sourced shape annotation properties

jfirebaugh:
This introduces the possibility to source the type and style properties of a shape annotation from a designated layer in the style.

:eyes: @brunoabinader


brunoabinader:
Looks great! One thing, though: I believe for classy annotations, we'd also need to check for `Update::Classes` in https://github.com/mapbox/mapbox-gl-native/blob/master/src/mbgl/map/map_context.cpp#L173 - do you agree?

PR #2737, opened 2015-10-22
IB inspectable for allowing tilt

1ec5:
This PR adds an inspectable to Interface Builder for enabling or disabling tilt gesture recognition. 42dc5336db5179c50b8d1edeca0bef7a27c8cefd added some of the code required for this inspectable, but it didn’t declare the inspectable itself. Additionally, the property was called `allowsPitching` – I changed it to `allowsTilting`, because “to pitch” has [many senses](https://en.wiktionary.org/wiki/pitch#Verb_2) that are more common than the intended aviation sense of the word. (One adjusts the `centerCoordinate` by scrolling or panning the map. Similarly, adjusts the map view’s `pitch` by tilting the map.)

/cc @friedbunny


friedbunny:
Sad to see the :baseball: go, but :+1:.

Issue #2684, opened 2015-10-20
Social


bleege:
/cc @incanus

bleege:
Integrated with Mapbox for Android 2.1.0.

bleege:
Officially announced today that the Mapbox iOS and Android SDKs will be [available via Twitter Fabric](https://www.mapbox.com/blog/fabric/).

![21718459683_996d47d486_b](https://cloud.githubusercontent.com/assets/989928/10651967/92c2fc4c-7807-11e5-9535-58d4643fdf6a.jpg)

1ec5:
:icecream:

PR #2742, opened 2015-10-23
refs #2684: integrate Twitter Fabric

incanus:
More to do here, but this is a start.

### DIFF
--- a/gyp/platform-ios.gypi
+++ b/gyp/platform-ios.gypi
@@ -92,6 +92,7 @@
         'OTHER_CPLUSPLUSFLAGS': [ '<@(cflags_cc)' ],
         'CLANG_ENABLE_OBJC_ARC': 'YES',
         'CLANG_ENABLE_MODULES': 'YES',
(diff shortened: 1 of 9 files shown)


incanus:
This is good once I get tests to pass. Will also update the [release process](https://github.com/mapbox/mapbox-gl-native/wiki/Releasing-Mapbox-GL-for-iOS) since we have to bump a version number as part of this.

mb12:
@incanus What functionality does this add?

incanus:
@mb12 This is related to our [partnership with Twitter](http://www.theverge.com/2015/10/22/9600792/twitter-fabric-developer-toolkit-stripe-aws) and only affects installs done via the forthcoming Fabric integration. It's dead code on other installs.

incanus:
In case it looks particularly lightweight, we also have backend code supporting the heavy lifting in conjunction with this code. 

See also https://github.com/mapbox/mapbox-gl-native/commit/bb6fff0c5db8091b4af43e51e3411fcaca161f18.

PR #2743, opened 2015-10-23
[core] Remove MapData dependency from StyleParser

jfirebaugh:
Third in a series of PRs that builds towards a full-fledged [programmatic style API](https://github.com/mapbox/mapbox-gl-native/issues/837). This is a big chunk of work, but I'm trying to break it up step by step and not create a long-lived branch with a lot of changes.

When working on other changes in this series, I found that having a `MapData` dependency in the style parsing code was problematic for other necessary refactoring. The parsing code is low level code that should not have a dependency on high-level objects such as `MapData`.

MapData was needed for two things: getting the default transition and fade durations, and detecting still rendering vs. continuous. This avoids the first by making transition durations optional, and supplying the default fallback value at evaluation time. The second did not fix the still rendering issue, and will be followed up with a different approach in a later commit.

### DIFF
--- a/src/mbgl/layer/background_layer.cpp
+++ b/src/mbgl/layer/background_layer.cpp
@@ -2,10 +2,10 @@
 
 namespace mbgl {
 
-RenderPass BackgroundLayer::applyStyleProperties(const float z, const TimePoint& now, const ZoomHistory& zoomHistory) {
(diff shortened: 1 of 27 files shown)


brunoabinader:
Looks great! :+1: Makes me wonder if we can have a similar class `PainterParameters` for `Painter` class.

Issue #2726, opened 2015-10-22
Local reference table overflow while adding markers in bulk


tobrun:
Documented at original ticket #2552 
Stacktrace is located [here](http://pastebin.com/BYH9TgQP)

After resolving JNI issue:
 - add a Spinner to add 100, 1000 or 10000 markers (cfr. iOS)
 - add a progress indicator for AsyncTask

ljbade:
@tobrun You need to call `DeleteLocalRef` on any objects that are created in a loop in the JNI.

JNI only deletes "local refs" (ie Objects) at the exit of a JNI method, so if you generate a lot of objects before you return, you can overflow the local ref table.

ljbade:
In this case from the log, it is `Sprite` and `LatLng` objects.

ljbade:
Deleting `jid` fixed the overflow.

1,000 markers works:
![device-2015-10-23-154345](https://cloud.githubusercontent.com/assets/1215089/10684767/f3a39148-799c-11e5-8735-53dfe04898fc.png)

10,000 markers freezes after it gets this far:
![device-2015-10-23-154345](https://cloud.githubusercontent.com/assets/1215089/10684771/0485ffa0-799d-11e5-8149-f977e18aa279.png)

ljbade:
Making a PR now.

Still leaves:
 - add a Spinner to add 100, 1000 or 10000 markers (cfr. iOS)
 - add a progress indicator for AsyncTask 
 - fix freeze with 10,000 markers

ljbade:
@tobrun Can we also add an option to switch between `addMarker()` one at a time, and `addMarkers()` to stress test both ways.

ljbade:
Been investigating the 10,000 stall.

I can repro with just 2,000 markers.

Loading the GeoJSON into `MarkerOption` list takes 1.3 seconds.
Adding the markers to the map with `addMarkers` takes 0.3 seconds.

This happens in a async task so doesn't freeze the Android UI.

What happens next is that the Core GL threads process the markers, and stalls `renderSync`, which in turn freezes the map.

@kkaefer Is this a known problem in CoreGL?

If we could solve the 30fps issue in Android then we could punt `renderSync` onto a separate Java thread and not freeze the rest of the app's UI while the map does it's thing. However this presents user location dot sync issues that I have not figure out yet.

ljbade:
@kkaefer by the way it only uses 1/8th of the CPU cores

[...]

PR #2745, opened 2015-10-23
[android] Fix JNI local ref overflow

ljbade:
Change test to 1000 markers
For #2726


ljbade:
Travis iOS hanging. Merging.

Issue #2022, opened 2015-08-10
Using source code (GYP integration)


xrayed:
I have tried to use Mapbox GL source, so I followed steps and used GYP,
but here are some problems when I ran following:

gyp ios/app/mapboxgl-app.gypi --depth=. -f xcode --generator-output=./build.xcodefiles/

1) gyp: name 'host' is not defined while evaluating condition 'host == "ios"' in ios/app/mapboxgl-app.gypi
 - it seems that this is an naming error, this file might have 'OS == "ios" instead of 'host == "ios"

2) When I tried with ['OS == "ios"] another error appeared:
gyp: Undefined variable platform_lib in ios/app/mapboxgl-app.gypi

I'm new to GYP, maybe I did something wrong

1ec5:
You shouldn't need to use gyp directly for developing Mapbox GL. To create the iOS Xcode workspace, run `make iproj`. See the readme for more details.

PR #2746, opened 2015-10-23
Replace bundled styles with mapbox: style URLs

### LINKED ISSUE #2239: unbundle default Mapbox styles from SDKs
Load these via the Mapbox Styles API instead. Bundling doesn't win us anything, because network access is still needed for sources, sprite, fonts, and tiles. Unbundling allows us to remove the styles submodule from this repository (and not replace it with an npm dependency).

ljbade:
Fixes #2239


1ec5:
We also need to deprecate or remove the bundledStyleURLs property from MGLMapView, since that property always returns an empty array. We can deprecate it with a compile-time message instructing developers to look at some webpage for a list of always-available styles, but we need to have that webpage ready.

ljbade:
@1ec5 can we make it return a list of `mapbox://` URLs?

1ec5:
That’s not a bad idea. We’d just base it off default_styles.cpp, which would enable us to nix the parallel list we’ve compiled in MBXViewController. We’d still need to rename the method, though: the styles aren’t “bundled”, just universally available, regardless of access token.

ljbade:
@1ec5 I was thinking more from source compatibility perspective but next release is breaking anyway right?

1ec5:
You’re right, we should mark `-bundledStyleURLs` as deprecated with a message to migrate to whatever we call it. `-universalStyleURLs`?

ljbade:
`-mapboxStyleURLs`?

1ec5:
I’ve completed the iOS side of this change. There’s a new class MGLStyle that – for now – does nothing but vend style URLs in a manner similar to UIColor. It would make for a great entry point an iOS runtime styling API as part of #837. Unlike Style.java on the Android side, MGLStyle gets its values directly from a refactored default_styles.hpp and uses static assertions to stay in sync. So that takes care of #1462. Objective-C++, oh happy day!

The upshot for other platforms is that there are now individual constants for each “default style” (that is, a style that’s available regardless of access token). There’s still a constant C array of them in case you need to iterate over them.

@jfirebaugh, does my C++ look reasonable? @incanus, can you forgive my macro shenanigans?

include/mbgl/util/default_styles.hpp
```diff
@@ -0,0 +1,32 @@
...
+
+#include <vector>
+#include <string>
+
+namespace mbgl {
+namespace util {
+namespace default_styles {
+
+struct DefaultStyle {
+    const char *url;
```
jfirebaugh: Our style for C++ is to space `*` and `&` with the rest of the type, so `const char* url`.

jfirebaugh:
LGTM other than the one nit.

include/mbgl/util/default_styles.hpp
1ec5: Old habits die hard: e13535eda83116ced526c6785e9596e25bdfaac2.

ljbade:
Excellent, when your happy @1ec5 I think we can squash, get Travis green, final set of eyes, then merge.

1ec5:
Running into #2022.

ljbade:
@1ec5 is that the right ticket number?

1ec5:
Yes, that was precisely what I and Travis were both hitting, but now it’s green again. Go figure.

1ec5:
I’ve come back around to the idea of deprecating and eventually removing `-bundledStyleURLs` from the iOS SDK (rather than just renaming it). I don’t see a real use case for host applications to iterate over these style URLs without any names that would be suitable for display in a UI. With the introduction of `-[MGLStyle streetsStyleURL]` et al., developers no longer need to rely on `-bundledStyleURLs` to [discover available styles](https://www.mapbox.com/ios-sdk/examples/custom-style/).

PR #2759, opened 2015-10-23
[android] Use hardware layer for compass fade

ljbade:
Set minimum API level to 14 (don't know why it 15)
Set test app version to 2.2.0


mb12:
@ljbade The following code will crash on all devices below jelly bean if withLayer method is not available. 

'''
if (android.os.Build.VERSION.SDK_INT >= Build.VERSION_CODES.JELLY_BEAN) {
                                    mFadeAnimator.withLayer();
                                }
'''

ljbade:
@mb12 That is what the `if` statement checks for. It wont run the function below jelly bean

mb12:
It will still fail in ClassLoader. The best option would be to write a helloworld testcase with just these 3 lines and test on below jellybean emulator. There was a similar crash <a href="https://github.com/mapbox/mapbox-gl-native/blob/master/android/java/MapboxGLAndroidSDK/src/main/java/com/mapbox/mapboxsdk/views/MapView.java#L613">here</a> as well. To resolve it, I believe the minsdkversion was bumped up.

ljbade:
I will switch in `ViewCompat` here.

ljbade:
None of the Android API 14 or 15 emulators support OpenGL so I can't test for support properly.

I am going to try and get an old Galaxy Nexus off ebay to wipe and install fresh Android 4.0 on

PR #2706, opened 2015-10-23
Test and document `willStartLoading` & `didFinishLoading` delegate methods

friedbunny:
With #630 and #2627, the map loading notifications should be being sent from core MBGL, which means that the `mapViewWillStartLoadingMap:` & `mapViewDidFinishLoadingMap:` delegate methods that already exist in `MGLMapViewDelegate` should be ready for use.

- [x] Make sure the current delegate methods [are actually hooked up](https://github.com/mapbox/mapbox-gl-native/blob/17121563ac807fbc68a203ac08c64de5711d8b92/platform/ios/MGLMapView.mm#L3001-L3016)
- [x] Update [core changelog](https://github.com/mapbox/mapbox-gl-native/blob/master/CHANGELOG.md)
- [x] Update [iOS docs](https://github.com/mapbox/mapbox-gl-native/blob/17121563ac807fbc68a203ac08c64de5711d8b92/include/mbgl/ios/MGLMapView.h#L437-L441)
- [ ] Is this testable? If so, write tests.

/cc @incanus @1ec5

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -38,6 +38,7 @@ Known issues:
 - A new delegate callback was added for observing taps to annotation callout views. ([#2596](https://github.com/mapbox/mapbox-gl-native/pull/2596))
 - `-mapViewRegionIsChanging:` is now sent to the map view’s delegate during gestures. ([#2700](https://github.com/mapbox/mapbox-gl-native/pull/2700))
 - Improved gesture recognition while the map is tilted. ([#2770](https://github.com/mapbox/mapbox-gl-native/pull/2770))
+- `-mapViewWillStartLoadingMap:` and `-mapViewDidFinishLoadingMap:` delegate methods now work. ([#2706](https://github.com/mapbox/mapbox-gl-native/pull/2706))
 - Removed CoreTelephony.framework dependency. ([#2581](https://github.com/mapbox/mapbox-gl-native/pull/2581))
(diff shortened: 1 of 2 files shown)


friedbunny:
Going to wait for the [`ios-improve-tests`](https://github.com/mapbox/mapbox-gl-native/tree/ios-improve-tests) branch to shake out before thinking about writing tests for this.

friedbunny:
This doesn't necessarily need to wait for the MapKitization proposed in #2775, but that should be on the radar here.

1ec5:
What’s left? Should this still be on the 3.0.0 milestone?

incanus:
Just to restate, the issue is that these don't work reliably for tests, or always? 

Option two sounds better to me providing these work for users, and we'll continue to up our testing game across the whole framework, though we'd need to update `CHANGELOG.md` to not be so bright-eyed and bushy-tailed, I think.

friedbunny:
> Just to restate, the issue is that these don't work reliably for tests, or always?

Per #2775, these two callbacks (and the related load failure callbacks) don't work as a MapKit developer would expect — they reliably fire, but at the wrong times.

incanus:
Ok, I think we can roll this PR in then for 3.0.0.

incanus:
Rebased & awaiting tests (even if this is just header doc changes).

PR #2770, opened 2015-10-23
improve gestures in perspective view

ansis:
with these changes:
- the location under your finger stays under your finger as you pan
- the location you double tap to zoom stays at the same point
- the location at the center of a pinch zoom stays at the center
- the location at the center of a rotation stays at the center

This is mostly a port from -js.

@1ec5 can you review?


1ec5:
:+1:

PR #2741, opened 2015-10-22
[android] Add satellite streets style

ljbade:
Fixes #2739
Fixes #2740

Issue #2740, opened 2015-10-22
Add all official styles to styles


ljbade:
From https://github.com/mapbox/mapbox-gl-native/issues/2739#issuecomment-150385722

> we need to update both iosapp and scripts/ios/package.sh.

/cc @1ec5

ljbade:
@1ec5 was looking at the iosapp code and it currently in `cycleStyles` assumes that the style file name is the lowercase+hyphenated version of the human name.

Satellite Streets == `satellite-hybrid-v8`

Breaks that.

ljbade:
@1ec5 Can you update the logic + name of the style in iOS for "Satellite Streets"

I don't have much of a clue in ObjC and I dont have iOS test setup handy

1ec5:
Taken care of in #2771.

PR #2771, opened 2015-10-23
More natural style names in iosapp

1ec5:
This followup to #2741 refactors iosapp to track each bundled style’s display name separately in parallel with the raw name. The upshot is that “Satellite-Hybrid” is now just “Hybrid”.

/ref #2740
/cc @ljbade


ljbade:
Excellent. Always interesting to see how things work in Obj C

PR #2778, opened 2015-10-24
Improve tvOS build; builds for correct platform, including dependencies.

corvino:
This is focused around improving the build system to create a tvOS project and build the dependencies for tvOS.

It is working well for my tvOS project. But the app and benchmark problems don't link against the system libs and sqlite properly.

Still quite a lot of progress on the tvOS front. I'd at least like to get feedback before pushing on.

I've tested using the tvOS-corvino branch for this and mason; because mason needs updating, I need extra commits to point mason at my fork. This pull request depends on https://github.com/mapbox/mason/pull/119; I believe together they should provide a working library for tvOS..


corvino:
This also needs https://github.com/mapbox/mason/pull/122 to update sqlite-system.

The machine I was building on had sqlite headers in /usr/include, so I hadn't seen this.

I'm updating my tvOS-corvino branches  as I find issues to (hopefully) build on clone. Requires switching to tvOS-corvino before initializing submodules, though.

friedbunny:
Cool, we appreciate your efforts here. I think we'll hold off on merging this into the `tvos` experimental branch until mason is sorted out.

My inclination is to say that you needn't worry about a tvOS benchmark app, but feel free to fiddle with that if you like.

jfirebaugh:
Closing since we're not likely to merge this in the near term; see https://github.com/mapbox/mapbox-gl-native/pull/2340#issuecomment-199980489.

Issue #2117, opened 2015-08-18
allow appledoc upgrade


incanus:
Currently we require appledoc 2.2v963 exactly for iOS docs generation. Is this still the case? If not, let's fix that so we can move on.

1ec5:
Last I checked, the main holdup was that newer versions of appledoc come with templates that mimick Apple’s 2014-era docsets. Those templates require JavaScript to work correctly, but Xcode’s Documentation window seems to disallow JavaScript in third-party docsets. This won’t be an issue for the [online documentation](https://www.mapbox.com/mapbox-gl-ios/api/) but will be for those who install the docset in Xcode (which I gather will be relatively few people).

I’d much prefer that we come up with new templates, like the ones used for [Mapbox GL JS documentation](https://www.mapbox.com/mapbox-gl-js/api/).

1ec5:
It’s more likely that we’ll end up switching to jazzy: #3203.

1ec5:
Bitrise is running appledoc 2.2.1. I just upgraded my machine from 2.2.0 and the templates are broken, which means we’ve been building broken documentation on Bitrise all this time. 2.2.1 comes with an impressive [58 commits](https://github.com/tomaz/appledoc/compare/v2.2-963...2.2.1).

1ec5:
The template breakage is tomaz/appledoc#561.

PR #2800, opened 2015-10-26
Recreate structure of documentation

### LINKED ISSUE #2799: Refactor documentation
The docs here could use a pass to become a bit more straightforward and followable for different sorts of people. Going to take a pass at refactoring these.

tmcw:
Fixes #2799

Trying to refactor this as much as possible, I'll record quibbles here:

> This process gives you a native iOS app for simulator or device which compiles the source to GL within it for fastest turnaround time during development

What does this mean? "compiles the source to GL"... does this JIT compile anything, or is it just saying "compiles mapbox-gl-native instead of using a bundle".

---

> Install [appledoc](http://appledoc.gentlebytes.com/appledoc/) 2.2v963 for API docs generation (only this exact version works right now ([#2117](https://github.com/mapbox/mapbox-gl-native/issues/2117)).

This links to a closed ticket that doesn't give a clear answer. Is this statement still true?

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,12 +1,12 @@
 # Contributing
 
-If you have a usage question for a product built on Mapbox GL (such as Mapbox Mobile toolkits like our SDKs), please visit https://www.mapbox.com/help/. 
+If you have a usage question for a product built on Mapbox GL (such as Mapbox Mobile toolkits like our SDKs), please visit https://www.mapbox.com/help/.
 
-If you want to contribute code: 
+If you want to contribute code:
 
-1. Please familiarize yourself with the [install process](INSTALL.md). 
+1. Please familiarize yourself with the [install process](INSTALL.md).
 
-1. Ensure that the [existing issues](https://github.com/mapbox/mapbox-gl-native/issues?utf8=✓&q=) don't already cover your question or contribution. 
+1. Ensure that the [existing issues](https://github.com/mapbox/mapbox-gl-native/issues?utf8=✓&q=) don't already cover your question or contribution.
 
 1. Pull requests gladly accepted. If there are any changes that developers using one of the GL SDKs should be aware of, please update the “master” section of the relevant changelog: [iOS](CHANGELOG.md) [Node.js](platform/node/CHANGELOG.md).
 
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -1,357 +1,75 @@
-# Install
+# Building & Developing Mapbox GL Native from Source
 
-This document describes the process for setting up Mapbox GL for development _on the framework itself_.
+**Just trying to use Mapbox GL Native? You don't need to read this stuff! We
+provide [easy-to-install, prebuilt versions of the Mapbox SDKs for iOS and Android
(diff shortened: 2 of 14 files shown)


tmcw:
Okay: I think this refactor is ready for review. cc @incanus

incanus:
Taking a look now @tmcw. 

> What does this mean? 

Ordinary consumers of GL link it into their app as a static library. For development, we actually integrate the source of GL right into the source of the test app, for compilation speediness reasons. More here: https://www.mapbox.com/ios-sdk/#source

> Is this statement still true?

I'm running `2.2.1 (1333)` so I think not. Will amend.

incanus:
I'm not sure about the approach of so many files, particularly `docs/TROUBLESHOOTING.md`. If I were having a problem, my first instinct wouldn't be to search the source for a file called `TROUBLESHOOTING.md`, particularly if one or more of the existing `DEVELOP_` files had troubleshooting sections already.

incanus:
We should probably bring in Node-specific instructions here as well, as "Linux on Linux" could be desktop, could be Node. /cc @mikemorris

incanus:
Otherwise I'm :+1: here @tmcw.

mikemorris:
Node.js docs have lived in https://github.com/mapbox/mapbox-gl-native/blob/master/platform/node/README.md for mostly historical reasons, we should consistify this between platforms - either breaks all platform implementations into `platform/%/README.md` or bring everything into `README.md`. Same goes for the Node.js [changelog](https://github.com/mapbox/mapbox-gl-native/blob/master/platform/node/CHANGELOG.md).

tmcw:
@mikemorris are there any instructions for building the node module from source? [The instructions in README.md](https://github.com/mapbox/mapbox-gl-native/tree/master/platform/node#installing) are

    npm install mapbox-gl-native

Which would install it from npm, right?

mikemorris:
@tmcw `npm install --build-from-source` to install node modules for initial source compile, `make node` (or [`make xnode`](https://github.com/mapbox/mapbox-gl-native/pull/2688)) for recompiling C++ files during development.

tmcw:
> This process gives you a native iOS app for simulator or device which compiles the source to GL within it for fastest turnaround time during development

Okay so this is trying to use "GL" as a stand-in for "the library". I'll improve this sentence.

tmcw:
I wrote [a section for the node.js bindings](https://github.com/mapbox/mapbox-gl-native/blob/redo-docs/docs/DEVELOP_NODE.md), documenting `--build-from-source`: if there's some more reasoned reason for using that versus `make node` versus `make xnode`, that's something for @mikemorris to write because it isn't clear from any of the issues/prs/code

tmcw:
Better API documentation, for all bindings, should be a second pass. This PR is just reorganizing our current build-stage docs.

mikemorris:
Cleaned up Node.js docs a bit, looks good to me now :+1:

tmcw:
rebasing into one commit & merging.

incanus:
Great work @tmcw. We've gone through some solid iterations of all this mess o' docs and this is the best yet.

PR #2804, opened 2015-10-26
Ensure URL normalization is safe for all input. Fixes #2695

### LINKED ISSUE #2695: Make URL normalization safe
https://github.com/mapbox/mapbox-gl-native/blob/master/src/mbgl/util/mapbox.cpp#L87-L90

All of the normalization methods in this file assume that `pathname` has the right arity before using parts of it constructing URLs. When it doesn't - when it's just a style id or something, or missing some chunk - then concatenating the strings leads to a malloc, as seen in http://stackoverflow.com/questions/33227894/how-do-i-show-a-map-from-a-mapbox-project-with-the-mapbox-ios-sdk

The methods should assert the correct arity before proceeding.

tmcw:
Check that a URL has enough segments before normalizing it.

cc @jfirebaugh for the review

### DIFF
--- a/src/mbgl/util/mapbox.cpp
+++ b/src/mbgl/util/mapbox.cpp
@@ -1,4 +1,5 @@
 #include <mbgl/util/mapbox.hpp>
+#include <mbgl/platform/log.hpp>
 
 #include <stdexcept>
 #include <vector>
@@ -50,6 +51,12 @@ std::string normalizeStyleURL(const std::string& url, const std::string& accessT
     }
 
     std::vector<std::string> pathname = getMapboxURLPathname(url);
+
+    if (pathname.size() < 3) {
+        Log::Error(Event::ParseStyle, "Invalid style URL");
+        return url;
+    }
+
     std::string user = pathname[1];
     std::string id = pathname[2];
     bool isDraft = pathname.size() > 3;
@@ -62,6 +69,12 @@ std::string normalizeSpriteURL(const std::string& url, const std::string& access
     }
 
     std::vector<std::string> pathname = getMapboxURLPathname(url);
+
+    if (pathname.size() < 3) {
+        Log::Error(Event::ParseStyle, "Invalid sprite URL");
+        return url;
+    }
+
     std::string user = pathname[1];
     bool isDraft = pathname.size() > 3;
 
@@ -85,6 +98,12 @@ std::string normalizeGlyphsURL(const std::string& url, const std::string& access
     }
 
     std::vector<std::string> pathname = getMapboxURLPathname(url);
+
+    if (pathname.size() < 4) {
+      Log::Error(Event::ParseStyle, "Invalid glyph URL");
+      return url;
+    }
+
     std::string user = pathname[1];
     std::string fontstack = pathname[2];
     std::string range = pathname[3];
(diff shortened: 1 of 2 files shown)


src/mbgl/util/mapbox.cpp
```diff
@@ -85,6 +95,11 @@ std::string normalizeGlyphsURL(const std::string& url, const std::string& access
     }
 
     std::vector<std::string> pathname = getMapboxURLPathname(url);
+
+    if (pathname.size() < 3) {
```
jfirebaugh: This one should be `pathname.size() < 4`, since `pathname[3]` is used unconditionally.

src/mbgl/util/mapbox.cpp
```diff
@@ -50,6 +50,11 @@ std::string normalizeStyleURL(const std::string& url, const std::string& accessT
...
+        return url;
```
1ec5: If a mapbox: URL has an unexpected format, wouldn’t this be a good opportunity to provide an assertion message to that effect? Otherwise you’ll wind up with a 404 but no other details. It’ll still be a crash but a helpful one.

tmcw: Log::Error(Event::ParseStyle, "Invalid URL for glyphs");

Would that be the appropriate kind of log?

tmcw:
Addressed @jfirebaugh's issue, and added log messages as suggested by @1ec5: is it possible to test our log output?

src/mbgl/util/mapbox.cpp
1ec5: In this particular case, `isDraft` assumes a pathname longer than 3. That’s dead code now.

1ec5:
At a glance, I think assertions would be better than log messages in this case, because client code may assume “normalized” to mean “no more mapbox: URLs”. But this is definitely an improvement over the current behavior.

src/mbgl/util/mapbox.cpp
tmcw: This is in the wrong section: I've adjusted the code to < 4 for the glyph endpoint

tmcw:
@jfirebaugh should I merge this as-is or switch to assertions?

ljbade:
One problem with asserts is that on Android it causes a native crash which confuses most Android devs as they are unlikely to have seen these before and will struggle to connect the crash to a malformed URL. I would much prefer logging, or even better a way to through a Java exception with a suitable error message.

/cc @bleege

1ec5:
I think it’s generally best to assert when the issue is likely to be developer error; logging and otherwise failing gracefully is better for issues out of the developer’s direct control, like network failures. The JNI code should have exception handling to catch assertions and present them as Java exceptions.

tmcw:
Sounds like we should do logging first, and then the JNI work, and then switch to exceptions

ljbade:
@1ec5 Unfortunately JNI was desinged before C++ exceptions were invented and there is no reliable way to pass them to Java. Even the exception's message is not printed on Android, all you get in the log is a SIGABRT error.

I have tried various libs that claim to handle C++ exceptions in JNI, but they were either outdated (did not support newer Android versions), unreliable (missed most exceptions), or just complicated debugging normal crashes with false stack traces.

Quite simply C++ exceptions are not compatible with Android.

bleege:
> The JNI code should have exception handling to catch assertions and present them as Java exceptions.

Agreed.  However, "what JNI should do" vs "what JNI can actually do" are often 2 different things.  :smile: Remember JNI is an imperfect (and usually one way) bridge between C++ code and Java code.

1ec5:
When you interface with C++ code, you have to be open to exceptions being thrown, and not just for the most catastrophic of reasons. So it sounds like we need to revisit this issue, probably as another layer beneath JNI that catches C++-style exceptions and turns them into C-style error codes. We'd do the same if we wanted to expose these error conditions as native exceptions on the iOS side too. But saying the C++ layer can't use C++ exceptions because of JNI is like the tail wagging the dog. 😊

I don't want to hold up this PR for a seemingly large task, so this is good to go.

jfirebaugh:
LGTM.

What @1ec5 says about exceptions is spot on -- when writing bindings to C++, it's common to need to blanket catch all exceptions and translate them into some other error mechanism that's compatible with the platform.

ljbade:
I can see our jni.cpp growing in size :-( we really need to refactor that file into smaller ones.

PR #2806, opened 2015-10-26
[core] Polymorphic layout and paint parsing

jfirebaugh:
Fourth in a series of PRs that builds towards a full-fledged [programmatic style API](https://github.com/mapbox/mapbox-gl-native/issues/837). This is a big chunk of work, but I'm trying to break it up step by step and not create a long-lived branch with a lot of changes.

This episode moves further towards being able to remove the generic `ClassProperties` in favor of concrete paint and layout classes like those found in [style_properties.hpp](https://github.com/mapbox/mapbox-gl-native/blob/bb6241844a092cf632fe740e6ac79c920117dd4e/include/mbgl/style/style_properties.hpp). The way that I'm going to do that is to continue refactoring until all usage of `ClassProperties` is in layer subclasses, and then substitute the concrete per-layer replacements.

Specifically, the code that needs to move to layer subclasses is layout and paint parsing (this PR) and bucket creation (next PR).

### DIFF
--- a/src/mbgl/layer/background_layer.cpp
+++ b/src/mbgl/layer/background_layer.cpp
@@ -1,6 +1,15 @@
 #include <mbgl/layer/background_layer.hpp>
+#include <mbgl/style/property_parsing.hpp>
 
 namespace mbgl {
+
+void BackgroundLayer::parsePaints(const JSVal& layer) {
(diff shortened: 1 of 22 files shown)


kkaefer:
Awesome!

PR #2815, opened 2015-10-27
[android] Fix crash on Marshmallow

### LINKED ISSUE #2813: Emulator crash part 2
Regression found for solution at #2807.

The fix solved the problem of running on ICS but created a crash on Marshmallow:
>E/libEGL: validate_display:255 error 3008 (EGL_BAD_DISPLAY)
E/mbgl: {Main}[OpenGL]: eglMakeCurrent(EGL_NO_CONTEXT) returned error 12296
A/libc: Fatal signal 6 (SIGABRT), code -6 in tid 29835 (apboxgl.testapp)

@ljbade @bleege @zugaldia

ljbade:
Fixes #2813

### DIFF
--- a/platform/android/native_map_view.cpp
+++ b/platform/android/native_map_view.cpp
@@ -423,6 +423,10 @@ void NativeMapView::createSurface(ANativeWindow *window_) {
              return reinterpret_cast<mbgl::gl::glProc>(eglGetProcAddress(name));
         });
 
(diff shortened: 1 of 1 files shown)


ljbade:
@jfirebaugh Both node and linux Travis jobs are being flakey with time outs and crashes even after restarting them.

This commit only touches files compiled in the Android jobs which pass so I am merging.

PR #2824, opened 2015-10-27
[android] #2224 inflate children map view through xml

### LINKED ISSUE #2224: Inflating MapView children using xml
For the MapView class we are currently adding child views programmatically. I want to test if using a layout inflater is as fast/efficient. The general idea is that this is the preferred way of creating views on Android and this would reduce ui code in the MapView class.

tobrun:
[android] #2224 inflate children map view through xml
[android] - #2224 make UserLocationView inflatable through xml. using setter to attach MapView
[android] #2224 - renamed fragment layout file, reordered views in layout file map view

closes #2224


tobrun:
Sigh, been trying to get it to pass on travis yesterday evening and this morning it is stil giving errors.. :disappointed:

tobrun:
Not touching any code of other OS'es so going to merge

tobrun:
Travis approved after my rebase! merging

ljbade:
@tobrun Travis is super unstable on Node/Linux last few days (@jfirebaugh is looking at it) - I've been merging as long as the Android jobs pass when a patch is only touching Android files.

jfirebaugh:
> I've been merging as long as the Android jobs pass when a patch is only touching Android files.

Please don't do this -- instead, make sure someone is aware of and working on the failure, and restart the job until it's green.

Yes, I know this is annoying. But a "merge if you think your changes aren't what caused the breakage" policy leeds to creeping complacency towards CI results, and eventually checking in changes which DO break the build. Then it takes longer to figure out what happened because we lose the direct connection to the cause. I spent most of the day yesterday debugging something that a) broke the first build it was introduced in and b) was fairly easy to spot once I tracked down that commit. But because I didn't know that to begin with, it took me all day to find.

ljbade:
@jfirebaugh OK I'll do that but it will seriously reduce my productivity.

PR #2829, opened 2015-10-27
always add labels to buffers in MapMode::Still

mikemorris:
To minimize label clipping.

/cc @bsudekum @ansis @miccolis

### DIFF
--- a/src/mbgl/layer/symbol_layer.cpp
+++ b/src/mbgl/layer/symbol_layer.cpp
@@ -114,7 +114,8 @@ bool SymbolLayer::recalculate(const StyleCalculationParameters& parameters) {
 
 std::unique_ptr<Bucket> SymbolLayer::createBucket(StyleBucketParameters& parameters) const {
     auto bucket = std::make_unique<SymbolBucket>(parameters.tileID.overscaling,
-                                                 parameters.tileID.z);
+                                                 parameters.tileID.z,
+                                                 parameters.mode);
 
     bucket->layout = layout;
 
--- a/src/mbgl/map/source.cpp
+++ b/src/mbgl/map/source.cpp
@@ -330,6 +330,7 @@ TileData::State Source::addTile(const TileID& id, const StyleUpdateParameters& p
                                                              std::move(monitor),
                                                              info.source_id,
                                                              parameters.style,
+                                                             parameters.mode,
                                                              callback);
         }
 
--- a/src/mbgl/map/tile_worker.cpp
+++ b/src/mbgl/map/tile_worker.cpp
@@ -21,13 +21,15 @@ TileWorker::TileWorker(TileID id_,
                        SpriteStore& spriteStore_,
                        GlyphAtlas& glyphAtlas_,
                        GlyphStore& glyphStore_,
-                       const std::atomic<TileData::State>& state_)
+                       const std::atomic<TileData::State>& state_,
+                       const MapMode mode_)
     : id(id_),
       sourceID(std::move(sourceID_)),
       spriteStore(spriteStore_),
       glyphAtlas(glyphAtlas_),
       glyphStore(glyphStore_),
-      state(state_) {
+      state(state_),
+      mode(mode_) {
 }
 
 TileWorker::~TileWorker() {
@@ -140,7 +142,8 @@ void TileWorker::parseLayer(const StyleLayer* layer, const GeometryTile& geometr
                                      spriteStore,
                                      glyphAtlas,
                                      glyphStore,
-                                     *collisionTile);
+                                     *collisionTile,
+                                     mode);
 
     std::unique_ptr<Bucket> bucket = layer->createBucket(parameters);
 
--- a/src/mbgl/map/tile_worker.hpp
+++ b/src/mbgl/map/tile_worker.hpp
@@ -3,6 +3,7 @@
 
 #include <mapbox/variant.hpp>
 
+#include <mbgl/map/mode.hpp>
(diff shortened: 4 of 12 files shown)


mikemorris:
This seems to fix some of the egregious clipping issues, but I'm still seeing some minor clipping.
- "Washington Monument & Grounds" clipped by "Maine Ave SW"
screen shot 2015-10-27 at 5 56 50 pm

- Clipping from tiles with missing labels
screen shot 2015-10-27 at 5 58 25 pm
screen shot 2015-10-27 at 6 00 28 pm

mikemorris:
Note that there are failing tests in the visual render suite because of this change. Is there an equivalent change that should be applied to mapbox-gl-js to generate updated fixtures, or is there an alternate workaround here @jfirebaugh?

jfirebaugh:
@mikemorris I believe the "Current" and "New" images are swapped. Should be:

**Current**
download-2

**New**
download-1

In the "New" image, you can see that many of the labels (those that come from points within tile buffer area) are drawn twice, or maybe more if they're near a tile corner. This is a problem for static renders.

To fix this, I believe we need to clip labels by tile boundaries, and at that point I think we should not have different layout and rendering pathways for Still vs Continuous mode:

* Always `addToBuffers` during layout
* Always clip labels to tile boundaries during painting

cc @ansis

kkaefer:
Ah dang, totally missed that the static API can render more than just individual tiles.

ansis:
> I believe we need to clip labels by tile boundaries

Yep, they need to be clipped by enabling the stencil test [here](https://github.com/mapbox/mapbox-gl-native/blob/1fcc606d271c0a643f788c5d79ff763528cfc6b5/src/mbgl/renderer/painter_symbol.cpp#L161-L170).

>  Is there an equivalent change that should be applied to mapbox-gl-js to generate updated fixtures

Yes, whatever changes are made in -native need to be added to -js.


> and at that point I think we should not have different layout and rendering pathways for Still vs Continuous mode

Separate paths allow for a pretty significant optimization. If labels are always clipped to tile boundaries, the quads used for rendering need to be included in multiple tiles not just one. Without clipping we can drop all the labels with anchors outside that tile's boundaries. Usually 50% - 66% of all labels are outside and can be dropped.

A single render path with clipping at tile boundaries would mean:
- 2x to 3x the amount of labels we need to generate quads for (slower initial loading)
- 2x to 3x bigger symbol buffers (how does this impact processed tile caching costs on mobile?)
- 2x to 3x more symbol geometries that are rendered each frame

jfirebaugh:
My concern with having a separate rendering pathway is that even if we fix this specific issue, it will still be possible to make a style where labels look fine in `MapMode::Continuous` but then get clipped when rendering static tiles. With large enough fonts or text spacing, that's always going to be an issue, right? So it would be better if you could spot style issues like that right away in Mapbox Studio or when testing your mobile app.

mb12:
@jfirebaugh Is still mode used to render vector tiles as PNG/JPG/raster tiles?

ansis:
@jfirebaugh that's a very good point. I hadn't thought about that

jfirebaugh:
My current thinking is that this approach is good enough as an expedient fix for 90% of the label rendering issues with static tiles. Let's get a test case for this in the test suite (hopefully City of London is a good one) and merge.

I expect that some cases will still have a rendering issue -- where a label is so big/long that it crosses over a tile despite not being anchored in the buffer area. This is probably what's happening with Colorado Springs in #3357. These will need to be dealt with by tweaking the style.

lucaswoj:
> These will need to be dealt with by tweaking the style.

Agreed. This is the whole point of vector tile buffers. We could go as far as to issue a style spec warning if `text-max-width` > buffer width

jfirebaugh:
Thinking about the Colorado Springs case again, I realized that it's probably the `MapMode::Continuous` analog of this same issue, and _not_ a "label too big/wide" issue. When the tile containing Colorado Springs is offscreen, we don't render the label _even if_ it's within the buffer of the tile immediately to the east, which is onscreen.

So I think we should merge this for the expedient fix, but then go back and make the more complex fix, where we always clip labels to tile boundaries during painting. In order to get some of the nice performance characteristics of the existing approach perhaps we can cull the symbol geometries to those that actually intersect the tile.

kkaefer:
Hmm, overall we'd like to move to a model that doesn't use clipping. In an interactive context, we can load adjacent tiles by widening the invisible buffer from the viewport to X pixels around it.

mikemorris:
Rebased this onto `master`, test suite failures all look like fixed clipped symbols (yay!) or overdrawing issues that would be fixed by `Always clip labels to tile boundaries during painting`, which still needs to be implemented.

mikemorris:
After clipping labels to tiles, I'm still seeing a single failing test cases locally for `text-rotate function`, wondering if the label clipping math fails to account for rotation properly.

ansis:
All tests are passing now. There was a small bug in the condition that determines when clipping happens. Merging

jfirebaugh:
Was there a test-suite test case added for this?

ansis:
> Was there a test-suite test case added for this?

No, most of the existing text tests have four tiles and already test tile crosstile clipping. I'm not sure how a clipping test would differ from any of the existing text tests. Do you have any ideas for tests that should be added?

jfirebaugh:
We should have a test case that was failing before the change, and passing after, like one of the regions from the screenshots above.

ansis:
pr for test: https://github.com/mapbox/mapbox-gl-native/pull/3457

Issue #1975, opened 2015-07-31
Performance regression between iOS 0.5.2 and 0.5.3


tomtaylor:
I've noticed a performance regression between iOS 0.5.2 and 0.5.3. I have a UI element which acts as a scrubber along a polyline ([video here](http://www.citycyclistapp.com)). I use a custom animation, driven by a CADisplayLink, to set the zoom and centre coordinate of the map every frame using a 3D spring behaviour.

In 0.5.2, the labels don't redraw until the animation stops, and during the animation the movement is smooth.

In 0.5.3, if the animation changes the zoom to the extent that the labels are due for refresh, it becomes jerky. If you pause the movement long enough for the labels to redraw, it becomes smooth again.

Let me know if you need any more information.

friedbunny:
This sounds like it would be related to @brunoabinader's label work in #1548.

/cc @incanus

brunoabinader:
> This sounds like it would be related to @brunoabinader's label work in #1548.

We have a fix for that in #1928, which landed on version `0.5.4`. @tomtaylor, is it possible to test it against version `0.5.4` instead of `0.5.3`? Also, I'd love to reproduce the custom animation if you have it handy.

tomtaylor:
Thanks - I'm away from my machine at the moment, but I'm pretty confident I tested against 0.5.4 with the same results. I'll get you the animation code when I get back.

tomtaylor:
OK, here's the animation driver code: https://gist.github.com/tomtaylor/91d29f47f281b9d96fcd

And I can confirm the same behaviour on 0.5.4. Let me know if I can give you any more info.

brunoabinader:
@tomtaylor I just merged #1979 into `master`; could you please check?

tomtaylor:
@brunoabinader happy to, do you think you could make me a new podspec file, pointing to a precompiled .zip version?

tomtaylor:
@brunoabinader I just managed to try this out on 2.1.0-pre.1, and I don't think it's quite fixed yet. I recorded a video through Quicktime from my 5s, which shows what I'm experiencing. For context, I'm touching on the scrubber at the bottom to zoom in at that point on the route, driven by the animation code I shared previously. Let me know if I can give you any more information.

https://s3-eu-west-1.amazonaws.com/files.tomtaylor.co.uk/data/mapbox-lag.mov

(On the flip side, it seems to perform even better when all the labels have rendered!)

[...]

PR #2852, opened 2015-10-29
Allow users to notify us when they are driving our maps via their own…

adam-mapbox:
… animation.

#1975 

@brunoabinader @kkaefer @incanus @tomtaylor

This is our best idea for how to support multiple use cases.  Basically the idea is this: it would be difficult and brittle for us to detect when we're being driven by an external animation source (e.g. CADisplayLink).  So we allow the user to set this hint, which improves their performance while animating.  Basically we can skip certain operations because we know that we're going to be forced to render every frame.  Then they clear it when done.  The only failure mode is that if they set this, and never clear it, we could end up in situations where fades don't fully perform because we're no longer being driven, and so we don't continue to render.  If they never set it and ignore it completely, the only behavior regression(s) are related to performance, not correctness.

Auto-detecting external animation would require us to detect that we're being modified every frame, and then force us to include an auto-canceling dispatch_async that detected when we were no longer being modified every frame.  It would be fragile and error-prone and neither I nor @incanus  were in favor of that option.

### DIFF
--- a/include/mbgl/ios/MGLMapView.h
+++ b/include/mbgl/ios/MGLMapView.h
@@ -304,6 +304,10 @@ IB_DESIGNABLE
 *   @param animated If `YES`, the callout view is animated offscreen. */
 - (void)deselectAnnotation:(id <MGLAnnotation>)annotation animated:(BOOL)animated;
 
+#pragma mark - User Animations
+
+- (void)setUserAnimating:(BOOL)userAnimating;
+
 #pragma mark - Adding Overlays
 
 /** @name Adding Overlays */
--- a/include/mbgl/map/map.hpp
+++ b/include/mbgl/map/map.hpp
@@ -91,6 +91,10 @@ public:
     bool isRotating() const;
     bool isScaling() const;
     bool isPanning() const;
+    
+    // User Animating
+    void setUserAnimating(bool);
+    bool isUserAnimating() const;
 
     // Camera
     void jumpTo(CameraOptions options);
--- a/platform/ios/MGLMapView.mm
+++ b/platform/ios/MGLMapView.mm
@@ -1646,6 +1646,11 @@ std::chrono::steady_clock::duration secondsAsDuration(float duration)
     return [NSSet setWithObject:@"camera"];
 }
 
+- (void)setUserAnimating:(BOOL)userAnimating
+{
+    _mbglMap->setUserAnimating(userAnimating);
+}
+
 - (double)zoomLevel
 {
     return _mbglMap->getZoom();
--- a/src/mbgl/map/map.cpp
+++ b/src/mbgl/map/map.cpp
@@ -130,6 +130,14 @@ bool Map::isGestureInProgress() const {
     return transform->isGestureInProgress();
 }
 
+void Map::setUserAnimating(bool userAnimating) {
+    context->invokeSync<void>(&MapContext::setUpdateHint, userAnimating ? UpdateHint::UserAnimating : UpdateHint::Nothing);
+}
+
+bool Map::isUserAnimating() const {
+    return context->invokeSync<UpdateHint>(&MapContext::getUpdateHint) == UpdateHint::UserAnimating;
+}
+    
 bool Map::isRotating() const {
     return transform->isRotating();
 }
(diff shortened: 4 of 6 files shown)


jfirebaugh:
@adam-mapbox Can you explain how this improves performance for the external animation case?

incanus:
The basic gist here is that it avoids the zoom-related painter animation changes (i.e., label fades) done as a result of positive `if (painter->needsAnimation())` checks. These checks are expensive because they examine `frameHistory` to make determinations about the trend of the animation we are currently in. However, when the user is animating externally (as with `CADisplayLink`), this trend doesn't matter as the user is going to animate anyway, i.e. we don't need to determine if the trend justifies our own, internal successive frames.

adam-mapbox:
I'll be the first to admit that I don't understand 100% of what's going on, but the gist is that when we know the user is responsible for feeding us updates every frame, we can skip a number of different checks and updates because they'll get taken care of as a consequence of just rendering every frame.  For example in this case we're tracking unfinished fades and making sure we finish them by continuing to draw, but we don't need to do that because we're going to keep drawing anyway.

jfirebaugh:
This feels to me like something that should be addressed by making our existing code paths more efficient, rather than adding a public API, which a user could easily overlook or forget to call.

Is it the actual code run in `painter->needsAnimation()` that's the problem, or something triggered by the subsequent `Update::Repaint`?

incanus:
> This feels to me like something that should be addressed by making our existing code paths more efficient, rather than adding a public API, which a user could easily overlook or forget to call.

Discussed this and we should _also_ make the existing code paths more efficient, but I'm not sure we can do that by the order of magnitude necessary to get under the 100ms frame times we're seeing right now. We should check, though; you're right. 

With regard to the user forgetting the API, this is a bit of an edge case (using something like an `CADisplayLink` to drive your own animation frame draws), though it's a legit one. In this case, doing such things and forgetting to use the appropriate API is on the user. That said, we should reduce or eliminate tripwires — we more wanted to start that discussion here.

adam-mapbox:
What @incanus said.  I'm absolutely in favor of optimizing the underlying pathways, but I'm not sure that doing so obviates the need for a way for users to inform us of their usage pattern.  There is precedent for this, for example in the OpenGL API, ref: https://www.opengl.org/wiki/GLAPI/glBufferData usage hints.  We'll always do a better job if we know what the user is doing with us than if we don't.  Again, if the user *completely ignores* this new API, we still do things correctly, just with lower performance.

adam-mapbox:
One further comment: as for this being an "edge case", I actually think that it should be something that we not only support but encourage.  Using external drivers to animate widgets is a time-honored concept, and it actually works pretty darn well and opens the door to a lot of creativity - if managed correctly.

jfirebaugh:
If we fully understand why performance is the way that it is, and have exhausted reasonable attempts to improve it, then indeed we should think about adding a public API. I just don't think we've met those preconditions yet. My gut says we can understand and fix the underlying issue.

incanus:
That sounds fair to me; let's keep pushing here.

adam-mapbox:
I respectfully disagree; I haven't heard a compelling argument (to me) that this PR is substantively unwise.  I think it's the right way to solve this issue at the moment and will be a worthwhile addition moving forward as well, for the reasons I outlined.

incanus:
One thing we _can_ do with PR right now is roll a build for @tomtaylor to check out in the context of his bug brought up in #1975, to get some :eyes: on it in parallel. How does that sound @tomtaylor? 

Otherwise, since he is the only one hitting it, and this PR code is not particularly likely to go way stale and/or hard to migrate, we can push on the underlying perf issue as a technical debt issue that may bite us elsewhere if we don't find it.

adam-mapbox:
I will take a look at why this call is so expensive and what might be done about it.

jfirebaugh:
It is wise to keep public APIs as small as possible -- unless there's a compelling need for an API function, it shouldn't be added, as doing so adds overhead in maintenance, testing, documentation, support, and end user comprehension.

If it's possible to address the root cause issue here, it will mean external animations "just work" without the need to twiddle a particular bit in mbgl. We won't have to write tests for that setting, document it, or field support questions from people who overlooked it and are wondering why performance is bad. Furthermore, depending on what the cause turns out to be, we may be able improve performance in non-animated scenarios as well.

tomtaylor:
@adam-mapbox @incanus @jfirebaugh just so you know, from my perspective this flag fixes the issue I'm seeing. Happy to keep testing if you choose a different approach for the public API.

incanus:
Fixed without external-facing API in https://github.com/mapbox/mapbox-gl-native/pull/2922.

PR #2859, opened 2015-10-29
[android] Allow apps to override map resources

### LINKED ISSUE #2858: Make GPS and compass resources public to allow for customisation
In the public resource XML we should allow apps to override these resources:
- compass bitmap
- GPS marker bitmaps (and the colour of the accuracy ring which should be moved to a resource instead of a static value)
- Mapbox logo?
- attribution button

The use case here is say an app wants a pink GPS marker to match their app's theme:
- they copy our GPS bitmap
- edit the colour to pink
- override the resources
- also change the accuracy ring colour

/cc @tobrun

ljbade:
Fixes #2858


ljbade:
Renamed.

PR #2863, opened 2015-10-29
[core] Set names to threads on Linux

tmpsantos:
Makes debugging/profiling easier.


ljbade:
This should benefit Android too.

PR #2866, opened 2015-10-29
[android] #2861 - double tap zoom gesture while tracking mode is enabled

### LINKED ISSUE #2861: Double tap zoom with location tracking zooms to wrong spot
When user location tracking is enabled, the double tap zoom does not zoom around the centre of the map.

/cc @tobrun

tobrun:
closes #2861


ljbade:
:+1:

PR #2867, opened 2015-10-29
[android] #2864 - improved compass update cycle, removed lowpassfilter

### LINKED ISSUE #2864: Compass filter is too aggressive
Comparing our compass to Google's we need to turn down the filtering a bit.

Is it adding quite a bit of delay currently.

/cc @tobrun

tobrun:
closes #2864

### DIFF
--- a/android/MapboxGLAndroidSDK/src/main/java/com/mapbox/mapboxsdk/views/UserLocationView.java
+++ b/android/MapboxGLAndroidSDK/src/main/java/com/mapbox/mapboxsdk/views/UserLocationView.java
@@ -37,8 +37,6 @@ import com.mapzen.android.lost.api.LocationRequest;
 import com.mapzen.android.lost.api.LocationServices;
 import com.mapzen.android.lost.api.LostApiClient;
 
-import java.util.ArrayDeque;
-
 final class UserLocationView extends View {
 
     private MapView mMapView;
@@ -382,8 +380,7 @@ final class UserLocationView extends View {
         private GeomagneticField mGeomagneticField;
 
         // Controls the sensor update rate in milliseconds
-        private static final int UPDATE_RATE_MS = 500;
-        private AngleLowPassFilter mLowPassFilter;
+        private static final int UPDATE_RATE_MS = 300;
 
         // Compass data
         private float mCompassBearing;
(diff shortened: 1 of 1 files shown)


android/MapboxGLAndroidSDK/src/main/java/com/mapbox/mapboxsdk/views/UserLocationView.java
```diff
@@ -414,6 +410,11 @@ public void onSensorChanged(SensorEvent event) {
                 return;
             }
 
+            long currentTime = System.currentTimeMillis();
```
jush: I think it would be better to avoid using `System.currentTimeMillis()` and instead use [SystemClock](http://developer.android.com/reference/android/os/SystemClock.html) `elapsedTime` method.

From the `SystemClock` javadoc: 
>System.currentTimeMillis() is the standard "wall" clock (time and date) expressing milliseconds since the epoch. The wall clock can be set by the user or the phone network (see setCurrentTimeMillis(long)), so the time may jump backwards or forwards unpredictably.

tobrun:
@jush 
Thank you for bringing that up, indeed the docs say it's better to use `elapsedTime` for this:
>  elapsedRealtime() and elapsedRealtimeNanos() return the time since the system was booted, and include deep sleep. This clock is guaranteed to be monotonic, and continues to tick even when the CPU is in power saving modes, so is the recommend basis for **general purpose interval timing**.

ljbade:
:+1:

android/MapboxGLAndroidSDK/src/main/java/com/mapbox/mapboxsdk/views/UserLocationView.java
```diff
@@ -392,12 +389,11 @@ public int getMyBearingTrackingMode() {
         public MyBearingListener(Context context) {
             mSensorManager = (SensorManager) context.getSystemService(Context.SENSOR_SERVICE);
             mSensorRotationVector = mSensorManager.getDefaultSensor(Sensor.TYPE_ROTATION_VECTOR);
-            mLowPassFilter = new AngleLowPassFilter();
         }
 
         public void onStart(Context context) {
             mRotationDevice = ((WindowManager) context.getSystemService(Context.WINDOW_SERVICE)).getDefaultDisplay().getRotation();
-            mSensorManager.registerListener(this, mSensorRotationVector, UPDATE_RATE_MS * 1000);
+            mSensorManager.registerListener(this, mSensorRotationVector, UPDATE_RATE_MS * 2500);
```
jush: out of curiosity why are you multiplying by `2500` instead of `1000` to get microseconds?

tobrun: @jush yeah, I should have created two separate variables: one for UI update rate and one for sensor update rate. Will refactor that in the near future.

PR #2869, opened 2015-10-29
Monitor annotation tiles

### LINKED ISSUE #1688: Prevent flickering when adding/removing annotations

jfirebaugh:
Fixes #1688.

:eyes: @kkaefer

### DIFF
--- a/src/mbgl/map/tile_worker.cpp
+++ b/src/mbgl/map/tile_worker.cpp
@@ -287,6 +287,6 @@ void TileWorker::createSymbolBucket(const GeometryTileLayer& layer,
 
 void TileWorker::insertBucket(const std::string& name, std::unique_ptr<Bucket> bucket) {
     if (bucket->hasData()) {
-        result.buckets.emplace_back(name, std::move(bucket));
+        result.buckets.emplace(name, std::move(bucket));
     }
 }
--- a/src/mbgl/map/tile_worker.cpp
+++ b/src/mbgl/map/tile_worker.cpp
@@ -18,10 +18,8 @@ using namespace mbgl;
 TileWorker::TileWorker(TileID id_,
                        std::string sourceID_,
                        Style& style_,
-                       std::vector<util::ptr<StyleLayer>> layers_,
                        const std::atomic<TileData::State>& state_)
-    : layers(std::move(layers_)),
-      id(id_),
+    : id(id_),
       sourceID(sourceID_),
       parameters(id.z),
       style(style_),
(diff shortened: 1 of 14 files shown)


jfirebaugh:
The `Annotations.NonImmediateAdd` test case is not generating the expected results. Investigating.

jfirebaugh:
It's failing because `TileWorker` has a copy of the `StyleLayers` from before the shape layer is added.

src/mbgl/map/tile_worker.cpp
```diff
@@ -91,11 +90,8 @@ void TileWorker::redoPlacement(
     // Reset the collision tile so we have a clean slate; we're placing all features anyway.
     collisionTile = std::make_unique<CollisionTile>(config);
 
-    for (auto i = layers.rbegin(); i != layers.rend(); i++) {
```
jfirebaugh: @kkaefer Do you see any issues here with iterating `buckets` directly, rather than `layers`?

jfirebaugh: Ah, yeah, feature placement is order dependent, of course,

src/mbgl/map/tile_worker.cpp
```diff
@@ -287,6 +287,6 @@ void TileWorker::createSymbolBucket(const GeometryTileLayer& layer,
 
 void TileWorker::insertBucket(const std::string& name, std::unique_ptr<Bucket> bucket) {
     if (bucket->hasData()) {
-        result.buckets.emplace_back(name, std::move(bucket));
+        result.buckets.emplace(name, std::move(bucket));
```
kkaefer: Are we ever overwriting buckets here?

src/mbgl/map/tile_worker.hpp
```diff
@@ -31,7 +31,7 @@ class GeometryTileLayer;
 class TileParseResultBuckets {
 public:
     TileData::State state = TileData::State::invalid;
-    std::vector<std::pair<std::string, std::unique_ptr<Bucket>>> buckets;
+    std::unordered_map<std::string, std::unique_ptr<Bucket>> buckets;
```
kkaefer: Why a map? I don't think we're using the index anywhere.

src/mbgl/map/tile_worker.cpp
jfirebaugh: I don't think so -- if we are, there's a bug in the bucket de-duping step. The name is the bucket name, which should be unique.

src/mbgl/map/tile_worker.hpp
jfirebaugh: I changed this so that I could do `buckets = std::move(resultBuckets.buckets);` on the `VectorTileData`-side result of `parseAllLayers`. In the "all layers" case, we want to remove any old buckets that no longer exist (e.g., removed annotations).

PR #2873, opened 2015-10-30
[core] Polymorphic bucket creation

### LINKED ISSUE #2820: Place symbols in correct order
As a regression from #2617 (see https://github.com/mapbox/mapbox-gl-native/pull/2617/files#r43035678), we no longer respect the order of placement during the initial parse step. Parsing order doesn't matter, but as part of the parse step, we're also placing labels.

We could either make sure that we stop placing labels after a symbol layer wasn't parsed yet, or stop parsing layers after a symbol layer wasn't parsed.

jfirebaugh:
Fifth in a series of PRs that builds towards a full-fledged [programmatic style API](https://github.com/mapbox/mapbox-gl-native/issues/837). This is a big chunk of work, but I'm trying to break it up step by step and not create a long-lived branch with a lot of changes.

This episode moves further towards being able to remove the generic `ClassProperties` in favor of concrete paint and layout classes like those found in [style_properties.hpp](https://github.com/mapbox/mapbox-gl-native/blob/bb6241844a092cf632fe740e6ac79c920117dd4e/include/mbgl/style/style_properties.hpp). The way that I'm going to do that is to continue refactoring until all usage of `ClassProperties` is in layer subclasses, and then substitute the concrete per-layer replacements.

This PR eliminates `StyleBucket` and moves **bucket creation** to layer subclasses. Along the way I believe to have fixed #2820, by restoring the previous use of a `partialParse` boolean.


kkaefer:
> restoring the previous use of a `partialParse` boolean.

This means that we don't parse any symbol buckets that are after one that is missing. We could still parse them, but not place them.

jfirebaugh:
I _think_ it's set up so that they are parsed, but not placed: `SymbolLayer::createBucket` calls `parseFeatures` unconditionally, but `addFeatures` only if `partialParse` is false:

https://github.com/mapbox/mapbox-gl-native/pull/2873/files#diff-511e12d83fd87b94fa15dcdbf0d27bbcR159

Then `TileWorker` adds the bucket to the pending set if `partialParse` is true and it's a symbol bucket:

https://github.com/mapbox/mapbox-gl-native/blob/ca25594d28a15835dca4655b502c6fb04439f298/src/mbgl/map/tile_worker.cpp#L140

Let me know if that looks correct.

kkaefer:
There are actually three steps:

* **`parseFeatures`**: determines the text we need to display by replacing tokens and applying text transformations, and the glyphs we need
* **`addFeatures`**: determines the potential position of labels
* **`placeFeatures`**: actually places them and adds them to a buffer

`placeFeatures` is currently called as the [last step in `addFeatures`](https://github.com/mapbox/mapbox-gl-native/blob/1688-monitor-annotation-tiles/src/mbgl/renderer/symbol_bucket.cpp#L269), but there's no good reason to; we could separate that out and call the `addFeatures` without placing as soon as we have the symbols. However, that means we'd have to introduce a third step beyond `pending`.

PR #2874, opened 2015-10-30
[android] When pinch zooming lock rotation

### LINKED ISSUE #813: Lock rotation when pinch-zooming
On Android, it is very easy to change the rotation of the map when zooming. iOS has some mechanism to prevent that. We should have the same behavior on Android too (Google Maps does the same).

ljbade:
Fixes #813


tobrun:
Awesome :+1:

ljbade:
This branch was merged but for some reason did not close automatically despite pushing the rebased branch.

Issue #1504, opened 2015-05-11
Annotation tap gesture recognizer queries hard-coded region, ignoring image sizes


jfirebaugh:
Tapping on a point annotation in the iOS test app seems to open a callout as expected in only 15% or so of taps.

incanus:
Can you get more specific about how close nearby annotations are visually, what point you are tapping (as logged in debug) and how that compares to the annotation you were trying to tap? 

Because annotations are in GL and aren't actual tappable views, we are querying a fuzzy region around the user's tap to try to guess which annotations they are referring to, as well as iterating overlapping annotations like MapKit does to enable selection of obscured markers. 

https://github.com/mapbox/mapbox-gl-native/blob/44cb91c656ecfa3a33340f837460b3f527874380/platform/ios/MGLMapView.mm#L993-L1117

In addition, this heuristic can vary depending on marker image size, as that affects the user's perceived tappable area while not changing the region query (though we [do weight the region](https://github.com/mapbox/mapbox-gl-native/blob/44cb91c656ecfa3a33340f837460b3f527874380/platform/ios/MGLMapView.mm#L1016-L1021) to account for top-heavy markers). 

In short, this has felt good for months now, so I'm curious what changed and what the specific scenario is that is triggering this impression.

incanus:
More background: https://github.com/mapbox/mapbox-gl-native/pull/1037#issuecomment-84521846

jfirebaugh:
I'm testing on device using the Emerald style and "Add 100 Points", and zooming into an area where there is only one marker. It looks like the problem is that the hitbox for the marker is about half the size of the actual marker in Emerald, so only taps on the lower half of the marker image register.

incanus:
This is probably a result then of: 

1. Insufficient tap-testing with Emerald
1. Emerald's marker being so much bigger than Streets' (and Light/Dark's)

Streets: 

![](https://raw.githubusercontent.com/mapbox/mapbox-gl-styles/mb-pages/sprites/mapbox-streets%402x.png)

Emerald: 

![](https://raw.githubusercontent.com/mapbox/mapbox-gl-styles/mb-pages/sprites/emerald%402x.png)

However, this highlights the difficult with the stopgap method of limiting marker imagery to the sprite sheet while simultaneously trying to minimize communications between the GL sprite parsing layer and the iOS code relation to gestures. 

Possible solutions: 

* Tweak annotation gesture algorithms
[...]

PR #2876, opened 2015-10-30
[android] Adjust annotation touch tolerance to match default marker

ljbade:
For #1504


tobrun:
Would it be possible to measure the annotation and provide dynamic touch targets?
PR review wise, this looks good to merge

ljbade:
@tobrun Yeah that is next step in #1504. Wanted to get this in master ASAP in case it takes a while to implement dynamic bit.

PR #2878, opened 2015-10-30
[android] Save added markers in test app

### LINKED ISSUE #2877: Test app should save markers
After https://github.com/mapbox/mapbox-gl-native/pull/2855 it should now be possible to save the added `MarkerOptions` so the markers remain after a screen rotation.

ljbade:
Fixes #2877


tobrun:
:+1:

PR #2889, opened 2015-10-30
Create UIActionSheet on demand

### LINKED ISSUE #2475: Attempting to load the view of a view controller while it is deallocating...
I receive the following error message in the console when navigating away from a view that contains a MGLMapView.

```
2015-09-30 17:45:48.716 TallyGo[6181:50165274] Attempting to load the view of a view controller while it is deallocating is not allowed and may result in undefined behavior (<UIAlertController: 0x7f875bfd9940>)
```

I believe I have tracked it down to the UIActionSheet being instantiated here:
https://github.com/mapbox/mapbox-gl-native/blob/master/platform/ios/MGLMapView.mm#L301-L309

Not sure the best way to fix. Commenting out these lines make the error go away.

1ec5:
Deferring the UIActionSheet’s initialization eliminates some console spew about potential undefined behavior.

Fixes #2475.

### DIFF
--- a/platform/ios/MGLMapView.mm
+++ b/platform/ios/MGLMapView.mm
@@ -276,22 +276,6 @@ std::chrono::steady_clock::duration secondsAsDuration(float duration)
     [self addSubview:_attributionButton];
     _attributionButtonConstraints = [NSMutableArray array];
 
-    _attributionSheet = [[UIActionSheet alloc] initWithTitle:@"Mapbox iOS SDK"
-                                                    delegate:self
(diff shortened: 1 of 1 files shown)


friedbunny:
Awesome. :+1:

systemed:
Ha, brilliant. I just this minute Googled this error string and was directed here by [StackOverflow](http://stackoverflow.com/questions/32704169/attempting-to-load-the-view-of-a-view-controller-while-it-is-deallocating-ui)... only to find you fixed it one hour ago. :trophy:

PR #2727, opened 2015-11-02
Move request retries to DefaultFileSource

kkaefer:
We currently have the logic to retry requests stuffed away in the platform-specific implementations. Instead, we should move this logic to the DefaultFileSource (it's currently the same anyway).


kkaefer:
@jfirebaugh @tmpsantos :eyes:

PR #2902, opened 2015-11-02
ignore -Wunused-variable around libtess2

mikemorris:
Fixes the annoying warning below.

```
../src/libtess2/sweep.c:1139:6: warning: unused variable 'fixedEdges' [-Wunused-variable]
        int fixedEdges = 0;
            ^
  CC(target) Release/obj.target/core/src/libtess2/tess.o
1 warning generated.
```


jfirebaugh:
I don't think this will fix that warning. The warning is produced when compiling the libtess sources (`sweep.c`), not mbgl sources.

mikemorris:
Oh, yea. :disappointed:

Issue #2332, opened 2015-09-15
Xcode 6.x cannot link against Bitcode-enabled libMapbox.a


incanus:
Since we landed Bitcode in `ios-v2.1.1`, device builds for this test project, as well as dependent projects, don't work with target build errors such as: 

```
duplicate symbol _llvm.cmdline in:
    /Users/incanus/Documents/Projects/Mapbox/sirius-ios/Pods/Mapbox-iOS-SDK/libMapbox.a(SMCalloutView.o)
    /Users/incanus/Documents/Projects/Mapbox/sirius-ios/Pods/Mapbox-iOS-SDK/libMapbox.a(NSProcessInfo+MGLAdditions.o)
duplicate symbol _llvm.embedded.module in:
    /Users/incanus/Documents/Projects/Mapbox/sirius-ios/Pods/Mapbox-iOS-SDK/libMapbox.a(SMCalloutView.o)
    /Users/incanus/Documents/Projects/Mapbox/sirius-ios/Pods/Mapbox-iOS-SDK/libMapbox.a(NSProcessInfo+MGLAdditions.o)
duplicate symbol _llvm.cmdline in:
    /Users/incanus/Documents/Projects/Mapbox/sirius-ios/Pods/Mapbox-iOS-SDK/libMapbox.a(SMCalloutView.o)
    /Users/incanus/Documents/Projects/Mapbox/sirius-ios/Pods/Mapbox-iOS-SDK/libMapbox.a(NSString+MGLAdditions.o)
duplicate symbol _llvm.embedded.module in:
    /Users/incanus/Documents/Projects/Mapbox/sirius-ios/Pods/Mapbox-iOS-SDK/libMapbox.a(SMCalloutView.o)
    /Users/incanus/Documents/Projects/Mapbox/sirius-ios/Pods/Mapbox-iOS-SDK/libMapbox.a(NSString+MGLAdditions.o)
ld: 36 duplicate symbols for architecture armv7
```

Per @friedbunny, this might be a lead: 

http://www.openradar.me/21884601

> I accidentally enabled bitcode for the simulator build of my library

/cc @1ec5

incanus:
Things are broke in Xcode 6.4 as well, since `ios-v2.1.1` was generated with Xcode 7 GM. 

Per the above OpenRadar link, it's not clear how you can conditionally enable Bitcode based on _architecture_ vs. _build configuration_. 

This might be an Xcode 7 GM issue, too.

[...]

PR #2584, opened 2015-10-09
Install GeoJSONVT 2.1.6 from mason

mikemorris:
Fixes #2226, #898 

- Defines `cxx_host` gyp variable to set `-fabi-version=0` on non-clang builds, fixes https://github.com/mapbox/mapbox-gl-native/issues/2502
- Sets `'GCC_SYMBOLS_PRIVATE_EXTERN': 'YES'` to prevent linker warnings on OS X builds using mason binaries built with [`-fvisibility-inlines-hidden=hidden`](https://github.com/mapbox/mason/blob/8a63d5a84f41c6686f0c7fd1ce49babff91df1bc/mason.sh#L56) or [`'GCC_SYMBOLS_PRIVATE_EXTERN': 'YES'`](https://github.com/mapbox/geojson-vt-cpp/blob/visibility-hidden/deps/common.gypi#L26)
- Links `libgeojsonvt.a` in `test/ios/ios-tests.xcodeproj` to fix an undefined symbol (unused in `libmbgl-core.a`) when linking iOS tests.
- Sets `Inline Methods Hidden` in `test/ios/ios-tests.xcodeproj` to fix linker warnings.

**TODO**:
- [ ] ~~https://github.com/mapbox/mapbox-gl-native/pull/2582 should be merged first for `gcc-5`, necessary for compatibility with mason libraries build with `clang-3.5` (https://github.com/mapbox/mapbox-gl-native/issues/2502)~~
- [x] Pull request and merge geojson-vt-cpp branch [`visibility-hidden`](https://github.com/mapbox/geojson-vt-cpp/tree/visibility-hidden) and tag a new release (v2.1.6.2).
- [x] Add a new mason package with the `-fvisibility=hidden` geojson-vt-cpp release
- [ ] After merging this, rebase https://github.com/mapbox/mapbox-gl-native/pull/2352, preserving @lucaswoj changes to build without a mason submodule

/cc @jfirebaugh @1ec5 @incanus @springmeyer

PR #2903, opened 2015-11-02
Force Bitcode in make ipackage

1ec5:
`make ipackage` now performs a Bitcode-enabled archive build for devices, making it possible to archive a Bitcode-enabled application that links against the Mapbox iOS SDK.

@incanus, can you verify that the resulting static library contains all the expected symbols? I did notice some additional missing symbol warnings in the `CreateUniversalBinary` step, such as the one below, but I think those are the same things we try to suppress when calling `libtool` with the `-no_warning_for_no_symbols` flag:

```
/Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/libtool: for architecture: x86_64 file: /Users/mxn/hub/mapbox-gl-native/build/mbgl.build/Release-iphonesimulator/core.build/Objects-normal/x86_64/libmbgl-core.a(camera.o) has no symbols
```

/ref #2332
/cc @mikemorris

### DIFF
--- a/scripts/ios/package.sh
+++ b/scripts/ios/package.sh
@@ -53,6 +53,8 @@ if [[ "${BUILD_FOR_DEVICE}" == true ]]; then
         ARCHS="arm64 armv7 armv7s" \
         ONLY_ACTIVE_ARCH=NO \
         GCC_GENERATE_DEBUGGING_SYMBOLS=${GCC_GENERATE_DEBUGGING_SYMBOLS} \
+        ENABLE_BITCODE=YES \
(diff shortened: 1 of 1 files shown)


mikemorris:
These are the same warnings I'm seeing in Xcode when building a CocoaPods project with `libMapbox.a` copied over from https://github.com/mapbox/mapbox-gl-native/pull/2584

incanus:
Looking good here.

incanus:
Merging ahead of Travis running here, since this only affects the iOS package script and we don't build iOS on Travis.

PR #2908, opened 2015-11-03
Speed up time to first render call (again)

kkaefer:
Seems like #2138 got reverted, but was never revisited. Since this boosts performance quite a bit, we should bring it back.


kkaefer:
@incanus :eyes: same patch as before, I can't find any crashes.

PR #2912, opened 2015-11-03
Strip whitespace and newlines from access token and style URL

1ec5:
Just in case browser weirdness or a tendency to triple-click-select is causing some cases of #2012.

Ordinarily we’d want this kind of logic down in mbgl, but I think it belongs up here too because:

* The access token is stored as an instance variable of MGLAccountManager, and I think that ivar should have the correct value too.
* The inspectable style URL string gets transformed into an NSURL in `-[MGLMapView setStyleURL__:]`. NSURL demands RFC 1738/2732 well-formedness, so if the string has leading whitespace, the transformation will fail and you’ll wind up w/ no URL (and the default style).

/cc @friedbunny

### DIFF
--- a/platform/ios/MGLAccountManager.m
+++ b/platform/ios/MGLAccountManager.m
@@ -60,6 +60,8 @@
 }
 
 + (void) setAccessToken:(NSString *) accessToken {
+    accessToken = [accessToken stringByTrimmingCharactersInSet:
+                   [NSCharacterSet whitespaceAndNewlineCharacterSet]];
     if ( ! [accessToken length]) return;
     
     [MGLAccountManager sharedManager].accessToken = accessToken;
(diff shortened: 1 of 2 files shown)


friedbunny:
Very prudent, thanks for the placement explanation. :heart:

ljbade:
@1ec5 Can you open similar for Android?

Also does this detect too-short access keys? A person in #2012 mentioned that happened to them.

1ec5:
> Also does this detect too-short access keys? A person in #2012 mentioned that happened to them.

No, and I think any stricter validation belongs in mbgl proper. This PR is only about URL well-formedness.

1ec5:
The Bitrise test failure is being tracked as #2920. It’s unrelated to this change; merging.

Issue #2926, opened 2015-11-04
iOS crash in variant.hpp


picciano:
Saw a new type of crash from yesterday's build. Crashlytics reported this as Crashed: Map
SIGABRT ABORT 0x0000000194e7b270

```
Thread : Crashed: Map
0  TallyGo                        4297510532 CLSProcessRecordAllThreads + 4297543300
1  TallyGo                        4297510532 CLSProcessRecordAllThreads + 4297543300
2  TallyGo                        4297511732 CLSProcessRecordAllThreads + 4297544500
3  TallyGo                        4297450860 CLSHandler + 4297483628
4  TallyGo                        4297504656 __CLSExceptionRecord_block_invoke + 4297537424
5  libdispatch.dylib              6791846228 _dispatch_client_callout + 16
6  libdispatch.dylib              6791885284 _dispatch_barrier_sync_f_invoke + 76
7  TallyGo                        4297503528 CLSExceptionRecord + 4297536296
8  TallyGo                        4297502344 CLSTerminateHandler() + 4297535112
9  libc++abi.dylib                6776642484 std::__terminate(void (*)()) + 16
10 libc++abi.dylib                6776640632 __cxxabiv1::exception_cleanup_func(_Unwind_Reason_Code, _Unwind_Exception*) + 134
11 TallyGo                        4298359200 mapbox::util::geojsonvt::Convert::calcRingBBox(mapbox::util::geojsonvt::ProjectedPoint&, mapbox::util::geojsonvt::ProjectedPoint&, mapbox::util::geojsonvt::ProjectedGeometryContainer const&) (variant.hpp:640)
[...]

PR #2930, opened 2015-11-04
[core] Simplify functions

jfirebaugh:
Sixth in a series of PRs that builds towards a full-fledged [programmatic style API](https://github.com/mapbox/mapbox-gl-native/issues/837). This is a big chunk of work, but I'm trying to break it up step by step and not create a long-lived branch with a lot of changes.

This episode simplifies the classes and structs used for representing property values. Previously, property values were represented as a variant type:

```
template <typename T>
using Function = mapbox::util::variant<
    std::false_type,
    ConstantFunction<T>,
    StopsFunction<T>
>;
```

This eliminates the variant in favor of a unified `Function<T>` class template with an internal array of stops. For constant values, the array contains a single element `{0, value}`. An indeterminate state (`std::false_type` in the variant) is not needed.

In addition, I eliminated `PiecewiseConstantFunction`. Contrary to its name, this was not actually used for piecewise constant functions -- it is actually used to implement the cross-fading behavior of the `*-pattern` and `line-dasharray` properties. Now this is accomplished by a partial specialization `Function<Faded<T>>` instead.


brunoabinader:
:+1: this is nice!

picciano:
You think this might fix https://github.com/mapbox/mapbox-gl-native/issues/2926?

jfirebaugh:
No, this is unrelated to #2926.

PR #2933, opened 2015-11-04
[core] Remove the ability to change cross-fade durations via transition

jfirebaugh:
We shouldn't overload transition duration to mean the duration of the cross-fade when moving between stops. If we want to allow styles to override the default 300ms fade duration, it should be a dedicated property like `raster-fade-duration`.

I don't think any styles use this functionality -- if they did, they were broken for `MapMode::Still` rendering.

:eyes: @kkaefer @ansis


ansis:
Sounds good to me. I think this also needs to be changed in -js

brunoabinader:
:thumbsup:

PR #2934, opened 2015-11-04
Disable callout view dimming when tap is a no-op

### LINKED ISSUE #2634: iOS callout now tappable like button all the time
As of https://github.com/mapbox/mapbox-gl-native/pull/2596 the whole callout taps like a button, including dimming. This should default to off somehow.

1ec5:
Fixes #2634.

@incanus, I left the dimming on if it looks like the delegate is going to respond to the tap (thereby treating it like a button). If you disagree, I can turn off the dimming in that case too.

### DIFF
--- a/platform/ios/MGLMapView.mm
+++ b/platform/ios/MGLMapView.mm
@@ -1453,6 +1453,11 @@ std::chrono::steady_clock::duration secondsAsDuration(float duration)
     }
 }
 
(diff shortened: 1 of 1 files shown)


incanus:
That looks great. FWIW Maps.app has a chevron for details, but the whole callout gets tapped.

PR #2941, opened 2015-11-05
[android] Check for location permission in MyLocationTrackingModeActivity

### LINKED ISSUE #2110: Support Android M permissions
Android M has new permission model. We need to support this.

From what I can tell this only affects the GPS location support.

More info in https://developer.android.com/preview/features/runtime-permissions.html

Currently in the Android M simulator the test app crashes as soon as you hit the GPS button.

ljbade:
Fixes #2110


tobrun:
:+1:

PR #2943, opened 2015-11-05
[core] Replace GL save/restore values mechanism with dirty flag

brunoabinader:
We should not depend on a previous GL state, but set all required GL values instead. Flagging a GL value as dirty means that the next time it is set (via `operator=`) we'll call the respective GL setter commands independent whether the cached value is the same.

/cc @kkaefer @jfirebaugh @tmpsantos @adam-mapbox


jfirebaugh:
:+1:

PR #2950, opened 2015-11-05
Trimming Whitespace In setAccessToken()

### LINKED ISSUE #2919: MapView.setAccessToken() should trim whitespace
Per #2912, [`MapView.setAccessToken()`](https://github.com/mapbox/mapbox-gl-native/blob/2b5b6fad96a38837da21300e8541a7c9a69c19a5/android/MapboxGLAndroidSDK/src/main/java/com/mapbox/mapboxsdk/views/MapView.java#L1564) should trim whitespace from either end before validating the access token. It doesn’t look like `MapView` needs to do any style URL validation independently of mbgl, since it traffics in `String`s the whole way.

/cc @ljbade

bleege:
Closes #2919

### DIFF
--- a/android/MapboxGLAndroidSDK/src/main/java/com/mapbox/mapboxsdk/views/MapView.java
+++ b/android/MapboxGLAndroidSDK/src/main/java/com/mapbox/mapboxsdk/views/MapView.java
@@ -1563,6 +1563,9 @@ public final class MapView extends FrameLayout {
     @UiThread
     public void setAccessToken(@NonNull String accessToken) {
         // validateAccessToken does the null check
+        if (!TextUtils.isEmpty(accessToken)) {
+            accessToken = accessToken.trim();
+        }
         validateAccessToken(accessToken);
         mNativeMapView.setAccessToken(accessToken);
     }


android/MapboxGLAndroidSDK/src/main/java/com/mapbox/mapboxsdk/views/MapView.java
```diff
@@ -1563,6 +1563,9 @@ private void validateAccessToken(String accessToken) {
     @UiThread
     public void setAccessToken(@NonNull String accessToken) {
         // validateAccessToken does the null check
+        if (!TextUtils.isEmpty(accessToken)) {
+            accessToken = accessToken.trim();
```
1ec5: To catch all-whitespace tokens, trim first then ask questions later.

bleege:
Rebased and merged.

android/MapboxGLAndroidSDK/src/main/java/com/mapbox/mapboxsdk/views/MapView.java
bleege: Gotta do a Null check first (which is what TextUtils.isEmpty() is doing).  If not when a null shows up the trim will crash things.

PR #2951, opened 2015-11-05
Update after final frame of transition

### LINKED ISSUE #2946: Programmatic animations fail to recalculate zoom-dependent properties
Animating a region change sometimes doesn't redraw the map's streets/text for the correct level.  When zoomed into Chicago and animating to the state of New York, the map initially renders like below:

newyork1

After slightly pinch zooming, the map will redraw itself at the correct level:

newyork2

This happens when calling `setCenterCoordinate:zoomLevel:animated:` with `YES` for animated.  We've seen it on v2.1.2 and v3.0.0-pre4.

1ec5:
The final frame is self-destructing – it destroys the frame and finish functions – so technically we’re no longer `inTransition()`. Yet there still needs to be an update after that final frame.

Also, this PR removes a redundant IsChanging notification that was being sent from the final frame of the transition. It would’ve come after the finish function’s DidChange notification.

Fixes #2946.

/cc @brunoabinader


1ec5:
Travis and Bitrise both pass; merging without waiting another time.

incanus:
Nice work!

PR #2955, opened 2015-11-06
[android] Make Annotation.getId() suitable for public use.

### LINKED ISSUE #2914: Make Marker.getId not internal
Currently `Marker.getId` is internal as we have a note as such in the JavaDoc. This was to provide flexibility to change how core GL tracks markers. Instead it is possible to use `hashMap()` to generate a unique ID.

However after a customer support request I noticed that Google does exposes a `String getId()`. From https://developers.google.com/android/reference/com/google/android/gms/maps/model/Marker.html#getId%28%29

We should document getId() as usable.

ljbade:
Fixes #2914


tobrun:
:+1:

PR #2964, opened 2015-11-06
[core] Eliminate ClassProperties

jfirebaugh:
Sixth in a series of PRs that builds towards a full-fledged [programmatic style API](https://github.com/mapbox/mapbox-gl-native/issues/837). This is a big chunk of work, but I'm trying to break it up step by step and not create a long-lived branch with a lot of changes.

This episode accomplishes my first major milestone -- eliminating the use of `ClassProperties`, the essentially untyped map of every possible style property. Now each layer type has a dedicated `*PaintProperties`, and layers with layout properties (Line and Symbol) have `*LayoutProperties`. Each of these classes contain the set of properties specific to that layer. For example, line properties:

```c++
class LineLayoutProperties {
public:
    LayoutProperty<CapType> cap = CapType::Butt;
    LayoutProperty<JoinType> join = JoinType::Miter;
    LayoutProperty<float> miterLimit = 2.0f;
    LayoutProperty<float> roundLimit = 1.0f;
};

class LinePaintProperties {
public:
    PaintProperty<float> opacity = 1.0f;
    PaintProperty<Color> color = { {{ 0, 0, 0, 1 }} };
    PaintProperty<std::array<float, 2>> translate = { {{ 0, 0 }} };
    PaintProperty<TranslateAnchorType> translateAnchor = TranslateAnchorType::Map;
    PaintProperty<float> width = 1;
    PaintProperty<float> gapWidth = 0;
    PaintProperty<float> blur = 0;
    PaintProperty<std::vector<float>, Faded<std::vector<float>>> dasharray = { {} };
    PaintProperty<std::string, Faded<std::string>> pattern = { "" };
};
```

The template classes `LayoutProperty<T>` and `PaintProperty<T>` encapsulate the logic for parsing, cascading, and calculating an _individual_ property. This is a departure from the previous approach of parsing, cascading, and calculating _sets_ of properties, but it works out much cleaner.

Because C++ has no good way to metaprogram or introspect over member data, we still need to write stuff like:

```c++
void LineLayer::cascade(const StyleCascadeParameters& parameters) {
    paint.opacity.cascade(parameters);
    paint.color.cascade(parameters);
    paint.translate.cascade(parameters);
    paint.translateAnchor.cascade(parameters);
    paint.width.cascade(parameters);
    paint.gapWidth.cascade(parameters);
    paint.blur.cascade(parameters);
    paint.dasharray.cascade(parameters);
    paint.pattern.cascade(parameters);
}
```

In the future, I may explore generating this code from the style specification. gl-js will be able to just loop over object properties.

At this point, I think style-related code is starting to get to a point where we can expose more of it in public APIs. I'm starting now to look at what threading-related work we'll need to do to make that possible.

:eyes: @brunoabinader @kkaefer


jfirebaugh:
Also cc @lucaswoj for related style/layer refactoring in gl-js.

src/mbgl/style/paint_property.hpp
```diff
@@ -0,0 +1,152 @@
...
+            break;
+        }
+
+        assert(cascaded);
+    }
+
+    bool calculate(const StyleCalculationParameters& parameters) {
+        assert(cascaded);
+        value = cascaded->calculate(parameters);
+        return cascaded->prior.operator bool();
```
brunoabinader: Shouldn't `return cascaded->prior;` do the same?

jfirebaugh: That's what I had originally but `std::unique_ptr<T>::operator bool()` is `explicit`.

brunoabinader:
This looks great @jfirebaugh! Amazing work, style code looks much more simpler now! Also nice that you followed the `hasTransitions` bool check on each style update.

PR #2967, opened 2015-11-06
fixes #2966: remove -ObjC linker flag & bring in categories via other means

### LINKED ISSUE #2966: get rid of -ObjC linker flag
This flag is used in order to ensure that categories are properly linked into the final product. Another approach here would be add class methods to each category, then call them each from a non-category to ensure they get pulled in as linked dependencies. 

https://developer.apple.com/library/mac/qa/qa1490/_index.html

> Passing the `-ObjC` option to the linker causes it to load all members of static libraries that implement any Objective-C class or category. This will pickup any category method implementations. But it can make the resulting executable larger, and may pickup unnecessary objects. For this reason it is not on by default.

This will decrease our binary size.

incanus:
This should reduce our binary size as well as obviate the need for the `-ObjC` linker flag.

### DIFF
--- a/platform/ios/NSBundle+MGLAdditions.h
+++ b/platform/ios/NSBundle+MGLAdditions.h
@@ -2,6 +2,8 @@
 
 NS_ASSUME_NONNULL_BEGIN
 
+void mgl_linkBundleCategory();
+
 @interface NSBundle (MGLAdditions)
 
 + (NSString *)mgl_resourceBundlePath;
--- a/platform/ios/NSBundle+MGLAdditions.m
+++ b/platform/ios/NSBundle+MGLAdditions.m
@@ -4,6 +4,8 @@
 
 @implementation NSBundle (MGLAdditions)
 
+void mgl_linkBundleCategory(){}
+
 + (NSString *)mgl_resourceBundlePath
 {
     NSString *resourceBundlePath = [[NSBundle bundleForClass:[MGLMapView class]] pathForResource:@"Mapbox" ofType:@"bundle"];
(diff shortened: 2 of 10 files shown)


platform/ios/NSBundle+MGLAdditions.h
```diff
@@ -2,6 +2,8 @@
 
 NS_ASSUME_NONNULL_BEGIN
 
+void linkBundleCategory();
```
1ec5: These functions should not be public. Will this still work if you declare them in MGLCategoryLoader.m (but still implement them where they are now)?

platform/ios/NSBundle+MGLAdditions.m
```diff
@@ -4,6 +4,8 @@
 
 @implementation NSBundle (MGLAdditions)
 
+void linkBundleCategory(){}
```
1ec5: These functions should have an `mgl_` prefix to avoid potential conflicts with client code that uses the same approach.

platform/ios/NSBundle+MGLAdditions.h
1ec5: Never mind, I thought this header was public.

platform/ios/NSBundle+MGLAdditions.m
incanus: Good call.

incanus: https://github.com/mapbox/mapbox-gl-native/commit/90066b9eebd93344b6164c35418e4fffdc7bad05

incanus:
This is looking good in project-internal as well as external app testing on sim and device.

PR #2959, opened 2015-11-06
Crash when changing styles too fast

kkaefer:
I could easily reproduce this on 179917ccb670d62125754a6b519aa3c89f16d25f by launching the Linux app and pressing `s` (change style) fast. Eventually it crashes. This has probably something to do with trying to render after cleaning up the style.

/cc @kkaefer 

```
#0  0x00000000004aee94 in mbgl::Painter::render (this=0x7f23600033c0, style=..., state_=..., frame_=...)
    at ../../src/mbgl/renderer/painter.cpp:158

158	                item.bucket->upload();

#1  0x0000000000462bb2 in mbgl::MapContext::renderSync (this=0x7f2369a18cc0, state=..., frame=...) at ../../src/mbgl/map/map_context.cpp:250
#2  0x000000000042871c in operator() (this=0x7f2360575438) at /usr/include/c++/4.9/functional:2439
#3  operator() (this=0x7f2360575430) at /usr/include/c++/4.9/future:1241
#4  std::_Function_handler<std::unique_ptr<std::__future_base::_Result_base, std::__future_base::_Result_base::_Deleter> (), std::__future_base::_Task_setter<std::unique_ptr<std::__future_base::_Result<bool>, std::__future_base::_Result_base::_Deleter>, bool> >::_M_invoke(std::_Any_data const&) (__functor=...) at /usr/include/c++/4.9/functional:2025
#5  0x0000000000426a52 in operator() (this=<optimized out>) at /usr/include/c++/4.9/functional:2439
#6  std::__future_base::_State_baseV2::_M_do_set(std::function<std::unique_ptr<std::__future_base::_Result_base, std::__future_base::_Result_base::_Deleter> ()>&, bool&) (this=0x286a2e0, __f=..., __set=@0x7f2369a153af: false) at /usr/include/c++/4.9/future:485
#7  0x00007f237549236b in __pthread_once_slow (once_control=0x286a34c, init_routine=0x7f2374869e60 <__once_proxy>) at pthread_once.c:114
#8  0x0000000000422342 in __gthread_once (__func=<optimized out>, __once=0x286a34c)
    at /usr/include/x86_64-linux-gnu/c++/4.9/bits/gthr-default.h:699
[...]

PR #2982, opened 2015-11-09
Revert caching render items

kkaefer:
It looks like #2908 still causes a few edge case issues which weren't fixed by #2959. I'm going to revert this change again until we have solid version of that patch.


jfirebaugh:
:+1:

There's no guarantee that there will be an event that triggers TileDataChanged after `Tile`s or `Bucket`s pointers are deleted. You could have a tile that simply becomes unused, and then end up with a `RenderItem` with dangling pointers.

jfirebaugh:
Committed as 7b8b2b1790863c7103067d40c690e273a8af5198 (messed up the branch push order).

PR #2984, opened 2015-11-09
Expose default fade duration

brunoabinader:
We should expose default fade duration to public Map API. This allows clients to e.g. disable label fade transitions by setting their duration to zero.

/cc @incanus @jfirebaugh @kkaefer @tmpsantos

### DIFF
--- a/include/mbgl/map/map.hpp
+++ b/include/mbgl/map/map.hpp
@@ -73,10 +73,13 @@ public:
     void setClasses(const std::vector<std::string>&);
     std::vector<std::string> getClasses() const;
 
-    void setDefaultTransitionDuration(const Duration& = Duration::zero());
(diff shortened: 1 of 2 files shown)


mikemorris:
This sounds good =)

ljbade:
An Android ticket should be created to add this to the API

tmpsantos:
:+1: neat, setting to 0 improves performance significantly

friedbunny:
Should this be plumbed through to iOS and Android?

incanus:
I don’t think we should expose it, unless in future we determine that @adam-mapbox’s fixes related to choosing 30/60fps aren’t good enough and we need a bigger performance boost somewhere. But in general, if we’re not seeing problems, I think we should stick with the more elegant label fade appearance.

PR #2986, opened 2015-11-09
Fetch tags on install, not ipackage

### LINKED ISSUE #2320: make ipackage requires network connection
For #1880, I added a `git fetch --tags` to the iOS package script so that AppleDoc can version the docs without erroring out on Travis. However, fetching tags requires an Internet connection. This step should be moved to something like `make iproj`, which at least sometimes requires a connection.

/cc @incanus

### LINKED ISSUE #2871: make ipackage error message out of date
Running `make ipackage` failed for me at the "Creating API Docs.." step with:

```
Unable to find appledoc. See https://github.com/mapbox/mapbox-gl-native/blob/master/INSTALL.md
```

But there is no mention of `appledoc` in that file, but rather its in https://github.com/mapbox/mapbox-gl-native/blob/master/docs/BUILD_IOS_OSX.md

1ec5:
This change removes a git command that requires a network connection from the ipackage target. It was used to ensure that CI has a tag to version the appledoc documentation with, so the same command has been added to install.sh, where a network connection is otherwise expected.

Along for the ride, I’ve also updated some references to the old build documentation.

Fixes #2320; fixes #2871.

/cc @adam-mapbox @incanus

### DIFF
--- a/ios/docs/install_docs.sh
+++ b/ios/docs/install_docs.sh
@@ -1,7 +1,7 @@
 #!/usr/bin/env bash
 
 if [ -z `which appledoc` ]; then
-    echo "Unable to find appledoc. See https://github.com/mapbox/mapbox-gl-native/blob/master/INSTALL.md"
+    echo "Unable to find appledoc. See https://github.com/mapbox/mapbox-gl-native/blob/master/docs/BUILD_IOS_OSX.md"
     exit 1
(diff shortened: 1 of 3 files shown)


incanus:
Haha! The solution surfaces. Cross this one off your bucket list @1ec5.

PR #2987, opened 2015-11-09
[ios] Fix for #2259 - perspective tilt and cameraForLatLngs

incanus:
@ansis mentioned he was going to check this out. 

/cc @adam-mapbox

### DIFF
--- a/src/mbgl/map/transform_state.hpp
+++ b/src/mbgl/map/transform_state.hpp
@@ -101,7 +101,7 @@ private:
     double x = 0, y = 0;
     double angle = 0;
     double scale = 1;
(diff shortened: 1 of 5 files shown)


src/mbgl/map/transform_state.hpp
```diff
@@ -101,7 +101,7 @@ class TransformState {
     double x = 0, y = 0;
     double angle = 0;
     double scale = 1;
-    double altitude = 1.5;
+    double FOV = 36.87;
```
1ec5: As of #3362, iOS-specific calculations assume a vertical FOV of 30°, which is consistent with MapKit.

PR #2988, opened 2015-11-09
Improve transitions in iOS 9 multitasking

### LINKED ISSUE #2158: Opening control center during pan animation causes map to disappear
I've only tested this on a physical device with iOS 7.1 using the sample app.

To reproduce:
-Zoom in far enough that you have plenty of panning room (max or close to it)
-Start panning very fast from bottom to top
-During one of the rapid upward swipes, start your finger below the screen so it brings up the control center
-The control center opens and the map is gone...the Mapbox logo and attribution are still there

It seems to only happen if the control center is opened while the panning animation is still running.

1ec5:
Leave GLKView rendering on when app is inactive, to get nice smooth transitions between sizes when using multitasking on iPad (iOS 9).

Fixes #2158. Incorporates #2309.

/cc @bwhtmn @incanus

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -33,6 +33,7 @@ Known issues:
 - Fixed an issue that prevented zoom level–dependent style properties from updating after zooming programmatically with animation. ([#2951](https://github.com/mapbox/mapbox-gl-native/pull/2951))
 - Fixed a rendering issue with styles that use the `background-pattern` property. ([#2531](https://github.com/mapbox/mapbox-gl-native/pull/2531))
 - Fixed a crash when reusing a single `MGLMapView` across multiple `UIViewController`s. ([#2969](https://github.com/mapbox/mapbox-gl-native/pull/2969))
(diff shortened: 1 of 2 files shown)


incanus:
According to [this docs](https://developer.apple.com/library/ios/documentation/UIKit/Reference/UIApplication_Class/index.html#//apple_ref/c/data/UIApplicationWillResignActiveNotification), wouldn't this cause the GL to keep rendering when the device is locked?

1ec5:
> According to [this docs](https://developer.apple.com/library/ios/documentation/UIKit/Reference/UIApplication_Class/index.html#//apple_ref/c/data/UIApplicationWillResignActiveNotification), wouldn't this cause the GL to keep rendering when the device is locked? 

On iOS 9, when I lock the phone, the application receives `UIApplicationWillResignActiveNotification` followed by `UIApplicationDidEnterBackgroundNotification`. When I unlock, it receives `UIApplicationWillEnterForegroundNotification` followed by `UIApplicationDidBecomeActiveNotification`.

incanus:
Ah, perfecto. LGTM then.

PR #2994, opened 2015-11-10
Map API inconsistences

brunoabinader:
We currently have some API inconsistencies in our Map API:

- Pixel points are sometimes represented separately (e.g. `double x, double y`) or inside `vec2<double>` structs - see `Map::scaleBy` and `Map::setLatLng` as examples;
- Inconsistent usage of `const` for return types and pass-by-value, pass-by-const-ref, pass-by-ref in function parameters;
- We should be checking more for `std::nan()` in `Transform` functions;

For the pixel handling, I propose a `PrecisionPoint` struct with two double _x_ and _y_ members in the same fashion as the other structures in `include/mbgl/util/geo.hpp`.

These changes shouldn't affect current behavior, but might break API compatibility with Android and iOS platforms.

/cc @mapbox/gl @mapbox/mobile

### DIFF
--- a/include/mbgl/map/map.hpp
+++ b/include/mbgl/map/map.hpp
@@ -135,8 +135,8 @@ public:
     uint16_t getHeight() const;
 
     // Projection
-    void getWorldBoundsMeters(ProjectedMeters& sw, ProjectedMeters& ne) const;
-    void getWorldBoundsLatLng(LatLng& sw, LatLng& ne) const;
+    MetersBounds getWorldBoundsMeters() const;
+    LatLngBounds getWorldBoundsLatLng() const;
 
     double getMetersPerPixelAtLatitude(double lat, double zoom) const;
     ProjectedMeters projectedMetersForLatLng(const LatLng&) const;
--- a/include/mbgl/util/geo.hpp
+++ b/include/mbgl/util/geo.hpp
@@ -54,7 +54,7 @@ struct LatLngBounds {
     LatLng sw = {90, 180};
     LatLng ne = {-90, -180};
 
-    inline LatLngBounds(LatLng sw_ = {90, 180}, LatLng ne_ = {-90, -180})
+    inline LatLngBounds(const LatLng& sw_ = {90, 180}, const LatLng& ne_ = {-90, -180})
         : sw(sw_), ne(ne_) {}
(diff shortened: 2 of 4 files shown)


src/mbgl/map/transform.cpp
```diff
@@ -55,67 +60,69 @@ bool Transform::resize(const std::array<uint16_t, 2> size) {
 
 #pragma mark - Position
 
-void Transform::jumpTo(const CameraOptions options) {
-    CameraOptions jumpOptions = options;
-    jumpOptions.duration.reset();
-    easeTo(jumpOptions);
+void Transform::jumpTo(CameraOptions& options) {
+    options.duration.reset();
```
1ec5: This line mutates the CameraOptions for local purposes. There are several instances of this in Map and Transform, which is why the CameraOptions was being passed in by value.

jfirebaugh: This method should receive by const reference and copy before resetting.

1ec5:
This was committed in:

e5c460f2db3f7506f4fc38b16747f0ace9151017
e2052ea3e46272f5b0eb26643c0d7e70bad46557
c78e7696352e73f94500fd94a5eaaafffb971be6
99ac8d8720fcd838141b291e4c85466aefbb88f6
5a24ac605b23ae1d28ac052eb6eeae559db3bb79
994ee519e810705e6b703f73b54ec7395cac9725
12e0e5eb0434e3243e0dd6fedf8be7e6c5d1cbca
cd0266f25cc96358e665b6f39e358a263f6d4ab8
aa8857643d404a5512b1bc3cec380e56f88b6f75
e31c96d885f07fc70dbbc10322c1cdc9d678e992

ljbade:
@brunoabinader @1ec5 for future reference if a PR does not get "merged" status properly, you only need to force push to the branch with a appropriately rebased tree. GH will then fixup closed PR to merged status. This works even if you merge to master and forget to update the branch first after rebase.

ljbade:
@jfirebaugh you might also find ^ tip handy if you spot a broken PR merge

1ec5:
@ljbade, actually, I think that's what happened here, which is why the PR lists zero commits.

PR #3006, opened 2015-11-11
Fix static analyzer warnings

1ec5:
This PR fixes the following static analyzer warnings introduced by 5a24ac6 for #2994 and 4d5c633 for #2727:

```
/path/to/mapbox-gl-native/src/mbgl/storage/default_file_source.cpp
/path/to/mapbox-gl-native/src/mbgl/storage/default_file_source.cpp:328:29: The result of the '<<' expression is undefined
/path/to/mapbox-gl-native/src/mbgl/storage/default_file_source.cpp:328:29: The result of the '<<' expression is undefined
/path/to/mapbox-gl-native/src/mbgl/storage/default_file_source.cpp:333:25: The result of the '<<' expression is undefined
/path/to/mapbox-gl-native/src/mbgl/storage/default_file_source.cpp:333:25: The result of the '<<' expression is undefined
/path/to/mapbox-gl-native/src/mbgl/map/transform.cpp
/path/to/mapbox-gl-native/src/mbgl/map/transform.cpp:328:9: Value stored to 'center_x' is never read
/path/to/mapbox-gl-native/src/mbgl/map/transform.cpp:328:9: Value stored to 'center_x' is never read
/path/to/mapbox-gl-native/src/mbgl/map/transform.cpp:329:9: Value stored to 'center_y' is never read
/path/to/mapbox-gl-native/src/mbgl/map/transform.cpp:329:9: Value stored to 'center_y' is never read
```

More details in the commit messages.

/cc @brunoabinader @kkaefer

### DIFF
--- a/src/mbgl/map/transform.cpp
+++ b/src/mbgl/map/transform.cpp
@@ -310,29 +310,24 @@ void Transform::rotateBy(const PrecisionPoint& first, const PrecisionPoint& seco
         return;
     }
 
-    double center_x = static_cast<double>(state.width) / 2.0, center_y = static_cast<double>(state.height) / 2.0;
+    PrecisionPoint center(state.width, state.height);
+    center /= 2;
 
-    const double begin_center_x = first.x - center_x;
-    const double begin_center_y = first.y - center_y;
-
-    const double beginning_center_dist =
-        std::sqrt(begin_center_x * begin_center_x + begin_center_y * begin_center_y);
+    const PrecisionPoint offset = first - center;
+    const double distance = std::sqrt(std::pow(2, offset.x) + std::pow(2, offset.y));
 
     // If the first click was too close to the center, move the center of rotation by 200 pixels
     // in the direction of the click.
-    if (beginning_center_dist < 200) {
-        const double offset_x = -200, offset_y = 0;
-        const double rotate_angle = std::atan2(begin_center_y, begin_center_x);
-        const double rotate_angle_sin = std::sin(rotate_angle);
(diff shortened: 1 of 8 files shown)


src/mbgl/map/transform.cpp
```diff
@@ -310,25 +310,28 @@ void Transform::rotateBy(const PrecisionPoint& first, const PrecisionPoint& seco
...
-        const double rotate_angle = std::atan2(first_y, first_x);
+        const double rotate_angle = std::atan2(begin_center_y, begin_center_x);
         const double rotate_angle_sin = std::sin(rotate_angle);
         const double rotate_angle_cos = std::cos(rotate_angle);
         center_x = first.x + rotate_angle_cos * offset_x - rotate_angle_sin * offset_y;
         center_y = first.y + rotate_angle_sin * offset_x + rotate_angle_cos * offset_y;
     }
 
+    const double first_x = first.x - center_x, first_y = first.y - center_y;
+    const double second_x = second.x - center_x, second_y = second.y - center_y;
```
1ec5: I simply reverted this method to before the refactoring and applied the minimal number of changes required by the new parameters. The refactoring had turned the `if` statement above here into a no-op, and I’m not sure `beginning_center_dist` was still correct.

brunoabinader:
@1ec5 I spent some time checking on this and took liberty to add 2 more patches here - one that makes `PrecisionPoint` inherit from `vec2<>` so we can take advantage of its operator impls, and other that cleans up `Transform::rotateBy` preserving the same behavior as before my changes.

PR #3008, opened 2015-11-12
Floating-point logical sprite image dimensions

### LINKED ISSUE #2198: MGLAnnotationImage with scale greater than 1.0 causes crash.

1ec5:
On HiDPI screens, the logical dimensions of a sprite image can be non-integral.

Fixes #2198.

/cc @incanus @friedbunny

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -33,6 +33,7 @@ Known issues:
 - Fixed an issue that prevented zoom level–dependent style properties from updating after zooming programmatically with animation. ([#2951](https://github.com/mapbox/mapbox-gl-native/pull/2951))
 - Fixed a rendering issue with styles that use the `background-pattern` property. ([#2531](https://github.com/mapbox/mapbox-gl-native/pull/2531))
 - Fixed a crash when reusing a single `MGLMapView` across multiple `UIViewController`s. ([#2969](https://github.com/mapbox/mapbox-gl-native/pull/2969))
+- Fixed a crash on Retina displays when displaying an annotation whose image has non-integral dimensions. ([#3008](https://github.com/mapbox/mapbox-gl-native/pull/3008))
 - Eliminated flickering when opening and closing an overlay, such as an alert or action sheet. ([#2309](https://github.com/mapbox/mapbox-gl-native/pull/2309))
 - Labels can now line wrap on hyphens and other punctuation. ([#2598](https://github.com/mapbox/mapbox-gl-native/pull/2598))
 - A new delegate callback was added for observing taps to annotation callout views. ([#2596](https://github.com/mapbox/mapbox-gl-native/pull/2596))
--- a/include/mbgl/annotation/sprite_image.hpp
+++ b/include/mbgl/annotation/sprite_image.hpp
@@ -13,11 +13,11 @@ namespace mbgl {
 class SpriteImage : private util::noncopyable {
 public:
     SpriteImage(
-        uint16_t width, uint16_t height, float pixelRatio, std::string&& data, bool sdf = false);
+        float width, float height, float pixelRatio, std::string&& data, bool sdf = false);
 
     // Logical dimensions of the sprite image.
-    const uint16_t width;
-    const uint16_t height;
+    const float width;
+    const float height;
 
     // Pixel ratio of the sprite image.
     const float pixelRatio;
--- a/platform/ios/MGLMapView.mm
+++ b/platform/ios/MGLMapView.mm
@@ -2289,9 +2289,9 @@ CLLocationCoordinate2D MGLLocationCoordinate2DFromLatLng(mbgl::LatLng latLng)
 
     // add sprite
     auto cSpriteImage = std::make_shared<mbgl::SpriteImage>(
-        uint16_t(annotationImage.image.size.width),
-        uint16_t(annotationImage.image.size.height),
-        float(annotationImage.image.scale),
+        annotationImage.image.size.width,
+        annotationImage.image.size.height,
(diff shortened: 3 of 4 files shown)


friedbunny:
Awesome that you tracked this down, fix works. Adding some decimals in [iosapp's marker image code](https://github.com/mapbox/mapbox-gl-native/blob/7621a411582b54ccba7f90d5c2f53a5960ec989c/ios/app/MBXViewController.mm#L371) is enough to confirm:

```diff
diff --git a/ios/app/MBXViewController.mm b/ios/app/MBXViewController.mm
index 6380406..61d39b8 100644
--- a/ios/app/MBXViewController.mm
+++ b/ios/app/MBXViewController.mm
@@ -368,7 +368,7 @@ static UIColor *const kTintColor = [UIColor colorWithRed:0.120 green:0.550 blue:
 
     if ( ! image)
     {
-        CGRect rect = CGRectMake(0, 0, 20, 15);
+        CGRect rect = CGRectMake(0, 0, 20.5, 15.5);
 
         UIGraphicsBeginImageContextWithOptions(rect.size, NO, [[UIScreen mainScreen] scale]);
 ```

jfirebaugh:
I'm suspicious of the float ⇢ integer conversions that will be happening after this change [here](https://github.com/mapbox/mapbox-gl-native/blob/b918b5167273ee46f7eaad8ed2384f3dd6cf1bfa/src/mbgl/geometry/sprite_atlas.cpp#L68) and [here](https://github.com/mapbox/mapbox-gl-native/blob/b918b5167273ee46f7eaad8ed2384f3dd6cf1bfa/src/mbgl/geometry/sprite_atlas.cpp#L95-L96). We should probably review our handling of physical/logical dimensions at a higher level. Always using integer pixel dimensions plus float pixel ratio makes the most sense to me, but I haven't thought about it a lot.

1ec5:
The upshot is that without floating point logical dimensions (expressed in "points" in iOS), you can't have an odd width or height after multiplying by the pixel ratio.

1ec5:
The padding in `allocateImage()` would seem to inoculate this particular change from any ill effects. `SpriteAtlas::getPosition()` is only used to render backgrounds, fills, and lines, so it doesn’t come into play when assigning images to markers. I realize that introducing the concept of non-integral dimensions to just `SpriteImage` is problematic in the long run. However, this seems like a workable stopgap fix for #2198 while we consider more disruptive changes such as eliminating the concept of pixel density outside the renderer.

bleege:
Also related to Android via #3031

jfirebaugh:
I'm working on a different fix along the lines of https://github.com/mapbox/mapbox-gl-native/pull/3008#issuecomment-156178454, and build on the new `PremultipliedImage` class.

bleege:
@jfirebaugh Sounds good!  Do you have a ticket or branch that we can link to so that we make sure this gets addressed / tickets closed in the iOS and Android bindings when it's ready?

jfirebaugh:
@bleege: #3164.

PR #3009, opened 2015-11-12
Set refresh interval to 60fps

### LINKED ISSUE #3005: iPhone 6 animation performance cut in half
To state the obvious, #2922 cuts the maximum framerate down from 60fps to 30fps. Prior to #2922, MGLMapView would pan smoothly at 60fps (as confirmed by Xcode) on my iPhone 6 running iOS 9.1. Now it maxes out at 30fps, causing every animation to be choppy to an extent that’s difficult to ignore. (An `MGLTargetFrameInterval` of 1 unsurprisingly feels great.) Perhaps 30fps is beneficial for the iPod touch @adam-mapbox was testing on, but that should not come at the cost of performance on more current or more powerful hardware, where there is a greater expectation of smooth animations.

/cc @adam-mapbox @incanus @friedbunny @jfirebaugh

1ec5:
Set the refresh interval to 60 frames per second, which performs well on a 64 GB iPhone 6. We can blacklist specific hardware models that fail to keep up as we come across them in testing, but I think this is a good starting point.

Fixes #3005.

/cc @adam-mapbox @incanus @friedbunny

### DIFF
--- a/platform/ios/MGLMapView.mm
+++ b/platform/ios/MGLMapView.mm
@@ -53,7 +53,7 @@ const CGFloat MGLMinimumPitch = 0;
 const CGFloat MGLMaximumPitch = 60;
 const CLLocationDegrees MGLAngularFieldOfView = M_PI / 6.;
 const std::string spritePrefix = "com.mapbox.sprites.";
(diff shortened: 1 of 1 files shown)


friedbunny:
:heart_decoration:

PR #3011, opened 2015-11-12
2989 move token file

### LINKED ISSUE #2989: Move Develpment Access Token ( token.txt ) Back To TestApp
As discussed in #2974 we should refactor `ApiAccess` so that we can move the Developer Access Token (aka `token.txt`) back to the TestApp.  This is because runtime configuration parameters should not be located in the SDK proper.  Instead they should always be located in external implementing apps.

This ticket will have similar goals as #1929 in that in order for it to be considered "done" it needs to support the following configurations:

* [x] Activity via XML Layout
* [x] Activity via Programmatically set (ie, MapView.setAccessToken())
* [x] App via `AndroidManifest.xml`
* [x] Fragment via XML
* [x] Fragment via Programmatically set

Access Tokens should be storable in the following manners:

* App's `R.string.*`
* App's Code
* App's `token.txt` (with corresponding .gitignore so it's not commit-able)

/cc @tobrun @ljbade @zugaldia

tobrun:
closes #2989


tobrun:
@bleege

bleege:
@tobrun Looks good!  Since `config.xml` is going to be developer specific, do think it'd be helpful if we renamed it to something like `developer-config.xml`?  Just trying to be overly clear that this is a day to day Developer thing.

tobrun:
@bleege good idea, I will rename it

tobrun:
fixed + rebased, now waiting on green build to merge

tobrun:
Green light, going to merge

PR #3012, opened 2015-11-12
[core] Throttle state updates

tmpsantos:
Throttle how often we update the tile state. Previously we were updating as fast as we could, but if we do it more often than the frame rate, it is just a waste of resources because you only see the last state before rendering a frame.

Note that we still send invalidate requests as much as we can. In my tests throttling the ::invalidate() rate to the same frequency as the ::update() didn't yield great results mostly because the clock tick is not so precise. Maybe it could be tunned so we don't spam the client with invalidate request and don't wake the main loop too much.

Finally, we should probably expose the update throttle as a public API because:

1. The client might want to reduce the update frequency due to performance reasons. I.e. wants to paint at 60FPS at the cost of not having a so up-to-date map in the frame.
2. The client might have a frame rate set to something lower than 60FPS so updating at 16ms is overkill.

### DIFF
--- a/platform/default/async_task.cpp
+++ b/platform/default/async_task.cpp
@@ -1,6 +1,7 @@
 #include <mbgl/util/async_task.hpp>
 
 #include <mbgl/util/run_loop.hpp>
+#include <mbgl/util/timer.hpp>
 
 #include <atomic>
 #include <functional>
@@ -38,15 +39,33 @@ public:
     }
 
     void maySend() {
-        // uv_async_send will do the call coalescing for us.
-        if (uv_async_send(async) != 0) {
-            throw std::runtime_error("Failed to async send.");
+        if (!queued.test_and_set()) {
+            if (uv_async_send(async) != 0) {
+                throw std::runtime_error("Failed to async send.");
+            }
         }
+    }
+
+    void runTask() {
+        queued.clear();
+        task();
+    }
+
+    void received() {
+        if (throttle == Duration::zero()) {
+            runTask();
+        } else {
+            timer.start(throttle, Duration::zero(), [this] { runTask(); });
+        }
+    }
+
+    void setThrottle(Duration timeout) {
+        throttle = timeout;
     }
 
 private:
     static void asyncCallback(UV_ASYNC_PARAMS(handle)) {
-        reinterpret_cast<Impl*>(handle->data)->task();
+        reinterpret_cast<Impl*>(handle->data)->received();
     }
 
     uv_handle_t* handle() {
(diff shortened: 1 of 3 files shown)


tmpsantos:
Would be great to get some feedback on how this PR performs on Android and iOS.

/cc @bleege @ljbade @incanus @adam-mapbox

incanus:
Sorry for the delay here @tmpsantos — iOS release-focused. 

This seems clean enough and I will give it a spin. If I'm understanding it properly, this would obsolete the need for `CADisplayLink` at the iOS level as `renderSync` would get throttled properly in core. Also, if we deemed it necessary, we could plump through the 60fps value out to `map.hpp` for fine tuning at the platform level — for example we've talked about maybe limiting certain older iOS devices to 30fps. 

/cc @adam-mapbox

adam-mapbox:
@incanus I don't believe this would eliminate the need for `CADisplayLink`.  There are many ways that a `renderSync` can get triggered.  But luckily we have a great way of settling this: we should just use ground truth.  :)

incanus:
My point is that `MapContext::renderSync` still goes through this new routine, so internally the core would decide when to act on it. So platform could fire as often as it wished, just as we were doing before the `CADisplayLink`.

tmpsantos:
> If I'm understanding it properly, this would obsolete the need for CADisplayLink at the iOS level as renderSync would get throttled properly in core.

We could throttle `renderSync` too, it is worth experimenting (and then my understanding is you can get rid of the `CADisplayLink` if it works). You can do that by adding a:

`+    asyncInvalidate.setThrottle(1000 / 60);`

What this patch is doing is throttling the `asyncUpdate` but we keep issuing `asyncInvalidate` as fast as we can (which causes the view to call `map::renderSync`). It assumes that the underlying platform that is receiving the signals for repainting the map will coalesce this signals and only call `map::renderSync` when it is appropriated at the frequency it can handle (but obviously everything over 60Hz is a waste). We try not no be intrusive on the application level.

So what this patch as it is really does is save CPU and decrease latency when rendering because right now we render on the `MapContext` thread and if the thread is busy doing `asyncUpdate`s it will decrease how fast we render things because we can only start rendering after an ongoing `update()` finishes.

incanus:
Ahhhh, awesome. So testing here for mobile would mostly involve just making sure perf looks good and no regressions are introduced, but nothing super concrete other than hopeful reduced CPU usage.

jfirebaugh:
Closing; stale. I'd like to see an approach where we replace the "update" async task with a view invalidation whenever required, and perform the necessary updates inline in the render pathway. Throttling then can be the job of the view implementation. This is the approach that mapbox-gl-js uses.

Issue #2211, opened 2015-08-31
Allow more annotation callout view customization


friedbunny:
Our annotation callout view is [currently hardcoded to a plain `SMCalloutView`](https://github.com/mapbox/mapbox-gl-native/blob/c6e3ed8c6629d0a42d984a4ced57eff558eac0ea/platform/ios/MGLMapView.mm#L2215-L2225) and only offers custom title/subtitle text and [customization of the left and right accessory views](https://github.com/mapbox/mapbox-gl-native/blob/c6e3ed8c6629d0a42d984a4ced57eff558eac0ea/include/mbgl/ios/MGLMapView.h#L408-L439).

We should expose our `SMCalloutView` object and/or provide a delegate method for setting a custom callout view on a per annotation basis.

/cc @1ec5 @incanus

wrutkowski:
Possible duplicate of https://github.com/mapbox/mapbox-gl-native/issues/1784

friedbunny:
> Possible duplicate of #1784

#1784 is a similar use case, but refers to adding arbitrary custom native views as annotations on top of the GL map view. This ticket refers to our existing callout view, which needs to be better exposed for customization.

Name-less:
Someone know if callout view customization will be avaible ?

robipresotto:
+1

ljbade:
@friedbunny Adding parity ticket as we already have this feature in Android release.

RomainQuidet:
Hi, we need this feature at Mappy, I'm working on adding an API to set a custom callout view

RomainQuidet:
new pull request done: #3322

YBthebest:
Hi,

is this possible with the new pull request? I'd like to display a UIView instead of the default callout, is this achievable?

Thanks!

byroncoetsee:
That's what I'm busy doing at the moment @YBthebest . You can do it by returning false for the `annotationCanShowCallout` delegate method and handling the annotation manually. For example show your custom UIView from that method with the information contained in the annotation (which already comes in as an argument in the `annotationCanShowCallout` method)

RomainQuidet:
@byroncoetsee it's already done by this pull request #3322
@YBthebest Yes just pull the branch dedicated to #3322 and go to the demo app, there is an example.

YBthebest:
@RomainQuidet sorry I'm fairly new to Github and its process, how can I pull your branch? I can't seem to find an URL

[...]

PR #3013, opened 2015-11-12
iOS Customize Callout Content

mappy-mobile:
Add the ability to customize callout views content via the mapView:calloutViewForAnnotation: callback of MGLMapViewDelegate.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,6 +23,7 @@ Known issues:
 
 ## iOS master
 
+- Add the ability to customize callout views content via the mapView:calloutViewForAnnotation: callback of MGLMapViewDelegate.
 - If you install this SDK via CocoaPods, CocoaPods version 0.38.0 or above is required. ([#2132](https://github.com/mapbox/mapbox-gl-native/pull/2132))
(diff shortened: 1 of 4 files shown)


friedbunny:
This would fix #2211.

mappy-mobile:
I didn't mention issue #2211 because it seems to me that was about totally replacing the callout view.
Maybe both systems can be implemented : 
 - either the user provides a full view ( as requested in #2211)
 - either the user provides only the content as implemented by this pull request

My callback could be renamed to 'calloutContentViewForAnnotation' and we could add a 'calloutViewForAnnotation' for the first use case ?

friedbunny:
@mappy-mobile Ah yes, you're totally right —`calloutViewForAnnotation` threw me, at first glance.

I think our intention should still be to allow complete replacement of the callout view (#2211), but `calloutContentViewForAnnotation`/`centerCalloutContentView` (this PR) may also be useful.

/cc @incanus @1ec5

1ec5:
Thank you for the pull request! We’re overhauling the callout API in #4948 to remove the SMCalloutView dependency. I’m not sure if we want to allow a halfway-custom callout view in the new API, but let’s see whether the implementation lends itself to something like that.

PR #3026, opened 2015-11-13
Constrain map on both axis

brunoabinader:
Currently we constrain the map boundaries only in the vertical axis. We should have the option to constrain also in horizontal axis, which would disable the possibility to infinitely pan the map horizontally, but also fixes an issue where annotations start disappearing once we reach the map horizontal boundaries.

/cc @mapbox/gl @mapbox/mobile

### DIFF
--- a/include/mbgl/map/map.hpp
+++ b/include/mbgl/map/map.hpp
@@ -45,7 +45,8 @@ class Map : private util::noncopyable {
 public:
     explicit Map(View&, FileSource&,
                  MapMode mapMode = MapMode::Continuous,
-                 GLContextMode contextMode = GLContextMode::Unique);
+                 GLContextMode contextMode = GLContextMode::Unique,
+                 ConstrainMode constrainMode = ConstrainMode::HeightOnly);
     ~Map();
 
     // Pauses the render thread. The render thread will stop running but will not be terminated and will not lose state until resumed.
--- a/include/mbgl/map/mode.hpp
+++ b/include/mbgl/map/mode.hpp
@@ -19,6 +19,13 @@ enum class GLContextMode : uint8_t {
     Shared,
 };
 
+// We can choose to constrain the map both horizontally or vertically, or only
+// vertically e.g. while panning.
+enum class ConstrainMode : uint8_t {
+    HeightOnly,
+    WidthAndHeight,
+};
+
 } // namespace mbgl
 
 #endif // MBGL_MAP_MODE
--- a/include/mbgl/util/geo.hpp
+++ b/include/mbgl/util/geo.hpp
@@ -41,11 +41,18 @@ struct ProjectedMeters {
 };
 
 struct LatLngBounds {
-    LatLng sw = {90, 180};
-    LatLng ne = {-90, -180};
+    LatLng sw = {-90, -180};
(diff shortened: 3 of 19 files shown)


mb12:
@brunoabinader Won't this give poor experience to customers on lager screens at zoom level 0 because most of the area would be covered by background color instead of  repeated continents  in horizontal direction as was the case earlier? Infinite horizontal scrolling seems like a better experience for end users. Google maps, openstreetmaps, here.com websites allow infinite horizontal scrolling on their web apps. Infinite horizontal scrolling is a desirable feature and not a bug.

brunoabinader:
> Won't this give poor experience to customers on lager screens at zoom level 0 because most of the area would be covered by background color instead of repeated continents in horizontal direction as was the case earlier

No, because when constraining both axis we also constrain the minimum scale to make sure the map is entirely filled either horizontally or vertically, except when e.g. rotating the map at minimum scale (example below):

Minimum scale view:
screen shot 2015-11-13 at 19 54 30
Minimum scale (rotated) view:
screen shot 2015-11-13 at 19 55 37

> Infinite horizontal scrolling seems like a better experience for end users. Google maps, openstreetmaps, here.com websites allow infinite horizontal scrolling on their web apps. Infinite horizontal scrolling is a desirable feature and not a bug.

Not every endpoint wants the same experience. That is why I've added an enum `ConstrainMode` to be passed in `Map` ctor so you can choose which behavior you can expect.

mb12:
@brunoabinader Thanks for the explanation regarding constraining minimum scale as well. This looks greate.

1ec5:
Please test this in a very wide and narrow screen, such as you'd find on a phone in landscape orientation. It should still be possible to zoom out to see the whole world, even if that means repeating up to about half of the world.

mikemorris:
Should we add some render tests in https://github.com/mapbox/mapbox-gl-test-suite to verify this behaves as expected?

brunoabinader:
> It should still be possible to zoom out to see the whole world, even if that means repeating up to about half of the world.

@1ec5 this is what happens when you select `ConstrainMode::ConstrainHeight`.

brunoabinader:
To clarify: If your _port_ requires vertical constrain only, pass `ConstrainMode::ConstrainHeight` to your `mbgl::Map` ctor. If you want to constrain on both axis, use `ConstrainMode::ConstrainWidthAndHeight` instead (default). We can make `ConstrainMode::ConstrainHeight` the default behaviour if that helps.

brunoabinader:
I've renamed the enum values to `ConstrainMode::WidthAndHeight` and `ConstrainMode::HeightOnly`.

`Constrain::HeightOnly` is now the default behaviour, maintaining compatibility with the previous implementation.

brunoabinader:
@jfirebaugh I removed the extra `TransformState` param to `MapContext`ctor, and added a default value for `ConstrainMode` in `TransformState` ctor.

brunoabinader:
@mikemorris I've added tests in `test/miscellaneous/transform.cpp`.

I've opened a ticket for render tests in https://github.com/mapbox/mapbox-gl-test-suite/issues/59 :+1:

tmpsantos:
:+1:

PR #3030, opened 2015-11-13
Add creds for bitrise

miccolis:
@bleege should be what you need. Let me know if it makes sense to you :)


bleege:
Looks great @miccolis!  I just rebased and merged it.  We can figure out how to get this into AWS CLI on Monday.  Thanks again!

PR #2334, opened 2015-09-15
use std::unique_ptr for NodeMap resources

mikemorris:
Properly clean up all resources in `NodeMap::release` to allow the `NodeMap`
object to be garbage collected.

Fixes a memory leak that likely started in https://github.com/mapbox/node-mapbox-gl-native/pull/143 when the `NodeFileSource` object was moved inside the `NodeMap` instead of being an external reference.

/cc @lucaswoj @jfirebaugh @miccolis

PR #3033, opened 2015-11-14
[core] Clean up FileSource

### LINKED ISSUE #2827: FileSource::request should return an owned pointer
Ownership of the resulting `Request` should belong to the caller. Releasing the `Request` should cancel it.

jfirebaugh:
:eyes: @tmpsantos @kkaefer @mikemorris 

Clean up FileSource by taking advantage of the safe callback and cancellation semantics provided by `RunLoop::invokeWithCallback`.

Fixes #2827, eliminates the complex logic around `Resource` destruction, eliminates `ResourceHolder`, moves us toward #2909.

Also allows simplification of the node implementation of `FileSource`: we can use `RunLoop` to handle dispatch back to the node event loop. I removed JS-level cancellation via the optional "cancel" property; it was untested and unused. (Note that `RunLoop::invokeWithCallback` handles cancellation at the C++ level; i.e. not calling the callback if the request has already been cancelled.)


include/mbgl/util/run_loop.hpp
```diff
@@ -64,14 +65,23 @@ class RunLoop : private util::noncopyable {
...
+        auto flag = std::make_shared<std::atomic<bool>>();
         *flag = false;
 
-        auto after = RunLoop::current.get()->bind([flag, callback] (auto&&... results) {
+        // Create a lambda L1 that invokes another lambda L2 on the current RunLoop R, that calls
+        // the callback C. Both lambdas check the flag before proceeding. L1 needs to check the flag
+        // because if the request was cancelled, then R might have been destroyed. L2 needs to check
+        // the flag because the request may have been cancelled after L2 was invoked but before it
+        // began executing.
+        auto after = [flag, current = RunLoop::current.get(), callback1 = std::move(callback)] (auto&&... results1) {
```
tmpsantos: Good catch.

include/mbgl/util/run_loop.hpp
```diff
@@ -64,14 +65,23 @@ class RunLoop : private util::noncopyable {
     template <class Fn, class Cb, class... Args>
     std::unique_ptr<WorkRequest>
     invokeWithCallback(Fn&& fn, Cb&& callback, Args&&... args) {
-        auto flag = std::make_shared<bool>();
+        auto flag = std::make_shared<std::atomic<bool>>();
```
tmpsantos: If it is just a flag, I would use `std::atomic_flag` which is guaranteed to be lock-free.

include/mbgl/storage/file_source.hpp
```diff
@@ -1,33 +1,33 @@
...
 
 #include <functional>
-
-typedef struct uv_loop_s uv_loop_t;
+#include <memory>
 
 namespace mbgl {
 
-class Request;
+class FileRequest : private util::noncopyable {
```
tmpsantos: This class should be `final` and implemented at `file_source.cpp` to enforce correctness on everyone implementing the `FileSource` interface. I would also name it `FileSourceWork` to avoid the confusion like `DefaultFileRequest` and `DefaultFileRequestImpl`.

tmpsantos:
Kudos @jfirebaugh, I really liked this.

include/mbgl/util/run_loop.hpp
jfirebaugh: I would but I don't think it's possible to portably use `std::shared_ptr<std::atomic_flag>` due to the initialization requirements of `ATOMIC_FLAG_INIT`.

include/mbgl/storage/file_source.hpp
jfirebaugh: I considered an approach like this but ultimately decided to leave it up to the `FileSource` implementation what the derived destructor does. The default implementation cancels the request, but the node implementation does not.

include/mbgl/util/run_loop.hpp
tmpsantos: Ugh, indeed.

mikemorris:
Aside from the following minor naming issue this is working great for me, and fixes the segfaults I was seeing previously in https://github.com/mapbox/mapbox-gl-native/pull/2334.
```
../../src/mbgl/storage/default_file_source.cpp: In member function ‘void mbgl::DefaultFileSource::cancel(const mbgl::Resource&, mbgl::FileRequest*)’:
../../src/mbgl/storage/default_file_source.cpp:74:78: error: declaration of ‘request’ shadows a member of 'this' [-Werror=shadow]
 void DefaultFileSource::cancel(const Resource& resource, FileRequest* request) {
                                                                              ^
```

@miccolis Is removing the JavaScript `cancel` interface an issue at the implementation level?

miccolis:
@mikemorris removing `cancel` is fine.

include/mbgl/storage/file_source.hpp
tmpsantos: Node request could have a no-op cancel maybe.

Issue #3037, opened 2015-11-15
Polyline and markers drawn underneath road in Streets style


tomtaylor:
I've just built CityCyclist against c35f18fdbc, and I managed to get it into a situation, shortly after launch, where the route polyline and marker were drawn underneath the roads, but above the background colour. I quit and restarted and the issue went away, and I can't reproduce it again.

![img_9536](https://cloud.githubusercontent.com/assets/565/11169700/320a3af0-8bb7-11e5-8af6-5a77807ef6b2.jpg)

friedbunny:
Interesting! /cc @jfirebaugh

jfirebaugh:
Looks like this bug can happen when annotations are added before the style finishes loading. In that scenario, it's possible for the annotation layers to get added first, and then when the style JSON loads, the style layers get appended.

tomtaylor:
@jfirebaugh that seems likely in this situation. Is it possible to serve the style JSON locally to reduce the likelihood of this?

friedbunny:
Loading styles locally is broken (#3050) at the moment.

Zakay:
+1 I'm having the same issue, is there no way to play with z indices as a temporary fix?

Zakay:
@tmpsantos This commit seems to have created this issue: 36581f3d3015d525db92248004e9dc7477705694

Zakay:
I have been rolling back commit after commit to find the commit that caused this issue, and I found a candidate: f9ebe54a6336431af98ebfb428d28a0513b2522d @jfirebaugh

Zakay:
This happens also when adding annotations after loading the map.

Zakay:
@jfirebaugh I've updated to latest and still get this issue, but much less now. I have re-place an annotation by long-press on the map. The annotation is sometimes visible and sometimes not.

tomtaylor:
Yeah, I also managed to get an invisible annotation that became visible when I removed and readded it.

Zakay:
To reproduce modify handleLongPress on the demo app to remove other annotations.
And then randomly place dropped pins on the map.

It's tiresome but sometimes it happens a lot and sometimes you need to give in some time.

This issue is the only blocking issue we have left before releasing a major version of our app, and really look forward to have this fixed. Rolling back is no option as latest commits also have fixed serious crashing issues we've had.

```objc
- (void)handleLongPress:(UILongPressGestureRecognizer *)longPress
{

[...]

PR #3049, opened 2015-11-17
Separate annotation sprite atlas from the style, proposed fix for #1488

adam-mapbox:
@kkaefer @jfirebaugh @incanus 

Comments about thread safety, etc. welcome.

### DIFF
--- a/src/mbgl/annotation/annotation_manager.cpp
+++ b/src/mbgl/annotation/annotation_manager.cpp
@@ -10,7 +10,11 @@ namespace mbgl {
 const std::string AnnotationManager::SourceID = "com.mapbox.annotations";
 const std::string AnnotationManager::PointLayerID = "com.mapbox.annotations.points";
 
-AnnotationManager::AnnotationManager() = default;
+AnnotationManager::AnnotationManager(float pixelRatio)
+    : spriteStore(pixelRatio),
+      spriteAtlas(512, 512, pixelRatio, spriteStore) {
+}
+
 AnnotationManager::~AnnotationManager() = default;
 
 AnnotationIDs
@@ -122,6 +126,7 @@ void AnnotationManager::updateStyle(Style& style) {
         layer->sourceLayer = PointLayerID;
         layer->layout.icon.image = std::string("{sprite}");
         layer->layout.icon.allowOverlap = true;
+        layer->spriteAtlas = &spriteAtlas;
 
         style.addLayer(std::move(layer));
     }
@@ -150,6 +155,16 @@ void AnnotationManager::addTileMonitor(AnnotationTileMonitor& monitor) {
 
 void AnnotationManager::removeTileMonitor(AnnotationTileMonitor& monitor) {
     monitors.erase(&monitor);
+}
+
+void AnnotationManager::setSprite(const std::string& name, std::shared_ptr<const SpriteImage> sprite) {
+    spriteStore.setSprite(name, sprite);
+    spriteAtlas.updateDirty();
+}
+
+double AnnotationManager::getTopOffsetPixelsForAnnotationSymbol(const std::string& name) {
+    auto sprite = spriteStore.getSprite(name);
+    return sprite ? -sprite->height / 2 : 0;
 }
 
 }
--- a/src/mbgl/annotation/annotation_manager.hpp
+++ b/src/mbgl/annotation/annotation_manager.hpp
@@ -4,6 +4,8 @@
 #include <mbgl/annotation/annotation.hpp>
 #include <mbgl/annotation/point_annotation_impl.hpp>
 #include <mbgl/annotation/shape_annotation_impl.hpp>
+#include <mbgl/sprite/sprite_store.hpp>
+#include <mbgl/sprite/sprite_atlas.hpp>
 #include <mbgl/util/geo.hpp>
 #include <mbgl/util/noncopyable.hpp>
 
@@ -21,7 +23,7 @@ class Style;
 
 class AnnotationManager : private util::noncopyable {
 public:
-    AnnotationManager();
+    AnnotationManager(float pixelRatio);
     ~AnnotationManager();
 
     AnnotationIDs addPointAnnotations(const std::vector<PointAnnotation>&, const uint8_t maxZoom);
@@ -30,6 +32,10 @@ public:
 
     AnnotationIDs getPointAnnotationsInBounds(const LatLngBounds&) const;
     LatLngBounds getBoundsForAnnotations(const AnnotationIDs&) const;
+
+    void setSprite(const std::string& name, std::shared_ptr<const SpriteImage>);
+    double getTopOffsetPixelsForAnnotationSymbol(const std::string& name);
+    SpriteAtlas& getSpriteAtlas() { return spriteAtlas; }
 
     void updateStyle(Style&);
 
@@ -48,6 +54,9 @@ private:
     ShapeAnnotationImpl::Map shapeAnnotations;
     std::vector<std::string> obsoleteShapeAnnotationLayers;
     std::set<AnnotationTileMonitor*> monitors;
+
+    SpriteStore spriteStore;
+    SpriteAtlas spriteAtlas;
 };
 
 }
--- a/src/mbgl/layer/symbol_layer.hpp
+++ b/src/mbgl/layer/symbol_layer.hpp
@@ -7,6 +7,8 @@
 
 namespace mbgl {
 
+class SpriteAtlas;
+
 class SymbolLayoutProperties {
 public:
     LayoutProperty<PlacementType> placement = PlacementType::Point;
@@ -94,6 +96,8 @@ public:
 
     SymbolLayoutProperties layout;
     SymbolPaintProperties paint;
+
+    SpriteAtlas* spriteAtlas;
 };
 
 }
--- a/src/mbgl/map/map_context.cpp
+++ b/src/mbgl/map/map_context.cpp
@@ -243,7 +243,7 @@ bool MapContext::renderSync(const TransformState& state, const FrameData& frame)
     glObjectStore.performCleanup();
 
     if (!painter) painter = std::make_unique<Painter>(data, transformState);
-    painter->render(*style, frame);
+    painter->render(*style, frame, data.getAnnotationManager()->getSpriteAtlas());
 
     if (data.mode == MapMode::Still) {
         callback(nullptr, std::move(view.readStillImage()));
@@ -269,12 +269,7 @@ bool MapContext::isLoaded() const {
 
 double MapContext::getTopOffsetPixelsForAnnotationSymbol(const std::string& symbol) {
     assert(util::ThreadContext::currentlyOn(util::ThreadType::Map));
-    auto sprite = style->spriteStore->getSprite(symbol);
-    if (sprite) {
-        return -sprite->height / 2;
-    } else {
-        return 0;
-    }
+    return data.getAnnotationManager()->getTopOffsetPixelsForAnnotationSymbol(symbol);
 }
 
 void MapContext::setSourceTileCacheSize(size_t size) {
@@ -299,14 +294,8 @@ void MapContext::onLowMemory() {
 }
 
 void MapContext::setSprite(const std::string& name, std::shared_ptr<const SpriteImage> sprite) {
-    if (!style) {
-        Log::Info(Event::Sprite, "Ignoring sprite without stylesheet");
-        return;
-    }
-
-    style->spriteStore->setSprite(name, sprite);
-
-    style->spriteAtlas->updateDirty();
+    assert(util::ThreadContext::currentlyOn(util::ThreadType::Map));
+    data.getAnnotationManager()->setSprite(name, sprite);
 }
 
 void MapContext::onTileDataChanged() {
--- a/src/mbgl/map/map_context.cpp
+++ b/src/mbgl/map/map_context.cpp
@@ -25,9 +25,10 @@
 
 namespace mbgl {
 
-MapContext::MapContext(View& view_, FileSource& fileSource, MapData& data_)
+MapContext::MapContext(View& view_, FileSource& fileSource, MapMode mode_, GLContextMode contextMode_, const float pixelRatio_)
     : view(view_),
-      data(data_),
+      dataPtr(std::make_unique<MapData>(mode_, contextMode_, pixelRatio_)),
+      data(*dataPtr),
       asyncUpdate([this] { update(); }),
       asyncInvalidate([&view_] { view_.invalidate(); }),
       texturePool(std::make_unique<TexturePool>()) {
@@ -57,6 +58,7 @@ void MapContext::cleanup() {
     style.reset();
     painter.reset();
     texturePool.reset();
+    dataPtr.reset();
 
     glObjectStore.performCleanup();
 
--- a/src/mbgl/map/map_context.hpp
+++ b/src/mbgl/map/map_context.hpp
@@ -5,6 +5,7 @@
 #include <mbgl/map/update.hpp>
 #include <mbgl/map/transform_state.hpp>
 #include <mbgl/map/map.hpp>
+#include <mbgl/map/map_data.hpp>
 #include <mbgl/style/style.hpp>
 #include <mbgl/util/async_task.hpp>
 #include <mbgl/util/gl_object_store.hpp>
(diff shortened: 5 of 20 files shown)


incanus:
Digging into this now @adam-mapbox, but some code review / understanding comments: 

1. We should follow the pattern `Style& style, SpriteAtlas* atlas` for arguments (i.e. consistent spacing). 

1. The member `sourceSpriteAtlas` has some potential for confusion as we have the concept of sources, but this doesn't have anything to do with them. 

1. In routines like `MapContext::setSprite`, we should codify assumptions about threading such as in nearby routines (`assert(util::ThreadContext::currentlyOn(util::ThreadType::Map));`). 

1. I'm not quite following why you'd fall back to times with no layer-specific sprite, such as `SpriteAtlas *useSpriteAtlas = layer.sourceSpriteAtlas ? layer.sourceSpriteAtlas : spriteAtlas;`. Can you clarify? That might help point to a more descriptive name for `sourceSpriteAtlas` above. 

And I'd be curious if @jfirebaugh or @kkaefer have any thoughts here about adding functionality to the `MapContext` given longer-term plans to fold that back into the `Map` proper. 

I'm still doing some investigations into thread safety. It seems like we might be able to indicate some `const` around this to clarify that things aren't mutated unexpectedly, too.

incanus:
> I'm not quite following why you'd fall back to times with no layer-specific sprite

Oh, I think I follow — the normal style sprite. And a single annotation sprite, which doesn't come into play if there are no annotations. 

In that case, do we / can we destroy it when all image annotations are removed?

incanus:
Why is `SpriteStore::Observer` implemented (with empty method bodies)? As a todo for later? If we don't need to run anything here now, do we need to implement the observer?

incanus:
Ah, so `SpriteAtlas::getImage()` mutates the internal `images` as a `Holder`, hence we can't pass a `const SpriteAtlas` around. Internally, that routine already uses a mutex, so I think we're clear here thread-wise.

incanus:
Also, yeah, totally weird failing test on Travis here (e.g. https://travis-ci.org/mapbox/mapbox-gl-native/jobs/91681030). It's expecting the custom sprite functionality to log 21 errors and none are. Any insight there @kkaefer since you wrote the test?

adam-mapbox:
The intent was that `sourceSpriteAtlas` is the sprite atlas that the `SymbolLayer` uses as the source for its symbol images.

adam-mapbox:
I'm at a loss about what to do about this failing test.  I'm out of my depth here; the test is failing on linux only, and I don't understand the test or what it's supposed to do, and I don't know how to reproduce it.  I could use some assistance.

incanus:
> I'm at a loss about what to do about this failing test. I'm out of my depth here; the test is failing on linux only, and I don't understand the test or what it's supposed to do, and I don't know how to reproduce it. I could use some assistance.

Agree, but the reason we're seeing it only on Linux is that's the [only place these core tests are run](https://github.com/mapbox/mapbox-gl-native/blob/0a8d9a3b6d686c1fce9dc48014d1b13fdd58cdb3/scripts/linux/run.sh#L28). So at least we're consistent (i.e. not a flapping test). 

Going to play with this test locally a bit.

incanus:
> The intent was that `sourceSpriteAtlas` is the sprite atlas that the `SymbolLayer` uses as the source for its symbol images.

Maybe `layerSpriteAtlas` since it's within the `SymbolLayer`?

adam-mapbox:
How do I run the core tests locally?  Can I?

incanus:
I think, as in [Travis](https://github.com/mapbox/mapbox-gl-native/blob/0a8d9a3b6d686c1fce9dc48014d1b13fdd58cdb3/scripts/linux/run.sh#L28), it's just `make test-*`. I don't know offhand how to run a single, but looking into it.

incanus:
Figured out the format here @adam-mapbox: 

```
make test-Sprite.CustomSpriteImages
```

http://stackoverflow.com/questions/7208070/googletest-how-to-skip-a-test was useful here.

incanus:
Clarified docs at b07d4c8.

jfirebaugh:
:+1: Will add some line by line comments for small improvements, but overall I like this approach a lot.

src/mbgl/annotation/annotation_manager.cpp
```diff
@@ -1,7 +1,7 @@
 #include <mbgl/annotation/annotation_manager.hpp>
 #include <mbgl/annotation/annotation_tile.hpp>
-#include <mbgl/style/style.hpp>
 #include <mbgl/layer/symbol_layer.hpp>
+#include <mbgl/style/style.hpp>
```
jfirebaugh: Avoid reordering the includes here.

src/mbgl/annotation/annotation_manager.cpp
```diff
@@ -105,7 +105,7 @@ std::unique_ptr<AnnotationTile> AnnotationManager::getTile(const TileID& tileID)
     return tile;
 }
 
-void AnnotationManager::updateStyle(Style& style) {
+void AnnotationManager::updateStyle(Style& style, SpriteAtlas* atlas) {
```
jfirebaugh: Take `SpriteAtlas` by reference here.

src/mbgl/annotation/annotation_manager.cpp
```diff
@@ -122,8 +122,14 @@ void AnnotationManager::updateStyle(Style& style) {
         layer->sourceLayer = PointLayerID;
         layer->layout.icon.image = std::string("{sprite}");
         layer->layout.icon.allowOverlap = true;
+        layer->layerSpriteAtlas = atlas;
 
         style.addLayer(std::move(layer));
+    } else {
```
jfirebaugh: Is the else branch necessary here? The only place that the point annotation layer gets added to the style is in the `if` branch above, so it shouldn't be possible for the style not to have the annotation source, but have a point annotation layer without its sprite atlas set.

src/mbgl/layer/symbol_layer.hpp
```diff
@@ -94,6 +95,8 @@ class SymbolLayer : public StyleLayer {
 
     SymbolLayoutProperties layout;
     SymbolPaintProperties paint;
+
+    SpriteAtlas *layerSpriteAtlas;
```
jfirebaugh: Let's just call this `spriteAtlas`, and set the default value to the regular `SpriteAtlas` (this can happen in `Style::addLayer`). Having it always set will be nice because then no conditional checks are necessary.

src/mbgl/annotation/annotation_manager.cpp
adam-mapbox: You guys don't alphabetize includes?

src/mbgl/map/map_context.cpp
```diff
@@ -32,7 +32,10 @@ MapContext::MapContext(View& view_, FileSource& fileSource, MapData& data_)
       data(data_),
       asyncUpdate(std::make_unique<uv::async>(util::RunLoop::getLoop(), [this] { update(); })),
       asyncInvalidate(std::make_unique<uv::async>(util::RunLoop::getLoop(), [&view_] { view_.invalidate(); })),
-      texturePool(std::make_unique<TexturePool>()) {
+      texturePool(std::make_unique<TexturePool>()),
+      savedPixelRatio(data.pixelRatio),
+      annotationSpriteStore(nullptr),
```
jfirebaugh: `nullptr` is the default value for `std::unique_ptr`, so it's not necessary to initialize `annotationSpriteStore` and `annotationSpriteStore` explicitly.

src/mbgl/annotation/annotation_manager.cpp
adam-mapbox: atlas is sometimes null.

src/mbgl/annotation/annotation_manager.cpp
adam-mapbox: No, that's not correct.  There is a flow whereby initially we pass in null for the sprite atlas (because we don't have any annotations), then later we set up annotations and set the sprite atlas.  You can see this flow if you open the sample app and add annotations.

src/mbgl/map/map_context.cpp
```diff
@@ -299,14 +310,16 @@ void MapContext::onLowMemory() {
 }
 
 void MapContext::setSprite(const std::string& name, std::shared_ptr<const SpriteImage> sprite) {
+    assert(util::ThreadContext::currentlyOn(util::ThreadType::Map));
     if (!style) {
```
jfirebaugh: It's no longer necessary to have a style before adding an annotation sprite -- we can now remove this block!

src/mbgl/map/map_context.hpp
```diff
@@ -68,7 +68,7 @@ class MapContext : public Style::Observer {
     void onResourceLoadingFailed(std::exception_ptr error) override;
 
     void dumpDebugLogs() const;
-
+
```
jfirebaugh: Spurious whitespace change.

jfirebaugh:
The annotation `SpriteAtlas` needs to be uploaded in the "upload pass" of `Painter::render`.

src/mbgl/layer/symbol_layer.hpp
adam-mapbox: I don't totally understand.  Can you clarify?

adam-mapbox:
How do I get at the annotation spriteatlas from inside `Painter::render`?  I don't seem to have access to the map context.

src/mbgl/map/map_context.cpp
adam-mapbox: Done.

src/mbgl/map/map_context.hpp
adam-mapbox: Fixed.

jfirebaugh:
Could we make `SpriteAtlas` lazy internally, instead of lazily creating the annotation `SpriteAtlas`? Having possibly null pointers threading through multiple parts of the code gives me pause.

adam-mapbox:
We certainly *could* do that.  Doesn't that just kind of push the problem around?  Making `SpriteAtlas` lazy is going to significantly increase its complexity and possibly introduce weird bugs to unrelated sections of the code.  I don't think it increases the overall quality of the code per se.

src/mbgl/annotation/annotation_manager.cpp
incanus: They're typically roughly in order of "importance", as best I can read. So in this case, we've got our own class header, then the thing we produce, then the thing we use to produce that, then a specific class of many that we need access to. At least that's my logic flow in looking at the old version. Maybe we should settle on a stricter system.

jfirebaugh:
Here's what an internally lazy `SpriteAtlas` looks like: 928128c4610f49a7a2fb2560ddedd17b3a03cde6.

I like the fact that this encapsulates the possibility of a null pointer in a single class and gains us the useful invariant that `SymbolLayer::spriteAtlas` is non-null (once added to the style) -- we trade six conditionals spread out through several files for two conditionals in a single file.

With a bit more work we could make `SymbolLayer::spriteAtlas` a reference and have the compiler check our invariant automatically. This would involve adding a `SpriteAtlas&` argument to the `SymbolLayer` constructor and threading an appropriate parameter through.

BTW, for the `Sprite.CustomSpriteImages` test failure, I encountered the same issue in an unrelated change. If you rebase on f2772309feb394b7caeaea436e13ef45ba86a4e7 it should be resolved.

src/mbgl/annotation/annotation_manager.cpp
jfirebaugh: Yeah, we don't really have a strict system, except that we try to group by path prefix.

adam-mapbox:
Hey @jfirebaugh OK that looks great.  I didn't realize it would be so easy to modify SpriteAtlas.  Thanks for doing that work.

adam-mapbox:
Are you comfortable with rebasing and merging into master at this point?

jfirebaugh:
I've encountered several issues in the course of working on annotations today (investigating #3037 and testing this PR):

* Annotation tests had an undiagnosed failure prior to this PR. Fixed in afa04de15286cdfa1cedc756d8fbe46a42e319a5 and filed a followup ticket in #3082.
* Tests and the GLFW app were still assuming they could use icons from the style sprite for point annotations. As of this change that's no longer the case -- all point annotations must have an icon specified via `Map::setSprite`. I added changes to this PR to fix that.
* Tests were failing [an assert in GLObjectStore](https://github.com/mapbox/mapbox-gl-native/blob/afa04de15286cdfa1cedc756d8fbe46a42e319a5/src/mbgl/util/gl_object_store.cpp#L21). Fixed by moving `MapData` storage to the `MapContext` (in this PR).
* Tests are [crashing](https://travis-ci.org/mapbox/mapbox-gl-native/jobs/92176509) in `~HeadlessView`. Still investigating.

tmpsantos:
> Why is SpriteStore::Observer implemented (with empty method bodies)? As a todo for later? If we don't need to run anything here now, do we need to implement the observer?

The tiles always arrive first and from the tiles we see what sprites are needed. Meanwhile we draw whatever we have. When the sprites finally arrive, the `SpriteStore::Observer` get a signal and repaints the map, this time with sprites. Same goes for the glyphs.

src/mbgl/annotation/annotation_manager.hpp
```diff
@@ -48,6 +54,9 @@ class AnnotationManager : private util::noncopyable {
     ShapeAnnotationImpl::Map shapeAnnotations;
     std::vector<std::string> obsoleteShapeAnnotationLayers;
     std::set<AnnotationTileMonitor*> monitors;
+
+    SpriteStore spriteStore;
```
tmpsantos: Wouldn't be easier to have these objects living the `MapContext` object? They would survive a style change anyway and would have a more clear scope. The crash I'm seeing on Linux is related to destruction order, not threading issues.

We are calling `glObjectStore.performCleanup()` on `MapContext::cleanup()` just before calling `~MapContext()` that will ultimately calls the `~MapData()` -> `~AnnotationManager()` -> `~SpriteAtlas()` that tries to dispose the textures but they never get cleaned up.

This patch fixes the crash on your branch: https://github.com/mapbox/mapbox-gl-native/commit/c15e1e6945dee649f2ef705211a9c22c1416854a

jfirebaugh: They are specific to annotations, so I'd rather have them owned by `AnnotationManager`.

Your analysis of the issue sounds correct; however I would rather fix it by using `std::unique_ptr<AnnotationManager>` in `MapData` and resetting that.

tmpsantos: > Your analysis of the issue sounds correct; however I would rather fix it by using std::unique_ptr<AnnotationManager> in MapData and resetting that.

That also should work.

incanus:
Sounds like from voice that @jfirebaugh is wrapping up test accuracy here, then we can crunch this a bit into a final proposed PR @adam-mapbox?

jfirebaugh:
I'm on it.

incanus:
:+1: Cool just syncing up.

jfirebaugh:
d55aa7929cb10d40a58b6b7a8ed73bddd4f0a407 introduced some changes that conflict with this work. I'm proposing to revert them in #3157, but this PR will now need to wait for that to get sorted out.

incanus:
:tada:

cowgp:
@incanus  Was this fix not included in iOS 3.0.1 (released after fix was committed) ?   I'm using 3.0.1 and still seeing this.

jfirebaugh:
@cowgp That's correct. The 3.0.1 release branch was created before this fix was implemented, and it was deemed too risky to backport. It will be in the next iOS SDK release.

cowgp:
@jfirebaugh  - thanks.  That obviously begs the question, what's an ETA on the next release?

1ec5:
Added to the changelog in 0d3af143756381fe5a96336fb5882a26301302ca.

@cowgp, the 3.1.0 release is coming soon. In the meantime, you can try out the fix in [this prerelease](https://github.com/mapbox/mapbox-gl-native/releases/tag/ios-v3.1.0-pre.2) ([podspec](https://github.com/mapbox/mapbox-gl-native/blob/ios-v3.0.0-pre.2/ios/Mapbox-iOS-SDK.podspec) for CocoaPods).

cowgp:
Thanks @1ec5 - I built a custom version off the master branch last week and pointed to a custom podspec for that.  I noticed the 3.1.0-pre.X podspecs have the deployment target bumped to 8.0.  Do you know why? 
Happy to see 3.1 is rolling out officially soon though - I'll swap back to official cocoapod once it releases.

1ec5:
> I noticed the 3.1.0-pre.X podspecs have the deployment target bumped to 8.0. Do you know why? 

Starting in 3.1.0, the pod will download a dynamic framework, which is for iOS 8.0 and above. If you need iOS 7.x support, you’ll need to download the static framework manually from the releases page.

Issue #1662, opened 2015-05-29
Create OpenGL on Android with GLSurfaceView


kkaefer:
Instead of creating the OpenGL context on the C++ side, we should use the `GLSurfaceView` object and only use the JNI bindings to implement the `GLSurfaceView.Renderer#onDrawFrame()` method.

hanchao:
+1 
See https://github.com/tangrams/tangram-es

incanus:
@ljbade has been researching this.

ljbade:
Progress being made in https://github.com/mapbox/mapbox-gl-native/tree/android-glsurfaceview

Currently compiles, have not tried running it yet.

Still lots of work remaining currently TODOed.

ljbade:
Progress now at the first native crash:
```
07-31 12:09:38.414  29174-29251/com.mapbox.mapboxgl.testapp A/libc﹕ Fatal signal 11 (SIGSEGV), code 1, fault addr 0x0 in tid 29251 (GLThread 165793)
```

ljbade:
Fixed that crash caused by something in the GL version logging. This will be moved to Java at some stage. In mean time I just commented it out.

I now get a black map but I can move it around and rotate it.

Building debug build to see what sort of GL errors we are getting.

ljbade:
Crashes somewhere in the GL thread at `log.cpp` line 38. Not sure why yet need to trace it.

But commenting out this line allows the us to get to an actual GL error which triggers an `assert()`.

ljbade:
Ah it looks like it is related to the thread context stuff.

```
#0  basic_string (this=<optimized out>, __str=...) at ../../src/mbgl/util/thread_context.hpp:39
#1  getName () at ../../src/mbgl/util/thread_context.hpp:39
#2  mbgl::Log::record (severity=mbgl::Debug, event=mbgl::JNI, code=4294967295, msg=...) at ../../src/mbgl/platform/log.cpp:52
#3  0xa1b2c774 in mbgl::Log::record (severity=mbgl::Debug, event=mbgl::JNI, format=<optimized out>) at ../../src/mbgl/platform/log.cpp:38
#4  0xa1adae00 in Record<char const (&) [21]> (severity=mbgl::Debug, event=mbgl::JNI, args=...) at ../../include/mbgl/platform/log.hpp:59
#5  Debug<char const (&) [21]> (event=mbgl::JNI, args=...) at ../../include/mbgl/platform/log.hpp:36
[...]

PR #2909, opened 2015-11-17
Eliminate separate "MapContext" thread

jfirebaugh:
We're currently rendering on the MapContext thread. However, when triggering a render call from the main thread, we're also blocking the main thread, so there's no advantage from parallelization here, but we have the overhead of managing the OpenGL context on the MapContext thread. Instead, we should split up Mapbox GL into two parts: the MapContext thread manages the state of the objects, coordinates tile loading and style updates, the Painter thread (which can run either on the MapContext thread, the main thread, or an entirely different thread) that just takes the state the MapContext thread produces, and renders a frame.

-----

TODO list:

* [x] Eliminate use of `ThreadContext` for obtaining FileSource / GLObjectStore
  * [x] Restore GLObjectStore, but pass it in wherever needed
* [x] iOS / OSX
 * [x] Find the right place to call `mbgl::gl::InitializeExtensions` (fixes "Not using Vertex Array Objects" warning)
 * [x] Convert RunLoop implementation to pure C++
 * [x] Figure out why glfw app exits immediately
* [x] Android
 * [x] Update `View` implementation
 * [x] Write native `RunLoop` / `Timer` / `AsyncTask` implementation
 * [x] Fix https://github.com/mapbox/mapbox-gl-native/issues/4682
* [x] Node

### DIFF
--- a/platform/android/src/native_map_view.hpp
+++ b/platform/android/src/native_map_view.hpp
@@ -24,10 +24,7 @@ public:
     std::array<uint16_t, 2> getFramebufferSize() const override;
     void activate() override;
     void deactivate() override;
-    void notify() override;
     void invalidate() override;
-    void beforeRender() override;
-    void afterRender() override;
 
     void notifyMapChange(mbgl::MapChange) override;
 
@@ -43,8 +40,7 @@ public:
     void createSurface(ANativeWindow *window);
     void destroySurface();
 
-    void resume();
-    void pause();
+    void render();
 
     void enableFps(bool enable);
     void updateFps();
--- a/platform/android/src/native_map_view.hpp
+++ b/platform/android/src/native_map_view.hpp
@@ -57,6 +57,7 @@ private:
 
 private:
     JavaVM *vm = nullptr;
+    JNIEnv *env = nullptr;
     jweak obj = nullptr;
 
     ANativeWindow *window = nullptr;
@@ -83,9 +84,6 @@ private:
 
     int availableProcessors = 0;
     size_t totalMemory = 0;
-
-    jboolean renderDetach = false;
-    JNIEnv *renderEnv = nullptr;
 
     // Ensure these are initialised last
     std::unique_ptr<mbgl::DefaultFileSource> fileSource;
--- a/include/mbgl/map/map.hpp
+++ b/include/mbgl/map/map.hpp
@@ -32,10 +32,6 @@ class ShapeAnnotation;
 struct CameraOptions;
 struct AnimationOptions;
 
-namespace util {
-template <class T> class Thread;
-} // namespace util
-
 class Map : private util::noncopyable {
     friend class View;
 
@@ -46,22 +42,15 @@ public:
                  ConstrainMode constrainMode = ConstrainMode::HeightOnly);
     ~Map();
 
-    // Pauses the render thread. The render thread will stop running but will not be terminated and will not lose state until resumed.
-    void pause();
-    bool isPaused();
-
-    // Resumes a paused render thread
-    void resume();
-
     // Register a callback that will get called (on the render thread) when all resources have
     // been loaded and a complete render occurs.
     using StillImageCallback = std::function<void (std::exception_ptr, PremultipliedImage&&)>;
     void renderStill(StillImageCallback callback);
 
-    // Triggers a synchronous render.
-    void renderSync();
+    // Main render function.
+    void render();
 
-    // Notifies the Map thread that the state has changed and an update might be necessary.
+    // Notifies the Map that the state has changed and an update might be necessary.
     void update(Update update);
 
     // Styling
@@ -189,7 +178,7 @@ public:
 private:
     View& view;
     const std::unique_ptr<Transform> transform;
-    const std::unique_ptr<util::Thread<MapContext>> context;
+    const std::unique_ptr<MapContext> context;
     MapData* data;
 
     enum class RenderState {
--- a/include/mbgl/map/view.hpp
+++ b/include/mbgl/map/view.hpp
@@ -30,43 +30,41 @@ enum MapChange : uint8_t {
 
 class View {
 public:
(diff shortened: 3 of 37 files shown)


jfirebaugh:
> `glfwPollEvents()`

Ah, good call. I pushed a proof of concept of that and converted this to a PR. It also includes stuff from `tmpsantos-libuv_abstracted (with a few fixes).

jfirebaugh:
Next step here is to peel off the parts from `tmpsantos-libuv_abstracted` related to removing libuv dependencies from the FileSource thread. That's a bigger chunk of work, mainly due to the use of libuv IO functionality, that we don't need to bite off right now.

The model for `render-main` will be that the main thread uses the native event loop, with a `RunLoop` wrapper that isn't actually `run()` except for the GLFW app, and other threads use a libuv loop.

jfirebaugh:
Scratch that, it doesn't actually work, it'd require conflicting `RunLoop` implementations. The main thread needs a `RunLoop` backed by the native event loop, while the FileSource and worker threads need a `RunLoop` backed by libuv. I think the answer is to remove the direct dependency of `MapContext` on `uv::async`/`AsyncTask` entirely. "Update" should be an operation the MapContext asks the view to do, and the view takes care of throttling / deduplicating updates in a platform-specific manner. (We already have something like this in the vestigial `View::notify` method.) So I'm not going to use most of `libuv_abstracted` (yet) after all.

jfirebaugh:
...except that _that_ doesn't work because the `FileSource` thread depends on the rendering thread having an active `RunLoop` for `invokeWithCallback`. Ugh, I keep thinking in circles here.

kkaefer:
> "Update" should be an operation the MapContext asks the view to do, and the view takes care of throttling / deduplicating updates in a platform-specific manner. 

Yeah, we are sort of having something like that already with `view.invalidate()`.

tmpsantos:
I'm working on this. The changes needed for adding a new platform `RunLoop` were pushed at #3139 and they don't change anything in the client API, thus can be already merged.

For rendering in the main thread, I'm also experimenting an alternative and less intrusive approach that would be a main loop integration (glib-style), but that would also depend on #3139.

tmpsantos:
I managed to create a main loop integration mechanism that is extremely simple and worked well. @jfirebaugh please have a look on the 5 last patches of this branch here (the rest of the delta with master is basically #3139):

https://github.com/tmpsantos/mapbox-gl-native/commits/render-main

jfirebaugh:
So this is the "Option 2" approach. How will `Timer` notifications wake the main thread with this approach?

tmpsantos:
> So this is the "Option 2" approach. How will Timer notifications wake the main thread with this approach?

They won't work. I documented it here: https://github.com/mapbox/mapbox-gl-native/commit/9a1128708e2d7219ed4d9c6f109eddf2f4471223

We don't use timers on the main thread anyway and the commit adds an assertion for that. But this is not final, I'm still experimenting with this approach.

jfirebaugh:
:+1:

jfirebaugh:
Making progress here; iOS and OS X are working, with the caveat that VAOs are not working on OS X at the moment because of an extension initialization issue which I haven't figured out.

Node is tough. Eliminating the map thread means there are potentially multiple GL contexts on a single thread. I suspect that Map does not correctly activate the view before all GL operations that can potentially interleave. We may need to activate not only at the beginning of every `MapContext::update` but potentially whenever we get a callback from an async operation like a FileSource request, if the callback triggers GL operations. @kkaefer @tmpsantos do you have a sense if that's required and the performance cost of adding many context switches?

This may well indicate that a similar problem exists on other platforms, if multiple Maps are created.

tmpsantos:
As per chat conversation, [Looper](http://developer.android.com/reference/android/os/Looper.html) seems to be the way to go for implementing the main loop for Android. I will give it a try.

/cc @zugaldia

jfirebaugh:
I rebased this on `master` today. Current status:

* iOS, OS X, and node are in good shape
* Blocked on Android by #4682 
* On linux, `Annotations.FillAnnotation` and `Annotations.NonImmediateAdd` core unit tests are failing due to an unknown issue. The "actual" rendering looks like this:
 image
 
 Expected:
 image

tmpsantos:
> On linux, Annotations.FillAnnotation and Annotations.NonImmediateAdd core unit tests are failing due to an unknown issue. The "actual" rendering looks like this:

Made progress today on #4682. On that branch, all tests are passing on Linux.

jfirebaugh:
Thanks @tmpsantos. I integrated your changes into this PR. However, when running the Android test app, I only see a white screen, no map.

platform/android/src/native_map_view.hpp
```diff
@@ -88,6 +89,7 @@ class NativeMapView : public mbgl::View, private mbgl::util::noncopyable {
     JNIEnv *renderEnv = nullptr;
 
     // Ensure these are initialised last
+    std::unique_ptr<mbgl::util::RunLoop> loop;
```
jfirebaugh: @tmpsantos Is it ok to create multiple `RunLoops` representing the main thread, if multiple map views are created? Or does this need to be a singleton somewhere else?

tmpsantos: Nice catch, it should be a singleton. I will fix that.

jfirebaugh:
> On linux, Annotations.FillAnnotation and Annotations.NonImmediateAdd core unit tests are failing due to an unknown issue.

This turned out to be caused by #3883.

tmpsantos:
> However, when running the Android test app, I only see a white screen, no map.

It works on Nexus 5x just fine but throws an exception on Nexus 5 when trying to swap buffers.

I don't think it has to do with the `RunLoop` anymore, but how we manage the GL context. It is now over complicated for rendering in the main thread and we should do it on the Java side. I will give #1662 a try.

jfirebaugh:
Android rendering issues were due to missing `eglSwapBuffers`; fixed. I think this is ready!

PR #3055, opened 2015-11-17
Allow setting pitch in the GLFW app

kkaefer:


kkaefer:
Hold down `Shift` and drag up/down to change pitch.

PR #3058, opened 2015-11-17
fixes #3054: move desktop test app keybindings output to startup log

### LINKED ISSUE #3054: Show keybindings in OSX/Linux applications
Instead of only showing our keybindings in Markdown documentation, we should make them obvious within the UI of the test applications themselves.

incanus:


platform/default/glfw_view.cpp
```diff
@@ -82,6 +82,20 @@ GLFWView::GLFWView(bool fullscreen_, bool benchmark_)
...
+    printf("- Press `S` to cycle through bundled styles\n");
+    printf("- Press `X` to reset the transform\n");
+    printf("- Press `N` to reset north\n");
+    printf("- Press `Tab` to toggle debug information\n");
+    printf("- Press `C` to toggle symbol collision debug boxes\n");
+    printf("- Press `Esc` to quit\n");
+    printf("- Press `1` through `6` to add increasing numbers of point annotations for testing\n");
+    printf("\n");
+    printf("================================================================================\n");
+    printf("\n");
```
kkaefer: Misses `Z`, `R`, `Q`, `P`, `W`, `7-9`, as well as right-click+ drag to rotate and shift+drag to tilt.

incanus: Whoa! Missed those. They've never been in the Markdown docs anyway! Case in point for having them in the same file as the actual keybinding definitions.

incanus: b78a55de4c6e86248c800858ef6f503a42916de8

PR #3059, opened 2015-11-17
[WIP] cleanup podspec, packaging script

friedbunny:
- Remove MGLDummy.m (#1426)
- Remove `-ObjC` linker flag (#2966)

Hold for now, this PR currently causes existing projects to break — for some reason CocoaPods is failing to add the Mapbox module, which breaks `#import <Mapbox/Mapbox.h>`.

/cc @1ec5 @incanus

### DIFF
--- a/ios/Mapbox-iOS-SDK.podspec
+++ b/ios/Mapbox-iOS-SDK.podspec
@@ -23,7 +23,7 @@ Pod::Spec.new do |m|
   m.requires_arc = true
 
   m.preserve_paths = '**'
-  m.source_files = 'Headers/*.h', 'MGLDummy.m'
+  m.source_files = 'Headers/*.h'
   m.resource_bundle = { 'Mapbox' => 'Mapbox.bundle/*' }
   m.vendored_library = 'libMapbox.a'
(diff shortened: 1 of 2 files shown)


friedbunny:
OK, we still need MGLDummy.m because of https://github.com/CocoaPods/CocoaPods/issues/3499. We distribute a static library with headers, which needs special massaging. Looks like if we switch to distributing a framework or source we could avoid this problem.

Closing here, will break out the removal of the `-ObjC` flag to another PR.

incanus:
We removed the `-ObjC` flag already in https://github.com/mapbox/mapbox-gl-native/commit/423cfc11e5ddc3c0464264c876c74f63c456ce5d.

friedbunny:
Our podspec has its own `-ObjC` flag: [Mapbox-iOS-SDK.podspec#L35](https://github.com/mapbox/mapbox-gl-native/blob/fcec19b8357e65d39d429a175001ef4fc1876686/ios/Mapbox-iOS-SDK.podspec#L35)

... removing it is also breaking things, though.

PR #3060, opened 2015-11-17
more clearly indicate Foundation imports to avoid recursion

incanus:
Due to https://github.com/mapbox/mapbox-gl-native/commit/0a8d9a3b6d686c1fce9dc48014d1b13fdd58cdb3, we recursively import `Mapbox.h` and we shouldn't do this. Instead, we should bring in any Apple Foundation imports directly, our own superclasses / category-extended classes if necessary, and forward-declare the rest. 

/cc @1ec5

### DIFF
--- a/include/mbgl/ios/MGLGeometry.h
+++ b/include/mbgl/ios/MGLGeometry.h
@@ -1,9 +1,8 @@
-#pragma once
-
-#import "MGLTypes.h"
-
+#import <Foundation/Foundation.h>
 #import <CoreLocation/CoreLocation.h>
 #import <CoreGraphics/CGBase.h>
+
+#import "MGLTypes.h"
 
 NS_ASSUME_NONNULL_BEGIN
 
--- a/include/mbgl/ios/MGLAccountManager.h
+++ b/include/mbgl/ios/MGLAccountManager.h
@@ -1,3 +1,5 @@
+#import <Foundation/Foundation.h>
+
 #import "MGLTypes.h"
 
 NS_ASSUME_NONNULL_BEGIN
--- a/include/mbgl/ios/MGLAnnotation.h
+++ b/include/mbgl/ios/MGLAnnotation.h
@@ -1,3 +1,4 @@
+#import <Foundation/Foundation.h>
 #import <CoreLocation/CoreLocation.h>
 
 #import "MGLTypes.h"
--- a/include/mbgl/ios/MGLMapCamera.h
+++ b/include/mbgl/ios/MGLMapCamera.h
@@ -1,6 +1,8 @@
-#import "Mapbox.h"
+#import <Foundation/Foundation.h>
+#import <CoreGraphics/CoreGraphics.h>
+#import <CoreLocation/CoreLocation.h>
 
-#pragma once
+#import "MGLTypes.h"
 
 NS_ASSUME_NONNULL_BEGIN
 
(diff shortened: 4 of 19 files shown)


incanus:
Here's what I see with b07d4c8ff61b869e1a01c445ae1dc479323fa2c3 without this: 

```
...Objects-normal/i386/MGLUserLocation.o

In file included from /Users/incanus/Documents/Projects/Mapbox/gl-native/platform/ios/MGLUserLocation.m:3:
In file included from /Users/incanus/Documents/Projects/Mapbox/gl-native/gyp/../include/mbgl/ios/MGLMapView.h:2:
In file included from /Users/incanus/Documents/Projects/Mapbox/gl-native/gyp/../include/mbgl/ios/MGLMapCamera.h:1:
In file included from /Users/incanus/Documents/Projects/Mapbox/gl-native/gyp/../include/mbgl/ios/Mapbox.h:7:
/Users/incanus/Documents/Projects/Mapbox/gl-native/gyp/../include/mbgl/ios/MGLMapView+IBAdditions.h:6:12: error: cannot define category for undefined class 'MGLMapView'
@interface MGLMapView (IBAdditions)
           ^
In file included from /Users/incanus/Documents/Projects/Mapbox/gl-native/platform/ios/MGLUserLocation.m:1:
/Users/incanus/Documents/Projects/Mapbox/gl-native/platform/ios/MGLUserLocation_Private.h:5:8: note: forward declaration of class here
@class MGLMapView;
       ^
1 error generated.
```

1ec5:
This is unnecessary. MGLMapView.h should have a `#pragma once` guard.

incanus:
I'm all for a simple fix! Will check it out.

include/mbgl/ios/MGLGeometry.h
```diff
@@ -1,10 +1,11 @@
 #pragma once
 
-#import "MGLTypes.h"
```
1ec5: This breaks compilation on Xcode 6.2, because MGLTypes.h is responsible for defining `NS_ASSUME_NONNULL_BEGIN` et al. on older Xcodes. I don’t know if we still want to support 6.2 anymore.

incanus:
This `#pragma once` required here because we're doing Objective-C++? Because I thought `#import` already had an import guard.

incanus:
I'm not seeing `#pragma once` help here. @1ec5 Do you see compilation problems with `master` as of https://github.com/mapbox/mapbox-gl-native/commit/b07d4c8ff61b869e1a01c445ae1dc479323fa2c3?

1ec5:
Hm, you’re right. On second thought, I do like the idea of explicitly declaring the imports like we used to, instead of relying on the umbrella header.

I’d still omit the Foundation import because MGLTypes.h is being imported into every file and it already imports Foundation (since its main job is to implement macros based on Foundation macros).

incanus:
> I’d still omit the Foundation import

It still feels "most correct" to bring in Foundation because of objects inheriting from `NSObject` or using `NSString` or what-have-you, but I could let it pass.

1ec5:
To me, it feels like something you’d put in a .pch, and MGLTypes.h feels like a similarly universal header, but it makes little difference to me either way. Merge away.

incanus:
I've long avoided precompiled headers due to http://qualitycoding.org/precompiled-headers/ (refs https://github.com/mapbox/mapbox-ios-sdk-legacy/issues/301). 

Will merge.

incanus:
Will rebase if all checks out here.

PR #3061, opened 2015-11-18
Ensure proper delegate callbacks for resetNorth

1ec5:
Any MGLMapView method that can programmatically move the map’s viewport, with or without animation, needs to explicitly send a didChange notification. Otherwise, mbgl will trigger notifications, but not necessarily in the right order. This includes the programmatic rotation that occurs due to tracked heading changes.

This PR also fixes a bug in which setting the direction programmatically would fail to boot the user out of course-tracking mode.

/ref #2948
/cc @friedbunny @incanus

### DIFF
--- a/platform/ios/MGLMapView.mm
+++ b/platform/ios/MGLMapView.mm
@@ -1555,7 +1555,7 @@ std::chrono::steady_clock::duration durationInSeconds(float duration)
 
 - (void)resetNorthAnimated:(BOOL)animated
 {
-    _mbglMap->setBearing(0, durationInSeconds(animated ? MGLAnimationDuration : 0));
+    [self setDirection:0 animated:animated];
 }
 
 - (void)resetPosition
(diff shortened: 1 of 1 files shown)


incanus:
LGTM

incanus:
Carrying over from https://github.com/mapbox/mapbox-gl-native/commit/05f81c861b63ae2934dd9a9b18db6585f5d01df0#commitcomment-14479567 for rebases: 

> Hmm, now I'm concerned with testing this sufficiently before potential release today. This feels more invasive than the previous iteration.

1ec5:
It is slightly more invasive: I roped in the rotation that happens programmatically when the tracked bearing changes, and programmatic rotations (other than tracking bearing changes) boot you out of course tracking.

incanus:
This is so small, let's set it for 3.0.0 with plans to revert if we have complications. 

Cherry-picked as fd0e21321e40d1bc503a1c2b6dadb677ea2c6940.

Issue #2775, opened 2015-10-23
Make iOS delegate notifications consistent with MapKit


friedbunny:
[MapKit](https://developer.apple.com/library/prerelease/ios/documentation/MapKit/Reference/MKMapViewDelegate_Protocol/index.html) `loading` delegate callbacks are exclusively about fetching map data, while `rendering` ones are exclusively the drawing of the map.

The flow for Mapkit looks like this on initialization:

```
2015-10-23 19:39:28.470 start RENDERING
2015-10-23 19:39:28.503 start LOADING
2015-10-23 19:39:29.435 finish LOADING
2015-10-23 19:39:30.470 finish RENDERING fullyRendered: YES
```

Panning around continues to issue the same notification flow.

**We don't follow MapKit conventions currently.** Instead, at the core-level:
- :x: `MapChangeWillStartLoading` notifications are fired [when the style is set](https://github.com/mapbox/mapbox-gl-native/blob/ffb02d1b83c28ef32176f92fe1dcea3abbfc5526/src/mbgl/map/map.cpp#L99-L107)
- :x: `MapChangeDidFinishLoading` is fired after the map is fully rendered, in [`Map::renderSync()`](https://github.com/mapbox/mapbox-gl-native/blob/ffb02d1b83c28ef32176f92fe1dcea3abbfc5526/src/mbgl/map/map.cpp#L79), not at the point where data is finished loading
- :x: `MapChangeWillStartRendering` is sent in [`Map::renderSync()`](https://github.com/mapbox/mapbox-gl-native/blob/ffb02d1b83c28ef32176f92fe1dcea3abbfc5526/src/mbgl/map/map.cpp#L58-62) but only initially (it should always come before `DidFinish`, but it does not)
- :white_check_mark: `MapChangeDidFinishRenderingMapFullyRendered` [is sent](https://github.com/mapbox/mapbox-gl-native/blob/54858dae8cb92ba3e111422cdc8188f6d38ab4a9/src/mbgl/map/map.cpp#L76) when rendering is done
  - :x: but does not handle the non-fully-rendered case (#1804)
- :x: `MapChangeDidFailLoadingMap` is not implemented (#2762)

Our initialization flow looks like:

```
2015-10-23 19:42:21.695 start RENDERING
2015-10-23 19:42:24.260 finish RENDERING fullyRendered: YES
2015-10-23 19:42:24.262 finish LOADING
```

[...]

Issue #2762, opened 2015-10-23
Implement MapChangeDidFailLoadingMap event notification


friedbunny:
#630 got us `willStartLoading` and `didFinishLoading`, but [`MapChangeDidFailLoadingMap`](https://github.com/mapbox/mapbox-gl-native/blob/ba4a2965c75c7455f133c78d5c0dcea192e4b874/include/mbgl/map/view.hpp#L22) is somewhat more complicated and remains to be implemented.

This delegate method already exists on [iOS](https://github.com/mapbox/mapbox-gl-native/blob/ba4a2965c75c7455f133c78d5c0dcea192e4b874/platform/ios/MGLMapView.mm#L2983-L2991) and probably does on Android (@ljbade?), it's the core logic that needs doing.

/cc @jfirebaugh

1ec5:
#1804

ljbade:
This failure should also include malformed URLs

ljbade:
And yes Android plumbing is done

friedbunny:
@1ec5 this issue refers to `mapViewDidFailLoadingMap:withError:`, while #1804 is `mapViewDidFinishRenderingMap:fullyRendered:NO`. These both exist in MapKit, though I'm unsure what the effective difference is right now.

bleege:
> And yes Android plumbing is done

@ljbade Could you provide a link to the issue / commit where this was done?  That way we all can track this better in the future.  :+1:

ljbade:
The two relevant PRs are https://github.com/mapbox/mapbox-gl-native/pull/2255 and https://github.com/mapbox/mapbox-gl-native/pull/2591

Although we will need to change the Android Javadoc for that event to indicate it is now implemented.

bleege:
Cool!  Thanks for the links @ljbade!  Makes remembering what was done much easier.  :+1:

tmpsantos:
`MapChangeDidFailLoadingMap` is currently never emitted by `mbgl-core`. I'm proposing on #6142 to emit it at least when the style fails to load.

friedbunny:
#6142 added this to core and #6145 finished up the iOS/macOS implementations.

[...]

PR #3063, opened 2015-11-18
[WIP] Throw/notify on map loading error

1ec5:
Throw an Objective-C exception whenever mbgl encounters an error. (Developers are trained to break on Objective-C exceptions as opposed to C++ exceptions.) If mbgl is unable to set up the map and the delegate implements the map loading failure callback method, call that method instead of throwing the exception.

/ref #2762, #2775
/cc @friedbunny @incanus

### DIFF
--- a/platform/ios/MGLMapView.mm
+++ b/platform/ios/MGLMapView.mm
@@ -42,6 +42,7 @@
 #import <unordered_set>
 
 class MBGLView;
+class MGLLogObserver;
 
 NSString *const MGLDefaultStyleMarkerSymbolName = @"default_marker";
 NSString *const MGLMapboxSetupDocumentationURLDisplayString = @"mapbox.com/guides/first-steps-ios-sdk";
@@ -202,6 +203,9 @@ std::chrono::steady_clock::duration durationInSeconds(float duration)
 - (void)commonInit
 {
     _isTargetingInterfaceBuilder = NSProcessInfo.processInfo.mgl_isInterfaceBuilderDesignablesAgent;
+    
+    std::unique_ptr<MGLLogObserver> logObserver = std::make_unique<MGLLogObserver>(self);
+    mbgl::Log::setObserver(std::move(logObserver));
 
     BOOL background = [UIApplication sharedApplication].applicationState == UIApplicationStateBackground;
     if (!background)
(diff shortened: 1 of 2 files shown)


platform/ios/MGLMapView.mm
```diff
@@ -202,6 +203,9 @@ - (void)setStyleURL:(nullable NSURL *)styleURL
 - (void)commonInit
 {
     _isTargetingInterfaceBuilder = NSProcessInfo.processInfo.mgl_isInterfaceBuilderDesignablesAgent;
+    
+    std::unique_ptr<MGLLogObserver> logObserver = std::make_unique<MGLLogObserver>(self);
+    mbgl::Log::setObserver(std::move(logObserver));
```
1ec5: `mbgl::Log` is a singleton, so this approach won’t scale to multiple `MGLMapView`s.

1ec5:
I considered calling `view.notifyMapChange(mbgl::MapViewDidFailLoadingMap)` directly from within `MapContext::setStyleURL()`, but that method provides no way to specify an error string.

platform/ios/MGLMapView.mm
```diff
@@ -3324,6 +3328,40 @@ void afterRender() override
...
+{
+public:
+    MGLLogObserver(MGLMapView *nativeView_)
+        : nativeView(nativeView_) {}
+    virtual ~MGLLogObserver() {}
+    
+    bool onRecord(mbgl::EventSeverity severity, mbgl::Event event, int64_t code, const std::string &msg)
+    {
+        __block bool didNotifyDelegate = false;
+        if (severity == mbgl::EventSeverity::Error && event == mbgl::Event::Setup)
```
friedbunny: How closely does `mbgl::Event::Setup` correlate to map loading errors (as opposed to networking errors)?

1ec5: Currently, it’s only used inside `MapContext::setStyleURL()`, when the style itself fails to load due to a failed HTTP request. Is that consistent with what you outlined in https://github.com/mapbox/mapbox-gl-native/issues/2775#issuecomment-150829090?

jfirebaugh:
Closing as stale.

PR #3072, opened 2015-11-19
fixes #3069, refs #2922: update display link in all common modes (inc. scrolling)

### LINKED ISSUE #3069: Performance regression from pre-3 to pre-8
Pre-3 fixed the flickering issue for adding/removing annotations on iOS, and the performance was stellar. A matter of milliseconds to make annotation changes and see them in the UI.

Using pre-8 (may have happened earlier, haven't updated since pre-3) annotation changes now take ~1 second to reflect in the UI. There is no flickering or main-thread hanging, that's still all good, but the redraw now takes over a second to complete async.

incanus:


incanus:
Cherry-picked to `master` as a0533d824438392cab504616be3d4afba674ca17.

Issue #898, opened 2015-02-17
default to -fvisibility=hidden


jfirebaugh:
Expose symbols only for the public API.

From @springmeyer:

one gocha is that `-fvisibility=hidden` interacts in interesting ways with  `-flto`. I've gotten mapnik working with both combined but it required some careful usage of `__attribute__ ((visibility ("default")))` and `extern template`

Refs https://github.com/mapbox/mapbox-gl-native/issues/759.

ljbade:
I did not realise you needed this flag Linux/OSX. I know on Windows the default is to only expose specifically marked symbols from shared lib.

mikemorris:
https://gcc.gnu.org/wiki/Visibility has an example of macros with Windows support

mikemorris:
Attempted setting this up in [`898-visibility-hidden`](https://github.com/mapbox/mapbox-gl-native/tree/898-visibility-hidden), seeing the follow two linker errors from Boost when running `make run-osx` that I'm attributing to the cross-lib exceptions issue with hidden symbols, but trying to track this down I can't find anywhere that we're using `boost::throw_exception`?

```
ld: warning: direct access in void boost::throw_exception<boost::bad_function_call>(boost::bad_function_call const&) to global weak symbol typeinfo for boost::exception_detail::clone_impl<boost::exception_detail::error_info_injector<boost::bad_function_call> > means the weak symbol cannot be overridden at runtime. This was likely caused by different translation units being compiled with different visibility settings.
ld: warning: direct access in boost::exception_detail::clone_impl<boost::exception_detail::error_info_injector<boost::bad_function_call> >::rethrow() const to global weak symbol typeinfo for boost::exception_detail::clone_impl<boost::exception_detail::error_info_injector<boost::bad_function_call> > means the weak symbol cannot be overridden at runtime. This was likely caused by different translation units being compiled with different visibility settings.
```

mikemorris:
Uh, my pull request unexpectedly passed CI when I assumed it wouldn't. Anyone want to review and see if it's actually doing what I think it is?

mikemorris:
[...]

PR #3080, opened 2015-11-19
reduce framework binary size with more stripping

incanus:
This needs some more testing, but this shaves off ~20MB on the "stripped" static binary we're producing already with `make ipackage-strip`. We should make sure it doesn't bust anything, but also work on this and other ways to reduce our footprint, especially for folks who don't care about debugging symbols anyway and want to use our framework as if it were closed source. 

Refs https://github.com/mapbox/mapbox-gl-native/issues/898

Also, from @1ec5: 

> There’s a `SEPARATE_STRIP` build setting in Xcode that invokes `strip`.
> 
> It’s off by default.
> 
> Instead, `COPY_PHASE_STRIP` and `STRIP_INSTALLED_PRODUCT` default to on.
> 
> `STRIP_INSTALLED_PRODUCT` only strips the product when you build for archive, something that we should already be doing via `DEPLOYMENT_POSTPROCESSING`.
> 
> Are you producing the framework using package.sh? Is `DEPLOYMENT_POSTPROCESSING` getting set?
> 
> package.sh only sets `DEPLOYMENT_POSTPROCESSING` when building for the device, since that’s where Bitcode bit us.
> 
> Maybe we should set `DEPLOYMENT_POSTPROCESSING` in all cases.

/cc @1ec5 @mikemorris @springmeyer

### DIFF
--- a/scripts/ios/framework.sh
+++ b/scripts/ios/framework.sh
@@ -23,6 +23,7 @@ cp -rv ../../../../ios/framework/Settings.bundle Mapbox.framework/Settings.bundl
 
 # binary
 cp -v ../static/libMapbox.a Mapbox.framework/Mapbox
+strip -Sx Mapbox.framework/Mapbox
 
 # module map
 mkdir Mapbox.framework/Modules


friedbunny:
I'll ask this here (as Slack is down today), mostly for my own edification: 

Why so much stripping — what’s the difference in also using `strip` and why is it separate?

PR #3081, opened 2015-11-19
Token defaults

1ec5:
Tokens in style properties can now fall back on defaults: for example, `{name_en:{name:?}}` produces the value of `name_en`, falling back on `name` if `name_en` is undefined, and finally falling back on the string `?` if neither is defined.

/ref mapbox/mapbox-gl-style-spec#104
/cc @jfirebaugh @ajashton

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## Android master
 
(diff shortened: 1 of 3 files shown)


jfirebaugh:
Per https://github.com/mapbox/mapbox-gl-style-spec/issues/362#issuecomment-161857532 we want to go with a different syntax.

PR #3087, opened 2015-11-20
Set iOS application bundle as asset root

### LINKED ISSUE #1208: iOS: No longer able to load local Sprite resources
I am setting a custom style in my app, as follows:

    NSString *jsonFile = [[NSBundle mainBundle] pathForResource:@"tallygo-v7" ofType:@"json"];
    NSURL *styleURL = [NSURL URLWithString:[NSString stringWithFormat:@"asset://%@", jsonFile]];
    [self.mapView setStyleURL:styleURL];

However, the sprite sheet (in the app bundle) is no longer loading. This URL was working in previous versions by inserting the URL into the raw JSON at runtime, but is now no longer found:

    asset:///Users/picciano1/Library/Developer/CoreSimulator/Devices/7A8973A8-E8B6-4D23-85BE-25B92FBC0067/data/Containers/Bundle/Application/4ECBD40F-A759-4279-AC15-C5D1C20B4316/TallyGo.app/tallygo

The files are located at the root level of the main bundle. I have tried various paths without success:

    asset://tallygo
    asset:///tallygo
    tallygo

I have even tried hard-coding the absolute path above into the style JSON.

Is there any way to do this now?

### LINKED ISSUE #3050: `asset://style.json` loading is broken
It appears to be impossible to load local assets — such as styles — using the `asset://foo.json` URL scheme. This is a result of path problems, which we used to be able to work around using `../../` to get back to the main app bundle.

We've started pushing online styles exclusively (`mapbox://styles/user/hash`), but it should still be possible to load styles locally.

Follows on from #1208, which was specific to sprite loading.

/cc @incanus @1ec5 @jfirebaugh @kkaefer

1ec5:
As of #2746, we no longer bundle any styles with the SDK, so the asset: URL scheme is unused. Instead of pointing asset: into Mapbox.bundle, point it to the application bundle for developer convenience and consistency with the Android and default asset roots. (asset: URLs also work this way for sprite URLs.) Also fixed an issue that prevented relative URLs from being treated as asset: URLs.

Fixes #1208, fixes #3050.

/cc @friedbunny @incanus


friedbunny:
:+1: ++

1ec5:
Cherry-picked to release-ios-v3.0.0 in c4558d48f2ac74b3c2413eb8c363688e430b1177.

incanus:
:+1:

PR #3088, opened 2015-11-20
Discourage code usage of styleURL__ property

1ec5:
A hack atop the hack added in #2632. Pre-declare `styleURL__` with an attribute that prevents it from appearing in code completion suggestions, while leaving it available to Interface Builder, which is unable to parse attributes.

/cc @incanus @friedbunny


friedbunny:
Would it be egregiously terrible form to jam this in v3.0.0? I don't think `.styleURL` is used too often, but heading off any potential confusion gives me a strong sense of inner calm.

1ec5:
Cherry-picked into release-ios-v3.0.0 in 06d400fb9a9a4f612f9fcd8ddd7a68964d01e27e.

PR #3086, opened 2015-11-20
Ensure consistent willChange/didChange callbacks

1ec5:
Coalescing willChange/didChange callbacks means keeping track of how many gestures are currently in progress; by contrast, tracking this state with a boolean opens the door to many race conditions. This change refactors the gesture recognizers, making them more consistent with each other, more consistent in the case where more than one of them has fired, and a whole lot easier to reason about. It also explicitly cancels transitions before all programmatic viewport-modifying methods, since mbgl only does so when animating.

Two behavioral changes result from this PR:

* When the map automatically rotates back to 0° at low zoom levels, that animation gets a separate willChange/didChange pair. Coalescing the two pairs will be possible again once we implement #1581.
* Drift-panning, drift-zooming, etc. result in an didChange callbacks, contrary to MapKit but consistent with MapKit’s documentation.

This PR also updates the tests to rely on didChange callbacks instead of timers. It wasn’t possible to do so before, because didChange would be called immediately after willChange (and before any animation) about half the time. This change should prevent the rotation test cases from flip-flopping going forward.

Fixes #2204, fixes #2313, fixes #2379, fixes #3062. Related to #2775.

/cc @friedbunny @incanus

PR #3090, opened 2015-11-20
Completion handlers for animated MGLMapView methods

### LINKED ISSUE #1581: Animated MGLMapView methods should accept completion handlers

1ec5:
Added transition frame and finish functions to `CameraOptions`. Added equivalent Objective-C blocks as completion handlers on the three main entry points to `easeTo()` in MGLMapView.

Fixes #1581. Depends on #3086.

/cc @incanus @friedbunny

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,6 +23,7 @@ Known issues:
 
 ## iOS master
 
+- `MGLMapView` methods that alter the viewport now accept optional completion handlers. ([#3090](https://github.com/mapbox/mapbox-gl-native/pull/3090))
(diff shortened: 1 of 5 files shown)


1ec5:
This PR is now unblocked.

PR #3096, opened 2015-11-21
Transparent GLKView

### LINKED ISSUE #3038: allow for transparent background on iOS
[GLKView can have transparency](http://stackoverflow.com/questions/10636176/opengl-es-2-0-glkit-with-transparent-background), so we should see how tricky it is for an `MGLMapView` to do this also, if parts of the style (e.g. water) have transparency.

1ec5:
Set the GLKView’s layer to be non-opaque. In rudimentary testing, this appears to have no noticeable performance impact on styles that have opaque background layers (which should be the vast majority of styles). Any performance impact I’m seeing with a transparent background style layer appears to be due to content beneath the map view.

transparent

Fixes #3038.

/cc @incanus

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -46,6 +46,7 @@ Known issues:
 - The user dot’s callout view is now centered above the user dot. It was previously offset slightly to the left. ([#3261](https://github.com/mapbox/mapbox-gl-native/pull/3261))
 - Fixed an issue with small map views not properly fitting annotations within bounds. (#[3407](https://github.com/mapbox/mapbox-gl-native/pull/3407))
 - The map will now snap to north. ([#3403](https://github.com/mapbox/mapbox-gl-native/pull/3403))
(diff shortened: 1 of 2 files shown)


incanus:
Neat. Will test. One thing that jumps out is possible performance implications of a non-opaque backing layer all the time.

incanus:
... which you mentioned :-)

1ec5:
Yes, I haven’t profiled the difference, and in particular I haven’t tried this out on older hardware, but it isn’t nearly as bad as I feared.

1ec5:
If there does turn out to be a significant performance impact, we can also make the transparency optional, by tying it to `-[UIView isOpaque]`.

friedbunny:
Going to rebase locally and try to quantify performance impact, see if I can find anything that would keep us from merging this.

1ec5:
If we honor `-[UIView isOpaque]`, which is `YES` by default, there will be no performance impact in the usual case.

friedbunny:
Found an unrelated, strange rendering issue with translucent water polygons, but this PR seems solid. :+1:

PR #3083, opened 2015-11-21
make annotation tests strict

jfirebaugh:
The annotation tests currently require manual verification. Predictably, they regressed without me noticing (my fault).

We need to:

* Port [pixelmatch](https://github.com/mapbox/pixelmatch) to C++ (`pixelmatch-cpp`?) /cc @mourner 
* Use it to verify expected rendering in the annotation tests (and possibly elsewhere)


jfirebaugh:
Ok, pixelmatch is ported and in use. Tests are passing locally by failing on CI. I think it's something to do with PNG color profiles?

Here's a sample failure.

Expected:
image

Actual:
image

Diff:
image

When I run the JS version of pixelmatch locally on those files, it reports they are identical.

@mourner Have you encountered this?

incanus:
Color profiles seems likely.

mourner:
> Ok, pixelmatch is ported and in use. Tests are passing locally by failing on CI. I think it's something to do with PNG color profiles?

@jfirebaugh what was it? Haven't encountered anything similar on JS side.

jfirebaugh:
It was undesired premultiplication behavior of libpng when a color profile is present in the source PNG, plus confusion in our code about where image data is premultiplied and where is is not. To fix the first, I disabled all premultiplication in libpng and manually premultiply using an algorithm that is reversed engineered to produce the same results as Core Graphics (premultiplication algorithms are surprisingly subtle). To fix the second, I rewrote most imaging code to work with explicit `PremultipliedImage` and `UnassociatedImage` (non-premultiplied) types. We should be using `PremultipliedImage` more or less everywhere, the exception being at the very edge of reading/writing PNGs, which are always unpremultiplied.

PR #3106, opened 2015-11-24
Don't double-tap zoom when quickzooming on Android

### LINKED ISSUE #3104: [Android] Double-tap zoom should not be called when quickzooming
When trying to quickzoom out (double-tap and drag up), the double-tap gesture is detected and the map immediately zooms in. This makes it nearly impossibly to quickzoom out.

brewin:
Fixes #3104

Also changed to zoom in on center of map when quickzooming, which is how Google Maps does it.


tobrun:
Taking a look at this, was able to reproduce the original issue

tobrun:
The issue seems to gone so great job @brewin 
Also great catch about the center of map zooming. 
Will merge it now.

FWIW, while testing I noticed that OnMapLongClickListener is being called, but this is also the case for current state of master. Will create a new ticket for that.

brewin:
Cool. I found a couple other problems with quickzoom too. I'll open new issues for them.

Issue #3110, opened 2015-11-24
[iOS] showAnnotations not working with two points


YBthebest:
When I add multiple annotations using dispatch_asynch, all the annotations are correctly added to the Map, and the basic zoom level shows the whole world map.

![simulator screen shot 24 nov 2015 17 03 38](https://cloud.githubusercontent.com/assets/1565896/11372393/8388f2c8-92cd-11e5-94f1-f838c33e3058.png)

Yet when I try to use:
```[mapBox showAnnotations:mapBox.annotations animated:YES];```

It zooms in on a street that IS in the same CITY as all the annotations, but it still clearly does not accomplish what it should.
![simulator screen shot 24 nov 2015 17 04 20](https://cloud.githubusercontent.com/assets/1565896/11372404/976b8d64-92cd-11e5-92db-c2d6c2076c3b.png)

**Comment by 1ec5** [comment]
In #3111, I modified iosapp to similarly add annotations using `dispatch_async()` then show annotations on the main thread. Are there any differences in how you’re adding and showing the annotations?

**Comment by 1ec5** [comment]
Correction: in that PR, iosapp is still adding the annotations on the main thread. But if you move that line into the surrounding block, it still zooms in to the right area.

**Comment by incanus** [comment]
Yeah, I would ensure that you are calling UI code on the main thread/queue, as well as possibly post your setup code here @YBthebest if you are still seeing problems.

**Comment by YBthebest** [comment]
```
    [mapBox removeAnnotations:[mapBox annotations]];
    dispatch_group_t group = dispatch_group_create();
    for(int i = 0;i<[tableNoms count];i++){
        dispatch_group_enter(group);
        MGLPointAnnotation *point = [[MGLPointAnnotation alloc] init];
        point.coordinate = coordinate;
        point.title = tableNoms[i];
        [mapBox addAnnotation:point];
        dispatch_group_leave(group);
    }
    while (dispatch_group_wait(group, DISPATCH_TIME_NOW)) {
        [[NSRunLoop currentRunLoop] runMode:NSDefaultRunLoopMode
                                 beforeDate:[NSDate dateWithTimeIntervalSinceNow:1.f]];
    }
    [mapBox showAnnotations:mapBox.annotations animated:YES];
    [self.myTableView reloadData];
```

This is my code. I have deleted some of the code that isn't important here.
What is your idea? Where should I put the showAnnotations line to be sure that it is called correctly after all the annotations are added?

Thanks

[...]

PR #3111, opened 2015-11-24
Demonstrate show annotations in iosapp

1ec5:
ref #3110


incanus:
There exists the ability to have long press-dropped annotations on the map already, so `showAnnotations` could move beyond the expected area. We should probably `showAnnotations:annotations` instead using the annotations in scope that we just added.

incanus:
Oh, never mind — we call `removeAnnotations` first before adding these new demo ones.

1ec5:
That’s a fine optimization: 2eb39e93de1d04503c1c8e25e47d049306184cc4.

PR #3114, opened 2015-11-24
fixes #3113: bring in map view in absence of -ObjC linker flag for IB

### LINKED ISSUE #3113: Interface Builder doesn't recognize classes without -ObjC linker flag
Discovered by @friedbunny and applies to any manual binary install method. You can't use `MGLMapView` from IB without referencing it elsewhere in code. 

Should be able to fix this with a combo of `+load` and calling `MGLMapView` from our category loader from #2967. 

/cc @friedbunny @1ec5

incanus:
@friedbunny @1ec5

### DIFF
--- a/platform/ios/MGLCategoryLoader.m
+++ b/platform/ios/MGLCategoryLoader.m
@@ -4,15 +4,21 @@
 #import "NSProcessInfo+MGLAdditions.h"
 #import "NSString+MGLAdditions.h"
 
-// https://github.com/mapbox/mapbox-gl-native/issues/2966
+#import "MGLMapView.h"
 
 @implementation MGLCategoryLoader
 
 + (void)loadCategories
 {
+    // https://github.com/mapbox/mapbox-gl-native/issues/2966
+    //
     mgl_linkBundleCategory();
     mgl_linkProcessCategory();
     mgl_linkStringCategory();
+
+    // https://github.com/mapbox/mapbox-gl-native/issues/3113
+    //
+    NSString *description = [MGLMapView description];
 }
 
 @end
--- a/platform/ios/MGLAccountManager.m
+++ b/platform/ios/MGLAccountManager.m
@@ -18,6 +18,9 @@
 #pragma mark - Internal
 
 + (void)load {
+    // Load all referenced categories due to absence of -ObjC linker flag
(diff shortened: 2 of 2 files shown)


platform/ios/MGLCategoryLoader.m
```diff
@@ -4,7 +4,16 @@
...
 #import "NSString+MGLAdditions.h"
 
+#import "MGLMapView.h"
+
 // https://github.com/mapbox/mapbox-gl-native/issues/2966
+// https://github.com/mapbox/mapbox-gl-native/issues/3113
+
+@interface MGLMapView (MGLCategoryLoader)
+
++ (void)mgl_linkMapView;
```
1ec5: It should be possible to just call an NSObject class method. For example, `[MGLMapView description];`.

incanus: Oh, good call. I was wondering about optimization there, and originally went with `+[UIView requiresConstraintBasedLayout]`, but something from `NSObject` would be safer.

1ec5: Come to think of it, ObjC dynamic message dispatch would probably prevent any optimization around whatever class method you’d call here, but I feel better about a method that has to return a string than one that really is a no-op.

friedbunny:
Still seeing `Unknown class MGLMapView in Interface Builder file.` when trying to use this framework with an IB-only project.

incanus:
Hmm, can you set a symbolic breakpoint on `+load` or maybe `+[MGLAccountManager load]`? Since you rolled the binary locally, you should still get symbols.

friedbunny:
Symbolic breakpoint for `+[MGLAccountManager load]` never trips.

incanus:
Hmm, and it does with a code-based path? I'm not as familiar with `+load` as @1ec5, but maybe `awakeFromNib` is more appropriate here?

incanus:
Eh, `awakeFromNib` would be a :chicken: & :white_circle: problem here, since nothing would be in the NIB/XIB to awake. Hmm.

incanus:
We don't yet have a fix here.

incanus:
While this doesn't solve the IB problem, it's still better code & docs and should go into 3.0.1. We're bringing back the `-ObjC` flag, anyway.

PR #3117, opened 2015-11-24
Attach compass gesture recognizer to compass view

### LINKED ISSUE #3099: [iOS] Compass does not receive touch events if its frame changes
If the user of the iOS framework changes the frame of the compass view (e.g. if part of the map is covered by UI elements) , the end user cannot tap on the compass to reset the north position.
The touch zone is still the default compass view position.

expected : update touch zone when changing compass view frame.

1ec5:
Fixes #3099.

/cc @incanus

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,7 +23,7 @@ Known issues:
 
 ## iOS master
 
-- Fixed an issue preventing the compass from responding to taps after the compass is moved programmatically. ([#3116](https://github.com/mapbox/mapbox-gl-native/pull/3116))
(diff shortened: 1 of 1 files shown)


1ec5:
I accidentally pushed f782f5eeb793bc4489a906969ab4774a2befbbb8 directly to master instead of on this branch, but consider that part of this PR.

incanus:
:+1: 

Dang, it bugs me where I can't find the version of this we had before (can't remember if GL or raster) but this was the fix there, too. 

Let's put this on 3.0.1 (bug fixes) since we already have at least one other ticket there.

1ec5:
Cherry-picked into release-ios-v3.0.1 as 8ba6f80c1b0d02f09ab787af7b145345b0f0244d dbe68760ec32938fcab408082345eadfbcf88b9b.

PR #3124, opened 2015-11-25
Fix GCC5 complaining about -Wunused-variable.

vosst:
GCC5 and later complains about unused variables in the files altered by this PR. Selectively disabling the warning solves the issue and leaves the forcing function of the variables untouched.

### DIFF
--- a/platform/default/image.cpp
+++ b/platform/default/image.cpp
@@ -11,6 +11,8 @@
 
 #include <mbgl/platform/default/image_reader.hpp>
 
+#pragma GCC diagnostic push
+#pragma GCC diagnostic ignored "-Wunused-variable"
 // Check png library version.
 const static bool png_version_check = []() {
     const png_uint_32 version = png_access_version_number();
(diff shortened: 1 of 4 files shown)


mikemorris:
It looks like the warning in GCC5 is correct, as we're not using these as variables anywhere else - is there a way to write them an anonymous lambdas instead? If not, this fix looks fine to me.

vosst:
I was thinking about that, too. The variable is required to force execution evaluation of the lambda, though. I also checked if asserting on the variable in at least one function would work. Unfortunately, as assert is a macro, the use of the variable is optimized out in release builds.

vosst:
Speaking of anonymous: I would propose to place the variables in an anonymous namespace to prevent from polluting the global linker namespace. Happy to adjust this change if you are fine with it.

jfirebaugh:
Declaring them as `static` has the same effect -- their visibility is limited to the translation unit.

PR #3100, opened 2015-11-25
Improve map debug options

brunoabinader:
Besides tile identification and status, I believe it would be a nice feat to have the associated tile _timestamp_ information.

By _timestamp_ I refer to the time point used together with expiry date to calculate if the cached tile data is still valid.

/cc @mapbox/gl

### DIFF
--- a/platform/default/sqlite_cache.cpp
+++ b/platform/default/sqlite_cache.cpp
@@ -118,9 +118,9 @@ void SQLiteCache::Impl::get(const Resource &resource, Callback callback) {
                 // Status codes > 1 indicate an error
                 response->error = std::make_unique<Response::Error>(Response::Error::Reason(status));
             }
-            response->modified = getStmt->get<int64_t>(1);
+            response->modified = Seconds(getStmt->get<Seconds::rep>(1));
             response->etag = getStmt->get<std::string>(2);
-            response->expires = getStmt->get<int64_t>(3);
+            response->expires = Seconds(getStmt->get<Seconds::rep>(3));
             response->data = std::make_shared<std::string>(std::move(getStmt->get<std::string>(4)));
             if (getStmt->get<int>(5)) { // == compressed
                 response->data = std::make_shared<std::string>(std::move(util::decompress(*response->data)));
@@ -177,9 +177,9 @@ void SQLiteCache::Impl::put(const Resource& resource, std::shared_ptr<const Resp
             putStmt->bind(2 /* status */, 1 /* success */);
         }
         putStmt->bind(3 /* kind */, int(resource.kind));
-        putStmt->bind(4 /* modified */, response->modified);
+        putStmt->bind(4 /* modified */, response->modified.count());
         putStmt->bind(5 /* etag */, response->etag.c_str());
-        putStmt->bind(6 /* expires */, response->expires);
+        putStmt->bind(6 /* expires */, response->expires.count());
 
         std::string data;
         if (resource.kind != Resource::SpriteImage && response->data) {
@@ -208,7 +208,7 @@ void SQLiteCache::Impl::put(const Resource& resource, std::shared_ptr<const Resp
     }
 }
 
-void SQLiteCache::Impl::refresh(const Resource& resource, int64_t expires) {
+void SQLiteCache::Impl::refresh(const Resource& resource, Seconds expires) {
     try {
         if (!db) {
             createDatabase();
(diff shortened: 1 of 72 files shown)


brunoabinader:
Example output:
screenshot from 2015-11-25 16-06-34

Please beware that you'll need to clear your current SQLite cache file, as these changes modifies the type for both `modified` and `expires` variables.

src/mbgl/renderer/debug_bucket.cpp
```diff
@@ -9,9 +10,20 @@
...
-DebugBucket::DebugBucket(const TileID id, const TileData::State state_) : state(state_) {
+DebugBucket::DebugBucket(const TileID id, const TileData::State state_, time_t modified_, time_t expires_)
+    : state(state_),
+      modified(modified_),
+      expires(expires_) {
     const std::string text = std::string(id) + " - " + TileData::StateToString(state);
     fontBuffer.addText(text.c_str(), 50, 200, 5);
+
+    if (modified > 0 && expires > 0) {
+        const std::string modifiedText = "modified: " + util::rfc1123(modified);
```
kkaefer: Can we find a shorter representation of the date, e.g. "Y-m-d H:i:s" and imply GMT?

platform/default/sqlite_cache.cpp
```diff
@@ -118,9 +118,9 @@ void SQLiteCache::Impl::get(const Resource &resource, Callback callback) {
                 // Status codes > 1 indicate an error
                 response->error = std::make_unique<Response::Error>(Response::Error::Reason(status));
             }
-            response->modified = getStmt->get<int64_t>(1);
+            response->modified = getStmt->get<time_t>(1);
```
kkaefer: I don't think we should change away from `int64_t`. `time_t` is a platform-specific typedef and is a 32 bit value on some systems, which will make us susceptible to the year 2038 problem. If it's defined to something else entirely, compilation may fail because `.get<time_t>()` may not be instantiated.

tmpsantos:
Neat :+1:

platform/default/sqlite_cache.cpp
tmpsantos: `std::chrone::time_point` ?

friedbunny:
Cool, this should be useful.

incanus:
As on GL JS, we should probably look into splitting this out separately from tile borders/labels and collision boxes. Things are getting crowded and turning one on means turning all on, which sometimes gets in the way.

tmpsantos:
> As on GL JS, we should probably look into splitting this out separately from tile borders/labels and collision boxes. Things are getting crowded and turning one on means turning all on, which sometimes gets in the way.

Maybe we should have something like `::enableDebug(DebugOptions)`

brunoabinader:
> Maybe we should have something like ::enableDebug(DebugOptions)

@incanus @tmpsantos good idea. I've added `MapDebugMode` enum in the last patch to specialize the debug information shown on each tile. `Map::toggleDebug()` now increases the amount of debug information on screen until it goes back to none. I've kept current behavior i.e. `MapDebugMode::TileBorders | MapDebugMode::ParseStatus` when debug is set for each port, but ports can now specialize this if they want/need.

The example below better illustrates the new behavior for `Map::toggleDebug()`:
test

tmpsantos:
> I've added MapDebugMode enum in the last patch to specialize the debug information shown on each tile. Map::toggleDebug() now increases the amount of debug information on screen until it goes back to none.

:heart:

brunoabinader:
In https://github.com/mapbox/mapbox-gl-native/commit/3acfa72f0c0a22b757eed516dc44a7cb70468bc8 text collision debug is now a flag `MapDebugMode::Collision`. I've noticed there is currently a bug in collision debug implementation that causes it to not properly repaint upon toggling - this will be handled in #3140.

src/mbgl/map/map_data.hpp
```diff
@@ -49,24 +50,25 @@ class MapData {
...
-    }
-    inline void setDebug(bool value) {
-        debug = value;
+    inline MapDebugMode getDebug() const {
+        return debugMode;
     }
 
-    inline bool getCollisionDebug() const {
-        return collisionDebug;
+    inline void toggleDebug() {
```
kkaefer: Can we rename `toggleDebug` to reflect what it actually does? It's not really toggling rather than cycling through the options.

brunoabinader: Good point @kkaefer :+1: In that sense `MapDebugMode` should be also called `MapDebugOptions` instead. I'll update the code accordingly.

PR #3092, opened 2015-11-25
Offset lines

kkaefer:
We should support offsetting lines as a paint property. The difference to `line-translate` is that it offsets the overall line, while this feature is about an offset perpendicular to the line.

### DIFF
--- a/src/mbgl/shader/line.vertex.glsl
+++ b/src/mbgl/shader/line.vertex.glsl
@@ -14,6 +14,7 @@ uniform mat4 u_matrix;
 // shared
 uniform float u_ratio;
 uniform vec2 u_linewidth;
+uniform float u_offset;
 
 uniform float u_extra;
 uniform mat2 u_antialiasingmatrix;
@@ -23,6 +24,7 @@ varying float v_gamma_scale;
 
 void main() {
     vec2 a_extrude = a_data.xy;
+    float a_direction = sign(a_data.z) * mod(a_data.z, 2.0);
 
     // We store the texture normals in the most insignificant bit
     // transform y so that 0 => -1 and 1 => 1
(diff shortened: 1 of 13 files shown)


kkaefer:
This patch implements rudimentary support for offsetting lines.

It doesn't properly offset line joins that are very sharp, and generally only works okay-ish for miter joins (bevel or round joins start to look weird at a certain distance).

This is missing:

- [ ] Tests
- [ ] Documentation in the style spec?

lucaswoj:
> Documentation in the style spec?

Is there any reason we wouldn't want to document this feature? 

If not, I'll ticket out the style spec and GL JS work.

kkaefer:
I looked at the failing rendering tests and they amount to this difference:

| Old | New | 
| --- | --- |
| out-18-old | out-18-new |

Both lines look fine when the lines doesn't overlap internally. There's no version of that join that looks more correct (both are broken due to our in-shader tessellation), so I'd update the test suite once we have a patch for JS.

src/mbgl/shader/line.vertex.glsl
```diff
@@ -36,11 +38,19 @@ void main() {
...
+    float u = 0.5 * a_direction;
+    float t = 1.0 - abs(u);
+    vec2 offset = u_offset * a_extrude * scale * normal.y * mat2(t, -u, u, t);
+
     // Remove the texture normal bit of the position before scaling it with the
     // model/view matrix. Add the extrusion vector *after* the model/view matrix
     // because we're extruding the line in pixel space, regardless of the current
     // tile's zoom level.
-    gl_Position = u_matrix * vec4(floor(a_pos * 0.5) + dist / u_ratio, 0.0, 1.0);
+    gl_Position = u_matrix * vec4(floor(a_pos * 0.5) + offset / u_ratio + dist / u_ratio, 0.0, 1.0);
```
lucaswoj: Is there a reason this isn't written as `(offset + dist) / u_ratio`?

kkaefer: Nope, that works too, and is probably marginally faster.

kkaefer: Fixed

Issue #3132, opened 2015-11-25
[Android] Double-taps are sometimes mistaken for quickzooms


brewin:
With ad0b37f I've noticed double-tap zooming is a little more difficult than it is on Google Maps. I think Google Maps measures the distance between double-tap ACTION_DOWN and ACTION_UP. If it's under a certain threshold (maybe getScaledTouchSlop) then it doesn't count as a quickzoom. Right now, if your finger strays a few pixels after ACTION_DOWN, it's considered a quickzoom even though it was probably just a sloppy double-tap.

tobrun:
Closed with https://github.com/mapbox/mapbox-gl-native/commit/60b0e78e0ef7357cf6ef260a6df4241b2c5503db

PR #3136, opened 2015-11-26
Improve quickzooming on Android

### LINKED ISSUE #3131: Quickzooming on Android causes onLongMapClickListener to be called
Quickzooming is a feature that allows the user to zoom the map with just one hand.
You can activate quickzooming by double tapping and holding you finger on the screen after second tap.
Moving your finger vertically will now scale the map.

While testing the PR from #3106, I noticed that onLongMapClickListener is called when activating QuickZooming. This should not be the case. 

Example see [here](https://www.youtube.com/watch?v=XxaRQVFqZQE&feature=youtu.be).

brewin:
Fixes #3131 and #3132

### DIFF
--- a/android/MapboxGLAndroidSDK/src/main/java/com/mapbox/mapboxsdk/views/MapView.java
+++ b/android/MapboxGLAndroidSDK/src/main/java/com/mapbox/mapboxsdk/views/MapView.java
@@ -2377,13 +2377,12 @@ public final class MapView extends FrameLayout {
 
             switch (e.getAction()) {
                 case MotionEvent.ACTION_DOWN:
-                    mQuickZoom = false;
                     break;
                 case MotionEvent.ACTION_MOVE:
-                    mQuickZoom = true;
                     break;
                 case MotionEvent.ACTION_UP:
                     if (mQuickZoom) {
+                        mQuickZoom = false;
                         break;
                     }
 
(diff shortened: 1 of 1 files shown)


tobrun:
@brewin just tried this out and this fixes #3131, awesome stuff!
As for #3132, I'm not seeing how this is being resolved. 
Is this resolved without introducing that construction you mentioned in the issue itself?

brewin:
I found that the fix I described was basically already implemented in ScaleGestureListener. The problem was that with mQuickZoom set to true in double-tap ACTION_MOVE, double-tap zooming would never occur even when "sloppy" taps were detected in ScaleGestureListener. Setting mQuickZoom to true in ScaleGestureListener fixes that. So now double-taps with very slight movement of the finger before ACTION_UP actually zoom in on the map instead of just being ignored. This behavior seems to be similar to how Google Maps works.

```java
            // If scale is large enough ignore a tap
            mScaleFactor *= detector.getScaleFactor();
            if ((mScaleFactor > 1.05f) || (mScaleFactor < 0.95f)) {
                mZoomStarted = true;
            }
            
            // Ignore short touches in case it is a tap
            // Also ignore small scales
            long time = detector.getEventTime();
            long interval = time - mBeginTime;
            if (!mZoomStarted && (interval <= ViewConfiguration.getTapTimeout())) {
                return false;
            }

            if (!mZoomStarted) {
                return false;
            }

            // Cancel any animation
            mNativeMapView.cancelTransitions();

            // Gesture is a quickzoom if there aren't two fingers
            mQuickZoom = !mTwoTap;
```

tobrun:
Merged with https://github.com/mapbox/mapbox-gl-native/commit/60b0e78e0ef7357cf6ef260a6df4241b2c5503db

@brewin Awesome work! thank you for contributing

PR #3137, opened 2015-11-26
Disable implicit location updating in background

maciekish:
If the `location` mode is added to `UIBackgroundModes` in an apps `Info.plist` the blue "Appname is using your location" bar never goes away after pressing the home button on iOS 8.1. This happens even if you set `showsUserLocation = NO` because it's not the `MGLMapView` that causes it, but `MGLMapBoxEvents`.

This fixes it by checking if the app is backgrounded in `MGLMapboxEvents.m`


friedbunny:
Thanks for the pull request, we hit this in #3671.

PR #3139, opened 2015-11-26
[core] Abstract libuv

tmpsantos:
This branch reduces the delta between master and the Qt branch that is under development. It enables the platform to reimplement the main loop and get rid of libuv in favor of, say, glib or the Qt main loop.

Wrapping is mostly done by reusing existing `util::` like `util::Thread` and also by 3 new classes:
 - `AsyncTask`: Send asynchronous messages to the main loop where the `AsyncTask` object lives, coalescing multiple messages.
 - `Timer`: an ordinary timer.
 - `RunLoop`: not exactly new but was refactored to allow reimplementation. Consists of a platform main loop, using `AsyncTask` internally for waking up.

All these new classes were implemented in a way to be almost a drop-in replacement for the libuv counterparts, making the changes as mild as possible. `AssetFS` and `AssetZIP` are exceptions, as they were heavily refactored to use `util::Thread` for doing threaded file I/O instead of libuv workers.

For all the newly introduced classes, unit tests were also added and in some cases, like for the `AssetZIP`, broken unit tests were fixed and more were added.

This PR also is also a enabler for #2909.


tmpsantos:
:eye: @kkaefer @brunoabinader @jfirebaugh

tmpsantos:
:eyes: @kkaefer @brunoabinader @jfirebaugh

linux/mapboxgl-app.gypi
```diff
@@ -39,6 +39,10 @@
         ],
       },
 
+      'include_dirs': [
+        '../src',
```
jfirebaugh: What's this necessary for? IIRC, we want to use the apps partly as as examples, which means they should avoid using internal includes.

include/mbgl/util/run_loop.hpp
```diff
@@ -15,21 +15,36 @@
...
+    };
 
-    static uv_loop_t* getLoop() {
-        return current.get()->get();
-    }
+    RunLoop(Type type = Type::Default);
+    ~RunLoop();
 
+    static RunLoop* Get();
+    static LOOP_HANDLE getLoopHandle();
```
jfirebaugh: Is there any way to make this an implementation-dependent typed pointer? `void *` and then casting back to a known type where used bypasses the type checker and will crash/corrupt memory if somehow the expected implementation is not the one actually in use.

src/mbgl/util/thread_context.cpp
```diff
@@ -1,30 +1,95 @@
 #include <mbgl/util/thread_context.hpp>
 
+#include <mbgl/util/thread_local.hpp>
+
+#include <cassert>
+
+namespace {
+
+using namespace mbgl::util;
+static tls<ThreadContext>& current = *new tls<ThreadContext>;
```
jfirebaugh: Any particular reason to `new` this and then dereference it, as opposed to using a static value? The current implementation is technically a memory leak.

linux/mapboxgl-app.gypi
tmpsantos: This one slipped. Removing.

src/mbgl/util/thread_local.hpp
```diff
@@ -0,0 +1,59 @@
...
+
+#include <stdexcept>
+
+#include <pthread.h>
+
+namespace mbgl {
+namespace util {
+
+template <class T>
+class tls : public noncopyable {
```
jfirebaugh: Let's avoid the TLA and call this `ThreadLocal`.

include/mbgl/util/run_loop.hpp
tmpsantos: I really don't see how. But IMO the risk is very low as it is only used in the `AsyncTask` and `Timer`.

jfirebaugh: Yeah, I agree, I'm not too worried. I'll see if anything springs to mind after all the refactorings here shake out.

src/mbgl/util/thread_context.cpp
tmpsantos: Yes and I did the same with `RunLoop`. I was hitting sometimes a bug that would sometimes the `static tls<RunLoop> current` before a `RunLoop` object on the heap. The object on the heap sets the `::current` to `nullptr` and crashes.

That reminded me that a practice we had at the Chromium browser to "leak" static variables to prevent this sort of issues: https://code.google.com/p/chromium/codesearch#chromium/src/base/macros.h&l=153

It is not exactly a leak because it is still referenced (so valgrind won't complain) and it would only be destructed at exit anyway.

src/mbgl/util/thread_local.hpp
tmpsantos: :+1:

PR #3142, opened 2015-11-27
[android] Remove autogenerated file from tree

tmpsantos:
developer-config.xml gets generated on build time with the token set for the env var MAPBOX_ACCESS_TOKEN. Because the file was checked to the git repo, it was not being regenerated and thus, crashing the test app accusing no token set.


tmpsantos:
cc @brunoabinader @bleege @tobrun

Issue #3115, opened 2015-11-24
Allowing multiple InfoWindows to be visible at a time


zugaldia:
At the moment, only one InfoWindow can be visible at a time. We want to be able to show multiple at once.

incanus:
To analog (verb?) the iOS side, the `selectAnnotations:` API there allows for multiple annotations to be passed to selection but, like Apple Maps, to only set `selected` state on the first one (see [here](https://www.mapbox.com/ios-sdk/api/3.0.0/Classes/MGLMapView.html#//api/name/selectedAnnotations)). 

I think it's best on iOS to probably stick to that, but since on Android we have some needs for multiple popups, we can maybe use that API as a model, albeit with slightly different behavior.

zugaldia:
In that case, and to mimic `selectedAnnotations` we could just move from a `private Marker mSelectedMarker;` to a `private List<Marker> mSelectedMarkers  = new ArrayList<>();` and change the internal `MapView` logic accordingly. API-wise, `public Marker getSelectedMarker()` will become `public List<Marker> getSelectedMarkers()`.

I'll try that approach in https://github.com/mapbox/mapbox-gl-native/tree/3115-multiple-infowindows

zugaldia:
Good progress on both keeping the InfoWindow open and also [allowing an InfoWindow to be open while panning](https://github.com/mapbox/mapbox-gl-native/issues/3121).

* It works for default InfoWindows: https://www.dropbox.com/s/yj6kzmmjvichkp4/movie1.mp4?dl=0
* As well as for custom InfoWindows: https://www.dropbox.com/s/kthg70njitwqm2l/movie2.mp4?dl=0

Before having a PR ready we need to do some small refactoring. [As discussed](https://github.com/mapbox/mapbox-gl-native/issues/3116), we don't see a performance hit  for such a small number of InfoWindows.

bleege:
@zugaldia @tobrun FWIW, let's leave the "default" mode to be just 1 `InfoWindow` open at a time per MapView as this is likely to be the majority of desired use cases for developers.  This means that we'll want to have a flag of some sort on the `MapView` to enable multiple visible `InfoWindow` support.  Something like `setAllowConcurrentMultipleOpenInfoWindows(boolean allow)`.  Thoughts?

tobrun:
@bleege seems logical, I consider having multiple open to be a special kind of requirement.
@zugaldia can you take this up? currently not on the project.

[...]

PR #3152, opened 2015-11-30
Added sample InfoWindowConcurrentActivity

zugaldia:
To illustrate the behavior of multiple InfoWindows visible at a time (https://github.com/mapbox/mapbox-gl-native/issues/3115).

### DIFF
--- a/android/MapboxGLAndroidSDKTestApp/src/main/res/layout/activity_infowindow_concurrent.xml
+++ b/android/MapboxGLAndroidSDKTestApp/src/main/res/layout/activity_infowindow_concurrent.xml
@@ -1,0 +1,24 @@
+<?xml version="1.0" encoding="utf-8"?>
+<LinearLayout xmlns:android="http://schemas.android.com/apk/res/android"
+    xmlns:app="http://schemas.android.com/apk/res-auto"
+    android:layout_width="match_parent"
+    android:layout_height="match_parent"
+    android:orientation="vertical">
+
+    <android.support.v7.widget.Toolbar
+        android:id="@+id/toolbar"
+        android:layout_width="match_parent"
+        android:layout_height="?attr/actionBarSize"
+        android:background="@color/primary"
+        android:theme="@style/ThemeOverlay.AppCompat.Dark.ActionBar" />
+
+    <com.mapbox.mapboxsdk.views.MapView
+        android:id="@+id/infoWindowConcurrentMapView"
+        android:layout_width="match_parent"
+        android:layout_height="match_parent"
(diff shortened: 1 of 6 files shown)


incanus:
This as well as the Info Window tracking to pans & zooms is :ok_hand: :100:

zugaldia:
@incanus right. both features ended up being implemented at pretty much the same time.

incanus:
Looking good! 

anigif-1448918357

One weird manifestation of this odd UI pattern is the final frame here: 

screen shot 2015-11-30 at 1 20 18 pm

Layering was done by which was above the other pre-rotation, but later rotations can throw this off.

zugaldia:
cc: @tobrun

tobrun:
@zugaldia 
Not really sure what best UX would be in this case.. hence this is an `odd ui pattern`.

In general: 
The parent layout is a FrameLayout so it will stack up its children in the way they were added to the ViewGroup. We could play around with this by using addView with an index or replace FrameLayout with a Relative so we can use bringToFront.

zugaldia:
I have added the layering issue @incanus has found to the original ticket (https://github.com/mapbox/mapbox-gl-native/issues/3115#issuecomment-161065419) but this doesn't affect this PR, as it's just the test app and the basic functionality is there.

Also, the CI fail above is unrelated to Android, so if it all looks good otherwise, we're ready to merge the branch.

/cc: @bleege

incanus:
> hence this is an `odd ui pattern`

android/MapboxGLAndroidSDKTestApp/src/main/res/layout/activity_infowindow_concurrent.xml
```diff
@@ -0,0 +1,24 @@
...
+
+    <android.support.v7.widget.Toolbar
+        android:id="@+id/toolbar"
+        android:layout_width="match_parent"
+        android:layout_height="?attr/actionBarSize"
+        android:background="@color/primary"
+        android:theme="@style/ThemeOverlay.AppCompat.Dark.ActionBar" />
+
+    <com.mapbox.mapboxsdk.views.MapView
+        android:id="@+id/mapView"
```
bleege: @zugaldia Let's change the id to something like `@+id/infoWindowConcurrentMapView`.  While there shouldn't be any runtime issues due to namespacing, I unfortunately got us into the habit of using `@+id/mapView` in all the TestApp examples which makes immediate sight recognition more difficult than it should be.

zugaldia:
@bleege sounds good, I'll include that before merging.

Issue #3112, opened 2015-11-24
Crash on launch if Core Telephony is linked — -[CTTelephonyNetworkInfo init]


skajake:
I added the SDK using Cocoapods.  I added the access key to the plist.  I added the Settings bundle for metrics opt out.

Upon running the application I get the following stack trace:

![Stack Trace](http://i.imgur.com/obpLjQ7.png)

incanus:
What's the exception thrown?

skajake:
No exception, just a bad access:

![Bad Access](http://i.imgur.com/tmvubei.jpg)

incanus:
Hmm. This functionality should be fine @skajake, so could you email help@mapbox.com since it's probably specific to your token string and we'll go from there?

1ec5:
What device is this? MGLMapboxEventsData attempts to get the network connection type using Core Telephony, but we’ve had problems with this on some devices in the past: #2687.

incanus:
Sorry about that @skajake — I may have jumped to a conclusion. Are you using the 3.0.0 version we released to CocoaPods yesterday? It fixes https://github.com/mapbox/mapbox-gl-native/issues/2687.

skajake:
This is happening in an Ipad Air 2 simulator.  Also happens on a hardware Ipad Air device.

Cocoapods pulled in version 3.0.0  hash:

   Mapbox-iOS-SDK: 5e7bafc554a15ebfc615d8b0fa0a0b5d31faac70

friedbunny:
I was able to replicate this crash with the iPad Air 2 simulator by manually linking CoreTelephony in a new v3.0.0 CocoaPods project. I'll look in to this, but, in the meantime, this can be avoided by simply not linking CoreTelephony.

```
thread #1: tid = 0x142f00, 0x0000000111de8fc9 CoreFoundation`CFStringCompare + 25, queue = 'com.apple.main-thread', stop reason = EXC_BAD_ACCESS (code=1, address=0x0)
  * frame #0: 0x0000000111de8fc9 CoreFoundation`CFStringCompare + 25
    frame #1: 0x000000010f437329 CoreTelephony`sCTEventForNotificationString(__CFString const*) + 77
    frame #2: 0x000000010f46ea60 CoreTelephony`_CTServerConnectionRegisterForNotification + 22
    frame #3: 0x000000010f456f8b CoreTelephony`-[CTTelephonyNetworkInfo setUpServerConnection] + 215
[...]

PR #3168, opened 2015-12-02
refs #3112: cleanups around Core Telephony runtime

incanus:
This doesn't solve our problems directly, but does make things clearer (I think) and avoid doing anything with CT on the simulator anyway, as I think that can be unpredictable. Basically I was seeing crashes from the root of #3112 on device, and I think on simulator because of weird behavior of CT there (so, not related directly).

/cc @friedbunny @1ec5


1ec5:
:+1:

incanus:
Also I'm avoiding the reuse of `CTTelephonyNetworkInfo` in our own namespace, just for safety.

Issue #2805, opened 2015-10-26
Android Camera API (aka Tilt)


bleege:
The [Core GL work to support perspective](https://github.com/mapbox/mapbox-gl-native/pull/2116) was done by @ansis at the end of August.  @1ec5 added iOS bindings for it #2159 and #2193.  Now that the initial version of the Android SDK (v 2.1.0) has been released (as of 21-Oct-2015) we should add support in Android.

/cc @mapbox/mobile @twbell

ljbade:
@1ec5 From my reading of relevant tickets, there are still a few bugs and functions that don't take into account tilt?

Just want to make sure we have a list of the ones that have not been fixed to include as known issues with tilt in the release notes.

1ec5:
Now that #2770 is merged, the biggest known issues are #2266, #2259, and possibly #2139. #256 is unimplemented.

bleege:
Looking at the how iOS implemented tilt in https://github.com/mapbox/mapbox-gl-native/pull/2159/files and https://github.com/mapbox/mapbox-gl-native/pull/2193/files it appears that the Core GL has the necessary API for Android implementation.  We _should_ be able to wire this up with JNI and have what we need for the Android bindings API.

The outstanding question is going to be how much will the Android bindings API look like the [Google Maps for Android Camera API](https://developers.google.com/maps/documentation/android-api/views)?  Google Maps Android API uses.  I _think_ we can get by (for now) with simply implementing `CameraPosition`, `CameraPosition.Builder`, and `MapView.animateCamera(CameraPosition cp)` as their example (see below) does.  The "extra" API that we don't need for now (as it'll change the Mapbox API) is `MapView.animateCamera(CameraUpdate cu)` as it deals with zoom, center location, etc.

```java
// Construct a CameraPosition focusing on Mountain View and animate the camera to that position.
CameraPosition cameraPosition = new CameraPosition.Builder()
    .target(MOUNTAIN_VIEW)      // Sets the center of the map to Mountain View
    .zoom(17)                   // Sets the zoom
    .bearing(90)                // Sets the orientation of the camera to east
    .tilt(30)                   // Sets the tilt of the camera to 30 degrees
    .build();                   // Creates a CameraPosition from the builder
map.animateCamera(CameraUpdateFactory.newCameraPosition(cameraPosition));
```

/cc @zugaldia @tobrun @1ec5

1ec5:
[...]

PR #3171, opened 2015-12-03
Implementation of flyTo.

adam-mapbox:
@1ec5 @jfirebaugh @incanus @bleege @mapbox/mobile 

Open for suggestions about the API on MGLMapView.

### DIFF
--- a/include/mbgl/ios/MGLMapView.h
+++ b/include/mbgl/ios/MGLMapView.h
@@ -218,6 +218,14 @@ IB_DESIGNABLE
 *   @param completion The block to execute after the animation finishes. */
 - (void)setCamera:(MGLMapCamera *)camera withDuration:(NSTimeInterval)duration animationTimingFunction:(nullable CAMediaTimingFunction *)function completionHandler:(nullable void (^)(void))completion;
 
+#pragma mark - flyTo
+
+/** Uses a ballistic parabolic motion to "fly" the viewpoint to a different location with respect to the map with an optional transition duration.
+ *   @param camera The new viewpoint.
+ *   @param duration The amount of time, measured in seconds, that the transition animation should take. Specify `0` to jump to the new viewpoint instantaneously.
+ *   @param completion The block to execute after the animation finishes. */
+- (void)flyToCamera:(MGLMapCamera *)camera withDuration:(NSTimeInterval)duration completionHandler:(nullable void (^)(void))completion;
+
 #pragma mark - Converting Map Coordinates
 
 /** @name Converting Map Coordinates */
--- a/src/mbgl/map/transform.cpp
+++ b/src/mbgl/map/transform.cpp
@@ -313,6 +313,107 @@ void Transform::_easeTo(const CameraOptions& options, double new_scale, double n
     }
 }
 
+void Transform::flyTo(const CameraOptions &options) {
+    CameraOptions flyOptions(options);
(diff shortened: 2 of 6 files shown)


adam-mapbox:
Supersedes https://github.com/mapbox/mapbox-gl-native/tree/1ec5-what-goes-up-256.

include/mbgl/ios/MGLMapView.h
```diff
@@ -218,6 +218,10 @@ IB_DESIGNABLE
 *   @param completion The block to execute after the animation finishes. */
 - (void)setCamera:(MGLMapCamera *)camera withDuration:(NSTimeInterval)duration animationTimingFunction:(nullable CAMediaTimingFunction *)function completionHandler:(nullable void (^)(void))completion;
 
+#pragma mark - flyTo
+
+- (void)flyTo:(CLLocationCoordinate2D *)coordinate zoomLevel:(double)zoomLevel direction:(CLLocationDirection)direction completionHandler:(nullable void (^)(void))completion;
```
1ec5: This should take an MGLMapCamera and be called `-flyToCamera:withDuration:completionHandler:`. This would also allow you to fly to a tilted viewpoint, looking at a particular coordinate from a particular coordinate, and not have to worry about calculating the pitch you want. We don't need more variations on `-setCenterCoordinate:`, and in fact I'm about to propose that we deprecate some of them.

1ec5: A further idea would be to allow the developer to specify a maximum airspeed velocity instead of a duration, but that can happen in a separate PR if desired.

1ec5: Also, this declaration needs a documentation comment. I would say something about the trajectory being parabolic versus -setCamera which transitions the viewport properties linearly.

src/mbgl/map/transform.cpp
```diff
@@ -313,6 +313,107 @@ void Transform::_easeTo(const CameraOptions& options, double new_scale, double n
     }
 }
 
+void Transform::flyTo(const CameraOptions &options) {
```
kkaefer: Can you please add documentation for all of the math things that go on here? It'd be cool to have comments that describe what you're doing on a high level.

adam-mapbox: I would love to, but a lot of this stuff I don't fully understand myself; I inherited it from @1ec5's branch, and he inherited it from the GL-JS world.  It's not actually clear who originally wrote some of this math.  I agree this is not an ideal situation.

twbell: > I agree this is not an ideal situation.

pinning that.

incanus: /cc @mourner @ansis here maybe for some backup. I've rev-eng'd some of the projection stuff over time, too, so I'll take a look when I review in depth.

incanus:
As on #3172, I think we should take a swipe at Android implementation soon after we're happy and land this. 

Refs https://github.com/mapbox/mapbox-gl-native/issues/2805

include/mbgl/ios/MGLMapView.h
adam-mapbox: Done.

bleege:
Android Implementation is going to be #3244

src/mbgl/map/transform.cpp
1ec5: Copious comments are being added in #3355.

PR #3135, opened 2015-11-25
OS X SDK and revamped OS X demo application

1ec5:
This PR replaces the minimal, glfw-based OS X support in this project with a proper SDK that mirrors (and in many places shares code with) the Mapbox iOS SDK. `make xproj` now creates an Xcode project with a framework target and a rewritten osxapp application target. The vestigial static library target is immediately repackaged as a bone fide dynamic framework (#828) – not the kind of faux framework that framework.sh creates for iOS. (In the future, we can cut out the intermediate static library step.) `make xpackage` simply builds the project that `make xproj` creates.

Inside the SDK is almost all the same functionality as you’d find in the iOS SDK: MGLMapView has support for gestures, animations, annotations, callouts, etc. It lacks support for the user dot, user location tracking, or telemetry. However, the optimized framework weighs in at only 5 MB compared to 290 MB for the iOS SDK. It’s also easier to install: you just plop Mapbox.framework into your application’s Xcode project, set the access token in Info.plist, and add the MGLMapView to MainMenu.xib or Main.storyboard.

osxapp is now a typical Cocoa application that links against Mapbox.framework and allows you to view any GL style in an MGLMapView. osxapp no longer depends on glfw. Go to Help ‣ Mapbox GL Help for an explanation of the various gestures that are supported.

Both SDKs share almost all of the model object classes and most geometry conversion code – basically anything that doesn’t touch AppKit or UIKit. They have separate implementations of view classes, annotation images (because I didn’t want to confuse appledoc with conditional compilation; see #3203), gesture recognition (because the Mac has different standard gestures), and callouts (because NSPopover is such an elegant solution to the problem).

The biggest architectural difference between the two SDKs, other than the use of a framework, is that the OS X version of MGLMapView holds a mapping from numeric “annotation tags” to annotation context objects that contain the annotations themselves (#3159). This optimizes nearly every operation involving annotations. It allowed me to fix longstanding bugs and performance bottlenecks such as #1504 right off the bat and implement some nifty Mac-only features like annotation tooltips, custom cursors, and proper hit testing for annotations. (A doughnut-shaped annotation can only be selected by clicking the dough, not the transparent hole in the middle.)

[...]

PR #3172, opened 2015-12-03
initial featuresAt implementation

### LINKED ISSUE #352: Feature eventing (native equivalent of queryRenderedFeatures and querySourceFeatures)

incanus:
This is ready for implementation review, then it can be squashed. The goal here is a minimal `featuresAt` functionality with configurable radius (owing to use in touch gesture scenarios) that returns basic information about queries on style layers marked `interactive`. 

Right now the values returned for each feature are: 

- Style layer name as string
- Style source name as string
- Properties as string keys & values

Couple things that can be added later: 

- Further qualifying features returned by query layer, feature type, or true point-in-polygon intersectionality as in GL JS. 
- More concrete and strongly-typed results, including style reference objects and feature geometries, when #837 and #2161 land. 

One note @jfirebaugh: I wanted to adjust your `PointAnnotationImpl` use of the `BOOST_GEOMETRY_REGISTER_POINT_2D` and `BOOST_GEOMETRY_REGISTER_BOX` macros to be more in line with my setup in `interactive_features_impl.hpp` and my adjustment of `collision_tile.hpp`. It should be straightforward but I just couldn't figure it out for named field types in the Boost R-tree indexing scheme. 

Quick demo on iOS: 

anigif-1449122597

Once this looks good and lands, I will add the Android bindings as well. 

Fixes #352.


incanus:
Another thing we can address after this lands: we can probably merge the existing annotations query functionality into it, since it's built around the abstract concept of `AnnotationTile` objects. Then, our traditional annotation query API can just qualify by layer and/or source type. This would be the first step towards a map that could blur the distinction between maps with user annotation data and maps with inherent embedded vector tile data (POIs etc.) that are both able to be queried with native popups.

incanus:
- Upside to this PR: We can have interactive things on iOS today, and Android very shortly afterwards.
- Downside: the return type contents for both C++ and platform will change when #2161 and #837 land.

We also could do some stuff with this PR and #3135, but I don't think that's a blocker to getting the feature in mobile. 

Could use a code review if there's time @kkaefer @jfirebaugh @mourner @1ec5.

include/mbgl/ios/MGLMapView.h
```diff
@@ -82,6 +82,14 @@ IB_DESIGNABLE
 /** The button shown in the lower-right of the map which when pressed displays the map attribution information. */
 @property (nonatomic, readonly) UIButton *attributionButton;
 
+/** Query the visible map for features at a given point. 
+*
+*   @warning Beta API. Return value subject to change in subsequent releases.
```
1ec5: If we’re considering this a “beta API”, I think it would be better to leave it undocumented (actually, just convert this block to a normal comment).

include/mbgl/ios/MGLMapView.h
```diff
@@ -82,6 +82,14 @@ IB_DESIGNABLE
...
+*
+*   @param point A point on screen, for example at a user gesture.
+*   @return An array of `NSDictionary` objects representing features near the point. Current fields provided are the style `layer`, the style `source`, and a nested dictionary of properties of the feature as `NSString` keys and values. */
+- (NS_ARRAY_OF(NSDictionary *) *)featuresAt:(CGPoint)point;
```
1ec5: Something more descriptive would be nice, like `-featureDescriptionsAtPoint:`. That way, when we replace this with something more strongly typed, we can put the new API alongside named `-featuresAtPoint:`.

incanus: That’s a good idea, and a good way to set us up for proper deprecation.

platform/ios/MGLMapView.mm
```diff
@@ -383,6 +383,40 @@ - (void)commonInit
     }];
 }
 
+- (NS_ARRAY_OF(NSDictionary *) *)featuresAt:(CGPoint)point
+{
+    // flip y for core
+    point.y = self.bounds.size.height - point.y;
```
1ec5: To avoid both this flip and having to pass the `TransformState` into the style, we should call `-convertPoint:toCoordinateFromView:` here and pass a coordinate in.

platform/ios/MGLMapView.mm
```diff
@@ -383,6 +383,40 @@ - (void)commonInit
...
+
+    mbgl::FeatureResults results = _mbglMap->featuresAt(mbgl::PrecisionPoint(point.x, point.y), 50);
+
+    NSMutableArray *features = [NSMutableArray arrayWithCapacity:results.size()];
+
+    for (const auto& result : results)
+    {
+        NSString *layerName  = [NSString stringWithUTF8String:std::get<0>(result).c_str()];
```
1ec5: Nit: You can also write this as `@(std::get<0>(result).c_str())`.

src/mbgl/map/map.cpp
```diff
@@ -407,6 +407,13 @@ LatLngBounds Map::getBoundsForAnnotations(const AnnotationIDs& annotations) {
 }
 
 
+#pragma mark - Features
+
+FeatureResults Map::featuresAt(const PrecisionPoint point, const uint8_t radius) const {
+    return context->invokeSync<std::vector<std::tuple<std::string, std::string, FeatureProperties>>>(&MapContext::featuresAt, point, radius);
```
1ec5: This can be `context->invokeSync<FeatureResults>`, no?

src/mbgl/map/tile_worker.cpp
```diff
@@ -136,6 +137,43 @@ void TileWorker::parseLayer(const StyleLayer& layer, const GeometryTile& geometr
 
     std::unique_ptr<Bucket> bucket = layer.createBucket(parameters);
 
+    if (layer.interactive) {
+        for (std::size_t i = 0; i < geometryLayer->featureCount(); i++) {
+            const auto feature = geometryLayer->getFeature(i);
+            FeatureBox featureBox = {{ 4096, 4096 }, { -1, -1 }};
```
1ec5: What’s the significance of 4,096? Can you put that in a named constant so it’s clear?

src/mbgl/style/style_layer.hpp
```diff
@@ -56,6 +56,7 @@ class StyleLayer : public util::noncopyable {
     float minZoom = -std::numeric_limits<float>::infinity();
     float maxZoom = std::numeric_limits<float>::infinity();
     VisibilityType visibility = VisibilityType::Visible;
+    bool interactive;
```
1ec5: I would’ve called this `isInteractive`, but I don’t know what the prevailing style is in mbgl.

src/mbgl/map/geometry_tile.hpp
```diff
@@ -30,6 +31,7 @@ class GeometryTileFeature : private util::noncopyable {
     virtual ~GeometryTileFeature() = default;
     virtual FeatureType getType() const = 0;
     virtual mapbox::util::optional<Value> getValue(const std::string& key) const = 0;
+    virtual std::unordered_map<std::string, std::string> getAllValues() const = 0;
```
kkaefer: This returns a copy; is that intentional?

src/mbgl/map/source.cpp
```diff
@@ -603,4 +606,58 @@ void Source::dumpDebugLogs() const {
...
+
+    // find the right tile
+    for (const auto& tile : getLoadedTiles()) {
+        if (tile->id != id) continue;
+
+        // query the tile
+        const auto& data = tile->data;
+        data->featureTree.query(boost::geometry::index::intersects(queryBox),
+                                boost::make_function_output_iterator([&](const auto& val) {
+            const std::string layer_id = std::get<1>(val);
```
kkaefer: Any reason for copying these?

src/mbgl/map/source.cpp
```diff
@@ -603,4 +606,58 @@ void Source::dumpDebugLogs() const {
     }
 }
 
+FeatureResults Source::featuresAt(const PrecisionPoint point, const uint8_t radius, const TransformState& transform) const {
```
kkaefer: Add more `const` modifiers to clarify intent.

kkaefer: Why are we using a `uint8_t` for radius?

src/mbgl/map/source.cpp
```diff
@@ -603,4 +606,58 @@ void Source::dumpDebugLogs() const {
...
+    // figure out tile (bounded by source max zoom)
+    LatLng p = transform.pointToLatLng(point);
+
+    const double sine = std::sin(p.latitude * M_PI / 180);
```
kkaefer: Please document math.

src/mbgl/map/source.cpp
```diff
@@ -603,4 +606,58 @@ void Source::dumpDebugLogs() const {
...
+    const int16_t right  = position.x + r;
+    const int16_t bottom = position.y - r;
+    const int16_t top    = position.y + r;
+
+    FeatureBox queryBox = {{ left, bottom }, { right, top }};
+
+    FeatureResults results;
+
+    // find the right tile
+    for (const auto& tile : getLoadedTiles()) {
```
kkaefer: This seems really inefficient; can we instead implement a function that just gets the correct tile from the unordered map?

src/mbgl/map/source.cpp
1ec5: At a glance, I think some of this math goes away entirely if we account for https://github.com/mapbox/mapbox-gl-native/pull/3172/files#discussion_r47417361.

src/mbgl/map/source.cpp
```diff
@@ -603,4 +606,58 @@ void Source::dumpDebugLogs() const {
...
 
+FeatureResults Source::featuresAt(const PrecisionPoint point, const uint8_t radius, const TransformState& transform) const {
+    // figure out tile (bounded by source max zoom)
+    LatLng p = transform.pointToLatLng(point);
+
+    const double sine = std::sin(p.latitude * M_PI / 180);
+    const double x = p.longitude / 360 + 0.5;
+    double y = 0.5 - 0.25 * std::log((1 + sine) / (1 - sine)) / M_PI;
+
+    y = y < -1 ? -1 : y > 1 ? 1 : y;
```
mourner: Why do we transform point to latlng and then immediately project it back?

src/mbgl/map/source.cpp
```diff
@@ -603,4 +606,58 @@ void Source::dumpDebugLogs() const {
...
+    const int16_t bottom = position.y - r;
+    const int16_t top    = position.y + r;
+
+    FeatureBox queryBox = {{ left, bottom }, { right, top }};
+
+    FeatureResults results;
+
+    // find the right tile
+    for (const auto& tile : getLoadedTiles()) {
+        if (tile->id != id) continue;
```
mourner: Do we only include results from a single tile the query point is on? It should consider all tiles within a radius of a query point (usually 1 but may be more).

src/mbgl/map/vector_tile.cpp
```diff
@@ -84,6 +84,21 @@ mapbox::util::optional<Value> VectorTileFeature::getValue(const std::string& key
     return mapbox::util::optional<Value>();
 }
 
+std::unordered_map<std::string, std::string> VectorTileFeature::getAllValues() const {
+    std::unordered_map<std::string, std::string> values;
+
+    for (const auto key_it : layer.keys) {
+        const auto& key = key_it.first;
+        const auto maybe_val = getValue(key);
```
kkaefer: We're using `camelCase` naming convention for local vars.

src/mbgl/style/style.cpp
```diff
@@ -265,6 +265,21 @@ void Style::emitResourceLoadingFailed(std::exception_ptr error) {
...
 }
 
+FeatureResults Style::featuresAt(const PrecisionPoint point, const uint8_t radius, const TransformState& transform) const {
+    assert(util::ThreadContext::currentlyOn(util::ThreadType::Map));
+
+    FeatureResults results;
+
+    for (const auto& source : sources) {
+        if (source->info.type == SourceType::Vector && source->isLoaded()) {
+            const auto source_results = source->featuresAt(point, radius, transform);
```
kkaefer: Maybe pass the output iterator to `featuresAt`?

src/mbgl/map/source.cpp
```diff
@@ -603,4 +606,58 @@ void Source::dumpDebugLogs() const {
...
+
+        // query the tile
+        const auto& data = tile->data;
+        data->featureTree.query(boost::geometry::index::intersects(queryBox),
+                                boost::make_function_output_iterator([&](const auto& val) {
+            const std::string layer_id = std::get<1>(val);
+            const FeatureProperties feature_properties = std::get<2>(val);
+            const auto result = std::make_tuple(layer_id, info.source_id, feature_properties);
```
kkaefer: Why is this a tuple? Can't we make a real type for it with named members? Index based members are a disaster waiting to happen.

src/mbgl/annotation/annotation_tile.cpp
```diff
@@ -19,6 +19,10 @@ mapbox::util::optional<Value> AnnotationTileFeature::getValue(const std::string&
     return mapbox::util::optional<Value>();
 }
 
+std::unordered_map<std::string, std::string> AnnotationTileFeature::getAllValues() const {
```
kkaefer: Shouldn't we implement the querying part in the Tile object itself rather than just stealing the features from it?

src/mbgl/map/tile_worker.cpp
```diff
@@ -136,6 +137,43 @@ void TileWorker::parseLayer(const StyleLayer& layer, const GeometryTile& geometr
...
+                    }
+                    if (point.y > max.get<1>()) {
+                        featureBox.max_corner().set<1>(::fmin(point.y, 4095));
+                    }
+                }
+            }
+
+            if (featureBox.min_corner().get<0>() < 4096 && featureBox.min_corner().get<1>() < 4096 &&
+                featureBox.max_corner().get<0>() > -1   && featureBox.max_corner().get<1>() > -1) {
+                const auto& values = feature->getAllValues();
```
kkaefer: This is a bug: `getAllValues` returns a value, but it's being stored in a reference

src/mbgl/map/source.cpp
1ec5: `radius` appears to be measured in pixels. Does it account for pixel density?

platform/ios/MGLMapView.mm
```diff
@@ -383,6 +383,40 @@ - (void)commonInit
     }];
 }
 
+- (NS_ARRAY_OF(NSDictionary *) *)featuresAt:(CGPoint)point
+{
+    // flip y for core
+    point.y = self.bounds.size.height - point.y;
+
+    mbgl::FeatureResults results = _mbglMap->featuresAt(mbgl::PrecisionPoint(point.x, point.y), 50);
```
1ec5: The radius should be configurable. A 50-pixel radius (is that 25 or 50 points?) may not make sense for all use cases.

include/mbgl/ios/MGLMapView.h
incanus: Let's document this, but I've renamed the API in https://github.com/mapbox/mapbox-gl-native/commit/0ea4f537bbb48a1babccc2702b3e16dce9379a0c to make room for our final, more-richly-typed API once style API lands.

include/mbgl/ios/MGLMapView.h
incanus: https://github.com/mapbox/mapbox-gl-native/commit/0ea4f537bbb48a1babccc2702b3e16dce9379a0c

platform/ios/MGLMapView.mm
incanus: Like the JS API, we do the work of figuring out the tile deep down in the source. Converting to a coordinate still requires us to access the transform in the source since we still need to use the `Transform::pointToCoordinate` API to determine the `TileCoordinate` (separate type from both `LatLng` and screen `PrecisionPoint`. We also need access to the current `Transform::scale` since we need are querying for features at the resolution of the tile extent (i.e. 4096) and the scale informs the query box size in terms of the radius passed.

platform/ios/MGLMapView.mm
incanus: https://github.com/mapbox/mapbox-gl-native/commit/262665e2250e4b51c8f4bec9b162ec689255b8f1

src/mbgl/map/map.cpp
incanus: https://github.com/mapbox/mapbox-gl-native/commit/b98b686b9e38ddd254098d671f0a93c0c54497d8

src/mbgl/map/tile_worker.cpp
incanus: https://github.com/mapbox/mapbox-gl-native/commit/4c639e5c01202cb489f44411cc555590225c8a14

src/mbgl/map/geometry_tile.hpp
incanus: Yes, because in building the abstract API for all `GeometryTile` classes, we have to support `VectorTile`, which creates its map from raw vector tile data. Applying a patch such as https://gist.github.com/incanus/d4302ba58616c5096442 to return references doesn't work because we need to create the map on the stack and get: 

```
.../src/mbgl/map/vector_tile.cpp:99:12: Reference to stack memory associated with local variable 'values' returned
```

src/mbgl/map/source.cpp
incanus: Fixed as part of https://github.com/mapbox/mapbox-gl-native/commit/505f929414e1b6e4491b90cbfbc0586921e438b1.

src/mbgl/map/source.cpp
incanus: > Add more `const` modifiers in this function to clarify intent.

@kkaefer Can you clarify what you mean? 

> Why are we using a `uint8_t` for radius?

Fixed in https://github.com/mapbox/mapbox-gl-native/commit/d123545de9239bcab735c9ed77abbf785a1a6e9c. 

> `radius` appears to be measured in pixels. Does it account for pixel density?

It does, as the point and radius refer to screen points.

src/mbgl/map/source.cpp
incanus: https://github.com/mapbox/mapbox-gl-native/commit/ac655cebe68065322eb773d5b96fa9dc80ddabc8

src/mbgl/map/source.cpp
incanus: https://github.com/mapbox/mapbox-gl-native/commit/ac655cebe68065322eb773d5b96fa9dc80ddabc8

src/mbgl/map/source.cpp
incanus: Added some comments in https://github.com/mapbox/mapbox-gl-native/commit/ac655cebe68065322eb773d5b96fa9dc80ddabc8, but see also https://github.com/mapbox/mapbox-gl-native/pull/3172#discussion_r47561929.

src/mbgl/map/source.cpp
incanus: Yes, just like on JS. 

- https://github.com/mapbox/mapbox-gl-js/blob/ba9d1095ce5509595a0014a18037c4a3ba3b879d/js/source/source.js#L92
- https://github.com/mapbox/mapbox-gl-js/blob/ba9d1095ce5509595a0014a18037c4a3ba3b879d/js/source/tile_pyramid.js#L349-L350

src/mbgl/map/vector_tile.cpp
incanus: https://github.com/mapbox/mapbox-gl-native/commit/b6ad5e2205aa9eb21cb4d7b47346b4cc1130d1c6

src/mbgl/style/style.cpp
incanus: @kkaefer Can you clarify what you mean?

src/mbgl/map/source.cpp
incanus: https://github.com/mapbox/mapbox-gl-native/commit/505f929414e1b6e4491b90cbfbc0586921e438b1

src/mbgl/map/tile_worker.cpp
incanus: https://github.com/mapbox/mapbox-gl-native/commit/beda70d55f02e8d207abae98564cf79b10593cfc

platform/ios/MGLMapView.mm
incanus: https://github.com/mapbox/mapbox-gl-native/commit/394c48403f11a27e8e89c412f245520ac1d28aeb

include/mbgl/ios/MGLMapView.h
```diff
@@ -82,6 +82,13 @@ IB_DESIGNABLE
 /** The button shown in the lower-right of the map which when pressed displays the map attribution information. */
 @property (nonatomic, readonly) UIButton *attributionButton;
 
+/** Query the visible map for features at a given point. 
+*
+*   @param point A point on screen, for example at a user gesture.
+*   @param radius A distance in which to query features around the point.
```
1ec5: Should clarify that the distance is measured in screen points, not a physical unit or coordinate delta.

src/mbgl/style/style_layer.hpp
incanus: https://github.com/mapbox/mapbox-gl-native/commit/6c86e789cb38c4f2e2d570a054720a05e8873f2f

include/mbgl/ios/MGLMapView.h
incanus: https://github.com/mapbox/mapbox-gl-native/commit/10d36f899410eb3f058077c62a9df12111bbfd31

incanus:
Cleaned up the demo a bit so that it works on all POI layers, regardless of zoom level, to be a little more self-evident. 

anigif-1450140277

src/mbgl/map/source.hpp
```diff
@@ -5,6 +5,8 @@
 #include <mbgl/map/source_info.hpp>
 
 #include <mbgl/util/mat4.hpp>
+#include <mbgl/util/interactive_features_impl.hpp>
```
jfirebaugh: This can be `#include <mbgl/util/interactive_features.hpp>` right?

jfirebaugh:
Tests?

include/mbgl/darwin/MGLTypes.h
```diff
@@ -26,6 +26,15 @@ typedef NS_ENUM(NSUInteger, MGLUserTrackingMode) {
     MGLUserTrackingModeFollowWithCourse,
 };
 
+/** The style layer name for a queried map feature. */
+extern NSString * const MGLFeatureLayerNameKey;
```
1ec5: Unlike `MGLUserTrackingMode`, these keys are only used by MGLMapView, so they can go in MGLMapView.h.

1ec5:
Would it be straightforward to include coordinates (either screen or geographic coordinates) in each feature description? The use case would be MapKit-style clickable POIs: you’d want the callout to be anchored on the POI, not the tap point, which may be a small distance away.

incanus:
We are dealing with more than just points here. Now that we support GeoJSON sources, the possibility exists to pass actual features, but the early intent here is the use case of tying to external identifier such as OSM ID. We can always make a more robust API in future, particularly when the style API lands. I'm going for basic functionality that I and others can iterate on going forward, but also meeting the use cases we are hearing about as blockers right now.

1ec5:
:+1:

1ec5:
GL JS has revamped (and renamed) the featuresAt feature in mapbox/mapbox-gl-js#2224.

jfirebaugh:
Closing here; we'll want an implementation that more closely follows the new paradigm established in Mapbox GL JS.

PR #3177, opened 2015-12-03
record app fg/bg/inactive state in telemetry

incanus:
To make data introspection more useful, we should know what state the app was in (`UIApplicationState`).

@bleege @friedbunny @1ec5 @camilleanne @mick @lxbarth


platform/ios/MGLMapboxEvents.m
```diff
@@ -655,6 +657,41 @@ - (NSString *) deviceOrientation {
...
+                result = @"Active";
+                break;
+            case UIApplicationStateInactive:
+                result = @"Inactive";
+                break;
+            case UIApplicationStateBackground:
+                result = @"Background";
+                break;
+            default:
+                result = @"Default - Unknown";
```
1ec5: The string keys don’t need to be user-presentable; the backend would take care of that.

PR #3174, opened 2015-12-04
Make redoPlacement faster

jfirebaugh:
With https://github.com/mapbox/mapbox-gl-native/commit/bafbf6c04b1bfd5da84411f52a72e26783f3bcb7, label placement is very slow. Label placement occurs when the map is rotated or tilted (but not on zoom or pan). It's slow because we're creating a copy of the StyleLayer* objects on essentially every frame when rotating.

/cc @jfirebaugh

### DIFF
--- a/src/mbgl/layer/background_layer.cpp
+++ b/src/mbgl/layer/background_layer.cpp
@@ -4,10 +4,7 @@
 namespace mbgl {
 
 std::unique_ptr<StyleLayer> BackgroundLayer::clone() const {
-    std::unique_ptr<BackgroundLayer> result = std::make_unique<BackgroundLayer>();
-    result->copy(*this);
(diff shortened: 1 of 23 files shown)


jfirebaugh:
:eyes: @kkaefer

kkaefer:
Looks great!

Something that is out of scope for this patch: Now that `StyleLayer` isn't noncopyable anymore, we can make it movable and remove the indirection introduced by wrapping it in `std::unique_ptr`

jfirebaugh:
`StyleLayer` is an abstract base class, so there still needs to be a type-erasing wrapper somewhere. I think `std::unique_ptr` is as good as any.

kkaefer:
Ah, alright. :+1:

PR #3200, opened 2015-12-05
Android Advanced / Dynamic Map Overlays

manimaul:
This PR add the SDK user ability to create advanced / dynamic Map View overlays on Android.
The existing compass and location views are updated to be Overlays. 

Note: updates to native map view and JNI to allow for zero java allocations on draw methods.
Note: public MapView.addOverlay(Overlay) method and public Overlay interface.


tobrun:
Thank you for reaching out and contributing, 
could you elaborate what you the benefit of having overlays would be versus using Views?

manimaul:
Thanks tobrun for following up with me.  The main benefit is exposing the ability to add an Overlay.  The Overlay interface does not exclude the possibility of an Overlay being a View.  In fact, it would be relatively simple to keep them as Views.  However, examples are sometimes the best documentation, so It seemed good to add an example.  

Additionally, SDK users like myself have lots of code from the previous MapBox Android SDK and also perhaps Osmdroid that leverage the flexible Overlay system there.  This PR allows us to make slight modifications to our existing Overlays and be able to use the new GL SDK.

Thanks,

Will

manimaul:
Closing PR due to lack of feedback.

mb12:
@manimaul The biggest advantage of the Overlay pattern is integrating with existing code. However if you have line, area or point markers, please use the annotation APIs instead of the Overlay pattern.  This will give you better performance and zero lag/jitter with respect to base map.

1.) All lines, areas and points are tesselated in GL and coalesced into glDrawElements calls. For e.g. if you add 10000 point annotations, this would translate  into 40000 vertices  and a single drawElements calls in each frame in mapbox GL. It would be very difficult to match performance of mapbox GL for annotations with Overlay pattern.  

2.) Another problem with Overlay pattern is synchronizing Android View/iOS UIView with GL MapView.  This is a problem if the Overlay needs to pan with the map. However if the overlay is static (position wise) like Scale bar, Overlay pattern works very well.

tobrun:
@manimaul 
Sorry for not getting back to you in time. Thank you for this contribution, it has given us some great insights to work with. We have discussed this API internally but feel it's not matching our current implementation or the direction we are heading.

As indicated by @mb12, there are 2 types of components to draw content on a map:
 - GL based components
 - Android View based components

If we would introduce a overlay system I would like to have all overlay items incorporated (both GL as Android View). I currently don't see a way to make that work.

For the Android View based components, we are thinking about creating a generic system #3276, which will incorporate some ideas shown from this PR.

PR #3202, opened 2015-12-06
Fix linker warnings about overridden styleURL__

1ec5:
Reapplied #3097, which was originally applied to the release-ios-v3.0.0 branch but was omitted in a merge back to master.

/cc @incanus

### DIFF
--- a/include/mbgl/ios/MGLMapView.h
+++ b/include/mbgl/ios/MGLMapView.h
@@ -259,9 +259,6 @@ IB_DESIGNABLE
 *   To display the default style, set this property to `nil`. */
 @property (nonatomic, null_resettable) NSURL *styleURL;
 
-/* Discourage programmatic usage of this IB-only property. Interface Builder skips over this declaration because it is unable to parse attributes. See the real declaration in MGLMapView+IBAdditions.h. */
-@property (nonatomic, nullable) IBInspectable NSString *styleURL__ __attribute__((unavailable("styleURL__ is for use within Interface Builder only. Use styleURL in code.")));
-
 /** Currently active style classes, represented as an array of string identifiers. */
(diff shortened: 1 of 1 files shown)


incanus:
> originally applied to the release-ios-v3.0.0 branch but was omitted in a merge back to master

Can you say more about this @1ec5? I'm not sure how this could have happened, and I want to avoid it in future.

incanus:
It appears to have been cherry-picked in https://github.com/mapbox/mapbox-gl-native/commit/c054441071d3ed050fb2a8ff393913aadf83cea7 under the `ios-v3.0.0` tree.

1ec5:
Half of the original change (to MGLMapView+IBAdditions.h) made it back to master, but the other half (MGLMapView.h) went missing in transit.

incanus:
If you look at https://github.com/mapbox/mapbox-gl-native/commits/ios-v3.0.0 though, you can see https://github.com/mapbox/mapbox-gl-native/commit/c054441071d3ed050fb2a8ff393913aadf83cea7 applied with changes in both. 

Forgive my persistence — I want to make sure we have good technique around branch management, and this could be indicative of missing changes we're _not_ seeing.

1ec5:
It wasn’t cherry-picked into release-ios-v3.0.0 – I merged it directly into that branch. When you merged that branch back to master, the merge strategy you used probably omitted the MGLMapView.h change. I think that’s because the property was originally added to master for #3088, then cherry-picked to release-ios-v3.0.0. When merging release-ios-v3.0.0 back to master, git must’ve recorded release-ios-v3.0.0 as having two changes that canceled each other out, whereas master had a seemingly unrelated change that was not negated and could thus live on. Hope that clears things up!

incanus:
Aha! Yes. The big variable here then is merging PR's directly into release branches. Got it.

PR #3066, opened 2015-12-07
Do not alter OpenGL viewport

kkaefer:
We're currently setting the OpenGL viewport for every frame. Instead, we should move this to the bindings since they know best when the viewport will change. This would also allow bindings to render the map into a specific area of the screen, rather than grabbing the full screen. In particular, there's no relation between the dimensions of the map (which must match the viewport dimensions or the map will be distorted) and the size of the OpenGL context/backbuffer.


1ec5:
Do the Android, iOS, and OS X frontends need to be updated? They don’t currently set the viewport themselves.

kkaefer:
@1ec5 I checked all of those and it seems that resizing the framebuffer automatically sets the viewport, and it's also set initially.

PR #3211, opened 2015-12-07
[core] Replace StyleLayerType with RTTI

jfirebaugh:
This eliminates a potential source of bugs: forgetting to set the layer's `type`. Now the type is entirely determined by the runtime type information inherent in the layer subclasses.

:eyes: @kkaefer


src/mbgl/map/tile_worker.cpp
```diff
@@ -65,22 +67,18 @@ TileParseResult TileWorker::parsePendingLayers() {
     // dependencies.
     for (auto it = pending.begin(); it != pending.end();) {
         auto& layer = *it->first;
-        auto& bucket = it->second;
+        auto bucket = dynamic_cast<SymbolBucket*>(it->second.get());
```
kkaefer: We should `continue` if `bucket == nullptr`.

src/mbgl/map/tile_worker.cpp
```diff
@@ -144,10 +142,11 @@ void TileWorker::parseLayer(const StyleLayer* layer, const GeometryTile& geometr
                                      *collisionTile);
 
     std::unique_ptr<Bucket> bucket = layer->createBucket(parameters);
+    const SymbolLayer* symbolLayer = layer->as<SymbolLayer>();
```
kkaefer: Maybe add a comment here that it's `null` when conversion fails, since it's not obvious that `.as<>` does a `dynamic_cast` internally.

src/mbgl/map/tile_worker.cpp
jfirebaugh: Only symbol layers can be pending, so it should never be null. I'll change the types to enforce this more strongly.

kkaefer: :+1:

jfirebaugh: Turns out there's not a good way to do a moving dynamic cast of a unique_ptr, so I'm going to comment the assert on the next line and leave it as-is.

kkaefer:
:+1:

PR #3165, opened 2015-12-08
Use ccache

kkaefer:
We should use ccache to get faster incremental compiles.

- [x] Enable on **Linux + node.js** builds

  Should be as simple as changing the `CXX` variable

- [ ] ~~Enable on **Xcode** builds ([blog post](https://pspdfkit.com/blog/2015/ccache-for-fun-and-profit/))~~

  Caveat: Travis doesn't cache build artifacts on OS X :cry: 

- [ ] ~~Enable on **Android** builds~~
  
  Caveat: Android NDK r10e [uses shell scripts](https://code.google.com/p/android/issues/detail?id=182301) in place of `clang++` executables, but they don't contain a shebang, so `ccache` fails to invoke them (`failed: Exec format error`).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -30,64 +30,64 @@ matrix:
     # OS X - Xcode 7 - Debug
     - os: osx
       osx_image: xcode7
-      compiler: ": xcode-debug"
+      compiler: clang
       env: FLAVOR=osx BUILDTYPE=Debug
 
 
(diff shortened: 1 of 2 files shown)


kkaefer:
ccache with Xcode works, but Travis doesn't cache on OS X :(

PR #3214, opened 2015-12-08
Set the bearing of the checked object to avoid a null exception

### LINKED ISSUE #3213: Null exception in onResume() when mocking GPS
This exception is found when setting a background service that mocks the GPS position for a route by using Android's `LocationManager.addTestProvider()`. If we have this enabled and following a route then we get  the following exception:

```
E/AndroidRuntime(16917): FATAL EXCEPTION: main
E/AndroidRuntime(16917): java.lang.NullPointerException: Attempt to invoke virtual method 'float android.location.Location.getBearing()' on a null object reference
E/AndroidRuntime(16917):    at com.mapbox.mapboxsdk.views.UserLocationView.setLocation(UserLocationView.java:577)
```

zugaldia:
Fixes #3213.

/cc: @bleege

### DIFF
--- a/platform/android/mapboxgl-app.gypi
+++ b/platform/android/mapboxgl-app.gypi
@@ -72,6 +72,10 @@
         './src/example_custom_layer.cpp',
       ],
 
+      'include_dirs': [
+        '../../include',
+      ],
+
       'variables': {
         'ldflags': [
           '-llog',
--- a/platform/android/src/example_custom_layer.cpp
+++ b/platform/android/src/example_custom_layer.cpp
@@ -1,6 +1,8 @@
 #include <jni.h>
 #include <GLES2/gl2.h>
 
+#include <mbgl/style/types.hpp>
+
 static const GLchar * vertexShaderSource = "attribute vec2 a_pos; void main() { gl_Position = vec4(a_pos, 0, 1); }";
 static const GLchar * fragmentShaderSource = "void main() { gl_FragColor = vec4(0, 1, 0, 1); }";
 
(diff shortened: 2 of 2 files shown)


android/MapboxGLAndroidSDK/src/main/java/com/mapbox/mapboxsdk/views/UserLocationView.java
```diff
@@ -574,7 +574,7 @@ private void setLocation(Location location) {
             mShowDirection = true;
             mGpsMarkerDirection = 0;
             if (location.hasBearing()) {
-                mMapView.setBearing(mUserLocation.getBearing());
+                mMapView.setBearing(location.getBearing());
```
bleege: @zugaldia I'm wondering if it might make more sense to move `mUserLocation = location` up in the method instead.  [Currently](https://github.com/mapbox/mapbox-gl-native/blob/a0f2553ab4af7e9723e1e0723bb070f665321aeb/android/MapboxGLAndroidSDK/src/main/java/com/mapbox/mapboxsdk/views/UserLocationView.java#L591) it's after this code.

zugaldia: @bleege I'm not very familiar with that class, but moving the assignment earlier will change the behavior of other blocks of code like https://github.com/mapbox/mapbox-gl-native/blob/a0f2553ab4af7e9723e1e0723bb070f665321aeb/android/MapboxGLAndroidSDK/src/main/java/com/mapbox/mapboxsdk/views/UserLocationView.java#L581-L589 that rely on a previous value of `mUserLocation`.

bleege: Makes sense.  Let's merge this as is then.

jfirebaugh:
@zugaldia Is this mergeable?

zugaldia:
Yes. I thought it had been merged already.

jfirebaugh:
Rebasing onto master resulted in zero commits so I think it had been.

zugaldia:
@jfirebaugh thank you. I should've cleaned up this branch earlier.

PR #3057, opened 2015-12-08
Fix font rendering at the top of the viewport

kkaefer:
When the map is tilted, glyphs at the top of the viewport are shaded incorrectly:

screen shot 2015-11-17 at 19 32 41

### DIFF
--- a/scripts/build-shaders.py
+++ b/scripts/build-shaders.py
@@ -1,6 +1,6 @@
 #!/usr/bin/env python
 
-import sys, re, os, errno
+import sys, re, os, errno, hashlib
 
 
 output_dir = sys.argv[1]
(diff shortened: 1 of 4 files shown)


1ec5:
Since #3135 landed, the new osxapp obscures the problematic labels by default with a translucent toolbar. You can still see the issue if you hide the toolbar though.

I see that this PR has a “do not merge” label on it – what’s left to do?

jfirebaugh:
@kkaefer What's left here? Can this be merged?

jfirebaugh:
It looks like one of the two commits was merged in 2700ea87a4e6abb9c8ea8520fe92b067be022935. @kkaefer, if you want the other commit ("only write shaders if the file content changed"), please rebase and merge in a new PR.

kkaefer:
@jfirebaugh 2700ea87a4e6abb9c8ea8520fe92b067be022935 is no longer necessary, since we're now generating the shaders individually with gyp rules, rather than always writing them.

Issue #2151, opened 2015-08-21
Add centerOffset and calloutOffset properties to MGLAnnotation


friedbunny:
MGLAnnotation needs the ability to shift the positioning of annotations and their callouts, akin to MapKit's [`centerOffset`](https://developer.apple.com/library/ios/documentation/MapKit/Reference/MKAnnotationView_Class/index.html#//apple_ref/occ/instp/MKAnnotationView/centerOffset) and [`calloutOffset`](https://developer.apple.com/library/ios/documentation/MapKit/Reference/MKAnnotationView_Class/index.html#//apple_ref/occ/instp/MKAnnotationView/calloutOffset).

Per @incanus in https://github.com/mapbox/mapbox-gl-native/issues/1121#issuecomment-88667899:

> We will hopefully be passing bitmap imagery across the wire instead of relying on pre-made sprite sheets, so this point is moot. At that point, since you are supplying imagery at runtime (which might be created on-the-fly anyway), passing things like [`calloutOffset`](https://developer.apple.com/library/ios/documentation/MapKit/Reference/MKAnnotationView_Class/index.html#//apple_ref/occ/instp/MKAnnotationView/calloutOffset) and [`centerOffset`](https://developer.apple.com/library/ios/documentation/MapKit/Reference/MKAnnotationView_Class/index.html#//apple_ref/occ/instp/MKAnnotationView/centerOffset) is much more natural.

This should probably happen in conjunction with fixing #1504.

/cc @1ec5

incanus:
We can maybe leverage the GL style spec's `translate` property here.

1ec5:
In #3135 for OS X, I’ve eschewed the mbgl::Map methods for looking up the annotation images’ alignment rects directly. Eventually, I’d like to reverse it so that MGLAnnotationImage keeps an `anchorPoint` and `calloutEdge`, by analogy with `-[NSCursor hotspot]`.

RomainQuidet:
Hi,
I'd like to work on this feature.

RomainQuidet:
center offset for annotations is done, need to offset the touch event and the callout too.

RomainQuidet:
checkout pull request #3220 for this to work (core + iOS code)

1ec5:
By way of an update, #3261 introduced support for alignment insets in annotation images, so the “center offset” effect can be achieved by padding the annotation image and insetting it to the visible portion.

BFTE:
[...]

Issue #3164, opened 2015-12-02
never scale images when copying them into the SpriteAtlas


jfirebaugh:
Currently we bilinearly scale images that are not at the native pixel ratio up or down when adding them to the SpriteAtlas. This has led to code that is more complex than necessary and produced issues like #3031 and #2198.

Instead, we should copy images into the SpriteAtlas pixel-for-pixel, and let up/down sampling happen in the shaders.

@kkaefer @ansis -- am I missing anything that would make this non-viable?

ansis:
> Instead, we should copy images into the SpriteAtlas pixel-for-pixel, and let up/down sampling happen in the shaders.
> @kkaefer @ansis -- am I missing anything that would make this non-viable?

I agree that we should do it this way. This is what we're still doing in -js. The scaling was never ported. relevant discussion: https://github.com/mapbox/mapbox-gl-js/issues/1124

jfirebaugh:
@kkaefer @ansis Is there any significance to the use of `2` here:

https://github.com/mapbox/mapbox-gl-js/blob/abe3512815032f02a0edefe38c732b4431535b3a/js/symbol/sprite_atlas.js#L114

While 1 is used everywhere else:

https://github.com/mapbox/mapbox-gl-js/blob/abe3512815032f02a0edefe38c732b4431535b3a/js/symbol/sprite_atlas.js#L91
https://github.com/mapbox/mapbox-gl-js/blob/abe3512815032f02a0edefe38c732b4431535b3a/js/symbol/sprite_atlas.js#L173
https://github.com/mapbox/mapbox-gl-native/blob/2a0d2b7f0bb655607f4104b1d22b23d4cd1d6e2b/src/mbgl/sprite/sprite_atlas.cpp#L99
https://github.com/mapbox/mapbox-gl-native/blob/2a0d2b7f0bb655607f4104b1d22b23d4cd1d6e2b/src/mbgl/sprite/sprite_atlas.cpp#L120
https://github.com/mapbox/mapbox-gl-native/blob/2a0d2b7f0bb655607f4104b1d22b23d4cd1d6e2b/src/mbgl/sprite/sprite_atlas.cpp#L134

If not, I'm going to use `1` everywhere and remove the duplication.

ansis:
I'm not sure, but I think the `2` means 1px padding on each of the two sides. I think it's correct but not clear.

jfirebaugh:
[...]

Issue #3159, opened 2015-12-01
MGLMapView should map annotation IDs to annotation metadata


1ec5:
MGLMapView holds an NSMapTable mapping MGLAnnotation objects to dictionaries containing IDs and other metadata. The problem is that IDs aren’t metadata; they’re the unique key. So many operations that should be simple lookups end up requiring expensive loops or are simply impossible. For example, [this loop through all annotation metadata](https://github.com/mapbox/mapbox-gl-native/blob/e7b6d72cd12aae845f56c92db53dad608217726a/platform/ios/MGLMapView.mm#L1103) can be very expensive, and it isn’t possible to put a fix for #1504 in there because a) we’re iterating over all the annotations, not just the nearby ones, and b) the annotation ID is inaccessible.

I ran into this issue while working on #1496 and again while trying to port the fix for #1504 back from OS X to iOS. For the OS X version of MGLMapView in #3135, I started out with a mapping from annotation IDs (`MGLAnnotationTag`) to annotation context objects, and it has worked out very well. There are couple places where I need to get an ID from an annotation, and performance is fine in those places.

/cc @incanus

tyoshii-uie:
After this change, removeAnnotation() and removeAnnotations() of MGLMapView crashes when annotation has not been added, with following assertion.

NSAssert(annotationTag != MGLAnnotationTagNotFound, @"No ID for annotation %@", annotation);

I think removeAnnotation can be called whether the annotation is added or not.

1ec5:
Thanks, @tyoshii-uie. Fixed in #3685.

t-yoshii:
Thank you for your quick response!

PR #3220, opened 2015-12-08
2151

RomainQuidet:
Add centerOffset and calloutOffset properties to MGLAnnotation
for issue #2151


ios/app/MBXViewController.mm
```diff
@@ -130,127 +158,58 @@ - (void)showSettings
...
-
-            free(polygonCoordinates);
-        }
-    }
-    else if (buttonIndex == actionSheet.firstOtherButtonIndex + 8)
-    {
-        [self.mapView removeAnnotations:self.mapView.annotations];
+    NSInteger settingChoice = buttonIndex - 1;      // cancel is index 0
+    switch (settingChoice) {
+        case settingIndexResetNorth:
```
1ec5: Thanks, the switch statement is a good idea here.

1ec5:
Getting the annotation image for an annotation is a couple simple lookups (first get the metadata dictionary for the annotation, then "dequeue" the annotation's reusable image). Since the callout is an iOS-specific implementation, it doesn't seem necessary to add a parallel field in mbgl code. @incanus, would that be desirable anyways?

RomainQuidet:
all callout offset update is only within iOS code. The only update needed was to take into account the annotation image center offset to position by default the callout y on the top of the image.
The annotation image offset needs to go deep into mgl core as it is rendered here.

1ec5:
Ah, makes sense then. :+1: Although I wonder what should happen if the UIImage has an alignment rect set. (On OS X, I’m using the alignment rect to help determine the hit target and callout offset.)

RomainQuidet:
You're right. The offset reveal a conception problem: the annotation touch event is not "really" linked to the image but to an arbitrary area around the geographical position. It's mostly working, but will fail if the offset is too hight or left/right.
I think this part needs a refactor. How would you like me to help ? What's the better way to work with you on this ?

RomainQuidet:
the tap issue is #1504

1ec5:
Correcting the hit testing will require #3159. It’s a big change, but I’ve done it before as part of #1496, which unfortunately went stale. I may have time either today or tomorrow to work on it, depending on how my other active PRs progress. Once that’s done, either of us can implement the approach taken in 05b039e1641d0ba81d57ee8c56d3f92e9ac14eee for #3135. (See `-installAnnotationImage:` and `-handleClickGesture:`.) I do think your approach is better in terms of not requiring the developer to manually pad the image with whitespace. If we can combine that with the hit testing improvements in 05b039e1641d0ba81d57ee8c56d3f92e9ac14eee, I think we’ll be in good shape for when #837 lands and additional properties become accessible.

RomainQuidet:
Ok, feel free to ask me when you need help on iOS / Core, I'm dedicated by Mappy company to help on that part.

1ec5:
SpriteImage assumes that images can only have integral width and height. I attempted to introduce the concept of non-integral width and height in #3008, but it turns out that we need to refactor more of mbgl to always work with pixel dimensions (which would be integral regardless) instead of point dimensions. We’re tracking this work in #3164.

/cc @jfirebaugh

1ec5:
Once #3561 lands, `SpriteImage` will be expressed entirely in terms of premultiplied pixels (backing pixels in Cocoa terminology). That’ll clear the way for this PR.

1ec5:
@jfirebaugh pointed out that we should take better advantage of the GL style system for the center offset functionality:

> If we want to offer this functionality, it should use the `icon-offset` style property. But in order to control that per-annotation, we need either data-driven styles or control over annotation grouping into layers.

The callout offset functionality doesn’t fit into the style system, and it’s implemented entirely within the iOS SDK, so that isn’t a concern. But I think I’d still like to see it expressed as a `calloutAnchorEdge` property of type UIRectEdge. (We already have support for aligning the callout view based on the image’s alignment insets, in case you need finer-grained control over the anchor’s placement.)

RomainQuidet:
I updated the pull request to work on v3.2.0

1ec5:
`centerOffset` was added to MGLAnnotationView in #5062. Still waiting on the runtime styling API before tackling this for MGLAnnotationImage.

PR #3223, opened 2015-12-08
[core] Ensure GL state is set to necessary values in more places

jfirebaugh:
:eyes: @brunoabinader


jfirebaugh:
Looks like the `check()` mechanism here produces too many false positives. There are cases where `Painter::renderFill` for example is a no-op: it doesn't set any state explicitly, but it also doesn't call any GL draw methods. If we want a robust check it would need to happen just before calls to `glDrawElements`, `glDrawArrays`, etc.

brunoabinader:
:+1:

Issue #3179, opened 2015-12-03
Daniel J H's Code Review: Master Ticket


daniel-j-h:
I'm in Berlin this week, where @joto @kkaefer and I are co-working for a couple of days, exchanging knowledge, tricks and sharing the pain. Today was mapbox-gl-native day for me.

The following is what I found looking over the code base (master@ e5f136a). I started with the platform specific code and could finish that in a handful of hours. There's more to come, as soon as I can spare some time and @mapbox/directions let's me do this :P
- Whole codebase:
  - [x] throw std::runtime_error{"Ooops"}; instead of throw new std::runtime_error{"Oops"};
  - [x] mark all classes and structs final where there is no virtual dtor (no intend of polymorphism) to avoid UB and help compiler in devirtualization
  - [ ] no need to mark member functions as inline when they are defined inside the class definition; maybe grep through all header for inline and check manually
  - [x] mark all operator bool explicit, otherwise you will experience funny implicit conversion issues
  - [x] catch exceptions by const ref
- In platform/android/native_map_view.cpp
  - [ ] configList is std::list, but you only push_back and then even .sort(). Use a std::vector, use .emplace_back() with the tuple's arguments and then std::sort that. Or even better, don't sort it at all, since you want the best => std::min_element. By the way, is this really the best, sorting the tuples lexicographically?
  - [ ] the ConfigProperties type is std::tuple with 6 types, some of which are the same: just make a small local type for that (e.g. struct with named member attributes)
- In platform/node/src/util/async_queue.hpp:
  - [ ] do as little as possible under the locks: std::make_unique(std::move(data)); (without specifying the type, after all type inference is why you are using make_unique) outside the lock, then lock and .push(std::move(uptr)); under the lock
  - [ ] process() spins on empty queue, contention on the lock, better: std::condition_variable
  - [ ] instead of storing the callback in a std::function that allocates memory, better template it typename Callback, then take it by value Callback callback, and store it in a member attribute of type Callback. This does not allocate memory, and is likely to get inlined when passing a lambda
  - [ ] the loop in process() should destroy the item after the callback ran; also, passing a T& to the callback is strange, when you really transfer ownership; better: callback takes a T by value, and then on the call site std::move the *item into the callback
  - [ ] maybe refactor with uv.hpp abstraction?
- In platform/node/src/node_log.cpp:
[...]

PR #3226, opened 2015-12-08
node code review cleanup

mikemorris:
Refs #3179

- In platform/node/src/util/async_queue.hpp:
  - Punting on this per https://github.com/mapbox/mapbox-gl-native/issues/3179#issuecomment-162964129

- In platform/node/src/node_log.cpp:
  - [x] NodeLogOserver's constructor: std::move arguments into their sink, e.g. the std::string to avoic a potential copy (probably everywhere in codebase)

- In platform/node/src/node_map.cpp:
  - [x] return by const value does not make sense, probably in the whole code base --- there should be compiler warnings for that
  - [x] NodeMap::ParseOptions returns a std::unique_ptr<RenderOptions>: this does not make sense, better: just return a RenderOptions type by value

- In platform/default/headless_display.cpp
  - [x] return after throw is dead code

- In platform/default/headless_view.cpp
  - [x] empty line at top, argh my ocd is killing me
  - [x] move the shared_ptrs into their sink in the constructors, otherwise you have an additional refcount increment and decrement
  - [x] isActive can be const

    ```
    error: 'const' type qualifier on return type has no effect
    ```
  - [ ] functions in the HeadlessView destructor can throw, throwing from destructor is bad: wrap in try-catch, log or exit on exception; I would check the whole codebase for that. I think if you mark them noexcept, the standard guarentees for std::terminate to be called; at least it's not UB.
    - Not quite sure what can throw or how to best implement this...

/cc @daniel-j-h

### DIFF
--- a/platform/default/headless_view.cpp
+++ b/platform/default/headless_view.cpp
@@ -1,4 +1,3 @@
-
 #include <mbgl/platform/default/headless_view.hpp>
 #include <mbgl/platform/default/headless_display.hpp>
 #include <mbgl/platform/log.hpp>
@@ -108,7 +107,7 @@ void HeadlessView::createContext() {
 #endif
 }
 
-bool HeadlessView::isActive() {
+bool HeadlessView::isActive() const {
     return std::this_thread::get_id() == thread;
 }
 
--- a/include/mbgl/platform/default/headless_view.hpp
+++ b/include/mbgl/platform/default/headless_view.hpp
@@ -48,7 +48,7 @@ private:
     void createContext();
     void loadExtensions();
     void clearBuffers();
-    bool isActive();
+    bool isActive() const;
 
 private:
     std::shared_ptr<HeadlessDisplay> display;
--- a/platform/default/headless_display.cpp
+++ b/platform/default/headless_display.cpp
@@ -24,11 +24,9 @@ HeadlessDisplay::HeadlessDisplay() {
     CGLError error = CGLChoosePixelFormat(attributes, &pixelFormat, &num);
     if (error != kCGLNoError) {
         throw std::runtime_error(std::string("Error choosing pixel format:") + CGLErrorString(error) + "\n");
-        return;
     }
     if (num <= 0) {
         throw std::runtime_error("No pixel formats found.");
-        return;
     }
 #endif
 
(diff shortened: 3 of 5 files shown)


platform/default/headless_view.cpp
```diff
@@ -19,7 +18,7 @@
 namespace mbgl {
 
 HeadlessView::HeadlessView(float pixelRatio_, uint16_t width, uint16_t height)
-    : display(std::make_shared<HeadlessDisplay>()), pixelRatio(pixelRatio_) {
+    : display(std::move(std::make_shared<HeadlessDisplay>())), pixelRatio(pixelRatio_) {
```
daniel-j-h: Here you don't have to `move` as `make_shared` already returns a rvalue, and thus will be moved into the `display` variable by default.

daniel-j-h:
@mikemorris

> functions in the HeadlessView destructor can throw, throwing from destructor is bad: wrap in try-catch, log or exit on exception; I would check the whole codebase for that. I think if you mark them noexcept, the standard guarentees for std::terminate to be called; at least it's not UB.
> 
>     Not quite sure what can throw or how to best implement this...

`activate()` for example can throw an exception, take a look at its implementation a few lines above.

So either mark the destructor as `noexcept`, as in `struct S { ~S() noexcept { mayThrow(); } };` and be aware that in the case when `mayThrow` actually throws, `std::terminate` is getting called, shutting down your program the hard way (without stack unwinding, that is without cleaning up by invoking all the destructors up to main).

Or wrap the destructor in a try-catch block, and e.g. only do some logging, and then / or call terminate yourself, as in `struct S { ~S() try { mayThrow(); } catch(...) { logErrAndTerminate(); } };`.

jfirebaugh:
> isActive can be const

@mikemorris This is referring to making the method const, rather than the return value. I added this.

 > std::move arguments into their sink

@daniel-j-h We don't care too much about using `std::move` absolutely everywhere it could be used, only in performance sensitive code.

> I think if you mark them noexcept, the standard guarentees for std::terminate to be called; at least it's not UB.

There is never undefined behavior related to throwing from constructors. If an exception is thrown during stack unwinding, the defined behavior is that `std::terminate` is called. For any of the failure cases in `~HeadlessView`, that's fine.

daniel-j-h:
> I think if you mark them noexcept, the standard guarentees for std::terminate to be called; at least it's not UB.
> 
> There is never undefined behavior related to throwing from constructors. If an exception is thrown during stack unwinding, the defined behavior is that std::terminate is called. For any of the failure cases in ~HeadlessView, that's fine.

Throwing from constructors is fine and encouraged to signal failed construction.
Throwing from destructors is dangerous and can lead to undefined behavior.

PR #3230, opened 2015-12-09
Enable Android test release signing

niedzielski:
Configure release signing for testing release builds. This enables
testing of the ProGuard'd apk and the installRelease Gradle target.


zugaldia:
@tobrun Could you review this? It's related to https://github.com/mapbox/mapbox-gl-native/issues/7608 that you've got ideas for.

tobrun:
Thank you for the contribution but after reviewing the PR I'm not so sure about the properties setup. I think something as http://stackoverflow.com/a/21020469/1281750 is more straightforward. Closing.

PR #3237, opened 2015-12-09
Decouple Location Provider

### LINKED ISSUE #2954: Decouple MapView from LOST
Currently MapView takes care of setting up LOST to retrieve location updates and handles things such as pausing updates when the app gets backgrounded.

Google Maps takes a different approach where you implement a `LocationSource` interface to feed in location data to the my location layer. They provide an implementation for Google Plays navigation library, but you could just as easily plug in LOST.

I could not see a benefit to this approach before. In fact even Google Maps used to take care of location data for you, but they have deprecated that API.

The major reason I can see for needing to separate this is the Android M location permissions model. See #2100. Basically only the app has control and insight into whether it has been granted the needed permissions, or if they have been revoked while the app was in the background.

We blindly assume you can resume GPS after return from background which is fine for pre-Android M. We also save the GPS on/off state in MapView's `onSaveInstanceState` which results in a crash if the user revokes a previously given permission.

Instead the app needs to be in charge of saving GPS on/off state and pausing/resuming the location provider. To do this will require us to add a `LocationSource` interface.

/cc @bleege @tobrun @zugaldia

bleege:
Closes #2954

### DIFF
--- a/android/MapboxGLAndroidSDK/src/main/java/com/mapbox/mapboxsdk/location/LocationListener.java
+++ b/android/MapboxGLAndroidSDK/src/main/java/com/mapbox/mapboxsdk/location/LocationListener.java
@@ -1,0 +1,13 @@
+package com.mapbox.mapboxsdk.location;
+
+import android.location.Location;
+
+public interface LocationListener {
+
+    /**
+     * Callback method for receiving location updates from LocationServices.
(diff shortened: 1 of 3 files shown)


bleege:
Bitrise approved, Rebased, and Merged.  :boom:

PR #2161, opened 2015-12-10
Support GeoJSON sources

kkaefer:
[GeoJSON sources](https://www.mapbox.com/mapbox-gl-style-spec/#geojson) are a documented part of the Mapbox GL style specification. mapbox-gl-native should have full and direct support for them, regardless of whatever other annotation/external data APIs it supports.


kkaefer:
- [x] parse GeoJSON specified inline in Source
- [ ] load GeoJSON files specified via URL
- [ ] enable GeoJSON tests for native
- [ ] more comprehensive test coverage

kkaefer:
@jfirebaugh :eyeglasses:

src/mbgl/tile/geojson_tile.cpp
```diff
@@ -0,0 +1,134 @@
...
+
+GeoJSONTileMonitor::GeoJSONTileMonitor(mapbox::geojsonvt::GeoJSONVT* geojsonvt_, const TileID& id)
+    : tileID(id), geojsonvt(geojsonvt_) {
+}
+
+GeoJSONTileMonitor::~GeoJSONTileMonitor() = default;
+
+// A monitor can have its GeoJSONVT object swapped out (e.g. when loading a new GeoJSON file).
+// In that case, we're sending new notifications to all observers.
+void GeoJSONTileMonitor::setGeoJSONVT(mapbox::geojsonvt::GeoJSONVT* vt) {
```
jfirebaugh: Is this actually used?

jfirebaugh:
Looks good. I can see that we're going to want to refactor to a `Source` base + subclasses for the styles API, but that can be longer term too.

> Align Tile/Geometry/Tag classes of GeoJSONVT and Mapbox GL so we can avoid a copy

How big a lift is this?

kkaefer:
>> Align Tile/Geometry/Tag classes of GeoJSONVT and Mapbox GL so we can avoid a copy

> How big a lift is this?

If we do this, we should probably move to a shared header-only module that defines c++ geometry types and is used across our gl native codebase

incanus:
:tada:

PR #3241, opened 2015-12-10
Update geojson-vt options

mourner:
- Fix the code to use the same simplification tolerance
- Remove outdated indexMaxZoom/Points values and rely on default
- Bump buffer 2x to fix bleeding linecaps on thick lines

cc @kkaefer


mourner:
Pushing to https://github.com/mapbox/mapbox-gl-native/pull/3232 instead.

PR #3256, opened 2015-12-11
[android] #3248 - enforce application context in a singleton

### LINKED ISSUE #3248: Enforce usage of ApplicationContext in LocationServices
When instantiating a Singleton with a Context, we should enforce the usage of ApplicationContext:

Replace:
```java
public static LocationServices getLocationServices(@NonNull final Context context) {
        if (instance == null) {
            instance = new LocationServices(context);
        }
        return instance;
    }
```
With:
```java
public static LocationServices getLocationServices(@NonNull final Context context) {
        if (instance == null) {
            instance = new LocationServices(context.getApplicationContext());
        }
        return instance;
    }
```

This will prevent leaking the Activity and corresponding view hierarchy.

tobrun:
closes #3248


tobrun:

PR #3261, opened 2015-12-11
Refactored iOS annotation selection

### LINKED ISSUE #1504: Annotation tap gesture recognizer queries hard-coded region, ignoring image sizes

### LINKED ISSUE #3027: Tapping point annotation does not select it at z0
Tapping a point annotation on iOS does not select it when the map is at zoom level 0.

/cc @incanus

### LINKED ISSUE #3159: MGLMapView should map annotation IDs to annotation metadata

### LINKED ISSUE #3206: Crash when tapping on disabled map annotations
In the Mapbox iOS SDK 3.0.0, tapping a map annotation that has its `MGLAnnotationImage`'s `enabled` property set to `NO` sometimes causes a crash.  Using the default of `YES` never crashes.  FWIW: we mix enabled and disabled annotations on the map, with the disabled ones being much smaller pin images.

Trace:
```objc
* thread #1: tid = 0x3beff1, 0x36750a6e libsystem_platform.dylib`_platform_memmove + 366, queue = 'com.apple.main-thread', stop reason = EXC_BAD_ACCESS (code=1, address=0x18e00000)
    frame #0: 0x36750a6e libsystem_platform.dylib`_platform_memmove + 366
    frame #1: 0x0166d832 Mapbox`-[MGLMapView handleSingleTapGesture:] + 2186
    frame #2: 0x28e6020a UIKit`_UIGestureRecognizerSendTargetActions + 142
    frame #3: 0x28aada82 UIKit`_UIGestureRecognizerSendActions + 170
    frame #4: 0x2893fda4 UIKit`-[UIGestureRecognizer _updateGestureWithEvent:buttonEvent:] + 1004
    frame #5: 0x28e61396 UIKit`___UIGestureRecognizerUpdate_block_invoke898 + 62
    frame #6: 0x28900afe UIKit`_UIGestureRecognizerRemoveObjectsFromArrayAndApplyBlocks + 298
    frame #7: 0x288fd098 UIKit`_UIGestureRecognizerUpdate + 2920
    frame #8: 0x2479c090 CoreFoundation`__CFRUNLOOP_IS_CALLING_OUT_TO_AN_OBSERVER_CALLBACK_FUNCTION__ + 20
    frame #9: 0x2479a386 CoreFoundation`__CFRunLoopDoObservers + 282
    frame #10: 0x2479a7c4 CoreFoundation`__CFRunLoopRun + 972
    frame #11: 0x246ed0d8 CoreFoundation`CFRunLoopRunSpecific + 516
    frame #12: 0x246ececc CoreFoundation`CFRunLoopRunInMode + 108
    frame #13: 0x2da62af8 GraphicsServices`GSEventRunModal + 160
    frame #14: 0x289762dc UIKit`UIApplicationMain + 144
```

1ec5:
This PR backports a lot of annotation-related improvements from the OS X port in #3135, reprising a large chunk of #1496. MGLMapView now maps “annotation tags” (Cocoa-speak for numeric annotation IDs, by analogy with view tags, tooltip tags, etc.) to context objects that pair annotations with their identifiers. Previously, MGLMapView mapped the annotation objects themselves to context objects that contained annotation IDs, but this caused most operations involving annotations to be rather expensive: since mbgl tracks annotations by ID, most of the time you know the ID and need either the annotation or some piece of metadata about it.

Tapping now selects annotations more reliably. Tapping near the top of a large annotation image now selects that annotation. An annotation image’s alignment insets influence how far away the user can tap and still select the annotation. For example, if your annotation image has a large shadow, you can keep that shadow from being tappable by excluding it from the image’s alignment rect. Hit testing now works much like in #3135, with the exception that UIKit doesn’t support hit testing based on pixel opacity.

@RomainQuidet, I was focused on porting existing changes and haven’t worked out how this PR affects #3220. In principle, both changes would be desirable because mine improves interaction with the annotation after it’s added, while yours affects the annotation’s placement in the first place. Let me know if you have any concerns.

The selection code uses `std::remove_if()` instead of `mbgl::util::erase_if()`, fixing #3206 but not the underlying crash. @jfirebaugh, is there a reason to keep `erase_if()` around? It seems to crash when there are a lot of elements and a lot of them are deleted.

The user dot’s callout view is now centered above the user dot. It was previously offset slightly to the left.

More geographic methods have been added internally. `-convertRect:toLatLngBoundsFromView:` returns a world-spanning bounds when the world is wrapping.

Fixes #1504, fixes #3027, fixes #3159, fixes #3206.

/cc @incanus @friedbunny

### DIFF
--- a/platform/ios/MGLMapView.mm
+++ b/platform/ios/MGLMapView.mm
@@ -40,11 +40,12 @@
 
 #import <algorithm>
 #import <cstdlib>
+#import <map>
 #import <unordered_set>
 
 class MBGLView;
+class MGLAnnotationContext;
 
-NSString *const MGLDefaultStyleMarkerSymbolName = @"default_marker";
 NSString *const MGLMapboxSetupDocumentationURLDisplayString = @"mapbox.com/help/first-steps-ios-sdk";
 
 const NSTimeInterval MGLAnimationDuration = 0.3;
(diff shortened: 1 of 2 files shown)


1ec5:
If you tap near a tight cluster of annotations, the annotation directly beneath your finger should be selected first. Previously, the annotations within the query bounds would be selected in order from oldest to newest, which was counterintuitive. The new behavior does sometimes make it more difficult to cycle through the entire cluster, but the previous behavior often got in the way anyways, as your finger would wind up atop a callout view for an annotation farther south. @incanus, since you’re more familiar with the old behavior, let me know if there are problems with the new approach.

platform/ios/MGLMapView.mm
```diff
@@ -2346,9 +2359,159 @@ - (void)removeOverlays:(NS_ARRAY_OF(id <MGLOverlay>) *)overlays
...
+    // This prefix is used to avoid collisions with style-defined sprites in
+    // mbgl, but reusable identifiers are never prefixed.
+    if ([identifier hasPrefix:MGLAnnotationSpritePrefix])
+    {
+        identifier = [identifier substringFromIndex:MGLAnnotationSpritePrefix.length];
+    }
+    return self.annotationImagesByIdentifier[identifier];
+}
+
+/** Returns the straight-line distance between two coordinates. */
```
1ec5: This isn’t quite true. This function returns a value that correlates weakly to the distance between two coordinates. It’s only useful for sorting coordinates by proximity without going through the trouble of converting back and forth between screen and geographic coordinates. I’ll probably just inline it.

PR #3264, opened 2015-12-11
[android] Move files to platform/android subdirectory

jfirebaugh:
Previous omnibus efforts at improving the organization of platform files (#899, #2146) got stalled and then went stale. I'm going to try doing this more piecemeal, platform-by-platform.

/cc @bleege @zugaldia @tobrun


jfirebaugh:
Blocked by #3265.

PR #3268, opened 2015-12-11
[android] #3191 - remove double direction state

### LINKED ISSUE #3191: Double direction state
Currently when the MapView is recreated with a savedstate. 
We are setting the direction twice as noticed by @zugaldia [here](https://github.com/mapbox/mapbox-gl-native/blob/3188-save-tilt/android/MapboxGLAndroidSDK/src/main/java/com/mapbox/mapboxsdk/views/MapView.java#L812).

The correct way to handle this is to take in account the different tracking modes:
 - MyLocationTracking 
 - MyLocationBearing

While restoring the correct direction/location.

tobrun:
closes #3191


tobrun:

PR #3292, opened 2015-12-15
don't use glClear optimization for layers that aren't the bottommost

jfirebaugh:
#2024 results in mis-rendered styles where a background layer is used to implement a semi-transparent overlay on top of other layers.


zugaldia:
@jfirebaugh just confirming that this fixes the issue. Just checked with a problematic style and it looks as expected now. Thanks!

PR #3315, opened 2015-12-15
Only build osxapp on Travis

### LINKED ISSUE #3313: osx build takes 25-30 mins on travis
With #3135, our build time on Travis for OS X went from ~11 minutes to 25 minutes+. OS X builds are now by far the critical path. We need to get those back to a reasonable time -- via #2776 or other means.

cc @1ec5

1ec5:
Building osxapp means building the SDK, so building the SDK independently is wasteful.

Fixes #3313.

/cc @jfirebaugh


1ec5:
Down to 12 min 21 sec.

PR #3316, opened 2015-12-15
Mark debug mask as restorable on OS X

### LINKED ISSUE #3287: osx: resets debug options
If I turn on "Show Tile Boundaries" and "Show Tile Info", they get turned off again when I relaunch the app.

1ec5:
Fixes #3287.

/cc @jfirebaugh


1ec5:
Builds passed before rebasing; merging now.

PR #3095, opened 2015-12-16
Avoid loading style source if its layer(s) are not shown

brunoabinader:
Given https://github.com/mapbox/mapbox-gl-styles/blob/master/styles/satellite-hybrid-v8.json as example of a style containing 2+ sources, if all layers related to a specific source are either disabled (e.g. due to class selections) or invisible, we should avoid _loading_ the style source entirely. This saves processing time and network bandwidth.

When the layers related to a specific style source are re-enabled, the burden of loading the style source could be reduced by using a cache file source.

/cc @mapbox/gl @mapbox/mobile

### DIFF
--- a/src/mbgl/map/source.cpp
+++ b/src/mbgl/map/source.cpp
@@ -61,6 +61,8 @@ void Source::load() {
         return;
     }
 
+    if (req) return;
+
     // URL may either be a TileJSON file, or a GeoJSON file.
     FileSource* fs = util::ThreadContext::getFileSource();
     req = fs->request({ Resource::Kind::Source, info.url }, [this](Response res) {
--- a/src/mbgl/map/source.cpp
+++ b/src/mbgl/map/source.cpp
@@ -39,9 +39,7 @@ Source::Source() {}
 Source::~Source() = default;
 
 bool Source::isLoaded() const {
-    if (!loaded) {
-        return false;
-    }
+    if (!loaded) return false;
 
     for (const auto& tile : tiles) {
         if (tile.second->data->getState() != TileData::State::parsed) {
@@ -50,6 +48,10 @@ bool Source::isLoaded() const {
     }
 
     return true;
+}
+
+bool Source::isLoading() const {
+    return !loaded && req.operator bool();
 }
 
 // Note: This is a separate function that must be called exactly once after creation
--- a/src/mbgl/style/style.cpp
+++ b/src/mbgl/style/style.cpp
@@ -77,7 +77,6 @@ Style::~Style() {
 
 void Style::addSource(std::unique_ptr<Source> source) {
     source->setObserver(this);
-    source->load();
     sources.emplace_back(std::move(source));
 }
 
@@ -184,11 +183,10 @@ void Style::recalculate(float z) {
         hasPendingTransitions |= layer->recalculate(parameters);
 
         Source* source = getSource(layer->source);
-        if (!source) {
-            continue;
+        if (source && layer->needsRendering()) {
+            source->enabled = true;
+            if (!source->loaded && !source->isLoading()) source->load();
         }
-
-        source->enabled = true;
     }
 }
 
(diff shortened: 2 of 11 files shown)


src/mbgl/style/style_layer.hpp
```diff
@@ -45,6 +45,8 @@ class StyleLayer {
     // Returns true if any paint properties have active transitions.
     virtual bool recalculate(const StyleCalculationParameters&) = 0;
 
+    virtual bool isHidden() const = 0;
```
jfirebaugh: Rename this to `isVisible` and reverse the boolean, for consistency with "visibility" as from the style spec.

jfirebaugh:
Instead of introducing more state, can we use `Source::enabled` for this?

brunoabinader:
> Instead of introducing more state, can we use Source::enabled for this?

Indeed, _enabling_ the source only if the layer is visible in `Style::recalculate()` is nicer than comparing strings in `Style::loadSources()`. However, we still require `Source::isLoading()` to avoid double requests and `StyleLayer::isVisible()` because each StyleLayer-derived class calculates its visibility in a different way. I'll update the code accordingly.

jfirebaugh:
There are some corner cases where a layer with a zero-opacity color still renders something, e.g. symbols with halo but transparent fill, background-pattern, outline-only fill layers. Can we limit the check to `opacity > 0`? Also we should add checks of the `*-visibility` layout property, which is the most common way to hide a layer completely.

jfirebaugh:
Can we make `layer.isVisible()` equivalent to `layer.passes != RenderPass::None`?

src/mbgl/map/source.cpp
```diff
@@ -52,6 +52,14 @@ bool Source::isLoaded() const {
     return true;
 }
 
+bool Source::isLoading() const {
+    if (loaded) {
+        return false;
+    }
+
+    return req.operator bool();
```
jfirebaugh: Should be a one liner: `return !loaded && req.operator bool();`

brunoabinader:
> Can we limit the check to opacity > 0?
> Also we should add checks of the *-visibility layout property, which is the most common way to hide a layer completely.

This is used in `Style::recalculate()` now:
```
bool StyleLayer::needsRendering() const {
    return passes != RenderPass::None && visibility != VisibilityType::None;
}
```

I had to fix/modify some style unit tests to adapt to the new scheme for source loading - instead of immediately load sources upon style JSON parsing, we wait to check whether a layer actually needs it.

src/mbgl/map/source.cpp
```diff
@@ -39,9 +39,7 @@ Source::Source() {}
 Source::~Source() = default;
 
 bool Source::isLoaded() const {
-    if (!loaded) {
```
jfirebaugh: Is this change necessary? I would expect `Source:isLoaded()` to return true in the case that at some zN, it was enabled and loaded, and then at zM the only layer that used it becomes hidden and the source was therefore disabled.

src/mbgl/style/style.cpp
```diff
@@ -66,6 +66,14 @@ void Style::setJSON(const std::string& json, const std::string&) {
     loaded = true;
 }
 
+void Style::loadSources() {
+    for (auto& source : sources) {
+        if (source->loaded) continue;
```
jfirebaugh: I think this is clearer if the conditions are all in the same direction:

```
if (source->enabled && !source->loaded && !source->isLoading()) {
    source->load();
}
```

src/mbgl/style/style.cpp
```diff
@@ -166,6 +173,8 @@ void Style::cascade() {
     for (const auto& layer : layers) {
         layer->cascade(parameters);
     }
+
+    loadSources();
```
jfirebaugh: I think `loadSources()` is not necessary here -- `cascade` doesn't change `source->enabled` so it can never trigger a source being needed to load.

src/mbgl/map/source.cpp
brunoabinader: Agreed about `enabled`, but I believe we also should check if there is a pending request, which indicates the source is in loading state.

jfirebaugh: Doesn't `req` imply `!loaded` though? `loaded` only gets set to true after all the requests are successful.

brunoabinader: Yes :dart: Good catch, indeed!

PR #3318, opened 2015-12-16
Map::isFullyLoaded should be called Map::isLoaded

brunoabinader:
This matches `Style`, `Source` and GL JS.

/cc @jfirebaugh


jfirebaugh:
It looks like this changes the iOS public API. The backward incompatibility there is probably not worth the naming consistency.

cc @1ec5

1ec5:
`-[MGLMapView isFullyLoaded]` is actually a private method that the ios-bench application redeclares for its own use in waiting for assets to load. It’s a hack, and I don’t think this change would impact iOS SDK versioning at all.

@kkaefer, can MBXBenchViewController make use of `-[MGLMapViewDelegate mapViewDidFinishRenderingMap:fullyRendered:]` instead?

brunoabinader:
Closing due to ongoing terminology discussions.

PR #3146, opened 2015-11-30
Allow annotation images to be updated

RomainQuidet:
Add possibility to delete / update a sprite in order to update an AnnotationImage

#2210
Fixes #2517

PR #3320, opened 2015-12-16
Fix annotation icon replacement

1ec5:
Another pass at #3146, including a unit test. `Map::removeAnnotationIcon()` was calling itself recursively instead of `MapContext::removeAnnotationIcon()`, and an iterator in `SpriteAtlas::updateDirty()` needs to be incremented before the current element is removed from the map. Thanks to @jfirebaugh for spotting these issues.

/cc @RomainQuidet


src/mbgl/sprite/sprite_atlas.cpp
```diff
@@ -189,11 +189,10 @@ void SpriteAtlas::updateDirty() {
             holder.texture = spriteIterator->second;
             if (holder.texture != nullptr) {
                 copy(holder, imageIterator->first.second);
+                ++imageIterator;
             } else {
-                images.erase(imageIterator);
+                images.erase(imageIterator++);
```
jfirebaugh: This should be prefix increment -- `images.erase(++imageIterator)`. (For anyone who doesn't know: the reason is that `std::map::erase` invalidates any iterators pointing to the erased element. Prefix `++` both returns the _original_ value for passing to `erase`, and increments just before the call. This is both very subtle _and_ a standard C++ idiom in this situation.)

jfirebaugh: And of course, after writing that I realized I had it backwards. The current code is correct!

RomainQuidet:
Thanks for that, I'll try to add unit tests next time and better double check after pull request update

PR #3323, opened 2015-12-16
Fix off-by-one error due to merge

1ec5:
This needs to be converted into a giant switch statement, but that wouldn’t prevent issues like this cropping up in a merge anyways.

/cc @jfirebaugh


jfirebaugh:
:+1:

1ec5:
The conversion to a switch statement is included in #3220.

1ec5:
I also reverted 081e4e44516937a6139fad5823dd5ec3200cc343 and folded in a workaround for the appledoc warnings.

PR #3328, opened 2015-12-16
Move platform scripts into platforms

jfirebaugh:
Continuing a piecemeal approach to #899 / #2146.

cc @kkaefer


test/ios/OHHTTPStubs/Rakefile
```diff
@@ -1,13 +1,13 @@
 # Build & test OHHTTPStubs lib from the CLI
 
 desc 'Build an iOS scheme'
-task :ios, [:scheme, :ios_version, :action] do |_,args|
+task :scripts, [:scheme, :ios_version, :action] do |_,args|
```
jfirebaugh: Not sure how this file got changed. Will revert.

PR #3337, opened 2015-12-17
[osx] 'make compdb' is failing

brunoabinader:
Error message:
```
ninja: error: obj/gyp/osxsdk.ninja:97: bad $-escape (literal $ must be written as $$)
    @rpath/Mapbox.framework/Versions/A/Mapbox -Wl,-rpath,$(inherited) $
                                                         ^ near here
```

### DIFF
--- a/platform/osx/app/mapboxgl-app.gypi
+++ b/platform/osx/app/mapboxgl-app.gypi
@@ -37,7 +37,7 @@
         'CLANG_ENABLE_OBJC_ARC': 'YES',
         'INFOPLIST_FILE': '../platform/osx/app/Info.plist',
         'LD_RUNPATH_SEARCH_PATHS': [
-          '$(inherited)',
+          '\${inherited}',
           '@executable_path/../Frameworks',
         ],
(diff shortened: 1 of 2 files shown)


platform/osx/app/mapboxgl-app.gypi
```diff
@@ -60,7 +60,7 @@
       
       'copies': [
         {
-          'destination': '<(PRODUCT_DIR)/$(FRAMEWORKS_FOLDER_PATH)',
+          'destination': '<(PRODUCT_DIR)/\${FRAMEWORKS_FOLDER_PATH}',
```
1ec5: This is incompatible with Xcode. The destination path becomes “\${FRAMEWORKS_FOLDER_PATH}”, and osxapp crashes on launch (when not launched through the debugger).

1ec5: To reproduce, run `make run-xosx`.

1ec5: Fixed in #3339.

1ec5:
What does `make compdb` do? Does it really need to build the OS X demo application?

Issue #2734, opened 2015-10-22
iOS testing failures around location permissions


friedbunny:
We had hangs on Travis during tests because the location permission dialog was going untapped (since #2612), which we addressed in #2645. This was tweaked somewhat in #2685 to speed things up (as waiting for the permission dialog takes 10 seconds).

There have been some different failures since #2685:

```
Test Case '-[MapViewTests testDelegatesStartStopLocatingUser]' started.
2015-10-22 13:47:54.330 Mapbox GL Tests[3963:13037] _BSMachError: (os/kern) invalid capability (20)
2015-10-22 13:47:54.334 Mapbox GL Tests[3963:13093] _BSMachError: (os/kern) invalid name (15)
2015-10-22 13:47:54.385 Mapbox GL Tests[3963:13037] _BSMachError: (os/kern) invalid capability (20)
2015-10-22 13:47:57.477 Mapbox GL Tests[3963:13037] tap point is not within the bounds of the screen
/Users/vagrant/git/test/ios/KIF/Classes/UIAutomationHelper.m:61: error: -[MapViewTests testDelegatesStartStopLocatingUser] : failed: caught "UIAExceptionBadPoint", "tap point is not within the bounds of the screen"
(
	0   CoreFoundation                      0x0000000105e4bf65 __exceptionPreprocess + 165
	1   libobjc.A.dylib                     0x00000001058c3deb objc_exception_throw + 48
	2   CoreFoundation                      0x0000000105e4be9d +[NSException raise:format:] + 205
	3   UIAutomation                        0x000000011c0992a3 -[UIATarget _tapOptionalObject:tapCount:touchCount:tapOffset:] + 466
	4   UIAutomation                        0x000000011c087f4b -[UIAElement tap] + 319
	5   Test Bundle                         0x00000001155d2835 -[UIAutomationHelper acknowledgeSystemAlert] + 203
	6   Test Bundle                         0x00000001155d26e4 +[UIAutomationHelper acknowledgeSystemAlert] + 48
	7   Test Bundle                         0x00000001155b739c -[MapViewTests approveLocationIfNeeded] + 72
	8   Test Bundle                         0x00000001155de186 -[KIFSystemTestActor waitForNotificationName:object:whileExecutingBlock:complete:] + 360
	9   Test Bundle                         0x00000001155ddfea -[KIFSystemTestActor waitForNotificationName:object:whileExecutingBlock:] + 74
[...]

PR #3341, opened 2015-12-18
[core] Fix background-pattern origin

jfirebaugh:
Regressed in 6e41664cb033ee5edf6ae5ac66ed518d9f0d1f89. This restores it to the same behavior as gl-js, backed by test-suite tests.


jfirebaugh:
Bitrise failure is https://github.com/mapbox/mapbox-gl-native/issues/2734.

Issue #3251, opened 2015-12-11
Directions Example In TestApp


bleege:
To help people better self serve let's build a simple Directions activity in the TestApp.  All it needs to do is draw markers for the starting and stopping points and a polyline for the route.  The data can be cached JSON from an actual request to [Mapbox Directions](https://www.mapbox.com/developers/api/directions/).

/cc @zugaldia @tobrun @mapbox/support

tobrun:
I will be using the example [JSON file](https://api.mapbox.com/v4/directions/mapbox.driving/-122.42,37.78;-77.03,38.91.json?access_token=pk.eyJ1IjoibnVyYm90IiwiYSI6ImNpaG5id2diODBvemF0cGx6ZDRnbzdmbTUifQ.vUB5Wvu6A1gVEFn039NWIQ) as shown in the documentation of Directions.
Will add it locally as an asset so we are independent of the API (could break due to api key)

tobrun:
Current progress:

![device-2015-12-18-144327](https://cloud.githubusercontent.com/assets/2151639/11898112/baebccf6-a595-11e5-9a53-de1e2cf9f643.png)

tobrun:
Running into the problem of not able to use VectorDrawable with SpriteFactory,
Going to PR current progress, will update after SpriteFactory issue is resolved.

@bleege 
 - could you have :eyes: on current progress? 
 - is this what you envisioned when creating the ticket?

zugaldia:
Isn't that route a bit unrealistic? Wouldn't it make more sense something
more local to a city where you can see the (Mapbox) streets?

tobrun:
@zugaldia 
Yeah maybe it is.. but it is the default one shown as an example on the directions page.
Any suggestion for another? :)

tobrun:
I'm currently waiting for #3350 to be merged before being able to merge this

bleege:
> Isn't that route a bit unrealistic?

Agreed.  This needs to be **a simple example of how to drive from Pt A to Pt B in a US city**.  It should have Markers at the starting and stopping points with a Polyline showing the route.  Let's use the following **to go from Pike Place Market to SAFECO Field in Seattle**.

**Directions URL:**

```
https://api.mapbox.com/v4/directions/mapbox.driving/-122.34067,47.60869;-122.33420,47.59028.json?access_token=<YOURACCESSTOKENHERE>
```

**JSON Response From Directions URL**

```json
{

    "origin": 

{

[...]

PR #3350, opened 2015-12-18
[android] #3107 - add support non BitmapDrawables to SpriteFactory

### LINKED ISSUE #3107: SpriteFactory.fromResource returns null for XML drawables
Code that currently returns a `null` sprite for xml drawable `ic_flag`:
```java
Sprite sprite = new SpriteFactory(mapView).fromResource(R.drawable.ic_flag);
```
Step into that code:

```java
public Sprite fromResource(int resourceId) {
   Bitmap bitmap = BitmapFactory.decodeResource(mMapView.getResources(), resourceId);
   return fromBitmap(bitmap);
}
```
It seems that decodeResource doesn't work for drawables defined in xml. It is trying to directly load the referenced file instead of loading the files referenced by the xml. See this [SO-post](http://stackoverflow.com/questions/8370316/android-decoderesource-returns-null-for-xml-bitmap-api-level-7).

Note: `ic_flag` is a vector drawable generated with Android Gradle Plugin 1.5.0

tobrun:
allow other type of drawable to be used from a resource drawable other than BitmapDrawables.
Closes #3107


tobrun:
CI approved going to merge, needed for #3251

Issue #584, opened 2014-11-10
Offline maps


ljbade:
It is a personal goal to have the ability to view maps without needing an Internet connection.

Very useful for activities such as bushwalking our mountain biking.

Plus we know there are others interested in this feature too.

hallahan:
Whatever happened with this issue? Is there now MBTiles support for Android? Is this done natively? This feature is a requirement for me to use this library as an Android app.

incanus:
There is not MBTiles support yet in this project. Offline support will be coming in a future release; pretty sure this will include MBTiles.

hallahan:
Darn, is there a branch this is being tried with? Is this going to be native C++ for all builds? For me, this is the most important feature I need. I'm interested in helping.

incanus:
There's no movement on it yet, AFAIK. Right now we are sprinting towards a series of iOS beta releases, after which we'll focus on Android and a broader feature set. I would expect it to be in C++, yes, with Java/ObjC bindings. 

Happy to take work in a branch, but it just likely won't get many eyeballs during March due to priorities elsewhere.

kristfal:
Hey,

I know this is a large feature that will be implemented later, but I'm very curious regd. the performance differences, so wondering:

Is there currently a more elegant way to load tiles bundled with an app besides doing [this](https://github.com/mapbox/mapbox-gl-native/blob/7827d4954cea5b4fcfce3f245aae2e8c66e0ccb2/test/fixtures/api/water.json) and then load individual .pbfs referenced through [a tilejson](https://github.com/mapbox/mapbox-gl-native/blob/7827d4954cea5b4fcfce3f245aae2e8c66e0ccb2/test/fixtures/tiles/streets.json)?

xwz:
I'm trying to have a go at providing a mbtiles source for offline use. There is some difference in what the sdk downloads from the tile server hosted by mapbox and that downloaded from the same url from the command line using curl. 

E.g. for the vector tile at `v4/mapbox.mapbox-streets-v6/14/15074/9831.vector.pbf` the sdk downloaded file 13405 bytes, the first few bytes are: (setting a break point some where in `http_request_nsurl.m`)

~~~
0000000: 1a83 080a 076c 616e 6475 7365 125f 0800  .....landuse._..
0000010: 1206 0000 0101 0202 1803 2251 0927 f615  .........."Q.'..
~~~

Where as download via the command line with curl it is 7842 bytes:
~~~
[...]

PR #2939, opened 2015-12-18
offline-capable file source

jfirebaugh:
First movements towards https://github.com/mapbox/mapbox-gl-native/issues/584. 

My current thinking here is to have `ThreadContext` return an array of ordered file sources, not just a lone one with which the map/map context/thread context was created. For example, you'd have an offline source, then the `DefaultFileSource`, and a fallback would occur if the offline one didn't have a resource (either by circumstance or design). 

This should work during e.g. `VectorTileMonitor::monitorTile()` request kickoffs as they query the `util::ThreadContext::getFileSource()` API for each request, so that means the array could be mutated at various points during the lifecycle and such an API could return the latest iteration of the array. 

Currently working on a basic mockup file source, then some structure to test this design out as the first steps. 

I do think we should keep offline sources clear of `DefaultFileSource` and its associated (SQLite-backed) cache complexity.

### DIFF
--- a/carryover.py
+++ b/carryover.py
@@ -1,0 +1,59 @@
+#!/usr/bin/env python
+
+import sqlite3
+import sys
+import zlib
+
+in_db = sqlite3.connect('cache.db')
+
+with in_db:
+    in_db.row_factory = sqlite3.Row
+    in_cur = in_db.cursor()
+    in_cur.execute('select url, data, kind, compressed from http_cache where status = 1 order by kind asc')
+    in_rows = in_cur.fetchall()
+    out_db = sqlite3.connect('cache.mbtiles')
+    with out_db:
+        out_cur = out_db.cursor()
+        out_cur.executescript("""
+            drop table if exists tiles;
+            create table tiles (zoom_level integer, tile_column integer, tile_row integer, tile_data blob);
+            create unique index tile_index on tiles (zoom_level, tile_column, tile_row);
+            drop table if exists metadata;
+            create table metadata (name text,value text);
+            create unique index name on metadata (name);
+            """)
+        for in_row in in_rows:
+            kind = in_row['kind']
+            data = in_row['data']
+            if in_row['compressed'] == 1:
+                data = zlib.decompress(data)
+            if kind == 3:
+                parts = in_row['url'].split('/')
+                z = int(parts[4])
+                x = int(parts[5])
+                y = int(parts[6].split('.')[0])
+                y = pow(2, z) - y - 1
+                out_cur.execute('insert into tiles (zoom_level, tile_column, tile_row, tile_data) ' \
+                                'values (?, ?, ?, ?)', (z, x, y, sqlite3.Binary(data)))
+                print 'inserted tile %s,%s,%s (%s bytes)' % (z, x, y, len(data))
+            elif kind in (2, 4, 5):
+                if kind == 2:
+                    prefix = 'gl_source'
+                elif kind == 4:
+                    prefix = 'gl_glyph'
+                else:
+                    prefix = 'gl_sprite_image'
+                out_cur.execute('insert into metadata (name, value) values (?, ?)', 
+                    (prefix + '_' + in_row['url'], sqlite3.Binary(data)))
+                label = prefix.replace('gl_', '').replace('_', ' ')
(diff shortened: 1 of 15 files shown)


jfirebaugh:
I updated this to latest `master`. Going to start looking at the core API more closely.

mb12:
@incanus Is there any plan to add routing and search to the offline roadmap? Will it be developed in the open like gl native client?  Its one big item that's missing from mapbox sdk.

twbell:
@mb12 -- ancillary offline services are on our radar; focusing on getting the basics in-play first.

jfirebaugh:
The architectural changes needed here are:

* Rename `DefaultFileSource` ⇢ `OnlineFileSource`
* Write a new `DefaultFileSource` class that implements the behavior of first checking the `OfflineFileSource`, then falling back to `OnlineFileSource` internally.
* Remove `handlesResource`. Instead of asking `OfflineFileSource` if it can handle a request, and then making the actual request (2x the queries), just make the actual request and handle a "not found" response by falling back to online.
* Remove the `offlineMapPath` option; use a fixed path instead.
* Revert the changes to `MGLMapView.{h,mm}`, `map.{hpp,cpp}`, `map_context.{hpp,cpp}`, `raster_tile_data.cpp`, `source.cpp`, `vector_tile.cpp`, `sprite_store.cpp`, `glyph_pbf.cpp`, and `thread_context.{hpp,cpp}`. Given the above modifications, none of the APIs used or exposed by these files will change.

incanus:
@jfirebaugh Keep in mind here the use case of multiple offline documents, the set of which can differ at runtime. A concrete use case for this is a parks or golf courses app that uses a paywall to give access to individual sections of offline area. At runtime, the dev just wants to pass in the supported documents/regions (or perhaps alter them later as new ones becomes available, or maybe even just say "use all of the downloaded regions"), but does not want to have to reinitialize or reconfigure the map view (perhaps jarringly visually) just to change the attached regions. This would be most evident when the map viewport is made to change to a new region and associated backend configuration / singular referenced file path would have to change in response, possibly delaying actual map rendering as file sources are reconfigured. 

My original goal here was an API that allows management of a _set_ of `OfflineFileSource` objects (add/remove/insert/replace), ordered according to the caller's preference, each backed by their own filesystem path for the backing document. Then, when a map view zooms to a new region, it can quickly check (linear _O(n)_ complexity) across the attached documents (especially if, [per the spec suggestion for performance](https://github.com/mapbox/mbtiles-spec/blob/master/1.2/spec.md#content), the `bounds` metadata field is used) to which one(s) can provide tiles and other assets for the current viewport. 

You can see a successful model of this in the old MBXMapKit idea of multiple `MBXRasterTileOverlay` layers on a map view at once, each of which controlled their own region and answered for it when relevant to the current viewport, or in the iOS SDK 1.x multiple tile source API, which did the same. The idea in _this_ SDK was to use what we learned there to finally have an API which supported multiple regions/documents configured at init time and just did the right thing, showing as much as it could based on what had been previously cached for offline use, while keeping visual ordering hierarchy in the map unified style across multiple style sources. 

Keep in mind also that while we haven't explicitly promised it, we may also want to allow for arbitrary MBTiles file path hookup (https://github.com/mapbox/mapbox-gl-native/issues/3053) alongside the internally-managed and SDK-downloaded documents. 

I think we need a multiple offline file source API for these reasons, and as the evolution of what we've learned through use and feedback of our past offline mobile APIs. 

With regard to 2x the queries, yes, we could optimize this a bit; however, using the (optional) `bounds` metadata field for a document is a single query, and so is just attempting to pull a tile; we may find that the tile query is less performant given the potential larger database table size and index complexity — not sure yet. This could be measured.

jfirebaugh:
Yes, good to keep those use cases in mind. However, I believe the management of "sets of offline resources" should be handled internally to `DefaultFileSource` and/or `OfflineFileSource`, and the `FileSource` and `ThreadContext` interfaces should not change.

incanus:
Yep, agree. My read on this was instead of obtaining `fs` synchronously and kicking off the request asynchronously, the whole thing should kick off asynchronously and the right file source(s) would be determined internally.

jfirebaugh:
Build issues are now fixed, down to test failures.

jfirebaugh:
⇢ #3715

Issue #3294, opened 2015-12-15
`make linux` should work on OSX


brunoabinader:
OSX also supports GLFW, which is the base for `linux` build target.

Extra: @kkaefer and @1ec5 spoke about changing that target name to `glfw` if that makes sense.

/cc @tmpsantos

brunoabinader:
`make linux` is working mostly fine in OSX, with exception to these interesting warning messages (might be specific to my environment):
> ld: warning: object file (./Release/libmbgl-cache-sqlite.a(sqlite_cache.o)) was built for newer OSX version (10.10) than being linked (10.9)

1ec5:
That’s similar to mapbox/mapbox-gl-native#1707. In #3135, I changed the minimum OS X deployment target from 10.9 to 10.10. It’s possible that I missed a spot.

brunoabinader:
`make linux` now works fine on OSX.

PR #3356, opened 2015-12-18
[osx] Use framebuffer width/height in GLFW port

brunoabinader:
Framebuffer width/height takes pixel ration in consideration.

Part of #3294.

/cc @tmpsantos


tmpsantos:
:+1: thanks for fixing

PR #3362, opened 2015-12-20
Better zoom level–altitude conversion

### LINKED ISSUE #2266: Inaccurate camera altitude
MGLMapView.camera.altitude doesn’t come close to the value that MKMapView would give at the same zoom level.

ref #2193

### LINKED ISSUE #2456: `MGLMapView setCamera:` does not check if new camera is the same as current camera
`-[MGLMapView setCamera:withDuration:animationTimingFunction:]` will happily pass along the current (unchanged) camera to `_mbglMap->easeTo()`, which eventually sends false region change notifications.

This is especially bothersome when trying to set the camera in one of the region change delegate methods, because any call to a `-[MGLMapView setCamera:]` variant will keep looping but accomplish nothing.

Example:

```objc
#import "ViewController.h"
#import "Mapbox.h"

@interface ViewController () <MGLMapViewDelegate>

@property (nonatomic) MGLMapView *mapView;
@property (nonatomic) MGLMapCamera *camera;

@end

@implementation ViewController

- (void)viewDidLoad {
    [super viewDidLoad];
    
    self.mapView = [[MGLMapView alloc] initWithFrame:self.view.bounds];
    self.mapView.autoresizingMask = UIViewAutoresizingFlexibleWidth | UIViewAutoresizingFlexibleHeight;
    
    self.mapView.delegate = self;
    
    self.camera = [MGLMapCamera cameraLookingAtCenterCoordinate:CLLocationCoordinate2DMake(51.002040, 3.324372) fromDistance:20000 pitch:60 heading:180];
    
    [self.mapView setCamera:self.camera];
    
    [self.view addSubview:self.mapView];
}

- (void)mapView:(MGLMapView *)mapView regionDidChangeAnimated:(BOOL)animated
{
    NSLog(@"regionDidChangeAnimated");
    
    [self.mapView setCamera:self.camera];
//               withDuration:2
//    animationTimingFunction:[CAMediaTimingFunction functionWithName:kCAMediaTimingFunctionEaseInEaseOut]];
}

@end
```

Which results in this strangely mesmerizing effect:

camera-pan-no-animation 1 44 58 pm

/cc @incanus @1ec5

### LINKED ISSUE #3286: osx: sometimes resets view to null island
Launch OS X, maximize the window on a largeish monitor, zoom all the way out, quit. Relaunch the app.

Expected behavior: shows the view you were looking at when you quit.
Actual behavior: resets to null island at z4.

1ec5:
Greatly simplified conversions between zoom level and altitude. Adjusted the (apparently vertical) field of view to more closely match MapKit. Conversions are now nearly identical to MapKit at all latitudes.

The conversion code is now shared between iOS and OS X. It isn’t in mbgl, because although the conversions are more or less consistent with MapKit, it isn’t clear that they’re consistent with mbgl itself. (mbgl also has its own concept of “altitude” that’s distinct from the conventional meaning of altitude.)

Building off this improvement, MGLMapView now has a variation of `-flyToCamera:…` that accepts a peak altitude parameter.

Fixes #2266, and probably also fixes #3286. Also fixed #2456 along the way.

/cc @friedbunny @adam-mapbox

### DIFF
--- a/platform/darwin/MGLMapCamera.mm
+++ b/platform/darwin/MGLMapCamera.mm
@@ -98,4 +98,28 @@
             NSStringFromClass([self class]), self, _centerCoordinate.latitude, _centerCoordinate.longitude, _altitude, _heading, _pitch];
 }
 
+- (BOOL)isEqual:(id)other
+{
+    if ( ! [other isKindOfClass:[self class]])
+    {
+        return NO;
+    }
+    if (other == self)
+    {
+        return YES;
+    }
+    
+    MGLMapCamera *otherCamera = other;
+    return (_centerCoordinate.latitude == otherCamera.centerCoordinate.latitude
+            && _centerCoordinate.longitude == otherCamera.centerCoordinate.longitude
+            && _altitude == otherCamera.altitude
+            && _pitch == otherCamera.pitch && _heading == otherCamera.heading);
+}
+
+- (NSUInteger)hash
+{
+    return (@(_centerCoordinate.latitude).hash + @(_centerCoordinate.longitude).hash
+            + @(_altitude).hash + @(_pitch).hash + @(_heading).hash);
+}
+
 @end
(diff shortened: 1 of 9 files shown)


platform/darwin/MGLMapCamera.mm
```diff
@@ -98,4 +98,28 @@ - (NSString *)description
...
+    }
+    
+    MGLMapCamera *otherCamera = other;
+    return (_centerCoordinate.latitude == otherCamera.centerCoordinate.latitude
+            && _centerCoordinate.longitude == otherCamera.centerCoordinate.longitude
+            && _altitude == otherCamera.altitude
+            && _pitch == otherCamera.pitch && _heading == otherCamera.heading);
+}
+
+- (NSUInteger)hash
```
friedbunny: Thanks for adding `isEqual:`, will look at adding this to the camera example.

friedbunny: (Grr, commented on the wrong line there.)

On this line, I meant to ask: where do you see `hash` coming in useful?

1ec5: Any class that implements `-isEqual:` [must also implement `-hash`](https://developer.apple.com/library/ios/documentation/Cocoa/Reference/Foundation/Protocols/NSObject_Protocol/#//apple_ref/occ/intfm/NSObject/isEqual:), because equal objects must have the same hash value.

friedbunny: Oooh, `NSObject`. :school:

PR #3370, opened 2015-12-21
AsyncTask.RequestCoalescingMultithreaded test fails on OS X with default ulimits

tmpsantos:
```
[ RUN      ] AsyncTask.RequestCoalescingMultithreaded
libc++abi.dylib: terminating with uncaught exception of type std::runtime_error: Failed to initialize loop.
./scripts/run_tests.sh: line 32: 80042 Abort trap: 6           (core dumped) ${CMD}
```

Default value of `ulimit -n` on OS X is 256. Raising it to 1024 allows the test to pass. Can we rewrite the test so that it doesn't try to open more than 256 files?


jfirebaugh:
:+1: Fixes the issue.

PR #3377, opened 2015-12-22
make osx with Xcode

### LINKED ISSUE #3293: make osx builds OS X framework without Xcode, fails
`make osx` attempts to build the OS X framework using make instead of `xcodebuild`. The framework target relies on Xcode-specific build settings, so it fails to build under make:

```
../../platform/osx/sdk/Mapbox.m:1:9: fatal error: 'Mapbox/Mapbox.h' file not found
#import <Mapbox/Mapbox.h>
        ^
1 error generated.
```

If `make osx` is useful in its own right, it should avoid building the framework and osxapp and limit itself to just the headless parts. Otherwise, we should remove the target because we already have `make ipackage` and `make xosx`.

/cc @brunoabinader @kkaefer

1ec5:
`make osx` now builds with Xcode instead of make, because building the OS X SDK and osxapp requires Xcode-specific build settings as of #3135. `make xosx` and `make nosx` are now aliases to `make osx`.

Fixes #3293.

/cc @kkaefer @jfirebaugh

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -22,11 +22,11 @@ default: ; @printf "You must specify a valid target\n"
 
 ifeq ($(BUILD),osx)
 .PHONY: osx xosx nosx run-osx run-xosx
-osx: ; $(RUN) HOST=osx HOST_VERSION=x86_64 Makefile/osxapp
-xosx: ; $(RUN) HOST=osx HOST_VERSION=x86_64 Xcode/osxapp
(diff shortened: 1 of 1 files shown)


1ec5:
I flubbed the change to `make run-xosx`. Fixed in 7bfb7766187da6e076deb2989f7757b3c8309f3b.

Issue #2231, opened 2015-09-02
Move HTTP caching to Java


ljbade:
Either via OkHTTP's built-in [`Cache`](http://square.github.io/okhttp/javadoc/com/squareup/okhttp/Cache.html) or implementing our current SQLite cache using the Android platform [`SQLiteDatabase`](http://developer.android.com/reference/android/database/sqlite/SQLiteDatabase.html)

Follows on from #823 so we can implement all of `FileSource` in Java. This allows us to drop the native sqlite dependency.

tobrun:
This landed with #2857, closing

jfirebaugh:
This ticket is basically about a native implementation of `SQLiteCache`, or some other means of removing the native SQLite dependency for Android, so it wasn't fixed with #2857. However, given how offline is shaping up I'm skeptical that we'll ever want to do that, so it's fine with me to leave it closed.

PR #3381, opened 2015-12-22
Preparation for OfflineFileSource, which depends on SQLite

jfirebaugh:
* Move OnlineFileSource to platform directory
* Merge SQLiteCache into the platform library

These changes prepare for adding `OfflineFileSource`, which will have a SQLite dependency. Since all platforms (except node) depend on SQLite currently, and we are looking to expand that dependency, I think it makes sense to merge the "cache" library into the platform library.

cc @kkaefer


kkaefer:
Will this still allow us to eventually replace C SQLite with a Java/JNI-based interface, according to https://github.com/mapbox/mapbox-gl-native/issues/2231?

jfirebaugh:
We could take several approaches for #2231:

* Reimplement `SQLiteCache` entirely for Android. In this case we'd rename it to `DefaultFileCache`, and the Android implementation would be written as JNI bindings to some Java cache store object.
* Continue to use the same `SQLiteCache` implementation on all platforms, but reimplement SQLite bindings for Android. In this case the Android implementation of the stuff in `sqlite3.hpp` would be written as JNI bindings to Java SQLite classes. `OfflineFileSource` would use this implementation too.

What I don't want to do is use a platform specific offline store format for Android; i.e. per-platform `OfflineFileSource` implementations. If we take the first approach above, we'll still have to either _also_ do platform-specific SQLite bindings, or just accept a C SQLite dependency.

PR #3386, opened 2015-12-23
[core] Make it possible to limit the size of the SQLite cache

tmpsantos:
We use the number of active pages allocated by SQLite to calculate the size of the cache. The restriction here is that we don't call VACUUM as it is very expensive and usually requires to copy the database which would use twice as much space for a moment.

In order to make more space for new entries as the cache grows, least accessed entries are purged.


platform/default/sqlite_cache.cpp
```diff
@@ -210,17 +414,13 @@ void SQLiteCache::Impl::put(const Resource& resource, std::shared_ptr<const Resp
...
-        if (!schema) {
-            createSchema();
-        }
+        initializeDatabase();
 
         if (!refreshStmt) {
-            refreshStmt = std::make_unique<Statement>( //        1               2
-                db->prepare("UPDATE `http_cache` SET `expires` = ? WHERE `url` = ?"));
+            refreshStmt = std::make_unique<Statement>(
+                db->prepare("UPDATE `http_cache` SET `accessed` = strftime('%s', 'now'), "
```
kkaefer: Why are we storing dates as strings?

kkaefer: We should make sure that we're indeed using integers rather than stringified numbers here. It seems like we do the latter?

tmpsantos: This gets converted to INTEGER at the end of the day and AFAIK is the best know method for getting the current Unix epoch inline in the SQL query. But I do agree that it looks odd and probably less efficient than simply binding the value. I will changed that.

PR #3387, opened 2015-12-23
Add SQLite, zlib linker flags

1ec5:
This fixes iosapp builds.

/cc @jfirebaugh @friedbunny


jfirebaugh:
:+1: Can you also ticket out or PR updating CI so this type of failure is caught?

1ec5:
#3388

friedbunny:
:bow:

PR #3394, opened 2015-12-23
[tests] Rewrite resource loading test

jfirebaugh:
Make it more readable and easier to debug.

cc @brunoabinader

PR #3399, opened 2015-12-23
[tests] Moved lazy source loading tests to ResourceLoading

brunoabinader:
Follow up of #3394, this change moves the lazy source loading tests from `test/style/unused-sources.cpp` into `test/style/resource_loading.cpp` to take advantage of @jfirebaugh's latest test improvements.

/cc @jfirebaugh @tmpsantos


jfirebaugh:
I prefer to keep all assertions (`EXPECT_EQ` etc) in the body of the test, rather than in helper functions. Assertions in helper functions tend to develop complex conditional logic, and when they fail, the cause is harder to track down because (a) the line number points into generic code rather than the specific test that's failing and (b) the expected/actual values that are printed come from somewhere else in the code and must be tracked down.

Write your test helpers like any other function: receive input values and return calculated results. Then assert on the results in the test body.

PR #3400, opened 2015-12-23
[core] Fix build script paths

brunoabinader:
Updated remaining script paths from `scritps/$(HOST)/$(HOST_VERSION)` to `platform/$(HOST)/scripts/$(HOST_VERSION)`.

/cc @jfirebaugh @tmpsantos


brunoabinader:
It seems some targets still use the deprecated `scripts/$(HOST)/$(HOST_VERSION)` path. We're going to check for both in `scripts/main.mk` to make Travis/Bitrise happy until we update all remaining targets scripts.

brunoabinader:
Commit 7bd4745cf10c504a4899a37016e87bce45e51472 is also part of this.

PR #3401, opened 2015-12-23
Various zoom level corrections

### LINKED ISSUE #1755: Changing device orientation causes user dot to wander
Rotating the device between landscape and portrait can cause the user dot to wander all over the world map if you zoom out to the minimum allowed zoom level.

Start out in portrait with the user dot visible, and note your location. Zoom all the way out to the minimum zoom level:

Rotate the device to landscape:

Zoom further out:

Rotate back to portrait:

The user dot has wandered hundreds of miles. Now pan left and right:

The user dot moves relative to the map content, sometimes disappears, and sometimes reappears on the other side of the screen.

The bug goes away as soon as you zoom in or out.

/cc @incanus

1ec5:
I started out trying to fix the OS X SDK’s Zoom Out button, which refused to disable even as you reach the effective minimum zoom level. I wound up fixing a few related issues:

* Two active sources can have different minimum and maximum zoom levels. Always use the nearest available zoom level for the source. `TransformState` has the final say over whether a zoom level is allowed.
* Even if `TransformState::constrainMode` is `HeightOnly`, there isn’t usually a reason to _show_ more than one full revolution of the world at a time. Before this change, osxapp effectively constrained the zoom level based on whether the window was shorter or taller than 512 points. (Maps.app constrains the zoom level even further in standard view, but that’s because its satellite view is projected onto a globe, so it’s more appropriate at low zoom levels.)
* Fixed `Transform::getMinZoom()` to account for the constrained zoom level.

These changes have the side effect of fixing an issue in the iOS SDK, in which the minimum zoom level in portrait can be circumvented by zooming out in landscape then rotating the device. Now the two orientations have consistent minimum zoom levels, which fully fixes #1755.

/cc @brunoabinader @friedbunny @zugaldia

### DIFF
--- a/platform/osx/src/MGLMapView.mm
+++ b/platform/osx/src/MGLMapView.mm
@@ -865,6 +865,12 @@ public:
     _mbglMap->setZoom(zoomLevel, MGLDurationInSeconds(animated ? MGLAnimationDuration : 0));
 }
 
+- (void)zoomBy:(double)zoomDelta animated:(BOOL)animated {
+    [self willChangeValueForKey:@"zoomLevel"];
+    _mbglMap->setZoom(self.zoomLevel + zoomDelta, MGLDurationInSeconds(animated ? MGLAnimationDuration : 0));
+    [self didChangeValueForKey:@"zoomLevel"];
+}
+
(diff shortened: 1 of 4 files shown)


friedbunny:
Tried this out on iOS and did not notice any issues.

PR #3403, opened 2015-12-23
[ios] snap to north if interaction ends within ±7°

### LINKED ISSUE #1064: Compass/bearing snapping fails when also zooming

friedbunny:
If a user pinch-rotates and the gesture ends within ±7° of north, reset to north (0°).

This behavior is the same as in MapKit.

First proposed in #938, implemented in #1059, reverted in #1064. (Closes #1064.)

/cc @1ec5

### DIFF
--- a/platform/ios/src/MGLMapView.mm
+++ b/platform/ios/src/MGLMapView.mm
@@ -3027,6 +3027,8 @@ std::chrono::steady_clock::duration MGLDurationInSeconds(float duration)
 //
 - (void)unrotateIfNeededAnimated:(BOOL)animated
 {
+    double snapTolerance = 7;
+
     // don't worry about it in the midst of pinch or rotate gestures
     //
     if (self.pinch.state  == UIGestureRecognizerStateChanged || self.rotate.state == UIGestureRecognizerStateChanged) return;
@@ -3056,6 +3058,10 @@ std::chrono::steady_clock::duration MGLDurationInSeconds(float duration)
         {
             [self resetNorthAnimated:NO];
         }
+    }
+    else if (self.direction < snapTolerance || self.direction > 360 - snapTolerance)
+    {
+        [self resetNorthAnimated:animated];
     }
 }
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -45,6 +45,7 @@ Known issues:
 - MGLMapCamera’s `altitude` values now match those of MKMapCamera. ([#3362](https://github.com/mapbox/mapbox-gl-native/pull/3362))
 - The user dot’s callout view is now centered above the user dot. It was previously offset slightly to the left. ([#3261](https://github.com/mapbox/mapbox-gl-native/pull/3261))
 - Fixed an issue with small map views not properly fitting annotations within bounds. (#[3407](https://github.com/mapbox/mapbox-gl-native/pull/3407))
+- The map will now snap to north. ([#3403](https://github.com/mapbox/mapbox-gl-native/pull/3403))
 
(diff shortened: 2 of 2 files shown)


friedbunny:
Going to leave this convenient `iosapp` debug code here:

```objc
- (void)mapViewRegionIsChanging:(__unused MGLMapView *)mapView
{
    [self updateDirectionDebug];
}
- (void)mapView:(__unused MGLMapView *)mapView regionDidChangeAnimated:(__unused BOOL)animated
{
    [self updateDirectionDebug];
}

- (void)updateDirectionDebug
{
    if (! _directionDebug)
    {
        _directionDebug = [[UILabel alloc] initWithFrame:CGRectMake(250, 100, 100, 50)];
        [self.view addSubview:_directionDebug];
    }

    _directionDebug.text = [NSString stringWithFormat:@"%.2f", self.mapView.direction];
}
```

1ec5:
Could this be done as part of `-unrotateIfNeededAnimated:`?

friedbunny:
I'm looking into integrating it in `-unrotateIfNeededAnimated:`. That does get called rather frequently...

friedbunny:
I went ahead and integrated the snapping with `-unrotateIfNeededAnimated:`. This will cover more cases and be called more frequently.

I've not copied the delayed animations and disabled user inaction that exists just above in  `-unrotateIfNeededAnimated:` because it does not appear to be necessary.

platform/ios/src/MGLMapView.mm
```diff
@@ -3053,6 +3055,10 @@ - (void)unrotateIfNeededAnimated:(BOOL)animated
             [self resetNorthAnimated:NO];
         }
     }
+    else if (self.compassView.alpha > 0 && (self.direction < snapTolerance || self.direction > 360 - snapTolerance))
```
1ec5: What if the developer hides the compass view? Shouldn’t the map view snap to north regardless?

friedbunny: Good point, will look at changing this.

friedbunny:
Ergh, accidentally included iosapp debug code. Will back that out.

friedbunny:
Rebased and squashed.

PR #3407, opened 2015-12-24
[ios] vary annotation bounds fit padding based on view size

### LINKED ISSUE #3392: "showAnnotations:animated:" bug in small map
Hi

I found bug. Method `showAnnotations:animated:` doesn't work properly on smaller map size.
Steps to reproduce:
- When you have mglmapview sized e.g. 128x128, 
- you create polyline annotation 
- execute showAnnotation method
-> mapview scroll to center of annotation but in such zoom scale it is not visible.

Here is prepared code https://bitbucket.org/sirljan/mapboxtrythings
I would expect to see this:
screen shot 2015-12-23 at 01 37 33
But I see this:
screen shot 2015-12-23 at 01 48 49

To see it in action unzip this video.
[showAnnotationBug.mov.zip](https://github.com/mapbox/mapbox-gl-native/files/70367/showAnnotationBug.mov.zip)

friedbunny:
`-[MGLMapView showAnnotations:animated:]` hardcoded the edge inset to 100 points on all sides, which meant that views with height or width <=200 points would max-zoom into the center point of the bounds, not show the entire bounds.

This fix varies the edge padding based on the map view's frame size — if it's less than the default padding, use a fraction of the map view frame's height/width as inset padding.

Fix #3392, #3110.

/cc @1ec5


friedbunny:
So tiny!

simulator screen shot dec 23 2015 7 57 18 pm

1ec5:
:+1:

PR #3183, opened 2015-12-04
Convert iOS SDK into dynamic framework

1ec5:
`make iproj` now produces a target that pulls together static libraries like core and platform-ios into a true dynamic framework (as opposed to a static library executable stuffed inside a framework bundle). iosapp is pretty much just a regular iOS application that links Mapbox.framework (except for the inclusion of default_styles.hpp).

~~For now, I’ve patched support for code signing files in Copy Files build phases into the ancient checked-in copy of gyp, pending #3178.~~

Dynamic framework distribution relies on a shell script copied from realm/realm-cocoa#2759. This script works around an [App Store bug](http://www.openradar.me/radar?id=6409498411401216).

Fixes #828, fixes #3161, paving the way for #1623. Based on #3135 for OS X.

/cc @incanus @friedbunny

PR #3416, opened 2015-12-25
Reorganize documentation

1ec5:
Rewrote some copy to provide more context and link to more GL-related repos, including GL JS. Put the various SDKs in a table so we can show all three Bitrise bots’ status.

/cc @jfirebaugh @friedbunny


jfirebaugh:
Looks good.

Should we merge the "Targets" list into the table?

1ec5:
I ended up reorganizing all the Markdown documentation in this repository. Previously, even though all the documentation was in one folder, it was a maze to navigate to the right documentation. I tried interlinking the documents, but there ended up being quite a lot of duplication. So instead, all the SDK documentation lives right alongside the SDKs, in platform/, in keeping with #899 #2146. One benefit is that directories like [platform/android/](https://github.com/mapbox/mapbox-gl-native/tree/master/platform/android/) show an introduction in addition to the file listing.

The documents now distinguish between the two primary use cases of integrating an SDK into an application and building an SDK for SDK development. INSTALL.md is still a bit of a bottleneck, but I’m happy with how things ended up.

jfirebaugh:
This is fantastic.

friedbunny:
Very nice, this feels more natural and self-contained.

One thing that may still trip folks up is `/ios` in the base of the repo, which seems like a logical place to look (if I'm not in to reading and just like clicking the first reasonableish link). Maybe we stick a brief readme there that redirects to `/platform/ios`?

1ec5:
Everything passes. Merging.

jfirebaugh:
@friedbunny `/ios` will be merged into `/platform/ios`. I've been waiting for some of @1ec5's PRs to land (#3183 for example) before doing so, to minimize potential conflicts.

1ec5:
I added a soft redirect anyways. We can toss it when ios/ gets merged.

PR #3424, opened 2016-01-03
[android] Allow running without any permissions

mpetroff:
Although a completely offline app can be created by placing the style and tiles in assets, the `ACCESS_NETWORK_STATE` permission is currently still required, and removing the `INTERNET` permission is currently risky (causes a crash if a request is attempted). This pull request makes both permissions optional.

### DIFF
--- a/platform/android/MapboxGLAndroidSDK/src/main/java/com/mapbox/mapboxsdk/maps/MapView.java
+++ b/platform/android/MapboxGLAndroidSDK/src/main/java/com/mapbox/mapboxsdk/maps/MapView.java
@@ -2232,9 +2232,13 @@ public class MapView extends FrameLayout {
     // Called when MapView is being created
     private boolean isConnected() {
         Context appContext = getContext().getApplicationContext();
-        ConnectivityManager connectivityManager = (ConnectivityManager) appContext.getSystemService(Context.CONNECTIVITY_SERVICE);
-        NetworkInfo activeNetwork = connectivityManager.getActiveNetworkInfo();
-        return (activeNetwork != null) && activeNetwork.isConnectedOrConnecting();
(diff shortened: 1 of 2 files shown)


tobrun:
This looks awesome, thank you @mpetroff for contributing.
@zugaldia are you able to take a look? 
Offline is right up your alley.

zugaldia:
Thank you @mpetroff -- I'll add a note to the offline ticket https://github.com/mapbox/mapbox-gl-native/issues/584.

mpetroff:
I just rebased this off the latest master. However, with the changes in 235f4ee6f679c9cdf2ab7a374782febee1db3003, I don't seem to be able to eliminate the need for the `INTERNET` permission anymore, so this pull request now only removes the need for the `ACCESS_NETWORK_STATE` permission.

jfirebaugh:
@zugaldia Can you review this PR?

zugaldia:
@mpetroff Thank you very much for the PR. Removing unnecesary permisions is always great, but in this case as you noted in your code, our `MapboxEventManager` needs the `ACCESS_NETWORK_STATE` permission or otherwise [this `ContextCompat` condition](https://github.com/mapbox/mapbox-gl-native/pull/3424/files#diff-81f6c0fdddba24b25aacfbb2e7b04ff2R529) will always default to false and it'll result in no events being sent.

I'm closing this one for now, happy to keep hearing from you any other approaches to improve the offline experience.

mpetroff:
Not sending any data is entirely the point. In the use case I have in mind, no Mapbox-hosted data or services are being used and all of the required map data and style files are bundled in the app's assets; therefore, I don't want the app communicating with any servers or using the network connection at all. I can understand if you don't want to include this for business-related reasons.

Issue #2445, opened 2015-09-29
Efficiently apply style changes


kkaefer:
In JS, we have [a way to compute the difference between two styles](https://github.com/mapbox/mapbox-gl-js/pull/1348), and only selectively reparse tiles. This is very useful in Native as well, as it'd allow us to rapidly switch styles without seeing a flash of an empty map. Also see https://github.com/mapbox/mapbox-gl-js/issues/1341.
- [ ] Figure out a way to reparse individual layers of a tile and how buffer uploads are handled
- [ ] Compute differences between style objects, e.g. by creating a hash of all properties of a layer

mourner:
cc @scothis

jfirebaugh:
Note that style diffing is not currently built in to mapbox-gl-js: it's an [external function](https://github.com/mapbox/mapbox-gl-style-spec/blob/mb-pages/lib/diff.js) which produces a set of commands to run in a [batch style change](https://www.mapbox.com/mapbox-gl-js/api/#map/batch).

scothis:
Batching style mutations is what gave us the greatest perf boost, followed by the style diff. Diff'ing without batching is most likely a perf hit rather than a gain (for a non-trivial delta).

I'd like to keep the diff format compatible between JS and native if at all possible, although the JS diff is biased towards the mapbox-gl-js API.

jfirebaugh:
> I'd like to keep the diff format compatible between JS and native if at all possible, although the JS diff is biased towards the mapbox-gl-js API.

Me too, which means the first part of this change is porting all the JS style mutation APIs. It's something we've been meaning to do for quite some time, but a substantial change in and of itself.

scothis:
It's also worth mentioning that mapbox-gl-js does not support every diff operation. In studio when we apply the diff, we fall back to a full setStyle if a particular operation is unsupported.

The most important operations to support are:
- addLayer
- removeLayer
- setLayoutProperty
- setPaintProperty

From there add support for any operation you see that frequently causes a setStyle fallback.

The semantics around add/removeLayer are particularly important to get right as the diff is optimized to avoid unnecessary work. For example, if you move a layer and change a property the diff will be two operations: removeLayer and addLayer.

jfirebaugh:
Depends on #837, more or less.

jfirebaugh:
⇢ #7893

Issue #2720, opened 2015-10-22
Use stale Glyph PBFs


kkaefer:
With #2617 in place, we now get stale data from the cache. Currently, stale Glyph data isn't used, but we should change this. Things to consider:

* The easy case is when the refreshed request returns a 304. In this case, we don't have to do anything.
* The hard case is when we get a different response. Glyphs that change may have different metrics. That means we'll have to reparse any layers that used glyphs from the updated glyph PBF
* The same is true for the Glyph Atlas: We need to update the Glyph stored in the atlas, and probably assign a different position.

On the other hand, glyphs that change dimensions are very rare and should not occur in practice. This could mean that we could just completely blow away the Glyph atlas, and all parsed buckets and do a complete reparse of all symbol buckets.

kkaefer:
Implemented as part of #2721.

PR #2721, opened 2016-01-04
Use stale styles

kkaefer:
With #2617 in place, we now get stale data from the cache. Currently, stale styles aren't used, but we should change this. This could be trivially implemented by regular style update (= treat it like setting a new style). Currently, this means that we have to reparse everything, and will see a short flicker. However, as soon as we implement style transitions (https://github.com/mapbox/mapbox-gl-native/issues/2445), we'll get a flicker-free update.


kkaefer:
:eyes: @jfirebaugh @tmpsantos

kkaefer:
For reference, this patch incorporates #2720, #2722 and #2724 as well.

jfirebaugh:
This PR has a lot of changes and is hard to review comprehensively. Is it possible to split it up into smaller units?

kkaefer:
Will roll this again after https://github.com/mapbox/mapbox-gl-native/pull/3480 is in.

kkaefer:
@jfirebaugh pls :eyes: (minus the first two commits)

Issue #3163, opened 2015-12-02
Storage.HTTPCancelMultiple test occasionally fails


jfirebaugh:
https://travis-ci.org/mapbox/mapbox-gl-native/jobs/94300803#L1439

```
../../test/storage/http_cancel.cpp:43: Failure
Value of: res.error
  Actual: 8-byte object <60-43 24-01 00-00 00-00>
Expected: nullptr
Which is: 8-byte object <00-00 00-00 00-00 00-00>
../../test/storage/http_cancel.cpp:45: Failure
Value of: res.data.get()
  Actual: false
Expected: true
```

jfirebaugh:
https://travis-ci.org/mapbox/mapbox-gl-native/jobs/100170656

The message from the unexpected error is "Couldn't connect to server: couldn't connect to host".

1ec5:
This test is failing for me at least as often on Travis as the iOS KIF tests do on Bitrise.

jfirebaugh:
The test server is initialized with gtest's [Shared Test Resource](https://github.com/google/googletest/blob/35fb11efbe1a2761ce923f49a9df1a430e5d16be/googletest/docs/AdvancedGuide.md#sharing-resources-between-tests-in-the-same-test-case) functionality -- it's started once, used for all `Storage` tests, then shut down. But while it's active, it will [shut itself down if there's no activity for 5 seconds](https://github.com/mapbox/mapbox-gl-native/blob/4d5c6333be52aae4a9c72f4b01941e16ead503f4/test/storage/server.js#L13-L23). As of e6dc4fe49d9e6ce59d4a888b82460a9ec89e3449 this is quite common. When that happens, any subsequent tests that try to connect receive an error.

jfirebaugh:
Fixed in 79368a0ecf8ac4b950f5a3349c0281317c49cb09. Implementing proper process control is up for grabs.

kkaefer:
→ #3432

PR #3427, opened 2016-01-04
Make osxapp document-based

1ec5:
Allow for multiple map windows open at a time. If windows aren’t kept when quitting, the last open window is restored at the next launch, just like in Maps.app.

Also fixed the zoom level used for the map feedback tool, and the URL handler now accepts a `pitch` argument.


1ec5:
I’m unable to get past the “random” failure in #3163, no matter how many times I retry (a dozen times today). Fortunately, this PR goes nowhere near the failing code.

/cc @jfirebaugh

Issue #3374, opened 2015-12-21
make Resource a variant type


jfirebaugh:
```
class FileSource : private util::noncopyable {
public:
    virtual std::unique_ptr<FileRequest> requestSource(const std::string& url, Callback) = 0;
    virtual std::unique_ptr<FileRequest> requestSprite(const std::string& url, Callback) = 0;
    virtual std::unique_ptr<FileRequest> requestGlyphs(const std::string& template, const std::string& fontstack, const GlyphRange&, Callback) = 0;
    virtual std::unique_ptr<FileRequest> requestTile(const std::string& template, const TileID&, Callback) = 0;
    // etc
};
```

This is necessary for offline: the `OfflineFileSource` needs access to actual `z`/`x`/`y` tile coordinates. It shouldn't try to guess them from the URL because there's no way to do that with 100% reliability.

jfirebaugh:
Retitled: we should do this with a variant, not with a ton of virtual methods.

PR #3431, opened 2016-01-05
[core] Remove dependence on Resource from FileCache

jfirebaugh:
`Resource` is going away as part of #3374.

The only place that `Resource#kind` was used was to bypass an attempt at compression for sprite images. But it's fine for that path to just do the compression and use it if it works -- even if not optimal, sprite images account for a relatively small portion of cache rows.

@kkaefer Does that sound ok? Also, is there a reason for `http_cache_kind_idx`? It doesn't seem like it would be utilized by any of the existing queries.

### DIFF
--- a/platform/default/sqlite_cache.cpp
+++ b/platform/default/sqlite_cache.cpp
@@ -68,7 +68,7 @@ void SQLiteCache::Impl::createSchema() {
         "CREATE TABLE IF NOT EXISTS `http_cache` ("
         "    `url` TEXT PRIMARY KEY NOT NULL,"
         "    `status` INTEGER NOT NULL," // The response status (Successful or Error).
-        "    `kind` INTEGER NOT NULL," // The kind of file.
+        "    `kind` INTEGER NOT NULL," // The kind of file. (No longer used; column remains only for backward compatibility.)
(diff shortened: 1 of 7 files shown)


kkaefer:
The idea of storing the kind was that we could discriminate in the cache based on type (e.g. not compress images when storing them in the cache). However, it's probably enough to do magic header detection for those specific use cases.

platform/default/sqlite_cache.cpp
```diff
@@ -68,7 +68,7 @@ void SQLiteCache::Impl::createSchema() {
         "CREATE TABLE IF NOT EXISTS `http_cache` ("
         "    `url` TEXT PRIMARY KEY NOT NULL,"
         "    `status` INTEGER NOT NULL," // The response status (Successful or Error).
-        "    `kind` INTEGER NOT NULL," // The kind of file.
+        "    `kind` INTEGER NOT NULL," // The kind of file. (No longer used; column remains only for backward compatibility.)
```
tmpsantos: You can probably remove this, as the cache will get recreated anyway after https://github.com/mapbox/mapbox-gl-native/pull/3429. Don't forget to bump the cache schema version.

jfirebaugh:
Given recent offline discussions I'm not sure this is necessary.

PR #3432, opened 2016-01-05
Make server tests more robust

kkaefer:
After 3a86321aeecbf4acb2c0b6040d1f332b616a823a, our test server is sometimes lingering around in case the main application crashed and failed to terminate the server. To prevent this from happening, we're going to use a second pipe that we're `dup`ing onto `stdin` in the server, and automatically terminate in case that pipe gets closed.


jfirebaugh:
:+1: Thanks for figuring out how to do this.

Issue #3336, opened 2015-12-17
[ios] In-app changes to MGLMapboxMetricsEnabled should take effect immediately


friedbunny:
In implementing an in-app telemetry on/off switch, we found that changing the `MGLMapboxMetricsEnabled` bool in `NSUserDefaults` will not take immediate effect because the telemetry manager is not actively monitoring that key for changes.

We should implement KVO/etc for this key and trigger `+[MGLMapboxEvents validate]` when it changes.

/cc @1ec5 @incanus

friedbunny:
We [actually do](https://github.com/mapbox/mapbox-gl-native/blob/69cf21cecef1b07c2b9742c5af7b9c1410a5c414/platform/ios/src/MGLMapboxEvents.m#L271-L279) already watch for `NSUserDefaults` changes and then call `+[MGLMapboxEvents validate]`. Verified to be working in iosapp.

PR #3445, opened 2016-01-06
Make telemetry on/off setting available in-app

friedbunny:
Changes the :information_source: action sheet to include a "Mapbox Telemetry" entry that pops up a modal dialog with the telemetry on/off setting. This replaces "Adjust privacy settings", which kicked the user out to the system Settings.app.

Because the telemetry on/off setting is now always available (so long as the :information_source: button is not hidden), there is no longer the requirement that the developer provide such a setting in their app's [`Settings.bundle`](https://github.com/mapbox/mapbox-gl-native/tree/69cf21cecef1b07c2b9742c5af7b9c1410a5c414/ios/app/Settings.bundle) or specify in `Info.plist` that they were otherwise providing this setting in-app.

Providing the telemetry setting in Settings.app (via `Settings.bundle`) is still a recommended best-practice and completely supported, it's just no longer a requirement.

~~Blocked by #3336, which causes this in-app setting changing to not take immediate effect.~~

blah2-compressor

/cc @1ec5


friedbunny:
I'll squash this down before merging, but I wanted to preserve some code for discussion.

1. I think we can remove `+[MGLAccountManager mapboxMetricsEnabledSettingShownInApp]` entirely, but for now I've made it deprecated and throw an exception. This should never have been used externally.
2. `-[MGLMapboxEvents init]` no longer has to care if `Settings.bundle` exists, so I ripped all of that out.

1ec5:
> for now I've made it deprecated and throw an exception

That would make the method unavailable, not deprecated. There’s an attribute for marking a method as unavailable, too, but I don’t think we need the backwards-incompatibility in this case. Just turn the method into a no-op with an `NSLog` about the deprecation.

> `-[MGLMapboxEvents init]` no longer has to care if `Settings.bundle` exists, so I ripped all of that out.

What if the :information_source: button is hidden? Presumably the developer will implement the attribution elsewhere in the application, but then we should clarify the `MGLMapView.attributionButton` documentation. Currently it only talks about copyright notices, but the user opt-out is also important to get right.

include/mbgl/ios/MGLAccountManager.h
```diff
@@ -26,12 +26,12 @@ NS_ASSUME_NONNULL_BEGIN
...
 + (nullable NSString *)accessToken;
 
-#pragma mark Providing User Metrics Opt-Out
+#pragma mark Providing User Telemetry Opt-Out
 
 /**
- Whether in-app user metrics opt-out is configured. If set to the default value of `NO`, a user opt-out preference is expected in a `Settings.bundle` that shows in the application's section within the system Settings app.
- */
-+ (BOOL)mapboxMetricsEnabledSettingShownInApp;
+ Whether in-app user telemetry opt-out is configured. If set to the default value of `NO`, a user opt-out preference was expected in a `Settings.bundle` that shows in the application's section within the system Settings app. Telemetry settings were added directly to the in-app ℹ️ menu, obseleting this method.
```
1ec5: Jazzy doesn’t recognize the deprecated attribute yet. Our convention from the appledoc days is to omit documentation for deprecated and unavailable methods, but if this information is still valuable, I’d stick the last sentence in a `@note`.

platform/ios/src/MGLMapView.mm
```diff
@@ -1421,10 +1418,47 @@ - (void)actionSheet:(UIActionSheet *)actionSheet didDismissWithButtonIndex:(NSIn
...
-    else if (buttonIndex == actionSheet.firstOtherButtonIndex + 4)
+    else if (buttonIndex == actionSheet.firstOtherButtonIndex + 3)
+    {
+        NSString *messageTense;
+        NSString *participate;
+        NSString *optOut;
+
+        if ([[NSUserDefaults standardUserDefaults] boolForKey:@"MGLMapboxMetricsEnabled"])
+        {
+            messageTense = @"are helping to";
```
1ec5: I know we haven’t revisited localization yet, but splitting strings like this complicates localization. The whole sentence should be kept together in one string.

platform/ios/src/MGLMapView.mm
```diff
@@ -1421,10 +1418,47 @@ - (void)actionSheet:(UIActionSheet *)actionSheet didDismissWithButtonIndex:(NSIn
...
+                                                       delegate:self
+                                              cancelButtonTitle:participate
+                                              otherButtonTitles:@"Tell Me More", optOut, nil];
+        [alert show];
+    }
+}
+
+- (void)alertView:(UIAlertView *)alertView didDismissWithButtonIndex:(NSInteger)buttonIndex
+{
+    if (buttonIndex == alertView.cancelButtonIndex)
```
1ec5: This should be a `switch` statement to prevent off-by-one errors after merges.

friedbunny:
> What if the :information_source: button is hidden?

We _could_ repurpose the `MGLMapboxMetricsEnabledSettingShownInApp` plist key, checking if :information_source: is hidden and shouting if the key is not set, but I don't think I'm in favor of that — improved docs for `MGLMapView.attributionButton` will hopefully suffice.

platform/ios/src/MGLMapView.mm
friedbunny: A `switch` requires constant values (`Case value is not a constant expression`), which here would mean hardcoding the indexes rather than using `alertView.cancelButtonIndex` and so on. Is that what you had in mind?

platform/ios/src/MGLMapView.mm
friedbunny: Will consolidate into full sentences. 🎌

include/mbgl/ios/MGLAccountManager.h
friedbunny: Ultimately killed the docs here.

platform/ios/src/MGLMapView.mm
1ec5: I actually had something like #3220 in mind, but that’s overkill for this simple three-option sheet. Never mind.

friedbunny: That is nice, though...

platform/ios/src/MGLAccountManager.m
```diff
@@ -60,7 +52,7 @@ + (instancetype) sharedManager {
 }
 
 + (BOOL) mapboxMetricsEnabledSettingShownInApp {
-    return [MGLAccountManager sharedManager].mapboxMetricsEnabledSettingShownInApp;
+    NSLog(@"mapboxMetricsEnabledSettingShownInApp is no longer necessary, the Mapbox iOS SDK has changed to always provide a telemetry setting in-app.");
```
1ec5: Nit: `,` → `;`.

1ec5:
Looks good!

friedbunny:
Incoming `MGLMapView.attributionButton` docs updates:

screen shot 2016-01-06 at 4 28 30 pm

The new, second note links to https://www.mapbox.com/help/metrics-opt-out-for-users/, which will need updating.

platform/ios/src/MGLMapboxEvents.h
```diff
@@ -66,7 +66,10 @@ typedef NS_MUTABLE_DICTIONARY_OF(NSString *, id) MGLMutableMapboxEventAttributes
 + (void) flush;
 
 // Main thread only
-+ (void)validate;
++ (void) validate;
+
+// Main thread only
++ (void) ensureMetricsOptoutExists;
```
1ec5: We should be asserting that we’re on the main thread, using `MGLAssertIsMainThread()`, instead of using these Java annotation–like comments, which mean nothing.

1ec5:
The setting is still called Mapbox Metrics in a [couple](https://github.com/mapbox/mapbox-gl-native/blob/master/ios/app/Settings.bundle/en.lproj/Root.strings) [places](https://github.com/mapbox/mapbox-gl-native/blob/master/ios/framework/Settings.bundle/en.lproj/Root.strings). (Don’t ask me why it’s duplicated…)

Mind updating the strings?

platform/ios/src/MGLMapboxEvents.m
```diff
@@ -770,6 +740,40 @@ + (BOOL) checkPushEnabled {
...
+                }
+            }
+        }
+
+        if ( ! defaultEnabledValue) {
+            [NSException raise:@"Telemetry opt-out missing" format:
+             @"End users must be able to opt out of Mapbox Telemetry in your app, either inside Settings (via Settings.bundle) or inside this app. "
+             @"By default, this opt-out control is included as a menu item in the attribution action sheet. "
+             @"If you reimplement the opt-out control inside this app, disable this assertion by setting MGLMapboxMetricsEnabledSettingShownInApp to YES in Info.plist."
+             @"\r\n\r\nSee https://www.mapbox.com/ios-sdk/#telemetry_opt_out for more information."
```
1ec5: `\r` is extraneous on POSIX-based systems. (On the Mac, it results in Windows-style selection, so I wouldn’t be surprised if VoiceOver trips over it in iOS.)

ios/app/Settings.bundle/Root.plist
```diff
@@ -18,7 +18,7 @@
 			<key>Key</key>
 			<string>MGLMapboxMetricsEnabled</string>
 			<key>Title</key>
-			<string>Mapbox Metrics</string>
+			<string>Mapbox Telemetry</string>
```
1ec5: Also update the corresponding [.strings files](https://github.com/mapbox/mapbox-gl-native/blob/master/ios/app/Settings.bundle/en.lproj/Root.strings).

platform/ios/src/MGLMapboxEvents.h
friedbunny: We do `MGLAssertIsMainThread()` in most cases here — I'll go ahead and do so for `ensureMetricsOptoutExists`, too.

platform/ios/src/MGLAccountManager.m
```diff
@@ -8,7 +8,6 @@
 
 @interface MGLAccountManager() <FABKit>
 
-@property (atomic) BOOL mapboxMetricsEnabledSettingShownInApp;
```
1ec5: Make sure to test Fabric compatibility.

PR #3322, opened 2015-12-16
2211 custom callout view

RomainQuidet:
Adding a callback to ask user for a custom callout view following mapbox protocol.
Demo available.

PR #3456, opened 2016-01-06
Custom callout views

1ec5:
This PR is a continuation of #3322 by @RomainQuidet with the following additional changes:

* Renamed MGLCalloutView to MGLCompactCalloutView and MGLCalloutViewProtocol to MGLCalloutView to avoid the awkward situation of having to put “Protocol” in a protocol name (for Swift support).
* Replaced individual `title` and `subtitle` properties in MGLCalloutView with a single `representedObject` property that lets you use custom annotation properties in the custom callout view.
* Overrode a problematic SMCalloutView method that treated its own class as an umbrella class. This SDK doesn’t support iOS 5, so there’s no need to support “classic” callouts.
* Documented new APIs.

/cc @friedbunny


bsudekum:
:pray:

RomainQuidet:
Great thanks.

1ec5:
No, thank you! 😊

1ec5:
Categorized `MGLCalloutView` and `MGLCalloutViewDelegate` in jazzy-generated documentation in cd94049537708ec6ef4d35739b0393230e3a5532.

gurumaheshPCS:
Hi Mapbox,
i have some custom annotation view. When user selects any annotation i need to show. in mabox V:3.1.0 we got new method for that in below.
- (nullable UIView <MGLCalloutView> *)mapView:(MGLMapView *)mapView calloutViewForAnnotation:(id <MGLAnnotation>)annotation;

How can i use this method? Can anyone provide some example for this...

1ec5:
Here’s the [documentation for `-mapView:calloutViewForAnnotation:`](https://www.mapbox.com/ios-sdk/api/3.1.1/Protocols/MGLMapViewDelegate.html#/c:objc%28pl%29MGLMapViewDelegate%28im%29mapView:calloutViewForAnnotation:). The iosapp demo application has an [example implementation](https://github.com/mapbox/mapbox-gl-native/blob/ios-v3.1.1/ios/app/MBXViewController.mm#L657-L667) using a custom [MBXCustomCalloutView](https://github.com/mapbox/mapbox-gl-native/blob/ios-v3.1.1/ios/app/MBXCustomCalloutView.m) class.

gurumaheshPCS:
Thanks for providing Demo App. But in my callout view i have two buttons. i am not able to get the button actions when i tap the button. Can you help how the button action will fire.
Thanks,
Gurumahesh Vunna
9703940400

1ec5:
Your MGLCalloutView-conforming class should somehow implement the [`leftAccessoryView`](https://www.mapbox.com/ios-sdk/api/3.1.1/Protocols/MGLCalloutView.html#/c:objc%28pl%29MGLCalloutView%28py%29leftAccessoryView) and `rightAccessoryView` methods. In the MBXCustomCalloutView example, those methods are implemented as properties, so the code that creates the MBXCustomCalloutView would be responsible for setting `leftAccessoryView` to the tappable button. In your case, however, your callout view class probably already knows about the button, so it can set `leftAccessoryView` to that button inside `-initWithFrame:`.

If you have any further questions about this feature, please [contact our Support team](https://www.mapbox.com/contact/), who would be happy to assist you. (They also help us write documentation, so please let them know about things that can be explained better.)

Issue #3128, opened 2015-11-25
Missing arm64 lib


leofarage:
I'm trying to use MapBox's version 2. in my project but when I try to show the Map an **UnsatisfiedLinkError** is thrown.

```java.lang.UnsatisfiedLinkError: dalvik.system.PathClassLoader[DexPathList[[zip file "/data/app/com.whittl.whittl.debug-1/base.apk"],nativeLibraryDirectories=[/data/app/com.whittl.whittl.debug-1/lib/arm64, /vendor/lib64, /system/lib64]]] couldn't find "libmapbox-gl.so"```

I saw an open issue that was discussing something related to that -- in the discussion it seemed that Android should fallback to the 32bit lib if no 64bit was found, but clearly this is not happening in my case.

Both my project and the **MapboxGLAndroidSDKTestApp** included in this repository throw the same error, but https://github.com/mapbox/mapbox-android-demo runs. 

I'm using the same gradle url for fetching MapBox as *mapbox-android-demo*.

I hope you guys can lend me a hand.

Thanks!

zugaldia:
Thank you @leofarage for the report. Just to clarify, you're only seeing this message on 64bits devices, right?

I just tried a fresh build on a Nexus 6p which is a 64 bits device and I couldn't reproduce the issue (see screenshot below). Could you tell us the steps you follow to reproduce this crash? That'd help us troubleshoot the root cause of it. Also, are you seeing this crash with every run, or just during the first time you try to launch the app?

![screenshot_20151201-093900](https://cloud.githubusercontent.com/assets/6964/11503677/c6d2f2f8-980f-11e5-8894-740379e4cbfa.png)

bleege:
@leofarage Yep, the Mapbox Android SDK don't ship an `arm64` .so as [part of the Android Library based on information from Google](https://github.com/mapbox/mapbox-gl-native/issues/1956#issuecomment-137496217).  Could you tell us more about how your project is setup?  Are there other native code bases / dependencies as part of it?

As some background, we regularly develop and test using 64 bit Android devices specifically the Samsung S6 and HTC M9.  That's not to say there isn't an issue but I just wanted to let you know that we regularly run it on 64 bit devices (in addition to 32 bit devices).

leofarage:
My project has a few other libraries that are being used -- possibly forcing the use of 64 bits .so?

[...]

PR #3460, opened 2016-01-07
[android] Fixed x86_64 crash on startup

Malinskiy:
chrono's second is not an int64_t on x86_x64 and there is only a template for int64_t for bind and get.

Seems to start the demo app and work on x86_64 emulator

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -91,7 +91,7 @@ android: android-lib
 
 # Builds all android architectures for distribution.
 apackage: android-lib-arm-v5 android-lib-arm-v7
-apackage: android-lib-x86
+apackage: android-lib-x86 android-lib-x86-64
 apackage: android-lib-mips
 	cd platform/android && ./gradlew --parallel-threads=$(JOBS) assemble$(BUILDTYPE)
 
--- a/platform/default/sqlite_cache.cpp
+++ b/platform/default/sqlite_cache.cpp
@@ -329,9 +329,9 @@ void SQLiteCache::Impl::get(const Resource &resource, Callback callback) {
                 // Status codes > 1 indicate an error
                 response->error = std::make_unique<Response::Error>(Response::Error::Reason(status));
             }
-            response->modified = Seconds(getStmt->get<Seconds::rep>(1));
+            response->modified = Seconds(getStmt->get<int64_t>(1));
             response->etag = getStmt->get<std::string>(2);
-            response->expires = Seconds(getStmt->get<Seconds::rep>(3));
+            response->expires = Seconds(getStmt->get<int64_t>(3));
             response->data = std::make_shared<std::string>(getStmt->get<std::string>(4));
             if (getStmt->get<int>(5)) { // == compressed
                 response->data = std::make_shared<std::string>(util::decompress(*response->data));
@@ -354,7 +354,7 @@ void SQLiteCache::Impl::get(const Resource &resource, Callback callback) {
                 accessedStmt->reset();
             }
 
-            accessedStmt->bind(1, toSeconds(SystemClock::now()).count());
+            accessedStmt->bind(1, int64_t(toSeconds(SystemClock::now()).count()));
             accessedStmt->bind(2, canonicalURL.c_str());
             accessedStmt->run();
         }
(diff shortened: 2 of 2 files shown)


tobrun:
This looks great, will review this soon.

tmpsantos:
I'm curious of what is the datatype. The standard seems to mandate "_A signed integral type of at least 35 bits_" [1]. The method in theory should handle both `int` and `int64_t` (aka `long` on x86_64).

[1] http://www.cplusplus.com/reference/chrono/seconds/

tmpsantos:
And `int` should be non-compliant as it is 32-bits long on x86_64.

Malinskiy:
If you know what type has Ix name in symbol - that's your answer (_ZN6mapbox6sqlite9Statement3get**Ix**EET_i). And chrono8duration**Ix** is the issue judging from the symbol table.

ps. Android may invented a new signed integer type, hehe

tmpsantos:
```
$ c++filt _ZN6mapbox6sqlite9Statement3getIxEET_i
long long mapbox::sqlite::Statement::get<long long>(int)
```

tobrun:
@Malinskiy 
I wanted to test out this PR and see if it solves your issue on running on a x86-64bit emulator.

Before testing I tried reproducing your issue:
- building the library against x86 using:
 
   > make android-lib-x86

- create a x86 emulator:

  screen shot 2016-01-12 at 13 08 38
 
- run testapp:
  
   screen shot 2016-01-12 at 13 07 28

@Malinskiy:
As seen in the screenshot, I'm able to run without any issue and unable to reproduce your problem. Could add some information how I can reproduce your issue?

Malinskiy:
@tobrun The problem with current stable release is the assumption that I don't use any external libraries. Right now my crash is happening because of another library that provides specific build for x86_64 and their x86 version crashes on x86_64 ABI.

I can't force Android to use x86 version if there are any x86_64 libs already in the jni libs. So there are two solutions in my case:
1. Strip the libs (that will cause the other lib to crash on startup on x86_64 since they advise specifically to use the x86_64 jni lib)
2. Build mapbox for x86_64 so that I don't have to strip libs and the other library will start correctly

tobrun:
@Malinskiy thank you for clarifying. 
Pinging @bleege and @zugaldia for exta :eyes: .

bleege:
Thanks for looking into this @Malinskiy.  You're absolutely correct about [the external libraries issue](https://github.com/mapbox/mapbox-gl-native/pull/3460#issuecomment-170894542).  For some background on this, we've gotten similar reports (arm64 in this case) of this ( #3128 ) but have left things as is due to it being [Google's statement on Android only needing 32 bit](http://developer.android.com/training/articles/perf-jni.html#64_bit).  That said [we also said that if someone could address this we'd be happy to consider adding support](https://github.com/mapbox/mapbox-gl-native/issues/3128#issuecomment-164473522).  It looks like that time is now.  :smiley: 

Here's how we'll go about doing that:

1. Have a more experienced C++ developer sign off on using `int64_t`.  /cc @tmpsantos @kkaefer @jfirebaugh 
1. Build `x86` only version of TestApp and run on 32bit Android device
1. Build `x86_64` only version of TestApp and run on 64bit Android device

If everything passes as expected we should also test this fix for `arm64` as well.

**NOTE:** This will noticeably increase the size of the Mapbox `.aar` file as compiled ABIs are between 5 MB and 6 MB in size.

screen shot 2016-01-12 at 9 37 46 am
_Current SNAPSHOT ABI File Sizes_

Malinskiy:
@bleege ok, gotcha.
arm64-v8a only build still crashes on startup on samsung device for me. I believe it was SIGILL in the native stacktrace.
With regards to increasing the aar size - more and more developers are using [splits by apk](http://tools.android.com/tech-docs/new-build-system/user-guide/apk-splits) when they have native packages, so the issue of increasing the apk size can be addressed.

bleege:
@kkaefer was able to review the code just now and is on board with it (see statement below for details).

> using `int64_t` in those places is good; we’re not going to change the internal representation `std::chrono` uses, but rather force the use of overloads where we actually need them

bleege:
@Malinskiy I just merged your fixes in for `x86_64` support and it should be in the next `3.1.0-SNAPSHOT` (currently running through CI and should be available in about 30 minutes or so).  Thank you so much for working on this and contributing to the project!  :+1: 

**NOTE**
_This only affects `x86` and NOT `arm` ABI._

`arm` will still only provide 32 bit ABI in the Mapbox Android `.aar`.  As demonstrated with this PR and as [said previously](https://github.com/mapbox/mapbox-gl-native/issues/3128#issuecomment-164473522) we happily accept PRs so if `arm` 64 bit ABI is desired please feel free to submit a PR.  Otherwise we'll continue to follow [Google's recommendation on 32 bit support](http://developer.android.com/training/articles/perf-jni.html#64_bit) for `arm`.

PR #3454, opened 2016-01-08
privatize SQLiteCache

jfirebaugh:
No code that uses `DefaultFileSource` wants to _not_ use `SQLiteCache`. We should make the `SQLiteCache` instance a member of `DefaultFileSource`, thus enforcing its use, and remove it and the `FileCache` base class from the public API.

cc @kkaefer


jfirebaugh:
:eyes: @kkaefer @tmpsantos 

Despite all the architecture discussion above, I actually had a very simple change in mind to start out with. WDYT?

tmpsantos:
> Despite all the architecture discussion above, I actually had a very simple change in mind to start out with. WDYT?

Looks good to me but we should handle something like empty path not creating a cache.

jfirebaugh:
@tmpsantos Do we need that feature anywhere? If not, let's leave it out.

tmpsantos:
> @tmpsantos Do we need that feature anywhere? If not, let's leave it out.

Not that I'm aware of.

PR #3472, opened 2016-01-08
[core] Simplify OnlineFileSource::Impl::reschedule

jfirebaugh:
I assume `OnlineFileSource::Impl::reschedule` should only ever reschedule a realRequest, and that in the case that the real request should commence immediately, commencing it after a zero-second timeout is also acceptable. If so, this is a desirable refactor because it removes a conditional case and allows `OnlineFileSource::Impl::update` to be inlined as a followup.

cc @kkaefer

### DIFF
--- a/platform/default/online_file_source.cpp
+++ b/platform/default/online_file_source.cpp
@@ -321,9 +321,7 @@ void OnlineFileSource::Impl::reschedule(OnlineFileRequestImpl& request) {
 
     const Seconds timeout = request.getRetryTimeout();
 
-    if (timeout == Seconds::zero()) {
-        update(request);
-    } else if (timeout > Seconds::zero()) {
+    if (timeout >= Seconds::zero()) {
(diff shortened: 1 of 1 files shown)


platform/default/online_file_source.cpp
```diff
@@ -332,9 +332,7 @@ void OnlineFileSource::Impl::reschedule(OnlineFileRequestImpl& request) {
 
     const Seconds timeout = request.getRetryTimeout();
 
-    if (timeout == Seconds::zero()) {
-        update(request);
```
tmpsantos: If you remove this, I guess `::update()` is no longer used anywhere.

jfirebaugh: It's used by `OnlineFileSource::Impl::add`. After I merge this I'll inline it there.

tmpsantos: :+1:

jfirebaugh:
Closing in favor of #3490.

PR #3473, opened 2016-01-08
[core] Simplify OnlineFileSource's Timer use

jfirebaugh:
There is no need for the timer to be created only under certain conditions. Stopping a non-started timer is a no-op (I added a test for this). Therefore it can simply be a direct member.

While here, rename the variable. It's a timer, not a request.

cc @kkaefer

### DIFF
--- a/platform/default/online_file_source.cpp
+++ b/platform/default/online_file_source.cpp
@@ -50,7 +50,7 @@ public:
     const Resource resource;
     std::unique_ptr<WorkRequest> cacheRequest;
     RequestBase* realRequest = nullptr;
-    std::unique_ptr<util::Timer> timerRequest;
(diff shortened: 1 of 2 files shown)


tmpsantos:
LGTM :+1:

tmpsantos:
The initial intention was probably to save some memory by not always allocating a `util::Timer`. This made me think about this -> https://github.com/mapbox/mapbox-gl-native/issues/3486

jfirebaugh:
Even if the memory usage of `Timer` is significant, which I doubt, the majority of requests are expected to go into "refresh mode" eventually, and thus force the allocation.

In general, let's avoid speculative optimizations at the cost code complexity until benchmarking shows them to be necessary.

PR #3474, opened 2016-01-08
Add missing coordinate conversion methods to iOS SDK

1ec5:
The coordinate bounds conversion methods on MGLMapView have been public in the OS X SDK all along. This change makes them public in the iOS SDK too.

/cc @friedbunny @jfirebaugh

### DIFF
--- a/platform/ios/src/MGLMapView.mm
+++ b/platform/ios/src/MGLMapView.mm
@@ -1926,6 +1926,19 @@ std::chrono::steady_clock::duration MGLDurationInSeconds(float duration)
     return MGLCoordinateBoundsFromLatLngBounds([self convertRect:rect toLatLngBoundsFromView:view]);
 }
 
+- (CGRect)convertCoordinateBounds:(MGLCoordinateBounds)bounds toRectToView:(nullable UIView *)view
+{
+    return [self convertLatLngBounds:MGLLatLngBoundsFromCoordinateBounds(bounds) toRectToView:view];
+}
+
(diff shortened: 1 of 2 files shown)


platform/ios/src/MGLMapView.mm
```diff
@@ -1926,6 +1926,19 @@ - (MGLCoordinateBounds)convertRect:(CGRect)rect toCoordinateBoundsFromView:(null
     return MGLCoordinateBoundsFromLatLngBounds([self convertRect:rect toLatLngBoundsFromView:view]);
 }
 
+- (CGRect)convertCoordinateBounds:(MGLCoordinateBounds)bounds toRectToView:(nullable UIView *)view
+{
+    return [self convertLatLngBounds:MGLLatLngBoundsFromCoordinateBounds(bounds) toRectToView:view];
+}
+
+/// Converts a geographic bounding box to a rectangle in the view’s coordinate
```
friedbunny: Have we been three-slashing this whole time and I've never noticed? :scream:

1ec5: Heh, not really. I started doing that in the OS X SDK for short doc comments in implementation files. Xcode’s Quick Help feature is the only consumer of these comments, and it supports the `///` syntax. (For fun, check out the local variable doc comments I added to [`flyTo()`](https://github.com/mapbox/mapbox-gl-native/blob/b6fca670896e6351a03916fd9fb646e1970d3ebe/src/mbgl/map/transform.cpp#L382-L389) – Quick Help picks those up too.)

PR #3475, opened 2016-01-08
[core] Merge OnlineFileRequestImpl::notify into OnlineFileRequestImpl::setResponse

jfirebaugh:
In every case where setResponse is called, notify was called soon after. In one case immediately, in another, after intervening functions which are safe to reorder after notifications.

The null-check of response is safe to eliminate because response must not be null inside setResponse.

cc @kkaefer


tmpsantos:
LGTM :+1:

PR #3477, opened 2016-01-08
Don't store asset:// requests in the cache

kkaefer:
We're currently storing `asset://` requests in the cache, but given that they're on the local file system anyway, we should never do that.


jfirebaugh:
:+1: 

Support for `asset://` URLs is another thing I can imagine moving into `OfflineFileSource`. Logically, they are an offline resource, and the fact that `OnlineFileSource` must handle them differently suggests that they shouldn't be its responsibility.

kkaefer:
Yeah, agreed.

PR #3478, opened 2016-01-08
Make test suite more reliable

kkaefer:


test/fixtures/util.cpp
```diff
@@ -108,12 +108,20 @@ uint64_t crc64(const std::string& str) {
     return crc64(str.data(), str.size());
 }
 
-PremultipliedImage render(Map& map) {
+PremultipliedImage render(Map& map, std::chrono::milliseconds timeout) {
```
jfirebaugh: What's this used for? I'm suspicious of it because it looks like it has the potential to introduce timing-dependent flakiness to tests.

jfirebaugh:
Can you provide more specifics of what "more reliable" means?

test/api/annotations.cpp
```diff
@@ -19,11 +19,15 @@ std::shared_ptr<SpriteImage> namedMarker(const std::string &name) {
     return std::make_shared<SpriteImage>(image.width, image.height, 1.0, std::string(reinterpret_cast<char*>(image.data.get()), image.size()));
 }
 
+namespace {
```
jfirebaugh: Should this anonymous namespace include `namedMarker` from above as well?

PR #3479, opened 2016-01-08
Only cache successful or NotFound responses

kkaefer:
Right now, we're storing all sorts of responses into the cache, potentially overwriting existing successful responses. This means that a bad connection could slowly remove all cache items by simply requesting them.

### DIFF
--- a/test/storage/cache_response.cpp
+++ b/test/storage/cache_response.cpp
@@ -50,3 +50,175 @@ TEST_F(Storage, CacheResponse) {
 
     loop.run();
 }
+
+// Make sure we don't store a bad server response into the cache
+TEST_F(Storage, CacheNotFound) {
+    SCOPED_TEST(CacheNotFound);
+
+    using namespace mbgl;
+
+    util::RunLoop loop;
+    SQLiteCache cache(":memory:");
+    OnlineFileSource fs(&cache);
+
+    const Resource resource{ Resource::Unknown, "http://127.0.0.1:3000/not-found" };
+
+    // Insert existing data into the cache that will be marked as stale.
+    auto response = std::make_shared<Response>();
+    response->data = std::make_shared<const std::string>("existing data");
+    cache.put(resource, response, FileCache::Hint::Full);
+
+    std::unique_ptr<FileRequest> req1;
+    std::unique_ptr<WorkRequest> req2;
+
+    int counter = 0;
+
+    // Then, request the actual URL and check that we're getting the rigged cache response first,
+    // then the connection error message.
+    req1 = fs.request(resource, [&](Response res) {
+        if (counter == 0) {
+            EXPECT_EQ(nullptr, res.error);
+            EXPECT_EQ(true, res.stale);
+            ASSERT_TRUE(res.data.get());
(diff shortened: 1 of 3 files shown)


friedbunny:
Nice! This fixes #2012, where 401s for bad access tokens were being cached.

test/storage/cache_response.cpp
```diff
@@ -50,3 +50,175 @@ TEST_F(Storage, CacheResponse) {
 
     loop.run();
 }
+
+// Make sure we don't store a bad server response into the cache
```
jfirebaugh: This comment should be "Make sure we do store 404 responses in the cache".

test/storage/cache_response.cpp
```diff
@@ -50,3 +50,175 @@ TEST_F(Storage, CacheResponse) {
...
+
+    using namespace mbgl;
+
+    util::RunLoop loop;
+    SQLiteCache cache(":memory:");
+    OnlineFileSource fs(&cache);
+
+    const Resource resource{ Resource::Unknown, "http://127.0.0.1:3000/not-found" };
+
+    // Insert existing data into the cache that will be marked as stale.
```
jfirebaugh: Is having existing data in the cache important to this test in some way? Seems like it could just make a request for an uncached resource that 404s and then verify that it was cached.

test/storage/cache_response.cpp
```diff
@@ -50,3 +50,175 @@ TEST_F(Storage, CacheResponse) {
...
+        } else {
+            FAIL() << "Got too many responses";
+        }
+        counter++;
+    });
+
+    loop.run();
+}
+
+// Make sure we don't store a bad server response into the cache
```
jfirebaugh: "Make sure we don't cache an error response when we can't connect to the server"

test/storage/cache_response.cpp
```diff
@@ -50,3 +50,175 @@ TEST_F(Storage, CacheResponse) {
...
+
+    using namespace mbgl;
+
+    util::RunLoop loop;
+    SQLiteCache cache(":memory:");
+    OnlineFileSource fs(&cache);
+
+    const Resource resource{ Resource::Unknown, "http://127.0.0.1:3001" };
+
+    // Insert existing data into the cache that will be marked as stale.
```
jfirebaugh: Same question about existing data.

test/storage/cache_response.cpp
kkaefer: Yeah, it's important because it verifies that it overwrites an existing "good" response (i.e. non-error response).

test/storage/cache_response.cpp
kkaefer: Same answer as above.

test/storage/cache_response.cpp
kkaefer: Fixed in 13fbf8b39b2aa3814b245fdec000048c973d90e4

test/storage/cache_response.cpp
kkaefer: Fixed in 13fbf8b39b2aa3814b245fdec000048c973d90e4

PR #3486, opened 2016-01-08
[core] ::unref() is probably no longer needed

tmpsantos:
Classes like `Timer` and `AsyncTask` have a `::ref()` and `::unref()` method which is legacy from `libuv` for keeping the main loop alive. Now that we have a `util::RunLoop` abstraction we don't need this and it just causes confusion.

Removing these methods also opens a door for optimization as we can lazy instantiate the `Impl` part of these objects and thus save some memory.

### DIFF
--- a/platform/default/async_task.cpp
+++ b/platform/default/async_task.cpp
@@ -28,6 +28,7 @@ public:
         }
 
         handle()->data = this;
+        uv_unref(handle());
     }
 
     ~Impl() {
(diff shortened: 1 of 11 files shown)


tmpsantos:
:eyes: @jfirebaugh 

It adds a bit of complexity to the `libuv` main loop at the cost of removing some code and making the equivalent for Qt, iOS and Android more simple and we don't have to deal with `unref()` anymore in the [core] code.

jfirebaugh:
:+1:

PR #3491, opened 2016-01-09
Fix stencil overflow bug by properly utilizing the 8 bits we have ava…

PaulZed:
This simplifies id generation and ensures that we aren't wasting space in our stencil buffer by using 1 bitfield for all tiles, instead of a segment of the bitfield per source.

### DIFF
--- a/src/mbgl/util/clip_id.cpp
+++ b/src/mbgl/util/clip_id.cpp
@@ -73,23 +73,19 @@ void ClipIDGenerator::update(std::forward_list<Tile *> tiles) {
         }
     }
 
-    if (!pool.empty()) {
-        const uint32_t bit_count = util::ceil_log2(pool.size() + 1);
-        const std::bitset<8> mask = uint64_t(((1ul << bit_count) - 1) << bit_offset);
+    uint32_t maxIdValue = (1ul << kStencilBufferBitLength) - 1;
 
-        // We are starting our count with 1 since we need at least 1 bit set to distinguish between
-        // areas without any tiles whatsoever and the current area.
-        uint8_t count = 1;
+    if (!pool.empty()) {
+      const std::bitset<8> mask = static_cast<uint64_t>(maxIdValue);
         for (auto& leaf : pool) {
             leaf.tile.clip.mask = mask;
-            leaf.tile.clip.reference = uint32_t(count++) << bit_offset;
+            leaf.tile.clip.reference = clipIdSerial++;
         }
 
-        bit_offset += bit_count;
         pools.push_front(std::move(pool));
     }
 
-    if (bit_offset > 8) {
+    if (clipIdSerial > maxIdValue) {
         Log::Error(Event::OpenGL, "stencil mask overflow");
     }
 }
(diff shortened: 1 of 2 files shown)


ansis:
With this approach I think we'll need to have a separate clipping mask for each source and redraw it whenever we switch between sources. Otherwise tiles in one source will block tiles in the other source from being rendered. I'll look into this more on monday

cc @kkaefer

PaulZed:
I'm not so sure I'm convinced. I don't see a compelling reason why sources would end up with different ids. The operation is for clipping to tile boundaries, so the only thing i can see that would end up causing the ID to be different for 2 collocated sources is sourceZ. That probably just means that tileID is the wrong thing to be using, or our comparator needs to drop sourceZ as a criterion.

ansis:
This is one of the case I'm thinking about:

There are two sources: A and B.
The map is showing zoom 0 and all the z0 tiles are loaded for both sources.
You zoom in to zoom 1.
All the z1 tiles for source A are loaded but source B is slower so they aren't loaded yet.
It renders the four z1 tiles from source A and the larger z0 tile from source B.

It needs some way of clipping the A tiles to each of their boundaries and the B tile to it's larger boundary that overlaps all the A tiles.

cc @kkaefer

PaulZed:
Well, the z coordinate would be different in that case (regular z, not sourceZ), so the clip mask should be fine. I guess my point was if we have Source A and Source B with the same tile coords x y z, then they represent the same region of the map, and we don't need to stencil it twice for the two different sources. they can test the same region.

If they're different in z, then they won't have the same clip ID, and the lower z tile, drawing first, should get clipped properly to its larger tile, and the higher z, drawing second, should get clipped to their appropriate tiles. Is there a use-case where tile coordinate z can be the same between 2 sources, yet sourceZ differs, requiring different clip masks? I can't think of anything.

ansis:
I don't think this would work. The two tiles from different sources in different zoom levels will have different IDs. Each fragment in the stencil buffer holds only one ID. Tiles are clipped by [checking for equality](https://github.com/mapbox/mapbox-gl-native/blob/c10f6489e61a52e3ae36af7cd2c40d120f9181da/src/mbgl/renderer/painter.cpp#L80). The check would match for only *one* of the two tiles so only *one* of them would be rendered at that fragment. However, since the tiles are from different sources you want to render *both* of them.

PaulZed:
Ah, I see. You're correct. I'll have a think on this then.

ansis:
Yeah, it's a messy problem

@kkaefer thinks that long-term we need to move away from stencil clipping by clipping the data before rendering.

billchen-tesla:
Agreed about clipping on the server. Stencil clipping will not work for 3D terrain as well.

PaulZed:
Yeah ideally this is something that can be addressed by not overworking the stencil test. I think if the geo generated on the server was clean enough that we didn't need to do fragment-level clipping, then we could feasibly use stencil only in the case that we have to resolve zoom related issues. We won't need nearly as many bits to accomplish this.

PR #3495, opened 2016-01-09
Send turnstile event when telemetry is paused

friedbunny:
The turnstile event is used solely to count monthly active users and should always be sent, regardless of whether the user has opted themselves out of telemetry.

The turnstile wasn't being turned if telemetry was paused because `_session` is invalided and `nil`'d in `pauseMetricsCollection`. This change instantiates a temporary `NSURLSession` for the turnstile event.

@1ec5 I could use your advice on my use of `strongSelf.session` here — before, we were straight-up calling `_session`, but inside of a background async block.

### DIFF
--- a/platform/ios/src/MGLMapboxEvents.m
+++ b/platform/ios/src/MGLMapboxEvents.m
@@ -543,7 +543,19 @@ const NSTimeInterval MGLFlushInterval = 60;
             [request setHTTPBody:jsonData];
 
             // Send non blocking HTTP Request to server
-            [[_session dataTaskWithRequest:request] resume];
+            if ( ! strongSelf.paused) {
+                [[strongSelf.session dataTaskWithRequest:request] resume];
+            } else {
+                for (MGLMapboxEventAttributes *event in events) {
+                    if ([event[@"event"] isEqualToString:MGLEventTypeAppUserTurnstile]) {
+                        NSURLSession *temporarySession = [NSURLSession sessionWithConfiguration:[NSURLSessionConfiguration defaultSessionConfiguration]
+                                                                                       delegate:strongSelf
+                                                                                  delegateQueue:nil];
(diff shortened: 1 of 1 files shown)


platform/ios/src/MGLMapboxEvents.m
```diff
@@ -543,7 +543,17 @@ - (void) postEvents:(NS_ARRAY_OF(MGLMapboxEventAttributes *) *)events {
...
+            if (strongSelf.session) {
+                [[strongSelf.session dataTaskWithRequest:request] resume];
+            } else {
+                // Session is invalidated when telemetry is paused, which means
+                // we have to manually manage its lifecycle here for the turnstile.
+                strongSelf.session = [NSURLSession sessionWithConfiguration:[NSURLSessionConfiguration defaultSessionConfiguration]
+                                                                   delegate:strongSelf
+                                                              delegateQueue:nil];
+                [[strongSelf.session dataTaskWithRequest:request] resume];
+                [strongSelf.session finishTasksAndInvalidate];
```
1ec5: Rather than checking for the existence of `MGLMapboxEvents.session`, I think it’d be better to check for paused state, or better yet, for an indication that this is a turnstile event. If we’re getting in here without a session for anything other than a turnstile event, that’s an error on our part, so the original code attempts to fail gracefully.

Additionally, could we keep this synchronous, turnstile-specific session a local variable, to avoid confusion about what kind of session is being stored in the ivar?

friedbunny: Making this more specific to the turnstile sounds good, forget why I didn't go with `isPaused` here but will look at casing this to the turnstile only.

:+1: on using a standalone `NSURLSession` here.

friedbunny:
Squashed and rebased, will merge when the tests go green.

PR #3496, opened 2016-01-09
Add telemetry on/off status to turnstile event

friedbunny:
Depends on #3495.

/cc @1ec5

### DIFF
--- a/platform/ios/src/MGLMapboxEvents.m
+++ b/platform/ios/src/MGLMapboxEvents.m
@@ -391,31 +391,27 @@ const NSTimeInterval MGLFlushInterval = 60;
 }
 
 - (void) pushTurnstileEvent {
-
     __weak MGLMapboxEvents *weakSelf = self;
 
     dispatch_async(_serialQueue, ^{
-
         MGLMapboxEvents *strongSelf = weakSelf;
 
         if ( ! strongSelf) return;
 
-            // Build only IDFV event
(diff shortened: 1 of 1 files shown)


platform/ios/src/MGLMapboxEvents.m
```diff
@@ -391,31 +391,27 @@ - (void)startUpdatingLocation {
...
-            [strongSelf flush];
+        NSString *idfv = [[[UIDevice currentDevice] identifierForVendor] UUIDString];
+        NSDictionary *vevt = @{
+            @"event" : MGLEventTypeAppUserTurnstile,
+            @"created" : [strongSelf.rfc3339DateFormatter stringFromDate:[NSDate date]],
+            @"appBundleId" : strongSelf.appBundleId,
+            @"vendorId": idfv,
+            @"version": @(version),
+            @"instance": strongSelf.instanceID,
+            @"telemetry": @([[strongSelf class] isEnabled])
```
1ec5: Perhaps a more descriptive key like `telemetryEnabled` would be more future-proof.

mikemorris:
:+1: Will be good to surface these stats.

platform/ios/src/MGLMapboxEvents.m
friedbunny: `enabled.telemetry` would fit with similar key names. 🤔

friedbunny:
Superceded by #3620.

PR #3355, opened 2015-12-18
Further refine flyTo() flight path

1ec5:
This is a followup to #3301 that further refines the flight path used by `Transform::flyTo()` to match GL JS exactly, even as the pitch varies between the endpoints. It fixes #3296 and ensures that the ending zoom level is respected regardless of the starting zoom level. Finally, I began documenting and renaming variables to reflect a better understanding of this algorithm, now that [this paper](https://www.win.tue.nl/%7Evanwijk/zoompan.pdf) has been rediscovered.

Other noteworthy changes:
* Began pushing mutators into `TransformState`, which will allow us to begin factoring redundant code out of `Transform`.
* Addresses mapbox/mapbox-gl-js#1853 by crossing the antimeridian if that would be a shorter path.
* Replaces the opaque `speed` and `curve` options with more intuitive options that are expressed in screen units.

/cc @bleege @zugaldia @mourner @lucaswoj

Issue #2600, opened 2015-10-13
Offset map center


maciekish:
Hi,
Is it possible to easily offset what the map treats as it centre? We want the map to follow the users position more like a navigation app where the "car" is much lower on the screen so that you can see further ahead.

I've looked for a value that defines the center but wasn't able to find anything. We have our own fork so it doesn't matter what the change is as long as it's possible.

<img src="http://f.cl.ly/items/1i3p0K1n0w2T3E1J2q2a/IMG_0051.PNG" width="134">

Thanks
Maciej

picciano:
+1

Zakay:
+1 Both Apple and Google uses this offset thing on their map apps. Would be very useful to have!

incanus:
> Both Apple and Google uses this offset thing on their map apps. 

Just to clarify, this is in their own user-facing apps, not their developer-facing frameworks. 

But agree.

1ec5:
This is in some ways blocked by #1041: currently “user tracking” consists of moving the user dot in response to user location changes, then synchronizing the map’s center point in response. A better approach would be to affix the user dot to a set location on-screen (usually the center point but in this case somewhere near the bottom) and pan the map directly in response to user location changes. The logic in `-[MGLMapView locationManager:didUpdateLocations:]` may need to be rewritten, especially due to tilting, which is typically used with course tracking mode.

ljbade:
@1ec5 Glueing the dot to the screen centre then moving the map was the approach @tobrun took in the Android code.

RomainQuidet:
Apple MapKit uses 2 APIs for this using insets: 
- (void)setVisibleMapRect:(MKMapRect)mapRect
              edgePadding:(UIEdgeInsets)insets
                 animated:(BOOL)animate

and

- (MKMapRect)mapRectThatFits:(MKMapRect)mapRect
                 edgePadding:(UIEdgeInsets)insets

1ec5:
We have `-setVisibleCoordinateBounds:edgePadding:animated:` already; a `-coordinateBoundsThatFits:edgePadding:` would certainly complement it. But I think this request is more about having MGLMapView offset the center when _automatically_ panning due to user location tracking.

[...]

PR #3497, opened 2016-01-11
Refactor Transform

1ec5:
This PR cleans up `Transform` considerably, making it easier to implement #2600 at this low level:

* `CameraOptions` only describes the destination viewpoint now. `AnimationOptions` describes the transition to that destination viewpoint.
* Eliminated many redundant methods on `Transform`.
* Minimized usage of `TransformState::x` and `TransformState::y`.
* Added convenience constructors for `AnimationOptions`. When scaling, passing in view’s origin as the anchor no longer anchors the scaling operation at the view’s center. To specify no anchor, use `NaN` instead of 0.
* Corrected `setBearing()` to assume an origin at the upper-left corner of the view rather than the lower-left corner.
* Added an `anchor` option to `CameraOptions` that pins a coordinate to a point within the view throughout a zoom, rotation, or tilt. This option is useful for ensuring that a pinch or rotation gesture is revolves around the gesture rather than the view’s center point. However, all the platform frontends currently use convenience methods for this purpose anyways.
* Easing across the antimeridian now travels the short way around the globe (see #3355).

/cc @tmpsantos @zugaldia

### DIFF
--- a/include/mbgl/map/camera.hpp
+++ b/include/mbgl/map/camera.hpp
@@ -17,6 +17,10 @@ struct CameraOptions {
     /** Coordinate at the center of the map. */
     mapbox::util::optional<LatLng> center;
     
+    /** Point of reference for `zoom` and `angle`, assuming an origin at the
+        top-left corner of the view. */
+    mapbox::util::optional<PrecisionPoint> anchor;
+    
     /** Zero-based zoom level. Constrained to the minimum and maximum zoom
         levels. */
     mapbox::util::optional<double> zoom;
--- a/include/mbgl/map/camera.hpp
+++ b/include/mbgl/map/camera.hpp
@@ -11,13 +11,9 @@
 
 namespace mbgl {
 
-/** Various options for describing the viewpoint of a map, along with parameters
-    for transitioning to the viewpoint with animation. All fields are optional;
-    the default values of transition options depend on how this struct is used.
- */
+/** Various options for describing the viewpoint of a map. All fields are
+    optional. */
 struct CameraOptions {
-    // Viewpoint options
-    
     /** Coordinate at the center of the map. */
     mapbox::util::optional<LatLng> center;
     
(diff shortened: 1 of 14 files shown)


mappy-mobile:
Hi @1ec5,
I just saw this pull-request related to #2600. You said that it will make easier to implement insets with your work. So I was wondering if you will work on #2600 after this pull-request ? Otherwise can you give us some clues about how you think it should be implemented with your refactoring.
For the moment we pause our work on #2600 because it seems there will be a lot of conflicts once you will merge this pull-request.
Regards,
Larry

1ec5:
Hi Larry, #2600 is definitely top of mind for me right now. I’m open to your ideas about where the insets should be implemented.

The refactoring in this PR is needed generally, but I wanted to make it possible to eventually implement the inset functionality down in `Transform` and/or `TransformState`, so that other platforms like Android can take advantage of them. I’m not sure yet whether the insets should be merely a camera option like `around` in GL JS (the least flexible approach), or an optional parameter like the `anchor` parameter in this PR, or an additional field in `TransformState` (the most complex but more comprehensive approach).

Do you think you could push what you have so far (even a gist with a high-level overview is fine), so we can hash out the details and avoid conflicts?

lgeromegnace:
Yes sure we can push our work. It is not working fine with all interactions impacted by the insets. We still have issues to deal with.

As you suggested in your comment we also implement the inset functionality in Transform and TransformState.

The base idea is to add the insets on center(x,y) point in easeTo method. 

To do so we must take into account the map's angle and pitch. Unfortunately we ended up with an issue with the pitch perspective.
Because of that we tried to modify the projection matrix in TransformState just before applying the rotation transformations by translating the center by the insets. The pitch was fine and the animation to center on the user position too. However this code broke the gestures excepting the pitching one by adding insets continuously. 

Finally we tried to detect the "pitching gesture" by adding an API to state to be aware of it and applied in TransformState the inset to the matrix only if we are pitching. On the other side, in Transform we applied the inset only if we are not in a "gesture in progress" state. Still not good : the pitch case is not ok.

At this moment I cannot give you a clear vision on what should be done to fulfill the inset functionality but here are some leads we have in mind: 

* I think it could help to have more info in state such as "start panning". With this it will be possible to synchronize the state.x|y with the insets (previously applied on the proj matrix) at the beginning of the gesture only. Indeed, we could compute a delta between start and current position.

* I see that the proj matrix is rebuilt each time. What about applying the transformations once ? Maybe you have constraints that lead you to this design ?

I prepare a pull-request with a "DO NOT MERGE" in the commit message if you want to have a look on what we tried so far.

1ec5:
Wonderful! I followed up in #3523. For this PR, I’m going to finish cleaning `Transform` up, then insets will likely be in a separate PR.

jfirebaugh:
This is looking great. :+1: :100:

tmpsantos:
Fantastic!

PR #3501, opened 2016-01-11
[core] Do not make a copy of the TransformState all the time

tmpsantos:
Return a const ref instead. We use this mostly for:

```
return transform->getState().getMinZoom();
return transform->getState().getMaxZoom();
return transform->getState().getWidth();
return transform->getState().getHeight();
return transform->getState().latLngToPoint(latLng);
return transform->getState().pointToLatLng(pixel);
```

And when passing it on `::invoke()` it gets copied when packing the arguments.

### DIFF
--- a/src/mbgl/map/map.cpp
+++ b/src/mbgl/map/map.cpp
@@ -71,7 +71,6 @@ public:
 
     Map::StillImageCallback callback;
     size_t sourceCacheSize;
-    TransformState transformState;
     TimePoint timePoint;
     bool loading = false;
 };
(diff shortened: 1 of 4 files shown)


tmpsantos:
:eyes: @jfirebaugh

jfirebaugh:
Seems ok. Is this a performance bottleneck though?

mb12:
@tmpsantos and @jfirebaugh Can we delete copy constructor and assignment operator of TransformState to ensure objects of this class are never copied? Or at the very least make the copy constructor explicit.

tmpsantos:
> @tmpsantos and @jfirebaugh Can we delete copy constructor and assignment operator of TransformState to ensure objects of this class are never copied? Or at the very least make the copy constructor explicit.

No we can't. It needs to get copied when we send it to another thread.

mb12:
@tmpsantos You can still delete the assignment operator and make the copy constructor explicit.

PR #3504, opened 2016-01-11
[tests] Fix how we launch the test server

tmpsantos:
`execl` uses varargs thus must be NULL terminated.


tmpsantos:
:eyes: @kkaefer

kkaefer:
Looks good!

Issue #3488, opened 2016-01-08
switch u_patternmatrix to u_patternscale in pattern shader


ansis:
the same way it was done here: https://github.com/mapbox/mapbox-gl-js/pull/1886

PR #3509, opened 2016-01-11
[core] seamless fill-pattern

kelvinabrokwa:
From:
image

To:
image

### DIFF
--- a/src/mbgl/renderer/painter_fill.cpp
+++ b/src/mbgl/renderer/painter_fill.cpp
@@ -65,7 +65,7 @@ void Painter::renderFill(FillBucket& bucket, const FillLayer& layer, const TileI
 
         // Image fill.
         if (pass == RenderPass::Translucent && posA && posB) {
-            float factor = 8.0 / std::pow(2, state.getIntegerZoom() - id.z) / id.overscaling;
+            float factor = (8.0 / std::pow(2, state.getIntegerZoom() - id.z)) / id.overscaling;
 
             mat3 patternMatrixA;
             matrix::identity(patternMatrixA);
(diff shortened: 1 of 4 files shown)


kelvinabrokwa:
@ansis @lucaswoj @jfirebaugh :eyes: before I squash, please

kelvinabrokwa:
oops, actually not quite there. Will re-ping when ready.

src/mbgl/renderer/painter_fill.cpp
```diff
@@ -90,6 +90,26 @@ void Painter::renderFill(FillBucket& bucket, const FillLayer& layer, const TileI
...
+                (int)((*posA).size[1] * properties.pattern.value.fromScale)
+            }};
+            std::array<int, 2> imageSizeScaledB = {{
+                (int)((*posB).size[0] * properties.pattern.value.toScale),
+                (int)((*posB).size[1] * properties.pattern.value.toScale)
+            }};
+
+            int tileSize = 512;
+
+            float offsetAx = ((tileSize % imageSizeScaledA[0]) * id.x) / (float)imageSizeScaledA[0];
```
lucaswoj: Is `w` not available on the native side? https://github.com/mapbox/mapbox-gl-js/pull/1908/files#diff-a406b8ffc7f2d9f303442aecebb9a2bdR164

src/mbgl/shader/pattern.vertex.glsl
```diff
@@ -9,6 +12,6 @@ varying vec2 v_pos_b;
 
 void main() {
     gl_Position = u_matrix * vec4(a_pos, 0, 1);
-    v_pos_a = (u_patternmatrix_a * vec3(a_pos, 1)).xy;
-    v_pos_b = (u_patternmatrix_b * vec3(a_pos, 1)).xy;
+    v_pos_a = (u_patternmatrix_a * vec3(a_pos, 1)).xy + u_offset_a;
+    v_pos_b = (u_patternmatrix_b * vec3(a_pos, 1)).xy + u_offset_b;
```
lucaswoj: It looks to me like u_`patternmatrix_*` are transformation matrices. Have you looked into applying the offset using the transformation matrix? That'd save us from having to add `u_offset_*` uniforms.

ansis: I think we should remove u_patternmatrix_ like with did in -js. https://github.com/mapbox/mapbox-gl-native/issues/3488

Or we could re-add it in -js but I think the way -js does it now is slightly better.

src/mbgl/renderer/painter_fill.cpp
kelvinabrokwa: On native, patterns are only seamless across the dateline when I you don't increment the `x` multiple by a factor of the `w` (which is the opposite from -js). To be completely honest, I really don't understand why... :cold_sweat:

src/mbgl/shader/pattern.vertex.glsl
kelvinabrokwa: I can definitely look into that

ansis: If removing the matrix like in -js sounds good, then it can be done separately from this pull request

src/mbgl/renderer/painter_fill.cpp
ansis: It looks like `x` is wrapped in -js: https://github.com/mapbox/mapbox-gl-js/blob/25b02a05d29130702e8f66e6a2bfe91333ac010f/js/source/tile_coord.js#L157

but not in -native: https://github.com/mapbox/mapbox-gl-native/blob/11062a01b40cb64c57d1b7b2ef8c0fc39c11d159/src/mbgl/util/tile_cover.cpp#L76-L77

so `x` already has the 'w' part built into it. I opened https://github.com/mapbox/mapbox-gl-native/issues/3529 about that

PR #3512, opened 2016-01-12
[core] Simplify asset:// implementation

jfirebaugh:
Followup to #3490.

* Move asset:// URL handling to `DefaultFileSource`.
* `AssetFileSource` implements `FileSource` interface and follows familiar implementation patterns.
* Move default implementation to `platform/default`, zip implementation to `platform/android`.
* Don't bother with modified / expires / etag -- assets are not cached so it doesn't matter.
* Don't bother with interleaving individual IO calls on the implementation thread. That adds a lot of complexity for very little benefit.
* Eliminate `platform::assetRoot()`.

cc @kkaefer @tmpsantos


mb12:
@jfirebaugh  without platform specific assetRoot, how do you figure out if assets have to be loaded from main bundle or MapboxGL.bundle? This is for iOS.

jfirebaugh:
There's still a platform specific root, it's merely passed as a normal parameter rather than looked up via a platform-specific implementation of `platform::assetRoot()`.

mb12:
Thanks for the clarification @jfirebaugh .

jfirebaugh:
Merged.

PR #3519, opened 2016-01-12
[android] #3498 - disable missing permission lint check inline

### LINKED ISSUE #3498: Disable MissingPermission lint at SDK level
Our SDK has a special requirement that doesn't go well with the Android M Permission model an associated lint warning of MissingPermssion. Our requirement is that the end-user of the library can choose to add or not add locational permissions.

Current implementation results in following error:

screen shot 2016-01-11 at 10 09 21

The only way to resolve this error is be enforcing the permission in the AndroidManifest form the SDK but this will automatically enforce the permission on users that are not interested in this permission. The only way to resolve this is to disable this lint error in `build.gradle` (as we do for the invalid package lint error).

tobrun:
Closes #3498


tobrun:

PR #3511, opened 2016-01-12
Test fail with "terminating with uncaught exception of type std::runtime_error: Failed to close loop."

tmpsantos:
```
[ RUN      ] Storage.HTTPLoad
libc++abi.dylib: terminating with uncaught exception of type std::runtime_error: Failed to close loop.
./scripts/run_tests.sh: line 32: 96090 Abort trap: 6           (core dumped) ${CMD}
make[1]: *** [test-Storage.HTTPLoad] Error 134
make: *** [test-Storage.HTTPLoad] Error 2
```

Presumably related to d7fdcc73bfcab39f63e547ce2af8435a9859cb08.

cc @tmpsantos

### DIFF
--- a/include/mbgl/util/run_loop.hpp
+++ b/include/mbgl/util/run_loop.hpp
@@ -41,6 +41,9 @@ public:
     void runOnce();
     void stop();
 
+    void ref();
+    void unref();
+
     // So far only needed by the libcurl backend.
     void addWatch(int fd, Event, std::function<void(int, Event)>&& callback);
     void removeWatch(int fd);
--- a/platform/darwin/http_request_nsurl.mm
+++ b/platform/darwin/http_request_nsurl.mm
@@ -5,6 +5,7 @@
 #include <mbgl/util/async_task.hpp>
 #include <mbgl/util/time.hpp>
 #include <mbgl/util/parsedate.h>
+#include <mbgl/util/run_loop.hpp>
 
 #import <Foundation/Foundation.h>
 
@@ -96,6 +97,11 @@ HTTPNSURLRequest::HTTPNSURLRequest(HTTPNSURLContext* context_,
       context(context_),
       existingResponse(existingResponse_),
       async([this] { handleResponse(); }) {
+    // Hold the main loop alive until the request returns. This
+    // is needed because completion handler runs in another
+    // thread and will notify this thread using AsyncTask.
+    util::RunLoop::Get()->ref();
+
     @autoreleasepool {
         NSURL* url = [NSURL URLWithString:@(url_.c_str())];
         if (context->accountType == 0 &&
@@ -129,6 +135,7 @@ HTTPNSURLRequest::HTTPNSURLRequest(HTTPNSURLContext* context_,
 }
 
 HTTPNSURLRequest::~HTTPNSURLRequest() {
+    util::RunLoop::Get()->unref();
     assert(!task);
 }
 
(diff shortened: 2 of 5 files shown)


tmpsantos:
I'm not happy with the "fix", as what really needs to be done is find out why the request is apparently not being destroyed.

brunoabinader:
Tested on OSX :+1:

brunoabinader:
FYI, after a few test runs, there was a one that made `Storage.HTTPLoad` run forever.

jfirebaugh:
I think I understand libuv a little better after looking at this: `uv_unref` does not mean you can call `uv_loop_close` without getting `UV_EBUSY` if there's still an active handle. It just means that handle won't prevent `uv_run` from returning.

The reason there's still an active handle is indeed that `HTTPNSURLRequest::cancel` doesn't immediately `delete this` if there's an active `task`. Execution of the completion handler happens on a different thread than `HTTPNSURLRequest::cancel`, and `delete this` would race with access to `this` within the completion handler.

`HTTPNSURLRequest::cancel` could do an immediate `delete` if it shared a mutex with the completion handler that is locked by the completion handler when using `this` and by `HTTPNSURLRequest::cancel` while doing `delete this`. (Note: a shared atomic flag would not be sufficient: it would still allow `delete this` to happen after the completion handler checked the flag but before it accessed `this`.)

tmpsantos:
> I think I understand libuv a little better after looking at this: uv_unref does not mean you can call uv_loop_close without getting UV_EBUSY if there's still an active handle. It just means that handle won't prevent uv_run from returning.

Exactly.

> HTTPNSURLRequest::cancel could do an immediate delete if it shared a mutex with the completion handler that is locked by the completion handler when using this and by HTTPNSURLRequest::cancel while doing delete this.

Alternatively, if it is safe, we could simple delete the util::AsyncTask on `cancel()` (we make it a `unique_ptr`).

jfirebaugh:
> Alternatively, if it is safe, we could simple delete the util::AsyncTask on `cancel()` (we make it a `unique_ptr`).

The completion task, running on another thread, also accesses the `AsyncTask`. So it's not safe unless we change how the completion task works.

kkaefer:
This error also happens when running with `ASSET=zip`

tmpsantos:
> This error also happens when running with ASSET=zip

On the Storage.Asset* tests?

jfirebaugh:
Bump @tmpsantos -- this is a nuisance for running tests locally.

tmpsantos:
> Bump @tmpsantos -- this is a nuisance for running tests locally.

Sorry, on it.

tmpsantos:
Closed by accident.

:eyes: @jfirebaugh I'm not super happy with this fix but the code for making the completion handler and the FileSource thread cooperate was getting way too complicated due to thread safety issues.

This pretty much restores the old behavior, although with a better semantic IMO.

tmpsantos:
BTW, tested `Storage.HTTPLoad` on OSX 150+ times with no incidents.

jfirebaugh:
:+1:

tmpsantos:
> uv_unref on a non-refed handle should be a nop

Sorry @kkaefer saw it now when deleting the branch. You talking about the RunLoop starting with ref == 1 and unref() stopping it?

annymsMthd:
Seeing this failure on android

tmpsantos:
@annymsMthd do you have the steps of how to reproduce?

PR #3523, opened 2016-01-12
2600 offset map center

lgeromegnace:
This pull-request is linked to #3497. This is a WIP so please DO NOT MERGE.

### DIFF
--- a/include/mbgl/ios/MGLMapView.h
+++ b/include/mbgl/ios/MGLMapView.h
@@ -82,6 +82,8 @@ IB_DESIGNABLE
 /** The button shown in the lower-right of the map which when pressed displays the map attribution information. */
 @property (nonatomic, readonly) UIButton *attributionButton;
 
+@property (nonatomic) UIEdgeInsets mapViewInsets;
+
 #pragma mark - Accessing the Delegate
 
 /** @name Accessing the Delegate */
--- a/include/mbgl/ios/MGLMapView.h
+++ b/include/mbgl/ios/MGLMapView.h
@@ -82,6 +82,7 @@ IB_DESIGNABLE
 /** The button shown in the lower-right of the map which when pressed displays the map attribution information. */
 @property (nonatomic, readonly) UIButton *attributionButton;
 
+/** Defines with insets an area where the contents (marker, user position) of the map will be centered.*/
 @property (nonatomic) UIEdgeInsets mapViewInsets;
 
 #pragma mark - Accessing the Delegate
--- a/include/mbgl/map/map.hpp
+++ b/include/mbgl/map/map.hpp
@@ -132,6 +132,7 @@ public:
     // Pitch
     void setPitch(double pitch, const Duration& = Duration::zero());
     double getPitch() const;
+    void setPitching(bool pitching);
 
     // North Orientation
     void setNorthOrientation(NorthOrientation);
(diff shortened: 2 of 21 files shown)


1ec5:
Thank you very much for this work. It’ll help a great deal in coming up with a solution that works on every platform. One of the challenges as I clean up `Transform` and `TransformState` will be to account for frontends that don’t take advantage of `Transform`’s built-in animation capabilities, instead calling the non-animated methods repeatedly using a platform-provided animation facility. To hand-wave for a moment, it’d be really cool to let UIKit Dynamics or Core Animation drive [custom animations](https://www.objc.io/issues/12-animations/animating-custom-layer-properties/) this way.

I’m curious what led to the decision to maintain the insets as a field on `TransformState`. While working on #3497, I was thinking that `Transform`’s methods are used in too many ways for there to be a good sense of when to automatically apply insets. The `gestureInProgress` state appears to have been a first step towards making `Transform` more aware of the caller’s intent, but I’m not sure that it’s even being used the way it was originally intended anymore.

A conservative first step would be to just have the platform-specific code (MGLMapView in the case of iOS) keep track of the view insets and pass them into `Transform` methods as an option in `CameraOptions`. MGLMapView has all the context it needs to determine whether to apply the inset as well as to report the shifted center coordinate to client code when appropriate.

I see you’ve thought about tilting too. What’s the expected interaction for a tilted map? mbgl uses a vertical field of view based on the view’s height. Perhaps the axis of tilt would be shifted downward, but I’m wary of introducing an additional degree of freedom within the scope of this feature. (It would be awesome to have that feature, though!)

lgeromegnace:
Hi @1ec5,

Thanks a lot for your feedback ! 

Your are right about keeping track of insets in `CameraOptions`. What led me to add them in TransformState was the fact that in Map there are others API like `MoveBy` that does not use the `CameraOptions` and also at some point of my attempts I needed the insets in `TransformState`. Howover I think it is okay to ignore the first reason I mentioned since these API are not exposed in `MapView`.

So with insets in `CameraOptions` I have a map that can apply insets to center since it is not "pitched". 

I saw you just finished your refactor in #3497. I'll have a look on it and I think I will plug this solution on it.

For the tilted map and insets I'll see if I can figure it out but as you said maybe this should be addressed in another PR.

1ec5:
@lgeromegnace, some good news and bad news. I see you recently pushed https://github.com/Mappy/mapbox-gl-native/tree/2600-insets. The bad news is that we had an internal project that needed insets right away on OS X, so I had to step on your toes a bit and start implementing insets in https://github.com/mapbox/mapbox-gl-native/tree/1ec5-padding-2600. The good news is that you and I chose almost identical approaches at the core level, and since you’re working on the iOS SDK and I’m working on the OS X SDK, conflicts between our branches should be minimal.

Apologies for any work duplication here. I really like what you’re doing on the iOS side though. :+1: Are you planning to work on shifting the center during user tracking mode also?

lgeromegnace:
@1ec5, no need to apologize. On the contrary I'm happy to have you on this field :+1:. I like the work you started, `moveLatLng` in `TransformState` is very interesting, I missed it from your previous refactor.  For the conflicts I assume you will be faster than me to handle them. However if you have any questions feel free to ask.

About shifting the center during user tracking mode : yes we have to apply inset automatically. So we expect to have the user position centered in the bounds defined by the insets.

tobrun:
@lgeromegnace 
I believe this has landed for both iOS and Android.
Can we close this?

lgeromegnace:
@tobrun you're right. Thanks for the reminder.

Issue #3513, opened 2016-01-12
icon offsets in sprite render tests


ansis:
<img width="1274" alt="screen shot 2016-01-11 at 4 19 22 pm" src="https://cloud.githubusercontent.com/assets/1421652/12250862/1edaebb0-b87f-11e5-90a3-8f920fbf074a.png">

ansis:
fixed by https://github.com/mapbox/mapbox-gl-native/pull/3530

Issue #3528, opened 2016-01-12
sprite padding artifact in sprites/2x-screen-1x-pattern


ansis:
(not in master yet. I'm creating the issue now so I can reference it)

<img width="494" alt="screen shot 2016-01-12 at 1 50 58 pm" src="https://cloud.githubusercontent.com/assets/1421652/12278047/a66ba218-b933-11e5-8bc5-fb11ef2721ed.png">

I suspect that this is because the sprite atlas uses `GL_CLAMP_TO_EDGE` and the sprite is directly next to the edge.

ansis:
fixed by https://github.com/mapbox/mapbox-gl-native/pull/3530

PR #3530, opened 2016-01-13
don't scale when adding to sprite atlas

### LINKED ISSUE #3164: never scale images when copying them into the SpriteAtlas

ansis:
Don't scale images when adding them to the sprite atlas and make the sprite atlas match -js.

fix #3164 and port https://github.com/mapbox/mapbox-gl-js/pull/1919/files

can you review @jfirebaugh?


jfirebaugh:
Also fixes #3513?

mb12:
@ansis Can SpriteAtlas::bind be unconditionally called with linear=true?

jfirebaugh:
What about [this scaling code](https://github.com/mapbox/mapbox-gl-native/blob/sprite-atlas-no-scaling/src/mbgl/sprite/sprite_atlas.cpp#L132-L152)?

ansis:
> Also fixes #3513?

Yes, except for https://github.com/mapbox/mapbox-gl-native/issues/3528

> @ansis Can SpriteAtlas::bind be unconditionally called with linear=true?

Using nearest neighbour filtering is clearer when icons copied at their original size. This small bit of extra clarity matters since icons are already pretty small.

> What about this scaling code?

yeah, that needs to be removed.

ansis:
The nearest neighbour scaling code is removed in fc20199

ansis:
@jfirebaugh I think this should be good now. Want to take another look?

src/mbgl/sprite/sprite_atlas.cpp
```diff
@@ -83,28 +83,48 @@ mapbox::util::optional<SpriteAtlasPosition> SpriteAtlas::getPosition(const std::
...
 
+void copyBitmap(const uint32_t *src, const uint32_t srcStride, const uint32_t srcX, const uint32_t srcY,
+        uint32_t *const dst, const uint32_t dstStride, const uint32_t dstX, const uint32_t dstY, int dstSize,
+        const int width, const int height, const bool wrap) {
+
+    int srcI = srcY * srcStride + srcX;
+    int dstI = dstY * dstStride + dstX;
+    int x, y;
+
+    if (wrap && false) {
```
jfirebaugh: ?

src/mbgl/util/scaling.hpp
```diff
@@ -10,10 +10,6 @@
 namespace mbgl {
 namespace util {
 
-void bilinearScale(const uint32_t* srcData, const vec2<uint32_t>& srcSize,
```
jfirebaugh: I believe this whole file can be deleted.

src/mbgl/sprite/sprite_atlas.cpp
ansis: oh, good catch. I'll remove and add a render test to catch this kind of bug: https://github.com/mapbox/mapbox-gl-test-suite/issues/72

ansis:
scaling.hpp is removed in 0930972

the wrapping bug is fixed in e16a487 and covered with a new render test: https://github.com/mapbox/mapbox-gl-test-suite/commit/a7e5668c001b6f63f618604629bb6fbbca935682

do you see anything else or can this be squashed and merged (once tests ci passes)?

jfirebaugh:
LGTM

PR #3516, opened 2016-01-14
have a more explicit Response representation for "304 Not Modified"

jfirebaugh:
Currently, the way that a 304 Not Modified response gets encoded into `Response` is that the `std::shared_ptr<const std::string>` data from the previous response is copied into the new one. This requires a subtle invariant to be preserved by `OnlineFileSource`, the resulting pointer equality comparisons can be mistaken for a bug, and it means that the raw response body must remain in memory in addition to any parsed representation.

Instead, we should have a way for a 304 Not Modified response to be more explicitly encoded into `Response`, and consumers should only have to maintain the etag and expires values to make a revalidating request and check if things changed.

cc @kkaefer @tmpsantos


jfirebaugh:
Converted to PR. CC @kkaefer, this should simplify #2721 a bit.

tmpsantos:
LGTM

PR #3543, opened 2016-01-14
place symbol layers in correct order

ansis:
Previously, the placement of symbol layers depended on the order in which the glyphs they need were loaded. The placement order should be based on the style.

To fix this, symbol layers are placed in `placementPending` after they are parsed. After all layers are parsed, symbol layers (stored in `placementPending`) are placed and returned.

@kkaefer can you review this?

### DIFF
--- a/src/mbgl/map/tile_worker.cpp
+++ b/src/mbgl/map/tile_worker.cpp
@@ -41,13 +41,11 @@ TileParseResult TileWorker::parseAllLayers(std::vector<std::unique_ptr<StyleLaye
                                            PlacementConfig config) {
     // We're doing a fresh parse of the tile, because the underlying data has changed.
     pending.clear();
+    placementPending.clear();
     partialParse = false;
 
     // Store the layers for use in redoPlacement.
     layers = std::move(layers_);
-
-    // Reset the collision tile so we have a clean slate; we're placing all features anyway.
-    collisionTile = std::make_unique<CollisionTile>(config);
 
     // We're storing a set of bucket names we've parsed to avoid parsing a bucket twice that is
     // referenced from more than one layer
@@ -62,10 +60,15 @@ TileParseResult TileWorker::parseAllLayers(std::vector<std::unique_ptr<StyleLaye
     }
 
     result.state = pending.empty() ? TileData::State::parsed : TileData::State::partial;
+
+    if (result.state == TileData::State::parsed) {
+        placeLayers(config);
+    }
+
     return std::move(result);
 }
 
-TileParseResult TileWorker::parsePendingLayers() {
+TileParseResult TileWorker::parsePendingLayers(const PlacementConfig config) {
     // Try parsing the remaining layers that we couldn't parse in the first step due to missing
     // dependencies.
     for (auto it = pending.begin(); it != pending.end();) {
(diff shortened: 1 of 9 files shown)


src/mbgl/map/tile_worker.cpp
```diff
@@ -142,14 +157,17 @@ void TileWorker::parseLayer(const StyleLayer* layer, const GeometryTile& geometr
...
 
-    if (layer->is<SymbolLayer>() && partialParse) {
-        // We cannot parse this bucket yet. Instead, we're saving it for later.
-        pending.emplace_back(layer->as<SymbolLayer>(), std::move(bucket));
+    if (layer->is<SymbolLayer>()) {
+        if (partialParse) {
+            // We cannot parse this bucket yet. Instead, we're saving it for later.
+            pending.emplace_back(layer->as<SymbolLayer>(), std::move(bucket));
+        } else {
+            placementPending.emplace(layer->bucketName(), std::move(bucket));
```
kkaefer: This will not overwrite existing data that has the same key in `placementPending`. Is this what you want?

ansis: Since label names are unique, there should never be anything to overwrite. Also, this is the same as [this](https://github.com/mapbox/mapbox-gl-native/blob/4515bb00c0fd6c90050cf3befc4e848aaa7362c0/src/mbgl/map/tile_worker.cpp#L1600)

kkaefer: :+1: just checking

ansis:
@jfirebaugh it would be great to get a third set of :eyes: on this if you have time

jfirebaugh:
Will do.

src/mbgl/map/tile_worker.cpp
```diff
@@ -41,14 +41,13 @@ TileParseResult TileWorker::parseAllLayers(std::vector<std::unique_ptr<StyleLaye
                                            PlacementConfig config) {
     // We're doing a fresh parse of the tile, because the underlying data has changed.
     pending.clear();
+    placementPending.clear();
     partialParse = false;
+    placementConfig = config;
```
jfirebaugh: Instead of storing a copy of `placementConfig`, would it be better to pass the correct value to `TileWorker::parsePendingLayers`, and then pass it through as a parameter to `placeLayers` everywhere?

src/mbgl/map/tile_worker.cpp
```diff
@@ -89,20 +92,32 @@ TileParseResult TileWorker::parsePendingLayers() {
...
     return std::move(result);
 }
 
+void TileWorker::placeLayers() {
+    redoPlacement(&placementPending, placementConfig);
+    for (auto &p : placementPending) {
+        p.second->swapRenderData();
+        insertBucket(p.first, std::move(p.second));
+    }
+}
```
jfirebaugh: Should this do `placementPending.clear()` before exiting? I don't think there's an execution path where `placementPending` can get reused afterwards but before `parseAllLayers` clears it, and we might as well reclaim the memory in the meantime.

src/mbgl/map/tile_worker.cpp
ansis: yes, that would be cleaner and it could avoid an extra `redoPlacement` in some cases. changed in 716b01c

src/mbgl/map/tile_worker.cpp
ansis: changed in a0ebbda

ansis:
is this ready for squashing and merging?

jfirebaugh:
LGTM

PR #3548, opened 2016-01-14
Support multiple paths in SharedSQLiteCache::get

kkaefer:
We're currently only supporting one path, and calling this function will always return the same cache, as long as it's held elsewhere. That means subsequent requests won't necessarily honor the path passed to `SharedSQLiteCache::get()`.

### DIFF
--- a/platform/default/sqlite_cache.cpp
+++ b/platform/default/sqlite_cache.cpp
@@ -12,6 +12,9 @@
 #include "sqlite3.hpp"
 #include <sqlite3.h>
 
+#include <unordered_map>
+#include <mutex>
+
 namespace {
 
 // The cache won't accept entries larger than this arbitrary size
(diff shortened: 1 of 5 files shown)


platform/default/sqlite_cache.cpp
```diff
@@ -470,16 +473,27 @@ void SQLiteCache::Impl::refresh(const Resource& resource, Seconds expires) {
     }
 }
 
-std::shared_ptr<SQLiteCache> SharedSQLiteCache::get(const std::string &path) {
-    std::shared_ptr<SQLiteCache> temp = masterPtr.lock();
-    if (!temp) {
-        temp.reset(new SQLiteCache(path));
-        masterPtr = temp;
+std::shared_ptr<SQLiteCache> SQLiteCache::getShared(const std::string &path) {
+    static std::mutex sharedMutex;
```
jfirebaugh: Function level statics are initialized upon first execution, and there are no thread-safety guarantees in case of simultaneous first execution from multiple threads. This can be made 100% safe by moving the static to file level or using something like `pthread_once`.

jfirebaugh:
:+1: other than that.

platform/default/sqlite_cache.cpp
kkaefer: Good point, I tried to mitigate that by locking on the mutex, but clearly didn't think this through.

PR #3549, opened 2016-01-14
[android] #3450 - fix rotation when enabling user location

### LINKED ISSUE #3450: Android Bug: the map rotates on enabling tracking
Hi,
We are working on adding insets to the map (#2600) and we notice an unexpected behavior with the test app.

### How to:
After rotating the map, press the user location button several times.

### Observed:
The map rotates each time we press the user location.

### Expected:
The map should not rotate.

tobrun:
Closes #3450


tobrun:
CI approved going to merge

PR #3554, opened 2016-01-14
[ios] [docs] Wrote up some iOS documentation to avoid common mistakes.  

roblabs:
#### Updating the docs so that others can learn the build chain for iOS.

DEVELOPING.md & INSTALL.md could probably be combined, as they both had similar instructions.  Instead of removing them, I made similar comments in the Build section.

Best way to compare the changes in this PR is to _cf._ by hitting these links

DEVELOPING.md 
* Original — https://github.com/mapbox/mapbox-gl-native/blob/master/platform/ios/DEVELOPING.md
* PR —https://github.com/roblabs/mapbox-gl-native/blob/master/platform/ios/DEVELOPING.md

INSTALLmd
* Original — https://github.com/mapbox/mapbox-gl-native/blob/master/platform/ios/INSTALL.md
* PR —https://github.com/roblabs/mapbox-gl-native/blob/master/platform/ios/INSTALL.md


1ec5:
Thanks, really appreciate your eyes on these documents. It can be difficult for folks like me to spot these omissions because we’re used the environment we set up.

INSTALL.md was intended to be a simpler guide for folks who just want to use the SDK (and can’t wait for a release), so we omitted any mention of `make iproj` and `make itest`. Maybe we can avoid the duplication by simply linking INSTALL.md to DEVELOPING.md as “further reading”.

By the way, some of these steps will change once #3183 lands (hopefully very soon).

roblabs:
@1ec5, excellent.  We would be happy to review #3183 docs when it arrives.

We were also looking for docs or code snippets on how to use the [tiles folder found in iOS/benchmark/assets](https://github.com/mapbox/mapbox-gl-native/tree/master/ios/benchmark/assets/tiles).   Any examples on how to use those assets?  We get that download.sh is for downloading, but how to use?  I am aware of #584, so this seemed like "offline", too.

1ec5:
That folder is only used for the iOS benchmarking application – solely for performance testing purposes. As far as I know, our initial implementation of offline support will be much more sophisticated than what you see in that folder. (cc @adam-mapbox)

Issue #3031, opened 2015-11-13
Custom icon potentially causing crash on certain devices


danswick:
While testing the Mapbox Android SDK [custom marker example](https://www.mapbox.com/android-sdk/custom-marker-icon/), I ran into some errors running a [simple version](https://www.mapbox.com/android-sdk/custom-marker-icon/) on a Nexus 5x. Here are the steps I took: 

- Open the app as-is in AVD (both API 21 and 23): App installs and opens without error. Custom marker is present.
- Open the app as-is on my Nexus 5x via ADB: app installs and opens, but force closes right away. Nothing is printed to logcat in Android Studio.
- Open the app on my Nexus 5x with `mimpap` instead of `drawable` in `R.drawable.ic_directions_boat_black_18dp` (and created corresponding image files in the proper `mipmap_` directories): same result as above.
- Open the app on my Nexus 5x without `.icon(icon)`: app installs and opens without error.

cc: @zugaldia @bleege

danswick:
Thanks, @zugaldia. Remembered my USB-C cable today and here's the unfiltered log: 

```
11-17 18:07:51.706 32137-32137/com.mycompany.myfirstglapp A/libc: Fatal signal 6 (SIGABRT), code -6 in tid 32137 (ny.myfirstglapp)
11-17 18:07:51.810 3249-3249/? A/DEBUG: *** *** *** *** *** *** *** *** *** *** *** *** *** *** *** ***
11-17 18:07:51.810 3249-3249/? A/DEBUG: Build fingerprint: 'google/bullhead/bullhead:6.0/MDB08L/2343525:user/release-keys'
11-17 18:07:51.810 3249-3249/? A/DEBUG: Revision: 'rev_1.0'
11-17 18:07:51.810 3249-3249/? A/DEBUG: ABI: 'arm'
11-17 18:07:51.810 3249-3249/? A/DEBUG: pid: 32137, tid: 32137, name: ny.myfirstglapp  >>> com.mycompany.myfirstglapp <<<
11-17 18:07:51.810 3249-3249/? A/DEBUG: signal 6 (SIGABRT), code -6 (SI_TKILL), fault addr --------
11-17 18:07:51.826 3249-3249/? A/DEBUG:     r0 00000000  r1 00007d89  r2 00000006  r3 f753ab7c
11-17 18:07:51.826 3249-3249/? A/DEBUG:     r4 f753ab84  r5 f753ab34  r6 00000002  r7 0000010c
[...]

Issue #3537, opened 2016-01-13
Release Android 3.1.0


bleege:
### New Highlights

* [Camera API Callback Improvements](https://github.com/mapbox/mapbox-gl-native/issues/3412)
* [Coordinate Deprecated For LatLng](https://github.com/mapbox/mapbox-gl-native/issues/3309)
* [Responsive User Dot Location Tracking](https://github.com/mapbox/mapbox-gl-native/issues/2049)

Full Change Log: https://github.com/mapbox/mapbox-gl-native/issues?q=milestone%3Aandroid-v3.1.0

Release Branch: https://github.com/mapbox/mapbox-gl-native/commits/release-android-v3.1.0

### The TODO list for making this happen:

- [x] Change Version Name to `3.1.0` in [`gradle.properties`](https://github.com/mapbox/mapbox-gl-native/blob/925687ab06892528f25fd4a79d27a55560634d96/platform/android/MapboxGLAndroidSDK/gradle.properties#L2)
- [x] Update version info in [Twitter Fabric metadata file](https://github.com/mapbox/mapbox-gl-native/blob/925687ab06892528f25fd4a79d27a55560634d96/platform/android/MapboxGLAndroidSDK/src/main/resources/fabric/com.mapbox.mapboxsdk.mapbox-android-sdk.properties#L2)
- [x] Move open issues to `android-v4.0.0` Milestone in GitHub
- [x] Publish `3.1.0` artifact to Maven Central
- [x] Updating Mapbox.com JavaDoc Documentation
- [x] Update [CHANGELOG.md](https://github.com/mapbox/mapbox-gl-native/blob/master/CHANGELOG.md)
- [x] Tag `android-v3.1.0` in GitHub
- [x] Close Milestone `android-v3.1.0` in GitHub
- [x] Blog Post on https://www.mapbox.com/blog/
- [x] Alert Twitter Fabric team of new release
- [x] Create ticket to merge `release-android-v3.1.0` branch back into `master`
- [x] Change Version Name _BACK to SNAPSHOT after release and release branch merge_ to support daily development.  In this case set to  `4.0.0-SNAPSHOT` in [`gradle.properties`](https://github.com/mapbox/mapbox-gl-native/blob/925687ab06892528f25fd4a79d27a55560634d96/platform/android/MapboxGLAndroidSDK/gradle.properties#L2)

/cc @mapbox/mobile @mapbox/support @mapbox/sales @gundersen @lxbarth  @twbell

[...]

PR #3561, opened 2016-01-15
[core][ios][osx][android] fix icons with non-integer width/height

### LINKED ISSUE #2198: MGLAnnotationImage with scale greater than 1.0 causes crash.

ansis:
ref #3031
ref #2198

For example, an icon that has:
- a pixel width of 10
- a pixel ratio of 3
- a scaled with of 3.333

is now supported.

SpriteImage is now constructed with the pixel width of the image instead of the width.

this works in android with @tobrun's https://github.com/mapbox/mapbox-gl-native/tree/3031-test-branch-android
@1ec5 I think this should fix #2198 - what's the best way to test that?

### DIFF
--- a/platform/android/src/jni.cpp
+++ b/platform/android/src/jni.cpp
@@ -24,6 +24,7 @@
 #include <mbgl/platform/event.hpp>
 #include <mbgl/platform/log.hpp>
 #include <mbgl/storage/network_status.hpp>
+#include <mbgl/util/exception.hpp>
 
 #pragma clang diagnostic ignored "-Wunused-parameter"
 
@@ -1225,16 +1226,18 @@ void JNICALL nativeAddAnnotationIcon(JNIEnv *env, jobject obj, jlong nativeMapVi
 
     const std::string symbolName = std_string_from_jstring(env, symbol);
 
-    jbyte* pixelData = env->GetByteArrayElements(jpixels, nullptr);
     jsize size = env->GetArrayLength(jpixels);
-    std::string pixels(reinterpret_cast<char*>(pixelData), size);
-    env->ReleaseByteArrayElements(jpixels, pixelData, JNI_ABORT);
+    mbgl::PremultipliedImage premultipliedImage(width, height);
+
+    if (premultipliedImage.size() != uint32_t(size)) {
+        throw mbgl::util::SpriteImageException("Sprite image pixel count mismatch");
+    }
+
+    env->GetByteArrayRegion(jpixels, 0, size, reinterpret_cast<jbyte*>(premultipliedImage.data.get()));
 
     auto iconImage = std::make_shared<mbgl::SpriteImage>(
-        uint16_t(width),
-        uint16_t(height),
-        float(scale),
-        std::move(pixels));
+        std::move(premultipliedImage),
+        float(scale));
 
     nativeMapView->getMap().addAnnotationIcon(symbolName, iconImage);
 }
--- a/platform/ios/src/MGLMapView.mm
+++ b/platform/ios/src/MGLMapView.mm
@@ -2346,8 +2346,8 @@ std::chrono::steady_clock::duration MGLDurationInSeconds(float duration)
 
     // add sprite
     auto cSpriteImage = std::make_shared<mbgl::SpriteImage>(
-        uint16_t(annotationImage.image.size.width),
-        uint16_t(annotationImage.image.size.height),
+        uint16_t(width),
+        uint16_t(height),
         float(annotationImage.image.scale),
         std::move(pixels));
 
--- a/platform/ios/src/MGLMapView.mm
+++ b/platform/ios/src/MGLMapView.mm
@@ -2334,22 +2334,19 @@ std::chrono::steady_clock::duration MGLDurationInSeconds(float duration)
     size_t width = CGImageGetWidth(image);
     size_t height = CGImageGetHeight(image);
     CGColorSpaceRef colorSpace = CGColorSpaceCreateDeviceRGB();
-    std::string pixels(width * height * 4, '\0');
+    mbgl::PremultipliedImage cPremultipliedImage(width, height);
     size_t bytesPerPixel = 4;
     size_t bytesPerRow = bytesPerPixel * width;
     size_t bitsPerComponent = 8;
-    char *pixelData = const_cast<char *>(pixels.data());
-    CGContextRef context = CGBitmapContextCreate(pixelData, width, height, bitsPerComponent, bytesPerRow, colorSpace, kCGImageAlphaPremultipliedLast);
+    CGContextRef context = CGBitmapContextCreate(cPremultipliedImage.data.get(), width, height, bitsPerComponent, bytesPerRow, colorSpace, kCGImageAlphaPremultipliedLast);
     CGContextDrawImage(context, CGRectMake(0, 0, width, height), image);
     CGContextRelease(context);
     CGColorSpaceRelease(colorSpace);
 
     // add sprite
     auto cSpriteImage = std::make_shared<mbgl::SpriteImage>(
-        uint16_t(width),
-        uint16_t(height),
-        float(annotationImage.image.scale),
-        std::move(pixels));
+        std::move(cPremultipliedImage), 
+        float(annotationImage.image.scale));
 
     // sprite upload
     NSString *symbolName = [MGLAnnotationSpritePrefix stringByAppendingString:annotationImage.reuseIdentifier];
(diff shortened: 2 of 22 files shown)


jfirebaugh:
Approach here looks sound. I would like to continue by changing the `SpriteImage` constructor signature to:

```
SpriteImage(PremultipliedImage&&, float pixelRatio, bool sdf = false)
```

And replacing these `SpriteImage` members:

```
    const uint16_t width;
    const uint16_t height;
    const uint16_t pixelWidth;
    const uint16_t pixelHeight;
    const std::string data;
```

With:

```
const PremultipliedImage image;
```

1ec5:
> I think this should fix #2198 - what's the best way to test that?

Change [this line](https://github.com/mapbox/mapbox-gl-native/blob/f62d41ae05adade69fb08cff3c071b960c25c857/ios/app/MBXViewController.mm#L520) to:

```objc
CGRect rect = CGRectMake(0, 0, 20, 15.5);
```

1ec5:
Can you add a line to the Android and iOS sections of the changelog, too? Lots of folks are eager to see this fixed. :clap:

bleege:
Tagging myself here so that we can cherry pick this over to `release-android-3.1.0` branch when this lands.

ansis:
@jfirebaugh SpriteImage and PremultipliedImage changes implemented in 9cef5a7

platform/android/src/jni.cpp
```diff
@@ -1227,14 +1228,19 @@ void JNICALL nativeAddAnnotationIcon(JNIEnv *env, jobject obj, jlong nativeMapVi
 
     jbyte* pixelData = env->GetByteArrayElements(jpixels, nullptr);
     jsize size = env->GetArrayLength(jpixels);
-    std::string pixels(reinterpret_cast<char*>(pixelData), size);
     env->ReleaseByteArrayElements(jpixels, pixelData, JNI_ABORT);
```
jfirebaugh: This line needs to be moved to after the `std::copy`. (Actually I'm reading now that `GetByteArrayRegion` is a better API to use: http://developer.android.com/training/articles/perf-jni.html.)

platform/ios/src/MGLMapView.mm
```diff
@@ -2304,11 +2304,11 @@ - (void)installAnnotationImage:(MGLAnnotationImage *)annotationImage
...
     size_t height = CGImageGetHeight(image);
     CGColorSpaceRef colorSpace = CGColorSpaceCreateDeviceRGB();
-    std::string pixels(width * height * 4, '\0');
+    mbgl::PremultipliedImage cPremultipliedImage(width, height);
     size_t bytesPerPixel = 4;
     size_t bytesPerRow = bytesPerPixel * width;
     size_t bitsPerComponent = 8;
-    char *pixelData = const_cast<char *>(pixels.data());
+    char *pixelData = reinterpret_cast<char *>(cPremultipliedImage.data.get());
     CGContextRef context = CGBitmapContextCreate(pixelData, width, height, bitsPerComponent, bytesPerRow, colorSpace, kCGImageAlphaPremultipliedLast);
```
jfirebaugh: Avoid the cast and pass `cPremultipliedImage.data.get()` directly to `CGBitmapContextCreate`.

src/mbgl/annotation/annotation_manager.cpp
```diff
@@ -165,7 +165,7 @@ void AnnotationManager::removeIcon(const std::string& name) {
 
 double AnnotationManager::getTopOffsetPixelsForIcon(const std::string& name) {
     auto sprite = spriteStore.getSprite(name);
-    return sprite ? -sprite->height / 2 : 0;
+    return sprite ? -(std::ceil(sprite->image.height / sprite->pixelRatio)) / 2 : 0;
```
jfirebaugh: I'm not sure if this should use `std::ceil` or not.

test/sprite/sprite_store.cpp
```diff
@@ -13,9 +13,10 @@ using namespace mbgl;
 TEST(SpriteStore, SpriteStore) {
     FixtureLog log;
 
-    const auto sprite1 = std::make_shared<SpriteImage>(8, 8, 2, std::string(16 * 16 * 4, '\0'));
-    const auto sprite2 = std::make_shared<SpriteImage>(8, 8, 2, std::string(16 * 16 * 4, '\0'));
-    const auto sprite3 = std::make_shared<SpriteImage>(8, 8, 2, std::string(16 * 16 * 4, '\0'));
+    PremultipliedImage image(16, 16);
+    const auto sprite1 = std::make_shared<SpriteImage>(std::move(image), 2);
```
jfirebaugh: `move`ing something twice isn't valid. Instead do `std::make_shared<SpriteImage>(PremultipliedImage(16, 16), 2)`.

ansis:
@jfirebaugh thanks for finding all the bugs, they should be fixed now

jfirebaugh:
:+1:

bleege:
Should this be cherry picked into the `release-android-v3.1.0` branch?

ansis:
@bleege I think so. at least the first commit (26faa6a5ade54c0a423aab84106876dc59be868f).

tobrun:
I verified that this fixed the initial issue of having a rectangle image as marker icon

bleege:
Cool!  @tobrun confirmed that he tested from `master` and both  26faa6a5ade54c0a423aab84106876dc59be868f and d34f8eb674b9753c47616f37ae88ff7a02f61ba0 made it in there.  I'll cherry pick both to `release-android-v3.1.0` for today's release at #3537 .

Thanks everyone!

bleege:
FYI... I fudged the cherry pick process for the first commit 26faa6a5ade54c0a423aab84106876dc59be868f and ended up creating a new commit after resolving the changes instead of doing a `git cherry-pick --continue` hence 52fd9d7b505559972654a9f7af05d6be9c244f73.  I did it the proper way for the second commit however.  Discussed with @zugaldia this morning and we decided to leave this as is (and not deal with force pushes and other crazy git commands to be named later) and deal with the pain when the merge of `release-android-v3.1.0` into `master` after release.

danswick:
Huge! Our mobile SDK users will be pumped when this lands. :clap:

bleege:
@danswick FYI... we had to pull this from Android `3.1.0` today due to C++ Pain.  It's now scheduled for `4.0.0`.

afathman:
When is the expected release of 4.0.0?

bleege:
@afathman We're still planning the specifics around `4.0.0`, but in the interim the fix for this issues is available immediately for use via the nighlty `4.0.0-SNAPSHOT` builds.

Issue #2150, opened 2015-08-21
Get tests to work on iOS/Android


kkaefer:
We should run all available test suites on each platform.

| Platform | [Core tests](https://github.com/mapbox/mapbox-gl-native/tree/master/test) | [Rendering tests](https://github.com/mapbox/mapbox-gl-test-suite) | Platform tests |
| --- | --- | --- | --- |
| Android | ✘ #4671 | ✘ #2488 | ✔︎ |
| iOS | [✔︎](https://github.com/mapbox/mapbox-gl-native/pull/4574) ([Regressed](https://github.com/mapbox/mapbox-gl-native/issues/6423)) | ✘ #2488 | ✘ #4495, #2734, etc. |
| OS X | ✔︎ | ✘ #2488 | ✔︎ |
| Linux | ✔︎ | ✘ #2488 | n/a |
| Node OS X | n/a | [✔︎](https://github.com/mapbox/mapbox-gl-native/pull/4560) | ✔︎ |
| Node Linux | n/a | ✔︎ | ✔︎ |

ljbade:
On Android side there have been a few tickets in this area: https://github.com/mapbox/mapbox-gl-native/issues/1667

ljbade:
And https://github.com/mapbox/mapbox-gl-native/issues/1898

tobrun:
Closing this down because we have UI testing frameworks for both platforms:
 - Espresso on Android 
 - KIF on iOS

jfirebaugh:
Technically his is still relevant: in a perfect world we would run [the headless rendering tests](https://github.com/mapbox/mapbox-gl-test-suite) on every platform. It's low priority / not actionable right now though.

tobrun:
@jfirebaugh on Android we have Unit tests using JUnit, these are executed on each commit to this repo.
Coverage is not that high but we are working on this to improve.

@1ec5 / @friedbunny does iOS have this?

friedbunny:
iOS has [functionality tests](https://github.com/mapbox/mapbox-gl-native/blob/c10f6489e61a52e3ae36af7cd2c40d120f9181da/test/ios/MapViewTests.m) using KIF.

kkaefer:
@tobrun KIF/JUnit are running integration tests. This issue is about unit tests, that independently test parts of the system. Integration tests may not cover everything tested by unit tests (and vice versa).

1ec5:
[...]

PR #3571, opened 2016-01-15
Fix ASSET=zip on OS X + Linux

kkaefer:
It's currently not even compiling, since we're not install libzip.


jfirebaugh:
I'd rather see the zip implementation become firmly Android-only and fix #2150.

tmpsantos:
It is Android-only after 86c8446d3a4390ff6577d070ac8b5fa3ad3c5cd1

PR #3575, opened 2016-01-15
Correctly update cache for responses that sent a modified/expires time of 0

kkaefer:
We're currently treating `0` as a special value meaning "not set" when we get 304 Not Modified response, but `0` is also a valid response value.

### DIFF
--- a/platform/darwin/http_request_nsurl.mm
+++ b/platform/darwin/http_request_nsurl.mm
@@ -231,15 +231,12 @@ void HTTPNSURLRequest::handleResult(NSData *data, NSURLResponse *res, NSError *e
             if (existingResponse) {
                 response->data = existingResponse->data;
 
-                if (response->expires == Seconds::zero()) {
-                    response->expires = existingResponse->expires;
-                }
-
-                if (response->modified == Seconds::zero()) {
+                // Only copy the existing headers if we didn't get one during this response.
+                if (!last_modified) {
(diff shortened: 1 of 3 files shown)


jfirebaugh:
Hmm, I find this solution confusing because it makes `Seconds::zero()` mean "no expiration" in some cases and "expires far in the past" in others.

kkaefer:
How so? We don't have a notion of "no expiration" when receiving HTTP responses. When the server doesn't send an expires header, we're using the value `0`, which we treat as "expires immediately".

jfirebaugh:
https://github.com/mapbox/mapbox-gl-native/blob/master/platform/default/online_file_source.cpp#L217-L218 - the comment there is based on my interpretation of this prior code: https://github.com/mapbox/mapbox-gl-native/blob/a06f28ddb70cd1963076d6df2ef3716d45b7a13e/platform/default/online_file_source.cpp#L430 and the fact that the default value for `Response::expires` is zero. Is it wrong?

I would like to clearly distinguish "response without an `Expires` or `Cache-Control` header" from "response with cache headers indicating immediate expiration", which is why I suggest using `optional`, with empty representing the former and 0 the latter.

jfirebaugh:
See #3611 for my preferred fix.

jfirebaugh:
Fixed by #3611.

PR #3576, opened 2016-01-15
Cleanup std::chrono usage

brunoabinader:
We have abstractions for `std::chrono` in `mbgl/util/chrono.hpp` for most common usages. This patch series reduces `std::chrono` cast complexity and usage to `include/mbgl/util/chrono.hpp`.

:eyes: @kkaefer @jfirebaugh @tmpsantos

### DIFF
--- a/include/mbgl/map/map.hpp
+++ b/include/mbgl/map/map.hpp
@@ -125,8 +125,8 @@ public:
     void setBearing(double degrees, const PrecisionPoint&, const Duration& = Duration::zero());
     void setBearing(double degrees, const EdgeInsets&, const Duration& = Duration::zero());
     double getBearing() const;
-    void resetNorth(const Duration& = std::chrono::milliseconds(500));
-    void resetNorth(const EdgeInsets&, const Duration& = std::chrono::milliseconds(500));
+    void resetNorth(const Duration& = Milliseconds(500));
+    void resetNorth(const EdgeInsets&, const Duration& = Milliseconds(500));
 
     // Pitch
     void setPitch(double pitch, const Duration& = Duration::zero());
--- a/ios/benchmark/MBXBenchViewController.mm
+++ b/ios/benchmark/MBXBenchViewController.mm
@@ -60,13 +60,6 @@
 
 }
 
-using Clock = std::chrono::steady_clock;
-using TimePoint = Clock::time_point;
-template <typename Duration>
-inline int64_t Microseconds(Duration d) {
-    return std::chrono::duration_cast<std::chrono::microseconds>(d).count();
-}
-
 size_t idx = 0;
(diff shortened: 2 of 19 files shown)


jfirebaugh:
We're using `std::chrono::seconds`, `std::chrono::milliseconds`, etc. wrong. The intent of introducing `mbgl::Duration` was to have _one_ duration type which we use everywhere a duration is required. Adding `mbgl::Seconds`, `mbgl::Milliseconds`, etc., and converters such as `asSeconds`, `toSeconds` is a mistake.

The _only_ times `std::chrono::seconds` and so on should be used are:

* When constructing a constant, e.g. `std::chrono::seconds(10)`. (Or, if you prefer, `using std::literals::chrono_literals` and `10s` instead.)
* When converting an integer type with implied units from an external data or an external API.

In both of those cases the expression should be immediately assigned to a `mbgl::Duration` instance, letting the constructor handle conversion.

brunoabinader:
The main goal for establishing aliases and helper functions is consistency. We either follow `mbgl` or `std::chrono` directly, but having a mix is bad.

> The intent of introducing mbgl::Duration was to have one duration type which we use everywhere a duration is required.

Yes, however, as you mentioned, using helpers like `std::chrono::{,micro,milli}seconds` or `std::literals::chrono_literals` are still required for data input.

> Adding mbgl::Seconds, mbgl::Milliseconds, etc., and converters such as asSeconds, toSeconds is a mistake.

Are you suggesting we should get rid of our helpers and use e.g. `std::chrono::seconds`, `std::chrono::duration_cast` directly?

> We're using std::chrono::seconds, std::chrono::milliseconds, etc. wrong.

What do you have in mind then for cases where output is required to be in a time unit that is not seconds - e.g. https://github.com/mapbox/mapbox-gl-native/blob/master/platform/default/timer.cpp#L69 ?

kkaefer:
> The main goal for establishing aliases and helper functions is consistency. We either follow mbgl or std::chrono directly, but having a mix is bad.

It's a little inconsistent, but given that we just have type aliases, it's not a problem per se.

I think @jfirebaugh wants to say that we should stay with or move back to `std::chrono::duration` (or a `mbgl::Duration` alias of it), instead of using `std::chrono::milliseconds` (or any aliased types) for storing time duration. When we need to output the actual number, we should cast directly at the place where need it. We could helper functions or aliases for this as well.

jfirebaugh:
Right @kkaefer.

* For points in time, use `mbgl::TimePoint`.
* For spans of time, use `mbgl::Duration`.
* If you need to convert to and from these types, do it as far towards the outer edge as possible -- immediately after input, immediately before output.

It looks like the main issue here is `Response::modified` and `Response::expires`. Modification time and expiration time are both _points in time_, not durations: they should be `mbgl::TimePoint`s.

kkaefer:
Looks good!

PR #3555, opened 2016-01-14
[ios] remove `testCompassTap` test

friedbunny:
Remove test that fails on CI because of KIF unreliability.

Refs 61615a44fd49336905e7b18c8ac665ce1eb7d4a4, #2769, #2734.

/cc @1ec5

PR #2769, opened 2015-10-23
fixes #2734: improve iOS testing

incanus:
- Got rid of the heading tracking test. Mocking this is proving too complicated (plus I turned up two GL library bugs in the process) and we do builds often enough to keep manual tabs on this functionality. I just don't think it's worth the hassle to automate, especially given the low risk of functionality breakage. 

- Fixed said tracking-related bugs in `MGLMapView`. 

- Removed the external location mocking library to make this more like other test suites. 

- Moved to `Debug` since it runs faster these days. 

- Test tweaks for more reliability. 

- Enabled all tests. In future, if tests are flapping, we should remove them and ticket replacement, not disable them. 

Going to squash/rebase/finalize etc. before merging. 

/cc @friedbunny @1ec5

PR #3581, opened 2016-01-15
[ios] remove location delegate test to improve reliability

friedbunny:
Continue the whack-a-mole with iOS tests.

Refs 61615a4, #2769, #2734, #3555

/cc @1ec5

### DIFF
--- a/test/ios/MapViewTests.m
+++ b/test/ios/MapViewTests.m
@@ -48,15 +48,6 @@
     tester.mapView.delegate = self;
 }
 
-- (void)approveLocationIfNeeded {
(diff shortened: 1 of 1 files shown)


1ec5:
Sigh.

friedbunny:
Alright, so this appears to have been the wrong approach and iOS is still failing on Bitrise:

```
✗ testDelegateRegionDidChange, The step timed out after 10.00 seconds: Application is ignoring interaction events
✗ testDelegateRegionDidChange, Failure in child step: The step timed out after 10.00 seconds: Application is ignoring interaction events
```

PR #3583, opened 2016-01-15
Content insets

### LINKED ISSUE #2600: Offset map center

### LINKED ISSUE #3574: [core] Map::latLngForPixel and Map::pixelForLatLng are using bottom/left coordinates
The other `Map::` methods are using top/left coordinates which is our convention.

/cc @1ec5

1ec5:
This PR implements content insets on iOS and OS X. By default, the content insets automatically adjust to account for translucent chrome (on iOS, the top and bottom bars when there’s extended layout; on OS X, the translucent toolbar of a window that has the map view as a full-window content view). However, it can be configured for any purpose, such as excluding a Google-style floating search bar from the logical viewport.

I attempted to mimic UIScrollView and NSScrollView on iOS and OS X, respectively. Therefore, there are a few minor differences in the implementations between the two platforms. Whereas the map view itself has an `automaticallyAdjustsContentInsets` property on OS X, it’s the containing view controller that feeds the map view its insets on iOS.

The underlying implementation in mbgl is designed to be usable by Android as well, and eventually course tracking mode will shift the center downward so that the user dot appears towards the bottom of the view.

Following this work, the rule is that `Map` and `Transform` assume a view origin at the top-left corner of the view (consistent with iOS, Android, and GLFW), while `TransformState` assumes a view origin at the bottom-left corner of the view (more convenient for converting to geographic coordinates). (OS X natively assumes an origin at the bottom-left.)

* [x] `CameraOptions.padding`
* [x] Inset content view on OS X for translucent toolbar
* [x] Flip origin for `Map::latLngForPixel()` and `Map::pixelForLatLng()` (fixes #3574)
  * @bleege and @tmpsantos, please ensure that I’ve updated the Android and GLFW code correctly.
* [x] Fix annotation origin flipping regression on iOS
* [x] Merge Mappy/mapbox-gl-native@2d7b54ed6defe1bbba597a02e9b71883669280b9 by @lgeromegnace for iOS support
* [x] Shift the center on iOS for top and bottom bars
* [x] Fix bouncing during user tracking mode
* [ ] ~~Shift the center downward during course tracking mode~~ (split out into #3589)
  * The center gravity should likely be configurable separately from the content insets.

/ref #2600, #3574
/cc @friedbunny


1ec5:
To keep the diffs in this PR coherent, I split out #3589 to push the user puck downward in course tracking mode.

src/mbgl/util/geo.cpp
```diff
@@ -28,4 +28,16 @@ LatLngBounds::LatLngBounds(const TileID& id)
...
+PrecisionPoint EdgeInsets::getCenter(uint16_t width, uint16_t height) const {
+    return {
+        (width - left - right) / 2.0f + left,
+        (height - top - bottom) / 2.0f + top,
+    };
+}
+
+void EdgeInsets::flip() {
+    std::swap(top, bottom);
+    std::swap(left, right);
```
1ec5: [Some discussion on this line](https://github.com/mapbox/mapbox-gl-native/commit/eaeef958a5357713dd19c7031789399d6a883a76#commitcomment-15509284).

platform/ios/src/MGLMapView.mm
```diff
@@ -770,6 +773,49 @@ - (void)layoutSubviews
     }
 }
 
+/// Updates `contentInset` to reflect the current window geometry.
+- (void)adjustContentInset
+{
```
mb12: Does this handle the case where  MGLMapView is nested more than 2 level deep in the UIViewController view hierarchy?

1ec5: It [doesn’t](https://github.com/mapbox/mapbox-gl-native/pull/3583/files#diff-20502b6bc02986a4398d0da84b2d4b94R779), for consistency with UIScrollView and `UIViewController.automaticallyAdjustsScrollViewInsets`. (This is already more robust than the UIScrollView behavior; a UIScrollView has to be the first subview.) Regardless, a developer can manually set the `contentInset` of a deeply nested MGLMapView.

lgeromegnace:
Hi @1ec5,
First of all: great work! Insets work fine :). 

I have a question: are you (or someone else) planning to implement this on the android side ? 
If not we can take care of it. Just let me know if you have specific needs for this platform.

1ec5:
Android support is tracked by #3627. I'd be happy to let someone more familiar with Android take this on. @tobrun, are there any special considerations for Android support?

tobrun:
@lgeromegnace feel free to take that up, it's on my list but currently not ranked in the top tier.

lgeromegnace:
Ok @1ec5, @tobrun.

PR #3588, opened 2016-01-16
Refactor unrotation

### LINKED ISSUE #3587: Bearing/course tracking mode disables when facing north
Turning your device so that it faces within 7° of true north kicks MGLMapView out of bearing or course tracking mode back into ordinary tracking mode. As a result, it’s difficult to stay in bearing or course tracking mode in the iOS Simulator, which never reports a bearing other than due north.

This regression was introduced in #3403. The extra step added to `-[MGLMapView unrotateIfNeededAnimated:]` should check the user tracking mode.

/cc @friedbunny

1ec5:
This PR refactors unrotation on iOS. There are two flavors of unrotation:

* At low zoom levels (below z3), the map unrotates to avoid showing too much emptiness in polar regions. However, this unrotation should only occur due to a gesture by the user, not due to the developer rotating the map programmatically (compare with MapKit).
* When the map is rotated within 7° of true north, the map unrotates for aesthetic reasons. However, this unrotation should only occur due to a gesture by the user, not due to programmatic rotation or an automatic rotation in heading or course tracking mode. We should avoid extraneous unrotation because it kicks the user out of heading or course tracking mode.

Fixes #3587.

/cc @friedbunny


platform/ios/src/MGLMapView.mm
```diff
@@ -3139,18 +3139,30 @@ - (BOOL)isRotationAllowed
...
+        && state != UIGestureRecognizerStateBegan
+        && state != UIGestureRecognizerStateChanged)
+    {
+        [self unrotateIfNeededAnimated:YES];
+        
+        // Snap to north.
+        if ((self.direction < MGLToleranceForSnappingToNorth
+             || self.direction > 360 - MGLToleranceForSnappingToNorth)
+            && self.userTrackingMode != MGLUserTrackingModeFollowWithHeading
+            && self.userTrackingMode != MGLUserTrackingModeFollowWithCourse)
```
friedbunny: Good catch. I was initially unsure about needing to check for the tracking mode here, but two-finger zoom and quickzoom both require it.

friedbunny:
Thanks! :ok_hand:

Issue #3154, opened 2015-11-30
Annotation callout closes automatically when changing the viewport


incanus:
Android now does this. 

Complicated by the fact that `SMCalloutView` doesn't support a move API, just a placement/initial layout. On raster we used to use this repeatedly during viewport changes to sync up, but it became non-performant as of last check. 

Refs https://github.com/mapbox/mapbox-gl-native/issues/1125 https://github.com/mapbox/mapbox-gl-native/pull/2412

1ec5:
Also leave it open when the user location changes: #2618.

silviokuehn:
I have the same problem. Auto-closing of the annotation callouts while after the map was moved (by app or manually by user interaction) seems to be a bad user experience. Moreover when you have a click event (button) within the callout bubble and user does a little bit move the finger while clicking, the pop-up disappears and the click event does not work.

As a workaround we tried a re-select of last selected annotation after a map move but that does not work really good.

1ec5:
As demonstrated by #1582, all we need to do is set `self.calloutViewForSelectedAnnotation.frame`, potentially inside a UIView animation block, instead of closing the callout view as we do now.

1ec5:
Fixed in #6676 on the v3.4.0 release branch.

Issue #2618, opened 2015-10-14
User dot callout view dismisses when user location changes slightly


1ec5:
On iOS, the user dot's callout view is automatically dismissed whenever the user location changes. It should not be dismissed when the location change is small, because that's usually just the result of Core Location refining its location.

/cc @friedbunny

1ec5:
Keeping the callout view open in this case does not expose any view syncing performance issues, because the user dot itself isn't animated (see #1582).

friedbunny:
Sounds good to me. We show the accuracy pulse (`haloLayer`) [when location accuracy is <10 meters](https://github.com/mapbox/mapbox-gl-native/blob/5b9deae95709ea557f28e4a5dcbe3ac8fd47023e/platform/ios/MGLMapView.mm#L2776-L2777), but I'm thinking perhaps an additional screen distance threshold may be better for the callout — unless potentially jumping all the way across the screen doesn't look that bad.

incanus:
We should just ensure that the dot isn't in fact moving perceptibly, since we can't yet reposition callouts (or other views) synced to the map yet (https://github.com/mapbox/mapbox-gl-native/issues/1125).

1ec5:
@incanus, I don't think that's a problem in this case: https://github.com/mapbox/mapbox-gl-native/issues/2618#issuecomment-148074417.

incanus:
Right, so I guess we use `-[MGLMapView metersPerPixelAtLatitude:]` combined with changes of < 1 pixel.

friedbunny:
Is this adequately addressed by your recent changes in #3589, @1ec5?

1ec5:
No, not until #1582 lands.

1ec5:
I stand corrected: even though the user dot and its callout don’t animate around yet outside of user tracking mode, the callout view does stay open. Closing.

PR #3589, opened 2016-01-17
Affix user dot on screen in user tracking mode

### LINKED ISSUE #1041: User location tracking is rough

### LINKED ISSUE #2600: Offset map center

1ec5:
Place the user dot at a fixed location on screen in user tracking mode, eliminating the jumping effect between user location updates. ~~In course tracking mode, weight the user puck down towards the bottom of the view, since it’s more important to see the road ahead than to see whoever’s tailgating you.~~ Added a new property to MGLMapView that controls the position of the user dot or user puck. By default, the user dot/puck is fixed in the center of the view, but it can be shifted to the top or bottom (e.g. for a GPS-style turn-by-turn view).

Conceptually, we now rely on entity interpolation to smooth out the large jumps between location updates. As in MapKit, the user dot may be up to one second behind where Core Location thinks the device is. When user tracking is initially enabled, MGLMapView “flies” to the first location reported by Core Location. This process may take longer than a second, but it gives the user more context as to their current location than the standard easing that MapKit performs.

* [x] Affix user dot
* [x] Fly to initial user location
* [x] Allow the user dot’s callout view to stay open (#2618, #3154)
* [x] Weight user puck
* [x] Make user dot location configurable

Depends on #3583. Fixes #1041 and fixes #2600. #1582 complements this PR by smoothing the user dot’s motion outside of user tracking mode, using view-based animation.

/cc @friedbunny @zugaldia


platform/ios/src/MGLMapView.mm
```diff
@@ -3175,7 +3276,9 @@ - (void)unrotateIfNeededAnimated:(BOOL)animated
             [self resetNorthAnimated:NO];
         }
     }
-    else if (self.direction < snapTolerance || self.direction > 360 - snapTolerance)
+    else if ((self.direction < snapTolerance || self.direction > 360 - snapTolerance)
+             && self.userTrackingMode != MGLUserTrackingModeFollowWithHeading
+             && self.userTrackingMode != MGLUserTrackingModeFollowWithCourse)
```
1ec5: Note to self to remove this change once #3588 lands.

1ec5:
Tentatively putting on 3.1.0, since it’s ready. Just needs lots of testing.

platform/ios/src/MGLMapView.mm
```diff
@@ -3024,50 +3090,65 @@ - (void)locationManager:(__unused CLLocationManager *)manager didUpdateLocations
...
-                    // assumes we won't disrupt tracking mode
-                    [self setVisibleCoordinateBounds:MGLCoordinateBoundsMake(desiredSouthWest, desiredNorthEast) edgePadding:UIEdgeInsetsZero direction:course animated:YES];
-                    [self unrotateIfNeededAnimated:YES];
-                }
+                self.userTrackingState = MGLUserTrackingStateBegan;
+                
+                MGLMapCamera *camera = self.camera;
+                camera.centerCoordinate = self.userLocation.location.coordinate;
+                camera.heading = course;
+                camera.altitude = newLocation.horizontalAccuracy;
```
1ec5: This approach seems to be more reliable, and it also works when the map is tilted, something unsupported by the old code, which was ported from the raster SDK, which didn’t support tilting.

platform/ios/src/MGLMapView.mm
```diff
@@ -3003,10 +3069,10 @@ - (void)locationManager:(__unused CLLocationManager *)manager didUpdateLocations
 
     if (! oldLocation || ! CLLocationCoordinate2DIsValid(oldLocation.coordinate) || [newLocation distanceFromLocation:oldLocation])
     {
-        self.userLocation.location = newLocation;
-
-        // deselect user if applicable since we don't do callout tracking yet
-        if ([self.selectedAnnotation isEqual:self.userLocation]) [self deselectAnnotation:self.userLocation animated:NO];
+        if (self.userTrackingState != MGLUserTrackingStateBegan)
+        {
+            self.userLocation.location = newLocation;
```
friedbunny: This means we potentially drop a fresher location if the old transition is still in progress?

platform/ios/src/MGLMapView.mm
friedbunny: Happy that this is simplified! The camera movement is much slicker, too.

One continued quibble is that highly-accurate location fixes zoom in too far and context is lost. This is somewhat mitigated by the camera animation, but perhaps we could also adopt a minimum altitude?

We could also consider only zooming in if the map view's zoom level is relatively low; i.e., Apple Maps doesn't change the altitude/zoom unless the map is zoomed out to ~z8.

friedbunny: I'm also seeing this zoom in beyond the accuracy ring, which is unintuitive.

friedbunny:
Quick zoom doesn't appear to take into account the content insets, which means the user dot is moved during this gesture while user tracking is enabled... and then moved back to the proper placement afterwards.

ezgif com-optimize

platform/ios/src/MGLMapView.mm
1ec5: Only for the transition to the initial location when user tracking is switched on. That transition may take longer than a second and shouldn't be interrupted.

platform/ios/src/MGLMapView.mm
1ec5: You're right, it does always seem to zoom into z18 for some reason. Perhaps it's because iOS-specific code assumes a field of view of 30 degrees, which is less than what some mbgl calculations assume.

1ec5:
Ah, good catch. All the gestures need to be updated to account for the content inset.

Additionally, heading changes, which come in much more frequently than location changes, tend to interrupt any running animation to the new location during heading tracking mode. The animation for heading changes needs to also pan to the current location, not just rotate.

1ec5:
> Quick zoom and two-finger zoom do not appear to take into account the content insets

Fixed in d32e0e6ed7d1a27b31b374e34a6fac1ea48551ee. In that change, for MapKit parity, I also removed the logic that kicks you out of user tracking mode if your zoom gesture is centered somewhere other than the user dot.

1ec5:
Failing builds are due to #1828.

1ec5:
> Additionally, heading changes, which come in much more frequently than location changes, tend to interrupt any running animation to the new location during heading tracking mode.

This is also the current behavior. Fixing it will require #3625 and is out of scope for this PR.

platform/ios/src/MGLMapView.mm
1ec5: Fixed in dbfc363a7a58307ebcdf4696a9bca89d0401048d. I haven’t nailed down exactly what MapKit does, but it does seem to like z14 quite a lot.

platform/ios/src/MGLMapView.mm
friedbunny: The first location is also likely to be relatively inaccurate and get preempted quickly by another update. Don't think this is a blocker here, just another reason to do #3610.

1ec5: MapKit does a simple, easeTo-style animation to the first location, which makes it easier to accommodate intervening updates. By contrast, this branch uses a Google-style flyTo, which would be easy to modify mid-flight even after implementing #3610.

In their own ways, both MapKit and Google Maps favor context and continuity over up-to-the-second precision. If a particular application needs to go immediately to the actual user location, it can pass NO into -setUserTrackingMode:animated:, which is being made public in this PR.

friedbunny:
YEAAAAAAAAAAH

1ec5:
Everything passes; merging.

PR #3599, opened 2016-01-18
WebP support

### LINKED ISSUE #3572: WebP support
[WebP](https://developers.google.com/speed/webp/?hl=en) is an image format employing both lossy and lossless compression. WebP is both efficient like JPEG and supports alpha like PNG.

We currently support WebP server-side, as well as [Mapbox GL JS](https://github.com/mapbox/mapbox-gl-js/issues/1725). [WebP 0.4.2](https://github.com/mapbox/mason/tree/webp-0.4.2) is the current version listed in our mason package manager. We should detect WebP support and use WebP tiles if the server advertises them.

Overall plan is:
- [x] Generate a mason branch for WebP 0.5.0 - latest stable version listed [here](https://storage.googleapis.com/downloads.webmproject.org/releases/webp/index.html).
- [x] Add WebP dependency in Mapbox GL native build system.
- [x] Provide support for WebP in `mbgl::decodeImage`.
- [ ] Modify sources for default styles in [Mapbox GL styles](https://github.com/mapbox/mapbox-gl-styles) to use WebP instead of PNG.

@mapbox/gl @mapbox/mobile

brunoabinader:
This provides WebP support for `linux`, `osx` and `ios` targets.

We use WebP library - currently in 0.5.0 version - to decode WebP and generate our premultiplied images. Since CoreFoundation does not decode WebP _yet_, we're adding WebP as a dependency for OS X & iOS builds.

Closes #3572.

@kkaefer @jfirebaugh @tmpsantos


friedbunny:
Have you noticed any practical performance differences (positive or negative) when using WebP?

1ec5:
I’d like to leave this change out of the upcoming release of Mapbox iOS SDK until we have a better understanding of the performance implications (memory, battery) of decoding images non-natively, particularly in raster styles like Satellite.

brunoabinader:
> Have you noticed any practical performance differences (positive or negative) when using WebP?

This just came out of the oven, so we have yet to perform performance evaluations, both in 1) network consumption and 2) runtime spent on decoding WebP tiles vs. PNG or JPEG.

> I’d like to leave this change out of the upcoming release of the Mapbox iOS SDK

Fair enough, I'll integrate the linux implementation and keep the iOS and OS X implementations on a separate branch until we receive the green signal :+1:.

PR #1732, opened 2016-01-19
Permit use of variant in public headers

jfirebaugh:
Currently, `style_properties.hpp` includes the variant code and is required for by the public API. This means the user has to include the variant as well. We should remove this requirement and make the variant code private.


jfirebaugh:
Going with #3616 for now. Will revisit if we need variant support later on.

PR #3617, opened 2016-01-19
[ios] Fixes crash when vendorId is nil

### LINKED ISSUE #3430: Crash in -[MGLMapboxEvents pushTurnstileEvent]
I'm guessing this has to do with the IDFV not being set, which [Apple says](https://developer.apple.com/library/ios/documentation/UIKit/Reference/UIDevice_Class/index.html#//apple_ref/occ/instp/UIDevice/identifierForVendor) can theoretically happen:

> If the value is nil, wait and get the value again later. This happens, for example, after the device has been restarted but before the user has unlocked the device.

```objc
Fatal Exception: NSInvalidArgumentException
*** -[__NSPlaceholderDictionary initWithObjects:forKeys:count:]: attempt to insert nil object from objects[3]

Thread : Fatal Exception: NSInvalidArgumentException
0  CoreFoundation                 0x30ed1f83 __exceptionPreprocess
1  libobjc.A.dylib                0x3b682ccf objc_exception_throw
2  CoreFoundation                 0x30e10413 -[__NSPlaceholderDictionary initWithObjects:forKeys:count:]
3  CoreFoundation                 0x30e101db +[NSDictionary dictionaryWithObjects:forKeys:count:]
4  appname                        0x2fd0ad __37-[MGLMapboxEvents pushTurnstileEvent]_block_invoke
5  libdispatch.dylib              0x3bb6a833 _dispatch_call_block_and_release
6  libdispatch.dylib              0x3bb6aded _dispatch_queue_drain$VARIANT$up
7  libdispatch.dylib              0x3bb6b297 _dispatch_queue_invoke$VARIANT$up
8  libdispatch.dylib              0x3bb7d88d _dispatch_root_queue_drain
9  libdispatch.dylib              0x3bb7db21 _dispatch_worker_thread2
10 libsystem_pthread.dylib        0x3bcacbd3 _pthread_wqthread
11 libsystem_pthread.dylib        0x3bcaca98 start_wqthread
```

From v2.1.2, but likely still possible.

/cc @1ec5

csmulhern:
This fixes #3430.

### DIFF
--- a/platform/ios/src/MGLMapboxEvents.m
+++ b/platform/ios/src/MGLMapboxEvents.m
@@ -402,6 +402,9 @@ const NSTimeInterval MGLFlushInterval = 60;
 
             // Build only IDFV event
             NSString *vid = [[[UIDevice currentDevice] identifierForVendor] UUIDString];
+
+            if (!vid) return;
+
             NSDictionary *vevt = @{
                 @"event" : MGLEventTypeAppUserTurnstile,
                 @"created" : [strongSelf.rfc3339DateFormatter stringFromDate:[NSDate date]],


friedbunny:
Thanks for this @csmulhern. We'll want to think more about deferring the turnstile event, but this fix is great in the meantime.

friedbunny:
Committed as 07f6dee616cda17e020b32a2f0b8caca76aa621d.

PR #3618, opened 2016-01-19
[ios] Updates configure to work when project is a submodule

csmulhern:
The configure bash script is currently checking for the existence of the .git directory to see if it should install mason dependencies within the current directory. We use mapbox-gl-native as a submodule which breaks this check.


1ec5:
Merged in a5889409cc9dd5e0cc973cdb81d69bd045bd6094. Thanks!

PR #3619, opened 2016-01-19
[ios] Adds the ability to disable map deceleration

csmulhern:
This adds the ability to disable the deceleration animation that occurs when the user stops panning. This feature can be useful when trying to move a marker on the map.

### DIFF
--- a/include/mbgl/ios/MGLMapView.h
+++ b/include/mbgl/ios/MGLMapView.h
@@ -19,6 +19,10 @@ NS_ASSUME_NONNULL_BEGIN
 @protocol MGLOverlay;
 @protocol MGLCalloutView;
 
+extern const CGFloat MGLMapViewDecelerationRateNormal;
+extern const CGFloat MGLMapViewDecelerationRateFast;
+extern const CGFloat MGLMapViewDecelerationRateImmediate;
+
 /** The vertical alignment of an annotation within a map view. */
 typedef NS_ENUM(NSUInteger, MGLAnnotationVerticalAlignment) {
     /** Aligns the annotation vertically in the center of the map view. */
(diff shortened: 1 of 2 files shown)


1ec5:
Thanks for this PR. We call this animation “drifting”. Can you provide a little more context about how you’re implementing marker dragging, and what role the added property plays in that implementation?

csmulhern:
I'd be happy to switch out the term for "drifting" if you feel it is appropriate. We do pin placement similar to Uber where a pin overlay is fixed in the center of the view and you move the map around it. In this mode of operation it can be frustrating to not get high accuracy pin placement because of the drifting animation. Many users also have trouble not panning the map when they lift their finger which is exacerbated by the drifting animation.

1ec5:
Ah, now I understand: you were referring to a view stationed above the map, rather than an MGLAnnotation. Thanks for clarifying!

UIScrollView has a [`decelerationRate`](https://developer.apple.com/library/ios/documentation/UIKit/Reference/UIScrollView_Class/#//apple_ref/occ/instp/UIScrollView/decelerationRate) property that controls how soon the map comes to a stop. It would be pretty elegant if we could expose something like that in MGLMapView, and maybe some constants similar to `UIScrollViewDecelerationRateFast` to make the API simple to use.

csmulhern:
No problem. Yeah that's a great idea. I'd be happy to investigate this some more and look at implementing a similar API.

csmulhern:
I've updated the implementation to more closely reflect the decelerationRate API exposed by UIScrollView.

1ec5:
Thanks for your patience on this one. Let’s continue this over in #5504, where I cherry-picked your change and tidied up a few things.

csmulhern:
No problem, thanks for getting to it.

PR #3620, opened 2016-01-19
Refactor telemetry keys

friedbunny:
- Add `pluggedIn` key to non-location events
- Add `enabled.telemetry` key to turnstile (supercedes #3496)
- Remove `instance`
- Remove `bundleId`
- Remove `lat`,`lng` from interaction events
- Remove obsolete `enabled.email` key
- Make `pushEnabled` an instance method
- Move most static keys to `map.load`

/cc @1ec5 @mapbox/mobiledata @boundsj

### DIFF
--- a/platform/ios/src/MGLMapView.mm
+++ b/platform/ios/src/MGLMapView.mm
@@ -1056,7 +1056,7 @@ std::chrono::steady_clock::duration MGLDurationInSeconds(float duration)
 
     if (pan.state == UIGestureRecognizerStateBegan)
     {
-        [self trackGestureEvent:MGLEventGesturePanStart forRecognizer:pan];
+        [self trackGestureEvent:MGLEventGesturePanStart];
 
         self.userTrackingMode = MGLUserTrackingModeNone;
         
@@ -1114,7 +1114,7 @@ std::chrono::steady_clock::duration MGLDurationInSeconds(float duration)
 
     if (pinch.state == UIGestureRecognizerStateBegan)
     {
-        [self trackGestureEvent:MGLEventGesturePinchStart forRecognizer:pinch];
+        [self trackGestureEvent:MGLEventGesturePinchStart];
 
         self.scale = _mbglMap->getScale();
         
@@ -1185,7 +1185,7 @@ std::chrono::steady_clock::duration MGLDurationInSeconds(float duration)
 
     if (rotate.state == UIGestureRecognizerStateBegan)
     {
-        [self trackGestureEvent:MGLEventGestureRotateStart forRecognizer:rotate];
+        [self trackGestureEvent:MGLEventGestureRotateStart];
 
         self.angle = MGLRadiansFromDegrees(_mbglMap->getBearing()) * -1;
 
@@ -1248,7 +1248,7 @@ std::chrono::steady_clock::duration MGLDurationInSeconds(float duration)
     {
         return;
     }
-    [self trackGestureEvent:MGLEventGestureSingleTap forRecognizer:singleTap];
+    [self trackGestureEvent:MGLEventGestureSingleTap];
 
     CGPoint tapPoint = [singleTap locationInView:self];
 
@@ -1293,7 +1293,7 @@ std::chrono::steady_clock::duration MGLDurationInSeconds(float duration)
 
     if (doubleTap.state == UIGestureRecognizerStateBegan)
     {
-        [self trackGestureEvent:MGLEventGestureDoubleTap forRecognizer:doubleTap];
+        [self trackGestureEvent:MGLEventGestureDoubleTap];
     }
     else if (doubleTap.state == UIGestureRecognizerStateEnded)
     {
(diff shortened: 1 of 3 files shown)


friedbunny:
This should also bump one or both of the [version numbers](https://github.com/mapbox/mapbox-gl-native/blob/20c069064cb36ee086ab72955d96fc738e164a2b/platform/ios/src/MGLMapboxEvents.m#L15-L16). @camilleanne how about setting the user agent to `MapboxEventsiOS/2.0` and `version` to `2`?

camilleanne:
@friedbunny could you post an example of what each of the new events will look like? I'm having a hard time reading through all this quickly. Also, yes, the version numbers should bump. When will they go live? I would need to set up validation in api-events for a v2

friedbunny:
@camilleanne Sure, I'll get you JSON. This will go live in a couple weeks when v3.1.0 is released.

platform/ios/src/MGLMapView.mm
```diff
@@ -1474,15 +1466,11 @@ - (BOOL)gestureRecognizer:(UIGestureRecognizer *)gestureRecognizer shouldRecogni
     return ([validSimultaneousGestures containsObject:gestureRecognizer] && [validSimultaneousGestures containsObject:otherGestureRecognizer]);
 }
 
-- (void)trackGestureEvent:(NSString *)gestureID forRecognizer:(UIGestureRecognizer *)recognizer
+- (void)trackGestureEvent:(NSString *)gestureID forRecognizer:(__unused UIGestureRecognizer *)recognizer
```
1ec5: Since this parameter is unused, just remove it.

platform/ios/src/MGLMapboxEvents.m
```diff
@@ -12,8 +12,8 @@
 
 #include <sys/sysctl.h>
 
-static const NSUInteger version = 1;
```
1ec5: What’s the difference between `version` and the version substring in `MGLMapboxEventsUserAgent`? Why was it possible for them to differ until now?

platform/ios/src/MGLMapboxEvents.m
```diff
@@ -476,47 +466,63 @@ - (void) pushEvent:(NSString *)event withAttributes:(MGLMapboxEventAttributes *)
...
+        [evt setValue:[strongSelf applicationState] forKey:@"applicationState"];
 
-        MGLReachability *reachability = [MGLReachability reachabilityForLocalWiFi];
-        [evt setValue:([reachability isReachableViaWiFi] ? @YES : @NO) forKey:@"wifi"];
+        // Send with every event, but exclude these keys from location & visit events.
+        if ( ! [event isEqualToString:MGLEventTypeLocation] && ! [event isEqualToString:MGLEventTypeVisit]) {
+            [evt setValue:[strongSelf deviceOrientation] forKey:@"orientation"];
 
-        [evt setValue:[strongSelf applicationState] forKey:@"applicationState"];
+            [evt setValue:@((int)(100 * [UIDevice currentDevice].batteryLevel)) forKey:@"batteryLevel"];
```
1ec5: Round this value before casting it to an `int`.

platform/ios/src/MGLMapboxEvents.m
friedbunny: Good question. I thought `version` was used in constructing the API endpoint, [but it is not](https://github.com/mapbox/mapbox-gl-native/blob/e33e58f390b6310d48775554211a64f63b912c54/platform/ios/src/MGLMapboxEvents.m#L587). As it stands, `version` is sent with every event (and that's the only place it is used).

We bumped the user agent version to 1.1 when the last round of changes happened, at @camilleanne's request.

friedbunny: @1ec5 And [you've pondered on this before](https://github.com/mapbox/mapbox-gl-native/pull/1642#discussion_r31078613), I see.

platform/ios/src/MGLMapboxEvents.m
friedbunny: e33e58f390b6310d48775554211a64f63b912c54

platform/ios/src/MGLMapView.mm
friedbunny: e87edebe0b19a9031f8ed10e2a484961fce4a9b0

platform/ios/src/MGLMapboxEvents.m
camilleanne: Instrumentile doesn't have control over userAgent, so it sends `version` -- we might be able to drop it for mobile events by checking the userAgent, but that's why it's around.

boundsj:
Updates: 

- We can reduce lat/lng precision to 7 digits.
-  it turns out that the version = 1 in mapbox events is no longer required because we no longer need to send that along in each event. I'm going to ignore it for now and come back to it when I work on telem diet / V2 events.
- Based on https://github.com/mapbox/telemetry/pull/5#issuecomment-187878392 we should not `Remove lat,lng from interaction events` as noted above.

In general, because so much has moved on in mapview and events, this PR should probably serve as a reference and we can cherry pick in the still relevant parts to a fresh branch.

boundsj:
Subsumed by https://github.com/mapbox/mapbox-gl-native/pull/4164

PR #3616, opened 2016-01-20
switch to std::experimental::optional

jfirebaugh:
`std::experimental::optional` is likely to become a non-experimental part of the standard library in the next revision, and is more complete than our `mapbox/variant`-based implementation. It's supported by gcc 4.9 and clang 3.5; need to check on iOS/OS X.

For brevity, I recommend we define `mbgl/util/optional.hpp`:

```
#include <experimental/optional>

namespace mbgl {
using optional = std::experimental::optional;
}
```

And then use unqualified `optional<T>` wherever needed in the `mbgl` namespace.


kkaefer:
:+1:

PR #3611, opened 2016-01-20
Type of Response::{modified,expired} should be optional<SystemTimePoint>

jfirebaugh:


jfirebaugh:
`nativeOnResponse` / `HTTPAndroidRequest::onResponse` do not properly distinguish between no modified/expires/etag and zero modified/zero expires/empty etag. I'll fix that next.

kkaefer:
Looks like libstdc++ doesn't have full `optional` support?

jfirebaugh:
The only incompatibility I've found so far is that whatever implementation Android uses appears to be missing `optional::operator!=` (workaround is to use `!(a == b)` instead).

The build errors here are legit and I'm working on them.

jfirebaugh:
:eyes: @kkaefer @brunoabinader @tobrun

kkaefer:
:+1:

Issue #3607, opened 2016-01-19
[core] Crash on Transform::setScale()


tmpsantos:
Also observed when calling `Map::setZoom()` using padding repeatedly which internally calls `Transform::setScale()`.

https://travis-ci.org/mapbox/mapbox-gl-native/jobs/103073819#L1353

```
Program received signal SIGSEGV, Segmentation fault.
std::_Function_handler<mbgl::Update (std::chrono::time_point<std::chrono::_V2::steady_clock, std::chrono::duration<long, std::ratio<1l, 1000000000l> > >), mbgl::Transform::startTransition(mbgl::CameraOptions const&, mbgl::AnimationOptions const&, std::function<mbgl::Update (double)>, std::chrono::duration<long, std::ratio<1l, 1000000000l> > const&)::{lambda(std::chrono::time_point<std::chrono::_V2::steady_clock, std::chrono::duration<long, std::ratio<1l, 1000000000l> > >)#1}>::_M_invoke(std::_Any_data const&, std::chrono::time_point<std::chrono::_V2::steady_clock, std::chrono::duration<long, std::ratio<1l, 1000000000l> > >) () at /usr/include/c++/4.9/functional:2275
2275		if (_M_manager)
Thread 1 (Thread 0x2aaaaaaf2580 (LWP 12797)):
#0  std::_Function_handler<mbgl::Update (std::chrono::time_point<std::chrono::_V2::steady_clock, std::chrono::duration<long, std::ratio<1l, 1000000000l> > >), mbgl::Transform::startTransition(mbgl::CameraOptions const&, mbgl::AnimationOptions const&, std::function<mbgl::Update (double)>, std::chrono::duration<long, std::ratio<1l, 1000000000l> > const&)::{lambda(std::chrono::time_point<std::chrono::_V2::steady_clock, std::chrono::duration<long, std::ratio<1l, 1000000000l> > >)#1}>::_M_invoke(std::_Any_data const&, std::chrono::time_point<std::chrono::_V2::steady_clock, std::chrono::duration<long, std::ratio<1l, 1000000000l> > >) () at /usr/include/c++/4.9/functional:2275
#1  0x00000000005df9d4 in mbgl::Transform::startTransition(mbgl::CameraOptions const&, mbgl::AnimationOptions const&, std::function<mbgl::Update (double)>, std::chrono::duration<long, std::ratio<1l, 1000000000l> > const&) () at /usr/include/c++/4.9/functional:2439
[...]

PR #3624, opened 2016-01-20
Simplify non-animated transitions

1ec5:
This PR short-circuits the complex machinery around animated transitions for transitions that occur instantaneously and synchronously. This is a speculative fix for #3607.

/cc @tmpsantos


1ec5:
This change also brings back an old behavior in which a non-animated change to angle or pitch doesn’t cancel a currently animating transition, although I can revert that part if it turns out to be problematic.

1ec5:
Stale.

PR #3632, opened 2016-01-20
Switch to libjpeg-turbo

kkaefer:
It's way faster.


friedbunny:
Cool! Do we need to update our license to comply with [their's](https://github.com/libjpeg-turbo/libjpeg-turbo/blob/master/LICENSE.md)?

PR #3641, opened 2016-01-21
Pause map's location manager on sleep

friedbunny:
- Factors out the location manager start/stop code into new methods.
- Calls these new methods on `wakeGL` and `sleepGL`, if necessary.
- Only asks for location permissions if not already granted/denied.

Stuck this on v3.1.0, but we can move it off if this proves bothersome or controversial.

/cc @1ec5 @boundsj

### DIFF
--- a/platform/ios/src/MGLMapView.mm
+++ b/platform/ios/src/MGLMapView.mm
@@ -179,6 +179,7 @@ public:
 @property (nonatomic) CGFloat quickZoomStart;
 @property (nonatomic, getter=isDormant) BOOL dormant;
 @property (nonatomic, readonly, getter=isRotationAllowed) BOOL rotationAllowed;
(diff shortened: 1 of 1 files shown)


1ec5:
This would be a lot simpler if the logic for enabling and disabling the location manager were consolidated in one place, since multiple factors can determine whether the location manager is paused. Take a look at [`-validateDisplayLink`](https://github.com/mapbox/mapbox-gl-native/blob/874ff24d8cda6be1a3463b9614862ec186db6f40/platform/ios/src/MGLMapView.mm#L905) as an example of this validation pattern.

friedbunny:
@1ec5 This has been refactored to a single validation-style method in f6f91e07f074e42757844d03ba21b8af275ee69c. Can't think of any edge cases, but maybe you can?

friedbunny:
→ #4034 for inclusion in v3.1.2.

PR #3642, opened 2016-01-21
[node] Build OS X on Bitrise

jfirebaugh:
The OS X node build is the last non-linux build on Travis. Moving it to Bitrise will reduce Travis queue pressure, and potentially allow us to move the linux builds to CircleCI. (The reason that's currently blocked is that CircleCI doesn't support build matrixes and the way to simulate them using parallelism does not support mixed OSs.)

@mikemorris @bsudekum @miccolis This would have the effect of building for node v4 only on OS X. How do you feel about that?

### DIFF
--- a/platform/node/bitrise.yml
+++ b/platform/node/bitrise.yml
@@ -1,0 +1,64 @@
+format_version: 1.1.0
+default_step_lib_source: https://github.com/bitrise-io/bitrise-steplib.git
+
+trigger_map:
+- pattern: "*"
+  is_pull_request_allowed: true
+  workflow: primary
+
+workflows:
+  primary:
+    steps:
+    - script:
+        title: Check for skipping CI
+        inputs:
+        - content: |-
+            #!/bin/bash
+            if [[ -n "$(echo $GIT_CLONE_COMMIT_MESSAGE_SUBJECT | sed -n '/\[skip ci\]/p')"  ||
+                  -n "$(echo $GIT_CLONE_COMMIT_MESSAGE_SUBJECT | sed -n '/\[ci skip\]/p')"  ||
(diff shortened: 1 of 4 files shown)


jfirebaugh:
Note that the GitHub ⇢ Bitrise webhook for triggering this build is currently disabled, because every branch but this one is going to fail until it's merged. I'm currently manually triggering builds and will enable the webhook after merging.

Builds are here: https://www.bitrise.io/app/55e3a9bf71202106

mikemorris:
> This would have the effect of building for node v4 only on OS X.

This isn't really an option, as we need Node v4 builds on Linux to publish binaries.

jfirebaugh:
@mikemorris What I mean is having the following matrix:

* Linux, node 4
* Linux, node 5
* OS X, node 4

mikemorris:
@jfirebaugh Ah okay - is there a reason we wouldn't be able to continue building OS X Node v5 binaries on bitrise?

jfirebaugh:
Bitrise doesn't support build matrixes. If we absolutely need v5 binaries I can investigate running v4 and v5 builds in serial. It means that the build will take approximately twice as long though. I don't think we want to create separate v4 and v5 apps on Bitrise; the maintenance overhead in terms of configuration and cancelling unwanted builds is already painful enough with three apps.

1ec5:
Remember to update the build buttons in README.md and platform/node/README.md.

jfirebaugh:
Picking this up again. Wait times for OS X builds on Travis are now on the order of hours at peak times.

jfirebaugh:
Merged in 231e0b5de1b5074a2c954bd6e203cc5f0e6a2f46. Timely, as for some reason the Travis OS X Node build started failing.

PR #3650, opened 2016-01-21
node-v3.0.0

mikemorris:
Document dropping support for Node.js v0.10.x, add `engines` field to `package.json` and update changelog.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "mapbox-gl-native",
-  "version": "2.2.2",
+  "version": "3.0.0",
   "description": "Renders map tiles with Mapbox GL",
   "keywords": [
     "mapbox",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -31,6 +31,9 @@
     "request": "^2.67.0",
     "tape": "^4.2.2"
   },
+  "engines": {
+    "node": ">=4.2.1"
```
bsudekum: @mikemorris Do versions before 4.2.1 _actually_ not work?

mikemorris: @bsudekum v4.2.0 introduced a bug that was fixed in https://github.com/nodejs/node/commit/b3cbd133400a73a54ca8eab5165efe866b26b4c5, and IIRC there was another bug in early v4.0.x releases. I could track down exactly which versions are incompatible, but this was simpler.

bsudekum: :+1: just curious

PR #3654, opened 2016-01-21
[core] Simplify chrono.hpp/cpp

jfirebaugh:
Reduce need for `SystemClock::{to,from}_time_t` and other forms of casting.

:eyes: @brunoabinader -- I kept around `Seconds` and `Milliseconds` for now, but I think my preference would be to switch to `std::literals::chrono_literals` syntax, e.g. `Milliseconds(300)` ⇢ `300ms`.


brunoabinader:
Looks good :+1:, I'll update #3576 accordingly.

PR #3657, opened 2016-01-21
[core] trigger repaint after placement is redone

### LINKED ISSUE #3140: Collision debug does not properly repaint upon toggling
Collision debug toggling should trigger a repaint once the symbol buckets have been updated. This doesn't seem to be the case as the collision boxes are only visible e.g. when enabling collision debug after a second call to `MapContext::update()`.

Current behavior _feels_ correct because usually we are causing more repaints due to e.g. mouse gestures. In GLFW example, if we keep pressing `C` key the behavior also _feels_ correct because each time we press the shortcut, the previous repaint gives the wrong impression that collision debug state is the same as visible on-screen.

/cc @mapbox/gl

ansis:
This adds `onPlacementRedone` to `Source::Observer` which is called when a tile is updated with new symbol placement. Style uses this to trigger a repaint.

fixes #3140

This also fixes a bug where placement could be redone with a stale PlacementConfig after the tile first finishes loading.

@jfirebaugh can you review?

### DIFF
--- a/package.json
+++ b/package.json
@@ -26,7 +26,7 @@
   ],
   "devDependencies": {
     "aws-sdk": "^2.2.21",
-    "mapbox-gl-test-suite": "mapbox/mapbox-gl-test-suite#2f4d8ed044a3c962a43d62de0608b752eb68052c",
+    "mapbox-gl-test-suite": "mapbox/mapbox-gl-test-suite#1b9035a02f8f23bf72a197c2a8d8f910d935346f",
     "node-gyp": "^3.2.1",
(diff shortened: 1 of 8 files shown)


jfirebaugh:
Code changes look good.

One render test failed: symbol-avoid-edges rotated-false.

image

It looks like other `symbol-avoid-edges` tests have similar diffs but are passing. Thoughts?

jfirebaugh:
http://mapbox.s3.amazonaws.com/mapbox-gl-native/render-tests/9415.3/index.html BTW

PR #3658, opened 2016-01-22
[core] Simplify revalidation (part 2)

jfirebaugh:
This implements changes to consumers of `FileSource::request` as described in https://github.com/mapbox/mapbox-gl-native/commit/7c6ae71791bf5e7171d1ecda1670eb0cae068150#commitcomment-15482416, namely always using `response.notModified` to check for unchanged responses, rather than retaining and comparing the prior response body.

Then it adds members to `Resource` such that it's possible to implement a self-contained revalidating `FileSource` (i.e. you don't have to pass in the whole prior response, which isn't possible in the `FileSource` API). This is necessary for further progress on offline APIs.

:eyes: @kkaefer

### DIFF
--- a/include/mbgl/storage/resource.hpp
+++ b/include/mbgl/storage/resource.hpp
@@ -1,12 +1,15 @@
 #ifndef MBGL_STORAGE_RESOURCE
 #define MBGL_STORAGE_RESOURCE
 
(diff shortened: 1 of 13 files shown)


jfirebaugh:
Merged.

kkaefer:
Looks good!

PR #3659, opened 2016-01-22
Update telemetry location manager to allow background location updates when required.

boundsj:
In iOS 9, it is now required to set the value of `allowsBackgroundLocationUpdates` on CLLocationManager to YES (in addition to previous requirements) to achieve location updates in the background. 

The Mapbox gl native telemetry location manager runs in the background to gather data for telemetry events for improving open map data. This PR updates that location manager to be compatible with iOS 9 for the applicable use cases. Bugs related to gaps in telemetry event data from devices running iOS 9 can probably be closed after we test this change.

cc @1ec5 @friedbunny

### DIFF
--- a/ios/app/app-info.plist
+++ b/ios/app/app-info.plist
@@ -43,6 +43,10 @@
 	<string>The map will ALWAYS display the user's location.</string>
 	<key>NSLocationWhenInUseUsageDescription</key>
 	<string>The map will display the user's location.</string>
+	<key>UIBackgroundModes</key>
+	<array>
+		<string>location</string>
+	</array>
 	<key>UILaunchStoryboardName</key>
 	<string>Default</string>
 	<key>UIRequiredDeviceCapabilities</key>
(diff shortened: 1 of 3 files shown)


boundsj:
Tagging this to future 3.2.0 so we can work on other enhancements first.

mb12:
@boundsj Can you please clarify the following?

1.) What is the behavior if the application using MGLMapView does not included Location in UIBackgroundModes in the plist file?

2.) Is it possible to make sure that requestAlwaysAuthorization is also called before the call to startLocationUpdates?

friedbunny:
@mb12:

1. Without `UIBackgroundModes Location`, this has no effect.
2. At this point, we're not going to require a minimum level of location permissions.

boundsj:
Note: this will need a correction to check if background mode.location has been enabled by the developer. It is a fatal error to set `allowsBackgroundLocationUpdates` to true if background mode.location is false.

boundsj:
Addressed by https://github.com/mapbox/mapbox-gl-native/pull/4115

PR #3663, opened 2016-01-22
Override tileSize, tile API urls for mapbox raster sources

yhahn:
Native port of PR: mapbox/mapbox-gl-js#1964. Would love your eyes, especially the best placement for `tileSize` override of Mapbox raster sources.

cc @kkaefer @jfirebaugh

### DIFF
--- a/src/mbgl/map/source.cpp
+++ b/src/mbgl/map/source.cpp
@@ -406,6 +406,14 @@ bool Source::update(const StyleUpdateParameters& parameters) {
         return allTilesUpdated;
     }
 
+    // Mapbox+raster specific override: the requested image tiles are always
+    // of size 512x512 on the v4 tile API (see mapbox::normalizeRasterTileURL)
+    // so disregard the source tileSize and instead use a tileSize based on
+    // the device pixelRatio.
+    if (type == SourceType::Raster && util::mapbox::isMapboxURL(url)) {
+        tileSize = parameters.pixelRatio > 1.0 ? 256 : 512;
+    }
+
     double zoom = coveringZoomLevel(parameters.transformState);
     std::forward_list<TileID> required = coveringTiles(parameters.transformState);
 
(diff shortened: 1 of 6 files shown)


src/mbgl/util/mapbox.cpp
```diff
@@ -135,11 +135,12 @@ std::string normalizeRasterTileURL(const std::string& url) {
         normalizedURL.replace(extensionIdx + 1, 3, "webp");
     }
 #endif // !defined(__ANDROID__) && !defined(__APPLE__)
-    normalizedURL.insert(extensionIdx, "{ratio}");
+    // Mapbox raster sources always use the @2x suffix on the v4 tile API
```
kkaefer: There's no check that this is indeed a v4 tile API.

src/mbgl/map/source.cpp
```diff
@@ -406,6 +406,14 @@ bool Source::update(const StyleUpdateParameters& parameters) {
...
     }
 
+    // Mapbox+raster specific override: the requested image tiles are always
+    // of size 512x512 on the v4 tile API (see mapbox::normalizeRasterTileURL)
+    // so disregard the source tileSize and instead use a tileSize based on
+    // the device pixelRatio.
+    if (type == SourceType::Raster && util::mapbox::isMapboxURL(url)) {
+        tileSize = parameters.pixelRatio > 1.0 ? 256 : 512;
+    }
+
```
kkaefer: We should move this check to a place where it is executed only once: when parsing the style in `style_parser.cpp`.

jfirebaugh:
At this point I feel like we should completely ignore the `tiles` property in Mapbox TileJSON and generate a single tile URL from scratch.

jfirebaugh:
Closing here because we realized that we can't safely use `@2x` tiles from one zoom level lower on `@1x` devices. Raster zoom levels just don't work like that in many cases.

Reverted in gl-js with https://github.com/mapbox/mapbox-gl-js/commit/971b956670d0e89039c22618410adb0d33cd1156.

PR #3669, opened 2016-01-22
[core] use nearest neigbour texture filtering for some line icons

ansis:
Use nearest neighbour texture filtering to draw sharper icons when the icons are aligned with the viewport. This matches -js.


ansis:
ded8b62 is the only commit merged with this pr. 0101fc0 is already in master (but wasn't when the pr was opened)

PR #3671, opened 2016-01-23
[ios] Avoid the blue background location status bar

### LINKED ISSUE #2945: iOS GL - Using Your Location blue status bar on iOS 8
After upgrading to GL, if the user sets location preferences to `While Using the App` on iOS 8, they will get the blue status bar that reads `xxx is Using Your Location` when they leave our app.  This does not happen on iOS 9.  This was also not happening on 8 with the non-GL version of MBXMapKit.

Is this expected behavior or can we get this banner to go away?  We've seen it on v2.1.2 and v3.0.0-pre4.

friedbunny:
Apps with `whenInUse` location permission will show a blue status bar when they continue to use location services after leaving the foreground. This is worrying and to be avoided, so let's disable telemetry location services in this situation.

Fixes #2945, refs #3641.

/cc @boundsj @1ec5 @jfirebaugh @mikemorris


friedbunny:
I looked into listening for `UIApplication` background/foreground notifications, but we already trigger `validate` and `validateUpdatingLocation` in the necessary cases.

mikemorris:
This would disable telemetry collection even when the app itself is triggering the blue bar, right?

friedbunny:
Yup, we have no way of knowing if another location manager elsewhere in the app is still active.

1ec5:
:+1:

friedbunny:
Rebased with an eye to merging. We'll revisit ways to better piggyback in the future, but for now let's err on the side of not annoying users.

PR #3673, opened 2016-01-23
[ios] Update IB designables message to remove Settings.bundle mention

friedbunny:
Adding Settings.bundle is no longer a requirement to get started.

/cc @1ec5


friedbunny:
@1ec5 I should note, I never got designables to actually work — `build failed` in iosapp and using the dynamic framework in a throwaway app.

1ec5:
:+1: Though this begs the question of whether we should even supply a designable given that IB doesn't support it in opaque frameworks where the source isn't part of the workspace. I guess we can leave it in for now, since there's no performance cost.

PR #3674, opened 2016-01-23
Custom edgeInsets for fitting in showAnnotations

jormon:
Keeps sensible defaults already in place, but allows callers to override.

### DIFF
--- a/include/mbgl/ios/MGLMapView.h
+++ b/include/mbgl/ios/MGLMapView.h
@@ -498,15 +498,33 @@ IB_DESIGNABLE
 
 /**
  Sets the visible region so that the map displays the specified annotations.
- 
+
  Calling this method updates the value in the visibleCoordinateBounds property
- and potentially other properties to reflect the new map region.
- 
+ and potentially other properties to reflect the new map region. Defaults to 
+ insetting the annotation bounds by 100,100,100,100, or if map is smaller than 
+ 200x200 in either dimension, sets the insets to 20% of that dimmension on
+ either side.
+
  @param annotations The annotations that you want to be visible in the map.
  @param animated `YES` if you want the map region change to be animated, or `NO`
     if you want the map to display the new region immediately without animations.
  */
 - (void)showAnnotations:(NS_ARRAY_OF(id <MGLAnnotation>) *)annotations animated:(BOOL)animated;
+
+/**
+ Sets the visible region so that the map displays the specified annotations with
+ custom insets.
+
+ Calling this method updates the value in the visibleCoordinateBounds property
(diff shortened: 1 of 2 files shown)


jormon:
I have not been able to set up your test app yet -- but I can confirm that  `make itest` passes.

1ec5:
Thanks! make itest might not test this particular API yet, but this looks good.

Should the contentInset be added to the passed-in insets?

include/mbgl/ios/MGLMapView.h
```diff
@@ -498,10 +498,10 @@ IB_DESIGNABLE
 
 /**
  Sets the visible region so that the map displays the specified annotations.
- 
+
  Calling this method updates the value in the visibleCoordinateBounds property
- and potentially other properties to reflect the new map region.
- 
+ and potentially other properties to reflect the new map region. Defaults to insetting the annotation bounds by 100,100,100,100, or if map is smaller than 200x200 in either dimmension, sets the insets to 20% of that dimmension on either side.
```
friedbunny: "Dimension" is spelled incorrectly. We'll also want to wrap this at 80 characters.

friedbunny:
Many thanks for the pull request, @jormon. Other than some slight changes to the docs, this looks great.

1ec5:
> Should the contentInset be added to the passed-in insets?

Never mind, `-setVisibleCoordinateBounds:` already adds `contentInset`.

jormon:
cleaned up warp and typos -- thanks for considering!

jormon:
I don't see much documentation or information re: your release schedule and testing procedures.  Any chance this can make the 3.1.0 boat?

jormon:
@friedbunny -- what are the next steps here?  can i get a "ready for review" tag?

the https://github.com/mapbox/mapbox-gl-native/wiki/Contributing guidelines could also do a better job of explaining the flow of pull requests, as well as the timing around planned versions and upcoming releases.

cc @geografa @ansis the closest people I could find to developer advocates on your team website.

(for context, I care and am taking the time to pester because we use mapbox-gl-ios and love it and need these improvements to make using it feasible in our production apps!)

1ec5:
@jormon, first of all, thanks for submitting this PR. @friedbunny and I have reviewed it and it looks good to go. :+1: We’re trying to lock down the 3.1.0 milestone and have only held off branching so that some performance fixes in core code could land first. However, this PR does look harmless and useful, so I’ll do a little copyediting and sneak it into 3.1.0.

jormon:
Thanks all -- excited for this, and all the other good stuff (map move completion handlers in particular) coming in 3.1.0

1ec5:
Continued in #3787.

PR #3675, opened 2016-01-23
[core] Final preparations for OfflineFileSource

jfirebaugh:
This results in `OnlineFileSource` containing precisely the logic we want for reuse by `OfflineFileSource` (which needs access to a `FileSource` from which to download), and no more.

:eyes: @kkaefer

### DIFF
--- a/include/mbgl/storage/default_file_source.hpp
+++ b/include/mbgl/storage/default_file_source.hpp
@@ -5,6 +5,8 @@
 
 namespace mbgl {
 
+class SQLiteCache;
+
 class DefaultFileSource : public FileSource {
 public:
     DefaultFileSource(const std::string& cachePath, const std::string& assetRoot);
@@ -18,8 +20,13 @@ public:
 
     std::unique_ptr<FileRequest> request(const Resource&, Callback) override;
 
+    // For testing purposes only.
+    SQLiteCache& getCache();
+
 private:
     class Impl;
+    friend class DefaultFileRequest;
+
     const std::unique_ptr<Impl> impl;
 };
 
--- a/include/mbgl/storage/online_file_source.hpp
+++ b/include/mbgl/storage/online_file_source.hpp
@@ -5,15 +5,13 @@
 
 namespace mbgl {
 
-class SQLiteCache;
-
 namespace util {
 template <typename T> class Thread;
 } // namespace util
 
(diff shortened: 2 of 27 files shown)


include/mbgl/storage/default_file_source.hpp
```diff
@@ -18,8 +20,13 @@ class DefaultFileSource : public FileSource {
 
     std::unique_ptr<FileRequest> request(const Resource&, Callback) override;
 
+    // For testing purposes only.
```
kkaefer: Can we make this private: https://github.com/google/googletest/blob/master/googletest/docs/AdvancedGuide.md#private-class-members

jfirebaugh: I don't think we want to add a `gtest.h` dependency to non-test headers. I tried various permutations of `friend class Storage;` and forward declarations but couldn't it to work.

jfirebaugh:
I believe 31dd6e7f69320ca437fec26d4dac5aea90f1e296 makes `util::mapbox::canonicalURL` unnecessary, since it effectively moves the normalization of `mapbox://` URLs _after_ the cache request. It will also change the effective URL key for many cached resources, since `util::mapbox::canonicalURL(util::mapbox::normalize*URL(url)) != url`. Should we increase the cache schema version to ensure consistency? This will blow away any existing caches, which we're likely to do for offline anyway.

jfirebaugh:
> 31dd6e7f69320ca437fec26d4dac5aea90f1e296 makes `util::mapbox::canonicalURL` unnecessary

This isn't quite true. Tile URLs still contain embedded access tokens. This will cease to be an issue once we cache tiles by TileID rather than URL.

kkaefer:
> Tile URLs still contain embedded access tokens.

...can we come up with a `mapbox://` tile URL and change the TileJSON our servers to output that URL?

include/mbgl/storage/default_file_source.hpp
kkaefer: We could either to forward declarations in the global namespace, and use `friend class ::Storage_Foo_Test` or something like this:

```cpp
#ifdef GTEST_TEST
public:
#else
private:
#endif
```

jfirebaugh:
Going to cut a fresh offline PR.

PR #3676, opened 2016-01-23
[core] Proposed offline interface

jfirebaugh:
Here's my proposal for the mbgl core offline interface. This is stable enough that I think we can start working on iOS and Android bindings for this interface, independent of the core implementation work.

Implementation nodes in code comments.

cc @1ec5 @zugaldia @adam-mapbox

### DIFF
--- a/include/mbgl/storage/default_file_source.hpp
+++ b/include/mbgl/storage/default_file_source.hpp
@@ -2,8 +2,134 @@
 #define MBGL_STORAGE_DEFAULT_FILE_SOURCE
 
 #include <mbgl/storage/file_source.hpp>
+#include <mbgl/util/geo.hpp>
+
+#include <vector>
 
 namespace mbgl {
+
+/*
+ * An offline region defined by a style URL, geographic bounding box, zoom range, and
+ * device pixel ratio.
+ */
+class OfflineTilePyramidRegionDefinition {
+public:
+    OfflineTilePyramidRegionDefinition(const std::string&, const LatLngBounds&, double, double, float);
+
+    std::string styleURL;
+    LatLngBounds bounds;
+    double minZoom;
+    double maxZoom;
+    float pixelRatio;
+};
+
+/*
+ * For the present, a tile pyramid is the only type of offline region. In the future,
+ * other definition types will be available and this will be a variant type.
+ */
+using OfflineRegionDefinition = OfflineTilePyramidRegionDefinition;
+
+/*
+ * Arbitrary binary region metadata. The contents are opaque to the mbgl implementation;
+ * it just stores and retrieves a BLOB. SDK bindings may enforce a higher-level data format
+ * if doing so is natural for the platform.
+ */
+using OfflineRegionMetadata = std::vector<uint8_t>;
+
+/*
+ * A region is either inactive (not downloading, but previously-downloaded
+ * resources are available for use), or active (resources are being downloaded
+ * or will be downloaded, if necessary, when network access is available).
+ * 
+ * This state is independent of whether or not the complete set of resources
+ * is currently available for offline use. To check if that is the case, use
+ * `OfflineRegionStatus::complete()`.
+ */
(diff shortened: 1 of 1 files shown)


include/mbgl/storage/default_file_source.hpp
```diff
@@ -2,9 +2,135 @@
...
 
+/*
+ * An offline region defined by a style URL, geographic bounding box, zoom range, and
+ * device pixel ratio.
+ */
+class OfflineTilePyramidRegionDefinition {
+public:
+    OfflineTilePyramidRegionDefinition(const std::string&, const LatLngBounds&, double, double, float);
+
+    std::string styleURL;
```
mb12: Why is styleUrl and pixelRatio part of this? Isn't the following use case more likely with vector data set?

- createOfflineRegion
- Now apply different styles (with possibly different sprites) to the same offline region.

include/mbgl/storage/default_file_source.hpp
```diff
@@ -18,6 +144,49 @@ class DefaultFileSource : public FileSource {
...
+     *
+     * When the initial database queries have completed, the provided callback will be
+     * executed on the database thread; it is the responsibility of the SDK bindings
+     * to re-execute a user-provided callback on the main thread.
+     *
+     * Note that the resulting region will be in an inactive download state; to begin
+     * downloading resources, call `setDownloadState(OfflineRegionDownloadState::Active)`,
+     * optionally registering an `OfflineRegionObserver` beforehand.
+     */
+    void createOfflineRegion(const OfflineRegionDefinition& definition,
```
mb12: How is update modeled here? Does createOfflineRegion  internally do update if called repeatedly for the same OfflineRegionDefinition?

include/mbgl/storage/default_file_source.hpp
1ec5: If I’m not mistaken, the style URL determines which data sources get downloaded. For example, taking the Satellite style offline shouldn’t pull in Mapbox Streets data.

jfirebaugh: An offline region must store _all_ resources necessary for that region. Those resources are necessarily tied to a style: two different styles may use different sources, sprites, and glyphs. Therefore the style is part of the region definition.

Resources that are common between two regions will be deduplicated -- downloaded and stored only once. If you want multiple offline styles, create a region for each one. The result is that union of all required resources will be available offline.

include/mbgl/storage/default_file_source.hpp
jfirebaugh: Updates will not be part of this API, at least in the first version. With deduplication, you can get much the same behavior by creating a new region and then deleting the old one.

include/mbgl/storage/default_file_source.hpp
```diff
@@ -2,9 +2,135 @@
...
 
 namespace mbgl {
 
+/*
+ * An offline region defined by a style URL, geographic bounding box, zoom range, and
+ * device pixel ratio.
+ */
+class OfflineTilePyramidRegionDefinition {
+public:
+    OfflineTilePyramidRegionDefinition(const std::string&, const LatLngBounds&, double, double, float);
```
mb12: It would be good to have a constructor where minZoom and maxZoom is optional since they can be automatically inferred for all sources from the style.

jfirebaugh: The intent is that they can be used to specify a subrange of the full zoom range. If you want the full range supported by sources, pass `minZoom = 0` and `maxZoom = Infinity`. I'm fine if the SDK bindings use those as default values.

1ec5: There’s precedent for this in how `Transform` clamps `CameraOptions::zoom` to the minimum and maximum zoom levels supported by the style. On the iOS side, we probably will have a convenience initializer without zoom arguments.

jfirebaugh:
Rolling this into other PRs.

PR #3683, opened 2016-01-25
Disable implicit animation of user dot bounds

1ec5:
This PR prevents Core Animation from implicitly animating changes to the user dot’s bounds. Therefore, the user dot no longer lags perceptibly when panning, zooming, or rotating the map.

We can similarly disable implicit animation of the accuracy circle’s size (#2045). However, that’ll have to happen in a different PR, because the accuracy circle’s animation is still useful for when the accuracy itself changes.

/ref #1125
/cc @friedbunny @kkaefer


friedbunny:
:+1: in theory, but I'm still struggling to perceive much/any practical effect.

1ec5:
For future reference, the behavior I disabled here is documented in “[CALayer Animatable Properties](https://developer.apple.com/library/ios/documentation/Cocoa/Conceptual/CoreAnimation_guide/AnimatableProperties/AnimatableProperties.html#//apple_ref/doc/uid/TP40004514-CH11-SW4)” as a quarter-second basic animation.

Issue #3627, opened 2016-01-20
Content insets


1ec5:
Content insets were implemented in mbgl and for iOS and OS X in #3583. MapView should expose some sort of inset or padding property so that the map’s center can be shifted downward to account for e.g. a translucent or floating toolbar at the top.

/cc @tobrun

bleege:
I'm going to review @lgeromegnace's PR in #3687 and then build an example using it in the TestApp so that we can verify it and demonstrate to others how it's used.

bleege:
The API build out from #3687 has now been been reviewed and merged into `master`.  Will build out example next.

bleege:
I set the location to be center on Nashville and then took screenshots of with and without a top offset of 250 pixels.  Not much has changed and this is likely due to density not being factored in to the API.

![device-2016-01-27-135809](https://cloud.githubusercontent.com/assets/989928/12626456/58db1fa2-c4fe-11e5-8c26-8ece21850706.png)
_No Content Offset_

![device-2016-01-27-135905](https://cloud.githubusercontent.com/assets/989928/12626464/63feec38-c4fe-11e5-872b-dfd9eb6fe28a.png)
_Top Content Offset Of 250px_

1ec5:
@bleege, this is because setting the insets alone does not move the map. [On iOS](https://github.com/mapbox/mapbox-gl-native/blob/master/platform/ios/src/MGLMapView.mm#L846-L869) and OS X, calling `-setContentInset:` causes the map to shift with optional animation.

bleege:
Thanks for the clarification @1ec5.  To help illustrate this better (and the fact that we still have API work todo), I added a Marker to the map.

![device-2016-01-27-145136](https://cloud.githubusercontent.com/assets/989928/12628077/0fe32be2-c507-11e5-8f82-b56857e83bc2.png)
_No Content Offset_

![device-2016-01-27-145226](https://cloud.githubusercontent.com/assets/989928/12628086/1917c984-c507-11e5-9eae-91f0addc8dc7.png)
_Top Content Offset Of 1000px_

bleege:
[...]

PR #3687, opened 2016-01-25
3627 android content insets

lgeromegnace:
This PR is meant to implement padding functionality on android platform described in this issue #3627.

* Add a `setPaddingContent` method on `MapView`
* Padding are stored in NativeMapView to be applied on any methods that can update current bounding box.


tobrun:
@lgeromegnace this is great, will review soon!

bleege:
This looks like a great start to the API.  I just rebased and merged it into `master`.  We'll still need to have this work with the Camera API and build an example in the TestApp.

Thanks for the contribution @lgeromegnace!

PR #3695, opened 2016-01-26
[android] #3692 - catch NullPointerException assign a generic message in that case

### LINKED ISSUE #3692: java.lang.NullPointerException: String cannot be null.
I got this exception

>`java.lang.NullPointerException: String cannot be null.
>	at com.mapbox.mapboxsdk.http.HTTPContext$HTTPRequest.nativeOnFailure(Native Method)
>	at com.mapbox.mapboxsdk.http.HTTPContext$HTTPRequest.onFailure(SourceFile:122)
>	at com.mapbox.mapboxsdk.http.HTTPContext$HTTPRequest.onResponse(SourceFile:99)
>	at a.aq$b.f(SourceFile:133)
>	at a.a.k.run(SourceFile:33)
>	at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1113)
>	at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:588)
>	at java.lang.Thread.run(Thread.java:818)`

I see, some one has commited this issue. but this issue is very serious to me. cause our APP is a ski tracker .I need use mapbox to display users' location and tracks.if user have unstable network this exception maybe occur.So . is there a solution can resolve this issue? thanks a lot

tobrun:
Closes #3692.

Initially I wanted to optimise `:std_string_from_jstring`, but I was not able to remove the throw of the NullPointerException since it's being used by other methods. I decided to check for a null value and replace that value with a generic message "Error processing the request". This should stop crashing the application if there is a bad internet connection. Not really sure if this is the correct way of handling this..

:eyes: @jfirebaugh

### DIFF
--- a/platform/android/src/http_request_android.cpp
+++ b/platform/android/src/http_request_android.cpp
@@ -221,7 +221,10 @@ void HTTPAndroidRequest::onResponse(JNIEnv* env, int code, jstring /* message */
 }
 
 void HTTPAndroidRequest::onFailure(JNIEnv* env, int type, jstring message) {
-    std::string messageStr = mbgl::android::std_string_from_jstring(env, message);
+    std::string messageStr = "Error processing the request";
+    if(message) {
+        messageStr = mbgl::android::std_string_from_jstring(env, message);
+    }
 
     response = std::make_unique<Response>();
     using Error = Response::Error;


jfirebaugh:
Is it expected that the message can be null? If so this looks like the right fix. Just wondering if there could be an issue upstream of this code.

tobrun:
@jfirebaugh looking now at the java code that is executed before the jni part.

```java
      @Override
        public void onFailure(Call call, IOException e) {
            Log.w(LOG_TAG, String.format("[HTTP] Request could not be executed: %s", e.getMessage()));

            int type = PERMANENT_ERROR;
            if ((e instanceof UnknownHostException) || (e instanceof SocketException) || (e instanceof ProtocolException) || (e instanceof SSLException)) {
                type = CONNECTION_ERROR;
            } else if ((e instanceof InterruptedIOException)) {
                type = TEMPORARY_ERROR;
            } else if (mCall.isCanceled()) {
                type = CANCELED_ERROR;
            }

            nativeOnFailure(mNativePtr, type, e.getMessage());
        }
```

I will be cleaner handling this the null value here.
Going to rework this.

tobrun:
Since I'm starting form scratch here,  going to create new PR for this

Issue #3625, opened 2016-01-20
Concurrent transition animations


1ec5:
`mbgl::Transform` only supports one transition animation at a time, and starting an animation ends any existing animation (see #3624). `mbgl::Transform` should be refactored to manage multiple concurrent animations, potentially with staggered start and end times. At a minimum, it should be possible for two animations that animate different properties to run without conflicting with one another.

/cc @ansis @lucaswoj

jfirebaugh:
Can you say more about why it needs to support this?

1ec5:
On iOS, in heading tracking mode, user location updates typically come in every second or so (less frequently when there’s poor GPS connectivity), while heading updates come in more or less in real time from the on-board magnetometer.

MGLMapView pans to the new location with animation and also independently rotates to the new heading with animation. However, if you’re in motion and aren’t keeping your device pointed in a constant direction, heading updates tend to interrupt animations to new locations. On master, this is perceived as the map jumping from location to location with smooth rotation. In #3589, this is perceived as a series of smooth rotations in one place, followed by rapid panning as the location updates catch up.

I considered less drastic approaches to addressing this issue, such as allowing callers to `easeTo()` to update the camera in the middle of the animation, but the resulting code was messier than what we would get by supporting multiple concurrent animations. To be clear, all the animations would continue to run on the Map thread, but we’d interleave their frame functions rather than canceling animations.

ansis:
> to update the camera in the middle of the animation

What does "update the camera" mean here? If I do map.getLatLng in the middle of an animation do I get the current location or the location of the beginning of the animation?

1ec5:
> What does "update the camera" mean here?

CameraOptions is passed in by reference, so I considered checking on each frame whether the CameraOptions fields had changed and adjust accordingly. But that would never yield a pleasing animation and introduces too much complexity into what are already complex `easeTo()` and `flyTo()` methods. I don’t see a whole lot of code in mbgl or the iOS or OS X SDKs that assumes only one animation can run at a time. The Android SDK does appear to rely on this assumption for its listeners (cc @tobrun), so there might need to be more significant changes there.

> If I do map.getLatLng in the middle of an animation do I get the current location or the location of the beginning of the animation?

You get the current location. That won’t change.

1ec5:
[...]

PR #3705, opened 2016-01-27
[ios] Improve heading tracking smoothness by filtering updates less

friedbunny:
Heading tracking was only delivering updates every 5°, which seems kind of infrequent. This changes the filter to 1°.

To-do:
- [ ] Investigate and profile performance to see if this puts additional strain on our map, especially on older devices.
- [ ] Confirm that `headingFilter` is the bottleneck I think it is, rather than the animation.

/cc @1ec5

### DIFF
--- a/platform/ios/src/MGLMapView.mm
+++ b/platform/ios/src/MGLMapView.mm
@@ -2930,7 +2930,7 @@ mbgl::Duration MGLDurationInSeconds(NSTimeInterval duration)
         }
 #endif
 
-        self.locationManager.headingFilter = 5.0;
+        self.locationManager.headingFilter = 1.0;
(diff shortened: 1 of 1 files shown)


friedbunny:
Turns out the default value for `CLLocationManager.headingFilter` _is_ `1`, so...

1ec5:
ref mapbox/mapbox-ios-sdk-legacy#84

1ec5:
Frequent heading changes already preempt location changes. This change all but ensures the user dot will never move in real world usage of heading tracking. #3625 should be implemented first, so that rotation due to heading updates can occur independently of panning due to location updates.

friedbunny:
Tabling this for now.

PR #3707, opened 2016-01-27
Fix Memory leak caused by duplicated add in LocationListener list

jaumeolba:
The addLocationListener method didn't check if the LocationListener was
already in the list.

From MapView, if we call setMyLocationEnabled(true), it adds a
LocationListener and onResume adds the same LocationListener again, but
is just removed by onPause.

So LocationListener is added twice but removed just once, so
LocationServices singleton keeps an instance of the the LocationListener
that is never released.


tobrun:
@jaumeolba 
Thank you for contributing!
I will merge this in in the upcoming days.

tobrun:
Merged with https://github.com/mapbox/mapbox-gl-native/commit/026e6a75600f7109828fc33ff97b117a3a5090a7

PR #3636, opened 2016-01-25
Hang on each tile load — VectorTileData, TileWorker::parseLayer

tmpsantos:
On iOS and to a lesser extent OS X, MGLMapView is unresponsive for a second or so every time a tile loads at mid zoom levels in the San Francisco Bay Area. This is a severe regression from probably late last week. These are the only threads that appear to be doing much of anything:

```
Map (13)#0	0x0000000183fc7f90 in __psynch_mutexwait ()
#1	0x000000018409239c in _pthread_mutex_lock_wait ()
#2	0x000000018409257c in _pthread_mutex_lock_slow ()
#3	0x0000000183a53a5c in std::__1::recursive_mutex::lock() ()
#4	0x00000001003cd3c0 in std::__1::lock_guard<std::__1::recursive_mutex>::lock_guard(std::__1::recursive_mutex&) [inlined] at /Applications/Xcode+1.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/../include/c++/v1/__mutex_base:83
#5	0x00000001003cd3ac in std::__1::lock_guard<std::__1::recursive_mutex>::lock_guard(std::__1::recursive_mutex&) [inlined] at /Applications/Xcode+1.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/../include/c++/v1/__mutex_base:83
#6	0x00000001003cd3a0 in _ZN4mbgl4util7RunLoop7InvokerIZNS0_6ThreadINS_6Worker4ImplEE4bindIMS5_FvPNS_10TileWorkerENSt3__16vectorINSA_10unique_ptrINS_10StyleLayerENSA_14default_deleteISD_EEEENSA_9allocatorISG_EEEENSC_INS_12GeometryTileENSE_ISK_EEEENS_15PlacementConfigENSA_8functionIFvN6mapbox4util7variantIJNS_22TileParseResultBucketsESt13exception_ptrEEEEEEEEEDaT_EUlDpOT_E_NSA_5tupleIJS9_SJ_SM_SN_ZNS1_18invokeWithCallbackIS14_RSW_JS9_SJ_SM_RSN_EEENSC_INS_11WorkRequestENSE_IS19_EEEEOS10_OT0_DpOT1_EUlS13_E_EEEE6cancelEv at /Users/mxn/hub/mapbox-gl-native-3/gyp/../include/mbgl/util/run_loop.hpp:142
[...]

PR #3711, opened 2016-01-27
WorkTasks: destruct data before invoking callback

kkaefer:
This is a short term fix for #3636. The reason for the hangs we were seeing was because the following happened:

- Work task completes, calls `callback`, which schedules the after callback on the main thread
- After callback was invoked and tried to cancel the work task first thing, which blocks on the mutex
- On the worker thread, we were destructing the arguments (in particular the `VectorTile` object) *after* we've called the callback in the first step; on slow devices, this could take 20-100ms
- Worker thread releases mutex, then main thread can finally acquire it to mark the task as canceled

This patch fixes this with the following changes:

- Move work task arguments from the tuple into the actual function, so they can be destructed there before the callback gets fired
- Actually destruct the arguments by either resetting, or moving the data into the invocation where it will get destructed

### DIFF
--- a/src/mbgl/map/tile_worker.cpp
+++ b/src/mbgl/map/tile_worker.cpp
@@ -37,7 +37,7 @@ TileWorker::~TileWorker() {
 }
 
 TileParseResult TileWorker::parseAllLayers(std::vector<std::unique_ptr<StyleLayer>> layers_,
-                                           const GeometryTile& geometryTile,
+                                           std::unique_ptr<const GeometryTile> geometryTile,
                                            PlacementConfig config) {
     // We're doing a fresh parse of the tile, because the underlying data has changed.
     pending.clear();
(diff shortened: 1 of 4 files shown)


src/mbgl/map/tile_worker.cpp
```diff
@@ -37,7 +37,7 @@ TileWorker::~TileWorker() {
 }
 
 TileParseResult TileWorker::parseAllLayers(std::vector<std::unique_ptr<StyleLayer>> layers_,
-                                           const GeometryTile& geometryTile,
+                                           std::unique_ptr<const GeometryTile> geometryTile,
```
kkaefer: by moving the entire unique_ptr, we ensure that it gets destructed when this function returns.

kkaefer:
With this fix, we almost never wait for WorkTask cancelations (= immediate lock acquisition), and if we do, it's only for a very short period of time.

jfirebaugh:
:+1:

Issue #509, opened 2014-10-27
max zoom > 18


incanus:
This was changed back in https://github.com/mapbox/mapbox-gl-native/issues/178 but should: 

1. Allow for zoom > 18
1. Allow for setting min/max from external API

manuelwenner:
What about max zoom Level in mapbox-gl-native for android? When i try to set zoom level >18.0 it's ignored. 

Thanks in advance for your answers.

tmpsantos:
> What about max zoom Level in mapbox-gl-native for android? When i try to set zoom level >18.0 it's ignored.
> 
> Thanks in advance for your answers

@manuelwenner It is currently hardcoded. Until we provide an API for that, you can tweak it at this variable here: https://github.com/mapbox/mapbox-gl-native/blob/master/src/mbgl/map/transform.hpp#L87

manuelwenner:
Thanks @tmpsantos that worked perfect for me.

petaren:
Is there an estimated time as to when an increase in max zoom level might be available? We would need a max zoom level of around 20 in my current project. Without it, it's unfortunately useless for us :(

tmpsantos:
> Is there an estimated time as to when an increase in max zoom level might be available? We would need a max zoom level of around 20 in my current project. Without it, it's unfortunately useless for us :(

Not on the roadmap AFAIK, but you can change that on the code for now.

petaren:
Okey, L87 in transform.hpp is blank. So what do I need to change to get it to work?

mourner:
It's hardcoded here now: https://github.com/mapbox/mapbox-gl-native/blob/e63fb444993c16b158cf833ba4c4e66f81943fba/src/mbgl/map/transform_state.hpp#L76-L77

I think we should address this issue sooner, shouldn't be hard and is an often requested feature.

franckt:
Hi, 
Any news about providing an api to set max zoom ?

ansis:
@franckt max zoom is implemented in master and should be available in the next release

bleege:
Re-opening as discussion with @kkaefer this morning revealed that the max zoom level in `mgbl` is actually 25.5.  We'll need to update the iOS and Android bindings added in PR #3712 to support this.

/cc @mapbox/mobile

bleege:
[...]

Issue #1690, opened 2015-06-04
POI disappears unnecessarily when zooming in or rotating


1ec5:
The “Restaurant Tsugaru” POI in the first screenshot disappears as soon as you zoom in to the next level, even though there doesn’t appear to be anything that would collide with it that wasn’t in the previous zoom level:

![out](https://cloud.githubusercontent.com/assets/1231218/7988544/e65e22c8-0a98-11e5-8f68-5f9fb92d70fd.PNG)

![in](https://cloud.githubusercontent.com/assets/1231218/7988559/f4daece6-0a98-11e5-8356-1ca808adf130.PNG)

When rotating, I see this POI possibly avoiding collisions with a nearby one (“Shinmori Optometry”), but at certain angles, neither POI shows up, probably because of each other:

![food](https://cloud.githubusercontent.com/assets/1231218/7988693/99baa04e-0a99-11e5-9f56-e9a8199f4474.PNG)

![nothing](https://cloud.githubusercontent.com/assets/1231218/7988708/a5b14362-0a99-11e5-9695-c46fb01898a2.PNG)

![sight](https://cloud.githubusercontent.com/assets/1231218/7988712/acb98bc4-0a99-11e5-9ec8-8e0990662a0e.PNG)

[In mapbox-gl-js](https://www.mapbox.com/mapbox-gl-styles/#17.69/37.34897/-121.89358/-16.9), the same POIs also blink in and out when zooming and rotating, though the thresholds are slightly different.

friedbunny:
Not the best gif, but Tsugaru Restaurant disappears at max zoom and the surrounding collision boxes become red.

![ezgif com-gif-maker](https://cloud.githubusercontent.com/assets/1198851/8491398/c9bb0e64-2105-11e5-8573-32a64cf1b130.gif)

friedbunny:
Still happens.

ansis:
This is expected right now

The label that disappears is blocked by other labels that will show up later in the zoom level (when you zoom in a bit more). It's implemented this way to reduce labels popping in and out all the time while zooming.

I agree that it would be better if the label didn't disappear but I'm going to treat this as an enhancement, not a bug.

I don't see us doing a major revision of the labelling algorithm anytime soon but I'm keeping track of all the algorithm suggestions in a separate master ticket. They all impact each other so I think it makes sense to close the individual issues and keep the ideas in one spot: https://github.com/mapbox/mapbox-gl/issues/14

[...]

PR #3712, opened 2016-01-27
add setters, getters for map's minZoom and maxZoom

ansis:
fixes https://github.com/mapbox/mapbox-gl-native/issues/509

- [x] @tobrun can you review the android commit?
- [x] @1ec5 can you review the ios and osx commits?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@
 * Fixed crash caused by annotation image with non-integer width or height ([#3031](https://github.com/mapbox/mapbox-gl-native/issues/3031))
 * Tracking Mode Reverses Bearing Fix ([#3664](https://github.com/mapbox/mapbox-gl-native/issues/3664))
 * GPS Extra Rotation Fix ([#3661](https://github.com/mapbox/mapbox-gl-native/issues/3661))
+* Added new methods for getting and setting the min and max zoom levels: `getMinZoom`, `setMinZoom`, `getMaxZoom`, `setMaxZoom`. ([#509](https://github.com/mapbox/mapbox-gl-native/issues/509))
 
 ## 3.1.0
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -75,6 +75,8 @@ Known issues:
 - Fixed crash caused by MGLAnnotationImage with non-integer width or height ([#2198](https://github.com/mapbox/mapbox-gl-native/issues/2198))
 - Fixed “include of non-modular header” errors in Swift projects managed by CocoaPods. ([#3679](https://github.com/mapbox/mapbox-gl-native/pull/3679)) 
 - Avoids triggering the blue background location status bar when user has granted "when in use" permission. ([#3671](https://github.com/mapbox/mapbox-gl-native/issues/3671))
+- Added new methods to MGLMapView for getting and setting the minimum and maximum zoom levels: `minimumZoomLevel`, `setMinimumZoomLevel`, `maximumZoomLevel`, `setMaximumZoomLevel`. ([#509](https://github.com/mapbox/mapbox-gl-native/issues/509))
+- The default `maximumZoomLevel` changed from 18 to 20.
 
 ## iOS 3.0.1
 
--- a/include/mbgl/ios/MGLMapView.h
+++ b/include/mbgl/ios/MGLMapView.h
@@ -464,6 +464,30 @@ IB_DESIGNABLE
 - (void)setZoomLevel:(double)zoomLevel animated:(BOOL)animated;
 
 /**
+ * The minimum zoom level at which the map can be shown.
+ *
+ * Depending on the map view’s aspect ratio, the map view may be prevented
+ * from reaching the minimum zoom level, in order to keep the map from
+ * repeating within the current viewport.
+ *
+ * If the value of this property is greater than that of the
+ * maximumZoomLevel property, the behavior is undefined.
+ *
+ * The default minimumZoomLevel is 0.
+ */
+@property (nonatomic) double minimumZoomLevel;
+
+/**
+ * The maximum zoom level the map can be shown at.
+ *
+ * If the value of this property is smaller than that of the
+ * minimumZoomLevel property, the behavior is undefined.
+ *
+ * The default maximumZoomLevel is 20.
+ */
+@property (nonatomic) double maximumZoomLevel;
+
+/**
  The heading of the map, measured in degrees clockwise from true north.
  
  The value `0` means that the top edge of the map view corresponds to true
--- a/include/mbgl/osx/MGLMapView.h
+++ b/include/mbgl/osx/MGLMapView.h
@@ -182,13 +182,29 @@ IB_DESIGNABLE
  */
 @property (nonatomic) double zoomLevel;
 
-/** The minimum zoom level that can be displayed by the receiver using the
-    current style. */
-@property (nonatomic, readonly) double maximumZoomLevel;
+/**
+ * The minimum zoom level at which the map can be shown.
+ *
+ * Depending on the map view’s aspect ratio, the map view may be prevented
+ * from reaching the minimum zoom level, in order to keep the map from
+ * repeating within the current viewport.
+ *
+ * If the value of this property is greater than that of the
+ * maximumZoomLevel property, the behavior is undefined.
+ *
+ * The default minimumZoomLevel is 0.
+ */
+@property (nonatomic) double minimumZoomLevel;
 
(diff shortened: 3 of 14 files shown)


include/mbgl/ios/MGLMapView.h
```diff
@@ -404,6 +404,30 @@ IB_DESIGNABLE
 - (void)setZoomLevel:(double)zoomLevel animated:(BOOL)animated;
 
 /**
+ * The minimum zoom level the can be shown at.
```
1ec5: “the **map** can be”

include/mbgl/osx/MGLMapView.h
```diff
@@ -184,11 +184,25 @@ IB_DESIGNABLE
...
     current style. */
-@property (nonatomic, readonly) double maximumZoomLevel;
+@property (nonatomic) double maximumZoomLevel;
+
+/**
+ * Changes the maximum zoom level of the map.
+ *
+ * @param maximumZoomLevel The new minimum zoom level.
+ */
+- (void)setMaximumZoomLevel:(double)maximumZoomLevel;
```
1ec5: These changes aren’t necessary. By removing the `readonly` keyword from the property, you implicitly added a `-setMaximumZoomLevel:` method to the class.

include/mbgl/ios/MGLMapView.h
1ec5: It’s worth noting here that the minimum zoom level can still be [constrained](https://github.com/mapbox/mapbox-gl-native/blob/8357b3716bfc563c2fb436a51a42dc6b6bfe3c32/src/mbgl/map/transform_state.cpp#L365-L368) regardless of this property to prevent showing multiple copies of the world at the same time. This is relevant whenever the view is much wider than it is tall.

include/mbgl/ios/MGLMapView.h
```diff
@@ -404,6 +404,30 @@ IB_DESIGNABLE
...
+ */
+@property (nonatomic) double minimumZoomLevel;
+
+/**
+ * Changes the minimum zoom level of the map.
+ *
+ * @param minimumZoomLevel The new minimum zoom level.
+ */
+- (void)setMinimumZoomLevel:(double)minimumZoomLevel;
```
1ec5: Same as https://github.com/mapbox/mapbox-gl-native/pull/3712/files#r51043942.

include/mbgl/ios/MGLMapView.h
```diff
@@ -404,6 +404,30 @@ IB_DESIGNABLE
...
+
+/**
+ * The maximum zoom level the can be shown at.
+ */
+@property (nonatomic) double maximumZoomLevel;
```
1ec5: We should document that the default value of this property is 20 (but could be further constrained by the style?).

CHANGELOG.md
```diff
@@ -67,6 +68,7 @@ Known issues:
 - Fixed crash caused by MGLAnnotationImage with non-integer width or height ([#2198](https://github.com/mapbox/mapbox-gl-native/issues/2198))
 - Fixed “include of non-modular header” errors in Swift projects managed by CocoaPods. ([#3679](https://github.com/mapbox/mapbox-gl-native/pull/3679)) 
 - Avoids triggering the blue background location status bar when user has granted "when in use" permission. ([#3671](https://github.com/mapbox/mapbox-gl-native/issues/3671))
+- Added new methods to MGLMapView for getting and setting the minimum and maximum zoom levels: `minimumZoomLevel`, `setMinimumZoomLevel`, `maximumZoomLevel`, `setMaximumZoomLevel`. ([#509](https://github.com/mapbox/mapbox-gl-native/issues/509))
```
1ec5: Also mention that the default maximum zoom level has been increased from 18 to 20.

1ec5:
@friedbunny @boundsj, should these properties be made inspectable in Interface Builder? Or would it be confusing that setting the minimum zoom level to 0 won’t necessarily allow the user to zoom out to z0 if the view is wide enough?

include/mbgl/ios/MGLMapView.h
ansis: how would you phrase this in the documentation? I'm having trouble coming up with a clear explanation

CHANGELOG.md
ansis: I'm not actually sure if the default change is a good idea. What do you think?

include/mbgl/ios/MGLMapView.h
1ec5: How about:

> The minimum zoom level at which the map can be shown.
>
> Depending on the map view’s aspect ratio, the map view may be prevented from reaching the minimum zoom level, in order to keep the map from repeating within the current viewport.
>
>  If the value of this property is greater than that of the `maximumZoomLevel` property, the behavior is undefined.

CHANGELOG.md
1ec5: What’s the default in GL JS? [22](https://github.com/mapbox/mapbox-gl-js/blob/739aa39e340f523fd9b7e431a0bb366b41a91711/js/geo/transform.js#L28)? Are the issues enumerated in https://github.com/mapbox/mapbox-gl-native/issues/509#issuecomment-61131373 still relevant?

include/mbgl/ios/MGLMapView.h
ansis: perfect, thanks

CHANGELOG.md
ansis: the actual default is 20, https://github.com/mapbox/mapbox-gl-js/blob/739aa39e340f523fd9b7e431a0bb366b41a91711/js/ui/map.js#L142 . The default in transform.js should probably be removed.

The issues in https://github.com/mapbox/mapbox-gl-native/issues/509#issuecomment-61131373 are gone (fixed by https://github.com/mapbox/mapbox-gl-native/pull/1666),

1ec5:
What happens if the map is already at z0 when the minimum zoom level is set to z2? It seems like the map would suddenly snap to z2 as soon as you touch it. We should look into handling this special case at the SDK level by first animating to the new minimum zoom level then officially setting that minimum. In any case, once the comments above are addressed, this PR will be good to go from my perspective.

ansis:
> What happens if the map is already at z0 when the minimum zoom level is set to z2? It seems like the map would suddenly snap to z2 as soon as you touch it. We should look into handling this special case at the SDK level by first animating to the new minimum zoom level then officially setting that minimum.

It will immediately snap to z2 when the new minimum zoom is set: https://github.com/mapbox/mapbox-gl-native/commit/196feb9a4ac4d21ae15879f04616ec185f51300d#diff-fd1625168ba48a6fb963e63343c0ac59R308

I think immediately jumping is fine since this isn't something you would usually call during use. With easing you'd have to worry about:
- only actually using the new minimum it after the animation runs
- what happens if the animation is cancelled
- what getMinZoom returns while the animation is running

CHANGELOG.md
1ec5: Yes, we should increase the default to 20, then. In dense, urban business districts at z18, POIs often collide each other out, leaving a pretty sparse map (as seen in #1690).

1ec5:
> only actually using the new minimum it after the animation runs
> what getMinZoom returns while the animation is running

When it comes to properties like this, on iOS and OS X, we’ve provided additional setters like `-setContentInset:animated:`, with the understanding that immediately calling `-contentInset` afterwards may not yield the final value.

> what happens if the animation is cancelled

The minimum or maximum zoom level would be set inside the `transitionFinishFn` of the animation options passed into `easeTo()`, which is guaranteed to be called even if the animation is canceled.

> It will immediately snap to z2 when the new minimum zoom is set

In that case, I’m fine with the behavior you’ve implemented. :+1:

ansis:
Thanks for the background on animated setters and transitions @1ec5. Good to know!

friedbunny:
>  should these properties be made inspectable in Interface Builder

I vote... yes? (Halfheartedly.)

That said, most of the zoom-related requests we've seen are related to bumping the maximum value, but if we set that to the effective maximum by default then that'd be satisfactory there.

1ec5:
> if we continue to set that to the effective maximum by default then that'd be satisfactory there

That’s a good point. Let’s revisit the inspectable idea later if the use cases become clear.

tobrun:
@ansis 
The Android commit looks great!

About this:
> I think immediately jumping is fine since this isn't something you would usually call during use. 

With the 4.0.0 release we are probably going to expose this as a configuration in a builder pattern. The developer will not be able to change the behaviour after the map is created.

1ec5:
> should these properties be made inspectable in Interface Builder?

Tracked in #4388.

PR #3713, opened 2016-01-27
[iOS] Generate and send a turnstile event when map view event is queued

boundsj:
This change adds logic to push a turnstile event whenever a map view
event is queued. It guards against sending the turnstile event more
than once per session.

Note that `pushTurnstileEvent` is actually a "force push" in that it
flushes the queue and, since it enqueues a turnstile event, the
flush's call to `postEvents` will attempt to send any queued events
even if telemetry is paused. This has the side effect of defeating
the other queue / timing / flush logic in place whenever map load
(and now implicitly turnstile) events are pushed.

This change does take care to continue to ensure that, in `pushEvent` even if the 
events instance is paused, the turnstile event is still pushed. 

This change drops the use of the priority background queue that used
to be used to dispatch async `pushTurnstileEvent`. I could not see how the
priority queue was required for that. The subsequently called implementations of 
`pushTurnstileEvent` and `flush` still use the custom serial queue
and the `postEvents` method dispatches using a priority background
queue.

cc @1ec5 @friedbunny

### DIFF
--- a/platform/ios/src/MGLMapboxEvents.m
+++ b/platform/ios/src/MGLMapboxEvents.m
@@ -275,11 +275,6 @@ const NSTimeInterval MGLFlushInterval = 60;
              MGLMapboxEvents *strongSelf = weakSelf;
              [strongSelf validate];
          }];
-
-        // Turn the Mapbox Turnstile to Count App Users
-        dispatch_async(dispatch_get_global_queue(DISPATCH_QUEUE_PRIORITY_BACKGROUND, 0), ^{
-            [self pushTurnstileEvent];
-        });
     }
     return self;
 }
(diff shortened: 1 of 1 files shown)


platform/ios/src/MGLMapboxEvents.m
```diff
@@ -429,10 +424,15 @@ - (void)startUpdatingLocation {
     }
 }
 
+static BOOL turnstileEventSent = NO;
+
 - (void) pushTurnstileEvent {
 
-    __weak MGLMapboxEvents *weakSelf = self;
+    // Turnstile events only need to be sent up once per application launch
+    if (turnstileEventSent) { return; }
```
friedbunny: We should make a call on `if () return` style in MGLMapboxEvents and harmonize the entire thing.

platform/ios/src/MGLMapboxEvents.m
```diff
@@ -429,10 +424,15 @@ - (void)startUpdatingLocation {
     }
 }
 
+static BOOL turnstileEventSent = NO;
```
friedbunny: Can we make this a property defined on MGLMapboxEvents?

platform/ios/src/MGLMapboxEvents.m
```diff
@@ -456,6 +456,8 @@ - (void) pushTurnstileEvent {
         
         // Flush
         [strongSelf flush];
```
friedbunny: When telemetry is enabled, the turnstile will always be sent roughly simultaneously with the map load event. I wonder if we should reevaluate always flushing the event queue here.

platform/ios/src/MGLMapboxEvents.m
1ec5: Instead of this global static variable or a property on MGLMapboxEvents, `-pushTurnstileEvent` should be wrapped in a `dispatch_once()` block guarded by a static local variable. More broadly, we should eliminate all this complexity around thread safety with #3717.

platform/ios/src/MGLMapboxEvents.m
boundsj: Agreed @friedbunny. For this PR, I just wanted to make sure that the turnstile event post continues to always be attempted, expediently, irrespective of the state of telemetry. I think that larger refactors to separate out networking and to remove threading complexity may make things clearer here, too. It might make sense to have a different queue or no queue at all for the turnstile event (i.e. just post it right away).

friedbunny:
🚧:+1:

1ec5:
:+1:

PR #3714, opened 2016-01-27
[ios] Remove incorrect tab whitespace in method

boundsj:
This just removes some whitespace that made the implementation of
`pushTurnstileEvent` in `MGLMapboxEvents` look a little strange.

The indentation was right below an early return conditional
(with no braces) so it felt like the right thing to do (for
readability) to clean it up. Choosing and enforcing a consistent style
for conditionals with one line might be a good idea, too, but this
small change does not go there.

cc @1ec5 @friedbunny


1ec5:
:+1:

Issue #3647, opened 2016-01-21
Enable HTTP pipelining on iOS


kkaefer:
The default is set to `NO`

/cc @1ec5

kkaefer:
This might explain why cURL is faster than NSURL*

jfirebaugh:
It sounds to me like HTTP pipelining is too new/dangerous to use. If the server you're connecting to (or proxy you're connecting through) doesn't implement it, you can get incorrect results without warning.

http://stackoverflow.com/questions/14810890/what-are-the-disadvantages-of-using-http-pipelining

kkaefer:
Interesting, I thought it was accepted standard practice, but apparently not.

friedbunny:
iOS 9 and OkHTTP both have HTTP/2 enabled by default, which implements multiplexing. When Mapbox content servers start supporting HTTP/2, we should get it for free.

PR #3715, opened 2016-01-27
Offline (mbgl-core implementation)

jfirebaugh:
TODO list:

* [x] Normalize Mapbox tile URLs to `mapbox://tiles` scheme before requesting, translate back to `api.mapbox.com` URLs inside `OnlineFileSource`.
* [x] Remove `sqlite_cache.{hpp,cpp}`
* [x] Style, source, sprite, glyphs downloading
* [x] Figure out progress semantics when complete file count is indeterminate
* [x] Font stack normalization
* [x] Verify that `LatLngBounds` enforces the correct invariants: https://github.com/mapbox/mapbox-gl-native/issues/3802
* [x] Tiles downloading
* [x] Tile ID wrapping
* [x] Region deletion
* [x] Test cases
  * [x] No sprite
  * [x] No glyphs
  * [x] Inline source
  * [x] GeoJSON source
* [x] Size reporting
* [x] Remove schema version logic; once shipped we need to write migrations rather than dumping existing dbs
* [x] Review use of `Request::Error::NotFound`
* [x] Get status of inactive download
* [x] Error handling
* [x] Resource deletion
* [x] Concurrency
* [x] Performance
  * [x] [Concurrent request throttling](https://github.com/mapbox/mapbox-gl-native/issues/3858)
  * ~~[ ] [Request pipelining](https://github.com/mapbox/mapbox-gl-native/issues/3647)~~
  * [x] Add indexes
* [x] Size limits / eviction
* [x] Polish schema
* [x] Consistently use `?NNN` syntax for prepared statements

#### Think about

* [x] Style spec migrations -- After we release v9, we'll need to continue supporting offline databases containing v8 styles, via version branches in the renderer, or JIT migration.
* [x] Static evaluation of font stacks vs. data driven styling -- The required font stacks need to be determined without loading data. Therefore we should either not support data-driven `"text-font"`, or require that all possible font stacks values be listed elsewhere, like we do with `"icon-image"`.
* [x] Stop tile pyramid recursion when encountering a nodata tile -- This is an potential future optimization to avoid storing nodata child tiles once we encounter a nodata parent. This depends on #123.


jfirebaugh:
This is largely complete and ready to receive code reviews. :eyes: @zugaldia @1ec5 @brunoabinader @tmpsantos @kkaefer.

jfirebaugh:
I wrote a simple test binary to download the Bay Area z0-15 using the core offline API. This completed without errors, resulting in 231.58 MB of downloads in 6 minutes (~650 kb/sec on office wifi), and a database that is 237.4 MB on-disk. :facepunch:

zugaldia:
Bra-vo.

tmpsantos:
@jfirebaugh amazing work. :+1: 

My only concern really is just that we can no longer limit the size of the cache. I know the implementation using `page_size` was not exactly the most efficient but SQLite is very limited on this regard.

Being able to set the cache on the `DefaultFileSource` would be a nice to have. I for instance was experimenting with a two tiers `SQLiteCache` (basically one in memory and the other one in disk).

jfirebaugh:
@tmpsantos I'm going to reimplement the size limitations today. As a first pass, it will apply to ambient caching only -- resources used by offline regions won't count towards any limits, and will never be evicted. This will make the behavior of OfflineDatabase identical to SQLiteCache in cases where you don't create any offline regions. An eviction strategy that does involve offline resources is a harder problem and will take some careful thought.

tmpsantos:
> This will make the behavior of OfflineDatabase identical to SQLiteCache

:+1: Thanks, let me know if I can help.

jfirebaugh:
@tmpsantos Check out the latest commit!

jfirebaugh:
@tmpsantos How about now?

tmpsantos:
> @tmpsantos How about now?

:shipit: kudos

jfirebaugh:
Notes from the "Think about" section:

* Style spec migrations -- After we release v9, we'll need to continue supporting offline databases containing v8 styles, via version branches in the renderer, or JIT migration.
* Static evaluation of font stacks vs. data driven styling -- The required font stacks need to be determined without loading data. Therefore we should either not support data-driven `"text-font"`, or require that all possible font stacks values be listed elsewhere, like we do with `"icon-image"`.
* Stop tile pyramid recursion when encountering a nodata tile -- This is an potential future optimization to avoid storing nodata child tiles once we encounter a nodata parent. This depends on #123.

bsudekum:
Congrats :tada:!

Issue #2644, opened 2015-10-16
~Actor<O> is blocking


kkaefer:
When canceling a `WorkTask`, we block until that task has completed its entire invocation. This isn't something we should do. You can make this behavior more visible when adding a `sleep(4)` in a worker task, (e.g. in `TileWorker::parse()`). Panning fast now stalls heavily, as partial work requests are getting canceled.

/cc @jfirebaugh

jfirebaugh:
This is by design. `WorkTask`s typically have references to objects which need to outlive the task, whose owner is in turn owning the task. If the task isn't canceled, then these lifetime invariants will not hold: the objects will be destroyed along with their owner, while the work task will continue running with dangling references.

jfirebaugh:
Worker tasks should be designed to check a flag in inner loops and exit early if the flag indicates the task has been cancelled. Most do; the exception I see is `SymbolBucket`, which has no early exit for the loops in `needsDependencies` and `addFeatures`.

tmpsantos:
> This is by design.

And I like the design here because protects us from bugs that are very difficult to debug that haunted us in the past. On #3636 suggested a special deleter for `TileData` that if turns out to be needed somewhere else we could make it more generic and call it `NonBlockerDeleter` and have a method like `::willBlockOnDestructor()` to postpone the deletion on objects with `WorkRequest`s.

kkaefer:
This is resurfacing: When parsing tiles, we're checking a flag after parsing every feature whether we still need the tile. However, our landuse layer, in particular on zoom level 7, contains some *very* complex features that take > 1 second to parse in their call to clipper + libtess2. These tasks are not interruptible. We should do two things:

1. Make tile parsing cancellation non-blocking. This means that parsing a tile must not access resources shared with the main thread or other threads.
2. Refactor our landuse layer to avoid single polygons that are too complex

/cc @mourner

jfirebaugh:
As of #6337 there's no cancellation of tasks, but destroying the actor can still block. We need #667 in order to remove that.

jfirebaugh:
It should be safe to change this now.

PR #3724, opened 2016-01-28
[core] Blocking the Map Thread when destroying TileData objects

tmpsantos:
We create and destroy TileData objects constantly, specially when zooming in/out fast, and we might block and destroying them. This issue is noticeable on iOS (tested on a iPad Mini 2 Retina) and can be reproduced two ways:

- Pinch the map fast, zooming in and out and you will see the map hanging sometimes.
- Move the map to a place somewhat far from the current location and click on the top right button to trigger a "fly to" animation, that should hang a few times (can be made worse with the map on full tilt).

While the first scenario is hardly a real life use case, the second is. The root cause is that we cannot cancel worker tasks once they are started (see #2644).

### DIFF
--- a/include/mbgl/util/run_loop.hpp
+++ b/include/mbgl/util/run_loop.hpp
@@ -1,8 +1,11 @@
 #ifndef MBGL_UTIL_RUN_LOOP
 #define MBGL_UTIL_RUN_LOOP
 
+#include <mbgl/platform/log.hpp>
+#include <mbgl/util/chrono.hpp>
 #include <mbgl/util/noncopyable.hpp>
 #include <mbgl/util/util.hpp>
+#include <mbgl/util/thread_context.hpp>
 #include <mbgl/util/work_task.hpp>
 #include <mbgl/util/work_request.hpp>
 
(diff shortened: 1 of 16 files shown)


tmpsantos:
:eyes: @jfirebaugh @kkaefer

tmpsantos:
Also easy to see using the "Start World Tour" option on the iOS app.

jfirebaugh:
This adds additional complexity to an area where reasoning about threading behavior and object lifetimes is already difficult. I would much prefer we continue to investigate and eliminate the direct causes of blocking, rather than delaying TileData destructors.

tmpsantos:
> This adds additional complexity to an area where reasoning about threading behavior and object lifetimes is already difficult. I would much prefer we continue to investigate and eliminate the direct causes of blocking, rather than delaying TileData destructors.

The alternative I see would be change how we use workers. We would always transfer the ownership of what is pass on the parameters to the worker, so cancellation would still need a mutex for flipping a flag, but executing the task on the worker would not acquire that mutex. The worker would only acquire the mutex for posting back the callback to the Map Thread or not if it was cancelled.

Canceling an ongoing task would be a matter of let the task go on and self destruct at the end. Only destroying the worker itself would block.

Good thing is these objects momentarily owned by the worker would be destroyed on the worker thread in this case, offloading the destruction job from the Map Thread, but I don't think this will make a difference. The advantage I see is cancelling a task can be considered effectively non-blocking (not truly because the cancellation flag mutex).

Bad thing is we need to enforce a safe design pattern that I don't think it can be done programmatically like it is today. We would only pass as arguments to the worker `std::unique_ptr` or `std::shared_ptr`. The later only to immutable objects.

1ec5:
Pulling off the 3.1.0 milestone given #3731.

kkaefer:
Thinking some more about this, I think we should actually implement what @tmpsantos is suggesting in https://github.com/mapbox/mapbox-gl-native/pull/3724#issuecomment-176742934. Having a custom deleter will add unnecessary complexity and essentially add a custom deferred shared pointer destruction that could be very complicated to debug.

This is what our workers currently use: 

- layers from the stylesheet (these are already copied ✓)
- actual tile data (they are moved ✓)
- placement configuration (already copied ✓)
- tile ID, source ID, map mode (already copied ✓)
- **SpriteStore, GlyphAtlas, GlyphStore**, these are currently references and shared among all workers and the main thread. In JS, these live on the main thread, and web workers get a copy when initializing and communicate with the main thread to mutate them. We should do this in native as well, and make it easier to communicate across thread boundaries.

jfirebaugh:
Agreed @kkaefer; that's been our long term goal since #667.

jfirebaugh:
Closing; I think we should investigate alternative solutions here per above comments.

PR #3731, opened 2016-01-28
fix redoPlacement blocking main thread, fix #3727

### LINKED ISSUE #3727: [core] Text placement might block the Map Thread
When redoing the text placement we try to cancel any ongoing task on the worker and that might block (see #2644) and cause the UI to hang.

We should never block and at the same time we must make sure that the text placement is deterministic, which is the reason why the current code tries to stop the ongoing redo placement task and start a new one synchronously.

/cc @ansis

ansis:
fix #3727

If a workRequest exists, don't cancel it and start an new request. It's ok to just do nothing if the request exists because a new call to redoPlacement will be triggered after the existing request finishes.

This fixes a regression in https://github.com/mapbox/mapbox-gl-native/pull/3543

@tmpsantos can you review?

### DIFF
--- a/src/mbgl/map/vector_tile_data.cpp
+++ b/src/mbgl/map/vector_tile_data.cpp
@@ -144,16 +144,15 @@ void VectorTileData::redoPlacement(const PlacementConfig newConfig, const std::f
     if (newConfig != placedConfig) {
         targetConfig = newConfig;
 
-        if (!workRequest) {
-            // Don't start a new placement request when the current one hasn't completed yet, or when
-            // we are parsing buckets.
-            redoPlacement(callback);
-        }
+        redoPlacement(callback);
     }
 }
 
 void VectorTileData::redoPlacement(const std::function<void()>& callback) {
-    workRequest.reset();
+    // Don't start a new placement request when the current one hasn't completed yet, or when
+    // we are parsing buckets.
+    if (workRequest) return;
+
     workRequest = worker.redoPlacement(tileWorker, buckets, targetConfig, [this, callback, config = targetConfig] {
         workRequest.reset();
 


src/mbgl/map/vector_tile_data.cpp
```diff
@@ -144,15 +144,15 @@ void VectorTileData::redoPlacement(const PlacementConfig newConfig, const std::f
...
+        redoPlacement(callback);
     }
 }
 
 void VectorTileData::redoPlacement(const std::function<void()>& callback) {
+    // Don't start a new placement request when the current one hasn't completed yet, or when
+    // we are parsing buckets.
+    if (workRequest) return;
+
     workRequest.reset();
```
jfirebaugh: This line can be removed.

ansis: that line, and two others removed in fa522e02fe8fe58382a82643f9e4884136c59b76

ansis: It's ok to remove the other two because `workRequest =` will destruct/cancel the old one, right?

jfirebaugh: It's safe to remove this one; `if (workRequest) return;` guarantees that `reset()` would be a no-op. For the other ones, there's a difference:

* reset, create new: Ensures the current task is ended and the callback for it will not be called, then starts a new one.
* create new, assign: The callback for the previous task could still be executed, which might put things into an unexpected state.

ansis: right, thanks! I should've checked. changed.

tmpsantos:
:+1:

tmpsantos:
I had the impression that by doing this we could potentially make text placement non deterministic.

I.e. if you rotate from angle A to B fast enough, in some cases we could skip the last `redoPlacement` because there was already a task going on with outdated values. So different "A -> B" could end up with labels in different places.

My solution was in a situation like there is a task ongoing, queue the `redoPlacement` for the next main loop interaction.

ansis:
If you're rotating fast enough it could skip some of the intermediate `redoPlacement` calls, but the last one should always be the same. Since the results of `redoPlacement` should not depend on previous placement, it should be deterministic.

Whenever the map is rotated `redoPlacement` is [called here](https://github.com/mapbox/mapbox-gl-native/blob/fix-redo-placement-blocking/src/mbgl/map/source.cpp#L499) which updates the [placement config the tile needs to eventually get to](https://github.com/mapbox/mapbox-gl-native/blob/fix-redo-placement-blocking/src/mbgl/map/vector_tile_data.cpp#L145). After a tile is [loaded](https://github.com/mapbox/mapbox-gl-native/blob/53dda0f75b180a4280a9276f6fd2e2e3923dae2e/src/mbgl/map/source.cpp#L548) or redoPlacement [finishes](https://github.com/mapbox/mapbox-gl-native/blob/53dda0f75b180a4280a9276f6fd2e2e3923dae2e/src/mbgl/map/vector_tile_data.cpp#L170), it is called again until the placement is calculated for the latest PlacementConfig.

tmpsantos:
:100: thanks a lot for explaining,

PR #3733, opened 2016-01-28
[core] Increment the schema version

jfirebaugh:
This is a followup to 1c21d0fd4cd30cbf6c5b863fd0179b227c28bc0b. Prior to that commit, resources without an expiration were stored with an expiration column value of 0. After it, they are stored with a null column value. Without a schema increment, existing resources in the cache with a 0 value would be treated as expiring at the epoch, causing endless revalidation.

cc @bleege


jfirebaugh:
Obsoleted by #3715.

PR #3737, opened 2016-01-28
Mark strip frameworks script executable

1ec5:
Xcode keeps unsetting the executable flag on this script, leading to the following build error in projects that embed the dynamic framework:

```
/Mapbox.app/Frameworks/Mapbox.framework/strip-frameworks.sh: Permission denied
```

/cc @bsudekum


1ec5:
Everything passes (other than the intermittent KIF failure). Merging.

PR #3738, opened 2016-01-29
Eliminate “flashlight mode”

### LINKED ISSUE #3710: Heading tracking mode sometimes fails to rotate map after device rotation
Haven't pinned down exact reproduction steps, but flipping through the tracking modes in iosapp and rotating the device back-and-forth, landscape-portrait will sometimes result in this behavior in heading tracking mode:

flashlight

... a non-rotating map and a jerky (but cool) "flashlight".

1ec5:
Avoid resetting user tracking state when going from location or course tracking mode to heading tracking mode.

Fixes #3710.

/cc @friedbunny


friedbunny:
We barely knew ye, flashlight mode.

PR #3740, opened 2016-01-29
[core] Rationalize Resource initialization

jfirebaugh:
Extract URL construction code for reuse in #3715.

### DIFF
--- a/include/mbgl/storage/resource.hpp
+++ b/include/mbgl/storage/resource.hpp
@@ -31,6 +31,18 @@ public:
     optional<SystemTimePoint> priorModified;
     optional<SystemTimePoint> priorExpires;
     optional<std::string> priorEtag;
+
+    // Includes auxiliary data if this is a tile request.
+
+    struct TileData {
+        std::string urlTemplate;
+        float pixelRatio;
+        int32_t x;
+        int32_t y;
+        int8_t z;
+    };
+
+    optional<TileData> tileData;
 };
 
 } // namespace mbgl
--- a/include/mbgl/storage/resource.hpp
+++ b/include/mbgl/storage/resource.hpp
@@ -20,20 +20,6 @@ public:
         SpriteJSON
     };
 
-    Resource(Kind kind_, const std::string& url_)
-        : kind(kind_),
-          url(url_) {
-    }
-
-    const Kind kind;
-    const std::string url;
-
-    optional<SystemTimePoint> priorModified;
-    optional<SystemTimePoint> priorExpires;
-    optional<std::string> priorEtag;
-
-    // Includes auxiliary data if this is a tile request.
-
     struct TileData {
         std::string urlTemplate;
         float pixelRatio;
@@ -42,7 +28,34 @@ public:
         int8_t z;
     };
 
+    Resource(Kind kind_, const std::string& url_, optional<TileData> tileData_ = {})
+        : kind(kind_),
+          url(url_),
+          tileData(tileData_) {
+    }
+
+    static Resource style(const std::string& url);
+    static Resource source(const std::string& url);
+    static Resource tile(const std::string& urlTemplate,
+                         float pixelRatio,
+                         int32_t x,
+                         int32_t y,
+                         int8_t z);
+    static Resource glyphs(const std::string& urlTemplate,
+                           const std::string& fontStack,
+                           const std::pair<uint16_t, uint16_t>& glyphRange);
+    static Resource spriteImage(const std::string& base, float pixelRatio);
+    static Resource spriteJSON(const std::string& base, float pixelRatio);
+
+    const Kind kind;
+    const std::string url;
+
+    // Includes auxiliary data if this is a tile request.
     optional<TileData> tileData;
+
+    optional<SystemTimePoint> priorModified = {};
+    optional<SystemTimePoint> priorExpires = {};
+    optional<std::string> priorEtag = {};
 };
 
 } // namespace mbgl
--- a/include/mbgl/storage/resource.hpp
+++ b/include/mbgl/storage/resource.hpp
@@ -22,7 +22,7 @@ public:
 
     struct TileData {
         std::string urlTemplate;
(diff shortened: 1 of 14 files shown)


jfirebaugh:
This is a partial reroll of #3675 (the least risky parts).

include/mbgl/storage/resource.hpp
```diff
@@ -20,17 +20,42 @@ class Resource {
         SpriteJSON
     };
 
-    Resource(Kind kind_, const std::string& url_)
+    struct TileData {
+        std::string urlTemplate;
+        float pixelRatio;
```
kkaefer: Does this need to be a float? Are we supporting assets that have a fractional pixel ratio?

include/mbgl/storage/resource.hpp
```diff
@@ -20,17 +20,42 @@ class Resource {
...
+    static Resource tile(const std::string& urlTemplate,
+                         float pixelRatio,
+                         int32_t x,
+                         int32_t y,
+                         int8_t z);
+    static Resource glyphs(const std::string& urlTemplate,
+                           const std::string& fontStack,
+                           const std::pair<uint16_t, uint16_t>& glyphRange);
+    static Resource spriteImage(const std::string& base, float pixelRatio);
+    static Resource spriteJSON(const std::string& base, float pixelRatio);
```
kkaefer: Why static functions rather than constructors that discriminate on `Kind`?

src/mbgl/map/raster_tile_data.cpp
```diff
@@ -9,23 +9,18 @@
...
-    cancel();
-}
-
-void RasterTileData::request(const std::string& url,
-                             const RasterTileData::Callback& callback) {
     state = State::loading;
 
-    FileSource* fs = util::ThreadContext::getFileSource();
-    req = fs->request({ Resource::Kind::Tile, url }, [url, callback, this](Response res) {
+    Resource resource = Resource::tile(urlTemplate, pixelRatio, id.x, id.y, id.sourceZ);
```
kkaefer: `const`

src/mbgl/map/vector_tile.cpp
```diff
@@ -178,13 +178,15 @@ util::ptr<const GeometryTileFeature> VectorTileLayer::getFeature(std::size_t i)
...
+VectorTileMonitor::VectorTileMonitor(const TileID& tileID_, float pixelRatio_, const std::string& urlTemplate_)
+    : tileID(tileID_),
+      pixelRatio(pixelRatio_),
+      urlTemplate(urlTemplate_) {
 }
 
 std::unique_ptr<FileRequest> VectorTileMonitor::monitorTile(const GeometryTileMonitor::Callback& callback) {
-    const std::string url = util::templateTileURL(urlTemplate, tileID);
-    return util::ThreadContext::getFileSource()->request({ Resource::Kind::Tile, url }, [callback, this](Response res) {
+    Resource resource = Resource::tile(urlTemplate, pixelRatio, tileID.x, tileID.y, tileID.sourceZ);
```
kkaefer: `const`

include/mbgl/storage/resource.hpp
kkaefer: I know we support fractional pixel ratios when rendering, but should we always round up to the nearest integer pixel ratio when requesting assets?

include/mbgl/storage/resource.hpp
jfirebaugh: This way enforces the appropriate arguments for each kind, and is more succinct.

```
const Resource style = Resource::style(url);
```

vs

```
const Resource style(Resource::Kind::Style, url);
const Resource tile(Resource::Kind::Tile, url); // wrong arguments
```

Better would be to make `Resource` a variant type, but that's a more invasive change than I want to make right now.

include/mbgl/storage/resource.hpp
jfirebaugh: It's a bit of a grey area because technically we don't support anything other than `@2x`, and that only for Mapbox sources. So if the parameter is to represent strictly what's supported _now_, it could be an integer or even `bool retinaTiles`.

In terms of offline, what I am planning to do is extract a `tileCover` function that takes a display pixel ratio, display bounding box, display zoom range, source type, source tile size, and `SourceInfo`, and returns a `std::vector<Resource>`, where the `Resources` are `Resource::Kind::Tile` with the appropriate `TileData`. (Or I might extract `Resource::TileData` to a non-nested class and return a vector of that.)

The `tileCover` function will:

* Determine the appropriate pixel ratio for the tile requests, given the display pixel ratio and supported pixel ratios of the source (currently 1.0 and 2.0 only for Mapbox sources)
* Determine the set of appropriate source zoom levels for the tile requests, given the display zoom range and taking into account
  * source min and max zoom 
  * tile size normalization (i.e. requesting z1 256px tiles at display zoom level 0)
  * different raster/vector float z ⇢ integer z rounding behavior
* Determine the appropriate set of tile `[x, y]`s for each zoom level

`Source` will use this function to calculate the covering tiles for the current bounding box and zoom range (a degenerate range where min=max). Offline will use this function to calculate the appropriate tiles to download given the region definition.

This function will be responsible for rounding to the nearest integer pixel when that's all the source supports. It should be the only place we need to change if we add support for additional pixel ratios  in the future (`@3x` or even non-integer ratios).

The only other place where the type of `TileData::pixelRatio` will matter is in the database schema. There I think I'd like to keep it floating point, just for future flexibility, even though right now the only values will be 1.0 and 2.0.

jfirebaugh: Actually, I'm comfortable saying we'll never support non-integer multipliers for tile resolution. I'll make it an integer here and in the database schema.

include/mbgl/storage/resource.hpp
kkaefer: Good point. Also allows us to write `const auto style = Resource::style(url);`

PR #3742, opened 2016-01-29
Debug mask

1ec5:
This change ports over `MGLMapDebugMaskOptions` from the OS X SDK, allowing you to toggle individual debugging aids. The old debug options have been deprecated and will be removed in a future release.

/ref #3100
/cc @brunoabinader


1ec5:
Squeezed this PR into 3.1.0 to avoid a major version bump according to semver rules. #3100 would’ve removed a public API; this PR restores that public API but marks it as deprecated.

Issue #250, opened 2014-05-26
Support layers with extent != 4096


kkaefer:
We currently hard code the assumption that vertices in a layer span the extent of 0..4096. This may not be true for all tiles, so we should either make our renderer aware of that, or normalize vertices to that extent when decoding layers that specify a different extent.

bsudekum:
I'm seeing a lot of errors coming up from tiles generated in [tippecanoe](https://github.com/mapbox/tippecanoe) as the default setting is not 4096.

@kkaefer has there been any headway on this issue?

kkaefer:
No

Issue #3765, opened 2016-01-29
make sorting symbols by y deterministic


ansis:
The order of symbols with the same y value (after rotation) is undefined, which is causing some differences with -js. https://github.com/mapbox/mapbox-gl-native/blob/0439d003135dbdd15a993efff712e27cd15ab89e/src/mbgl/renderer/symbol_bucket.cpp#L382-L385

ansis:
done 2cfc624

Issue #3764, opened 2016-01-29
fix circle rendering in still mode


ansis:
The circle bucket currently drops all points that are not strictly within the tile boundaries. This works for continuous rendering, but not in still mode. In still mode it should add all points, not just the ones strictly in the tile boundaries.

PR #3766, opened 2016-01-29
[core] support tiles with non-4096 extents

ansis:
Convert all geometries to the maximum extent supported by our buffers and then use that constant extent everywhere else.

ports 1da1e5ab2bd10ea4d427d4f8dacd77e54f47e0b9 from -js

fixes https://github.com/mapbox/mapbox-gl-native/issues/250

Only the line extent render test is re-enabled. The rest of the extent render tests have other differences causing them to still fail. I'll work on these separately: https://github.com/mapbox/mapbox-gl-native/issues/3764 https://github.com/mapbox/mapbox-gl-native/issues/3765

:eyes: @jfirebaugh

### DIFF
--- a/include/mbgl/util/constants.hpp
+++ b/include/mbgl/util/constants.hpp
@@ -10,6 +10,7 @@ namespace mbgl {
 namespace util {
 
 extern const float tileSize;
+extern const int32_t EXTENT;
 
 extern const double DEG2RAD;
 extern const double RAD2DEG;
--- a/package.json
+++ b/package.json
@@ -26,7 +26,7 @@
   ],
   "devDependencies": {
     "aws-sdk": "^2.2.21",
-    "mapbox-gl-test-suite": "mapbox/mapbox-gl-test-suite#7a17d43bd8482a01dc165de6fff6ae4c33c4fc5d",
+    "mapbox-gl-test-suite": "mapbox/mapbox-gl-test-suite#67fe48d058c23dbb4a8ac36ce0270185da3045a8",
     "node-gyp": "^3.2.1",
     "request": "^2.67.0",
     "tape": "^4.2.2"
--- a/src/mbgl/annotation/annotation_tile.hpp
+++ b/src/mbgl/annotation/annotation_tile.hpp
@@ -17,6 +17,7 @@ public:
     FeatureType getType() const override { return type; }
     optional<Value> getValue(const std::string&) const override;
     GeometryCollection getGeometries() const override { return geometries; }
+    uint32_t getExtent() const override { return 4096; }
 
     const FeatureType type;
     const std::unordered_map<std::string, std::string> properties;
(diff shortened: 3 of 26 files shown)


jfirebaugh:
Are you going to port https://github.com/mapbox/mapbox-gl-js/commit/282a12cfcc38784e6947e686ed782383f7930480? It seems like a good idea.

bsudekum:
:bow:

ansis:
@jfirebaugh I think it should be ported separately, later. -native doesn't have an equivalent to `Tile` that it passes to draw functions. `pixelsToTileUnits` could be added to TileID in -native or somewhere else, but I think it should be part of a separate push to make -native and -js Tile/TileID/TileCoord objects more similar. https://github.com/mapbox/mapbox-gl-native/issues/3768

mb12:
@ansis Can you please kindly clarify the following?

1.) Since all coordinates are normalized to 13 bit values, is it correct to say that using tiles with extent > 13 bits is not advisable since the higher precision won't be used by the client for rendering?

2.) Under what scenarios should a source return tiles with extent  <  8192? Does a lower tile extent reduce size of vector tiles? Is there a specific end user case/bug that this change addresses? I also looked at the gl-js link added by @jfirebaugh but couldn't figure it out.

3.) Should the default value be changed to 8192 here as well?

https://github.com/mapbox/vector-tile-spec/blob/master/2.1/vector_tile.proto

4.) Can you please also confirm that a source can return tiles with variable extents and the client would render them correctly? For e.g. return tiles with extent 512 for zoom less than maxzoom and 8192 for maxzoom.

ansis:
@mb12 sure,

1) Yes, since the client won't use the higher precision there's no point in using tiles with extent > 13 bits (yet). We might be able to bump it up to 16384 or 32768 in the future (see https://github.com/mapbox/mapbox-gl-js/pull/2010#issue-129332346)

2) The main reason for implementing support here is that the spec supports non-4096 extents and there are existing tilesets with non-4096 extents (for example, those produced by https://github.com/mapbox/tippecanoe). Yes, I think smaller extents do reduce the size of vector tiles.

3) Nope. The reasons for 4096 being the default (compromise between precision and size) don't change because of this. It's completely fine to have tiles with extents smaller than the one supported internally.

4) Yes, different tiles can have different extents. A single tile could even have layers with different extents.

jfirebaugh:
Are we at all performance sensitive about the floating point operations in `getGeometries`? That's the only thing I would be concerned about.

ansis:
The extra work done by `getGeometries` takes around 1.5% of LineBucket creation time and about 0.5% of FillBucket creation time. Switching to integers speeds it up by about a third (1%, 0.3%) but breaks tiles with extent > util::EXTENT and all theoretical extents that are powers of two. I'm not sure how precise this comparison is because the proportion is different when different data is loaded.

This seems ok, right?

PR #3782, opened 2016-02-01
[tests] Refactor and make MockFileSource more general

jfirebaugh:
Now it works more like StubStyleObserver: you can assign std::functions to specific slots based on resource type.

Rewrite resource loading tests in that style, making them less like integration tests of Style and more like unit tests of Source, GlyphStore, and SpriteStore.

:eyes:  @brunoabinader


tmpsantos:
:+1:

tmpsantos:
I'm sad to see the `PendingResources` go. This one caught some many bugs like crash/blocking on style switch or creation/destruction of the Map object (that happens a lot on the node port).

The "LoadingCancel" test might offer similar functionality, but I think we should still do it at the `mbgl::Map` level too.

PR #3787, opened 2016-02-02
Show annotations with custom edge padding

1ec5:
This PR rebases #3674 atop master for fast-forward merging. Along the way, I revised @jormon’s change to fix a syntax error that crept in and also performed a bit of copyediting.


jormon:
:heart:

PR #3788, opened 2016-02-02
Add custom @2x handling for mapbox v4 tile API when tileSize: 512

yhahn:
Native port of https://github.com/mapbox/mapbox-gl-js/pull/2044.

cc @jfirebaugh @kkaefer @brunoabinader

### DIFF
--- a/src/mbgl/util/mapbox.cpp
+++ b/src/mbgl/util/mapbox.cpp
@@ -119,7 +119,7 @@ std::string normalizeTileURL(const std::string& url, const std::string& accessTo
     return baseURL + "v4/" + url.substr(sizeof("mapbox://tiles/") - 1) + "?access_token=" + accessToken;
 }
 
-std::string canonicalizeTileURL(const std::string& url, SourceType type) {
+std::string canonicalizeTileURL(const std::string& url, SourceType type, uint16_t tileSize) {
     auto tilesetStartIdx = url.find("/v4/");
     if (tilesetStartIdx == std::string::npos) {
         return url;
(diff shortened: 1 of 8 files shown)


src/mbgl/util/mapbox.cpp
```diff
@@ -135,7 +135,19 @@ std::string normalizeRasterTileURL(const std::string& url) {
...
-    normalizedURL.insert(extensionIdx, "{ratio}");
+
+    // Mapbox raster sources always use the @2x suffix on the v4 tile API
+    // to ensure a maximum 512 image size.
+    std::string::size_type versionIdx = url.rfind("/v4/");
+    if (tileSize == 512 &&
+        versionIdx != std::string::npos &&
+        (normalizedURL.compare(extensionIdx + 1, 3, "png") == 0 ||
+        normalizedURL.compare(extensionIdx + 1, 3, "jpg") == 0 ||
+        normalizedURL.compare(extensionIdx + 1, 4, "webp") == 0)) {
```
kkaefer: A better comparison would be add this function to `string.hpp`:

```cpp
inline bool beginsWith(const std::string& string, const std::string &prefix) {
    return std::equal(prefix.begin(), prefix.end(), string.begin());
}
```

jfirebaugh:
Can we hold this until after #3715 lands? There's going to be significant conflicts.

brunoabinader:
> Can we hold this until after #3715 lands?

:+1:

jfirebaugh:
Updated for today's new offline world. @yhahn @tmpsantos Look good?

tmpsantos:
> Updated for today's new offline world. @yhahn @tmpsantos Look good?

:+1:

PR #3803, opened 2016-02-03
Improve LatLngBounds

jfirebaugh:
Replace:
* `LatLngBounds()` (constructor)
* `LatLngBounds::getExtendable()` (static)
* `LatLngBounds(sw, ne)` (constructor)

With:
* `LatLngBounds::world()` (static)
* `LatLngBounds::empty()` (static)
* `LatLngBounds::fromCorners(a, b)` (static, swaps coordinates as necessary)

### DIFF
--- a/include/mbgl/util/geo.hpp
+++ b/include/mbgl/util/geo.hpp
@@ -44,21 +44,47 @@ public:
 
 class LatLngBounds {
 public:
-    LatLng sw = {-90, -180};
-    LatLng ne = {90, 180};
+    // Return a bounds covering the entire (unwrapped) world.
+    static LatLngBounds world() {
+        return LatLngBounds({-90, -180}, {90, 180});
+    }
 
-    LatLngBounds() {}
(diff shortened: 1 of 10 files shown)


include/mbgl/util/geo.hpp
```diff
@@ -44,21 +44,39 @@ class ProjectedMeters {
...
+    static LatLngBounds empty() {
+        LatLngBounds bounds = world();
+        std::swap(bounds.sw, bounds.ne);
+        return bounds;
+    }
 
-    static LatLngBounds getExtendable() {
-        LatLngBounds bounds;
-        return { bounds.ne, bounds.sw };
+    static LatLngBounds hull(const LatLng& a, const LatLng& b) {
```
1ec5: The naming of this method isn’t incredibly obvious. Could you add a documentation comment so its purpose is clear?

1ec5:
The new cardinal direction getters for `LatLngBounds` are lovely. :+1:

PR #3805, opened 2016-02-03
[core] Make bool conversions explicit

jfirebaugh:
Implicit bool conversions are bad; they'll be used e.g. for a == b and a != b if those operators are not defined. This was happening at https://github.com/mapbox/mapbox-gl-native/blob/032c8fba3c8e3c122dd399b5c9341d92ad9d286f/src/mbgl/map/transform.cpp#L132-L132, for example.


1ec5:
:+1:

PR #3809, opened 2016-02-03
Fix designable warning when showing user location

1ec5:
Turn the Shows User Location inspectable into a no-op when running in the designables agent. Otherwise, we complain that the agent lacks Location Services permissions:

```
file:///path/to/TestApplication/TestApplication/Storyboard.storyboard: warning: IB Designables: Ignoring user defined runtime attribute for key path "showsUserLocation" on instance of "TestApplication.MapView". Hit an exception when attempting to set its value: In iOS 8 and above, this app must have a value for NSLocationWhenInUseUsageDescription or NSLocationAlwaysUsageDescription in its Info.plist.
```

/cc @friedbunny


friedbunny:
Thanks. 😁

Issue #3529, opened 2016-01-13
make TileID more similar to -js's TileCoord


ansis:
- `x` values are wrapped by worlds in -js but not in -native
- `w` value doesn't make sense in -native
- names are different

jfirebaugh:
Regarding tile wrapping, the changes I would like to see are:

* Two different types: `TileID` (wrapped x coordinate; prefer this type unless you specifically need an unwrapped ID) and `UnwrappedTileID` (unwrapped x coordinate + `w`)
* `UnwrappedTileID` has `TileID wrap() const` method; no other conversions are available
* `tileCover` returns `UnwrappedTileID`s
* Remove `Source::getLoadedTiles()` and `Source::getTiles()`
* Add `Source::getCoveringUnwrappedTileIDs()`, `Source::getTile(const TileID&)` and _maybe_ `Source::getTile(const UnwrappedTileID&)`
* Change `TransformState::matrixFor` parameter to `UnwrappedTileID`
* Fix ripples from those changes

jfirebaugh:
The names `TileCoordinate` and `UnwrappedTileCoordinate` would be ok with me too, but we'll need to clear away the existing `TileCoordinate` type first.

jfirebaugh:
⇢ #3834

PR #3810, opened 2016-02-03
[core] Refactor tileCover

jfirebaugh:
:eyes: @ansis


test/util/tile_cover.cpp
```diff
@@ -0,0 +1,118 @@
...
+            TileID(1, 1, 0, 1),
+            TileID(1, 0, 0, 1)
+        }}),
+        (set {
+            result.begin(),
+            result.end()
+        }));
+}
+
+//TEST(TileCover, SingletonZ0) {
```
jfirebaugh: `tileCover(LatLngBounds::singleton(...))` is always empty; IMO this is a bug in `tileCover` that should be fixed. It should return the containing tile.

test/util/tile_cover.cpp
```diff
@@ -0,0 +1,118 @@
...
+            TileID(10, 163, 396, 10),
+            TileID(10, 164, 396, 10)
+        }}),
+        (set {
+            result.begin(),
+            result.end()
+        }));
+}
+
+//TEST(TileCover, OrderedByDistanceToCenter) {
```
jfirebaugh: Our sort appears to be buggy; it's probably not sorting by exactly |(bounds center - tile center)|.

test/util/tile_cover.cpp
```diff
@@ -0,0 +1,118 @@
...
+//    ASSERT_EQ(  12, result[1].z);
+//    ASSERT_EQ( 655, result[1].x);
+//    ASSERT_EQ(1583, result[1].y);
+//}
+//
+//static const LatLngBounds sanFranciscoWrapped = LatLngBounds::hull(
+//    { 37.6609, 238.5744 },
+//    { 37.8271, 238.3204 });
+//
+//TEST(TileCover, SanFranciscoZ0Wrapped) {
```
jfirebaugh: See https://github.com/mapbox/mapbox-gl-native/issues/3529#issuecomment-179486390 for proposed wrapping behavior.

jfirebaugh:
Included in #3715.

PR #3813, opened 2016-02-04
[ios] Explain prerelease testing in docs

jormon:
Show how to point a podspec at an officially tagged pre-release.


1ec5:
Thanks, we definitely do need to document this workflow. However, DEVELOPING.md is intended for people who are comfortable with building the SDK from source. Perhaps INSTALL.md or README.md would be a better place for this information.

jormon:
deprecated in favor of #3828

PR #3823, opened 2016-02-04
Remove code duplication in gl_helper.hpp

kkaefer:
In `gl_helper.hpp`, we're retrieving and storing GL state, which is encapsulated in `gl_config.hpp` as well. Instead of duplicating the code, we should just reuse it there.


jfirebaugh:
⇢ @brunoabinader

PR #3790, opened 2016-02-04
Fix clipping mask rendering

kkaefer:
We're currently not painting clipping masks correctly in a few circumstances:

- Tiles are still being loaded
- Multiple sources use the same tile IDs

### DIFF
--- a/include/mbgl/platform/default/glfw_view.hpp
+++ b/include/mbgl/platform/default/glfw_view.hpp
@@ -52,6 +52,9 @@ private:
     makeSpriteImage(int width, int height, float pixelRatio);
 
     void nextOrientation();
+    void toggleClipMasks();
+
+    void renderClipMasks();
 
     void addRandomPointAnnotations(int count);
     void addRandomShapeAnnotations(int count);
@@ -80,6 +83,8 @@ private:
     int fbWidth;
     int fbHeight;
     float pixelRatio;
+
+    bool showClipMasks = false;
 
     double lastX = 0, lastY = 0;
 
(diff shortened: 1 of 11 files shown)


kkaefer:
:eyes: @jfirebaugh @ansis

ansis:
> Multiple sources use the same tile IDs

Can you explain this part more? I'm having trouble understanding what's going on from just the code

kkaefer:
The code changes the way stencil masks are painted to the stencil buffer. Previously, we'd loop through all sources, and for every active tile in that source, we'd draw the stencil mask. This was inefficient, since we always have at least two sources active, the regular tile source, and the auto-generated annotations source. In non-overzooming cases, these contained the same tile IDs, which lead to the stencil mask being painted twice. Instead, we're now combining all tiles that we need to create a mask for, and merge in all overlapping parents from other sources, so we only have to paint a tile once.

In addition to the performance improvement, this PR also fixes a rendering issue with multiple sources (which is always the case!): When the second source has tiles that are parents of tiles in the first source, that tile isn't properly clipped:

Source 1:
* 1/0/0, produces ref `00000001` with mask `00000001`

Source 2:
* 0/0/0, produces ref `00000010` with mask `00000010`
* 1/0/0, reuses ref `00000001` with mask `00000001` from Source 1

In this case, `0/0/0` of the second source isn't properly clipped by `1/0/0`, because it only looks at at the value of the second to last bit. Instead, the `mask needs to be expanded to `00000011`, so that pixels where the stencil mask is `00000011` (superposition of `0/0/0 and 1/0/0`) aren't shaded when painting `0/0/0`.

Here's a visualization from actual screencaps:

before

before

ansis:
@kkaefer thanks for the explanation! looks good to me

PR #3828, opened 2016-02-04
[ios] Explain prerelease testing in docs

jormon:
Show how to point a podspec at an officially tagged pre-release.


1ec5:
Thanks!

PR #3831, opened 2016-02-05
[android] #3830 added mocked parcel implementation

### LINKED ISSUE #3830: Add mocked Parcel implementation for unit tests
Unit testing on Android is considered to be difficult because unit tests for Android are run on the JVM of your local development machine. These tests are run against the `android.jar` that doesn't contain any actual implementations. It just throws an exception when called. To bypass this we need to mock some parts of the Android SDK. In this issue I will try to setup a mock for the Parcelable interface. I will base my solution from [this](https://gist.github.com/Sloy/d59a36e6c51214d0b131).

The reason why I want this, is that:
 - we are sure our objects are recreated correctly
 - we get better insights in code coverage, currently Parceable obfuscate our results.

tobrun:
closes #3830


tobrun:

Issue #3185, opened 2015-12-04
If there are too many annotation images, some annotation images are invisible


skajake:
My use case:  
I have 19,000 annotations spread across the US.
Each annotation has a unique image.
I only show 20 annotations at a time as the user pans.

Problem:
After panning for awhile and loading a few hundred annotations (not all at once), it fails to show any more.

Here is how I remove and add annotations as the user pans:

``` objc
-(void)mapView:(MGLMapView *)mapView regionDidChangeAnimated:(BOOL)animated {
    //This method only returns 20 objects
    NSArray *myObjects = [dummyService getObjectsInVisibleRect:[mapView visibleCoordinateBounds].sw ne:[mapView visibleCoordinateBounds].ne];
    NSMutableArray *newAnnotations = [[NSMutableArray alloc] init];
    for(MyObject *myObject in myObjects) {
        MGLPointAnnotation *point = [[MGLPointAnnotation alloc] init];
        point.coordinate = [[myObject location] coordinate];
        point.title = [myObject myIdentifier];
        point.subtitle = [myObject myName];
        [newAnnotations addObject:point];
    }
    NSArray *existingAnnotations = [_mapView annotations];
    NSMutableArray *annotationsToRemove = [[NSMutableArray alloc] init];
    for(MGLPointAnnotation *existingAnnotation in existingAnnotations) {
        //I actually subclass MGLPointAnnotation, such that isEqual and hash are implemented.  In any case, the containsObject logic here works as intended
        if([newAnnotations containsObject:existingAnnotation]) {
            [newAnnotations removeObject:existingAnnotation];
        } else {
            [annotationsToRemove addObject:existingAnnotation];
        }
    }
    if([annotationsToRemove count] > 0) {
        [_mapView removeAnnotations:annotationsToRemove];
    }
    if([newAnnotations count] > 0) {
        [_mapView addAnnotations:newAnnotations];
    }
}
```

To be clear, the imageForAnnotation method fires, and a valid MGLAnnotationImage (inited with UIImage and a unique reuseIdentifier) is returned.  However, the annotations stop showing up on the map.

1ec5:
> I actually subclass MGLPointAnnotation, such that isEqual and hash are implemented.

Oh, good catch. Please feel free to submit a pull request with that functionality.

[...]

PR #3835, opened 2016-02-05
Update annotations and annotation images dynamically

### LINKED ISSUE #1980: Updating "MGLPointAnnotation" coordinate does not change annotation placement on a map
Mapbox GL 0.5.1

1ec5:
MGLMapView observes changes to the `coordinate` property of each MGLAnnotation added to it. Changing the `coordinate` property in a KVO-compliant way causes the annotation to be relocated and its callout view, if present, to be dismissed. To avoid observing the same annotation twice yet also avoid expensive lookups when adding or removing annotations, MGLMapView indexes added point annotations in an NSMutableSet.

Added the ability to delete an unused annotation image’s images on iOS. (The OS X SDK doesn’t yet support changing an annotation image’s image.) When you nil out the image of an MGLAnnotationImage, MGLMapView deletes the sprite from the style and resets any annotation associated with the MGLAnnotationImage instance to use the default red pin icon under the hood. The annotation remains associated with the MGLAnnotationImage, which comes in handy in the event that the annotation image’s image becomes non-nil.

In iosapp, tapping a callout view moves the selected annotation to the center of the screen and deselects it. Deselecting an annotation resets its image to the default; deselecting it again restores the image.

Fixes #1980.

/ref #3185
/cc @boundsj @friedbunny

### DIFF
--- a/platform/ios/src/MGLMapView.mm
+++ b/platform/ios/src/MGLMapView.mm
@@ -135,9 +135,8 @@ mbgl::Color MGLColorObjectFromUIColor(UIColor *color)
 class MGLAnnotationContext {
 public:
     id <MGLAnnotation> annotation;
-    /// mbgl-given identifier for the annotation image used by this annotation.
-    /// Based on the annotation image’s reusable identifier.
-    NSString *symbolIdentifier;
+    /// The annotation’s image’s reuse identifier.
+    NSString *imageReuseIdentifier;
 };
 
 #pragma mark - Private -
@@ -2371,11 +2370,17 @@ mbgl::Duration MGLDurationInSeconds(NSTimeInterval duration)
 
 - (void)addAnnotations:(NS_ARRAY_OF(id <MGLAnnotation>) *)annotations
 {
+    [self addAnnotations:annotations withAnnotationImage:nil];
+}
+
+- (void)addAnnotations:(NS_ARRAY_OF(id <MGLAnnotation>) *)annotations withAnnotationImage:(MGLAnnotationImage *)explicitAnnotationImage
+{
     if ( ! annotations) return;
     [self willChangeValueForKey:@"annotations"];
 
     std::vector<mbgl::PointAnnotation> points;
     std::vector<mbgl::ShapeAnnotation> shapes;
+    NSMutableArray *annotationImages = [NSMutableArray arrayWithCapacity:annotations.count];
 
     BOOL delegateImplementsImageForPoint = [self.delegate respondsToSelector:@selector(mapView:imageForAnnotation:)];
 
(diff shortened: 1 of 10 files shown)


ios/app/MBXViewController.mm
```diff
@@ -169,6 +169,7 @@ - (void)showSettings
                                                                 @"Add Test Shapes",
                                                                 @"Start World Tour",
                                                                 @"Add Custom Callout Point",
+                                                                @"Reset Annotation Images",
```
friedbunny: I'm not seeing a new handler for this action.

1ec5: Oops, I didn’t end up needing that action. I’ll remove it.

1ec5:
Implementing the annotation replacement in mbgl, as in #3885, would probably be more efficient.

1ec5:
Moving to 3.3.0 milestone in light of #4556.

1ec5:
I added support for atomically relocating an annotation without having to replace the annotation outright. The PR description has been updated.

platform/ios/src/MGLMapView.mm
```diff
@@ -2426,9 +2450,12 @@ - (void)addAnnotations:(NS_ARRAY_OF(id <MGLAnnotation>) *)annotations
         
         for (size_t i = 0; i < pointAnnotationTags.size(); ++i)
         {
+            MGLAnnotationImage *annotationImage = annotationImages[i];
+            annotationImage.styleIconIdentifier = @(points[i].icon.c_str());
+            
             MGLAnnotationContext context;
             context.annotation = annotations[i];
-            context.symbolIdentifier = @(points[i].icon.c_str());
+            context.imageReuseIdentifier = annotationImage.reuseIdentifier;
```
boundsj: Great! I made a similar refactor on my annotation branch. This is better and includes the new and better property name on MGLAnnotationContext

platform/ios/src/MGLMapView.mm
```diff
@@ -2995,11 +3047,55 @@ - (void)showAnnotations:(NS_ARRAY_OF(id <MGLAnnotation>) *)annotations edgePaddi
...
+    {
+        // Add the default icon to the style if necessary.
+        annotationImage.styleIconIdentifier = fallbackIconIdentifier;
+        if ( ! [self dequeueReusableAnnotationImageWithIdentifier:MGLDefaultStyleMarkerSymbolName])
+        {
+            [self installAnnotationImage:self.defaultAnnotationImage];
+        }
+        
+        // Update any annotations associated with the annotation image.
+        for (auto &pair : _annotationContextsByAnnotationTag)
```
boundsj: nit: some repeated code here, could be method with "icon identifier" parameter.

boundsj:
👍 this looks great! I will base the change I'm about to make for a custom view off of this branch until it lands in master.

platform/ios/src/MGLMapView.mm
1ec5: Fixed in 3976448479d85cab2f0554995acf92fa0403deda.

t-yoshii:
Hi, 
Does this feature cover MGLPolyLine?

Removing MGLPolyline object from MGLMapView causes app crash.

	override func viewDidLoad() {
		super.viewDidLoad()

		mapView = MGLMapView(frame: view.bounds)
		mapView.autoresizingMask = [.FlexibleWidth, .FlexibleHeight]

		var coordinates: [CLLocationCoordinate2D] = [CLLocationCoordinate2DMake(40.7326808, -73.9843407), CLLocationCoordinate2DMake(40.7326808, -73.9943407)]
		polyline = MGLPolyline(coordinates: &coordinates, count: 2)

		// set the map's center coordinate
		mapView.setCenterCoordinate(CLLocationCoordinate2D(latitude: 40.7326808,
			longitude: -73.9843407),
			zoomLevel: 12, animated: false)

		mapView.addAnnotation(polyline!)

		mapView.removeAnnotation(polyline!)  /// CRASH HAPPENS HERE!
		view.addSubview(mapView)
	}

Exception detail:

2016-04-20 18:11:23.932 MapboxTest[364:56042] *** Terminating app due to uncaught exception 'NSRangeException', reason: 'Cannot remove an observer <MGLMapView 0x15ee754d0> for the key path "coordinate" from <MGLPolyline 0x15ee602a0> because it is not registered as an observer.'

1ec5:
@t-yoshii, good catch. It still isn’t possible to change a polyline or polygon’s location or form on the fly. A fix for this crash is in #4766.

t-yoshii:
@1ec5 Thank you for your reply! I will try the version which includes #4766.

incanus:
@1ec5 I realized I undid your deselection behavior over in https://github.com/mapbox/mapbox-gl-native/pull/5786 (see opening description). How important is that to bring back?

1ec5:
Here’s what this PR implemented originally:

> In iosapp, tapping a callout view moves the selected annotation to the center of the screen and deselects it. Deselecting an annotation resets its image to the default; deselecting it again restores the image.

We should have something to exercise both pieces of functionality, but feel free to find a better way to exercise them.

PR #3840, opened 2016-02-06
[node] add debug options to node bindings

### LINKED ISSUE #3579: Add debug options to node bindings
Add a `debug` option to [`map.render`](https://github.com/mapbox/mapbox-gl-native/tree/master/platform/node#rendering-a-map-tile) with the following sub-options, corresponding to [`MapDebugOptions`](https://github.com/mapbox/mapbox-gl-native/blob/a88e58ef3a90225fcfafca1efab996de6a827b72/include/mbgl/map/mode.hpp#L31-L37):

```
debug: {
  tileBorders: true/false,
  parseStatus: true/false,
  timestamps: true/false,
  collision: true/false
}
```

cc @mikemorris @bsudekum

ansis:
fix #3579 

:eyes: @jfirebaugh @mikemorris

### DIFF
--- a/platform/node/src/node_map.cpp
+++ b/platform/node/src/node_map.cpp
@@ -25,6 +25,7 @@ struct NodeMap::RenderOptions {
     unsigned int width = 512;
     unsigned int height = 512;
     std::vector<std::string> classes;
+    mbgl::MapDebugOptions debugOptions = mbgl::MapDebugOptions::NoDebug;
 };
 
 ////////////////////////////////////////////////////////////////////////////////////////////////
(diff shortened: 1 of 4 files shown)


platform/node/src/node_map.cpp
```diff
@@ -25,6 +25,15 @@ struct NodeMap::RenderOptions {
     unsigned int width = 512;
     unsigned int height = 512;
     std::vector<std::string> classes;
+    bool debugTileBorders = false;
```
jfirebaugh: Why not use `mbgl::MapDebugOptions` here?

ansis: To `mbgl` stuff out of `NodeMap::ParseOptions`. Should I switch it?

I replaced these three bools with the nested debug struct and forgot to remove them.

mikemorris: Pulling in `mbgl::MapDebugOptions` here seems :+1: to me.

ansis: switched to using `mbgl::MapDebugOptions` 1934258

platform/ios/vendor/SMCalloutView
```diff
@@ -1 +1 @@
-Subproject commit 2aede5d8d1577101bf18405246220e7a710df607
+Subproject commit 717078fb6b0d02b944a12f4d40fc15567489d23b
```
jfirebaugh: Accidental submodule update?

ansis: yeah, thanks for catching that

jfirebaugh:
:+1:

PR #3848, opened 2016-02-08
[android] Fixed error with asset URLs with URL encoded spaces.

jonbe242:
Found bug in android version of asset_file_source.cpp not decoding URL encoded spaces.


jfirebaugh:
Thanks! Committed in 040fc09ea791921b1b86d96b20989d6735c7c3af.

Issue #3597, opened 2016-01-18
Add per HTTP request callback


jonbe242:
For both Android and iOS it would be nice to be able to pre process the HTTP request (okhttp3.Request  and NSMutableURLRequest) before the request is sent to the webserver to be able to for example add custom headers required for your own map sources.

While at it, I think it would be a good idea to be able to construct the OkHttpClient (Android) and the NSURLSession (iOS) from custom code for similar reasons.

jonbe242:
Looking at it I might give this a try on the Android side (at least for now). But I'd like feedback on the preferred way to do this.

Either
- I do as suggested above and just expose a setter for the OkHttpClient in HTTPContext (easy) or
- hide the fact that OkHttpClient is used in the public API and abstracts how HTTPRequest.start and .cancel is implemented and moves current implementation into a default implementation that a user of the SDK can choose to replace.

Any other suggestions?

tobrun:
@1ec5 @boundsj @friedbunny 
What are your thoughts on this for iOS?
Related Android discussion can be found in #3849

friedbunny:
This would best be handled with a delegate method on iOS and not by directly exposing `NSURLSession`, as @boundsj suggested in chat.

I'm not convinced of the need for this, however. We haven't provided such low-level networking hooks in the past and the current level of demand wouldn't appear to warrant expending much/any effort on supporting this feature.

jonbe242:
@jfirebaugh I'm wondering in what direction your internal discussion went regarding an interceptor API.
Are we talking about an interceptor that can only manipulate e.g. headers or an interceptor that can replace the entire HTTP client implementation (my second suggestion/bullet above)?

jonbe242:
Depending on the wanted direction we might be interested in helping out on this.

alexmcroberts:
Has any further consideration been given to this?

1ec5:
> for example add custom headers required for your own map sources

I can see this being useful if you’re maintaining your own tile server or hosting your own styles. Unfortunately, the Mapbox team is unable to go out of our way to focus on these relatively minor use cases.

[...]

PR #3849, opened 2016-02-08
[android] Allow a custom configured OkHttpClient to be injected…

jonbe242:
… into the HTTPContext.

This should resolve  issue #3597 for Android according my option no 1.

### DIFF
--- a/platform/android/MapboxGLAndroidSDK/src/main/java/com/mapbox/mapboxsdk/http/HTTPContext.java
+++ b/platform/android/MapboxGLAndroidSDK/src/main/java/com/mapbox/mapboxsdk/http/HTTPContext.java
@@ -20,7 +20,7 @@ import okhttp3.OkHttpClient;
 import okhttp3.Request;
 import okhttp3.Response;
 
-class HTTPContext {
+public class HTTPContext {
 
     private static final int CONNECTION_ERROR = 0;
     private static final int TEMPORARY_ERROR = 1;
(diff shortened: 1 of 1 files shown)


jfirebaugh:
Given that we've switched back and forth between OkHttp and other implementations a couple times, I'm not sure that we want to commit to exposing an OkHttp dependency in the public API. @tobrun what do you think?

jonbe242:
I can sympathize with you not wanting to expose OkHttp like that. But then I'd like you to consider my option no 2 in #3597.

An easy way for any SDK user to monitor and adjust HTTP requests is a must have in my opinion. Going back to a native solution would make that much harder, so I don't want that.

tobrun:
@jonbe242 first of all, thank you for reaching out and contributing to this repo.

As @jfirebaugh noted, this would indicate that we are committing to always expose a configure the `http client library`-API in future releases. Moving back to a native would make this very hard since we have to configure the client through JNI. I understand the need for this when you combine that with the requirement of having an own tile server, but I'm not sure how this translates to our own requirements and goals.

Before going further with this, I'm going to ping iOS developers in the original issue #3597.
Also pinging @bleege and @zugaldia for :eyes: and roadmap related remarks.

zugaldia:
@jonbe242 @jfirebaugh @tobrun Though I don't see ourselves going away from OkHttp any time soon (we also opted for it in our directions and geocoder libraries) I agree we don't necessarily want to expose the OkHttp client directly.

If the use case is to be able to attach a custom interceptor (that will allow setting headers, or log requests) I'd prefer to expose a generic API to add and remove interceptors ([the interface](https://github.com/square/okhttp/blob/master/okhttp/src/main/java/okhttp3/Interceptor.java) is here).

jfirebaugh:
Ok, per the discussion here, closing in favor of #3597, where further discussion can happen about exposing an interceptor API that's not tied to a particular implementation.

PR #3872, opened 2016-02-10
Set dynamic framework minimum deployment target to iOS 8.0

### LINKED ISSUE #3870: Mapbox 3.1 dynamic framework has minimum deployment target of iOS 9.3; should be 8.0
When trying to archive for TestFlight today after updating MapBox, I received a validation warning when submitting to the App Store. Looks like MapBox is referencing 9.3 in the target? Was the dynamic framework you're referencing in the podspec created using the Xcode beta?

screen shot 2016-02-09 at 5 05 10 pm

1ec5:
Fixes #3870.

/cc @boundsj @friedbunny @parrots

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -51,6 +51,7 @@ Known issues:
 
 ## iOS 3.1.1
 
+- Corrected the dynamic framework’s minimum deployment target to iOS 8.0. ([#3872](https://github.com/mapbox/mapbox-gl-native/pull/3872))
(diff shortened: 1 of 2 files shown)


mb12:
@1ec5  Is Mapbox dropping support for iOS 7.x?

1ec5:
No, but iOS 7.x doesn't support dynamic frameworks. This PR affects only the dynamic framework; the static framework continues to have a minimum deployment target of iOS 7.0.

Issue #3785, opened 2016-02-02
too many coordinates / bounds classes


jfirebaugh:
We represent coordinates using no less than 8 different types in different parts of the code:

* `LatLng`
* `ProjectedMeters`
* `PrecisionPoint`
* `TileCoordinate`
* `TileID`
* `std::array<float, 2>`
* `vec2`
* `Coordinate`

This is three more than we should have. We should have:

* One type for spherical geographic coordinates (`LatLng`)
* One type for projected cartesian coordinates (`ProjectedMeters`, but the naming is off -- this class should be unit-less)
* One type for screen pixel coordinates (`PrecisionPoint`, but should be renamed to something like `ScreenCoordinate`)
* One type for tile z/x/y coordinates (`TileID`)
* One type for [normalized](https://github.com/mapbox/mapbox-gl-native/pull/3766) vector tile coordinates (`Coordinate`)

So the ones to eliminate are:

* [ ] `std::array<float, 2>`
* [ ] `vec2`
* [ ] `TileCoordinate`

For bounds, we have:

* `LatLngBounds`
* `MetersBounds`
* `box` (I'm in the process of removing this.)

jfirebaugh:
We also have at least three duplicated and inconsistent copies of projection math:

* In `projection.hpp`
* In `transform_state.cpp`
* In `geo.cpp`

1ec5:
The projection math in TransformState is often dependent on the current scale, but it would be great to factor some of that to be based on Projection, which is more readily available to code at the SDK level. That would hep to eliminate some of the additional redundancy at the SDK level.

brunoabinader:
Good point @jfirebaugh :+1: 

I believe we shouldn't use `vec2<T>` directly (we could enforce this by e.g. enforcing a private ctor). Instead, we'd use meaningful aliases like `PrecisionPoint`. It seems to me that `Coordinate` and `PrecisionPoint` are correlates, so we might get rid of `Coordinate`. From what I see, `std::array<float, 2>` can also be easily transformed into a `PrecisionPoint`. In sum, we'd get rid of `vec2<T>` (directly), `Coordinate` and `std::array<float, 2>`.

PR #3873, opened 2016-02-10
Clean up coordinate types

brunoabinader:
As per #3785, this PR main changes are:
- [x] Replace `TileCoordinate` with `TileID`
- [x] Make `TileID` coordinates use double precision
- [x] Remove duplicated `vec4` implementation
- [x] Renamed `PrecisionPoint` to `ScreenCoordinate`
- [x] Cleanup usage of `vec{2,3,4}` and `std::array<float, 2>` - they are still used in some places though no longer in coordinate-related logic

:eyes: @jfirebaugh @tmpsantos @1ec5 @kkaefer


src/mbgl/map/transform_state.cpp
```diff
@@ -256,42 +255,42 @@ double TransformState::worldSize() const {
     return scale * util::tileSize;
 }
 
-PrecisionPoint TransformState::latLngToPoint(const LatLng& latLng) const {
+ScreenCoordinate TransformState::latLngToPoint(const LatLng& latLng) const {
```
1ec5: This method should be renamed latLngToScreenCoordinate(), right?

brunoabinader: Good catch :+1: I'll update the `TransformState` function names accordingly.

kkaefer:
> Make `TileID` coordinates use double precision

I strongly disagree with this change. These are integer because tile endpoints are integer only. Making them floating point may introduce subtle epsilon bugs when comparing tile IDs.

src/mbgl/map/source_info.hpp
```diff
@@ -16,11 +15,11 @@ class TileID;
...
 public:
     std::vector<std::string> tiles;
-    uint16_t minZoom = 0;
-    uint16_t maxZoom = 22;
+    double minZoom = util::MIN_ZOOM;
+    double maxZoom = util::MAX_ZOOM;
     std::string attribution;
-    std::array<float, 3> center = { { 0, 0, 0 } };
-    std::array<float, 4> bounds = { { -180, -90, 180, 90 } };
+    vec3<> center = { 0, 0, 0 };
```
jfirebaugh:
```
LatLng center;
double zoom;
LatLngBounds bounds;
```

include/mbgl/util/geo.hpp
```diff
@@ -2,14 +2,20 @@
...
 #include <cmath>
+#include <vector>
 
 namespace mbgl {
 
 class TileID;
 
-using PrecisionPoint = vec2<double>;
+// Normalized vector tile coordinate.
+using Coordinate = vec2<int16_t>;
```
jfirebaugh: Can we rename this to `GeometryCoordinate` and move it to `geometry_tile.hpp`?

include/mbgl/util/geo.hpp
```diff
@@ -54,7 +60,8 @@ class LatLngBounds {
 public:
     // Return a bounds covering the entire (unwrapped) world.
     static LatLngBounds world() {
-        return LatLngBounds({-90, -180}, {90, 180});
+        return { { -util::LATITUDE_MAX, -util::LONGITUDE_MAX },
```
jfirebaugh: This is a change in semantics because `LATITUDE_MAX` is 85.05...

src/mbgl/annotation/point_annotation_impl.cpp
```diff
@@ -13,11 +13,11 @@ void PointAnnotationImpl::updateLayer(const TileID& tileID, AnnotationTileLayer&
     featureProperties.emplace("sprite", point.icon.empty() ? std::string("default_marker") : point.icon);
 
     const uint16_t extent = 4096;
-    const mbgl::PrecisionPoint pp = point.position.project();
```
jfirebaugh: This code was odd to begin with and should probably be replaced wholesale with something that makes more sense. It shouldn't round trip through screen coordinates, for one thing. And stuff like `pp.x * z2 - x`, where `x = pp.x * z2`, looks nonsensical.

jfirebaugh:
I agree with @kkaefer, I don't think should replace TileCoordinate with TileID and make TileID coordinates double. Instead, we should reduce the usage of TileCoordinate to just `tileCover` and then inline it there, and some set of TileID changes suggested in #3834 and #3529, but keep it integer based.

brunoabinader:
@jfirebaugh @kkaefer this is ready for review now. This PR is getting too big to fit all changes from https://github.com/mapbox/mapbox-gl-native/issues/3529 so I'd prefer to implement the rest of `TileID` changes in a followup PR.

`TileCoordinate` is now scoped to tile cover implementation. I had some issues while clearing it from `TransformState::moveLatLng` but behavior is kept the same. `LatLng` now can be `wrapped()` when needed. This helped me find a small precision loss issue wrt usage of `TransformState` conversion functions (tests for that are attached).

brunoabinader:
Update: I'm investigating a regression in the test suite expectations, will let you know once I have more information.

kkaefer:
@brunoabinader let's remove the last commit ("Initial setup for UnwrappedTileID") and commit this before it gets unwieldy.

Issue #3775, opened 2016-02-01
[android] Change marker icon without remove/add workaround


mpuchala:
Hi Guys,

I'm using your map with more or less 500 - 1k markers at a time. There are few types of icons. Some of them are just made from resources and cached, but I'm using clustering as well. That’s why many of them have to be created as a new Bitmap.

My issue is about changing marker icon during the runtime. On Github I saw this thread: https://github.com/mapbox/mapbox-gl-native/pull/3146. Does it mean it would be possible to change icon on Android in different way than removing/changing/adding new marker? In fact this solution is problematic  - it can take up to 1000ms and looks awful (tested on Nexus 4, Android 5.1). It blocks UI thread so despite the fact clustering is slow and setting different icons according to zoom doesn’t work smooth it affects another views (ex. animations). Actually even this code can take up to 200ms:

```
mMapView.removeAnnotation(mMarkerOptions.getMarker());
mMarkerOptions.icon(...);
mMapView.addMarker(mMarkerOptions);
```

 Maybe you have any idea how can I improve it?

1ec5:
The iOS SDK supports changing marker images without adding and removing markers (#3146, #3320). [`-[MGLMapView annotationImageNeedsRedisplay:]`](https://github.com/mapbox/mapbox-gl-native/blob/e8b8f5e49306e3cc80c2692575703e49129dcb84/platform/ios/src/MGLMapView.mm#L2989-L2996) should be ported to the Android SDK.

socketbind:
Hi,

I have been actually working on this (+ the ability to set the position also) but my code is a bit crude (see referenced pull request). I would love to improve upon it and make this happen.

PR #3885, opened 2016-02-10
[core] [android] Update marker position and icon

socketbind:
Hi,

This pull request enables updating a marker's position and icon in place without resorting to workarounds like removing and adding the marker. I have implemented this functionality because I needed it for an application that displays neighboring cards, traffic lights with animated marker icons.

As my C++ knowledge is a bit rusty the code might be a little bit crude. I would love to improve it to be more performant (it can get really slow with a lot of markers even on powerful devices) and make it so that it conforms better to your design principles, code quality.

Relates to #3775 (change marker icon without workarounds) but also adds option for changing position.

### DIFF
--- a/include/mbgl/map/map.hpp
+++ b/include/mbgl/map/map.hpp
@@ -161,6 +161,8 @@ public:
     AnnotationID addShapeAnnotation(const ShapeAnnotation&);
     AnnotationIDs addShapeAnnotations(const std::vector<ShapeAnnotation>&);
 
+    void updatePointAnnotation(AnnotationID, const PointAnnotation&);
+
     void removeAnnotation(AnnotationID);
     void removeAnnotations(const AnnotationIDs&);
 
--- a/platform/android/MapboxGLAndroidSDK/src/main/java/com/mapbox/mapboxsdk/annotations/Marker.java
+++ b/platform/android/MapboxGLAndroidSDK/src/main/java/com/mapbox/mapboxsdk/annotations/Marker.java
@@ -61,8 +61,17 @@ public final class Marker extends Annotation {
         return infoWindowShown;
     }
 
-    void setPosition(LatLng position) {
+    /**
+     * Sets the position.
+     *
+     * @param position new position
+     */
+    public void setPosition(LatLng position) {
         this.position = position;
+        MapboxMap map = getMapboxMap();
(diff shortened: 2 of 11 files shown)


tobrun:
@socketbind 
thank you reach out and contributing! 
I'm going review the java code and will add some inline comments.
I will try to track someone down to validate the C++ code.

socketbind:
Hi @tobrun,
Thank you for the review. I have replaced HashMap in MapView with LongSparseArray per your suggestion and also made an attempt to make the C++ code a bit more performant by trying to reduce the number of operations required to update the marker (calling the appropriate AnnotationTileMonitor::update method in annotation_manager.cpp to update and coalescing the marker update calls into a single one when possible). It seems to be working properly but I'm unsure about its correctness (especially with regards to thread safety) because my knowledge is still very limited about the engine and C++14 itself. Could you please help me validate it?

I'm still trying to make things even faster as my application does a lot of marker updates in a short time but I think the code included here could be appropriate for basic functionality (given that it is actually correct and the design is acceptable).

jfirebaugh:
Thanks for the contribution @socketbind. I've merged the core C++ parts of this in #3933 (using your initial implementation rather than the more complex version in 23c357a5b210c8d4a45d32c947009ead1e18a057) and added a test. Can you rebase the Android changes and squash then down to a single commit for @tobrun to do a final review?

socketbind:
Rebased changes as requested (messed up the first rebase on these commits, you might have received some strange emails, sorry about that). 

I have also removed 23c357a as it turned out to be unreliable. Still, I would l love to improve performance somehow. I have been experimenting with batched updates (using a window of 100 ms on the Java side to save on JNI calls and applying changes first then issuing a single update call) but it only helped slightly, `update(Update::Annotations)` seems to be very costly for now.

tobrun:
@socketbind, np, I have fixed the rebase issue and pushed with #3943. For the Android code I'm also going to add some tests in that PR. Will also add an example activity demonstrating this awesome feature.

tobrun:
@socketbind https://github.com/mapbox/mapbox-gl-native/commit/b3953029909cbb6d4d8cc48fc95acf71523f941b landed with https://github.com/mapbox/mapbox-gl-native/pull/3943. Thank you again for contributing!

PR #3887, opened 2016-02-10
Only constrain after adding to a window

### LINKED ISSUE #3868: Configuring latitude value with a storyboard is broken
2016-02-09 at 9 36 am

It doesn't matter what you put in the latitude field, the map is always at the equator

1ec5:
Introduced a setter/getter for constrain mode. On iOS and OS X, the zoom level inspectable causes the zoom level to be set independently from the longitude and latitude. Thus, the latitude inspectable had no effect because the latitude was constrained to 0 at z0. Temporarily removing the heightwise constraint allows the map to center on the intended location before zooming, which is the usual case for storyboards and XIBs. On iOS, the only guarantee we have timing-wise is that all the inspectables are applied after initialization but before the view is added to a window. So we reimpose the heightwise constraint as soon as the view is added to a window, that is, before the user has a chance to pan the map out of bounds.

Fixes #3868.

/cc @friedbunny @brunoabinader @tmcw

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -54,6 +54,7 @@ Known issues:
 - Corrected the dynamic framework’s minimum deployment target to iOS 8.0. ([#3872](https://github.com/mapbox/mapbox-gl-native/pull/3872))
 - Fixed Fabric compatibility. ([#3847](https://github.com/mapbox/mapbox-gl-native/pull/3847))
 - Fixed a crash that can occur when reselecting an annotation. ([#3881](https://github.com/mapbox/mapbox-gl-native/pull/3881))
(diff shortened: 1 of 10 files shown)


1ec5:
Although this fix reaches down into mbgl, I think it’s a very straightforward fix for a case that many users will likely hit when just getting started with the SDK. It’s also a regression that we failed to catch in 3.0.0.

tmcw:
:+1: fix makes sense

brunoabinader:
@1ec5 if it turns out that the `ConstrainWidthAndHeight` logic is not worth the burden, we could remove it entirely from the engine, otherwise I'm good with it.

PR #3914, opened 2016-02-11
less slanted dashes near sharp corners

### LINKED ISSUE #2607: Dotted lines aren't consistently round
Dotted lines ([raster example](http://tile.cycle.travel/topoclassical/16/32500/21679.png)) are traditionally (Illustrator, CartoCSS, etc.) achieved with zero-width dashes, round line caps, and a significant gap, like so:

```ruby
	layer(:paths, :zoom=>13..22) {
		line_color 0x777777
		line_width 10
		line_dasharray [0,2]
		line_cap :round
	}
```

([Glug](https://github.com/systemed/glug)-formatted but you get the idea.)

Doing this in Mapbox GL can produce really entertainingly misshaped dots:

screen shot 2015-10-13 at 22 06 43

(Much though I like the :heart: :heart: :heart: linestyle!)

The dots look fine when the line is at a multiple of 45°, but become lozengy or <3-like at other angles. Even at 45° multiples, the dots get stretched at fractional zoom levels - I guess the 'smart resizing' should only stretch gaps, not dashes, if the dash length is 0.

Tested with 2.1.2 on iOS; I haven't checked Mapbox GL JS to see if it exhibits the same behaviour.

ansis:
port https://github.com/mapbox/mapbox-gl-js/pull/2043 and https://github.com/mapbox/mapbox-gl-js/pull/2113

fix #2607

:eyes: @jfirebaugh

### DIFF
--- a/include/mbgl/util/math.hpp
+++ b/include/mbgl/util/math.hpp
@@ -88,6 +88,11 @@ inline T dist(const S1& a, const S2& b) {
 }
 
 template <typename T>
+inline T round(const T& a) {
+    return T(::round(a.x), ::round(a.y));
+}
+
+template <typename T>
 inline T length(T a, T b) {
     return std::sqrt(a * a + b * b);
 }
--- a/package.json
+++ b/package.json
@@ -26,7 +26,7 @@
   ],
   "devDependencies": {
     "aws-sdk": "^2.2.21",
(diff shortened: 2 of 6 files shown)


jfirebaugh:
Per #3882 I believe we have to use `::round` rather than `std::round`. cc @tmpsantos

ansis:
switched from `std::round` to `::round`.

tmpsantos:
> switched from std::round to ::round.

Thank you! Reason being an oversight in the <cmath> header.

jfirebaugh:
:+1:

PR #3930, opened 2016-02-12
[ios] Update linked library extension to .tbd

friedbunny:
Update some outdated docs that managed to escape my attention. Xcode 7 switched to `.tbd` wrappers around `.dylib`.


friedbunny:
Let's do this against the v3.1.1 release branch instead.

friedbunny:
Committed to the `release-ios-v3.1.0` branch as c5e3b2089181ff2467252dd7cf196df5f1f479d3.

PR #3913, opened 2016-02-11
{ios,osx} ⇢ platform

jfirebaugh:
Fixes #3702.

:eyes: @1ec5

PR #3934, opened 2016-02-13
[ios] Fix screenshot link in generated docs

friedbunny:
We use absolute links to `master` in our docs, so #3913 affects our v3.1.1 release.

This will need to be merged back to `master`.

/cc @1ec5


friedbunny:
To be clear, this is a Friday-night bandaid — there's a better way to do this than link to the raw `master`.

PR #3936, opened 2016-02-13
[core] mbgl::util::noncopyable default dtor should be virtual

brunoabinader:
FWIW `noncopyable` is just a trait we use for disabling copy semantics. However, we are often using non-virtual dtors for its derived classes. This change allows all further derived classes to properly call their respective dtors. Plus for enforcing private inheritance for derived classes.

:eyes: @kkaefer @tmpsantos @jfirebaugh

### DIFF
--- a/include/mbgl/util/noncopyable.hpp
+++ b/include/mbgl/util/noncopyable.hpp
@@ -1,24 +1,20 @@
 #ifndef MBGL_UTIL_NONCOPYABLE
 #define MBGL_UTIL_NONCOPYABLE
 
-namespace mbgl { namespace util {
+namespace mbgl {
+namespace util {
 
-namespace non_copyable_
-{
-
-class noncopyable
-{
+class noncopyable {
 protected:
     constexpr noncopyable() = default;
-    ~noncopyable() = default;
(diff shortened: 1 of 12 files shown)


jfirebaugh:
This would force every object that inherits from `noncopyable` to have a vtable, which in many cases is undesired. It's better to have the class that derives from noncopyable declare the virtual destructor, if and when it's actually necessary.

brunoabinader:
> This would force every object that inherits from noncopyable to have a vtable

Yes, that's exactly the deal: Guarantee that all derived classes from the class the immediately derives from `noncopyable` dtor calls are made vs the associated cost of having a vtable for all `noncopyable`-derived classes. That, IMO, can be considered one of the `rule-of-N` pitfalls.

One alternative would be to have `noncopyable` as a macro - e.g. http://code.qt.io/cgit/qt/qt.git/tree/src/corelib/global/qglobal.h#n2638 - but I'm not much keen for macros in general.

I don't have a strong opinion on which one of the two deals above is better - so if you believe we can deal with the virtual dtors on derived classes on a case-by-case basis, that's :+1: for me.

Issue #2337, opened 2015-09-15
Write and use RAII wrappers for all GL resources


jfirebaugh:
Anything that uses `glGen*` should use a RAII wrapper that takes care of the corresponding `glDelete*`.

cc @brunoabinader

brunoabinader:
I've started digging on this and it seems we already started doing so in `TexturePool` and `GLObjectStore`, however we probably need to implement a destructor for both classes that cleans all remaining textures up (something like this: https://gist.github.com/brunoabinader/57d813f3a69fb10871fd ).

PR #3945, opened 2016-02-15
Implement GL object wrappers

brunoabinader:
Main changes:
- Implement the following wrappers: `ProgramObject`, `ShaderObject`, `BufferObject`, `TextureObject` and `VAOObject`. These wrappers are used inside `std::unique_ptr` to provide RAII.
- Rewrite `TexturePool` to provide proper texture object cleanup.
- Simplify `GLObjectStore` code.
- Rewrite `CustomLayer` tests to verify GL object wrappers implementation.

/cc @jfirebaugh @kkaefer @tmpsantos

### DIFF
--- a/include/mbgl/gl/gl.hpp
+++ b/include/mbgl/gl/gl.hpp
@@ -110,6 +110,32 @@ public:
 using glProc = void (*)();
 void InitializeExtensions(glProc (*getProcAddress)(const char *));
 
+static gl::ExtensionFunction<
+    void (GLuint array)>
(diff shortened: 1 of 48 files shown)


brunoabinader:
FIxes #2336 and #2337.

kkaefer:
Can you please change all of the commit messages to say `Added mbgl::util::*Holder` rather than `*Object`?

It may also be a good idea to move those functions to the `gl` namespace and into the `gl` folder.

PR #3958, opened 2016-02-15
Fix coordinate bounds documentation.

tmcw:
2016-02-15 at 12 43 pm

4-space indentation was being intepreted as a Markdown code block


1ec5:
:+1:

PR #3965, opened 2016-02-16
[core] reset line distance when it is near max

ansis:
port https://github.com/mapbox/mapbox-gl-js/pull/2114

ref https://github.com/mapbox/mapbox-gl-js/issues/2099

When the distance of the current vertex is > half the maximum distance that can be represented, reset the distance. This fixes the original problem where a line with many segments is longer than the maximum distance.

We still need to make sure we can represent the distance of the longest possible segment so we need to make two more changes:

- improve how linesofar is packed into the buffer so that we have two more bits
- scale the distance down before adding it to the buffer to increase the maximum distance that can be represented

I don't like how var distance changed to this.distance but addCurrentVertex needs to be able to reset the distance and I'm not seeing a better way to do this.

### DIFF
--- a/package.json
+++ b/package.json
@@ -26,7 +26,7 @@
   ],
   "devDependencies": {
     "aws-sdk": "^2.2.21",
-    "mapbox-gl-test-suite": "mapbox/mapbox-gl-test-suite#8859b504ef241bca49d4c2f9a79ff4d0bfdf81a1",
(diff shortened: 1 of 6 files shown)


jfirebaugh:
Fixes #4492.

ansis:
I completely forgot about this pull request and accidentally rewrote it: https://github.com/mapbox/mapbox-gl-native/pull/4503

Closing this in favour of the new one.

PR #3983, opened 2016-02-17
[android] #2816 - make save instance state robuster

### LINKED ISSUE #2816: Check if savedInstanceState bundle contains MapView keys
To make the MapView class a bit more robust, we should check if the bundle from `onCreate` contains  any MapView keys before trying to load them. This will prevent the app from crashing as in #2801 and will result in ignoring saved state in case that the end developer has not called the onSaveInstanceState method.

tobrun:
closes #2816


tobrun:

PR #3984, opened 2016-02-17
[androd] #3119 - use resourcePrefix for library resources

### LINKED ISSUE #3119: Rename access_token resource
Been thinking a bit about this and it might be a good idea to rename `access_token` resource:

```xml
<resources>
    <string name="access_token">null</string>
</resources>
```

to

```xml
<resources>
    <string name="mapbox_access_token">null</string>
</resources>
```

Their is always the possibility that the end user or another library, that the end user is using, is overriding the resource by using resource merging. By prepending mapbox to the key we are making a more unique identifier.

tobrun:
closes #3119


tobrun:

PR #3986, opened 2016-02-17
Generate unit tests coverage report via lcov

### LINKED ISSUE #3973: Test coverage metrics
Whilst using GCC, there is a tool called [gcov](https://gcc.gnu.org/onlinedocs/gcc/Gcov.html) that generates test coverage data. Another tool, [lcov](http://ltp.sourceforge.net/coverage/lcov.php), generates HTML pages out of gcov data containing the source code annotated with coverage information - see this [example](http://ltp.sourceforge.net/coverage/lcov/output/index.html) of the generated HTML:
![screen shot 2016-02-16 at 2 10 44 pm]
(https://cloud.githubusercontent.com/assets/76133/13075874/2002ff38-d4b7-11e5-926c-a10fdf1dfdc5.png)
My proposal is: Add a new build targets: `check`, that ultimately runs all unit tests, collects coverage data and generates these handy HTML pages.

/cc @mapbox/mobile @mapbox/gl

brunoabinader:
This PR:
- Instruments `Debug` target with coverage data (supported by GCC and clang)
- Adds a new build target `check` that collects unit tests coverage data and reports it in HTML

Fixes #3973.

/cc @tmpsantos @zugaldia @1ec5 @jfirebaugh @kkaefer

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,6 +4,8 @@
 *.actual.png
 *.diff.png
 *.pyc
+*.gcno
+*.gcda
 offline.db
 /platform/android/debug
 /platform/android/sdk
--- a/Makefile
+++ b/Makefile
@@ -117,6 +117,9 @@ ifeq ($(BUILD),osx)
 xtest: ; $(RUN) HOST=osx HOST_VERSION=x86_64 Xcode/test
 endif
 
+.PHONY: check
+check: ; $(RUN) BUILDTYPE=Debug check
+
 .PHONY: render
 render: ; $(RUN) Makefile/mbgl-render
 
--- a/gyp/common.gypi
+++ b/gyp/common.gypi
@@ -88,7 +88,18 @@
     ],
     'configurations': {
       'Debug': {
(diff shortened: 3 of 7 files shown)


tmpsantos:
Great job, this is really going to be useful to map what areas of our code are lacking utests.

I'm only missing `platform/`. I added it here locally and lcov will only add to the report files that are compiled in the binary.

tmpsantos:
Another minor detail is that my report has the canonical path of each file (which is quite long) instead of the relative path to the root of the tree (preferred).

brunoabinader:
> Another minor detail is that my report has the canonical path of each file (which is quite long) instead of the relative path to the root of the tree (preferred)

Interesting, that should have been fixed by the `--prefix ${PWD}` argument in the `genhtml` call. Here it shows exactly based on the root of the tree.

jfirebaugh:
This is great! Does this mean we can hook the repo up to https://coveralls.io/ and get a +/- report with every PR?

brunoabinader:
>  Does this mean we can hook the repo up to https://coveralls.io/ and get a +/- report with every PR?

Yes :) We already have a Travis CI target which builds in debug mode and runs all unit tests that we could use.

PR #3994, opened 2016-02-17
[core] libuv unref() not really needed

tmpsantos:
The run loop will be kept alive because it has an `AsyncTask`. We also can simple stop the loop with `uv_stop()`.

The `RunLoop` is still gonna be the last object to be destroyed because it is the first object to be created in the thread, so by design it won't outlive `Timer`s and `AsyncTask`s.

This patch won't change the current behavior, will just make the code simpler.

### DIFF
--- a/platform/default/async_task.cpp
+++ b/platform/default/async_task.cpp
@@ -28,7 +28,6 @@ public:
         }
 
         handle()->data = this;
-        uv_unref(handle());
     }
 
     ~Impl() {
--- a/platform/default/run_loop.cpp
+++ b/platform/default/run_loop.cpp
@@ -13,12 +13,6 @@ namespace {
 using namespace mbgl::util;
 static ThreadLocal<RunLoop>& current = *new ThreadLocal<RunLoop>;
 
-#if UV_VERSION_MAJOR == 0 && UV_VERSION_MINOR <= 10
-void dummyCallback(uv_async_t*, int) {};
-#else
-void dummyCallback(uv_async_t*) {};
-#endif
-
 } // namespace
 
 namespace mbgl {
@@ -62,18 +56,7 @@ RunLoop* RunLoop::Get() {
 
 class RunLoop::Impl {
 public:
-    void closeHolder() {
-        uv_close(holderHandle(), [](uv_handle_t* h) {
-            delete reinterpret_cast<uv_async_t*>(h);
-        });
-    }
-
-    uv_handle_t* holderHandle() {
-        return reinterpret_cast<uv_handle_t*>(holder);
-    }
-
     uv_loop_t *loop = nullptr;
-    uv_async_t* holder = new uv_async_t;
 
     RunLoop::Type type;
     std::unique_ptr<AsyncTask> async;
@@ -99,12 +82,6 @@ RunLoop::RunLoop(Type type) : impl(std::make_unique<Impl>()) {
         break;
     }
 
-    // Just for holding a ref to the main loop and keep
-    // it alive as required by libuv.
-    if (uv_async_init(impl->loop, impl->holder, dummyCallback) != 0) {
-        throw std::runtime_error("Failed to initialize async.");
-    }
-
     impl->type = type;
 
     current.set(this);
@@ -113,10 +90,6 @@ RunLoop::RunLoop(Type type) : impl(std::make_unique<Impl>()) {
 
 RunLoop::~RunLoop() {
     current.set(nullptr);
-
-    // Close the dummy handle that we have
-    // just to keep the main loop alive.
-    impl->closeHolder();
 
     if (impl->type == Type::Default) {
         return;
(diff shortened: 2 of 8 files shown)


tmpsantos:
BTW, I just realized that this patch has the best code removal/addition ratio I ever made for a refactoring that still keeps the same functionality.

jfirebaugh:
Hehe :+1:

tmpsantos:
Argh, node is hanging now. I have a few ideias why, but I will look tomorrow.

jfirebaugh:
Probably need to remove the `stop` from [this line](https://github.com/mapbox/mapbox-gl-native/blob/a331e94b1b8d7fcdffe7ef757f4f4c56c9307e8e/platform/node/src/node_mapbox_gl_native.cpp#L26).

tmpsantos:
:eyes: @mikemorris

mikemorris:
@tmpsantos The biggest change here looks like calling `map.release()` may now be required? Previously it was just a faster way to deallocate resources instead of waiting around for garbage collection.

tmpsantos:
@mikemorris libuv keeps the `uv_loop` running until there is no references to the `uv_loop`. `mbgl::RunLoop` will keep one until the last `NodeMap` gets collected by the gc or if you explicitly call `.release()`.

Previously on the tests, node was forcing the gc at the end because it wanted to close the main loop. It won't do it anymore, but the following would work:
`node --expose-gc platform/node/test/js/map.test.js`

```
diff --git a/platform/node/test/js/map.test.js b/platform/node/test/js/map.test.js
index 48bd6fe..e8e8603 100644
--- a/platform/node/test/js/map.test.js
+++ b/platform/node/test/js/map.test.js
@@ -318,8 +318,8 @@ test('Map', function(t) {
             });
             map.load(style);
             map.render({ zoom: 1 }, function(err, data) {
-                map.release();
                 t.ok(data, 'no error');
+                global.gc();
                 t.end();
             });
         });
```

If you want to restore the old behavior, you can. You can force `loop->reset()` when the object refcount drops to zero. `nan` always gets me a bit confused with the traditional V8 API, but I think `NodeMap` is a persistent, and you can set a callback `v8::Persistent::SetWeak` for that.

I could give it a try if you think this is how it should be done. This callback could also call `NodeMap::release()` for you, so you would never need to call it manually (not even expose it).

Note that if we go all-in (i.e. call `release()`), it would make the `NodeMap` destruction almost completely synchronous when the refcount drops to zero and the gc would be merely cleaning up V8 stuff. But that is what we do today manually, right?

mikemorris:
> I think `NodeMap` is a persistent

It's not, the only thing keeping it alive should be references in JavaScript land. I think there's a trick where we can hold a reference to the `loop` with an internal field instead of a shared pointer?

> This callback could also call NodeMap::release() for you

We can't do this, as `NodeMap::release()` destroys the underlying `mbgl::Map` object - we'll often be holding `NodeMap` objects in a cache to reuse them.

> Note that if we go all-in (i.e. call release()), it would make the NodeMap destruction almost completely synchronous when the refcount drops to zero and the gc would be merely cleaning up V8 stuff. But that is what we do today manually, right?

Yep, this is a manual speedup we can do when we know we're done with a `NodeMap` object and we want to immediately dispose of it's resources.

tmpsantos:
> It's not, the only thing keeping it alive should be references in JavaScript land

This is exactly my understanding of what a weak V8::Persistent is. It is like a handle, but heap allocated.

> we'll often be holding NodeMap objects in a cache to reuse them

If you are holding in the cache, it is referenced, so `release()` never gets called by the `::SetWeak` callback.

mikemorris:
> It is like a handle, but heap allocated.

Oh, I wasn't thinking about the internal representation in V8, just the explicit node bindings code.

> never gets called by the ::SetWeak callback

Hmm, was thinking you meant the `map.render` callback in the original post. I can take a closer look tomorrow.

jfirebaugh:
@tmpsantos This doesn't look correct to me. Unless there is a pending still render, `map.release()` should never be required to get a program to exit. Can you please revisit?

tmpsantos:
> Unless there is a pending still render, map.release() should never be required to get a program to exit.

We don't really need to explicitly call `map.release()`, but all NodeMaps need to get `gc()`'d. I was thinking of ways of how to approach this but it all point back to not holding any references to the default `uv_loop` on the main thread.

 - The `SetWeak` solution doesn't work because it also depends on the `gc()` being called. So the application won't exit immediately (like now).
 - Converting how file request is implemented (which is the reason why we need a `mbgl::RunLoop` in the first place) to use a pure `uv_async` would work but...
 - When we move to rendering in the main thread, we will have timers and asyncs holding references again.

I vote for revert and get back to how it was before as it will work with https://github.com/mapbox/mapbox-gl-native/pull/2909.

tmpsantos:
Revert pull request -> https://github.com/mapbox/mapbox-gl-native/pull/4080

I also added a test that explicitly doesn't call `map.release()` to make sure we are not holding references.

mikemorris:
> The SetWeak solution doesn't work because it also depends on the gc() being called.

This would be pretty much the same functionality as before IIRC - I used `map.release()` everywhere in the tests to ensure they exited immediately, but the tests would still exit safely after a `gc()` cleanup even without that. The primary concern here is avoided a lock where `gc()` can't clean up a `NodeMap` object because of a strong reference to the libuv loop, and the tests hang permanently unless `map.release()` is called.

Issue #3720, opened 2016-01-28
Add OnPolygonClickListener and OnPolylineClickListener


buzoherbert:
Hi, first, I'm very happy with the Mapbox products in general!

I'm just implementing the new Android library and I was wondering if there is a way to add a listener to clicks on Polylines. I know this is posible in de js API, but I can't find how to do this in Android.

This would be to change their style and show more information on them.

Is this possible?

Thank you!

tobrun:
@buzoherbert 

Thank you reaching out and requesting this feature, currently we do not expose this as an API but is definitely something we are going to add in future releases. 

@bleege @zugaldia:

Google maps SDK  exposes:
![screen shot 2016-01-28 at 13 21 48](https://cloud.githubusercontent.com/assets/2151639/12644407/ec038d18-c5c2-11e5-9db8-1039580ae6ea.png)

To make this possible we should refactor `SingleTapConfirmed`, it now only takes Markers in account:

```java
        @Override
        public boolean onSingleTapConfirmed(MotionEvent e) {
            // Open / Close InfoWindow
            PointF tapPoint = new PointF(e.getX(), e.getY());

            final float toleranceSides = 15 * mScreenDensity;
            final float toleranceTop = 20 * mScreenDensity;
            final float toleranceBottom = 5 * mScreenDensity;

            RectF tapRect = new RectF(tapPoint.x - toleranceSides, tapPoint.y + toleranceTop,
                    tapPoint.x + toleranceSides, tapPoint.y - toleranceBottom);

            List<LatLng> corners = Arrays.asList(
                    fromScreenLocation(new PointF(tapRect.left, tapRect.bottom)),
                    fromScreenLocation(new PointF(tapRect.left, tapRect.top)),
                    fromScreenLocation(new PointF(tapRect.right, tapRect.top)),
                    fromScreenLocation(new PointF(tapRect.right, tapRect.bottom))
            );

            BoundingBox tapBounds = BoundingBox.fromLatLngs(corners);
            List<Marker> nearbyMarkers = getMarkersInBounds(tapBounds);
            long newSelectedMarkerId = -1;

            if (nearbyMarkers!=null && nearbyMarkers.size() > 0) {
                Collections.sort(nearbyMarkers);
                for (Marker nearbyMarker : nearbyMarkers) {
                    boolean found = false;
                    for (Marker selectedMarker : mSelectedMarkers) {
                        if(selectedMarker.equals(nearbyMarker)){
[...]

PR #3999, opened 2016-02-18
[WIP] [android] #3176 -  marker click only raised once

### LINKED ISSUE #3176: MarkerClick event is only raised once
The concept of a tap/click input _event_ and the default _response_ to that event (select the marker and show info window) seem conflated in the current implementation. This causes problems when the input events are important, but the "selection" state is not.

The most critical problem is that a tap/click on a marker _always_ moves the marker into the "selected" state (regardless of the `MarkerClick` event handler's return value), and once in this state, additional tap/click input no longer raises the `MarkerClick` event. When selection state is not important/visible to the user, it is not obvious to them how to fix this. In addition, there is no easy way at the code level to prevent it from happening.

Some specifics contributing to the confusion:
- A tap/click on an unselected marker _always_ moves it to the "selected" state, even if `mOnMarkerClickListener.onMarkerClick` returns false, indicating the input event was "handled".
- A tap/click on a selected marker does _not_ raise the `MarkerClick` event, even though the associated input event clearly occurred.
- Calling `selectMarker` raises the `MarkerClick` event, even though the method name doesn't say anything about simulating a particular input event, and the event name suggests a particular input event occurred (which it did not).
- The default visual indicator of a marker's selection state is the presence or absence of an info window. However, this is not reliable because an info window may not be defined, it may be prevented from appearing due to the `MarkerClick` event returning `false`, and it may be shown and hidden independently by calling methods on the `Marker` object directly, none of which have any effect on the marker's selection state.

Unfortunately, because the current implementation is so confusing, one cannot be sure exactly how the code is currently being used, and therefore almost any fix could be considered a breaking change for some scenarios.

In the long term, though:
- `MapView.onSingleTapConfirmed` should raise the `MarkerClick` event any time the input event occurs, regardless of the marker's selection state.
- If the `MarkerClick` event handler returns `false`, the _entire_ default response to that input, including any selection state change, should be skipped.
- A new `MarkerSelectionChanged` event should be introduced to allow input events and selection state changes to be handled separately.
- `MapView.selectMarker` should raise the new `MarkerSelectionChanged` event, and should _not_ raise `MarkerClick`.

tobrun:
Work in progress to close #3176

### DIFF
--- a/platform/android/MapboxGLAndroidSDK/src/main/java/com/mapbox/mapboxsdk/annotations/InfoWindow.java
+++ b/platform/android/MapboxGLAndroidSDK/src/main/java/com/mapbox/mapboxsdk/annotations/InfoWindow.java
@@ -226,7 +226,7 @@ public class InfoWindow {
             if(listener!=null){
                 listener.onInfoWindowClose(getBoundMarker());
             }
-            mapboxMap.deselectMarker(getBoundMarker());
+            mapboxMap.deselectAnnotation(getBoundMarker());
             setBoundMarker(null);
         }
     }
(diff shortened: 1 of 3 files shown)


tobrun:
Contains some definitions for #3720

tobrun:
stale

PR #4012, opened 2016-02-18
Include application and operating system name and version in user agent string

### LINKED ISSUE #3997: User agent strings are too generic
Mapbox GL and the various SDKs built atop it are sending overly generic user agent strings in style and tile requests:

* Mapbox iOS and OS X SDKs: [`MapboxGL`](https://github.com/mapbox/mapbox-gl-native/blob/a331e94b1b8d7fcdffe7ef757f4f4c56c9307e8e/platform/darwin/src/http_request_nsurl.mm#L64)
* Mapbox Android SDK: [`MapboxGL/1.0`](https://github.com/mapbox/mapbox-gl-native/blob/20be32f217b860bb0b5af3b32cea2d928b091d29/platform/android/src/http_request_android.cpp#L128), where 1.0 is hard-coded and remains constant across versions
* All other Mapbox GL: [`MapboxGL/1.0`](https://github.com/mapbox/mapbox-gl-native/blob/20be32f217b860bb0b5af3b32cea2d928b091d29/platform/default/http_request_curl.cpp#L390), where 1.0 is hard-coded and remains constant across versions

This stands in contrast to the legacy Mapbox iOS SDK, which [allowed applications to fully customize the user agent string](https://github.com/mapbox/mapbox-ios-sdk-legacy/blob/78ee633d76ed7f088c8747801f60af3d86eb52cc/MapView/Map/RMConfiguration.h#L63), and the legacy Mapbox Android SDK, which used [`Mapbox Android SDK/0.7.0`](https://github.com/mapbox/mapbox-android-sdk-legacy/blob/26d8857786856e5bd933ef6ad28b1d359b2a342a/MapboxAndroidSDK/src/main/java/com/mapbox/mapboxsdk/constants/MapboxConstants.java#L15), where 0.7.0 was the SDK version.

Telemetry events are currently sent over a different code path that specifies a more specific user agent string: `MyTelemetricApp/2.0/3737 MapboxEventsPlatform/1.1`, where MyTelemetryApp is the bundle identifier, 2.0 is the bundle version, 3737 is the build number, _Platform_ is either `iOS` or `Android`, and 1.1 gets incremented every time we change the event format.

The SDKs should be using more specific user agent strings that at least identify the platform and SDK version. Some other attributes (such as the application name and version and the hardware model) can easily be obtained and are customarily included in the user agent string; however, these attributes are already being sent as part of the turnstile event and telemetry.

/cc @mapbox/gl @mapbox/telemetry

1ec5:
The user agent string for all non-telemetry requests in the iOS and OS X SDKs now conforms to the following format:

```
MyApp/0.1.2 Mapbox/3.4.5 MapboxGL/6.7.8 (deadfeed) iOS Simulator/9.0.1 (x86_64)
^ app       ^ sdk        ^ core                    ^ os                 ^ arch
```

This file only links against the cross-platform CoreFoundation and Foundation frameworks, not the platform-specific AppKit or UIKit frameworks, so the options for identifying the operating system and device are fairly limited. (I didn’t want to go through [such](https://github.com/WebKit/webkit/blob/67985c34ffc405f69995e8a35f9c38618625c403/Source/WebCore/page/mac/UserAgentMac.mm) [great](https://github.com/WebKit/webkit/blob/0d5a0900084c2babb65ff85dedcb1a42d854f54c/Source/WebCore/page/ios/UserAgentIOS.mm#L71) [lengths](http://stackoverflow.com/a/5462277/4585461) to include all those details.)

Fixes #3997.

@kkaefer for review. @boundsj, can you confirm that this beefed-up user agent string will be compatible with the style API and dashboard?

### DIFF
--- a/platform/darwin/src/http_file_source.mm
+++ b/platform/darwin/src/http_file_source.mm
@@ -5,9 +5,17 @@
 #include <mbgl/util/http_header.hpp>
 #include <mbgl/util/async_task.hpp>
 
+#include "version.hpp"
+
 #import <Foundation/Foundation.h>
 
 #include <mutex>
+
+@interface MBGLBundleCanary : NSObject
+@end
+
+@implementation MBGLBundleCanary
+@end
 
 namespace mbgl {
 
@@ -82,8 +90,7 @@ public:
 
             session = [NSURLSession sessionWithConfiguration:sessionConfig];
 
-            // Write user agent string
-            userAgent = @"MapboxGL";
+            userAgent = getUserAgent();
 
             accountType = [[NSUserDefaults standardUserDefaults] integerForKey:@"MGLMapboxAccountType"];
         }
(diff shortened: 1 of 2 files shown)


kkaefer:
Can we use `TargetConditionals.h` to know more about the platform we're building for? So instead of `Darwin`, we'd use `iOS`, `iOS Simulator`, or `Mac OS X`.

kkaefer:
Ah sorry, the last commit does that.

Can we cache the value of the the User Agent string composition?

1ec5:
@kkaefer, do you know why the version.hpp include would fail only on clean builds for iOS (but not OS X)?

mikemorris:
> do you know why the version.hpp include would fail only on clean builds for iOS (but not OS X)?

Because it's generated and the iOS build might be missing this step? https://github.com/mapbox/mapbox-gl-native/blob/master/gyp/version.gypi

kkaefer:
> since we can detect it from the architecture

what are we going to do when Apple introduces an x86_64 iOS device?

1ec5:
> what are we going to do when Apple introduces an x86_64 iOS device?

We’re going to build an SDK for it and update this logic. :smile:

1ec5:
> Because it's generated and the iOS build might be missing this step? https://github.com/mapbox/mapbox-gl-native/blob/master/gyp/version.gypi

@mikemorris, [including a `version` dependency](https://github.com/mapbox/mapbox-gl-native/blob/3e1ecb83852a467d5b14e79e84297597f3fde709/gyp/platform-ios.gypi#L10) should be enough, right?

Issue #3275, opened 2015-12-13
Migrate iOS targets to checked-in Xcode projects


1ec5:
Currently, on OS X, all libraries and executables are built by Xcode projects generated by gyp. This system is ill-suited for iOS and OS X development: for instance, to add a file to the iOS SDK, you need to move the file to the right folder, `git add` the file, and finally edit the appropriate .gypi file to include it. The process gets even more involved when adding a new target, such as a test bundle: you need to essentially reverse-engineer Xcode’s template for that target to ensure that all the required build settings are persisted across invocations of `make iproj`.

We should migrate the platform-ios and iosapp targets to a checked-in Xcode project, similar to the [ios-tests.xcodeproj](https://github.com/mapbox/mapbox-gl-native/tree/master/test/ios/ios-tests.xcodeproj) that’s already checked in. Adding a new file thus becomes a matter of adding it in Xcode (which handles git and any required build phase updates automatically). If folks are uncomfortable with the idea of maintaining build settings in a .pbxproj file, we can put those settings in .xcconfig files ([example](https://github.com/realm/realm-cocoa/blob/master/Configuration/Realm/Realm.xcconfig)), which are less prone to conflicts.

The new Xcode project would continue to depend on the output of gyp-generated core projects. We’d probably check in a workspace that ties together all the projects to ensure that they get built together. With these changes, we’ll also be able to remove much of the shell scripting that forms the final part of the iOS SDK build process.

/cc @mapbox/gl

1ec5:
GitHub, that was the subjunctive mood, not the indicative.

kkaefer:
lol

PR #3989, opened 2016-02-18
Run unit tests on iOS

kkaefer:
Follow up to https://github.com/mapbox/mapbox-gl-native/issues/2150


kkaefer:
@1ec5 how can we run this on Bitrise? There's now a new target called "test" in the default Xcode project for iOS. That target works on both actual devices and the simulator and all tests are passing.

@jfirebaugh There are two OfflineDatabase tests that fail on iOS: `PutEvictsLeastRecentlyUsedResources` and `PutFailsWhenEvictionInsuffices`. Do you have any idea why they might be failing?

1ec5:
> how can we run this on Bitrise?

The Bitrise steps are defined in [bitrise.yml](https://github.com/mapbox/mapbox-gl-native/blob/87f28f3f4456892de07d67fcdba8e9ebaed2243a/platform/ios/bitrise.yml). It essentially does two things: it runs `make ipackage-sim` on the command line and then tests ios-tests.xcodeproj using Xcode. Since the test target has been added to a gyp-generated project (#3275), you’ll have to add a make target that builds and runs the tests. That make target could invoke gyp, followed by a shell script that looks like [test.sh](https://github.com/mapbox/mapbox-gl-native/blob/87f28f3f4456892de07d67fcdba8e9ebaed2243a/platform/ios/scripts/test.sh). (test.sh is invoked by `make itest`, but Bitrise doesn’t use it.)

jfirebaugh:
The eviction tests were tuned for a database page size of 1024, while on iOS the page size is 4096. I pushed a change that makes it work for both.

kkaefer:
Ah interesting. Reading the patch, it's unclear to me how this would fix the issue. Can you please explain? If our eviction is based on page size, we should enforce page size with `PRAGMA page_size = 1024;` (or whatever value we chose) so that all platforms behave the same.

jfirebaugh:
The [eviction algorithm](https://github.com/mapbox/mapbox-gl-native/blob/master/platform/default/mbgl/storage/offline_database.cpp#L464-L521) tries to keep the total size of the database below a specified limit. The tests were setting this limit to `1024 * 25` = 25,600 bytes. SQLite appears to enforce a minimum of 15 pages in a database. In cases where the page size is 1024, a limit of 25,600 bytes is above this minimum size of `1024 * 15`, but if the page size is 4096, then even an empty database is larger than 25,600 bytes. Increasing the limit to `1024 * 100` ensures that the eviction limit is greater than the minimum database size in both cases.

A better way to write this test would be to query the page size and minimum number of pages dynamically, and then create a fresh database with an eviction limit greater than the product of those two. But I took the lazier route of just bumping the numbers.

PR #4028, opened 2016-02-18
Set default polyline and polygon color to match app

### LINKED ISSUE #3927: Default polyline color should match tint color
The default polyline color, which is currently blue, should match the tint color on iOS or the text selection color on OS X.

/cc @friedbunny @boundsj

friedbunny:
**iOS**
- Set the poly annotation stroke and fill color to the map view's tint color

**OS X**
- Set the poly annotation stroke and fill color to the system menu item selection color
- Add polygon and polyline examples

Fixes #3927.

/cc @1ec5

### DIFF
--- a/platform/osx/app/MainMenu.xib
+++ b/platform/osx/app/MainMenu.xib
@@ -1,7 +1,7 @@
 <?xml version="1.0" encoding="UTF-8" standalone="no"?>
-<document type="com.apple.InterfaceBuilder3.Cocoa.XIB" version="3.0" toolsVersion="9531" systemVersion="15C50" targetRuntime="MacOSX.Cocoa" propertyAccessControl="none" useAutolayout="YES" customObjectInstantitationMethod="direct">
+<document type="com.apple.InterfaceBuilder3.Cocoa.XIB" version="3.0" toolsVersion="10109" systemVersion="15E39d" targetRuntime="MacOSX.Cocoa" propertyAccessControl="none" useAutolayout="YES" customObjectInstantitationMethod="direct">
     <dependencies>
-        <plugIn identifier="com.apple.InterfaceBuilder.CocoaPlugin" version="9531"/>
+        <plugIn identifier="com.apple.InterfaceBuilder.CocoaPlugin" version="10109"/>
     </dependencies>
     <objects>
         <customObject id="-2" userLabel="File's Owner" customClass="NSApplication">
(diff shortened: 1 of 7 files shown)


platform/osx/app/MainMenu.xib
```diff
@@ -473,10 +473,16 @@
                                     <action selector="dropManyPins:" target="-1" id="Rtv-8N-3Z8"/>
                                 </connections>
                             </menuItem>
-                            <menuItem title="Remove All Pins" id="6rC-68-vk0">
+                            <menuItem title="Draw Example Polygon and Polyline" id="DVr-vT-lpe">
```
1ec5: Should be called “Add Polygon and Polyline”, since it isn’t an interactive drawing tool.

platform/osx/app/MapDocument.m
```diff
@@ -457,8 +488,8 @@ - (BOOL)validateMenuItem:(NSMenuItem *)menuItem {
     if (menuItem.action == @selector(dropManyPins:)) {
         return YES;
     }
-    if (menuItem.action == @selector(removeAllPins:)) {
-        return self.mapView.annotations.count;
+    if (menuItem.action == @selector(removeAllAnnotations:)) {
+        return self.mapView.annotations.count ? YES : NO;
```
1ec5: Eh, this isn’t strictly necessary. If you prefer to be explicit, it’d be better to return `self.mapView.annotations.count > 0`.

friedbunny:
simulator screen shot feb 18 2016 5 43 10 pm|simulator screen shot feb 18 2016 5 42 09 pm|
|---|---|
||

screen shot 2016-02-18 at 5 48 08 pm

platform/osx/app/MapDocument.m
friedbunny: Did this because `self.mapView.annotations.count` alone wasn't doing the job and `YES`/`NO` is pleasing semantically.

platform/osx/app/MainMenu.xib
friedbunny: Good point.

platform/osx/app/MapDocument.m
friedbunny: Switched to your suggestion in https://github.com/mapbox/mapbox-gl-native/commit/21ce6fbdf96e80272fa0893640f7da85497aa9aa#diff-1478e1e346e69c2437356a66353ddbf8R492.

PR #4029, opened 2016-02-18
Import required frameworks with static framework umbrella

1ec5:
This change corrects a regression introduced in #3183, when the umbrella header used in version 3.0.1 was removed. The umbrella header inside the static framework now imports the required iOS SDK frameworks. With this change, it’s no longer necessary to manually link any frameworks other than Mapbox.framework.

/cc @friedbunny


1ec5:
This change was partially backed out in 7bed2cebd2d78762650dbe2cbf895fc1673ba8da. If you start out with a storyboard and the static framework, you won’t be able to link the project until you import the module. To work around that, you need to link the various and sundry framework dependencies.

Issue #3832, opened 2016-02-05
Improve unit testing of geometry package


tobrun:
Since #3830 it is possible to unit test the Android `Parceable` interface using mockito.
This allowed us to achieve 100% code coverage on the `LatLng` model class.
We should aim to bring same coverage to the others classes in the `geometry` package as well.

Current state:
 
![screen shot 2016-02-05 at 10 24 22](https://cloud.githubusercontent.com/assets/2151639/12842454/3cc3ac98-cbf3-11e5-8d4e-823ca1d3b691.png)

tobrun:
Putting a note here that mocking Parceable will require some extra work to get in.

tobrun:
I was able to make above work for all types of Parceable:

![screen shot 2016-03-30 at 15 56 04](https://cloud.githubusercontent.com/assets/2151639/14144430/f47bdcc2-f68f-11e5-92cf-385b624b930d.png)

PR #4031, opened 2016-02-19
Hit missed lines in equality tests

tmcw:
Refs #3832 cc @tobrun for the review.


tobrun:
These are great! I will rebase and merge

PR #4034, opened 2016-02-19
[ios] Refactor location services

friedbunny:
Continuation of #3641, rebased onto the `release-ios-v3.1.0` branch for the upcoming v3.1.2 release.

- Pause location services when the app goes to sleep, address iOS 8 blur bar issue.
- Only ask for location permissions if not already granted/denied.

Partially addresses #2945.

/cc @1ec5 @boundsj


platform/ios/src/MGLMapView.mm
```diff
@@ -3022,45 +3026,31 @@ - (void)annotationImageNeedsRedisplay:(MGLAnnotationImage *)annotationImage
 
 #pragma mark - User Location -
 
-- (void)setShowsUserLocation:(BOOL)showsUserLocation
+- (void)validateLocationServices
 {
-    if (showsUserLocation == _showsUserLocation || _isTargetingInterfaceBuilder) return;
-
-    _showsUserLocation = showsUserLocation;
+    BOOL shouldEnableLocationServices = !self.locationManager && self.showsUserLocation;
```
1ec5: Validation methods shouldn’t have preconditions. This flag should additionally check for `self.dormant`, such that you could call this method twice in a row without producing the opposite result.

boundsj:
@friedbunny :+1: although for "Only ask for location permissions if not already granted/denied" the CLLocationManager API [already guards against redundant requests](https://developer.apple.com/library/ios/documentation/CoreLocation/Reference/CLLocationManager_Class/#//apple_ref/occ/instm/CLLocationManager/requestAlwaysAuthorization) although this certainly does not hurt anything.

friedbunny:
@boundsj Aye, checking for `kCLAuthorizationStatusNotDetermined` isn't strictly necessary, but doing so is more explicit and we can avoid an extra little bit of work in most situations if we do check the authorization status ourselves.

1ec5:
:+1:

PR #4042, opened 2016-02-19
Don't use strings as IDs

kkaefer:
We're using strings as unique IDs in many places, but strings are dynamically allocated, take up unnecessary space and are costly to compare/sort. We should instead switch to numeric IDs.

### DIFF
--- a/test/util/id.cpp
+++ b/test/util/id.cpp
@@ -1,0 +1,82 @@
+#include "../fixtures/util.hpp"
+
+#include <mbgl/util/id.hpp>
+
+#include <set>
+#include <sstream>
+#include <unordered_set>
+
+using namespace mbgl;
+
+class Foo;
+
+TEST(ID, Creation) {
+    const util::ID<Foo> foo("foo");
+    const util::ID<Foo> bar = { "bar" };
+    const util::ID<Foo> baz;
+    const util::ID<Foo> boo = { "" };
+
+    EXPECT_TRUE(foo.valid());
+    EXPECT_TRUE(bar.valid());
+    EXPECT_FALSE(baz.valid());
+    EXPECT_FALSE(boo.valid());
+}
+
+TEST(ID, Assignment) {
+    const util::ID<Foo> foo("foo");
+    util::ID<Foo> bar("bar");
+    util::ID<Foo> baz;
+    bar = "test";
+    bar = foo;
+    baz = "baz";
+
+    EXPECT_TRUE(foo.valid());
+    EXPECT_TRUE(bar.valid());
+    EXPECT_TRUE(baz.valid());
+}
+
+TEST(ID, Comparison) {
+    const util::ID<Foo> foo("foo");
+    const util::ID<Foo> baz;
+
+    EXPECT_EQ(foo, util::ID<Foo>("foo"));
+    EXPECT_NE(foo, util::ID<Foo>("bar"));
+
+    EXPECT_FALSE(baz == foo);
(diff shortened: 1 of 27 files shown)


jfirebaugh:
Does this show significant benefits in real world profiling benchmarks that justify the additional complexity?

test/util/id.cpp
```diff
@@ -0,0 +1,76 @@
...
+    EXPECT_TRUE(baz != foo);
+    EXPECT_TRUE(foo != baz);
+    EXPECT_TRUE(baz != baz);
+
+    EXPECT_EQ("foo", foo.str());
+    EXPECT_STREQ("foo", foo.c_str());
+    EXPECT_EQ("", baz.str());
+}
+
+TEST(ID, Sorting) {
```
tmpsantos: Maybe you should test the `operator<` and make sure that sort ordering is really the expected.

kkaefer:
Yeah; a good chunk of `renderSync` is eaten up by looking up things with strings:

Before:
instruments2 2016-02-23 14-29-16


After:
instruments2 2016-02-23 14-30-22

kkaefer:
Same picture on iOS; running times with repeated executions are very stable, at exactly 100 rendered frames:

Before:
instruments2 2016-02-23 16-49-02

After:
instruments2 2016-02-23 16-49-41

jfirebaugh:
My concerns with this are:

* Introduction of potential lock contention on code paths that do Dictionary lookup.
* Unbounded memory growth: there's no mechanism for evicting entries from the Dictionary when they're no longer needed. It will always contain the set of all source and layer IDs ever used by the program.

Are there algorithmic approaches we could take instead, to avoid doing so many lookups into `std::string`-keyed hashes? (I assume we're already using `unordered_map` everywhere rather than `map`.)

kkaefer:
> Are there algorithmic approaches we could take instead

I looked into this as well. There are two hot paths what do lookups by name currently, both in `Style::getRenderData()`

- `getSource(name)` for every layer does a lookup by name. We could probably intercept all changes to source objects in `Style::addSource/addLayer` and store a direct pointer to the Source object in the `StyleLayer` object
- but even hotter is `TileData::getBucket(name)`, which is a virtual call to `VectorTileData/RasterTileData` that performs a lookup by name. It's done once per tile per layer per rendered frame. Buckets are changing ocassionally, (e.g. when they are reparsed due to expiry or because we initially loaded stale data).

Instead of calling `getRenderData()` for every frame, we could also attempt to only call it whenever something in the buckets/sources changed. I implemented that originally, but it was eventually reverted due to side effects that made the app crash.

brunoabinader:
Great idea @kkaefer! A few comments below:

This can be useful for storing class names in `MapData` as well.

> Unbounded memory growth

Memory-wise, I believe the odds of having a bunch of individual string objects per item is higher than having a non-evicting dictionary. New source and layer names are scoped to style loading, which doesn't happen too often, and different class names are usually very low.

I believe we could also pre-populate a list of known source/layer/class names, in a similar way how Servo's [string-cache](https://github.com/servo/string-cache/blob/master/src/static_atom_list.rs) does.

jfirebaugh:
> New source and layer names are scoped to style loading, which doesn't happen too often

This won't be true when the styles API lands. With a styles API, there are valid use cases for using many unique layer names. Doing so shouldn't produce memory leaks in the library.

jfirebaugh:
Closing a stale and per above.

Issue #3717, opened 2016-01-28
MGLMapboxEvents should run on the main thread


1ec5:
MGLMapboxEvents is far more complex than it needs to be. There are `atomic` properties and lots of wrapper methods and synchronization to ensure that individual methods can be called from any thread, yet the class’s entry points are only ever called from MGLMapView and MGLAccountManager on the main thread. If there’s any concern about rapid-fire events code hurting performance, then we should isolate the potentially costly code (like the code that interfaces with System Configuration or sets up the network request) and run only that code asynchronously.

/ref #3704
/cc @boundsj @bleege

bleege:
> MGLMapboxEvents is far more complex than it needs to be.

+1 to this statement.  Now that the use case has matured for iOS Telemetry lets look to simplify it for better developer understanding (internally and externally) and better performance.  Hope you brought your :machete: and :pithhelmet: @boundsj!

PR #4046, opened 2016-02-19
Reduce async complexity

boundsj:
Since MGLMapboxEvents is only accessed via the main thread the
complexity to ensure thread safety is not required. This removes
that extra code so that future work can benefit from being based
on a simpler foundation.

The main changes in this commit are:

- Untangle nested calls to GCD dispatch_async (i.e. `pushEvents` used to
dispatch onto a global (concurrent) background queue from a public
class method and then those blocks would be dispatched onto a custom
serial queue). Now, the only calls to dispatch_async that are left are
in `- postEvents:` and `- writeEventToLocalDebugLog:` and the
performance requirement to have even that dispatching done is still
unclear. I'm keeping it for now only because it allows for non-blocking
JSON serialization and pre-validation (network requests are already
non-blocking anyway)
- Since the async dispatches are pushed down to just wrap around
the JSON serialization / networking, this commit also removes the main
thread marshaling that was done in many places (i.e. guarding access
to UIKit methods that provide data for events that used to be
(synchronously) gathered via the main thread)
- Remove assertions that we are running on the main thread.
- Pushed some bools used for guards (i.e. `debugLoggingEnabled`) down
into the methods that really need to know about the bool value to
eliminate some repetitive boilerplate guarding all over the place
- Make progress towards a consistent code style

cc @1ec5 @friedbunny

### DIFF
--- a/platform/ios/src/MGLMapboxEvents.m
+++ b/platform/ios/src/MGLMapboxEvents.m
@@ -1,15 +1,12 @@
 #import "MGLMapboxEvents.h"
-
 #import <UIKit/UIKit.h>
 #import <CoreLocation/CoreLocation.h>
-
-#include <mbgl/platform/darwin/reachability.h>
-
 #import "MGLAccountManager.h"
 #import "NSProcessInfo+MGLAdditions.h"
 #import "NSBundle+MGLAdditions.h"
 #import "NSException+MGLAdditions.h"
 
+#include <mbgl/platform/darwin/reachability.h>
 #include <sys/sysctl.h>
 
 static const NSUInteger version = 1;
@@ -53,14 +50,11 @@ const NSTimeInterval MGLFlushInterval = 60;
 
 @interface MGLMapboxEventsData : NSObject
 
-// All of the following properties are written to only from
-// the main thread, but can be read on any thread.
-//
-@property (atomic) NSString *vendorId;
-@property (atomic) NSString *model;
-@property (atomic) NSString *iOSVersion;
-@property (atomic) NSString *carrier;
-@property (atomic) CGFloat scale;
+@property (nonatomic) NSString *vendorId;
+@property (nonatomic) NSString *model;
+@property (nonatomic) NSString *iOSVersion;
+@property (nonatomic) NSString *carrier;
+@property (nonatomic) CGFloat scale;
 
 @end
 
@@ -81,10 +75,7 @@ const NSTimeInterval MGLFlushInterval = 60;
     return self;
 }
 
-// Can be called from any thread.
-//
-- (NSString *)sysInfoByName:(char *)typeSpecifier
-{
+- (NSString *)sysInfoByName:(char *)typeSpecifier {
     size_t size;
     sysctlbyname(typeSpecifier, NULL, &size, NULL, 0);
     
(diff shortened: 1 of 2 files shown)


boundsj:
@1ec5 this attempts to address https://github.com/mapbox/mapbox-gl-native/issues/3717

platform/ios/src/MGLMapboxEvents.m
```diff
@@ -180,20 +135,16 @@ + (BOOL)isEnabled {
             [[NSUserDefaults standardUserDefaults] integerForKey:@"MGLMapboxAccountType"] == 0);
 }
 
++ (BOOL)debugLoggingEnabled {
```
1ec5: This convenience method appears to be dead code. We can remove it.

platform/ios/src/MGLMapboxEvents.m
```diff
@@ -357,253 +306,198 @@ + (void)pauseMetricsCollection {
...
         return;
     }
-    _locationManager = [[CLLocationManager alloc] init];
-    _locationManager.desiredAccuracy = kCLLocationAccuracyKilometer;
-    _locationManager.distanceFilter = 10;
-    _locationManager.delegate = self;
+    self.locationManager = [[CLLocationManager alloc] init];
+    self.locationManager.desiredAccuracy = kCLLocationAccuracyKilometer;
+    self.locationManager.distanceFilter = 10;
+    self.locationManager.delegate = self;
```
1ec5: Nit: Let’s reduce the repeated calls to `-locationManager` with a local variable that gets assigned to the property at the end.

platform/ios/src/MGLMapboxEvents.m
```diff
@@ -357,253 +306,198 @@ + (void)pauseMetricsCollection {
...
+    // mapbox-events-ios stock attributes
+    [evt setValue:self.data.model forKey:@"model"];
+    [evt setValue:self.data.iOSVersion forKey:@"operatingSystem"];
+    [evt setValue:@(self.data.scale) forKey:@"resolution"];
+    MGLReachability *reachability = [MGLReachability reachabilityForLocalWiFi];
+    [evt setValue:([reachability isReachableViaWiFi] ? @YES : @NO) forKey:@"wifi"];
+    [evt setValue:[self deviceOrientation] forKey:@"orientation"];
+    [evt setValue:@([self batteryLevel]) forKey:@"batteryLevel"];
+    [evt setValue:[self applicationState] forKey:@"applicationState"];
+    [evt setValue:@([self contentSizeScale]) forKey:@"accessibilityFontScale"];
```
1ec5: There’s inconsistent usage of `-setObject:forKey:` versus `-setValue:forKey:` here. Moreover, since all of the value-setting is unconditional, this should all be collapsed into a call to `-[NSMutableDictionary dictionaryWithObjectsAndKeys:]` and merged with `attributeDictionary` using `-addEntriesFromDictionary:`.

platform/ios/src/MGLMapboxEvents.m
```diff
@@ -964,70 +781,58 @@ - (void)URLSession:(NSURLSession *)session didReceiveChallenge:(NSURLAuthenticat
...
 
-    if ( ! _debugLogSerialQueue) {
+    if (!self.debugLogSerialQueue) {
         NSString *uniqueID = [[NSProcessInfo processInfo] globallyUniqueString];
-        _debugLogSerialQueue = dispatch_queue_create([[NSString stringWithFormat:@"%@.%@.events.debugLog", _appBundleId, uniqueID] UTF8String], DISPATCH_QUEUE_SERIAL);
+        self.debugLogSerialQueue = dispatch_queue_create([[NSString stringWithFormat:@"%@.%@.events.debugLog", _appBundleId, uniqueID] UTF8String], DISPATCH_QUEUE_SERIAL);
     }
 
-    dispatch_sync(_debugLogSerialQueue, ^{
+    dispatch_async(self.debugLogSerialQueue, ^{
```
1ec5: Good call.

platform/ios/src/MGLMapboxEvents.m
friedbunny: Yeah, this should go away. At one point I had some debug logging in MGLMapView, but those keys didn't make the final cut.

platform/ios/src/MGLMapboxEvents.m
friedbunny: I think we may have originally been relying on `-setValue:forKey:` to not create keys if we fed in `nil` values. Definitely we can do this better, but perhaps as a part of the larger refactor.

1ec5: Ah, that makes sense.

boundsj:
Closed with https://github.com/mapbox/mapbox-gl-native/commit/f3ff38d1259a4949c6ac416fa451644477b044ef

Issue #3940, opened 2016-02-14
Offline support doesn't work correctly with tile expiration


brewin:
Nice work on adding offline support. It's much appreciated! Unfortunately, it is causing some problems with displaying layers that are updated very frequently, such as weather layers.
- Expires is set to -1 on the tile server, but Mapbox GL does not honor it. The tiles are cached anyway and will load with airplane mode enabled.
- The tiles are downloaded continuously and indefinitely. (Related to #3667)
- There is no way to manually refresh a layer, either by invalidating the cache or just by updating the layer on screen if there is no cache.

jfirebaugh:
I'm not sure what you mean by "Expires is set to -1". If you mean that your tile server is sending `Expires: -1` as a response header, that is not a valid `Expires` value. The value must be an RFC 1123 formatted date.

The intended behavior for frequently updated layers requires the cooperation of the tile server:

* The tile server should send caching headers (`Expires` or `Cache-Control`) that reflect when new data will be available. I.e., an `Expires` header with the date when a new tile will be available, or a `Cache-Control` `max-age` value that is the actual refresh interval.
* When one of those headers is present, mapbox-gl-native will refresh the tile accordingly, even while that tile is visible on screen.

Sending no cache headers (or an invalid cache header such as `Expires: -1`) doesn't give mapbox-gl-native any information about when to refresh the tile, so it can only fall back to refreshing it when the map is recreated or the application is relaunched.

brewin:
Thanks for the help. Sorry, I meant `expires` is set to -1 in the nginx configuration. According to the [docs](http://nginx.org/en/docs/http/ngx_http_headers_module.html#expires), this causes `Cache-Control: no-cache` to be sent and the `Expires` time is always the time when the tile is accessed. 

Here are all the headers sent:
```
HTTP/1.1 200 OK
Server: nginx/1.4.6 (Ubuntu)
Date: Mon, 15 Feb 2016 18:23:57 GMT
Content-Type: image/png
Content-Length: 320
Last-Modified: Mon, 15 Feb 2016 18:22:16 GMT
Connection: keep-alive
ETag: "56c21758-140"
Expires: Mon, 15 Feb 2016 18:23:56 GMT
Cache-Control: no-cache
Accept-Ranges: bytes
```
With these headers, the following happens:
[...]

PR #4019, opened 2016-02-21
[core] HTTP replies with expiration time already expired should be treated as errors

tmpsantos:
The `OnlineFileSource` should treat a HTTP reply with expiration time already expired as an error and apply the same delay rules for retrying that we have in place for other errors, otherwise a faulty server can trigger an endless number of consecutive requests.

/cc @jfirebaugh

### DIFF
--- a/include/mbgl/util/constants.hpp
+++ b/include/mbgl/util/constants.hpp
@@ -1,6 +1,8 @@
 #ifndef MBGL_UTIL_CONSTANTS
 #define MBGL_UTIL_CONSTANTS
 
+#include <mbgl/util/chrono.hpp>
+
 #include <cmath>
 #include <string>
 #include <vector>
@@ -22,6 +24,8 @@ extern const double MIN_ZOOM;
 extern const double MAX_ZOOM;
 
 extern const uint64_t DEFAULT_MAX_CACHE_SIZE;
+
+extern const Duration CLOCK_SKEW_RETRY_TIMEOUT;
 
 } // namespace util
 
--- a/platform/default/online_file_source.cpp
+++ b/platform/default/online_file_source.cpp
@@ -5,6 +5,7 @@
 #include <mbgl/storage/response.hpp>
 #include <mbgl/platform/log.hpp>
 
+#include <mbgl/util/constants.hpp>
 #include <mbgl/util/thread.hpp>
 #include <mbgl/util/mapbox.hpp>
 #include <mbgl/util/exception.hpp>
@@ -29,7 +30,7 @@ public:
     ~OnlineFileRequestImpl();
 
     void networkIsReachableAgain(OnlineFileSource::Impl&);
-    void schedule(OnlineFileSource::Impl&, bool forceImmediate = false);
+    void schedule(OnlineFileSource::Impl&, optional<SystemTimePoint> expires);
     void completed(OnlineFileSource::Impl&, Response);
 
     FileRequest* key;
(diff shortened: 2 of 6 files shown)


tmpsantos:
:eyes: @jfirebaugh 

> Enforce a minimum revalidation interval. This would need to be relatively short, to not interfere with legitimate use cases for frequent revalidation.

This could be easily done with something like this, but I'm not sure if we should enforce it ATM. The only use case I can think of is a server serving tiles with a very short TTL, but that would have to be very close to the minimum we define (that is also going to be no higher than 5s). It will probably hurt more the legit usage than help.

`timeout = std::max(timeout, mbgl::minimumRetryTimeout);`

jfirebaugh:
We need to think about what will happen in cases of clock skew with servers configured like in https://github.com/mapbox/mapbox-gl-native/issues/3940#issuecomment-184442025, serving resources with `Expires` set to the server's idea of the current time, or potentially the server's idea of a "near future" time.

If the server clock is slightly ahead of the device's clock, then the responses will be interpreted as having a very short lifetime. They'll be revalidated on a period equal to the difference between the device clock and the server clock, without exponential backoff. The device has no way to distinguish this case from the case that where the clocks are in sync and it's _intended_ that the resource be revalidated in the near future. Therefore we can accept this behavior. If, in fact, the desired behavior is that the resource _not_ be revalidated, `Cache-Control` needs to be used, and mbgl should support this as I suggested in https://github.com/mapbox/mapbox-gl-native/issues/3940#issuecomment-186030198.

If the server clock is slightly _behind_ the device's clock, then the resource will be interpreted as always being expired when received, and the new exponential backoff behavior will be triggered, even though the value of `Expires` is changing with each response. Again, the device has no way to distinguish this case from the case where the clocks are in sync and it's intended that the resource be revalidated in the near future. But that's a valid use case for `Expires` for which exponential backoff is inappropriate.

My conclusion: in order to support the valid latter case, the condition for exponential backoff should be that the `Expires` value is in the past and _not changing_, not merely that it is in the past.

This leaves open the question of what the revalidation schedule should be for  `Expires` values that are in the past, but changing. I think we need to pick a short, fixed period (~seconds), and use that.

jfirebaugh:
[Slight edits to previous comment to correct a mistake.]

tmpsantos:
> My conclusion: in order to support the valid latter case, the condition for exponential backoff should be that the Expires value is in the past and not changing, not merely that it is in the past.

Thanks for the write-up, very insightful.

> This leaves open the question of what the revalidation schedule should be for Expires values that are in the past, but changing. I think we need to pick a short, fixed period (~seconds), and use that.

We could use the delta from the previous expiration time and the current expiration time when it changes and interpolate the intended expiration time. 

Note that what is going to happen is:

- Current time: 10:26:00 - attempt 1 (expired at 10:25)
- Current time: 10:26:04 - attempt 2 (expired at 10:25) exponential backoff
- Current time: 10:26:08 - attempt 3 (expired at 10:25) exponential backoff
...
- Current time: 10:43:04 - attempt N (expired at 10:35) delta 10 min, schedule next retry for 10:45:04 local time. ~8 minutes have passed already and the probably the next expiration time is in ~2 minutes).

What do you think?

jfirebaugh:
I still think that to avoid potentially hammering servers that are (mis-)configured to send `Expires` set to the server's idea of the current time, and no `Cache-Control`, we should have a fixed lower bound on the revalidation timer of few seconds. The delta between `Expires` that you suggest could be used to increase that lower bound.

tmpsantos:
Take 2 - :eyes: @jfirebaugh

jfirebaugh:
:+1:

include/mbgl/util/constants.hpp
```diff
@@ -23,6 +23,8 @@ extern const double MAX_ZOOM;
 
 extern const uint64_t DEFAULT_MAX_CACHE_SIZE;
 
+extern const int8_t CLOCK_SKEW_RETRY_TIMEOUT;
```
jfirebaugh: Oh, make this a `Duration`, so the units are explicit.

Issue #3778, opened 2016-02-01
[iOS] Wrong map insets behavior while tracking user location


RomainQuidet:
Bug visible on current master (3.1.0 pre 3)
Show a map, set the map contentInset. The map moves accordingly. This is correct.
Then set the user tracking to follow. The map then "moves back" to its original position and only the user location is inset. This makes a double error : the map is not correctly inset and the user location is incorrect.

1ec5:
Does this bug reproduce in pre.2 or pre.1?

1ec5:
There are a number of possible reasons why the content inset might appear to be reverted. I’m unable to reproduce this issue by setting the content inset then setting the user tracking mode. Are you using the plain property setters or the `animated:` methods? Does the view controller have a top or bottom bar that might be overriding the content inset? Is the user tracking mode set immediately (synchronously) after setting the content inset, or is there some delay?

RomainQuidet:
Hi. I'm trying to inset on the right and bottom.
The view controller managing the map is part of a navigation controller so there's a navigation bar.
I did some tests, it always behave the same : I'm setting the inset to the map. I tried both with animation api or directly with the setter. Then I set userTrackingMode to follow. Even if I set it after 2s, there is a map shift which leads to an incorrect position of the marker vs the map. The user marker is correctly centered in the inset area, but the map suffers from an extra offset. 
I also just tried to set the user mode to follow on map appearance, then switch on/off the insets. Here are some screenshots to show the behavior.

![simulator screen shot 2 fevr 2016 10 39 54](https://cloud.githubusercontent.com/assets/8179069/12745265/ba7467f2-c998-11e5-96f9-6f76129333ba.png)
![simulator screen shot 2 fevr 2016 10 40 00](https://cloud.githubusercontent.com/assets/8179069/12745264/ba7261c8-c998-11e5-82c9-039bb1822bc6.png)
![simulator screen shot 2 fevr 2016 10 40 38](https://cloud.githubusercontent.com/assets/8179069/12745263/ba724e5e-c998-11e5-9176-d2f6463d679e.png)
![simulator screen shot 2 fevr 2016 10 40 42](https://cloud.githubusercontent.com/assets/8179069/12745262/ba70afae-c998-11e5-83dc-0ef092d5b82a.png)

RomainQuidet:
In case I just set the map region or center coordinate before setting / unsetting the insets, it works correctly.

[...]

PR #4067, opened 2016-02-22
[iOS] Using contentInset for setCenterCoordinate: and flyToCamera met…

RomainQuidet:
Using contentInset for setCenterCoordinate: and flyToCamera methods everywhere.

### DIFF
--- a/platform/ios/src/MGLMapView.mm
+++ b/platform/ios/src/MGLMapView.mm
@@ -3296,7 +3296,7 @@ mbgl::Duration MGLDurationInSeconds(NSTimeInterval duration)
 - (void)didUpdateLocationIncrementallyAnimated:(BOOL)animated
 {
     [self _setCenterCoordinate:self.userLocation.location.coordinate
-                   edgePadding:self.contentInset
+                   edgePadding:self.edgePaddingForFollowing
                      zoomLevel:self.zoomLevel
                      direction:self.directionByFollowingWithCourse
                       duration:animated ? MGLUserLocationAnimationDuration : 0
@@ -3323,7 +3323,7 @@ mbgl::Duration MGLDurationInSeconds(NSTimeInterval duration)
     
     __weak MGLMapView *weakSelf = self;
     [self _flyToCamera:camera
-           edgePadding:self.contentInset
+           edgePadding:self.edgePaddingForFollowing
           withDuration:animated ? -1 : 0
           peakAltitude:-1
      completionHandler:^{
--- a/platform/ios/src/MGLMapView.mm
+++ b/platform/ios/src/MGLMapView.mm
@@ -3375,17 +3375,23 @@ mbgl::Duration MGLDurationInSeconds(NSTimeInterval duration)
 /// Returns the edge padding to apply when moving the map to a tracked location.
 - (UIEdgeInsets)edgePaddingForFollowing
 {
-    // Center on user location unless we're already centered there (or very close).
-    CGPoint correctPoint = self.userLocationAnnotationViewCenter;
-    
-    // Shift the entire frame upward or downward to accommodate a shifted user
-    // location annotation view.
-    CGRect bounds = self.bounds;
-    CGRect boundsAroundCorrectPoint = CGRectOffset(bounds,
-                                                   correctPoint.x - CGRectGetMidX(bounds),
-                                                   correctPoint.y - CGRectGetMidY(bounds));
-    return UIEdgeInsetsMake(CGRectGetMinY(boundsAroundCorrectPoint) - CGRectGetMinY(bounds), 0,
-                            CGRectGetMaxY(bounds) - CGRectGetMaxY(boundsAroundCorrectPoint), 0);
+    UIEdgeInsets result;
+    if (self.userLocationVerticalAlignment == MGLAnnotationVerticalAlignmentCenter)
+    {
+        result = self.contentInset;
+    }
+    else
+    {
+        // Center on user location unless we're already centered there (or very close).
+        CGPoint correctPoint = self.userLocationAnnotationViewCenter;
+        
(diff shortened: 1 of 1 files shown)


1ec5:
If edgePaddingForFollowing doesn't include the content inset, it should.

RomainQuidet:
@1ec5 What do you mean it should? I thought the logic behind edgePaddingForFollowing was for a former usage, I simply replaced it by the contentInset as everywhere else in the mapView class. It's working according to my basic tests. 
If I must use this method, what is the logic behind it? Thanks.

1ec5:
`-edgePaddingForFollowing` should be used here, but its implementation is incorrect. The method is meant to account for the fact that the user dot may be somewhere other than the center of the map (even after accounting for content insets). For example, if you set `userLocationVerticalAlignment` to `MGLAnnotationVerticalAlignmentTop` or `MGLAnnotationVerticalAlignmentBottom`, this method should return quite a large top inset. However, [this line](https://github.com/mapbox/mapbox-gl-native/blob/7b40cbcfd949371e04944a23ffe08b69862f6a1a/platform/ios/src/MGLMapView.mm#L3384) is incorrect: we should start out with `contentFrame`, not `bounds`. `bounds` should only be used to calculate the final return value, since edge padding is always based on the view’s non-inset bounds.

RomainQuidet:
I agree that - (UIEdgeInsets)edgePaddingForFollowing implementation is wrong and a little bit strange for me. Nowhere it uses the userLocationVerticalAlignment...
I'll try to adapt if to return the correct insets.
Everywhere else the method _setCenterCoordinate is used, it simply pass the self.contentInset to the edgePadding parameter. (see https://github.com/mapbox/mapbox-gl-native/blob/7b40cbcfd949371e04944a23ffe08b69862f6a1a/platform/ios/src/MGLMapView.mm#L1763)

RomainQuidet:
correct edgePaddingForFollowing, using contentInset and vertical alignment.

boundsj:
> Everywhere else the method _setCenterCoordinate is used, it simply pass the self.contentInset to the edgePadding parameter.

I agree. It seems to me like the fix should actually be https://github.com/mapbox/mapbox-gl-native/pull/4067/commits/ccbdeaa451c7ec125e5b7b7af0f4eaa32ac8cce8, the original commit in this PR. 

This rect created in `userLocationAnnotationViewCenter` is based off of `contentFrame` and [already accounts for user location vertical alignment](https://github.com/mapbox/mapbox-gl-native/blob/master/platform/ios/src/MGLMapView.mm#L4704-L4713) which is the concern @1ec5 brought up above. The paths `[MGLMapView glkView:drawInRect:]` -> `[MGLMapView updateUserLocationAnnotationView]` ->  `[MGLMapView updateUserLocationAnnotationViewAnimatedWithDuration:]` and also `[MGLMapView locationManager:didUpdateLocations:]`  -> `[MGLMapView locationManager:didUpdateLocations:animated:]` -> `[MGLMapView updateUserLocationAnnotationViewAnimatedWithDuration:]` all end up using `userLocationAnnotationViewCenter` to get the `userPoint` used to position the user dot view via its `center` property.

Therefore, I think that `[MGLMapView _setCenterCoordinate:edgePadding:zoomLevel:direction:duration:animationTimingFunction:completionHandler:]` should expect the edgePadding to be the `contentInset` for the case of updating for contentInset changes like it does for all other cases.

**Note: I'm using an orange, semi-transparent UIView to visualize the `contentFrame` that is created [here](https://github.com/mapbox/mapbox-gl-native/blob/master/platform/ios/src/MGLMapView.mm#L4695).** I update the `contentInset` on a timer every 5 seconds. With only https://github.com/mapbox/mapbox-gl-native/pull/4067/commits/ccbdeaa451c7ec125e5b7b7af0f4eaa32ac8cce8 applied, adjusting the map's contentInset works like I would expect for the following cases:

### `MGLAnnotationVerticalAlignmentCenter`, contentInset oscillated between left=200 and left=0 on a loop:
contentinsetfix

### `MGLAnnotationVerticalAlignmentBottom`, contentInset oscillated between left=200 and left=0 on a loop:
contentinsetfixbottom

### `MGLAnnotationVerticalAlignmentTop`, contentInset oscillated between left=200 and left=0 on a loop:
contentinsetfixtop

If I use the implementation of `edgePaddingForFollowing` we have in the current release version or update the implementation of `edgePaddingForFollowing` as suggested in https://github.com/mapbox/mapbox-gl-native/pull/4067#issuecomment-187941664 the result is incorrect because the [implementation](https://github.com/mapbox/mapbox-gl-native/blob/7b40cbcfd949371e04944a23ffe08b69862f6a1a/platform/ios/src/MGLMapView.mm#L3384-L3388) of `edgePaddingForFollowing` results in `UIEdgeInsetsZero`. This causes the map to not move based on the `contentInset` input by the developer even though the user dot does move. This matches the [original issue report](https://github.com/mapbox/mapbox-gl-native/issues/3778):

> Then set the user tracking to follow. The map then "moves back" to its original position and only the user location is inset. 

contetinsetincorrect

In any case, there is no reason that I can see to use `edgePaddingForFollowing` since the user dot is moved as noted above and the map just needs the raw `contentInset` value. The final commit in this PR causes incorrect behavior when something other than `MGLAnnotationVerticalAlignmentCenter` is used.

I propose that we reduce this PR to just https://github.com/mapbox/mapbox-gl-native/pull/4067/commits/ccbdeaa451c7ec125e5b7b7af0f4eaa32ac8cce8 (the first commit) and merge it.

1ec5:
OK, I’m convinced. 😄 Thanks for the visual verification.

boundsj:
Subsumed by https://github.com/mapbox/mapbox-gl-native/pull/6216

1ec5:
This caused a regression that reproduces when the user dot is top- or bottom-aligned: #6257.

PR #4070, opened 2016-02-22
Fix crash when we get a HTTP 200 with no data

tmpsantos:
We assume `response->data` an empty string instead of a null pointer.


jfirebaugh:
:+1: Thanks for the catch.

Did you looks at linux, iOS/OSX, and node implementations too? (Looking forward to having the storage tests run for all of these platforms.)

bleege:
Awesome!  Thanks @tmpsantos!

tmpsantos:
Yes, they are all good. libcurl is creating a ` std::string()`. OSX/iOS and Node are constructing a `std::string(nullptr, 0)` which is also valid.

PR #4079, opened 2016-02-23
Fix iOS benchmark app

kkaefer:


platform/ios/benchmark/MBXBenchViewController.mm
```diff
@@ -1,6 +1,6 @@
 #import "MBXBenchViewController.h"
 
-#import <mbgl/ios/MGLMapView.h>
+#import "MGLMapView.h"
```
1ec5: If this depends on the iossdk target, which is a dynamic framework, you should be able to import the umbrella header, as `<Mapbox/Mapbox.h>`.

1ec5:
:+1:

platform/ios/benchmark/MBXBenchViewController.mm
kkaefer: Changed.

PR #4080, opened 2016-02-23
[node, core] Revert #3994

tmpsantos:
We cannot hold any references to the libuv's default main loop otherwise node won't exit. The prior implementation assumed that the gc() would be called frequently to collect the unused NodeMaps holding these references but this is simply not true.

/cc @jfirebaugh @mikemorris


tmpsantos:
Merged.

Issue #3704, opened 2016-01-27
Refactor networking in MGLMapboxEvents


boundsj:
MGLMapboxEvents manages instances of NSURLSession (and friends) in order to send events back to an events API. This issue is to refactor the networking concerns out of MGLMapboxEvents into a utility that MGLMapboxEvents (and other future classes) can be a client of.

Separating networking concerns from MGLMapboxEvents (including being a delegate for TLS callbacks and JSON serialization for the events themselves) should make the events class less complex and easier to add features to in the future. In addition, new features that require networking won't have to reach into events or roll their own networking implementations.

cc @1ec5 @friedbunny

PR #4081, opened 2016-02-23
Introduce API Client

boundsj:
This introduces a new utility class that wraps networking via NSURLSession. All related code that used to live inside the telemetry MGLMapboxEvents class has been pulled into the new MGLAPIClient. An API client instance is used as a service by telemetry and can be reused in the future when and if our networking needs grow or become more complex.

There are some nice side effects as a result of this refactor such as:

- The turnstile event can now be guaranteed to have a send attempt without having to be special cased and searched for inside a queue of other events
- We can react to failures (for now I've just added debug logging but in the future it would be nice to have a back off strategy when the network just doesn't work)
- Future work to make the MGLMapboxEvents class more testable will be easier since the networking client can be mocked and injected under test so tests are more controllable and determinate
- In flight requests can be cancelled

The main benefit of the api client is the API it presents to its clients. The public interface is:

```
- (void)postEvents:(nonnull NS_ARRAY_OF(MGLMapboxEventAttributes *) *)events completionHandler:(nullable void (^)(NSError * _Nullable error))completionHandler;
- (void)postEvent:(nonnull MGLMapboxEventAttributes *)event completionHandler:(nullable void (^)(NSError * _Nullable error))completionHandler;
- (void)cancelAll;
```

An example usage of this for sending the turnstile event specifically is:

```
[self.apiClient postEvent:eventAttributes completionHandler:^(NSError * _Nullable error) {
    if (error) { /* log error and return */ }
    [strongSelf writeEventToLocalDebugLog:turnstileEventAttributes];
}];
```

### DIFF
--- a/gyp/platform-ios.gypi
+++ b/gyp/platform-ios.gypi
@@ -46,6 +46,8 @@
         '../platform/darwin/src/MGLMapCamera.mm',
         '../platform/ios/src/MGLMapboxEvents.h',
         '../platform/ios/src/MGLMapboxEvents.m',
+        '../platform/ios/src/MGLAPIClient.h',
+        '../platform/ios/src/MGLAPIClient.m',
         '../platform/ios/src/MGLMapView.mm',
         '../platform/ios/src/MGLAccountManager_Private.h',
         '../platform/ios/src/MGLAccountManager.m',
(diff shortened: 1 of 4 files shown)


boundsj:
@1ec5, @friedbunny this is still a WIP but I wanted to get :eyes: on it sooner rather than later to get your feedback.

1ec5:
> in the future it would be nice to have a back off strategy when the network just doesn't work

We're already using Reachability to react to changing network conditions for mbgl-level requests. The Reachability object and callback are owned by MGLMapView, but there's nothing stopping this API client object from owning its own instance.

PR #4082, opened 2016-02-23
Introduce API Client

boundsj:
This introduces a new utility class that wraps networking via NSURLSession. All related code that used to live inside the telemetry MGLMapboxEvents class has been pulled into the new MGLAPIClient. An API client instance is used as a service by telemetry and can be reused in the future when and if our networking needs grow or become more complex.

There are some nice side effects as a result of this refactor such as:

- The turnstile event can now be guaranteed to have a send attempt without having to be special cased and searched for inside a queue of other events
- We can react to failures (for now I've just added debug logging but in the future it would be nice to have a back off strategy when the network just doesn't work)
- Future work to make the MGLMapboxEvents class more testable will be easier since the networking client can be mocked and injected under test so tests are more controllable and determinate
- In flight requests can be cancelled

The main benefit of the api client is the API it presents to its clients. The public interface is:

```
- (void)postEvents:(nonnull NS_ARRAY_OF(MGLMapboxEventAttributes *) *)events completionHandler:(nullable void (^)(NSError * _Nullable error))completionHandler;
- (void)postEvent:(nonnull MGLMapboxEventAttributes *)event completionHandler:(nullable void (^)(NSError * _Nullable error))completionHandler;
- (void)cancelAll;
```

An example usage of this for sending the turnstile event specifically is:

```
[self.apiClient postEvent:eventAttributes completionHandler:^(NSError * _Nullable error) {
    if (error) { /* log error and return */ }
    [self writeEventToLocalDebugLog:turnstileEventAttributes];
}];
```

### DIFF
--- a/platform/ios/src/MGLAPIClient.m
+++ b/platform/ios/src/MGLAPIClient.m
@@ -1,0 +1,191 @@
+#import "MGLAPIClient.h"
+#import "NSBundle+MGLAdditions.h"
+#import "MGLAccountManager.h"
+
+static NSString * const MGLAPIClientUserAgent = @"MapboxEventsiOS/1.1";
+static NSString * const MGLAPIClientBaseURL = @"https://api.tiles.mapbox.com";
+
+static NSString * const MGLAPIClientHeaderFieldUserAgentKey = @"User-Agent";
+static NSString * const MGLAPIClientHeaderFieldContentTypeKey = @"Content-Type";
+static NSString * const MGLAPIClientHeaderFieldContentTypeValue = @"application/json";
+static NSString * const MGLAPIClientHTTPMethodPost = @"POST";
+
+@interface MGLAPIClient ()
+
+@property (nonatomic, copy) NSURLSession *session;
+@property (nonatomic, copy) NSString *baseURL;
+@property (nonatomic, copy) NSData *digicertCert;
+@property (nonatomic, copy) NSData *geoTrustCert;
+@property (nonatomic, copy) NSData *testServerCert;
+@property (nonatomic, copy) NSString *userAgent;
+@property (nonatomic) NSMutableArray *dataTasks;
+@property (nonatomic) BOOL usesTestServer;
+
+@end
+
+@implementation MGLAPIClient
+
+- (instancetype)init {
+    self = [super init];
+    if (self) {
+        _session = [NSURLSession sessionWithConfiguration:[NSURLSessionConfiguration defaultSessionConfiguration]
+                                                 delegate:self delegateQueue:nil];
+        _dataTasks = [NSMutableArray array];
+        [self loadCertificates];
+        [self setupBaseURL];
+        [self setupUserAgent];
+    }
+    return self;
+}
+
+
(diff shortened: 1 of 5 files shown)


boundsj:
@1ec5 @friedbunny oops please look over here instead (based off of the wrong branch before). Still looking to get :eyes: on this so I can polish it up as required today.

platform/ios/src/MGLAPIClient.m
```diff
@@ -0,0 +1,191 @@
...
+
+#pragma mark Utilities
+
+- (NSURLRequest *)requestForEvents:(NS_ARRAY_OF(MGLMapboxEventAttributes *) *)events {
+    NSString *url = [NSString stringWithFormat:@"%@/events/v1?access_token=%@", self.baseURL, [MGLAccountManager accessToken]];
+    NSMutableURLRequest *request = [[NSMutableURLRequest alloc] initWithURL:[NSURL URLWithString:url]];
+    [request setValue:self.userAgent forHTTPHeaderField:MGLAPIClientHeaderFieldUserAgentKey];
+    [request setValue:MGLAPIClientHeaderFieldContentTypeValue forHTTPHeaderField:MGLAPIClientHeaderFieldContentTypeKey];
+    [request setHTTPMethod:MGLAPIClientHTTPMethodPost];
+    NSData *jsonData = [NSJSONSerialization dataWithJSONObject:events options:NSJSONWritingPrettyPrinted error:nil];
```
boundsj: oops. I forgot to remove the "pretty print" option. That does nothing for us since we are sending this data over they wire and not printing it for a human

platform/ios/src/MGLAPIClient.m
```diff
@@ -0,0 +1,191 @@
...
+
+- (void)cancelAll {
+    [self.dataTasks makeObjectsPerformSelector:@selector(cancel)];
+    [self.dataTasks removeAllObjects];
+}
+
+#pragma mark Utilities
+
+- (NSURLRequest *)requestForEvents:(NS_ARRAY_OF(MGLMapboxEventAttributes *) *)events {
+    NSString *url = [NSString stringWithFormat:@"%@/events/v1?access_token=%@", self.baseURL, [MGLAccountManager accessToken]];
```
friedbunny: With this refactor and the v2 telem endpoint coming up, it would be good to consolidate our `version` strings. If the `version = 1` [in MGLMapboxEvents](https://github.com/mapbox/mapbox-gl-native/pull/4082/files#diff-6e8e31cd0ffe52aeda9c8b2671bcea19R13) means the telemetry version, it should be the same as the telemetry API URL here.

boundsj: Good point. This was sort of discussed in https://github.com/mapbox/mapbox-gl-native/pull/3620 (telem diet), too. 

I just talked to @camilleanne and it turns out that the `version = 1` in mapbox events is no longer required because we no longer need to send that along in each event. I'm going to ignore it for now and come back to it when I work on telem diet / V2 events.

PR #4084, opened 2016-02-23
Remove dynamic_casts

kkaefer:
Since https://github.com/mapbox/mapbox-gl-native/commit/e04c5f650648d150e78da3602745b6190ec4aee3, we're using `dynamic_cast`s in a few areas, in particular the `SymbolLayer`. However, these are very slow, especially on iOS, eating up to 10% of our render time for Mapbox Streets. I think we should go back to storing an explicit enum.

/cc @jfirebaugh

### DIFF
--- a/src/mbgl/layer/background_layer.hpp
+++ b/src/mbgl/layer/background_layer.hpp
@@ -15,6 +15,7 @@ public:
 
 class BackgroundLayer : public StyleLayer {
 public:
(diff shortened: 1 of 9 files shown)


jfirebaugh:
:+1: -- nice solution.

tmpsantos:
:+1: nice indeed

Issue #4043, opened 2016-02-19
Update JNI bindings to use modern C++ practices


jfirebaugh:
The JNI bindings should:

* Use RAII class wrappers for all JNI resources, e.g. GlobalRefs, thread attachments, ...?
* Use method wrappers that do error checking and throw exceptions
* (stretch goal) Use type-safe method wrappers for `Call*Method` invocations

These wrappers should be designed to avoid the items on this list, of which we are guilty of several:

http://android-developers.blogspot.com/2011/11/jni-local-reference-changes-in-ics.html

bleege:
We'll also want to investigate how using JavaCPP could make this better or worse.

https://github.com/bytedeco/javacpp

/cc @zugaldia @tmpsantos

zugaldia:
I fully support this one and I'm happy to help with the effort.

> We'll also want to investigate how using JavaCPP could make this better or worse.

This is the library used by  OpenCV, FFmpeg, and Apple Music on Android, so it has a strong industry support.

I know @jfirebaugh is looking into a few more options. Just adding two more that have been discussed in the past at some point:
* [Djinni](https://github.com/dropbox/djinni), "a tool for generating cross-language type declarations and interface bindings", by Dropbox. It supports Java and Objective-C, and Python is experimental.
* [SWIG](http://www.swig.org), this one mentioned by @tmpsantos supporting a larger number of languages, not only Java.

tmpsantos:
Regardless of what tool we use, we should definitely try to generate the bindings. I know that not everybody is a big fan of SWIG, but I just mentioned it because I have done some quick python bindings with it and it worked just fine. The benefit is we would get other languages for free.

If we find something more specialized to Java that will deliver better interfaces and performance, I'm all for it too.

bleege:
Agreed all around.  The current setup of crafting artisanal JNI bindings is a huge PITA not to mention tends to lead to undesirable code.

FWIW, @kkaefer and I looked at Djinni last summer and passed on it.  Still, it'd be worth a second look to see what's changed since then.

jfirebaugh:
I don't think we should try to generate the bindings.

* Every binding generator I've seen produced reams of low-quality code, and resulted in bindings with bad ergonomics, bindings that don't follow the conventions of the target language, or bindings that leaked  underlying abstraction details.
[...]

PR #4085, opened 2016-02-23
[android] - Implements Android bindings for offline API

### LINKED ISSUE #3891: Android bindings for offline API
Previously: #584, #3715.

We now have an mbgl-core offline API for which Android bindings need to be written.

zugaldia:
Fixes #3891 

/cc: @jfirebaugh

### DIFF
--- a/platform/android/MapboxGLAndroidSDK/src/main/java/com/mapbox/mapboxsdk/offline/OfflineManager.java
+++ b/platform/android/MapboxGLAndroidSDK/src/main/java/com/mapbox/mapboxsdk/offline/OfflineManager.java
@@ -1,0 +1,184 @@
+package com.mapbox.mapboxsdk.offline;
+
+import android.content.Context;
+import android.os.Handler;
+import android.os.Looper;
+import android.support.annotation.NonNull;
+
+import java.io.File;
+
+/**
+ * The offline manager is the main entry point for offline-related functionality.
+ * It'll help you list and create offline regions.
+ */
+public class OfflineManager {
+
+    // Default database name
+    private final static String OFFLINE_DATABASE_NAME = "mbgl-offline.db";
+
+    /*
+     * The maximumCacheSize parameter is a limit applied to non-offline resources only,
+     * i.e. resources added to the database for the "ambient use" caching functionality.
+     * There is no size limit for offline resources.
+     */
+    private final static long DEFAULT_MAX_CACHE_SIZE = 50 * 1024 * 1024;
+
+    // Holds the pointer to JNI DefaultFileSource
+    private long mDefaultFileSourcePtr = 0;
+
+    // Makes sure callbacks come back to the main thread
+    private Handler handler;
+
+    // This object is implemented as a singleton
+    private static OfflineManager instance;
+
+    /*
+     * Callbacks
+     */
+
+    public interface ListOfflineRegionsCallback {
+        void onList(OfflineRegion[] offlineRegions);
+        void onError(String error);
+    }
+
+    public interface CreateOfflineRegionCallback {
+        void onCreate(OfflineRegion offlineRegion);
+        void onError(String error);
+    }
+
+    /*
+     * Constructors
+     */
+
+    private OfflineManager(Context context) {
+        // Get a pointer to the DefaultFileSource instance
+        String assetRoot = context.getFilesDir().getAbsolutePath();
+        String cachePath = assetRoot  + File.separator + OFFLINE_DATABASE_NAME;
+        mDefaultFileSourcePtr = createDefaultFileSource(cachePath, assetRoot, DEFAULT_MAX_CACHE_SIZE);
+    }
+
+    public static synchronized OfflineManager getInstance(Context context) {
+        if (instance == null) {
+            instance = new OfflineManager(context);
+        }
+
+        return instance;
+    }
+
+    /*
+     * Access token getter/setter
+     */
+    public void setAccessToken(String accessToken) {
+        setAccessToken(mDefaultFileSourcePtr, accessToken);
+    }
+
+    public String getAccessToken() {
+        return getAccessToken(mDefaultFileSourcePtr);
+    }
+
+    private Handler getHandler() {
+        if (handler == null) {
+            handler = new Handler(Looper.getMainLooper());
+        }
+
+        return handler;
+    }
+
+    /**
+     * Retrieve all regions in the offline database.
+     *
+     * The query will be executed asynchronously and the results passed to the given
+     * callback on the main thread.
+     */
+    public void listOfflineRegions(@NonNull final ListOfflineRegionsCallback callback) {
+        listOfflineRegions(mDefaultFileSourcePtr, new ListOfflineRegionsCallback() {
+            @Override
+            public void onList(final OfflineRegion[] offlineRegions) {
+                getHandler().post(new Runnable() {
+                    @Override
+                    public void run() {
+                        callback.onList(offlineRegions);
+                    }
+                });
+            }
+
+            @Override
+            public void onError(final String error) {
+                getHandler().post(new Runnable() {
+                    @Override
+                    public void run() {
+                        callback.onError(error);
+                    }
+                });
+            }
+        });
+    }
+
+    /**
+     * Create an offline region in the database.
+     *
+     * When the initial database queries have completed, the provided callback will be
+     * executed on the main thread.
+     *
+     * Note that the resulting region will be in an inactive download state; to begin
+     * downloading resources, call `OfflineRegion.setDownloadState(DownloadState.STATE_ACTIVE)`,
+     * optionally registering an `OfflineRegionObserver` beforehand.
+     */
+    public void createOfflineRegion(
+            @NonNull OfflineRegionDefinition definition,
+            @NonNull OfflineRegionMetadata metadata,
+            @NonNull final CreateOfflineRegionCallback callback) {
+
+        createOfflineRegion(mDefaultFileSourcePtr, definition, metadata, new CreateOfflineRegionCallback() {
+            @Override
+            public void onCreate(final OfflineRegion offlineRegion) {
+                getHandler().post(new Runnable() {
+                    @Override
+                    public void run() {
+                        callback.onCreate(offlineRegion);
+                    }
+                });
+            }
+
+            @Override
+            public void onError(final String error) {
+                getHandler().post(new Runnable() {
+                    @Override
+                    public void run() {
+                        callback.onError(error);
+                    }
+                });
+            }
+        });
+    }
+
+    /*
+    * Changing or bypassing this limit without permission from Mapbox is prohibited
+    * by the Mapbox Terms of Service.
+    */
+    public void setOfflineMapboxTileCountLimit(long limit) {
+        setOfflineMapboxTileCountLimit(mDefaultFileSourcePtr, limit);
+    }
+
+
+    /*
+     * Native methods
+     */
+
+    private native long createDefaultFileSource(
+            String cachePath, String assetRoot, long maximumCacheSize);
+
+    private native void setAccessToken(long defaultFileSourcePtr, String accessToken);
+    private native String getAccessToken(long defaultFileSourcePtr);
+
+    private native void listOfflineRegions(
+            long defaultFileSourcePtr, ListOfflineRegionsCallback callback);
+
+    private native void createOfflineRegion(
+            long defaultFileSourcePtr, OfflineRegionDefinition definition,
+            OfflineRegionMetadata metadata, CreateOfflineRegionCallback callback);
+
+    private native void setOfflineMapboxTileCountLimit(
+            long defaultFileSourcePtr, long limit);
+
+}
--- a/platform/android/MapboxGLAndroidSDK/src/main/java/com/mapbox/mapboxsdk/offline/OfflineRegion.java
+++ b/platform/android/MapboxGLAndroidSDK/src/main/java/com/mapbox/mapboxsdk/offline/OfflineRegion.java
@@ -1,0 +1,278 @@
+package com.mapbox.mapboxsdk.offline;
+
+import android.os.Handler;
+import android.os.Looper;
+import android.support.annotation.IntDef;
+import android.support.annotation.NonNull;
+import android.util.Log;
+
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+
+/**
+ * To use offline maps on mobile, you will first have to create an offline region.
+ * Use OfflineManager.createOfflineRegion() to create a new offline region.
+ */
+public class OfflineRegion {
+
+    private final static String LOG_TAG = "OfflineRegion";
+
+    // Parent OfflineManager
+    private OfflineManager offlineManager;
+
+    // Members
+    private long mId = 0;
+    private OfflineRegionDefinition mDefinition = null;
+    private OfflineRegionMetadata mMetadata = null;
+
+    // Holds the pointer to JNI OfflineRegion
+    private long mOfflineRegionPtr = 0;
+
+    // Makes sure callbacks come back to the main thread
+    private Handler handler;
+
+    /**
+     * A region can have a single observer, which gets notified whenever a change
+     * to the region's status occurs.
+     */
+    public interface OfflineRegionObserver {
+        /**
+         * Implement this method to be notified of a change in the status of an
+         * offline region. Status changes include any change in state of the members
+         * of OfflineRegionStatus.
+         *
+         * This method will be executed on the main thread.
+         */
+        void onStatusChanged(OfflineRegionStatus status);
+
+        /**
+         * Implement this method to be notified of errors encountered while downloading
+         * regional resources. Such errors may be recoverable; for example the implementation
+         * will attempt to re-request failed resources based on an exponential backoff
+         * algorithm, or when it detects that network access has been restored.
+         *
+         * This method will be executed on the main thread.
+         */
+        void onError(OfflineRegionError error);
+
+        /*
+         * Implement this method to be notified when the limit on the number of Mapbox
+         * tiles stored for offline regions has been reached.
+         *
+         * Once the limit has been reached, the SDK will not download further offline
+         * tiles from Mapbox APIs until existing tiles have been removed. Contact your
+         * Mapbox sales representative to raise the limit.
+         *
+         * This limit does not apply to non-Mapbox tile sources.
+         *
+         * This method will be executed on the main thread.
+         */
+        void mapboxTileCountLimitExceeded(long limit);
+    }
+
+    /*
+     * Callbacks
+     */
+
+    public interface OfflineRegionStatusCallback {
+        void onStatus(OfflineRegionStatus status);
+        void onError(String error);
+    }
+
+    public interface OfflineRegionDeleteCallback {
+        void onDelete();
+        void onError(String error);
+    }
+
+    /**
+     * A region is either inactive (not downloading, but previously-downloaded
+     * resources are available for use), or active (resources are being downloaded
+     * or will be downloaded, if necessary, when network access is available).
+     *
+     * This state is independent of whether or not the complete set of resources
+     * is currently available for offline use. To check if that is the case, use
+     * `OfflineRegionStatus.isComplete()`.
+     */
+
+    @IntDef({STATE_INACTIVE, STATE_ACTIVE})
+    @Retention(RetentionPolicy.SOURCE)
+    public @interface DownloadState {}
+
+    public static final int STATE_INACTIVE = 0;
+    public static final int STATE_ACTIVE = 1;
+
+    /*
+     * Constructor
+     */
+
+    private OfflineRegion() {
+        // For JNI use only, to create a new offline region, use
+        // OfflineManager.createOfflineRegion() instead.
+    }
+
+    /*
+     * Getters
+     */
+
+    public long getID() {
+        return mId;
+    }
+
+    public OfflineRegionDefinition getDefinition() {
+        return mDefinition;
+    }
+
+    public OfflineRegionMetadata getMetadata() {
+        return mMetadata;
+    }
+
+    private Handler getHandler() {
+        if (handler == null) {
+            handler = new Handler(Looper.getMainLooper());
+        }
+
+        return handler;
+    }
+
+    /**
+     * Register an observer to be notified when the state of the region changes.
+     */
+    public void setObserver(@NonNull final OfflineRegionObserver observer) {
+        setOfflineRegionObserver(this, new OfflineRegionObserver() {
+            @Override
+            public void onStatusChanged(final OfflineRegionStatus status) {
+                getHandler().post(new Runnable() {
+                    @Override
+                    public void run() {
+                        observer.onStatusChanged(status);
+                    }
+                });
+            }
+
+            @Override
+            public void onError(final OfflineRegionError error) {
+                getHandler().post(new Runnable() {
+                    @Override
+                    public void run() {
+                        observer.onError(error);
+                    }
+                });
+            }
+
+            @Override
+            public void mapboxTileCountLimitExceeded(final long limit) {
+                getHandler().post(new Runnable() {
+                    @Override
+                    public void run() {
+                        observer.mapboxTileCountLimitExceeded(limit);
+                    }
+                });
+            }
+        });
+    }
+
+    /**
+     * Pause or resume downloading of regional resources.
+     */
+    public void setDownloadState(@DownloadState int state) {
+        setOfflineRegionDownloadState(this, state);
+    }
+
+    /**
+     * Retrieve the current status of the region. The query will be executed
+     * asynchronously and the results passed to the given callback which will be
+     * executed on the main thread.
+     */
+    public void getStatus(@NonNull final OfflineRegionStatusCallback callback) {
+        getOfflineRegionStatus(this, new OfflineRegionStatusCallback() {
+            @Override
+            public void onStatus(final OfflineRegionStatus status) {
+                getHandler().post(new Runnable() {
+                    @Override
+                    public void run() {
+                        callback.onStatus(status);
+                    }
+                });
+            }
+
+            @Override
+            public void onError(final String error) {
+                getHandler().post(new Runnable() {
+                    @Override
+                    public void run() {
+                        callback.onError(error);
+                    }
+                });
+            }
+        });
+    }
+
+    /**
+     * Remove an offline region from the database and perform any resources evictions
+     * necessary as a result.
+     *
+     * Eviction works by removing the least-recently requested resources not also required
+     * by other regions, until the database shrinks below a certain size.
+     *
+     * When the operation is complete or encounters an error, the given callback will be
+     * executed on the main thread.
+     */
+    public void delete(@NonNull final OfflineRegionDeleteCallback callback) {
+        deleteOfflineRegion(this, new OfflineRegionDeleteCallback() {
+            @Override
+            public void onDelete() {
+                getHandler().post(new Runnable() {
+                    @Override
+                    public void run() {
+                        callback.onDelete();
+                        OfflineRegion.this.finalize();
+                    }
+                });
+            }
+
+            @Override
+            public void onError(final String error) {
+                getHandler().post(new Runnable() {
+                    @Override
+                    public void run() {
+                        callback.onError(error);
+                    }
+                });
+            }
+        });
+    }
+
+    @Override
+    protected void finalize() {
+        try {
+            super.finalize();
+            destroyOfflineRegion(mOfflineRegionPtr);
+            mOfflineRegionPtr = 0;
+        } catch (Throwable throwable) {
+            Log.e(LOG_TAG, "Failed to finalize OfflineRegion: " + throwable.getMessage());
+        }
+    }
+
+    /*
+     * Native methods
+     */
+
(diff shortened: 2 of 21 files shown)


platform/android/MapboxGLAndroidSDK/src/main/java/com/mapbox/mapboxsdk/offline/OfflineManager.java
```diff
@@ -0,0 +1,172 @@
...
+
+/**
+ * The offline manager is the main entry point for offline-related functionality.
+ * It'll help you list and create offline regions.
+ */
+public class OfflineManager {
+
+    // Default database name and size (1GB)
+    private final static String OFFLINE_DATABASE_NAME = "mbgl-offline.db";
+    private final static long OFFLINE_MAX_CACHE_SIZE = 1024 * 1024 * 1024;
```
jfirebaugh: FYI: this limit is applied to _non-offline_ resources only, i.e. resources added to the database for the "ambient use" caching functionality. There is no size limit for offline resources. So 1GB is probably too high; if a user never creates any offline regions, we want the database to remain fairly small (order tens or low hundreds of megabytes).

platform/android/MapboxGLAndroidSDK/src/main/java/com/mapbox/mapboxsdk/offline/OfflineManager.java
```diff
@@ -0,0 +1,172 @@
...
+            handler = new Handler(Looper.getMainLooper());
+        }
+
+        return handler;
+    }
+
+    /**
+     * Retrieve all regions in the offline database.
+     *
+     * The query will be executed asynchronously and the results passed to the given
```
jfirebaugh: You should tailor the documentation comments to be reflective of the interface exposed by the SDK bindings, rather than the core mbgl interface. So in this case, this note about threading details is not relevant; it's an implementation detail between the core interface and the SDK binding implementation.

platform/android/MapboxGLAndroidSDK/src/main/java/com/mapbox/mapboxsdk/offline/OfflineManager.java
```diff
@@ -0,0 +1,172 @@
...
+
+    /**
+     * Create an offline region in the database.
+     *
+     * When the initial database queries have completed, the provided callback will be
+     * executed on the database thread; it is the responsibility of the SDK bindings
+     * to re-execute a user-provided callback on the main thread.
+     *
+     * Note that the resulting region will be in an inactive download state; to begin
+     * downloading resources, call `setOfflineRegionDownloadState(OfflineRegionDownloadState::Active)`,
```
jfirebaugh: Translate this into Android SDK APIs.

platform/android/MapboxGLAndroidSDK/src/main/java/com/mapbox/mapboxsdk/offline/OfflineRegion.java
```diff
@@ -0,0 +1,247 @@
...
+import android.os.Handler;
+import android.os.Looper;
+import android.support.annotation.IntDef;
+import android.support.annotation.NonNull;
+
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+
+/**
+ * For the present, a tile pyramid is the only type of offline region. In the future,
```
jfirebaugh: This is the description for `OfflineRegionDefinition`, rather than `OfflineRegion`.

platform/android/MapboxGLAndroidSDK/src/main/java/com/mapbox/mapboxsdk/offline/OfflineRegionDefinition.java
```diff
@@ -0,0 +1,65 @@
...
+ * device pixel ratio.
+ *
+ * Both minZoom and maxZoom must be ≥ 0, and maxZoom must be ≥ minZoom.
+ *
+ * maxZoom may be ∞, in which case for each tile source, the region will include
+ * tiles from minZoom up to the maximum zoom level provided by that source.
+ *
+ * pixelRatio must be ≥ 0 and should typically be 1.0 or 2.0.
+ */
+public class OfflineRegionDefinition {
```
jfirebaugh: Let's structure the Java class definitions so that we can introduce new region definition types in the future without breaking code that creates the current type of region definition.

```
public interface OfflineRegionDefinition {}

class LatLngBoundsOfflineRegionDefinition implements OfflineRegionDefinition {
  ...
}
```

platform/android/MapboxGLAndroidSDK/src/main/java/com/mapbox/mapboxsdk/offline/OfflineRegionError.java
```diff
@@ -0,0 +1,47 @@
+package com.mapbox.mapboxsdk.offline;
+
+/**
+ * Created by antonio on 2/21/16.
+ */
+public class OfflineRegionError {
+
+    /**
+     * Success = 1
```
jfirebaugh: I don't think we want to make these specific integer values (which ultimately come from [here](https://github.com/mapbox/mapbox-gl-native/blob/c91b44b02f658548bdc70a625eb54d0ea86f6737/include/mbgl/storage/response.hpp#L41-L45)) part of the public API. What would be an idiomatic way to expose these error types on Android?

platform/android/MapboxGLAndroidSDK/src/main/java/com/mapbox/mapboxsdk/offline/OfflineRegionMetadata.java
```diff
@@ -0,0 +1,65 @@
...
+import java.io.ObjectOutputStream;
+
+/**
+ * Arbitrary binary region metadata. The contents are opaque to the mbgl implementation;
+ * it just stores and retrieves a BLOB. SDK bindings should leave the interpretation of
+ * this data up to the application; they _should not_ enforce a higher-level data format.
+ * In the future we want offline database to be portable across target platforms, and a
+ * platform-specific metadata format would prevent that.
+ */
+public class OfflineRegionMetadata {
```
jfirebaugh: Is it worth having a class for this? Versus just `byte[] OfflineRegion::getMetadata()`?

platform/android/MapboxGLAndroidSDK/src/main/java/com/mapbox/mapboxsdk/offline/OfflineRegionStatus.java
```diff
@@ -0,0 +1,83 @@
...
+
+    private OfflineRegionStatus() {
+        // For JNI use only
+    }
+
+    /*
+     * Is the region complete?
+     */
+
+    public boolean complete() {
```
jfirebaugh: `isComplete`?

platform/android/MapboxGLAndroidSDK/src/main/java/com/mapbox/mapboxsdk/offline/OfflineRegionStatus.java
```diff
@@ -0,0 +1,83 @@
...
+
+    public long getCompletedResourceSize() {
+        return completedResourceSize;
+    }
+
+    public long getRequiredResourceCount() {
+        return requiredResourceCount;
+    }
+
+    public boolean isRequiredResourceCountIsIndeterminate() {
```
jfirebaugh: There's a double "is" in this method name, but I also think that the mbgl core property could be better named. Right now it is named in the negative, and names in the positive are almost always better. How about I rename the core property to `requiredResourceCountIsPrecise` (with a reversed sense), and this method can be named `isRequiredResourceCountPrecise`?

platform/android/src/jni.cpp
```diff
@@ -373,6 +435,77 @@ std::pair<mbgl::AnnotationSegment, mbgl::ShapeAnnotation::Properties> annotation
...
+
+void jniExceptionCheck(JNIEnv * env) {
+    if (!env) {
+        mbgl::Log::Error(mbgl::Event::JNI, "Empty environment.");
+        abort();
+    }
+
+    const jthrowable e(env->ExceptionOccurred());
+    if (e) {
+        env->ExceptionClear();
```
jfirebaugh: This doesn't look like the desired semantics given how this function is used. This definition will clear any pending Java exception and allow C++ code that calls it to continue as if nothing has gone wrong.

In most cases, the desired behavior if a Java exception is detected is that C++ code unwinds, without doing further work, back to the JVM boundary. And then after returning to JVM control, Java exception handling takes over, processing the still-pending Java exception.

This can be accomplished by throwing an exception of a specific type, say `PendingJavaException`, here, and catching and discarding it in all native binding methods just before returning to JVM control. This is the approach that the JNI library I'm writing will take.

platform/android/src/jni.cpp
```diff
@@ -373,6 +435,77 @@ std::pair<mbgl::AnnotationSegment, mbgl::ShapeAnnotation::Properties> annotation
...
+        };
+
+        std::unique_ptr<uint8_t, decltype(deleter)> ptr(
+            reinterpret_cast<uint8_t*>(jniEnv->GetPrimitiveArrayCritical(j, nullptr)),
+            deleter
+        );
+
+        if (ptr) {
+            // Construct and then move-assign. This copies the elements only once,
+            // and avoids having to initialize before filling (as with resize())
```
jfirebaugh: Don't worry about the overhead of `resize()`. `resize()` + `GetByteArrayRegion` is preferable to `GetPrimitiveArrayCritical` / `ReleasePrimitiveArrayCritical`, and the compiler likely optimizes `std::vector< uint8_t>` initialization to a `memset` anyway.

platform/android/src/jni.cpp
```diff
@@ -373,6 +435,77 @@ std::pair<mbgl::AnnotationSegment, mbgl::ShapeAnnotation::Properties> annotation
...
+    mbgl::Log::Error(mbgl::Event::JNI, "jbyteArray conversion completed.");
+    return ret;
+}
+
+static jbyteArray metadata_from_native(JNIEnv* jniEnv, const std::vector<uint8_t>& c) {
+    assert(c.size() <= std::numeric_limits<jsize>::max());
+    jbyteArray j = jniEnv->NewByteArray(static_cast<jsize>(c.size()));
+    jniExceptionCheck(jniEnv);
+
+    // Using .data() on an empty vector is UB
```
jfirebaugh: This is not true. From http://en.cppreference.com/w/cpp/container/vector/data:

> The pointer is such that range `[data(); data() + size())` is always a valid range, even if the container is empty.

platform/android/MapboxGLAndroidSDK/src/main/java/com/mapbox/mapboxsdk/offline/OfflineManager.java
zugaldia: Gotcha. Clarified and corrected with https://github.com/mapbox/mapbox-gl-native/commit/3e4f8f758ec2e424c7b1a9cf9897703298183aec, I'm using the default value in `src/mbgl/util/constants.cpp`.

platform/android/src/jni.cpp
```diff
@@ -1620,6 +1753,363 @@ void JNICALL nativeRemoveCustomLayer(JNIEnv *env, jobject obj, jlong nativeMapVi
...
+
+    // Checks
+    assert(defaultFileSourcePtr != 0);
+    if (listCallback == nullptr) {
+        mbgl::Log::Error(mbgl::Event::JNI, "Callback has to be set.");
+        return;
+    }
+
+    // Makes sure the callback doesn't get GC'ed
+    listCallback = reinterpret_cast<jobject>(env->NewGlobalRef(listCallback));
```
jfirebaugh: This needs to be released at some point.

platform/android/MapboxGLAndroidSDK/src/main/java/com/mapbox/mapboxsdk/offline/OfflineManager.java
zugaldia: Addressed by https://github.com/mapbox/mapbox-gl-native/commit/0e37aa2d703b2d7f2a79ce987d00ea2d524defe5.

platform/android/MapboxGLAndroidSDK/src/main/java/com/mapbox/mapboxsdk/offline/OfflineManager.java
zugaldia: Addressed by https://github.com/mapbox/mapbox-gl-native/commit/0e37aa2d703b2d7f2a79ce987d00ea2d524defe5.

platform/android/src/jni.cpp
```diff
@@ -1620,6 +1753,363 @@ void JNICALL nativeRemoveCustomLayer(JNIEnv *env, jobject obj, jlong nativeMapVi
...
+        if (renderDetach) {
+            mbgl::Log::Debug(mbgl::Event::JNI, "Attached.");
+        }
+
+        if (error) {
+            std::string message = mbgl::util::toString(error);
+            env2->CallVoidMethod(listCallback, listOnErrorMethodId, std_string_to_jstring(env2, message));
+        } else if (regions) {
+            // Build the OfflineManager object
+            jobject jmanager = env2->NewObject(offlineManagerClass, offlineManagerClassConstructorId);
```
jfirebaugh: Why is this allocating a new manager? Shouldn't it be reusing the `OfflineManager` on which `listOfflineRegions()` was called?

platform/android/src/jni.cpp
```diff
@@ -1620,6 +1753,363 @@ void JNICALL nativeRemoveCustomLayer(JNIEnv *env, jobject obj, jlong nativeMapVi
...
+            env2->SetLongField(jmanager, offlineManagerClassPtrId, defaultFileSourcePtr);
+
+            // Build jobjectArray
+            jsize index = 0;
+            jobjectArray jregions = env2->NewObjectArray(regions->size(), offlineRegionClass, NULL);
+            for (std::vector<mbgl::OfflineRegion>::iterator region = regions->begin(); region != regions->end(); ++region) {
```
jfirebaugh: `for (auto& region : regions)`

platform/android/src/jni.cpp
```diff
@@ -1620,6 +1753,363 @@ void JNICALL nativeRemoveCustomLayer(JNIEnv *env, jobject obj, jlong nativeMapVi
...
+                // Build the Region object
+                jobject jregion = env2->NewObject(offlineRegionClass, offlineRegionConstructorId);
+                env2->SetObjectField(jregion, offlineRegionOfflineManagerId, jmanager);
+                env2->SetLongField(jregion, offlineRegionIdId, region->getID());
+                env2->SetObjectField(jregion, offlineRegionMetadataId, metadata_from_native(env2, region->getMetadata()));
+
+                // Moves the region on the stack into a heap-allocated one
+                env2->SetLongField(jregion, offlineRegionPtrId,
+                    reinterpret_cast<jlong>(new mbgl::OfflineRegion(std::move(*region))));
```
jfirebaugh: `OfflineRegion` needs to define a `finalize` method that deallocates this.

platform/android/MapboxGLAndroidSDK/src/main/java/com/mapbox/mapboxsdk/offline/OfflineRegionStatus.java
zugaldia: This is an IDE artifact. The getter generated for a boolean `requiredResourceCountIsIndeterminate` is by default the name of the field prefixed by `is`. I'm gonna rename the method already to make sure the API doesn't change independently of when the core property gets renamed: https://github.com/mapbox/mapbox-gl-native/commit/36a47ae9f009544e312db101de837ece73f0526e.

platform/android/src/jni.cpp
```diff
@@ -1620,6 +1753,363 @@ void JNICALL nativeRemoveCustomLayer(JNIEnv *env, jobject obj, jlong nativeMapVi
...
+    // Definition fields
+    jstring jStyleURL = reinterpret_cast<jstring>(env->GetObjectField(definition_, offlineRegionDefinitionStyleURLId));
+    std::string styleURL = std_string_from_jstring(env, jStyleURL);
+    jobject jBounds = env->GetObjectField(definition_, offlineRegionDefinitionBoundsId);
+    jdouble jMinZoom = env->GetDoubleField(definition_, offlineRegionDefinitionMinZoomId);
+    jdouble jMaxZoom = env->GetDoubleField(definition_, offlineRegionDefinitionMaxZoomId);
+    jfloat jPixelRatio = env->GetFloatField(definition_, offlineRegionDefinitionPixelRatioId);
+
+    // Bounds fields
+    jdouble mLatNorth = env->GetDoubleField(jBounds, latLngBoundsLatNorthId);
```
jfirebaugh: We should extract a method that converts Java `LatLngBounds` to C++ `LatLngBounds` for use here and in `nativeGetAnnotationsInBounds`.

platform/android/MapboxGLAndroidSDK/src/main/java/com/mapbox/mapboxsdk/offline/OfflineRegionMetadata.java
zugaldia: My thinking to keep this independent is twofold: 1) let devs use the class as a base class for their higher level implementation, and 2) provide some utils to help with the `byte[]` conversion (see `serialize` and `deserialize`).

platform/android/MapboxGLAndroidSDK/src/main/java/com/mapbox/mapboxsdk/offline/OfflineRegionStatus.java
zugaldia: Done.

platform/android/MapboxGLAndroidSDK/src/main/java/com/mapbox/mapboxsdk/offline/OfflineRegion.java
zugaldia: Good catch, I updated the docs in https://github.com/mapbox/mapbox-gl-native/commit/a15175c46e98c4d2cdf53c94180ab0f6edc2e1d3.

platform/android/MapboxGLAndroidSDK/src/main/java/com/mapbox/mapboxsdk/offline/OfflineRegionDefinition.java
zugaldia: Sounds good, this is done in https://github.com/mapbox/mapbox-gl-native/commit/a15175c46e98c4d2cdf53c94180ab0f6edc2e1d3. I took a note on the JNI side (it doesn't affect the higher level API), that part can be addressed when we have more than one definition and https://github.com/mapbox/mapbox-gl-native/issues/4043 is up and running.

platform/android/MapboxGLAndroidSDK/src/main/java/com/mapbox/mapboxsdk/offline/OfflineRegionError.java
zugaldia: The idiomatic version in Android is to use the `@IntDef` annotation like I had used for `OfflineRegion.DownloadState` (we try to avoid enums at the SDK level). Changes are in https://github.com/mapbox/mapbox-gl-native/commit/9eb74359f07da2f13807a807f043f55b47f68137.

platform/android/src/jni.cpp
zugaldia: Right: https://github.com/mapbox/mapbox-gl-native/commit/a0ca0bd2a61935f56803b96c78cf6ccb7e1f81b8.

platform/android/src/jni.cpp
zugaldia: Done, though it required a few more changes (`regions` is wrapped by a `mbgl::optional`): https://github.com/mapbox/mapbox-gl-native/commit/86096b728bb14d2e9d64b9958b3cbe0a8dbdc68b.

platform/android/src/jni.cpp
zugaldia: @jfirebaugh: `mbgl::OfflineRegion`, right?

platform/android/src/jni.cpp
zugaldia: @jfirebaugh Would this https://github.com/mapbox/mapbox-gl-native/commit/f67efbf701829ae7ab121e4f7bcfcf7f497de4bc approach would work?

platform/android/MapboxGLAndroidSDK/src/main/java/com/mapbox/mapboxsdk/offline/OfflineRegionError.java
jfirebaugh: My concern was more about tying these to specific integers that must match `Response::Error::Reason`. We are fairly likely to change `Response::Error::Reason` (to remove the current gap in numbering, for instance), which would break Java code that compared with, say, `REASON_OTHER`.

platform/android/MapboxGLAndroidSDK/src/main/java/com/mapbox/mapboxsdk/offline/OfflineRegionMetadata.java
jfirebaugh: :+1:

platform/android/src/jni.cpp
jfirebaugh: No, the Java `OfflineRegion`. Wherever you have a Java instance with ownership of a corresponding C++ instance, as you do here with `OfflineRegion` and `mbgl::OfflineRegion`, the Java instance needs to release the C++ instance in its finalizer.

The other place this applies is `OfflineManager`, which owns a native reference to a `mbgl::DefaultFileSource`. It's less of a concern there though since `OfflineManager` is a singleton.

platform/android/src/jni.hpp
```diff
@@ -103,6 +103,62 @@ extern jclass httpRequestClass;
 extern jmethodID httpRequestStartId;
 extern jmethodID httpRequestCancelId;
 
+// Offline declarations start
+
+extern jclass offlineManagerClass;
+extern jmethodID offlineManagerClassConstructorId;
```
jfirebaugh: This is now unused.

platform/android/MapboxGLAndroidSDK/src/main/java/com/mapbox/mapboxsdk/offline/OfflineRegion.java
```diff
@@ -0,0 +1,241 @@
...
+    }
+
+    /**
+     * Remove an offline region from the database and perform any resources evictions
+     * necessary as a result.
+     *
+     * Eviction works by removing the least-recently requested resources not also required
+     * by other regions, until the database shrinks below a certain size.
+     *
+     * Note that this method takes ownership of the input, reflecting the fact that once
```
jfirebaugh: This sentence isn't relevant to the Android bindings.

platform/android/src/jni.cpp
jfirebaugh: Another wrinkle with the above is that `OfflineRegion.delete(OfflineRegionDeleteCallback)` puts the region into a state where it is invalid to call further methods on it, because the native side releases ownership of the `mbgl::OfflineRegion` ([here](https://github.com/mapbox/mapbox-gl-native/blob/f67efbf701829ae7ab121e4f7bcfcf7f497de4bc/platform/android/src/jni.cpp#L2085)). This has a couple of implications:

* `deleteOfflineRegion` (the native method) should reset the value of the `mOfflineRegionPtr` field to 0.
* Other `OfflineRegion` methods should raise an exception if the value of `mOfflineRegionPtr` is 0, in case someone tries to call the method after calling `OfflineRegion.delete`.

platform/android/src/jni.cpp
jfirebaugh: Just spent some time looking into this and I think that is the best approach for now. Please do carry it through to the other locations.

platform/android/src/jni.cpp
zugaldia: I have simplified this code in https://github.com/mapbox/mapbox-gl-native/commit/7f7f492219c29e800d190d54aa656c984a917538 and I found the cause of the crash, it was actually in a different spot where I was overwriting the actual class, not its `byte[]` object.

platform/android/src/jni.cpp
zugaldia: I've removed this code that was coming from Dropbox's djinni and redone it in https://github.com/mapbox/mapbox-gl-native/commit/7f7f492219c29e800d190d54aa656c984a917538.

platform/android/src/jni.cpp
zugaldia: As before, this has been rewritten: https://github.com/mapbox/mapbox-gl-native/commit/7f7f492219c29e800d190d54aa656c984a917538. Comments much appreciated.

jfirebaugh:
Remaining tasks:

* [ ] Resolve https://github.com/mapbox/mapbox-gl-native/pull/4085#discussion_r54009098
* [ ] Carry https://github.com/mapbox/mapbox-gl-native/pull/4085#discussion_r53966553 through to similar cases.
* [ ] Resolve https://github.com/mapbox/mapbox-gl-native/pull/4085#discussion_r53855426
* [ ] Resolve https://github.com/mapbox/mapbox-gl-native/pull/4085#discussion_r53855960
* [ ] Resolve https://github.com/mapbox/mapbox-gl-native/pull/4085#discussion_r54012466
* [ ] Resolve https://github.com/mapbox/mapbox-gl-native/pull/4085#discussion_r54012836
* [ ] Rebase on master and:
 * [ ] Handle [this rename](https://github.com/mapbox/mapbox-gl-native/commit/65c06e62707a9982dedf39a88904bce04df1e227)
 * [ ] Add bindings for https://github.com/mapbox/mapbox-gl-native/commit/065cda83fab02aea85573d9c2d78e9121296582d

platform/android/MapboxGLAndroidSDK/src/main/java/com/mapbox/mapboxsdk/offline/OfflineRegionError.java
zugaldia: Gotcha. How about https://github.com/mapbox/mapbox-gl-native/commit/8c1ad36249fddfae316f3be0b0749019110c36fd?

platform/android/src/jni.cpp
zugaldia: Ready for your :eyes: in https://github.com/mapbox/mapbox-gl-native/commit/2fe6aa121c80f6d977800b14416ddca1b2d95c27.

platform/android/src/jni.hpp
zugaldia: Removed in https://github.com/mapbox/mapbox-gl-native/commit/4e1101d967f999b5bb35c397721f9649416ba938.

platform/android/MapboxGLAndroidSDK/src/main/java/com/mapbox/mapboxsdk/offline/OfflineRegion.java
zugaldia: Removed in https://github.com/mapbox/mapbox-gl-native/commit/21b52c6d7c376d6e7e14e5b10fcdb40187c5f868.

zugaldia:
> Handle [this rename](https://github.com/mapbox/mapbox-gl-native/commit/65c06e62707a9982dedf39a88904bce04df1e227)

Handled here: https://github.com/mapbox/mapbox-gl-native/commit/ef6a8fbc34c39cc4b28ec4f9c5b655d1f2478c9b

zugaldia:
> Add bindings for https://github.com/mapbox/mapbox-gl-native/commit/065cda83fab02aea85573d9c2d78e9121296582d

Added here https://github.com/mapbox/mapbox-gl-native/commit/9c96e26f94d10ddcce9f97a0c46a1d0a55782d60.

platform/android/src/jni.cpp
zugaldia: Method extracted in https://github.com/mapbox/mapbox-gl-native/commit/0cc6a9b68668a9d10d645fc389233435b074a382.

platform/android/MapboxGLAndroidSDK/src/main/java/com/mapbox/mapboxsdk/offline/OfflineRegionError.java
jfirebaugh: That still depends on the definition of `Response::Error::Reason`. Using numeric values in the Java interface is fine. They key point is that we need to use _identifiers_ like `Response::Error::Reason::NotFound` to translate to the numeric values that the Java API uses, not just rely on the _values_ of those identifiers lining up. [Same idea as what you do here.](https://github.com/mapbox/mapbox-gl-native/blob/0cc6a9b68668a9d10d645fc389233435b074a382/platform/android/src/jni.cpp#L2036-L2044)

platform/android/src/jni.cpp
zugaldia: Is something like https://github.com/mapbox/mapbox-gl-native/commit/96883fccf2a00af25b9ba9452ebed6e796de27e6 what you have in mind? I followed the pattern we use for the native map view.

jfirebaugh: Perfect.

PR #4087, opened 2016-02-24
[core] Offline tweaks

jfirebaugh:
See commits for details.

:eyes: @zugaldia @1ec5

Note that this includes a small schema change -- if you have previously generated an offline database, you should delete and rebuild it with the new code (cc @tmpsantos). This is (hopefully) the last schema change before release. After release, we'll absolutely have to use migrations instead. It will be painful.


tmpsantos:
Thanks for the heads-up.

zugaldia:
All changes SGTM. I brought the more descriptive comment about `maximumCacheSize` to the Android binding too. Otherwise, bindings are ready to handle this change without API breaks (we'll still need to implement the extra `tileCountLimitExceeded` on the observer class.

PR #4107, opened 2016-02-24
[ios] Introduce API Client

boundsj:

This introduces a new utility class that wraps networking via
NSURLSession. All related code that used to live inside the telemetry
MGLMapboxEvents class has been pulled into the new MGLAPIClient.
An API client instance is used as a service by telemetry and can be
reused in the future when and if our networking needs grow or become
more complex.


1ec5:
:+1:

boundsj:
Closed with https://github.com/mapbox/mapbox-gl-native/commit/9a97a4d34f03e2c6266ae9b38c4dacda342520e1

Issue #4030, opened 2016-02-19
Ensure that location update validation is called when host app is going to background


boundsj:
In https://github.com/mapbox/mapbox-gl-native/pull/3671 we added logic to turn off the telemetry location manager the when the host application is in the background and the user has granted `when in use` permissions. This logic (in the `validateUpdatingLocation` implementation) is currently triggered when the host app modifies `NSUserDefaults` and when it implicitly calls `wakeGL:` in MGLMapView when it comes into the foreground.

However, there is currently nothing in place that ensures the call to `validateUpdatingLocation` when an app goes into the background.

We should ensure that `validateUpdatingLocation` is called when the host application goes into the background.

Blocked on https://github.com/mapbox/mapbox-gl-native/issues/3717 where refactoring will be done to simplify the events class. After the refactoring is done, we can consider monitoring application foreground/background events and take the appropriate action to fix this issue.

cc @1ec5 @friedbunny

boundsj:
Addressed by https://github.com/mapbox/mapbox-gl-native/pull/4115

Issue #3916, opened 2016-02-12
Stop monitoring for and sending visits and associated events.


boundsj:
CLLocation visits are currently tracked as telemetry events. Visits are not valuable as events so we should remove them.

boundsj:
Addressed by https://github.com/mapbox/mapbox-gl-native/pull/4115

Issue #1424, opened 2015-05-05
Use Geofencing To Reduce Power Usages


bleege:
Monitor device velocity to make an educated guess as to when it's in a relatively static location so that more battery intensive location services (GPS) can be switched off to save device power.  When the device leaves a specific radius (aka "geofence") re-enable location services so that the more accurate location services (GPS) can be re-enabled for map building.

The Geofences can be built using iOS's [Region Monitoring API](https://developer.apple.com/library/ios/documentation/UserExperience/Conceptual/LocationAwarenessPG/RegionMonitoring/RegionMonitoring.html#//apple_ref/doc/uid/TP40009497-CH9-SW1).

@mapbox/mobile @mapbox/mobiledata

friedbunny:
I'll be really interested to see which strategy is the best for determining stops while preserving battery. Simple coordinate/time comparison? [Core Motion activity API](https://developer.apple.com/library/ios/documentation/CoreMotion/Reference/CMMotionActivity_class/index.html)? Direct accelerometer access? Significant change API?

sbma44:
@friedbunny we've spoken to some experienced mobile devs who said that they keep a running figure for user velocity. Their use case/user experience emphasizes detecting stops quickly, though, so it's possible we will be able to do something simpler or better.

bleege:
Moving to Beta 3 to focus on map UI features in the short term.

bleege:
So this has been now officially moved from the iOS Beta 3 to iOS 3.2.0 release `#BetterLateThanNever`

boundsj:
Addressed by https://github.com/mapbox/mapbox-gl-native/pull/4115

PR #4115, opened 2016-02-26
Introduce Location Manager Utility and Optimize Background Telemetry

boundsj:
##### Background

When Mapbox telemetry is used in a host application that has background.location permissions AND `always` permissions granted by the user, the location manager utility will attempt to gather anonymous location data in the background to [improve map quality](https://www.mapbox.com/telemetry). Previously, the Mapbox SDK did little to optimize background data collection for the case of stationary devices. Running location services in the background, even in a passive way, consumes resources and contributes to battery drain and background activity time of the host application in the iOS battery report.

##### What's in this PR?

This change makes background data gathering more efficient by disabling standard location updates when the device has been stationary for at least five minutes. It also establishes region monitoring and significant location change monitoring so that if the device appears to be in motion again then background telemetry data collection can resume. All of this reduces the amount of time required for telemetry data collection to the time the device is in motion only. It also only applies to host apps that already run in the background and have the `always` location permission from  their users.

This also includes some changes to make the internal pause/resume API of the `MGLMapboxEvents` class less complex and autonomous. The side effects of the map view waking or sleeping are no longer required for mapbox events to work as intended.

##### Addresses:
- https://github.com/mapbox/mapbox-gl-native/issues/1422
- https://github.com/mapbox/mapbox-gl-native/issues/1424
- https://github.com/mapbox/mapbox-gl-native/issues/3916
- https://github.com/mapbox/mapbox-gl-native/issues/4030

##### Subsumes:
- https://github.com/mapbox/mapbox-gl-native/pull/3659

cc @1ec5 @friedbunny 
fyi @bleege @mick @camilleanne @willwhite

### DIFF
--- a/gyp/platform-ios.gypi
+++ b/gyp/platform-ios.gypi
@@ -48,6 +48,8 @@
         '../platform/ios/src/MGLMapboxEvents.m',
         '../platform/ios/src/MGLAPIClient.h',
         '../platform/ios/src/MGLAPIClient.m',
+        '../platform/ios/src/MGLLocationManager.h',
+        '../platform/ios/src/MGLLocationManager.m',
         '../platform/ios/src/MGLMapView.mm',
         '../platform/ios/src/MGLAccountManager_Private.h',
         '../platform/ios/src/MGLAccountManager.m',
--- a/platform/ios/src/MGLLocationManager.h
+++ b/platform/ios/src/MGLLocationManager.h
@@ -1,0 +1,25 @@
+#import <Foundation/Foundation.h>
+#import <CoreLocation/CoreLocation.h>
+
+@protocol MGLLocationManagerDelegate;
+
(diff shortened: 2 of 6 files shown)


platform/ios/src/MGLLocationManagerUtility.h
```diff
@@ -0,0 +1,25 @@
+#import <Foundation/Foundation.h>
+#import <CoreLocation/CoreLocation.h>
+
+@protocol MGLLocationManagerUtilityDelegate;
+
+@interface MGLLocationManagerUtility : NSObject <CLLocationManagerDelegate>
```
1ec5: Maybe call this MGLLocationManager or MGLTelemetryLocationManager, since it functions like a replacement for CLLocationManager rather than a complement to it.

willwhite:
:clap:

PR #4144, opened 2016-02-29
Removing Partial Wake Lock

### LINKED ISSUE #4142: Remove Partial Wake Lock From Telemetry
We've decided to remove the use of the app session based Partial Wake Lock in Telemetry.  While Internal testing has shown that it is comparable to Google Play Services in terms of battery usage and performance, removing it will further minimize configuration for developers.

/cc @mapbox/telemetry

bleege:
Closes #4142


bleege:
Rebased and Merged!

Issue #4078, opened 2016-02-23
Message of exceptions redirecting users to .md files


tobrun:
This type of message is found in `InvalidAccessTokenException`:
```
/**
 * A {@code InvalidAccessTokenException} is thrown by {@link MapView} when there is either no access
 * token set before {@link MapView#onCreate(Bundle)} or an invalid access token is set in {@link MapView#setAccessToken(String)}
 *
 * @see MapView#onCreate(Bundle)
 * @see MapView#setAccessToken(String)
 */
public class InvalidAccessTokenException extends RuntimeException {

    public InvalidAccessTokenException() {
        super("Using MapView requires setting a valid access token. See the INSTALL.md");
    }

}
```

And in `TelemetryServiceNotConfiguredException`:

```
/**
 * A {@code TelemetryServiceNotConfiguredException} is thrown by {@link MapView} when it checks and finds that
 * TelemetryService has not been configured in the app's AndroidManifest.xml {@link MapView#onCreate(Bundle)}
 *
 * @see MapView#onCreate(Bundle)
 */
public class TelemetryServiceNotConfiguredException extends RuntimeException {

    public TelemetryServiceNotConfiguredException() {
        super("Using Mapbox Android SDK requires configuring TelemetryService. See the INSTALL.md");
    }

}
```

This is giving our users a bad user experience. Take for example someone that installs our SDK through `mapbox.com`. They will not know where to locate this `INSTALL.md` file. Unexperienced users might even not know what an `.md` file is. The general rule I apply on writing good exception message is that a user need to know directly **what to fix** and **where to fix it**.

@danswick: can you reflect on this with things you are seeing on support side?

cc @bleege @zugaldia

cammace:
Removes referring to install.md when access token and telemetry exception occurs. Instead it now describes the steps to take in order to resolve the problem.

PR #4161, opened 2016-03-01
[android] #4078 fixed exceptions referring to install.md

cammace:
Removes referring to install.md when access token and telemetry exception occurs. Instead it now describes the steps to take in order to resolve the problem.

### DIFF
--- a/platform/android/MapboxGLAndroidSDK/src/main/java/com/mapbox/mapboxsdk/exceptions/InvalidAccessTokenException.java
+++ b/platform/android/MapboxGLAndroidSDK/src/main/java/com/mapbox/mapboxsdk/exceptions/InvalidAccessTokenException.java
@@ -13,7 +13,7 @@ import com.mapbox.mapboxsdk.maps.MapView;
 public class InvalidAccessTokenException extends RuntimeException {
 
     public InvalidAccessTokenException() {
-        super("Using MapView requires setting a valid access token. See the INSTALL.md");
+        super("\nUsing MapView requires setting a valid access token. Please see https://www.mapbox.com/help/create-api-access-token/ to learn how to create one.\n     Lastly, follow the steps in this guide https://www.mapbox.com/help/first-steps-android-sdk/#access-tokens in order to include the access token in your app.");
     }
 
 }
--- a/platform/android/MapboxGLAndroidSDK/src/main/java/com/mapbox/mapboxsdk/exceptions/TelemetryServiceNotConfiguredException.java
+++ b/platform/android/MapboxGLAndroidSDK/src/main/java/com/mapbox/mapboxsdk/exceptions/TelemetryServiceNotConfiguredException.java
@@ -12,7 +12,7 @@ import com.mapbox.mapboxsdk.maps.MapView;
 public class TelemetryServiceNotConfiguredException extends RuntimeException {
 
     public TelemetryServiceNotConfiguredException() {
-        super("Using Mapbox Android SDK requires configuring TelemetryService. See the INSTALL.md");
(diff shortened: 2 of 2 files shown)


tobrun:
I'm going to pull this in our repo to allow it to be tested by our CI.

tobrun:
Related tot the PR: 
I think it a big step forward versus what we had before. Not validated, but I think the shown URLs in the logcat are clickable for our ursers, which is a huge improvement!

Just to validate current message:
 - The user now knows more about what has gone wrong
 - We aren't giving the user a direct solution in the exception message, we are still redirecting them.

I think this improvement is a good enough improvement to merge in without tackling the second item mentioned above. 

@bleege @zugaldia 
- Do you think a clickable link is enough? Rather have the exception message explain next steps?

tobrun:
I pushed this to our repo, you can see the test running here:
screen shot 2016-03-02 at 12 53 47

platform/android/MapboxGLAndroidSDK/src/main/java/com/mapbox/mapboxsdk/exceptions/TelemetryServiceNotConfiguredException.java
```diff
@@ -12,7 +12,7 @@
 public class TelemetryServiceNotConfiguredException extends RuntimeException {
 
     public TelemetryServiceNotConfiguredException() {
-        super("Using Mapbox Android SDK requires configuring TelemetryService. See the INSTALL.md");
+        super("\nUsing Mapbox Android SDK requires configuring TelemetryService. This only requires two steps:\n    1. Include the WAKE_LOCK permission within your applications AndroidManifest.xml\n    2. Add the \"com.mapbox.mapboxsdk.telemetry.TelemetryService\" service in your applications AndroidManifest.xml\n    For more help, visit https://www.mapbox.com/android-sdk/");
```
bleege: Let's remove the phrase "this only" as it's subjective.  The WAKE_LOCK requirement is now gone.  I like the style of this though:

1. State the problem
1. State the technical solution
1. For more help click here...

platform/android/MapboxGLAndroidSDK/src/main/java/com/mapbox/mapboxsdk/exceptions/InvalidAccessTokenException.java
```diff
@@ -13,7 +13,7 @@
 public class InvalidAccessTokenException extends RuntimeException {
 
     public InvalidAccessTokenException() {
-        super("Using MapView requires setting a valid access token. See the INSTALL.md");
+        super("\nUsing MapView requires setting a valid access token. Please see https://www.mapbox.com/help/create-api-access-token/ to learn how to create one.\n     Lastly, follow the steps in this guide https://www.mapbox.com/help/first-steps-android-sdk/#access-tokens in order to include the access token in your app.");
```
bleege: Should be updated to follow the 3 steps:

1. State the problem
1. State the technical solution
1. For more help click here...

platform/android/MapboxGLAndroidSDK/src/main/java/com/mapbox/mapboxsdk/exceptions/TelemetryServiceNotConfiguredException.java
```diff
@@ -12,7 +12,7 @@
...
     }
```
mb12: @bleege we should get rid of the service altogether.  Instead just convert it into a regular class hooked with life cycle of the MapView.

tobrun: @mb12 that is indeed a valid point, not sure what future plans are for telemetry.

tobrun:
Going to merge this with #4230

PR #4164, opened 2016-03-02
[iOS] Update iOS Telemetry for V2

### LINKED ISSUE #4148: Convert iOS Telemetry To V2 Data Model
This was started in https://github.com/mapbox/mapbox-gl-native/pull/3620. We should update the event model serialization to be compatible with the V2 spec and post the the V2 endpoint.

##### Tasks

- [x] Update turnstile event
- [x] Update map load event
- [x] Update map click event
- [x] Update map drag end event
- [x] Update location event
- [x] Stop sending visits (already done in master)
- [x] Reduce lat/lng precision to 7 digits
- [x] Update the user agent to include the SDK version
- [x] Pull in anything from https://github.com/mapbox/mapbox-gl-native/pull/3620/commits that still makes sense
- [x] Use location manager timestamp on location events instead of artificial one
- [x] Decide if we want core telephony for `carrier` and `cellularNetworkType` (see https://github.com/mapbox/mapbox-gl-native/issues/3295)
- [x] Change session id based on time boundary like [MapboxEventManager.java](https://github.com/mapbox/mapbox-gl-native/blob/5ac6fc523dea6374da7ba15a7c0433b115987ed2/platform/android/MapboxGLAndroidSDK/src/main/java/com/mapbox/mapboxsdk/telemetry/MapboxEventManager.java#L375-L381)
- [x] Add hook to catch network requests that receive a troublesome status code
- [x] Make the flush timeouts slightly less aggressive (20 events or 60 seconds -> 180 events or 180 seconds)

cc @camilleanne @mick

boundsj:
Closes #4148

### DIFF
--- a/platform/ios/src/MGLAPIClient.m
+++ b/platform/ios/src/MGLAPIClient.m
@@ -2,8 +2,10 @@
 #import "NSBundle+MGLAdditions.h"
 #import "MGLAccountManager.h"
 
-static NSString * const MGLAPIClientUserAgent = @"MapboxEventsiOS/1.1";
-static NSString * const MGLAPIClientBaseURL = @"https://api.tiles.mapbox.com";
+static NSString * const MGLAPIClientUserAgentBase = @"MapboxEventsiOS";
+static NSString * const MGLAPIClientBaseURL = @"https://api.mapbox.com";
+static NSString * const MGLAPIClientEventsPath = @"events/v2";
+static NSString * const MGLAPIClientDomain = @"com.mapbox.mglnative";
 
 static NSString * const MGLAPIClientHeaderFieldUserAgentKey = @"User-Agent";
 static NSString * const MGLAPIClientHeaderFieldContentTypeKey = @"Content-Type";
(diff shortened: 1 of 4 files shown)


platform/ios/src/MGLAPIClient.m
```diff
@@ -106,7 +107,8 @@ - (void)setupUserAgent {
     NSString *appName = [[NSBundle mainBundle] objectForInfoDictionaryKey:@"CFBundleName"];
     NSString *appVersion = [[NSBundle mainBundle] objectForInfoDictionaryKey:@"CFBundleShortVersionString"];
     NSString *appBuildNumber = [[NSBundle mainBundle] objectForInfoDictionaryKey:@"CFBundleVersion"];
-    _userAgent = [NSString stringWithFormat:@"%@/%@/%@ %@", appName, appVersion, appBuildNumber, MGLAPIClientUserAgent];
+    NSString *sdkVersion = [NSBundle mgl_frameworkBundle].infoDictionary[@"CFBundleShortVersionString"];
```
1ec5: Come to think of it, what about using the `MGLSemanticVersionString` key instead, which will include, e.g., `-pre.1` for prereleases? Would distinguishing between final release and prerelease be useful?

boundsj: I think that would be very useful @1ec5 

@camilleanne @mick would that be ok? For our prerelease builds, it would mean the UA would look something like `Mapbox GL/0.0.2/7877 MapboxEventsiOS/3.2.0-pre.1`, ``Mapbox GL/0.0.2/7877 MapboxEventsiOS/3.2.0-pre.2`, etc..

mick: :+1:

platform/ios/src/MGLMapboxEvents.h
```diff
@@ -4,29 +4,20 @@
 
 NS_ASSUME_NONNULL_BEGIN
 
+/// Event types
```
1ec5: `///` creates a documentation comment. You probably didn’t intend to document `MGLEventTypeAppUserTurnstile` individually with this comment.

boundsj: Oops. Fixed https://github.com/mapbox/mapbox-gl-native/commit/aff672c20575acaa9e87cba69d6170f385a5e123

PR #4169, opened 2016-03-02
[android] #4166 - move style url to map view for improved async loading

### LINKED ISSUE #4166: Moving setStyle back to MapView
@bleege @zugaldia 

**TL;DR**:
  - We need to invoke `OnMapReadyCallback` after map has finished loading. 
  - a map can only be loaded if we have set the related style url.
  - `mMapview.setStyle/setStyleUrl` must be called before `mMapview.onCreate` 
  - An exception similar to `InvalidAccessTokenException` will be thrown if not

With the introduction of the `OnMapReadyCallback` we have introduced a async callback that is invoked when we the Map is ready to interact with. Recently I started to notice small side effects when executing methods immediately after this callback has occurred (one strange side effect documented in #4074). It seems that some methods need to be executed before we can interact with the map. Our current implementation is too "quick".

Current implementation just post a message on the messagequeu:

```java
    /**
     * Sets a callback object which will be triggered when the {@link MapboxMap} instance is ready to be used.
     *
     * @param callback The callback object that will be triggered when the map is ready to be used.
     */
    @UiThread
    public void getMapAsync(@NonNull final OnMapReadyCallback callback) {
        post(new Runnable() {
            @Override
            public void run() {
                callback.onMapReady(mMapboxMap);
            }
        });
    }
```

I'm now looking into the possibility to invoke the callback when the Map has finished loading as indicated by the `OnMapChangeListener`. The `getMapAsync` method will look like:

```java

    /**
     * Sets a callback object which will be triggered when the {@link MapboxMap} instance is ready to be used.
     *
     * @param callback The callback object that will be triggered when the map is ready to be used.
     */
    @UiThread
    public void getMapAsync(@NonNull final OnMapReadyCallback callback) {
        mMapReadyCallback = callback;
    }
```

while the `OnMapReadyCallback` will look like:

```java
        // onCreate of MapView
        addOnMapChangedListener(new OnMapChangedListener() {
            @Override
            public void onMapChanged(@MapChange int change) {
                if (change == DID_FINISH_LOADING_MAP) {
                    reloadIcons();
                    reloadMarkers();
                    adjustTopOffsetPixels();
                    mMapReadyCallback.onMapReady(mMapboxMap);
                }
            }
        });
```

Doing this solves these small side effects since markers, icons and offsets are reloaded before interacting with the map but this introduces an extra requirement for the end developer. The `OnMapChangeListener` above will only be invoked if we `setStyle/setStyleUrl` was called.

We used to do it in `4.0.0-SNAPSHOT`with:

```java
        mMapView = (MapView) findViewById(R.id.cameraMapView);
        mMapView.setAccessToken(ApiAccess.getToken(this));
        mMapView.onCreate(savedInstanceState);
        mMapView.getMapAsync(new OnMapReadyCallback() {
            @Override
            public void onMapReady(@NonNull final MapboxMap mapboxMap) {
                // set a style
                mapboxMap.setStyle(Style.MAPBOX_STREETS);
           }        
```

But will now require the end developer to:

```java
        mMapView = (MapView) findViewById(R.id.cameraMapView);
        mMapView.setAccessToken(ApiAccess.getToken(this));
        mMapView.setStyleUrl(Style.MAPBOX_STREETS);
        mMapView.onCreate(savedInstanceState);
        mMapView.getMapAsync(new OnMapReadyCallback() {
            @Override
            public void onMapReady(@NonNull final MapboxMap mapboxMap) {
               // customise map
            }
```

I want to add a note here that this pattern of setting style with accesstoken has been a `semi-requirement` since the beginning of this SDK. The accesstoken was more visible for the end developer since it threw its own exception. Not setting a style was less visible because the developer did not see the app crashing and just saw a black map. Since setting a styleurl will become a more fixed requirement I'm going to introduce a similar check as with accesstoken and throw an exception.

tobrun:
closes #4166 

:eyes: @bleege @zugaldia

### DIFF
--- a/platform/android/MapboxGLAndroidSDK/src/main/java/com/mapbox/mapboxsdk/maps/MapFragment.java
+++ b/platform/android/MapboxGLAndroidSDK/src/main/java/com/mapbox/mapboxsdk/maps/MapFragment.java
@@ -9,6 +9,7 @@ import android.view.View;
 import android.view.ViewGroup;
 
 import com.mapbox.mapboxsdk.R;
+import com.mapbox.mapboxsdk.constants.Style;
 import com.mapbox.mapboxsdk.utils.ApiAccess;
 
 /**
--- a/platform/android/MapboxGLAndroidSDK/src/main/java/com/mapbox/mapboxsdk/maps/MapView.java
+++ b/platform/android/MapboxGLAndroidSDK/src/main/java/com/mapbox/mapboxsdk/maps/MapView.java
@@ -55,6 +55,7 @@ import android.widget.FrameLayout;
 import android.widget.ImageView;
 import android.widget.ListView;
 import android.widget.ZoomButtonsController;
+
 import com.almeros.android.multitouch.gesturedetectors.RotateGestureDetector;
 import com.almeros.android.multitouch.gesturedetectors.ShoveGestureDetector;
 import com.almeros.android.multitouch.gesturedetectors.TwoFingerGestureDetector;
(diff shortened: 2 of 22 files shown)


tobrun:
CI approved, going to fixup commits with a rebase and merge!

tobrun:
Forgot to remove `InvalidStyleUrlException`...

tobrun:

PR #4171, opened 2016-03-02
[offline] Accept parameters

tmpsantos:
The defaults remains the same: Bay area at zoom 15.
The key, if not supplied, is taken from the environment.

```
Allowed options:
  -s [ --style ] URL                    Map stylesheet
  --north degrees (=37.200000000000003) North latitude
  --west degrees (=-122.8)              West longitude
  --south degrees (=38.100000000000001) South latitude
  --east degrees (=-121.7)              East longitude
  --minZoom number (=0)                 Min zoom level
  --maxZoom number (=15)                Max zoom level
  --pixelRatio number (=1)              Pixel ratio
  -t [ --token ] key                    Mapbox access token
  -o [ --output ] file (=offline.db)    Output database file name
```

### DIFF
--- a/bin/offline.cpp
+++ b/bin/offline.cpp
@@ -1,3 +1,4 @@
+#include <mbgl/util/default_styles.hpp>
 #include <mbgl/util/run_loop.hpp>
 #include <mbgl/util/string.hpp>
 #include <mbgl/util/io.hpp>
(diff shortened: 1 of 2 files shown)


jfirebaugh:
* How about `north`, `south`, `east`, `west` instead?
* `minZoom` and `maxZoom`

Otherwise :+1:.

tmpsantos:
> How about north, south, east, west instead?
> minZoom and maxZoom

Sounds better, gonna change that. Thanks.

PR #4174, opened 2016-03-02
[core] Map flashing to clear color when interacting

brunoabinader:
Reverting https://github.com/mapbox/mapbox-gl-native/pull/3873 seems to fix the issue.

Testing on a Linux x86_64.

out

/cc @brunoabinader


tobrun:
FIWIW I was now also able to reproduce this, let me see if rebuilding resolves this.

tobrun:
Yes rebuilding fixed it!

brunoabinader:
@tobrun awesome, thanks for testing!

PR #3645, opened 2016-03-03
Adding a class adds an infinite cascade loop

brunoabinader:
In the `linuxapp`, press "R" to toggle night styles and observe how `Style::cascade` gets called ceaselessly.


brunoabinader:
:eyes: @kkaefer @jfirebaugh

PR #4193, opened 2016-03-04
Use jni.hpp

### LINKED ISSUE #2225: condense JNI boilerplate
Not critical right now (so leaving off of even the 0.2.0 Android milestone), but we should think about refactoring and condensing much of the setup & teardown in `jni.cpp` per https://github.com/mapbox/mapbox-gl-native/pull/2209#issuecomment-136782502. 

Here is but a small sample: 

https://github.com/mapbox/mapbox-gl-native/blob/244fe93f0e6610cd7d66e1f9fd08ca84cb0dd7b2/android/cpp/jni.cpp#L1816-L1860

### LINKED ISSUE #2825: Catch C++ exceptions before returning to JNI
Per #2804, the Android SDK suffers from a native crash any time there’s an uncaught C++ exception at the mbgl level. An uncaught exception there usually indicates developer error, so the Android SDK needs to present that error in a way that’s actionable without spelunking into mbgl code. The Android bindings cannot rely on an assumption that the C++ code won’t throw exceptions.

platform/android/jni.cpp needs to have C-style exception handling. For example, [`nativeSetStyleURL()`](https://github.com/mapbox/mapbox-gl-native/blob/f3ffb45cf6adbdca375217c645eb455ff47991ec/platform/android/jni.cpp#L563) should be wrapped in a try-catch statement. When an exception is caught, this function should return some kind of error code. For functions that already need to return a value, this would imply some kind of out parameter or a tuple-typed return value.

Though this could be a fair amount of work, it pays off in the form of Android bindings that can actually be debugged.

/cc @ljbade @bleege @jfirebaugh

### LINKED ISSUE #3438: [Android] Remove unused parameters in HTTPAndroidRequest
Via @mikemorris, there seem to be some unused parameters in `HTTPAndroidRequest` creating some warnings.

Related commit: https://github.com/mapbox/mapbox-gl-native/commit/03cf403be7677d5420d0b91351029cd8397d8cb6#commitcomment-14539204

/cc: @bleege @tobrun

### LINKED ISSUE #4043: Update JNI bindings to use modern C++ practices

jfirebaugh:
Still working on the integration here, but the basics are in place.

Fixes #2225.
Fixes #2825.
Fixes #4043.
Fixes #3438.

:eyes: @zugaldia @bleege @tobrun


jfirebaugh:
Tested in the test app:

* Basic panning and zooming
* Toggling a custom layer
* Creating an offline region
* Listing offline regions

jfirebaugh:
The followup work here is to convert the NativeMapView and offline bindings to [jni.hpp high-level wrappers](https://github.com/mapbox/jni.hpp#high-level-wrappers) and generally clean up that code. I'll ticket / PR that separately.

zugaldia:
@jfirebaugh This. is. fantastic. Thank you very much for putting the effort into building this. My favorite part is the almost 2k lines that are removed with this PR so far, it speaks mountains to the major cleanup that this work represents.

I'm only starting to get familiar with `jni.hpp` but from reading the README and samples, it all makes sense so far overall.

One question regarding the lack of `DeleteGlobalRef` calls. I read in one of the samples that _"The global references created during registration are purposefully leaked"_. I've read the linked post and I understand the rationale, I wonder if this could affect somehow the way we're building the higher level Android API. Should we be in the look out for leaks?

Related, I see that `DeleteGlobalRef` remains for the offline callbacks. Is this temporary? How could we tackle https://github.com/mapbox/mapbox-gl-native/issues/4121 in the light of `jni.hpp`?

> The followup work here is to convert the NativeMapView and offline bindings to jni.hpp high-level wrappers and generally clean up that code.

I'd really appreciate if in the process of doing this we could document the new logic to do mbgl <---> Android communication with `jni.hpp`. This has classically been, and I include myself here, one of the major difficulties that Java developers have found to contribute to this project.

bleege:
I love that we're getting to clean up the JNI spiderweb.  Thank you for doing this @jfirebaugh.

One request though, would it be possible to hold off merging this into `master` until after `4.0.0` ships?  The reason being is that we're very close to shipping and this would be a large variable thrown into the mix at a late stage.  Let's voice about this during scrum later today to weigh the pros / cons.

jfirebaugh:
> I wonder if this could affect somehow the way we're building the higher level Android API. Should we be in the look out for leaks?

The upshot is that `JNI_OnUnload` is basically never called, and you pretty much have to give up and just permanently leak the global class references obtained via `FindClass` + `NewGlobalRef`. That's the approach this PR takes, in particular with the commit that removes `JNI_OnUnload`.

bleege:
:bow: :bow:

kkaefer:
:100:

PR #4194, opened 2016-03-04
Add package 'offline' to proguard keeps

saschpe:
To avoid release build crashes at runtime due to missing classes.


jfirebaugh:
Thanks! Merged as 661e9e7f49d6aadbd9c14a99b213d8fa0653b469.

PR #4204, opened 2016-03-05
[android] #4195 - make on touch robuster

### LINKED ISSUE #4195: [android] NullPointerException in MapView.onTouchEvent
I've seen this crash twice now (in crashlytics), but I don't have steps to reproduce it because it is happening on a customer's device:

```
Fatal Exception: java.lang.NullPointerException: Attempt to invoke virtual method 'void com.mapbox.mapboxsdk.views.NativeMapView.setGestureInProgress(boolean)' on a null object reference
       at com.mapbox.mapboxsdk.views.MapView.onTouchEvent(MapView.java:2845)
       at android.view.View.dispatchTouchEvent(View.java:8471)
       at android.view.ViewGroup.dispatchTransformedTouchEvent(ViewGroup.java:2373)
```

I took a look at the onTouchEvent code, and I think it would make sense to have a `mNativeMapView  != null` check before calling `setGestureInProgress`

tobrun:
closes #4195

### DIFF
--- a/platform/android/MapboxGLAndroidSDK/src/main/java/com/mapbox/mapboxsdk/maps/MapView.java
+++ b/platform/android/MapboxGLAndroidSDK/src/main/java/com/mapbox/mapboxsdk/maps/MapView.java
@@ -1285,8 +1285,11 @@ public class MapView extends FrameLayout {
     // Called when user touches the screen, all positions are absolute
     @Override
     public boolean onTouchEvent(@NonNull MotionEvent event) {
-        // Check and ignore non touch or left clicks
+        if (mNativeMapView == null) {
+            return false;
+        }
 
+        // Check and ignore non touch or left clicks
         if ((event.getButtonState() != 0) && (event.getButtonState() != MotionEvent.BUTTON_PRIMARY)) {
             return false;
         }


mb12:
@tobrun This does not look right. Native map View is created in the constructor of mapview. It should never be null.

tobrun:
@mb12 I know this should never occur but it sometimes does as mentioned in https://github.com/mapbox/mapbox-gl-native/issues/4195. Something I noticed over the years doing Android is that you can't explain all the exceptions coming in from your crash reporting, the only thing in such situations is avoid the crash.

tobrun:
Thinking a bit more about this, this will probably occur when the Mapview is being destroyed and the framework is still sending us touch events.

Here we are setting NativeMapView to null https://github.com/mapbox/mapbox-gl-native/blob/master/platform/android/MapboxGLAndroidSDK/src/main/java/com/mapbox/mapboxsdk/maps/MapView.java#L484

mb12:
Okay, that would explain it. Would it be safer to return "super.onTouchEvent(event)" instead of "false" in those cases?

tobrun:
@mb12 that would be a bit cleaner, imo the changes of us moving to another parent view group that isn't part of the Android SDK will be very small (all of those return false when called `onTouchEvent`.
I have picked up your suggestion in #4212 and I'm tackling more occurrences similar to this one there. 

Thanks again for the :eyes: @mb12!

PR #4209, opened 2016-03-06
Minor MGLAPIClient cleanup

1ec5:
This change replaces `MGLAPIClientDomain` with `MGLErrorDomain`, for consistency with errors raised elsewhere in the SDK. `NSLocalizedString()` calls were also moved, because once we tackle localization (#949), we would only localize the format strings, not every possible expanded string.

/cc @boundsj


boundsj:
:+1: thanks!

Issue #1985, opened 2015-08-04
Remove {get,set}DefaultTransition{Duration,Delay}


jfirebaugh:
These values should come from the style -- https://www.mapbox.com/mapbox-gl-style-spec/#transition.

Existing, class-mutation-related use of these properties should be eliminated as described in #1025.

Refs #302.

ljbade:
Yeah seems weird to have a global parameter for this.

brunoabinader:
I wonder if we might find scenarios where the user agent is willing to modify the default values for such values (i.e. old devices and/or devices without much processing power).

I might be wrong here, but I see this in the same level as the possibility to modify the default font type/size/color in web browsers. Would that make sense then, [to provide a way for user agents to provide their own style default values in a higher order level](https://developer.mozilla.org/en-US/docs/Web/CSS/Cascade#Cascading_order)?

jfirebaugh:
> I wonder if we might find scenarios 

YAGNI -- if this actually comes up in the future, we can think about it then.

brunoabinader:
> if this actually comes up in the future, we can think about it then.

Fair enough :+1:

PR #4213, opened 2016-03-06
[core] Remove default fade/transition duration from MapData

### LINKED ISSUE #1025: map class-related methods should accept transition options
`addClass`, `removeClass` and `setClasses` should accept options to control the transition duration or disable transitions entirely.

`MGLMapView::setStyleClasses` should use this option rather than calling `setDefaultTransitionDuration`.

brunoabinader:
In the process of replacing MapData entirely, this PR:
- Removes default transition {duration,delay} from `MapData`, `Map` and platform-specific APIs
- Adds optional property transition parameters to `addClass`, `removeClass` and `setClasses`
- Replaces default fade duration from `MapData` with a constant value

Among other minor cleanups. Further steps for fully replacing `MapData` with scoped parameters passing will proceed on a separate PR.

Fixes #1025 and #1985.

:eyes: @jfirebaugh @kkaefer


jfirebaugh:
This looks good, but I think we should hold refactoring of this scale until after the Android 4.0.0 and iOS 3.2.0 releases. We should be in high-stability mode for the next week+.

src/mbgl/map/map.cpp
```diff
@@ -92,9 +92,7 @@ void Map::renderSync() {
 }
 
 void Map::update(Update flags) {
-    if (flags & Update::Dimensions) {
-        transform->resize(view.getSize());
-    }
+    if (flags & Update::Dimensions) transform->resize(view.getSize());
```
kkaefer: Please don't remove braces.

PR #4219, opened 2016-03-07
3697 compass insets

RomainQuidet:
Updating position of compass, logo and attribution button according to the contentInsets. 
The contentInsets by default is already defined by the top and bottom layout guides, so this PL simplifies the code.
Less constraints update is also better for speed. Compass size does not need to be updated each time.

### DIFF
--- a/platform/ios/src/MGLMapView.mm
+++ b/platform/ios/src/MGLMapView.mm
@@ -376,6 +376,20 @@ mbgl::Duration MGLDurationInSeconds(NSTimeInterval duration)
     _compassView.userInteractionEnabled = YES;
     [_compassView addGestureRecognizer:[[UITapGestureRecognizer alloc] initWithTarget:self action:@selector(handleCompassTapGesture:)]];
     _compassView.translatesAutoresizingMaskIntoConstraints = NO;
+    [NSLayoutConstraint constraintWithItem:_compassView
+                              attribute:NSLayoutAttributeWidth
+                              relatedBy:NSLayoutRelationEqual
+                                 toItem:nil
+                              attribute:NSLayoutAttributeNotAnAttribute
+                             multiplier:1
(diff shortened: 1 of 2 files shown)


postmechanical:
Any progress on this one? Will it make the ios-v3.3.0 release?

postmechanical:
@boundsj  Any progress on this one? Will it make the ios-v3.3.0 release?

boundsj:
@postmechanical the [changelog for 3.3.0](https://github.com/mapbox/mapbox-gl-native/blob/release-ios-v3.3.0/platform/ios/CHANGELOG.md) is already very large and we want to focus on getting what is already in that release shipped as soon as possible. We can land this PR in a followup patch release.

platform/ios/src/MGLMapView.mm
```diff
@@ -374,10 +375,23 @@ - (void)commonInit
...
+                              attribute:NSLayoutAttributeNotAnAttribute
+                             multiplier:1
+                               constant:_compassView.image.size.width];
+    [NSLayoutConstraint constraintWithItem:_compassView
+                              attribute:NSLayoutAttributeHeight
+                              relatedBy:NSLayoutRelationEqual
+                                 toItem:nil
+                              attribute:NSLayoutAttributeNotAnAttribute
+                             multiplier:1
+                               constant:_compassView.image.size.height];
```
frederoni: These constraints are never added to a view so ARC will have them removed after this line.
They're  also unnecessary since the UIImageView is initialized with `-initWithImage` which handles the size in `-intrinsicContentSize` IIRC. This also means that line: 374 `_compassView.frame = CGRectMake(` is redundant and can be removed.

frederoni:
Closing this in favour of [#5671](https://github.com/mapbox/mapbox-gl-native/pull/5671)

Issue #4155, opened 2016-03-01
PrecisionPoint TransformState::latLngToPoint dateline issue


tobrun:
Issue in `mbgl` when calculating points based on LatLng from the left side of the dateline after crossing the dateline from left to right. This is a follow up issue for https://github.com/mapbox/mapbox-gl-native/issues/3995.

An example:

Take this LatLng:
> LatLng [longitude=-179.64576449708883, latitude=-16.83408606795382, altitude=0.0]

before crossing the dateline `latLngToPoint` is giving us Point[**x=455.80133**, y= 746.6456]
while after crossing the dateline `latLngToPoint` is giving us Point[**x=393667.44**, y=747.1837].
Normally we should see a small decrease in value not a big increase.

tobrun:
Related method in `transform_state.cpp`:

```cpp
PrecisionPoint TransformState::latLngToPoint(const LatLng& latLng) const {
    return coordinateToPoint(latLngToCoordinate(latLng));
}
```

which on his turn calls:

```cpp
PrecisionPoint TransformState::coordinateToPoint(const TileCoordinate& coord) const {
    mat4 mat = coordinatePointMatrix(coord.zoom);
    matrix::vec4 p;
    matrix::vec4 c = {{ coord.column, coord.row, 0, 1 }};
    matrix::transformMat4(p, c, mat);
    return { p[0] / p[3], height - p[1] / p[3] };
}
```
and 
```cpp
TileCoordinate TransformState::latLngToCoordinate(const LatLng& latLng) const {
    const double tileZoom = getZoom();
    const double k = zoomScale(tileZoom) / worldSize();
    return {
        lngX(latLng.longitude) * k,
        latY(latLng.latitude) * k,
        tileZoom
    };
}
```

1ec5:
[...]

PR #4214, opened 2016-03-07
Map jumps from one position to another

brunoabinader:
When panning the map jumps from on location to another.
I'm guessing it's something related to the dateline.
You can see the bug in action [here](https://youtu.be/t3FsKKc5meY)

### DIFF
--- a/src/mbgl/map/transform.cpp
+++ b/src/mbgl/map/transform.cpp
@@ -95,12 +95,17 @@ void Transform::easeTo(const CameraOptions& camera, const AnimationOptions& anim
     if (camera.padding) {
         padding = *camera.padding;
     }
-    const LatLng startLatLng = getLatLng(padding);
+
+    LatLng startLatLng = getLatLng(padding);
+    startLatLng.unwrapForShortestPath(latLng);
+
+    // Make sure the end coordinate always remains valid.
+    latLng.wrap();
+
     const ScreenCoordinate startPoint = {
         state.lngX(startLatLng.longitude),
         state.latY(startLatLng.latitude),
     };
-    latLng.unwrapForShortestPath(getLatLng());
     const ScreenCoordinate endPoint = {
         state.lngX(latLng.longitude),
         state.latY(latLng.latitude),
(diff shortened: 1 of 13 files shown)


brunoabinader:
:eyes: @kkaefer

1ec5:
@brunoabinader, can you add a blurb to the Android and iOS sections of the changelog? Thanks.

tobrun:
I retested but still seeing the issue:
https://youtu.be/gC489mSfS44

brunoabinader:
@tobrun I've updated the PR and tested it on GLFW & iOS. Could you please check it on Android? Thanks!

tobrun:
@brunoabinader yes this issue is no longer reproducible :bow:

tobrun:
FWIW  I retested https://github.com/mapbox/mapbox-gl-native/issues/4155 and this is still an issue

brunoabinader:
Fix for #4155 is in 949ed9ed53debd1d54ef6f1243f2b15a25b6013b (together with unit tests). I have tested it on both Android Nexus 5X and iPad 2 :+1: .

src/mbgl/map/transform.cpp
```diff
@@ -517,7 +527,7 @@ double Transform::getAngle() const {
 #pragma mark - Pitch
 
 void Transform::setPitch(double pitch, const Duration& duration) {
-    setPitch(pitch, {NAN, NAN}, duration);
+    setPitch(pitch, ScreenCoordinate {}, duration);
```
1ec5: So if you don’t specify an `anchor` for a `setPitch()` call, we assume the gesture is anchored in the top-left corner of the view?

brunoabinader: If we don't specify a valid anchor (by _valid_ I mean an anchor that does not contain any NaN and contains at least one non-zero value on its axis),  these are simply not taken in consideration in `startTransition()`.

1ec5: I originally made the change to rely on NaN for anchors because anchoring at the origin (or somewhere on the top or left side) may be a valid use case. Though it may be difficult to trigger such a transition via a gesture, it may still be possible to do so in the iOS SDK with some combination of edge insets, user location or course tracking, and zooming or rotation gestures. Moreover, it’s only a matter of time before we need to allow arbitrary anchor points for programmatic animations in the mobile SDKs.

brunoabinader: You are right @1ec5 - I am reverting this particular change in https://github.com/mapbox/mapbox-gl-native/pull/4277.

PR #4221, opened 2016-03-07
Offline viewing in iOS and OS X

### LINKED ISSUE #3892: iOS bindings for offline API
Previously: #584, #3715.

We now have an mbgl-core offline API for which iOS bindings need to be written.

1ec5:
This PR implements APIs for offline viewing in the iOS and OS X SDKs, based on the work in #3715. Additionally, iosapp’s interface has been partially rewritten using an Interface Builder storyboard and an interface for managing offline downloads has been added.

Here’s a high-level overview of the new classes:

* MGLTilePyramidOfflineRegion, as the only class currently allowed to conform to MGLOfflineRegion, encapsulates `mbgl::OfflineRegionDefinition`, acting as a “bill of materials” for an MGLOfflineTask.
* MGLOfflineTask encapsulates `mbgl::OfflineRegion` and provides methods for pausing and resuming a download. It pairs an MGLOfflineRegion with an opaque context data blob (analogous to `mbgl::OfflineRegionMetadata`).
* A singleton MGLOfflineStorage object encapsulates `mbgl::DefaultFileSource` and provides methods for creating and destroying MGLOfflineTask objects.

A word about naming: I chose to diverge from the offline class names used by mbgl and the Android SDK in #4085. The intent is to keep the SDK’s naming intuitive for developers who are familiar with Cocoa development but not necessarily with maps, to avoid the kind of confusion we’ve been seeing around MGLAnnotationImage. In particular, along with MapKit, Core Location, and other Apple-provided frameworks, the iOS SDK already uses the word “region” to refer to a [more-or-less geographic model object](https://www.mapbox.com/ios-sdk/api/3.1.2/Protocols/MGLMapViewDelegate.html#/c:objc%28pl%29MGLMapViewDelegate%28im%29mapView:regionWillChangeAnimated:) rather than a task or controller. Additionally, in keeping with strong Cocoa conventions, “context” replaces “metadata” and “delegate” replaces “observer”.

To ensure KVO compliance, MGLOfflineTask registers an internal `mbgl::OfflineRegionObserver` for its entire lifetime that dispatches progress updates to an optional MGLOfflineTaskDelegate. There is a bit of an impedance mismatch with `mbgl::OfflineRegion`, which expects all progress information to be requested and communicated to the delegate asynchronously via callbacks. You can see how this gets complicated with `-[MGLOfflineTask requestProgress]`. I’m not completely opposed to removing the `state` and `progress` properties and routing all progress information through an asynchronous API. However, we should then consider some kind of affordance for KVO compliance in the near future, because this is a very common paradigm for frequently changing properties in Cocoa. (KVO would help with hooking up MGLOfflineTask to NSProgress, for example.)

In the iosapp Downloads view, tapping the + button adds and begins a download of the main map’s current viewport, from the current zoom level all the way to the maximum zoom level. It is recommended that you zoom in before using this feature, to avoid maxing out the tile limit. Tapping an in-progress download pauses and resumes the download, while tapping a completed download switches to its associated style and navigates to the downloaded viewport.

Fortuitously, my Internet connection went kaput last night as I was preparing 34611a7801fab053bbbb73ec5af61c63eda4ce38, allowing me to test much of the offline functionality in a real-world environment.

Fixes #3892.

/cc @jfirebaugh @boundsj @friedbunny @zugaldia


1ec5:
@jfirebaugh suggested calling `MGLOfflineTask` `MGLOfflineRecord`, `MGLOfflineArchive`, or `MGLOfflineEntry` instead. `MGLOfflineRecord` sounds the best to me; it would fit right in with `MGLOfflineStorage`. `MGLOfflineArchive` could also work, but it’s a little close to the unrelated concepts of keyed archiving and unarchiving. `MGLOfflineEntry` sounds like the role that a resource plays in the current `mbgl::OfflineRegion`.

jfirebaugh:
Further terminology discussion in https://github.com/mapbox/mapbox-gl-native/issues/4250.

platform/darwin/src/MGLOfflineStorage.mm
```diff
@@ -0,0 +1,151 @@
...
+    self.mbglFileSource->deleteOfflineRegion(std::move(*task.mbglOfflineRegion), [&, task, completion](std::exception_ptr exception) {
+        NSError *error;
+        if (exception) {
+            error = [NSError errorWithDomain:MGLErrorDomain code:-1 userInfo:@{
+                NSLocalizedDescriptionKey: @(mbgl::util::toString(exception).c_str()),
+            }];
+        }
+        if (completion) {
+            dispatch_async(dispatch_get_main_queue(), [&, task, completion, error](void) {
+                [task invalidate];
```
jfirebaugh: I believe `[task invalidate]` should be the first line in this method -- as soon as `mbglOfflineRegion` is moved out of `task`, it's invalidated.

platform/darwin/src/MGLOfflineTask.mm
```diff
@@ -0,0 +1,189 @@
...
+        _mbglOfflineRegionObserver = new MBGLOfflineRegionObserver(self);
+        
+        mbgl::DefaultFileSource *mbglFileSource = [[MGLOfflineStorage sharedOfflineStorage] mbglFileSource];
+        mbglFileSource->setOfflineRegionObserver(*_mbglOfflineRegion, std::make_unique<MBGLOfflineRegionObserver>(*_mbglOfflineRegionObserver));
+    }
+    return self;
+}
+
+- (void)dealloc {
+    delete _mbglOfflineRegionObserver;
```
jfirebaugh: This should be `mbglFileSource->setOfflineRegionObserver(*_mbglOfflineRegion, nullptr)`, and `_mbglOfflineRegionObserver` should go away. The `mbglFileSource` owns the observer.

1ec5: Oh, in that case, the `_mbglOfflineRegion` ivar might need to be marked `unsafe_unretained`, to keep ARC from holding its own reference to the observer.

platform/darwin/src/MGLOfflineTask.mm
```diff
@@ -0,0 +1,189 @@
...
+    MGLAssertOfflineTaskIsValid();
+    
+    mbgl::DefaultFileSource *mbglFileSource = [[MGLOfflineStorage sharedOfflineStorage] mbglFileSource];
+    
+    __weak MGLOfflineTask *weakSelf = self;
+    mbglFileSource->getOfflineRegionStatus(*_mbglOfflineRegion, [&, weakSelf](__unused std::exception_ptr exception, mbgl::optional<mbgl::OfflineRegionStatus> status) {
+        if (status) {
+            mbgl::OfflineRegionStatus checkedStatus = *status;
+            dispatch_async(dispatch_get_main_queue(), ^{
+                MGLOfflineTask *strongSelf = weakSelf;
```
jfirebaugh: It it possible the weak reference can be invalid by this point?

1ec5: The weak reference is either valid or `nil`, in which case the following line is a no-op.

Or are you referring to the fact that I skipped doing a strong/weak dance directly inside the outer lambda? I might be mistaken, but I don’t think an extra level of weak indirection is required.

platform/darwin/src/MGLOfflineStorage.mm
1ec5: What happens if the removal errors out? Would the region continue to be valid, or is it still invalid?

platform/darwin/src/MGLOfflineTask.mm
jfirebaugh: Ah, messaging `nil` is a no-op in Objective C. :+1:

platform/darwin/src/MGLOfflineStorage.mm
jfirebaugh: It's still invalid. As soon as `deleteOfflineRegion` is called, the callee relinquishes ownership of the region. You can get notified of errors, but not do anything about them, other than obtaining another reference to the region via `listOfflineRegions`.

1ec5:
@boundsj pointed out that MGLOfflineStorage should KVO-observe access tokens just like MGLMapView does, in case client code sets the access token programmatically at a later point.

jfirebaugh:
Setting the access token after activating an offline download isn't going to work very well. In that situation, resource downloads may have already been attempted, and `OnlineFileSource` is not equipped to handle changes to the access token after the fact. We should encourage people to set the token as early as possible.

1ec5:
Oh, I forgot I already implemented [KVO on the access token](https://github.com/mapbox/mapbox-gl-native/pull/4221/files#diff-3f0caf5e71e8443a455628a8f6824fa7R69).

> We should encourage people to set the token as early as possible.

Agreed. Even for MGLMapView, setting the access token after initialization doesn’t have guaranteed behavior around caching, but we use KVO observation because it’s the most reliable way for the shared MGLAccountManager object, all instances of MGLMapView, the shared MGLMapboxEvents object, and now the shared MGLOfflineStorage object to stay on the same page with the same access token.

platform/darwin/src/MGLOfflineTask.mm
```diff
@@ -0,0 +1,198 @@
...
+    return nil;
+}
+
+- (instancetype)initWithMBGLRegion:(mbgl::OfflineRegion *)region {
+    if (self = [super init]) {
+        _mbglOfflineRegion = region;
+        _state = MGLOfflineTaskStateUnknown;
+        
+        mbgl::DefaultFileSource *mbglFileSource = [[MGLOfflineStorage sharedOfflineStorage] mbglFileSource];
+        MBGLOfflineRegionObserver *mbglObserver = new MBGLOfflineRegionObserver(self);
```
jfirebaugh: You can write these two lines more succinctly and with only one allocation, as:

```C++
mbglFileSource->setOfflineRegionObserver(*_mbglOfflineRegion, std::make_unique<MBGLOfflineRegionObserver>(self));
```

platform/darwin/src/MGLTilePyramidOfflineRegion.mm
```diff
@@ -0,0 +1,66 @@
...
+
+- (instancetype)initWithStyleURL:(NSURL *)styleURL bounds:(MGLCoordinateBounds)bounds fromZoomLevel:(double)minimumZoomLevel toZoomLevel:(double)maximumZoomLevel {
+    if (self = [super init]) {
+        if (!styleURL) {
+            styleURL = [MGLStyle streetsStyleURL];
+        }
+        
+        if (!styleURL.scheme) {
+            // Assume a relative path into the application bundle.
+            styleURL = [NSURL URLWithString:[@"asset://" stringByAppendingString:styleURL.absoluteString]];
```
jfirebaugh: `asset://` URLs aren't supported for offline.

platform/darwin/src/MGLOfflineTask.mm
```diff
@@ -0,0 +1,198 @@
...
+    const mbgl::OfflineRegionMetadata &metadata = _mbglOfflineRegion->getMetadata();
+    return [NSData dataWithBytes:&metadata[0] length:metadata.size()];
+}
+
+- (void)resume {
+    MGLAssertOfflineTaskIsValid();
+    
+    mbgl::DefaultFileSource *mbglFileSource = [[MGLOfflineStorage sharedOfflineStorage] mbglFileSource];
+    mbglFileSource->setOfflineRegionDownloadState(*_mbglOfflineRegion, mbgl::OfflineRegionDownloadState::Active);
+    self.state = MGLOfflineTaskStateActive;
```
jfirebaugh: I'm concerned about `MGLOfflineTask` having its own `state` ivar that's not necessarily in sync with the underlying `mbgl::OfflineRegion`'s state.

One way the two will get out of sync is a consequence of the fact that `setOfflineRegionDownloadState` is asynchronous. It's possible to receive a few `statusChanged` notifications with a `downloadState` of `Active` even after calling `setOfflineRegionDownloadState(Inactive)`, prior the async state change being processed. Once it is processed, the implementation guarantees to send you a final `statusChanged` notification with a `downloadState` of `Inactive`. In this scenario, the current iOS implementation will flip `self.state` from `Active` ⇢ `Inactive` ⇢ `Active` ⇢ `Inactive`.

I suggest making `self.state` entirely determined by the value last received in `statusChanged`, with two exceptions:

* The initial state is always `MGLOfflineTaskStateUnknown`
* Once the state goes to `MGLOfflineTaskStateInvalid`, further state changes from `statusChanged` are ignored

platform/darwin/src/MGLTilePyramidOfflineRegion.mm
1ec5: Oh, that’s good to know. This line was copy-pasted from the MGLMapView implementation.

If a developer has a local style but wants to download online resources required by it, I suppose the workaround is to create a copy of the style in their Mapbox account and use it just for MGLTilePyramidOfflineRegion. I’ll document that workaround if you think it’s correct.

platform/darwin/src/MGLOfflineTask.mm
1ec5: :+1:

jfirebaugh:
LGTM. Providing this build passes, :shipit:.

1ec5:
Updated jazzy symbol categories in 3622c2d25c5f96e9933bc35e261d3fccef052274.

PR #4230, opened 2016-03-07
[android] #4078 - improve exception message

### LINKED ISSUE #4078: Message of exceptions redirecting users to .md files

tobrun:
closes #4078

### DIFF
--- a/platform/android/MapboxGLAndroidSDK/src/main/java/com/mapbox/mapboxsdk/exceptions/IconBitmapChangedException.java
+++ b/platform/android/MapboxGLAndroidSDK/src/main/java/com/mapbox/mapboxsdk/exceptions/IconBitmapChangedException.java
@@ -20,8 +20,8 @@ import com.mapbox.mapboxsdk.maps.MapView;
 public class IconBitmapChangedException extends RuntimeException {
 
     public IconBitmapChangedException() {
-        super("The added Marker has a Icon with a Bitmap that has been modified. You cannot modufy" +
-                "a Icon after it has been added in a Marker.");
+        super("The added Marker has an Icon with a bitmap that has been modified. An Icon cannot be modified" +
+                "after it has been added to the map in a Marker.");
     }
 
 }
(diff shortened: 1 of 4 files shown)


danswick:
These new messages are a huge improvement, @tobrun! I made a couple of small changes to `IconBitmapChangedException.java` and `TooManyIconsException.java` as well. 

Is there a hard limit on the number of `Icon` objects or is it device-specific? If it's a hard limit, I wonder if we should add the limit to our documentation and link to it in the exception message.

tobrun:
@danswick I did not implement TooManyIconsException but looking at the code this will happen if we have assigned `2147483647` icons. After assigning the last icon with an id of `2147483647`, the int primitive datatype overflows and we will try assigning `-2147483648` as an id. It that case we will throw the `TooManyIconsException`. I'm actually not sure that this will ever happen because it will probably run into a OutOfMemory Exception problem before hitting that amount of icons.

PR #4234, opened 2016-03-08
[core] Add client defined network status

tmpsantos:
This API will let the client force a network status. If set to Offline, we won't make network requests.

When set make to Online, it will trigger the pending requests and try to fetch tiles from the network.

### DIFF
--- a/include/mbgl/storage/default_file_source.hpp
+++ b/include/mbgl/storage/default_file_source.hpp
@@ -102,7 +102,6 @@ public:
 
     // For testing only.
     void put(const Resource&, const Response&);
-    void goOffline();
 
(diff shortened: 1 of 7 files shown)


tmpsantos:
:eyes: @jfirebaugh

jfirebaugh:
No (more) timing-based tests, please. Please figure out how to write the test without using timeouts.

tmpsantos:
> No (more) timing-based tests, please. Please figure out how to write the test without using timers.

:+1: This one just needs to make sure that the request only gets answered after setting NetworkStatus back to Online.

tmpsantos:
Merged.

PR #4241, opened 2016-03-08
[android] #4240 - add flag to resize only when needed.

### LINKED ISSUE #4240: Add flag to indicate resizeView
Capturing from chat with @kkaefer:

Currently we are calling `glViewport` before rendering in:

```cpp
void NativeMapView::beforeRender() {
    mbgl::Log::Debug(mbgl::Event::Android, "NativeMapView::beforeRender()");
    glViewport(0, 0, fbWidth, fbHeight);
}
```
this should only be needed in cases when the size has changed.

We should set a flag in:

```cpp
void NativeMapView::resizeView(int w, int h) {
    width = w;
    height = h;
    map->update(mbgl::Update::Dimensions);
}
```

This flag should be checked in beforeRender, if true execute `glViewport`.

tobrun:
closes #4240


tobrun:

Issue #4243, opened 2016-03-08
[android] single taping on a Marker does not work


mayardb:
Hello Mapbox team!

By taping on the top of a Marker, I do not have any IndoWindow prompting.

Reading the code (from `MapView` and copied below), I notice some things:

- We are using a hard coded tolerance `RectF` around the `tapPoint`. Where do these dimension come from ? (30 * 25 dip, with only 5 dip below the tapPoint)
- We are looking for Markers in this `RectF`. This means that we are looking for Marker position that are currently in it. However if the Marker is high, when taping on the top of the marker we miss the tap using this algorithm.
For Example : try to tap on the top of the default_marker.

```
@Override
public boolean onSingleTapConfirmed(MotionEvent e) {
    // Open / Close InfoWindow
    PointF tapPoint = new PointF(e.getX(), e.getY());

    List<Marker> selectedMarkers = mMapboxMap.getSelectedMarkers();

    final float toleranceSides = 15 * mScreenDensity;
    final float toleranceTop = 20 * mScreenDensity;
    final float toleranceBottom = 5 * mScreenDensity;

    RectF tapRect = new RectF(tapPoint.x - toleranceSides, tapPoint.y + toleranceTop,
            tapPoint.x + toleranceSides, tapPoint.y - toleranceBottom);

    LatLngBounds.Builder builder = new LatLngBounds.Builder();
    builder.include(fromScreenLocation(new PointF(tapRect.left, tapRect.bottom)));
    builder.include(fromScreenLocation(new PointF(tapRect.left, tapRect.top)));
    builder.include(fromScreenLocation(new PointF(tapRect.right, tapRect.top)));
    builder.include(fromScreenLocation(new PointF(tapRect.right, tapRect.bottom)));

    List<Marker> nearbyMarkers = getMarkersInBounds(builder.build());
    long newSelectedMarkerId = -1;

    if (nearbyMarkers != null && nearbyMarkers.size() > 0) {
        Collections.sort(nearbyMarkers);
        for (Marker nearbyMarker : nearbyMarkers) {
            boolean found = false;
            for (Marker selectedMarker : selectedMarkers) {
                if (selectedMarker.equals(nearbyMarker)) {
                    found = true;
                }
            }
            if (!found) {
                newSelectedMarkerId = nearbyMarker.getId();
                break;
            }
        }
    }

    if (newSelectedMarkerId >= 0) {
        List<Annotation> annotations = mMapboxMap.getAnnotations();
[...]

Issue #3643, opened 2016-01-21
getAnnotationsInBounds not returning in bounds annotations


tobrun:
#### Prior
In the Android release 3.1.0 we made the touch target of touch event smaller #3156. 

![2d48ad00-be8c-11e5-94a4-57b2f20b8cd8](https://cloud.githubusercontent.com/assets/2151639/12476275/8adbeda4-c028-11e5-82e1-ea93ac5717ab.png)

This was needed to be able to properly select annotations when they are just side by side. 

![screen shot 2016-01-21 at 10 22 47](https://cloud.githubusercontent.com/assets/2151639/12476340/f61f570e-c028-11e5-833b-f001a524f30a.png)

See the result from that issue [here](https://www.dropbox.com/s/ycucb7t92c7nu1s/device-2016-01-19-101337.mp4?dl=0&preview=device-2016-01-19-101337.mp4).

#### Issue
While making touch target smaller was great for selecting annotations side by side. It did make selecting annotations in general more difficult. I believe the source of this problem can be found in C++ implementation of `getAnnotationsInBounds`. 

#### Visual
In next screenshot I visualised the `BoundingBox` passed to the `getAnnotationsInBounds`: 

![screen shot 2016-01-21 at 10 39 54](https://cloud.githubusercontent.com/assets/2151639/12476734/768e4830-c02b-11e5-8590-62ee33b6585f.png)

In that case an `InfoWindow` is not shown because `getAnnotationsInBounds` returns an empty list.

More examples shown in this [clip](https://www.dropbox.com/s/3hg9ju7glc02613/device-2016-01-21-095454.mp4?dl=0).

#### Problem
I'm currently not sure where to pinpoint this on:
 -  Is the C++ code not taking the size of the Marker image in account? only the base of the marker?
 - Is the C++ code not taking in account the current zoom level when determining?
 - Should the java code calculate a different bounding box based on zoom level?

tobrun:
Let's take a trip down to C++ land and figure this out:

 - `getAnnotationsInBounds()` is calling `nativeGetAnnotationsInBounds()` in jni.cpp
 - `nativeGetAnnotationsInBounds()` is calling `getMap().getPointAnnotationsInBounds()` on native_map_view.cpp / map.cpp
 - `getPointsAnnotationsInBounds()` is calling `getPointAnnotationsInBounds` in annotation_manager.cpp

```cpp
[...]

PR #4244, opened 2016-03-08
[android] #4243 #3643 - refactoring the onSingleTapConfirmed to tap on Markers anywhere on its Icon Bitmap

mayardb:
I manage the click anywhere on the markers.

Tested in the "Press For Marker" sample (manually).

I still have a problem with the `Collections.sort(allMarkers);` which sorts according to the Marker ids.
So with this code, if two markers are approximately at the same `LatLng`, the first added will always received the tap event.

Is there a way to get the marker ahead ?


mayardb:
In the issue #3643:
> bleege modified the milestone: android-v4.1.0 4 days ago

tobrun:
@mayardb we changed the way we are handling milestones as a result for now we removed all the milestones from all the issues. The reason behind this is that we don't want to keep on changing the milestone from on to the next without handing it. We want to clearly communicate publicly what we are going to do.

Related to this issue. this is something that will be likely to be handled in `4.1.0`. We just need to start selecting issues for `4.1.0`.

PR #3834, opened 2016-03-08
Clarify TileID value semantics

kkaefer:
When overzooming tiles, we currently have `TileID` objects that have a `z` value that is out of sync with the `x` and `y` value. The real `z` value is stored in `sourceZ`, but all of the functions (like `.children()` or `.isChildOf()` aren't aware of that. Instead of changing these values, we should flip the semantics and store the overzoomed `z` value in field that is not `z`.


kkaefer:
Remaining tasks here are:

* [ ] Move `tileCover` to use the new IDs
* [ ] Calculating the set of tiles to render currently relies on a combination of overscaled *and* unwrapped. We should get rid of this though by changing the workflow to first ensuring that all ideal tiles (=Overscaled) are getting loaded, then iterating through the Unwrapped covering tiles and selecting the best data (or generating new unwrapped tiles in case the data for an ideal unwrapped tile isn't loaded yet.

kkaefer:
@jfirebaugh @tmpsantos 👀

test/algorithm/generate_clip_ids.cpp
```diff
@@ -0,0 +1,427 @@
...
+//                   << (int64_t(pair.first.canonical.x) +
+//                       pair.first.wrap * (1ll << pair.first.canonical.z))
+//                   << ", " << pair.first.canonical.y << " }, ClipID{ \"" << pair.second.mask
+//                   << "\", \"" << pair.second.reference << "\" } }," << std::endl;
+//     }
+//     std::cout << "          })," << std::endl;
+//     std::cout << "          stencils);" << std::endl;
+// }
+
+} // end namespace
```
brunoabinader: Do we need to keep this commented code?

src/mbgl/renderer/painter_background.cpp
```diff
@@ -59,7 +59,8 @@ void Painter::renderBackground(const BackgroundLayer& layer) {
 
     for (auto &id: tileIDs) {
         mat4 vtxMatrix;
-        state.matrixFor(vtxMatrix, id, id.z);
+        const UnwrappedTileID tileID(id.z, id.x, id.y);
```
brunoabinader: I wonder if it makes sense to have a factory for ```UnwrappedTileID``` within ```CanonicalTileID``` e.g. ```UnwrappedTileID CanonicalTileID::unwrapped()```?

test/algorithm/generate_clip_ids.cpp
kkaefer: Yeah, I left it there for debugging purposes.

src/mbgl/util/tile_cover.cpp
```diff
@@ -98,21 +100,19 @@ int32_t coveringZoomLevel(double zoom, SourceType type, uint16_t tileSize) {
...
 }
 
-std::vector<TileID> tileCover(const TileCoordinate& tl,
-                              const TileCoordinate& tr,
-                              const TileCoordinate& br,
-                              const TileCoordinate& bl,
-                              const TileCoordinate& c,
-                              int32_t z,
-                              int32_t actualZ) {
+std::set<UnwrappedTileID> tileCover(const TileCoordinate& tl,
```
tmpsantos: We could use this opportunity to add the `util::` namespace to this function.

src/mbgl/util/tile_cover.cpp
```diff
@@ -124,19 +124,10 @@ std::vector<TileID> tileCover(const TileCoordinate& tl,
     scanTriangle(tl, tr, br, 0, tiles, scanLine);
     scanTriangle(br, bl, tl, 0, tiles, scanLine);
 
-    t.sort();
-    t.unique();
-
-    t.sort([&](const TileID& a, const TileID& b) {
```
tmpsantos: :-1: My understanding here is we no longer have the "spiral" effect when loading tiles. I tested it here and the tiles are loading from left to right.

kkaefer: Correct, but I have never seen a spiral loading effect. Are you?

tmpsantos: Spiral is maybe was a poor choice of words, but tiles currently get loaded from the center to the corners.

It is very noticeable when you change to a satellite style with the window maximized. IMO this is a very nice thing to have because it starts loading exactly where the user is looking at.

screenshot from 2016-04-26 17-26-11

jfirebaugh: Yeah, loading from the center outward is a nice feature. Can we keep it?

jfirebaugh:
This looks great overall. Having the distinction between different ID types makes the intent much clearer.

src/mbgl/tile/tile_id.hpp
```diff
@@ -0,0 +1,261 @@
...
+class CanonicalTileID;
+class UnwrappedTileID;
+
+// Has integer z/x/y coordinates
+// All tiles must be derived from 0/0/0 (=no tiles outside of the main tile pyramid)
+// Used for requesting data; represents data tiles that exist out there.
+// z is never larger than the source's maxzoom
+class CanonicalTileID {
+public:
+    inline CanonicalTileID(uint8_t z, uint32_t x, uint32_t y);
```
jfirebaugh: Can `inline` go on the in-header definition, rather than the declaration?

jfirebaugh:
Is now a good time to define hash functions for `*TileID` and switch to `std::unordered_{map,set}` where appropriate (#3980)?

kkaefer:
@jfirebaugh Before we do that, we should check whether this actually makes sense; `*TileID`s are only 3-4 numbers which should be sufficiently fast to compare.

src/mbgl/util/tile_cover.cpp
kkaefer: Done

src/mbgl/util/tile_cover.cpp
kkaefer: Added this back.

kkaefer:
@jfirebaugh @tmpsantos can I please get a final review on the changes?

tmpsantos:
From my side, :ship:

jfirebaugh:
Offline wants the spiral loading order as well -- the assumption is that if downloading an offline region is interrupted, it's better to have the center of the region available than the outskirts.

kkaefer:
@jfirebaugh Offline should use an iterator generator instead of holding all tile IDs in memory?

jfirebaugh:
Yes, it should. Is that something that looks feasible as part of this PR?

kkaefer:
Hmm, I think that's out of scope for this one. It's been sitting around for quite a while and I'd like to finally get it merged to pave the way for https://github.com/mapbox/mapbox-gl-native/issues/123

jfirebaugh:
Ok, so revert more fully to spiral order, than merge?

kkaefer:
Added back spiral loading

PR #4247, opened 2016-03-08
Calling RunLoop::run() a second time doesn't work as expected

jfirebaugh:
If you call `RunLoop::run()` twice, the second call is equivalent to `RunLoop::runOnce()`: it only runs the loop once. The expected behavior is that the loop runs until `RunLoop::stop()` is called again.

The reason is that `holderHandle` isn't re-`uv_ref`ed. It should be, every time `RunLoop::run()` is called.


jfirebaugh:
:eyes: @tmpsantos

tmpsantos:
:+1: Nice.

PR #4249, opened 2016-03-08
Moving QuickZoom Event Recording

### LINKED ISSUE #4183: Tracking gesture event QUICK_ZOOM
I just noticed that QUICK_ZOOM tracking event is placed in the double tap event:

```java
  // Called for double taps
        @Override
        public boolean onDoubleTapEvent(MotionEvent e) {
            UiSettings uiSettings = mMapboxMap.getUiSettings();
            if (!uiSettings.isZoomGesturesEnabled()) {
                return false;
            }

            switch (e.getAction()) {
                case MotionEvent.ACTION_DOWN:
                    break;
                case MotionEvent.ACTION_MOVE:
                    break;
                case MotionEvent.ACTION_UP:
                    if (mQuickZoom) {
                        mQuickZoom = false;
                        break;
                    }

                    if (uiSettings.isZoomOnCenter() || !mMapboxMap.getTrackingSettings().isLocationTrackingDisabled()) {
                        // Zoom in around center map with adjusted padding
                        PointF centerPoint = mMapboxMap.getCenterPoint();
                        zoom(true, centerPoint.x, centerPoint.y);
                    } else {
                        // Zoom in on gesture
                        zoom(true, e.getX(), e.getY());
                        trackGestureEvent(MapboxEvent.GESTURE_QUICK_ZOOM, e.getX(), e.getY());
                    }
                    break;
            }
            trackGestureEvent(MapboxEvent.GESTURE_DOUBLETAP, e.getX(), e.getY());

            return true;
        }
```

I think this should be placed in:

```java
  // This class handles two finger gestures and double-tap drag gestures
  private class ScaleGestureListener extends ScaleGestureDetector.SimpleOnScaleGestureListener {
```

cc @bleege

bleege:
Closes #4183

### DIFF
--- a/platform/android/MapboxGLAndroidSDK/src/main/java/com/mapbox/mapboxsdk/maps/MapView.java
+++ b/platform/android/MapboxGLAndroidSDK/src/main/java/com/mapbox/mapboxsdk/maps/MapView.java
@@ -1505,7 +1505,6 @@ public class MapView extends FrameLayout {
                     if (mMapboxMap.getTrackingSettings().isLocationTrackingDisabled()) {
                         // Zoom in on gesture
                         zoom(true, e.getX(), e.getY());
-                        trackGestureEvent(MapboxEvent.GESTURE_QUICK_ZOOM, e.getX(), e.getY());
                     } else {
                         // Zoom in on user location view
                         PointF centerPoint = mUserLocationView.getMarkerScreenPoint();
@@ -1752,6 +1751,7 @@ public class MapView extends FrameLayout {
                     PointF centerPoint = mUserLocationView.getMarkerScreenPoint();
                     mNativeMapView.scaleBy(detector.getScaleFactor(), centerPoint.x / mScreenDensity, centerPoint.y / mScreenDensity);
                 }
+                trackGestureEvent(MapboxEvent.GESTURE_QUICK_ZOOM, detector.getFocusX(), detector.getFocusY());
             }
             return true;
         }


platform/android/MapboxGLAndroidSDK/src/main/java/com/mapbox/mapboxsdk/maps/MapView.java
```diff
@@ -1743,6 +1742,7 @@ public boolean onScale(ScaleGestureDetector detector) {
             if (uiSettings.isScrollGesturesEnabled() && !mQuickZoom && trackingSettings.isLocationTrackingDisabled()) {
                 // around gesture
                 mNativeMapView.scaleBy(detector.getScaleFactor(), detector.getFocusX() / mScreenDensity, detector.getFocusY() / mScreenDensity);
+                trackGestureEvent(MapboxEvent.GESTURE_QUICK_ZOOM, detector.getFocusX(), detector.getFocusY());
```
tobrun: @bleege 
the if check above indicated !quickzoom and a place like this will flood the eventmanager. 
I'm guessing the correct place for this is were we are setting `mQuickZoom = false;` aka the end of the event.

bleege: @tobrun Yeah, this is a weird one.  When I placed log statements in between the first if and else as well as the if / else in the else statement (who's on first?) and did the single finger double tap and drag this was the code the fired.  I'm thinking there may be an issue with how `mTwoTap` gets set.

bleege:
Squashed, Force Pushed to branch, Rebased, Merged!

PR #4237, opened 2016-03-09
Deactivate OfflineDownload when the tile count limit is hit

jfirebaugh:


platform/default/mbgl/storage/offline_download.cpp
```diff
@@ -230,6 +230,7 @@ void OfflineDownload::ensureResource(const Resource& resource, std::function<voi
             && util::mapbox::isMapboxURL(resource.url)
             && offlineDatabase.offlineMapboxTileCountLimitExceeded()) {
             observer->mapboxTileCountLimitExceeded(offlineDatabase.getOfflineMapboxTileCountLimit());
+            setState(OfflineRegionDownloadState::Inactive);
```
1ec5: Should this be flipped? Now `mapboxTileCountLimitExceeded()` should be a signal to client code that the download has stopped, not that it will stop. (For context, in the iOS SDK implementation in #4221, I’d like to set `MGLOfflineTask.state` to `MGLOfflineTaskStateInactive` inside the callback.)

jfirebaugh: `setState(OfflineRegionDownloadState::Inactive)` triggers a `statusChanged` notification to the observer with the new `downloadState`. This is related to my comment here: https://github.com/mapbox/mapbox-gl-native/pull/4221#discussion_r55572448.

1ec5: :+1:

PR #4268, opened 2016-03-10
[core] Check for NaNs in mbgl::util::{min,max,clamp}

brunoabinader:
Tighten `min`, `max` and `clamp` functions to check for NaNs using stdlib min/max. Also switched redefinitions of min/max with a variadic template for each.

/cc @kkaefer @1ec5


jfirebaugh:
LGTM but let's hold on merging until after the 4.0.0/3.2.0 release branch is created.

PR #4271, opened 2016-03-10
[core] Make vec{2,3,4} operator bool explicit

brunoabinader:
TSIA.

/cc @jfirebaugh


jfirebaugh:
This isn't what I meant. I meant, change:

```
 inline operator bool() const { ... }
```

to:

```
 explicit operator bool() const { ... }
```

(`inline` is unnecessary)

jfirebaugh:
https://en.wikibooks.org/wiki/More_C%2B%2B_Idioms/Safe_bool

PR #4274, opened 2016-03-10
[android] #4273 - library resource prefix espresso update

### LINKED ISSUE #4273: Espresso test still use library resource prefix
Something that was unnoticed when reverting, since it's not linked when building the testapp.

tobrun:
closes #4273


tobrun:

PR #4275, opened 2016-03-11
[core] Fix screen coordinates when crossing the antimeridian (#2)

### LINKED ISSUE #4155: PrecisionPoint TransformState::latLngToPoint dateline issue

brunoabinader:
Added missing case. Really fixes #4155 this time.

:eyes: @tobrun

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -52,7 +52,7 @@ Known issues:
 ## iOS master
 
 - Offline packs can now be downloaded to allow users to view specific regions of the map offline. A new MGLOfflineStorage class provides APIs for managing MGLOfflinePacks. ([#4221](https://github.com/mapbox/mapbox-gl-native/pull/4221))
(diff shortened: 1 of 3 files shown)


tobrun:
Will be testing this in the morning!

tobrun:
It works, nice!

PR #4279, opened 2016-03-11
Deactivate offline region upon completion

1ec5:
I think it’s reasonable for the SDKs to expect that the `mbgl::OfflineRegion` will automatically deactivate when complete.

/cc @jfirebaugh @bleege @boundsj


jfirebaugh:
Tests need to be updated / added, but I agree with the approach.

PR #4285, opened 2016-03-11
[core] Harden Transform anchor & padding usage

brunoabinader:
Enforce invalid anchors via `ScreenCoordinate::null()` when neeeded, backed up by unit tests. This patch reverts some more changes from b33b2f15 related to the default value for Transform functions. As pointed out by Minh, we use invalid screen coordinates to tell the affected transform functions to use the origin instead.

/cc @1ec5 @kkaefer

### DIFF
--- a/include/mbgl/map/map.hpp
+++ b/include/mbgl/map/map.hpp
@@ -1,6 +1,7 @@
 #ifndef MBGL_MAP_MAP
 #define MBGL_MAP_MAP
 
+#include <mbgl/util/optional.hpp>
 #include <mbgl/util/chrono.hpp>
 #include <mbgl/util/image.hpp>
 #include <mbgl/map/update.hpp>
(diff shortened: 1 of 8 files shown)


jfirebaugh:
We should be using `optional<ScreenCoordinate>` for optional `ScreenCoordinate`s, not a `vec2<double>` containing NaNs.

brunoabinader:
**TL;DR**: If gesturing, do what the user wants, instead take the shortest path.

@1ec5 this fixes the shortest path code I broke in 026b6d4c01a3d96af9629cc1790373137ecab950, together with an issue where a use gesture of scrolling around the world would have unexpected behavior due to the _shortest path_ algorithm. We use `isGestureInProgress()` to ignore that and instead transfer the world rounds from the end longitude into start, so we still get the scrolling effect while assuring that the end longitude remains _wrapped_.

brunoabinader:
Annotation tooltips are shown in both sides of the dateline with wrapped longitudes now:
new

brunoabinader:
@bleege @tobrun I believe this should be on android-v4.0.0 milestone too.

1ec5:
That makes a lot of sense. Thanks @brunoabinader!

bleege:
:+1: @brunoabinader

Issue #4287, opened 2016-03-12
MGLOfflineStorage should strongly own MGLOfflinePacks


1ec5:
If you follow [this example](https://www.mapbox.com/help/mobile-offline/) for creating an offline pack on iOS, you’ll run into a crash because the MGLOfflinePack is only held in a local variable. It is expected that the completion handler would store the pack in an ivar of the surrounding view controller, but this is still a poor experience for someone who follows the documentation to the letter.

MGLOfflineStorage should strongly own an MGLOfflinePack from the time it is added to the time it is removed, to prevent the object from being deleted via ARC.

/cc @boundsj @bsudekum @jfirebaugh

jfirebaugh:
Will this flip the problem from dangling pointers to memory leaks? What if the `MGLOfflinePack` isn't removed? What about the packs from `[[MGLOfflineStorage sharedOfflineStorage] getPacksWithCompletionHandler:]`?

1ec5:
These are good points. Maybe then what needs to happen is that MBGLOfflineRegionObserver should unregister itself as the mbgl::OfflineRegion's observer when it detects that the pack has gone away. Then someone following the help document as written will end up with code that will never trigger a call to the delegate methods, but at least it won't crash.

Still, we won't have an answer for situations where the application wants to follow the traditional delegate pattern. In that paradigm, the delegate doesn't need to separately hold strong references to offline packs but should still have access to the packs via the first parameter of the delegate methods.

1ec5:
In other words, we've designed an API that makes it look like MGLOfflineStorage owns MGLOfflinePacks, but in reality they're owned by the class that uses MGLOfflineStorage, which is often an MGLOfflinePackDelegate in practice.

1ec5:
Closing in favor of the safer fix in #4293.

1ec5:
#4293 isn’t enough: if the surrounding class fails to capture a strong reference to the pack from within the completion handler, the pack’s underlying `mbgl::OfflineRegion` does begin downloading without crashing, but the delegate is never notified because the MGLOfflinePack instance itself is deallocated immediately after the completion handler finishes executing.

1ec5:
@jfirebaugh is correct to be concerned about memory leaks, because currently only one object ever gets notified about progress updates and state changes for a given offline pack. There’s no way for MGLOfflineStorage to know what packs are active at a given time without preventing other classes from getting that information.

[...]

PR #4293, opened 2016-03-12
Remove extraneous nil check from MGLOfflineRegionObserver

1ec5:
Due to ARC, MGLOfflinePack can be deallocated at any time following its initialization. `-dealloc` ensures that the offline region observer is unset, but that step is asynchronous; there is no guarantee that it completes before the first status change update. In fact, the pack is deallocated before the first status change update unless it is captured outside of the MGLOfflinePackAdditionCompletionHandler. The pack nil check is unnecessary because the subsequent code would be a no-op.

This is an alternative fix for #4287.

/cc @jfirebaugh @boundsj @bsudekum


1ec5:
Given the call to `mbgl::DefaultFileSource::setOfflineRegionObserver()` inside `-dealloc`, there won’t be a memory leak because `MBGLOfflineRegionObserver` only holds a weak reference to MGLOfflinePack. However, @bsudekum was seeing that the `pack` pointer would sometimes be garbage rather than `nil` when encountering this race condition. This is surprising because `pack` is a weak pointer.

Issue #3801, opened 2016-02-03
port background rendering from -js


ansis:
https://github.com/mapbox/mapbox-gl-js/blob/master/js/render/draw_background.js

This is needed to make background patterns work correctly in perspective view.

PR #4303, opened 2016-03-14
Fix background pattern

kkaefer:
Currently, background pattern doesn't work, because it's always painted above other non-translucent layers.


kkaefer:
More fixes are needed: #3801

ansis:
replaced by https://github.com/mapbox/mapbox-gl-native/pull/4499

PR #4329, opened 2016-03-15
Copy .der files into static framework

### LINKED ISSUE #4316: [iOS] static framework does not include .der files
On master, make ipackage builds the dynamic and static framework. In the static framework directory, there is no .der files as in the dynamic one.
Is it an error or the .der files are not necessary ?

1ec5:
#4104 deleted redundant .der files from the iOS resources directory, exposing a bug in which we failed to copy the shared resources directory into the static framework bundle.

Fixes #4316.

/cc @RomainQuidet


1ec5:
iOS and Android builds passed. Rebasing and merging now.

PR #4325, opened 2016-03-15
Reduce ClipID usage

kkaefer:
We're currently allocating Clip IDs for tiles we don't actually need them for, in particular for tiles that are empty (e.g. empty annotation tiles), and for raster tiles that don't need clipping.

- [x] don't allocate clip IDs for empty tiles (e.g. from annotations)
- [x] don't allocate clip IDs for raster tiles; we don't need to clip them when they're opaque anyway

### DIFF
--- a/src/mbgl/renderer/bucket.hpp
+++ b/src/mbgl/renderer/bucket.hpp
@@ -38,6 +38,8 @@ public:
 
     virtual bool hasData() const = 0;
 
+    virtual bool needsClipping() const = 0;
+
     inline bool needsUpload() const {
         return !uploaded;
     }
--- a/src/mbgl/renderer/circle_bucket.cpp
+++ b/src/mbgl/renderer/circle_bucket.cpp
@@ -31,6 +31,10 @@ bool CircleBucket::hasData() const {
     return !triangleGroups_.empty();
 }
 
+bool CircleBucket::needsClipping() const {
+    return true;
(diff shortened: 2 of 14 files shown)


brunoabinader:
This is great @kkaefer!

mb12:
@kkaefer and @brunoabinader  This change seems incorrect for raster tiles for the following scenario. Can you please clarify how the following will work correctly with no clipping of raster tiles? 

 If all the tiles covering a zoom level have not been loaded, one of the parent tiles would have to be overzoomed and clipped to the boundary of tile being covered by the overzoomed tile. For e.g. if there are four tiles covering a viewport at zoom 15, but only three of them have been loaded. The fourth tile would have to be created by overzooming a parent tile and clipping it to the boundary of fourth tile.

kkaefer:
@mb12 in that situation, the z15 tiles would be drawn first (reversed bucket order during opaque pass), but they also draw into the depth buffer. Once we draw the z14 tile, the depth test fails for the regions that already have z15 tiles, so the z14 tile is only drawn in locations where no raster data exists yet.

kkaefer:
@mb12 @jfirebaugh any objections to merging?

jfirebaugh:
🚢

mb12:
Looks good. Thank you very much @kkaefer  for answering my question in detail.

Issue #4330, opened 2016-03-15
[Android] Crash when starting app in offline mode


erikandre:
When starting an app that that is using the latest 4.0.0 snapshot build (com.mapbox.mapboxsdk:mapbox-android-sdk:4.0.0-SNAPSHOT@aar) of the Mapbox Android SDK I'm getting this crash. It can be avoided if I very quickly move to an area for which I already have downloaded offline map data, but as soon as I leave that area it will crash.

The stacktrace below is coming from a virtual Genymotion device running Android 5.10 (x86).

```
W/com.mapbox.mapboxsdk.http.HTTPRequest(22745): [HTTP] Request could not be executed: failed to connect to /192.168.0.4 (port 8080) after 10000ms
E/mbgl    (22745): {Map}[Setup]: loading style failed: failed to connect to /192.168.0.4 (port 8080) after 10000ms
F/libc    (22745): Fatal signal 6 (SIGABRT), code -6 in tid 22780 (Map Thread)
I/DEBUG   (   84): *** *** *** *** *** *** *** *** *** *** *** *** *** *** *** ***
I/DEBUG   (   84): Build fingerprint: 'generic/vbox86p/vbox86p:5.1/LMY47D/buildbot11172007:userdebug/test-keys'
I/DEBUG   (   84): Revision: '0'
I/DEBUG   (   84): ABI: 'x86'
I/DEBUG   (   84): pid: 22745, tid: 22780, name: Map Thread  >>> org.erikandre.sverigekartangl <<<
I/DEBUG   (   84): signal 6 (SIGABRT), code -6 (SI_TKILL), fault addr --------
I/DEBUG   (   84):     eax 00000000  ebx 000058d9  ecx 000058fc  edx 00000006
I/DEBUG   (   84):     esi e1d07db8  edi 0000000b
I/DEBUG   (   84):     xcs 00000023  xds 0000002b  xes 0000002b  xfs 000000b7  xss 0000002b
I/DEBUG   (   84):     eip f7525ef6  ebp 000058fc  esp e1d04200  flags 00200282
I/DEBUG   (   84):
I/DEBUG   (   84): backtrace:
I/DEBUG   (   84):     #00 pc 00073ef6  /system/lib/libc.so (tgkill+22)
I/DEBUG   (   84):     #01 pc 000221ab  /system/lib/libc.so (pthread_kill+155)
I/DEBUG   (   84):     #02 pc 00023a24  /system/lib/libc.so (raise+36)
[...]

Issue #4202, opened 2016-03-05
Crash When OfflineRegion.delete() Is Called


bleege:
When I try to delete a previously saved Offline Region using the `OfflineRegion.delete()` method it always crashes.  The last time this happened the following (super undetailed) crash report was shown:

```java
03-04 19:02:13.996 23931-25068/com.mapbox.sirius A/libc: Fatal signal 6 (SIGABRT), code -6 in tid 25068 (Offline Thread)
```

/cc @zugaldia @jfirebaugh

jfirebaugh:
More complete crash report:

```
03-04 02:38:09.621 16028-16251/com.mapbox.mapboxsdk.testapp A/libc: Fatal signal 11 (SIGSEGV), code 1, fault addr 0xdeadcab1 in tid 16251 (Offline Thread)
03-04 02:38:09.737 963-963/? A/DEBUG: *** *** *** *** *** *** *** *** *** *** *** *** *** *** *** ***
03-04 02:38:09.738 963-963/? A/DEBUG: Build fingerprint: 'Android/sdk_google_phone_x86/generic_x86:6.0/MASTER/2469028:userdebug/test-keys'
03-04 02:38:09.738 963-963/? A/DEBUG: Revision: '0'
03-04 02:38:09.738 963-963/? A/DEBUG: ABI: 'x86'
03-04 02:38:09.738 963-963/? A/DEBUG: pid: 16028, tid: 16251, name: Offline Thread  >>> com.mapbox.mapboxsdk.testapp <<<
03-04 02:38:09.738 963-963/? A/DEBUG: signal 11 (SIGSEGV), code 1 (SEGV_MAPERR), fault addr 0xdeadcab1
03-04 02:38:09.742 963-963/? A/DEBUG:     eax a3840100  ebx a3970cc0  ecx a3840100  edx 00000001
03-04 02:38:09.742 963-963/? A/DEBUG:     esi 9fc7ef40  edi 9fc7efb0
03-04 02:38:09.742 963-963/? A/DEBUG:     xcs 00000073  xds 0000007b  xes 0000007b  xfs 000000ff  xss 0000007b
03-04 02:38:09.742 963-963/? A/DEBUG:     eip a3840116  ebp 9fc7eeb8  esp 9fc7eea0  flags 00210296
[...]

Issue #3280, opened 2015-12-14
Map crashes in airplane mode (iOS v3.0.1).


DejanRistic:
The map will crash every time when initializing when in airplane mode if the map has been loaded at least once. The map is just black if its a fresh install with airplane mode, but if its been loaded once before we get a crash. 

Error: 
```
libc++abi.dylib: terminating with uncaught exception of type mbgl::util::SpriteLoadingException: Failed to load [mapbox://sprites/roadtrippers/cignsecsw001aa6ksebcdpmrl@2x.json]: The operation couldn’t be completed. (NSURLErrorDomain error -1009.)
```

StackTrace:
```
* thread #19: tid = 0x21a7, 0x31af6dfc libsystem_kernel.dylib`__pthread_kill + 8, name = 'Map', stop reason = signal SIGABRT
  * frame #0: 0x31af6dfc libsystem_kernel.dylib`__pthread_kill + 8
    frame #1: 0x31b74d36 libsystem_pthread.dylib`pthread_kill + 62
    frame #2: 0x31a96908 libsystem_c.dylib`abort + 76
    frame #3: 0x30dd29c8 libc++abi.dylib`abort_message + 88
    frame #4: 0x30dec63a libc++abi.dylib`default_terminate_handler() + 214
    frame #5: 0x314b0f48 libobjc.A.dylib`_objc_terminate() + 228
    frame #6: 0x30de9de2 libc++abi.dylib`std::__terminate(void (*)()) + 78
    frame #7: 0x30de9ee0 libc++abi.dylib`std::terminate() + 156
    frame #8: 0x30d8d16e libc++.1.dylib`std::rethrow_exception(std::exception_ptr) + 14
    frame #9: 0x004736f4 Roadtrippers`mbgl::Style::emitResourceLoadingFailed(std::exception_ptr) + 108
    frame #10: 0x00473be0 Roadtrippers`non-virtual thunk to mbgl::Style::onSpriteLoadingFailed(std::exception_ptr) + 92
    frame #11: 0x00457fca Roadtrippers`mbgl::SpriteStore::emitSpriteLoadingFailed(std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char> > const&) + 234
    frame #12: 0x00458910 Roadtrippers`___lldb_unnamed_function327$$Roadtrippers + 464
    frame #13: 0x0045d6f8 Roadtrippers`mbgl::Request::notifyCallback() + 156
[...]

PR #4339, opened 2016-03-15
[core] Assert and guard against null std::exception_ptr in toString

jfirebaugh:
It appears that if `std::rethrow_exception` is called with a null `std::exception_ptr`, it calls `std::terminate`. Such a crash was confirmed in #4202 and may be the cause of the crashes in #3280 and #4330.

We don't expect `util::toString` to be called with a null `std::exception_ptr` and would like to be notified in debug builds if that happens. But we'd also prefer not to terminate the process in release builds. Therefore, split the difference and add both an `assert` and a guard clause.

#4330 is on the `android-v4.0.0` milestone, so targeting this change at `release-ios-3.2.0-android-4.0.0`.

:eyes: @zugaldia @1ec5 @kkaefer

### DIFF
--- a/include/mbgl/util/string.hpp
+++ b/include/mbgl/util/string.hpp
@@ -28,6 +28,12 @@ inline std::string toString(uint8_t num) {
 }
 
 inline std::string toString(std::exception_ptr error) {
+    assert(error);
+
+    if (!error) {
+        return "(null)";
+    }
+
     try {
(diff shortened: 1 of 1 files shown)


1ec5:
This was committed as 46830f1ea5e1054d35466731345e8b1f2b5dd380.

PR #4342, opened 2016-03-16
[core] Implement a vacuum strategy for the offline database

### LINKED ISSUE #4340: Better vacuum behavior for offline database
The offline database currently has no vacuum behavior: once it reaches a certain size on disk, it will never get smaller, even if you delete offline regions. (It will however reuse emptied pages within the database.)

If the database size becomes larger than the target size set by the `maximumCacheSize` parameter (default 50 MB), the following warning message is printed:

> Current size is larger than the maximum size. Database won't get truncated.

This behavior was inherited from the previous cache database implementation. The original reasoning was that [SQLite `VACUUM`](https://sqlite.org/lang_vacuum.html) was an expensive operation. With offline functionality now included, it's time to revisit this decision. If you create a large offline region, and then delete it, the on disk-size should return to roughly the previous level.

One approach would be to enable [`auto_vacuum`](https://sqlite.org/pragma.html#pragma_auto_vacuum).

jfirebaugh:
Enable `PRAGMA auto_vacuum = INCREMENTAL`, and perform a `PRAGMA incremental_vacuum` when deleting an offline region.

I'm targeting this for the release because enabling auto vacuum on an existing database has some pitfalls, and doing before offline sees widespread adoption will minimize the number of affected databases.

Fixes #4340.

:eyes: @tmpsantos @1ec5


platform/default/mbgl/storage/offline_database.cpp
```diff
@@ -458,6 +466,7 @@ void OfflineDatabase::deleteRegion(OfflineRegion&& region) {
     stmt->run();
 
     evict(0);
+    db->exec("PRAGMA incremental_vacuum");
```
tmpsantos: I think calling vacuum should be a separated API and not called automatically. I can think of 2 use cases for that:
 - In your app you want to remove several regions. Each deletion would be super slow because the database would be recreated on each VACUUM. Instead, the interface would delete first and at the end call `OffilineDatabase::vacuum()`
 - This API could be used for vacuum the dangling pages from the cache.

tmpsantos:
@jfirebaugh :+1: + a comment.

platform/default/mbgl/storage/offline_database.cpp
jfirebaugh: Note that `PRAGMA incremental_vacuum` is not the same as `VACUUM`. In particular, my interpretation of the documentation is that `PRAGMA incremental_vacuum` is much cheaper: it works with a page structure that has already been prepared for incremental vacuum via `PRAGMA auto_vacuum`, and it doesn't compact partially filled pages of the database nor rewrite the entire database file like `VACUUM` does. I am not sure that a single `PRAGMA incremental_vacuum` after deleting several regions would be significantly cheaper than calling `PRAGMA incremental_vacuum` after each deletion. In fact, it might block other database commands for a longer continuous period of time.

Regarding the second point, we expect there the number of free pages for the ambient cache to be low. The expected behavior is that the ambient cache fills up to the maximum allotted size, and then stays at that size. Vacuuming free pages from it would in fact be a deoptimization, because those pages would just need to be reallocated as soon as the cache filled back up to its maximum size.

I'd prefer to keep the vacuum strategy a private implementation detail for now. There are certainly some unknowns about how it will perform, but it's not clear to me that providing API control will help.

tmpsantos: Thanks for clarifying.

PR #4343, opened 2016-03-16
[core] Stop offline download at tile limit

boundsj:
The tile limit guard (when used) stops a download from continuing when the tile limit is reached. This wraps the guard in a method and employs it in both places currently necessary to ensure the guard has a chance to function.

cc @jfirebaugh @1ec5

### DIFF
--- a/platform/default/mbgl/storage/offline_download.cpp
+++ b/platform/default/mbgl/storage/offline_download.cpp
@@ -229,12 +229,8 @@ void OfflineDownload::ensureResource(const Resource& resource, std::function<voi
 
             return;
         }
-
-        if (resource.kind == Resource::Kind::Tile
-            && util::mapbox::isMapboxURL(resource.url)
-            && offlineDatabase.offlineMapboxTileCountLimitExceeded()) {
-            observer->mapboxTileCountLimitExceeded(offlineDatabase.getOfflineMapboxTileCountLimit());
-            setState(OfflineRegionDownloadState::Inactive);
+        
+        if (checkTileCountLimit(resource)) {
             return;
         }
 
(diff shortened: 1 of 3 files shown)


platform/default/mbgl/storage/offline_download.hpp
```diff
@@ -49,7 +49,8 @@ class OfflineDownload {
      */
     void ensureResource(const Resource&, std::function<void (Response)> = {});
     void ensureTiles(SourceType, uint16_t, const SourceInfo&);
-
+    bool hasExceededTileCount(const Resource& resource);
```
jfirebaugh: This method has side effects, but `hasExceededTileCount` sounds like the name of a pure function. How about `checkTileCountLimit`?

platform/default/mbgl/storage/offline_download.cpp
```diff
@@ -263,4 +263,18 @@ void OfflineDownload::ensureResource(const Resource& resource, std::function<voi
...
 
+bool OfflineDownload::hasExceededTileCount(const Resource& resource) {
+    bool tileCountExceeded = false;
+    
+    if (resource.kind == Resource::Kind::Tile
+        && util::mapbox::isMapboxURL(resource.url)
+        && offlineDatabase.offlineMapboxTileCountLimitExceeded()) {
+        observer->mapboxTileCountLimitExceeded(offlineDatabase.getOfflineMapboxTileCountLimit());
+        setState(OfflineRegionDownloadState::Inactive);
+        tileCountExceeded = true;
```
jfirebaugh: MBGL core convention is to prefer not to use temporary variables if possible, so use `return true;` here and `return false;` down below, and remove the `tileCountExceeded` variable.

platform/default/mbgl/storage/offline_download.cpp
```diff
@@ -250,12 +246,16 @@ void OfflineDownload::ensureResource(const Resource& resource, std::function<voi
             if (callback) {
                 callback(onlineResponse);
             }
-
+
```
jfirebaugh: Avoid the whitespace change.

jfirebaugh:
Good catch! You're completely correct about the logic here. In addition to the code adjustment, we should add a regression test. Want to give that a shot?

jfirebaugh:
@boundsj The test and implementation changes look great. I checked them out locally and verified that the test fails before the change and passes after. :ship:

boundsj:
Thanks @jfirebaugh. Thanks for pointing out how the tests can hang, too. I made sure the new test will fail (and not hang in the run loop) if the required implementation is removed.

boundsj:
Merged in https://github.com/mapbox/mapbox-gl-native/commit/bcb3bb0931cbe0a85cb0f4ee9a311eceec7689c7

PR #4347, opened 2016-03-16
Make constants constexpr

kkaefer:
We're currently declaring all of our constants in `constants.hpp` as `extern` and actually initialize them in the corresponding C++ file. Instead, we should declare them as constexpr and assign the value in the header file directly. The fact that they are defined elsewhere is an old debugging construct that allowed me to quickly change tile sizes without recompilation.


brunoabinader:
:+1:

PR #4349, opened 2016-03-16
[iOS] Use CFBundleIdentifier instead of CFBundleName in UA string

boundsj:
Although often recommended, CFBundleName is not guaranteed to exist in an app's configuration. As an identifier of telemetry data, even when it is present it can be inconsistent because of localization.

CFBundleIdentifier, although possibly longer, will be consistent across locales and for the lifetime of an app in the App Store even if the app name changes in any locale.

cc @1ec5 @mick @camilleanne

### DIFF
--- a/platform/ios/src/MGLAPIClient.m
+++ b/platform/ios/src/MGLAPIClient.m
@@ -115,7 +115,7 @@ static NSString * const MGLAPIClientHTTPMethodPost = @"POST";
 }
 
 - (void)setupUserAgent {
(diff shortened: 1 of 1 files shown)


1ec5:
A bundle identifier makes sense as a fallback in the absence of an application name, but I'm not sure we need to eliminate the name entirely.

1ec5:
Per chat, we’re going to go with the bundle identifier to ensure unique user agent strings for every application that uses the SDK and avoid complications around localization.

Issue #3802, opened 2016-02-03
LatLng and LatLngBounds should enforce invariants


jfirebaugh:
Remove `operator bool()` from both classes and enforce the following invariants.
- [ ] **`LatLng`** constructor _throws_ if any of the following are violated:
  - `std::isfinite(latitude)`
  - `std::isfinite(longitude)`
  - `-90 <= latitude <= 90`
- [x] **`LatLngBounds`** constructor _swaps coordinates_ as necessary to enforce:
  - `sw.latitude <= ne.latitude`
  - `sw.longitude <= ne.longitude`

jfirebaugh:
It's useful to have an "empty" state for `LatLngBounds`: #3803. We can keep a `bool empty() const` method around for that.

jfirebaugh:
Just discovered an example of how dangerous a non-explicit `operator bool` can be: currently `LatLng(a, b) == LatLng(c, d)` is `true` for any non-NaN a, b, c, d.

PR #4355, opened 2016-03-17
Remove LatLng::null, vec2::null, and operator bool

jfirebaugh:
Instead of using NaN, 0 or other special values to indicate the lack of a real `LatLng` or `vec2` value, we should use `optional<LatLng>` or `optional<vec2>`.

Refs #3802, #4285.

cc @brunoabinader

### DIFF
--- a/include/mbgl/util/geo.hpp
+++ b/include/mbgl/util/geo.hpp
@@ -15,17 +15,19 @@ using ScreenCoordinate = vec2<double>;
 
 class LatLng {
 public:
-    struct null {};
-
     double latitude;
     double longitude;
 
     enum WrapMode : bool { Unwrapped, Wrapped };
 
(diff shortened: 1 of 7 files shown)


brunoabinader:
:+1: Glad to see one of the best idioms I learned while learning Rust (together with type traits) in C++!

include/mbgl/util/geo.hpp
```diff
@@ -65,12 +63,8 @@ class ProjectedMeters {
     double northing = 0;
     double easting = 0;
 
-    ProjectedMeters(double n = 0, double e = 0)
+    ProjectedMeters(double n, double e)
```
brunoabinader: Shouldn't `ProjectedMeters` ctor params have the same NaN/infinite checks as in `LatLng`?

PR #4372, opened 2016-03-17
Place offline cache in Application Support

### LINKED ISSUE #4371: Storing offline.db in user caches directory bad idea on OS X
The caches directory may get purged by the system at any time. It is not hard to imagine a case where a user is offline, depending on their offline map regions, and fills up their device in some other way (e.g., taking a lot of photos). In such a case the system could and would purge the user cache directory and remove the Mapbox SDK's offline db.

See: https://github.com/mapbox/mapbox-gl-native/blob/1ee7264b0edf6601a8170a30c40af4282679f495/platform/darwin/src/MGLOfflineStorage.mm#L33-L34

1ec5:
Unlike the ambient cache maintained by MGLMapView, the offline cache should live in a subdirectory of Application Support, where the SDK has control over the file’s lifetime. The subdirectory is already named after the host application’s bundle identifier, ensuring that each Mapbox-powered application has an independent tile limit.

Fixes #4371.

/cc @jfirebaugh

### DIFF
--- a/platform/darwin/src/MGLOfflineStorage.mm
+++ b/platform/darwin/src/MGLOfflineStorage.mm
@@ -33,7 +33,7 @@
         NSArray *paths = NSSearchPathForDirectoriesInDomains(NSDocumentDirectory, NSUserDomainMask, YES);
         NSString *cachePath = [paths.firstObject stringByAppendingPathComponent:fileName];
 #elif TARGET_OS_MAC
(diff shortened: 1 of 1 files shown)


jfirebaugh:
I don't understand why you are still putting "ambient cache" and "offline cache" in different categories. They need to be the same file, or "offline" doesn't work.

PR #4375, opened 2016-03-18
[iOS] Increase distance filter

boundsj:
This increases the distance filter to five meters. The additional location data captured at 1 meter is not required.

cc @1ec5

### DIFF
--- a/platform/ios/src/MGLLocationManager.m
+++ b/platform/ios/src/MGLLocationManager.m
@@ -1,9 +1,10 @@
 #import "MGLLocationManager.h"
 #import <UIKit/UIKit.h>
 
-static const NSTimeInterval fiveMinuteTimeInterval = 300.0;
-static const NSTimeInterval fiveSecondTimeInterval = 5.0;
-static const CLLocationDistance regionRadiusLocationDistance = 300.0;
+static const NSTimeInterval MGLLocationManagerHibernationTimeout = 300.0;
+static const NSTimeInterval MGLLocationManagerHibernationPollInterval = 5.0;
+static const CLLocationDistance MGLLocationManagerHibernationRadius = 300.0;
+static const CLLocationDistance MGLLocationManagerDistanceFilter = 5.0;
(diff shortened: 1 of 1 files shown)


platform/ios/src/MGLLocationManager.m
```diff
@@ -1,9 +1,10 @@
 #import "MGLLocationManager.h"
 #import <UIKit/UIKit.h>
 
-static const NSTimeInterval fiveMinuteTimeInterval = 300.0;
-static const NSTimeInterval fiveSecondTimeInterval = 5.0;
-static const CLLocationDistance regionRadiusLocationDistance = 300.0;
+static const NSTimeInterval MGLLocationManagerFiveMinuteTimeInterval = 300.0;
```
1ec5: While we’re at it, can we give these constants more meaningful names? If not, they should be inlined.

1ec5:
:+1: aside from the one comment.

platform/ios/src/MGLLocationManager.m
```diff
@@ -1,9 +1,9 @@
 #import "MGLLocationManager.h"
 #import <UIKit/UIKit.h>
 
-static const NSTimeInterval MGLLocationManagerFiveMinuteTimeInterval = 300.0;
-static const NSTimeInterval MGLLocationManagerFiveSecondTimeInterval = 5.0;
-static const CLLocationDistance MGLLocationManagerRegionRadiusLocationDistance = 300.0;
+static const NSTimeInterval MGLLocationManagerHibernationTimeout = 300.0;
+static const NSTimeInterval MGLLocationManagerHibernationPollInterval = 5.0;
+static const CLLocationDistance MGLLocationManagerHibernationRadius = 300.0;
```
boundsj: @1ec5 these will probably change again in the future but hopefully this is a bit clearer. I'd like to keep them as constants, for now, since this class might someday evolve be something more configurable and be used by other parts of the SDK or even by developers who have similar battery sensitive location manager needs.

1ec5: These names are great, thanks!

Issue #4376, opened 2016-03-18
Add API to clear ambient cache


1ec5:
As of #4376, all cached resources are located in a file in Application Support instead of Caches, so any disk space taken up by ambient caching cannot be automatically reclaimed by iOS when disk space runs low. These resources are only removed when the application is deleted. This would become more of a problem if the application is permitted to raise the ambient cache size (#3437).

Perhaps MGLOfflineStorage should have a method that clears the ambient cache contents in the unified cache file, leaving any downloaded offline content intact. The application could provide UI for reclaiming the space taken up by ambient caching.

/cc @jfirebaugh

zugaldia:
Adding the `android` tag as the current situation affects Android similarly.

tsemerad:
Is there still no way to clear the ambient cache, other than reinstalling the app? I'm finding that when I update a tileset, my iOS client is still holding onto the previous tiles. Previously-cached tiles still show the old data, but fetching new tiles show the new data.

1ec5:
Is this tile set hosted by Mapbox or by you? If you’re hosting your own tile set, you can set a caching header in the HTTP response to force regular refreshes: https://github.com/mapbox/mapbox-gl-native/issues/4446#issuecomment-200583990.

1ec5:
Is this tile set hosted by Mapbox or by you? If you’re hosting your own tile set, you can set a caching header in the HTTP response to force regular refreshes: https://github.com/mapbox/mapbox-gl-native/issues/4446#issuecomment-200583990.

tsemerad:
It's hosted on Mapbox. Do you know what `Cache-Control: max-age` header Mapbox serves the tiles with?

Hosting the tiles myself wouldn't be ideal, but I can do so if necessary. But doesn't the new offline caching functionality (where you set a region) require that the tiles be hosted by Mapbox? Maybe I'm misremembering.

zugaldia:
Just noting that, on Android, this is something that can easily be implemented by the developer today. For reference, we currently have a private method [`deleteAmbientDatabase`](https://github.com/mapbox/mapbox-gl-native/blob/master/platform/android/MapboxGLAndroidSDK/src/main/java/com/mapbox/mapboxsdk/offline/OfflineManager.java#L168-L185) that deletes the database that we used to use before offline landed. (This method name needs to be clarified once this new API is introduced.)

[...]

Issue #4338, opened 2016-03-15
Compiler warnings with installed static framework — Module debugging should be disabled when shipping static libraries


1ec5:
When I install the Mapbox iOS SDK static framework into a project and build in Xcode 7.3b5, I get 21 errors that look like this:

```
while processing /Users/mxn/Documents/Playroom/Sedan/Mapbox.framework/Mapbox(reachability.o):
warning: /var/folders/cl/8cgpg37n5cl78pjkbpgml94w0000gn/C/org.llvm.clang.mxn/ModuleCache/2M9FX5S5W7EA5/Foundation-3DFYNEBRQSXST.pcm: No such file or directory
note: Module debugging should be disabled when shipping static libraries.
```

I’m unaware of any other symptoms, but we should set `CLANG_ENABLE_MODULE_DEBUGGING=NO` when building the static framework with debug symbols:

> When this setting is enabled, clang will use the shared debug info available in clang modules and precompiled headers. Results in smaller build artifacts, faster compile times, and more complete debug info. This should only be disabled when building static libraries with debug info for distribution. [`CLANG_ENABLE_MODULE_DEBUGGING`, `-gmodules`]

/cc @boundsj @friedbunny

boundsj:
@1ec5 I'm unable to repro when building the SDK with Xcode 7.3 and compiling the static Mapbox.framework in client Objective-C or Swift apps in Xcode 7.2.x or Xcode 7.3. I'm also unable to repro the issue when building the SDK with Xcode 7.2.x and compiling in client apps in 7.2 or 7.3.

Closing now and will reopen and look again if a clear repro is found.

PR #4377, opened 2016-03-18
Unify offline and ambient caches

1ec5:
Now there’s only one instance of `mbgl::OfflineFileSource`, created when the shared MGLOfflineStorage object is initialized. Also create and use the shared MGLOfflineStorage object when initializing an MGLMapView object.

The unified cache file is located in a subdirectory of Application Support, where the SDK has control over the file’s lifetime. The subdirectory is already named after the host application’s bundle identifier, ensuring that each Mapbox-powered application has an independent tile limit.

If there’s an ambient cache in a subdirectory of Caches, delete it. If there’s an offline cache in a subdirectory of Documents on iOS or Caches on OS X, move it to the unified cache location in a subdirectory of Application Support.

There are a couple consequences to this approach:

1. On iOS, it isn’t possible for a user to export their offline cache to another device or import it from another device. The application would have to move the file into Documents first.
1. If disk space runs low on the iOS device, the unified cache will continue to eat up disk space, potentially due to ambiently cached content, and the only way to reclaim that space would be to delete the application (along with all the offline content). See #4376 for a possible solution.
1. On OS X, it isn’t possible for two Mapbox-powered applications to share an ambient cache. The unified cache’s path must include the application’s bundle identifier, because otherwise one application’s offline packs would count against another application’s limits.

Fixes the iOS/OS X side of #4338.

/cc @jfirebaugh @boundsj

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -57,9 +57,10 @@ Known issues:
 - Black Screen On Ice Cream Sandwich and Jelly Bean devices ([#2802](https://github.com/mapbox/mapbox-gl-native/issues/2802))
   - Resolved in 2.2.0
 
-## iOS master
+## iOS 3.2.0
 
 - Offline packs can now be downloaded to allow users to view specific regions of the map offline. A new MGLOfflineStorage class provides APIs for managing MGLOfflinePacks. ([#4221](https://github.com/mapbox/mapbox-gl-native/pull/4221))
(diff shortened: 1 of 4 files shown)


jfirebaugh:
:+1: 

I got tripped up by simulator behavior, but after figuring that out, I can confirm that the migration works and offline packs are used when offline.

1ec5:
Builds are passing; merging now.

Issue #4362, opened 2016-03-17
iOS, Android are not using the offline database


jfirebaugh:
iOS creates an offline database named ["offline.db"](https://github.com/mapbox/mapbox-gl-native/blob/ff9c3a57b05fa3287a126ffd7c66cddc7e0929dc/platform/darwin/src/MGLOfflineStorage.mm#L25-L25), but then uses the name ["cache.db"](https://github.com/mapbox/mapbox-gl-native/blob/c2cd9335e9c9ff599cd84723eada6a503d1e5c80/platform/ios/src/MGLMapView.mm#L312-L312) for the actual map view. These need to be the same name, if not the same instance of `DefaultFileSource`.

jfirebaugh:
It looks like the path, as well as the name, is different.

1ec5:
I copied the Android SDK, which is doing the same thing: [mbgl-cache.db](https://github.com/mapbox/mapbox-gl-native/blob/20c3706536ac9e927f80a50e9ec8906dfd0d2baf/platform/android/src/native_map_view.cpp#L79), [mbgl-offline.db](https://github.com/mapbox/mapbox-gl-native/blob/55cacfafd47d81204161e789c4282e0537152591/platform/android/MapboxGLAndroidSDK/src/main/java/com/mapbox/mapboxsdk/offline/OfflineManager.java#L17).

1ec5:
On iOS, per #4371, the unified cache definitely shouldn’t go in the Caches directory, where the ambient cache currently resides. If we put it in the Documents directory, where the offline cache currently resides, a .db file will be listed in iTunes as a shareable document for every Mapbox-powered application, which is a poor user experience. As a compromise, we can put the unified cache in Application Support, but there are two caveats:

1. It isn’t possible for a user to export their offline cache to another device or import it from another device. For example, if a user installs both the iOS and OS X versions of an application, they can currently drag their OS X offline cache into iTunes, and the iOS application should automatically be able to make use of it.
1. If disk space runs low on the iOS device, the unified cache will continue to eat up disk space, potentially due to ambiently cached content, and the only way to reclaim that space would be to delete the application.

1ec5:
#4372 puts the OS X offline cache in a subfolder of Application Support, but neither caveat in https://github.com/mapbox/mapbox-gl-native/issues/4362#issuecomment-198131488 applies to OS X.

1ec5:
[...]

PR #4383, opened 2016-03-18
[android] Unify cache and offline databases

zugaldia:
Per #4362, we're now unifying the ambient and the offline databases in one location. This PR:

1. Moves them both to a permanent storage folder.
2. Deletes any previous ambient database file.

/cc: @jfirebaugh @1ec5

### DIFF
--- a/platform/android/MapboxGLAndroidSDK/src/main/java/com/mapbox/mapboxsdk/offline/OfflineManager.java
+++ b/platform/android/MapboxGLAndroidSDK/src/main/java/com/mapbox/mapboxsdk/offline/OfflineManager.java
@@ -4,6 +4,7 @@ import android.content.Context;
 import android.os.Handler;
 import android.os.Looper;
 import android.support.annotation.NonNull;
+import android.util.Log;
 
 import java.io.File;
 
@@ -13,8 +14,10 @@ import java.io.File;
  */
 public class OfflineManager {
 
+    private final static String LOG_TAG = "OfflineManager";
+
     // Default database name
-    private final static String OFFLINE_DATABASE_NAME = "mbgl-offline.db";
+    private final static String DATABASE_NAME = "mbgl-offline.db";
 
     /*
      * The maximumCacheSize parameter is a limit applied to non-offline resources only,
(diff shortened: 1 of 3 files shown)


1ec5:
Noting for posterity that, with this change, the Android SDK continues to have multiple `DefaultFileSource` instances, whereas the iOS SDK now maintains only one `DefaultFileSource` instance as of #4372. As far as I know, the main consequence is that the limit of 20 simultaneous connections is enforced per map view / offline manager rather than per process.

jfirebaugh:
On the iOS side, #4377 ensures that the beta.1 offline database is preserved. It looks like this PR will not take that measure, since it changes the file name used by `OfflineManager` from "mbgl-offline.db" to "mbgl-cache.db".

Options:

* Keep the PR as-is, and let people know that they'll need to recreate offline regions in beta.2. (If we do this, we should ensure that "mbgl-offline.db" is removed.)
* Restore "mbgl-offline.db" as the name used by `OfflineManager`, and change [the name here](https://github.com/mapbox/mapbox-gl-native/blob/fff13d0976015650c2c79348dd8d51d25ce0b2d1/platform/android/src/native_map_view.cpp#L79) to match.
* Keep using "mbgl-cache.db", but write a migration that moves "mbgl-offline.db" to "mbgl-cache.db".

I like the idea of preserving existing beta.1 offline databases in one of the latter two ways, but I don't feel super strongly about it. What do you think?

zugaldia:
> Restore "mbgl-offline.db" as the name used by OfflineManager, and change the name here to match.

Good catch. Tackled by https://github.com/mapbox/mapbox-gl-native/pull/4383/commits/f5b750b2ad92c0fb255bb285812b8971f5839079.

platform/android/MapboxGLAndroidSDK/src/main/java/com/mapbox/mapboxsdk/offline/OfflineManager.java
```diff
@@ -53,8 +56,30 @@
...
+        deleteAmbientDatabase(context);
+    }
+
+    private void deleteAmbientDatabase(final Context context) {
+        // Delete the file in a separate thread to avoid affecting the UI
+        new Thread(new Runnable() {
+            @Override
+            public void run() {
+                try {
+                    String path = context.getCacheDir().getAbsolutePath() + File.separator + DATABASE_NAME;
```
jfirebaugh: Now `DATABASE_NAME` needs to be replaced with `"mbgl-cache.db"` here.

zugaldia: Yes, indeed: https://github.com/mapbox/mapbox-gl-native/pull/4383/commits/7fcfc2ce9831e56b240c9a79e29e6cd3de3e0e84.

jfirebaugh:
:+1: 

Tested and confirmed that an existing offline region is preserved, and then used when there is not network access.

PR #4390, opened 2016-03-18
[iOS] Send turnstile event as much as once per day

boundsj:
Previously, the post of the turnstile event was a side effect of the
map.load event and would happen when the host app was started from a
terminated state at the first initialization of MGLMapView. The
turnstile event post was guarded with dispatch_once so that it would
only ever be sent once during the lifecycle of the application after
it started from the terminated state.

This changes that behavior by:

- Allowing a new turnstile event to be posted as much as once per
day
- Only triggering a map.view event when a map is initialized if the map
is loaded outside of the 'background' application state
- Sending a map.view event whenever a MGLMapView is instantiated
and the application did become active or enters foreground

iOS turnstile event sending behavior is now:

- A map.view event is generated whenever a map is loaded and also
when a map is shown again when the host app becomes active and moves
into the foreground
- Turnstile events are still triggered by mapviews and will be sent as
frequently as once per day for a host app that is not terminated and
also whenever a host app is terminated and restarted

cc @1ec5

### DIFF
--- a/platform/ios/src/MGLMapView.mm
+++ b/platform/ios/src/MGLMapView.mm
@@ -439,7 +439,9 @@ mbgl::Duration MGLDurationInSeconds(NSTimeInterval duration)
     _pendingLongitude = NAN;
     _targetCoordinate = kCLLocationCoordinate2DInvalid;
 
-    [MGLMapboxEvents pushEvent:MGLEventTypeMapLoad withAttributes:@{}];
+    if ([UIApplication sharedApplication].applicationState != UIApplicationStateBackground) {
+        [MGLMapboxEvents pushEvent:MGLEventTypeMapLoad withAttributes:@{}];
+    }
 }
 
 - (void)createGLView
@@ -982,6 +984,8 @@ mbgl::Duration MGLDurationInSeconds(NSTimeInterval duration)
         _displayLink.paused = NO;
 
         [self validateLocationServices];
+        
+        [MGLMapboxEvents pushEvent:MGLEventTypeMapLoad withAttributes:@{}];
     }
 }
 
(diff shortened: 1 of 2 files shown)


platform/ios/src/MGLMapView.mm
```diff
@@ -450,7 +450,9 @@ - (void)commonInit
     _pendingLongitude = NAN;
     _targetCoordinate = kCLLocationCoordinate2DInvalid;
 
-    [MGLMapboxEvents pushEvent:MGLEventTypeMapLoad withAttributes:@{}];
+    if ([UIApplication sharedApplication].applicationState != UIApplicationStateBackground) {
+        [MGLMapboxEvents pushEvent:MGLEventTypeMapLoad withAttributes:@{}];
+    }
 }
```
boundsj: @1ec5 I kept this in commonInit since, when map view is first initialized it cannot have started to observe life cycle events and therefore will never call `wakeGL:` the first time it is loaded. I did however add the guard so that the 
turnstile event will not be attempted if this happens in the background.

platform/ios/src/MGLMapboxEvents.m
```diff
@@ -314,38 +315,36 @@ - (void)flush {
...
-                                                                                   @"error": error}];
-                return;
-            }
-            [strongSelf writeEventToLocalDebugLog:turnstileEventAttributes];
-        }];
-    });
+        [strongSelf writeEventToLocalDebugLog:turnstileEventAttributes];
+        NSTimeInterval twentyFourHourTimeInterval = 24 * 3600;
+        strongSelf.turnstileSendDate = [[NSDate date] dateByAddingTimeInterval:twentyFourHourTimeInterval];
+    }];
```
boundsj: Note: we **will** not consider a turnstile event to have been sent in a session unless the server responds favorably.

Issue #4387, opened 2016-03-18
MGLOfflineStorage sharedOfflineStorage throws exception


postmechanical:
Calling `[MGLOfflineStorage sharedOfflineStorage]` throws an exception with Mapbox-iOS-SDK `3.2.0-pre.2`:

    <_NSCallStackArray 0x7feb326af210>(
    0   ???                                 0x000000011a0f8a05 0x0 + 4732193285,
    1   App                                 0x000000010407bac0 main + 0,
    2   libc++.1.dylib                      0x000000010534d8ff _ZNSt3__118condition_variable4waitERNS_11unique_lockINS_5mutexEEE + 47,
    3   libc++.1.dylib                      0x000000010534e22e _ZNSt3__117__assoc_sub_state10__sub_waitERNS_11unique_lockINS_5mutexEEE + 46,
    4   libc++.1.dylib                      0x000000010534e189 _ZNSt3__117__assoc_sub_state4copyEv + 43,
    5   libc++.1.dylib                      0x000000010534e3f3 _ZNSt3__16futureIvE3getEv + 25,
    6   Mapbox                              0x000000010779bc88 _ZN4mbgl4util6ThreadINS_17DefaultFileSource4ImplEEC2IJRKNSt3__112basic_stringIcNS6_11char_traitsIcEENS6_9allocatorIcEEEERyEEERKNS0_13ThreadContextEDpOT_ + 216,
    7   Mapbox                              0x00000001077993e2 _ZN4mbgl17DefaultFileSourceC2ERKNSt3__112basic_stringIcNS1_11char_traitsIcEENS1_9allocatorIcEEEES9_y + 130,
    8   Mapbox                              0x00000001077bf0d0 -[MGLOfflineStorage initWithFileName:] + 464,
    9   Mapbox                              0x00000001077beebd __41+[MGLOfflineStorage sharedOfflineStorage]_block_invoke + 45,
    10  libdispatch.dylib                   0x000000010b02649b _dispatch_client_callout + 8,
    11  libdispatch.dylib                   0x000000010b011e28 dispatch_once_f + 543,
    12  Mapbox                              0x00000001077bee86 +[MGLOfflineStorage sharedOfflineStorage] + 102,
    13  App                                 0x0000000104097a17 __30+[MapboxHelper sharedInstance]_block_invoke + 359,
    14  libdispatch.dylib                   0x000000010b02649b _dispatch_client_callout + 8,
[...]

Issue #4382, opened 2016-03-18
[Crash] Map crashes on start


mpuchala:
Map crashes on start - here is my log:

```
03-18 17:41:47.257 461-461/? A/DEBUG: *** *** *** *** *** *** *** *** *** *** *** *** *** *** *** ***
03-18 17:41:47.257 461-461/? A/DEBUG: Build fingerprint: '...'
03-18 17:41:47.257 461-461/? A/DEBUG: Revision: '0'
03-18 17:41:47.258 461-461/? A/DEBUG: ABI: 'arm'
03-18 17:41:47.258 461-461/? A/DEBUG: pid: 23479, tid: 23554, name: ...
03-18 17:41:47.258 461-461/? A/DEBUG: signal 6 (SIGABRT), code -6 (SI_TKILL), fault addr --------
03-18 17:41:47.277 461-461/? A/DEBUG:     r0 00000000  r1 00005c02  r2 00000006  r3 98dcf978
03-18 17:41:47.277 461-461/? A/DEBUG:     r4 98dcf980  r5 98dcf930  r6 00000002  r7 0000010c
03-18 17:41:47.277 461-461/? A/DEBUG:     r8 b90742e0  r9 b90587f0  sl b7ed6200  fp 98dcf2f8
03-18 17:41:47.278 461-461/? A/DEBUG:     ip 00000006  sp 98dcf298  lr b6ce3471  pc b6ce515c  cpsr 400f0010
03-18 17:41:47.297 461-461/? A/DEBUG: backtrace:
03-18 17:41:47.297 461-461/? A/DEBUG:     #00 pc 0004415c  /system/lib/libc.so (tgkill+12)
03-18 17:41:47.297 461-461/? A/DEBUG:     #01 pc 0004246d  /system/lib/libc.so (pthread_kill+32)
03-18 17:41:47.297 461-461/? A/DEBUG:     #02 pc 0001b87f  /system/lib/libc.so (raise+10)
03-18 17:41:47.297 461-461/? A/DEBUG:     #03 pc 00018a31  /system/lib/libc.so (__libc_android_abort+34)
03-18 17:41:47.298 461-461/? A/DEBUG:     #04 pc 00016830  /system/lib/libc.so (abort+4)
03-18 17:41:47.298 461-461/? A/DEBUG:     #05 pc 003f3530  /data/app/.../lib/arm/libmapbox-gl.so
[...]

PR #4395, opened 2016-03-19
[core] Add additional logging for exceptions during database connection / creation

jfirebaugh:
Endeavor to obtain the `what()` from exceptions like #4382 and #4387.

:eyes: @1ec5


1ec5:
:+1:

PR #4400, opened 2016-03-20
Reset the translation to zero without adjusting for view

danwexler:
Pan translations are broken for any MGLMapView with a non-zero frame origin. Fix this by using the 'nil' view when resetting the pan gesture recognizer translation.

To test this bug, set the leading-edge constraint of the map view in the test application to any non-zero value and try to pan the map.


1ec5:
Thanks!

PR #4415, opened 2016-03-21
[glfw] Main loop integration

tmpsantos:
This code will work on both rendering on the Map thread and on
the Main thread and it is needed by #2909.


tmpsantos:
:eyes: @jfirebaugh

jfirebaugh:
:+1:

PR #4420, opened 2016-03-21
upload new symbol buffers in upload pass

### LINKED ISSUE #4131: symbol updates don't get uploaded in the upload pass
When the map is rotated and symbol buffers are recalculated we don't reset `uploaded` so the buffers don't get uploaded during the upload pass.

I think we need an `uploaded = false;` here:
https://github.com/mapbox/mapbox-gl-native/blob/09c91008910746667b26d04eca679dfa5bb6b03b/src/mbgl/renderer/symbol_bucket.cpp#L571

ansis:
fix #4131 

When the map is rotated and symbol buffers are recalculated we don't reset `uploaded` so the buffers don't get uploaded during the upload pass. This fixes that by resetting `uploaded`.


ansis:
@brunoabinader can you review and merge this?

brunoabinader:
Looks good @ansis :+1:

PR #4427, opened 2016-03-22
Fix up center coordinate after lopsided pinch gesture

### LINKED ISSUE #4315: One-finger pinch gesture moves zoom origin in opposite direction
- **Steps to reproduce:** put two fingers on the map and then pinch by moving one of the fingers
- **Expected behavior:** geographic locations under both fingers stay approximately the same as the map zooms
- **Actual behavior:** the locations are off under the fingers, so it appears to be using an incorrect zoom origin

You can confirm that this is a bug by doing the same in Apple/Google maps — they behave correctly in this case.

I'm not sure whether this is reproducible on Android since I'm only testing on iOS, but it's worth making sure too. cc @bleege 

cc @1ec5

1ec5:
The gesture recognizer only reports the gesture’s current center point, so use the previous center point to anchor the transition.

Fixes #4315.

/cc @mourner @friedbunny


friedbunny:
This gets a :+1: from me — the coordinate under the finger that’s held on the screen can still change (based on rotation and if you move from high to low zoom), but this is essentially the same behavior as Maps.app.

PR #4431, opened 2016-03-22
[core] Style class property transitions are optional

brunoabinader:
We should use `optional<PropertyTransition>` following the trend on Map API.

/cc @jfirebaugh

### DIFF
--- a/src/mbgl/map/map_context.cpp
+++ b/src/mbgl/map/map_context.cpp
@@ -297,15 +297,15 @@ bool MapContext::hasClass(const std::string& className) const {
     return style->hasClass(className);
 }
 
-void MapContext::addClass(const std::string& className, const PropertyTransition& properties) {
+void MapContext::addClass(const std::string& className, optional<PropertyTransition> properties) {
     if (style->addClass(className, properties)) updateAsync(Update::Classes);
 }
 
-void MapContext::removeClass(const std::string& className, const PropertyTransition& properties) {
+void MapContext::removeClass(const std::string& className, optional<PropertyTransition> properties) {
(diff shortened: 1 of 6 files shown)


src/mbgl/map/map_context.cpp
```diff
@@ -297,15 +297,15 @@ bool MapContext::hasClass(const std::string& className) const {
...
 }
 
-void MapContext::removeClass(const std::string& className, const PropertyTransition& properties) {
+void MapContext::removeClass(const std::string& className, optional<PropertyTransition> properties) {
     if (style->removeClass(className, properties)) updateAsync(Update::Classes);
 }
 
-void MapContext::setClasses(const std::vector<std::string>& classNames, const PropertyTransition& properties) {
+void MapContext::setClasses(const std::vector<std::string>& classNames, optional<PropertyTransition> properties) {
     style->setClasses(classNames, properties);
```
tmpsantos: The crash is because style could be `null` here.

brunoabinader: This is not related to the style crash - this is an API cosmetics.

jfirebaugh:
It looks to me like `PropertyTransition` is like `CameraOptions`: a set of parameters, each of which is individually `optional`. I'm sure that making the object as a whole `optional` is beneficial.

jfirebaugh:
Edit: missed a "not".

brunoabinader:
> It looks to me like PropertyTransition is like CameraOptions: a set of parameters, each of which is individually optional. I'm not sure that making the object as a whole optional is beneficial.

Agreed - this was mostly cosmetics to keep up to par with the other functions available in `Map`, but your argument on its members being `optional` already makes this irrelevant.

PR #4433, opened 2016-03-22
Place unit test fixtures in correct place when creating iOS project

friedbunny:
`make iproj` was producing a project with a broken `$()../test` folder containing a missing `fixtures` item. 

@kkaefer The "gyp doesn't prepend $(SRCROOT)” hack you mention doesn’t appear to be necessary — at least not on my system, running Xcode 7.3.

screen-shot-2016-03-22-at-6_15_29-pm

Refs https://github.com/mapbox/mapbox-gl-native/commit/c587eefbb1290cc01e3e14a0ae18a2d316a498fd#diff-6df1c496c702460f8a63d53bce86b8c7R139

### DIFF
--- a/test/test.gypi
+++ b/test/test.gypi
@@ -136,8 +136,7 @@
           },
           'copies': [
             { 'destination': '<(PRODUCT_DIR)/$(WRAPPER_NAME)/test',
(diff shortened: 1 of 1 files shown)


friedbunny:
I can’t verify that this allows unit tests to pass on iOS because #4434 fails before they can complete.

PR #4235, opened 2016-03-24
Unify FileRequest and WorkRequest

jfirebaugh:
* Replace `FileRequest` with a more generically named `AsyncRequest`.
* `WorkRequest` should inherit from `AsyncRequest` and be an implementation detail. `RunLoop::invoke{Cancellable,WithCallback}` should return `std::unique_ptr<AsyncRequest>`.

This will allow code that makes either or both kinds of async requests to treat them uniformly.


jfirebaugh:
:eyes: @tmpsantos

tmpsantos:
:+1:

Saw it just now, I was on vacations :surfer:

PR #4455, opened 2016-03-24
[ios Fix static lib issues for signed device builds and iTunes connect uploads

### LINKED ISSUE #4118: Cannot submit app with static lib to App Store
When trying to upload to the App Store, I get the following message:

`Couldn't find platform family in Info.plist CFBundleSupportedPlatforms or Mach-O LC_VERSION_MIN for Mapbox.framework.`

I'm using the static library (version 3.1.0) since we need iOS 7 support.

boundsj:
Add a flavor of static build that:

- Puts resources in a Mapbox.bundle file
- Puts Mapbox.bundle and Settings.bundle at the same level as Mapbox.framework in the /static output folder
- Removes CFBundleExecutable and CFBundlePackageType from the Info.plist in Mapbox.bundle that cause errors when uploading to iTunes Connect

This also updates the `binary` instructions in INSTALL.md to reflect these changes.

Fixes #4118

cc @1ec5

### DIFF
--- a/platform/ios/INSTALL.md
+++ b/platform/ios/INSTALL.md
@@ -95,7 +95,15 @@ If your application targets iOS 7.x, you’ll need to install the static framewo
 
 1. Build from source manually per above.
 
-1. Open the project editor and select your application target. Drag `build/ios/pkg/static/Mapbox.framework` into the “Embedded Binaries” section of the General tab. (Don’t drag it into the “Linked Frameworks and Libraries” section; Xcode will add it there automatically.) In the sheet that appears, make sure “Copy items if needed” is checked, then click Finish.
+1. Drag the Mapbox.bundle and Mapbox.framework files in `build/ios/pkg/static` into the Project navigator, checking "Copy items if needed". It should happen automatically, but ensure that:
+   
+   - `Mapbox.framework` is listed in your `Link Binary With Libraries` build phase.
+   - Your *Framework Search Paths* (`FRAMEWORK_SEARCH_PATHS`) build setting includes the directory that contains `Mapbox.framework`. For most projects, the default value of `$(inherited) $(PROJECT_DIR)` should be sufficient.  
+   - `Mapbox.bundle` is in your target's *Copy Bundle Resources* build phase.
+
+1. **Optional** As [noted in the documentation](https://www.mapbox.com/ios-sdk/#telemetry_opt_out) you may use the provided Settings.bundle to provide a Telemetry opt out for users. To use the provided file, drag the `Settings.bundle` file in `build/ios/pkg` into the Project navigator, checking "Copy items if needed". It should happen automatically, but ensure that:
+
+   - `Settings.bundle` is in your target's *Copy Bundle Resources* build phase. 
 
 1. Add the following Cocoa Touch frameworks and libraries to the “Linked Frameworks and Libraries” section:
 
--- a/platform/ios/scripts/package.sh
+++ b/platform/ios/scripts/package.sh
@@ -27,13 +27,28 @@ if [[ ${BUNDLE_RESOURCES} ]]; then
     BUNDLE_PATH="/${NAME}.bundle"
 fi
 
+PLACE_RESOURCE_BUNDLES_OUTSIDE_FRAMEWORK=${PLACE_RESOURCE_BUNDLES_OUTSIDE_FRAMEWORK:-}
+STATIC_BUNDLE_PATH=
+if [[ ${PLACE_RESOURCE_BUNDLES_OUTSIDE_FRAMEWORK} ]]; then
+    STATIC_BUNDLE_PATH="${OUTPUT}/static${BUNDLE_PATH}"
+else
+    STATIC_BUNDLE_PATH="${OUTPUT}/static/${NAME}.framework${BUNDLE_PATH}"
+fi
+
+STATIC_SETTINGS_DIRECTORY=
+if [[ ${PLACE_RESOURCE_BUNDLES_OUTSIDE_FRAMEWORK} ]]; then
+    STATIC_SETTINGS_DIRECTORY="${OUTPUT}"
+else
+    STATIC_SETTINGS_DIRECTORY="${OUTPUT}/static/${NAME}.framework"
+fi
+
 SDK=iphonesimulator
 if [[ ${BUILD_FOR_DEVICE} == true ]]; then
     SDK=iphoneos
 fi
 IOS_SDK_VERSION=`xcrun --sdk ${SDK} --show-sdk-version`
 
-echo "Configuring ${FORMAT:-dynamic and static} ${BUILDTYPE} framework for ${SDK}; symbols: ${GCC_GENERATE_DEBUGGING_SYMBOLS}; Bitcode: ${ENABLE_BITCODE}; Mapbox.bundle: ${BUNDLE_RESOURCES}"
+echo "Configuring ${FORMAT:-dynamic and static} ${BUILDTYPE} framework for ${SDK}; symbols: ${GCC_GENERATE_DEBUGGING_SYMBOLS}; Bitcode: ${ENABLE_BITCODE}; Mapbox.bundle: ${BUNDLE_RESOURCES} bundle.outside: ${PLACE_RESOURCE_BUNDLES_OUTSIDE_FRAMEWORK}"
 
 function step { >&2 echo -e "\033[1m\033[36m* $@\033[0m"; }
 function finish { >&2 echo -en "\033[0m"; }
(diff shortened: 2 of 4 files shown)


boundsj:
@1ec5 @friedbunny I've tested the following scenarios (all with a simple Objc client app linked to the static lib created with the proposed changes):

- [ ] Build and run on simulator
- [ ] Build and run on device with development cert and profile
- [ ] Archive, verify, and upload to iTunes connect with iOS 7 as the Deployment Target  (also distributed with Testflight)
- [ ] Archive, verify, and upload to iTunes connect with iOS 8 as the Deployment Target  (also distributed with Testflight)
- [ ] Archive, verify, and upload to iTunes connect with iOS 9 as the Deployment Target  (also distributed with Testflight)

All of this was done with Xcode 7.3. Before these changes, only first test "Build and run on simulator" passed.

screen shot 2016-03-23 at 9 22 35 pm

platform/ios/INSTALL.md
```diff
@@ -95,8 +95,13 @@ If your application targets iOS 7.x, you’ll need to install the static framewo
 
 1. Build from source manually per above.
 
-1. Open the project editor and select your application target. Drag `build/ios/pkg/static/Mapbox.framework` into the “Embedded Binaries” section of the General tab. (Don’t drag it into the “Linked Frameworks and Libraries” section; Xcode will add it there automatically.) In the sheet that appears, make sure “Copy items if needed” is checked, then click Finish.
-
+1. Copy the contents of `build/ios/pkg/static` into your project. It should happen automatically, but ensure that:
```
1ec5: Developers have indicated that they’d like this step to be as explicit as possible, so let’s say to drag those files into the Project navigator, checking “Copy items if needed”.

platform/ios/INSTALL.md
```diff
@@ -95,8 +95,13 @@ If your application targets iOS 7.x, you’ll need to install the static framewo
...
+   
+   - `Mapbox.framework` is listed in your `Link Binary With Libraries` build phase.
+   - The path to the project's local copy of `Mapbox.framework` is in your *Framework Search Paths* (`FRAMEWORK_SEARCH_PATHS`) build setting.   
+   - `Mapbox.bundle` is in your target's *Copy Bundle Resources* build phase.
+   - `Settings.bundle` is in your target's *Copy Bundle Resources* build phase.
```
1ec5: Settings.bundle is in pkg/, not pkg/static/. It’s also optional.

platform/ios/INSTALL.md
boundsj: Thanks. Will fix.

platform/ios/INSTALL.md
boundsj: Good catch. Will fix and note that it is optional.

boundsj:
@1ec5 I'm not sure why the node / android builds are now failing. I've tried rerunning but no luck so far. Will ignore since I'm certain this change set is unrelated to those builds.

platform/ios/scripts/package.sh
```diff
@@ -207,17 +222,18 @@ SEM_VERSION=$( git describe --tags --match=ios-v*.*.* --abbrev=0 | sed 's/^ios-v
 SHORT_VERSION=${SEM_VERSION%-*}
 if [[ ${BUNDLE_RESOURCES} ]]; then
     cp -pv LICENSE.md "${OUTPUT}/static/${NAME}.framework"
-    cp -rv platform/ios/app/Settings.bundle "${OUTPUT}/static/${NAME}.framework"
+    cp -rv platform/ios/app/Settings.bundle ${STATIC_SETTINGS_DIRECTORY}
```
1ec5: We need to rethink `$BUNDLE_RESOURCES`. I added this variable to support `make ifabric`, which outputs an all-in-one static framework and nothing else. But now `make ipackage` needs to simultaneously output a static framework, a dynamic framework, and some auxiliary resources like Settings.bundle.

Before this change, `make ipackage` would place the LICENSE.md and Settings.bundle directly under pkg/, so that users of the dynamic framework can also access those resources easily. Now they’re hidden inside the static framework.

1ec5:
Remember to update [pod-README.md](https://github.com/mapbox/mapbox-gl-native/blob/boundsj-fix-static-lib/platform/ios/docs/pod-README.md#static-framework), which gets copied into the build output as README.md. This is the primary place for developers to read the installation instructions.

platform/ios/scripts/package.sh
boundsj: Understood and I agree. I'll think about how to make this better. Will check with you with some ideas tomorrow.

1ec5:
> I'm not sure why the node / android builds are now failing.

@jfirebaugh added a bitrise.yml to master in 231e0b5de1b5074a2c954bd6e203cc5f0e6a2f46 and set up Bitrise to look for that configuration file. So that change needs to be cherry-picked into this branch; otherwise, the Node build is guaranteed to fail. Alternatively, we might be able to change the workflow script’s `trigger_map.pattern` to be `master` instead of `*` for now, to prevent that build from running on any other branches.

boundsj:
@1ec5 I'll make a note in the INSTALL doc that libc++.tbd, libsqlite3.tbd, libz.tbd are actually dylib files in older xcode versions

boundsj:
Passed this test case with Xcode 6.4

- [x] Archive, verify, and upload to iTunes connect with iOS 7 as the Deployment Target  (also distributed with Testflight)

platform/ios/INSTALL.md
boundsj: Actually, because of https://github.com/mapbox/mapbox-gl-native/blob/boundsj-fix-static-lib/platform/ios/scripts/package.sh#L39-L43, Settings.bundle is indeed in `pkg/static` (instead of pkg/static/Mapbox.framework like it would be for fabric). I will note that it is optional though.

platform/ios/scripts/package.sh
boundsj: Ok, I think I'm finally with you @1ec5. So we would like something like (focusing on the static side of the fence and the files moving around in this change):

- ifabric
  - pkg/static/Mapbox.framework (holds **everything**)

- ipackage* flavors
  - pkg/
    - Mapbox.bundle
    - Settings.bundle
    - static/
  - pkg/static/
    - Mapbox.framework

I'll adjust accordingly.

1ec5: More like:

- ifabric
  - pkg/static/Mapbox.framework (holds **everything**)

- ipackage* flavors
  - pkg/
    - Settings.bundle
    - static/
    - dynamic/
  - pkg/static/
    - Mapbox.framework
    - Mapbox.bundle (resources in here)
  - pkg/dynamic/
    - Mapbox.framework (resources in here)

boundsj:
@1ec5 took another stab at this. Please let me know if you think this will work or if a slightly larger refactor of package.sh is required. I have tested that ipackage* and ifrabric produce results as outlined above.

platform/ios/INSTALL.md
```diff
@@ -95,7 +95,18 @@ If your application targets iOS 7.x, you’ll need to install the static framewo
 
 1. Build from source manually per above.
 
-1. Open the project editor and select your application target. Drag `build/ios/pkg/static/Mapbox.framework` into the “Embedded Binaries” section of the General tab. (Don’t drag it into the “Linked Frameworks and Libraries” section; Xcode will add it there automatically.) In the sheet that appears, make sure “Copy items if needed” is checked, then click Finish.
+1. Drag the Mapbox.bundle and Mapbox.framework files in `build/ios/pkg/static` into the Project navigator, checking "Copy items if needed". It should happen automatically, but ensure that:
+   
+   - `Mapbox.framework` is listed in your `Link Binary With Libraries` build phase.
+   - The path to the project's local copy of `Mapbox.framework` is in your *Framework Search Paths* (`FRAMEWORK_SEARCH_PATHS`) build setting.
```
1ec5: Is this accurate? `FRAMEWORK_SEARCH_PATHS` is automatlly set to `$(inherited) $(PROJECT_DIR)` in my test project and it seems to be working just fine.

1ec5: I think the wording is just imprecise. How about:

```md
Your *Framework Search Paths* (`FRAMEWORK_SEARCH_PATHS`) build setting includes the directory that contains `Mapbox.framework`. For most projects, the default value of `$(inherited) $(PROJECT_DIR)` should be sufficient.
```

platform/ios/INSTALL.md
```diff
@@ -95,7 +95,18 @@ If your application targets iOS 7.x, you’ll need to install the static framewo
...
 1. Build from source manually per above.
 
-1. Open the project editor and select your application target. Drag `build/ios/pkg/static/Mapbox.framework` into the “Embedded Binaries” section of the General tab. (Don’t drag it into the “Linked Frameworks and Libraries” section; Xcode will add it there automatically.) In the sheet that appears, make sure “Copy items if needed” is checked, then click Finish.
+1. Drag the Mapbox.bundle and Mapbox.framework files in `build/ios/pkg/static` into the Project navigator, checking "Copy items if needed". It should happen automatically, but ensure that:
+   
+   - `Mapbox.framework` is listed in your `Link Binary With Libraries` build phase.
+   - The path to the project's local copy of `Mapbox.framework` is in your *Framework Search Paths* (`FRAMEWORK_SEARCH_PATHS`) build setting.   
+   - `Mapbox.bundle` is in your target's *Copy Bundle Resources* build phase.
+
+<img width="322" alt="screen shot 2016-03-24 at 6 42 14 pm" src="https://cloud.githubusercontent.com/assets/63186/14035855/4744d476-f1f0-11e5-9392-6b19a5658835.png">
```
1ec5: The alt text should be more descriptive.

As for the image itself, I think it’d be better to link to the image on raw.githubusercontent.com, as we do on pod-README.md, rather than cloud.githubusercontent.com. The former breaks when the file is moved within the repository (but you can address that by pinning to a revision), while the latter breaks when GitHub garbage-collects images that are no longer needed for issues and comments.

1ec5:
Can you add a blurb about these fixes to the changelog as well?

boundsj:
@1ec5 I think all issues are addressed. If there is nothing else, I'll merge after a +1. Thanks again!

CHANGELOG.md
```diff
@@ -84,6 +84,7 @@ Known issues:
 - The Improve This Map tool now uses the same zoom level that is currently being shown in the map view. ([#4068](https://github.com/mapbox/mapbox-gl-native/pull/4068))
 - Fixed a formatting issue in the documentation for `MGLCoordinateBoundsIsEmpty()`. ([#3958](https://github.com/mapbox/mapbox-gl-native/pull/3958))
 - The maximum maximum zoom level is now 21. ([#4417](https://github.com/mapbox/mapbox-gl-native/pull/4417)))
+- Fixed issues with configuration and documentation that caused problems when installing the SDK as a static binary ([#4455](https://github.com/mapbox/mapbox-gl-native/pull/4455))
```
1ec5: Let’s point out that this should also fix App Store submission problems, since the installation itself appears to work without a hitch until you go to release the application.

boundsj: Done. Also rebased and added this blurb to the ios specific changelog.

1ec5:
:+1:

1ec5:
This will also fix #4455.

PR #4459, opened 2016-03-24
TelemetryService: Catch IllegalArgumentException exception when unregisterReceiver

vacxe:
If the receiver was already unregistered or was not registered, then call to unregisterReceiver throws IllegalArgumentException. In your case you need to just put special try/catch for this exception and ignore it.


bleege:
@vacxe Thanks for adding this!  We always appreciate PRs!

I've merged this into `master`, but will now cherry pick it into `release-ios-3.2.0-android-4.0.0` so that it makes the next release.

/cc @jfirebaugh

Issue #4429, opened 2016-03-22
NonNull annotation needed in onMapReady MapboxMap parameter


cammace:
Writing this: 

    mapView.getMapAsync(new OnMapReadyCallback() {
                @Override
                public void onMapReady(final MapboxMap mapboxMap) {

gives the warning 

> Not annotated parameter overrides @NonNull parameter

but adding the @NonNull annotation resolves the Android Studio warning. Can we get rid of the `@NonNull` annotation requirement?

@zugaldia @bleege @tobrun

tobrun:
It's actually already there..
It should automatically add that..

```java
package com.mapbox.mapboxsdk.maps;

import android.support.annotation.NonNull;

/**
 * Interface definition for a callback to be invoked when the map is ready to be used.
 * <p>
 * Once an instance of this interface is set on a {@link MapFragment} or {@link MapView} object,
 * the onMapReady(MapboxMap) method is triggered when the map is ready to be used and provides a non-null instance of {@link MapboxMap}.
 * </p>
 */
public interface OnMapReadyCallback {

    /**
     * Called when the map is ready to be used.
     *
     * @param mapboxMap A non-null instance of a MapboxMap associated with the {@link MapFragment}
     *                  or {@link MapView} that defines the callback.
     */
    void onMapReady(@NonNull MapboxMap mapboxMap);
}


```

tobrun:
I have no issue removing it

tobrun:
Talked with @cammace about it, he is going to pick this up :rocket:

bleege:
Rebased and merged into `release-ios-3.2.0-android-4.0.0`.

PR #4460, opened 2016-03-24
[android] #4429 Removed NonNull from onMapReady

cammace:
As @tobrun and I discussed in issue #4429, I've removed the NonNull annotation from onMapReady as it isn't needed.


cammace:
branched master instead of release-ios-3.2.0-android-4.0.0, closing PR

PR #4462, opened 2016-03-24
[core] Update time point in map context on every update

### LINKED ISSUE #4439: OnMapReadyCallback not called
```java
        mapView = new MapView(this, options);
        ViewGroup viewGroup = (ViewGroup) findViewById(R.id.parentLayout);
        viewGroup.addView(mapView);

        mapView.getMapAsync(this);
```

When creating a mapview as shown above, I'm not receiving the map async callback. 
Only when I zoom out a bit I'm receiving the callback.

An example, the maps is shown but callback is not invoked:

device-2016-03-23-113451

Only when you interact (zoom/pan) with the map, the callback is invoked and content is shown:

device-2016-03-23-113538

We need some kind of invalidation or some sort in onStart/onResume to trigger the update?

brunoabinader:
Previous behavior prevented view from notifying when map was fully loaded, because the style sources relies on a future time point for checking if their tiles are updated. This was first caught only in still mode, but also happens in continuous mode for platforms that rely on `View::notifyMapChange` for updating the view.

Fixes #4439.

/cc @tobrun @jfirebaugh

### DIFF
--- a/src/mbgl/map/map_context.cpp
+++ b/src/mbgl/map/map_context.cpp
@@ -164,6 +164,11 @@ void MapContext::update() {
         return;
     }
 
+    // This time point is used to:
+    // - Calculate style property transitions;
+    // - Hint style sources to notify when all its tiles are loaded;
+    frameData.timePoint = Clock::now();
+
     if (style->loaded && updateFlags & Update::Annotations) {
         data.getAnnotationManager()->updateStyle(*style);
         updateFlags |= Update::Classes;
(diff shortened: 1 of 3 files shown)


tobrun:
Nice!

src/mbgl/map/map_context.cpp
```diff
@@ -164,6 +164,11 @@ void MapContext::update() {
         return;
     }
 
+    // This time point is used to:
+    // - Calculate style property transitions;
+    // - Hint style sources to notify when all its tiles are loaded;
```
jfirebaugh: Where does this happen?

brunoabinader: Here: https://github.com/mapbox/mapbox-gl-native/blob/master/src/mbgl/source/source.cpp#L344

brunoabinader: This would be easier to look when we add some utests to check the different notifications from `View` at some point.

Issue #4444, opened 2016-03-23
Out of bounds interaction telemetry events on Android


camilleanne:
Seeing some Android interaction events (`map.click`) come in that have out of bounds values for `lng` -- in api-events we strictly enforce `lat` to between -90 and 90, and `lng` to -180 to 180 so these events are getting filtered out.

e.g.

`lat: 39.130515262056804,
  lng: -187.88018805382615`

`lat: -64.25024633537475,
  lng: 191.59464321880097`

cc @bleege

bleege:
I've been able to reproduce the issue for Longitude.  It seems centered around the International Date Line.  I'll continue looking to see if I can find it elsewhere.

```java
03-24 15:08:28.594 14948-14948/com.mapbox.mapboxsdk.testapp I/MapView: x = 503.0, y = 765.52454; tapLatLng = LatLng [longitude=-170.73755048183057, latitude=20.631945963786507, altitude=0.0]
03-24 15:08:28.624 14948-14948/com.mapbox.mapboxsdk.testapp I/MapView: x = 510.02267, y = 764.4962; tapLatLng = LatLng [longitude=-170.54494119212745, latitude=20.65833841256945, altitude=0.0]
03-24 15:08:28.634 14948-14948/com.mapbox.mapboxsdk.testapp I/MapView: x = 513.0, y = 764.0; tapLatLng = LatLng [longitude=-170.46328199459674, latitude=20.67107244593123, altitude=0.0]
03-24 15:08:28.634 14948-14948/com.mapbox.mapboxsdk.testapp I/MapView: x = 513.0, y = 764.0; tapLatLng = LatLng [longitude=-170.46328199459674, latitude=20.67107244593123, altitude=0.0]
03-24 15:08:34.514 14948-14948/com.mapbox.mapboxsdk.testapp I/MapView: x = 531.0, y = 794.0; tapLatLng = LatLng [longitude=-175.15328030718524, latitude=20.438555684695217, altitude=0.0]
03-24 15:08:39.494 14948-14948/com.mapbox.mapboxsdk.testapp I/MapView: x = 376.0, y = 844.0; tapLatLng = LatLng [longitude=-180.48651391197888, latitude=19.76890837502343, altitude=0.0]
[...]

PR #4464, opened 2016-03-24
Wrap coordinates converted from points

1ec5:
The release-ios-3.2.0-android-4.0.0 branch includes #4275 but not #4285, meaning that `-[MGLMapView convertPoint:toCoordinateFromView:]` and dependent methods return latitudes that aren’t wrapped to ±180°. This change restores the wrapping on an interim basis.

@jfirebaugh has made the point that this method shouldn’t wrap anyways, but we don’t want to bump the major version number so close to a release. Although the wrapping isn’t documented anywhere, it’s consistent with MapKit, and it’s very likely that developers have come to rely on this behavior (as has iosapp). We can revisit unwrapping in a future release.

If a developer needs an unwrapped coordinate, they can detect this edge case by comparing the return values of `-[MGLMapView convertPoint:toCoordinateFromView:]` for two different points on screen (and accounting for rotation and pitch of course). In fact, this is [what `-convertRect:toRectToView:` does](https://github.com/mapbox/mapbox-gl-native/blob/fc76689a5b379f3d340652a65422c36a01ab35bb/platform/ios/src/MGLMapView.mm#L2196-L2212) (although that may be dead code due to 63820d1b4f96b4e2370fd6567a8f77149117620e for #3401).

/ref #4444
/cc @brunoabinader @bleege @boundsj

### DIFF
--- a/platform/ios/src/MGLMapView.mm
+++ b/platform/ios/src/MGLMapView.mm
@@ -2150,7 +2150,7 @@ mbgl::Duration MGLDurationInSeconds(NSTimeInterval duration)
 - (mbgl::LatLng)convertPoint:(CGPoint)point toLatLngFromView:(nullable UIView *)view
 {
     CGPoint convertedPoint = [self convertPoint:point fromView:view];
-    return _mbglMap->latLngForPixel(mbgl::ScreenCoordinate(convertedPoint.x, convertedPoint.y));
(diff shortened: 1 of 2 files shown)


boundsj:
:+1:

1ec5:
The Android build failure is #4473. Merging.

brunoabinader:
Looks good, @1ec5!

PR #4472, opened 2016-03-25
[Android] SupportMapFragment: Check KEY_META_DATA

vacxe:
If I try add SupportMapFragment in XML (i know it not good way, but any people can try do this) that than has NPE, because Bundle is empty. I think it simple solution to fix set access token. After, user can set style and other parameters when he get map async.

### DIFF
--- a/platform/android/MapboxGLAndroidSDK/src/main/java/com/mapbox/mapboxsdk/maps/SupportMapFragment.java
+++ b/platform/android/MapboxGLAndroidSDK/src/main/java/com/mapbox/mapboxsdk/maps/SupportMapFragment.java
@@ -1,9 +1,12 @@
 package com.mapbox.mapboxsdk.maps;
 
+import android.content.pm.ApplicationInfo;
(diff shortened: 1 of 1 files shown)


tobrun:
@vacxe will review soon

tobrun:
Going to pick this up, it seems valid suggestion to fallback to `meta-data` key even if we have removed the `APIAccess` class in previous release. This should also open up the possibility of having a `newInstance` method a parameter.

tobrun:
I'm going to pull this in our repo so our CI tests can be run, this will also require me to open up another PR. You can follow along there.

tobrun:
@vacxe separate PR created in #4610, I added some extra checks, ported code to MapFragment and allowed the possibility to use the `mapbox_access_token` string resource.

vacxe:
@tobrun Thanks, nice improvement.

tobrun:
@vacxe I just merged in #4610. Thank you for the contribution and great job on this!

PR #4477, opened 2016-03-25
Wrapping Telemetry Longitude Data

### LINKED ISSUE #4475: Introduce LatLng.wrap()
As part of #4444 wrapping of LatLng at the JNI level was introduced to help solve issue with the International Date Line / antimeridian.  After some more internal discussion with @tobrun @zugaldia and based on [feedback from @1ec5 ](https://github.com/mapbox/mapbox-gl-native/issues/4444#issuecomment-201053694) and @jfirebaugh  it makes the most sense to introduce a `LatLng.wrap()` method instead of having to remember to call `wrap()` on every JNI method.  This keeps the Java API and the Core GL API in sync.  It also won't negatively impact apps migrating from Google Maps to Mapbox as the Google Map LatLng object is already wrapped.

/cc @mapbox/mobile

bleege:
Closes #4475

### DIFF
--- a/platform/android/MapboxGLAndroidSDK/src/main/java/com/mapbox/mapboxsdk/geometry/LatLng.java
+++ b/platform/android/MapboxGLAndroidSDK/src/main/java/com/mapbox/mapboxsdk/geometry/LatLng.java
@@ -6,6 +6,7 @@ import android.os.Parcelable;
 
 import com.mapbox.mapboxsdk.constants.GeoConstants;
 import com.mapbox.mapboxsdk.constants.MathConstants;
+import com.mapbox.mapboxsdk.utils.MathUtils;
 
 import java.io.Serializable;
 
@@ -116,6 +117,17 @@ public class LatLng implements ILatLng, Parcelable {
     @Override
     public double getAltitude() {
         return altitude;
+    }
+
+    /**
+     * Return a new LatLng object with a wrapped Longitude.  This allows original data object
+     * to remain unchanged.
+     * @return New LatLng object with wrapped Longitude
+     */
+    public LatLng wrap(){
+        LatLng wrappedVersion = new LatLng(this);
+        wrappedVersion.setLongitude(MathUtils.wrap(wrappedVersion.getLongitude(), GeoConstants.MIN_LONGITUDE, GeoConstants.MAX_LONGITUDE));
+        return wrappedVersion;
     }
 
     @Override
--- a/platform/android/MapboxGLAndroidSDK/src/main/java/com/mapbox/mapboxsdk/geometry/LatLng.java
+++ b/platform/android/MapboxGLAndroidSDK/src/main/java/com/mapbox/mapboxsdk/geometry/LatLng.java
@@ -126,7 +126,10 @@ public class LatLng implements ILatLng, Parcelable {
      */
     public LatLng wrap(){
         LatLng wrappedVersion = new LatLng(this);
-        wrappedVersion.setLongitude(MathUtils.wrap(wrappedVersion.getLongitude(), GeoConstants.MIN_LONGITUDE, GeoConstants.MAX_LONGITUDE));
+        double lon = wrappedVersion.getLongitude();
+        if (lon < GeoConstants.MIN_LONGITUDE || lon > GeoConstants.MAX_LONGITUDE) {
+            wrappedVersion.setLongitude(MathUtils.wrap(wrappedVersion.getLongitude(), GeoConstants.MIN_LONGITUDE, GeoConstants.MAX_LONGITUDE));
+        }
         return wrappedVersion;
     }
 
--- a/platform/android/MapboxGLAndroidSDK/src/main/java/com/mapbox/mapboxsdk/telemetry/MapboxEventManager.java
+++ b/platform/android/MapboxGLAndroidSDK/src/main/java/com/mapbox/mapboxsdk/telemetry/MapboxEventManager.java
@@ -558,8 +558,10 @@ public class MapboxEventManager {
                     if (evt.containsKey(MapboxEvent.KEY_LONGITUDE)) {
                         double lon = (double)evt.get(MapboxEvent.KEY_LONGITUDE);
                         Log.i(TAG, "Original Lon = " + lon);
-                        lon = MathUtils.wrap(lon, GeoConstants.MIN_LONGITUDE, GeoConstants.MAX_LONGITUDE);
-                        Log.i(TAG, "Wrapped Lon = " + lon);
+                        if ((lon < GeoConstants.MIN_LONGITUDE) || (lon > GeoConstants.MAX_LONGITUDE)) {
+                            lon = MathUtils.wrap(lon, GeoConstants.MIN_LONGITUDE, GeoConstants.MAX_LONGITUDE);
+                            Log.i(TAG, "Wrapped Lon = " + lon);
+                        }
                         jsonObject.put(MapboxEvent.KEY_LONGITUDE, lon);
                     }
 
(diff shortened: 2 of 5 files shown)


platform/android/MapboxGLAndroidSDKTestApp/src/test/java/com/mapbox/mapboxsdk/geometry/LatLngTest.java
```diff
@@ -2,19 +2,17 @@
...
 import android.os.Parcel;
-
 import com.mapbox.mapboxsdk.geometry.LatLng;
 import com.mapbox.mapboxsdk.utils.MockParcel;
-
 import org.junit.Test;
-
 import java.util.Objects;
-
+import dalvik.annotation.TestTargetClass;
```
zugaldia: The addition of this import seems unnecessary.

platform/android/MapboxGLAndroidSDKTestApp/src/test/java/com/mapbox/mapboxsdk/geometry/LatLngTest.java
```diff
@@ -202,4 +200,11 @@ public void testDescribeContents() {
         assertEquals("contents should be 0", 0, latLng.describeContents(), DELTA);
     }
 
+    @Test
+    public void testWrapped() {
+        LatLng latLng = new LatLng(45.0, -185.0);
+        LatLng wrapped = latLng.wrap();
+        assertTrue(!latLng.equals(wrapped));
+        assertTrue(wrapped.getLongitude() == 175.0);
```
zugaldia: For consistency, this should be `assertEquals("longitude wrapped value", wrapped.getLongitude(), 175.0, DELTA);`

platform/android/MapboxGLAndroidSDK/src/main/java/com/mapbox/mapboxsdk/utils/MathUtils.java
```diff
@@ -24,4 +24,23 @@ public static float clamp(float value, float min, float max) {
         return Math.max(min, Math.min(max, value));
     }
 
+    /**
+     * Constrains n to the given range (including min, excluding max) via modular arithmetic.
```
zugaldia: "Constrains value"

platform/android/MapboxGLAndroidSDK/src/main/java/com/mapbox/mapboxsdk/geometry/LatLng.java
```diff
@@ -118,6 +119,17 @@ public double getAltitude() {
...
     }
 
+    /**
+     * Return a new LatLng object with a wrapped Longitude.  This allows original data object
+     * to remain unchanged.
+     * @return New LatLng object with wrapped Longitude
+     */
+    public LatLng wrap(){
+        LatLng wrappedVersion = new LatLng(this);
+        wrappedVersion.setLongitude(MathUtils.wrap(wrappedVersion.getLongitude(), GeoConstants.MIN_LONGITUDE, GeoConstants.MAX_LONGITUDE));
```
zugaldia: Should we include a check similar to the one in telemetry (`if ((lon < GeoConstants.MIN_LONGITUDE) || (lon > GeoConstants.MAX_LONGITUDE)) {`) to avoid creating a new object if it's unnecessary?

bleege: I think adding the range check to avoid wrapping if we don't have to makes sense.  As for the new object part, I kinda like the idea that the `wrap()` method is consistently giving a new object so that there's never any doubt that the base object data still remains.

zugaldia: :+1:

bleege:
Just made the changes recommended by @zugaldia.  When CI approves I'll merge this into `release-ios-3.2.0-android-4.0.0`.

bleege:
CI is happy!  Time to merge!

bleege:
Rebased and merged!

platform/android/MapboxGLAndroidSDK/src/main/java/com/mapbox/mapboxsdk/telemetry/MapboxEventManager.java
```diff
@@ -551,7 +553,16 @@ protected Void doInBackground(Void... voids) {
                     jsonObject.putOpt(MapboxEvent.ATTRIBUTE_SOURCE, evt.get(MapboxEvent.ATTRIBUTE_SOURCE));
                     jsonObject.putOpt(MapboxEvent.ATTRIBUTE_SESSION_ID, evt.get(MapboxEvent.ATTRIBUTE_SESSION_ID));
                     jsonObject.putOpt(MapboxEvent.KEY_LATITUDE, evt.get(MapboxEvent.KEY_LATITUDE));
-                    jsonObject.putOpt(MapboxEvent.KEY_LONGITUDE, evt.get(MapboxEvent.KEY_LONGITUDE));
```
jfirebaugh: This should be wrapped using `LatLng.wrap` [here](https://github.com/mapbox/mapbox-gl-native/blob/0d816916b108bfe830bddf1c01ebe9e12caa88ba/platform/android/MapboxGLAndroidSDK/src/main/java/com/mapbox/mapboxsdk/maps/MapView.java#L1362-L1362) rather than duplicating the conditionals for wrapping.

jfirebaugh:
Did you audit the other places `MapView.fromScreenLocation` is used to determine whether or not they need wrapping?

platform/android/MapboxGLAndroidSDK/src/main/java/com/mapbox/mapboxsdk/telemetry/MapboxEventManager.java
bleege: Normally I'd agree that conditional wrapping should be avoided, however in this case it's needed.  Reason being is that Latitude and Longitude data can also come from other events besides `location`.  Specifically gesture events like [`map.click`](https://github.com/mapbox/mapbox-gl-native/blob/b864fd4ea543b85daf48380fc36388f8cedf0032/platform/android/MapboxGLAndroidSDK/src/main/java/com/mapbox/mapboxsdk/maps/MapView.java#L1391) and [`map.dragend`](https://github.com/mapbox/mapbox-gl-native/blob/b864fd4ea543b85daf48380fc36388f8cedf0032/platform/android/MapboxGLAndroidSDK/src/main/java/com/mapbox/mapboxsdk/maps/MapView.java#L1412).  This is handled automatically via `json.putOpt()` for Latitude because it's value (if one exists) can always be used as is.

jfirebaugh: @bleege My point is that we should reuse `LatLng.wrap`, which already has all the correct logic, whenever we add coordinates to the event queue. There's no need to duplicate the code.

bleege:
The primary goal was to minimally impact the code base this late in the release process, so this was deliberately scoped to the Telemetry issue from #4444.  We can and should do a further investigation for the next release.

platform/android/MapboxGLAndroidSDK/src/main/java/com/mapbox/mapboxsdk/telemetry/MapboxEventManager.java
bleege: @jfirebaugh Ah... my mistake.  I saw the comment attached to Line 554 and thought you were referring to the `jsonObject.putOpt()` for Longitude and didn't follow that you were referring to wrap tests itself.  I follow your thinking now about creating a new `LatLng` with a constant Latitude and the specific Longitude to be wrapped.

PR #4490, opened 2016-03-28
[ios] Update the pod-README and changelog regarding static lib changes

boundsj:
Update the pod-README so generated documentation is correct and make a note in the changelog
regarding the updated / reverted procedure for installing the SDK as a static framework.

cc @1ec5 @friedbunny

### DIFF
--- a/platform/ios/docs/pod-README.md
+++ b/platform/ios/docs/pod-README.md
@@ -33,7 +33,17 @@ bash "${BUILT_PRODUCTS_DIR}/${FRAMEWORKS_FOLDER_PATH}/Mapbox.framework/strip-fra
 
 If your application targets iOS 7.x, you’ll need to install the static framework instead:
 
-1. Open the project editor and select your application target. Drag `Mapbox.framework` from the `static` folder into the “Embedded Binaries” section of the General tab. (Don’t drag it into the “Linked Frameworks and Libraries” section; Xcode will add it there automatically.) In the sheet that appears, make sure “Copy items if needed” is checked, then click Finish.
+1. Build from source manually per above.
+
+1. Drag the Mapbox.bundle and Mapbox.framework files in `build/ios/pkg/static` into the Project navigator, checking "Copy items if needed". It should happen automatically, but ensure that:
+   
+   - `Mapbox.framework` is listed in your `Link Binary With Libraries` build phase.
+   - Your *Framework Search Paths* (`FRAMEWORK_SEARCH_PATHS`) build setting includes the directory that contains `Mapbox.framework`. For most projects, the default value of `$(inherited) $(PROJECT_DIR)` should be sufficient.  
+   - `Mapbox.bundle` is in your target's *Copy Bundle Resources* build phase.
+
+1. **Optional** As [noted in the documentation](https://www.mapbox.com/ios-sdk/#telemetry_opt_out) you may use the provided Settings.bundle to provide a Telemetry opt out for users. To use the provided file, drag the `Settings.bundle` file in `build/ios/pkg` into the Project navigator, checking "Copy items if needed". It should happen automatically, but ensure that:
+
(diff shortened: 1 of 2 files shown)


1ec5:
Took the liberty of moving the entry up to the top, since it’s a major change for static framework users.

platform/ios/docs/pod-README.md
```diff
@@ -33,7 +33,17 @@ bash "${BUILT_PRODUCTS_DIR}/${FRAMEWORKS_FOLDER_PATH}/Mapbox.framework/strip-fra
 
 If your application targets iOS 7.x, you’ll need to install the static framework instead:
 
-1. Open the project editor and select your application target. Drag `Mapbox.framework` from the `static` folder into the “Embedded Binaries” section of the General tab. (Don’t drag it into the “Linked Frameworks and Libraries” section; Xcode will add it there automatically.) In the sheet that appears, make sure “Copy items if needed” is checked, then click Finish.
+1. Build from source manually per above.
```
friedbunny: Which step “above” are we referring to? Unless we’re going to stop distributing a zipped static build (which would be fine by me), these instructions should point the user to the GitHub releases page and the prebuilt static build, not tell them how to build from source.

1ec5: This document is only distributed with the prebuilt static build, so this bullet point should go away.

platform/ios/docs/pod-README.md
```diff
@@ -33,7 +33,17 @@ bash "${BUILT_PRODUCTS_DIR}/${FRAMEWORKS_FOLDER_PATH}/Mapbox.framework/strip-fra
...
+
+1. Drag the Mapbox.bundle and Mapbox.framework files in `build/ios/pkg/static` into the Project navigator, checking "Copy items if needed". It should happen automatically, but ensure that:
```
1ec5: The path should be `static`, not `build/ios/pkg/static`.

platform/ios/docs/pod-README.md
```diff
@@ -33,7 +33,17 @@ bash "${BUILT_PRODUCTS_DIR}/${FRAMEWORKS_FOLDER_PATH}/Mapbox.framework/strip-fra
...
+   
+   - `Mapbox.framework` is listed in your `Link Binary With Libraries` build phase.
+   - Your *Framework Search Paths* (`FRAMEWORK_SEARCH_PATHS`) build setting includes the directory that contains `Mapbox.framework`. For most projects, the default value of `$(inherited) $(PROJECT_DIR)` should be sufficient.  
+   - `Mapbox.bundle` is in your target's *Copy Bundle Resources* build phase.
+
+1. **Optional** As [noted in the documentation](https://www.mapbox.com/ios-sdk/#telemetry_opt_out) you may use the provided Settings.bundle to provide a Telemetry opt out for users. To use the provided file, drag the `Settings.bundle` file in `build/ios/pkg` into the Project navigator, checking "Copy items if needed". It should happen automatically, but ensure that:
```
1ec5: Settings.bundle is in the ZIP file’s root folder, not under `build/ios/pkg`.

platform/ios/docs/pod-README.md
boundsj: @friedbunny @1ec5 

Sorry, this was copy pasta from the INSTALL.md file. I will remove this line. I see that there is language referencing the releases page above in the `## Installation` section. Is that good enough or do we need to be more specific here and in the DYNAMIC template just above?

1ec5: That line above is good enough.

boundsj:
Thanks @friedbunny @1ec5. I've updated.

1ec5:
:+1:

boundsj:
I'm going to merge. The Travis failure is only on `tidy-clang38-release` and unrelated to this change.

PR #4493, opened 2016-03-28
Exclude offline cache from backups

1ec5:
Avoid backing up the offline cache onto iCloud, because it can be redownloaded and would easily take up a significant portion of a standard iCloud account. Ideally, we’d even put ambiently cached resources in Caches, so it can be reclaimed by the system when disk space runs low. But unfortunately it has to live in the same file as offline resources.

/cc @friedbunny


friedbunny:
:+1:

1ec5:
The iOS test failure doesn’t reproduce locally or in a manually-run build on Bitrise, so I’m satisfied that it wasn’t introduced in this PR.

PR #4496, opened 2016-03-28
[android] Remove onStart() and onStop() lifecycle methods from MapView

justasm:
- Removes `onStart()` and `onStop()` from MapView and UserLocationView.
- Moves sensor event listener registering / unregistering from UserLocationView `onStart()` and `onStop()` to `onResume()` and `onPause()`, respectively.

Motivation:
- Parity with the [Google Maps Android API v2 regarding MapView lifecycle requirements](https://developers.google.com/android/reference/com/google/android/gms/maps/MapView).
- Simpler API for users who choose to manage MapView directly.
- Android guidelines [recommend registering and unregistering sensor event listeners in `onResume()` and `onPause()`](http://developer.android.com/guide/topics/sensors/sensors_overview.html#sensors-monitor) to reduce battery drain in situations where the Activity is paused but not stopped, e.g. when the user locks the device and the screen is off.

### DIFF
--- a/platform/android/MapboxGLAndroidSDK/src/main/java/com/mapbox/mapboxsdk/maps/MapFragment.java
+++ b/platform/android/MapboxGLAndroidSDK/src/main/java/com/mapbox/mapboxsdk/maps/MapFragment.java
@@ -76,7 +76,6 @@ public final class MapFragment extends Fragment {
     @Override
     public void onStart() {
         super.onStart();
-        mMap.onStart();
         mMap.getMapAsync(mOnMapReadyCallback);
     }
 
@@ -115,7 +114,6 @@ public final class MapFragment extends Fragment {
     @Override
     public void onStop() {
         super.onStop();
-        mMap.onStop();
     }
 
     /**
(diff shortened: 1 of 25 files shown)


tobrun:
to be able to run our CI tests and put it into the release branch versus master. I created #4510 and followed up with #4511 PR. Going to close this one in favour of those mentioned above.

bleege:
@justasm Thank you for this contribution!  We all really like how it simplifies things and continues to move the SDK closer to Google Map compatibility.  We need to do some internal discussion around timing first before proceeding however.  We should have more info on this soon.  Thanks again!

/cc @zugaldia @jfirebaugh

bleege:
@justasm Thanks for your patience.  We've all discussed this internally and decided that we want to get it into the `4.0.0` release.  Since we're using the `release-ios-3.2.0-android-4.0.0` branch instead of `master` to generate the builds / run test infrastructure I'm going to close this PR down again and use #4511.   @tobrun has cherrypicked your commit to it so you're contribution will definitely be noted.

Thanks again for making this contribution!  Please keep up the great work!

PR #4499, opened 2016-03-28
fix background-pattern rendering in perspective view

### LINKED ISSUE #3488: switch u_patternmatrix to u_patternscale in pattern shader

### LINKED ISSUE #3801: port background rendering from -js

ansis:
fix #3801
fix #3488
:eyes: @jfirebaugh

### DIFF
--- a/package.json
+++ b/package.json
@@ -26,7 +26,7 @@
   ],
   "devDependencies": {
     "aws-sdk": "^2.2.21",
-    "mapbox-gl-test-suite": "mapbox/mapbox-gl-test-suite#8859b504ef241bca49d4c2f9a79ff4d0bfdf81a1",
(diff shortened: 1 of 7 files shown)


jfirebaugh:
Does this fix #3488 too?

ansis:
Yeah, this fixes #3488

jfirebaugh:
:+1:

mb12:
@ansis Is plainshader needed for cases when the background is not a pattern? glClearColor is already set to background color.

ansis:
@mb12 there is an uncommon but possible case where there are multiple background layers

PR #4504, opened 2016-03-29
[ios] Don't use negative content insets

### LINKED ISSUE #4440: Doesn't respect height of view
In `3.2.0 beta 2` on iOS, the map view doesn't seem to respect the height of the view, meaning it doesn't show the full world map and the centre is off. So when zoomed in and you set the centre coordinate, it will not be visible.

The following screens illustrate the problem.

img_2117
3.0.1

img_2116
3.2.0 beta 2

The actual `MGLMapView` is of correct size in both instances according to `frame.size.height`.

friedbunny:
When a map view was smaller than the entire viewport, negative content insets would be applied. This had the effect of making the visual "center" of every map the center of the viewport, ignoring the map view’s true center. Negative content insets would have only been valid if the map view extended outside of its frame, which cannot happen.

Applying negative content insets would erroneously offset `showAnnotations:`, `setCenterCoordinate:`, “follow” user tracking mode, and so on.

Fixes #4440.

9e7aad7d00ef09cf74f6d08f4d1e4189d613ac5a also slightly tweaks annotation padding behavior so that smaller views use proportional padding.

/cc @1ec5


platform/ios/src/MGLMapView.mm
```diff
@@ -860,6 +861,13 @@ - (CGPoint)contentCenter
     return CGPointMake(CGRectGetMidX(contentFrame), CGRectGetMidY(contentFrame));
 }
 
+/// Returns edge insets with negative values replaced with 0.
+- (UIEdgeInsets)removeNegativeInsets:(UIEdgeInsets)insets
+{
+    return UIEdgeInsetsMake((insets.top > 0 ? insets.top : 0), (insets.left > 0 ? insets.left : 0),
```
1ec5: This should be expressed with the MIN macro for brevity. And since it's only used in one place, it should be inclined.

1ec5: Of course I mean the MAX macro. Definitely keep the comment. 😄

friedbunny: I will incline the MAXes. :wink:

friedbunny: Force pushed these changes at 0b3c67d.

friedbunny:
The iOS build on Bitrise [is failing](https://www.bitrise.io/build/c00b583c60ccbfde) with `xcode test exit code: 65` (and nothing else). :disappointed:

friedbunny:
Seeing tests fail intermittently on my machine, too:

```
Failing tests:
	-[MapViewTests testContentInsetsWithTinyMapView]
** TEST FAILED **

make: *** [itest] Error 65
```

friedbunny:
Tests pass reliably (five in a row on Bitrise) after switching to KVO in 9e91e6c886dc15257093d032df3f6c83e947df84, which I rebased back into https://github.com/mapbox/mapbox-gl-native/pull/4504/commits/183f55eb5e9cad3f342f3bcc8ba8fc18e38243fe.

platform/ios/test/MapViewTests.m
```diff
@@ -424,6 +424,29 @@ - (void)testInsetMapView {
...
+    [self keyValueObservingExpectationForObject:tester.mapView keyPath:@"contentInset" handler:^BOOL(id observedObject, NSDictionary *change) {
+        XCTAssertEqual(tester.mapView.contentInset.top,
+                       0,
+                       @"map should not have top content inset");
+        XCTAssertEqual(tester.mapView.contentInset.bottom,
+                       0,
+                       @"map should not have bottom content inset");
+        return YES;
+    }];
+    [self waitForExpectationsWithTimeout:5.0 handler:nil];
```
1ec5: Five seconds seems rather generous. Is it necessary to make this test take 10 seconds for a frame change that should be instant?

friedbunny: I’ll lower this to `1.0`.

This is mainly a buffer against slow CI machines. `waitForExpectationsWithTimeout:handler:` should never add more than a tick or two to the execution time, no matter the timeout length — if exceeds the timeout, it’s a failure and everything stops.

I think xcpretty is swallowing our test output and the elapsed time per test on Bitrise, but locally this test completes in the blink of an eye.

friedbunny: → https://github.com/mapbox/mapbox-gl-native/commit/fde8a4c59de08559b9125297f80589a0c188b516#diff-64665744fc510b27e594fbb086ae26d5R438

1ec5:
Merged as d20795b891a6dfbde55333c95234c0a6d4e82131 and 593f3d511a89acb226979f4a504d2449d9a96969.

friedbunny:
This PR got jumped by another into `release-ios-3.2.0-android-4.0.0` and neglected to force-push again into this branch. Merged per above.

PR #4506, opened 2016-03-29
enable circle-opacity render tests

### LINKED ISSUE #2647: circle rendering differences
This is visible in all the circle rendering tests, but only actually fails the opacity ones. It looks like native is rendering a few circles that js isn't. Filing here, but need to verify which of the two is correct.

image

ansis:
fix #2647

I think the tests changed with https://github.com/mapbox/mapbox-gl-js/commit/7658e00da603bc9721ff57f258239997888adf5f in -js.

### DIFF
--- a/package.json
+++ b/package.json
@@ -26,7 +26,7 @@
   ],
   "devDependencies": {
     "aws-sdk": "^2.2.21",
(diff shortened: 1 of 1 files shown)


ansis:
replaced by https://github.com/mapbox/mapbox-gl-native/pull/4507

PR #4507, opened 2016-03-29
[core] fix circle rendering in still mode 

### LINKED ISSUE #2647: circle rendering differences

### LINKED ISSUE #3764: fix circle rendering in still mode

ansis:
fix #3764

Also switch circle rendering tests to the poi_label source layer. Still mode rendering requires that points are in the same spots in all tiles. This was not the case with the road layer.

fix #2647 - no changes were needed. I think some -js change already fixed these tests.

### DIFF
--- a/package.json
+++ b/package.json
@@ -26,7 +26,7 @@
   ],
   "devDependencies": {
     "aws-sdk": "^2.2.21",
-    "mapbox-gl-test-suite": "mapbox/mapbox-gl-test-suite#10924ef1cea1bea0cb1fa0aac4233a037a4daecc",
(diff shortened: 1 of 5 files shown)


ansis:
:eyes: @jfirebaugh

jfirebaugh:
:+1:

PR #4512, opened 2016-03-29
[android] #4500 - OnMapReady not invoked for offline maps

### LINKED ISSUE #4500: onMapReady not getting called when offline when a (largish) offline region is present
Following up on #4486: Turns out the issue that I encountered only seems to occur when a offline region larger than a particular size has been downloaded. In my test case this occurs with a region of 87.353 megabytes but **not** with a region of 74.243 megabytes.

Testing steps are mostly the same as I outlined in #4486. The difference being that a OfflineRegion of ~87MB (or probably larger) is downloaded the first time application is loaded with the WiFi enabled.

@zugaldia, @jfirebaugh does that make sense?

I am putting together a test Activity in the TestApp to demonstrate.

tobrun:
closes #4500.

cc @jfirebaugh @zugaldia @bleege

### DIFF
--- a/platform/android/MapboxGLAndroidSDK/src/main/java/com/mapbox/mapboxsdk/maps/MapView.java
+++ b/platform/android/MapboxGLAndroidSDK/src/main/java/com/mapbox/mapboxsdk/maps/MapView.java
@@ -400,7 +400,7 @@ public class MapView extends FrameLayout {
         addOnMapChangedListener(new OnMapChangedListener() {
             @Override
             public void onMapChanged(@MapChange int change) {
-                if (change == DID_FINISH_LOADING_MAP) {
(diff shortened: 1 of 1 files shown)


jfirebaugh:
I believe this change is safe and will have the intended effect. I want to try to understand why `DID_FINISH_LOADING_MAP` isn't working as expected though. That may indicate an issue that affects iOS and node as well as Android.

bleege:
Picking this up now.

bleege:
I just successfully tested by:

* Opening TestApp
* Opening Offline Map Activity
* Downloading Region
* Quiting Test App
* Turning WiFi Off (no cellular connection)
* Opening TestApp
* Opening Offline Map Activity
* Load Saved Region From `List Regions` Button

Will merge this into `release-ios-3.2.0-android-4.0.0` now.

bleege:
Rebased and Merged!

PR #4519, opened 2016-03-29
[core] fix background-pattern rendering in perspective view

jfirebaugh:
Cherry-pick of #4499 to the release branch.


1ec5:
Is there anything worth mentioning in the changelogs?

PR #4524, opened 2016-03-29
[core] Localize and simplify state for MapChangeDidFinishLoadingMap event

### LINKED ISSUE #4500: onMapReady not getting called when offline when a (largish) offline region is present

jfirebaugh:
Fixes #4500 for iOS.

Only `Map` needs to track the loaded state, and it can simply be set when beginning to load a style and reset when finished.

The only commit related to `MapData::loading` that I could find was 2ee7b0218929d56c9331ded5ef695fc6bc59f70b. I thought I remembered some additional adjustments to how this variable was used based on past bugs, but could not find any. There is one branch that references it, [`tmpsantos-MapChangeDidFinishLoadingMap`](https://github.com/mapbox/mapbox-gl-native/tree/tmpsantos-MapChangeDidFinishLoadingMap), which looks to be obsolete. @tmpsantos do you remember what this was about? Is anyone else aware of subtleties relating to `MapData::loading` or `MapChangeDidFinishLoadingMap` that I forgot to take into account?

This provides an alternative fix for #4512 on Android, but the least risky approach would be to continue using that workaround for the release and perhaps revert it on the main branch.

I attempted to write a test for `MapChangeDidFinishLoadingMap`, but it is very difficult. It will be easier once #2909 lands.

:eyes: @1ec5 @brunoabinader @tmpsantos

### DIFF
--- a/include/mbgl/map/map.hpp
+++ b/include/mbgl/map/map.hpp
@@ -205,6 +205,7 @@ private:
     };
 
     RenderState renderState = RenderState::never;
+    bool loading = false;
 };
 
 } // namespace mbgl
--- a/src/mbgl/map/map.cpp
+++ b/src/mbgl/map/map.cpp
@@ -78,8 +78,8 @@ void Map::renderSync() {
     } else if (renderState != RenderState::fully) {
         renderState = RenderState::fully;
         view.notifyMapChange(MapChangeDidFinishRenderingMapFullyRendered);
-        if (data->loading) {
-            data->loading = false;
+        if (loading) {
+            loading = false;
             view.notifyMapChange(MapChangeDidFinishLoadingMap);
         }
     }
@@ -101,11 +101,13 @@ void Map::update(Update flags) {
 #pragma mark - Style
 
 void Map::setStyleURL(const std::string &url) {
+    loading = true;
     view.notifyMapChange(MapChangeWillStartLoadingMap);
     context->invoke(&MapContext::setStyleURL, url);
 }
 
 void Map::setStyleJSON(const std::string& json, const std::string& base) {
+    loading = true;
     view.notifyMapChange(MapChangeWillStartLoadingMap);
     context->invoke(&MapContext::setStyleJSON, json, base);
 }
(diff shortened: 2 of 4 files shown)


jfirebaugh:
I should have included my reasoning for why this is thread safe. The relevant events are:

| Main thread | Map thread |
| ----- | ----- |
| `map.setStyleURL`/`setStyleJSON` |    |
|    |  invocation of `context.setStyleURL` / `setStyleJSON` |
| `map.renderSync` | sync invocation of `context.renderSync` |

Invocations, including synchronous ones, are FIFO, so the `fullyLoaded` result of the sync invocation of `context.renderSync` is guaranteed to reflect the loaded status of the most recently set style. The following execution schedule is not possible, for example:

| Main thread | Map thread |
| ----- | ----- |
| `map.setStyleURL(A)` |    |
| `map.setStyleURL(B)` |    |
|    | invocation of `context.setStyleURL(A)` |
| `map.renderSync` | sync invocation of `context.renderSync` |
| `MapChangeDidFinishLoadingMap` emitted for A, `loading` is reset to false |   |
|    |  invocation of `context.setStyleURL(B)` |
| `map.renderSync` |   |
| `MapChangeDidFinishLoadingMap` should be emitted for B, but isn't |   |

tmpsantos:
> @tmpsantos do you remember what this was about?

That branch is obsolete. Just deleted it.

> Is anyone else aware of subtleties relating to MapData::loading or MapChangeDidFinishLoadingMap that I forgot to take into account?

We have several methods that are currently crashing because `setStyleURL` is async. For instance `addLayer`, `addClass`, `isLoaded` and several others. We should have a signal (`MapChangeDidFinishLoadingStyle`?) emitted right after `loadStyleJSON`.

On Android for instance `MapChangeDidFinishLoadingMap` is used to indicate that the style was loaded as a whole but we need something to indicate that the style JSON arrived (and the style object was created). Otherwise we might ended up rendering a partial state and later add a custom layer or class.

tmpsantos:
I started implementing this new signal and realized that it would be better implemented after we get rid of the context thread because we can safely call `view.notifyMapChange` from `MapContext::setStyleURL` after that.

tmpsantos:
:+1:  for this one, created #4527 for my other comments. Sorry for the noise.

1ec5:
:+1:

brunoabinader:
Excellent! :+1:

PR #4528, opened 2016-03-30
increase geojson buffer size, parse overscaled geojson

### LINKED ISSUE #3838: fix geojson/reparse-overscaled render test
screen shot 2016-02-05 at 4 09 21 pm

-js fix: https://github.com/mapbox/mapbox-gl-js/pull/2029

### LINKED ISSUE #3928: fix line-dasharray/long-segment render test
The test is failing because the line is clipped by geojson-vt differently in -native than in -js. The buffer parameter for geojson-vt needs to be increased to match -js which means https://github.com/mapbox/geojson-vt-cpp/issues/31 needs to be done.

screen shot 2016-02-12 at 1 56 21 pm

ansis:
fix #3928
fix #3838

:eyes: @jfirebaugh


src/mbgl/style/style_parser.cpp
```diff
@@ -242,16 +241,21 @@ void StyleParser::parseSources(const JSValue& value) {
     }
 }
 
-std::unique_ptr<mapbox::geojsonvt::GeoJSONVT> StyleParser::parseGeoJSON(const JSValue& value) {
+std::unique_ptr<mapbox::geojsonvt::GeoJSONVT> StyleParser::parseGeoJSON(const JSValue& value, std::unique_ptr<SourceInfo>& info) {
```
jfirebaugh: `std::unique_ptr<SourceInfo>&` ⇢ `SourceInfo&`. There's no ownership transfer happening here, so `std::unique_ptr` should not be used.

src/mbgl/style/style_parser.cpp
```diff
@@ -242,16 +241,21 @@ void StyleParser::parseSources(const JSValue& value) {
...
     using namespace mapbox::geojsonvt;
 
+    Options options;
+    options.buffer = util::EXTENT / util::tileSize * 128;
+    options.extent = util::EXTENT;
+    info->maxZoom = options.maxZoom;
+
     try {
-        return std::make_unique<GeoJSONVT>(Convert::convert(value, 0));
+        return std::make_unique<GeoJSONVT>(Convert::convert(value, 0), options);
```
jfirebaugh: Does `GeoJSONVT` provide access to the `options` passed to the constructor? If so, you could avoid passing `SourceInfo` here and instead have the caller set `maxZoom` based on the options from the `GeoJSONVT` return value.

ansis: I changed this in geojsonvt, released a new version and switched to that. Is there anything else that needs to be changed before merging?

src/mbgl/style/style_parser.cpp
```diff
@@ -224,8 +224,10 @@ void StyleParser::parseSources(const JSValue& value) {
                 continue;
             }
 
-            // We always assume the default configuration for GeoJSON sources.
             info = std::make_unique<SourceInfo>();
```
jfirebaugh: Move this line to the top of the `SourceType::GeoJSON` case, and then set `info->maxZoom = geojsonvt->options.maxZoom` right after `geojsonvt = ...`, so that the `if (geojsonvt)` conditional isn't necessary.

jfirebaugh:
Good to merge after addressing the above.

PR #4540, opened 2016-03-30
[core] make sprite metrics checking stricter

### LINKED ISSUE #714: Check that sprite image is large enough
When getting images from the sprite image, we should check that it is at least as big as the position indicated by the JSON metadata.

ansis:
- check that sprite metrics are within the bounds of the sprite. fix #714
- make sprite metrics problems Errors instead of Warnings
- check for negative image width, height

:eyes: @jfirebaugh

### DIFF
--- a/src/mbgl/sprite/sprite_parser.cpp
+++ b/src/mbgl/sprite/sprite_parser.cpp
@@ -20,9 +20,10 @@ SpriteImagePtr createSpriteImage(const PremultipliedImage& image,
                                  const double ratio,
                                  const bool sdf) {
     // Disallow invalid parameter configurations.
-    if (width == 0 || height == 0 || ratio <= 0 || ratio > 10 || width > 1024 ||
-        height > 1024) {
-        Log::Warning(Event::Sprite, "Can't create sprite with invalid metrics");
+    if (width <= 0 || height <= 0 || width > 1024 || height > 1024 ||
+        ratio <= 0 || ratio > 10 ||
+        srcX + width > image.width || srcY + height > image.height) {
+        Log::Error(Event::Sprite, "Can't create sprite with invalid metrics");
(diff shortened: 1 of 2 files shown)


src/mbgl/sprite/sprite_parser.cpp
```diff
@@ -20,9 +20,10 @@ SpriteImagePtr createSpriteImage(const PremultipliedImage& image,
                                  const double ratio,
                                  const bool sdf) {
     // Disallow invalid parameter configurations.
-    if (width == 0 || height == 0 || ratio <= 0 || ratio > 10 || width > 1024 ||
-        height > 1024) {
-        Log::Warning(Event::Sprite, "Can't create sprite with invalid metrics");
+    if (width <= 0 || height <= 0 || width > 1024 || height > 1024 ||
+        ratio <= 0 || ratio > 10 ||
+        srcX < 0 || srcY < 0 || srcX + width > image.width || srcY + height > image.height) {
```
jfirebaugh: Can `srcX + width` or `srcY + height` overflow? Should we check that `width` and `height` are multiples of `ratio`?

ansis: > Can srcX + width or srcY + height overflow?

I think it can't because it will automatically get promoted to an `unsigned int`? I'm not completely sure. When I pass in `2^16 - 2` for srcX and `50` for width the comparison happens without overflow but I'm not sure if that's guaranteed. I tried reading `4.5 Integral promotions` in the spec but it's going over my head a bit.

> Should we check that width and height are multiples of ratio?

No, we currently support icons where this isn't true: https://github.com/mapbox/mapbox-gl-native/pull/3561

jfirebaugh:
:+1:

Issue #4542, opened 2016-03-30
create github issue template


ansis:
https://help.github.com/articles/creating-an-issue-template-for-your-repository/

It's been working well in mapbox-gl-js.

friedbunny:
[GL-JS’s issue template:](https://github.com/mapbox/mapbox-gl-js/blob/7b61aa9888bd98c3b6be0a9150225a1871fdae31/ISSUE_TEMPLATE.md)

```
<!--
Hello! Thanks for contributing. To help us diagnose your problem quickly, please:

 - include a link to a minimal demonstration of the bug; we recommend using https://jsbin.com
 - ensure you can reproduce the bug using the latest release
 - only post to report a bug or request a feature; direct all other questions to https://stackoverflow.com/questions/tagged/mapbox-gl-js
-->

**mapbox-gl-js version**:

### Steps to Trigger Behavior

 1.
 2.
 3.

### Expected Behavior

### Actual Behavior
```

We would want to ask about platform, as well. Anything else?

1ec5:
Beyond that, it really depends on the platform, so probably not. I don't think we should make contributors wade through a flowchart to report an issue. So just the platform is fine.

friedbunny:
Thanks for the kick @ansis, the GL-JS template looks really solid (and the commented instructions block is :100:). Let’s move implementation discussion to the pull request I just opened: #4545.

PR #4545, opened 2016-03-31
Add GitHub issue template!

friedbunny:
Add an issue template that everyone will see when creating a new issue here on GitHub. (Pull requests will not see this template, that would need a separate file: `PULL_REQUEST_TEMPLATE.md`)

Cribbed from [Mapbox GL JS](https://github.com/mapbox/mapbox-gl-js/blob/7b61aa9888bd98c3b6be0a9150225a1871fdae31/ISSUE_TEMPLATE.md), per @ansis’s suggestion in #4542.

**Raw template:**

screen shot 2016-03-30 at 8 13 58 pm

**Markdown preview:**

screen shot 2016-03-30 at 8 14 07 pm

/cc @mapbox/gl

### DIFF
--- a/ISSUE_TEMPLATE.md
+++ b/ISSUE_TEMPLATE.md
@@ -1,0 +1,20 @@
+<!--
+Hello and thanks for contributing! To help us diagnose your problem quickly, please:
+
+ - Include a minimal demonstration of the bug, including code, logs, and screenshots.
+ - Ensure you can reproduce the bug using the latest release.
+ - Only post to report a bug or request a feature; direct all other questions to: https://stackoverflow.com/questions/tagged/mapbox
+-->
+
+**Platform:**
+**Mapbox SDK version:**
+
+### Steps to trigger behavior
+
+ 1.
+ 2.
+ 3.
+
+### Expected behavior
+
+### Actual behavior


mb12:
@friedbunny Can the version be a drop down menu that one could select instead of typing it in?

mourner:
@mb12 GitHub doesn't support dropdown menus in templates as far as I know.

friedbunny:
@mb12 A little more sophistication would be nice, but Markdown is as much as we’re allowed to do in GitHub issue and pull request bodies — only text formatting HTML tags.

jfirebaugh:
:+1:

bleege:
So happy to not have to sound like a broken record anymore.  :smiley:

brunoabinader:
> Ensure you can reproduce the bug using the latest release.

Can we suggest the user to include the exact commit hash?

This is awesome @friedbunny!

1ec5:
> Can we suggest the user to include the exact commit hash?

That’s quite difficult for an SDK user to find. The SDK version is enough for us to trace back to a commit, since we tag each release in this repository by SDK version.

tmpsantos:
:+1:

PR #4547, opened 2016-03-31
[core] simplify conversions from pixels to tile units

### LINKED ISSUE #3768: simplify conversion from pixels to tile units
port https://github.com/mapbox/mapbox-gl-js/commit/282a12cfcc38784e6947e686ed782383f7930480

ansis:
fix #3768

:eyes: @jfirebaugh 

This is slightly different than the -js implementation. In -js `pixelsToTileUnits` its just a regular function. Here it's a method on TileID. I think this makes sense until we refactor TileID/TileCoord and get rid of the -js and -native differences.


jfirebaugh:
:+1:

Issue #4501, opened 2016-03-29
Hang on launch after downloading 776.6 MB of offline resources — mbgl::OfflineDatabase::migrateToVersion3()


1ec5:
iosapp hangs for about a minute on every launch after downloading the large amount of offline data described in https://github.com/mapbox/mapbox-gl-native/pull/4482#issuecomment-202633356, pictured below:

![california-cities](https://cloud.githubusercontent.com/assets/1231218/14094240/1c52e11a-f509-11e5-85c9-c7086a1ede61.PNG)

Below is the stack track in iosapp. The hang is triggered by the first MGLMapView being loaded, which causes the shared MGLOfflineStorage object to be created. Even though osxapp shares virtually all of the same offline code, it doesn’t appear to be affected by the hang at all.

/cc @jfirebaugh

```
Thread 1 Queue : com.apple.main-thread (serial)
#5	0x00000001826f4130 in std::__1::future<void>::get() ()
#6	0x00000001006d4ecc in mbgl::util::Thread<mbgl::DefaultFileSource::Impl>::Thread<std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char> > const&, unsigned long long&>(mbgl::util::ThreadContext const&, std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char> > const&&&, unsigned long long&&&) at /path/to/mapbox-gl-native/gyp/../src/mbgl/util/thread.hpp:109
#7	0x00000001006d4ca4 in mbgl::util::Thread<mbgl::DefaultFileSource::Impl>::Thread<std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char> > const&, unsigned long long&>(mbgl::util::ThreadContext const&, std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char> > const&&&, unsigned long long&&&) at /path/to/mapbox-gl-native/gyp/../src/mbgl/util/thread.hpp:90
[...]

PR #4502, opened 2016-03-31
Remove blocking on future in Thread constructor

jfirebaugh:
The `Thread` constructor has [this line](https://github.com/mapbox/mapbox-gl-native/blob/8112ac3371230d0d99733c4b71fb2fac3d610404/src/mbgl/util/thread.hpp#L109-L109), which blocks the launching thread until construction of the `Object` template parameter has completed on the launched thread. This is not necessary, and should be removed so as to avoid potentially blocking the main thread for an indeterminate amount of time.

It isn't necessary because there's no way that launching thread can access `object` directly. It can only ask the `Thread` to `invoke` methods, and execution of the invocation is guaranteed to happen, on the launched thread, only after `object` is constructed.

This was a factor in #4501.

### DIFF
--- a/include/mbgl/util/run_loop.hpp
+++ b/include/mbgl/util/run_loop.hpp
@@ -47,7 +47,7 @@ public:
 
     // Invoke fn(args...) on this RunLoop.
     template <class Fn, class... Args>
-    void invoke(Fn&& fn, Args&&... args) {
+    void invoke(Fn fn, Args... args) {
         auto tuple = std::make_tuple(std::move(args)...);
         auto task = std::make_shared<Invoker<Fn, decltype(tuple)>>(
             std::move(fn),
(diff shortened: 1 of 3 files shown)


jfirebaugh:
Turns out the `running` promise/future _is_ necessary, because the launching thread needs to at least wait for the launched thread's `RunLoop` to be constructed, and `RunLoop` doesn't currently support cross-thread ownership transfer. But fulfillment of the promise can happen before constructing `Object`.

cc @tmpsantos

tmpsantos:
:+1: 

> doesn't currently support cross-thread ownership transfer.

This is probably not worth doing because some main loops implementations backing the `RunLoop` make assumptions of which thread they get initialized.

jfirebaugh:
Never got this to pass tests. Abandoning.

jfirebaugh:
Never got this to pass tests. Abandoning.

PR #4557, opened 2016-03-31
[glfw] Annotations are misplaced when zooming out

brunoabinader:
Steps to reproduce:

1. Run GLFW client `make run-linux`
2. Zoom out e.g. via mouse scroll
3. Press `3` to add some point annotations - usually these are randomly placed in both latitude and longitude, but they all appear at the horizontal edges of the screen (screenshot below):
screen shot 2016-03-31 at 3 49 59 pm

This issue seems old (if not always present), as I bisected it up to 3 months old commits. I suspect the following:
- We're not properly clamping the scale/zoom value when issuing `scaleBy`; or
- There's a bug in `PointAnnotationImpl::updateLayer`; or
- `GLFWView::makeRandomPoint` is bogus for the following scenario.

/cc @mapbox/gl

### DIFF
--- a/platform/default/glfw_view.cpp
+++ b/platform/default/glfw_view.cpp
@@ -211,13 +211,9 @@ void GLFWView::onKey(GLFWwindow *window, int key, int /*scancode*/, int action, 
 }
 
 mbgl::LatLng GLFWView::makeRandomPoint() const {
-    const auto nw = map->latLngForPixel({ 0, 0 });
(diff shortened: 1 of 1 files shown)


brunoabinader:
Rationale for this fix is that the previous calculation assumed that screen boundaries always had opposite longitude coordinates, which is not true for cases e.g. when the whole world is shown.

jfirebaugh:
:+1:

PR #4559, opened 2016-03-31
[ios, osx] Replace libuv with native implementations

jfirebaugh:
Progress toward #2909.

:eyes: @1ec5 @tmpsantos


tmpsantos:
:+1: This ended up cleaner than I was expecting.

jfirebaugh:
This is passing Node OS X builds on Bitrise while failing locally. I believe it should be failing on Bitrise -- with this implementation as it stands, the Node bindings on OS X are using a mix of libuv and Cocoa APIs, which may produce unpredictable results. Instead,  the Node bindings on OS X should continue to use libuv as the loop implementation. I'm investigating.

jfirebaugh:
#4560

1ec5:
:+1:

Issue #4365, opened 2016-03-17
TestApp refactor


tobrun:
#### Test App Refactor

Currently our test nav drawer is like:

![options](https://cloud.githubusercontent.com/assets/2151639/13855821/88127c36-ec48-11e5-850e-fa5354894afb.png)
![device-2016-03-17-140230](https://cloud.githubusercontent.com/assets/2151639/13855897/ea7674cc-ec48-11e5-9963-e74c18f83489.png)
![device-2016-03-17-140046](https://cloud.githubusercontent.com/assets/2151639/13855898/ea79a516-ec48-11e5-9cd8-7178ba03fe42.png)

The main goal here is to hit the following things:
 - [x] make more clear feature activities for our users
 - [x] organise it in categories and add some explanation
 - [x] make it easier for our users to play around with the app without having to build from source

##### Make more clear feature activities for our users
Simplify the examples we expose, make them a bit more smart eg:

> InfoWindowActivity and InfoWindowConcurrentActivity can be merged together and we can define an OptionsMenu to switch the state

Update outdated examples eg:

> CameraActivity is only show casing ease, move etc. and is not show casing how to use it for other CameraUpdateFactory or allow to change variables.

##### Organise it in categories and add some explanation

Currently the testapp  shows a map, we discussed that having a list of different feature activities, showing 2 lines, could be more usefull for external people. It also removes the need to having different icons for our testactivity and add the test case that people are coming from and moving back to an Activity. Just in the same way our users are doing. 

An example of Material design list with 2 lines + header:
![components_lists_usage3](https://cloud.githubusercontent.com/assets/2151639/13855712/2040f060-ec48-11e5-929f-8f15fb865e40.png)

##### Make it easier for our users to play around with the app without having to build from source

Find a way to expose a ready build apk for our users. This step must be automated and must not interfere with git. The most feasible solution would be to upload the the testapk from Bitrise when doing a nightly build.

tobrun:
cc @bleege @zugaldia @cammace @danswick

zugaldia:
:+1: :+1: :+1:

tobrun:
First up to tackle is `providing a downloadble apk file`:
 - make it easy for users to test out our example app without building from source
 - will be part of our bitrise SNAPSHOT build
 - versioning could be a nice to have (eg. git bisect)
 - may not interfere with our git workflow

Their are many ways to do this.

- Serve it as release on github
[...]

Issue #4531, opened 2016-03-30
Add User Location example to the test app


tobrun:
Our old `MainActivity` was removed in favour of a `FeatureOverviewActivity` in #4365.
Currently we have no way of showcasing the possibility to toggle the user location on/off.
We should add a separate Feature activity covering this.

I'm suggesting to add a new Category `User Location` and move the `UserLocationTracking` with the new created Activity for this feature to that category.

This issue should also implement a clean way to handle runtime permissions and port that solution to `UserLocationTracking` example.

bleege:
Closing as the work is being done in #4835 .

PR #4570, opened 2016-04-01
4365 test app refactor

tobrun:
Refs #4365.

Final cleanup of Test App.

* category packages
untitled

* ApiAcces was removed from SDK in `4.0.0` and is now replaced with a simple getString().
* Bugfix for SnapShotActivity

:eyes: @bleege @zugaldia


bleege:
Yep, I'll be happy to :eyes: this later today.

bleege:
I built the new TestApp from source and overall it's looking great!  That said, I did find some things that need some more work.

* [ ] `Polygone` is spelled wrong
* [ ] `PolygonActivity` doesn't have an AppBar (see screenshot)
* [ ] `CameraAnimationTypeActivity` has it's buttons on the top (see screenshot)
* [ ] `MapFragmentActivity` and `SupportMapFragmentActivity` should use the same UI to demonstrate that they both work the same.
* [ ] `GeocoderActivity` does geocoding where black marker is, not where i tap.  The on screen instructions imply the opposite.
* [ ] `MyLocationTrackingModeActivity` - Modes are reset when i tap on the map.
* [ ] Rename app to `Mapbox Android SDK TestApp`

One other thing that I think we need mull over is that even though using Java subpackages is Standard Operation Procedure for apps, in this case it may working against one of the primary use cases for the TestApp; namely making discoverability more difficult for developers looking for example code.  Having a single `activities` package where all the example code lives would let developers see all examples immediately.  Thoughts?

device-2016-04-01-160146
_Polygone_

device-2016-04-01-160551
_AppBar Is AWOL_

device-2016-04-01-160532
_Buttons Have Gone Rogue_

bleege:
I also received the following MemoryLeak report from LeakCanary when testing:

```java
In com.mapbox.mapboxsdk.testapp:4.0.0:8.
* FAILURE:
java.lang.OutOfMemoryError: OutOfMemoryError thrown while trying to throw OutOfMemoryError; no stack available

* Reference Key: 0f38b58a-b60a-4edc-9699-afce8929482d
* Device: samsung samsung SAMSUNG-SM-G900A klteuc
* Android Version: 5.0 API: 21 LeakCanary: 1.4-beta1 02804f3
* Durations: watch=5082ms, gc=170ms, heap dump=5330ms, analysis=19365ms
* Excluded Refs:
| Field: android.app.ActivityThread$ActivityClientRecord.nextIdle
| Field: android.view.inputmethod.InputMethodManager.mNextServedView
| Field: android.view.inputmethod.InputMethodManager.mServedView
| Field: android.view.inputmethod.InputMethodManager.mServedInputConnection
| Field: android.view.inputmethod.InputMethodManager.mCurRootView
| Field: android.animation.LayoutTransition$1.val$parent
| Field: android.view.textservice.SpellCheckerSession$1.this$0
| Field: android.support.v7.internal.widget.ActivityChooserModel.mActivityChoserModelPolicy
| Field: android.widget.ActivityChooserModel.mActivityChoserModelPolicy
| Field: android.accounts.AccountManager$AmsTask$Response.this$1
| Field: android.media.MediaScannerConnection.mContext
| Field: android.os.UserManager.mContext
| Field: android.appwidget.AppWidgetHost$Callbacks.this$0
| Field: android.sec.clipboard.ClipboardUIManager.mContext
| Field: android.media.AudioManager$1.this$0
| Field: android.view.Choreographer$FrameDisplayEventReceiver.mMessageQueue (always)
| Static field: android.media.session.MediaSessionLegacyHelper.sInstance
| Static field: android.text.TextLine.sCached
| Static field: android.widget.TextView.mLastHoveredView
| Thread:FinalizerWatchdogDaemon (always)
| Thread:main (always)
| Thread:LeakCanary-Heap-Dump (always)
| Class:java.lang.ref.WeakReference (always)
| Class:java.lang.ref.SoftReference (always)
| Class:java.lang.ref.PhantomReference (always)
| Class:java.lang.ref.Finalizer (always)
| Class:java.lang.ref.FinalizerReference (always)
| Root Class:android.os.Binder (always)
```

tobrun:
@bleege thank you for the helpful :eyes:, those are great catches! 
Most of them are implemented. Below you will find some additional feedback.

#### Geocoder
> GeocoderActivity does geocoding where black marker is, not where i tap. The on screen instructions imply the opposite.

The current instruction text is `Tap the map to trigger the geocoder.`. No able to see the link between tap location and geocoding location. Feel free to provide an alternative text.

#### MyLocationTracking
> MyLocationTrackingModeActivity - Modes are reset when i tap on the map.

This is by design and is configurable in `TrackingSettings.java`

```java
 void setDismissTrackingOnGesture(boolean dismissTrackingOnGesture)
```
The default behaviour was set to true to have parity with iOS to support tracking for ReactNative. 

**Update** I just added a checkbox options menu item to toggle the state of dismissTrackingOnGesture

untitled

#### App name
>  Rename app to Mapbox Android SDK TestApp

Both in app title as title in application menu has been changed. (used to be Mapbox GL)

#### Packaging

Could you share some more information about the requirement of having all activities in one package?

The current type makes the most logical categorisation and follows the order shown in the testapp:

image

tobrun:
There is one ticket remaining for this in #4531

bleege:
The updates are looking good @tobrun!  Here's my second round of feedback for 

**Geocoder**
I made the text more descriptive so that it's a bit more clear what to do and what's happening.  Originally it was `Tap the map to trigger the geocoder.` and now it's `Tap Map To Geocode Where Black Marker Is Currently Located`.  

**MyLocationTracking**

> The default behaviour was set to true to have parity with iOS to support tracking for ReactNative. 

Can you provide more detail on this?  I'm just curious of the original iOS logic behind this as it's seems odd to me that Bearing UI should disappear simply due to a zoom out.  Maybe I'm wrong though.  Regardless, the toggle in the short term is a good workaround.

**App Name**

Yep, the old name had been around for too long.  We don't even refer to things as "GL" anymore.  Thanks for updating it!  :+1: 

**Packaging**

> Could you share some more information about the requirement of having all activities in one package?
> The current type makes the most logical categorisation and follows the order shown in the testapp:

Yep, this is definitely one of those things that are more "how do humans work" / behavioral economics issues than logical app design.  While the graphic above is absolutely :100: accurate, it's more likely that developers (aka customers) will be searching through the Github repo looking for code examples by name than first setting up the app, looking at the menu, and then trying to find the matching code.  In this case it makes it much easier to present a single source of content that's immediately all visible (and searchable within the browser) to them rather than having content "hidden" in sub-packages.  Here's an example:

Starting at the same spot in the repo package structure, which is easier to find something related to `MyLocation`?  Is it obviously clear to anyone not familiar with the TestApp to look in the proper subpackage for it or to know that an example even exists?  If they don't think there is an example what will that do to their impression of Mapbox Android SDK?  Will they still view it as complete?  Will they still view it as a useful tool for solving their needs?  Will they think it's easy to use?

Does this make more sense now?

screen shot 2016-04-04 at 2 55 35 pm
_Sample Code In Multiple Sub Packages_

screen shot 2016-04-04 at 1 57 31 pm
_Single Example Code Package_

bleege:
@tobrun Let's keep #4531 separate from #4365 and this PR.  We can always add it afterwards as it'll be more helpful to land this first.  From my perspective we just need to find an agreed upon solution for the package structure question and then we can merge.  What do you think?

tobrun:
We are going with full package so we close this PR

bleege:
Discussed internally with @tobrun and we decided to ship with the subpackages for activities, but will rename them to be more atomic and intuitively named.

bleege:
I just rearranged the examples subpackage so that the the content is more descriptive / more discoverable when the code is viewed on GitHub.  I also updated the category layout within the app so that it mirrors the new subpackages [as before to keep app to code way finding possible](https://github.com/mapbox/mapbox-gl-native/pull/4570#issuecomment-205201837).

screen shot 2016-04-05 at 11 24 11 am

bleege:
Rebased and Merged!

PR #4574, opened 2016-04-01
[ios] Run core unit tests in CI

jfirebaugh:


1ec5:
How is `ios-sim` being installed? Is it provided by Bitrise?

jfirebaugh:
@1ec5 I don't know. I haven't successfully got the build to that point yet. It seems to be hanging on the first `xcodebuild` command for some reason:

https://www.bitrise.io/build/2ec0b5f61e76c0ef

Any ideas?

1ec5:
Try nixing the package.sh invocation for now. Maybe the normal SDK build is interfering somehow?

jfirebaugh:
Nope. Still hangs.

https://www.bitrise.io/build/951cf169a62f3b51

jfirebaugh:
http://devcenter.bitrise.io/docs/step-hangs-times-out-after-a-period-without-any-logs-on-bitrise

I guess I need to figure out how to get gyp to generate a scheme that's marked as "shared" when running `make Xcode/ios`.

1ec5:
Ah, in that case, check out platform/osx/scripts/test.sh (`make xctest`). I have quite the hack for you in there.

jfirebaugh:
Heh, wow, quite the hack.

I was going to go with [a slightly different approach](https://github.com/mapbox/mapbox-gl-native/pull/4574/commits/f1c970375fff24cf2f18b41ec66cf69c837fac67), of checking in the `.xcscheme` file from my local checkout, and copying it into place in the build script. That would be a temporary hack until #3275. Any reason to prefer the `open / sleep` approach instead?

1ec5:
If xcodebuild is able to use that scheme, then great! I’m all for finding alternatives to my hack.

jfirebaugh:
Well, it's successfully running the tests on iOS CI, but we can't see the output. 😢

1ec5:
The simulator does support output to a file descriptor, but I'm not sure if ios-sim supports it. If it doesn't, you could also log to a file and print that file in a subsequent step that runs unconditionally.

jfirebaugh:
ios-sim is supposed to handle that automatically. I discovered that there's [a bug](https://github.com/phonegap/ios-sim/issues/134) in the 3.1.1 version of ios-sim in homebrew which was fixed in 3.2.0. I submitted a [homebrew pull request](https://github.com/Homebrew/homebrew-core/pull/171) to upgrade.

PR #4576, opened 2016-04-01
[node] Properly inherit from EventEmitter

vkurchatkin:
`process.EventEmitter` is deprecated, so we should use `require('events')` instead.

Also added call to `EventEmitter` to initialise instance properties.


mikemorris:
Merged in https://github.com/mapbox/mapbox-gl-native/commit/e886eb95455335818972cafd65923a7fe8d83e49, thanks!

PR #4580, opened 2016-04-02
[build] Be not afraid of standard Makefile syntax, whitespace, and build output

jfirebaugh:


friedbunny:
This makes a huge difference in readability — thanks @jfirebaugh.

Issue #4586, opened 2016-04-04
Get correct aws user rights for publishing test apk


tobrun:
Currently our daily snapshots are failing because uploading to s3 in't working due to missing user rights.
 - [x] Remove publish to `s3` step in Bitrise buildscript
 - [x] Fix user rights and restore buildscript

tobrun:
Closing as https://github.com/mapbox/mapbox-android-demo fulfills this use-case.

PR #4587, opened 2016-04-04
[android] #4586 - disable publish s3

tobrun:
Refs #4586


tobrun:
CI approved, let's merge this is in before SNAPSHOT is being build

PR #4589, opened 2016-04-04
[build] Apply Makefile changes for {ios,isim,ibench}

brunoabinader:
Prevents the following error output on Linux:
```Makefile:73: *** recipe commences before first target.  Stop.```

:eyes: @jfirebaugh

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -67,15 +67,19 @@ iproj: Xcode/ios
 	open ./build/ios-all/gyp/ios.xcodeproj
 
 .PHONY: ios ibench isim
-ios ibench isim: export XCODEBUILD_ARGS += PROVISIONING_PROFILE="$$(PROVISIONING_PROFILE)"
+ios ibench isim:
+	export XCODEBUILD_ARGS += PROVISIONING_PROFILE="$$(PROVISIONING_PROFILE)"
 
-ios: export XCODEBUILD_ARGS += -sdk iphoneos ARCHS="arm64 armv7 armv7s"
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

**Comment by jfirebaugh** [comment]
I don't think this is correct. The `export XCODEBUILD_ARGS += ...` syntax is [Makefile target-specific  variable syntax](https://www.gnu.org/software/make/manual/html_node/Target_002dspecific.html#Target_002dspecific), not recipe (shell) syntax.

**Comment by brunoabinader** [comment]
This is what the [docs says about the given error](https://www.gnu.org/software/make/manual/html_node/Error-Messages.html):
> This means the first thing in the makefile seems to be part of a recipe: it begins with a recipe prefix character and doesn’t appear to be a legal make directive (such as a variable assignment). Recipes must always be associated with a target.

If I understood it correctly, this means there was an illegal character preventing the directive from being legal. Putting a next line to it could have solved it, but the fact that the docs link you gave have this example - ```prog : CFLAGS = -g``` - puzzles me on why Linux make is complaining.

**Comment by jfirebaugh** [comment]
I'm getting these warnings after this change:

```
Makefile:74: warning: overriding commands for target `ios'
Makefile:71: warning: ignoring old commands for target `ios'
Makefile:78: warning: overriding commands for target `isim'
Makefile:71: warning: ignoring old commands for target `isim'
Makefile:82: warning: overriding commands for target `ibench'
Makefile:71: warning: ignoring old commands for target `ibench'
```

PR #4599, opened 2016-04-04
[core] warn when a geometry exceeds the maximum allowed values

ansis:
port https://github.com/mapbox/mapbox-gl-js/pull/2383

:eyes: @jfirebaugh

### DIFF
--- a/src/mbgl/util/get_geometries.cpp
+++ b/src/mbgl/util/get_geometries.cpp
@@ -1,9 +1,15 @@
 #include <mbgl/util/get_geometries.hpp>
+#include <mbgl/platform/log.hpp>
 #include <mbgl/util/constants.hpp>
 
 #include <cmath>
 
 namespace mbgl {
+
+// the minimum position we can store in the buffers
+constexpr int32_t EXTENT_MIN = util::EXTENT * -2;
+// // the maximum position we can store in the buffers
+constexpr int32_t EXTENT_MAX = (util::EXTENT * 2) - 1;
 
 GeometryCollection getGeometries(const GeometryTileFeature& feature) {
     const float scale = float(util::EXTENT) / feature.getExtent();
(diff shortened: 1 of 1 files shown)


src/mbgl/util/get_geometries.hpp
```diff
@@ -2,9 +2,15 @@
 #define MBGL_UTIL_GET_GEOMETRIES
 
 #include <mbgl/tile/geometry_tile.hpp>
+#include <mbgl/util/constants.hpp>
 
 namespace mbgl {
 
+// the minimum position we can store in the buffers
```
jfirebaugh: Move these constants to the .cpp file, since they're only used there.

ansis: thanks, done

mb12:
@ansis  This does not look right.  A valid vector tile should have coordinates within [0, extent).  The invariant is even enforced by the  mapnik-vector-tile writer (Its a Mapbox project that's called mapnik for probably legacy reasons). Since it uses ClipperLib, in practice it actually generates [0, extent].

ansis:
> @ansis This does not look right. A valid vector tile should have coordinates within [0, extent). The invariant is even enforced by the mapnik-vector-tile writer (Its a Mapbox project that's called mapnik for probably legacy reasons). Since it uses ClipperLib, in practice it actually generates [0, extent].

Geometries can (and very often do) extend past the tile boundaries:

> A layer MUST contain an extent that describes the width and height of the tile in integer coordinates. The geometries within the Vector Tile MAY extend past the bounds of the tile's area as defined by the extent. Geometries that extend past the tile's area as defined by extent are often used as a buffer for rendering features that overlap multiple adjacent tiles.

https://github.com/mapbox/vector-tile-spec/tree/master/2.1#41-layers

The current implementation in mapbox-gl-native only supports geometries that extend a limited distance from the tile edges. This distance is limited by the precision we have for storing positions in the vertex buffers. This adds a warning for when geometries extend past that limit.

Eventually we may want to clip geometries instead of warning.

mb12:
@ansis Is there any advantage to generating geometries outside the tile boundaries? Is it correct (or not) that everything outside the tile boundary is stenciled out?

ansis:
> ansis Is there any advantage to generating geometries outside the tile boundaries? Is it correct (or not) that everything outside the tile boundary is stenciled out?

Yes, there are advantages to having this data in the vector tiles. It is correct that everything outside the tile boundary is stenciled out.

- the geometries for symbols outside tile boundaries are needed to do cross-tile label placement without collisions. This non-global approach for label placement is needed so that we can render png tiles with cross-tile labels
- the buffer is also needed to correctly draw lines and circles for png tiles. Sometimes a geometry is entirely within one tile but covers pixels in another tile when styled.

mb12:
Thank you very much @ansis for the detailed explanation.

jfirebaugh:
This warning turned out to be problematic in gl-js (https://github.com/mapbox/mapbox-gl-js/issues/2568), and ultimately not very useful. Let's not bother.

PR #4601, opened 2016-04-05
Build http_request_nsurl.mm with ARC

jfirebaugh:
We should build http_request_nsurl.mm with ARC, like the rest of our Objective C code.

Flip [this line](https://github.com/mapbox/mapbox-gl-native/blob/65acf35dcfa3784de7e522926469083068599550/gyp/http-nsurl.gypi#L28) and fix the compile errors.

### DIFF
--- a/platform/darwin/src/http_request_nsurl.mm
+++ b/platform/darwin/src/http_request_nsurl.mm
@@ -19,7 +19,6 @@ class HTTPNSURLContext;
 class HTTPNSURLRequest : public HTTPRequestBase {
 public:
     HTTPNSURLRequest(HTTPNSURLContext*, Resource, Callback);
-    ~HTTPNSURLRequest();
 
     void cancel() final;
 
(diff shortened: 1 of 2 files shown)


jfirebaugh:
👀  @1ec5

mb12:
@jfirebaugh If you want to be superrrr cautious, just do a quick sanity test with leaks in xcode. If there is anything obvious, it would show up.

platform/darwin/src/http_request_nsurl.mm
```diff
@@ -147,11 +126,7 @@
 }
 
 void HTTPNSURLRequest::cancel() {
-    if (task) {
-        [task cancel];
-        [task release];
-        task = nullptr;
-    }
+    [task cancel];
```
1ec5: It's good practice to still nil out the task after you cancel it here.

jfirebaugh:
Restored nil assignment; no leaks reported in Instruments.

PR #4608, opened 2016-04-06
[WIP] [build] Refactor and simplify build system

### LINKED ISSUE #1372: mesa version is specified in three different places
https://github.com/mapbox/mapbox-gl-native/blob/c27f84e75364bc3fa4562ab8a399c6e11dd2e363/configure#L55-L56
https://github.com/mapbox/mapbox-gl-native/blob/2b633cc51cd44be3f0224f0b105b18f0a3d7d657/scripts/linux/install.sh#L13-13
https://github.com/mapbox/mapbox-gl-native/blob/2b633cc51cd44be3f0224f0b105b18f0a3d7d657/scripts/linux/run.sh#L31-31

### LINKED ISSUE #3388: Build iosapp on Bitrise
Just like Travis runs `make run-xosx` to ensure that everything all the way up to osxapp builds and runs, we need a `make run-ios` that builds and runs iosapp. `make itest` is good for ensuring that the iOS SDK integrates with ordinary iOS applications, but iosapp is… special. (It will become markedly less special once #3183 lands.)

/ref #3387
/cc @jfirebaugh @incanus

### LINKED ISSUE #3452: iOS benchmarking app fails to link — mbgl::SharedSQLiteCache::get, mapbox::geojsonvt::Convert::convert
The ios-bench target fails to link with the following errors:

```
Undefined symbols for architecture arm64:
  "mbgl::SharedSQLiteCache::get(std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char> > const&)", referenced from:
      -[MGLMapView commonInit] in libmbgl-platform-ios.a(MGLMapView.o)
  "mapbox::geojsonvt::Convert::convert(rapidjson::GenericValue<rapidjson::UTF8<char>, rapidjson::MemoryPoolAllocator<rapidjson::CrtAllocator> > const&, double)", referenced from:
      mbgl::SourceInfo::parseGeoJSON(rapidjson::GenericValue<rapidjson::UTF8<char>, rapidjson::MemoryPoolAllocator<rapidjson::CrtAllocator> > const&) in libmbgl-core.a(source_info.o)
ld: symbol(s) not found for architecture arm64
```

/cc @kkaefer

### LINKED ISSUE #3539: flaky Android build
The Android build will sometimes fail with the following output:

```
make[2]: *** No rule to make target `Release/obj.target', needed by `/bitrise/src/platform/android/MapboxGLAndroidSDK/src/main/obj.target'.  Stop.
make[2]: *** Waiting for unfinished jobs....
  CXX(target) Release/obj.target/example-custom-layer-lib/platform/android/src/example_custom_layer.o
../../platform/android/src/http_request_android.cpp:191:59: warning: unused parameter 'message' [-Wunused-parameter]
void HTTPAndroidRequest::onResponse(int code, std::string message, std::string etag, std::string modified, std::string cacheControl, std::string expires, std::string body) {
                                                          ^
1 warning generated.
make[2]: Leaving directory `/bitrise/src/build/android-arm-v7'
make[1]: Leaving directory `/bitrise/src'
make[1]: *** [Makefile/androidapp] Error 2
make: *** [android-lib] Error 2
==> Script finished with exit code: 2
ERRO[21:59:51] Step (Build Android SDK For arm) failed, error: (exit status 2) 
```

I don't think the "warning: unused parameter" is relevant (although we should fix that). I have no idea what "No rule to make target `Release/obj.target'" means, but it seems to be the cause. There's nothing before that.

### LINKED ISSUE #3821: Rename 'linux' target to 'glfw'
We've been able to build `linux` target in OS X for a while now. This target uses [GLFW](http://www.glfw.org/index.html) for rendering, and thus I feel having `glfw` target would be less confusing.

/cc @mapbox/gl

### LINKED ISSUE #4582: Eliminate gyp/build output
Some build products get output to `gyp/build` rather than `build`. This is probably a historical accident or quirk of gyp. We should adjust the build to eliminate this special case output directory -- everything should go to `build`.

### LINKED ISSUE #4600: Bitrise OS X node tests should fail if output fails
Bitrise reports https://www.bitrise.io/build/6933d45bafef6b66 as successful despite having failing test comparisons (`14 failed (4.7%)`)

jfirebaugh:
Fixes #3821
Fixes #3539
Fixes #3452
Fixes #1372
Fixes #4600
Fixes #4582 
Fixes #3388


tmpsantos:
Thanks for working on this. 

My understanding is that now each platform is completely self contained. So adding a new platform (i.e. GTK+ or Qt) would be a matter of creating a a `platform/[gtk,qt,etc]/platform.gyp`. The `Makefile` will handle the default platforms but you could still create your own platform and build using `PLATFORM=gtk make`. This way third parties could maintain their platforms easily as a git submodule. Correct?

Makefile
```diff
@@ -210,32 +176,14 @@ compdb:
 tidy:
 	$(RUN) tidy
 
-.PHONY: clear_xcode_cache
-clear_xcode_cache:
-ifeq ($(BUILD), osx)
```
boundsj: @jfirebaugh @1ec5 is it really ok that we will no longer clear the derived data folder for osx builds?

1ec5: We should continue to clear the derived data. (It’s possible to clear out data derived from building core and the SDK without clearing out data derived from gyp project generation.)

However, I don’t think the derived data is being laid down in ~/Library/Developer/Xcode/DerivedData/{mapboxgl-app,osxapp}-* anyways. It seems to be going to ~/Library/Developer/Xcode/DerivedData/osx-* instead, at least when you build inside Xcode. Maybe it goes in ~/Library/Developer/Xcode/DerivedData/{mapboxgl-app,osxapp}-* if you run `make run-osx`?

jfirebaugh: Output can go to a variety of places depending on the build command, but as @1ec5 notes, none of those places are `~/Library/Developer/Xcode/DerivedData/{mapboxgl-app,osxapp}-*`. Sorting this out, and getting the output to _always_ go to `build`, is #4582. I hope to get to that as part of this PR.

jfirebaugh:
Note that I dropped support for builds such as `HTTP=curl LOOP=uv make glfw-app` with this change. I tried several different ways to continue support for this before concluding that the complexity they added just wasn't worth it.

If you are super dependent on being able to use the curl HTTP implementation or libuv loop implementation on OS X, let's talk.

mikemorris:
Is there a recommended method for customizing builds now then? Specifically asking in context of https://github.com/mapbox/mapbox-gl-native/pull/4638, which will provide an alternate rendering pipeline for Linux machines without a graphics card.

PR #4614, opened 2016-04-06
[android] #3643 - improved touch target with getAnnotationsInBounds

### LINKED ISSUE #3643: getAnnotationsInBounds not returning in bounds annotations

tobrun:
closes #3643

### DIFF
--- a/platform/android/MapboxGLAndroidSDK/src/main/java/com/mapbox/mapboxsdk/maps/MapView.java
+++ b/platform/android/MapboxGLAndroidSDK/src/main/java/com/mapbox/mapboxsdk/maps/MapView.java
@@ -23,7 +23,6 @@ import android.location.Location;
 import android.net.ConnectivityManager;
 import android.net.NetworkInfo;
 import android.net.Uri;
-import android.os.AsyncTask;
 import android.os.Bundle;
 import android.support.annotation.CallSuper;
 import android.support.annotation.FloatRange;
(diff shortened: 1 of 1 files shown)


tobrun:
CI approved, let's merge this in

PR #4615, opened 2016-04-06
[Android] Telemetry Service will be disable at first launch

vacxe:
If user simple add MapView (or fragment with them) him don't need use TelemetryService

At first time it fix necessity declare service in manifest.xml 
At second time if user really need use telemetry him call setTelemetryEnabled(boolean) method
At third at secure this beforehand  https://github.com/mapbox/mapbox-gl-native/commit/9324862aad5d71b43679dd59a5e48087a45446c5 

@tobrun @bleege


vacxe:
Sorry, im read documentation and understood why it is needed

bleege:
No problem @vacxe!  Thanks for contributing!

PR #4616, opened 2016-04-06
Final merge of iOS 3.2.0 into Android 4.0.0

boundsj:
@1ec5 @friedbunny since not much has changed in the release branch since the last merge, we can go ahead and close out this branch simply using a PR and Github merge.

cc @mapbox/mobile

### DIFF
--- a/platform/ios/Mapbox-iOS-SDK-symbols.podspec
+++ b/platform/ios/Mapbox-iOS-SDK-symbols.podspec
@@ -1,7 +1,7 @@
 Pod::Spec.new do |m|
 
   m.name    = 'Mapbox-iOS-SDK'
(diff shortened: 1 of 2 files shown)


jfirebaugh:
You should still do the merge from the command line. GitHub merge will rebase and squash the commits, but in this case we want to preserve the release branch commits.

boundsj:
> GitHub merge will rebase and squash

Thanks @jfirebaugh I missed somehow that we've opted out of `Allow merge commits ` but I see we did that. I'll use the command line. Before that, I guess I need to investigate why master iOS bitrise is failing with `Run make ipackage-sim (exit code: 2)`

Issue #4196, opened 2016-03-04
Scheduled nightly Bitrise builds


1ec5:
We should have Bitrise build the iOS SDK and iosapp nightly via a scheduled build and keep the artifacts around. The Android SDK already has nightly snapshots hooked up this way.

Blocked by #3388 and #3275.

/cc @zugaldia @boundsj @tmcw

friedbunny:
👍👍👍👍👍

1ec5:
We can also consider running the static analyzer in this nightly build (reviving #1562), as long as it doesn’t cause builds to time out.

1ec5:
Nightly builds are implemented on Bitrise, but the artifacts currently live on Bitrise. Leaving this issue open until we have a way to upload the artifacts somewhere more discoverable.

PR #4618, opened 2016-04-06
Add script to automate s3 build and publish process

boundsj:
A large part of the iOS SDK deploy process is creating the various binary flavors, uploading them to s3, and then testing that those uploads worked by downloading them -- those binaries are used in the next step of the process when they are attached to the github release.

This script automates that build, upload, and download process and can be run after `mbx auth`.

If this is useful at all, it'll probably be less so after https://github.com/mapbox/mapbox-gl-native/issues/4196. However, I'm making the PR since it won't hurt and might be helpful in the meantime.

cc @1ec5 @friedbunny

### DIFF
--- a/platform/ios/scripts/deploy-packages.sh
+++ b/platform/ios/scripts/deploy-packages.sh
@@ -1,0 +1,84 @@
+#!/usr/bin/env bash
+
+set -e
+set -o pipefail
+set -u
+
+usage() {
+cat <<EOF
+# Usage: sh $0 version target_directory [argument]
+
+# version: The semver version plus optional alpha beta distinction (i.e. {major.minor.patch}{-alpha.N})
+# target_directory: The directory where build output should be placed
+
+# argument:
+#     -g: Upload to github
+
+# environment variables and dependencies: 
+#     - You must run "mbx auth ..." before running
+#     - Set GITHUB_TOKEN to a GitHub API access token in your environment to use GITHUB_RELEASE
+#     - "wget" is required for downloading the zip files from s3
+#     - The "github-release" gem is required to use GITHUB_RELEASE
+EOF
+}
+
+buildPackageStyle() {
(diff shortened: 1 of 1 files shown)


boundsj:
If we decide to go with this (even if only temporarily) I will update https://github.com/mapbox/mapbox-gl-native/wiki/Releasing-the-Mapbox-iOS-SDK to note that this script exists.

1ec5:
Thanks, this looks good!

I’ve also been experimenting with automating the GitHub release process via Bitrise. I think the following invocation would be useful as part of the script:

```bash
GITHUB_USER=mapbox
GITHUB_REPO=mapbox-gl-native

PUBLISH_PRE_FLAG=''
if [[ $( echo ${PUBLISH_VERSION} | awk '/[0-9]-/' ) ]]; then
  PUBLISH_PRE_FLAG='--pre-release'
fi
github-release --verbose release --tag "ios-v${PUBLISH_VERSION}" --name "ios-v${PUBLISH_VERSION}" --draft ${PUBLISH_PRE_FLAG}
```

Then in `buildPackageStyle()`:

```bash
local file_name=mapbox-ios-sdk-${PUBLISH_VERSION}-${style}.zip
github-release --verbose upload --tag "ios-v${PUBLISH_VERSION}" --name ${file_name} --file "$BITRISE_SOURCE_DIR/build/ios/${file_name}"
```

`github-release` requires you to set `GITHUB_TOKEN` to a GitHub API access token in your environment.

boundsj:
Thanks @1ec5 for the great suggestion. The latest commit adds the github piece and some usage output.

1ec5:
:+1:

friedbunny:
Cherry-picked into 3.2.1 as https://github.com/mapbox/mapbox-gl-native/commit/89db01fef0abe258e152eeb2013358c74f8bfeec.

PR #4625, opened 2016-04-07
Convert iosapp to pure Objective-C

1ec5:
Converted MBXViewController from Objective-C++ to Objective-C. In the process, the dependency on default_styles.hpp was replaced by a hardcoded list of style URLs along with runtime assertions that the list of styles is the right length.

/cc @friedbunny @boundsj


boundsj:
:+1:

friedbunny:
Fiddled with the default styles and this fails as you would expect when adding a new style to core. :+1:

PR #4627, opened 2016-04-07
[Android] MapboxEventManager: operator size optimized

vacxe:
Little operator calling optimization

@tobrun @bleege


bleege:
Looks good!  Thanks @vacxe!

PR #4634, opened 2016-04-07
Dispense Telemetry Events If Offline

### LINKED ISSUE #4629: Telemetry events buffer should be capped
We've started noticing `OutOfMemoryError` with the following stack trace:

screen shot 2016-04-07 at 14 07 59

It seems that it happens when device is offline for too long and have collected too much events. Even if it succeedeed, I'm not sure that Mapbox servers would be happy with parsing enormous JSONs. 

I see a couple of options:

 - use capped buffer for events (maybe some flavour of ring buffer); there is a chance to drop some events, but if we crash we drop all of them anyway;
 - send events in chunks; it would solve serialization problem, but  `OutOfMemoryError` could still happen because of excessive buffer growth;
 - flush some part of events to flash, while offline.

**Platform:** Android (but probably applies to iOS as well)
**Mapbox SDK version:** 4.0.0

bleege:
Closes #4629

### DIFF
--- a/platform/android/MapboxGLAndroidSDK/src/main/java/com/mapbox/mapboxsdk/telemetry/MapboxEventManager.java
+++ b/platform/android/MapboxGLAndroidSDK/src/main/java/com/mapbox/mapboxsdk/telemetry/MapboxEventManager.java
@@ -82,7 +82,7 @@ public class MapboxEventManager {
     private long mapboxSessionIdLastSet = 0;
     private static long hourInMillis = 1000 * 60 * 60;
     private static long flushDelayInitialInMillis = 1000 * 10;  // 10 Seconds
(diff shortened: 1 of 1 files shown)


vkurchatkin:
@bleege maybe it makes sense also to call `events.removeAllElements()` in case of failed requests?

bleege:
@vkurchatkin That should still happen regardless of the HTTP response.

vkurchatkin:
@bleege what about `IOException`?

bleege:
@vkurchatkin Ah... I see what you mean.  I just moved it to a `finally` block so that we can be super sure that it gets cleaned up.

bleege:
Rebased and Merged!

Issue #4632, opened 2016-04-07
GeoJSON layers vanishing when rendering a map tile in zoom >= 15


trasch:
I am trying to render a map with some GeoJSON line on it (using the Node.js bindings), here is the example:

https://gist.github.com/trasch/2ad54fd13565da72e4ce9ae6662ca7db

The curious thing is that it works for any zoom level < 15, but the line vanishes in zoom levels >= 15.

### Steps to trigger behavior

Run the script above for zoom 14.9 and 15.

### Expected behavior

There should be a line on the rendered image for both zoom levels.

### Actual behavior

There is a line for zoom 14.9 but not for 15.

**Platform: Mac OS X**
**Mapbox SDK version: 3.0.2**

mikemorris:
Welp, I can confirm the behavior you're reporting in `node-v3.0.2`, but it looks to be fixed in `master` now. Not quite sure what originally caused this, but I'll tag a new release tomorrow.

mikemorris:
Looks like this was fixed in 387f37d035d08a1832ad693e8d9401d7f29857bd

Issue #3678, opened 2016-01-24
Segfault: 11 using node platform


abuchanan:
On Mac OSX 10.11.2
Node v5.0.0 (also tried v4.2.6)

This is the script: https://gist.github.com/abuchanan/5f3c3737469661314975

Error:
```
node js/init.js
PID 62870 received SIGSEGV for address: 0x880
0   segfault-handler.node               0x00000001049f1ecb _ZL16segfault_handleriP9__siginfoPv + 283
1   libsystem_platform.dylib            0x00007fff9b963eaa _sigtramp + 26
2   ???                                 0x0000000500000100 0x0 + 21474836736
3   mapbox-gl-native.node               0x0000000103a0af75 _ZN9node_mbgl15NodeLogObserverD0Ev + 53
4   libsystem_c.dylib                   0x00007fff97e6246b __cxa_finalize_ranges + 345
5   libsystem_c.dylib                   0x00007fff97e6276f exit + 55
6   node                                0x000000010000263b start + 59
7   ???                                 0x0000000000000002 0x0 + 2
```

Directory:
```
tree -I node_modules .
.
├── crash.log
├── fixtures
│   └── tiles
│       └── 0-0-0.vector.pbf
├── image.png
├── js
│   └── init.js
└── package.json
```

I've tried both plain ```npm install mapbox-gl-native``` and downloading/building the code. ```npm test``` passes, but when I pull the test code (from "renders image") out, it does segfault also, that's just not reflected by the test results.

It does write ```image.png``` but it's not an actual PNG and can't be opened.

I'm stumped.

mikemorris:
I think this was an issue introduced in v3.0.0, I'm working on tracking this down.

jfirebaugh:
There's a bug in the node README that presumably got propagated to the script.

```
map.render({}, function(err, image) {
    if (err) throw err;
    fs.writeFileSync('image.png', image);
});
```

This doesn't write a valid PNG file, because `map.render` produces a raw RGBA buffer. You need to encode it to the desired image format yourself.

mikemorris:
Oops, yea, thought this was a different issue. `map.render` used to return a PNG, looks like the examples were never updated to reflect that it returns raw RGBA now.

abuchanan:
Thanks guys. I am able to write an image now (although it took me awhile to find an image library that worked. I ended up with picha)

Still segfaults though.

mikemorris:
[...]

PR #4640, opened 2016-04-07
clarify API documentation

mikemorris:
`map.render()` returns raw pixel buffer and `map.release()` can safely be called earlier. Addresses confusion in https://github.com/mapbox/mapbox-gl-native/issues/3678#issuecomment-174620932 and uses `sharp` encoding example from https://github.com/mapbox/mapbox-gl-native/issues/4632 (thanks @trasch!)

### DIFF
--- a/platform/node/README.md
+++ b/platform/node/README.md
@@ -28,11 +28,29 @@ npm test
 ## Rendering a map tile
 
 ```js
+var mbgl = require('mapbox-gl-native');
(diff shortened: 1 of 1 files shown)


mikemorris:
Merged in https://github.com/mapbox/mapbox-gl-native/commit/f5d66f362272db034a311d2077dbdb2937c9bbdf

Issue #4578, opened 2016-04-02
`run-linux` exits immediately on osx


ansis:
### Steps to trigger behavior

 1. `make linux run-linux`

### Expected behavior

It runs.

### Actual behavior

It exits immediately without error.

`make linux run-linux` is supposed to work on osx, right?

It broke for me in 879382f72405edd1840d2f37744bcba49131859a . It looks like the view's RunLoop exits immediately. @jfirebaugh

jfirebaugh:
I think this is essentially the same issue as #3634. `make linux` should build with linux-appropriate `loop_lib`, `http_lib`, etc -- not the host defaults.

kkaefer:
This only seems to be an issue when building with `loop=darwin`. In particular, it looks like the single timer isn't enough to keep the cfrunloop alive.

kkaefer:
In a similar vein, node.js tests on OS X (when built with Xcode and using `loop=darwin` are failing to complete and hang indefinitely when rendering.

jfirebaugh:
Normally a timer is enough to keep a CFRunLoop alive. We have existing unit tests that verify that. The issue with the GLFW app seems to be specifically triggered by the use of `glfwCreateWindow` -- it must do something internally to the main run loop that triggers this behavior.

jfirebaugh:
I set a breakpoint on `CFRunLoopStop` and this is what I see:

```
#0	0x00007fff8e81c050 in CFRunLoopStop ()
#1	0x0000000100515547 in _glfwPlatformPostEmptyEvent at /Users/john/Development/mapbox-gl-native/mason_packages/.build/glfw-3.1.2/src/cocoa_window.m:1142
#2	0x0000000100515429 in -[GLFWApplicationDelegate applicationDidFinishLaunching:] at /Users/john/Development/mapbox-gl-native/mason_packages/.build/glfw-3.1.2/src/cocoa_window.m:287
#3	0x00007fff8e8a770c in __CFNOTIFICATIONCENTER_IS_CALLING_OUT_TO_AN_OBSERVER__ ()
#4	0x00007fff8e8a767f in ___CFXRegistrationPost_block_invoke ()
#5	0x00007fff8e8a6d47 in _CFXRegistrationPost ()
#6	0x00007fff8e8a6ab2 in ___CFXNotificationPost_block_invoke ()
#7	0x00007fff8e8a0d42 in -[_CFXNotificationRegistrar find:object:observer:enumerator:] ()
[...]

PR #4651, opened 2016-04-08
[build] Introduce optional FORCE_LIBUV for OS X

brunoabinader:
Following up https://github.com/mapbox/mapbox-gl-native/pull/4608#issuecomment-207162791 - GLFW test app using CoreFoundation's `CFRunLoop` is currently broken (see https://github.com/mapbox/mapbox-gl-native/issues/4578 for details). Our current working fallback is the `libuv` implementation. This patch re-enables it when passing an optional `FORCE_LIBUV=1` envvar.

Example usage in OS X:
```$ FORCE_LIBUV=1 make glfw-app```

/cc @jfirebaugh @kkaefer @tmpsantos


mikemorris:
Can this just be `LOOP=uv` (like https://github.com/mapbox/mapbox-gl-native/commit/d46f24cdf8ccc577beec0f1a75e64f6599f25875) instead of a significantly different syntax?

brunoabinader:
> Can this just be LOOP=uv (like d46f24c) instead of a significantly different syntax?

It can be done, I avoided it because cases like e.g. `LOOP=darwin` would have no effect. I'm happy to change it to the previous syntax otherwise.

jfirebaugh:
How about we just fix #4578?

brunoabinader:
> How about we just fix #4578?

Agreed - this is meant only as a workaround until we figure out #4578 👍

kkaefer:
Fixing #4578 is important, but I'd love to see a patch like this as well because it makes debugging libuv issues way easier.

jfirebaugh:
When would you need to debug libuv issues nowadays? The node bindings are the only real platform where it's used now, and those always built with libuv, even on OS X. So you can debug normally without any special flags.

jfirebaugh:
Closing; merged #4666 instead.

Issue #4655, opened 2016-04-08
what is the intended constrain behaviour?


ansis:
https://github.com/mapbox/mapbox-gl-native/blob/35b579fde196ec3dd2a487ab5d528c8662571a43/src/mbgl/map/transform_state.cpp#L297-L299

``` cpp
scale_ = util::max(scale_,
                       static_cast<double>((rotatedNorth() ? height : width) / util::tileSize),
                       static_cast<double>((rotatedNorth() ? width : height) / util::tileSize));
```

looks like it's equivalent to

``` cpp
scale_ = util::max(scale_,
                       static_cast<double>(width / util::tileSize),
                       static_cast<double>(height / util::tileSize));
```

@1ec5 Is this the intended behaviour?

It's constraining based on the width even if the constrain mode is `HeightOnly`.

1ec5:
Yes, I made this change in #3401, so that the only difference between the two modes is that WidthAndHeight prevents you from seeing both sides of the antimeridian at a time, while HeightOnly allows scrolling seamlessly across it. The current HeightOnly behavior matches other SDKs and avoids the situation where the same coordinate appears twice on screen. If there is a need for tiling the world multiple times, we should add another constraining mode.

1ec5:
This part of #3401 was motivated by a bug in the OS X SDK, but in principle you would've seen the same thing in the iOS SDK on an iPad:

> Before this change, osxapp effectively constrained the zoom level based on whether the window was shorter or taller than 512 points.

I suppose I could've fixed that issue by turning off all constraints, but we don't show anything beyond certain latitudes.

PR #4656, opened 2016-04-08
implement queryRenderedFeatures

ansis:
`queryRenderedFeatures` is mostly implemented. 48/50 query tests are passing. Two are skipped because different [constrain behaviour](https://github.com/mapbox/mapbox-gl-native/issues/4655) is causing errors.

#### Questions
- What should `queryRenderedFeatures` argument signatures be? [js api here](https://www.mapbox.com/mapbox-gl-js/api/#Map.queryRenderedFeatures)
- What should the return type be? This will use geometry.hpp, right?
- Should we support filters? Can [`FilterExpression`](https://github.com/mapbox/mapbox-gl-native/blob/master/src/mbgl/style/filter_expression.hpp) be made public? 

**If any of the above are not quick to decide or quick to implement, can we punt and merge this as is?** The current external api is not something we want to have people use but it would be good to have this in master with tests running on ci. It's already a huge pull request.

`querySourceFeatures` is not implemented because it depends heavily on the return type question. It can be done in a separate pr once that's answered.

---
@jfirebaugh can you review or assign reviewing to someone else?

---

my other remaining tasks are:
- [x] fix this build error. Any ideas? https://www.bitrise.io/build/108bf7df608dc23c
- [x] track down this segfault https://www.bitrise.io/build/e63b43e20310d3e0
- [x] implement querying for symbols with `-ignore-placement: true`

### DIFF
--- a/include/mbgl/map/map.hpp
+++ b/include/mbgl/map/map.hpp
@@ -158,6 +158,10 @@ public:
                         const char* before = nullptr);
     void removeCustomLayer(const std::string& id);
 
+    // Feature queries
+    std::vector<std::string> queryRenderedFeatures(const ScreenCoordinate&, const optional<std::vector<std::string>>& layerIDs = {});
+    std::vector<std::string> queryRenderedFeatures(const std::array<ScreenCoordinate, 2>&, const optional<std::vector<std::string>>& layerIDs = {});
+
     // Memory
     void setSourceTileCacheSize(size_t);
     void onLowMemory();
--- a/src/mbgl/geometry/feature_index.cpp
+++ b/src/mbgl/geometry/feature_index.cpp
@@ -1,0 +1,236 @@
+#include <mbgl/geometry/feature_index.hpp>
+#include <mbgl/util/math.hpp>
+#include <mbgl/style/style.hpp>
+#include <mbgl/style/style_layer.hpp>
+#include <mbgl/layer/symbol_layer.hpp>
+#include <mbgl/util/get_geometries.hpp>
+#include <mbgl/text/collision_tile.hpp>
+#include <mbgl/util/rapidjson.hpp>
+#include <rapidjson/writer.h>
+
+#include <cassert>
+#include <string>
+
+using namespace mbgl;
+
+FeatureIndex::FeatureIndex() {}
+
+void FeatureIndex::insert(const GeometryCollection& geometries, std::size_t index,
+        const std::string& sourceLayerName, const std::string& bucketName) {
+
+    auto sortIndex = treeBoxes.size();
+
+    for (auto& ring : geometries) {
+
+        float minX = std::numeric_limits<float>::infinity();
+        float minY = std::numeric_limits<float>::infinity();
+        float maxX = -std::numeric_limits<float>::infinity();
+        float maxY = -std::numeric_limits<float>::infinity();
+        for (auto& p : ring) {
+            const float x = p.x;
+            const float y = p.y;
+            minX = util::min(minX, x);
+            minY = util::min(minY, y);
+            maxX = util::max(maxX, x);
+            maxY = util::max(maxY, y);
+        }
+
+        treeBoxes.emplace_back(
+            TreeBox {
+                TreePoint { minX, minY },
+                TreePoint { maxX, maxY }
+            },
+            IndexedSubfeature { index, sourceLayerName, bucketName, sortIndex }
+        );
+    }
+}
+
+void FeatureIndex::loadTree() {
+    tree.insert(treeBoxes.begin(), treeBoxes.end());
+}
+
+bool vectorContains(const std::vector<std::string>& vector, const std::string& s) {
+    return std::find(vector.begin(), vector.end(), s) != vector.end();
+}
+
+bool vectorsIntersect(const std::vector<std::string>& vectorA, const std::vector<std::string>& vectorB) {
+    for (auto& a : vectorA) {
+        if (vectorContains(vectorB, a)) return true;;
+    }
+    return false;
+}
+
+
+bool topDown(const FeatureTreeBox& a, const FeatureTreeBox& b) {
+    return std::get<1>(a).sortIndex > std::get<1>(b).sortIndex;
+}
+
+bool topDownSymbols(const IndexedSubfeature& a, const IndexedSubfeature& b) {
+    return a.sortIndex < b.sortIndex;
+}
+
+void FeatureIndex::query(
+        std::unordered_map<std::string, std::vector<std::string>>& result,
+        const GeometryCollection& queryGeometry,
+        const float bearing,
+        const double tileSize,
+        const double scale,
+        const optional<std::vector<std::string>>& filterLayerIDs,
+        const GeometryTile& geometryTile,
+        const Style& style) const {
+
+    const float pixelsToTileUnits = util::EXTENT / tileSize / scale;
+
+    float additionalRadius = style.getQueryRadius() * pixelsToTileUnits;
+
+    float minX = std::numeric_limits<float>::infinity();
+    float minY = std::numeric_limits<float>::infinity();
+    float maxX = -std::numeric_limits<float>::infinity();
+    float maxY = -std::numeric_limits<float>::infinity();
+
+    for (auto& ring : queryGeometry) {
+        for (auto& p : ring) {
+            minX = util::min<float>(minX, p.x);
+            minY = util::min<float>(minY, p.y);
+            maxX = util::max<float>(maxX, p.x);
+            maxY = util::max<float>(maxY, p.y);
+        }
+    }
+
+    TreeBox queryBox = {
+        TreePoint { minX - additionalRadius, minY - additionalRadius },
+        TreePoint { maxX + additionalRadius, maxY + additionalRadius }
+    };
+
+    // query circle, line, fill features
+    std::vector<FeatureTreeBox> matchingBoxes;
+    tree.query(bgi::intersects(queryBox), std::back_inserter(matchingBoxes));
+    std::sort(matchingBoxes.begin(), matchingBoxes.end(), topDown);
+
+    size_t previousSortIndex = std::numeric_limits<size_t>::max();
+    for (auto& matchingBox : matchingBoxes) {
+        auto& indexedFeature = std::get<1>(matchingBox);
+
+        // If this feature is the same as the previous feature, skip it.
+        if (indexedFeature.sortIndex == previousSortIndex) continue;
+        previousSortIndex = indexedFeature.sortIndex;
+
+        addFeature(result, indexedFeature, queryGeometry, filterLayerIDs, geometryTile, style, bearing, pixelsToTileUnits);
+    }
+
+    // query symbol features
+    assert(collisionTile);
+    std::vector<IndexedSubfeature> symbolFeatures = collisionTile->queryRenderedSymbols(minX, minY, maxX, maxY, scale);
+    std::sort(symbolFeatures.begin(), symbolFeatures.end(), topDownSymbols);
+    for (auto& symbolFeature : symbolFeatures) {
+        addFeature(result, symbolFeature, queryGeometry, filterLayerIDs, geometryTile, style, bearing, pixelsToTileUnits);
+    }
+}
+
+void FeatureIndex::addFeature(
+    std::unordered_map<std::string, std::vector<std::string>>& result,
+    const IndexedSubfeature& indexedFeature,
+    const GeometryCollection& queryGeometry,
+    const optional<std::vector<std::string>>& filterLayerIDs,
+    const GeometryTile& geometryTile,
+    const Style& style,
+    const float bearing,
+    const float pixelsToTileUnits) const {
+
+    auto& layerIDs = bucketLayerIDs.at(indexedFeature.bucketName);
+
+    if (filterLayerIDs && !vectorsIntersect(layerIDs, *filterLayerIDs)) return;
+
+    auto sourceLayer = geometryTile.getLayer(indexedFeature.sourceLayerName);
+    assert(sourceLayer);
+    auto feature = sourceLayer->getFeature(indexedFeature.index);
+    assert(feature);
+
+    for (auto& layerID : layerIDs) {
+
+        if (filterLayerIDs && !vectorContains(*filterLayerIDs, layerID)) continue;
+
+        auto styleLayer = style.getLayer(layerID);
+        if (!styleLayer) continue;
+
+        if (!styleLayer->is<SymbolLayer>()) {
+            auto geometries = getGeometries(*feature);
+            if (!styleLayer->queryIntersectsGeometry(queryGeometry, geometries, bearing, pixelsToTileUnits)) continue;
+        }
+
+        auto& layerResult = result[layerID];
+
+        auto properties = feature->getProperties();
+        rapidjson::StringBuffer buffer;
+        buffer.Clear();
+        rapidjson::Writer<rapidjson::StringBuffer> writer(buffer);
+
+        writer.StartObject();
+        writer.Key("type");
+        writer.String("Feature");
+        auto id = feature->getID();
+        if (id) {
+            writer.Key("id");
+            writer.Double(feature->getID());
+        }
+        writer.Key("properties");
+        writer.StartObject();
+        for (auto& prop : properties) {
+            std::string key = prop.first;
+            Value& value = prop.second;
+
+            writer.Key(key.c_str());
+
+            if (value.is<std::string>()) {
+                writer.String(value.get<std::string>().c_str());
+            } else if (value.is<bool>()) {
+                writer.Bool(value.get<bool>());
+            } else if (value.is<int64_t>()) {
+                writer.Int64(value.get<int64_t>());
+            } else if (value.is<uint64_t>()) {
+                writer.Uint64(value.get<uint64_t>());
+            } else if (value.is<double>()) {
+                writer.Double(value.get<double>());
+            }
+        }
+        writer.EndObject();
+        writer.EndObject();
+
+        layerResult.push_back(buffer.GetString());
+    }
+}
+
+optional<GeometryCollection> FeatureIndex::translateQueryGeometry(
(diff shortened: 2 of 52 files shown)


mb12:
@ansis  Can you please kindly answer the following?

1.) It looks like all geometries are being inserted in the index by default (irrespective of whether or not the application uses query feature or not)? Is that correct (or not) and intentional? Can tree construction (loadTree call ) for a tile be postponed until the application actually uses query feature?  

2.) In 'Source:: Source::queryRenderedFeatures' , under what scenarios would 'tileQueries' have more than one element? We can reduce the number of rtree lookups by picking only one tile per Source. 

3.) Is it possible to explain what geometries are actually inserted in the index for symbols(esp. with text)?

ansis:
Thanks for digging in @mb12 

> 1.) It looks like all geometries are being inserted in the index by default (irrespective of whether or not the application uses query feature or not)? Is that correct (or not) and intentional? Can tree construction (loadTree call ) for a tile be postponed until the application actually uses query feature?

Yes, all geometries that are rendered are inserted into the index. It is partially intentional. I need to profile to see if index creation time is significant. One of advantage of doing it at tile load time is that it doesn't block the main thread. On the mapbox-gl-js side we use a different kind of index where insertions are really cheap: https://github.com/mapbox/grid-index

> 2.) In 'Source:: Source::queryRenderedFeatures' , under what scenarios would 'tileQueries' have more than one element? 

When the first argument of `queryRenderedFeatures` is a box instead of a point

> We can reduce the number of rtree lookups by picking only one tile per Source.

This is already done [here](https://github.com/mapbox/mapbox-gl-native/blob/eb20d195fbbdb1a25a6c5fc59f5fb0ea293a3b65/src/mbgl/source/source.cpp#L526). Tiles that don't intersect the query box are skipped.

> 3.) Is it possible to explain what geometries are actually inserted in the index for symbols(esp. with text)?

Symbol querying uses the index that was already created during the symbol placement process. Nothing is inserted into the main FeatureIndex tree. The CollisionTile tree holds a bunch of axis-aligned boxes that are anchored around points. You can render these boxes with `MapDebugOptions::Collision`.

include/mbgl/map/map.hpp
```diff
@@ -174,6 +174,10 @@ class Map : private util::noncopyable {
                         const char* before = nullptr);
     void removeCustomLayer(const std::string& id);
 
+    // Feature queries
+    std::vector<std::string> queryRenderedFeatures(const ScreenCoordinate&, const optional<std::vector<std::string>>& layerIDs = {});
+    std::vector<std::string> queryRenderedFeatures(const std::vector<ScreenCoordinate>&, const optional<std::vector<std::string>>& layerIDs = {});
```
jfirebaugh: `const std::vector<ScreenCoordinate>&` ⇢ `const std::array<ScreenCoordinate, 2>&`

jfirebaugh: How about the following single-method interface:

```
    using QueryGeometry = mapbox::util::variant<ScreenCoordinate, std::array<ScreenCoordinate, 2>>;
    std::vector<std::string> queryRenderedFeatures(const QueryGeometry&, const optional<std::vector<std::string>>& layerIDs = {});
```

src/mbgl/map/map_context.cpp
```diff
@@ -306,6 +307,38 @@ void MapContext::setClasses(const std::vector<std::string>& classNames, const Pr
...
+    for (auto& p : queryPoints) {
+        queryGeometry.push_back(TileCoordinate::fromScreenCoordinate(transformState, 0, { p.x, transformState.getHeight() - p.y }));
+    }
+    return queryGeometry;
+}
+
+std::vector<std::string> MapContext::queryRenderedFeatures(const ScreenCoordinate& point, const optional<std::vector<std::string>>& layerIDs) {
+    auto queryGeometry = pointsToCoordinates({ point }, transformState);
+
+    if (!style) return {};
```
jfirebaugh: Move up to first statement.

src/mbgl/map/map_context.cpp
```diff
@@ -306,6 +307,38 @@ void MapContext::setClasses(const std::vector<std::string>& classNames, const Pr
...
+        { box[0].x, box[0].y },
+        { box[1].x, box[0].y },
+        { box[1].x, box[1].y },
+        { box[0].x, box[1].y },
+        { box[0].x, box[0].y }
+    };
+
+    auto queryGeometry = pointsToCoordinates(queryPoints, transformState);
+
+    if (!style) return {};
```
jfirebaugh: Move up to first statement.

include/mbgl/map/map.hpp
jfirebaugh: Nah, never mind, overloading is better.

src/mbgl/style/style.cpp
```diff
@@ -328,6 +328,43 @@ RenderData Style::getRenderData() const {
     return result;
 }
 
+std::vector<std::string> Style::queryRenderedFeatures(
+        const std::vector<TileCoordinate>& queryGeometry,
+        const double zoom,
+        const double bearing,
+        const optional<std::vector<std::string>>& layerIDs) {
+    std::vector<std::unordered_map<std::string, std::vector<std::string>>> sourceResults;
```
jfirebaugh: Can this method be implemented like this?

```
    std::vector<std::string> features;

    for (const auto& source : sources) {
        for (auto& result : source->queryRenderedFeatures(queryGeometry, zoom, bearing, layerIDs)) {
            features.emplace_back(std::move(result));
        }
    }

    return features;
```

src/mbgl/source/source.cpp
```diff
@@ -473,6 +474,87 @@ void Source::updateTilePtrs() {
...
+    std::unordered_map<uint64_t, TileQuery> tileQueries;
+
+    for (auto& tilePtr : tilePtrs) {
+        auto& tile = *tilePtr;
+        uint64_t integerID = tile.id.to_uint64();
+
+        auto tileSpaceBoundsMin = coordinateToTilePoint(tile.id, { minX, minY, z });
+        auto tileSpaceBoundsMax = coordinateToTilePoint(tile.id, { maxX, maxY, z });
+
+        if (tileSpaceBoundsMin.x < util::EXTENT && tileSpaceBoundsMin.y < util::EXTENT &&
```
jfirebaugh: Reverse the condition and use `continue`; unindent the happy path that follows.

src/mbgl/source/source.cpp
```diff
@@ -473,6 +474,87 @@ void Source::updateTilePtrs() {
...
+    for (auto& c : queryGeometry) {
+        minX = util::min(minX, c.x);
+        minY = util::min(minY, c.y);
+        maxX = util::max(maxX, c.x);
+        maxY = util::max(maxY, c.y);
+    }
+
+    std::unordered_map<uint64_t, TileQuery> tileQueries;
+
+    for (auto& tilePtr : tilePtrs) {
```
jfirebaugh: Do we need two loops and temporary here? Or can we use a single loop?

src/mbgl/util/tile_coordinate.hpp
```diff
@@ -0,0 +1,39 @@
...
+#include <mbgl/style/types.hpp>
+#include <mbgl/map/transform_state.hpp>
+
+
+namespace mbgl {
+
+class TransformState;
+
+// Has floating point x/y coordinates.
+// Used for computing the tiles that need to be visible in the viewport.
```
jfirebaugh: I still don't really understand why this class exists, what exactly it represents, or what makes it distinct from the other coordinate classes. Can you flesh out this comment more?

src/mbgl/tile/tile_data.hpp
```diff
@@ -82,6 +84,16 @@ class TileData : private util::noncopyable {
...
     virtual void redoPlacement(const std::function<void()>&) {}
 
+    virtual void queryRenderedFeatures(
+            std::unordered_map<std::string, std::vector<std::string>>&,
+            const GeometryCollection&,
+            const double,
+            const double,
+            const double,
+            const optional<std::vector<std::string>>&) {
+        return;
```
jfirebaugh: Omit `return;`.

src/mbgl/tile/tile_data.hpp
```diff
@@ -82,6 +84,16 @@ class TileData : private util::noncopyable {
     virtual void redoPlacement(PlacementConfig, const std::function<void()>&) {}
     virtual void redoPlacement(const std::function<void()>&) {}
 
+    virtual void queryRenderedFeatures(
+            std::unordered_map<std::string, std::vector<std::string>>&,
```
jfirebaugh: Unless their purpose is obvious from the type, provide names for parameter declarations. This applies to all the parameters here.

src/mbgl/tile/tile_worker.hpp
```diff
@@ -32,6 +33,8 @@ class TileParseResultBuckets {
 public:
     TileData::State state = TileData::State::invalid;
     std::unordered_map<std::string, std::unique_ptr<Bucket>> buckets;
+    std::unique_ptr<FeatureIndex> featureIndex;
```
jfirebaugh: The additional members make `TileParseResultBuckets` (even more of) a misnomer. Probably what we need here is a radical refactor so that parsing a vector tile _returns_ `VectorTileData`. But in the absence of that, can we rename `TileParseResultBuckets` to `TileParseSuccess` or something?

src/mbgl/tile/vector_tile_data.cpp
```diff
@@ -77,6 +79,11 @@ VectorTileData::VectorTileData(const TileID& id_,
                 // existing buckets in case we got a refresh parse.
                 buckets = std::move(resultBuckets.buckets);
 
+                if (state == State::parsed) {
```
jfirebaugh: In what situations does `parseGeometryTile` give you `TileParseResult` whose state isn't `parsed`?

include/mbgl/map/map.hpp
ansis: done d13f7b96423ac5fa78fb6e1996a9c9e32c47c239

src/mbgl/map/map_context.cpp
ansis: done d13f7b96423ac5fa78fb6e1996a9c9e32c47c239

src/mbgl/style/style.cpp
ansis: Nope. Features need to be returned in order based on the style order. Since layers from a source aren't all consecutive two steps are needed to get this sorting right.

added a short comment: 0d180639f92e3fc739472f33315b851b405ecd94

src/mbgl/source/source.cpp
ansis: done a0eb0141631763a00dde2b7492f560b11ac03725

src/mbgl/source/source.cpp
ansis: The two loops and `TileQuery` objects are needed for queries that cross the international date line. These queries that cross the date line can intersect a tile in multiple spots. TileQuery` objects are used to collect these separate query geometry parts into a single multi-geometry query. A single multi-geometry query let's us avoid returning duplicates without having a separate dedupe step later.

src/mbgl/tile/vector_tile.cpp
```diff
@@ -84,6 +84,27 @@ optional<Value> VectorTileFeature::getValue(const std::string& key) const {
     return optional<Value>();
 }
 
+std::unordered_map<std::string,Value> VectorTileFeature::getProperties() const {
+    std::unordered_map<std::string,Value> properties;
+    pbf tags = tags_pbf;
+    while (tags) {
+        uint32_t tag_key = tags.varint();
+        uint32_t tag_val = tags.varint();
+        // this shouldn't use a linear searach of a map
```
jfirebaugh: We could make `VectorTileLayer::keys` a `std::vector<std::string>`, and then make an auxiliary `std::map<std::string, uint32_t>` for use where `keys` is currently used. Ideally the latter would be an `unordered_map`, and the keys would be references into the vector rather than copies of the same strings.

jfirebaugh:
From what I understand of the abstract algorithm so far, it seems like the simplest natural expression is (pseudocode):

```
    // do whatever computation is needed for dateline logic, tile cover, etc.
    auto queryRegion = computeQueryRegion(input.query);
    std::vector<result> results;

    for (layer in style.layers) {
        if (layer in input.layers) {
            results.concat(layer.source.queryRenderedFeatures(layer, queryRegion, ...));
        }
    }

    return results;
```

What fundamentals am I missing, or what would it take to be able to implement it this way?

jfirebaugh:
As context, I would find it very appealing if this could be implemented as an [online algorithm](https://en.wikipedia.org/wiki/Online_algorithm). Not so much for efficiency's sake, just that they tend in the end to be the simplest expression of a lot of algorithms.

ansis:
> What fundamentals am I missing, or what would it take to be able to implement it this way?

All features are currently indexed in a single index. For each layer's query you'd be querying the index with the same region again and again and would need to filter out features from all other layers. I haven't tried, but I'm guessing this would be slower. It might be fine though since reading features from the pbf and doing geometry intersections is the expensive part.

Why not index layers individually? They were indexed in a group index earlier and I just stuck with that. I guessed it could be cheaper for queries that cover all layers (the default and worse case) and it could have a lower memory footprint for the grid index (which was written so that we could transfer it). Now that I think about it, it probably could be fine. Symbols use the existing collision index though, which would need to be split up into separate indexes once parsing is done.

I agree that an online algorithm would be simpler. Do you think it's worth the effort to try to switch?

src/mbgl/tile/tile_data.hpp
ansis: They weren't named to avoid unused variable warnings.
Is this better? ea80c70bca770fe9901403a1d32694cfb3e958b3
or should I move that to raster_tile_data.cpp?

src/mbgl/tile/tile_worker.hpp
ansis: is `TileParseResultData` ok? 3b251aa `Data` doesn't really mean anything I'm not seeing how `Success` is a good fit either

src/mbgl/tile/vector_tile_data.cpp
ansis: I'm not 100% sure I understand how partial tile parsing works, but:
- `parseGeometryTile` [calls](https://github.com/mapbox/mapbox-gl-native/blob/7fa32a0d083b9c828d2e487f547d55967bd43817/src/mbgl/util/worker.cpp#L37) `parseAllLayers`
- `parseAllLayers` only returns `parsed` if it [manages to parse all layers](https://github.com/mapbox/mapbox-gl-native/blob/7fa32a0d083b9c828d2e487f547d55967bd43817/src/mbgl/tile/tile_worker.cpp#L62) which I think happens when it doesn't need to load any glyphs or icons

src/mbgl/tile/vector_tile.cpp
ansis: done in 8bfd1f62f5969de3030cea4e761522da6a6113d2

I stored the original strings in the map and references in the vector because reference_wrapper was causing problems with unordered_map's key requirements. It might be possible to get it to work but this was easier and should work just as well.

src/mbgl/geometry/feature_index.cpp
```diff
@@ -0,0 +1,230 @@
...
+                TreePoint { minX, minY },
+                TreePoint { maxX, maxY }
+            },
+            IndexedSubfeature { index, sourceLayerName, bucketName, sortIndex }
+        );
+    }
+}
+
+void FeatureIndex::loadTree() {
+    tree.insert(treeBoxes.begin(), treeBoxes.end());
```
mb12: Should this loop be made cancellable/interruptible (or not)?
See the comment here:

https://github.com/mapbox/mapbox-gl-native/issues/2644#issuecomment-210258384

src/mbgl/util/tile_coordinate.hpp
ansis: It's main purpose is represent euclidean world coordinates and help convert them to tile coordinates at different zoom levels. It does this in a weird way though, since it also has a z parameter, which gives it an extra dimension.

It should probably be replaced by a zoom-less world coordinate type with transform methods that convert it to tile units or tile ids.

It was removed in the coordinate cleanup pull request but those changes never made their way to -js. -js uses this class for a couple of things:
- for calculating tile cover
- for helping convert lat/lng to tile coordinates
- in Transform.setLocationAtPoint which is used by easeTo

This class could be removed now and a latlng -> tile units method could be added but I'd rather do this as part of a larger Transform cleanup and synchronization between -js and -native.

Do you see this as a blocker for this pr or as something that can be cleaned up later?

ansis:
@jfirebaugh I think all your notes are addressed except https://github.com/mapbox/mapbox-gl-native/pull/4656#discussion_r59475796 which I think we should leave for later. Can you review the fixes?

jfirebaugh:
Can you push the rebased / squashed version?

ansis:
squashed version pushed

jfirebaugh:
👍  Let's merge once CI is passing.

ansis:
CI is passing.

The rtree is really slow when built in Debug mode. Creating the index adds a couple *hundred* ms. In Release mode it takes 10-15ms. Is this ok, or is this something that needs to be optimized before merging?

jfirebaugh:
That's somewhat worrisome. Besides being an annoyance for testing a debug build, it's often the case that slow debug performance on a desktop will translate to slow release performance on a mobile device. Can you test this and optimize accordingly?

mb12:
@ansis and @jfirebaugh On mobile devices with use model similar to (Google Maps, Nokia Here, maps.me or Apple Maps) linear search through the geometries should be good enough( query is asynchronous and/or number of queries is too low because of long touch/tap use model).

This is what the use model looks likes:

- Long touch or tap (Drops a pin or draw a circle around point of query)
- Panel slides up from bottom or the side
- The panel gets asynchronously populated

ansis:
@jfirebaugh I ported GridIndex from -js: daab075
This should be fast enough now. Can you review that commit?

@mb12 thanks for the suggestion

PR #4666, opened 2016-04-11
[glfw] Workaround for OS X

### LINKED ISSUE #4578: `run-linux` exits immediately on osx

brunoabinader:
Based on https://github.com/mapbox/mapbox-gl-native/issues/4578#issuecomment-207586376 - this is a proposed workaround I've been using in OS X. 

Fixes #4578.

/cc @jfirebaugh @kkaefer

### DIFF
--- a/platform/default/glfw_view.cpp
+++ b/platform/default/glfw_view.cpp
@@ -448,7 +448,11 @@ void GLFWView::run() {
     };
 
     frameTick.start(mbgl::Duration::zero(), mbgl::Milliseconds(1000 / 60), callback);
+#if defined(__APPLE__)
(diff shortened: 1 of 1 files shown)


brunoabinader:
FIY I tested this solution with no noticeable performance decrease due to the busy wait. When asked to quit, the app also does so without memory leaks or crashes.

kkaefer:
Also see #4669

PR #4674, opened 2016-04-11
[ios] Restart user location heading updates on validation

### LINKED ISSUE #4668: Mapbox loses CoreLocation Heading Updates when coming back from background
**Platform:** iOS 9.3.1 (13E238)
**Mapbox SDK version:** 3.1.0
### Abstract
   While Mapbox is in a `MGLUserTrackingModeFollowWithHeading` tracking mode, backgrounding the app causes the app to lose heading updates.
### Steps to trigger behavior

 1. Build, Install, and run the Mapbox Test app.
 2. Set tracking mode to `MGLUserTrackingModeFollowWithHeading`.
 3. Background the App.
 4. Foreground the App.

### Expected behavior
Expecting to see Mapbox restored to its previous state before backgrounding.
### Actual behavior
Although the tracking mode doesn't change, the map no longer updates its orientation to match the orientation of the phone. Re-toggling the tracking mode fixes the problem.
### Suggested Fix
in `MGLMapView.mm`, Add the following to `wakeGL:` (or do what I did and make a new `handleWake:` method that calls `wakeGL:` and includes the following code:)

```if (self.userTrackingMode == MGLUserTrackingModeFollowWithHeading) {
        [self.locationManager startUpdatingHeading];
    }```

friedbunny:
User location heading updates would not automatically restart when an app became active again.

Fixes #4668. Thanks to @JThramer for reporting.

/cc @1ec5

### DIFF
--- a/platform/ios/CHANGELOG.md
+++ b/platform/ios/CHANGELOG.md
@@ -4,6 +4,8 @@ Mapbox welcomes participation and contributions from everyone.  If you’d like 
 
 ## master
 
+- User location heading updates now resume properly when an app becomes active again. ([#4674](https://github.com/mapbox/mapbox-gl-native/pull/4674))
+
 ## 3.2.0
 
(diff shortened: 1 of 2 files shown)


1ec5:
:+1:

friedbunny:
Added the first entry to the post v3.2.0 changelog.

friedbunny:
Cherry-picked into 3.2.1 as https://github.com/mapbox/mapbox-gl-native/commit/023be536d5f61f1404a85fe394b1e9435cb456e9.

PR #4676, opened 2016-04-11
[ios, osx] The return of xcpretty, part II

jfirebaugh:
Bring back xcpretty. We've encountered issues with it in the past, but I'm willing to go back to it because the alternative -- immensely verbose build logs in which it is virtually impossible to find error output -- is worse.


jfirebaugh:
@1ec5 @friedbunny Thoughts?

1ec5:
:+1: Raw xcodebuild output is sometimes useful, but it has also been obscuring some warnings like the one I finally noticed while working on #4678 and fixed in 1b09898e03aad3029bb276bc3fd429a102348c35.

PR #4677, opened 2016-04-11
[core] Fix race condition that could lead to a UNIQUE constraint failure

### LINKED ISSUE #4422: OfflineDatabase.ConcurrentUse: UNIQUE constraint failed: resources.url
https://travis-ci.org/mapbox/mapbox-gl-native/jobs/117598855

```
[ RUN      ] OfflineDatabase.ConcurrentUse
terminate called after throwing an instance of 'mapbox::sqlite::Exception'
  what():  UNIQUE constraint failed: resources.url
```

This indicates both threads ran `UPDATE resources` and saw that the row didn't exist, then both tried to run the `INSERT INTO resources` statement. I had thought this would be impossible because only one thread could get the exclusive lock to run the `UPDATE`, and then that exclusive lock would be held for the remainder of `OfflineDatabase::putResource` because the `Statement` remains in scope. Apparently not.

jfirebaugh:
Fixes #4422

### DIFF
--- a/platform/default/mbgl/storage/offline_database.cpp
+++ b/platform/default/mbgl/storage/offline_database.cpp
@@ -239,6 +239,10 @@ bool OfflineDatabase::putResource(const Resource& resource,
 
     // We can't use REPLACE because it would change the id value.
 
+    // Begin an immediate-mode transaction to ensure that two writers do not attempt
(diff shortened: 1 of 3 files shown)


jfirebaugh:
👀  @kkaefer

Issue #4665, opened 2016-04-11
API.Offline is flaky on Bitrise OS X


brunoabinader:
```
[ RUN      ] API.Offline
make: *** [test-osx] Segmentation fault: 11
```

See https://www.bitrise.io/build/9c38311ed1453d25 for full log.

👀  @jfirebaugh @tmpsantos

brunoabinader:
Caught this also:
```
[ RUN      ] OfflineDatabase.ConcurrentUse
libc++abi.dylib: terminating with uncaught exception of type mapbox::sqlite::Exception: UNIQUE constraint failed: resources.url
make: *** [test-osx] Abort trap: 6
```
Ref: https://www.bitrise.io/build/8731103bc21b6812

brunoabinader:
One more:
```
[ RUN      ] AsyncTask.ThreadSafety
/Users/vagrant/git/test/util/async_task.cpp:125: Failure
Expected: (count) > (1), actual: 1 vs 1
[  FAILED  ] AsyncTask.ThreadSafety (47 ms)```
Ref: https://www.bitrise.io/build/0a43b6b55c4a457c

jfirebaugh:
> make: *** [test-osx] Segmentation fault: 11

For this one, we need to get a stack trace. For linux, we do this via running under scripted gdb. @ansis was working on something similar for OS X / lldb.

> libc++abi.dylib: terminating with uncaught exception of type mapbox::sqlite::Exception: UNIQUE constraint failed: resources.url

This one is #4422.

> /Users/vagrant/git/test/util/async_task.cpp:125: Failure
> Expected: (count) > (1), actual: 1 vs 1

[Famous last words](https://github.com/mapbox/mapbox-gl-native/blob/5471122079b9e882917c148fa443024d3918ddec/test/util/async_task.cpp#L123-L124)

jfirebaugh:
Looks like the lldb that comes with Xcode does not support `-batch` or `--batch` (if it does, I can't find any documentation -- it's not listed, e.g. [in the man page](https://developer.apple.com/library/mac/documentation/Darwin/Reference/ManPages/man1/lldb.1.html)).

Another approach would be to `cat` the most recently created file in `~/Library/Logs/DiagnosticReports/` after the crash occurs. @1ec5 does this sound reasonable or is there a better way?

1ec5:
LLDB is scripted via Python: http://lldb.llvm.org/scripting.html

[...]

PR #4679, opened 2016-04-11
[test] This "very unlikely" condition in fact happens regularly in CI

jfirebaugh:
Refs #4665


tmpsantos:
I was hoping for it. :trollface:

PR #4678, opened 2016-04-12
Speed up iOS CI build

jfirebaugh:
The iOS build is topping out our build times at 30+ minutes.

We should investigate how to bring this down to the range of other builds (10-15 minutes).

Here's one thing to try: https://labs.spotify.com/2013/11/04/shaving-off-time-from-the-ios-edit-build-test-cycle/


jfirebaugh:
Down to ~500 seconds with these changes. 🎉

PR #4685, opened 2016-04-12
Scroll to zoom on OS X

### LINKED ISSUE #4670: Allow scrollwheel zooming
On OS X, the default behavior of the trackpad is to pan the map. However, this is counter to how all in-browser maps work and causing constant frustration to me, especially when switching back and forth. Could we please add a preference that disables panning with the trackpad and instead zooms in/out when scrolling on the trackpad?

/cc @1ec5

1ec5:
Added a hidden preference to MGLMapView that enables “unnatural zooming”, in which the scroll wheel zooms in and out, as in a Web browser, instead of scrolling the map, as is the platform convention on iOS and OS X.

Added a checkbox to osxapp’s preferences window that toggles this hidden preference via a binding. You can use this checkbox, or invoke `defaults write com.mapbox.MapboxGL MGLScrollWheelZoomsMapView -bool YES`, or pass `-MGLScrollWheelZoomsMapView YES` into osxapp as command line arguments.

Fixes #4670.

/cc @kkaefer


1ec5:
The OS X build failure is https://github.com/mapbox/mapbox-gl-native/issues/4665#issuecomment-208595647. Merging.

PR #4687, opened 2016-04-12
Darwin RunLoop: abort signaled AsyncTasks

kkaefer:
When signaling an `AsyncTask`, then destroying the task before it fired, the Darwin `RunLoop` implementation currently crashes.


jfirebaugh:
👍

Issue #3621, opened 2016-01-19
Exclude iOS simulator location events from telemetry


friedbunny:
/cc @1ec5 @boundsj

1ec5:
We can `#ifdef` out some entry point for MGLMapboxEvents when compiled for x86. That’ll prevent any iOS Simulator location updates from reaching the telemetry endpoint. There’s still a possibility that Xcode’s Simulated Location feature would cause telemetry data to be sent too. We can address that by detecting the presence of a debugger, but I think that’d be lower priority, since Xcode’s default simulations are single motionless points.

PR #4691, opened 2016-04-12
[iOS] Convert telemetry test server setting to user default

friedbunny:
- Previously looked in app's Info.plist, now looks at NSUserDefaults
- Deprecation should be unnecessary for this internal test setting
- ~~Harmonized name with other keys (added 'Mapbox’)~~ Key name is `MGLTelemetryTestServerURL`

/cc @boundsj @1ec5

### DIFF
--- a/platform/ios/app/MBXViewController.m
+++ b/platform/ios/app/MBXViewController.m
@@ -40,9 +40,8 @@ static const CLLocationCoordinate2D WorldTourDestinations[] = {
             @"MBXUserTrackingMode": @(MGLUserTrackingModeNone),
             @"MBXShowsUserLocation": @NO,
             @"MBXDebug": @NO,
+            @"MGLTelemetryTestServerURL": @"https://cloudfront-staging.tilestream.net",
         }];
-        
-        
     }
 }
 
--- a/platform/ios/src/MGLAPIClient.m
+++ b/platform/ios/src/MGLAPIClient.m
@@ -87,7 +87,7 @@ static NSString * const MGLAPIClientHTTPMethodPost = @"POST";
 }
 
 - (void)setupBaseURL {
-    NSString *testServerURL = [[NSBundle mainBundle] objectForInfoDictionaryKey:@"MGLMetricsTestServerURL"];
+    NSString *testServerURL = [[NSUserDefaults standardUserDefaults] stringForKey:@"MGLTelemetryTestServerURL"];
     if (testServerURL) {
         _baseURL = testServerURL;
         _usesTestServer = YES;
--- a/platform/ios/app/app-info.plist
+++ b/platform/ios/app/app-info.plist
@@ -24,8 +24,6 @@
 	<string>7877</string>
 	<key>LSRequiresIPhoneOS</key>
 	<true/>
-	<key>MGLMetricsTestServerURL</key>
-	<string>https://cloudfront-staging.tilestream.net</string>
 	<key>NSAppTransportSecurity</key>
 	<dict>
 		<key>NSExceptionDomains</key>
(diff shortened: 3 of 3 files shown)


friedbunny:
Should iosapp default to the production server, instead?

platform/ios/src/MGLAPIClient.m
```diff
@@ -87,7 +87,7 @@ - (NSURLRequest *)requestForEvents:(NS_ARRAY_OF(MGLMapboxEventAttributes *) *)ev
 }
 
 - (void)setupBaseURL {
-    NSString *testServerURL = [[NSBundle mainBundle] objectForInfoDictionaryKey:@"MGLMetricsTestServerURL"];
+    NSString *testServerURL = [[NSUserDefaults standardUserDefaults] stringForKey:@"MGLMapboxMetricsTestServerURL"];
```
1ec5: How about using [`-[NSUserDefaults URLForKey:]`](https://developer.apple.com/library/ios/documentation/Cocoa/Reference/Foundation/Classes/NSUserDefaults_Class/index.html#//apple_ref/occ/instm/NSUserDefaults/URLForKey:)? You can register the default with an NSURL instance; it'll be stored as a `<data>` value.

platform/ios/app/MBXViewController.m
```diff
@@ -40,6 +40,7 @@ + (void)initialize
             @"MBXUserTrackingMode": @(MGLUserTrackingModeNone),
             @"MBXShowsUserLocation": @NO,
             @"MBXDebug": @NO,
+            @"MGLMapboxMetricsTestServerURL": @"https://cloudfront-staging.tilestream.net",
```
1ec5: Since you're already renaming the default, how about replacing "MapboxMetrics" with "Telemetry"?

friedbunny: I thought about that, but we have a variety of other legacy keys that are user-facing and we don’t want to change. Seems like it may be confusing to not be consistent.

platform/ios/src/MGLAPIClient.m
friedbunny: Didn’t think of that — I’ll give that a try now.

1ec5:
Not until #3621 is fixed.

platform/ios/src/MGLAPIClient.m
friedbunny: `MGLAPIClientBaseURL` (the production server) is defined as `static NSString * const`, which means a switch to NSURL for `_baseURL` and `testServerURL` wouldn’t make sense here.

1ec5: Yeah, I'm thinking we should register a default for this default wherever we're currently calling `-[NSUserDefaults registerDefaults:]`.

friedbunny: Is [this diff](https://gist.github.com/friedbunny/cbc10c75da27dbda4dce5c69df08f388) what you had in mind?

- We have to `registerDefaults` separately in MGLAPIClient because `MGLAPIClientBaseURL` is defined there, not in MGLMapboxEvents.
- NSURL has to be converted to NSData before it can be registered. Is there a better way to do this than NSKeyedArchiver?
- We still have to check if we’re using a non-standard telemetry server (for the certificate checks later), so now there’s a string comparison between the URL and the production URL.

friedbunny: > You can register the default with an NSURL instance; it'll be stored as a `<data>` value.

Doing this (without first wrapping it in an NSData):

```objc
    NSURL *defaultTelemetryServerURL = [NSURL URLWithString:MGLAPIClientBaseURL];
    [[NSUserDefaults standardUserDefaults] registerDefaults:@{
        @"MGLTelemetryServerURL": defaultTelemetryServerURL,
    }];
```

Results in this:

```
Attempt to set a non-property-list object https://api.mapbox.com as an NSUserDefaults/CFPreferences value for key MGLTelemetryServerURL
*** Terminating app due to uncaught exception 'NSInvalidArgumentException', reason: 'Attempt to insert non-property list object https://api.mapbox.com for key MGLTelemetryServerURL’
```

1ec5: Ah, I guess that’s probably too much hassle, then. Sorry about that. It turns out that you can use `-URLForKey:` and `-setURL:forKey:` without going through NSData, but you have to wrap in NSData to register a default.

1ec5:
:+1:

platform/ios/app/MBXViewController.m
friedbunny: On second though, PROGRESS++! Changed to `MGLTelemetryTestServerURL` in ff425b9.

PR #4701, opened 2016-04-14
[Android] SupportMapFragment: add  newInstance method with onMapReadyCallback

vacxe:
I think this easy way add callback, when you create fragment in code.

@tobrun @bleege what's you think about this?

### DIFF
--- a/platform/android/MapboxGLAndroidSDK/src/main/java/com/mapbox/mapboxsdk/maps/SupportMapFragment.java
+++ b/platform/android/MapboxGLAndroidSDK/src/main/java/com/mapbox/mapboxsdk/maps/SupportMapFragment.java
@@ -50,12 +50,42 @@ public class SupportMapFragment extends Fragment {
      * @return MapFragment created.
      */
     public static SupportMapFragment newInstance(@Nullable MapboxMapOptions mapboxMapOptions) {
-        SupportMapFragment mapFragment = new SupportMapFragment();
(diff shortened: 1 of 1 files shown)


tobrun:
@vacxe the way we currently set up a fragment is on par with [MapFragment](https://developers.google.com/android/reference/com/google/android/gms/maps/SupportMapFragment) Google Maps SDK. 
Is there another reason behind this PR besides convenience?

vacxe:
@tobrun no, just i think it be convenience. This solution not conflict with Google Maps SDK, but expands him. If you set strategy Mapbox develop of absolute migration gmSDK im agree with you (Not problem, close PR). 

Simply i'm just like make some things better )

tobrun:
@vacxe 
> Simply i'm just like make some things better 

That is a great motivation! keep this up.

Related to the issue, For me it boils down to the responsibility an object (in this case a fragment) has when being created with a `newInstance` method. The object itself is responsible for managing the state introduced with `newInstance` and is responsible for making sure the state is restored after being destroyed/recreated (for this  a fragment using a bundle with arguments). This is a key difference between an `OnMapChangeListener` and `MapboxMapOptions`.  For the former it is not possible to restore/destroy/recreate it, for the latter it is.

vacxe:
@tobrun im close this PR? Please look https://github.com/mapbox/mapbox-gl-native/pull/4735

PR #4709, opened 2016-04-14
Store glActiveTexture in GLConfig

kkaefer:
We're currently calling this function manually in some places, but we should move that value to `GLConfig` so that it can be reset properly if the need arises.


kkaefer:
@ansis can I merge this as-is?

ansis:
🚢

PR #4716, opened 2016-04-15
[android] Restore prior GLContext if exists

tmpsantos:
Apparently some versions of Android, notable 4.4.4 running on my Nexus 5, creates a GLContext on the Android UIThread which is the one we are now using for rendering after #2909.

If the context is not restored, nothing gets rendered (or sometimes partially, or artifacts) because Android will try to do GL stuff on the context used by Mapbox GL and mess things up.

### DIFF
--- a/platform/android/src/native_map_view.cpp
+++ b/platform/android/src/native_map_view.cpp
@@ -127,6 +127,11 @@ std::array<uint16_t, 2> NativeMapView::getFramebufferSize() const {
 void NativeMapView::activate() {
     mbgl::Log::Debug(mbgl::Event::Android, "NativeMapView::activate");
 
+    oldDisplay = eglGetCurrentDisplay();
+    oldReadSurface = eglGetCurrentSurface(EGL_READ);
+    oldDrawSurface = eglGetCurrentSurface(EGL_DRAW);
+    oldContext = eglGetCurrentContext();
+
     assert(vm != nullptr);
 
     if ((display != EGL_NO_DISPLAY) && (surface != EGL_NO_SURFACE) && (context != EGL_NO_CONTEXT)) {
@@ -150,7 +155,13 @@ void NativeMapView::deactivate() {
 
     assert(vm != nullptr);
 
-    if (display != EGL_NO_DISPLAY) {
+    if (oldContext != context && oldContext != EGL_NO_CONTEXT) {
+        if (!eglMakeCurrent(oldDisplay, oldDrawSurface, oldReadSurface, oldContext)) {
+            mbgl::Log::Error(mbgl::Event::OpenGL, "eglMakeCurrent() returned error %d",
+                             eglGetError());
+            throw std::runtime_error("eglMakeCurrent() failed");
+        }
+    } else if (display != EGL_NO_DISPLAY) {
         if (!eglMakeCurrent(display, EGL_NO_SURFACE, EGL_NO_SURFACE, EGL_NO_CONTEXT)) {
             mbgl::Log::Error(mbgl::Event::OpenGL, "eglMakeCurrent(EGL_NO_CONTEXT) returned error %d",
                              eglGetError());
--- a/platform/android/src/native_map_view.hpp
+++ b/platform/android/src/native_map_view.hpp
@@ -61,6 +61,12 @@ private:
     jweak obj = nullptr;
 
     ANativeWindow *window = nullptr;
+
+    EGLDisplay oldDisplay = EGL_NO_DISPLAY;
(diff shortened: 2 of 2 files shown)


tmpsantos:
:eyes: @jfirebaugh @tobrun

platform/android/src/native_map_view.cpp
```diff
@@ -196,6 +201,14 @@ void NativeMapView::render() {
     }
 
     deactivate();
+
+    if (currentContext != context && currentContext != EGL_NO_CONTEXT) {
```
mb12: Why don't you just comment out the call to deactivate instead? It's equivalent.

tmpsantos: :+1: 

But I think it is better `deactivate()` if there is no prior GLContext. Gonna change that.

platform/android/src/native_map_view.cpp
```diff
@@ -195,7 +200,15 @@ void NativeMapView::render() {
...
 
-    deactivate();
+    if (currentContext != context && currentContext != EGL_NO_CONTEXT) {
+        if (!eglMakeCurrent(currentDisplay, currentDrawSurface, currentReadSurface, currentContext)) {
+            mbgl::Log::Error(mbgl::Event::OpenGL, "eglMakeCurrent() returned error %d",
+                             eglGetError());
+            throw std::runtime_error("eglMakeCurrent() failed");
+        }
+    } else {
+        deactivate();
```
tmpsantos: For the record, on my Nexus 5x Android 6.0.1 `deactivate()` gets called but on my Nexus 5 Android 4.4.4 the context gets replaced with the prior context.

mb12: @tmpsantos My guess is that  on "Nexus 5 Android 4.4.4", either the View Framework or some other opengl view is changing the opengl state prior to call to NativeMapView::render. Is it possible to do the following experiment in your sandbox with the following two changes?  If the 'screen is white bug'  goes away, its an issue with not handling the opengl state correctly.

1. Instantiate mbgl::Map with GLContextMode::Shared as the fourth argument  [here](https://github.com/mapbox/mapbox-gl-native/blob/master/platform/android/src/native_map_view.cpp#L83). This will ensure that gl state is reset every time by mbgl::Map.
2.  Call glViewPort unconditionally by commenting out if(sizeChanged)  (The current code  will consistently cause crashes/wrong behavior if there are multiple Opengl views in the same application. Its also missing the glError call that MBGL_ERROR_CHECK provides.).

jfirebaugh:
If saving and restoring the specific display, context, and surfaces is necessary, it seems like that should be part of `activate` / `deactivate` themselves, to be consistent wherever the `activate` / `deactivate` pair is used.

mb12:
@tmpsantos and @jfirebaugh  Is there any plan to fix this? All this complexity will go away when this is fixed.
https://github.com/mapbox/mapbox-gl-native/issues/1662

tmpsantos:
> @tmpsantos and @jfirebaugh Is there any plan to fix this? All this complexity will go away when this is fixed.

I still want to work at this issue, but I don't have a timeline for that.

Now that Android renders on the same thread as the mbgl::Map object that is feasible (we need to create the mbgl::Map on the GLSurfaceView.Renderer thread). The challenge is to proxy every getter and setter from the UIThread to this renderer thread, specially because some of the getters are synchronous.

platform/android/src/native_map_view.cpp
tmpsantos: > @tmpsantos My guess is that on "Nexus 5 Android 4.4.4", either the View Framework or some other opengl view is changing the opengl state prior to call to NativeMapView::render. Is it possible to do the following experiment in your sandbox with the following two changes? If the 'screen is white bug' goes away, its an issue with not handling the opengl state correctly.
> 
> Instantiate mbgl::Map with GLContextMode::Shared as the fourth argument here. This will ensure that gl state is reset every time by mbgl::Map.
> Call glViewPort unconditionally by commenting out if(sizeChanged) (The current code will consistently cause crashes/wrong behavior if there are multiple Opengl views in the same application. Its also missing the glError call that MBGL_ERROR_CHECK provides.).

What really happens is there is another GL context in use that we cannot use/share because is not configured with the attributes we expect. `GLContextMode::Shared` didn't work in this case (I tested). We really have to use our own context and after we are done rendering, restore the old context that Android expects to be active.

@mb12 thanks for the feedback!

tmpsantos:
> it seems like that should be part of activate / deactivate themselves

:+1:

PR #4641, opened 2016-04-07
Migrate iOS and OS X targets to checked-in Xcode projects

1ec5:
This PR replaces the various iOS-specific targets in ios.xcodeproj and OS X–specific targets in osx.xcodeproj with checked-in targets created within Xcode. The core targets remain in the GYP-generated platform.xcodeproj projects. New Xcode workspaces bind ios.xcodeproj to platform.xcodeproj and manages dependencies between the two; likewise for osx.xcodeproj. The ios.xcworkspace also contains ios-test.xcodeproj and KIF.xcodeproj for convenience. (Although we really should be pulling in KIF via a package manager like CocoaPods: #1522.)

You still set up your development environment with `make iproj` and `make xproj`, but now those make rules open the workspace. Other workflows have been streamlined:

* When you add or remove an SDK file, you can do so inside Xcode. No more build failures because you forgot to also update platform.gypi and rerun `make iproj`.
* You can leave the workspace open in Xcode while performing Git operations that add and remove SDK files. You can also run `make iproj` or `make xproj` with the workspace still open.
* You can debug the iOS SDK and core code while running KIF-based integration tests.
* Derived data locations are shared between the Xcode GUI and `xcodebuild` command line tool, so you no longer have to rebuild from scratch when switching between the two. Cleaning in one environment automatically affects the other environment.
* Incremental builds of the SDK are much faster.
* `make clean` is more thorough, wiping away Xcode’s index and module cache along with the intermediate files.
* You can add additional test bundles or demo applications inside Xcode without having to reverse-engineer undocumented GYP features.

libmbgl-platform-ios.a has been reduced to a very small library that contains just the iOS-specific code needed by core code. (The same was done for libmbgl-platform-osx.a in #4608.) The iOS SDK dynamic framework and iOS SDK static library link the core static libraries directly, making it possible for the two to have differing content should the need ever arise. To significantly reduce binary sizes and build times, the iOS SDK no longer builds the `armv7s` slice; iPhone 5 and iPhone 5c will fall back to the `armv7` slice automatically without a noticeable performance hit.

`make test-ios` additionally runs the same “Darwin” unit tests that we’ve been running against the OS X SDK. It’d be trivial to add functional tests (whether the existing KIF tests or future XCUITest tests) to this make rule.

The workspaces come with disabled Objective-C exception breakpoints by default. They also include schemes that exactly replicate what gets built on Bitrise and what gets built by package.sh for distribution.

[...]

Issue #4191, opened 2016-03-03
Allow C99 extensions in iOS and OS X SDK targets


1ec5:
C99 extensions are currently disallowed throughout the project due to pedantic warnings being enabled. This makes no sense in Objective-C code, since we’ve been able to depend on C99 support for years. At the least, the iOS and OS X SDK targets should get a `-Wno-c99-extensions` flag. This will enable us to remove a fair amount of redundant code in MGLMapView.

If any non-SDK Objective-C code needs to use C99-specific features like designated initializers or the Elvis operator, we can selectively turn off the Clang warning via `#pragma clang diagnostic ignored "-Wc99-extensions"` on a case-by-case basis.

/cc @jfirebaugh @boundsj

boundsj:
:+1:

PR #4722, opened 2016-04-17
Take advantage of C99

1ec5:
Now that #4641 has removed the prohibition against C99 extensions, this change removes many pragma statements that suppress warnings about variadic macros and designated initializers. Designated initializers and Elvis operators have been introduced wherever appropriate.

/ref #4191
/cc @boundsj @friedbunny


1ec5:
These changes touch one public header, MGLTypes.h. I don’t consider it a backwards-incompatible change because Objective-C projects have had C99 enabled by default since Xcode 3.x.

friedbunny:
Cool. 👍

PR #4724, opened 2016-04-18
Make userTrackingMode KVO-compliant

### LINKED ISSUE #4363: User tracking mode is not KVO-compliant
Hey guys,

# The Problem
Apple requires any class that overrides a property setter to call NSKeyValueObserving lifecycle methods before and after setting the iVar. `MGLMapView`'s `- (void)setUserTrackingMode:` does not do this. Thus, anyone who tries to observe the `userTrackingMode` key on `MGLMapView` will never see their `observeValueForKeyPath:ofObject:change:context:` method fire.

# The Fix
In `MGLMapView.m`, in `- (void)setUserTrackingMode:`, add the following lines:
*  Before the ` _userTrackingMode = mode;` assignment on line `3150`, add `    [self willChangeValueForKey:NSStringFromSelector(@selector(userTrackingMode))];`
*  After the ` _userTrackingMode = mode;` assignment on line `3150`, add `    [self didChangeValueForKey:NSStringFromSelector(@selector(userTrackingMode))];`

1ec5:
Fixes #4363.

/cc @friedbunny


friedbunny:
👍

Issue #2822, opened 2015-10-27
write iOS metrics tests


incanus:
Refs https://github.com/mapbox/mapbox-gl-native/commit/1421755434963cd4a35e7a4f757bf400f90528b7 and https://github.com/mapbox/mapbox-gl-native/pull/2769.

1ec5:
Blocked by #3155.

1ec5:
The non-UI, XCTest-based unit tests added to OS X in #3406 and to iOS in #4641 offer a way forward for this functionality. We should be able to mock user location and HTTP requests so that MGLMapboxEvents can operate more or less independently of MGLMapView. The tricky part will be managing MGLMapboxEvents’ lifecycle, since it’s a singleton.

jfirebaugh:
Closing; this has been open for a long time without progress and isn't likely to happen.

PR #4726, opened 2016-04-18
Use staging server in simulator

### LINKED ISSUE #3621: Exclude iOS simulator location events from telemetry

1ec5:
~~Use the staging telemetry server when compiled for the iOS Simulator. An alternative would’ve been to disable telemetry altogether in this scenario, but I think leaving it enabled and sending it to the staging endpoint leaves open the possibility of testing telemetry with mocked requests (#2822).~~

Disable telemetry when compiled for the iOS Simulator.

Fixes #3621.

/cc @boundsj @friedbunny @bleege


boundsj:
👍 thanks. and also +1 to leaving it enabled but pointed to staging.

friedbunny:
👍

bleege:
Yep, this makes sense.  We do the same thing for the Android TestApp.

1ec5:
Per chat with @bsudekum, we’ve decided to remove references to the staging endpoint from the iOS SDK. Therefore, this PR now disables telemetry on the simulator. We’ll need to create an override mechanism when we’re ready to test telemetry.

PR #4728, opened 2016-04-18
Prompt for access token in iosapp

### LINKED ISSUE #3102: Reloading a failed MGLMapView
If a `MGLMapView` fails to load (due to not being able to fetch the styles, usually for network reasons), is there a recommended way of trying again, other than unsetting and setting the `styleURL`?

### LINKED ISSUE #4696: iosapp should prompt for an access token
iosapp currently requires that you set a MAPBOX_ACCESS_TOKEN the first time you launch after a clean build; after that point, the access token is persisted as a user default. People (myself included) always forget to set this environment variable.

When iosapp is unable to find an access token in user defaults, it should prompt for one. This is what osxapp does, and it's a lot more convenient than what we have now.

If there's a need to automate setting the access token outside of checked-in files, we should add a Run Script build phase that reads the contents of ~/.mapbox and inserts the access token into the Info.plist in the build product (which is not checked in). Alternatively, you could add a command line argument in the scheme editor (right above the UI for setting environment variables) to the effect of `-MBXMapboxAccessToken sk.feedcafedeadbeefbadebede`. (You can already do this for any user default.)

/cc @kkaefer @friedbunny

1ec5:
If no access token is found in user defaults at launch, iosapp prompts the user for an access token. Renamed the user defaults key for the access token to conform to Cocoa naming conventions. The `MAPBOX_ACCESS_TOKEN` environment variable is still honored.

Added an action to MGLMapView that reloads the current style. Note that it doesn’t clear the cache; it’s pretty much only useful for ensuring that a change in access token is reflected on the map or for forcing the map view to retry loading tiles independently of the exponential back-off strategy.

Fixes #3102 and fixes #4696.

/cc @friedbunny @kkaefer

### DIFF
--- a/platform/ios/CHANGELOG.md
+++ b/platform/ios/CHANGELOG.md
@@ -8,6 +8,7 @@ Mapbox welcomes participation and contributions from everyone.  If you’d like 
 - Removed the `armv7s` slice from the SDK to reduce its size. iPhone 5 and iPhone 5c automatically use the `armv7` slice instead. ([#4641](https://github.com/mapbox/mapbox-gl-native/pull/4641))
 - The user dot now moves smoothly between user location updates while user location tracking is disabled. ([#1582](https://github.com/mapbox/mapbox-gl-native/pull/1582))
 - User location heading updates now resume properly when an app becomes active again. ([#4674](https://github.com/mapbox/mapbox-gl-native/pull/4674))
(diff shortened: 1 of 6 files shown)


friedbunny:
As you noted, users will expect a `-reloadStyle:` method to fetch the current style from the network and display changes. Until we do have such an implementation, can we keep this as a private method or as a category/extension where necessary?

If we make this public now, let’s add a warning to the inline docs stating that it does not affect cached resources.

1ec5:
I’d rather that iosapp rely solely on public API, and it does fetch the current style when the original requests failed due to network conditions. So let’s document the caveats.

PR #4732, opened 2016-04-18
added projMatrix, nativeMatrix and extrudeMatrix to CustomLayerRender…

mozinator:
This change passes the projmatrix, nativematrix and extrudematrix via the customlayer parameters. which makes it easier to draw items on the map in a custom layer.

### DIFF
--- a/include/mbgl/style/types.hpp
+++ b/include/mbgl/style/types.hpp
@@ -210,6 +210,9 @@ struct CustomLayerRenderParameters {
     double bearing;
     double pitch;
     double altitude;
+    double projMatrix[16];
+    double nativeMatrix[16];
+    double extrudeMatrix[16];
 };
 
 /**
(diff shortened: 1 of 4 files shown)


1ec5:
These parameters should probably be plumbed through to the SDK APIs too, for example https://github.com/mapbox/mapbox-gl-native/blob/master/platform/ios/src/MGLMapView.mm#4231. Perhaps CATransform3D would be an appropriate type for this on iOS.

/cc @jfirebaugh

mozinator:
Why not just a double array ? That GLKit or CoreGraphics users are both helped

1ec5:
We could also include a utility function in MGLGeometry.h to convert between the two, but a double array is reasonable. 👍

mozinator:
Just tried it out with a double array, which makes it ugly with swift. 
UnsafePointer<Double>
I think GLKMatrix4 and/or a utility function in MGLGeometry makes it nicer

mozinator:
something like this, maybe ?

jfirebaugh:
Thanks for the contribution. Given that these are implementation details that can and do change (we're in the process of removing `extrudeMatrix` at the moment), and also that custom layers in general is not a publicly supported API, I'm going to opt to keep the existing API.

PR #4737, opened 2016-04-18
Checked-in Xcode project followups

jfirebaugh:
cc @1ec5


platform/ios/.gitignore
```diff
@@ -0,0 +1,2 @@
+ios.xcodeproj/xcuserdata
+ios.xcworkspace/xcuserdata
```
1ec5: No `xcuserdata` should ever be committed in my opinion. The shared settings live alongside these directories, in `xcshareddata`.

jfirebaugh: Are you suggesting broadening the ignore rule to (unprefixed) `xcuserdata`?

1ec5: Yes.

friedbunny:
screen shot 2016-04-18 at 3 45 18 pm

On every fresh repo clone, anyone who wants to run the iosapp demo will now have to create the scheme themselves (#4741). This is not difficult, but is certainly less than friendly.

1ec5:
Once #4728 lands, maybe we should instead discourage folks from setting this environment variable. As for ios-bench, since that application needs user intervention anyways to download the assets locally (#4740), we should have it wait until you set your access token inside the application, then start benchmarking.

PR #4740, opened 2016-04-18
Use downloaded local assets in ios-bench if present

1ec5:
If you’ve gone through the trouble of running ios-bench’s download.sh scripts, use those assets instead of fetching them at runtime from the server.

This fixes a regression introduced in abf93c008b993f178df327e7681418df51e72844 for #2746.

/cc @kkaefer

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -20,3 +20,5 @@ xcuserdata
 /test/fixtures/**/diff.png
 /test/output
 /node_modules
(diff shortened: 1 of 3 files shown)


1ec5:
I experimented with adding a Run Script build phase that ran the two download.sh scripts, but they depend on the `MAPBOX_ACCESS_TOKEN` environment variable, and there isn’t a good way to set a build setting for personal use to avoid committing the access token.

PR #4741, opened 2016-04-18
Update the DEVELOPING docs

boundsj:
https://github.com/mapbox/mapbox-gl-native/pull/4737 broke the docs we use to walk potential devs through getting the development env set up. Since we no longer share (and check in) the scheme, setting up the env for iOS development now requires the extra step of generating a scheme. This PR updates the docs to reflect that.

cc @jfirebaugh @1ec5 @friedbunny

### DIFF
--- a/platform/ios/DEVELOPING.md
+++ b/platform/ios/DEVELOPING.md
@@ -12,7 +12,7 @@ The Mapbox iOS SDK and iosapp demo application build against the iOS 7.0 SDK and
 
     make iproj
 
-Which will create and open an Xcode workspace that includes both the SDK source and some Objective-C test applications. Before building, use the scheme picker button in the toolbar to change the scheme to “iosapp” and the destination to one of the simulators or connected devices listed in the menu.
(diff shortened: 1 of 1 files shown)


1ec5:
The schemes should be there if you have Autocreate Schemes set in the Manage Schemes sheet.

boundsj:
I do have that set @1ec5 and a scheme for iosapp did not appear out of the box for me and it sounds like not for @friedbunny, too.

jfirebaugh:
Yeah, I would have expected the iosapp/osxapp scheme to be autocreated, but it doesn't seem to be.

friedbunny:
Aye, “Autocreate schemes” was enabled and failed to work for me. In addition to fresh clones, updating existing repos to the scheme-less master looks like it’s going to require (most?) developers to recreate the relevant schemes.

1ec5:
I think we should undo bc3512773ffe942e6f70c43596490a6fcf8920a4 and remove support for `MAPBOX_ACCESS_TOKEN` from iosapp, or at least remove any documentation about it.

jfirebaugh:
Does osxapp prompt as well? If so, 👍 .

1ec5:
Yes, osxapp pops up an alert, and the default button takes you to the Preferences window, where you can set an access token.

1ec5:
Closing in favor of #4743.

PR #4742, opened 2016-04-19
[osx] Create shared schemas for mbgl-offline, mbgl-render, and test targets that set a working directory

jfirebaugh:
A change I've been wanting to make for awhile was unlocked by #4641.

These targets expect the working directory to be set to the repository root.

cc @1ec5


1ec5:
:+1:

PR #4743, opened 2016-04-19
Prompt for access token in ios-bench; restore iosapp scheme

1ec5:
ios-bench now prompts for an access token on first launch just as iosapp and osxapp do (#4728). ios-bench waits to start benchmarking until after the access token is entered. If this application needs to run without user intervention, the `MAPBOX_ACCESS_TOKEN` environment variable is still supported.

Removed the `MAPBOX_ACCESS_TOKEN` environment variable from iOS contributor documentation. Since the tendency to inadvertently commit this environment variable was the reason for removing the iosapp scheme in #4737, I’ve reverted that change, so the contributor documentation is once again accurate.

/cc @jfirebaugh @boundsj @friedbunny @kkaefer

### DIFF
--- a/platform/ios/DEVELOPING.md
+++ b/platform/ios/DEVELOPING.md
@@ -31,13 +31,7 @@ You can customize the build output by passing the following arguments into the `
 
 ### Access tokens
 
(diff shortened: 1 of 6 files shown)


1ec5:
If you do prefer to set the environment variable still, I recommend duplicating the provided iosapp or ios-bench scheme and setting the variable on that unshared scheme.

jfirebaugh:
👍

Issue #4738, opened 2016-04-18
Output from builds in Xcode is not going to build directory


jfirebaugh:
From a clean checkout, I ran `make xproj` and then built the `osxapp` target. I expected derived data to go to `build/osx-x86_64`, but instead it's going to `~/Library/Developer/Xcode/DerivedData/osx-boomqfdljdofupbolhbelbzoedvb`.

cc @1ec5

1ec5:
Looks like there’s no real way to force a workspace’s derived data settings for all users. I tried sharing the settings, but Xcode isn’t recognizing the shared settings. You can go to File ‣ Workspace Settings, change Derived Data to “Workspace-relative Location”, and set the location to `../../build/DerivedData`.

jfirebaugh:
Can we force it on a per-target basis?

1ec5:
Unfortunately, no. There’s a “legacy” setting in Workspace Settings, but even that is per-user. I haven’t looked into whether upgrading the projects from 3.x format to 6.x format would help, but gyp is generating 3.x projects anyways.

1ec5:
A hacky way would be to copy our desired WorkspaceSettings.xcsettings to a path based on the current user name at the end of the make rule.

jfirebaugh:
> A hacky way would be to copy our desired WorkspaceSettings.xcsettings to a path based on the current user name at the end of the make rule.

I think it would be worth doing that. We rely on getting build output to `build` so that command line and Xcode builds share the same output, and `make clean` works as advertised.

jfirebaugh:
Related: can we flip `OSX_DERIVED_DATA_PATH = build/DerivedData/osx` in the Makefile to `build/osx-x86_64/DerivedData`, or even simply `build/osx-x86_64`, keeping it consistent with other output for the platform?

1ec5:
> Related: can we flip `OSX_DERIVED_DATA_PATH = build/DerivedData/osx` in the Makefile to `build/osx-x86_64/DerivedData`, or even simply `build/osx-x86_64`, keeping it consistent with other output for the platform?

If we do that, Xcode will never match `xcodebuild`, because Xcode puts the name of the workspace there. I guess we could rename the workspace. Note that `build/ios/` (used by package.sh) also doesn’t match `build/ios-all` (used by gyp).

[...]

PR #4744, opened 2016-04-19
Rationalize ios and osx build output

jfirebaugh:
Refs #4738 

cc @1ec5 -- want to add what you mentioned in https://github.com/mapbox/mapbox-gl-native/issues/4738#issuecomment-211516320?

### DIFF
--- a/.ycm_extra_conf.py
+++ b/.ycm_extra_conf.py
@@ -37,7 +37,7 @@ import ycm_core
 
 compilation_database_folders = [
   'build/linux-x86_64/Release',
-  'build/osx-x86_64/Release',
+  'build/osx/Release',
 ]
 
 subprocess.call(['make compdb'], shell=True)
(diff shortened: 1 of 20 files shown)


1ec5:
I added the hack to overwrite the user’s workspace settings file. Technically, this is more aggressive than it should be, because the workspace settings file can contain other settings, like whether to show live errors. The only settings we really care about are `CustomBuildLocationType`, `DerivedDataCustomLocation`, and `DerivedDataLocationStyle`. But we can revisit that if people complain.

One other caveat to this hack is that it doesn’t work if you just delete your WorkspaceSettings.xcsettings file but not the rest of your .xcuserdatad folder: in that case, Xcode will overwrite the WorkspaceSettings.xcsettings we lay down with a WorkspaceSettings.xcsettings with default settings.

PR #4745, opened 2016-04-19
Fix hang observing user defaults on background queue

1ec5:
Fixed a deadlock that could occur if the host application attempts to set user defaults on a background queue. Now we observe user defaults changes synchronously.

/cc @jfirebaugh @boundsj

### DIFF
--- a/platform/ios/src/MGLMapboxEvents.m
+++ b/platform/ios/src/MGLMapboxEvents.m
@@ -240,6 +240,7 @@ const NSTimeInterval MGLFlushInterval = 180;
 }
 
 - (void)userDefaultsDidChange:(NSNotification *)notification {
+    MGLAssertIsMainThread();
     [self pauseOrResumeMetricsCollectionIfRequired];
 }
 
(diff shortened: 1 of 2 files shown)


boundsj:
👍

platform/ios/src/MGLMapboxEvents.m
```diff
@@ -249,6 +239,11 @@ - (NSString *)instanceID {
     return _instanceID;
 }
 
+- (void)userDefaultsDidChange:(NSNotification *)notification {
+    MGLAssertIsMainThread();
```
postmechanical: Why the assert here? Could `pauseOrResumeMetricsCollectionIfRequired` not just be dispatched back onto main thread if needed instead?

1ec5: Good point. Fixed in 3c4049c796270772184e4dd647c05e73774445e4.

1ec5:
One-off [build](https://github.com/mapbox/mapbox-gl-native/files/226296/mapbox-ios-sdk.1ec5-defaults-queue-2.zip) and [podspec](https://gist.github.com/1ec5/b2b029b64c6f55b3da4ed4325ebcfa62) for verification.

1ec5:
Cherry-picked into 3.2.1 as b8ecbf6b357a20ffd0b9d40c7425aac2a7b5d409...78ddcd76ee4b3ca1f3f958cf3cb089bacd0d4df3.

PR #4712, opened 2016-04-19
zooming is janky

kkaefer:
Zooming (easiest using Shift + drag) in the osxapp is noticeably janky, even in a Release build.

(Doesn't seem to have been caused by #2909; it was janky before that PR too.)

### DIFF
--- a/platform/osx/src/MGLMapView.mm
+++ b/platform/osx/src/MGLMapView.mm
@@ -381,7 +381,6 @@ public:
 
 /// Adds gesture recognizers for manipulating the viewport and selecting annotations.
 - (void)installGestureRecognizers {
-    self.acceptsTouchEvents = YES;
     _scrollEnabled = YES;
     _zoomEnabled = YES;
     _rotateEnabled = YES;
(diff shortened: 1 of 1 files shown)


kkaefer:
Issue was that the view object was eating many low-delta scroll events, interpreting them as touch events instead.

kkaefer:
@jfirebaugh will merge as soon as you give the go-ahead that this fixes your issues.

jfirebaugh:
@kkaefer It's a definite improvement, but still lacking the smoothness and fluidity of glfw-app. If you compare them side-by-side, the difference is striking: with glfw-app, you feel like you're directly manipulating the map surface. With osxapp, input is noticeably delayed and janky.

platform/osx/src/MGLMapView.mm
```diff
@@ -375,7 +375,6 @@ - (void)installAttributionView {
 
 /// Adds gesture recognizers for manipulating the viewport and selecting annotations.
 - (void)installGestureRecognizers {
-    self.acceptsTouchEvents = YES;
```
1ec5: This breaks touch events, right? Even if we conditionalize it on `MGLScrollWheelZoomsMapViewDefaultKey`, gestures like rotation and two-finger tap will be broken.

1ec5:
@jfirebaugh, is this remaining jankiness limited to zooming, or is it consistent with what you see when rotating or scrolling?

jfirebaugh:
With these changes, it's about equally noticeable across all gestures.

kkaefer:
> With these changes, it's about equally noticeable across all gestures.

Does that mean that scrolling becomes worse for you with this patch?

platform/osx/src/MGLMapView.mm
kkaefer: I tested both of these before submitting, and both rotation and two-finger tap still work.

1ec5: I just tried out this one change (without the rest of the changes in this PR), and all the gestures continue to work fine. In fact, performance improves dramatically with this one change. I think `acceptsTouchEvents` is about raw touch events only, but MGLMapView uses gesture recognizers, which are unaffected by this property.

Issue #4260, opened 2016-03-09
Map is black after returning from any test app "Individual Test"


jfirebaugh:
1. Run Android TestApp
2. Run an individual test such as "MapboxMap"
3. Press ⬅  button in top left
4. Map is black

Possibly relevant log messages:

```
03-09 03:04:07.565 4055-4055/com.mapbox.mapboxsdk.testapp E/Surface: getSlotFromBufferLocked: unknown buffer: 0x9f271270
03-09 03:04:07.573 4055-4106/com.mapbox.mapboxsdk.testapp E/Surface: getSlotFromBufferLocked: unknown buffer: 0x9f26ffa0
03-09 03:04:07.885 4055-6975/com.mapbox.mapboxsdk.testapp I/mbgl: {Map}[Android]: Not activating as we are not ready
03-09 03:04:07.923 4055-6975/com.mapbox.mapboxsdk.testapp E/libEGL: call to OpenGL ES API with no current context (logged once per thread)
```

tobrun:
I'm not seeing this behaviour on my nexus 5, are you testing on an emulator?

jfirebaugh:
Yeah, Android Studio + x86 emulator.

tobrun:
Adding a note that I'm not seeing this on Genymotion

tobrun:
I was able to reproduce this with the mentioned emulator above. I have tested this with the `Nexus 5` image, which is the same image as currently found in my real world Nexus 5 device. Since I'm not seeing any issues there. I'm going to pin this on an emulator issue and move it to the `4.1.0` release since this will not impact users from client applications.

tobrun:
In the meantime we have refactored the testapp and we don't have 2 activities after each other with both MapViews in them. For testing purposes we should include this.

tobrun:
I added an example in 2ce4ca5, was not able to reproduce the issue. Closing for now.

PR #4756, opened 2016-04-19
[android] #4260 - added example of 2 map activities

tobrun:
Refs #4260 

Add the possibility to have 2 Activities after each other both containing a mapview.


tobrun:

PR #4759, opened 2016-04-20
Version default style URL APIs; deprecate Emerald

### LINKED ISSUE #4577: Deprecate Emerald
`-[MGLStyle emeraldStyleURL]` should be marked deprecated in the iOS and OS X SDKs, because it’s expected to be removed from the API when the v9 styles are released. The Android SDK currently has this style hard-coded as well. Eventually, we’ll need to remove this style or alias it to Streets for backwards compatibility.

/cc @friedbunny @danswick @tobrun

### LINKED ISSUE #4702: Add version argument to class members of Style, deprecate older ones
> ~~Add a version argument to the Style class members. Each member documentation would say what the current latest version is, “as of” a certain date. Deprecate the unversioned methods.~~

**Update**:

For iOS:
> Add a version argument to the MGLStyle class methods. Each method’s documentation would say what the current latest version is, “as of” a certain date. Deprecate the unversioned methods.

For Android:
> Use same system as iOS, deprecate old `Style.java` class. 

In line with deprecation of the Emerald style in #4577 .

1ec5:
Updated default styles from v8 to v9 in the centralized default style constants used by the iOS and OS X SDKs as well as by the GLFW application. Deprecated the MGLMapView class methods in favor of new methods that take a version parameter. Replaced usage of the unversioned MGLStyle methods with the corresponding versioned methods and MGLStyleCurrentVersion to ensure consistency.

Expanded MGLStyle unit tests to also assert that MGLStyle has the right number of versioned style URL methods and that they’re all public.

* [x] Deprecate unversioned style URL methods in favor of versioned methods (#4702)
* [x] Deprecate Emerald (#4577)
* [x] Rename Hybrid to Satellite Streets
* [x] Figure out how to make MGLStyleCurrentVersion not be a trap: https://github.com/mapbox/mapbox-gl-native/pull/4759#discussion_r60455953
* [x] Keep unversioned style URL methods on v8: https://github.com/mapbox/mapbox-gl-native/pull/4812#discussion_r61304626
* [ ] Revise style descriptions to match www.mapbox.com

Fixes #4577 and fixes #4702. Hold until the v9 styles near release.

/cc @jfirebaugh @friedbunny

### DIFF
--- a/include/mbgl/util/default_styles.hpp
+++ b/include/mbgl/util/default_styles.hpp
@@ -14,16 +14,18 @@ struct DefaultStyle {
 };
 
 extern const DefaultStyle streets;
-extern const DefaultStyle emerald;
+extern const DefaultStyle outdoors;
 extern const DefaultStyle light;
 extern const DefaultStyle dark;
 extern const DefaultStyle satellite;
 extern const DefaultStyle hybrid;
 
 const DefaultStyle orderedStyles[] = {
-    streets, emerald, light, dark, satellite, hybrid,
+    streets, outdoors, light, dark, satellite, hybrid,
 };
 const size_t numOrderedStyles = sizeof(orderedStyles) / sizeof(DefaultStyle);
+
+static const unsigned currentVersion = 9;
 
 } // end namespace default_styles
 } // end namespace util
--- a/include/mbgl/util/default_styles.hpp
+++ b/include/mbgl/util/default_styles.hpp
@@ -18,10 +18,10 @@ extern const DefaultStyle outdoors;
 extern const DefaultStyle light;
 extern const DefaultStyle dark;
 extern const DefaultStyle satellite;
-extern const DefaultStyle hybrid;
+extern const DefaultStyle satelliteStreets;
 
 const DefaultStyle orderedStyles[] = {
-    streets, outdoors, light, dark, satellite, hybrid,
+    streets, outdoors, light, dark, satellite, satelliteStreets,
 };
 const size_t numOrderedStyles = sizeof(orderedStyles) / sizeof(DefaultStyle);
 
--- a/platform/android/MapboxGLAndroidSDK/src/main/java/com/mapbox/mapboxsdk/constants/Style.java
+++ b/platform/android/MapboxGLAndroidSDK/src/main/java/com/mapbox/mapboxsdk/constants/Style.java
@@ -1,11 +1,14 @@
 package com.mapbox.mapboxsdk.constants;
 
+import android.support.annotation.IntRange;
 import android.support.annotation.StringDef;
 
 import com.mapbox.mapboxsdk.maps.MapView;
 
 import java.lang.annotation.Retention;
 import java.lang.annotation.RetentionPolicy;
+import java.util.Locale;
+
 /**
  * <p>
  * Style provides URLs to several professional styles designed by Mapbox.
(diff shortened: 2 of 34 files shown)


platform/darwin/include/MGLStyle.h
```diff
@@ -4,50 +4,115 @@
 
 NS_ASSUME_NONNULL_BEGIN
 
-/** A collection of convenience methods for creating style URLs of default styles provided by Mapbox. These instances of NSURL are cached. */
+/**
+ A version number identifying the latest released version of the suite of default styles provided by Mapbox. This version number may be passed into one of the “StyleURLWithVersion” class methods of MGLStyle.
+ 
+ The value of this constant is current as of the date on which this SDK was published. Consult the [Mapbox Styles API documentation](https://www.mapbox.com/api-documentation/#styles) for the most up-to-date style versioning information.
```
friedbunny: Per #4760, we’ll want to HTML'ize this link.

platform/darwin/include/MGLStyle.h
```diff
@@ -4,50 +4,115 @@
 
 NS_ASSUME_NONNULL_BEGIN
 
-/** A collection of convenience methods for creating style URLs of default styles provided by Mapbox. These instances of NSURL are cached. */
+/**
+ A version number identifying the latest released version of the suite of default styles provided by Mapbox. This version number may be passed into one of the “StyleURLWithVersion” class methods of MGLStyle.
+ 
+ The value of this constant is current as of the date on which this SDK was published. Consult the <a href="https://www.mapbox.com/api-documentation/#styles">Mapbox Styles API documentation</a> for the most up-to-date style versioning information.
+ 
+ @warning The value of this constant may change in a future release of the SDK. If you use any feature that depends on a specific implementation detail in a default style, you may use the current value of this constant or the underlying style URL, but do not use the constant itself. Such implementation details may change significantly from version to version.
```
jfirebaugh: Let me go on the record stating that I think including a `MGLStyleCurrentVersion` constant is a trap and that we should require people to explicitly specify a version number.

But it we do include it, we should not use the term "implementation detail" -- this suggests that we don't consider style features to be part of an API contract, when in fact the very reason we are using versions is to indicate changes to documented semantics.

1ec5: I was thinking of it being used for sanity-checking more than anything. For example, a developer might want to assert in a unit test that the current version is still 9, as a reminder to upgrade the application to use the newest default style with the runtime styling API. We’d avoid using this constant anywhere else in documentation or guides. It’s sort of like how this framework (and all others) provides a `MapboxVersionNumber` constant just in case.

But you’re right: the way I’ve worded references to this constant in e.g. `-streetsStyleURLWithVersion:` could definitely lead developers to use it without reading this documentation and understanding the ramifications. If we can’t settle on a safer way to treat this constant, I agree that we should make it private.

platform/darwin/include/MGLStyle.h
```diff
@@ -4,50 +4,115 @@
...
+ Returns the URL to the current version of the Emerald style.
+ 
+ Mapbox Emerald is a versatile style with emphasis on road networks and public transportation.
+ */
++ (NSURL *)emeraldStyleURL __attribute__((deprecated("Use <mapbox://styles/mapbox/emerald-v8>.")));
+
+/**
+ Returns the URL to the given version of the Outdoors style.
+ 
+ Mapbox Outdoors is a rugged style that emphasizes physical terrain and outdoor activities.
```
1ec5: ⛏ @lucaswoj

tobrun:
Android equivalent targeting to merge with this PR in #4812

tobrun:
Android equivalent was merged with this PR

platform/darwin/include/MGLStyle.h
1ec5: How does 1e3b1f2a8fce7d22993deb90e584310dd39b3116 look to you? The constant is now `MGLStyleDefaultVersion`, with documentation that positions it as “here’s what MGLMapView chooses by default” rather than “here’s what you should pass into MGLStyle”. There are no public references to this constant; instead, documentation for each method spells out the current version “as of publication”, and a unit test verifies that it’s in sync with the constant.

1ec5:
As long as everything checks out, I’m going to branch off the v3.2.1 tag and cherry-pick these changes into that branch (or at least everything that fits after resolving conflicts), then release v3.2.2 from that branch.

friedbunny:
This looks good to me and works in practice (merged into a local v3.2.1 branch), even if the new styles aren’t available yet.

`MGLStyleDefaultVersion` is public, but I think the inline docs should be sufficient warning to developers about the perils of using it in production. I’m fine with explicitly not using `MGLStyleDefaultVersion` in examples and always hard-coding versions in code we want the general public to emulate.

friedbunny:
Running `make iframework`, there are warnings about NSInteger casting for each style:

```
⚠️  /mapbox-gl-native/platform/darwin/src/MGLStyle.mm:25:1: values of type 'NSInteger' should not be used as format arguments; add an explicit cast to 'long' instead [-Wformat]

MGL_DEFINE_STYLE(streets, streets)
^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```

1ec5:
> Running `make iframework`, there are warnings about NSInteger casting for each style (L26-30):

Just fixed and rebased. Thanks!

1ec5:
Cherry-picked the changes into the [release-ios-v3.2.2](https://github.com/mapbox/mapbox-gl-native/tree/release-ios-v3.2.2) branch as 6fd80aebde12a4ed34f75a751f702522853923d0...a5c7b2a18e643f69ea08d9e11a715cc4e2ed9c36. I omitted 0845158e5c662417fdf23ca4668c10b9fbb63a4e because only the iOS SDK will be built from this branch, and there were lots of Android conflicts that I didn’t quite follow.

PR #4766, opened 2016-04-20
Fix crash removing multipoint or duplicate annotation

1ec5:
Fixed a crash removing a polyline or polygon by only removing the `coordinate` key path observer from non-MGLMultiPoints. Noted in the changelog that you still need to replace an MGLMultiPoint to have it appear in a different location or in a different form.

The annotation tag is now passed in as the observation context. This fixes a crash when an annotation is added to the map view more than once, then removed more than once.

/ref https://github.com/mapbox/mapbox-gl-native/pull/3835#issuecomment-212342112
/cc @boundsj @friedbunny

### DIFF
--- a/platform/ios/src/MGLMapView.mm
+++ b/platform/ios/src/MGLMapView.mm
@@ -1612,12 +1612,21 @@ mbgl::Duration MGLDurationInSeconds(NSTimeInterval duration)
     {
         id <MGLAnnotation> annotation = object;
         MGLAnnotationTag annotationTag = (MGLAnnotationTag)(NSUInteger)context;
-        const mbgl::LatLng latLng = MGLLatLngFromLocationCoordinate2D(annotation.coordinate);
-        MGLAnnotationImage *annotationImage = [self imageOfAnnotationWithTag:annotationTag];
-        _mbglMap->updatePointAnnotation(annotationTag, { latLng, annotationImage.styleIconIdentifier.UTF8String ?: "" });
-        if (annotationTag == _selectedAnnotationTag)
+        // We can get here because a subclass registered itself as an observer
+        // of the coordinate key path of a non-multipoint annotation but failed
+        // to handle the change. This check deters us from treating the
+        // subclass’s context as an annotation tag. If the context happens to
+        // match a valid annotation tag, the annotation will be unnecessarily
+        // but safely updated.
+        if (annotation == [self annotationWithTag:annotationTag])
         {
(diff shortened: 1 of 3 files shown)


platform/ios/src/MGLMapView.mm
```diff
@@ -1610,19 +1608,16 @@ - (void)observeValueForKeyPath:(NSString *)keyPath ofObject:(id)object change:(N
             [MGLMapboxEvents ensureMetricsOptoutExists];
         }
     }
-    else if ([keyPath isEqualToString:@"coordinate"] && [object conformsToProtocol:@protocol(MGLAnnotation)])
+    else if ([keyPath isEqualToString:@"coordinate"] && [object conformsToProtocol:@protocol(MGLAnnotation)] && ![object isKindOfClass:[MGLMultiPoint class]])
     {
         id <MGLAnnotation> annotation = object;
-        MGLAnnotationTag annotationTag = [self annotationTagForAnnotation:annotation];
-        if (annotationTag != MGLAnnotationTagNotFound)
+        MGLAnnotationTag annotationTag = (MGLAnnotationTag)(NSUInteger)context;
```
friedbunny: I take it MGLMultiPoint doesn’t return a valid `annotationTag`, which is why we can assume the tag exists (now that we’re no longer trying to support multipoint here)?

friedbunny:
👍

platform/ios/src/MGLMapView.mm
1ec5: We do give MGLMultiPoints annotation tags, but we don’t bother to observe the `coordinate` key path on MGLMultiPoints, because MGLMultiPoints are immutable.

This `context` usage could potentially be unsafe if a developer subclasses MGLMapView and registers it as an observer for a non-multipoint annotation’s `coordinate` key path. The `context` can be used to distinguish multiple observations on the same key path and object, or to distinguish observation by multiple levels of the inheritance hierarchy, [but not both](https://mikeash.com/pyblog/key-value-observing-done-right.html).

1ec5: Fixed in 1f23e4536d6719012886649cc6ad6fc97857437d.

Even if the context in that case happens to match the annotation’s tag – say, because the developer passed in `NULL` (which is equal to 0) for the first added annotation (which would have a tag of 0) – the result would be an unnecessary but harmless update. There could still be trouble around removing the observer, but this is such an edge case as to be programmer error.

PR #4767, opened 2016-04-20
[Qt] Upstreaming of  Qt platform

tmpsantos:
Initial support for building Mapbox GL Native for the Qt Framework:

 * It should build on Linux and OSX and support Qt4 and Qt5 API.
 * Currently only Qt4 is built and tested on the bots but we should enable Qt5 too.
 * We expose for now a low level C++ API, but QML components should follow.
 * Missing API documentation and API tests (unit tests are all passing).

/cc @brunoabinader


brunoabinader:
Really happy to see this finally coming upstream! ❤️

PR #4768, opened 2016-04-20
[core] remove unused code in value.hpp

springmeyer:
Noticed this dead code while reviewing `mapbox::variant` usage with @artemp


jfirebaugh:
👍  Thanks!

Issue #1997, opened 2015-08-05
Make MapView's setStyleJson public


erf:
I would like to have setStyleJson as a public method in MapView, like so:

<pre>
    public void setStyleJson(String json){
        mNativeMapView.setStyleJson(json);
    }
</pre>

1ec5:
For reference, on the iOS side, we removed `-setStyleJSON:` in #1299 but are definitely open to reintroducing granular styling in some form (#837).

ljbade:
I'd much rather see #837 as the public API instead of this as setting the JSON directly is messy and I recall it causing issues a while ago.

@jfirebaugh Should we expose this to developers?

jfirebaugh:
Yep, #837 covers what our plans are for public APIs in this area. @erf, let us know if for some reason your use case wouldn't be satisfied with that approach.

erf:
@jfirebaugh #837 looks promising - has work on this begun for Android ?

jfirebaugh:
No, we haven't started on it for any native platform yet.

Issue #1367, opened 2015-04-29
Alternative method for setStyleJSON?


robipresotto:
mbglMap->setStyleJSON((std::string)[styleJSON cStringUsingEncoding:NSUTF8StringEncoding]

I can not see more this method on public api.

kkaefer:
Can you please elaborate?

1ec5:
As explained in #1299, I got rid of the `styleJSON` API for now, per https://github.com/mapbox/mapbox-gl-native/issues/1147#issuecomment-88171503 and #900. We'll eventually need some way to interact with styles with that kind of granularity for #837.

robipresotto:
still some crashes on json serialisation,

```ruby
10 Foundation                     0x00000001838ec6f0 +[NSJSONSerialization JSONObjectWithData:options:error:] + 80
```

1ec5:
On the other hand, if you’re seeing a crash in `+[NSJSONSerialization JSONObjectWithData:options:error:]`, it’s likely an exception caused by malformed JSON or a JSON object of a different type than GL expects. You should set an exception breakpoint and see where and why the exception is thrown.

robipresotto:
:+1:

Issue #1299, opened 2015-04-20
Response class in not NSHTTPURLResponse


robipresotto:
The api accept now only NSURL to set the json style,
How is possible download the json style, change it locally and reuse the data on mapstyle?

I got it all the time...
[ERROR] {0}{Map}[Setup]: loading style failed: response class is not NSHTTPURLResponse

kkaefer:
You can use `asset://...` to load from disk, or use `.setStyleJSON()` to load styles directly from memory.

robipresotto:
the setStyleJSON is private on master branch... as some reason for that?

kkaefer:
I don't think it's private: https://github.com/mapbox/mapbox-gl-native/blob/2831073f85b502ed0b00c3fb335fadfd705524a5/include/mbgl/map/map.hpp#L105

Maybe you're running an outdated version?

robipresotto:
yeah, i know... but nothing on MGLMapView.h

1ec5:
I got rid of the `styleJSON` API for now, per https://github.com/mapbox/mapbox-gl-native/issues/1147#issuecomment-88171503 and #900. We'll eventually need some way to interact with styles with that kind of granularity for #837.

/cc @incanus

1ec5:
Just to close the loop, if you pass a relative URL into `-setStyleURL:`, we’ll treat that as a local file relative to the main bundle. Absolute file URLs still lead to an error; see #3505.

Issue #900, opened 2015-02-18
get/setStyleJSON should not be public API


ljbade:
We only want to support setting a style via the URL (and `asset://`) mechanism provided by `get/setStyleURL`

Helps avoid issues with a style URL loading after the JSON is set directly.

Currently `get/setStyleJSON` is only needed in the tests to rewrite port numbers. We should find a way to expose this to the tests but not make it a public API.

/cc @kkaefer

mikemorris:
How are you defining public API @ljbade? I'm currently using `setStyleJSON` at https://github.com/mapbox/node-mapbox-gl-native/blob/master/src/node_map.cpp#L132 to set data loaded from a custom `FileSource` implementation.

incanus:
> Does `setStyleURL` correctly use the map's `FileSource` to load data?

Yes, this API now supports both regular HTTP URLs as well as `mapbox://user.style` map IDs.

mikemorris:
What I was asking is if it supports the refactored `FileSource` interface @incanus, such that when used with Node.js bindings I can plug in a custom `FileSource` rather than having something like `curl` baked in.

friedbunny:
We did this.

PR #4769, opened 2016-04-20
Expose setStyleJson from native class to MapboxMap and MapboxView

smartkiwi:
Here is the use case:
I need app the be able load style json from internal storage. So it could be updated periodically without publishing new version of functionality or publishing new style to mapbox studio (the phone might not have wireless connection during the day).
I've tried setStyleUrl - but it supports only http://. mapbox:// or assists//, but reject to load information from file://

`E/mbgl: {Map}[Setup]: loading style failed: unexpected url: file:/storage/emulated/0/Download/test_aoi_style.json`

### DIFF
--- a/platform/android/MapboxGLAndroidSDK/src/main/java/com/mapbox/mapboxsdk/maps/MapView.java
+++ b/platform/android/MapboxGLAndroidSDK/src/main/java/com/mapbox/mapboxsdk/maps/MapView.java
@@ -811,6 +811,23 @@ public class MapView extends FrameLayout {
     public String getStyleUrl() {
         return mStyleUrl;
     }
+    
+        /**
+     * <p>
+     * Loads a new map style from the specified bundled style.
+     * </p>
+     *
+     * @param newStyleJson The bundled style json code.
+     * @param base one of 'mapbox://', 'http://', 'https://', 'assets/'.     
+     * @see Style
+     */
+    @UiThread
+    public void setStyleJson(@NonNull String newStyleJson, @NonNull String base) {
+        if (mDestroyed) {
+            return;
+        }
+        mNativeMapView.setStyleJson(newStyleJson, base);
+    }
 
     //
     // Access token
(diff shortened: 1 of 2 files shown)


bleege:
@smartkiwi Thanks for the PR and the insight into what technically you're trying to accomplish.  In this case we're reluctant to merge this into the project as we're currently working on a [Runtime Styling API](https://github.com/mapbox/mapbox-gl-native/issues/837) that would likely (as best we can tell) accomplish the same thing technically and don't want to introduce an additional API with a short self life.  Can you tell us more about what the feature you're trying to build so that we can make sure that the Runtime Styling API will be able to meet those needs?  Thanks!

jfirebaugh:
Refs #900, #1299, #1367, #1997 for previous discussions about `setStyleJSON`.

smartkiwi:
Thank you for reviewing and providing context.
Sounds like Runtime Styling API would help with my use case, any idea when will it be ready?
The end goal is to highlight some area on a map and update that area every day.
We've tried to use polygon loaded from geojson but we couldn't make it transparent and still visible.
We were able to accomplish need visual result by creating layer in style and setting transparency for it.
The application would run on a phones without permanent wireless connection - so we would want to be able load updated style from local storage.

Actually we don't need whole style to be updated - just polygon for one of the layers.

I was hoping that loading style from Json would solve the task at hand - build proof of content app to be used internally.

But unfortunately I have problem building and/or importing of my fork of the Mapbox SDK.

vicapow:
Any guesses why the filesystem path to the style wouldn't work?

smartkiwi:
Here is the code I use to load style from URL on a local filesystem which produces the exception I've posted before

`E/mbgl: {Map}[Setup]: loading style failed: unexpected url: file:/storage/emulated/0/Download/test_aoi_style.json`

```
public void showAoiGeoJsonLayers() {
    if (mMap == null) {
        return;
    }
    File internalFile = new File(
            Environment.getExternalStoragePublicDirectory(
                Environment.DIRECTORY_DOWNLOADS
            ), "test_aoi_style.json"
    );
    Log.i(TAG, internalFile.toString());
    Log.i(TAG, "file exists: " + internalFile.exists());
    try {
        String internalFileUrl = internalFile.toURI().toURL().toString();
        Log.i(TAG, "URL: " + internalFileUrl);
        mMap.setStyleUrl(internalFileUrl);
    } catch (MalformedURLException e) {
        Log.e(TAG, e.getStackTrace().toString());
    }

}
```

bleege:
The forthcoming Style API is going to be the preferred solution for the SDK.  Thank you though for the PR @smartkiwi.

PR #4771, opened 2016-04-20
[core] port the pattern-fill antialiasing fix from gl-js

mollymerp:
just like [this](https://github.com/mapbox/mapbox-gl-js/pull/2372)


mollymerp:
Render tests from Travis are here: https://mapbox.s3.amazonaws.com/mapbox-gl-native/render-tests/11966.1/index.html

It looks like this change is affecting more of the render tests than it should be. The condition I'm using to see whether to apply the outline-pattern shader is `if (properties.antialias && stroke_color == fill_color) {}`.

@jfirebaugh mentioned in chat that I should try the conditional `stroke_color[3] == -1` to test if the stroke-color property of a fill layer is not set, and if it isn't, apply the outlinepattern shader. That didn't seem to work (i.e. there was no change from the master branch in the rendering tests)

I will continue poking around, but if anyone has suggestions I'd love to hear them!

jfirebaugh:
@ansis Can you help @mollymerp here? 🙇

ansis:
The render tests look right to me. The antialiasing doesn't match -js but that's expected because -js renders antialiasing wrong in headless testing: https://github.com/mapbox/mapbox-gl-js/issues/2080

I think the failing outline test can be disabled in -native until we fix the antialising issue.

src/mbgl/renderer/painter_fill.cpp
```diff
@@ -108,6 +109,44 @@ void Painter::renderFill(FillBucket& bucket, const FillLayer& layer, const TileI
...
+                    1.0f / id.pixelsToTileUnits(imageSizeScaledB[1], state.getIntegerZoom())
+                }};
+
+                outlinePatternShader->u_offset_a = std::array<float, 2>{{offsetAx, offsetAy}};
+                outlinePatternShader->u_offset_b = std::array<float, 2>{{offsetBx, offsetBy}};
+
+                config.activeTexture = GL_TEXTURE0;
+                spriteAtlas->bind(true, glObjectStore);
+
+                setDepthSublayer(0);
```
ansis: I think this should be `setDepthSublayer(2);` like the non-pattern antialiasing call.

What this does is draws the outline at a further depth so that it doesn't get drawn on top of the inner part of the fill.

ansis:
Once https://github.com/mapbox/mapbox-gl-native/pull/4771/files#r61337105 is fixed and tests failing because of antialiasing differences are disabled I think this should be good to merge.

@jfirebaugh do you still want to review anything here?

jfirebaugh:
LGTM!

mollymerp:
Updated the render test for `test-Map.Offline` that was failing:

_old expected_ 
screen shot 2016-04-28 at 11 28 17 am

_new expected_
screen shot 2016-04-28 at 11 28 41 am

thank you to @tmpsantos for helping me trouble shoot the failing builds! 💯 

once everything is green, this is good to 🚢  @ansis ?

ansis:
:ship:!

Issue #2184, opened 2015-08-26
Use Square Leak Canary To Review Performance


bleege:
Let's use [Square Leak Canary](https://github.com/square/leakcanary) to help track memory usage / guard against leaky 1s and 0s.

https://corner.squareup.com/2015/05/leak-canary.html

ljbade:
@bleege Only thing to remember is I don't think this will catch leaks caused in the native side, or track memory used by native objects held by Java objects.

Still useful however.

tobrun:
I added LeakCanary to the sample app but it is not playing nice with the native code.
In certain cases, it shows the dialog to dump memory but it never closes.. 
Only killing the app resolves this.

![leakcanary2](https://cloud.githubusercontent.com/assets/2151639/10016505/d5523bdc-6127-11e5-90cd-cea8eaa5abff.png)
![leakcanary](https://cloud.githubusercontent.com/assets/2151639/10016506/d7a402e4-6127-11e5-9b1e-d2adfe32f999.png)



Haven't had this issue on while developing trivial apps.

incanus:
:snowflake:

ljbade:
Perhaps it doesn't know how to free non-Java threads that are created on the C++ side.

Do they have NDK support/integration that we need to enable?

ljbade:
There is an issue on this here:
https://github.com/square/leakcanary/issues/285

Looks like we would need to bring back the external storage permission as well as some permission checks for Marshmallow.

pyricau:
LeakCanary `1.4-SNAPSHOT` has a fix for this (asking for permission at runtime on M). You'll need the storage permission on your dev builds. You could also implement your own `LeakDirectoryProvider` and use a different storage.

tobrun:
I reintegrated this in the TestApp and :snowflake: problem from above seems to be resolved. 
Will PR so we can close this issue.

PR #4776, opened 2016-04-21
Change leak canary version from beta to stable

pedrocactus:
1.4-beta leak canary version is not stable. Might be safer to use the last stable version.


tobrun:
@pedrocactus thank you for reaching out and contributing! I completely agree that it would be better to use the stable version but in this particular case we had issues with the 1.3 version.Those issues were addressed in a 1.4-SNAPSHOT. Read more about this in https://github.com/mapbox/mapbox-gl-native/issues/2184#issuecomment-169682455. Since we can't go back to 1.3, going to close this PR.

PR #4780, opened 2016-04-21
[Qt] Remove duplicated mbgl::Map initialization

brunoabinader:
We were initializing `mbgl::Map` twice 😨 

/cc @tmpsantos


tmpsantos:
:+1:

Issue #1108, opened 2015-03-25
Language support


tatsvc:
How will language support work in the new SDK? Will this be the country's language by default, or can you set it to the language of the OS (this is what Apple Maps does, they have a toggle to always have it in English)? Just something to think about.

cc. @incanus @1ec5 @peterqliu

friedbunny:
As another point of comparison, Google Maps (iOS app and SDK) defaults to device language with no exposed setting or API for switching. Labels in "foreign" places will display both your device language (where available) and the appropriate local name.

![screen shot 2015-03-28 at 11 04 16 am](https://cloud.githubusercontent.com/assets/1198851/6882211/4f1b2ec6-d53a-11e4-88d8-460cbc79f7a2.png)

Meanwhile, Apple will try to draw labels in the device language and (almost always) fallback to local.

![screen shot 2015-03-28 at 11 11 56 am](https://cloud.githubusercontent.com/assets/1198851/6882254/897c87f8-d53b-11e4-832d-7651fee55cb8.png)
 
Google's data is far superior to Apple's, but both have outdated names for things like stadiums. It also appears that Google has done some machine translation of common place names. OSM has very, very few localized (non-English) names in Portland and setting a -native map to strictly Japanese results in no labels at all.

incanus:
Proper support here, including RTL and complex text shaping is a bit out of scope, so moving off of b2 milestone for now.

1ec5:
To be clear, this ticket is about dynamically showing labels in whatever system language is being used at runtime. Currently that would require a large number of styles or a large number of classes in a single style (although we don't currently expose style classes to iOS SDK clients), to be toggled manually by client code. This is because label text is baked into the Mapbox Streets source, rather than decoupled into a separate source that can be parameterized by locale.

jfirebaugh:
@1ec5 Not sure I understand your comment. Did you look at how the example works? It doesn't use multiple styles or style classes.

jfirebaugh:
Oops, I thought this was a gl-js ticket. For gl-native, the upshot is that once we have a [style API](https://github.com/mapbox/mapbox-gl-native/issues/837) in native, we can do the same as the [gl-js example](https://www.mapbox.com/mapbox-gl-js/example/language-switch/) -- changing the feature property used for the label.

1ec5:
[...]

PR #4783, opened 2016-04-21
Make iOS and OS X SDKs localizable

1ec5:
This PR makes the iOS and OS X SDKs localizable. It separates strings from logic and establishes some infrastructure for localizing the SDKs, but actually adding localizations or support for a specific localization tool is out of scope for this PR. Also out of scope for this PR is any form of map localization: as explained in #1108, the map’s localization is currently controlled by the style and source; dynamically switching languages would require #837 and potentially more language support in the source. So this PR is about the “chrome” only.

Hard-coded English strings are now wrapped in calls to `NSLocalizedStringWithDefaultValue()` that pair the English strings with unique identifiers, so that changes to the English strings don’t affect other localizations. In code shared between the iOS and OS X SDKs, the strings come from the “Foundation” table. The `NSLocalizedString()`, `NSLocalizedStringFromTable()`, and `NSLocalizedStringWithDefaultValue()` macros are redefined in NSBundle+MGLAdditions.h to look in the SDK bundle rather than the main application bundle. This redefinition has no effect on the host application, because it’s in an internal header. That header must be included in all files that contain localizable strings. See the header for a full explanation.

Removed the text from the iOS SDK compass view’s image, leaving only the compass plate. A localizable string is drawn atop it at runtime, along the lines of #949. Unlike in that PR, no translations are hard-coded (other than the default English string), and the system font is used.

Added a make rule that runs `genstrings` and ensures UTF-8 output, since `genstrings` insists on UTF-16. The encoding is converted in place using the built-in `textutil` utility. The workflow is that, whenever you add a string that should be localized, you’ll use the `NSLocalizedStringWithDefaultValue()` macro (making sure to import NSBundle+MGLAdditions.h), run `make genstrings`, and commit. Eventually, if we start using a localization service, we could use Xcode or `xcodebuild` to export an XLIFF.

With iOS SDK resources coming from three different directories, the resource-copying part of package.sh has gotten unwieldy. Instead of copying individual resource files around to build the static framework, package.sh now moves Mapbox.bundle into the appropriate directory. Mapbox.bundle is produced by a new build target. Now, when you add a new SDK resource, make sure to add it to both the “dynamic” and “bundle” targets in Xcode.

Replaced `$BUNDLE_RESOURCES` and `$PLACE_RESOURCE_BUNDLES_OUTSIDE_FRAMEWORK` with a single variable, `$SELF_CONTAINED`, that determines whether the output static framework is self-contained – that is, whether it contains Mapbox.bundle, Settings.bundle, and LICENSE.md. Eliminated the ability to produce a static framework that has the same layout as a dynamic framework. Now, resources are always in Mapbox.bundle; self-containment only determines where that bundle resides.

Added a separate source Info.plist to be placed in both the static framework and its resource bundle, but only after we give the build system a chance to expand variables within it. Now nearly all the Info.plist keys are filled in by the build system rather than using `plutil` in package.sh.

~~Implemented a localizable `CLLocationCoordinate2D` formatter in iosapp for use with dropped pin callouts.~~ (Split out into #4802.) Rely on subclasses of MGLPointAnnotation instead of annotation titles to distinguish between different kinds of point annotations.

* [x] Adopt `NSLocalizedString()` throughout shared “Foundation” code
* [x] Improve localizability of coordinate and direction formatters
* [x] Adopt `NSLocalizedString()` throughout iOS SDK
* [x] Adopt `NSLocalizedString()` throughout OS X SDK
* [ ] ~~Adopt `NSLocalizedString()` throughout iosapp~~
 * iosapp requires far more strings than the SDKs for little gain.
* [x] Pass unique identifiers into `NSLocalizedString()` and add `-a` to `genstrings`
* [x] Get strings from the SDK bundle instead of the main bundle
* [x] Add make rule to run `genstrings`
* [x] Share formatter localizations between iOS and OS X SDKs
* [x] Ensure that .strings files use UTF-8 to facilitate diffing and merging
* [x] Draw iOS compass text at runtime
* [x] Add Mapbox.bundle target
* [x] Streamline package.sh
* [x] Squash 🎃

/cc @friedbunny @boundsj


1ec5:
Translations for the coordinate formats and compass indicator are available for a wealth of languages in the CLDR package ([example](http://unicode.org/cldr/trac/browser/tags/release-29/common/main/ja.xml#L8859).

friedbunny:
> Translations for the coordinate formats and compass indicator are available for a wealth of languages in the CLDR package ([example](http://unicode.org/cldr/trac/browser/tags/release-29/common/main/ja.xml#L8859)).

Coordinate format strings are definitely available (e.g. the ones you linked to, 北緯35度32分 東経135度0分 — N35°32”, E135°0”), though I don’t see the string I would expect for a compass: plain ole 北.

1ec5:
That’s true, but I don’t foresee us automatically pulling translations from CLDR, so the choice of compass text would be up to the translator. CLDR would just be a starting point or backfill.

1ec5:
Since I first filed this PR, I’ve made a number of changes tangentially related to localizability and made the OS X SDK localizable to boot. The additional localization resources forced my hand in refactoring and cleaning up package.sh. I left half a spaghetti dinner in there for anyone who’s interested in cleaning up the rest. 🍝

With our increased reliance on Xcode for project management, it isn’t clear to newcomers what each target is used for and where each build product ends up. I’ve written up a brief guide to common workflows like adding source code files and resources. Folks who’ve contributed to the iOS or OS X SDKs in the past should also take note of recent changes by reading DEVELOPING.md in the respective platform/ directories.

1ec5:
57f0a211251a9378fa004ee22b2c6f4d21e2bd59 fixes an issue causing `make genstrings` to reconvert a UTF-8-encoded file from UTF-16 to UTF-8, garbling it.

PR #4785, opened 2016-04-21
Naming regularity

jfirebaugh:
This is a boring PR that just renames a bunch of things such that names derived from the style specification rigorously follow a convention. Mostly it ensures that an `type-hyphen-property` in the style spec (e.g. `fill-color`) maps to a `typeCamelizedProperty` (e.g. `fillColor`) in mapbox-gl-native.

This will be important for the code generation part of the runtime styling API.


src/clipper/clipper.hpp
```diff
@@ -121,7 +121,7 @@ typedef void (*TZFillCallback)(IntPoint& z1, IntPoint& z2, IntPoint& pt);
 #endif
 
 enum InitOptions {ioReverseSolution = 1, ioStrictlySimple = 2, ioPreserveCollinear = 4};
-enum JoinType {jtSquare, jtRound, jtMiter};
+enum LineJoinType {jtSquare, jtRound, jtMiter};
```
jfirebaugh: Whoops, over-broad find-and-replace. I'll revert this part.

PR #4789, opened 2016-04-21
Constrain zoom level when selecting offline pack

### LINKED ISSUE #4787: selecting an offline pack should show the map at the minimum zoom level of the pack
When you select an offline pack in osxapp, it opens a new window with the map bounds set to the bounds of the offline pack. This is good, but even more good would be to ensure that the zoom level is at least the minimum zoom level of the offline pack, since there won't be any tiles in the pack below that level.

1ec5:
When selecting an offline pack for display in iosapp or osxapp, constrain the zoom level to the pack’s minimum and maximum zoom levels.

Fixes #4787.

/cc @jfirebaugh


jfirebaugh:
👌 🙇

1ec5:
e898b0a459c5db5149e5b774cc22add5aceee983 follows up with a fix for an issue in which new windows would always be constrained to the last selected offline pack’s minimum and maximum zoom levels.

PR #4790, opened 2016-04-22
cameraThatFitsCoordinateBounds

1ec5:
Added an API to get a camera that you can pass into `-[MGLMapView setCamera:]` that fits the given coordinate bounds, by analogy with `-[MKMapView regionThatFits:]` or `-[MKMapView mapRectThatFits:edgePadding:]`.

Added `mbgl::Map::getCameraOptions()` for getting the current camera options in fewer lines of code – less opportunity to forget a field.

@friedbunny, can you double-check that I haven’t flubbed anything with respect to edge insets / padding?

### DIFF
--- a/include/mbgl/map/map.hpp
+++ b/include/mbgl/map/map.hpp
@@ -69,6 +69,7 @@ public:
     bool isPanning() const;
 
     // Camera
+    CameraOptions getCameraOptions(optional<EdgeInsets>) const;
     void jumpTo(const CameraOptions&);
     void easeTo(const CameraOptions&, const AnimationOptions&);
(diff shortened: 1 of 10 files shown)


1ec5:
Note that the result of `-cameraThatFitsCoordinateBounds:` still isn’t quite perfect when the map is tilted, due to #2259.

1ec5:
df0d1187da91537ef55fefc154159f1a6fd747ac fixes a regression this change introduced that prevented the map’s previous direction and pitch from being persisted by state restoration. Thankfully, centralizing this code means we don’t have to do so much of this conversion anymore.

Issue #4352, opened 2016-03-16
Animating map view contraints is jerky and does not update properly


friedbunny:
The proper way to animate MGLMapView is to modify its auto layout constraints (and not the frame), but this doesn’t behave correctly:
- White gaps at top and bottom jump as the map frame instantly shrinks to fit, rather than animate to the new, smaller constraint. The logo and attribution subviews do animate correctly, which indicates this is an issue with the GLKView that provides the map itself.
- User location annotation does not move, it just sits around waiting to be in the correct spot again.

![animating-constraint-jumping](https://cloud.githubusercontent.com/assets/1198851/13827612/fc0ed78e-eb93-11e5-9fba-411f4b7c0cae.gif)

Preview of the [test project](https://github.com/mapbox/mapbox-gl-native/files/176840/animationtest.zip):

![screen shot 2016-03-16 at 3 10 26 pm](https://cloud.githubusercontent.com/assets/1198851/13827615/feb32922-eb93-11e5-8e25-b4e7831fa6bd.png)

``` swift
import UIKit
import Mapbox

class ViewController: UIViewController {

    @IBOutlet weak var mapView: MGLMapView!
    @IBOutlet weak var heightConstraint: NSLayoutConstraint!

    override func viewDidLoad() {
        super.viewDidLoad()
        mapView.showsUserLocation = true
    }

    @IBAction func animate(sender: AnyObject) {
        UIView.animateWithDuration(1.5, delay: 0, options: .CurveEaseOut, animations: {
            self.mapView.userInteractionEnabled = false

            self.heightConstraint.constant = self.heightConstraint.constant > 200 ? 200 : self.view.frame.height/2

            self.view.layoutIfNeeded()
        }, completion: { finished in
            self.mapView.userInteractionEnabled = true
        })
    }

}
```

/cc @1ec5 @boundsj

harri121:
+1

harri121:
Is there a workaround?

1ec5:
I’m not sure about a workaround, but I suspect the problem may be that `-layoutSubviews` only gets called at the beginning of the animation, so `mbgl::Map::update()` gets called only once with the final size. Meanwhile, the user dot appears to stay stationary because it’s actually animating in the opposite direction to compensate. (If the map view is flush with the bottom of the view controller, showing and hiding the bottom bar will cause the user dot to visibly slide into place.)

1ec5:
[...]

Issue #3276, opened 2015-12-14
Expose a generic Android View Marker API


tobrun:
Current Marker API on android is fairly limited because Markers are Gl-drawn components. I'm hearing about requests for animating markers or having the ability to do more typical Android SDK visual things with it (animations, selectors etc.).

`Concurrent InfoWindows` from #3127 and `UserLocationView` showcased that we are able to sync Android Views with the underlying MapView. The logic is their but we do not expose a generic system. 

## API proposal:
For native annotations views in the Android Mapbox SDK. I don't see any reason to stray far from the adapter concept what an Android developer uses on daily basis. Combining this with the existing Mapbox API for adding annotations will look like:

```java
 // combined API
 addMarker();
 addMarkers();
 removeMarker();
 removeMarkers();
 selectMarker();
```

```java
 // new API
 setMarkerViewAdapter();
 setOnMarkerViewClickListener();
 setMarkerViewItemAnimation();
```

### Example API

Integration of the API will follow the same path as the current Marker API.

```java
  // Add country markers
  List<BaseMarkerOptions> countries = new ArrayList<>();
  countries.add(new CountryMarkerOptions().title("China").abbrevName("ch").flagRes(R.drawable.ic_china).position(new LatLng(31.230416, 121.473701)));
  countries.add(new CountryMarkerOptions().title("United States").abbrevName("us").flagRes(R.drawable.ic_us).position(new LatLng(38.907192, -77.036871)));
  countries.add(new CountryMarkerOptions().title("Brazil").abbrevName("br").flagRes(R.drawable.ic_brazil).position(new LatLng(-15.798200, -47.922363)));
  countries.add(new CountryMarkerOptions().title("Germany").abbrevName("de").flagRes(R.drawable.ic_germany).position(new LatLng(52.520007, 13.404954)));
  mapboxMap.addMarkers(countries);
```

On top of that it will expose some new APIs to make ViewMarkers possible:

```java
  // Add view marker adapter
  mapboxMap.setMarkerViewAdapter(new CountryAdapter(this));

  // Add a view marker click listener
  mapboxMap.setOnMarkerViewClickListener(new MapboxMap.OnMarkerViewClickListener() {
            @Override
            public void onMarkerClick(@NonNull Marker marker, @NonNull View view) {
                Log.d(MapboxConstants.TAG, "Country clicked " + ((CountryMarker) marker).getAbbrevName());
            }
        });
```

[...]

PR #4801, opened 2016-04-22
[ios] Add platform annotation views

### LINKED ISSUE #1784: iOS annotation views
Our current iOS user dot is a custom view (made of three parts; the dot and, mutually exclusively, either the pulsing halo or the accuracy radius). 

ios simulator screen shot jun 23 2015 3 39 49 pm

We also do this with the (singular) callout view: 

ios simulator screen shot jun 23 2015 3 40 25 pm

Though we currently punt on _tracking_ this callout view to the map pan/zoom since there are margin and layout issues that are more complex than the mere user dot. 

We eventually want to support other, user-provided views atop the GL map view for several reasons: 

- Users are already creating views in projects and want to use them on maps. 
- Follows expected MapKit APIs. 
- Allows use of existing libraries expecting to work with views such as [kingpin](https://github.com/itsbonczek/kingpin) (related: https://github.com/mapbox/mapbox-gl-native/issues/320). 
- Native views allow animatability with Cocoa APIs (`CABasicAnimation`, `UIView` animation blocks, etc.)

Somewhat blocked by / could be improved upon with https://github.com/mapbox/mapbox-gl-native/issues/1125, which I think is at the core of a successful implementation. 

Tasks: 

- Memory-efficient reuse identifier-based queueing approach for view vending from delegate (`-[MGLMapViewDelegate mapView:viewForAnnotation:]`, `-[MGLMapView dequeueReusableAnnotationViewWithIdentifier]`, etc.)

- Improved syncing/tracking performance in #1125 as mentioned above, including research into possible frame-based view/layer OpenGL compositing/texture uploading. 

- Branching current tap gesture hit testing to account for native-side views as well as core-side symbols.

boundsj:
Implement platform annotation view (UIViews) for iOS.

## API Proposal

For native annotation views in the Mapbox SDK, I don't see any reason to stray too far from what a developer coming from the Apple's MapKit might expect. Combined with the existing Mapbox API for adding annotations this would look like:

```
// existing annotation model management API
-addAnnotation(s):
-removeAnnotation(s):
-selectAnnotation:animated:
...

// a new method to dequeue a reusable annotation view
-dequeueReusableAnnotationViewWithIdentifier:

// a new method in MGLMapViewDelegate protocol
-mapView:viewForAnnotation:
```

Below is an example of how the new methods noted above might be used in a future implementation. Note that MGLMapView will expect its delegate to supply a UIView that is a subclass of a new `MGLAnnotationView`. `MGLAnnotationView` will hold additional meta information about the annotation, will allow for reuse of code to deal with touch interactions, and will help when view reuse is eventually
added.

```
// In the MGLMapView's delegate implementation
func mapView(mapView: MGLMapView, viewForAnnotation annotation: MGLAnnotation) -> MGLAnnotationView? {
    var customAnnotationView = mapView.dequeueReusableAnnotationViewWithIdentifier("reuse-me") as? SomeMGLAnnotationViewSubclass
   
    if ((customAnnotationView == nil)) {
        customAnnotationView = SomeMGLAnnotationViewSubclass(annotation: annotation, reuseIdentifier: "reuse-me")
    } else {
        customAnnotationView?.annotation = annotation
    }
    
    return customAnnotationView
}
```

### Next steps:

- [x] Investigate where imageForAnnotation is currently called in map view -- this might inform what happens next...
- [x] Investigate Minh's PR for updating an annotation w/o removing it
- [x] Introduce new protocol method for obtaining a view
- [x] Introduce logic to use native pin if no view or image, use image if image, use view if view
- [x] Support for existing interaction API
- [x] Introduce view reuse queue
- [x] Explore using protocol to handle interactions for MGLAnnotationView annotations

cc @1ec5 @friedbunny @tobrun

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -23,3 +23,4 @@ xcuserdata
 /node_modules
 /platform/ios/benchmark/assets/glyphs/DIN*
 /platform/ios/benchmark/assets/tiles/mapbox.mapbox-terrain-v2,mapbox.mapbox-streets-v6
+**/token
--- a/platform/ios/Mapbox.playground/Contents.swift
+++ b/platform/ios/Mapbox.playground/Contents.swift
@@ -1,0 +1,133 @@
+import UIKit
+import XCPlayground
+import Mapbox
+
+let width: CGFloat = 700
+let height: CGFloat = 800
+
+//: A control panel
+let panelWidth: CGFloat = 200
+let panel = UIView(frame: CGRect(x: width - panelWidth, y: 0, width: 200, height: 100))
+panel.alpha = 0.8
+panel.backgroundColor = UIColor.whiteColor()
+let deleteSwitchLabel = UILabel(frame: CGRect(x: 0, y: 0, width: 100, height: 30))
+deleteSwitchLabel.adjustsFontSizeToFitWidth = true
+deleteSwitchLabel.text = "Delete Markers"
+let deleteMarkerSwitchView = UISwitch(frame: CGRect(x: panelWidth - panelWidth / 2.0, y:0, width: 100, height: 50))
+panel.addSubview(deleteSwitchLabel)
+panel.addSubview(deleteMarkerSwitchView)
+let hideSwitchLabel = UILabel(frame: CGRect(x: 0, y: 30, width: 100, height: 30))
+hideSwitchLabel.adjustsFontSizeToFitWidth = true
+hideSwitchLabel.text = "Hide Markers"
+let hideMarkerSwitchView = UISwitch(frame: CGRect(x: panelWidth - panelWidth / 2.0, y: 30, width: 100, height: 50))
+panel.addSubview(hideSwitchLabel)
+panel.addSubview(hideMarkerSwitchView)
+
+//: # Mapbox Maps
+
+/*:
+ Put your access token into a plain text file called `token`. Then select the “token” placeholder below, go to Editor ‣ Insert File Literal, and select the `token` file.
+ */
+var accessToken = try String(contentsOfURL: <#token#>)
+MGLAccountManager.setAccessToken(accessToken)
+
+class PlaygroundAnnotationView: MGLAnnotationView {
+    
+    override func prepareForReuse() {
+        hidden = hideMarkerSwitchView.on
+    }
+    
+}
+
+//: Define a map delegate
+
+class MapDelegate: NSObject, MGLMapViewDelegate {
+    
+    var annotationViewByAnnotation = [MGLPointAnnotation: PlaygroundAnnotationView]()
+    
+    func mapView(mapView: MGLMapView, viewForAnnotation annotation: MGLAnnotation) -> MGLAnnotationView? {
+        
+        var annotationView = mapView.dequeueReusableAnnotationViewWithIdentifier("annotation") as? PlaygroundAnnotationView
+        
+        if (annotationView == nil) {
+            let av = PlaygroundAnnotationView(reuseIdentifier: "annotation")
+            av.frame = CGRect(x: 0, y: 0, width: 30, height: 30)
+            let centerView = UIView(frame: CGRectInset(av.bounds, 3, 3))
+            centerView.backgroundColor = UIColor.whiteColor()
+            av.addSubview(centerView)
+            av.backgroundColor = UIColor.purpleColor()
+            annotationView = av
+        } else {
+            annotationView!.subviews.first?.backgroundColor = UIColor.greenColor()
+        }
+       
+        annotationViewByAnnotation[annotation as! MGLPointAnnotation] = annotationView
+        
+        return annotationView
+    }
+    
+    func mapView(mapView: MGLMapView, didSelectAnnotation annotation: MGLAnnotation) {
+        let pointAnnotation = annotation as! MGLPointAnnotation
+        let annotationView: PlaygroundAnnotationView  = annotationViewByAnnotation[pointAnnotation]!
+        
+        for view in annotationViewByAnnotation.values {
+            view.layer.zPosition = -1
+        }
+        
+        annotationView.layer.zPosition = 1
+        
+        UIView.animateWithDuration(1.25, delay: 0, usingSpringWithDamping: 0.4, initialSpringVelocity: 0.6, options: .CurveEaseOut, animations: {
+            annotationView.transform = CGAffineTransformMakeScale(1.8, 1.8)
+        }) { _ in
+            annotationView.transform = CGAffineTransformMakeScale(1, 1)
+            
+            if deleteMarkerSwitchView.on {
+                mapView.removeAnnotation(pointAnnotation)
+                return
+            }
+            
+            if hideMarkerSwitchView.on {
+                annotationView.hidden = true
+            }
+        }
+    }
+    
+    func handleTap(press: UILongPressGestureRecognizer) {
+        let mapView: MGLMapView = press.view as! MGLMapView
+        
+        if (press.state == .Recognized) {
+            let coordiante: CLLocationCoordinate2D = mapView.convertPoint(press.locationInView(mapView), toCoordinateFromView: mapView)
+            let annotation = MGLPointAnnotation()
+            annotation.title = "Dropped Marker"
+            annotation.coordinate = coordiante
+            mapView.addAnnotation(annotation)
+            mapView.showAnnotations([annotation], animated: true)
+        }
+    }
+    
+}
+
+//: Create a map and its delegate
+
+let lat: CLLocationDegrees = 37.174057
+let lng: CLLocationDegrees = -104.490984
+let centerCoordinate = CLLocationCoordinate2D(latitude: lat, longitude: lng)
+
+let mapView = MGLMapView(frame: CGRect(x: 0, y: 0, width: width, height: height))
+mapView.frame = CGRect(x: 0, y: 0, width: width, height: height)
+
+XCPlaygroundPage.currentPage.liveView = mapView
+
+let mapDelegate = MapDelegate()
+mapView.delegate = mapDelegate
+
+let tapGesture = UILongPressGestureRecognizer(target: mapDelegate, action: #selector(mapDelegate.handleTap))
+mapView.addGestureRecognizer(tapGesture)
+
+//: Zoom in to a location
+
+mapView.setCenterCoordinate(centerCoordinate, zoomLevel: 12, animated: false)
+
+//: Add control panel
+
+mapView.addSubview(panel)
--- a/platform/ios/app/MBXAnnotationView.m
+++ b/platform/ios/app/MBXAnnotationView.m
@@ -1,0 +1,28 @@
+#import "MBXAnnotationView.h"
+
+@interface MBXAnnotationView ()
+
+@property (nonatomic) UIView *centerView;
+
+@end
+
+@implementation MBXAnnotationView
+
+- (void)layoutSubviews {
+    [super layoutSubviews];
+    if (!self.centerView) {
+        self.backgroundColor = [UIColor blueColor];
+        self.centerView = [[UIView alloc] initWithFrame:CGRectInset(self.bounds, 5.0, 5.0)];
+        self.centerView.backgroundColor = self.centerColor;
+        [self addSubview:self.centerView];
+    }
+}
+
+- (void)setCenterColor:(UIColor *)centerColor {
+    if (![_centerColor isEqual:centerColor]) {
+        _centerColor = centerColor;
+        self.centerView.backgroundColor = centerColor;
+    }
+}
+
+@end
--- a/platform/ios/src/MGLAnnotationView.m
+++ b/platform/ios/src/MGLAnnotationView.m
@@ -1,0 +1,40 @@
+#import "MGLAnnotationView.h"
+#import "MGLAnnotationView_Private.h"
+
+@interface MGLAnnotationView ()
+
+@property (nonatomic) id<MGLAnnotation> annotation;
+@property (nonatomic, readwrite, nullable) NSString *reuseIdentifier;
+
+@end
+
+@implementation MGLAnnotationView
+
+- (instancetype)initWithReuseIdentifier:(NSString *)reuseIdentifier
+{
+    self = [super init];
+    
+    if (self)
+    {
+        _reuseIdentifier = [reuseIdentifier copy];
+    }
+    
+    return self;
+}
+
+- (void)prepareForReuse
+{
+    // Intentionally left blank. The default implementation of this method does nothing.
+}
+
+- (id<CAAction>)actionForLayer:(CALayer *)layer forKey:(NSString *)event
+{
+    // Allow mbgl to drive animation of this view’s bounds.
+    if ([event isEqualToString:@"bounds"])
+    {
+        return [NSNull null];
+    }
+    return [super actionForLayer:layer forKey:event];
+}
+
+@end
--- a/platform/ios/src/MGLAnnotationView_Private.h
+++ b/platform/ios/src/MGLAnnotationView_Private.h
@@ -1,0 +1,13 @@
+#import "MGLAnnotationView.h"
+#import "MGLAnnotation.h"
+
+NS_ASSUME_NONNULL_BEGIN
+
+@interface MGLAnnotationView (Private)
+
+@property (nonatomic) id<MGLAnnotation> annotation;
+@property (nonatomic, readwrite, nullable) NSString *reuseIdentifier;
+
+@end
+
+NS_ASSUME_NONNULL_END
--- a/platform/ios/src/MGLMapView.mm
+++ b/platform/ios/src/MGLMapView.mm
@@ -37,6 +37,7 @@
 #import "MGLUserLocationAnnotationView.h"
 #import "MGLUserLocation_Private.h"
 #import "MGLAnnotationImage_Private.h"
+#import "MGLAnnotationView_Private.h"
 #import "MGLMapboxEvents.h"
 #import "MGLCompactCalloutView.h"
 
@@ -167,6 +168,8 @@ public:
     /// The annotation’s image’s reuse identifier.
     NSString *imageReuseIdentifier;
     MGLAnnotationAccessibilityElement *accessibilityElement;
+    MGLAnnotationView *annotationView;
+    NSString *viewReuseIdentifier;
 };
 
 /** An accessibility element representing the MGLMapView at large. */
@@ -240,10 +243,12 @@ public:
     BOOL _opaque;
 
     NS_MUTABLE_ARRAY_OF(NSURL *) *_bundledStyleURLs;
-    
+   
     MGLAnnotationContextMap _annotationContextsByAnnotationTag;
     /// Tag of the selected annotation. If the user location annotation is selected, this ivar is set to `MGLAnnotationTagNotFound`.
     MGLAnnotationTag _selectedAnnotationTag;
+    NS_MUTABLE_DICTIONARY_OF(NSString *, NS_MUTABLE_ARRAY_OF(MGLAnnotationView *) *) *_annotationViewReuseQueueByIdentifier;
+    
     BOOL _userLocationAnnotationIsSelected;
     /// Size of the rectangle formed by unioning the maximum slop area around every annotation image.
     CGSize _unionedAnnotationImageSize;
@@ -272,6 +277,8 @@ public:
     BOOL _delegateHasLineWidthsForShapeAnnotations;
     
     MGLCompassDirectionFormatter *_accessibilityCompassFormatter;
+    
+    CGSize _largestAnnotationViewSize;
 }
 
 #pragma mark - Setup & Teardown -
@@ -404,6 +411,7 @@ mbgl::Duration MGLDurationInSeconds(NSTimeInterval duration)
     // Set up annotation management and selection state.
     _annotationImagesByIdentifier = [NSMutableDictionary dictionary];
     _annotationContextsByAnnotationTag = {};
+    _annotationViewReuseQueueByIdentifier = [NSMutableDictionary dictionary];
     _selectedAnnotationTag = MGLAnnotationTagNotFound;
     _annotationsNearbyLastTap = {};
 
@@ -1402,6 +1410,13 @@ mbgl::Duration MGLDurationInSeconds(NSTimeInterval duration)
         {
             [self selectAnnotation:self.userLocation animated:YES];
         }
+        return;
+    }
+   
+    MGLAnnotationView *hitAnnotationView = [self annotationViewAtPoint:tapPoint];
+    if (hitAnnotationView)
+    {
+        [self selectAnnotation:hitAnnotationView.annotation animated:YES];
         return;
     }
     
@@ -2713,6 +2728,11 @@ mbgl::Duration MGLDurationInSeconds(NSTimeInterval duration)
     {
         return pair.second.annotation;
     });
+    
+    annotations.erase(std::remove_if(annotations.begin(), annotations.end(),
+                                     [](const id <MGLAnnotation> annotation) { return annotation == nullptr; }),
+                      annotations.end());
+    
     return [NSArray arrayWithObjects:&annotations[0] count:annotations.size()];
 }
 
@@ -2763,8 +2783,11 @@ mbgl::Duration MGLDurationInSeconds(NSTimeInterval duration)
 
     std::vector<mbgl::PointAnnotation> points;
     std::vector<mbgl::ShapeAnnotation> shapes;
-    NSMutableArray *annotationImages = [NSMutableArray arrayWithCapacity:annotations.count];
+    
+    NSMutableDictionary *annotationImagesForAnnotation = [NSMutableDictionary dictionary];
+    NSMutableDictionary *annotationViewsForAnnotation = [NSMutableDictionary dictionary];
 
+    BOOL delegateImplementsViewForAnnotation = [self.delegate respondsToSelector:@selector(mapView:viewForAnnotation:)];
     BOOL delegateImplementsImageForPoint = [self.delegate respondsToSelector:@selector(mapView:imageForAnnotation:)];
 
     for (id <MGLAnnotation> annotation in annotations)
@@ -2777,32 +2800,51 @@ mbgl::Duration MGLDurationInSeconds(NSTimeInterval duration)
         }
         else
         {
-            MGLAnnotationImage *annotationImage;
-            if (delegateImplementsImageForPoint)
+            MGLAnnotationView *annotationView;
+            NSString *symbolName;
+            NSValue *annotationValue = [NSValue valueWithNonretainedObject:annotation];
+            
+            if (delegateImplementsViewForAnnotation)
             {
-                annotationImage = [self.delegate mapView:self imageForAnnotation:annotation];
-            }
-            if ( ! annotationImage)
-            {
-                annotationImage = [self dequeueReusableAnnotationImageWithIdentifier:MGLDefaultStyleMarkerSymbolName];
-            }
-            if ( ! annotationImage)
-            {
-                annotationImage = self.defaultAnnotationImage;
+                annotationView = [self annotationViewForAnnotation:annotation];
+                if (annotationView)
+                {
+                    annotationViewsForAnnotation[annotationValue] = annotationView;
+                    annotationView.center = [self convertCoordinate:annotation.coordinate toPointToView:self];
+                    [self.glView addSubview:annotationView];
+                }
             }
             
-            NSString *symbolName = annotationImage.styleIconIdentifier;
-            if ( ! symbolName)
-            {
-                symbolName = [MGLAnnotationSpritePrefix stringByAppendingString:annotationImage.reuseIdentifier];
-                annotationImage.styleIconIdentifier = symbolName;
+            if ( ! annotationView) {
+                MGLAnnotationImage *annotationImage;
+                
+                if (delegateImplementsImageForPoint)
+                {
+                    annotationImage = [self.delegate mapView:self imageForAnnotation:annotation];
+                }
+                if ( ! annotationImage)
+                {
+                    annotationImage = [self dequeueReusableAnnotationImageWithIdentifier:MGLDefaultStyleMarkerSymbolName];
+                }
+                if ( ! annotationImage)
+                {
+                    annotationImage = self.defaultAnnotationImage;
+                }
+                
+                symbolName = annotationImage.styleIconIdentifier;
+                
+                if ( ! symbolName)
+                {
+                    symbolName = [MGLAnnotationSpritePrefix stringByAppendingString:annotationImage.reuseIdentifier];
+                    annotationImage.styleIconIdentifier = symbolName;
+                }
+                if ( ! self.annotationImagesByIdentifier[annotationImage.reuseIdentifier])
+                {
+                    [self installAnnotationImage:annotationImage];
+                }
+               
+                annotationImagesForAnnotation[annotationValue] = annotationImage;
             }
-            
-            if ( ! self.annotationImagesByIdentifier[annotationImage.reuseIdentifier])
-            {
-                [self installAnnotationImage:annotationImage];
-            }
-            [annotationImages addObject:annotationImage];
 
             points.emplace_back(MGLLatLngFromLocationCoordinate2D(annotation.coordinate), symbolName.UTF8String ?: "");
         }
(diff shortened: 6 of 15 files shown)


platform/ios/app/MBXAnnotationView.m
```diff
@@ -0,0 +1,78 @@
+#import "MBXAnnotationView.h"
+
+static const CGFloat size = 35.0;
+static const CGFloat padding = 7.5;
```
1ec5: Nit: Globals should be named the same way as classes, in UpperCamelCase with a prefix.

platform/ios/src/MGLMapView.mm
```diff
@@ -750,6 +753,20 @@ - (void)glkView:(__unused GLKView *)view drawInRect:(__unused CGRect)rect
...
         [self updateUserLocationAnnotationView];
+        
+        // Hide all annotation views
+        for (MGLAnnotationView *annotationView in _annotationViews) {
+            annotationView.hidden = YES;
+        }
+        
+        // Update visible annotation views
+        std::vector<MGLAnnotationTag> annotationTags = [self annotationTagsInRect:CGRectInset(self.bounds, 0, 0)];
+        for (std::size_t i = 0; i != annotationTags.size(); ++i) {
```
1ec5: A more convenient way to express this is:

```cpp
for (auto &pair : _annotationContextsByAnnotationTag) {
    // pair.first is the annotation tag
    // pair.second is the annotation context
}
```

platform/ios/src/MGLMapView.mm
```diff
@@ -750,6 +753,20 @@ - (void)glkView:(__unused GLKView *)view drawInRect:(__unused CGRect)rect
...
 
         [self updateUserLocationAnnotationView];
+        
+        // Hide all annotation views
+        for (MGLAnnotationView *annotationView in _annotationViews) {
+            annotationView.hidden = YES;
+        }
+        
+        // Update visible annotation views
+        std::vector<MGLAnnotationTag> annotationTags = [self annotationTagsInRect:CGRectInset(self.bounds, 0, 0)];
```
1ec5: `CGRectInset(self.bounds, 0, 0)` is the same as `self.bounds`.

platform/ios/src/MGLMapView.mm
```diff
@@ -750,6 +753,20 @@ - (void)glkView:(__unused GLKView *)view drawInRect:(__unused CGRect)rect
...
+        for (std::size_t i = 0; i != annotationTags.size(); ++i) {
+            MGLAnnotationContext &annotationContext = _annotationContextsByAnnotationTag[annotationTags[i]];
+            CGPoint annotationPoint = [self convertCoordinate:annotationContext.annotation.coordinate toPointToView:self.glView];
```
1ec5: Nit: We tend to use `self` as the reference view when calling the conversion methods. This is normally OK, since `glView` is constrained to fill the map view, but there are cases like #4352 where the bounds won’t quite match. We should be consistent in the meantime.

platform/ios/src/MGLMapView.mm
```diff
@@ -2447,16 +2491,23 @@ - (void)addAnnotations:(NS_ARRAY_OF(id <MGLAnnotation>) *)annotations
         
         for (size_t i = 0; i < annotationTags.size(); ++i)
         {
-            MGLAnnotationTag annotationTag = annotationTags[i];
-            MGLAnnotationImage *annotationImage = annotationImages[i];
-            annotationImage.styleIconIdentifier = @(points[i].icon.c_str());
-            id <MGLAnnotation> annotation = annotations[i];
+            id<MGLAnnotation> annotation = annotations[i];
+            NSValue *annotationValue = [NSValue valueWithNonretainedObject:annotation];
```
1ec5: The existing code is already more indirect than it should be: we should build the `MGLAnnotationContext`s upfront, in the loop up above, and store them in a `vector`. We can assume that `Map::addPointAnnotations()` adds the annotations in order, so this loop would only need to associate the new tags with their annotation contexts.

platform/ios/src/MGLMapView.mm
```diff
@@ -2594,6 +2645,12 @@ - (void)removeAnnotations:(NS_ARRAY_OF(id <MGLAnnotation>) *)annotations
         {
             continue;
         }
+       
+        MGLAnnotationContext &annotationContext = _annotationContextsByAnnotationTag[annotationTag];
```
1ec5: Always call `count()` on the `map` before indexing into it, or call `at()` instead of indexing into it directly. As written, if `_annotationContextsByAnnotationTag` has no entry for `annotationTag`, this code [will create one](http://www.cplusplus.com/reference/map/map/operator[]/) using the default constructor, throwing off logic everywhere else. ([`at()`](http://www.cplusplus.com/reference/map/map/at/) throws an exception if no entry is found.)

platform/ios/src/MGLMapView.mm
```diff
@@ -2698,17 +2755,29 @@ - (MGLAnnotationTag)annotationTagAtPoint:(CGPoint)point persistingResults:(BOOL)
...
             NSAssert(annotation, @"Unknown annotation found nearby tap");
             
-            MGLAnnotationImage *annotationImage = [self imageOfAnnotationWithTag:annotationTag];
-            if ( ! annotationImage.enabled)
+            MGLAnnotationContext annotationContext = _annotationContextsByAnnotationTag[annotationTag];
+           
+            CGRect annotationRect;
+            if (annotationContext.annotationView)
+            {
+                annotationRect = annotationContext.annotationView.frame;
```
1ec5: Real views can do their own hit-testing. We [let the user dot do its own hit testing](https://github.com/mapbox/mapbox-gl-native/blob/a4b3ff55403e648e86a8539503bf2732454a242c/platform/ios/src/MGLMapView.mm#L1294) because it’s round. An [irregularly shaped annotation view](https://www.mapbox.com/ios-sdk/examples/marker-image/) would also benefit from this treatment.

platform/ios/MGLAnnotationView.h
```diff
@@ -0,0 +1,5 @@
+#import <UIKit/UIKit.h>
```
1ec5: We always import MGLTypes.h and insert the `NS_ASSUME_NONNULL_BEGIN` and `NS_ASSUME_NONNULL_END` macros. That way you don’t have to think so much about nullability as you fill out the interface in the future.

platform/ios/src/MGLMapView.mm
boundsj: Thanks. I think that, in this case, I need the `i` since I want to take the intersection of the `annotationTagsInRect` and the annotation tags in `_annotationContextsByAnnotationTag`. I want to get the context from `_annotationContextsByAnnotationTag` with the key of `annotationTags[i]`. Only those visible contexts need to have their view update to reflect where they should now show on the map. I will use the `for (auto &pair ...` expression in my next commit for view reuse though. Please stay tuned!

platform/ios/src/MGLMapView.mm
boundsj: Understood. This is a WIP. I will be using something different than 0, 0 in a future commit. Please stay tuned.

platform/ios/src/MGLMapView.mm
boundsj: Thanks. I've updated in my next commit

platform/ios/src/MGLMapView.mm
boundsj: Agreed. I will refactor in a future commit.

platform/ios/src/MGLMapView.mm
1ec5: That would still be possible, unless I’m missing something:

```obj
MGLAnnotationContext &annotationContext = _annotationContextsByAnnotationTag[pair.first];
```

platform/ios/src/MGLMapView.mm
```diff
@@ -750,6 +753,20 @@ - (void)glkView:(__unused GLKView *)view drawInRect:(__unused CGRect)rect
         _mbglMap->render();
 
         [self updateUserLocationAnnotationView];
+        
+        // Hide all annotation views
```
1ec5: This code runs 60 times a second due to the CADisplayLink. Intuitively, I’d expect hiding, repositioning, and showing views in this relatively hot code to be a bit of a drag on performance. Can we move this into a separate method that gets called when appropriate, for example in `-notifyMapChange:`, as well as in `-observeValueForKeyPath:ofObject:change:context:` when an annotation’s `coordinate` key path changes?

platform/ios/src/MGLMapView.mm
boundsj: I'm sure I'm missing something! 😄 

But, I do want to loop over the vector result of `annotationTagsInRect:` and not the the entire contents of `_annotationContextsByAnnotationTag`. In the spirit of not conjuring up the index when I don't need it, how about:

```
std::vector<MGLAnnotationTag> annotationTags = [self annotationTagsInRect:CGRectInset(self.bounds, 0, 0)];
for(auto const& annotationTag: annotationTags)
 {
    MGLAnnotationContext &annotationContext = _annotationContextsByAnnotationTag[annotationTag];
...
```

platform/ios/src/MGLMapView.mm
boundsj: Yes. I have a task in this PR to investigate and refactor to use view hit testing. Will fix.

platform/ios/src/MGLMapView.mm
boundsj: Thanks!

platform/ios/src/MGLMapView.mm
1ec5: Oh, sorry, that’s what I meant!

platform/ios/src/MGLMapView.mm
boundsj: Definitely. I'll work on that next.

platform/ios/src/MGLMapView.mm
```diff
@@ -3931,6 +3945,64 @@ - (void)updateUserLocationAnnotationView
...
+        MGLAnnotationView *annotationView = annotationContext.annotationView;
+        NSMutableArray *annotationViewReuseQueue = [self annotationViewReuseQueueForIdentifier:annotationContext.viewReuseIdentifier];
+        
+        if (annotationView.hidden && ![annotationViewReuseQueue containsObject:annotationView])
+        {
+            [annotationViewReuseQueue addObject:annotationView];
+            annotationContext.annotationView = NULL;
+        }
+        
+        NSLog(@"================> queue: %@, count: %ld", annotationContext.viewReuseIdentifier, (long)annotationViewReuseQueue.count);
```
boundsj: oops. will remove!

platform/ios/src/MGLMapView.mm
boundsj: @1ec5 fyi https://github.com/mapbox/mapbox-gl-native/pull/4801/commits/4db0d0f82352c453ff47fae6fcf6b66f078404ec

tobrun:
@boundsj I have been running at the Android equivalent and looking into how we can efficiently manage which annotations are currently visible by using `MapChangeRegionIsChanging` map change event as seen in this PR.

I noticed the following scenario: for the following screen I received the `MapChangeRegionIsChanging` event and was able to determine the marker shown below:

device-2016-04-27-105605

When I slowly pan to the bottom, I'm not receiving the required map change event and I'm not able to determine that we have 4 point annotations in bounds instead of 1.

device-2016-04-27-105638

For the reference my [logs](https://gist.github.com/tobrun/fab1b9c7bb9144f14dff93cdd9c2a597).
In other cases of panning normally or doing a double tap zoom, I'm receiving the events as expected.

@boundsj are you seeing the same behaviour on iOS?

PS. note that when a region change is detected, I'm receiving multiple ones after each other. We might want to look into creating an idling system that will only act on one of these events and will ignore other ones for a certain timeout time.  For Android this would make sense because it would limit the amount JNI trips.

cc @1ec5

platform/ios/src/MGLAnnotationView_Private.h
```diff
@@ -0,0 +1,23 @@
...
+
+- (void)annotationView:(MGLAnnotationView *)annotationView didReceiveTapAtPoint:(CGPoint)point;
+
+@end
+
+@interface MGLAnnotationView (Private)
+
+@property (nonatomic, weak) id<MGLAnnotationViewDelegate> delegate;
+
+// Identifier used to specifiy a reuse queue for the annotation view type.
```
1ec5: Nit: `///` for single-line documentation comments.

platform/ios/src/MGLMapView.mm
```diff
@@ -194,7 +197,7 @@ @implementation MGLMapView
     MGLAnnotationContextMap _annotationContextsByAnnotationTag;
     /// Tag of the selected annotation. If the user location annotation is selected, this ivar is set to `MGLAnnotationTagNotFound`.
     MGLAnnotationTag _selectedAnnotationTag;
-    NS_MUTABLE_ARRAY_OF(MGLAnnotationView *) *_annotationViews;
+    NS_MUTABLE_DICTIONARY_OF(NSString *, NSMutableArray *) *_annotationViewReuseQueueByIdentifier;
```
1ec5: The NSMutableArray type should be further constrained to `MGLAnnotationView *`, if that’s possible with the macros we’re using.

platform/ios/src/MGLMapView.mm
```diff
@@ -189,10 +194,11 @@ @implementation MGLMapView
     BOOL _opaque;
 
     NS_MUTABLE_ARRAY_OF(NSURL *) *_bundledStyleURLs;
-    
+   
     MGLAnnotationContextMap _annotationContextsByAnnotationTag;
     /// Tag of the selected annotation. If the user location annotation is selected, this ivar is set to `MGLAnnotationTagNotFound`.
     MGLAnnotationTag _selectedAnnotationTag;
+    NS_MUTABLE_DICTIONARY_OF(NSString *, NSMutableArray *) *_annotationViewReuseQueueByIdentifier;
```
1ec5: The NSMutableArray type should be further constrained to `MGLAnnotationView *`, if that’s possible with the macros we’re using.

1ec5:
> PS. note that when a region change is detected, I'm receiving multiple ones after each other. We might want to look into creating an idling system that will only act on one of these events and will ignore other ones for a certain timeout time. For Android this would make sense because it would limit the amount JNI trips.

I’m not sure whether the multiple events in succession reflects a problem in mbgl or in the Android SDK itself, but I haven’t personally witnessed this issue for gestures or programmatic animations. The iOS SDK uses a different pattern in which there’s only one listener for map change notifications – the `mbgl::View` subclass owned by MGLMapView – rather than a separate `OnMapChangedListener` for every little thing that might need reacting to. So I’m not sure an idling system would be needed on the iOS side.

We should pay special attention to where we update the annotation layer. Currently, in this PR, all annotations are being updated in `-[MGLMapView glkView:drawInRect:]`, which gets called 60 times per second as long as the view is in the foreground. Even if the annotation layer appears to be in sync with the map, the annotations are probably bogging down the map itself, causing it to skip frames. It would be valuable to move this work to `-notifyMapChange:`, specifically in response to both `MapChangeDidFinishRenderingFrame` and `MapChangeDidFinishRenderingFrameFullyRendered`. If responding to these two notifications isn’t sufficient to ensure that the views are synchronized with the map, that’s an mbgl bug we should prioritize.

platform/ios/src/MGLMapView.mm
```diff
@@ -3850,6 +3952,75 @@ - (void)updateUserLocationAnnotationView
...
+        const bool tagIsNotVisible = visibleTags.find(annotationTag) == visibleTags.end();
+        
+        if (annotationView && tagIsNotVisible)
+        {
+            annotationView.hidden = YES;
+            
+            if (annotationContext.viewReuseIdentifier)
+            {
+                NSMutableArray *annotationViewReuseQueue = [self annotationViewReuseQueueForIdentifier:annotationContext.viewReuseIdentifier];
+                if (![annotationViewReuseQueue containsObject:annotationView])
```
1ec5: Do the semantics of the reuse queue require this search to form a pointer comparison or object comparison on each view in the reuse queue? If the latter, MGLAnnotationView should have custom implementations of `-isEqual:` and `-hash`.

platform/ios/Mapbox.playground/Contents.swift
```diff
@@ -0,0 +1,33 @@
+//: # Mapbox Maps
+
+import UIKit
+import XCPlayground
+import Mapbox
+
+let width = 400
+let height = 400
+
+MGLAccountManager.setAccessToken("<Mapbox Access Token>")
```
1ec5: Let’s use `<#your access token#>` so that Xcode displays the string as a placeholder token and warns about it.

platform/ios/app/MBXAnnotationView.m
```diff
@@ -2,24 +2,20 @@
...
-    }
-    return self;
-}
-
 - (void)layoutSubviews {
     [super layoutSubviews];
-    self.centerView.layer.cornerRadius = CGRectGetWidth(self.centerView.bounds) / 2.0;
+    if (!self.centerView) {
+        self.backgroundColor = [UIColor blueColor];
+        self.centerView = [[UIView alloc] initWithFrame:CGRectMake(5,5, CGRectGetWidth(self.bounds) - 10, CGRectGetWidth(self.bounds) - 10)];
```
1ec5: This can be simpler if you inset `self.bounds` using `CGRectInset()`.

platform/ios/app/MBXAnnotationView.m
```diff
@@ -0,0 +1,28 @@
...
+
+@end
+
+@implementation MBXAnnotationView
+
+- (void)layoutSubviews {
+    [super layoutSubviews];
+    if (!self.centerView) {
+        self.backgroundColor = [UIColor blueColor];
+        self.centerView = [[UIView alloc] initWithFrame:CGRectMake(5,5, CGRectGetWidth(self.bounds) - 10, CGRectGetWidth(self.bounds) - 10)];
```
1ec5: This can be simpler if you inset `self.bounds` using `CGRectInset()`.

platform/ios/src/MGLAnnotationView_Private.h
```diff
@@ -0,0 +1,23 @@
...
+@property (nonatomic, copy, nullable) NSString *reuseIdentifier;
+
+@property (nonatomic) id<MGLAnnotation> annotation;
```
1ec5: MKAnnotationView exposes this property publicly. Is there a reason for keeping it private?

platform/ios/src/MGLAnnotationView_Private.h
```diff
@@ -0,0 +1,23 @@
...
+
+@protocol MGLAnnotationViewDelegate <NSObject>
+
+- (void)annotationView:(MGLAnnotationView *)annotationView didReceiveTapAtPoint:(CGPoint)point;
+
+@end
+
+@interface MGLAnnotationView (Private)
+
+@property (nonatomic, weak) id<MGLAnnotationViewDelegate> delegate;
```
1ec5: Due to this delegate, an MGLAnnotationView instance can only be associated with one MGLMapView at a time. Instead of a dedicated delegate property, could we assume (or assert) that view’s superclass is an MGLMapView? What does MapKit do when you try to insert an MKAnnotationView into a generic UIView’s subview list?

platform/ios/src/MGLMapView.mm
```diff
@@ -2496,6 +2536,22 @@ - (MGLAnnotationImage *)defaultAnnotationImage
     return annotationImage;
 }
 
+- (MGLAnnotationView *)annotationViewForAnnotation:(id<MGLAnnotation>)annotation
+{
+    MGLAnnotationView *annotationView = (MGLAnnotationView *)[self.delegate mapView:self viewForAnnotation:annotation];
```
1ec5: This cast and the assertion below are unnecessary, because `-[MGLMapViewDelegate mapView:viewForAnnotation:]` returns an MGLAnnotationView.

platform/ios/src/MGLAnnotationView.m
```diff
@@ -0,0 +1,72 @@
...
+    {
+        [self commonInit];
+    }
+    return self;
+}
+
+- (void)commonInit
+{
+    UITapGestureRecognizer *tapGestureRecognizer = [[UITapGestureRecognizer alloc] initWithTarget:self action:@selector(didTapView:)];
+    [self addGestureRecognizer:tapGestureRecognizer];
```
1ec5: I suspect that having a separate gesture recognizer for each annotation would have a nontrivial impact on performance. We already have a gesture recognizer for the overall MGLMapView that calls `-handleSingleTapGesture:`. Can we call `-hitTest:` on each of the annotations there?

platform/ios/src/MGLMapView.mm
```diff
@@ -2932,17 +2929,34 @@ - (void)selectAnnotation:(id <MGLAnnotation>)annotation animated:(BOOL)animated
...
-    // The annotation can’t be selected if no part of it is hittable.
-    CGRect positioningRect = [self positioningRectForCalloutForAnnotationWithTag:annotationTag];
+    MGLAnnotationContext &annotationContext = _annotationContextsByAnnotationTag.at(annotationTag);
+    CGRect positioningRect;
+   
+    if (annotationContext.annotationView)
+    {
+        // Annotations represented by views use the view frame as the positioning rect.
+        positioningRect = annotationContext.annotationView.frame;
+        positioningRect.origin.y -= 0.5;
```
1ec5: Can you add a comment explaining why this line is needed?

platform/ios/src/MGLMapView.mm
```diff
@@ -2854,17 +2929,34 @@ - (void)selectAnnotation:(id <MGLAnnotation>)annotation animated:(BOOL)animated
...
+    }
+    
+    MGLAnnotationContext &annotationContext = _annotationContextsByAnnotationTag.at(annotationTag);
+    CGRect positioningRect;
+   
+    if (annotationContext.annotationView)
+    {
+        // Annotations represented by views use the view frame as the positioning rect.
+        positioningRect = annotationContext.annotationView.frame;
+        positioningRect.origin.y -= 0.5;
```
1ec5: Can you add a comment explaining why this line is needed?

platform/ios/src/MGLMapView.mm
```diff
@@ -2932,17 +2929,34 @@ - (void)selectAnnotation:(id <MGLAnnotation>)annotation animated:(BOOL)animated
...
+     // The client can request that any annotation be selected (even ones that are offscreen).
+     // The annotation can’t be selected if no part of it is hittable.
     if ( ! CGRectIntersectsRect(positioningRect, self.bounds) && annotation != self.userLocation)
     {
         return;
     }
 
     self.selectedAnnotation = annotation;
-
+
```
1ec5: Nit: stray whitespace. (I’m fine with whitespace on blank lines, but it should line up with the surrounding code.)

boundsj:
> @tobrun When I slowly pan to the bottom, I'm not receiving the required map change event and I'm not able to determine that we have 4 point annotations in bounds instead of 1. are you seeing the same behaviour on iOS?

I see that all `MapChangeRegionIsChanging` events are received for all manner of gestures including a slow pan to the bottom.

platform/ios/Mapbox.playground/Contents.swift
```diff
@@ -0,0 +1,123 @@
...
+import XCPlayground
+import Mapbox
+
+//: # Mapbox Maps
+
+/*:
+### Setting the Access Token
+ 
+Add a PlaygroundInfo.plist file to the Resources folder in this playground. Then add an 
+access token key value pair:
```
1ec5: Here’s another approach that would have the same effect of placing an ignored file in the Resources folder. I’m not sure if it’s more or less convenient, but perhaps we could use this approach for other purposes in the future:

```swift
/*:
Put your access token into a plain text file called `token`. Then select the “token” placeholder below, go to Editor ‣ Insert File Literal, and select the `token` file.
*/
let accessToken = String(contentsOfURL: <#token#>)
```

tobrun:
@boundsj: @1ec5 responded on above in https://github.com/mapbox/mapbox-gl-native/issues/3276#issuecomment-215108070 and everything is working as on iOS.

platform/ios/src/MGLMapView.mm
boundsj: It's a pointer comparison. This guard just protects against adding a reference to any exact same view object that was previously added to the queue. I used an array for the queue since that API feels better (as a queue) in most places -- except here where a set might make things clearer. In any case, the intention is to capture any view object that goes offscreen in the appropriate queue so that it can be reused

platform/ios/src/MGLAnnotationView_Private.h
boundsj: Good catch! thanks

boundsj: Like MKAnnotationView, we can expose it publicly, readonly and keep a writable internal available. since the map view needs to write to it.

platform/ios/src/MGLMapView.mm
boundsj: I did not intend to commit this. Will remove.

platform/ios/src/MGLAnnotationView_Private.h
boundsj: I checked and an MKAnnotationView can be added to any arbitrary view. But, I think it is a great idea to remove the constraint of a single delegate. In 984cfb828f9028c2cb9d95e534c8f661cdb1f894 I was able to remove the need for the delegate entirely since your previous suggestion of a hit test in the map view's gesture recognizer is now implemented (and the only need for delegation so far was for the tap interaction). If the need for a delegate comes back for any other reason though, I will go with checking that the containing view is a mapview and casting as such.

platform/ios/src/MGLMapView.mm
```diff
@@ -2727,6 +2727,11 @@ - (void)removeStyleClass:(NSString *)styleClass
     {
         return pair.second.annotation;
     });
+    
+    annotations.erase(std::remove_if(annotations.begin(), annotations.end(),
+                                     [](const id <MGLAnnotation> annotation) { return annotation == nullptr; }),
+                      annotations.end());
+    
     return [NSArray arrayWithObjects:&annotations[0] count:annotations.size()];
```
boundsj: @1ec5 @friedbunny please note this fix for a crash I noticed that I think is already possible in master. Repro is to use iosapp, show the callout on the userdot, then add 100 annotations (which calls `annotations`). Although I think this fix
may be worth keeping, I wanted to call it out since I was thinking it might illustrate a slightly deeper issue (i.e. weird representation of userdot in the _annotationContextsByAnnotationTag map).

1ec5: That map should never contain the user location annotation, since it never gets a tag from mbgl.

Can you assert instead of removing the annotation in this case? I think you've uncovered a case where we're using [subscript] notation on a std::map without first checking for existence – that creates a bogus entry in the map, which is not what you'd expect.

platform/ios/src/MGLMapView.mm
```diff
@@ -278,6 +278,9 @@ @implementation MGLMapView
     BOOL _delegateHasLineWidthsForShapeAnnotations;
     
     MGLCompassDirectionFormatter *_accessibilityCompassFormatter;
+    
+    CGFloat _largestAnnotationViewHeight;
+    CGFloat _largestAnnoationViewWidth;
```
1ec5: Let’s use a `CGSize`, as with `_unionedAnnotationImageSize` above.

platform/ios/src/MGLMapView.mm
```diff
@@ -2924,6 +2927,8 @@ - (MGLAnnotationView *)annotationViewForAnnotation:(id<MGLAnnotation>)annotation
         annotationView.reuseIdentifier = self.lastAnnotationViewReuseIdentifier;
         annotationView.annotation = annotation;
         self.lastAnnotationViewReuseIdentifier = nil;
+        _largestAnnoationViewWidth = CGRectGetWidth(annotationView.frame);
+        _largestAnnotationViewHeight = CGRectGetHeight(annotationView.frame);
```
1ec5: Do we need to worry about the view’s `alignmentRectInsets`?

platform/ios/src/MGLMapView.mm
```diff
@@ -4393,7 +4400,7 @@ - (void)updateAnnotationViews
         }
        
         annotationContext.annotationView.center = annotationPoint;
-        annotationContext.annotationView.hidden = NO;
+        annotationContext.annotationView.hidden = annotationContext.annotationView.hidden ?: NO;
```
1ec5: `-[UIView isHidden]` returns a `BOOL`, so the `?: NO` has no effect.

.gitignore
```diff
@@ -23,3 +23,4 @@ xcuserdata
 /node_modules
 /platform/ios/benchmark/assets/glyphs/DIN*
 /platform/ios/benchmark/assets/tiles/mapbox.mapbox-terrain-v2,mapbox.mapbox-streets-v6
+token
```
1ec5: Use a relative path instead of a file name so other files that happen to be named “token” don’t get ignored.

platform/ios/src/MGLMapView.mm
```diff
@@ -2864,6 +2918,22 @@ - (MGLAnnotationImage *)defaultAnnotationImage
...
+{
+    MGLAnnotationView *annotationView = [self.delegate mapView:self viewForAnnotation:annotation];
+    
+    if (annotationView)
+    {
+        annotationView.reuseIdentifier = self.lastAnnotationViewReuseIdentifier;
+        annotationView.annotation = annotation;
+        self.lastAnnotationViewReuseIdentifier = nil;
+        _largestAnnoationViewWidth = CGRectGetWidth(annotationView.frame);
+        _largestAnnotationViewHeight = CGRectGetHeight(annotationView.frame);
```
1ec5: Do we need to worry about the view’s `alignmentRectInsets`?

platform/ios/src/MGLMapView.mm
```diff
@@ -4231,6 +4361,84 @@ - (void)updateUserLocationAnnotationView
...
+                {
+                    [self.glView addSubview:annotationView];
+                }
+                
+                annotationContext.annotationView = annotationView;
+            }
+        }
+       
+        annotationContext.annotationView.center = annotationPoint;
+        annotationContext.annotationView.hidden = annotationContext.annotationView.hidden ?: NO;
```
1ec5: `-[UIView isHidden]` returns a `BOOL`, so the `?: NO` has no effect.

platform/ios/src/MGLMapView.mm
boundsj: Good catch @1ec5. Also I was not halving the dimensions (the buffer only needs to be half the size of the width or height). This works better:

```
CGRect bounds = UIEdgeInsetsInsetRect({ CGPointZero, annotationView.frame.size }, annotationView.alignmentRectInsets);
_largestAnnotationViewSize = CGSizeMake(bounds.size.width / 2.0, bounds.size.height / 2.0);
```

platform/ios/src/MGLMapView.mm
boundsj: oops, I did not mean to commit this. I was experimenting with ways to maintain the hidden state of a view for an application developer but I eventually came to the conclusion that it is best to let the developer have complete control over hidden while the view is visible. When it goes off screen we enqueue / hide it. When it is dequeued from the reuse queue we unhide it (an application developer can re-hide it if required when they get it back from the reuse queue). Tail work might be to investigate why `annotationTagsInRect` (and the underlying boost query) does not return some tags that are definitely visible when the map is tilted  (which is why we are currently hiding as a workaround)

1ec5: > Tail work might be to investigate why `annotationTagsInRect` (and the underlying boost query) does not return some tags that are definitely visible when the map is tilted  (which is why we are currently hiding as a workaround)

When the map is tilted, the visible region isn’t really a rectangle: https://github.com/mapbox/mapbox-gl-native/issues/2259#issuecomment-149659328. Depending on the rotation, `visibleCoordinateBounds` may end up being inscribed in or circumscribed around the actual visible region. I think this PR can land without handling that situation perfectly, but we’ll definitely want to revisit before release.

platform/ios/src/MGLMapView.mm
```diff
@@ -4364,56 +4364,38 @@ - (void)updateAnnotationViews
...
+            MGLAnnotationView *annotationView = [self annotationViewForAnnotation:annotationContext.annotation];
             if (annotationView)
             {
                 // If the annotation view has no superview it means it was never used before so add it
                 if (!annotationView.superview)
                 {
                     [self.glView addSubview:annotationView];
                 }
-                
+                annotationView.center = [self convertCoordinate:annotationContext.annotation.coordinate toPointToView:self];;
```
1ec5: Nit: extra semicolon.

friedbunny:
popper

1ec5:
Yes! Filed some issues for tail work: #5035 #5036 #5037 #5038.

1ec5:
Updated the changelog in 54381d20a045daddac6c3bdec7eb9f1f6f4033cc.

boundsj:
I added an additional issue to the tail work pile: https://github.com/mapbox/mapbox-gl-native/issues/5059

PR #4802, opened 2016-04-22
Geometry formatters and value methods

### LINKED ISSUE #4485: Add category methods on NSValue for geometry types
MapKit has [convenient category methods](https://developer.apple.com/library/mac/documentation/Cocoa/Reference/Foundation/Classes/NSValue_Class/#//apple_ref/occ/clm/NSValue/valueWithMKCoordinate:) on NSValue for storing and retrieving `CLLocationCoordinate2D` (called “MKCoordinate” for some reason) and `MKCoordinateSpan` types. We should provide similar methods on NSValue for `CLLocationCoordinate2D` and `MGLCoordinateBounds` at the least. For `CLLocationCoordinate2D`, this is just a matter of moving [some methods](https://github.com/mapbox/mapbox-gl-native/blob/3afdfc15a182707d24d288dcd32166c4a111d732/platform/osx/app/NSValue%2BAdditions.m) from osxapp into a file shared between the iOS and OS X SDKs.

/cc @friedbunny @tmcw

1ec5:
Implemented a degree-minute-second coordinate formatter for general use. This formatter replaces redundant implementations in iosapp and osxapp. Implemented NSFormatters for formatting `CLLocationDirection` values as localizable display strings: one for compass directions, such as “south”, and one for relative [clock positions](https://en.wikipedia.org/wiki/Clock_position), such as “6 o’clock”.

These formatters are a cleaner implementation of the format strings implemented in #1496 for VoiceOver support. They may also be useful for navigation use cases.

Added category methods on NSValue for converting to and from the structure types defined in MGLGeometry.h.

Fixes #4485.

/cc @friedbunny @boundsj


friedbunny:
👍

1ec5:
The new `NSValue (MGLGeometryAdditions)` category exposed realm/jazzy#539: it isn’t possible to separately document two categories of the same external class (in this case NSValue). The workaround is to unify those categories in a dedicated header. Unfortunately, that’ll prevent us from putting the category methods close to related classes under the `custom_categories` we’ve defined for jazzy.

1ec5:
#4783 and #4819 rework the coordinate and clock direction formatters to be more flexible and localizable.

PR #4810, opened 2016-04-22
[build] Depend on directories, so that projects are regenerated when files are added or removed

jfirebaugh:
cc @1ec5 @boundsj


1ec5:
Neat!

PR #4832, opened 2016-04-25
Update frame size when changing dot/puck mode

### LINKED ISSUE #4823: Enlarge user puck’s accessibility frame
The user puck has the same accessibility frame as the user dot, even though the former appears larger and has a much larger hit target:

puck

/cc @friedbunny

friedbunny:
dotpuck

The user dot/puck was keeping whichever frame size it was initialized with. This had no apparent ill effects, except that the accessibility frame would be incorrect when the user location annotation type changed.

- Updates the accessibility frame (by proxy) when changing between dot and puck
- Keeps the same center point while updating the frame
- `updateFrameWithSize:` does not take CGSize because there is not yet
  support for non-square user location annotations

Fixes #4823.

/cc @1ec5

### DIFF
--- a/platform/ios/src/MGLUserLocationAnnotationView.m
+++ b/platform/ios/src/MGLUserLocationAnnotationView.m
@@ -181,6 +181,22 @@ const CGFloat MGLUserLocationAnnotationArrowSize = MGLUserLocationAnnotationPuck
     }
 }
 
+- (void)updateFrameWithSize:(CGFloat)size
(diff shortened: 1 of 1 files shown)


1ec5:
Nice! This also means we had a bug in 3.2.0 where the puck’s hit target was too small. This happens to be fixed on master because we now hit test against the presentation layer rather than the view itself.

PR #4836, opened 2016-04-25
use Clock::time_point::max() instead of frame time

mikemorris:
For `Style::cascade` and `Style::recalculate`, when rendering with `MapMode::Still`. Fixes a subtle race condition with animated transitions exacerbated by cd1a06c2dc209da81b3d745c088e568b3b14809f.

/cc @brunoabinader @bsudekum @miccolis


brunoabinader:
LGTM 👍

Issue #4794, opened 2016-04-22
Replace KIF with XCUITest


1ec5:
We should replace the KIF-based iOS SDK functional tests with similar tests based on XCUITest. The KIF tests have long suffered from unreliability such as #2734. Random failures forced us to disable tests one by one (e.g., #2734 and #3555) until we finally removed the entire suite from CI in #4594. We’ve had good results so far with the non-UI unit tests added to OS X in #3406 and to iOS in #4641. Hopefully XCUITest will be similarly reliable due to its tighter system integration.

Thanks to improvements in the build system in #4641, it’s quite easy to add a UI test bundle to the iOS project. We can replicate the gesture tests using [`-[XCUICoordinate pressForDuration:thenDragToCoordinate:]`](http://masilotti.com/ui-testing-cheat-sheet/#how-to-quot-pull-to-refresh-quot) and some UILabels that we continuously update with the map view’s current state. XCUITest operates on accessibility elements, whereas our existing annotation-related tests rely on access to the actual MGLMapView object, so converting them is blocked by #1496.

/cc @friedbunny @boundsj

PR #4837, opened 2016-04-25
iosapp UI tests

1ec5:
Added a UI test bundle that uses XCUITest to test iosapp. The bundle contains some basic tests of annotation functionality. Gather code coverage data in the iosapp and dynamic schemes.

Added accessibility identifiers for key controls in the SDK and iosapp.
- [x] Add UI test bundle
- [x] Suppress access token prompt if necessary
- [x] Test gestures
- [x] Test point annotations
- [ ] ~~Test shape annotations (blocked by #4838)~~
- [x] Test callout views
- [x] Test compass
- [x] Test user tracking modes
- [x] Test user dot
- [x] Test ℹ️ actions
- [ ] Test offline pack table view controller
- [x] Add UI test bundle to CI scheme

Working towards #4794.

/cc @friedbunny @boundsj

### DIFF
--- a/platform/ios/DEVELOPING.md
+++ b/platform/ios/DEVELOPING.md
@@ -91,7 +91,7 @@ To instead run the tests in Xcode:
 
 1. Run `make iproj` to set up the workspace.
 1. Change the scheme to “test (platform project)” and press Command-R to run core unit tests.
-1. Change the scheme to “CI” and press Command-U to run SDK integration tests.
+1. Change the scheme to “iosapp” and press Command-U to run SDK unit and UI tests.
 
 ## Access tokens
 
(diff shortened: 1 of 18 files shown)


1ec5:
The three existing test suites currently take 113.7 seconds. Once we implement all the tests, the tests might end up taking too long to run on Bitrise for every push, forcing us to run the tests only in the nightly workflow. The biggest time sink right now is the fact that XCUITest [relaunches the application](https://openradar.appspot.com/23115791) before every single test.

friedbunny:
> The biggest time sink right now is the fact that XCUITest relaunches the application before every single test.

That’s been my experience elsewhere, too. For now, we could write fewer, larger tests.

jfirebaugh:
Stale.

Issue #3254, opened 2015-12-11
private-implementation pattern


jfirebaugh:
With the [styles API](https://github.com/mapbox/mapbox-gl-native/issues/837), a lot more API surface area will be publicly exposed:

* `Style`
* `Source` and future subclasses for specific source types
* `StyleLayer` and subclasses for specific layer types
* Paint and layout properties for all layer types

So far, we've been using the following technique for separating public APIs from private implementation:

* Public headers live in `include`, private headers live in `src`
* Public headers use forward declarations and pointers to avoid needing to include private headers

For the styles API, I don't think this is going to be enough. For example, I don't think we want to simply make `style_layer.hpp` a public include file -- it would cascade into making other headers public as well, and expose implementation details we want to be free to change. The current `StyleLayer` and subclasses depend on rapidjson headers (not easily forward declared) and have methods that are conceptually public to other mbgl internal classes, but private to API consumers (e.g. `cascade`, `recalculate`, `createBucket`).

Therefore we need to use more heavy-duty techniques to separate the public API from private implementation. I propose we adopt a particular implementation of the [opaque pointer](https://en.wikipedia.org/wiki/Opaque_pointer) pattern, and follow this convention throughout portions of the codebase exposed to the public API:

* Public header files expose objects in the `mbgl` namespace. (Already doing this.)
* These objects hold a pointer to a corresponding instance of a class with the same name, but in the `mbgl::impl` namespace.
* All implementation code moves to the `mbgl::impl` namespace and `src/mbgl/impl` directory tree.

Example:

**include/mbgl/map.hpp**

```
namespace mbgl {
namespace impl { class Map; }

class Map {
public:
    void doSomething();

private:
    std::unique_ptr<impl::Map> impl;
};

}
```

**src/mbgl/map.cpp**

```
namespace mbgl {

void Map::doSomething() {
   impl->a.doSomething();
   impl->b.doSomethingElse();
}

}
```

**src/mbgl/impl/map.hpp**

```
#include <mbgl/map.hpp>
#include <mbgl/impl/foo.hpp>
#include <mbgl/impl/bar.hpp>

namespace mbgl {
namespace impl {

class Map {
public:
    Foo a;
    Bar b;
};

}
}
```

[...]

PR #4839, opened 2016-04-25
Add and remove layers dynamically

jfirebaugh:
This PR implements the first bit of the core [runtime styling API](https://github.com/mapbox/mapbox-gl-native/issues/837): the ability to dynamically add and remove layers.

As appropriate for a C++ API, the layers API is _strongly typed_: there are subclasses for each layer type, with correctly-typed accessors for each style property. This results in a large API surface area. Fortunately, this is automated, by generating the API – and the regular portion of the implementation – from the style specification.

The layers API makes a distinction between public API and internal implementation using [the `Impl` idiom](https://github.com/mapbox/mapbox-gl-native/issues/3254) seen elsewhere in the codebase. Here, it takes the form of two parallel class hierarchies:

* `Layer` and its subclasses form the public API.
* `Layer::Impl` and its subclasses form the internal API.

As well as forming the boundary between public and internal, these two class hierarchies form the boundary between generated code and handwritten code. Except for `CustomLayer` and `CustomLayer::Impl`:

* `Layer` subclasses are entirely generated. (`Layer` itself is handwritten.)
* `Layer::Impl` and its subclasses are entirely handwritten.

--------

There is still work to be done before this is ready to be exposed via SDK bindings:

* [x] The types relating to property values need to be cleaned up. Conceptually, a property value is a variant with three possibilities: not set (equivalent to a default value), constant value, or function value. The property API needs to reflect this. Currently every value is treated as a function value.
* [x] The namespacing needs to be rationalized. Everything style-related should live in the `mbgl::style` namespace and `mbgl/style` directory. (Beneath that directory there can be a `layer` directory for organization.)
* [x] Accessors for the layer filter and zoom range need to be added.
* [ ] There needs to be a layer getter. I'm leaning towards `map.getStyle().getLayer()` as the API. `Map::{add,remove}Layer` would also move to `Style`.


include/mbgl/layer/background_layer.hpp
```diff
@@ -0,0 +1,39 @@
...
+
+    Function<std::string> getBackgroundPattern() const;
+    void setBackgroundPattern(Function<std::string>);
+
+    Function<float> getBackgroundOpacity() const;
+    void setBackgroundOpacity(Function<float>);
+
+    // Private implementation
+
+    std::unique_ptr<Layer> clone() const final;
```
kkaefer: Any way we can move this to `Impl`?

scripts/generate-style-code.js
```diff
@@ -0,0 +1,339 @@
...
+    return property.default;
+  }
+}
+
+const layerHpp = ejs.compile(`<%
+  const type = locals.type;
+  const layoutProperties = locals.layoutProperties;
+  const paintProperties = locals.paintProperties;
+-%>
+// This file is generated. Do not edit.
```
kkaefer: Can we add hints to how to regenerate the files in there?

include/mbgl/layer/background_layer.hpp
```diff
@@ -0,0 +1,39 @@
+// This file is generated. Do not edit.
```
kkaefer: Can we generate these files during the build, like we we with shader files?

kkaefer:
Looks good! As for the generated files, it's probably useful to have them pre-generated to be able to look at them. We could still add the generation script to the build system, and check whether files were modified and fail if that is the case.

1ec5:
> We could still add the generation script to the build system, and check whether files were modified and fail if that is the case.

How about running the generation script as part of the build, but having the script output to the checkout, as it does now?

jfirebaugh:
For now I'm going to go with running the generator manually when you make a change. I looked at including it as part of the build step but outputting to the src directory (which I want to do) didn't fit in cleanly with our use of gyp.

jfirebaugh:
The last commit here involves a lot of renaming and re-namespacing. Would be great to get a second opinion on my approach:

* Everything "style-related" (roughly: implementing a concept from the style specification) lives in the `style` subdirectory and `mbgl::style` namespace.
* Inside the `mbgl::style` namespace, don't prefix class names with `Style` (except for the `Style` itself).
* Outside the `mbgl::style`, use forward declarations in headers where available.
* Use `using namespace style` in cpp files (not in headers).

PR #4842, opened 2016-04-26
Hide compass from VoiceOver when pointing north

### LINKED ISSUE #4820: Compass should be inaccessible when hidden
The compass is invisible when it points north, so it should be inaccessible to VoiceOver in that state. Alternatively, it should remain visible but allow upward and downward swipes to rotate the map.

/ref #1496
/cc @friedbunny

1ec5:
Fixes #4820.


platform/ios/src/MGLMapView.mm
```diff
@@ -4278,12 +4277,14 @@ - (CGPoint)userLocationAnnotationViewCenter
 
 - (void)updateCompass
 {
-    CLLocationDirection degrees = mbgl::util::wrap(-self.direction, 0., 360.);
-    self.compassView.transform = CGAffineTransformMakeRotation(MGLRadiansFromDegrees(degrees));
+    CLLocationDirection direction = self.direction;
+    CLLocationDirection plateDirection = mbgl::util::wrap(-self.direction, 0., 360.);
```
friedbunny: You _could_ also update this to line to use the de-`self`’d `direction`. 🤓

PR #4826, opened 2016-04-26
[Qt] libqmapboxgl.so is linking with QtGui

tmpsantos:
Only the Qt app should link with `QtGui`. We need to dismember the `qt_ldflags` variable.

Same goes to the Qt test app that is linking with a lot of Qt libraries that are not in use directly.

/cc @brunoabinader


tmpsantos:
Merged.

PR #4856, opened 2016-04-26
Fix ℹ️ accessibility and make the user dot less noisy

1ec5:
The user location annotation view is created lazily, so if the user dot isn’t within view at launch, it doesn’t exist by the time UIKit first queries MGLMapView for child accessibility elements. This caused console spew about missing accessibility elements and prevented the ℹ️ from being visible to XCUITest.

Constantly posting accessibility layout change notifications on every location change made it very difficult to stay focused on an accessibility element, be it the user dot or some other element. Now we only post the notification when the user dot is focused, user tracking mode is off (so the user dot actually has a chance to move around), and the application is active (not in the middle of application switching). We also keep the focus on the user dot in that case, so you don’t have to find it again.

/ref #1496
/cc @friedbunny

### DIFF
--- a/platform/ios/resources/Base.lproj/Localizable.strings
+++ b/platform/ios/resources/Base.lproj/Localizable.strings
@@ -38,7 +38,7 @@
 "INFO_A11Y_LABEL" = "About this map";
 
 /* Accessibility label */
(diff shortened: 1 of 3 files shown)


friedbunny:
👍

PR #4858, opened 2016-04-26
User Location Toggle In TestApp

### LINKED ISSUE #4835: User Location On / Off Test Example
The UserLocationView was refactored last week (yay!) in #4396.  I've noticed some difficulties / sluggishness getting it to work in an internal app but it's unclear if the issue is with the implementing app or with the SDK.  There are currently 2 TestApp examples that deal with User Location (`MyLocationCustomisationActivity` and `MyLocationTrackingModeActivity`) however neither of them allow for a simple toggle on / toggle off.  Instead of overloading these two activities, the goal will be to create a simple FAB based toggle so that the issue can be sandboxed.

bleege:
Closes #4835


bleege:
Rebased and Merged!

Issue #4821, opened 2016-04-25
Announce zoom level and visible features after zooming via VoiceOver


1ec5:
After zooming in or out using VoiceOver’s upward or downward swipe gestures, MGLMapView should announce the new zoom level and summarize the visible features. For comparison, MapKit provides a readout similar to the following:

> Zoom 15x. Places visible: AutoZone, McDonald’s. 16 roads visible.

Announcing the zoom level would be easy, but I don’t know how to manually trigger a readout that doesn’t correspond to an element on-screen. Summarizing the visible roads or points of interest would require #352. In the meantime, we’d want to summarize the visible annotations, as the fact that they’re visible is perhaps more intentional than the fact that some road’s label happens to be visible.

/ref #1496
/cc @friedbunny

1ec5:
An implementation is ready in #9950.

PR #4859, opened 2016-04-26
Announce zoom level and visible annotations after zooming via VoiceOver

1ec5:
Announce the zoom level and number of visible annotations after zooming using VoiceOver’s swipe up or swipe down gesture.

/ref #4821
/cc @friedbunny


1ec5:
1dd7492fee228055d1e1dac2ccaffec5871b4fd1 moves the .stringsdict file to the correct location in en.lproj rather than Base.lproj.

PR #4864, opened 2016-04-27
[android] #4850 - user location view is show instantly when toggled on

### LINKED ISSUE #4850: UserLocationView only shown after second location update
With #4396 we landed integration of entity interpolation between location updates to have smooth animation when we are changing the location of the map/userdot. As a side effect this resulted in showing the UserLocationView only after the second location update. The view should be shown instantly with the first location update.

tobrun:
Closes #4850


tobrun:
Going to rebase and merge

PR #4870, opened 2016-04-27
[ios] Add a property for bytes used by offline tiles

bwhtmn:
Having a separate property for the bytes used by offline tiles allows their data usage to be distinguished from the style data (which is often shared between multiple offline packs). I was finding it to be confusing/misleading when even a region containing a handful of times had a large size, due to the shared resources.


jfirebaugh:
Thanks for the contribution! I've done some followup work and created a branch in this repo so that we can run CI against the changes: #4874.

Issue #4846, opened 2016-04-26
[Qt] Implement QQuickMapboxGL 


tmpsantos:
A QML Item exposing an interface compatible with QtLocation [Map](http://doc.qt.io/qt-5/qml-qtlocation-map.html).

At the moment the plan is to use [QQuickFramebufferObject](http://doc.qt.io/qt-5/qquickframebufferobject.html) which will set our minimum Qt version supported at 5.2.

The Qt5 build bot was already upgraded https://github.com/mapbox/mapbox-gl-native/pull/4827.

/cc @brunoabinader

brunoabinader:
We've progressed substantially the past week, and now `QQuickMapboxGL` already has a working example:
![ca5b077c-113f-11e6-99b9-9eca75cc7d41](https://cloud.githubusercontent.com/assets/76133/14986318/01fd108a-1154-11e6-9052-efe0a9f747d1.png)
We are keeping up API parity with Qt's [Map QML Type](http://doc.qt.io/qt-5/qml-qtlocation-map.html) in our `QQuickMapboxGL` QML Type, based on `QQuickFramebufferObject` + `QMapboxGL`. A QML example application is provided and can be built/executed via `$ make run-qt-qml-app`

`QQuickMapboxGL` implements the following:
- [Qt Map Type](http://doc.qt.io/qt-5/qml-qtlocation-map.html)-like properties
  - Fully supported: `minimumZoomLevel`, `maximumZoomLevel`, `zoomLevel`, `center`, `error`, `errorString`, `visibleRegion`
  - Replaced with `style`: `activeMapType`, `supportedMapTypes`
  - Not needed/missing: `gesture`, `plugin`, `mapItems`, `copyrightsVisible`, `color` (depends on https://github.com/mapbox/mapbox-gl-native/issues/837)
- `QQuickMapboxGL`-specific properties: `style`, `bearing`, `pitch`

PR #4872, opened 2016-04-27
[Qt] Expose center and zoom properties

tmpsantos:
Initial implementation of the Qt Location Map API.

Part of #4846


brunoabinader:
👍

PR #4874, opened 2016-04-27
Add tile-only count and size properties to offline status

jfirebaugh:
Followup to #4870.

### DIFF
--- a/include/mbgl/storage/offline.hpp
+++ b/include/mbgl/storage/offline.hpp
@@ -93,12 +93,20 @@ public:
     uint64_t completedResourceCount = 0;
 
     /**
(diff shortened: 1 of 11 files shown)


1ec5:
Awesome, this is worth mentioning in the [iOS SDK changelog](https://github.com/mapbox/mapbox-gl-native/blob/bwhtmn-tile-size/platform/ios/CHANGELOG.md#master). Also, [MGLOfflinePackTests](https://github.com/mapbox/mapbox-gl-native/blob/bwhtmn-tile-size/platform/darwin/test/MGLOfflinePackTests.m#L21) should be updated too.

jfirebaugh:
@1ec5 Can I hand off to you to hit those things?

PR #4879, opened 2016-04-27
Fix 'make linux' build on OS X

springmeyer:
The "make linux" target is really just "build an app using GLFW that runs on unix". As such this works on OS X and well as Linux, however the OS X build appears to have regressed. This pull fixes this error:

```
$ make linux
/Applications/Xcode.app/Contents/Developer/usr/bin/make -f scripts/main.mk Makefile/glfw-app
* Recreating project...
deps/run_gyp platform/osx/platform.gyp -I./build/osx-x86_64/config.gypi -Dcoverage=0 -Dcxx_host="clang" -Goutput_dir=. --depth=. --generator-output=./build/osx-x86_64 -f make
gyp: build/osx/config.gypi not found (cwd: /Users/dane/projects/mbgl) while reading includes of platform/osx/platform.gyp
make[1]: *** [Makefile/__project__] Error 1
make: *** [glfw-app] Error 2
```

Question: can we run `make linux` on OS X CI to ensure this does not break again? /cc @1ec5 @jfirebaugh

### DIFF
--- a/platform/osx/bitrise.yml
+++ b/platform/osx/bitrise.yml
@@ -33,6 +33,8 @@ workflows:
             export BUILDTYPE=Debug
             make osx
             make test-osx
+            make linux
+            make test-linux
         - is_debug: 'yes'
     - slack:
         title: Post to Slack
--- a/Makefile
+++ b/Makefile
@@ -45,7 +45,7 @@ $(OSX_OUTPUT_PATH)/mbgl.xcconfig: $(OSX_OUTPUT_PATH)/config.gypi
 	./scripts/export-xcconfig.py $< $@
 
 $(OSX_PROJ_PATH): platform/osx/platform.gyp $(OSX_OUTPUT_PATH)/config.gypi $(OSX_OUTPUT_PATH)/mbgl.xcconfig $(GYP_DEPENDENCIES)
-	./deps/run_gyp -f xcode --depth=. --generator-output=$(OSX_OUTPUT_PATH) $<
+	./deps/run_gyp -f xcode --depth=. -I$(OSX_OUTPUT_PATH)/config.gypi --generator-output=$(OSX_OUTPUT_PATH) $<
 
 osx: $(OSX_PROJ_PATH)
 	set -o pipefail && xcodebuild \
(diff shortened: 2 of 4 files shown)


jfirebaugh:
`make linux` is no longer intended to work on OS X. (As you can see, this change breaks the OS X build.) To run the GLFW app on OS X, use `make xproj` and build the "glfw-app" target.

1ec5:
We removed the architecture from the build output path so that builds within Xcode and on the command line would go to the same place.

springmeyer:
@1ec5 @jfirebaugh - latest work in this branch keeps both builds passing. Let me know if this looks okay to and then I'll rebase and merge.

platform/osx/bitrise.yml
```diff
@@ -33,6 +33,8 @@ workflows:
             export BUILDTYPE=Debug
             make osx
             make test-osx
+            make linux
+            make test-linux
```
1ec5: How much time does this add to the build? (We’re trying to keep build times down.) With these invocations, do we end up rebuilding everything from scratch?

1ec5: From https://www.bitrise.io/build/80b3fbddc8ce5bda, it looks like core code would indeed get rebuilt from scratch. Generally, builds of master have been taking 5–10 minutes, but we expect the build time to go up if we decide to run iOS SDK UI tests on Bitrise: #4837.

jfirebaugh:
Hmm, I think I am opposed to this approach as well. I think `make linux`, if supported, should be a cross-compile operation, like `make android`, rather than a way to do a Makefile-based build of OS X platform binaries.

springmeyer:
Will close. My next actions are:

  - Using `test-node` currently to have an easy way to run a build quickly that respects CXXFLAGS, LDFLAGS
  - When I have more time to debug I'll try to figure out how to pass LD_OTHER_LDFLAGS down to xcode for the `make osx` build. Currently it looks like `CXXFLAGS` are magically respected (don't know how) but not `LDFLAGS` (as expected).

1ec5:
Typically you'd pass `OTHER_CPPFLAGS` and `OTHER_LDFLAGS` into each `xcodebuild` invocation. See platform/ios/package.sh for examples of passing build options into `xcodebuild`.

PR #4881, opened 2016-04-28
Respect V arg to make to disable xcpretty

springmeyer:
Doing `make V=1` is a common way of asking make file based builds to show all compiler arguments. 

This pull passes that option down to disable the use of `xcpretty` for osx/ios command line builds.

The goal here is to quickly be able to see all the compiler arguments to ensure any custom flags are being respected (without a developer needing to find and hack the `Makefile` to do this).

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -9,6 +9,12 @@ else
   $(error Cannot determine host platform)
 endif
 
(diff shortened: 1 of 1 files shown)


jfirebaugh:
👍

PR #4882, opened 2016-04-28
[core] Eliminate interpolate special cases

jfirebaugh:
This changes some of the overloads to return the second value rather than the first. Returning the second value is consistent with -js. It also probably doesn't really matter in practice.

cc @ansis


ansis:
Where does interpolate with these non-interpolatable values? I know I added this but I can't seem to find it.

jfirebaugh:
@ansis: [Here](https://github.com/mapbox/mapbox-gl-native/blob/8507b8663de780e0ad1b6631550a77965e4c3358/src/mbgl/style/paint_property.hpp#L131)

jfirebaugh:
I reworked this to use the interpolating version by default. Non-interpolating types now need to explicitly opt out via a partial specialization.

PR #4887, opened 2016-04-28
[android] WIP - running on Kazam trooper, fixes black screen issue

tobrun:


platform/android/src/native_map_view.cpp
```diff
@@ -391,6 +391,11 @@ void NativeMapView::createSurface(ANativeWindow *window_) {
     if (!firstTime) {
         firstTime = true;
 
+        oldDisplay = eglGetCurrentDisplay();
```
tmpsantos: Probably you no longer need to get the oldDisplay, surface, etc again on `::activate()`

tobrun: Will look into this in #4892

tobrun:
Closing this in favour of #4892

PR #4892, opened 2016-04-28
[android] - fix black screen

tobrun:
This fixes the black screen issue we have seen on multiple devices. It reimplements a fix we used to have for these devices but was removed during the lift for rendering on main thread. 

👀  @tmpsantos 

cc @bleege @zugaldia @cammace @danswick

### DIFF
--- a/platform/android/src/native_map_view.cpp
+++ b/platform/android/src/native_map_view.cpp
@@ -391,6 +391,8 @@ void NativeMapView::createSurface(ANativeWindow *window_) {
     if (!firstTime) {
         firstTime = true;
 
+        activate();
+
         if (!eglMakeCurrent(display, surface, surface, context)) {
             mbgl::Log::Error(mbgl::Event::OpenGL, "eglMakeCurrent() returned error %d",
                              eglGetError());
@@ -410,6 +412,8 @@ void NativeMapView::createSurface(ANativeWindow *window_) {
         mbgl::gl::InitializeExtensions([] (const char * name) {
              return reinterpret_cast<mbgl::gl::glProc>(eglGetProcAddress(name));
         });
+
+        deactivate();
     }
 }
 


tobrun:
@tmpsantos  indicated we could remove this line:

```
        oldDisplay = eglGetCurrentDisplay();
```

Since it will be filled up in activate();

tmpsantos:
> @tmpsantos indicated we could remove the following lines:
> 
>         oldDisplay = eglGetCurrentDisplay();
>         oldReadSurface = eglGetCurrentSurface(EGL_READ);
>         oldDrawSurface = eglGetCurrentSurface(EGL_DRAW);
>         oldContext = eglGetCurrentContext();
> Since it will be filled up in activate();
> 
> update: I tried removing those but this results in a black screen again

It is the other way around. You keep the onde on the `createSurface()` and remove on `activate()`. :smile:

tobrun:
> It is the other way around. You keep the onde on the createSurface() and remove on activate(). :smile:

:face_palm:

tobrun:
@tmpsantos 
Tested it out and it works 🙇

platform/android/src/native_map_view.cpp
```diff
@@ -127,11 +127,6 @@ std::array<uint16_t, 2> NativeMapView::getFramebufferSize() const {
 void NativeMapView::activate() {
     mbgl::Log::Debug(mbgl::Event::Android, "NativeMapView::activate");
 
-    oldDisplay = eglGetCurrentDisplay();
```
jfirebaugh: This doesn't look correct to me -- we want `deactivate` to restore the display and surfaces that were active when `activate` was called. We shouldn't assume that's the same display and surfaces that were active when `createSurface` was called.

I think rather what we should do is call `activate` / `deactivate` _inside_ `createSurface`, around the portion of the code where we want our context active.

tobrun:
@jfirebaugh I implemented the fix from https://github.com/mapbox/mapbox-gl-native/pull/4892#discussion_r61494850 in 13d1149:

This results in:
  - resolving black screen on Kazam Trooper
  - not introducing regressions on Nexus 6P or Nexus 5X (x86)

Could you validate if this was the solution you had in mind? 

cc @tmpsantos

jfirebaugh:
Perfect. 🚢

tmpsantos:
:shipit:

tobrun:
Co-workers approved, CI approved, going to rebase and merge 🚀

bleege:
\o/  Great job @tobrun!  🚀

PR #4897, opened 2016-04-28
[android] added user dot customization example

### LINKED ISSUE #4896: Add MyLocation Customization Example To TestApp
The recent rebuild of the MyLocation API (aka User Dot) has create the ability to more fully customize it's appearance (see #4878 ).  To show case how this works as well as help the project visually verify potential regressions going forward, there should be an a standalone example in the TestApp.  The example should allow the user to toggle / switch between the 3 primary use cases.

1. Regular (Basic MapView setup)
  * Blue User Dot and Blue Accuracy Ring
1. Changed Dot Tint (that also changes Accuracy Ring by default too)
  * Green User Dot and Green Accuracy Ring
1. Changed Accuracy Ring Tint (that leaves the User Dot it's current color)
  * Blue User Dot and Red Accuracy Ring

/cc @cammace @danswick @tobrun @zugaldia

cammace:
Fixes #4896 example for the user dot. It changes the color depending on which button you push at the bottom. 

cc: @zugaldia @tobrun @bleege

### DIFF
--- a/platform/android/MapboxGLAndroidSDKTestApp/src/main/AndroidManifest.xml
+++ b/platform/android/MapboxGLAndroidSDKTestApp/src/main/AndroidManifest.xml
@@ -148,6 +148,14 @@
                 android:value="@string/category_userlocation" />
         </activity>
         <activity
+            android:name=".activity.userlocation.MyLocationDotColor"
+            android:description="@string/description_user_location_dot_color"
+            android:label="@string/activity_user_dot_color">
+            <meta-data
+                android:name="@string/category"
+                android:value="@string/category_userlocation" />
+        </activity>
+        <activity
             android:name=".activity.userlocation.MyLocationToggleActivity"
             android:description="@string/description_user_location_toggle"
             android:label="@string/activity_user_location_toggle">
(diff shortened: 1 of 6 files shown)


tobrun:
[Suggestion](http://jeroenmols.com/img/blog/resourcenaming/resourcenaming_cheatsheet.pdf) for convention on naming resources.

cammace:
Fixed id naming to use all lowercase and underscores (i.e. mapboxGreen became mapbox_green). User permissions check is added within the activity like the example @bleege showed. I'm not sure if this was the best way to check as the entire activity revolves around the user location so maybe checking in `FeatureOverviewActivity` is better? 

Two possible bugs I noticed:
- When changing between user dot and accuracy colors, it sometimes wont immediately show the change. Maybe this has to do with location becoming stale or invalidate needing to be called?

- Tilting the map causes the accuracy ring becomes uncentered around the user dot.
screenshot_20160502-124843

tobrun:
Great find @cammace! this is definitely a serious bug. Don't hesitate to ticket these directly else there is a chance of them getting lost in the firehose.

bleege:
@cammace Good find of the bug.  👍  That said, it's an issue with the User Dot and not your example code for the TestApp.  If everything else is ready to go I'd recommend merging this into `master` so that we can use it to help fix the separate bug that you found.

tobrun:
I'm going to restart the failing test and merge this so I can start taking a look at resolving #4930 and #4929. For future occasions @cammace it's handy if you link the issues/PR this makes it easier to navigate through Github.

tobrun:
CI approved, going to rebase/squash and merge this

PR #4901, opened 2016-04-28
[linux] Build with GCC 5

### LINKED ISSUE #2553: gcc-5
Any objections to moving from `gcc-4.9` to `gcc-5` for the gl-native Travis matrix? i haven’t been able to track down why `gcc-4.9/stdlibc++` [isn’t working with `clang/stdlibc++` binaries](https://github.com/mapbox/mapbox-gl-native/pull/2489), but `gcc-5` is compiling with them just fine.

/cc @mapbox/gl

jfirebaugh:
Fixes #2553

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,19 +8,19 @@ addons_shortcuts:
       sources: [ 'ubuntu-toolchain-r-test', 'llvm-toolchain-precise-3.5' ]
       packages: [ 'gdb', 'clang-3.5', 'libstdc++-4.9-dev', 'libstdc++6', 'libllvm3.4', 'xutils-dev',
                   'libxxf86vm-dev', 'x11proto-xf86vidmode-dev', 'mesa-utils' ]
-  addons_gcc49: &gcc49
+  addons_gcc5: &gcc5
     apt:
       sources: [ 'ubuntu-toolchain-r-test' ]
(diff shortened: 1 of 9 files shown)


jfirebaugh:
I expected to get linker errors in gtest and geojson-vt-cpp due to the ABI change, but didn't. 😕 

cc @tmpsantos

tmpsantos:
> I expected to get linker errors in gtest and geojson-vt-cpp due to the ABI change, but didn't.

That is because we are probably installing the compiler but linking with an older `libstdc++` (not the new one that supposedly ships with GCC 5 and caused the ABI breakage).

I still get the same linking errors on my Ubuntu 16.04 with this branch. Let's go ahead and merge this compiler upgrade and proceed with https://github.com/mapbox/mason/issues/157.

jfirebaugh:
Yeah, I think it's either that or that the `ubuntu-toolchain-r` GCC 5 packages for precise are configured with to use `_GLIBCXX_USE_CXX11_ABI=0` by default. Can't find any documentation of that but the [GCC dual ABI docs](https://gcc.gnu.org/onlinedocs/libstdc++/manual/using_dual_abi.html) do say "some GNU/Linux distributions configure GCC 5 differently so that the default value of the macro is 0".

PR #2696, opened 2016-04-29
Use Travis Ubuntu Trusty 14.04 build env

jfirebaugh:
If you just want to get started using our Trusty beta, then add the following to your `.travis.yml`.

```
sudo: required
dist: trusty
```

http://blog.travis-ci.com/2015-10-14-opening-up-ubuntu-trusty-beta/

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,3 +1,7 @@
+os: linux
+sudo: required
+dist: trusty
+
 git:
   submodules: false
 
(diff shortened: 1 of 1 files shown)


jfirebaugh:
Close. Need to figure out why the clang debug build hangs when launching the tests.

tmpsantos:
> Close. Need to figure out why the clang debug build hangs when launching the tests.

The issue was with running the tests inside `gdb`. Not installing `gdb` for LLVM bots fixes.

I migrate the LLVM bots to Trusty, but I recommend we wait until Trusty gets "containerized" like Precise before migrating the rest. It feels to me the bots have a significantly slower startup time.

The LLVM bots needed migration because llvm.org is no longer serving clang ubuntu packages.

PR #4905, opened 2016-04-29
[ios] Fix analyzer's nullability complaints

friedbunny:
Thanks for walking me through this, @1ec5.

- Redefine as local variables
- Check/assert that these local variables are not null

Analyzer warnings:

screen shot 2016-04-29 at 3 57 50 pm
screen shot 2016-04-29 at 3 54 51 pm

### DIFF
--- a/platform/ios/src/MGLMapView.mm
+++ b/platform/ios/src/MGLMapView.mm
@@ -594,7 +594,11 @@ mbgl::Duration MGLDurationInSeconds(NSTimeInterval duration)
     [_attributionButton removeObserver:self forKeyPath:@"hidden"];
     
     // Removing the annotations unregisters any outstanding KVO observers.
-    [self removeAnnotations:self.annotations];
+    NSArray *annotations = self.annotations;
+    if (annotations)
+    {
+        [self removeAnnotations:annotations];
+    }
     
     [self validateDisplayLink];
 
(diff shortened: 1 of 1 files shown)


platform/ios/src/MGLMapView.mm
```diff
@@ -1567,7 +1571,9 @@ - (void)handleCalloutAccessoryTapGesture:(UITapGestureRecognizer *)tap
     if ([self.delegate respondsToSelector:@selector(mapView:annotation:calloutAccessoryControlTapped:)])
     {
         NSAssert([tap.view isKindOfClass:[UIControl class]], @"Tapped view %@ is not a UIControl", tap.view);
-        [self.delegate mapView:self annotation:self.selectedAnnotation
+        id <MGLAnnotation> selectedAnnotation = self.selectedAnnotation;
+        NSAssert(selectedAnnotation != NULL, @"Selected annotation should not be NULL.");
```
1ec5: Use `nil` instead of `NULL`. `id <MGLAnnotation>` is an Objective-C type, not a C type.

friedbunny: What would you say to omitting either?

1ec5: Oh yeah, that’s fine too.

friedbunny: 👍 → e8689d691c0bff0b22ff0cf421e94e55f33bb732

PR #4916, opened 2016-05-02
[core] Replace boost::lexical_cast with std::to_string

jfirebaugh:
This removes the only boost dependency from public headers.


springmeyer:
💯

PR #4917, opened 2016-05-02
[core] Use gyp's direct_dependent_settings to propagate variant cflags

jfirebaugh:
variant.hpp is used for types that are included in the public API, so all dependents must be able to include the header.

cc @springmeyer


springmeyer:
Nice 💯

PR #4919, opened 2016-05-02
Boost headers are no longer needed in project files after #4916

springmeyer:


1ec5:
Not sure what the Node failure is about, but the project changes look good to me.

PR #4925, opened 2016-05-03
[Qt] Improve Qt QML app example

tmpsantos:
- Add slider for controlling pitch.
- Rounded corners on the map.
- Add slider to flip the map to a satellite (in fact we have two maps in sync).

### DIFF
--- a/platform/qt/include/qquickmapboxgl.hpp
+++ b/platform/qt/include/qquickmapboxgl.hpp
@@ -51,7 +51,6 @@ public:
     void setZoomLevel(qreal zoomLevel);
     qreal zoomLevel() const;
 
-    void setCenter(const QGeoCoordinate &center);
     QGeoCoordinate center() const;
 
     QGeoServiceProvider::Error error() const;
@@ -103,6 +102,9 @@ signals:
 
     void styleChanged();
     void bearingChanged(qreal angle);
+
+public slots:
+    void setCenter(const QGeoCoordinate &center);
 
 private:
     qreal m_minimumZoomLevel = 0;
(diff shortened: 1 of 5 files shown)


tmpsantos:
:eyes: @brunoabinader

platform/qt/include/qquickmapboxgl.hpp
```diff
@@ -83,6 +87,7 @@ class Q_DECL_EXPORT QQuickMapboxGL : public QQuickFramebufferObject
         StyleNeedsSync   = 0x04,
         PanNeedsSync     = 0x08,
         BearingNeedsSync = 0x16,
+        PitchNeedsSync   = 0x32,
```
brunoabinader: Can we use `1 << {2,3,4...}` notation instead? I think it is easier to maintain than using raw values.

platform/qt/qmlapp/main.qml
```diff
@@ -1,35 +1,220 @@
...
+import QtQuick 2.0
+import QtQuick.Controls 1.0
+import QtQuick.Layouts 1.0
 
 import QQuickMapboxGL 1.0
 
 ApplicationWindow {
-    id: window
-    title: qsTr("Mapbox GL QML example")
+    title: "Mapbox GL QML example"
```
brunoabinader: Can we keep the `qsTr`? Who knows we'll need this translated some day :-)

brunoabinader:
Minor comments, but overall 👍 this is looking awesome!

platform/qt/qmlapp/main.qml
tmpsantos: Does it make sense to have it without a translation file?

PR #4928, opened 2016-05-03
[Qt] Removed unused resource

brunoabinader:
Removes a prior resource image that is no longer in use.

/cc @tmpsantos


tmpsantos:
:+1:

Issue #4847, opened 2016-04-26
ScreenshotUtil for tests is broken


tobrun:
When trying to take a screenshot using espresso (using ScreenshotUtil.java). The utility crashes with:

```java
04-26 11:33:23.300 15948-15978/com.mapbox.mapboxsdk.testapp A/art: art/runtime/java_vm_ext.cc:466] JNI DETECTED ERROR IN APPLICATION: thread Thread[12,tid=15978,Native,Thread*=0x7c65215800,peer=0x12cdc820,"Instr: android.support.test.runner.AndroidJUnitRunner"] using JNIEnv* from thread Thread[12,tid=15978,Native,Thread*=0x7c65215800,peer=0x12cdc820,"Instr: android.support.test.runner.AndroidJUnitRunner"]
04-26 11:33:23.300 15948-15978/com.mapbox.mapboxsdk.testapp A/art: art/runtime/java_vm_ext.cc:466]     in call to CallVoidMethodV
04-26 11:33:23.300 15948-15978/com.mapbox.mapboxsdk.testapp A/art: art/runtime/java_vm_ext.cc:466]     from void com.mapbox.mapboxsdk.maps.NativeMapView.nativeRender(long)
04-26 11:33:23.300 15948-15978/com.mapbox.mapboxsdk.testapp A/art: art/runtime/java_vm_ext.cc:466] "Instr: android.support.test.runner.AndroidJUnitRunner" prio=5 tid=12 Runnable
04-26 11:33:23.300 15948-15978/com.mapbox.mapboxsdk.testapp A/art: art/runtime/java_vm_ext.cc:466]   | group="main" sCount=0 dsCount=0 obj=0x12cdc820 self=0x7c65215800
04-26 11:33:23.300 15948-15978/com.mapbox.mapboxsdk.testapp A/art: art/runtime/java_vm_ext.cc:466]   | sysTid=15978 nice=-8 cgrp=default sched=0/0 handle=0x7c703b7450
04-26 11:33:23.300 15948-15978/com.mapbox.mapboxsdk.testapp A/art: art/runtime/java_vm_ext.cc:466]   | state=R schedstat=( 159305621 2259425 55 ) utm=4 stm=11 core=2 HZ=100
04-26 11:33:23.300 15948-15978/com.mapbox.mapboxsdk.testapp A/art: art/runtime/java_vm_ext.cc:466]   | stack=0x7c702b5000-0x7c702b7000 stackSize=1037KB
[...]

PR #4931, opened 2016-05-03
Integrating ScreenshotActivityTest

bleege:
Prep work for #4847 to be resolved.  To be able to diagnose and resolve the breaking code the breaking test needs to be in the code base so that everyone can work from the same starting point.


bleege:
Rebased and Merged!

Issue #4909, opened 2016-05-02
[core] Crash when launching glfw app on Linux


tmpsantos:
Release mode, crashes right after the window pops up.

Reverting #4900 fixes.

```
[Thread debugging using libthread_db enabled]
Using host libthread_db library "/lib/x86_64-linux-gnu/libthread_db.so.1".
Core was generated by `./mapbox-glfw'.
Program terminated with signal SIGSEGV, Segmentation fault.
bt#0  0x000000000051d37e in mbgl::GeometryTileFeatureExtractor::getValue (this=0x7f10e92aaf70, key=...)
    at ../../src/mbgl/tile/geometry_tile.cpp:10
10	    return feature.getValue(key);
[Current thread is 1 (Thread 0x7f10e92af700 (LWP 27117))]
(gdb) bt
#0  0x000000000051d37e in mbgl::GeometryTileFeatureExtractor::getValue (this=0x7f10e92aaf70, key="scalerank")
    at ../../src/mbgl/tile/geometry_tile.cpp:10
#1  0x00000000004cbfc5 in mbgl::FilterEvaluator<mbgl::GeometryTileFeatureExtractor>::operator() (this=this@entry=0x7f10e92aaf60, 
    filter=...) at ../../src/mbgl/style/filter_evaluator.hpp:51
#2  0x00000000004c6e28 in mapbox::util::detail::dispatcher<mbgl::FilterEvaluator<mbgl::GeometryTileFeatureExtractor>&, mapbox::util::variant<mbgl::NullFilter, mbgl::EqualsFilter, mbgl::NotEqualsFilter, mbgl::LessThanFilter, mbgl::LessThanEqualsFilter, mbgl::GreaterThanFilter, mbgl::GreaterThanEqualsFilter, mbgl::InFilter, mbgl::NotInFilter, mbgl::AnyFilter, mbgl::AllFilter, mbgl::NoneFilter>, bool, mbgl::InFilter, mbgl::NotInFilter, mbgl::AnyFilter, mbgl::AllFilter, mbgl::NoneFilter>::apply_const (f=..., v=...)
    at /opt/tmpsantos/Projects/mapbox-gl-native/mason_packages/headers/variant/1.1.0/include/mapbox/variant.hpp:302
[...]

PR #4935, opened 2016-05-04
[core] Simplify FilterEvaluator and fix crash

jfirebaugh:
Speculative fix for #4909. Previous implementation was assigning a temporary to `FilterEvaluator<T>::extractor`.

cc @tmpsantos


brunoabinader:
Thank you @jfirebaugh - that fixed the crash we were experiencing on Ubuntu Xenial 16.04 + GCC/libstdc++ 5 :+1:

tmpsantos:
Worked for me to. :+1:

PR #4914, opened 2016-05-02
Fix a bug in applying padding to the camera.

bwhtmn:
NE padding was applied as if it was SW, thus causing the view's center to be offset.

**Example**

Originally detected the issue when using the iOS method
`- (void)setVisibleCoordinateBounds:(MGLCoordinateBounds)bounds edgePadding:(UIEdgeInsets)insets animated:(BOOL)animated;` to show the black polyline annotation with insets of top = 20, left = 20, bottom = 80, right = 20.

Before this fix:
screen shot 2016-05-02 at 9 44 05 am

After this fix:
screen shot 2016-05-02 at 9 42 15 am

PR #4937, opened 2016-05-04
Fix incorrect offset when fitting to bounds with padding

### LINKED ISSUE #4862: Fitting viewport to annotations bounds does not center properly

1ec5:
This is a rebase of #4914 so that we can fast-forward merge the change and run tests against it on Bitrise. Thanks to @bwhtmn for the fix.

Fixes #4862.

/cc @friedbunny @tobrun

### DIFF
--- a/src/mbgl/map/map.cpp
+++ b/src/mbgl/map/map.cpp
@@ -523,7 +523,7 @@ CameraOptions Map::cameraForLatLngs(const std::vector<LatLng>& latLngs, optional
             padding->left / minScale,
             padding->bottom / minScale,
         };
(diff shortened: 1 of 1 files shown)


1ec5:
Actually, since the tests here pass, I’m going to fast-forward merge @bwhtmn’s original PR, #4914, using GitHub’s spiffy new squash-merge feature.

PR #4947, opened 2016-05-04
Upgrade to geojsonvt@5.0.0

springmeyer:
This brings in an optimized geojsonvt release: https://github.com/mapbox/geojson-vt-cpp/releases/tag/v5.0.0.

This is the first small step to using geojsonvt that is based on [geometry.hpp](https://github.com/mapbox/geometry.hpp). I think this is worth merging now to get the new 5.x release more testing.

More work remains and will trigger further upgrades: https://github.com/mapbox/geojson-vt-cpp/issues/40


src/mbgl/annotation/shape_annotation_impl.cpp
```diff
@@ -88,16 +88,15 @@ void ShapeAnnotationImpl::updateTile(const TileID& tileID, AnnotationTile& tile)
         const double tolerance = baseTolerance / (maxAmountOfTiles * GeometryTileFeature::defaultExtent);
 
         geojsonvt::ProjectedRings rings;
-        std::vector<geojsonvt::LonLat> points;
+        mapbox::geometry::linear_ring<double> points;
```
jfirebaugh: Can this be `std::vector<mapbox::geometry::point<double>>` instead? `linear_ring` is not technically correct because this is processing both line and polygon geometry, and `linear_ring` is an appropriate type only for the latter.

jfirebaugh:
This no longer builds due to an ABI break -- the precompiled geojson-vt-cpp binaries were built with geometry.hpp 0.1.0, while mapbox-gl-native is now using 0.5.0, and the two are incompatible.

I suggest we hold off on a geojson-vt-cpp upgrade until after https://github.com/mapbox/geojson-vt-cpp/pull/45 lands.

jfirebaugh:
We're on geojson-vt-cpp 6.1.0 now.

Issue #4392, opened 2016-03-18
Move callout customization away from MGLMapViewDelegate towards custom views


1ec5:
The current annotation callout API is implemented as a series of delegate methods in MGLMapViewDelegate that configure SMCalloutView. SMCalloutView provides a lot of customization options, but we’re only exposing a few basic ones.

SMCalloutView is great for callout views that need a certain layout, but many developers need more flexibility and don’t mind setting up the layout and drawing themselves. The current API for custom callout views can be very confusing for a developer who wants something more freeform – “Why do I have to implement accessory views?” And you’re basically on your own in terms of styling as soon as you go the custom callout route.

Let’s drop the dependency on SMCalloutView and deprecate the current callout API – both the delegate methods and the protocol – in favor of a more deliberate, coherent callout API. The proposal below is aimed at aligning the iOS SDK’s annotations API with that of the OS X SDK, which manages callouts with NSPopoverControllers and allows you to specify your own.

---

You should be able to drag a custom view into a storyboard, turn it into an MGLMapView, and design the custom callout views directly inside the storyboard. You’d drag a custom view into the scene dock above the view controller and optionally set the custom class to MGLCompactCalloutView, which provides the standard bubble look and feel (surfaced in the storyboard via a designable). You’d specify a reuse identifier for the callout view, then drag a connection from the callout view to the map view’s `prototypeCalloutViews` outlet collection. The callout view can be customized extensively without any work on our part:
- To set the callout view’s background color, you’d use the Background Color control available in the Attributes inspector for any view. If you want to get fancy, you can insert a UIVisualEffectView for a vibrant blur effect.
- You’d use the Size inspector to make the callout view taller, or even use Auto Layout constraints or a UIScrollView to ensure that the callout view fits its contents. The callout view’s size and contents could vary by size class.
- If tapping on the right side of the callout view should take the user to another view controller, you’d drag a UIButton into the right side of the callout view and drag a segue from the button to the destination view controller. Optionally, you’d connect the button to the callout view’s `rightAccessoryView` outlet so that `-[MGLMapViewDelegate mapView:annotation:calloutAccessoryControlTapped:]` would get called.

[...]

PR #4948, opened 2016-05-05
Make custom callout views easier to use [WIP]

friedbunny:
Following on from #4392, this pull request aims to simplify custom callout view creation and usage.
#### `MGLCalloutView` protocol
- Simplified.
- Most properties are optional, including accessory views and title strings.
- Adds `-detailAccessoryView`.
#### `MGLCalloutViewDelegate`
- Removed.
#### `MGLAnnotation` (Really, only `MGLPointAnnotation`)
- Adds `-canShowCallout` boolean property.
- Adds `-calloutView` property that conforms to `MGLCalloutView` protocol.
#### `MGLCompactCalloutView`
- Replaced `SMCalloutView` with custom implementation.
  - Current placeholder is based on the existing [custom callout view example](https://www.mapbox.com/ios-sdk/examples/custom-callout/):

screen shot 2016-05-04 at 9 45 43 pm
### Todo
- [ ] Interface Builder support.
- [ ] Accessibility.
- [ ] Hook-up accessory views.
- [ ] Handle gestures for callouts.
  - [ ] Re-enable `-[MGLMapView mapView:annotation:calloutAccessoryControlTapped:]`.
  - [ ] Add custom gesture handlers to custom callout example.
- [ ] Fit the default callout view to the viewport.
- [ ] Add `MGLCalloutView.annotation`?
  - Instead of manually re-setting the `title` and `subtitle` properties, maybe let the callout implementation do that directly with a passed-in `annotation`.
- [ ] Audit `MGLMapView` for dead delegate methods; remove or deprecate.
  - `-mapView:annotationCanShowCallout:`
  - `-mapView:calloutViewForAnnotation:`
  - `-mapView:tapOnCalloutForAnnotation:`
  - `-mapView:rightCalloutAccessoryViewForAnnotation:`
  - `-mapView:leftCalloutAccessoryViewForAnnotation:`
#### Stretch goals
- [ ] Make the default callout view swanky.
  - Animations and other polish.
- [ ] Eliminate need for public `MGLCalloutView` protocol.
- ~~Enable callout views for `MGLMultiPoint` (polylines, polygons)~~ Punted per https://github.com/mapbox/mapbox-gl-native/pull/4948#discussion_r62155897.
- [ ] Allow callout views to remain open while panning.

/cc @1ec5 @boundsj

### DIFF
--- a/platform/darwin/src/MGLAnnotation.h
+++ b/platform/darwin/src/MGLAnnotation.h
@@ -2,6 +2,10 @@
 #import <CoreLocation/CoreLocation.h>
 #import <TargetConditionals.h>
 
+#ifdef TARGET_OS_IPHONE
+#import <UIKit/UIKit.h>
+#endif
+
 #import "MGLTypes.h"
 
 NS_ASSUME_NONNULL_BEGIN
@@ -48,6 +52,14 @@ NS_ASSUME_NONNULL_BEGIN
  This string is displayed in the callout for the associated annotation.
  */
 @property (nonatomic, readonly, copy, nullable) NSString *subtitle;
+
+#ifdef TARGET_OS_IPHONE
+
+@property (nonatomic) BOOL canShowCallout;
+
+@property (nonatomic, nullable) UIView *callout;
+
+#endif
 
 #if !TARGET_OS_IPHONE
 
--- a/platform/darwin/src/MGLAnnotation.h
+++ b/platform/darwin/src/MGLAnnotation.h
@@ -2,7 +2,7 @@
 #import <CoreLocation/CoreLocation.h>
 #import <TargetConditionals.h>
 
-#ifdef TARGET_OS_IPHONE
+#if TARGET_OS_IPHONE
 #import <UIKit/UIKit.h>
 #import "MGLCalloutView.h"
 #endif
@@ -54,7 +54,7 @@ NS_ASSUME_NONNULL_BEGIN
  */
 @property (nonatomic, readonly, copy, nullable) NSString *subtitle;
 
-#ifdef TARGET_OS_IPHONE
+#if TARGET_OS_IPHONE
 
 @property (nonatomic) BOOL canShowCallout;
 
--- a/platform/darwin/src/MGLAnnotation.h
+++ b/platform/darwin/src/MGLAnnotation.h
@@ -4,6 +4,7 @@
 
 #ifdef TARGET_OS_IPHONE
 #import <UIKit/UIKit.h>
+#import "MGLCalloutView.h"
 #endif
 
 #import "MGLTypes.h"
@@ -57,7 +58,7 @@ NS_ASSUME_NONNULL_BEGIN
 
 @property (nonatomic) BOOL canShowCallout;
 
-@property (nonatomic, nullable) UIView *callout;
+@property (nonatomic, nullable) UIView <MGLCalloutView> *calloutView;
 
 #endif
 
--- a/platform/darwin/src/MGLShape.h
+++ b/platform/darwin/src/MGLShape.h
@@ -1,10 +1,5 @@
 #import <Foundation/Foundation.h>
 
-#if TARGET_OS_IPHONE
-#import <UIKit/UIKit.h>
-#import "MGLCalloutView.h"
-#endif
-
 #import "MGLAnnotation.h"
 
 #import "MGLTypes.h"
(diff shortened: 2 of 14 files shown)


friedbunny:
This is still a big work-in-progress; b6c0c9fe929c8db6301beb7bea7e4f0b7621d9f1 is my stopping point for tonight.

My approach so far has been to go through with a machete and wanton disregard for backwards compatibility, but clearly we’ll have to address deprecation one way or another before this is ready to go.

platform/ios/src/MGLCalloutView.h
```diff
@@ -42,35 +22,27 @@ NS_ASSUME_NONNULL_BEGIN
...
- */
-- (void)calloutViewWillAppear:(UIView<MGLCalloutView> *)calloutView;
+@property (nonatomic, strong) UIView *detailAccessoryView;
 
 /**
- Called after the callout view appears on screen, or after the appearance animation is complete.
+ An object conforming to the `MGLCalloutViewDelegate` method that receives messages related to the callout view’s interactive subviews.
  */
-- (void)calloutViewDidAppear:(UIView<MGLCalloutView> *)calloutView;
+@property (nonatomic, weak) id<MGLCalloutViewDelegate> delegate;
```
friedbunny: This is slated for removal.

platform/darwin/src/MGLAnnotation.h
```diff
@@ -36,6 +41,14 @@ NS_ASSUME_NONNULL_BEGIN
  */
 @property (nonatomic, readonly, copy, nullable) NSString *subtitle;
 
+#if TARGET_OS_IPHONE
+
+@property (nonatomic) BOOL canShowCallout;
```
1ec5: Come to think of it, these view-related properties don’t belong in a model protocol if we’re going to respect a conventional MVC pattern. MapKit places this and other callout-related properties on MKAnnotationView, which prevents the developer from customizing the callout view for custom model objects that are used as point annotations. Once #4801 lands, it may be worth moving these properties over to MGLAnnotationView.

platform/darwin/src/MGLShape.h
```diff
@@ -15,6 +20,14 @@ NS_ASSUME_NONNULL_BEGIN
 /** The subtitle of the shape annotation. The default value of this property is `nil`. */
 @property (nonatomic, copy, nullable) NSString *subtitle;
 
+#if TARGET_OS_IPHONE
+
+@property (nonatomic) BOOL canShowCallout;
+
+@property (nonatomic, nullable) UIView <MGLCalloutView> *calloutView;
```
1ec5: I’m not sure supporting callout views on shapes should be within the scope of this PR. We currently rely on `mbgl::AnnotationManager` to perform the spatial query that tells us which annotation is near a tap. That query only returns point annotations, not shapes or lines. Once we reimplement `AnnotationManager` atop a GeoJSON layer – #4839 is a start in that direction – we’ll be able to use #352 to query for shapes and lines as well. We’ll still need to figure out what callout positioning should look like for non-point features.

platform/ios/src/MGLCompactCalloutView.m
```diff
@@ -1,31 +1,127 @@
...
+        {
+            [self removeFromSuperview];
+        }
+    }
+}
+
+
+#pragma mark - Custom view styling
+
+- (UIColor *)backgroundColorForCallout
```
1ec5: Since we’re overriding `-drawRect:`, we can use the existing `backgroundColor` property to indicate the visible portion’s background color.

friedbunny: Good point — removed this method in b416ac0.

platform/darwin/src/MGLShape.h
friedbunny: Fair enough, this is already a fairly ambitiously scoped pull request. I moved these properties to `MGLPointAnnotation` in 260ed7e.

platform/darwin/src/MGLAnnotation.h
friedbunny: Where these properties live will need to be reevaluated after https://github.com/mapbox/mapbox-gl-native/pull/4801, aye. Though, if callouts only existed on `MGLAnnotationView`, wouldn’t GL (e.g., `MGLPointAnnotation`) annotations not have (custom) callouts?

1ec5: That's a good point. Maybe we should put them in a separate protocol, MGLCalloutSource, that both MGLPointAnnotation and MGLAnnotationImage could conform to.

friedbunny:
Pushing this off of v3.3.0 — haven’t made enough progress and there’s a lot of other validation to do for that release. I may keep picking at this in the meantime, though.

andrewstay:
Please clarify if that means I have to allocate all the callout views for all the annotations on the map and keep those views in the memory even though user may never lick on those?

1ec5:
@andrewstay, by the time this feature lands, it’ll work similarly to annotation views. Callout views will go into a reuse queue, so the number of allocated views will be limited.

friedbunny:
This has been rebased onto the current `master` — will be picking up work on this again this week.

friedbunny:
So stale. Hopefully this will happen eventually!

PR #4950, opened 2016-05-05
[Qt] Make it possible to disable libwebp and libjpeg-turbo at build time

tmpsantos:
Pull less dependencies. Useful when building for Yocto.

/cc @brunoabinader


brunoabinader:
👍

PR #4951, opened 2016-05-05
[core] Use geometry.hpp's point type

jfirebaugh:


jfirebaugh:
cc @kkaefer @springmeyer

platform/osx/osx.xcodeproj/project.pbxproj
```diff
@@ -1018,6 +1022,10 @@
 				HEADER_SEARCH_PATHS = ../../include;
 				INFOPLIST_FILE = test/Info.plist;
 				LD_RUNPATH_SEARCH_PATHS = "$(inherited) @executable_path/../Frameworks @loader_path/../Frameworks";
+				OTHER_CPLUSPLUSFLAGS = (
+					"$(OTHER_CFLAGS)",
+					"$(geometry_cflags)",
+				);
```
springmeyer: @jfirebaugh how come we need this here in addition to https://github.com/mapbox/mapbox-gl-native/blob/geometry.hpp/platform/osx/osx.xcodeproj/project.pbxproj#L985?

springmeyer:
overall 💯

platform/osx/osx.xcodeproj/project.pbxproj
jfirebaugh: There are two targets which wind up including mapbox/geometry.hpp (via map.hpp): this one is the osx `test` target, and the other one is the `dynamic` target that builds a framework.

PR #4952, opened 2016-05-05
Implement "has" / "!has" filter operators 

jfirebaugh:
ref https://github.com/mapbox/feature-filter/pull/15

cc @1ec5 @jfirebaugh @mourner

### DIFF
--- a/src/mbgl/style/filter.hpp
+++ b/src/mbgl/style/filter.hpp
@@ -20,7 +20,9 @@ typedef variant<
     class NotInFilter,
     class AnyFilter,
     class AllFilter,
-    class NoneFilter
+    class NoneFilter,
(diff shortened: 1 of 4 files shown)


lucaswoj:
@mb12 "has" returns true if a feature provides a value for `key`. `in` returns true of the value for `key` is in a provided set of values.


```js
['has', 'mapbox'] // returns true if the feature has a value for "mapbox"
```

```js
['in', 'mapbox', 'neat', 'swell', 'great'] // returns true if the value for "mapbox" is 'neat', 'swell', or 'great'
```

lucaswoj:
It is similar to but not quite the same as `['==', 'mapbox', undefined]`

PR #4963, opened 2016-05-06
Avoid recomputing tile cache size on every redraw

### LINKED ISSUE #4026: Avoid recomputing tile cache size on every redraw
The tile cache size is [recomputed and reset on every redraw](https://github.com/mapbox/mapbox-gl-native/blob/13e2acbc754893efb945fe02d20824698415dcdb/platform/ios/src/MGLMapView.mm#L764-L770), which seems wasteful considering that the minimum and maximum zoom levels, processor count, processor memory, and view size rarely change. (We don’t support hot-swapping RAM on your iPhone!) We should recompute the tile size only when `-setMinimumZoomLevel:`, `-setMaximumZoomLevel:`, `-setFrame:`, and `-setBounds:` are called.

/cc @ansis @tmpsantos

1ec5:
The tile cache size is only recomputed when an input changes.

This code was originally introduced in #1281 (actually 52d00686d4c160e2890a7dd9cdcaa4a1fa14d5bd for #1255). `Map::setSourceTileCacheSize()` bails if the new size is no different than the current one, but by then we’ve already gotten the zoom limits and process info. By contrast, the Android SDK sets the tile cache size just once, in the `NativeMapView` constructor. This is part of a push to minimize the work we’re doing at the SDK level in the drawing routines, currently the hottest-running SDK code on iOS and OS X.

Fixes #4026.

/cc @jfirebaugh


jfirebaugh:
👍

PR #4965, opened 2016-05-06
[core] Remove trailing zeros in float→string conversion

### LINKED ISSUE #998: Text from floating point data rounding error
Hello,

we're seeing some issues with text rendering in the iOS client. See screenshot bellow of iOS:

ios simulator screen shot 13 mar 2015 09 24 16

the same place in the js-client:

screenshot 2015-03-13 09 29 22

The depth soundings are fetched from the tile data with "text-field": "{}". The data is initially is from a single precision float column in a postgis db. The data is rounded to one decimal in the database using round(field::numeric, 1). The JS client handles this as expected, but the native iOS client does not.

Also, the text has some overflow issues, where it bleeds as illustrated on the picture. This only happens when these floating point issues are onscreen.

Let me know if you need more info.

friedbunny:
With the replacement of `boost::lexical_cast` with `std:to_string` in #4916, symbol labels that used floating point numbers could end up looking like “12345.000000” when the passed in number was actually just “12345”.

The fix here is two-fold:
- Cast sufficiently int-like floats to int.
- Otherwise, let `std::to_string` convert the float to a string, then trim trailing zeros and hanging decimal points.

If this passes muster, we should consider including it in iOS v3.2.2 and the next Android release.

Fixes #998.

/cc @jfirebaugh @1ec5 @tobrun

### DIFF
--- a/include/mbgl/util/string.hpp
+++ b/include/mbgl/util/string.hpp
@@ -4,6 +4,7 @@
 #include <string>
 #include <cassert>
 #include <exception>
+#include <cmath>
 
 namespace mbgl {
 namespace util {
@@ -14,7 +15,7 @@ inline std::string toString(T t) {
 }
 
 inline std::string toString(double num) {
-    if (num - (int)num > 0.0 || num + (int)num < 0.0) {
+    if (std::fmod(num, 1)) {
         std::string str = std::to_string(num);
         str.erase(str.find_last_not_of('0') + 1, std::string::npos);
         if (str.back() == '.') {
--- a/include/mbgl/util/string.hpp
+++ b/include/mbgl/util/string.hpp
@@ -4,7 +4,7 @@
 #include <string>
 #include <cassert>
 #include <exception>
-#include <cmath>
+#include <sstream>
 
 namespace mbgl {
 namespace util {
@@ -15,16 +15,9 @@ inline std::string toString(T t) {
 }
 
 inline std::string toString(double num) {
-    if (std::fmod(num, 1)) {
-        std::string str = std::to_string(num);
-        str.erase(str.find_last_not_of('0') + 1, std::string::npos);
-        if (str.back() == '.') {
-            str.pop_back();
-        }
-        return str;
-    } else {
-        return std::to_string(int(num));
-    }
+    std::ostringstream formatted;
+    formatted << num;
+    return formatted.str();
 }
 
 inline std::string toString(int8_t num) {
--- a/include/mbgl/util/string.hpp
+++ b/include/mbgl/util/string.hpp
@@ -10,15 +10,19 @@ namespace util {
 
 template <class T>
 inline std::string toString(T t) {
-    if (t - (int)t > 0.0 || t + (int)t < 0.0) {
-        std::string str = std::to_string(t);
+    return std::to_string(t);
+}
+
+inline std::string toString(double num) {
+    if (num - (int)num > 0.0 || num + (int)num < 0.0) {
+        std::string str = std::to_string(num);
         str.erase(str.find_last_not_of('0') + 1, std::string::npos);
         if (str.back() == '.') {
             str.pop_back();
         }
         return str;
     } else {
-        return std::to_string(int(t));
+        return std::to_string(int(num));
     }
 }
 
--- a/include/mbgl/util/string.hpp
+++ b/include/mbgl/util/string.hpp
@@ -10,7 +10,16 @@ namespace util {
 
 template <class T>
 inline std::string toString(T t) {
-    return std::to_string(t);
+    if (t - (int)t > 0.0 || t + (int)t < 0.0) {
+        std::string str = std::to_string(t);
+        str.erase(str.find_last_not_of('0') + 1, std::string::npos);
+        if (str.back() == '.') {
+            str.pop_back();
+        }
+        return str;
+    } else {
+        return std::to_string(int(t));
+    }
 }
 
 inline std::string toString(int8_t num) {
--- a/test/util/number_conversions.cpp
+++ b/test/util/number_conversions.cpp
@@ -6,6 +6,12 @@
 using namespace mbgl;
 
 TEST(NumberConversions, number_to_string) {
+    EXPECT_EQ("0", util::toString(0));
+    EXPECT_EQ("10", util::toString(10));
+    EXPECT_EQ("-10", util::toString(-10));
+    EXPECT_EQ("10", util::toString(10.0));
+    EXPECT_EQ("-10", util::toString(-10.0));
+
     EXPECT_EQ("1", util::toString(1));
     EXPECT_EQ("-1", util::toString(-1));
     EXPECT_EQ("32768", util::toString(32768));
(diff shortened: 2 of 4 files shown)


test/util/number_conversions.cpp
```diff
@@ -0,0 +1,38 @@
...
+
+#include <mbgl/util/string.hpp>
+
+using namespace mbgl;
+
+TEST(NumberConversions, number_to_string) {
+    EXPECT_EQ("1", util::toString(1));
+    EXPECT_EQ("-1", util::toString(-1));
+    EXPECT_EQ("32768", util::toString(32768));
+    EXPECT_EQ("-32768", util::toString(-32768));
```
1ec5: Other things to test:

* `0` → `0`
* `-0` → `-0` (not sure about this one)
* `10` → `10`
* `-10` → `-10`

include/mbgl/util/string.hpp
```diff
@@ -10,7 +10,16 @@ namespace util {
 
 template <class T>
 inline std::string toString(T t) {
-    return std::to_string(t);
+    if (t - (int)t > 0.0 || t + (int)t < 0.0) {
```
1ec5: `std::to_string()` can also be used with numeric types much larger than `int`, as well as unsigned types. Casting an `unsigned` or `long double` to an `int` risks overflow and other unexpected behavior.

include/mbgl/util/string.hpp
```diff
@@ -10,7 +10,16 @@ namespace util {
...
+        std::string str = std::to_string(t);
+        str.erase(str.find_last_not_of('0') + 1, std::string::npos);
+        if (str.back() == '.') {
```
1ec5: According to the [`std::to_string()`](http://www.cplusplus.com/reference/string/to_string/) documentation, the output matches that of `printf()`. `printf()` is affected by the current locale, which isn’t necessarily the default C locale. If you were to call `std::locale::global(std::locale("fr"))`, the decimal separator would be `,`, not `.`.

I actually have no idea if other parts of mbgl assume the C locale as this code does, but I suspect that would be problematic on some of the platforms we support.

friedbunny: Interesting, I definitely did not consider these strings as being localizable — is this a realistic scenario right now?

A possible fix would be to remove any non-number that’s the final character.

include/mbgl/util/string.hpp
friedbunny: This was one of my uncertainties, working on this — I don’t know what types are ever likely to pass through here. Should we be checking for larger numeric types in this function, or break out floats into a more specific `toString()` of its own?

include/mbgl/util/string.hpp
friedbunny: One data point: iOS system region currently has no effect on what decimal separator is used here. The separator is still a decimal point with the region set to German, which is “4.567,89” for 4,567.89 in the US.

include/mbgl/util/string.hpp
1ec5: `std::to_string()` can be used with a variety of numeric types. This function ideally shouldn’t restrict the domain of that function at all. The particular bug you’re trying to fix occurs by way of [these lines](https://github.com/mapbox/mapbox-gl-native/blob/5837ba1ad3cd84954711c6d42ccc1358da8a09d3/src/mbgl/renderer/symbol_bucket.cpp#L137-L142), particularly the `double` case. In fact, if you look below here, there are already overrides for `int8_t` and `uint8_t`: I think a proper fix would come in the form of an override for `double`, not a rewrite of this generic function.

include/mbgl/util/string.hpp
1ec5: > iOS system region currently has no effect on what decimal separator is used here.

Yes, this is more likely to be a problem on desktop systems, that sort of thing. There’s also potentially a distinction between the current C locale and current C++ locale, and the C locale is more likely to be affected.

Rather than delve into the complexities of number formatting, I think it’d be more robust to rely on existing string formatters. I’m sure we can find something between C-style `sprintf()` and C++-style [`stringstream`](http://www.cplusplus.com/reference/sstream/stringstream/).

test/util/number_conversions.cpp
friedbunny: Did this in d09cdb4a2dbc53ff2515e09777b17beed2c57b8f.

include/mbgl/util/string.hpp
friedbunny: Yeah, it makes much more sense to have `toString(double num)`. I did this in e9d16dc07f5b1380ea6e66601548007860572bd4 and it seems to be working at least as well.

`long double` is not specifically handled ([see this test line](https://github.com/mapbox/mapbox-gl-native/pull/4965/files#diff-cb902609ef74233feb733fddb90c8b86R48)), but I think that’s an edge case we’re OK not covering.

include/mbgl/util/string.hpp
```diff
@@ -13,6 +13,19 @@ inline std::string toString(T t) {
     return std::to_string(t);
 }
 
+inline std::string toString(double num) {
+    if (num - (int)num > 0.0 || num + (int)num < 0.0) {
```
1ec5: I think this can be expressed as `if (std::fmod(num, 1))`. `std::fmod()` and `fmod()` are floating-point-friendly versions of the `%` operator.

friedbunny: Nice, I originally went with this (silly) way because `num` didn’t have a reliable type, but now we can make this better.

include/mbgl/util/string.hpp
```diff
@@ -13,6 +13,19 @@ inline std::string toString(T t) {
...
+        std::string str = std::to_string(num);
+        str.erase(str.find_last_not_of('0') + 1, std::string::npos);
+        if (str.back() == '.') {
+            str.pop_back();
+        }
+        return str;
+    } else {
+        return std::to_string(int(num));
```
1ec5: `DBL_MAX` ≫ `INT_MAX`

1ec5:
@jfirebaugh, would there be any disadvantage to using `std::ostringstream` instead of `std::to_string`? It defaults to the equivalent of the `%g` printf specifier, which seems to be what we want for just about any numeric type.

1ec5:
Remember to update all the relevant changelogs.

jfirebaugh:
I think using `std::ostringstream` would be fine. However, what this suggests to me is that `toString` is probably not a useful abstraction. What I propose instead is:

* Drop `toString.
* Use `std::to_string` directly where we really don't care much about presentation (e.g. debug logging).
* Create a special purpose formatter, using `std::ostringstream` or whatever we land on, [in `SymbolBucket`](https://github.com/mapbox/mapbox-gl-native/blob/7d0d09e7dc5157338c4c9bcfd1488f27cdb6338f/src/mbgl/renderer/symbol_bucket.cpp#L129-L144).

1ec5:
> If this passes muster, we should consider including it in iOS v3.2.2 and the next Android release.

Has anyone been able to reproduce this issue in iOS SDK v3.2.1? If not, there’s no need to backport the fix, it may be too late for 3.2.2 anyways.

friedbunny:
We don’t need this for v3.2.x — I think I had a bork build at one point that did suffer from this issue, but subsequent builds and the RC work as expected.

kkaefer:
Can we use RapidJSON's [dtoa.h](https://github.com/miloyip/rapidjson/blob/master/include/rapidjson/internal/dtoa.h) implementation instead of `std::stringstream`? AFAIK, `dtoa.h` is much faster than `stringstream`

1ec5:
If we're concerned about performance, perhaps `sprintf()` or similar would be worth trying. It's literally the `%g` specifier that we're after.

kkaefer:
@1ec5 `sprintf` is really slow as well.

See https://github.com/miloyip/dtoa-benchmark and https://github.com/miloyip/itoa-benchmark for benchmarks.

kkaefer:
I'm interested in optimizing this because I've seen this be a bottleneck in the past, because we're doing a lot of number => string conversions when parsing tiles.

friedbunny:
Unassigned myself — this can be implemented more efficiently and precisely by someone else.

1ec5:
Superseded by #5054.

PR #4966, opened 2016-05-07
[core] Normalize tile coordinates in VectorTile

jfirebaugh:
Previously, vector tile coordinates were being normalized to `util::EXTENT` in each place they were used, and for annotation and GeoJSON sources that don't require normalization.


jfirebaugh:
@ansis Do you see any issues with this approach?

PR #4980, opened 2016-05-09
[Qt] Use C++11 ABI packages

### LINKED ISSUE #3123: Compilation with gcc5 fails due to ABI break
When trying to build mapbox-gl-native with gcc/g++5 or later, the build fails with a linker error:
```
./Release/obj.target/gyp/libmbgl-core.a(shape_annotation_impl.o): In function `mbgl::ShapeAnnotationImpl::updateTile(mbgl::TileID const&, mbgl::AnnotationTile&)':
/tmp/mapbox-gl-native/build/linux-x86_64/../../src/mbgl/annotation/shape_annotation_impl.cpp:108: undefined reference to `mapbox::util::geojsonvt::Convert::create(std::map<std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> >, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> >, std::less<std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > >, std::allocator<std::pair<std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > > > >, mapbox::util::geojsonvt::ProjectedFeatureType, mapbox::util::variant<mapbox::util::geojsonvt::ProjectedPoint, mapbox::util::geojsonvt::ProjectedGeometryContainer>)'
```
The issue is likely caused by the ABI break introduced by gcc/g++5 for supporting C++11 and the respective adjustments to std::string/std::list. The respective symbol names are mangled with a std::__cxx11 prefix. The geojsonvt static library managed via mason is likely built with gcc/g++ 4.9 and the required symbols are not named correctly and thus not resolved by the linker.

One possible solution could be to force gcc 4.9 in the build system setup and error out if gcc 4.9 is not available on the build machine.

tmpsantos:
Was causing linking errors on more recent Linux distros.

Fixes #3123.


tmpsantos:
:eyes:  @jfirebaugh @brunoabinader

brunoabinader:
Clever solution @tmpsantos - there seems to be no easier way other than using `readelf` 👍

jfirebaugh:
This seems fine, but I'm also interested to know what it would take to just cut over to cxx11abi universally. Would installing a newer libstdc++ on Travis and compiling everything with `_GLIBCXX_USE_CXX11_ABI=1` do it?

tmpsantos:
> This seems fine, but I'm also interested to know what it would take to just cut over to cxx11abi universally. Would installing a newer libstdc++ on Travis and compiling everything with _GLIBCXX_USE_CXX11_ABI=1 do it?

Probably yes. In fact, recompiling everything on a Ubuntu 16.04 with no special flags should do it, but everyone else using 14.04 won't be happy.

I suggest we merge this for now until 16.04 gets a broader adoption or until any other package gives us headache. IMO we are good because most of our deps are header-only.

PR #4981, opened 2016-05-09
Fix context menu on OS X

### LINKED ISSUE #3493: Missing context menu
Since #3426, right-clicking in osxapp no longer displays a context menu.

1ec5:
This PR fixes the context menu on map views in osxapp so that the menu appears and is enabled in the right situations. These connections were lost in #3427.

drop
remove

As in MapKit, a two-finger tap (the “walking fingers” gesture) should be interpreted as a secondary click to open the context menu, if applicable. It should not zoom out; rather, a double-tap on a Magic Mouse should zoom out (“smart magnification”).

This PR also fixes a longstanding issue that made the osxapp context menu only apply to the right-clicked pin after first selecting the pin. It also happened to be a latent bug in the iOS implementation.

Fixes #3493.

/cc @kkaefer

### DIFF
--- a/platform/ios/src/MGLMapView.mm
+++ b/platform/ios/src/MGLMapView.mm
@@ -3142,9 +3142,9 @@ mbgl::Duration MGLDurationInSeconds(NSTimeInterval duration)
             }
             
             // Choose the first nearby annotation.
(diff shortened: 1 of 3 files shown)


kkaefer:
Behavior is correct 👍

PR #4990, opened 2016-05-10
Don't lookup sources constantly

kkaefer:
Every time we're accessing a layer's source, we're looking it up by it's ID. However, they never change, so we can just store the pointer directly in the Layer object.


kkaefer:
👀  @jfirebaugh

jfirebaugh:
We used to do this, but I switched to ID-based indirection to eliminate `Source` as a type of object that's shared with worker threads (#667). I don't think it's actually needed by worker threads, but if it's accessible on layer, than it risks being used accidentally.

jfirebaugh:
Closing; I think the layer of indirection here is overall a good thing.

PR #4995, opened 2016-05-10
Merge iOS SDK 3.2.2 release branch into master

1ec5:
Looks like this ropes in some additional changes for v3.2.1 that were never merged back to master. I decided to keep the changes added in 8ecb20610bc633ee6201bc2b0a668301140d8fc1 to make sure we can handle arbitrary versions.

/cc @boundsj @friedbunny


friedbunny:
Thanks. 💪

PR #4997, opened 2016-05-10
[ios] fix potential for very slow animations of user location dot

bwhtmn:
It seems there was a logic error in calculating the duration for animating the user's current location dot. Since the code used the MAX of 1 second, or the time between the current location and previous location, it could easily result in animations of 30 seconds or more in cases where there's a gap in location updates. This could happen in common situations such as when the user is indoors. The result was a very slow and strange movement of the location dot. Pretty sure MIN is more appropriate here, as it puts a cap on the animation duration.


1ec5:
Thank you!

PR #4998, opened 2016-05-11
Fix nested framework

1ec5:
An error in the packaging script caused Mapbox.framework to be placed inside a folder named Mapbox.framework. This change moves the framework up one level.


1ec5:
This change overcorrected, putting all the framework bundle’s contents one level too high. 5f1144f05a8391cf2c01dc7a67312a27ef32805b corrects the overcorrection.

PR #5001, opened 2016-05-11
[core] [Qt] Added viewport mode

### LINKED ISSUE #4866: [Qt] Map is flipped on the Y axis for Qt versions < 5.6
Qt assumes a top-down coordinate system, which goes against traditional OpenLG bottom-up coordinate system currently in use by Mapbox GL. This results in a Y-axis flipped map for the Qt Quick component:
screen shot 2016-04-27 at 3 55 47 pm
Qt provides a property [QQuickFramebufferObject::mirrorVertically](http://doc.qt.io/qt-5/qquickframebufferobject.html#mirrorVertically-prop) to address this, but it is only available in Qt 5.6. Thus, we need to find a solution for using the Qt Quick component between versions 5.2 to 5.6.

/cc @tmpsantos @kkaefer

brunoabinader:
_Viewport mode_ satisfies embedding platforms that requires the viewport coordinate systems to be set according to its standards e..g. Qt Quick requires third-party rendering code to follow its standards with regards to viewport transforms.
 
Qt 5.6 adds a [mirrorVertically](http://doc.qt.io/qt-5/qquickframebufferobject.html#mirrorVertically-prop) property that solves the solution for that version of Qt onwards. Telling Mapbox GL core to flip the Y axis makes `QQuickMapboxGL` item available from Qt 5.2 onwards.

Fixes #4866.

:eyes: @kkaefer @tmpsantos

### DIFF
--- a/include/mbgl/map/map.hpp
+++ b/include/mbgl/map/map.hpp
@@ -34,7 +34,8 @@ public:
     explicit Map(View&, FileSource&,
                  MapMode mapMode = MapMode::Continuous,
                  GLContextMode contextMode = GLContextMode::Unique,
-                 ConstrainMode constrainMode = ConstrainMode::HeightOnly);
+                 ConstrainMode constrainMode = ConstrainMode::HeightOnly,
+                 ViewportMode viewportMode = ViewportMode::Default);
     ~Map();
 
     // Register a callback that will get called (on the render thread) when all resources have
@@ -117,10 +118,14 @@ public:
     // North Orientation
     void setNorthOrientation(NorthOrientation);
     NorthOrientation getNorthOrientation() const;
-    
+
     // Constrain mode
     void setConstrainMode(ConstrainMode);
     ConstrainMode getConstrainMode() const;
+
+    // Viewport mode
+    void setViewportMode(ViewportMode);
+    ViewportMode getViewportMode() const;
 
     // Size
     uint16_t getWidth() const;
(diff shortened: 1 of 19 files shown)


include/mbgl/map/map.hpp
```diff
@@ -34,6 +34,7 @@ class Map : private util::noncopyable {
     explicit Map(View&, FileSource&,
                  MapMode mapMode = MapMode::Continuous,
                  GLContextMode contextMode = GLContextMode::Unique,
+                 ViewportMode viewportMode = ViewportMode::Default,
```
kkaefer: Any reason to add this before `ConstraintMode`?

brunoabinader: Rationale is that properties that affect code rendering gets higher priority in order of appearance - but I don't hold a strong opinion on this.

brunoabinader: On a second thought - for the sake of adding new features at the tail - I'm going to add this after `ConstrainMode` 👍

tmpsantos:
:+1:

1ec5:
OS X puts the origin in the bottom-left. Should the OS X SDK set the viewport mode too?

kkaefer:
@1ec5 that's the way OpenGL renders by default; there's nothing OS X-specific about it. For Qt, it looks like they're trying to be smart and somehow mirror the coordinate system without giving you the option to use the OpenGL-default until 5.6, so we'll manually have to alter our rendering.

PR #5002, opened 2016-05-11
Remove View dependency in Transform

kkaefer:
The `Transform` object currently depends on `View` to deliver notifications. We should remove this and simply pass in a function pointer/`std::function` that can be optional.


brunoabinader:
Oops, PR #5001 added a new parameter to `Transform` ctor - rebase?

PR #5008, opened 2016-05-11
[android] Animate MylocationView direction

### LINKED ISSUE #4942: Animate bearing drawable when direction changes with MyLocationView
Currently we set the direction of the bearing drawable immediately when the direction changes (due to compass or a GPS sensor update). We should animate this using an Animator instead as we used to do with previous implementation. The code from there could not directly be ported in #4396 so I opted to not implement it at that time and create a follow up issue instead.

cammace:
Fixes #4942 added ObjectAnimator to the MyLocationView direction. Worked on my phone but occasionally jumps around still. 

@tobrun @zugaldia @bleege

### DIFF
--- a/platform/android/MapboxGLAndroidSDK/src/main/java/com/mapbox/mapboxsdk/maps/widgets/MyLocationView.java
+++ b/platform/android/MapboxGLAndroidSDK/src/main/java/com/mapbox/mapboxsdk/maps/widgets/MyLocationView.java
@@ -429,17 +429,18 @@ public class MyLocationView extends View {
 
     private class CompassListener implements SensorEventListener {
 
-        private boolean paused;
         private SensorManager mSensorManager;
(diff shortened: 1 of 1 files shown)


tobrun:
@cammace 88e01d0 fixes the jumpiness and additionally some fixes to GPS bearing. To close this issue, we need to look into one more rotation issue. I'm currently not able to rotate the compass 180 degrees. It's always stuck pointing somewhere from east-south-west, never pointing to north direction. At this point I'm not sure if it's a bug introduced with the new way of setting rotation or my device sensors misbehaving.

tobrun:
Did another test and seeing the same results as mentioned in https://github.com/mapbox/mapbox-gl-native/pull/5008#issuecomment-218977966. Will need to dig deeper and find the cause of this.

tobrun:

PR #5011, opened 2016-05-12
[android] - fixes style init bug, added unit test to prevent regression

tobrun:
Fixes the issue that the style field within the initialiser is set to null.
Backed up these changes with unit tests.


tobrun:

PR #5013, opened 2016-05-12
Remove indirection in DefaultFileSource

kkaefer:
In `DefaultFileSource`, we're constructing a new object called `Task`, which just holds another `unique_ptr`. We can just do everything inline.

### DIFF
--- a/platform/default/default_file_source.cpp
+++ b/platform/default/default_file_source.cpp
@@ -25,29 +25,6 @@ namespace mbgl {
 
 class DefaultFileSource::Impl {
 public:
-    class Task {
-    public:
-        Task(Resource resource, FileSource::Callback callback, DefaultFileSource::Impl* impl) {
-            auto offlineResponse = impl->offlineDatabase.get(resource);
-
-            Resource revalidation = resource;
-
-            if (offlineResponse) {
-                revalidation.priorModified = offlineResponse->modified;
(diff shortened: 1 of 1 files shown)


kkaefer:
cc @jfirebaugh

platform/default/default_file_source.cpp
```diff
@@ -132,7 +123,7 @@ class DefaultFileSource::Impl {
 
     OfflineDatabase offlineDatabase;
     OnlineFileSource onlineFileSource;
-    std::unordered_map<AsyncRequest*, std::unique_ptr<Task>> tasks;
+    std::unordered_map<AsyncRequest*, std::unique_ptr<AsyncRequest>> tasks;
```
kkaefer: Why is this an `unordered_map`? Pointers can be compared very fast and have their own well-defined order already, so creating a hash of a pointer just adds a level of indirection.

jfirebaugh:
👍  I don't know why I added the indirection to begin with.

platform/default/default_file_source.cpp
jfirebaugh: Per chat: using `unordered_map` is usually preferable when you don't need ordering because search, insertion, and removal are amortized constant time complexity, compared to `map`'s logarithmic complexity.

PR #5022, opened 2016-05-13
[ios] Check that telemetry test server string is URL-like

friedbunny:
Do a _little_ bit more validation on the telemetry test server URL. HTTPS-only because this deals in location information.

Follows up on #4691.

/cc @boundsj @1ec5

### DIFF
--- a/platform/ios/src/MGLAPIClient.m
+++ b/platform/ios/src/MGLAPIClient.m
@@ -14,7 +14,7 @@ static NSString * const MGLAPIClientHTTPMethodPost = @"POST";
 @interface MGLAPIClient ()
 
 @property (nonatomic, copy) NSURLSession *session;
-@property (nonatomic, copy) NSString *baseURL;
+@property (nonatomic, copy) NSURL *baseURL;
 @property (nonatomic, copy) NSData *digicertCert;
 @property (nonatomic, copy) NSData *geoTrustCert;
 @property (nonatomic, copy) NSData *testServerCert;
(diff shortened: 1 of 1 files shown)


platform/ios/src/MGLAPIClient.m
```diff
@@ -88,7 +88,7 @@ - (NSURLRequest *)requestForEvents:(NS_ARRAY_OF(MGLMapboxEventAttributes *) *)ev
 
 - (void)setupBaseURL {
     NSString *testServerURL = [[NSUserDefaults standardUserDefaults] stringForKey:@"MGLTelemetryTestServerURL"];
-    if (testServerURL) {
+    if (testServerURL && [testServerURL hasPrefix:@"https://"]) {
```
1ec5: We should check whether the URL is well-formed, by checking whether an NSURL can be initialized from it. Also, with the NSURL, we can compare its `scheme` property.

friedbunny: I went whole hog and refactored everything to use NSURL in 2510d116bd88dc5a7c5e46665d4b63a63d9cc0ee.

friedbunny:
This expanded a bit with 2510d116bd88dc5a7c5e46665d4b63a63d9cc0ee to include refactoring `baseURL` from NSString to NSURL.

boundsj:
👍

platform/ios/src/MGLAPIClient.m
```diff
@@ -76,8 +76,9 @@ - (void)cancelAll {
 #pragma mark Utilities
 
 - (NSURLRequest *)requestForEvents:(NS_ARRAY_OF(MGLMapboxEventAttributes *) *)events {
-    NSString *url = [NSString stringWithFormat:@"%@/%@?access_token=%@", self.baseURL, MGLAPIClientEventsPath, [MGLAccountManager accessToken]];
-    NSMutableURLRequest *request = [[NSMutableURLRequest alloc] initWithURL:[NSURL URLWithString:url]];
+    NSString *path = [NSString stringWithFormat:@"%@?access_token=%@", MGLAPIClientEventsPath, [MGLAccountManager accessToken]];
```
1ec5: In the future, we should consider making the user default hold only the hostname, so that we can built the URL more robustly with NSURLComponents. (For example, we’d set the `scheme` to `https`, no questions asked.)

friedbunny: That’s a fine idea — I could go ahead and just do that now, this code is already in flux for v3.3.0 and we’re the only ones that use it.

friedbunny: Actually, going to punt on NSURLComponents for now as the current code has proven workable.

PR #5023, opened 2016-05-13
Separate frontend from backend FileSources

kkaefer:
We're currently using the FileSource interface for three distinct things:

- The "frontend" part that users can implement (e.g. in the node bindings) and that `DefaultFileSource` provides
- The "backend" parts that `OnlineFileSource`, `AssetFileSource` implement
- The `HTTPFileSource`, which is second-tier backend to `OnlineFileSource`

There's no reason for the latter two to use the same interface and go through the virtual calls.


jfirebaugh:
Strong 👎 here. Using a consistent interface for requesting resources made the refactoring required for offline much easier, keeps these components loosely coupled (relying only on an abstract interface and not concrete implementations), and allows us to pass in stub/mock implementations of `FileSource` for testing. The virtual function dispatch is not meaningful for performance at the frequency at which file requests are made.

kkaefer:
Reason for this is that they're conceptually very different: The `DefaultFileSource` does various lookups, and can call the callback multiple times, whereas `AssetFileSource` and `HTTPFileSource` never do this.

My ultimate goal is to be able to split up `FileSource::request()` into a method that only loads "optional" resources (e.g. from cache, or filesystem) and "required" resources, which tries to request them from the server as well.

jfirebaugh:
> My ultimate goal is to be able to split up `FileSource::request()` into a method that only loads "optional" resources (e.g. from cache, or filesystem) and "required" resources, which tries to request them from the server as well.

Is this something that we can accomplish by adding an optional hint parameter to `Resource` -- alongside the existing options like prior request headers -- which particular `FileSource` implementations can examine if they choose?

I view the fact that `FileSource::request` is so generic as an advantage: it allows a wide variety of implementations to fit behind it, while presenting a simple and uniform interface to `Map`. We take advantage of this particularly for the node bindings, but I can imagine use cases where it would be useful to be able to pass say `HTTPFileSource` or `AssetFileSource` as the `Map`'s `FileSource`, e.g. maps on a device with a read-only filesystem, or permanently offline devices. So if possible I'd like to keep everything inheriting from a single interface, with a single method.

kkaefer:
Thanks, those are good reasons. I went down the route you suggest in #5027

PR #4889, opened 2016-05-13
[Qt] QMapboxGL pixel ratio is wrong for Retina displays

brunoabinader:
screen shot 2016-04-28 at 2 50 45 pm
Some issues to consider:
- `Map::Impl` only reads the pixel ratio once in the ctor. In my environment, I have an external  monitor connected to my MacBook. When moving the application window between them, the pixel ratio should be updated.
- We're hardcoding the pixel ratio to 1 in https://github.com/mapbox/mapbox-gl-native/blob/master/platform/qt/src/qmapboxgl.cpp#L597.

/cc @tmpsantos

### DIFF
--- a/platform/qt/app/mapwindow.cpp
+++ b/platform/qt/app/mapwindow.cpp
@@ -149,13 +149,16 @@ void MapWindow::initializeGL()
 
 void MapWindow::resizeGL(int w, int h)
 {
-    m_map.resize(QSize(w, h));
+    QSize size(w, h);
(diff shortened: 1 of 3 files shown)


brunoabinader:
Correct aspect rations on Retina display:

screen shot 2016-05-13 at 4 31 33 pm
screen shot 2016-05-13 at 4 31 46 pm

Remaining issue related to pixel ratio moved to #5029.

PR #5028, opened 2016-05-13
Move from SystemTimePoint to Timestamp

kkaefer:
We're currently using a timepoint type that has microsecond precision, but since all of our timestamps are from HTTP fields, we don't need that. In fact, it's causing subtle bugs because seemingly identical timestamps differ on the fractional level, so comparison fails.

### DIFF
--- a/src/mbgl/util/http_header.cpp
+++ b/src/mbgl/util/http_header.cpp
@@ -25,12 +25,8 @@ CacheControl CacheControl::parse(const std::string& value) {
     return result;
 }
 
-optional<SystemTimePoint> CacheControl::toTimePoint() const {
-    // Round trip through time_t to truncate fractional seconds.
-    return maxAge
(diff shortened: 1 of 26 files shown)


src/mbgl/util/http_header.cpp
```diff
@@ -25,12 +25,8 @@ CacheControl CacheControl::parse(const std::string& value) {
     return result;
 }
 
-optional<SystemTimePoint> CacheControl::toTimePoint() const {
-    // Round trip through time_t to truncate fractional seconds.
```
kkaefer: This is no longer necessary, since we're not storing any fractional seconds.

jfirebaugh:
Sounds good to me. Curious though, where are we making exact equality comparisons?

kkaefer:
Ran into this issue while working on https://github.com/mapbox/mapbox-gl-native/issues/5027

Issue #4844, opened 2016-04-26
Simplify HTTP resource requesting


kkaefer:
Currently, the requests we issue to obtain data via HTTP first get an (optional) response served from the cache, no matter whether the data in the cache was expired. They also (optionally) get a second response with data that is always fresh. Instead of having two responses per request, we should always exactly return one response and change the behavior so that the first request is always served from cache (if possible). When the client got a stale resource, it should make a second request with appropriate headers that force a non-stale response.

jfirebaugh:
I don't think I understand the change you're proposing, or what part of HTTP requesting it would simplify. Can you explain again?

kkaefer:
We're currently outsourcing the retry logic into the HTTP source. However, for #123 to work, we need to make the decision of whether to attempt to verify data freshness directly inside the renderer. I'm envisioning this order of attempting to load resources:

![a](https://cloud.githubusercontent.com/assets/52399/15153085/35a472f0-16d8-11e6-8e86-73985a285b84.png)

This would mean that we'll need closer control over the order these requests happen in.

kkaefer:
In particular, cache lookups and HTTP requests need to be decoupled so that we can first do a cache lookup, and once the tile becomes an ideal tile (through zooming), we'll want to go the second step and ensure freshness.

kkaefer:
Blocked by #5023

jfirebaugh:
#5027 implemented what we needed for #5143.

PR #5027, opened 2016-05-13
Allow requesting optional resources

kkaefer:
Right now, when calling `FileSource::request()`, we're making every effort to obtain the data. However, for #123, we need a more relaxed way to request things that can be obtained with low effort (e.g. by loading it from a cache, but *not* downloading it from the internet).

### DIFF
--- a/platform/default/default_file_source.cpp
+++ b/platform/default/default_file_source.cpp
@@ -82,21 +82,35 @@ public:
     }
 
     void request(AsyncRequest* req, Resource resource, Callback callback) {
-        auto offlineResponse = offlineDatabase.get(resource);
-
         Resource revalidation = resource;
 
-        if (offlineResponse) {
-            revalidation.priorModified = offlineResponse->modified;
-            revalidation.priorExpires = offlineResponse->expires;
-            revalidation.priorEtag = offlineResponse->etag;
-            callback(*offlineResponse);
+        const bool hasPrior = resource.priorEtag || resource.priorModified || resource.priorExpires;
+        if (!hasPrior || resource.necessity == Resource::Optional) {
+            auto offlineResponse = offlineDatabase.get(resource);
+
+            if (resource.necessity == Resource::Optional && !offlineResponse) {
+                // Ensure there's always a response that we can send, so the caller knows that
+                // there's no optional data available in the cache.
+                offlineResponse.emplace();
+                offlineResponse->noContent = true;
+                offlineResponse->error = std::make_unique<Response::Error>(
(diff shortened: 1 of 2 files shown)


jfirebaugh:
Looks good! (For anyone else reviewing, the addition of `Resource::necessity` was in cd65a43855f33555eca5f3e3ad5d37661253209f.)

Checking my understanding of how this will be used (c.f. https://github.com/mapbox/mapbox-gl-native/issues/4844):

1. Source will make an `Optional` request for the ideal tile.
2. If unfulfilled, it will make a `Required` request for the ideal tile. Concurrently, it will make `Optional` requests for successive parent tiles until either one of those is fulfilled, or the `Required` request for the ideal tile is fulfilled.
3. Otherwise, if fulfilled, it will make a `Required` request for the ideal tile, passing the `prior` headers so that the request doesn't go back through the cache pathway.

Is this right? The (minor) gap I see with this strategy is that in step 2, it looks like the `Required` request for the ideal tile will go back through the cache pathway, even though there's known to be nothing there.

kkaefer:
> For anyone else reviewing, the addition of Resource::necessity was in cd65a43

Yeah, that was an oversight; sorry about that.

> Is this right?

Yup, correct. Essentially it'll climb up the tile pyramid to see if something is in the cache without requesting from the internet that is not an ideal tile.

> The (minor) gap I see with this strategy is that in step 2, it looks like the Required request for the ideal tile will go back through the cache pathway, even though there's known to be nothing there.

In step 2, it knows that the request is unfulfilled and can set `priorExpires` to `0`, which will make `DefaultFileSource` skip the cache. I'm wondering if we should make this behavior more explicit, e.g. by changing `necessity` to something that explicitly tells the `FileSource` that it should make a cache request xor http request, however, I'm not sure how to name such a thing without also implying that something must be loaded from the internet.

jfirebaugh:
Another potential approach would be to make the initial request `Required`, have the cache miss reported as `noContent` regardless of `necessity`, and add some other flag to the response that allows the requester to distinguish between a cache hit or miss:

1. Source makes an `Required` request for the ideal tile.
2. If the cache hits, it's done.
3. If the cache misses, it keeps the request open, but also makes `Optional` requests for successive parent tiles until either one of those is fulfilled, or the `Required` request for the ideal tile is fulfilled.

kkaefer:
@jfirebaugh We need to transition the requests between `Required` and `Optional` state when zooming in/out, so in that case, we don't have a way to ensure that no HTTP request is made after transitioning from `Required` to `Optional`.

PR #5048, opened 2016-05-17
Check For Asyc Approval Of Location Permissions

### LINKED ISSUE #5045: Add Checks To Start Telemetry In Android 6+ After Permissions Granted
Prior to Android 6.0 user granted permissions were granted at app install time which means that Ambient Telemetry would start immediately as part of the Telemetry start up process.  This is not the case with Android 6.0 as permissions are now approved in app via dialog (similar to iOS).  This change makes it likely that app's starting for the first time may not have had location permissions approved, thus making it that Ambient Telemetry mode doesn't start the first time the app is run.

Unfortunately the Android API for listening for Permissions granted is [tied specifically to Activities](http://developer.android.com/reference/android/support/v4/app/ActivityCompat.OnRequestPermissionsResultCallback.html#onRequestPermissionsResult%28int,%20java.lang.String[],%20int[]%29) and not generic callbacks (ie, Anonymous Classes).  This leaves two options available for implementing this is in a manor that doesn't impose any responsibilities upon implementing app developers:

* Run a timed loop at startup (if permission not already generated) that checks for permissions granted status
* Have Mapbox ask for Location Permissions

I believe the timed loop option is the better choice as it doesn't force the developer's hand as to when the Location permissions would be requested (ie, shown to the user) in Android 6+.

bleege:
Closes #5045

### DIFF
--- a/platform/android/MapboxGLAndroidSDK/src/main/java/com/mapbox/mapboxsdk/location/LocationServices.java
+++ b/platform/android/MapboxGLAndroidSDK/src/main/java/com/mapbox/mapboxsdk/location/LocationServices.java
@@ -74,8 +74,7 @@ public class LocationServices implements com.mapzen.android.lost.api.LocationLis
      * @param enableGPS true if GPS is to be enabled, false if GPS is to be disabled
      */
     public void toggleGPS(boolean enableGPS) {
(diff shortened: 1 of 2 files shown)


bleege:
Testing this morning confirmed what was need to be confirmed.  I'm going to Squash and Merge now!

PR #5053, opened 2016-05-17
[ios] Make MGLCoordinateInCoordinateBounds() public

friedbunny:
Makes the formerly-private `MGLCoordinateInCoordinateBounds()` function publicly available.

- The old private function converted `MGLCoordinateBounds`→`mbgl::LatLngBounds` and `CLLocationCoordinate2D`→`mbgl:LatLng`.

- The new public function compares `CLLocationCoordinate2D` and `MGLCoordinateBounds` directly.

/cc @1ec5

### DIFF
--- a/platform/darwin/src/MGLGeometry.h
+++ b/platform/darwin/src/MGLGeometry.h
@@ -64,6 +64,14 @@ NS_INLINE BOOL MGLCoordinateBoundsEqualToCoordinateBounds(MGLCoordinateBounds bo
             bounds1.ne.longitude == bounds2.ne.longitude);
 }
 
+/** Returns `YES` if the coordinate is within the coordinate bounds. */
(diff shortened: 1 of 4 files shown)


friedbunny:
One potential issue I’ve encountered with this simplified function is that it strictly requires the bounds to be NE, SW. The old function would fix screwy bounds (e.g., where NW/SE was used) by creating a convex hull from that bound's points — essentially, always recreating the bounds.

1ec5:
Looks good to me.

> The old function would fix screwy bounds (e.g., where NW/SE was used) by creating a convex hull from that bound's points — essentially, always recreating the bounds.

We've also discussed flipped coordinates as a way of indicating a bounds that spans the antimeridian. Whatever the case, perhaps we need an MGLCoordinateBoundsStandardize(), by analogy with CGRectStandardize().

PR #5054, opened 2016-05-17
Use rapidjson's dtoa implementation to stringify floats

### LINKED ISSUE #998: Text from floating point data rounding error

kkaefer:
Fixes #998 and excess trailing zeros introduced by #4916.

Supersedes #4965.

cc @friedbunny @1ec5

### DIFF
--- a/src/mbgl/util/dtoa.cpp
+++ b/src/mbgl/util/dtoa.cpp
@@ -1,0 +1,97 @@
+#include "dtoa.hpp"
+
+#include <rapidjson/internal/dtoa.h>
+
+namespace mbgl {
+namespace util {
+
+namespace {
+
+// From https://github.com/miloyip/rapidjson/blob/master/include/rapidjson/internal/dtoa.h
+
+inline char* Prettify(char* buffer, int length, int k) {
+    constexpr int maxDecimalPlaces = 324;
+    const int kk = length + k;  // 10^(kk-1) <= v < 10^kk
+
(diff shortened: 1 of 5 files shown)


src/mbgl/util/dtoa.cpp
```diff
@@ -0,0 +1,97 @@
+#include "dtoa.hpp"
+
+#include <rapidjson/internal/dtoa.h>
```
1ec5: Is it OK to be including an internal header like this?

kkaefer: It's internal, but RapidJSON is header only, so we're not going to run into ABI-incompatibilities, and we'll notice as soon as they change the internal function signatures (which is unlikely).

friedbunny:
Thanks, @kkaefer. 🎯

PR #5062, opened 2016-05-18
[ios] Added centerOffset to MGLAnnotationView

### LINKED ISSUE #5059: Add centerOffset property to MGLAnnotationView
MapKit has a [centerOffset property available on MKAnnotationView](https://developer.apple.com/library/ios/documentation/MapKit/Reference/MKAnnotationView_Class/#//apple_ref/occ/instp/MKAnnotationView/centerOffset) that allows one "to reposition the annotation view as needed" from the default center point of the annotation model object itself. We should add this feature to MGLAnnotationView.

fyi @1ec5 @friedbunny @tobrun

frederoni:
closes #5059

### DIFF
--- a/platform/ios/CHANGELOG.md
+++ b/platform/ios/CHANGELOG.md
@@ -26,6 +26,7 @@ Mapbox welcomes participation and contributions from everyone.  Please read [CON
 - Removed unused SVG files from the SDK’s resource bundle. ([#4641](https://github.com/mapbox/mapbox-gl-native/pull/4641))
 - Deprecated `-[MGLMapView emptyMemoryCache]`. ([#4725](https://github.com/mapbox/mapbox-gl-native/pull/4725))
 - Added `MGLCoordinateInCoordinateBounds()`, a function that tests whether or not a coordinate is in a given bounds. ([#5053](https://github.com/mapbox/mapbox-gl-native/pull/5053))
+- An MGLAnnotationView can be repositioned by changing its `centerOffset` property. ([#5059](https://github.com/mapbox/mapbox-gl-native/issues/5059))
 
 ## 3.2.2
 
--- a/platform/ios/CHANGELOG.md
+++ b/platform/ios/CHANGELOG.md
@@ -26,7 +26,7 @@ Mapbox welcomes participation and contributions from everyone.  Please read [CON
 - Removed unused SVG files from the SDK’s resource bundle. ([#4641](https://github.com/mapbox/mapbox-gl-native/pull/4641))
 - Deprecated `-[MGLMapView emptyMemoryCache]`. ([#4725](https://github.com/mapbox/mapbox-gl-native/pull/4725))
 - Added `MGLCoordinateInCoordinateBounds()`, a function that tests whether or not a coordinate is in a given bounds. ([#5053](https://github.com/mapbox/mapbox-gl-native/pull/5053))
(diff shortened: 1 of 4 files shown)


platform/ios/CHANGELOG.md
```diff
@@ -26,6 +26,7 @@ Mapbox welcomes participation and contributions from everyone.  Please read [CON
 - Removed unused SVG files from the SDK’s resource bundle. ([#4641](https://github.com/mapbox/mapbox-gl-native/pull/4641))
 - Deprecated `-[MGLMapView emptyMemoryCache]`. ([#4725](https://github.com/mapbox/mapbox-gl-native/pull/4725))
 - Added `MGLCoordinateInCoordinateBounds()`, a function that tests whether or not a coordinate is in a given bounds. ([#5053](https://github.com/mapbox/mapbox-gl-native/pull/5053))
+- An MGLAnnotationView can be repositioned by changing its `centerOffset` property. ([#5059](https://github.com/mapbox/mapbox-gl-native/issues/5059))
```
1ec5: Let's mention that this offset is relative to the `coordinate` property, so people don't think you set this property to completely move the annotation. (You'd set `MGLAnnotation.coordinate` in that case.)

platform/ios/src/MGLAnnotationView.h
```diff
@@ -26,6 +26,14 @@ NS_ASSUME_NONNULL_BEGIN
  */
 @property (nonatomic, readonly, nullable) NSString *reuseIdentifier;
 
+/**
+ Annotation view is centered at the coordinate point of the associated annotation.
+ 
+ By changing this property you can reposition the view as needed. The offset is measured in points.
+ Positive offset moves the annotation view towards the bottom right, while negative offset moves it towards the top left.
+ */
+@property (nonatomic) CGPoint centerOffset;
```
1ec5: CGPoint is fine, but [CGVector](https://developer.apple.com/library/ios/documentation/GraphicsImaging/Reference/CGGeometry/index.html#//apple_ref/c/tdef/CGVector) would be better since we only support iOS 7 and above.

platform/ios/src/MGLAnnotationView.m
```diff
@@ -27,6 +27,13 @@ - (void)prepareForReuse
     // Intentionally left blank. The default implementation of this method does nothing.
 }
 
+- (void)setCenter:(CGPoint)center
+{
+    center.x += _centerOffset.x;
+    center.y += _centerOffset.y;
```
1ec5: Setting `centerOffset` after adding it to the map view should immediately move the view rather than waiting until the next call to `-setCenter:`. A custom `-setCenterOffset:` would do the trick.

platform/ios/src/MGLAnnotationView.h
```diff
@@ -26,6 +26,14 @@ NS_ASSUME_NONNULL_BEGIN
  */
 @property (nonatomic, readonly, nullable) NSString *reuseIdentifier;
 
+/**
+ Annotation view is centered at the coordinate point of the associated annotation.
+ 
+ By changing this property you can reposition the view as needed. The offset is measured in points.
+ Positive offset moves the annotation view towards the bottom right, while negative offset moves it towards the top left.
```
1ec5: Thanks for documenting these details. 👍

PR #5064, opened 2016-05-18
std::forward instead of std::move in invokeWithCallback

mikemorris:
Change use of `std::move` to `std::forward` in `RunLoop` per @daniel-j-h's suggestion. However, I'm now hitting an odd segfault in `npm run test-suite`...
```
* passed line-width default
* passed line-width function
* passed line-width literal
* passed line-width very-overscaled
* passed linear-filter-opacity-edge literal
May 17 20:13:54  node[80997] <Error>: ImageIO: PNG Not a PNG file
May 17 20:13:54  node[80997] <Error>: ImageIO: PNG Not a PNG file
Process 80997 stopped
* thread #1: tid = 0xf0df7, 0x00007fff9dab7be8 libsystem_malloc.dylib`tiny_free_list_add_ptr + 228, queue = 'com.apple.main-thread', stop reason = EXC_BAD_ACC
ESS (code=1, address=0x100)
    frame #0: 0x00007fff9dab7be8 libsystem_malloc.dylib`tiny_free_list_add_ptr + 228
libsystem_malloc.dylib`tiny_free_list_add_ptr:
->  0x7fff9dab7be8 <+228>: movq   %rax, (%r9)
    0x7fff9dab7beb <+231>: jmp    0x7fff9dab7c0b            ; <+263>
    0x7fff9dab7bed <+233>: movl   $0x1, %eax
    0x7fff9dab7bf2 <+238>: movb   %r8b, %cl
(lldb) bt
* thread #1: tid = 0xf0df7, 0x00007fff9dab7be8 libsystem_malloc.dylib`tiny_free_list_add_ptr + 228, queue = 'com.apple.main-thread', stop reason = EXC_BAD_ACCESS (code=1, address=0x100)
  * frame #0: 0x00007fff9dab7be8 libsystem_malloc.dylib`tiny_free_list_add_ptr + 228
    frame #1: 0x00007fff9dab5c5f libsystem_malloc.dylib`tiny_malloc_from_free_list + 1443
    frame #2: 0x00007fff9dab4705 libsystem_malloc.dylib`szone_malloc_should_clear + 292
    frame #3: 0x00007fff9dab45a1 libsystem_malloc.dylib`malloc_zone_malloc + 71
    frame #4: 0x00007fff9dab30cc libsystem_malloc.dylib`malloc + 42
    frame #5: 0x00007fff9b0db33e libc++abi.dylib`operator new(unsigned long) + 30
    frame #6: 0x000000010034f1db node`std::_Rb_tree<void*, std::pair<void* const, unsigned long>, std::_Select1st<std::pair<void* const, unsigned long> >, std::less<void*>, std::allocator<std::pair<void* const, unsigned long> > >::_M_insert_unique(std::_Rb_tree_iterator<std::pair<void* const, unsigned long> >, std::pair<void* const, unsigned long> const&) + 395
    frame #7: 0x00000001003294e8 node`v8::internal::Heap::RegisterNewArrayBuffer(bool, void*, unsigned long) + 280
    frame #8: 0x0000000100559ce7 node`v8::internal::Runtime::SetupArrayBuffer(v8::internal::Isolate*, v8::internal::Handle<v8::internal::JSArrayBuffer>, bool, void*, unsigned long, v8::internal::SharedFlag) + 295
    frame #9: 0x000000010055a10a node`v8::internal::Runtime_ArrayBufferInitialize(int, v8::internal::Object**, v8::internal::Isolate*) + 522
    frame #10: 0x000032832f30963b
    frame #11: 0x000032832f67f76c
    frame #12: 0x000032832f331147
    frame #13: 0x000032832f55cc2c
    frame #14: 0x000032832f94e8f3
    frame #15: 0x000032832f309ff7
    frame #16: 0x000032832f331143
    frame #17: 0x000032832f348c83
    frame #18: 0x000032832f5e761e
    frame #19: 0x000032832f309ff7
    frame #20: 0x000032832f331143
    frame #21: 0x000032832f5a2e87
    frame #22: 0x000032832f3318fd
    frame #23: 0x000032832f315b62
    frame #24: 0x00000001002dac88 node`v8::internal::Invoke(bool, v8::internal::Handle<v8::internal::JSFunction>, v8::internal::Handle<v8::internal::Object>, int, v8::internal::Handle<v8::internal::Object>*) + 728
    frame #25: 0x0000000100166c84 node`v8::Function::Call(v8::Local<v8::Context>, v8::Local<v8::Value>, int, v8::Local<v8::Value>*) + 276
    frame #26: 0x00000001006557a9 node`node::AsyncWrap::MakeCallback(v8::Local<v8::Function>, int, v8::Local<v8::Value>*) + 547
    frame #27: 0x000000010068a59b node`node::ZCtx::After(uv_work_s*, int) + 251
    frame #28: 0x0000000100789f99 node`uv__work_done + 178
    frame #29: 0x000000010078bba5 node`uv__async_event + 198
    frame #30: 0x000000010078bd38 node`uv__async_io + 136
    frame #31: 0x000000010079aa2a node`uv__io_poll + 1554
    frame #32: 0x000000010078c1eb node`uv_run + 321
    frame #33: 0x0000000100668124 node`node::Start(int, char**) + 583
    frame #34: 0x0000000100000c34 node`start + 52
```

It looks like this is code you wrote a year ago @jfirebaugh, any idea what could be going on here?

### DIFF
--- a/include/mbgl/util/run_loop.hpp
+++ b/include/mbgl/util/run_loop.hpp
@@ -48,9 +48,9 @@ public:
     // Invoke fn(args...) on this RunLoop.
     template <class Fn, class... Args>
     void invoke(Fn&& fn, Args&&... args) {
-        auto tuple = std::make_tuple(std::move(args)...);
+        auto tuple = std::make_tuple(std::forward<Args>(args)...);
         auto task = std::make_shared<Invoker<Fn, decltype(tuple)>>(
-            std::move(fn),
+            std::forward<Fn>(fn),
             std::move(tuple));
 
         push(task);
@@ -63,9 +63,9 @@ public:
         auto flag = std::make_shared<std::atomic<bool>>();
         *flag = false;
 
-        auto tuple = std::make_tuple(std::move(args)...);
+        auto tuple = std::make_tuple(std::forward<Args>(args)...);
         auto task = std::make_shared<Invoker<Fn, decltype(tuple)>>(
-            std::move(fn),
+            std::forward<Fn>(fn),
             std::move(tuple),
             flag);
 
@@ -86,19 +86,19 @@ public:
         // because if the request was cancelled, then R might have been destroyed. L2 needs to check
         // the flag because the request may have been cancelled after L2 was invoked but before it
         // began executing.
-        auto after = [flag, current = RunLoop::Get(), callback1 = std::move(callback)] (auto&&... results1) {
+        auto after = [flag, current = RunLoop::Get(), callback1 = std::forward<Cb>(callback)] (auto&&... results1) mutable {
             if (!*flag) {
-                current->invoke([flag, callback2 = std::move(callback1)] (auto&&... results2) {
+                current->invoke([flag, callback2 = std::forward<Cb>(callback1)] (auto&&... results2) {
                     if (!*flag) {
-                        callback2(std::move(results2)...);
+                        callback2(std::forward<decltype(results2)>(results2)...);
                     }
-                }, std::move(results1)...);
+                }, std::forward<decltype(results1)>(results1)...);
             }
         };
 
-        auto tuple = std::make_tuple(std::move(args)..., after);
+        std::tuple<Args..., decltype(after)> tuple(std::forward<Args>(args)..., std::move(after));
         auto task = std::make_shared<Invoker<Fn, decltype(tuple)>>(
-            std::move(fn),
+            std::forward<Fn>(fn),
             std::move(tuple),
             flag);
 


jfirebaugh:
I think `std::move` is correct as-is in all these places -- `invoke` and friends take ownership of the input, in order to execute the given function with the given arguments on a separate thread.

daniel-j-h:
> I think std::move is correct as-is in all these places -- invoke and friends take ownership of the input, in order to execute the given function with the given arguments on a separate thread.

Small self-contained example why calling `std::move` on forwarding references is a bad idea:

```c++
#include <cassert>
#include <utility>

struct S {

  // Default Constructible
  S() = default;

  // Copyable
  S(S&) = default;
  S& operator=(S&) = default;

  // Moveable
  S(S&& o) noexcept {
    o.dead = true;
  }

  S& operator=(S&& o) noexcept {
    o.dead = true;
    return *this;
  }

  // Assert if we are a moved-from object
  void fn() {
    assert(!dead);
  }

  bool dead = false;
};


template <typename T>
void fwd_wrong(T&& t) {
  auto tmp = std::move(t);
  (void)tmp;
}

template <typename T>
void fwd_right(T&& t) {
  auto tmp = std::forward<T>(t);
  (void)tmp;
}

int main() {
  S x;

  // Although we pass a lvalue we always call std::move, invalidating the callsite although it should have been a copy
  fwd_wrong(x);
  x.fn(); // triggers, x is dead now

  // Perfectly forwards, we pass x as lvalue, therefore a copy is made
  //fwd_right(x);
  //x.fn();
}
```

`F&&` in a type-deducing context is a forwarding reference, you have to `std::forward` it along.

In cases the `F&&` really is a rvalue reference (which you don't know at this point) `std::forward` will lower to `std::move` anyway doing the Right Thing (tm).

jfirebaugh:
The intent is that `RunLoop::invoke*` is a sink, accepting only rvalue references. So I would rather go with something like https://github.com/mapbox/mapbox-gl-native/commit/bf3a963d4ddf602ace6d3c568ccd7307dd68a202.

daniel-j-h:
Yep either restrict the forwarding reference then via `enable_if` or: in the function still `std::forward` (since this is how to perfectly forward, `std::move` on a forwarding reference is technically wrong) and provide the user with a clear error message instead of obscure errors from SFINAE.

    static_assert(std::is_rvalue_reference<T&&>(), "Takes Ownership");

jfirebaugh:
Going with @daniel-j-h's recommendation in #6262. 579c31c546684c05bcbee806f0450a329f8e6b7d is similar to the change here but passes tests. I think the key difference is sticking with `std::make_tuple` so that the resulting tuple does not have any reference members, as `std::tuple<Args..., decltype(after)>` might.

PR #5067, opened 2016-05-18
[ios] Allow annotation views to be flattened

frederoni:


1ec5:
Filed #5089 and #5090 to improve the tilting effect and the interaction with callout views.

1ec5:
#5546 backs out this feature due to #5090. We’ll revisit for v3.4.0.

Issue #1572, opened 2015-05-16
MGLMapView fails to bind EAGLDrawable when initialised with CGRect with no width and height


duemunk:
Initialising a `MGLMapView` with CGRectZero:

``` swift
let map = MGLMapView(frame: CGRectZero, accessToken: "_")
```

Crashes:

``` console
2015-05-16 16:56:35.782 SomeApp[35232:1946290] Failed to bind EAGLDrawable: <CAEAGLLayer: 0x17403a940> to GL_RENDERBUFFER 1
2015-05-16 16:56:35.783 SomeApp[35232:1946290] Failed to make complete framebuffer object 8cd6
2015-05-16 16:56:35.800 SomeApp[35232:1946361] [ERROR] {0}{Map}[Shader]: Program failed to validate: Validation Failed: Current draw framebuffer is invalid.
```

Current workaround

``` swift
let map = MGLMapView(frame: CGRect(x: 0, y: 0, width: 1, height: 1), accessToken: "_")
```

1ec5:
Via #1654, `MGLMapView` should also support `CGRectNull`.

1ec5:
`-[MGLMapView wakeGL:]` should guard on `CGRectIsEmpty()`, and probably `CGRectIsInfinite()` while we’re at it. `-setFrame:` and `-setBounds:` should trigger this check and cause binding to occur if the frame changes from being empty (or infinite) to a reasonable value.

1ec5:
Per #4065, initializing the map view with an empty frame no longer crashes. It looks like the validation step that was failing was removed in #2147.

1ec5:
Per #5457, the crash has returned.

1ec5:
We’re still unable to reproduce the crash, yet we do get reports about it in the wild. The fix would likely involve changes to mbgl, which we aren’t willing to take this close to a release. Punting to 3.4.0.

DsrMedia:
[...]

PR #5074, opened 2016-05-19
[core] [Qt] Avoid NaNs in TransformState unit conversions

brunoabinader:
Fix cases where e.g. state has either zero width or height, causing the unit conversion functions to return NaNs.

:eyes: @kkaefer @jfirebaugh @ansis

### DIFF
--- a/src/mbgl/map/transform.cpp
+++ b/src/mbgl/map/transform.cpp
@@ -45,24 +45,23 @@ Transform::Transform(std::function<void(MapChange)> callback_,
 #pragma mark - Map View
 
 bool Transform::resize(const std::array<uint16_t, 2> size) {
-    if (state.width != size[0] || state.height != size[1]) {
-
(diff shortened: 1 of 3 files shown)


src/mbgl/map/transform_state.hpp
```diff
@@ -84,7 +84,7 @@ class TransformState {
     NorthOrientation orientation = NorthOrientation::Upwards;
 
     // logical dimensions
-    uint16_t width = 0, height = 0;
+    uint16_t width = 1, height = 1;
```
kkaefer: Why `1`?

brunoabinader: Any unsigned number != 0. Either this or we could go the other way and use `std::numeric_limits::max`.

1ec5: I think this fixes #1572.

PR #5082, opened 2016-05-20
[android] #5081 - added rotation to View Marker

### LINKED ISSUE #5081: Add rotation to MarkerView
Google exposes customization of **rotation**:

> The orientation of the marker, specified in degrees clockwise. The default position changes if the marker is flat. The default position for a flat marker is north aligned. When the marker is not flat, the default position is pointing up and the rotation is such that the marker is always facing the camera.

Now that #3276 landed we can implement this in MarkerView/MarkerViewOptions.

tobrun:
closes #5081


tobrun:

Issue #5040, opened 2016-05-14
When map is tilted, annotation views don’t scale by distance from camera


1ec5:
When the map is tilted, an GL icon or point annotation grows as the camera moves toward it and shrinks as the camera moves away from it (#2186). However, annotation views maintain their size. This behavior is consistent with the user dot, but scaling would help by reinforcing the illusion of a tilted map.

/cc @boundsj @friedbunny

friedbunny:
This will be a good option to have. MapKit maintains constant size, while Google scales by distance — neither gives you a choice.

1ec5:
I think the default should be to scale, for consistency with GL annotations.

boundsj:
@1ec5 @friedbunny while https://github.com/mapbox/mapbox-gl-native/pull/5085 offers a solution it is not one that can be shared with @tobrun. It also defaults to not scale by default -- for consistency with other SDKs and also because [there is an ongoing discussion and it seems like things may eventually change in GL annotation land](https://github.com/mapbox/mapbox-gl-native/issues/2186).

Example of 1000 scaled, flattened annotation views:

![scaled_flat_annotations](https://cloud.githubusercontent.com/assets/63186/15440991/374ecc4c-1e8d-11e6-9422-c716a750a450.gif)

1ec5:
> It also defaults to not scale by default -- for consistency with other SDKs and also because [there is an ongoing discussion and it seems like things may eventually change in GL annotation land](https://github.com/mapbox/mapbox-gl-native/issues/2186).

I’ve come around to the idea that unscaled annotations are more useful to more use cases than scaled annotations, despite scaling making for better demos. However, point annotations are still rendered by GL by default, and GL-rendered annotations are unconditionally scaled. I think it would be a jarring experience for you to adopt annotation views and suddenly see behaviors like scaling change.

friedbunny:
I don’t think we should constrain view-based annotations based on the current (in)capabilities of GL annotations. The difference may be potentially jarring, but unscaled is the default we’re converging on for both, even if we don’t know when GL will support it.

I wonder how common mix-and-match views-and-GL annotations will be. If a developer uses both and thinks the scale difference is unsightly, they’re likely to seek out the `MGLViewAnnotation.scaled` option. I’m not sure that the opposite will be true — I’d be afraid developers won’t know the unscaled option exists.

1ec5:
[...]

PR #5085, opened 2016-05-20
[ios] Add annotation view scaling

boundsj:
Fixes https://github.com/mapbox/mapbox-gl-native/issues/5040

cc @1ec5 @friedbunny


platform/ios/src/MGLAnnotationView.h
```diff
@@ -40,6 +40,11 @@ NS_ASSUME_NONNULL_BEGIN
  */
 @property (nonatomic, assign, getter=isFlat) BOOL flat;
 
+/**
+ Setting this property to YES will cause the annotation to shrink as it approaches the horizon and grow as it moves away from the
+ horizon when the associated map view is tilted.
+ */
+@property (nonatomic, assign, getter=isScaled) BOOL scaled;
```
1ec5: “Scaling” happens for a lot of reasons on iOS. This property should be more descriptively named so that people don’t expect it to relate to the device’s pixel ratio or `UIImage.scale` somehow. Maybe `scalesByViewingDistance` or something?

platform/ios/src/MGLMapView.mm
```diff
@@ -4383,7 +4383,7 @@ - (void)updateAnnotationViews
                 }
                 
                 CGPoint center = [self convertCoordinate:annotationContext.annotation.coordinate toPointToView:self];
-                [annotationView setCenter:center pitch:self.camera.pitch];
+                [annotationView setCenter:center pitch:self.camera.pitch maxPitch:MGLDegreesFromRadians(mbgl::util::PITCH_MAX)];
```
1ec5: `mbgl::util::PITCH_MAX` is constant, so instead of passing it into this method, use it directly in `-setCenter:pitch:`. You’ll need to rename MGLAnnotationView.m to have a .mm extension so you can use Objective-C++ namespaces.

platform/ios/src/MGLAnnotationView.m
```diff
@@ -56,6 +62,18 @@ - (void)updatePitch:(CGFloat)pitch
     self.layer.transform = t;
 }
 
+- (void)updateScaleForPitch:(CGFloat)pitch maxPitch:(CGFloat)maxPitch
+{
+    CGFloat superViewHeight = CGRectGetHeight(self.superview.frame);
+    if (superViewHeight > 0.0) {
+        CGFloat scaleReduction = 1.0 - self.center.y / superViewHeight;
+        CGFloat pitchIntensity = MAX(pitch, 0) / maxPitch;
+        CGFloat pitchAdjustedScale = 1.0 - scaleReduction * pitchIntensity;
```
1ec5: Can you add comments explaining the rationale behind this formula, or alternatively where this formula came from?

platform/ios/src/MGLAnnotationView.h
```diff
@@ -13,7 +13,7 @@ NS_ASSUME_NONNULL_BEGIN
  @param reuseIdentifier The string that identifies that this annotation view is reusable.
  @return The initialized annotation view object or `nil` if there was a problem initializing the object.
```
1ec5: The return type isn’t marked `nullable`. I realize the MKAnnotationView documentation also indicates that `nil` can be returned, but I think that’s a holdover from the old days before nullability was a concept in Objective-C. Is this view that much more likely to fail to initialize than an ordinary UIView?

1ec5:
👍

platform/ios/src/MGLAnnotationView.h
boundsj: I agree and thanks for helping me understand this a bit better. I'm going to remove the confusing language in the comment. Seems like MK and older MB code needs some additional comment cleaning (i.e. https://github.com/mapbox/mapbox-gl-native/blob/master/platform/ios/src/MGLAnnotationImage.h#L17)

PR #4359, opened 2016-05-21
[core] Wireframe debug mode

brunoabinader:
Together with `MapDebugOptions::Collision`, provides _wireframe_ rendering output so all drawn objects are visible. This mode ignores the line width, background and fill colors for better visualization.

/cc @mapbox/gl

### DIFF
--- a/include/mbgl/map/mode.hpp
+++ b/include/mbgl/map/mode.hpp
@@ -42,6 +42,7 @@ enum class MapDebugOptions : EnumType {
     ParseStatus = 1 << 2,
     Timestamps  = 1 << 3,
     Collision   = 1 << 4,
+    Wireframe   = 1 << 5,
 };
 
 inline MapDebugOptions operator| (const MapDebugOptions& lhs, const MapDebugOptions& rhs) {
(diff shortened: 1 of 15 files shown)


brunoabinader:
Example output:

wireframe

:eyes: @kkaefer @jfirebaugh @tmpsantos

1ec5:
Awesome – this is the mode @kkaefer requested in #4669, right? It’ll be pretty easy to hook this up to the iOS and OS X SDKs now that there’s a `MapDebugOptions` option for it.

brunoabinader:
@1ec5 I think this is what @kkaefer mentioned - https://github.com/mapbox/mapbox-gl-native/blob/master/platform/default/glfw_view.cpp#L483-L520 - which provides a different output - but can be added as a new `MapDebugOptions` as such.

1ec5:
22ef615f360bb5e0ea58491620f0fc59da8f491b exposes this new mode in the iOS and OS X SDKs and their demo applications.

brunoabinader:
Thank you @1ec5!

kkaefer:
@brunoabinader stencil mask mode is different from the wireframe mode; the stencil mode shows the checkerboard-style pattern that we're using to clip tiles.

brunoabinader:
@kkaefer yes, that's exactly what I mentioned to @1ec5 in the last comment :) We should move the stencil mask code to core and add a new enum for it in `MapDebugOptions`.

PR #5091, opened 2016-05-21
Cleanup iOS demo app

friedbunny:
- Remove unsupported custom style layer example.
- Only show debug log print/delete options if debug logging is enabled.

/cc @1ec5 @jfirebaugh


1ec5:
👍

friedbunny:
FF-merged and pushed this in master, but the ongoing GitHub outage means that it isn’t showing up yet.

jfirebaugh:
👍

PR #5098, opened 2016-05-23
Move Observers to their own files

kkaefer:
We should move the observers to their own header files and separate an observer from the actual object the observer template is for. This is to reduce include clutter and inadvertent dependencies.


kkaefer:
:eyes: @jfirebaugh

jfirebaugh:
👍

Issue #4562, opened 2016-03-31
port improved symbol fading


ansis:
port https://github.com/mapbox/mapbox-gl-js/pull/2376

PR #5105, opened 2016-05-23
[core] Shared GL context mode rendering glyph atlas instead of the actual map

tmpsantos:
screenshot from 2016-05-23 16-59-08

Regression caused by https://github.com/mapbox/mapbox-gl-native/issues/4562

### DIFF
--- a/src/mbgl/renderer/painter_symbol.cpp
+++ b/src/mbgl/renderer/painter_symbol.cpp
@@ -289,4 +289,5 @@ void Painter::renderSymbol(SymbolBucket& bucket,
 
     }
 
+    config.activeTexture = GL_TEXTURE0;
 }


tmpsantos:
:eyes: @ansis

tmpsantos:
Bit more of context here (no pun intended), the Share Context Mode is when Mapbox GL Native is drawing in the same thread/render pass as the application embedding the map, like this example using Qt.

kkaefer:
Looks good!

brunoabinader:
Tested on Qt QML app in OS X 👍

tmpsantos:
Merged.

jfirebaugh:
This doesn't look like the correct fix. As with all GL state, the active texture should be set to the necessary value wherever it's needed, not reset to a default value in an ad hoc manner.

tmpsantos:
> This doesn't look like the correct fix. As with all GL state, the active texture should be set to the necessary value wherever it's needed, not reset to a default value in an ad hoc manner.

We do that.

IMO the problem is that we were leaking the active texture set to 1. This could be fixed on the client side by setting the active texture unit back to 0 as in:

```
    mapObj->render();
    glActiveTexture(GL_TEXTURE0);
```

jfirebaugh:
If we need to reset state for third party compatibility, let's do it in `gl::Config::setDirty`.

PR #5092, opened 2016-05-23
Updating Android NDK To 11c

kkaefer:
Google released [Android NDK 11c in March 2016](https://developer.android.com/ndk/downloads/index.html).  The current version in use by the Mapbox Android SDK is [NDK 10e which is from May 2015](https://developer.android.com/ndk/downloads/revision_history.html).

The steps for making this happen are:

1. Add Android NDK 11c to Mason
1. Update Mapbox Android SDK GYP build scripts to use it

/cc @mapbox/mobile @jfirebaugh @kkaefer @tmpsantos @brunoabinader


bleege:
Thanks @kkaefer @tmpsantos !

bleege:
We ended up having to revert this in 6abd3d70beb9233262fbbb2de1f780cdfc29bb5c as there were runtime issues on Android.  Will create a new ticket to do the upgrade that we can work on after `4.1.0 FINAL` ships.

PR #4669, opened 2016-05-23
Stencil debugging view is gone

brunoabinader:
The GLFW app had a debug view that visualized the stencil buffer rather than the color buffer. The OS X app no longer has that.

### DIFF
--- a/include/mbgl/gl/gl_values.hpp
+++ b/include/mbgl/gl/gl_values.hpp
@@ -269,6 +269,10 @@ struct PixelZoom {
     }
 };
 
+inline bool operator!=(const PixelZoom::Type& a, const PixelZoom::Type& b) {
+    return a.xfactor != b.xfactor || a.yfactor != b.yfactor;
+}
+
 struct RasterPos {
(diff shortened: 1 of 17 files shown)


brunoabinader:
:eyes: @1ec5 @kkaefer

1ec5:
Added to the iOS and OS X SDKs in e83924f0eca538f63f2a5a1cdad812ee96824937. However, the iOS SDK fails to compile because `GL_STENCIL_INDEX`, `PixelZoom`, and `RasterPos` are all undefined. [This StackOverflow answer](http://stackoverflow.com/a/8150719/4585461) suggests a workaround for [the lack of `GL_STENCIL_INDEX` on iOS](https://developer.apple.com/library/ios//documentation/OpenGLES/Conceptual/OpenGLESHardwarePlatformGuide_iOS/OpenGLESPlatforms/OpenGLESPlatforms.html#//apple_ref/doc/uid/TP40012935-CH106-SW4).

brunoabinader:
Support for OpenGL ES 2.0 tracked here: #5117.

Issue #1729, opened 2015-06-16
add support for multipolygons and polygon interior holes


incanus:
Punted from #1655. Will need to be exposed in `MGLPolygon` as well on iOS.

jfirebaugh:
[GeoJSON sources](https://github.com/mapbox/mapbox-gl-native/pull/2161) support the full range of GeoJSON geometries. However if we want to continue expanding the annotations API, we could support this there too.

Zakay:
+1 really need this!

TYRONEMICHAEL:
+1

maciekish:
+1

dariav:
+1

MaximAlien:
+1

aemchuk:
+1 Will be very useful

caHarkness:
We have switched from Google Maps to Mapbox for the rendering speed of polygons, but this feature is essential. We absolutely need holes!

nealsanche:
+1 I am currently working with a mapping app with very complex polygons with holes that Google Maps API is stumbling on quite profoundly. I'd dearly love to replace it with MapBox and then found out that holes are not currently supported. If I have to, I'd work to contribute a PR to the project so that this could be supported. Would you please contact me, so that I might be about to assess the work that might be involved in making that happen?

1ec5:
#5110 adds support for holes in polygons to the iOS and OS X SDKs. Note that, unlike in MapKit, a MGLPolygon cannot be nested more than one level deep; a concentric polygon, as it were, would be represented by an MGLMultiPolygon, which currently cannot be added to the map as a single annotation. True multipolygon support will have to wait for #5158.

nealsanche:
Last night I was able to render a polygon with holes on Android. Added the
API for adding holes to Java. Processed it in the JNI layer, and added
support in the rendering for the new data.

On Fri, May 27, 2016, 1:35 AM Minh Nguyễn <notifications@github.com> wrote:

> #5110 <https://github.com/mapbox/mapbox-gl-native/pull/5110> adds support
> for holes in polygons to the iOS and OS X SDKs. Note that, unlike in
> MapKit, a MGLPolygon cannot be nested more than one level deep; a
> concentric polygon, as it were, would be represented by an MGLMultiPolygon,
> which currently cannot be added to the map as a single annotation. True
> multipolygon support will have to wait for #5158
> <https://github.com/mapbox/mapbox-gl-native/issues/5158>.
>
> —
[...]

Issue #3630, opened 2016-01-20
Provide a GeoJSON serialization and deserialization utility


tobrun:
This is something that has already been discussed multiple times but was not ticketed out yet. 
Our SDK should expose an API so our users can work with GeoJSON.

Since those discussions, GeoJSON support has landed natively in https://github.com/mapbox/mapbox-gl-native/pull/2161. 
I'm not really sure how that translates to this requirement:
- should we use that API and expose it in Android/iOS SDK?
- should we implement parsers/builders etc in the Android/iOS SDK ourselves?
  - We used android-geojson library in Android Raster SDK. 
  - We could consider using a library like mattt/GeoJSONSerialization for iOS

1ec5:
> We could consider using a library like mattt/GeoJSONSerialization for iOS

GeoJSONSerialization converts GeoJSON into MapKit-specific types, which are incompatible with the our iOS SDK. We could derive a Mapbox-compatible library from it, however.

friedbunny:
Serializing JSON can be a definite pain (ugh, Swift), but there are a variety of third-party libraries out there for the folks who need to do it... so I'm not sure that we need to provide this ourselves for every user.

1ec5:
I can see the value-add for something like GeoJSONSerialization, though. Ordinary JSON comes with no semantics, so using a generic JSON library conjures up memories of raw XML parsing and validation using DTDs and XML Schemas. :grin: By contrast, a GeoJSON-specific library can convert to actual Mapbox SDK types.

For iOS, I think it makes more sense right now to have an independent GeoJSON-to-MGL* library rather than folding the APIs directly into the Mapbox iOS SDK. A developer using React Native Mapbox GL would want to use a more idiomatic React Native component anyways.

1ec5:
Note that basic GeoJSON support has landed in the native SDKs, in the form of GeoJSON sources and more-or-less GeoJSON-compatible geometry classes. This ticket remains open to track ideas for streamlining GeoJSON-related workflows. For example, a built-in GeoJSON (de)serialization feature (think MGLGeoJSONSerialization on iOS and macOS) is now eminently feasible.

tobrun:
obsolete, closing

1ec5:
For iOS and macOS, an MGLGeoJSONSerialization utility would still be desirable. Reopening and removing the Android label.

1ec5:
[...]

PR #5110, opened 2016-05-23
Feature querying; more geometry types

1ec5:
Added methods that return the visible features at a given point or in a given rectangle on the map. Each feature is represented by a class conforming to the MGLFeature protocol and, conveniently enough, the MGLAnnotation protocol. Yup, that means you can feed the features right back to the map in the form of point and shape annotations. These classes derive from the usual suspects (MGLPointAnnotation, MGLPolyline, etc.), but there are several new classes to represent geometry types supported by geometry.hpp but not by MapKit (MGLFeatureCollection, MGLMultiPolyline, etc.). Complex geometry types like multipolygon cannot yet be added as overlays, but this PR does add support for holes in simple polygons.

In addition to the usual annotation properties, each feature object has an identifier and a dictionary of attributes (properties) that come from the source. The dictionary contains boxed Foundation types such as NSString and NSNumber, as appropriate. Tags are also represented as NSNumbers: mbgl allows features to have IDs within the entire `uint64_t` space ~~and the Mapbox GL Vector Tile Spec doesn’t specify the ID data type~~, so there’s no room for magic numbers.

In iosapp and osxapp, dropping a pin (by long-pressing or via the context menu) shows the usual callout view, but now its title is the name of a feature under the cursor, if available, rather than “Dropped Pin”. Choose “Select Features” from the context menu to highlight all the features under the cursor.

scpa
downtown
jungle-jims

More to come:

* [x] Convert `mbgl::geometry::value`s to Cocoa types
* [x] Specialize SDK shape classes for feature querying results
* [x] Write SDK types for multipolylines, multipolygons, feature collections
* [x] Convert `mbgl::Feature`s to `MGLFeature`s
* [x] Query rendered features at an `NSPoint` on OS X
* [x] Support all feature shape classes in the annotation layer
* [x] Move SDK feature types into their own files
* [x] Make all SDK feature types inherit from MGLShape
* [x] Move feature Cocoa-ization into factory method on MGLShape in darwin/
* [x] Add unit tests for feature Cocoa-ization
* [x] Query rendered features at a `CGPoint` on iOS
* [x] Add a way to filter by layer
* [x] Support interior rings (holes) in polygons (fixes the iOS/OS X side of #1729)
* [ ] ~~Migrate `mbgl::AnnotationManager` to `mbgl::geometry` types?~~ (split out into #5158)
* [x] Fix crash querying features in the presence of `mbgl::AnnotationManager`-managed features
* [x] Query rendered features within `NSRect`/`CGRect`

Fixes the iOS/OS X side of #352.


jfirebaugh:
> the Mapbox GL Vector Tile Spec doesn’t specify the ID data type

It [specifies them as `uint64`](https://github.com/mapbox/vector-tile-spec/blob/969d1ec77e0c9d83cc4d2aee791238bc21e6c132/2.1/vector_tile.proto#L34). However, I might switch the mbgl core type to `mbgl::Value`, to accommodate all possible [GeoJSON `"id"` values](https://tools.ietf.org/html/draft-ietf-geojson-03#section-3.2).

> Dropping a pin via the menu or context menu item shows the usual dropped pin, but the pin’s title is now the name of a feature under the cursor, if available

Nice!

jfirebaugh:
> * Support line rings and interior holes? (#1729)
> * Migrate `mbgl::AnnotationManager` to `mbgl::geometry` types?

👍  These two would go well together.

1ec5:
>> the Mapbox GL Vector Tile Spec doesn’t specify the ID data type

> It [specifies them as `uint64`](https://github.com/mapbox/vector-tile-spec/blob/969d1ec77e0c9d83cc4d2aee791238bc21e6c132/2.1/vector_tile.proto#L34). However, I might switch the mbgl core type to `mbgl::Value`, to accommodate all possible [GeoJSON `"id"` values](https://tools.ietf.org/html/draft-ietf-geojson-03#section-3.2).

Ah, I mistakenly thought the prose readme was the canonical specification. 👍 With `mbgl::Value` as the underlying `id` type, I think the most appropriate type for `MGLFeature.tag` would be `id`, confusingly enough!

1ec5:
I need to investigate a crash querying features in a tile that also has annotations. It has to do with the indexed feature having no name but the annotation layers having names like `com.mapbox.annotations.points`:

```
#0	0x0000000100170774 in mbgl::AnnotationTile::getLayer(std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char> > const&) const at /path/to/mapbox-gl-native-2/src/mbgl/annotation/annotation_tile.cpp:29
#1	0x000000010018f2c6 in mbgl::FeatureIndex::addFeature(std::__1::unordered_map<std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char> >, std::__1::vector<mbgl::Feature, std::__1::allocator<mbgl::Feature> >, std::__1::hash<std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char> > >, std::__1::equal_to<std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char> > >, std::__1::allocator<std::__1::pair<std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char> > const, std::__1::vector<mbgl::Feature, std::__1::allocator<mbgl::Feature> > > > >&, mbgl::IndexedSubfeature const&, mbgl::GeometryCollection const&, std::experimental::fundamentals_v1::optional<std::__1::vector<std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char> >, std::__1::allocator<std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char> > > > > const&, mbgl::GeometryTile const&, mbgl::CanonicalTileID const&, mbgl::Style const&, float, float) const at /path/to/mapbox-gl-native-2/src/mbgl/geometry/feature_index.cpp:105
#2	0x000000010018f0b2 in mbgl::FeatureIndex::query(std::__1::unordered_map<std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char> >, std::__1::vector<mbgl::Feature, std::__1::allocator<mbgl::Feature> >, std::__1::hash<std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char> > >, std::__1::equal_to<std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char> > >, std::__1::allocator<std::__1::pair<std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char> > const, std::__1::vector<mbgl::Feature, std::__1::allocator<mbgl::Feature> > > > >&, mbgl::GeometryCollection const&, float, double, double, std::experimental::fundamentals_v1::optional<std::__1::vector<std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char> >, std::__1::allocator<std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char> > > > > const&, mbgl::GeometryTile const&, mbgl::CanonicalTileID const&, mbgl::Style const&) const at /path/to/mapbox-gl-native-2/src/mbgl/geometry/feature_index.cpp:85
#3	0x000000010044ce0d in mbgl::VectorTileData::queryRenderedFeatures(std::__1::unordered_map<std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char> >, std::__1::vector<mbgl::Feature, std::__1::allocator<mbgl::Feature> >, std::__1::hash<std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char> > >, std::__1::equal_to<std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char> > >, std::__1::allocator<std::__1::pair<std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char> > const, std::__1::vector<mbgl::Feature, std::__1::allocator<mbgl::Feature> > > > >&, mbgl::GeometryCoordinates const&, mbgl::TransformState const&, std::experimental::fundamentals_v1::optional<std::__1::vector<std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char> >, std::__1::allocator<std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char> > > > > const&) at /path/to/mapbox-gl-native-2/src/mbgl/tile/vector_tile_data.cpp:202
#4	0x000000010030e9d7 in mbgl::Source::queryRenderedFeatures(mbgl::StyleQueryParameters const&) const at /path/to/mapbox-gl-native-2/src/mbgl/source/source.cpp:367
#5	0x000000010036bcea in mbgl::Style::queryRenderedFeatures(mbgl::StyleQueryParameters const&) const at /path/to/mapbox-gl-native-2/src/mbgl/style/style.cpp:324
#6	0x00000001002511d0 in mbgl::Map::queryRenderedFeatures(mapbox::geometry::point<double> const&, std::experimental::fundamentals_v1::optional<std::__1::vector<std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char> >, std::__1::allocator<std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char> > > > > const&) at /path/to/mapbox-gl-native-2/src/mbgl/map/map.cpp:739
#7	0x0000000100069e75 in -[MGLMapView visibleFeaturesAtPoint:inStyleLayersWithIdentifiers:] at /path/to/mapbox-gl-native-2/platform/osx/src/MGLMapView.mm:2280
#8	0x0000000100069a03 in -[MGLMapView visibleFeaturesAtPoint:] at /path/to/mapbox-gl-native-2/platform/osx/src/MGLMapView.mm:2263
#9	0x00000001000065a8 in -[MapDocument handlePressGesture:] at /path/to/mapbox-gl-native-2/platform/osx/app/MapDocument.m:411
```

1ec5:
Split out #5159 for the crash querying annotation tiles.

1ec5:
@jfirebaugh, would you mind reviewing the Objective-C++ changes in MGLFeature.mm, MGLFeatureTests.mm, and MGLMapView.mm, plus the C++ changes in shape_annotation_impl.cpp? Thanks!

@boundsj, would you mind reviewing the rest of the geometry-related changes in platform/darwin/? Thanks!

platform/darwin/src/MGLFeature.mm
```diff
@@ -0,0 +1,299 @@
...
+}
+
+@end
+
+/**
+ Recursively transforms a C++ type into the corresponding Foundation type.
+ */
+class PropertyValueEvaluator {
+public:
+    PropertyValueEvaluator() {}
```
jfirebaugh: This default constructor can be removed; the compiler will generate it automatically.

platform/darwin/src/MGLFeature.mm
```diff
@@ -0,0 +1,299 @@
...
+    }
+    
+    id operator()(const std::string &value) const {
+        return @(value.c_str());
+    }
+    
+    id operator()(const std::vector<mbgl::Value> &values) const {
+        std::vector<id> objects;
+        objects.reserve(values.size());
+        std::transform(values.begin(), values.end(), std::back_inserter(objects), ^id (const mbgl::Value &value) {
```
jfirebaugh:
```
for (const auto& v : values) {
    objects.push_back(mbgl::Value::visit(v, *this));
}
```

platform/darwin/src/MGLFeature.mm
```diff
@@ -0,0 +1,299 @@
...
+    
+    id operator()(const std::unordered_map<std::string, mbgl::Value> &items) const {
+        std::vector<NSString *> keys;
+        keys.reserve(items.size());
+        std::vector<id> objects;
+        objects.reserve(items.size());
+        for (auto &item : items) {
+            keys.push_back(@(item.first.c_str()));
+            PropertyValueEvaluator evaluator;
+            objects.push_back(mbgl::Value::visit(item.second, evaluator));
```
jfirebaugh: Reuse `*this` here too.

platform/darwin/src/MGLFeature.mm
```diff
@@ -0,0 +1,299 @@
...
+    }
+    
+    id operator()(const std::vector<mbgl::Value> &values) const {
+        std::vector<id> objects;
+        objects.reserve(values.size());
+        std::transform(values.begin(), values.end(), std::back_inserter(objects), ^id (const mbgl::Value &value) {
+            PropertyValueEvaluator evaluator;
+            return mbgl::Value::visit(value, evaluator);
+        });
+        return [NSArray arrayWithObjects:&objects[0] count:objects.size()];
```
jfirebaugh: Is there no way to build up the `NSArray` directly, rather than building a temporary `std::vector` and then copying it in?

platform/darwin/src/MGLFeature.mm
```diff
@@ -0,0 +1,299 @@
...
+        std::vector<NSString *> keys;
+        keys.reserve(items.size());
+        std::vector<id> objects;
+        objects.reserve(items.size());
+        for (auto &item : items) {
+            keys.push_back(@(item.first.c_str()));
+            PropertyValueEvaluator evaluator;
+            objects.push_back(mbgl::Value::visit(item.second, evaluator));
+        }
+        return [NSDictionary dictionaryWithObjects:&objects[0] forKeys:&keys[0] count:keys.size()];
```
jfirebaugh: Same question for `NSDictionary`.

platform/darwin/src/MGLFeature.mm
```diff
@@ -0,0 +1,299 @@
...
+class GeometryEvaluator {
+public:
+    MGLShape <MGLFeaturePrivate> * operator()(const mapbox::geometry::point<T> &geometry) const {
+        MGLPointFeature *feature = [[MGLPointFeature alloc] init];
+        feature.coordinate = coordinateFromPointGeometry(geometry);
+        return feature;
+    }
+    
+    MGLShape <MGLFeaturePrivate> * operator()(const mapbox::geometry::line_string<T> &geometry) const {
+        std::vector<CLLocationCoordinate2D> coordinates;
```
jfirebaugh: Factor out a helper method: `std::vector<CLLocationCoordinate2D> toLocationCoordinates2D(const std::vector<mbgl::Point>&) const`.

platform/darwin/src/MGLFeature.mm
```diff
@@ -0,0 +1,299 @@
...
+};
+
+/**
+ Transforms an `mbgl::geometry::geometry` type into an instance of the
+ corresponding Objective-C geometry class.
+ */
+template <typename T>
+class GeometryEvaluator {
+public:
+    MGLShape <MGLFeaturePrivate> * operator()(const mapbox::geometry::point<T> &geometry) const {
```
jfirebaugh: Prefer the `mbgl` typedefs: `mbgl::Point<T>`, `mbgl::LineString<T>`, etc. The exception is `mapbox::geometry::geometry_collection`, for which we still need to clear away [the name](https://github.com/mapbox/mapbox-gl-native/blob/78fe86fcb14a08a625ad7e476c721f46f93e520b/src/mbgl/tile/geometry_tile.hpp#L40).

platform/darwin/src/MGLFeature.mm
1ec5: Removed the vestigial constructor and force-pushed.

src/mbgl/annotation/shape_annotation_impl.cpp
```diff
@@ -88,21 +88,22 @@ void ShapeAnnotationImpl::updateTile(const CanonicalTileID& tileID, AnnotationTi
         const double tolerance = baseTolerance / (maxAmountOfTiles * util::EXTENT);
 
         geojsonvt::ProjectedRings rings;
-        std::vector<geojsonvt::LonLat> points;
```
jfirebaugh: These changes will conflict with https://github.com/mapbox/mapbox-gl-native/compare/5158-minimal but the resolution should be straightforward.

platform/darwin/src/MGLFeature.mm
1ec5: Building up an `std::vector` can be faster than building up an NSMutableArray, but this is just what’s left from some earlier tortured logic rather than the product of a profile. The optimization wouldn’t matter much in this case anyways.

1ec5:
Here’s the new feature’s documentation:

mglmapview

mglfeature

jfirebaugh:
How would the API change if we added the the `filter` option supported by gl-js? Or other options? Is it extensible without breaking backward compatibility?

1ec5:
> How would the API change if we added the the `filter` option supported by gl-js? Or other options? Is it extensible without breaking backward compatibility?

We’d add additional variations of this method, like `-visibleFeaturesInRect:inStyleLayersWithIdentifiers:filteredWithPredicates:` (or something to that effect). That won’t scale too well if we have more than 3–4 options, but there aren’t any good alternatives in Objective-C.

The typical solution in C++ would be to pass in a struct representing the query, but structs in Objective-C don’t support ARC and thus can’t hold Objective-C objects like NSString. (This is fixed in Swift, but without any backwards compatibility for Objective-C.) We went down this road in the past with `-setCenterCoordinate:…` before we introduced `-setCamera:`. But MGLMapCamera is a full-fledged class, with all the cognitive and syntactic overhead that entails.

I don’t think we should introduce a class with only one property. If lots more options become available down the road, or if MGLStyleLayer becomes a type the developer can interact with, we can do the deprecation dance with these methods at that point.

1ec5:
I’ve tweaked the Swift names of these MGLMapView methods to be briefer and accommodate additional options in the future. Hopefully it won’t be too confusing for the Swift names to be slightly different than the Objective-C names. Eventually, we can also [refine the declarations for Swift](https://developer.apple.com/library/ios/documentation/Swift/Conceptual/BuildingCocoaApps/MixandMatch.html#//apple_ref/doc/uid/TP40014216-CH10-ID188), marking each labeled parameter as optional so you can specify any permutation of these parameters.

platform/darwin/src/MGLFeature.h
```diff
@@ -0,0 +1,133 @@
...
+ */
+@property (nonatomic, copy, nullable, readonly) id identifier;
+
+/**
+ A dictionary of attributes specified by the vector tile source for this
+ feature.
+ 
+ The keys and values of this dictionary are determined by the vector tile
+ source. Vector tiles are stored in <a href="http://geojson.org/">GeoJSON</a>
+ format; they additionally conform to the
```
lucaswoj: What does it mean to store a vector tile in GeoJSON format?

 > Vector tiles are stored in <a href="http://geojson.org/">GeoJSON</a> format?

platform/ios/src/MGLMapView.h
```diff
@@ -1021,6 +1032,182 @@ IB_DESIGNABLE
...
 
+#pragma mark Accessing the Underlying Map Data
+
+/**
+ Returns an array of rendered map features that intersect with a given point.
+ 
+ Each object in the returned array represents a feature rendered by the
+ current style and provides access to attributes specified by the relevant
+ vector tile sources. The returned array only includes features specified in
+ vector tile sources, not raster tile sources or other kinds of sources.
```
lucaswoj: Does this method support GeoJSON sources?

1ec5: Yes, although we don’t have a way to add GeoJSON sources at runtime yet: #3630.

platform/ios/src/MGLMapView.h
```diff
@@ -1021,6 +1032,182 @@ IB_DESIGNABLE
...
+ Only visible features are returned. For example, suppose the current style uses
+ the
+ <a href="https://www.mapbox.com/vector-tiles/mapbox-streets/">Mapbox Streets source</a>,
+ but none of its style layers includes features that have the `maki` property
+ set to `bus`. If you pass a point corresponding to the location of a bus stop
+ into this method, the bus stop feature does not appear in the resulting array.
+ On the other hand, if the style does include bus stops, an `MGLFeature` object
+ representing that bus stop is returned and its `featureAttributes` dictionary
+ has the `maki` key set to `bus` (along with other attributes). The dictionary
+ does not indicate how the feature is rendered by the current style.
```
lucaswoj: This explanation is good and correct. As a thought exercise, I wonder what it'd look like to frame this description more in terms of what the API _does_ rather than what it _doesn't_ do. 

This API returns the set of features that are rendered on the map at a point.

1ec5: I think the first part of the paragraph explains what it does do – or at least, that’s what I was trying to do. The point I was trying to make at the end is that you shouldn’t expect paint properties to be included in the attribute dictionary.

platform/darwin/src/MGLFeature.h
1ec5: Oops, that’s incorrect and it’s the next part, about the vector tile spec, that we should be emphasizing anyways.

PR #5122, opened 2016-05-24
Share shaders with gl-js

jfirebaugh:
See https://github.com/mapbox/mapbox-gl-js/pull/2599.

Note that this adds a build dependency on npm to all platforms. Previously, it was a build dependency for node, and a test dependency for node, osx, qt, and linux. If there's resistance to this, I can explore using a submodule instead. But I'd prefer to use npm.


1ec5:
> Note that this adds a build dependency on npm to all platforms.

Can you update INSTALL.md to explain how to get `npm`? The platform-specific INSTALL.md documents refer to the main INSTALL.md for core dependencies.

INSTALL.md
```diff
@@ -13,15 +13,10 @@ depending on your operating system and what platform you want to develop for.
...
 ## 1: Getting the Source
 
-Mapbox GL Native uses git submodules, so you'll need to initialize and update
-them after cloning the source from Git.
+Clone the git repository:
 
     git clone https://github.com/mapbox/mapbox-gl-native.git
     cd mapbox-gl-native
-    git submodule init
-    git submodule update
```
1ec5: Submodules are still used for the KIF-based iOS SDK UI tests, so I moved this step to the iOS SDK testing guide in 102e62ae5582ec5cad5d8bbe3d42f99b875d143e. Once enough of the KIF tests have been ported to XCUITest in #4837, we can delete the KIF tests outright, and the two remaining submodules will be no more.

PR #5124, opened 2016-05-24
[ios] Set backup exclusion flag after initializing DefaultFileSource

### LINKED ISSUE #5112: MGLOfflineStorage iCloud Backup
We're using MGLOfflineStorage to download offline maps as shown in your example code. And it seems that downloaded data gets backed up to iCloud. Which is against Apple guidelines. Actually we found it out after being rejected. 

Steps to reproduce:

・Check iCloud backup size for an app in settings before downloading offline region.
・Check iCloud backup size after

friedbunny:
Speculatively fixes #5112, where apps mistakenly backup the offline/ambient cache.

This moves the backup exclusion flag to a point just after `DefaultFileSource` has been created/initialized, ensuring that this flag is set on every launch.

As with the other offline storage tests, the new `-testBackupExclusion` must be run in alphabetical order — or, at least after the cache has been initialized. The cache file will persist in the simulator when you run these tests locally.

/cc @1ec5

### DIFF
--- a/platform/darwin/test/MGLOfflineStorageTests.m
+++ b/platform/darwin/test/MGLOfflineStorageTests.m
@@ -116,7 +116,7 @@
     [cacheURL getResourceValue:&exclusionFlag
                         forKey:NSURLIsExcludedFromBackupKey
                          error:&error];
-    XCTAssertTrue([exclusionFlag boolValue], @"Backup exclusion flag should be set for cache database.");
+    XCTAssertTrue(exclusionFlag && [exclusionFlag boolValue], @"Backup exclusion flag should be set for cache database.");
     XCTAssertNil(error, @"No errors should be returned when checking backup exclusion flag.");
 }
 
--- a/platform/darwin/test/MGLOfflineStorageTests.m
+++ b/platform/darwin/test/MGLOfflineStorageTests.m
@@ -97,6 +97,29 @@
     [self waitForExpectationsWithTimeout:1 handler:nil];
 }
 
+- (void)testBackupExclusion {
+    NSURL *cacheDirectoryURL = [[NSFileManager defaultManager] URLForDirectory:NSApplicationSupportDirectory
+                                                                      inDomain:NSUserDomainMask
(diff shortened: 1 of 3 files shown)


platform/darwin/test/MGLOfflineStorageTests.m
```diff
@@ -97,6 +97,29 @@ - (void)testAddPack {
...
+
+    NSURL *cacheURL = [cacheDirectoryURL URLByAppendingPathComponent:@"cache.db"];
+    XCTAssertTrue([[NSFileManager defaultManager] fileExistsAtPath:cacheURL.path], @"Cache database should exist.");
+
+    NSError *error = nil;
+    NSNumber *exclusionFlag = nil;
+    [cacheURL getResourceValue:&exclusionFlag
+                        forKey:NSURLIsExcludedFromBackupKey
+                         error:&error];
+    XCTAssertTrue([exclusionFlag boolValue], @"Backup exclusion flag should be set for cache database.");
```
1ec5: Nit: check for `nil` before downcasting to a `BOOL`. Either would be a failure in this case, but this is a reminder in case we’ve made a similar assumption elsewhere.

friedbunny: Which would be the runtime behavior when `exclusionFlag == nil`?

```
(lldb) po [exclusionFlag boolValue]
<nil>

(lldb) p [exclusionFlag boolValue]
(BOOL) $2 = NO
```

`nil`, no?

1ec5: When `exclusionFlag` is `nil`, `[exclusionFlag boolValue]` is `NO`, just as if `exclusionFlag` had been equal to `@NO`. By invoking the `po` command, you’re treating the `BOOL` as an object pointer and attempting to print its `-description`, which is `nil`.

friedbunny: Added an extra check that `exclusionFlag` exists in 6becc07.

PR #5125, opened 2016-05-24
MapboxAccountManager And Offline Support

### LINKED ISSUE #5058: Offline manager access token
`MapboxAccountManager` has been added in `4.1.0` but we still ask for access token when setting up the `OfflineManager` object. We should look into using account manager to handle this in the background so users don't need to call `setAccessToken` still and write something like this:

```java
offlineManager.setAccessToken(MapboxAccountManager.getInstance().getAccessToken());
```

bleege:
Closes #5058


bleege:
Rebased and Merged!

PR #5012, opened 2016-05-25
Move updateRenderables to clean API

kkaefer:
the `updateRenderables` algorithm currently expects specific maps and collections; We should instead hide this behind an API to simplify testing and expansion.


kkaefer:
This converts the `updateRenderables` algorithm to take a few lambdas that perform the respective actions of obtaining a TileData object, creating, retaining and rendering it. In the test suite, we can inject a log recorder into these to verify that the actions are executed in the correct order.

Prerequisite for #123

jfirebaugh:
Didn't go too deep on this but it seems like a logical path to making it more testable. Should we rename `update_renderables_impl.hpp` to `update_renderables.hpp`, since everything includes the `impl.hpp` anyway?

PR #5139, opened 2016-05-25
Remove TileData::State

kkaefer:
It's time to remove the `TileData::State` enum. This enum holds linear state and should represent the lifecycle of a tile, but it's a stretch to say that it's linear. This removes it completely in favor of a clear API and a separate `obsolete` variable for `VectorTileData` objects.


jfirebaugh:
Awesome. Looks great! 👍

PR #5141, opened 2016-05-25
Fix texture leak in TexturePoolHolder

### LINKED ISSUE #5136: Raster/TexturePoolHolder leaks textures
We're seeing a lot of "out of memory"-crashes after a while of heavy panning (iPad Air 2, a raster only stylesheet containing two raster-layers, both using 1024px tiles). 

Trying to track down what's eating all the memory I found that in gl_object_store.cpp, `TexturePoolHolder::reset()` it seems textures are never deleted due to the line [`if (!bool()) return;`](https://github.com/mapbox/mapbox-gl-native/blob/master/src/mbgl/gl/gl_object_store.cpp#L63).
There's also the line [`if (bool()) return;`](https://github.com/mapbox/mapbox-gl-native/blob/master/src/mbgl/gl/gl_object_store.cpp#L57) which to me seems to have no purpose?
Or am I missing something?

I can also add that, with those lines removed `TexturePoolHolder::reset()` seems to always be called having a NULL `objectStore` when called from [`Raster::~Raster()`](https://github.com/mapbox/mapbox-gl-native/blob/master/src/mbgl/util/raster.cpp#L18)...

brunoabinader:
Fixes a misused `operator bool()` plus usage of an undefined pointer in `{GL,TexturePool}Holder` code.

:eyes: @kkaefer @jfirebaugh 

Fixes #5136.

### DIFF
--- a/src/mbgl/gl/gl_object_store.cpp
+++ b/src/mbgl/gl/gl_object_store.cpp
@@ -6,61 +6,61 @@ namespace mbgl {
 namespace gl {
 
 void ProgramHolder::create(GLObjectStore& objectStore_) {
-    if (id) return;
+    if (created()) return;
     objectStore = &objectStore_;
     id = MBGL_CHECK_ERROR(glCreateProgram());
 }
 
 void ProgramHolder::reset() {
-    if (!id) return;
+    if (!created()) return;
     objectStore->abandonedPrograms.push_back(id);
     id = 0;
 }
 
 void ShaderHolder::create(GLObjectStore& objectStore_) {
-    if (id) return;
+    if (created()) return;
     objectStore = &objectStore_;
     id = MBGL_CHECK_ERROR(glCreateShader(type));
 }
 
 void ShaderHolder::reset() {
-    if (!id) return;
+    if (!created()) return;
     objectStore->abandonedShaders.push_back(id);
     id = 0;
 }
 
 void BufferHolder::create(GLObjectStore& objectStore_) {
-    if (id) return;
+    if (created()) return;
     objectStore = &objectStore_;
     MBGL_CHECK_ERROR(glGenBuffers(1, &id));
 }
 
 void BufferHolder::reset() {
-    if (!id) return;
+    if (!created()) return;
     objectStore->abandonedBuffers.push_back(id);
     id = 0;
 }
 
 void TextureHolder::create(GLObjectStore& objectStore_) {
-    if (id) return;
+    if (created()) return;
     objectStore = &objectStore_;
     MBGL_CHECK_ERROR(glGenTextures(1, &id));
 }
 
 void TextureHolder::reset() {
-    if (!id) return;
+    if (!created()) return;
     objectStore->abandonedTextures.push_back(id);
     id = 0;
 }
 
 void TexturePoolHolder::create(GLObjectStore& objectStore_) {
-    if (bool()) return;
+    if (created()) return;
     objectStore = &objectStore_;
     MBGL_CHECK_ERROR(glGenTextures(TextureMax, ids.data()));
 }
 
 void TexturePoolHolder::reset() {
-    if (!bool()) return;
+    if (!created()) return;
     for (GLuint& id : ids) {
         if (id == 0) continue;
         objectStore->abandonedTextures.push_back(id);
@@ -69,13 +69,13 @@ void TexturePoolHolder::reset() {
 }
 
 void VAOHolder::create(GLObjectStore& objectStore_) {
-    if (id) return;
+    if (created()) return;
     objectStore = &objectStore_;
     MBGL_CHECK_ERROR(gl::GenVertexArrays(1, &id));
 }
 
 void VAOHolder::reset() {
-    if (!id) return;
+    if (!created()) return;
     objectStore->abandonedVAOs.push_back(id);
     id = 0;
 }
--- a/src/mbgl/gl/gl_object_store.cpp
+++ b/src/mbgl/gl/gl_object_store.cpp
@@ -61,12 +61,11 @@ void TexturePoolHolder::create(GLObjectStore& objectStore_) {
 
 void TexturePoolHolder::reset() {
     if (!bool()) return;
-    for (GLuint id : ids) {
-        if (id) {
-            objectStore->abandonedTextures.push_back(id);
-        }
-    }
-    ids.fill(0);
+    for (GLuint& id : ids) {
+        if (id == 0) continue;
+        objectStore->abandonedTextures.push_back(id);
+        id = 0;
+    };
 }
 
 void VAOHolder::create(GLObjectStore& objectStore_) {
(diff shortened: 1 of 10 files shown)


src/mbgl/gl/gl_object_store.cpp
```diff
@@ -61,12 +61,11 @@ void TexturePoolHolder::create(GLObjectStore& objectStore_) {
 
 void TexturePoolHolder::reset() {
     if (!bool()) return;
-    for (GLuint id : ids) {
-        if (id) {
-            objectStore->abandonedTextures.push_back(id);
-        }
-    }
-    ids.fill(0);
+    std::for_each(ids.begin(), ids.end(), [&](GLuint& id) {
```
jfirebaugh: Why the switch from range-based `for` to `for_each`?

brunoabinader: You are right, I can achieve the same thing capturing a non-const reference in a range-based `for`.

jfirebaugh:
The intended invariant is that if `{GLObjectStore,TexturePoolHolder}` is owning a resource, its `objectStore` is not null. It shouldn't be necessary to check both the resource and `objectStore`.

I think the actual bug lies in the implementations of the move constructors and `operator=`s. They need to transfer `objectStore` along with the resource.

jfirebaugh:
Here's how I would do if implementing this pattern from scratch:

* Pull in a polyfill for C++XX `unique_resource` (https://github.com/okdshin/unique_resource)
* `using UniqueProgram = std::unique_resource<GLuint, ProgramDeleter>;`

```
struct ProgramDeleter {
    GLObjectStore* store;
    void operator()(GLuint id) {
        assert(store);
        store->abandonedPrograms.push_back(id);
    }
};
```
* `UniqueProgram GLObjectStore::createProgram()`

brunoabinader:
@jfirebaugh my previous implementation used `std::unique_ptr` but as suggested by @kkaefer in https://github.com/mapbox/mapbox-gl-native/commit/02b3771a5891207f0243d1755347d011633bbfb3#commitcomment-16093745 we ditched that to save on lesser redirections. I'm :+1: on revisiting this if we all agree on the implementation pros/cons.

src/mbgl/gl/gl_object_store.cpp
```diff
@@ -6,77 +6,76 @@ namespace mbgl {
...
-    if (!bool()) return;
-    for (GLuint id : ids) {
-        if (id) {
-            objectStore->abandonedTextures.push_back(id);
-        }
-    }
-    ids.fill(0);
+    if (!created()) return;
+    for (GLuint& id : ids) {
+        if (id == 0) return;
```
jfirebaugh: The `return` needs to revert back to a conditional.

src/mbgl/gl/gl_object_store.hpp
```diff
@@ -107,10 +108,11 @@ class TexturePoolHolder : private util::noncopyable {
     TexturePoolHolder() { ids.fill(0); }
     ~TexturePoolHolder() { reset(); }
 
-    TexturePoolHolder(TexturePoolHolder&& o) noexcept : ids(std::move(o.ids)) {}
-    TexturePoolHolder& operator=(TexturePoolHolder&& o) noexcept { ids = std::move(o.ids); return *this; }
+    TexturePoolHolder(TexturePoolHolder&& o) noexcept : ids(std::move(o.ids)), objectStore(o.objectStore) { o.ids.fill(0); }
+    TexturePoolHolder& operator=(TexturePoolHolder&& o) noexcept { ids = std::move(o.ids); objectStore = o.objectStore; return *this; }
 
-    explicit operator bool() { return std::none_of(ids.begin(), ids.end(), [](int id) { return id == 0; }); }
+    operator bool() = delete;
```
jfirebaugh: There is no compiler-generated default `operator bool()`, so you don't need this.

src/mbgl/gl/gl_object_store.cpp
brunoabinader: Ah yes, meant `continue` there.

src/mbgl/gl/gl_object_store.hpp
```diff
@@ -38,10 +38,10 @@ class GLHolder : private util::noncopyable {
 public:
     GLHolder() {}
 
-    GLHolder(GLHolder&& o) noexcept : id(o.id) { o.id = 0; }
-    GLHolder& operator=(GLHolder&& o) noexcept { id = o.id; o.id = 0; return *this; }
+    GLHolder(GLHolder&& o) noexcept : id(o.id), objectStore(o.objectStore) { id = 0; }
```
jfirebaugh: Why did you change `o.id = 0` to `id = 0`? Isn't the former correct?

src/mbgl/gl/gl_object_store.hpp
```diff
@@ -38,10 +38,10 @@ class GLHolder : private util::noncopyable {
...
+    GLHolder& operator=(GLHolder&& o) noexcept { id = o.id; objectStore = o.objectStore; return *this; }
```
jfirebaugh: Likewise, this still needs to set `o.id = 0`.

brunoabinader: Bad rebase, my bad.

brunoabinader:
We'll proceed with @jfirebaugh's suggestion from https://github.com/mapbox/mapbox-gl-native/pull/5141#issuecomment-221719872 in a separate PR.

haberbyte:
> Here's how I would do if implementing this pattern from scratch:
> 
>     * Pull in a polyfill for C++XX `unique_resource` (https://github.com/okdshin/unique_resource)
> 
>     * `using UniqueProgram = std::unique_resource<GLuint, ProgramDeleter>;`
> 
> 
> ```
> struct ProgramDeleter {
>     GLObjectStore* store = nullptr;
>     void operator()(GLuint id) {
>         assert(store);
>         store->abandonedPrograms.push_back(id);
>     }
> };
> ```
> 
> ```
> UniqueProgram GLObjectStore::createProgram() {
>     return std::make_unique_resource(MBGL_CHECK_ERROR(glCreateProgram()), ProgramDeleter(this));
> }
> ```

Hey @jfirebaugh, i came across this recently and was wondering if you would still do it this way.
I am wondering if the same thing could be accomplished by simply using std::unique_ptr, either wrapping each type in its own class or even by passing a deleter to the unique_ptr, e.g.
```struct TextureDeleter {
    Context* context;
    TextureDeleter(Context* c) : context(c) {}
    void operator()(TextureID*) const;
};
```

Why is the unique_resource.hpp needed? Would love to understand this, i am probably missing something!

kkaefer:
@haberbyte `unique_resource` is similar, but not identical to `unique_ptr`: It manages resources that are represented as integers (e.g. as returned by OpenGL API calls) instead of pointers. You could create a wrapper class for every resource type that stores the integer handle, but that means that you'd have to dereference a pointer every time you want to access the resource (i.e. you essentially store a pointer to an integer). In reality though, the performance benefit of storing the integer directly is probably pretty small.

haberbyte:
Ah, that makes sense, thank you! Unfortunate, that it is not even part of C++17  🤷‍♂

PR #5142, opened 2016-05-25
[ios] Fix points/shapes typo that caused crash

friedbunny:
https://github.com/mapbox/mapbox-gl-native/pull/5097 introduced different arrays for points and shapes, but only ever attempted to use the points array. Adding a shape would crash.

/cc @1ec5


1ec5:
👍

PR #5145, opened 2016-05-25
[earcut] Drop smallest inner rings for earcut performance

yhahn:
Port of JS version + unit tests: https://github.com/mapbox/mapbox-gl-js/pull/2622

cc @jfirebaugh @mourner @jakepruitt


src/mbgl/renderer/fill_bucket.cpp
```diff
@@ -35,7 +35,11 @@ FillBucket::~FillBucket() {
 }
 
 void FillBucket::addGeometry(const GeometryCollection& geometry) {
-    for (const auto& polygon : classifyRings(geometry)) {
+    // Max number of rings to allow per polygon.
+    // Optimizes polygons with many interior rings for earcut tesselation.
+    unsigned maxRings = 500;
```
yhahn: @jfirebaugh not sure where the right place for this is.

mourner: Let's use `uint32_t` type instead of `unsigned` everywhere — it's more predictable across platforms.

src/mbgl/tile/geometry_tile.cpp
```diff
@@ -100,16 +105,39 @@ std::vector<GeometryCollection> classifyRings(const GeometryCollection& rings) {
...
+    if (!polygon.empty()) {
+        std::sort(innerRings.begin(), innerRings.end(), sortByArea);
+        unsigned ringCount = 1;
+        for (const auto& pair : innerRings) {
+            if (maxRings > 0 && ringCount > maxRings - 1) break;
+            polygon.push_back(pair.second);
+            ringCount++;
+        }
         polygons.push_back(polygon);
+    }
```
mourner: Can we avoid duplicating code here, like in JS version?

mourner:
Similar to my JS PR comment, we should use [`std::nth_element`](http://en.cppreference.com/w/cpp/algorithm/nth_element) instead of `std::sort` here for performance and convenience.

jfirebaugh:
Going to fix the above issues and merge this into the `earcut` branch.

PR #5147, opened 2016-05-25
[android] Remove logging of location due to privacy leak.

roblabs:
* Location logs are not just kept on the device.
* Some logging packages, e.g., Firebase, now have the option to upload logs to cloud.
* This closes a potential privacy leak.

### DIFF
--- a/platform/android/MapboxGLAndroidSDK/src/main/java/com/mapbox/mapboxsdk/location/LocationServices.java
+++ b/platform/android/MapboxGLAndroidSDK/src/main/java/com/mapbox/mapboxsdk/location/LocationServices.java
@@ -123,7 +123,7 @@ public class LocationServices implements com.mapzen.android.lost.api.LocationLis
      */
     @Override
     public void onLocationChanged(Location location) {
-        Log.d(TAG, "onLocationChanged()..." + location);
+//        Log.d(TAG, "onLocationChanged()..." + location);
         this.lastLocation = location;
 
         // Update Listeners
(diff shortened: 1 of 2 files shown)


friedbunny:
Thanks for this, though I wonder if we’ll want to put such logging behind a debug setting of some sort. /cc @bleege @tobrun

roblabs:
Yes, I agree on the debug logging.  I left it as a comment, instead of removing, in case the original author wanted to use it for something.

Your solution idea sounds best.

tobrun:
Thank you @roblabs for bringing this to our attention.

@friedbunny 
 - does iOS have debug logging for this? 
 - is something like that automatically disabled when you create release build?

bleege:
Thanks @roblabs.

> Yes, I agree on the debug logging. I left it as a comment, instead of removing, in case the original author wanted to use it for something.

Yep, this sounds like the right approach to me.

friedbunny:
@tobrun Yes and yes. iOS has a telemetry debug logging setting that logs to file and the console. Console logs [have coordinates redacted](https://github.com/mapbox/mapbox-gl-native/blob/8ace0d37d10e8174268f259dec38cf88e2494ef0/platform/ios/src/MGLMapboxEvents.m#L717-L718). If we think the app build is for app store distribution or wide-release, [we disable the debug logging](https://github.com/mapbox/mapbox-gl-native/blob/8ace0d37d10e8174268f259dec38cf88e2494ef0/platform/ios/src/MGLMapboxEvents.m#L725-L765).

The debug logging on iOS is only for telemetry right now, but @boundsj has indicated an interest in eventually expanding it.

friedbunny:
I should mention: the debug logging on iOS, to this point, hasn’t really been for public consumption and the setting is undocumented.

bleege:
Discussed with @tobrun internally and we're going to ship this in Android 4.1.0.  Since this PR is against `master` due to the release schedule yesteday I'll first merge it there and then cherry pick it over to `release-android-v4.1.0`.

bleege:
Rebased and Merged on `master`.

bleege:
Cherry picked on to `release-android-v4.1.0`.  Thanks @roblabs for contributing to the project!

screen shot 2016-05-27 at 10 52 31 am

Issue #5140, opened 2016-05-25
Add a container view for UIView based annotations


boundsj:
In https://github.com/mapbox/mapbox-gl-native/pull/4801 we added the ability for application code to provide `UIView` based representations for annotations. Currently, the annotation views are added as subviews of the map view. This [has potential performance implications](http://blog.benjamin-encz.de/post/disassembling-uikit-tintcolor-visitor) and is probably the source if issues [like this](https://github.com/mapbox/mapbox-gl-native/issues/5089).

We should add a container view similar to MapKit to contain the annotation views and perhaps other containers for other UIView based visuals. This should help reduce the complexity of layering UIView based visuals over the map.

#### MapKit annotation container

![screen shot 2016-05-25 at 12 09 37 pm](https://cloud.githubusercontent.com/assets/63186/15553003/ee16ded2-2271-11e6-9014-f1167a45d6e0.png)

boundsj:
@1ec5 @friedbunny I created WIP PR (https://github.com/mapbox/mapbox-gl-native/pull/5148) for your review.

Issue #5089, opened 2016-05-21
Flat annotation views pierce through callout views


1ec5:
When the map is tilted, annotation views with the `flat` property set to true pierce through any open callout view:

![piercing-pins](https://cloud.githubusercontent.com/assets/1231218/15446411/2f2a1ee8-1ed5-11e6-9d7c-696bfacd72d7.PNG)

/cc @boundsj @friedbunny

PR #5148, opened 2016-05-26
[ios] Add annotation container view

boundsj:
Fixes https://github.com/mapbox/mapbox-gl-native/issues/5140
Fixes https://github.com/mapbox/mapbox-gl-native/issues/5089

Adds a new UIView subclass to contain MGLAnnotationView instances and an API for helping to manage the addition of new annotations. To mitigate [known performance issues with `UIView:addSubview:`](http://blog.benjamin-encz.de/post/disassembling-uikit-tintcolor-visitor), this batches up the addition of subviews to the container _before_ adding the container itself to the view hierarchy. The organization of the views in the separate container also naturally avoids the "piercing" issue seen before between annotations and callouts.


platform/ios/app/MBXAnnotationView.m
```diff
@@ -12,7 +12,7 @@ - (void)layoutSubviews {
     [super layoutSubviews];
     if (!self.centerView) {
         self.backgroundColor = [UIColor blueColor];
-        self.centerView = [[UIView alloc] initWithFrame:CGRectInset(self.bounds, 5.0, 5.0)];
+        self.centerView = [[UIView alloc] initWithFrame:CGRectInset(self.bounds, 1.0, 1.0)];
```
1ec5: MBXAnnotationView is also used for dropped pins, but this is kind of small for that demo. Let’s stop using an annotation view for dropped pins, so we get better visibility on GL annotations, which are still supported. MBXAnnotationView would continue to be used for other demos like “Add _n_ Point Annotations”.

platform/ios/app/MBXViewController.m
```diff
@@ -553,10 +553,18 @@ - (MGLAnnotationView *)mapView:(MGLMapView *)mapView viewForAnnotation:(id<MGLAn
     if (!annotationView)
     {
         annotationView = [[MBXAnnotationView alloc] initWithReuseIdentifier:MBXViewControllerAnnotationViewReuseIdentifer];
-        annotationView.frame = CGRectMake(0, 0, 40, 40);
+        annotationView.frame = CGRectMake(0, 0, 10, 10);
```
1ec5: MBXAnnotationView is also used for dropped pins, but this is kind of small for that demo. Let’s stop using an annotation view for dropped pins, so we get better visibility on GL annotations, which are still supported. MBXAnnotationView would continue to be used for other demos like “Add _n_ Point Annotations”.

platform/ios/src/MGLAnnotationContainerView.h
```diff
@@ -0,0 +1,17 @@
...
+
+#import "MGLTypes.h"
+
+@class MGLAnnotationView;
+
+NS_ASSUME_NONNULL_BEGIN
+
+@interface MGLAnnotationContainerView : UIView
+
++ (instancetype)annotationContainerViewWithAnnotationContainerView:(nonnull MGLAnnotationContainerView *)annotationContainerView;
```
1ec5: `nonnull` is unnecessary here because of the `NS_ASSUME_NONNULL_BEGIN` guard above.

platform/ios/src/MGLMapView.mm
```diff
@@ -2909,6 +2912,23 @@ - (void)addAnnotations:(NS_ARRAY_OF(id <MGLAnnotation>) *)annotations
...
+    {
+        // reload any previously added views
+        newAnnotationContainerView = [MGLAnnotationContainerView annotationContainerViewWithAnnotationContainerView:self.annotationContainerView];
+        [self.annotationContainerView removeFromSuperview];
+    }
+    else
+    {
+        newAnnotationContainerView = [[MGLAnnotationContainerView alloc] initWithFrame:self.bounds];
+    }
+    newAnnotationContainerView.autoresizingMask = UIViewAutoresizingFlexibleWidth | UIViewAutoresizingFlexibleHeight;
```
1ec5: Also set the `contentMode` to `UIViewContentModeCenter` like we do for the GLKView. That way the content remains centered correctly as the device rotates and probably in some other edge cases as well.

1ec5:
The saga continues in #5194.

PR #5155, opened 2016-05-26
[android] #3629 - smooth animation for non location tracking

### LINKED ISSUE #3629: Smooth animation in user tracking mode
Currently on Android the UserLocationView animates from one location update to another.
You can clearly see the different location updates coming in and the it doesn't seem that the user is traveling at a constant speed. To properly support the idea of implementing navigation, we should streamline location updates/animation.

There are two ways to improve this user experience:
 - Entity interpolation
 - Dead reckoning

On iOS they have chosen to implement Entity interpolation which means that the user dot is actually one step behind the actual location, it is always trying to catch up.

@1ec5 @bleege @zugaldia

tobrun:
closes #3629


tobrun:

PR #5156, opened 2016-05-26
[ios] canBecomeFirstResponder

halset:
I want to use a UIMenuController on a MGLMapView in my app. For that to work, MGLMapView#canBecomeFirstResponder need to return YES.


1ec5:
Thanks!

PR #5157, opened 2016-05-26
[Qt] QMapboxGLPrivate::getPixelRatio() cleanup

brunoabinader:
From http://doc.qt.io/qt-5/qguiapplication.html#devicePixelRatio, `QWindow::devicePixelRatio` is the most reliable pixel ratio because `QGuiApplication` returns the maximum pixel ratio of all available `QScreen` objects. This is not valid for cases e.g. where two or more `QScreen` objects with different pixel ratios are present and the application window appears on the screen with lower pixel ratio.

This patch also moved `glViewport` call to QMapboxGL domain.

:eyes: @tmpsantos @kkaefer


tmpsantos:
:+1:

PR #5162, opened 2016-05-26
[node] Don't fire callback when the AsyncRequest was canceled

kkaefer:
When requesting a resource in the node bindings, we're currently [returning a plain `AsyncRequest` object](https://github.com/mapbox/mapbox-gl-native/blob/c111250fcccaf5e57154852606ee740f0db242c1/platform/node/src/node_map.cpp#L536). This means that when we cancel the file request, we're not canceling the associated JS `request()` call. We can't do that because it's immediate, but that immediate function call is async, so it could fire the passed `callback` function later on, at a point when the `AsyncRequest` has long been destroyed.


jfirebaugh:
👍 

Thanks for tracking this down.

Issue #4838, opened 2016-04-25
Multipoint annotations should be accessible to VoiceOver


1ec5:
Point annotations are now accessible to VoiceOver, but multipoints (polylines and polygons) are not. Multipoints have no callout views, but it should still be possible to summarize them as, for instance, “{title}, northeast to southwest”.

/cc @friedbunny

1ec5:
This is blocked on a runtime layer API (#4839), rewriting `AnnotationManager` based on a style layer inserted at runtime, and feature eventing (#352).

1ec5:
#352 and #5200 have landed, so this issue is now blocked by #5165.

PR #5165, opened 2016-05-27
Replace getPointAnnotationsInBounds() w/ queryPointAnnotations()

### LINKED ISSUE #5151: getPointAnnotationsInBounds with a rotated map
This is a follow up issue on #5051. In that issue we implemented that camera animations and projection should take MapView padding into account. While testing my changes for projection, I started to notice that `getAnnotationsInBounds` was not returning the actual annotations for the projection. 

After some debugging and playing around with some values. I started to notice that having `tilt = 0` and `direction = 0` returns the correct annotations in bounds, When the map is rotated we aren't receiving the correct values anymore.

### Example
In below gif I'm seeing the top left marker as an annotations in bounds even though it's not a part of the current projection. When I make the map face north again it becomes a part of the projection.

ezgif com-video-to-gif 23

1ec5:
`queryPointAnnotations()` accepts a screen rectangle instead of a geographic bounding box. If you pass in the bounds (in screen coordinates) of a rotated, tilted map view, the query is restricted to the trapezoidal region visible in the map view.

Fixes #5151.

@jfirebaugh to review the C++ side. @tobrun to review the Java side, which I edited blind without a Java IDE. 😬

### DIFF
--- a/src/mbgl/map/map.cpp
+++ b/src/mbgl/map/map.cpp
@@ -700,10 +700,6 @@ void Map::removeAnnotation(AnnotationID annotation) {
     update(Update::AnnotationStyle | Update::AnnotationData);
 }
 
-AnnotationIDs Map::getPointAnnotationsInBounds(const LatLngBounds& bounds) {
-    return impl->annotationManager->getPointAnnotationsInBounds(bounds);
-}
-
 #pragma mark - Feature query api
 
 std::vector<Feature> Map::queryRenderedFeatures(const ScreenCoordinate& point, const optional<std::vector<std::string>>& layerIDs) {
@@ -730,6 +726,18 @@ std::vector<Feature> Map::queryRenderedFeatures(const ScreenBox& box, const opti
         impl->transform.getState(),
         layerIDs
     });
+}
+
+AnnotationIDs Map::queryPointAnnotations(const ScreenBox& box) {
+    auto features = queryRenderedFeatures(box, {{ AnnotationManager::PointLayerID }});
+    AnnotationIDs ids;
+    ids.reserve(features.size());
+    for (auto &feature : features) {
+        assert(feature.id);
+        assert(*feature.id <= std::numeric_limits<AnnotationID>::max());
+        ids.push_back(static_cast<AnnotationID>(feature.id->get<uint64_t>()));
+    }
+    return ids;
 }
 
 #pragma mark - Style API
--- a/include/mbgl/map/map.hpp
+++ b/include/mbgl/map/map.hpp
@@ -148,8 +148,6 @@ public:
     void updateAnnotation(AnnotationID, const Annotation&);
     void removeAnnotation(AnnotationID);
 
-    AnnotationIDs getPointAnnotationsInBounds(const LatLngBounds&);
-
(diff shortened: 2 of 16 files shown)


1ec5:
This isn’t going to work: annotation views on iOS and marker views on Android work by inserting a point annotation that has no icon and is thus prevented from rendering at the GL level. But as the name suggests, `queryRenderedFeatures()` only returns GL-rendered features, so if the application is using only annotation views or marker views, this new method always returns an empty vector.

This approach will only work if we use `querySourceFeatures()`, which is unimplemented in mbgl but needed for parity with GL JS.

1ec5:
It would be possible to hack around the issue by setting each annotation/marker view’s icon to a single transparent pixel. It’s unclear to me whether handling rotation and pitch is important enough to use such a hack.

src/mbgl/map/map.cpp
```diff
@@ -760,6 +756,18 @@ std::vector<Feature> Map::queryRenderedFeatures(const ScreenBox& box, const opti
     });
 }
 
+AnnotationIDs Map::queryPointAnnotations(const ScreenBox& box) {
+    auto features = queryRenderedFeatures(box, {{ AnnotationManager::PointLayerID }});
+    AnnotationIDs ids;
+    ids.reserve(features.size());
+    std::transform(features.begin(), features.end(), std::back_inserter(ids), [](auto& feature) -> AnnotationID {
+        assert(feature.id);
+        assert(*feature.id <= UINT32_MAX);
```
jfirebaugh: `assert(*feature.id <= std::numeric_limits<AnnotationID>::max())`

src/mbgl/map/map.cpp
```diff
@@ -760,6 +756,18 @@ std::vector<Feature> Map::queryRenderedFeatures(const ScreenBox& box, const opti
     });
 }
 
+AnnotationIDs Map::queryPointAnnotations(const ScreenBox& box) {
+    auto features = queryRenderedFeatures(box, {{ AnnotationManager::PointLayerID }});
+    AnnotationIDs ids;
+    ids.reserve(features.size());
+    std::transform(features.begin(), features.end(), std::back_inserter(ids), [](auto& feature) -> AnnotationID {
```
jfirebaugh: We generally prefer a range-based for loop + `push_back` over `std::transform` + `std::back_inserter`.

src/mbgl/map/map.cpp
```diff
@@ -760,6 +756,18 @@ std::vector<Feature> Map::queryRenderedFeatures(const ScreenBox& box, const opti
...
+        assert(feature.id);
+        assert(*feature.id <= UINT32_MAX);
+        return (AnnotationID)*feature.id;
```
jfirebaugh: C++ style cast: `AnnotationID(*feature.id)`

1ec5:
This will unblock #4838, #5467, #2082, and #5787.

1ec5:
The issue in https://github.com/mapbox/mapbox-gl-native/pull/5165#issuecomment-222034113 was resolved by associating every view-backed annotation with a spacer image: #5509.

PR #5172, opened 2016-05-27
[core] Use unique_resource for GL objects

brunoabinader:
Source: https://github.com/okdshin/unique_resource
    
These replace the complexity of manually handling moveable-RAII objects with a type specific for that purpose.

As suggested in https://github.com/mapbox/mapbox-gl-native/pull/5141#issuecomment-221719872.

:eyes: @jfirebaugh @kkaefer @tmpsantos @1ec5


src/mbgl/gl/gl_object_store.hpp
```diff
@@ -3,136 +3,75 @@
...
 #include <vector>
 
+#include <functional>
+
 namespace mbgl {
 namespace gl {
 
+constexpr GLsizei TextureMax = 64;
+
+using UniqueObject = std_experimental::unique_resource<GLuint, std::function<void(GLuint)>>;
```
jfirebaugh: Rather than using `std::function` polymorphism, create a unique `*Deleter` functor and `Unique*` typedef for each resource type as sketched in https://github.com/mapbox/mapbox-gl-native/pull/5141#issuecomment-221719872. This adds additional type safety: mistaken use of the wrong resource type will be a compile time error.

src/mbgl/renderer/frame_history.hpp
```diff
@@ -31,7 +31,7 @@ class FrameHistory {
     bool firstFrame = true;
     bool changed = true;
 
-    gl::TextureHolder texture;
+    gl::UniqueObject texture = gl::UniqueObject(0, [](auto) {});
```
jfirebaugh: The default constructor should suffice here.

brunoabinader: Yes, but `unique_resource` does not have a default constructor.

brunoabinader:
Changes since last review:
- `s/gl::GLObjectStore/gl::ObjectStore/`
- Using `*Deleter` functors and `Unique*` aliases for `unique_resource` usage.
- Dealt with absence of copy assignment in `gl::ObjectStore` reference inside all `*Deleter`'s using `std::reference_wrapper`
- Dealt with absence of default constructor in `unique_resource` with `mbgl::optional` where needed (e.g. in places where the GL object is lazily created).

src/mbgl/geometry/buffer.hpp
```diff
@@ -36,11 +37,11 @@ class Buffer : private util::noncopyable {
 
     // Transfers this buffer to the GPU and binds the buffer to the GL context.
     void bind(gl::ObjectStore& store) {
-        if (buffer.created()) {
-            MBGL_CHECK_ERROR(glBindBuffer(bufferType, getID()));
+        if (buffer) {
+            MBGL_CHECK_ERROR(glBindBuffer(bufferType, (*buffer).get()));
```
jfirebaugh: Shouldn't `*buffer` work, using [the `unique_resource` implicit conversion](https://github.com/okdshin/unique_resource/blob/cba309e92ec79a95be2aa5a324a688a06af8d40a/unique_resource.hpp#L76)?

src/mbgl/gl/object_store.hpp
```diff
@@ -3,135 +3,114 @@
...
-    friend class VAOHolder;
+class ObjectStore;
 
-    std::vector<GLuint> abandonedPrograms;
-    std::vector<GLuint> abandonedShaders;
-    std::vector<GLuint> abandonedBuffers;
-    std::vector<GLuint> abandonedTextures;
-    std::vector<GLuint> abandonedVAOs;
+struct ProgramDeleter {
+    std::reference_wrapper<ObjectStore> store;
```
jfirebaugh: How about a plain `ObjectStore*` pointer instead of `std::reference_wrapper`?

src/mbgl/geometry/buffer.hpp
brunoabinader: I'd kept `.get()` to keep in sync with the other `Unique*` usage elsewhere, but your suggestion looks better, thanks!

tmpsantos:
@brunoabinader this is looking really nice. Do you think it would be possible to make utests for these holders? (not just that, but also the other wrappers we have for GL stuff)

tmpsantos:
Neat, thanks for adding the tests.

PR #5175, opened 2016-05-27
[core] Fill the initial `changeTimes` with the reference timestamp of the first frame

### LINKED ISSUE #5109: Point annotations disappear at low zoomlevels
When you zoom out, point annotations disappears. Zooming back in restores them.

ezgif com-video-to-gif 20

This was noticed on Android and was validated for iOS by @1ec5 with 3.3.0-alpha.3

tmpsantos:
On the Javascript implementation the array is automatically filled with
zeroes which is not the same thing as `TimePoint::min()`. Here we need
to get our own "zero" which is the timestamp of the first frame.

Fixes #5109

### DIFF
--- a/src/mbgl/renderer/frame_history.cpp
+++ b/src/mbgl/renderer/frame_history.cpp
@@ -4,7 +4,6 @@
 using namespace mbgl;
 
 FrameHistory::FrameHistory() {
-    changeTimes.fill(TimePoint::min());
     changeOpacities.fill(0);
     opacities.fill(0);
 };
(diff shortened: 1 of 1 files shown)


tmpsantos:
:eyes: @jfirebaugh @bleege @ansis

bleege:
Tested on Android and [confirmed it worked](https://github.com/mapbox/mapbox-gl-native/issues/5109#issuecomment-222254305).  I'm going to merge this into `master` and then cherry pick it to the `release-android-v4.1.0` branch so that Android can make use of it.

PR #5180, opened 2016-05-27
[ios] Base turnstile send time guard on calendar day

boundsj:
Previously, the SDK sent the turnstile every time the map loaded but not more than once every 24 hours. This change updates the logic so that the guard is based on the the calendar day and not a 24 hour interval. This addresses the issue where a turnstile event might be sent at 11pm on day 1, not at all on day 2 (if the event could have only been sent before 11pm on day 2), and then again on day 3. Basing off of the calendar day, the event can be sent at least once per day.

cc @mick @friedbunny @1ec5


1ec5:
👍 to basing it on the calendar day instead of a time interval.

boundsj:
Updated the original PR description to reflect what this change actually is.

PR #5183, opened 2016-05-29
Store feature keys in same order as in vector tile

### LINKED ISSUE #5182: Random rendered features have mismatched properties
Random features returned by `mbgl::Map::queryRenderedFeatures()` have mismatched or missing properties.

When the current style is Mapbox Streets (which uses Mapbox Streets source), calling `mbgl::Map::queryRenderedFeatures()` with the location of [this OSM node](http://www.openstreetmap.org/node/154075308) at z14 returns a point feature in which the `scalerank` and `type` are conflated. Here’s the output from `MGLPointFeature.attributes`:

```plist
{
    ldir = E;
    localrank = 1;
    name = Loveland;
    "name_de" = Loveland;
    "name_en" = Loveland;
    "name_es" = Loveland;
    "name_fr" = Loveland;
    "name_ru" = "\U041b\U0430\U0432\U043b\U0435\U043d\U0434";
    "name_zh" = "\U6d1b\U5f17\U5170\U5fb7";
    scalerank = town;
}
```

In GL JS v0.18.0, `queryRenderedFeatures()` doesn’t mismatch any tags, but `scalerank` is nowhere to be seen:

```js
{
  ldir: "E",
  localrank: 1,
  name: "Loveland",
  name_de: "Loveland",
  name_en: "Loveland",
  name_es: "Loveland",
  name_fr: "Loveland",
  name_ru: "Лавленд",
  name_zh: "洛弗兰德",
  type: "town"
}
```

Querying rendered features at 39°5′43″ north, 84°31′20″ west (in the 13/2172/3128 tile) yields several features in which the `layer`, `oneway`, `structure`, and `type` properties are all off by one:

```plist
{
    class = motorway;
    layer = true;
    oneway = bridge;
    structure = motorway;
    type = 2;
}
```

Compare with GL JS:

```js
{
  class: "motorway",
  layer: 2,
  oneway: "true",
  structure: "bridge",
  type: "motorway"
}
```

/cc @jfirebaugh

1ec5:
The keys in the vector tile may not be in alphabetical order. Building a vector of keys by looping over `std::map<std::string, …>` effectively sorts the keys by alphabetical order without sorting the associated values. ~~This change removes the `map` in favor of logic that [more closely matches vector-tile-js](https://github.com/mapbox/vector-tile-js/blob/91b134985381bb66fbd778720d3bbd94c3642cad/lib/vectortilelayer.js#L30).~~ This change inserts keys in the same order in which they appear in the vector tile.

Fixes #5182.

/cc @jfirebaugh

### DIFF
--- a/package.json
+++ b/package.json
@@ -20,7 +20,7 @@
     "aws-sdk": "^2.3.5",
     "express": "^4.11.1",
     "mapbox-gl-shaders": "mapbox/mapbox-gl-shaders#36278b864e60e1dba937a6863064c03d69526854",
-    "mapbox-gl-test-suite": "mapbox/mapbox-gl-test-suite#b3441d9a285ffbe9b876677acb13d7df07e5b975",
+    "mapbox-gl-test-suite": "mapbox/mapbox-gl-test-suite#f45fd7aba98650c7f3bf778c9cbbfd3b548a4ee8",
     "node-gyp": "^3.3.1",
     "request": "^2.72.0",
     "tape": "^4.5.1"
--- a/src/mbgl/tile/vector_tile.cpp
+++ b/src/mbgl/tile/vector_tile.cpp
@@ -191,7 +191,10 @@ VectorTileLayer::VectorTileLayer(protozero::pbf_reader layer_pbf) {
             features.push_back(layer_pbf.get_message());
             break;
         case 3: // keys
-            keysMap.emplace(layer_pbf.get_string(), keysMap.size());
+            {
+                auto iter = keysMap.emplace(layer_pbf.get_string(), keysMap.size());
+                keys.emplace_back(std::reference_wrapper<const std::string>(iter.first->first));
+            }
             break;
         case 4: // values
             values.emplace_back(parseValue(layer_pbf.get_message()));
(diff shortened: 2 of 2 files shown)


jfirebaugh:
The map was so that `getValue` could avoid the linear search to convert a key string to an index. vector-tile-js [doesn't create the property map lazily](https://github.com/mapbox/vector-tile-js/blob/91b134985381bb66fbd778720d3bbd94c3642cad/lib/vectortilefeature.js#L29-L37) so doesn't have an exact analog. I'm not sure how important the algorithmic approach here is. The regression was in 61d14089e0dd742719328fd74c693bcae6274a4b, and the most conservative change would be to keep the bidirectional index but properly initialize it by inserting into `keys` immediately after inserting into `keysMap`. This is safe because insertions in `map` (and `unordered_map`, should we switch) do not invalidate references.

1ec5:
> the most conservative change would be to keep the bidirectional index but properly initialize it by inserting into `keys` immediately after inserting into `keysMap`.

Thanks for the tip. I implemented this conservative change in 053b517d4b9afe98e4438720a825949ed29f8ec3.

jfirebaugh:
👍 

Can you add a query test to the test-suite that would have caught #5182?

jfirebaugh:
Looks like the test case is failing because it wants to load some tile fixtures that don't exist:

```
[Error: Not Found]
ERROR (Style): Failed to load tile 14/4356/6245=>14 for source mapbox: Not Found
ERROR (Style): Failed to load tile 14/4357/6246=>14 for source mapbox: Not Found
ERROR (Style): Failed to load tile 14/4356/6246=>14 for source mapbox: Not Found
```

PR #5185, opened 2016-05-30
Add MGLAnnotationView to jazzy TOC

1ec5:
Added MGLAnnotationView to jazzy’s table of contents.

/ref #4801
/cc @boundsj


boundsj:
👍

PR #5194, opened 2016-05-31
[ios] Add annotation container view

boundsj:
Fixes https://github.com/mapbox/mapbox-gl-native/issues/5140
Fixes https://github.com/mapbox/mapbox-gl-native/issues/5089

Adds a new UIView subclass to contain MGLAnnotationView instances and an API for helping to manage the addition of new annotations. To mitigate [known performance issues with `UIView:addSubview:`](http://blog.benjamin-encz.de/post/disassembling-uikit-tintcolor-visitor), this batches up the addition of subviews to the container _before_ adding the container itself to the view hierarchy. The organization of the views in the separate container also naturally avoids the "piercing" issue seen before between annotations and callouts.

### DIFF
--- a/platform/ios/ios.xcodeproj/project.pbxproj
+++ b/platform/ios/ios.xcodeproj/project.pbxproj
@@ -12,6 +12,9 @@
 		4018B1C91CDC288A00F666AF /* MGLAnnotationView_Private.h in Headers */ = {isa = PBXBuildFile; fileRef = 4018B1C31CDC277F00F666AF /* MGLAnnotationView_Private.h */; };
 		4018B1CA1CDC288E00F666AF /* MGLAnnotationView.h in Headers */ = {isa = PBXBuildFile; fileRef = 4018B1C51CDC277F00F666AF /* MGLAnnotationView.h */; settings = {ATTRIBUTES = (Public, ); }; };
 		4018B1CB1CDC288E00F666AF /* MGLAnnotationView.h in Headers */ = {isa = PBXBuildFile; fileRef = 4018B1C51CDC277F00F666AF /* MGLAnnotationView.h */; settings = {ATTRIBUTES = (Public, ); }; };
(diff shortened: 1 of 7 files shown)


platform/ios/ios.xcodeproj/project.pbxproj
```diff
@@ -693,6 +698,21 @@
...
+				DA8848341CBAFB8500AB86E3 /* MGLAnnotationImage.h */,
+				DA8848401CBAFB9800AB86E3 /* MGLAnnotationImage_Private.h */,
+				DA8848411CBAFB9800AB86E3 /* MGLAnnotationImage.m */,
+				DA8848421CBAFB9800AB86E3 /* MGLAPIClient.h */,
+				DA8848431CBAFB9800AB86E3 /* MGLAPIClient.m */,
+				DA8848351CBAFB8500AB86E3 /* MGLCalloutView.h */,
+				DA8848441CBAFB9800AB86E3 /* MGLCompactCalloutView.h */,
+				DA8848451CBAFB9800AB86E3 /* MGLCompactCalloutView.m */,
+				DA8848461CBAFB9800AB86E3 /* MGLLocationManager.h */,
+				DA8848471CBAFB9800AB86E3 /* MGLLocationManager.m */,
```
1ec5: This is a bad merge. I moved most of these files to the Annotations group. You should move MGLAnnotationContainerView.h/.m to that group but otherwise omit this diff.

platform/ios/src/MGLMapView.mm
```diff
@@ -2805,6 +2806,8 @@ - (void)addAnnotations:(NS_ARRAY_OF(id <MGLAnnotation>) *)annotations
 
     BOOL delegateImplementsViewForAnnotation = [self.delegate respondsToSelector:@selector(mapView:viewForAnnotation:)];
     BOOL delegateImplementsImageForPoint = [self.delegate respondsToSelector:@selector(mapView:imageForAnnotation:)];
+    
+    NSMutableArray *newAnnotationViews = [NSMutableArray array];
```
1ec5: Nit: I think you can set the capacity to the size of the annotation array.

1ec5:
Looks good, other than the bad merge in the project file.

PR #5198, opened 2016-05-31
[android] Fix x86_64 build

tmpsantos:
sysroot path not being passed to the compiler and linker


tmpsantos:
:eyes: @bleege

bleege:
Successfully tested the original in Mason --> https://github.com/mapbox/mason/pull/178

Issue #5108, opened 2016-05-23
Android TestApp Crashing When MapView Is Loaded


bleege:
After updating to the latest on `master` 25c4436d37d84cb8eb800385b412e784a0789087 and running the TestApp on a Samsung S5 (API 21) the app crashes when any Activity with a `MapView` is loaded.  The error (see stacktrace below) points to an issue with the `geojsonvt`, and not the still open issue about switching arm64 to [use `gold` instead of `bfd`](https://github.com/mapbox/mason/issues/174) as the Samsung S5 is 32 bit.  I've tried checking out the previous version of `mason` c03b19241ad6450dab16bf06466f849cb5a15757 and rebuilding with the same issue happening.  Based on that it's unclear to me if this is related to the new NDK version in #5092 .

**Steps To Reproduce**

1. Update code to 25c4436d37d84cb8eb800385b412e784a0789087
1. Update Git submodules
1. Make Android
  1.  `make clean`
  1. `make android`
1. Run TestApp on arm device
1. Choose Polygon activity (any will do it, but for consistency in debugging let's focus on this one)

```java
Caused by: java.lang.UnsatisfiedLinkError: dlopen failed: cannot locate symbol "_ZN6mapbox9geojsonvt7Convert11projectRingERKNSt6__ndk16vectorINS0_6LonLatENS2_9allocatorIS4_EEEEd" referenced by "libmapbox-gl.so"...
         at java.lang.Runtime.loadLibrary(Runtime.java:371)
         at java.lang.System.loadLibrary(System.java:989)
         at com.mapbox.mapboxsdk.maps.NativeMapView.<clinit>(NativeMapView.java:44)
         at com.mapbox.mapboxsdk.maps.MapView.initialize(MapView.java:203)
         at com.mapbox.mapboxsdk.maps.MapView.<init>(MapView.java:166)
         at java.lang.reflect.Constructor.newInstance(Native Method) 
         at java.lang.reflect.Constructor.newInstance(Constructor.java:288) 
         at android.view.LayoutInflater.createView(LayoutInflater.java:614) 
         at android.view.LayoutInflater.createViewFromTag(LayoutInflater.java:750) 
         at android.view.LayoutInflater.rInflate(LayoutInflater.java:813) 
         at android.view.LayoutInflater.inflate(LayoutInflater.java:511) 
         at android.view.LayoutInflater.inflate(LayoutInflater.java:415) 
[...]

Issue #3288, opened 2015-12-14
What is MultiPoint


tmcw:
Something seems off about this

https://github.com/mapbox/mapbox-gl-native/blob/master/android/MapboxGLAndroidSDK/src/main/java/com/mapbox/mapboxsdk/annotations/MultiPoint.java
- The docs refer to it as being a 'polyline'
- There's a MultiPoint.java, but not a Point.java
- Besides geometry, it also manages alpha, but just alpha.

What's up with this?

tmcw:
Reading into this, yeah, it's a coordinate-container abstract class that's extended by Polygon and Polyline. That shouldn't be the case, it should be named something else and more obviously documented.

tobrun:
@tmcw 
> The docs refer to it as being a 'polyline'

Can you show me what you which docs you are referring to? 

> it's a coordinate-container abstract class that's extended by Polygon and Polyline.

Where are you reading this, this is not part of the javadoc. 
I created #3338 to improve javadoc for the annotations package.

tmcw:
It's referred to as a polyline here: https://github.com/mapbox/mapbox-gl-native/blob/master/platform/android/MapboxGLAndroidSDK/src/main/java/com/mapbox/mapboxsdk/annotations/MultiPoint.java#L28

It's extended at https://github.com/mapbox/mapbox-gl-native/blob/master/platform/android/MapboxGLAndroidSDK/src/main/java/com/mapbox/mapboxsdk/annotations/Polyline.java and https://github.com/mapbox/mapbox-gl-native/blob/master/platform/android/MapboxGLAndroidSDK/src/main/java/com/mapbox/mapboxsdk/annotations/Polygon.java

There's isn't necessarily a logic problem here, but the usage of MultiPoint really conflicts with its usage elsewhere, in the Simple Features Spec, GeoJSON, and other geo standards.

tobrun:
@tmcw 
Current gameplan is to refactor annotations with the `4.0.0` release.

I'm looking into the possibility to:
 - remove Multipoint & Annotation base classes (not that logical)
 - remove Icon and Iconfactory (obsolete, not needed)
 - streamline with requirements coming from geo standards

zugaldia:
[...]

PR #5200, opened 2016-05-31
[core] Refactor core annotations API

### LINKED ISSUE #5158: mbgl::AnnotationManager should use mapbox::geometry types
`mbgl::AnnotationManager` should use the `mapbox::geometry` types that are being used e.g. as return values for `mbgl::Map::queryRenderedFeatures()` (#5066). This would give us the ability to add any kind of geometry to the annotation layer, unblocking #1729.

/cc @jfirebaugh

jfirebaugh:
* Use geometry.hpp types (fixes #5158).
* Rename annotation types based on the underlying style layer type they use: Symbol, Line, or Fill.
* Introduce a unified variant `Annotation` type; eliminate type-specific add/update methods.
* Eliminate separate array-oriented methods. They are not significantly faster than simply calling the singleton method in a loop.

* [x] @1ec5, can you please review the osx and ios changes?
* [x] @tobrun, can you please review the android changes?
* [x] @brunoabinader or @tmpsantos, can you please make the necessary qt changes? I'm not sure what the best approach is for this platform.


platform/darwin/src/MGLMultiPoint.mm
```diff
@@ -103,22 +103,6 @@ - (BOOL)intersectsOverlayBounds:(MGLCoordinateBounds)overlayBounds
...
-    
-    mbgl::AnnotationSegment segment;
-    segment.reserve(count);
-    for (NSUInteger i = 0; i < count; i++) {
-        segment.push_back(MGLLatLngFromLocationCoordinate2D(coordinates[i]));
-    }
-    return { segment };
-}
-
-- (mbgl::ShapeAnnotation::Properties)shapeAnnotationPropertiesObjectWithDelegate:(__unused id <MGLMultiPointDelegate>)delegate {
```
1ec5: Currently, if you call `-[MGLMapView visibleFeaturesAtPoint:]` at the location of a multipoint (in geometry.hpp terms), you get an MGLMultiPoint object back. MGLMultiPoint conforms to MGLAnnotation, so you can add this MGLMultiPoint object to the map using `-[MGLMapView addAnnotations:]`. With these changes, MGLMapView will end up [calling `-shapeAnnotationObjectWithDelegate:`](https://github.com/mapbox/mapbox-gl-native/blob/98b9946f36df267fa04a04dcf2e6138ded0cb428/platform/ios/src/MGLMapView.mm#L2820) on an MGLMultiPoint that has no implementation of `-shapeAnnotationObjectWithDelegate:`, resulting in an Objective-C exception.

If it isn’t straightforward to create an `mbgl::ShapeAnnotation` representing a multipoint geometry at this time, I think it’d be reasonable to implement a `-shapeAnnotationObjectWithDelegate:` that asserts for now.

jfirebaugh: `MGLMultiPoint` is trying to do too many things. It should represent the geometric "multi point" concept, and not also try to be a base class for `MGLPolygon` and `MGLPolyline`. (Our Java classes [make a similar mistake](https://github.com/mapbox/mapbox-gl-native/issues/3288).)

1ec5: Agreed. Unfortunately, I think we’re stuck with [MapKit’s concept of a multipoint](https://developer.apple.com/library/mac/documentation/MapKit/Reference/MKMultiPoint_class/) (as any shape with multiple points) for backwards compatibility. What could we call the class that represents a true multipoint? Or should we bite the bullet and bump the major version number for simplicity?

1ec5: MGLSparseMultiPoint? 😬

jfirebaugh: If I make 56989784ec77a99e47ddd2bb688767b094e7426b part of this PR, how hard would it be to refactor things so that `-annotationObjectWithDelegate:` is implemented by every class that implements `MGLAnnotation`?

1ec5: MGLAnnotation is a public protocol, so we can’t use any C++ types or expect the developer to implement this method themselves for their custom MGLAnnotation classes.

The base model class for things that get turned into `mbgl::ShapeAnnotation`s is currently MGLMultiPoint. There’s no base model class for things that get turned into `mbgl::PointAnnotation`s, because any Objective-C object can conform to the MGLAnnotation protocol and thus be added as a point annotation.

I like the idea of centralizing this shape conversion logic in one place. MGLMultiPoint is a natural place to do this, but the most flexible approach would be to implement a standalone C function similar to `MGLFeaturesFromMBGLFeatures()`.

tmpsantos:
On it.

tmpsantos:
@jfirebaugh made it so it will build, but we are not really using shape annotations for Qt ATM and we are likely to rewrite it completely anyway.

tobrun:
@jfirebaugh will review later today

jfirebaugh:
Merged the larger refactor into this PR and updated the description.

platform/darwin/src/MGLGeometry_Private.h
```diff
@@ -6,6 +6,7 @@
 #endif
 
 #import <mbgl/util/geo.hpp>
+#import <mbgl/util/geometry.hpp>
```
1ec5: This is causing the Darwin unit tests to fail to compile. Those tests need only a couple of the methods in this header. We should split this header into MGLGeometry_Private.h (private visibility, so it's visible to test target) and MGLGeometry_Internal.h (project visibility, so it's visible only to the dynamic and static targets).

1ec5: Alternatively, you can put `MGLPointFromLocationCoordinate2D()` in MGLFeature_Private.h, which already houses some geometry type conversion code.

jfirebaugh: The issue was that `geometry.hpp` was still in `src` rather than `include`.

tobrun:
@jfirebaugh 

I tried building this locally with:
```
git submodule update
git make clean
git make android-lib-arm-v8
```

I have no issues with building but at runtime I'm running into:

>Caused by: java.lang.UnsatisfiedLinkError: dlopen failed: cannot locate symbol "_ZN6mapbox9geojsonvt9GeoJSONVTC1ENSt6__ndk16vectorINS0_16ProjectedFeatureENS2_9allocatorIS4_EEEENS0_7OptionsE" referenced by "libmapbox-gl.so"...

I validated that this is not happening on master.

----

The code itself LGTM and reading the C++ was really useful/insightful for me.

LMK if I can help out or need to retest.

jfirebaugh:
@tobrun Looks to me like this crash does happen on `master` too.

jfirebaugh:
It's #5108 again. 😠

PR #5205, opened 2016-06-01
Add Google Benchmark library

kkaefer:
Similar to the Google Test framework, we should add [Google Benchmark](https://github.com/google/benchmark) to have a facility in place that we can use to benchmark performance impact of changes.


tobrun:
I haven't looked into this but this was shown during the NDK talk at Google I/O. 

screen shot 2016-06-01 at 18 08 49

Sorry if this isn't relevant for the related use-case.

kkaefer:
@tobrun These are Java based benchmarking frameworks; Google Benchmark is about C++ microbenchmarks.

tmpsantos:
@kkaefer does it make sense to run it on travis? We probably need a dedicated machine (not a VM) to do a per-commit track of performance results.

kkaefer:
@tmpsantos no; the idea is that you can use this tool locally to test changes.

PR #5207, opened 2016-06-01
initial draft of Node.js bindings memory usage test

mikemorris:
`mbgl.Map.release` and `gc()` seem to be doing the opposite of what I would expect (acquiring more memory instead of freeing)? Unless my math is backwards.

I think `mbgl.Map.load` is async (not sure if the problem in https://github.com/cutting-room-floor/tilelive-gl/issues/11#issuecomment-91659773 still exists though?), which makes the timer a bit silly.

/cc @miccolis @brunoabinader

### DIFF
--- a/package.json
+++ b/package.json
@@ -18,10 +18,13 @@
   },
   "devDependencies": {
     "aws-sdk": "^2.3.5",
+    "d3-queue": "^2.0.3",
     "express": "^4.11.1",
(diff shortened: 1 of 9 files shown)


mikemorris:
Implemented memory usage test in https://github.com/mapbox/mapbox-gl-native/pull/5529

PR #5208, opened 2016-06-01
[android] Cancel other gestures during shove.

tobrun:
From @boundsj 

> This adds a guard so that when a shove gesture is occurring, no other
gestures take place. This helps avoid the disorienting feeling of
having the map move left and right and rotate while trying to tilt it
and looks more like Google maps.

To @boundsj:
> join the dark side we have cookies

Or at least Marshmallow, Lollipop, KitKat,..


bleege:
Should this be added to the Android 4.1.0 milestone?

tobrun:
@bleege  I currently don't have the bandwith to test this throughly to include this into `4.1.0`, if we drop it into master we can iterate on it and time will tell but feel free to look into this yourself if you want to include it in next release 🚀

tobrun:
Rebased and running CI checks again

PR #5217, opened 2016-06-02
[node] Expose map.onLowMemory()

brunoabinader:
:eyes: @mikemorris @miccolis


brunoabinader:
As per conversations with @mikemorris, this hasn't proven to cause much impact in memory consumption, thus closing until further notice.

mikemorris:
@brunoabinader I might actually have to rebase and test this again to verify, as it looks like this branch contains the leak introduced in 61dad6eae59e284fdba63fcd76d9192aba900db4 and fixed in https://github.com/mapbox/mapbox-gl-native/pull/5269

brunoabinader:
Is this still revelant @mikemorris?

mikemorris:
Rebased and checked again, confirmed that this doesn't appear to do anything to reign in memory growth.

PR #5227, opened 2016-06-03
[android] #5171 - introduce new select/deselect animation View Markers

### LINKED ISSUE #5171: Make select/deselect animation more flexible on MarkerView
Currently MarkerView exposes select/deselect animations as follows:

```java
public void setSelectAnimRes(int selectAnimRes) {
        this.selectAnimRes = selectAnimRes;
}

public void setDeselectAnimRes(int deselectAnimRes) {
    this.deselectAnimRes = deselectAnimRes;
}
```

This limits end users to only providing animations through an animator created in xml. 
To be more flexible I'm proposing to provide a callback interface instead and manage a selected state inside of MarkerView. This callback will provide the developer with an instance of the View and this gives the developer more flexibility to do animations (or even do non animations things as for example hiding the MarkerView).

Side notes:
 - View reuse could be affected by this. It's possible that we need to look into the same `prepareForReuse` callback as iOS.
 - There are other issues related to selection state that should be solved to completely get a correct select/deselect user experience. Will look into identifying and linking them back.

tobrun:
closes #5171


tobrun:

PR #5231, opened 2016-06-03
[android] Fix style urls

### LINKED ISSUE #5193: Style URLs in Android SDK
With Android, we have a two ways to set the style you'd like, either in XML or java. We prefer users setting in XML to prevent loading in the default street style before changing to the one they prefer. In the past we've used constants that users can call instead of manually plugging in the default style url themselves. These constants had the style version hardcoded in them so using the constant resulted in always automatically updating the map to the latest style when the SDK was updated. Previous chats with @tobrun and @bleege, we've discussed various ways to handle removing the version number from the constants, or removing the constants all together. This resulted in deprecating all the style strings and creating java methods with the version number as the parameter. This works great for the java side of things but xml doesn't use methods and doesn't even show the style constants as deprecated. After our weekly strategy call today, we've come up with this solution:

For java we will keep the current methods for setting the map style i.e. `Style.getOutdoorsStyleUrl(9)` which gives devs flexibility when choosing which version they'd like to use. Style wont be changed without the developer updating the version parameter.  

XML will remain using the string constants i.e. `mapbox:style_url="@string/style_mapbox_streets"`. When we update the style version in the SDK, the users map style will be updated as well. This has the benefit of handling style updates on behalf of the dev. 

in both cases, the developer always has the option to pass in the full style URL themselves. The main concern here is that by keeping the constants we still keep the ease of quickly adding correct style urls quickly. If this sounds 👍 I'll self-assign and clean up the constant files. 

cc: @zugaldia @jfirebaugh

cammace:
Fixes #5193

### DIFF
--- a/platform/android/MapboxGLAndroidSDK/src/main/java/com/mapbox/mapboxsdk/constants/Style.java
+++ b/platform/android/MapboxGLAndroidSDK/src/main/java/com/mapbox/mapboxsdk/constants/Style.java
@@ -1,7 +1,9 @@
 package com.mapbox.mapboxsdk.constants;
 
 import android.support.annotation.StringDef;
+
 import com.mapbox.mapboxsdk.maps.MapView;
+
 import java.lang.annotation.Retention;
 import java.lang.annotation.RetentionPolicy;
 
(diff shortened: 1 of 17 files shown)


platform/android/MapboxGLAndroidSDK/src/main/java/com/mapbox/mapboxsdk/constants/Style.java
```diff
@@ -15,157 +17,33 @@
...
-     * </ul>
-     * <p>
-     * More information on the Mapbox styles API can be found on https://www.mapbox.com/api-documentation/#styles
-     * </p>
-     *
-     * @param version the version of the style.
-     * @return uri to load style from
+     * Indicates the parameter accepts one of the values from {@link Style}. Using one of these
+     * constants means your map style will always use the latest version and may change as we
+     * improve the style
```
bleege: I love this documentation.  Is there anyway we could make it bolder though via fonts?  We should strive to make sure there is NO DOUBT that things will change on them.

platform/android/MapboxGLAndroidSDK/src/main/res/values/strings.xml
```diff
@@ -11,11 +11,14 @@
     <string name="infoWindowAddress">Address</string>
 
     <!-- these are public -->
-    <!-- {@deprecated Use Style.getXStyleUrl(int version) instead.} -->
+    <!-- Using one of these constants means your map style will always use the latest version and
+     may change as we improve the style. -->
```
bleege: Same here about being super clear that this means Style will change.  Maybe we should put some language in both the Java and XML side about these being "Convenience" Style URLs, and that the preferred route is to use regular Style URLs

tobrun:
Travis is failing with `The command "curl -sSL "http://llvm.org/apt/llvm-snapshot.gpg.key" | sudo -E apt-key add -" failed and exited with 2 during .` This is an problem with reaching llvm. I double checked that this PR only contains java code that will not have any impact on core or any other OS'es. From my end good to merge here.

jfirebaugh:
@tobrun That was fixed with 83ba835168858ba5cdcef0f4d6fdde8624a11e9f on the main branch. You may want to cherry pick that to the release branch.

tobrun:
Done with 4094987fa09561c4c69b1b4cf8a6cbf9d3dd0433, PR already merged with 32a3d98be041f6ecd31af80570f78e802887225b. closing

Issue #5229, opened 2016-06-03
How to run the demo app?


YoungxHelsinki:
**Platform: iOS**
**Mapbox SDK version: `commit 623da8bfeb7adcbf1774a14c9a0dc46469891cef`**

Hi all,

I would like to contribute to Mapbox ios SDK and I tried to run the demo app inside
the mapbox-gl-native. How I ever I got multiple errors. The first one was about 
SMCalloutView.m/h missing. I figured that the SMCalloutView directory was linked 
to [other repo](https://github.com/nfarina/calloutview/tree/2aede5d8d1577101bf18405246220e7a710df607). 
I copied the corresponding files to the "right" path, i.e., 
`/mapbox-gl-native/platform/ios/vendor/SMCalloutView/`

Then I got another error:
`The file “mbgl.xcconfig” couldn’t be opened because there is no such file. (/mapbox-gl-native/build/ios/mbgl.xcconfig)`
![screen shot 2016-06-03 at 17 24 31](https://cloud.githubusercontent.com/assets/5738363/15781744/3653e5c4-29b0-11e6-9478-553ebfa99b99.png)
This time I couldn't get any clue about what that file is.

### In short:
1. How can I resolve all those red things in my workspace?
2. Why are they deliberately missing from the repo?

Thanks.

friedbunny:
Thanks for digging in! Our [developing instructions](https://github.com/mapbox/mapbox-gl-native/blob/e9dcfa0ed627ecb8783d1ac7f38a24db51cf8cb8/platform/ios/DEVELOPING.md) appear to be lacking an important setup step:

```
git submodule init
git submodule update
```

This will get you SMCalloutView and KIF, which you shouldn’t install manually. I’ll add this omitted step to our instructions.

As for the other red files, these are build products that will be created when you build the various app and test targets. The `iosapp` target is the demo app that you’ll likely be most interested in.

YoungxHelsinki:
Thanks @friedbunny for the reply. Actually I did learn about the submodule thing on irc channel but it didn't solve the issue. I still get this error:
`/mapbox-gl-native/include/mbgl/util/geo.hpp:6:10: 'mapbox/geometry/point.hpp' file not found`
![screen shot 2016-06-03 at 18 16 50](https://cloud.githubusercontent.com/assets/5738363/15783618/9392a52a-29b7-11e6-9463-33e5705df9c2.png)

jfirebaugh:
[...]

PR #5233, opened 2016-06-03
[ios] Reorganize and update development instructions

friedbunny:
Prompted by #5229.

- Add subsections to the “Building” section.
- Add more detail to `make` packaging commands and emphasize that Debug
  is the default build type.
- ~~Manually checking out git submodules is no longer necessary.~~ Removed in #5232.

/cc @1ec5 @jfirebaugh

### DIFF
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -54,3 +54,7 @@ See the relevant SDK documentation for next steps:
 * [Mapbox Qt SDK](platform/qt/)
 * [Mapbox GL Native on Linux](platform/linux/)
 * [node-mapbox-gl-native](platform/node/)
+
+## 4: Keeping up to date
+
+This repository uses Git submodules, which should be automatically checked out when you first run a `make` command for one of the above platforms. These submodules are not updated automatically and we recommended that you run `git submodule update` after pulling down new commits to this repository.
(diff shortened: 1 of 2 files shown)


1ec5:
I just updated a checkout from 64f4be635bc83d0925018a344404f9f387ba482d to 7a43f8bd1eb0df0d519d48d81996e02691a02817, and running either `make xproj` or `make iproj` got me this:

```
$ make iproj 
./configure platform/ios/scripts/configure.sh build/ios/config.gypi ios
* Using Python 2.7.11
bash: /Users/mxn/hub/mapbox-gl-native-2/.mason/scripts/unique_resource/dev/script.sh: No such file or directory
make: *** [build/ios/config.gypi] Error 127
```

Meanwhile, the .mason submodule is still on mapbox/mason@34cb8a83904d873dfcee6e3543c7527e83816d43.

jfirebaugh:
@1ec5 The build scripts handle initializing the submodules, but not updating them. If you fetch or pull changes that include a submodule update, you'll need to ensure that happens. If you can find a way to have the build scripts do this automatically when needed, I'd definitely welcome that.

friedbunny:
@jfirebaugh also put in some edits to these docs at #5232, which took care of the now-unnecessary submodule commands. Rebased this PR onto that.

friedbunny:
@1ec5 So we should add a note about remembering to `git submodule update` when you pull down new changes.

friedbunny:
Added a section to the main `INSTALL.md` about updating git submodules.

PR #5235, opened 2016-06-03
Make annotation views accessible to VoiceOver

### LINKED ISSUE #5038: Annotation views are inaccessible to VoiceOver
As of #4801, annotations can be presented as either GL annotations or annotation views. Whereas MGLMapView synthesizes an accessibility element for each GL annotation, it doesn’t do so for annotation views. MGLMapView should include annotation views as they are in its child accessibility elements.

/cc @boundsj

1ec5:
Corrected the accessibility frames of annotation views to match the views’ bounds instead of the bounds of the default annotation image (which could be empty if no default GL point annotation has been added).

~~Ideally, we’d figure out how to make the annotation view its own accessibility element, so that it or a subclass can specify a custom frame or Bézier path. But I’m punting on that for now to keep things simple.~~ Annotation views are now their own accessibility elements. A subclass can define a custom accessibility path.

Fixes #5038.

/cc @boundsj @friedbunny

### DIFF
--- a/platform/ios/src/MGLAnnotationContainerView.m
+++ b/platform/ios/src/MGLAnnotationContainerView.m
@@ -35,4 +35,18 @@
     }
 }
 
+#pragma mark UIAccessibility methods
+
+- (UIAccessibilityTraits)accessibilityTraits {
+    return UIAccessibilityTraitAdjustable;
+}
+
+- (void)accessibilityIncrement {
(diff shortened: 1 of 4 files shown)


boundsj:
👍

1ec5:
I’ve redone the PR to use annotation views themselves as their own accessibility elements.

friedbunny:
👍

Issue #5218, opened 2016-06-02
Flat markers should have screen-centered vanishing point


1ec5:
As seen in https://github.com/mapbox/mapbox-gl-native/issues/3276#issuecomment-223452965, the vanishing point for a flat marker view on a tilted map is always somewhere directly above the marker view. So in the screenshot above, the flag to the right is symmetrical across the y axis, whereas you’d expect both the left and right sides of the flag to parallel the north-south streets they’re being drawn atop:

![device-2016-06-02-183106](https://cloud.githubusercontent.com/assets/989928/15764390/339ee61e-28f0-11e6-98d5-4838d101579b.png)

If the perspective transformation is executed correctly, you shouldn’t be able to tell that the flag is a marker view. As you pan the map from side to side, the streets shouldn’t appear to be moving relative to the flag. In graphics terms, the vanishing point for the transform should be horizontally centered; it shouldn’t vary by the x position of the marker view.

Due to #5090, we haven’t run into this issue on iOS yet, but it’s something to be aware of.

/cc @tobrun @bleege @boundsj @friedbunny

tobrun:
Thank for the 👀  @1ec5, getting this in will probably require the addition of a container view since we need to be able to control the `onDraw` and do the projection ourselves. (instead of relying on setRotationX of the user provided view).

tobrun:
with #9782 we have deprecated MarkerView, closing.

Issue #5090, opened 2016-05-21
When map is tilted, flat annotation views should be skewed


1ec5:
Both the user dot/puck and the flat annotation views introduced in #5067 rely on `CATransform3DRotate()` to rotate the annotation view in 3D space. But the transform is based on `CATransform3DIdentity`, so the annotations appear to be squashed, whereas the map is skewed with a perspective transform:

![untilted](https://cloud.githubusercontent.com/assets/1231218/15446485/70eb6ace-1ed7-11e6-90b5-399abfebd9ee.png)
&nbsp;&nbsp;&nbsp;&nbsp;_Untilted_

![tilted](https://cloud.githubusercontent.com/assets/1231218/15446486/792f066e-1ed7-11e6-8cac-46fb4095eb60.png)
&nbsp;&nbsp;&nbsp;&nbsp;_Tilted_

Notice how the left and right edges of each annotation remain perpendicular to the top and bottom edges, even as the north-south roads come closer together towards the top of the view. (This is the same area depicted in #1266, by the way.)

/cc @boundsj @friedbunny

1ec5:
http://stackoverflow.com/questions/6887786/understanding-catransform3d
https://developer.apple.com/library/mac/documentation/Cocoa/Conceptual/CoreAnimation_guide/AdvancedAnimationTricks/AdvancedAnimationTricks.html#//apple_ref/doc/uid/TP40004514-CH8-SW13

1ec5:
#5245 isn’t going to make iOS SDK v3.3.0, so we need to remove the `flat` property from MGLAnnotationView, because that property doesn’t produce the intended, documented effect. It isn’t such a big deal with the user dot/puck, which is round and relatively small, so we won’t touch that for v3.3.0. We’ll leave this bug open to track the real fix in v3.4.0.

1ec5:
#5546 backs out the `flat` property, pending #5245.

1ec5:
The commits above are from #6686.

1ec5:
#5546 removed the flat annotation view feature, but we’re continuing to track a correct implementation of this feature in #10498.

Issue #5169, opened 2016-05-27
When map is tilted, Marker Views don’t scale by distance from camera


tobrun:
Android equivalent of #5040, capturing from that issue:

> When the map is tilted, an GL icon or point annotation grows as the camera moves toward it and shrinks as the camera moves away from it (#2186). However, annotation views maintain their size. This behavior is consistent with the user dot, but scaling would help by reinforcing the illusion of a tilted map.

Also capturing that scaling should be default to follow the style spec:

> Given that the consensus in mapbox/mapbox-gl-js#2541 was that scaling by default was better, I suspect it will always be the style spec default.

dylan-chong:
Any progress on this for Android? It looks like the iOS has been updated a while ago  https://github.com/mapbox/mapbox-gl-native/pull/5085

tobrun:
MarkerView is going to be deprecated with https://github.com/mapbox/mapbox-gl-native/pull/9782. We are no longer maintaining this API and will remove it in future versions of the SDK. As noted in the deprecation message. We are advising users to leverage SymbolLayer instead (you can use the sample code in #9782 to convert Android SDK views to an Image to be used as a Symbol). While we no longer maintain the API we are happy to accept any PRs on it.

Issue #3668, opened 2016-01-22
[Question] Is this possible to match the MGL view point in another opengl/webgl view? 


dagatsoin:
This is not an issue or maybe a lack of information about the rendering design. (Or I miss some docs?)

My goal is to put 3D models on the map. I know it is not possible to draw 3D objects for now in Mapbox.

I plan to use another view on top of the MGLView. I will start with a cordova wkwebview + webgl canvas for cross plateform concern.

This 3D canvas will have a camera that I need to match to the MGLView.

I would like to know what should be a good starting point to do this with mapboxgl design in mind.

jfirebaugh:
Unfortunately this isn't something that's possible with the iOS SDK at this time.

1ec5:
To clarify, it isn't possible to place 3D models directly in the map, to have the Mapbox iOS SDK render it for you. If you wanted to synchronize some other view with the map view, that would be a pretty manual affair, but take a look at the documentation for MGLMapCamera as well as the geometric conversion methods in MGLMapView; maybe you could piece it together with those APIs.

dagatsoin:
@1ec5 It is exactly my point. I am aware that mapboxgl can't render 3D objects. It is for this reason that I need an interface to retrieve the global space and camera information to reproduce a 3D overlay with another view (wkwebview + tree.js or native gl View). Could you plz leave this issue opened until I figure it out?

dagatsoin:
Are you talking about the doc in the source code or somewhere else?

1ec5:
Yes, I’m referring to the documentation in the source code, which is also [available online](https://www.mapbox.com/ios-sdk/api/3.1.2/). Specifically, take a look at:

* [`-[MGLMapViewDelegate mapViewRegionIsChanging:]`](https://www.mapbox.com/ios-sdk/api/3.1.2/Protocols/MGLMapViewDelegate.html#/c:objc%28pl%29MGLMapViewDelegate%28im%29mapViewRegionIsChanging:)
* [`-[MGLMapView camera]`](https://www.mapbox.com/ios-sdk/api/3.1.2/Classes/MGLMapView.html#/c:objc%28cs%29MGLMapView%28py%29camera)
* [MGLMapCamera Class Reference](https://www.mapbox.com/ios-sdk/api/3.1.2/Classes/MGLMapCamera.html)

dagatsoin:
Thx you, this is a good starting point.

suzukieng:
[...]

PR #5245, opened 2016-06-06
[WIP] Configurable degrees of freedom for annotation views

### LINKED ISSUE #991: allow for annotation selection z-order change
Piggybacking off of https://github.com/mapbox/mapbox-gl-native/issues/988, when you mark an annotation as selected, it should come to the front of the z-order. This is how MapKit behaves. For state tracking, we can do that per-platform (such as determining whether single or multiple selection is allowed), but we need to be able to change individual ordering on demand for the visual component. 

Need to see how difficult this might be in the shaders in #988 first.

### LINKED ISSUE #2528: Align annotation rotation and pitch with map
Annotations maintain their perspective even as you rotate and pitch the map. But sometimes you want an annotation to stay fixed, so that it rotates and tilts with the map. A contrived example would be an arrow that points in a particular direction – you’d want the direction to be in sync with the map’s direction.

/cc @incanus @ansis

### LINKED ISSUE #5230: MGLAnnotationView not update location immediately on setCoordinate:
**Platform:** iOS
**Mapbox SDK version:** 3.3 alpha-3

### Steps to trigger behavior

 1. [annotation setCoordinate:CLLocationCoordinate2DMake(...,...)];

### Expected behavior
Updating position of MGLAnnotationView on map view

### Actual behavior
Position changing only if map region changed, other annotations added or other manipulations with map

1ec5:
Replaced MGLAnnotationView’s `flat` property with a `freeAxes` property that allows 0–2 degrees of freedom (pitch and/or rotation) for various “billboard” effects. (Fixes #2528.) Currently, y-free annotation views are rendered incorrectly when rotated and tilted. Other than that issue and #5090, most of the transforms are working:

| Rotated | Tilted | Neither free | x free | y free | Both free |
| --- | --- | --- | --- | --- | --- |
| ✖️ | ✖️ | none unrotated untilted | x unrotated untilted | y unrotated untilted | xy unrotated untilted |
| ✔️ | ✖️ | none rotated untilted | x rotated untilted | y rotated untilted | xy rotated untilted |
| ✖️ | ✔️ | none unrotated tilted | x unrotated tilted | y unrotated tilted | xy unrotated tilted |
| ✔️ | ✔️ | none rotated tilted | x rotated tilted | y rotated tilted | xy rotated tilted |

This PR also applies perspective to x-free annotation views. (#5090) The effect is equivalent to #5218. The fix will require exposing the coordinate-to-point transformation matrix in `mbgl::TransformState` so that the SDK can convert it into a `CATransform3D`.

At times, I’m seeing an orange screen of death when there are a thousand annotations and the views start getting recycled:

orange

To do:
- [ ] Plumb projection matrix through to MGLAnnotationView
- [ ] Fix orange screen bug seen in iosapp with 1,000 point annotations
- [ ] Fix y-free annotation views when tilted and rotated

This PR previously contained the following changes that have been split out:
- ~~`-[MGLMapView addAnnotations:]` ignores multipoints (other than polylines and polygons) like it ignores multipolylines, multipolygons, and shape collections.~~ Split out into #5262.
- ~~Fixed an issue that caused a default annotation icon to appear beneath an annotation view when relocating the annotation.~~ Split out into #5263.
- ~~Bring annotation view to the front when selected. (Fixes #991.)~~ Split out into #5264.
- ~~Reformatted and copyedited MGLAnnotationView documentation. Removed the unnecessary custom getter on scalesWithViewingDistance.~~ Split out into #5545.
- ~~Fixed a bug causing a flat annotation view’s pitch to be reset to 0° when setting the view’s center offset.~~ Split out into #5550.
- ~~Made annotation views’ positions animatable. In iosapp, tapping the callout view of a D.C. fire hydrant annotation causes the annotation view to slide to the center of the screen; the view still moves instantaneously with the map as you manipulate it with gestures. (#5230)~~ Split out into #5550.
- ~~In iosapp, eliminated the center view of MBXAnnotationView in favor of applying a background color and border to the annotation view’s layer.~~ Split out into #5550.

/cc @boundsj @friedbunny

### DIFF
--- a/include/mbgl/map/map.hpp
+++ b/include/mbgl/map/map.hpp
@@ -30,6 +30,8 @@ class Image;
 class Style;
 } // namespace style
 
+using mat4 = std::array<double, 16>;
+
 class Map : private util::noncopyable {
 public:
     explicit Map(Backend&,
@@ -139,6 +141,7 @@ public:
     // Projection
     ScreenCoordinate pixelForLatLng(const LatLng&) const;
     LatLng latLngForPixel(const ScreenCoordinate&) const;
+    void getProjMatrix(mat4& matrix, uint16_t nearZ = 1) const;
 
     // Annotations
     void addAnnotationImage(std::unique_ptr<style::Image>);
--- a/platform/ios/Mapbox.playground/Contents.swift
+++ b/platform/ios/Mapbox.playground/Contents.swift
@@ -118,7 +118,7 @@ class MapDelegate: NSObject, MGLMapViewDelegate {
             let av = PlaygroundAnnotationView(reuseIdentifier: "annotation")
             av.frame = CGRect(x: 0, y: 0, width: 30, height: 30)
             av.centerOffset = CGVector(dx: -15, dy: -15)
-            av.flat = true
+            av.freeAxes = .X;
             let centerView = UIView(frame: CGRectInset(av.bounds, 3, 3))
             centerView.backgroundColor = UIColor.whiteColor()
             av.addSubview(centerView)
--- a/platform/ios/Mapbox.playground/Contents.swift
+++ b/platform/ios/Mapbox.playground/Contents.swift
@@ -118,6 +118,7 @@ class MapDelegate: NSObject, MGLMapViewDelegate {
             let av = PlaygroundAnnotationView(reuseIdentifier: "annotation")
             av.frame = CGRect(x: 0, y: 0, width: 30, height: 30)
             av.centerOffset = CGVector(dx: -15, dy: -15)
+            av.flat = true
             let centerView = UIView(frame: CGRectInset(av.bounds, 3, 3))
             centerView.backgroundColor = UIColor.whiteColor()
             av.addSubview(centerView)
--- a/platform/ios/app/MBXAnnotationView.m
+++ b/platform/ios/app/MBXAnnotationView.m
@@ -17,8 +17,8 @@
 {
     [super setSelected:selected animated:animated];
 
-    self.layer.borderColor = selected ? [UIColor blackColor].CGColor : [UIColor whiteColor].CGColor;
-    self.layer.borderWidth = selected ? 2.0 : 0;
+    self.layer.borderColor = selected ? [UIColor blackColor].CGColor : [UIColor blueColor].CGColor;
+    self.layer.borderWidth = selected ? 2.0 : 1.0;
 }
 
 - (void)setDragState:(MGLAnnotationViewDragState)dragState animated:(BOOL)animated
--- a/platform/ios/app/MBXViewController.m
+++ b/platform/ios/app/MBXViewController.m
@@ -1720,6 +1720,10 @@ typedef NS_ENUM(NSInteger, MBXSettingsMiscellaneousRows) {
         // method in this class to make draggable annotation views play nice.
         annotationView.draggable = YES;
 
+        // uncomment to flatten the annotation view against the map when the map is tilted
+        // this currently causes severe performance issues when more than 2k annotations are visible
+        // annotationView.flat = YES;
+        
         // Uncomment to force annotation view to maintain a constant size when
         // the map is tilted. By default, annotation views will shrink and grow
         // as they move towards and away from the horizon. Relatedly, annotations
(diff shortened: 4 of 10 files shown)


1ec5:
Based on [this guide](https://gist.github.com/JeOam/94e833bcefd738d805cc#the-vanishing-point), I attempted to also address the iOS side of #5218 by translating to the view center before rotating, but that caused the annotations to drift with respect to the map when panning from side to side.

At this point, I’m eyeing a solution in which `mbgl::Map` exposes `TransformState`’s projection matrix or `coordinatePointMatrix` to the SDK. It should be possible to convert from that matrix to a CATransform3D that can be applied to the annotation views’ layers, although some [additional involved steps](http://stackoverflow.com/a/10952866/4585461) may be required.

/cc @jfirebaugh @tobrun

tobrun:
@1ec5 That looks like a great approach to me. While I feel the same when it comes to holding back on tilt. I don't see an issue with rotate or scale. These are currently implemented without side effects and can be ported to use the solution mentioned above later on. Side note for this is that a lift is needed to start implementing the `CATransform3D` alternative on Android for View Annotations (Camera graphics object from the Android SDK as we already use on MyLocationView).

1ec5:
> now I'm thinking this could also be related to impact of scale of view annotations when zooming out in #5169 (which is not implemented on android yet)

Correct. Scaling based on viewing distance is implemented in the iOS SDK, but the effect is suboptimal. For one thing, all annotations appear to shrink as the pitch increases, because they're all sizes based on the annotations farthest back rather than at the center of the view (where there is supposed to be no distortion). Moreover, I don't feel good about relying on magic values throughout these transform calculations. Without basing the transforms on mbgl's matrix, it'll be difficult to make a view-backed annotation behave like a GL annotation when it comes to scaling.

1ec5:
I split out and merged the less risky parts of this PR as #5262, #5263, #5264. The remaining changes either need further work to produce the desired effect or need us to better understand the performance implications.

1ec5:
The naming and design of the `freeAxes` property and `MGLAnnotationViewBillboardAxis` enum was based on SceneKit’s `SCNBillboardConstraint.freeAxes` and `SCNBillboardAxis`. I thought it was an elegant approach at first, but it has been difficult to keep the different axes straight: which axis affects whether the view tilts with the map? Does it tilt when `MGLAnnotationViewBillboardAxisX` is on or off? On top of everything else, MGLMapView doesn’t support banking the camera, so `MGLAnnotationViewBillboardAxisZ` ends up being meaningless.

So I’m thinking of restoring the original API with more descriptive names, like `tiltsToMatchCamera` and `rotatesToMatchCamera` (with `UITextField.adjustsFontSizeToFitWidth` as precedent for such a long name).

1ec5:
> At this point, I’m eyeing a solution in which mbgl::Map exposes TransformState’s projection matrix or coordinatePointMatrix to the SDK. It should be possible to convert from that matrix to a CATransform3D that can be applied to the annotation views’ layers, although some additional involved steps may be required.

The SceneKit integration described in https://github.com/mapbox/mapbox-gl-native/issues/3668#issuecomment-286742893 would benefit from access to the map's transform matrix as well. For that particular example (which is awesome!), an orthographic projection may be desirable, but something more realistic might require a perspective transform.

/cc @suzukieng

1ec5:
peterqliu/threebox#12 looks like it addresses the same use case as this PR. In particular, it [synthesizes a projection matrix](https://github.com/peterqliu/threebox/pull/12/files#diff-f5dcf9e56e693f30034246d3d043f7e3R45) that matches the underlying map. If this matrix manages to account for horizontal translation on a tilted plane, that would be a good candidate to be ported here.

/cc @kronick

1ec5:
Transforming the annotation container view instead of individual annotation views ensures that each annotation view appears to match the underlying map’s transform. However, this approach redundantly applies the map’s perspective matrix to the annotations’ positions, causing them to appear at the wrong locations. Additionally, it prevents individual annotation views from opting out of lying flat against the map.

flattened hydrants

Along the lines of https://github.com/mapbox/mapbox-gl-native/pull/5245#issuecomment-223891387, a more correct approach would be to apply the perspective rotation transform to each individual annotation view, but the anchor point (vanishing point) would need to be moved to the center of the container view to avoid this effect:

uncentered hydrants

jfirebaugh:
Stale.

1ec5:
Tracking a revival of this PR in #10498.

PR #5246, opened 2016-06-06
Split out MGLMapViewDelegate into separate header

### LINKED ISSUE #1385: refactor MGLMapViewDelegate into its own header

1ec5:
Reprise of #1620, but without any documentation changes. This change makes the iOS and OS X versions of MGLMapViewDelegate a little more consistent with one another.

Fixes #1385.

/cc @friedbunny


friedbunny:
🎉

PR #5250, opened 2016-06-06
5076 optimise view reuse cache

### LINKED ISSUE #5076: Optimise view reuse cache with Marker Views
Marker view integration on Android landed with #3276.

This integration is backed up by an adapter approach that uses view recycling. Currently there is no limit on the amount of Views in the cache (child views of MapView) and there is a behaviour when panning in and out the viewport that results in growing the cache while it should be reusing the items in the cache.

**Example**:
We show 100 view marker but panning quickly a couple of times results in a cache with 147 views.

device-2016-05-20-132241

tobrun:
closes #5076


tobrun:

PR #5253, opened 2016-06-06
Enum naming convention

kkaefer:
map.cpp uses lower case enum values, but we're using TitleCase values in the rest of the codebase.


tmpsantos:
:+1:

PR #5257, opened 2016-06-06
Move OpenGL code away from Buckets

kkaefer:
We currently have some OpenGL rendering code in the buckets, but the buckets shouldn't be part of the renderer (they're part of the TileData object), so I'd like to separate them.


kkaefer:
Superseded by #6468

PR #2336, opened 2016-06-06
TexturePool should use RAII

brunoabinader:
TexturePool is a very risky class at the moment. To make it safe:

* The `getTextureID` method should return an RAII object that releases the texture back to the pool on destruction.
* `TexturePool` should release all texture resources in its destructor. Before doing so, it should assert that there are no active acquisitions -- the pool itself must outlive the RAII'ed acquisitions.

### DIFF
--- a/src/mbgl/gl/object_store.hpp
+++ b/src/mbgl/gl/object_store.hpp
@@ -20,27 +20,27 @@ class ObjectStore;
 
 struct ProgramDeleter {
     ObjectStore* store;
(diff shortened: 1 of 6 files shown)


jfirebaugh:
Nit: `PooledTexture` would be a better name for the resource. In the domain of ownership objects, the term "shared" usually means that the object implements shared-ownership semantics (e.g. `shared_ptr`).

PR #5266, opened 2016-06-07
Rename VectorTileData => GeometryTileData

kkaefer:
We're using the term "vector tile" to refer to the encoded vector tiles and "geometry tile" to the parsed representation that is made up of geometries. This aligns the naming in our code base.


kkaefer:
Will be merged as part of #5143

PR #5269, opened 2016-06-07
[core] Fix memory leak in Raster image data

brunoabinader:
Revert 61dad6eae59e284fdba63fcd76d9192aba900db4. No one owned the data after release(), so this wasn't being free'd. `reset()` does that, thus this patch causes a huge memory leak.

:eyes: @tmpsantos @kkaefer

### DIFF
--- a/include/mbgl/util/image.hpp
+++ b/include/mbgl/util/image.hpp
@@ -1,5 +1,7 @@
 #pragma once
 
+#include <mbgl/util/noncopyable.hpp>
+
 #include <string>
 #include <memory>
 #include <algorithm>
@@ -12,7 +14,7 @@ enum ImageAlphaMode {
 };
 
 template <ImageAlphaMode Mode>
-class Image {
+class Image : private util::noncopyable {
 public:
     Image() = default;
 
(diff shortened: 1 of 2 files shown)


include/mbgl/util/image.hpp
```diff
@@ -26,6 +28,18 @@ class Image {
           height(h),
           data(std::move(data_)) {}
 
+    Image(Image&& o)
+        : width(o.width),
+          height(o.height),
+          data(std::move(o.data)) {}
+
+    Image& operator=(Image&& o) {
```
tmpsantos: Not in use.

brunoabinader: There's actually some places where the default copy operator (now deleted) was in use:

https://github.com/mapbox/mapbox-gl-native/blob/master/src/mbgl/sprite/sprite_parser.cpp#L96

jfirebaugh: That was using the default move operator. Objects with non-copyable members such as `std::unique_ptr` don't have a default copy operator.

I don't think any of the changes to image.hpp here were necessary. What was the rationale here?

brunoabinader: >  What was the rationale here?

Like other `mbgl::utill::noncoypable` classes, I think we want to prevent copying an `Image` by design.

jfirebaugh: It was already non-copyable. And the added move constructor and move assignment operator are what the compiler would provide automatically anyway. (You should almost never write an explicit move constructor or move assignment operator.)

Issue #5036, opened 2016-05-14
Draggable annotation views


1ec5:
Like MKAnnotationView in MapKit, MGLAnnotationView should have a `draggable` property that, when set to `YES`, allows the user the relocate the annotation by long-pressing on it and dragging it. During the dragging gesture, the map remains still. (MKPinAnnotationView additionally levitates slightly above the map while being dragged and animates the beginning and end of the gesture.)

The Android SDK equivalent of this feature is being tracked in #2450.

/cc @boundsj

1ec5:
Until this feature is implemented, the workaround is to keep track of the annotation yourself; when the user long-presses at a given location, check if the gesture took place at the annotation’s coordinates. If it did, set the map view’s `scrollEnabled` to `NO` and install your own pan gesture recognizer that changes the annotation’s coordinates.

frederoni:
Just pushed a branch that seems to be working well. It's still WiP as I'm not aware of all the implications.
I'm hesitating about if the view should handle updating the position or if that should be delegated back to the implementer through `MGLMapViewDelegate` or `MGLAnnotationViewDelegate` (`annotationViewWasDraggedToLocation:`)

I do get the concerns about using two gesture recognizers per view but I don't see any performance decrease comparing 1000 draggable vs 1000 non-draggable annotation views.

Not a fan of updating the `mapView` property on the annotationView in `MGLMapView -updateAnnotationViews` so let me know if you have any good ideas how to get around that.

Issue #5130, opened 2016-05-24
Add annotationView to mapView's didSelectAnnotation and didDeselectAnnotation protocol methods


boundsj:
MapKit passes back the annotation view that was (de)selected in `- mapView:did(De)SelectAnnotationView`. To maintain backwards compatibility with annotation views not backed by an MGLAnnotationView, the Mapbox API currently (and for now must continue to) pass the annotation object itself as in: `- mapView:did(De)SelectAnnotation:`. However, the Mapbox API can gain the `annotationView` for a fully formed `- mapView:did(De)SelectAnnotation:annotationView:`. The annotationView parameter can be nil when it does not make sense to send in (in the context of a non-MGLAnnotationView backed annotation).

This will make it easier for application developers to interact with views that are involved with selection and deselection.

1ec5:
`- mapView:did(De)SelectAnnotationView:` would be more elegant once #5129 is implemented.

Issue #5129, opened 2016-05-24
Add annotation property to MGLAnnotationView


boundsj:
MapKit as an [`annotation` property](https://developer.apple.com/library/ios/documentation/MapKit/Reference/MKAnnotationView_Class/#//apple_ref/occ/instp/MKAnnotationView/annotation) that represents the annotation object currently associated with the view. For application developers' convenience MGLAnnotationView could have a similar property.

Issue #5128, opened 2016-05-24
Add selected property to MGLAnnotationView


boundsj:
MapKit has a [`selected` property](https://developer.apple.com/library/ios/documentation/MapKit/Reference/MKAnnotationView_Class/#//apple_ref/occ/instp/MKAnnotationView/selected) on the annotation view base class. This property is used to indicate if a callout is currently shown for a particular annotation view. MGLAnnotationView should have a similar property.

PR #5272, opened 2016-06-07
[ios] [WiP] #5036 draggable annotation views

frederoni:
This PR is still WiP or even PoC as I'm trying to figure out all implications of it and aligning it with [#5245](https://github.com/mapbox/mapbox-gl-native/pull/5245)

**TODO**
- [ ] Figure out if this solution will work at all.
- [ ] Applying transformations is being omitted while dragging which could lead to unexpected behaviour if `freeAxes`/`flat` is set.
- [ ] Current animation only scales the annotation view, but the most common use case will likely be a pin annotation being lifted. Add this functionality and demo in `MBXViewController`

***TBC***

/cc @friedbunny @1ec5 @boundsj

### DIFF
--- a/platform/darwin/src/MGLAnnotation.h
+++ b/platform/darwin/src/MGLAnnotation.h
@@ -27,7 +27,7 @@ NS_ASSUME_NONNULL_BEGIN
  The center point (specified as a map coordinate) of the annotation. (required)
  (read-only)
  */
-@property (nonatomic, readonly) CLLocationCoordinate2D coordinate;
+@property (nonatomic, readwrite) CLLocationCoordinate2D coordinate;
 
 @optional
 
--- a/platform/ios/src/MGLAnnotationView.h
+++ b/platform/ios/src/MGLAnnotationView.h
@@ -4,6 +4,14 @@
 
 NS_ASSUME_NONNULL_BEGIN
 
+typedef NS_ENUM(NSUInteger, MGLAnnotationViewDragState) {
+    MGLAnnotationViewDragStateNone = 0,     // View is sitting on the map
+    MGLAnnotationViewDragStateStarting,     // View is beginning to drag and animation starts
+    MGLAnnotationViewDragStateDragging,     // View is being dragged
+    MGLAnnotationViewDragStateCanceling,    // View dragging was cancelled and will be returned to its starting positon.
+    MGLAnnotationViewDragStateEnding        // View was dragged, new coordinate has been set and stop animation is finishing
+};
+
 /** The MGLAnnotationView class is responsible for representing point-based annotation markers as a view. Annotation views represent an annotation object, which is an object that corresponds to the MGLAnnotation protocol. When an annotation’s coordinate point is visible on the map view, the map view delegate is asked to provide a corresponding annotation view. If an annotation view is created with a reuse identifier, the map view may recycle the view when it goes offscreen. */
 @interface MGLAnnotationView : UIView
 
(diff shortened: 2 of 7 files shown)


platform/darwin/src/MGLAnnotation.h
```diff
@@ -27,7 +27,7 @@ NS_ASSUME_NONNULL_BEGIN
  The center point (specified as a map coordinate) of the annotation. (required)
  (read-only)
  */
-@property (nonatomic, readonly) CLLocationCoordinate2D coordinate;
+@property (nonatomic, readwrite) CLLocationCoordinate2D coordinate;
```
friedbunny: Not all `MGLAnnotation` implementations allow for `coordinate` to be changed, so we’ve been making this mutable on an as-necessary basis.

platform/ios/src/MGLAnnotationView.h
```diff
@@ -49,6 +57,23 @@ NS_ASSUME_NONNULL_BEGIN
...
+/**
+ All states are handled automatically when `draggable` is set to YES.
+ Custom animations can be achieved by overriding setDragState:animated:
+ */
+@property (nonatomic, readonly) MGLAnnotationViewDragState dragState;
+
+/**
+ Implementer may override this method to use a custom animation for the different MGLAnnotationViewDragState.
+ */
+- (void)setDragState:(MGLAnnotationViewDragState)dragState animated:(BOOL)animated __attribute__((objc_requires_super));
```
1ec5: Nit: use `NS_REQUIRES_SUPER` instead of the raw attribute.

platform/ios/src/MGLAnnotationView.mm
```diff
@@ -95,6 +103,129 @@ - (void)updateScaleForPitch:(CGFloat)pitch
...
+    _dragState = dragState;
+    
+    switch (dragState) {
+        case MGLAnnotationViewDragStateNone:
+            break;
+        case MGLAnnotationViewDragStateStarting: {
+            if (animated)
+            {
+                [UIView animateWithDuration:.20 animations:^{
+                    self.transform = CGAffineTransformScale(CGAffineTransformIdentity, 2, 2);
```
1ec5: Does `-[MKAnnotationView setDragState:animated:]` have any built-in animation like this? I thought only MKPinAnnotationView implements that method, while MKAnnotationView only has it as a stub. Putting a built-in animation here means that a draggable annotation view always behaves like a pin, even if it doesn’t look like one. Maybe we need to finally create an MGLPinAnnotationView (which would be the ideal place to implement #2250).

1ec5:
Should long-pressing an annotation view select the associated annotation and trigger the selection delegate methods (#5130)?

frederoni:
`MKAnnotationView` cannot be dragged until it has been selected. So to initiate a drag, you have to tap -> long-press -> pan. Many examples use a workaround to make annotation views draggable immediately after long-pressing by setting the `selected` property to true in `MKMapViewDelegate didChangeDragState, didAddAnnotationViews and didDeselectAnnotationView`.

I will postpone this issue due to more critical ( [#5127] (https://github.com/mapbox/mapbox-gl-native/issues/5127), [#5128] (https://github.com/mapbox/mapbox-gl-native/issues/5128) and [#5129] (https://github.com/mapbox/mapbox-gl-native/issues/5129) ) which this issue depends on.
Will take another stab here when `selected` and `enabled` is working properly.

platform/ios/src/MGLAnnotationView.mm
frederoni: You are absolutely right @1ec5 the animation stuff should not be included in `MGLAnnotationView`

PR #5277, opened 2016-06-07
Enable Update Of InfoWindow Title And Snippet

### LINKED ISSUE #5237: Make InfoWindow dynamically updatable
We expose setTitle and setSnippet on Marker to set the contents of the default InfoWindow. At this point you can only set them once. Calling these methods consecutively will not always result in changing the contents of the InfoWindow (eg. when it's already open).  We have been hearing about the requirement to do this and this follows the line set in `4.1.0` to make other Marker attributes updatable. 

We should also look into the possibility to expose an InfoWindowAdapter equivalent for this use-case by exposing an invalidate method on InfoWindowAdapter (this will result in recreating the InfoWindow).

bleege:
Closes #5237


bleege:
Squashed and Merged!

Issue #3249, opened 2015-12-10
-[MGLMapView selectAnnotation:animated:] does not pan map to fit callout view


1ec5:
When YES is passed into `-[MKMapView selectAnnotation:animated:]`, MKMapView automatically pans so that the entire callout view fits in the viewport. Instead, `-[MGLMapView selectAnnotation:animated:]`’s `animated` parameter controls whether the callout view fades in.

A fix for this discrepancy may have to wait until we fix #3154 and #2618.

/cc @incanus @friedbunny

1ec5:
Now that #6676 has landed, this is no longer blocked.

1ec5:
This could just be a matter of calling `-showAnnotation:` within `-selectAnnotation:animated:`. However, we’d have to take care not to reenter `-selectAnnotation:animated:` and cause infinite recursion.

1ec5:
`-showAnnotation:` puts the annotation in the center of the viewport, which for point annotations would lead to an undesirable wandering effect as the user selects various annotations. But we could craft a simple method that pans the map only enough to show the point annotation and its callout.

1ec5:
https://github.com/mapbox/mapbox-gl-native/issues/2600#issuecomment-357845624 illustrates what the fix for this issue might look like inside MGLMapView.

julianrex:
Closing since this was addressed in #11442

PR #5283, opened 2016-06-08
[ios] fixes #5128 selected property to MGLAnnotationView

### LINKED ISSUE #5128: Add selected property to MGLAnnotationView

frederoni:
Fixes [#5128](https://github.com/mapbox/mapbox-gl-native/issues/5128)
Simply added `selected` property to `MGLAnnotationView` and hooked it up.
Also notice the small change in `selectAnnotation:animated:` to make deselection animatable.

### DIFF
--- a/platform/ios/src/MGLAnnotationView.h
+++ b/platform/ios/src/MGLAnnotationView.h
@@ -41,6 +41,12 @@ NS_ASSUME_NONNULL_BEGIN
 @property (nonatomic, assign, getter=isFlat) BOOL flat;
 
 /**
+ Defaults to NO. Becomes YES when clicked on in the map view.
+ */
+@property (nonatomic, assign, getter=isSelected) BOOL selected;
+- (void)setSelected:(BOOL)selected animated:(BOOL)animated;
+
+/**
  Setting this property to YES will cause the annotation view to shrink as it approaches the horizon and grow as it moves away from the
  horizon when the associated map view is tilted. Conversely, setting this property to NO will ensure that the annotation view maintains
  a constant size even when the map view is tilted. To maintain consistency with annotation representations that are not backed by an
--- a/platform/ios/src/MGLAnnotationView.mm
+++ b/platform/ios/src/MGLAnnotationView.mm
@@ -37,6 +37,16 @@
     self.center = self.center;
 }
 
+- (void)setSelected:(BOOL)selected
+{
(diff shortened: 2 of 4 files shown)


platform/ios/src/MGLAnnotationView.h
```diff
@@ -41,6 +41,12 @@ NS_ASSUME_NONNULL_BEGIN
 @property (nonatomic, assign, getter=isFlat) BOOL flat;
 
 /**
+ Defaults to NO. Becomes YES when clicked on in the map view.
```
1ec5: On iOS, we say "tap" instead of "click".

platform/ios/src/MGLAnnotationView.mm
```diff
@@ -37,6 +37,16 @@ - (void)setCenterOffset:(CGVector)centerOffset
...
 }
 
+- (void)setSelected:(BOOL)selected
+{
+    [self setSelected:selected animated:NO];
+}
+
+- (void)setSelected:(BOOL)selected animated:(BOOL)animated
+{
+    _selected = selected;
```
1ec5: In order to be KVO-compliant, we need to bracket this line with calls to `-willChangeValueForKey:` and `-didChangeValueForKey:`.

platform/ios/src/MGLMapView.mm
```diff
@@ -3401,6 +3401,8 @@ - (void)selectAnnotation:(id <MGLAnnotation>)annotation animated:(BOOL)animated
             positioningRect = annotationView.frame;
             
             [annotationView.superview bringSubviewToFront:annotationView];
+            
+            [annotationView setSelected:YES animated:animated];
```
1ec5: The `animated` property was intended to determine whether the map view pans to the selected annotation with animation: #3249. I don't think MapKit uses it to determine whether other things like the annotation view's visual appearance should animate.

frederoni:
continue in [#5297](https://github.com/mapbox/mapbox-gl-native/pull/5297)

PR #5286, opened 2016-06-08
[core] Tighten LatLng and other geo.hpp class invariants

### LINKED ISSUE #3802: LatLng and LatLngBounds should enforce invariants

jfirebaugh:
- Remove LatLng::null and enforce non-NaN, non-infinite coordinates and -90 <= lat <= 90
- Remove unnecessary operator bool()

This has caught two bugs so far:
- lat/lng swapping in TileJSON center and bounds
- Wrong w₀ value in `Transform::flyTo`

Fixes #3802. Reroll of #4355.

### DIFF
--- a/include/mbgl/util/geo.hpp
+++ b/include/mbgl/util/geo.hpp
@@ -9,6 +9,7 @@
 #include <mapbox/geometry/box.hpp>
 
 #include <cmath>
+#include <stdexcept>
 
 namespace mbgl {
 
@@ -21,17 +22,29 @@ using ScreenBox        = mapbox::geometry::box<double>;
 
 class LatLng {
 public:
-    struct null {};
-
     double latitude;
     double longitude;
 
     enum WrapMode : bool { Unwrapped, Wrapped };
 
-    LatLng(null) : latitude(std::numeric_limits<double>::quiet_NaN()), longitude(latitude) {}
-
     LatLng(double lat = 0, double lon = 0, WrapMode mode = Unwrapped)
-        : latitude(lat), longitude(lon) { if (mode == Wrapped) wrap(); }
+        : latitude(lat), longitude(lon) {
+        if (std::isnan(lat)) {
+            throw std::domain_error("latitude must not be NaN");
+        }
+        if (std::isnan(lon)) {
+            throw std::domain_error("longitude must not be NaN");
+        }
+        if (std::abs(lat) > 90.0) {
+            throw std::domain_error("latitude must be between -90 and 90");
+        }
+        if (!std::isfinite(lon)) {
+            throw std::domain_error("longitude must not be infinite");
+        }
+        if (mode == Wrapped) {
+            wrap();
+        }
+    }
 
     LatLng wrapped() const { return { latitude, longitude, Wrapped }; }
 
@@ -46,10 +59,6 @@ public:
         if (delta < util::LONGITUDE_MAX || delta > util::DEGREES_MAX) return;
         if (longitude > 0 && end.longitude < 0) longitude -= util::DEGREES_MAX;
         else if (longitude < 0 && end.longitude > 0) longitude += util::DEGREES_MAX;
-    }
-
-    explicit operator bool() const {
-        return !(std::isnan(latitude) || std::isnan(longitude));
     }
 
     // Constructs a LatLng object with the top left position of the specified tile.
@@ -72,10 +81,6 @@ public:
 
     ProjectedMeters(double n = 0, double e = 0)
         : northing(n), easting(e) {}
-
-    explicit operator bool() const {
-        return !(std::isnan(northing) || std::isnan(easting));
-    }
 };
 
 constexpr bool operator==(const ProjectedMeters& a, const ProjectedMeters& b) {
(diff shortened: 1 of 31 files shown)


brunoabinader:
@jfirebaugh want to apply this before #8601?

jfirebaugh:
I rebased this on master, but I think it needs attention at the SDK binding point for LatLng -- we want the SDK to enforce the same invariants, but report errors using a platform-appropriate exception mechanism (not `std::domain_error`).

platform/darwin/src/MGLGeometry_Private.h
```diff
@@ -13,6 +13,18 @@
 CGRect MGLExtendRect(CGRect rect, CGPoint point);
 
 NS_INLINE mbgl::LatLng MGLLatLngFromLocationCoordinate2D(CLLocationCoordinate2D coordinate) {
+    if (std::isnan(coordinate.latitude)) {
+        [NSException raise:NSInvalidArgumentException format:@"latitude must not be NaN"];
+    }
+    if (std::isnan(coordinate.longitude)) {
+        [NSException raise:NSInvalidArgumentException format:@"longitude must not be NaN"];
+    }
+    if (std::abs(coordinate.latitude) > 90.0) {
```
1ec5: I think we’ll want to push these checks further up. In many places where we call this function, the analogous MapKit API fails silently when given out-of-bounds coordinates; an exception would be heavy-handed. In other cases, a debug assertion is all that’s needed.

src/mbgl/map/transform.cpp
```diff
@@ -198,9 +196,9 @@ void Transform::flyTo(const CameraOptions &camera, const AnimationOptions &anima
 
     /// w₀: Initial visible span, measured in pixels at the initial scale.
     /// Known henceforth as a <i>screenful</i>.
-    double w0 = padding ? std::max(state.size.width, state.size.height)
-                        : std::max(state.size.width - padding.left - padding.right,
-                                   state.size.height - padding.top - padding.bottom);
+    double w0 = padding ? std::max(state.size.width - padding->left - padding->right,
+                                   state.size.height - padding->top - padding->bottom)
+                        : std::max(state.size.width, state.size.height);
```
1ec5: Wow, great catch! I wonder if this makes #8300 more difficult to hit. Hard to say just from inspecting the code, since `r()` is a bit convoluted.

include/mbgl/util/geo.hpp
```diff
@@ -21,17 +22,29 @@ using ScreenBox        = mapbox::geometry::box<double>;
 
 class LatLng {
 public:
-    struct null {};
-
     double latitude;
```
brunoabinader: Can we mark `latitude` and `longitude` as `const` here? We validate their values when initializing `LatLng`, but these are free to be modified after without checks.

Another option is to move these checks to a setter, and having these private.

platform/darwin/src/MGLGeometry_Private.h
jfirebaugh: Can you be more specific about what "further up" means in this context, or what specific changes you'd recommend?

1ec5: My suggestion was to move these checks to individual call sites, or to add a “checked” version of this function that wraps `MGLLatLngFromLocationCoordinate2D()`, adding these checks.

I reviewed the existing call sites to `MGLLatLngFromLocationCoordinate2D()` and tested similar functionality in MapKit using invalid coordinates. Based on MapKit behavior, having the exceptions where they are is OK, although we’ll want to add identical preflight checks in methods like `-[MGLMapView convertCoordinate:toPointToView:]` that return things like `CGPointMake(NAN, NAN)`.

Mapbox call site | MapKit equivalent | MapKit behavior | Mapbox expected behavior
----|----|----|----
`-[MGLMapView handlePinchGesture:]` | — | Constrained | `NSAssert` (but `NSException` is fine)
`-[MGLMapView _setCenterCoordinate:edgePadding:zoomLevel:direction:duration:animationTimingFunction:completionHandler:]` | `-[MKMapView setRegion:animated:]` | `NSException` | `NSException`
`-[MGLMapView cameraOptionsObjectForAnimatingToCamera:edgePadding:]` | `-[MKMapView setCamera:animated:]` | `NSException` | `NSException`
`-[MGLMapView convertCoordinate:toPointToView:]` | `-[MKMapView convertCoordinate:toPointToView:]` | `CGPoint(NaN, NaN)` | `CGPoint(NaN, NaN)`
`-[MGLMapView showAnnotations:edgePadding:animated:]` | `-[MKMapView showAnnotations:animated:]` | Moves to invalid part of map, in ocean | `NSException` is OK
`-[MGLMapView directionByFollowingWithCourse]` | — | — | `NSException`

1ec5: ea20dd58e66981e39da87e3dfea663494d746011 adds preflight checks to some of the call sites above that avoid the exception and instead return special values indicating invalidity. If you try to convert an invalid coordinate to a point, you get an invalid point. You can create a camera with an invalid coordinate, but you’ll get an exception when you try to move the map to that camera. You can create a polyline, polygon, or point collection with an invalid coordinate, and the overlay will be invalid, but you’ll get an exception when you try to add that shape to the map.

PR #5264, opened 2016-06-07
Bring annotation view to front when selected

1ec5:
Fixes #991. Split out from #5245.

/cc @boundsj

PR #5287, opened 2016-06-08
Update annotation views before informing delegate

1ec5:
On each frame update, update annotation views before informing the delegate of the new frame. This allows the delegate to manipulate annotation views without fear of the SDK overriding those modifications.

For example, if you want a particular view-backed annotation other than the selected annotation to appear in front of any other view-backed annotations, your MGLMapViewDelegate can hold a strong reference to the view it provides for that annotation and implement this method:

```objc
- (void)mapViewDidFinishRenderingFrame:(MGLMapView *)mapView fullyRendered:(BOOL)fullyRendered {
    // Make way for Its Preeminence!
    [self.preeminentAnnotationView.superview bringSubviewToFront:self.preeminentAnnotationView];
}
```

/ref #991 #5264
/cc @frederoni @boundsj @zugaldia

### DIFF
--- a/platform/ios/src/MGLMapView.mm
+++ b/platform/ios/src/MGLMapView.mm
@@ -4454,11 +4454,11 @@ mbgl::Duration MGLDurationInSeconds(NSTimeInterval duration)
         case mbgl::MapChangeDidFinishRenderingFrame:
         case mbgl::MapChangeDidFinishRenderingFrameFullyRendered:
         {
+            [self updateAnnotationViews];
             if ([self.delegate respondsToSelector:@selector(mapViewDidFinishRenderingFrame:fullyRendered:)])
             {
                 [self.delegate mapViewDidFinishRenderingFrame:self fullyRendered:(change == mbgl::MapChangeDidFinishRenderingFrameFullyRendered)];
             }
-            [self updateAnnotationViews];
             break;
         }
     }


1ec5:
If the developer needs the previous behavior, in which the delegate gets notified before MGLMapView arranges its MGLAnnotationViews, MGLMapViewDelegate can implement `-mapViewWillStartRenderingFrame:` instead of `-mapViewDidFinishRenderingFrame:fullyRendered:`.

boundsj:
👍

andrewstay:
@1ec5 Please note that this approach is extremely risky as the ```MGLAnnotationView``` gets reused for different annotations to be rendered. Therefore clients cannot hold a strong reference to it.

1ec5:
> Therefore clients cannot hold a strong reference to it.

This change is unrelated to how MGLAnnotationView gets used or reused. Instead of holding a strong reference to the view in application code, consider calling `-[MGLMapView viewForAnnotation:]` to get a fresh reference to the view anytime you need it.

> Moreover, in the SDK you already move to the front the selected annotation if it sits behind the other ones. Please consider exposing this to the map view interface.

Are you referring to the functionality added in #5264? Please feel free to open a new issue requesting a public API for controlling it; I’m not quite sure what kind of configurability you’re looking for.

PR #5288, opened 2016-06-08
text-pitch-alignment

yhahn:
Native port of https://github.com/mapbox/mapbox-gl-js/pull/2668.
native-port

### Backlog

- [x] Render tests pass
- [x] Code review
- [x] Coordinate with https://github.com/mapbox/mapbox-gl-js/pull/2668, style spec attribute

cc @lucaswoj @jfirebaugh @mollymerp @lxbarth

### DIFF
--- a/scripts/generate-style-code.js
+++ b/scripts/generate-style-code.js
@@ -21,7 +21,7 @@ global.propertyType = function (property) {
   if (/-translate-anchor$/.test(property.name)) {
     return 'TranslateAnchorType';
   }
-  if (/-rotation-alignment$/.test(property.name)) {
+  if (/-(rotation|pitch)-alignment$/.test(property.name)) {
     return 'RotationAlignmentType';
   }
   switch (property.type) {
--- a/scripts/generate-style-code.js
+++ b/scripts/generate-style-code.js
@@ -6,14 +6,12 @@ const spec = require('mapbox-gl-style-spec').latest;
 var parseCSSColor = require('csscolorparser').parseCSSColor;
 
 global.camelize = function (str) {
-  str = str === undefined ? 'undefined' : str;
   return str.replace(/(?:^|-)(.)/g, function (_, x) {
     return x.toUpperCase();
   });
 }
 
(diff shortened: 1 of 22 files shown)


src/mbgl/renderer/painter_symbol.cpp
```diff
@@ -48,7 +48,7 @@ void Painter::renderSDF(SymbolBucket &bucket,
     bool alignedWithMap = rotationAlignment == RotationAlignmentType::Map;
     float gammaScale = 1.0f;
 
-    if (alignedWithMap) {
+    if (false && alignedWithMap) {
```
yhahn: I'll fix this, was poking but this whole codepath can be dropped.

brunoabinader:
Good stuff @yhahn! Do you know if this affects the recently added `mbgl::ViewportMode::FlippedY`?

yhahn:
@jfirebaugh :pray: ready for your review when you have a chance : )

scripts/generate-style-code.js
```diff
@@ -6,12 +6,14 @@ const spec = require('mapbox-gl-style-spec').latest;
 var parseCSSColor = require('csscolorparser').parseCSSColor;
 
 global.camelize = function (str) {
+  str = str === undefined ? 'undefined' : str;
```
jfirebaugh: What prompted this change, and the corresponding one in `camelizeWithLeadingLowercase`?

yhahn: This was to have `Undefined` be the value for `text-pitch-alignment` when it's not defined in the style and be able to detect when inheritance should kick in (per https://github.com/mapbox/mapbox-gl-style-spec/pull/460#issuecomment-225238097).

It's definitely the main eyebrow-raiser in this PR :smile: would love to know if there's a better way to achieve this.

jfirebaugh: Yeah, I see the `Alignment::Undefined`. Why does a JS `undefined` value pass through `camelize` though?

jfirebaugh:
👌

PR #5293, opened 2016-06-09
[core] Singleton pattern for mbgl::gl::{ObjectStore,TexturePool}

brunoabinader:
These objects can be safely using singleton pattern, which reduces a lot of redundant code and improves readability. Both objects resides on the same thread thus no need for ThreadLocal storage.

:eyes: @tmpsantos @kkaefer @jfirebaugh

### DIFF
--- a/src/mbgl/geometry/buffer.hpp
+++ b/src/mbgl/geometry/buffer.hpp
@@ -36,11 +36,11 @@ public:
     }
 
     // Transfers this buffer to the GPU and binds the buffer to the GL context.
-    void bind(gl::ObjectStore& store) {
+    void bind() {
         if (buffer) {
             MBGL_CHECK_ERROR(glBindBuffer(bufferType, *buffer));
         } else {
-            buffer = store.createBuffer();
+            buffer = gl::ObjectStore::get().createBuffer();
             MBGL_CHECK_ERROR(glBindBuffer(bufferType, *buffer));
             if (array == nullptr) {
                 Log::Debug(Event::OpenGL, "Buffer doesn't contain elements");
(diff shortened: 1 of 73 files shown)


kkaefer:
Did you test this with more than one map view being present, e.g. on iOS or Android? Seems like this could cause issues.

tmpsantos:
It was like this before (actually using TLS), the reason why it isn't anymore is because we moved rendering to the main thread so all map objects would unintentionally share the same store.

brunoabinader:
Yes, having multiple map/view objects is exactly one of the target scenarios:
- Instead of cleaning up the objects from a single map instance, we now clear them all (specially useful for `onLowMemory()`
- TexturePool can now act in a much broader scale

However, @kkaefer is right - there is one corner case in which I didn't account for: when calling for `TexturePool::clear()` inside the Map`s dtor. I guess we could solve that by using a shared_ptr to store the TexturePool on its factory. Thoughts?

brunoabinader:
> ~~Both objects resides on the same thread thus no need for ThreadLocal storage.~~

This isn't true in cases e.g. where multiple Map/View instances resides in different threads. Capturing the conversion with @kkaefer:

> so, in overall, the idea is 1) reduce the size of each batch of texture pools from 64 to a smaller value, and 2) have the object store and texture pool being shared among map/view instances as long as they reside on the same thread and share the same gl context

jfirebaugh:
👎 . Not using a singleton pattern for these was very much intentional. The singleton pattern is typically an anti-pattern, making the code harder to modify (what if it turns out you _do_ need more than one) and test (you frequently _do_ need more than one in tests). In addition, being required to have access to a `gl::ObjectStore` in order to create GL resources provides an assurance that we don't try to do so on background threads.

jfirebaugh:
9c26f063187b129218910dbb86eb21215a2cdf40

PR #5297, opened 2016-06-09
[ios] fixes #5127, #5128, #5130 added enabled and selected property to MGLAnnotationView

### LINKED ISSUE #5127: Add enabled property to MGLAnnotationView
MapKit has an [`enabled` property](https://developer.apple.com/library/ios/documentation/MapKit/Reference/MKAnnotationView_Class/#//apple_ref/occ/instp/MKAnnotationView/enabled) on the annotation view base class. This property informs touch behavior and subclasses can use it has a hint for content display. MGLAnnotationView should have a similar property.

### LINKED ISSUE #5128: Add selected property to MGLAnnotationView

frederoni:
Fixes  #5127 #5128 #5130 

Added `selected` and `enabled` property to `MGLAnnotationView` and hooked it up.
Added `mapView:did(De)SelectAnnotationView:` protocol methods.

@boundsj 
I kept the annotation and annotation views selection protocol methods separated but let me know if you want me to combine them into `mapView:did(De)SelectAnnotation:annotationView:`

/cc @1ec5 @friedbunny

### DIFF
--- a/platform/ios/src/MGLAnnotationView.h
+++ b/platform/ios/src/MGLAnnotationView.h
@@ -41,6 +41,26 @@ NS_ASSUME_NONNULL_BEGIN
 @property (nonatomic, assign, getter=isFlat) BOOL flat;
 
 /**
+ Defaults to NO and becomes YES when the view is tapped on.
+ 
+ Selecting another view will first deselect the currently selected view.
+ This property should not be changed directly.
+ */
+@property (nonatomic, assign, getter=isSelected) BOOL selected;
+
+/**
+ Subclasses may override this method in order to customize appearance.
+ This method should not be called directly.
+ */
+- (void)setSelected:(BOOL)selected animated:(BOOL)animated;
+
+/*
+ This property defaults to YES. Setting it to NO will cause the annotation view to ignore all touch events.
+ Subclasses may use this property to customize the appearance.
+ */
+@property (nonatomic, assign, getter=isEnabled) BOOL enabled;
+
+/**
  Setting this property to YES will cause the annotation view to shrink as it approaches the horizon and grow as it moves away from the
  horizon when the associated map view is tilted. Conversely, setting this property to NO will ensure that the annotation view maintains
  a constant size even when the map view is tilted. To maintain consistency with annotation representations that are not backed by an
--- a/platform/ios/src/MGLAnnotationView.mm
+++ b/platform/ios/src/MGLAnnotationView.mm
@@ -22,6 +22,7 @@
     {
         _reuseIdentifier = [reuseIdentifier copy];
         _scalesWithViewingDistance = YES;
+        _enabled = YES;
     }
     return self;
 }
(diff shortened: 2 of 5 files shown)


platform/ios/src/MGLAnnotationView.h
```diff
@@ -41,6 +41,19 @@ NS_ASSUME_NONNULL_BEGIN
 @property (nonatomic, assign, getter=isFlat) BOOL flat;
 
 /**
+ This property defaults to NO and becomes YES when the view is tapped/clicked on.
```
1ec5: Nit: Let’s remove the word “clicked”; that term is only appropriate for OS X, and this file isn’t shared with the OS X SDK.

platform/ios/src/MGLAnnotationView.mm
```diff
@@ -37,6 +38,25 @@ - (void)setCenterOffset:(CGVector)centerOffset
...
+- (void)setSelected:(BOOL)selected animated:(BOOL)animated
+{
+    [self willChangeValueForKey:@"selected"];
+    _selected = selected;
+    [self didChangeValueForKey:@"selected"];
+}
+
+- (void)setEnabled:(BOOL)enabled
+{
+    [self willChangeValueForKey:@"enabled"];
```
1ec5: `-willChangeValueForKey:`/`-didChangeValueForKey:` is only necessary when you set the ivar in a method other than a standard setter. So in the case of `-setSelected:animated:`, you need those calls, because it isn’t called `-setSelected:`.

Moreover, it’s unnecessary to implement `-setEnabled:` just to set the ivar; Objective-C can synthesize this method for you automatically.

platform/ios/src/MGLAnnotationView.h
1ec5: As with the `MKAnnotationView.selected` property, the documentation for this property should mention that you shouldn’t set it directly.

platform/ios/src/MGLAnnotationView.h
```diff
@@ -41,6 +41,19 @@ NS_ASSUME_NONNULL_BEGIN
...
+ Selecting another view will first deselect the currently selected view.
+ */
+@property (nonatomic, assign, getter=isSelected) BOOL selected;
+- (void)setSelected:(BOOL)selected animated:(BOOL)animated;
```
1ec5: Document this method also, and mention that you shouldn’t call this method directly.

frederoni:
That should cover it. Let me know if there's anything else @1ec5

platform/ios/src/MGLAnnotationView.h
```diff
@@ -41,10 +41,17 @@ NS_ASSUME_NONNULL_BEGIN
 @property (nonatomic, assign, getter=isFlat) BOOL flat;
 
 /**
- This property defaults to NO and becomes YES when the view is tapped/clicked on.
+ This property should not be changed directly.
```
1ec5: Thanks, this line should actually be at the end of the comment, rather than the beginning. The first several words of a documentation comment appear in the code completion popup, so it should be descriptive and reasonably unique.

Though it’s intended for Swift and Markdown instead of Objective-C and HeaderDoc, [this Swift API design guideline document](https://swift.org/documentation/api-design-guidelines/#write-doc-comment) has a great guideline on forming documentation comments (scroll up and click “Expand all details now” to see it).

1ec5:
This looks great, other than the one comment about comments.

1ec5:
👍

PR #5298, opened 2016-06-09
[android] #5296 - Add a container ViewGroup for MarkerViews

### LINKED ISSUE #5296: Marker views float over the attribution controls
The new Marker Views float over the attribution controls where the controls should always remain on top.

screen shot 2016-06-09 at 10 52 17

The markers are appended to the same FrameLayout as all the other components. With FrameLayout the order of the children determines the drawing order. An inner layout to contain the Marker Views would do the trick.

ivovandongen:
Fixes #5296 

@zugaldia


tobrun:
LGTM :shipit:

PR #5270, opened 2016-06-09
Memory leak bot

tmpsantos:
We should perform memory checks as part of our automated build bots process. The idea is to add a new Makefile target similar to `run-valgrind-glfw-app` that could run e.g. unit tests and analyzes the data from [valgrind's memcheck](http://valgrind.org/docs/manual/mc-manual.html).

/cc @mapbox/gl

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -18,13 +18,13 @@ addons_shortcuts:
   addons_qt4: &qt4
     apt:
       sources:  [ 'ubuntu-toolchain-r-test' ]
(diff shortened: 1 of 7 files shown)


tmpsantos:
:eyes: @brunoabinader @kkaefer

brunoabinader:
Looks great @tmpsantos! Once I fix the mason recipe for valgrind on OS X, we can activate it also on Bitrise. 🚢

PR #5302, opened 2016-06-09
[ios] Support for a customizable zoom gesture centering strategy.

revetkn:
We introduce a new MGLZoomGestureCentering enum and corresponding zoomGestureCentering property on MGLMapView.

Currently, Mapbox automatically adjusts the center of the map relative to the user's gesture location, but for some applications (e.g Uber, Lyft) it is preferable to keep the map center "locked" in response to zoom gestures.

Exisiting code will retain its current behavior without modification and will implicitly use the default value of MGLZoomGestureCenteringFollowsTouch. New code can specify MGLZoomGestureCenteringLockedInPlace to get the new "locked" behavior.

### DIFF
--- a/platform/ios/src/MGLMapView.h
+++ b/platform/ios/src/MGLMapView.h
@@ -31,6 +31,17 @@ typedef NS_ENUM(NSUInteger, MGLAnnotationVerticalAlignment) {
     MGLAnnotationVerticalAlignmentBottom,
 };
 
+/**
+  How the map view should adjust its center coordinate in response to
+  a user-initiated zoom gesture.
+ */
+typedef NS_ENUM(NSUInteger, MGLZoomGestureCentering) {
+    /** Adjusts the center of the map relative to the user's touch position. */
+    MGLZoomGestureCenteringFollowsTouch = 0,
+    /** Locks the zoom to the center of the map view. */
+    MGLZoomGestureCenteringLockedInPlace,
+};
+
 /** Options for enabling debugging features in an MGLMapView instance. */
 typedef NS_OPTIONS(NSUInteger, MGLMapDebugMaskOptions) {
     /** Edges of tile boundaries are shown as thick, red lines to help diagnose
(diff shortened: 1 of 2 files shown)


1ec5:
Besides the pinch/zoom gesture, the rotation, double-tap, two-finger tap, and double-tap-drag (“quickzoom”) gestures are also anchored to the user location when user tracking mode is on. Currently, these gesture recognizers make use of the `-userLocationAnnotationViewCenter` method or `-locationInView:` depending on the user tracking mode. I think we should make all these gestures respect the property you’ve added while simplifying the code: make them all use a new method, `-anchorPointForGesture:`, that uses either `-userLocationAnnotationViewCenter` or `-locationInView:` or `-contentCenter` depending on the user tracking mode and this new property.

An enum for this property is clearer than a simple Boolean. However, I wonder if it’s flexible enough for the use cases you describe. For example, what if the location you want to lock around is neither the user’s location nor the center of the view? It could be the selected annotation, for example. In that case, I think we could avoid adding a `zoomGestureCentering` property; instead, we’d have this `-anchorPointForGesture:` method use `-userLocationAnnotationViewCenter` or `-locationInView:` as the individual gesture recognizers do now. Applications that need custom behavior, such as anchoring to the view’s inset center, could subclass MGLMapView and override -`anchorPointForGesture:` to suit their own needs. Does that sound reasonable?

revetkn:
What a fast and well-thought-out response...thank you.  I agree that overriding a publicly-visible `-anchorPointForGesture:` method will give users more flexibility than having a fixed-value property.  I'll go ahead and rework this pull request.

revetkn:
OK!  Hopefully this is more in line with what you're thinking.

1ec5:
Perfect, thanks!

1ec5:
Updated changelog in 6998479d05497baf57eb264e92807d327920f7d6. Cherry-picked to the v3.3.0 release branch in dd565b701b20cac69f7a84a02e7bc8a783c92bb6 and 383df9dd21e518d3894072390a459a80487348ef.

PR #5307, opened 2016-06-10
[ios] fixes #5129 added annotation property to MGLAnnotationView

### LINKED ISSUE #5129: Add annotation property to MGLAnnotationView

frederoni:
Also added a test case to make sure the `annotation` property is handled correctly but there's no reference to the `platform/ios/test/MGLAnnotationViewTests.m` file in the project file as I'm figuring out the details around [#4641](https://github.com/mapbox/mapbox-gl-native/pull/4641).

### DIFF
--- a/platform/ios/src/MGLAnnotationView.h
+++ b/platform/ios/src/MGLAnnotationView.h
@@ -4,6 +4,8 @@
 
 NS_ASSUME_NONNULL_BEGIN
 
+@protocol MGLAnnotation;
+
 /** The MGLAnnotationView class is responsible for representing point-based annotation markers as a view. Annotation views represent an annotation object, which is an object that corresponds to the MGLAnnotation protocol. When an annotation’s coordinate point is visible on the map view, the map view delegate is asked to provide a corresponding annotation view. If an annotation view is created with a reuse identifier, the map view may recycle the view when it goes offscreen. */
 @interface MGLAnnotationView : UIView
 
@@ -14,6 +16,13 @@ NS_ASSUME_NONNULL_BEGIN
  @return The initialized annotation view object.
  */
 - (instancetype)initWithReuseIdentifier:(nullable NSString *)reuseIdentifier;
+
+/**
+ This property will be set to the associated annotation when the view is visible.
+ 
+ When the view is queued and waiting to be reused, the value will be set to nil.
+ */
+@property (nonatomic, readonly, nullable) id <MGLAnnotation> annotation;
 
 /**
  The string that identifies that this annotation view is reusable. (read-only)
--- a/platform/ios/test/MGLAnnotationViewTests.m
+++ b/platform/ios/test/MGLAnnotationViewTests.m
@@ -1,0 +1,62 @@
+#import <Mapbox/Mapbox.h>
+#import <XCTest/XCTest.h>
+
+static NSString * const MGLTestAnnotationReuseIdentifer = @"MGLTestAnnotationReuseIdentifer";
+
+@interface MGLTestAnnotation : NSObject <MGLAnnotation>
+@property (nonatomic, assign) CLLocationCoordinate2D coordinate;
+@end
+
+@implementation MGLTestAnnotation
+@end
+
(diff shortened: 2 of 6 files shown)


platform/ios/test/MGLAnnotationViewTests.m
```diff
@@ -0,0 +1,62 @@
...
+@implementation MGLAnnotationViewTests
+
+- (void)testAnnotationView
+{
+    _mapView = [[MGLMapView alloc] initWithFrame:CGRectZero];
+    _mapView.delegate = self;
+    
+    _expectation = [self expectationWithDescription:@"test annotation property"];
+    
+    MGLTestAnnotation *annotation = [MGLTestAnnotation new];
```
1ec5: But: use alloc/init rather than new. In general, using init directly makes it easier to add parameters to an initializer. (Admittedly, it used to matter more in the days before ARC.) In this codebase specifically, where many of the files are Objective-C++, we prefer to leave the new keyword to C++ usage.

platform/ios/test/MGLAnnotationViewTests.m
```diff
@@ -0,0 +1,62 @@
...
+    [_mapView addAnnotation:annotation];
+    
+    [self waitForExpectationsWithTimeout:1 handler:nil];
+}
+
+- (void)performTestWithAnnotationView:(MGLAnnotationView *)annotationView
```
1ec5: The flow in this test class is a bit hard to reason about. I'm adding XCUITests that would make this sort of thing much easier in #4837. In the meantime, I think a better structure for this test would be to perform these checks at the end of -testAnnotationView, after the wait. -mapView:viewForAnnotation: should fulfill the expectation that the delegate callback gets called, allowing -testAnnotationView to continue running.

1ec5:
#4837 adds a UI test bundle to the iOS project. Until then, you may be able to add this test to the existing unit test bundle, which otherwise only tests non-UI code that's shared between the iOS and OS X SDKs.

platform/ios/src/MGLAnnotationView.h
```diff
@@ -16,6 +18,14 @@ NS_ASSUME_NONNULL_BEGIN
 - (instancetype)initWithReuseIdentifier:(nullable NSString *)reuseIdentifier;
 
 /**
+ This property will be set to the associated annotation when the view is visible.
+ 
+ When the view is queued and waiting to be reused, the value will be set to nil.
+ This property should not be changed directly.
+ */
+@property (nonatomic, strong, nullable) id <MGLAnnotation> annotation;
```
boundsj: It does not look like there is a good reason for us to use the MK approach of warning about changing this property in the comment but not making it readonly. Let's go ahead and make it readonly.

1ec5: I wonder if the reason MapKit leaves this property read-write is that a subclass might want to override `-setAnnotation:` to do something special. At least, that’s the reason MapKit exposes `-setSelected:animated:` publicly with a note saying never to call it (as seen in #5297).

frederoni: I can think of one case where you would need to override `setAnnotation:` until we landed an equivalent for `-mapView:didAddAnnotationViews`
(which would also improve the test case in this PR). It's possible with a workaround but e.g animating an annotation view immediately after adding an annotation is not very straight forward. You can override `setAnnotation:` to achieve that now. So I'd suggest we make this property readonly and implement `-mapView:didAddAnnotationViews`

frederoni: The annotation corresponds to the visibility of the view so it could be useful to override in order to animate the adding of the annotations views. However, I'd also prefer to make this property readonly and maybe expose `-mapView:didAddAnnotationViews:`

1ec5:
👍

PR #5315, opened 2016-06-10
[tests] Add tests for detecting memory leaks

tmpsantos:
Simple tests that will render a raster and a vector map and hopefully the valgrind bot will detect something wrong if any.

The other test will look for application lifetime leaks that cannot be detected with valgrind. It can also detect GL leaks if they are very expressive. A threshold is needed because memory will grow anyway due to fragmentation.


tmpsantos:
:eyes: @mikemorris @brunoabinader

brunoabinader:
Coverage tests jumped to ~74% ❤️

PR #5316, opened 2016-06-10
[core] Use default transition ease for style transition

### LINKED ISSUE #363: Change style transition easing
Currently, we are doing linear value transitions when applying/removing classes. We should change this to use a `0, 0, 0.25, 1` bezier curve easing.

brunoabinader:
Use a shared default `UnitBezier` transition ease for both transform and style transitions. `UnitBezier` is now `constexpr` and all its functions can be safely marked `const`.

Fixes #363.

👀 @jfirebaugh @kkaefer @1ec5

### DIFF
--- a/include/mbgl/util/unitbezier.hpp
+++ b/include/mbgl/util/unitbezier.hpp
@@ -31,32 +31,31 @@ namespace mbgl {
 namespace util {
 
 struct UnitBezier {
-    UnitBezier(double p1x, double p1y, double p2x, double p2y) {
-        // Calculate the polynomial coefficients, implicit first and last control points are (0,0) and (1,1).
-        cx = 3.0 * p1x;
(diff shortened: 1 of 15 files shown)


brunoabinader:
👀 @tmpsantos for the Qt changes.

include/mbgl/util/unitbezier.hpp
```diff
@@ -100,18 +99,18 @@ struct UnitBezier {
         return t2;
     }
 
-    double solve(double x, double epsilon) {
+    double solve(double x, double epsilon) const {
         return sampleCurveY(solveCurveX(x, epsilon));
     }
 
 private:
-    double ax;
```
kkaefer: Can we make them `const`?

kkaefer:
👍

PR #5317, opened 2016-06-10
[core] Make enum ↔ string conversion more generic-friendly

jfirebaugh:
Rewrite enum.hpp in such a way that parseConstant can be defined generically for all enumerated types.

While there, fix validation of enumerated types ~~and make enum.hpp a private header~~.


jfirebaugh:
Rolling into #5318.

PR #5318, opened 2016-06-10
Partial implementation of runtime styling API for node bindings

jfirebaugh:
Enough to pass the three basic runtime styling tests in test-suite. 🎉 

Still TODO:

* [ ] `setLayoutProperty`
* [ ] `setPaintProperty` with a function
* [ ] Support color, translate, dash-array, and font-stack properties


jfirebaugh:
There will be some conflicts with #5288; I'm going to wait for the merge there and then rebase and resolve them here.

PR #5325, opened 2016-06-12
Re-enable tidy checks

brunoabinader:
Re-enable `clang-tidy` checks. Background:
> clang-tidy is a clang-based C++ “linter” tool. Its purpose is to provide an extensible framework for diagnosing and fixing typical programming errors, like style violations, interface misuse, or bugs that can be deduced via static analysis. 

The following checks are being performed (list is not final):
- [llvm-namespace-comment](http://clang.llvm.org/extra/clang-tidy/checks/llvm-namespace-comment.html)
- [modernize-use-nullptr](http://clang.llvm.org/extra/clang-tidy/checks/modernize-use-nullptr.html)
- [modernize-pass-by-value](http://clang.llvm.org/extra/clang-tidy/checks/modernize-pass-by-value.html)
- [modernize-use-override](http://clang.llvm.org/extra/clang-tidy/checks/modernize-use-override.html)
- [modernize-use-default](http://clang.llvm.org/extra/clang-tidy/checks/modernize-use-default.html)
- [modernize-use-auto](http://clang.llvm.org/extra/clang-tidy/checks/modernize-use-auto.html)
- [modernize-loop-convert](http://clang.llvm.org/extra/clang-tidy/checks/modernize-loop-convert.html)
- [misc-static-assert](http://clang.llvm.org/extra/clang-tidy/checks/misc-static-assert.html)

/cc @mapbox/gl

### DIFF
--- a/.clang-tidy
+++ b/.clang-tidy
@@ -1,1 +1,1 @@
-Checks: '-clang-analyzer-security.insecureAPI.rand'
+Checks: 'modernize-*,misc-static-assert,llvm-namespace-comment,-clang-analyzer-security.insecureAPI.rand,-clang-analyzer-core.uninitialized.UndefReturn'
--- a/Makefile
+++ b/Makefile
@@ -289,7 +289,7 @@ check: test
 	scripts/collect-coverage.sh $(LINUX_OUTPUT_PATH)/$(BUILDTYPE)
 
 # Generates a compilation database with ninja for use in clang tooling
-compdb: compdb-$(HOST_PLATFORM)
+compdb: node_modules compdb-$(HOST_PLATFORM)
 
 compdb-linux: platform/linux/platform.gyp $(LINUX_OUTPUT_PATH)/config.gypi
 	$(GYP) -f ninja -I $(LINUX_OUTPUT_PATH)/config.gypi \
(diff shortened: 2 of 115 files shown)


brunoabinader:
Mason binaries for `clang-tidy 3.8` are missing for linux-x86_64. I'll handle this before resuming this PR.

brunoabinader:
Hmm, the clang-tidy build bot took ~22min to finish. We could reduce that by 1) splitting the tidy scan in 2-3 bots or 2) making it a daily scheduled run. Thoughts?

brunoabinader:
> ~~Hmm, the clang-tidy build bot took ~22min to finish.~~

Managed to reduce time to ~15min (removed redundant `make osx` build).

jfirebaugh:
Nice cleanups! Is there a check for unnecessary `inline` in class declarations? That would be a useful one for our code.

What's the reason to use Bitrise here? I think if we could get this running on Travis it would be best to avoid the overhead of another Bitrise app.

brunoabinader:
> Is there a check for unnecessary inline in class declarations?

It seems `readability-redundant-inline` is just going out of the oven: http://lists.llvm.org/pipermail/cfe-commits/Week-of-Mon-20160411/155258.html

> What's the reason to use Bitrise here? I think if we could get this running on Travis it would be best to avoid the overhead of another Bitrise app.

We use Bitrise here just as a workaround until we get proper `clang-tidy` v3.8 mason binary packages for Linux.

jfirebaugh:
> proper clang-tidy v3.8 mason binary packages for Linux

Is there a ticket for that?

brunoabinader:
> Is there a ticket for that?

Just did 👍 https://github.com/mapbox/mason/issues/188

PR #5328, opened 2016-06-13
[android] #5276 - icon should be optional for MarkerView

### LINKED ISSUE #5276: MarkerViewOptions.icon() Should Be Optional
Position should be the only thing that needs to be passed to `MarkerViewOptions()`. Currently `.icon` is also required instead of just using the default red marker icon `MarkerOptions()` uses. In other words, setting up a simple marker view currently looks like this:

```java
 mapboxMap.addMarker(new MarkerViewOptions()
                        .position(new LatLng(-33.85699436, 151.21510684))
                        .icon(icon)); // This line shouldn't be required
```

cc: @bleege @tobrun @zugaldia @ivovandongen

tobrun:
This PR resolves the crash when trying to add a MarkerView without an icon. This PR introduces a default icon for MarkerView. We can't reuse the icon from Marker since that has extra transparent pixels
closes #5276

cc @bleege @cammace @ivovandongen @zugaldia


bleege:
Thanks for picking this up @tobrun.  One question though, why bundle and use `default_markerview.png` and not use the the default marker that Core GL already provides?

tobrun:
@bleege 

The gl marker:

screen shot 2016-06-13 at 16 34 55


The view marker:

screen shot 2016-06-13 at 16 36 15


The problem we had with GL markers is that we didn't have a way to do anchoring/offsets. To be able to put the bottom of the marker at the correct position, we had to add extra transparent pixels. Now with view markers we have anchors and they are by default placed at the bottom of the image. 

This is what @cammace was indicating in https://github.com/mapbox/mapbox-gl-native/issues/5276#issuecomment-224404025 with:
> BTW the default marker image has a bunch of bottom padding right now and can either be cropped out or .anchor(0.5f,0.5f) needs to be used

tobrun:

tobrun:
rebased against master instead of release and only noticed after force pushing, follow up PR in #5348

bleege:
> The problem we had with GL markers is that we didn't have a way to do anchoring/offsets. To be able to put the bottom of the marker at the correct position, we had to add extra transparent pixels. Now with view markers we have anchors and they are by default placed at the bottom of the image. 

Thanks for the details in the thinking behind this.  Can you help me understand why the solution wasn't to fix the Core GL markers then instead of duplicating default markers?

PR #5332, opened 2016-06-13
4924 make gesture focal point configurable

### LINKED ISSUE #4924: Make gesture focal point configurable
**tldr we need to expose a user configurable focal point. A focal represents the x,y value on which a gesture works. In normal cases the x,y from the MotionEvent is used. We need to make this configurable**

Currently we already expose an API that works similar to a  focal point implementation. When you enable the tracking mode on MyLocationView you also change the focal point to follow that view. An example how that now looks:

```java
                    // Single finger double tap
                    if (mMapboxMap.getTrackingSettings().isLocationTrackingDisabled()) {
                        // Zoom in on gesture
                        zoom(true, e.getX(), e.getY());
                    } else {
                        // Zoom in on user location view
                        // getCenterX also takes in account padding + insets
                        zoom(true, mMyLocationView.getCenterX(), mMyLocationView.getCenterY());
                    }
```

To be able to support all different use-cases users are requesting we need to expose an API as:
 
```java
mapboxMap.getUiSettings().setFocalPoint(x,y);
```

People will mostly use this to center the focal point 

```java
mapboxMap.getUiSettings().setFocalPoint(mapView.getWidth()/2,mapView.getHeight()2));
```

It would be nice to be able to limit if-else in the gesture code above and use a factory pattern to provide an implementation instead.

tobrun:
Closes #4924

### DIFF
--- a/platform/android/MapboxGLAndroidSDK/src/main/java/com/mapbox/mapboxsdk/maps/MapView.java
+++ b/platform/android/MapboxGLAndroidSDK/src/main/java/com/mapbox/mapboxsdk/maps/MapView.java
@@ -152,6 +152,8 @@ public class MapView extends FrameLayout {
     private int mContentPaddingRight;
     private int mContentPaddingBottom;
 
+    private PointF mFocalPoint;
+
     private StyleInitializer mStyleInitializer;
 
     private List<OnMapReadyCallback> mOnMapReadyCallbackList;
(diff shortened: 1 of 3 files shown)


bleege:
Restarted tests and now good to go!  Will squash and merge.

bleege:
Squashed and then while rebasing on to `release-android-v4.1.0` discovered a merge conflict.  Will address next.

bleege:
Rebased after squashing `UiSettings.java` conflict and then merged!

PR #5334, opened 2016-06-13
Support for icon-text-fit, icon-text-fit-padding

yhahn:
Native port of https://github.com/mapbox/mapbox-gl-js/pull/2720.

### Backlog

- [x] Style spec coordination
- [x] Code review


mb12:
@yhahn Can you please kindly clarify the following?

1. Should the default value of icon-text-fit be both instead of none?  This would ensure that the shield symbolizer icon is scaled by default to cover the text on it.

2. Also is it possible to explain what is the following number 24.0F? I've noticed this several times in the symbol placement code but I do not understand what it is for etc.. 

auto size = layout.textSize / 24.0f;

yhahn:
> Should the default value of icon-text-fit be both instead of none? This would ensure that the shield symbolizer icon is scaled by default to cover the text on it.

No since this would alter the behavior of all existing icons in styles. This is an opt-in mode so it defaults to `none` and then if a style specifies one of the `both/width/height` values it enables icon-sizing/positioning based on text.

> Also is it possible to explain what is the following number 24.0F? I've noticed this several times in the symbol placement code but I do not understand what it is for etc.

Yes, this is the current fixed size of the glyph SDF bitmaps used by Mapbox GL and is a "magic number" that glyph metrics are currently related to. You can think of text shaping as occurring on a 24.0-based scale independently of text size, etc.

To convert between shaped text and actual size of quads/pixels on screen the 24.0 magic number needs to be scaled by text size.

src/mbgl/style/property_parsing.cpp
```diff
@@ -39,7 +39,21 @@ optional<std::string> parseConstant(const char* name, const JSValue& value) {
 
 template <>
 optional<Color> parseConstant(const char* name, const JSValue& value) {
```
yhahn: @jfirebaugh help : )

Defining

```
optional<std::array<float, 4>> parseConstant(const char* name, const JSValue& value) {
```

is considered a redefinition of `optional<Color>`. Is there a noninvasive way around this or do we just decide on the semantics here and maybe go with `optional<std::array<float, 4>>` as less misleading?

Other options?

jfirebaugh: We should introduce a real class definition for `Color`, rather than using a typedef. I've been meaning to do this for a while. Want to take a stab?

yhahn: Yep, will poke you when I have questions.

yhahn:
@jfirebaugh ready for review.

jfirebaugh:
👍

Issue #5282, opened 2016-06-08
Build Location Picker Example In TestApp


bleege:
A common use case for apps is to have a Location Picker over a map and while the TestApp has the basics of this already in the `GeocoderActivity` it can be enhanced to mirror the functionality that traditionally is part of this use case as well as made more discoverable.  Specifically, this example should have:

* A Marker (pin) that is centered in the middle of the screen that moves as the user pans around the map
* An animation should be triggered on the UserLocationView when the Marker is moved over it to indicate that the Marker and User Location have come into alignment.
* An InfoWindow for the Marker that displays the address of the selected location
  * The address is generated using the Mapbox Geocoding API in [Mapbox Android Services](https://github.com/mapbox/mapbox-java).
* Once a location is selected via Marker all zoom events should be centered on the Marker's location.

bleege:
When getting ready to hook up the Geocoder I noticed that the TestApp is still running the old school (and deprecated) Mapbox Directions and Mapbox Geocoding libraries and not [Mapbox Android Services](https://www.mapbox.com/android-sdk/#mapbox_android_services) as it should be these days.  Before I can build the Location Picker I need to do a little house cleaning and get things ported to MAS.

bleege:
I've got the basic UI of Location Picker built, namely pan map using a marker, select location, geocode location, display address with Marker and opened InfoWindow.  To finish this part of the requirements off I need to:

* [ ] Add Reset To Geocoding Search Mode UI
* [ ] Fix Pin Alignment so that Search Pin to Results Pin is seamless
* [ ] Improve Address Formatting in the InfoWindow

Once these are done the next steps are to add the User Dot, User Dot animation, and Zoom events to be centered on Results Pin location.

![2016-06-10 14_52_00](https://cloud.githubusercontent.com/assets/989928/15976862/eee813e8-2f1a-11e6-9df6-7740e64cd6e2.gif)


/cc @zugaldia

bleege:
Shortening the displayed geocoded data in the InfoWindow from the full address to simply the Street Number and Street Name makes the InfoWindow more manageable as well as better fit the use case.  However, it also showed that fixing the empty snippet line is going to be a priority as well.  This is documented in --> https://github.com/mapbox/mapbox-gl-native/issues/4959

bleege:
[...]

PR #5346, opened 2016-06-14
4153 circle pulse animation

tobrun:
@bleege 

This PR contains code on how to do a pulse animation when you hover the camera over a marker:

ezgif com-video-to-gif 34

[Here](https://gist.github.com/tobrun/394d14b1b1bd6eaa7edcccef4b6fae80) is also a simplified example:

ezgif com-video-to-gif 33

cc @cammace @ivovandongen @zugaldia


bleege:
Thanks for putting this together @tobrun.  I'm not sure how much of it we'll be able to use as the use case is for the `MyLocationView` in #5282 and not the generic Marker / Annotations.

zugaldia:
@bleege @tobrun Agreed that the `MyLocationView` usecase is different from the marker view usecase. At the same time, I can see how devs could be interested in either depending on their apps needs. Can we keep both approaches around for reference (as different activities, or as two settings in an new activity)?

I think this should probably be kept separate from the geocoder activity.

tobrun:
Closing, code was only to showcase this functionality.

Issue #1555, opened 2015-05-14
headless tests on travis don't antialias fills


jfirebaugh:
Actual:
![image](https://cloud.githubusercontent.com/assets/98601/7624280/20dc7742-f99b-11e4-908a-b8cfc99845b9.png)

Expected:
![image](https://cloud.githubusercontent.com/assets/98601/7624277/1838bdbc-f99b-11e4-83a5-152d48947067.png)

jfirebaugh:
This appears to be an issue with `glLineWidth` / `GL_LINES`. If I switch this test case to just draw the outline, I get the following result locally:

![image](https://cloud.githubusercontent.com/assets/98601/7695601/b3d9e976-fda6-11e4-9681-78db0237a7f6.png)

And this result on linux:

![image](https://cloud.githubusercontent.com/assets/98601/7695602/bcfd9e08-fda6-11e4-9f4d-a587a87b7cef.png)

Fiddling with the line width on linux doesn't seem to produce any better results. E.g., here line width = 10:

![image](https://cloud.githubusercontent.com/assets/98601/7695617/e02d314a-fda6-11e4-9647-b4366963cbe3.png)

`glGet` reports `GL_LINE_WIDTH_RANGE` = 1 - 10 and `GL_LINE_WIDTH_GRANULARITY` = 0.1.

@kkaefer Any ideas?

jfirebaugh:
Here's a hint: if I remove the multiplication by `alpha` [here](https://github.com/mapbox/mapbox-gl-native/blob/abafb52f37beb5659efc2105ccd1568e1f754898/src/mbgl/shader/outline.fragment.glsl#L8-8), it looks like this:

![image](https://cloud.githubusercontent.com/assets/98601/7695671/d815c8ea-fda7-11e4-9fb0-10034af34c9b.png)

It looks like perhaps every other segment is simply not being drawn.

kkaefer:
I can reproduce this behavior on Ubuntu 15.05, but my output looks different:

![actual](https://cloud.githubusercontent.com/assets/52399/7723560/feb3ec82-fee9-11e4-9a03-8ddaccd1f29e.png)

kkaefer:
Playing with this a bit further, I'm seeing only mostly horizontal lines:

![image](https://cloud.githubusercontent.com/assets/52399/7731190/120d50ea-ff1f-11e4-8a99-1c99de1796ad.png)

kkaefer:
[...]

PR #5357, opened 2016-06-14
[core] Separate pattern size from draw-time variables

brunoabinader:
This is a step towards implementing data driven fill styles.

Part of https://github.com/mapbox/mapbox-gl-shaders/issues/1.

:eyes: @jfirebaugh @kkaefer

### DIFF
--- a/include/mbgl/util/convert.hpp
+++ b/include/mbgl/util/convert.hpp
@@ -1,0 +1,17 @@
+#include <array>
+#include <type_traits>
+#include <utility>
+
+namespace mbgl {
+namespace util {
+
+template<typename To, typename From, std::size_t Size,
+         typename = std::enable_if_t<std::is_convertible<From, To>::value>>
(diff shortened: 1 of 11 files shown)


jfirebaugh:
Rather than ignoring the test-suite regressions (https://github.com/mapbox/mapbox-gl-test-suite/commit/f185b8874a7c6633a8c4663749d1f2cd38806c56, https://github.com/mapbox/mapbox-gl-test-suite/commit/f83ab256e508daa9d656cf698687c6a91f08641d), please investigate why they regressed and address the underlying issue. Since these tests were previously passing, it's either something introduced by this change, or a latent issue that was exposed by it. Either way, we should understand it before merging.

brunoabinader:
@jfirebaugh I'm now getting `diff: 0` locally for the two remaining fill tests. I'm running these on OS X, so I don't see any other causes besides https://github.com/mapbox/mapbox-gl-native/issues/1555 for these.

PR #5362, opened 2016-06-15
[ios] fixes #5036 draggable annotation views

### LINKED ISSUE #5036: Draggable annotation views

frederoni:
Fixes #5036 

Setting MGLAnnotationView's `draggable` property to YES will make it draggable.
In order to make it stick to the location where it was dropped, you have to implement;
`-mapView:didDragAnnotationView:toLocation:`
and update the `coordinate` property of the associated annotation.

### DIFF
--- a/platform/ios/src/MGLAnnotationView.mm
+++ b/platform/ios/src/MGLAnnotationView.mm
@@ -1,15 +1,19 @@
 #import "MGLAnnotationView.h"
 #import "MGLAnnotationView_Private.h"
+#import "MGLAnnotation.h"
 #import "MGLMapView_Internal.h"
 
 #import "NSBundle+MGLAdditions.h"
 
 #include <mbgl/util/constants.hpp>
 
-@interface MGLAnnotationView ()
+@interface MGLAnnotationView () <UIGestureRecognizerDelegate>
 
 @property (nonatomic, readwrite, nullable) NSString *reuseIdentifier;
 @property (nonatomic, readwrite, nullable) id <MGLAnnotation> annotation;
+@property (nonatomic, weak) UIPanGestureRecognizer *panGestureRecognizer;
+@property (nonatomic, weak) UILongPressGestureRecognizer *longPressRecognizer;
+@property (nonatomic, weak) MGLMapView *mapView;
 
 @end
 
@@ -62,6 +66,11 @@
     
     [super setCenter:center];
     
+    // Omit applying a new transformation while the view is being dragged.
+    if (self.dragState == MGLAnnotationViewDragStateDragging) {
+        return;
+    }
+    
     if (self.flat)
     {
         [self updatePitch:pitch];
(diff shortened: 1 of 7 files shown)


platform/ios/src/MGLAnnotationView.mm
```diff
@@ -108,6 +117,122 @@ - (void)updateScaleForPitch:(CGFloat)pitch
...
+    }
+}
+
+- (void)enableDrag
+{
+    if (!_longPressRecognizer)
+    {
+        UILongPressGestureRecognizer *recognizer = [[UILongPressGestureRecognizer alloc] initWithTarget:self action:@selector(handleLongPress:)];
+        recognizer.delegate = self;
+        [self addGestureRecognizer:recognizer];
```
1ec5: If you enable dragging, disable it, then enable it again, the gesture recognizer is never re-added to the map view.

platform/ios/src/MGLAnnotationView.mm
```diff
@@ -108,6 +117,122 @@ - (void)updateScaleForPitch:(CGFloat)pitch
...
+        case UIGestureRecognizerStateChanged:
+            self.dragState = MGLAnnotationViewDragStateDragging;
+            break;
+        case UIGestureRecognizerStateCancelled:
+            self.dragState = MGLAnnotationViewDragStateCanceling;
+            break;
+        case UIGestureRecognizerStateEnded:
+            self.dragState = MGLAnnotationViewDragStateEnding;
+            break;
+        default:
```
1ec5: Nit: omit the default case (replacing it with UIGestureRecognizerStatePossible) so we can get a compiler warning here if another enum value is ever added.

platform/ios/src/MGLAnnotationView.mm
```diff
@@ -108,6 +117,122 @@ - (void)updateScaleForPitch:(CGFloat)pitch
...
+    if (dragState == MGLAnnotationViewDragStateStarting)
+    {
+        [self.superview bringSubviewToFront:self];
+    }
+    
+    if (dragState == MGLAnnotationViewDragStateEnding)
+    {
+        if ([self.mapView.delegate respondsToSelector:@selector(mapView:didDragAnnotationView:toCoordinate:)])
+        {
+            CLLocationCoordinate2D coordinate = [self.mapView convertPoint:self.center toCoordinateFromView:self.mapView];
```
1ec5: `center` doesn't account for `offset`. Should it?

platform/ios/src/MGLMapViewDelegate.h
```diff
@@ -300,6 +300,18 @@ NS_ASSUME_NONNULL_BEGIN
  */
 - (void)mapView:(MGLMapView *)mapView didDeselectAnnotationView:(MGLAnnotationView *)annotationView;
 
+/**
+ Tells the delegate that one if its annotation views was dragged to a new coordinate.
+ 
+ You should update the coordinate of the associated annotation when this method is called.
```
1ec5: What happens if you don't? Does the annotation go back to the original location automatically?

platform/ios/src/MGLAnnotationView.mm
frederoni: @1ec5 
Just tested and that doesn't seem to be the case.
- `-enableDrag` creates a gesture recognizer and adds it to the view which holds a strong reference to it. (the property holds a weak ref to the recognizer)
- `-disableDrag` removes the gesture recognizer from the view and since the property only holds a weak ref, it gets nil'd out by `[self removeGestureRecognizer:_longPressRecognizer];`
- Next time `-enableDrag` is called, a new gesture recognizer will be created and added to the view because the property is nil.

platform/ios/src/MGLMapViewDelegate.h
frederoni: Correct. Next time `-updateAnnotationViews` is called, it will go back to the original location.

platform/ios/src/MGLAnnotationView.mm
1ec5: Ah, I missed that the ivar was a weak reference. 👍

PR #5343, opened 2016-06-13
Rename OS X SDK to macOS SDK

1ec5:
What’s old is new again!

This change renames as many references to “OS X” as possible to “macOS” in the project structure, build system, and documentation. The Mapbox OS X SDK is now Mapbox macOS SDK, for consistency with the Mapbox iOS SDK. osxapp is now macosapp. I’ve left Mason on “osx” for now, and the SDK identifier is still technically `macosx`. The build folder is now build/macos/ instead of build/osx/.

/cc @jfirebaugh @friedbunny @frederoni

PR #5369, opened 2016-06-16
[ios] Disable telemetry while Low Power Mode is enabled

friedbunny:
Disable (the already extremely power-efficient) telemetry while the user has [Low Power Mode](https://developer.apple.com/library/ios/documentation/Performance/Conceptual/EnergyGuide-iOS/LowPowerMode.html) enabled. This feature is iOS 9+, so some availability checks must be made.

/cc @boundsj @1ec5 @frederoni

### DIFF
--- a/platform/ios/src/MGLMapboxEvents.m
+++ b/platform/ios/src/MGLMapboxEvents.m
@@ -147,10 +147,16 @@ const NSTimeInterval MGLFlushInterval = 180;
 #if TARGET_OS_SIMULATOR
     return NO;
 #else
+    BOOL isLowPowerModeEnabled = NO;
+    if ([NSProcessInfo instancesRespondToSelector:@selector(isLowPowerModeEnabled)]) {
+        isLowPowerModeEnabled = [[NSProcessInfo processInfo] isLowPowerModeEnabled];
+    }
     return ([[NSUserDefaults standardUserDefaults] boolForKey:@"MGLMapboxMetricsEnabled"] &&
-            [[NSUserDefaults standardUserDefaults] integerForKey:@"MGLMapboxAccountType"] == 0);
+            [[NSUserDefaults standardUserDefaults] integerForKey:@"MGLMapboxAccountType"] == 0 &&
+            !isLowPowerModeEnabled);
 #endif
 }
+
 
 - (BOOL)debugLoggingEnabled {
     return (self.canEnableDebugLogging &&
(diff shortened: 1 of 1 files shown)


friedbunny:
@1ec5: [Test failure](https://www.bitrise.io/build/cbb25d3ce1d50284) is due to the macOS renaming not having happened yet on this release branch:

> FATA[23:58:56] Failed to create bitrise config, error: Config (path:platform/macos/bitrise.yml) in not valid: No file found at path: platform/macos/bitrise.yml

platform/ios/src/MGLMapboxEvents.m
```diff
@@ -245,16 +249,28 @@ - (void)userDefaultsDidChange:(NSNotification *)notification {
     });
 }
 
+- (BOOL)isLowPowerModeEnabled {
```
friedbunny: This isn’t a property, but it is named after the getter for a property — should we name it differently?

platform/ios/src/MGLMapboxEvents.m
```diff
@@ -278,16 +286,24 @@ - (void)pauseMetricsCollection {
...
     }
-    
+
     self.paused = NO;
     self.data = [[MGLMapboxEventsData alloc] init];
     
     [self.locationManager startUpdatingLocation];
 }
 
+- (BOOL)isLowPowerModeEnabled {
```
1ec5: This can be a class method. In fact, it may make sense to consolidate this check with `+isEnabled`.

platform/ios/src/MGLMapboxEvents.m
```diff
@@ -252,9 +256,13 @@ - (void)pauseOrResumeMetricsCollectionIfRequired {
         [self pauseMetricsCollection];
         return;
     }
+
+    if ([self isLowPowerModeEnabled]) {
+        [self pauseMetricsCollection];
```
1ec5: Do we resume once the charge reaches a suitable level and Low Power Mode automatically shuts off?

platform/ios/src/MGLMapboxEvents.m
friedbunny: I ran with the suggestion to integrate `isLowPowerModeEnabled` into `+isEnabled` in b9adf31 — much simpler.

platform/ios/src/MGLMapboxEvents.m
friedbunny: Yes, we register for [`NSProcessInfoPowerStateDidChangeNotification`](https://developer.apple.com/reference/foundation/nsprocessinfopowerstatedidchangenotification?language=objc), which is sent when Low Power Mode is enabled or disabled. See [line 215](https://github.com/mapbox/mapbox-gl-native/pull/5369/files#diff-6e8e31cd0ffe52aeda9c8b2671bcea19R215).

When we receive this notification, we call `-pauseOrResumeMetricsCollectionIfRequired` and revalidate whether or not telemetry should be enabled. After b9adf31, the LPM check occurs in `+isEnabled`.

boundsj:
👍

1ec5:
I’ve cherry-picked #5343 into the release branch; the build failure should go away once you rebase.

friedbunny:
Rebased and looking to merge this once tests pass.

Issue #115, opened 2014-03-25
crossfade raster textures


incanus:
When fading in raster tiles, currently we fade from blank, but we should crossfade if there is an existing texture in the screenspace, e.g. when a higher-zoom tile fades in over a lower-zoom overzoomed one.

jfirebaugh:
⇢ #5543

PR #5372, opened 2016-06-16
[core] Update RasterShader to support crossfading

brunoabinader:
So far this has been the trickiest pending task from https://github.com/mapbox/mapbox-gl-shaders/issues/1. I'm afraid this will require help from someone with more GL expertise. I've started updating `RasterShader` according to the new shader code and am assuming that `tile == parentTile` for testing purposes.

Question: If `u_buffer_scale` is always 1 - why do we need to expose this as a uniform? I assume we want to eventually expose this as as data-driven style property?

Part of https://github.com/mapbox/mapbox-gl-native/issues/115.

:eyes: @kkaefer @lucaswoj @ansis @jfirebaugh

### DIFF
--- a/src/mbgl/renderer/painter.hpp
+++ b/src/mbgl/renderer/painter.hpp
@@ -242,6 +242,11 @@ private:
     VertexArrayObject backgroundPatternArray;
     VertexArrayObject backgroundArray;
 
+    VertexArrayObject coveringPlaingOverdrawArray;
+    VertexArrayObject coveringRasterOverdrawArray;
+    VertexArrayObject backgroundPatternOverdrawArray;
+    VertexArrayObject backgroundOverdrawArray;
+
     // Set up the tile boundary lines we're using to draw the tile outlines.
     StaticVertexBuffer tileBorderBuffer = {
         { 0, 0 },
@@ -252,8 +257,6 @@ private:
     };
 
     VertexArrayObject tileBorderArray;
-
-    VertexArrayObject coveringRasterOverdrawArray;
 };
 
 } // namespace mbgl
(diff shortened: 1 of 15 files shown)


brunoabinader:
> ~~Where/How can we populate the a_texture_pos array?~~

Solved by adding a new Buffer `StaticRasterVertexBuffer` and using `GL_TRIANGLE_STRIP` when drawing the arrays 🎉

brunoabinader:
Voilà:
screen shot 2016-06-19 at 8 55 10 pm

Missing: Use `raster-fade-duration` to provide transition between raster tiles.

src/mbgl/renderer/painter.hpp
```diff
@@ -242,6 +242,11 @@ class Painter : private util::noncopyable {
     VertexArrayObject backgroundPatternArray;
     VertexArrayObject backgroundArray;
 
+    VertexArrayObject coveringPlaingOverdrawArray;
```
kkaefer: typo?

brunoabinader: We apply the same _alias_ technique for selecting original/overdraw VAOs, as seen in https://github.com/mapbox/mapbox-gl-native/pull/5372/files/f94b354b53ba2fc5d12aade7a82790965784afd9..d8a10ac509d5f56120eb878b5e330dcb4c7abf1b#diff-bf2327a9287221e6bae1de29dc0c3400R16 for example.

kkaefer: I'm specifically referring to coveringPlain**g**OverdrawArray`

brunoabinader: Fixed in https://github.com/mapbox/mapbox-gl-native/pull/5557.

PR #5373, opened 2016-06-16
[ios] fixes #5036 draggable annotation views

### LINKED ISSUE #5036: Draggable annotation views

frederoni:
Fixes #5036

Setting MGLAnnotationView's `draggable` property to YES will make it draggable.
In order to make it stick to the location where it was dropped, you have to implement;
`-mapView:didDragAnnotationView:toLocation:`
and update the `coordinate` property of the associated annotation.

Made a few changes compared to previous PR which didn't take center offset into account.
- Circle offset is 0,0
- Rectangle offset is "bottom-left"

drag_small

also updated doc in MGLMapViewDelegate, I hope it's more clear now.

### DIFF
--- a/platform/ios/app/MBXAnnotationView.m
+++ b/platform/ios/app/MBXAnnotationView.m
@@ -33,4 +33,32 @@
     self.layer.borderWidth = selected ? 2.0 : 0;
 }
 
+- (void)setDragState:(MGLAnnotationViewDragState)dragState animated:(BOOL)animated
+{
+    [super setDragState:dragState animated:NO];
+    
+    switch (dragState) {
+        case MGLAnnotationViewDragStateNone:
+            break;
+        case MGLAnnotationViewDragStateStarting: {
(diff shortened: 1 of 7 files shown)


1ec5:
The macOS build failed because this branch is based on release-ios-3.3.0, which doesn't have the OS X-to-macOS renaming from #5343. I've added a special Bitrise workflow for that release branch, but macOS builds will continue to fail for new PRs made against that branch.

@boundsj @friedbunny @kkaefer, maybe we should cherry-pick #5343 into the release branch to keep things sane. The macOS builds are important to iOS SDK development because the two SDKs share so much code.

friedbunny:
I'm 👍 on cherry-picking macOS into the release branch.

1ec5:
I’ve cherry-picked #5343 into the release branch; the build failure should go away once you rebase.

PR #5375, opened 2016-06-16
[android] Dynamic update of InfowWindow with InfoWindowAdapter example

### LINKED ISSUE #5353: Make InfoWindowAdapter dynamically updateable
Analog to https://github.com/mapbox/mapbox-gl-native/issues/5237.

Add the possibility to invalidate the `InfoWindowAdapter` so that the content is refreshed on the map. As the `InfoWindowAdapter` is an interface and we don't want to break the api, the easiest way to provide this would probably be add an abstract class extending `InfoWindowAdapter`  with an  `invalidate()` method that may be called so that the infowindow contents is refreshed.

ivovandongen:
Fixes #5353

@zugaldia Correct PR (hopefully), previous was based on master

### DIFF
--- a/platform/android/MapboxGLAndroidSDK/src/main/java/com/mapbox/mapboxsdk/annotations/InfoWindow.java
+++ b/platform/android/MapboxGLAndroidSDK/src/main/java/com/mapbox/mapboxsdk/annotations/InfoWindow.java
@@ -259,9 +259,18 @@ public class InfoWindow {
         View view = mView.get();
         if (mapboxMap != null && marker != null && view != null) {
             mCoordinates = mapboxMap.getProjection().toScreenLocation(marker.getPosition());
-            view.setX(mCoordinates.x + mViewWidthOffset - mMarkerWidthOffset);
+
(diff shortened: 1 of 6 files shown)


platform/android/MapboxGLAndroidSDK/src/main/java/com/mapbox/mapboxsdk/maps/MapboxMap.java
```diff
@@ -1601,7 +1601,10 @@ public void invalidateCustomLayers() {
         mMapView.invalidateCustomLayers();
     }
 
-    MapView getMapView() {
+    /**
+     * Do not use this method, used by Mapbox internally
+     */
+    public MapView getMapView() {
```
zugaldia: @ivovandongen This seems unnecessary for this PR, I'm making the visibility private again. Is it needed somewhere else?

zugaldia:
@ivovandongen I made a couple of smaller changes, otherwise it's all looking good. Great work, getting it ready to merge now.

PR #5376, opened 2016-06-16
5242 animate marker view

### LINKED ISSUE #5242: Animate MarkerView from current to another position
When we currently call `setPosition` on `MarkerView` the position changes immediately. It would be nice to support a similar feature as with `setRotation` and `setAlpha` to perform an Android SDK animation when the position changes. 

### Animating LatLng values

One approach for this is what we use in `AnimatedMarkerActivity`:

```
  final Marker marker = mapboxMap.addMarker(new MarkerOptions().position(brussels));
  ValueAnimator markerAnimator = ValueAnimator.ofObject(new LatLngEvaluator(), (Object[]) new LatLng[]{brussels, washington});
   markerAnimator.setDuration(5000);
  markerAnimator.setRepeatCount(ValueAnimator.INFINITE);
  markerAnimator.setRepeatMode(ValueAnimator.REVERSE);
  markerAnimator.setInterpolator(new AccelerateDecelerateInterpolator());
  markerAnimator.addUpdateListener(new ValueAnimator.AnimatorUpdateListener() {
          @Override
          public void onAnimationUpdate(ValueAnimator animation) {
              if (marker != null) {
                 marker.setPosition((LatLng) animation.getAnimatedValue());
              }
           }
  });
  markerAnimator.start();

  private class LatLngEvaluator implements TypeEvaluator<LatLng> {

        private LatLng mLatLng = new LatLng();

        @Override
        public LatLng evaluate(float fraction, LatLng startValue, LatLng endValue) {
            mLatLng.setLatitude(startValue.getLatitude() + (endValue.getLatitude() - startValue.getLatitude()) * fraction);
            mLatLng.setLongitude(startValue.getLongitude() + (endValue.getLongitude() - startValue.getLongitude()) * fraction);
            return mLatLng;
        }
    }

```

While this works, it's not resulting in an smooth animation because the marker needs to go through jni for each update. I have been profiling the code the related code to make it run smooth but at this point, after doing a couple of improvements, I'm not sure if we will ever get the right UX.

### Animating Pixel Coordinate values

Another approach is to use pixel coordinates instead of the `LatLng` coordinates above. This has the benefit that we don't need to pass through JNI to get the correct pixel coordinate value for each animation update but has the downside that we need to manage camera changes manually.

tobrun:
closes #5242

### DIFF
--- a/platform/android/MapboxGLAndroidSDK/src/main/java/com/mapbox/mapboxsdk/annotations/MarkerView.java
+++ b/platform/android/MapboxGLAndroidSDK/src/main/java/com/mapbox/mapboxsdk/annotations/MarkerView.java
@@ -1,12 +1,16 @@
 package com.mapbox.mapboxsdk.annotations;
 
+import android.animation.AnimatorSet;
 import android.graphics.Bitmap;
+import android.graphics.PointF;
 import android.support.annotation.FloatRange;
 import android.support.annotation.NonNull;
 import android.support.annotation.Nullable;
 import android.view.View;
+import android.view.animation.AnimationUtils;
 
 import com.mapbox.mapboxsdk.constants.MapboxConstants;
+import com.mapbox.mapboxsdk.geometry.LatLng;
 import com.mapbox.mapboxsdk.maps.MapboxMap;
 
 /**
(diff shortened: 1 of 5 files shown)


tobrun:
CI approved, except for renmaing macOS issue, going to rebase and merge

PR #5385, opened 2016-06-17
Optimize animated annotations

jfirebaugh:
I added an "Add Animated Annotation" menu item in the macOS test app, and then profiled it, which revealed several gross inefficiencies. After fixing those, the profile is heavily dominated by rendering performance -- updating the annotation is less than 3% of runtime.

@tobrun, can you test this out on Android?

/cc @1ec5 for macOS test app changes.


platform/macos/app/MapDocument.m
```diff
@@ -405,6 +407,25 @@ - (IBAction)drawPolygonAndPolyLineAnnotations:(id)sender {
     [self.mapView addAnnotation:line];
 }
 
+- (IBAction)drawAnimatedAnnotation:(id)sender {
+    DroppedPinAnnotation *annotation = [[DroppedPinAnnotation alloc] init];
+    [self.mapView addAnnotation:annotation];
+
+    [NSTimer scheduledTimerWithTimeInterval:1/60
```
1ec5: The time interval is 0 since no floats are involved.

platform/macos/app/MapDocument.m
```diff
@@ -5,6 +5,8 @@
 
 #import <Mapbox/Mapbox.h>
 
+#import <math.h>
```
1ec5: This is unnecessary in Objective-C.

tobrun:
@jfirebaugh this a major improvement, we aren't dropping any frames anymore with this fix.
https://www.dropbox.com/s/z2vmjhldi4pndgl/device-2016-06-17-105618.mp4?dl=0

tobrun:
I'm having some issues bringing this from `master` to the `release-v4.1.0`. We have diverged a lot and I'm not sure which core pieces I need to pick and which to ignore.

To test, I reversed the approach and cherry picked the animating car example from `release-v4.1.0` into this branch:

ezgif com-video-to-gif 8

Here is also an example with +100 markers:

ezgif com-video-to-gif 7

cc @cammace @ivovandongen @pveugen @zugaldia

jfirebaugh:
Ported by hand to the Android release branch in [`animated-annotation-android-4.1.0`](https://github.com/mapbox/mapbox-gl-native/tree/animated-annotation-android-4.1.0). @zugaldia can you test?

platform/macos/app/MapDocument.m
jfirebaugh: It was copied from `dropManyPins:`. I'll fix both occurrences.

zugaldia:
@jfirebaugh This is the `AnimatedMarkerActivity` _without your changes_ in the current release branch:

without

This is the same activity brought to your ` animated-annotation-android-4.1.0` branch:

with

Bottom line, your port is improving substantially the performance. Let's merge it into the release branch 👍 .

zugaldia:
@jfirebaugh Bringing `animated-annotation-android-4.1.0` over to the release branch via https://github.com/mapbox/mapbox-gl-native/pull/5392.

cammace:
getting a bunch of 
```
06-17 15:42:25.385 718-718/com.mapbox.mapboxsdk.testapp D/mbgl: [JNI]: nativeUpdateMarker
``` 
output in logcat which makes it impossible to read without filtering. Anyway we could remove this from printing?

tobrun:
@cammace you can disable this type of logging if you build with BUILDTYPE=Release

PR #5389, opened 2016-06-17
[core] Do not use pthread_getname_np before Glibc 2.11

brunoabinader:
Follows https://github.com/mapbox/mapbox-gl-native/blob/master/src/mbgl/util/thread.hpp#L98-L104.

:eyes: @kkaefer @tobrun @tmpsantos


jfirebaugh:
Want to factor this to `static Thread::getCurrentThreadName()` or similar?

brunoabinader:
@jfirebaugh sounds good, same applies for when setting a thread name.

PR #5400, opened 2016-06-18
[core] Factor out thread name setter/getter

brunoabinader:
Follow-up to #5389. Unfortunately Apple headers defines a different `pthread_getname_np` from Glibc, so creating a common `#define` would eventually add more code lines than the contrary. I left them out of `Thread` because it is a template class.

:eyes: @jfirebaugh @kkaefer

### DIFF
--- a/src/mbgl/util/thread.hpp
+++ b/src/mbgl/util/thread.hpp
@@ -87,6 +87,29 @@ private:
     RunLoop* loop = nullptr;
 };
 
+inline std::string getCurrentThreadName() {
+    char name[32] = "unknown";
+#if defined(__APPLE__)
+    pthread_getname_np(pthread_self(), name, sizeof(name));
+#elif defined(__GLIBC__) && defined(__GLIBC_PREREQ)
+#if __GLIBC_PREREQ(2, 12)
+    pthread_getname_np(pthread_self(), name, sizeof(name));
+#endif
+#endif
+    return name;
+}
+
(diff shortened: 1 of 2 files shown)


src/mbgl/util/thread.hpp
```diff
@@ -87,6 +87,34 @@ class Thread {
     RunLoop* loop = nullptr;
 };
 
+inline std::string getCurrentThreadName() {
+#if defined(__APPLE__)
+    char name[32];
+    pthread_getname_np(pthread_self(), name, sizeof(name));
+    return name;
+#elif defined(__GLIBC__) && defined(__GLIBC_PREREQ)
+#if __GLIBC_PREREQ(2, 12)
```
jfirebaugh: Shouldn't this be part of the `elif` condition on the previous line? Otherwise the function will be missing a `return` statement in the case that you are compiling with a glibc that doesn't meet the prerequisite.

brunoabinader: True about the missing `return` statement, but we can't join lines 95 and 96 together because otherwise `__GLIBC_PREREQ` is not always defined and causes the preprocessor to abort with an error.

brunoabinader: @jfirebaugh I moved `name` out of the ifdefs so it'll return `"unknown"` by default.

kkaefer:
👍

Issue #5371, opened 2016-06-16
Use overdraw inspector for wireframe mode


brunoabinader:
Our [native implementation for wireframe mode](https://github.com/mapbox/mapbox-gl-js/pull/2489) differs from js significantly. While native implementation modifies the fragment color before passing it to shader code, js uses a `OVERDRAW_INSPECTOR` define that forces fragment color to e.g. black for fills. While this means a couple extra shader programs to compute, this would contribute to clean up native's Painter code.

Ref: https://github.com/mapbox/mapbox-gl-shaders/issues/1

:eyes: @jfirebaugh @kkaefer @lucaswoj

PR #5403, opened 2016-06-18
[core] Refactor wireframe to match JS overdraw

brunoabinader:
This PR:
- Adds a new `BlendColor` GL configuration value
- Generates an additional overdraw fragment shader & program for each shader type
- Refactors the wireframe mode to behave like JS _overdraw_ mode, where each render pass over the same place increases the color by 1/8 from black to white.

Fixes https://github.com/mapbox/mapbox-gl-native/issues/5371.

Snapshot:
screen shot 2016-06-18 at 7 33 39 pm

Notes:
- Although similar, it seems that `Wireframe` and `Overdraw` serves different purposes - we could add a new debug mode instead of replacing `wireframe` implementation.
- Do we really need to overdraw SDFs?

:eyes: @mapbox/gl

### DIFF
--- a/platform/ios/src/MGLMapView.h
+++ b/platform/ios/src/MGLMapView.h
@@ -43,9 +43,8 @@ typedef NS_OPTIONS(NSUInteger, MGLMapDebugMaskOptions) {
     /** Edges of glyphs and symbols are shown as faint, green lines to help
         diagnose collision and label placement issues. */
     MGLMapDebugCollisionBoxesMask = 1 << 4,
-    /** Line widths, backgrounds, and fill colors are ignored to create a
-        wireframe effect. */
-    MGLMapDebugWireframesMask = 1 << 5,
+    /** Overdraw inspector. */
+    MGLMapDebugOverdrawsMask = 1 << 5,
 };
 
 /**
--- a/include/mbgl/gl/gl_values.hpp
+++ b/include/mbgl/gl/gl_values.hpp
@@ -214,6 +214,19 @@ constexpr bool operator!=(const BlendFunc::Type& a, const BlendFunc::Type& b) {
     return a.sfactor != b.sfactor || a.dfactor != b.dfactor;
 }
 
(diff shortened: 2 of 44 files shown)


brunoabinader:
I suspect the `debug/overdraw` render test differs between -js and -native because in -native we optimize the background layer via `glClear`. However, I'm yet to find a proper solution for that.

brunoabinader:
When 1) disabling the `glClear` optimization for the bottommost solid background layer and 2) not resetting the blend function in the translucent render pass we are now getting 1:1 overdraw results between -js and -native 🎉:
screen shot 2016-06-19 at 6 19 12 pm

kkaefer:
Awesome!

platform/ios/src/MGLMapView.h
```diff
@@ -43,9 +43,8 @@ typedef NS_OPTIONS(NSUInteger, MGLMapDebugMaskOptions) {
     /** Edges of glyphs and symbols are shown as faint, green lines to help
         diagnose collision and label placement issues. */
     MGLMapDebugCollisionBoxesMask = 1 << 4,
-    /** Line widths, backgrounds, and fill colors are ignored to create a
-        wireframe effect. */
-    MGLMapDebugWireframesMask = 1 << 5,
+    /** Overdraw inspector. */
+    MGLMapDebugOverdrawsMask = 1 << 5,
```
1ec5: This enum value needs a more descriptive name, and the comment is somewhat nonsensical to an iOS developer.

1ec5:
@friedbunny, @brunoabinader renamed the public enum value, so we either need to revert the Objective-C changes or backport them to the release branch. Otherwise, the release after 3.3.0 will break backwards compatibility.

friedbunny:
@1ec5 I suppose just porting the Obj-C changes without including the refactored implementation itself is a thing we can do. Functionality would change after v3.3.0, but the API would remain the same.

PR #5404, opened 2016-06-18
[core] Remove redundant inline syntax

### LINKED ISSUE #5352: Remove redundant inline syntax
Most (if not all) our current `inline` usage occurs while defining a body within a class definition or together with `constexpr` - which are all considered redundant. We should get rid of those.

While a `clang-tidy` check for this is being developed - see http://reviews.llvm.org/D18914 - we should manually get rid of them for now.

/cc @jfirebaugh

brunoabinader:
- Removed all `inline` usage from functions implemented inside a class definition
- Replaced `inline` with `constexpr` for non-member functions defined in headers whenever possible

The remaining `inline` usage in the code is either a non-member function defined in header that uses a _non-constexpr function_ e.g. `mbgl::util::now()` or a member function defined in a header, but outside of the class definition e.g. `*TileID` functions in order to comply with the [one definition rule](https://en.wikipedia.org/wiki/One_Definition_Rule).

Fixes #5352.

:eyes: @jfirebaugh @kkaefer @tmpsantos @1ec5

### DIFF
--- a/include/mbgl/gl/gl.hpp
+++ b/include/mbgl/gl/gl.hpp
@@ -60,14 +60,14 @@ namespace gl {
 #endif
     
 struct Error : ::std::runtime_error {
-    inline Error(GLenum err, const std::string &msg) : ::std::runtime_error(msg), code(err) {};
+    Error(GLenum err, const std::string &msg) : ::std::runtime_error(msg), code(err) {};
     const GLenum code;
 };
 
 void checkError(const char *cmd, const char *file, int line);
 
 #if defined(DEBUG)
-#define MBGL_CHECK_ERROR(cmd) ([&]() { struct __MBGL_C_E { inline ~__MBGL_C_E() { ::mbgl::gl::checkError(#cmd, __FILE__, __LINE__); } } __MBGL_C_E; return cmd; }())
+#define MBGL_CHECK_ERROR(cmd) ([&]() { struct __MBGL_C_E { ~__MBGL_C_E() { ::mbgl::gl::checkError(#cmd, __FILE__, __LINE__); } } __MBGL_C_E; return cmd; }())
 #else
 #define MBGL_CHECK_ERROR(cmd) (cmd)
 #endif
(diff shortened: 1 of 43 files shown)


brunoabinader:
Dang, Clang doesn't complain but GCC complains about using `constexpr` in `Layer::is<T>()` functions. because `Layer`'s destructor is not trivial. Unfortunately we cannot get rid of it being virtual because we store layers as `unique_ptr<Layer>` inside the style. So one more exception for `inline` at least until GCC relaxes its `constexpr` constraints the same way Clang does.

platform/ios/uitest/OHHTTPStubs/OHHTTPStubs/Pods/AFNetworking/AFNetworking/AFURLConnectionOperation.m
```diff
@@ -74,7 +74,7 @@ static dispatch_queue_t url_request_operation_completion_queue() {
 typedef NSCachedURLResponse * (^AFURLConnectionOperationCacheResponseBlock)(NSURLConnection *connection, NSCachedURLResponse *cachedResponse);
 typedef NSURLRequest * (^AFURLConnectionOperationRedirectResponseBlock)(NSURLConnection *connection, NSURLRequest *request, NSURLResponse *redirectResponse);
 
-static inline NSString * AFKeyPathFromOperationState(AFOperationState state) {
+static NSString * AFKeyPathFromOperationState(AFOperationState state) {
```
1ec5: AFNetworking is a third-party library being pulled in as part of OHHTTPStubs, another third-party library. Let's avoid changes to the OHHTTPStubs directory; if it is necessary to make these changes in Objective-C code, let's contribute it upstream.

src/mbgl/style/tile_source_impl.hpp
```diff
@@ -18,7 +18,7 @@ namespace style {
 */
 class TileSourceImpl : public Source::Impl {
 public:
-    // A tile source can either specify a URL to TileJSON, or inline TileJSON.
+    // A tile source can either specify a URL to TileJSON, or TileJSON.
```
jfirebaugh: Restore the "inline" in the comment here.

jfirebaugh:
👍

PR #5411, opened 2016-06-20
[build] Added 'lint' Make target

### LINKED ISSUE #5342: Enforce code style with clang-format
We've [re-enabled clang-tidy checks](https://github.com/mapbox/mapbox-gl-native/pull/5325) as part of establishing good practices in our code changes. Keeping in line with that, we should start using `clang-format` - it automatically formats C++ code according to a set of configuration options - we actually maintain a [.clang-format](https://github.com/mapbox/mapbox-gl-native/blob/master/.clang-format) config already.

As @kkaefer suggested, we should have a bot that:

1. Runs `clang-format -i` over `mapbox-gl-native` codebase;
2. Runs `git diff`, then fails if there are any differences (showing diff output for reference).

:eyes: @mapbox/gl

brunoabinader:
Introduce a new Makefile target named `lint` - it is meant to run both `clang-tidy` and `clang-format` checks only on modified code from your local branch.

Example:
1. Rachel has a local branch named `new-feature`, which tracks `origin/master`.
2. Rachel commits all her code changes and wants to check lints and code formatting issues.
3. Rachel runs `make lint` and if there any changes made by either `clang-tidy` or `clang-format`, Rachel reviews and applies the changes before submitting upstream.

Caveats:
- This script assumes that your remote repository is tracked as `origin`.
- We currently have a `.clang-format` in GL native, but we might want to share it with https://github.com/mapbox/cpp-bootstrap/pull/7.

Fixes #5342.

:eyes: @mapbox/gl


brunoabinader:
Note: We might want to disable `clang-format` checks before doing a one-time run on all source code.

brunoabinader:
Per conversation with @tmpsantos - I'll rename the current `check` target to `coverage`, and use `check` instead of `lint` for this new proposed target.

Makefile
```diff
@@ -338,15 +338,29 @@ compdb-macos: platform/macos/platform.gyp $(MACOS_OUTPUT_PATH)/config.gypi
...
+clang-tools-macos:
 	if test -z $(CLANG_TIDY); then .mason/mason install clang-tidy 3.8.0; fi
+	if test -z $(CLANG_FORMAT); then .mason/mason install clang-format 3.8.0; fi
 	deps/ninja/ninja-macos -C $(MACOS_OUTPUT_PATH)/$(BUILDTYPE) headers
-	scripts/clang-tidy.sh $(MACOS_OUTPUT_PATH)/$(BUILDTYPE)
+
+tidy-macos: clang-tools-macos
+	scripts/clang-tools.sh $(MACOS_OUTPUT_PATH)/$(BUILDTYPE)
+
+lint: compdb lint-$(BUILD_PLATFORM)
```
tmpsantos: if `compdb` is not a dep of `lint-linux`, running `make lint-linux` could fail no?

brunoabinader: We shouldn't need to call `make lint-linux` directly - `make lint` is smart enough to trigger it's platform-specific variant.

PR #5416, opened 2016-06-20
Rename overdraw to overdraw visualization

1ec5:
A verb doesn’t make sense in this enum value. Besides, overdrawing is the bug that this effect exposes, not the effect itself. Instead, use a descriptive, Cocoa-compliant word for this effect and describe it in detail in the documentation comment. The core implementation continues to use the same term as GL JS.

/ref #5403
/cc @brunoabinader @friedbunny

### DIFF
--- a/platform/ios/CHANGELOG.md
+++ b/platform/ios/CHANGELOG.md
@@ -4,6 +4,8 @@ Mapbox welcomes participation and contributions from everyone.  Please read [CON
 
 ## master
 
+## 3.3.0
+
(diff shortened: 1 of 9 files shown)


1ec5:
This change and #5403 will need to be cherry-picked into the release-ios-v3.3.0 branch to avoid flip-flopping on the option's name in public API.

jfirebaugh:
We already use the term "X-Ray" to refer to [a different drawing mode](https://www.mapbox.com/blog/turbo-xray-mode/). Let's pick something different.

1ec5:
Oh, right! Maybe "AdditiveFills" or something along those lines?

1ec5:
Since X-ray is taken, let’s use “overdraw visualization” instead, since that appears to be an actual term.

1ec5:
Cherry-picked to the release-ios-v.3.3.0 branch in f5f39c8b3adfed4c4c102dca5d5be4d6cfc0b69f. I avoided cherry-picking #5403 due to the change it made to the test suite version pin.

Issue #3048, opened 2015-11-16
Benchmark OkHTTP vs curl


ljbade:
From #2400 

@mb12 would like us to benchmark the difference between the two libraries

/cc @tobrun @zugaldia @bleege

zugaldia:
Just for background purposes, this is what @mb12 suggests:

> Is it possible to measure the performance overhead of OkHttp as well? In this case we will be copying 25K to 100K+ bytes for each tile from java to native using JNI versus the older one where this copy and the additional overhead of managed code(GC, etc.) is missing.

PR #5417, opened 2016-06-21
[core] Replace boost iostream with std::{i,s}stream

### LINKED ISSUE #4915: Remove boost iostreams dependency
jpeg_reader.cpp and png_reader.cpp use boost iostreams for simple buffer reads. Bringing in boost requires some warning-disabling boilerplate and even then causes a bunch of compile spew about deprecated headers.

It looks like the use of boost can be replaced with a simple buffer and current location; let's do so.

brunoabinader:
Fixes #4915.

:eyes: @jfirebaugh @springmeyer

### DIFF
--- a/platform/default/png_reader.cpp
+++ b/platform/default/png_reader.cpp
@@ -2,15 +2,8 @@
 #include <mbgl/util/premultiply.hpp>
 #include <mbgl/platform/log.hpp>
 
-#pragma GCC diagnostic push
-#pragma GCC diagnostic ignored "-Wunknown-pragmas"
-#pragma GCC diagnostic ignored "-Wunused-local-typedefs"
-#pragma GCC diagnostic ignored "-Wshadow"
-#include <boost/iostreams/stream.hpp>
-#pragma GCC diagnostic pop
-
-#include <boost/iostreams/device/file.hpp>
-#include <boost/iostreams/device/array.hpp>
+#include <istream>
+#include <sstream>
 
 extern "C"
 {
@@ -18,9 +11,6 @@ extern "C"
 }
 
 namespace mbgl {
-
-using source_type = boost::iostreams::array_source;
-using input_stream = boost::iostreams::stream<source_type>;
 
 static void user_error_fn(png_structp, png_const_charp error_msg) {
     throw std::runtime_error(std::string("failed to read invalid png: '") + error_msg + "'");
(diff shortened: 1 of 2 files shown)


springmeyer:
👍

jfirebaugh:
🙇

platform/default/png_reader.cpp
```diff
@@ -54,8 +44,8 @@ struct png_struct_guard {
 };
 
 PremultipliedImage decodePNG(const uint8_t* data, size_t size) {
-    source_type source(reinterpret_cast<const char*>(data), size);
-    input_stream stream(source);
+    std::istringstream source(std::string(reinterpret_cast<const char*>(data), size));
```
artemp: ^ creates a new std::string and copies data!

platform/default/png_reader.cpp
```diff
@@ -54,8 +44,8 @@ struct png_struct_guard {
...
+    std::istream stream(source.rdbuf());
```
artemp: hmm and again ^.

artemp:
Better solution is to implement custom stream buffer : https://artofcode.wordpress.com/2010/12/12/deriving-from-stdstreambuf/
/cc @brunoabinader @jfirebaugh @springmeyer

artemp:
@brunoabinader  This is what II did in mapnik https://github.com/mapnik/mapnik/commit/2e8c0d36c2237f2815d8004c1b96bad909056eb9
Note: for `tiff_reader` to work correctly stream buffer must implement 
`c++
pos_type pubseekoff( off_type off, ios_base::seekdir dir,
                     ios_base::openmode which = ios_base::in | ios_base::out );
```
http://en.cppreference.com/w/cpp/io/basic_streambuf/pubseekoff

/cc @springmeyer @jfirebaugh

jfirebaugh:
@artemp I was wondering about that, thanks for the links.

mb12:
@artemp and @jfirebaugh Is the concern here that duplicate copy of image data in std::string can cause performance issue? This may not be perceptible to the end user at all esp for Vector maps where this code will kick in only once when loading the sprite. #3048 is more likely to be a performance bottleneck arising due to copying data multiple times (But it has not been measured).

When we switched from Curl to OkHttp for network access, it introduced multiple copies of tile data as it moves across JNI boundary. (Issue #3048).

https://github.com/mapbox/mapbox-gl-native/issues/3048

brunoabinader:
Thank you @artemp - I wonder if we could add `char_array_buffer` on its own header-only library so we could share code and avoid duplication? /cc @jfirebaugh

PR #5436, opened 2016-06-22
Remove TexturePool dependency from Raster

kkaefer:
The `Raster` object currently has a construction-time dependency on `TexturePool`, but all we do is store the reference until we need it when calling `.bind()`. Instead, we can just pass it then.


kkaefer:
@brunoabinader 👀

brunoabinader:
👍

PR #5453, opened 2016-06-23
5450 fix locational runtime permissions

### LINKED ISSUE #5450: Fix request permission in feature activities related to location
Currently most activities aren't incorporating runtime permissions correctly. We need to migrate the code found in `MyLocationViewTintActivity` to the other feature Activities that contain locational updates.

tobrun:
Closes #5450

### DIFF
--- a/platform/android/MapboxGLAndroidSDKTestApp/src/main/java/com/mapbox/mapboxsdk/testapp/activity/FeatureOverviewActivity.java
+++ b/platform/android/MapboxGLAndroidSDKTestApp/src/main/java/com/mapbox/mapboxsdk/testapp/activity/FeatureOverviewActivity.java
@@ -1,5 +1,6 @@
 package com.mapbox.mapboxsdk.testapp.activity;
 
+import android.Manifest;
 import android.content.ComponentName;
 import android.content.Intent;
 import android.content.pm.ActivityInfo;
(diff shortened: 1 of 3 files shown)


zugaldia:
@tobrun Instead of adding a check to every activity, why don't we add it once to `FeatureOverviewActivity`? In the case that the user decides to deny access to location, we disable the examples that requires the permission.

tobrun:
@zugaldia that is a great idea. Will change it accordingly.

tobrun:
@zugaldia In e7ce3c0 I'm handling the runtime permissions in `FeatureOverviewActivity`. I'm currently thinking of leaving the other permission checks in the other activities. This will cover the use case of a user disabling the location permissions in settings and returning to the app without them enabled. (this would result in a SecurityException, now it will result in showing the permission dialog again).

zugaldia:
> This will cover the use case of a user disabling the location permissions in settings and returning to the app without them enabled.

@tobrun Unlikely but certainly possible. Maybe we could extract some utils in the future, [like we do in MAS](https://github.com/mapbox/mapbox-java/blob/master/libandroid/lib/src/main/java/com/mapbox/services/android/utils/PermissionsUtils.java), to avoid code duplication. Either way, this PR is good to solve the current issue. I'm gonna get this ready to merge.

PR #5452, opened 2016-06-23
Key bindings for macosapp

kkaefer:
We should add a few keybindings to the Debug menu for the macOS application.

Here's my proposal:
- Tile boundaries: **⌘B**
- Tile info: **⌘I**
- Timestamps: **⌘T**
- Collision: **⌘⇧C**
- Overdraw: **⌘⇧O**
- Stencil buffer: **⌘⌥S**
- Color buffer: **⌘⌥C**

### DIFF
--- a/platform/macos/app/Base.lproj/MainMenu.xib
+++ b/platform/macos/app/Base.lproj/MainMenu.xib
@@ -128,13 +128,14 @@
                                 </connections>
                             </menuItem>
                             <menuItem isSeparatorItem="YES" id="aJh-i4-bef"/>
-                            <menuItem title="Page Setup…" keyEquivalent="P" id="qIS-W8-SiK">
-                                <modifierMask key="keyEquivalentModifierMask" shift="YES" command="YES"/>
+                            <menuItem title="Page Setup…" id="qIS-W8-SiK">
+                                <modifierMask key="keyEquivalentModifierMask" control="YES"/>
                                 <connections>
                                     <action selector="runPageLayout:" target="-1" id="Din-rz-gC5"/>
                                 </connections>
                             </menuItem>
-                            <menuItem title="Print…" keyEquivalent="p" id="aTl-1u-JFS">
+                            <menuItem title="Print…" id="aTl-1u-JFS">
+                                <modifierMask key="keyEquivalentModifierMask" control="YES"/>
                                 <connections>
                                     <action selector="print:" target="-1" id="qaZ-4w-aoO"/>
                                 </connections>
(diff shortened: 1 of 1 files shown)


1ec5:
Great idea, but let’s use the command key to make the key bindings more Mac-like. Otherwise, if we ever add a text field to the window, such as for a search bar or “jump to coordinate” field, the key binding will only work when the map view is focused, which can be unpredictable.

kkaefer:
Added the ⌘ key, and in some instances the ⇧ key to avoid collisions with default commands like Copy.

platform/macos/app/Base.lproj/MainMenu.xib
```diff
@@ -128,13 +128,14 @@
...
-                                <modifierMask key="keyEquivalentModifierMask" shift="YES" command="YES"/>
+                            <menuItem title="Page Setup…" id="qIS-W8-SiK">
+                                <modifierMask key="keyEquivalentModifierMask" control="YES"/>
                                 <connections>
                                     <action selector="runPageLayout:" target="-1" id="Din-rz-gC5"/>
                                 </connections>
                             </menuItem>
-                            <menuItem title="Print…" keyEquivalent="p" id="aTl-1u-JFS">
+                            <menuItem title="Print…" id="aTl-1u-JFS">
+                                <modifierMask key="keyEquivalentModifierMask" control="YES"/>
```
1ec5: Why change this from the standard ⌘P shortcut?

platform/macos/app/Base.lproj/MainMenu.xib
```diff
@@ -512,15 +506,15 @@
                                     <action selector="drawAnimatedAnnotation:" target="-1" id="CYM-WB-s97"/>
                                 </connections>
                             </menuItem>
-                            <menuItem title="Remove All Annotations" id="6rC-68-vk0">
-                                <modifierMask key="keyEquivalentModifierMask"/>
+                            <menuItem title="Remove All Annotations" keyEquivalent="q" id="6rC-68-vk0">
+                                <modifierMask key="keyEquivalentModifierMask" option="YES" command="YES"/>
```
1ec5: This conflicts with the standard shortcut for Quit but Keep Windows.

platform/macos/app/Base.lproj/MainMenu.xib
```diff
@@ -494,14 +490,12 @@
                                 </connections>
                             </menuItem>
                             <menuItem isSeparatorItem="YES" id="Sl5-nE-kHd"/>
-                            <menuItem title="Blanket Map With Pins" id="LMZ-oe-Ngh">
-                                <modifierMask key="keyEquivalentModifierMask"/>
+                            <menuItem title="Blanket Map With Pins" keyEquivalent="p" id="LMZ-oe-Ngh">
```
1ec5: This should have more modifier keys.

platform/macos/app/Base.lproj/MainMenu.xib
kkaefer: I figured printing a map from our debug application isn't super common.

1ec5: It may not be, but I don’t really think we should break muscle memory for standard platform shortcuts. There are plenty of other keys available for adding polygons. 😉

1ec5: macosapp is the closest thing we have to an example for how to integrate the SDK into an application, so I’m hoping to [follow the HIG](https://developer.apple.com/library/prerelease/content/documentation/UserExperience/Conceptual/OSXHIGuidelines/InteractivityInput.html#//apple_ref/doc/uid/20000957-CH10-SW4) as much as possible:

> Respect the standard keyboard shortcuts and create app-specific shortcuts for frequently used commands.

platform/macos/app/Base.lproj/MainMenu.xib
```diff
@@ -437,52 +438,47 @@
...
+                                <modifierMask key="keyEquivalentModifierMask" option="YES" command="YES"/>
                                 <connections>
                                     <action selector="showStencilBuffer:" target="-1" id="WkN-t9-Mpv"/>
                                 </connections>
                             </menuItem>
                             <menuItem isSeparatorItem="YES" id="dYw-bb-tr1"/>
-                            <menuItem title="Show Tooltips on Dropped Pins" id="uir-Rx-zmw">
-                                <modifierMask key="keyEquivalentModifierMask"/>
+                            <menuItem title="Show Tooltips on Dropped Pins" keyEquivalent="t" id="uir-Rx-zmw">
+                                <modifierMask key="keyEquivalentModifierMask" option="YES" command="YES"/>
```
1ec5: This conflicts with View ‣ Show Toolbar.

PR #5459, opened 2016-06-23
[android] #5421 - recalculate offset with new anchor

### LINKED ISSUE #5421: [Android] Incorrect Anchor when reusing MarkerViewOptions
I'm experiencing "moving" Markers because the Anchor of a MarkerView seems faulty when reusing once set MarkerViewOptions when the map orientation changes, e.g. while using the locationtracking/bearingtracking

**Platform:**
Android 23
**Mapbox SDK version:**
4.1.0-SNAPSHOT

### Steps to trigger behavior

 1. Set up MarkerViewOptions
 2. Add new MarkerView to MapView
 3. Remove it
 4. Do it in a loop

### Expected behavior
The MarkerView (here they are the location pins) stays at a stable position, it works when i constantly set up new MarkerViewOptions for adding a MarkerView.
stable_markerview

### Actual behavior
In this example the anchor is completly wrong and it is moving. The clips are quite short but you can take the lake or street as reference. I assume the true anchor point does not really change and just sticks with the initial one when setting up new MarkerViewOptions().
unstable_markerview

So for example if you want to reuse MarkerViewOptions, maybe for multiple locations that only differ in their coordinates you would always have to set up new MarkerViewOptions. I haven't tested yet whether this also affects the visibility option that would usually be the better practice than constantly recreating MarkerViews.

tobrun:
closes #5421

### DIFF
--- a/platform/android/MapboxGLAndroidSDK/src/main/java/com/mapbox/mapboxsdk/annotations/MarkerView.java
+++ b/platform/android/MapboxGLAndroidSDK/src/main/java/com/mapbox/mapboxsdk/annotations/MarkerView.java
@@ -86,6 +86,7 @@ public class MarkerView extends Marker {
     public void setAnchor(@FloatRange(from = 0.0, to = 1.0) float u, @FloatRange(from = 0.0, to = 1.0) float v) {
         this.anchorU = u;
         this.anchorV = v;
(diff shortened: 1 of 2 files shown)


zugaldia:
@tobrun Looks good to me. Waiting for CI to merge.

PR #5460, opened 2016-06-23
[ios] Cycle through both UIView and UIImage backed annotations

### LINKED ISSUE #5166: Tapping bunched up point annotation views should cycle among them
When you tap repeatedly near a bunch of closely-spaced or overlapping GL-rendered point annotations, a different annotation is selected each time due to [this cycling logic](https://github.com/mapbox/mapbox-gl-native/blob/8ace0d37d10e8174268f259dec38cf88e2494ef0/platform/ios/src/MGLMapView.mm#L3226-L3235). MapKit does the same with MKAnnotationViews. This behavior is important to ensure that every annotation is reachable by the user. But MGLAnnotationView is a bit more haphazard, because we just select the first annotation view we happen to find in the query.

We should cycle among all the nearby annotations, whether they’re GL annotations, annotation views, or a mixture of both. This probably means we need to stop short-circuiting `-annotationTagAtPoint:persistingResults:` with `-annotationViewAtPoint:`. But until #5165 lands, I’m not sure how we’d be able to reconcile the two methods’ queries. (One of them is limited to the tap point, the other to the entire viewport.)

/cc @boundsj

boundsj:
Fixes #5166 

This changes the way the the calculation for the largest visual annotation representation (stored in
_unionedAnnotationRepresentationSize) is done so that it includes both images and views. The `annotationTagAtPoint` method is then able to find annotation objects that are in a query rect formed from the sizes of all manner of point annotation visual representations.

The `annotationTagAtPoint` method also gains new logic in a conditional to handle annotations backed by MGLAnnotationViews. Similarly to the image case, it filters out disabled annotations and calculates
a centered rect for the view. The hit test logic remains the same but now works correctly for the centered rects of both images and views.

This also fixes a bug where the `largestAnnotationViewSize` was always holding the last annotation view size and not necessarily the largest. This affects when annotation views are put into a reuse queue since the size is used to help determine when an annotation view has gone offscreen.

cc @1ec5 @frederoni

### DIFF
--- a/platform/ios/src/MGLMapView.mm
+++ b/platform/ios/src/MGLMapView.mm
@@ -255,8 +255,9 @@ public:
     NS_MUTABLE_DICTIONARY_OF(NSString *, NS_MUTABLE_ARRAY_OF(MGLAnnotationView *) *) *_annotationViewReuseQueueByIdentifier;
     
     BOOL _userLocationAnnotationIsSelected;
-    /// Size of the rectangle formed by unioning the maximum slop area around every annotation image.
-    CGSize _unionedAnnotationImageSize;
+    /// Size of the rectangle formed by unioning the maximum slop area around every annotation image and annotation image view.
+    CGSize _unionedAnnotationRepresentationSize;
+    CGSize _largestAnnotationViewSize;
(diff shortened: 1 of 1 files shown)


1ec5:
👍

PR #5461, opened 2016-06-23
[ios] Use transparent image for annotations backed by views

### LINKED ISSUE #5210: The presence of image backed annotations should not trigger log messages
Once a GL sprite backed annotation is added to the map, the message `[INFO] {Worker}[Sprite]: Can't find sprite named 'default_marker'` is seen for every tile parse that has a UIView backed annotation in it. The message comes from [this line in sprite_store.cpp](https://github.com/mapbox/mapbox-gl-native/blob/7272d7c84cf6e874b417ecd165f46f26ae7729c9/src/mbgl/sprite/sprite_store.cpp#L145) but the underlying issue is that annotations that are are actually backed by UIViews are not fully understood by core implementation.

boundsj:
The current implementation of view backed annotations mostly relies on the existing annotation model manager and the associated underlying boost query implementations for annotation management. However since, in the case of view backed annotations, the representation of the model is a UIView, an image was not installed. Because the underlying annotation management implementation is not entirely comfortable working without an image, it would report `[INFO] {Worker}[Sprite]: Can't find sprite named 'default_marker'` when annotation models without an image were encountered.

This updates the annotation adding logic to create (or reuse) and install a small, invisible image for each view backed annotation.

This stops the issue noted in #5210 for iOS.

cc @1ec5

### DIFF
--- a/platform/ios/src/MGLMapView.mm
+++ b/platform/ios/src/MGLMapView.mm
@@ -88,6 +88,10 @@ const CLLocationDirection MGLToleranceForSnappingToNorth = 7;
 /// Reuse identifier and file name of the default point annotation image.
 static NSString * const MGLDefaultStyleMarkerSymbolName = @"default_marker";
 
+/// Reuse identifier and file name of the invisible point annotation image used
+/// by annotations that are visually backed by MGLAnnotationView objects
+static NSString * const MGLInvisibleStyleMarkerSymbolName = @"invisible_marker";
+
 /// Prefix that denotes a sprite installed by MGLMapView, to avoid collisions
 /// with style-defined sprites.
 NSString *const MGLAnnotationSpritePrefix = @"com.mapbox.sprites.";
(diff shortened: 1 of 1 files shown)


boundsj:
@1ec5 although I think this is a workable solution for https://github.com/mapbox/mapbox-gl-native/pull/5165 I don't think we actually need #5165 for the ios-v3.3.0 milestone since a solution for https://github.com/mapbox/mapbox-gl-native/issues/5166 landed in https://github.com/mapbox/mapbox-gl-native/pull/5460.

1ec5:
A bit kludgy, but this is exactly what I had in mind. The benefit of leaving mbgl responsible for these annotation views’ positions is that, in the future, it might be possible for annotation views to collide out labels, for instance. (That would require us to vary the spacer image’s size by the view’s size, but that isn’t necessary just yet.)

> although I think this is a workable solution for #5165 I don't think we actually need #5165 for the ios-v3.3.0 milestone since a solution for #5166 landed in #5460.

Does #5165 address any other issues like #5151? If not, then it’s fine to leave that for 3.4.0.

platform/ios/src/MGLMapView.mm
```diff
@@ -2962,6 +2973,23 @@ - (MGLAnnotationImage *)defaultAnnotationImage
     return annotationImage;
 }
 
+- (MGLAnnotationImage *)invisibleAnnotationImage
+{
+    MGLAnnotationImage *annotationImage = [self dequeueReusableAnnotationImageWithIdentifier:MGLInvisibleStyleMarkerSymbolName];
+    
+    if (!annotationImage)
+    {
+        UIGraphicsBeginImageContext(CGSizeMake(1, 1));
```
tobrun: Is this creating an image with width/height 1/1 and using this in core as image for the pointannotation? As far as I can see this is the same hack I'm using on Android: https://github.com/mapbox/mapbox-gl-native/blob/master/platform/android/MapboxGLAndroidSDK/src/main/java/com/mapbox/mapboxsdk/annotations/MarkerView.java#L299

boundsj: @tobrun yes

PR #5466, opened 2016-06-24
[ios] fixes #5465 map view now informs about added annotation views

### LINKED ISSUE #5465: [ios] Add -mapView:didAddAnnotationViews
For parity with MapKit, we should add:
```
- (void)mapView:(MGLMapView *)mapView didAddAnnotationViews:(NSArray<MGLAnnotationView *> *)views
```

frederoni:
fixes #5465

### DIFF
--- a/platform/ios/src/MGLMapViewDelegate.h
+++ b/platform/ios/src/MGLMapViewDelegate.h
@@ -131,15 +131,6 @@ NS_ASSUME_NONNULL_BEGIN
 #pragma mark Managing the Display of Annotations
 
 /**
- Returns a view object to use for the marker for the specified point annotation object.
- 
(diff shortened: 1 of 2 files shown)


platform/ios/src/MGLMapViewDelegate.h
```diff
@@ -300,6 +300,18 @@ NS_ASSUME_NONNULL_BEGIN
  */
 - (void)mapView:(MGLMapView *)mapView didDeselectAnnotationView:(MGLAnnotationView *)annotationView;
 
+#pragma mark Managing annotation views
```
1ec5: Should -mapView:viewForAnnotation: also be under this section?

PR #5476, opened 2016-06-24
Reuse image buffer in HeadlessView

mikemorris:
Took a shot at wiring @tmpsantos' work on reusing the image buffer in `mbgl::HeadlessView` into the Node.js bindings, and something is definitely not right, causing segfaults during a benchmark - I'm not entirely sure how to properly handle memory management here, as I can't call `image->data.release()` with `std::shared_ptr<const mbgl::PremultipliedImage>` due to `const`, and some other stuff changed around to use `std::unique_ptr::swap` instead of `std::move`.

Can you take a closer look at this @jfirebaugh @tmpsantos, particularly the interaction between [`Nan::NewBuffer`](https://github.com/nodejs/nan/blob/master/doc/buffers.md#api_nan_new_buffer) and [`mbgl::PremultipliedImage::data`](https://github.com/mapbox/mapbox-gl-native/blob/4294b0f83a6894cd6a1679c7fb427eefeac5a7a6/include/mbgl/util/image.hpp#L54) in `NodeMap::renderFinished`?

### DIFF
--- a/platform/default/headless_view.cpp
+++ b/platform/default/headless_view.cpp
@@ -51,6 +51,8 @@ void HeadlessView::resize(const uint16_t width, const uint16_t height) {
 }
 
 std::shared_ptr<const PremultipliedImage> HeadlessView::readStillImage() {
+    MBGL_VERIFY_THREAD(tid);
+
     assert(active);
     assert(dimensions[0] <= dimensionsLimits[0]);
     assert(dimensions[1] <= dimensionsLimits[1]);
--- a/platform/default/headless_view.cpp
+++ b/platform/default/headless_view.cpp
@@ -4,6 +4,12 @@
 #include <cassert>
 #include <cstring>
 
+namespace {
+
+std::array<uint16_t, 2> dimensionsLimits = {{ 4096, 4096 }};
+
+}
+
 namespace mbgl {
 
 HeadlessView::HeadlessView(float pixelRatio_, uint16_t width, uint16_t height)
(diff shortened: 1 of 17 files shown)


jfirebaugh:
I don't think we want to bring `std::shared_ptr` into this. If you want the view to reuse the `PremultipliedImage` buffer, then make the callback type `void (std::exception_ptr, const PremultipliedImage&)`.

platform/default/headless_view.cpp
```diff
@@ -32,26 +38,41 @@ HeadlessView::~HeadlessView() {
...
 
-PremultipliedImage HeadlessView::readStillImage() {
+std::shared_ptr<const PremultipliedImage> HeadlessView::readStillImage() {
+    MBGL_VERIFY_THREAD(tid);
+
     assert(active);
+    assert(dimensions[0] <= dimensionsLimits[0]);
+    assert(dimensions[1] <= dimensionsLimits[1]);
+
+    static auto image = std::make_shared<PremultipliedImage>(
```
jfirebaugh: `static` means that there is one buffer for the entire process.

mikemorris: Ack, didn't catch that - this doesn't seem workable given how we have have up to 8 `NodeMap` objects per `node` process currently, each with their own `HeadlessView`?

tmpsantos: > static means that there is one buffer for the entire process.

This is exactly the idea. I added an assert to make sure this is always called from the same thread.

tmpsantos:
> I don't think we want to bring std::shared_ptr

No, we don't. I replaced with `std::shared_ptr` because I wanted to prototype if these allocations were causing fragmentation. On my tests on "core only" it didn't make much difference.

tmpsantos:
@mikemorris for testing this branch you have to synchronously encode the buffer into a PNG/JPEG on the `renderStill` callback which is probably not practical because you want to do that asynchronous on a node worker thread, right?

platform/default/headless_view.cpp
jfirebaugh: I don't think we want to limit `HeadlessView` to single-threaded programs. We could easily have each `HeadlessView` own a distinct buffer and still reuse that buffer.

tmpsantos: I will close this before we spend more bandwidth on it. This branch was never intended to upstream and was initially pushed to my own fork (https://github.com/tmpsantos/mapbox-gl-native/tree/tmpsantos-read_still_image). I suggest we keep experimental branchs on private forks.

tmpsantos:
@mikemorris I will leave the `Delete branch`for you in case you don't have a copy of this branch.

PR #5477, opened 2016-06-25
[core] Merge TexturePool into ObjectStore; pool all textures

jfirebaugh:
cc @kkaefer @brunoabinader -- Any reason not to make this simplification?


kkaefer:
:+1:

src/mbgl/gl/object_store.hpp
```diff
@@ -75,8 +66,13 @@ class ObjectStore : private util::noncopyable {
...
 
     UniqueTexture createTexture() {
-        GLuint id = 0;
-        MBGL_CHECK_ERROR(glGenTextures(1, &id));
+        if (pooledTextures.empty()) {
+            pooledTextures.resize(TextureMax);
+            MBGL_CHECK_ERROR(glGenTextures(TextureMax, pooledTextures.data()));
+        }
+
+        GLuint id = pooledTextures.back();
```
brunoabinader: I think this this will fail once the amount of textures in use reaches `TextureMax + 1`.

Previous texture pool logic uses a _pool of pools_, where a new texture pool would be generated when the current pool size reaches its limit - and old & unused pools would be eventually free'd.

jfirebaugh: If there are `TextureMax` textures in use, then `pooledTextures` will be `empty()` and the next `createTexture()` request will create `TextureMax` more textures.

brunoabinader: Got it now 👍 Indeed, after re-reading this patch I guess we didn't need to keep track of previous generated pools as we did before.

Issue #3979, opened 2016-02-16
gzip telemetry events


willwhite:
Make sure that we gzip telemetry events before sending them. gzip could reduce the time the mobile radio is working, especially on slower connections, and thus reduce battery impact.

/cc @boundsj

friedbunny:
I'm surprised that we don't do this already.

tobrun:
If content encoding is set to gzip, this is automatically handled in the background by `Okhttp` on Android.
You can verify this is the related [unit-test](https://github.com/square/okhttp/blob/00c9db7859ed4edac380b25c3be61afd9ece0e5d/okhttp-tests/src/test/java/okhttp3/CallTest.java#L1830-L1856), here they are showing Okhttp transparent gzipping in action.

boundsj:
There are 3rd party libs that make this transparent for iOS, too (https://github.com/AFNetworking/AFgzipRequestSerializer/blob/master/AFgzipRequestSerializer/AFgzipRequestSerializer.m?#L57) but we don't use a 3rd party dep in the iOS bindings

mawajdak:
الشبكات
لا توجد شبكات

friedbunny:
Resources that we can draw upon for our own GZIP implementation:

- https://github.com/mattt/Godzippa (used in AFNetworking, per above)
- https://github.com/nicklockwood/GZIP

tobrun:
What I mentioned in https://github.com/mapbox/mapbox-gl-native/issues/3979#issuecomment-185063771 is not true. 

To GZIP our communication we need to implement the following request interceptor on OkHttp: https://github.com/square/okhttp/blob/3f7a3344a4c85aa3bbb879dabac5ee625ab987f3/samples/guide/src/main/java/okhttp3/recipes/RequestBodyCompression.java#L73

ivovandongen:
@tobrun We're using something similar in Human too. Can take a look at that too if you want.

mick:
The api already excepts gzip'd JSON as long as the following headers are set on the POST request:

```
Content-type: application/json
Content-Encoding: gzip
```

bleege:
I've ticketed out the Android work in #5536.

friedbunny:
This is done on the iOS side (#5490) and will be in the final v3.3.0 release.

zugaldia:
[...]

Issue #5472, opened 2016-06-24
Require Xcode 7.3 due to NS_SWIFT_NAME


1ec5:
We’re using the `NS_SWIFT_NAME` macro in a few places in MGLMapView.h, breaking compilation in versions of Xcode before 7.3. MGLTypes.h should `#define NS_SWIFT_NAME` to nothing if that macro is undefined.

Alternatively, we could require Xcode 7.3 and simultaneously inline all our uses of `NS_ARRAY_OF` and company, which are intended to keep Xcode 6.x happy.

/cc @friedbunny

1ec5:
Defining `NS_SWIFT_NAME` to nothing will avoid an error on Xcode 6.x but not Xcode 7.0–7.2, in which the macro is defined but can’t be used on methods they way we’re using them.

At this point, MGLTypes.h is so chock full of workarounds for Xcode 6.x that we should just require Xcode 7.3; it isn’t much trouble to upgrade 7.0 to 7.3 anyhow. If @mapbox/gl agrees, the next steps are:

* [ ] Document Xcode 7.3 is a requirement for building the iOS and macOS SDKs
* [ ] Inline `NS_ARRAY_OF` and company and remove these aliases from MGLTypes.h

kkaefer:
I'm fine with requiring Xcode 7.3. Do we have statistics on what versions are still in use out there? Given that Xcode typically auto-updates, I imagine there aren't many older versions out there?

friedbunny:
Xcode 7.3 is evidently already a requirement — trying to use a lower version will throw `Only factory methods can have the swift_name attribute.`.

friedbunny:
Per chat with @1ec5, we still need to have [different `NS_ARRAY_OF` definitions](https://github.com/mapbox/mapbox-gl-native/blob/bca9d091805dc01a4456ab3f24e9de87f9b4aa48/platform/darwin/src/MGLTypes.h#L59) to support iOS 8 and macOS 10.10.

friedbunny:
Theoretically something like this could work:

```
// Versions of Xcode lower than 7.3 do not fully support NS_SWIFT_NAME.
#if defined(__apple_build_version__) && __apple_build_version__ < 703029
    #undef NS_SWIFT_NAME
    #define NS_SWIFT_NAME(_)
#endif
```

... but we’ve already broken Xcode 6 support in other ways, so I’m going to forge ahead with requiring Xcode 7.3.1.

PR #5490, opened 2016-06-27
[ios] #3979 compress telemetry events

frederoni:
##### Things to consider:
 - ~~License for dependency ([zlib](https://en.wikipedia.org/wiki/Zlib_License))~~ 
 - Threshold for compression
 - Build request in a background thread

An event in the following test case is `lat, lng, timestamp` with randomized values.

| Events        |  Raw size   | Compressed size | Compression  |
| ------------- | -----------:|  ------:|        -----:|
| 1 event*    |  65 bytes   | 85 bytes   | 1.3076  |
| 2 events   |  117 bytes  | 100 bytes  | 0.8547 |
| 5 events   |  273 bytes  | 126 bytes  | 0.4615 |
| 180 events |  9125 bytes | 1290 bytes | 0.1413 |

*Compressing one event actually increases the size of data on top of using unnecessary CPU cycles.

 👀 @boundsj @friedbunny @1ec5


friedbunny:
Awesome, these results look quite promising!

You should go ahead and add a copy of NSData+GZIP's license to [our list of licenses](https://github.com/mapbox/mapbox-gl-native/blob/master/LICENSE.md). Looks like the license [in the repo](https://github.com/nicklockwood/GZIP/blob/5602735f71634052d186383d91bda1ecff77b92a/LICENCE.md) is slightly different and more up-to-date. We also already include the original zlib license, so I think we’re covered there.

@1ec5 mentioned prefixing as a thing we should do, as we’ve done with [Reachability](https://github.com/mapbox/mapbox-gl-native/blob/b8326870e1f3960b169746b0473326060bb6cf54/platform/darwin/src/reachability.m#L87), which became `MGLReachability`. In this case, we’ll want to prefix the category methods.

NSData+GZIP’s files should also probably live in a subfolder of [`platform/ios/vendor`](https://github.com/mapbox/mapbox-gl-native/tree/master/platform/ios/vendor), rather than commingled with our own source in `platform/ios/src`.

1ec5:
So it turns out we already have a simple interface to zlib in the codebase, in src/mbgl/util/compression.hpp. (That's part of the reason we already include the zlib license in our LICENSE file.) The header is currently private to mbgl, but I don't see any reason we couldn't move it to include/mbgl/util/ and use it within the SDK. compression.hpp uses std::string to hold the compressed data, but it's trivial to dump that buffer into an NSData.

frederoni:
Exposed the zlib interface we already had an interface for in include/mbgl/util/compression.hpp and got some interesting results

| Events        |  Raw size   | Compressed size | Compression  |
| ------------- | -----------:|  ------:|        -----:|
| 1 event*    |  52 bytes   | 52 bytes   | 1  |
| 2 events   |  103 bytes  | 105 bytes  | 1.0194 |
| 5 events   |  251 bytes  | 115 bytes  | 0.4581 |
| 180 events |  9136 bytes | 267 bytes | 0.0292 |

frederoni:
Noticed that the compressed data was corrupt so even though we don't need it for the iOS SDK, I had to add `-mgl_decompressedData` just to be able to write a test case.

platform/darwin/src/NSData+MGLAdditions.mm
```diff
@@ -0,0 +1,33 @@
...
+
+#include <mbgl/util/compression.hpp>
+
+@implementation NSData (MGLAdditions)
+
+- (NSData *)mgl_compressedData
+{
+    NSString *dataString = [[NSString alloc] initWithData:self encoding:NSUTF8StringEncoding];
+    
+    std::string *string = new std::string([dataString UTF8String]);
```
1ec5: C++ types should be allocated on the stack, whenever possible, rather than the heap.

platform/darwin/src/NSData+MGLAdditions.mm
```diff
@@ -0,0 +1,33 @@
+#import "NSData+MGLAdditions.h"
+
+#include <mbgl/util/compression.hpp>
+
+@implementation NSData (MGLAdditions)
+
+- (NSData *)mgl_compressedData
+{
+    NSString *dataString = [[NSString alloc] initWithData:self encoding:NSUTF8StringEncoding];
```
1ec5: You can avoid this step by passing `self.bytes` into a `std::string` constructor.

platform/darwin/src/NSData+MGLAdditions.mm
```diff
@@ -0,0 +1,33 @@
...
+    delete string;
+    
+    return [NSData dataWithBytes:(void *) compressed_string.c_str() length:compressed_string.length()];
+}
+
+- (NSData *)mgl_decompressedData
+{
+    unsigned char buffer[[self length]];
+    
+    [self getBytes:buffer length:[self length]];
```
1ec5: You can avoid this step by accessing `self.bytes`. Since `mbgl::util::decompress()` only reads that memory within the scope of this method, that otherwise unsafe property is fine to use locally.

platform/darwin/src/NSData+MGLAdditions.h
```diff
@@ -0,0 +1,13 @@
...
+
+@interface NSData (MGLAdditions)
+
+- (NSData *)mgl_compressedData;
+
+- (NSData *)mgl_decompressedData;
+
+@end
+
+NS_ASSUME_NONNULL_END
```
1ec5: Nit: missing newline at the end.

platform/darwin/src/NSData+MGLAdditions.h
```diff
@@ -0,0 +1,13 @@
+#import <Foundation/Foundation.h>
+
+NS_ASSUME_NONNULL_BEGIN
```
1ec5: Headers that use `NS_ASSUME_NONNULL_BEGIN` should also import MGLTypes.h, or Xcode 6 will fail to compile the code. We’re considering phasing out this requirement, however: #5472.

platform/ios/ios.xcodeproj/project.pbxproj
```diff
@@ -7,7 +7,12 @@
 	objects = {
 
 /* Begin PBXBuildFile section */
+		35305D481D22AA680007D005 /* NSData+MGLAdditions.mm in Sources */ = {isa = PBXBuildFile; fileRef = 35305D471D22AA450007D005 /* NSData+MGLAdditions.mm */; };
+		35305D491D22AA680007D005 /* NSData+MGLAdditions.mm in Sources */ = {isa = PBXBuildFile; fileRef = 35305D471D22AA450007D005 /* NSData+MGLAdditions.mm */; };
+		35305D4A1D22AA6A0007D005 /* NSData+MGLAdditions.h in Headers */ = {isa = PBXBuildFile; fileRef = 35305D461D22AA450007D005 /* NSData+MGLAdditions.h */; settings = {ATTRIBUTES = (Public, ); }; };
+		353794D01D22B3BD002C281C /* NSData+MGLAdditions.h in Headers */ = {isa = PBXBuildFile; fileRef = 35305D461D22AA450007D005 /* NSData+MGLAdditions.h */; settings = {ATTRIBUTES = (Public, ); }; };
```
1ec5: This header should be Private, not Public. Private allows other targets in this workspace, like the test bundle, to see the header, while avoiding shipping the header publicly.

include/mbgl/util/compression.hpp
```diff
@@ -0,0 +1,12 @@
+#pragma once
+
+#include <string>
+
+namespace mbgl {
+    namespace util {
```
1ec5: Nit: the convention inside mbgl is to avoid extra indentation due to namespaces.

include/mbgl/util/compression.hpp
```diff
@@ -0,0 +1,12 @@
...
+        
+        std::string compress(const std::string &raw);
+        std::string decompress(const std::string &raw);
```
1ec5: Nit: the convention inside mbgl is to place `&` and `*` next to the type rather than the symbol name, so `const std::string& raw`. This is different than the convention in our Objective-C++ code at the SDK level.

1ec5:
It looks like you’ve copied src/mbgl/util/compression.hpp to include/mbgl/util/compression.hpp. Instead, you should move the file there. The convention is that headers in include/ are public (to SDK code), while headers in src/ can only be used in the SDK. Once you delete src/mbgl/util/compression.hpp, you’ll need to update compression.cpp to include `<mbgl/util/compression.hpp>` instead of `"compression.hpp"`.

platform/ios/test/MGLNSDataAdditionsTests.m
```diff
@@ -0,0 +1,44 @@
...
+        [array addObject:@{@"lat": @(@([self randomFloatBetween:-90 and:90]).integerValue),
+                           @"lng": @(@([self randomFloatBetween:-180 and:180]).integerValue),
+                           @"timestamp": @(@([[NSDate date] timeIntervalSince1970]).integerValue)}];
+    }
+    return array;
+}
+
+- (float)randomFloatBetween:(float)lowerBound and:(float)upperBound {
+    float diff = upperBound - lowerBound;
+    return (((float) (arc4random() % ((unsigned)RAND_MAX + 1)) / RAND_MAX) * diff) + lowerBound;
```
boundsj: nit: you might consider using [arc4random_uniform](https://developer.apple.com/legacy/library/documentation/Darwin/Reference/ManPages/man3/arc4random_uniform.3.html)

friedbunny: Hrm, good to know. 💭

frederoni: Good one but since I want lat/lng I'm probably gonna go with `drand48`.
I'm grabbing the value as integer after that because NSArray/NSDictionary `-isEqualToDictionary/Array` returns false despite the content being the same. Any idea why?

1ec5: Comparisons between floats often fail due to floating-point representations being imprecise. That’d be my first guess. If that turns out to be the case, try multiplying the latitude and longitude by enough decimal places that the imprecision wouldn’t matter much in reality, then floor the values. See [this table](https://en.wikipedia.org/wiki/Decimal_degrees) to get an idea of how much precision you need.

platform/darwin/src/NSData+MGLAdditions.mm
```diff
@@ -0,0 +1,25 @@
...
+#include <mbgl/util/compression.hpp>
+
+@implementation NSData (MGLAdditions)
+
+- (NSData *)mgl_compressedData
+{
+    std::string string(static_cast<const char*>(self.bytes), self.length);
+    std::string compressed_string = mbgl::util::compress(string);
+    
+    return [NSData dataWithBytes:(void *) compressed_string.c_str() length:compressed_string.length()];
```
1ec5: A more idiomatic way of doing this would be `&compressed_string[0]`.

1ec5:
👍

kkaefer:
It's worth noting that this is a non-standard way of using HTTP; theoretically our client cannot know that the server supports `Content-Encoding: gzip`, but given that the endpoint is always a server we control, we can deploy this anyway.

frederoni:
You're right @kkaefer. I guess we could verify supported  `Content-Encoding` by making a `HEAD` request with `Accept-Encoding: gzip` (or in our case: `deflate` since we're not compressing with a gzip wrapper but zlib deflate compression) to verify that the server/endpoint supports it. But the endpoint doesn't support `HEAD` requests and it's not guaranteed even if it did. Only more likely if it's supported at http server level. And as you mentioned, the endpoint is always a server we control.

kkaefer:
@frederoni not saying we should do anything else, I just wanted to make everyone aware that this is something we explicitly need to support in our server (and we already do).

Issue #5495, opened 2016-06-28
After dragging an annotation `mapView:didDragAnnotationView:toCoordinate:` is not always called.


boundsj:
The reference an annotation view currently keeps to the mapview it is shown in is currently set up only in `updateAnnotations` and not when the annotation is first created. Therefore, if an annotation is dragged
before it is processed by `updateAnnotations:` it will not have a mapview and will not call the `mapView:didDragAnnotationView:toCoordinate:` delegate method.

We should set the mapview property when an annotation view is first created, too.

boundsj:
Fixed by https://github.com/mapbox/mapbox-gl-native/pull/5496

Issue #5494, opened 2016-06-28
MGLAnnotationView does not close callout view when dragging starts.


boundsj:
When dragging and dropping an annotation view, the callout popover does not close automatically when dragging starts. This results in a visually detached callout if one is shown before its annotation is dragged.

<img width="248" alt="screen shot 2016-06-27 at 5 30 53 pm" src="https://cloud.githubusercontent.com/assets/63186/16400084/22d7f93c-3c8d-11e6-8ea6-5d7678e9af43.png">

boundsj:
Fixed by https://github.com/mapbox/mapbox-gl-native/pull/5496

PR #5496, opened 2016-06-28
[ios] Always set map view reference when annotation view is created.

boundsj:
Closes https://github.com/mapbox/mapbox-gl-native/issues/5494
Closes https://github.com/mapbox/mapbox-gl-native/issues/5495

This fixes two issues with annotation views that were exposed when dragging them:

1) The annotation view's mapView property was not set until it was moved (in `updateAnnotations:`) so it did not always call its delegate back about the map view delegate dragging methods. This adds a call to
set the property when an annotation view is created so the delegate methods are always called.

2) Since the mapView is now always set, a call to the map view's callout view for selected annotation was added to the annotation view drag handler. This forces the callout to close and eliminates an issue where the callout would remain visible as an annotation was dragged away which had the visual effect of a detached callout.

To make 2 possible, the map views selected callout view was exposed in the private header.

@frederoni @1ec5 this adds test to verify that the issues are fixed and continues the pattern in `MGLAnnotationViewTests` of manually stubbing out dependencies and spying on interaction methods. If we are going to unit test at this level, does it makes sense to link OCMock to `SDK Tests`?

### DIFF
--- a/platform/ios/src/MGLAnnotationView.mm
+++ b/platform/ios/src/MGLAnnotationView.mm
@@ -187,7 +187,7 @@
 {
     CGPoint center = [sender locationInView:sender.view.superview];
     [self setCenter:center pitch:self.mapView.camera.pitch];
-    
+
     if (sender.state == UIGestureRecognizerStateEnded) {
         self.dragState = MGLAnnotationViewDragStateNone;
     }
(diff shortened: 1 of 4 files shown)


platform/ios/src/MGLAnnotationView.mm
```diff
@@ -187,7 +187,7 @@ - (void)handlePan:(UIPanGestureRecognizer *)sender
 {
     CGPoint center = [sender locationInView:sender.view.superview];
     [self setCenter:center pitch:self.mapView.camera.pitch];
-    
+
```
boundsj: fyi I'm using AppCode which removes whitespace issues when a file is saved.

PR #5497, opened 2016-06-28
Specify ultra light weight literally

### LINKED ISSUE #5486: crashed when I init a MapView
**Platform: iOS 7
**Mapbox SDK version: v3.3.0-beta2

-  ##Is it not support iOS7 at beta version?

1ec5:
Fixed a crash on iOS 8.1 and below due to specifying the compass legend’s font weight using a font weight constant that was introduced in iOS 8.2. `UIFontWeightUltraLight` happens to equal −0.8 on iOS 9 in the default configuration. I think it’s reasonable to hardcode this value as a fallback; we still use the system constant when it’s available.

Fixes #5486, a regression introduced in #4783.

/cc @friedbunny


friedbunny:
👍

PR #5502, opened 2016-06-28
[ios] Adds the ability to get selection events for shape annotations

csmulhern:
This should not break any existing behavior, as the implementation makes shapes default to `enabled = NO`.

### DIFF
--- a/include/mbgl/map/map.hpp
+++ b/include/mbgl/map/map.hpp
@@ -149,6 +149,7 @@ public:
     void removeAnnotation(AnnotationID);
 
     AnnotationIDs getPointAnnotationsInBounds(const LatLngBounds&);
+    AnnotationIDs getShapeAnnotationsInBounds(const LatLngBounds&);
 
     // Sources
     style::Source* getSource(const std::string& sourceID);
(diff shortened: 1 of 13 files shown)


1ec5:
Thank you for this PR – this is absolutely important functionality to have. I think we’ll ultimately want to combine your SDK-level changes with #5165, which avoids having to pull in `boost::geometry`. (With #5110 and #352, we finally have a feature index that can be used for everything, not just points and not just annotations.) I’m currently focusing on the 3.3.0 release right now, but I’ll pick up #5165 again right afterwards so we can get rid of the existing mbgl code that’s limited to points.

include/mbgl/map/map.hpp
```diff
@@ -149,6 +149,7 @@ class Map : private util::noncopyable {
     void removeAnnotation(AnnotationID);
 
     AnnotationIDs getPointAnnotationsInBounds(const LatLngBounds&);
+    AnnotationIDs getShapeAnnotationsInBounds(const LatLngBounds&);
```
1ec5: #5165 landed, so now we can implement a `queryShapeAnnotations()` that calls `queryRenderedFeatures()` – no custom querying logic is needed anymore. Unlike in `queryPointAnnotations()`, the call to `queryRenderedFeatures()` will need to pass in the layer IDs of all the shape annotation layers.

1ec5:
This was implemented in #9984 with the approach described in https://github.com/mapbox/mapbox-gl-native/pull/5502#discussion_r104299116. Thank you for this early iteration, though!

Issue #5501, opened 2016-06-28
MGLMapView does not update while a UIScrollView is scrolling


kvurd:
**Platform: iOS**
**Mapbox SDK version: 3.3.0 beta2** (note this is not an issue in 3.2.3)

### Steps to trigger behavior

 1. Create an `MGLMapView` and a `UIScrollView`.
 2. Change the `centerCoordinate` or `zoomLevel` of the map view while scrolling in the scroll view (via `scrollViewDidScroll` or similar)

### Expected behavior
Map updates while scrolling in scroll view.

### Actual behavior
Map does not update while scrolling in scroll view, but does update once scrolling finishes and finger is lifted.

### More info
I think this might be a run loop issue in `platform/darwin/src/async_task.cpp`. Perhaps `kCFRunLoopDefaultMode` should be `kCFRunLoopCommonModes` to ensure updates while the scroll view is scrolling. Not sure if this will cause side effects elsewhere...

boundsj:
Thanks @kvurd for bringing this to our attention! I think you are correct and I've made the change in https://github.com/mapbox/mapbox-gl-native/pull/5503. This should land in our next release candidate build for [version 3.3.0](https://github.com/mapbox/mapbox-gl-native/milestones/ios-v3.3.0)

PR #5503, opened 2016-06-28
[ios, macos] Use common instead of default mode for run loop source

boundsj:
The map events instigated by certain actions (i.e. updates from a UIScrollView) are not processed by the run loop with `kCFRunLoopDefaultMode`until the thread is idle. This changes the mode to `kCFRunLoopCommonModes` to avoid this issue.

cc @1ec5 @jfirebaugh


boundsj:
Closes #5501

1ec5:
👍

PR #5504, opened 2016-06-28
Adjustable map deceleration rate

1ec5:
This is a continuation of #3619 that makes the “drift” animation after a user gesture configurable.

/cc @csmulhern @frederoni


friedbunny:
Thanks, @csmulhern — this is pretty cool.

csmulhern:
Looks good @1ec5, thanks.

PR #5514, opened 2016-06-29
[core] geometry@0.8.0 / geojsonvt@6.0.0

yhahn:
`geometry.hpp`-based geojsonvt.

- [x] Shape annotation tests are failing. I need to dig into this
- [x] Update all platform build configs
- [x] Tests passing, IRL test with glfw-app
- [x] Remaining code adjustments

------

### geometry-0.8.0

- [x] rebase in geometry 0.8.0
- [x] update geojson-vt dep

---

- [x] Code review


include/mbgl/util/geometry.hpp
```diff
@@ -12,6 +12,16 @@ enum class FeatureType : uint8_t {
     Polygon = 3
 };
 
+struct ToFeatureType {
+    FeatureType operator()(mapbox::geometry::point<int16_t>) const { return FeatureType::Point; }
```
jfirebaugh: These should all take their argument as `const __ &` to avoid copying, and use the CamelCase aliases defined below.

src/mbgl/tile/geometry_tile_data.hpp
```diff
@@ -73,4 +74,72 @@ Feature convertFeature(const GeometryTileFeature&, const CanonicalTileID&);
...
+    GeometryCollection operator()(const mapbox::geometry::point<int16_t>& geom) const {
+        GeometryCollection collection;
+        GeometryCoordinates coordinates;
+        coordinates.emplace_back(geom.x, geom.y);
+        collection.push_back(coordinates);
+        return collection;
+    }
+    GeometryCollection operator()(const mapbox::geometry::multi_point<int16_t>& geom) const {
+        GeometryCollection collection;
+        GeometryCoordinates coordinates;
```
jfirebaugh: As a performance optimization, wherever you copy vector-to-vector with `emplace_back` in a loop, do `coordinates.reserve(geom.size())` beforehand to pre-allocate the exact size.

src/mbgl/annotation/shape_annotation_impl.cpp
```diff
@@ -19,84 +18,21 @@ ShapeAnnotationImpl::ShapeAnnotationImpl(const AnnotationID id_, const uint8_t m
...
-        const double tolerance = baseTolerance / (maxAmountOfTiles * util::EXTENT);
-
-        std::vector<geojsonvt::ProjectedFeature> features = {
-            ShapeAnnotationGeometry::visit(geometry(), ToGeoJSONVT(tolerance))
-        };
-
+        ToGeometry toGeometry;
+        mapbox::geometry::feature_collection<double> features;
+        mapbox::geometry::feature<double> feature = { apply_visitor(toGeometry, geometry()) };
+        features.emplace_back(feature);
```
jfirebaugh: I _think_ you should be able to write this using a generic lambda as the visitor:

```C++
FeatureCollection features;
features.emplace_back(ShapeAnnotationGeometry::visit(geometry(), [] (auto&& geom) {
    return Feature(geom);
}));
```

We don't have a `FeatureCollection` alias in `mbgl` but we should.

src/mbgl/tile/geojson_tile.cpp
```diff
@@ -49,52 +49,28 @@ class GeoJSONTileData : public GeometryTileData {
...
-            }
+            GeometryCollection geometry = apply_visitor(toGeometryCollection, feature.geometry);
 
             // https://github.com/mapbox/geojson-vt-cpp/issues/44
             if (featureType == FeatureType::Polygon) {
                 geometry = fixupPolygons(geometry);
             }
 
-            Feature::property_map properties{ feature.tags.begin(), feature.tags.end() };
+            Feature::property_map properties{ feature.properties.begin(), feature.properties.end() };
```
jfirebaugh: `feature.properties` is now the same type as `Feature::property_map`, I think? If so this can be a direct assignment?

src/mbgl/annotation/shape_annotation_impl.cpp
```diff
@@ -106,34 +42,20 @@ void ShapeAnnotationImpl::updateTileData(const CanonicalTileID& tileID, Annotati
     }
 
     const auto& shapeTile = shapeTiler->getTile(tileID.z, tileID.x, tileID.y);
-    if (!shapeTile)
+    if (shapeTile.features.size() == 0)
```
jfirebaugh: `size() == 0` ⇢ `empty()`

src/mbgl/annotation/shape_annotation_impl.cpp
```diff
@@ -19,121 +18,37 @@ ShapeAnnotationImpl::ShapeAnnotationImpl(const AnnotationID id_, const uint8_t m
...
-            ShapeAnnotationGeometry::visit(geometry(), ToGeoJSONVT(tolerance))
-        };
-
+        mapbox::geometry::feature_collection<double> features;
+        features.emplace_back(ShapeAnnotationGeometry::visit(geometry(), [] (auto&& geom) {
+            return Feature(std::move(geom));
+        }));
         mapbox::geojsonvt::Options options;
-        options.maxZoom = maxZoom;
+        options.maxZoom = mbgl::util::clamp((int)maxZoom, 0, 18);
```
yhahn: Shape annotation tests were failing without this clamp -- the `maxZoom` passed here is the map's mapZoom (20). I need to dig further in master to understand what's changed (was geojson-vt-cpp clamping this?)

mourner:
@brunoabinader can you check if this fixes that maxZoom issue?

src/mbgl/annotation/shape_annotation_impl.cpp
```diff
@@ -19,121 +18,37 @@ ShapeAnnotationImpl::ShapeAnnotationImpl(const AnnotationID id_, const uint8_t m
...
-    if (!shapeTile)
+    if (shapeTile.features.empty())
         return;
 
     AnnotationTileLayer& layer = *data.layers.emplace(layerID,
         std::make_unique<AnnotationTileLayer>(layerID)).first->second;
 
+    ToGeometryCollection toGeometryCollection;
+    ToFeatureType toFeatureType;
     for (auto& shapeFeature : shapeTile.features) {
```
brunoabinader: I think we can replace this `auto&` with a `const auto&`.

src/mbgl/annotation/shape_annotation_impl.cpp
brunoabinader: This looks similar to the issue I am debugging in https://github.com/mapbox/mapbox-gl-native/pull/5517. I've previously lowered the maximum zoom range for `AnnotationSource` from 22 to 18 alleviate the issue of disappearing annotations in z19 [here](https://github.com/mapbox/mapbox-gl-native/pull/5418).

brunoabinader:
>@brunoabinader can you check if this fixes that maxZoom issue?

Just did a rebase of my patches from https://github.com/mapbox/mapbox-gl-native/pull/5517 on top of this branch. I removed the zoom clamping introduced by @yhahn otherwise the app crashes inside GeoJSONVT's `getTile`:
```
#0  0x00007ffff56f9418 in __GI_raise (sig=sig@entry=6) at ../sysdeps/unix/sysv/linux/raise.c:54
#1  0x00007ffff56fb01a in __GI_abort () at abort.c:89
#2  0x00007ffff5d3284d in __gnu_cxx::__verbose_terminate_handler() () from /usr/lib/x86_64-linux-gnu/libstdc++.so.6
#3  0x00007ffff5d306b6 in ?? () from /usr/lib/x86_64-linux-gnu/libstdc++.so.6
#4  0x00007ffff5d30701 in std::terminate() () from /usr/lib/x86_64-linux-gnu/libstdc++.so.6
#5  0x00007ffff5d30919 in __cxa_throw () from /usr/lib/x86_64-linux-gnu/libstdc++.so.6
#6  0x000000000089bbf4 in mapbox::geojsonvt::GeoJSONVT::getTile (this=0x227a1b0, z=20 '\024', x_=167740, y=405294)
```

Although the unit test I introduce passes (all zoom levels returns non-empty features when queried with `queryRenderedFeatures`) - I've noticed erratic annotation behavior as shown:

out

This comes along with a console filled with ```[DEBUG] {mapbox-glfw}[OpenGL]: Buffer doesn't contain elements``` output. This weird behavior is also present also without my patches applied.

yhahn:
Once :green_apple: this PR will be ready for another review! Here are some known issues/open questions remaining:

- [ ] geojsonvt maxzoom issue - determine if this work should wait until @brunoabinader's annotation maxzoom PR fix lands.
- [ ] Determine what to do about two known API behavior changes:
  - Shape annotation API now must receive closed rings for polygons to render correctly. There's no validation/assertion in place for this, and the behavior of this has changed from `master`. (https://github.com/mapbox/mapbox-gl-native/blob/geojsonvt/platform/default/glfw_view.cpp#L275-L276)
  - `geojson.cpp` is now stricter and requires a `properties` key when parsing GeoJSON features. (https://github.com/mapbox/mapbox-gl-test-suite/pull/120)
  - **Question:** do we need to smooth these changes out for developers or should we nudge them toward more strictness?

cc @jfirebaugh @mourner @brunoabinader

mourner:
> Shape annotation API now must receive closed rings for polygons to render correctly.

What that the rendering issue you were seeing earlier? For the annotations API, it would be easy enough to close any unclosed polygon rings, right?

> `geojson.cpp` is now stricter and requires a properties key when parsing GeoJSON features. 

Note that features with `properties: null` are still a valid GeoJSON.

yhahn:
> What that the rendering issue you were seeing earlier? For the annotations API, it would be easy enough to close any unclosed polygon rings, right?

Yes. I started to poke at this but it looks like a slippery slope to me -- we would need to add an early closed-ring validator + corrector for fixing any polygons/multipolygons (ie. before shape annotations are passed to `geojsonvt` for clipping.) Do we do this elsewhere in mbgl?

> Note that features with properties: null are still a valid GeoJSON.

Yep, this condition is handled properly.

yhahn:
After re-reading the discussion on https://github.com/mapbox/mapbox-gl-native/pull/5517 I think we should land this first (get fully up to date on `geojson-vt`) and then I can reroll @brunoabinader's testcase and test.

src/mbgl/style/sources/geojson_source_impl.cpp
```diff
@@ -18,14 +20,30 @@ using namespace mapbox::geojsonvt;
...
+
+struct ToFeatureCollection {
+    mapbox::geojson::feature_collection operator()(const mapbox::geojson::feature_collection value) const {
+        return value;
+    }
+    mapbox::geojson::feature_collection operator()(const mapbox::geojson::feature value) const {
+        return mapbox::geojson::feature_collection { value };
+    }
+    mapbox::geojson::feature_collection operator()(const mapbox::geojson::geometry value) const {
+        return mapbox::geojson::feature_collection { { value } };
```
mourner: C++ should deduct the return type so this can be just `return { ... }`, right?

mourner:
Looked through the code and this looks great to me! Looking forward to this being merged. I'll follow-up with supercluster.hpp integration after that.

src/mbgl/style/sources/geojson_source_impl.cpp
```diff
@@ -18,14 +20,30 @@ using namespace mapbox::geojsonvt;
 namespace mbgl {
 namespace style {
 namespace conversion {
+
+struct ToFeatureCollection {
+    mapbox::geojson::feature_collection operator()(const mapbox::geojson::feature_collection value) const {
```
springmeyer: The `value` in each visitor here could be passed by `const &` to avoid an unneeded copy. That said, the compiler will likely optimize away the copy so its not certain my suggestion will run any faster.

src/mbgl/tile/geojson_tile.cpp
```diff
@@ -49,52 +49,28 @@ class GeoJSONTileData : public GeometryTileData {
 std::unique_ptr<GeoJSONTileData> convertTile(const mapbox::geojsonvt::Tile& tile) {
     std::shared_ptr<GeoJSONTileLayer> layer;
 
-    if (tile) {
+    if (tile.features.size()) {
```
springmeyer: if `(!tile.features.empty()) {` should be faster per https://github.com/mapbox/mapbox-gl-native/pull/1842

jfirebaugh:
> Shape annotation API now must receive closed rings for polygons to render correctly.

My sense is this will be an unexpected / breaking change for users and we should retain the loose API for annotations. We've never documented the annotations API as requiring closed rings or strict GeoJSON geometry.

> `geojson.cpp` is now stricter and requires a `properties` key when parsing GeoJSON features. (https://github.com/mapbox/mapbox-gl-test-suite/pull/120)

In contrast this seems like an acceptable change. It's reasonable to require a GeoJSON source have valid GeoJSON data. Making the parser more conformant can be construed as a bug fix. And use of GeoJSON sources is low in any case.

src/mbgl/annotation/shape_annotation_impl.cpp
jfirebaugh: `util::clamp<uint8_t>(maxZoom, 0, 18)`

src/mbgl/style/sources/geojson_source_impl.cpp
```diff
@@ -18,14 +20,30 @@ using namespace mapbox::geojsonvt;
...
     Options options;
     options.buffer = util::EXTENT / util::tileSize * 128;
     options.extent = util::EXTENT;
 
     try {
-        return GeoJSON { std::make_unique<GeoJSONVT>(Convert::convert(value, 0), options) };
+        ToFeatureCollection toFeatureCollection;
+        const auto geojson = mapbox::geojson::convert(value);
+        const auto features = apply_visitor(toFeatureCollection, geojson);
+        return GeoJSON { std::make_unique<GeoJSONVT>(features, options) };
```
jfirebaugh: It's awkward to be unable to directly compose `mapbox::geojson::convert` and `mapbox::GeoJSONVT::GeoJSONVT `. We should add a `GeoJSONVT` constructor that accepts `mapbox::util::variant<geometry, feature, feature_collection>`. cc @mourner

src/mbgl/tile/geometry_tile_data.hpp
```diff
@@ -73,4 +74,70 @@ Feature convertFeature(const GeometryTileFeature&, const CanonicalTileID&);
...
 
+struct ToGeometryCollection {
+    GeometryCollection operator()(const mapbox::geometry::point<int16_t>& geom) const {
+        return { { geom } };
+    }
+    GeometryCollection operator()(const mapbox::geometry::multi_point<int16_t>& geom) const {
+        GeometryCoordinates coordinates;
+        coordinates.reserve(geom.size());
+        for (const auto& point : geom) {
+            coordinates.emplace_back(point.x, point.y);
```
jfirebaugh: `s/emplace_back(point.x, point.y)/emplace_back(point)/g`

src/mbgl/tile/geometry_tile_data.hpp
```diff
@@ -73,4 +74,70 @@ Feature convertFeature(const GeometryTileFeature&, const CanonicalTileID&);
...
+    GeometryCollection operator()(const mapbox::geometry::multi_line_string<int16_t>& geom) const {
+        GeometryCollection collection;
+        collection.reserve(geom.size());
+        for (const auto& ring : geom) {
+            GeometryCoordinates coordinates;
+            coordinates.reserve(ring.size());
+            for (const auto& point : ring) {
+                coordinates.emplace_back(point.x, point.y);
+            }
+            collection.push_back(coordinates);
```
jfirebaugh: `push_back(std::move(coordinates))`

1ec5:
> Shape annotation API now must receive closed rings for polygons to render correctly.

This is fine from the iOS and macOS SDKs’ perspective: for compatibility with MapKit, we’ve documented a requirement that MGLPolygon be closed, and the behavior is undefined otherwise.

yhahn:
@jfirebaugh ready whenever you have a chance for another :eyes: Only major change besides the cleanup from notes is the addition of the ring-closer for fill annotations:

- https://github.com/mapbox/mapbox-gl-native/pull/5514/commits/dd8ceb44720ead7907484215459e58922161aed0

jfirebaugh:
Ring closing should only apply to polygons and multipolygons, not line strings.

1ec5:
8ca8e2c5093db331d592b8bd4abe695cde2e77ab updates the MGLFeature documentation in the iOS and macOS SDKs to reflect the fact that a feature identifier may be one of several types, not necessarily an NSNumber. We deliberately made the identifier [as broadly typed as possible](https://github.com/mapbox/mapbox-gl-native/pull/5110#issuecomment-221121610) (`id`, amusingly enough) in anticipation of this change, so no backwards compatibility issues are expected.

PR #5517, opened 2016-06-30
[core] Restore AnnotationSource maximum zoom range to 22

### LINKED ISSUE #5505: Annotation source should use zoom range 0...22
We experienced some rendering issues - see https://github.com/mapbox/mapbox-gl-native/issues/5138 - that prevented us from having `AnnotationSource`'s zoom range set to 0..22 - thus requiring a downgrade for the maximum zoom to 18 - see https://github.com/mapbox/mapbox-gl-native/pull/5418.

However, this is not ideal as we're losing precision in situations where the map zoom level is higher than 18. We need to figure out why this is happening.

:eyes: @mapbox/gl

brunoabinader:
This is a work-in-progress PR.

So far:
- Implemented a unit test that reproduces the missing annotation on z19: `Annotations.AnnotationZoomRange`
- Fixed tolerance level when generating the GeoJSONVT shape tiler in `ShapeAnnotationImpl::updateTileData` - this change makes the subsequent query for `shapeTile.features` return a non-empty `TileFeature`.
- However, `TileFeature` still contains an empty tile ring in its geometry. This is what makes `queryRenderedFeatures` return no features in the test case, and causes the annotations to disappear in zoom 19.

For the last item, I'm suspecting either 1) an erroneous parameter passed out to `GeoJSONVT` via `Options` struct or 2) an unknown bug inside `geojson-vt-cpp`.

Fixes #5505.

:eyes: @mourner @kkaefer @jfirebaugh


brunoabinader:
I've compiled `geojson-vt-cpp` 4.1.2 (current version @ GL Native) in debug mode and got the following output:
```
tile z18-131071-131072 (features: 1, points: 2, simplified: 2)
creation: 0.01ms
drilling down: 0.05ms
shapeFeature.tileGeometry.shapeRing.size: 2
drilling down to z18-131071-131071
found parent tile z17-65535-65535
clipping: 0.01ms
tile z18-131071-131071 (features: 1, points: 2, simplified: 2)
creation: 0.01ms
drilling down: 0.05ms
shapeFeature.tileGeometry.shapeRing.size: 2
drilling down to z18-131072-131072
found parent tile z17-65536-65536
clipping: 0.01ms
tile z18-131072-131072 (features: 1, points: 2, simplified: 2)
creation: 0.01ms
drilling down: 0.04ms
shapeFeature.tileGeometry.shapeRing.size: 2
drilling down to z18-131072-131071
found parent tile z17-65536-65535
clipping: 0.02ms
tile z18-131073-131070 (features: 1, points: 2, simplified: 2)
creation: 0.01ms
tile z18-131072-131071 (features: 1, points: 2, simplified: 2)
creation: 0.01ms
tile z18-131072-131070 (features: 1, points: 2, simplified: 2)
creation: 0.00ms
drilling down: 0.07ms
shapeFeature.tileGeometry.shapeRing.size: 2
drilling down to z19-262144-262143
found parent tile z18-131072-131071
clipping: 0.03ms
tile z19-262145-262142 (features: 1, points: 2, simplified: 0)
creation: 0.01ms
tile z19-262144-262143 (features: 1, points: 2, simplified: 0)
creation: 0.01ms
tile z19-262144-262142 (features: 1, points: 2, simplified: 0)
creation: 0.01ms
drilling down: 0.11ms
shapeFeature.tileGeometry.shapeRing.size: 0
drilling down to z19-262144-262144
found parent tile z18-131072-131072
clipping: 0.01ms
tile z19-262144-262144 (features: 1, points: 2, simplified: 0)
creation: 0.01ms
drilling down: 0.04ms
shapeFeature.tileGeometry.shapeRing.size: 0
drilling down to z19-262143-262143
found parent tile z18-131071-131071
clipping: 0.01ms
tile z19-262143-262143 (features: 1, points: 2, simplified: 0)
creation: 0.01ms
drilling down: 0.04ms
shapeFeature.tileGeometry.shapeRing.size: 0
drilling down to z19-262143-262144
found parent tile z18-131071-131072
clipping: 0.01ms
tile z19-262143-262144 (features: 1, points: 2, simplified: 0)
creation: 0.01ms
drilling down: 0.03ms
shapeFeature.tileGeometry.shapeRing.size: 0
```

As it seems, tile generation for z19 returns empty simplified features. Interesting to check with latest geojson-vt-cpp to see if the same behavior happens (/cc @yhahn).

mourner:
Capturing from chat:

> @brunoabinader: given the unit test I added yesterday to geojson-vt-cpp doesn't capture the issue I see in https://github.com/mapbox/mapbox-gl-native/pull/5517#issuecomment-229644335 - I'm ought to believe that this has been fixed between versions 4.1.2 and 6.0.0

src/mbgl/style/source_impl.cpp
```diff
@@ -207,16 +207,23 @@ static Point<int16_t> coordinateToTilePoint(const UnwrappedTileID& tileID, const
 }
 
 std::unordered_map<std::string, std::vector<Feature>> Source::Impl::queryRenderedFeatures(const QueryParameters& parameters) const {
+    std::unordered_map<std::string, std::vector<Feature>> result;
+    if (renderTiles.empty()) {
```
jfirebaugh: The algorithm works fine without checking these boundary conditions right? Why increase the amount of code?

src/mbgl/style/style.cpp
```diff
@@ -337,14 +337,17 @@ RenderData Style::getRenderData(MapDebugOptions debugOptions) const {
...
+    std::vector<Feature> result;
     std::unordered_map<std::string, std::vector<Feature>> resultsByLayer;
 
     for (const auto& source : sources) {
         auto sourceResults = source->baseImpl->queryRenderedFeatures(parameters);
         std::move(sourceResults.begin(), sourceResults.end(), std::inserter(resultsByLayer, resultsByLayer.begin()));
     }
 
-    std::vector<Feature> result;
+    if (resultsByLayer.empty()) {
```
jfirebaugh: Ditto.

src/mbgl/util/grid_index.cpp
```diff
@@ -28,9 +29,10 @@ void GridIndex<T>::insert(T&& t, const BBox& bbox) {
     auto cx2 = convertToCellCoord(bbox.max.x);
     auto cy2 = convertToCellCoord(bbox.max.y);
 
-    for (int32_t x = cx1; x <= cx2; x++) {
-        for (int32_t y = cy1; y <= cy2; y++) {
-            auto cellIndex = d * y + x;
+    int32_t x, y, cellIndex;
```
jfirebaugh: What's the reason for this change?

src/mbgl/style/source_impl.cpp
brunoabinader: This is an early return - avoids unnecessary creation of `queryGeometry`.

src/mbgl/style/style.cpp
brunoabinader: Same reason as before - this is an early return.

PR #5527, opened 2016-06-30
[node] Fix leak and add test

tmpsantos:
This PR changes the API, ideally it should stay the same.


mikemorris:
Moving forward with #5529 instead.

PR #5529, opened 2016-07-01
refactor NodeRequest to stop leaking

mikemorris:
Stupid fancy refactored code still fails leak test. Will investigate further tomorrow.

/cc @tmpsantos


platform/node/src/node_request.cpp
```diff
@@ -3,158 +3,184 @@
...
-    Nan::Set(instance, Nan::New("kind").ToLocalChecked(), Nan::New<v8::Integer>(int(resource.kind)));
-
-    return scope.Escape(instance);
+NodeRequestWorker::NodeRequestWorker(
+    v8::Local<v8::Object> nodeMapHandle_,
+    const mbgl::Resource& resource,
+    mbgl::FileSource::Callback fileSourceCallback_)
+    : AsyncWorker(new Nan::Callback([&] {
+        Nan::EscapableHandleScope scope;
+        return scope.Escape(Nan::New<v8::Function>(
```
tmpsantos: My understanding that `v8::Function` has its lifetime bound to the context.

mikemorris:
This pull request now implements an alternative to https://github.com/mapbox/mapbox-gl-native/pull/5527 that solves the `v8::FunctionTemplate` memory leak while preserving the callback-passing function factory API, but at the cost of [creating a new `v8::Context`](https://twitter.com/indutny/status/751650877279236096) for each callback it returns to JavaScript-land.

This is roughly equivalent to creating an `iframe` to run each function in, and seems to be necessary because of the architecture of `v8::FunctionTemplate` and `v8::Function`, _which are designed to only be called **once** in a single context_ (the global context by default), and calling the function simply creates a new _instance_ of it.

The problem here is that attaching data (like a reference to the `NodeRequest` object) must be done early, during the creation of the `v8::FunctionTemplate`, and an instance of the function cannot be created in C++ then passed on to JavaScript (to attach data later with `Nan::SetInternalFieldPointer` or `v8::SetHiddenValue`) as calling `v8::Function::NewInstance` actually executes the function.

> In terms of CPU time and memory, it might seem an expensive operation to create a new execution context given the number of built-in objects that must be built. However, V8's extensive caching ensures that, while the first context you create is somewhat expensive, subsequent contexts are much cheaper.

According to the [V8 Embedder's Guide](https://developers.google.com/v8/embed#contexts), this shouldn't be TOO costly, and has been benchmarked to verify that performance is acceptable.

Changing the API to `req.respond`, as in https://github.com/mapbox/mapbox-gl-native/pull/5527, would still likely be the more performant option, albeit less idiomatic JavaScript, and may be worth considering in the future.

/cc @miccolis @bsudekum @tmpsantos @brunoabinader @jfirebaugh

jfirebaugh:
This looks good to me.

I wonder what the reason is for the v8 design decision to make FunctionTemplates non-garbage collectable... pretty crazy that you have to make an entire context just to get a temporary function.

PR #5532, opened 2016-07-01
Code style cleanups

kkaefer:
- puts function definitions in a `namespace ... {}` rather than `using namespace ...;`
- remove trailing whitespace
- add trailing newline
- protect SQL statements from being formatted by clang-format

PR #5534, opened 2016-07-01
[core] Fix overdraw mode on Linux

### LINKED ISSUE #5435: [linux] Overdraw mode crashes with GL_INVALID_OPERATION
Console output:
```
[ERROR] {mapbox-glfw}[OpenGL]: GL_DEBUG_SOURCE_API GL_DEBUG_TYPE_ERROR 1282 GL_DEBUG_SEVERITY_HIGH - GL_INVALID_OPERATION error generated. <location> is invalid.
terminate called after throwing an instance of 'mbgl::gl::Error'
  what():  glUniformMatrix4fv(location, 1, GL_FALSE, t.data()): Error GL_INVALID_OPERATION - ../../src/mbgl/shader/uniform.cpp:49
Aborted (core dumped)
```

Problem seems related to the location of uniforms from a shader using two quasi-identical programs. In macOS, it looks like those uniforms locations are the same, but it doesn't seem to be the case on Linux.

brunoabinader:
- Use glBindAttribLocation for GLSL attributes.
- Create a separate shader for each shader that supports overdraw.
  Needed because each uniform location must be known for every program.
- Create a separate VAO for each shader inside buckets.
  Needed because we can only bind a VAO to a specific shader.

Grabbed some shader cleanups from https://github.com/mapbox/mapbox-gl-native/pull/5532 - thanks @kkaefer!

Fixes #5435.

:eyes: @kkaefer @jfirebaugh @zugaldia @tmpsantos

### DIFF
--- a/package.json
+++ b/package.json
@@ -23,7 +23,7 @@
     "express": "^4.11.1",
     "mapbox-gl-shaders": "mapbox/mapbox-gl-shaders#59e998295d548f208ee3ec10cdd21ff2630e2079",
     "mapbox-gl-style-spec": "mapbox/mapbox-gl-style-spec#194fc55b6a7dd54c1e2cf2dd9048fbb5e836716d",
-    "mapbox-gl-test-suite": "mapbox/mapbox-gl-test-suite#cb733d358a95d46344c2212f1546ed94022718f7",
+    "mapbox-gl-test-suite": "mapbox/mapbox-gl-test-suite#2e5addbbd6b3eafaa2984bfd863fd28071f2e481",
     "node-gyp": "^3.3.1",
     "request": "^2.72.0",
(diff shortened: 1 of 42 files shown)


kkaefer:
Any reason for renaming the shader variables?

brunoabinader:
> Any reason for renaming the shader variables?

You mean https://github.com/mapbox/mapbox-gl-native/pull/5534/commits/a17b0bcd82ebe8628b69588d19773d26596b86da#diff-c4566dff12c9f7a41bf5c7e847a3fa95R24 ?

If so, the reason was because we needed an _alias_ for the current shader in use, which could be either the original or overdraw shader. I used a different name to avoid name clashing.

PR #5533, opened 2016-07-01
Update to Android NDK r12b

kkaefer:


kkaefer:
Looks like `log2` is broken in the current Android NDK/stdlib.

bleege:
Thanks @kkaefer!  🚀

Issue #3841, opened 2016-02-06
fix debug/collision,tile,tile-overscaled render tests


ansis:
Some of the changes might need to be made in -js.

it looks like it's a bunch of things:
- a collision-box difference box. Probably something to do with float/integer conversions
- no tile id in -native
- native also has annotation tiles rendering borders

![screen shot 2016-02-05 at 6 07 56 pm](https://cloud.githubusercontent.com/assets/1421652/12863946/8f4e32b8-cc33-11e5-9b1c-116f540bb03b.png)
![screen shot 2016-02-05 at 6 08 06 pm](https://cloud.githubusercontent.com/assets/1421652/12863947/8f603e22-cc33-11e5-8e27-bf3aef007561.png)

jfirebaugh:
See https://github.com/mapbox/mapbox-gl-native/issues/6410#issuecomment-249002027 for potential causes of the first issue, and https://github.com/mapbox/mapbox-gl-test-suite/tree/99844eaf09622b0972e740f87cd9d73487e245f5/render-tests/text-rotation-alignment for the version of the test that revealed those issues.

PR #5540, opened 2016-07-02
[core] Enforce solid fragColor for debug frame/text

brunoabinader:
Fixes an issue with rendering debug frame/text being too opaque in certain zooms.

| Before | After |
|:--:|:--:|
| screen shot 2016-07-02 at 1 03 54 pm | screen shot 2016-07-02 at 1 04 07 pm |

:eyes: @kkaefer @tmpsantos


brunoabinader:
This partially fixes `debug/tile-overscaled` render test - see  https://github.com/mapbox/mapbox-gl-native/issues/3841 - but debug text is still missing.

PR #5542, opened 2016-07-02
[android] Fix gestures when MapView is not full screen

### LINKED ISSUE #5528: Rotation doesn't work on Android when map is not full screen
From my debugging, `RotateGestureDetector` doesn't seem to detect any gesture (`onRotateBegin` never gets called).

The issue is in `TwoFingerGestureDetector.getRawX()`. See my comment below.

**Platform:** Android - Samsung Galaxy S5
**Mapbox SDK version:** 4.1.0

### Steps to trigger behavior

 1. Position MapView in the bottom half of the screen
 2. Run app

### Expected behavior

MapView should rotate when rotating 2 pinched fingers on the screen

### Actual behavior

MapView just zooms with no rotation

dapetcu21:
This fixes #5528

### DIFF
--- a/platform/android/MapboxGLAndroidSDK/src/main/java/com/almeros/android/multitouch/gesturedetectors/TwoFingerGestureDetector.java
+++ b/platform/android/MapboxGLAndroidSDK/src/main/java/com/almeros/android/multitouch/gesturedetectors/TwoFingerGestureDetector.java
@@ -133,7 +133,7 @@ public abstract class TwoFingerGestureDetector extends BaseGestureDetector {
      * @return Raw x value or 0
      */
     protected static float getRawX(MotionEvent event, int pointerIndex) {
-        float offset = event.getX() - event.getRawX();
+        float offset = event.getRawX() - event.getX();
         if (pointerIndex < event.getPointerCount()) {
             return event.getX(pointerIndex) + offset;
         }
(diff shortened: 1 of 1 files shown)


tobrun:
@dapetcu21 thank you for reaching out and contributing! Will review soon.

tobrun:
I'm going to push the related branch to our repo so our tests can run

tobrun:
@dapetcu21 you can follow along the tests in #5562. 
Closing this in favor of #5562.

mb12:
@tobrun and @dapetcu21 gesture detection code has been copied from this github repository. It would be good to open an issue there as well linking to this one.

https://github.com/Almeros/android-gesture-detectors

tobrun:
@mb12 you are correct, I didn't notice this was found in the code of the dependency, will need to investigate more thoroughly before merging. Thanks for the ping.

dapetcu21:
Cool, but that project seems to be abandoned. Last commit was 2 years ago and nobody seems to respond to issues. I think opening an issue/PR over there wouldn't hurt, but I don't think we need to block this issue for that one.

tobrun:
Noting that we are thinking of creating our own gesture lib that is based or not on Almeros library. Main idea behind this is that the project is abandoned and we're already shipping custom versions of the library inside our SDK. I'm suggesting to go ahead and merge https://github.com/mapbox/mapbox-gl-native/pull/5562 and if we are going to separate the maps sdk from the gesture library we should use the one found in this library as base.

cc @zugaldia

zugaldia:
All: Any good (and maintained) gesture library handling that we should be thinking about besides Almeros?

PR #5554, opened 2016-07-04
mapbox: URL containing query string causes infinite loop

jingsam:
If `getMapboxURLPathname` parse a url with querystring, such as `mapbox://p/a/t/h?query=aa`, an infinite loop happens and crash your computer. This is a serious bug which will affect all platforms.  You should fix it ASAP!

### DIFF
--- a/test/util/mapbox.cpp
+++ b/test/util/mapbox.cpp
@@ -7,34 +7,65 @@
 
 using namespace mbgl;
 
-// TODO: correct all EXPECT_EQ(actual, expected) to EXPECT_EQ(expected, actual)
 
 TEST(Mapbox, SourceURL) {
-    EXPECT_EQ(mbgl::util::mapbox::normalizeSourceURL("mapbox://user.map", "key"), "https://api.mapbox.com/v4/user.map.json?access_token=key&secure");
-    EXPECT_EQ(mbgl::util::mapbox::normalizeSourceURL("mapbox://user.map", "token"), "https://api.mapbox.com/v4/user.map.json?access_token=token&secure");
-    EXPECT_THROW(mbgl::util::mapbox::normalizeSourceURL("mapbox://user.map", ""), std::runtime_error);
+    EXPECT_EQ(
+        "https://api.mapbox.com/v4/user.map.json?access_token=key&secure",
+        mbgl::util::mapbox::normalizeSourceURL("mapbox://user.map", "key"));
+    EXPECT_EQ(
+        "https://api.mapbox.com/v4/user.map.json?access_token=token&secure",
+        mbgl::util::mapbox::normalizeSourceURL("mapbox://user.map", "token"));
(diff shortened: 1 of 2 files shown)


test/util/mapbox.cpp
```diff
@@ -7,34 +7,65 @@
...
 TEST(Mapbox, SourceURL) {
-    EXPECT_EQ(mbgl::util::mapbox::normalizeSourceURL("mapbox://user.map", "key"), "https://api.mapbox.com/v4/user.map.json?access_token=key&secure");
-    EXPECT_EQ(mbgl::util::mapbox::normalizeSourceURL("mapbox://user.map", "token"), "https://api.mapbox.com/v4/user.map.json?access_token=token&secure");
-    EXPECT_THROW(mbgl::util::mapbox::normalizeSourceURL("mapbox://user.map", ""), std::runtime_error);
+    EXPECT_EQ(
+        "https://api.mapbox.com/v4/user.map.json?access_token=key&secure",
+        mbgl::util::mapbox::normalizeSourceURL("mapbox://user.map", "key"));
+    EXPECT_EQ(
+        "http://path",
+        mbgl::util::mapbox::normalizeSourceURL("http://path", "key");
```
1ec5: This line is missing a closing parenthesis.

test/util/mapbox.cpp
```diff
@@ -7,34 +7,65 @@
...
+    EXPECT_THROW(
+        std::runtime_error,
+        mbgl::util::mapbox::normalizeSourceURL("mapbox://user.map", ""));
```
1ec5: Unlike `EXPECT_EQ`, the order of the parameters in `EXPECT_THROW` matter: the error type needs to be in the second parameter, not the first.

jingsam:
Should be OK now. @1ec5

1ec5:
Verified that the tests build and pass locally. Thanks for the fix!

jingsam:
I checked that the implement in `mapbox.cpp` would have some potential bugs when URL contains query string. So I plan to refactor this file to fully compatible with query string and hash tags

jingsam:
Note that my development environment is an Ubuntu server VM on Windows 10. Mapbox-gl-native can not run on my VM, it is a bit of un-convenient to test my modification

1ec5:
Unfortunately, I’m not very familiar with our Linux build support, but [here are instructions](https://github.com/mapbox/mapbox-gl-native/tree/master/platform/linux#build) in case you haven’t seen them. Otherwise, another option might be to [build the Android SDK](https://github.com/mapbox/mapbox-gl-native/blob/master/platform/android/CONTRIBUTING_LINUX.md) on Linux, although that’s a bit further removed from the code you’re working with.

1ec5:
Updated the iOS and macOS changelogs in a53dfcd675049943443def6b65cc8b654faa7a90.

PR #5561, opened 2016-07-04
[ios] fixes #5544 support canceling a drag operation and not by default

### LINKED ISSUE #5544: Cancel dragging annotation view by setting drag state
In MapKit, you cancel a drag operation by setting the annotation view’s `dragState` property to [`MKAnnotationViewDragStateCanceling`](https://developer.apple.com/reference/mapkit/mkannotationviewdragstate/mkannotationviewdragstatecanceling). In the Mapbox iOS SDK, as of v3.3.0-beta.2, a drag operation is canceled by default; you have to implement `-[MGLMapViewDelegate mapView:didDragAnnotationView:toCoordinate:]` to manually set the new coordinate if you want the drag to succeed. Meanwhile, from reading the code, setting `MGLAnnotationViewDragStateCanceling` appears to have no effect. I think MapKit’s approach is less prone to error, since it does what the developer expects by default.

/cc @frederoni @boundsj

frederoni:
This PR simplifies implementing draggable annotation views by removing `didDragAnnotationView:toCoordinate` and instead call `setCoordinate:` on the associated MGLAnnotation. It also adds support for canceling a drag operation by setting `dragState` to `MGLAnnotationViewDragStateCanceling` as mentioned in [#5544](https://github.com/mapbox/mapbox-gl-native/issues/5544)

👀 @1ec5 @boundsj

### DIFF
--- a/platform/ios/src/MGLAnnotationView.mm
+++ b/platform/ios/src/MGLAnnotationView.mm
@@ -212,13 +212,21 @@
         [self.mapView.calloutViewForSelectedAnnotation dismissCalloutAnimated:animated];
         [self.superview bringSubviewToFront:self];
     }
-
-    if (dragState == MGLAnnotationViewDragStateEnding)
+    else if (dragState == MGLAnnotationViewDragStateCanceling)
     {
-        if ([self.mapView.delegate respondsToSelector:@selector(mapView:didDragAnnotationView:toCoordinate:)])
+        self.panGestureRecognizer.enabled = NO;
+        self.longPressRecognizer.enabled = NO;
+        self.center = [self.mapView convertCoordinate:self.annotation.coordinate toPointToView:self.mapView];
+        self.panGestureRecognizer.enabled = YES;
+        self.longPressRecognizer.enabled = YES;
+        self.dragState = MGLAnnotationViewDragStateNone;
+    }
+    else if (dragState == MGLAnnotationViewDragStateEnding)
+    {
+        if ([self.annotation respondsToSelector:@selector(setCoordinate:)])
         {
             CLLocationCoordinate2D coordinate = [self.mapView convertPoint:self.center toCoordinateFromView:self.mapView];
-            [self.mapView.delegate mapView:self.mapView didDragAnnotationView:self toCoordinate:coordinate];
+            [(NSObject *)self.annotation setValue:[NSValue valueWithMGLCoordinate:coordinate] forKey:@"coordinate"];
         }
     }
 }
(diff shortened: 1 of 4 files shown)


platform/ios/src/MGLAnnotationView.mm
```diff
@@ -212,13 +212,21 @@ - (void)setDragState:(MGLAnnotationViewDragState)dragState animated:(BOOL)animat
...
+    }
+    else if (dragState == MGLAnnotationViewDragStateEnding)
     {
-        if ([self.mapView.delegate respondsToSelector:@selector(mapView:didDragAnnotationView:toCoordinate:)])
+        if ([self.annotation respondsToSelector:@selector(setCoordinate:)])
         {
             CLLocationCoordinate2D coordinate = [self.mapView convertPoint:self.center toCoordinateFromView:self.mapView];
-            [self.mapView.delegate mapView:self.mapView didDragAnnotationView:self toCoordinate:coordinate];
+            [(NSObject *)self.annotation setValue:[NSValue valueWithMGLCoordinate:coordinate] forKey:@"coordinate"];
         }
```
frederoni: Will this work if the class that conforms to MGLAnnotation is a native Swift base class?

platform/ios/src/MGLAnnotationView.mm
```diff
@@ -212,13 +212,21 @@ - (void)setDragState:(MGLAnnotationViewDragState)dragState animated:(BOOL)animat
...
+        self.dragState = MGLAnnotationViewDragStateNone;
+    }
+    else if (dragState == MGLAnnotationViewDragStateEnding)
     {
-        if ([self.mapView.delegate respondsToSelector:@selector(mapView:didDragAnnotationView:toCoordinate:)])
+        if ([self.annotation respondsToSelector:@selector(setCoordinate:)])
         {
             CLLocationCoordinate2D coordinate = [self.mapView convertPoint:self.center toCoordinateFromView:self.mapView];
-            [self.mapView.delegate mapView:self.mapView didDragAnnotationView:self toCoordinate:coordinate];
+            [(NSObject *)self.annotation setValue:[NSValue valueWithMGLCoordinate:coordinate] forKey:@"coordinate"];
```
1ec5: Clever. I wonder if there’s any difference between this and using NSInvocation.

frederoni: @1ec5 I hesitated between this and NSInvocation. Wasn't sure setValue:forKey would work if the class would be Swift native base class but fortunately, the protocol ensures NSObject.

1ec5: Right, fortunately, we can assume Objective-C dispatch even in Swift, thanks to the MGLAnnotation protocol. I think this KVO approach will be fine as long as the class doesn’t do any fancy KVO tricks to forward the `coordinate` key path to a selector other than `-setCoordinate:`.

Theoretically, if it’s an NSProxy that happens to respond to `-setValue:forKey:`, casting to `NSObject *` could be problematic, but that’s really going off into asking-for-trouble land.

PR #5562, opened 2016-07-04
[android] Fix gestures when MapView is not full screen

### LINKED ISSUE #5528: Rotation doesn't work on Android when map is not full screen

tobrun:
closes #5528, refs #5542


tobrun:
Capturing from @mb12 in https://github.com/mapbox/mapbox-gl-native/pull/5542#issuecomment-230300334 that we need to look into resolving this upstream.

PR #5569, opened 2016-07-05
[ios] fixes #4536 include podspec in iOS releases

### LINKED ISSUE #4536: Include podspec for static framework in iOS releases
Including the podspec file in iOS releases will make it simpler to install releases as local development pods.

frederoni:
Dynamic and static framework w/ and w/o symbols seems to be working except for one case;
- `platform :ios, '7.0'` in Podfile
- `Deployment target set to 7.0` in Xcode (not using `use_frameworks!`)

Yields:
```
[!] Unable to satisfy the following requirements:

- `Mapbox-iOS-SDK (from `../mapbox-gl-native/build/ios/pkg/dynamic/Mapbox-iOS-SDK-symbols.podspec`)` required by `Podfile`

It seems like you've changed the constraints of dependency `Mapbox-iOS-SDK` inside your development pod `Mapbox-iOS-SDK`.
You should run `pod update Mapbox-iOS-SDK` to apply changes you've made.
```
- `pod update Mapbox-iOS-SDK` yields the same error message.

Podfile
```
platform :ios, '7.0'
target 'podtest' do
pod 'Mapbox-iOS-SDK', :path => "../mapbox-gl-native/build/ios/pkg/dynamic/Mapbox-iOS-SDK-symbols.podspec"
end
```
👀 @boundsj

### DIFF
--- a/platform/ios/scripts/package.sh
+++ b/platform/ios/scripts/package.sh
@@ -178,11 +178,25 @@ if [[ "${GCC_GENERATE_DEBUGGING_SYMBOLS}" == false ]]; then
     fi
 fi
 
+function create_local_podspec {
+    step "Creating local podspec"
+    POD_SOURCE_PATH='    :path => ".",'
+    POD_FRAMEWORKS="  m.vendored_frameworks = '"${NAME}".framework'"
+    [[ $SYMBOLS = YES ]] && POD_SUFFIX="-symbols" || POD_SUFFIX=""
+    POD_LOCALSPEC=${OUTPUT}/$1/${NAME}-iOS-SDK${POD_SUFFIX}.podspec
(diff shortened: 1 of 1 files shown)


1ec5:
The minimum deployment target should be iOS 8. iOS 7 doesn’t officially support dynamic frameworks.

1ec5:
The dynamic podspecs still specify iOS 7.0 as the minimum deployment target? It should be 8.0.

Also, the static podspecs don't refer to the resource bundle at all, so I think there'd be a crash on launch unless the resources are getting pulled in some other way.

1ec5:
#5595 backs out the podspecs for the static framework. We’ll revisit that in v3.3.1.

Issue #5037, opened 2016-05-14
Custom user location annotation view


1ec5:
MGLMapView should call `-[MGLMapViewDelegate mapView:viewForAnnotation:]` to get the MGLAnnotationView to display for MGLUserLocation. This would allow client code to completely customize the user dot’s appearance. If `-mapView:viewForAnnotation:` returns `nil`, MGLMapView should use an MGLUserLocationAnnotationView by default.

/cc @boundsj @friedbunny

1ec5:
This will be a lot more elegant once we implement #5039.

KevinHu2014:
@1ec5 
How about Android?
Is it possible to custom user location annotation view?

tobrun:
@KevinHu2014 at this point we haven't migrated the `MyLocationView` component on Android to use annotation views yet but we expose some API to customise it. Take a look at [MyLocationViewSettings](https://github.com/mapbox/mapbox-gl-native/blob/master/platform/android/MapboxGLAndroidSDK/src/main/java/com/mapbox/mapboxsdk/maps/widgets/MyLocationViewSettings.java), this class contains our public API to change the appearance of `MyLocationView`.

KevinHu2014:
@tobrun Thanks a lot~

friedbunny:
Fixed by #5882.

Issue #5571, opened 2016-07-05
Tapping annotation within user location accuracy halo selects user location annotation


matrejek:
**Platform:** iOS 9.3
**Mapbox SDK version:** 3.3.0 beta 3

### Steps to trigger behavior
(Screenshot below) 
 1. Add annotation near user location 
 2. Set it to show callout view
 3. When annotation is in range of user location accuracy (blue circle) it is impossible to show callout view (marker is not tappable). Instead of that we get callout for user location annotation

### Expected behavior

When annotation in blue circle is tapped it should show its callout. 

### Actual behavior

When annotation in blue circle is tapped user location annotation is presented. 

![aaa](https://cloud.githubusercontent.com/assets/1874050/16579222/ffe034a4-429e-11e6-8c32-2999e9fe3484.png)

1ec5:
This issue reproduces in v3.3.0 but not v3.2.0.

1ec5:
The regression was introduced in #1582. We now hit-test the user location annotation view based on its `layer.presentationLayer`, since the view can move around with implicit animation in response to the user location changing. We apparently should be hit-testing based on the `presentationLayer` of `_puckDot` or `_dotLayer`.

friedbunny:
A fix for this should probably be done in conjunction with @frederoni’s refactor in #5573 or wait for that to land.

1ec5:
I think we can land a more targeted fix in v3.3.2.

friedbunny:
Fixed by #5816.

fabioknoedt:
The problem is still happening in the v3.3.3. I can't select an annotation within the user location's accuracy range.

friedbunny:
True, #5816 excluded the pulsing halo layer but not the static accuracy ring. 

e4a9173a2a5c585dda3fa7a0f955cff1552f8d25 in #5882 includes a more comprehensive fix that specifically hit-checks the main dot. That will go in v3.4.0, but we could consider porting a similar fix to yet another patch release... @1ec5?

fabioknoedt:
It would be awesome to have that fix as soon as possible. Now I'm stuck to the release v3.2.3 because of that. I want to start using MGLAnnotationView with the `selected` state support but I can't have this problem in the app. Thanks!

friedbunny:
This was fixed in #5894 and is now available in the [v3.3.4 release](https://github.com/mapbox/mapbox-gl-native/releases/tag/ios-v3.3.4).

[...]

PR #5573, opened 2016-07-05
[ios] #5039 refactored MGLUserLocationAnnotationView

### LINKED ISSUE #5039: MGLUserLocationAnnotationView should inherit from MGLAnnotationView
Now that MGLAnnotationView has landed in #4801, we should make MGLUserLocationAnnotationView, inherit from it. This will be important for consistency once we implement #5037.

/cc @boundsj @friedbunny

frederoni:
Fixes #5039 and #5037
`MGLUserLocationAnnotationView` now inherits from `MGLAnnotationView`.
`MGLMapView` calls `-[MGLMapViewDelegate mapView:viewForAnnotation:]` for `MGLUserLocation` to support a custom user dot. If the implementer still wants the default user dot, they should return nil in case annotation equals `mapView.userLocation`


frederoni:
**This proposal is very much work in progress.**

The UI part of the code in `MGLUserLocationAnnotationView` has been moved to `MGLFaux3DUserLocationAnnotationView` (naming suggestions are more than welcome) leaving only crucial parts like accessibility and references to userLocation and mapView in the base class. With this abstraction we can untangle some calls from `MGLMapView` like updating the visibility of the haloLayer and setting up layers. This can now be achieved by overriding `-didUpdateUserLocation:` from a `MGLUserLocationAnnotationView`subclass. It is a bit redundant since map view already informs the delegate with `-mapView:didUpdateUserLocation:` but it makes it a bit cleaner when you subclass a user annotation view.

`MGLUserLocationAnnotationView` would become public and the new `MGLFaux3DUserLocationAnnotationView` or whatever we decide to name it would be private.

👀 @friedbunny @boundsj @1ec5

platform/ios/app/MBXUserLocationAnnotationView.h
```diff
@@ -0,0 +1,5 @@
+#import "MGLUserLocationAnnotationView.h"
```
1ec5: You should be able to just import the Mapbox umbrella header or module. But this PR is missing changes to Mapbox.h and the project file that would expose MGLUserLocationAnnotationView.h publicly.

1ec5:
MGLUserLocationAnnotationView.h needs to be exposed publicly. See [these instructions](https://github.com/mapbox/mapbox-gl-native/blob/master/platform/ios/DEVELOPING.md#adding-a-source-code-file) for details.

platform/ios/src/MGLUserLocationAnnotationView.h
```diff
@@ -9,15 +10,15 @@ NS_ASSUME_NONNULL_BEGIN
 @class MGLUserLocation;
 
 /** View representing an MGLUserLocation on screen. */
-@interface MGLUserLocationAnnotationView : UIView
+@interface MGLUserLocationAnnotationView : MGLAnnotationView
 
 @property (nonatomic, weak) MGLMapView *mapView;
-@property (nonatomic) MGLUserLocation *annotation;
-@property (nonatomic, readonly, nullable) CALayer *haloLayer;
+@property (nonatomic, readonly, nullable) MGLUserLocation *userLocation;
```
1ec5: Now that this header is public, it needs documentation for each symbol.

platform/ios/src/MGLUserLocationAnnotationView.h
```diff
@@ -9,15 +10,15 @@ NS_ASSUME_NONNULL_BEGIN
...
 
-- (instancetype)initInMapView:(MGLMapView *)mapView NS_DESIGNATED_INITIALIZER;
-- (void)setupLayers;
+- (instancetype)initInMapView:(MGLMapView *)mapView userLocation:(MGLUserLocation *)userLocation;
```
1ec5: I know we’ve done it this way for a long time, but I’m not sure we want to make public an API that’s based on a backpointer. It’s just too tempting for subclasses to abuse this backpointer in ways that’ll leak memory. Instead, we should probably figure out a way to push the necessary state to this class.

platform/ios/src/MGLFaux3DUserLocationAnnotationView.m
```diff
@@ -0,0 +1,522 @@
+#import "MGLFaux3DUserLocationAnnotationView.h"
+
+#import "MGLMapView.h"
+#import "MGLUserLocation_Private.h"
```
1ec5: Does this file use MGLUserLocation_Private.h for anything, or can it import MGLUserLocation instead? Ideally, a developer could theoretically implement anything that’s in this class in their own custom subclass of MGLUserLocationAnnotationView.

1ec5:
Since you’re exposing MGLUserLocationAnnotationView publicly for the first time, you’ll also need to add it to jazzy’s table of contents. In d0766120ca2513e44584f12014283de4c361372b, I updated the contributing guide to point this out. Feel free to add anything that appears to be missing.

platform/ios/app/MBXUserLocationAnnotationView.m
```diff
@@ -9,46 +10,168 @@ @interface MBXUserLocationAnnotationView()
 
 @implementation MBXUserLocationAnnotationView
 
+- (instancetype)initWithFrame:(CGRect)frame
+{
+    self = [super initWithFrame:frame];
```
1ec5: Nit: The structure of this implementation works, but we should adhere to the format used in the initializer code snippet. Consistency with initializers makes it easier for us to spot errors like missing assignments to self or return statements.

```objc
if (self = [super initWithFrame:frame]) {
    self.backgroundColor = [UIColor clearColor];
}
return self;
```

1ec5:
What’s a good way to address https://github.com/mapbox/mapbox-gl-native/issues/5571#issuecomment-234372579? Do we need to expose a new property on MGLUserLocationAnnotationView for the tappable layer or Bézier path?

frederoni:
The frame is correct and the halo layer is drawn out of bounds.
screenshot 2016-07-25 09 49 46
Keep it simple and just hit-test the view's frame?

1ec5:
The problem is that the view is animated using implicit animation. As the user moves in a straight line, the frame jumps to the CLLocation while the visible dot animates smoothly. The user should be able to tap the dot without having to guess where it’ll wind up.

friedbunny:
I rebased this on master and then force-pushed. This PR targets the v3.3.0 release branch, so I’m closing it in favor of https://github.com/mapbox/mapbox-gl-native/pull/5882.

Issue #5456, opened 2016-06-23
Run core tests on Android


ivovandongen:
In order to test the android bindings for core-gl we need to run the core unit tests on Android.

There two ways to run the tests:
- Create a native executable, linked with the tests and execute from the adb shell
- Create an apk with a small Java class to kick things of like #4671

Furthermore, there are a couple types of tests:
- 'Simple' tests - no dependencies on fs, network, etc
- Fixture tests - need fixures from the fs to run
- Network tests - need an http server
- Render tests - need a headless renderer

ivovandongen:
I started of with the a native executable running only the 'simple' tests. With great help from @tmpsantos we got some 20 unit tests to run on the device. However, it seems that these executables have no access to any kind of vm (as they're not started from dalvik/art). Proceeding with the approach in #4671.

Also, it only compiles with `arm-v7`. @jfirebaugh suggested updating the ndk. Adding that to the list.

tobrun:
Linking https://github.com/mapbox/mapbox-gl-native/issues/2150

ivovandongen:
Progress - Runs on a vm now, properly linked with gtest in a shared lib, from the command line. But only for a subset of the tests still. A couple of hacks were needed to get things going:
- Got rid of the `-whole-archive` in test/test.gyphi as it gives linking errors when linking into a shared library - need to make that dependent on the target os.
- Commented out [one line](https://github.com/mapbox/mapbox-gl-native/blob/android-core-tests/platform/android/src/jni.cpp#L1725) in jni.cpp as it made it crash - no ideas yet
- Still commented out tests that need external resources or a headless renderer - next steps

ivovandongen:
Concerning the crash in jni.cpp:

```
********** Crash dump: **********
Build fingerprint: 'google/bullhead/bullhead:6.0.1/MTC19V/2862947:user/release-keys'
pid: 32683, tid: 32683, name: main  >>> dalvikvm32 <<<
signal 6 (SIGABRT), code -6 (SI_TKILL), fault addr --------
Stack frame #00 pc 00042374  /system/lib/libc.so (tgkill+12)
Stack frame #01 pc 0003ff81  /system/lib/libc.so (pthread_kill+32)
Stack frame #02 pc 0001c73f  /system/lib/libc.so (raise+10)
Stack frame #03 pc 000198f1  /system/lib/libc.so (__libc_android_abort+34)
[...]

PR #5575, opened 2016-07-05
[android] Fix bugs on low level platform abstractions

tmpsantos:
With #5456 started, we could run the basic unit tests on Android and already found some bugs fixed by this PR.

/cc @ivovandongen


ivovandongen:
@tmpsantos Thanks so much for picking this up! Looks good. Tested on x86 and arm-v7 with the test-app and I'm not noticing any regressions.

Issue #5577, opened 2016-07-05
On first app run, cache.db is backed up to iCloud


dapetcu21:
I found the following bug:

When you first open the app, there is no `cache.db` to begin with. `setResourceValue:forKey:NSURLIsExcludedFromBackupKey` is getting [called](https://github.com/mapbox/mapbox-gl-native/blob/9c9140595e9c1d9a12120e167421034f6036b08a/platform/darwin/src/MGLOfflineStorage.mm#L94) on the file, but you have no guarrantees that the file is created at that point and the call won't fail silently.

So, on the first run, `NSURLIsExcludedFromBackupKey` might be set before `cache.db` is created, hence being left with `cache.db` being backed up by iCloud, automatically rejecting your app when reviewed on the App Store.

A simple and effective solution (that I employed in my app) is to set `NSURLIsExcludedFromBackupKey` on the cache directory, which is [ensured to exist](https://github.com/mapbox/mapbox-gl-native/blob/9c9140595e9c1d9a12120e167421034f6036b08a/platform/darwin/src/MGLOfflineStorage.mm#L60), instead of on the file.

I'll submit a PR in a sec.

**Platform:** iOS
**Mapbox SDK version:** 3.2.0

### Steps to trigger behavior

 1. Use offline maps in your app
 2. Download a big enough offline pack in the first run of the app
 3. Publish the app to the App Store

### Expected behavior

Though big in size, `cache.db` will be marked with `NSURLIsExcludedFromBackupKey`, so it will not be backed up. Hence, your app will pass the review process with flying colors.

### Actual behavior

Your app will get rejected because `cache.db` is not properly excluded and you'd end up saving hundreds of MB on the user's iCloud account.

dapetcu21:
If you want to try to reproduce, I suggest using
```
xattr -l ~/Library/Developer/CoreSimulator/Devices/${DEVICE_UUID}/data/Containers/Data/Application/${APP_UUID}/Library/Application\ Support/${APP_BUNDLE_ID}/cache.db
```
to test.

friedbunny:
Thanks again for reporting this and proposing #5578, @dapetcu21.

PR #5578, opened 2016-07-05
[ios] Set backup exclusion attribute on cache directory

dapetcu21:
As explained in #5577, it's safer to set NSURLIsExcludedFromBackupKey to the cache directory instead of the cache file. This must be done before anything is created/moved into the directory so as to prevent iCloud backup from kicking in at any point (during the legacy db move, for example).

### DIFF
--- a/platform/darwin/src/MGLOfflineStorage.mm
+++ b/platform/darwin/src/MGLOfflineStorage.mm
@@ -64,6 +64,12 @@ NSString * const MGLOfflinePackMaximumCountUserInfoKey = @"MaximumCount";
         NSURL *cacheURL = [cacheDirectoryURL URLByAppendingPathComponent:MGLOfflineStorageFileName];
         NSString *cachePath = cacheURL ? cacheURL.path : @"";
         
+        // Avoid backing up the offline cache onto iCloud, because it can be
+        // redownloaded. Ideally, we’d even put the ambient cache in Caches, so
+        // it can be reclaimed by the system when disk space runs low. But
+        // unfortunately it has to live in the same file as offline resources.
+        [cacheDirectoryURL setResourceValue:@YES forKey:NSURLIsExcludedFromBackupKey error:NULL];
+        
         // Move the offline cache from v3.2.0-beta.1 to a location that can also
         // be used for ambient caching.
 #if TARGET_OS_IPHONE || TARGET_OS_SIMULATOR
@@ -86,12 +92,6 @@ NSString * const MGLOfflinePackMaximumCountUserInfoKey = @"MaximumCount";
         }
         
         _mbglFileSource = new mbgl::DefaultFileSource(cachePath.UTF8String, [NSBundle mainBundle].resourceURL.path.UTF8String);
-
-        // Avoid backing up the offline cache onto iCloud, because it can be
-        // redownloaded. Ideally, we’d even put the ambient cache in Caches, so
-        // it can be reclaimed by the system when disk space runs low. But
-        // unfortunately it has to live in the same file as offline resources.
-        [cacheURL setResourceValue:@YES forKey:NSURLIsExcludedFromBackupKey error:NULL];
 
         // Observe for changes to the global access token (and find out the current one).
         [[MGLAccountManager sharedManager] addObserver:self


1ec5:
@friedbunny, v3.3.0 is only a few days away, but I think we should cherry-pick this change into the release branch, given that #5124 was speculative and @dapetcu21 has provided verification steps in https://github.com/mapbox/mapbox-gl-native/issues/5577#issuecomment-230595628.

1ec5:
So there's one problem with this PR: it assumes the developer won't need to put anything else in the application's subfolder in Application Support. I don't think we can make that assumption. Perhaps we need to move the cache yet again to a subfolder within that folder that we can set the exclusion flag on.

/cc @jfirebaugh

friedbunny:
This behavior has changed since v3.2.0 — see #5124. We were indeed failing to set the `NSURLIsExcludedFromBackupKey` flag on the first launch because we tried to set it before cache.db existed. #5124 moved the flagging step to immediately after cache.db is created, which should have been created synchronously.

You can verify that cache.db exists on the first launch and that we’re setting the flag correctly in v3.3.0:

```objc
NSAssert(![[NSFileManager defaultManager] fileExistsAtPath:cachePath], @"Cache database file shouldn't already exist.");

_mbglFileSource = new mbgl::DefaultFileSource(cachePath.UTF8String, [NSBundle mainBundle].resourceURL.path.UTF8String);

NSAssert([[NSFileManager defaultManager] fileExistsAtPath:cachePath], @"Cache database file should exist.");

// Avoid backing up the offline cache onto iCloud, because it can be
// redownloaded. Ideally, we’d even put the ambient cache in Caches, so
// it can be reclaimed by the system when disk space runs low. But
// unfortunately it has to live in the same file as offline resources.
[cacheURL setResourceValue:@YES forKey:NSURLIsExcludedFromBackupKey error:NULL];

NSError *error = nil;
NSNumber *exclusionFlag = nil;
[cacheURL getResourceValue:&exclusionFlag forKey:NSURLIsExcludedFromBackupKey error:&error];
NSAssert(exclusionFlag && [exclusionFlag boolValue], @"Backup exclusion flag should be set for cache database.");
```

We don't set the backup exclusion flag on the directory because it may interfere with legitimately backup-able files that the developer also stores there. If we were to set the flag on a directory, it would need to be one that we’re certain to own exclusively — i.e., a new subdirectory and not the generic `~/Library/Application Support/tld.app.bundle.id` that we use now.

friedbunny:
I should also say: thanks for helping out, @dapetcu21. 🙇

If you’re still seeing the backup exclusion flag not being set properly on the `master` or `release-ios-v3.3.0` branches (as opposed to the v3.2.0 release build), then we’ll have to go ahead and move cache.db into a subdirectory and set the flag on that subdirectory.

dapetcu21:
Wait. I'm using a precompiled `v3.2.0`. I didn't check with `master` or `v3.3.0`. I agree with all the above. Setting it on the dir might not be such a great idea after all since other apps might use that directory. 

When submitting this PR I made 2 assumptions:

a) That the code didn't change between my build and master (which is wrong)
b) That initializing `DefaultFileSource` didn't create the file synchronously. (which, again, might be wrong)

So if b) is wrong, I guess it should work and we should't merge this.

1ec5:
Ah, sorry for the confusion then. I just assumed a PR made against master would fix a bug in master, my bad.

I don’t know whether `DefaultFileSource` creates the file synchronously or asynchronously. If you merely call `-[MGLOfflineStorage sharedOfflineStorage]` to exercise this code but don’t initialize an MGLMapView or refer to the shared offline storage’s `packs` property until later, does the cache.db file get created right away?

friedbunny:
> does the cache.db file get created right away?

Yes, in my observation it is created immediately via the call to `new mbgl::DefaultFileSource()` [here](https://github.com/mapbox/mapbox-gl-native/blob/903d609b40b6d0f4873f7bb46d96f4a06d7b17d6/platform/darwin/src/MGLOfflineStorage.mm#L88).

Frankly, we’re not completely confident in when exactly cache.db being created, but @kkaefer speculated that the SQLite database having [`PRAGMA synchronous = FULL;`](https://www.sqlite.org/pragma.html#pragma_synchronous) enabled by default ([as it is](https://cloud.githubusercontent.com/assets/1198851/16633327/bd91fb7a-4396-11e6-85ba-0cd291bf71a2.png)) should cause the database to be written to disk immediately.

> a) That the code didn't change between my build and master (which is wrong)

Apologies for the confusion, @dapetcu21 — between releases we merge changes into master, then create a short-lived branch for pre-releases and the eventual final release (currently: `release-ios-v3.3.0`) before eventually merging that back to master.

I’m going to close this pull request and #5577, but we’d quite appreciate if you could help to verify the fix for this issue by giving the [current v3.3.0 beta release](https://github.com/mapbox/mapbox-gl-native/releases/tag/ios-v3.3.0-beta.3) a try. Thanks for time you've put into this, @dapetcu21.

PR #5581, opened 2016-07-05
Sets the loading flag in the style request handler to ensure that the…

PaulZed:
… correct Map event fires when a stale style gets refreshed from the server.

### DIFF
--- a/src/mbgl/map/map.cpp
+++ b/src/mbgl/map/map.cpp
@@ -307,6 +307,7 @@ void Map::setStyleURL(const std::string& url) {
         } else if (res.notModified || res.noContent) {
             return;
         } else {
+            impl->loading = true;
             impl->loadStyleJSON(*res.data);
         }
     });


jfirebaugh:
I'm pretty sure we don't want to change the semantics of `MapChangeDidFinishLoadingMap` such that it fires more than once per call to `setStyle{URL,JSON}`. That would likely violate the assumptions of existing users of the event.

What's the underlying need you're trying to satisfy here?

PaulZed:
The problem here is that setStyle doesn't get called when a resource expires and is refetched. This means that the map reloads, but the MapChangeDidFinishLoadingMap event never fires, which is wrong. If someone is relying on that event to say, add a custom layer to the style, then when the style expires that custom layer will disappear forever, until the style is explicitly changed.

PaulZed:
Perhaps a cleaner solution is to not keep the same request around to handle resource expiration,  and explicitly call `setStyleURL` to load the new style when the resource expires.

PaulZed:
@tmpsantos @brunoabinader thoughts?

brunoabinader:
I suggest we add a new enum e.g. `MapChangeDidFinishRefreshStyle` - this ensures that the semantic concerns mentioned by @jfirebaugh are sustained while providing a mean for the client to be notified about it.

tmpsantos:
> MapChangeDidFinishRefreshStyle

I like the idea.

PaulZed:
I don't really understand what the difference is. If you have something that you want to happen whenever the map finishes loading, why wouldn't you want to do it when the map reloads? Can anyone actually come up with an example where semantics are being violated, because i'm struggling to.

jfirebaugh:
@PaulZed Thanks for bringing this to my attention. This exposed a wider issue with the intersection of runtime styling and style revalidation that I think needs discussion independent of this specific proposal. Opened #5665 for discussion.

Issue #5042, opened 2016-05-15
[ios] Restrict panning of map with offline downloaded map


frederik-jacques:
When you download an offline map you provide the `MGLTilePyramidOfflineRegion` initialiser a bounds struct which downloads a specific part of a map.

When you use the offline map, the user is able to scroll around and sees blank tiles for parts outside the map that have been downloaded.

It should be possible to restrict the map to only pan within certain bounds.

You could implement `- (void)mapView:(MGLMapView *)mapView regionDidChangeAnimated:(BOOL)animated`, but then you have to manually check if the current visible screen bounds is outside the downloaded bounds.

In the previous version of the Mapbox SDK this was possible with a certain method where you would pass in the NE en SW coordinate and you couldn't scroll past those.

**Platform:** iOS
**Mapbox SDK version:** 3.2.2
### Steps to trigger behavior
1. Download an offline map
2. Show it on a map
3. Scroll around
### Expected behavior

You can only scroll within the bounds that have been downloaded.
### Actual behavior

You can scroll everywhere, so you get to see blank tiles when the app runs in offline mode.

1ec5:
Previous discussion in #2457. This is complicated by the fact that this SDK allows rotation and tilting, whereas that wasn’t possible with the old SDK. In some cases, it would be OK for the user to see beyond the bounds by rotating and tilting, whereas for the case you describe, that would be suboptimal. Another consideration is that we intend to eventually allow offline regions to be non-rectangular, at which point the `-mapView:regionDidChangeAnimated:` approach would be easier to work with, and #2457 would be even easier.

dsmitchell:
Hey, I'm new here, but what about restricting the MGLCamera's possible coordinate? So then rotating and tilting doesn't really matter.

andrewstay:
This is one of the critical features for stay.com.
Our logic is based around a particular city. We load annotations and map for one city. So we need to restrict user from panning outside the city area.

1ec5:
For starters, I’ve implemented the simplest possible API for restricting movement in #5584. With this API, it should be possible to implement logic to restrict gestures based on regions already downloaded for offline use, but we should keep this ticket open to investigate more convenient or automatic approaches.

andrewstay:
@1ec5 I cannot implement it using `-mapView:regionDidChangeAnimated:` approach as it is blocked by [#7119](https://github.com/mapbox/mapbox-gl-native/issues/7119)

[...]

PR #5584, opened 2016-07-05
Delegate method to restrict movement

### LINKED ISSUE #2457: Let MGLMapViewDelegate cancel region changes
`-[MGLMapViewDelegate mapView:regionWillChangeAnimated:]` should return a Boolean, so that delegates can cancel impending viewport changes by return `NO`. If the viewport change was initiated programmatically, perhaps the return value should be ignored, but it should at least be honored for viewport changes initiated by gesture recognizers.

/cc @friedbunny @incanus

1ec5:
Added a way for the delegate to restrict where the user can move within the map using gestures. This approach allows the developer to restrict to a non-rectangular region or vary the region by zoom level or pitch as necessary, such as to prevent the user from seeing beyond a certain region regardless of the degree of tilt.

This PR intentionally keeps things simple to unblock some key use cases. Future improvements could include implementing a “rubber band” effect, allowing the delegate to “redirect” the camera to some other camera (https://github.com/mapbox/mapbox-gl-native/issues/2457#issuecomment-222051074), and automatically restricting the camera to downloaded regions (#5042).

Before we can land this feature:

* [ ] Restrict all the other gestures on iOS: https://github.com/mapbox/mapbox-gl-native/pull/5584#issuecomment-230885270
* [ ] Add a warning about performance: https://github.com/mapbox/mapbox-gl-native/pull/5584#r69783094
  * Or maybe find a way to call the method less frequently?
  * Or calculate the final camera beforehand and avoid the set-and-revert dance?

Fixes #2457.

/cc @boundsj @frederoni @friedbunny

### DIFF
--- a/platform/ios/CHANGELOG.md
+++ b/platform/ios/CHANGELOG.md
@@ -109,10 +109,6 @@ Mapbox welcomes participation and contributions from everyone. Please read [CONT
 * Fixed an issue that caused an assertion failure if a `MGLShapeCollection` (a GeoJSON GeometryCollection) was created with an empty array of shapes. ([#7632](https://github.com/mapbox/mapbox-gl-native/pull/7632))
 * Improved the precision of annotations at zoom levels greater than 18. ([#5517](https://github.com/mapbox/mapbox-gl-native/pull/5517))
 
-### User interactivity
-
-
-
 ### Networking and offline maps
 
 * Fixed an issue preventing an MGLMapView from loading tiles while an offline pack is downloading. ([#6446](https://github.com/mapbox/mapbox-gl-native/pull/6446))
--- a/platform/ios/CHANGELOG.md
+++ b/platform/ios/CHANGELOG.md
@@ -108,6 +108,12 @@ Mapbox welcomes participation and contributions from everyone. Please read [CONT
 * Fixed an issue that caused an assertion failure if a `MGLShapeCollection` (a GeoJSON GeometryCollection) was created with an empty array of shapes. ([#7632](https://github.com/mapbox/mapbox-gl-native/pull/7632))
 * Improved the precision of annotations at zoom levels greater than 18. ([#5517](https://github.com/mapbox/mapbox-gl-native/pull/5517))
 
+### User interactivity
+
+* Added a method to MGLMapViewDelegate, `-mapView:shouldChangeFromCamera:toCamera:`, that you can implement to restrict which parts the user can navigate to using gestures. ([#5584](https://github.com/mapbox/mapbox-gl-native/pull/5584))
+* The UITapGestureRecognizer on MGLMapView that is used for selecting annotations now fails if a tap does not select an annotation. ([#7246](https://github.com/mapbox/mapbox-gl-native/pull/7246))
+* As the user zooms in, tiles from lower zoom levels are scaled up until tiles for higher zoom levels are loaded. ([#5143](https://github.com/mapbox/mapbox-gl-native/pull/5143))
+
 ### Networking and offline maps
 
 * Fixed an issue preventing an MGLMapView from loading tiles while an offline pack is downloading. ([#6446](https://github.com/mapbox/mapbox-gl-native/pull/6446))
@@ -121,14 +127,9 @@ Mapbox welcomes participation and contributions from everyone. Please read [CONT
 
 ### Other changes
 
-* Raster tiles such as those from Mapbox Satellite are now cached, eliminating flashing while panning back and forth. ([#7091](https://github.com/mapbox/mapbox-gl-native/pull/7091))
-* Improved the performance of symbol style layers. ([#7025](https://github.com/mapbox/mapbox-gl-native/pull/7025))
-* As the user zooms in, tiles from lower zoom levels are scaled up until tiles for higher zoom levels are loaded. ([#5143](https://github.com/mapbox/mapbox-gl-native/pull/5143))
 * Notification names and user info keys are now string enumeration values for ease of use in Swift. ([#6794](https://github.com/mapbox/mapbox-gl-native/pull/6794))
 * MGLMapDebugOverdrawVisualizationMask no longer has any effect in Release builds of the SDK. This debug mask has been disabled for performance reasons. ([#5555](https://github.com/mapbox/mapbox-gl-native/pull/5555))
 * Fixed a typo in the documentation for the MGLCompassDirectionFormatter class. ([#5879](https://github.com/mapbox/mapbox-gl-native/pull/5879))
-* The UITapGestureRecognizer on MGLMapView that is used for selecting annotations now fails if a tap does not select an annotation. ([#7246](https://github.com/mapbox/mapbox-gl-native/pull/7246))
-* Fixed issues related to the visibility of sources in viewports less than 512 pixels wide or tall. ([#7438](https://github.com/mapbox/mapbox-gl-native/pull/7438))
 
 ## 3.3.7 - November 17, 2016
 
--- a/platform/ios/CHANGELOG.md
+++ b/platform/ios/CHANGELOG.md
@@ -37,6 +37,7 @@ Mapbox welcomes participation and contributions from everyone. Please read [CONT
 * Fixed an issue where translucent, non-view-backed point annotations along tile boundaries would be drawn darker than expected. ([#6832](https://github.com/mapbox/mapbox-gl-native/pull/6832))
 * Fixed flickering that occurred when panning past the antimeridian. ([#7574](https://github.com/mapbox/mapbox-gl-native/pull/7574))
 * Added a `MGLDistanceFormatter` class for formatting geographic distances. ([#7888](https://github.com/mapbox/mapbox-gl-native/pull/7888))
+* Added a method to MGLMapViewDelegate, `-mapView:shouldChangeFromCamera:toCamera:`, that you can implement to restrict which parts the user can navigate to using gestures. ([#5584](https://github.com/mapbox/mapbox-gl-native/pull/5584))
 
 ## 3.4.1 - January 25, 2017
 
@@ -110,9 +111,7 @@ Mapbox welcomes participation and contributions from everyone. Please read [CONT
 
 ### User interactivity
 
-* Added a method to MGLMapViewDelegate, `-mapView:shouldChangeFromCamera:toCamera:`, that you can implement to restrict which parts the user can navigate to using gestures. ([#5584](https://github.com/mapbox/mapbox-gl-native/pull/5584))
-* The UITapGestureRecognizer on MGLMapView that is used for selecting annotations now fails if a tap does not select an annotation. ([#7246](https://github.com/mapbox/mapbox-gl-native/pull/7246))
-* As the user zooms in, tiles from lower zoom levels are scaled up until tiles for higher zoom levels are loaded. ([#5143](https://github.com/mapbox/mapbox-gl-native/pull/5143))
+
 
 ### Networking and offline maps
 
--- a/platform/ios/CHANGELOG.md
+++ b/platform/ios/CHANGELOG.md
@@ -122,9 +122,14 @@ Mapbox welcomes participation and contributions from everyone. Please read [CONT
 
 ### Other changes
 
+* Raster tiles such as those from Mapbox Satellite are now cached, eliminating flashing while panning back and forth. ([#7091](https://github.com/mapbox/mapbox-gl-native/pull/7091))
+* Improved the performance of symbol style layers. ([#7025](https://github.com/mapbox/mapbox-gl-native/pull/7025))
+* As the user zooms in, tiles from lower zoom levels are scaled up until tiles for higher zoom levels are loaded. ([#5143](https://github.com/mapbox/mapbox-gl-native/pull/5143))
 * Notification names and user info keys are now string enumeration values for ease of use in Swift. ([#6794](https://github.com/mapbox/mapbox-gl-native/pull/6794))
 * MGLMapDebugOverdrawVisualizationMask no longer has any effect in Release builds of the SDK. This debug mask has been disabled for performance reasons. ([#5555](https://github.com/mapbox/mapbox-gl-native/pull/5555))
 * Fixed a typo in the documentation for the MGLCompassDirectionFormatter class. ([#5879](https://github.com/mapbox/mapbox-gl-native/pull/5879))
+* The UITapGestureRecognizer on MGLMapView that is used for selecting annotations now fails if a tap does not select an annotation. ([#7246](https://github.com/mapbox/mapbox-gl-native/pull/7246))
+* Fixed issues related to the visibility of sources in viewports less than 512 pixels wide or tall. ([#7438](https://github.com/mapbox/mapbox-gl-native/pull/7438))
 
 ## 3.3.7 - November 17, 2016
 
--- a/platform/ios/src/MGLMapView.mm
+++ b/platform/ios/src/MGLMapView.mm
@@ -1636,6 +1636,7 @@ public:
 
     _mbglMap->cancelTransitions();
     MGLMapCamera *oldCamera = self.camera;
+    BOOL didCameraChanged = NO;
     if (twoFingerDrag.state == UIGestureRecognizerStateBegan)
     {
         [self trackGestureEvent:MGLEventGesturePitchStart forRecognizer:twoFingerDrag];
@@ -1652,7 +1653,7 @@ public:
         CGPoint centerPoint = [self anchorPointForGesture:twoFingerDrag];
 
         _mbglMap->setPitch(pitchNew, mbgl::ScreenCoordinate { centerPoint.x, centerPoint.y });
-
+        didCameraChanged = YES;
         [self notifyMapChange:mbgl::MapChangeRegionIsChanging];
     }
     else if (twoFingerDrag.state == UIGestureRecognizerStateEnded || twoFingerDrag.state == UIGestureRecognizerStateCancelled)
@@ -1661,7 +1662,8 @@ public:
         [self unrotateIfNeededForGesture];
     }
     
-    if ([self.delegate respondsToSelector:@selector(mapView:shouldChangeFromCamera:toCamera:)]
+    if (didCameraChanged
+        && [self.delegate respondsToSelector:@selector(mapView:shouldChangeFromCamera:toCamera:)]
         && ![self.delegate mapView:self shouldChangeFromCamera:oldCamera toCamera:self.camera])
     {
         self.camera = oldCamera;
--- a/platform/ios/src/MGLMapView.mm
+++ b/platform/ios/src/MGLMapView.mm
@@ -1271,7 +1271,7 @@ public:
     _mbglMap->cancelTransitions();
 
     CGPoint centerPoint = [self anchorPointForGesture:pinch];
-
+    MGLMapCamera *oldCamera = self.camera;
     if (pinch.state == UIGestureRecognizerStateBegan)
     {
         [self trackGestureEvent:MGLEventGesturePinchStart forRecognizer:pinch];
@@ -1341,6 +1341,12 @@ public:
 
         [self unrotateIfNeededForGesture];
     }
+    
+    if ([self.delegate respondsToSelector:@selector(mapView:shouldChangeFromCamera:toCamera:)]
+        && ![self.delegate mapView:self shouldChangeFromCamera:oldCamera toCamera:self.camera])
+    {
+        self.camera = oldCamera;
+    }
 
     _previousPinchCenterCoordinate = [self convertPoint:centerPoint toCoordinateFromView:self];
     _previousPinchNumberOfTouches = pinch.numberOfTouches;
@@ -1353,7 +1359,7 @@ public:
     _mbglMap->cancelTransitions();
 
     CGPoint centerPoint = [self anchorPointForGesture:rotate];
-
+    MGLMapCamera *oldCamera = self.camera;
     if (rotate.state == UIGestureRecognizerStateBegan)
     {
         [self trackGestureEvent:MGLEventGestureRotateStart forRecognizer:rotate];
@@ -1378,10 +1384,16 @@ public:
             newDegrees = fminf(newDegrees,  30);
             newDegrees = fmaxf(newDegrees, -30);
         }
-
-        _mbglMap->setBearing(newDegrees, mbgl::ScreenCoordinate { centerPoint.x, centerPoint.y });
-
-        [self notifyMapChange:mbgl::MapChangeRegionIsChanging];
+        
+        if ([self.delegate respondsToSelector:@selector(mapView:shouldChangeFromCamera:toCamera:)]
+            && ![self.delegate mapView:self shouldChangeFromCamera:oldCamera toCamera:self.camera])
+        {
+            self.camera = oldCamera;
+        } else {
+            _mbglMap->setBearing(newDegrees, mbgl::ScreenCoordinate { centerPoint.x, centerPoint.y });
+            [self notifyMapChange:mbgl::MapChangeRegionIsChanging];
+        }
+        
     }
     else if (rotate.state == UIGestureRecognizerStateEnded || rotate.state == UIGestureRecognizerStateCancelled)
     {
@@ -1410,6 +1422,8 @@ public:
             [self unrotateIfNeededForGesture];
         }
     }
+    
+    
 }
 
 - (void)handleSingleTapGesture:(UITapGestureRecognizer *)singleTap
--- a/platform/ios/src/MGLMapView.mm
+++ b/platform/ios/src/MGLMapView.mm
@@ -1206,7 +1206,6 @@ public:
     _mbglMap->cancelTransitions();
 
     MGLMapCamera *oldCamera = self.camera;
-    BOOL didChangeCamera = NO;
     if (pan.state == UIGestureRecognizerStateBegan)
     {
         [self trackGestureEvent:MGLEventGesturePanStart forRecognizer:pan];
@@ -1220,7 +1219,6 @@ public:
         CGPoint delta = [pan translationInView:pan.view];
         _mbglMap->moveBy({ delta.x, delta.y });
         [pan setTranslation:CGPointZero inView:pan.view];
-        didChangeCamera = YES;
 
         [self notifyMapChange:mbgl::MapChangeRegionIsChanging];
     }
@@ -1254,8 +1252,7 @@ public:
         }];
     }
     
-    if (didChangeCamera
-        && [self.delegate respondsToSelector:@selector(mapView:shouldChangeFromCamera:toCamera:)]
+    if ([self.delegate respondsToSelector:@selector(mapView:shouldChangeFromCamera:toCamera:)]
         && ![self.delegate mapView:self shouldChangeFromCamera:oldCamera toCamera:self.camera])
     {
         self.camera = oldCamera;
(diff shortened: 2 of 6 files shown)


1ec5:
I think it’s too late for this feature to make the impending v3.3.0 final release, but we could easily backport it to v3.3.1 if it takes awhile to release v3.4.0.

1ec5:
Note that this API does not affect programmatic movement of the viewport, such as via `-flyToCamera:…`, so we won’t run into mapbox/mapbox-gl-js#2521.

platform/macos/src/MGLMapViewDelegate.h
```diff
@@ -61,6 +61,25 @@ NS_ASSUME_NONNULL_BEGIN
...
 - (void)mapView:(MGLMapView *)mapView cameraDidChangeAnimated:(BOOL)animated;
 
+/**
+ Asks the delegate whether the map view should be allowed to change from the
+ existing camera to the new camera in response to a user gesture.
+ 
+ This method is called as soon as the user gesture is recognized. It is not
+ called in response to a programmatic camera change, such as by setting the
+ `centerCoordinate` property or calling `-flyToCamera:completionHandler:`.
+
```
friedbunny: Should we add the standard, “this method is called many times during gesticulation so you should keep it as lightweight as possible”?

friedbunny:
This only seems to prevent simple pans, so far. With the new delegate method set to always return `NO`, it’s still possible to move the map via flick, pinch + pan, and rotate + pan.

1ec5:
> With the new delegate method set to always return `NO`, it’s still possible to move the map via flick, pinch + pan, and rotate + pan. Quick zoom also works.

Oh, you’re right! I took care of all the gesture recognizers for macOS but forgot to follow up for iOS.

1ec5:
Another caveat about this approach is that, if the delegate decides to block the camera change, the camera actually does change (even sending the `-mapView:didChangeRegionAnimated:` message to the delegate) before the map view resets the camera.

Zakay:
👍

maciekish:
Any chance to see this in v3.3.1 please?

1ec5:
v3.3.1 has already been released. This work is currently targeted for v3.4.0. However, lately I’ve been focused on the runtime styling API ahead of the first beta, since it has the potential for a lot more API churn.

1ec5:
> This feels a bit more useful if we can determine more about the viewport than just the newCamera center coordinate

Yep, there’s definitely a lot more potential here. Future smarts probably belong in mbgl, whereas I tried to keep things light in this PR. For example, if the use case is to prevent a certain tile from being shown (with rotation and pitch gestures enabled), there needs to be some nearest-point-in-polygon logic in mbgl, but for now, developers can bring their own implementation of that logic.

maciekish:
This does not work when flicking the map. Delegate is never called.

1ec5:
> This does not work when flicking the map. Delegate is never called.

Yes, this is on my to-do list, per https://github.com/mapbox/mapbox-gl-native/pull/5584#issuecomment-230885270. All the gesture recognizers are hooked up on macOS, in case you want to try it out in the meantime.

Sealos:
Any chance for this feature in 3.4.0?

boundsj:
Moved this to 3.4.1

boundsj:
Moving to 3.5.0 cc @1ec5

boundsj:
@incanus can you please have another look at this and release your hold on it if it looks ok?

platform/ios/CHANGELOG.md
```diff
@@ -99,6 +99,12 @@ Mapbox welcomes participation and contributions from everyone. Please read [CONT
 * Fixed an issue that caused an assertion failure if a `MGLShapeCollection` (a GeoJSON GeometryCollection) was created with an empty array of shapes. ([#7632](https://github.com/mapbox/mapbox-gl-native/pull/7632))
 * Improved the precision of annotations at zoom levels greater than 18. ([#5517](https://github.com/mapbox/mapbox-gl-native/pull/5517))
 
+### User interactivity
+
+* Added a method to MGLMapViewDelegate, `-mapView:shouldChangeFromCamera:toCamera:`, that you can implement to restrict which parts the user can navigate to using gestures. ([#5584](https://github.com/mapbox/mapbox-gl-native/pull/5584))
```
boundsj: This changelog entry should be moved to reflect the actual release it is going out in. Alternatively, this PR could be retargeted to the `master` branch and this entry could go in the "master" section. cc @1ec5

platform/macos/src/MGLMapView.mm
```diff
@@ -1372,15 +1382,25 @@ - (void)handlePanGesture:(NSPanGestureRecognizer *)gestureRecognizer {
...
                 _mbglMap->setBearing(newDirection, center);
+                didChangeCamera = YES;
                 [self didChangeValueForKey:@"direction"];
             }
             if (self.pitchEnabled) {
                 _mbglMap->setPitch(_pitchAtBeginningOfGesture + delta.y / 5, center);
+                didChangeCamera = YES;
+            }
+            
+            if (didChangeCamera
```
boundsj: @fabian-guerra @1ec5 is there a reason we don't move this up and short circuit the bearing and pitch changes to the map completely?

platform/ios/CHANGELOG.md
1ec5: This PR is already targeted at master, so this blurb needs to be moved up to “Other changes” under the existing master section.

platform/macos/src/MGLMapView.mm
fabian-guerra: I realized that we change `didChangeCamera` only when `pan.state == UIGestureRecognizerStateChanged` this introduced a bug that caused the camera to move when `pan.state == UIGestureRecognizerStateEnded` I think we can remove this variable and do the short circuit instead. What are your thoughts @1ec5?

1ec5: I alluded to this up above:

> Or calculate the final camera beforehand and avoid the set-and-revert dance?

The reason I originally implemented it this way was that I expected to eventually implement a rubber-banding effect, in which we would have to set the camera (with animation) then revert. But we can backfill that behavior later without affecting the API, so short-circuiting should be fine.

platform/ios/src/MGLMapView.mm
```diff
@@ -1337,6 +1337,12 @@ - (void)handlePinchGesture:(UIPinchGestureRecognizer *)pinch
 
         [self unrotateIfNeededForGesture];
     }
+    
+    if ([self.delegate respondsToSelector:@selector(mapView:shouldChangeFromCamera:toCamera:)]
+        && ![self.delegate mapView:self shouldChangeFromCamera:oldCamera toCamera:self.camera])
+    {
+        self.camera = oldCamera;
+    }
```
incanus: This seems to happen too late in the process? i.e. `setScale()` has already been called above, versus the way the way other gestures are handled in this PR.

platform/macos/src/MGLMapView.mm
1ec5: The other reason I used a set-and-revert strategy was simplicity: we don’t know `newCamera` until we perform the transformation, and `mbgl::Map` lacks a “dry run” version of these transformation methods.

platform/ios/src/MGLMapView.mm
1ec5: This is consistent with all the gesture handlers I implemented on macOS, as well as `-handlePanGesture:` above, but you’re right that it’s inconsistent with `-handleRotateGesture:` below.

platform/ios/src/MGLMapView.mm
```diff
@@ -1364,10 +1380,16 @@ - (void)handleRotateGesture:(UIRotationGestureRecognizer *)rotate
             newDegrees = fminf(newDegrees,  30);
             newDegrees = fmaxf(newDegrees, -30);
         }
-
-        _mbglMap->setBearing(newDegrees, mbgl::ScreenCoordinate { centerPoint.x, centerPoint.y });
-
-        [self notifyMapChange:mbgl::MapChangeRegionIsChanging];
+        
+        if ([self.delegate respondsToSelector:@selector(mapView:shouldChangeFromCamera:toCamera:)]
+            && ![self.delegate mapView:self shouldChangeFromCamera:oldCamera toCamera:self.camera])
```
1ec5: `newCamera` equals `oldCamera` because we haven’t called `mbgl::Map::setBearing()` yet. `newCamera` should instead reflect the camera _after_ applying the rotation.

platform/macos/CHANGELOG.md
```diff
@@ -67,6 +67,7 @@ This version of the Mapbox macOS SDK corresponds to version 3.4.0 of the Mapbox
 * Improved style parsing performance. ([#6170](https://github.com/mapbox/mapbox-gl-native/pull/6170))
 * Improved feature querying performance. ([#6514](https://github.com/mapbox/mapbox-gl-native/pull/6514))
 * Fixed an issue where shapes that cannot currently be visually represented as annotations were still shown on the map as point annotations. ([#6764](https://github.com/mapbox/mapbox-gl-native/issues/6764))
+* Raster tiles such as those from Mapbox Satellite are now cached, eliminating flashing while panning back and forth. ([#7091](https://github.com/mapbox/mapbox-gl-native/pull/7091))
```
1ec5: Did you mean to add this entry?

platform/ios/src/MGLMapView.mm
```diff
@@ -1615,6 +1656,12 @@ - (void)handleTwoFingerDragGesture:(UIPanGestureRecognizer *)twoFingerDrag
         [self notifyGestureDidEndWithDrift:NO];
         [self unrotateIfNeededForGesture];
     }
+    
+    if ([self.delegate respondsToSelector:@selector(mapView:shouldChangeFromCamera:toCamera:)]
```
1ec5: Unless you check the gesture recognizer's state (or didChangeCamera), the camera will be reset to oldCamera before the gesture is complete.

platform/macos/CHANGELOG.md
fabian-guerra: eh.. no

platform/ios/src/MGLMapView.mm
```diff
@@ -1533,17 +1560,25 @@ - (void)handleTwoFingerTapGesture:(UITapGestureRecognizer *)twoFingerTap
...
-
-        mbgl::ScreenCoordinate center(gesturePoint.x, gesturePoint.y);
-        _mbglMap->scaleBy(0.5, center, MGLDurationInSecondsFromTimeInterval(MGLAnimationDuration));
-
-        __weak MGLMapView *weakSelf = self;
-
-        [self animateWithDelay:MGLAnimationDuration animations:^
+        MGLMapCamera *oldCamera = self.camera;
+        if ([self.delegate respondsToSelector:@selector(mapView:shouldChangeFromCamera:toCamera:)]
+            && ![self.delegate mapView:self shouldChangeFromCamera:oldCamera toCamera:self.camera])
```
1ec5: self.camera at this point is equal to oldCamera. We need to pass in the destination camera as the third parameter. Could you create a copy of the camera and modify it to reflect the result after zooming out?

platform/ios/CHANGELOG.md
```diff
@@ -99,6 +99,12 @@ Mapbox welcomes participation and contributions from everyone. Please read [CONT
...
+* The UITapGestureRecognizer on MGLMapView that is used for selecting annotations now fails if a tap does not select an annotation. ([#7246](https://github.com/mapbox/mapbox-gl-native/pull/7246))
```
1ec5: Can you undo all these unrelated changes?

platform/macos/CHANGELOG.md
```diff
@@ -81,6 +82,10 @@ This version of the Mapbox macOS SDK corresponds to version 3.4.0 of the Mapbox
 * Fixed an issue that caused an assertion failure if a `MGLShapeCollection` (a GeoJSON GeometryCollection) was created with an empty array of shapes. ([#7632](https://github.com/mapbox/mapbox-gl-native/pull/7632))
 * Improved the precision of annotations at zoom levels greater than 18. ([#5517](https://github.com/mapbox/mapbox-gl-native/pull/5517))
 
+### User interactivity
+
+* The NSClickGestureRecognizer on MGLMapView that is used for selecting annotations now fails if a click does not select an annotation. ([#7246](https://github.com/mapbox/mapbox-gl-native/pull/7246))
```
1ec5: There are still a number of unrelated changes in both changelogs.

platform/ios/src/MGLMapView.mm
```diff
@@ -1364,10 +1377,10 @@ - (void)handleRotateGesture:(UIRotationGestureRecognizer *)rotate
             newDegrees = fminf(newDegrees,  30);
             newDegrees = fmaxf(newDegrees, -30);
         }
-
+
```
incanus: Nitpick: get rid of extraneous whitespace here.

platform/ios/src/MGLMapView.mm
```diff
@@ -1607,14 +1669,21 @@ - (void)handleTwoFingerDragGesture:(UIPanGestureRecognizer *)twoFingerDrag
         CGPoint centerPoint = [self anchorPointForGesture:twoFingerDrag];
 
         _mbglMap->setPitch(pitchNew, mbgl::ScreenCoordinate { centerPoint.x, centerPoint.y });
-
+        didCameraChanged = YES;
```
incanus: Nitpick: should probably call this `didChangeCamera`.

1ec5:
> Do we care that various metrics are still reported even when the old camera is reset, as the viewport would not end up changing but the change would be recorded?

Given the gesture lifecycle, this is an inevitable consequence of our set-and-revert strategy. I think it's OK, though, because the events are meant to track user intent, not necessarily tile views.

1ec5:
Indeed, the delegate method is still unable to stop drift-panning/zooming/rotating on iOS. Consider this scenario:

1. The delegate method is implemented so that the user may not pan beyond the antimeridian.
1. The user quickly pans westward toward the antimeridian, letting go before reaching it, but there's enough momentum that the map continues drift-panning.
1. Drift-pannning is implemented by calling mbgl::Map::panBy() with a duration. The animation runs asynchronously, so the delegate method gets called at the moment the animation begins, with a newCamera equal to oldCamera.
1. The delegate method returns YES because newCamera (which incorrectly matches oldCamera) still has a longitude greater than −180°.

The only gesture recognizer that currently handles this scenario correctly is the double-tap gesture recognizer, because it predicts the final camera instead of performing the set-and-revert strategy. In order to fix this issue comprehensively, we'd have to follow a similar approach with all the gesture recognizers. The code will be a little messier, but not by much.

The macOS implementation is unaffected by this issue, because the drift-panning animation is driven by AppKit instead of mbgl. The gesture recognizer method gets called incrementally after the user lets go.

platform/ios/src/MGLMapView.mm
```diff
@@ -1508,18 +1573,36 @@ - (void)handleDoubleTapGesture:(UITapGestureRecognizer *)doubleTap
...
-
-        mbgl::ScreenCoordinate center(gesturePoint.x, gesturePoint.y);
-        _mbglMap->scaleBy(2, center, MGLDurationInSecondsFromTimeInterval(MGLAnimationDuration));
-
-        __weak MGLMapView *weakSelf = self;
-
-        [self animateWithDelay:MGLAnimationDuration animations:^
+        MGLMapCamera *oldCamera = self.camera;
+        
+        mbgl::EdgeInsets padding = MGLEdgeInsetsFromNSEdgeInsets(self.contentInset);
```
boundsj: nitpick: these several lines about getting padding, camera options, setting a zoom adjustment, clamping, and getting a `toCamera` from camera options are repeated in the `handleDoubleTapGesture:`, `handleTwoFingerTapGesture:` and `handleQuickZoomGesture:` implementations. This is similar to what we had before where the previous (less complex because no estimate) logic was repeated, but this PR could be a chance to refactor this into a private utility method. I think anything we can do to make the gesture handling more readable and easier to refactor in the future would be 👍 so that is why I mention this.

platform/ios/src/MGLMapView.mm
```diff
@@ -1321,14 +1353,26 @@ - (void)handlePinchGesture:(UIPinchGestureRecognizer *)pinch
...
+        zoom = log2(newScale);
+        toCamera = [self currentCameraWithEstimatedZoom:zoom];
+        
+        if ([self.delegate respondsToSelector:@selector(mapView:shouldChangeFromCamera:toCamera:)]
+            && ![self.delegate mapView:self shouldChangeFromCamera:oldCamera toCamera:toCamera])
         {
-            _mbglMap->setScale(newScale, mbgl::ScreenCoordinate { centerPoint.x, centerPoint.y }, MGLDurationInSecondsFromTimeInterval(duration));
+            self.camera = oldCamera;
+            drift = NO;
+        }else {
```
1ec5: Nit: missing space before `else`.

platform/ios/src/MGLMapView.mm
```diff
@@ -1619,6 +1748,73 @@ - (void)handleTwoFingerDragGesture:(UIPanGestureRecognizer *)twoFingerDrag
...
+        width/2.,
+        height/2.,
+    };
+    
+    mbgl::ScreenCoordinate centerOffset = {
+        endPoint.x,
+        -endPoint.y,
+    };
+    
+    mbgl::ScreenCoordinate centerPoint = screenCenter - centerOffset;
```
1ec5: You can simplify this logic by creating a `CGPoint` directly instead of a `ScreenCoordinate`:

```objc
CGPoint centerPoint = CGPointMake(CGRectGetMidX(self.bounds), CGRectGetMidY(self.bounds));
```

platform/ios/src/MGLMapView.mm
```diff
@@ -1619,6 +1748,73 @@ - (void)handleTwoFingerDragGesture:(UIPanGestureRecognizer *)twoFingerDrag
         [self notifyGestureDidEndWithDrift:NO];
         [self unrotateIfNeededForGesture];
     }
+
+}
+
+- (MGLMapCamera *)currentCameraWithEstimatedEndPoint:(CGPoint)endPoint panGesture:(UIPanGestureRecognizer *)pan
```
1ec5: `pan` is unused.

A better place for this logic would be a private `-[MGLMapCamera cameraByPanningWithTranslation:inMapView:]` (which would have parameters similar to `-[UIPanGestureRecognizer setTranslation:inView:]`). You’d need to create a new MGLMapCamera_Private.h for these methods.

This way you can also share this and similar methods between iOS and macOS. The downside is that our non-UI code would have some dependency on our UI code, but that won’t introduce any practical problems unless we decide to split the Darwin code into a separate framework down the line.

platform/ios/src/MGLMapView.mm
```diff
@@ -1619,6 +1748,73 @@ - (void)handleTwoFingerDragGesture:(UIPanGestureRecognizer *)twoFingerDrag
...
+
+- (MGLMapCamera *)currentCameraWithEstimatedPitch:(CGFloat)pitch anchorPoint:(CGPoint)anchorPoint
+{
+    mbgl::EdgeInsets padding = MGLEdgeInsetsFromNSEdgeInsets(self.contentInset);
+    mbgl::CameraOptions currentCameraOptions = _mbglMap->getCameraOptions(padding);
+    
+    MGLMapCamera *camera;
+    
+    mbgl::ScreenCoordinate anchor = mbgl::ScreenCoordinate { anchorPoint.x, anchorPoint.y };
+    currentCameraOptions.pitch = pitch * mbgl::util::DEG2RAD;
```
1ec5: Pitch has no anchor point – it’s always the center of the view, disregarding any padding – so this method only needs to copy the camera and set its pitch to the new value.

platform/ios/src/MGLMapView.mm
```diff
@@ -1619,6 +1748,73 @@ - (void)handleTwoFingerDragGesture:(UIPanGestureRecognizer *)twoFingerDrag
...
+    
+    mbgl::ScreenCoordinate centerPoint = screenCenter - centerOffset;
+    CGPoint endCameraPoint = CGPointMake(centerPoint.x, centerPoint.y);
+    CLLocationCoordinate2D panCoordinate = [self convertPoint:endCameraPoint toCoordinateFromView:pan.view];
+    panCamera.centerCoordinate = panCoordinate;
+    
+    return panCamera;
+}
+
+- (MGLMapCamera *)currentCameraWithEstimatedZoom:(double)zoom
```
1ec5: A pinch gesture performs a zoom anchored at the center of the pinch gesture. So this method needs something similar to the rotation method.

platform/ios/src/MGLMapView.mm
1ec5: If moving these methods to MGLMapCamera would be too much hassle because of dependencies on internal MGLMapView methods, we can leave the methods here, but let’s give them better names:

* `-currentCameraWithEstimatedEndPoint:panGesture:` → `-cameraByPanningWithTranslation:`
* `-currentCameraWithEstimatedZoom:` → `-cameraByZoomingToZoomLevel:`
* `-currentCameraWithEstimatedDegrees:anchorPoint:` → `-cameraByRotatingToDirection:aroundAnchorPoint:`
* `-currentCameraWithEstimatedPitch:anchorPoint:` → `-cameraByTiltingToPitch:`

platform/ios/src/MGLMapView.mm
```diff
@@ -1619,6 +1736,86 @@ - (void)handleTwoFingerDragGesture:(UIPanGestureRecognizer *)twoFingerDrag
...
+    
+    mbgl::ScreenCoordinate anchor = mbgl::ScreenCoordinate { anchorPoint.x, anchorPoint.y };
+    currentCameraOptions.pitch = pitch * mbgl::util::DEG2RAD;
+    currentCameraOptions.anchor = anchor;
+    camera = [self cameraForCameraOptions:currentCameraOptions];
+    
+    return camera;
+}
+
+- (double)currentCameraZoom
```
1ec5: This method is unnecessary; use `self.zoomLevel` instead.

platform/ios/src/MGLMapView.mm
```diff
@@ -1619,6 +1736,86 @@ - (void)handleTwoFingerDragGesture:(UIPanGestureRecognizer *)twoFingerDrag
...
+    mbgl::CameraOptions currentCameraOptions = _mbglMap->getCameraOptions(padding);
+    MGLMapCamera *camera;
+    
+    currentCameraOptions.zoom = mbgl::util::clamp(zoom, self.minimumZoomLevel, self.maximumZoomLevel);
+    camera = [self cameraForCameraOptions:currentCameraOptions];
+    
+    return camera;
+}
+
+- (MGLMapCamera *)currentCameraWithEstimatedDegrees:(CGFloat)degrees anchorPoint:(CGPoint)anchorPoint
```
1ec5: Use `CLLocationDirection` to represent the map’s rotation in degrees.

platform/ios/src/MGLMapView.mm
```diff
@@ -1216,8 +1219,17 @@ - (void)handlePanGesture:(UIPanGestureRecognizer *)pan
...
         CGPoint delta = [pan translationInView:pan.view];
-        _mbglMap->moveBy({ delta.x, delta.y });
-        [pan setTranslation:CGPointZero inView:pan.view];
+
+        toCamera = [self currentCameraWithEstimatedEndPoint:delta panGesture:pan];
+        
+        if ([self.delegate respondsToSelector:@selector(mapView:shouldChangeFromCamera:toCamera:)]
+            && ![self.delegate mapView:self shouldChangeFromCamera:oldCamera toCamera:toCamera])
+        {
+            self.camera = oldCamera;
```
fabian-guerra: Correct, this only emphases we won't perform any gesture.

1ec5: OK, let’s remove this line and similar lines, then. Setting the camera also cancels any pending animations, which could be unexpected.

platform/ios/src/MGLMapView.mm
```diff
@@ -1275,9 +1297,19 @@ - (void)handlePinchGesture:(UIPinchGestureRecognizer *)pinch
...
 
         if (log2(newScale) < _mbglMap->getMinZoom()) return;
-
-        _mbglMap->setScale(newScale, mbgl::ScreenCoordinate { centerPoint.x, centerPoint.y });
-
+        
+        // Calculates the final camera zoom, has no effect within current map camera.
+        MGLMapCamera *toCamera;
+        double zoom = log2(newScale);
+        toCamera = [self cameraByZoomingToZoomLevel:zoom aroundAnchorPoint:centerPoint];
```
1ec5: Declare and set `toCamera` on the same line.

platform/ios/src/MGLMapView.mm
```diff
@@ -1216,8 +1219,17 @@ - (void)handlePanGesture:(UIPanGestureRecognizer *)pan
     else if (pan.state == UIGestureRecognizerStateChanged)
     {
         CGPoint delta = [pan translationInView:pan.view];
-        _mbglMap->moveBy({ delta.x, delta.y });
-        [pan setTranslation:CGPointZero inView:pan.view];
+
+        toCamera = [self cameraByPanningWithTranslation:delta panGesture:pan];
```
1ec5: Move the declaration of `toCamera` down to this line.

platform/ios/src/MGLMapView.mm
```diff
@@ -1275,9 +1297,19 @@ - (void)handlePinchGesture:(UIPinchGestureRecognizer *)pinch
         CGFloat newScale = self.scale * pinch.scale;
 
         if (log2(newScale) < _mbglMap->getMinZoom()) return;
-
-        _mbglMap->setScale(newScale, mbgl::ScreenCoordinate { centerPoint.x, centerPoint.y });
-
+        
+        // Calculates the final camera zoom, has no effect within current map camera.
+        MGLMapCamera *toCamera;
+        double zoom = log2(newScale);
```
1ec5: Nit: declare this a few lines further up to avoid saying `log2(newScale)` twice.

platform/ios/src/MGLMapView.mm
```diff
@@ -1508,18 +1573,32 @@ - (void)handleDoubleTapGesture:(UITapGestureRecognizer *)doubleTap
...
         CGPoint gesturePoint = [self anchorPointForGesture:doubleTap];
-
-        mbgl::ScreenCoordinate center(gesturePoint.x, gesturePoint.y);
-        _mbglMap->scaleBy(2, center, MGLDurationInSecondsFromTimeInterval(MGLAnimationDuration));
-
-        __weak MGLMapView *weakSelf = self;
-
-        [self animateWithDelay:MGLAnimationDuration animations:^
+        
+        MGLMapCamera *toCamera = [self cameraByZoomingToZoomLevel:newZoom aroundAnchorPoint:gesturePoint];
```
1ec5: Nit: s/newZoom/self.zoomLevel + 1/

platform/ios/src/MGLMapView.mm
```diff
@@ -1216,8 +1218,16 @@ - (void)handlePanGesture:(UIPanGestureRecognizer *)pan
     else if (pan.state == UIGestureRecognizerStateChanged)
     {
         CGPoint delta = [pan translationInView:pan.view];
-        _mbglMap->moveBy({ delta.x, delta.y });
-        [pan setTranslation:CGPointZero inView:pan.view];
+
+        MGLMapCamera *toCamera = [self cameraByPanningWithTranslation:delta panGesture:pan];
+        
+        if (![self.delegate respondsToSelector:@selector(mapView:shouldChangeFromCamera:toCamera:)] ||
+            ([self.delegate respondsToSelector:@selector(mapView:shouldChangeFromCamera:toCamera:)]
```
1ec5: This is redundant: `!A || (A & C)` is equivalent to `!A || C`.

Issue #5526, opened 2016-06-30
Archive for iTunes Connect fails with `./Mapbox.framework/Mapbox' was built without full bitcode`


roblabs:
Attempting to build Mapbox iOS 3.3.0 for sending app out for TestFlight, then to customers on the App Store.

I am receiving the below error regarding 'not built with full bitcode'.  Our last App Store build was 3.2.2 and `make ipackage` worked without any flags.  Now we are seeing the issue, and the flag `BITCODE=YES` does not solve anything.

Any suggestions or workarounds?  We have bitcode on in our app, and we prefer to have bitcode on [due to its advantages from Apple](https://developer.apple.com/library/prerelease/content/documentation/IDEs/Conceptual/AppDistributionGuide/AppThinning/AppThinning.html#//apple_ref/doc/uid/TP40012582-CH35-SW2). 

Issue looks similar to #5448.  

*Platform: iOS*
*Mapbox SDK version: 3.3.0*
*SHA = 5389f04da5f66c5e26d71e761a0fcda4d8382581*
*Xcode Version 7.3.1 (7D1014)*

### Steps to trigger behavior

 1. `make clean`
 2. [Make with BITCODE=YES](https://github.com/mapbox/mapbox-gl-native/blob/bc24da11b27b6e180ab1716e57caa751eb4f5f00/Makefile#L154)
 3. `make ipackage BITCODE=YES`
 4.  copy `mapbox-gl-native/build/ios/pkg/dynamic` to my project
 5.  From Xcode for my target, `Product` > `Archive`

### Expected behavior
Happy Archive, ready for iTunes Connect and the App Store

### Actual behavior
Archive failed with this error

```
ld: bitcode bundle could not be generated because '/Users/me/bitbucket/project/Mapbox.framework/Mapbox' was built without full bitcode. 
All frameworks and dylibs for bitcode must be generated from Xcode Archive or Install build for architecture armv7
clang: error: linker command failed with exit code 1 (use -v to see invocation)
```

1ec5:
`make ipackage` produces a fat binary that contains slices for the simulator as well as the device. Simulator slices can’t be built with Bitcode support, because the x86_64 architecture doesn’t support it. In our [installation instructions](https://www.mapbox.com/ios-sdk/#manual), one of the steps is to add a Run Script build phase that runs strip-frameworks.sh. That script strips the simulator slices when archiving.

If you add that build phase, archive, and resubmit, do you still get the same error?

[...]

Issue #5448, opened 2016-06-23
No Bitcode in simulator removes bitcode in fat dynamic library


kenji21:
**Platform:** iOS
**Mapbox SDK version:** master@3459b83736f0e4eedfb95a1394d9e4ee1523b917

### Steps to trigger behavior

 1. `git clone in /tmp`
 2. `export BITCODE=YES`
 3. `export BUILDTYPE=Release`
 4. `make iframework`
 5. `for file in $(ls ./build/ios/Build/Products/Release-*/*.a ./build/ios/Build/Products/Release-*/Mapbox.framework/Mapbox build/ios/pkg/dynamic/Mapbox.framework/Mapbox  ) ; do echo $file ; otool -l $file |grep __LLVM |uniq ; done`

### Expected behavior

Last command should show 'segname __LLVM' for each dependencies (.a / framework), like for libmbgl-core.a:

```
./build/ios/Build/Products/Release-iphonesimulator/libmbgl-core.a
   segname __LLVM
```

Grep comes from an apple member indications (https://forums.developer.apple.com/thread/3971) :

>Searching for a “bitcode” section is not a reliable way to detect if your files contain embedded bitcode. If you want to do that, search for the “__LLVM” segment. You should be aware that a normal build with the -fembed-bitcode-marker option will produce minimal size embedded bitcode sections without any real content. This is done as a way of testing the bitcode-related aspects of your build without slowing down the build process. The actual bitcode content is included when you do an Archive build.

### Actual behavior

````
$ for file in $(ls ./build/ios/Build/Products/Release-*/*.a ./build/ios/Build/Products/Release-*/Mapbox.framework/Mapbox build/ios/pkg/dynamic/Mapbox.framework/Mapbox  ) ; do echo $file ; otool -l $file |grep __LLVM |uniq ; done
./build/ios/Build/Products/Release-iphoneos/Mapbox.framework/Mapbox
  segname __LLVM
   segname __LLVM
  segname __LLVM
   segname __LLVM
./build/ios/Build/Products/Release-iphoneos/libmbgl-core.a
   segname __LLVM
./build/ios/Build/Products/Release-iphoneos/libmbgl-platform-ios.a
   segname __LLVM
./build/ios/Build/Products/Release-iphonesimulator/Mapbox.framework/Mapbox
./build/ios/Build/Products/Release-iphonesimulator/libmbgl-core.a
[...]

PR #5586, opened 2016-07-06
Add support for full Bitcode builds

boundsj:
Fixes https://github.com/mapbox/mapbox-gl-native/issues/5526
Fixes https://github.com/mapbox/mapbox-gl-native/issues/5448

In https://github.com/mapbox/mapbox-gl-native/commit/f791eb32b9ebe96845e87e5a6acd3ee46182c1d5 we simplified the package script and intended to remove support for building the SDK without Bitcode support. The intention was to use the ENABLE_BITCODE project setting to do this and remove the control of that variable in the
script. However, since we use `xcodebuild build` this only uses bitcode `marker` by default and causes a failure when an app that uses the Mapbox SDK is archived.

This adds the `BITCODE_GENERATION_MODE` and `bitcode` value to the necessary iOS SDK and mbgl and platform targets. If an app that uses the Mapbox SDK is built with these flags set, it will be successfully processed in the bitcode phase in an app archive operation.

This also removes the `BITCODE` flag from the Makefile since the intention is to not support building the SDK without bitcode support and the presence of that option might be confusing.

cc @1ec5 @frederoni @friedbunny

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -125,23 +125,23 @@ test-ios: ios
 	  -workspace $(IOS_WORK_PATH) -scheme CI test $(XCPRETTY)
 
 ipackage: $(IOS_PROJ_PATH)
-	BITCODE=$(BITCODE) FORMAT=$(FORMAT) BUILD_DEVICE=$(BUILD_DEVICE) SYMBOLS=$(SYMBOLS) \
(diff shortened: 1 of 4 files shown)


1ec5:
:+1:

PR #5593, opened 2016-07-06
Use gesture recognizer for context menu

### LINKED ISSUE #5078: Context menu disables click and pan gesture recognizers
After bringing up the context menu, and subsequently dismissing it, panning and rotation with the mouse does not work anymore.

This only happens when using the right mouse button of a physical mouse,  or the two-finger tap on a trackpad, but NOT when using Ctrl+Tap or Ctrl+Left Click.

Zooming still works with the keyboard, the trackpad, or the mouse scrollwheel.

The map can be panned again after using pinch-to-zoom.

/cc @1ec5

1ec5:
NSView’s built-in context menu handling preempts non-right-click gesture recognizers, so preempt it with a right-click gesture recognizer that does essentially the same thing. With this change, two-finger double-tapping on a trackpad also zooms out, consistent with MapKit.

Fixes #5078.

/cc @kkaefer

### DIFF
--- a/platform/macos/CHANGELOG.md
+++ b/platform/macos/CHANGELOG.md
@@ -1,6 +1,8 @@
 # Changelog for Mapbox macOS SDK
 
 ## master
+
+* Right-clicking to open MGLMapView’s context menu no longer prevents the user from subsequently panning the map by clicking and dragging. ([#5593](https://github.com/mapbox/mapbox-gl-native/pull/5593))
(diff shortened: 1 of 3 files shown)


1ec5:
Tested on a trackpad and a Magic Mouse. Still need to test on a standard USB mouse.

kkaefer:
Awesome! Confirmed that this fixes the issue with a USB mouse as well.

PR #5585, opened 2016-07-06
Offline storage size API

1ec5:
Added a property to MGLOfflineStorage that indicates the disk space occupied by all cached and offline resources.

Fixes the iOS side of #5580.

/cc @friedbunny @jfirebaugh @zugaldia

PR #5601, opened 2016-07-07
[ios] Move cache.db to a subdirectory

friedbunny:
Per confirmation from @jfirebaugh, we cannot guarantee that the offline and ambient cache.db will be created in a timely/synchronous manner, so it's safer to set the backup exclusion key on its containing directory.

But, because we placed cache.db in the base of the app's Application Support directory (where backups are expected to be allowed), we need to move cache.db into a directory that we control and can exclude from backups with impunity.

This is an amalgam of https://github.com/mapbox/mapbox-gl-native/pull/5578 (directory exclusion) and https://github.com/mapbox/mapbox-gl-native/pull/5585 (refactored URL methods, but without the new database size method).

/cc @1ec5 @boundsj @dapetcu21


platform/darwin/src/MGLOfflineStorage.mm
```diff
@@ -40,58 +40,98 @@ + (instancetype)sharedOfflineStorage {
...
+ Returns the file URL to the offline cache.
+
+ The cache is located in a directory specific to the application, so that packs
+ downloaded by other applications don’t count toward this application’s limits.
+
+ The cache is located at:
+ ~/Library/Application Support/tld.app.bundle.id/mapbox/cache.db
+ */
++ (NSURL *)cacheURL {
+    return [[self class] cacheURLIncludingSubdirectory:YES];
```
1ec5: `self` is already the class.

1ec5: Might as well inline this method since it's so simple.

platform/darwin/src/MGLOfflineStorage.mm
```diff
@@ -40,58 +40,98 @@ + (instancetype)sharedOfflineStorage {
...
+                                                                        create:YES
+                                                                         error:nil];
+    NSString *bundleIdentifier = [NSBundle mainBundle].bundleIdentifier;
+    if (!bundleIdentifier) {
+        // There’s no main bundle identifier when running in a unit test bundle.
+        bundleIdentifier = [NSBundle bundleForClass:self].bundleIdentifier;
+    }
+    cacheDirectoryURL = [cacheDirectoryURL URLByAppendingPathComponent:bundleIdentifier];
+    if (useSubdirectory) {
+        cacheDirectoryURL = [cacheDirectoryURL URLByAppendingPathComponent:@"mapbox"];
```
1ec5: I'd go with something even less likely to conflict with developer code, such as `.mapbox`. For that matter, I wonder if dot-directories are even backed up to iCloud or Time Machine in the first place.

platform/darwin/src/MGLOfflineStorage.mm
```diff
@@ -40,58 +40,98 @@ + (instancetype)sharedOfflineStorage {
...
++ (NSURL *)cacheURL {
+    return [[self class] cacheURLIncludingSubdirectory:YES];
+}
+
+/**
+ Returns the file URL to the offline cache, with the option to omit the "mapbox"
+ subdirectory for legacy (v3.2.0 - v3.2.3) migration purposes.
+
+ The subdirectory-less cache was located at:
+ ~/Library/Application Support/tld.app.bundle.id/cache.db
```
1ec5: Also mention the new path. This comment is handy to link to in tickets.

1ec5:
:+1:

boundsj:
👍

Issue #3235, opened 2015-12-09
Implement gl::Value::reset as operator=(T::Default)


jfirebaugh:
```
    inline void reset() {
        operator=(T::Default);
    }
```

This will:

* avoid making a no-op change
* ensure `dirty` is `false`, which is correct (the GL state winds up set to a known value)

I tried this briefly as part of #3223 but there was a compilation error and I moved on.

cc @brunoabinader

PR #5611, opened 2016-07-07
Remove redundant GL state calls

kkaefer:
We frequently call our wrapped GL state/value system with `.reset()` when we want to set the value to its default value. However, `.reset()` currently forces the value without checking the previous state. This changes the behavior of `.reset()` to "set default value" (which does nothing when the value was already the default value).

### DIFF
--- a/include/mbgl/gl/gl_values.hpp
+++ b/include/mbgl/gl/gl_values.hpp
@@ -140,6 +140,10 @@ struct StencilOp {
     }
 };
 
+constexpr bool operator!=(const StencilOp::Type& a, const StencilOp::Type& b) {
+    return a.sfail != b.sfail || a.dpfail != b.dpfail || a.dppass != b.dppass;
+}
+
 struct DepthRange {
     struct Type { GLfloat near, far; };
     static const Type Default;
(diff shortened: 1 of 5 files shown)


kkaefer:
In my testing, this reduces the amount of GL calls in a 4-tile scene on iOS from ~1650 to ~1250, or around 30%.

jfirebaugh:
👍  I wanted to do this earlier (#3235) but tests were failing -- I probably missed the painter.cpp or test changes.

kkaefer:
Follow up at #5612

tmpsantos:
> In my testing, this reduces the amount of GL calls in a 4-tile scene on iOS from ~1650 to ~1250, or around 30%.

@kkaefer This is a very interesting metric to keep track of. Do you think it is worth having a bot to make sure we don't grow much these numbers? We have already a test rendering a reference vector tile, we could wrap it with `apitrace`.

kkaefer:
@tmpsantos yeah, that sounds useful, something like https://arewefastyet.com/ would be great

brunoabinader:
Nice!

PR #5612, opened 2016-07-07
Track bound textures

kkaefer:
Our GL state tracking system currently doesn't include bound textures, so we're repeatedly setting the bound texture as well as changing the active texture to do so.B


kkaefer:
Note that this currently limits the number of textures to 2, if we need more at some point, we'll just have to increase the array size in the `Config` object.

kkaefer:
Using the same scene as in #5611, this further reduces GL state calls from ~1250 to ~1140.

PR #5614, opened 2016-07-07
[node] Also Update::Classes when setting a paint property

tmpsantos:
Otherwise it has no effect. Tests might be working because a Update::Classes is also scheduled when adding a new layer.


tmpsantos:
:eyes: @jfirebaugh @yhahn 

This explains why this code on Qt was only working if `foobar` layer was being added. The code was based on the `node` code and was only doing a `RecalculateStyle`.

```
            QVariantMap foobar;
            foobar["id"] = "foobar";
            foobar["type"] = "fill";
            foobar["source"] = "composite";

            // Commenting this line was making the next statements stop to work.
            // Adding Update::Classes to ::setPaintProperty() fixed.
            m_map.addLayer(foobar);

            m_map.setPaintProperty("water", "fill-color", "red");
            m_map.setPaintProperty("building", "fill-color", "red");
            m_map.setPaintProperty("road-secondary-tertiary", "line-color", "red");
```

jfirebaugh:
Any idea why the [`paint-property-*` tests](https://github.com/mapbox/mapbox-gl-test-suite/tree/master/render-tests/runtime-styling) don't surface this issue?

tmpsantos:
> Any idea why the paint-property-* tests don't surface this issue?

Correct me I'm wrong, but these are not render tests are they? Internally the layer gets changed correctly, I think the issue is the render buckets don't get updated.

PR #5620, opened 2016-07-08
Include VAO and buffers in GL state tracking

kkaefer:
The remaining items that we don't track yet are buffers and vertex array objects.


jfirebaugh:
I'm 👍  on this if you want to get it rebased and passing CI.

kkaefer:
Fixed and rebased

PR #5624, opened 2016-07-08
[core] Flip implied cxx abi for mason packages

yhahn:
Switches to new behavior:

- No suffix means `cxx11abi` in `mason`
- Explicit `cxx03abi` suffix required

cc @jfirebaugh @tmpsantos


scripts/check-cxx11abi.sh
```diff
@@ -13,7 +13,7 @@ fi
 LIBSTDCPP=$(ldd $(dirname $0)/check-cxx11abi.dat |grep libstdc++ |cut -d' ' -f3)
```
tmpsantos: I think now that we inverted the semantics of this script, we should rename it to `check-cxx03abi`

platform/qt/scripts/configure.sh
```diff
@@ -8,7 +8,7 @@ BOOST_VERSION=1.60.0
 GEOMETRY_VERSION=0.8.0
 GEOJSON_VERSION=0.1.4${CXX11ABI:-}
 GEOJSONVT_VERSION=6.1.0
-GTEST_VERSION=1.7.0${CXX11ABI:-}
+GTEST_VERSION=1.7.0-1${CXX11ABI:-}
```
tmpsantos: What about `CXX11ABI` -> `CXX03ABI`?

tmpsantos:
The binary you uploaded for this one is crashing.

jfirebaugh:
Closing; we couldn't find a solution here and the work here went stale.

PR #5628, opened 2016-07-09
Quadkey tokens in tile URL templates, limited WMS support

### LINKED ISSUE #822: WMS/TMS Overlay
is there any function to make add as many WMS/TMS layers that have transperence.

### LINKED ISSUE #5485: Pass custom parameters to styleURL
I need to integrate the Bing Maps datasource into the styleURL, however instead of /z/x/y.png it uses quadKey indexing, where quadKey can be defined as:

```javascript
_quadKey: function (x, y, z) {
    var quadKey = [];
    for (var i = z; i > 0; i--) {
        var digit = '0';
        var mask = 1 << (i - 1);
        if ((x & mask) != 0) {
            digit++;
        }
        if ((y & mask) != 0) {
            digit++;
            digit++;
        }
        quadKey.push(digit);
    }
    return quadKey.join('');
}
```

How can I make the styleURL eat a custom parameter in iOS and Android SDK? say for example something like:
`https://somedatasource.com/{quadKey}.png`

1ec5:
Ported mapbox/mapbox-gl-js#2805 and mapbox/mapbox-gl-js@0877cac1d065a57c6ff2ce32f6b1201376ff8116 to support [quadkey](https://msdn.microsoft.com/en-us/library/bb259689.aspx) tokens in tile URL templates.

Ported https://github.com/mapbox/mapbox-gl-js/pull/2612#issuecomment-223595580 along with the necessary parts of mapbox/whoots-js@f2ca0caf9889581c1eda2de649eff32542c61eea for limited WMS scheme support. Added a local style to macosapp that you can use to debug WMS support in the macOS SDK: zoom in to New Jersey, then go to View ‣ Custom Style and enter `wms.json`.

Fixes #822 and fixes #5485.

/cc @lucaswoj @tmcw @bhousel

### DIFF
--- a/src/mbgl/storage/resource.cpp
+++ b/src/mbgl/storage/resource.cpp
@@ -1,7 +1,11 @@
+#include <mapbox/geometry/box.hpp>
 #include <mbgl/storage/resource.hpp>
+#include <mbgl/util/constants.hpp>
 #include <mbgl/util/string.hpp>
 #include <mbgl/util/token.hpp>
 #include <mbgl/util/url.hpp>
+
+#include <cmath>
 
 namespace mbgl {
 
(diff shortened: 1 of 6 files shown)


1ec5:
The WMS test is failing due to what appears to be a double rounding discrepancy. Does it matter that (x0, y0, z1) produces:

```
-20037508.342789245,0,0,20037508.342789245
```

instead of the GL JS expected value of:

```
-20037508.342789244,0,0,20037508.342789244
```

src/mbgl/storage/resource.cpp
```diff
@@ -55,6 +89,8 @@ Resource Resource::tile(const std::string& urlTemplate,
                         int8_t z,
                         Necessity necessity) {
     bool supportsRatio = urlTemplate.find("{ratio}") != std::string::npos;
+    auto quadKey = getQuadKey(x, y, z);
```
jfirebaugh: Move `getQuadKey` and `getTileBBox` into the lambda conditionals, so they are only called if needed.

1ec5: 👍 Fixed.

1ec5:
Going to assume the discrepancy doesn’t matter because these are projected meters. If anyone has a problem with gl-native being a nanometer off gl-js in both directions, we can address that as tail work.

kkaefer:
@1ec5 given that these units are Spherical Mercator meters, the slight discrepancy doesn't matter, since it's just 1 nanometer at the equator.

PR #5632, opened 2016-07-10
Fix the example in node/README.md

jingsam:


platform/node/README.md
```diff
@@ -29,13 +29,25 @@ npm run test-suite
...
+  },
+  ratio: 1
+};
+
+var map = new mbgl.Map(options);
 
 map.load(require('./test/fixtures/style.json'));
 
-map.render({}, function(err, buffer) {
+map.render({zoom: 0}, function(err, buffer) {
```
mikemorris: This shouldn't be necessary, as a zoom level of `0` is the default [when no options are specified](https://github.com/mapbox/mapbox-gl-native/blob/08d0a9a0e2e771b87018bcb3d4d124c52fed1689/platform/node/src/node_map.cpp#L24).

mikemorris:
This looks 👍, working examples are always better =)

platform/node/README.md
jingsam: I think empty options are a little subtle to understand, so set a zoom should be OK.

I also noticed that a [debugOptions](https://github.com/mapbox/mapbox-gl-native/blob/08d0a9a0e2e771b87018bcb3d4d124c52fed1689/platform/node/src/node_map.cpp#L32) which is undocumented in node/readme.md. Does it available in node-mapbox-gl-native?

mikemorris: Ah, yea we should probably document `debugOptions`, it's parsed out from a JavaScript object passed in the render options at https://github.com/mapbox/mapbox-gl-native/blob/08d0a9a0e2e771b87018bcb3d4d124c52fed1689/platform/node/src/node_map.cpp#L272-L299, just exposes a few things like drawing tile boundaries and label collision boxes to help debug issues during core development.

mikemorris:
Merged in https://github.com/mapbox/mapbox-gl-native/commit/5300488883b276cafca42fd48694a50eb0c4b847

PR #5638, opened 2016-07-11
5541 mapbox account manager

### LINKED ISSUE #5541: MapboxAccountManager not correctly initialised
Capturing from @geohacker that some people were having trouble setting up a working Mapview using the MapboxAccountManager setup. They were running into an invalid access token exception. This is the code they were using:

```java
    @Override
    protected void onCreate(Bundle savedInstanceState) {
        super.onCreate(savedInstanceState);
        setContentView(R.layout.activity_main);
        MapboxAccountManager.start(MainActivity.this, "somekey");
        mapView  = (MapView) findViewById(R.id.mapview);
    }
```

I asked if they could move the initialisation code to the Application object but that didn't resolve the issue. We need to look into reproducing this by setting up a new android project from scratch using `4.1.0` dependency.

tobrun:
Proposed solution that closes #5541:
 - remove requirement to call `MapboxAccountManager` before `setContentView`
 - introduces a concrete `MapboxAccountManagerNotStartedException`

This solution was tested against: 
 - `MapboxAccountManager.start` after setContentView in Activity
 - setting up an accesstoken in xml
 - setting up an accesstoken via MapboxMapOptions
 - setting up an accesstoken via `MapboxAccountManager.start` in Application object

cc @bleege @cammace @danswick @ivovandongen @zugaldia

### DIFF
--- a/platform/android/MapboxGLAndroidSDK/src/main/java/com/mapbox/mapboxsdk/MapboxAccountManager.java
+++ b/platform/android/MapboxGLAndroidSDK/src/main/java/com/mapbox/mapboxsdk/MapboxAccountManager.java
@@ -2,8 +2,10 @@ package com.mapbox.mapboxsdk;
 
 import android.content.Context;
 import android.text.TextUtils;
+
 import com.mapbox.mapboxsdk.constants.MapboxConstants;
 import com.mapbox.mapboxsdk.exceptions.InvalidAccessTokenException;
+import com.mapbox.mapboxsdk.exceptions.MapboxAccountManagerNotStartedException;
 import com.mapbox.mapboxsdk.telemetry.MapboxEventManager;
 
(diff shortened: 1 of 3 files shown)


bleege:
Just to clarify, even though this will enable more flexibility as to where `MapboxAccountManager.start` can be called so that things will work, for the purpose of consistency in our documentation let's continue to use the recommendation of "first Mapbox code in your app".  This means recommending that it's the first Mapbox code in the app in either `Application.onCreate()` or `Activity.onCreate()` as described https://github.com/mapbox/mapbox-gl-native/issues/5541#issuecomment-231419084

zugaldia:
@bleege Sounds good to me. Is this PR good to merge otherwise?

bleege:
> Is this PR good to merge otherwise?

Deferring to @tobrun on this.

PR #5551, opened 2016-07-03
[qt] Implement Qt runtime style API

brunoabinader:
Implements:
- [x] Set filters
- [x] Non-custom layers (line, circle, symbol, raster, background)
- [x] Sources (raster, vector, GeoJSON*)
- [x] Set paint properties
- [x] Set layout properties

Caveats:
- Use `QVariant` to abstract variant types
- Strict type check for filter values via `Qmapbox::FilterValueType`
- Implemented Qt-specific conversion helpers
- Replace `QMapboxGL::removeCustomLayer()` with generic `QMapboxGL::removeLayer()`
- `QMapboxGL` is getting big - we might want to move the conversion functions to a separate file

Missing:
- Inline GeoJSON data input for GeoJSON source
- Static assertion checks

Fixes #5511.

:eyes: @yhahn @tmpsantos

PR #5642, opened 2016-07-11
[Qt] Implement the runtime style API

tmpsantos:
Based on #5551, but using the core datatype converters.


tmpsantos:
:eyes: @brunoabinader

platform/qt/include/qmapboxgl.hpp
```diff
@@ -174,6 +174,8 @@ class Q_DECL_EXPORT QMapboxGL : public QObject
 
     void removeAnnotation(QMapbox::AnnotationID);
 
+    void setLayoutProperty(const QString& layer, const QString& property, const QVariant& value);
```
brunoabinader: Let's use the Qt style on the header files e.g. put the `&` together with the param name.

platform/qt/src/qt_conversion.hpp
```diff
@@ -0,0 +1,106 @@
...
+        }
+
+        ++iter;
+    }
+
+    return {};
+}
+
+inline optional<bool> toBool(const QVariant& value) {
+    if (value.type() == QVariant::Bool) {
```
brunoabinader: I think a ternary operator would suit fine here, but this is fine too.

platform/qt/src/qt_conversion.hpp
```diff
@@ -0,0 +1,106 @@
...
+        return value.toBool();
+    } else {
+        return {};
+    }
+}
+
+inline optional<float> toNumber(const QVariant& value) {
+    if (value.type() == QVariant::Double) {
```
brunoabinader: Ditto.

platform/qt/src/qt_conversion.hpp
```diff
@@ -0,0 +1,106 @@
...
+        return value.toFloat();
+    } else {
+        return {};
+    }
+}
+
+inline optional<std::string> toString(const QVariant& value) {
+    if (value.type() == QVariant::String) {
```
brunoabinader: Can we use `switch (value.type())` here instead?

platform/qt/src/qt_conversion.hpp
```diff
@@ -0,0 +1,106 @@
...
+        return value.toString().toStdString();
+    } else if (value.type() == QVariant::Color) {
+        return value.value<QColor>().name().toStdString();
+    } else {
+        return {};
+    }
+}
+
+inline optional<Value> toValue(const QVariant& value) {
```
brunoabinader: Ditto.

platform/qt/src/http_request.cpp
```diff
@@ -39,7 +39,8 @@ QNetworkRequest HTTPRequest::networkRequest() const
     req.setRawHeader("User-Agent", "MapboxGL/1.0 [Qt]");
 
     if (m_resource.priorEtag) {
-        req.setRawHeader("If-None-Match", QByteArray(m_resource.priorEtag->data(), m_resource.priorEtag->size()));
+        auto etag = m_resource.priorEtag;
```
brunoabinader: Can we use `const auto` here?

brunoabinader:
Minor comments above, but this is awesome! :tada:

platform/qt/src/qt_conversion.hpp
tmpsantos: List initialization would have to be explicit :cry:

platform/qt/src/qt_geojson.hpp
```diff
@@ -0,0 +1,55 @@
...
+    rapidjson::GenericDocument<rapidjson::UTF8<>, rapidjson::CrtAllocator> d;
+    if (data.endsWith(char(0))) {
+        d.Parse<0>(value.toByteArray().data());
+    } else {
+        d.Parse<0>(value.toByteArray().constData());
+    }
+
+    // Needed to keep the error message alive
+    // when we go out of this scope.
+    static std::string error;
```
jfirebaugh: Can you open a ticket to change `Error` so we can remove this?

tmpsantos: https://github.com/mapbox/mapbox-gl-native/pull/5655

Issue #5438, opened 2016-06-22
Incorrect MyLocationView transformation


justasm:
**Platform:** Android
**Mapbox SDK version:** 4.1.0-beta.3

### Steps to trigger behavior

 1. Enable `MyBearingTracking.COMPASS` tracking mode.
 2. Tilt map around x axis (i.e. pitch) using two-finger shove gesture.
 3. Rotate map around z axis (i.e. bearing, aka direction) using two-finger rotate gesture.

### Expected behavior
![screenshot_20160622-182419](https://cloud.githubusercontent.com/assets/1081815/16272594/0215e932-38a7-11e6-96de-75254290f02a.png)

### Actual behavior
![screenshot_20160622-182544](https://cloud.githubusercontent.com/assets/1081815/16272598/0667dc84-38a7-11e6-950a-5af5fdb0d2d6.png)

### Discussion
The 3D transformation calculation for `MyLocationView` is

1. incorrect - the compass bearing / direction rotation is applied as `M' = R(degrees) * M` using [`Matrix#postRotate(float)`][mylocationview-postrotate-bearing], whereas it should be calculated as `M' = M * R(degrees)`, e.g. by using `Matrix#preRotate(float)`.
2. incomplete - map rotation around the z axis, i.e. map bearing / direction, is not taken into account. The compass bearing indicator is therefore incorrect when the map is not pointing north. Map rotation gestures should be propagated to `MyLocationView`, similar to [how tilt is][mylocationview-settilt], and accounted for when calculating the transformation.
3. needlessly complex - it uses both the `Camera` and `Matrix` classes, whereas the entire transformation could be calculated using just `Camera` and applied using `Camera#applyToCanvas(canvas)`.

[mylocationview-postrotate-bearing]: https://github.com/mapbox/mapbox-gl-native/blob/062c4ce773079fc6b408a50f64ab57c280dacdd9/platform/android/MapboxGLAndroidSDK/src/main/java/com/mapbox/mapboxsdk/maps/widgets/MyLocationView.java#L272
[mylocationview-settilt]: https://github.com/mapbox/mapbox-gl-native/blob/062c4ce773079fc6b408a50f64ab57c280dacdd9/platform/android/MapboxGLAndroidSDK/src/main/java/com/mapbox/mapboxsdk/maps/widgets/MyLocationView.java#L302

tobrun:
[...]

PR #5643, opened 2016-07-11
[android] - remove duplicate bearing/direction API from MapView

tobrun:
We currently support the same API twice on the MapView object. Removing direction api in favor of bearing. Bearing has been the terminology used for internal API while direction reflects our public API. The methods contained in MapView are for internal use only.


tobrun:
I need this PR to continue work in #5438, CI approved, going to rebase and merge

PR #5644, opened 2016-07-11
[android] latlngbounds includes added in builder

### LINKED ISSUE #5574: Add a fromList method in LatLngBound
Currently the only way to add points to a LatLngBound object is to using the `.include()` method which only takes in individual latlng points. It would be useful to also expose a method that consumes a list of latlng points. An example use would be when you want to fit the map to a polyline, you would only need to pass in the list of points rather then creating a for loop and using the include. 

cc: @tobrun

cammace:
Fixes #5574 

cc: @tobrun


tobrun:
@cammace looking great, one suggestion is to back this change up with an unit test?

tobrun:
Test was added, going to rebase and merge

PR #5646, opened 2016-07-12
[ios] Pass along `animated` property when selecting annotation views

friedbunny:
Fixes a bug where the selection of annotation views would never be animated, because `-[MGLMapView selectAnnotation:animated:]` wasn’t passing along its `animated` value.

This change means that tapped view annotations now properly receive `animated = YES` — e.g., when a custom view annotation class overrides `-setSelected:animated:`.

I’ve put this on the v3.3.0 milestone because it’s a minor change that shouldn’t need serious testing. We could push this to v3.3.1, though, if we don’t think there’s enough time for another RC before release.

/cc @boundsj @frederoni @1ec5


1ec5:
👍 Let’s get this into v3.3.0.

friedbunny:
Per chat with @1ec5, we’re going to try to get this in v3.3.0.

PR #5648, opened 2016-07-12
Wrapped world is blank

brunoabinader:
When panning past the antimeridian, the wrapped copy of the world is blank until the antimeridian goes past the center of the view.

macos
ios

Reproduces in fe99295970a1f5e2d8834b35ce327938a1738240 on macOS and the iOS Simulator.

/cc @jfirebaugh

### DIFF
--- a/src/mbgl/map/transform_state.cpp
+++ b/src/mbgl/map/transform_state.cpp
@@ -16,7 +16,7 @@ TransformState::TransformState(ConstrainMode constrainMode_, ViewportMode viewpo
 #pragma mark - Matrix
 
 void TransformState::matrixFor(mat4& matrix, const UnwrappedTileID& tileID) const {
-    const uint32_t tileScale = 1u << tileID.canonical.z;
+    const uint64_t tileScale = 1ull << tileID.canonical.z;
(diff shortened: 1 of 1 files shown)


brunoabinader:
Nice catch @1ec5 - fixing it by reverting a line in https://github.com/mapbox/mapbox-gl-native/commit/1d46e83e3087f0d392c74c40d6b562359ed91576 - `tileScale` indeed needs to be `uint64_t` not because of its input value, but because when adding it together in `matrix::translate` it overflows.

jfirebaugh:
@brunoabinader Can you please add a test case to https://github.com/mapbox/mapbox-gl-test-suite/tree/master/render-tests/regressions for this?

brunoabinader:
> Can you please add a test case to https://github.com/mapbox/mapbox-gl-test-suite/tree/master/render-tests/regressions for this?

Will do 👍

brunoabinader:
Regression test added in https://github.com/mapbox/mapbox-gl-test-suite/commit/1589dc7fa1f76473b07d882da70b3a4d99b0d8ce and https://github.com/mapbox/mapbox-gl-native/pull/5670.

PR #5655, opened 2016-07-12
[core] Report conversion errors using std::string

tmpsantos:
char* increases the risk of pointing to a invalid reference. Qt had to use a static variable as retainer to workaround.


jfirebaugh:
👍

PR #5658, opened 2016-07-12
[qt] if `qmake` isn't available, try `qmake-qt5`

akikoskinen:
At least on the Fedora system I'm using the qt5 development packages don't provide a binary named `qmake` (at least not one that would be in `$PATH`). This is a simple change that uses `qmake-qt5` if `qmake` isn't found. If `qmake` is available the behaviour is as before.

It does NOT test whether `qmake-qt5` is available. That is similar to the behaviour this script had before: if `qmake` wasn't found, it just exits with an error.

### DIFF
--- a/platform/qt/scripts/configure.sh
+++ b/platform/qt/scripts/configure.sh
@@ -38,7 +38,12 @@ fi
 function print_qt_flags {
     mason install Qt system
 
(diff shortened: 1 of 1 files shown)


brunoabinader:
Good stuff, kiitos @akikoskinen!

PR #5464, opened 2016-07-12
mbgl::FillAnnotation::outlineColor should be an optional<Color>

brunoabinader:
Ref: [include/mbgl/annotation/annotation.hpp#L41](https://github.com/mapbox/mapbox-gl-native/blob/master/include/mbgl/annotation/annotation.hpp#L41)

This implies adapting all platform-specific code to return an optional color value for this, specifically:
- Android: [platform/android/src/jni.cpp#L788](https://github.com/mapbox/mapbox-gl-native/blob/master/platform/android/src/jni.cpp#L788)
- Darwin: [platform/darwin/src/MGLPolygon.mm#L48](https://github.com/mapbox/mapbox-gl-native/blob/master/platform/darwin/src/MGLPolygon.mm#L48)

Then we could cleanup the outline color check done in [src/mbgl/renderer/painter_fill.cpp#L27-L31](https://github.com/mapbox/mapbox-gl-native/blob/master/src/mbgl/renderer/painter_fill.cpp#L27-L31).

/cc @tobrun @1ec5 @jfirebaugh

### DIFF
--- a/platform/ios/src/MGLMapViewDelegate.h
+++ b/platform/ios/src/MGLMapViewDelegate.h
@@ -247,7 +247,7 @@ NS_ASSUME_NONNULL_BEGIN
  @param annotation The annotation being rendered.
  @return A line width for the polyline, measured in points.
  */
-- (CGFloat)mapView:(MGLMapView *)mapView lineWidthForPolylineAnnotation:(MGLPolyline *)annotation;
+- (float)mapView:(MGLMapView *)mapView lineWidthForPolylineAnnotation:(MGLPolyline *)annotation;
 
 #pragma mark Managing Annotation Views
 
--- a/src/mbgl/style/layers/background_layer_properties.hpp
+++ b/src/mbgl/style/layers/background_layer_properties.hpp
@@ -17,7 +17,7 @@ public:
     void cascade(const CascadeParameters&);
     bool recalculate(const CalculationParameters&);
 
-    PaintProperty<Color> backgroundColor { { 0, 0, 0, 1 } };
+    PaintProperty<Color> backgroundColor { Color::black() };
     PaintProperty<std::string, CrossFadedPropertyEvaluator> backgroundPattern { "" };
     PaintProperty<float> backgroundOpacity { 1 };
 };
--- a/src/mbgl/style/layers/fill_layer_properties.hpp
+++ b/src/mbgl/style/layers/fill_layer_properties.hpp
@@ -20,7 +20,7 @@ public:
     PaintProperty<bool> fillAntialias { true };
     PaintProperty<float> fillOpacity { 1 };
     PaintProperty<Color> fillColor { { 0, 0, 0, 1 } };
-    PaintProperty<Color> fillOutlineColor { { 0, 0, 0, -1 } };
+    PaintProperty<Color> fillOutlineColor { {} };
     PaintProperty<std::array<float, 2>> fillTranslate { {{ 0, 0 }} };
     PaintProperty<TranslateAnchorType> fillTranslateAnchor { TranslateAnchorType::Map };
     PaintProperty<std::string, CrossFadedPropertyEvaluator> fillPattern { "" };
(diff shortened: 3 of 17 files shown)


src/mbgl/style/layers/background_layer_properties.hpp
```diff
@@ -17,7 +17,7 @@ class BackgroundPaintProperties {
     void cascade(const CascadeParameters&);
     bool recalculate(const CalculationParameters&);
 
-    PaintProperty<Color> backgroundColor { { 0, 0, 0, 1 } };
+    PaintProperty<Color> backgroundColor { Color::black() };
```
jfirebaugh: This change will get blown away by the next run of `make style-code`.

src/mbgl/style/layers/fill_layer_properties.hpp
```diff
@@ -19,8 +19,8 @@ class FillPaintProperties {
 
     PaintProperty<bool> fillAntialias { true };
     PaintProperty<float> fillOpacity { 1 };
-    PaintProperty<Color> fillColor { { 0, 0, 0, 1 } };
-    PaintProperty<Color> fillOutlineColor { { 0, 0, 0, -1 } };
+    PaintProperty<Color> fillColor { Color::black() };
+    PaintProperty<Color> fillOutlineColor { {} };
```
jfirebaugh: Ditto.

src/mbgl/style/paint_property.hpp
```diff
@@ -39,8 +39,12 @@ class PaintProperty {
         return *this;
     }
 
+    bool isUndefined() const {
+        return values.find(ClassID::Default) == values.end();
+    }
+
     const PropertyValue<T>& get() const {
-        return values.at(ClassID::Default);
+        return isUndefined() ? values.at(ClassID::Fallback) : values.at(ClassID::Default);
```
jfirebaugh: `get()` is mainly used for implementing `Layer` subclass property getters, e.g. `BackgroundLayer::getBackgroundColor()`. I intended for these to return an undefined `PropertyValue<T>` if the value wasn't set, but it looks like that's always been broken, because `values.at(ClassID::Default)` would throw in that case.

However, returning `values.at(ClassID::Fallback)` doesn't provide the desired semantics. I think the correct implementation is something like:

```
     PropertyValue<T> get() const {
         auto it = values.find(ClassID::Default);
         return it == value.end() ? PropertyValue<T>() : it->second;
     }
```

jfirebaugh:
@1ec5 Can you review the Objective-C changes here?

platform/ios/src/MGLMapViewDelegate.h
```diff
@@ -247,7 +247,7 @@ NS_ASSUME_NONNULL_BEGIN
  @param annotation The annotation being rendered.
  @return A line width for the polyline, measured in points.
  */
-- (CGFloat)mapView:(MGLMapView *)mapView lineWidthForPolylineAnnotation:(MGLPolyline *)annotation;
+- (float)mapView:(MGLMapView *)mapView lineWidthForPolylineAnnotation:(MGLPolyline *)annotation;
```
1ec5: This would be a backwards-incompatible change on 64-bit systems, where `CGFloat` is a typedef for `double`. Values as large as `FLT_MAX` are nonsensical, but it’s a convention in Cocoa APIs to use `CGFloat` for any value measured in screen coordinates. This should be reverted to `CGFloat` for consistency with all the other geometry APIs in this SDK and in Cocoa frameworks.

src/mbgl/style/layers/background_layer_properties.hpp
brunoabinader: Sorry I didn't paid attention to the disclaimer on the first line. Will update `layer_properties.hpp.ejs` file accordingly.

platform/ios/src/MGLMapViewDelegate.h
brunoabinader: Thanks for the explanation @1ec5 - `PropertyValue<float>` doesn't like to convert from double, so I guess I'll need to add a static_cast in that case.

PR #5660, opened 2016-07-12
Reorganize jazzy’s navigation categories; cleanup docs

friedbunny:
This pull request reorganizes jazzy’s navigation categories by demoting less commonly used parts of our SDK to generic “Other” categories at the bottom of the list. In particular, this makes the Maps and Offline sections easier to understand.

I’ve also cleaned up some typos in the inline documentation and added two back-references to enum docs about which MGLMapView properties they are used with.

mapbox-ios-sdk-reference- 20160712

/cc @1ec5

### DIFF
--- a/platform/ios/src/MGLMapView.h
+++ b/platform/ios/src/MGLMapView.h
@@ -31,8 +31,8 @@ extern const CGFloat MGLMapViewDecelerationRateFast;
 extern const CGFloat MGLMapViewDecelerationRateImmediate;
 
 /**
- The vertical alignment of an annotation within a map view. Used with the
- `MGLMapView.userLocationVerticalAlignment` property.
+ The vertical alignment of an annotation within a map view. Used with
+ `MGLMapView.userLocationVerticalAlignment`.
  */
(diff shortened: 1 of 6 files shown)


1ec5:
Can you make the equivalent changes to the macOS SDK’s jazzy configuration file, while you’re at it?

platform/ios/src/MGLMapView.h
```diff
@@ -30,7 +30,12 @@ extern const CGFloat MGLMapViewDecelerationRateFast;
 /** Disables decleration in a map view. */
 extern const CGFloat MGLMapViewDecelerationRateImmediate;
 
-/** The vertical alignment of an annotation within a map view. */
+/**
+ The vertical alignment of an annotation within a map view.
+
+ @note This is currently used with the `MGLMapView.userLocationVerticalAlignment`
```
1ec5: No need for `@note`.

friedbunny:
Will do that now.

1ec5:
👍

PR #5671, opened 2016-07-13
[ios] fixes #3697 compass, logo and attribution now respects insets

### LINKED ISSUE #3697: Map content insets should offset compass, logo, attribution beyond layout guides
With current version the map insets on iOS work (user center is moved correctly) but the compass view is not moved inside the insets.
Same with the mapbox logo.
Insets are mostly used because another view hides temporally a part of the map. Users still want to see its compass.

frederoni:
Compass, logo and attribution button new respects the map view's content inset.

👀 @1ec5 @friedbunny @boundsj

### DIFF
--- a/platform/ios/src/MGLMapView.mm
+++ b/platform/ios/src/MGLMapView.mm
@@ -741,17 +741,6 @@ mbgl::Duration MGLDurationInSeconds(NSTimeInterval duration)
     [constraintParentView removeConstraints:self.compassViewConstraints];
     [self.compassViewConstraints removeAllObjects];
 
-    if (viewController)
-    {
-        [self.compassViewConstraints addObject:
-         [NSLayoutConstraint constraintWithItem:compassContainer
-                                      attribute:NSLayoutAttributeTop
-                                      relatedBy:NSLayoutRelationGreaterThanOrEqual
-                                         toItem:viewController.topLayoutGuide
-                                      attribute:NSLayoutAttributeBottom
-                                     multiplier:1
-                                       constant:5 + self.contentInset.top]];
-    }
     [self.compassViewConstraints addObject:
      [NSLayoutConstraint constraintWithItem:compassContainer
                                   attribute:NSLayoutAttributeTop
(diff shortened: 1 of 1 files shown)


1ec5:
When the `contentInset` property changes, the constraints need to be updated immediately.

frederoni:
**Note:** `automaticallyAdjustsScrollViewInsets` must be set to NO on the VC for the `contentInset` to be persistent. If not, `contentInset` will be adjusted automatically and logo, attribution and compass will fallback to its original position next time `-layoutSubviews` is called. This is consistent with how UIScrollView works.  @1ec5  👀

platform/ios/src/MGLMapView.mm
```diff
@@ -974,6 +975,9 @@ - (void)setContentInset:(UIEdgeInsets)contentInset animated:(BOOL)animated
     {
         [self didUpdateLocationWithUserTrackingAnimated:animated];
     }
+    
+    // Compass, logo and attribution button constraints needs to be updated.
+    [self setNeedsUpdateConstraints];
```
1ec5: `-layoutSubviews` calls `-adjustContentInset`, which calls `-setContentInset:` and `-setContentInset:animated:`, which calls `-setNeedsUpdateConstraints`. I’m not quite sure, but are there any cases where that’ll cause a feedback loop?

frederoni: Not that I can find. `-setNeedsUpdateConstraints` will make sure a future call to `-updateConstraintsIfNeeded` calls `-updateConstraints` = no loop. If `-adjustContentInset` would get triggered in a loop, the loop is interrupted either because the previous `contentInset` is equal to the new or in case you're setting your own `contentInset`, it would get interrupted before that because of `automaticallyAdjustsScrollViewInsets=NO`

1ec5: :+1:

postmechanical:
Will this make the 3.3.0 release?

1ec5:
@postmechanical, no, v3.3.0 has already been released. 😃 But it will be in v3.3.1.

PR #5675, opened 2016-07-13
[core] Revert workaround for std::atomic on ARMv5

tmpsantos:
Got fix on r12.

https://github.com/android-ndk/ndk/issues/31


tmpsantos:
As side note, the `log2` workaround is still needed.

PR #5677, opened 2016-07-13
[ios] Also pass `animated` property when deselecting annotation views

friedbunny:
This is the view annotation deselection counterpart to https://github.com/mapbox/mapbox-gl-native/pull/5646's selection (that I neglected to check). Going to put this on v3.3.1, as this is a minor fix and we can live without it in the short term.

/cc @1ec5 @boundsj @frederoni


friedbunny:
In case anyone needs a work around, my suggestion is to just ignore the incoming `animated` property in your MGLAnnotationView subclass’ implementation of `-setSelected:selected animated:`.

PR #5681, opened 2016-07-14
Event Queue Cap

### LINKED ISSUE #5680: Send Android Telemetry Events Automatically When Count Reaches 1000
The Android Telemetry systems should add a cap to the number of events stored before automatically triggering a push to the server.  Currently, the Android Telemetry system attempts to send data every 3 minutes in an effort to minimize the number of power consuming networking calls.  However this can have the effect of allowing large numbers of events to queue up resulting in larger data sets, longer networking calls, and issues processing the number of events at the server.

/cc @mapbox/telemetry

bleege:
Closes #5680


bleege:
Squashed and Merged!

PR #5689, opened 2016-07-14
Update color layer properties code generator

brunoabinader:
Follow-up to review comments from https://github.com/mapbox/mapbox-gl-native/pull/5464.

:eyes: @1ec5 @jfirebaugh

### DIFF
--- a/src/mbgl/style/layers/layer_properties.hpp.ejs
+++ b/src/mbgl/style/layers/layer_properties.hpp.ejs
@@ -37,7 +37,13 @@ public:
 <% if (/-pattern$/.test(property.name) || property.name === 'line-dasharray') { -%>
     PaintProperty<<%- propertyType(property) %>, CrossFadedPropertyEvaluator> <%- camelizeWithLeadingLowercase(property.name) %> { <%- defaultValue(property) %> };
 <% } else if (property.name === 'fill-outline-color') { -%>
-    PaintProperty<<%- propertyType(property) %>> <%- camelizeWithLeadingLowercase(property.name) %> { { 0, 0, 0, -1 } };
+    PaintProperty<<%- propertyType(property) %>> <%- camelizeWithLeadingLowercase(property.name) %> { {} };
+<% } else if (property.name.endsWith('color') && defaultValue(property) === '{ 0, 0, 0, 0 }') { -%>
+    PaintProperty<<%- propertyType(property) %>> <%- camelizeWithLeadingLowercase(property.name) %> { {} };
+<% } else if (property.name.endsWith('color') && defaultValue(property) === '{ 0, 0, 0, 1 }') { -%>
+    PaintProperty<<%- propertyType(property) %>> <%- camelizeWithLeadingLowercase(property.name) %> { Color::black() };
+<% } else if (property.name.endsWith('color') && defaultValue(property) === '{ 1, 1, 1, 1 }') { -%>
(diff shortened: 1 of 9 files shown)


1ec5:
Thanks. 👍 In the future, we can look into changing the return type of `-mapView:alphaForShapeAnnotation:` to `float`, since that would be more consistent with Core Graphics. This is good for now though.

src/mbgl/style/layers/layer_properties.hpp.ejs
```diff
@@ -37,7 +37,13 @@ public:
 <% if (/-pattern$/.test(property.name) || property.name === 'line-dasharray') { -%>
     PaintProperty<<%- propertyType(property) %>, CrossFadedPropertyEvaluator> <%- camelizeWithLeadingLowercase(property.name) %> { <%- defaultValue(property) %> };
 <% } else if (property.name === 'fill-outline-color') { -%>
-    PaintProperty<<%- propertyType(property) %>> <%- camelizeWithLeadingLowercase(property.name) %> { { 0, 0, 0, -1 } };
+    PaintProperty<<%- propertyType(property) %>> <%- camelizeWithLeadingLowercase(property.name) %> { {} };
+<% } else if (property.name.endsWith('color') && defaultValue(property) === '{ 0, 0, 0, 0 }') { -%>
```
jfirebaugh: I'd rather just leave these as `{ 0, 0, 0, 0 }`, as they appear in the spec, than have special cases here that translate them to something else.

src/mbgl/style/layers/layer_properties.hpp.ejs
```diff
@@ -37,7 +37,13 @@ public:
 <% if (/-pattern$/.test(property.name) || property.name === 'line-dasharray') { -%>
     PaintProperty<<%- propertyType(property) %>, CrossFadedPropertyEvaluator> <%- camelizeWithLeadingLowercase(property.name) %> { <%- defaultValue(property) %> };
 <% } else if (property.name === 'fill-outline-color') { -%>
-    PaintProperty<<%- propertyType(property) %>> <%- camelizeWithLeadingLowercase(property.name) %> { { 0, 0, 0, -1 } };
+    PaintProperty<<%- propertyType(property) %>> <%- camelizeWithLeadingLowercase(property.name) %> { {} };
```
jfirebaugh: Given https://github.com/mapbox/mapbox-gl-native/blob/c670b08ecb431b1bfd02da748de3d9fa16a757ec/src/mbgl/renderer/painter_fill.cpp#L27-L27, this should expand to `PaintProperty<Color> fillOutlineColor { };` so it's in the "undefined" state, not a fully-transparent color.

brunoabinader: I tried this before, but the current implementation of `PaintProperty` does not allow an empty constructor like `PropertyValue` does:
```
../../src/mbgl/style/layers/fill_layer_properties.hpp:23:43: error: no matching constructor for initialization of 'PaintProperty<mbgl::Color>'
    PaintProperty<Color> fillOutlineColor {};
                                          ^~
```
Are you proposing then to have `PaintProperty` use the same variant mechanism to store its default value?

jfirebaugh: Ah, I see. `PaintProperty<Color>({}).isUndefined()` is still `true`. 👍

PR #5696, opened 2016-07-14
Add an API to enable/disable delivering messages when the region goes inactive

### LINKED ISSUE #5538: Stop delivering status updates when offline download is canceled
**Platform:** iOS and Android
**Mapbox SDK version:** latest of each

### Steps to trigger behavior

 1. Initiate an offline download
 2. Cancel the download

### Expected behavior

Download immediately stops and no further messages are sent. 

### Actual behavior

Messages continue to be sent. 

Related issues: https://github.com/mapbox/mapbox-gl-native/issues/5531, https://github.com/mapbox/mapbox-gl-native/issues/4418, https://github.com/mapbox/mapbox-gl-native/issues/4229

zugaldia:
This PR adds a new API to the `OfflineRegion` object to enable or disable the delivery of messages once a region state is set as `STATE_INACTIVE`. Note that it doesn't change the internal state of the region, this only changes whether the callback will emit or not notifications.

From now on, by default, the API won't provide updates when a region is set inactive although this behavior can be changed via `OfflineRegion.setDeliverInactiveMessages(true)`.

/cc: @danswick @1ec5 

This handles the Android side of #5538.

### DIFF
--- a/platform/android/MapboxGLAndroidSDK/src/main/java/com/mapbox/mapboxsdk/offline/OfflineRegion.java
+++ b/platform/android/MapboxGLAndroidSDK/src/main/java/com/mapbox/mapboxsdk/offline/OfflineRegion.java
@@ -140,6 +140,34 @@ public class OfflineRegion {
     public static final int STATE_INACTIVE = 0;
     public static final int STATE_ACTIVE = 1;
 
+    // Keep track of the region state
+    private int state = STATE_INACTIVE;
+
+    private boolean deliverInactiveMessages = false;
+
+    /**
+     * Gets whether or not the `OfflineRegionObserver` will continue to deliver messages even if
+     * the region state has been set as STATE_INACTIVE.
+     */
+    public boolean isDeliveringInactiveMessages() {
+        return deliverInactiveMessages;
+    }
+
+    /**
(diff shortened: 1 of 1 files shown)


1ec5:
There’s nothing inherently platform-specific about this approach, is there? I think suppressing updates is something mbgl should be responsible for rather than every SDK individually.

platform/android/MapboxGLAndroidSDK/src/main/java/com/mapbox/mapboxsdk/offline/OfflineRegion.java
```diff
@@ -140,6 +140,34 @@
...
+    // Keep track of the region state
+    private int state = STATE_INACTIVE;
+
+    private boolean deliverInactiveMessages = false;
+
+    /**
+     * Gets whether or not the `OfflineRegionObserver` will continue to deliver messages even if
+     * the region state has been set as STATE_INACTIVE.
+     */
+    public boolean isDeliverInactiveMessages() {
```
1ec5: This sounds grammatically incorrect. Is it necessary to say “isDeliver” here instead of “isDelivering” or “delivers”?

PR #5555, opened 2016-07-15
Add ability to disable overdraw shaders 

brunoabinader:
#5371 and #5534 add a second set of shaders for overdraw mode. We should completely disable this with `ifdefs` or so when we're building a release build for publication to avoid additional shader compilation overhead. Overdraw mode is a debugging tool and shouldn't be in the SDK.

/cc @brunoabinader

### DIFF
--- a/platform/ios/CHANGELOG.md
+++ b/platform/ios/CHANGELOG.md
@@ -9,6 +9,7 @@ Mapbox welcomes participation and contributions from everyone.  Please read [CON
 * The `text-pitch-alignment` property is now supported in stylesheets for improved street label legibility on a tilted map. ([#5288](https://github.com/mapbox/mapbox-gl-native/pull/5288))
 * The `icon-text-fit` and `icon-text-fit-padding` properties are now supported in stylesheets, allowing the background of a shield to automatically resize to fit the shield’s text. ([#5334](https://github.com/mapbox/mapbox-gl-native/pull/5334))
 * Improved the performance of relocating a non-view-backed point annotation by changing its `coordinate` property. ([#5385](https://github.com/mapbox/mapbox-gl-native/pull/5385))
+* MGLMapDebugOverdrawVisualizationMask does nothing in Release builds of the SDK. This is disabled for performance reasons. ([#5555](https://github.com/mapbox/mapbox-gl-native/pull/5555))
 
 ## 3.3.1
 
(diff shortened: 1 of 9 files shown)


brunoabinader:
Moved discussion about debug options in release mode to #5699.

brunoabinader:
Node render test `debug/overdraw` fails because we are running the render tests on a Travis node build bot in release mode. @mikemorris any objections on switching this bot to debug mode?

mikemorris:
@brunoabinader 👍 to `BUILDTYPE=Debug` on that bot

brunoabinader:
@1ec5 @mikemorris could you please :eyes: on the changelog changes before 🚢 ?

platform/ios/CHANGELOG.md
```diff
@@ -6,6 +6,8 @@ Mapbox welcomes participation and contributions from everyone.  Please read [CON
 
 * Added [quadkey](https://msdn.microsoft.com/en-us/library/bb259689.aspx) support and limited WMS support in raster tile URL templates. ([#5628](https://github.com/mapbox/mapbox-gl-native/pull/5628))
```
1ec5: Nit: remove this blank line.

platform/ios/CHANGELOG.md
```diff
@@ -6,6 +6,8 @@ Mapbox welcomes participation and contributions from everyone.  Please read [CON
...
 
+* Map debug options 'overdraw' and 'stencil clip' are now disabled (no-ops) in release mode. ([#5555](https://github.com/mapbox/mapbox-gl-native/pull/5555))
```
1ec5: Refer to the actual symbol names. Also, no need to say "in release mode", as that's the only build configuration we support.

1ec5:
For iOS and macOS, can you mark the enum values with ` __attribute__((deprecated("…")));` and point out that they're no-ops in the documentation comment? You can condition the deprecated attribute on the `DEBUG` build variable.

If you think it's still useful to have these options in iosapp, you can conditionally compile the code that adds and carries out the action. For macosapp, change the menu item's validation code to conditionally return YES for debug and NO otherwise.

1ec5:
If deprecation and conditional compiling is too much trouble, it'd be fine to simply document the two enum options as being nonfunctional.

PR #5700, opened 2016-07-15
Introduce a getDatabasePath() method that enables switching storage paths

### LINKED ISSUE #5589: Enable external storage to host the tiles database
Currently, the Android SDK chooses a default location in the internal storage via [`Context.getFilesDir().getAbsolutePath()`](https://github.com/mapbox/mapbox-gl-native/blob/master/platform/android/MapboxGLAndroidSDK/src/main/java/com/mapbox/mapboxsdk/offline/OfflineManager.java#L90-L93) to save the database used to store offline and ambient caching tiles.

We'd like to add a setting to enable this location to be in [external storage](https://developer.android.com/guide/topics/data/data-storage.html#filesExternal), for example an SD Card. Note that this doesn't mean that multiple concurrent sources would be supported, tiles still need to be stored in the same database file on disk.

Related tickets: https://github.com/mapbox/mapbox-gl-native/issues/4386 https://github.com/mapbox/mapbox-gl-native/issues/4223

/cc: @jfirebaugh @1ec5 @hjudge

zugaldia:
In order to enable this functionality, the developer can now set a `SetStorageExternal` boolean flag in the `AndroidManifest.xml` file to switch to external storage:

```
<meta-data
	android:name="com.mapbox.SetStorageExternal"
	android:value="true" />
```

The default value for storage remains internal storage (`false`), like until now. 

Finally, note that this PR does not enable multiple sources for storage, tiles still need to be stored in the same database file on disk. The storage path, internal or external, is shared between the `NativeMapView` and the `OfflineManager` objects and, because it's a Manifest value, cannot be changed programmatically at runtime.

Fixes #5589.

/cc: @cammace @danswick @jfirebaugh @1ec5 @hjudge

### DIFF
--- a/platform/android/MapboxGLAndroidSDK/src/main/java/com/mapbox/mapboxsdk/constants/MapboxConstants.java
+++ b/platform/android/MapboxGLAndroidSDK/src/main/java/com/mapbox/mapboxsdk/constants/MapboxConstants.java
@@ -31,6 +31,16 @@ public class MapboxConstants {
     public static final String KEY_META_DATA_STAGING_ACCESS_TOKEN = "com.mapbox.TestEventsAccessToken";
 
     /**
+     * Key used to switch storage to external in AndroidManifest.xml
+     */
+    public final static String KEY_META_DATA_SET_STORAGE_EXTERNAL = "com.mapbox.SetStorageExternal";
+
+    /**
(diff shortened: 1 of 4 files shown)


tobrun:
LGTM :shipit:

zugaldia:
@tobrun Thanks for the 👀 . Merged.

PR #5704, opened 2016-07-15
NodeRequest member function instead of v8::Context

mikemorris:
Resurrects #5527, but with a JavaScript shim in front to preserve the callback-passing public API.

/cc @tmpsantos @jfirebaugh @miccolis

### DIFF
--- a/package.json
+++ b/package.json
@@ -6,7 +6,7 @@
     "mapbox",
     "gl"
   ],
-  "main": "lib/mapbox-gl-native.node",
+  "main": "platform/node/index.js",
   "repository": {
     "type": "git",
     "url": "git://github.com/mapbox/mapbox-gl-native.git"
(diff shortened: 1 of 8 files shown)


jfirebaugh:
To bulletproof this for production, we should probably tweak the shim so that the exported `Map.prototype` has all the same properties as before, in case somebody is using  `Map.prototype.load.call(map, ...)` or something like that.

I haven't actually tested it, but the following might work:

```
var Map = mbgl.Map;
var originalConstructor = Map.prototype.constructor;
Map.prototype.constructor = function(options) {
    ...
}
```

mikemorris:
Rebased this and updated the shim to export the proper prototype, added tests (which were passing on `master`, failing here before but fixed now) to confirm expected behavior.

PR #5706, opened 2016-07-15
[core] Add support for $id key to filters

jfirebaugh:
https://github.com/mapbox/mapbox-gl-style-spec/issues/391


1ec5:
:+1:

1ec5:
Actually, the Darwin tests need to be updated to construct the feature using an argument list too.

jfirebaugh:
⇢ #5708.

Issue #4106, opened 2016-02-24
Support an OfflineRegionDefinition defined by an arbitrary shape


zugaldia:
At this moment we only support an `OfflineTilePyramidRegionDefinition` defined by a bounding box. We should let developers use arbitrary shapes to match, for example, the shape of a state, province, etc.

/cc: @jfirebaugh @1ec5

jfirebaugh:
The pieces that we would need to support this are:

* A set of C++ types representing geographic features. Probably strongly GeoJSON influenced so we could reuse these types for `featuresAt` results and `geoJSONLayer.setData()`. Could potentially be extracted from geojsonvt-cpp. (cc @springmeyer)
* For each platform, bindings to the above.
* An implementation of `tileCover` for the above.

twbell:
Polygons are P1; let's also consider polyline-as-input in this ticket as a P2.  The primary use case for the latter is pre-caching tiles along a route.

1ec5:
#5707 and #8012 take slightly different approaches to enabling the developer to define an arbitrary shape for offline download.

1ec5:
#11447 adds mbgl and Android support for offline regions that cover non-pyramidal geometries.

1ec5:
Fixed in #11447. Thanks @ivovandongen!

PR #5707, opened 2016-07-15
add polygonal offline region support

rosecodym:
Added here is the capability to calculate tile coverage of arbitrary polygonal areas in addition to the basic bounding boxes that were already supported, as well as an additional offline region definition to support polygonal regions. (See #4106) The offline region definition hierarchy has been fiddled with a bit to enable this, and should be easily expandable to future region types, such as ones generated from linear features.


jfirebaugh:
Replaced by #11267 and followups to come.

PR #5708, opened 2016-07-16
[core] Add support for $id key to filters

jfirebaugh:
https://github.com/mapbox/mapbox-gl-style-spec/issues/391

Reroll of #5706 because Bitrise doesn't like branches named `$id`.


kkaefer:
Can we add a test with a feature that has an ID, as well as a tag called `id` and verify that it filters against the right one?

PR #5710, opened 2016-07-16
Restore shape annotation z-order

### LINKED ISSUE #5691: Shape annotation z-order is reversed
We have a proof-of-concept application that draws an arrow on a route line at the location of each maneuver. Because the arrow’s stem must follow the path of the route even as it twists and winds, the arrow is implemented as a couple of polylines instead of a point annotation or annotation view. Because a polyline can only be stroked, not outlined, we’ve implemented the arrow as two similar pairs of polylines, a thinner, white pair of polylines atop a thicker, blue pair.

[Here’s a patch to iosapp and macosapp](https://gist.github.com/1ec5/fcd98632654a75ccc52999f1dd3e04c3) that draws the arrow the way this proof-of-concept application does. This is what it’s supposed to look like, seen here in aa1a54c577a95082824f2a5a6bdf4948506fcaa9 (ignore the fact that it’s backwards):

aa1a54c

8985b1311b7d00cf761752bd9291566325ae207a introduced a regression that incorrectly closed each polyline:

cfd6757

0fba70d5a8465499b0dce900e5aa74f7189e4594 introduced another regression that reversed the z-order of the polylines:

0fba70d

As of 44406e91de97132e18494fb1d3353ce00726f192, the polyline-closing bug has been fixed, but the z-order bug remains:

master

In some cases, we’ve seen mixed behavior:

screen shot 2016-07-11 at 6 59 18 pm

What should happen is that each added annotation is drawn above the annotation that was added before it.

/ref #5200
/cc @jfirebaugh

1ec5:
Shapes are once again always added to the top z-index.

fixed

Fixes #5691.

* [ ] Cherry-pick to release-ios-v3.3.0 branch


test/api/annotations.cpp
```diff
@@ -79,6 +79,24 @@ TEST(Annotations, FillAnnotation) {
...
+    Polygon<double> polygon = {{ {{ { 0, 0 }, { 0, 45 }, { 45, 45 }, { 45, 0 } }} }};
+    FillAnnotation underlaidAnnotation { polygon };
+    underlaidAnnotation.color = { { 0, 255, 0, 1 } };
+    FillAnnotation overlaidAnnotation { polygon };
+    overlaidAnnotation.color = { { 255, 0, 0, 1 } };
+
+    test.map.setStyleJSON(util::read_file("test/fixtures/api/empty.json"));
+    test.map.addAnnotation(underlaidAnnotation);
+    test.map.addAnnotation(overlaidAnnotation);
+    test.checkRendering("fill_annotation");
```
jfirebaugh: `"fill_annotation"` here ends up being the base name of the file used for comparison / written to disk if the test fails. Let's use a different name than the prior test.

test/api/annotations.cpp
```diff
@@ -79,6 +79,24 @@ TEST(Annotations, FillAnnotation) {
...
+
+    test.map.setZoom(test.map.getMaxZoom());
```
jfirebaugh: I think we can remove this block. It's a good check for the basic fill annotation test above but not necessary here.

test/api/annotations.cpp
1ec5: 👍 I copied the fixture to decouple the two tests.

test/api/annotations.cpp
1ec5: 👍 Removed.

jfirebaugh:
👍 

Whether the test fails if the `annotation_manager.hpp` change is reverted depends on implementation details of `unordered_map`, right? To be safe, can you also add a comment in `annotation_manager.hpp` explaining why `unordered_map` isn't used?

Then 🚢  on 🍏 .

1ec5:
Cherry-picked into the release-ios-v3.3.0 branch as 8179f2f6b2c185c16745c6a3faf4d4bd9683ef16.

PR #5724, opened 2016-07-19
GeoJSON point clustering

### LINKED ISSUE #320: point clustering
Once we have #170 GeoJSON with arbitrary points complete, then we need to be able to cluster overlay points for parity with existing mobile APIs.

mourner:
Closes #320, a work in progress. cc @jfirebaugh 

- [x] add `Supercluster` type to `urlOrGeoJSON` variant
- [x] instantiate Supercluster index in `GeoJSONSource::Impl` `setGeoJSON` if `cluster` is `true` in options and use it in `createTile`
- [x] ~~see if we can avoid adding `geojson` to include paths and public headers~~
- [x] bump supercluster.hpp and kdbush.hpp to fix compiler warnings in iOS
- [x] fix cluster wrapping near the date line
- [x] make sure all platforms are building
- [x] clean up the code
- [x] un-ignore clustering test in render suite and make sure it passes
- [x] review by @jfirebaugh 
- [x] rebase and add `[core]` prefix to commits
- [x] rejoice!


mourner:
OK, clustering now seems to be working!

include/mbgl/util/geojson.hpp
```diff
@@ -1,22 +0,0 @@
...
-
-namespace mapbox {
-namespace geojsonvt {
-class GeoJSONVT;
-} // namespace geojsonvt
-} // namespace mapbox
-
-namespace mbgl {
-
-class GeoJSON {
```
jfirebaugh: Let's keep this header file and keep using the `GeoJSON` symbol, but change the definition to:

```
namespace mbgl {

using GeoJSON = mapbox::geojson::geojson;

} // namespace mbgl
```

This follows a convention established by `mbgl::Geometry`, `mbgl::Feature`, etc. It keeps the naming conventions within the project more consistent, and allows for definition changes (like this one) to happen with minimal changes to downstream code.

include/mbgl/style/sources/geojson_source.hpp
```diff
@@ -1,17 +1,46 @@
...
 class GeoJSONSource : public Source {
 public:
-    GeoJSONSource(const std::string& id);
+    GeoJSONSource(const std::string& id, const GeoJSONOptions& options_ = GeoJSONOptions());
 
     void setURL(const std::string& url);
-    void setGeoJSON(GeoJSON&&);
+    void setGeoJSON(const mapbox::geojson::geojson&);
+
+    std::string getURL();
```
jfirebaugh: Since a GeoJSON source might not have a URL, let's make this `optional<std::string> getURL()` so that downstream code has to explicitly handle the non-present case. Conveys the intent better than a possibly-empty string.

mourner: Currently it's only called when `loaded` is `false`, in which case there's always a URL, so the empty string code path is never taken actually. Although it's probably right to make it bulletproof.

include/mbgl/util/geojson.hpp
mourner: Sure, will do.

include/mbgl/style/sources/geojson_source.hpp
1ec5: In that case, the getter should assert that the URL isn’t empty.

mourner:
The tests passed, woohoo! Now what's left is rebase + cleanup + merge.

platform/ios/ios.xcodeproj/project.pbxproj
```diff
@@ -1705,6 +1705,7 @@
 					"$(OTHER_CFLAGS)",
 					"$(variant_cflags)",
 					"$(geometry_cflags)",
+					"$(geojson_cflags)",
```
1ec5: Should there be corresponding changes in macos.xcodeproj?

mourner: Maybe — let me add them. I wonder why it's not caught by the builds though.

leekuo:
Any idea when this will make it into the iOS SDK release?

1ec5:
@leekuo, this functionality is now in [ios-v3.4.0-alpha.1](https://github.com/mapbox/mapbox-gl-native/releases/tag/ios-v3.4.0-alpha.1). Please note the caveats in https://github.com/mapbox/mapbox-gl-native/issues/320#issuecomment-235799557.

Issue #5626, opened 2016-07-08
Runtime styling API for iOS/macOS


1ec5:
#837 tracks the overall work to bring a runtime styling API to all the Mapbox GL Native libraries. This ticket tracks the work to design and implement the public interface in the iOS and macOS SDKs. (We’re planning similar APIs for Android in #5610 and for Qt in #5642.)

Conceptually, the runtime styling API brings all the settings found in Mapbox Studio into the SDK, allowing the developer to manipulate the style and its sources dynamically at runtime rather than at design time, without having to switch styles. The runtime styling API consists of:

* Classes to represent different kinds of sources
* Methods for adding and removing sources from a map’s style, including adding and removing raw data
* Classes to represent different kinds of layers
* Methods for adding and removing layers
* Methods for modifying a layer’s feature filter
* Methods for modifying a layer’s layout properties
* Methods for modifying a layer’s paint properties

## Design goals

Here are some proposed goals for the Cocoa API’s design and possible approaches to meeting those design goals:

* Use standard Foundation types wherever possible:
  * The developer can set a paint property to an NSString, NSNumber, or NSArray as appropriate. UIColor/NSColor would be an exception to the Foundation rule but easily typedef’d.
  * The developer can filter features [using an NSPredicate](https://gist.github.com/1ec5/3d81399e11ff45ee86d482cac25a0493).
* Balance type safety and dynamicism
* Minimize the surface area of MGLMapView:
  * MGLMapView’s sprawling API could get far larger if we require all style and layer operations to go through MGLMapView (`-setString:forPaintProperty:forLayerWithIdentifier:` or, worse, `-setFillOutlineColor:forPaintProperty:forLayerWithIdentifier:`). Instead, we should take the approach described in https://github.com/mapbox/mapbox-gl-native/issues/5610#issuecomment-231435943: MGLMapView vends instances of MGLSource and MGLStyleLayer that have backpointers to MGLMapView; you set properties on MGLSource and MGLStyleLayer rather than MGLMapView.

## Open questions

* How specific should property methods be?
[...]

Issue #5610, opened 2016-07-07
Android - Runtime styling Api


ivovandongen:
Now that the [Runtime styling Api](https://github.com/mapbox/mapbox-gl-native/issues/837) is taking shape in core, we can get started on the android side of things.

Major parts:

- Add/remove/manipulate layers
   - Expose layers; Background/Fill/Line/Symbol/Raster/Circle
   - Control paint properties (incl visibility)
   - Control layout properties
   - Control filtering
- Add/remove sources

It's been [suggested](https://github.com/mapbox/mapbox-gl-native/issues/837#issuecomment-226627258) to generate the Java bindings from the [style specification](https://www.mapbox.com/mapbox-gl-style-spec/). We can use the same approach as taken for the node bindings, using [ejs templates](http://www.embeddedjs.com/) or go for a more Java central approach by using a gradle plugin that transforms json into java classes, like [jsonschema2pojo](https://github.com/joelittlejohn/jsonschema2pojo/tree/master/jsonschema2pojo-gradle-plugin). Depending on integration in the main build or the Android sub-build (gradle).

Because the types (and manipulation of the types) need to go through jni, it's best to make the api flat (no 1-n relations). All manipulation/querying goes through the `MapboxMap` object. For some of the types we can use subclasses or type IntDef's if the types have little specialisation.

**Possible api**

Layer Types (generated):

- `BackgroundLayer`
- `CircleLayer`
- `FillLayer`
- `LineLayer`
- `RasterLayer`
- `SymbolLayer`

Related types:

- `LayoutProperty` - Each layer supports a set of layout properties
- `PaintProperty` - Each layer supports a set of paint properties
- `Filter` - Each layer can have a [filter](https://www.mapbox.com/mapbox-gl-style-spec/#types-filter)

Layer methods on `MapboxMap`:

- `setFilter()`
- `addLayer()`
- `removeLayer()`
- `setPaintProperty()`

Source types:

- `GeoJsonSource`
- `RasterSource`
- `VectorSource`  

Source methods on `MapboxMap`

- `addSource()`
- `removeSource()`

ivovandongen:
Ok, so we have a decision to make as to the level of typing we want in this api. The core api is beautifully typed, which is great to work with. For the android side of things there are some considerations though:
- A completely typed api adds quite a bit of api surface
- Passing rich types over JNI is not what we want, performance wise. So it would be translated to simpler types before crossing the boundary, where it would again be converted to core types.

[...]

PR #5727, opened 2016-07-19
Runtime styling API for iOS/macOS

frederoni:
WIP #5626 

Runtime styling for iOS and macOS is taking shape. Here's an overview if what's done, what's left to do and what's TBD.

- [ ] Sources
    - [x] Vector
    - [x] Raster
    - [x] GeoJSON
    - [ ] TileJSON
- [ ] Layers
    - [x] Background
    - [x] Fill
    - [x] Line
    - [x] Symbol
    - [x] Raster
    - [x] Circle
    - [ ] Custom
- [ ] Types
    - [x] Color
    - [ ] Enum (missing function getter)
    - [x] String
    - [x] Boolean
    - [x] Number
    - [x] Array
    - [x] Function
    - [ ] Filter
- [ ] Documentation
    - [x] Paint and layout properties (parsed from style spec and not particularly idiomatic)
    - [ ] Classes and methods


platform/ios/app/MBXViewController.m
```diff
@@ -407,6 +413,12 @@ - (void)presentAnnotationWithCustomCallout
     [self.mapView showAnnotations:@[annotation] animated:YES];
 }
 
+- (void)testRuntimeStyling
+{
+    MGLFillStyleLayer *waterLayer = (MGLFillStyleLayer *)[self.mapView.style layerWithIdentifier:@"water"];
+    waterLayer.fillColor = [UIColor redColor];
```
friedbunny: Would be nice to make this a toggle.

friedbunny:
Is the convention that we’ll run `make style-code` and the various `make style-code-{platform}` manually when the spec changes?

/cc @jfirebaugh

frederoni:
@friedbunny Android is generating style code for each build [#5734-184](https://github.com/mapbox/mapbox-gl-native/pull/5734/files#diff-b67911656ef5d18c4ae36cb6741b7965R184). It's convenient but easier for bugs to slip through. Eventually, we could generate tests from the style spec.

 ios/macos styles are being generated with `make style-code-ios` but since that part of the implementation lives in `platform/darwin` I guess `style-code-darwin` would make more sense.

1ec5:
We should place generated files in their own directory.

Instead of (or in addition to) a make rule for generating the source files, how about a target and scheme in the iOS and macOS workspaces?

platform/ios/scripts/generate-style-code.js
```diff
@@ -0,0 +1,82 @@
...
+
+global.isObject = function (property) {
+    switch (property.type) {
+        case 'string':
+            return true;
+        case 'color':
+            return true;
+        case 'array':
+            return true;
+        case 'enum': // (enum will be converted to NSString hence isObject(enum) == true)
```
1ec5: If we’re going with the strongly typed design proposed in #5626, each enumeration type in the style specification would ideally be mapped to a C enum (`NS_ENUM`) rather than NSString.

platform/ios/scripts/generate-style-code.js
```diff
@@ -0,0 +1,82 @@
...
+            return 'NSString';
+        case 'enum':
+            return 'NSString';
+        case 'color':
+            return `MGLColor`;
+        case 'array':
+            if (property.length) {
+                return `NSMutableArray`;
+            } else {
+                return `CGVector`;
```
1ec5: NSMutableArray and CGVector are two _very_ different types. Can you explain the rationale here, perhaps in a comment?

platform/ios/scripts/generate-style-code.js
```diff
@@ -0,0 +1,82 @@
...
+            return false;
+    }
+}
+
+global.propertyType = function (property) {
+    switch (property.type) {
+        case 'boolean':
+            return 'BOOL';
+        case 'number':
+            return 'float';
```
1ec5: We may want `CGFloat` or `double` precision in some cases, but potentially `NSInteger` or `NSUInteger` behavior in other places. Can we choose an appropriate type automatically? If not, maybe NSNumber is a better choice.

platform/darwin/src/MGLStyleLayer.mm
```diff
@@ -0,0 +1,38 @@
...
+
+@interface MGLStyleLayer()
+@property (nonatomic, weak) MGLMapView *mapView;
+@property (nonatomic) mbgl::Map *mbglMap;
+@property (nonatomic) mbgl::style::Layer *layer;
+@end
+
+@implementation MGLStyleLayer
+
+mbgl::Color MGLColorFromColor(MGLColor *color)
```
1ec5: This function is confusingly named.

platform/darwin/src/MGLStyleLayer.mm
```diff
@@ -0,0 +1,38 @@
...
+{
+    if (!color)
+    {
+        return { 0, 0, 0, 0 };
+    }
+    CGFloat r, g, b, a;
+    [color getRed:&r green:&g blue:&b alpha:&a];
```
1ec5: This can crash on iOS 10 and macOS if the color belongs to a color space other than `NSCalibratedRGBColorSpace`. You’ll have to first convert to that color space using `-[MGLColor colorUsingColorSpaceName:]`. That conversion can fail in some cases, for instance in the case of a pattern color.

platform/darwin/src/MGLStyle_Private.hpp
```diff
@@ -0,0 +1,11 @@
+#import "MGLStyle.h"
+
+#import "MGLStyleLayer.h"
+#import "MGLFillStyleLayer.h"
+#import <mbgl/util/default_styles.hpp>
+#include <mbgl/mbgl.hpp>
+
+@interface MGLStyle (Private)
+@property (nonatomic, weak) MGLMapView *mapView;
+@property (nonatomic) mbgl::Map *mbglMap;
```
1ec5: Let’s avoid keeping pointer references to `mbgl::Map` outside of MGLMapView. MGLMapView should have a private method that returns the `mbgl::Map`, or perhaps MGLMapView should have private methods for each of the things you’d want to do with that object.

platform/darwin/src/MGLStyleLayer.mm
1ec5: Also, we already have this code as `MGLColorObjectFromUIColor()` and `MGLColorObjectFromNSColor()`. We should find a good place for this function so that MGLMapView can use it too.

platform/ios/scripts/generate-style-code.js
frederoni: CGFloat seems to suit every number I've seen so far (width, padding, spacing, angle, blur, size) Let's start with that and see if we find any edge cases we should handle differently.

platform/darwin/src/MGLStyle_Private.hpp
frederoni: Good point. What if we just expose the `Map::update(Update flags)` method in a private header? I don't know about sources but we should be covered in terms of layers and paint properties.

platform/ios/scripts/generate-style-code.js
frederoni: There is at least one case where the array represents an offset (iconOffset) which is why I added CGVector but the implementation is mixed up and not ready. Also, the array shouldn't be mutable.

1ec5: In that case, I think we should use `CGVector` for specific properties, based on the property name. The style specification uses its `Array` type for a variety of things that Cocoa conventionally represents with different types. For example:

* [`center`](https://www.mapbox.com/mapbox-gl-style-spec/#root-center) as `CLLocationCoordinate2D`
* [`fill-translate`](https://www.mapbox.com/mapbox-gl-style-spec/#paint-fill-translate), [`line-translate`](https://www.mapbox.com/mapbox-gl-style-spec/#paint-line-translate), [`icon-offset`](https://www.mapbox.com/mapbox-gl-style-spec/#layout-symbol-icon-offset), [`text-offset`](https://www.mapbox.com/mapbox-gl-style-spec/#layout-symbol-text-offset), [`icon-translate`](https://www.mapbox.com/mapbox-gl-style-spec/#paint-icon-translate), [`text-translate`](https://www.mapbox.com/mapbox-gl-style-spec/#paint-text-translate), [`circle-translate`](https://www.mapbox.com/mapbox-gl-style-spec/#paint-circle-translate) as `CGVector`
  * `text-offset` is a good example of why `CGVector` is preferable to `NSPoint`/`CGPoint`: its unit is ems rather than screen points
* [`line-dasharray`](https://www.mapbox.com/mapbox-gl-style-spec/#paint-line-dasharray) as `NSArray<NSNumber *> *` or `CGFloat *`
  * I’m wary of introducing C arrays into this API. If we use NSNumber, maybe we should use NSNumber for [all the other `Number`-typed properties](https://github.com/mapbox/mapbox-gl-native/pull/5727/files/2ab2ab222fc4b4aa5eaaeaba7c57f0c535accb34#r71622482) too.
* [`icon-text-fit-padding`](https://www.mapbox.com/mapbox-gl-style-spec/#layout-symbol-icon-text-fit-padding) as `NSEdgeInsets`/`UIEdgeInsets`
* [`text-font`](https://www.mapbox.com/mapbox-gl-style-spec/#layout-symbol-icon-offset) as `NSArray<NSString *> *` (or `NSArray<NSFontDescriptor *> *` / `NSArray<UIFontDescriptor *> *`?)

platform/darwin/src/MGLStyleLayer.h.ejs
```diff
@@ -0,0 +1,29 @@
...
+// This file is generated. Do not edit.
+
+#import <Mapbox/Mapbox.h>
+
+#import "MGLTypes.h"
+
+@interface MGL<%- camelize(type) %>StyleLayer : MGLStyleLayer
+
+<% if (layoutProperties.length) { -%>
+// Layout properties
```
1ec5: Nit: Use `#pragma mark Layout Properties` so jazzy can create a new section for the symbols below.

1ec5: Better yet, “Accessing the Layout Attributes”, since we follow the Apple documentation convention of labeling each section in the documentation with a task. Since “property” has a more generic meaning in Objective-C and Swift, “attribute” is usually used to describe the kind of property that affects presentation etc. We [use “attribute” in MGLFeature](https://github.com/mapbox/mapbox-gl-native/blob/03ebae0f2ed0e5ff2c4fe284d8d3db8960ae4938/platform/darwin/src/MGLFeature.h#L86), for example.

platform/darwin/src/MGLStyleLayer.mm
1ec5: One option would be to create a new header per platform, NSColor+MGLAdditions.h and UIColor+MGLAdditions.h, that defines a `-mbgl_color` method on NSColor and UIColor, respectively.

platform/ios/scripts/generate-style-code.js
1ec5: As a stretch goal, we could use the appropriate typedefs for certain properties:

* [`bearing`](https://www.mapbox.com/mapbox-gl-style-spec/#root-bearing), [`icon-rotate`](https://www.mapbox.com/mapbox-gl-style-spec/#layout-symbol-icon-rotate), [`text-rotate`](https://www.mapbox.com/mapbox-gl-style-spec/#layout-symbol-text-rotate) as `CLLocationDirection`
* [`duration`](https://www.mapbox.com/mapbox-gl-style-spec/#transition-duration), [`delay`](https://www.mapbox.com/mapbox-gl-style-spec/#transition-delay), [`raster-fade-duration`](https://www.mapbox.com/mapbox-gl-style-spec/#paint-raster-fade-duration) as `NSTimeInterval`

platform/darwin/src/MGLStyleLayer.h.ejs
```diff
@@ -0,0 +1,29 @@
+<%
+  const type = locals.type;
+  const layoutProperties = locals.layoutProperties;
+  const paintProperties = locals.paintProperties;
+-%>
+// This file is generated. Do not edit.
```
1ec5: We should strip out this comment when packaging the SDK. Also say what generates the file, or how to generate it.

platform/darwin/src/MGLStyleLayer.h.ejs
```diff
@@ -0,0 +1,29 @@
...
+
+#import "MGLTypes.h"
+
+@interface MGL<%- camelize(type) %>StyleLayer : MGLStyleLayer
+
+<% if (layoutProperties.length) { -%>
+// Layout properties
+
+<% for (const property of layoutProperties) { -%>
+@property (nonatomic) <%- propertyType(property) %> <% if (isObject(property)) { -%>*<% } -%><%- camelizeWithLeadingLowercase(property.name) %>;
```
1ec5: We should get some documentation in here. 😉 Every property in the specification has a `doc` property alongside the `type` property.

platform/darwin/src/MGLStyleLayer.h.ejs
```diff
@@ -0,0 +1,29 @@
...
+
+<% } -%>
+<% } -%>
+// Paint properties
+
+<% for (const property of paintProperties) { -%>
```
1ec5: Putting paint properties at the top level of this object means we can only support default paint properties. Given that the iOS SDK [supports style classes](https://github.com/mapbox/mapbox-gl-native/blob/03ebae0f2ed0e5ff2c4fe284d8d3db8960ae4938/platform/ios/src/MGLMapView.h#L214) (and the macOS SDK could trivially do so as well), do we also need to support [class-specific paint properties](https://www.mapbox.com/mapbox-gl-style-spec/#layer-paint.*)?

/cc @jfirebaugh

platform/darwin/src/MGLSymbolStyleLayer.h
```diff
@@ -20,55 +72,55 @@
 
 @property (nonatomic) BOOL iconOptional;
```
1ec5: Should we add a custom getter, so `-isIconOptional`? Or would it be too difficult to ensure that all the getters remain grammatically correct?

frederoni: I think we should give it a try at least, if we decide not to go with `NSNumber`

platform/darwin/src/MGLCircleStyleLayer.h
```diff
@@ -15,20 +16,41 @@ typedef NS_ENUM(NSUInteger, MGLCircleStyleLayerCirclePitchScale) {
...
+*/
 @property (nonatomic) NSArray *circleTranslate;
 
+/**
+  Control whether the translation is relative to the map (north) or viewport (screen)
+*/
 @property (nonatomic) MGLCircleStyleLayerCircleTranslateAnchor circleTranslateAnchor;
 
+/**
+  Controls the scaling behavior of the circle when the map is pitched. The value `map` scales circles according to their apparent distance to the camera. The value `viewport` results in no pitch-related scaling.
```
1ec5: We’re going to need a way to map `map` to `MGLCircleStyleLayerCirclePitchScaleMap` and so on.

Makefile
```diff
@@ -162,6 +162,9 @@ ifabric: $(IOS_PROJ_PATH)
 idocument:
 	OUTPUT=$(OUTPUT) ./platform/ios/scripts/document.sh
 
+style-code-ios:
+	node platform/ios/scripts/generate-style-code.js
```
1ec5: The script generates an API for both iOS and macOS. Let’s call it `style-code-darwin` and put the script in the platform/darwin/ directory.

platform/darwin/src/MGLFillStyleLayer.h
```diff
@@ -11,20 +12,41 @@ typedef NS_ENUM(NSUInteger, MGLFillStyleLayerFillTranslateAnchor) {
...
-// Paint properties
+#pragma mark - Accessing the Paint Attributes
 
+/**
+  Whether or not the fill should be antialiased.
+*/
 @property (nonatomic) BOOL fillAntialias;
 
+/**
+  The opacity of the entire fill layer. In contrast to the fill-color, this value will also affect the 1px stroke around the fill, if the stroke is used.
```
1ec5: We need to avoid saying “pixel” when we mean “point”, since a point is 2–3 pixels on a Retina display. This is one area where Apple platforms differ from the other Mapbox GL platforms.

platform/darwin/src/MGLFillStyleLayer.h
```diff
@@ -11,20 +12,41 @@ typedef NS_ENUM(NSUInteger, MGLFillStyleLayerFillTranslateAnchor) {
...
+*/
 @property (nonatomic) CGFloat fillOpacity;
 
+/**
+  The color of the filled part of this layer. This color can be specified as rgba with an alpha component and the color's opacity will not affect the opacity of the 1px stroke, if it is used.
+*/
 @property (nonatomic) MGLColor *fillColor;
 
+/**
+  The outline color of the fill. Matches the value of `fill-color` if unspecified.
```
1ec5: We need a way to map `fill-color` to `fillColor`.

platform/darwin/src/MGLFillStyleLayer.mm
```diff
@@ -0,0 +1,25 @@
+#import "MGLFillStyleLayer.h"
+
+#include "MGLStyleLayer_Private.hpp"
```
1ec5: So far, we’ve named Objective-C++ headers with .h extensions, because .hpp implies C++. I can see how using .hpp would be useful, to avoid accidentally making C++-infected code part of our public API. But that should be a separate change applied consistently to the whole codebase.

platform/darwin/src/MGLLineStyleLayer.h
```diff
@@ -0,0 +1,100 @@
+// This file is generated. 
+// Edit platform/ios/scripts/generate-style-code.js, then run `make style-code-ios`.
+
+#import <Mapbox/Mapbox.h>
+
+#import "MGLTypes.h"
+
+typedef NS_ENUM(NSUInteger, MGLLineStyleLayerLineCap) {
```
1ec5: We’ll need documentation comments for enums and their values. Otherwise, they won’t show up in jazzy. We could either write the comments manually, as overrides, or we could have the comment simply say `@see -[MGL…Layer lineCap]`.

platform/darwin/src/MGLSymbolStyleLayer.h
```diff
@@ -0,0 +1,307 @@
...
+*/
+@property (nonatomic) CGFloat iconSize;
+
+/**
+  Position and scale an icon by the its corresponding text.
+*/
+@property (nonatomic) MGLSymbolStyleLayerIconTextFit iconTextFit;
+
+/**
+  Size of padding area around the text-fit size in clockwise order: top, right, bottom, left.
```
1ec5: If we do adopt `NSEdgeInsets`/`UIEdgeInsets` for properties such as this, we’ll need to rearrange the values from clockwise (top, right, bottom, left) to counterclockwise (top, left, bottom, right).

platform/darwin/src/MGLSymbolStyleLayer.h
```diff
@@ -0,0 +1,307 @@
...
+*/
+@property (nonatomic) CGFloat textPadding;
+
+/**
+  If true, the text may be flipped vertically to prevent it from being rendered upside-down.
+*/
+@property (nonatomic) BOOL textKeepUpright;
+
+/**
+  Specifies how to capitalize text, similar to the CSS `text-transform` property.
```
1ec5: It’s unfortunate that we’d be talking about CSS instead of NSString APIs.

platform/ios/src/Mapbox.h
```diff
@@ -32,6 +32,8 @@ FOUNDATION_EXPORT const unsigned char MapboxVersionString[];
 #import "MGLShape.h"
 #import "MGLShapeCollection.h"
 #import "MGLStyle.h"
+#import "MGLStyleLayer.h"
+#import "MGLFillStyleLayer.h"
```
1ec5: We also need to import all the generated headers into the umbrella headers.

1ec5:
`MGLColorObjectFromNSColor()` and `MGLColorObjectFromUIColor()` should be removed in favor of the new category methods.

jfirebaugh:
In addition to being constant values, style properties can be [functions](https://www.mapbox.com/mapbox-gl-style-spec/#types-function). In core, this is expressed via the [`PropertyValue` template](https://github.com/mapbox/mapbox-gl-native/blob/master/include/mbgl/style/property_value.hpp), which can represent an undefined, constant, or function value of underlying type `T`. We'll need an equivalent representation for the SDK bindings. How are templated or generic types typically handled in Objective C?

platform/darwin/src/MGLCircleStyleLayer.mm
```diff
@@ -0,0 +1,79 @@
...
+{
+    circleLayer->setCircleColor(circleColor.mbgl_color);
+}
+- (MGLColor *)circleColor
+{
+    return [MGLColor mbgl_color:circleLayer->getCircleColor().asConstant()];
+}
+- (void)setCircleBlur:(CGFloat)circleBlur
+{
+    circleLayer->setCircleBlur(circleBlur);
```
1ec5: There’s a lot of repetition in these implementation files. Even though it’s all generated, it could potentially make the generated code easier to reason about if we factor the boilerplate code out into macros, so that each pair of `CGFloat`-typed getters and setters is ultimately represented by:

```c
DEFINE_LAYER_PROP_NUMBER(circle, CircleBlur, CircleBlur)
```

These macros would be a little tricker to debug, but I’d expect them to be pretty straightforward callthroughs to other code anyways.

1ec5:
That’s a great point, @jfirebaugh. Unlike Swift, Objective-C doesn’t support generics, other than some “lightweight generics” for container types. I can think of a few options:

1. Change all property types to `id`, shifting all type checking to runtime. We’d have to use object types for everything, so enumerations would be inconveniently wrapped in NSValue objects, for example.
  1. Change all property types to `id <MGLStyleAttributeValue>` and extend NSNumber (for `Number` and `Bool`), NSValue (`Enum`), NSString (`String`), NSArray (`Array`), and MGLStyleAttributeFunction (`Function`) to conform to the MGLStyleAttributeValue property. This would restore some type safety: you can’t even attempt to set an attribute to a UIView, but you can attempt to set the `textTransform` property to a floating-point number. This approach is largely in line with the solution for https://github.com/mapbox/mapbox-gl-native/pull/5727#issuecomment-234290240, which we also need to solve.
  2. Change property types to NSValue and add category methods on NSValue for MGLStyleAttributeFunction. This would also restore some type safety in exchange for additional verbosity.
  3. Replace the individual properties with dictionary-style getters and setters, as proposed in #5626. This would have even less type safety – you could attempt to set an valid attribute – but developers might be more comfortable with verbosity from an API that looks vaguely like KVO.
2. Replace each property with two setters – one for a constant value and one for a function value – and a getter that could return either a constant or a function via an out parameter. This would be an awkward design, and I don’t know of any API that uses it as systematically as we’d have to.
3. Replace all the properties with a single setter and getter that takes an MGLStyleAttribute object, which could be generically typed in a manner similar to what the Android SDK is doing. This would be unfortunate for its verbosity and, to a lesser degree, memory management overhead.

Of these options, only 1(i) strikes me as a reasonably ergonomic solution.

1ec5:
Design-wise, there’s some prior art in cutting-room-floor/mapbox-gl-cocoa#27 and cutting-room-floor/mapbox-gl-cocoa#31. But I’m not too confident that we’ll be able to glean anything from it, because in cutting-room-floor/mapbox-gl-cocoa#31 we essentially hit the same wall we’re hitting now. Fortunately, the mbgl interface for runtime styling is a lot different than it was back then, so we at least have some ability to perform compile-time type checking, even if it isn’t as airtight as it would be in a language that supports generics.

platform/darwin/src/MGLStyleAttributeValue.h
```diff
@@ -0,0 +1,31 @@
+#import <Foundation/Foundation.h>
+
+#import "MGLTypes.h"
+
+typedef struct {
+    float red;
+    float green;
+    float blue;
+    float alpha;
+} MGLAttributeColor;
```
1ec5: Instead of defining our own color struct type, why not extend NSColor and UIColor to conform to MGLStyleAttributeValue? Then `-[MGLStyleAttributeValue colorValue]` would return an instance of NSColor or UIColor (depending on the platform, using an ifdef).

platform/macos/src/NSColor+MGLAdditions.hpp
```diff
@@ -0,0 +1,17 @@
...
+
+/**
+ Converts the color into an mbgl::Color in calibrated RGB space.
+ */
+- (mbgl::Color)mbgl_color;
+
+/**
+ Instantiates `NSColor` from an `mbgl::Color`
+ */
++ (NSColor *)mbgl_color:(mbgl::Color)color;
```
1ec5: This should be `+mbgl_colorWithColor:` so it’s clear that it initializes a new color object.

platform/darwin/src/MGLStyleAttributeValue.h
1ec5: Ah, I see that you’re already extending NSColor and UIColor that way. So `-[MGLStyleAttributeValue colorValue]` should return an NSColor or UIColor, and whatever calls `-colorValue` would use `-[NSColor(MGLAdditions) mbgl_color]` to convert to an `mbgl::Color` for use with core code.

incanus:
Apologies if I missed past discussed here, but I noticed in testing that we need to call this after map style changes: 

```objc
self.mapView.mbglMap->update(mbgl::Update::RecalculateStyle | mbgl::Update::Classes);
```

That's not the case with @ivovandongen's Android API over in #5610. 

Can we avoid that?

frederoni:
@incanus That's definitely up for discussion. I guess it all depends on how you gonna use the API but if you intend to update 100 layers and 100 paint/layout properties, that'd be a lot of update calls to `mbglMap` so we're considering a transaction model or at least a way to disable automatic updating of styles and classes.

incanus:
Yeah, I like the idea of a transaction model. Can we bound it a là `CATransaction` or `UIView` animation commit blocks though? Even if we don't allow nesting and/or don't _need_ an opener? 

BTW perf on this at least as far as the sample app three-layer change looks ok.

platform/ios/src/UIColor+MGLAdditions.mm
```diff
@@ -0,0 +1,38 @@
+#import "UIColor+MGLAdditions.hpp"
+
+@implementation UIColor (MGLAdditions)
+
+- (mbgl::Color)mbgl_color
+{
+    if (!self)
```
1ec5: If `self` is `nil`, this method won’t execute.

platform/ios/src/UIColor+MGLAdditions.mm
```diff
@@ -0,0 +1,38 @@
...
+    mbgl::Color color = self.mbgl_color;
+    return {{ color.r, color.g, color.b, color.a }};
+}
+
++ (UIColor *)mbgl_color:(mbgl::Color)color
+{
+    return [UIColor colorWithRed:color.r green:color.g blue:color.b alpha:color.a];
+}
+
++ (UIColor *)mbgl_propertyValue:(mbgl::style::PropertyValue<mbgl::Color>)color
```
1ec5: Nit: `+mbgl_colorWithPropertyValue:` or `+colorWithMBGLPropertyValue:`.

platform/darwin/src/NSNumber+MGLStyleAttributeAdditions_Private.hpp
```diff
@@ -0,0 +1,15 @@
+#import <Foundation/Foundation.h>
+
+#import "MGLStyleAttributeValue.h"
+
+#include <mbgl/style/property_value.hpp>
+
+@interface NSNumber (MGLStyleAttributeAdditions_Private) <MGLStyleAttributeValue>
```
1ec5: This inheritance strategy looks sound. It’s consistent with our strategy with MGLOfflineRegion and MGLTilePyramidOfflineRegion: MGLTilePyramidOfflineRegion publicly conforms to MGLOfflineRegion and privately conforms to MGLOfflineRegion_Private (which also conforms to MGLOfflineRegion). MGLOfflineStorage imports MGLOfflineRegion_Private.h, so it has access to, for instance, `-initWithOfflineRegionDefinition:` (although it still needs to call `-conformsToProtocol:` first).

platform/darwin/src/NSNumber+MGLStyleAttributeAdditions_Private.hpp
```diff
@@ -0,0 +1,15 @@
...
+
+- (mbgl::style::PropertyValue<bool>)mbgl_booleanPropertyValue;
+
++ (NSNumber *)mbgl_booleanWithPropertyValue:(mbgl::style::PropertyValue<bool>)propertyValue;
```
1ec5: This is fine, but if you ever find it tedious to support multiple `-mbgl_…withPropertyValue:` methods, it might be simpler to have a single `-mbgl_numberWithPropertyValue:` method that takes an `mbgl::style::PropertyValue<MGLNSNumberConvertible>`? `MGLNSNumberConvertible` would be typedef’d to `variant<bool, float, uint32_t, int32_t>`.

platform/darwin/scripts/generate-style-code.js
```diff
@@ -0,0 +1,133 @@
...
+
+global.getterImplementation = function(property, layerType = null) {
+    switch (property.type) {
+        case 'boolean':
+            return `return [NSNumber mbgl_booleanWithPropertyValue:${layerType}Layer->get${camelize(property.name)}()];`;
+        case 'number':
+            return `return [NSNumber mbgl_numberWithPropertyValue:${layerType}Layer->get${camelize(property.name)}()];`;
+        case 'string':
+            //return `return [NSNumber mbgl_numberWithPropertyValue:${layerType}Layer->get${camelize(property.name)}()];`;
+            return `return @"";`;
```
1ec5:
```objc
return @(${layerType}Layer->get${camelize(property.name)}().c_str());
```

platform/darwin/scripts/generate-style-code.js
```diff
@@ -0,0 +1,133 @@
...
+        case 'enum':
+            //return `return [[MGLStyleAttribute alloc] init]; //return ${prefix}${camelize(layerType)}${suffix}${camelize(property.name)}${camelize(property.values[0])};`
+            return `return @0;`;
```
1ec5:
```objc
auto rawValue = ${layerType}Layer->get${camelize(property.name)}();
const char *type = ${prefix}${camelize(layerType)}${suffix}${camelize(property.name)};
return [NSValue value:&rawValue withObjCType:@encode(type)];
```

1ec5:
I just rebased onto master in order to resolve conflicts and remove references to #5626 in commit messages that were polluting that issue page on GitHub. If you want to return to the state of the branch pre-rebase but keep the commits I made tonight, you can force-push 06f6d8689fb406f2ada9ee34f9829468be188f17.

platform/darwin/src/MGLStyle.h
```diff
@@ -158,7 +164,45 @@ static const NSInteger MGLStyleDefaultVersion = 9;
  */
 + (NSURL *)satelliteStreetsStyleURLWithVersion:(NSInteger)version;
 
-- (instancetype)init NS_UNAVAILABLE;
+#pragma mark Runtime Styling
+
+/**
+ Returns a layer that conforms to MGLStyleLayer if any layer with the given identifier was found.
```
friedbunny: We’ve recently settled on wrapping these lines at 80 characters. Xcode has a convenient “Page guide at column” setting at in Text Settings.

friedbunny: Style-wise, MGLStyleLayer here should also have the code ticks, i.e.:

```
`MGLStyleLayer`
```

platform/darwin/src/MGLStyle.h
```diff
@@ -158,7 +164,45 @@ static const NSInteger MGLStyleDefaultVersion = 9;
...
+ Adds a new source to the map view.
+ 
+ @param source The source to add to the map view.
+ */
+- (void)addSource:(MGLSource *)source;
+
+/**
+ Removes a source from the map view.
+ 
+ @param source The source to remove
```
friedbunny: Nit: needs a period at the end of this sentence.

platform/darwin/src/MGLStyleAttributeValue_Private.hpp
```diff
@@ -17,4 +19,7 @@
 - (mbgl::style::PropertyValue<std::vector<float> >)mbgl_numberArrayPropertyValue;
 - (mbgl::style::PropertyValue<uint8_t>)mbgl_enumPropertyValue;
 
+// Convert mbgl types to darwin types
+- (id <MGLStyleAttributeValue>)mbgl_colorPropertyValueWith:(mbgl::style::PropertyValue<mbgl::Color>)color;
```
1ec5: `-mbgl_colorPropertyValueWith:` doesn’t conform to Cocoa naming conventions. How about `-mbgl_colorPropertyValueWithColor:`?

platform/darwin/src/MGLBaseStyleLayer.h
```diff
@@ -0,0 +1,26 @@
+#import <Foundation/Foundation.h>
+
+@interface MGLBaseStyleLayer : NSObject
```
1ec5: This interface declaration needs a documentation comment.

platform/darwin/src/MGLBaseStyleLayer.h
```diff
@@ -0,0 +1,26 @@
...
+
+@property (nonatomic, assign, getter=isVisible) BOOL visible;
+
+/**
+ The maximum zoom level on which the layer gets parsed and appears on.
+ */
+@property (nonatomic, assign) float maxZoom;
```
1ec5: “max” should always be written out as “maximum” in API.

platform/darwin/src/MGLBaseStyleLayer.h
```diff
@@ -0,0 +1,26 @@
...
+
+/**
+ The minimum zoom level on which the layer gets parsed and appears on.
+ */
+@property (nonatomic, assign) float minZoom;
```
1ec5: “min” should always be written out as “minimum” in API.

platform/darwin/src/MGLBaseStyleLayer.h
```diff
@@ -0,0 +1,26 @@
...
+
+/**
+ Updates the layers layout and paint properties.
```
friedbunny: Grammar cop: layers → layer’s

platform/darwin/src/MGLBaseStyleLayer.h
```diff
@@ -0,0 +1,26 @@
...
+ 
+ @discussion The properties are recalculated upon managing sources and
+ layers but in order to update a layout or paint property you have to call
+ this method manually.
+ */
+- (void)update;
```
1ec5: As tail work, let’s consider a way to update immediately by default but offer a block-based batch updating mechanism for use cases that involve changing multiple properties in tandem.

platform/darwin/src/MGLBaseStyleLayer.h
```diff
@@ -0,0 +1,26 @@
...
+
+/**
+ The minimum zoom level on which the layer gets parsed and appears on.
+ */
+@property (nonatomic, assign) float minZoom;
+
+/**
+ Updates the layers layout and paint properties.
+ 
+ @discussion The properties are recalculated upon managing sources and
```
friedbunny: Hrm, I’m not sure what “managing sources” means in this context — is there a more clear/specific way to write this?

platform/darwin/src/MGLStyle.h
```diff
@@ -32,6 +36,8 @@ static const NSInteger MGLStyleDefaultVersion = 9;
  */
 @interface MGLStyle : NSObject
 
+#pragma mark Default Style URL's
```
friedbunny: Nit: please use smart quotes: URL's → URL’s

platform/darwin/src/MGLStyle.h
```diff
@@ -162,22 +164,46 @@ static const NSInteger MGLStyleDefaultVersion = 9;
...
-
-#pragma mark - Runtime style API
+#pragma mark Runtime Styling
 
-- (id <MGLStyleLayer>)layerWithIdentifier:(NSString *)identifier;
+/**
+ Returns a layer that conforms to MGLStyleLayer if any layer with the given identifier was found.
+ 
+ @return layer A layer of type `MGLBackgroundStyleLayer`, `MGLCircleStyleLayer`,
+ `MGLFillStyleLayer`, `MGLLineStyleLayer`, `MGLRasterStyleLayer` or `MGLSymbolStyleLayer`.
```
1ec5: I don’t think we need a full list of classes here.

platform/darwin/src/MGLStyle.h
```diff
@@ -162,22 +164,46 @@ static const NSInteger MGLStyleDefaultVersion = 9;
...
+ */
+- (nullable id <MGLStyleLayer>)layerWithIdentifier:(NSString *)identifier;
 
+/**
+ Adds a new layer to the map view.
```
1ec5: Does the new layer appear on top of the existing layers, or below it?

platform/darwin/src/MGLBaseStyleLayer.h
1ec5: The `@discussion` keyword is also unnecessary. Anything after the first blank line is treated as discussion by both Quick Help and jazzy.

platform/macos/src/NSColor+MGLAdditions.hpp
```diff
@@ -0,0 +1,20 @@
...
+
+@interface NSColor (MGLAdditions)
+
+/**
+ Converts the color into an mbgl::Color in calibrated RGB space.
+ */
+- (mbgl::Color)mbgl_color;
+
+/**
+ Instantiates `NSColor` from an `mbgl::Color`
```
friedbunny: Nit: period at the end of this sentence.

platform/darwin/src/MGLSource.h
```diff
@@ -0,0 +1,10 @@
+#import <Foundation/Foundation.h>
+
+@interface MGLSource : NSObject
+
```
friedbunny: Should we document this interface (and its implementations, such as MGLGeoJSONSource)?

1ec5:
Don’t forget to add MGLStyleFilter.h/.m to macos.xcodeproj too.

platform/darwin/src/MGLRuntimeStylingTests.m.ejs
```diff
@@ -0,0 +1,39 @@
...
+
+#import "MGLStyleLayerTests.h"
+
+@interface MGL<%- camelize(type) %>LayerTests : MGLStyleLayerTests
+@end
+
+@implementation MGL<%- camelize(type) %>LayerTests
+
+- (void)test<%- camelize(type) %>Layer {
+    NSURL *geoJSONURL = [NSURL URLWithString:@"https://dl.dropboxusercontent.com/u/5285447/amsterdam.geojson"];
```
friedbunny: We should check this fixture in to this repo and then use the GitHub raw URL, if we need to fetch it remotely.

platform/darwin/src/MGLStyle.h
friedbunny: If we wanted to include a list, we could use `@see`. Something like this:

```
 @see `MGLStyle`, `MGLMapView.styleURL`, `delegate`
```

Ends up looking like this in jazzy:

screen shot 2016-08-10 at 3 05 51 pm

platform/darwin/src/MGLStyleLayer.h.ejs
```diff
@@ -0,0 +1,57 @@
+<%
+  const type = locals.type;
+  const layoutProperties = locals.layoutProperties;
+  const paintProperties = locals.paintProperties;
+-%>
+// This file is generated. 
+// Edit platform/darwin/scripts/generate-style-code.js, then run `make style-code-darwin`.
+
+#import <Mapbox/Mapbox.h>
```
friedbunny: Importing Mapbox.h isn’t necessary and it breaks jazzy doc generation.

>mapbox-gl-native/platform/darwin/src/MGLLineStyleLayer.h:4:9: fatal error: 'Mapbox/Mapbox.h' file not found

platform/darwin/src/MGLStyle.h
friedbunny: Whoops, I didn’t really read this line — there should be no apostrophe at all.

platform/darwin/src/MGLStyle.h
friedbunny: This is how it looks it jazzy now:
screen shot 2016-08-10 at 3 23 03 pm

`@return` does not take a parameter — it should be prose.

friedbunny:
`MGLStyleLayer` and friends aren’t being included in the jazzy docs:

screen shot 2016-08-10 at 3 33 54 pm

platform/darwin/src/MGLRasterSource.h
```diff
@@ -0,0 +1,10 @@
+#import <Mapbox/Mapbox.h>
```
friedbunny: This header should import MGLSource.h, not Mapbox.h.

friedbunny: Looks like removing Mapbox.h also requires either UIKit or CoreGraphics to be added.

```
#import "MGLSource.h"
#import <CoreGraphics/CoreGraphics.h>
```

platform/darwin/src/MGLStyleAttributeValue_Private.hpp
frederoni: I think we're both off track 😄 Since objc doesn't support generics, naming a method with a generic type param is very confusing. How about `-mbgl_colorWithPropertyValue:` which theoretically should be `-mbgl_colorWithPropertyValueColor:` but that seems excessive. 
same goes for all methods in [MGLStyleAttribute.hpp](https://github.com/mapbox/mapbox-gl-native/blob/5626-runtime-styling/platform/darwin/src/MGLStyleAttribute.hpp)

1ec5: Oh, right, `-mbgl_colorWithPropertyValueColor:` makes sense. I got confused about the return type because of the existing name. 😄 Also, if you ever find the lack of generics to be limiting, we can also break out a standalone, templated C++ function and call that instead of the Objective-C category method.

platform/darwin/src/MGLRuntimeStylingTests.m.ejs
1ec5: This test would be more robust if we check the GeoJSON fixture into this repository, add it to the test bundle target as a resource, and read it in from a local file URL.

1ec5: On second thought, that would require mbgl to recognize the local file URL. Maybe we should [support asset: URLs](https://github.com/mapbox/mapbox-gl-native/blob/c354ae523a00b4c0e0577c4784b53553ea1a1cb7/platform/ios/src/MGLMapView.mm#L355-L359) here?

platform/darwin/src/MGLStyleLayer.h.ejs
```diff
@@ -0,0 +1,57 @@
...
+
+#import "MGLTypes.h"
+#import "MGLStyleAttributeValue.h"
+#import "MGLBaseStyleLayer.h"
+
+<% for (const property of layoutProperties) { -%>
```
1ec5: The generated headers should be audited for nullability.

platform/darwin/src/MGLBaseStyleLayer.h
frederoni: I think this method should be moved to `MGLStyle`. You call `-update` on the layer but it will actually recalculate styles and update classes for all layers.

platform/darwin/src/MGLStyleLayer.h.ejs
1ec5: → #5945

platform/ios/scripts/generate-style-code.js
1ec5: → #5947

platform/ios/scripts/generate-style-code.js
1ec5: We ended up using NSNumber, so this is no longer applicable. However, if in the future we introduce any primitive-typed number, we should keep these types in mind.

platform/darwin/src/MGLStyleLayer.h.ejs
1ec5: → #5948

platform/darwin/src/MGLSymbolStyleLayer.h
1ec5: We went with NSNumber, so it makes less sense to customize the getter.

platform/darwin/src/MGLCircleStyleLayer.h
1ec5: → #5949

platform/darwin/src/MGLFillStyleLayer.h
1ec5: → #5950

platform/darwin/src/MGLFillStyleLayer.h
1ec5: → #5951

platform/darwin/src/MGLFillStyleLayer.mm
1ec5: We need to go back to .h: #5952.

platform/darwin/src/MGLLineStyleLayer.h
1ec5: → #5953

platform/darwin/src/MGLSymbolStyleLayer.h
1ec5: https://github.com/mapbox/mapbox-gl-native/issues/5947#issuecomment-239262950

platform/ios/src/Mapbox.h
1ec5: This happened, but I think we can automate this step: #5955.

platform/darwin/src/MGLBaseStyleLayer.h
1ec5: → #5957

platform/darwin/src/MGLStyleAttributeValue_Private.hpp
1ec5: → #5958

platform/darwin/src/MGLBaseStyleLayer.h
1ec5: → #5959

platform/darwin/src/MGLSource.h
1ec5: → #5960

platform/darwin/src/MGLStyleLayer.h.ejs
1ec5: → #5945

1ec5:
Filed lots of tail work for what’s been implemented so far: #5943 #5947 #5948 #5949 #5950 #5951 #5952 #5953 #5954 #5955 #5956 #5957 #5958 #5959 #5960 #5961.

#5626 remains open for the remaining unimplemented portions of the runtime styling API.

platform/darwin/src/MGLRuntimeStylingTests.m.ejs
1ec5: This happened in #6029.

PR #5752, opened 2016-07-21
[ios] Reformat inline documentation with ticks, smart quotes, breaks

friedbunny:
- Makes the use of apostrophes in the iOS SDK documentation consistent by converting “dumb” `'` into smart `’`.
- Adds ticks around class/property names/values.
- Breaks lines at 80 characters.

This isn’t strictly necessary, is probably rather pedantic, and it may mildly disturb a couple of @frederoni’s inflight PR’s, but let’s yank the bandaid off and do this all at once.

/cc @1ec5 @boundsj @frederoni @incanus

### DIFF
--- a/platform/ios/src/MGLCalloutView.h
+++ b/platform/ios/src/MGLCalloutView.h
@@ -8,32 +8,38 @@ NS_ASSUME_NONNULL_BEGIN
 @protocol MGLAnnotation;
 
 /**
- A protocol for a `UIView` subclass that displays information about a selected annotation near that annotation.
+ A protocol for a `UIView` subclass that displays information about a selected
+ annotation near that annotation.
  */
 @protocol MGLCalloutView <NSObject>
 
 /**
- An object conforming to the `MGLAnnotation` protocol whose details this callout view displays.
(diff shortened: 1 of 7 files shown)


platform/ios/src/MGLCalloutView.h
```diff
@@ -8,32 +8,38 @@ NS_ASSUME_NONNULL_BEGIN
...
 /**
- An object conforming to the `MGLCalloutViewDelegate` method that receives messages related to the callout view’s interactive subviews.
+ An object conforming to the `MGLCalloutViewDelegate` method that receives
+ messages related to the callout view’s interactive subviews.
  */
 @property (nonatomic, weak) id<MGLCalloutViewDelegate> delegate;
 
 /**
- Presents a callout view by adding it to `inView` and pointing at the given rect of `inView`’s bounds. Constrains the callout to the bounds of the given view.
+ Presents a callout view by adding it to `inView` and pointing at the given rect
```
1ec5: While you’re at it, can you rename `inView` to just `view`? `inView:` is part of the selector; `view` is the name of the parameter.

1ec5:
Nice, thanks for taking care of this.

PR #5753, opened 2016-07-22
[core] Don't allow style mutations to be overwritten by revalidation

### LINKED ISSUE #5665: style revalidation vs. runtime styling
Currently, a style loaded via URL is revalidated on a schedule defined by the caching-related headers on the HTTP response. If a revalidation indicates that the style has changed, a new style object is created, replacing the old one wholesale. This poses a challenge for the runtime styling API. In the current naive implementation, any mutations to the style via runtime styling APIs are discarded when revalidation results in a new style:

* Changes to style properties revert to the original values
* Custom sources or layers disappear
* Pointers and references to `Source` or `Layer` objects in the existing style are invalidated

All of these effects are likely to vex SDK users, many of whom are unaware of the revalidation behavior, and often not in control of the caching headers in any case. See #5581 for an early example.

To mitigate this, there are several approaches we could pursue:

* Document and keep the current behavior, but add a "style reloaded" event, with the intent that users attach a handler that restores their desired modifications whenever a reload occurs. This is the approach suggested in #5581, but it doesn't strike me as very user-friendly because it doesn't address the root issue of the default behavior being surprising.
* Turn off revalidation of styles that have been mutated via runtime styling APIs. This is a relatively simple change which would avoid the surprising behavior, and likely not be detrimental to many users. The intersection of use cases that rely on both style revalidation and runtime styling is probably small. With this approach, we would hopefully still be able to do an initial revalidation when a map loads a previously-cached style, prior to any mutations.
* Turn off midstream revalidation of styles entirely -- like the above, but not conditioned upon use of runtime styling. The rationale for this is that revalidation of _tiles_ is good, and revalidation of the style _at initial load_ is good, but I don't know that anyone actually expects or depends on periodic revalidation of the style throughout the lifetime of a map. My impression is we implemented that more for consistency than user demand.
* Replace the wholesale reload with a "diff and patch" based approach, similar to that discussed in https://github.com/mapbox/mapbox-gl-js/issues/1989. This is a far more complex approach, with many unknown unknowns, and several thorny known unknowns like:
  * What do you do if a user mutates a style property, and then an update to that style also mutates that property?
  * What do you do if a user adds a layer referencing source _S_ originally found in the style, and then an update to that style removes _S_?

I currently lean toward one of the two middle options.

jfirebaugh:
Fixes #5665.

This fix is lenient about the case where the style is mutated between when the loading of a style URL is requested and when it completes. GL JS [throws](https://github.com/mapbox/mapbox-gl-js/blob/de61bbfa21f070e7afbe4fd42e24e8e44e7e25f1/js/style/style.js#L265-L269) in this case, and native probably should prohibit it as well.


jfirebaugh:
Rebased on `master`. This PR does not currently include the class list mutators (`addClass`, etc.) in the set of methods that trigger cancelling revalidation, although they do control state held by `Style`. We need to make a decision on #6118 regarding this state.

jfirebaugh:
@kkaefer, @tmpsantos and @brunoabinader can one of you review please?

src/mbgl/map/map.cpp
```diff
@@ -67,6 +67,7 @@ class Map::Impl : public style::Observer {
 
     std::string styleURL;
     std::string styleJSON;
+    bool styleMutated = false;
```
kkaefer: Can we move this to the `Style` object and have it keep track of whether it was mutated itself?

tmpsantos: :+1:

src/mbgl/map/map.cpp
```diff
@@ -746,22 +760,27 @@ AnnotationIDs Map::queryPointAnnotations(const ScreenBox& box) {
 #pragma mark - Style API
 
 style::Source* Map::getSource(const std::string& sourceID) {
+    impl->styleMutated = true;
     return impl->style ? impl->style->getSource(sourceID) : nullptr;
 }
 
 void Map::addSource(std::unique_ptr<style::Source> source) {
+    impl->styleMutated = true;
     impl->style->addSource(std::move(source));
```
tmpsantos: Just realized that if checks for if the style is !null on the getters but not on the setters.

src/mbgl/map/map.cpp
jfirebaugh: I wanted to do it that way, but the issue is that we want some style mutations to be excluded from this logic, specifically those that `AnnotationManager::updateStyle` performs. Adding annotations should not cancel the initial revalidation. (I'll add a test for this.)

src/mbgl/map/map.cpp
jfirebaugh: I opened #6121 as a followup issue for handling attempts to mutate an unloaded style.

PR #5754, opened 2016-07-22
Explicit conversion

GretaCB:
This PR explicitly handles type conversion, instead of allowing for silent type conversion. This was discovered when @springmeyer noticed that an `int` was being converted to an `bool` value incorrectly.

cc @springmeyer @jakepruitt @jfirebaugh


GretaCB:
hey @jfirebaugh @mikemorris , this PR is ready to roll as a node fix. After taking a look at CONTRIBUTING.md, do I need to make another commit with `[node]` in the message?

Curious what else needs to be done before/after merging:
- [ ] Update node [changelog](https://github.com/mapbox/mapbox-gl-native/tree/master/platform/node)
- [ ] ?

friedbunny:
@GretaCB I’m not sure about additional steps for node, but the green button in the repo uses squash-merging and gives you an opportunity to reword/prefix the commit at that time.

mikemorris:
> the green button in this repo uses squash-merging and gives you an opportunity to reword/prefix the commit at that time

Whoa, I actually had no idea it did that @friedbunny

Thanks for adding this @GretaCB, didn't realize this would be necessary!

PR #5757, opened 2016-07-22
[Qt] Use Qt text codec instead of nunicode

tmpsantos:
Remove one extra dependency. Qt has some bugs on the text codec it seems, but it is also important to keep the behavior aligned with the client API.


brunoabinader:
👍 one less dependency!

PR #5758, opened 2016-07-22
[default] uv_run should use UV_RUN_NOWAIT

tmpsantos:
UV_RUN_ONCE might block and that is not the semantics we expect from RunLoop::runOnce().

### DIFF
--- a/platform/default/run_loop.cpp
+++ b/platform/default/run_loop.cpp
@@ -158,7 +158,7 @@ void RunLoop::run() {
 void RunLoop::runOnce() {
     MBGL_VERIFY_THREAD(tid);
 
-    uv_run(impl->loop, UV_RUN_ONCE);
+    uv_run(impl->loop, UV_RUN_NOWAIT);
 }
 
 void RunLoop::stop() {


mikemorris:
@tmpsantos Under what circumstances would this block, and would blocking be (tangentially) dependent on [libuv threadpool size](http://docs.libuv.org/en/v1.x/threadpool.html)?

What would the downstream implications of blocking be? Could slow/hanging network I/O for requesting resources be a symptom of this?

tmpsantos:
@mikemorris this change probably won't affect downstream because `RunLoop::runOnce` is only used by the unit tests and when you destroy a worker thread.

`RUN_ONCE` will block until there are no pending callbacks. I realized that this function was buggy when I was testing Mapbox GL Native with Unity3D and using the `libuv` main loop as a "foreign main loop", and it was blocking a lot. The idea of `RunLoop::runOnce` is "process whatever is on the queue", thus the change to `NOWAIT`.

There we call `RunLoop::runOnce` all the time, so we don't need to wait for the pending callbacks because they will get processed later eventually.

mikemorris:
Recapping from chat:

> @mikemorris could destroying a map object (and its corresponding worker thread) block the libuv event loop?

> @tmpsantos: ugh, it could, maybe
should I say "oooops" :disappointed:

> @mikemorris: i wouldnt expect anything to really show up in the tests, im more thinking about situations we’ve seen where a map object is destroyed because of a rendering timeout, and that triggers subsequent render timeouts (and map object destruction) on _other_ map objects

> @tmpsantos: could that be caused because the HTTP stack is blocking somehow?
that would explain other map objects timeouting
if the default main loop on the main thread would be blocked, you wouldn't be able to start new requests in the first place

Issue #5201, opened 2016-06-01
MGLMultiPoint is both an abstract superclass and a concrete class for multipoints


jfirebaugh:
As noted in https://github.com/mapbox/mapbox-gl-native/pull/5200#discussion_r65283275, `MGLPolyline` and `MGLPolygon` should not inherit from `MGLMultiPoint`. Neither a polyline nor a polygon [_is-a_](https://en.wikipedia.org/wiki/Is-a) multipoint.

1ec5:
Is it important that the class that represents the Simple Features/GeoJSON multipoint be called “MGLMultiPoint”, as opposed to “MGLSparseMultiPoint” or “MGLPointCollection”? If so, we’d need to bump the major version number to acknowledge this breaking change. It would also be a deviation from MapKit’s terminology, but I think it’s an acceptable and reasonable one, considering that developers rarely interact with MGLMultiPoint by name.

/cc @boundsj @friedbunny

1ec5:
We already don't have a strict correspondence between Simple Features names and class names, for instance MGLShapeCollection, so named for consistency with the MGLShape protocol and to avoid confusion with the MGLFeature protocol.

incanus:
From https://github.com/mapbox/mapbox-gl-native/pull/6565#pullrequestreview-2627888, when #6565 is landed, we should clean up the abstract class checks mentioned there as part of this cleanup.

1ec5:
Now that #6565 has landed, we no longer treat MGLMultiPoint as an abstract class. The next step is to either:

1. Stop overloading MGLMultiPoint for GeoJSON multipoints, instead creating a new class for this purpose, as suggested in https://github.com/mapbox/mapbox-gl-native/issues/5201#issuecomment-222858385. This would be backwards-compatible and allow us to remove various `isMemberOfClass` checks.
2. Make MGLPolyline and MGLPolygon inherit directly from MGLShape, as suggested in https://github.com/mapbox/mapbox-gl-native/issues/5201#issue-157796755. Existing client code might use MGLMultiPoint as an umbrella class for both MGLPolyline and MGLPolygon, so this change would be backwards-incompatible. We’d have to wait until the next major version.

[...]

PR #5770, opened 2016-07-22
[ios] Do not allow multipoint annotations to be assigned views

friedbunny:
#### Changes
- Limits view annotations to single-point annotations.
- Refactors iosapp’s test shapes into a separate function.
 - Adds detail to the descriptions of individual shapes.
 - Makes the holey shape square (rather than trapezoidal, as below).

#### Issue
In v3.3.x, it was possible for any annotation class to be assigned an annotation view. If your implementation of `-mapView:viewForAnnotation:` didn't check if `annotation` was a kind of `MGLPointAnnotation`, you could end up drawing view annotations for `MGLMultiPoint` subclasses, such as `MGLPolygon`:

screen shot 2016-07-22 at 5 00 07 pm

As this fix uses the `MGLMultiPoint` superclass to exclude polygons and polylines, we will have to reevaluate if/when the proposed geometry class hierarchy changes in #5201 happen.

/cc @boundsj @1ec5 @incanus @frederoni

### DIFF
--- a/platform/ios/src/MGLMapView.mm
+++ b/platform/ios/src/MGLMapView.mm
@@ -4547,7 +4547,13 @@ mbgl::Duration MGLDurationInSeconds(NSTimeInterval duration)
         
         MGLAnnotationContext &annotationContext = pair.second;
         MGLAnnotationView *annotationView = annotationContext.annotationView;
-       
+
+        // Defer to the shape/polygon styling delegate methods
+        if ([annotationContext.annotation isKindOfClass:[MGLMultiPoint class]])
+        {
+            continue;
+        }
+
         if (!annotationView)
         {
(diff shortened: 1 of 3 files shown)


platform/ios/src/MGLMapView.mm
```diff
@@ -4547,7 +4547,13 @@ - (void)updateAnnotationViews
         
         MGLAnnotationContext &annotationContext = pair.second;
         MGLAnnotationView *annotationView = annotationContext.annotationView;
-       
+
+        // Defer to the shape/polygon styling delegate methods
+        if ([annotationContext.annotation isKindOfClass:[MGLMultiPoint class]])
```
1ec5: You could instead check for conformance to MGLShape if you're concerned about future class hierarchy chances.

1ec5:
:+1:

Issue #5763, opened 2016-07-22
[core] Line annotation disappears and/or clips to tile boundaries from zoom 3 onwards


brunoabinader:
In `macosapp` example, press ⌘L and zoom in until a piece of the example annotation disappears: 

![giphy 1](https://cloud.githubusercontent.com/assets/76133/17063372/920ef5b0-503f-11e6-8a92-027a51d94705.gif)

I suspect https://github.com/mapbox/mapbox-gl-native/pull/5514 could be the cause - need to investigate/bisect further to pinpoint the issue. From my experiments, https://github.com/mapbox/mapbox-gl-native/pull/5517 does not fix the issue.

:eyes: @yhahn @1ec5 @mourner

brunoabinader:
Also reproducible in `glfw-app`:
![glfw-app](https://cloud.githubusercontent.com/assets/76133/17077599/ce44a32e-50dd-11e6-876f-8d6c6906a74c.gif)

Here you can see also an interesting artifact: the line annotations are somehow clipped when crossing tile boundaries:

![out](https://cloud.githubusercontent.com/assets/76133/17077591/95f7453a-50dd-11e6-863d-57b50083db9d.gif)

brunoabinader:
Fix added in #5517.

PR #5774, opened 2016-07-23
[glfw] Added 'addRandomLineAnnotations'

brunoabinader:
Added `addRandomLineAnnotations` as part of investigating #5763.

### DIFF
--- a/platform/default/glfw_view.cpp
+++ b/platform/default/glfw_view.cpp
@@ -103,6 +103,7 @@ GLFWView::GLFWView(bool fullscreen_, bool benchmark_)
     printf("\n");
     printf("- Press `Q` to remove annotations\n");
     printf("- Press `P` to add a random custom runtime imagery annotation\n");
+    printf("- Press `L` to add a random line annotation\n");
     printf("- Press `W` to pop the last-added annotation off\n");
     printf("\n");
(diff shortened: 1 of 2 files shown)


tmpsantos:
:+1:

platform/default/glfw_view.cpp
```diff
@@ -164,9 +165,12 @@ void GLFWView::onKey(GLFWwindow *window, int key, int /*scancode*/, int action,
         case GLFW_KEY_Q:
             view->clearAnnotations();
             break;
-        case GLFW_KEY_P: {
+        case GLFW_KEY_P:
             view->addRandomCustomPointAnnotations(1);
-        } break;
+            break;
+        case GLFW_KEY_L:
+            view->addRandomLineAnnotations(1);
```
tmpsantos: shouldn't this be called `addRandomCustomLineAnnotations` for consistency with Point?

tmpsantos: or the other way around

brunoabinader: I believe that is meant to be _custom point_ because we're using a random, custom sprite image for the point symbols, whereas in this line annotation we use regular, non-custom line annotations.

PR #5776, opened 2016-07-23
[android] fixes #5762. Passing pixel float points to core.

### LINKED ISSUE #5762: Markers are not added at the correct location
Adding a gl marker to the map ontap/click just after starting up the map places a marker at the wrong location. See this example gif where it is first placed wrong and second attempt the behaviour is correct.

**Note** that this is a different issue as #5717.

ezgif com-video-to-gif 49

BharathMG:


tobrun:
Merged with  e16c906

bleege:
Good catch @BharathMG!

Issue #5745, opened 2016-07-21
[Android] Calling selectMarker from MapboxMap doesn't select custom MarkerView


BharathMG:
When calling 

mapboxMap.selectMarker(customMarkerView);

It does not set selected state of this Marker. It just calls onMarkerClick event. I need to return views from my Custom MarkerViewAdapter depending on this selection state. Pls help.

tobrun:
@BharathMG 
You need to use the MarkerViewManager for that:
```
mapboxMap.getMarkerViewManager().selectMarker(customMarkerView);
```

We need to route the `selectMarker` to use MarkerView equivalent.

Thank you for reaching out and reporting this issue.

BharathMG:
Thank you!

PR #5778, opened 2016-07-24
[android] Fixes #5777. deselectMarker is delegated to MapboxMap.

### LINKED ISSUE #5777: [android] deselectMarker has to be called on MapboxMap from MarkerViewManager like selectMarker
Currently, selectMarker delegates to MapboxMap when called on MarkerViewManager instance. But deselectMarker doesn't.

**Platform: Android**
**Mapbox SDK version: 4.2.0-SNAPSHOT**

### Steps to trigger behavior

 1. Call getMarkerViewManager().deselectMarker(marker).

### Expected behavior
 Selected State should be changed to false and delegated to MapboxMap deselectMarker method.
### Actual behavior
 Marker selection state is still selected. MapboxMap deselectMarker(marker) doesn't get triggered.

BharathMG:
Issue #5777.


tobrun:
@BharathMG thank you for the contribution and following up on #5745. I debugged the interaction and was able to verify that the select/deselect interaction works correctly now!

PR #5786, opened 2016-07-25
refs 654f015: bring back sprite-backed annotation option for comparison

incanus:
Background in https://github.com/mapbox/mapbox-gl-native/commit/654f0156397b68420bc48dfa2d53589ea7c8042e#commitcomment-18366764. 

We should have sprite-backed annotations as an option still for comparison to view-backed until such time that view-backed performance is equal or of negligible difference. 

This more or less reverses https://github.com/mapbox/mapbox-gl-native/commit/654f0156397b68420bc48dfa2d53589ea7c8042e, though it doesn't bring back `-mapView:didDeselectAnnotation:`, of which I'm not sure what the point was—something about changing to pins if we don't have a sprite to reuse? Open to feedback here. 

Subjectively, since last being in this code, I am noticing a pretty marked performance decrease for annotation sprites, possibly related to z-ordering that I see happening during map rotation. Might be something else. I'm going to profile out of curiosity separately from this—this PR is just about the functionality. But these should be a lot faster. 

/cc @1ec5 @boundsj

### DIFF
--- a/platform/ios/app/MBXViewController.m
+++ b/platform/ios/app/MBXViewController.m
@@ -639,6 +639,8 @@ static NSString * const MBXViewControllerAnnotationViewReuseIdentifer = @"MBXVie
         return nil; // use default marker
     }
 
+    NSAssert([annotation isKindOfClass:[MBXSpriteBackedAnnotation class]], @"Annotations should be sprite-backed.");
+
     NSString *title = [(MGLPointAnnotation *)annotation title];
     if (!title.length) return nil;
     NSString *lastTwoCharacters = [title substringFromIndex:title.length - 2];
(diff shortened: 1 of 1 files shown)


platform/ios/app/MBXViewController.m
```diff
@@ -301,6 +316,7 @@ - (void)parseFeaturesAddingCount:(NSUInteger)featuresCount
                 MGLPointAnnotation *annotation = [MGLPointAnnotation new];
                 annotation.coordinate = coordinate;
                 annotation.title = title;
+                annotation.subtitle = [NSString stringWithFormat:@"View: %i", useViews];
```
1ec5: Instead of using a magic string to detect a view-backed annotation, subclass MGLPointAnnotation and check whether it’s a kind of that subclass below. The class check is slightly more efficient than a string comparison and more self-documenting.

incanus: Yeah, was thinking about that but trying to avoid overhead of a class. Probably a good idea, though.

incanus:
Merged manually into `master` after rebase as of https://github.com/mapbox/mapbox-gl-native/commit/4e61b2e506f31682d67708832d409d38a7cb6ac0. 

Sorry, forgot the `[ios]` out front; will get that next time.

incanus:
Forgot to force push this into the branch after rebase; remedied as of https://github.com/mapbox/mapbox-gl-native/commit/4e61b2e506f31682d67708832d409d38a7cb6ac0.

PR #5791, opened 2016-07-25
Fix crash drawing compass on iOS 7

### LINKED ISSUE #5780: Unrecognized selector exception on initialization — -compassImage, +[UIFont systemFontOfSize:weight:]
#5497 didn’t completely fix #5486: not only is `UIFontWeightUltraLight` new to iOS 8.2, but so is `+[UIFont systemFontOfSize:weight:]`. As a result, initializing an MGLMapView in Mapbox iOS SDK v3.3.0 causes an uncaught exception and crash on iOS 8.1 and below.

Given the severity of this issue, I created an ios-v3.3.2 milestone to expedite the fix.

/cc @friedbunny @boundsj @frederoni @incanus

1ec5:
Fixes #5780.

/cc @friedbunny

### DIFF
--- a/platform/ios/CHANGELOG.md
+++ b/platform/ios/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 Mapbox welcomes participation and contributions from everyone.  Please read [CONTRIBUTING.md](../../CONTRIBUTING.md) to get started.
 
+## 3.3.2
+
+* Fixed a crash that occurred when initializing an MGLMapView on iOS 8.1 or below. ([#5791](https://github.com/mapbox/mapbox-gl-native/pull/5791))
+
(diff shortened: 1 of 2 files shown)


1ec5:
The “N” ends up being quite faint in iOS 8.1:

1ec5:
@incanus, can you try out this fix on an iOS 7 device, just in case?

friedbunny:
It is quite difficult to validate this fix: no one at Mapbox has access to an iOS 7 device and the iOS 7 simulator no longer runs on the current version of macOS. This fix is therefore **speculative** — we’re confident, but cannot be absolutely sure that it solves the issue.

Issue #3886, opened 2016-02-10
Ask HTTP implementations not to do content decoding


jfirebaugh:
We want to store compressed resources in the offline database. Our implementation ends up decompressing and then recompressing. Instead, we should ask NSURLSession, curl, and OkHttp to give us the gzipped body back, put the compressed body directly into the database, and only then decompress it for downstream consumers.

Specifically, we need to tell our HTTP libraries to as for and pass through gzip content encoding, since that's the only format we know how to decompress.

If we get back an uncompressed body (i.e. the server doesn't support content encoding), we should still try to compress it before storing it in the database.

This will improve performance of offline downloading significantly. Right now almost all of the CPU used during offline downloads goes toward compression and decompression, neither of which is necessary.

jfirebaugh:
* :white_check_mark:  [OkHttp](https://github.com/square/okhttp/wiki/Calls): If you add your own `Accept-Encoding: gzip` header to the request, you take responsibility for decoding.
* :white_check_mark: [Curl](https://curl.haxx.se/libcurl/c/CURLOPT_ACCEPT_ENCODING.html): "You can also opt to just include the Accept-Encoding: header in your request with CURLOPT_HTTPHEADER but then there will be no automatic decompressing when receiving data."
* :x: NSURLSession: Apparently not possible to take responsibility for decoding. :sob:

kkaefer:
This is indeed the right approach (but it still means we'll have to decompress in the SQLite source, and ideally move that to another thread).

Regarding `NSURLSessions`, I tried http://stackoverflow.com/a/16705063/331379 and indeed got the compressed response, but of course it's a private API so we can't use it :(

tmpsantos:
Agreed, it is the right approach. :white_check_mark: for Qt, it is done the same way as OkHttp.

I also second @kkaefer that we should move `decompress` to a separate thread.

jfirebaugh:
http://openradar.appspot.com/radar?id=4931101243998208

1ec5:
It’s pretty low-level, but does [`NSInputStream`](https://developer.apple.com/library/mac/documentation/Cocoa/Reference/Foundation/Classes/NSInputStream_Class/#//apple_ref/occ/clm/NSInputStream/inputStreamWithURL:) pass through gzip-encoded data?

incanus:
[...]

PR #5796, opened 2016-07-26
[core] Update SQLite schema with WAL journal mode and normal sync

friedbunny:
This pull request changes our offline/ambient cache database schema to use write-ahead logging and normal sync, which greatly increases the speed that resources can be written and read.

### Journal modes

SQLite offers several [journal modes](https://www.sqlite.org/pragma.html#pragma_journal_mode) — we were using the default, `DELETE`, which continuously writes a “rollback journal” that is then deleted at the end of the transaction. This mode offers good data safety and is [atomic](https://www.sqlite.org/atomiccommit.html), but is subsequently somewhat slow.

[Write-head logging](https://www.sqlite.org/wal.html) (`WAL`) is a different type of journaling that offers better performance but marginally reduced data safety. WAL effectively writes to a separate cache and periodically flushes back to the main database.

The SQLite docs go into great detail about the pros and cons of WAL, so I won’t rehash them here.

### Synchronous modes

SQLite also [offers several modes](https://www.sqlite.org/pragma.html#pragma_synchronous) of write synchronization. `FULL` is the default (which we were using) and is quite safe, but slow. `NORMAL` syncs less often (but still at “critical moments”) and is faster, but comes with a “non-zero” chance of data loss. `NORMAL` is the suggested mode to be used in conjunction with WAL and is what this PR uses.

`NORMAL` only offers a minimal performance increase, so far as I could tell.

### Performance tests with offline resources

Bottom-line: WAL offers 2× to 5× performance for us on iOS, depending on device vintage. Speeds appear to scale linearly — two rounds of tests were done with ~1000 resources and ~6000 resources. Initial writes/downloads were on my home internet connection, which was not saturated.

| | iPhone 6s (iOS 9.3.2, SQLite 3.8.10.2) | iPhone 5 (iOS 10b2, SQLite 3.13.0) |
|---|---|---|
| `DELETE` + `FULL` | 2.5 MB/s writing, 4.0 MB/s verifying | 1.2 MB/s writing, 1.8 MB/s verifying |
| `WAL` + `NORMAL` | **4.4 MB/s** writing, **21.9 MB/s** verifying | 1.2 MB/s writing, **4.2 MB/s** verifying |

The iPhone 5 appeared to be severely CPU-bound and was pegged at 100%.

### Availability

WAL was first supported in SQLite 3.7.0, which is available from iOS 5 and Android API 11 — so it should be available everywhere.

### Unknowns

- This PR will increase the chances of data loss, but it’s unknown whether or not this increase is significant.
- If data loss were to occur, the failure mode is unknown.
- Android performance is untested.
- The impact on other platforms?

/cc @mapbox/gl

### DIFF
--- a/platform/default/mbgl/storage/offline_database.cpp
+++ b/platform/default/mbgl/storage/offline_database.cpp
@@ -48,7 +48,8 @@ void OfflineDatabase::ensureSchema() {
             case 0: break; // cache-only database; ok to delete
             case 1: break; // cache-only database; ok to delete
             case 2: migrateToVersion3(); // fall through
-            case 3: return;
+            case 3: migrateToVersion4(); // fall through
+            case 4: return;
             default: throw std::runtime_error("unknown schema version");
             }
 
@@ -79,8 +80,10 @@ void OfflineDatabase::ensureSchema() {
 
         // If you change the schema you must write a migration from the previous version.
         db->exec("PRAGMA auto_vacuum = INCREMENTAL");
+        db->exec("PRAGMA synchronous = NORMAL");
+        db->exec("PRAGMA journal_mode = WAL");
         db->exec(schema);
-        db->exec("PRAGMA user_version = 3");
+        db->exec("PRAGMA user_version = 4");
     } catch (...) {
         Log::Error(Event::Database, "Unexpected error creating database schema: %s", util::toString(std::current_exception()).c_str());
         throw;
@@ -109,6 +112,12 @@ void OfflineDatabase::migrateToVersion3() {
     db->exec("PRAGMA auto_vacuum = INCREMENTAL");
     db->exec("VACUUM");
     db->exec("PRAGMA user_version = 3");
+}
+
+void OfflineDatabase::migrateToVersion4() {
+    db->exec("PRAGMA synchronous = NORMAL");
+    db->exec("PRAGMA journal_mode = WAL");
+    db->exec("PRAGMA user_version = 4");
 }
 
 OfflineDatabase::Statement OfflineDatabase::getStatement(const char * sql) {
--- a/platform/default/mbgl/storage/offline_database.hpp
+++ b/platform/default/mbgl/storage/offline_database.hpp
@@ -60,6 +60,7 @@ private:
     void ensureSchema();
     void removeExisting();
     void migrateToVersion3();
+    void migrateToVersion4();
 
     class Statement {
     public:
--- a/platform/ios/CHANGELOG.md
+++ b/platform/ios/CHANGELOG.md
@@ -40,6 +40,7 @@ Mapbox welcomes participation and contributions from everyone. Please read [CONT
 ### Other changes
 
 * Fixed a crash that sometimes occurred when initializing an MGLMapView. ([#5932](https://github.com/mapbox/mapbox-gl-native/pull/5932))
+* Improved offline and ambient cache database performance. ([#5796](https://github.com/mapbox/mapbox-gl-native/pull/5796))
 * As the user zooms in, tiles from lower zoom levels are scaled up until tiles for higher zoom levels are loaded. ([#5143](https://github.com/mapbox/mapbox-gl-native/pull/5143))
 * MGLMapDebugOverdrawVisualizationMask no longer has any effect in Release builds of the SDK. This debug mask has been disabled for performance reasons. ([#5555](https://github.com/mapbox/mapbox-gl-native/pull/5555))
 * Fixed a typo in the documentation for the MGLCompassDirectionFormatter class. ([#5879](https://github.com/mapbox/mapbox-gl-native/pull/5879))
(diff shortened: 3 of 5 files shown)


kkaefer:
> This PR will increase the chances of data loss, but it’s unknown whether or not this increase is significant.

When I experimented with WAL dbs, I never saw any sort of data loss. Even if it happens, we're writing a cache, so if we lose a couple of records, it's not a big deal.

> If data loss were to occur, the failure mode is unknown.

Our current layer is implemented so that it deletes the cache and starts over if the DB file is unreadable.

zugaldia:
> Our current layer is implemented so that it deletes the cache and starts over if the DB file is unreadable.

Considering that the cached tiles and the offline tiles share the same database, wouldn't this delete as well any offline tiles the user has previously downloaded, not just cached ones?

friedbunny:
>Considering that the cached tiles and the offline tiles share the same database, wouldn't this delete as well any offline tiles the user has previously downloaded, not just cached ones?

This is a good point and is a shortcoming of our combined ambient+offline approach. If the risk of data loss is not meaningfully increased by WAL+Normal, then this won’t be any more of a problem than it already is.

It’s also worth noting that Core Data [switched to WAL](https://developer.apple.com/library/mac/qa/qa1809/_index.html) for its backing SQLite database in iOS 7 and OS X 10.9 in 2013, so this is a well-tested configuration on Apple systems.

The test failures on CI look legit — I’ll investigate updating them for this PR.

friedbunny:
Test failures were because `OfflineDatabase.MigrateFromV2Schema` had a hard-coded expectation of schema `v3`.

incanus:
Couple notes based on reviewing this, some discussed offline: 

- I'm not seeing much mention of data loss—@friedbunny where are you reading up on that? 

- Docs mention adverse performance: 

  > WAL might be very slightly slower (perhaps 1% or 2% slower) than the traditional rollback-journal approach in applications that do mostly reads and seldom write.

  I wonder if we should concretely test this (also on multiple platforms) for the more general use case of general caching. 

- We should flag in code comments notes around read-only media just so we don't get bitten in future. 

  > Hence, SQLite databases should always be converted to PRAGMA journal_mode=DELETE prior to being transferred to read-only media.

- Should we also consider `TRUNCATE` mode, which might be a more middle-ground compromise between the default `DELETE` and `WAL`? 

  https://www.sqlite.org/pragma.html#pragma_journal_mode

- Lastly, and sorry if I missed this elsewhere, but can we get a crosslink to a master ticket for the main driver behind this, presumably offline performance on older devices? I know another issue (https://github.com/mapbox/mapbox-gl-native/issues/3886) that is holding up better offline performance. If we can ticket this holistically and make measurable gains over time, it will be easier to gauge priorities on individual tickets.

friedbunny:
> I'm not seeing much mention of data loss—@friedbunny where are you reading up on that?

Many of the examples [here](https://www.sqlite.org/howtocorrupt.html) mention WAL as a potential aggravator. There are some circumstances where WAL is an improvement (e.g., out of order writes), though.

> Docs mention adverse performance

I haven’t seen slower performance in any circumstances (read or write) so far, but yeah, we should test the ambient cacheing use case more throughly.

> Should we also consider `TRUNCATE` mode, which might be a more middle-ground compromise between the default `DELETE` and `WAL`?

I’d prefer to go full-hog and use WAL, which seems like it should be the best option for us, but I haven’t tested `TRUNCATE` yet.

> Lastly, and sorry if I missed this elsewhere, but can we get a crosslink to a master ticket for the main driver behind this

There’s no specific ticket related to database performance AFAIK. This pull request offers a fairly significant performance increase for everyone, but there’s no particular impetus other than that. Do we need more than [GitHub’s “offline” + “performance” tags](https://github.com/mapbox/mapbox-gl-native/issues?utf8=✓&q=label%3Aoffline%20label%3Aperformance%20)?

jfirebaugh:
This sounds good to me.

> The iPhone 5 appeared to be severely CPU-bound and was pegged at 100%.

This is due to https://github.com/mapbox/mapbox-gl-native/issues/3886.

friedbunny:
Cool! On iOS, we’re feeling good about getting this in v3.4.0 and would like to merge this week.

This will affect other platforms, so it’s important that we get sign-off from the Android and Qt folks. Any concerns about switching to WAL, @tmpsantos @brunoabinader @zugaldia @tobrun?

zugaldia:
@friedbunny 👍

incanus:
Is it possible at all to have some sort of baseline test for speed so we can attach numbers to this?

tmpsantos:
> Any concerns about switching to WAL?

There is one corner case that we need to keep in mind which is this only works with write access to the cache, so if you have a read-only offline database on a read-only partition, it won't work.

That said, :shipit:. This is a fairly easy change to revert in a custom build if we ever need to address this corner case.

friedbunny:
Thanks everybody, sounds like we’re good to go. I’m going to rebase this, let the tests pass, and then merge today.

PR #5807, opened 2016-07-27
[Qt] Implement QQuickMapboxGL color property

brunoabinader:
Implements background `color` property for `QQuickMapboxGL` after runtime style API has been implemented (#837 and #5642).

Example:
qml-color-change2


tmpsantos:
:+1:

PR #5813, opened 2016-07-27
[ios] Limit annotation view pan gesture check to own recognizer

### LINKED ISSUE #5539: Initiating a pan from an annotation view should move the map
**Version**: v3.3.0-beta.2 at 388fdd2ff7bfc652d0bd40a614a31e6a235e1328

**Reproduction:**
1. Find an annotation _view_ on the map.
2. Try panning the map by dragging from that annotation view.
3. Map fails to pan (nothing happens).

I’m unsure if this is a regression brought on by draggable view annotations, but I’m inclined to think that it is not. MGLAnnotationImage does pan the map when you start dragging from on top of one, as do other iOS map libraries’ native views.

/cc @boundsj @frederoni @1ec5

friedbunny:
Fixes #5539. Allows non-dragging pan gestures that are initiated on view annotations to pan the underlying map.

(This PR is now filed against the v3.3.0 release branch and supersedes #5812.)

Annotation views each have long press and pan gesture recognizers. If a view is not draggable or not in a positive dragging state, the pan gesture recognizer should not begin. This commit makes the criteria for rejecting pans more strict — only `_panGestureRecognizer` should be rejected, not the entire `UIPanGestureRecognizer` class.

/cc @frederoni @1ec5

### DIFF
--- a/platform/ios/CHANGELOG.md
+++ b/platform/ios/CHANGELOG.md
@@ -5,6 +5,7 @@ Mapbox welcomes participation and contributions from everyone.  Please read [CON
 ## 3.3.2
 
 * Fixed a crash that occurred when initializing an MGLMapView on iOS 8.1 or below. ([#5791](https://github.com/mapbox/mapbox-gl-native/pull/5791))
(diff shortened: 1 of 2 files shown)


1ec5:
👍

Issue #5721, opened 2016-07-19
MapView: eglSwapBuffers Error


ivovandongen:
When chaining some operations on the MapView the map, applications [crash](https://github.com/mapbox/mapbox-gl-native/issues/5610#issuecomment-233123509) due to an error while rendering. Specifically:

```
07-16 11:25:32.255 7133-7133/com.mapbox.mapboxsdk.testapp E/libEGL: call to OpenGL ES API with no current context (logged once per thread)
07-16 11:25:32.258 7133-7133/com.mapbox.mapboxsdk.testapp D/mbgl: {unknown}[Android]: NativeMapView::notifyMapChange()
07-16 11:25:32.259 7133-7133/com.mapbox.mapboxsdk.testapp E/mbgl: {unknown}[OpenGL]: eglSwapBuffers() returned error 12301
07-16 11:25:32.259 7133-7133/com.mapbox.mapboxsdk.testapp D/AndroidRuntime: Shutting down VM
07-16 11:25:32.260 7133-7133/com.mapbox.mapboxsdk.testapp E/AndroidRuntime: FATAL EXCEPTION: main
                                                                            Process: com.mapbox.mapboxsdk.testapp, PID: 7133
                                                                            java.lang.Error: eglSwapBuffers() failed
                                                                                at com.mapbox.mapboxsdk.maps.NativeMapView.nativeRender(Native Method)
                                                                                at com.mapbox.mapboxsdk.maps.NativeMapView.render(NativeMapView.java:131)
                                                                                at com.mapbox.mapboxsdk.maps.MapView.onDraw(MapView.java:1302)
                                                                                at android.view.View.draw(View.java:16184)
                                                                                at android.view.View.updateDisplayListIfDirty(View.java:15180)
```

As @jfirebaugh [suggests](https://github.com/mapbox/mapbox-gl-native/issues/5610#issuecomment-233480297), it might be that the the MapView doesn't reactivate the GL Context properly. We need to check that the the context is deactivated and re-activated properly. Atm, it's easily re-produced by removing a layer within a camera animation finish callback.

It might be a good idea to isolate from issues like this by placing callbacks on the main looper. This ensures that the callbacks always happen on the main thread as well. A similar solution is also used on the QT bindings I believe (cc @tmpsantos).

cc @tobrun @zugaldia @bleege

ivovandongen:
[...]

PR #5834, opened 2016-07-29
fixes #5810: post animateCamera()/easeCamera() callbacks on main thread

### LINKED ISSUE #5810: Camera Callbacks Not Returning On Main UI Thread
`MapboxMap.animateCamera()` and `MapboxMap.easeCamera()` are firing their Callbacks on the Map Thread and not the main UI Thread as the JavaDocs comments say.

incanus:
Naive approach here based on my experience, but I think that it's correct. Please tell me it really is this simple! 

/cc @bleege @zugaldia @cammace @ivovandongen

### DIFF
--- a/platform/android/MapboxGLAndroidSDK/src/main/java/com/mapbox/mapboxsdk/maps/MapboxMap.java
+++ b/platform/android/MapboxGLAndroidSDK/src/main/java/com/mapbox/mapboxsdk/maps/MapboxMap.java
@@ -380,7 +380,11 @@ public class MapboxMap {
             @Override
             public void onCancel() {
                 if (callback != null) {
-                    callback.onCancel();
+                    mMapView.post(new Runnable() {
+                        @Override public void run() {
+                            callback.onCancel();
+                        }
+                    });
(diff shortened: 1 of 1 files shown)


ivovandongen:
@incanus Let's discuss this first on Monday. There is more to this than reported in #5810 as I described and am working on in #5721.

incanus:
Sounds good—figured, as with threading issues almost all of the time, that it's more complex than first blush.

ivovandongen:
@incanus There's actually more going on still. Just created an activity to re-create the issue and ran into a stack overflow exception. I'm going to make a separate ticket for that as well so we can discuss the root cause.

incanus:
After offline discussion with @bleege @ivovandongen, broader context here is that with rendering on the main thread now, this shouldn't be an issue as such and this PR is irrelevant. We may still yet have race condition issues around state and clearing internal cyclical callbacks, e.g. https://github.com/mapbox/mapbox-gl-native/issues/5836, but something like what is proposed here purely in Java land does not need to be happening.

PR #5838, opened 2016-08-01
build and publish Node.js binaries with BUILDTYPE=Release

brunoabinader:
Commit mapbox/mapbox-gl-native@29ee60a caused npm packages to be built in debug mode.

/cc @mikemorris


brunoabinader:
👍

PR #4638, opened 2016-04-07
[wip] add OSMesa HeadlessView and build target

mikemorris:
Fixes #4498 

Needs some further cleanup to be used alongside the GLX HeadlessView (instead of breaking the GLX build), but this is now compiling and passing `npm test` and `npm run test-suite` on a local VM with OpenSWR on OSMesa.

Build and test string I was using:

```
CXX=g++-4.9 HEADLESS=osmesa npm install --build-from-source && LD_LIBRARY_PATH=/usr/local/lib/osmesa/lib npm test
```

/cc @zmully

Issue #4498, opened 2016-03-28
Add Linux OSMesa support to HeadlessView


mikemorris:
http://www.mesa3d.org/osmesa.html

For testing OpenSWR rendering. Is building GLFW from source with Wayland/EGL support necessary too?

~~http://www.glfw.org/docs/latest/compile.html#compile_manual~~

/cc @zmully

mikemorris:
OMG I think I'm hitting https://bugs.chromium.org/p/chromium/issues/detail?id=482787, I can't get `OSMesaMakeCurrent` to release a context successfully...

mikemorris:
This is fun https://cgit.freedesktop.org/mesa/mesa/tree/src/mesa/drivers/osmesa/osmesa.c#n983

brunoabinader:
I believe I found a way to fix our anti-aliasing issues (#1555) by __using OSMesa + post processing__.

### Replacing GLX with EGL

My efforts towards EGL usage (based on #5843) were unsuccessful - the EGL swrast driver works in non-VM environments, but not otherwise - it fails to find a suitable display, and `EGL_DEFAULT_DISPLAY` cannot be used in those cases.

Using mesa 12.0.3 or 13.0.0 with GLX (LLVMpipe) also produces the very same render test results as previous mesa versions, and I think I understood _why_: The software renderer must explicitly enable [post-processing](http://www.mesa3d.org/postprocess.html) effects such as MLAA. However, post-processing is only available when using Gallium drivers. Gallium support for GLX library only comes with `gallium-xlib`, which involves using X and thus not desirable for headless support.

### Enter [OSMesa](http://www.mesa3d.org/osmesa.html)

I took liberty of rebasing/updating @mikemorris PR in #4638. I've also re-arranged the mason scripts for mesa 13.0.0 to produce a minimal binary, enabling Gallium implementation of the OSMesa library. Being OSMesa a headless, _simple_ layer on top of OpenGL and by using Gallium to provide _post-processing_ effects such as MLAA, I was then able to __obtain 42 passing render tests that are currently ignored__ (same number as in macOS).

Post-processing can be enabled by calling e.g. ```OSMesaPostprocess(glContext, "pp_jimenezmlaa", 8);``` right after creating the context - this enables Jimenez' [MLAA](http://www.iryoku.com/mlaa/) filter to be applied.

In sum:
 - OSMesa is truly headless - no extra dependencies besides LLVM 3.8 (available via apt-get on Travis)
[...]

PR #5843, opened 2016-08-02
Replace GLX by EGL platform for headless view

tiagovignatti:
Calling X11 window system is superfluous for headless rendering. This patch implements EGL platform using GBM, which is slightly more simple than the GLX path when using X11.

In principle there are no big advantages in terms of performance etc. My motivation behind this was to get in touch with the code and the project. For testing I'm using:

`$ unset DISPLAY && make`

Also, it removes headless GLX code and use EGL platform for Qt and Node.


brunoabinader:
cc @kkaefer

mikemorris:
Hmmm, neat! I've got a few questions:

1. Am I correct in thinking that EGL is just the higher-level interface, and wouldn't directly affect anything lower level, like software rendering with the `llvmpipe` driver for Mesa on Linux?
2. Any idea on how this approach would compare to rendering with [OSMesa](http://www.mesa3d.org/osmesa.html)? It seems appealing as an alternative that would be supported on MacOS and Linux?
3. Would this allow us to [drop the `xvfb` dependency](https://github.com/mapbox/mapbox-gl-native/blob/6c05e51cbc0781d6826370a618dc61855b79bb4f/scripts/travis_setup.sh#L33-L41) for software rendering?

tiagovignatti:
EGL is more like the glue between the rendering API and what is providing the buffers to be rendered. So changing the headless now to use EGL doesn't mean much besides clean up -- with GLX there was this indirection to the X (X11) server, which is a *window* system but, well, headless doesn't care about windows. 

So there shouldn't be any impact whether your system is rasterizing with software or not. I think you should be able to remove xvfb dependency now from your build bot (but again: it doesn't mean that its rendering method will be changed).

I hope I've answered the three questions, @mikemorris ?

mikemorris:
@tiagovignatti Was specifically referring to https://github.com/mapbox/mapbox-gl-native/pull/4638 re OSMesa, as an alternative means of eliminating the window server. I'm 👍 in general on this though, as EGL seems to [be in better shape than the OSMesa API](https://github.com/mapbox/mapbox-gl-native/issues/4498#issuecomment-207052374).

mikemorris:
This is currently failing CI tests because we're running against a version of Mesa [compiled without EGL](https://github.com/mapbox/mason/blob/master/scripts/mesa/11.2.2/script.sh#L19-L28). I can work on publishing a release of Mesa [12.0.1](http://www.mesa3d.org/relnotes/12.0.1.html) with [EGL enabled](http://www.mesa3d.org/egl.html).

tiagovignatti:
@mikemorris I think these are orthogonal problems: replace GLX by EGL would simply remove the window system underneath which is useless for the headless configuration, while your suggestion of using libOSMesa means that headless would also be using a software rasterizer instead.

Let me know how the new Mesa integration with the CI is going. Thanks!

tiagovignatti:
One more thing @mikemorris. In case your CI bot doesn't have a GPU hw available, then you will need to set Mesa to be compiled using ```--with-gallium-drivers=swrast```

platform/default/headless_display.cpp
```diff
@@ -29,40 +32,41 @@ HeadlessDisplay::HeadlessDisplay() {
     }
 #endif
 
-#if MBGL_USE_GLX
-    if (!XInitThreads()) {
-        throw std::runtime_error("Failed to XInitThreads.");
+#if MBGL_USE_EGL
+    // TODO(vignatti): it should search for a valid render* handle instead.
+    const char device_name[] = "/dev/dri/renderD128";
```
mikemorris: How would we update this to search for a valid handle?

platform/default/headless_view_egl.cpp
```diff
@@ -4,51 +4,62 @@
...
+    glContext = eglCreateContext(dpy, config, EGL_NO_CONTEXT, NULL);
+    if (glContext == EGL_NO_CONTEXT) {
+        mbgl::Log::Error(mbgl::Event::OpenGL, "eglCreateContext() returned error 0x%04x",
+                         eglGetError());
+        throw std::runtime_error("Error creating GL context object");
     }
 
-    if (glContext == nullptr) {
-        throw std::runtime_error("Error creating GL context object.");
+    // TODO(vignatti): think about an optimal size here.
```
mikemorris: Any thoughts here? Are we actually rendering to the gbm surface here now (as opposed to the `pbuffer` we were only using to initialize the context)?

/cc @kkaefer

mikemorris:
> replace GLX by EGL would simply remove the window system underneath which is useless for the headless configuration

👍 

> your suggestion of using libOSMesa means that headless would also be using a software rasterizer instead. In case your CI bot doesn't have a GPU hw available, then you will need to set Mesa to be compiled using `--with-gallium-drivers=swrast`

Yep, this is exactly the case - we need to support hardware rendering as well as software rendering because Travis CI boxes don't have GPUs. I got a Mesa 12.0.1 build configured and packaged up at https://github.com/mapbox/mason/compare/mesa-12.0.1, and have mapbox-gl-native building against it successfully on Travis now by applying https://github.com/mapbox/mapbox-gl-native/commit/a6899b5ba87d4baf8e41479cbea16a4c2bf92d05 to your branch @tiagovignatti.

CI tests are currently failing on a runtime error, which I'm guessing is expected at this point judging by your `todo` notes, would be great to figure out how to get this wrapped up =)

```
terminate called after throwing an instance of 'std::runtime_error'
  what():  Couldn't open drm device.
Aborted (core dumped)
```

platform/default/headless_display.cpp
tiagovignatti: done.

platform/default/headless_view_egl.cpp
tiagovignatti: done.

I actually removed completely the EGL surface code because that's not needed in this single-window case.

tiagovignatti:
hey back @mikemorris.

I've just rebased now the two commits and fixed the TODOs that you commented about. They're living [here](https://github.com/tiagovignatti/mapbox-gl-native/commits/egl-headless-view_v2) (```egl-headless-view_v2``` branch).

I thought about rebasing https://github.com/mapbox/mapbox-gl-native/commit/04fc3cb88389ee1ad607fceefd2fd2549addbb09 as well but I can't test it cause I don't have access to the Travis machine. I'm sure you can roll this commit quite easy on top of my new branch though :) Let me know if the CI tests are alright then.

tmpsantos:
> use EGL platform for Qt

Qt has now a "Qt headless view" using only Qt stuff, which could be potentially used on Windows.

brunoabinader:
Follow-up: #7007.

PR #5847, opened 2016-08-02
Migrate playground to Swift 3

### LINKED ISSUE #5846: Upgrade playground to Swift 3
We should conditionally use Swift 3 syntax in the [iOS playground](https://github.com/mapbox/mapbox-gl-native/blob/master/platform/ios/Mapbox.playground/Contents.swift).

/cc @boundsj

1ec5:
Left in the old code for those using Xcode 7.3.

Fixes #5846.

/cc @friedbunny @boundsj @frederoni


boundsj:
👍

Issue #5858, opened 2016-08-03
Runtime style api: add convenience methods


ivovandongen:
In building out the examples (https://github.com/mapbox/mapbox-android-demo/issues/115) @cammace noted some things that could be smoothed out:

- Getting a color int from color properties directly instead of an rgba string
- Adding a fluent api to the Layers.

ivovandongen:
Merged

PR #5860, opened 2016-08-03
[android] #5858 - added convenience methods for color properties layer

ivovandongen:
Some convenience methods the runtime style api


tobrun:
LGTM :shipit:

PR #5862, opened 2016-08-03
fixes #5238: Android marker view z-ordering

### LINKED ISSUE #5238: Add View Annotations Z ordering
We have been hearing about the need to be able to order Markers. For View based Markers an end developer should be able to indicate to a marker to be placed on top of another marker.  This system must integrate with other already shown View elements as CompasView, MyLocationView etc. This could be made part or based on current implementation of MarkerViewAdapter, as a separate ViewOverlay implementation or exposed in Marker itself.

This should solve the issues we have been seeing related to overlapping views.
This issue doesn't cover z-ordering or layering from core in https://github.com/mapbox/mapbox-gl-native/issues/991

incanus:
First phase of #5238, allowing addition of marker views meant to always be above another, named marker view. 

I'll squash etc. once this looks good to go. 

@tobrun @zugaldia

### DIFF
--- a/platform/android/MapboxGLAndroidSDK/src/main/java/com/mapbox/mapboxsdk/annotations/BaseMarkerViewOptions.java
+++ b/platform/android/MapboxGLAndroidSDK/src/main/java/com/mapbox/mapboxsdk/annotations/BaseMarkerViewOptions.java
@@ -21,6 +21,7 @@ public abstract class BaseMarkerViewOptions<U extends MarkerView, T extends Base
     protected String snippet;
     protected String title;
     protected Icon icon;
+    protected MarkerView aboveMarkerView;
     protected boolean flat;
     protected float anchorU = 0.5f;
     protected float anchorV = 1f;
@@ -78,6 +79,17 @@ public abstract class BaseMarkerViewOptions<U extends MarkerView, T extends Base
      */
     public T icon(Icon icon) {
         this.icon = icon;
+        return getThis();
+    }
+
+    /**
+     * Set the MarkerView above which this MarkerView is shown.
+     *
+     * @param aboveMarkerView the MarkerView above which to show
+     * @return the object for which the method was called
+      */
+    public T aboveMarkerView(MarkerView aboveMarkerView) {
+        this.aboveMarkerView = aboveMarkerView;
         return getThis();
     }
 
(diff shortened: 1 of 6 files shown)


incanus:
Something that occurred to me here is my relative inexperience with memory management in Java, so I would welcome thoughts on any sort of issues with `aboveMarkerView` references as a result of this.

tobrun:
@incanus this is great! just one remark is that the builder creating the MarkerView must comply to the Parceable interface. The Parceable interface is used to restore the builder and its variables after a configuration change (eg. orientation change). At this point the MarkerView itself does not implement the Parceable interface and we won't be able to recreate this interaction correctly (the set MarkerView will be null). 

There are a couple of ways to workaround this:
 - expose the functionality only on MarkerView (it's the user responsibility to restore correctly)
 - expose a method on `MarkerViewManager` that takes in 2 MarkerViews
 - only save the id of the MarkerView and lookup the concrete MarkerView when needed through MarkerViewManager

incanus:
> only save the id of the MarkerView and lookup the concrete MarkerView when needed through MarkerViewManager

I'm not used to objects having this, what sounds like a global id, that can be used. I'll look into this.

tobrun:
> only save the id of the MarkerView and lookup the concrete MarkerView when needed through MarkerViewManager

Now thinking of this, this won't be an option since the id will only be available after we have added the marker to the map (this id would be equal to the core gl marker id).

incanus:
Backing this up a moment @tobrun, I am seeing things ordered ok after device orientation change. Is this truly a problem?

incanus:
@tobrun and I voiced on this at length yesterday and worked through some consideration of both an API design like I've proposed as well as a callback-based sort comparator. The short is that it probably makes sense to skip right to the comparator as fetching it from a callback is a lesser consideration on performance than the actual sorting. Additionally, we could add `Parcelable` to `Marker` in order to make the proposed solution work, but the API is still limiting in referring to only one other marker, when marker add order needs to remain flexible. 

I will start on this soon, but it's not as high priority right now.

incanus:
Closing this PR per https://github.com/mapbox/mapbox-gl-native/pull/5862#issuecomment-239164783 since we're go the comparator route instead.

PR #5870, opened 2016-08-04
5859 generate sanity tests

### LINKED ISSUE #5859: Generate espresso activity sanity tests
In same line as we are now generating runtime style tests. We could generate the activity sanity tests for classes found in the `com.mapbox.mapboxsdk.activity` package as a separate build task.

 - [x] create a separate generation tests task in our build
 - [x] create a ejs template for it
 - [x] generate a testsuite
 - [x] rename attributes as name and id that they are equal for all Activities
 - [x] git ignored generated test
 - [x] integration as bitrise task
 - [x] integration in AWS device farm through gradle plugin

tobrun:
closes #5859


zugaldia:
@tobrun This is awesome, good thinking on bringing automatic generation here. Could you document this (how to write a new test, how to generate, how to run) in [`tests`](https://github.com/mapbox/mapbox-gl-native/tree/master/platform/android/tests)? Thanks!

tobrun:
Implementation is done! now only need to update the docs before we can merge

tobrun:
@ivovandongen @zugaldia care to review?

zugaldia:
@tobrun @ivovandongen should the generated code be committed?

zugaldia:
@tobrun @ivovandongen sorry, clicked comment too fast. I'm wondering if it would't be better to have the generated code automatically created when we do e.g. `make android`.

tobrun:
> should the generated code be committed?

The problem with this is that these will constantly come up as changes when you execute make. 

> would't be better to have the generated code automatically created when we do e.g. make android

I really like this approach, I would like to look into this when I work on CI integration for UI tests. Either way the tests should be generated again at PR time so incorrect changes/not correctly implemented activities are automatically flagged. For now I'm going to rebase and resolve conflicts.

ivovandongen:
@zugaldia @tobrun 

> should the generated code be committed?

For the runtime style api, I opted to commit the generated classes as it makes exploring the codebase easier, especially when using GitHub.

> would't be better to have the generated code automatically created when we do e.g. make android

Again, for the runtime style api, I opted not to do this as the spec doesn't change very frequently and it makes build times longer. In this case it makes sense. If we'd do that though, we would ideally make it smart enough that it doesn't unnecessarily increases build times.

tobrun:
Been thinking a bit more about this and I suggest gitignoreing the folder and generating them whenever the test need to be executed (with this setup it's than also executed with `make android` I believe. Committing and making them versioned is not necessary and results in unneeded overhead.

zugaldia:
@tobrun Agreed. Per @mapbox/android convo, let's keep this autogenerated tests out of version control and generate them when need it. This does not affect the automated classes and tests included in https://github.com/mapbox/mapbox-gl-native/issues/5610.

tobrun:
cbcc691 introduces a `gen` pacakge that contains the generated test files. This folder was added to the gitignore file and as a result generated files will not show anymore as changes.

tobrun:
9a14237 introduces Bitrise configuration to run tests on AWS + AWS Device Farm gradle plugin

tobrun:
4edf35c removes bug of having the gradle.properties set for AWS credentials

ivovandongen:
Looks good 👍

zugaldia:
> remove python from android sdk

😢 but the right thing to do now that we've got a better way.

tobrun:
updated docs about aws gradle plugin. Going to wait for CI and merge it.

tobrun:
cc @mapbox/android

PR #5872, opened 2016-08-05
[Qt] Added QQuickMapboxGLStyle

brunoabinader:
This PR introduces a new QML item: `QQuickMapboxGLStyle`. This item is responsible for managing the Mapbox GL style, as seen on the QML example below:
```qml
...
MapboxMap {
    style: MapboxStyle {
        id: styleStreets
        style: "mapbox://styles/mapbox/streets-v9"
    }
    ...   
```

~~Now every `QQuickMapboxGLStyleProperty` can be a child item of either `QQuickMapboxGL` or `QQuickMapboxGLStyle`.~~ (as per discussion with @tmpsantos, we felt like style properties should only be children of the style to simplify the API).

A `QQuickMapboxGLStyleProperty` being a child of `QQuickMapboxGLStyle` assures that only properties associated with that particular style will reflect on the map - and that happens only when the style is the current set via `style` property.

Other changes in this PR includes:
- Invert parent/child notification mechanism: Now every parent item e.g. `QQuickMapboxGL` and `QQuickMapboxGLStyle` is responsible for connecting/disconnecting signals and slots for its child items upon insertion/removal, and not the opposite.
- Other minor cleanups and API simplifications.

### DIFF
--- a/platform/qt/include/qquickmapboxglstyle.hpp
+++ b/platform/qt/include/qquickmapboxglstyle.hpp
@@ -1,0 +1,39 @@
+#ifndef QQUICKMAPBOXGLSTYLE_H
+#define QQUICKMAPBOXGLSTYLE_H
+
+#include <QQuickItem>
+
+class Q_DECL_EXPORT QQuickMapboxGLStyle : public QQuickItem
+{
+    Q_OBJECT
+
+    Q_PROPERTY(QString url READ url WRITE setUrl NOTIFY urlChanged)
+    Q_PROPERTY(QString styleClass READ styleClass WRITE setStyleClass NOTIFY classChanged)
+
+public:
+    QQuickMapboxGLStyle(QQuickItem *parent = Q_NULLPTR);
+    virtual ~QQuickMapboxGLStyle() {}
+
+    // QQuickItem implementation
+    virtual void itemChange(QQuickItem::ItemChange, const QQuickItem::ItemChangeData &);
+
+    void setUrl(const QString &);
+    QString url() const;
+
+    void setStyleClass(const QString &);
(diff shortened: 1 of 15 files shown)


brunoabinader:
**Note**: `QQuickMapboxGLStyle` opens the way for future `QQuickMapboxGLStyleLayer`, `QQuickMapboxGLStyleSource` and `QQuickMapboxGLStyleFilter` implementations.

tmpsantos:
> Now every QQuickMapboxGLStyleProperty can be a child item of either QQuickMapboxGL or QQuickMapboxGLStyle

QQuickMapboxGLStyle should own a QQuickMapboxGLStyleProperty:

1. otherwise we have 2 ways of doing the same thing
2. semantically I don't think the former is correct.

brunoabinader:
> `QQuickMapboxGLStyle` should own a `QQuickMapboxGLStyleProperty`

You are right - this simplifies the code logic and avoids ambiguous corner cases 👍

platform/qt/include/qquickmapboxglstyle.hpp
```diff
@@ -0,0 +1,39 @@
+#ifndef QQUICKMAPBOXGLSTYLE_H
+#define QQUICKMAPBOXGLSTYLE_H
+
+#include <QQuickItem>
+
+class Q_DECL_EXPORT QQuickMapboxGLStyle : public QQuickItem
+{
+    Q_OBJECT
+
+    Q_PROPERTY(QString style READ style WRITE setStyle NOTIFY styleChanged)
```
tmpsantos: s/style/url ?

tmpsantos:
:+1:

PR #5874, opened 2016-08-05
[core] Refcount the view activations

### LINKED ISSUE #5721: MapView: eglSwapBuffers Error

### LINKED ISSUE #5836: Chained animations lead to StackOverFlowException
In creating an activity to reproduce the error in #5721 I ran into a situation where the test app crashes due to a StackOverflowException. The relevant code in question:

```
    private void animateToNextPosition(final int iterations) {
        //Setup the target camera position
        final CameraPosition cameraPosition = new CameraPosition.Builder()
                .target(getNextLatLng())
                .bearing(270)
                .tilt(20)
                .build();

        //Create a callback
        final MapboxMap.CancelableCallback callback = new MapboxMap.CancelableCallback() {
            @Override
            public void onCancel() {
                Log.i(TAG, String.format("Duration onCancel Callback called on Thread id: %s, name %s", Thread.currentThread().getId(), Thread.currentThread().getName()));
                //Toast.makeText(CameraAnimationChainingActivity.this, "Duration onCancel Callback called.", Toast.LENGTH_SHORT).show();
            }

            @Override
            public void onFinish() {
                Log.i(TAG, String.format("Duration onFinish Callback called on Thread id: %s, name %s", Thread.currentThread().getId(), Thread.currentThread().getName()));
                //Toast.makeText(CameraAnimationChainingActivity.this, "Duration onFinish Callback called.", Toast.LENGTH_SHORT).show();
                if (iterations > 0) {
                    animateToNextPosition(iterations - 1);
                }
            }
        };

        mapView.getMapAsync(new OnMapReadyCallback() {
            @Override
            public void onMapReady(MapboxMap mapboxMap) {
                mapboxMap.animateCamera(CameraUpdateFactory.newCameraPosition(cameraPosition), 7500, callback);
            }
        });
    }
```

Resulting in:

```
...
07-30 16:03:34.426  4707  4707 I CameraAnimationChainingActivity: Duration onFinish Callback called on Thread id: 1, name main
07-30 16:03:34.427  4707  4707 I CameraAnimationChainingActivity: Duration onFinish Callback called on Thread id: 1, name main
07-30 16:03:34.427  4707  4707 I CameraAnimationChainingActivity: Duration onFinish Callback called on Thread id: 1, name main
07-30 16:03:34.595  4707  4707 W art     : JNI WARNING: java.lang.StackOverflowError thrown while calling printStackTrace
07-30 16:03:34.625  4707  4707 W System.err: java.lang.StackOverflowError: stack size 8MB
```

The actual issue is caused when cancelTransitions() is called in [MapView#flyto()](https://github.com/mapbox/mapbox-gl-native/blob/master/platform/android/MapboxGLAndroidSDK/src/main/java/com/mapbox/mapboxsdk/maps/MapView.java#L1252), which eventually reaches [Transform::cancelTransition()](https://github.com/mapbox/mapbox-gl-native/blob/master/src/mbgl/map/transform.cpp#L638) in core where the previous `transitionFinishFn` is still set:

```
    if (transitionFinishFn) {
        transitionFinishFn();
    }

    transitionFrameFn = nullptr;
    transitionFinishFn = nullptr;
```

Might be solved by clearing the `transitionFn` and then calling it. A similar situation exists in [startTransition()](https://github.com/mapbox/mapbox-gl-native/blob/master/src/mbgl/map/transform.cpp#L602) which might instigate the loop.

Possibly related: #5833

cc @tmpsantos @zugaldia

tmpsantos:
Fixes #5836
Fixes #5721

### DIFF
--- a/platform/android/src/native_map_view.cpp
+++ b/platform/android/src/native_map_view.cpp
@@ -125,6 +125,10 @@ std::array<uint16_t, 2> NativeMapView::getFramebufferSize() const {
 }
 
 void NativeMapView::activate() {
+    if (active++) {
+        return;
+    }
+
     mbgl::Log::Debug(mbgl::Event::Android, "NativeMapView::activate");
 
     oldDisplay = eglGetCurrentDisplay();
@@ -151,6 +155,10 @@ void NativeMapView::activate() {
 }
 
 void NativeMapView::deactivate() {
+    if (--active) {
+        return;
+    }
+
     mbgl::Log::Debug(mbgl::Event::Android, "NativeMapView::deactivate");
 
     assert(vm != nullptr);
--- a/platform/android/src/native_map_view.hpp
+++ b/platform/android/src/native_map_view.hpp
@@ -97,6 +97,8 @@ private:
     std::unique_ptr<mbgl::DefaultFileSource> fileSource;
     std::unique_ptr<mbgl::Map> map;
     mbgl::EdgeInsets insets;
+
+    unsigned active = 0;
 };
 }
 }


tmpsantos:
:eyes: @kkaefer @ivovandongen @mikemorris

tmpsantos:
On android we still have to activate/deactivate the context because on some phones you have multiple contexts around.

But it is now "refcounted" so it supports nested invocations.

tmpsantos:
Reducing the scope just for android, gonna be one week out and this could potentially have some unwanted side effects.

kkaefer:
Why are you refcounting? Shouldn't it be something like "activate when not activated" and "deactivate when still active" and make all other operations no-ops? Rather than count the number of activations, is there a way to fix the underlying issue so that they aren't activated/deactivated in the first place?

tmpsantos:
> Why are you refcounting? Shouldn't it be something like "activate when not activated" and "deactivate when still active" and make all other operations no-ops? Rather than count the number of activations, is there a way to fix the underlying issue so that they aren't activated/deactivated in the first place?

That would not work. The buggy code was:

```
Render
    Activate
        Signal (mapped to OnLoaded on Android)
            Client calls AddLayer
                Activate
                Do some GL
                Deactivate
    Do some more GL
    Deactivate
```

So the `Do some more GL` would have a deactivated context. Another way of fixing I tried was not calling activate/deactivate at all and assume the context is always valid inside mbgl-core. We would transfer this responsibility to client/platform. What do you think?

kkaefer:
Well, we could only call Activate/Deactivate after AddLayer when the context is not active?

PR #5877, opened 2016-08-05
[android] - 5861 better compass

### LINKED ISSUE #5861: Update CompassListener logic in MyLocationView
Our current compass integration inside of MyLocationView is not really great. It was based on a older Google example found [here](https://android.googlesource.com/platform/development/+/master/samples/Compass/src/com/example/android/compass/CompassActivity.java). In meantime @cammace and I found a more recent  example used for Google glass [here](https://github.com/googleglass/gdk-compass-sample/blob/master/app/src/main/java/com/google/android/glass/sample/compass/OrientationManager.java). We should look into porting this to `MyLocationView`.

cammace:
Changed compass sensor from accelerometer and magnetometer to using the rotation vector. Also fixed bug, we were only updating the camera bearing when the user location moved. Now when `MyLocationTracking.TRACKING_FOLLOW` is enabled as well as compass, we constantly rotate the camera to point same direction as user.

Closes #5861 

cc: @zugaldia @tobrun

### DIFF
--- a/platform/android/MapboxGLAndroidSDK/src/main/java/com/mapbox/mapboxsdk/maps/widgets/MyLocationView.java
+++ b/platform/android/MapboxGLAndroidSDK/src/main/java/com/mapbox/mapboxsdk/maps/widgets/MyLocationView.java
@@ -96,6 +96,9 @@ public class MyLocationView extends View {
     private float bearing;
     private float tilt;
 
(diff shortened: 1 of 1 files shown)


tobrun:
LGTM :shipit:

cammace:
rebased, waiting for CI approval and then i'll merge

PR #5359, opened 2016-06-14
Switch build system to CMake

kkaefer:
revert #28

We want to switch to CMake from gyp because:

* gyp isn't maintained
* Android Studio has explicit CMake integration
* Qt Creator has explicit CMake integration
* CMake automatically rebuilds the project if CMake files changed
* Less complicated build system, fewer bash files

PR #5879, opened 2016-08-05
minor docs updates

incanus:
1. iOS docs bug reported via user. 
1. Update dependencies for CMake since https://github.com/mapbox/mapbox-gl-native/pull/5359 landed. (look ok @kkaefer?)

### DIFF
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -26,7 +26,7 @@ targets.
  - Modern C++ compiler that supports `-std=c++14`
    - clang++ 3.5 or later _or_
    - g++-5 or later
- - [CMake](https://cmake.org/) (for build only)
+ - [CMake](https://cmake.org/) 3.1 or later (for build only)
  - [cURL](https://curl.haxx.se) (for build only)
  - [Node.js](https://nodejs.org/) (for build only)
  - [`pkg-config`](https://wiki.freedesktop.org/www/Software/pkg-config/) (for build only)
--- a/platform/darwin/src/MGLCompassDirectionFormatter.h
+++ b/platform/darwin/src/MGLCompassDirectionFormatter.h
@@ -12,7 +12,7 @@ NS_ASSUME_NONNULL_BEGIN
  
  Use this class to create localized heading strings when displaying directions
  irrespective of the user’s current location. To format a direction relative to
- the user’s current location, use `MGLCompassDirectionFormatter` instead.
+ the user’s current location, use `MGLClockDirectionFormatter` instead.
  */
 @interface MGLCompassDirectionFormatter : NSFormatter
 


1ec5:
👍

INSTALL.md
```diff
@@ -26,6 +26,7 @@ targets.
  - Modern C++ compiler that supports `-std=c++14`
    - clang++ 3.5 or later _or_
    - g++-5 or later
+ - CMake (for build only)
  - Python 2.x (for build only)
```
kkaefer: We don't need python anymore afaik.

friedbunny:
It would also be great to verify that [the test instructions](https://github.com/mapbox/mapbox-gl-native/blob/91fd87f2a05ae310537f50d49e78c1da9b1285d6/platform/ios/DEVELOPING.md#testing) still work and are accurate.

mikemorris:
It appears that we specifically require CMake v3.x+, which on Ubuntu 12.04 requires adding a backport ppa:
```
sudo apt-add-repository ppa:george-edison55/precise-backports
sudo apt-get update
```

Attempting to install with  CMake 2.8.7 yields the following error:
```
(cd build/linux-x86_64/Debug && cmake -G Ninja ../../.. \
                -DCMAKE_BUILD_TYPE=Debug \
                -DCMAKE_EXPORT_COMPILE_COMMANDS=ON \
                -DWITH_CXX11ABI=OFF \
                -DWITH_COVERAGE=)
CMake Error: Could not create named generator Ninja
make: *** [build/linux-x86_64/Debug/build.ninja] Error 1
```

INSTALL.md
```diff
@@ -26,6 +26,7 @@ targets.
...
  - [Node.js](https://nodejs.org/) (for build only)
  - [`pkg-config`](https://wiki.freedesktop.org/www/Software/pkg-config/) (for build only)
```
mikemorris: It appears that we're no longer using `pkg-config` either? I was previously using this to ensure we were compiling and linking against a mason-installed version of Mesa, ran into trouble rebasing https://github.com/mapbox/mapbox-gl-native/commit/04fc3cb88389ee1ad607fceefd2fd2549addbb09 onto `master` because now it can't find `EGL/egl.h` (which it previously found through `pkg-config --cflags`)

kkaefer: We are not using `find_package` in CMake. However, we are using it indirectly because we are invoking `mason` (which could call `pkg-config` based on the `script.sh`) when there is no `mason.ini` file yet in the package. As we move on and more of the packages we use contain a `mason.ini` in the precompiled tarball download, we can remove the on-demand `mason.ini` creation, which removes the `pkg-config` requirement for Mapbox GL Native.

For what it's worth, we only require Python for creating the shader files and version headers, but we can easily move this to node.js, which we require anyway because of our use of `npm`.

kkaefer: created ticket for Python removal: https://github.com/mapbox/mapbox-gl-native/issues/6010

kkaefer:
Yes, we require CMake 3.1 due to features we used that are not present before 3.1.

1ec5:
@bleege committed the CMake change directly to master in bbd6cfcf8a51881c1d6879fa92e161313b513741.

bleege:
I rebased the development branch onto `master` and then added an update to explicitly say CMake 3.1 or later per @kkaefer's note.  Once CI 👍 I'll merge this into `master`.

bleege:
CI Tested and Approved.  We're documented once again!

PR #5902, opened 2016-08-08
[android] - fix scheduled build with cmake dependency

tobrun:
Fixes our daily snapshot build by adding the cmake dependency.


tobrun:
CI approved, let's rebase and merge

PR #5919, opened 2016-08-09
[android]Polyline click listener

SylvainHocq:
This is a sample evolution in PolylineActivity to show a solution to do an PolylineClickListener system.


tobrun:
Hey @SylvainHocq, thank you for reaching out and contributing, we really appreciate the contribution but we prefer handling this kind of functionality in core/c++ instead of the android/java bindings. In core we can rely on geometry objects and using intersection methods oppose doing the calculations manually. This will not only be a more performant solution but will resolve the issue for multiple platforms at once.

SylvainHocq:
Hello @tobrun, I have try this because i need this functionality.
And it seem to be enougth efficient even if it's code in Java.
The method getProximityPolyline take around 80ms, it's not so bad...

But yes i understand it's better to code this in C++ cause of more efficient and , for portability of all platform,  but i'm not very confortable in c++  :-(

So do you have an idea of when this feature will be coded?
Thanks.

tobrun:
Capturing from @1ec5 on the iOS side:
> It's on the roadmap for `iOS SDK v3.4.0`,  First we need https://github.com/mapbox/mapbox-gl-native/pull/5165, then https://github.com/mapbox/mapbox-gl-native/pull/5502 will be much simpler. 

This will normally translate to a similar timeline for Android so `4.2.0` or the release afterwards. I feel your need for this functionality since I created the #3720 a while back. Sorry for not being able to get this functionality in at this point, we aware this is highly requested feature.

PR #5928, opened 2016-08-10
[android] #5238 - marker view z ordering example

tobrun:
An example activity for Marker View Z ordering.


tobrun:
Closing as we are not moving towards the solution proposed in `5238-android-marker-view-z-ordering`.

PR #5931, opened 2016-08-10
[ios] Account for center offset when tapping annotation views

### LINKED ISSUE #5820: [iOS] AnnotationView tap detection does not take into account the center offset
**Platform: iOS**
**Mapbox SDK version: latest branch master**

### Steps to trigger behavior

 1. Create an annotation view
 2. Set a center offset different from 0,0
 3. Tap on the annotation view

### Expected behavior
tap is only detected over the annotation view

### Actual behavior
tap is detected on the annotation view original position (without offset)

boundsj:
Fixes #5820 

This adjusts the tap point for the map's single tap gesture recognizer to use the coordinate to point translated point of the actual annotation the view represents. This fixes an issue where views that were offset in the point coordinate system were not selected when tapped since their associated annotations were actually at the origin of the offset.

An alternative would've been to use a gesture recognizer on each view but we've already stated that we don't want to go down that route since it would not work well with the existing logic we have for handling single taps and finding annotations near those taps.

Note: as before, if a mixture of annotations views with and without centerOffset is used, the selection behavior may look strange due to the cycling logic we use to handle the case of overlapping annotations.

cc @friedbunny


friedbunny:
👍

PR #5938, opened 2016-08-11
Restore FBO and Viewport on GL reset

kkaefer:
When resetting the GL configuration object, we are currently not resetting the Framebuffer binding and the viewport.

### DIFF
--- a/include/mbgl/gl/gl.hpp
+++ b/include/mbgl/gl/gl.hpp
@@ -21,6 +21,7 @@
     #elif TARGET_OS_MAC
         #include <OpenGL/OpenGL.h>
         #include <OpenGL/gl.h>
+        #include <OpenGL/glext.h>
     #else
         #error Unsupported Apple platform
     #endif
--- a/include/mbgl/gl/gl_values.hpp
+++ b/include/mbgl/gl/gl_values.hpp
@@ -271,6 +271,31 @@ struct ActiveTexture {
     }
 };
 
+struct BindFramebuffer {
+    using Type = GLint;
+    static void Set(const Type& value) {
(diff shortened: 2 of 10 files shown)


jfirebaugh:
The state management makes sense, but is there a reason why we need to add `CustomLayerRenderParameters::framebuffer`?

kkaefer:
The idea of adding `framebuffer` is to allow custom layers to bind a different framebuffer, then restore the original (main) framebuffer.

jfirebaugh:
In general we don't require that custom layers restore the original GL state. Instead, we call [`config.setDirty()`](https://github.com/mapbox/mapbox-gl-native/blob/0df60c46517648b18e65d860f5a9356ba751ca11/src/mbgl/renderer/painter.cpp#L267), which ensures that our own code resets the state when necessary. Can we follow the same pattern for the framebuffer?

kkaefer:
@jfirebaugh we don't expect that they reset the state, but exposing the original framebuffer ID allows custom layers to switch to a different (offscreen) FBO, render to that FBO, then switch back to the original FBO and composite the offscreen buffer.

jfirebaugh:
Ah, I understand now. Is `glGetIntegerv(GL_FRAMEBUFFER_BINDING, &id)` too inefficient to use for that use case?

jfirebaugh:
Also, this addition would need to be plumbed through `MGLCustomStyleLayerDrawingHandler` in the iOS SDK, which I believe would be a breaking API change. Android uses the core `CustomLayerRenderParameters` struct directly, and because the addition is at the end of the struct I think the change is even ABI compatible.

kkaefer:
Calling `glGet*` typically stalls the rendering pipeline and introduces considerable delay.

kkaefer:
This was mostly merged already.

PR #5944, opened 2016-08-11
Fix compiler warning about std::move

### LINKED ISSUE #5943: Semantic issue — Moving a temporary object prevents copy elision
Following #5727 and #5935, I’m now seeing this compiler warning in the macOS workspace:

```
Semantic Issue Group
/path/to/mapbox-gl-native/platform/darwin/src/MGLStyle.mm:145:37: Moving a temporary object prevents copy elision
```

/cc @frederoni

1ec5:
Fixes #5943.

@frederoni, do you recall why this `move` was added?

### DIFF
--- a/platform/darwin/src/MGLStyle.mm
+++ b/platform/darwin/src/MGLStyle.mm
@@ -142,7 +142,7 @@ static NSURL *MGLStyleURL_emerald;
 
 - (void)addSource:(MGLSource *)source
 {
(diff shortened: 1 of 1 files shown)


frederoni:
It should've been moved instead of copied to the various `MGLSource`s.
It's now being moved twice.

Cherry-picked this commit into [#5980](https://github.com/mapbox/mapbox-gl-native/pull/5980)

PR #5986, opened 2016-08-12
[core] Compare unsigned int to unsigned int to pass tidy check

boundsj:
This fixes the travis error we are seeing (i.e. https://travis-ci.org/mapbox/mapbox-gl-native/jobs/151947796) from a test added in https://github.com/mapbox/mapbox-gl-native/pull/5165

cc @1ec5


1ec5:
👍

Issue #5489, opened 2016-06-27
Map view lags behind annotation views when panning


Paladinfeng:
*_Platform: iOS 9.3.2
*_Mapbox SDK version: v3.3.0-beta2

I found the MGLAnnotationView scrolling like jello when I use the new iOS SDK. The MGLAnnotationView can't follow the MapView smoothly. I have tested at iPhone 5s and iPhone 6.

friedbunny:
Please post more information about the view you’re using — code and a gif would be ideal. Thanks.

Paladinfeng:
ok, I'll save something as a gif video tomorrow. 
it just like rolling shutter in the field of photography.

Paladinfeng:
![rolling_shutter](https://cloud.githubusercontent.com/assets/15520106/16401810/0541ca64-3d1a-11e6-8756-b94c9ff5cfca.gif)
can u see it? The MGLAnnotationView can't follow the mapView totally.

friedbunny:
Thanks for the image, @Paladinfeng. Syncing the movement of native views with the map is an ongoing challenge, certainly.

/cc @boundsj @1ec5

1ec5:
I suspect #5245 may yield some improvement, specifically disabling animations while updating annotations and coalescing the frame updates into one Core Animation transaction. I’ll need to pull those changes out into a separate PR and see if they improve things on their own, without the other matrix-related changes.

1ec5:
The changes in #5245 don’t address this issue after all.

As far as I can tell, the map is lagging behind the annotation views by one frame every other frame or so. In the screen recording below, keep an eye on the annotation view that initially covers up the word “Washington”. When I scroll up, you can occasionally see “Washington” appear to the south of the annotation view; when I scroll down, you can occasionally see “Washington” appear to the north of the annotation view:

![gelatin](https://cloud.githubusercontent.com/assets/1231218/16543877/392697b0-40a3-11e6-8454-d48e31733d4d.gif)

1ec5:
One potential way to eliminate the "tearing" effect might be to put the annotation updating phase into a custom style layer drawing callback. This would ensure that the Core Graphics and Core Animation drawing commands occur closer to the OpenGL drawing commands and in the same run loop iteration. Ironically, the annotation views would be less performant, but the perceived performance would improve.

incanus:
Naive post to related issue, before further investigation: https://github.com/mapbox/mapbox-gl-native/pull/1813

[...]

PR #5987, opened 2016-08-13
Optimize annotation view updates

### LINKED ISSUE #5787: unnecessary performance overhead for non-view annotations on iOS
Looking separately into the perf issues mentioned in https://github.com/mapbox/mapbox-gl-native/pull/5786, it seems that `-[MGLMapView updateAnnotationViews]` is called as a result of `-[MGLMapView notifyMapChange]`. This appears to run through every annotation on every frame update, doing unnecessary logic for annotations which are sprite-backed. In my testing of an entirely sprite-backed map, I am getting about 16% of time profiling spent on rendering the whole map and _the same amount_ just to run through `-[MGLMapView convertCoordinate:toPointToView:]` when this is completely unnecessary for thousands of annotations. 

screen shot 2016-07-25 at 12 00 44 pm

/cc @1ec5 @boundsj

boundsj:
Now that https://github.com/mapbox/mapbox-gl-native/pull/5165 has landed, we can address the issues reported in https://github.com/mapbox/mapbox-gl-native/issues/5787 and again in https://github.com/mapbox/mapbox-gl-native/issues/5489#issuecomment-239492522. This PR will:
- [x] Only update annotations in that are backed by `MGLAnnotationView` subclasses
- [x] Make use of `queryPointAnnotations` (from https://github.com/mapbox/mapbox-gl-native/pull/5165) to operate on potentially less view backed annotations when updating them

Fixes #5787 

cc @incanus @1ec5

### DIFF
--- a/platform/ios/app/MBXViewController.m
+++ b/platform/ios/app/MBXViewController.m
@@ -102,11 +102,19 @@ typedef NS_ENUM(NSInteger, MBXSettingsMiscellaneousRows) {
                                  UITableViewDataSource,
                                  MGLMapViewDelegate>
 
+
 @property (nonatomic) IBOutlet MGLMapView *mapView;
+@property (weak, nonatomic) IBOutlet UILabel *hudLabel;
 @property (nonatomic) NSInteger styleIndex;
 @property (nonatomic) BOOL debugLoggingEnabled;
 @property (nonatomic) BOOL customUserLocationAnnnotationEnabled;
+
+@end
+
+@interface MGLMapView (MBXViewController)
+
 @property (nonatomic) BOOL usingLocaleBasedCountryLabels;
+@property (nonatomic) NSDictionary *annotationViewReuseQueueByIdentifier;
 
 @end
 
@@ -1506,9 +1514,23 @@ typedef NS_ENUM(NSInteger, MBXSettingsMiscellaneousRows) {
 - (void)mapView:(MGLMapView *)mapView didFinishLoadingStyle:(MGLStyle *)style
 {
     // Default Mapbox styles use {name_en} as their label language, which means
+    NSUInteger queuedAnnotations = 0;
     // that a device with an English-language locale is already effectively
+    {
     // using locale-based country labels.
+    }
     _usingLocaleBasedCountryLabels = [[self bestLanguageForUser] isEqualToString:@"en"];
+}
+
+- (void)mapViewRegionIsChanging:(MGLMapView *)mapView
+{
+    return;
+    NSUInteger queuedAnnotations = 0;
+    for (NSArray *queue in self.mapView.annotationViewReuseQueueByIdentifier.allValues)
+    {
+        queuedAnnotations += queue.count;
+    }
+    self.hudLabel.text = [NSString stringWithFormat:@"Visible: %ld  Queued: %ld", (long)mapView.visibleAnnotations.count, (long)queuedAnnotations];
 }
 
 @end
--- a/platform/ios/app/MBXViewController.m
+++ b/platform/ios/app/MBXViewController.m
@@ -109,6 +109,7 @@ typedef NS_ENUM(NSInteger, MBXSettingsMiscellaneousRows) {
 @property (nonatomic) NSInteger styleIndex;
 @property (nonatomic) BOOL debugLoggingEnabled;
 @property (nonatomic) BOOL customUserLocationAnnnotationEnabled;
+@property (nonatomic) BOOL usingLocaleBasedCountryLabels;
 @property (nonatomic) BOOL reuseQueueStatsEnabled;
 
 @end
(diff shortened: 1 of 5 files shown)


boundsj:
This first commit addresses https://github.com/mapbox/mapbox-gl-native/issues/5787 specifically. After this change, profiling shows that drawing icons at the GL level is responsible for most of the 
CPU usage when 10k annotations backed by sprites are added to the map and then the map is continuously panned.

screen shot 2016-08-12 at 5 15 17 pm

cc @incanus

boundsj:
Noting that the parts of this PR not blocked by #6055 have been PRd in https://github.com/mapbox/mapbox-gl-native/pull/6199

boundsj:
The fix of #6055 freed us up to make some long desired changes to the way annotation views are updated and queued for reuse. The summary of changes here so far is:

- Use query point annotations to create subsets of **visible** and **offscreen** annotations
- Update the centers of the visible annotations as the map is moved
- Dequeue annotations that are offscreen
- Work around and a few (what seems to be) remaining issues / limitations with symbol querying in mbgl core where the no annotations come back in the query -- sometimes this seems to happen randomly but rarely during rotation / changing zoom. It also seems to be a limitation that a query box larger than the viewport becomes increasingly problematic as pitch increases. But again, this PR works around those issues so that the performance benefit of using mbgl's query can still be used
- A performance refactor to avoid `MGLMapView annotationTagForAnnotation:` since the linear performance characteristics of that method become problematic with thousands of annotation views. The tradeoff is space complexity and another `std::map` to make annotation context retrieval time constant

When testing for 1 minute with 10,000 annotation views, high zoom (to trigger heavy reuse queue usage), and a constant panning gesture, the performance increase is significant:

### Before: total weight of ~50 seconds and `updateAnnotationViews` being ~88% of that
screenshot 2016-10-30 23 21 49

### Now: total weight of ~8 seconds and `updateAnnotationViews` being ~50% of that
screenshot 2016-10-30 23 24 44

This commit also includes an optional debugger view in iosapp to confirm the reuse queue is working as expected:

fullsizerender


This addresses concerns raised in #5787 and https://github.com/mapbox/mapbox-gl-native/issues/6173. I believe it is a more desirable and complete solution than the one proposed in https://github.com/mapbox/mapbox-gl-native/pull/6840.

cc @1ec5 @incanus @frederoni

boundsj:
Noting also that while this is not a complete fix for https://github.com/mapbox/mapbox-gl-native/issues/5489, I'm seeing much better results when scrolling since, I think, the main thread as a whole is taxed far less than before.

platform/ios/src/MGLMapView.mm
```diff
@@ -129,7 +129,9 @@ typedef NS_ENUM(NSUInteger, MGLUserTrackingState) {
 
 /// Mapping from an annotation tag to metadata about that annotation, including
 /// the annotation itself.
-typedef std::map<MGLAnnotationTag, MGLAnnotationContext> MGLAnnotationContextMap;
+typedef std::map<MGLAnnotationTag, MGLAnnotationContext> MGLAnnotationTagContextMap;
+
+typedef std::map<id<MGLAnnotation>, MGLAnnotationContext> MGLAnnotationObjectContextMap;
```
1ec5: What if we instead have a mapping from annotation objects to annotation tags? It would be an additional hop, but also less parallel state to keep in sync.

platform/ios/src/MGLMapView.mm
```diff
@@ -129,7 +129,9 @@ typedef NS_ENUM(NSUInteger, MGLUserTrackingState) {
 
 /// Mapping from an annotation tag to metadata about that annotation, including
 /// the annotation itself.
-typedef std::map<MGLAnnotationTag, MGLAnnotationContext> MGLAnnotationContextMap;
+typedef std::map<MGLAnnotationTag, MGLAnnotationContext> MGLAnnotationTagContextMap;
```
1ec5: Rename this type in the macOS implementation as well to keep the two implementations from diverging too much.

boundsj: Done in fc8d77a

platform/ios/src/MGLMapView.mm
```diff
@@ -189,6 +192,7 @@ - (void)accessibilityDecrement
     MGLAnnotationAccessibilityElement *accessibilityElement;
     MGLAnnotationView *annotationView;
     NSString *viewReuseIdentifier;
+    MGLAnnotationTag annotationTag;
```
1ec5: This is unused, right?

boundsj: Fixed in 8e67bef

platform/ios/app/MBXViewController.m
```diff
@@ -1506,9 +1527,24 @@ - (void)mapView:(MGLMapView *)mapView tapOnCalloutForAnnotation:(id <MGLAnnotati
 - (void)mapView:(MGLMapView *)mapView didFinishLoadingStyle:(MGLStyle *)style
 {
     // Default Mapbox styles use {name_en} as their label language, which means
+    NSUInteger queuedAnnotations = 0;
```
boundsj: Thanks. Fixed in https://github.com/mapbox/mapbox-gl-native/pull/6882

PR #5988, opened 2016-08-13
[android] - make query point annotations depend on screendensity

tobrun:
Corrects implementation of #5165 for Android. 
 We need to take in account screen density when we query for annotations.


tobrun:
I went ahead and merged this since the underlying logic for `onSingleTapConfirmed` had changed in #5165 and wasn't working correctly anymore. This change from #5165  will enable a lot of improvements to ViewMarkers and make things as clickable shape annotations possible.

\cc @mapbox/android

PR #5999, opened 2016-08-13
Source-driven attribution

### LINKED ISSUE #2723: Derive attribution from sources in current style
We should implement callbacks that notify the embedding applications of changes in attributions. Attributions are included in Sources, and they may change when enabling/disabling tiles from a source depending on the stylesheet, or when setting a new stylesheet that includes different sources.

### LINKED ISSUE #5775: Move Improve This Map to macOS SDK
macosapp currently implements an Improve This Map item under the Help menu. This design is HIG-compliant, and it’s where users would expect to find this functionality. But developers shouldn’t have to reinvent this code in each application they write.

[`-[MapDocument giveFeedback:]`](https://github.com/mapbox/mapbox-gl-native/blob/4b1eeada9eaa416585b8e908265240c409c1368f/platform/macos/app/MapDocument.m#L467-L472) should be implemented as an action on MGLMapView, so that developers can simply connect a menu item to the first responder’s `-giveFeedback:` action. Going a step further, maybe MGLMapView could append an Improve This Map item to the Help menu if one isn’t found. But that might be too aggressive and make too many assumptions about the host application, so let’s stick to just the action for now.

1ec5:
The attribution view on macOS and the ℹ️ action sheet on iOS now display the correct attribution based on all the sources used in the current style, instead of the Mapbox and OpenStreetMap attribution that was previously hard-coded in each implementation of MGLMapView. This means DigitalGlobe is credited for Mapbox Satellite and Satellite Streets, and attribution for third-party sources added in Mapbox Studio or using the runtime styling API is also shown.

The attribution view on macOS updates immediately whenever the style changes and a new source is loaded. Developers can (should) now hook up an Improve This Map menu item to a new `giveFeedback:` action on MGLMapView, which opens in a Web browser any feedback URL it finds in a source’s attribution HTML code.

The ℹ️ action sheet on iOS fetches the current style’s attribution information on demand.

macos

~~Implemented observer callbacks so the style knows when the source’s attribution changes and the map knows when the style’s attribution changes. Also implemented a getter for a tile source’s attribution.~~ Split out as #6431.

More to come:
- [x] ~~Attribution observer callbacks~~ (split out as #6431)
- [x] ~~Attribution accessors~~
- [x] Parse attribution HTML source code into structured data
- [x] Tests for parsing attribution strings
- [x] Dedupe attribution strings
- [x] Tests for deduping attribution strings
- [x] Update macOS attribution view dynamically (fixes the macOS side of #2723)
- [x] Use dynamic attribution in iOS attribution action sheet (fixes the iOS side of #2723)
- [x] Replace built-in “Improve This Map” link with customized/localized link from style (fixes #5775)

Depends on #6431.

/cc @kkaefer @friedbunny

### DIFF
--- a/platform/darwin/src/MGLAttributionInfo.h
+++ b/platform/darwin/src/MGLAttributionInfo.h
@@ -21,7 +21,7 @@ NS_ASSUME_NONNULL_BEGIN
  */
 + (NS_ARRAY_OF(MGLAttributionInfo *) *)attributionInfosFromHTMLString:(NSString *)htmlString fontSize:(CGFloat)fontSize linkColor:(nullable MGLColor *)linkColor;
 
-- (instancetype)initWithTitle:(NSAttributedString *)title URL:(NSURL *)URL;
+- (instancetype)initWithTitle:(NSAttributedString *)title URL:(nullable NSURL *)URL;
 
 @property (nonatomic) NSAttributedString *title;
 @property (nonatomic, nullable) NSURL *URL;
--- a/platform/darwin/src/MGLAttributionInfo.h
+++ b/platform/darwin/src/MGLAttributionInfo.h
@@ -1,4 +1,5 @@
 #import <Foundation/Foundation.h>
+#import <CoreGraphics/CoreGraphics.h>
 
 #import "MGLTypes.h"
 
@@ -15,12 +16,16 @@ NS_ASSUME_NONNULL_BEGIN
 /**
  Parses and returns the attribution infos contained in the given HTML source
  code string.
+ 
+ @param htmlString The HTML source code to parse.
+ @param fontSize The default text size in points.
+ @param linkColor The default link color.
  */
-+ (NS_ARRAY_OF(MGLAttributionInfo *) *)attributionInfosFromHTMLString:(NSString *)htmlString;
++ (NS_ARRAY_OF(MGLAttributionInfo *) *)attributionInfosFromHTMLString:(NSString *)htmlString fontSize:(CGFloat)fontSize linkColor:(nullable MGLColor *)linkColor;
 
-- (instancetype)initWithTitle:(NSString *)title URL:(NSURL *)URL;
+- (instancetype)initWithTitle:(NSAttributedString *)title URL:(NSURL *)URL;
 
-@property (nonatomic) NSString *title;
+@property (nonatomic) NSAttributedString *title;
 @property (nonatomic) NSURL *URL;
 
 @end
--- a/platform/darwin/src/MGLAttributionInfo.h
+++ b/platform/darwin/src/MGLAttributionInfo.h
@@ -3,8 +3,6 @@
 
 #import "MGLTypes.h"
 
-#include <vector>
-
 NS_ASSUME_NONNULL_BEGIN
 
 /**
@@ -27,6 +25,29 @@ NS_ASSUME_NONNULL_BEGIN
 
 @property (nonatomic) NSAttributedString *title;
 @property (nonatomic, nullable) NSURL *URL;
+
+@end
+
+@interface NSMutableArray (MGLAttributionInfoAdditions)
+
+/**
+ Adds the given attribution info object to the receiver as long as it isn’t
+ redundant to any object already in the receiver. Any existing object that is
+ redundant to the given object is replaced by the given object.
+ 
+ @param info The info object to add to the receiver.
+ @return True if the given info object was added to the receiver.
+ */
+- (void)growArrayByAddingAttributionInfo:(MGLAttributionInfo *)info;
+
+/**
+ Adds each of the given attribution info objects to the receiver as long as it
+ isn’t redundant to any object already in the receiver. Any existing object that
+ is redundant to the given object is replaced by the given object.
+ 
+ @param infos An array of info objects to add to the receiver.
+ */
+- (void)growArrayByAddingAttributionInfosFromArray:(NS_ARRAY_OF(MGLAttributionInfo *) *)infos;
 
 @end
 
--- a/platform/darwin/src/MGLAttributionInfo.h
+++ b/platform/darwin/src/MGLAttributionInfo.h
@@ -1,0 +1,28 @@
+#import <Foundation/Foundation.h>
+
+#import "MGLTypes.h"
+
+#include <vector>
+
+NS_ASSUME_NONNULL_BEGIN
+
+/**
+ Information about an attribution statement, usually a copyright or trademark
+ statement, associated with a map source.
+ */
+@interface MGLAttributionInfo : NSObject
+
+/**
+ Parses and returns the attribution infos contained in the given HTML source
+ code string.
+ */
++ (NS_ARRAY_OF(MGLAttributionInfo *) *)attributionInfosFromHTMLString:(NSString *)htmlString;
+
+- (instancetype)initWithTitle:(NSString *)title URL:(NSURL *)URL;
+
+@property (nonatomic) NSString *title;
+@property (nonatomic) NSURL *URL;
+
+@end
+
+NS_ASSUME_NONNULL_END
--- a/platform/darwin/src/MGLAttributionInfo.h
+++ b/platform/darwin/src/MGLAttributionInfo.h
@@ -26,7 +26,7 @@ NS_ASSUME_NONNULL_BEGIN
 - (instancetype)initWithTitle:(NSAttributedString *)title URL:(NSURL *)URL;
 
 @property (nonatomic) NSAttributedString *title;
-@property (nonatomic) NSURL *URL;
+@property (nonatomic, nullable) NSURL *URL;
 
 @end
 
--- a/platform/darwin/src/MGLAttributionInfo.h
+++ b/platform/darwin/src/MGLAttributionInfo.h
@@ -1,5 +1,6 @@
 #import <Foundation/Foundation.h>
 #import <CoreGraphics/CoreGraphics.h>
+#import <CoreLocation/CoreLocation.h>
 
 #import "MGLTypes.h"
 
@@ -25,6 +26,9 @@ NS_ASSUME_NONNULL_BEGIN
 
 @property (nonatomic) NSAttributedString *title;
 @property (nonatomic, nullable) NSURL *URL;
+@property (nonatomic, getter=isFeedbackLink) BOOL feedbackLink;
+
+- (nullable NSURL *)feedbackURLAtCenterCoordinate:(CLLocationCoordinate2D)centerCoordinate zoomLevel:(double)zoomLevel;
 
 @end
 
--- a/platform/darwin/src/MGLAttributionInfo.mm
+++ b/platform/darwin/src/MGLAttributionInfo.mm
@@ -64,8 +64,9 @@ static NSString * const MGLAttributionFeedbackURLString = @"https://www.mapbox.c
         }
         
         // Omit whitespace-only strings.
-        NSAttributedString *title = [attributedString attributedSubstringFromRange:range];
-        if (![title.string stringByTrimmingCharactersInSet:[NSCharacterSet whitespaceAndNewlineCharacterSet]].length) {
+        NSAttributedString *title = [[attributedString attributedSubstringFromRange:range]
+                                     mgl_attributedStringByTrimmingCharactersInSet:[NSCharacterSet whitespaceAndNewlineCharacterSet]];
+        if (!title.length) {
             return;
         }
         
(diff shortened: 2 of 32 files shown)


friedbunny:
This is a great change. 👍

Should the feedback item be contingent on the current style being Mapbox-hosted?

I also keep coming back to the UI, but adding more entries to the action sheet is something I think we should try to limit or avoid. The items that exist there now are all very important (and are all presented as equally important), but adding an arbitrary number of new items dilutes existing ones.

If a map were to have many sources and/or many attributions, this list could quickly grow out of control. A custom view is one way around this, but we could also consider following the current telemetry opt-out alert style: add a generic “Map data providers” item to the action sheet, which presents the user with an alert with items that link to the different map data providers.

1ec5:
> Should the feedback item be contingent on the current style being Mapbox-hosted?

This branch currently filters out the feedback link coming from TileJSON in favor of the built-in feedback item. On iOS, we could definitely drop the built-in item in favor of the TileJSON one, but we’ll need to customize the URL based on the current viewport and potentially localize the item title.

On macOS, the built-in “Improve This Map” item belongs in the Help menu. As part of upstreaming the item from macosapp to the macOS SDK (#5775), we could enable and disable the menu item based on the link’s presence in TileJSON. However, I’m still unsure how the SDK would even automatically add a menu item in the first place.

> adding an arbitrary number of new items dilutes existing ones

Agreed. However, in practice, the number of attribution links in a source is always kept to a minimum in order to fit horizontally on a standard webpage. For parity with GL JS, I’ve implemented deduping logic in the iOS/macOS SDK to avoid showing redundant items when, for instance, multiple Mapbox-copyrighted sources are being used. Among the default styles, Satellite and Satellite Streets are currently the worst case:

iOS

macOS

(On iOS, “Improve This Map” would swap places with the DigitalGlobe attribution if we replace the built-in feedback item with the one coming from TileJSON.)

A single “Map Data Providers” item would be more elegant, but it would divorce attribution even further from the map view. UI design changes will require considerable thought, so let’s hash that out in a separate PR.

1ec5:
macOS now supports rich formatting besides links, including inline images:

OSM logo

1ec5:
The Darwin unit tests are failing on iOS (but not macOS) due to network issues while running `-[MGLOfflineStorageTests testAAALoadPacks]`:

```
2016-08-14 12:08:30.022 xctest[78390:1809420] [ERROR] {}[Setup]: loading style failed: HTTP status code 401
```

I’m not sure why initializing an MGLOfflineStorage would trigger network requests, and none of this is related to the changes in this branch.

1ec5:
Tracking the iOS test failure in #6006.

src/mbgl/style/style.cpp
```diff
@@ -240,6 +251,26 @@ Source* Style::getSource(const std::string& id) const {
     return it != sources.end() ? it->get() : nullptr;
 }
 
+std::vector<std::string> Style::getAttributions() const {
+    std::vector<std::string> result;
+    result.reserve(sources.size());
```
tmpsantos: Not needed IMO.

tmpsantos:
:+1:

include/mbgl/map/map.hpp
```diff
@@ -149,9 +149,11 @@ class Map : private util::noncopyable {
     void removeAnnotation(AnnotationID);
 
     // Sources
+    std::vector<const style::Source*> getSources() const;
```
tmpsantos: We don't need this API yet, do we?

src/mbgl/style/style.cpp
```diff
@@ -232,6 +234,15 @@ void Style::recalculate(float z, const TimePoint& timePoint, MapMode mode) {
     }
 }
 
+std::vector<const Source*> Style::getSources() const {
```
tmpsantos: Is this needed?

src/mbgl/style/style.cpp
```diff
@@ -240,6 +251,26 @@ Source* Style::getSource(const std::string& id) const {
...
+            result.push_back(tileSource->getAttribution());
+        }
+            
+        case SourceType::GeoJSON:
+        case SourceType::Video:
+        case SourceType::Annotations:
+            break;
+        }
+    }
+    return result;
```
tmpsantos: You can probably `std::move()` the result.

src/mbgl/style/style.cpp
tmpsantos: To clarify, this will always allocated more memory than we need since we always have a annotation source that doesn't have attribution.

src/mbgl/map/map.cpp
```diff
@@ -861,6 +870,10 @@ void Map::onLowMemory() {
     impl->view.invalidate();
 }
 
+void Map::Impl::onSourceAttributionChanged(style::Source&, std::string) {
```
tmpsantos: Why the parameters if not in use?

src/mbgl/style/style.cpp
brunoabinader: Wouldn't `std::move()` prevent [RVO](https://en.wikipedia.org/wiki/Return_value_optimization)?

test/src/mbgl/test/stub_style_observer.hpp
```diff
@@ -34,6 +34,10 @@ class StubStyleObserver : public style::Observer {
         if (sourceError) sourceError(source, error);
     }
 
+    void onSourceAttributionChanged(Source& source, std::string attribution) override {
```
tmpsantos: Neat, would be cool if we could have some utests for this.

src/mbgl/map/map.cpp
1ec5: I struggled with this one. Semantically, I think it makes plenty of sense for the string to go along with the callback, for consistency with the other callbacks. But we aren’t currently plumbing the string all the way through because the `notifyMapChange()` mechanism doesn’t support additional data beyond a single `MapChange` enum.

src/mbgl/style/style.cpp
```diff
@@ -240,6 +242,28 @@ Source* Style::getSource(const std::string& id) const {
...
+        switch (source->baseImpl->type) {
+        case SourceType::Vector:
+        case SourceType::Raster: {
+            style::TileSourceImpl* tileSource = static_cast<style::TileSourceImpl*>(source->baseImpl.get());
+            auto attribution = tileSource->getAttribution();
+            if (!attribution.empty()) {
+                result.push_back(attribution);
+            }
+        }
+
```
tmpsantos: Spaces.

src/mbgl/style/style.cpp
tmpsantos: True, I take it back. :)

test/src/mbgl/test/stub_style_observer.hpp
1ec5: Added a test case to source.cpp.

1ec5:
A Valgrind failure on Qt appears to have been uncovered by the addition of a simple unit test:

```
==10075== Conditional jump or move depends on uninitialised value(s)

==10075==    at 0x4C307C2: __memcmp_sse4_1 (in /home/travis/build/mapbox/mapbox-gl-native/mason_packages/linux-x86_64/valgrind/latest/lib/valgrind/vgpreload_memcheck-amd64-linux.so)

==10075==    by 0x75AB4E: mbgl::util::mapbox::isMapboxURL(std::string const&) (in /home/travis/build/mapbox/mapbox-gl-native/build/qt-linux-x86_64/Release/mbgl-test)

==10075==    by 0x6A8259: mbgl::style::TileSourceImpl::parseTileJSON(std::string const&, std::string const&, mbgl::SourceType, unsigned short) (in /home/travis/build/mapbox/mapbox-gl-native/build/qt-linux-x86_64/Release/mbgl-test)

==10075==    by 0x6A9F1D: std::_Function_handler<void (mbgl::Response), mbgl::style::TileSourceImpl::load(mbgl::FileSource&)::{lambda(mbgl::Response)#1}>::_M_invoke(std::_Any_data const&, mbgl::Response&&) (in /home/travis/build/mapbox/mapbox-gl-native/build/qt-linux-x86_64/Release/mbgl-test)
```

The test uses `test` as a tile URL, just like some of the other Source tests.

tmpsantos:
@1ec5 what about we split this in two PRs [core] + [osx/ios]? This way I can do the Qt work in parallel.

1ec5:
OK, I split out the core parts as #6431.

1ec5:
#6431 landed with the core support I split out. There will be an additional warning or two in master until this branch lands.

1ec5:
Since `mbgl::Map::getAttributions()` was removed in #6465, this PR is now blocked by #6097, which would provide a way to enumerate the current style’s layers, whether in mbgl or at the SDK level.

1ec5:
#6097 has landed, so this PR is unblocked.

1ec5:
The iOS test failure is still here, and it appears to be an uncaught exception on [this line](https://github.com/mapbox/mapbox-gl-native/blob/930801b9b6be04bb277da79dddf427473614fc23/platform/darwin/src/MGLOfflineStorage.mm#L199) that’s being hidden by the active XCTestExpectation:

```
2016-11-29 11:47:49.754 xctest[63439:4496363] *** Assertion failure in -[_XCKVOExpectation fulfill], /Library/Caches/com.apple.xbs/Sources/XCTest_Sim/XCTest-11753/XCTestFramework/Classes/XCTestExpectation.m:99
```

```
#0	0x000000010dfa91ee in objc_exception_throw ()
#1	0x000000010e54be42 in +[NSException raise:format:arguments:] ()
#2	0x000000010db3e66d in -[NSAssertionHandler handleFailureInMethod:object:file:lineNumber:description:] ()
#3	0x000000010d9ccfd4 in -[XCTestExpectation fulfill] ()
#4	0x000000010d9cf1e3 in -[_XCKVOExpectation observeValueForKeyPath:ofObject:change:context:] ()
#5	0x000000010daaadec in NSKeyValueNotifyObserver ()
#6	0x000000010daaa687 in NSKeyValueDidChange ()
#7	0x000000010dabceb5 in -[NSObject(NSKeyValueObserverNotification) didChange:valuesAtIndexes:forKey:] ()
#8	0x000000010db2c79f in -[NSKeyValueNotifyingMutableArray addObject:] ()
#9	0x000000011a5226b4 in ::__68-[MGLOfflineStorage addPackForRegion:withContext:completionHandler:]_block_invoke(MGLOfflinePack *, NSError *) at /Users/mxn/hub/mapbox-gl-native-2/platform/darwin/src/MGLOfflineStorage.mm:199
#10	0x000000011a526ec4 in -[MGLOfflineStorage _addPackForRegion:withContext:completionHandler:]::$_0::operator()(std::exception_ptr, std::experimental::fundamentals_v1::optional<mbgl::OfflineRegion>) const::'lambda'()::operator()() const at /Users/mxn/hub/mapbox-gl-native-2/platform/darwin/src/MGLOfflineStorage.mm:228
#11	0x000000011a526e99 in ::___ZZ69-[MGLOfflineStorage _addPackForRegion:withContext:completionHandler:]ENK3$_0clESt13exception_ptrNSt12experimental15fundamentals_v18optionalIN4mbgl13OfflineRegionEEE_block_invoke() at /Users/mxn/hub/mapbox-gl-native-2/platform/darwin/src/MGLOfflineStorage.mm:227
#12	0x000000011124e808 in _dispatch_call_block_and_release ()
#13	0x000000011127012e in _dispatch_client_callout ()
#14	0x00000001112574fb in _dispatch_main_queue_callback_4CF ()
#15	0x000000010e50be49 in __CFRUNLOOP_IS_SERVICING_THE_MAIN_DISPATCH_QUEUE__ ()
#16	0x000000010e4d137d in __CFRunLoopRun ()
#17	0x000000010e4d0884 in CFRunLoopRunSpecific ()
#18	0x000000010d9af266 in -[XCTestExpectationWaiter primitiveWait:] ()
#19	0x000000010d9aec47 in -[XCTestExpectationWaiter wait:forExpectations:enforceOrder:] ()
#20	0x000000010d9ce859 in -[XCTestCase(AsynchronousTesting) waitForExpectationsWithTimeout:handler:] ()
#21	0x000000011a23620d in -[MGLOfflineStorageTests testAddPack] at /Users/mxn/hub/mapbox-gl-native-2/platform/darwin/test/MGLOfflineStorageTests.m:67
```

1ec5:
A fix for the iOS test failure is ready in #7228. This PR, meanwhile, is ready for review.

1ec5:
Feedback links are now derived from the current style’s sources instead of being hard-coded, as suggested in https://github.com/mapbox/mapbox-gl-native/pull/5999#issuecomment-239639757. So in iOS applications displaying the Mapbox Satellite Streets style, the Improve This Link button in the ℹ️ sheet appears between the OpenStreetMap and DigitalGlobe attribution buttons. macOS applications can easily create a similar menu item by hooking it up to the first responder’s `giveFeedback:` action, which MGLMapView validates.

This PR is ready for review. The remaining test failure is due to mbgl test coverage declining on the release branch independently of these changes.

1ec5:
I was thinking about making `MGLAttributionInfo` public and turning `MGLTileSet.attribution` into an array of them, but I’ll address that separately in #7361.

PR #6007, opened 2016-08-14
Replace testAAALoadPacks with a setUp method

### LINKED ISSUE #6006: -[MGLOfflineStorageTests testAAALoadPacks] fails
`-[MGLOfflineStorageTests testAAALoadPacks]` – so named because it’s supposed to run before all the other tests in the test case – is running last for some reason, leading to the following failure on iOS (but not macOS for some reason):

```
Test Case '-[MGLOfflineStorageTests testAAALoadPacks]' started.
/Users/mxn/hub/mapbox-gl-native-3/platform/darwin/test/MGLOfflineStorageTests.m:22: error: -[MGLOfflineStorageTests testAAALoadPacks] : Asynchronous wait failed: Exceeded timeout of 1 seconds, with unfulfilled expectations: "Expect change to packs of <MGLOfflineStorage: 0x7fcacbf0edd0>".
Test Case '-[MGLOfflineStorageTests testAAALoadPacks]' failed (1.030 seconds).
Test Suite 'MGLOfflineStorageTests' failed at 2016-08-14 12:13:45.269.
```

The failure just started happening on Bitrise last night for some reason, but I can reproduce it locally.

/cc @friedbunny @boundsj

1ec5:
XCTest apparently no longer executes tests in alphabetical order, so naming the test `testAAALoadPacks` doesn’t ensure it gets run first. Instead, put the pack loading test inside `-setUp`, but also put it inside a `dispatch_once()` block to ensure that it only gets run before the first offline storage test that gets run in the current process.

Fixes #6006.

/cc @friedbunny @boundsj

### DIFF
--- a/platform/darwin/test/MGLOfflineStorageTests.m
+++ b/platform/darwin/test/MGLOfflineStorageTests.m
@@ -8,20 +8,24 @@
 
 @implementation MGLOfflineStorageTests
 
-- (void)testSharedObject {
-    XCTAssertEqual([MGLOfflineStorage sharedOfflineStorage], [MGLOfflineStorage sharedOfflineStorage], @"There should only be one shared offline storage object.");
+- (void)setUp {
+    [super setUp];
+    
+    static dispatch_once_t onceToken;
+    dispatch_once(&onceToken, ^{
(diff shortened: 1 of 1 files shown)


friedbunny:
👍

PR #6010, opened 2016-08-15
Remove Python dependency from build

kkaefer:
Now that we have gotten rid of `gyp`, we can eliminate the Python dependency altogether. The remaining python files are

- [x] `scripts/build-shaders.py`
- [x] `scripts/build-version.py`
- [x] `scripts/export-xcconfig.py`

and maybe

- [ ] `platform/android/tests/script/devicefarm.py`

We can replace them with node.js scripts, given that we require node.js to be present anyway.


kkaefer:
Not removing `devicefarm.py` given https://github.com/mapbox/mapbox-gl-native/pull/6010#issuecomment-239775079

PR #6002, opened 2016-08-15
Parse and expose more style JSON root properties

kkaefer:
[`mbgl::style::Parser::parse()`](https://github.com/mapbox/mapbox-gl-native/blob/master/src/mbgl/style/parser.cpp#L23) should parse the [`name`](https://www.mapbox.com/mapbox-gl-style-spec/#root-name) and [`metadata`](https://www.mapbox.com/mapbox-gl-style-spec/#root-metadata) root properties out of the style JSON and expose them with accessors on the `mbgl::style::Style` class. The SDK could read the name of the style aloud in accessibility contexts.

Similarly, root properties such as [`center`](https://www.mapbox.com/mapbox-gl-style-spec/#root-center) and [`zoom`](https://www.mapbox.com/mapbox-gl-style-spec/#root-zoom) may be useful for establishing a default viewport for the style other than the Null Island default used in the SDK currently. This would enable the developer to set the default viewport once, in Mapbox Studio, and not have to set it again programmatically or in a storyboard.

/cc @jfirebaugh @kkaefer

### DIFF
--- a/include/mbgl/map/map.hpp
+++ b/include/mbgl/map/map.hpp
@@ -158,6 +158,13 @@ public:
     void addLayer(std::unique_ptr<style::Layer>, const optional<std::string>& beforeLayerID = {});
     void removeLayer(const std::string& layerID);
 
+    // Defaults
+    std::string getStyleName() const;
+    LatLng getDefaultLatLng() const;
+    double getDefaultZoom() const;
+    double getDefaultBearing() const;
+    double getDefaultPitch() const;
+
     // Feature queries
     std::vector<Feature> queryRenderedFeatures(const ScreenCoordinate&, const optional<std::vector<std::string>>& layerIDs = {});
     std::vector<Feature> queryRenderedFeatures(const ScreenBox&,        const optional<std::vector<std::string>>& layerIDs = {});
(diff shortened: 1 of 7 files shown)


kkaefer:
@1ec5 This patch parses the properties, but they aren't yet exposed in the SDK. You can obtain the values with `Map::getName()`, `Map::getDefaultLatLng()`, `Map::getDefaultZoom()`, `Map::getDefaultBearing()`, and `Map::getDefaultPitch()`.

Do you have any specific use case for `metadata` values?

include/mbgl/map/map.hpp
```diff
@@ -158,6 +158,13 @@ class Map : private util::noncopyable {
     void addLayer(std::unique_ptr<style::Layer>, const optional<std::string>& beforeLayerID = {});
     void removeLayer(const std::string& layerID);
 
+    // Defaults
+    std::string getName() const;
```
1ec5: Should this be getStyleName()?

1ec5:
> Do you have any specific use case for metadata values?

No, I figured this would eventually be opened up to developers, but we can definitely leave it out, especially since the structure is undefined.

1ec5:
:+1:

include/mbgl/map/map.hpp
kkaefer: Yeah; renamed.

PR #2270, opened 2016-08-15
Raster tile overlay with TMS coordinates (non-flipped Y)

kkaefer:
Hi,
We want to provide our own raster overlay for a certain area but MapBox flips the y coordinate while our tile service does not. We are providing a `style.json` and a `tileset.json` file like this:

style.json

        {
      "version": 7,
      "sources": {
        "simple-tiles": {
          "type": "raster",
          "url": "http://localhost:9000/tileset.json",
          "tileSize": 256
        }
      },
      "layers": [
        {
          "id": "background",
          "type": "background",
          "paint": {
            "background-color": "rgb(140,215,225)"
          }
        },
        {
          "id": "simple-tiles",
          "type": "raster",
          "source": "simple-tiles"
        }
      ]
    }

tileset.json

    {
      "tiles": [
        "http://localhost:9000/proxy.php?z={z}&x={x}&y={y}"
      ],
      "tilejson": "2.1.0",
      "scheme": "tms"
    }

The proxy-php currently only flips the y-coordinate back, we want to get rid of the proxy.php file. How can we achieve non-flipped-y tile loading?

### DIFF
--- a/include/mbgl/storage/resource.hpp
+++ b/include/mbgl/storage/resource.hpp
@@ -3,6 +3,7 @@
 #include <mbgl/storage/response.hpp>
 #include <mbgl/util/optional.hpp>
 #include <mbgl/util/font_stack.hpp>
+#include <mbgl/util/tileset.hpp>
 
 #include <string>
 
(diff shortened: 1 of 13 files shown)


1ec5:
👍 Would you do the honors and add a blurb to the iOS and macOS changelogs?

kkaefer:
JS-side PR was already merged: https://github.com/mapbox/mapbox-gl-js/pull/2565

Issue #6017, opened 2016-08-15
allow for more efficient batch runtime style API changes


incanus:
We want a transaction block model such as `CATransaction` which will defer calls to `-[MGLBaseStyleLayer update]` until the transaction is completed, allowing for many setters to be called which are all made effective with one `-update` call. 

A model for design here might be a singleton which counts to `1` when the transaction `begin`s, decrements to `0` when the transaction `commit`s, and throws an exception if it is ever tried to be nested with a secondary or later `begin` (though, that would probably be harmless—`commit` could just always reset to `0`). 

/cc @frederoni @1ec5

1ec5:
(Spun out from #5957.)

1ec5:
> A model for design here might be a singleton which counts to `1` when the transaction `begin`s, decrements to `0` when the transaction `commit`s, and throws an exception if it is ever tried to be nested with a secondary or layer `begin`.

With a block-based design, we can allow the counter to increment past 1 and only update when it falls to 0. Nesting transactions should be no problem.

frederoni:
Perfect case for the recently introduced class properties:
```
@property (class) NSUInteger batchedTransactions;
```
Though, it's unfortunate that we can't make use of this yet as it would mean dropping support for Xcode 7.x

jfirebaugh:
I don't think we want an explicit transactional API; we had that in gl-js but replaced it with automatic batching, which works well and requires less API surface area (https://github.com/mapbox/mapbox-gl-js/pull/2355).

jfirebaugh:
Automatic batching is part of #6201, I'm going to mark this to close when that's merged.

Issue #5952, opened 2016-08-11
Rename Objective-C++ headers back to .h


1ec5:
Xcode refuses to highlight Objective-C keywords in .hpp files. Indexing is also broken. Let’s rename the Objective-C++ headers back to .h.

/ref https://github.com/mapbox/mapbox-gl-native/pull/5727#discussion_r71796994
/cc @frederoni

1ec5:
In #5989, we inadvertently exposed an Objective-C++ header publicly, even though that was the problem we were trying to avoid with the .hpp extension. Given the downsides, I’m going ahead and renaming back to .h.

PR #6018, opened 2016-08-15
iOS immediate style 😎

### LINKED ISSUE #5957: Runtime styling changes should take effect immediately
Per https://github.com/mapbox/mapbox-gl-native/pull/5727#commitcomment-18577738 and https://github.com/mapbox/mapbox-gl-native/pull/5727#discussion_r74302320, changes to layer and source attributes should take effect immediately. We should offer a block-based transaction API for cases in which the developer needs to change multiple attributes in tandem.

/cc @incanus @frederoni

incanus:
Fixes #5957 by introducing an automatic call to `-[MGLBaseStyleLayer update]` for all setters. 

https://github.com/mapbox/mapbox-gl-native/issues/6017 tracks a more efficient transaction block mechanism, which doesn't necessarily have to be a part of the same release. 

/cc @frederoni @1ec5

### DIFF
--- a/platform/darwin/scripts/generate-style-code.js
+++ b/platform/darwin/scripts/generate-style-code.js
@@ -100,22 +100,31 @@ global.initLayer = function (layerType) {
 }
 
 global.setterImplementation = function(property, layerType) {
+    var implementation = '';
     switch (property.type) {
         case 'boolean':
-            return `self.layer->set${camelize(property.name)}(${objCName(property)}.mbgl_boolPropertyValue);`;
+            implementation = `self.layer->set${camelize(property.name)}(${objCName(property)}.mbgl_boolPropertyValue);`;
+            break;
         case 'number':
-            return `self.layer->set${camelize(property.name)}(${objCName(property)}.mbgl_floatPropertyValue);`;
+            implementation = `self.layer->set${camelize(property.name)}(${objCName(property)}.mbgl_floatPropertyValue);`;
+            break;
         case 'string':
-            return `self.layer->set${camelize(property.name)}(${objCName(property)}.mbgl_stringPropertyValue);`;
+            implementation = `self.layer->set${camelize(property.name)}(${objCName(property)}.mbgl_stringPropertyValue);`;
+            break;
         case 'enum':
             var objCType = `${prefix}${camelize(layerType)}${suffix}${camelize(property.name)}`;
-            return `MGLSetEnumProperty(${objCName(property)}, ${camelize(property.name)}, ${mbglType(property)}, ${objCType});`; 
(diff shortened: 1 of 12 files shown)


1ec5:
There should no longer be a use case for calling `-[MGLBaseStyleLayer update]`, so that method should be made private.

incanus:
Ok, attempted `-update` privacy in https://github.com/mapbox/mapbox-gl-native/pull/6018/commits/8f98d836000c115738cee2d07837cc518d970038 @1ec5. Curious about your thoughts on whether we should have a separate `MGLBaseStyleLayer_Private.hpp` for this or if this suffices. I originally tried something with an additionally protocol that the child layer types could adopt, but that doesn't allow them to call back to `super`'s implementation, but rather warns that they don't implement it themselves.

1ec5:
MGLBaseStyleLayer_Private should go in a separate MGLBaseStyleLayer_Private.h file. Also, note that I’m about to go through and rename the .hpp files to .h: #5952.

incanus:
Feel good about this then @1ec5 @frederoni? If so I'll squash & merge. I think it's in 🆗 shape.

1ec5:
Looks good. Interesting that no change was needed in MapDocument.m. I take that as evidence that developers would likely forget to call `-update`, so this is a good API change to make.

frederoni:
Adding/removing a source or a layer also recalculates styles and classes, that's why no `-update` call was needed in MapDocument.m

This looks good to me 👍

incanus:
Dang, keep forgetting the `[ios]`; sorry about that. Next time!

1ec5:
MGLBaseStyleLayer_Private.hpp wasn’t added to ios.xcodeproj or macos.xcodeproj. The projects build, because it was only included privately anyways, but we should add it to the projects.

1ec5:
> MGLBaseStyleLayer_Private.hpp wasn’t added to ios.xcodeproj or macos.xcodeproj.

Fixed in #6020.

incanus:
Oh weird; I didn't realize we did this manually now vs. getting picked up by the build system based on location and file extension.

1ec5:
CMake does automatically pull core files into the project it generates, but the SDK-level projects are checked into the repository, so we use the standard Xcode workflow for that. Feel free to use Xcode’s New File sheet etc.

kkaefer:
> CMake does automatically pull core files into the project it generates

No, it doesn't. If you add/remove files, we have a convenience script at `./scripts/generate-core-files.sh` and `./scripts/generate-test/files.sh` that updates the list.

1ec5:
I stand corrected. 😀 That's actually better than what we had with GYP, because we had to manually add SDK files to the .gypi file. (It was only automatic based on the file extension and location for core files and for resource files. But never for SDK source files or headers, because Xcode's build system doesn't support target membership based on folder references.)

PR #5882, opened 2016-08-06
Refactored MGLUserLocationAnnotationView (redux)

friedbunny:
This is a replacement PR for https://github.com/mapbox/mapbox-gl-native/pull/5573 that targets `master`.

Additionally, I’ve rewritten https://github.com/mapbox/mapbox-gl-native/pull/5816 so that the hit testing layer can be set by any subclass of `MGLUserLocationAnnotationView`.

/cc @frederoni @1ec5 @incanus @boundsj

PR #6019, opened 2016-08-15
Improve user location annotation performance

friedbunny:
User location annotation performance wasn’t bad previously, but there are some small tweaks we can make to improve both perceived and actual performance.

This PR is proposed against `5039-MGLUserLocationAnnotationView-refactor` until #5882 lands. We can now [change the base branch of existing pull requests](https://github.com/blog/2224-change-the-base-branch-of-a-pull-request), so I’ll try that out before merging this.

### Accuracy ring
The biggest improvement is to the accuracy ring: instead of queueing up implicit animations while zooming, now we only animate changes to the ring’s size (i.e., when the horizontal accuracy changes). This fixes the wobbling effect you could see at the edges of the ring when zooming.

We’re able to disable the implicit animation during zooms because we already update the user dot ourselves in `-[MGLMapView updateUserLocationAnnotationViewAnimatedWithDuration:]` — sometimes almost every frame.

### Shadow paths
Per [Apple](https://developer.apple.com/library/ios/documentation/Cocoa/Conceptual/CoreAnimation_guide/ImprovingAnimationPerformance/ImprovingAnimationPerformance.html), pre-calculating the paths of shadows can provide better performance.

### Rasterize pulsating inner dot
This stops the dot from being rendered off-screen and smoothes its edges (which could at times appear squarish).

/cc @1ec5 @boundsj @frederoni @incanus

### DIFF
--- a/platform/ios/src/MGLFaux3DUserLocationAnnotationView.m
+++ b/platform/ios/src/MGLFaux3DUserLocationAnnotationView.m
@@ -148,6 +148,7 @@ const CGFloat MGLUserLocationAnnotationArrowSize = MGLUserLocationAnnotationPuck
         _puckDot.backgroundColor = [[UIColor whiteColor] CGColor];
         _puckDot.shadowColor = [[UIColor blackColor] CGColor];
         _puckDot.shadowOpacity = 0.25;
+        _puckDot.shadowPath = [[UIBezierPath bezierPathWithOvalInRect:_puckDot.bounds] CGPath];
 
         if (self.mapView.camera.pitch)
         {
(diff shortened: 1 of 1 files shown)


platform/ios/src/MGLFaux3DUserLocationAnnotationView.m
```diff
@@ -291,13 +292,21 @@ - (void)drawDot
         if (accuracyRingSize > MGLUserLocationAnnotationDotSize + 15)
         {
             _accuracyRingLayer.hidden = NO;
+
+            id shouldNotAnimateAccuracyRingChange = (_oldHorizontalAccuracy == self.userLocation.location.horizontalAccuracy) ? (id)kCFBooleanTrue : (id)kCFBooleanFalse;
```
1ec5: Nit: always express variable names in the positive, for easier reasoning. How about `shouldDisableActions`, along with a comment?

friedbunny: Sounds good to me... fought with myself a bit here and I like your solution. 👍

friedbunny:
Changing the target branch was a bit rougher than I hoped — it didn’t recognize that #5882 had been squash-merged and showed all of the squashed commits again here.

Fixed and rebased onto master. Will merge after tests go green.

friedbunny:
Merged without Travis (re-)approval because Travis is having a major outage today.

PR #6028, opened 2016-08-16
Remove MGLSource.sourceType

### LINKED ISSUE #6027: Remove MGLSource.sourceType
MGLSource’s `sourceType` property is unnecessary, because the same information can be ascertained through a runtime `isKindOfClass` check. (Besides, it should be an enumeration, not a string.) The property should be removed.

/cc @frederoni

1ec5:
Fixes #6027.

/cc @frederoni


frederoni:
👍

Issue #6014, opened 2016-08-15
add test for Android runtime styling non-string type filters


incanus:
https://github.com/mapbox/mapbox-gl-native/pull/5979 was meant to fix filters of this sort: 

```java
LineLayer segments = mapboxMap.getLayerAs("segments");
segments.setFilter(eq("uniqid", new Integer(10191020)));
segments.setProperties(lineColor(Color.GREEN),
    lineWidth(10f),
    lineOpacity(0.5f));
```

However, it still doesn't seem to work (nor does `eq("uniqid", 10191020)`). Regardless, we should add a test scenario for a non-string type filter so that [this one](https://github.com/mapbox/mapbox-gl-native/blob/52134eb8fd99527e084c6fc9ec2926057c593d4d/platform/android/MapboxGLAndroidSDKTestApp/src/main/java/com/mapbox/mapboxsdk/testapp/activity/style/RuntimeStyleActivity.java#L248) isn't the only thing being exercised (as well as fix the core issue, if it's really still happening). 

/cc @ivovandongen

ivovandongen:
@incanus See the example with numerical filters here: https://github.com/mapbox/mapbox-gl-native/blob/master/platform/android/MapboxGLAndroidSDKTestApp/src/main/java/com/mapbox/mapboxsdk/testapp/activity/style/GeoJsonClusteringActivity.java#L170

But I'll happily add another one.

ivovandongen:
Merged example

PR #6030, opened 2016-08-16
[android] #6014 - add example with numerical filter

ivovandongen:
Adds an example on `RuntimeStyleActivity` with numerical filters.

@tobrun review please


tobrun:
LGTM 🚢

ivovandongen:
cc @mapbox/android

platform/android/MapboxGLAndroidSDKTestApp/src/main/java/com/mapbox/mapboxsdk/testapp/activity/style/RuntimeStyleActivity.java
```diff
@@ -244,8 +244,8 @@ private void addParksLayer() {
                 fillAntialias(true)
         );
 
-        //Only show me parks
-        layer.setFilter(eq("type", "park"));
+        //Only show me parks (except westerpark with stroke-width == 3)
```
zugaldia: @ivovandongen what do you mean with `except westerpark`, is it because we don't have that data in the layer, or is it because there's a filter to exclude that park somewhere else?

Issue #4974, opened 2016-05-09
Shape overlays are unable to cross antimeridian


1ec5:
A polyline that straddles the antimeridian ends up taking the long way around the globe:

```objc
CLLocationCoordinate2D coords[2] = {
    CLLocationCoordinate2DMake(0, -170),
    CLLocationCoordinate2DMake(0, 170),
};
MGLPolyline *polyline = [MGLPolyline polylineWithCoordinates:coords count:sizeof(coords) / sizeof(coords[0])];
[mapView addOverlay:polyline];
```

![mapbox](https://cloud.githubusercontent.com/assets/1231218/15105013/bdcc44a8-1572-11e6-8b0e-089e17ab62a4.png)

The corresponding MapKit code does straddle the antimeridian:

![mapkit](https://cloud.githubusercontent.com/assets/1231218/15105034/e7582454-1572-11e6-9ddc-a1bc0c07e44f.png)

Also reproduces with polygons.

This is a very old bug that has reproduced in the iOS SDK since shape annotations were first implemented in #1655, but I was unable to find a ticket that adequately addresses it. The diagnosis in https://github.com/mapbox/mapbox-gl-native/issues/3879#issuecomment-183149498 probably applies here, even though point annotations are apparently unaffected as of #4285.

/cc @jfirebaugh @tobrun @bleege

bleege:
Confirming that this also impacts the Android SDK as the corresponding code produces a similar screenshot.

```java
// Android Binding Code
List<LatLng> points = Arrays.asList(new LatLng(0, -170), new LatLng(0, 170));
mapboxMap.addPolyline(new PolylineOptions().addAll(points));
```

This translates to the Core GL level by `MapboxMap.addPolyline()` calling `MapView.addPolyline()` which in turn calls [`addShapeAnnotations()` in Core GL to do the rendering](https://github.com/mapbox/mapbox-gl-native/blob/7d1a4c5ec146493bb3e192f7fc20e3ae3b626011/platform/android/src/jni.cpp#L806-L838).

![android-antimeridian-polyline](https://cloud.githubusercontent.com/assets/989928/15124005/e70016e6-15eb-11e6-98b7-4a43a3941a5a.png)

tmcw:
Can you do

```objc
CLLocationCoordinate2D coords[2] = {
    CLLocationCoordinate2DMake(0, -170),
    CLLocationCoordinate2DMake(0, -190),
};
[...]

Issue #5600, opened 2016-07-06
circle-radius/antimeridian test is failing


jfirebaugh:
Revealed during the upgrade to pixelmatch4. It was passing before because the old pixelmatch had a bug where black vs. white differences weren't detected.

![image](https://cloud.githubusercontent.com/assets/98601/16638219/6b97b9c8-439a-11e6-8b2a-777c8a2fdc1a.png)

brunoabinader:
This looks similar to the symbol clip issue in https://github.com/mapbox/mapbox-gl-native/issues/5419.

mourner:
The real reason is that circles are explicitly not clipped to tile boundary in GL JS https://github.com/mapbox/mapbox-gl-js/blob/7ec9bbba94f60f188c98b0032aca57371be285fa/js/render/draw_circle.js#L15-L17, unlike the GL Native behavior. 

Should we disable clipping in GL Native too, or the other way around? I'm not sure the "leaking" off the tile is the correct behavior here. cc @tmcw

mourner:
OK, it looks like the real reason is not what I initially thought — stencil test is disabled for circles on native as well; however, for the 512x512 zoom 0 render test which is supposed to load z0 tile, GL JS render suite also loads adjacent "wrapped" tiles `-1,0,0` and `1,0,0`, and GL Native does not.

mourner:
Heh, I found out that the real reason is most likely the fact we're using `MapMode::Still` for render tests, which disables leaking for circles and symbols. https://github.com/mapbox/mapbox-gl-native/blob/master/src/mbgl/renderer/painter_circle.cpp#L22

PR #6043, opened 2016-08-16
[WIP] Add solution for shape overlays to cross antimeridian

boundsj:
#4974 points out how GL Native cannot support a shape overlay that crosses the antimeridian. This PR is to provide support that use case by:

- [x] Adding tests that describe the problem
- [x] Changing Core GL to support shapes going over the antimeridian

The change in core should be a continuation of the work done in https://github.com/mapbox/mapbox-gl-native/commit/026b6d4c01a3d96af9629cc1790373137ecab950 which fixed several issues related to points surrounding the antimeridian but removed support for expressing shapes crossing the antimerdian with longitudinal values +-180º.


boundsj:
The diff in the new `AntimeridianAnnotationSmall` test illustrates the case we would like to support but currently cannot. A shape with points greater than or less than +- 180º is not fully rendered. It is cut off at the antimeridian.

diff

cc @brunoabinader

brunoabinader:
This looks related: https://github.com/mapbox/mapbox-gl-native/issues/5600

brunoabinader:
Another observation: the full triangle is drawn when we duplicate each polyline coordinate with its 360 degrees negative counterpart, so the following workaround draws the full small triangle:
```c++
mbgl::Polygon<double> polygon = {{
    {{ { 190, 0 }, { 170, 10 }, { 170, -10 } }},
    {{ { -170, 0 }, { -190, 10 }, { -190, -10 } }}
}};
```

mourner:
The triangle bug above will be fixed by https://github.com/mapbox/geojson-vt-cpp/pull/57

brunoabinader:
Cherry-picked this commit in #6088 with a minor update to the expectation from `AntimeridianAnnotationLarge`: because we are now properly painting on both sides, the red line fill for the right tile now appears (you can notice it is bolder now):
expected

PR #5942, opened 2016-08-17
Merge iproj and xproj?

kkaefer:
#5727 is showing that it’s difficult to keep track of shared sources between the iOS and macOS SDKs. Should we combine ios.xcworkspace and macos.xcworkspace (which also hosts the Node schemes) into one workspace that contains both ios.xcodeproj and macos.xcodeproj? Should we go a step further and combine the .xcodeproj files, so that a file can directly be targeted at either platform or both platforms as needed? We do this in the Mapbox Swift library projects without any notable problems. It would also make it much easier to stand up a tvOS SDK in the future (#2340 #2778).

/cc @kkaefer @frederoni @friedbunny @boundsj @incanus @mikemorris


kkaefer:
@1ec5 I added an attempt at merging the projects, but I'm still not convinced it is a good idea since the workspace becomes pretty big; it takes a long time to have Xcode index everything and I'm unsure about the benefits (remember to add files in both projects IMHO is a weak reason).

kkaefer:
Also, once we merged iOS and macOS, should we also merge in the remaining separate Qt project?

1ec5:
We add files very frequently to the iOS and macOS projects. Because we check in the projects instead of generating them with CMake, the usual workflow is to use Xcode's Add Files feature.

Remembering to add files is indeed a weak argument, but we've had several incidents lately where a file was missing from either project. CI can't always catch these issues because headers can still be found via the header search path.

If the indexing time would become a burden for folks not working on the iOS or macOS SDK, I suppose we can table this work and see if the missing file reference issue subsides after the runtime styling work draws down.

> should we also merge in the remaining separate Qt project?

If there's a desire for that, then sure, I guess. But keeping Qt separate doesn't have the same downsides as keeping the iOS and macOS projects separate, does it?

1ec5:
> it takes a long time to have Xcode index everything

If I’m not mistaken, this is mostly because the same workspace contains two copies of mbgl.xcodeproj and thus two sets of file references. Instead, the workspace should contain only one copy of mbgl.xcodeproj, and that project should have two copies of the mbgl-platform-core target, one with iOS as the base SDK and one with macOS as the base SDK. With only one set of references, a particular scheme will have the same number of file references that it does on master, so indexing will take roughly the same amount of time.

1ec5:
> Instead, the workspace should contain only one copy of mbgl.xcodeproj, and that project should have two copies of the mbgl-core target, one with iOS as the base SDK and one with macOS as the base SDK.

This would only reduce the number of file references if we’re able to deduplicate file references; that is, if we’re able to have a single file be a member of multiple targets. Apparently CMake doesn’t support that. I wonder if this is something we could fix up after the fact using the [`xcodeproj` gem](https://rubygems.org/gems/xcodeproj) or [`xcode` npm module](https://www.npmjs.com/package/xcode).

/cc @incanus

1ec5:
> This would only reduce the number of file references if we’re able to deduplicate file references; that is, if we’re able to have a single file be a member of multiple targets. [Apparently CMake doesn’t support that](https://gitlab.kitware.com/cmake/cmake/issues/17805#note_384884). I wonder if this is something we could fix up after the fact using the [`xcodeproj` gem](https://rubygems.org/gems/xcodeproj) or [`xcode` npm module](https://www.npmjs.com/package/xcode).

[This gist](https://gist.github.com/ratazzi/f6d9217654d6605450a0) claims to duplicate a target using the `xcodeproj` gem. I’m not sure if it works as well as the Duplicate command in Xcode, but if so, that would be very useful for merging the iOS and macOS projects without burdening everyone with additional indexing. It would be even more useful for #9319.

1ec5:
This was originally an issue that got commandeered for a pull request. The pull request has stalled, but I’m hopeful the work in #9319 will get us closer to a merged iOS/macOS project.

PR #6047, opened 2016-08-17
Remove ios.xcscmblueprint

kkaefer:
https://github.com/mapbox/mapbox-gl-native/commit/96d3ccb0225ec784e6620b1a0f865bb9e466db41 added the [`ios.xcscmblueprint` file](https://github.com/mapbox/mapbox-gl-native/blob/fff0ebaa1754aabd9584c994624b7b948bec95af/platform/ios/ios.xcworkspace/xcshareddata/ios.xcscmblueprint), but it is platform specific and I'm seeing changes in this file that shouldn't be in the VCS.


.gitignore
```diff
@@ -25,3 +25,4 @@ xcuserdata
 /platform/ios/benchmark/assets/tiles/mapbox.mapbox-terrain-v2,mapbox.mapbox-streets-v6
 **/token
 /platform/macos/macos.xcworkspace/xcshareddata/macos.xcscmblueprint
+/platform/ios/ios.xcworkspace/xcshareddata/ios.xcscmblueprint
```
1ec5: We could also ignore any .xcscmblueprint.

1ec5:
Good catch, thanks!

Issue #5973, opened 2016-08-12
Add predicate property to MGLStyleLayer


1ec5:
MGLStyleLayer should have a `predicate` property that translates to a filter in mbgl. The developer would create a predicate like so:

```objc
[NSPredicate predicateWithFormat:@"name == 'USA'"]
```

The returned value would be an instance of NSComparisonPredicate with `leftExpression.keyPath` being `name`, `predicateOperatorType` being `NSEqualToPredicateOperatorType`, and `rightExpression.constantValue` being `USA`. We’d translate that into an `mbgl::style::EqualsFilter`.

[Here’s a full correspondence](https://gist.github.com/1ec5/3d81399e11ff45ee86d482cac25a0493) between NSPredicate types and style specification filter forms.

/cc @frederoni @incanus

boundsj:
This was completed by @frederoni in https://github.com/mapbox/mapbox-gl-native/pull/6049 🎉

PR #6049, opened 2016-08-17
MGLStyleLayer predicates (filters)

frederoni:
wip #5973.

/cc @incanus @1ec5

### DIFF
--- a/platform/darwin/src/MGLCircleStyleLayer.h
+++ b/platform/darwin/src/MGLCircleStyleLayer.h
@@ -18,6 +18,14 @@ typedef NS_ENUM(NSUInteger, MGLCircleStyleLayerCirclePitchScale) {
 
 @interface MGLCircleStyleLayer : MGLBaseStyleLayer <MGLStyleLayer>
 
+/**
+ A predicate that corresponds to the layer's <a href='https://www.mapbox.com/mapbox-gl-style-spec/#types-filter'>filter</a>.
+ 
+ The predicate's left expression must be a string that identifies a feature
+ property, or one of the special keys.
+ */
+@property (nonatomic, nullable) NSPredicate *predicate;
+
 #pragma mark - Accessing the Paint Attributes
 
 /**
--- a/platform/darwin/src/MGLFillStyleLayer.h
+++ b/platform/darwin/src/MGLFillStyleLayer.h
@@ -13,6 +13,14 @@ typedef NS_ENUM(NSUInteger, MGLFillStyleLayerFillTranslateAnchor) {
 
 @interface MGLFillStyleLayer : MGLBaseStyleLayer <MGLStyleLayer>
 
+/**
+ A predicate that corresponds to the layer's <a href='https://www.mapbox.com/mapbox-gl-style-spec/#types-filter'>filter</a>.
+ 
+ The predicate's left expression must be a string that identifies a feature
+ property, or one of the special keys.
+ */
+@property (nonatomic, nullable) NSPredicate *predicate;
+
 #pragma mark - Accessing the Paint Attributes
 
 /**
--- a/platform/darwin/src/MGLStyleLayer.h.ejs
+++ b/platform/darwin/src/MGLStyleLayer.h.ejs
@@ -35,6 +35,16 @@ typedef NS_ENUM(NSUInteger, MGL<%- camelize(type) %>StyleLayer<%- camelize(prope
 <% } -%>
 @interface MGL<%- camelize(type) %>StyleLayer : MGLBaseStyleLayer <MGLStyleLayer>
 
+<% if (type !== 'background' && type !== 'raster') { -%>
+/**
+ A predicate that corresponds to the layer's <a href='https://www.mapbox.com/mapbox-gl-style-spec/#types-filter'>filter</a>.
+ 
+ The predicate's left expression must be a string that identifies a feature
+ property, or one of the special keys.
+ */
+@property (nonatomic, nullable) NSPredicate *predicate;
+
+<% } -%>
 <% if (layoutProperties.length) { -%>
 #pragma mark - Accessing the Layout Attributes
 
@@ -53,6 +63,7 @@ typedef NS_ENUM(NSUInteger, MGL<%- camelize(type) %>StyleLayer<%- camelize(prope
 
 <% } -%>
 <% } -%>
+<% if (paintProperties.length) { -%>
 #pragma mark - Accessing the Paint Attributes
 
 <% for (const property of paintProperties) { -%>
@@ -68,6 +79,7 @@ typedef NS_ENUM(NSUInteger, MGL<%- camelize(type) %>StyleLayer<%- camelize(prope
  */
 @property (nonatomic<% if (!property.required) { %>, null_resettable<% } %>) <%- propertyType(property, false, type) %> <%- camelizeWithLeadingLowercase(property.name) %>;
 
+<% } -%>
 <% } -%>
 @end
 
--- a/platform/darwin/src/NSComparisonPredicate+MGLAdditions.h
+++ b/platform/darwin/src/NSComparisonPredicate+MGLAdditions.h
@@ -1,0 +1,7 @@
+#import <Foundation/Foundation.h>
+
+#include <mbgl/style/filter.hpp>
+
+@interface NSComparisonPredicate (MGLAdditions)
+
+@end
--- a/platform/darwin/src/NSComparisonPredicate+MGLAdditions.mm
+++ b/platform/darwin/src/NSComparisonPredicate+MGLAdditions.mm
@@ -1,0 +1,67 @@
+#import "NSComparisonPredicate+MGLAdditions.h"
+
+#import "NSPredicate+MGLAdditions.h"
+#import "NSExpression+MGLAdditions.h"
+
+@implementation NSComparisonPredicate (MGLAdditions)
+
+- (mbgl::style::Filter)mgl_filter
+{
+    switch (self.predicateOperatorType) {
+        case NSEqualToPredicateOperatorType: {
+            auto filter = mbgl::style::EqualsFilter();
+            filter.key = self.leftExpression.keyPath.UTF8String;
+            filter.value = self.rightExpression.mgl_filterValue;
+            return filter;
+        }
+        case NSNotEqualToPredicateOperatorType: {
+            auto filter = mbgl::style::NotEqualsFilter();
+            filter.key = self.leftExpression.keyPath.UTF8String;
+            filter.value = self.rightExpression.mgl_filterValue;
+            return filter;
+        }
+        case NSGreaterThanPredicateOperatorType: {
+            auto filter = mbgl::style::GreaterThanFilter();
+            filter.key = self.leftExpression.keyPath.UTF8String;
+            filter.value = self.rightExpression.mgl_filterValue;
+            return filter;
+        }
+        case NSGreaterThanOrEqualToPredicateOperatorType: {
+            auto filter = mbgl::style::GreaterThanEqualsFilter();
+            filter.key = self.leftExpression.keyPath.UTF8String;
+            filter.value = self.rightExpression.mgl_filterValue;
+            return filter;
+        }
+        case NSLessThanPredicateOperatorType: {
+            auto filter = mbgl::style::LessThanFilter();
+            filter.key = self.leftExpression.keyPath.UTF8String;
+            filter.value = self.rightExpression.mgl_filterValue;
+            return filter;
+        }
+        case NSLessThanOrEqualToPredicateOperatorType: {
+            auto filter = mbgl::style::LessThanEqualsFilter();
+            filter.key = self.leftExpression.keyPath.UTF8String;
+            filter.value = self.rightExpression.mgl_filterValue;
+            return filter;
+        }
+        case NSInPredicateOperatorType: {
+            auto filter = mbgl::style::InFilter();
+            filter.key = self.leftExpression.keyPath.UTF8String;
+            filter.values = self.rightExpression.mgl_filterValues;
+            return filter;
+        }
+        case NSMatchesPredicateOperatorType:
+        case NSLikePredicateOperatorType:
+        case NSBeginsWithPredicateOperatorType:
+        case NSEndsWithPredicateOperatorType:
+        case NSCustomSelectorPredicateOperatorType:
+        case NSContainsPredicateOperatorType:
+        case NSBetweenPredicateOperatorType:
+            [NSException raise:@"Unsupported operator type"
+                        format:@"NSPredicateOperatorType:%lu is not supported.", (unsigned long)self.predicateOperatorType];
+    }
+    
+    return {};
+}
+
+@end
--- a/platform/darwin/src/NSCompoundPredicate+MGLAdditions.mm
+++ b/platform/darwin/src/NSCompoundPredicate+MGLAdditions.mm
@@ -1,0 +1,56 @@
+#import "NSCompoundPredicate+MGLAdditions.h"
+
+#import "NSPredicate+MGLAdditions.h"
+#import "NSExpression+MGLAdditions.h"
+
+@implementation NSCompoundPredicate (MGLAdditions)
+
+- (std::vector<mbgl::style::Filter>)mgl_subfilters
+{
+    std::vector<mbgl::style::Filter>filters;
+    for (NSPredicate *predicate in self.subpredicates) {
+        filters.push_back(predicate.mgl_filter);
+    }
+    return filters;
+}
+
+- (mbgl::style::Filter)mgl_filter
+{
+    switch (self.compoundPredicateType) {
+        case NSNotPredicateType: {
+            
+            // Translate a nested NSComparisonPredicate with operator type NSInPredicateOperatorType into a flat mbgl::NotIn filter.
+            NSArray<NSComparisonPredicate *> *comparisonPredicates = [self.subpredicates filteredArrayUsingPredicate:[NSPredicate predicateWithFormat:@"class == %@", [NSComparisonPredicate class]]];
+            NSArray<NSComparisonPredicate *> *notInPredicates = [comparisonPredicates filteredArrayUsingPredicate:[NSPredicate predicateWithBlock:^BOOL(NSComparisonPredicate *_Nonnull predicate, NSDictionary<NSString *,id> * _Nullable bindings) {
+                return predicate.predicateOperatorType == NSInPredicateOperatorType;
+            }]];
+            
+            if (notInPredicates.count) {
+                auto filter = mbgl::style::NotInFilter();
+                filter.key = notInPredicates.firstObject.leftExpression.keyPath.UTF8String;
+                filter.values = notInPredicates.firstObject.rightExpression.mgl_filterValues;
+                return filter;
+            } else {
+                auto filter = mbgl::style::NoneFilter();
+                filter.filters = [self mgl_subfilters];
+                return filter;
+            }
+        }
+        case NSAndPredicateType: {
+            auto filter = mbgl::style::AllFilter();
+            filter.filters = [self mgl_subfilters];
+            return filter;
+        }
+        case NSOrPredicateType: {
+            auto filter = mbgl::style::AnyFilter();
+            filter.filters = [self mgl_subfilters];
+            return filter;
+        }
+    }
+    
+    [NSException raise:@"Compound predicate type not handled"
+                format:@""];
+    return {};
+}
+
+@end
--- a/platform/darwin/src/NSExpression+MGLAdditions.mm
+++ b/platform/darwin/src/NSExpression+MGLAdditions.mm
@@ -1,0 +1,43 @@
+#import "NSExpression+MGLAdditions.h"
+
+@implementation NSExpression (MGLAdditions)
+
+- (std::vector<mbgl::Value>)mgl_filterValues
+{
+    if ([self.constantValue isKindOfClass:NSArray.class]) {
+        NSArray *values = self.constantValue;
+        std::vector<mbgl::Value>convertedValues;
+        for (id value in values) {
+            convertedValues.push_back([self mgl_convertedValueWithValue:value]);
+        }
+        return convertedValues;
+    }
+    [NSException raise:@"Values not handled" format:@""];
+    return { };
+}
+
+- (mbgl::Value)mgl_filterValue
+{
+    return [self mgl_convertedValueWithValue:self.constantValue];
+}
+
+- (mbgl::Value)mgl_convertedValueWithValue:(id)value
+{
+    if ([value isKindOfClass:NSString.class]) {
+        return { std::string([(NSString *)value UTF8String]) };
+    } else if ([value isKindOfClass:NSNumber.class]) {
+        NSNumber *number = (NSNumber *)value;
+        if((strcmp([number objCType], @encode(int))) == 0) {
+            return { number.intValue };
+        } else if ((strcmp([number objCType], @encode(double))) == 0) {
+            return { number.doubleValue };
+        } else if ((strcmp([number objCType], @encode(bool))) == 0) {
+            return { number.boolValue };
+        }
+    }
+    [NSException raise:@"Value not handled"
+                format:@"Can’t convert %s:%@ to mbgl::Value", [value objCType], value];
+    return { };
+}
+
+@end
--- a/platform/darwin/src/NSExpression+MGLAdditions.mm
+++ b/platform/darwin/src/NSExpression+MGLAdditions.mm
@@ -31,7 +31,7 @@
             return { number.intValue };
         } else if ((strcmp([number objCType], @encode(double))) == 0) {
             return { number.doubleValue };
-        } else if ((strcmp([number objCType], @encode(bool))) == 0) {
+        } else if ((strcmp([number objCType], @encode(char))) == 0) {
             return { number.boolValue };
         }
     }
--- a/platform/darwin/src/NSPredicate+MGLAdditions.h
+++ b/platform/darwin/src/NSPredicate+MGLAdditions.h
@@ -1,0 +1,13 @@
+#import <Foundation/Foundation.h>
+
+#import "NSExpression+MGLAdditions.h"
+#include <mbgl/style/filter.hpp>
+
+@interface NSPredicate (MGLAdditions)
+
+- (mbgl::style::Filter)mgl_filter;
+
++ (instancetype)mgl_predicateWithFilter:(mbgl::style::Filter)filter;
+
+@end
+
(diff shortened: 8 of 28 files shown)


platform/darwin/src/MGLBaseStyleLayer.h
```diff
@@ -14,4 +14,11 @@
  */
 @property (nonatomic, assign) float minimumZoomLevel;
 
+@property (nonatomic) NSPredicate *predicate;
+
+/**
+ Updates the layer’s layout and paint properties.
+ */
+- (void)update;
```
1ec5: We made this method private in #6018. Any reason you’re reintroducing this method in the public interface?

1ec5:
To convert from NSPredicate to the various `mbgl::style::Filter` classes, we should use a visitor pattern like [the ones in MGLFeature.mm](https://github.com/mapbox/mapbox-gl-native/blob/bca9d091805dc01a4456ab3f24e9de87f9b4aa48/platform/darwin/src/MGLFeature.mm#L119). Whereas `PropertyValueEvaluator` converts a recursive (arbitrarily nested) C++ structure into an Objective-C structure, we want to convert a recursive Objective-C structure into a C++ structure. We can nonetheless rely on operator overloading in C++ to keep things clean without a series of if/else-if statements.

platform/darwin/src/MGLBaseStyleLayer.h
frederoni: No reason. It was in my git stash from last week. I'll remove it.

platform/darwin/src/NSPredicate+MGLAdditions.mm
```diff
@@ -0,0 +1,63 @@
+#import "NSPredicate+MGLAdditions.h"
+
+class FilterEvaluator {
+public:
+    
+    mbgl::Value getValue(id obj) {
```
1ec5: Instead of a catch-all `getValue()` method, use the Visitor pattern that we’re already inside. Define operators for NSString and NSNumber in a FilterExpressionEvaluator, then apply the FilterExpressionEvaluator to each `constantValue` you encounter.

platform/darwin/src/NSPredicate+MGLAdditions.mm
```diff
@@ -0,0 +1,63 @@
...
+        if ([obj isKindOfClass:NSString.class]) {
+            NSString *string = (NSString *)obj;
+            return std::string(string.UTF8String);
+        } else if ([obj isKindOfClass:NSString.class]) {
```
1ec5: I think you meant NSNumber here.

platform/darwin/src/NSPredicate+MGLAdditions.mm
```diff
@@ -0,0 +1,63 @@
...
+                filter.value = getValue(predicate.rightExpression.constantValue);
+                return filter;
+            }
+            case NSNotEqualToPredicateOperatorType: {
+                auto filter = mbgl::style::NotEqualsFilter();
+                filter.key = predicate.leftExpression.keyPath.UTF8String;
+                filter.value = getValue(predicate.rightExpression.constantValue);
+                return filter;
+            }
+            default:
```
1ec5: Instead of a `default` case that raises, add cases for all of the `NSPredicateOperatorType` values. If Foundation ever adds a new `NSPredicateOperatorType` value, we’ll get a compiler warning, but not if there’s a `default` case.

platform/darwin/src/NSPredicate+MGLAdditions.mm
```diff
@@ -0,0 +1,63 @@
...
+        }
+        return mbgl::style::NoneFilter();
+    }
+};
+
+@implementation NSPredicate (MGLAdditions)
+
+- (mbgl::style::Filter)mgl_filter
+{
+    if ([self isKindOfClass:NSCompoundPredicate.class]) {
```
1ec5: Instead of this check, implement an operator in FilterEvaluator for NSCompoundPredicate and raise the “not yet supported” exception there.

platform/darwin/src/MGLCircleStyleLayer.h
```diff
@@ -17,6 +17,8 @@ typedef NS_ENUM(NSUInteger, MGLCircleStyleLayerCirclePitchScale) {
 
 @interface MGLCircleStyleLayer : MGLBaseStyleLayer <MGLStyleLayer>
 
+@property (nonatomic) NSPredicate *predicate;
```
1ec5: #6076 has landed, so without a nullability qualifier, the compiler will assume that the predicate is non-`nil`.

1ec5:
Also make sure to test some edge cases where you may need to escape the format string or use format string specifiers:

* [ ] Keys with hyphens are converted correctly
* [ ] `$type` and `$id` keys (#5708) are converted correctly

platform/darwin/src/NSPredicate+MGLAdditions.mm
```diff
@@ -0,0 +1,63 @@
...
+        [NSException raise:@"Nested predicates (NSCompoundPredicate) are not yet supported" format:@""];
+    }
+    
+    NSComparisonPredicate *predicate = (NSComparisonPredicate *)self;
+    FilterEvaluator evaluator;
+    auto filter = evaluator(predicate);
```
1ec5: A C++ evaluator may not be a great tool for converting from NSPredicate to `mbgl::style::Filter` after all. We generally use this pattern with `mapbox::util::variant`, not an Objective-C umbrella class. Instead, let’s make this method do nothing but assert, then implement `-mgl_filter` overrides in categories on NSComparisonPredicate and NSCompoundPredicate.

The logic for converting a key or value to an `mbgl::Value` should live in a category on NSExpression.

A C++ evaluator remains a good idea for efficiently and compactly converting an `mbgl::style::Filter` into an NSPredicate. We could have the evaluator call into category initializers on NSComparisonPredicate, NSCompoundPredicate, and NSExpression, if that would make the code easier to understand.

platform/darwin/src/MGLCircleStyleLayer.h
frederoni: will fix that after rebase

platform/darwin/src/NSPredicate+MGLAdditions.h
```diff
@@ -1,9 +1,108 @@
...
 @end
+
+namespace mbgl {
+namespace darwin {
+
+class FilterEvaluator {
+public:
+
+mbgl::style::Filter operator()(NSComparisonPredicate *predicate) {
+    switch (predicate.predicateOperatorType) {
```
1ec5: The entire FilterEvaluator definition shouldn't be inlined in the header. Declare the operators here but define them in the implementation file.

platform/darwin/src/NSComparisonPredicate+MGLAdditions.h
```diff
@@ -0,0 +1,9 @@
+#import <Foundation/Foundation.h>
+
+#include <mbgl/style/filter.hpp>
+
+@interface NSComparisonPredicate (MGLAdditions)
+
+- (mbgl::style::Filter)mgl_filter;
```
1ec5: Since this is already declared on NSPredicate(MGLAdditions), there's no need to declare it here and in NSCompoundPredicate(MGLAdditions).

platform/darwin/src/NSPredicate+MGLAdditions.h
```diff
@@ -0,0 +1,108 @@
...
+@interface NSPredicate (MGLAdditions)
+
+- (mbgl::style::Filter)mgl_filter;
+
+@end
+
+namespace mbgl {
+namespace darwin {
+
+class FilterEvaluator {
```
1ec5: This class is no longer needed. Fold the NSComparisonPredicate operator into `-[NSComparisonPredicate(MGLAdditions) mgl_filter]` and the NSComound Predicate operator into `-[NSCompoundPredicate(MGLAdditions) mgl_filter]`.

platform/darwin/src/NSPredicate+MGLAdditions.mm
```diff
@@ -0,0 +1,11 @@
+#import "NSPredicate+MGLAdditions.h"
+
+@implementation NSPredicate (MGLAdditions)
+
+- (mbgl::style::Filter)mgl_filter
+{
+    [NSException raise:@"A specific predicate must override this method." format:@""];
```
1ec5: The raise parameter should be a short title. The format parameter should be a complete sentence describing the problem.

platform/darwin/src/NSExpression+MGLAdditions.mm
```diff
@@ -0,0 +1,43 @@
...
+    [NSException raise:@"Values not handled" format:@""];
+    return std::vector<mbgl::Value>();
+}
+
+- (mbgl::Value)mgl_filterValue
+{
+    return [self convertedValueWithValue:self.constantValue];
+}
+
+- (mbgl::Value)convertedValueWithValue:(id)value
```
1ec5: Always prefix with mgl_ when adding a category method on a class we don't own.

platform/darwin/src/NSPredicate+MGLAdditions.h
```diff
@@ -0,0 +1,108 @@
...
+
+mbgl::style::Filter operator()(NSComparisonPredicate *predicate) {
+    switch (predicate.predicateOperatorType) {
+        case NSMatchesPredicateOperatorType:
+        case NSLikePredicateOperatorType:
+        case NSBeginsWithPredicateOperatorType:
+        case NSEndsWithPredicateOperatorType:
+        case NSCustomSelectorPredicateOperatorType:
+        case NSContainsPredicateOperatorType:
+        case NSBetweenPredicateOperatorType:
```
1ec5: These are all operators we don't support. They should not be conflated with the equals operator. Instead of falling through, raise an exception explaining that the operator is unsupported.

1ec5: As it happens, we can implement between as AllFilter(GreaterThanEqualsFilter, LessThanEqualsFilter).

platform/darwin/src/NSPredicate+MGLAdditions.h
```diff
@@ -0,0 +1,108 @@
...
+public:
+
+mbgl::style::Filter operator()(NSComparisonPredicate *predicate) {
+    switch (predicate.predicateOperatorType) {
+        case NSMatchesPredicateOperatorType:
+        case NSLikePredicateOperatorType:
+        case NSBeginsWithPredicateOperatorType:
+        case NSEndsWithPredicateOperatorType:
+        case NSCustomSelectorPredicateOperatorType:
+        case NSContainsPredicateOperatorType:
```
1ec5: This is the same thing as NSInPredicateOperatorType, except with the left and right expressions swapped.

platform/darwin/test/MGLFilterTests.m
```diff
@@ -0,0 +1,69 @@
...
+
+- (void)tearDown
+{
+    [self.mapView.style removeLayer:layer];
+    [self.mapView.style removeSource:source];
+}
+
+- (NSArray<NSPredicate *> *)predicates
+{
+    NSPredicate *equalPredicate = [NSPredicate predicateWithFormat:@"type == 'neighborhood'"];
```
1ec5: Mapbox Streets v7 uses the British spelling, `neighbourhood`, for consistency with OpenStreetMap.

platform/macos/macos.xcodeproj/project.pbxproj
```diff
@@ -613,6 +626,7 @@
 			isa = PBXGroup;
 			children = (
 				DA8F257C1D51C5F40010E6B5 /* Layers */,
+				35C5D84B1D6DD75B00E95907 /* MGLFilterTests.m */,
 			);
 			name = Styling;
 			path = ../../ios/test;
```
1ec5: This group should correspond to platform/macos/test/, not platform/ios/test/.

platform/darwin/src/MGLStyleLayer.h.ejs
```diff
@@ -35,6 +35,10 @@ typedef NS_ENUM(NSUInteger, MGL<%- camelize(type) %>StyleLayer<%- camelize(prope
 <% } -%>
 @interface MGL<%- camelize(type) %>StyleLayer : MGLBaseStyleLayer <MGLStyleLayer>
 
+<% if (type !== 'background' && type !== 'raster') { -%>
+@property (nonatomic, nullable) NSPredicate *predicate;
```
1ec5: Needs documentation.

platform/darwin/src/NSComparisonPredicate+MGLAdditions.mm
```diff
@@ -0,0 +1,76 @@
+#import "NSComparisonPredicate+MGLAdditions.h"
+
+#import "NSPredicate+MGLAdditions.h"
+#import "NSExpression+MGLAdditions.h"
+
+class FilterEvaluator {
+public:
+
+mbgl::style::Filter operator()(NSComparisonPredicate *predicate) {
```
1ec5: This function should be inlined into `-[NSComparisonPredicate(MGLAdditions) mgl_filter]`.

platform/darwin/src/NSComparisonPredicate+MGLAdditions.mm
```diff
@@ -0,0 +1,76 @@
...
+        }
+        case NSMatchesPredicateOperatorType:
+        case NSLikePredicateOperatorType:
+        case NSBeginsWithPredicateOperatorType:
+        case NSEndsWithPredicateOperatorType:
+        case NSCustomSelectorPredicateOperatorType:
+        case NSContainsPredicateOperatorType:
+        case NSBetweenPredicateOperatorType:
+            [NSException raise:@"Operator type not handled"
+                        format:@""];
```
1ec5: For debugging purposes, the format string should say which unsupported operator type is being used here.

platform/darwin/src/NSCompoundPredicate+MGLAdditions.mm
```diff
@@ -0,0 +1,51 @@
...
+    
+    std::vector<mbgl::style::Filter>getFilters (NSArray<NSPredicate *> *predicates) {
+        std::vector<mbgl::style::Filter>filters;
+        for (NSPredicate *predicate in predicates) {
+            filters.push_back(predicate.mgl_filter);
+        }
+        return filters;
+    }
+    
+    mbgl::style::Filter operator()(NSCompoundPredicate *predicate) {
```
1ec5: This function should be inlined into `-[NSCompoundPredicate(MGLAdditions) mgl_filter]`.

platform/darwin/src/NSCompoundPredicate+MGLAdditions.mm
```diff
@@ -0,0 +1,51 @@
+#import "NSCompoundPredicate+MGLAdditions.h"
+
+#import "NSPredicate+MGLAdditions.h"
+#import "NSExpression+MGLAdditions.h"
+
+class PredicateEvaluator {
+public:
+    
+    std::vector<mbgl::style::Filter>getFilters (NSArray<NSPredicate *> *predicates) {
```
1ec5: This function should become `-[NSCompoundPredicate(MGLAdditions) mgl_subfilters]`.

platform/darwin/src/NSExpression+MGLAdditions.mm
```diff
@@ -0,0 +1,60 @@
+#import "NSExpression+MGLAdditions.h"
+
+class FilterExpressionEvaluator {
+public:
+    id operator()(mbgl::Value value) {
```
1ec5: Instead of a single `mbgl::Value` operator with a number of `if…is` clauses, there should be a separate operator for each type you want to convert. So, for example:

```objc
id operator()(bool value) {
    return @(value.get<bool>());
}
```

1ec5: The implementation of FilterExpressionEvaluator would look identical to PropertyValueEvaluator in MGLFeature.mm. We should make that class accessible to the whole project, then reusing it instead of reinventing the wheel here.

platform/darwin/src/NSCompoundPredicate+MGLAdditions.mm
```diff
@@ -0,0 +1,51 @@
...
+    mbgl::style::Filter operator()(NSCompoundPredicate *predicate) {
+        switch (predicate.compoundPredicateType) {
+            case NSNotPredicateType: {
+                auto filter = mbgl::style::NoneFilter();
+                filter.filters = getFilters(predicate.subpredicates);
+                return filter;
+            }
+            case NSAndPredicateType: {
+                auto filter = mbgl::style::AllFilter();
+                filter.filters = getFilters(predicate.subpredicates);
```
1ec5: The NSCompoundPredicate documentation mentions this edge case:

> An AND predicate with no subpredicates evaluates to TRUE.

Does `mbgl::style::AllFilter` have the same behavior?

platform/darwin/src/NSCompoundPredicate+MGLAdditions.mm
```diff
@@ -0,0 +1,51 @@
...
+                return filter;
+            }
+            case NSOrPredicateType: {
+                auto filter = mbgl::style::AnyFilter();
+                filter.filters = getFilters(predicate.subpredicates);
```
1ec5: The NSCompountPredicate documentation mentions this edge case:

> An OR predicate with no subpredicates evaluates to FALSE.

Does `mbgl::style::AnyFilter` have the same behavior?

platform/darwin/src/NSCompoundPredicate+MGLAdditions.mm
```diff
@@ -0,0 +1,51 @@
...
+        for (NSPredicate *predicate in predicates) {
+            filters.push_back(predicate.mgl_filter);
+        }
+        return filters;
+    }
+    
+    mbgl::style::Filter operator()(NSCompoundPredicate *predicate) {
+        switch (predicate.compoundPredicateType) {
+            case NSNotPredicateType: {
+                auto filter = mbgl::style::NoneFilter();
```
1ec5: This should be `NotFilter`, not `NoneFilter`. Also, for performance reasons, we should have a special case in which an `NSNotPredicateType` predicate containing an `NSInPredicateOperatorType` should use `mbgl::style::NotInFilter`.

platform/darwin/test/MGLFilterTests.m
```diff
@@ -0,0 +1,106 @@
...
+{
+    [self.mapView.style removeLayer:layer];
+    [self.mapView.style removeSource:source];
+}
+
+- (NSArray<NSPredicate *> *)predicates
+{
+    NSPredicate *equalPredicate = [NSPredicate predicateWithFormat:@"type == 'neighbourhood'"];
+    NSPredicate *notEqualPredicate = [NSPredicate predicateWithFormat:@"type != 'park'"];
+    NSPredicate *greaterThanPredicate = [NSPredicate predicateWithFormat:@"%K > %@", @"stroke-width", @2.1];
```
1ec5: Nit: it should also be possible to inline the 2.1 into the format string.

platform/darwin/test/MGLFilterTests.m
```diff
@@ -0,0 +1,106 @@
...
+             typePredicate,
+             idPredicate,
+             specialCharsPredicate];
+}
+
+- (void)testAllPredicates
+{
+    for (NSPredicate *predicate in self.predicates) {
+        layer.predicate = predicate;
+        XCTAssertEqualObjects(layer.predicate, predicate);
```
1ec5: This ensures we can round-trip the predicates, which is a good start. However, the tests don’t verify that for instance `specialCharsPredicate` resolves to the right `mbgl::style::Filter`. What if the special characters get mangled during the conversion but get unmangled during the conversion back to NSPredicate?

platform/darwin/src/MGLCircleStyleLayer.h
```diff
@@ -18,6 +18,16 @@ typedef NS_ENUM(NSUInteger, MGLCircleStyleLayerCirclePitchScale) {
 
 @interface MGLCircleStyleLayer : MGLBaseStyleLayer <MGLStyleLayer>
 
+/**
+ A predicate that corresponds to the layer's filter.
+ 
+ The predicate's left expression must be a string that identifies a feature
+ property, or one of the following special keys:
+    "$type" - identifies the feature by type.
```
1ec5: I don’t think this will look like a list in jazzy. We should either insert our own Doxygen or HTML formatting or override this documentation.

platform/darwin/src/NSExpression+MGLAdditions.mm
```diff
@@ -0,0 +1,43 @@
...
+{
+    if ([value isKindOfClass:NSString.class]) {
+        return { std::string([(NSString *)value UTF8String]) };
+    } else if ([value isKindOfClass:NSNumber.class]) {
+        NSNumber *number = (NSNumber *)value;
+        if((strcmp([number objCType], @encode(int))) == 0) {
+            return { number.intValue };
+        } else if ((strcmp([number objCType], @encode(double))) == 0) {
+            return { number.doubleValue };
+        } else if ((strcmp([number objCType], @encode(bool))) == 0) {
```
boundsj: @frederoni  I expect to be able to write something like `[NSPredicate predicateWithFormat:@"%K != %@", @"cluster", [NSNumber numberWithBool:YES]]]` however this is not caught by this implementation (ref: http://stackoverflow.com/questions/23425187/nsnumber-with-bool-is-no-bool-on-64bit). Using `else if ((strcmp([number objCType], @encode(char))) == 0)` worked for me.

frederoni: Did it raise an exception or returned `mbgl::Value` with incorrect type? Feel free to push the fix.

boundsj: It triggered the exception since it was not caught in the previous conditions. Fixed here https://github.com/mapbox/mapbox-gl-native/pull/6049/commits/eefe5cb050789078b7fae0f05f13fbc929731240

boundsj:
👍

platform/darwin/src/MGLFillStyleLayer.h
```diff
@@ -13,6 +13,14 @@ typedef NS_ENUM(NSUInteger, MGLFillStyleLayerFillTranslateAnchor) {
 
 @interface MGLFillStyleLayer : MGLBaseStyleLayer <MGLStyleLayer>
 
+/**
+ A predicate that corresponds to the layer's <a href='https://www.mapbox.com/mapbox-gl-style-spec/#types-filter'>filter</a>.
+ 
+ The predicate's left expression must be a string that identifies a feature
+ property, or one of the special keys.
```
1ec5: Possibly tail work, but we need to document inline which NSExpression operator types we support, as well as the valid values of `$type` and `$id`. Linking to the style specification will suffice for now, but it’s suboptimal because the JavaScript syntax and terminology is very foreign to the NSPredicate-based interface we ended up exposing.

platform/darwin/src/NSPredicate+MGLAdditions.mm
```diff
@@ -0,0 +1,107 @@
...
+        if (predicates.count > 1) {
+            NSCompoundPredicate *predicate = [NSCompoundPredicate andPredicateWithSubpredicates:predicates];
+            return [NSCompoundPredicate notPredicateWithSubpredicate:predicate];
+        } else {
+            return [NSCompoundPredicate notPredicateWithSubpredicate:predicates.firstObject];
+        }
+    }
+    
+    NSPredicate* operator()(mbgl::style::HasFilter filter) {
+        [NSException raise:@"Unsupported filter type"
```
1ec5: This should be translated as `key != nil`.

platform/darwin/src/NSPredicate+MGLAdditions.h
1ec5: → #6244

platform/darwin/src/NSPredicate+MGLAdditions.h
1ec5: → #6244

platform/darwin/src/NSCompoundPredicate+MGLAdditions.mm
1ec5: [Yes, it does](https://github.com/mapbox/mapbox-gl-native/blob/6a39cf5aaece81c7a531b12321dd503004cc45b8/include/mbgl/style/filter_evaluator.hpp#L102).

platform/darwin/src/NSCompoundPredicate+MGLAdditions.mm
1ec5: [Yes, it does](https://github.com/mapbox/mapbox-gl-native/blob/6a39cf5aaece81c7a531b12321dd503004cc45b8/include/mbgl/style/filter_evaluator.hpp#L93).

platform/darwin/src/MGLFillStyleLayer.h
1ec5: Oh, I’m seeing now that this documentation was originally inline but was removed in response to https://github.com/mapbox/mapbox-gl-native/pull/6049#discussion_r76650575. I meant that we should describe the `$type` and `$id` keys but do so in an HTML definition list.

platform/darwin/test/MGLFilterTests.mm
```diff
@@ -0,0 +1,139 @@
...
+    NSPredicate *greaterThanOrEqualToPredicate = [NSPredicate predicateWithFormat:@"%K >= %@", @"stroke-width", @2.1];
+    NSPredicate *lessThanOrEqualToPredicate = [NSPredicate predicateWithFormat:@"%K <= %@", @"stroke-width", @2.1];
+    NSPredicate *lessThanPredicate = [NSPredicate predicateWithFormat:@"%K < %@", @"stroke-width", @2.1];
+    NSPredicate *inPredicate = [NSPredicate predicateWithFormat:@"type IN %@", @[@"park", @"neighbourhood"]];
+    NSPredicate *notInPredicate = [NSPredicate predicateWithFormat:@"NOT (type IN %@)", @[@"park", @"neighbourhood"]];
+    NSPredicate *inNotInPredicate = [NSPredicate predicateWithFormat:@"type IN %@ AND NOT (type IN %@)", @[@"park"], @[@"neighbourhood", @"test"]];
+    NSPredicate *typePredicate = [NSPredicate predicateWithFormat:@"%K == %@", @"$type", @"Feature"];
+    NSPredicate *idPredicate = [NSPredicate predicateWithFormat:@"%K == %@", @"$id", @"1234123"];
+    NSPredicate *specialCharsPredicate = [NSPredicate predicateWithFormat:@"%K == %@", @"ty-’pè", @"sŒm-ethįng"];
+    NSPredicate *booleanPredicate = [NSPredicate predicateWithFormat:@"%K != %@", @"cluster", [NSNumber numberWithBool:YES]];
```
1ec5: This can be inlined as `%K != YES`.

PR #6052, opened 2016-08-17
Gradle And Android Plugin Updates For Security Fix

### LINKED ISSUE #6050: Update Gradle And Android Plugin For Security Fix
Google [released Android Studio 2.1.3 earlier this week](https://developer.android.com/studio/releases/index.html#Revisions) and is strongly recommending people update their projects versions of Gradle and the Android plugin as there's a ["Local privilege escalation when using the Daemon"](https://docs.gradle.org/2.14/release-notes#local-privilege-escalation-when-using-the-daemon) security bug.  While we tend not to use the daemon internally at Mapbox and thus likely not affected, this is something we should do regardless and the sooner the better to keep things running smoothly.

screen shot 2016-08-17 at 9 10 21 am

/cc @mapbox/android

bleege:
Closes #6050


bleege:
Rebased and Merged!

PR #6058, opened 2016-08-17
Make Xcode requirement clear on install requirements cc @friedbunny f…

tmcw:
…or review

via https://github.com/mapbox/mapbox-gl-native/issues/5472


friedbunny:
Thanks again, @tmcw. I’m going to close here and roll this into #6059.

PR #6059, opened 2016-08-17
[ios, macos] Make Xcode 7.3 the minimum required version

### LINKED ISSUE #5472: Require Xcode 7.3 due to NS_SWIFT_NAME

friedbunny:
This commit makes Xcode 7.3 the minimum version requirement to build and integrate the iOS SDK.

This removes `#import “MGLTypes.h”` from headers where it was only being used to conditionally define `NS_ASSUME_NONNULL_*` and various nullability traits, which were necessary to support Xcode 6.

Xcode 7.3 becomes the minimum supported version because of `NS_SWIFT_NAME`, which is not fully supported by lower versions (and we were already using it).

@frederoni I also removed the MGLTypes.h import from the runtime styling generation template and regenerated.

Fixes #5472.

/cc @1ec5 @tmcw

### DIFF
--- a/platform/ios/CHANGELOG.md
+++ b/platform/ios/CHANGELOG.md
@@ -6,6 +6,7 @@ Mapbox welcomes participation and contributions from everyone. Please read [CONT
 
 * A new runtime styling API allows you to adjust the style and content of the base map dynamically. All the options available in [Mapbox Studio](https://www.mapbox.com/studio/) are now exposed via MGLStyle and subclasses of MGLStyleLayer and MGLSource. ([#5727](https://github.com/mapbox/mapbox-gl-native/pull/5727))
 * The user location annotation is now customizable via the newly added `MGLUserLocationAnnotationView` class. ([#5882](https://github.com/mapbox/mapbox-gl-native/pull/5882))
+* Xcode 7.3 or higher is now required to use this SDK. ([#6059](https://github.com/mapbox/mapbox-gl-native/issues/6059))
 * Simulator architecture slices are included in the included dSYM file, allowing you to symbolicate crashes that occur in the Simulator. ([#5740](https://github.com/mapbox/mapbox-gl-native/pull/5740))
 * As the user zooms in, tiles from lower zoom levels are scaled up until tiles for higher zoom levels are loaded. ([#5143](https://github.com/mapbox/mapbox-gl-native/pull/5143))
 * Fixed an issue causing the wrong annotation view to be selected when tapping an annotation view with a center offset applied. ([#5931](https://github.com/mapbox/mapbox-gl-native/pull/5931))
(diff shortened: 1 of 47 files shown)


friedbunny:
This PR doesn’t handle macOS yet.

platform/ios/CHANGELOG.md
```diff
@@ -6,6 +6,7 @@ Mapbox welcomes participation and contributions from everyone. Please read [CONT
 
 * A new runtime styling API allows you to adjust the style and content of the base map dynamically. All the options available in [Mapbox Studio](https://www.mapbox.com/studio/) are now exposed via MGLStyle and subclasses of MGLStyleLayer and MGLSource. ([#5727](https://github.com/mapbox/mapbox-gl-native/pull/5727))
 * The user location annotation is now customizable via the newly added `MGLUserLocationAnnotationView` class. ([#5882](https://github.com/mapbox/mapbox-gl-native/pull/5882))
+* Xcode 7.3 or higher is now officially required to build and integrate this SDK. ([#6059](https://github.com/mapbox/mapbox-gl-native/issues/6059))
```
1ec5: Nit: remove “officially”. Also, “install” or “link against” would be more relevant to most readers than “build and integrate”.

friedbunny: How about “use this SDK”?

1ec5:
* [ ] Update the minimum required Xcode version in the [build requirements](https://github.com/mapbox/mapbox-gl-native/blob/bca9d091805dc01a4456ab3f24e9de87f9b4aa48/platform/ios/DEVELOPING.md#requirements) (also for the corresponding macOS document)
* [ ] While you’re at it, mention that headers should be audited for nullability in the [new source file guide](https://github.com/mapbox/mapbox-gl-native/blob/bca9d091805dc01a4456ab3f24e9de87f9b4aa48/platform/ios/DEVELOPING.md#adding-a-source-code-file) (also for the corresponding macOS document)

platform/ios/CHANGELOG.md
1ec5: Sure.

friedbunny:
> Update the minimum required Xcode version in the build requirements 

This was already done for iOS; now it’s also done for macOS.

1ec5:
👍

PR #6060, opened 2016-08-17
fix #5983: ignore viewport no-ops

### LINKED ISSUE #5983: Avoid animating to the same camera options
If you attempt to change the camera to an `mbgl::CameraOptions` that’s identical to the current `CameraOptions`, `mbgl::Transform` still goes through the entire animation machinery. We do [avoid modifying the `TransformState`](https://github.com/mapbox/mapbox-gl-native/blob/532c659b22c73eac04c7d643aff6ac5f1cc9f55a/src/mbgl/map/transform.cpp#L151) in that case, but by then it’s too late: we’ve already kicked off an animation timer and all the calculations for each step of the animation. The animation timer updates the map on a schedule regardless of whether the transform state changes.

There are multiple places where we can short-circuit the animation, and we should consider all of them to avoid severe energy consumption in navigation use cases, which require constant animation.

Incidentally, when we bail for a no-op animation, should the completion callback by called immediately, or still on a delay?

/cc @kkaefer @friedbunny

incanus:
Work in progress towards fixing #5983. We may want to do some of this not just at the mobile bindings layer, but also in core for completeness. 

/cc @1ec5 @friedbunny


platform/darwin/src/MGLMapCamera.mm
```diff
@@ -98,28 +98,26 @@ - (NSString *)description
...
-    {
-        return YES;
-    }
-    
-    MGLMapCamera *otherCamera = other;
-    return (_centerCoordinate.latitude == otherCamera.centerCoordinate.latitude
-            && _centerCoordinate.longitude == otherCamera.centerCoordinate.longitude
-            && _altitude == otherCamera.altitude
-            && _pitch == otherCamera.pitch && _heading == otherCamera.heading);
+    NSAssert([otherCamera isKindOfClass:[self class]], @"must compare two cameras");
```
1ec5: The method takes an id, which implies you can pass in any object. Do other classes assert when you attempt to compare to something of a different class?

incanus: Good point. 5fba4e6d961e7900db82071c49ac19c0bc6a8266

platform/darwin/src/MGLGeometry_Private.h
```diff
@@ -34,6 +34,14 @@ NS_INLINE mbgl::LatLngBounds MGLLatLngBoundsFromCoordinateBounds(MGLCoordinateBo
                                     MGLLatLngFromLocationCoordinate2D(coordinateBounds.ne));
 }
 
+NS_INLINE void MGLLocationCoordinatesFromCoordinateBounds(MGLCoordinateBounds coordinateBounds, CLLocationCoordinate2D *coordinates) {
+    assert(sizeof(coordinates) == 4 * sizeof(CLLocationCoordinate2D));
```
1ec5: Nit: use NSAssert.

incanus: Tried, but can't do that in `NS_INLINE`.

1ec5: My bad, you want NSCAssert, because we're in a C function that has no access to self.

1ec5:
Closing in favor of #7125, #7119, and #7724.

Issue #6055, opened 2016-08-17
Map::queryPointAnnotations() -> Map::queryRenderedFeatures() returns unexpected result as map is moved


boundsj:
In https://github.com/mapbox/mapbox-gl-native/pull/5165 we added a new method `queryPointAnnotations` that calls `queryRenderedFeatures` with a `ScreenBox` representing the view port for which you want to query for all point annotations. In the iOS SDK, this is used by:

- -[MGLMapView annotationTagAtPoint:persistingResults:] // used for annotation selection
- -[MGLMapView accessibilityAnnotationCount]
- -[MGLMapView accessibilityElementAtIndex:]
- -[MGLMapView indexOfAccessibilityElement:]

I was planing to use `queryPointAnnotations` in https://github.com/mapbox/mapbox-gl-native/pull/5987 to help clean up and optimize the implementation in the iOS SDK's `updateAnnotationViews`. However, based on the visual behavior I'm seeing after doing that, I don't know that it'll be possible to rely on the underlying `queryRenderedFeatures` for this purpose.

For example, if I give `queryPointAnnotations\queryRenderedFeatures` a view port of:

`viewPort == mapView.bounds //(origin = (x = 0, y = 0), size = (width = 375, height = 667))`

I see that the annotation view update method stops updating views ~almost~ when I would expect it to, at first (the views get stuck near the edge of the view port when their annotations move offscreen). However, when I change the zoom level, the view port / ScreenBox seems to reflect only the upper left section of the view. This is illustrated by native views getting stuck on the map even when the underlying annotation is still technically visible -- at this point, `queryPointAnnotations` is returning less annotations than I would expect (and sometimes 0 annotations). 

![bounds](https://cloud.githubusercontent.com/assets/63186/17745086/6398f53a-645f-11e6-886f-05ede8539fbb.gif)

Another issue is that the result of `queryPointAnnotations\queryRenderedFeatures` seems to lag behind the map movement. This is particularly noticeable when changing zoom level. Annotation views will maintain their `center` for the last zoom level and then snap back suddenly.

![zoom](https://cloud.githubusercontent.com/assets/63186/17745173/aee74136-645f-11e6-9382-dae033d281e0.gif)

[...]

PR #6061, opened 2016-08-17
Introduce visible annotations API

### LINKED ISSUE #5467: Accessor for visible annotation views in the viewport
After MGLAnnotationView was introduced, it became possible to implement advanced presentations for annotations in the map. One of the possibilities is to adjust transparency of some layers in multilayered image depending on the zoom level (which is important feature for Stay.com).
At the same time to achieve such effect client needs to re-add the same annotations on every event that may trigger the need to refresh the UI (pan/zoom).

Please consider adding the following methods:
```
-[MGLMapView visibleAnnotationViewsInBounds:]
```
or
```
-[MGLMapView reloadAnnotationsinBounds:]
```
The first one is more preferable as it will not require the client to reallocate MGLAnnotationView

boundsj:
This adds an API for getting all annotations that are currently visible on the map. We can also add an additional API to query by a specific rect. That API can be used internally to power this one.

This also adds a new option to iosapp to test the feature. 

Fixes #5467

### DIFF
--- a/platform/ios/app/MBXViewController.m
+++ b/platform/ios/app/MBXViewController.m
@@ -46,6 +46,7 @@ typedef NS_ENUM(NSInteger, MBXSettingsAnnotationsRows) {
     MBXSettingsAnnotations10000Sprites,
     MBXSettingsAnnotationsTestShapes,
     MBXSettingsAnnotationsCustomCallout,
+    MBXSettingsAnnotationsQueryAnnotations,
     MBXSettingsAnnotationsRemoveAnnotations,
 };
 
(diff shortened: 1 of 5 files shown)


1ec5:
MGLMapView's accessibility hint also says the number of visible annotations. It currently does so using a different API. Maybe it should be migrated over to this API? Not sure.

boundsj:
Since `-[MGLMapView accessibilityAnnotationCount]` just goes after the count, it may be better to just leave it as is instead of using the new `visibleAnnotations` method that creates, populates, and returns an NSArray of annotations.

boundsj:
I rebased to include #6627 and this is a lot closer. The visible annotations query result works as expected except when the map is tilted. In that case, annotations are still returned as visible even when they are slightly offscreen. This does work now for all other cases (i.e. not tilted, with or without rotation)

boundsj:
Resuming this since https://github.com/mapbox/mapbox-gl-native/issues/6055 is fixed.

PR #6054, opened 2016-08-17
[android] Runtime Styling API: mutable GeoJsonSource

ivovandongen:
Fixes #5906
Fixes #6237

PR #6071, opened 2016-08-18
[core] extract GeoJSONOptions conversion from GeoJSONSource conversion

ivovandongen:
For re-use in https://github.com/mapbox/mapbox-gl-native/pull/6054

@tmpsantos Could you check if this impacts QT at all? Checked ios/tests.

### DIFF
--- a/include/mbgl/style/conversion/source.hpp
+++ b/include/mbgl/style/conversion/source.hpp
@@ -2,6 +2,7 @@
 
 #include <mbgl/style/conversion.hpp>
 #include <mbgl/style/conversion/geojson.hpp>
+#include <mbgl/style/conversion/geojson_options.hpp>
 #include <mbgl/style/conversion/tileset.hpp>
 #include <mbgl/style/source.hpp>
 #include <mbgl/style/sources/geojson_source.hpp>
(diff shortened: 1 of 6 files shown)


include/mbgl/style/conversion/source.hpp
```diff
@@ -103,63 +104,12 @@ struct Converter<std::unique_ptr<Source>> {
...
-
-        const auto clusterRadiusValue = objectMember(value, "clusterRadius");
-        if (clusterRadiusValue) {
-            if (toNumber(*clusterRadiusValue)) {
-                options.clusterRadius = static_cast<double>(*toNumber(*clusterRadiusValue));
-            } else {
-                return Error{ "GeoJSON source clusterRadius value must be a number" };
-            }
+        Result<GeoJSONOptions> options = convert<GeoJSONOptions>(value);
+        if (!options) {
```
mourner: Is this condition ever `true`?

include/mbgl/style/conversion/source.hpp
```diff
@@ -103,63 +104,12 @@ struct Converter<std::unique_ptr<Source>> {
...
+            return Error{ "Could not parse GeoJSONOptions" };
```
jfirebaugh: This should return the error from `options` rather than this more generic error message.

test/src/mbgl/test/conversion_stubs.hpp
```diff
@@ -0,0 +1,111 @@
...
+#include <mbgl/util/variant.hpp>
+
+#include <string>
+#include <map>
+
+namespace mbgl {
+namespace style {
+namespace conversion {
+
+class Holder;
```
tmpsantos: Can we rename `Holder` to something like `Value`? Usually on mbgl `Holder` is a temporary variable used to prevent something like a refcounted pointer to go out of scope.

tmpsantos:
:+1: thanks a lot for adding more utests!

ivovandongen:
@tmpsantos Ok to merge with the latest changes?

tmpsantos:
> @tmpsantos Ok to merge with the latest changes?

From my side yes, I would just squash commits that are related to each other.

mourner:
Looks great to me, feel free to merge after rebase.

PR #6072, opened 2016-08-18
[Qt] Implement a Qt headless renderer

tmpsantos:
Using legacy QGLWidget because we support Qt4.


brunoabinader:
:ship:

tmpsantos:
Android test that is failing is flaky. /cc @tobrun

PR #6075, opened 2016-08-18
[ios, macos] fixes #6056 backwards compatibility macro

### LINKED ISSUE #6056: MGLStyleAttributeFunction.stops should use NS_DICTIONARY_OF macro
[MGLStyleAttributeFunction.stops](https://github.com/mapbox/mapbox-gl-native/blob/bca9d091805dc01a4456ab3f24e9de87f9b4aa48/platform/darwin/src/MGLStyleAttributeFunction.h#L16) is currently defined as a plain `NSDictionary<NSNumber *, id>`, but for backwards compatibility we need it to use our [`NS_DICTIONARY_OF`](https://github.com/mapbox/mapbox-gl-native/blob/bca9d091805dc01a4456ab3f24e9de87f9b4aa48/platform/darwin/src/MGLTypes.h#L65) macro.

/cc @frederoni @1ec5

frederoni:
Fixes #6056 

👀  @friedbunny

### DIFF
--- a/platform/darwin/src/MGLStyleAttributeFunction.h
+++ b/platform/darwin/src/MGLStyleAttributeFunction.h
@@ -8,7 +8,7 @@ NS_ASSUME_NONNULL_BEGIN
 
 @interface MGLStyleAttributeFunction : NSObject <MGLStyleAttributeValue>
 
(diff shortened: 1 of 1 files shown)


friedbunny:
👍

Issue #5970, opened 2016-08-12
Add category convenience methods to NSNumber for style values (was: Runtime styling properties are too loosely typed)


1ec5:
We currently use the MGLStyleAttributeValue protocol to allow any style attribute property to be set to the appropriate value type or an instance of MGLStyleAttributeFunction. While this approach is slightly more strongly typed than `id`, preventing the developer from setting a property to e.g. a UIButton, it doesn’t keep the developer from setting the property to an NSString where an NSNumber is expected or an NSArray where an NSValue is expected. The attribute properties’ types should be more specific, allowing the compiler to catch common developer errors.

/cc @frederoni @incanus @jfirebaugh

1ec5:
## Solution 1: More protocols

We’d need type-specific protocols that each conform to MGLStyleAttributeValue:

* MGLColorStyleAttributeValue (implemented by NSColor, UIColor, and MGLStyleAttributeFunction)
* MGLNumberStyleAttributeValue (implemented by NSNumber and MGLStyleAttributeFunction)
* MGLArrayStyleAttributeValue (implemented by NSArray and MGLStyleAttributeFunction)
* Etc.

This isn’t a huge departure from what we’re already doing. It’s just a matter of defining some additional protocols and having the existing categories and properties adopt them.

1ec5:
## Solution 2: Lightweight generics

All properties would be of type MGLStyleAttributeValue, which should be a class with a single type parameter that determines the primitive underlying type of the value. Perhaps something like this:

```objc
@interface MGLStyleAttributeValue<T> : NSObject

+ (instancetype)valueWithRawValue:(T)rawValue;
+ (instancetype)valueWithFunctionType:(MGLStyleAttributeFunctionType)functionType property:(NSString *)property base:(NSNumber *)base stops:(NSDictionary<NSNumber *, MGLStyleAttributeValue<T> *> *)stops;

// MGLStyleAttributeFunctionTypeNone if created using +valueWithRawValue:
@property (nonatomic) MGLStyleAttributeFunctionType functionType;

- (T)rawValueAtZoomLevel:(double)zoomLevel;

@end
```

The conversion from Objective-C types to C++ types would occur in each individual property, but we’d factor out almost everything into macros similar to the current `MGLSetEnumProperty()` and `MGLGetEnumProperty()`. (I think we should consider doing that anyways.)

[...]

PR #6076, opened 2016-08-18
Expand and clean up runtime styling documentation

### LINKED ISSUE #5945: MGLStyleLayer headers should be audited for nullability
Per https://github.com/mapbox/mapbox-gl-native/pull/5727/files#r74335948, headers generated by MGLStyleLayer.h.ejs should be audited for nullability. That is, they should have `NS_ASSUME_NONNULL_BEGIN` and `NS_ASSUME_NONNULL_END` at either end, and each parameter and return type in between should be reviewed for whether they can be `nil`.

/cc @frederoni

### LINKED ISSUE #5949: References to enumeration values in runtime styling documentation are unprefixed
Per https://github.com/mapbox/mapbox-gl-native/pull/5727#discussion_r71796286, references to enumeration values in runtime styling API documentation comments are unprefixed and unformatted. For example, [the `circlePitchScale` documentation](https://github.com/mapbox/mapbox-gl-native/blob/a6ec5d1a5a25887669d31e1824e7a243de031a94/platform/darwin/src/MGLCircleStyleLayer.h#L53) refers to “The value `map`”, which should be “`MGLCircleStyleLayerCirclePitchScaleMap`”.

/cc @frederoni

### LINKED ISSUE #5950: Runtime styling documentation refers to pixels instead of points
Per https://github.com/mapbox/mapbox-gl-native/pull/5727#discussion_r71796539, runtime styling API documentation comments measure screen distances in “[pixels](https://github.com/mapbox/mapbox-gl-native/blob/a6ec5d1a5a25887669d31e1824e7a243de031a94/platform/darwin/src/MGLLineStyleLayer.h#L87)” and “[px](https://github.com/mapbox/mapbox-gl-native/blob/a6ec5d1a5a25887669d31e1824e7a243de031a94/platform/darwin/src/MGLFillStyleLayer.h#L23)”. This is incorrect on Apple platforms, where the fundamental screen unit is a point and there are 2–3 pixels to a point. We should replace any occurrence of “pixel” or “px” in the documentation with “point” and “pt”, respectively.

/cc @frederoni

### LINKED ISSUE #5951: References to attributes in runtime styling documentation use kebab-case
Per https://github.com/mapbox/mapbox-gl-native/pull/5727#discussion_r71796641, runtime styling API documentation comments refer to other attributes using [kebab-case](https://github.com/mapbox/mapbox-gl-native/blob/a6ec5d1a5a25887669d31e1824e7a243de031a94/platform/darwin/src/MGLFillStyleLayer.h#L23), which matches the style specification but not the actual API. The API uses camelCase; hyphens are disallowed in identifiers.

/cc @frederoni

1ec5:
This PR improves the runtime styling API’s generated documentation in several ways:

* Source and layer headers have been audited for nullability. Properties representing optional attributes are qualified with `null_resettable`. The default value is stated in documentation. (Fixes #5945.)
* References to other attributes now use camelCase for consistency with the Objective-C/Swift property rather than kebab-case as defined in the style specification. (Fixes #5951.)
* References to valid enumeration values are expanded into the full Objective-C symbol name for clarity. (Fixes #5949.)
* Documentation expresses screen distances in points instead of pixels. (Fixes #5950.)
* The documentation for each numeric property indicates its unit.
* Interdependencies between properties are described in documentation comments.
* Along the way, the code to specialize array types has been refactored. Previously, any new array-typed property we didn’t specifically account for would be treated as an offset, which wouldn’t necessarily be correct.

/cc @frederoni @friedbunny

### DIFF
--- a/platform/darwin/scripts/generate-style-code.js
+++ b/platform/darwin/scripts/generate-style-code.js
@@ -37,7 +37,7 @@ global.testImplementation = function (property, layerType) {
             return `layer.${objCName(property)} = MGLRuntimeStylingHelper.testColor;`;
         case 'array':
             return testArrayImplementation(property);
-        default: throw new Error(`unknown type for ${property.name}`)
+        default: throw new Error(`unknown type for ${property.name}`);
     }
 }
 
@@ -57,7 +57,7 @@ global.testGetterImplementation = function (property, layerType) {
             return `XCTAssertEqualObjects(gLayer.${objCName(property)}, MGLRuntimeStylingHelper.testColor);`;
         case 'array':
             return testGetterArrayImplementation(property);
-        default: throw new Error(`unknown type for ${property.name}`)
+        default: throw new Error(`unknown type for ${property.name}`);
     }
 }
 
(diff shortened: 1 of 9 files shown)


1ec5:
I’m not entirely sure about the use of `null_resettable` yet. To be sure, mbgl does return a default value for any property that is undefined in the style JSON but defined to have a default value in the style specification. But what happens when you attempt to set such a property to undefined at runtime? Does it revert to that default value or officially become undefined?

Also, we’ll need to revisit this decision if we go with solution 3 in #5970.

1ec5:
`null_resettable` semantics aren’t being honored by optional style properties, and in fact the semantics are pretty weird right now: #6095. For the time being, I’m going to revert all the properties to be `nullable`.

PR #6087, opened 2016-08-19
Add option to show depth buffer

kkaefer:
We are [showing the stencil buffer](https://github.com/mapbox/mapbox-gl-native/pull/4669) in the debug app, but there's no way to show the depth buffer.

PR #6094, opened 2016-08-19
[ios, macos] Move debug mask options to shared header

friedbunny:
~~This brings the depth buffer (#6087) and stencil buffer (#4669) debug modes to iOS~~ by moving `MGLMapDebugMaskOptions` to MGLTypes.h, which is shared between macOS and iOS targets.

~~These alternative buffer modes aren’t exposed in iosapp, but they are available in code.~~ These modes are excluded from iOS per https://github.com/mapbox/mapbox-gl-native/pull/6094#issuecomment-241101240.

The main advantage of this PR is now that we have slightly less duplicate code. 😁

/cc @kkaefer @1ec5

### DIFF
--- a/platform/darwin/src/MGLTypes.h
+++ b/platform/darwin/src/MGLTypes.h
@@ -58,7 +58,7 @@ typedef NS_OPTIONS(NSUInteger, MGLMapDebugMaskOptions) {
         drawing operations appear more prominent to help diagnose overdrawing.
         @note This option does nothing in Release builds of the SDK. */
     MGLMapDebugOverdrawVisualizationMask = 1 << 5,
-#if TARGET_OS_OSX
+#if !TARGET_OS_IPHONE
     /** The stencil buffer is shown instead of the color buffer.
         @note This option does nothing in Release builds of the SDK. */
     MGLMapDebugStencilBufferMask = 1 << 6,
--- a/platform/darwin/src/MGLTypes.h
+++ b/platform/darwin/src/MGLTypes.h
@@ -42,6 +42,30 @@ typedef NS_ENUM(NSUInteger, MGLUserTrackingMode) {
     MGLUserTrackingModeFollowWithCourse,
 };
 
+/** Options for enabling debugging features in an `MGLMapView` instance. */
+typedef NS_OPTIONS(NSUInteger, MGLMapDebugMaskOptions) {
+    /** Edges of tile boundaries are shown as thick, red lines to help diagnose
+        tile clipping issues. */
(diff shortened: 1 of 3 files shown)


1ec5:
This effectively adds `MGLMapDebugStencilBufferMask` and MGLMapDebugDepthBufferMask` options to the iOS SDK that should be accounted for in `-[MGLMapView setDebugMask:]`.

friedbunny:
Indeed, I’ll port that logic now.

1ec5:
This also puts pressure on us to maintain parity between the two SDKs when it comes to debug options, whereas the trend lately is to add options that aren’t currently supported in OpenGL ES. The extra two buffer options fall into that category. So for this change, we should use compiler macros to keep the options out of iOS. Otherwise, it wouldn’t make sense to add nonfunctional API.

friedbunny:
> Otherwise, it wouldn’t make sense to add nonfunctional API.

Excellent point. 😉

platform/darwin/src/MGLTypes.h
```diff
@@ -42,6 +42,32 @@ typedef NS_ENUM(NSUInteger, MGLUserTrackingMode) {
...
+        drawing operations appear more prominent to help diagnose overdrawing.
+        @note This option does nothing in Release builds of the SDK. */
+    MGLMapDebugOverdrawVisualizationMask = 1 << 5,
+#if TARGET_OS_OSX
+    /** The stencil buffer is shown instead of the color buffer.
+        @note This option does nothing in Release builds of the SDK. */
+    MGLMapDebugStencilBufferMask = 1 << 6,
+    /** The depth buffer is shown instead of the color buffer.
+        @note This option does nothing in Release builds of the SDK. */
+    MGLMapDebugDepthBufferMask = 1 << 7,
```
1ec5: So this is fine for now, but what happens when we do need to add an option that’s supported by both iOS and macOS? Should the value be `1 << 8`?

friedbunny: Is there a disadvantage to potentially skipping numbers? If they’re just an implementation detail, I think the centralization and explicitness of what is supported where is a win.

1ec5: I don’t think there’s necessarily a problem with skipping numbers. When we do implement these two options for OpenGL ES, I suppose we’ll be able to remove this guard pretty cleanly.

friedbunny:
Even though this PR became progressively less useful, I still think it’s a solid (if minor) refactor. Any objections to merging?

PR #6097, opened 2016-08-19
MGLStyle key-value coding compliance

### LINKED ISSUE #6003: Add layers property to MGLStyle
MGLStyle should have a KVO-compliant `layers` property of type NSMutableArray. The array accessor should call a new `mbgl::Map::getLayers()` that calls the existing `mbgl::style::Style::getLayers()`. KVO provides for [many getters and setters for individual items in a to-many property](https://developer.apple.com/library/ios/documentation/Cocoa/Conceptual/KeyValueCoding/Articles/AccessorConventions.html#//apple_ref/doc/uid/20002174-SW11), such as `-replaceObjectInLayersAtIndex:withObject:`.

/cc @frederoni

### LINKED ISSUE #6096: MGLMapView.style is not key-value compliant
MGLMapView’s `style` method is implemented as a factory method, returning a newly allocated MGLStyle object every time it’s called. But it should be a property and a KVO-compliant one at that: MGLMapView should own an MGLStyle object that it holds onto until `styleURL` changes.

/cc @frederoni

### LINKED ISSUE #6931: Menu item to toggle native/localized labels
macosapp should have a View ‣ Labels submenu that allows the user to switch between displaying labels in the local language and displaying them in the system language. This would largely be a port of #6819, except that it needs to be more robust than just modifying country label layers, since macosapp supports arbitrary styles.

1ec5:
This PR makes the runtime styling API’s entry points [key-value coding compliant](https://developer.apple.com/library/content/documentation/Cocoa/Conceptual/KeyValueCoding/Compliant.html#//apple_ref/doc/uid/20002172).

The `-[MGLMapView style]` factory method has been replaced with a KVC-compliant `style` property whose value changes only when reloading the style or changing the style URL. This makes the API more convenient in Swift and more predictable overall.

New `layers` and `sources` properties have been added to MGLStyle that make it easy to enumerate style layers or sources without having to hard-code style-specific identifiers in application code. See #5999, #7030, and #7031 for examples of how these properties would be used.

The `layers` array is ordered from front to back, for consistency with Cocoa, as opposed to mbgl’s back-to-front ordering. There are complementary `-insertLayer:aboveLayer:` and `-insertLayer:belowLayer:` methods, for consistency with MapKit’s `-[MKMapView insertOverlay:aboveOverlay:]` and `-[MKMapView insertOverlay:belowOverlay:]`. Please double-check my index arithmetic to make sure I didn’t make any off-by-one errors. The `sources` property is a unordered set. A dictionary mapping identifiers to source objects would better conform to the style specification but would encourage some antipatterns we want to avoid; see https://github.com/mapbox/mapbox-gl-native/issues/7011#issuecomment-259835261 for details.

The mutable `layers` and `sources` collections comply with KVC conventions for to-many relationships, which entails additional methods that `mbgl::Map` and `mbgl::style::Style` wouldn’t ordinarily provide. These KVC methods could be optimized further with corresponding methods in mbgl; however, memory management would become a lot less clear. For example, who owns an `mbgl::style::Layer` if it’s already part of a style but is included in an array passed into `-[MGLStyle setLayers:]`? Rather than muddle the pointer ownership model or implement complex array diffing, this PR sticks to the safest, most straightforward implementation despite a possible performance tradeoff.

As a real-world stress test of KVC compliance, a sidebar has been added to macosapp that displays the layers in the current style and updates the list automatically using Cocoa bindings. The sidebar can be opened and closed via View ‣ Show/Hide Layers or ⇧⌘L. Layers are shown with their identifiers and icons representing their types. Hidden layers are shown in italics. To toggle one or more layers’ visibility, select them and double-click on them. Or use the context menu, which also has an option to delete the selected layers.

Finally, to illustrate one common use of the new APIs, macosapp now has an option to localize all labels in the user’s preferred language, and the Style toolbar button displays the name of the current style when a custom style is shown.

outofplace  
_Hiding layers using the new Layers sidebar in macosapp._

* [x] Make MGLMapView’s `style` property KVC-compliant
* [x] Make layers API KVC-compliant
* [x] Reverse layers to go from front to back
* [x] Implement `-[MGLStyle insertLayer:aboveLayer:]` (reprising #6730)
* [x] Make sources API KVC-compliant
* [x] Fix race condition involving change notifications when switching styles
* [x] Issue change notifications for layers when adding shape annotations
* [x] Add layers sidebar to macosapp
* [x] Allow showing/hiding layers sidebar

Fixes #6003, fixes #6096, fixes #6931.

/cc @frederoni @jfirebaugh


include/mbgl/map/map.hpp
```diff
@@ -154,7 +154,9 @@ class Map : private util::noncopyable {
     void removeSource(const std::string& sourceID);
 
     // Layers
+    std::vector<style::Layer*> getLayers();
     style::Layer* getLayer(const std::string& layerID);
+    void setLayers(std::vector<style::Layer*>&);
```
jfirebaugh: `Map`/`Style` must have ownership of the layers, so this signature needs to be `void setLayers(std::vector<std::unique_ptr<style::Layer>>)`.

1ec5: Fixed in e43fbf4a20ab981261b422c37a583e5aac871d00.

src/mbgl/map/map.cpp
```diff
@@ -794,6 +798,16 @@ style::Layer* Map::getLayer(const std::string& layerID) {
     return nullptr;
 }
 
+void Map::setLayers(std::vector<std::unique_ptr<style::Layer>> layers) {
+    impl->view.activate();
```
jfirebaugh: Needs an `if (!impl->style)` guard and `impl->styleMutated = true;` like the other mutators.

platform/darwin/src/MGLStyle.mm
```diff
@@ -96,21 +100,133 @@ + (NSURL *)emeraldStyleURL {
...
+        [styleLayers addObject:styleLayer];
+    }
+    return styleLayers;
+}
+
+- (void)setLayers:(NS_MUTABLE_ARRAY_OF(MGLStyleLayer *) *)layers {
+    std::vector<std::unique_ptr<mbgl::style::Layer>> rawLayers;
+    rawLayers.reserve(layers.count);
+    for (MGLStyleLayer *layer in layers.reverseObjectEnumerator) {
+        rawLayers.push_back(std::unique_ptr<mbgl::style::Layer>(layer.layer));
```
jfirebaugh: Constructing a `std::unique_ptr` from anything other than the result of `new` is usually incorrect (and even then, you should be using `std::make_unique`). What's the ownership model of `layer.layer`?

1ec5: It’s complicated: #6254. 😕 Maybe we need to tackle that issue first.

1ec5: To respect ownership of the layers passed into `-setLayers:` – some of which may exist in the style and some of which may be new – we need to ability to remove a layer from the style and regain ownership of it in order to re-add it. So we’re blocked on `mbgl::Map::removeLayer()` returning ownership of the layer: #6959.

1ec5:
After rebasing, the point annotation layer `com.mapbox.annotations.points` no longer shows up when the first style loads, but it does appear as soon as you switch styles. Unfortunately, doing anything with the point annotation layer crashes or raises an uncaught exception because the raw layer pointer has gone stale.

```
2016-11-21 03:18:57.436 Mapbox GL[87889:2777666] An uncaught exception was raised
2016-11-21 03:18:57.436 Mapbox GL[87889:2777666] [<NSPopUpButton 0x608000164680> valueForUndefinedKey:]: this class is not key value coding-compliant for the key layers.
```

```
Thread 1Queue : com.apple.main-thread (serial)
#0	0x000000010043126b in mbgl::style::Layer::getVisibility() const at /path/to/mapbox-gl-native/src/mbgl/style/layer.cpp:18
#1	0x00000001000eccdc in ::-[MGLStyleLayer isVisible]() at /path/to/mapbox-gl-native/platform/darwin/src/MGLStyleLayer.mm:30
```

1ec5:
The crash was caused by the `MGLMapView.style` property getting set too early. Because setting via the property generates KVO change notifications, the application was enumerating the sources and layers while the old style was being torn down and the new one was being loaded. The fix was to bracket the changes to `style` and `styleURL` together.

frederoni:
Changing style twice throws an exception. Looks like an issue in the Style Object Controller?

```
2016-11-22 16:17:18.817048 Mapbox GL[17730:253771] [General] An uncaught exception was raised
2016-11-22 16:17:18.817099 Mapbox GL[17730:253771] [General] [<NSPopUpButton 0x608000162640> valueForUndefinedKey:]: this class is not key value coding-compliant for the key layers.
2016-11-22 16:17:18.836627 Mapbox GL[17730:253771] [General] (
	0   CoreFoundation                      0x00007fff9d9fe6fb __exceptionPreprocess + 171
	1   libobjc.A.dylib                     0x00007fffb2205a2a objc_exception_throw + 48
	2   CoreFoundation                      0x00007fff9d9fe649 -[NSException raise] + 9
	3   Foundation                          0x00007fff9f4e5842 -[NSObject(NSKeyValueCoding) valueForUndefinedKey:] + 226
	4   Foundation                          0x00007fff9f3b74dc -[NSObject(NSKeyValueCoding) valueForKey:] + 283
	5   Foundation                          0x00007fff9f3b70aa -[NSObject(NSKeyValueCoding) valueForKeyPath:] + 260
	6   AppKit                              0x00007fff9b65bd61 -[NSBinder valueForBinding:resolveMarkersToPlaceholders:] + 164
	7   AppKit                              0x00007fff9b821884 -[NSArrayDetailBinder _refreshDetailContentInBackground:] + 256
	8   Foundation                          0x00007fff9f3b2dbd NSKeyValueNotifyObserver + 350
	9   Foundation                          0x00007fff9f42a004 -[NSObject(NSKeyValueObservingPrivate) _notifyObserversForKeyPath:change:] + 990
	10  AppKit                              0x00007fff9b7d0161 -[NSController _notifyObserversForKeyPath:change:] + 194
	11  AppKit                              0x00007fff9b9a2863 -[NSObjectController setContent:] + 310
	12  AppKit                              0x00007fff9b9a2689 -[NSObjectDetailBinder refreshDetailContent] + 223
	13  Foundation                          0x00007fff9f3b2dbd NSKeyValueNotifyObserver + 350
	14  Foundation                          0x00007fff9f42a004 -[NSObject(NSKeyValueObservingPrivate) _notifyObserversForKeyPath:change:] + 990
	15  AppKit                              0x00007fff9b7d0161 -[NSController _notifyObserversForKeyPath:change:] + 194
	16  AppKit                              0x00007fff9b7d9016 -[NSController observeValueForKeyPath:ofObject:change:context:] + 919
	17  Foundation                          0x00007fff9f3b2dbd NSKeyValueNotifyObserver + 350
	18  Foundation                          0x00007fff9f3b2654 NSKeyValueDidChange + 486
	19  Foundation                          0x00007fff9f4f12d7 -[NSObject(NSKeyValueObservingPrivate) _changeValueForKeys:count:maybeOldValuesDict:usingBlock:] + 944
	20  Foundation                          0x00007fff9f3763c9 -[NSObject(NSKeyValueObservingPrivate) _changeValueForKey:key:key:usingBlock:] + 60
	21  Foundation                          0x00007fff9f3dec3b _NSSetObjectValueAndNotify + 261
	22  libsystem_trace.dylib               0x00007fffb2d14c41 _os_activity_initiate + 61
	23  AppKit                              0x00007fff9bd12d8e -[NSApplication(NSResponder) sendAction:to:from:] + 456
	24  AppKit                              0x00007fff9b85d5c4 -[NSControl sendAction:to:] + 86
	25  AppKit                              0x00007fff9b85d4ec __26-[NSCell _sendActionFrom:]_block_invoke + 136
	26  libsystem_trace.dylib               0x00007fffb2d14c41 _os_activity_initiate + 61
	27  AppKit                              0x00007fff9b85d444 -[NSCell _sendActionFrom:] + 128
	28  AppKit                              0x00007fff9b89fd6c -[NSButtonCell _sendActionFrom:] + 98
	29  libsystem_trace.dylib               0x00007fffb2d14c41 _os_activity_initiate + 61
	30  AppKit                              0x00007fff9b9b2c3b -[NSPopUpButtonCell _popUpItemAction:] + 87
	31  libsystem_trace.dylib               0x00007fffb2d14c41 _os_activity_initiate + 61
	32  AppKit                              0x00007fff9bd12d8e -[NSApplication(NSResponder) sendAction:to:from:] + 456
	33  AppKit                              0x00007fff9b84bf7e -[NSMenuItem _corePerformAction] + 324
	34  AppKit                              0x00007fff9b84bcea -[NSCarbonMenuImpl performActionWithHighlightingForItemAtIndex:] + 114
	35  libsystem_trace.dylib               0x00007fffb2d14c41 _os_activity_initiate + 61
	36  AppKit                              0x00007fff9b8d50a9 -[NSMenu performActionForItemAtIndex:] + 121
	37  AppKit                              0x00007fff9b8d5020 -[NSMenu _internalPerformActionForItemAtIndex:] + 94
	38  AppKit                              0x00007fff9b8d4e4d -[NSCarbonMenuImpl _carbonCommandProcessEvent:handlerCallRef:] + 107
	39  AppKit                              0x00007fff9b78bb44 NSSLMMenuEventHandler + 986
	40  HIToolbox                           0x00007fff9ceea125 _ZL23DispatchEventToHandlersP14EventTargetRecP14OpaqueEventRefP14HandlerCallRec + 1708
	41  HIToolbox                           0x00007fff9cee9396 _ZL30SendEventToEventTargetInternalP14OpaqueEventRefP20OpaqueEventTargetRefP14HandlerCallRec + 428
	42  HIToolbox                           0x00007fff9cefee14 SendEventToEventTarget + 40
	43  HIToolbox                           0x00007fff9cf4aeb6 _ZL18SendHICommandEventjPK9HICommandjjhPKvP20OpaqueEventTargetRefS5_PP14OpaqueEventRef + 411
	44  HIToolbox                           0x00007fff9cf764db SendMenuCommandWithContextAndModifiers + 59
	45  HIToolbox                           0x00007fff9cf7648a SendMenuItemSelectedEvent + 188
	46  HIToolbox                           0x00007fff9cf7635d _ZL19FinishMenuSelectionP13SelectionDataP10MenuResultS2_ + 96
	47  HIToolbox                           0x00007fff9cf5563b _ZL19PopUpMenuSelectCoreP8MenuData5PointdS1_tjPK4RecttjS4_S4_PK14__CFDictionaryPK10__CFStringPP13OpaqueMenuRefPt + 2007
	48  HIToolbox                           0x00007fff9cf546a3 _ZL26_HandlePopUpMenuSelection8P13OpaqueMenuRefP14OpaqueEventRefj5PointtjPK4RecttS6_S6_PK14__CFDictionaryPK10__CFStringPS0_Pt + 612
	49  HIToolbox                           0x00007fff9cf5427b _HandlePopUpMenuSelectionWithDictionary + 287
	50  AppKit                              0x00007fff9b8cbd7f _NSSLMPopUpCarbonMenu3 + 6218
	51  AppKit                              0x00007fff9b8ca47b -[NSCarbonMenuImpl popUpMenu:atLocation:width:forView:withSelectedItem:withFont:withFlags:withOptions:] + 456
	52  AppKit                              0x00007fff9b945a39 -[NSPopUpButtonCell trackMouse:inRect:ofView:untilMouseUp:] + 564
	53  AppKit                              0x00007fff9b85a6e3 -[NSControl mouseDown:] + 832
	54  AppKit                              0x00007fff9be736f3 -[NSWindow(NSEventRouting) _handleMouseDownEvent:isDelayedEvent:] + 6341
	55  AppKit                              0x00007fff9be6ff10 -[NSWindow(NSEventRouting) _reallySendEvent:isDelayedEvent:] + 1942
	56  AppKit                              0x00007fff9be6f3ae -[NSWindow(NSEventRouting) sendEvent:] + 541
	57  AppKit                              0x00007fff9bd0f4cd -[NSApplication(NSEvent) sendEvent:] + 1145
	58  AppKit                              0x00007fff9b5f0589 -[NSApplication run] + 1002
	59  AppKit                              0x00007fff9b5bb1ad NSApplicationMain + 1237
	60  Mapbox GL                           0x000000010000cb52 main + 34
	61  libdyld.dylib                       0x00007fffb2ae2255 start + 1
)
```

frederoni:
Layers on map and in the table seems to be off by one:
Street: 192/191
Dark: 152/151

1ec5:
> Changing style twice throws an exception. Looks like an issue in the Style Object Controller?

Yes, I’m looking into the issue now. The Layers sidebar works by binding to the map view’s style’s layers array, which means didChange notifications have to occur at the right time or the KVO system will raise exceptions like this. It’s a great stress test for catching race conditions, but unfortunately bindings make those race conditions difficult to debug.

> Layers on map and in the table seems to be off by one:
> Nvm, it was just an invisible symbol layer. 👍

This is the built-in point annotation layer. It used to show up reliably in `layers` when I issued the didChange notification in response to `mbgl::MapChangeDidFinishLoadingMap`, but it no longer appears now that the didChange notification is issued in `mbgl::MapChangeDidFinishLoadingStyle`.

1ec5:
Fixed the style-switching crash in 0656bfdd7955cdf3c624d2f1f9af095dc93e5cb5. It was related to the point annotation layer, as expected. The fix is to issue change notifications on `layers` both when the style finishes loading and when the map finishes loading, since `mbgl::AnnotationManager` adds that layer to the map behind the scenes before the map finishes loading.

1ec5:
And six hours later, the crash came back, even for the commit that wasn’t reproducing it before. I’ve ruled out anything involving the change notifications at this point. `MGLMapView.style` is inexplicably set to an NSMenuItem or NSPopUpButton!

Here’s what I see having broken at `-[NSObject(NSKeyValueCoding) valueForKeyPath:]` just before the exception is raised:

```
(lldb) po $arg1
<NSObjectController: 0x608000121b80>[object class: MGLMapView]

(lldb) po 0x608000121b80
<NSObjectController: 0x608000121b80>[object class: MGLMapView]

(lldb) po [0x608000121b80 selectedObjects]
<__NSArrayI 0x600000031ea0>(
<MGLMapView: 0x104302130>
)


(lldb) po [0x608000121b80 selectedObjects]
<__NSArrayI 0x600000031ea0>(
<MGLMapView: 0x104302130>
)


(lldb) po [[0x608000121b80 selectedObjects] firstObject]
<MGLMapView: 0x104302130>

(lldb) po [[[0x608000121b80 selectedObjects] firstObject] styleURL]
mapbox://styles/1ec5/civg44q2d00192kpis50oafog

(lldb) po [[[0x608000121b80 selectedObjects] firstObject] style]
<NSMenuItem: 0x6000000a1a40 Streets>
```

1ec5:
I found the root cause by setting a symbolic breakpoint on `-[MGLMapView setStyle:]`. The issue is that the menu items and popup button are bound to the first responder’s `-setStyle:` method. The idea was that MapDocument would be the first responder; this is the only way something in MainMenu.xib can trigger a MapDocument action. However, it turns out that MGLMapView can sometimes appear before MapDocument in the responder chain.

Publicly, MGLMapView’s `style` method is read-only, so application code can’t assume the presence of an `-[MGLMapView setStyle:]` method that would preempt `-[MapDocument setStyle:]`. However, it can’t assume that method’s absence, either. Indeed, [`style` is privately redeclared](https://github.com/mapbox/mapbox-gl-native/blob/297337ae8980e2838d759737e2899e747ac97c4b/platform/macos/src/MGLMapView.mm#L155) as a read-write property, which means `-[MGLMapView setStyle:]` does get autosynthesized.

The fix, then, is to rename `-[MapDocument setStyle:]` to something more distinct to ensure the first responder–targeted action always reaches MapDocument.

frederoni:
Great work here @1ec5 
This PR will help a lot with debugging and taking runtime styling even further.

src/mbgl/map/map.cpp
```diff
@@ -825,15 +829,34 @@ std::unique_ptr<Source> Map::removeSource(const std::string& sourceID) {
...
     if (impl->style) {
         impl->styleMutated = true;
         return impl->style->getLayer(layerID);
     }
     return nullptr;
 }
 
+void Map::setLayers(std::vector<std::unique_ptr<Layer>> layers) {
+    if (!impl->style) {
+        impl->styleMutated = true;
```
jfirebaugh: This goes in the opposite branch -- we're only mutating the style if we have one.

platform/darwin/src/MGLBackgroundStyleLayer.mm
```diff
@@ -51,6 +51,9 @@ - (void)addToMapView:(MGLMapView *)mapView belowLayer:(MGLStyleLayer *)otherLaye
 - (void)removeFromMapView:(MGLMapView *)mapView
 {
     auto removedLayer = mapView.mbglMap->removeLayer(self.identifier.UTF8String);
+    if (!removedLayer) {
+        return;
+    }
     _pendingLayer = std::move(reinterpret_cast<std::unique_ptr<mbgl::style::BackgroundLayer> &>(removedLayer));
```
jfirebaugh: Not part of this PR, but this cast is undefined behavior. You need to write something like:

```
template <class T, class U>
std::unique_ptr<T> dynamic_ptr_cast(std::unique_ptr<U> p) {
    T* t = dynamic_cast<T*>(p.get());
    if (!t) return nullptr;
    p.release();
    return t;
}
```

And be aware that this will return a null `unique_ptr` (and deallocate the input) if the `dynamic_cast` fails, i.e. if the input is not the anticipated derived type. In this particular code, it looks like this could happen if the user was aliasing `*StyleLayer` objects while at the same time removing a layer and adding another layer with the same identifier but a different layer type.

src/mbgl/map/map.cpp
1ec5: 👍 I ended up not using this method in the SDK due to the memory management conundrum in the PR description. Should I keep this method around to complement `getLayers()` or remove it?

1ec5: Fixed.

jfirebaugh: Yes, please remove it if it's not used.

platform/darwin/src/MGLBackgroundStyleLayer.mm
1ec5: → #7211

src/mbgl/map/map.cpp
1ec5: Removed.

Issue #6066, opened 2016-08-18
Offsets should be flipped vertically on macOS


1ec5:
In the style specification, the `y` value of a property like `fill-translate` is defined as the number of pixels from the top. By contrast, offsets are normally measured from the bottom of the screen or view on macOS. In order to simplify calculations, the value exposed to the developer via the runtime styling API should be consistent with the system’s standard origin. This will become especially important once we replace the NSArray representation with a `CGVector` (#5947).

/cc @frederoni

jfirebaugh:
Adding to the 3.4.0 milestone -- we need to implement this now or keep the existing behavior at least until a semver major version.

1ec5:
This only applies to macOS; the origin is at the top-left on iOS. The macOS SDK hasn't reached v1.0.0 yet, so I'm not terribly concerned about backwards compatibility there.

boundsj:
In that case, I'm going to remove the 3.4.0 milestone.

1ec5:
Fixed in #7636 on the release-ios-v3.4.0 branch.

PR #6107, opened 2016-08-22
Override style specification documentation in runtime styling API

### LINKED ISSUE #5954: Runtime styling documentation overrides
For #5953, https://github.com/mapbox/mapbox-gl-native/pull/5727#discussion_r71797844, and #5960, we should have an override table for documentation coming from the style specification. It could be a JSON directly inside platform/darwin/ that gets included by the code generation script.

/cc @frederoni @friedbunny

frederoni:
Fixes #5954 - Override some gl-style-spec documentation to make it more idiomatic.

👀  @1ec5 @friedbunny

### DIFF
--- a/platform/darwin/scripts/generate-style-code.js
+++ b/platform/darwin/scripts/generate-style-code.js
@@ -2,8 +2,9 @@
 
 const fs = require('fs');
 const ejs = require('ejs');
-const spec = require('mapbox-gl-style-spec').latest;
+const _ = require('lodash');
 const colorParser = require('csscolorparser');
+const spec = _.merge(require('mapbox-gl-style-spec').latest, require('./style-spec-overrides-v8.json'));
 
 const prefix = 'MGL';
 const suffix = 'StyleLayer';
--- a/platform/darwin/scripts/style-spec-overrides-v8.json
+++ b/platform/darwin/scripts/style-spec-overrides-v8.json
@@ -1,0 +1,55 @@
+{
+  "layout_symbol": {
+    "icon-text-fit-padding": {
+      "doc": "Size of the additional area added to dimensions determined by `icon-text-fit`."
+    },
+    "icon-offset": {
+      "doc": "Offset distance of icon from its anchor."
+    },
+    "text-offset": {
+      "doc": "Offset distance of text from its anchor."
+    },
+    "text-transform": {
+      "doc": "Specifies how to capitalize text."
+    }
+  },
+  "paint_background": {
+    "background-pattern": {
+      "doc": "Name of image in style images to use for drawing an image background. For seamless patterns, image width and height must be a factor of two (2, 4, 8, ..., 512)."
+    }
+  },
+  "paint_fill": {
+    "fill-pattern": {
+      "doc": "Name of image in style images to use for drawing image fills. For seamless patterns, image width and height must be a factor of two (2, 4, 8, ..., 512)."
+    }
+  },
+  "paint_line": {
+    "line-pattern": {
+      "doc": "Name of image in style images to use for drawing image lines. For seamless patterns, image width must be a factor of two (2, 4, 8, ..., 512)."
+    },
+    "line-translate": {
+      "doc": "The geometry's offset."
+    }
+  },
+  "paint_circle": {
+    "circle-translate": {
+      "doc": "The geometry's offset."
+    }
+  },
+  "paint_fill": {
+    "fill-translate": {
+      "doc": "The geometry's offset."
+    },
+    "fill-color": {
+      "doc": "The color of the filled part of this layer."
+    }
+  },
+  "paint_symbol": {
+    "icon-translate": {
+      "doc": "Distance that the icon's anchor is moved from its original placement."
+    },
+    "text-translate": {
+      "doc": "Distance that the text's anchor is moved from its original placement."
+    }
+  }
+}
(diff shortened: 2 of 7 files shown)


platform/darwin/scripts/generate-style-code.js
```diff
@@ -97,6 +97,16 @@ global.testArrayImplementation = function (property) {
     }
 };
 
+global.enumDoc = function (property, value) {
+    if ('docs' in property) {
+        return `    /**
+     ${property.docs[property.values.indexOf(value)]}
+     */`;
+    } else {
+        return "#warning documentation missing.";
```
1ec5: These warnings might be easier to track if they included the names of the enum values that are missing documentation.

platform/darwin/src/MGLSymbolStyleLayer.h
```diff
@@ -6,64 +6,109 @@
 
 NS_ASSUME_NONNULL_BEGIN
 
+/** Label placement relative to its geometry. `line` can only be used on LineStrings and Polygons. */
```
1ec5: These documentation comments need the same transformations we put in for properties in #6076.

platform/darwin/src/MGLCircleStyleLayer.h
```diff
@@ -6,13 +6,19 @@
 
 NS_ASSUME_NONNULL_BEGIN
 
+/** Control whether the translation is relative to the map (north) or viewport (screen) */
```
1ec5: Do we need the property's entire documentation? Something like "Circle translate anchor values for use with the `circleTranslateAnchor` property of MGLCirclrStyleLayer" might do the trick.

1ec5:
All the `#warning`s for enum values will conflict (but can go away) once #6508 lands. 🎉

/cc @incanus

frederoni:
Rebased and cleaned. `icon-text-fit-padding` is the only style spec documentation I could find that contradicts itself. @1ec5 👀

platform/darwin/scripts/generate-style-code.js
```diff
@@ -2,12 +2,28 @@
...
 
 const prefix = 'MGL';
 const suffix = 'StyleLayer';
 
+global.isObjectNotArray = function(object) {
+    return (typeof object === 'object' && !Array.isArray(object));
+}
+
+global.override = function(base, overrides) {
+    Object.keys(overrides).forEach(function(key) {
```
1ec5: This can be expressed as a for-of loop:

```js
for (var key of overrides) { … }
```

platform/darwin/scripts/generate-style-code.js
```diff
@@ -2,12 +2,28 @@
...
 const colorParser = require('csscolorparser');
 
 const prefix = 'MGL';
 const suffix = 'StyleLayer';
 
+global.isObjectNotArray = function(object) {
+    return (typeof object === 'object' && !Array.isArray(object));
+}
+
+global.override = function(base, overrides) {
```
1ec5: Instead of rolling our own function, import `lodash` and use [`_.assign()`](https://lodash.com/docs/4.16.6#assign) or [`_.assignIn()`](https://lodash.com/docs/4.16.6#assignIn).

platform/darwin/scripts/style-spec-overrides-v8.json
```diff
@@ -0,0 +1,7 @@
+{
+  "layout_symbol": {
+    "icon-text-fit-padding": {
+      "doc": "Size of the additional area added to dimensions determined by `icon-text-fit`, in counterclockwise order: top, left, bottom, right."
```
1ec5: How about:

> Padding added to the dimensions determined by the `iconTextFit` property.

We already mention in the generated documentation that the NSValue contains an NSEdgeInsets or UIEdgeInsets. The developer can look up what the expected order is for those types.

1ec5:
Some additional properties that we should fix:

* [ ] Instead of “sprite”, which is a term we’re trying to phase out, refer to “style images” and `-[MGLStyle setImage:forName:]`.
  * [ ] `MGLBackgroundStyleLayer.backgroundPattern`
  * [ ] `MGLFillStyleLayer.fillPattern`
  * [ ] `MGLLineStyleLayer.linePattern`
* [ ] Remove the “Values are…” sentence, which talks about an array; moreover, the negative Y part will be incorrect for macOS (#6066). The developer can always refer to CGVector documentation.
  * [ ] `MGLCircleStyleLayer.circleTranslate`
  * [ ] `MGLFillStyleLayer.fillTranslate`
  * [ ] `MGLLineStyleLayer.lineTranslate`
* [ ] Remove the “Positive values…” sentence, which will be incorrect on macOS.
  * [ ] `MGLSymbolStyleLayer.iconOffset`
  * [ ] `MGLSymbolStyleLayer.textOffset`
  * [ ] `MGLSymbolStyleLayer.iconTranslate`
  * [ ] `MGLSymbolStyleLayer.textTranslate`
* [ ] Remove “as `rgba`”, which is CSS syntax.
  * [ ] `MGLFillStyleLayer.fillColor`
* [ ] Remove “similar to the CSS `text-transform` property”.
  * [ ] `MGLTextTransform`
  * [ ] `MGLSymbolStyleLayer.textTransform`

It may be possible to fix these properties without an override, similar to how we replace “pixels” with “points”.

PR #6110, opened 2016-08-22
Correctly composite CustomLayers

kkaefer:
We are currently not setting the correct depth settings for custom layers, so they are potentially painting over layers that are displayed above it.

Here's a test case (blue water = below, yellow landcover = above, transparent green triangle = custom layer in the middle).

Before:
actual

After:
expected

This happens due to the way we are drawing some layers: Solid fills are drawn first top to bottom (!) to take advantage of fragment culling when the topmost fragment is opaque. However, CustomLayers are always treated as potentially translucent since we can't know what parts of the viewport they cover, so they are drawn in the second bottom-to-top rendering phase and could potentially cover fragments from the original solid phase.


jfirebaugh:
LGTM, modulo the CI failure.

PR #6053, opened 2016-08-22
Add ability to render to offscreen texture

kkaefer:
We should add facilities that allows rendering individual parts to an offscreen texture and composite them later.


kkaefer:
Please review

PR #6120, opened 2016-08-23
Stop exaggerating heading accuracy indicator size

### LINKED ISSUE #3718: Heading accuracy indicator is too wide
The user dot’s heading accuracy indicator appears to be twice as wide as it should be, assuming MapKit’s representation is correct:

mapbox mapkit

/cc @friedbunny

friedbunny:
This change removes the 1.5× exaggeration of the user location annotation’s heading indicator.

Also cleaned up some of the heading mask calculations while I was at it.

Fixes #3718.

/cc @1ec5


friedbunny:
Unsure if this rises to changelog status. Probably? Haven’t added anything yet.

1ec5:
> Unsure if this rises to changelog status.

Sure, why not. There’s visual impact to the user.

1ec5:
👍

friedbunny:
Added a changelog entry.

I also realized that I hadn’t changed the maximum accuracy clipping size, so I went back and updated that to 20° (`90 - 70`). We now appear as accurate as Apple:

img_4474

(Google doesn’t represent heading accuracy and uses a static heading arrow.)

friedbunny:
Note: I’m seeing reported accuracy as low as 10°, but we should still clip to 20° for display purposes.

PR #6122, opened 2016-08-23
Disable implicit animation of user annotation pitch changes

friedbunny:
Subtly improves the performance of the dot/puck’s shadow by disabling its implicit animation — it no longer lags behind when you’re executing the pitch gesture.

Follows on from #6019.

/cc @1ec5


1ec5:
👍

PR #6124, opened 2016-08-23
[core] Add static getters for property default values

### LINKED ISSUE #6117: mbgl::Layer subclasses should have accessors for implicit default values
Per https://github.com/mapbox/mapbox-gl-native/issues/6095#issuecomment-241550595, `mbgl::Layer`’s subclasses should have accessors for the implicit default values used when the actual values of layout or paint properties are set to undefined. This would enable the iOS SDK to implement `null_resettable` semantics for the runtime styling API, as described in #6095.

/cc @jfirebaugh @frederoni @ivovandongen

jfirebaugh:
Fixes #6117.

/cc @1ec5 @frederoni @ivovandongen

### DIFF
--- a/include/mbgl/style/layers/background_layer.hpp
+++ b/include/mbgl/style/layers/background_layer.hpp
@@ -18,12 +18,15 @@ public:
 
     // Paint properties
 
+    static PropertyValue<Color> getDefaultBackgroundColor();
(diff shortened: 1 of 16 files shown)


1ec5:
👍

PR #6125, opened 2016-08-23
Distinguish between blackColor and clearColor

1ec5:
Distinguish between `{NS,UI}Color.blackColor` and `{NS,UI}Color.clearColor` in documentation comments for runtime styling API properties. In #6124, generate-style-code.js has the same special cases for clear, black, and white.

/cc @frederoni


frederoni:
👍

PR #6127, opened 2016-08-23
Expose name and default position of current style

1ec5:
This is a followup to #6002 that exposes the current style’s name and makes `-[MGLMapView resetPosition]` reset to the style’s default viewport rather than the SDK default.

/cc @friedbunny

### DIFF
--- a/platform/ios/src/MGLMapView.mm
+++ b/platform/ios/src/MGLMapView.mm
@@ -1899,9 +1899,13 @@ public:
 
 - (void)resetPosition
 {
-    MGLMapCamera *camera = [MGLMapCamera camera];
-    camera.altitude = MGLAltitudeForZoomLevel(0, 0, 0, self.frame.size);
-    self.camera = camera;
+    CGFloat pitch = _mbglMap->getDefaultPitch();
+    CLLocationDirection heading = mbgl::util::wrap(_mbglMap->getDefaultBearing(), 0., 360.);
+    CLLocationDistance distance = MGLAltitudeForZoomLevel(_mbglMap->getDefaultZoom(), pitch, 0, self.frame.size);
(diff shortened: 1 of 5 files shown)


1ec5:
I punted on jumping to the style’s default viewport on initialization, because the style may take an undefined amount of time to load. Would it be desirable to jump to that position as soon as the style finishes loading? We’d need to avoid clobbering any developer- or user-initiated viewport modifications. The same complications would arise if we were to respect the style’s default viewport when switching styles.

jfirebaugh:
> Would it be desirable to jump to that position as soon as the style finishes loading? We’d need to avoid clobbering any developer- or user-initiated viewport modifications.

That's the behavior of GL JS. My preference would be to implement this behavior in core.

1ec5:
Split out #6129 to consider jumping to the default viewport on initialization.

platform/ios/src/MGLMapView.mm
```diff
@@ -1899,9 +1899,13 @@ - (void)resetNorthAnimated:(BOOL)animated
 
 - (void)resetPosition
 {
-    MGLMapCamera *camera = [MGLMapCamera camera];
-    camera.altitude = MGLAltitudeForZoomLevel(0, 0, 0, self.frame.size);
-    self.camera = camera;
+    CGFloat pitch = _mbglMap->getDefaultPitch();
+    CLLocationDirection heading = mbgl::util::wrap(_mbglMap->getDefaultBearing(), 0., 360.);
```
boundsj: It looks like studio wraps bearings. Do we do this since we don't know where the style will actually come from? (just curious)

boundsj:
👍

platform/ios/src/MGLMapView.mm
1ec5: Correct, we can’t assume the style was authored in Studio. Alternatively, we could establish a convention that `mbgl::Map` wraps the return values of `getDefeaultBearing()` and `getBearing()`, but we can tackle that separately from this PR.

PR #6140, opened 2016-08-24
[ios] Fix typo

tmpsantos:
s/Reponse/Response


frederoni:
Should be done in the dependency but this is correct since it's copied/checked in.
(https://github.com/AliSoftware/OHHTTPStubs)

👍

tmpsantos:
> Should be done in the dependency but this is correct since it's copied/checked in.
> (https://github.com/AliSoftware/OHHTTPStubs)

Thanks for pointing out. I will make a PR there too.

tmpsantos:
The API was fixed upstream already, just the docs were left.

https://github.com/AliSoftware/OHHTTPStubs/pull/198

friedbunny:
Do you recall why OHHTTPStubs isn’t a git submodule, @1ec5? Submodules can be a headache, but we’re probably always going to be reliant on them in this project.

1ec5:
I don't think there's a good reason it's checked in. We could certainly turn it into a submodule.

PR #6142, opened 2016-08-24
[core] Emit MapChangeDidFailLoadingMap when the style cannot be loaded or parsed

tmpsantos:
Currently this signal is never emitted, which can cause the Still mode to starve in case of an invalid style or failed request.


1ec5:
Awesome, I think this will partially fix #2762.

/cc @friedbunny

jfirebaugh:
👍

1ec5:
Took care of some tail work on iOS and macOS in #6145.

PR #6145, opened 2016-08-24
Document -mapViewDidFailLoadingMap:withError:

1ec5:
Documented `-[MGLMapViewDelegate mapViewDidFailLoadingMap:withError:]` on iOS and ported the declaration and call site over to macOS. Also mentioned the sudden new use of this API in the changelogs.

/cc @friedbunny


friedbunny:
Awesome. 👍

PR #6151, opened 2016-08-25
[node] Throw on invalid style JSON

### LINKED ISSUE #6143: [node] Style parse errors are not handled
Parsing errors won't throw an exception but rather a `MapChange` signal. Node bindings will never detect this error and potentially starve waiting.

Depends on #6142

/cc @mikemorris

tmpsantos:
Fixes #6143

### DIFF
--- a/platform/node/src/node_map.cpp
+++ b/platform/node/src/node_map.cpp
@@ -740,6 +740,12 @@ NodeMap::NodeMap(v8::Local<v8::Object> options) :
     map(std::make_unique<mbgl::Map>(view, *this, mbgl::MapMode::Still)),
     async(new uv_async_t) {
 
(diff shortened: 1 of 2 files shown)


mikemorris:
Hrm, it would be nice if `map.load` took a callback arg that it could pass an error to instead of needing to wrap this in a try/catch in implementing code...

tmpsantos:
> I think it's actually async already despite the API design?

The underlying `Map::setStyleJSON` so that makes `map.load` sync too. Note that the exception is thrown at https://github.com/mapbox/mapbox-gl-native/blob/8c2b185ce111cde7c87e588dde8fcdd1eb3e128b/platform/node/src/node_map.cpp#L207

mikemorris:
:shipit:

Issue #6133, opened 2016-08-24
Anchoring not applied with builder


tobrun:
The following is not working correctly:

```java
PulseMarkerViewOptions background = new PulseMarkerViewOptions().position(new LatLng(38.897654, -77.036589)).anchor(0.5f,0.5f);
PulseMarkerView markerView = (PulseMarkerView) mapboxMap.addMarker(background);
```

As workaround we can use:

```java
PulseMarkerViewOptions background = new PulseMarkerViewOptions().position(new LatLng(38.897654, -77.036589));
PulseMarkerView markerView = (PulseMarkerView) mapboxMap.addMarker(background);
markerView.setAnchor(0.5f,0.5f);
```

Issue #6132, opened 2016-08-24
Alpha not applied in MarkerView builder


tobrun:
Sample code that not works:
```java
PulseMarkerViewOptions background = new PulseMarkerViewOptions().position(new LatLng(38.897904, -77.036915)).alpha(0.5f);
PulseMarkerView markerView = (PulseMarkerView) mapboxMap.addMarker(background);
```

Workaround for now:
```java
PulseMarkerViewOptions background = new PulseMarkerViewOptions().position(new LatLng(38.897904, -77.036915)).alpha(0.3f);
PulseMarkerView markerView = (PulseMarkerView) mapboxMap.addMarker(background);
markerView.setAlpha(0.5f);
```

tobrun:
The issue was two fold: One the value was not applied by the parent constructor, two the alpha value was reset by an animation.

PR #6153, opened 2016-08-25
[android] added missing rotation, cleanup duplicate entries

tobrun:
Refs #6132 and #6133 

Review @cammace


ivovandongen:
LGTM!

PR #6163, opened 2016-08-25
Ignore non-conforming user location annotation views

### LINKED ISSUE #6100: -[MGLMapViewDelegate mapView:viewForAnnotation:] causes assertion failure if upgrading from v3.3.x
Prior to #5882, `-[MGLMapViewDelegate mapView:viewForAnnotation:]` was only called for “user-contributed” annotations, never for the user location annotation. Now that it’s additionally called for the user location annotation, the most obvious implementation of this method will trip [this assertion](https://github.com/mapbox/mapbox-gl-native/blob/98c6987af4bca8815c9dbafb7da83c0e7c0a7a5f/platform/darwin/src/MGLSymbolStyleLayer.h#L160) unless MGLUserLocationAnnotationView or a user-contributed annotation class is specifically accounted for. Instead of asserting, we should simply ignore the returned annotation view. We can consider making it a hard assertion in a future release.

/cc @friedbunny @frederoni

friedbunny:
Prior to v3.4.0, a developer could indiscriminately return any UIView-conforming object in `mapView:viewForAnnotation:` and everything would be fine. With custom user location annotation views, we now also pass in a `MGLUserLocation` annotation, which requires a view that conforms to `MGLUserLocationAnnotationView`.

This pull request ignores views that the developer nominated for the user location annotation view that are not actually a kind of `MGLUserLocationAnnotationView`. This changes the alpha.2 behavior of asserting when the inheritance was wrong. When we ignore a view, the default user dot is used in its stead.

I’ve also documented that `mapView:viewForAnnotation:` can be used to customize the user location annotation view, because this is a somewhat unexpected new behavior.

Fixes #6100.

/cc @1ec5 @frederoni

### DIFF
--- a/platform/ios/src/MGLMapView.mm
+++ b/platform/ios/src/MGLMapView.mm
@@ -3824,10 +3824,14 @@ public:
         if ([self.delegate respondsToSelector:@selector(mapView:viewForAnnotation:)])
         {
             userLocationAnnotationView = (MGLUserLocationAnnotationView *)[self.delegate mapView:self viewForAnnotation:self.userLocation];
-            if (userLocationAnnotationView)
+            if (userLocationAnnotationView && ! [userLocationAnnotationView isKindOfClass:MGLUserLocationAnnotationView.class])
             {
-                NSAssert([userLocationAnnotationView.class isSubclassOfClass:MGLUserLocationAnnotationView.class],
-                         @"User location annotation view must be a subclass of MGLUserLocationAnnotationView");
+                static dispatch_once_t onceToken;
+                dispatch_once(&onceToken, ^{
+                    NSLog(@"Ignoring user location annotation view with type %@. User location annotation view must be a kind of MGLUserLocationAnnotationView. This warning is only shown once and will become an error in a future version.", NSStringFromClass(userLocationAnnotationView.class));
+                });
+
+                userLocationAnnotationView = nil;
             }
         }
         
--- a/platform/ios/src/MGLMapViewDelegate.h
+++ b/platform/ios/src/MGLMapViewDelegate.h
@@ -273,6 +273,11 @@ NS_ASSUME_NONNULL_BEGIN
  Touch frameworks. On the other hand, static annotation images use less memory
  and draw more quickly than annotation views.
  
+ The user location annotation view can also be customized via this method. When
+ `annotation` is an instance of `MGLUserLocation` (or equal to the map view’s
+ `userLocation` property), return an instance of `MGLUserLocationAnnotationView`
+ (or a subclass thereof).
+
  @param mapView The map view that requested the annotation view.
  @param annotation The object representing the annotation that is about to be
     displayed.
--- a/platform/ios/app/MBXViewController.m
+++ b/platform/ios/app/MBXViewController.m
@@ -732,15 +732,17 @@ static NSString * const MBXViewControllerAnnotationViewReuseIdentifer = @"MBXVie
         annotationView = [[MBXAnnotationView alloc] initWithReuseIdentifier:MBXViewControllerAnnotationViewReuseIdentifer];
         annotationView.frame = CGRectMake(0, 0, 10, 10);
         annotationView.backgroundColor = [UIColor whiteColor];
-        
-        // uncomment to make the annotation view draggable
-        // also note that having two long press gesture recognizers on overlapping views (`self.view` & `annotationView`) will cause weird behaviour
-        // comment out the pin dropping functionality in the handleLongPress: method in this class to make draggable annotation views play nice
+
+        // Note that having two long press gesture recognizers on overlapping
+        // views (`self.view` & `annotationView`) will cause weird behaviour.
+        // Comment out the pin dropping functionality in the handleLongPress:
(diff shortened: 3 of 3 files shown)


platform/ios/src/MGLMapViewDelegate.h
```diff
@@ -273,6 +273,10 @@ NS_ASSUME_NONNULL_BEGIN
  Touch frameworks. On the other hand, static annotation images use less memory
  and draw more quickly than annotation views.
  
+ The user location annotation view can also be customized via this method. When
+ `annotation` is `MGLUserLocation`, return an instance of a custom
+ `MGLUserLocationAnnotationView` subclass.
```
1ec5: Returning an MGLUserLocationAnnotationView should be acceptable too, thanks to https://github.com/mapbox/mapbox-gl-native/pull/6163/files#diff-20502b6bc02986a4398d0da84b2d4b94R3827.

platform/ios/src/MGLMapView.mm
```diff
@@ -3824,10 +3824,12 @@ - (void)setShowsUserLocation:(BOOL)showsUserLocation
...
         {
             userLocationAnnotationView = (MGLUserLocationAnnotationView *)[self.delegate mapView:self viewForAnnotation:self.userLocation];
-            if (userLocationAnnotationView)
+            if ( ! [userLocationAnnotationView isKindOfClass:MGLUserLocationAnnotationView.class])
             {
-                NSAssert([userLocationAnnotationView.class isSubclassOfClass:MGLUserLocationAnnotationView.class],
-                         @"User location annotation view must be a subclass of MGLUserLocationAnnotationView");
+                // Ignore the user-provided annotation if it is not a kind of
+                // MGLUserLocationAnnotationView (and use the default user
+                // location annotation instead).
```
1ec5: Should we log a message to the console that this will be disallowed in a future release? Or we are fine with this behavior going forward?

platform/ios/src/MGLMapViewDelegate.h
friedbunny: You are correct! How about this, then:

```
 The user location annotation view can also be customized via this method. When
 `annotation` is `MGLUserLocation`, return an instance of
 `MGLUserLocationAnnotationView` (or a subclass thereof).
```

1ec5: Sounds good. To be pedantic:

s/is MGLUserLocation/is an instance of MGLUserLocation (or equal to the map view's userLocation property)/

platform/ios/src/MGLMapView.mm
friedbunny: I think we should raise an exception in a future version, but I was worried about log spam. In testing, this message should only be shown when the user dot is initialized.

Here’s the log message I came up with:

> Ignoring incorrectly-typed user location annotation view. User location annotation view must be a kind of MGLUserLocationAnnotationView. This warning will become an error in a future version.

platform/ios/src/MGLMapViewDelegate.h
friedbunny: screen shot 2016-08-26 at 11 50 43 am

platform/ios/src/MGLMapView.mm
friedbunny: Added the log message in https://github.com/mapbox/mapbox-gl-native/pull/6163/commits/384274bc705eca374e99b966089e05f1ed0b88df#diff-20502b6bc02986a4398d0da84b2d4b94R3836.

friedbunny: Oh, that’s dumb — setting `userLocationAnnotationView = nil` and then checking to see if it exists.

platform/ios/src/MGLMapView.mm
```diff
@@ -3824,10 +3824,14 @@ - (void)setShowsUserLocation:(BOOL)showsUserLocation
...
+            if (userLocationAnnotationView && ! [userLocationAnnotationView isKindOfClass:MGLUserLocationAnnotationView.class])
             {
-                NSAssert([userLocationAnnotationView.class isSubclassOfClass:MGLUserLocationAnnotationView.class],
-                         @"User location annotation view must be a subclass of MGLUserLocationAnnotationView");
+                // Ignore the user-provided annotation if it is not a kind of
+                // MGLUserLocationAnnotationView (and use the default user
+                // location annotation instead).
+                userLocationAnnotationView = nil;
+
+                NSLog(@"Ignoring incorrectly-typed user location annotation view. User location annotation view must be a kind of MGLUserLocationAnnotationView. This warning will become an error in a future version.");
```
1ec5: To avoid spamming the console, this message should only occur once per map view or once per process (and it should say so).

1ec5: Also, this message should include the view’s actual class, to aid in debugging.

friedbunny: https://github.com/mapbox/mapbox-gl-native/pull/6163/commits/af771e0294543d707b596ad84ae126e000170583 limits the warning to being shown once and adds the name of the incorrect view to the text.

Issue #5701, opened 2016-07-15
Setting a layout property needs to trigger tile reparse


jfirebaugh:
When a layout property is modified, all existing tiles for that source need to be reparsed, so that vertex arrays that are calculated based on that property are recalculated.

This will fix https://github.com/mapbox/mapbox-gl-native/issues/5610#issuecomment-233004625.

Additionally, once data-driven-styling lands, modifying a data-driven paint property also needs to trigger a reparse.

We should port the general strategy for batched updates from gl-js -- see https://github.com/mapbox/mapbox-gl-js/pull/2355 and the current [style.js](https://github.com/mapbox/mapbox-gl-js/blob/master/js/style/style.js).

cc @ivovandongen @mourner @kkaefer

kkaefer:
Right now, we're throwing away the parsed representation of a tile after we created all buckets, and when reparsing a tile, we'll have to do a new request (which is fast because it'll likely be served from cache, but still slower because it needs to be decompressed etc.). We should change this behavior and instead have all tiles/buckets hang on to the data they loaded so that we can do faster reparses.

mkv123:
Is there some workaround to force the reparse after changing runtime styles until this gets fixed?

incanus:
Adding and removing layers seems to cause reloads, so perhaps a dummy (possibly even unstyled?) layer could help here.

mkv123:
Unfortunately I can't seem to get things to re-render even by adding a layer. However in trying it I got this simple way to replicate the issue, at least on my Nexus 5X with 6.0.1.

```java
map.getMapAsync(new com.mapbox.mapboxsdk.maps.OnMapReadyCallback() {
            @Override
            public void onMapReady(final MapboxMap mapboxMap) {

                mapboxMap.setCameraPosition(new CameraPosition.Builder().target(new LatLng(60.186132, 24.827881)).zoom(18).build());
                addOnMapChangedListener(new com.mapbox.mapboxsdk.maps.MapView.OnMapChangedListener() {
                    @Override
                    public void onMapChanged(int change) {
                        if(change == DID_FINISH_LOADING_MAP) {
                            FillLayer layer = new FillLayer("force-refresh", "composite");
                            layer.setSourceLayer("landuse");
                            layer.setFilter(Filter.eq("class", "grass"));
                            layer.setProperties(PropertyFactory.fillColor("#ff0000"));
                            layer.setMinZoom(9);
                            layer.setMaxZoom(22);
[...]

PR #6164, opened 2016-08-25
Preparatory cleanup for tile reloading

jfirebaugh:
Various cleanups to code I was reviewing prior to working on #5701.


mourner:
The GeoJSON-related changes look great to me.

PR #6166, opened 2016-08-26
avoid altering annotation view transforms set by user

incanus:
I unearthed a bug where trying to set an annotation view `transform` of your own gets rapidly reset nearly every frame while annotation view `center` is changing due to the `CATransform3DIdentity` that is repeatedly set in our `setCenter:` override. 

One case this doesn't cover is when you _do_ have draggable views but still want to set a `transform`. Perhaps we could create an alternate setter called `setCenterDuringAnnotationUpdate:` which is called by `-[MGLMapView updateAnnotationViews]` to bypass any of this drag-related `transform` setting entirely? 

@1ec5 @friedbunny @boundsj

### DIFF
--- a/platform/ios/CHANGELOG.md
+++ b/platform/ios/CHANGELOG.md
@@ -39,6 +39,7 @@ Mapbox welcomes participation and contributions from everyone. Please read [CONT
 * To make an MGLPolyline or MGLPolygon span the antimeridian, specify coordinates with longitudes greater than 180° or less than −180°. ([#6088](https://github.com/mapbox/mapbox-gl-native/pull/6088))
 * Improved the performance of relocating a non-view-backed point annotation by changing its `coordinate` property. ([#5385](https://github.com/mapbox/mapbox-gl-native/pull/5385))
 * Improved the precision of annotations at zoom levels greater than 18. ([#5517](https://github.com/mapbox/mapbox-gl-native/pull/5517))
+* Fixed an issue that could reset user-added transformations on annotation views. ([#6166](https://github.com/mapbox/mapbox-gl-native/pull/6166))
 
 ### Other changes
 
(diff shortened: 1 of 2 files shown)


1ec5:
`-updateTransform` isn’t about dragging (and is already a no-op while dragging). It’s about `scalesWithViewingDistance`, which is on by default and which this PR breaks.

1ec5:
If resetting to `CATransform3DIdentity` is a problem, one solution would be:

1. Keep track of the `CATransform3D` that we apply within `-updateTransform`.
1. Every time we call `-updateTransform`, reverse the previously-applied transform and apply the new one, without ever resetting to `CATransform3DIdentity`.

incanus:
Yeah I am thinking the manual bookkeeping might be necessary. Will keep on this.

incanus:
I think I fix things in https://github.com/mapbox/mapbox-gl-native/pull/6166/commits/c0edb621ac91699f968d5eeb3722dc0141158970. However @1ec5 I would also appreciate an eye on https://github.com/mapbox/mapbox-gl-native/pull/6166/commits/51aeb89e07d7d0a4a5e56c0d026e517ecfcf12cb to make sure that I didn't lose some intended effect by simply getting rid of the unconditional identify transform there.

1ec5:
I _think_ that should be OK. Incidentally, [`CALayer.sublayerTransform`](https://developer.apple.com/reference/quartzcore/calayer/1410888-sublayertransform) seems purpose-made for our intended effect, but it’s iOS 8+ API.

1ec5:
Also, would you do the honors and update the changelog?

incanus:
Not sure why Android test is failing, but going to rebase on master and merge this.

PR #6079, opened 2016-08-18
Offsets as vectors, padding as edge insets

1ec5:
An offset style attribute is now exposed publicly as an NSValue representing a `CGVector` instead of an NSArray of NSNumbers. Likewise, a padding style attribute is now exposed publicly as an NSValue representing an `NSEdgeInsets` or `UIEdgeInsets` instead of an NSArray of NSNumbers.

This change also fixes round-tripping of padding values due to a difference between the style specification and Foundation regarding the order of edges around a box. It uses a designated initializer on `NSEdgeInsets`/`UIEdgeInsets` to ensure correct order when converting from C++ to Objective-C.

Fixes #5947, fixes #6065. Depends on #6076.

/cc @frederoni

PR #6168, opened 2016-08-26
Make style properties null-resettable; fix property function roundtripping

### LINKED ISSUE #6126: MGLStyleLayer properties should be null_resettable or nonnull
As of #6076, all the layout and paint properties on MGLStyleLayer subclasses are marked as `nullable`, but `nil` has an implicit meaning that we describe in documentation comments. Relying on documentation to assign a meaning to `nil` is fragile. Now that #6124 has landed, MGLStyleLayer properties should be either `null_resettable` or `nonnull`. `null_resettable` is pretty rare among Cocoa APIs, but `nullable` with an implicit meaning assigned to `nil` is even rarer.

On the plus side, the default behavior with `null_resettable` is easier to document and intuit, and we could even potentially change many NSNumber- and NSValue-typed properties back to primitive integers and enumerations, depending on how #5970 shakes out. (If we do go back to primitive types, we’ll need to use `assign` semantics, which precludes `null_resettable` behavior.)

On the other hand, the SDK will be unable to distinguish between an unset property and one that happens to be the default. Any attempted serialization from MGLStyleLayers to style JSON would be lossy, but it isn’t clear to me that this contrived use case would outweigh the benefits.

/cc @frederoni @incanus @friedbunny @boundsj @jfirebaugh

### LINKED ISSUE #6161: Getters for enumeration-typed style properties return nil instead of functions
The `MGLGetEnumProperty()` macro only implements constant values, [not function values](https://github.com/mapbox/mapbox-gl-native/blob/05ba97b2b7c0b15d31b42fdd14094eab015c9c90/platform/darwin/src/MGLStyleAttributeFunction_Private.h#L36), so if an enumeration-typed style property is set to a function, the getter always returns `nil`.

/cc @frederoni

### LINKED ISSUE #6162: Test setting style properties to functions
The generated unit tests for style property setters and getters only exercise constant values, not functions. We should also test roundtripping functions too. That would’ve caught #6161.

/cc @frederoni

1ec5:
This PR includes several interwoven changes, all related to the runtime styling API:

* MGLStyleLayer properties representing optional style attributes are all `null_resettable` in Objective-C and therefore implicitly unwrapped optionals in Swift. Instead of returning `nil` when unset, these properties now return the intrinsic default value. (Fixes #6126.)
* Each documentation comment spells out the type of the default value, in anticipation of a fix for #5970.
* When an enumeration-typed property can have either of two values and is set to a constant, its getter returns the right value instead of random garbage.
* When an enumeration-typed property is set to a function, its getter returns that function instead of `nil`. (Fixes #6161.)
* Fixed an issue in which setting a padding-typed property to a function would erroneously assert (due to a change omitted from #6079).
* Fixed an issue in which getting a padding-typed property would yield NSArray-typed stops (another followup to #6079).
* Added unit tests that caught the four previous issues. (Fixes #6162.)

/cc @frederoni


1ec5:
Every enumeration-typed property test is failing on the iOS build bot:

```
testSymbolLayer, (([(NSValue *)gLayer.iconTextFit isEqualToValue:[MGLRuntimeStylingHelper testEnum:MGLSymbolStyleLayerIconTextFitHeight type:@encode(MGLSymbolStyleLayerIconTextFit)]]) is true) failed
```

However, I’m unable to reproduce this failure locally.

1ec5:
The test failures turned out to be legitimate. The tests caught yet another roundtripping issue.

platform/darwin/scripts/generate-style-code.js
```diff
@@ -157,26 +136,27 @@ global.parseColor = function (str) {
...
-            return '`' + `${objCType}${camelize(value)}` + '`';
+            return 'an `NSValue` object containing `' + `${objCType}${camelize(value)}` + '`';
         case 'color':
             let color = parseColor(value);
             if (!color) {
                 throw new Error(`unrecognized color format in default value of ${property.name}`);
             }
             if (color.r === 0 && color.g === 0 && color.b === 0 && color.a === 0) {
-                return '`clearColor`';
+                return '`NSColor.clearColor` or `UIColor.clearColor`';
```
friedbunny: Nit: it seems to me that the iOS variant should come before macOS’s.

friedbunny:
Other than [my iOS partisanship](https://github.com/mapbox/mapbox-gl-native/pull/6168#discussion_r76509663), this LGTM. 👍

platform/darwin/scripts/generate-style-code.js
1ec5: Rest assured, I was being completely objective and dispassionate by alphabetizing the items in this list. 😉

1ec5: The entire declaration and its comment are now duplicated once per platform using conditional compilation macros. Each comment says only NSColor or UIColor as appropriate. Out of respect for iOS’s primacy, the UIKit version goes first.

PR #6170, opened 2016-08-26
Upgrade RapidJSON

kkaefer:
[RapidJSON 1.1.0 was released](https://github.com/miloyip/rapidjson/releases/tag/v1.1.0) and has some goodies:

* C++11 range loop syntax
* Comments(!) and trailing commas, as well as NaN/Infinity
* Reduced memory usage


jfirebaugh:
For future reference, can you include a rundown of what the clang-tidy issue was and how we worked around it here or in the commit message?

kkaefer:
clang-tidy issue is that the code path for allocating memory in RapidJSON is pretty complicated and I think clang-tidy trips over this. I followed the code path with Xcode's visual analyze feature and think that it is correct. Therefore, I added a `.clang-tidy` file that disables these checks with in the folder where RapidJSON's write features are used (`offline.cpp`)

kkaefer:
Can't test linking of libgeojson on Android due to https://github.com/mapbox/mapbox-gl-native/issues/6193. I'll refrain from merging for now.

kkaefer:
@brunoabinader @jfirebaugh please review

PR #6175, opened 2016-08-26
Check glGetError in a loop

mikemorris:
As recommended in https://www.opengl.org/wiki/GLAPI/glGetError

/cc @tmpsantos

### DIFF
--- a/include/mbgl/gl/gl.hpp
+++ b/include/mbgl/gl/gl.hpp
@@ -59,9 +59,8 @@ namespace gl {
     void mbx_trapExtension(const char *name, GLuint array);
 #endif
 
-struct Error : ::std::runtime_error {
-    Error(GLenum err, const std::string &msg) : ::std::runtime_error(msg), code(err) {};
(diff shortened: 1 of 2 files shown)


kkaefer:
Any reason for moving the error message construction to a separate function?

mikemorris:
@kkaefer Just seemed slightly cleaner that way, I can refactor it back into `checkError` quickly if you'd prefer.

mikemorris:
@kkaefer Reworked this to move error construction back into `checkError`

kkaefer:
The [previous code](https://github.com/mapbox/mapbox-gl-native/blob/f8f3f276c8dc2bd9cf032df5d0c180267e68374b/src/mbgl/gl/gl.cpp) always allocated a `std::vector` and a `std::string`, even when no error occurred (which is the vast majority of all cases). I updated the code to do the allocation lazily, and to print the errors in the order they were received, instead of backwards, and factored out the enum to string conversion.

Issue #5722, opened 2016-07-19
Memory issue on init


type-it:
Hi,
I have installed MapBox via Fabric.
When i go an trigger the init method like this

- (void)mapBox:(UIView*)mapView
{
    self.map = [MGLMapView new];
    [self.map setFrame:mapView.bounds];
    [self.map setCenterCoordinate:CLLocationCoordinate2DMake(40.712791, -73.997848)
                        zoomLevel:12
                         animated:NO];
    [mapView addSubview:self.map];
}

it doesn't load and I get a crash like this
![zrzut ekranu 2016-07-19 o 09 58 56](https://cloud.githubusercontent.com/assets/20536013/16946970/9893e5fc-4dac-11e6-8790-da3b7894e79f.png)
![zrzut ekranu 2016-07-19 o 10 00 56](https://cloud.githubusercontent.com/assets/20536013/16946971/9895019e-4dac-11e6-9ba7-9f4bf1b7aefb.png)
![zrzut ekranu 2016-07-19 o 09 59 17](https://cloud.githubusercontent.com/assets/20536013/16946973/9895a266-4dac-11e6-9278-26fa230cfcd3.png)
![zrzut ekranu 2016-07-19 o 09 59 31](https://cloud.githubusercontent.com/assets/20536013/16946972/989536d2-4dac-11e6-9dbd-b6ae7f1c17a9.png)

AppName(3849,0x16e087000) malloc: *** mach_vm_map(size=805306368) failed (error code=3)
*** error: can't allocate region
*** set a breakpoint in malloc_error_break to debug
libc++abi.dylib: terminating with uncaught exception of type std::bad_alloc: std::bad_alloc
(lldb) 

PLEASE HELP

wanggang316:
me to

Paladinfeng:
it affected me too. the memmory increase to hundreds of MB. the thread would be killed

type-it:
I Found that using
 pod "Mapbox-iOS-SDK", "~> 3.2.0"

Works correctly ;-)

Paladinfeng:
the service was resumed at 7:00pm (time zone: Beijing)

friedbunny:
We’re investigating, thanks for your patience.

1ec5:
We’re tracking the fix in #5723 and will release the fix today in v3.3.1.

HappyDr0id:
[...]

Issue #943, opened 2015-03-04
Make network interface more resilient to captive portals


kkaefer:
Some captive portals hijack DNS and use self-signed certificates to redirect the user to a log in. In this case, our requests will likely fail. We should treat certificate errors as temporary connection problems and back off exponentially to see whether the issue has been fixed.

ljbade:
It would be interesting to see how Google and Apple maps behave with bad certificates.

jfirebaugh:
We do use exponential backoff now.

PR #6176, opened 2016-08-26
[core] Corrupted resource stays in the cache until expiration

tmpsantos:
Any resource that makes to the cache is only evicted when it expires or the cache runs out of space. In case of critical resources like the style or tile source, this can potentially "brick" the map.

We should evict corrupted resources as we find them.

### DIFF
--- a/platform/default/mbgl/storage/offline_database.cpp
+++ b/platform/default/mbgl/storage/offline_database.cpp
@@ -241,7 +241,13 @@ optional<std::pair<Response, uint64_t>> OfflineDatabase::getResource(const Resou
     if (!data) {
         response.noContent = true;
     } else if (stmt->get<int>(4)) {
-        response.data = std::make_shared<std::string>(util::decompress(*data));
+        try {
+            response.data = std::make_shared<std::string>(util::decompress(*data));
+        } catch (std::exception const& ex) {
+            Log::Error(Event::Database, "Error decompressing resource: %s", ex.what());
+            removeResource(resource);
+            return {};
+        }
         size = data->length();
     } else {
         response.data = std::make_shared<std::string>(*data);
@@ -410,7 +416,13 @@ optional<std::pair<Response, uint64_t>> OfflineDatabase::getTile(const Resource:
     if (!data) {
         response.noContent = true;
     } else if (stmt->get<int>(4)) {
-        response.data = std::make_shared<std::string>(util::decompress(*data));
+        try {
+            response.data = std::make_shared<std::string>(util::decompress(*data));
+        } catch (std::exception const& ex) {
+            Log::Error(Event::Database, "Error decompressing tile: %s", ex.what());
+            removeTile(tile);
+            return {};
+        }
         size = data->length();
     } else {
         response.data = std::make_shared<std::string>(*data);
--- a/platform/default/mbgl/storage/offline_database.cpp
+++ b/platform/default/mbgl/storage/offline_database.cpp
@@ -130,6 +130,15 @@ OfflineDatabase::Statement OfflineDatabase::getStatement(const char * sql) {
     return Statement(*statements.emplace(sql, std::make_unique<mapbox::sqlite::Statement>(db->prepare(sql))).first->second);
 }
 
+void OfflineDatabase::remove(const Resource& resource) {
+    if (resource.kind == Resource::Kind::Tile) {
+        assert(resource.tileData);
+        removeTile(*resource.tileData);
+    } else {
+        removeResource(resource);
+    }
+}
+
 optional<Response> OfflineDatabase::get(const Resource& resource) {
     auto result = getInternal(resource);
     return result ? result->first : optional<Response>();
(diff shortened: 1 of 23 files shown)


tmpsantos:
Work in progress, will finish the implementation and write tests.

platform/default/mbgl/storage/offline_database.cpp
```diff
@@ -184,6 +184,15 @@ std::pair<bool, uint64_t> OfflineDatabase::putInternal(const Resource& resource,
     return { inserted, size };
 }
 
+void OfflineDatabase::remove(const Resource& resource) {
+    // clang-format off
+    Statement stmt = getStatement("DELETE FROM resources WHERE url = ?1");
```
jfirebaugh: Should handle tiles too.

tmpsantos:
:eyes: @jfirebaugh 

I'm not particularly happy with 137cdfd4fa2f6f, but did not find a better way since unlike the raster tiles, we parse vector tiles on a worker.

jfirebaugh:
I'm not convinced that this adds enough value to justify the additional complexity, especially given that the added code path is infrequently exercised and therefore has a relatively high chance of regressions that worsen a failure case than recover from it. I'm imagining something like a bug in the `reportBad` pathway that turns what would have been "map doesn't render / is missing tiles" into "app crashes".

It wouldn't have saved us in the case of #5722 -- the crash there happened before `reportBad` would have been called. And even in the case that it does result in successfully evicting a bad resource from the cache, what prevents the bad resource simply being re-downloaded from the server and re-cached, ad infinitum?

tmpsantos:
@jfirebaugh I do agree that this code won't be frequently exercised but I believe we have enough unit tests for it and we can probably add more if needed.

Currently without this patch Mapbox GL Native will crash if it fails to decompress something from the cache. This is very unlikely but nevertheless a possible scenario.

Also on the unlikely event that we get a tile source we cannot parse, that will pretty much brick the map until it expires.

The current approach is very conservative and will simply purge the item from the cache and won't cause a retry. If we are talking a about a mobile app here, it will only unbrick the map if you set the style again (probably will require you to close/open the app).

The follow-up from this will be report "resource loading failed" as a signal to the client, so it can decide to reload the map. Currently we only do that when failing to load the style.

mb12:
@tmpsantos and @jfirebaugh several private wifi networks (esp on airports and coffee shops) always return an http sign in page for every https/http request until one signs in. Since the offline download does not do any validation,  downloads initiated under such network conditions would result in this code path being used quite often.

1ec5:
> The follow-up from this will be report "resource loading failed" as a signal to the client, so it can decide to reload the map.

#1804

> several private wifi networks (esp on airports and coffee shops) always return an http sign in page for every https/http request until one signs in.

Previous discussion about captive networks: #943.

tmpsantos:
:eyes: @kkaefer

test/storage/offline_database.cpp
```diff
@@ -224,15 +224,38 @@ TEST(OfflineDatabase, PutResourceNoContent) {
...
     response.noContent = true;
 
-    db.put(resource, response);
-    auto res = db.get(resource);
-    EXPECT_EQ(nullptr, res->error);
-    EXPECT_TRUE(res->noContent);
-    EXPECT_FALSE(res->data.get());
+    Resource style { Resource::Style, "http://example.com/" };
+    db.put(style, response);
+    EXPECT_FALSE(db.get(style));
```
kkaefer: Could you please add a note about why `EXPECT_FALSE` is correct here. This is not obvious from just reading this code.

kkaefer:
Instead of tacking on a `std::function` to every Response object, we should funnel the notification that a resource is bad directly to the filesource.

I think the captive portal scenario that @mb12 describes is valid, and we should add tests for this. Instead of providing a callback that allows reporting bad data, we could also introduce very simple sanity checks directly in the `OnlineFileSource` (e.g. make sure JSON files start with a valid JSON character, images have a magic header, and PBF files are parseable. Of course, this means that we could still get invalid data delivered from our server, but as @jfirebaugh says, in the case of #5722, this wouldn't have helped us.

jfirebaugh:
> Instead of tacking on a std::function to every Response object, we should funnel the notification that a resource is bad directly to the filesource.

I'd be more amenable to this feature if we did that. [Code like this](https://github.com/mapbox/mapbox-gl-native/blob/416cdab5460cb2eee1a99483f958d044c3a11874/platform/default/default_file_source.cpp#L88-L90) makes me think "how do we know the `this` bound in the lamba is still valid when it's called"?

tmpsantos:
@jfirebaugh @kkaefer thanks for reviewing. With the new Actor model we can do much better here, gonna rewrite it.

kkaefer:
I still believe that this is a valid issue, in particular bad captive portals poisoning the cache. If we implement some way of deleting bad resources from the cache, we're also going to trigger a re-request. This means we'll have to make sure we don't enter a loop that in quick succession requests + deletes the same item over and over again. An exponential back off (like we have for failing requests) could work here.

tmpsantos:
Abandoned.

PR #6182, opened 2016-08-26
Persist querystrings in mapbox:// resource urls to API requests

jakepruitt:
Carries over any query strings from `mapbox://`-prefixed styles urls, source urls and tile urls to the `api.mapbox.com` requests.

cc/ @mapsam @1ec5 @jfirebaugh @yhahn @GretaCB

### DIFF
--- a/src/mbgl/util/mapbox.cpp
+++ b/src/mbgl/util/mapbox.cpp
@@ -11,14 +11,15 @@ namespace util {
 namespace mapbox {
 
 const std::string protocol = "mapbox://";
-    
+const std::size_t protocolLength = protocol.length();
+
 bool isMapboxURL(const std::string& url) {
     return std::equal(protocol.begin(), protocol.end(), url.begin());
 }
 
-std::vector<std::string> getMapboxURLPathname(const std::string& url) {
+static std::vector<std::string> getMapboxURLPathname(const std::string& url) {
     std::vector<std::string> pathname;
-    auto startIndex = protocol.length();
+    auto startIndex = protocolLength;
     auto end = url.find_first_of("?#");
     if (end == std::string::npos) {
         end = url.length();
@@ -34,6 +35,20 @@ std::vector<std::string> getMapboxURLPathname(const std::string& url) {
     return pathname;
 }
 
+static std::pair<std::string, std::size_t> normalizeQuery(const std::string& url) {
+    std::string query;
+
+    auto queryIdx = url.find("?");
+    if (queryIdx != std::string::npos) {
+        query = url.substr(queryIdx + 1, url.length() - queryIdx + 1);
+        if (!query.empty()) {
+            query = "&" + query;
+        }
+    }
+
+    return std::make_pair(query, queryIdx);
+}
+
 std::string normalizeSourceURL(const std::string& baseURL, const std::string& url, const std::string& accessToken) {
     if (!isMapboxURL(url)) {
         return url;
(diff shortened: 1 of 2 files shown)


src/mbgl/util/mapbox.cpp
```diff
@@ -44,7 +44,13 @@ std::string normalizeSourceURL(const std::string& url, const std::string& access
         throw std::runtime_error("You must provide a Mapbox API access token for Mapbox tile sources");
     }
 
-    return baseURL + "v4/" + url.substr(protocol.length()) + ".json?access_token=" + accessToken + "&secure";
+    auto queryIdx = url.find("?");
+    std::string query = "";
```
jfirebaugh: Omit ` = ""` -- it'll be initialized to an empty string automatically.

src/mbgl/util/mapbox.cpp
```diff
@@ -44,7 +44,13 @@ std::string normalizeSourceURL(const std::string& url, const std::string& access
...
+    if (queryIdx != std::string::npos) {
+        query = "&" + url.substr(queryIdx + 1, url.length() - queryIdx + 1);
```
jfirebaugh: What will this code do if `?` is the last character in the URL? That  "should never happen" with `mapbox://` URLs but [as we've seen](https://github.com/mapbox/mapbox-gl-native/issues/5722), sometimes the supposedly impossible happens. Please add a test for this case and adjust the code if necessary.

src/mbgl/util/mapbox.cpp
```diff
@@ -44,7 +44,16 @@ std::string normalizeSourceURL(const std::string& url, const std::string& access
         throw std::runtime_error("You must provide a Mapbox API access token for Mapbox tile sources");
     }
 
-    return baseURL + "v4/" + url.substr(protocol.length()) + ".json?access_token=" + accessToken + "&secure";
+    auto queryIdx = url.find("?");
+    std::string query;
+    if (queryIdx != std::string::npos) {
+        query = url.substr(queryIdx + 1, url.length() - queryIdx + 1);
+        if (query.length() > 0) {
```
springmeyer: Generally `if (!query.empty()) {` is better practice since it is faster than `query.size()`.

springmeyer:
Made a refactoring pass and rebased this /cc @jakepruitt. Ready for review by others, notably @jfirebaugh @1ec5

src/mbgl/util/mapbox.cpp
```diff
@@ -11,15 +11,16 @@ namespace util {
 namespace mapbox {
 
 const std::string protocol = "mapbox://";
+const std::size_t protocol_length = protocol.length();
```
jfirebaugh: Camel case please.

src/mbgl/util/mapbox.cpp
```diff
@@ -11,15 +11,16 @@ namespace util {
...
 const std::string baseURL = "https://api.mapbox.com/";
 
 bool isMapboxURL(const std::string& url) {
     return std::equal(protocol.begin(), protocol.end(), url.begin());
 }
 
-std::vector<std::string> getMapboxURLPathname(const std::string& url) {
+inline std::vector<std::string> getMapboxURLPathname(const std::string& url) {
```
jfirebaugh: In this codebase, `inline` is used only where strictly necessary (i.e. non-template standalone definitions in header files).

src/mbgl/util/mapbox.cpp
```diff
@@ -35,6 +36,18 @@ std::vector<std::string> getMapboxURLPathname(const std::string& url) {
     return pathname;
 }
 
+std::pair<std::string,std::size_t> normalizeQuery(const std::string& url) {
```
jfirebaugh: Space after comma.

src/mbgl/util/mapbox.cpp
```diff
@@ -121,7 +135,8 @@ std::string normalizeTileURL(const std::string& url, const std::string& accessTo
         return url;
     }
 
-    return baseURL + "v4/" + url.substr(sizeof("mapbox://tiles/") - 1) + "?access_token=" + accessToken;
+    auto query = normalizeQuery(url);
+    return baseURL + "v4/" + url.substr(sizeof("mapbox://tiles/") - 1, query.second - sizeof("mapbox://tiles/") + 1) + "?access_token=" + accessToken + normalizeQuery(url).first;
```
jfirebaugh: If no query is present, `query.second` is `std::string::npos`, `query.second - sizeof("mapbox://tiles/") + 1` will be a huge number, and `url.substr(...)` will throw `std::out_of_range`. Please handle and add a test case for this.

jfirebaugh: Hmm, actually it won't throw, that's only if the first argument to `substr` is greater than the size. But can you add a test case anyway?

mapsam:
Thank you @jfirebaugh!

PR #6185, opened 2016-08-27
[core] Fix updates of line and fill annotations

### LINKED ISSUE #5844: AnnotationManager::update doesn't update LineAnnotation / FillAnnotation
I have been looking into implementing updatable Polyline/Polygons on the Android bindings.
For this I'm hooking into the following methods exposed in AnnotationManager:

```cpp
    Update update(const AnnotationID&, const LineAnnotation&, const uint8_t);
    Update update(const AnnotationID&, const FillAnnotation&, const uint8_t);
```

When I execute the update method, I'm  seeing the polygon being removed but not readded:

ezgif com-video-to-gif 23

The code used here is equivalent to the one used for SymbolAnnotation.

jfirebaugh:
Fixes #5844.

There were two issues:

* The "remove-and-readd" logic was neglecting to account for `obsoleteShapeAnnotationLayers`, causing updates to remove the style layer.
* `{Fill,Line}AnnotationImpl::updateStyle` did not update the properties of previously-existing style layer.

cc @tobrun


tobrun:
Tested and works great. Removing current workaround in https://github.com/mapbox/mapbox-gl-native/pull/6187.

Issue #6092, opened 2016-08-19
Unrecognized selector crash when suspending and removing MGLOfflinePack


Jackhammer42:
Platform: iOS 9.3.4
Mapbox SDK version: 3.3.1
### Steps to trigger behavior

When I try to cancel the download of an MGLOfflinePack, i get this error:

```
-[NSTaggedPointerString offlineRegionStatusDidChange:]: unrecognized selector sent to instance 0xa0000000148344a4
2016-08-19 16:53:08.187 ChemPark[659:57661] *** Terminating app due to uncaught exception 'NSInvalidArgumentException', reason: '-[NSTaggedPointerString offlineRegionStatusDidChange:]: unrecognized selector sent to instance 0xa0000000148344a4'
*** First throw call stack:
(0x181aa6db0 0x18110bf80 0x181aadc4c 0x181aaabec 0x1819a8c5c 0x100f484e4 0x102ba9a7c 0x102ba9a3c 0x102baf4e4 0x181a5cd50 0x181a5abb8 0x181984c50 0x18326c088 0x186c6e088 0x1002b1c70 0x1815228b8)
libc++abi.dylib: terminating with uncaught exception of type NSException
```

I am using this code for the task:

```
let packs = MGLOfflineStorage.sharedOfflineStorage().packs

        for pack in packs! {

            pack.suspend()

            MGLOfflineStorage.sharedOfflineStorage().removePack(pack, withCompletionHandler: nil)

        }
```

It seems similar to https://github.com/mapbox/mapbox-gl-native/issues/4394, except this issue only seems to happen if i cancel the download, not after successfully completing the download.
### Expected behavior

I was hoping to stop the download.
### Actual behavior

Crash

1ec5:
Related to #5538.

1ec5:
Does the crash occur if you only remove the pack instead of canceling it? Removing the pack should cancel it automatically, if asynchronously.

Jackhammer42:
I tried 

```
let packs = MGLOfflineStorage.sharedOfflineStorage().packs

        for pack in packs! {


            MGLOfflineStorage.sharedOfflineStorage().removePack(pack, withCompletionHandler: nil)

        }
```

Same crash, after a few starting and canceling cycles.


When I trie

```
let packs = MGLOfflineStorage.sharedOfflineStorage().packs
        
        for pack in packs! {
            
            
            let priority = DISPATCH_QUEUE_PRIORITY_DEFAULT
[...]

PR #6186, opened 2016-08-27
Ignore progress updates after suspending

### LINKED ISSUE #5538: Stop delivering status updates when offline download is canceled

1ec5:
This is a cosmetic fix that makes an offline pack appear to stop downloading as soon as `-suspend` is called on it.

Note that this change does not address the crash in #6092, which occurs because the `MBGLOfflineRegionObserver`’s weak pointer to the MGLOfflinePack object has gone stale by the time late notifications come in from the `mbgl::OfflineRegion`. `MBGLOfflineRegionObserver` is owned by `mbgl::OfflineRegion` and MGLOfflinePack doesn’t currently have a reference to it, although that might be one future direction for improvement.

Fixes #5538.

/cc @jfirebaugh

### DIFF
--- a/platform/darwin/src/MGLOfflinePack.mm
+++ b/platform/darwin/src/MGLOfflinePack.mm
@@ -43,7 +43,9 @@ private:
 
 @end
 
-@implementation MGLOfflinePack
+@implementation MGLOfflinePack {
+    BOOL _isSuspending;
+}
 
 - (instancetype)init {
     if (self = [super init]) {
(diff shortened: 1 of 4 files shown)


jfirebaugh:
I'll delegate review here to someone more familiar with `platform/darwin` code.

1ec5:
@boundsj @incanus, does this look safe and effective to you?

platform/darwin/src/MGLOfflinePack.mm
```diff
@@ -93,6 +95,11 @@ - (void)resume {
 - (void)suspend {
     MGLAssertOfflinePackIsValid();
     
+    if (self.state == MGLOfflinePackStateActive) {
```
boundsj: Does `resume` need to set _isSuspending to `NO`? I'm not certain but it seems like a resumed offline pack may no longer report progress even if the state is set to `MGLOfflinePackStateActive` because of a region status change.

1ec5: Fixed in 3b546b964609d0f596dac32e155b1489bb85645e.

PR #6207, opened 2016-08-30
Move class handling to Style

### LINKED ISSUE #6118: Move class state back to Map
01edd4503c5b78a73ac8ce1ee1de8b403d76d4fc moved the state holding the list of active classes to `Style`. In gl-js, we had had previously (https://github.com/mapbox/mapbox-gl-js/issues/911, https://github.com/mapbox/mapbox-gl-js/pull/916) decided that `Map` was the correct holder of this state, because it is desirable for it to persist across style loads or reloads. This is particularly true on native, where reloads of a style can happen implicitly during revalidation. In that case, having the class list state reset, as currently happens, is unexpected.

If we make this change, we'll need to revert #5974.

cc @brunoabinader @1ec5 @frederoni

brunoabinader:
Fixes #6118.

- [x] Map::Impl now owns class data.
- [x] This prevents class state being reset upon `mbgl::Style` state changes.
- [x] Adds new `mbgl::setClassTransition` to cleanup class API.
 - Updated GLFW and darwin API usage accordingly
- [x] Update Qt QML: `classes` property now belongs to `QQuickMapboxGL` (map) instead of `QQuickMapboxGLStyle` (style)

### DIFF
--- a/test/map/map.cpp
+++ b/test/map/map.cpp
@@ -228,3 +228,37 @@ TEST(Map, RemoveLayer) {
 
     test::checkImage("test/fixtures/map/remove_layer", test::render(map));
 }
+
+TEST(Map, Classes) {
+    MapTest test;
+
+    Map map(test.view, test.fileSource, MapMode::Still);
+    map.setStyleJSON(util::read_file("test/fixtures/api/empty.json"));
+
+    EXPECT_FALSE(map.getTransitionOptions().duration);
+
+    auto duration = mbgl::Duration(mbgl::Milliseconds(300));
+    map.setTransitionOptions({ duration });
+    EXPECT_EQ(map.getTransitionOptions().duration, duration);
+
+    map.addClass("test");
+    EXPECT_TRUE(map.hasClass("test"));
+
+    map.removeClass("test");
+    EXPECT_TRUE(map.getClasses().empty());
+
+    std::vector<std::string> classes = { "foo", "bar" };
+    map.setClasses(classes);
+    EXPECT_FALSE(map.hasClass("test"));
+    EXPECT_TRUE(map.hasClass("foo"));
+    EXPECT_TRUE(map.hasClass("bar"));
+
+    // Does nothing - same style JSON.
+    map.setStyleJSON(util::read_file("test/fixtures/api/empty.json"));
+    EXPECT_TRUE(map.hasClass("foo"));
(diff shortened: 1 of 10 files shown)


brunoabinader:
Tagging @tmpsantos for reviewing the Qt changes 👍

jfirebaugh:
@brunoabinader Sorry for the flip-flop, but after reviewing how style batch updates happen in GL JS and will need to happen in native for #5701, I think you had the right idea moving classes to `Style`, and we should stick with that (and change JS to match -- https://github.com/mapbox/mapbox-gl-js/issues/2385). I'm prepping a PR for JS that demonstrates the benefit pretty clearly.

👍  on `setClassTransition` though, the existing transition argument behaves strangely (only most recent value is respected).

tmpsantos:
Reviewed offline.

brunoabinader:
@jfirebaugh rewrote the PR, keeping class handling in Style and applying the transition options changes. @tmpsantos we can forward the QML style property behavior changes to the upcoming QML style layer PR.

test/map/map.cpp
```diff
@@ -228,3 +228,39 @@ TEST(Map, RemoveLayer) {
...
+    MapTest test;
+
+    Map map(test.view, test.fileSource, MapMode::Still);
+    map.setStyleJSON(util::read_file("test/fixtures/api/empty.json"));
+
+    EXPECT_EQ(map.getTransitionOptions().duration->count(), 0);
+
+    auto duration = mbgl::Milliseconds(300);
+    map.setTransitionOptions({ { duration } });
+    EXPECT_EQ(map.getTransitionOptions().duration->count(),
```
jfirebaugh: Can you just do `EXPECT_EQ(map.getTransitionOptions().duration, duration)` here? Ditto line 260.

jfirebaugh:
LGTM, modulo the nit in the test code and the CI failures.

test/map/map.cpp
brunoabinader: Yes and no - with the current implementation I get `invalid operands to binary expression ('const std::experimental::fundamentals_v1::optional<std::__1::chrono::duration<long long, std::__1::ratio<1, 1000000000> > >' and 'const std::__1::chrono::duration<long long, std::__1::ratio<1, 1000> >')` - but if I encapsulate `duration` inside `mbgl::Duration` then it works fine 👍 Thanks!

Issue #6115, opened 2016-08-22
Add support for pole-of-inaccessibility symbol placement


lucaswoj:
GL JS now supports pole-of-inaccessibility symbol placement for polygon geometries. We need to port this feature over to GL Native. 
- https://github.com/mapbox/mapbox-gl-js/issues/1231
- https://github.com/mapbox/mapbox-gl-js/pull/2678
- https://github.com/mapbox/mapbox-gl-test-suite/pull/111

cc @blanchg @jfirebaugh @mourner

1ec5:
The pole of inaccessibility algorithm will also come in handy for callout placement on selected shape annotations, as envisioned for #2082.

1ec5:
The first step is to port [Polylabel](https://github.com/mapbox/polylabel/) to C++ for use in this project.

PR #6211, opened 2016-08-30
Remove misleading MGLOverlay documentation

friedbunny:
This removes misleading information from MGLOverlay’s documentation ~~and adds specific docs about the current behavior of the `coordinate` property of MGLPolygon and MGLPolyline~~.

MGLMultiPoint [currently provides the implementation](https://github.com/mapbox/mapbox-gl-native/blob/22cc375da0e2f950ae0de805b5e2b9ab01c15708/platform/darwin/src/MGLMultiPoint.mm#L61) of `coordinate`, which simply returns the first coordinate in its array of coordinates. Documenting this behavior in MGLMultiPoint doesn’t override MGLOverlay when using Xcode’s Quick Help.

This is a stopgap, until we fix this behavior and return the visual center of overlays in https://github.com/mapbox/mapbox-gl-native/issues/6115.

/cc @1ec5

### DIFF
--- a/platform/darwin/src/MGLOverlay.h
+++ b/platform/darwin/src/MGLOverlay.h
@@ -29,7 +29,7 @@ NS_ASSUME_NONNULL_BEGIN
 @protocol MGLOverlay <MGLAnnotation>
 
 /**
- The approximate center point of the overlay area. (required) (read-only)
+ A coordinate representing the overlay. (required) (read-only)
  */
 @property (nonatomic, readonly) CLLocationCoordinate2D coordinate;
 
--- a/platform/darwin/src/MGLOverlay.h
+++ b/platform/darwin/src/MGLOverlay.h
@@ -30,10 +30,6 @@ NS_ASSUME_NONNULL_BEGIN
 
 /**
  The approximate center point of the overlay area. (required) (read-only)
- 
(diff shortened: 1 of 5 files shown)


platform/darwin/src/MGLOverlay.h
```diff
@@ -30,10 +30,6 @@ NS_ASSUME_NONNULL_BEGIN
 
 /**
  The approximate center point of the overlay area. (required) (read-only)
```
1ec5: Isn’t this misleading too? For now, just say that it’s a coordinate representing the overlay.

platform/darwin/src/MGLPolygon.h
```diff
@@ -17,6 +17,14 @@ NS_ASSUME_NONNULL_BEGIN
 @interface MGLPolygon : MGLMultiPoint <MGLOverlay>
 
 /**
+ The first coordinate of the polygon.
```
1ec5: We essentially consider this behavior to be a bug (unimplemented functionality). Instead of documenting it as a feature, we should speak about this API in general terms https://github.com/mapbox/mapbox-gl-native/pull/6211/files#r76847571 and avoid talking about future plans in the source code.

1ec5:
👍

PR #6213, opened 2016-08-30
Update INSTALL doc for -ObjC; use local podspec path

### LINKED ISSUE #6195: Static framework requires -ObjC linker flag
**Platform:** iOS
**Mapbox SDK version:** v3.3.4 through v3.4.0-alpha.2

Attempting to use the static framework results in a blank view and `Unknown class MGLMapView in Interface Builder file.`. Adding the `-ObjC` linker flag in Build Settings fixes this.

We’ve flipped back and forth between requiring this flag or not. [INSTALL.md](https://github.com/mapbox/mapbox-gl-native/blob/8deb7aaa79ccb96812d94d3d49e13bf49fe9308f/platform/ios/INSTALL.md) and [pod-README.md](https://github.com/mapbox/mapbox-gl-native/blob/8deb7aaa79ccb96812d94d3d49e13bf49fe9308f/platform/ios/docs/pod-README.md) don’t currently mention it, but they evidently should.

friedbunny:
- Adds -ObjC flag for static lib.
- Fixes link to iOS SDK site.
- Pushes CocoaPods down the list.
- Changes CocoaPods to use local path, not HTTP.
- Emphasizes that INSTALL.md is only for custom builds.

Fixes #6195.

/cc @1ec5

### DIFF
--- a/platform/ios/INSTALL.md
+++ b/platform/ios/INSTALL.md
@@ -1,6 +1,6 @@
-# Integrating the Mapbox iOS SDK into your application
+# Integrating custom builds of the Mapbox iOS SDK into your application
 
-This document explains how to build a development version of Mapbox iOS SDK for use in your own Cocoa Touch application. To use a production-ready version of the SDK, see the [Mapbox iOS SDK homepage](https://github.com/mapbox/ios-sdk).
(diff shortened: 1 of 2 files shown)


frederoni:
Can we add
```
m.xcconfig = { 'OTHER_LDFLAGS' => '-ObjC' }
```
to `Mapbox-iOS-SDK-static-part.podspec` ?

1ec5:
Good point – tracking in #6195.

PR #6214, opened 2016-08-30
[android] introduce checkstyle

cammace:
added new checkstyle rules and gradle task. Using Googles checkstyle rules with javadoc checks commented out (can be reintroduced later) and more rules adjusted to match [squares](https://github.com/square/retrofit/blob/master/checkstyle.xml). Next steps to merge this, and add step to Bitrise to run task similar to what's been done in the demo app. Right now I have everything set to warning only so builds shouldn't fail. Plan on changing to error though once all checkstyle issues in repo are resolved making builds fail (enforcing code style)

cc: @mapbox/android


tobrun:
This is great @cammace, note that adding a step to bitrise can be made part of this PR by adding it to the `bitrise.yml`. LGTM :shipit:

zugaldia:
> note that adding a step to bitrise can be made part of this PR by adding it to the bitrise.yml

In favor of this approach too, otherwise 👍 .

PR #6215, opened 2016-08-30
Allow implicit animation of annotation views

### LINKED ISSUE #6203: User dot doesn’t animate between locations
#1582 has regressed. While user tracking mode is off, the user location annotation view no longer animates between locations as location updates come in.

This issue reproduces in v3.4.0-alpha.3 but not v3.4.0-alpha.2.

/cc @friedbunny @incanus

friedbunny:
We were disabling the implicit animation of `position` changes for annotation views (bb2f627 in #5550), but this meant that UIView animations were also being canceled. Once the user location annotation started inheriting from MGLAnnotationView, this meant that non-tracking changes to the user dot became unanimated.

This change shouldn’t affect the sync between map view and annotation views, as we already disable implicit animations when we update their position in [-[MGLMapView updateAnnotationViews]](https://github.com/mapbox/mapbox-gl-native/blob/c075386a984cd5de21f6201341d360a5c1039b67/platform/ios/src/MGLMapView.mm#L4548).

This also removes redundant implicit animation disabling in MGLUserLocationAnnotationView.

@incanus, I could definitely use some help confirming that this doesn’t actually regress any of our annotation view sync/performance improvements. 😁

Fixes #6203.

/cc @1ec5

### DIFF
--- a/platform/ios/src/MGLAnnotationView.mm
+++ b/platform/ios/src/MGLAnnotationView.mm
@@ -246,16 +246,6 @@
     return otherGestureRecognizer == _longPressRecognizer || otherGestureRecognizer == _panGestureRecognizer;
 }
 
-- (id<CAAction>)actionForLayer:(CALayer *)layer forKey:(NSString *)event
-{
-    // Allow mbgl to drive animation of this view’s bounds.
-    if ([event isEqualToString:@"bounds"] || [event isEqualToString:@"position"])
-    {
-        return [NSNull null];
-    }
-    return [super actionForLayer:layer forKey:event];
-}
-
 #pragma mark UIAccessibility methods
 
 - (BOOL)isAccessibilityElement {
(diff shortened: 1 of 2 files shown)


incanus:
> I could definitely use some help confirming that this doesn’t actually regress any of our annotation view sync/performance improvements.

We really haven't made any recently, and based on my testing, we are 👌 here.

incanus:
Shouldn't we also just get rid of the `bounds` disable, too?

1ec5:
Yeah, I don’t think we need the override at all.

friedbunny:
Removed the `bounds` override — we’re ✈️ by the seat of our 👖 now.

friedbunny:
After passing yesterday several times, the rebase is failing because of unrelated [Android](https://www.bitrise.io/build/d0b4920a846e202c) and [Qt](https://travis-ci.org/mapbox/mapbox-gl-native/jobs/156589001) issues.

friedbunny:
Android build failure was fixed by https://github.com/mapbox/mapbox-gl-native/pull/6225, while Qt remains open in https://github.com/mapbox/mapbox-gl-native/issues/6226.

PR #6216, opened 2016-08-30
Use contentInset to set center coordinate for contentInset updates

### LINKED ISSUE #3778: [iOS] Wrong map insets behavior while tracking user location

boundsj:
Fixes #3778 
Subsumes #4067 

This pulls just https://github.com/mapbox/mapbox-gl-native/pull/4067/commits/ccbdeaa451c7ec125e5b7b7af0f4eaa32ac8cce8 from https://github.com/mapbox/mapbox-gl-native/pull/4067 to fix the issue where the map did not correctly update to reflect the centerOffset when the user location was being tracked.


boundsj:
I'm not sure why this qt test failed:

> [ RUN      ] Memory.Fragmentation
> ../../../test/util/memory.cpp:168: Failure

merging.

1ec5:
This caused a regression that reproduces when the user dot is top- or bottom-aligned: #6257.

PR #6217, opened 2016-08-31
[ios, macos] fixes #5962 added geojson options to support clustering

### LINKED ISSUE #5962: Clustering support in MGLGeoJSONSource
MGLGeoJSONSource needs additional parameters or properties to support the point clustering functionality added in #5724 for #320. (The Android equivalent of this issue is #5852.)

/cc @frederoni

frederoni:
wip #5962

~~- [ ] Reuse default values from core~~
- [x] Write documentation
- [x] Add unit tests

~~- [ ] Documentation for mobile environment~~ (will happen in https://github.com/mapbox/mapbox-gl-native/issues/6236)

👀 @boundsj


1ec5:
Thinking ahead to when we document these options: `buffer` and `clusterRadius` will unfortunately be difficult APIs for developers to reason about. They’re expressed in units of “⅟₅₁₂ of a tile”, which don’t correspond to any fixed real-world or screen distances, yet we don’t offer any way for the developer to find out the size of the tile at a given zoom level.

Fortunately, I don’t think this needs to be one of those “guess-and-check” APIs that we try to avoid exposing in the SDK. Rather than the formulation used in the style specification documentation, I think we should talk about the buffer and cluster radius being expressed in screen points (which is true at integral zoom levels), with a caveat that they may be scaled up to 2× between integral zoom levels. (The offset threshold implemented in #1843 is only for raster and video sources, so it won’t affect clustering.)

@mourner, a couple questions for you:

1. The style specification talks about “⅟₅₁₂ of a tile” units. I assume that means a tile is 512 of these units tall and 512 of these units wide, not that there are 512 to a tile, correct? If so, that should be an easy documentation fix.
1. The style specification describes `tolerance` as a “Douglas-Peucker simplification tolerance”, but in this case that means it’s also measured in ⅟₅₁₂-of-a-tile units, right? Most descriptions of the Douglas–Peucker algorithm discuss curve simplification specifically, so it’ll be unclear to the developer how the tolerance relates to clustering. We can probably clarify the documentation a bit by talking about a minimum distance between points.

platform/darwin/src/MGLGeoJSONOptions.h
```diff
@@ -0,0 +1,13 @@
+#import <Foundation/Foundation.h>
+
+@interface MGLGeoJSONOptions : NSObject
```
1ec5: It’s ironic for me to say this in light of #5970, but I think we could get away with a simple NSDictionary and some global string constants instead of a full-blown options object. That’s the way most Cocoa APIs accept options, especially when not all of the options are required (for example, NSAttributedString). Moving to a loosely-typed dictionary also frees us from having to [hard-code default values](https://github.com/mapbox/mapbox-gl-native/pull/6217/files#diff-cbb5293af6f3a871ebb2bf9565fe13fdR10).

platform/darwin/src/MGLGeoJSONOptions.h
```diff
@@ -0,0 +1,13 @@
...
+
+@property (nonatomic, assign) NSInteger maximumZoom;
```
1ec5: Thanks for spelling out “maximum”. For consistency with `-[MGLMapView maximumZoomLevel]`, I think this should be called `maximumZoomLevel` too.

boundsj:
I actually already started this in another branch. I noticed that when I try to combine similar changes in this commit with the changes in the 6049 branch and filter on `point_count` (like this [example from the Android demo app](https://github.com/mapbox/mapbox-android-demo/blob/d5b8dd962e8ff8fa3afdf334b340baf223043eb8/MapboxAndroidDemo/src/main/java/com/mapbox/mapboxandroiddemo/examples/styles/CreateHeatmapPointsActivity.java#L121-L125)) no data is rendered. I'm going to investigate that more now

boundsj:
https://github.com/mapbox/mapbox-gl-native/pull/6217/commits/249514389a77cdb15eee779a332876eea06cdcfa implements this with an NSDictionary as suggested above.

Regarding https://github.com/mapbox/mapbox-gl-native/pull/6217#issuecomment-243697979 and the discussion in https://github.com/mapbox/mapbox-gl-js/issues/2985; this PR has minimum viable documentation and I'd like to land it sooner for an alpha release. I'll create another ticket to capture the work to make the documentation better.

cc @frederoni @1ec5

boundsj:
Tail work ticket for documentation opened: https://github.com/mapbox/mapbox-gl-native/issues/6236

platform/darwin/src/MGLGeoJSONSource.h
```diff
@@ -6,6 +6,49 @@ NS_ASSUME_NONNULL_BEGIN
 
 @protocol MGLFeature;
 
+/**
+ The value of this option is an `NSNumber` object containing a boolean. If the data
+ is a collection of point features, setting this to true clusters the points by 
+ radius into groups. The default value is false.
```
1ec5: Nit: our convention is currently to use Objective-C Boolean values, so `NO` (in backticks) rather than `false`.

platform/darwin/src/MGLGeoJSONSource.h
```diff
@@ -6,6 +6,49 @@ NS_ASSUME_NONNULL_BEGIN
...
+ */
+extern NSString * const MGLGeoJSONClusterOptionName;
+
+/**
+ The value of this option is an `NSNumber` object containing an integer. It controls
+ the radius of each cluster when clustering points, measured in 1/512ths of a tile.
```
1ec5: Nit: `<sup>1</sup>/<sub>512</sub>ths`.

platform/darwin/src/MGLGeoJSONSource.h
```diff
@@ -6,6 +6,49 @@ NS_ASSUME_NONNULL_BEGIN
...
+ The default value is 50.
+ */
+extern NSString * const MGLGeoJSONClusterRadiusOptionName;
+
+/**
+ The value of this option is an `NSNumber` object containing an integer. It controls 
+ the maximum zoom to cluster points on. Defaults to one zoom less than the value
```
1ec5: Nit: “zoom” → “zoom level”.

platform/darwin/src/MGLGeoJSONSource.h
```diff
@@ -6,6 +6,49 @@ NS_ASSUME_NONNULL_BEGIN
...
+ of `MGLGeoJSONMaxZoomLevelOptionName` (so that last zoom features are not clustered).
+ */
+extern NSString * const MGLGeoJSONClusterMaximumZoomLevelOptionName;
+
+/**
+ The value of this option is an `NSNumber` object containing an integer. It controls
+ the maximum zoom level at which to create vector tiles (higher means greater detail 
+ at high zoom levels). The default value is 18.
+ */
+extern NSString * const MGLGeoJSONMaxZoomLevelOptionName;
```
1ec5: “Max” → “Maximum”.

platform/darwin/src/MGLGeoJSONSource.h
```diff
@@ -47,6 +90,15 @@ NS_ASSUME_NONNULL_BEGIN
 - (instancetype)initWithSourceIdentifier:(NSString *)sourceIdentifier geoJSONData:(NSData *)data NS_DESIGNATED_INITIALIZER;
 
 /**
+ Initializes a source with the given identifier, GeoJSON data, and a dictionary of options for the source.
+ 
+ @param sourceIdentifier A string that uniquely identifies the source.
+ @param geoJSONData An NSData object representing GeoJSON source code.
+ @param options An NSDictionary attributes for this source specified by the <a href="https://www.mapbox.com/mapbox-gl-style-spec/#sources-geojson">the style specification</a>.
```
1ec5: “An NSDictionary attributes” → “An `NSDictionary` of attributes”.

platform/darwin/src/MGLGeoJSONSource.mm
```diff
@@ -7,25 +7,113 @@
...
+{
+    if (! [options objectForKey:key]) return;
+   
+    id keyValue = [options objectForKey:key];
+    BOOL isKindOfNumber = [keyValue isKindOfClass:[NSNumber class]];
+    
+    // if we are dealing with an NSNumber then determine the selector to extract
+    // its value based on its type and then call that selector, assigning the value
+    // to the passed in mbgl options value
+    if (isKindOfNumber)
```
1ec5: Presumably we’ll replace this with `ValueEvaluator` once #6049 lands.

boundsj: Maybe I'm just too tired but it looks like `ValueEvaluator` (renamed from `PropertyValueEvaluator`) [does the opposite](https://github.com/mapbox/mapbox-gl-native/pull/6049/files#diff-9661184742108500b609e44c1341e16dR6), transforming C++ -> to Foundation. `setOptionForKeyWithValue` in this PR converts from Foundation to C++ just like [NSExpression+ MGLAdditions'  mgl_convertedValueWithValue](https://github.com/mapbox/mapbox-gl-native/pull/6049/files#diff-698c604c23220e922e9df03699e408abR24) that is also in #6049.

 [You mentioned something along these lines few weeks ago](https://github.com/mapbox/mapbox-gl-native/pull/6049#issuecomment-240527548). Maybe you mean we should implement the reverse logic using the visitor pattern so we have a way to go from Foundation to C++. We could use such an `Evaluator` here in that case. In any case, we should definitely dry up the duplication in this PR and #6049 when they both land.

Relatedly https://github.com/mapbox/mapbox-gl-native/pull/6217/commits/4213ab9b1b6f6ab01e53cf5ec995fce1ee267faa removes some of the dynamic weirdness in this PR so the two implementations will be easier to reduce however we decide to do it.

1ec5: Sorry, you’re right, I was thinking of the opposite direction.

Looking at this again, I don’t think this method should even exist. For example, we document that `MGLGeoJSONToleranceOptionName` is an NSNumber containing a double. So we should just set `options.tolerance` to `[options[@"MGLGeoJSONToleranceOptionName"] doubleValue]` and possibly assert if it isn’t an NSNumber. That way, if the developer passes in an NSNumber containing an integer, that value is automatically converted to a double.

boundsj: > As it is, if the developer passes in a string as the tolerance, nothing will happen;

I tested NSAttributedString and it works this way. But I agree with you and we can do better 😄 . I will update this to work as you described.

1ec5: True, if you pass in a bogus value for a character attribute in `-[NSAttributedString initWithString:attributes:]`, nothing will happen. I was thinking of a more traditional option dictionary, such as the second parameter in `-[NSAttributedString initWithHTML:options:documentAttributes]`. In those cases, you will get an uncaught exception as the initializer attempts to send a message to an object of the wrong type.

boundsj:
Great points @1ec5. Thanks for the help. I've simplified this and will land it later today if it looks ok.

incanus:
> we don’t offer any way for the developer to find out the size of the tile at a given zoom level.

AFAICT, isn't this just `FractionalZoom * 512`?

1ec5:
> > we don’t offer any way for the developer to find out the size of the tile at a given zoom level.

> AFAICT, isn't this just `FractionalZoom * 512`? 

Not quite: zoom levels are logarithmic, not linear. But we don’t have an API for it anyways, so I don’t think we should hand-wave about it in documentation.

incanus:
> zoom levels are logarithmic, not linear

Ah, right. But then: `log2(FractionalZoom) * 512`?

PR #6222, opened 2016-08-31
[android][build] Only strip the resulting libraries

### LINKED ISSUE #6221: Use strip on the final libraries only so we can use ndk-stack again
When changing the build to CMake I added stripping in the wrong place ([here](https://github.com/mapbox/mapbox-gl-native/blob/5ac95b9dcfc6e71f81f4c940df422b35f25a4d0b/platform/android/scripts/toolchain.sh#L20-L20)). This results in all the object files being stripped which prevents us in using `ndk-stack` to diagnose crashes easily.

Ideally we'll only strip the resulting libraries in the [copy target](https://github.com/mapbox/mapbox-gl-native/blob/5ac95b9dcfc6e71f81f4c940df422b35f25a4d0b/platform/android/config.cmake#L157-L157). We can get the `strip` command from mason with:
```echo "set(STRIP_COMMAND \"`which $(mason env STRIP)`\")"``` (thanks @kkaefer)

ivovandongen:
Fixes #6221

cc @kkaefer


zugaldia:
@ivovandongen 👍

ivovandongen:
cc @mapbox/android

PR #6223, opened 2016-08-31
[core] rate limit handling on online file source

### LINKED ISSUE #5821: Offline region download hits Mapbox API rate limits for sprites
**Platform:** Android
**Mapbox SDK version:** 4.1.0

### Steps to trigger behavior

 1. Create a new offline region of any size using a style that relies on Mapbox-hosted sprites and tilesets.
 2. Start downloading it.
 3. Watch for error message `Failed to load sprite: HTTP status code 429`

### Expected behavior
SDK does not trigger API rate limits.

### Actual behavior
Sometimes, but not always, you see the 429 response, which must mean that the SDK is kicking off requests faster than the API allows.

### Question
We are killing the offline download whenever an error occurs. For errors like this, does the SDK retry downloading the asset after reporting the 429 response, or is killing on the first error the correct way to handle them?

ivovandongen:
Fixes #5821

### DIFF
--- a/include/mbgl/storage/response.hpp
+++ b/include/mbgl/storage/response.hpp
@@ -2,6 +2,7 @@
 
 #include <mbgl/util/chrono.hpp>
 #include <mbgl/util/optional.hpp>
+#include <mbgl/util/variant.hpp>
 
 #include <string>
 #include <memory>
@@ -45,15 +46,18 @@ public:
         NotFound = 2,
         Server = 3,
         Connection = 4,
+        RateLimit = 5,
         Other = 6,
     } reason = Reason::Other;
 
     // An error message from the request handler, e.g. a server message or a system message
     // informing the user about the reason for the failure.
     std::string message;
+    
+    optional<Timestamp> retryAfter;
 
 public:
-    Error(Reason, std::string = "");
+    Error(Reason, std::string = "", optional<Timestamp> = {});
 };
 
 std::ostream& operator<<(std::ostream&, Response::Error::Reason);
--- a/platform/default/http_file_source.cpp
+++ b/platform/default/http_file_source.cpp
@@ -4,6 +4,7 @@
 #include <mbgl/platform/log.hpp>
 
 #include <mbgl/util/util.hpp>
+#include <mbgl/util/optional.hpp>
 #include <mbgl/util/run_loop.hpp>
 #include <mbgl/util/string.hpp>
 #include <mbgl/util/timer.hpp>
@@ -79,6 +80,9 @@ private:
     // Will store the current response.
     std::shared_ptr<std::string> data;
     std::unique_ptr<Response> response;
+
+    optional<std::string> retryAfter;
+    optional<std::string> xRateLimitReset;
 
     CURL *handle = nullptr;
     curl_slist *headers = nullptr;
@@ -325,6 +329,10 @@ size_t HTTPRequest::headerCallback(char *const buffer, const size_t size, const 
     } else if ((begin = headerMatches("expires: ", buffer, length)) != std::string::npos) {
         const std::string value { buffer + begin, length - begin - 2 }; // remove \r\n
         baton->response->expires = Timestamp{ Seconds(curl_getdate(value.c_str(), nullptr)) };
+    } else if ((begin = headerMatches("retry-after: ", buffer, length)) != std::string::npos) {
+        baton->retryAfter = std::string(buffer + begin, length - begin - 2); // remove \r\n
+    } else if ((begin = headerMatches("x-rate-limit-reset: ", buffer, length)) != std::string::npos) {
+        baton->xRateLimitReset = std::string(buffer + begin, length - begin - 2); // remove \r\n
     }
 
     return length;
@@ -372,6 +380,10 @@ void HTTPRequest::handleResult(CURLcode code) {
         } else if (responseCode == 404) {
             response->error =
                 std::make_unique<Error>(Error::Reason::NotFound, "HTTP status code 404");
+        } else if (responseCode == 429) {
+            response->error =
+                std::make_unique<Error>(Error::Reason::RateLimit, "HTTP status code 429",
+                                        http::parseRetryHeaders(retryAfter, xRateLimitReset));
         } else if (responseCode >= 500 && responseCode < 600) {
             response->error =
                 std::make_unique<Error>(Error::Reason::Server, std::string{ "HTTP status code " } +
--- a/platform/qt/src/http_request.cpp
+++ b/platform/qt/src/http_request.cpp
@@ -3,6 +3,7 @@
 
 #include <mbgl/storage/response.hpp>
 #include <mbgl/util/chrono.hpp>
+#include <mbgl/util/optional.hpp>
 #include <mbgl/util/http_header.hpp>
 #include <mbgl/util/string.hpp>
 
@@ -68,6 +69,8 @@ void HTTPRequest::handleNetworkReply(QNetworkReply *reply)
     }
 
     QPair<QByteArray, QByteArray> line;
+    optional<std::string> retryAfter;
+    optional<std::string> xRateLimitReset;
     foreach(line, reply->rawHeaderPairs()) {
         QString header = QString(line.first).toLower();
 
@@ -79,6 +82,10 @@ void HTTPRequest::handleNetworkReply(QNetworkReply *reply)
             response.expires = http::CacheControl::parse(line.second.constData()).toTimePoint();
         } else if (header == "expires") {
             response.expires = util::parseTimestamp(line.second.constData());
+        } else if (header == "retry-after") {
+            retryAfter = std::string(line.second.constData(), line.second.size());
+        } else if (header == "x-rate-limit-reset") {
+            xRateLimitReset = std::string(line.second.constData(), line.second.size());
         }
     }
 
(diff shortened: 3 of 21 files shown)


ivovandongen:
I've implemented the core and platform specifics. All platforms, expect for Android, have the same unit test. It would be nice to also test this against an actual server with a low rate limit as well, just to be sure.

- @jfirebaugh Can you take a look at the core stuff?
- @tmpsantos Same for linux/QT?
- @frederoni iOS/macOS?
- @zugaldia Android?

include/mbgl/storage/response.hpp
```diff
@@ -45,15 +46,24 @@ class Response::Error {
...
 
 public:
-    Error(Reason, std::string = "");
+    //Either; a HttpDate, seconds from now or a Timestamp
+    using RetryAfter = mbgl::variant<std::string, std::chrono::seconds, Timestamp>;
+    
+    Error(Reason, std::string = "", optional<RetryAfter> = {});
+    
+private:
+    optional<Timestamp> parseRetryAfter(const optional<RetryAfter> retryAfter_) const;
```
jfirebaugh: Can you move this to `http_header.hpp` and make the signature:

```
optional<Timestamp> parseRetryHeaders(const optional<std::string>& retryAfter,
                                      const optional<std::string>& xRateLimitReset);
```

Advantages:
* No need for `RetryAfter` type
* Factor conditional and integer conversion logic from individual SDK implementations

test/storage/online_file_source.cpp
```diff
@@ -359,3 +359,56 @@ TEST(OnlineFileSource, TEST_REQUIRES_SERVER(NetworkStatusOnlineOffline)) {
 
     loop.run();
 }
+
+TEST(OnlineFileSource, TEST_REQUIRES_SERVER(RateLimit)) {
```
jfirebaugh: We shouldn't add any more timing-dependent tests to the test suite. The ones we have already spuriously fail intermittently (as will most any timing-dependent test, eventually).

Instead, how about making `errorRetryTimeout` public and unit testing it instead?

ivovandongen:
@jfirebaugh Thanks for taking a look. I've changed the implementations and will rework the tests.

ivovandongen:
@jfirebaugh 
- I've extracted the timeout functions to a header in src/util so it is testable, didn't want to add it to the public header as it is not quite public api. Let me know if this isn't what you had in mind.
- I kept the integration tests so that the platform bindings are tested as well.
- I've added unit tests on the timeout functions directly. 

When writing the unit tests I ran into a little issue where the the timeout became negative on high number of errors as the sign bit flipped ([here](https://github.com/mapbox/mapbox-gl-native/blob/e1a6795a73af844ff63d99b35fb4695177d9494d/platform/default/online_file_source.cpp#L206) among others). Changed all three exponential backoffs so that doesn't happen (`1` -> `1u`) [here](https://github.com/mapbox/mapbox-gl-native/blob/1eebf712623fb7f9cd85a6aac2ba124663216a84/src/mbgl/util/http_timeout.hpp#L14)

src/mbgl/util/http_header.cpp
```diff
@@ -28,6 +28,24 @@ CacheControl CacheControl::parse(const std::string& value) {
 optional<Timestamp> CacheControl::toTimePoint() const {
     return maxAge ? util::now() + Seconds(*maxAge) : optional<Timestamp>{};
 }
+    
+optional<Timestamp> parseRetryHeaders(const optional<std::string>& retryAfter,
+                                      const optional<std::string>& xRateLimitReset) {
+    optional<Timestamp> result;
```
jfirebaugh: Prefer early returns to temporary variables:

```
if (!retryAfter) {
    return ...;
}

try {
    return ...;
} catch ... {
    return ...;
}
```

src/mbgl/util/http_header.cpp
```diff
@@ -28,6 +28,24 @@ CacheControl CacheControl::parse(const std::string& value) {
...
+    
+    if(retryAfter) {
+        try {
+            auto secs = std::chrono::seconds(std::stoi(*retryAfter));
+            result = std::chrono::time_point_cast<Seconds>(std::chrono::system_clock::now() + secs);
+        } catch (std::invalid_argument) {
```
jfirebaugh: `stoi` can also throw `std::out_of_range`. It's probably best to use `catch (...)` here.

src/mbgl/util/http_header.cpp
```diff
@@ -28,6 +28,24 @@ CacheControl CacheControl::parse(const std::string& value) {
...
+            result = util::parseTimestamp((*retryAfter).c_str());
+        }
+    } else if(xRateLimitReset) {
+        result = util::parseTimestamp(std::stoi(*xRateLimitReset));
```
jfirebaugh: Need to handle exceptions from `std::stoi` here.

jfirebaugh: Hmm, actually, do we want `parseRetryHeaders` to swallow malformed header values? I'm not sure. What does existing code do in the case of malformed `Cache-Control`, etc.?

src/mbgl/util/http_timeout.hpp
```diff
@@ -0,0 +1,43 @@
+#pragma once
+
+#include <mbgl/storage/response.hpp>
+#include <mbgl/util/optional.hpp>
+#include <mbgl/util/chrono.hpp>
+
+namespace mbgl {
+namespace http {
+
+static Duration errorRetryTimeout(Response::Error::Reason failedRequestReason, uint32_t failedRequests, optional<Timestamp> retryAfter = {}) {
```
jfirebaugh: Using an internal header for the method declaration is 👍 , but the definition should go in a sibling http_timeout.cpp file. And then remove `static` here.

src/mbgl/util/http_timeout.hpp
```diff
@@ -0,0 +1,43 @@
...
+
+    if (failedRequestReason == Response::Error::Reason::Server) {
+        // Retry after one second three times, then start exponential backoff.
+        return Seconds(failedRequests <= 3 ? 1 : 1u << std::min(failedRequests - 3, 31u));
```
jfirebaugh: Nice catch on the signed integer overflow!

src/mbgl/util/http_header.cpp
```diff
@@ -28,6 +28,26 @@ CacheControl CacheControl::parse(const std::string& value) {
 optional<Timestamp> CacheControl::toTimePoint() const {
     return maxAge ? util::now() + Seconds(*maxAge) : optional<Timestamp>{};
 }
+    
+optional<Timestamp> parseRetryHeaders(const optional<std::string>& retryAfter,
+                                      const optional<std::string>& xRateLimitReset) {
+    if(retryAfter) {
```
tmpsantos: nit, missing space after `if`

src/mbgl/util/http_header.cpp
```diff
@@ -28,6 +28,26 @@ CacheControl CacheControl::parse(const std::string& value) {
...
+        try {
+            auto secs = std::chrono::seconds(std::stoi(*retryAfter));
+            return std::chrono::time_point_cast<Seconds>(std::chrono::system_clock::now() + secs);
+        } catch (...) {
+            return util::parseTimestamp((*retryAfter).c_str());
+        }
+    } else if(xRateLimitReset) {
```
tmpsantos: ditto.

src/mbgl/util/http_timeout.cpp
```diff
@@ -0,0 +1,39 @@
...
+    } else if (failedRequestReason == Response::Error::Reason::Connection) {
+        // Immediate exponential backoff.
+        assert(failedRequests > 0);
+        return Seconds(1u << std::min(failedRequests - 1, 31u));
+    } else if (failedRequestReason == Response::Error::Reason::RateLimit) {
+        if (retryAfter) {
+            return *retryAfter - util::now();
+        } else {
+            //Default
+            return Seconds(5);
```
tmpsantos: Move the magic constant to `./include/mbgl/util/constants.hpp`. We already have similar things like `CLOCK_SKEW_RETRY_TIMEOUT` there.

platform/qt/src/http_request.cpp
```diff
@@ -79,7 +82,12 @@ void HTTPRequest::handleNetworkReply(QNetworkReply *reply)
             response.expires = http::CacheControl::parse(line.second.constData()).toTimePoint();
         } else if (header == "expires") {
             response.expires = util::parseTimestamp(line.second.constData());
+        } else if (header == "retry-after") {
+            retryAfter = std::string(line.second.constData(), line.second.size());
+        } else if (header == "x-rate-limit-reset") {
+            xRateLimitReset = std::string(line.second.constData(), line.second.size());
         }
+
```
tmpsantos: Not needed.

platform/default/http_file_source.cpp
```diff
@@ -325,8 +329,13 @@ size_t HTTPRequest::headerCallback(char *const buffer, const size_t size, const
     } else if ((begin = headerMatches("expires: ", buffer, length)) != std::string::npos) {
         const std::string value { buffer + begin, length - begin - 2 }; // remove \r\n
         baton->response->expires = Timestamp{ Seconds(curl_getdate(value.c_str(), nullptr)) };
+    } else if ((begin = headerMatches("retry-after: ", buffer, length)) != std::string::npos) {
+        baton->retryAfter = std::string(buffer + begin, length - begin - 2); // remove \r\n
+    } else if ((begin = headerMatches("x-rate-limit-reset: ", buffer, length)) != std::string::npos) {
+        baton->xRateLimitReset = std::string(buffer + begin, length - begin - 2); // remove \r\n
     }
 
+
```
tmpsantos: Not needed.

tmpsantos:
Thanks for adding more tests!

ivovandongen:
@tmpsantos Thanks for the review! Made the changes as suggested. Waiting on CI and then I'll merge.

PR #6231, opened 2016-09-01
[ios] Silence MGLCalloutView related warnings in tests

boundsj:
The annotation view test uses a fake MGLTestCalloutView that declares conformance to the MGLCalloutView protocol. However, several properties and a method were not implemented in the test which caused several warnings that were visible if you compiled the SDK tests. This change stubs out the properties and method so the warnings go away.

cc @frederoni @1ec5


1ec5:
👍

PR #6249, opened 2016-09-05
Improve naming in Source

kkaefer:
There are some very generic terms like `update` used in `Source` member functions names. Instead, we should explicitly name these functions after what they are updating.

### DIFF
--- a/src/mbgl/style/source_impl.cpp
+++ b/src/mbgl/style/source_impl.cpp
@@ -77,11 +77,9 @@ const std::map<UnwrappedTileID, RenderTile>& Source::Impl::getRenderTiles() cons
     return renderTiles;
 }
 
-bool Source::Impl::update(const UpdateParameters& parameters) {
-    bool allTilesUpdated = true;
-
-    if (!loaded || parameters.animationTime <= updated) {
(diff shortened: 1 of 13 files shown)


src/mbgl/style/source_impl.cpp
```diff
@@ -77,11 +77,9 @@ const std::map<UnwrappedTileID, RenderTile>& Source::Impl::getRenderTiles() cons
     return renderTiles;
 }
 
-bool Source::Impl::update(const UpdateParameters& parameters) {
-    bool allTilesUpdated = true;
-
-    if (!loaded || parameters.animationTime <= updated) {
```
jfirebaugh: Why did you remove the second conditional?

kkaefer: Because it's never triggered

jfirebaugh: Ok, I thought I remembered this being added somewhat recently for a specific reason, but it seems to date back to ba4294ae0c06c1ed160f0be47863c9515da8d403, which isn't clear about the purpose. If you assert it's no longer necessary I believe you. 😄

Issue #5683, opened 2016-07-14
Highway shield text is tilted


1ec5:
When the map is tilted, Interstate and state route shields remain upright, as expected, but the text inside them tilts with the map:

![gl-native](https://cloud.githubusercontent.com/assets/1231218/16829798/a497b150-494f-11e6-9b2c-180d237c8b17.png)

Reproduces in Streets, Outdoors, and Satellite Streets styles in 5fb0e878b4eb1b2d9555860e86d720e4904de07e on iOS and macOS. Does not reproduce in mapbox/mapbox-gl-js@742b46d7bc46c9489a4bb081036b2a6add7beb40:

![gl-js](https://cloud.githubusercontent.com/assets/1231218/16829793/9df0bfcc-494f-11e6-95ad-497e8ff117fa.png)

Exit shields are unaffected. The difference is likely that highway shields have line placement, whereas exit shields have point placement.

1ec5:
Bisect implicates ecd4aa14397ed081a521d8412557724e52f277f3 (#5288).

/cc @yhahn

yhahn:
@1ec5 sorry for the delay -- I started digging into this yesterday and need to vet what I think is a subtle difference in behavior between the way an undefined pitch alignment value is handled in the style:

- Native https://github.com/mapbox/mapbox-gl-native/blob/master/src/mbgl/style/layers/symbol_layer_impl.cpp#L39-L47
- JS https://github.com/mapbox/mapbox-gl-js/blob/master/js/style/style_layer/symbol_style_layer.js#L25-L38

Will keep trying to get to the bottom of this today.

1ec5:
Just a hunch, but in gl-native, if `symbolPlacement` is `line`, it doesn’t look like it’s possible for either `iconRotationAlignment` or `textRotationAlignment` to be `viewport` for the purpose of a `textPitchAlignment` fallback, even if the stylesheet sets it explicitly. The corresponding code in GL JS only defaults `text-rotation-alignment` to `map` if it isn’t specified in the stylesheet; likewise for `icon-rotation-alignment`.

1ec5:
Indeed, falling back to a temporarily stored value of `text-rotation-alignment` does fix the issue. Here’s what [1ec5-shields-upright-5683](https://github.com/mapbox/mapbox-gl-native/tree/1ec5-shields-upright-5683) looks like:

[...]

PR #6253, opened 2016-09-06
Better handling for undefined icon|text-rotation-alignment

yhahn:
I believe this addresses the root cause behind https://github.com/mapbox/mapbox-gl-native/issues/5683.

- Updates GL style spec to remove an explicit `default` from `text-rotation-alignment` and `icon-rotation-alignment`. This value was a lie given that renderers currently derive the default in coordination with `symbol-placement`
- Update logic in GL native to set the default value for these properties only when they are undefined.

### Remaining backlog 

- [x] @yhahn Render tests to lockdown expected behavior
- [x] @yhahn Vet style spec change, esp possible effect on GL JS and other integrators
- [x] Review @1ec5 @jfirebaugh


src/mbgl/style/layers/symbol_layer_impl.cpp
```diff
@@ -36,12 +36,27 @@ std::unique_ptr<Bucket> SymbolLayer::Impl::createBucket(BucketParameters& parame
...
-        bucket->layout.textRotationAlignment.value = AlignmentType::Map;
+
+    bucket->layout.iconRotationAlignment.calculate(p);
+    if (bucket->layout.iconRotationAlignment.value == AlignmentType::Undefined) {
+        if (bucket->layout.symbolPlacement.value == SymbolPlacementType::Line) {
+            bucket->layout.iconRotationAlignment.value = AlignmentType::Map;
+        } else {
+            bucket->layout.iconRotationAlignment.value = AlignmentType::Viewport;
+        }
+    };
```
1ec5: Nit: extra semicolon.

src/mbgl/style/layers/symbol_layer.cpp
```diff
@@ -125,7 +125,7 @@ void SymbolLayer::setIconOptional(PropertyValue<bool> value) {
     impl->layout.iconOptional.set(value);
 }
 PropertyValue<AlignmentType> SymbolLayer::getDefaultIconRotationAlignment() {
-    return { AlignmentType::Viewport };
+    return { AlignmentType::Undefined };
```
1ec5: Looks like this will cause the SDKs to potentially return a bogus value for these properties by default. `AlignmentType::Undefined` has no analogue in the style specification, so [`MGLSymbolStyleLayerTextRotationAlignment`](https://github.com/mapbox/mapbox-gl-native/blob/7e208c46cebe5b49d4b7ead9826bbe4229ebabd3/platform/darwin/src/MGLSymbolStyleLayer.h#L31) for instance has no corresponding value. On iOS and macOS, this means `-[MGLSymbolStyleLayer textRotationAlignment]` will return a value of 3 that isn’t describable as an `MGLSymbolStyleLayerTextRotationAlignment`.

These default getters were added in order for the iOS and macOS SDKs to implement `null_resettable` semantics (#6126). It would be unfortunate if these SDKs had to special-case these two properties, but I’m not sure how else to avoid having yet another layer of getters for these style properties.

/cc @frederoni

yhahn:
Quick discussion with @1ec5 -- one possible approach to address the special-caseness of these properties would be to add an explicit enum value to the spec like `auto` or `inherit` for `text-rotation-alignment`, `icon-rotation-alignment` and `text-pitch-alignment`.

- Add `auto` enum value
- Set default value to `auto`
- Update logic in GL native/JS to handle this appropriately

I'm fairly certain this would be a breaking change though -- styles explicitly using the value `auto` would not parse (?) with older renderers.

jfirebaugh:
Yes, I think we need to add `auto` for all three of these properties, and treat this new value like we would treat the addition of a new style property -- add a row in the spec property support table documenting the minimum versions requirements.

src/mbgl/style/layers/symbol_layer.cpp
boundsj: Note: this file is generated so we will need to update [generate-style-code.js](https://github.com/mapbox/mapbox-gl-native/blob/master/scripts/generate-style-code.js) to make the required change here (which will probably be to make this [conform to the new `auto` value](https://github.com/mapbox/mapbox-gl-style-spec/pull/493)

jfirebaugh:
Changes here look good. Do we have a test case for the original issue #5683?

boundsj:
@jfirebaugh [a PR for GL JS tests](https://github.com/mapbox/mapbox-gl-test-suite/pull/144) is started and is blocked by https://github.com/mapbox/mapbox-gl-js/pull/3203. We still need to add render tests to this PR.

jfirebaugh:
Style-spec, test-suite, and gl-js have been updated. Remaining work here is to get the tests passing.

jfirebaugh:
Two of the icon-rotation-alignment tests are failing due to very slight positioning differences. 😖  I'll finish up tomorrow.

PR #6255, opened 2016-09-06
Use enum bool classes for better readability

kkaefer:
In a few places, we are using `bool` to denote flags, but it's often hard to guess what true/false actually means, in particular at the call site. Instead, we should move to explicit enums in the style of 

```cpp
enum class TileLoadState : bool {
    First = true,
    Subsequent = false,
};
```

This makes the meaning and intent a lot clearer.


jfirebaugh:
👍

PR #6262, opened 2016-09-06
Extract and clean up WorkTaskImpl

jfirebaugh:
Preparation for #1471. Fixes #5064.


jfirebaugh:
@tmpsantos Can you review here?

tmpsantos:
:+1:, good call on callback as last, IMO makes the code more legible.

PR #6268, opened 2016-09-07
[android] - make using TextureView optional

tobrun:
WIP, allow end user to select between SurfaceView and TextureView.


tobrun:
Closing for now

Issue #6252, opened 2016-09-06
Add initializer to set sourceLayer in MGLStyleLayer


boundsj:
In the [GL JS example for adding a third party vector source](https://www.mapbox.com/mapbox-gl-js/example/third-party) and also in [this Android demo app example](https://github.com/mapbox/mapbox-android-demo/blob/master/MapboxAndroidDemo/src/main/java/com/mapbox/mapboxandroiddemo/examples/styles/VectorSourceActivity.java#L39) the `source-layer` property can be set to a specific layer.

The iOS MGL*StyleLayer classes do not offer a way to set this value either in the initializers in the `MGLStyleLayer` protocol or via a property or setter method. This feature is important for styling sources that have many layers.

cc @1ec5 @frederoni

jfirebaugh:
Note that the core setter isn't hooked up internally to cause a refresh, so it only works when called prior to adding the source to the style. I don't foresee any technical blocker to allowing on-the-fly mutation, but it's probably not a big priority (GL JS doesn't have such an API for example).

boundsj:
Thanks for pointing that out @jfirebaugh. I adjusted the title of this issue and also proposed a fix (to add an init method only, no setter) in https://github.com/mapbox/mapbox-gl-native/pull/6269

Issue #6250, opened 2016-09-05
Using `initWithLayerIdentifier:` with MGL*StyleLayer throws exception


boundsj:
All MGL style layer classes (i.e. `MGLBackgroundStyleLayer`) implement the `MGLStyleLayer` protocol. That protocol declares both `-[MGLStyleLayer initWithLayerIdentifier:]` and `-[MGLStyleLayer initWithLayerIdentifier:sourceIdentifier:]` as optional.

If a developer initializes an `MGLBackgroundStyleLayer` (or any other) with just a layer identifier, their application will compile but crash at runtime if that code path is exercised. We should make the init methods non-optional or at least implement `-[MGLStyleLayer initWithLayerIdentifier:]` in all of the MGL style layer classes.

@1ec5 @frederoni

Issue #6254, opened 2016-09-06
Source and layer ownership is inconsistent (from style versus from developer)


1ec5:
MGLSource and the MGLStyleLayer classes manage their underlying C++ objects inconsistently depending on the task. For example, when getting an MGLSource object from an MGLStyle, the MGLSource object holds a strong reference to an `mbgl::style::Source` object in its `source` property. On the other hand, when the developer initializes an MGLSource object from scratch in order to add it to an MGLMapView, `-[MGLSource mbglSource]` creates the `mbgl::style::Source` object lazily when adding it to the style and never holds a reference to it. If the developer attempts to get an MGLSource from the style and add it back to the style, a brand-new `mbgl::style::Source` object is created and added to the map, discarding any modifications that the developer may have made on the MGLSource object.

We should remove `-[MGLSource mbglSource]`. MGLSource should create the `mbgl::style::Source` upfront in its initializers. (There should be an additional private initializer that accepts an `mbgl::style::Source`.) This means all the source options need to be immutable. Currently, the only mutable source option is the `sourceIdentifier` property, but I think that should be read-only anyways.

The catch is that a source added to the map really shouldn’t be owned by the MGLSource object. Could we treat the `source` property as holding only a “pending” source and always access the `mbgl::style::Source` object via `mbgl::Map::getSource()`? What happens if the developer attempts to add the same MGLSource to the MGLStyles of two different MGLMapViews?

/cc @frederoni @boundsj

jfirebaugh:
To expand on https://github.com/mapbox/mapbox-gl-native/pull/6269#issuecomment-245364132, objects wrapping `mbgl::style::Layer` and `mbgl::style::Source` domain objects need to have two internal members:

* A `std::unique_ptr<>`, which is present only for objects that were created independently (not obtained by `Map#getLayer` or `Map#getSource`) and have not yet been added via `Map#addLayer` or `Map#addSource`.
* A raw pointer to the `mbgl` object, which is always initialized, either to the result of `Map#getLayer` or `Map#getSource`, or for independently created objects, to the pointer value held in the `unique_ptr`. In the latter case, this raw pointer value stays even after _ownership_ of the object is transferred via `Map#addLayer` or `Map#addSource`.

This is the approach that the Android bindings follow.

jfirebaugh:
[...]

PR #6269, opened 2016-09-07
[ios, macos] Put MGLStyleLayer inits in respective classes

boundsj:
Previously, we declared MGLStyleLayer initializers in the MGLStyleLayer protocol as optional methods. In the generated implementation, this made it easier to opt in and out of initializers that did or or did not make sense for the subclass. However, this approach was dangerous since it was easy for an application developer to initialize an MGLStyleLayer subclass with an init method that was actually not implemented in that class causing an exception at runtime.

This commit moves the init methods that each subclass supports to each subclass so xcode (and the compiler) help the developer avoid the previously possible runtime exception.

In addition, a new init method is added that takes the source layer and passes that on to `mbgl::style:Layer` (but only on classes where this is possible in core). This allows an application developer to style a specific source layer (i.e. the contour lines of the mapbox terrain vector source).

Fixes https://github.com/mapbox/mapbox-gl-native/issues/6250
Fixes  https://github.com/mapbox/mapbox-gl-native/issues/6252

@frederoni @1ec5 do you think this approach is reasonable or did I miss a requirement that the initializers should be in the protocol and optional?

### Todo:

- [x] confirm tests work
- [x] test that contour example works
- [x] review from @1ec5 @frederoni 
- [x] clean up commented code
- [x] clean up code generation script

### DIFF
--- a/platform/darwin/scripts/generate-style-code.js
+++ b/platform/darwin/scripts/generate-style-code.js
@@ -187,12 +187,24 @@ global.propertyType = function (property, _private) {
     return _private ? `id <MGLStyleAttributeValue, MGLStyleAttributeValue_Private>` : `id <MGLStyleAttributeValue>`;
 };
 
+global.initLayerIdentifierOnly = function (layerType) {
+    return `_layer = new mbgl::style::${camelize(layerType)}Layer(layerIdentifier.UTF8String);`
+}
+
 global.initLayer = function (layerType) {
     if (layerType == "background") {
-        return `_layer = new mbgl::style::${camelize(layerType)}Layer(layerIdentifier.UTF8String);`
+       return `_layer = new mbgl::style::${camelize(layerType)}Layer(layerIdentifier.UTF8String);` 
     } else {
-        return `_layer = new mbgl::style::${camelize(layerType)}Layer(layerIdentifier.UTF8String, sourceIdentifier.UTF8String);`
+        return `_layer = new mbgl::style::${camelize(layerType)}Layer(layerIdentifier.UTF8String, source.sourceIdentifier.UTF8String);`
     }
+}
+
+global.initLayerWithSourceLayer = function(layerType) {
+    return `_layer = new mbgl::style::${camelize(layerType)}Layer(layerIdentifier.UTF8String, source.sourceIdentifier.UTF8String);`  
+}
+
+global.setSourceLayer = function() {
+   return `_layer->setSourceLayer(sourceLayer.UTF8String);`
 }
 
 global.setterImplementation = function(property, layerType) {
--- a/platform/darwin/src/MGLBackgroundStyleLayer.mm
+++ b/platform/darwin/src/MGLBackgroundStyleLayer.mm
@@ -1,6 +1,7 @@
 // This file is generated. 
 // Edit platform/darwin/scripts/generate-style-code.js, then run `make style-code-darwin`.
 
+#import "MGLSource.h"
 #import "NSPredicate+MGLAdditions.h"
 #import "MGLStyleLayer_Private.h"
 #import "MGLStyleAttributeValue.h"
@@ -13,6 +14,7 @@
 @property (nonatomic) mbgl::style::BackgroundLayer *layer;
 @property (nonatomic, readwrite) NSString *layerIdentifier;
 @property (nonatomic, readwrite) NSString *sourceIdentifier;
+@property (nonatomic, readwrite) NSString *sourceLayerIdentifier;
 
 @end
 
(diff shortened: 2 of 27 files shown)


boundsj:
With this change's `-[MGLFillStyleLayer initWithLayerIdentifier:sourceIdentifier:sourceLayer:]` it is now possible to style a specific layer for a source, like:

```Objc
NSURL *url = [[NSURL alloc] initWithString:@"mapbox://mapbox.mapbox-terrain-v2"];
MGLVectorSource *vectorSource = [[MGLVectorSource alloc] initWithSourceIdentifier:@"terrain-data" URL:url];
[self.mapView.style addSource:vectorSource];
    
// style the "contour" source layer specifically
MGLLineStyleLayer *lineLayer = [[MGLLineStyleLayer alloc] initWithLayerIdentifier:@"terrain-data" sourceIdentifier:@"terrain-data" sourceLayer:@"contour"];

NSUInteger lineJoineValue = (NSUInteger)MGLLineStyleLayerLineJoinRound;
    [lineLayer setLineJoin:[NSValue value:&lineJoineValue withObjCType:@encode(MGLLineStyleLayerLineJoin)]];
NSUInteger lineCapValue = (NSUInteger)MGLLineStyleLayerLineCapRound;
[lineLayer setLineCap:[NSValue value:&lineCapValue withObjCType:@encode(MGLLineStyleLayerLineCap)]];
[lineLayer setLineColor:[UIColor redColor]];
[self.mapView.style addLayer:lineLayer];
```

screen shot 2016-09-06 at 9 59 37 pm

platform/darwin/scripts/generate-style-code.js
```diff
@@ -187,14 +187,22 @@ global.propertyType = function (property, _private) {
...
 };
 
+global.initLayerIdentifierOnly = function (layerType) {
+    return `_layer = new mbgl::style::${camelize(layerType)}Layer(layerIdentifier.UTF8String);`
+}
+
 global.initLayer = function (layerType) {
     if (layerType == "background") {
-        return `_layer = new mbgl::style::${camelize(layerType)}Layer(layerIdentifier.UTF8String);`
+       return `_layer = new mbgl::style::${camelize(layerType)}Layer(layerIdentifier.UTF8String);`
```
frederoni: We're leaking this layer and all other `mbgl::style::layers`. Should we delete it in `-dealloc` or rewrite the getters and setters to make them use the layer in a lazy way?

1ec5: I think we need to take a step back and address this leak in #6254. Until then, any patches will only add to the confusion.

platform/darwin/src/MGLBackgroundStyleLayer.mm
```diff
@@ -13,14 +13,25 @@ @interface MGLBackgroundStyleLayer ()
...
+- (instancetype)initWithLayerIdentifier:(NSString *)layerIdentifier
+{
+    if (self = [super init]) {
+        _layerIdentifier = layerIdentifier;
+        _layer = new mbgl::style::BackgroundLayer(layerIdentifier.UTF8String);
+    }
+    return self;
+}
+
+- (instancetype)initWithLayerIdentifier:(NSString *)layerIdentifier sourceIdentifier:(NSString *)sourceIdentifier
```
1ec5: Let's take this opportunity to replace the sourceIdentifier parameter with a source parameter that takes an MGLSource.

1ec5:
How does the developer discover the layers of a source? Maybe we need a new method on MGLSource for getting an MGLSourceLayer by identifier or a property that returns all the source layers.

boundsj:
@1ec5 how do developers do source layer discovery on JS and Android? I don't see any mention of it in examples like [the GL JS example for adding a third party vector source](https://www.mapbox.com/mapbox-gl-js/example/third-party) or in [this Android demo app example](https://github.com/mapbox/mapbox-android-demo/blob/master/MapboxAndroidDemo/src/main/java/com/mapbox/mapboxandroiddemo/examples/styles/VectorSourceActivity.java#L39). 

Source discovery sounds great but also like a different feature. So far, I just assumed that we expect developers to read the documentation of the source (i.e. https://www.mapbox.com/vector-tiles/mapbox-terrain/)

jfirebaugh:
You should never use `new` or `delete` directly in C++ code, always `std::make_unique` or `std::make_shared`. Layer ownership is not shared, so `std::make_unique<LayerType>(...)`, then `std::move` ownership into `Map#addLayer` when the layer is added. You will likely want to retain a raw pointer for the `MGLStyleLayer`'s use, and for consistency with `style::Layer*`s obtained via `Map#getLayer`.

platform/darwin/src/MGLBackgroundStyleLayer.mm
boundsj: This is done in https://github.com/mapbox/mapbox-gl-native/pull/6269/commits/939cca1a3099b0598e2cdd9276a46790bbb9ed11

boundsj:
I added https://github.com/mapbox/mapbox-gl-native/issues/6285 as a placeholder for adding source layer discovery at some point.

Regarding https://github.com/mapbox/mapbox-gl-native/pull/6269#discussion_r77798514 and https://github.com/mapbox/mapbox-gl-native/pull/6269#issuecomment-245364132 let's take @1ec5's advice and handle that in https://github.com/mapbox/mapbox-gl-native/issues/6254

This PR should be ready to go now.

boundsj:
Ok I think we are good here. Merging.

PR #6086, opened 2016-08-19
[android] #6083 - add circle example

tobrun:
Refs #6083,
add circle example
Review @ivovandongen

PR #6281, opened 2016-09-07
[android] - throw concrete exception when generated test cases fail 

### LINKED ISSUE #6279: Throw concrete exception when IdlingResources times out
Auto generated test cases landed with #5859, to correctly support this feature when adding a new Activity we need to add MapboxMap as a member of that class. When not following this rule an IdlingResourceException is thrown ([example](https://cloud.githubusercontent.com/assets/2151639/17808168/d001c458-660e-11e6-98e2-fbf668dc2f98.png)). This exception doesn't indicate what actions can be taken to resolve the issue. I'm proposing to throw a more concrete exception.

tobrun:
Closes #6279, improves experience around the issue documented in https://github.com/mapbox/mapbox-gl-native/pull/6086#issuecomment-240996903.

Review @zugaldia


zugaldia:
@tobrun 👍

PR #6289, opened 2016-09-08
[core] - add unit test for wrap

tobrun:
Adds core unit tests for `wrap.hpp`.

cc @jfirebaugh @tmpsantos

### DIFF
--- a/wrap.cpp
+++ b/wrap.cpp
@@ -6,21 +6,21 @@
 using namespace mbgl;
 
 TEST(Math, WrapHigherValue) {
-    ASSERT_DOUBLE_EQ(4, util::wrap(16, 0, 12));
+    ASSERT_DOUBLE_EQ(4.0, util::wrap(16.0, 0.0, 12.0));
 }
 
 TEST(Math, WrapLowerValue) {
-    ASSERT_DOUBLE_EQ(9, util::wrap(-3, 0, 12));
+    ASSERT_DOUBLE_EQ(9.0, util::wrap(-3.0, 0.0, 12.0));
 }
 
 TEST(Math, WrapInRangeValue) {
-    ASSERT_DOUBLE_EQ(4, util::wrap(4, 0, 12));
+    ASSERT_DOUBLE_EQ(4.0, util::wrap(4.0, 0.0, 12.0));
 }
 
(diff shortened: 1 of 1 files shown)


jfirebaugh:
Note that `wrap` is templated on the input type, and when you call it without specifying the template parameter -- e.g. `wrap(1, 2, 3)` rather than `wrap<double>(1, 2, 3)` -- the value of the template parameter is _deduced_ from the type of the parameters provided. In this case, the integer literals you are using have type `int`, so what you're testing is `wrap<int>`.

Looking at the definition of `wrap`, `wrap<int>` will call the fourth variant of `std::fmod` listed [here](http://en.cppreference.com/w/cpp/numeric/math/fmod), which promotes the result type to a `double`. _However_, because the _return type_ of `wrap` is defined as `T`, the same as the inputs, `wrap<int>` will end up truncating this `double` back to an `int`! But it looks like truncation is safe, because all the _intermediate_ states of computation will still be done on `double`s, and the end result of wrapping integers should never have a fractional component. Phew!

In practice, I believe `wrap` is only instantiated with floating point types. If you intend to test floating point wrapping, it's best to write floating point literals (`4.0`, `16.0`, etc). And if you intend to test integer wrapping, then it's best to use `ASSERT_EQ` as opposed to `ASSERT_DOUBLE_EQ`.

Otherwise, this looks great!

tobrun:
Thank you for 👀 and even more for the explanation! 
I adjusted the code accordingly to use doubles instead of integers.  

> Note that wrap is templated on the input type, and when you call it without specifying the template parameter -- e.g. wrap(1, 2, 3) rather than wrap<double>(1, 2, 3) -- the value of the template parameter is deduced from the type of the parameters provided. In this case, the integer literals you are using have type int, so what you're testing is wrap<int>.

Do you prefer setting this explicitly?

jfirebaugh:
> Do you prefer setting this explicitly?

No, generally it's best to rely on template type deduction whenever possible. These tests are great, merge on 🍏 .

tmpsantos:
:rocket:

friedbunny:
@tobrun This added `wrap.cpp` to the base directory of the repo.

tobrun:
Thanks for noticing, fixing up

Issue #6244, opened 2016-09-02
Translate more NSComparisonPredicate operators to filters


1ec5:
Per https://github.com/mapbox/mapbox-gl-native/pull/6049#discussion_r76087268 and https://github.com/mapbox/mapbox-gl-native/pull/6049#discussion_r77412484, we should translate a few more NSPredicate operators to `mbgl::style::Filter`s to prevent information loss:

* [x] NSComparisonPredicate with an operator type of `NSContainsPredicateOperatorType` should be represented by an `mbgl::style::InFilter` with the right expression as the key and the left expression as the values.
* [x] NSComparisonPredicate with an operator type of `NSBetweenPredicateOperatorType` should be represented by an `mbgl::style::AllFilter` of `mbgl::style::GreaterThanEqualsFilter` (corresponding to the first element of the right expression) and `mbgl::style::LessThanEqualsFilter` (the second element of the right expression).
* [x] NSComparisonPredicate with an operator type of `NSEqualToPredicateOperatorType` and a right expression of `nil` should be represented by an `mbgl::style::NotHasFilter`.
* [x] NSComparisonPredicate with an operator type of `NSNotEqualToPredicateOperatorType` and a right expression of `nil` should be represented by an `mbgl::style::HasFilter`.
* [x] NSPredicate with a format of `NSTRUEPREDICATE` should be represented as an `mbgl::style::AllFilter` with no subfilters (same as an NSComparisonPredicate with a type of `NSAndPredicateType`).
* [x] NSPredicate with a format of `NSFALSEPREDICATE` should be represented as an `mbgl::style::AnyFilter` with no subfilters (same as an NSComparisonPredicate with a type of `NSOrPredicateType`).

/cc @frederoni @boundsj

1ec5:
Per https://github.com/mapbox/mapbox-gl-native/issues/6248#issuecomment-245182840, if you try to get the predicate of an existing layer that has one of these operators (quite common in Streets), the SDK crashes.

PR #6290, opened 2016-09-08
fixes #6248: handle all NSNumber types

### LINKED ISSUE #6248: Unable to round-trip NSInteger in predicate
Hello! I'm facing some unexpected behavior of the MGLStyleLayer predicate using ios-v3.4.0-alpha.4.

1) I load a style json with some filter inside like this:
`"filter": ["all",["in","id",2467,2269,559,2906,852,189,3027,5597,188,765,190,633,146]]`

Then I take a predicate from a current style after loading the app and try to apply it back again:
```
MGLLineStyleLayer *roadLayer = (MGLLineStyleLayer *)[_mapView.style layerWithIdentifier:@"stagnation"];
NSPredicate *predicateOld = roadLayer.predicate;
roadLayer.predicate = predicateOld;
```
And here is the exception rising at once:
`*** Terminating app due to uncaught exception 'Value not handled', reason: 'Can’t convert q:2467 to mbgl::Value'`

All the elements in array are coming with (long) format.

2) Next case is when I load a style json again with a filter:
`"filter": ["all",["in","id",2467,2269,559,2906,852,189,3027,5597,188,765,190,633,146]]`

Then I build a predicate just in a few millisecs (asynchronosly) after mapView initializing (while the map is rendering on the screen) and try to apply it:
```
MGLLineStyleLayer *roadLayer = (MGLLineStyleLayer *)[_mapView.style layerWithIdentifier:@"stagnation"];
NSPredicate *predicateNew = [NSPredicate predicateWithFormat:@"id IN %@", @[@2467, @2269, @559, @2906, @852, @189, @3027, @5597, @188, @765, @190, @633, @146]];
roadLayer.predicate = predicateNew;
```
And after that the layer disappears from the map.

3) Same as case 2 but now I try to change a predicate in several seconds after map initializing. A layer disappears later. Looks like it exists in cache for a short period of time and then goes off.

In general my goal is to change selected road segment colors on demand based on road traffic data.

incanus:


platform/darwin/src/NSExpression+MGLAdditions.mm
```diff
@@ -27,12 +27,37 @@ @implementation NSExpression (MGLAdditions)
...
+        } else if (strcmp([number objCType], @encode(long long)) == 0) {
+            return { number.longLongValue };
+        } else if (strcmp([number objCType], @encode(unsigned long long)) == 0) {
+            return { number.unsignedLongLongValue };
+        } else if (strcmp([number objCType], @encode(float)) == 0) {
+            return { number.floatValue };
+        } else if (strcmp([number objCType], @encode(double)) == 0) {
             return { number.doubleValue };
-        } else if ((strcmp([number objCType], @encode(char))) == 0) {
+        } else if (strcmp([number objCType], @encode(bool)) == 0) {
```
1ec5: BOOL is another type that is represented differently between 32- and 64-bit systems. 7e208c46cebe5b49d4b7ead9826bbe4229ebabd3 broke Boolean conversions on 32-bit systems while fixing it for 64-bit systems; this change does the opposite, breaking 64-bit but fixing 32-bit. We need both cases, ideally predicated (hah!) on the architecture via conditional compilation.

boundsj: As best I can tell, https://github.com/mapbox/mapbox-gl-native/commit/7e208c46cebe5b49d4b7ead9826bbe4229ebabd3 is actually ok and we want `... if ((strcmp([number objCType], @encode(char))) == 0)`.

We used to compare `[number objCType]` (returns `c` for boolean on both 32 and 64 bit system to `@encode(bool)` which returns `c` on 32 bit and `b` on 64 bit. `7e208c4` changed to compare `[number objCType]` to `@encode(char)` which always returns `c`.

boundsj:
> Can you add tests exercising all the additional code paths?

`MGLFilterTests` is probably a good place to add tests for the new types.

platform/darwin/src/NSExpression+MGLAdditions.mm
boundsj: I tested this branch with https://gist.github.com/boundsj/c39973346a27104a3fad1647a37606e6 (it uses a predicate with a boolean value) and this commit works by accident in that applied example since the boolean is converted into a `charValue` (first conversion in the new code above) and passed to core and core treats that as `true`. However, I think something like `[unclusteredLayer setPredicate:[NSPredicate predicateWithFormat:@"%K != NO", @"cluster"]];` would not work as expected since the `NO` value (when converted to a `charValue`) would still be `true`.

incanus: Yeah there are some weird things going on with booleans still, more to come. Been digging into this.

incanus: Re: weird things, namely, regardless of platform and handling, there is currently no way to tell if a `char` value (e.g., `17`) should be handled as a number and not a boolean. When we _do_ actually get a boolean parsed out of an `NSExpression`, the class cluster type inside `NSNumber` is `NSCFBoolean`, so I'm currently working on inspecting that a bit better.

incanus:
Ok, https://github.com/mapbox/mapbox-gl-native/pull/6290/commits/30434096ad90307771474687e95e97e393caae5e fixes things, I think. Well, it gets the types right now. Filters still aren't working in my use case (`@"uniqid == 10191020"` or the formatted argument `NSNumber` variants like `@"uniqid = %@", @10191020`). 

I _think_ we want the behavior in this commit for `float` values on 32-bit platforms. Since our `mbgl::geometry::value` only supports `double` for decimal values, on 32-bit, `float` values as passed to an `NSExpression` by `-[NSNumber numberWithFloat:]` get cast to `double` and this can introduce imprecision. For example, I had a value like `23.23430` that got passed down to core as `23.23430000232`. If we avoid `float` use on 32-bit, we avoid this. Most folks will probably use `-[NSNumber numberWithDouble:]` or boxed literals (`@(23.23430)`) anyway, which get interpreted as `double`. 

Regardless of 32- or 64-bit, we interpret both `char` (`c`, 32) and `BOOL` (`B`, 64) as booleans now. 

Note also the commentary in https://github.com/mapbox/mapbox-gl-native/pull/6290/commits/30434096ad90307771474687e95e97e393caae5e#diff-698c604c23220e922e9df03699e408abR58 about how we never seem to get back actually unsigned `uint64_t` from `NSExpression`, but I am leaving the code path in to cover our bases for completeness. I could be missing a subtlety here. 

More to come. 

@boundsj @1ec5 @frederoni @jfirebaugh

platform/darwin/src/NSExpression+MGLAdditions.mm
```diff
@@ -27,37 +29,47 @@ @implementation NSExpression (MGLAdditions)
...
-        } else if (strcmp([number objCType], @encode(NSInteger)) == 0) {
-            return { number.integerValue };
-        } else if (strcmp([number objCType], @encode(NSUInteger)) == 0) {
-            return { number.unsignedIntegerValue };
+            // Double values are interpreted precisely on all platforms.
+            return { (double)number.doubleValue };
+        } else if (thirtyTwoBit && strcmp([number objCType], @encode(float)) == 0) {
+            // Float values are converted to double on 32-bit platforms and
+            // introduce precision problems.
+            [NSException raise:@"Float values introduce imprecision on 32-bit systems"
```
1ec5: This might be too harsh for what could end up being a pretty common developer scenario. Can we avoid the overprecision issue by using `doubleValue` instead of `floatValue`? As a last resort, how about logging a one-time warning instead of raising an exception?

platform/darwin/src/NSExpression+MGLAdditions.mm
```diff
@@ -27,37 +29,47 @@ @implementation NSExpression (MGLAdditions)
...
+            // of these types by NSExpression regardless of signedness.
+            return { (int64_t)number.integerValue };
+        } else if (strcmp([number objCType], @encode(unsigned char))      == 0 ||
+                   strcmp([number objCType], @encode(unsigned short))     == 0 ||
+                   strcmp([number objCType], @encode(unsigned int))       == 0 ||
+                   strcmp([number objCType], @encode(unsigned long))      == 0 ||
+                   strcmp([number objCType], @encode(unsigned long long)) == 0 ||
+                   strcmp([number objCType], @encode(NSUInteger))         == 0) {
+            // This code path does not seem to get run in current NSExpression
+            // implementations (even with e.g. [NSNumber numberWithLong:INT_MAX]),
```
1ec5: What about `@(NSIntegerMax)`, which should be much larger than `@(INT_MAX)`?

platform/darwin/src/NSExpression+MGLAdditions.mm
```diff
@@ -27,37 +29,47 @@ @implementation NSExpression (MGLAdditions)
...
-            return { number.longLongValue };
-        } else if (strcmp([number objCType], @encode(unsigned long long)) == 0) {
-            return { number.unsignedLongLongValue };
-        } else if (strcmp([number objCType], @encode(float)) == 0) {
-            return { number.floatValue };
+        NSLog(@"NSNumber %@ (%s)", number, number.objCType);
+
+        // 32/64-bit check based on CGFloat storage size
+        // per https://developer.apple.com/library/ios/documentation/General/Conceptual/CocoaTouch64BitGuide/
+        BOOL thirtyTwoBit = sizeof(CGFloat) == sizeof(float);
```
1ec5: Since this is an Objective-C++ file, we may be able to perform this same size comparison at compile time with `constexpr`s.

Incidentally, NSObjCRuntime.h performs this check at compile-time with `__LP64__ || (TARGET_OS_EMBEDDED && !TARGET_OS_IPHONE) || TARGET_OS_WIN32 || NS_BUILD_32_LIKE_64`.

incanus: I found a number of other ways that work, but this definitely does, reads cleanly, and has a concrete documentation link explaining the approach. So maybe just keep it?

platform/darwin/src/NSExpression+MGLAdditions.mm
incanus: Taking the `doubleValue` effectively gets the same result as the `double` cast, so no-go. Using `float` is just dangerous territory here, but yeah, a log warning could suffice.

platform/darwin/src/NSExpression+MGLAdditions.mm
incanus: Nice, got one! On 64-bit this triggers the unsigned code path with a `Q` (and a value of `18446744073709551615` in my testing).

incanus: Comments updated in 18e84ca5e98fc73dc480a7b009492b98051716d8. Now I feel better about leaving this code path in. 😄

platform/darwin/src/NSExpression+MGLAdditions.mm
```diff
@@ -27,12 +29,45 @@ @implementation NSExpression (MGLAdditions)
...
-            return { number.doubleValue };
-        } else if ((strcmp([number objCType], @encode(char))) == 0) {
-            return { number.boolValue };
+        NSLog(@"NSNumber %@ (%s)", number, number.objCType);
+
+        // 32/64-bit check based on CGFloat storage size
+        // per https://developer.apple.com/library/ios/documentation/General/Conceptual/CocoaTouch64BitGuide/
+        BOOL thirtyTwoBit = sizeof(CGFloat) == sizeof(float);
+        BOOL sixtyFourBit = sizeof(CGFloat) == sizeof(double);
+        NSAssert((thirtyTwoBit || sixtyFourBit) && ! (thirtyTwoBit && sixtyFourBit), @"Fault in 32/64-bit determination");
```
1ec5: This should be a `static_assert()` so the build would fail if this turned out to be false for some reason.

platform/darwin/src/NSExpression+MGLAdditions.mm
incanus: Eh, this happens on 64-bit, too. I just got `23.232300` as a `-numberWithFloat:` turned into `23.2322998046875`. So we need to warn regardless of platform.

incanus: 90ff91581bcf44d42a4c1fc150cf3a553fb1f4b9

platform/darwin/src/NSExpression+MGLAdditions.mm
incanus: Doesn't work here, it's a non-constant expression and requires runtime to ascertain the `sizeof` bits.

incanus:
> Can you add tests exercising all the additional code paths?

Begun in https://github.com/mapbox/mapbox-gl-native/pull/6290/commits/c3f90c3e4cc2a036f2bb8463db91e547fe745e8c.

1ec5:
* [ ] Also test the minimum and maximum possible values for each type

platform/darwin/test/MGLFilterTests.mm
```diff
@@ -20,7 +20,6 @@ - (void)setUp
     source = [[MGLGeoJSONSource alloc] initWithSourceIdentifier:@"test-source" geoJSONData:geoJSONData];
     [self.mapView.style addSource:source];
     layer = [[MGLLineStyleLayer alloc] initWithLayerIdentifier:@"test-layer" sourceIdentifier:@"test-source"];
-    [self.mapView.style addLayer:layer];
```
boundsj: @incanus I just realized that the reason the simple test was breaking is that, since the `tearDown` method calls `removeLayer` and since that code path throws an exception if you attempt to remove a layer that has not been added, adding any test that does not add the tests `layer` ivar will cause the error we saw. I think we can just chalk this up to shared setup and cleanup methods sometimes causing issues like this. In any case, breaking the tests apart was the way to go here.

incanus:
This is looking basically done, pending some test adjustments in order to pass on 32-bit systems. Getting some failures around the `uint64_t` type test maximums.

platform/darwin/src/NSExpression+MGLAdditions.mm
```diff
@@ -27,12 +29,39 @@ @implementation NSExpression (MGLAdditions)
...
+            return { (uint64_t)number.unsignedIntegerValue };
+        } else if (strcmp([number objCType], @encode(double)) == 0) {
+            // Double values on all platforms are interpreted precisely.
+            return { (double)number.doubleValue };
+        } else if (strcmp([number objCType], @encode(float)) == 0) {
+            // Float values when taken as double introduce precision problems,
+            // so warn the user to avoid them. This would require them to
+            // explicitly use -[NSNumber numberWithFloat:] arguments anyway.
+            // We still do this conversion in order to provide a valid value.
+            mbgl::Log::Warning(mbgl::Event::Style, "float values are converted to double and can introduce imprecision; please use double values explicitly in predicate arguments");
```
1ec5: Why `mbgl::Log`? Since this is in platform-specific code, we can issue an `NSLog()` directly. We should also stick it in a `dispatch_once()`, like we did in #6163, to avoid potentially spamming the console.

1ec5:
By the way, we currently test the conversion in the other direction, from `mbgl::Value` to NSExpression in `ValueEvaluator`, in `-[MGLFeatureTests testPropertyConversion]`. (We even have a `(TARGET_OS_IPHONE && __LP64__)  ||  TARGET_OS_WATCH` conditional in there.) Should we break these tests out into MGLFilterTests too, or should we leave them as integration tests that involve feature querying?

platform/darwin/src/NSExpression+MGLAdditions.mm
incanus: 7fdd5ae16b370a0d8402bb075daf10c415069b5f

incanus:
Ok, things are much cleaner, more correct, and tests are now passing on both 32- and 64-bit test simulators as of f6350f6 and https://github.com/mapbox/mapbox-gl-native/pull/6290/commits/d0762f77539649108630c117b491cd10b3384eb6. 

@jfirebaugh helped me realize that `mbgl::Value` _tells us_ how to decide whether integers will be encoded as `int64_t` (negative) or `uint64_t` (positive or zero), which simplifies things somewhat and makes them more predictable. I also figured out that we always need to deal with the (signed or unsigned) `long long` value in order to avoid overflow, since `mbgl::Value` is meant to handle the largest of integers. This helps us avoid all of the introspection and `@encode` comparison for integers of any type other than `char` and `BOOL` (which we treat as 32- and 64-bit booleans) and just look at the sign. 

@waphumanoid, would you mind confirming whether this fixes https://github.com/mapbox/mapbox-gl-native/issues/6248 for you? Assuming this checks out and tests pass for https://github.com/mapbox/mapbox-gl-native/pull/6290/commits/b9c1e1e7defb3fc6517261591efba47a0d499ffb, this will be ready to 🚢 .

platform/darwin/test/MGLExpressionTests.mm
```diff
@@ -44,184 +51,126 @@ - (void)testExpressionConversionBooleanFalse
...
-    XCTAssertEqual(convertedValue.get<int64_t>(), LLONG_MIN);
+    // Because we can't guarantee precision when using float, and because
+    // we warn the user to this effect in mgl_convertedValueWithValue:,
+    // just always pass this test.
 
-    predicate = [self equalityComparisonPredicateWithRightConstantValue:[NSNumber numberWithLongLong:LLONG_MAX]];
-    convertedValue = predicate.rightExpression.mgl_filterValue;
-    XCTAssert(convertedValue.is<int64_t>() == true);
-    XCTAssertEqual(convertedValue.get<int64_t>(), LLONG_MAX);
+    XCTAssertTrue(YES);
```
1ec5: We can still assert that certain values like -1 are round-tripped, and we can still do imprecise comparisons to make sure it's in the right ballpark.

platform/darwin/src/NSExpression+MGLAdditions.mm
```diff
@@ -27,12 +27,33 @@ @implementation NSExpression (MGLAdditions)
...
+            // Double values on all platforms are interpreted precisely.
+            return { (double)number.doubleValue };
+        } else if (strcmp([number objCType], @encode(float)) == 0) {
+            // Float values when taken as double introduce precision problems,
+            // so warn the user to avoid them. This would require them to
+            // explicitly use -[NSNumber numberWithFloat:] arguments anyway.
+            // We still do this conversion in order to provide a valid value.
+            static dispatch_once_t onceToken;
+            dispatch_once(&onceToken, ^{
+                NSLog(@"Float values are converted to double and can introduce imprecision; please use double values explicitly in predicate arguments.");
```
1ec5: In the message, say that the message won't be shown again. This is a common pattern for ~~limited-time only offers~~ one-time log messages.

platform/darwin/test/MGLExpressionTests.mm
```diff
@@ -0,0 +1,176 @@
...
+    XCTAssertEqual(convertedValue.get<double>(), DBL_MAX);
+}
+
+- (void)testExpressionConversionFloat
+{
+    // Because we can't guarantee precision when using float, and because
+    // we warn the user to this effect in mgl_convertedValueWithValue:,
+    // just always pass this test.
+
+    XCTAssertTrue(YES);
```
1ec5: We should at least exercise the code path to ensure we convert to something in the right ballpark.

waphumanoid:
@incanus, Sorry I can't check it since I'm not able to compile all the project :( . I use it via podspecs

incanus:
@waphumanoid Not a huge deal, though you can do this if you want. 

1. Clone the repo & install dependencies. 
1. `make iframework`
1. Use `pod 'Mapbox-iOS-SDK', :path => '<clone folder>/build/ios/pkg/dynamic/Mapbox-iOS-SDK.podspec'`

platform/darwin/test/MGLExpressionTests.mm
incanus: a486f3390543684c7fd63478938a00f3e52431b6

platform/darwin/src/NSExpression+MGLAdditions.mm
incanus: 05874a4154ea9f687e98c783e2a11316a0ac6510

platform/darwin/test/MGLExpressionTests.mm
incanus: a486f3390543684c7fd63478938a00f3e52431b6

incanus:
Ok, gonna rebase on `master` and squash this, then I think it's good to go! I may review it myself, but anyone else is welcome to, also.

incanus:
Squashed into https://github.com/mapbox/mapbox-gl-native/pull/6290/commits/fc9aa3028184fe0b4ff70f7f311843ad9918a62b. When that passes, I'll merge locally and push up to close this out.

incanus:
It passed, but based on elapsed time I had to rebase _again_ and then still couldn't merge at the command line as things are blocked by this new review system. Can I get someone to review this as of https://github.com/mapbox/mapbox-gl-native/pull/6290/commits/63041032654f15970708316675078ab4fbb72780 so I can land this?

waphumanoid:
@incanus, Thanks, I did compile the master brunch (with a little issue which I will post separately) and  give it some tests. It works as expected so far. Still testing and looking forward to seeing an implementation for `[NSException raise:@"Unsupported filter type"
format:@"Cannot convert mbgl::style::NotHasFilter to NSPredicate"];`

1ec5:
> looking forward to seeing an implementation for `[NSException raise:@"Unsupported filter type"
format:@"Cannot convert mbgl::style::NotHasFilter to NSPredicate"];`

This is being tracked in #6244.

PR #6291, opened 2016-09-08
Implement SQLite error logging

### LINKED ISSUE #6267: Implement SQLite error/warning logging callback
Per http://www.sqlite.org/errlog.html:

>SQLite can be configured to invoke a callback function containing an error code and a terse error message whenever anomalies occur. This mechanism is very helpful in tracking obscure problems that occur rarely and in the field. Application developers are encouraged to take advantage of the error logging facility of SQLite in their products, as it is very low CPU and memory cost but can be a huge aid for debugging.

This will be helpful in diagnosing and ultimately fixing #5651, #6193, and other SQLite exceptions that might occur.

/cc @jfirebaugh @tobrun @kkaefer

friedbunny:
Implements [SQLite error logging](http://www.sqlite.org/errlog.html) in a very verbose fashion — for now, everything gets logged. As we see what comes down the tube, we can selectively silence common or irrelevant messages.

**Notes**
- `sqlite3_config()` must be called [before the database is initialized](http://www.sqlite.org/c3ref/config.html) (which happens automatically via `sqlite3_open_v2()`).
- I’m very unconfident about my type choices, would appreciate feedback.

Fixes #6267.

/cc @jfirebaugh @kkaefer @tobrun

### DIFF
--- a/platform/default/sqlite3.cpp
+++ b/platform/default/sqlite3.cpp
@@ -7,6 +7,8 @@
 #include <chrono>
 #include <experimental/optional>
 
+#include <mbgl/platform/log.hpp>
+
 const static bool sqliteVersionCheck __attribute__((unused)) = []() {
     if (sqlite3_libversion_number() / 1000000 != SQLITE_VERSION_NUMBER / 1000000) {
         char message[96];
@@ -26,6 +28,7 @@ template <typename T>
 using optional = std::experimental::optional<T>;
 
 Database::Database(const std::string &filename, int flags) {
+    sqlite3_config(SQLITE_CONFIG_LOG, errorLogCallback, nullptr);
     const int err = sqlite3_open_v2(filename.c_str(), &db, flags, nullptr);
     if (err != SQLITE_OK) {
         const auto message = sqlite3_errmsg(db);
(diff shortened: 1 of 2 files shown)


friedbunny:
Clang-tidy failed:

```
Caught clang-tidy warning/error:
platform/default/sqlite3.cpp:61:39: warning: unused parameter 'arg' [clang-diagnostic-unused-parameter]
void Database::errorLogCallback(void *arg, const int err, const char *msg) {
```

... which is legit, because `arg` is unused. This is because we pass `nullptr` for the custom user data parameter:

> The third argument to the sqlite3_config(SQLITE_CONFIG_LOG,...) interface (the "pData" argument in the example above) is a pointer to arbitrary data. SQLite passes this pointer through to the first argument of the error logger callback. The pointer can be used to pass application-specific setup or state information, if desired. Or it can simply be a NULL pointer which is ignored by the callback.

What’s our preferred way of dealing with unused parameters?

friedbunny:
Here’s the only message I’ve seen so far, as an example of the output:

>[INFO] {DefaultFileSource}[Database]: recovered 211 frames from WAL file /Library/Developer/CoreSimulator/.../com.mapbox.MapboxGL/.mapbox/cache.db-wal (Code 283)

This message seems to come every time the database is started (i.e., on app launch).

More on SQLite result codes [here](https://www.sqlite.org/rescode.html) — `283` is [`SQLITE_NOTICE_RECOVER_WAL`](https://www.sqlite.org/rescode.html#notice_recover_wal).

jfirebaugh:
> What’s our preferred way of dealing with unused parameters?

Omit the argument name: `void Database::errorLogCallback(void *, const int err, const char *msg)`.

LGTM. This should help immensely.

friedbunny:
Thanks @jfirebaugh, I’ve removed the argument name and pushed a rebased branch.

PR #6294, opened 2016-09-08
6277 disable leak canary during instrumenation tests

### LINKED ISSUE #6277: Disable strictmode and leak canary for CI testing
We are using both LeakCanary and Strict mode to make our code more robust but these tools interfere with our automated tests on CI. We need to look into disabling these during tests.

tobrun:
closes #6277, aims to make instrumentation tests more robust

Review @zugaldia


zugaldia:
@tobrun 👍

Issue #6295, opened 2016-09-08
Fix/silence SQLite log messages at first launch


friedbunny:
The new SQLite logging from #6291 now shows us a few messages on database creation. We should decide if these represent bugs or just messages that we can silence.
- [`SQLITE_CANTOPEN`](https://www.sqlite.org/rescode.html#cantopen) (14) is expected because of how we fail-open the database on first launch.
- [`SQLITE_MISUSE`](https://www.sqlite.org/rescode.html#misuse) (21) isn’t clear and worries me a bit. → #6296

iOS (simulator):

```
[INFO] {DefaultFileSource}[Database]: cannot open file at line 38073 of [91e811f51e] (Code 14)
[INFO] {DefaultFileSource}[Database]: [91e811f51e]:38073: (3) open(/Library/Developer/CoreSimulator/Devices/.../com.mapbox.MapboxGL/.mapbox/cache.db) -  (Code 14)
[INFO] {DefaultFileSource}[Database]: misuse at line 141285 of [91e811f51e] (Code 21)
[INFO] {DefaultFileSource}[Database]: misuse at line 141285 of [91e811f51e] (Code 21)
```

Android (emulator):

```
{DefaultFileSour}[Database]: cannot open file at line 32438 of [767c1727fe] (Code 14)
{DefaultFileSour}[Database]: os_unix.c:32438: (2) open(/data/user/0/com.mapbox.mapboxsdk.testapp/files/mbgl-offline.db) - No such file or directory (Code 14)
{DefaultFileSour}[Database]: misuse at line 131142 of [767c1727fe] (Code 21)
{DefaultFileSour}[Database]: misuse at line 131142 of [767c1727fe] (Code 21)
```

/cc @jfirebaugh

jfirebaugh:
`SQLITE_MISUSE` sounds like we could be doing something we shouldn't.

Do you have any idea what the magic numbers like `141285` and `91e811f51e` refer to?

jfirebaugh:
I set a breakpoint inside `errorLogCallback`, and can see that `SQLITE_MISUSE` is coming from the second call to `sqlite3_config`. I think we're failing to obey this part of the documentation:

> The sqlite3_config() interface may only be invoked prior to library initialization using sqlite3_initialize() or after shutdown by sqlite3_shutdown(). If sqlite3_config() is called after sqlite3_initialize() and before sqlite3_shutdown() then it will return SQLITE_MISUSE.

[...]

PR #6296, opened 2016-09-08
Move SQLite logging setup to before database initialization

friedbunny:
Fixes the `SQLITE_MISUSE` warning on the first launch of an app — half of #6295.

Follows-up on the initial SQLite logging implementation in #6291.

/cc @jfirebaugh

### DIFF
--- a/platform/default/sqlite3.cpp
+++ b/platform/default/sqlite3.cpp
@@ -9,6 +9,12 @@
 
 #include <mbgl/platform/log.hpp>
 
+namespace mapbox {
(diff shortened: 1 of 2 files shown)


jfirebaugh:
👍 

If you want you can move that whole static initializer expression into the namespace to avoid qualifying `errorLogCallback`.

friedbunny:
Moving `sqliteVersionCheck` into the namespace resulted in a slightly strange diff, but it works. 😉

Issue #2851, opened 2015-10-28
rethink partial tiles


jfirebaugh:
The logic around partial tiles is complex and brittle. We need to rethink this, maybe do it in a totally different way.

cc @tmpsantos @kkaefer

kkaefer:
What aspect of it is causing problems for you?

jfirebaugh:
Our current implementation is [highly stateful](https://github.com/mapbox/mapbox-gl-native/blob/d20bcb07957edc008d452db8ef5c91baa117bce6/src/mbgl/map/tile_data.hpp#L22-L60), and it is difficult to reason about the order or timing of state transitions. This has given rise to pattern of issues where tiles either never finish parsing or are parsed _ad infinitum_:

* https://github.com/mapbox/mapbox-gl-native/issues/2808
* https://github.com/mapbox/mapbox-gl-native/issues/1565
* https://github.com/mapbox/mapbox-gl-native/issues/1458
* https://github.com/mapbox/mapbox-gl-native/pull/2393
* https://github.com/mapbox/tilelive-gl/issues/25

These issues are often very difficult and time consuming to debug.

Yet despite its complexity, our current strategy is still inefficient:

* https://github.com/mapbox/mapbox-gl-native/issues/2511

I'm not sure what the right approach is yet, however.

tmpsantos:
We need a way of making tiles subscribe to the resources they need and get notified when they arrived. And I'm all in for getting rid of tile states. They will probably still exist, but IMO they could be inferred at runtime.

kkaefer:
Partially addressed by #5139

tmpsantos:
> Partially addressed by #5139

:+1:

jfirebaugh:
Lots of refactoring here lately. Fixing #667 will likely fix #2511 and then things will be pretty clean.

Issue #6263, opened 2016-09-06
Fix race condition with GeometryTile::redoLayout


jfirebaugh:
Followup from #6201.

`GeometryTile::redoLayout` assumes that the worker has tile data from a previous call to `worker.parseGeometryTile(...)`, but it also cancels the pending work request, which might be the very request that provides the worker with the tile data.

The way I'd like to fix this is to change the worker API to:

```
void setGeometryTileData(worker, data); // invoked in "fire and forget" fashion, not cancellable
Request startLayout(worker, ..., callback); // replaces parseGeometryTile, cancellable
Request continueLayout(worker, ..., callback); // replaces parsePendingGeometryTileLayers, cancellable
```

However, this depends on serial processing order of worker messages, so that we can guarantee that `setGeometryTileData` has completed by the time `startLayout` begins. We don't have this right now: it's possible for `setGeometryTileData` to get dispatched to one thread and `startLayout` to get dispatched to another and start executing before `setGeometryTileData`.

So I'm going to work on fixing #1471, which if implemented in the way I have in mind, would solve this problem as well.

jfirebaugh:
This is currently causing Android test failures, in the form of crashes with the following stack:

```
Build fingerprint: 'google/shamu/shamu:5.1.1/LMY48Y/2364368:user/release-keys'
pid: 12922, tid: 14188, name: Worker  >>> com.mapbox.mapboxsdk.testapp <<<
signal 11 (SIGSEGV), code 1 (SEGV_MAPERR), fault addr 0x0
Stack frame #00 pc 00129614  /data/app/com.mapbox.mapboxsdk.testapp-1/lib/arm/libmapbox-gl.so: Routine mbgl::TileWorker::parseLayer(mbgl::style::Layer const*) at /bitrise/src/build/android-arm-v7/Debug/../../../src/mbgl/tile/tile_worker.cpp:139
[...]

PR #6298, opened 2016-09-09
[core] Extract SymbolLayout from SymbolBucket

### LINKED ISSUE #2884: Rename SymbolBucket::addFeatures and SymbolBucket::placeFeatures
From https://github.com/mapbox/mapbox-gl-native/pull/2873#issuecomment-152588322:

There are actually three steps:

* **`parseFeatures`**: determines the text we need to display by replacing tokens and applying text transformations, and the glyphs we need
* **`addFeatures`**: determines the potential position of labels
* **`placeFeatures`**: actually places them and adds them to a buffer

`placeFeatures` is currently called as the [last step in `addFeatures`](https://github.com/mapbox/mapbox-gl-native/blob/1688-monitor-annotation-tiles/src/mbgl/renderer/symbol_bucket.cpp#L269), but there's no good reason to; we could separate that out and call the `addFeatures` without placing as soon as we have the symbols. Then we would have less work to do later. However, it means we'd have to introduce a third step beyond `pending`.

jfirebaugh:
I still need to debug this, but I want to get @ansis and @kkaefer 👀  on it as soon as possible.

This PR splits `SymbolBucket` in two. The new class `SymbolLayout` lives on the worker thread only and contains the persistent data needed for repeated placement. `SymbolBucket` contains the data generated during placement, and is transferred to the main thread for rendering.

This eliminates the risky sharing of `GeometryTile::buckets` between the main thread and worker thread during `TileWorker::redoPlacement`. `SymbolBucket::swapRenderData` is no longer necessary.

While here, rationalize the names of states a `SymbolLayout` may be in:

* `Pending`: Waiting for the necessary glyphs or icons to be available.
* `Prepared`: The potential positions of text and icons have been determined.
* `Placed`: The final positions have been determined, taking into account prior layers.

The related `SymbolLayout` methods reflect this terminology, which fixes #2884.

In `TileWorker`, all `SymbolLayouts` are now stored in a single vector. Each `SymbolLayout` knows what state it is in, and `TileWorker` can more straightforwardly determine how much progress it can make toward a final result. I think the advancement of a tile through the stages of processing is clearer now, making further progress towards #2851.

### DIFF
--- a/src/mbgl/renderer/symbol_bucket.cpp
+++ b/src/mbgl/renderer/symbol_bucket.cpp
@@ -40,7 +40,8 @@ void SymbolBucket::render(Painter& painter,
 }
 
 bool SymbolBucket::hasData() const {
-    return hasTextData() || hasIconData();
+    assert(false); // Should be calling SymbolLayout::hasSymbolInstances() instead.
+    return false;
 }
 
 bool SymbolBucket::hasTextData() const {
--- a/src/mbgl/renderer/symbol_bucket.cpp
+++ b/src/mbgl/renderer/symbol_bucket.cpp
@@ -1,86 +1,32 @@
 #include <mbgl/renderer/symbol_bucket.hpp>
-#include <mbgl/style/filter_evaluator.hpp>
-#include <mbgl/style/layers/symbol_layer.hpp>
-#include <mbgl/tile/geometry_tile_data.hpp>
-#include <mbgl/sprite/sprite_image.hpp>
-#include <mbgl/sprite/sprite_store.hpp>
-#include <mbgl/sprite/sprite_atlas.hpp>
-#include <mbgl/geometry/text_buffer.hpp>
-#include <mbgl/geometry/icon_buffer.hpp>
-#include <mbgl/geometry/glyph_atlas.hpp>
-#include <mbgl/geometry/anchor.hpp>
-#include <mbgl/text/get_anchors.hpp>
 #include <mbgl/renderer/painter.hpp>
-#include <mbgl/text/glyph_store.hpp>
-#include <mbgl/text/glyph_set.hpp>
-#include <mbgl/platform/log.hpp>
-#include <mbgl/text/collision_tile.hpp>
+#include <mbgl/style/layers/symbol_layer.hpp>
 #include <mbgl/shader/sdf_shader.hpp>
 #include <mbgl/shader/icon_shader.hpp>
 #include <mbgl/shader/collision_box_shader.hpp>
-
-#include <mbgl/util/utf.hpp>
-#include <mbgl/util/token.hpp>
-#include <mbgl/util/math.hpp>
-#include <mbgl/util/merge_lines.hpp>
-#include <mbgl/util/clip_lines.hpp>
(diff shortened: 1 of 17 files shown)


jfirebaugh:
I'd like to merge this today, so I can continue progress on fixing #6263. Reviews appreciated!

jfirebaugh:
@ansis @kkaefer Please review.

kkaefer:
Moving most of the code + changes in one commit makes this really hard to review.

kkaefer:
> The new class `SymbolLayout` lives on the worker thread only and contains the persistent data needed for repeated placement. `SymbolBucket` contains the data generated during placement, and is transferred to the main thread for rendering.

Really liking this separation. While reviewing, I didn't spot any logic errors. I think this is good to go.

kkaefer:
Here's a suggestion for reducing the number of `SymbolBuckets that are gratuitously created, then discarded:

```diff
diff --git a/src/mbgl/tile/tile_worker.cpp b/src/mbgl/tile/tile_worker.cpp
index f079847..53f6acb 100644
--- a/src/mbgl/tile/tile_worker.cpp
+++ b/src/mbgl/tile/tile_worker.cpp
@@ -164,9 +164,9 @@ TilePlacementResult TileWorker::redoPlacement(const PlacementConfig& config) {
 
     for (auto& symbolLayout : symbolLayouts) {
         symbolLayout->state = SymbolLayout::Placed;
-        std::unique_ptr<Bucket> bucket = symbolLayout->place(*result.collisionTile);
-        if (bucket->hasData() || symbolLayout->hasSymbolInstances()) {
-            result.buckets.emplace(symbolLayout->bucketName, std::move(bucket));
+        if (symbolLayout->hasSymbolInstances()) {
+            result.buckets.emplace(symbolLayout->bucketName,
+                                   symbolLayout->place(*result.collisionTile));
         }
     }
 
```

jfirebaugh:
Sorry for the poor diff. [Here](https://gist.github.com/jfirebaugh/582573a0b192752635eb8d26aa96daca) is a diff between symbol_bucket.cpp before and symbol_layout.cpp after, so you can see what changed in the moved code:

* Updated `#include` lists
* Changed constructor arguments
* Changed some variables to camel-case
* `needsDependencies` ⇢ `canPrepare` (and negated the sense of the return value)
* `addFeatures` ⇢ `prepare`
* `placeFeatures` ⇢ `place`, and it returns a bucket
* `addToDebugBuffers` gets passed `SymbolBucket&` rather than updating a member variable

src/mbgl/renderer/symbol_bucket.cpp
```diff
@@ -40,7 +40,8 @@ void SymbolBucket::render(Painter& painter,
 }
 
 bool SymbolBucket::hasData() const {
-    return hasTextData() || hasIconData();
+    assert(false); // Should be calling SymbolLayout::hasSymbolInstances() instead.
+    return false;
```
kkaefer: We could move `hasSymbolInstances()` to this function.

mourner:
@jfirebaugh nice! Does it make sense to split the two in GL JS as well?

PR #6301, opened 2016-09-09
v10 default styles

1ec5:
Upgraded from v9 default styles to v10 wherever the developer expects to get the latest and greatest, as well as in a couple tests where it may be beneficial to ensure that we can handle a two-digit version number in the style URL.

Added the [Traffic Day v2 and Traffic Night v2 styles](https://www.mapbox.com/blog/live-traffic-maps/), which are now documented as part of the styles API in [Mapbox API Documentation](https://www.mapbox.com/api-documentation/#styles).

This PR also partially reverses the decision in #4702 to require iOS and macOS developers to specify a style version number when calling one of MGLStyle’s style URL factory methods. The versioned methods are still there, but the unversioned methods are no longer deprecated. This change makes the iOS and macOS SDKs consistent with the Android and Qt SDKs, which never followed suit and kept their unversioned constants.

~~Hold until the v10 styles are live.~~

/cc @tobrun @friedbunny @tmpsantos @amyleew


tmpsantos:
@1ec5 thanks!

Just food for thought, not exactly related to this PR, but if a style change also upgrades the tile source, we need to think about a way of not rendering tiles on the offline database useless.

1ec5:
Can you elaborate? Is the concern that the user would have to redownload a region's tiles if the v10 Streets style uses a newer version of the Streets source (say, v8) than the v9 Streets style? Styles are somewhat tightly coupled to the sources they reference, so I'm not sure how we'd address that issue without adding custom tile migration logic for individual sources.

tmpsantos:
> Styles are somewhat tightly coupled to the sources they reference

Makes sense. Yeah, in this case, if you wanna keep the offline database, better not upgrade the style. Thanks!

amyleew:
@1ec5 - New `v10` styles and `v9` updates are now live. 

Please note all styles were not updated to a `v10` version, only Mapbox Streets, Mapbox Outdoors, Mapbox Satellite Streets now have `v10` versions. Please reference the [CHANGELOG](https://github.com/mapbox/mapbox-gl-private-styles/blob/master/CHANGELOG.md) for specifics.

zugaldia:
@1ec5 Thanks for updating the Android versions as well, is there any tail work required on Android once this PR gets merged?

1ec5:
v10 has been released for Streets, Outdoors, and Satellite Streets. Apparently the styles are versioned separately, so Light, Dark, and Satellite are still stuck on v9. This PR will need to be reworked to reflect that fact:

* [ ] Remove `mbgl::util::default_styles::currentVersion` and figure out what to do about code that uses it
* [ ] Tweak `MGLStyleDefaultVersion`’s documentation to say it’s the version of the default style, not “the default version of the suite of default styles”
* [ ] Hard-code version numbers in tests, making it much more likely that versions will get out of sync in the future

1ec5:
Not critical for v3.4.1, and no Mapbox-wide timetable for releasing the v10 styles. Moving off the v3.4.1 milestone.

platform/default/mbgl/util/default_styles.cpp
```diff
@@ -4,12 +4,12 @@ namespace mbgl {
...
-const DefaultStyle outdoors         = { "mapbox://styles/mapbox/outdoors-v9",          "Outdoors" };
-const DefaultStyle light            = { "mapbox://styles/mapbox/light-v9",             "Light" };
-const DefaultStyle dark             = { "mapbox://styles/mapbox/dark-v9",              "Dark" };
-const DefaultStyle satellite        = { "mapbox://styles/mapbox/satellite-v9",         "Satellite" };
-const DefaultStyle satelliteStreets = { "mapbox://styles/mapbox/satellite-streets-v9", "Satellite Streets" };
+const DefaultStyle streets          = { "mapbox://styles/mapbox/streets-v10",           "Streets" };
+const DefaultStyle outdoors         = { "mapbox://styles/mapbox/outdoors-v10",          "Outdoors" };
+const DefaultStyle light            = { "mapbox://styles/mapbox/light-v10",             "Light" };
+const DefaultStyle dark             = { "mapbox://styles/mapbox/dark-v10",              "Dark" };
+const DefaultStyle satellite        = { "mapbox://styles/mapbox/satellite-v10",         "Satellite" };
```
tmpsantos: We don't have a v10 for light, dark and satellite.

1ec5: I originally authored this PR before any of the v10 styles went public, so it wasn’t clear to me. https://github.com/mapbox/mapbox-gl-native/pull/6301#issuecomment-252031292 tracks other things that need to happen before this PR can land.

1ec5:
The Qt SDK’s (unversioned) style literals were updated downstream in [change 192085](https://codereview.qt-project.org/#/c/192085/).

1ec5:
In 98890a10c46283265272e49a6450bf408920a69f, I partially reversed the decision in #4702 to require the developer to specify a style version number when calling one of MGLStyle’s style URL factory methods. The versioned methods are still there, but the unversioned methods are no longer deprecated. This change makes the iOS and macOS SDKs consistent with the Android and Qt SDKs, which never followed suit and kept their unversioned constants.

To mitigate the risk of developers using the unversioned methods with the runtime styling API, I’ve added warnings in the documentation that spell out the risk. However, we may want to go a step further:

* We could log a warning to the console that provides the current style version or the full style URL when one of the style URL factory methods is called.
* We could append a query parameter like `?x-darwin-versioned=false` to the returned URL; MGLMapView could track the usage of such URLs and disable or booby-trap the runtime styling API. Once #6386 is implemented, we would rely on a flag on MGLStyle instead of an ad-hoc URL query parameter.

/cc @jfirebaugh

kkaefer:
> We could log a warning to the console that provides the current style version or the full style URL when one of the style URL factory methods is called.

Can we log that to the console for Debug builds only? I think disabling runtime styling for certain URLs could also lead to very weird behavior and increases the complexity.

1ec5:
> Can we log that to the console for Debug builds only? I think disabling runtime styling for certain URLs could also lead to very weird behavior and increases the complexity.

Yes, a log message would certainly be doable. If our goal is to alert about programmer error, though, maybe it should be present in release builds as well, but once per session, like [the warning about using floats in predicates](https://github.com/mapbox/mapbox-gl-native/blob/ea0e45b4019ea18c03718e67a3397202080a644a/platform/darwin/src/NSExpression%2BMGLAdditions.mm#L50-L53).

As far as disabling runtime styling for unversioned URLs, I think this is something that MGLMapView would be able to handle: `-[MGLMapView setStyleURL:]` would detect the `x-darwin-versioned` parameter and potentially strip it, and `-[MGLMapView style]` would remain `nil` even after `-[MGLMapViewDelegate mapView:styleDidFinishLoading:]` is called. But you’re right that a log message would be friendlier. Disabling the runtime styling API would be rather draconian, only remotely desirable if we find that everyone is using unversioned style URLs with the runtime styling API.

platform/ios/CHANGELOG.md
```diff
@@ -26,7 +28,12 @@ Mapbox welcomes participation and contributions from everyone. Please read [CONT
 ### Other changes
 
 * Xcode 8.0 or higher is now recommended for using this SDK. ([#8775](https://github.com/mapbox/mapbox-gl-native/pull/8775))
+<<<<<<< HEAD
```
kkaefer: Merge conflict

1ec5: Whoops, I missed that one – good catch! Fixed in 1d0e203f18ba315d2f5166d671624992335d3a7c.

1ec5:
> We could append a query parameter like `?x-darwin-versioned=false` to the returned URL; MGLMapView could track the usage of such URLs and disable or booby-trap the runtime styling API.

Tracking as possible tail work in #8971.

1ec5:
Cherry-picking into the release-ios-v3.6.0-android-v5.1.0 branch in #8972.

PR #6308, opened 2016-09-12
[core] using 32bits (instead of 8) for the stencil masks, fixes stencil mask overflows

### LINKED ISSUE #962: stencil mask overflow.
I've seen this message in Xcode debugger spew sometimes. Not sure if this is a real issue or not.

"stencil mask overflow"

rastapasta:
fixes #962

OpenGL's glStencilFunc [supports](https://www.opengl.org/sdk/docs/man/html/glStencilMask.xhtml) a stencial mask in form of a GLuint -> 32 bit long mask.

Using just 8 bits as before is limiting and leads to avoidable stencil overflow errors.

Cheers!


kkaefer:
While the function signature uses a 32 bit type, almost all GPUs only support an 8 bit buffer in practice, so setting more bits won't have any effect on the actual hardware.

PR #6309, opened 2016-09-12
[core] [android] - configurable API endpoint

tobrun:
This PR enables configuration of a custom URL base endpoint. 
Review c++: @jfirebaugh @tmpsantos 
Review java: @zugaldia 

Side notes vs GL-JS implementation:
 -  I went for using APIBaseURL over APIURL for readability 
 -  I initially mimicked the Config class concept but removed it after discussion with @tmpsantos

### DIFF
--- a/include/mbgl/util/constants.hpp
+++ b/include/mbgl/util/constants.hpp
@@ -45,6 +45,8 @@ constexpr UnitBezier DEFAULT_TRANSITION_EASE = { 0, 0, 0.25, 1 };
     
 constexpr int DEFAULT_RATE_LIMIT_TIMEOUT = 5;
 
+constexpr const char* API_BASE_URL = "https://api.mapbox.com";
+    
 } // namespace util
 
 namespace debug {
--- a/platform/android/MapboxGLAndroidSDK/src/main/java/com/mapbox/mapboxsdk/http/HTTPRequest.java
+++ b/platform/android/MapboxGLAndroidSDK/src/main/java/com/mapbox/mapboxsdk/http/HTTPRequest.java
@@ -56,7 +56,7 @@ class HTTPRequest implements Callback {
 
             HttpUrl httpUrl = HttpUrl.parse(resourceUrl);
             final String host = httpUrl.host().toLowerCase(MapboxConstants.MAPBOX_LOCALE);
-            if (host.equals("mapbox.com") || host.endsWith(".mapbox.com")) {
+            if (host.equals("mapbox.com") || host.endsWith(".mapbox.com") || host.equals("mapbox.cn") || host.endsWith(".mapbox.cn")) {
                 if (httpUrl.querySize() == 0) {
                     resourceUrl = resourceUrl + "?";
                 } else {
--- a/platform/android/MapboxGLAndroidSDK/src/main/java/com/mapbox/mapboxsdk/maps/MapboxMapOptions.java
+++ b/platform/android/MapboxGLAndroidSDK/src/main/java/com/mapbox/mapboxsdk/maps/MapboxMapOptions.java
@@ -7,15 +7,12 @@ import android.graphics.Canvas;
 import android.graphics.Color;
 import android.graphics.drawable.BitmapDrawable;
 import android.graphics.drawable.Drawable;
-import android.graphics.drawable.GradientDrawable;
-import android.graphics.drawable.VectorDrawable;
 import android.os.Parcel;
 import android.os.Parcelable;
 import android.support.annotation.ColorInt;
 import android.support.annotation.IntRange;
 import android.support.annotation.NonNull;
 import android.support.annotation.Nullable;
-import android.support.graphics.drawable.VectorDrawableCompat;
 import android.support.v4.content.ContextCompat;
 import android.util.AttributeSet;
 import android.view.Gravity;
@@ -79,6 +76,7 @@ public class MapboxMapOptions implements Parcelable {
     private int myLocationAccuracyTintColor;
     private int myLocationAccuracyAlpha;
 
+    private String apiBaseUrl;
     private String style;
     @Deprecated
     private String accessToken;
@@ -123,12 +121,12 @@ public class MapboxMapOptions implements Parcelable {
         }
 
         Bitmap foregroundBearingBitmap = in.readParcelable(getClass().getClassLoader());
-        if(foregroundBearingBitmap!=null) {
+        if (foregroundBearingBitmap != null) {
             myLocationForegroundBearingDrawable = new BitmapDrawable(foregroundBearingBitmap);
         }
 
         Bitmap backgroundBitmap = in.readParcelable(getClass().getClassLoader());
-        if(backgroundBitmap!=null){
+        if (backgroundBitmap != null) {
             myLocationBackgroundDrawable = new BitmapDrawable(backgroundBitmap);
         }
 
@@ -140,6 +138,7 @@ public class MapboxMapOptions implements Parcelable {
 
         style = in.readString();
         accessToken = in.readString();
+        apiBaseUrl = in.readString();
     }
 
     public static Bitmap getBitmapFromDrawable(Drawable drawable) {
@@ -172,6 +171,7 @@ public class MapboxMapOptions implements Parcelable {
 
             mapboxMapOptions.accessToken(typedArray.getString(R.styleable.MapView_access_token));
             mapboxMapOptions.styleUrl(typedArray.getString(R.styleable.MapView_style_url));
+            mapboxMapOptions.apiBaseUrl(typedArray.getString(R.styleable.MapView_api_base_url));
 
             mapboxMapOptions.zoomGesturesEnabled(typedArray.getBoolean(R.styleable.MapView_zoom_enabled, true));
             mapboxMapOptions.scrollGesturesEnabled(typedArray.getBoolean(R.styleable.MapView_scroll_enabled, true));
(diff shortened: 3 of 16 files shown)


platform/default/online_file_source.cpp
```diff
@@ -152,30 +152,34 @@ OnlineFileSource::~OnlineFileSource() = default;
 
 std::unique_ptr<AsyncRequest> OnlineFileSource::request(const Resource& resource, Callback callback) {
     Resource res = resource;
+    
+    if(apiBaseURL.empty()){
```
jfirebaugh: Set the default value to `mbgl::util::API_BASE_URL` instead of checking for the empty string.

include/mbgl/util/constants.hpp
```diff
@@ -43,6 +43,8 @@ constexpr Seconds CLOCK_SKEW_RETRY_TIMEOUT { 30 };
 
 constexpr UnitBezier DEFAULT_TRANSITION_EASE = { 0, 0, 0.25, 1 };
 
+constexpr const char* API_BASE_URL = "https://api.mapbox.com/";
```
jfirebaugh: Let's remove the trailing slash from this URL. That matches GL JS and is more correct -- a trailing slash is a path, and in this case the point is that this is a base URL to which a path will be appended.

src/mbgl/util/mapbox.cpp
```diff
@@ -35,19 +34,19 @@ std::vector<std::string> getMapboxURLPathname(const std::string& url) {
...
     if (!isMapboxURL(url)) {
         return url;
     }
 
     if (accessToken.empty()) {
         throw std::runtime_error("You must provide a Mapbox API access token for Mapbox tile sources");
     }
-
+    
     return baseURL + "v4/" + url.substr(protocol.length()) + ".json?access_token=" + accessToken + "&secure";
```
jfirebaugh: After removing the slash from the default value, add a slash here (`"/v4/"`) and in each of the other normalization functions.

tobrun:
Thanks @jfirebaugh for 👀 , the remarks have been fixed in ad5f363. 
I will now look into resolving this for the other bindings.

platform/android/MapboxGLAndroidSDK/src/main/java/com/mapbox/mapboxsdk/maps/MapboxMapOptions.java
```diff
@@ -957,6 +978,8 @@ public boolean equals(Object o) {
         if (!Arrays.equals(myLocationBackgroundPadding, options.myLocationBackgroundPadding))
             return false;
         if (style != null ? !style.equals(options.style) : options.style != null) return false;
+        if (apiBaseUrl != null ? !apiBaseUrl.equals(options.apiBaseUrl) : options.apiBaseUrl != null)
```
zugaldia: @tobrun This is very hard to read, could you split it up?

platform/android/MapboxGLAndroidSDK/src/main/java/com/mapbox/mapboxsdk/http/HTTPRequest.java
```diff
@@ -56,15 +56,15 @@ private HTTPRequest(long nativePtr, String resourceUrl, String userAgent, String
 
             HttpUrl httpUrl = HttpUrl.parse(resourceUrl);
             final String host = httpUrl.host().toLowerCase(MapboxConstants.MAPBOX_LOCALE);
-            if (host.equals("mapbox.com") || host.endsWith(".mapbox.com")) {
+            if (host.contains("mapbox.")) {
```
zugaldia: @tobrun Not sure about this one, [from the docs](https://square.github.io/okhttp/3.x/okhttp/okhttp3/HttpUrl.html) the host can be something like `www.google.com` so if a third-party names their server `mapbox.mycompany.com` this will be `true` too.

platform/android/MapboxGLAndroidSDK/src/main/java/com/mapbox/mapboxsdk/http/HTTPRequest.java
```diff
@@ -56,15 +56,15 @@ private HTTPRequest(long nativePtr, String resourceUrl, String userAgent, String
...
                 if (httpUrl.querySize() == 0) {
                     resourceUrl = resourceUrl + "?";
                 } else {
                     resourceUrl = resourceUrl + "&";
                 }
                 resourceUrl = resourceUrl + "events=true";
             }
-
+
```
tmpsantos: nit

platform/android/src/jni.cpp
```diff
@@ -424,6 +424,13 @@ jni::jobject* nativeGetClasses(JNIEnv *env, jni::jobject* obj, jlong nativeMapVi
     return std_vector_string_to_jobject(env, nativeMapView->getMap().getClasses());
 }
 
+void nativeSetAPIBaseURL(JNIEnv *env, jni::jobject* obj, jlong nativeMapViewPtr, jni::jstring* url){
```
tmpsantos: `) {`

platform/default/default_file_source.cpp
```diff
@@ -28,6 +28,10 @@ class DefaultFileSource::Impl {
     Impl(const std::string& cachePath, uint64_t maximumCacheSize)
         : offlineDatabase(cachePath, maximumCacheSize) {
     }
+    
+    void setAPIBaseURL(const std::string& url){
```
tmpsantos: `) {`

platform/default/online_file_source.cpp
```diff
@@ -152,30 +152,30 @@ OnlineFileSource::~OnlineFileSource() = default;
 
 std::unique_ptr<AsyncRequest> OnlineFileSource::request(const Resource& resource, Callback callback) {
     Resource res = resource;
-
+
```
tmpsantos: nit

src/mbgl/util/mapbox.cpp
```diff
@@ -35,19 +34,19 @@ std::vector<std::string> getMapboxURLPathname(const std::string& url) {
...
     if (!isMapboxURL(url)) {
         return url;
     }
 
     if (accessToken.empty()) {
         throw std::runtime_error("You must provide a Mapbox API access token for Mapbox tile sources");
     }
-
-    return baseURL + "v4/" + url.substr(protocol.length()) + ".json?access_token=" + accessToken + "&secure";
+
```
tmpsantos: nit

tmpsantos:
This PR is missing unit tests for a new core API. ;-)

platform/android/MapboxGLAndroidSDK/src/main/java/com/mapbox/mapboxsdk/maps/MapboxMapOptions.java
tobrun: This is generated code from intelij or actually it's copied code from intelij. The correct way should be to regenerate the hashcode/equals method.. Will pick this up with other remarks but I'm not taking readability into account with generated code.

platform/android/MapboxGLAndroidSDK/src/main/java/com/mapbox/mapboxsdk/http/HTTPRequest.java
tobrun: After convo, we need to explicitly set the different supported domains.

platform/android/MapboxGLAndroidSDK/src/main/java/com/mapbox/mapboxsdk/maps/MapboxMapOptions.java
zugaldia: @tobrun Sorry, I missed this was generated code. Scratch my comment.

tobrun:
@tmpsantos I fixed up the comments and added a tests in https://github.com/mapbox/mapbox-gl-native/pull/6309/commits/4db1f6d9ab3bf8b84ec6b8baeb139519946ee558.

1ec5:
Tracking the necessary iOS/macOS SDK changes in #6346.

tobrun:
Need to fix the tests for glfw, the tests are missing runloop

tobrun:
@tmpsantos thanks for the hint on RunLoop. 
This is now implemented in 5825a6c and the tests are passing 😎

test/storage/online_file_source.cpp
```diff
@@ -409,10 +409,9 @@ TEST(OnlineFileSource, ChangeAPIBaseURL){
     util::RunLoop loop;
     OnlineFileSource fs;
     
-    loop.run();
```
tmpsantos: Would be great if you could split the PR in two or three commits:

- One adding the API to core
- One adding the API to android
- One adding the tests

This makes reviewing a lot easier, since reviewing is usually done "per commit". You can do this next time, but for now I'm fine squashing everything into a single commit. The last two are fixing something on the first one, so they don't really add value to the PR.

test/util/mapbox.cpp
```diff
@@ -10,85 +11,85 @@ using namespace mbgl;
 TEST(Mapbox, SourceURL) {
     EXPECT_EQ(
         "https://api.mapbox.com/v4/user.map.json?access_token=key&secure",
-        mbgl::util::mapbox::normalizeSourceURL("mapbox://user.map", "key"));
+              mbgl::util::mapbox::normalizeSourceURL(util::API_BASE_URL, "mapbox://user.map", "key"));
```
tmpsantos: Nit, extra tab

PR #6310, opened 2016-09-12
Move unused icon burning script

friedbunny:
Removes an icon burning script that appears to be unused. If this script is indeed still used, we should move it to `platform/android/someplace`.

/cc @bleege @tobrun

### DIFF
--- a/platform/android/MapboxGLAndroidSDKTestApp/update_icons.sh
+++ b/platform/android/MapboxGLAndroidSDKTestApp/update_icons.sh
@@ -1,7 +1,7 @@
 #!/bin/bash
 
 if [[ -z `which iconoblast` ]]; then
-    echo "Requires https://github.com/mapbox/iconoblast!"
+    echo "Requires https://github.com/mapbox/iconoblast"
     exit 1
 fi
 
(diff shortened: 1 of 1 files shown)


bleege:
@friedbunny Let's leave the Icon Burning in there for Android too.  While @cammace has dramatically upped the quality of the icon I really like the idea of being able to visibly track the state of the TestApp build by SHA like iOS does.

friedbunny:
@bleege OK. Where should it be moved to?

friedbunny:
Incidentally, we no longer use icon burning on iOS (and I’m not sure we ever did in this repo?).

friedbunny:
Updated this to move the script to `platform/android/MapboxGLAndroidSDKTestApp/update_icons.sh`.

friedbunny:
@bleege Please review/approve. 🙇

tobrun:
@bleege
 if we are not using it and we are not planning to use it, let's remove it? 
cc @mapbox/android

incanus:
We did use this for a time, before we had other internal test apps, but I am 👍 with getting it out of this repo.

bleege:
> but I am 👍 with getting it out of this repo. 

Same here.  Let's do it.

BTW, I'm noticing a brand new Merge UI requiring a "Review" before merging can technically happen (see image below).  ¯\_(ツ)_/¯  I'll see if I can officially bless this in a way to make our Github overlords happy.

screen shot 2016-09-16 at 10 22 35 am

PR #6311, opened 2016-09-12
Move wrap.cpp test to correct directory 

tobrun:
Follow up from #6289 where I wrongly placed the test in the root directory.

cc @friedbunny


friedbunny:
Thanks. <3

PR #6313, opened 2016-09-12
Use content inset to calculate edge padding

### LINKED ISSUE #6257: Map not centered over bottom-aligned user puck
**Platform:** iOS
**Mapbox SDK version:** 3.4.0-alpha.4 (51b2033)

It's looking like we're not putting the user puck on the actual location of the user. For example:

Has me down an alley off 15th street, but actually I'm standing right at the corner of 14th and S.

It's very easy to generate this behavior in a test app, so I'm happy to provide more screenshots.

As you move around and the POV changes the puck swings around and has a very disorienting behavior.

boundsj:
This reverts a previous change that used only the `contentInset` when setting the center coordinate of the map view. Instead of `contentInset`, `edgePaddingForFollowing` is now used again. This fixes
the regression introduced in https://github.com/mapbox/mapbox-gl-native/pull/6216 that caused the user dot annotation to be misplaced when the map was configured to show the annotation in any place other than in the center of the map.

In addition, `edgePaddingForFollowing` is updated to account for `contentInset` when it calculates its `UIEdgeInsets`. This fixes a bug where the user dot would move to the center of the content frame
(when it was adjusted for content insets) but the map would not scroll along with it yielding the effect of the user dot being misplaced.

Fixes #6257 
Reverts #6216

cc @1ec5

### DIFF
--- a/platform/ios/src/MGLMapView.mm
+++ b/platform/ios/src/MGLMapView.mm
@@ -4085,7 +4085,7 @@ public:
 - (void)didUpdateLocationIncrementallyAnimated:(BOOL)animated
 {
     [self _setCenterCoordinate:self.userLocation.location.coordinate
-                   edgePadding:self.contentInset
+                   edgePadding:self.edgePaddingForFollowing
                      zoomLevel:self.zoomLevel
                      direction:self.directionByFollowingWithCourse
                       duration:animated ? MGLUserLocationAnimationDuration : 0
@@ -4112,7 +4112,7 @@ public:
     
     __weak MGLMapView *weakSelf = self;
     [self _flyToCamera:camera
-           edgePadding:self.contentInset
+           edgePadding:self.edgePaddingForFollowing
           withDuration:animated ? -1 : 0
           peakAltitude:-1
      completionHandler:^{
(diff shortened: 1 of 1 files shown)


platform/ios/src/MGLMapView.mm
```diff
@@ -4173,8 +4173,14 @@ - (UIEdgeInsets)edgePaddingForFollowing
     CGRect boundsAroundCorrectPoint = CGRectOffset(bounds,
                                                    correctPoint.x - CGRectGetMidX(bounds),
                                                    correctPoint.y - CGRectGetMidY(bounds));
-    return UIEdgeInsetsMake(CGRectGetMinY(boundsAroundCorrectPoint) - CGRectGetMinY(bounds), 0,
-                            CGRectGetMaxY(bounds) - CGRectGetMaxY(boundsAroundCorrectPoint), 0);
+    UIEdgeInsets insets = UIEdgeInsetsMake(CGRectGetMinY(boundsAroundCorrectPoint) - CGRectGetMinY(bounds), 0, CGRectGetMaxY(bounds) - CGRectGetMaxY(boundsAroundCorrectPoint), 0);
+    
+    insets.top += self.contentInset.top;
```
1ec5: These lines could be folded into the line above, but it’s your call as to which would be easier to reason about.

boundsj:
Here is a visual of this change in effect.

screen shot 2016-09-12 at 4 06 49 pm

boundsj:
Note: the above screenshots were created using a map view controller layout where "adjusts scroll view insets" was set to false.

PR #6314, opened 2016-09-13
[android] - do not update marker while creating

### LINKED ISSUE #5915: Markers can get out of sync when added/removed regularly
I'm writing code to add/show markers on my map based on a zoom threshold. e.g. one set of markers is shown until zoom > 2.0, then show the other set.

I've coded up a demo project with essentially what I'm doing:

https://github.com/sddamico/mapbox-marker-bug

The issue seems to be that the markers that are "removed" cannot be removed again when they are added a second time. I've provided an example gif as well:

zoom-bug

**Platform:** Android
**Mapbox SDK version:** 4.2.0-SNAPSHOT

### Steps to trigger behavior

 1. Run example app above (after inserting own access token into start call)
 2. Zoom in and out

### Expected behavior
Based on zoom level, one marker is shown and the other is not

### Actual behavior
Both markers end up showing

### LINKED ISSUE #6208: Visual artifact of Marker still rendered on map when removed for a 2nd time (Android)
I have a toggle feature to show/hide groups of markers on a map (iterating each list and making add/remove calls to MapboxMap). I have noticed that I can add and remove markers once without issue however after adding markers the 2nd time then the artifact of the marker is rendered on the map forevermore (even though you cannot click on the marker to get an infowindow etc).

**Platform:**
Android 6.0.1 (Nexus 5X) 

**Mapbox SDK version:**
4.2.0 beta2

### Steps to trigger behavior

```
Marker marker = mMapboxMap.addMarker(getMarkerOpt());
mMapboxMap.removeMarker(marker);
Marker marker2 = mMapboxMap.addMarker(getMarkerOpt());
mMapboxMap.removeMarker(marker2);
```

```
private MarkerOptions getMarkerOpt() {
return new MarkerOptions()
                                    .position(latLng)
                                    .title(name)
                                    .snippet(type + " " + subcat)
                                    .icon(poi.icon);
}
```

### Expected behavior
I would expect the marker icon to be removed from the map each time I call remove marker

### Actual behavior
Only first call to remove marker actually removes the rendered marker from the map. 
N.B. The act of adding / removing the marker technically works i.e. if I call remove for a 2nd time then I can no longer click on the marker to show the info window

tobrun:
closes #6208 
closes #5915

This PR removes unnecessary `updateAnnotation` calls. 
This was resulting in having ghost annotations. 
This PR also adds a test activity to prevent regression.

Review @ivovandongen / @zugaldia 
cc @jfirebaugh

### DIFF
--- a/src/mbgl/annotation/annotation_manager.cpp
+++ b/src/mbgl/annotation/annotation_manager.cpp
@@ -40,7 +40,9 @@ AnnotationManager::~AnnotationManager() = default;
 
 AnnotationID AnnotationManager::addAnnotation(const Annotation& annotation, const uint8_t maxZoom) {
     AnnotationID id = nextID++;
-    updateAnnotation(id, annotation, maxZoom);
+    Annotation::visit(annotation, [&] (const auto& annotation_) {
+        this->add(id, annotation_, maxZoom);
+    });
     return id;
 }
 
@@ -57,6 +59,8 @@ void AnnotationManager::removeAnnotation(const AnnotationID& id) {
     } else if (shapeAnnotations.find(id) != shapeAnnotations.end()) {
         obsoleteShapeAnnotationLayers.insert(shapeAnnotations.at(id)->layerID);
         shapeAnnotations.erase(id);
+    } else {
+        assert(false); // Should never happen
     }
 }
 
@@ -85,13 +89,14 @@ void AnnotationManager::add(const AnnotationID& id, const StyleSourcedAnnotation
 }
 
 Update AnnotationManager::update(const AnnotationID& id, const SymbolAnnotation& annotation, const uint8_t maxZoom) {
+    Update result = Update::Nothing;
+    
     auto it = symbolAnnotations.find(id);
     if (it == symbolAnnotations.end()) {
-        removeAndAdd(id, annotation, maxZoom);
-        return Update::AnnotationData | Update::AnnotationStyle;
+        assert(false); // Attempt to update a non-existent symbol annotation
+        return result;
     }
 
-    Update result = Update::Nothing;
     const SymbolAnnotation& existing = it->second->annotation;
 
     if (existing.geometry != annotation.geometry) {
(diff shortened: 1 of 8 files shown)


jfirebaugh:
As a gl-core exercise, how about also adding asserts to `AnnotationManager` that trigger on attempts to update or remove a non-existent annotation? That probably would have caught this issue sooner.

tobrun:
@jfirebaugh I added asserts in 722053c.

For updating I'm checking if a key is available:
 -  `assert(symbolAnnotations.count(id));`

For removing I'm failing when a key is not available:
 -  `assert(false)` after an else check

jfirebaugh:
Asserts look good. However `addAnnotation` will require some refactoring to avoid triggering them. Instead of having it call `updateAnnotation`, try:

```
    return Annotation::visit(annotation, [&] (const auto& annotation_) {
        return this->add(id, annotation_, maxZoom);
    });
```

And then update `AnnotationManager::update` to make it clear the `it == symbolAnnotations.end()` case is not expected.

tobrun:
Adapted the addAnnotation to:

```cpp
AnnotationID AnnotationManager::addAnnotation(const Annotation& annotation, const uint8_t maxZoom) {
    AnnotationID id = nextID++;
    Annotation::visit(annotation, [&] (const auto& annotation_) {
        this->add(id, annotation_, maxZoom);
    });
    return id;
}
```
The return type is AnnotationID vs Update in https://github.com/mapbox/mapbox-gl-native/pull/6314#issuecomment-246879845. Is this a valid construction?

src/mbgl/annotation/annotation_manager.cpp
```diff
@@ -72,6 +76,7 @@ void AnnotationManager::add(const AnnotationID& id, const StyleSourcedAnnotation
 }
 
 Update AnnotationManager::update(const AnnotationID& id, const SymbolAnnotation& annotation, const uint8_t maxZoom) {
+    assert(symbolAnnotations.count(id));
```
jfirebaugh: `AnnotationManager::update` used to be expected to have the ability to add new annotations, but now it doesn't. Let's make this clear by changing it to:

```C++
 Update AnnotationManager::update(const AnnotationID& id, const SymbolAnnotation& annotation, const uint8_t maxZoom) {
     auto it = symbolAnnotations.find(id);
     if (it == symbolAnnotations.end()) {
         assert(false); // Attempt to update a non-existent symbol annotation
         return Update::Nothing;
     }
     ...
```

ivovandongen:
Besides CI failing, Android side looks good.

tobrun:
@jfirebaugh assertions have been changed as requested in 027f607. 
I also took the time to add some unit tests around this in 0a98ca4.

jfirebaugh:
I appreciate the thoroughness! However we don't usually test assertions -- assertions are themselves sort of a runtime test, tend to change frequently, and their exact message is platform specific, which I think is why CI is failing. Let's go ahead and leave those tests out.

tobrun:
> However we don't usually test assertions 

👍 GTK, removed that commit and cleaned the others.

while writing the tests I noticed that we only have render test for annotations. Do we want to be able to test the business logic? If yes, what is your opinion on adding code to be able unit test them? (we currently don't expose a getters for symbol/shape annotations).

jfirebaugh:
Is there functionality we could test that isn't covered by the rendering tests, and could we test it without adding accessors? In general, the rendering tests work well for us because they test fairly stable APIs and don't depend on adding accessors just for the sake of testing.

tobrun:
We are currently testing the overall feature through UI tests. While these tests detect a feature being broken, there are scenarios that changes to business logic would remain undetected. 

An example for this in this PR. We changed the internal logic of annotation manager but these changes did not reflect in any core tests failing. The java tests did however [flagged](https://www.bitrise.io/build/3e6606401ed2b475) some changes. 

For me, coming from another language and not being confident about changes introduced, writing unit tests help me validate my assumptions and help me remove the fear of changing code. Those are my 2 cents on the topic. I know this fits into a larger picture discussion about public API design and how you apply testing to it. It also comes close to the mistake of testing framework internals versus testing our own code.

jfirebaugh:
> We changed the internal logic of annotation manager but these changes did not reflect in any core tests failing.

They did, actually -- 722053c61264e5998506c8e47d1685071a03bbec was failing in CI before the additional changes to `AnnotationManager`.

> The java tests did however flagged some changes.

Can you say more about what exactly they flagged?

tobrun:
> Can you say more about what exactly they flagged?

They flagged misuse in the test code.. they did not flag the business logic change..

For me, testing is validating if the code works and thus also capturing the edge cases in tests. What we have now is that these cases are now encapsulated by doing runtime tests (assertions). 

Without extra 👀  I would have written:
 - addAnnotationTest
 - addRemoveAnnotationTest
 - updateAnnotationTest
 - updateNotExisitingAnnotationTest - expect assert
 - ...

I'm taking notes here and seeing that it can be done differently than what I'm used to. A benefit from current approach is that they test the feature as a black box and these will sustain change better than my white box testing approach.

The only remark is that I feel I'm still required to run the bindings to verify my changes. While with the white box testing approach I would feel confident doing this without.

Issue #6236, opened 2016-09-01
Clarify MGLGeoJSONSource option units


boundsj:
In https://github.com/mapbox/mapbox-gl-native/pull/6217#issuecomment-243697979 we discussed some possibilities to make the documentation for VT and cluster options clearer for native mobile developers. This issue is to continue that discussion and implement improved documentation since #6217 itself is functional and has good enough documentation.

cc @1ec5 @mourner @frederoni

boundsj:
Ref https://github.com/mapbox/mapbox-gl-js/issues/2985

1ec5:
Duplicate of #5960?

boundsj:
I'm happy with #5960, just wanted to capture the specifics you brought up about MGLGeoJSONSource and absolve #5962. Closing!

boundsj:
Actually going to keep this open and block the beta with it so we can have the discussion about how to make these options mean something.

1ec5:
To recap:

> Thinking ahead to when we document these options: `buffer` and `clusterRadius` will unfortunately be difficult APIs for developers to reason about. They’re expressed in units of “⅟₅₁₂ of a tile”, which don’t correspond to any fixed real-world or screen distances, yet we don’t offer any way for the developer to find out the size of the tile at a given zoom level.

> Fortunately, I don’t think this needs to be one of those “guess-and-check” APIs that we try to avoid exposing in the SDK. Rather than the formulation used in the style specification documentation, I think we should talk about the buffer and cluster radius being expressed in screen points (which is true at integral zoom levels), with a caveat that they may be scaled up to 2× between integral zoom levels. (The offset threshold implemented in #1843 is only for raster and video sources, so it won’t affect clustering.)

> @mourner, a couple questions for you:

> 1. The style specification talks about “⅟₅₁₂ of a tile” units. I assume that means a tile is 512 of these units tall and 512 of these units wide, not that there are 512 to a tile, correct? If so, that should be an easy documentation fix.
> 1. On Retina displays, do these ⅟₅₁₂-of-a-tile units remain ⅟₅₁₂-of-a-tile units, or do they become ⅟₁₀₂₄-of-a-tile units?
[...]

PR #6316, opened 2016-09-13
Add tile template option to raster and vector sources

### LINKED ISSUE #6245: MGLRasterSource and MGLVectorSource tile sets
MGLRasterSource needs support for the `tiles`, `minzoom`, and `maxzoom` style specification properties. We’ll need to add `tileURLTemplates`, `minimumZoomLevel`, and `maximumZoomLevel` properties to MGLRaster(Vector)Source, which we’ll use to create an `mbgl::style::TileSet` to pass into the `mbgl::style::RasterSource` constructor.

/cc @frederoni @boundsj

boundsj:
- [x] introduce MGLTileSet
- [x] add tileset initializer to MGLVectorSource
- [x] test MGLVectorSource in iosapp
- [x] add tileset initializer to MGLRasterSource 
- [x] test MGLRasterSource in iosapp
- [x] unit tests
- [x] documentation updates

Fixes #6245

### DIFF
--- a/platform/darwin/src/MGLRasterSource.h
+++ b/platform/darwin/src/MGLRasterSource.h
@@ -1,19 +1,20 @@
 #import "MGLSource.h"
 #import "MGLTypes.h"
 
+@class MGLTileSet;
+
 NS_ASSUME_NONNULL_BEGIN
 
 @interface MGLRasterSource : MGLSource
 
 @property (nonatomic, readonly, copy) NSURL *URL;
 @property (nonatomic, readonly, assign) CGFloat tileSize;
-@property (nonatomic, readonly, copy) NS_ARRAY_OF(NSString *) *tileURLTemplates;
-@property (nonatomic, readonly) uint8_t minimumZoomLevel;
-@property (nonatomic, readonly) uint8_t maximumZoomLevel;
+@property (nonatomic, readonly, nullable) MGLTileSet *tileSet;
 
 - (instancetype)initWithSourceIdentifier:(NSString *)sourceIdentifier URL:(NSURL *)url tileSize:(CGFloat)tileSize;
 
-- (instancetype)initWithSourceIdentifier:(NSString *)sourceIdentifier tileURLTemplates:(NS_ARRAY_OF(NSString *) *)tileURLTemplates minimumZoomLevel:(NSUInteger)minimumZoomLevel maximumZoomLevel:(NSUInteger)maximumZoomLevel;
+- (instancetype)initWithSourceIdentifier:(NSString *)sourceIdentifier tileSize:(CGFloat)tileSize tileSet:(MGLTileSet *)tileSet;
+
 @end
 
 NS_ASSUME_NONNULL_END
--- a/platform/darwin/src/MGLRasterSource.h
+++ b/platform/darwin/src/MGLRasterSource.h
@@ -1,10 +1,19 @@
 #import "MGLSource.h"
+#import "MGLTypes.h"
+
+NS_ASSUME_NONNULL_BEGIN
 
 @interface MGLRasterSource : MGLSource
 
 @property (nonatomic, readonly, copy) NSURL *URL;
 @property (nonatomic, readonly, assign) CGFloat tileSize;
+@property (nonatomic, readonly, copy) NS_ARRAY_OF(NSString *) *tileURLTemplates;
+@property (nonatomic, readonly) uint8_t minimumZoomLevel;
+@property (nonatomic, readonly) uint8_t maximumZoomLevel;
 
 - (instancetype)initWithSourceIdentifier:(NSString *)sourceIdentifier URL:(NSURL *)url tileSize:(CGFloat)tileSize;
 
+- (instancetype)initWithSourceIdentifier:(NSString *)sourceIdentifier tileURLTemplates:(NS_ARRAY_OF(NSString *) *)tileURLTemplates minimumZoomLevel:(NSUInteger)minimumZoomLevel maximumZoomLevel:(NSUInteger)maximumZoomLevel;
 @end
+
+NS_ASSUME_NONNULL_END
--- a/platform/darwin/src/MGLRasterSource.h
+++ b/platform/darwin/src/MGLRasterSource.h
@@ -8,7 +8,7 @@ NS_ASSUME_NONNULL_BEGIN
 @interface MGLRasterSource : MGLSource
 
 @property (nonatomic, readonly, copy) NSURL *URL;
-@property (nonatomic, readonly, assign) CGFloat tileSize;
+@property (nonatomic, readonly, assign) NSUInteger tileSize;
 @property (nonatomic, readonly, nullable) MGLTileSet *tileSet;
 
 - (instancetype)initWithSourceIdentifier:(NSString *)sourceIdentifier URL:(NSURL *)url tileSize:(CGFloat)tileSize;
--- a/platform/darwin/src/MGLRasterSource.mm
+++ b/platform/darwin/src/MGLRasterSource.mm
@@ -6,7 +6,8 @@
 
 @implementation MGLRasterSource
 
-- (instancetype)initWithSourceIdentifier:(NSString *)sourceIdentifier URL:(NSURL *)url tileSize:(CGFloat)tileSize {
+- (instancetype)initWithSourceIdentifier:(NSString *)sourceIdentifier URL:(NSURL *)url tileSize:(CGFloat)tileSize
+{
     if (self = [super initWithSourceIdentifier:sourceIdentifier]) {
         _URL = url;
         _tileSize = tileSize;
@@ -14,10 +15,39 @@
     return self;
 }
 
-- (std::unique_ptr<mbgl::style::Source>)mbglSource {
-    auto source = std::make_unique<mbgl::style::RasterSource>(self.sourceIdentifier.UTF8String,
-                                                              self.URL.absoluteString.UTF8String,
-                                                              uint16_t(self.tileSize));
+- (instancetype)initWithSourceIdentifier:(NSString *)sourceIdentifier tileURLTemplates:(NS_ARRAY_OF(NSString *) *)tileURLTemplates minimumZoomLevel:(NSUInteger)minimumZoomLevel maximumZoomLevel:(NSUInteger)maximumZoomLevel
+{
+    if (self = [super initWithSourceIdentifier:sourceIdentifier])
+    {
+        _tileURLTemplates = tileURLTemplates;
+        _minimumZoomLevel = minimumZoomLevel;
+        _maximumZoomLevel = maximumZoomLevel;
+    }
+    return self;
+}
+
+- (std::unique_ptr<mbgl::style::Source>)mbglSource
+{
+    std::unique_ptr<mbgl::style::RasterSource> source;
+    
+    if (self.URL)
+    {
+        source = std::make_unique<mbgl::style::RasterSource>(self.sourceIdentifier.UTF8String,
+                                                             self.URL.absoluteString.UTF8String,
+                                                             uint16_t(self.tileSize));
+    }
+    else
+    {
+        mbgl::Tileset tileSource;
+        tileSource.zoomRange = mbgl::Range<uint8_t>(self.minimumZoomLevel, self.maximumZoomLevel);
+        for (NSString *tileURLTemplate in self.tileURLTemplates)
+        {
+            tileSource.tiles.push_back(tileURLTemplate.UTF8String);
+        }
+        source = std::make_unique<mbgl::style::RasterSource>(self.sourceIdentifier.UTF8String, tileSource, uint16_t(self.tileSize));
+        
+    }
+    
     return std::move(source);
 }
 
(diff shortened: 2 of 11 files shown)


platform/darwin/src/MGLRasterSource.h
```diff
@@ -1,10 +1,19 @@
...
+#import "MGLTypes.h"
+
+NS_ASSUME_NONNULL_BEGIN
 
 @interface MGLRasterSource : MGLSource
 
 @property (nonatomic, readonly, copy) NSURL *URL;
 @property (nonatomic, readonly, assign) CGFloat tileSize;
+@property (nonatomic, readonly, copy) NS_ARRAY_OF(NSString *) *tileURLTemplates;
+@property (nonatomic, readonly) uint8_t minimumZoomLevel;
```
1ec5: These should be `NSInteger`-typed. In general, we should prefer Foundation typealiases over the compiler’s built-in numeric primitives in Objective-C code. It’s fine if the whole range of `NSInteger` won’t work; if it fails hard, we can clamp the value to whatever range is supported.

platform/darwin/src/MGLRasterSource.h
```diff
@@ -1,10 +1,19 @@
 #import "MGLSource.h"
+#import "MGLTypes.h"
+
+NS_ASSUME_NONNULL_BEGIN
 
 @interface MGLRasterSource : MGLSource
 
 @property (nonatomic, readonly, copy) NSURL *URL;
 @property (nonatomic, readonly, assign) CGFloat tileSize;
```
1ec5: `-mbglSource` casts this value to `uint16_t`, so this property shouldn’t be a floating-point number. (Or is the internal tile size in pixels rather than points?)

platform/darwin/src/MGLRasterSource.mm
```diff
@@ -6,18 +6,48 @@
...
+}
+
+- (std::unique_ptr<mbgl::style::Source>)mbglSource
+{
+    std::unique_ptr<mbgl::style::RasterSource> source;
+    
+    if (self.URL)
+    {
+        source = std::make_unique<mbgl::style::RasterSource>(self.sourceIdentifier.UTF8String,
+                                                             self.URL.absoluteString.UTF8String,
```
1ec5: Call `-[NSURL(MGLAdditions) mgl_URLByStandardizingScheme]` in order to canonicalize asset: URLs and such.

platform/darwin/src/MGLRasterSource.mm
```diff
@@ -6,18 +6,48 @@
...
+                                                             uint16_t(self.tileSize));
+    }
+    else
+    {
+        mbgl::Tileset tileSource;
+        tileSource.zoomRange = mbgl::Range<uint8_t>(self.minimumZoomLevel, self.maximumZoomLevel);
+        for (NSString *tileURLTemplate in self.tileURLTemplates)
+        {
+            tileSource.tiles.push_back(tileURLTemplate.UTF8String);
```
1ec5: Do we need something similar to `-mgl_URLByStandardizingScheme` for URL template strings? I guess it’d be pretty impractical to store tiles locally and include them via a URL template.

1ec5:
We should probably add an MGLTileSet class that exposes all the options `mbgl::style::TileSet` does and accept that in the MGLRasterSource initializer instead of the URL templates. Otherwise, it wouldn’t be possible to set any of the other TileJSON properties such as `data` or `attribution` (which will become important once #5999 lands).

boundsj:
> We should probably add an MGLTileSet class

This work has started in c0fb939

platform/darwin/src/MGLTileSet.mm
```diff
@@ -0,0 +1,72 @@
+#import "MGLTileSet.h"
+
+#include <mbgl/util/tileset.hpp>
+
+@implementation MGLTileSet
+{
+    BOOL _minimumZoomLevelIsSet;
```
1ec5: If the minimum and maximum zoom levels are optional, then do they need to be represented as NSNumbers publicly? Otherwise, if you use the `-initWithTileURLTemplates:` initializer, these properties end up being 0.

boundsj:
With the latest changes here, you can now mash up vector and raster sources created with tile sets:

```ObjC
    // Use raster tile URL templates directly (i.e. mapbox://tiles/mapbox.satellite/{z}/{x}/{y}{ratio}.png)
    // This is Stamen's terrain
    MGLTileSet *rasterTileSet = [[MGLTileSet alloc] initWithTileURLTemplates:@[@"http://a.tile.stamen.com/terrain-background/{z}/{x}/{y}.jpg"]];
    MGLRasterSource *rasterSource = [[MGLRasterSource alloc] initWithSourceIdentifier:@"raster-source" tileSize:256 tileSet:rasterTileSet];
    [self.mapView.style addSource:rasterSource];
    
    MGLRasterStyleLayer *rasterLayer = [[MGLRasterStyleLayer alloc] initWithLayerIdentifier:@"raster-layer" source:rasterSource];
    [self.mapView.style addLayer:rasterLayer];
    
    // Use vector tile URL templates directly. This is Mapzen's vector tile service
    MGLTileSet *vectorTileSet = [[MGLTileSet alloc] initWithTileURLTemplates:@[@"https://vector.mapzen.com/osm/all/{z}/{x}/{y}.mvt?api_key=akey"]];
    MGLVectorSource *vectorSource = [[MGLVectorSource alloc] initWithSourceIdentifier:@"mapzen" tileSet:vectorTileSet];
    [self.mapView.style addSource:vectorSource];
    
    MGLFillStyleLayer *buildingsFillLayer = [[MGLFillStyleLayer alloc] initWithLayerIdentifier:@"filllayer" source:vectorSource sourceLayer:@"buildings"];
    buildingsFillLayer.fillColor = [UIColor magentaColor];
    [self.mapView.style addLayer:buildingsFillLayer];
```

#### Not exactly "Streets"
screen shot 2016-09-21 at 3 00 31 pm

The current implementation does not attempt to reflect the mbgl object's default minimum and maximum zoom values if the Objective-C object's properties are nil. However, documentation was added that states that the defaults will be used if the Objective-C tile set object is used to initialize a source and the min/max values are nil.

cc @1ec5

platform/darwin/src/MGLTileSet.h
```diff
@@ -0,0 +1,78 @@
...
+/**
+ The `MGLTileSet` class holds the tile URL template strings and associated
+ configuration for those strings. It can be passed to a `MGLVectorSource` or
+ `MGLRasterSource` instead of an `NSURL` representing a TileJSON URL to create a 
+ source.
+ */
+@interface MGLTileSet : NSObject
+
+/**
+ A REQUIRED `NSArray` of `NSString` objects that represent the tile templates.
```
1ec5: Optionality should be indicated through nullability qualifiers; shouting isn't needed here.

platform/darwin/src/MGLTileSet.h
```diff
@@ -0,0 +1,78 @@
...
+ 
+ @param tileURLTemplates An `NSArray` of `NSString` objects that represent the 
+ tile templates.
+ @param minimumZoomLevel An `NSNumber` object containing an integer; specifies 
+ the minimum zoom level at which to display tiles.
+ @param maximumZoomLevel An `NSNumber` object containing an integer; specifies
+ the maximum zoom level at which to display tiles.
+ @return The initialized tile set object.
+ */
+- (instancetype)initWithTileURLTemplates:(NS_ARRAY_OF(NSString *) *)tileURLTemplates minimumZoomLevel:(NSNumber *)minimumZoomLevel maximumZoomLevel:(NSNumber *)maximumZoomLevel;
```
1ec5: The zoom level parameters should either be nullable or, better yet, primitive typed.

platform/ios/test/MGLTileSetTests.mm
```diff
@@ -0,0 +1,58 @@
...
+    // a tile set that provides an mbgl tile set
+    MGLTileSet *tileSet = [[MGLTileSet alloc] initWithTileURLTemplates:@[@"tile.1",
+                                                                         @"tile.2",
+                                                                         @"tile.3"]];
+    mbgl::Tileset mbglTileset = [tileSet mbglTileset];
+    
+    // has the correct URL templates
+    XCTAssertEqual(mbglTileset.tiles[0], "tile.1");
+    XCTAssertEqual(mbglTileset.tiles[1], "tile.2");
+    XCTAssertEqual(mbglTileset.tiles[2], "tile.3");
```
1ec5: Also assert on the tile set count.

platform/ios/test/MGLTileSetTests.mm
```diff
@@ -0,0 +1,58 @@
...
+    // when the tile set has min and/or max zoom level set
+    tileSet.minimumZoomLevel = @(1);
+    tileSet.maximumZoomLevel = @(2);
+    
+    // the mbgl object reflects the set values for min and max zoom level
+    XCTAssertEqual([tileSet mbglTileset].zoomRange.min, 1);
+    XCTAssertEqual([tileSet mbglTileset].zoomRange.max, 2);
+    
+    // when the tile set has an attribution
+    tileSet.attribution = @"my tileset";
```
1ec5: Also test that Unicode characters are preserved.

platform/ios/test/MGLTileSetTests.mm
```diff
@@ -0,0 +1,58 @@
...
+    tileSet.minimumZoomLevel = nil;
+    tileSet.maximumZoomLevel = nil;
+    
+    // the mbgl object has default values for min and max zoom level
+    XCTAssertEqual([tileSet mbglTileset].zoomRange.min, 0);
+    XCTAssertEqual([tileSet mbglTileset].zoomRange.max, 22);
+    
+    // when the tile set has min and/or max zoom level set
+    tileSet.minimumZoomLevel = @(1);
+    tileSet.maximumZoomLevel = @(2);
```
1ec5: If the developer sets a maximum zoom level less than the minimum, what happens? Would it be possible to test for an Objective-C exception being thrown (ideal)?

boundsj: 0fc9cbc throws Objective-C exceptions if values are crossed initially or later on in setters. This avoids a more shocking C++ exception that happens later when the doomed threshold is crossed when the map is zoomed. 

Alternatively, we could ignore invalid values and log a warning avoiding an exception altogether.

platform/darwin/src/MGLRasterSource.h
boundsj: f7a4427 changes this to an `NSUInteger`. I think the tile size here represents "tile units" as discussed in https://github.com/mapbox/mapbox-gl-native/issues/6236

platform/darwin/src/MGLTileSet.h
```diff
@@ -11,43 +11,43 @@ typedef NS_ENUM(NSUInteger, MGLTileSetScheme) {
...
 @property (nonatomic, copy) NS_ARRAY_OF(NSString *) *tileURLTemplates;
 
 /**
- An OPTIONAL `NSNumber` object containing an integer; specifies the minimum zoom
- level at which the source will display tiles. The value should be in the range 
- of 0 to 22. The default value is 0 and the source will use the default value
+ An `NSNumber` object containing an integer; specifies the minimum zoom level at 
+ which the source will display tiles. The value should be in the range of 0 to 
+ 22. The default value is 0 and the source will use the default value
  if `minimumZoomLevel` is nil;
```
1ec5: This sentence and two more below end in semicolons instead of periods.

platform/darwin/src/MGLTileSet.h
```diff
@@ -11,43 +11,43 @@ typedef NS_ENUM(NSUInteger, MGLTileSetScheme) {
...
+ which to display tiles. The value should be in the range of 0 to 22 and greater 
+ than `minimumZoomLevel`. The default value is 22 and the source will use the 
+ default value if `maximumZoomLevel` is nil;
  */
 @property (nonatomic, nullable) NSNumber *maximumZoomLevel;
 
 /**
- An OPTIONAL `NSString` object; Contains an attribution to be displayed
- when the map is shown to a user. The default value is nil;
+ An `NSString` object; Contains an attribution to be displayed when the map is
```
1ec5: Nit: replace the semicolon with “that” and lowercase “contains” (plus one more below).

platform/ios/test/MGLTileSetTests.mm
```diff
@@ -43,7 +44,7 @@ - (void)testMBGLTileSet {
     XCTAssertEqual([tileSet mbglTileset].zoomRange.max, 2);
     
     // when the tile set has an attribution
-    tileSet.attribution = @"my tileset";
+    tileSet.attribution = @"my tileset ©";
```
1ec5: © is a decent test, since it falls outside of ASCII. In #5999, I used ©️, which pairs © with a variation selector for an emoji look. But something outside the Basic Multilingual Plane, for example in the [Miscellaneous Symbols and Pictographs](https://en.wikipedia.org/wiki/Miscellaneous_Symbols_and_Pictographs) block, would be even better for detecting mojibake.

boundsj: Thanks! I think 0e31030 should probably wrap this up

PR #6322, opened 2016-09-13
[core] Band-aid to prevent GeometryTile::redoLayout race condition from causing a crash

jfirebaugh:
This provides temporary prevention for crashing due to #6263, which will hopefully help stabilize our Android builds prior to stuff like #1471 landing.

Instead of a crash, the race condition will produce a blank tile. Note that the race is run only if a layout property or filter is modified. I don't see any way for it to be triggered in other scenarios. So it won't become a possible cause of "blank tiles" in general.

### DIFF
--- a/src/mbgl/tile/tile_worker.cpp
+++ b/src/mbgl/tile/tile_worker.cpp
@@ -136,6 +136,12 @@ void TileWorker::parseLayer(const Layer* layer) {
     if (obsolete)
         return;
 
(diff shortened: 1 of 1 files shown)


tmpsantos:
:+1:

mikemorris:
Could this potentially return blank tiles instead of errors in `MapMode::Still` as well? No immediate plans for a new Node.js bindings release and not too concerned given limited scope, just want to be aware.

jfirebaugh:
@mikemorris Only if you use `map.setLayoutProperty` or `map.setFilter`.

PR #6324, opened 2016-09-13
Unordered Map

### LINKED ISSUE #3980: Revisit std::map and std::set usage
As flagged by https://github.com/mapbox/mapbox-gl-native/issues/3179, we use `std::set` and `std::map` in places where ordering is not important. In these cases, we should replace with the `unordered` versions of these containers (faster).

/cc @daniel-j-h

bleege:
Fixes #3980

### DIFF
--- a/platform/macos/src/MGLMapView.mm
+++ b/platform/macos/src/MGLMapView.mm
@@ -30,7 +30,7 @@
 #import <mbgl/util/constants.hpp>
 #import <mbgl/util/chrono.hpp>
 
-#import <map>
+#import <unordered_map>
 #import <unordered_set>
 
 #import "NSBundle+MGLAdditions.h"
@@ -102,7 +102,7 @@ enum { MGLAnnotationTagNotFound = UINT32_MAX };
 
 /// Mapping from an annotation tag to metadata about that annotation, including
 /// the annotation itself.
-typedef std::map<MGLAnnotationTag, MGLAnnotationContext> MGLAnnotationContextMap;
+typedef std::unordered_map<MGLAnnotationTag, MGLAnnotationContext> MGLAnnotationContextMap;
 
 /// Returns an NSImage for the default marker image.
 NSImage *MGLDefaultMarkerImage() {
--- a/src/mbgl/algorithm/generate_clip_ids.hpp
+++ b/src/mbgl/algorithm/generate_clip_ids.hpp
@@ -3,9 +3,9 @@
 #include <mbgl/tile/tile_id.hpp>
 #include <mbgl/util/clip_id.hpp>
 
-#include <set>
+#include <unordered_set>
 #include <vector>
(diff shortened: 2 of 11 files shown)


platform/macos/src/MGLMapView.mm
```diff
@@ -102,7 +102,7 @@
 
 /// Mapping from an annotation tag to metadata about that annotation, including
 /// the annotation itself.
-typedef std::map<MGLAnnotationTag, MGLAnnotationContext> MGLAnnotationContextMap;
+typedef std::unordered_map<MGLAnnotationTag, MGLAnnotationContext> MGLAnnotationContextMap;
```
1ec5: 👍 This is a good choice for what’s essentially a sparse array. By the way, the iOS SDK implementation of MGLMapView defines its own `MGLAnnotationContextMap` and uses it the same way as this implementation.

bleege:
I reviewed the failing Android build on Bitrise to determine the cause of the issue.  According to the run log the issue that caused the error was during the testing on the Google Cloud Service and not the local build nor local unit tests.  I was then able to review the logs on the Google Cloud Service and discovered that the only stacktraces from errors were caused by system level components of the device and the jacoco.  In other words not the Mapbox SDK nor TestApp.  This was also confirmed by watching the recorded test video of the run.

Next step is to build this version of the branch locally and run on a local Android device to ensure that it indeed works.  If it does, then that'll confirm the Google Cloud Services issue as a false positive and we'll be good to merge.

/cc @lucaswoj @jfirebaugh

bleege:
I've built this branch locally and successfully tested the entire Android TestApp suite on a local Samsung S5 running API 21 (aka the same thing that the automated system should be doing).  This makes me think this is indeed a false positive from Google Cloud Services and that we're safe to merge.

Next steps is to squash and merge.

bleege:
Squashed, Rebased on to `master`, and then merged to `master`.

bleege:
Oops, it looks like my rebase onto `master` accidentally removed a line of code from some work that @jfirebaugh had done.  My goal was to take the entirely new version from @jfirebaugh's work as it'd actually removed any need for `unordered_map` refactoring.  I'm going to re-add this line of code in a new PR.

screen shot 2016-09-14 at 4 46 27 pm

bleege:
Also, here's the list of all the commits / files changed during this PR before it was squashed.  This may be helpful in debugging potential future issues.

screen shot 2016-09-14 at 4 24 37 pm

bleege:
I just put the previously stomped on `friend class SymbolLayout;` back into `master` in #6336 so everything should be good to go again.

platform/macos/src/MGLMapView.mm
1ec5: Can you make the same change to the iOS SDK for consistency?

PR #6326, opened 2016-09-14
[android] - add default value to camera position builder

### LINKED ISSUE #6312: Empty CameraPosition.Builder crashes
When adding an empty camera position object:

```java
// configure inital map state
MapboxMapOptions options = new MapboxMapOptions()
        .styleUrl(Style.MAPBOX_STREETS)
        .camera(new CameraPosition.Builder().build());
```

It crashes with:

```java
 Caused by: java.lang.NullPointerException: Attempt to read from field 'com.mapbox.mapboxsdk.geometry.LatLng com.mapbox.mapboxsdk.camera.CameraPosition.target' on a null object reference
```

tobrun:
closes #6312 

Review @ivovandongen

### DIFF
--- a/platform/android/MapboxGLAndroidSDK/src/main/java/com/mapbox/mapboxsdk/camera/CameraPosition.java
+++ b/platform/android/MapboxGLAndroidSDK/src/main/java/com/mapbox/mapboxsdk/camera/CameraPosition.java
@@ -239,6 +239,9 @@ public final class CameraPosition implements Parcelable {
          * @return CameraPosition
          */
         public CameraPosition build() {
+            if (target == null) {
+                throw new NullPointerException("Target can't be null, provide a non-null value through CameraPosition.Builder#target(LatLng)");
+            }
             return new CameraPosition(target, zoom, tilt, bearing);
         }
 


ivovandongen:
@tobrun As this is most probably due to a programmer error, doesn't it make more sense to throw an runtime exception like `IllegalStateException`? (Fail early, etc) Might lead to tricky to debug issues otherwise.

cc @zugaldia

zugaldia:
Also, curious to know how the Google Maps SDK behaves in this case.

tobrun:
Google throws a NullPointer:

screen shot 2016-09-15 at 11 19 38


I will fix this up and replace setting the default LatLng to a throwing a NullPointerException.

zugaldia:
@tobrun 👍 thanks for looking into it.

tobrun:
@zugaldia added in 47dd653

tobrun:
Hmm, our java unit tests don't like our change:

```java
    @Test
    public void testZoomUpdateBuilder() {
        float zoomLevel = 5;
        CameraPosition.Builder builder = new CameraPosition.Builder(
                (CameraUpdateFactory.ZoomUpdate) CameraUpdateFactory.zoomTo(zoomLevel));
        assertEquals("zoom should match", zoomLevel, builder.build().zoom, 0);
    }
```

Our current implementation relies on the ability of having a null value for target. This makes both the original suggestion as the adapted suggestion incorrect. We should look into throwing the exception further down the chain. Luckily we have unit tests catching these problems upfront 😎

zugaldia:
> We should look into throwing the exception further down the chain.

@tobrun Do you mean throwing the `NullPointerException` when we use the `CameraPosition` in `animateCamera`, but not when we `build()` the `CameraPosition`?

clydebarrow:
Looking at the code, the `CameraPosition.Builder` uses a null target for no change in center, and -1 for no change to the other parameters. At least one of these should be set to a valid value so how about this:
````
        public CameraPosition build() {
            if(target == null && zoom == -1 && tilt == -1 && bearing == -1)
                throw new IllegalArgumentException("CameraPosition is empty");
            return new CameraPosition(target, zoom, tilt, bearing);
        }
````
I suggest using `IllegalArgumentException` as this appropriately implies that the arguments are invalid, not that an attempt was made to dereference a null pointer. Just because Google Maps has that behaviour does not mean it's intentional or desirable.

tobrun:
Since we opted to throw a nullpointer instead of proposed default value, this matches current behaviour. Closing.

Issue #6303, opened 2016-09-10
Text label with zoom stops behaves incorrectly on Android


clydebarrow:
**Platform: Android:**
**Mapbox SDK version 4.2.0-beta.2:**

### Steps to trigger behavior

 1. I have a style that changes labels for a point based on zoom level:

                "text-field": {
                    "base": 1,
                    "stops": [
                        [
                            0,
                            "{code}"
                        ],
                        [
                            10,
                            "{name}"
                        ],
                        [
                            22,
                            "{name}"
                        ]
                    ]
                },
Previewing this map in Studio, or using the sharing URL gives the expected behaviour, i.e. the label displays the {code} property at low zoom levels, and the {name} property at high zoom levels: 
![screen shot 2016-09-10 at 8 48 39 pm](https://cloud.githubusercontent.com/assets/2366188/18410538/40e70d4e-77a9-11e6-8a35-76f8fcfd5f9d.png)
. When using this map in an Android app only the higher zoom label ({name} in the above example) is displayed, irrespective of the zoom level.
![screenshot_2016-09-10-20-43-26](https://cloud.githubusercontent.com/assets/2366188/18410544/5adc0088-77a9-11e6-8b17-e6c15e08f32f.png)

clydebarrow:
After some experimentation I found that adding an extra stop to the function makes it work on Android. Changing the above example to:

```xml
                "text-field": {
                    "base": 1,
                    "stops": [
                        [
                            0,
                            "{code}"
                        ],
                        [
                            9,
                            "{code}"
                        ],
                        [
                            10,
                            "{name}"
                        ],
                        [
                            22,
                            "{name}"
                        ]
                    ]
                }
```

results in {code} being shown up to zoom level 10.

jfirebaugh:
Possibly same root cause as #5823.

clydebarrow:
This begs the question as to what is the expected behaviour. With numeric functions it's possible to interpolate, but where the values are discrete there needs to be some other algorithm. Is this specified anywhere? If not I propose the following for functions with discrete values:

```
Select the value from the nearest zoom stop with a value less than or equal to the current zoom;
If no such stop exists, select the value from the nearest zoom stop with a value greater than the current zoom.
If no such stop exists the function is invalid.
```

This shouldn't break anything existing, but will make my original example and that from #5823 work as expected.

clydebarrow:
[...]

Issue #5823, opened 2016-07-28
`text-allow-overlap` isn't correctly set on zoom functions


tristen:
I have the following properties set for [`text-allow-overlap`](https://www.mapbox.com/mapbox-gl-style-spec/#layout-symbol-text-allow-overlap):

```json
"text-allow-overlap": {
  "base": 1,
  "stops": [
    [0, false],
    [10, true]
  ]
}
```

But if I test the published style natively on iOS at lower zooms where overlap should _not_ be set it's appearing:

![img_0484](https://cloud.githubusercontent.com/assets/61150/17233042/a0d613fc-54fa-11e6-8a0c-6290e35630e7.PNG)

See: __mapbox://styles/tristen/cir6y946p0000g8m7jvp8ucfe__ for reference.

jfirebaugh:
Fixed in #6328.

PR #6328, opened 2016-09-14
Fix for stepwise functions 

clydebarrow:
This change fixes the bugs reported in #6303 and #5823. It swaps the behaviour of the interpolate function for discrete values to return the value just less than the current zoom level, in accordance with the Mapbox style documentation.

### DIFF
--- a/src/mbgl/util/interpolate.hpp
+++ b/src/mbgl/util/interpolate.hpp
@@ -57,8 +57,8 @@ public:
 
 struct Uninterpolated {
     template <class T>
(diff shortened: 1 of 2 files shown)


jfirebaugh:
Ah, so this regressed in #4882. My bad.

Can you add a test [here](https://github.com/mapbox/mapbox-gl-native/blob/5f49f28d3ecf49bca62ade6e21be4cf96b0acfca/test/style/functions.cpp)? Thanks for the PR!

clydebarrow:
Ok, test added and verified.

PR #6331, opened 2016-09-14
Initial style from MAPBOX_STYLE_URL if set

brunoabinader:
There is no quick and easy way of setting a custom style in our demo applications - current process involves e.g. replacing one of the default styles with the custom style. I've been using these patches for my own debugging and thought it could be a good addition for GL native developers that frequently needs to test/debug non-default styles.

__Usage__: Export your custom style URL as a `MAPBOX_STYLE_URL` environment variable prior to running the demo application. Once the application starts, it'll use the url given by the envvar if set, otherwise use default styles.

Currently supports GLFW and Qt apps.

### DIFF
--- a/bin/glfw.cpp
+++ b/bin/glfw.cpp
@@ -148,9 +148,15 @@ int main(int argc, char *argv[]) {
 
     // Load style
     if (style.empty()) {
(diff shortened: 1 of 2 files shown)


1ec5:
Can you add this usage information to the Linux and Qt SDK readmes?

Issue #3982, opened 2016-02-17
worker usage should be customizable


jfirebaugh:
For node, we don't want to spawn four threads for each Map instance. Instead, the node thread pool should be used.

Worker needs to be more like FileSource -- an abstract interface that can be replaced with different implementations depending on the platform.

tmpsantos:
Wouldn't make `mbgl::Worker` thread-safe (it is almost already) and share it across all the map objects a more maintainable option?

jfirebaugh:
It wouldn't be ideal for node. Node already has a built-in concept of a thread worker pool, which native extensions are expected to use for CPU-intensive tasks. The right thing for an extension to do is use that pool, not launch its own threads.

jfirebaugh:
We now have an abstract interface that can facilitate this: [`Scheduler`](https://github.com/mapbox/mapbox-gl-native/blob/740c995c2b37bd896c87b2d8e22d644e09bc3e42/src/mbgl/actor/scheduler.hpp). We could, for example, write a subclass that implements `Scheduler::schedule` via `Nan::AsyncQueueWorker`. The worker `Execute` implementation would do:

```
if (auto locked = mailbox.lock()) {
    locked->receive();
}
```

(Actually this logic is already repeated [here](https://github.com/mapbox/mapbox-gl-native/blob/740c995c2b37bd896c87b2d8e22d644e09bc3e42/include/mbgl/util/run_loop.hpp#L86-L88) and [here](https://github.com/mapbox/mapbox-gl-native/blob/740c995c2b37bd896c87b2d8e22d644e09bc3e42/src/mbgl/actor/thread_pool.cpp#L25-L27) and we should extract it as `static void Mailbox::maybeReceive(std::weak_ptr<Mailbox>)`.)

The other thing we need to do is externalize the dependency, passing a `Scheduler&` into the `Map` constructor as we do `View&` and `FileSource&`, rather than creating it internally. This will need to be a coordinated change across all SDKs.

PR #6337, opened 2016-09-15
[core] Use an actor model for tile worker concurrency

### LINKED ISSUE #1471: Worker pool should share queue instead of round-robin distribution

### LINKED ISSUE #3722: Move WorkTask callback invocation after task invocation
#3711 showed that our work task API is hard to use correctly because after the task invocation, side effects like object destruction can still occur after the caller was notified. This is because the task itself is responsible for notifying the caller. Instead, the task function should return the result, our work task API should then delete the task function (which also deletes any potentially bound objects), before it notifies the caller. This will also guarantee that the task function itself is less likely to perform any actions after it called the callback.

/cc @jfirebaugh

### LINKED ISSUE #6263: Fix race condition with GeometryTile::redoLayout

jfirebaugh:
`GeometryTile` and `RasterTile` now communicate with their asynchronous worker according to the [actor model](https://en.wikipedia.org/wiki/Actor_model).

This has the following benefits:

* Source-specific processing lives in source-specific worker classes (`GeometryTileWorker` and `RasterTileWorker`).
* No need to spawn a new set of worker threads when the style changes. We could switch to a process-wide worker pool easily.
* No need to write `Worker` wrappers for every `TileWorker` method.
* Impossible to access worker state from the tile object.
* "Invocation" is now 100% fire-and-forget: there is no support for cancelling a previous message and thus no possibility of unexpectedly blocking on cancellation.
* Introduces a more flexible model for asynchronous messaging, making #667 more tractable.
* Provides a clear path for using the built-in thread pool on node (#3982).
* Fixes #1471.
* Fixes #3722.
* Fixes #6263 and removes the band-aids (#6322, #6333).

## Code docs

An `Actor<O>` is an owning reference to an asynchronous object of type `O`: an "actor".
Communication with an actor happens via message passing: you send a message to the object
(using `invoke`), passing a pointer to the member function to call and arguments which
are then forwarded to the actor.

The actor receives messages sent to it asynchronously, in a manner defined its `Scheduler`.
To store incoming messages before their receipt, each actor has a `Mailbox`, which acts as
a FIFO queue. Messages sent from actor S to actor R are guaranteed to be processed in the
order sent. However, relative order of messages sent by two *different* actors S1 and S2
to R is *not* guaranteed (and can't be: S1 and S2 may be acting asynchronously with respect
to each other).

Construction and destruction of an actor is currently synchronous: the corresponding `O`
object is constructed synchronously by the `Actor` constructor, and destructed synchronously
by the `~Actor` destructor, after ensuring that the `O` is not currently receiving an
asynchronous message. (Construction and destruction may change to be asynchronous in the
future.)

An `Actor<O>` can be converted to an `ActorRef<O>`, a non-owning value object representing
a (weak) reference to the actor. Messages can be sent via the `Ref` as well.

It's safe -- and encouraged -- to pass `Ref`s between actors via messages. This is how two-way
communication and other forms of collaboration between multiple actors is accomplished.

It's safe for a `Ref` to outlive its `Actor` -- the reference is "weak", and does not extend
the lifetime of the owning Actor, and sending a message to a `Ref` whose `Actor` has died is
a no-op. (In the future, a dead-letters queue or log may be implemented.)

Please don't send messages that contain shared pointers or references. That subverts the
purpose of the actor model: prohibiting direct concurrent access to shared state.

A `Scheduler` is responsible for coordinating the processing of messages by
one or more actors via their mailboxes. It's an abstract interface. Currently,
the following concrete implementations exist:

* `ThreadPool` can coordinate an unlimited number of actors over any number of
  threads via a pool, preserving the following behaviors:

  - Messages from each individual mailbox are processed in order
  - Only a single message from a mailbox is processed at a time; there is no
    concurrency within a mailbox

  Subject to these constraints, processing can happen on whatever thread in the
  pool is available.

* `RunLoop` is a `Scheduler` that is typically used to create a mailbox and
  `ActorRef` for an object that lives on the main thread and is not itself wrapped
  as an `Actor`:

 ```
    auto mailbox = std::make_shared<Mailbox>(*util::RunLoop::Get());
    Actor<Worker> worker(threadPool, ActorRef<Foo>(*this, mailbox));
 ```

## Notes on implementation patterns

The actor model is a constrained environment. All you can do is send messages! Implementing behaviors that would be easy in a single threaded or shared memory environment takes a degree of cleverness, or at least knowledge of certain patterns.

Here are the patterns that `GeometryTile` / `GeometryTileWorker` use to get their job done:

* **[Correlation Identifier](http://www.enterpriseintegrationpatterns.com/patterns/messaging/CorrelationIdentifier.html)** allows `GeometryTile` to know when the worker has finished processing the most recent data it's been sent, such that the tile can be considered "fully loaded".
* `GeometryTileWorker` is a **[State Machine](https://engineering.shopify.com/17488160-why-developers-should-be-force-fed-state-machines)**: it transitions between states depending on which messages it receives, and its behavior for messages of a certain type depends on what state it's in.
* `GeometryTileWorker` creates a **Self-Sent Message**: a message it sends to itself, asynchronously. Its state transitions are defined in such a way that this allows it to coalesce the other messages it receives, to avoid doing obsolete work.

If we use actors more, which I think we should, we'll probably reuse these patterns and more.

## TODO

Now:

* [x] Write unit tests for the new types. They're integration tested via existing tests, but should be independently tested.

Future:

* Convert `Thread` users to actors, and remove `Thread`. This is a strictly better model.
* Think about adding a "supervisor" concept/type and more clearly defined exception handling behavior.


kkaefer:
\o/

jfirebaugh:
Added unit tests and did some high-intensity panning, zooming, rotating, and style switching on my iPhone. Everything is looking good.

@tmpsantos, @kkaefer, want to review?

src/mbgl/tile/raster_tile_worker.cpp
```diff
@@ -0,0 +1,27 @@
+#include <mbgl/tile/raster_tile_worker.hpp>
+#include <mbgl/tile/raster_tile.hpp>
+#include <mbgl/renderer/raster_bucket.cpp>
+#include <mbgl/actor/actor.hpp>
+
+namespace mbgl {
+
+RasterTileWorker::RasterTileWorker(ActorRef<RasterTileWorker>, ActorRef<RasterTile> parent_)
```
tmpsantos: It is not clear to me what is the first parameter on this constructor used for.

src/mbgl/actor/thread_pool.cpp
```diff
@@ -0,0 +1,48 @@
...
+
+                if (terminate.load()) {
+                    return;
+                }
+
+                auto mailbox = queue.front();
+                queue.pop();
+                lock.unlock();
+
+                if (auto locked = mailbox.lock()) {
```
jfirebaugh: This is the idiomatic use of `weak_ptr`; see e.g. [here](http://en.cppreference.com/w/cpp/memory/weak_ptr). In general, declaring on a separate line would require an extra set of braces and indentation to preserve the same region of locking:

```
{
    auto locked = mailbox.lock();
    if (locked) {
        ...
    }
}
```

src/mbgl/actor/thread_pool.cpp
```diff
@@ -0,0 +1,48 @@
...
+
+    for (auto& thread : threads) {
+        thread.join();
+    }
+}
+
+void ThreadPool::schedule(std::weak_ptr<Mailbox> mailbox) {
+    std::lock_guard<std::mutex> lock(mutex);
+    queue.push(mailbox);
+    cv.notify_one();
```
jfirebaugh: It's safe because we use the version of `condition_variable::wait` which accepts a predicate. If all threads are busy, on the subsequent loop whichever thread obtains the mutex first will check the predicate `!queue.empty() || terminate.load()`, see that it's true, and not wait on the condition variable.

src/mbgl/map/map.cpp
```diff
@@ -99,6 +101,7 @@ Map::Impl::Impl(View& view_,
       contextMode(contextMode_),
       pixelRatio(view.getPixelRatio()),
       asyncUpdate([this] { update(); }),
+      workerThreadPool(4),
```
jfirebaugh: Agreed. I think we can target that for a followup that also makes the thread pool global, instead of per-`Map`.

src/mbgl/tile/raster_tile_worker.cpp
jfirebaugh: Good point, I'll document this.

Issue #4344, opened 2016-03-16
Allow changing offline metadata


1ec5:
`mbgl::OfflineRegion` supports getting the region’s metadata but not setting it. An offline region’s metadata typically contains a name. An application that allows the user to choose the name would typically also allow the user to change the name.

/cc @jfirebaugh @zugaldia

jfirebaugh:
This should be a fairly straightforward addition to `OfflineDatabase`, followed by plumbing the new API from the `OfflineDatabase` core outward through `DefaultFileSource::Impl`, `DefaultFileSource`, and then SDK bindings.

1ec5:
Specifically, on iOS/macOS, MGLOfflinePack's `context` property should be writable. Hopefully it'll be possible to set the metadata synchronously. If it must be asynchronous, we can certainly add an asynchronous setter with a completion block, but unfortunately it won't be possible to maintain KVO compliance for this property.

1ec5:
I know it may not be technically correct or entirely safe, but I think we should consider making this new API synchronous at the SDK level, perhaps by waiting on the background thread that performs this update. From the developer’s perspective, there’s so much downside to making an API asynchronous for very little gain. I liken it to how NSFileManager is completely synchronous despite file operations technically being asynchronous.

jfirebaugh:
This will need to be async in core. If you want to make the SDK binding for it do a fire-and-forget... that seems risky, but your call.

1ec5:
Core support landed in #6338. Now we need to plumb support through to the Android and iOS/macOS SDKs.

> This will need to be async in core. If you want to make the SDK binding for it do a fire-and-forget... that seems risky, but your call.

Let’s make the property read-write, with a documented caveat that writes may not take effect immediately. We can also add an asynchronous `-setContext:completionHandler:` method for developers who need more certainty.

zugaldia:
Android work tracked in https://github.com/mapbox/mapbox-gl-native/issues/6402.

tobrun:
Core integration has landed, bindings can hook into this.

PR #6338, opened 2016-09-15
[core] allow offline region metadata to be updated

mollymerp:
WIP #4344 

cc @tobrun

### DIFF
--- a/platform/default/mbgl/storage/offline_database.cpp
+++ b/platform/default/mbgl/storage/offline_database.cpp
@@ -523,6 +523,19 @@ OfflineRegion OfflineDatabase::createRegion(const OfflineRegionDefinition& defin
     return OfflineRegion(db->lastInsertRowid(), definition, metadata);
 }
 
+OfflineRegionMetadata OfflineDatabase::updateMetadata(const int64_t regionID, const OfflineRegionMetadata& metadata) {
+    // clang-format off
+    Statement stmt = getStatement(
+                                  "UPDATE regions SET description = ?1"
+                                  "WHERE id = ?2");
+    // clang-format on
+    stmt->bindBlob(1, metadata);
+    stmt->bind(2, regionID);
+    stmt->run();
+    
+    return metadata;
+}
+
 void OfflineDatabase::deleteRegion(OfflineRegion&& region) {
     // clang-format off
     Statement stmt = getStatement(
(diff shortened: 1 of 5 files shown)


platform/default/mbgl/storage/offline_database.cpp
```diff
@@ -523,6 +523,19 @@ OfflineRegion OfflineDatabase::createRegion(const OfflineRegionDefinition& defin
...
+    // clang-format off
+    Statement stmt = getStatement(
+                                  "UPDATE regions SET description = ?1"
+                                  "WHERE id = ?2");
+    // clang-format on
+    stmt->bindBlob(1, metadata);
+    stmt->bind(2, region.getID());
+    stmt->run();
+    
+    return metadata;
```
1ec5: Does this method need to return the metadata? Under what circumstances would it differ from the metadata that is passed in?

mollymerp: It doesn't -- I just put it there to keep an error from popping up but it won't be in the final version of this PR 😏

tobrun:
I identified that `OfflineRegion` is the reason @mollymerp and I were having trouble get the code to compile. The offline region is `Move-only` and therefor can only be used as an rValue and can only be constructed from the friend class `OfflineDatabase`. 

I tried optimising the code for this 15ae7df but I'm now hitting an template error in thread.cpp: 

> /Users/Nurbot/work/src/mbgl/util/thread.hpp:68:34: Non-const lvalue reference to type 'vector<[2 * ...]>' cannot bind to a temporary of type 'vector<[2 * ...]>'

Which indicates that I'm not complying to the requirements..
Not sure what I'm doing wrong here.

cc @jfirebaugh @tmpsantos

jfirebaugh:
Hmm, tricky, the API was kind of designed around fully immutable regions, so this is definitely awkward. Maybe `void DefaultFileSource::updateOfflineMetadata(int64_t regionID, OfflineRegionMetadata, std::function<void (std::exception_ptr)>)`?

tobrun:
Re https://github.com/mapbox/mapbox-gl-native/pull/6338#issuecomment-247694944:
Yes, this is a lot cleaner. Will fix this while providing the binding integration. 

Re https://github.com/mapbox/mapbox-gl-native/pull/6338#issuecomment-247652638:
Got some 👀 from @tmpsantos and it all makes sense now. It's not because of the move only class but due not adding `const` to the metadata when passing this into a thread invoke.

tobrun:
I fixed up the commits and this PR is ready for review

PR #6345, opened 2016-09-15
Do not render tiles from disabled layers

kkaefer:
When a source has a `minzoom` and `maxzoom`, tiles from it should be be rendered when they are outside the range. However, when building the list of tiles to render, we are not skipping them correctly.


kkaefer:
Missing tests

tmpsantos:
> Missing tests

:+1: after adding utests

kkaefer:
This actually turned out to be a different problem:

So far, we didn't properly disable layers that are outside the zoom range. This means that we rendered layers that should not have been rendered, albeit we didn't make any attempt to load tiles for those layers. However, when zooming in/out, existing tiles might already have been loaded in the source which continued to be rendered. In most cases they weren't actually visible because either the matrices weren't updated, or the clip IDs weren't set so that they would be "rendered" off-screen and clipped completely. In any case, we did way too much work.

kkaefer:
There are a number of follow up issues that I found while working on this:

* When no layer of a source is visible anymore, we are currently disabling the source, but we don't evict tiles that are still stored in that source.

kkaefer:
Urgh, the color is off by one on Linux. On Mac, I get `#C1DEB7`, but on Linux, it's `#C1DEB6`.

jfirebaugh:
> So far, we didn't properly disable layers that are outside the zoom range.

I thought #5828 was supposed to fix that?

kkaefer:
#5828 didn't resolve this fully. It didn't actually stop rendering the layers, it just stopped updating them. This lead to a situation where they were still rendered, but their position matrix and stencil mask wasn't updated anymore. In most situations, this resulted either in a matrix that was off-screen, or a stencil mask that was all zeros, but we still issued all of the draw calls.

This patch introduces an `enabled` field that is updated when `recalculate` is called on the layer and we read this field when iterating through the layers, e.g. when building the render list.

jfirebaugh:
Do we need to introduce more state (`enabled` member), or can we just call `needsRendering()` in more places?

kkaefer:
Changed to call `needsRendering()` directly.

Issue #6264, opened 2016-09-06
Reload source tiles when GeoJSON data is mutated


jfirebaugh:
Followup from #6201.

jfirebaugh:
GeoJSON tile reloading will be similar to reloading for style changes in #6201, but also different.

The pathway for reloading for style changes is:

1. A layout property or layer filter is changed
1. Triggering `LayerObserver::onLayer{Filter,Visibility}Changed
1. Triggering `Style::onLayer{Filter,Visibility}Changed
1. Style queues the layer source ID in `updateBatch.sourceIDs` and triggers `style::Observer::onUpdate` with `Update::Layout`
1. `Map::Impl::onUpdate` adds `Update::Layout` and triggers an asyncUpdate signal (if not already triggered)
1. The signal is processed; `Map::Impl::update` is called
1. Triggering `Style::relayout`
1. Loop over `updateBatch.sourceIDs`, calling `Source::Impl::reloadTiles` for each
1. `Source::Impl::reloadTiles` loops over tiles, calling `Tile::redoLayout` for each
1. Which is overridden in `GeometryTile::redoLayout` to trigger the workers to reprocess with a new copy of the style layers

All this only handles changes to style layers. It's fairly complex in order to support batching multiple updates into a single `redoLayout` operation. For GeoJSON data changes, maybe we can get away without batching, and instead just add a loop over tiles in `GeoJSONSource::Impl::setGeoJSON`, calling `GeometryTile::setData` on each one with the new data.

tobrun:
I have been working with @mollymerp to get this feature in. 
We were able to get the the following working:

> For GeoJSON data changes, maybe we can get away without batching, and instead just add a loop over tiles in GeoJSONSource::Impl::setGeoJSON, calling GeometryTile::setData on each one with the new data.

An example of this can be seen in:

![ezgif com-video-to-gif 5](https://cloud.githubusercontent.com/assets/2151639/18559618/f3d24672-7b2c-11e6-99b4-fe463d89054d.gif)


Which is the equivalent of the GL-JS example [here](https://www.mapbox.com/mapbox-gl-js/example/live-geojson/).

tobrun:
@jfirebaugh would you like to see the improved version happening? 
(implementing the observer approach with batching)

jfirebaugh:
No, I think we're good here unless we encounter a concrete performance
issue.

PR #6347, opened 2016-09-15
6264 mutable geojson

tobrun:
References #6264, this PR follows approach as documented in OP:

>  For GeoJSON data changes, maybe we can get away without batching, and instead just add a loop over tiles in GeoJSONSource::Impl::setGeoJSON, calling GeometryTile::setData on each one with the new data.

cc @mollymerp

### DIFF
--- a/src/mbgl/style/sources/geojson_source_impl.cpp
+++ b/src/mbgl/style/sources/geojson_source_impl.cpp
@@ -65,6 +65,24 @@ void GeoJSONSource::Impl::setGeoJSON(const GeoJSON& geoJSON) {
         geoJSONOrSupercluster =
             std::make_unique<mapbox::supercluster::Supercluster>(features, clusterOptions);
     }
+    
+    for (auto const &item : tiles) {
+        GeoJSONTile* geoJSONTile = static_cast<GeoJSONTile*>(item.second.get());
+        setTileData(*geoJSONTile, geoJSONTile->id);
+    }
+}
+
+void GeoJSONSource::Impl::setTileData(GeoJSONTile& tile, const OverscaledTileID& tileID) {
+    if (geoJSONOrSupercluster.is<GeoJSONVTPointer>()) {
+        tile.updateData(geoJSONOrSupercluster.get<GeoJSONVTPointer>()->getTile(tileID.canonical.z,
+                                                                               tileID.canonical.x,
+                                                                               tileID.canonical.y).features);
+    } else {
+        assert(geoJSONOrSupercluster.is<SuperclusterPointer>());
+        tile.updateData(geoJSONOrSupercluster.get<SuperclusterPointer>()->getTile(tileID.canonical.z,
+                                                                                  tileID.canonical.x,
+                                                                                  tileID.canonical.y));
+    }
 }
 
 void GeoJSONSource::Impl::loadDescription(FileSource& fileSource) {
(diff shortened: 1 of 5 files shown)


jfirebaugh:
This approach looks good! Can you refactor it so that the `geoJSONOrSupercluster.is<...>()` conditionals and `get<GeoJSONVTPointer>()->getTile(...)` code isn't duplicated between `GeoJSONSource::Impl::setGeoJSON` and `GeoJSONSource::Impl::createTile`? This can be done by removing the `const mapbox::geometry::feature_collection<int16_t>& features ` argument from the `GeoJSONTile` constructor, adding a method `void GeoJSONSource::Impl::setTileData(GeoJSONTile&, const OverscaledTileID& tileID)` which calls `GeoJSONTile::updateData`, and calling `setTileData` from both `setGeoJSON` and `createTile`.

include/mbgl/style/sources/geojson_source.hpp
```diff
@@ -42,6 +43,7 @@ class GeoJSONSource : public Source {
 
     void setURL(const std::string& url);
     void setGeoJSON(const GeoJSON&);
+    void setTileData(GeoJSONTile&, const OverscaledTileID& tileID);
```
jfirebaugh: I don't believe a `GeoJSONSource::setTileData` method is necessary, only `GeoJSONSource::Impl::setTileData`. Remember, `GeoJSONSource` = public core API, `GeoJSONSource::Impl` = internal API.

src/mbgl/style/sources/geojson_source_impl.cpp
```diff
@@ -65,6 +65,24 @@ void GeoJSONSource::Impl::setGeoJSON(const GeoJSON& geoJSON) {
         geoJSONOrSupercluster =
             std::make_unique<mapbox::supercluster::Supercluster>(features, clusterOptions);
     }
+    
+    for(auto const &item : tiles) {
```
jfirebaugh: Spacing nit: `for (auto const& item : tiles)`.

src/mbgl/style/sources/geojson_source_impl.cpp
```diff
@@ -126,20 +144,10 @@ Range<uint8_t> GeoJSONSource::Impl::getZoomRange() {
...
-                tileID.canonical.y).features);
-    } else {
-        assert(geoJSONOrSupercluster.is<SuperclusterPointer>());
-        return std::make_unique<GeoJSONTile>(tileID, base.getID(), parameters,
-            geoJSONOrSupercluster.get<SuperclusterPointer>()->getTile(
-                tileID.canonical.z,
-                tileID.canonical.x,
-                tileID.canonical.y));
-    }
+    std::unique_ptr<Tile> tilePointer = std::make_unique<GeoJSONTile>(tileID, base.getID(), parameters);
```
jfirebaugh: Avoid the need for a `static_cast` here by assigning to a `auto tilePointer = ...`, so that the inferred type is `std::unique_ptr<GeoJSONTile>`. (If the compiler then complains about the `return` statement, make it `return std::move(tilePointer)`.)

src/mbgl/tile/geojson_tile.cpp
```diff
@@ -77,9 +77,11 @@ class GeoJSONTileData : public GeometryTileData,
 
 GeoJSONTile::GeoJSONTile(const OverscaledTileID& overscaledTileID,
                          std::string sourceID_,
-                         const style::UpdateParameters& parameters,
-                         const mapbox::geometry::feature_collection<int16_t>& features)
+                         const style::UpdateParameters& parameters)
     : GeometryTile(overscaledTileID, sourceID_, parameters) {
+}
+    
+void GeoJSONTile::updateData(const mapbox::geometry::feature_collection<int16_t>& features){
```
jfirebaugh: Nit: space before `{`.

src/mbgl/style/sources/geojson_source_impl.cpp
tobrun: This is awesome!

PR #6355, opened 2016-09-16
Format all C++ code

kkaefer:
This changeset is a preview of how it would look if we formatted all of the code with clang-format.

Please go through this and add notes if you think there are things that should be formatted differently or not at all.

/cc @mapbox/gl


scripts/clang-tools.sh
```diff
@@ -54,13 +54,18 @@ if [[ -n $2 ]] && [[ $2 == "--diff" ]]; then
     if [[ -n $DIFF_FILES ]]; then
         echo "${DIFF_FILES}" | xargs -I{} -P ${JOBS} bash -c 'check_tidy --fix' {}
         # XXX disabled until we run clang-format over the entire codebase.
```
brunoabinader: This comment is no longer needed :)

include/mbgl/annotation/annotation.hpp
```diff
@@ -20,26 +20,23 @@ class SymbolAnnotation {
...
-    Polygon<double>,
-    MultiLineString<double>,
-    MultiPolygon<double>>;
+using ShapeAnnotationGeometry =
+    variant<LineString<double>, Polygon<double>, MultiLineString<double>, MultiPolygon<double>>;
 
 class LineAnnotation {
 public:
     ShapeAnnotationGeometry geometry;
-    style::PropertyValue<float> opacity { 1.0f };
```
brunoabinader: I'm a bit confused here - isn't `Cpp11BracedListStyle: false` supposed to preserve the whitespace here?

brunoabinader:
Once we merge this, we could then make our clang-tidy bot to also run clang-format. Have you thought about a simple mechanism to enforce developers to check for format prior to submitting a PR? Maybe some custom crafted git pre-hook?  In WebKit, we've used to run a script that both checks for format and submits the patches for review.

platform/ios/benchmark/locations.cpp
```diff
@@ -4,15 +4,10 @@ namespace mbgl {
...
 const std::vector<Location> locations = {
-    { "paris",             2.3411,   48.8664, 11,   0.0 },
-    { "paris2",            2.3516,   48.8356, 13, 273.8 },
-    { "alps",             10.6107,   46.9599,  6,   0.0 },
-    { "us east",         -84.3395,   36.9400,  5,   0.0 },
-    { "greater la",     -117.9529,   34.0259,  9,   0.0 },
-    { "sf",             -122.4202,   37.7625, 14,   0.0 },
-    { "oakland",        -122.2328,   37.8267, 12,   0.0 },
-    { "germany",           9.2280,   50.9262,  6,   0.0 },
+    {"paris", 2.3411, 48.8664, 11, 0.0},        {"paris2", 2.3516, 48.8356, 13, 273.8},
```
1ec5: This code uses fewer lines, but it’s a bit less readable than the manually tabulated struct we had been using. Not the end of the world for this particular case, but worth pointing out in case there are more severe cases.

include/mbgl/map/update.hpp
```diff
@@ -5,15 +5,15 @@
 namespace mbgl {
 
 enum class Update {
-    Nothing                   = 0,
```
jfirebaugh: Can we keep the aligned formatting?

include/mbgl/storage/default_file_source.hpp
```diff
@@ -98,7 +100,7 @@ class DefaultFileSource : public FileSource {
      * executed on the database thread; it is the responsibility of the SDK bindings
      * to re-execute a user-provided callback on the main thread.
      */
-    void deleteOfflineRegion(OfflineRegion&&, std::function<void (std::exception_ptr)>);
```
jfirebaugh: Can we keep the space between return value and argument list in callable types?

include/mbgl/storage/resource.hpp
```diff
@@ -11,15 +11,7 @@ namespace mbgl {
 
 class Resource {
 public:
-    enum Kind : uint8_t {
```
jfirebaugh: Can we keep enum values on separate lines always?

include/mbgl/style/conversion.hpp
```diff
@@ -37,7 +37,8 @@ namespace conversion {
       * `arrayMember(v)` -- called only if `isArray(v)`; returns `V` or `V&`
 
       * `isObject(v)` -- returns a boolean indicating whether `v` represents a JSON object
-      * `objectMember(v, name)` -- called only if `isObject(v)`; `name` is `const char *`; return value:
+      * `objectMember(v, name)` -- called only if `isObject(v)`; `name` is `const char *`; return
+   value:
```
jfirebaugh: Indentation level here should be deeper.

include/mbgl/style/conversion/constant.hpp
```diff
@@ -19,7 +19,7 @@ struct Converter<bool> {
     Result<bool> operator()(const V& value) const {
         optional<bool> converted = toBool(value);
         if (!converted) {
-            return Error { "value must be a boolean" };
```
jfirebaugh: Other curly brace usage is with a space; I'd prefer to keep the same here.

include/mbgl/style/conversion/layer.hpp
```diff
@@ -21,37 +21,41 @@ optional<Error> setLayoutProperty(Layer& layer, const std::string& name, const V
...
-        if (paintName.compare(0, 6, "paint.") == 0) {
-            klass = paintName.substr(6);
-        }
-
-        return eachMember(paintValue, [&] (const std::string& k, const V& v) {
-            return setPaintProperty(layer, k, v, klass);
-        });
-    });
+    return eachMember(value,
+                      [&](const std::string& paintName, const V& paintValue) -> optional<Error> {
```
jfirebaugh: Too much rightward drift -- prior formatting was better.

include/mbgl/style/conversion/make_property_setters.hpp
```diff
@@ -23,7 +23,6 @@ auto makeLayoutPropertySetters() {
 
     result["visibility"] = &setVisibility<V>;
 
-
```
jfirebaugh: Can we get it to exclude generated files?

include/mbgl/style/layer.hpp
```diff
@@ -17,17 +17,20 @@ class BackgroundLayer;
 class CustomLayer;
 
 /**
- * The runtime representation of a [layer](https://www.mapbox.com/mapbox-gl-style-spec/#layers) from the Mapbox Style
+ * The runtime representation of a [layer](https://www.mapbox.com/mapbox-gl-style-spec/#layers) from
+ * the Mapbox Style
```
jfirebaugh: Here we'd want to join the next line.

include/mbgl/style/property_value.hpp
```diff
@@ -8,30 +8,50 @@ namespace style {
...
 
-    template <class S> friend bool operator==(const PropertyValue<S>&, const PropertyValue<S>&);
+    template <class S>
+    friend bool operator==(const PropertyValue<S>&, const PropertyValue<S>&);
 
 public:
-    PropertyValue()                     : value()         {}
-    PropertyValue(         T  constant) : value(constant) {}
-    PropertyValue(Function<T> function) : value(function) {}
+    PropertyValue() : value() {
```
jfirebaugh: Previous formatting was better because it made the parallelism clear.

include/mbgl/storage/default_file_source.hpp
kkaefer: I could not find a clang-format option that does this. Even `SpaceBeforeParens: Always` does not insert them.

include/mbgl/storage/resource.hpp
kkaefer: I could not find a clang-format option that does this, but we could assign values (we probably should do that anyway), which will keep them on a separate line.

include/mbgl/style/conversion/constant.hpp
kkaefer: Where? This style emphasizes the fact that a constructor is being called.

jfirebaugh:
Closing -- I don't think we should do this project wide unless we can get more satisfactory results.

Issue #6356, opened 2016-09-16
Adjust logging level for failed OkHttp requests


zugaldia:
Currently, all failed OkHttp requests are [logged as a warning](https://github.com/mapbox/mapbox-gl-native/blob/master/platform/android/MapboxGLAndroidSDK/src/main/java/com/mapbox/mapboxsdk/http/HTTPRequest.java#L141):

```
 Log.w(LOG_TAG, String.format("[HTTP] Request could not be executed: %s", e.getMessage()));
```

This doesn't make sense, for example, when the device is not connected to the network (offline mode) and the `NoRouteToHostException` is expected. Instead, I suggest we log according to the levels defined in the `HTTPRequest` class (`CONNECTION_ERROR`, `TEMPORARY_ERROR`, `PERMANENT_ERROR`). Something like:

```
if (type == PERMANENT_ERROR) {
    Log.w(...);
} else if (type == TEMPORARY_ERROR) {
    Log.i(...);
} else if (type == CONNECTION_ERROR) {
    Log.d(...);
}
```

Thoughts @mapbox/android ?

PR #6363, opened 2016-09-18
Suppress "success" messages in logcat

clydebarrow:
When syncing offline maps the Android logcat gets spammed with thousands of messages relating to routine http requests to download resources. This change suppresses those messages, while preserving messages when errors occur.


zugaldia:
@clydebarrow Thanks for the PR and great timing. Last week I opened https://github.com/mapbox/mapbox-gl-native/issues/6356 to adjust the level of logging for HTTP requests, which I also think it's too verbose.

I wouldn't recommend removing all logging in `HTTPRequest` like your PR suggests, but I like your idea of reducing the logging produced by `MapboxAccountManager` too. I'll take a pass at this.

clydebarrow:
@zugaldia Very good. The Android logcat is already so polluted with other data that I rarely use it, but the performance must suffer from what I consider unnecessary "it worked" messages.
Also, the number of callbacks in OfflineRegion.OfflineRegionObserver#onStatusChanged() seems excessive though this may reflect a more basic issue in that the number of http requests being made (many of which appear to return no useful data) also seems excessive. I played with offline regions a few months back but shelved it because of performance issues and crashes, but am now looking at it again.

zugaldia:
@clydebarrow Makes sense, and thanks again for the feedback. As discussed, I'm gonna close this PR in favor of https://github.com/mapbox/mapbox-gl-native/pull/6378 which achieves a similar goal and should help with the logs produced by `OfflineRegion.OfflineRegionObserver#onStatusChanged()`.

For longer term solutions, @bleege just opened https://github.com/mapbox/mapbox-gl-native/issues/6372, we'd love your thoughts on that.

Issue #1825, opened 2015-07-02
Remove dependency on pkg-config


kkaefer:
When using mason, we're currently relying on `pkg-config`. Instead, the mason packages should not rely on `pkg-config` and have the correct configuration flags baked into the package files already.

springmeyer:
:+1: I'd be happy to ban pkg-config for mason packages overall: https://github.com/mapbox/mason/issues/68

springmeyer:
Solved by #5359

PR #6364, opened 2016-09-18
Install pkg-config on Bitrise

1ec5:
Bitrise’s Xcode 8 stack doesn’t have pkg-config installed, causing Mason to fail with:

```
CMake Error at .mason/mason.cmake:149 (message):
  [Mason] Could not get configuration for package glfw 3.1.2:
  /Users/vagrant/git/.mason/mason.sh: line 584: pkg-config: command not found

Call Stack (most recent call first):
  platform/macos/config.cmake:3 (mason_use)
  CMakeLists.txt:46 (include)
```

This change installs pkg-config it via Homebrew. I didn’t add it to the iOS build bot’s configuration because the iOS targets don’t appear to need it.

/cc @jfirebaugh @brunoabinader

PR #6367, opened 2016-09-19
Upgrade GLFW to 3.2.1

kkaefer:


kkaefer:
This solves #6364 by using newly built mason packages that include a `mason.ini` file. This prompts the build system to no longer invoke mason (which in turn called `pkg-config` to get flags), since the flags are now bundled with the prebuilt packages.

jfirebaugh:
Does this resolve #1825?

kkaefer:
I think so. Although ideally, we'd remove the mason repo entirely and just use Cmake/curl to download the packages.

Issue #6032, opened 2016-08-16
onMapReady() callback called before style is loaded.


ivovandongen:
When using the runtime style api to manipulate layers / sources it seems that the `onMapReady` callback is called to early when the style is not yet in cache. For example, in the following snippet the fill layer will not be colored when the cache is cleared for the app (and thus all cached resources, incl the style are cleared). However, once the style is cached, the example works properly.

```
    @Override
    protected void onCreate(Bundle savedInstanceState) {
        super.onCreate(savedInstanceState);
        setContentView(R.layout.activity_lab_land_use_styling);

        mapView = (MapView) findViewById(R.id.mapView);
        mapView.onCreate(savedInstanceState);
        mapView.getMapAsync(new OnMapReadyCallback() {
            @Override
            public void onMapReady(final MapboxMap mapboxMap) {

                FillLayer schoolLayer = new FillLayer("school-layer", "composite");
                schoolLayer.setSourceLayer("landuse");
                schoolLayer.setProperties(
                        fillColor(Color.RED)
                );
                schoolLayer.setFilter(eq("class", "school"));
                mapboxMap.addLayer(schoolLayer);
            }
        });
    }
```

cc @jfirebaugh

frederoni:
This doesn't affect iOS and probably not macOS either.
The `onMapReady` iOS equivalent is `mapViewDidFinishLoadingMap` which is triggered by `-mbgl::View::notifyMapChange MapChangeDidFinishLoadingMap`

ivovandongen:
Thanks for checking @frederoni! The issue probably lies in this callback here then: https://github.com/mapbox/mapbox-gl-native/blob/master/platform/android/MapboxGLAndroidSDK/src/main/java/com/mapbox/mapboxsdk/maps/MapView.java#L442

We might want to refactor this a bit to make the intention clearer as well.

frederoni:
This doesn't affect iOS and probably not macOS either.
The `onMapReady` iOS equivalent is `mapViewDidFinishLoadingMap` which is triggered by `-mbgl::View::notifyMapChange MapChangeDidFinishLoadingMap` and it gets called at the right moment whether the style is cached or not.

tobrun:
Two items impacted that implementation:
 - issues with offline regions not being created
 - gl markers where reloaded and lost their offset issue?

Need to do some digging if you need concrete ticket info but these are some items to test for.

ivovandongen:
[...]

PR #6371, opened 2016-09-19
[core] style loaded signal

### LINKED ISSUE #4527: [core] Several methods assume style object is created
We have several methods that are currently crashing because `map.setStyleURL` is async. For instance `addLayer`, `addClass`, `isLoaded` and several others. We should have a signal (`MapChangeDidFinishLoadingStyle`?) emitted right after `MapContext::loadStyleJSON`.

On Android for instance `MapChangeDidFinishLoadingMap` is used to indicate that the style was loaded as a whole but we need something to indicate that the style JSON arrived (and the style object was created). Otherwise we might ended up rendering a partial state and later add a custom layer or class.

This new signal would be better implemented after we get rid of the context thread (#2909) because we can safely call `view.notifyMapChange` from `MapContext::loadStyleJSON` after that.

Finally, we should maybe throw an runtime exception when calling the methods that depend on a style when the style json is not yet loaded.

/cc @brunoabinader @jfirebaugh

ivovandongen:
Fixes #4527

Needed to properly time callbacks in #6032

### DIFF
--- a/src/mbgl/map/map.cpp
+++ b/src/mbgl/map/map.cpp
@@ -39,6 +39,7 @@ public:
     Impl(View&, FileSource&, MapMode, GLContextMode, ConstrainMode, ViewportMode);
 
     void onUpdate(Update) override;
+    void onStyleLoaded() override;
     void onStyleError() override;
     void onResourceError(std::exception_ptr) override;
 
@@ -964,6 +965,10 @@ void Map::Impl::onUpdate(Update flags) {
 
     updateFlags |= flags;
     asyncUpdate.send();
+}
+    
+void Map::Impl::onStyleLoaded() {
+    view.notifyMapChange(MapChangeDidFinishLoadingStyle);
 }
 
 void Map::Impl::onStyleError() {
--- a/test/map/map.cpp
+++ b/test/map/map.cpp
@@ -202,6 +202,26 @@ TEST(Map, StyleEarlyMutation) {
     EXPECT_NE(nullptr, map.getLayer("water"));
 }
 
+TEST(Map, StyleLoadedSignal) {
+    MapTest test;
+    Map map(test.view, test.fileSource, MapMode::Still);
+    
+    // The map should emit a signal on style loaded
+    bool emitted = false;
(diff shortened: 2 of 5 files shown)


src/mbgl/map/map.cpp
```diff
@@ -352,6 +352,8 @@ void Map::Impl::loadStyleJSON(const std::string& json) {
 
     // force style cascade, causing all pending transitions to complete.
     style->cascade(Clock::now(), mode);
+    
+    view.notifyMapChange(MapChangeDidFinishLoadingStyle);
```
tmpsantos: We probably need a signal for style loading failures. In case of parsing error, the `MapChangeDidFinishLoadingStyle` will be emitted here.

ivovandongen: In that case, does it make more sense to add the signals to the Style itself? 

In https://github.com/mapbox/mapbox-gl-native/blob/master/src/mbgl/style/style.cpp#L95 we can best check for parse errors.

tmpsantos: Yes, I think so.

ivovandongen: @tmpsantos to keep things clean, I've added a method to the observer on Style and kept the signal emitting in Map.

test/map/map.cpp
```diff
@@ -202,6 +202,31 @@ TEST(Map, StyleEarlyMutation) {
...
+        if (change == mbgl::MapChangeDidFinishLoadingStyle) {
+            emitted = true;
+        }
+    });
+    map.setStyleJSON(util::read_file("test/fixtures/api/empty.json"));
+    EXPECT_TRUE(emitted);
+    
+    // But not when the style couldn't be parsed
+    emitted = false;
+    test.view.setMapChangeCallback([&](MapChange change) {
```
tmpsantos: No need to redefine this callback.

ivovandongen: Good point...

ivovandongen:
@tmpsantos I've committed the changes you requested. Let me know if there's anything else or that it can be merged.

tmpsantos:
:shipit:

PR #6375, opened 2016-09-19
[core] Add interface to add images to the style

tmpsantos:
Unlike annotations, they will go away when we change the style. This API can also be used to change a sprite downloaded by a style at runtime.


include/mbgl/map/map.hpp
```diff
@@ -161,6 +161,10 @@ class Map : private util::noncopyable {
     void addLayer(std::unique_ptr<style::Layer>, const optional<std::string>& beforeLayerID = {});
     void removeLayer(const std::string& layerID);
 
+    // Add icon, bound to the style
+    void addIcon(const std::string&, std::shared_ptr<const SpriteImage>);
```
jfirebaugh: Should be `std::unique_ptr` for consistency with `addLayer` and `addSource`.

tmpsantos:
@jfirebaugh Thanks for reviewing. Changes made.

> Want to start with that naming convention here?

Makes sense.

1ec5:
Tracking iOS/macOS support for style image management in #6436.

/cc @zugaldia

zugaldia:
@1ec5 Thanks for the heads up. Android's work in now tracked in https://github.com/mapbox/mapbox-gl-native/issues/6459.

PR #6376, opened 2016-09-19
[core] Allow queryRenderedFeatures prior to symbol placement

### LINKED ISSUE #6374: Panning and zooming -> Crash with assertion "collisionTile" failed'
**Platform:**  Android 
**Mapbox SDK version:** Built from source, commit 1014a503a22dc47e4e6ec3c37d034fd729873345

### Steps to trigger behavior

 1. Build sdk and test app from source
 2. Start draw circle example (as it is simple)
 3. Pan and zoom

### Expected behavior
Snappy panning and zooming.

### Actual behavior
Crash after panning and zooming for maybe 30s with 1s breaks between gestures.

```Abort message: '../../../src/mbgl/geometry/feature_index.cpp:82: void mbgl::FeatureIndex::query(std::unordered_map<std::string, std::vector<Feature> > &, const mbgl::GeometryCollection &, const float, const double, const double, const optional<std::vector<std::string> > &, const mbgl::GeometryTileData &, const mbgl::CanonicalTileID &, const style::Style &) const: assertion "collisionTile" failed'```

Crash log:
[mapbox_crash_3.txt](https://github.com/mapbox/mapbox-gl-native/files/480306/mapbox_crash_3.txt)

jfirebaugh:
Fixes #6374.

### DIFF
--- a/src/mbgl/geometry/feature_index.cpp
+++ b/src/mbgl/geometry/feature_index.cpp
@@ -78,8 +78,11 @@ void FeatureIndex::query(
         addFeature(result, indexedFeature, queryGeometry, filterLayerIDs, geometryTileData, tileID, style, bearing, pixelsToTileUnits);
     }
 
-    // query symbol features
-    assert(collisionTile);
+    // Query symbol features, if they've been placed.
+    if (!collisionTile) {
+        return;
+    }
+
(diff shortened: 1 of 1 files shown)


ivovandongen:
@jfirebaugh Thanks for the quick fix. Resolves the issue on Android.

Issue #6379, opened 2016-09-19
beef up runtime styling test cases


incanus:
On both iOS and Android in-repo test apps, we currently test: 

- Source/layer addition (GeoJSON, raster)
- Existing Studio-made layer style changes (colors, antialiasing, widths)

However, we don't yet exercise filters, which would have revealed #6233 earlier. 

I'm still possibly seeing an issue with filters and styling of existing layers even after #6233 was fixed, so rather than build a test case, I'm going to add one to the framework app. 

/cc @ivovandongen @frederoni @jfirebaugh

jfirebaugh:
If you want to get really beefy, we could write a harness that can execute the [test-suite runtime styling tests](https://github.com/mapbox/mapbox-gl-test-suite/tree/master/render-tests/runtime-styling) against the platform SDK APIs.

incanus:
Apparently the subject line wasn't enough to auto-close, so: got this in https://github.com/mapbox/mapbox-gl-native/pull/6433 with https://github.com/mapbox/mapbox-gl-native/pull/6433/commits/b846f543b02d088f68a32b1d24b29530c53ab346.

incanus:
Re: test harness @jfirebaugh, will keep this in mind as we continue to near runtime styling release.

PR #6389, opened 2016-09-20
revamp iOS settings UI

incanus:
This was a bit of a yak shave as a result of getting frustrated starting work on #6379 (adding a new runtime styling action), but I think it's long overdue anyway and helps us scale. 

The goal here is a reorganization of our iOS settings UI, which is currently an action sheet and here a modal table view. Here's why: 

1. The action sheet paradigm dates from a time[<sup>1</sup>](https://github.com/mapbox/mapbox-gl-native/commit/d4d4cd44151423d374e373798f09d12157babc27#diff-33b34b5f099a16158c8462652a0c3688R129) that only had **three** actions, all of which would always fit on the screen. Action sheets are ideal for this sort of UI. 

1. At current `master` we now have 18-20 (depending on `debugLoggingEnabled`) actions. Even if we cull a few, no matter how we slice it, we're going to have to scroll. Might as well use a UI that's designed for it: a table view. We're probably going to keep adding actions. 

1. Speaking of adding actions, this separates the UI (section & row labels) from the code that performs the actions, putting the actions all in one routine [<sup>2</sup>](https://github.com/mapbox/mapbox-gl-native/blob/6ade86ad55a712f13126882a47f485e40383719f/platform/ios/app/MBXViewController.m#L280-L391). This breaks out some of the inline routines around simple runtime styling, custom callout, custom user dot, and some others. This is more scalable and easier to read. 

1. Breaking out runtime styling into individually-labeled actions is better because:

  - Each could possibly benefit from a custom viewport set (out of scope for this here revamp). Right now you need to know that tapping `Manipulate Style` does like eight things, including styling buildings and removing parks, which might not be relevant to the current viewport, making it look like nothing happened. 

  - It's easier to remember what the test UI is capable of with labeled actions right there to remind you, helping us maintain the list going forward. 

1. Scrolling _any_ list as long as this action one is a pain. Here I use a group-style table view to break things up a bit by functional task. Now we can add new actions [<sup>3</sup>](https://github.com/mapbox/mapbox-gl-native/blob/6ade86ad55a712f13126882a47f485e40383719f/platform/ios/app/MBXViewController.m#L215-L273) as well as whole new sections [<sup>4</sup>](https://github.com/mapbox/mapbox-gl-native/blob/6ade86ad55a712f13126882a47f485e40383719f/platform/ios/app/MBXViewController.m#L201-L206) and things still stay tidy and easy to recall from muscle memory. 

1. This avoids the need for a new view controller class by doing some minimal `UITableViewController` setup along with making our main app view controller the table `delegate` and `dataSource`. I started down a different route of settings label & action abstraction in a bigass dictionary, but this is cleaner and is what the delegate pattern is designed for. 

1. This actually performs the debug action after the table view completes dismissal. With the action sheet, its requisite animated dismissal sometimes covered up or distracted from the actual debug action being performed. Now you get a clean look at the viewport right before the chosen action happens. 

1. This opens up the possibility of new kinds of interaction UI such as sliders, switches, color pickers, etc. for debugging. 

anigif-1474336839

/cc @1ec5 @friedbunny @frederoni

### DIFF
--- a/platform/ios/app/MBXViewController.m
+++ b/platform/ios/app/MBXViewController.m
@@ -5,6 +5,7 @@
 #import "MBXOfflinePacksTableViewController.h"
 #import "MBXAnnotationView.h"
 #import "MBXUserLocationAnnotationView.h"
+
 #import "MGLFillStyleLayer.h"
 
 #import <Mapbox/Mapbox.h>
@@ -19,6 +20,52 @@ static const CLLocationCoordinate2D WorldTourDestinations[] = {
 };
 
 static NSString * const MBXViewControllerAnnotationViewReuseIdentifer = @"MBXViewControllerAnnotationViewReuseIdentifer";
+
+typedef NS_ENUM(NSInteger, MBXSettingsSections) {
+    MBXSettingsCoreRendering = 0,
+    MBXSettingsAnnotations,
+    MBXSettingsRuntimeStyling,
+    MBXSettingsMiscellaneous,
+};
+
+typedef NS_ENUM(NSInteger, MBXSettingsCoreRenderingRows) {
+    MBXSettingsCoreRenderingResetPosition = 0,
+    MBXSettingsCoreRenderingTileBoundaries,
+    MBXSettingsCoreRenderingTileInfo,
+    MBXSettingsCoreRenderingTimestamps,
+    MBXSettingsCoreRenderingCollisionBoxes,
+    MBXSettingsCoreRenderingOverdrawVisualization,
+};
+
+typedef NS_ENUM(NSInteger, MBXSettingsAnnotationsRows) {
+    MBXSettingsAnnotations100Views = 0,
+    MBXSettingsAnnotations1000Views,
+    MBXSettingsAnnotations10000Views,
+    MBXSettingsAnnotations100Sprites,
+    MBXSettingsAnnotations1000Sprites,
+    MBXSettingsAnnotations10000Sprites,
+    MBXSettingsAnnotationsTestShapes,
+    MBXSettingsAnnotationsCustomCallout,
+    MBXSettingsAnnotationsRemoveAnnotations,
+};
+
+typedef NS_ENUM(NSInteger, MBXSettingsRuntimeStylingRows) {
+    MBXSettingsRuntimeStylingWater = 0,
+    MBXSettingsRuntimeStylingRoads,
+    MBXSettingsRuntimeStylingRaster,
+    MBXSettingsRuntimeStylingGeoJSON,
+    MBXSettingsRuntimeStylingSymbols,
+    MBXSettingsRuntimeStylingBuildings,
+    MBXSettingsRuntimeStylingFerry,
+    MBXSettingsRuntimeStylingParks,
+};
+
+typedef NS_ENUM(NSInteger, MBXSettingsMiscellaneousRows) {
+    MBXSettingsMiscellaneousWorldTour = 0,
+    MBXSettingsMiscellaneousCustomUserDot,
+    MBXSettingsMiscellaneousPrintLogFile,
+    MBXSettingsMiscellaneousDeleteLogFile,
+};
 
 @interface MBXDroppedPinAnnotation : MGLPointAnnotation
 @end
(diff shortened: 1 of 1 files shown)


incanus:
Grr, of _course_ emoji branch names give Travis trouble. This should be only for core, which this doesn't touch, so I've learned my lesson this time but will let it go for this PR.

incanus:
> replacing the index literals with enum values or adding some assertions?

Makes sense @1ec5, I thought about that but you have have good reasons to make it happen. 👍

1ec5:
#6401 tracks the storyboard approach proposed in https://github.com/mapbox/mapbox-gl-native/pull/6389#pullrequestreview-678053.

1ec5:
* [ ] Add a section to [Contributing](https://github.com/mapbox/mapbox-gl-native/blob/master/platform/ios/DEVELOPING.md#contributing) that explains how to add a new action to iosapp

incanus:
I'm not concerned about the storyboard in #6401 yet since we don't add actions _that often_, but I replaced the integers with `NS_ENUM` stuff and rebased above. If this looks good @1ec5 I'll 🚢 once tests pass.

platform/ios/app/MBXViewController.m
```diff
@@ -167,153 +216,263 @@ - (void)prepareForSegue:(UIStoryboardSegue *)segue sender:(__unused id)sender {
...
+            if (self.debugLoggingEnabled)
+            {
+                [settingsTitles addObjectsFromArray:@[
+                    @"Print Telemetry Logfile",
+                    @"Delete Telemetry Logfile",
+                ]];
+            };
+            break;
+        default:
+            break;
```
1ec5: Should be this an assertion, in case a contributor forgets to add a case here or adds it to the wrong switch statement?

incanus: Or a thrown exception?

1ec5: This is in iosapp, so there’s no need to throw exceptions. If this were SDK code, an exception would be a fine idea.

incanus: Handled in the rebased 80079729e8d8fbab67ec08f557f8d4392a77a348.

incanus:
I think I'm going to pass on: 

> - [ ] Add a section to Contributing that explains how to add a new action to iosapp

It should be pretty apparent and this document is more geared towards SDK contributions, not the test app.

incanus:
BTW Travis fails here because of the Emoji branch name. Lesson learned. It's all non-Cocoa tests, though, which this PR doesn't impact, so going to proceed with merge anyway.

.mason
```diff
@@ -1 +1 @@
-Subproject commit 7cc64b1f32930d0de99381e21dd2cab9f474fa4f
+Subproject commit 508d7add9bde4afc9c085ebe1c8395a6c867bf5c
```
1ec5: Revert this.

incanus: Whoops.

incanus: 479919b3cb36ebbd7b7502c6c7ba457e1d5628cb

platform/ios/ios.xcworkspace/xcshareddata/xcdebugger/Breakpoints_v2.xcbkptlist
```diff
@@ -11,14 +11,6 @@
             continueAfterRunningActions = "No"
             scope = "1"
             stopOnStyle = "0">
-            <Actions>
```
1ec5: Eh? Probably worth breaking out into a different PR if you don’t like the default (but disabled) breakpoint. 😉

incanus: I feel like early on this was something I reversed in `master` during this process. I'll take it out of this PR; didn't mean to change anything currently.

incanus:
Ok, one more time with this 🎪 in 8c5ccff69907eebad32b61bfc9f1c54043e3578c!

incanus:
👉 https://github.com/mapbox/mapbox-gl-native/pull/6421 to avoid Emoji pain.

PR #6373, opened 2016-09-20
Remove tiles from source when source is disabled

kkaefer:
As a followup to #6345: When no layer of a source is visible anymore, we are currently disabling the source, but we don't evict tiles that are still stored in that source.

### DIFF
--- a/src/mbgl/style/source_impl.cpp
+++ b/src/mbgl/style/source_impl.cpp
@@ -146,6 +146,19 @@ void Source::Impl::updateTiles(const UpdateParameters& parameters) {
         cache.setSize(conservativeCacheSize);
     }
 
+    removeStaleTiles(retain);
+
+    const PlacementConfig config { parameters.transformState.getAngle(),
+                                   parameters.transformState.getPitch(),
+                                   parameters.debugOptions & MapDebugOptions::Collision };
+
+    for (auto& pair : tiles) {
+        pair.second->setPlacementConfig(config);
+    }
+}
+
+// Moves all tiles to the cache except for those specified in the retain set.
+void Source::Impl::removeStaleTiles(const std::set<OverscaledTileID>& retain) {
     // Remove stale tiles. This goes through the (sorted!) tiles map and retain set in lockstep
     // and removes items from tiles that don't have the corresponding key in the retain set.
     auto tilesIt = tiles.begin();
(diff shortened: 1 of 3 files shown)


kkaefer:
There's no shared tile cache among sources, which means that there's no global limit for keeping around old tiles. That means that when a source is not used anymore because the user zoomed in or out, we'll still keep the tiles around and will never evict them. Maybe a time-based approach is something we could consider here?

jfirebaugh:
Could this be implemented entirely by checking `enabled` within `Source::Impl::updateTiles`, without a change to `Style`?

> That means that when a source is not used anymore because the user zoomed in or out, we'll still keep the tiles around and will never evict them.

How common is it for a source to transition multiple times between enabled and disabled? Would it be acceptable to clear the cache for any source that becomes disabled?

kkaefer:
When a Source is disabled, it doesn't receive `updateTiles` anymore.

kkaefer:
@jfirebaugh you're right in that we are currently always calling `Source::Impl::updateTiles` even when there is no layer that actually needs a tile from that source. However, inside `updateTiles`, it's impossible to know whether there's a layer that needs that tile, so we have to prevent this call (#6461) and instead make the Source drop the existing tiles in a separate call like in this PR.

kkaefer:
> we don't evict tiles that are still stored in that source.

That is only half true: Moving those tiles to the cache means that they *will* get evicted on low-memory warnings.

jfirebaugh:
> However, inside `updateTiles`, it's impossible to know whether there's a layer that needs that tile

Why is it impossible? Isn't `enabled` tracking exactly that state? In other words, why can't `updateTiles` be defined as:

```
    if (!enabled) {
        ... clear tiles ...
        return;
    }

    ... existing rest of code ...
```

kkaefer:
It's not impossible, but https://github.com/mapbox/mapbox-gl-native/pull/6461 removes calling `updateTiles` for sources that aren't needed to display any of the currently active layers. It's more of a question of how we want to treat sources that aren't needed: Do they still get all of the notifications and just do no-ops? The only way for the Source to know whether it's relevant is to check the `enabled` field, but that field is also set by actors outside its own control. So instead of intruding into the Source object, we ought to keep track of whether a Source is required outside of the object and only interact with it when it is actually needed?

kkaefer:
Please review

PR #6226, opened 2016-09-20
Memory.Fragmentation test failure: Abnormal memory growth detected

tmpsantos:
**Platform:** Qt4
**Mapbox SDK version:** `master` @ fe28b57fe08bcd1069dad8fd47827da6bd835285

[Tests on Travis are failing with](https://travis-ci.org/mapbox/mapbox-gl-native/jobs/156589001): 

```
../../../test/util/memory.cpp:168: Failure
Expected: (memoryFragments) < (10 * 1024 * 1024), actual: 10678272 vs 10485760
        Abnormal memory growth detected.
[  FAILED  ] Memory.Fragmentation (203157 ms)
```

/cc @tmpsantos @brunoabinader

Issue #3175, opened 2015-12-03
icon alignment differences in tests


jfirebaugh:
Moved from https://github.com/mapbox/mapbox-gl-js/issues/1569 now that this happens only in native.

icon-offset/literal:

http://mapbox.s3.amazonaws.com/mapbox-gl-native/render-tests/8219.6/index.html

![image](https://cloud.githubusercontent.com/assets/98601/11576748/e444e890-99cd-11e5-8e92-f124a58ef964.png)

brunoabinader:
Possibly Fixed in 5e7736ce87f558e2d57869efef9512bcc3d2d1a5.

brunoabinader:
Fixed in #6899.

PR #6403, opened 2016-09-20
fix bad conditional in .travis.yml

mikemorris:
To actually run Node.js tests, refs https://github.com/mapbox/mapbox-gl-native/pull/6226#issuecomment-248322166. Why are conditionals in bash so complicated.

@tmpsantos `npm run test-memory` is currently not included in `make test-node` - should it be? This slows downs the tests on Travis substantially, but appears to be passing.

There is currently a single failing test in `npm run test-suite` (passing locally on OS X though). Is this just an anti-aliasing difference where we could bump up the [`allowed`](https://github.com/mapbox/mapbox-gl-test-suite/blob/master/render-tests/text-font/chinese/style.json#L7) threshold slightly, or is there something wrong I'm not seeing @amyleew?

download

download 1

- diff: 0.00057220458984375
- zoom: 11
- center: -70.751953125,43.38904828677382
- width: 512
- height: 512

/cc @tmpsantos

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -49,7 +49,7 @@ script:
   - make run-test
 after_script:
   - ccache --show-stats
-  - ./platform/linux/scripts/after_script.sh ${TRAVIS_JOB_NUMBER} ${TRAVIS_TAG:-}
+  - ./platform/linux/scripts/after_script.sh ${TRAVIS_JOB_NUMBER}
 
 matrix:
   include:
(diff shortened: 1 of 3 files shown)


jfirebaugh:
@mikemorris I think that's #3175.

mikemorris:
@jfirebaugh That seems likely - should I ignore the one failing test or bump up the threshold slightly?

jfirebaugh:
Yeah, ignore the test with a link to that issue. At some point I'm going to do a pass and simplify a bunch of tests so that they don't trigger this condition.

Issue #6361, opened 2016-09-17
Segfault calling layerWithIdentifier after initializing MGLMapView


waphumanoid:
Current master branch, iOS platform.
If I call layerWithIdentifier method right after _mapView init, it will crash the app.
```
NSURL *styleURL = [NSURL URLWithString:@"http://...json"];// = [NSURL URLWithString:@"demo_styles.json"];
_mapView = [[MGLMapView alloc] initWithFrame:self.view.bounds styleURL:styleURL];
[self.view addSubview:_mapView];
MGLLineStyleLayer *roadLayer = (MGLLineStyleLayer *)[_mapView.style layerWithIdentifier:@"stagnation"];
```

Crash EXC_BAD_ACCESS in
```
template <>
inline bool Layer::is<FillLayer>() const {
    return type == Type::Fill;
}
```
called from 
```
- (id <MGLStyleLayer>)layerWithIdentifier:(NSString *)identifier
{
    auto layer = self.mapView.mbglMap->getLayer(identifier.UTF8String);
    
    Class clazz = [self classFromLayer:layer]; <---  layer = NULL
```

1ec5:
It isn’t possible to reliably interact with a style immediately after initializing the map, since style loading is asynchronous. The most reliable place to get the road layer would be in `-[MGLMapViewDelegate mapViewDidFinishLoadingMap:]`. We should throw an exception or fail gracefully rather than segfaulting like this, though.

Related: #6121.

1ec5:
This will be fixed by #6408.

PR #6408, opened 2016-09-21
proper use of nullability for source & layer querying

incanus:
The core functions here can return `nullptr` if queries aren't found, yet we proceed merrily until the user gets a runtime crash. We should instead follow Cocoa convention and return `nil` early. 

This also touches up source querying to also support `nullable`. 

Also curious about the background of the dead code. 

/cc @1ec5 @frederoni

### DIFF
--- a/platform/darwin/src/MGLStyle.mm
+++ b/platform/darwin/src/MGLStyle.mm
@@ -92,19 +92,11 @@ static NSURL *MGLStyleURL_emerald;
     return @(self.mapView.mbglMap->getStyleName().c_str());
 }
 
-- (mbgl::style::Layer *)mbglLayerWithIdentifier:(NSString *)identifier
-{
-    return self.mapView.mbglMap->getLayer(identifier.UTF8String);
-}
-
-- (mbgl::style::Source *)mbglSourceWithIdentifier:(NSString *)identifier
-{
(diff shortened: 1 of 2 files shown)


platform/darwin/src/MGLStyle.mm
```diff
@@ -92,19 +92,11 @@ - (NSString *)name {
...
-- (mbgl::style::Source *)mbglSourceWithIdentifier:(NSString *)identifier
-{
-    return self.mapView.mbglMap->getSource(identifier.UTF8String);
-}
-
 - (id <MGLStyleLayer>)layerWithIdentifier:(NSString *)identifier
 {
     auto layer = self.mapView.mbglMap->getLayer(identifier.UTF8String);
+
+    if (layer == nullptr) return nil;
```
1ec5: (By the way, this can be expressed more succinctly as `if (layer) return nil;`.)

1ec5:
This fixes #6361.

platform/darwin/src/MGLStyle.mm
incanus: You mean `!layer`?

incanus: Grabbed this in the rebased c99ef575d4c23279d85c4203d538b1ce0f777a35.

PR #6413, opened 2016-09-21
[android] - invalidate focal point when setting MyLocationView padding

### LINKED ISSUE #6343: Location view with padding moves with camera
Adding my location view padding and rotating/zooming the camera causes the marker to move:

ezgif com-video-to-gif 3

Reproducible using [this example](https://github.com/mapbox/mapbox-android-demo/blob/master/MapboxAndroidDemo/src/main/java/com/mapbox/mapboxandroiddemo/examples/location/CustomizeUserLocationActivity.java) from demo app

cc: @tobrun

tobrun:
closes #6343

### DIFF
--- a/platform/android/MapboxGLAndroidSDK/src/main/java/com/mapbox/mapboxsdk/maps/MapView.java
+++ b/platform/android/MapboxGLAndroidSDK/src/main/java/com/mapbox/mapboxsdk/maps/MapView.java
@@ -1269,6 +1269,12 @@ public class MapView extends FrameLayout {
 
     public void invalidateContentPadding() {
         setContentPadding(contentPaddingLeft, contentPaddingTop, contentPaddingRight, contentPaddingBottom);
+
+        if (!mapboxMap.getTrackingSettings().isLocationTrackingDisabled()) {
+            setFocalPoint(new PointF(myLocationView.getCenterX(), myLocationView.getCenterY()));
+        } else {
+            setFocalPoint(null);
+        }
     }
 
     double getMetersPerPixelAtLatitude(@FloatRange(from = -180, to = 180) double latitude) {


tobrun:
Review @cammace

tobrun:
We need to take in account the x,y translation provided with the content insets

tobrun:
I identified that we need to apply a pre translation to calculate in our padding when we do a transformation in `onDraw`. This works nice for tilt but messes with the rotation gestures (the set Focal point). Need to look further into this.

tobrun:
Let me document current investigation and what I was able to locate:
In the gifs below you will see a small indicator (black dot). This shows the current configured focal point (the point used as a reference point when performing gestures).

### Current configuration:

ezgif com-video-to-gif 6

2 notes:
 - You can see the issue from https://github.com/mapbox/mapbox-gl-native/pull/6413#pullrequestreview-970386 in action 
 - You can see that the focal point stays correctly positioned on the MyLocationView.


### Current progress:

ezgif com-video-to-gif 7

2 notes:
 - You can see the issue from https://github.com/mapbox/mapbox-gl-native/pull/6413#pullrequestreview-970386  is resolved with applying https://github.com/mapbox/mapbox-gl-native/pull/6413#issuecomment-249782576. This results in adding a `matrix.preTranslate` with using calculated offset from contentpadding.
 - You can see that the focalpoint is still following the old projection, which results in faulty gestures/camera changes. We need to be able to calculate the offsetted tilt value ourselves.

tobrun:
ezgif com-video-to-gif 8

Here you can see that I was able to calculate the offset based on the used matrix projection. While this outputs the correct value offset. It's not updating smoothly. It's really important to get this synchronisation correct since an incorrect sync will results in wrongly applied gestures as shown in previous post.

tobrun:
Proposed solution for issue above in #6594.

@cammace the code found in this PR is actually unrelated to #6594. Can you do a review of this separately? (usecase: setting padding should always update correctly)

zugaldia:
I added some padding like suggested in https://github.com/mapbox/mapbox-gl-native/issues/6343#issuecomment-248520452 and it all seems to behave as expected:

video

PR #6414, opened 2016-09-21
[android] - disable tracking when a camera change occurs,

### LINKED ISSUE #5737: User location moves with camera while tracking enabled
When user location tracking's enabled, the marker will follow where ever the camera moves to until a user performs a map gesture disabling tracking. In the gif below, I first don't have tracking enabled and works as expected, then I enable and the bug occurs. 

ezgif com-video-to-gif 3

cc: @tobrun

tobrun:
Closes #5737.

Treat camera change equal to a gesture

Review @cammace 

cc @zugaldia


zugaldia:
Thanks @tobrun , merging now.

PR #6421, opened 2016-09-21
[ios] revamp settings/debug UI in test app

incanus:
Carryover of https://github.com/mapbox/mapbox-gl-native/pull/6389 to avoid Emoji in the branch name, which is busting Travis. 

😭 

/cc @1ec5


incanus:
screen shot 2016-09-21 at 3 08 42 pm

incanus:
For future reference: @jfirebaugh says you can re-push a branch with a new name and the same commits and GitHub will be smart enough to update the pull request in place, i.e. #6389 in this case. 

Next time!

Issue #6412, opened 2016-09-21
Notify MGLMapViewDelegate when style loading completes


1ec5:
As of #6371, MGLMapView.mm has a compiler warning due to the new `MapChangeDidFinishLoadingStyle` signal not being handled by an MGLMapViewDelegate method:

```
Semantic Issue Group
/Users/mxn/hub/mapbox-gl-native/platform/macos/src/MGLMapView.mm:753:13: Enumeration value 'MapChangeDidFinishLoadingStyle' not handled in switch
```

We’ll need to add `-[MGLMapView mapView:didFinishLoadingStyle:]` to both the iOS and macOS SDKs. The second parameter would be an MGLStyle, at least once #6097 lands and an instance of MGLStyle sticks around a bit longer.

Adding this method will take us a little further from the set of delegate methods exposed by MapKit (#2775), but MapKit doesn’t have the concept of remote map types anyhow (at least publicly).

/cc @friedbunny

1ec5:
A block-based style setter (https://github.com/mapbox/mapbox-gl-native/issues/6386#issuecomment-249033669) would be a fine alternative, but we need this delegate method in the meantime.

Issue #6383, opened 2016-09-19
Replace MapChange enum and `Backend::notifyMapChange` with dedicated observer interface


jfirebaugh:
The `MapChange` enumeration values have their roots in MapKit semantics -- yet are increasingly deviating from those semantics, or expanding on them. We should replace them with an API that is less influenced by the API of a particular platform, and more influenced by the pattern we use elsewhere for observation, e.g. [the style observer](https://github.com/mapbox/mapbox-gl-native/blob/23eaf6fb86644f036d8f09e5f8f80a116518bb84/src/mbgl/style/observer.hpp). It should be general enough that MapKit-like change notifications can be built on top of it.

1ec5:
In particular, it should be possible to include relevant information, such as an error pointer, along with a change notification. Awhile back, I started experimenting with making map change notifications more object oriented in the [1ec5-delegate-fail-loading-6142](https://github.com/mapbox/mapbox-gl-native/tree/1ec5-delegate-fail-loading-6142) branch, but imitating the style observer would be even better.

jfirebaugh:
Related: https://github.com/mapbox/mapbox-gl-js/pull/3244

PR #6431, opened 2016-09-22
Core support for source-driven attribution

1ec5:
Implemented observer callbacks so the style knows when the source’s attribution changes and the map knows when the style’s attribution changes. Also implemented a getter for a tile source’s attribution.

Working towards #2723. Blocks #5999.

/cc @tmpsantos

### DIFF
--- a/platform/qt/include/qmapboxgl.hpp
+++ b/platform/qt/include/qmapboxgl.hpp
@@ -95,7 +95,6 @@ public:
         NorthLeftwards,
     };
 
-
     QMapboxGL(QObject *parent = 0, const QMapboxGLSettings& = QMapboxGLSettings());
     virtual ~QMapboxGL();
 
--- a/src/mbgl/style/source_observer.hpp
+++ b/src/mbgl/style/source_observer.hpp
@@ -17,6 +17,7 @@ public:
     virtual ~SourceObserver() = default;
 
     virtual void onSourceLoaded(Source&) {}
+    virtual void onSourceAttributionChanged(Source&, const std::string&) {}
     virtual void onSourceError(Source&, std::exception_ptr) {}
 
     virtual void onTileChanged(Source&, const OverscaledTileID&) {}
--- a/src/mbgl/style/style.cpp
+++ b/src/mbgl/style/style.cpp
@@ -18,6 +18,7 @@
 #include <mbgl/style/update_parameters.hpp>
 #include <mbgl/style/cascade_parameters.hpp>
 #include <mbgl/style/calculation_parameters.hpp>
+#include <mbgl/style/tile_source_impl.hpp>
 #include <mbgl/sprite/sprite_atlas.hpp>
 #include <mbgl/text/glyph_atlas.hpp>
 #include <mbgl/geometry/line_atlas.hpp>
@@ -25,6 +26,7 @@
 #include <mbgl/renderer/render_tile.hpp>
 #include <mbgl/util/constants.hpp>
 #include <mbgl/util/string.hpp>
+#include <mbgl/util/tileset.hpp>
 #include <mbgl/platform/log.hpp>
 #include <mbgl/math/minmax.hpp>
 
@@ -293,6 +295,28 @@ Source* Style::getSource(const std::string& id) const {
     return it != sources.end() ? it->get() : nullptr;
 }
 
+std::vector<std::string> Style::getAttributions() const {
+    std::vector<std::string> result;
+    for (const auto& source : sources) {
+        switch (source->baseImpl->type) {
+        case SourceType::Vector:
+        case SourceType::Raster: {
+            style::TileSourceImpl* tileSource = static_cast<style::TileSourceImpl*>(source->baseImpl.get());
+            auto attribution = tileSource->getAttribution();
+            if (!attribution.empty()) {
+                result.push_back(std::move(attribution));
+            }
+        }
+
+        case SourceType::GeoJSON:
+        case SourceType::Video:
+        case SourceType::Annotations:
+            break;
+        }
+    }
+    return result;
+}
+
 bool Style::hasTransitions() const {
     return hasPendingTransitions;
 }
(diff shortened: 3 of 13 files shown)


platform/qt/src/qmapboxgl.cpp
```diff
@@ -76,6 +76,7 @@ static_assert(mbgl::underlying_type(QMapboxGL::MapChangeDidFinishRenderingFrameF
 static_assert(mbgl::underlying_type(QMapboxGL::MapChangeWillStartRenderingMap) == mbgl::underlying_type(mbgl::MapChangeWillStartRenderingMap), "error");
 static_assert(mbgl::underlying_type(QMapboxGL::MapChangeDidFinishRenderingMap) == mbgl::underlying_type(mbgl::MapChangeDidFinishRenderingMap), "error");
 static_assert(mbgl::underlying_type(QMapboxGL::MapChangeDidFinishRenderingMapFullyRendered) == mbgl::underlying_type(mbgl::MapChangeDidFinishRenderingMapFullyRendered), "error");
+static_assert(mbgl::underlying_type(QMapboxGL::MapChangeSourceAttributionDidChange) == mbgl::underlying_type(mbgl::MapChangeSourceAttributionDidChange), "error");
```
tmpsantos: This `static_assert` will fail because we did not add yet `MapChangeDidFinishLoadingStyle` on Qt. Can you add that one too?

src/mbgl/style/style.cpp
```diff
@@ -293,6 +295,28 @@ Source* Style::getSource(const std::string& id) const {
...
+std::vector<std::string> Style::getAttributions() const {
+    std::vector<std::string> result;
+    for (const auto& source : sources) {
+        switch (source->baseImpl->type) {
+        case SourceType::Vector:
+        case SourceType::Raster: {
+            style::TileSourceImpl* tileSource = static_cast<style::TileSourceImpl*>(source->baseImpl.get());
+            auto attribution = tileSource->getAttribution();
+            if (!attribution.empty()) {
+                result.push_back(attribution);
```
tmpsantos: `std::move(attribution)`

src/mbgl/style/style.cpp
```diff
@@ -465,6 +489,10 @@ void Style::onSourceLoaded(Source& source) {
     observer->onUpdate(Update::Repaint);
 }
 
+void Style::onSourceAttributionChanged(Source& source, std::string attribution) {
```
tmpsantos: `const std::string&`

test/src/mbgl/test/stub_style_observer.hpp
```diff
@@ -30,6 +30,10 @@ class StubStyleObserver : public style::Observer {
         if (sourceLoaded) sourceLoaded(source);
     }
 
+    void onSourceAttributionChanged(Source& source, std::string attribution) override {
```
tmpsantos: `const std::string&`

platform/qt/src/qmapboxgl.cpp
1ec5: I added it [here](https://github.com/mapbox/mapbox-gl-native/blob/5a26cd47f8f2bffce7963ebddde4ac030494379e/platform/qt/include/qmapboxgl.hpp#L115), since `MapChangeDidFinishRenderingMapFullyRendered` was also defined there. Do I need to add it anywhere else?

src/mbgl/style/source_observer.hpp
```diff
@@ -17,6 +17,7 @@ class SourceObserver {
     virtual ~SourceObserver() = default;
 
     virtual void onSourceLoaded(Source&) {}
+    virtual void onSourceAttributionChanged(Source&, std::string) {}
```
1ec5: Done.

src/mbgl/style/style.cpp
1ec5: Done.

src/mbgl/style/style.cpp
1ec5: Done.

test/src/mbgl/test/stub_style_observer.hpp
1ec5: Done.

1ec5:
> Instead of adding `Map::getAttributions()` and `Style::getAttributions()`, can we add `Source::getAttribution()`? That leverages the existing runtime source API better.

That would make a lot of sense, though it means we’ll need more infrastructure for bubbling source-related changes up to the SDK level. The macOS implementation of MGLMapView, for instance, would need to register itself as a source observer as sources are added and removed in order to accomplish the dynamically updating attribution control in #5999. (Presumably the Qt SDK would have to do likewise, since an on-demand attribution control is only an option for mobile UIs.) How will this fit into the `MapChange` revamp proposed in #6383?

platform/qt/src/qmapboxgl.cpp
tmpsantos: > Do I need to add it anywhere else?

That is not it, this code [won't compile on Qt](https://travis-ci.org/mapbox/mapbox-gl-native/jobs/161966172#L1279) because we are missing the `MapChangeDidFinishLoadingStyle` signal on the API that must be declared before `MapChangeSourceAttributionDidChange` otherwise the enum won't fully match the counterpart which is exactly what the `static_assert` is asserting.

platform/qt/include/qmapboxgl.hpp
```diff
@@ -111,7 +111,8 @@ class Q_DECL_EXPORT QMapboxGL : public QObject
         MapChangeDidFinishRenderingFrameFullyRendered,
         MapChangeWillStartRenderingMap,
         MapChangeDidFinishRenderingMap,
-        MapChangeDidFinishRenderingMapFullyRendered
+        MapChangeDidFinishRenderingMapFullyRendered,
+        MapChangeSourceAttributionDidChange
```
tmpsantos: This code will fail on the `static_assert` because we don't have `MapChangeDidFinishLoadingStyle` before `MapChangeSourceAttributionDidChange`.

platform/qt/src/qmapboxgl.cpp
1ec5: Ah, now I understand. Done in c1ef372be4b29f5889fe80fb357423c47447bbea. (Something similar is still missing from the iOS and macOS SDKs: #6412.)

tmpsantos:
@1ec5 seems like valgrind found an issue (not exactly related to this PR). I have a fix on #6455.

tmpsantos:
> I have a fix on #6455.

Fix landed, bots should be :green_apple: if you rebase.

jfirebaugh:
Hmm, I guess I should have disapproved the PR. I don't want to commit the core API to `Map::getAttributions()`.

PR #6432, opened 2016-09-22
preserve tile query params

mapsam:
Preserves any custom parameters for the Tile URL, strip out the first parameter (access token) and append the rest back into the tile URL, if they exist.

cc @jakepruitt @jfirebaugh @incanus


mapsam:
Btw, this assumes access token will always be first, which is probably not a good thing.

mapsam:
Updated the PR to search specifically for `access_token` anywhere in the query, remove it, and rebuild the query to be added to the final tile URL.

src/mbgl/util/mapbox.cpp
```diff
@@ -187,6 +187,20 @@ std::string canonicalizeTileURL(const std::string& url, SourceType type, uint16_
     }
 
     result += "." + extension;
+
+    // get the query and remove access_token, if more parameters exist, add them to the final result
+    if (queryIdx != url.length()) {
+        const auto query = url.substr(queryIdx);
+        // get access token index
+        const auto accessTokenIdx = (query.find("access_token") == 0) ? 0 : query.find("access_token") - 1; // if it's not the first param, grab the & as well
+        const auto accessTokenEndIdx = (accessTokenIdx == 0) ? query.find("&", accessTokenIdx) + 1 : query.find("&", accessTokenIdx);
```
mapsam: This line assumes there is a second parameter by checking for `query.find("&", accessTokenIdx)`. If the response from this is `0`, we need to grab the end of the query for `accessTokenEndIdx`.

mapsam:
@jfirebaugh yep, tests were indeed failing. I've updated the URL work to use a `std::regex` instead as well as added some tests to `Mapbox.CanonicalURL` to check for different orders of URL parameters.

mapsam:
@jfirebaugh any chance I could get a review here? Much appreciated!

mapsam:
Thank you @jfirebaugh!

Issue #6386, opened 2016-09-20
Initializers for MGLStyle


1ec5:
Now that MGLStyle behaves similar to a model object (actually a window into MGLMapView), we should add initializers to MGLStyle and replace MGLMapView's styleURL property with a style property.

We'd want some subset of the following initializers:
- -initWithURL: for parity with the current API on MGLMapView
- The oft-requested -initWithJSONData:
- A more object-oriented -initWithSources:layers:

This enhancement would shift some of the complexity from MGLMapView to MGLStyle and reunite the default style API with the style URL code path. MGLOfflineRegion would continue to use style URLs. The existing "Style URL" IB inspectable will remain, because IB3 plugins _still_ haven't made a comeback. :wink:

The upshot is that styles will be fully object oriented, instead of the current situation where the developer initially deals with an NSURL only to switch to MGLStyle as soon as they want to make modifications. I'm told this approach is even a bit more Reactive, although that isn't a design goal.

This has the potential to be a fairly disruptive change, so we may want to deprecate the existing NSURL-based API instead of removing it outright.

/cf mapbox/mapbox-gl-js#1989
/cc @jfirebaugh @incanus @bsudekum @frederoni

1ec5:
-initWithJSONData: would restore the functionality I took out as part of an MGLMapView spring cleaning last year. I remain convinced that there are usually cleaner, more performant, less circuitous ways to load a style, including referring to a local file URL, but I've included the JSON initializer in this proposal for completeness' sake. I think the added complexity in MGLStyle would be manageable.

Previous discussion: #900, https://github.com/mapbox/mapbox-gl-native/issues/1147#issuecomment-88171503, #1299, #1367, #1997, #4769.

1ec5:
Combined with #6383, we could also expose an `-[MGLMapView setStyle:completionHandler:]` in addition to the `MGLMapView.style` property envisioned by this proposal.

1ec5:
@incanus points out in https://github.com/mapbox/mapbox-gl-native/pull/6636#pullrequestreview-3370778 that, once we expose any API that makes use of `setStyleJSON()`, we’ll need to make any style URL API nullable.

1ec5:
Putting on the v3.4.1 milestone for the initializers themselves. The completion block–based API proposed in https://github.com/mapbox/mapbox-gl-native/issues/6386#issuecomment-249033669 would be a somewhat longer-term goal.

[...]

Issue #6251, opened 2016-09-06
Cannot add MGL*StyleLayer for custom vector sources


boundsj:
**Platform:** iOS
**Mapbox SDK version:** 3.4.0 alpha 4

### Steps to trigger behavior

 1. Create and locally host a TileJSON document that points to a valid vector tileset (i.e. [https://vector.mapzen.com/osm/all/{z}/{x}/{y}.mvt]
 1. Create an `MGLVectorSource` that uses a URL pointed at the previously created TileJSON (adjust ATS permissions to allow this on iOS)
 1. Create a style layer (i.e. `MGLFillStyleLayer`) for the vector source and add it to the map's style

### Expected behavior

You see the vector source data visually represented by the added style.

### Actual behavior

No change to the map.

cc @1ec5 @frederoni

boundsj:
With the fix for https://github.com/mapbox/mapbox-gl-native/issues/6252 came the ability to address and style vector source layers. Since [3.4.0 alpha 4](https://github.com/mapbox/mapbox-gl-native/releases/tag/ios-v3.4.0-alpha.4), you can style the layers of a vector source that you host. For example, assuming you are hosting [a TileJSON document](https://github.com/mapbox/tilejson-spec/tree/master/2.1.0/example) locally that points to [Mapzen's vector tile service](https://mapzen.com/documentation/vector-tiles/), you can write:

```Objc
- (void)mapViewDidFinishLoadingMap:(MGLMapView *)mapView {
    NSURL *url = [NSURL URLWithString:@"http://localhost:8000/osm.layer"];
    MGLVectorSource *vectorSource = [[MGLVectorSource alloc] initWithSourceIdentifier:@"mapzen" URL:url];
    [self.mapView.style addSource:vectorSource];
    
    MGLFillStyleLayer *fillLayer = [[MGLFillStyleLayer alloc] initWithLayerIdentifier:@"filllayer" source:vectorSource sourceLayer:@"buildings"];
    fillLayer.fillColor = [UIColor magentaColor];
    [self.mapView.style addLayer:fillLayer];
}
```

and see the filled buildings:

<img width="377" alt="screen shot 2016-09-12 at 6 05 45 pm" src="https://cloud.githubusercontent.com/assets/63186/18458151/91e26994-7913-11e6-83ad-e1fc41bec452.png">

boundsj:
Correction: the ability to address a sources layer does exist on master but not in a published alpha release. The next alpha release will have this functionality.

Issue #6040, opened 2016-08-16
Android runtime style API bug with filters & numbers


incanus:
Per extended chat with @ivovandongen—writing this up so we have a concrete place for it on which to hang commits and eventually mark it resolved. 

Here is a sample data file: 

https://gist.github.com/incanus/0607dccb035cd9b55ff95faa1b60cf8c

I've added this to a basic Mapbox Studio style at `mapbox://styles/justin/cirnit09r000bg1mb4kxqmvwi`. 

Each segment of the data file is addressable by a `uniqid` numeric field, e.g. `10191020`. Something like this should be possible: 

```java
LineLayer segments = mapboxMap.getLayerAs("segments");
        segments.setFilter(eq("uniqid", 10191020));
        segments.setProperties(lineColor(Color.GREEN),
                lineWidth(10f),
                lineOpacity(0.5f));
```

This is the analog to doing it via the GUI in Studio: 

<img width="1340" alt="screen shot 2016-08-11 at 5 29 53 pm" src="https://cloud.githubusercontent.com/assets/17722/17711971/d8f42028-63a8-11e6-93e5-1b67c0156787.png">

However, all segments will get their style changed, not just the single one of the ~20 with that `uniqid`. The filter seems to be ignored. 

Numeric comparison, either types or precision, seems to be at fault. @ivovandongen tried a fix at https://github.com/mapbox/mapbox-gl-native/commit/ddf0d25845177d3a3e865dd7e64cf98cf6b8a7d7 but it still doesn't seem to do it. 

I will also test https://github.com/mapbox/mapbox-gl-native/commit/c34130ac6d968a3ff636af71e19893d22c2faa99 and see why it might be working but my example isn't. Is it maybe the size of the number in my data field? 

/cc @frederoni just for when we might hit this on iOS filters.

incanus:
Confirmed that I am seeing the parks example in c34130a work just fine in my setup, so it's something more unique about my data set.

ivovandongen:
@incanus I've tried out your json and code snippets by working it into the test app and it just seems to work. Can't say what's going wrong with your example based on the information you provided. See the patch here for more information: https://gist.github.com/ivovandongen/f6ac2d5edd66844f853126759134bc82

![segments](https://cloud.githubusercontent.com/assets/64527/17730234/313fbd9a-6468-11e6-9b75-e8518d7c8fc7.gif)

Hope that helps you.

incanus:
[...]

PR #6433, opened 2016-09-22
fix #6379: beef up runtime styling manual tests

### LINKED ISSUE #6379: beef up runtime styling test cases

incanus:
Work in progress to test filtered fills & lines.

### DIFF
--- a/platform/ios/app/MBXViewController.m
+++ b/platform/ios/app/MBXViewController.m
@@ -58,6 +58,9 @@ typedef NS_ENUM(NSInteger, MBXSettingsRuntimeStylingRows) {
     MBXSettingsRuntimeStylingBuildings,
     MBXSettingsRuntimeStylingFerry,
     MBXSettingsRuntimeStylingParks,
+    MBXSettingsRuntimeStylingFilteredFill,
+    MBXSettingsRuntimeStylingFilteredLines,
+    MBXSettingsRuntimeStylingNumericFilteredFill,
 };
 
 typedef NS_ENUM(NSInteger, MBXSettingsMiscellaneousRows) {
@@ -290,14 +293,17 @@ typedef NS_ENUM(NSInteger, MBXSettingsMiscellaneousRows) {
             break;
         case MBXSettingsRuntimeStyling:
             [settingsTitles addObjectsFromArray:@[
-                @"Apply Water Functions",
-                @"Apply Road Line Functions",
-                @"Create Raster & Apply Function",
-                @"Create GeoJSON & Apply Fill",
-                @"Apply Symbol Color",
-                @"Apply Building Fill Color",
-                @"Apply Ferry Line Color",
-                @"Remove Park Layer",
+                @"Style Water With Function",
+                @"Style Roads With Function",
+                @"Add Raster & Apply Function",
+                @"Add GeoJSON & Apply Fill",
+                @"Style Symbol Color",
+                @"Style Building Fill Color",
+                @"Style Ferry Line Color",
+                @"Remove Parks",
+                @"Style Fill With Filter",
+                @"Style Lines With Filter",
+                @"Style Fill With Numeric Filter",
             ]];
             break;
         case MBXSettingsMiscellaneous:
@@ -413,6 +419,15 @@ typedef NS_ENUM(NSInteger, MBXSettingsMiscellaneousRows) {
                     break;
                 case MBXSettingsRuntimeStylingParks:
                     [self removeParkLayer];
+                    break;
+                case MBXSettingsRuntimeStylingFilteredFill:
+                    [self styleFilteredFill];
+                    break;
+                case MBXSettingsRuntimeStylingFilteredLines:
+                    [self styleFilteredLines];
+                    break;
+                case MBXSettingsRuntimeStylingNumericFilteredFill:
+                    [self styleNumericFilteredFills];
                     break;
                 default:
                     NSAssert(NO, @"All runtime styling setting rows should be implemented");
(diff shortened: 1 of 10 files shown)


incanus:
Got some examples bolted in for styling fills & lines using filters. Both feature Studio-created layers that are well under the normal map layers, showing off some of our runtime styling flexibility. 

*Map with states fill layer: fill in Texas*

simulator screen shot sep 22 2016 12 33 01 pm

*Map with counties line layer: restyle all `Washington` counties*

simulator screen shot sep 22 2016 12 32 32 pm

Other than some minor flickering upon styling, things look good. 

Next up is adding a numeric-type filter test to flex https://github.com/mapbox/mapbox-gl-native/issues/6040 and the iOS side.

platform/ios/app/MBXViewController.m
```diff
@@ -709,6 +724,79 @@ - (void)removeParkLayer
...
 }
 
+- (void)styleFilteredFill
+{
+    // set style and focus on Texas
+    [self.mapView setStyleURL:[NSURL URLWithString:@"mapbox://styles/justin/citc7ya7w00002jpa0z16cjah"]];
+    [self.mapView setCenterCoordinate:CLLocationCoordinate2DMake(31, -100) zoomLevel:3 animated:NO];
+
+    // after slight delay, fill in Texas
+    dispatch_after(dispatch_time(DISPATCH_TIME_NOW, (int64_t)(2 * NSEC_PER_SEC)), dispatch_get_main_queue(), ^
```
1ec5: A hard-coded timeout would be unfortunate. We don't have a completion block–based way to set the style; in the meantime, you can do this in didFinishLoadingMap or the forthcoming didFinishLoadingStyle perhaps.

platform/ios/app/MBXViewController.m
```diff
@@ -709,6 +724,79 @@ - (void)removeParkLayer
     [self.mapView.style removeLayer:parkLayer];
 }
 
+- (void)styleFilteredFill
+{
+    // set style and focus on Texas
+    [self.mapView setStyleURL:[NSURL URLWithString:@"mapbox://styles/justin/citc7ya7w00002jpa0z16cjah"]];
```
1ec5: This will only work with your access token, not the access token that the user enters. Instead, check in the style JSON as part of this project and refer to the bundled JSON file with a file URL or relative path.

platform/ios/app/MBXViewController.m
```diff
@@ -709,6 +724,79 @@ - (void)removeParkLayer
...
+    [self.mapView setStyleURL:[NSURL URLWithString:@"mapbox://styles/justin/citerrmcg00662hogpcv4o3w8"]];
+    [self.mapView setCenterCoordinate:CLLocationCoordinate2DMake(40, -97) zoomLevel:5 animated:NO];
+
+    // after slight delay, change styling for all regions 200-299
+    dispatch_after(dispatch_time(DISPATCH_TIME_NOW, (int64_t)(2 * NSEC_PER_SEC)), dispatch_get_main_queue(), ^
+    {
+        MGLFillStyleLayer *regionsLayer = (MGLFillStyleLayer *)[self.mapView.style layerWithIdentifier:@"regions"];
+
+        // filter
+        regionsLayer.predicate = [NSCompoundPredicate andPredicateWithSubpredicates:@[
```
1ec5: This can be expressed as a single predicate format string using the `AND` operator.

platform/ios/app/MBXViewController.m
incanus: It splits functionality up fairly crappily, but I see your point.

1ec5: https://github.com/mapbox/mapbox-gl-native/issues/6386#issuecomment-249033669 would mitigate this pain considerably, but it’s aways off.

boundsj:
Thanks for doing this. While you are at it, I think it'd be great to add cases for the soon to land https://github.com/mapbox/mapbox-gl-native/pull/6316 to exercise tile template URLs and also something to cover styling 3rd party vector (and raster) source layers. Support for the later was added in the fix for https://github.com/mapbox/mapbox-gl-native/issues/6251. A single example can cover this (see https://github.com/mapbox/mapbox-gl-native/pull/6316#issuecomment-248774954)

incanus:
Got this pretty well set on the Android side too. 

@boundsj On the ones you mention, is it more useful to visually test the functionality or to just confirm that something happens? i.e. I'm wondering how much we can push that to automated tests. The things I'm doing here are specifically things that won't ever throw or error, but might not look right if a bug crept in.

incanus:
screenshot_20160922-171050
screenshot_20160922-171112

boundsj:
> @incanus might not look right if a bug crept in

I think that definition covers a lot of runtime styling cases including what I suggested. For example the `MGLTileSet` class that will be introduced in https://github.com/mapbox/mapbox-gl-native/pull/6316 has its own set of options that affect how the map looks. Feel free to ignore my suggestion though if you don't see value in adding that case at this time.

platform/ios/app/MBXViewController.m
incanus: c3c9ca40aece3b08b9e63c2a13030e469452a845

platform/ios/app/MBXViewController.m
incanus: 63cdd45580602e9a7206dd6d39a76fd2a40887e4

platform/ios/app/MBXViewController.m
incanus: Looked into this with a patch such as the following: 

https://gist.github.com/1ab15bce31631f5ba032a731bd0bed38

However, it happens too quickly, essentially immediately after the style finishing tiling in, causing continual flashing between that happening and then the filter style changes. I want to eliminate the possibility that anything is done as part of the initial style fill, hence the delay so that the user can clearly see things in one state, then the other, to confirm that the filtered styling is applied.

1ec5: OK, in that case, can you document the intent here? Now that we’re getting the style from a local resource, I’m not as concerned about the time it’ll take to load it, but someone using iosapp as a learning resource shouldn’t be misled into thinking this is how you use the runtime styling API normally.

incanus:
Squashed and force pushed. This 🚢 will be ready to ⛵ once tests pass.

platform/ios/app/MBXViewController.m
incanus: 89bfeaa88eec7840ecee64c15f22e34d7f6bcbdf

incanus:
Ugh, fixing Android local style JSON also first.

PR #6434, opened 2016-09-22
[Qt] Add API for adding icon to the style

tmpsantos:
Also update the runtime style example to use the new API.


tmpsantos:
:eyes: @brunoabinader

PR #6437, opened 2016-09-23
[tests] Update test-suite

jfirebaugh:
Gets a few more ignored tests passing.


jfirebaugh:
Yes, going to roll this into #6403.

PR #6441, opened 2016-09-23
[android] user agent string

### LINKED ISSUE #1993: Android user agent string
From #823 current curl implementation just uses `MapboxGL/1.0`

At @kkaefer suggestion should:
- [ ] Indicate platform e.g. MapboxGL Android, Linux etc. using build macros?
- [ ] use actual version from `version.hpp`

ivovandongen:
Fixes #1993 

Example user agent string:

```
com.mapbox.mapboxsdk.testapp/4.1.0 (9) Mapbox/4.2.0-SNAPSHOT (0c25615) Android/24 (armeabi-v7a)
```
eg: 

```
<app packagename>/<app version string> (<app version code>) Mapbox/<SDK version> (<git revision>) Android/<API level> (ABI)
```

cc @mick

### DIFF
--- a/platform/android/MapboxGLAndroidSDK/build.gradle
+++ b/platform/android/MapboxGLAndroidSDK/build.gradle
@@ -41,6 +41,9 @@ android {
     defaultConfig {
         minSdkVersion Integer.parseInt(project.ANDROID_MIN_SDK)
         targetSdkVersion Integer.parseInt(project.ANDROID_BUILD_TARGET_SDK_VERSION)
+        buildConfigField "String", "MAPBOX_EVENTS_USER_AGENT_BASE", String.format("\"MapboxEventsAndroid/%s\"", project.VERSION_NAME)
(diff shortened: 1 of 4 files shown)


platform/android/MapboxGLAndroidSDK/build.gradle
```diff
@@ -157,6 +158,12 @@ def getRepositoryPassword() {
             (hasProperty('NEXUS_PASSWORD') ? NEXUS_PASSWORD : "")
 }
 
+def getGitRevision() {
```
1ec5: The other SDKs get this from a macro in mbgl. Probably not a big deal, but worth making sure the value is consistent at least.

ivovandongen: Yes, I saw that. But this way we can initialise the string statically instead of passing the revision through JNI in the constructor of each http request. Minor optimisation, but it makes things a little cleaner as well.

PR #6442, opened 2016-09-23
Color interpolation with configurable color spaces

tmcw:
This is just a start so I can get an idea from the C++ illuminati
about whether it's the right direction. Approach here is
- Since colors are arrays in JS and objects in C++, instead of a
  rgbToLab() method, there's a to_lab() method on the color object

---

Connected to https://github.com/mapbox/mapbox-gl-js/pull/3245 - the GL JS side

### DIFF
--- a/include/mbgl/style/function.hpp
+++ b/include/mbgl/style/function.hpp
@@ -23,14 +23,17 @@ public:
     explicit Function(Stops stops_, float base_)
         : base(base_), stops(std::move(stops_)) {}
 
+    explicit Function(Stops stops_, float base_, ColorSpace colorSpace_)
+        : base(base_), stops(std::move(stops_)), colorSpace(colorSpace_) {}
+
     float getBase() const { return base; }
     ColorSpace getColorSpace() const { return colorSpace; }
     const std::vector<std::pair<float, T>>& getStops() const { return stops; }
 
 private:
     float base = 1;
-    ColorSpace colorSpace = ColorSpace::RGB;
     std::vector<std::pair<float, T>> stops;
+    ColorSpace colorSpace = ColorSpace::RGB;
 
     template <class S> friend bool operator==(const Function<S>&, const Function<S>&);
 };
--- a/include/mbgl/style/function.hpp
+++ b/include/mbgl/style/function.hpp
@@ -3,6 +3,14 @@
 #include <vector>
 #include <utility>
 
+using EnumType = uint32_t;
+
+enum class ColorSpace : EnumType {
+    RGB,
+    LAB,
+    HCL
+};
+
 namespace mbgl {
 namespace style {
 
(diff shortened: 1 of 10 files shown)


jfirebaugh:
I think eventually we'll need a set of color-space-specific classes -- `class RGBColor`, `class LABColor`, `class HSLColor` -- and a variant type to unify them. The latter will probably want to be defined as:

```
using ColorBase = variant<RBGColor, LABColor, HSLColor>;
class Color : public ColorBase {
public:
    using ColorBase::ColorBase;
    // now it's possible to define methods such as:
    RGBColor asRGB() const;
};
```

tmcw:
@jfirebaugh want to take a look at the code and the approach? This now passes unit tests locally, but there are at least three ways I could implement the to&from RGB/LCH conversion, and I chose the one I could get to work, not necessarily the most idiomatic.

include/mbgl/style/function.hpp
```diff
@@ -3,6 +3,14 @@
 #include <vector>
 #include <utility>
 
+using EnumType = uint32_t;
+
+enum class ColorSpace : EnumType {
```
jfirebaugh: Delete `EnumType`; let the compiler choose.

include/mbgl/style/function.hpp
```diff
@@ -15,19 +23,26 @@ class Function {
     explicit Function(Stops stops_, float base_)
         : base(base_), stops(std::move(stops_)) {}
 
+    explicit Function(Stops stops_, float base_, ColorSpace colorSpace_)
```
jfirebaugh: Combine with the previous constructor using a default parameter (`ColorSpace colorSpace_ = ColorSpace::RGB`).

include/mbgl/util/color.hpp
```diff
@@ -14,9 +16,14 @@ class Color {
     float b = 0.0f;
     float a = 0.0f;
 
+    ColorLAB to_lab();
```
jfirebaugh: `toLAB`

include/mbgl/util/color.hpp
```diff
@@ -14,9 +16,14 @@ class Color {
...
+    ColorHCL to_hcl();
```
jfirebaugh: `toHCL`

include/mbgl/util/color.hpp
```diff
@@ -14,9 +16,14 @@ class Color {
...
+
     static constexpr Color black() { return { 0.0f, 0.0f, 0.0f, 1.0f }; };
     static constexpr Color white() { return { 1.0f, 1.0f, 1.0f, 1.0f }; };
 
+    static Color from_lab(const ColorLAB);
```
jfirebaugh: `fromLAB`/`fromHCL`

Parameter type should be `const ColorLAB&`

1ec5:
Tracking an iOS/macOS SDK implementation for the runtime styling API in #6721.

/cc @zugaldia

1ec5:
Looks like the changes requested in https://github.com/mapbox/mapbox-gl-native/pull/6442#pullrequestreview-4522523 have been implemented. @tmcw, would you mind rebasing this PR and resolving the conflicts?

jfirebaugh:
This requires some non-trival rework now that data-driven properties have landed.

PR #6446, opened 2016-09-23
[core] Incremental offline downloads

### LINKED ISSUE #4414: Downloading an offline pack blocks the visible map from loading
If the active style of a map view is changed while an offline pack is downloading, the new style will fail to load (i.e., it only shows a black screen) while that download is active.

1. Start downloading an offline pack
2. Switch the visible style on your map view
3. The new style will not load until the offline pack finishes (or pauses) downloading

/cc @jfirebaugh @1ec5 @zugaldia

jfirebaugh:
Don't allow OfflineDownload to flood the request queue. Fixes #4414.

### DIFF
--- a/platform/default/mbgl/storage/offline_download.cpp
+++ b/platform/default/mbgl/storage/offline_download.cpp
@@ -3,6 +3,7 @@
 #include <mbgl/storage/offline_download.hpp>
 #include <mbgl/storage/resource.hpp>
 #include <mbgl/storage/response.hpp>
+#include <mbgl/storage/http_file_source.hpp>
 #include <mbgl/style/parser.hpp>
 #include <mbgl/style/sources/geojson_source_impl.hpp>
 #include <mbgl/style/tile_source_impl.hpp>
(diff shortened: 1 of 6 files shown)


tobrun:
I was thinking of picking up this `gl-school` ticket. Really cool to see it happening and need to admit I wouldn't have come up with same code. 👀  and 📓 ✏️ 'ing.

platform/default/mbgl/storage/offline_download.cpp
```diff
@@ -217,30 +182,56 @@ void OfflineDownload::activateDownload() {
...
         }
 
-        for (const auto& resource : glyphResources(parser)) {
-            ensureResource(resource);
+        if (!parser.spriteURL.empty()) {
+            queueResource(Resource::spriteImage(parser.spriteURL, definition.pixelRatio));
+            queueResource(Resource::spriteJSON(parser.spriteURL, definition.pixelRatio));
         }
+
+        continueDownload();
```
tmpsantos: Shouldn't we also call `continueDownload()` on error?

platform/default/mbgl/storage/offline_download.cpp
```diff
@@ -299,9 +286,7 @@ void OfflineDownload::ensureResource(const Resource& resource,
                 return;
             }
 
-            if (status.complete()) {
-                setState(OfflineRegionDownloadState::Inactive);
-            }
+            continueDownload();
```
tmpsantos: Shouldn't we call `continueDownload()` also on error?

jfirebaugh: A request that errors will be retried after some delay. So in that sense it's still "active" and consuming resources, notably the request object, its timer, and network resources when the timer fires.

We _could_ try to squeeze in subsequent requests while we wait for the errored request to retry. But that risks overloading the upstream request queue -- defeating our own metering here -- if there are a lot of errored requests that all come up for retry at the same time.

Many times, the cause of a request error will apply to many requests of the same type. For instance if a server is unreachable, all the requests to that host are going to error. In that case, continuing to try subsequent resources after the first few errors is fruitless anyway.

I'll add the above rationale as a comment.

platform/default/mbgl/storage/offline_download.cpp
```diff
@@ -152,18 +113,21 @@ OfflineRegionStatus OfflineDownload::getStatus() const {
         }
     }
 
-    result.requiredResourceCount += spriteResources(parser).size();
-    result.requiredResourceCount += glyphResources(parser).size();
+    if (!parser.glyphURL.empty()) {
+        result.requiredResourceCount += parser.fontStacks().size() * 256;
```
1ec5: Factor 256 (used multiple times here) into a global constant?

PR #6448, opened 2016-09-23
Doc gen handles arrays as required property values

davidtheclark:
Ref https://github.com/mapbox/mapbox-gl-style-spec/issues/504.

Modifies the style documentation generation code to allow for an array as the required property value. An array means that the property must have one of multiple possible values.

I tried to match the sentence-style syntax and punctuation I saw elsewhere. Here's what we end up with in a few scenarios:

```
"requires": [ 
  {
    "icon-text-fit": ["both", "width", "height"]
  },
],
``` 
> This property is only applied to the style if `iconTextFit` is set to either an `NSValue` object containing `MGLSymbolStyleLayerIconTextFitBoth`, or an `NSValue` object containing `MGLSymbolStyleLayerIconTextFitWidth`, or an `NSValue` object containing `MGLSymbolStyleLayerIconTextFitHeight`. Otherwise, it is ignored.

```
"requires": [
  {
    "icon-text-fit": [
      "both",
      "width"
    ]
  }
],
```

> This property is only applied to the style if `iconTextFit` is set to either an `NSValue` object containing `MGLSymbolStyleLayerIconTextFitBoth` or an `NSValue` object containing `MGLSymbolStyleLayerIconTextFitWidth`. Otherwise, it is ignored.

```
"requires": [
  {
    "icon-text-fit": "both"
  }
],
```

> This property is only applied to the style if `iconTextFit` is set to an `NSValue` object containing `MGLSymbolStyleLayerIconTextFitBoth`. Otherwise, it is ignored.

(Looks like I also stripped some superfluous end-of-line whitespace.)

cc @1ec5

### DIFF
--- a/platform/darwin/scripts/generate-style-code.js
+++ b/platform/darwin/scripts/generate-style-code.js
@@ -107,7 +107,7 @@ global.propertyReqs = function (property, layoutPropertiesByName, type) {
             return '`' + camelizeWithLeadingLowercase(req['!']) + '` is set to `nil`';
         } else {
             let name = Object.keys(req)[0];
-            return '`' + camelizeWithLeadingLowercase(name) + '` is set to ' + describeValue(req[Object.keys(req)[0]], layoutPropertiesByName[name], type);
+            return '`' + camelizeWithLeadingLowercase(name) + '` is set to ' + describeValue(req[name], layoutPropertiesByName[name], type);
         }
     }).join(', and ') + '. Otherwise, it is ignored.';
 };
@@ -131,8 +131,21 @@ global.describeValue = function (value, property, layerType) {
         case 'string':
             return 'the string `' + value + '`';
         case 'enum':
-            let objCType = `${prefix}${camelize(layerType)}${suffix}${camelize(property.name)}`;
-            return 'an `NSValue` object containing `' + `${objCType}${camelize(value)}` + '`';
+            let displayValue;
+            if (Array.isArray(value)) {
+              let separator = (value.length === 2) ? ' ' : ', ';
+              displayValue = value.map((possibleValue, i) => {
+                let conjunction = '';
+                if (value.length === 2 && i === 0) conjunction = 'either ';
+                if (i === value.length - 1) conjunction = 'or ';
+                let objCType = `${prefix}${camelize(layerType)}${suffix}${camelize(property.name)}`;
+                return `${conjunction}\`${objCType}${camelize(possibleValue)}\``;
+              }).join(separator);
+            } else {
+              let objCType = `${prefix}${camelize(layerType)}${suffix}${camelize(property.name)}`;
+              displayValue = `\`${objCType}${camelize(value)}\``;
+            }
+            return `an \`NSValue\` object containing ${displayValue}`;
         case 'color':
             let color = parseColor(value);
             if (!color) {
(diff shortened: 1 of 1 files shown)


davidtheclark:
Modified the text based on chat feedback from @1ec5. Added semicolons between required properties and removed "either" if there are more than 2 possibilities. Now the doc for `icon-text-fit-padding` for `icon-text-fix` values would look like this ...

... with 2 possible values:
> This property is only applied to the style if `iconImage` is non-`nil`; and `textField` is non-`nil`; and `iconTextFit` is set to either an `NSValue` object containing `MGLSymbolStyleLayerIconTextFitBoth` or an `NSValue` object containing `MGLSymbolStyleLayerIconTextFitWidth`. Otherwise, it is ignored.

... with 3 possible values (the real thing):
> This property is only applied to the style if `iconImage` is non-`nil`; and `textField` is non-`nil`; and `iconTextFit` is set to an `NSValue` object containing `MGLSymbolStyleLayerIconTextFitBoth`, an `NSValue` object containing `MGLSymbolStyleLayerIconTextFitWidth`, or an `NSValue` object containing `MGLSymbolStyleLayerIconTextFitHeight`. Otherwise, it is ignored.

davidtheclark:
Had another idea for reducing verbosity, by cutting the repetition of "an NSValue object containing". After latest commit, here's the example output ...

... with 2 possible values:

> This property is only applied to the style if `iconImage` is non-`nil`; and `textField` is non-`nil`; and `iconTextFit` is set to an `NSValue` object containing either `MGLSymbolStyleLayerIconTextFitBoth` or `MGLSymbolStyleLayerIconTextFitHeight`. Otherwise, it is ignored.

... with 3 possible values:

>  This property is only applied to the style if `iconImage` is non-`nil`; and `textField` is non-`nil`; and `iconTextFit` is set to an `NSValue` object containing `MGLSymbolStyleLayerIconTextFitBoth`, `MGLSymbolStyleLayerIconTextFitWidth`, or `MGLSymbolStyleLayerIconTextFitHeight`. Otherwise, it is ignored.

platform/darwin/scripts/generate-style-code.js
```diff
@@ -107,9 +107,9 @@ global.propertyReqs = function (property, layoutPropertiesByName, type) {
             return '`' + camelizeWithLeadingLowercase(req['!']) + '` is set to `nil`';
         } else {
             let name = Object.keys(req)[0];
-            return '`' + camelizeWithLeadingLowercase(name) + '` is set to ' + describeValue(req[Object.keys(req)[0]], layoutPropertiesByName[name], type);
+            return '`' + camelizeWithLeadingLowercase(name) + '` is set to ' + describeValue(req[name], layoutPropertiesByName[name], type);
         }
-    }).join(', and ') + '. Otherwise, it is ignored.';
+    }).join('; and ') + '. Otherwise, it is ignored.';
```
1ec5: This semicolon might be excessive if none of the requirements includes a comma.

davidtheclark: I can remove that.

davidtheclark:
@1ec5 I reverted the semicolon to a comma and rebased & squashed into a single commit.

There is and should be no changes to the generated documentation *yet* — not until the a new spec version is released with the change discussed in mapbox/mapbox-gl-style-spec#504.

platform/darwin/scripts/generate-style-code.js
1ec5: Oh, a semicolon is definitely helpful if a requirement does include a comma, though. It isn’t wrong either way, since you’ve simplified the verbiage a bit, just a stylistic choice.

1ec5:
Oh, one more thing: if you don’t mind rebasing again, can you prepend `[ios, macos] ` to the commit message? Thanks!

friedbunny:
(Or you could add that to the commit message when you squash merge.)

1ec5:
> There is and should be no changes to the generated documentation _yet_ — not until the a new spec version is released with the change discussed in mapbox/mapbox-gl-style-spec#504.

We update the gl-native project frequently between official specification releases, so the corresponding changes are being handled in mapbox/mapbox-gl-native#6508.

PR #6449, opened 2016-09-24
[Qt] QML runtime style 2.0

brunoabinader:
Implements Qt QML runtime style via `MapParameter`, as part of future integration with Qt upstream. Now, we only expose two different QML types: `QQuickMapboxGL` and `QQuickMapboxGLMapParameter`. The latter is used for specifying Mapbox-specific features:

Every map parameter starts with a property `name` - which is used to identify the parameter type. Accepted parameter types are: `style`, `paint`, `layout`, `layer`, `source`, `filter`, `image`, `bearing` and `pitch` - examples below:

### Style: `style`
Properties:
- `url`: Specifies the style URL e.g. `"mapbox://styles/mapbox/streets-v9"`

### Style paint properties: `paint`
Properties:
- `layer`: Specifies the style layer.
- `class` (optional): Specifies the style class.
Style paint properties are then defined in camelCase e.g. `fill-color` becomes `fillColor`:
``` qml
property var fillColor: Qt.rgba(1, 1, 1, 1)
``` 

### Style layout properties: `layout`
Properties:
- `layer`: Specifies the style layer.
Style layout properties are then defined in camelCase e.g. `line-join` becomes `lineJoin`.

### Style layer: `layer`
Properties:
- `name`: Specifies the layer name (aka. `id`).
- `layerType`: Specifies the layer type e.g. `line`.
- `source`: Specifies the layer source.

### Style source: `source`
Properties:
- `name`: Specifies the source name (aka. `id`).
- `sourceType`: Specifies the layer type e.g. `geojson`.
- `data`: Specifies the source file path.

### Style filter: `filter`
Properties:
- `layer`: Specifies the associated style layer.
- `filter`: Specifies the filter as a list of strings e.g.:
``` qml
[ "==", "$type", "Point" ]
```

### Style image: `image`
Properties:
- `name`: Specifies the image name.
- `sprite`: Specifies the image file path. `QQuickImage` items are not supported because these are not exposed via public API.

### Map bearing: `bearing`
Properties:
- `angle`: Specifies the map bearing angle (in degrees).

### Map bearing: `pitch`
Properties:
- `angle`: Specifies the map pitch angle (in degrees).


brunoabinader:
Qt QML example:
QML example

platform/qt/include/qquickmapboxgl.hpp
```diff
@@ -34,10 +34,6 @@ class Q_DECL_EXPORT QQuickMapboxGL : public QQuickFramebufferObject
     Q_PROPERTY(bool copyrightsVisible READ copyrightsVisible WRITE setCopyrightsVisible NOTIFY copyrightsVisibleChanged)
     Q_PROPERTY(QColor color READ color WRITE setColor NOTIFY colorChanged)
 
-    // MapboxGL QML Type interface.
-    Q_PROPERTY(qreal bearing READ bearing WRITE setBearing NOTIFY bearingChanged)
-    Q_PROPERTY(qreal pitch READ pitch WRITE setPitch NOTIFY pitchChanged)
```
tmpsantos: you forgot to remove the signals:
```
    // Mapbox-specific signals.
    void bearingChanged(qreal angle);
    void pitchChanged(qreal angle);
```

platform/qt/src/qquickmapboxgl.cpp
```diff
@@ -182,132 +203,254 @@ void QQuickMapboxGL::pan(int dx, int dy)
...
+    if (needsUpdate) update();
 }
 
-QPointF QQuickMapboxGL::swapPan()
+void QQuickMapboxGL::onParameterPropertyUpdated(const QString &propertyName)
 {
-    QPointF oldPan = m_pan;
+    QQuickMapboxGLMapParameter *param = qobject_cast<QQuickMapboxGLMapParameter *>(sender());
+    if (!param) {
+        qWarning() << "Warning: Sender for QQuickMapboxGL::onParameterPropertyUpdated is not a QQuickMapboxGLMapParameter.";
```
tmpsantos: Error API.

platform/qt/src/qquickmapboxgl.cpp
```diff
@@ -182,132 +203,254 @@ void QQuickMapboxGL::pan(int dx, int dy)
...
+        return;
+    }
 
-    m_pan = QPointF(0, 0);
+    if (propertyName == "type") {
+        qWarning() << "Error: Cannot modify property 'type'.";
```
tmpsantos: Ditto, let's use the error API.

platform/qt/src/qquickmapboxgl.cpp
```diff
@@ -182,132 +203,254 @@ void QQuickMapboxGL::pan(int dx, int dy)
...
+        return;
+    }
 
-    return oldPan;
+    bool needsUpdate = false;
+    switch (s_parameterTypes.indexOf(param->property("type").toString())) {
+    case -1:
+        qWarning() << "Error: Invalid value for property 'type'.";
```
tmpsantos: Ditto.

platform/qt/src/qquickmapboxgl.cpp
```diff
@@ -182,132 +203,254 @@ void QQuickMapboxGL::pan(int dx, int dy)
...
+    if (sourceType == "vector" || sourceType == "raster") {
+        source["url"] = param->property("url");
+        m_sourceChanges << source;
+    } else if (sourceType == "geojson") {
+        QFile geojson(param->property("data").toString());
+        geojson.open(QIODevice::ReadOnly);
+        source["data"] = geojson.readAll();
+        m_sourceChanges << source;
+    } else {
+        qWarning() << "Error: Unsupported source type '" << sourceType << "'.";
```
tmpsantos: I see this a nice fit to the error API we inherited from the QML Map. Also I find this message a bit confusing for the end user.

```
Error { NoError, NotSupportedError, UnknownParameterError, MissingRequiredParameterError, ConnectionError }
```

platform/qt/src/qquickmapboxgl.cpp
```diff
@@ -182,132 +203,254 @@ void QQuickMapboxGL::pan(int dx, int dy)
...
 }
 
-qreal QQuickMapboxGL::pitch() const
+void QQuickMapboxGL::processMapParameter(QQuickMapboxGLMapParameter *param)
 {
-    return m_pitch;
+    bool needsUpdate = false;
+    switch (s_parameterTypes.indexOf(param->property("type").toString())) {
+    case -1:
+        qWarning() << "Error: Invalid value for property 'type'.";
```
tmpsantos: Ditto.

platform/qt/qt.cmake
```diff
@@ -57,7 +57,7 @@ include_directories(
 
 # C++ app
 add_executable(mbgl-qt
-    platform/qt/app/app.qrc
+    platform/qt/common.qrc
```
brunoabinader: Thanks, removed `platform/qt/qmlapp/source.geojson` and moved `common.qrc` to `platform/qt/resources` folder, removing the need for an `alias` on QRC.

platform/qt/qmlapp/main.qml
```diff
@@ -61,51 +61,122 @@ ApplicationWindow {
...
+                            property var name: "routeCase"
+                            property var layerType: "line"
+                            property var source: "routeSource"
                         },
                         MapParameter {
                             property var type: "paint"
-                            property var layer: "testLayer"
-                            property var fillColor: Qt.rgba(0, 0, 1, 1)
+                            property var layer: "routeCase"
+                            property color lineColor: "white"
```
brunoabinader: Indeed - I must have confused this with some other error I had at the time I was implementing this - good to know Qt is smart enough on this ;)

PR #6438, opened 2016-09-23
[android] - remove logging from snapshot builds

tobrun:
WIP - snapshot and release aar files contain logging

PR #6452, opened 2016-09-26
[android] - disable general purpose logging during transformation change

tobrun:
closes #6438.
Proposal to disable logs that are currently executed when users perform gestures.
These logs were written both in Release and Debug mode and we have received request to disable them. While this is not a long term solution it enables us to ship in short term. I'm suggesting to rethink this for the `5.0.0` refactor/release.

cc @cammace @ivovandongen @zugaldia


tobrun:
Changes have been implemented in 9c01cf0

PR #6456, opened 2016-09-26
6402 update offline metadata android

### LINKED ISSUE #6402: Allow changing offline metadata
`OfflineRegion` supports getting the region’s metadata via `byte[] getMetadata()` but not setting it. An offline region’s metadata typically contains a name. An application that allows the user to choose the name would typically also allow the user to change the name.

The corresponding iOS ticket is in https://github.com/mapbox/mapbox-gl-native/issues/4344.

/cc: @1ec5

tobrun:
closes #6402 

Provides Android binding integration in the core feature of updating metadata.

ezgif com-video-to-gif 62

Hasn't been added to a specific milestone (needs discussion-label).

Review @zugaldia

### DIFF
--- a/platform/android/MapboxGLAndroidSDKTestApp/src/main/java/com/mapbox/mapboxsdk/testapp/activity/offline/OfflineUtils.java
+++ b/platform/android/MapboxGLAndroidSDKTestApp/src/main/java/com/mapbox/mapboxsdk/testapp/activity/offline/OfflineUtils.java
@@ -1,0 +1,42 @@
+package com.mapbox.mapboxsdk.testapp.activity.offline;
+
+import android.support.annotation.NonNull;
+import android.util.Log;
+
+import com.mapbox.mapboxsdk.offline.OfflineRegion;
+
+import org.json.JSONObject;
+
+import static com.mapbox.mapboxsdk.testapp.activity.offline.OfflineActivity.JSON_CHARSET;
+import static com.mapbox.mapboxsdk.testapp.activity.offline.OfflineActivity.JSON_FIELD_REGION_NAME;
+
+class OfflineUtils {
+
+    static String getRegionName(@NonNull  OfflineRegion offlineRegion) {
+        String regionName;
+        try {
+            byte[] metadata = offlineRegion.getMetadata();
+            String json = new String(metadata, JSON_CHARSET);
+            JSONObject jsonObject = new JSONObject(json);
+            regionName = jsonObject.getString(JSON_FIELD_REGION_NAME);
+        } catch (Exception exception) {
+            Log.e("OfflineUtils", "Failed to decode metadata: " + exception.getMessage());
+            regionName = "Region " + offlineRegion.getID();
+        }
+        return regionName;
+    }
+
+    static byte[] convertRegionName(String regionName) {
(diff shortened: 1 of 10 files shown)


platform/android/MapboxGLAndroidSDKTestApp/src/main/java/com/mapbox/mapboxsdk/testapp/activity/offline/OfflineUtils.java
```diff
@@ -0,0 +1,42 @@
...
+import com.mapbox.mapboxsdk.offline.OfflineRegion;
+
+import org.json.JSONObject;
+
+import static com.mapbox.mapboxsdk.testapp.activity.offline.OfflineActivity.JSON_CHARSET;
+import static com.mapbox.mapboxsdk.testapp.activity.offline.OfflineActivity.JSON_FIELD_REGION_NAME;
+
+class OfflineUtils {
+
+    static String getRegionName(@NonNull  OfflineRegion offlineRegion) {
```
zugaldia: Let's make this method symmetric with `byte[] convertRegionName(String regionName)` below and change its signature to `String getRegionName(@NonNull byte[] metadata)`. We don't need the `OfflineRegion` reference here. After all, this is just a generic util to wrap from/to `String` <-> `JSONObject` <-> `byte[]`.

tobrun:
20facaf fixed the requested changes

platform/android/MapboxGLAndroidSDKTestApp/src/main/java/com/mapbox/mapboxsdk/testapp/utils/OfflineUtils.java
```diff
@@ -0,0 +1,41 @@
+package com.mapbox.mapboxsdk.testapp.utils;
+
+import android.support.annotation.NonNull;
+import android.util.Log;
+
+import com.mapbox.mapboxsdk.offline.OfflineRegion;
```
zugaldia: This can be dropped now.

platform/android/MapboxGLAndroidSDKTestApp/src/main/java/com/mapbox/mapboxsdk/testapp/utils/OfflineUtils.java
```diff
@@ -0,0 +1,41 @@
...
+
+    public static String convertRegionName(@NonNull  byte[] metadata) {
+        String regionName;
+        try {
+            String json = new String(metadata, JSON_CHARSET);
+            JSONObject jsonObject = new JSONObject(json);
+            regionName = jsonObject.getString(JSON_FIELD_REGION_NAME);
+        } catch (Exception exception) {
+            Log.e("OfflineUtils", "Failed to decode metadata: " + exception.getMessage());
+            regionName = "Region x";
```
zugaldia: For consistency with `convertRegionName` let's return a `null` here too if this fails and let the `UpdateMetadataActivity` handle it (what you had with `"Region: %d ", OfflineRegion.getId()` was fine as `Region x` will look like the same region is duplicated if several regions fail here).

tobrun:
I rebased fixes in d5112cb

PR #6457, opened 2016-09-26
[Qt] Simplify QML example

tmpsantos:
It is going to be part of our documentation, so make it KISS as possible.

### DIFF
--- a/platform/qt/qmlapp/main.qml
+++ b/platform/qt/qmlapp/main.qml
@@ -2,8 +2,8 @@ import QtGraphicalEffects 1.0
 import QtPositioning 5.0
 import QtQuick 2.0
 import QtQuick.Controls 1.0
-import QtQuick.Layouts 1.0
 import QtQuick.Dialogs 1.0
+import QtQuick.Layouts 1.0
 
 import QQuickMapboxGL 1.0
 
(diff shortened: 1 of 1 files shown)


platform/qt/qmlapp/main.qml
```diff
@@ -13,417 +13,290 @@ ApplicationWindow {
...
+        anchors.right: parent.right
+        anchors.top: parent.top
+        anchors.margins: 30
 
-                    value: 0
-                    minimumValue: 0
-                    maximumValue: 60
-                }
+        Slider {
+            id: bearingSlider
```
brunoabinader: Let's get advantage of this opportunity to put some labels on these sliders.

platform/qt/qmlapp/main.qml
```diff
@@ -13,417 +13,290 @@ ApplicationWindow {
...
+            anchors.left: parent.left
+            anchors.right: parent.right
+            value: 0
+            minimumValue: 0
+            maximumValue: 180
+        }
 
-                    Layout.fillHeight: true
-                    orientation: Qt.Vertical
+        Slider {
```
brunoabinader: Ditto.

tmpsantos:
@brunoabinader thanks for reviewing. I also made the style a radio group.

screenshot from 2016-09-26 18-08-24

PR #6461, opened 2016-09-26
Don't load tiles if no layer requires them

kkaefer:
As a follow-up to #6373, this change stops our renderer from requesting tiles from sources that don't have a layer that actually requires it. Currently, if a source has been loaded before, but no layer is enabled due to the current zoom level, we are still happily loading all of the tiles even thought they are never used.


kkaefer:
Please review

Issue #5960, opened 2016-08-11
MGLSource and subclasses need documentation


1ec5:
Per https://github.com/mapbox/mapbox-gl-native/pull/5727#discussion_r74305314 and https://github.com/mapbox/mapbox-gl-native/pull/5727#issuecomment-238978850, MGLSource and its symbols, as well as each of MGLSource’s subclasses, need documentation comments. This may require #5954.

/cc @frederoni @friedbunny

1ec5:
Until this documentation is added, developers are going to get confused by MGLRasterStyle’s `URL` property, thinking it’s for tile sets (#6245).

boundsj:
https://github.com/mapbox/mapbox-gl-native/pull/6316 adds the ability to create raster and vector sources with a new `MGLTileSet` object. The documentation in `MGLTileSet` plus the new initializers in `MGLRasterSource` and `MGLVectorSource` may help mitigate confusion until documentation for the sources lands.

Issue #5959, opened 2016-08-11
MGLBaseStyleLayer needs documentation


1ec5:
Per https://github.com/mapbox/mapbox-gl-native/pull/5727#discussion_r74301831, MGLBaseStyleLayer and its `visible` property are missing documentation comments.

/cc @frederoni

incanus:
We need to hit the base style layer subclasses as well (fill, line, etc.). Relies on https://github.com/mapbox/mapbox-gl-native/issues/5954.

boundsj:
> We need to hit the base style layer subclasses

@incanus note that this started in https://github.com/mapbox/mapbox-gl-native/pull/6107

incanus:
Subclasses handled in ticket at https://github.com/mapbox/mapbox-gl-native/issues/6467 for clarity.

PR #6466, opened 2016-09-26
flesh out iOS runtime styling docs

incanus:
Hitting #5959 and #5960. 

/cc @boundsj @1ec5 @frederoni

### DIFF
--- a/platform/darwin/src/MGLBaseStyleLayer.h
+++ b/platform/darwin/src/MGLBaseStyleLayer.h
@@ -8,7 +8,9 @@ NS_ASSUME_NONNULL_BEGIN
  The base style layer class from which all other layer types inherit. 
  Style layers allow runtime customization of all map styling properties. 
  
- You should use the concrete subclasses of MGLBaseStyleLayer (which conform to MGLStyleLayer) to style fill, line, symbol, and other layer types. 
+ You should use the concrete subclasses of MGLBaseStyleLayer (which 
+ conform to MGLStyleLayer) to style fill, line, symbol, and other layer 
+ types.
  */
 @interface MGLBaseStyleLayer : NSObject
 
--- a/platform/darwin/src/MGLBaseStyleLayer.h
+++ b/platform/darwin/src/MGLBaseStyleLayer.h
@@ -4,17 +4,24 @@
 
 NS_ASSUME_NONNULL_BEGIN
 
+/**
+ The base style layer class from which all other layer types inherit. 
+ Style layers allow runtime customization of all map styling properties. 
+ */
 @interface MGLBaseStyleLayer : NSObject
 
+/**
+ The display of the layer. A value of `NO` hides the layer.
+ */
 @property (nonatomic, assign, getter=isVisible) BOOL visible;
 
 /**
- The maximum zoom level on which the layer gets parsed and appears on.
+ The maximum zoom level at which the layer gets parsed and appears.
  */
 @property (nonatomic, assign) float maximumZoomLevel;
 
 /**
- The minimum zoom level on which the layer gets parsed and appears on.
+ The minimum zoom level at which the layer gets parsed and appears.
  */
 @property (nonatomic, assign) float minimumZoomLevel;
 
--- a/platform/darwin/src/MGLBaseStyleLayer.h
+++ b/platform/darwin/src/MGLBaseStyleLayer.h
@@ -5,11 +5,12 @@
 NS_ASSUME_NONNULL_BEGIN
 
 /**
- The base style layer class from which all other layer types inherit. 
- Style layers allow runtime customization of all map styling properties. 
+ The base style layer class from which all other style layer classes 
+ inherit. Style layers allow runtime customization of all map styling 
+ properties.
  
- You should use the concrete subclasses of MGLBaseStyleLayer (which 
- conform to MGLStyleLayer) to style fill, line, symbol, and other layer 
+ You should use the concrete subclasses of `MGLBaseStyleLayer` (which
+ conform to `MGLStyleLayer`) to style fill, line, symbol, and other layer 
  types.
  */
 @interface MGLBaseStyleLayer : NSObject
--- a/platform/darwin/src/MGLBaseStyleLayer.h
+++ b/platform/darwin/src/MGLBaseStyleLayer.h
@@ -7,6 +7,8 @@ NS_ASSUME_NONNULL_BEGIN
 /**
  The base style layer class from which all other layer types inherit. 
  Style layers allow runtime customization of all map styling properties. 
+ 
+ You should use the concrete subclasses of MGLBaseStyleLayer (which conform to MGLStyleLayer) to style fill, line, symbol, and other layer types. 
  */
 @interface MGLBaseStyleLayer : NSObject
 
--- a/platform/darwin/src/MGLRasterSource.h
+++ b/platform/darwin/src/MGLRasterSource.h
@@ -7,14 +7,51 @@
 
 NS_ASSUME_NONNULL_BEGIN
 
+/**
+ A raster tile source.
+ 
+ @see <a href="https://www.mapbox.com/mapbox-gl-style-spec/#sources-raster">The
+ style specification.</a>
+ */
 @interface MGLRasterSource : MGLSource
 
+/**
+ A URL to a TileJSON resource. Supported protocols are `http:`, `https:`, and 
+ `mapbox://<mapid>`.
+ */
 @property (nonatomic, readonly, copy) NSURL *URL;
+
+/**
+ The minimum visual size to display tiles for this source. 
+ */
 @property (nonatomic, readonly, assign) NSUInteger tileSize;
+
+/**
+ A tile set holds the raster tile URL template strings and associated
+ configuration for those strings. It can be passed in place of the URL
+ to TileJSON in order to create a source configured to download tiles
+ from ordinary web URLs.
+ */
 @property (nonatomic, readonly, nullable) MGLTileSet *tileSet;
 
+/**
+ Initializes a source with the given identifier, TileJSON configuration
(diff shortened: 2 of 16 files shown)


platform/darwin/src/MGLBaseStyleLayer.h
```diff
@@ -4,17 +4,24 @@
 
 NS_ASSUME_NONNULL_BEGIN
 
+/**
+ The base style layer class from which all other layer types inherit. 
+ Style layers allow runtime customization of all map styling properties. 
+ */
 @interface MGLBaseStyleLayer : NSObject
```
boundsj: I think it'd be a good idea to, like we do for `MGLMultiPoint`, mention that this is an abstract class and that subclasses should be used in practice. Similarly, we should document that using the properties of this class is intended.

incanus: da949f56c51ff39d3ca0826e2198f793a7a719d7

incanus: da949f56c51ff39d3ca0826e2198f793a7a719d7

platform/darwin/src/MGLBaseStyleLayer.h
```diff
@@ -4,17 +4,28 @@
 
 NS_ASSUME_NONNULL_BEGIN
 
+/**
+ The base style layer class from which all other layer types inherit.
```
1ec5: Nit: s/layer types/style layer classes/

platform/darwin/src/MGLBaseStyleLayer.h
```diff
@@ -4,17 +4,28 @@
 
 NS_ASSUME_NONNULL_BEGIN
 
+/**
+ The base style layer class from which all other layer types inherit. 
+ Style layers allow runtime customization of all map styling properties. 
+ 
+ You should use the concrete subclasses of MGLBaseStyleLayer (which
```
1ec5: Use backticks so jazzy can automatically link to relevant documentation.

platform/darwin/src/MGLRasterSource.h
```diff
@@ -1,18 +1,57 @@
...
+#import <CoreGraphics/CoreGraphics.h>
+
 @class MGLTileSet;
 
 NS_ASSUME_NONNULL_BEGIN
 
+/**
+ A raster tile source.
+ 
+ @see <a href="https://www.mapbox.com/mapbox-gl-style-spec/#sources-raster">The
```
1ec5: This is fine for now, but ideally we would copy the relevant documentation here, adapt it to Cocoa terminology, and not even need to link to the style specification.

platform/darwin/src/MGLRasterSource.h
```diff
@@ -1,18 +1,57 @@
...
+ style specification.</a>
+ */
 @interface MGLRasterSource : MGLSource
 
+/**
+ A URL to a TileJSON resource. Supported protocols are `http:`, `https:`, and
```
1ec5: We should link to the TileJSON specification, in case the developer wants to author their own.

platform/darwin/src/MGLRasterSource.h
```diff
@@ -1,18 +1,57 @@
...
 @interface MGLRasterSource : MGLSource
 
+/**
+ A URL to a TileJSON resource. Supported protocols are `http:`, `https:`, and 
+ `mapbox://<mapid>`.
+ */
 @property (nonatomic, readonly, copy) NSURL *URL;
+
+/**
+ The minimum visual size to display tiles for this source.
```
1ec5: Is this in points? Pixels? Is the same size used for width and height?

incanus:
Since these are just straight documentation adds, unlike https://github.com/mapbox/mapbox-gl-native/issues/6467 which is blocked by some Jazzy hacking in https://github.com/mapbox/mapbox-gl-native/issues/5954, what say we squash & land this to hit both https://github.com/mapbox/mapbox-gl-native/issues/5959 & https://github.com/mapbox/mapbox-gl-native/issues/5960 immediately, then tackle the harder work in a new PR? 

@boundsj @1ec5

platform/darwin/src/MGLRasterSource.h
```diff
@@ -1,18 +1,57 @@
...
+ */
 @property (nonatomic, readonly, copy) NSURL *URL;
+
+/**
+ The minimum visual size to display tiles for this source. 
+ */
 @property (nonatomic, readonly, assign) NSUInteger tileSize;
+
+/**
+ A tile set holds the raster tile URL template strings and associated
```
1ec5: Start off with "The tile set that the source uses to locate and present tiles." Or something like that. Just to make the relationship to the class clear.

boundsj:
> what say we squash & land this to hit both #5959 & #5960 immediately, then tackle the harder work in a new PR?

I think that is fine. Although, I'm probably missing something because I think that we could "just" add the minimal documentation for the layer subclasses (class and initializer documentation) so that the classes show up in jazzy docs. This probably means updating the `platform/darwin/src/MGLStyleLayer.h.ejs` template and adding a javascript function so that each subclass gets its documentation. Then property overrides to make each properties documentation more correct for the darwin platform could land later.

platform/darwin/src/MGLRasterSource.h
```diff
@@ -1,18 +1,57 @@
...
+ */
 - (instancetype)initWithSourceIdentifier:(NSString *)sourceIdentifier URL:(NSURL *)url tileSize:(CGFloat)tileSize;
 
+/**
+ Initializes a source with the given identifier, tile size, and tile 
+ URL template set.
+
+ @param sourceIdentifier A string that uniquely identifies the source.
+ @param tileSize The minimum visual size to display tiles for the source.
+ @param tileSet A tile set describing where to download tiles.
```
1ec5: Maybe these two parameters should be swapped, to better parallel the other initializer.

platform/darwin/src/MGLSource.h
```diff
@@ -1,8 +1,15 @@
 #import <Foundation/Foundation.h>
-#import <CoreGraphics/CoreGraphics.h>
 
+/**
+ A source supplies data to be shown on the map. Sources don't contain styling details like color or width. Use style layers to give visual representation to sources.
```
1ec5: Refer to "subclasses of `MGLStyleLayer`" instead of "style layer" so developers can easily navigate in the documentation.

platform/darwin/src/MGLSource.h
```diff
@@ -1,8 +1,15 @@
...
+ You should use the concrete subclasses of MGLSource to create vector, raster, GeoJSON, and other source types.
+ */
 @interface MGLSource : NSObject
 
+/**
+ A string that uniquely identifies the source.
+ */
 @property (nonatomic, copy) NSString *sourceIdentifier;
 
 - (instancetype)initWithSourceIdentifier:(NSString *)sourceIdentifier;
```
1ec5: This initializer needs documentation too.

platform/ios/jazzy.yml
```diff
@@ -51,6 +51,21 @@ custom_categories:
       - MGLPolygonFeature
       - MGLPolylineFeature
       - MGLShapeCollectionFeature
+  - name: Runtime Styling
```
1ec5: Runtime styling is an (intentionally) awkward public name for a feature. "Style Layers" and "Data Sources", as two separate categories, would better differentiate these classes from the other categories.

incanus:
Ah, yeah, I don't know the intricacies of the template stuff yet but I'll take a peek and hit here if so.

incanus:
@boundsj I think 44f30c73db3e94e7c04bcc2007e8c301e86bfb81 takes care of what we need at a base level to get style layer subclasses to show up.

platform/darwin/src/MGLBaseStyleLayer.h
incanus: a6429e5e9879551cb4f016dba2e2ce5eeb5a63ff

platform/darwin/src/MGLBaseStyleLayer.h
incanus: a6429e5e9879551cb4f016dba2e2ce5eeb5a63ff

platform/darwin/src/MGLRasterSource.h
incanus: 7fdca0df8537b8c2ff54bdbc06eaf63bce0332f2

platform/darwin/src/MGLRasterSource.h
incanus: 7be4bdb5073212e6fe79e57ca5cb0a074dd7143c

platform/darwin/src/MGLRasterSource.h
incanus: 47ceca8d730816fbe13572c6f8de784d03665559

platform/darwin/src/MGLRasterSource.h
incanus: 0597834cddc70c645236327f5f31c6227a119fb8

platform/darwin/src/MGLSource.h
incanus: b53d14bf709151be1487dfcc4b2562d8685ad559

platform/darwin/src/MGLSource.h
incanus: 965e50acffd1f8b165ddf2dea2043f36476c160a

platform/ios/jazzy.yml
incanus: a1ed5697a93ef25298817cd6b3e9e2bc4812aa8e

incanus:
Ok I think I hit all of @1ec5's requests, so this is again ready to go.

Issue #5731, opened 2016-07-20
Crash rendering the map — mbgl::FillBucket::drawElements, glrAGXRenderVertexArray


MikeKasperlik:
**Platform:** iOS
**Mapbox SDK version:** 3.3.1

We see a high crash rate for this on fabric.

I can't reproduce this crash but it happens randomly when the app is in foreground and doing something with the map.

We use a custom style & draw a single polyline on the map

```
Crashed: com.apple.main-thread
0  libsystem_platform.dylib       0x181e5a2a0 _platform_memmove + 176
1  AGXGLDriver                    0x193d93f6c glrAGXRenderVertexArray(GLDContextRec*, unsigned int, unsigned int, int, int, unsigned int, void const*, int, void const*)
2  AGXGLDriver                    0x193d93f6c glrAGXRenderVertexArray(GLDContextRec*, unsigned int, unsigned int, int, int, unsigned int, void const*, int, void const*)
3  GLEngine                       0x194b36ea4 glDrawElements_ACC_ES2Exec
4  Mapbox                         0x1017739e4 mbgl::FillBucket::drawElements(mbgl::PlainShader&, mbgl::gl::ObjectStore&) + 284
5  Mapbox                         0x10177dd7c mbgl::Painter::renderFill(mbgl::FillBucket&, mbgl::style::FillLayer const&, mbgl::UnwrappedTileID const&, std::__1::array<double, 16ul> const&) + 3852
6  Mapbox                         0x10177abb0 void mbgl::Painter::renderPass<std::__1::reverse_iterator<std::__1::__wrap_iter<mbgl::RenderItem const*> > >(mbgl::RenderPass, std::__1::reverse_iterator<std::__1::__wrap_iter<mbgl::RenderItem const*> >, std::__1::reverse_iterator<std::__1::__wrap_iter<mbgl::RenderItem const*> >, int, signed char) + 688
7  Mapbox                         0x10177a244 mbgl::Painter::render(mbgl::style::Style const&, mbgl::FrameData const&, mbgl::SpriteAtlas&) + 1424
8  Mapbox                         0x1017682ac mbgl::Map::Impl::render() + 160
9  Mapbox                         0x10176811c mbgl::Map::render() + 128
10 Mapbox                         0x1016ec2a8 -[MGLMapView glkView:drawInRect:] + 48
11 GLKit                          0x18a89136c -[GLKView _display:] + 332
[...]

PR #6468, opened 2016-09-27
C++ bindings for OpenGL

jfirebaugh:
This PR introduces mid-level bindings for OpenGL, specifically a procedural API that wraps a single primitive draw operation, with all state affecting the result provided via method parameters rather than the OpenGL default of implicit state and bindings. The primary design reference for this API was [glium](http://tomaka.github.io/glium/glium/index.html), with secondary inspiration from [regl](http://regl.party/api).

This parameters for a draw operation are the following:

* **Program**: A compiled vertex and fragment shader.
* **Parameters**: General OpenGL state such as depth test, stencil test, and blend mode. The variety of OpenGL methods for configuring the depth test, stencil test, and blend mode have been rationalized with an algebraic data type for each of [`Depth`](https://github.com/mapbox/mapbox-gl-native/blob/d45bca40523b4e187bf5bc8acd7c280e2a180ec8/src/mbgl/gl/parameters.hpp#L9-L29), [`Stencil`](https://github.com/mapbox/mapbox-gl-native/blob/d45bca40523b4e187bf5bc8acd7c280e2a180ec8/src/mbgl/gl/parameters.hpp#L31-L83), and [`Blend`](https://github.com/mapbox/mapbox-gl-native/blob/d45bca40523b4e187bf5bc8acd7c280e2a180ec8/src/mbgl/gl/parameters.hpp#L85-L136).
* **AttributeSources**: A collection of sources of vertex attribute data. A source can be a vertex buffer or a slice of one (`glVertexAttribPointer`), or a constant scalar or vector value (`glVertexAttrib*`). (It remains to be seen if we use the latter for DDS or go the route of GL JS and generate shader source code that uses attributes or uniforms as appropriate.) A buffer or buffer slice comes with a specification of its layout.
* **Uniforms**: A collection of uniform locations and values. Values are typed, and valid types include scalars and vectors as well as texture objects. For the latter, the implementation handles activating an unused texture unit.
* **Primitives**: The specification of what geometry to draw. This concept unifies unindexed (`glDrawArrays`) and indexed (`glDrawElements`) drawing. Input can be array buffer or element buffer objects or slices thereof. Buffer types are templated on their geometry, e.g. `ElementBuffer<Triangles, uint16_t>`.

Type safety is generally provided at the level of individual parameters, and notably _not_ at the level of matching types between shader source and uniform or attribute types, or between attribute types and C++ object layout. Some additional safety can be obtained by writing higher-level bindings atop this API, e.g. along the lines of the [API for uniforms proposed by @kkaefer here](https://github.com/mapbox/mapbox-gl-native/commit/119100bd069e7b2baa925d46bff50ef1a0eb3701#diff-0f0fd078dcbda458c5b1ca0ebb108ff8L85).

Likewise, we can layer a declarative, object-based API more like regl atop this one, as suggested in [this gl-js issue](https://github.com/mapbox/mapbox-gl-js/issues/145). Doing so may simplify the communication pathway between workers and the renderer, and provide a clearer path to use VAOs transparently than with a purely procedural approach.

### TODOs

The remaining items are numerous; I'll edit to flesh out this section.

* [x] Probably the first item is to land the unification of `gl::ObjectStore` and `gl::Config` (#6470). The `prepare` and `draw` functions I've sketched here should be methods on this object, and refer to its internal state to avoid no-op OpenGL operations as `gl::Config` currently does.

cc @kkaefer @lucaswoj


include/mbgl/gl/gl_values.hpp
```diff
@@ -76,7 +76,7 @@ struct DepthMask {
     }
 };
 
-struct ColorMask {
+struct ColorMaskValue {
```
jfirebaugh: Renames here to avoid conflicts with new type names. I envision that all the structs in gl_values.hpp will eventually go away.

kkaefer: The goal of these values, and their surrounding `Value` objects is to prevent duplicate state from being set. It seems like this is no longer the case with this code? Will it come back as work progresses?

src/mbgl/renderer/painter_clipping.cpp
```diff
@@ -5,41 +5,44 @@
...
+                         gl::Stencil::Replace
+                     },
+                     gl::Blend::disabled(),
+                     gl::ColorMask::disabled()
+                 },
+                 gl::AttributeSources {},
+                 gl::Uniforms {
+                     gl::UniformMatrix<4> { program.u_matrix, matrix }
+                 },
+                 tileStencilBuffer);
```
kkaefer: How does this draw call know that it should draw triangles?

src/mbgl/renderer/painter_clipping.cpp
```diff
@@ -5,41 +5,44 @@
...
+                     gl::Stencil {
+                         gl::Stencil::Always(),
+                         GLint(clip.reference.to_ulong()),
+                         0b11111111,
+                         gl::Stencil::Keep,
+                         gl::Stencil::Keep,
+                         gl::Stencil::Replace
+                     },
+                     gl::Blend::disabled(),
+                     gl::ColorMask::disabled()
```
kkaefer: Do all of these have to be specified? Can we use some defaults here?

src/mbgl/renderer/painter_clipping.cpp
```diff
@@ -5,41 +5,44 @@
...
+                 },
+                 gl::AttributeSources {},
+                 gl::Uniforms {
+                     gl::UniformMatrix<4> { program.u_matrix, matrix }
+                 },
```
kkaefer: Is there a way to ensure that all uniforms in a program can be set? In the past, we've seen issues with not all uniforms for a draw call being set so we didn't have predictable results.

src/mbgl/gl/attributes.hpp
```diff
@@ -0,0 +1,154 @@
...
+template <typename T> struct AttributeType;
+
+template <> struct AttributeType<int8_t>   : std::integral_constant<GLenum, GL_BYTE> {};
+template <> struct AttributeType<uint8_t>  : std::integral_constant<GLenum, GL_UNSIGNED_BYTE> {};
+template <> struct AttributeType<int16_t>  : std::integral_constant<GLenum, GL_SHORT> {};
+template <> struct AttributeType<uint16_t> : std::integral_constant<GLenum, GL_UNSIGNED_SHORT> {};
+template <> struct AttributeType<int32_t>  : std::integral_constant<GLenum, GL_INT> {};
+template <> struct AttributeType<uint32_t> : std::integral_constant<GLenum, GL_UNSIGNED_INT> {};
+template <> struct AttributeType<float>    : std::integral_constant<GLenum, GL_FLOAT> {};
+template <> struct AttributeType<double>   : std::integral_constant<GLenum, GL_DOUBLE> {};
```
kkaefer: At some point, we should hide the `GL_*`, and `gl*` symbols so users of this library don't need to include `gl.h`.

src/mbgl/gl/attributes.hpp
```diff
@@ -0,0 +1,154 @@
...
+    ConstantVector<4, float>,
+    ConstantVector<2, double>,
+    ConstantVector<3, double>,
+    ConstantVector<4, double>
+    */
+    >;
+
+using AttributeSources = std::vector<AttributeSource>;
+
+namespace detail {
```
kkaefer: Ideally, we'd move all of the `detail` code that calls actual GL functions to an implementation file

src/mbgl/gl/parameters.hpp
```diff
@@ -0,0 +1,234 @@
...
+       return Blend { Replace(), {} };
+    }
+};
+
+class ColorMask {
+public:
+    bool r;
+    bool g;
+    bool b;
+    bool a;
```
kkaefer: For types like this, we should either default them, or write explicit initializers that require all parameters to prevent accidental misuse, which results in uninitialized values. Valgrind may catch them, but it's always better to have a compiler error slapped in the face.

lucaswoj:
This is excellent work @jfirebaugh! I am excited to iterate on GL Native and GL JS rendering architecture together as we work on https://github.com/mapbox/mapbox-gl-js/issues/145.

src/mbgl/renderer/painter_clipping.cpp
jfirebaugh: Yes, we could allow defaults. However, given that part of the motivation for moving to this API is to eliminate errors caused by forgetting to specify the desired state, I suggest that we start by making all parameters required. We can define additional helpers in the vein of `gl::*::disabled()` to package up common values. For example I anticipate we'll want `gl::Blend::alphaTransparency()` and functions for the most common depth and stencil settings.

src/mbgl/renderer/painter_clipping.cpp
jfirebaugh: Yes, I think we can accomplish this with a higher-level API that layers on top of this one.

src/mbgl/renderer/painter_clipping.cpp
jfirebaugh: I was thinking that vertex buffer types would be tagged with a primitive type, but in some cases we want to reuse vertexes in multiple primitive types. So I think this API will probably need to be revised to accept vertex sources and index source in separate parameters, with a marker value for the latter to indicate unindexed drawing (e.g. `Unindexed<Triangles>()`). This is what the glium API does.

include/mbgl/gl/gl_values.hpp
jfirebaugh: Yep, I was just waiting for #6470 to land.

src/mbgl/gl/attributes.hpp
jfirebaugh: I think this is doable. It means that all polymorphic input types have to be wrapped in a `variant` at some point -- no full monomorphization via template methods. I think that's fine, it was already mostly that way.

jfirebaugh: Or did you just mean move them to an `_impl.hpp` file? I assumed you meant move them to a `.cpp` file.

src/mbgl/gl/parameters.hpp
jfirebaugh: clang gives a warning if you omit parameters in brace initialization:

```
warning: missing field 'b' initializer [-Wmissing-field-initializers]
        return ColorMask { false, false };
                                        ^
```

This will be an error once we fix #6487.

src/mbgl/renderer/painter_clipping.cpp
kkaefer: Yeah, in some cases we're using the same buffer for different draw operations.

src/mbgl/gl/attributes.hpp
kkaefer: Yeah, `.cpp` file.

src/mbgl/gl/parameters.hpp
brunoabinader: For consecutive `bool` values we could save some memory space by using [bit field](http://en.cppreference.com/w/cpp/language/bit_field) e.g.:
``` c++
bool r : 1;
bool g : 1;
bool b : 1;
bool a : 1;
```

jfirebaugh:
This is finally in good shape and (hopefully) passing CI -- it's ready for another round of review.

Major changes since the initial prototype:

* @kkaefer merged `gl::ObjectStore` and `gl::Config`, which means that this PR can now use all the `State` values in `gl::Config` to avoid no-op GL calls.
* I refactored how we work with GL buffers, and rebuilt this PR to use `gl::VertexBuffer` and `gl::IndexBuffer`.
* We must draw _segments_ of the vertex and index buffers when dealing with more vertices than can fit in a `uint16_t` index, and therefore most draw calls involve a loop over these segments. This loop is now pushed down into `Context::draw`.
* `Context::draw` now accepts a single argument of type `gl::Drawable`. This `Drawable` class holds all state required for the draw call. The `Drawable` constructor does static verification -- like checking that the shader type, vertex type, and uniform values all match up -- and then stores type-erased values internally. This allows `Context::draw` to be template-free and entirely defined in the `.cpp` file.
* This in turn removes the `gl.hpp` dependency from more files.
* Use of vertex array objects is now handled internally and transparently in the `Context::draw` implementation. Clients don't have to worry about creating/binding/verifying VAOs.

jfirebaugh:
@kkaefer Rebased on top of #6596 and ready for re-review.

src/mbgl/renderer/painter_debug.cpp
```diff
@@ -5,162 +5,59 @@
...
-    plainShader.u_opacity = 1.0f;
-
-    // draw tile outline
-    tileBorderArray.bind(plainShader, tileLineStripVertexBuffer, BUFFER_OFFSET_0, context);
-    plainShader.u_color = { 1.0f, 0.0f, 0.0f, 1.0f };
-    context.lineWidth = 4.0f * frame.pixelRatio;
-    MBGL_CHECK_ERROR(glDrawArrays(GL_LINE_STRIP, 0, static_cast<GLsizei>(tileLineStripVertexBuffer.vertexCount)));
-}
-
-#ifndef NDEBUG
```
jfirebaugh: I removed `Painter::renderClipMasks` and `Painter::renderDepthBuffer` in an early step to avoid dealing with them at the time, and need to go back and convert them now.

src/mbgl/gl/attribute.hpp
```diff
@@ -27,10 +25,8 @@ class AttributeBinding {
           type(DataTypeOf<T>::value),
           count(N),
           offset(O) {
-        static_assert(std::is_standard_layout<Vertex>::value, "vertex type must use standard layout");
```
jfirebaugh: This one moved to [the `Drawable` constructor](https://github.com/mapbox/mapbox-gl-native/blob/ca28e8319522829dc9315f4c94d92d5a759ceb9c/src/mbgl/gl/drawable.hpp#L105). I just dropped "vertex type is too big", because it was kind of silly.

src/mbgl/gl/context.cpp
```diff
@@ -327,11 +262,189 @@ void Context::setDirtyState() {
...
+}
+
+void Context::setColor(const Color& color) {
+    if (color.blendFunction.is<Color::Replace>()) {
+        blend = false;
+    } else {
+        blend = true;
+        blendColor = color.blendColor;
+        apply_visitor([&] (const auto& blendFunction) {
+            // TODO: blendEquation = blendFunction.equation;
```
jfirebaugh: Previous code didn't call `glBlendEquation` anywhere and we always want the default of `GL_FUNC_ADD`. But we should set it explicitly right? In case a custom layer changes it.

src/mbgl/gl/context.cpp
```diff
@@ -327,11 +262,189 @@ void Context::setDirtyState() {
...
+
+    setDepth(drawable.depth);
+    setStencil(drawable.stencil);
+    setColor(drawable.color);
+
+    program = drawable.program;
+
+    drawable.bindUniforms();
+
+    for (const auto& segment : drawable.segments) {
```
jfirebaugh: I think it'll be short circuited prior to getting here, but I'll add an internal guard as well.

src/mbgl/gl/context.cpp
```diff
@@ -327,11 +262,189 @@ void Context::setDirtyState() {
...
+            VertexArrayID id = 0;
+            MBGL_CHECK_ERROR(gl::GenVertexArrays(1, &id));
+            vertexArrayObject = id;
+            vaos.emplace(vaoKey, UniqueVertexArray(std::move(id), { this }));
+
+            return true;
+        };
+
+        if (needAttributeBindings()) {
+            vertexBuffer.setDirty();
```
jfirebaugh: It's for the case where a VAO is being initialized. VAOs don't inherit the existing buffer bindings; we need to ensure that `glBindBuffer` actually gets called. But it can be moved up into the `needAttributeBindings` lambda and commented. I'll do so.

src/mbgl/gl/mode.hpp
```diff
@@ -0,0 +1,34 @@
...
+
+using Mode = variant<
+    Points,
+    Lines,
+    LineStrip,
+    Triangles,
+    TriangleStrip>;
+
+} // namespace gl
+} // namespace mbgl
```
jfirebaugh: `PrimitiveMode`?

src/mbgl/gl/drawable.hpp
```diff
@@ -0,0 +1,124 @@
...
+          attributeBindings(AttributeBindings<Shader, typename Subject::VertexType>()(shader))
+    {
+        static_assert(std::is_standard_layout<typename Subject::VertexType>::value, "vertex type must use standard layout");
+        static_assert(std::is_same<typename Shader::VertexType, typename Subject::VertexType>::value, "vertex type mismatch");
+    }
+
+    Mode mode;
+    Depth depth;
+    Stencil stencil;
+    Color color;
```
jfirebaugh: I chose `Color` for parallelism with `Depth` and `Stencil` -- each holds the configuration that affects the respective GL buffer attachment of that type. For `Color`, that's both blend settings and the color mask.

src/mbgl/shader/circle_uniforms.hpp
```diff
@@ -0,0 +1,21 @@
+#pragma once
+
+#include <mbgl/shader/uniforms.hpp>
+
+namespace mbgl {
+
+MBGL_DEFINE_UNIFORM_SCALAR(float, u_radius);
```
jfirebaugh: Where multiple shaders use the same uniform name, I've hoisted the definition to [a shared file](https://github.com/mapbox/mapbox-gl-native/blob/ca28e8319522829dc9315f4c94d92d5a759ceb9c/src/mbgl/shader/uniforms.hpp).

src/mbgl/shader/circle_uniforms.hpp
```diff
@@ -0,0 +1,21 @@
...
+    u_matrix,
+    u_opacity,
+    u_color,
+    u_radius,
+    u_blur,
+    u_scale_with_map,
+    u_extrude_scale,
+    u_devicepixelratio> {};
+
+} // namespace mbgl
```
jfirebaugh: There's not a one-to-one relationship -- e.g. `FillShader`, `FillPatternShader`, `FillOutlineShader`, `FillOutlinePatternShader`. I think there's some consolidation to be done -- perhaps replacing explicit shader subclasses with a single `Shader` template -- but I'd like to defer it to a future PR.

src/mbgl/gl/context.cpp
```diff
@@ -327,11 +262,189 @@ void Context::setDirtyState() {
...
+            auto it = vaos.find(vaoKey);
+            if (it != vaos.end()) {
+                vertexArrayObject = it->second;
+                return false;
+            }
+
+            VertexArrayID id = 0;
+            MBGL_CHECK_ERROR(gl::GenVertexArrays(1, &id));
+            vertexArrayObject = id;
+            vaos.emplace(vaoKey, UniqueVertexArray(std::move(id), { this }));
```
jfirebaugh: Well, the static design is the previously existing one, where each buffer `Segment` holds a corresponding VAO, and binding is done explicitly. [That approach seems to be error-prone.](https://github.com/mapbox/mapbox-gl-native/issues/5731)

src/mbgl/shader/circle_uniforms.hpp
kkaefer: Yeah, I've seen that; I'm referring to the situation when two shaders use the same name for different uniform types.

src/mbgl/gl/mode.hpp
kkaefer: Either work for me. Maybe adding the `Mode` to all avoids ambiguity?

src/mbgl/shader/circle_uniforms.hpp
jfirebaugh: Ok -- there aren't any cases of that currently, and I think we should avoid introducing any. Besides link errors, it would just be confusing to use the same uniform name for different types.

kkaefer: Okay; but it still feels weird to have them in the main `mbgl` namespace so we have symbols like `mbgl::u_radius`. Can we at least move them to another namespace?

kkaefer:
Can you add the `[core]` prefix to the two last commits?

jfirebaugh:
Squashed one, prefixed the other.

I think it's prudent for this _not_ to go into iOS v3.4.0 or Android v4.2.0, as it's an invasive change. iOS is on a release branch already and not planning any more merges from `master`. @mapbox/android, do you have plans to create a release branch soon, so that this PR can be merged to `master`?

tobrun:
@jfirebaugh feel free to merge to master, we have a release branch called `release-android-v4.2.0`.

kkaefer:
🎉

PR #6477, opened 2016-09-27
[Qt] Header cleanup

tmpsantos:
Make some headers private, so we only commit to the QML API.


tmpsantos:
:eyes: @brunoabinader

PR #6470, opened 2016-09-27
Merge gl::ObjectStore and gl::Config

kkaefer:
These are currently two distinct objects, but they're almost always used together. Passing multiple objects around just makes the code longer and less readable. I'm not seeing any gains from keeping them separated.


kkaefer:
There are a few tasks that aren't necessarily part of this PR:

* [ ] move gl::debugging into gl::Context
* [ ] hide `gl*` calls in implementation files
* [ ] cleanup `gl.hpp` and `gl.cpp` and move various functions to their own headers

jfirebaugh:
Oh, didn't look at Qt5 macOS failure, that's something more serious:

```
[ RUN      ] Annotations.UpdateSymbolAnnotationIcon
mbgl-test(11337,0x7fff7acf4000) malloc: *** error for object 0x7f9a1bcc8180: incorrect checksum for freed object - object was probably modified after being freed.
*** set a breakpoint in malloc_error_break to debug
make: *** [run-qt-test-*] Abort trap: 6
```

kkaefer:
Yeah, I think we were destroying the `Context` before the `AnnotationManager`.

PR #6480, opened 2016-09-27
Optionable textureview

tobrun:
This PR allows to use the older TextureView approach in favor of SurfaceView.
This feature is annotated with deprecation, it could possibly be removed in future releases.

Review @zugaldia

### DIFF
--- a/platform/android/MapboxGLAndroidSDKTestApp/src/main/res/layout/activity_marker_view.xml
+++ b/platform/android/MapboxGLAndroidSDKTestApp/src/main/res/layout/activity_marker_view.xml
@@ -18,7 +18,6 @@
         android:layout_below="@id/toolbar"
         app:center_latitude="38.907192"
         app:center_longitude="-77.036871"
-        app:texture_mode="true"
         app:style_url="@string/style_mapbox_streets"
         app:zoom="12" />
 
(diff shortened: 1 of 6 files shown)


zugaldia:
Looking good. We're gonna need to add Javadoc first to make CI happy.

tobrun:
I still had to migrate the Hungarian notation, this is now fixed with dcf5a1a, feel free to review.

platform/android/MapboxGLAndroidSDKTestApp/src/main/res/layout/activity_marker_view.xml
```diff
@@ -18,6 +18,7 @@
         android:layout_below="@id/toolbar"
         app:center_latitude="38.907192"
         app:center_longitude="-77.036871"
+        app:texture_mode="true"
```
zugaldia: @tobrun Is it intended to leave this activity and `PolygonActivity` with a `TextureView` by default? If so, we need to clearly say it. A better way might be to a) have a specific `TextureViewActivity` example or b) leave the `SurfaceView` as the default in both activities, but add a clear comment that with that line you'd enable the now deprecated `TextureView`.

tobrun: I added explicitly a deprecation comment in PolygonActivity ( this should be obvious for end users hence the line is als strikethrough). I removed it from the MarkerViewActivity since it was less obvious.

zugaldia:
cc: @danswick : This happened.

Issue #6458, opened 2016-09-26
Custom annotation views disappear after scrolling map down and then return back


aksonov:
**Platform:** iOS
**Mapbox SDK version:** latest

### Steps to trigger behavior

 1. Create custom annotation class
```
@interface CustomAnnotation : MGLAnnotationView<MGLAnnotation>

@property (nonatomic, readonly) CLLocationCoordinate2D coordinate;

@end

@implementation CustomAnnotation

-(id)init {
    self = [super init];
    self.frame = CGRectMake(0, 0, 300, 300);
    self.backgroundColor = [UIColor blackColor];
    UILabel *label = [[UILabel alloc] initWithFrame:CGRectMake(0, 0, 300, 300)];
    label.text = @"TEST";
    [self addSubview:label];
    return self;
}

-(CLLocationCoordinate2D)coordinate {
    CLLocationCoordinate2D coords;
    coords.latitude = 37.33233141;
    coords.longitude = -122.0312186;
    return coords;
}
```
2. Add custom annotation to map:
```
- (nullable MGLAnnotationView *)mapView:(MGLMapView *)mapView viewForAnnotation:(id <MGLAnnotation>)annotation {
  return annotation;
}

- (void)viewDidLoad {
  [super viewDidLoad];
  NSURL *styleURL = [NSURL URLWithString:@"mapbox://styles/mapbox/streets-v8"];
  
  MGLMapView *mapView = [[MGLMapView alloc] initWithFrame:self.view.bounds styleURL:styleURL];
  
  mapView.autoresizingMask = UIViewAutoresizingFlexibleWidth | UIViewAutoresizingFlexibleHeight;
  mapView.delegate = self;
  
  // Set the map’s center coordinate and zoom level.
  [mapView setCenterCoordinate:CLLocationCoordinate2DMake(37.33233141, -122.0312186)
                     zoomLevel:17
                      animated:NO];
  [mapView addAnnotation:[[CustomAnnotation alloc] init]];
  [self.view addSubview:mapView];
}
```

3. Run, saw annotation successfully, but it disappears once scroll it away and then scroll back

### Expected behavior
Custom annotation still visible

### Actual behaviour
Annotation disappears

Am I doing something wrong?

boundsj:
[...]

PR #6485, opened 2016-09-27
[ios] Always show annotation view even when view reuse is not used

boundsj:
This fixes an issue noted in https://github.com/mapbox/mapbox-gl-native/issues/6458#issuecomment-249808011 where the map view would drop annotation views
if the developer did not make full use of the reuse queue. 

The fixes are:

- Guard against enqueuing a view for reuse if it does not have a `viewReuseIdentifier`. Previously, if the app developer created a new annotation view with no reuse id each time `[MGLMapViewDelegate mapView:viewForAnnotation:]` was called and then the view was scrolled offscreen, we  attempted to add the view to the reuse queue (for no good reason) and  did not update the views center ever again.  This change avoids the reuse queue and just updates the
center always.

- Add any annotation view that should be displayed but is not in the annotation container view to the annotation container view. Previously, if the app developer created a new annotation view with a reuse id each time `[MGLMapViewDelegate mapView:viewForAnnotation:]` was called,
and then if the annotation was scrolled off and then on screen again, we did not add the new view to the container view because it was not happening in the normal `addAnnotation` cycle. This adds add logic to catch that case and add annotation views created in this way to the view.

Nonte that it is still advisable to use the resuse queue as shown in our [annotation view example](https://www.mapbox.com/ios-sdk/examples/annotation-views). Also, once https://github.com/mapbox/mapbox-gl-native/issues/6055 lands, we will be able to clean this up even more. But, for now, this restores the unoptimized annotation view functionality that I think we were unintentionally not supporting.

cc @incanus @1ec5 @frederoni


boundsj:
Thanks for the reminder. I updated the changelog in 4e0a026

PR #6479, opened 2016-09-28
Add .test.cpp suffix to all test case files

kkaefer:
I'm frequently running into the issue of using my editor's quick open feature and selecting the wrong file (e.g. `test/map/map.cpp` instead of `src/mbgl/map/map.cpp`). Can we change the test case filenames to indicate that they contain tests only?

/cc @jfirebaugh @tmpsantos @brunoabinader @mikemorris


kkaefer:
Waiting for more feedback given that I changed this from a `test_` prefix to a `.test` suffix.

mikemorris:
👍 The `.test.cpp` suffix matches [a pattern we were already using for Node.js tests](https://github.com/mapbox/mapbox-gl-native/tree/master/platform/node/test).

PR #6495, opened 2016-09-28
[core] Inherit from mapbox::geometry::point<double> in {LatLng,ProjectedMeters}

brunoabinader:
Inherit from `mapbox::geometry::point<double>` makes it possible for `mbgl::utill::{LatLng,ProjectedMeters}` to reuse the arithmetic and logic operators.

### DIFF
--- a/test/util/geo.test.cpp
+++ b/test/util/geo.test.cpp
@@ -6,6 +6,11 @@
 
 using namespace mbgl;
 
+TEST(LatLng, MixedCoordinateTypes) {
+    static_assert(std::is_assignable<LatLng, ScreenCoordinate>::value == false, "LatLng and ScreenCoordinate cannot be interchangeable.");
+    static_assert(std::is_assignable<LatLng, ProjectedMeters>::value == false, "LatLng and ProjectedMeters cannot be interchangeable.");
+}
+
 TEST(LatLngBounds, World) {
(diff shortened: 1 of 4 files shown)


test/util/geo.test.cpp
```diff
@@ -6,6 +6,11 @@
 
 using namespace mbgl;
 
+TEST(LatLng, MixedCoordinateTypes) {
+    static_assert(std::is_assignable<LatLng, ScreenCoordinate>::value == false, "LatLng and ScreenCoordinate cannot be interchangeable.");
+    static_assert(std::is_assignable<LatLng, ProjectedMeters>::value == false, "LatLng and ProjectedMeters cannot be interchangeable.");
+}
```
jfirebaugh: Try it the other way around -- assigning a `LatLng` to a `ScreenCoordinate` variable, or passing a `LatLng` to a function that expects `ScreenCoordinate`.

brunoabinader: @jfirebaugh true - there is no simple way of forbidding implicit conversions like this besides deleting the base class assignment operator, which is not desired.

brunoabinader:
Closing as per comment from @jfirebaugh that this change is error prone in terms that `ScreenCoordinate` and `LatLng` shouldn't be allowed to be interoperable.

PR #6497, opened 2016-09-28
LocalFileSource

### LINKED ISSUE #6273: [android] Using asset:// URL with path outside of APK
**Platform:** Android
**Mapbox SDK version:** react-native-mapbox-gl 5.0.0 which uses mapbox-android-sdk 4.1.1

### Issue

The only way to access local files on Android via `asset://` is if it's bundled with the APK.  Since you can't modify the APK from the app, Android currently doesn't support assets that have been downloaded after the app is installed.

On iOS, I've been able to specify a path where files were downloaded onto the user's device, so no issues there.

### My Use Case

In the app I'm developing, a user has to log in.  Afterwards, the app will use their account information to generate the user's style and tilejson files along with downloading a set of vector tiles.

On iOS, I can point the `asset://` URL to these generated and downloaded files, and everything works great.  But on Android, there's not a way to shove these files into the signed APK programatically.

Am I missing something obvious to make this use case work on Android?  Or am I hampered by `asset://` only allowing access to pre-bundled APK assets?

ivovandongen:
Fixes #6273

### DIFF
--- a/platform/default/default_file_source.cpp
+++ b/platform/default/default_file_source.cpp
@@ -1,5 +1,6 @@
 #include <mbgl/storage/default_file_source.hpp>
 #include <mbgl/storage/asset_file_source.hpp>
+#include <mbgl/storage/local_file_source.hpp>
 #include <mbgl/storage/online_file_source.hpp>
 #include <mbgl/storage/offline_database.hpp>
 #include <mbgl/storage/offline_download.hpp>
@@ -164,7 +165,8 @@ DefaultFileSource::DefaultFileSource(const std::string& cachePath,
                                      uint64_t maximumCacheSize)
     : thread(std::make_unique<util::Thread<Impl>>(util::ThreadContext{"DefaultFileSource", util::ThreadPriority::Low},
             cachePath, maximumCacheSize)),
-      assetFileSource(std::make_unique<AssetFileSource>(assetRoot)) {
+      assetFileSource(std::make_unique<AssetFileSource>(assetRoot)),
+      localFileSource(std::make_unique<LocalFileSource>()) {
 }
 
 DefaultFileSource::~DefaultFileSource() = default;
@@ -203,6 +205,8 @@ std::unique_ptr<AsyncRequest> DefaultFileSource::request(const Resource& resourc
 
     if (isAssetURL(resource.url)) {
         return assetFileSource->request(resource, callback);
+    } else if (LocalFileSource::acceptsURL(resource.url)) {
+        return localFileSource->request(resource, callback);
     } else {
         return std::make_unique<DefaultFileRequest>(resource, callback, *thread);
     }
--- a/platform/default/local_file_source.cpp
+++ b/platform/default/local_file_source.cpp
@@ -1,0 +1,66 @@
+#include <mbgl/storage/local_file_source.hpp>
+#include <mbgl/storage/response.hpp>
+#include <mbgl/util/string.hpp>
+#include <mbgl/util/thread.hpp>
+#include <mbgl/util/url.hpp>
+#include <mbgl/util/util.hpp>
+#include <mbgl/util/io.hpp>
+
+#include <sys/types.h>
+#include <sys/stat.h>
+#include <unistd.h>
+
+namespace {
+
+const char* protocol = "file://";
+const std::size_t protocolLength = 7;
+    
+} // namespace
+
+namespace mbgl {
+
+class LocalFileSource::Impl {
+public:
+    void request(const std::string& url, FileSource::Callback callback) {
+        //Cut off the protocol
+        std::string path = mbgl::util::percentDecode(url.substr(protocolLength));
+
+        Response response;
+
+        struct stat buf;
+        int result = stat(path.c_str(), &buf);
+
+        if (result == 0 && S_ISDIR(buf.st_mode)) {
+            response.error = std::make_unique<Response::Error>(Response::Error::Reason::NotFound);
+        } else if (result == -1 && errno == ENOENT) {
+            response.error = std::make_unique<Response::Error>(Response::Error::Reason::NotFound);
+        } else {
+            try {
+                response.data = std::make_shared<std::string>(util::read_file(path));
+            } catch (...) {
+                response.error = std::make_unique<Response::Error>(
+                    Response::Error::Reason::Other,
+                    util::toString(std::current_exception()));
+            }
+        }
+
+        callback(response);
+    }
+
+};
+
+LocalFileSource::LocalFileSource()
+    : thread(std::make_unique<util::Thread<Impl>>(util::ThreadContext{"LocalFileSource", util::ThreadPriority::Low})) {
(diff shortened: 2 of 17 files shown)


ivovandongen:
@tmpsantos Could you review please (assuming the ci completes)?

I'm not very happy with the name `FileFileSource`, but couldn't come up with a better one.

1ec5:
> I'm not very happy with the name `FileFileSource`, but couldn't come up with a better one.

How about `LocalFileSource`?

1ec5:
With these changes, is it still necessary for the default `AssetFileSource` to know how to handle absolute paths? Also, [this iOS/macOS special case](https://github.com/mapbox/mapbox-gl-native/blob/1777b8757a437d6f27928c2bb3d821fc8679cf20/platform/darwin/src/NSURL%2BMGLAdditions.m#L10-L17) can be removed now that file: URLs are supported in core.

brunoabinader:
> How about LocalFileSource?

I like this one - though we should be aware that the [file URI scheme](https://en.wikipedia.org/wiki/File_URI_scheme) can also point to files on a network location.

tmpsantos:
> How about LocalFileSource?

:+1:

ivovandongen:
> How about LocalFileSource?

Seems popular :) I've changed this.

> is it still necessary for the default AssetFileSource to know how to handle absolute paths? 

I assumed we need to leave it in there now for backwards compatibility. 

> this iOS/macOS special case can be removed now that file: URLs are supported in core.

Removed

platform/default/default_file_source.cpp
```diff
@@ -19,6 +20,12 @@ bool isAssetURL(const std::string& url) {
     return std::equal(assetProtocol.begin(), assetProtocol.end(), url.begin());
 }
 
+const std::string fileProtocol = "file://";
```
tmpsantos: Better avoid this. See for alternatives: https://github.com/mapbox/mapbox-gl-native/pull/6455

test/storage/local_file_source.test.cpp
```diff
@@ -0,0 +1,101 @@
...
+#include <mbgl/platform/platform.hpp>
+#include <mbgl/util/run_loop.hpp>
+
+#include <unistd.h>
+#include <limits.h>
+#include <gtest/gtest.h>
+
+namespace {
+
+std::string toAbsoluteUrl(const std::string& fileName) {
```
tmpsantos: Nit: we use URL on the internal bits rather then Url.

test/storage/local_file_source.test.cpp
```diff
@@ -0,0 +1,101 @@
...
+    char buff[PATH_MAX + 1];
+    char* cwd = getcwd( buff, PATH_MAX + 1 );
+    return "file:://" + std::string(cwd) + "/test/fixtures/storage/assets/" + fileName;
```
tmpsantos: `://`

platform/default/local_file_source.cpp
```diff
@@ -0,0 +1,55 @@
...
+#include <sys/stat.h>
+#include <unistd.h>
+
+namespace mbgl {
+
+class LocalFileSource::Impl {
+public:
+    void request(const std::string& url, FileSource::Callback callback) {
+        //Cut off the protocol
+        std::string path = mbgl::util::percentDecode(url.substr(7));
```
tmpsantos: Make this like #6455 with the size a global together with the scheme variable.

test/storage/local_file_source.test.cpp
```diff
@@ -0,0 +1,101 @@
...
+        EXPECT_EQ(Response::Error::Reason::NotFound, res.error->reason);
+        ASSERT_FALSE(res.data.get());
+        // Do not assert on platform-specific error message.
+        loop.stop();
+    });
+
+    loop.run();
+}
+
+TEST(LocalFileSource, URLEncoding) {
```
tmpsantos: For the sake we are no missing anything, can you make a test with a extremely large file name, like `memset(filename, 'x', PATH_MAX);`

Wondering what would happen if you try to read something you don't have permission to (like `/root/foo.txt`).

ivovandongen:
@tmpsantos Thanks for the review. I've cleared up all items.

>Wondering what would happen if you try to read something you don't have permission to 

It returns `Response::Error::Reason::Other`. Not quite sure how to test this in a platform independent way though.

tmpsantos:
Core part looks fine to me. /cc @tobrun for the Android bits.

platform/default/local_file_source.cpp
```diff
@@ -0,0 +1,63 @@
...
+#include <mbgl/util/util.hpp>
+#include <mbgl/util/io.hpp>
+
+#include <sys/types.h>
+#include <sys/stat.h>
+#include <unistd.h>
+
+namespace mbgl {
+    
+const char* protocol = "file://";
```
tmpsantos: This could go to the anonymous namespace.

ivovandongen: Thanks! Changed it.

platform/default/local_file_source.cpp
```diff
@@ -0,0 +1,62 @@
...
+#include <mbgl/util/thread.hpp>
+#include <mbgl/util/url.hpp>
+#include <mbgl/util/util.hpp>
+#include <mbgl/util/io.hpp>
+
+#include <sys/types.h>
+#include <sys/stat.h>
+#include <unistd.h>
+
+const char* protocol = "file://";
```
tmpsantos:
```
namespace {

const char* protocol = "file://";
const std::size_t protocolLength = 7;

} // namespace
```

1ec5:
>> is it still necessary for the default AssetFileSource to know how to handle absolute paths?

> I assumed we need to leave it in there now for backwards compatibility. 

True, removing that support might (?) break asset: assets that have been cached, I suppose, and it’s pretty harmless.

ivovandongen:
@tmpsantos I fixed the valgrind errors. If all else is good, could you press the review button?

tobrun:
:shipit:

bigsassy:
I just wanted to say to @ivovandongen, thank you!  Oh, and also @1ec5 and the rest of the mapbox team as well 👍

zugaldia:
cc: @cammace this looks like a good candidate for a new sample, or a required update to https://www.mapbox.com/android-sdk/examples/custom-raster/.

SeanChristopherConway:
I am using ios SDK 3.4.0-beta.6 and am not able to use the file:// style to load my style JSON or in the tiles section of my sources in the style.json.  When I use asset:// for the style it loads, but neither asset nor file seem to work within the styles section.

1ec5:
It should be possible to use an _absolute_ file: URL as an MGLMapView’s style URL. But I’m not sure if it’s supposed to be possible to use a file: URL inside the JSON file itself. In any event, this closed PR likely isn’t being watched by folks who know the answer for sure, so please open a new issue, ideally with more details about your style JSON.

PR #6499, opened 2016-09-28
Pull OpenGL extension loading into their own headers

kkaefer:
They're currently all in the gl.hpp/cpp files.

### DIFF
--- a/cmake/core-files.cmake
+++ b/cmake/core-files.cmake
@@ -81,6 +81,8 @@ set(MBGL_CORE_FILES
     src/mbgl/gl/context.hpp
     src/mbgl/gl/debugging.cpp
     src/mbgl/gl/debugging.hpp
+    src/mbgl/gl/extension.cpp
+    src/mbgl/gl/extension.hpp
     src/mbgl/gl/gl.cpp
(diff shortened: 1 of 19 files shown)


cmake/core-files.cmake
```diff
@@ -81,12 +81,16 @@ set(MBGL_CORE_FILES
...
+    src/mbgl/gl/extension.cpp
+    src/mbgl/gl/extension.hpp
     src/mbgl/gl/gl.cpp
     src/mbgl/gl/object.cpp
     src/mbgl/gl/object.hpp
     src/mbgl/gl/state.hpp
     src/mbgl/gl/value.cpp
     src/mbgl/gl/value.hpp
+    src/mbgl/gl/vertex_array.cpp
+    src/mbgl/gl/vertex_array.hpp
```
jfirebaugh: Let's merge `src/mbgl/geometry/vao.{hpp,cpp}` into these files.

kkaefer: For the time being, I'd like to have all of the files within the `gl` namespace independent of the rest. We're going to refactor the VAO object as well at some point to remove the dependency on other code in MBGL.

PR #6508, opened 2016-09-29
update style generation code for new style spec enum values docs

incanus:
Downstream work for https://github.com/mapbox/mapbox-gl-style-spec/pull/510. 

- [x] core
- [x] Darwin
- [x] Android

/cc @1ec5 @frederoni @ivovandongen


incanus:
Only thing I'm not crazy about with https://github.com/mapbox/mapbox-gl-native/pull/6508/commits/e9be66b52fb78cf2c00511d1e256d1f0a0557f3d is failure to fully get `NS_ENUM` values, e.g.: 

```objc
typedef NS_ENUM(NSUInteger, MGLSymbolStyleLayerIconRotationAlignment) {
    /**
     When `symbolPlacement` is set to `point`, aligns icons east-west. When `symbolPlacement` is set to `line`, aligns icon x-axes with the line.
     */
    MGLSymbolStyleLayerIconRotationAlignmentMap,
    /**
     Produces icons whose x-axes are aligned with the x-axis of the viewport, regardless of the value of `symbolPlacement`.
     */
    MGLSymbolStyleLayerIconRotationAlignmentViewport,
    /**
     When `symbolPlacement` is set to `point`, this is equivalent to `viewport`. When `symbolPlacement` is set to `line`, this is equivalent to `map`.
     */
    MGLSymbolStyleLayerIconRotationAlignmentAuto,
};
```

Note the ``point``, ``line``, ``viewport``, etc. 

Do we get this nitpicky?

1ec5:
> Do we get this nitpicky? 

[Yes](https://github.com/mapbox/mapbox-gl-native/blob/1777b8757a437d6f27928c2bb3d821fc8679cf20/platform/darwin/scripts/generate-style-code.js#L81-L90). The existing logic is robust enough to handle values of the property currently being described (first-person values), but not enough to handle values of another property being referenced (third-person values).

1ec5:
We can maybe detect constructions such as “x is set to y” and find candidate values accordingly.

Getting this right will ensure that jazzy can autolink the referenced symbols. I think autolinking is important for enabling users to effectively navigate our documentation, which is spread out among many headers or many webpages.

platform/android/src/style/conversion/types_string_values.hpp.ejs
```diff
@@ -31,7 +31,8 @@ namespace conversion {
     //<%- property.name %>
     inline std::string toString(mbgl::style::<%- propertyNativeType(property) %> value) {
         switch (value) {
-<% for (const value of property.values) { -%>
+<% for (const value in property.values) { -%>
+          //<%- property.values[value].doc %>
```
ivovandongen: This is a left-over comment?

incanus:
@ivovandongen Curiious about your feedback again now after https://github.com/mapbox/mapbox-gl-native/pull/6508/commits/0c8cc787a56dd59284e96a4b5fa7d257fff2b7e7 and https://github.com/mapbox/mapbox-gl-native/pull/6508/commits/c58f90abb2b47486d40dc187d98f40a65cef6f96.

platform/android/src/style/conversion/types_string_values.hpp.ejs
incanus: https://github.com/mapbox/mapbox-gl-native/pull/6508/commits/0c8cc787a56dd59284e96a4b5fa7d257fff2b7e7

incanus:
Working on some final Darwin-side `propertyDoc()` enhancements still. I may try something similar on the Android side after, but considering we didn't even have `enum` values documented at all in Java until https://github.com/mapbox/mapbox-gl-native/pull/6508/commits/c58f90abb2b47486d40dc187d98f40a65cef6f96, this may not be high priority.

platform/darwin/src/MGLCircleStyleLayer.h
```diff
@@ -6,13 +6,31 @@
 
 NS_ASSUME_NONNULL_BEGIN
 
+/**
+ Controls the translation reference point.
```
1ec5: Since the enums are on a different page than the properties that use them, perhaps this documentation could mention those properties.

platform/darwin/src/MGLSymbolStyleLayer.h
```diff
@@ -6,67 +6,199 @@
...
+     */
     MGLSymbolStyleLayerSymbolPlacementLine,
 };
 
+/**
+ In combination with `symbolPlacement`, determines the rotation behavior of icons.
+ */
 typedef NS_ENUM(NSUInteger, MGLSymbolStyleLayerIconRotationAlignment) {
+    /**
+     When `symbolPlacement` is set to `point`, aligns icons east-west. When `symbolPlacement` is set to `line`, aligns icon x-axes with the line.
```
1ec5: Hopefully we can rely on this “`x` is set to `y`” phrase and expand `y` to the right enum value.

platform/darwin/src/MGLSymbolStyleLayer.h
```diff
@@ -178,7 +310,7 @@ typedef NS_ENUM(NSUInteger, MGLSymbolStyleLayerTextTranslateAnchor) {
  
  The default value of this property is an `NSValue` object containing `NSEdgeInsetsZero` or `UIEdgeInsetsZero`. Set this property to `nil` to reset it to the default value.
  
- This property is only applied to the style if `iconImage` is non-`nil`, and `iconTextFit` is non-`nil`, and `textField` is non-`nil`. Otherwise, it is ignored.
+ This property is only applied to the style if `iconImage` is non-`nil`, and `textField` is non-`nil`, and `iconTextFit` is set to an `NSValue` object containing `MGLSymbolStyleLayerIconTextFitBoth`, `MGLSymbolStyleLayerIconTextFitWidth`, or `MGLSymbolStyleLayerIconTextFitHeight`. Otherwise, it is ignored.
```
1ec5: 👍 This takes care of https://github.com/mapbox/mapbox-gl-native/pull/6448#issuecomment-249333715.

platform/android/MapboxGLAndroidSDK/src/main/java/com/mapbox/mapboxsdk/style/layers/FillLayer.java
```diff
@@ -143,7 +143,7 @@ public FillLayer withProperties(@NonNull Property<?>... properties) {
         return (PropertyValue<String>) new PropertyValue(nativeGetFillColor());
     }
      /**
```
1ec5: Some funky indentation going on here.

incanus:
> We can maybe detect constructions such as “x is set to y” and find candidate values accordingly.

I think this is going to be the best route, with the following rules: 

- `foo is set to bar`: recognize `bar` as as a peer value in the same layer type but another property
- `foo is equivalent to bar`: recognize `bar` as a peer value in the same property

platform/android/MapboxGLAndroidSDK/src/main/java/com/mapbox/mapboxsdk/style/layers/FillLayer.java
incanus: Think I got this in caeee4fcad54196fd97b0bd0fe9b248fda59671b.

incanus:
Got the last mile of Darwin docs in e08bc5a1b68e9253baff6118a225b6a48f8186e6. Formatting rules documented in `generate-style-code.js` itself.

incanus:
@ivovandongen To build on https://github.com/mapbox/mapbox-gl-native/pull/6508#issuecomment-250538933, https://github.com/mapbox/mapbox-gl-native/pull/6508/commits/132512740ae0e158329d45811d887237ccc37e8f brings over some of the logic from Darwin where we properly format symbol names in the `PropertyFactory` docs. 

Before: 

screen shot 2016-09-29 at 4 21 02 pm

After: 

screen shot 2016-09-29 at 4 21 20 pm

Note how we go from referring to `fill-color` (GL style spec) to `fillColor` (Android property name). 

Previously, unlike on Darwin, we didn't refer to symbols as they actually appear in the Android API, which can be confusing and lose people. 

A few more questions about how much further we should go on this: 

- Is there a way in JavaDoc that we can cross-link these symbols for better discoverability rather than just surround them in backticks? If so, should we do this? 

- As of changes like https://github.com/mapbox/mapbox-gl-style-spec/commit/742bc7a72a9a00e7f5ec3369339a7f78cb2b85d2, we move certain documentation from the `enum` property names to the  _values_. However, on Android, since this is through `PropertyFactory` and uses primitives like `string`, `float`, etc., we don't actually document the existence of `enum` values (again, unlike on Darwin; see for example for each of `butt`, `round`, and `square` [here](https://github.com/mapbox/mapbox-gl-native/blob/132512740ae0e158329d45811d887237ccc37e8f/platform/darwin/src/MGLLineStyleLayer.h#L9-L25)). I think that we should, but I'm not sure where this sort of thing would go. We should be documenting for users the range of possibilities to use in the runtime styling API. 

  Compare iOS: 

  ```objc
  /**
   Label placement relative to its geometry.
   */
  typedef NS_ENUM(NSUInteger, MGLSymbolStyleLayerSymbolPlacement) {
      /**
       The label is placed at the point where the geometry is located.
       */
      MGLSymbolStyleLayerSymbolPlacementPoint,
      /**
       The label is placed along the line of the geometry. Can only be used on `LineString` and `Polygon` geometries.
       */
      MGLSymbolStyleLayerSymbolPlacementLine,
  };
  ```

  To Android: 

  ```java
  /**
   * Label placement relative to its geometry.
   *
   * @param function a wrapper function for String
   * @return property wrapper around a String function
   */
  public static Property<Function<String>> symbolPlacement(Function<String> function) {
      return new LayoutProperty<>("symbol-placement", function);
  }
  ```

Note how we lost context like the fact that `line` can only be used on `LineString` and `Polygon`.  

/cc @tobrun @zugaldia

platform/darwin/scripts/generate-style-code.js
```diff
@@ -77,17 +83,33 @@ global.testHelperMessage = function (property, layerType, isFunction) {
     }
 };
 
-global.propertyDoc = function (property, layerType) {
-    let doc = property.doc.replace(/`(.+?)`/g, function (m, symbol, offset, str) {
+global.propertyDoc = function (propertyName, property, layerType) {
+    // Match references to other property names & values. 
+    // Requires the format 'When `foo` is set to `bar`,'.
+    let doc = property.doc.replace(/When `(.+?)` is set to `(.+?)`,/g, function (m, peerPropertyName, propertyValue, offset, str) {
```
1ec5: Is "When" needed? If so, can it be case-insensitive using the `i` flag? Is the comma needed? I think we should try to avoid coupling too tightly to how things are worded right now, because the style spec repo doesn't have any documentation that would inform a contributor that they need to worry about these details.

incanus: I tried without the `When` and comma, but ran into issues that crossed sentence boundaries. I can try to refine it further.

1ec5:
> Is there a way in JavaDoc that we can cross-link these symbols for better discoverability rather than just surround them in backticks? If so, should we do this?

`{@SNAKES_ON_A_LAYER}`

> We should be documenting for users the range of possibilities to use in the runtime styling API.

How about a `<dl>` in the existing comments for the property factory methods?

platform/darwin/scripts/generate-style-code.js
1ec5: If the `+?` is matching too greedily for some reason, you can replace `.` with:

```[^`]```

File: platform/android/MapboxGLAndroidSDK/src/main/java/com/mapbox/mapboxsdk/style/layers/Property.java [hunk 7]
```diff
@@ -11,7 +11,9 @@
  */
 public abstract class Property<T> {
 
-    //visibility
+    /**
+     * Layer visibility
+     */
     public static final String VISIBLE = "visible";
```
ivovandongen: @incanus If you document in this way, it is only applied to the first constant `VISIBLE`. The `//` I used don't show up at all in javadoc. I use those more as you'd use `#pragma mark`, to organize the source, not to document. If you apply javadoc, you need to do it on all constants I'm afraid.

platform/android/MapboxGLAndroidSDK/src/main/java/com/mapbox/mapboxsdk/style/layers/Property.java
```diff
@@ -22,9 +24,14 @@
     @Retention(RetentionPolicy.SOURCE)
     public @interface VISIBILITY {}
 
-    //line-cap
+    /**
+     * `line-cap`: The display of line endings.
+     */
+    //A cap with a squared-off end which is drawn to the exact endpoint of the line.
```
ivovandongen: Same thing here. I would invert this, start the header with `//` and the individual comments with `/** */`

ivovandongen:
> Is there a way in JavaDoc that we can cross-link these symbols for better discoverability rather than just surround them in backticks? If so, should we do this?

http://docs.oracle.com/javase/8/docs/technotes/tools/windows/javadoc.html#CHDDIECH

platform/darwin/scripts/generate-style-code.js
incanus: Specifically I tried things like `.[^\.]+?` in the regex to avoid sentence-spanning over periods. Will take another crack.

1ec5:
```js
> "When `foo` is set to `bar`, `baz` is set to `boo`.".match(/`(.+?)` is set to `(.+?)`/g)
[ "`foo` is set to `bar`", "`baz` is set to `boo`" ]
> "When `foo` is set to `bar`, `baz` becomes `boo`. When `moo` is set to `goo`, `oog` becomes `oop`.".match(/`(.+?)` is set to `(.+?)`/g)
[ "`foo` is set to `bar`", "`baz` becomes `boo`. When `moo` is set to `goo`" ]
> "When `foo` is set to `bar`, `baz` becomes `boo`. When `moo` is set to `goo`, `oog` becomes `oop`.".match(/`([^`]+?)` is set to `([^`]+?)`/g)
[ "`foo` is set to `bar`", "`moo` is set to `goo`" ]
```

incanus:
Still pushing on Android finishing work here per discussion with @ivovandongen and more learnin'  in the JavaDocs department.

incanus:
The JavaDoc stuff gets tricky since we don't know the argument quantity or types, which only lets us do something like `{@link symbolPlacement}` and not the `{@link symbolPlacement(string)}` that is required for actual linking. Per http://www.oracle.com/technetwork/java/javase/documentation/index-137868.html we could just use `<code>symbolPlacement</code>` instead. Doing some experiments.

platform/darwin/src/MGLCircleStyleLayer.h
incanus: Not sure I follow @1ec5.

1ec5: For example:

> Used by the `MGLCircleStyleLayer.circleTranslateAnchor` property.

platform/android/MapboxGLAndroidSDK/src/main/java/com/mapbox/mapboxsdk/style/layers/Property.java
incanus: 357f6a80692efe45952d243fdd1b30ce7cf6ea49

platform/android/MapboxGLAndroidSDK/src/main/java/com/mapbox/mapboxsdk/style/layers/Property.java
incanus: 357f6a80692efe45952d243fdd1b30ce7cf6ea49

incanus:
I'm going to flag some things above in response to comments as tail work. The main goal here is to get use of `enum` docs on the books and improve over current `master`. Then we can incrementally make these all better across all platforms.

platform/darwin/src/MGLCircleStyleLayer.h
incanus: Tail work per https://github.com/mapbox/mapbox-gl-native/pull/6508#issuecomment-250857940.

platform/darwin/src/MGLSymbolStyleLayer.h
incanus: Done pending improvements in https://github.com/mapbox/mapbox-gl-native/pull/6508#discussion_r81259119.

platform/darwin/scripts/generate-style-code.js
incanus: Tail work per https://github.com/mapbox/mapbox-gl-native/pull/6508#issuecomment-250857940.

incanus:
>> We should be documenting for users the range of possibilities to use in the runtime styling API.
> 
> How about a `<dl>` in the existing comments for the property factory methods?

Tail work per https://github.com/mapbox/mapbox-gl-native/pull/6508#issuecomment-250857940.

incanus:
Tail work tickets: https://github.com/mapbox/mapbox-gl-native/issues/6540, https://github.com/mapbox/mapbox-gl-native/issues/6541, https://github.com/mapbox/mapbox-gl-native/issues/6542. 

@1ec5 Mind reviewing again and getting this ready to 🚢?

1ec5:
The Qt build failed; looks like a spurious core test failure:

```
[ RUN      ] Timer.StartOverrides
../../../test/util/timer.test.cpp:165: Failure
Expected: (totalTime) <= (expectedTotalTime * 1.2), actual: 8-byte object <B4-05 00-00 00-00 00-00> vs 8-byte object <00-00 00-00 00-80 76-40>
[  FAILED  ] Timer.StartOverrides (1469 ms)
```

incanus:
Restarted tests and all now pass. Working on merge now.

Issue #5792, opened 2016-07-26
Query all features in sources (querySourceFeatures)


incanus:
Breaking out porting `querySourceFeatures` from JS land separate from the all-encompassing ticket https://github.com/mapbox/mapbox-gl-native/issues/352. 
- Larger PR bringing it to JS: https://github.com/mapbox/mapbox-gl-js/pull/2224
- JS API docs: https://www.mapbox.com/mapbox-gl-js/api/#Map#querySourceFeatures

cammace:
My understanding of this method is that it is needed so we can get the complete feature information query by feature ID. Currently this is happening:

![ezgif com-video-to-gif 5](https://cloud.githubusercontent.com/assets/5652865/17716732/ef5990a8-63d7-11e6-91ea-2cc5c15ac2fd.gif)

cc: @mapbox/android

tobrun:
@cammace that is a separate issue about rendered features and unrelated to source features. I created a follow up ticket in https://github.com/mapbox/mapbox-gl-native/issues/6044.

zugaldia:
@cammace @tobrun Cool if we close this in favor of #6044 then?

tobrun:
@zugaldia, this feature still needs to be implemented ;)

zugaldia:
@tobrun Oops, I mistakenly took @cammace's comment in https://github.com/mapbox/mapbox-gl-native/issues/5792#issuecomment-240250580 as OP. I said nothing :)

PR #6514, opened 2016-09-29
[core] Pre filter sources when querying by layer

tmpsantos:
Feature querying is extremelly slow because it goes through all the sources and these sources can be dense. This PR makes it possible to filter by source.

Annotations will benefit from it because they live in their own "virtual" source. Also if you have a low density source for say, road disruptions, you can make the query extremely responsive by querying only on this source.


src/mbgl/style/source_impl.cpp
```diff
@@ -199,6 +199,10 @@ std::unordered_map<std::string, std::vector<Feature>> Source::Impl::queryRendere
         return result;
     }
 
+    if (parameters.sourceIDs && parameters.sourceIDs->find(id) == parameters.sourceIDs->end()) {
```
brunoabinader: I'd prefer this check to be done in `Style::queryRenderedFeatures` - this way we can avoid the creation of `result` for sources not present in `sourceIDs`.

zugaldia:
@tmpsantos This is great. cc: @ivovandongen for adapting the Android API.

tmpsantos:
For the record Annotations will get a perf bump for free.

tmpsantos:
> Rather than adding new API surface area, can we optimize this case similarly to mapbox/mapbox-gl-js#2877

Not sure about pre-filtering automatically. Seems like a valid optimization when we don't set the source filter, but I like the filter parameter specially for annotations and the use case that motivated me write this PR in the first place: query features on from a source I created using the dataset editor. I wanted these features to get in sync with native views, so they needed to be queried extremely fast.

Looking at the code, the layers are stored in a `std::vector` and we would have to move it to a `std::set` to get reasonable performance. Is there a reason for having it in a `std::vector` in the first place?

1ec5:
> query features on from a source I created using the dataset editor

Wouldn’t #5792 be a more direct API for this use case?

tmpsantos:
> Wouldn’t #5792 be a more direct API for this use case?

The low level core API I'm proposing could be used by the platform bindings for implementing this API.

1ec5:
>> Wouldn’t #5792 be a more direct API for this use case?

> The low level core API I'm proposing could be used by the platform bindings for implementing this API.

This would be a departure from how GL JS implements feature querying: [`querySourceFeatures()`](https://www.mapbox.com/mapbox-gl-js/api/#Map#querySourceFeatures) has quite a few nuances that make it good for the annotation querying use case but not a great candidate for folding into `queryRenderedFeatures()`.

tmpsantos:
@jfirebaugh I re-enabled the benchmarks tests on this PR and wrote a benchmark for querying features. Pre-filtering sources gives a 64x performance bump when the layer you are looking for is in a source not so dense (like annotations or a geojson source).

Explicitly specifying the source makes not measurable difference compared to the pre-filter approach and exposes a new API surface. So let's do how you suggested. :+1:

tmpsantos:
Pre-filtering sources:

```
Benchmark                              Time(ns)    CPU(ns) Iterations
---------------------------------------------------------------------
API_queryRenderedFeatures              14167512   14160000         50                                 
API_queryRenderedFeaturesLayerFilter      21569      21563      33019                                 
```

Bofore the patch:

```
Benchmark                              Time(ns)    CPU(ns) Iterations
---------------------------------------------------------------------
API_queryRenderedFeatures              14428871   14400000         50                                 
API_queryRenderedFeaturesLayerFilter    1309461    1304854        515 
```

PR #6511, opened 2016-09-29
Remove access to gl.h for most code

kkaefer:
We currently liberally include `gl.h`, even when we're just using a few types it defines in code that otherwise doesn't actually use any GL-related code. Instead, we should not include that file except in implementation files that actually do OpenGL calls.


kkaefer:
Given that #6504 is a separate issue, I spend as little time as possible to make `buffer.hpp` gl.h-free.

kkaefer:
The goal of this pull request is to remove inclusion of `gl.hpp` (and in turn `gl.h`) from all header files. They are still included in many implementation files, but we'll have to wait for further refactoring in #6468 to consolidate all OpenGL API use in fewer places.

PR #6521, opened 2016-09-29
Use correct nullity annotation for convertView

athornz:
This annotation was incorrect, causing a crash when using the adapter with Kotlin due to a null convertView argument being passed to getView().


tobrun:
This is correct, not sure where this change happened since the MarkerViewActivity already had `@NonNull` annotation..

PR #6522, opened 2016-09-29
Automate build process for Android SDK releases

zugaldia:
@mapbox/android -

This is the first iteration of a script to helps us automate our release process. For now, you can start using it to build snapshots, although it should also support betas and final releases.

A couple of examples:

- Publish a snapshot from master (`release.py` uses the current branch)

```
$ git branch
* master
  1234-fix-crash
$ python release.py --stage snapshot
```

- Publish a snapshot from a feature branch (same as before, just switch branchs with git):

```
$ git branch
  master
* 1234-fix-crash
$ python release.py --stage snapshot
```

I'm also working on updating our release process docs to details the install process and usage. Any thoughts meanwhile?

### DIFF
--- a/platform/android/scripts/release.py
+++ b/platform/android/scripts/release.py
@@ -1,27 +1,66 @@
 '''
-Utility to schedule SDK builds in Bitrise
+Utility to schedule SDK builds in Bitrise.
+
+Examples:
+
+- Publish a snapshot from master (release.py uses the current branch)
+
+	$ git branch
+	* master
+	  1234-fix-crash
+	$ python release.py --stage snapshot
+
+- Publish a snapshot from a feature branch (same as before, just switch branchs with git):
+
+	$ git branch
+	  master
+	* 1234-fix-crash
+	$ python release.py --stage snapshot
+
+- Publish a beta from a pre-release branch:
+
+	$ git branch
+	  master
+	* release-android-420-beta1
+	$ python release.py --stage beta --version 4.2.0-beta.1
+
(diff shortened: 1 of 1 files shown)


zugaldia:
Could someone from @mapbox/android review? Try, for example, building a SNAPSHOT using the command listed in the OP.

tobrun:
After executing: `python platform/android/scripts/release.py --stage snapshot`
I'm running into:

```python
Traceback (most recent call last):
  File "platform/android/scripts/release.py", line 43, in <module>
    import requests
ImportError: No module named requests

```

tobrun:
I seems that I'm only able to build from the current git branch? could we add an option that would allow building from another eg. `--branch master` ?  For example when executing the script in this PR it results in:

screen shot 2016-10-06 at 11 58 55

Being able to change all the options through command line options would be a nice to have!

zugaldia:
> or add some documentation to it?

Definitely need to improve the docs 👍 .

> could we add an option that would allow building from another eg. --branch master ?

I avoided this on purpose for this release because depending on what you're trying to do, you need to be in the actual branch you want to publish to make some necessary changes (version updates). For now, the simple workaround is to switch to the branch you want to release as a snapshot and then execute the same command.

zugaldia:
@tobrun Correction, you can actually do this with `python release.py --type=snapshot --branch=1234-crash-fix`, but the script will complain if you try to release a non-snapshot from a branch you are not in (because changes are necessary).

Issue #6302, opened 2016-09-09
MGLGeoJSONSource should have a features initializer


1ec5:
MGLGeoJSONSource currently has four initializers: two that accept a GeoJSON-formatted NSData object, and two that accept a URL to a local or remote GeoJSON file. MGLGeoJSONSource should have an additional initializer that accepts an array of MGLFeatures and would accompany the existing read-only `features` property.

Consider this basic use case: the developer wants to create a GeoJSON source containing a particular polyline feature. Currently, they have to either compose a GeoJSON string and stuff it into an NSData object or write out a GeoJSON file and refer to it by its path. Instead, it would be much more convenient to create the source using an MGLPolyline object directly.

To simplify MGLGeoJSONSource, we should remove `-initWithSourceIdentifier:geoJSONData:` and `-initWithSourceIdentifier:URL:` and make the `options` parameter on the remaining initializers nullable. MGLGeoJSONSource would be left with three initializers corresponding to the three properties `features`, `geoJSONData`, and `URL`.

We should also make `MGLFeature.attributes` read-write so that it’s possible to create a feature with specific attributes that a layer might filter for.

/cc @boundsj @incanus @frederoni

1ec5:
This will require functionality for writing out MGLFeature objects as `mapbox::geojson` types or, alternatively, as GeoJSON strings. Each of the MGLFeature subclasses would implement a `-geoJSONFeature` method similar to the existing `-annotationObjectWithDelegate:` method, but using the `attributes` dictionary instead of MGLMultiPointDelegate.

In order to convert the attribute dictionary into a `mapbox::geojson::property_map`, we can implement category methods similar to `-[NSNumber(MGLStyleAttributeAdditions) mbgl_boolPropertyValue]`. Or, if that proves to be too cumbersome, a single recursive method would also work.

1ec5:
> This will require functionality for writing out MGLFeature objects as `mapbox::geojson` types

The manual MGLFeature-to-GeoJSON conversion in mapbox/react-native-mapbox-gl#419 could serve as a template for the kind of transformation we’d need to do here.

PR #6524, opened 2016-09-29
Add features initializer to MGLGeoJSONSource

ituaijagbone:
fixes https://github.com/mapbox/mapbox-gl-native/issues/6302


platform/darwin/src/MGLFeature.mm
```diff
@@ -37,11 +49,30 @@ @implementation MGLPolylineFeature
...
 
 - (id)attributeForKey:(NSString *)key {
     return self.attributes[key];
 }
 
+- (NS_DICTIONARY_OF(NSString *, id) *)featureDictionary {
+    
+    NSMutableArray *coordinates = [NSMutableArray array];
+    
+    for (int index = 0; index < self.pointCount; index++) {
```
1ec5: int might not be large enough for the number of coordinates that may be in this polyline. Use the same type as pointCount.

platform/darwin/src/MGLGeoJSONSource.h
```diff
@@ -131,6 +131,9 @@ extern NSString * const MGLGeoJSONToleranceOption;
  */
 - (instancetype)initWithSourceIdentifier:(NSString *)sourceIdentifier URL:(NSURL *)url options:(NS_DICTIONARY_OF(NSString *, id) *)options NS_DESIGNATED_INITIALIZER;
 
+
+- (instancetype)initWithSourceIdentifier:(NSString *)sourceIdentifier Features:(NS_ARRAY_OF(id<MGLFeature>) *)features options:(nullable NS_DICTIONARY_OF(NSString *, id) *)options NS_DESIGNATED_INITIALIZER;
```
1ec5: Nit: lowercase the Features parameter.

platform/darwin/src/MGLGeoJSONSource.mm
```diff
@@ -118,11 +127,28 @@ - (void)validateValue:(id)value
...
         source->setURL(url.absoluteString.UTF8String);
-    } else {
+    } else if (self.geoJSONData) {
         NSString *string = [[NSString alloc] initWithData:self.geoJSONData encoding:NSUTF8StringEncoding];
         const auto geojson = mapbox::geojson::parse(string.UTF8String).get<mapbox::geojson::feature_collection>();
         source->setGeoJSON(geojson);
         _features = MGLFeaturesFromMBGLFeatures(geojson);
+    } else {
+    
+        NSMutableArray<NS_DICTIONARY_OF(NSString *, id) *> *featuresArray = [NSMutableArray array];
```
1ec5: Use the macros consistently: either change NSMutableArray to NS_MUTABLE_ARRAY_OF, or change NS_DICTIONARY_OF to NSDictionary.

platform/darwin/src/MGLGeoJSONSource.mm
```diff
@@ -118,11 +127,28 @@ - (void)validateValue:(id)value
...
+        for (id<MGLFeature> feature in self.features) {
+            [featuresArray addObject:[feature featureDictionary]];
+        }
+        
+        NS_DICTIONARY_OF(NSString *, id) *featureCollection = @{@"type":@"FeatureCollection",
+                                            @"features":featuresArray};
+        
+        NSError *error;
+        NSData *featuresJSONData = [NSJSONSerialization dataWithJSONObject:featureCollection options:NSJSONWritingPrettyPrinted error:&error];
```
1ec5: Pretty printing is good for debugging purposes, but it'll need to be removed before merging because it increases memory consumption.

platform/darwin/src/MGLFeature.mm
```diff
@@ -188,10 +195,47 @@ - (id)attributeForKey:(NSString *)key {
...
+            [geometries addObject:@{
+                @"type":@"GeometryCollection",
+                @"geometries":[self geometryCollection:((MGLShapeCollectionFeature *)shape).shapes]}];
+        }
+    }
+    
+    return geometries;
+}
+
+- (void)addTypeAndCoordinatesFrom:(id<MGLFeature>)feature to:(NS_MUTABLE_ARRAY_OF(NS_DICTIONARY_OF(NSString *, id) *)**)geometries {
```
1ec5: This is a questionable refactoring. First of all, since the array is mutable, you can simply pass a pointer to it rather than a double-pointer. Secondly, if all the classes that implement MGLFeature have a `-featureDictionary` method, it would be cleaner to declare that method on MGLFeature. That way, `-geometryCollection:` can simply call `feature.featureDictionary` without caring which specific class `feature` happens to be.

platform/darwin/src/MGLFeature.mm
```diff
@@ -188,10 +195,47 @@ - (id)attributeForKey:(NSString *)key {
...
+    return @{@"type":@"Feature",
+             @"properties":(self.attributes) ? self.attributes : [NSDictionary dictionary],
+             @"geometry":@{
+                     @"type":@"GeometryCollection",
+                     @"geometries":[self geometryCollection:self.shapes]
+                     }
+             };
 }
 
+- (NS_MUTABLE_ARRAY_OF(NS_DICTIONARY_OF(NSString *, id) *) *)geometryCollection:(NS_ARRAY_OF(MGLShape *) *)shapes {
```
1ec5: MGLShapeCollectionFeature already knows about its shapes, so there’s no need to pass it in here. Just use `self.shapes` inside the method.

platform/darwin/src/MGLFeature.mm
ituaijagbone: Yes You are right. Calling -featureDictionary should work directly. Thank you.

platform/ios/test/MGLGeoJSONSourceTests.mm
```diff
@@ -36,4 +37,222 @@ - (void)testMGLGeoJSONSourceWithOptions {
...
+    std::unique_ptr<mbgl::style::Source> mbglSource = [source mbglSource];
+    
+    XCTAssertNotNil(source.features);
+    XCTAssertEqual(source.features.count, 1);
+    XCTAssertTrue([source.features.firstObject isMemberOfClass:[MGLPolylineFeature class]]);
+}
+
+
+- (void)testMGLGeoJSONSourceWithPolygonFeatures {
+    // TODO: Create feature
```
boundsj: nit remove the scaffolding comments

platform/darwin/src/MGLGeoJSONSource.mm
```diff
@@ -118,11 +109,29 @@ - (void)validateValue:(id)value
...
+        for (id<MGLFeature> feature in self.features) {
+            [featuresArray addObject:[feature featureDictionary]];
+        }
+        
+        NS_DICTIONARY_OF(NSString *, id) *featureCollection = @{
+            @"type":@"FeatureCollection",
+            @"features":featuresArray};
+        
+        NSError *error;
+        NSData *featuresJSONData = [NSJSONSerialization dataWithJSONObject:featureCollection options:0 error:&error];
```
boundsj: https://github.com/mapbox/mapbox-gl-native/pull/6524#discussion_r81261103 suggested removing the pretty printing. Noting also that this will be refactored away in a future commit where we eventually just set `mbgl::GeoJSON{ featureCollection }` and avoid the trip to create JSON data.

boundsj:
We should probably wait for https://github.com/mapbox/mapbox-gl-native/pull/6588 to land before continuing with this because of the the duplicate work and conflicts in `MGLGeoJSONSource` and related tests

cc @1ec5 @ituaijagbone

platform/darwin/src/MGLFeature.mm
boundsj: Done in https://github.com/mapbox/mapbox-gl-native/pull/6524/commits/087f0137087557536ff09b02b680c04c4fd867cc

boundsj:
@incanus @1ec5 @frederoni please feel free to have another look at this when you have a chance. I'm not sure why the CI builds started failing today but it does not seem to be related to any changes in this PR.

Note the comment in https://github.com/mapbox/mapbox-gl-native/pull/6524/commits/70962eb4fb3a562f9bb6c80187cdcb7319cdeeca. I'm up for suggestions about where / how to share the `mbgl::Value` conversion logic.

boundsj:
Example of a query for features in rect that are then used to initialize a GeoJSON source and styled with a fill layer.

screen shot 2016-10-11 at 4 36 55 pm

platform/darwin/src/MGLFeature.mm
```diff
@@ -5,14 +5,105 @@
...
 
-@property (nonatomic, copy, nullable, readwrite) id identifier;
-@property (nonatomic, copy, readwrite) NS_DICTIONARY_OF(NSString *, id) *attributes;
++ (mbgl::Value)convertedValueWithValue:(id)value;
+
+@end
+
+@implementation MGLFeatureUtility
+
+// TODO: This is copied from NSExpression+MGLAdditions. Where should a shared implementation go?
```
1ec5: How about stuffing the value in an NSExpression and calling the existing category method?

platform/darwin/src/MGLFeature.mm
```diff
@@ -28,6 +119,18 @@ - (id)attributeForKey:(NSString *)key {
     return self.attributes[key];
 }
 
+- (NS_DICTIONARY_OF(NSString *, id) *)featureDictionary {
+    return @{@"type": @"Feature",
+             @"properties": (self.attributes) ? self.attributes : @{},
+             @"geometry" :@{@"type":@"Point",
```
1ec5: Nit: funky spacing.

platform/darwin/src/MGLFeature.mm
```diff
@@ -5,14 +5,105 @@
...
+            // We use long long here to avoid any truncation.
+            return { (int64_t)number.longLongValue };
+        }
+    }
+    [NSException raise:@"Value not handled"
+                format:@"Can’t convert %s:%@ to mbgl::Value", [value objCType], value];
+    return { };
+}
+
++ (mbgl::PropertyMap)propertyMapForDictionary:(NSDictionary *)dictionary propertyMap:(mbgl::PropertyMap *)propertyMap {
```
1ec5: Pass C++ objects around by reference rather than as pointers.

platform/darwin/src/MGLFeature.mm
```diff
@@ -5,14 +5,105 @@
...
+    for (NSString *key in dictionary.allKeys) {
+        if ([dictionary[key] isKindOfClass:[NSDictionary class]]) {
+            std::unordered_map<std::string, mbgl::Value> unorderedMap;
+            [self propertyMapForDictionary:dictionary[key] propertyMap:&unorderedMap];
+            (*propertyMap)[[key cStringUsingEncoding:NSUTF8StringEncoding]] = unorderedMap;
```
1ec5: Use .UTF8String.

platform/darwin/src/MGLFeature.mm
```diff
@@ -5,14 +5,105 @@
...
+            (*vector).push_back([MGLFeatureUtility convertedValueWithValue:value]);
+        }
+    }
+    
+    return *vector;
+}
+
++ (mbgl::Feature)mbglFeatureWithPropertyMapForFeature:(mbgl::Feature)feature mglAttributes:(NSDictionary *)attributes {
+    mbgl::PropertyMap propertyMap;
+    [MGLFeatureUtility propertyMapForDictionary:attributes propertyMap:&propertyMap];
```
1ec5: Would it be more convenient for NSDictionary to know how to turn itself into a PropertyMap?

platform/darwin/src/MGLFeature.mm
```diff
@@ -5,14 +5,105 @@
...
+    feature.properties = propertyMap;
```
1ec5: How does the feature's identifier get set?

platform/darwin/src/MGLFeature.mm
```diff
@@ -70,6 +197,24 @@ - (id)attributeForKey:(NSString *)key {
     return self.attributes[key];
 }
 
+- (NSDictionary *)featureDictionary {
+    NSMutableArray *coordinates = [NSMutableArray array];
```
1ec5: Nit: initialize with a capacity.

platform/darwin/src/MGLFeature.mm
```diff
@@ -108,13 +287,47 @@ @implementation MGLShapeCollectionFeature
...
+            [geometries addObject:@{@"type": geometry[@"type"],
+                                    @"coordinates": geometry[@"coordinates"] }];
+        }
+    }
+    
+    return [geometries copy];
+}
 
+- (mbgl::Feature)mbglFeature {
+    [NSException raise:@"Method unavailable" format:@"%s is not available on %@. Please use a concrete subclass of %@", __PRETTY_FUNCTION__, [self class], [self class]];
```
1ec5: MGLShapeCollectionFeature is a concrete class.

platform/darwin/src/MGLMultiPoint.h
```diff
@@ -6,16 +6,24 @@
...
 /**
- The `MGLMultiPoint` class is an abstract superclass used to define shapes
- composed of multiple points. You should not create instances of this class
- directly. Instead, you should create instances of the `MGLPolyline` or
- `MGLPolygon` classes. However, you can use the method and properties of this
- class to access information about the specific points associated with the line
- or polygon.
+ The `MGLMultiPoint` class is used to define shapes composed of multiple points. 
+ You can use the method and properties of this class to access information about 
+ the specific points associated with a line or polygon.
```
1ec5: Mention that this class also doubles as a superclass of MGLPolyline and MGLPolygon, despite being a distinct geometry.

platform/darwin/src/MGLMultiPoint.h
```diff
@@ -6,16 +6,24 @@
...
 @interface MGLMultiPoint : MGLShape
 
 /**
+ Creates and returns an `MGLMultiPoint` object from the specified set of
+ coordinates.
+ 
+ @param coords The array of coordinates defining the shape. The data in this
+ array is copied to the new object.
+ @param count The number of items in the `coords` array.
+ @return A new MultiPoint object.
```
1ec5: Nit: lowercase.

platform/darwin/src/MGLPolygon+MGLAdditions.m
```diff
@@ -0,0 +1,27 @@
+#import "MGLPolygon+MGLAdditions.h"
+
+@implementation MGLPolygon (MGLAdditions)
+
+- (NS_ARRAY_OF(id) *)mgl_coordinates {
+    NS_MUTABLE_ARRAY_OF(NS_MUTABLE_ARRAY_OF(NS_ARRAY_OF(NSNumber *) *) *) *coordinates = [NSMutableArray array];
```
1ec5: This is really unwieldy. Just use an unqualified mutable array.

1ec5: Also initialize it with a capacity.

platform/ios/ios.xcodeproj/project.pbxproj
```diff
@@ -166,6 +166,7 @@
 		404C26E71D89C55D000AA13D /* MGLTileSet_Private.h in Headers */ = {isa = PBXBuildFile; fileRef = 404C26E61D89C515000AA13D /* MGLTileSet_Private.h */; };
 		404C26E81D89C55D000AA13D /* MGLTileSet_Private.h in Headers */ = {isa = PBXBuildFile; fileRef = 404C26E61D89C515000AA13D /* MGLTileSet_Private.h */; };
 		4085AF091D933DEA00F11B22 /* MGLTileSetTests.mm in Sources */ = {isa = PBXBuildFile; fileRef = 4085AF081D933DEA00F11B22 /* MGLTileSetTests.mm */; };
+		40CF6DBB1DAC3C6600A4D18B /* MGLShape_Private.h in Headers */ = {isa = PBXBuildFile; fileRef = 40CF6DBA1DAC3C1800A4D18B /* MGLShape_Private.h */; };
```
1ec5: A bunch of these files are missing from the macOS project.

platform/darwin/src/MGLFeature.mm
boundsj: Fixed in 3a6be23e3a922e739b6e7b1687d5630621bb7ba6

platform/darwin/src/MGLFeature.mm
boundsj: Fixed in 5b6d989b6d5be439797433a06de027656dc0d28e

platform/ios/ios.xcodeproj/project.pbxproj
boundsj: Fixed in 4268576d1e5dc05514ebf412a6b9723ab4623b93

platform/darwin/src/MGLFeature.mm
boundsj: Fixed in 3770a036d68a4bcca7ffa91a4ff11036262d491a

platform/darwin/src/MGLFeature.mm
boundsj: Fixed in 3770a036d68a4bcca7ffa91a4ff11036262d491a

platform/darwin/src/MGLFeature.mm
boundsj: Fixed in 3770a036d68a4bcca7ffa91a4ff11036262d491a

platform/darwin/src/MGLFeature.mm
boundsj: Fixed in 3770a036d68a4bcca7ffa91a4ff11036262d491a

platform/darwin/src/MGLFeature.mm
boundsj: Fixed in 3770a036d68a4bcca7ffa91a4ff11036262d491a

platform/darwin/src/MGLPolygon+MGLAdditions.m
boundsj: Fixed in 3770a036d68a4bcca7ffa91a4ff11036262d491a

boundsj:
Since yesterday afternoon, I'm still getting `npm ERR! Permission denied (publickey).` for all CI builds and I have no idea why.

platform/darwin/src/MGLFeature.mm
```diff
@@ -5,14 +5,30 @@
...
+#import "MGLPolyline+MGLAdditions.h"
+#import "MGLPolygon+MGLAdditions.h"
+#import "NSDictionary+MGLAdditions.h"
+
+#import "NSExpression+MGLAdditions.h"
 
 #import <mbgl/util/geometry.hpp>
+#import <mapbox/geometry/feature.hpp>
+
+@interface MGLFeatureUtility: NSObject
```
1ec5: This class doesn't seem to be adding much. The lone function can be a standalone C++ function.

platform/darwin/src/MGLFeature.mm
```diff
@@ -28,6 +44,17 @@ - (id)attributeForKey:(NSString *)key {
     return self.attributes[key];
 }
 
+- (NS_DICTIONARY_OF(NSString *, id) *)featureDictionary {
+    return @{@"type": @"Feature",
+             @"properties": (self.attributes) ? self.attributes : @{},
+             @"geometry": @{@"type": @"Point",
```
1ec5: The non-feature direct subclasses of MGLShape should have geoJSONDictionary methods that encapsulate these geometry structures.

1ec5: Once that happens, the bodies of most of these featureDictionary methods can be factored out.

platform/darwin/src/MGLFeature.mm
```diff
@@ -108,13 +206,47 @@ @implementation MGLShapeCollectionFeature
...
+            [geometries addObject:@{@"type": @"GeometryCollection",
+                                    @"geometries": [self geometryCollection]}];
+        } else {
+            NSDictionary *geometry = shape.featureDictionary[@"geometry"];
+            [geometries addObject:@{@"type": geometry[@"type"],
+                                    @"coordinates": geometry[@"coordinates"] }];
+        }
+    }
+    
+    return [geometries copy];
```
1ec5: +[NSArray arrayWithArray:] creates a non-mutable array.

platform/darwin/src/NSDictionary+MGLAdditions.mm
```diff
@@ -0,0 +1,51 @@
...
+#import "NSExpression+MGLAdditions.mm"
+
+@implementation NSDictionary (MGLAdditions)
+
+- (mbgl::PropertyMap)mbgl_propertyMap {
+    mbgl::PropertyMap propertyMap;
+    return [self propertyMapWithPropertyMap:propertyMap dictionary:self];
+}
+
+- (mbgl::PropertyMap)propertyMapWithPropertyMap:(mbgl::PropertyMap &)propertyMap dictionary:(NSDictionary *)dictionary {
```
1ec5: Passing in the property map by non-const reference is similar to an inout C parameter: you're actually mutating the map that's being passed in. So there's no need to return anything. In fact, this method could be renamed to "populatePropertyMap" or somesuch. But if an empty property map is always passed in, why not create it locally instead of accepting it as a parameter?

1ec5: If the intent is to convert a dictionary into a property map, this could be a category method on NSDictionary that takes no parameters and returns a property map.

platform/darwin/src/NSDictionary+MGLAdditions.mm
```diff
@@ -0,0 +1,51 @@
...
+        } else {
+            NSExpression *expression = [NSExpression expressionForConstantValue:dictionary[key]];
+            propertyMap[[key UTF8String]] = [expression mgl_filterValue];
+        }
+    }
+    
+    return propertyMap;
+}
+
+- (std::vector<mbgl::Value>)vectorForNSArray:(NSArray *)array vector:(std::vector<mbgl::Value> &)vector {
```
1ec5: Similarly, this would be a category method on NSArray that returns a vector of values, and the method name would be something like mgl_valueVector.

platform/darwin/src/MGLFeature.mm
boundsj: This is intentional. The method says that it returns a non-mutable array even though it uses a mutable array internally to create the collection

> (NSArray *)geometryCollection

platform/darwin/src/MGLFeature.mm
boundsj: Fixed in https://github.com/mapbox/mapbox-gl-native/pull/6524/commits/206af315fd00efbe3f4b400aa7bd6317db408ac8

platform/darwin/src/MGLFeature.mm
boundsj: I took this I idea and ran with it a bit in https://github.com/mapbox/mapbox-gl-native/pull/6524/commits/ad4e54d6a5a37bd8a589672cd35ebdd04acb711c

platform/darwin/src/NSDictionary+MGLAdditions.mm
boundsj: This was fixed in https://github.com/mapbox/mapbox-gl-native/pull/6524/commits/bc1dda36f6fabb272f273749b8b7cd6121608c48. Note that`propertyMapWithDictionary:` and `vectorForArray:` (name change in https://github.com/mapbox/mapbox-gl-native/pull/6524/commits/95d548b9f2e437e9a3eae2b10860bd0b39fbef04) still take `NSDictionary`s and `NSArray`s so that the entire dictionary can be traversed (recursively) including if any node is a dictionary or array. The `(mbgl::PropertyMap)mbgl_propertyMap` method already exists and is the only public method of the category. It is intended to take no parameters and return a property map.

platform/darwin/src/MGLFeature.h
```diff
@@ -91,6 +91,15 @@ NS_ASSUME_NONNULL_BEGIN
...
 - (nullable id)attributeForKey:(NSString *)key;
 
+/**
+ Returns a dictionary that can be serialized as a GeoJSON Feature representation
+ of the `MGLFeature` subclass.
+ 
+ The dictionary includes the required geometry and properties dictionaries along
+ with an "id" key and value if the underlying `MGLFeature` has an identifier.
+ */
+- (NS_DICTIONARY_OF(NSString *, id) *)geoJSONFeatureDictionary;
```
1ec5: The word “feature” in this method name is redundant.

platform/darwin/src/MGLFeature.h
```diff
@@ -91,6 +91,15 @@ NS_ASSUME_NONNULL_BEGIN
  */
 - (nullable id)attributeForKey:(NSString *)key;
 
+/**
+ Returns a dictionary that can be serialized as a GeoJSON Feature representation
+ of the `MGLFeature` subclass.
```
1ec5: Pedantry: it’s a representation of an instance of the class, not of the class itself.

platform/darwin/src/MGLFeature.h
```diff
@@ -91,6 +91,15 @@ NS_ASSUME_NONNULL_BEGIN
...
+ 
+ The dictionary includes the required geometry and properties dictionaries along
```
1ec5: > The dictionary includes a `geometry` key corresponding to the receiver’s underlying geometry data, a `properties` key corresponding to the receiver’s `attributes` property, and an `id` key corresponding to the receiver’s `identifier` property.

platform/darwin/src/MGLFeature.mm
```diff
@@ -42,6 +50,14 @@ - (id)attributeForKey:(NSString *)key {
     return self.attributes[key];
 }
 
+- (NSDictionary *)geoJSONFeatureDictionary {
```
1ec5: Rename `-geoJSONGeometryDictionary` to `-geoJSONDictionary` and `-geoJSONFeatureDictionary` to `geoJSONDictionary`. Then this method can call super’s implementation and wrap it in `NSDictionaryFeatureForGeometry()`.

platform/darwin/src/MGLFeature.mm
```diff
@@ -224,3 +286,23 @@ static CLLocationCoordinate2D toLocationCoordinate2D(const mbgl::Point<T> &point
...
+    }
+    feature.properties = [attributes mbgl_propertyMap];
+    return feature;
+}
+
+NS_DICTIONARY_OF(NSString *, id) *NSDictionaryFeatureForGeometry(NSDictionary *geometry, NSDictionary *attributes, id identifier) {
+    NSMutableDictionary *feature = [@{@"type": @"Feature",
+                                      @"properties": (attributes) ?: [NSNull null],
+                                      @"geometry": geometry} mutableCopy];
+    if (identifier) {
```
1ec5: Nit: the `if` statement is unnecessary; it’s perfectly fine to set to `nil` using subscript notation.

platform/darwin/src/NSDictionary+MGLAdditions.mm
```diff
@@ -0,0 +1,47 @@
...
+
+#import "NSExpression+MGLAdditions.mm"
+
+@implementation NSDictionary (MGLAdditions)
+
+- (mbgl::PropertyMap)mbgl_propertyMap {
+    return [self propertyMapWithDictionary:self];
+}
+
+- (mbgl::PropertyMap)propertyMapWithDictionary:(NSDictionary *)dictionary {
```
1ec5: Remove the `dictionary` parameter and use `self` instead. Below, call the method recursively on `dictionary[key]` rather than on `self`. Then rename this method to `-mgl_propertyMap` (not `-mbgl_propertyMap` – MGL is the class prefix for this target).

platform/darwin/src/NSExpression+MGLAdditions.mm
```diff
@@ -61,4 +61,30 @@ @implementation NSExpression (MGLAdditions)
...
+        } else if ( strcmp([number objCType], @encode(double)) == 0 ||
+                    strcmp([number objCType], @encode(float)) == 0) {
+            return mbglValue.get<double>();
+        } else if ([number compare:@(0)] == NSOrderedDescending ||
+                   [number compare:@(0)] == NSOrderedSame) {
+            return mbglValue.get<uint64_t>();
+        } else if ([number compare:@(0)] == NSOrderedAscending) {
+            return mbglValue.get<int64_t>();
+        }
+    }
```
1ec5: Should we throw an exception if we encounter an unknown type?

platform/darwin/src/MGLFeature.mm
boundsj: If I remove that conditional, I get a runtime exception when `[NSExpression expressionForConstantValue:identifier];` is called. I'm using NSExpression for the conversion of the id to the mbgl variant .

platform/darwin/src/NSExpression+MGLAdditions.mm
boundsj: I thought about this and decided to not throw an exception when I added it yesterday. 

The reason is that the first thing `mgl_featureIdentifier` does is call `mgl_convertedValueWithValue` which will throw an exception if there is an unknown type. `mgl_featureIdentifier` exists only to be able to get the correct raw value from `mbgl::Value` returned by `mgl_convertedValueWithValue` and the only reason we are doing any of this here is to share the conversion logic that happens to currently live in the `NSExpression` category.

I think you could make the case that we should catch an exception thrown by `mgl_convertedValueWithValue` and re-raise it with a different message (instead of the `Can’t convert %s:%@ to mbgl::Value`) that a developer will see if they send an invalid value for the `id` / `identifier`. However, I was thinking that it might make sense to wait and refactor this if we introduce a category method on NSObject as you've suggested in conversation. It is a bit weird at the moment though.

platform/darwin/src/MGLFeature.h
boundsj: Fixed in 0c317fa

platform/darwin/src/MGLFeature.h
boundsj: Fixed in 0c317fa

platform/darwin/src/MGLFeature.mm
boundsj: Fixed in 0c317fa

platform/darwin/src/NSDictionary+MGLAdditions.mm
boundsj: Fixed in 0c317fa

boundsj: And 3909852

platform/darwin/src/MGLFeature.mm
boundsj: Sorry, I was in `mbglFeature()` not `NSDictionaryFeatureForGeometry()`. Fixed in acf792b

platform/darwin/src/NSDictionary+MGLAdditions.mm
```diff
@@ -0,0 +1,42 @@
...
+            propertyMap[[key UTF8String]] = [self vectorForArray:array];
+        } else {
+            NSExpression *expression = [NSExpression expressionForConstantValue:self[key]];
+            propertyMap[[key UTF8String]] = [expression mgl_filterValue];
+        }
+    }
+    return propertyMap;
+}
+
+- (std::vector<mbgl::Value>)vectorForArray:(NSArray *)array {
```
1ec5: Remove this parameter too.

platform/darwin/src/NSDictionary+MGLAdditions.mm
```diff
@@ -0,0 +1,42 @@
...
+    std::vector<mbgl::Value> vector;
+    vector.reserve(array.count);
+    for (id value in array) {
+        if ([value isKindOfClass:[NSArray class]]) {
```
1ec5: Once we refactor this so that any object knows how to "valueize" itself, we'll be able to remove these conditionals too.

platform/darwin/test/MGLFeatureTests.mm
```diff
@@ -158,4 +158,169 @@ - (void)testPropertyConversion {
...
+                                               @[@[@[@(coord1.longitude), @(coord1.latitude)],
+                                                   @[@(coord2.longitude), @(coord2.latitude)],
+                                                   @[@(coord3.longitude), @(coord3.latitude)]]],
+                                                   @[@[@[@(coord1.longitude), @(coord1.latitude)],
+                                                   @[@(coord2.longitude), @(coord2.latitude)],
+                                                   @[@(coord3.longitude), @(coord3.latitude)]]]]};
+    XCTAssertEqualObjects(geoJSONFeature[@"geometry"], expectedGeometry);
+}
+
+- (void)testShapeCollectionFeatureFeatureGeoJSONDictionary {
```
1ec5: Nit: feature feature! :pizza:

platform/darwin/src/NSDictionary+MGLAdditions.mm
boundsj: Fixed in 61f4d69

platform/ios/ios.xcodeproj/project.pbxproj
```diff
@@ -157,6 +157,9 @@
 		35E79F201D41266300957B9E /* MGLStyleLayer_Private.h in Headers */ = {isa = PBXBuildFile; fileRef = 35E79F1F1D41266300957B9E /* MGLStyleLayer_Private.h */; };
 		35E79F211D41266300957B9E /* MGLStyleLayer_Private.h in Headers */ = {isa = PBXBuildFile; fileRef = 35E79F1F1D41266300957B9E /* MGLStyleLayer_Private.h */; };
 		36F1153D1D46080700878E1A /* libmbgl-core.a in Frameworks */ = {isa = PBXBuildFile; fileRef = 36F1153B1D46080700878E1A /* libmbgl-core.a */; };
+		400533011DB0862B0069F638 /* NSArray+MGLAdditions.h in Headers */ = {isa = PBXBuildFile; fileRef = 400532FF1DB0862B0069F638 /* NSArray+MGLAdditions.h */; };
```
1ec5: These files are missing from macos.xcodeproj. Also, NSArray+MGLAdditions.mm appears to be in Sources twice.

boundsj: Will fix macos. I used the standard mechanism in xcode for adding files. I guess I'll manually edit the file but seems strange that I need to do this.

boundsj: Looks like this is common. I think every impl file is in there twice and often headers are too. Look just below:
```		404C26E21D89B877000AA13D /* MGLTileSet.h in Headers */ = {isa = PBXBuildFile; fileRef = 404C26E01D89B877000AA13D /* MGLTileSet.h */; settings = {ATTRIBUTES = (Public, ); }; };
		404C26E31D89B877000AA13D /* MGLTileSet.h in Headers */ = {isa = PBXBuildFile; fileRef = 404C26E01D89B877000AA13D /* MGLTileSet.h */; settings = {ATTRIBUTES = (Public, ); }; };
		404C26E41D89B877000AA13D /* MGLTileSet.mm in Sources */ = {isa = PBXBuildFile; fileRef = 404C26E11D89B877000AA13D /* MGLTileSet.mm */; };
		404C26E51D89B877000AA13D /* MGLTileSet.mm in Sources */ = {isa = PBXBuildFile; fileRef = 404C26E11D89B877000AA13D /* MGLTileSet.mm */; };```

Maybe this is a bug related to adding to both static and dynamic targets? 

I'm not going to touch this in this PR. Let's open a new issue to clean this up if we need to.

boundsj: Added to macos in c4ea541

boundsj:
Noting that although the Qt and Node builds show that they were unsuccessful above, they actually passed if you click through to bitrise.

platform/ios/ios.xcodeproj/project.pbxproj
1ec5: Oh, you’re right, I thought I was looking at the project structure but I was looking at target membership instead.

Regarding the macOS side, adding a file to ios.xcodeproj won’t automatically add it to macos.xcodeproj. I realize it’s a bit inconvenient. #5942 was an attempt at merging the two projects so that you always get presented with all three targets when adding a file. Unfortunately it slowed down the index too much for comfort.

boundsj: Yeah that's fine. I just forgot

boundsj:
image

PR #6504, opened 2016-09-30
Buffer refactor

jfirebaugh:
As part of the general OpenGL code refactor effort, we should redesign how we handle buffers. The main design goal is to split the current responsibilities of the current `Buffer` class template.

* Building/filling an array of data (in the worker) should be done with plain `std::vector<Vertex>` (or `std::array<Vertex, N>` for fixed-sized arrays). `Vertex` is a stand in for any user-defined, plain-old-struct type with appropriate data members for the vertex attributes.
* Separately, (on the main thread) create an object representing the OpenGL buffer, passing the vector or array. Creation of this object handles OpenGL buffer allocation and data upload in one step.

Secondary design goals:

* Separate, dedicated types for vertex buffers and index buffers. (Technically we have this already since `Buffer` is templated on `GLenum target`, but we want to remove `GLenum` from the public interface.)
* Set up for the vertex buffer type to carry layout information about its elements. (Necessary for #6468.)

API sketch:

```C++
// Vertex buffers

template <class ElementType>
class VertexBufferSlice {
public:
    std::size_t offset() const;
    std::size_t size() const;

    const VertexBuffer<ElementType>& buffer;
};

template <class ElementType>
class VertexBuffer {
public:
    std::size_t size() const;
    VertexBufferSlice<ElementType> slice(std::size_t offset, std::size_t len) const;
    operator VertexBufferSlice<ElementType> () const { return slice(0, size()); }

private:
    UniqueBuffer buffer;
};

template <class Arraylike> // see https://github.com/mapbox/jni.hpp/blob/master/include/jni/arraylike.hpp
VertexBuffer<ArraylikeElement<ArrayLike>> Context::createVertexBuffer(const Arraylike& data);

// Index (a.k.a. element) buffers

template <class Primitive, class IndexType = uint16_t>
class IndexBufferSlice { ... };

template <class Primitive, class IndexType = uint16_t>
class IndexBuffer { ... };

// Where Primitive is one of `Points`, `Lines`, `LineStrip`, `Triangles`:
// https://github.com/mapbox/mapbox-gl-native/blob/3bf08b4cd89b71010bc25dbf74fe3c7752918faa/src/mbgl/gl/primitives.hpp#L14-L17

template <class Primitive, class IndexType>
using IndexBufferElementType;

template <> using IndexBufferElementType<Points> = IndexType;
template <> using IndexBufferElementType<Lines> = IndexType;
template <> using IndexBufferElementType<LineStrip> = IndexType;
template <> using IndexBufferElementType<Triangles> = std::array<IndexType, 3>;

template <class Primitive, class Arraylike>
IndexBuffer<Primitive, IndexType> Context::createIndexBuffer(const Arraylike& data) {
    static_assert(ArraylikeElement<ArrayLike> == IndexBufferElementType<Primitive>);
    ...
}
```


src/mbgl/renderer/circle_bucket.hpp
```diff
@@ -28,10 +27,14 @@ class CircleBucket : public Bucket {
     void drawCircles(CircleShader&, gl::Context&);
 
 private:
-    CircleVertexBuffer vertexBuffer_;
-    TriangleElementsBuffer elementsBuffer_;
+    std::vector<CircleVertex> vertexes;
+    std::vector<gl::Triangle<uint16_t>> indexes;
```
jfirebaugh: @kkaefer Should we hard-code `uint16_t` as the index type? So a declaration like this would be:

```C++
std::vector<gl::Triangle> indexes;
```

kkaefer: OpenGL only supports 16bit unsigned indices anyway

src/mbgl/gl/bytes.hpp
```diff
@@ -0,0 +1,36 @@
+#pragma once
+
+namespace mbgl {
+namespace gl {
+
+struct ElementOffset {
+    std::size_t value;
```
kkaefer: `size_t` is 8 bytes on 64-bit platforms, can we use `uint32_t` here?

src/mbgl/gl/index_buffer.hpp
```diff
@@ -0,0 +1,33 @@
...
+struct Triangle {
+    static constexpr std::size_t IndexCount = 3;
+    Index a;
+    Index b;
+    Index c;
+};
+
+template <class Primitive>
+class IndexBuffer {
+public:
```
kkaefer: Could we `static_assert` that `Primitive` is one of `Line` and `Triangle`?

src/mbgl/gl/vertex_buffer.hpp
```diff
@@ -0,0 +1,18 @@
...
+#include <mbgl/gl/bytes.hpp>
+#include <mbgl/gl/object.hpp>
+
+namespace mbgl {
+namespace gl {
+
+template <class Vertex>
+class VertexBuffer {
+public:
+    static constexpr std::size_t vertexSize = sizeof(Vertex);
```
kkaefer: Could we use `using vertexSize = std::integral_constant<std::size_t, sizeof(Vertex)>;` to avoid static data members that need initialization?

src/mbgl/layout/symbol_layout.cpp
```diff
@@ -464,18 +464,22 @@ void SymbolLayout::addSymbols(Buffer &buffer, const SymbolQuads &symbols, float
...
 
         // add the two triangles, referencing the four coordinates we just inserted.
-        buffer.triangles.add(triangleIndex + 0, triangleIndex + 1, triangleIndex + 2);
-        buffer.triangles.add(triangleIndex + 1, triangleIndex + 2, triangleIndex + 3);
+        buffer.triangles.push_back({static_cast<uint16_t>(triangleIndex + 0),
+                                    static_cast<uint16_t>(triangleIndex + 1),
+                                    static_cast<uint16_t>(triangleIndex + 2)});
+        buffer.triangles.push_back({static_cast<uint16_t>(triangleIndex + 1),
+                                    static_cast<uint16_t>(triangleIndex + 2),
+                                    static_cast<uint16_t>(triangleIndex + 3)});
```
kkaefer: why `push_back` and not `emplace_back`?

src/mbgl/renderer/symbol_bucket.hpp
```diff
@@ -46,25 +47,35 @@ class SymbolBucket : public Bucket {
...
-    typedef ElementGroup<4> IconElementGroup;
-    typedef ElementGroup<1> CollisionBoxElementGroup;
+    typedef gl::ElementGroup<2> TextElementGroup;
+    typedef gl::ElementGroup<4> IconElementGroup;
+    typedef gl::ElementGroup<1> CollisionBoxElementGroup;
 
     struct TextBuffer {
-        TextVertexBuffer vertices;
-        TriangleElementsBuffer triangles;
+        std::vector<TextureRectVertex> vertexes;
```
kkaefer: Why switch from `vertices`/`indices` to `vertexes`/`indexes`?

src/mbgl/shader/circle_vertex.cpp
```diff
@@ -0,0 +1,13 @@
+#include <mbgl/shader/circle_vertex.hpp>
+#include <mbgl/shader/shader.hpp>
+#include <mbgl/gl/gl.hpp>
+
+namespace mbgl {
+
+void CircleVertex::bind(const int8_t* offset) {
+    static_assert(sizeof(CircleVertex) == 4, "expected CircleVertex size");
+
+    MBGL_BIND_VERTEX_ATTRIBUTE(CircleVertex, a_pos, offset);
```
kkaefer: This is an incorrect place for using `a_pos`; those names/attributes are shader-specific and not Vertex type specific.

include/mbgl/gl/gl.hpp
```diff
@@ -36,12 +38,31 @@ struct Error : std::runtime_error {
...
+
+template <class V, class E, std::size_t N>
+void bindVertexAttribute(AttributeLocation location, E (V::*)[N], const int8_t* offset, std::size_t memberOffset) {
+    MBGL_CHECK_ERROR(glEnableVertexAttribArray(location));
+    MBGL_CHECK_ERROR(glVertexAttribPointer(location, N, AttributeType<E>::value, false, sizeof(V), offset + memberOffset));
+}
+
+// This has to be a macro because it uses the offsetof macro, which is the only legal way to get a member offset.
+#define MBGL_BIND_VERTEX_ATTRIBUTE(VertexType, member, offset) \
+    ::mbgl::gl::bindVertexAttribute(Shader::member, &VertexType::member, offset, offsetof(VertexType, member))
```
kkaefer: We should be using `std::is_standard_layout` to ensure that we can safely use `offsetof`. We should also assert that all values we get from `offsetof` are [multiples of 4](https://developer.apple.com/library/content/documentation/3DDrawing/Conceptual/OpenGLES_ProgrammingGuide/TechniquesforWorkingwithVertexData/TechniquesforWorkingwithVertexData.html#//apple_ref/doc/uid/TP40008793-CH107-SW7). 

It also assumes that the shader and the vertex type have identically named member names. I feel a little bit uneasy about that, but can't quite put my finger on it.

src/mbgl/shader/raster_vertex.hpp
```diff
@@ -0,0 +1,25 @@
...
+              x,
+              y
+          },
+          a_texture_pos {
+              tx,
+              ty
+          } {}
+
+    int16_t a_pos[2];
+    int16_t a_texture_pos[2];
```
kkaefer: Can we make all of these members `const`?

src/mbgl/shader/raster_shader.hpp
```diff
@@ -5,11 +5,18 @@
...
+
+class RasterVertex;
+
 class RasterShader : public Shader {
 public:
     RasterShader(gl::Context&, Defines defines = None);
 
-    void bind(int8_t* offset) final;
+    void bind(const gl::VertexBuffer<RasterVertex>&,
+              const int8_t* offset);
```
kkaefer: Can we rename `offset` to something like `begin` to avoid confusion of the offset within a buffer with the offset of an attribute within a vertex?

src/mbgl/renderer/circle_bucket.hpp
jfirebaugh: Ah, this is an OpenGL ES limitation and I was looking at generic OpenGL documentation for `glDrawElements`. OpenGL ES also supports `uint8_t` indexes, but I don't think we'll ever use those. So I'll deparameterize this.

src/mbgl/layout/symbol_layout.cpp
jfirebaugh: `emplace_back` doesn't work with aggregate initialization. I'll add constructors to `Line` and `Triangle` so we can use it.

src/mbgl/renderer/symbol_bucket.hpp
jfirebaugh: For `indexes` I actually meant to use `lines` or `triangles` depending on the primitive. I'll switch back to `vertices`.

src/mbgl/gl/bytes.hpp
jfirebaugh: Why? If you want to match `GLsizei`, it should be `int32_t`, or actually `uint16_t` because this must be a non-negative value.

src/mbgl/shader/circle_vertex.cpp
jfirebaugh: Vertex attribute binding involves some shader-specific things and some vertex-specific things:

* Shader-specific: attribute name, index, and GLSL type.
* Vertex-specific: member name, offset, and C++ type.

I'd like to come up with a design where:

* A given C++ vertex type can be used with multiple shaders.
* A shader requires one particular vertex type, enforced at compile time.
* Attribute indexes are defined per-`Shader`-subclass, rather than in `Shader`, as they are now. Not all shaders have all attributes.
* Debug builds verify that the GLSL and native types are consistent, using `glGetActiveAttrib` to introspect the GLSL type. (We should do this for uniforms as well.)
* I'm open to a design that doesn't enforce that the GLSL attribute name matches the C++ member name, but it also doesn't seem bad to do so. Why would we need them to differ?

I think the starting point here is to introduce a counterpart to `uniform.{hpp,cpp}` for attributes and add members to each `Shader` subclass -- e.g.:

```
AttributeVector<int16_t, 2> a_pos = { "a_pos", *this };
```

Then see where that leads the binding code.

src/mbgl/shader/raster_shader.hpp
jfirebaugh: My plan here is to replace both parameters with `const gl::VertexBufferSlice<RasterVertex>&`, but I'd like to put that off until a followup PR.

src/mbgl/shader/circle_vertex.cpp
jfirebaugh: [Sketched here](https://github.com/mapbox/mapbox-gl-native/commit/2c52f5959aca7f9bfa5f9ebad02534b3506c38d0) but currently stuck on a tricky problem with type erasing pointer-to-members.

jfirebaugh: ✅

jfirebaugh:
@kkaefer Re-review?

include/mbgl/gl/gl.hpp
```diff
@@ -36,12 +40,40 @@ struct Error : std::runtime_error {
...
+template <> struct AttributeType<int32_t>  : std::integral_constant<GLenum, GL_INT> {};
+template <> struct AttributeType<uint32_t> : std::integral_constant<GLenum, GL_UNSIGNED_INT> {};
+template <> struct AttributeType<float>    : std::integral_constant<GLenum, GL_FLOAT> {};
+
+template <std::size_t memberOffset, class V, class E, std::size_t N>
+void bindVertexAttribute(AttributeLocation location, const E (V::*)[N], const int8_t* offset) {
+    static_assert(std::is_standard_layout<V>::value, "vertex type must use standard layout");
+    static_assert(memberOffset % 4 == 0, "vertex attribute must be optimally aligned");
+    static_assert(1 <= N && N <= 4, "count must be 1, 2, 3, or 4");
+    static_assert(sizeof(V) <= std::numeric_limits<GLsizei>::max(), "vertex type is too big");
```
kkaefer: lol

src/mbgl/gl/element_group.hpp
```diff
@@ -0,0 +1,25 @@
...
+
+template <uint8_t count>
+struct ElementGroup : public util::noncopyable {
+    std::array<VertexArrayObject, count> array;
+    uint32_t vertex_length;
+    uint32_t elements_length;
+
+    ElementGroup(uint32_t vertex_length_ = 0, uint32_t elements_length_ = 0)
+        : vertex_length(vertex_length_)
+        , elements_length(elements_length_)
```
kkaefer: Our current formatting guide does not have leading commas in initializer lists. We should either change this formatting or the formatting guidelines.

src/mbgl/renderer/debug_bucket.cpp
```diff
@@ -1,54 +1,101 @@
...
-                         const bool complete_,
-                         optional<Timestamp> modified_,
-                         optional<Timestamp> expires_,
-                         MapDebugOptions debugMode_)
-    : renderable(renderable_),
-      complete(complete_),
-      modified(std::move(modified_)),
-      expires(std::move(expires_)),
-      debugMode(debugMode_) {
+std::vector<PlainVertex> buildTextVertexes(const OverscaledTileID& id,
```
kkaefer: There are still a few instances of `vertexes` in the code.

src/mbgl/shader/circle_vertex.cpp
kkaefer: We still have the issue of Vertex types having to have the same names as the shaders. In our code, we (almost?) always do it like that, but there's no reason why we need to do it like that. I.e. `VertexType::a_pos` could be bound to `Shader::a_pos`, but we could also name the position in the `VertexType` something else.

PR #6526, opened 2016-09-30
[android] - convert bearing from core to degrees

### LINKED ISSUE #6492: Restore bearing after orientation change
Bearing is not correctly restored:

ezgif com-video-to-gif 66

tobrun:
Closes #6492 

Review @ivovandongen

### DIFF
--- a/platform/android/MapboxGLAndroidSDK/src/main/java/com/mapbox/mapboxsdk/camera/CameraPosition.java
+++ b/platform/android/MapboxGLAndroidSDK/src/main/java/com/mapbox/mapboxsdk/camera/CameraPosition.java
@@ -199,7 +199,7 @@ public final class CameraPosition implements Parcelable {
             super();
             if (values != null && values.length == 5) {
                 this.target = new LatLng(values[0], values[1]);
-                this.bearing = (float) values[2];
+                this.bearing = Math.toDegrees(values[2]);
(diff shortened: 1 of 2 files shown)


platform/android/MapboxGLAndroidSDK/src/main/java/com/mapbox/mapboxsdk/camera/CameraPosition.java
```diff
@@ -199,7 +199,7 @@ public Builder(double[] values) {
             super();
             if (values != null && values.length == 5) {
                 this.target = new LatLng(values[0], values[1]);
-                this.bearing = (float) values[2];
+                this.bearing = Math.toDegrees(values[2]);
```
zugaldia: @tobrun I've noticed that `Math.toDegrees(x)`, when `x > pi radians`, returns a value that it's also `> 180 degrees`. However, we expect the bearing to be between `-180` and `180`. Would it make sense to add a check here to make sure that the returned value is within the right limits or, otherwise, adjust it with something like `while (bearing < -180) bearing += 360` and the equivalent for `bearing > -180`?

zugaldia: @tobrun 👍

PR #6529, opened 2016-09-30
[iOS] Debug build to include dSYM for use in Instruments

halset:
I use CocoaPods in my app. When using a pod built with "make ipackage", the Mapbox GL Native symbols did not show up in Instruments. This patch fixes that problem.


friedbunny:
Thanks for the PR! This is interesting — I’m not sure why the default Xcode setting in Debug for `DEBUG_INFORMATION_FORMAT` is `DWARF`, rather than `DWARF + dSYM` (as you propose).

Compilation speed, @1ec5?

screen shot 2016-09-30 at 6 10 12 am

platform/ios/ios.xcodeproj/project.pbxproj
```diff
@@ -2167,7 +2167,7 @@
 				CURRENT_COMMIT_HASH = deadbeef;
 				CURRENT_SEMANTIC_VERSION = 1.0.0;
 				CURRENT_SHORT_VERSION = 1.0;
-				DEBUG_INFORMATION_FORMAT = dwarf;
+				DEBUG_INFORMATION_FORMAT = "dwarf-with-dsym";
```
1ec5: Compilation speed is indeed part of the reason that the default is `dwarf` for Debug builds. I think we should keep it that way for builds within Xcode. A dSYM is unnecessary because the SDK is built with debug symbols.

`make iproj` builds a Debug build by default but strips debug symbols in a post-build step. Therefore, package.sh should be modified to override this build setting with `dwarf-with-dsym` if symbols would be stripped (see also #6531).

Please make corresponding changes for the macOS SDK. Thanks!

halset:
Closing this. I will just turn on dsym when I need them when playing with Instruments.

PR #6543, opened 2016-09-30
Loosen style doc transform regex

### LINKED ISSUE #6541: make Darwin enum peer cross-references more robust
Tail work from #6508 referenced in https://github.com/mapbox/mapbox-gl-native/pull/6508#pullrequestreview-2252593. 

We want to more robustly regex-match cross-references _between_ properties in a layer type. 

As an example, the output we have as of #6508 looks like this: 

```objc
/**
 In combination with `symbolPlacement`, determines the rotation behavior of icons.
 */
typedef NS_ENUM(NSUInteger, MGLSymbolStyleLayerIconRotationAlignment) {
    /**
     When `symbolPlacement` is set to `MGLSymbolStyleLayerSymbolPlacementPoint`, aligns icons east-west. When `symbolPlacement` is set to `MGLSymbolStyleLayerSymbolPlacementLine`, aligns icon x-axes with the line.
     */
    MGLSymbolStyleLayerIconRotationAlignmentMap,
    /**
     Produces icons whose x-axes are aligned with the x-axis of the viewport, regardless of the value of `symbolPlacement`.
     */
    MGLSymbolStyleLayerIconRotationAlignmentViewport,
    /**
     When `symbolPlacement` is set to `MGLSymbolStyleLayerSymbolPlacementPoint`, this is equivalent to `MGLSymbolStyleLayerIconRotationAlignmentViewport`. When `symbolPlacement` is set to `MGLSymbolStyleLayerSymbolPlacementLine`, this is equivalent to `MGLSymbolStyleLayerIconRotationAlignmentMap`.
     */
    MGLSymbolStyleLayerIconRotationAlignmentAuto,
};
```

Per the reference comment, we could make this less brittle in our code gen script since the upstream style spec isn't necessarily aware of the conditions. 

/cc @1ec5

1ec5:
Mildly decoupled the transformation from the style specification to the format required for iOS/macOS documentation so it’s harder for style specification references to other properties’ values to fall through the cracks.

Fixes #6541.

/cc @incanus

### DIFF
--- a/platform/darwin/scripts/generate-style-code.js
+++ b/platform/darwin/scripts/generate-style-code.js
@@ -86,10 +86,10 @@ global.testHelperMessage = function (property, layerType, isFunction) {
 global.propertyDoc = function (propertyName, property, layerType) {
     // Match references to other property names & values. 
     // Requires the format 'When `foo` is set to `bar`,'.
-    let doc = property.doc.replace(/When `(.+?)` is set to `(.+?)`,/g, function (m, peerPropertyName, propertyValue, offset, str) {
+    let doc = property.doc.replace(/`([^`]+?)` is set to `([^`]+?)`/g, function (m, peerPropertyName, propertyValue, offset, str) {
         let otherProperty = camelizeWithLeadingLowercase(peerPropertyName);
         let otherValue = objCType(layerType, peerPropertyName) + camelize(propertyValue);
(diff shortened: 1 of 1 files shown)


incanus:
This PR should have the updated style code included, too, which doubles as a way to make sure that it works on the current spec.

1ec5:
I ran `make style-code-darwin` on this branch, but there were no additional changes, which I guess means it works. I filed #6544 to track adding tests, linting, or some kind of verification step to this script.

Issue #6462, opened 2016-09-26
Unable to use LOST API in conjunction with Mapbox Android SDK


dmeignan:
**Platform:** Android
**Mapbox SDK version:** 4.1.1

The Android Mapbox SDK use the [LOST API](https://github.com/mapzen/lost) version 1.1.0 (see `compile 'com.mapzen.android:lost:1.1.0'` in [Gradle dependencies of MapboxGLAndroidSDK](https://github.com/mapbox/mapbox-gl-native/blob/master/platform/android/MapboxGLAndroidSDK/build.gradle)). I also use in my app the LOST API. However, the [version 1.1.0 of the LOST API does not manage correctly multiple location listeners](https://github.com/mapzen/lost/issues/62). In particular, when the `LocationServices` of Mapbox requests the `locationClient` to disconnect (in [LocationServices.java](https://github.com/mapbox/mapbox-gl-native/blob/master/platform/android/MapboxGLAndroidSDK/src/main/java/com/mapbox/mapboxsdk/location/LocationServices.java)), then it shuts down the LOST location engine for all location listeners. This problem has been [solved in version 1.1.1 of the LOST API](https://github.com/mapzen/lost/releases/tag/lost-parent-1.1.1).

The Mapbox Android SDK should use the version 2.0.0 or at least 1.1.1 of the LOST API for a better integration in apps that also use LOST API.

tobrun:
I did a gradle sync to see how big the semver change of 2.0.0 is:
![screen shot 2016-09-27 at 13 30 23](https://cloud.githubusercontent.com/assets/2151639/18871649/a5e580ee-84b6-11e6-9de9-b409a5a35afd.png)

Since we are close to release I'm going to assign it to `android-future`. We are currently thinking of rewriting the way we are handling these location udpdates. For now I'm going to update the dependency to `1.1.1` as requested. This should introduce any changes in code. Thank you for reaching out!

tobrun:
Update to 1.1.1 got merged in #6476. We will upgrade other dependencies with next release cycle.

clydebarrow:
See PR #6548

tobrun:
closing this in favor of #6537

Issue #6197, opened 2016-08-29
Target API 24 (Android Nougat)


zugaldia:
The Mapbox Android SDK [currently targets](https://github.com/mapbox/mapbox-gl-native/blob/master/platform/android/MapboxGLAndroidSDK/gradle.properties) API level 23 (Marshmallow) but Android Nougat (API level 24) has just been released. We should eventually set `android:targetSdkVersion` to API 24. Let's use this ticket to track this progress

A few things to note:
1. Targeting API 24 will force us to update to version 24.x of the Android Support libraries accordingly. This requires changes from developers (i.e. update to API 24 or add exclusion rules).
2. In order to properly target API level we need API breaking changes (e.g. https://github.com/mapbox/mapbox-gl-native/issues/5910). Therefore, we won't be doing this as part of the 4.x series.
3. Note that this does not affect `android:minSdkVersion` which remains at API level 15.

/cc: @mapbox/android

zugaldia:
Related, we might be able to optimize our use of the support libraries (it's being chopped up into smaller modules): https://commonsware.com/blog/2016/08/30/breaking-up-artifact-is-not-that-hard-to-do.html.

zugaldia:
Related: Update to LOST 2.0 https://github.com/mapbox/mapbox-gl-native/issues/6537.

zugaldia:
Speaking of Android 7.0, [Android 7.1 will be entering into Developer Preview](http://android-developers.blogspot.com/2016/10/android-71-developer-preview.html) mode later this month.

tobrun:
Only lost in #6537 needs to be fixed before closing this ticket, other related tickets have been PR'ed in #6390 and #6830.

tobrun:
Lost is PR'ed in https://github.com/mapbox/mapbox-gl-native/pull/7494. No actions left to pick up. Closing.

Issue #4331, opened 2016-03-15
Add LocationSource


tobrun:
[LocationSource](https://developers.google.com/android/reference/com/google/android/gms/maps/LocationSource) is our equivalent of `LocationServices`.

bleege:
Can you describe in more detail what problem this issue would solve?

mkv123:
Currently the user location marker is internally hardcoded to use the device's GPS positioning to define where it is. Imho this is architecurally a bit at odds with separating data and visualization, but also in a very practical sense it makes the position marker very inflexible. 

For example let's say you want a virtual walk through of the route, like a play button to preview your journey. Currently we'd have to re-implement the user position marker for this functionality, instead of simply setting its location. Another case would be for using any 3rd party location source, ranging from IP address to airplane schedules to qr codes.

mkv123:
@tobrun Btw, why the SEMVER-MAJOR tag? If implemented in the same way as in google-play-services, where you provide a default one, but allow overriding it, it wouldn't break API as far as I can tell.

tobrun:
@mkv123 sorry for only getting back to you now. The initial focus of this ticket was to rename `LocationServices` to `LocationSource` (reason behind the `semver`-label). Since posting this some requirements have changed and including a pluggable system for anything location related is on our list of long term improvements to the SDK.

tobrun:
Looking a bit more into this it seems that there are two requirements here:
 - general refactor of LocationServices to be a default LocationSource
 - make pluggable system using the LocationSource interface

tobrun:
Been working on this a bit and need to conclude that this requires a bigger lift as initially thought. Making the system pluggable requires bigger refactors to LocationServices and Telemetry. We ideally pick this up when we introduce a separate telemetry module. I'm going to submit some fixes that I made to telemetry separately.

zugaldia:
We have recently made some progress here:

> general refactor of LocationServices to be a default LocationSource

With https://github.com/mapbox/mapbox-java/issues/269 we now have an abstract `LocationEngine` with sample implementations for Lost and Play Services. We could tweak the former as the default implementation for our `LocationSource`. It conforms to `LocationSource` while expanding its functionality.

> make pluggable system using the LocationSource interface

That's achieved with the newly introduced `LocationEngine` interface.

zugaldia:
Fixed via https://github.com/mapbox/mapbox-gl-native/pull/7892.

PR #6548, opened 2016-10-01
[android] Update Android SDK, gradle plugin and mapzen lib

clydebarrow:
Update Android SDK version to 24 (Android 7)
 Update mapzen lost library to 2.0.0 with required code changes
 Update gradle Android plugin to 2.2


zugaldia:
@clydebarrow Thank you very much for this PR, this is something that we're definitely interested in bringing to the SDK.

I'd note though that this PR, as you mention in the OP, changes the `targetSdkVersion` to Nougat, as well as increments the support libraries versions. This means that SEMVER is affected and we won't be able to merge this into the SDK as part of the 4.x series. Let's track this PR as part of the larger work to improve our support of Android 7 in https://github.com/mapbox/mapbox-gl-native/issues/6197. Sounds good?

clydebarrow:
The LOST upgrade doesn't depend on the `targetSdkVersion` change, so I could split it out into a separate PR if you like.

zugaldia:
@clydebarrow Thank you for the offer but we'll still need to wait. LOST 2.0 itself brings some changes that will require further changes from us. See, for example, https://github.com/mapbox/mapbox-gl-native/issues/6462#issuecomment-249840081 and https://github.com/mapbox/mapbox-gl-native/issues/4331.

PR #6553, opened 2016-10-02
Fix CollisionTile::queryRenderedSymbols query geometry

### LINKED ISSUE #6055: Map::queryPointAnnotations() -> Map::queryRenderedFeatures() returns unexpected result as map is moved

brunoabinader:
Due to a bug in `CollisionTile` - see https://github.com/mapbox/mapbox-gl-native/issues/6055#issuecomment-249328483 for details - the results for `queryRenderedFeatures` for symbol features was wrong in some cases e.g. when map bearing is updated. This PR implements the proposed fix and adds both query and unit tests to verify.

In order to support the proposed query tests, two new Node.js functions are proposed: `setCenter` and `setBearing`. These are exemplified via `operations` inside `style.json`.

Fixes #6055.


brunoabinader:
Next steps:
- [ ] Test the proposed query tests on GL JS to verify if provides the same results
- [ ] Investigate the [failing render tests](http://mapbox.s3.amazonaws.com/mapbox-gl-native/render-tests/16300.1/index.html):
 - 'text-font/chinese'
 - 'text-pitch-alignment/map-text-depthtest'
 - 'text-pitch-alignment/viewport-text-depthtest'

brunoabinader:
Commit https://github.com/mapbox/mapbox-gl-native/pull/6553/commits/86e409879c420ac9ac44d15da8008daa06710b05 fixes an issue I found out while testing for queries: when symbols were placed on tile edges e.g. `{ 0, 0 }` they were _spawning_ 3 other symbols on the surrounding tiles (example below):
screen shot 2016-10-03 at 3 10 42 pm

brunoabinader:
Heads up: Though the fixes on this PR are valid, I am still obtaining the wrong amount of features while querying for visible features as I simply zoom in and out of the map. I'll continue my investigations around the relations between `FeatureIndex` and `CollisionTile`.

brunoabinader:
Commit 86e4098 fixes https://github.com/mapbox/mapbox-gl-native/issues/3563.

brunoabinader:
This PR is getting too big and addressing many different issues. I'm closing this in favor of separated PRs to handle each bug fix specifically.

brunoabinader:
This PR has been split in the following PRs and/or issues:

Fixed and/or pending review:
- Annotating Null Island also annotates its antipode - #3563 
- Map::setLatLng differs from Map::setLatLngZoom if zoom is same - #6625 
- Add unit tests for TileCoordinate and TilePoint - #6626 
- CollisionTile::queryRenderedSymbols should receive the query geometry - #6627
- Annotation symbols should ignore icon placement - #6629 

What's left:
- Collision detection for symbol features is broken - #6628 
- Querying rendered features returns wrong values in perspective view - #6630

PR #6555, opened 2016-10-03
    [android] Added dynamic scale widget to map view.

clydebarrow:
```
Scale is enabled by default, can be disabled in xml or programmatically.
Scale is styleable via xml or programmatically for position, color and distance unit.
Test activity added to TestApp in map layout category.
Sundry refactoring and minor bug fixes found while working in MapView.java.
```


clydebarrow:
screenshot 2016-10-03 09 29 16
scale1

tobrun:
Thank you for contributing, this seems great! 
Putting on my radar to review.

snodnipper:
As an outsider, my initial thought on this PR was whether a standalone widget may keep the SDK code tighter, similar to [MAS](https://github.com/mapbox/mapbox-java/tree/master/libandroid).  AFAIK, Google Maps doesn't have a scale bar within the Android SDK.

For our mbgl demo app, the UX team wanted the scale bar to move etc. based on the chrome around the map - so I had to move the compass and add one of our (branded) android scale bars (because gravity alone wasn't going to work).

Slightly topical given https://github.com/mapbox/mapbox-java/issues/131 _but_ we used RXJava to keep the code fairly clean.  For example:

```java
mMetersPerPixelObservable.subscribe(mpp -> mScaleBar.setMetersPerPixel(mpp););
```

clydebarrow:
@snodnipper A valid question and one I had thought about.

Firstly there is definitely some demand for the feature - there is more than one feature request lodged here for something similar, and it was something I kept looking for on my maps.

Originally I implemented it outside the SDK by [subclassing `Mapview`](http://stackoverflow.com/questions/39739250/mapbox-touchlistner-for-map/39813925#39813925) but there were a few details that anyone wanting to use that had to get right to make it work. Nothing show-stopping, but not quite plug and play.

Looking at the SDK it became apparent that there were already models of similar widgets - the compass and the attribution - that the scale bar would sit nicely alongside both in concept and implementation.

Finally, the [MAS](https://github.com/mapbox/mapbox-java/tree/master/libandroid) link you provided illustrates that at present there is, AFAIK,  no established framework for stand-alone addons to mapbox. It's not immediately obvious (without reading the source) what that code does and whatever it is, it's just an example, not a plug-and-play widget.

One question I'm ambivalent on is whether the scale bar should be enabled by default (in my PR it is.) On balance I think that's fine, because it enables easy discovery that the feature exists.

tobrun:
Thanks you guys for the remarks, I ticketed the feedback into #6570.

tobrun:
I'm going to add the android-future for now, we are currently planning a new release and don't want to impact stability by introducing new features at this point. We are currently thinking of merging this once the release lands. Probably in combination with refactoring in #6570. Thanks again for the contribution @clydebarrow!

tobrun:
This is going to be implemented as a plugin in https://github.com/mapbox/mapbox-plugins-android/issues/11

Issue #6512, opened 2016-09-29
NPE when MapView is recreated from savedInstanceSate


clydebarrow:
**Platform: Android**
**Mapbox SDK version: locally built from current git HEAD**

I'm seeing crashes like this on devices in the field:

````
java.lang.NullPointerException
	at com.mapbox.mapboxsdk.maps.NativeMapView.nativeSetStyleUrl(Native Method)
	at com.mapbox.mapboxsdk.maps.NativeMapView.setStyleUrl(NativeMapView.java:204)
	at com.mapbox.mapboxsdk.maps.MapView.setStyleUrl(MapView.java:854)
	at com.mapbox.mapboxsdk.maps.MapboxMap.setStyleUrl(MapboxMap.java:637)
	at com.mapbox.mapboxsdk.maps.MapView.onCreate(MapView.java:424)
	at com.controlj.naips.view.DivMapFragment.onCreateView(DivMapFragment.java:105)
	at android.support.v4.app.Fragment.performCreateView(Fragment.java:2080)
````
This appears to be occurring when `MapView.onCreate()` is called with a non-null `savedInstanceState`, but the value for `STATE_STYLE_URL` in that bundle is null. The bundle is stored in `onSaveInstanceState()` and the value for  `STATE_STYLE_URL` is saved from the MapView field `styleUrl`. This _can_ be null if onResume() has never been called, since `styleUrl` is not set prior to `onResume()` unless an explicit call to `setStyle()` has been made. This would imply that the `MapView` had been created, but never entered the resumed state before being destroyed. I have not been able to verify that this is possible, but there appears to be no other explanation.

I have a plan for a fix and will raise a pull request shortly.

PR #6556, opened 2016-10-03
[android] Rework setStyleUrl() logic to avoid NPE as per #6512

clydebarrow:
The MapView class has logic to set the map style from a url loaded from xml, or by an explicit call to `setStyle()` or `setStyleUrl()` (these are equivalent except for some annotations - one just calls the other) or by defaulting to `Style.MAPBOX_STREETS`. The actual setting of the style on the native map is delayed until `onResume()`, presumably for some good reason. This patch retains the logic but ensures that `styleUrl` is never null, so the NPE reported in #6512  cannot occur.

### DIFF
--- a/platform/android/MapboxGLAndroidSDK/src/main/java/com/mapbox/mapboxsdk/maps/MapView.java
+++ b/platform/android/MapboxGLAndroidSDK/src/main/java/com/mapbox/mapboxsdk/maps/MapView.java
@@ -155,8 +155,8 @@ public class MapView extends FrameLayout {
 
     private PointF focalPoint;
 
-    private String styleUrl;
-    private String initalStyle;
+    private String styleUrl = Style.MAPBOX_STREETS;
+    private boolean styleWasSet = false;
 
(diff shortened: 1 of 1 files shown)


tobrun:
Thank you for contributing, putting on my radar to review

tobrun:
> The actual setting of the style on the native map is delayed until onResume(), presumably for some good reason

The construction there is to load a default style (mapboxstreets) if no style has been set. If a user would have set the style it would already have been loaded. FWIW the fastest way to load a style is through xml or via MapboxMapOptions.

clydebarrow:
For Nougat this will have to be moved to `onStart()` at least.  I can't see any reason why `onCreate()` can't deal with it. Anyway, that's for another time. Right now I just don't want to see crash reports.

tobrun:
I pulled in your changes and pushed them to this repo so our CI tests can run. 
I fixed up the PR with adding some brackets and fixed up a comment.
Thank you for contributing. Closing in favor of #6574.

PR #6559, opened 2016-10-03
Make features and annotations conform to NSSecureCoding

### LINKED ISSUE #6200: Annotation classes don’t conform to NSCoding
The following classes don’t conform to NSSecureCoding but probably should (and could without much fuss):
- MGLAnnotationImage
- MGLAnnotationView
- MGLShape
- MGLUserLocation
- MGLUserLocationAnnotationView

The developer may want to archive an instance of one of these classes for state restoration or some other kind of persistence, but they’ll run into a missing implementation in doing so:

```
-[MGLUserLocation encodeWithCoder:]: unrecognized selector sent to instance 0x60800024b820
```

/cc @friedbunny

frederoni:
This PR fixes #6200 

All feature and annotation classes now conform to NSSecureCoding which means they can be serialized and deserialized. This is useful for [UIStateRestoring](https://developer.apple.com/reference/uikit/uistaterestoring) among other things.

- [x] MGLAnnotationImage
- [x] MGLAnnotationView
- [x] MGLUserLocation
- [x] MGLUserLocationAnnotationView
- [x] MGLShape
- [x] MGLPointAnnotation
- [x] MGLMultiPoint
- [x] MGLPolyline
- [x] MGLPolygon
- [x] MGLPointCollection
- [x] MGLMultiPolyline
- [x] MGLMultiPolygon
- [x] MGLShapeCollection
- [x] MGLPointFeature
- [x] MGLPolylineFeature
- [x] MGLPolygonFeature
- [x] MGLPointCollectionFeature
- [x] MGLMultiPolylineFeature
- [x] MGLMultiPolygonFeature
- [x] MGLShapeCollectionFeature

@incanus @1ec5 👀

### DIFF
--- a/platform/darwin/src/MGLFeature.mm
+++ b/platform/darwin/src/MGLFeature.mm
@@ -10,6 +10,7 @@
 #import "MGLPolyline+MGLAdditions.h"
 #import "MGLPolygon+MGLAdditions.h"
 #import "NSDictionary+MGLAdditions.h"
+#import "NSArray+MGLAdditions.h"
 
 #import "NSExpression+MGLAdditions.h"
 
@@ -24,6 +25,10 @@
 
 @synthesize identifier;
 @synthesize attributes;
+
+MGL_DEFINE_FEATURE_INIT_WITH_CODER();
+MGL_DEFINE_FEATURE_ENCODE();
+MGL_DEFINE_FEATURE_IS_EQUAL();
 
 - (id)attributeForKey:(NSString *)key {
     return self.attributes[key];
@@ -47,6 +52,10 @@
 @synthesize identifier;
 @synthesize attributes;
 
+MGL_DEFINE_FEATURE_INIT_WITH_CODER();
+MGL_DEFINE_FEATURE_ENCODE();
+MGL_DEFINE_FEATURE_IS_EQUAL();
+
 - (id)attributeForKey:(NSString *)key {
     return self.attributes[key];
 }
@@ -68,6 +77,10 @@
 
 @synthesize identifier;
 @synthesize attributes;
+
+MGL_DEFINE_FEATURE_INIT_WITH_CODER();
+MGL_DEFINE_FEATURE_ENCODE();
+MGL_DEFINE_FEATURE_IS_EQUAL();
 
 - (id)attributeForKey:(NSString *)key {
     return self.attributes[key];
@@ -91,6 +104,10 @@
 @synthesize identifier;
 @synthesize attributes;
 
+MGL_DEFINE_FEATURE_INIT_WITH_CODER();
+MGL_DEFINE_FEATURE_ENCODE();
+MGL_DEFINE_FEATURE_IS_EQUAL();
+
 - (id)attributeForKey:(NSString *)key {
     return self.attributes[key];
 }
@@ -113,6 +130,10 @@
 @synthesize identifier;
 @synthesize attributes;
 
+MGL_DEFINE_FEATURE_INIT_WITH_CODER();
+MGL_DEFINE_FEATURE_ENCODE();
+MGL_DEFINE_FEATURE_IS_EQUAL();
+
 - (id)attributeForKey:(NSString *)key {
     return self.attributes[key];
 }
@@ -134,6 +155,10 @@
 
 @synthesize identifier;
 @synthesize attributes;
+
+MGL_DEFINE_FEATURE_INIT_WITH_CODER();
+MGL_DEFINE_FEATURE_ENCODE();
+MGL_DEFINE_FEATURE_IS_EQUAL();
 
 - (id)attributeForKey:(NSString *)key {
     return self.attributes[key];
@@ -162,6 +187,10 @@
 + (instancetype)shapeCollectionWithShapes:(NSArray *)shapes {
     return [super shapeCollectionWithShapes:shapes];
 }
+
+MGL_DEFINE_FEATURE_INIT_WITH_CODER();
+MGL_DEFINE_FEATURE_ENCODE();
+MGL_DEFINE_FEATURE_IS_EQUAL();
 
 - (id)attributeForKey:(NSString *)key {
     return self.attributes[key];
--- a/platform/darwin/src/MGLFeature_Private.h
+++ b/platform/darwin/src/MGLFeature_Private.h
@@ -37,3 +37,31 @@ mbgl::Feature mbglFeature(mbgl::Feature feature, id identifier, NSDictionary *at
 NS_DICTIONARY_OF(NSString *, id) *NSDictionaryFeatureForGeometry(NSDictionary *geometry, NSDictionary *attributes, id identifier);
 
 NS_ASSUME_NONNULL_END
+
+#define MGL_DEFINE_FEATURE_INIT_WITH_CODER() \
+    - (instancetype)initWithCoder:(NSCoder *)decoder { \
+        if (self = [super initWithCoder:decoder]) { \
+            NSSet<Class> *identifierClasses = [NSSet setWithArray:@[[NSString class], [NSNumber class]]]; \
+            identifier = [decoder decodeObjectOfClasses:identifierClasses forKey:@"identifier"]; \
+            attributes = [decoder decodeObjectOfClass:[NSDictionary class] forKey:@"attributes"]; \
+        } \
+        return self; \
+    }
+
+#define MGL_DEFINE_FEATURE_ENCODE() \
+    - (void)encodeWithCoder:(NSCoder *)coder { \
+        [super encodeWithCoder:coder]; \
+        [coder encodeObject:identifier forKey:@"identifier"]; \
+        [coder encodeObject:attributes forKey:@"attributes"]; \
+    }
+
+#define MGL_DEFINE_FEATURE_IS_EQUAL() \
+    - (BOOL)isEqual:(id)other { \
+        if (other == self) return YES; \
+        if (![other isKindOfClass:[self class]]) return NO; \
+        __typeof(self) otherFeature = other; \
+        return [super isEqual:other] && [self geoJSONObject] == [otherFeature geoJSONObject]; \
+    } \
+    - (NSUInteger)hash { \
+        return [super hash] + [[self geoJSONDictionary] hash]; \
+    }
--- a/platform/darwin/src/MGLMultiPoint.mm
+++ b/platform/darwin/src/MGLMultiPoint.mm
@@ -1,9 +1,8 @@
 #import "MGLMultiPoint_Private.h"
 #import "MGLGeometry_Private.h"
+#import "MGLShape_Private.h"
+#import "NSCoder+MGLAdditions.h"
 #import "MGLTypes.h"
-
-#include <mbgl/util/geo.hpp>
-#include <mbgl/util/optional.hpp>
 
 @implementation MGLMultiPoint
 {
@@ -25,6 +24,39 @@
     }
 
     return self;
+}
+
+- (instancetype)initWithCoder:(NSCoder *)decoder
+{
+    if (self = [super initWithCoder:decoder]) {
+        _coordinates = [decoder mgl_decodeLocationCoordinates2DForKey:@"coordinates"];
+    }
+    return self;
+}
+
+- (void)encodeWithCoder:(NSCoder *)coder
+{
+    [super encodeWithCoder:coder];
+    [coder mgl_encodeLocationCoordinates2D:_coordinates forKey:@"coordinates"];
+}
+
+- (BOOL)isEqual:(id)other
+{
+    if (self == other) return YES;
+    if (![other isKindOfClass:[MGLMultiPoint class]]) return NO;
+           
+    MGLMultiPoint *otherMultipoint = other;
+    return ([super isEqual:otherMultipoint]
+            && _coordinates == otherMultipoint->_coordinates);
+}
+
+- (NSUInteger)hash
+{
+    NSUInteger hash = [super hash];
+    for (auto coord : _coordinates) {
+        hash += @(coord.latitude+coord.longitude).hash;
+    }
+    return hash;
 }
 
 - (CLLocationCoordinate2D)coordinate
--- a/platform/darwin/src/MGLShape.mm
+++ b/platform/darwin/src/MGLShape.mm
@@ -2,6 +2,15 @@
 
 #import "MGLFeature_Private.h"
 
+#import "NSString+MGLAdditions.h"
+#import "MGLTypes.h"
+
+#import <mbgl/util/geo.hpp>
+
+bool operator==(const CLLocationCoordinate2D lhs, const CLLocationCoordinate2D rhs) {
+    return lhs.latitude == rhs.latitude && lhs.longitude == rhs.longitude;
+}
+
 @implementation MGLShape
 
 + (nullable MGLShape *)shapeWithData:(NSData *)data encoding:(NSStringEncoding)encoding error:(NSError * _Nullable *)outError {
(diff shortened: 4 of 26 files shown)


platform/darwin/src/MGLPointAnnotation.m
```diff
@@ -4,6 +4,45 @@ @implementation MGLPointAnnotation
...
+    [coder encodeDouble:coordinate.longitude forKey:@"coordinateLongitude"];
+}
+
+- (BOOL)isEqual:(id)other
+{
+    if ( ! [other isKindOfClass:[self class]])
+    {
+        return NO;
+    }
+    if (other == self)
```
1ec5: This should be the first check for better performance.

platform/darwin/src/MGLPointAnnotation.m
```diff
@@ -4,6 +4,45 @@ @implementation MGLPointAnnotation
...
+}
+
+- (void)encodeWithCoder:(NSCoder *)coder
+{
+    [super encodeWithCoder:coder];
+    [coder encodeDouble:coordinate.latitude forKey:@"coordinateLatitude"];
+    [coder encodeDouble:coordinate.longitude forKey:@"coordinateLongitude"];
+}
+
+- (BOOL)isEqual:(id)other
```
1ec5: If you override -isEqual:, you also need to override -hash.

1ec5:
Individual shape subclasses need custom implementations of these methods, because they can't be represented by only one coordinate. The MGLFeature classes also need overrides to check for attribute dictionary equality.

boundsj:
> Individual shape subclasses need custom implementations of these methods, because they can't be represented by only one coordinate. The MGLFeature classes also need overrides to check for attribute dictionary equality.


The `geoJSONGeometryDictionary` method added to `MGLShape` and `geoJSONFeatureDictionary` method added to `MGLFeature` coming in https://github.com/mapbox/mapbox-gl-native/pull/6524 may help.

frederoni:
MGLFeature is going to be quite bloated. Thinking about splitting it up in one class per file.

platform/darwin/src/MGLFeature.mm
```diff
@@ -25,6 +25,44 @@ @implementation MGLPointFeature
...
+    return YES;
+}
+
+- (instancetype)initWithCoder:(NSCoder *)decoder
+{
+    self = [super initWithCoder:decoder];
+    if (self)
+    {
+        NSSet<Class> *identifierClasses = [NSSet setWithArray:@[NSString.class, NSNumber.class]];
+        identifier = [decoder decodeTopLevelObjectOfClasses:identifierClasses forKey:@"identifier" error:nil];
```
frederoni: iOS9 only

platform/darwin/src/MGLMultiPoint.mm
```diff
@@ -5,6 +5,37 @@
 #include <mbgl/util/geo.hpp>
 #include <mbgl/util/optional.hpp>
 
+bool compareByCoordinate(CLLocationCoordinate2D lhs, CLLocationCoordinate2D rhs)
+{
+    return lhs.latitude+lhs.longitude < rhs.latitude+rhs.longitude;
```
frederoni: This means coordinate{0,10} is the same as {10,0} but it's only used for sorting and as long as lhs and rhs is sorted the same way it shouldn't matter.

platform/darwin/src/MGLShape.mm
```diff
@@ -1,7 +1,62 @@
...
+#endif
+}
+
+- (BOOL)isEqual:(id)other
+{
+    if (other == self) { return YES; }
+    id <MGLAnnotation> annotation = other;
+    
+#if TARGET_OS_MACOS
+    return (MGLIsEqualToString(_title, [annotation title])
```
1ec5: Both annotations’ titles being nil should not cause this test to fail.

platform/darwin/src/MGLTypes.h
```diff
@@ -18,6 +18,12 @@
 #define MGLColor NSColor
 #endif
 
+#define TARGET_OS_MACOS (TARGET_OS_MAC && !(TARGET_OS_IOS || TARGET_OS_TV || TARGET_OS_WATCH))
+
+#define MGLIsEqual(lhs,rhs) ((!lhs && !rhs) || [lhs isEqual:rhs])
+#define MGLIsEqualToString(lhs,rhs) ((!lhs && !rhs) || [lhs isEqualToString:rhs])
+#define MGLIsEqualToDictionary(lhs,rhs) ((!lhs && !rhs) || [lhs isEqualToDictionary:rhs])
```
1ec5: This SDK should not get into the business of providing general-purpose Foundation equality macros to developers.

platform/darwin/src/MGLTypes.h
```diff
@@ -18,6 +18,12 @@
 #define MGLColor NSColor
 #endif
 
+#define TARGET_OS_MACOS (TARGET_OS_MAC && !(TARGET_OS_IOS || TARGET_OS_TV || TARGET_OS_WATCH))
```
1ec5: Let’s stick to a simple `!TARGET_OS_IPHONE` check at each call site for now. By the time we introduce support for tvOS and/or watchOS, we’ll’ve dropped support for macOS 10.10 and will be able to use the standard `TARGET_OS_OSX` macro anyways.

platform/darwin/src/MGLMultiPoint.mm
```diff
@@ -27,6 +58,61 @@ - (instancetype)initWithCoordinates:(const CLLocationCoordinate2D *)coords count
...
+        [coordinates addObject:@[@(coord.latitude), @(coord.longitude)]];
+    }
+    [coder encodeObject:coordinates forKey:@"coordinates"];
+}
+
+- (BOOL)isEqual:(id)other
+{
+    if (self == other) return YES;
+    
+    MGLMultiPoint *otherMultipoint = other;
```
1ec5: What if `other` is an instance of some other class that implements `coordinates` as an NSSet?

platform/darwin/src/MGLMultiPoint.mm
```diff
@@ -27,6 +58,61 @@ - (instancetype)initWithCoordinates:(const CLLocationCoordinate2D *)coords count
...
+    CLLocationCoordinate2D *otherCoordinates = [otherMultipoint coordinates];
+    std::vector<CLLocationCoordinate2D> otherCoords;
+    otherCoords.reserve([otherMultipoint pointCount]);
+    
+    for (NSUInteger i = 0; i < [otherMultipoint pointCount]; i++) {
+        otherCoords.push_back(otherCoordinates[i]);
+    }
+    
+    return ([super isEqual:otherMultipoint]
+            && _coordinates == otherCoords);
```
1ec5: This could be written as `_coordinates == otherCoords->_coordinates` to avoid having to build a copy of the vector in the first place.

platform/darwin/src/MGLMultiPoint.mm
```diff
@@ -27,6 +58,61 @@ - (instancetype)initWithCoordinates:(const CLLocationCoordinate2D *)coords count
...
+}
+
+- (BOOL)isEqual:(id)other
+{
+    if (self == other) return YES;
+    
+    MGLMultiPoint *otherMultipoint = other;
+    
+    CLLocationCoordinate2D *otherCoordinates = [otherMultipoint coordinates];
+    std::vector<CLLocationCoordinate2D> otherCoords;
```
1ec5: This can be written as `otherCoords = { otherCoordinates, [otherMultiPoint pointCount] }` to copy the array with the same efficiency as a `memcpy()`.

platform/darwin/src/MGLMultiPoint.mm
```diff
@@ -27,6 +58,61 @@ - (instancetype)initWithCoordinates:(const CLLocationCoordinate2D *)coords count
...
+    return self;
+}
+
+- (void)encodeWithCoder:(NSCoder *)coder
+{
+    [super encodeWithCoder:coder];
+    
+    NSMutableArray *coordinates = [NSMutableArray array];
+    for (auto coord : _coordinates) {
+        [coordinates addObject:@[@(coord.latitude), @(coord.longitude)]];
```
1ec5: We should never store a coordinate pair as an array if we can help it. Instead, represent it as a dictionary so that the order is never ambiguous.

1ec5: Better yet, stuff the CLLocationCoordinate2D in an NSValue, which already conforms to NSSecureCoding.

frederoni: `@encode` CLLocationCoordinate2D would work as long as we don't use it in combination with NSKeyed(Un)Archiver. It only gets the layout of the struct but no information about the name of the fields which is causing the archiver to raise an exception when trying to archive structs. Will make it unambiguous by representing it as a dictionary instead.

platform/darwin/src/MGLShape.mm
```diff
@@ -1,7 +1,95 @@
...
+
+bool operator!=(const CLLocationCoordinate2D lhs, const CLLocationCoordinate2D rhs) {
+    return lhs.latitude != rhs.latitude || lhs.longitude != rhs.longitude;
+}
+
+bool operator==(std::vector<CLLocationCoordinate2D>& lhs, std::vector<CLLocationCoordinate2D>& rhs)
+{
+    if (lhs.size() != rhs.size()) return false;
+    
+    std::sort(lhs.begin(), lhs.end(), compareByCoordinate);
```
1ec5: Order matters. Two coincident lines, one of which goes in the opposite direction as the other, are not equal.

platform/darwin/src/MGLMultiPoint.mm
1ec5: CLLocationCoordinate2D does work with keyed archivers and unarchivers, as long as you use NSValue to wrap it. Archiving a series of CLLocationCoordinate2Ds would be more efficient than archiving a series of dictionaries, and I don't think there's any ambiguity with that approach because we don't define an MGLSwappedCoordinate2D struct anywhere.

platform/darwin/src/NSArray+MGLAdditions.mm
```diff
@@ -23,4 +23,29 @@ @implementation NSArray (MGLAdditions)
     return vector;
 }
 
++ (NSArray *)mgl_coordinatesFromCoordinates:(std::vector<CLLocationCoordinate2D>)coords {
```
1ec5: Let's categorize NSCoder to have -encodeMGLLocationCoordinate2D:forKey: and -decodeMGLLocationCoordinate2DForKey:, as has been done for other structs like CGSize and UIEdgeInsets.

platform/darwin/src/NSCoder+MGLAdditions.h
```diff
@@ -0,0 +1,12 @@
+#import <Foundation/Foundation.h>
+#import <CoreLocation/CoreLocation.h>
+
+#import <mbgl/util/feature.hpp>
+
+@interface NSCoder (MGLAdditions)
+
+- (void)mgl_encodeLocationCoordinates2D:(std::vector<CLLocationCoordinate2D>)coordinates forKey:(NSString *)key;
```
1ec5: I think it would be useful to publicly expose a method that encodes a single CLLocationCoordinate2D. It would be called `-encodeMGLCoordinate:`, for consistency with existing category methods on NSCoder. (See also `-[NSValue valueWithMGLCoordinate:]`.)

platform/darwin/src/NSCoder+MGLAdditions.mm
```diff
@@ -0,0 +1,16 @@
+#import "NSCoder+MGLAdditions.h"
+
+#import "NSArray+MGLAdditions.h"
+
+@implementation NSCoder (MGLAdditions)
+
+- (void)mgl_encodeLocationCoordinates2D:(std::vector<CLLocationCoordinate2D>)coordinates forKey:(NSString *)key {
+    [self encodeObject:[NSArray mgl_coordinatesFromCoordinates:coordinates] forKey:key];
```
1ec5: Now that we have dedicated methods for working with `CLLocationCoordinate2D`s, let’s encode the result of `-[NSValue(MGLAdditions) valueWithMGLCoordinate:]` instead of a dictionary.

frederoni: I use NSKeyedArchiver for testing NSCoding which can't encode structs, not even structs wrapped in NSValue encoded with `@encode(CLLocationCoordinate2D)`. (```[NSKeyedArchiver encodeValueOfObjCType:at:]: this archiver cannot encode structs```) Any other idea how to solve this? UIStateRestoring will probably still work if we switch to NSValue but a solution that works with archiving as well would be preferred.

1ec5: I stand corrected. Apparently the [root cause](http://stackoverflow.com/a/12292033) is that CLLocationCoordinate2D is a typedef to an anonymous struct, so `@encode` can't do anything useful.

We could hard-code the type string `{dd}`, as many Swift developers have had to do, or we could define our own non-anonymous struct based on CLLocationCoordinate2D solely for the purpose of archiving, or we could eat the performance cost and use dictionaries here.

1ec5: Yeah, looks like you’re right. Unfortunately we’ll have to stick to an array of dictionaries. If performance does turn out to be a major issue, we can flatten the array, at the cost of triggering another debate about latitude-longitude order versus longitude-latitude order (or is it longitude-latitude order versus latitude-longitude order?).

platform/darwin/src/MGLFeature_Private.h
```diff
@@ -37,3 +37,31 @@ mbgl::Feature mbglFeature(mbgl::Feature feature, id identifier, NSDictionary *at
 NS_DICTIONARY_OF(NSString *, id) *NSDictionaryFeatureForGeometry(NSDictionary *geometry, NSDictionary *attributes, id identifier);
 
 NS_ASSUME_NONNULL_END
+
+#define MGLFeatureInitCoding() \
```
1ec5: Macros should be named in SCREAMING_SNAKE_CASE. 📣🐍 In this case, since this macro is defining a method, it should be named with a verb, like `MGL_DEFINE_FEATURE_INIT_WITH_CODER()`.

platform/darwin/src/MGLFeature_Private.h
```diff
@@ -37,3 +37,31 @@ mbgl::Feature mbglFeature(mbgl::Feature feature, id identifier, NSDictionary *at
...
+- (instancetype)initWithCoder:(NSCoder *)decoder { \
+if (self = [super initWithCoder:decoder]) { \
```
1ec5: Indent the macro to aid with readability. This line should start with two tabs.

platform/darwin/src/MGLFeature_Private.h
```diff
@@ -37,3 +37,31 @@ mbgl::Feature mbglFeature(mbgl::Feature feature, id identifier, NSDictionary *at
...
+[super encodeWithCoder:coder]; \
+[coder encodeObject:identifier forKey:@"identifier"]; \
+[coder encodeObject:attributes forKey:@"attributes"]; \
+}
+
+#define MGLFeatureIsEqual(klass) \
+- (BOOL)isEqual:(id)other { \
+if (other == self) return YES; \
+if (![other isKindOfClass:[self class]]) return NO; \
+klass *otherFeature = other; \
```
1ec5: `typeof(self)` also works in place of `klass *`, since we’ve already asserted that `other` is of the same kind as `self`.

platform/darwin/src/NSCoder+MGLAdditions.h
```diff
@@ -0,0 +1,10 @@
+#import <Foundation/Foundation.h>
+#import <CoreLocation/CoreLocation.h>
+
+@interface NSCoder (MGLAdditions)
+
+- (void)encodeMGLCoordinate:(CLLocationCoordinate2D)coordinate forKey:(NSString *)key;
```
1ec5: Publicly vending these methods would’ve made more sense if we could come up with a way to encode and decode the coordinate as a struct. But with the dictionary-based implementation is dependent on the keys we choose, so unfortunately it isn’t portable enough to expose publicly. MGLCode+MGLAdditions_Private.h can be folded into this header, which can be marked with Project visibility. Sorry for the detour.

platform/darwin/src/MGLFeature_Private.h
```diff
@@ -37,3 +37,31 @@ mbgl::Feature mbglFeature(mbgl::Feature feature, id identifier, NSDictionary *at
 NS_DICTIONARY_OF(NSString *, id) *NSDictionaryFeatureForGeometry(NSDictionary *geometry, NSDictionary *attributes, id identifier);
 
 NS_ASSUME_NONNULL_END
+
+#define MGL_DEFINE_FEATURE_INIT_WITH_CODER() \
+- (instancetype)initWithCoder:(NSCoder *)decoder { \
```
1ec5: Also indent this line and the closing brace.

PR #6565, opened 2016-10-03
[ios, macos] fixes #6160: allow updating multipoint coordinates

### LINKED ISSUE #6160: MGLPolygon, MGLPolyline coordinates should be mutable
Just as it’s very convenient to be able to change an MGLPointAnnotation’s `coordinate` property and have the annotation move on the fly, it should be possible to change the shape of an MGLPolygon or MGLPolyline. This would be useful for GPS and fitness tracking applications as well as applications that allow the user to trace features over the map.

MGLMultiPoint has a `coordinates` property that – due to Objective-C language limitations – holds a C array of CLLocationCoordinate2Ds, rather than an NSArray that we’d easily be able to make mutable. Nonetheless, we could pair this property with one that holds an NSMutableArray of NSValues representing CLLocationCoordinate2Ds, then implement all the to-many getters and setters expected by KVO. Alternatively, we could add a `-setCoordinates:range:` method. Either approach will be somewhat messy in Objective-C and even messier when bridged to Swift.

This is the iOS/macOS equivalent of #5224 on Android. I consider this feature blocked by #5844. The awkwardness of the API we’d add would outweigh any benefits to implementing a stopgap API like in #5848 on Android.

/cc @incanus @pveugen

incanus:
This adds `-[MGLMultiPoint updateCoordinates:count:]` and pipes to core, which immediately updates display of `MGLPolyline` or `MGLPolygon` annotations. 

Several design considerations: 

- Went with `updateCoordinates:...` naming since merely making the `coordinates` property `readwrite` (effectively) as an analog to `-[MGLPointAnnotation coordinate]` is not doable with C arrays of `CLLocationCoordinate2D` structs. 

- Was a little duplicative with internal comments and deselection in `MGLMapView` for the sake of clarity and safety. 

- Tried to keep `free` and `malloc` together in scope for future maintenance safety. 

/cc @1ec5 @friedbunny @boundsj


incanus:
You can test this with a small patch such as the following: 

```diff
diff --git a/platform/ios/app/MBXViewController.m b/platform/ios/app/MBXViewController.m
index feebb59..66a8158 100644
--- a/platform/ios/app/MBXViewController.m
+++ b/platform/ios/app/MBXViewController.m
@@ -170,6 +170,20 @@ typedef NS_ENUM(NSInteger, MBXSettingsMiscellaneousRows) {
         }
         [self presentViewController:alertController animated:YES completion:nil];
     }
+
+    [self performSelector:@selector(addTestShapes) withObject:nil afterDelay:1];
+}
+
+- (void)mapViewRegionIsChanging:(MGLMapView *)mapView
+{
+    CLLocationCoordinate2D center = mapView.centerCoordinate;
+    CLLocationCoordinate2D coordinates[2] = {
+        CLLocationCoordinate2DMake(45, -122),
+        center
+    };
+
+    MGLPolyline *line = (MGLPolyline *)mapView.annotations.firstObject;
+    [line updateCoordinates:coordinates count:2];
 }
 
 - (void)saveState:(__unused NSNotification *)notification
@@ -539,6 +553,24 @@ typedef NS_ENUM(NSInteger, MBXSettingsMiscellaneousRows) {
 
 - (void)addTestShapes
 {
+    CLLocationCoordinate2D center = self.mapView.centerCoordinate;
+    CLLocationCoordinate2D lineCoordinates[2] = {
+        CLLocationCoordinate2DMake(45, -122),
+        center
+    };
+    MGLPolyline *line = [MGLPolyline polylineWithCoordinates:lineCoordinates count:2];
+    [self.mapView addAnnotation:line];
+
+//    CLLocationCoordinate2D polygonCoordinates[3] = {
+//        CLLocationCoordinate2DMake(44, -122),
+//        CLLocationCoordinate2DMake(46, -122),
+//        center
+//    };
+//    MGLPolygon *shape = [MGLPolygon polygonWithCoordinates:polygonCoordinates count:3];
+//    [self.mapView addAnnotation:shape];
+
+    return;
+
     // Pacific Northwest triangle
     //
     CLLocationCoordinate2D triangleCoordinates[3] =
```

1ec5:
For consistency with -getCoordinates:inRange:, the new method should be called -setCoordinates:inRange:. This allows the developer to update the polyline or polygon piecemeal.

platform/darwin/src/MGLMultiPoint.mm
```diff
@@ -93,6 +100,22 @@ - (void)getCoordinates:(CLLocationCoordinate2D *)coords range:(NSRange)range
     }
 }
 
+- (void)updateCoordinates:(CLLocationCoordinate2D *)coords count:(NSUInteger)count
+{
+    if ([self isMemberOfClass:[MGLMultiPoint class]])
+    {
+        [[NSException exceptionWithName:@"MGLAbstractClassException"
+                                 reason:@"MGLMultiPoint is an abstract class"
```
1ec5: MGLMultiPoint is no longer technically an abstract class, since we're overloading it to represent the GeoJSON MultiPoint concept as well. The exception here is a good idea, but we should move the declaration of this method to MGLPolyline and MGLPolygon individually.

incanus:
Going to implement @1ec5's suggestion; this is better API.

platform/darwin/src/MGLMultiPoint.mm
```diff
@@ -93,6 +100,22 @@ - (void)getCoordinates:(CLLocationCoordinate2D *)coords range:(NSRange)range
...
+                               userInfo:nil] raise];
+    }
+
+    assert(_count > 0);
```
1ec5: For internal errors in Objective-C code, prefer NSAssert over a C++ assert. In this case, we're catching a developer error, so it should be a full-blown NSException.

incanus:
What about also having just `setCoordinates:count:`for simplicity's sake? A `count` is easier to form than a `range`.

1ec5:
> What about also having just -setCoordinates:count:for simplicity's sake? A count is easier to form than a range.

With a range, the developer could potentially append to the polyline without having to rebuild the whole polyline from scratch.

incanus:
Sorry, I mean in addition: 

```objc
- (void)setCoordinates:(CLLocationCoordinate2D *)coords range:(NSRange)range;
- (void)setCoordinates:(CLLocationCoordinate2D *)coords count:(NSUInteger)count;
```

friedbunny:
@incanus Please also add a changelog entry. I imagine we should create a port-to-macOS ticket as well, for API additions there.

platform/darwin/src/MGLMultiPoint.mm
incanus: Was following the pattern in `-getCoordinates:range:`: 

```objc
assert(range.location + range.length <= _count);
```

1ec5: OK, I guess we missed that too when we overloaded this class to represent multipoints.

platform/ios/src/MGLMapView.mm
```diff
@@ -2840,6 +2863,8 @@ - (void)addAnnotations:(NS_ARRAY_OF(id <MGLAnnotation>) *)annotations
             MGLAnnotationContext context;
             context.annotation = annotation;
             _annotationContextsByAnnotationTag[annotationTag] = context;
+
+            [(NSObject *)annotation addObserver:self forKeyPath:@"coordinates" options:0 context:(void *)(NSUInteger)annotationTag];
```
1ec5: For MGLPolygon, also observe changes to interiorPolygons.

1ec5: Although this could be tail work, since it would require making the interiorPolygons property mutable or at least settable.

incanus:
Ok, range-based updating works. Here's constantly adding the map center to a polyline as the map moves: 

weird

incanus:
Updated testing patch: 

```diff
diff --git a/platform/ios/app/MBXViewController.m b/platform/ios/app/MBXViewController.m
index feebb59..443265d 100644
--- a/platform/ios/app/MBXViewController.m
+++ b/platform/ios/app/MBXViewController.m
@@ -170,6 +170,28 @@ typedef NS_ENUM(NSInteger, MBXSettingsMiscellaneousRows) {
         }
         [self presentViewController:alertController animated:YES completion:nil];
     }
+
+    [self performSelector:@selector(addTestShapes) withObject:nil afterDelay:1];
+}
+
+- (void)mapViewRegionIsChanging:(MGLMapView *)mapView
+{
+    // update tail coordinate
+//    CLLocationCoordinate2D center = mapView.centerCoordinate;
+//    MGLPolyline *line = (MGLPolyline *)mapView.annotations.firstObject;
+//    CLLocationCoordinate2D *coordinates = malloc(2 * sizeof(CLLocationCoordinate2D));
+//    [line getCoordinates:coordinates range:NSMakeRange(0, 2)];
+//    coordinates[1] = center;
+//    [line setCoordinates:coordinates range:NSMakeRange(0, 2)];
+//    free(coordinates);
+
+    // append coordinate to tail
+    CLLocationCoordinate2D center = mapView.centerCoordinate;
+    MGLPolyline *line = (MGLPolyline *)mapView.annotations.firstObject;
+    CLLocationCoordinate2D *coordinates = malloc(1 * sizeof(CLLocationCoordinate2D));
+    coordinates[0] = center;
+    [line setCoordinates:coordinates range:NSMakeRange(line.pointCount, 1)];
+    free(coordinates);
 }
 
 - (void)saveState:(__unused NSNotification *)notification
@@ -539,6 +561,24 @@ typedef NS_ENUM(NSInteger, MBXSettingsMiscellaneousRows) {
 
 - (void)addTestShapes
 {
+    CLLocationCoordinate2D center = self.mapView.centerCoordinate;
+    CLLocationCoordinate2D lineCoordinates[2] = {
+        CLLocationCoordinate2DMake(45, -122),
+        center
+    };
+    MGLPolyline *line = [MGLPolyline polylineWithCoordinates:lineCoordinates count:2];
+    [self.mapView addAnnotation:line];
+
+//    CLLocationCoordinate2D polygonCoordinates[3] = {
+//        CLLocationCoordinate2DMake(44, -122),
+//        CLLocationCoordinate2DMake(46, -122),
+//        center
+//    };
+//    MGLPolygon *shape = [MGLPolygon polygonWithCoordinates:polygonCoordinates count:3];
+//    [self.mapView addAnnotation:shape];
+
+    return;
+
     // Pacific Northwest triangle
     //
     CLLocationCoordinate2D triangleCoordinates[3] =
```

platform/darwin/src/MGLMultiPoint.mm
incanus: e7888e1886cf50cce4397635473adfe3adf0d321

platform/darwin/src/MGLMultiPoint.mm
incanus: @1ec5 Do you have more context on this? I'm not sure I understand the GeoJSON concept you are referring to.

incanus:
macOS is set as of 0349363186c3d109ecafdda50c146947004d2460. Test patch here: 

```diff
diff --git a/platform/macos/app/MapDocument.m b/platform/macos/app/MapDocument.m
index 42398ff..2f96522 100644
--- a/platform/macos/app/MapDocument.m
+++ b/platform/macos/app/MapDocument.m
@@ -99,6 +99,37 @@ NS_ARRAY_OF(id <MGLAnnotation>) *MBXFlattenedShapes(NS_ARRAY_OF(id <MGLAnnotatio
     [self.mapView addGestureRecognizer:pressGestureRecognizer];
     
     [self applyPendingState];
+
+    [self performSelector:@selector(addTestShape) withObject:nil afterDelay:1];
+}
+
+- (void)addTestShape {
+    CLLocationCoordinate2D center = self.mapView.centerCoordinate;
+    CLLocationCoordinate2D lineCoordinates[2] = {
+        CLLocationCoordinate2DMake(45, -122),
+        center
+    };
+    MGLPolyline *line = [MGLPolyline polylineWithCoordinates:lineCoordinates count:2];
+    [self.mapView addAnnotation:line];
+}
+
+- (void)mapView:(MGLMapView *)mapView cameraDidChangeAnimated:(BOOL)animated {
+//    update tail coordinate
+//    CLLocationCoordinate2D center = mapView.centerCoordinate;
+//    MGLPolyline *line = (MGLPolyline *)mapView.annotations.firstObject;
+//    CLLocationCoordinate2D *coordinates = malloc(2 * sizeof(CLLocationCoordinate2D));
+//    [line getCoordinates:coordinates range:NSMakeRange(0, 2)];
+//    coordinates[1] = center;
+//    [line setCoordinates:coordinates range:NSMakeRange(0, 2)];
+//    free(coordinates);
+
+    // append coordinate to tail
+    CLLocationCoordinate2D center = mapView.centerCoordinate;
+    MGLPolyline *line = (MGLPolyline *)mapView.annotations.firstObject;
+    CLLocationCoordinate2D *coordinates = malloc(1 * sizeof(CLLocationCoordinate2D));
+    coordinates[0] = center;
+    [line setCoordinates:coordinates range:NSMakeRange(line.pointCount, 1)];
+    free(coordinates);
 }
 
 - (NSWindow *)window {
```

platform/darwin/src/MGLMultiPoint.mm
```diff
@@ -89,7 +89,7 @@ - (void)getCoordinates:(CLLocationCoordinate2D *)coords range:(NSRange)range
                                userInfo:nil] raise];
     }
 
-    assert(range.location + range.length <= _count);
+    NSAssert(range.location + range.length <= _count, @"Invalid coordinate range");
```
1ec5: These should be NSExceptions. NSAssert is for errors we make, whereas NSException is for errors the developer makes.

1ec5: Also describe the invalid range and indicate the correct count in the exception message, to aid in debugging.

platform/ios/CHANGELOG.md
```diff
@@ -40,6 +40,7 @@ Mapbox welcomes participation and contributions from everyone. Please read [CONT
 
 ### Annotations
 
+* MGLPolyline and MGLPolygon annotations are now able to be mutated, part or all, and changes are displayed immediately. ([#6565](https://github.com/mapbox/mapbox-gl-native/pull/6565))
```
1ec5: Since we don’t respond to mutations to inner polygons, let’s say that MGLPolyline and the exterior of MGLPolygon are mutable.

platform/darwin/src/MGLMultiPoint.mm
1ec5: For the purposes of feature querying (but not annotations yet), we use this class to represent GeoJSON [`MultiPoint`s](http://geojson.org/geojson-spec.html#multipoint), which are just collections of points with no vertices connecting them. #5201 tracks cleaning up the class hierarchy so that MGLPolyline and MGLPolygon no longer inherit from MGLMultiPoint.

platform/darwin/src/MGLMultiPoint.h
```diff
@@ -35,6 +35,20 @@ NS_ASSUME_NONNULL_BEGIN
...
 - (void)getCoordinates:(CLLocationCoordinate2D *)coords range:(NSRange)range;
 
+/**
+ Updates one or more coordinates for the shape, which will instantaneously 
+ cause the shape to be redrawn if it is currently visible on the map.
+ 
+ @param coords The array of coordinates defining the shape. The data in this
+ array is copied to the object.
+ @param range The range of points to update. The `location` field indicates the
+ first point you are requesting, with `0` being the first point, `1` being
```
1ec5: Nit: we’re updating points, not requesting them.

platform/darwin/src/MGLMultiPoint.mm
```diff
@@ -93,6 +98,45 @@ - (void)getCoordinates:(CLLocationCoordinate2D *)coords range:(NSRange)range
...
+    {
+        [[NSException exceptionWithName:@"MGLAbstractClassException"
+                                 reason:@"MGLMultiPoint is an abstract class"
+                               userInfo:nil] raise];
+    }
+
+    NSAssert(range.length > 0, @"Empty coordinate range");
+
+    [self willChangeValueForKey:@"coordinates"];
+    if (range.location >= _count)
```
1ec5: There are a few different cases we need to consider. The current code doesn’t address all of them. I think this code would be simpler with the following algorithm:

1. The new range is entirely contained within `_coordinates`, in which case `NSMaxRange(range) <= _count`. Patch up `_coordinates` using `memcpy()`. (Note that the `num` parameter to `memcpy()` is the number of bytes.)
2. The new range begins within `_coordinates` but extends beyond it, in which case `NSMaxRange(range) > _count`. Allocate a new array on the heap, `memcpy()` `_coordinates` to the beginning of it, then set `_coordinates` to that new array.
3. The new range doesn’t overlap `_coordinates` at all, in which case `range.location > _count`. Raise an exception.

Since (1) will reuse the memory that `_coordinates` already points to, we should avoid freeing that memory unless we know we’re in (2) and we’re already done with the old array.

platform/darwin/src/MGLMultiPoint.mm
1ec5: For the purposes of feature querying (but not annotations yet), we use this class to represent GeoJSON [`MultiPoint`s](http://geojson.org/geojson-spec.html#multipoint), which are just collections of points with no vertices connecting them. #5201 tracks cleaning up the class hierarchy so that MGLPolyline and MGLPolygon no longer inherit from MGLMultiPoint.

incanus: 👉 https://github.com/mapbox/mapbox-gl-native/issues/5201#issuecomment-251512297

1ec5: Well, no, once 5201 lands, this exception and the others like it won’t even be necessary anymore. But in the meantime, all I’m suggesting is that we say:

```objc
[NSException raise:@"Unsupported method" format:@"The coordinates of a member of MGLMultiPoint are read-only."];
```

platform/darwin/src/MGLMultiPoint.h
incanus: a24482a7bebf671c77579d077b9533bf71b9987e

platform/ios/CHANGELOG.md
incanus: 608115e07d97bc200606f9f0c170172256f0bd21

platform/darwin/src/MGLMultiPoint.mm
incanus: Let me know how 2d46f20d8b0713a35d7fa97729fb5db285cb098a looks @1ec5. I think what I did there is kosher.

incanus:
Ok, lots of work today and ready for review again. Let's 🚢this 🦃.

platform/darwin/src/MGLMultiPoint.mm
```diff
@@ -93,6 +84,47 @@ - (void)getCoordinates:(CLLocationCoordinate2D *)coords range:(NSRange)range
     }
 }
 
+- (void)setCoordinates:(CLLocationCoordinate2D *)coords range:(NSRange)range
+{
+    if ([self isMemberOfClass:[MGLMultiPoint class]])
```
1ec5: Oof, I totally forgot this method is defined on MGLMultiPoint rather than MGLPolyline and MGLPolygon individually. In that case, this check should go away too, because there’s no reason you couldn’t get and set the coordinates of an MGLMultiPoint instance.

platform/darwin/src/MGLMultiPoint.mm
```diff
@@ -93,6 +84,47 @@ - (void)getCoordinates:(CLLocationCoordinate2D *)coords range:(NSRange)range
...
+    {
+        [NSException raise:@"Unsupported method" format:@"The coordinates of a member of MGLMultiPoint are read-only."];
+    }
+
+    if (range.length == 0)
+    {
+        [[NSException exceptionWithName:NSRangeException
```
1ec5: This would be more concise with a call to `+[NSException raise:format:]`.

platform/darwin/src/MGLMultiPoint.mm
```diff
@@ -93,6 +84,47 @@ - (void)getCoordinates:(CLLocationCoordinate2D *)coords range:(NSRange)range
...
+                               userInfo:nil] raise];
+    }
+
+    [self willChangeValueForKey:@"coordinates"];
+    if (NSMaxRange(range) <= _count)
+    {
+        // replacing existing coordinate(s)
+        memcpy(_coordinates + range.location, coords, range.length * sizeof(CLLocationCoordinate2D));
+    }
+    else if (NSMaxRange(range) > _count)
```
1ec5: Nit: this can just be `else`.

platform/darwin/src/MGLMultiPoint.mm
```diff
@@ -93,6 +84,47 @@ - (void)getCoordinates:(CLLocationCoordinate2D *)coords range:(NSRange)range
...
+    else if (range.location > _count)
+    {
+        [[NSException exceptionWithName:NSRangeException
+                                 reason:[NSString stringWithFormat:
+                                            @"Invalid range %@ for existing coordinate count %zu",
+                                            NSStringFromRange(range), _count]
+                               userInfo:nil] raise];
+    }
+
+    [self willChangeValueForKey:@"coordinates"];
```
1ec5: `pointCount` can also change. Might be worth implementing `+keyPathsForValuesAffectingPointCount:`.

platform/darwin/src/MGLMultiPoint.mm
```diff
@@ -93,6 +84,47 @@ - (void)getCoordinates:(CLLocationCoordinate2D *)coords range:(NSRange)range
...
+    {
+        // replacing existing coordinate(s)
+        memcpy(_coordinates + range.location, coords, range.length * sizeof(CLLocationCoordinate2D));
+    }
+    else if (NSMaxRange(range) > _count)
+    {
+        // appending new coordinate(s)
+        NSUInteger newCount = NSMaxRange(range);
+        CLLocationCoordinate2D *newCoordinates = (CLLocationCoordinate2D *)malloc(newCount * sizeof(CLLocationCoordinate2D));
+        memcpy(newCoordinates, _coordinates, _count * sizeof(CLLocationCoordinate2D));
```
1ec5: If the developer is overwriting the tail end of the array while also extending it, we should avoid unnecessary work: take the minimum of `_count` and `range.location`.

1ec5: Now that the address that the `coordinates` property points to can change, we need a warning on that property’s documentation along the lines of [`-[NSString UTF8String]`](https://developer.apple.com/reference/foundation/nsstring/1411189-utf8string?language=objc) (another example of an `NS_RETURNS_INNER_POINTER` method).

platform/darwin/src/MGLMultiPoint.mm
```diff
@@ -93,6 +84,47 @@ - (void)getCoordinates:(CLLocationCoordinate2D *)coords range:(NSRange)range
...
+                                            @"Invalid range %@ for existing coordinate count %zu",
+                                            NSStringFromRange(range), _count]
+                               userInfo:nil] raise];
+    }
+
+    [self willChangeValueForKey:@"coordinates"];
+    if (NSMaxRange(range) <= _count)
+    {
+        // replacing existing coordinate(s)
+        memcpy(_coordinates + range.location, coords, range.length * sizeof(CLLocationCoordinate2D));
```
1ec5: [`memcpy()` documentation](http://www.cplusplus.com/reference/cstring/memcpy/) states that the destination and source of a copy mustn’t overlap, or Bad Things™ can happen. So we should avoid calling `memcpy()` in the case that either `coords` or `coords + coords.length` falls between `_coordinates + range.location` and `_coordinates + NSMaxRange(range)`. Since this is a bit of an edge case – you’d have to pass in the `coordinates` property as `coords` – it’s fine to punt on that by raising an exception, but we should ticket it out as tail work.

platform/darwin/src/MGLMultiPoint.h
```diff
@@ -35,6 +35,22 @@ NS_ASSUME_NONNULL_BEGIN
...
+ array is copied to the object.
+ @param range The range of points to update. The `location` field indicates the
+ first point you are replacing, with `0` being the first point, `1` being
+ the second point, and so on. The `length` field indicates the number of
+ points to update. You can append to an existing array of coordinates
+ by specifying a range with a `location` at the end of the existing array 
+ and/or a `length` which extends past the end of the existing array. The array 
+ in _`coords`_ must be equal in number to the length of the range.
+ */
+- (void)setCoordinates:(CLLocationCoordinate2D *)coords range:(NSRange)range;
```
1ec5: Something didn’t feel right to me, and now I know why: we’re only allowing the developer to overwrite or extend the multipoint but not shrink it. They can splice _in_ but not splice _out_. We’re assuming that `coords`’ range is equal to `range.length`, whereas the length would have to be passed in explicitly.

I don’t think we should take the time right now to implement all the possible splicing operations, but we should try to be consistent with [naming convention for KVO indexed accessors](https://developer.apple.com/library/content/documentation/Cocoa/Conceptual/KeyValueCoding/Articles/AccessorConventions.html#//apple_ref/doc/uid/20002174-SW11), which NSMutableArray and others adhere to. Ideally, we’d call this method `-replaceCoordinatesInRange:withCoordinates:count:` and have separate `-setCoordinates:count`, `-insertCoordinates:count:atIndex:`, and `-removeCoordinates:range:` methods. But for now, let’s call this method `-replaceCoordinatesInRange:withCoordinates:`. When we eventually implement `-replaceCoordinatesInRange:withCoordinates:count:`, we can deprecate this method.

1ec5:
#6580 would eliminate the manual C memory management and make it a lot easier (think one-liners) to implement the rest of the operations laid out in https://github.com/mapbox/mapbox-gl-native/pull/6565#discussion_r81871217.

platform/darwin/src/MGLMultiPoint.mm
incanus: ea0c900ae9e7ef920fb0d239da2a69dc72002209

platform/darwin/src/MGLMultiPoint.mm
incanus: d240ba93bfb019ef5cc530a9840d53e71b2c9ec9

platform/darwin/src/MGLMultiPoint.mm
incanus: a2377ed9d66b9bd6e9c35ad0f1425a30107cd46a

platform/darwin/src/MGLMultiPoint.mm
incanus: b9c10e373e165af7f4d2c8ff1d64e988a1c98a49

platform/darwin/src/MGLMultiPoint.mm
incanus: fcd5ada1cc784becba4bd09663f9985321e0b53a

incanus: I'll consider this handled by the https://github.com/mapbox/mapbox-gl-native/issues/6580 refactor and the fact that https://github.com/mapbox/mapbox-gl-native/commit/fcd5ada1cc784becba4bd09663f9985321e0b53a says `yet` as a cue.

platform/darwin/src/MGLMultiPoint.h
incanus: 138ae55fb8f5beacb1ae29ff987c939ab2128996

incanus:
Again all good to go. 😅

platform/darwin/src/MGLMultiPoint.mm
1ec5: This will catch the case in which the developer tries to close a polyline:

```objc
[polyline setCoordinates:polyline.coordinates range:NSMakeRange(polyline.count, 1)];
```

but not if they try to have it return to the midpoint:

```objc
[polyline setCoordinates:polyline.coordinates + polyline.pointCount / 2 range:NSMakeRange(polyline.count, 1)];
```

I think the check should be expanded to something like:

```objc
(coords >= _coordinates + range.location && coords < pointCount + NSMaxRange(range))
|| (coords + range.length >= pointCount + range.location && coords + range.length < pointCount + NSMaxRange(range))
```

By the way, in that case, it would be safe to call [`memmove()`](http://www.cplusplus.com/reference/cstring/memmove/) instead of `memcpy()`.

incanus: Will do. I'm going to avoid `memmove()` right now since it'll go away with the refactor.

incanus: The suggested comparison is **really** hard to follow, plus it doesn't compile since we can't do arithmetic on `void` pointers and integers, so I'm working on how to do this more clearly and safely. I think it's the last bit though?

incanus: I _think_ I got this ok in 7f6049b748d179afc281270d6bd8529141ceb296?

1ec5: 👌

platform/darwin/src/MGLMultiPoint.h
1ec5: → #6583

1ec5:
Thanks for the great work here @incanus! I’ve filed #6580 for the other half of this feature.

platform/darwin/src/MGLMultiPoint.mm
boundsj: I think this check may be off by 1 since `_count` is not 0 indexed and, when added to the array, it points at memory somewhere off the end of the array.

boundsj: Actually, the comparison is `<` so it is probably fine. Sorry!

nitrag:
Need guidance...I'm trying to add/update a Polyline on the fly.

```swift
var recordLine: MGLPolyline? = nil
func updateRecordingLine(_ coords: [CLLocationCoordinate2D]){
    var recordLineCoords = coords
    if(self.recordLine == nil){
        //add first time
        self.recordLine = MGLPolyline(coordinates: &recordLineCoords, count: UInt(recordLineCoords.count))
        self.mapView.addAnnotation(self.recordLine!)
    }else{
        //update
        self.recordLine?.replaceCoordinates(in: NSMakeRange(0, recordLineCoords.count), withCoordinates: &recordLineCoords)
    }
}
```

JesseCrocker:
Not being able to update a line with fewer coordinates than it previously had seems like an odd limitation.

incanus:
@JesseCrocker 👉 https://github.com/mapbox/mapbox-gl-native/issues/6583

PR #6571, opened 2016-10-04
Rename Raster => Texture

kkaefer:
We're currently using Raster as the name for our texture wrapper object, but we should standardize on `Texture`.


src/mbgl/util/offscreen_texture.cpp
```diff
@@ -55,12 +54,12 @@ void OffscreenTexture::bind(gl::Context& context,
...
 
-Raster& OffscreenTexture::getTexture() {
-    return raster;
+gl::Texture& OffscreenTexture::getTexture() {
+    return *texture;
 }
 
 std::array<uint16_t, 2> OffscreenTexture::getSize() const {
-    return raster.getSize();
+    return texture ? texture->size : std::array<uint16_t, 2>{{ 0, 0 }};
```
tmpsantos: Puzzles me why clang finds this ambiguous.

PR #6573, opened 2016-10-04
Update Runtime Style JavaDoc

### LINKED ISSUE #6510: javadoc runtime style
We are missing some javadoc for the new features.

Some lines in the style.layer are not documented:
 - inner classes
 - annotations

screen shot 2016-09-29 at 10 47 30

The explanation used in style.source is not always coherent when objects relate to each other. 
When a class has a sibling relation. The class should have the same language construct in javadoc. 

screen shot 2016-09-29 at 10 48 41

In general we can make the description of each class a bit longer. 
Examples are also nice additions when it comes to runtime styling.

cc @ivovandongen

### LINKED ISSUE #6542: more clearly list Android enum values docs in JavaDoc
Tail work from #6508 referenced in https://github.com/mapbox/mapbox-gl-native/pull/6508#issuecomment-250626817. 

We want better cross-linking between symbols in the JavaDoc. An example: 

```java
    //ICON_ROTATION_ALIGNMENT: In combination with `symbol-placement`, determines the rotation behavior of icons.

    /**
     * When `symbol-placement` is set to `point`, aligns icons east-west. When `symbol-placement` is set to `line`, aligns icon x-axes with the line.
     */
    public static final String ICON_ROTATION_ALIGNMENT_MAP = "map";
    /**
     * Produces icons whose x-axes are aligned with the x-axis of the viewport, regardless of the value of `symbol-placement`.
     */
    public static final String ICON_ROTATION_ALIGNMENT_VIEWPORT = "viewport";
    /**
     * When `symbol-placement` is set to `point`, this is equivalent to `viewport`. When `symbol-placement` is set to `line`, this is equivalent to `map`.
     */
    public static final String ICON_ROTATION_ALIGNMENT_AUTO = "auto";

    @StringDef({
            ICON_ROTATION_ALIGNMENT_MAP,
            ICON_ROTATION_ALIGNMENT_VIEWPORT,
            ICON_ROTATION_ALIGNMENT_AUTO,
        })
    @Retention(RetentionPolicy.SOURCE)
    public @interface ICON_ROTATION_ALIGNMENT {}
```

Here, `symbol-placement` should be `symbolPlacement` and should `@link` properly. Also, `point` should be `SYMBOL_PLACEMENT_POINT`, `line` should be `SYMBOL_PLACEMENT_LINE`, and `viewport` should self-identify and be `ICON_ROTATION_ALIGNMENT_VIEWPORT`. 

We already do this on Darwin (as of #6508) but considering Android was ready to go even without spec `enum` documentation, this can be release tail work and #6508 gets us an improvement on the books. 

/cc @ivovandongen @zugaldia @1ec5

ivovandongen:
Fixes #6510
Fixes #6542

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -447,9 +447,13 @@ test-node: node
 ANDROID_ENV = platform/android/scripts/toolchain.sh
 ANDROID_ABIS = arm-v5 arm-v7 arm-v8 x86 x86-64 mips
 
+.PHONY: style-code-android
+style-code-android: $(BUILD_DEPS)
+	node platform/android/scripts/generate-style-code.js
+
 define ANDROID_RULES
 
-build/android-$1/$(BUILDTYPE): $(BUILD_DEPS)
+build/android-$1/$(BUILDTYPE): style-code-android
 	mkdir -p build/android-$1/$(BUILDTYPE)
 
 build/android-$1/$(BUILDTYPE)/toolchain.cmake: platform/android/scripts/toolchain.sh build/android-$1/$(BUILDTYPE)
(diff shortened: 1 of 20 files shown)


ivovandongen:
@tobrun Think I got all the issues from #5610 and #6542. Maybe not perfect yet (could indeed use some examples here and there), but I think we should iterate on that. Let me know what you think.

PR #6572, opened 2016-10-04
[android] Revise handling of tracking modes and interactions with touch gestures.

tobrun:
Follow up on #6569, changed exposed API from original PR (closes #6549, closes #6567)

> This fixes issues #6549 and #6567. Also pertinent to #6557.
Additional code has been added to the test app (Activity "User Tracking
Mode") to test. A potential race condition in the MapboxMap#easeCamera() methods where
a camera change generated by tracking which could have reset the tracking
modes has been eliminated by factoring out a new method
MapboxMap#easeCameraInternal().

Review @ivovandongen

PR #6574, opened 2016-10-04
Rework setStyleUrl() logic to avoid NPE as per #6512 

### LINKED ISSUE #6512: NPE when MapView is recreated from savedInstanceSate

tobrun:
Follow up on #6556.
Closes #6512.

Review @ivovandongen

### DIFF
--- a/platform/android/MapboxGLAndroidSDK/src/main/java/com/mapbox/mapboxsdk/maps/MapView.java
+++ b/platform/android/MapboxGLAndroidSDK/src/main/java/com/mapbox/mapboxsdk/maps/MapView.java
@@ -594,10 +594,10 @@ public class MapView extends FrameLayout {
         nativeMapView.update();
         myLocationView.onResume();
 
-        // Presumably there is (or was) a good reason for delaying this until onResume.
-        // A comment explaining why would be helpful.
-        if(!styleWasSet)
-            setStyleUrl(styleUrl);    // will set styleWasSet
+        // In case that no style was set or was loaded through MapboxMapOptions
+        if (!styleWasSet) {
+            setStyleUrl(styleUrl);
+        }
     }
 
     void setFocalPoint(PointF focalPoint) {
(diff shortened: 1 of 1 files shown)


tobrun:
Sorry @clydebarrow, it seems that merging through github UI has lost your contribution on this PR. I command line merged your PR in https://github.com/mapbox/mapbox-gl-native/pull/6572 and this seems the work better.

PR #6581, opened 2016-10-04
iOS deploy script enhancements

friedbunny:
Spent some time converting @boundsj’s `deploy-packages.sh` into a more automated `make` command:

- **`make ideploy` is now the one-stop shop for deploying releases.**
 - Automatically detects latest `ios-vX.X.X-pre.X` tag.
 - Fails if version has already been published on GitHub.
 - Fails if tag does not exist, checks it out if it does.
 - Specify a version and skip GitHub publishing: `make ideploy VERSION_TAG=ios-v9.8.7 GITHUB_RELEASE=false`
- Does not allow your computer to go to sleep (`caffeinate`), which would ruin the deploy.
- Adds more informative steps, formatted like the build `make` commands.
- Shows upload progress for s3.
- Removes overly-verbose `github-release` output.
 - Tried to filter this instead, but `sed` would occasionally kill the entire process.
- Installs `github-release` from homebrew.
- Re-downloaded packages are now stored in `build/ios/deploy`.
 - Can override this via: `make ideploy BINARY_DIRECTORY='~/Desktop'`

Another future enhancement would be to check if you have `aws s3` push privileges.

/cc @boundsj @1ec5 @incanus @frederoni

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -254,6 +254,10 @@ ifabric: $(IOS_PROJ_PATH)
 	FORMAT=static BUILD_DEVICE=$(BUILD_DEVICE) SYMBOLS=NO SELF_CONTAINED=YES \
 	./platform/ios/scripts/package.sh
 
+.PHONY: ideploy
+ideploy:
+	caffeinate -i ./platform/ios/scripts/deploy-packages.sh
+
 .PHONY: idocument
 idocument:
 	OUTPUT=$(OUTPUT) ./platform/ios/scripts/document.sh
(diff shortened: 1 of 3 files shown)


friedbunny:
Thanks for the review, @boundsj. I’ve also updated the [release instructions wiki page](https://github.com/mapbox/mapbox-gl-native/wiki/Releasing-the-Mapbox-iOS-SDK).

Incidentally, the more I look at Fastlane, the more appealing it becomes for packaging tasks like this (even if it means another set of external dependencies). I got a bit caught up in enhancing our current deploy script, but the next phase is probably to ditch it for Fastlane. 😉

PR #6588, opened 2016-10-05
Revamp MGLStyleLayer, MGLSource inheritance

### LINKED ISSUE #6406: Put predicate property in protocol
A `predicate` property is declared redundantly on four out of six MGLBaseStyleLayer classes that conform to MGLStyleLayer. The property should instead be defined on an MGLSourceFiltering protocol that these four classes conform to.

/cc @frederoni @incanus

### LINKED ISSUE #6407: Predicate documentation should go into more detail
The documentation comment for four MGLStyleLayer subclass’s [`predicate`](https://github.com/mapbox/mapbox-gl-native/blob/7df9a21f415aa269e7ae435d1500f4a3c7d607e1/platform/darwin/src/MGLCircleStyleLayer.h#L27) property doesn’t go into too much detail, relying instead on a link to the style specification’s [`Filter`](https://www.mapbox.com/mapbox-gl-style-spec/#types-filter) documentation to provide more details. But the `Filter` documentation is expressed in a syntax completely foreign to iOS/macOS developers, who must create an NSPredicate instead of a nested array of strings. Though Apple’s NSPredicate documentation is thorough when it comes to expressable operators, there are a lot of expressable operators that implementations can choose not to implement, and indeed we haven’t implemented operators such as `LIKE`, `BEGINSWITH`, and blocks. Moreover, the style specification and Supercluster reserve the following keys:

* `$id` (#5708)
* `$type`
* `point_count` (mapbox/mapbox-gl-js#2985)

All these nuances would cause the `predicate` property documentation to grow unwieldy. To make matters worse, the property is declared redundantly on four individual MGLStyleLayer subclass. We should consolidate them into a single property: #6406. Beyond that, we could add our first appendix on the subject (realm/jazzy#435), giving predicates a little more discoverability.

/cc @friedbunny @frederoni

### LINKED ISSUE #6582: MGLBackgroundStyleLayer has a sourceIdentifier property
MGLBackgroundStyleLayer has `sourceIdentifier` and `sourceLayerIdentifier` properties, plus an `-initWithLayerIdentifier:source:` initializer, all of which are meaningless for a background style layer.

/ref #6269
/cc @boundsj

1ec5:
This PR cleans up the runtime styling API’s class hierarchy and completes the documentation thereof.

The MGLStyleLayer protocol has been removed, because almost none of its members was actually implemented in every class that adopted the protocol, and because all the concrete classes ultimately derived from the same base class. The unused `mapView` backpointer property has been removed with no replacement, although we’ll eventually have to bring it back – privately – for #6254. MGLBaseStyleLayer has been renamed to MGLStyleLayer. New intermediate abstract classes, MGLForegroundStyleLayer and MGLVectorStyleLayer, cover subsets of style layer classes with like functionality. To summarize, the style layer class hierarchy looks like this now (classes in _italics_ are abstract):

* _MGLStyleLayer_ (private properties in a `Private` category)
  * MGLBackgroundStyleLayer
  * _MGLForegroundStyleLayer_
    * MGLRasterStyleLayer
    * _MGLVectorStyleLayer_
      * MGLCircleStyleLayer
      * MGLFillStyleLayer
      * MGLLineStyleLayer
      * MGLSymbolStyleLayer

Each MGLBaseStyleLayer initializer and the corresponding properties have been pushed down to an abstract subclass such that the initializer makes sense for all concrete subclasses. Additional initializers and the predicate property have been pushed up to MGLVectorStyleLayer to eliminate duplication among the concrete subclasses. These initializers are all marked as designated initializers, and exceptions prevent developers from misusing them.

The word “source” or “layer” before “identifier” has been removed wherever the type of identifier is apparent. Extra MGLGeoJSONSource initializer variants have been removed in favor of nullable parameters (overlapping a little with #6524 for #6302).

Copious documentation comments have been added for source and style layer classes, including several previously undocumented methods and properties. In particular, the documentation now includes important preconditions and postconditions, as well as details about the subset of the predicate API that we support. Pragma marks break up the jazzy documentation pages into sections. Unfortunately, the fancy, marked-up documentation comments show up as plain text in Xcode 8: [rdar://problem/28628571](http://openradar.appspot.com/radar?id=5758406015057920). At least they look good in jazzy:

predicate

Fixes #6406, fixes #6407, fixes #6582. Working towards #5970.

/cc @frederoni @incanus @boundsj


platform/darwin/src/MGLForegroundStyleLayer.m
```diff
@@ -0,0 +1,13 @@
+#import "MGLForegroundStyleLayer.h"
+#import "MGLSource.h"
+
+@implementation MGLForegroundStyleLayer
+
+- (instancetype)initWithIdentifier:(NSString *)identifier source:(MGLSource *)source {
+    if (self = [super initWithIdentifier:identifier]) {
```
boundsj: should this initializer raise an exception?

platform/darwin/src/MGLForegroundStyleLayer.h
```diff
@@ -0,0 +1,45 @@
+#import <Foundation/Foundation.h>
+
+#import "MGLStyleLayer.h"
+
+NS_ASSUME_NONNULL_BEGIN
+
+@class MGLSource;
+
+/**
+ `MGLSourcedStyleLayer` is an abstract superclass for style layers whose content
```
boundsj: `MGLForegroundStyleLayer` is ...

1ec5: Fixed.

platform/darwin/src/MGLVectorStyleLayer.h
```diff
@@ -0,0 +1,127 @@
+#import <Foundation/Foundation.h>
+
+#import "MGLForegroundStyleLayer.h"
+
+NS_ASSUME_NONNULL_BEGIN
+
+/**
+ `MGLSourcedStyleLayer` is an abstract superclass for style layers whose content
```
1ec5: Fixed.

platform/darwin/src/MGLForegroundStyleLayer.m
boundsj: Thanks. That makes perfect sense for the abstract classes.

boundsj:
Would it make sense to override `initWithIdentifier:source:` in all of the concrete subclasses of `MGLForegroundStyleLayer` and raise an exception so that something like `[[MGLFillStyleLayer alloc] initWithIdentifier:@"some-id"];` fails faster? I'm thinking about the edge case where the developer initializes the concrete layer subclass "correctly" but creates a ⏲ 💣 if the instance is stored and used later on. Alternatively, I suppose we could document this edge case in the `MGLStyleLayer` header but it seems like it might be confusing since `initWithIdentifier:source:` is perfectly fine for `MGLBackgroundStyleLayer`.

1ec5:
> Would it make sense to override `initWithIdentifier:source:` in all of the concrete subclasses of `MGLForegroundStyleLayer` and raise an exception so that something like `[[MGLFillStyleLayer alloc] initWithIdentifier:@"some-id"];` fails faster? I'm thinking about the edge case where the developer initializes the concrete layer subclass "correctly" but creates a ⏲ 💣 if the instance is stored and used later on.

Since `-initWithIdentifier:source:` is the designated initializer for MGLForegroundStyleLayer, calling `-[MGLFillStyleLayer initWithIdentifier:source:]` results in a compile-time error or warning.

Unfortunately, we can’t do anything about the developer calling `-[MGLForegroundStyleLayer initWithIdentifier:source:]`, other than the ⏲💣 behavior noted above. This is an intrinsic limitation to abstract superclasses in Objective-C and the only real downside to using abstract superclasses instead of protocols. Because `-initWithIdentifier:source:` is the designated initializer, the concrete subclasses’ implementations of that method must call the abstract superclass’s implementation. The same pitfall exists for developers who attempt to initialize an NSControl or UIControl using `-initWithFrame:`. The only difference is that we don’t currently support custom subclasses of MGLStyleLayer. (If in the future we do support something along those lines, we’d document the methods that superclasses would want to override.)

> Alternatively, I suppose we could document this edge case in the `MGLStyleLayer` header but it seems like it might be confusing since `initWithIdentifier:source:` is perfectly fine for `MGLBackgroundStyleLayer`.

The comments for all the abstract superclasses already say that the developer must initialize a concrete subclass, not the abstract superclass. I reiterate the point in the exception messages.

1ec5:
> Since `-initWithIdentifier:source:` is the designated initializer for MGLForegroundStyleLayer, calling `-[MGLFillStyleLayer initWithIdentifier:]` results in a compile-time error or warning.

I was mistaken: `NS_DESIGNATED_INITIALIZER` only forces the subclass’s implementation to call the superclass’s, but the non-designated initializer becomes a convenience initializer rather than becoming unavailable. In 0bb46eaf0cac468ccbb0d42f7855640916d862c2, I explicitly marked the incomplete initializers unavailable to prevent the very likely case of a developer calling `-[MGLFillStyleLayer init]`.

PR #6589, opened 2016-10-05
Simplify style property enumeration names 

### LINKED ISSUE #6540: cross-reference Darwin enum docs to properties that use them
Tail work from #6508 referenced in https://github.com/mapbox/mapbox-gl-native/pull/6508#pullrequestreview-2230223. 

For example, the docs for `MGLCircleStyleLayerCircleTranslateAnchor` could reference `MGLCircleStyleLayer.circleTranslateAnchor` for clarity. 

/cc @1ec5

### LINKED ISSUE #6577: Style attribute enumerations redundantly include layer class names
The style layer classes’ paint attributes are named somewhat redundantly, for instance `-[MGLCircleStyleLayer circleOpacity]`. Since there’s no possibility that, say, `lineOpacity` would ever exist on the same class, we should remove the first word from this property and any property like it. It’s totally fine if, say, `width` means something slightly different on circle and line style layers, since the meaning on each is consistent with developer expectations.

Putting on the v3.4.0 milestone because we won’t be able to change this later without breaking backwards compatibility.

/cc @frederoni @incanus

1ec5:
The name and values of each enumeration type no longer refers to the layer type that uses the type. Instead, the specific class and property that uses the type is mentioned in the documentation comment, to be cross-linked in the jazzy-generated documentation.

Fixes #6540, fixes #6577.

/cc @incanus


1ec5:
One consequence of this change is that, going forward, the style specification shouldn’t allow any two layer types to have identically named properties.

1ec5:
>> One consequence of this change is that, going forward, the style specification shouldn’t allow any two layer types to have identically named properties.

> Shouldn't be a problem though, given that properties still have a prefix of the layer type (e.g. MGLCircle...?

Theoretically, the style specification could be extended so that the `circle` layer type has a `text-field` property just like the `symbol` layer types does today. That would break this assumption, because Objective-C has no namespacing.

incanus:
> Theoretically, the style specification could be extended so that the circle layer type has a icon-rotation property just like the symbol layer type does today.

Wait, why wouldn't we have `MGLCircleIconRotation` and `MGLSymbolIconRotation`? 

This PR has both layer-less properties like `MGLIconRotationAlignment` as well as layered ones like `MGLSymbolPlacement`. Why is `MGLIconRotationAlignment` not `MGLSymbolIconRotationAlignment`?

incanus:
Actually, scratch that—I get it, those properties just happen to refer to the layer type in their name. Never mind. 

/cc @jfirebaugh on this for some 👀. It's less invasive than what was initially proposed in https://github.com/mapbox/mapbox-gl-native/issues/6577 (and which I agree with @jfirebaugh on about accepting some redundancy).

1ec5:
> This PR has both layer-less properties like `MGLIconRotationAlignment` as well as layered ones like `MGLSymbolPlacement`. Why is `MGLIconRotationAlignment` not `MGLSymbolIconRotationAlignment`? 

Because the properties are called [`icon-rotation-alignment`](https://www.mapbox.com/mapbox-gl-style-spec/#layout-symbol-icon-rotation-alignment) and [`symbol-placement`](https://www.mapbox.com/mapbox-gl-style-spec/#layout-symbol-symbol-placement) in the style specification. In #6577, I had originally proposed that we drop redundant words from the style properties wherever possible, so that we’d end up with `MGLSymbolIconRotationAlignment` and `MGLSymbolPlacement`, but that idea didn’t fly.

PR #6598, opened 2016-10-05
Use Pixel Buffer Object for async glReadPixels

### LINKED ISSUE #477: Async glReadPixels() in HeadlessView?
Does this make sense or seem worthwhile @kkaefer?

https://www.opengl.org/discussion_boards/showthread.php/144238-asynchronous-readback-that-call-glReadPixels%28%29-with-PBO?highlight=multithread%2A

mikemorris:
Fixes #477 

I'm not sure if the `GLsync` fence use here is a good idea or not, would appreciate a review and needs more testing downstream before we could consider merging this. Need to add logging for why `glClientWaitSync` unblocked too (should warn on `GL_TIMEOUT_EXPIRED` or `GL_WAIT_FAILED`).

/cc @kkaefer @jfirebaugh

### DIFF
--- a/platform/default/headless_view.cpp
+++ b/platform/default/headless_view.cpp
@@ -1,6 +1,7 @@
 #include <mbgl/platform/default/headless_view.hpp>
 #include <mbgl/platform/default/headless_display.hpp>
 #include <mbgl/gl/types.hpp>
+#include <mbgl/platform/log.hpp>
 
 #include <cassert>
 #include <cstring>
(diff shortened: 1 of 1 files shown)


jfirebaugh:
I'm not clear on why this would be any better than `glReadPixels` without a PBO -- it's just doing manually what the API would do internally, right?

The point of an async `glReadPixels` is that you can  _do other work_ while the GPU is doing its thing, and this doesn't seem to accomplish that.

mikemorris:
@jfirebaugh The intention behind this work was to avoid GL pipeline stalls, I'm not too familiar on the technical details here, but there's some background discussion in https://github.com/mapbox/mapbox-gl-native/issues/477#issuecomment-242856572

Issue #4860, opened 2016-04-27
Port data-driven styling (property functions) from GL JS


ansis:
Port data-driven styling from GL JS (see mapbox/mapbox-gl-js#1626). DDS is currently exposed as [property functions](https://www.mapbox.com/mapbox-gl-style-spec/#types-function) in the style specification.

Master list of what's needed. Not necessarily in implementation order.

* [x] Refactor existing `LayoutProperty` and `PaintProperty` classes such that unevaluated and evaluated properties live in separate structures (in progress, @jfirebaugh).
* [x] `gl::Drawable`/`gl::Context` need support for multiple vertex buffers.
* [x] Rename `Function` ⇢ `ZoomFunction`.
* [x] Add `PropertyFunction` and `ZoomPropertyFunction`. Add evaluators for them.
* [x] Add `PropertyFunction` and `ZoomPropertyFunction` as variant types for `PropertyValue`.
  * Might need to split `PropertyValue` into one variant for properties that are known to be feature-constant (undefined, constant, or zoom function) and one for properties that are data-driven (property function or zoom property function).
* [x] Paint properties can no longer be fully evaluated by `Style`. Instead, pass cascaded paint properties into the bucket. Bucket figures out which are feature constant and which are data-driven, producing a `ProgramConfiguration` (or `PaintConfiguration`?). In the `ProgramConfiguration`:
  * constant properties have a `ZoomFunction` which will be evaluated by the main thread and the result bound to a uniform
  * data-driven properties have a vertex buffer
* [x] Decision: for feature-constant properties, should we use `glDisableVertexAttribArray`/`glVertexAttrib*` or dynamically-generated shader uniforms?
* [x] Decision: single "interleaved" vertex buffer for all data-driven paint properties of a layer (requires implementing a custom vector type with runtime-sized elements), or one vertex buffer per data-driven property (can keep using `std::vector`).
* [x] Changing a paint property needs to do a worker round trip (if changing to or from data-driven property or updating a data-driven property), or update the `PaintConfiguration` for each bucket in place (if not).
* [x] Implement support in SDK bindings.

nitrag:
Awesome!

So this will this allow MGLShapeSource parsing (mapbox built-in) to dynamically style a MGLPolyline based on `feature.properties.line-color` from the geojson source?

https://www.mapbox.com/mapbox-gl-js/style-spec/#paint-line-color

boundsj:
[...]

PR #6601, opened 2016-10-05
Revamp MGLStyleLayer property types

### LINKED ISSUE #5970: Add category convenience methods to NSNumber for style values (was: Runtime styling properties are too loosely typed)

### LINKED ISSUE #6078: Runtime styling category methods are prefixed with mbgl_
Category methods added as part of the runtime styling API implementation are prefixed with `mbgl_`. They should be prefixed with `mgl_` instead, for consistency with our class prefix.

/cc @frederoni

1ec5:
This PR reworks how style attribute values are represented in the runtime styling API. The intricate system of protocols adopted by Foundation classes via categories has been replaced with a simple umbrella value type, MGLStyleValue, that adopts lightweight generics. The parallel – and often incomplete – private protocols adopted by Foundation classes via categories has been replaced by a single templated C++ class, MGLStyleValueTransformer. C++ templates were chosen over C macros because they’re far easier to debug (once they compile, that is).

Everywhere the developer previously used a string literal, for instance, they must now wrap that string literal in an `MGLStyleValue<NSString *>`. That includes style functions: MGLStyleFunction, one of the classes under the MGLStyleValue umbrella, expects stop values to be wrapped in MGLStyleValues. See https://github.com/mapbox/mapbox-gl-native/pull/6601#issuecomment-252302160 for a concrete example in both Swift and Objective-C. All this is necessary because the Objective-C language lacks variant support.

The goal is to add a veneer of type safety, even when working with style functions, yet avoid a proliferation of abstract classes or private protocols. This approach should also extend rather straightforwardly to property functions and future extensions to the style specification syntax. More importantly, this PR ensures _internal_ type safety, avoiding what are in essence unchecked casts throughout the API implementation.

Fixes #5970. #6588 cleans up the style layer classes themselves, whereas this PR cleans up each style layer class’s properties.

Fixes #6078.

Lots to do:

* [x] Implement value classes
* [x] Prototype conversions between Objective-C and C++ types
* [x] Get the changes to link
* [x] Apply the changes to all the style attributes
* [x] Get array getters/setters to compile
* [x] Get tests to pass
* [x] Remove unused `-[MGLStyleFunction rawValueAtZoomLevel:]`
* [x] Update property documentation
* [x] Document MGLStyleValue classes

/cc @frederoni @boundsj @incanus


1ec5:
Here are the current linker errors:

```
Apple Mach-O Linker (ld) Error Group
  "mbgl::style::PropertyValue<mbgl::style::LineJoinType> MBGLValueFromMGLStyleValue<NSValue* __strong, mbgl::style::LineJoinType>(MGLStyleValue<NSValue* __strong>*)", referenced from:

  "mbgl::style::PropertyValue<float> MBGLValueFromMGLStyleValue<NSNumber* __strong, float>(MGLStyleValue<NSNumber* __strong>*)", referenced from:

  "MGLStyleValue<NSValue* __strong>* MGLStyleValueFromMBGLValue<mbgl::style::LineJoinType, NSValue* __strong>(mbgl::style::PropertyValue<mbgl::style::LineJoinType>&)", referenced from:

  "MGLStyleValue<NSNumber* __strong>* MGLStyleValueFromMBGLValue<float, NSNumber* __strong>(mbgl::style::PropertyValue<float>&)", referenced from:

```

1ec5:
The mangled version:

```
  "__Z26MBGLValueFromMGLStyleValueIU8__strongP7NSValueN4mbgl5style12LineJoinTypeEENS4_13PropertyValueIT0_EEP13MGLStyleValueIT_E", referenced from:
      -[MGLLineStyleLayer setLineJoin:] in MGLLineStyleLayer.o
  "__Z26MBGLValueFromMGLStyleValueIU8__strongP8NSNumberfEN4mbgl5style13PropertyValueIT0_EEP13MGLStyleValueIT_E", referenced from:
      -[MGLLineStyleLayer setLineMiterLimit:] in MGLLineStyleLayer.o
  "__Z26MGLStyleValueFromMBGLValueIN4mbgl5style12LineJoinTypeEU8__strongP7NSValueEP13MGLStyleValueIT0_ERNS1_13PropertyValueIT_EE", referenced from:
      -[MGLLineStyleLayer lineJoin] in MGLLineStyleLayer.o
  "__Z26MGLStyleValueFromMBGLValueIfU8__strongP8NSNumberEP13MGLStyleValueIT0_ERN4mbgl5style13PropertyValueIT_EE", referenced from:
      -[MGLLineStyleLayer lineMiterLimit] in MGLLineStyleLayer.o
```

Indeed, there’s no trace of `MBGLValueFromMGLStyleValue` at all in MGLStyleAttributeFunction.o.

1ec5:
As I work through the linker issues, a design question for everyone: am I swinging the pendulum too far towards type safety at the expense of concision? Something like this:

```swift
let layer = mapView.style().layer(identifier: "motorway") as? MGLLineStyleLayer
layer?.lineColor = UIColor.blue
let fn = MGLStyleAttributeFunction()
fn.stops = [1: 5, 18: 3]
layer?.lineWidth = fn
```

will probably turn into this:

```swift
let layer = mapView.style().layer(identifier: "motorway") as? MGLLineStyleLayer
layer?.lineColor = MGLStyleValue<UIColor>(rawValue: .blue)
let stops = [
    1: MGLStyleValue<Float>(rawValue: 5),
    18: MGLStyleValue<Float>(rawValue: 3),
]
layer?.lineWidth = MGLStyleValue<Float>(stops: stops)
```

(Swift may be able to infer the `<UIColor>` and `<Float>`s in this case, but Objective-C won’t.)

This recursive use of MGLStyleValue is pretty ugly, although it’s flexible enough to accommodate property functions (#4860) and future extensions to the style specification syntax, such as token defaults (mapbox/mapbox-gl-style-spec#104) and conditionals (mapbox/mapbox-gl-style-spec#402).

incanus:
The new Swift implementation above is _slightly_ more awkward. Things like this, mentioning `raw`, rub me a little funny: 

```swift
layer?.lineColor = MGLStyleValue(rawValue: .blue)
```

The new Objective-C implementation feels _a lot_ more awkward. 

```objc
layer.lineColor = [MGLStyleValue<UIColor> valueWithRawValue:UIColor.blueColor];
```

This is merely setting a property (a short one at that) to a color, but it's verbose even for Objective-C. 

Something I've been working with lately is SceneKit, which features a number of types that can be passed as `contents` on `SCNMaterialProperty` objects. 

screen shot 2016-10-07 at 11 52 16 am

Notably: 

> # Discussion

> For details on each visual property and the ways their contents affect a material’s appearance, see SCNMaterial.

> You can set a value for this property using any of the following types:

> - A color (`NSColor`/`UIColor` or `CGColorRef`), specifying a constant color across the material’s surface

> - An image (`NSImage`/`UIImage` or `CGImageRef`), specifying a texture to be mapped across the material’s surface

> - An `NSString` or `NSURL` object specifying the location of an image file

> - A Core Animation layer (`CALayer`)

> - A texture (`SKTexture`, `MDLTexture`, `MTLTexture`, or `GLKTextureInfo`)

> A SpriteKit scene (`SKScene`)

> - A specially formatted image or array of six images, specifying the faces of a cube map

This is even broader than our property APIs, yet still relies on an untyped `id` for the `contents` property. 

These are presumably meant to be used dynamically at runtime, else `SCNMaterialProperty` would only feature initializers that take `contents` at instantiation and then make them immutable in future. 

If you use a wrong type, _it's just silently ignored_. I don't feel like this is user-hostile and really keeps things simple, no? 

```swift
override func viewDidLoad() {
    super.viewDidLoad()

    let ball = SCNSphere(radius: 0.5)
    let node = SCNNode(geometry: ball)
    let sceneView = SCNView(frame: view.bounds)
    sceneView.scene = SCNScene()
    sceneView.scene?.rootNode.addChildNode(node)
    view.addSubview(sceneView)

    ball.firstMaterial?.diffuse.contents = UIColor.red // good
    ball.firstMaterial?.specular.contents = UIView(frame: view.bounds) // bad, ignored
}
```

Are we overthinking this API?

1ec5:
Thank you for this example @incanus. It certainly speaks to a concern I’ve had as we’ve been building out this API. There’s a lot to be said for the simplicity of just using `id` or even putting all the attributes into an untyped NSDictionary, as do many Cocoa APIs like NSAttributedString. That said, SceneKit’s API was originally released in 2012, long before Swift launched with [associated types](https://developer.apple.com/library/content/documentation/Swift/Conceptual/Swift_Programming_Language/Enumerations.html#//apple_ref/doc/uid/TP40014097-CH12-ID148) (variants) and Objective-C gained support for lightweight generics. I wonder if the API would’ve been this fast and loose with types if Swift had been out by then, given that it runs counter to Swift’s preference for a strict type system.

One alternative I proposed awhile back, in https://github.com/mapbox/mapbox-gl-native/issues/5970#issuecomment-239372568, was to rely just as heavily on protocols as we do now, but make the protocols more specific. If we go that route, I think we need to avoid the mess of private categories that we currently rely on: it’s too error-prone on our end and so easy to misuse on the developer’s.

1ec5:
> Things like this, mentioning `raw`, rub me a little funny

I’m not wedded to the “raw value” terminology, although it’s consistent with raw values of Swift enumerations. For this PR, I was considering some helper classes like MGLStyleNumber, the same way NSNumber is a convenience subclass for NSValue.

incanus:
FWIW `SCNMaterialProperty` is from iOS 8+, meaning 2014. So, a bit newer and probably simultaneously / caught by surprise with Swift. Still, a data point 😄 

I'll dig around a bit to see if Apple is introducing any new APIs like this, just as more data. 

Still thinking on this a bit.

frederoni:
Great rework @1ec5 
It's way more robust and future proof than before but indeed, not very concise.
Why not add a few convenient instance methods on UIColor, NSNumber etc in order to turn this
```objc
layer.lineGapWidth = [MGLStyleValue<NSNumber *> valueWithRawValue:@5];
layer.lineColor = [MGLStyleValue<UIColor> valueWithRawValue:UIColor.blueColor];
MGLStyleAttributeFunction *fn = [[MGLStyleAttributeFunction alloc] init];
NSDictionary *stops = @{
    @1: [MGLStyleValue<NSNumber *> valueWithRawValue:@5],
    @18: [MGLStyleValue<NSNumber *> valueWithRawValue:@3],
};
```
into this:
```objc
layer.lineGapWidth = @5.mglValue;
layer.lineColor = UIColor.blueColor.mglColor;
MGLStyleAttributeFunction *fn = [[MGLStyleAttributeFunction alloc] init];
NSDictionary *stops = @{
    @1: @5.mglValue,
    @18: @3.mglValue,
};
```

1ec5:
> Why not add a few convenient instance methods on UIColor, NSNumber

That’s a clever idea. `-mgl_styleConstant` wouldn’t be especially discoverable, but that’s OK because it’s just a shortcut for the more verbose syntax that would be discoverable. On the other hand, such a method has the potential to muddle the developer’s idea of what exactly is being passed in. It’s worth exploring, certainly.

1ec5:
With @boundsj’s help, I was able to get past the linker issues over the weekend. I also fixed a number of silly mistakes related to switching a method to use an inout parameter instead of a return value.

This has been the sole remaining beta blocker for awhile now. The criteria for a beta blocker is that it prevents us from being able to encourage developers to start developing against a new feature, because we expect to break their code significantly in a future prerelease. Given the problems we’ve already identified with the current API, like https://github.com/mapbox/mapbox-gl-native/issues/5970#issuecomment-240808179, I believe #5970 still deserves to be a beta blocker – we need to change _something_.

How about we land these changes, verbosity and all, and release the beta? Feedback we get from the beta can inform whether we:

1. Introduce the Foundation category conveniences proposed in https://github.com/mapbox/mapbox-gl-native/pull/6601#issuecomment-252479972;
2. Dial down to unqualified `id`s as proposed in https://github.com/mapbox/mapbox-gl-native/pull/6601#issuecomment-252335096; or
3. Introduce type-specific `id`s like `MGLBoolRepresentable` as proposed in https://github.com/mapbox/mapbox-gl-native/issues/5970#issuecomment-239372568.

We can layer all three approaches atop the changes in this PR. After all, the bulk of this PR is about improving _internal_ type safety, which we want regardless of external ergonomics.

1ec5:
Thanks everyone for your feedback. I’m going to finish the documentation and land this PR for the beta. Then we’ll continue the discussion in #5970.

1ec5:
Filed #6712 about the Qt5 build bot failure.

boundsj:
🚢 🚀

PR #6608, opened 2016-10-06
JumpTo style camera position

### LINKED ISSUE #6129: Jump to style’s default viewport on initialization
The map should jump to the default center coordinates, zoom level, bearing, and pitch as soon as the first style loads after initialization. Switching from one style to another probably shouldn’t change the viewport.

Complicating matters, the initial viewport may be overridden for various reasons. For example, state restoration would take place upon initialization. Or the developer may programmatically set the viewport immediately after initializing the map view but before the style loads. The style shouldn’t be able to clobber a viewport set this way.

In the iOS and macOS SDKs, MGLMapView’s initial center coordinates, zoom level, and direction can be specified in a storyboard. Unfortunately, storyboard-defined properties are set [asynchronously, in piecemeal fashion](https://github.com/mapbox/mapbox-gl-native/blob/63cce78cf1597ba5879227f2298bda52e91e4e04/platform/ios/src/MGLMapView.mm#L4983-L4988), after the MGLMapView object is initialized and thus sometime after `mbgl::Map` is initialized. The style’s default viewport shouldn’t clobber this asynchronously set viewport, either.

In https://github.com/mapbox/mapbox-gl-native/pull/6127#issuecomment-241899812, @jfirebaugh suggests implementing the initial viewport behavior in core. That could work for the iOS and macOS SDKs, as long as we account for the cases above. Perhaps methods like `mbgl::Map::setLatLng()` and `easeTo()` could shut off any style-default-viewport functionality for the lifetime of the map.

/cc @friedbunny @tobrun

tobrun:
WIP, closes #6129 

This PR enables loading the default camera position from a style using the following items from the [style-spec](https://www.mapbox.com/mapbox-gl-style-spec/#root-center). If the user changes the camera position while a style is being loaded, the camera position from that style will be ignored.

Currently covers Android use-cases since we only expose `jumpTo`, `easeTo` and `flyTo` as public API that could conflict with the camera position from the style. Will need to look into other bindings since they will probably expose other methods as Map::setLatLng or Map::setZoom as public API.

Would love some feedback on current approach & need to rename `setIgnoreInitialCameraOptions`.

**Question**: Would an approach hooking into MapChange events be a cleaner solution? 

cc @jfirebaugh @1ec5 @incanus

### DIFF
--- a/src/mbgl/map/map.cpp
+++ b/src/mbgl/map/map.cpp
@@ -43,6 +43,8 @@ public:
     void onStyleLoaded() override;
     void onStyleError() override;
     void onResourceError(std::exception_ptr) override;
+    
+    void setIgnoreInitialCameraOptions();
 
     void update();
     void render();
@@ -54,6 +56,7 @@ public:
 
     RenderState renderState = RenderState::Never;
     Transform transform;
+    std::unique_ptr<CameraOptions> initialCameraOptions;
 
     const MapMode mode;
     const GLContextMode contextMode;
@@ -104,6 +107,7 @@ Map::Impl::Impl(View& view_,
       asyncUpdate([this] { update(); }),
       workerThreadPool(4),
       annotationManager(std::make_unique<AnnotationManager>(pixelRatio)) {
+      initialCameraOptions = std::make_unique<CameraOptions>();
 }
 
 Map::~Map() {
@@ -400,16 +404,19 @@ CameraOptions Map::getCameraOptions(optional<EdgeInsets> padding) const {
 }
 
 void Map::jumpTo(const CameraOptions& camera) {
+    impl->setIgnoreInitialCameraOptions();
     impl->transform.jumpTo(camera);
     update(camera.zoom ? Update::RecalculateStyle : Update::Repaint);
 }
 
 void Map::easeTo(const CameraOptions& camera, const AnimationOptions& animation) {
+    impl->setIgnoreInitialCameraOptions();
     impl->transform.easeTo(camera, animation);
     update(camera.zoom ? Update::RecalculateStyle : Update::Repaint);
 }
 
 void Map::flyTo(const CameraOptions& camera, const AnimationOptions& animation) {
+    impl->setIgnoreInitialCameraOptions();
     impl->transform.flyTo(camera, animation);
     update(Update::RecalculateStyle);
 }
@@ -997,6 +1004,22 @@ void Map::Impl::onUpdate(Update flags) {
     
 void Map::Impl::onStyleLoaded() {
     view.notifyMapChange(MapChangeDidFinishLoadingStyle);
+    
+    if (initialCameraOptions == nullptr) {
+        // user manipulated camera before style was loaded
+        return;
+    }
+    
+    // move camera to default camera of style
+    initialCameraOptions->center = style->getDefaultLatLng();
+    initialCameraOptions->zoom = style->getDefaultZoom();
+    initialCameraOptions->angle = style->getDefaultBearing();
+    initialCameraOptions->pitch = style->getDefaultPitch();
+    transform.jumpTo(*initialCameraOptions);
+}
+
+void Map::Impl::setIgnoreInitialCameraOptions() {
+    initialCameraOptions = nullptr;
 }
 
 void Map::Impl::onStyleError() {
--- a/platform/android/MapboxGLAndroidSDK/src/main/java/com/mapbox/mapboxsdk/camera/CameraPosition.java
+++ b/platform/android/MapboxGLAndroidSDK/src/main/java/com/mapbox/mapboxsdk/camera/CameraPosition.java
@@ -14,6 +14,8 @@ import com.mapbox.mapboxsdk.utils.MathUtils;
  */
 public final class CameraPosition implements Parcelable {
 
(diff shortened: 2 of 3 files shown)


src/mbgl/map/map.cpp
```diff
@@ -54,6 +56,7 @@ class Map::Impl : public style::Observer {
 
     RenderState renderState = RenderState::Never;
     Transform transform;
+    std::unique_ptr<CameraOptions> initialCameraOptions;
```
1ec5: `CameraOptions` is just a struct, so there’s no need to have a pointer to it. It looks like you’re using `std::unique_ptr` as a way of allowing the options to be null. In that case, use `optional` instead.

src/mbgl/map/map.cpp
```diff
@@ -43,6 +43,8 @@ class Map::Impl : public style::Observer {
     void onStyleLoaded() override;
     void onStyleError() override;
     void onResourceError(std::exception_ptr) override;
+    
+    void setIgnoreInitialCameraOptions();
```
1ec5: Since the SDKs should have no need to call this method, make it private.

src/mbgl/map/map.cpp
```diff
@@ -997,6 +1004,22 @@ void Map::Impl::onUpdate(Update flags) {
...
+    
+    // move camera to default camera of style
+    initialCameraOptions->center = style->getDefaultLatLng();
+    initialCameraOptions->zoom = style->getDefaultZoom();
+    initialCameraOptions->angle = style->getDefaultBearing();
+    initialCameraOptions->pitch = style->getDefaultPitch();
+    transform.jumpTo(*initialCameraOptions);
+}
+
+void Map::Impl::setIgnoreInitialCameraOptions() {
```
1ec5: Inline this method. Seeing `initialCameraOptions` get cleared out is pretty self-explanatory; no need for a separate method.

src/mbgl/map/map.cpp
```diff
@@ -997,6 +1004,22 @@ void Map::Impl::onUpdate(Update flags) {
...
+    
+    if (initialCameraOptions == nullptr) {
+        // user manipulated camera before style was loaded
+        return;
+    }
+    
+    // move camera to default camera of style
+    initialCameraOptions->center = style->getDefaultLatLng();
+    initialCameraOptions->zoom = style->getDefaultZoom();
+    initialCameraOptions->angle = style->getDefaultBearing();
```
1ec5: In the style specification, the [`bearing`](https://www.mapbox.com/mapbox-gl-style-spec/#root-bearing) property is measured in degrees clockwise from north. However, `CameraOptions::angle` is measured in radians counterclockwise from north. We need to convert either here or in `mbgl::Style::setJSON()`, where we set the default bearing based on the value parsed out of the style JSON. We should also document the units and direction of `getDefaultBearing()`.

src/mbgl/map/map.cpp
```diff
@@ -997,6 +1004,22 @@ void Map::Impl::onUpdate(Update flags) {
...
+    initialCameraOptions->pitch = style->getDefaultPitch();
```
1ec5: Similarly, in the style specification, the [`pitch`](https://www.mapbox.com/mapbox-gl-style-spec/#root-pitch) property is measured in degrees increasing front and up, whereas `CameraOptions::pitch` is measured in radians increasing front and up.

src/mbgl/map/map.cpp
```diff
@@ -400,16 +404,19 @@ CameraOptions Map::getCameraOptions(optional<EdgeInsets> padding) const {
 }
 
 void Map::jumpTo(const CameraOptions& camera) {
```
1ec5: What about methods such as `moveBy()`, `setLatLng()`, `scaleBy()`, `setLatLngZoom()`, `rotateBy()`, etc.?

src/mbgl/map/map.cpp
1ec5: `getDefaultBearing()` should probably stay in degrees clockwise from north. Conventionally, public-facing APIs like `mbgl::Map` use degrees, while internal classes like `TransformState` and `Transform` use radians. However, `CameraOptions` is passed around up and down the stack, and it’s in radians. It would be a very large project to make all of mbgl use radians, so for this PR, I think you should convert to radians right here and perhaps document `getDefaultBearing()`’s current behavior.

src/mbgl/map/map.cpp
tobrun: that is correct, I changed it with `optional` and using `std::experimental::nullopt` instead of `nullptr` as a reset value.

src/mbgl/map/map.cpp
tobrun: Yes, as mentioned in OP: `Will need to look into other bindings since they will probably expose other methods as Map::setLatLng or Map::setZoom as public API.`

jfirebaugh:
@tobrun, thanks for taking a shot at this. I ended up needing #6129 implemented for another issue, and took a different tack on it in #6989. So the mbgl core changes here aren't needed. I'm not sure if the Android changes are or not -- can you take a look?

tobrun:
@jfirebaugh thank you for picking this up, didn't have the time to finish this.
Closing this PR and will gladly review #6989. Thanks again!

PR #6579, opened 2016-10-04
[core] Tiles that error on load are not renderable

jfirebaugh:
In 41bbd4e4f7d66465433e370ca024ab0239fcace3 I merged two error pathways that need to be distinct:

* If an error occurs while loading a tile, the tile should not be treated as renderable
* If an error occurs while creating the tile layout, the tile can be considered renderable (there might be partial data, or the data from a prior layout may still be used)

Fixes #6506
Fixes #6547

@halset, thanks for locating the offending commit, it made the cause of this regression easy to find.

PR #6613, opened 2016-10-06
Explicitly initialize default runloop on Darwin

kkaefer:
https://github.com/mapbox/mapbox-gl-native/pull/6579 introduced a unit test crash on macOS that exposed a larger issue: In the Darwin `RunLoop` implementation, we are statically initializing a `RunLoop` object for the main thread. However, this object is overwritten, and subsequently destructed/nulled by most unit tests that use their own `RunLoop`. This moves the static initialization to the macOS and iOS SDK. The GLFW implementation which also runs on macOS already creates its own `RunLoop`.

/cc @jfirebaugh

### DIFF
--- a/platform/darwin/src/run_loop.cpp
+++ b/platform/darwin/src/run_loop.cpp
@@ -8,7 +8,6 @@ namespace mbgl {
 namespace util {
 
 static ThreadLocal<RunLoop>& current = *new ThreadLocal<RunLoop>;
-static RunLoop mainRunLoop;
 
 class RunLoop::Impl {
 public:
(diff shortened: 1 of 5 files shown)


platform/darwin/src/run_loop.cpp
```diff
@@ -22,6 +21,7 @@ RunLoop* RunLoop::Get() {
 
 RunLoop::RunLoop(Type)
   : impl(std::make_unique<Impl>()) {
+    assert(!current.get());
```
kkaefer: Allows us to quickly catch duplicate `RunLoop`s in the future.

test/tile/raster_tile.test.cpp
```diff
@@ -15,6 +16,7 @@ class RasterTileTest {
 public:
     FakeFileSource fileSource;
     TransformState transformState;
+    util::RunLoop loop;
```
kkaefer: Tests were missing an explicit `RunLoop`. If running these tests separately, this is not an issue because we had the default `RunLoop`, but when running after other unit tests that create/destroy a `RunLoop`, we got a null `RunLoop` further down the line.

PR #6614, opened 2016-10-06
add NodeMap reference to req paramter passed to request function

mikemorris:
To give resource requests visibility into the `NodeMap` object that triggered them.


jfirebaugh:
⚠️  Given the history with memory leaks related to the map/request relationship, I think we should be _extremely_ diligent and cautious with this change. ⚠️

mikemorris:
Think we can get by without this.

PR #6618, opened 2016-10-06
Introduce a setConnected method to manually set a connected flag

### LINKED ISSUE #6617: Add method to set connectivity status manually
We've heard from a few SDK users that the recently added [ConnectivityManager check](https://github.com/mapbox/mapbox-gl-native/blob/2b05776f978c7759824bce1c4fc89d67cc00d332/platform/android/MapboxGLAndroidSDK/src/main/java/com/mapbox/mapboxsdk/MapboxAccountManager.java#L92-L102) in Android is too aggressive for devices using alternate methods to connect to a network. These users would benefit from a method that manually sets connectivity status. @tobrun recommended looking into https://developer.android.com/training/monitoring-device-state/manifest-receivers.html as a starting point. 

cc @zugaldia @ivovandongen

zugaldia:
Currently, the PR only covers the usecase where an app handles its connectivity state internally and bypasses `ConnectivityManager`.

Fixes #6617.

/cc: @danswick @mapbox/android

### DIFF
--- a/platform/android/MapboxGLAndroidSDK/src/main/java/com/mapbox/mapboxsdk/MapboxAccountManager.java
+++ b/platform/android/MapboxGLAndroidSDK/src/main/java/com/mapbox/mapboxsdk/MapboxAccountManager.java
@@ -109,7 +109,7 @@ public class MapboxAccountManager {
      *
      * @return true if there is an Internet connection, false otherwise
      */
-    public boolean isConnected() {
+    public Boolean isConnected() {
         if (connected != null) {
             // Connectivity state overridden by app
             return connected;
--- a/platform/android/MapboxGLAndroidSDK/src/main/java/com/mapbox/mapboxsdk/MapboxAccountManager.java
+++ b/platform/android/MapboxGLAndroidSDK/src/main/java/com/mapbox/mapboxsdk/MapboxAccountManager.java
@@ -18,6 +18,8 @@ public class MapboxAccountManager {
     private final String accessToken;
     private final Context applicationContext;
 
+    private Boolean connected = null;
+
     /**
      * MapboxAccountManager should NOT be instantiated directly.
      * Use @see MapboxAccountManager#getInstance() instead.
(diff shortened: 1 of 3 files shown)


zugaldia:
For testing purposes, a new SNAPSHOT build has been triggered via https://www.bitrise.io/build/a66f375f4c7a5408 and it's now available in [`mapbox-android-sdk-4.2.0-20161006.204020-83.aar`](https://oss.sonatype.org/content/repositories/snapshots/com/mapbox/mapboxsdk/mapbox-android-sdk/4.2.0-SNAPSHOT/mapbox-android-sdk-4.2.0-20161006.204020-83.aar).

zugaldia:
> This covers the HTTPRequest case, but not the core connectivity. The receiver that is started here will still inform the core about connectivity, through NativeConnectivityListener

Yes. Let's not merge this PR until we take care of the `ConnectivityReceiver` too (maybe instead of exposing a `setConnected` method, we can just simply have another constructor for `MapboxAccountManager` where `start(Context context, String accessToken)` accepts a flag?).

ivovandongen:
> maybe instead of exposing a setConnected method, we can just simply have another constructor for MapboxAccountManager where start(Context context, String accessToken) accepts a flag

Definitely. It will resume less quickly probably and needs some testing to be sure. But this is by far the easiest way for now.

zugaldia:
@ivovandongen Thinking about it, the problem with the constructor approach is that it can only be set once, and if the app connectivity state changes, there's no way to inform `MapboxAccountManager ` of the update. I've added a quick check inside `ConnectivityReceiver` which helps with the original issue in OP. This still has good room for refactoring before merging.

zugaldia:
Updated build happened via https://www.bitrise.io/build/5a1b3138c006b2e2 and it's now available here: [`mapbox-android-sdk-4.2.0-20161007.193849-85.aar`](https://oss.sonatype.org/content/repositories/snapshots/com/mapbox/mapboxsdk/mapbox-android-sdk/4.2.0-SNAPSHOT/mapbox-android-sdk-4.2.0-20161007.193849-85.aar).

zugaldia:
The previous snapshot was accidentally built from the wrong branch (`master`), an correct build happened via https://www.bitrise.io/build/337c5e7fdb9d04a7 and it's now available here: [`mapbox-android-sdk-4.2.0-20161010.174709-89.aar`](https://oss.sonatype.org/content/repositories/snapshots/com/mapbox/mapboxsdk/mapbox-android-sdk/4.2.0-SNAPSHOT/mapbox-android-sdk-4.2.0-20161010.174709-89.aar).

PR #6596, opened 2016-10-07
Refactor GL context creation and headless rendering

kkaefer:
We currently have a `View` object, which is a weird mix of GL context creating, framebuffer management, as well as size + pixel ratio management. Instead, I propose refactoring this as follows:

- `Backend` creates/destroys an OpenGL context (if needed). Contains the `gl::Context` object. Must be used to initialize `Map` and `Canvas` objects. A `Backend` can support multiple `Map` objects.
- `Map` objects are initialized with a fixed pixel ratio; we don't support changing pixel ratio on-the-fly anyway because it is tied to what resources are being loaded.
- `Map` objects can be resized, in logical units (unrelated to the pixel ratio). This affects what tiles are being loaded. It is not tied to a framebuffer or `View` object.
- `Map` objects can take a lambda/`std::function` that is called when the map should be redrawn
- `Canvas`: abstract class that represents something that we can draw upon. Every `render/renderStill` call gets passed a `Canvas` object. The `Map` and `Canvas` objects must be initialized with the same `Backend` object.
  - `DefaultCanvas` is present on platform bindings like iOS and macOS, this encapsulates the default framebuffer
  - `FramebufferCanvas` is an implementation that draws to the framebuffer. Supports methods for obtaining the image rendered.

Open questions:
- Continuously rendering map objects need a way to "invalidate" the map, i.e. tell the rendering driver to request a new frame
- Multiple headless maps should be able to run on the same thread, and when calling `renderStillImage`, it doesn't need to block as long as the event loop continues to run. We should only issue GL calls when we are finally drawing the image

/cc @mikemorris @jfirebaugh


kkaefer:
@jfirebaugh @tmpsantos @brunoabinader @mikemorris This is finally in a good shape and ready for review. I implemented pretty much exactly what is described above.

In particular, the only remaining (new) method of a `View` object is `bind()`. It is called whenever the Map needs to bind that view because it wants to draw on it. When the system provides a default framebuffer (i.e. all UI platforms), we are changing the `gl::Context`'s state to reflect the framebuffer and viewport binding. In particular, the `bindFramebuffer` and `viewport` state values in `gl::Context` have a special treatment because they are always used in conjunction in our code. We previously didn't have a way to explicitly bind the default view because we never bound other framebuffers so far, but we are going to in the future.

This pull request unlocks a few other tasks that aren't part of this PR yet:
- Remove `activate()`/`deactivate()` from the `Backend` object(?)
- Share `Backend` objects across multiple `Map`/`View` objects to prevent excessive context switching in the Node.js use case (where every Map object currently has its own GL context)
- Move `invalidate()` and `notifyMapChange` elsewhere; they shouldn't be part of `Backend`.

include/mbgl/map/backend.hpp
```diff
@@ -0,0 +1,47 @@
+#pragma once
+
+#include <mbgl/map/change.hpp>
+
+#include <memory>
+
+namespace mbgl {
+
+class Map;
```
jfirebaugh: Unused.

include/mbgl/map/view.hpp
```diff
@@ -1,63 +1,22 @@
...
-#include <mbgl/util/chrono.hpp>
 #include <mbgl/util/image.hpp>
 
 #include <array>
-#include <functional>
-#include <memory>
 
 namespace mbgl {
 
 class Map;
```
jfirebaugh: Unused.

include/mbgl/map/view.hpp
```diff
@@ -1,63 +1,22 @@
 #pragma once
 
-#include <mbgl/map/change.hpp>
-#include <mbgl/util/chrono.hpp>
 #include <mbgl/util/image.hpp>
 
 #include <array>
```
jfirebaugh: Unused.

include/mbgl/map/view.hpp
```diff
@@ -1,63 +1,22 @@
 #pragma once
 
-#include <mbgl/map/change.hpp>
-#include <mbgl/util/chrono.hpp>
 #include <mbgl/util/image.hpp>
```
jfirebaugh: Unused.

jfirebaugh:
> - Move `invalidate()` and `notifyMapChange` elsewhere; they shouldn't be part of `Backend`.

Refs #6383

benchmark/api/query.benchmark.cpp
```diff
@@ -2,8 +2,8 @@
 
 #include <mbgl/benchmark/util.hpp>
 #include <mbgl/map/map.hpp>
-#include <mbgl/platform/default/headless_display.hpp>
-#include <mbgl/platform/default/headless_view.hpp>
+#include <mbgl/platform/default/headless_backend.hpp>
+#include <mbgl/platform/default/offscreen_view.hpp>
```
kkaefer: I didn't rename this to have a difference in naming: Offscreen rendering doesn't necessarily imply headless rendering (and vice versa): We can use Offscreen views/texturs in regular rendering as well.

platform/android/src/style/layers/custom_layer.cpp
```diff
@@ -26,7 +26,8 @@ namespace android {
     void CustomLayer::update(jni::JNIEnv&) {
         Log::Debug(mbgl::Event::JNI, "Updating map");
         if (map) {
-            map->update(mbgl::Update::Repaint);
+            // TODO: trigger a rerender
```
kkaefer: implemented

kkaefer:
@jfirebaugh please review again

PR #6624, opened 2016-10-07
[android] - image opt android layout editor image

tobrun:
Cleanup resources for 4.2.0 release. 
Compresses image from 1.1 mb to 166kb, aar-armv7 from 5.7mb to 4.8mb

Result in layout editor:
screen shot 2016-10-07 at 14 28 25


ivovandongen:
@tobrun Great! Are there any other assets we can optimise while we're at it?

tobrun:
I looked at the other image resources, these are all in the 1-6kb range. 
We could at some point introduce svg resources? 
This allows us to ship multiple density folders.

PR #6626, opened 2016-10-07
Add unit tests for TileCoordinate and TilePoint

brunoabinader:
Previously part of #6553.

### DIFF
--- a/src/mbgl/style/source_impl.cpp
+++ b/src/mbgl/style/source_impl.cpp
@@ -176,14 +176,13 @@ void Source::Impl::reloadTiles() {
     cache.clear();
 
     for (auto& pair : tiles) {
-        auto tile = pair.second.get();
-        tile->redoLayout();
+        pair.second->redoLayout();
     }
 }
 
 std::unordered_map<std::string, std::vector<Feature>> Source::Impl::queryRenderedFeatures(const QueryParameters& parameters) const {
     std::unordered_map<std::string, std::vector<Feature>> result;
-    if (renderTiles.empty()) {
+    if (renderTiles.empty() || parameters.geometry.empty()) {
         return result;
     }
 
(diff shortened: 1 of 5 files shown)


src/mbgl/style/source_impl.cpp
```diff
@@ -22,8 +22,12 @@
 namespace mbgl {
 namespace style {
 
+namespace {
```
jfirebaugh: Not necessary here; `static` data and functions are already non-visible beyond their translation unit.

src/mbgl/style/source_impl.cpp
```diff
@@ -168,7 +172,8 @@ void Source::Impl::updateTiles(const UpdateParameters& parameters) {
                                    parameters.debugOptions & MapDebugOptions::Collision };
 
     for (auto& pair : tiles) {
-        pair.second->setPlacementConfig(config);
+        auto& tile = pair.second;
+        tile->setPlacementConfig(config);
```
jfirebaugh: Why introduce a temporary?

src/mbgl/tile/tile.hpp
```diff
@@ -25,8 +27,22 @@ namespace style {
 class Layer;
 } // namespace style
 
+using TilePoint = Point<int16_t>;
```
jfirebaugh: We have a typedef for this already, currently spelled `GeometryCoordinate`. I think `TilePoint` is better, but let's wait to introduce it until we can do a universal rename and at the same time remove `GeometryCoordinates` and `GeometryCollection` in favor of geometry.hpp types.

src/mbgl/tile/tile.hpp
```diff
@@ -25,8 +27,22 @@ namespace style {
...
+
 class Tile : private util::noncopyable {
 public:
+    static TilePoint pointFromTileCoordinate(const UnwrappedTileID& tileID, const TileCoordinatePoint& p) {
```
jfirebaugh: This is more closely related to to `TileCoordinate` than `Tile` -- let's move it there.

test/map/transform.test.cpp
```diff
@@ -514,3 +520,62 @@ TEST(Transform, DefaultTransform) {
     ASSERT_FALSE(transform.resize({{ max, max }}));
     testConversions(nullIsland, center);
 }
+
+TEST(Transform, TileCoordinate)
```
jfirebaugh: Let's put all the `TileCoordinate`-related tests -- the ones here and in tile_point.test.cpp -- in tile_coordinate.test.cpp.

I've been campaigning to get rid of `TileCoordinate` for awhile ([1](https://github.com/mapbox/mapbox-gl-native/pull/3873#issuecomment-182679439), [2](https://github.com/mapbox/mapbox-gl-native/issues/3785), [3](https://github.com/mapbox/mapbox-gl-native/issues/3529#issuecomment-179493928)) -- keeping it and its tests each confined to a single file at least limits the spread as much as possible.

src/mbgl/style/source_impl.cpp
brunoabinader: Intention is to improve readability (`tile` vs. `pair.second`).

If that's agreeable, we can move the object scope out of the for-loop so it only gets updated instead of created/deleted on each pass.

brunoabinader:

src/mbgl/style/source_impl.cpp
jfirebaugh: I prefer the original -- `pair.second->setPlacementConfig(config);` -- over introducing any temporary variables. It's clear from the previous line that you're iterating over tiles.

src/mbgl/style/source_impl.cpp
```diff
@@ -194,35 +193,29 @@ std::unordered_map<std::string, std::vector<Feature>> Source::Impl::queryRendere
...
     }
 
-    if (queryGeometry.empty()) {
-        return result;
-    }
-
     mapbox::geometry::box<double> box = mapbox::geometry::envelope(queryGeometry);
 
-    for (const auto& tilePtr : renderTiles) {
-        const RenderTile& tile = tilePtr.second;
```
kkaefer: I'd keep this; it's `pair` isn't a name that says anything.

Issue #1673, opened 2015-06-02
add padding to annotation tiles


ansis:
In order to properly sort markers across tiles (#988) we need annotation tiles to also include features that are outside of the tile's boundaries, but within a certain distance of the tile edge.

This will let us clip markers to tile boundaries at draw time to avoid incorrect overlap across tiles. Since markers near edges are included in both tiles they will be drawn seamlessly. Half the marker will be drawn by one tile and half by the other. This is how the -js implementation works.

This should be done after https://github.com/mapbox/mapbox-gl-native/pull/1666 and https://github.com/mapbox/mapbox-gl-native/pull/1655 are merged.

This will fix 
![screen shot 2015-06-02 at 11 28 57 am](https://cloud.githubusercontent.com/assets/1421652/7939740/96799a1e-091a-11e5-8250-82dedc0d1415.png)

After implementing, change these two lines:
https://github.com/mapbox/mapbox-gl-native/blob/3dfa317cf9689c5bbd0994f86bdf6b068b1c6716/src/mbgl/renderer/symbol_bucket.cpp#L326
https://github.com/mapbox/mapbox-gl-native/blob/3dfa317cf9689c5bbd0994f86bdf6b068b1c6716/src/mbgl/renderer/painter_symbol.cpp#L122

incanus:
This is stale, right @ansis?

ansis:
I think this still needs to be done. But it is unblocked now

PR #3563, opened 2016-10-07
Annotating Null Island also annotates its antipode

brunoabinader:
If I add an annotation at the coordinates 0°, 0°, an annotation also appears at the coordinates 0°, 180°, but only at z1.

antipodes

### DIFF
--- a/src/mbgl/util/tile_cover.cpp
+++ b/src/mbgl/util/tile_cover.cpp
@@ -147,13 +147,12 @@ std::vector<UnwrappedTileID> tileCover(const LatLngBounds& bounds_, int32_t z) {
         { std::max(bounds_.south(), -util::LATITUDE_MAX), bounds_.west() },
         { std::min(bounds_.north(),  util::LATITUDE_MAX), bounds_.east() });
 
-    const TransformState state;
     return tileCover(
-        TileCoordinate::fromLatLng(state, z, bounds.northwest()).p,
-        TileCoordinate::fromLatLng(state, z, bounds.northeast()).p,
-        TileCoordinate::fromLatLng(state, z, bounds.southeast()).p,
-        TileCoordinate::fromLatLng(state, z, bounds.southwest()).p,
-        TileCoordinate::fromLatLng(state, z, bounds.center()).p,
+        TileCoordinate::fromLatLng(z, bounds.northwest()).p,
+        TileCoordinate::fromLatLng(z, bounds.northeast()).p,
+        TileCoordinate::fromLatLng(z, bounds.southeast()).p,
+        TileCoordinate::fromLatLng(z, bounds.southwest()).p,
+        TileCoordinate::fromLatLng(z, bounds.center()).p,
         z);
 }
 
--- a/include/mbgl/util/projection.hpp
+++ b/include/mbgl/util/projection.hpp
@@ -2,9 +2,9 @@
 
 #include <mbgl/util/constants.hpp>
 #include <mbgl/util/geo.hpp>
+#include <mbgl/util/geometry.hpp>
 #include <mbgl/math/clamp.hpp>
 
-#include <cmath>
 
 namespace mbgl {
 
--- a/include/mbgl/util/projection.hpp
+++ b/include/mbgl/util/projection.hpp
@@ -46,6 +46,22 @@ public:
 
         return LatLng(latitude, longitude);
     }
+
+    static Point<double> project(const LatLng& latLng, double scale) {
+        return Point<double> {
+            util::LONGITUDE_MAX + latLng.longitude,
+            util::LONGITUDE_MAX - util::RAD2DEG * std::log(std::tan(M_PI / 4 + latLng.latitude * M_PI / util::DEGREES_MAX))
+        } * worldSize(scale) / util::DEGREES_MAX;
+    }
+
+    static LatLng unproject(const Point<double>& p, double scale, LatLng::WrapMode wrapMode = LatLng::Unwrapped) {
+        auto p2 = p * util::DEGREES_MAX / worldSize(scale);
+        return LatLng {
+            util::DEGREES_MAX / M_PI * std::atan(std::exp((util::LONGITUDE_MAX - p2.y) * util::DEG2RAD)) - 90.0,
+            p2.x - util::LONGITUDE_MAX,
+            wrapMode
+        };
+    }
 };
 
 } // namespace mbgl
--- a/include/mbgl/util/projection.hpp
+++ b/include/mbgl/util/projection.hpp
@@ -8,14 +8,20 @@
 
 namespace mbgl {
 
+// Spherical Mercator projection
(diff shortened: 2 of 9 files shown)


1ec5:
Please avoid commandeering an issue opened by someone else by converting it into a PR. The resulting page is confusing for anyone who hasn’t been following the conversation, moreso if for whatever reason the PR gets rejected.

jfirebaugh:
I'm not sure this is the right fix. Per #5419 and #1673, we actually _want_ the annotation on the border of a tile to appear in all tiles that share that border. So isn't the issue rather that the subsequent projection math places the feature at the wrong tile coordinate for some of the tiles?

brunoabinader:
Apologies @1ec5 - this was part of a PR creation batch and I should've paid more attention to this not being a recently created issue of mine.

brunoabinader:
You are correct @jfirebaugh - we were actually doing the wrong coordinate system conversions in `SymbolAnnotationImpl::updateLayer` - so in fact #5419 is just a different manifestation of this issue :tada: :tada: :tada:

Going from `Point<double>` in `SymbolAnnotation` to `GeometryCoordinate` involves:
1. Convert `Point<double>` (aka. `LonLat`) to `TileCoordinate` via `TileCoordinate::fromLatLng`
2. Convert `Point<double>` from `TileCoordinate.p` to `GeometryCoordinate` (aka. `TilePoint`) via `TileCoordinate::toGeometryCoordinate`

For 2), we require #6626 - thus I'm rebasing this PR against that. We can either merge #6626 first and rebase this against that or just review and merge this and close #6626, I'm fine with either option.

brunoabinader:
Fixes #5419.

brunoabinader:
Because features can now appear in more than one tile when they intersect with other tile edges, `queryRenderedFeatures` would return wrong values. With https://github.com/mapbox/mapbox-gl-native/pull/3563/commits/a3c3f781e1219345afdea2774343c325a39f15e5 we now filter these results by removing features with the same identifier.

brunoabinader:
Filtering out duplicated annotations affects the results of some query render tests e.g. `edge-cases/null-island` now outputs 1 feature instead of 4. I believe we shouldn't output duplicated features and thus, update the expectations for these.

brunoabinader:
Another side-effect: opaque annotations on tile edges are overdrawn:
screen shot 2016-10-09 at 9 20 47 pm

brunoabinader:
One solution to the opaque feature overdraw is to always enable symbol bucket clipping:
screen shot 2016-10-10 at 10 20 11 am

This potentially fixes #1673 and #5055.

brunoabinader:
> This potentially fixes #1673 and #5055.

Not true for #1673. Features that are not exactly on the tile edges, but close enough are still clipped as you can see below (notice the cross pattern):
screen shot 2016-10-10 at 10 48 56 am

brunoabinader:
i've had an insightful talk with @kkaefer today about the possibility to get rid of stencil clipping for symbols. In sum, these are the options we have so far:

# Option 1: Get rid of stencil clipping
- Filter symbol features in which geometries intersects more than one tile, so only one feature gets inside symbol layer
 - Implies disabling stencil tests
- __Pros__:
  - Fixes collision tile rendering
  - Fixes overdrawn features on tile edges
- __Cons__:
  - Because `symbol-avoid-edges` depends on the stencil clip, it stops working
  - 5 failing render-tests:
    - symbol-avoid-edges/default
    - symbol-avoid-edges/rotated-false
    - symbol-spacing/line-close
    - symbol-spacing/line-far
    - text-tile-edge-clipping/default

# Option 2: Always clip symbol buckets
- This is the default implementation for `MapMode::Still`
- __Pros__:
 - All tests pass
 - Fixes collision tile rendering
- __Cons__:
 - [Clipped features on tile edges](https://github.com/mapbox/mapbox-gl-native/pull/3563#issuecomment-252554312)
 - `queryRenderedFeatures` returns duplicated features - this is desired (?)
  - See e.g. https://github.com/mapbox/mapbox-gl-test-suite/blob/master/query-tests/edge-cases/null-island/expected.json
  - Depends on stencil clip - which @kkaefer suggested to get rid of in the future

brunoabinader:
Pinging @jfirebaugh for evaluating the options.

jfirebaugh:
Pretty sure that for now we want to go with Option 2: Always clip symbol buckets. Getting rid of stencil clipping for symbols is a good long-term goal but there are a number of tricky issues that we will need to solve for that to be viable, and we should fix this bug in the meantime.

brunoabinader:
One way of fixing the clipped features close to tile edges is [increasing the bounds when querying for renderable features within tile bounds](https://github.com/mapbox/mapbox-gl-native/blob/master/src/mbgl/annotation/annotation_manager.cpp#L164-L167) - downside is that we'll eventually insert features to render tiles that won't be visible at all due to stencil clipping.

brunoabinader:
Per conversion with @jfirebaugh, I've decided to split the fixes previously contained in this PR in separate ones:
- [core] Skip duplicated IDs when querying point annotations - #6655 
- [core] Cleanup SymbolLayout - #6666 
- [core] Disable stencil test when rendering collision boxes - #6669

The [commit left for review on this PR](https://github.com/mapbox/mapbox-gl-native/pull/3563/commits/dc81a13bc0f3f3b7173b8f5134e5994dbaa00dec) fixes the immediate cause for the issue initially mentioned on this issue (now PR).

In order to [fix rendering of opaque symbols near tile edges](https://github.com/mapbox/mapbox-gl-native/pull/3563#issuecomment-252503046), I'm moving this discussion to a new issue: #6670

jfirebaugh:
This introduces a `TransformState` to a bunch of functions in order to satisfy a composition of functions -- essentially `TileCoordinate::toGeometryCoordinate(TileCoordinate::fromLatLng(latLng))` -- in which `TransformState` should "cancel out" and be irrelevant to the result.

Can we instead write a method such as `GeometryCoordinate CanonicalTileID::project(const LatLng&)` which does the geographic coordinate ⇢ tile coordinate projection directly? (It doesn't necessarily have to be a member of `CanonicalTileID` -- a standalone function would also be fine.)

brunoabinader:
> Can we instead write a method such as GeometryCoordinate CanonicalTileID::project(const LatLng&) which does the geographic coordinate ⇢ tile coordinate projection directly?

Totally @jfirebaugh - this actually ended up in moving `{un,}project()` into `Projection` class - cleaning up a lot of code in between and moving us one step closer into switchable projection types in the future 😉

src/mbgl/util/tile_cover.cpp
```diff
@@ -149,11 +149,11 @@ std::vector<UnwrappedTileID> tileCover(const LatLngBounds& bounds_, int32_t z) {
 
     const TransformState state;
```
jfirebaugh: Unused.

1ec5:
This would be nice to get on the new release-ios-v3.4.0 branch before it diverges.

PR #6627, opened 2016-10-07
CollisionTile::queryRenderedSymbols should receive the query geometry

brunoabinader:
As per @ansis comments in #6055:
> The entire query geometry should be passed to CollisionTile#queryRenderedSymbols. It should
> rotate each point of the query using rotationMatrix and then take the bbox of that. The fix for
> this should be tested in mapbox-gl-test-suite and should be ported to -gl-js.


src/mbgl/text/collision_tile.cpp
```diff
@@ -153,12 +155,27 @@ Box CollisionTile::getTreeBox(const Point<float>& anchor, const CollisionBox& bo
...
-std::vector<IndexedSubfeature> CollisionTile::queryRenderedSymbols(const mapbox::geometry::box<int16_t>& box, const float scale) {
+std::vector<IndexedSubfeature> CollisionTile::queryRenderedSymbols(const GeometryCollection& queryGeometry, const float scale) {
 
     std::vector<IndexedSubfeature> result;
+    if (queryGeometry.empty()) return result;
+
     std::unordered_map<std::string, std::unordered_set<std::size_t>> sourceLayerFeatures;
 
-    auto anchor = util::matrixMultiply(rotationMatrix, convertPoint<float>(box.min));
+    mapbox::geometry::box<float> box {
```
jfirebaugh: This would be more clearly expressed as a rotating transform (using `std::transform` or a [geometry.hpp specific version](https://github.com/mapbox/geometry.hpp/issues/31)) followed by `mapbox::geometry::envelope`.

src/mbgl/text/collision_tile.cpp
```diff
@@ -153,12 +155,27 @@ Box CollisionTile::getTreeBox(const Point<float>& anchor, const CollisionBox& bo
     };
 }
 
-std::vector<IndexedSubfeature> CollisionTile::queryRenderedSymbols(const mapbox::geometry::box<int16_t>& box, const float scale) {
+std::vector<IndexedSubfeature> CollisionTile::queryRenderedSymbols(const GeometryCollection& queryGeometry, const float scale) {
```
jfirebaugh: I think we can simplify the parameter here, and various other places, to `const GeometryCoordinates& queryGeometry`. There's no need to support multipolygon query geometries: https://github.com/mapbox/mapbox-gl-native/commit/d7a816db0cdea3f2afc0ef234f71abfc721ad56b.

brunoabinader: I perceived this too, though didn't wanted to change behavior - good to know 👍

src/mbgl/text/collision_tile.cpp
brunoabinader: Intuit was to avoid traversing through the geometries twice - first for transforming each point, and then later for obtaining the bounding box. However, I acknowledge the need for readability in complex code like this, so doing as suggested.

brunoabinader:

brunoabinader:
Unit test rendering check failure in `Annotations.VisibleFeatures` looks like a GL driver issue:
screen shot 2016-10-11 at 12 01 04 pm

Thus, I'm skipping rendering checks on `Annotations.VisibleFeatures` for now.

brunoabinader:
Render test failures were caused by rebasing the newly-added query tests `symbol/panned-after-insert` and `symbol/rotated-after-insert` against `mapbox-gl-test-suite` master upstream, which contains tests not yet passable on GL native. I've added those tests on top @jfirebaugh's `native-fixes` branch there.

Issue #6628, opened 2016-10-07
Collision detection for symbol features is broken


brunoabinader:
While debugging #6055, I noticed that symbol features in which placement calculation is performed - `icon-ignore-placement` is set to `false` - all features are rendered, including ones that should have been hidden due to collision with other visible features.

This is probably caused by a bug in `CollisionTile` implementation - which currently contains a few discrepancies from GL JS implementation - e.g. https://github.com/mapbox/mapbox-gl-native/issues/6410#issuecomment-249024186.

/cc @jfirebaugh @ansis

brunoabinader:
Other particularity I've noticed is that this issue mostly happens when we're on a fractional zoom level - I'm attaching a unit test to this to clarify.

brunoabinader:
It turns out most of the issues I've perceived while debugging collision detection were caused either by [symbol annotations not having a buffer on each neighbor tile to improve collision detection](https://github.com/mapbox/mapbox-gl-native/pull/6629) or [duplicated collision results caused by a corner case when a symbol anchor is placed on tile edges](https://github.com/mapbox/mapbox-gl-native/pull/3563).

This only leaves some issues related with perspective collision detection, which are handled in https://github.com/mapbox/mapbox-gl-native/issues/6630.

PR #6629, opened 2016-10-07
Annotation symbols should ignore icon placement

brunoabinader:
We currently set `icon-allow-overlap` to `true` [when creating the annotation symbol layer](https://github.com/mapbox/mapbox-gl-native/blob/master/src/mbgl/annotation/annotation_manager.cpp#L187) but we also need `icon-ignore-placement` set to `true`.

Ignoring icon placements causes symbol annotations to skip calculating the placement scale - which is currently broken (see #6628) and thus returning the correct values for rendered features.

### DIFF
--- a/src/mbgl/annotation/annotation_manager.cpp
+++ b/src/mbgl/annotation/annotation_manager.cpp
@@ -185,6 +185,7 @@ void AnnotationManager::updateStyle(Style& style) {
         layer->setSourceLayer(PointLayerID);
         layer->setIconImage({"{sprite}"});
         layer->setIconAllowOverlap(true);
+        layer->setIconIgnorePlacement(true);
 
         layer->impl->spriteAtlas = &spriteAtlas;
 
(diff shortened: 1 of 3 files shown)


jfirebaugh:
Setting `setIconIgnorePlacement(true)` makes sense, but can you explain the collision_tile.cpp part of this PR?

brunoabinader:
> can you explain the collision_tile.cpp part of this PR?

In `CollisionTile`, features coming from layers in which placement is ignored - `icon-ignore-placement: true` - are inserted in the [ignoredTree](https://github.com/mapbox/mapbox-gl-native/blob/master/src/mbgl/text/collision_tile.cpp#L134-L135) - which [ignores placement scale](https://github.com/mapbox/mapbox-gl-native/blob/master/src/mbgl/text/collision_tile.cpp#L87-L95) - notice how placement scale is calculated only for features from the `tree` container.

This means these features are always visible and can bypass the [placement scale check](https://github.com/mapbox/mapbox-gl-native/blob/master/src/mbgl/text/collision_tile.cpp#L172-L174).

brunoabinader:

PR #1749, opened 2015-06-18
Static analyzer warning in Sprite::parseJSON()

kkaefer:
Caught by Xcode 7:

```
/path/to/mapbox-gl-native/src/mbgl/map/sprite.cpp
/path/to/mapbox-gl-native/src/mbgl/map/sprite.cpp:148:17: Function call argument is an uninitialized value
/path/to/mapbox-gl-native/src/mbgl/map/sprite.cpp:48:13: Calling 'Sprite::parseJSON'
/path/to/mapbox-gl-native/src/mbgl/map/sprite.cpp:120:1: Entered call from 'operator()'
/path/to/mapbox-gl-native/src/mbgl/map/sprite.cpp:130:75: Entering loop body
/path/to/mapbox-gl-native/src/mbgl/map/sprite.cpp:131:13: 'name' initialized to a garbage value
/path/to/mapbox-gl-native/src/mbgl/map/sprite.cpp:148:17: Function call argument is an uninitialized value
```

/cc @kkaefer

PR #6632, opened 2016-10-07
Add admonition about hub pull-request

1ec5:
`hub pull-request` is a nifty tool, but it must be used with care. Commandeering an older issue opened by another contributor creates confusion about who fixed the issue and when. It can be especially problematic if the pull request is declined or fails to fix the original issue. Other contributors are then forced to redo the original issue report, and visitors must follow a trail of derailed issues-turned-PRs.

/ref https://github.com/mapbox/mapbox-gl-native/pull/1749#issuecomment-118731547 https://github.com/mapbox/mapbox-gl-native/pull/3563#issuecomment-252282525

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -13,6 +13,8 @@ If you want to contribute code:
   * [Mapbox iOS SDK](platform/ios/CHANGELOG.md)
   * [Mapbox macOS SDK](platform/macos/CHANGELOG.md)
   * [node-mapbox-gl-native](platform/node/CHANGELOG.md)
+ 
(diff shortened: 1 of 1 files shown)


jfirebaugh:
I assume external contributors will not have the permissions necessary to run `hub pull-request -i` on someone else's issue. So while I agree with the sentiment, I'm not sure it's necessary to add it to CONTRIBUTING.md, whose primary audience is _external_ contributors.

1ec5:
Oh, I hadn’t considered that. Closing – Mapboxers, considered yourselves warned. 😛

PR #6635, opened 2016-10-07
Sample multimap activity

zugaldia:
Adds a new activity to the TestApp that shows multiple maps in the same activity. It uses `SupportMapFragment`.

Current iteration:

video

/cc: @tobrun @cammace

### DIFF
--- a/platform/android/MapboxGLAndroidSDKTestApp/src/main/AndroidManifest.xml
+++ b/platform/android/MapboxGLAndroidSDKTestApp/src/main/AndroidManifest.xml
@@ -123,6 +123,14 @@
                 android:value="@string/category_fragment" />
         </activity>
         <activity
+            android:name=".activity.fragment.MultiMapActivity"
+            android:description="@string/description_multimap"
+            android:label="@string/activity_multimap">
+            <meta-data
+                android:name="@string/category"
+                android:value="@string/category_fragment" />
+        </activity>
+        <activity
             android:name=".activity.camera.ManualZoomActivity"
             android:description="@string/description_camera_zoom"
             android:label="@string/activity_camera_zoom">
@@ -356,7 +364,6 @@
                 android:name="@string/category"
                 android:value="@string/category_style" />
         </activity>
-
         <activity
             android:name=".activity.imagegenerator.PrintActivity"
             android:description="@string/description_print"
@@ -416,7 +423,6 @@
                 android:value="@string/category_annotation" />
         </activity>
 
-
         <!-- For Unit tests -->
         <activity android:name=".activity.style.RuntimeStyleTestActivity" />
         <activity android:name=".activity.style.RuntimeStyleTimingTestActivity" />
@@ -435,7 +441,6 @@
         <!-- android:value="true" /> -->
 
         <service android:name="com.mapbox.mapboxsdk.telemetry.TelemetryService" />
-
     </application>
 
 </manifest>
--- a/platform/android/MapboxGLAndroidSDKTestApp/src/main/java/com/mapbox/mapboxsdk/testapp/activity/fragment/MultiMapActivity.java
+++ b/platform/android/MapboxGLAndroidSDKTestApp/src/main/java/com/mapbox/mapboxsdk/testapp/activity/fragment/MultiMapActivity.java
@@ -1,0 +1,15 @@
+package com.mapbox.mapboxsdk.testapp.activity.fragment;
+
+import android.support.v7.app.AppCompatActivity;
+import android.os.Bundle;
+
+import com.mapbox.mapboxsdk.testapp.R;
+
+public class MultiMapActivity extends AppCompatActivity {
+
+    @Override
+    protected void onCreate(Bundle savedInstanceState) {
+        super.onCreate(savedInstanceState);
+        setContentView(R.layout.activity_multi_map);
+    }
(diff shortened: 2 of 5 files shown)


tobrun:
@zugaldia the following autogenerated test failed:
> com.mapbox.mapboxsdk.activity.gen.fragment.MultiMapActivityTest
java.lang.RuntimeException: Could not launch intent Intent { act=android.intent.action.MAIN flg=0x10000000 cmp=com.mapbox.mapboxsdk.testapp/.activity.fragment.MultiMapActivity } within 45 seconds. Perhaps the main thread has not gone idle within a reasonable amount of time? There could be an animation or something constantly repainting the screen. Or the activity is doing network calls on creation? See the threaddump logs. For your reference the last time the event queue was idle before your activity launch request was 1475867001489 and now the last time the queue went idle was: 1475867001506. If these numbers are the same your activity might be hogging the event queue.
	at android.support.test.runner.MonitoringInstrumentation.startActivitySync(MonitoringInstrumentation.java:362)
	at android.support.test.rule.ActivityTestRule.launchActivity(ActivityTestRule.java:219)
	at android.support.test.rule.ActivityTestRule$ActivityStatement.evaluate(ActivityTestRule.java:255)
	at org.junit.rules.RunRules.evaluate(RunRules.java:20)
	at org.junit.runners.ParentRunner.runLeaf(ParentRunner.java:325)
	at org.junit.runners.BlockJUnit4ClassRunner.runChild(BlockJUnit4ClassRunner.java:78)
	at org.junit.runners.BlockJUnit4ClassRunner.runChild(BlockJUnit4ClassRunner.java:57)
	at org.junit.runners.ParentRunner$3.run(ParentRunner.java:290)
	at org.junit.runners.ParentRunner$1.schedule(ParentRunner.java:71)
	at org.junit.runners.ParentRunner.runChildren(ParentRunner.java:288)
	at org.junit.runners.ParentRunner.access$000(ParentRunner.java:58)
	at org.junit.runners.ParentRunner$2.evaluate(ParentRunner.java:268)
	at org.junit.runners.ParentRunner.run(ParentRunner.java:363)
	at org.junit.runners.Suite.runChild(Suite.java:128)
	at org.junit.runners.Suite.runChild(Suite.java:27)
	at org.junit.runners.ParentRunner$3.run(ParentRunner.java:290)
	at org.junit.runners.ParentRunner$1.schedule(ParentRunner.java:71)
	at org.junit.runners.ParentRunner.runChildren(ParentRunner.java:288)
	at org.junit.runners.ParentRunner.access$000(ParentRunner.java:58)
	at org.junit.runners.ParentRunner$2.evaluate(ParentRunner.java:268)
	at org.junit.runners.ParentRunner.run(ParentRunner.java:363)
	at org.junit.runner.JUnitCore.run(JUnitCore.java:137)
	at org.junit.runner.JUnitCore.run(JUnitCore.java:115)
	at android.support.test.internal.runner.TestExecutor.execute(TestExecutor.java:54)
	at android.support.test.runner.AndroidJUnitRunner.onStart(AndroidJUnitRunner.java:240)
	at android.app.Instrumentation$InstrumentationThread.run(Instrumentation.java:1853)

This is because the MultiMapActivity doesn't have a `mapboxmap` field. The autogenerated test case needs to wait for an onMapReady callback to be executed. This is currently checked with reflection on the `mapboxmap` field of the hosting Activity. To make the test pass, allow setting a mapboxmap value inside MultiMapActivity OR disable generation a test case for this activity. For the latter option, add an entry in `const excludeActivities` in `generate-style-code.js`.

LMK if you need extra 👀  or know a way to improve current generation/testing system.

cc @mapbox/android

zugaldia:
@tobrun Thanks for the 👀 , I opted for excluding the activity from the tests as it doesn't have the concept of a `mapboxmap` field and CI is happy now. Good to merge?

haydenkaizeta:
Is it possible to do this when using `MapView` instead of `MapFragment` ?

tobrun:
yes

haydenkaizeta:
@tobrun Is there an example somewhere of this? Whenever I create my second one it appears looking like the first and is frozen.

haydenkaizeta:
@tobrun I figured out the issue is because I was trying to lay them over each other, which seems to break it. Is there a way to lay maps over each other (such as a full-screen map with a mini-map in the corner)?

tobrun:
See example [here](https://github.com/mapbox/mapbox-gl-native/blob/master/platform/android/MapboxGLAndroidSDKTestApp/src/main/java/com/mapbox/mapboxsdk/testapp/activity/maplayout/DoubleMapActivity.java).

Key to achieve that effect is https://github.com/mapbox/mapbox-gl-native/blob/master/platform/android/MapboxGLAndroidSDKTestApp/src/main/java/com/mapbox/mapboxsdk/testapp/activity/maplayout/DoubleMapActivity.java#L128.

haydenkaizeta:
@tobrun awesome thank you! Is there a simple explaination as to why this is not on by default? Performance issues? I would just like to know because I would enable this in multiple places if not.

haydenkaizeta:
Ah answer to my own question is that you can only have one map layered over other maps

tobrun:
@haydenkaizeta no performance issue, this is just the way SurfaceView works (SurfaceView is used to render gl content).

PR #6636, opened 2016-10-07
Add -mapView:didFinishLoadingStyle: delegate methods

### LINKED ISSUE #6412: Notify MGLMapViewDelegate when style loading completes

friedbunny:
Fixes #6412 by implementing `-mapView:didFinishLoadingStyle:` delegate methods, made possible at the core level by #6371.

Also adds `MGLStyle.description` and a private `MGLStyle.URL` property, which are useful when debugging:

```
<MGLStyle: 0x60800001e0a0; name = "Mapbox Dark", URL = "mapbox://styles/mapbox/dark-v9”>
```

/cc @1ec5

### DIFF
--- a/platform/darwin/src/MGLStyle.mm
+++ b/platform/darwin/src/MGLStyle.mm
@@ -34,6 +34,7 @@
 
 @interface MGLStyle()
 @property (nonatomic, weak) MGLMapView *mapView;
+@property (readonly, copy, nullable) NSURL *URL;
 @end
 
 @implementation MGLStyle
@@ -90,6 +91,10 @@ static NSURL *MGLStyleURL_emerald;
 
 - (NSString *)name {
     return @(self.mapView.mbglMap->getStyleName().c_str());
+}
+
+- (NSURL *)URL {
+    return [NSURL URLWithString:@(self.mapView.mbglMap->getStyleURL().c_str())];
 }
 
 - (MGLStyleLayer *)layerWithIdentifier:(NSString *)identifier
@@ -254,5 +259,12 @@ static NSURL *MGLStyleURL_emerald;
     }
 }
 
+- (NSString *)description
+{
+    return [NSString stringWithFormat:@"<%@: %p; name = %@, URL = %@>",
+            NSStringFromClass([self class]), (void *)self,
+            self.name ? [NSString stringWithFormat:@"\"%@\"", self.name] : self.name,
+            self.URL ? [NSString stringWithFormat:@"\"%@\"", self.URL] : self.URL];
+}
 
 @end
--- a/platform/ios/src/MGLMapViewDelegate.h
+++ b/platform/ios/src/MGLMapViewDelegate.h
@@ -126,6 +126,23 @@ NS_ASSUME_NONNULL_BEGIN
  */
 - (void)mapViewDidFinishRenderingFrame:(MGLMapView *)mapView fullyRendered:(BOOL)fullyRendered;
 
+/**
+ Tells the delegate that the map has just finished loading a style.
+ 
+ This method is called during the initialization of the map view and after any
+ subsequent loading of a new style. This method is called between the
+ `-mapViewWillStartRenderingMap:` and `-mapViewDidFinishRenderingMap:` delegate
+ methods. Changes to sources or layers of the current style do not cause this
+ method to be called.
+ 
+ This method is the earliest opportunity to modify the layout or appearance of
+ the current style before the map view is displayed to the user.
+
+ @param mapView The map view that has just loaded a style.
+ @param style The style that was loaded.
+ */
+- (void)mapView:(MGLMapView *)mapView didFinishLoadingStyle:(MGLStyle *)style;
+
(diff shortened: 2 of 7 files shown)


platform/darwin/src/MGLStyle.mm
```diff
@@ -34,6 +34,7 @@
 
 @interface MGLStyle()
 @property (nonatomic, weak) MGLMapView *mapView;
+@property (readonly, copy, nullable) NSURL *URL;
```
1ec5: Fine for now, but it would be nice to make this public at some point, especially once we have #6386.

platform/ios/src/MGLMapViewDelegate.h
```diff
@@ -115,17 +115,34 @@ NS_ASSUME_NONNULL_BEGIN
...
+/**
+ Tells the delegate that the map has just finished loading a style.
+ 
+ This method is called during the initialization of the map view and after any
+ subsequent loading of a new style. This method is called between the
+ `-mapViewWillStartRenderingMap:` and `-mapViewDidFinishRenderingMap:` delegate
+ methods. Changes to sources or layers of the current style do not cause this
+ method to be called.
+ 
+ This method can be used to initiate runtime styling changes that will be
```
1ec5: We don’t use the term “runtime styling” in API or API documentation. How about “modify the layout or appearance of the current style before the map view is displayed to the user”?

1ec5: Maybe also emphasize that this is the earliest opportunity to make those changes.

1ec5:
> We should add some internal commentary around the fact that the setStyleJSON() functionality in core performs a reset on the internal impl->styleURL, which means this SDK call can come back empty (or whatever the equivalent is in the bindings). 

We don’t currently expose any way for the developer to use `setStyleJSON()`. Once we do in #6386, we may want to allow style URLs to be nullable everywhere. But this PR publicly deals in MGLStyle objects, not style URLs, and the private `URL` property is safely nullable.

incanus:
> We don’t currently expose any way for the developer to use `setStyleJSON()`. 

Right, but my point is to catch things later, when we need to expose this and it looks like a simpler core wiring but actually breaks the functionality in this PR.

1ec5:
OK, I’ve added a note to https://github.com/mapbox/mapbox-gl-native/issues/6386#issuecomment-252351471 to track the expected fallout. I see all the necessary `nullable` qualifiers in the code added by this PR, in any case.

platform/ios/src/MGLMapViewDelegate.h
friedbunny: Went with your suggested wording in 335fa5f106a904fa591e0ee916a5fc0210fe7012.

friedbunny:
How does this look to y’all now, @incanus and @1ec5?

1ec5:
* [ ] Update the iOS and macOS changelogs

platform/darwin/src/MGLStyle.mm
```diff
@@ -92,6 +93,10 @@ - (NSString *)name {
     return @(self.mapView.mbglMap->getStyleName().c_str());
 }
 
+- (NSURL *)URL {
```
incanus: What I was thinking of in https://github.com/mapbox/mapbox-gl-native/pull/6636#pullrequestreview-3370778 is just a comment here, something like: 

```objc
    // note that if `mbglMap->setStyleJSON()` is ever wired up, this can return unexpected results
```

friedbunny: Unexpected in what sense? The private definition of this property says it’s `nullable`. Is there an expectation that `URL` will ends up returning something in the future that isn’t a URL or `null`?

friedbunny:
How’s this for changelogging?
> * Added the `-mapView:didFinishLoadingStyle:` delegate method, which offers the earliest opportunity to modify the layout or appearance of the current style before the map view is displayed to the user. ([#6636](https://github.com/mapbox/mapbox-gl-native/pull/6636))

1ec5:
There's more than one delegate protocol in the library, so best to mention MGLMapViewDelegate by name. Otherwise, that's perfect.

friedbunny:
Added changelog entries and squashed back down to the basic commits.

Issue #6436, opened 2016-09-22
Add API for managing a style’s icon images


1ec5:
MGLStyle should have methods for managing a style’s images based on the mbgl support added in #6375. This will make it easier to work with style properties such as `iconImage`.

We could either implement `-imageNames`, `-image:forName:`, and `-setImage:forName:` or a single, mutable `imagesByName` dictionary property along the lines of #6097. Images would be straight-up NSImages (on macOS) or UIImages (on iOS). Style images are very closely related to `MGLAnnotationImage.image`, but I don’t think we should tangle this API with MGLAnnotationImage at all, because MGLAnnotationImage is sort of a proxy view object while style images only last for the lifetime of the style.

/cc @incanus @frederoni

rmnblm:
I am facing this problem right now. Styling the points with a `MGLCircleStyleLayer` works perfectly:

```
let geoJSONURL = Bundle.main.url(forResource: "siliconvalley", withExtension: "geojson")!

let geoJSONSource = MGLGeoJSONSource(sourceIdentifier: "sv", url: geoJSONURL)
mapView.style().add(geoJSONSource)

let cafeStyleLayer = MGLCircleStyleLayer(layerIdentifier: "sv-cafe-layer", sourceIdentifier: "sv")!
cafeStyleLayer.predicate = NSPredicate(format: "amenity == 'cafe'")
cafeStyleLayer.circleColor = UIColor.brown
mapView.style().add(cafeStyleLayer)

...
```

I tried to replace the `MGLCircleStyleLayer` with a `MGLSymbolStyleLayer` and setting the `iconImage` but couldn't get it to work. What is the recommended way to do this atm?

One possible solution: I could parse the GeoJSON manually (like this: https://github.com/rmnblm/mapbox-ios-examples/blob/master/src/Examples/AsyncPinDroppingViewController.swift) and filter each annotation inside the delegate method `public func mapView(_ mapView: MGLMapView, imageFor annotation: MGLAnnotation) -> MGLAnnotationImage?`. Any other ideas?

1ec5:
Note that iconImage is actually the name of a sprite in the style's sprite sheet, not any arbitrary image in the resource bundle. The feature proposed here would add a way to add an image from your resource bundle to the sprite sheet programmatically.

In the meantime, you could use point annotations if you're OK with the differences in behavior.

rmnblm:
[...]

PR #6637, opened 2016-10-09
[ios, macos] possibility to set custom images to style

rmnblm:
With the changes to the core in https://github.com/mapbox/mapbox-gl-native/pull/6375 we can now add custom images to the sprite atlas. This PR resolves https://github.com/mapbox/mapbox-gl-native/issues/6436 to add the possibility to add custom images from bundles in iOS and macOS.

As proposed from @1ec5 in #6436, I've first named the method `setImage:forName:` but to stay consistent with all other methods of `MGLStyle.h` I've decided to name the method `addImage:forName:`.

**Usage:**

``` swift
let imageURL = Bundle.main.url(forResource: "animal-dog", withExtension: "png")
mapView.style().addImage(imageURL!.path, forName: "dog")

let styleLayer  = MGLSymbolStyleLayer(identifier: "restaurants", source: geoJSONSource)
styleLayer.predicate = NSPredicate(format: "amenity == 'restaurant'")
styleLayer.iconImage = "dog" as MGLStyleAttributeValue!
mapView.style().add(styleLayer)
```

img_0011

**Remarks:**
This being my first PR and I'm relatively new to the codebase of Mapbox GL Native I hope I did everything correctly following the code and contribution guidelines.

During development I had to decide **what** is the argument type for the image being passed to 
As we all know, to use images from a .xcassets folder, you can use `NSImage` on macOS or `UIImage` on iOS but there is no base class for both to use as an argument type in 

```
- (void)addImage:(  -->  NSImage/UIImage  <--   *)image forName:(NSString *)name
```

That's why I've decided to use a `NSString *` which is the absolute path of the image. This solution implies that you only can use images part of the bundle and not part of a .xcassets folder. 

```
- (void)addImage:(NSString *)imagePath forName:(NSString *)name
```

/cc @nitrag

### DIFF
--- a/platform/darwin/src/MGLStyle.h
+++ b/platform/darwin/src/MGLStyle.h
@@ -259,17 +259,19 @@ static const NSInteger MGLStyleDefaultVersion = 9;
 - (void)removeStyleClass:(NSString *)styleClass;
 
 /**
- Sets an image to the sprite atlas with a unique identifier.
+ Sets the name and image to the style.
+ The image can be used by setting the MGLSymbolStyleLayer.iconImage
+ property to the name passed in here
 
- @param image
- @param name The unique identifier of this image
+ @param image The image for the name.
+ @param name The name of the image to set to the style.
  */
 - (void)setImage:(MGLImage *)image forName:(NSString *)name;
 
 /**
- Removes an image from the sprite atlas by its name.
+ Removes a name and its associated image from the style.
 
- @param name The unique identifier of the image to remove
+ @param name The name of the image to remove.
  */
 - (void)removeImageForName:(NSString *)name;
 
--- a/platform/darwin/src/MGLStyle.h
+++ b/platform/darwin/src/MGLStyle.h
@@ -258,6 +258,21 @@ static const NSInteger MGLStyleDefaultVersion = 9;
  */
 - (void)removeStyleClass:(NSString *)styleClass;
 
+/**
+ Adds an image to the sprite atlas by it's bundle URL.
+
+ @param imageURL The URL to the bundled image
+ @param name The unique identifier of this image
+ */
+- (void)addImage:(NSString *)imagePath forName:(NSString *)name;
+
+/**
+ Removes an image from the sprite atlas by its name.
+
+ @param name The unique identifier of the image to remove
+ */
+- (void)removeImage:(NSString *)name;
+
 @end
 
 NS_ASSUME_NONNULL_END
--- a/platform/darwin/src/MGLStyle.h
+++ b/platform/darwin/src/MGLStyle.h
@@ -12,14 +12,14 @@ NS_ASSUME_NONNULL_BEGIN
  A version number identifying the default version of the suite of default styles
  provided by Mapbox. This version number may be passed into one of the
  “StyleURLWithVersion” class methods of MGLStyle.
- 
+
  The value of this constant generally corresponds to the latest released version
  as of the date on which this SDK was published. You can use this constant to
  ascertain the style used by `MGLMapView` and `MGLTilePyramidOfflineRegion` when
  no style URL is specified. Consult the
  <a href="https://www.mapbox.com/api-documentation/#styles">Mapbox Styles API documentation</a>
  for the most up-to-date style versioning information.
- 
+
  @warning The value of this constant may change in a future release of the SDK.
     If you use any feature that depends on a specific aspect of a default style
     – for instance, the minimum zoom level that includes roads – you may use the
@@ -30,8 +30,8 @@ NS_ASSUME_NONNULL_BEGIN
 static const NSInteger MGLStyleDefaultVersion = 9;
 
 /**
- The proxy object for the current map style for customization purposes and a 
- set of convenience methods for creating style URLs of default styles provided 
+ The proxy object for the current map style for customization purposes and a
+ set of convenience methods for creating style URLs of default styles provided
  by Mapbox.
  <a href="https://www.mapbox.com/maps/">Learn more about Mapbox default styles</a>.
  */
@@ -42,9 +42,9 @@ static const NSInteger MGLStyleDefaultVersion = 9;
 /**
  Returns the URL to version 8 of the
  <a href="https://www.mapbox.com/maps/streets/">Mapbox Streets</a> style.
- 
+
  Streets is a general-purpose style with detailed road and transit networks.
- 
+
  `MGLMapView` and `MGLTilePyramidOfflineRegion` use Mapbox Streets when no style
  is specified explicitly.
  */
@@ -53,12 +53,12 @@ static const NSInteger MGLStyleDefaultVersion = 9;
 /**
  Returns the URL to the given version of the
  <a href="https://www.mapbox.com/maps/streets/">Mapbox Streets</a> style.
- 
+
  Streets is a general-purpose style with detailed road and transit networks.
- 
+
  `MGLMapView` and `MGLTilePyramidOfflineRegion` use Mapbox Streets when no style
  is specified explicitly.
- 
+
  @param version The style’s latest released version. As of publication, the
     current version is `9`.
  */
@@ -67,7 +67,7 @@ static const NSInteger MGLStyleDefaultVersion = 9;
 /**
  Returns the URL to version 8 of the
  <a href="https://www.mapbox.com/blog/emerald-gl/">Mapbox Emerald</a> style.
- 
+
  Emerald is a tactile style with subtle textures and dramatic hillshading.
  */
 + (NSURL *)emeraldStyleURL __attribute__((deprecated("Create an NSURL object with the string “mapbox://styles/mapbox/emerald-v8”.")));
@@ -75,9 +75,9 @@ static const NSInteger MGLStyleDefaultVersion = 9;
 /**
  Returns the URL to the given version of the
  <a href="https://www.mapbox.com/maps/outdoors/">Mapbox Outdoors</a> style.
- 
+
  Outdoors is a general-purpose style tailored to outdoor activities.
- 
+
  @param version The style’s latest released version. As of publication, the
     current version is `9`.
  */
@@ -86,7 +86,7 @@ static const NSInteger MGLStyleDefaultVersion = 9;
 /**
  Returns the URL to version 8 of the
  <a href="https://www.mapbox.com/maps/light-dark/">Mapbox Light</a> style.
- 
+
  Light is a subtle, light-colored backdrop for data visualizations.
  */
 + (NSURL *)lightStyleURL __attribute__((deprecated("Use -lightStyleURLWithVersion:.")));
@@ -94,9 +94,9 @@ static const NSInteger MGLStyleDefaultVersion = 9;
 /**
  Returns the URL to the given version of the
  <a href="https://www.mapbox.com/maps/light-dark/">Mapbox Light</a> style.
- 
+
  Light is a subtle, light-colored backdrop for data visualizations.
- 
+
  @param version The style’s latest released version. As of publication, the
     current version is `9`.
  */
@@ -105,7 +105,7 @@ static const NSInteger MGLStyleDefaultVersion = 9;
 /**
  Returns the URL to version 8 of the
  <a href="https://www.mapbox.com/maps/light-dark/">Mapbox Dark</a> style.
- 
+
  Dark is a subtle, dark-colored backdrop for data visualizations.
  */
 + (NSURL *)darkStyleURL __attribute__((deprecated("Use -darkStyleURLWithVersion:.")));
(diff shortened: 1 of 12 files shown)


1ec5:
> As we all know, to use images from a .xcassets folder, you can use `NSImage` on macOS or `UIImage` on iOS but there is no base class for both to use as an argument type

MGLTypes.h declares `MGLColor` as a typedef of NSColor on macOS and UIColor on iOS. Alternatively, you can use conditional compilation (`#if`) to declare two variations of the method, one per platform. It would be far better for the method to take an image object than a string – after all, an NSColor or UIColor can be generated programmatically, without involving any file.

1ec5:
> I've first named the method `setImage:forName:` but to stay consistent with all other methods of `MGLStyle.h` I've decided to name the method `addImage:forName:`.

Methods beginning with “add” typically append an object to an ordered collection, as in the case of `-[NSMutableArray addObject:]` or indeed `-[MGLStyle addLayer:]`, or add an object to an unordered collection, as in the case of `-[MGLMutableSet addObject:]` or `-[MGLStyle addSource:]`. In these cases, either the objects lack unique identifiers or the unique identifier is part of the object being added (MGLStyleLayer, MGLSource). The style image API you’re implementing is more analogous to NSMutableDictionary and NSUserDefaults, which have `-setObject:forKey:` methods.

rmnblm:
Good point about the "add" in front of methods and I totally agree with out but then again, it's inconsistent with the `map.hpp`/`map.cpp` names

``` cpp
void addImage(const std::string&, std::unique_ptr<const SpriteImage>);
void removeImage(const std::string&);
```

while the implementation of `addImage` does indeed **set** the sprite in the atlas

``` cpp
impl->style->spriteAtlas->setSprite(name, std::move(image));
```

> A `-setImage:forKey:` method would accept `nil` to remove an image, so a separate -removeImage: method would be unnecessary.

I can do this but in my opinion this would be misleading. Wouldn't it be better to also follow the API convention of a [NSMutableDictionary](https://developer.apple.com/reference/foundation/nsmutabledictionary)

``` objective-c
- (void)removeImageForName:(NSString *)name;
```

rmnblm:
I just added commit 762f0a39574606f29d670f688ed5fba6c7bee83a with the following changes:
* I added a new type inside `MGLTypes.h` called `MGLImage` which is either an `UIImage` on iOS or a `NSImage` on macOS
* I renamed the method as you proposed, now there are

    ```
    - (void)setImage:(MGLImage *)image forName:(NSString *)name;
    - (void)removeImageForName:(NSString *)name;
    ```
* I reused the code from `MGLMapView.mm` to get a `mbgl::SpriteImage` from a `UIImage`/`NSImage`. This was done by adding two new source files for iOS (`UIImage+MGLAdditions`) and macOS (`NSImage+MGLAdditions`) each.
* I refactored `MGLMapView.mm` in both iOS and macOS to use the new implementations to avoid duplicate code

platform/darwin/src/MGLStyle.h
```diff
@@ -258,6 +258,21 @@ static const NSInteger MGLStyleDefaultVersion = 9;
  */
 - (void)removeStyleClass:(NSString *)styleClass;
 
+/**
+ Sets an image to the sprite atlas with a unique identifier.
```
1ec5: “Sprite atlas” is jargon that most developers would find unintuitive. It’s probably sufficient to describe the image being added to the style (since it lasts until the style URL changes), then mention that the developer can use the image by setting the `MGLSymbolStyleLayer.iconImage` property to the name passed in here.

platform/darwin/src/MGLStyle.h
```diff
@@ -258,6 +258,21 @@ static const NSInteger MGLStyleDefaultVersion = 9;
...
+
+ @param image
```
1ec5: Need to document what `image` is.

platform/darwin/src/MGLStyle.mm
```diff
@@ -254,5 +257,21 @@ - (void)removeStyleClass:(NSString *)styleClass
     }
 }
 
+- (void)setImage:(MGLImage *)image forName:(NSString *)name
+{
+    if (image && name)
```
1ec5: Assert instead of failing gracefully. These two parameters are implied to be `nonnull` in the header.

platform/darwin/src/MGLStyle.mm
```diff
@@ -254,5 +257,21 @@ - (void)removeStyleClass:(NSString *)styleClass
...
+    {
+        auto spriteImage = [MGLImage mbgl_spriteImageFromImage:image];
+        self.mapView.mbglMap->addImage([name UTF8String], std::move(spriteImage));
+    }
+}
+
+- (void)removeImageForName:(NSString *)name
+{
+    if (name)
```
1ec5: Similarly, assert here.

platform/darwin/src/MGLTypes.h
```diff
@@ -3,6 +3,14 @@
 #pragma once
 
 #if TARGET_OS_IPHONE
+@class UIImage;
```
1ec5: 👍

(Sorry, for some reason I thought we had already defined MGLImage, but in fact it was MGLColor that we implemented.)

platform/darwin/src/MGLStyle_Private.h
```diff
@@ -5,6 +5,12 @@
 #import <mbgl/util/default_styles.hpp>
 #include <mbgl/mbgl.hpp>
 
+#if TARGET_OS_IPHONE
```
1ec5: Can we move this over to MGLStyle.mm?

platform/darwin/src/MGLStyle.mm
```diff
@@ -1,5 +1,7 @@
 #import "MGLStyle.h"
 
+#import "MGLTypes.h"
```
1ec5: This should be unnecessary as long as MGLStyle.h imports MGLTypes.h.

platform/macos/macos.xcodeproj/project.pbxproj
```diff
@@ -7,6 +7,8 @@
 	objects = {
 
 /* Begin PBXBuildFile section */
+		30E5781B1DAA857E0050F07E /* NSImage+MGLAdditions.h in Headers */ = {isa = PBXBuildFile; fileRef = 30E578141DAA7D920050F07E /* NSImage+MGLAdditions.h */; settings = {ATTRIBUTES = (Public, ); }; };
```
1ec5: This header’s visibility should be Project, not Public. Making it public would expose C++ syntax to pure Objective-C projects using the SDK.

platform/ios/src/UIImage+MGLAdditions.h
```diff
@@ -0,0 +1,9 @@
+#import <UIKit/UIKit.h>
+
+#include <mbgl/sprite/sprite_image.hpp>
+
+@interface UIImage (MGLAdditions)
+
++ (std::unique_ptr<mbgl::SpriteImage>)mbgl_spriteImageFromImage:(UIImage *)image;
```
1ec5: It would be more natural to have an instance method in this case: `-mgl_spriteImage`. (Also, for consistency’s sake, the class prefix is `mgl` rather than `mbgl`.)

platform/macos/macos.xcodeproj/project.pbxproj
rmnblm: I first had `UIImage+MGLAdditions`/`NSImage+MGLAdditions` public but then realized it wasn't necessary, might have missed to set it back to Project here.

platform/darwin/src/MGLStyle.mm
1ec5: By default, passing in `nil` for a `nonnull` parameter is only a warning rather than an error, so `NSAssert()` is for the benefit of developers who ignore their warnings. (Tsk, tsk!) On the other hand, we haven’t been disciplined about asserting in similar cases – `-addStyleClass:` and `-removeStyleClass:` are more examples of this – so it’s a bit unfair of me to nitpick about this detail. 😊

rmnblm: By the way: Do I have to `NSAssert()` now inside both methods because `(MGLImage *)image` and `(NSString *)name` can't be null anyway and as a side quesstion: What is your recommended way to assert for null? 

``` objc
NSAssert([name isEqual:[NSNull null]], "Name cannot be null");

or 

NSAssert(name == NULL, "Name cannot be null");
```

rmnblm:
I just noticed that I wasn't consistent with my last two commit messages (delete/remove). :expressionless:

platform/darwin/src/MGLStyle.h
```diff
@@ -259,17 +259,19 @@ static const NSInteger MGLStyleDefaultVersion = 9;
 - (void)removeStyleClass:(NSString *)styleClass;
 
 /**
- Sets an image to the sprite atlas with a unique identifier.
+ Sets the name and image to the style.
+ The image can be used by setting the MGLSymbolStyleLayer.iconImage
+ property to the name passed in here
```
1ec5: This method can be used to both add a new image or override an existing one, so we should make that clear. Also, when [jazzy](https://github.com/realm/jazzy/) generates the HTML documentation, it automatically links symbols enclosed in backticks, to help readers navigate the multi-page documentation. This wording would be more verbose, but it leaves nothing open to interpretation:

> Adds or overrides an image used by the style’s layers.
> 
> To use an image in a style layer, give it a unique name using this method, then set the `iconImage` property of an `MGLSymbolStyleLayer` object to that name.

platform/ios/src/UIImage+MGLAdditions.h
```diff
@@ -4,6 +4,6 @@
 
 @interface UIImage (MGLAdditions)
 
-+ (std::unique_ptr<mbgl::SpriteImage>)mbgl_spriteImageFromImage:(UIImage *)image;
++ (std::unique_ptr<mbgl::SpriteImage>)mgl_spriteImage:(UIImage *)image;
```
1ec5: What I meant was that this would become a parameter-less instance method, like this:

```objc
- (std::unique_ptr<mbgl::SpriteImage>)mgl_spriteImage;
```

Within the implementation of an instance method, we have access to all the members of `self`, such as `self.CGImage`. Then you can simply say `annotationImage.image.mgl_spriteImage` to get the `unique_ptr`.

1ec5:
> I just noticed that I wasn't consistent with my last two commit messages (delete/remove). :expressionless:

No worries. Normally we use GitHub’s “Squash and merge” feature, so the title of this PR will become the first line of the squashed commit message. If it bothers you, though, you can use `git rebase -i` to revise your commit messages, then `git push --force-with-lease` to override what you’ve previously pushed to this branch.

1ec5:
It’s a bit unusual to have a setter and remover without an `-imageForName:` getter. `mbgl::Map` doesn’t currently expose a way to get an installed image, so that’ll be a more invasive change. This getter wouldn’t be a high priority, so it’s fine to punt on it for now.

rmnblm:
I also thought about a getter like `-imageForName` but knew that there wasn't such an implementation in `mbgl::Map`. Though it's unusual to have a setter without a getter, I can't think of a use case where I'd need such getter. But somebody might need it in the future, who knows... 🌝

platform/darwin/src/MGLStyle.mm
1ec5: The first form will crash unless `name` is `[NSNull null]` specifically. Since `name` is of type `NSString *` rather than `id`, there is only one way to avoid the crash: set `name` to `nil`. Calling any method on `nil` does nothing and resolves to a `NO` or 0.

The second form will crash unless `name` is `NULL`, which is the opposite of what you’d do in this case. Moreover, in Objective-C, `nil` is conventionally used for Objective-C objects, `NULL` is used for C pointers and double-pointers, and `Nil` is used for Objective-C `Class`es. I recommend reading [this NSHipster article](http://nshipster.com/nil/) to learn about the various ways to say “nothing” in Objective-C.

In this case, you can simply `NSAssert(name)`.

rmnblm: Thanks for the article! If I can follow, you're expecting this:

``` objc
NSAssert(image);
NSAssert(name);
```

1ec5: Yep.

rmnblm: Just pushed it. I had to add a second argument with description otherwise I got error `too few arguments provided to function-like macro invocation`.

1ec5:
Sorry, this ended up conflicting with #6636, which just got merged. Can you resolve the (minor) conflict or check the “Allow edits from maintainers” box so I push the resolution before merging? Thanks!

rmnblm:
I enabled edits from maintainers, the stage is yours 🏟

1ec5:
Thanks! 🎉

friedbunny:
This commit to master [failed on CI](https://www.bitrise.io/build/4e1f733f6c4a8703):

```
 Compiling NSImage+MGLAdditions.mm

❌  /Users/vagrant/git/platform/macos/src/NSImage+MGLAdditions.mm:10:6: use of undeclared identifier 'image'

    [image lockFocus];
     ^



❌  /Users/vagrant/git/platform/macos/src/NSImage+MGLAdditions.mm:12:6: use of undeclared identifier 'image'

    [image unlockFocus];
     ^
```

1ec5:
Thanks @friedbunny. Bitrise CI doesn’t run on third-party PRs, unfortunately. Incoming fix in #6690.

Issue #6638, opened 2016-10-10
map.getProjection().fromScreenLocation modifies actual PointF value


vanshg:
The Android map.getProjection().fromScreenLocation method seems to be flawed. If you pass in a value from a list consecutively, the fromScreenLocation method modifies the original value of the point in the List and therefore you get two different values (and your list is unknowingly modified).

Reproducible with the following code:
```
List<PointF> pointsDrawn = new ArrayList<> ();
pointsDrawn.add(new PointF(500, 500));
LatLng p1 = map.getProjection().fromScreenLocation(pointsDrawn.get(0));
LatLng p2 = map.getProjection().fromScreenLocation(pointsDrawn.get(0));
```

p1 and p2 will have different values, when they should actually be the same value.
Furthermore, `pointsDrawn.get(0)` will return different values before running `fromScreenLocation` and after running `fromScreenLocation`

The code with the issue exists in the MapView#fromScreenLocation
```
LatLng fromScreenLocation(@NonNull PointF point) {
        if (destroyed) {
            return new LatLng();
        }
        point.set(point.x / screenDensity, point.y / screenDensity);
        return nativeMapView.latLngForPixel(point);
    }
```

Proposed fix:
```
LatLng fromScreenLocation(@NonNull PointF point) {
        if (destroyed) {
            return new LatLng();
        }
        PointF copy = new PointF(point.x / screenDensity, point.y / screenDensity);
        return nativeMapView.latLngForPixel(copy);
    }
```

tobrun:
thanks for flagging @vanshg, my only remark would be to reuse a point instead of calling new each time. There are a couple of use-cases where` fromScreenLocation` is called often and limiting object allocation would be great to have. Feel free to PR this.

tobrun:
Will be resolved when https://github.com/mapbox/mapbox-gl-native/pull/6652 is merged.

PR #6646, opened 2016-10-10
Fixed bug where fromScreenLocation would modify the actual Point bein…

vanshg:
…g passed in


vanshg:
https://github.com/mapbox/mapbox-gl-native/issues/6638

tobrun:
I created a follow up PR in #6652 so our tests could and I could fix some code up. 
Thank you for the contribution @vanshg.

PR #6655, opened 2016-10-11
[core] Skip duplicated IDs when querying point annotations

brunoabinader:
Extracted out of https://github.com/mapbox/mapbox-gl-native/pull/3563.

/cc @ivovandongen

### DIFF
--- a/src/mbgl/map/map.cpp
+++ b/src/mbgl/map/map.cpp
@@ -750,13 +750,15 @@ std::vector<Feature> Map::queryRenderedFeatures(const ScreenBox& box, const opti
 
 AnnotationIDs Map::queryPointAnnotations(const ScreenBox& box) {
     auto features = queryRenderedFeatures(box, {{ AnnotationManager::PointLayerID }});
-    AnnotationIDs ids;
-    ids.reserve(features.size());
+    std::set<AnnotationID> set;
     for (auto &feature : features) {
         assert(feature.id);
         assert(*feature.id <= std::numeric_limits<AnnotationID>::max());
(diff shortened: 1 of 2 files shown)


src/mbgl/map/map.cpp
```diff
@@ -755,7 +755,10 @@ AnnotationIDs Map::queryPointAnnotations(const ScreenBox& box) {
     for (auto &feature : features) {
         assert(feature.id);
         assert(*feature.id <= std::numeric_limits<AnnotationID>::max());
-        ids.push_back(static_cast<AnnotationID>(feature.id->get<uint64_t>()));
+        AnnotationID id = static_cast<AnnotationID>(feature.id->get<uint64_t>());
+        if (std::find(ids.begin(), ids.end(), id) == ids.end()) {
```
jfirebaugh: This is an O(n²) algorithm. Better use a set to do deduplication.

brunoabinader:
@1ec5 can you please review the relevant iOS and macOS parts?

src/mbgl/map/map.cpp
brunoabinader: > Now using AnnotationIDs as a typedef for std::set<AnnotationID>.

It turned out that platform code logic depended on particularities of `std::vector` e.g. ability to reassign member values via `std::remove_if` that would end up adding more code complexity by converting it to an approach in which `std::set` is compatible with.

Thus, I'm keeping `AnnotationIDs` as a `std::vector`. I'm using `std::set` internally when issuing the query and later moving those values to a `std::vector`, preserving linear complexity.

PR #6658, opened 2016-10-11
[core] update shaders for line property functions

mollymerp:
attempting to make native work with the shaders update at https://github.com/mapbox/mapbox-gl-shaders/pull/24

~~Currently the macOS build is 'succeeding' but it freezes/times out and the actual app never opens so I'm definitely doing something wrong -- possibly that you can't `#define` macros like that in OpenGL or something?~~ 🔧 fixed 

~~I'm planning on having the device pixel ratio not be hard-coded in the `build-shaders` script, but I figured I'd try to get that working as a first step 💭~~ 🔧 fixed

cc @lucaswoj


scripts/build-shaders.js
```diff
@@ -33,12 +33,15 @@ var data = data.split('\n').map(function(line) {
         } else {
             return params[1] + params[3] + 'p ' + params[4] + ' ' + params[5] + ' = u_' + params[5] + ';';
         }
+    } else if (line.match(pixel_ratio_regex)){
+        return '\n#define DEVICE_PIXEL_RATIO 2.0\n' + line;
```
lucaswoj: This line replaces any occurrences of `/(.*)#define(.*)DEVICE_PIXEL_RATIO(.*)/` in the shader source with `#define DEVICE_PIXEL_RATIO 2.0`. Are there occurrences of `/(.*)#define(.*)DEVICE_PIXEL_RATIO(.*)/` in the shader source?

mollymerp: It actually inserts the hardcoded definition before (`#define DEVICE_PIXEL_RATIO 2.0`) before the matching line in the line-*.vertex shaders. so the output is

```glsl
#define DEVICE_PIXEL_RATIO 2.0
#define ANTIALIASING 1.0 / DEVICE_PIXEL_RATIO / 2.0
```

I've confirmed the build-shaders script is producing the `.hpp` shaders with the expected result but at least one of the shaders is not compiling and I can't find a helpful error message ☹️

lucaswoj: @mollymerp There are some shaders which use the `DEVICE_PIXEL_RATIO` outside of `#define` statements. I suggest following the precedent set in GL JS and adding this `#define` to the top of all shaders.

mollymerp:
Ok I fixed a bug in mapbox/mapbox-gl-shaders#24 and now everything is compiling/building, and I think all I have left to do is figure out how to access `view.getPixelRatio()` from `src/mbgl/gl/shader.cpp` so that I can dynamically `#define` the device pixel ratio instead of hard-coding it at 2.0.

mollymerp:
Soooo this is more or less working, but the render tests show that (I think) antialiasing is messed up and maybe color too? I'll need to investigate that more.

src/mbgl/gl/shader.cpp
```diff
@@ -21,20 +21,24 @@ Shader::Shader(const char* name_,
...
+               gl::ShaderParameters& shaderParameters)
     : name(name_),
       program(context.createProgram()),
       vertexShader(context.createVertexShader()),
       fragmentShader(context.createFragmentShader()) {
     util::stopwatch stopwatch("shader compilation", Event::Shader);
 
-    if (!compileShader(vertexShader, vertexSource)) {
+    std::string vertex(vertexSource);
+    vertex.replace(vertex.find_first_of('\n'), 1,"\n#define DEVICE_PIXEL_RATIO " + std::to_string(shaderParameters.pixelRatio) + "\n");
```
kkaefer: I'm not sure hard-coding the device pixel ratio is a good idea. On desktop, we can switch windows between screens with differing pixel ratio. We don't reload any sprites, but we still do a best effort to render them at the correct size.

src/mbgl/gl/shader.cpp
```diff
@@ -21,20 +21,24 @@ Shader::Shader(const char* name_,
...
+    if (!compileShader(vertexShader, vertex.c_str())) {
         Log::Error(Event::Shader, "Vertex shader %s failed to compile: %s", name, vertexSource);
         throw util::ShaderException(std::string { "Vertex shader " } + name + " failed to compile");
     }
 
+
     std::string fragment(fragmentSource);
-    if (defines & Defines::Overdraw) {
+    fragment.replace(fragment.find_first_of('\n'), 1,"\n#define DEVICE_PIXEL_RATIO "+ std::to_string(shaderParameters.pixelRatio) + "\n");
+    if ((strcmp(name, "collision_box") != 0) & shaderParameters.overdraw) {
```
kkaefer: Why is this hard-coded?

src/mbgl/renderer/painter.cpp
```diff
@@ -91,6 +84,17 @@ void Painter::render(const Style& style, const FrameData& frame_, View& view, Sp
...
+    
+    #ifndef NDEBUG
+        gl::debugging::enable();
+    #endif
+        shaderParameters = gl::ShaderParameters{ frame.pixelRatio, false };
+        shaders = std::make_unique<Shaders>(context, shaderParameters);
+    #ifndef NDEBUG
+        shaderParametersOverdraw = gl::ShaderParameters{ frame.pixelRatio, true };
+        overdrawShaders = std::make_unique<Shaders>(context, shaderParametersOverdraw);
+    #endif
```
kkaefer: `#if/endif` should not be indented; and the code it surrounds needs to be unindented. You can use `clang-format` to apply our standard code formatting.

src/mbgl/renderer/painter.cpp
```diff
@@ -91,6 +84,17 @@ void Painter::render(const Style& style, const FrameData& frame_, View& view, Sp
...
     }
+    
+    #ifndef NDEBUG
+        gl::debugging::enable();
+    #endif
+        shaderParameters = gl::ShaderParameters{ frame.pixelRatio, false };
+        shaders = std::make_unique<Shaders>(context, shaderParameters);
+    #ifndef NDEBUG
+        shaderParametersOverdraw = gl::ShaderParameters{ frame.pixelRatio, true };
+        overdrawShaders = std::make_unique<Shaders>(context, shaderParametersOverdraw);
```
kkaefer: Why did you move this to the `render()` method? This means the shaders are initialized on every frame, which isn't efficient.

src/mbgl/renderer/painter.hpp
```diff
@@ -151,6 +153,8 @@ class Painter : private util::noncopyable {
     LineAtlas* lineAtlas = nullptr;
 
     FrameHistory frameHistory;
+    gl::ShaderParameters shaderParameters;
+    gl::ShaderParameters shaderParametersOverdraw;
```
kkaefer: `ShaderParameters` are only used to initialize the shaders; there's no need to keep them around.

src/mbgl/gl/shader.hpp
```diff
@@ -31,7 +32,7 @@ class Shader : private util::noncopyable {
            const char* vertex,
            const char* fragment,
            Context&,
-           Defines defines = Defines::None);
```
kkaefer: Since `Defines` isn't used anymore, you should also remove the `enum` definition.

kkaefer:
There seem to be a few commits in here that track your progress rather than individual self-contained changes (and some commented out code added, then removed subsequently). Can you please merge/split commits so that the changes represent self-contained units? Our general rule of thumb is that every commit in gl-native should have all tests passing. This helps bigly when using `git bisect` to find regressions later on.

kkaefer:
There seem to be a few commits in here that track your progress rather than individual self-contained changes (and some commented out code added, then removed subsequently). Can you please merge/split commits so that the changes represent self-contained units? Our general rule of thumb is that every commit in gl-native should have all tests passing. This helps bigly when using `git bisect` to find regressions later on.

kkaefer:
There seem to be a few commits in here that track your progress rather than individual self-contained changes (and some commented out code added, then removed subsequently). Can you please merge/split commits so that the changes represent self-contained units? Our general rule of thumb is that every commit in gl-native should have all tests passing. This helps bigly when using `git bisect` to find regressions later on.

kkaefer:
There seem to be a few commits in here that track your progress rather than individual self-contained changes (and some commented out code added, then removed subsequently). Can you please merge/split commits so that the changes represent self-contained units? Our general rule of thumb is that every commit in gl-native should have all tests passing. This helps bigly when using `git bisect` to find regressions later on.

src/mbgl/gl/shader.cpp
kkaefer: I'll take that back; we are initializing `Map` with a constant pixel ratio that can't be changed after the fact anyway.

src/mbgl/gl/shader.cpp
mollymerp: Because I couldn't figure out how to modify things such that the `collision_box_shader` is always initialized with `overdraw = false` 😳

src/mbgl/renderer/painter.cpp
mollymerp: When I put it originally in the `Painter` initialization, `DEVICE_PIXEL_RATIO` was getting set to 0 all the time 😬

mollymerp:
Thanks for the review @kkaefer ! I will squash my commits when this is closer to merging because I don't think any of these commits have all tests passing so far.

mollymerp:
Ok I think this is finally ready for another glance @kkaefer @jfirebaugh 👀

package.json
```diff
@@ -20,10 +20,9 @@
     "csscolorparser": "^1.0.2",
     "ejs": "^2.4.1",
     "express": "^4.11.1",
-    "lodash": "^4.16.4",
```
jfirebaugh: Revert the lodash removal -- this is still used in a couple places.

src/mbgl/gl/context.cpp
```diff
@@ -40,7 +40,7 @@ Context::~Context() {
 
 UniqueShader Context::createShader(ShaderType type, const std::string& source) {
     UniqueShader result { MBGL_CHECK_ERROR(glCreateShader(static_cast<GLenum>(type))), { this } };
-
+
```
jfirebaugh: Revert the whitespace changes in this file.

src/mbgl/programs/line_program.cpp
```diff
@@ -12,13 +12,11 @@ static_assert(sizeof(LineAttributes::Vertex) == 8, "expected LineVertex size");
 
 template <class Values, class...Args>
 Values makeValues(const style::LinePaintProperties& properties,
-                  float pixelRatio,
                   const RenderTile& tile,
                   const TransformState& state,
                   Args&&... args) {
     // the distance over which the line edge fades out.
```
jfirebaugh: Remove the comments here; they were attached to the line that was deleted.

src/mbgl/programs/program.hpp
```diff
@@ -1,33 +1,38 @@
...
+    Program(gl::Context& context, const ProgramParameters& programParameters)
+        : ParentType(context, vertexSource(programParameters), fragmentSource(programParameters))
         {}
 
-    static std::string fragmentSource(ProgramDefines defines) {
+    static std::string fragmentSource(const ProgramParameters& programParameters) {
         std::string fragment = Shaders::fragmentSource;
-        if (defines == ProgramDefines::Overdraw) {
+        
+        fragment.replace(fragment.find_first_of('\n'), 1,"\n#define DEVICE_PIXEL_RATIO "+ std::to_string(programParameters.pixelRatio) + "\n");
```
jfirebaugh: What's the purpose of finding the first newline here? Could it instead simply prepend `"#define DEVICE_PIXEL_RATIO ...\n"` to the source?

src/mbgl/programs/programs.hpp
```diff
@@ -7,26 +7,30 @@
...
-          line(context, defines),
-          lineSDF(context, defines),
-          linePattern(context, defines),
-          raster(context, defines),
-          symbolIcon(context, defines),
-          symbolIconSDF(context, defines),
-          symbolGlyph(context, defines),
-          debug(context, ProgramDefines::None),
-          collisionBox(context, ProgramDefines::None) {
+    Programs(gl::Context& context, ProgramParameters& programParameters)
```
jfirebaugh: `const ProgramParameters&`

src/mbgl/programs/programs.hpp
```diff
@@ -7,26 +7,30 @@
 #include <mbgl/programs/symbol_program.hpp>
 #include <mbgl/programs/debug_program.hpp>
 #include <mbgl/programs/collision_box_program.hpp>
+#include <mbgl/programs/program_parameters.hpp>
+
 
 namespace mbgl {
 
+
+
```
jfirebaugh: Revert this added whitespace.

src/mbgl/renderer/painter.cpp
```diff
@@ -21,6 +22,7 @@
 #include <mbgl/geometry/line_atlas.hpp>
 #include <mbgl/text/glyph_atlas.hpp>
 
+#include <mbgl/programs/program_parameters.hpp>
```
jfirebaugh: Duplicate `#include`.

src/mbgl/programs/program_parameters.hpp
```diff
@@ -0,0 +1,18 @@
+#pragma once
+
+namespace mbgl {
+
+class ProgramParameters {
```
jfirebaugh: Formatting nits:

```C++
class ProgramParameters {
public:
    ProgramParameters(float pixelRatio_ = 1.0, bool overdraw_ = false)
      : pixelRatio(pixelRatio_),
        overdraw(overdraw_) {}

    float pixelRatio;
    bool overdraw;
};
```

src/mbgl/renderer/painter.hpp
```diff
@@ -151,6 +152,7 @@ class Painter : private util::noncopyable {
 
     FrameHistory frameHistory;
 
+
```
jfirebaugh: Revert whitespace change.

src/mbgl/renderer/painter_circle.cpp
```diff
@@ -45,9 +45,7 @@ void Painter::renderCircle(PaintParameters& parameters,
                     pixelsToGLUnits[0] * state.getAltitude(),
                     pixelsToGLUnits[1] * state.getAltitude()
                   }}
-                : pixelsToGLUnits },
-            uniforms::u_devicepixelratio::Value{ frame.pixelRatio },
-        },
+                : pixelsToGLUnits }        },
```
jfirebaugh: Formatting nit:

```C++
                : pixelsToGLUnits }
         },
         *bucket.vertexBuffer,
```

mollymerp:
Thanks @jfirebaugh -- fixed and squashed!

mollymerp:
Launch steps:

- [ ] merge shaders PR https://github.com/mapbox/mapbox-gl-shaders/pull/24
- [ ] merge test-suite PR https://github.com/mapbox/mapbox-gl-test-suite/pull/136
- [ ] update SHAs in native + gl-js https://github.com/mapbox/mapbox-gl-js/pull/3033
- [ ] merge native + gl-js

src/mbgl/programs/program.hpp
```diff
@@ -1,33 +1,38 @@
...
+    Program(gl::Context& context, const ProgramParameters& programParameters)
+        : ParentType(context, vertexSource(programParameters), fragmentSource(programParameters))
         {}
 
-    static std::string fragmentSource(ProgramDefines defines) {
+    static std::string fragmentSource(const ProgramParameters& programParameters) {
         std::string fragment = Shaders::fragmentSource;
-        if (defines == ProgramDefines::Overdraw) {
+        
+        fragment.insert(0, "\n#define DEVICE_PIXEL_RATIO "+ std::to_string(programParameters.pixelRatio) + "\n");
```
jfirebaugh:
```C++
static std::string pixelRatioDefine(const ProgramParameters& parameters) {
    return std::string("#define DEVICE_PIXEL_RATIO ") + std::to_string(parameters.pixelRatio) + "\n";
}

static std::string fragmentSource(const ProgramParameters& parameters) {
    std::string source = pixelRatioDefine(parameters) + Shaders::fragmentSource;
    ...
}

static std::string vertexSource(const ProgramParameters& parameters) {
    return pixelRatioDefine(parameters) + Shaders::vertexSource;
}
```

mollymerp:
@jfirebaugh ✅

jfirebaugh:
@kkaefer Have all your requested changes been addressed?

src/mbgl/renderer/painter.cpp
```diff
@@ -86,9 +87,12 @@ Painter::Painter(gl::Context& context_, const TransformState& state_)
     gl::debugging::enable();
 #endif
 
-    programs = std::make_unique<Programs>(context);
+    ProgramParameters programParameters = ProgramParameters{ pixelRatio, false };
```
kkaefer: Can be shortened to `ProgramParameters programParameters{ pixelRatio, false };`

kkaefer:
Looks like the mapbox-gl-shaders repository branch didn't get into master after this branch got merged?

PR #6660, opened 2016-10-11
[android] - correct android to core conversion of camera options.

tobrun:
Fixes issue of Android not correctly converting the android camera position to the core CameraOptions equivalent as mentioned in https://github.com/mapbox/mapbox-gl-native/pull/6656.

Review @cammace

PR #6671, opened 2016-10-12
[android] - fixup MyLocationView heading

tobrun:
in https://github.com/mapbox/mapbox-gl-native/pull/6660 we updated the way we get and update our camera options. This now needs to be reflected in MyLocationView code when we track the user heading.

 - Using `setCompass(0)` always shows the MyLocationView facing north when we track the user location and the heading.
 - Using ` setCompass(gpsDirection + bearing)` not only shows the direction  of the gps but recalculates it based in current heading of the map.

Review @zugaldia / @cammace

### DIFF
--- a/platform/android/MapboxGLAndroidSDK/src/main/java/com/mapbox/mapboxsdk/maps/widgets/MyLocationView.java
+++ b/platform/android/MapboxGLAndroidSDK/src/main/java/com/mapbox/mapboxsdk/maps/widgets/MyLocationView.java
@@ -55,13 +55,9 @@ public class MyLocationView extends View {
     private float contentPaddingX;
     private float contentPaddingY;
 
-    private Location location;
     private LatLng latLng;
-    private LatLng interpolatedLocation;
-    private LatLng previousLocation;
+    private Location location;
     private long locationUpdateTimestamp;
-
-    private float gpsDirection;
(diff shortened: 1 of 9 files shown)


zugaldia:
@cammace has been working closely with the `MyLocationView` object, I'll let him take the deal in this review.

platform/android/MapboxGLAndroidSDK/src/main/java/com/mapbox/mapboxsdk/maps/widgets/MyLocationView.java
```diff
@@ -702,7 +702,7 @@ void updateLatLng(@NonNull Location location) {
                     builder.bearing(location.getBearing());
                 }
                 gpsDirection = location.getBearing();
```
zugaldia: Isn't this assignment unnecessary after we change to `setCompass(0);`?

tobrun: I followed up this request with some other cleanup and corrections, want to review again?

tobrun:
This ends up fixing multiple parts of the camera API we currenty exposed in combination with fixes for MyLocationView.

PR #6569, opened 2016-10-04
[android] Revise handling of tracking modes and interactions with touch gestures.

clydebarrow:
This fixes issues #6549 and #6567. Also pertinent to #6557.
Additional code has been added to the test app (Activity "User Tracking
Mode") to test.
A potential race condition in the MapboxMap#easeCamera() methods where
a camera change generated by tracking which could have reset the tracking
modes has been eliminated by factoring out a new method
MapboxMap#easeCameraInternal().

PR #6672, opened 2016-10-12
[android] - reset bearing tracking mode while panning the map

tobrun:
It seems that #6569 has changed the business logic when it comes to disabling bearing tracking mode when you pan the map. Since upcoming release is a non-semver release, we can't change this behaviour now. I'm up to discuss this more for future releases but for now, let's disable bearing when you pan the map (FWIW users can easily override this behaviour).

Review @zugaldia / @cammace

### DIFF
--- a/platform/android/MapboxGLAndroidSDK/src/main/java/com/mapbox/mapboxsdk/maps/MapView.java
+++ b/platform/android/MapboxGLAndroidSDK/src/main/java/com/mapbox/mapboxsdk/maps/MapView.java
@@ -1965,7 +1965,7 @@ public class MapView extends FrameLayout {
             requestDisallowInterceptTouchEvent(true);
 
             // reset tracking if needed
-            resetTrackingModesIfRequired(true, false);
+            resetTrackingModesIfRequired(true, true);
             // Cancel any animation
             nativeMapView.cancelTransitions();
 


clydebarrow:
Two comments; was the previous behaviour documented? If not, I would consider it a bug, as I don't see any particular reason why a straight pan should need to change the bearing tracking mode.
Secondly, it's not immediately obvious to me how you would override the behaviour other than by disabling panning. Have I missed something?

tobrun:
> override the behaviour other than by disabling panning. Have I missed something?

You can disable the whole dismissal of tracking modes, and handle this interaction yourself. I'm onboard changing the interaction but see below why.

>  was the previous behaviour documented? If not, I would consider it a bug

This is not how you should approach semver changes. An application will behave differently between 2 versions. What you are proposing is a change in logic, this is something we can't do that now.

cammace:
I've added documentation to the onScroll event to better describe what happens with user location tracking's enabled.

zugaldia:
Qt build is failing though:

```
-- [Mason] Unpacking package to mason_packages/headers/geojson/0.3.2...
CMake Error at /usr/local/lib/cmake/Qt5Core/Qt5CoreConfig.cmake:15 (message):
  The imported target "Qt5::Core" references the file

     "/usr/local/.//mkspecs/macx-clang"

  but this file does not exist.  Possible reasons include:

  * The file was deleted, renamed, or moved to another location.

  * An install or uninstall procedure did not complete successfully.

  * The installation package was faulty and contained

     "/usr/local/lib/cmake/Qt5Core/Qt5CoreConfigExtras.cmake"

  but not all the files it references.

Call Stack (most recent call first):
  /usr/local/lib/cmake/Qt5Core/Qt5CoreConfigExtras.cmake:50 (_qt5_Core_check_file_exists)
  /usr/local/lib/cmake/Qt5Core/Qt5CoreConfig.cmake:141 (include)
  platform/qt/qt5.cmake:1 (find_package)
  platform/qt/qt.cmake:72 (include)
  platform/qt/config.cmake:1 (include)
  CMakeLists.txt:51 (include)
```

tobrun:
I think it's related to https://github.com/mapbox/mapbox-gl-native/pull/6716
Let me rebase to be sure.

clydebarrow:
@tobrun 
> This is not how you should approach semver changes. An application will behave differently between 2 versions. 

Isn't that the effect of all bug fixes?

The previous behaviour was that both bearing and location tracking would be reset on _any_ gesture (assuming the appropriate flags were set.) If that behaviour is to be preserved for now then _all_ calls to `resetTrackingModesIfRequired(boolean translate, boolean rotate)` should pass `true` for both arguments.

tobrun:
Looked into the code from iOS and they are doing the exact thing @clydebarrow is doing.

Issue #5388, opened 2016-06-17
[Android] Dispatch touch event from MarkerView to MapView


Hugoo13:
Platform: Android
Mapbox SDK version: 4.1.0-beta.2

### Steps to trigger behavior
I tried to use MarkerView instead of Marker. When I want to scroll the map, if I touch a marker view, the scroll doesn't work. See here:
![ezgif com-video-to-gif 1](https://cloud.githubusercontent.com/assets/3592767/16150644/a61a7970-3498-11e6-99c7-d443a1bafd6b.gif)

### Expected behavior
When I begin to scroll the map with a finger on a MarkerView, the touch should be delegated to the map.

tobrun:
@Hugoo13 thank you for reaching out. You are correct that this behaviour isn't correct. Will try patching this up and go for the same approach as with iOS. eg. handling click in core instead of the View object itself.

tobrun:
I'm removing this from the `4.1.0` milestone, this change impacts the code a bit too much just before  release.

tobrun:
Looking into this now, solution will be to let core determine pressed marker views instead of using the Android SDK onClick event.

tobrun:
I'm working on a proposed solution in #5639

tobrun:
This is on hold to merge until underlying logic is being sorted out

tobrun:
Explicitly noting that issue depends on #6055 getting resolved first

tobrun:
#6773 got merged to resolve #6055, going to rebase my PR in #6673 and see if it still works.

tobrun:
#5639 was merged, closing

PR #5639, opened 2016-07-11
[android] #5388 - moves touch handling of MarkerViews to core

tobrun:
Refers #5388.

Notes:
- Moves touch interaction of MarkerViews to core
- Replaced average icon width/height for touch surface padding to a fixed size, since this was giving incorrect results. Will need to revisit later to fix this.
- Noticed some strange behaviour that click events were not triggered and had to wait a couple of seconds before being able to click the map.

PR #6673, opened 2016-10-12
[android] Dispatching touch events from MarkerView to MapView

tomlocks:
This is simple solution for [5388](https://github.com/mapbox/mapbox-gl-native/issues/5388). Just dispatch the touch event from layout with markers back to MapView.


tobrun:
Thanks for 👀  and proposing this to unblock us. We currently prefer waiting for #6055 to get resolved and following the approach in https://github.com/mapbox/mapbox-gl-native/pull/5639. Main thinking behind this is that #5639 mimics the exact same behaviour as iOS.

tobrun:
#6055 is resolved and unblocks #5639 to be merged. 
Opting for that solution for now. 
Thanks again for willing to contribute!

PR #6681, opened 2016-10-12
Don't force-update tiles when setting GeoJSON data

kkaefer:
https://github.com/mapbox/mapbox-gl-native/commit/7df9a21f415aa269e7ae435d1500f4a3c7d607e1 introduced a loop that updates all existing tiles when changing the GeoJSON source. This could break in a few ways:

- doing too much on the main thread
- the source could have tiles in the cache, which are not updated
- the source is unused, but still update its tiles

Instead, we should just invalidate all tiles and have our mechanism for querying tiles automatically generate new ones when they are needed.

### DIFF
--- a/platform/android/src/style/sources/source.cpp
+++ b/platform/android/src/style/sources/source.cpp
@@ -35,19 +35,6 @@ namespace android {
         return jni::Make<jni::String>(env, source.getID());
     }
 
-    void Source::updateStyle(jni::jboolean updateClasses) {
-        //Update the style only if attached
-        if (ownedSource == nullptr) {
-            Update flags = mbgl::Update::RecalculateStyle;
-            if(updateClasses) {
-                flags = flags | mbgl::Update::Classes;
-            }
-            map->update(flags);
-        } else {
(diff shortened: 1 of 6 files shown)


platform/android/src/style/sources/source.cpp
```diff
@@ -35,19 +35,6 @@ namespace android {
...
-        if (ownedSource == nullptr) {
-            Update flags = mbgl::Update::RecalculateStyle;
-            if(updateClasses) {
-                flags = flags | mbgl::Update::Classes;
-            }
-            map->update(flags);
-        } else {
-            mbgl::Log::Debug(mbgl::Event::JNI, "Not updating as source is not attached to map (yet)");
-        }
-    }
```
kkaefer: This is removed because it is automatically triggered when updating the GeoJSON data

jfirebaugh:
This introduces flickering, right? It removes the existing data, then asynchronously loads new data. Flickering is the reason I didn't suggest this approach in #6264.

kkaefer:
Ah, good point.

Issue #6651, opened 2016-10-11
Performance issues with hundreds of offline packs


halset:
**Platform:** iOS
**Mapbox SDK version:** latest from master
### Steps to trigger behavior
1. Create two styles with 500+ small offline packs for each of them
2. Resume the packs, but kill the app before the packs are complete.
3. Resume all the packs on app launch
### Expected behavior

Acceptable performance. 
### Actual behavior

Having lots of offline packs for a small number of styles are very slow for a number of reasons. Here are the main issues I have found with profiling and experimenting.
- getCompletedTileCountAndSize is called for each region. Perhaps it could be called once instead of once pr region? That makes it faster, but still rather slow.
- When gathering status of a non-complete pack/region, getTile and getResource are called. There should instead be similar hasTile and hasResource that would not need to copy (and even decompress) the data out of the database.
- Each pack/region has a small number of similar small resources that are fetched from the database a huge number of times. The most busy small resources should be cached in memory for increased performance.
- OfflineDownload::getStatus() parses the style file for each of the regions.

halset:
By adding tiles_count and tiles_size to the regions-table, getCompletedTileCountAndSize got a 1 minute and 15 second speedup on an old iPad 2 with 800+ offline regions and around 800MB of tiles.  

Do you think adding those fields to the regions table would be okay? Or is it not acceptable to double-store such information?

halset:
#6691 handle all of the above points except for the repeated parsing of the same style. After #6691, the repeated style parsing are now the biggest cpu time consumer on startup of my app.

halset:
Updated status.
* hasResource got accepted :) #6707
* Resource cache not accepted #6739.
* Faster getCompletedTileCountAndSize by adding resources.tiles_{count,size} not accepted. I have isolated the change in a [separate branch](https://github.com/halset/mapbox-gl-native/commits/offline-performance-regionCompletedStatus) and added transaction for robustness and use a separate user_version numbering scheme to handle future schema changes from master.
* The repeated parsing of the same style json files are the next bottleneck.

I am using the resource cache and the faster getCompletedTileCountAndSize in my app to get acceptable performance. Hopefully, the future will bring a better solution.

@jfirebaugh mentioned that #4106 with a MultiPolygon could be an alternative to using lots of separate regions. It might work, but my offline regions change in location and number from time to time.

nitrag:
Thank you for your contributions! Following.

halset:
[...]

PR #6691, opened 2016-10-13
offline performance improvements

halset:
This solves most of the issues in #6651
 * use hasTile instead of getTile when figuring out if a tile is present
 * add regions.tiles_{count,size} to speed up getCompletedTileCountAndSize
 * cache used resources in memory


halset:
Should I add resources_{count,size} to regions as well? for consistency? Those are not that important for performance in my app.

jfirebaugh:
Thanks for the contributions @halset! Taking these one-by-one:

* use hasTile instead of getTile when figuring out if a tile is present -- Looks good, I think we can merge this with only minor changes. Would you mind opening a separate PR with just this change?
* add regions.tiles_{count,size} to speed up getCompletedTileCountAndSize -- This will require caution. It's a form of database denomalization, which means that consistency becomes a concern. What if, e.g. a device loses power between `INSERT OR IGNORE INTO region_tiles ...` and `UPDATE regions  SET tiles_count = ...`. We at least need to introduce transactions, and then carefully test their effect on various usage scenarios.
* cache used resources in memory -- I'm afraid we can't accept this change, as it introduces a cache without an expiration policy, a.k.a. a memory leak. In addition, the SQLite page cache should already largely be serving the same purpose. If this is a significant performance improvement, we can consider alternate downstream changes with a similar effect.

halset:
@jfirebaugh Thank you for looking into this.

I have added a separate pull-request for hasTile & co. See #6707.

The getCompletedTileCountAndSize performance (in master) is a problem in my project with lots of offline packs, so I hope we will be able to find an acceptable solution. The regions.tiles_{count,size} fix saved more than a minute of the startup time of my app on an old iPad with all the tiles already loaded.

PR #6707, opened 2016-10-14
added hasRegionResource when the content are not needed by the caller

halset:
When a offline download continues with lots of existing tiles, it is too expensive to get the tile and even decompress it to only check if it exist. This add hasRegionResource, hasResource and hasTile to the existing get-methods. This solve one part of #6651 and is a subset of the earlier #6691.

### DIFF
--- a/platform/default/mbgl/storage/offline_database.cpp
+++ b/platform/default/mbgl/storage/offline_database.cpp
@@ -151,6 +151,15 @@ optional<std::pair<Response, uint64_t>> OfflineDatabase::getInternal(const Resou
     }
 }
 
+optional<int64_t> OfflineDatabase::hasInternal(const Resource& resource) {
+    if (resource.kind == Resource::Kind::Tile) {
+        assert(resource.tileData);
+        return hasTile(*resource.tileData);
+    } else {
+        return hasResource(resource);
+    }
+}
+
 std::pair<bool, uint64_t> OfflineDatabase::put(const Resource& resource, const Response& response) {
     return putInternal(resource, response, true);
 }
@@ -234,6 +243,20 @@ optional<std::pair<Response, uint64_t>> OfflineDatabase::getResource(const Resou
     }
 
     return std::make_pair(response, size);
+}
+
+optional<int64_t> OfflineDatabase::hasResource(const Resource& resource) {
+    // clang-format off
+    Statement stmt = getStatement("SELECT length(data) FROM resources WHERE url = ?");
+    // clang-format on
+
+    stmt->bind(1, resource.url);
+    if (!stmt->run()) {
+        return {};
+    }
+
+    optional<int64_t> size = stmt->get<optional<int64_t>>(0);
+    return size;
 }
 
 bool OfflineDatabase::putResource(const Resource& resource,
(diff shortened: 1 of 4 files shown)


platform/default/mbgl/storage/offline_database.cpp
```diff
@@ -236,6 +245,20 @@ optional<std::pair<Response, uint64_t>> OfflineDatabase::getResource(const Resou
...
+    // clang-format off
+    Statement stmt = getStatement("SELECT length(data) FROM resources WHERE url = ?");
+    // clang-format on
+
+    stmt->bind(1, resource.url);
+    if (!stmt->run()) {
+        return {};
+    }
+
+    optional<int64_t> size = stmt->get<optional<int64_t>>(0);
```
jfirebaugh: Combine this with the subsequent `return` to eliminate the temporary variable.

platform/default/mbgl/storage/offline_database.cpp
```diff
@@ -386,6 +409,32 @@ optional<std::pair<Response, uint64_t>> OfflineDatabase::getTile(const Resource:
...
+    stmt->bind(2, tile.pixelRatio);
+    stmt->bind(3, tile.x);
+    stmt->bind(4, tile.y);
+    stmt->bind(5, tile.z);
+
+    if (!stmt->run()) {
+        return {};
+    }
+
+    optional<int64_t> size = stmt->get<optional<int64_t>>(0);
```
jfirebaugh: Likewise eliminate the temporary variable here.

platform/default/mbgl/storage/offline_download.cpp
```diff
@@ -253,23 +253,40 @@ void OfflineDownload::ensureResource(const Resource& resource,
     *workRequestsIt = util::RunLoop::Get()->invokeCancellable([=]() {
         requests.erase(workRequestsIt);
 
-        optional<std::pair<Response, uint64_t>> offlineResponse =
-            offlineDatabase.getRegionResource(id, resource);
-        if (offlineResponse) {
-            if (callback) {
-                callback(offlineResponse->first);
-            }
+        if (callback) {
```
jfirebaugh: Can you refactor this change so that the common code in each branch of this conditional is not repeated? E.g.:

```c++
auto getResourceSizeInDatabase = [&] () -> optional<uint64_t> {
    if (!callback) {
        return offlineDatabase.hasRegionResource(id, resource);
    }
    optional<std::pair<Response, uint64_t>> response =
        offlineDatabase.getRegionResource(id, resource);
    if (!response) {
        return {};
    }
    callback(response->first);
    return response->second;
}

optional<uint64_t> offlineResponse = getResourceSizeInDatabase();
if (offlineResponse) {
    ...
}
```

platform/default/mbgl/storage/offline_database.cpp
```diff
@@ -562,6 +611,16 @@ optional<std::pair<Response, uint64_t>> OfflineDatabase::getRegionResource(int64
     return response;
 }
 
+optional<int64_t> OfflineDatabase::hasRegionResource(int64_t regionID, const Resource& resource) {
```
jfirebaugh: Please add a test for this [here](https://github.com/mapbox/mapbox-gl-native/blob/master/test/storage/offline_database.test.cpp).

platform/default/mbgl/storage/offline_database.cpp
```diff
@@ -236,6 +245,20 @@ optional<std::pair<Response, uint64_t>> OfflineDatabase::getResource(const Resou
     return std::make_pair(response, size);
 }
 
+optional<int64_t> OfflineDatabase::hasResource(const Resource& resource) {
```
jfirebaugh: This method should update `resources.accessed`, as `getResource` does. You can keep it to a single query by checking `db->changes()` after the update -- if it's 0, you know the resource does not exist.

platform/default/mbgl/storage/offline_database.cpp
```diff
@@ -386,6 +409,32 @@ optional<std::pair<Response, uint64_t>> OfflineDatabase::getTile(const Resource:
     return std::make_pair(response, size);
 }
 
+optional<int64_t> OfflineDatabase::hasTile(const Resource::TileData& tile) {
```
jfirebaugh: Likewise, update `tiles.accessed` in this method.

platform/default/mbgl/storage/offline_database.cpp
jfirebaugh: Actually, I take that back -- it's probably better not to bump the accessed value in the situations where `hasResources` is called. It's a very weak form of access and probably shouldn't effect the eviction priority.

halset:
@jfirebaugh Thank you for the code review. I have tried to fix the issues.

platform/default/mbgl/storage/offline_database.cpp
halset: Skipping this as of your comment for the same on `resources.accessed`.

platform/default/mbgl/storage/offline_database.cpp
halset: Done

platform/default/mbgl/storage/offline_database.cpp
halset: Fine.

Issue #6346, opened 2016-09-15
Configurable API hostname


1ec5:
#6309 is adding the ability to customize the meaning of mapbox: URLs to mbgl. We should offer a convenient way for the developer to modify the API hostname. Two options:

1. Add a `hostname` property to MGLOfflineStorage whose setter calls `mbgl::DefaultFileSource::setAPIBaseURL()`.
2. Have MGLOfflineStorage call `mbgl::DefaultFileSource::setAPIBaseURL()` automatically based on an `MGLMapboxAPIHostname` user default.
3. Same as (2), but as an Info.plist entry.

I prefer either (2) or (3) because MGLOfflineStorage is largely asynchronous and therefore easy to use incorrectly.

It’d be nice to call this option a “hostname” for consistency with the Mapbox Swift libraries’ existing API hostname options. However, “base URL” would be more correct since it looks like you can provide an arbitrary path and scheme in addition to the hostname. MGLAPIClient already has support for a `MGLTelemetryTestServerURL` user default that defaults to `MGLAPIClientBaseURL`, so I guess we can follow that naming scheme.

/cc @boundsj @tobrun

1ec5:
Also, if we end up making this a user default, we should provide constants for `api.mapbox.com` and any alternative hostnames we might need to support publicly.

incanus:
Thoughts on moving this off of the milestone though @1ec5?

PR #6709, opened 2016-10-14
add support for configurable API base URL

incanus:
I think this fixes https://github.com/mapbox/mapbox-gl-native/issues/6346. It's fairly identical to how access tokens work in the SDK. 

> I prefer either (2) or (3) because MGLOfflineStorage is largely asynchronous and therefore easy to use incorrectly.

The `DefaultFileSource::setAPIBaseURL()` that is called under the hood is synchronous. I think we can expect that if someone changes the API base URL midstream during a download operation, unexpected things will happen. This is the same with the access token. Is that what you meant @1ec5? Am I missing something as to why this can't be this simple? 

Builds on @tobrun's work in https://github.com/mapbox/mapbox-gl-native/pull/6309. 

Todo: 

- [x] Finish docs
- [x] Mirror to macOS
- [x] Add tests


incanus:
Also, I thought it best to add a new singleton class `MGLNetworkConfiguration` since going at `MGLOfflineStorage` is non-obvious, particularly when you are not using offline functionality. Much like `MGLAccountManager`, it's single-purpose and clearly indicates what it's for. It's a bit more API surface area but I think it's the right thing to do.

incanus:
One item up for debate here is if `+[MGLNetworkConfiguration apiBaseURL]` should return the default, internal `mbgl::util::API_BASE_URL` value of (currently) `https://api.mapbox.com` or just `nil` in the default case. 

To fit the current setup (returning `nil`) I added proper nullability support in debf1579e4216b63622c7843cc79061c7a385908.

1ec5:
> The `DefaultFileSource::setAPIBaseURL()` that is called under the hood is synchronous. I think we can expect that if someone changes the API base URL midstream during a download operation, unexpected things will happen. This is the same with the access token. Is that what you meant @1ec5?

Yes, that’s what I meant. An Info.plist-based implementation is deal because the developer has an easy way to set it before MGLOfflineStorage is initialized, and because changing the base URL midstream results in undefined behavior.

> Also, I thought it best to add a new singleton class `MGLNetworkConfiguration` since going at `MGLOfflineStorage` is non-obvious, particularly when you are not using offline functionality. Much like `MGLAccountManager`, it's single-purpose and clearly indicates what it's for. It's a bit more API surface area but I think it's the right thing to do.

MGLMapView uses MGLOfflineStorage for online functionality too. “Offline” in this context means fetching and caching, that’s all.

I wonder if it’s really necessary to expose a formal API around this setting. I always imagined this setting to be more similar to the telemetry base URL Info.plist entry we had originally (before turning it into a user default): just a simple dictionary read in MGLOfflineStorage right when we initialize the file source. The only benefit to a formal API would be that people can set the base URL in their application delegate instead of in the Info.plist, but I still think that’s a risky proposition.

incanus:
Ok, well this works out of the `Info.plist` just fine, so I think we can indeed remove the public API—it's easy enough to add if requested in future. 

> “Offline” in this context means fetching and caching, that’s all.

Right, but that's non-obvious as well 😄 

I think I'll just remove the public methods from `MGLNetworkConfiguration`.

incanus:
I just got rid of the setter, keeping the getter, in https://github.com/mapbox/mapbox-gl-native/pull/6709/commits/3ec5510b486e611d4ec55f0c0ccff84891be35db and included docs to indicate that this will return any _custom_ API base URL used (`nil` if none), including from `Info.plist`, which can aid in any debugging.

platform/darwin/src/MGLNetworkConfiguration.h
```diff
@@ -0,0 +1,14 @@
+#import <Foundation/Foundation.h>
+
+NS_ASSUME_NONNULL_BEGIN
+
+@interface MGLNetworkConfiguration : NSObject
+
+/**
+ Returns any custom API base URL in use by instances of MGLMapView and MGLOfflineStorage in the current application.
+ */
++ (nullable NSURL *)apiBaseURL;
```
1ec5: Without a programmatic gett, there’s no point to having a setter, because the developer can just as easily call `[[NSBundle mainBundle] objectForInfoDictionaryKey:@"MGLMapboxAPIBaseURL"]` themselves.

incanus:
How about 85206fda47af830b848dcc2151b0c87d80a3c731 then, which just takes the class private now. Note that I've named the (only) header `MGLNetworkConfiguration_Private.h` and naturally it is not included in `Mapbox.h`. But it does not declare a class extension; the filename is merely a reminder that this is not public API. Look good?

platform/darwin/src/MGLOfflineStorage.mm
```diff
@@ -2,6 +2,7 @@
 
 #import "MGLAccountManager_Private.h"
 #import "MGLGeometry_Private.h"
+#import "MGLNetworkConfiguration_Private.h"
```
1ec5: If there’s no public version of this header, there’s no need for `_Private`.

incanus:
> MGLNetworkConfiguration could be a nice place to hang reachability stuff off of in the future.

My thoughts exactly.

platform/darwin/src/MGLOfflineStorage.mm
incanus: 5c4aa6bd8d9f60141ed90d5905666c87cad53525

1ec5:
Qt builds are failing reliably with the following error, which is certainly unrelated to these changes:

```
-- [Mason] Unpacking package to mason_packages/headers/geojson/0.3.2...
CMake Error at /usr/local/lib/cmake/Qt5Core/Qt5CoreConfig.cmake:15 (message):
  The imported target "Qt5::Core" references the file

     "/usr/local/.//mkspecs/macx-clang"

  but this file does not exist.  Possible reasons include:

  * The file was deleted, renamed, or moved to another location.

  * An install or uninstall procedure did not complete successfully.

  * The installation package was faulty and contained

     "/usr/local/lib/cmake/Qt5Core/Qt5CoreConfigExtras.cmake"

  but not all the files it references.

Call Stack (most recent call first):
  /usr/local/lib/cmake/Qt5Core/Qt5CoreConfigExtras.cmake:50 (_qt5_Core_check_file_exists)
  /usr/local/lib/cmake/Qt5Core/Qt5CoreConfig.cmake:141 (include)
  platform/qt/qt5.cmake:1 (find_package)
  platform/qt/qt.cmake:72 (include)
  platform/qt/config.cmake:1 (include)
  CMakeLists.txt:51 (include)


-- Configuring incomplete, errors occurred!
See also "/Users/vagrant/git/build/qt-macos-x86_64/Debug/CMakeFiles/CMakeOutput.log".
make: *** [build/qt-macos-x86_64/Debug/build.ninja] Error 1
```

1ec5:
Filed #6712 about the Qt5 build bot failure.

PR #6726, opened 2016-10-17
Made telemetry dialog material

cammace:
cc: @zugaldia @tobrun for review


cammace:
Following the material spec, this is what the attribution menu looks like now:



The only thing I was unsuccessful getting to work was tinting the 4 buttons found in the dialog depending on what the apps primary color is.

cammace:
Note, before merging we need to decide on fixing the font color for the dialogs 4 buttons (was trying to get the color to depend on the apps primaryColor) but had issues changing the color outside the XML file.

PR #6727, opened 2016-10-18
[ios, macos] Use swift_name for MGLMultiPoint and subclasses

boundsj:
In https://github.com/mapbox/mapbox-gl-native/pull/6524 we added a class factory method to `MGLMultiPoint`. Since `MGLMultiPoint` is a superclass of `MGLPolyline` and `MGLPolygon`, when [modern Swift 3 translation rules](https://github.com/apple/swift-evolution/blob/master/proposals/0005-objective-c-name-translation.md) are applied, the result is ambiguous class factory method names for MultiPoint -> PolyLine and Polygon.

This change decorates the class factory methods with `NS_SWIFT_NAME({shape_name}(coordinates:count:))` so that, at the call site in a modern Swift app, we can write:

```Swift
MGLPolyline.polyline(coordinates: &coordinates, count: UInt(coordinates.count))
```

cc @friedbunny @1ec5

### DIFF
--- a/platform/darwin/src/MGLMultiPoint.h
+++ b/platform/darwin/src/MGLMultiPoint.h
@@ -22,7 +22,7 @@ NS_ASSUME_NONNULL_BEGIN
  @param count The number of items in the `coords` array.
  @return A new multipoint object.
  */
-+ (instancetype)multiPointWithCoordinates:(CLLocationCoordinate2D *)coords count:(NSUInteger)count;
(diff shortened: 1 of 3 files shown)


1ec5:
We discovered that this change would break some existing Swift applications. So we’re going to fix #5201 for the v3.4.0 release, eliminating the need for the `+multiPointWithCoordinates:count:` method that caused this ambiguity in the first place.

PR #6728, opened 2016-10-18
[ios, macos] Make feature attributes and identifier writable

boundsj:
Instances of MGLFeature subclasses can now have their attributes and identifier properties mutated.

Follow up to https://github.com/mapbox/mapbox-gl-native/issues/6302

cc @1ec5


platform/darwin/src/MGLFeature.h
```diff
@@ -50,7 +50,7 @@ NS_ASSUME_NONNULL_BEGIN
  <a href="https://www.mapbox.com/vector-tiles/mapbox-streets/">Mapbox Streets</a>
  layer reference.
```
1ec5: Point out that you can change this property and `attributes` on an MGLFeature obtained using feature querying, but doing so has no effect.

boundsj:
@1ec5 40681d4 just removes a category in a test that I just remembered. It is no longer required.

PR #6729, opened 2016-10-18
Update docs on gradle configuration

tobrun:
CI doesn't apply heap size configuration from gradle.properties, setting in build.gradle instead.

eg. still seeing:

> To run dex in process, the Gradle daemon needs a larger heap.
It currently has approximately 1662 MB.
For faster builds, increase the maximum heap size for the Gradle daemon to more than 2048 MB.
To do this set org.gradle.jvmargs=-Xmx2048M in the project gradle.properties.
For more information see https://docs.gradle.org/current/userguide/build_environment.html

Even though we applied the suggestion here in #6678.

Review @ivovandongen

### DIFF
--- a/platform/android/CONTRIBUTING_MACOS.md
+++ b/platform/android/CONTRIBUTING_MACOS.md
@@ -46,6 +46,15 @@ export ANDROID_HOME=/<installation location>/android-sdk-macosx
 This environment variable configuration should go into a file that's read on
 your shell's startup, like `~/.profile`.
 
+## Speeding up gradle builds
+
+To optimise you development machine for faster gradle builds create the following `/Users/name/.gradle/gradle.properties`  file:
+
+```
+org.gradle.daemon=true
+org.gradle.jvmargs=-Xmx3072M
+```
+
 ## Running The TestApp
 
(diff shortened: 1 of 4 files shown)


tobrun:
Now with this configuration enabled:

> Running dex as a separate process.

To run dex in process, the Gradle daemon needs a larger heap.
It currently has 1024 MB.
For faster builds, increase the maximum heap size for the Gradle daemon to at least 2560 MB (based on the dexOptions.javaMaxHeapSize = 2g).
To do this set org.gradle.jvmargs=-Xmx2560M in the project gradle.properties.
For more information see https://docs.gradle.org/current/userguide/build_environment.html

So we need to update gradle.properties to `org.gradle.jvmargs=-Xmx2560M`

tobrun:
As a result the java build goes from:

> Total time: 39.107 secs

to

> Total time: 3.335 secs

tobrun:
It's seems that these configurations are not picked up by CI at all. 
For this we should look into fixing this with #6734 or contact Bitrise to improve their docker image. 
For I opted to add these configurations as part of the installation process.

PR #6732, opened 2016-10-18
[android] - update map change events javadoc

tobrun:
In our Map change events we had some:

```
     * <p>
     * This {@link MapChange} is currently not implemented.
     * </p>
```

While they actually are implemented. 

Review @cammace


tobrun:
screen shot 2016-10-18 at 14 51 00 1

screen shot 2016-10-18 at 14 51 35 1

Issue #6612, opened 2016-10-06
Adding/removing a GeoJSON source doesn't show in layer till map is zoomed in/out


cammace:
**Platform:** Nexus 5x running Nougat
**Mapbox SDK version:** `4.2.0-beta.3`
### Steps to trigger behavior
1. Have a map style with a GeoJSON source that is visible.
2. Replace the GeoJSON source with a new one.
3. The layer won't show update till you zoom in/out
### Expected behavior

Updated source shows up instantly when added to map
### Actual behavior

new Source won't show up till after zooming map in or out (invalidating the map?)

cc: @ivovandongen @zugaldia

rmnblm:
Might be the same issue as I had with iOS, see https://github.com/mapbox/mapbox-gl-native/issues/6177#issuecomment-251378211

1ec5:
This sounds different than #6177. As of beta 3, we’re correctly invalidating the map when a layout or paint property or filter is set; however, swapping out a source is probably distinct from changing a filter.

@jfirebaugh, does this sound like something that should’ve been fixed by #6201, or is it another special case?

ivovandongen:
@cammace Just to clarify, this issue is not about updating a `GeoJsonSource` (like the title suggests), but replacing it all together with add/removeSource right?

jfirebaugh:
This is exposing some lax behavior in our runtime styling APIs -- it's not actually intended to be possible to remove a source that's currently in use by one or more layers, add another source with the same ID but different data, and have the layers update to the new data.

Instead, removing an in-use source should either:

* Produce an error indicating that the layers that use it need to be removed first, or
* Automatically remove those layers

ivovandongen:
@jfirebaugh Agreed. I would prefer an error in this situation to make it abundantly clear that this is an unsupported operation.

I removed this from the upcoming release milestone as it is not a blocker.

sddamico:
It should be noted that the same behavior is seen if one changes the backing data in a geojson source
(e.g. layers not not invalidating until the map is invalidated by zoom or some other event)

```
String serializedGeoJsonData = getNewGeoJsonData();
final GeoJsonSource source = mapboxMap.getSourceAs("geoJsonData");
source.setGeoJson(serializedGeoJsonData);
```

sddamico:
My current workaround is `            mapboxMap.moveCamera(CameraUpdateFactory.zoomBy(0.0001f));`

jfirebaugh:
[...]

PR #6735, opened 2016-10-18
Reload tiles on adding sprite image

### LINKED ISSUE #6536: Updating sprite image is not reflected in currently rendered layers
In https://github.com/mapbox/mapbox-gl-native/pull/6375 an api is added to add and remove Sprite images.  While implementing the bindings for Android (https://github.com/mapbox/mapbox-gl-native/pull/6513) I noticed that updating a sprite doesn't affect anything visually.

**Initial situation**
- Add a sprite image
- Add a geojson source with a point
- Add a symbol layer with that source and `source-image` set to the sprite image identifier

_Sprite is displayed correctly,_

**Updating the sprite by adding a new image with the same identifier**
_Image is not updated_

As the size of the sprite cannot be changed, a log entry is shown, nothing more. This might be confusing for developers. See: https://github.com/mapbox/mapbox-gl-native/blob/master/src/mbgl/sprite/sprite_atlas.cpp#L132

**Update the sprite by first removing and then adding a sprite with the same identifier**
_Image is not updated_

In this case; no log entry is shown, nothing happens visually.

cc @tmpsantos

ivovandongen:
Fixes #6536


ivovandongen:
The current way of reloading introduces some flickering on symbols (most notably labels) as all sources for current symbol layers are reloaded. Not sure if there is around this so that only sources are updated that actually reference the sprite image in the currently loaded tiles.

cc @jfirebaugh

jfirebaugh:
As I noted in https://github.com/mapbox/mapbox-gl-native/issues/6536#issuecomment-250790550, I believe we should _prohibit_ the use of `addImage` to update an existing image. Due to the use of a texture atlas and asynchronous layout, I don't think there is an easy way to support the ability to update an existing image on the fly

ivovandongen:
@jfirebaugh 
> I believe we should prohibit the use of addImage to update an existing image.

Agreed. The title of the PR is somewhat misleading (changing now). This is actually fixing a case where you remove an image and later add one with the same id.

ivovandongen:
@jfirebaugh  If you want, I can add an exception in the case where an image is added for an existing id to prevent confusion there (now there is a log warning, only in the case the dimensions don't match).

jfirebaugh:
> This is actually fixing a case where you remove an image and later add one with the same id.

Yeah, but there's a symbol layer that references that ID the whole time right? It's functionally the same as updating an existing image.

Basically, I think the same rules should apply here as in https://github.com/mapbox/mapbox-gl-native/issues/6612#issuecomment-252299676 -- no removing an image that's in use.

ivovandongen:
@jfirebaugh 
> Yeah, but there's a symbol layer that references that ID the whole time right? It's functionally the same as updating an existing image.

Yes

> Basically, I think the same rules should apply here as in #6612 (comment) -- no removing an image that's in use.

I agree on the principle. However, I do see people running in to this quite quickly when using this api. The first example I made, I ran into it myself. Wondering why the image didn't update. It also adds the need to always update the source and the sprite instead of just the sprite.

How about adding an explicit update call then indeed? This would be the most intuitive to use.

This does leave us with the original issues: what to do on `removeImage`. Currently, it removes the sprite, but the rendered tiles will still show it, this feels buggy as it will disappear / appear when zooming for example. We can do two things here; either throw an exception on `removeImage` when the sprite is in use (although this might require some additional bookkeeping) or reload the rendered tiles, which is probably more in line with what the developer would expect.

So the changes would be:
- Adding a check on `addImage` so it throws an exception when adding a sprite with an existing id
- Updating rendered tiles on `removeImage` / `addImage` to avoid "stale" tiles.
- Adding an `updateImage` call that will update an existing image (or add it in case it wasn't before)

jfirebaugh:
The problem is that rendering is in an undefined state in the interval between removing the image from the `SpriteAtlas` and the tiles completing the reload. During that period, the existing buffers for symbol layers may contain references to coordinates within the sprite texture that no longer contain the expected image data -- portions that could have, for example, been overwritten by an unrelated image, possibly with different dimensions.

We could allow in place updates where the image dimensions do not change. This case should not require reloading tiles, only updating the texture and repainting.

PR #6737, opened 2016-10-18
[ios] Optimize event uploading

boundsj:
Upload any events collected when the app was in use for apps that only have "when in use" location permissions. Also, for all apps, avoid posting event data if there is only a single event.

### DIFF
--- a/platform/ios/src/MGLMapboxEvents.m
+++ b/platform/ios/src/MGLMapboxEvents.m
@@ -129,6 +129,7 @@ const NSTimeInterval MGLFlushInterval = 180;
 
 @implementation MGLMapboxEvents {
     NSString *_instanceID;
+    UIBackgroundTaskIdentifier _backgroundTaskIdentifier;
 }
 
 + (void)initialize {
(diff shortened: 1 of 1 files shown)


platform/ios/src/MGLMapboxEvents.m
```diff
@@ -489,6 +504,9 @@ - (void)postEvents:(NS_ARRAY_OF(MGLMapboxEventAttributes *) *)events {
             }
             [strongSelf pushDebugEvent:MGLEventTypeLocalDebug withAttributes:@{MGLEventKeyLocalDebugDescription: @"post",
                                                                                     @"debug.eventsCount": @(events.count)}];
+            
+            [[UIApplication sharedApplication] endBackgroundTask:_backgroundTaskIdentifier];
+            _backgroundTaskIdentifier = UIBackgroundTaskInvalid;
```
friedbunny: It looks like the background task won’t be invalidated if there was a network error (see [L503](https://github.com/mapbox/mapbox-gl-native/pull/6737/files#diff-6e8e31cd0ffe52aeda9c8b2671bcea19R503)), which would mean the app continues on in the background until it’s killed.

boundsj: Thanks! Fixed 7688bcc

friedbunny:
This was originally based on the current `master`, but the `release-ios-v3.4.0` is one Android commit behind that — could you rebase on the release branch?

platform/ios/src/MGLMapboxEvents.m
```diff
@@ -485,10 +500,15 @@ - (void)postEvents:(NS_ARRAY_OF(MGLMapboxEventAttributes *) *)events {
             if (error) {
                 [strongSelf pushDebugEvent:MGLEventTypeLocalDebug withAttributes:@{MGLEventKeyLocalDebugDescription: @"Network error",
                                                                                         @"error": error}];
+                [[UIApplication sharedApplication] endBackgroundTask:_backgroundTaskIdentifier];
+                _backgroundTaskIdentifier = UIBackgroundTaskInvalid;
                 return;
```
friedbunny: If we removed this `return` and then enclosed the subsequent `pushDebugEvent` in an `else` statement, I think we could avoid duplicating the background task invalidation.

boundsj: Thanks again. d648594 refactors as you described and squashes this all into a single commit on top of the release branch.

PR #6739, opened 2016-10-18
[core] cache offline resources in memory

halset:
When working with lots of offline packs, OfflineDatabase::getResource is a bottleneck. The same resources are requested a lot of times. This solve one part of #6651.

I tried a similar pull-request before in #6691, but it did not have a limit on the memory usage and it was coupled with some other performance improvements. I hope this one is a bit better, @jfirebaugh.

### DIFF
--- a/platform/default/mbgl/storage/offline_database.cpp
+++ b/platform/default/mbgl/storage/offline_database.cpp
@@ -201,6 +201,16 @@ std::pair<bool, uint64_t> OfflineDatabase::putInternal(const Resource& resource,
 }
 
 optional<std::pair<Response, uint64_t>> OfflineDatabase::getResource(const Resource& resource) {
+    
+    // looking for cached resource.
+    auto it = resourceCache.find(resource.url);
+    if (it != resourceCache.end()) {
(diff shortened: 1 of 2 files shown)


halset:
Starting up my app with all the tiles/resources already loaded takes some time. Here is a screen shot from Instruments without this patch.
screen shot 2016-10-18 at 22 33 53

And then a screen shot with this patch. As you can see, getResource is no longer there and getRegionCompletedStatus is the next bottleneck.
screen shot 2016-10-18 at 22 09 06

jfirebaugh:
@halset I'm still pretty uncomfortable adding caching to `OfflineDatabase`. Would #4106 support your use case? I.e. if you could define a single region whose geometry was a MultiPolygon, could you use that instead?

halset:
@jfirebaugh Okay. 

My rectangles move around from time to time and grow in number from day to day. And each time a rectangle is added or moved, I will have to keep the previous version until the new is fully loaded.

Would it be better to cache the resources in OfflineDownload? I see that there are one OfflineDownload for each region, but the cache would have to be used across all regions in some way.

PR #6747, opened 2016-10-19
[core] - initialise frame_history with an out of range zoom value

### LINKED ISSUE #6667: Zoomlevel 0, not becoming idle
When looking into the Android CI tests being flaky in #6546. I was able to pinpoint that a map at zoomlevel 0 keeps using CPU resources and allocating memory while it should be in an idle state (as it is with other zoomlevels).
### Analysis

App startup, not showing a map:
screen shot 2016-10-12 at 09 53 07

Open screen with a map at zoom 0:
screen shot 2016-10-12 at 10 02 05

Double tap the map to go to zoom 1:
screen shot 2016-10-12 at 10 06 32 1
### Impact on CI Espresso tests

The Android espresso testing framework is smart enough to wait until the app under test becomes idle (waits for message qeue to be empty and threads to halt). When the test app is idle, it will check if our map has been correctly loaded and notifies that the tests can run.

I noticed that tests showing zoomlevel 0, never start. It seems the app never returns to an idle state. When this occurs, the application stays open for 45 seconds until it espresso times out. Strange behaviour here is if I double tap while the application is timing out the tests will run.

tobrun:
Closes #6667.

This PR initialises `previousZoomIndex` with -1 instead of 0 in `frame_history.cpp`. Solves the issue that `painter->needsAnimation()` keeps on returning true when a map is initialised with the default camera location. This currently results in constantly rerendering the map.

Below you can see it stops using cpu resources vs [the before screenshot](https://cloud.githubusercontent.com/assets/2151639/19302211/81c444c6-9063-11e6-8dd7-5d0236460a6f.png). 

screen shot 2016-10-19 at 10 02 29

More information in #6667.

Review @kkaefer

### DIFF
--- a/src/mbgl/renderer/frame_history.cpp
+++ b/src/mbgl/renderer/frame_history.cpp
@@ -20,6 +20,7 @@ void FrameHistory::record(const TimePoint& now, float zoom, const Duration& dura
         for (int16_t z = 0; z <= zoomIndex; z++) {
             opacities[z] = 255u;
         }
+        previousTime = now;
         firstFrame = false;
     }
 
--- a/src/mbgl/renderer/frame_history.hpp
+++ b/src/mbgl/renderer/frame_history.hpp
@@ -30,7 +30,7 @@ private:
     std::array<uint8_t, 256> changeOpacities;
     std::array<uint8_t, 256> opacities;
 
-    int16_t previousZoomIndex = 0;
+    int16_t previousZoomIndex = -1;
     TimePoint previousTime = TimePoint::min();
     TimePoint time = TimePoint::min();
     bool firstFrame = true;


kkaefer:
There's already a variable called `firstFrame`. Can we use that to detect the presence of the first frame rendered?

tobrun:
@kkaefer thank you for 👀 , will attempt to make it work with `firstFrame`.

tobrun:
@kkaefer in f2b8339 I proposed an alternative solution. The underlying issue is that the following if check never returns true and thus never setting the previous time.

```cpp
    if (zoomIndex != previousZoomIndex) {
        previousZoomIndex = zoomIndex;
        previousTime = now;
    }
```

In f2b8339 I'm setting the previousTime when the `firstFrame` is going to be rendered. This solves the issue that `needsAnimation` keeps returning true. 

Do you feel this is a better approach?

tobrun:
closing in favour of #6979

Issue #5035, opened 2016-05-14
Annotation views on macOS


1ec5:
The iOS and OS X versions of MGLMapView had nearly identical annotation implementations up until annotation views landed on iOS in #4801. @boundsj did a great job keeping the implementation clean, so porting it to OS X should mostly be a copy-paste job.

Along the way, it may be possible to directly share MGLAnnotationImage and MGLAnnotationView between the two SDKs in platform/darwin/ the way the other annotation classes are shared. The downside to sharing MGLAnnotationView would be limited Interface Builder support. We’d need to conditionally make MGLAnnotationView inherit from UIView on iOS and NSView on OS X with conditional compilation, yet IB lost the ability to understand preprocessor commands in Xcode 6.

Zakalicious:
Am I correct that MGLAnnotationViews have not made it into the 0.15.1 release of the macos SDK?

PR #6756, opened 2016-10-19
[6749] fix annotation tap detection (smaller area for annotation imag…

RomainQuidet:
I know tap detection is quite sensitive, my tests show it works well on different annotation images and views sizes, with offsets.


1ec5:
@boundsj @incanus, can you review?

1ec5:
Also, if you don’t mind, please make the corresponding changes in the macOS implementation of MGLMapView. Apart from annotation views (#5035), we’re trying to keep the annotation code consistent between the two platforms.

RomainQuidet:
Ok I'll do it tomorrow for MacOS.

RomainQuidet:
Fun to see that method `(MGLAnnotationTag)annotationTagAtPoint:(NSPoint)point persistingResults:(BOOL)persist` from MGLMapView.mm in macos target as already the correct queryRect size I patch in iOS, meaning the way I patched iOS is correct! As macos does not handle yet MGLAnnotationViews, there's no need to update macos code. The click detection works well on this platform.

1ec5:
Ah, OK. I may be mistaken, but the reason may be because clicking on macOS is much more precise than tapping on iOS, so there was never a need to introduce a “slop area” around an annotation.

margvl:
Hello, in our iOS application we are migrating from MKMapKit to Mapbox, however we are facing an issue, which may be related to this branch. Our annotations selection area is to big and user often selects a different annotations, we did not have this problem in MKMapKit.
Just wanted to know the status of this branch, because it was created in 2016.
Can I expect that it will be merged in near future?

PR #6763, opened 2016-10-19
reduce public header code for shared color type

incanus:
I don't feel super strongly about this, but we could cut down some public API code by merging things like this, leveraging the work in https://github.com/mapbox/mapbox-gl-native/pull/6637. 

Similarly, we could do the same for `UIEdgeInsets` and `NSEdgeInsets`. Speaking of, we call it `MGLEdgeInsetsFromNSEdgeInsets` on both platforms despite taking a `UIEdgeInsets` argument on iOS. 

/cc @1ec5 @friedbunny

### DIFF
--- a/platform/darwin/src/MGLBackgroundStyleLayer.h
+++ b/platform/darwin/src/MGLBackgroundStyleLayer.h
@@ -18,25 +18,14 @@ NS_ASSUME_NONNULL_BEGIN
 
 #pragma mark - Accessing the Paint Attributes
 
-#if TARGET_OS_IPHONE
 /**
  The color with which the background will be drawn.
  
- The default value of this property is an `MGLStyleValue` object containing `UIColor.blackColor`. Set this property to `nil` to reset it to the default value.
+ The default value of this property is an `MGLStyleValue` object containing the color black. Set this property to `nil` to reset it to the default value.
 
  This property is only applied to the style if `backgroundPattern` is set to `nil`. Otherwise, it is ignored.
  */
 @property (nonatomic, null_resettable) MGLStyleValue<MGLColor *> *backgroundColor;
-#else
-/**
- The color with which the background will be drawn.
- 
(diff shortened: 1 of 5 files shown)


1ec5:
This is generated code, so your changes will be wiped out the next time someone runs `make style-code-darwin`.

I guess I don’t see a downside to having the declarations duplicated, since it’s all generated code. Whether using jazzy, Quick Help, or Jump to Definition, everyone sees the most specific content for their platform.

1ec5:
> Speaking of, we call it `MGLEdgeInsetsFromNSEdgeInsets` on both platforms despite taking a `UIEdgeInsets` argument on iOS.

Yeah, that’s been bugging me too. Care to fix that?

incanus:
> This is generated code, so your changes will be wiped out

Ah right, good point. 

> I guess I don’t see a downside to having the declarations duplicated, since it’s all generated code. 

Just to get rid of `ifdef`'s in headers? Like I said, not a huge deal, but it seems odd to declare the new `MGLColor` now but then still branch some public code to handle both `UIColor` and `NSColor`.

1ec5:
> Just to get rid of `ifdef`'s in headers? Like I said, not a huge deal, but it seems odd to declare the new `MGLColor` now but then still branch some public code to handle both `UIColor` and `NSColor`. 

We did this to avoid having Quick Help and jazzy say “MGLColor” and not know how to link it: https://github.com/mapbox/mapbox-gl-native/pull/6168#r76509663.

incanus:
> We did this to avoid having Quick Help and jazzy say “MGLColor” and not know how to link it

Ah, ok, let's scrap the original intent here then. 

What say I repurpose for the `MGLEdgeInsetsFromNSEdgeInsets` problem then? That's all private usage so it's a non-breaking change.

1ec5:
Sure, or even rename it to MGLEdgeInsetsFromObjCEdgeInsets. 😉

1ec5:
Flipping CGVector for macOS has made conditional compilation unavoidable in these headers: #7636.

1ec5:
Stale.

PR #6776, opened 2016-10-20
Fix problems with map view UI layout constraints

### LINKED ISSUE #5549: Compass, logo, info button have ambiguous vertical positions
In iosapp, the compass, logo view, and ℹ️ button each have ambiguous vertical positions at runtime, as reported by Xcode 8:

```
Layout Issues Group
x-xcode-debug-views://7fe5680d30f0: runtime: Layout Issues: Vertical position is ambiguous for UIImageView.

x-xcode-debug-views://7fe5680d30f0: runtime: Layout Issues: Vertical position is ambiguous for UIButton.

x-xcode-debug-views://7fe5680d30f0: runtime: Layout Issues: Vertical position is ambiguous for UIView.
```

/cc @friedbunny

### LINKED ISSUE #6755: CompassView doesn't update position when turning device from landscape to portrait mode
**Platform:** iOS 10.1
**Mapbox SDK version:** 3.4.0-beta.1

Startup the device in **portrait** mode and make the compass view appear:

portrait1

Turn the device to **landscape** mode, the compass view still has the same distance to top like in portrait mode, which is wrong:

landscape

Turn the device back to **portrait**, the compass view distance to top would now be correct when in landscape, but is again wrong in portrait:

portrait2

friedbunny:
This addresses two problems with the map view’s UI elements (compass, attribution button, logo):

### Fixes ambiguous constraints

Fixes #5549.

By declaring that the vertical constraint relationships were `NSLayoutRelationGreaterThanOrEqual`, this meant that there were many positions that could potentially satisfy the constraints. This is not what we want (and not what was being shown) — the vertical placement of these views should be `NSLayoutRelationEqual` to the exact values we calculate.

### Fixes iOS 10 giving us incorrect `topLayoutGuide` values

Fixes #6755, where the vertical offset of the compass would be incorrect after device rotation.

Compared to iOS 9, iOS 10 is giving incorrect values for `topLayoutGuide` when the device is rotated. For reasons I don’t understand, re-adding a direct constraint against `topLayoutGuide` fixes the issue. This walks back a small part of #5671, where our accounting for layout guides moved entirely to `-adjustContentInset`.

This fix isn’t great, but it’s backwards compatible and it works.

/cc @frederoni @1ec5

### DIFF
--- a/platform/ios/src/MGLMapView.mm
+++ b/platform/ios/src/MGLMapView.mm
@@ -757,7 +757,7 @@ public:
     [self.compassViewConstraints addObject:
      [NSLayoutConstraint constraintWithItem:compassContainer
                                   attribute:NSLayoutAttributeTop
-                                  relatedBy:NSLayoutRelationGreaterThanOrEqual
+                                  relatedBy:NSLayoutRelationEqual
                                      toItem:self
                                   attribute:NSLayoutAttributeTop
                                  multiplier:1
@@ -800,7 +800,7 @@ public:
     [self.logoViewConstraints addObject:
      [NSLayoutConstraint constraintWithItem:self
                                   attribute:NSLayoutAttributeBottom
-                                  relatedBy:NSLayoutRelationGreaterThanOrEqual
+                                  relatedBy:NSLayoutRelationEqual
                                      toItem:self.logoView
                                   attribute:NSLayoutAttributeBaseline
                                  multiplier:1
@@ -824,7 +824,7 @@ public:
     [self.attributionButtonConstraints addObject:
      [NSLayoutConstraint constraintWithItem:self
                                   attribute:NSLayoutAttributeBottom
-                                  relatedBy:NSLayoutRelationGreaterThanOrEqual
+                                  relatedBy:NSLayoutRelationEqual
                                      toItem:self.attributionButton
                                   attribute:NSLayoutAttributeBaseline
                                  multiplier:1
--- a/platform/ios/src/MGLMapView.mm
+++ b/platform/ios/src/MGLMapView.mm
@@ -754,14 +754,18 @@ public:
     [constraintParentView removeConstraints:self.compassViewConstraints];
     [self.compassViewConstraints removeAllObjects];
 
+    // Work around an apparent iOS 10 bug where the top layout guide isn't always updated
+    // unless there's an active constraint against it. This means we cannot constrain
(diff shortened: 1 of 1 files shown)


frederoni:
Assumed the Swift equivalent would be something like `po UIApplication.shared.delegate?.window._autotraceLayout` but apparently not. 😕

friedbunny:
As a private method, I think it’s not exposed in Swift (which is what lldb expects when you’re in a Swift app). [Looks like](http://nshint.io/blog/2015/08/17/autolayout-breakpoints/) `expr -l objc++ -O -- [[UIWindow keyWindow] _autolayoutTrace]` is how you’d do it. :wink:

1ec5:
The inequality constraints were originally added in #1328. Can you verify that these changes don’t regress #1327?

friedbunny:
The bottom elements continue to move with the map view, but the compass has gone MIA:

img_4568 2

frederoni:
If you're using a map view that is not full size, you should set `automaticallyAdjustsScrollViewInsets` to false. I rebased this PR and added a forced re-layout after the device has been rotated which takes care of the layout bug in iOS 10.

The following cases has been tested in iOS 8, 9 and 10:
- ✅  Full size map view, w/ navbar and tabbar w/ `automaticallyAdjustsScrollViewInsets=YES`.
- ✅  Half size map view w/ `automaticallyAdjustsScrollViewInsets=NO`.

ketz:
Sorry if I'm being thick, but I'm having the same problem when resizing the map view to show an info panel with a POI's details.

Where exactly should I set automaticallyAdjustsScrollViewInsets:NO? I'm not using auto layout, and all my views are created programatically.

thanks

1ec5:
> Where exactly should I set automaticallyAdjustsScrollViewInsets:NO?

`automaticallyAdjustsScrollViewInsets` is a property on the view controller that contains MGLMapView.

1ec5:
> If you're using a map view that is not full size, you should set automaticallyAdjustsScrollViewInsets to false.

@frederoni, just to be clear, though, we haven’t regressed #1327, have we? The compass should never float outside of the map view, even if you have a non-full-screen map view and `automaticallyAdjustsScrollViewInsets` set to `YES`.

1ec5:
I reopened #6954 to track the regression exacerbated by this PR.

PR #6790, opened 2016-10-21
[core] Transform{,State} cleanups

brunoabinader:
Reuse `TransformState::{zoomScale,scaleZoom,getZoomFraction}` and `util::log2` whenever possible.

### DIFF
--- a/src/mbgl/layout/symbol_layout.cpp
+++ b/src/mbgl/layout/symbol_layout.cpp
@@ -418,6 +418,7 @@ std::unique_ptr<SymbolBucket> SymbolLayout::place(CollisionTile& collisionTile) 
 
 template <typename Buffer>
 void SymbolLayout::addSymbols(Buffer &buffer, const SymbolQuads &symbols, float scale, const bool keepUpright, const style::SymbolPlacementType placement, const float placementAngle) {
+    constexpr const uint16_t vertexLength = 4;
     const float placementZoom = util::max(util::log2(scale) + zoom, 0.0f);
 
     for (const auto& symbol : symbols) {
@@ -447,16 +448,15 @@ void SymbolLayout::addSymbols(Buffer &buffer, const SymbolQuads &symbols, float 
             minZoom = 0;
         }
 
-        const int glyph_vertex_length = 4;
-
-        if (buffer.segments.empty() || buffer.segments.back().vertexLength + glyph_vertex_length > 65535) {
+        if (buffer.segments.back().vertexLength + vertexLength > std::numeric_limits<uint16_t>::max()) {
             buffer.segments.emplace_back(buffer.vertices.size(), buffer.triangles.size());
         }
 
         // We're generating triangle fans, so we always start with the first
         // coordinate in this polygon.
         auto& segment = buffer.segments.back();
-        size_t index = segment.vertexLength;
+        assert(segment.vertexLength <= std::numeric_limits<uint16_t>::max());
+        uint16_t index = segment.vertexLength;
 
         // Encode angle of glyph
         uint8_t glyphAngle = std::round((symbol.glyphAngle / (M_PI * 2)) * 256);
@@ -472,14 +472,10 @@ void SymbolLayout::addSymbols(Buffer &buffer, const SymbolQuads &symbols, float 
                             minZoom, maxZoom, placementZoom, glyphAngle);
 
         // add the two triangles, referencing the four coordinates we just inserted.
-        buffer.triangles.emplace_back(static_cast<uint16_t>(index + 0),
-                                      static_cast<uint16_t>(index + 1),
-                                      static_cast<uint16_t>(index + 2));
-        buffer.triangles.emplace_back(static_cast<uint16_t>(index + 1),
-                                      static_cast<uint16_t>(index + 2),
-                                      static_cast<uint16_t>(index + 3));
+        buffer.triangles.emplace_back(index + 0, index + 1, index + 2);
+        buffer.triangles.emplace_back(index + 1, index + 2, index + 3);
 
-        segment.vertexLength += glyph_vertex_length;
+        segment.vertexLength += vertexLength;
         segment.primitiveLength += 2;
     }
 }
--- a/src/mbgl/layout/symbol_layout.cpp
+++ b/src/mbgl/layout/symbol_layout.cpp
@@ -11,12 +11,12 @@
 #include <mbgl/util/constants.hpp>
 #include <mbgl/util/utf.hpp>
 #include <mbgl/util/token.hpp>
-#include <mbgl/util/math.hpp>
 #include <mbgl/util/std.hpp>
 #include <mbgl/util/constants.hpp>
 #include <mbgl/util/string.hpp>
 #include <mbgl/math/clamp.hpp>
 #include <mbgl/math/minmax.hpp>
+#include <mbgl/math/log2.hpp>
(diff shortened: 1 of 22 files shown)


src/mbgl/renderer/painter_symbol.cpp
```diff
@@ -256,8 +256,7 @@ void Painter::renderSymbol(PaintParameters& parameters,
         auto& collisionBoxShader = shaders->collisionBox;
         context.program = collisionBoxShader.getID();
         collisionBoxShader.u_matrix = tile.matrix;
-        // TODO: This was the overscaled z instead of the canonical z.
-        collisionBoxShader.u_scale = std::pow(2, state.getZoom() - tile.id.canonical.z);
+        collisionBoxShader.u_scale = state.zoomScale(state.getZoomFraction());
```
jfirebaugh: Is this correct in the case of overzoomed tiles?

brunoabinader: Good question - I'm not sure. I'll revert this particular change.

src/mbgl/tile/geometry_tile.cpp
```diff
@@ -139,7 +139,7 @@ void GeometryTile::queryRenderedFeatures(
                         queryGeometry,
                         transformState.getAngle(),
                         util::tileSize * id.overscaleFactor(),
-                        std::pow(2, transformState.getZoom() - id.overscaledZ),
+                        transformState.zoomScale(transformState.getZoomFraction()),
```
jfirebaugh: (Same question here BTW.)

brunoabinader: Yes, I guess I didn't account for overscaled zoom on my tests. We can revisit it again with proper unit tests to verify later.

brunoabinader:
Took liberty to include a few other refactors and bug fixes related. Rebased after #6468 (/cc @jfirebaugh).

src/mbgl/layout/symbol_layout.cpp
```diff
@@ -418,8 +418,8 @@ std::unique_ptr<SymbolBucket> SymbolLayout::place(CollisionTile& collisionTile)
 
 template <typename Buffer>
 void SymbolLayout::addSymbols(Buffer &buffer, const SymbolQuads &symbols, float scale, const bool keepUpright, const style::SymbolPlacementType placement, const float placementAngle) {
-
-    const float placementZoom = ::fmax(std::log(scale) / std::log(2) + zoom, 0);
+    static const uint16_t vertexLength = 4;
+    const float placementZoom = util::max(std::log2(scale) + zoom, 0.0f);
```
kkaefer: Not using `util::log2`

src/mbgl/layout/symbol_layout.cpp
```diff
@@ -428,15 +428,14 @@ void SymbolLayout::addSymbols(Buffer &buffer, const SymbolQuads &symbols, float
...
-        float maxZoom = util::min(static_cast<float>(zoom + log(symbol.maxScale) / log(2)), 25.0f);
+        float minZoom = util::max(zoom + util::log2(symbol.minScale), placementZoom);
+        float maxZoom = util::min(zoom + util::log2(symbol.maxScale), util::MAX_ZOOM_F);
         const auto &anchorPoint = symbol.anchorPoint;
 
         // drop upside down versions of glyphs
         const float a = std::fmod(symbol.anchorAngle + placementAngle + M_PI, M_PI * 2);
         if (keepUpright && placement == style::SymbolPlacementType::Line &&
-            (a <= M_PI / 2 || a > M_PI * 3 / 2)) {
+            (a <= M_PI_2 || a > M_PI_2 * 3)) {
```
kkaefer: Not sure this makes sense for readability. The compiler almost certainly optimizes this anyway.

src/mbgl/math/log2.cpp
```diff
@@ -21,15 +21,9 @@ uint32_t ceil_log2(uint64_t x) {
...
-// log2() is producing wrong results on ARMv5 binaries
-// running on ARMv7+ CPUs.
-#if defined(__ANDROID__)
-    return std::log(x) / 0.6931471805599453; // log(x) / log(2)
-#else
-    return ::log2(x);
-#endif
-}
+template <> float log2(float x);
+template <> double log2(double x);
```
kkaefer: Why are these needed? The implementation is all in the header anyway.

src/mbgl/layout/symbol_layout.cpp
```diff
@@ -449,16 +448,14 @@ void SymbolLayout::addSymbols(Buffer &buffer, const SymbolQuads &symbols, float
...
-        if (buffer.segments.empty() || buffer.segments.back().vertexLength + glyph_vertex_length > 65535) {
+        if (buffer.segments.back().vertexLength + vertexLength > std::numeric_limits<uint16_t>::max()) {
             buffer.segments.emplace_back(buffer.vertices.size(), buffer.triangles.size());
         }
 
         // We're generating triangle fans, so we always start with the first
         // coordinate in this polygon.
         auto& segment = buffer.segments.back();
-        size_t index = segment.vertexLength;
+        uint16_t index = uint16_t(segment.vertexLength);
```
kkaefer: Maybe assert here that it's `<= 65535`?

src/mbgl/layout/symbol_layout.cpp
```diff
@@ -418,6 +418,7 @@ std::unique_ptr<SymbolBucket> SymbolLayout::place(CollisionTile& collisionTile)
 
 template <typename Buffer>
 void SymbolLayout::addSymbols(Buffer &buffer, const SymbolQuads &symbols, float scale, const bool keepUpright, const style::SymbolPlacementType placement, const float placementAngle) {
+    static const uint16_t vertexLength = 4;
```
kkaefer: Can be `constexpr`

src/mbgl/layout/symbol_layout.cpp
brunoabinader: Oops - forgot to update these, thanks!

src/mbgl/layout/symbol_layout.cpp
brunoabinader: Fair enough 👍 From reading http://stackoverflow.com/posts/1610346/revisions it seems that using `M_PI_2` actually decreases a bit of precision in favor of pre computation.

src/mbgl/math/log2.cpp
brunoabinader: This accounts for [explicit instantiation](http://en.cppreference.com/w/cpp/language/template_specialization), which from my understanding _instantiates_ these template specializations to be reused by each compilation unit without duplications.

src/mbgl/layout/symbol_layout.cpp
brunoabinader: Sure - btw this accounts for the longest type declaration in our codebase: `static constexpr const uint16_t` 😄

src/mbgl/math/log2.cpp
kkaefer: Yeah, but they are only needed when you want to force certain instances to exist even if they're not used, or when the template definition is in an implementation file.

jfirebaugh:
Can you please revert the `GL_LINEAR` change? Enabling `GL_LINEAR` during pan-only operations was the intended behavior -- it prevents icons from "wiggling" during the pan.

PR #6795, opened 2016-10-22
Adopt string enumerations

### LINKED ISSUE #6791: Adopt string enumerations
We have various sets of string constants that could be better organized with `NS_STRING_ENUM` or `NS_EXTENSIBLE_STRING_ENUM`. Constants like `MGLGeoJSONClusterOption` would become instances of `MGLGeoJSONSourceOption`, which would allow Swift code to say `MGLGeoJSONSource.Option.cluster`. We’ll need to use conditionally compiled `#define`s to avoid breaking projects compiled using Xcode 7.x.

/cc @friedbunny @frederoni

1ec5:
MGLGeoJSONSource options, notification names and user info keys posted by MGLOfflineStorage, and MGLErrorDomain are instances of string enumerations for ease of use in Swift. `MGLGeoJSONSourceOption` and `MGLOfflinePackUserInfoKey` constants have been renamed for clarity. (The old `MGLOfflinePackUserInfoKey` constants are deprecated.) Added some shims for projects built with Xcode 7.x.

Fixes #6791.

/cc @friedbunny @frederoni


platform/darwin/src/MGLGeoJSONSource.h
```diff
@@ -7,48 +7,53 @@ NS_ASSUME_NONNULL_BEGIN
...
+ */
+typedef NSString *MGLGeoJSONSourceOption NS_STRING_ENUM;
+
+/**
  An `NSNumber` object containing a Boolean enabling or disabling clustering.
  If the `features` property contains point features, setting this option to
  `YES` clusters the points by radius into groups. The default value is `NO`.
  */
-extern NSString * const MGLGeoJSONClusterOption;
+extern MGLGeoJSONSourceOption const MGLGeoJSONClusterOption;
```
boundsj: `MGLGeoJSONSourceOption` combined with `MGLGeoJSON{Foo}Option` will make it look like `MGLGeoJSONSourceOption.bufferOption`. Can we use the root `MGLGeoJSONSourceOption` plus noun (i.e. Cluster, ClusterRadius, etc.) and "limit our options"? 😄

platform/darwin/src/MGLOfflineStorage.h
```diff
@@ -24,7 +24,7 @@ NS_ASSUME_NONNULL_BEGIN
  alternatively observe KVO change notifications to the pack’s `progress` key
  path.
  */
-extern NSString * const MGLOfflinePackProgressChangedNotification;
+extern NSNotificationName const MGLOfflinePackProgressChangedNotification;
```
boundsj: Same comment here. MGLOfflinePackProgressChangedNotification namespaces it so I don't think we need the word `Notification` in the name. Also, along the lines of `swift_name`, does this change things enough for some bleeding edge users that we would want to hold off until a larger semver change?

platform/darwin/src/MGLGeoJSONSource.h
1ec5: Good point. Looks like the Swift compiler can trim common suffixes from Objective-C enumerations and `NS_OPTION` fields but not string enumerations. We’ll have to use `NS_SWIFT_NAME`.

platform/darwin/src/MGLOfflineStorage.h
1ec5: I stealthily edited [our versioning policy](https://github.com/mapbox/mapbox-gl-native/wiki/Versions-and-tagging#mapbox-ios-sdk) to carve out an exemption for Swift name refinement. The idea is that, with each major Swift revision, developers on the latest Swift will have to put up with a short period of API instability as we get our act together. It’s sort of like how you can upgrade to the latest iOS release the day it comes out, but application updates trickle in for weeks after that. Maybe that’s wrongheaded, but I see no other way we can keep up with MapKit, which has a latest-Swift-compatible release available the day the next version of Swift is officially released.

We can definitely target these changes to only affect Swift 3 code, by the way.

boundsj: 👍  sounds great. I missed the note about Swift in our versioning policy.

1ec5: Swift 3 bridges this symbol as `NSNotification.Name.MGLOfflinePackProgressChanged`, which is consistent with how AppKit and UIKit string enumeration values are being bridged.

platform/darwin/src/MGLGeoJSONSource.h
1ec5: f2bfc2c95d76a1066552220d4658cebb111bca97 renames these constants. Since they’re new to v3.4.0, I didn’t bother to keep the old names around.

PR #6770, opened 2016-10-20
[core] Reload geojson source on url change

ivovandongen:
Fixes #6757 

Review @tmpsantos  / @brunoabinader ?

PR #6798, opened 2016-10-23
[core] geojson source - reset req on setURL

ivovandongen:
Fixes https://github.com/mapbox/mapbox-gl-native/pull/6770#discussion_r84544664

@jfirebaugh added a check on an existing request in `setURL` as well so that any pending request is canceled when changing the url, not only when it was previously done loading.

### DIFF
--- a/src/mbgl/style/sources/geojson_source_impl.cpp
+++ b/src/mbgl/style/sources/geojson_source_impl.cpp
@@ -40,8 +40,9 @@ void GeoJSONSource::Impl::setURL(std::string url_) {
     url = std::move(url_);
 
     //Signal that the source description needs a reload
-    if (loaded) {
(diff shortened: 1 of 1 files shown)


ivovandongen:
@jfirebaugh We might want to do the same on [setGeoJSON](https://github.com/mapbox/mapbox-gl-native/blob/ad44188bd2173a787d0970ace92905397128d1d7/src/mbgl/style/sources/geojson_source_impl.cpp#L54) when called explicitly by the user (as opposed to the request from `loadDescription` finishing) so that the content is not overwritten by completion of a request in progress or a refresh on caching headers.

jfirebaugh:
@ivovandongen Good call, you're absolutely right. Feel free to either make that change here or in a separate PR.

PR #6806, opened 2016-10-24
[node] add test illustrating 256px raster sources missing tileSize

mikemorris:
Using a 256px raster source without setting `tileSize: 256` will request and rescale tiles from the zoom level to be rendered instead of requesting tiles from the next zoom level and rendering them as a mosaic. 

This is expected behavior, and this test will simply serve as an example.

/cc @bsudekum @miccolis


mikemorris:
This would be better handled as a test in mapbox-gl-test-suite, closing.

PR #6819, opened 2016-10-25
[ios] Add country label language toggle demo

friedbunny:
Adds the option to toggle between Chinese and English country labels in ios-app.

/cc @boundsj @pveugen @1ec5

### DIFF
--- a/platform/ios/app/MBXViewController.m
+++ b/platform/ios/app/MBXViewController.m
@@ -140,12 +140,7 @@ typedef NS_ENUM(NSInteger, MBXSettingsMiscellaneousRows) {
     [self restoreState:nil];
 
     self.debugLoggingEnabled = [[NSUserDefaults standardUserDefaults] boolForKey:@"MGLMapboxMetricsDebugLoggingEnabled"];
-}
 
-- (void)viewDidAppear:(BOOL)animated
-{
-    [super viewDidAppear:animated];
-    
     if ([MGLAccountManager accessToken].length)
     {
         self.styleIndex = -1;
--- a/platform/ios/app/MBXViewController.m
+++ b/platform/ios/app/MBXViewController.m
@@ -70,6 +70,7 @@ typedef NS_ENUM(NSInteger, MBXSettingsRuntimeStylingRows) {
     MBXSettingsRuntimeStylingUpdateGeoJSONSourceFeatures,
     MBXSettingsRuntimeStylingVectorSource,
     MBXSettingsRuntimeStylingRasterSource,
+    MBXSettingsRuntimeStylingCountryLabels,
 };
 
 typedef NS_ENUM(NSInteger, MBXSettingsMiscellaneousRows) {
@@ -105,6 +106,7 @@ typedef NS_ENUM(NSInteger, MBXSettingsMiscellaneousRows) {
 @property (nonatomic) NSInteger styleIndex;
 @property (nonatomic) BOOL debugLoggingEnabled;
 @property (nonatomic) BOOL customUserLocationAnnnotationEnabled;
+@property (nonatomic) BOOL usingLocaleBasedCountryLabels;
 
 @end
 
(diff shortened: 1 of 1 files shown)


1ec5:
Better yet, let’s turn this into a switch between the local language and the system’s current language, similar to the one that Apple Maps has. (If the system language is set to something that isn’t [included in Mapbox Streets vector tiles](https://www.mapbox.com/vector-tiles/mapbox-streets-v7/#overview), fall back to English.)

friedbunny:
Changed in 0e51b729a9c7b60903ccf6e5fb8c294cb208e273 to grab to the user’s most-favored language (as defined by the device’s locale settings) and then toggle between that and `{name}`.

- If the user’s language is `en`, it will initially toggle to `{name}`.
- If the user’s language is `zh`, it will initially toggle to `{name_zh}`.

friedbunny:
Rebased this onto the newly-synced-with-master release branch.

platform/ios/app/MBXViewController.m
```diff
@@ -908,6 +912,33 @@ - (void)styleDynamicPointCollection
...
+}
+
+- (NSString *)bestLanguageForUser
+{
+    NSArray *supportedLanguages = @[ @"en", @"es", @"fr", @"de", @"ru", @"zh" ];
+    NSArray<NSString *> *preferredLanguages = [NSLocale preferredLanguages];
+    NSString *bestLanguage;
+
+    for (NSString *language in preferredLanguages) {
+        NSString *thisLanguage = [language substringToIndex:2];
```
1ec5: What if the user’s preferred language is set to Aromanian (`rup`)? 😉 FYI, this would be more robust, but what you’ve got is fine for a demo:

```objc
[NSLocale localeWithIdentifier:language].languageCode
```

1ec5: Yeah, not sure if Aromanian speakers would know what to do with a Russian map.

platform/ios/app/MBXViewController.m
```diff
@@ -310,6 +311,7 @@ - (void)dismissSettings:(__unused id)sender
                 @"Style Query For GeoJSON",
                 @"Style Feature",
                 @"Style Dynamic Point Collection",
+                @"Style Country Label Language"
```
friedbunny: Got this in 89bc8efc1564f61a7741e2c55383c1ca4d5ac1c9.

platform/ios/app/MBXViewController.m
```diff
@@ -908,6 +912,33 @@ - (void)styleDynamicPointCollection
     [self.mapView.style addLayer:layer];
 }
 
+- (void)styleCountryLabelLanguage
+{
+    NSString *bestLanguageForUser = [NSString stringWithFormat:@"{name_%@}", [self bestLanguageForUser]];
+    MGLSymbolStyleLayer *countryLayer = (MGLSymbolStyleLayer *)[self.mapView.style layerWithIdentifier:@"country-label-lg"];
```
friedbunny: WIP in e9b24c1. Doesn’t support functions yet. Sets the bool every iteration, which could lead to inconsistency and should be unnecessary.

friedbunny:
Most of the way there. Still need to implement function-based text fields and find a better way to set `_usingLocaleBasedCountryLabels`.

friedbunny:
Implemented zoom functions. This is ready for review again.

platform/ios/app/MBXViewController.m
```diff
@@ -1049,6 +1054,67 @@ - (void)styleRasterSource
...
+        @"country-label-lg",
+        @"country-label-md",
+        @"country-label-sm",
+    ];
+    [self styleLabelLanguageForLayersNamed:labelLayers];
+}
+
+- (void)styleLabelLanguageForLayersNamed:(NSArray<NSString *> *)layers
+{
+    _usingLocaleBasedCountryLabels = !_usingLocaleBasedCountryLabels;
```
1ec5: FYI, this flag will get out of sync as soon as the user switches styles.

friedbunny: True... 😒

PR #6826, opened 2016-10-26
Use animated rotation value for MarkerView.setRotate

### LINKED ISSUE #6816: Marker view rotation not working correctly
Marker view rotation is off by ~90 degrees from the expected value. Reproducible in the [`OffRouteActivity`](https://github.com/mapbox/mapbox-android-demo/blob/master/MapboxAndroidDemo/src/main/java/com/mapbox/mapboxandroiddemo/labs/OffRouteActivity.java) in the demo app

cc: @mapbox/android

tobrun:
closes #6816, targeting `release-android-v4.2.0`. This PR cancels ongoing rotation animations and uses the current animated value instead of using the previous set rotation. If you would call `MarkerView#setRotate` while an animation was ongoing you would end up with a wrong rotation value. 

Review @ivovandongen


cammace:
Tested in Demo app, looks good now:

ezgif com-video-to-gif 5

Issue #6796, opened 2016-10-23
Still mode should clear the rendered result upon recall


brunoabinader:
I believe this is what causes the duplicated painting in `regressions/mapbox-gl-js#3365`:

| actual | expected |
| :-: | :-: |
| ![actual](https://cloud.githubusercontent.com/assets/76133/19623674/d3c9c7d4-98db-11e6-85af-9004c0b64a83.png) | ![expected](https://cloud.githubusercontent.com/assets/76133/19623675/d6220f64-98db-11e6-8897-f942ca08a67a.png) |

jfirebaugh:
This test case is designed to test that text collisions are recalculated when rotating the map (a.k.a. `redoPlacement`). The "actual" result looks like what happens if that doesn't happen.

PR #6828, opened 2016-10-26
Line-break ideographic text by character

### LINKED ISSUE #1223: Chinese labels should be wrapped by character with balanced lines
the font is "Unifont Medium” .The english is ok

1ec5:
Automatically insert a line break after any supported Chinese, Japanese, or Yi character in a point-placed label as needed to stay within the `text-max-width`. Balance the lines unless non-ideographic text such as Latin letters are present.

golf
_Compare the label on the map to the popover title, which shows the queried feature’s name. A line break is absent from the feature but applied automatically by the renderer._

Ported from mapbox/mapbox-gl-js#3420. Fixes #1223. Depends on mapbox/mapbox-gl-test-suite#153.

Before merging:

* [x] Revert mapbox/mapbox-gl-test-suite@3b005f1318fe4530f75caa3516b6d2b2a8014e37
* [ ] Update gl-test-suite version pin once [1ec5-換行-1223](https://github.com/mapbox/mapbox-gl-test-suite/tree/1ec5-%E6%8F%9B%E8%A1%8C-1223) gl-test-suite branch lands

/cc @lucaswoj @xrwang

### DIFF
--- a/src/mbgl/text/glyph_set.cpp
+++ b/src/mbgl/text/glyph_set.cpp
@@ -128,8 +128,7 @@ void GlyphSet::lineWrap(Shaping &shaping, const float lineHeight, float maxWidth
                     // Collapse invisible characters.
                     uint32_t breakGlyph = positionedGlyphs[lastSafeBreak].glyph;
                     uint32_t lineEnd = lastSafeBreak;
-                    if (breakGlyph == 0x20 /* space */
-                        || breakGlyph == 0x200b /* zero-width space */) {
+                    if (util::i18n::isVisible(breakGlyph)) {
                         lineEnd--;
                     }
 
@@ -144,16 +143,7 @@ void GlyphSet::lineWrap(Shaping &shaping, const float lineHeight, float maxWidth
 
             // Ideographic characters, spaces, and word-breaking punctuation that often appear without surrounding spaces.
             if (useBalancedIdeographicBreaking
-                || shape.glyph == 0x20 /* space */
-                || shape.glyph == 0x26 /* ampersand */
-                || shape.glyph == 0x2b /* plus sign */
-                || shape.glyph == 0x2d /* hyphen-minus */
-                || shape.glyph == 0x2f /* solidus */
-                || shape.glyph == 0xad /* soft hyphen */
-                || shape.glyph == 0xb7 /* middle dot */
-                || shape.glyph == 0x200b /* zero-width space */
-                || shape.glyph == 0x2010 /* hyphen */
-                || shape.glyph == 0x2013 /* en dash */
+                || util::i18n::allowsWordBreaking(shape.glyph)
                 || util::i18n::allowsIdeographicBreaking(shape.glyph)) {
                 lastSafeBreak = i;
             }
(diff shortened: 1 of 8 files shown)


src/mbgl/util/i18n.hpp
```diff
@@ -0,0 +1,18 @@
+#pragma once
+
+#include <string>
+
+namespace mbgl {
+namespace util {
+namespace i18n {
+
+/// Returns whether a line break can be inserted after any character in the given string.
+/// If false, line breaking should occur on word boundaries instead.
```
kkaefer: We're not using triple slashes in core

src/mbgl/text/glyph_set.cpp
```diff
@@ -143,7 +153,8 @@ void GlyphSet::lineWrap(Shaping &shaping, const float lineHeight, const float ma
                 || shape.glyph == 0xb7 /* middle dot */
                 || shape.glyph == 0x200b /* zero-width space */
                 || shape.glyph == 0x2010 /* hyphen */
-                || shape.glyph == 0x2013 /* en dash */) {
+                || shape.glyph == 0x2013 /* en dash */
```
kkaefer: We could move all of those as well to a sister function of `allowsIdeographicBreaking` in the `i18n` namespace

1ec5:
The failing test is:

* failed regressions [mapbox-gl-js#3365](https://github.com/mapbox/mapbox-gl-js/issues/3365)

brunoabinader:
Worth noticing I am temporarily ignoring `text-max-width/ideographic-breaking` for GL native as part of updating mesa to use OSMesa: https://github.com/mapbox/mapbox-gl-test-suite/commit/51546ca001ee22acf8e11b771184e8b268c43be0. We can undo it once this PR gets merged.

jfirebaugh:
The test-suite failure was prexisting. It's tracked in https://github.com/mapbox/mapbox-gl-native/issues/6796 and the test case is now ignored for native. If you rebase this PR, I expect test-suite will pass.

Was that the only remaining obstacle to merging this?

1ec5:
> Was that the only remaining obstacle to merging this?

I also need to address @kkaefer’s feedback, but that should be it.

kkaefer:
Are there any tests that specifically test breaking by character?

1ec5:
There are rendering tests in the test suite for the overall ideographic breaking feature, but no comprehensive test that goes character by character. The rendering tests were disabled on the native side in mapbox/mapbox-gl-test-suite@3b005f1318fe4530f75caa3516b6d2b2a8014e37, but I verified that they passed before pinning to that commit.

kkaefer:
Can we enable these tests when merging?

1ec5:
Yes, it's on the to-do list above.

1ec5:
Putting this on the Android SDK v4.2.0 milestone because it blocks #6996.

kkaefer:
@1ec5 any reason why we're not reenabling the tests as part of this PR rather than doing to after merge?

1ec5:
> any reason why we're not reenabling the tests as part of this PR rather than doing to after merge?

OK, I’ve pointed this PR to the [1ec5-換行-1223](https://github.com/mapbox/mapbox-gl-test-suite/tree/1ec5-%E6%8F%9B%E8%A1%8C-1223) branch in gl-test-suite, in which I’ve reenabled the tests. Once I get the green light, I’ll fast-forward-merge that branch, then update the commit hash in package.json, then merge this PR.

PR #6390, opened 2016-09-20
[android] - Update to Android Studio 2.2.2 with tools/dependencies

tobrun:
Android Studio 2.2 has been officially released in the stable channel:
This as a trigger to update other dependencies/tools in the project as upgrading the target SDK version to Nougat (officially released a month a go and now running on multiple devices in the wild). This change results in updating the support lib versions and android gradle tools version. Also updated the test app version number to reflect the 4.2.0 release.

@mapbox/android any objections?

PR #6830, opened 2016-10-26
Android Nougat compatibility: onstart/onstop on MapView

### LINKED ISSUE #5910: Move listener logic to onStart/onStop
Android N introduces multiwindow support and to get the correct behaviour we need to move our listener logic from `onResume/onPause` to `onStart/onStop` instead. This currently affects the `MyLocationView` since switching focus between different windows invokes the `onResume/onPause` lifecycle events. If we don't move the listener logic, the `MyLocationView` will stop updating the moment the window loses focus. 

More information in this [Google I/O session](https://www.youtube.com/watch?v=yEEy_48hoXI).

tobrun:
Closes #5910.

This PR makes the SDK compatible with Android Nougat by moving the location listener logic to the  Activity onStart/onStop instead of onResume/onPause. This change is needed to support multiwindow feature introduced in Android nougat.

To make migration easier for users, I opted to implement an exception that is thrown when the newly introduced MapView#onStart or MapView#onStop isn't called. It's similar to SuperNotCalledException from the Android SDK. This will enforce Users to migrate to the new methods. 

Review @ivovandongen

### DIFF
--- a/platform/android/MapboxGLAndroidSDK/src/main/java/com/mapbox/mapboxsdk/maps/MapView.java
+++ b/platform/android/MapboxGLAndroidSDK/src/main/java/com/mapbox/mapboxsdk/maps/MapView.java
@@ -160,6 +160,9 @@ public class MapView extends FrameLayout {
     private List<OnMapReadyCallback> onMapReadyCallbackList;
     private SnapshotRequest snapshotRequest;
 
+    private boolean onStartCalled;
+    private boolean onStopCalled;
+
     @UiThread
     public MapView(@NonNull Context context) {
         super(context);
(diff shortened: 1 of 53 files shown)


tobrun:
Generated the activity sanity tests locally to validate changes:

screen shot 2016-10-26 at 15 18 21

tobrun:
Target SDK set to Nougat has landed on master with #6390, this PR is ready for review.

tobrun:
@zugaldia updated `changelog.md` as requested.

zugaldia:
@tobrun Thanks. Note that I haven't reviewed the rest yet, I'll let @ivovandongen do it per OP.

platform/android/MapboxGLAndroidSDK/src/main/java/com/mapbox/mapboxsdk/maps/MapView.java
```diff
@@ -573,50 +576,80 @@ public void onSaveInstanceState(@NonNull Bundle outState) {
...
+            // TODO: 26/10/16, can be removed after 5.0.0 release
+            throw new IllegalStateException("MapView#onStart() was not called. " +
+                    "You must call this method from the parent's {@link Activity#onStart()} or {@link Fragment#onStart()}.");
+        }
     }
 
     /**
      * You must call this method from the parent's {@link Activity#onPause()} or {@link Fragment#onPause()}.
      */
     @UiThread
```
ivovandongen: Since we're doing a semver release anyway. Can't we remove this all together? Or at least deprecate it?

tobrun: I implemented this way to make sure users are correctly migrating from an older version. If a user doesn't implement the new onStart/onStop, they will not be able to receive any location updates. This is something that will go unnoticed during "migration" and will only show when you are trying to a location fix. I have no problem with removing this but I felt it was an improvement over the other possible UX here.

ivovandongen: 👍

Issue #6670, opened 2016-10-12
Annotation symbols are overdrawn if located on tile edges


brunoabinader:
Spin-off from the discussions around stencil clip usage from #3563.

Fixing the position of repeated features near tile edges caused a new issue to rise: opaque symbols are now overdrawn:
<img width="1136" alt="screen shot 2016-10-09 at 9 20 47 pm" src="https://cloud.githubusercontent.com/assets/76133/19222648/720a04e8-8e66-11e6-9f15-7df9440775cc.png">

brunoabinader:
As [previously mentioned in #3563](https://github.com/mapbox/mapbox-gl-native/pull/3563#issuecomment-252650498), fixing this issue requires either removing stencil clip usage for symbols (option 1), or enforcing stencil clip for all symbol buckets (option 2).

__Option 1__ is tricky because it contains some corner cases in which I have no proper fix in mind:
- Style property [symbol-avoid-edges](https://www.mapbox.com/mapbox-gl-style-spec/#layout-symbol-symbol-avoid-edges) depends on stencil clip functionality
- Requires enforcing non-duplicated entries when _updating tile layers_, _layout_ and _querying_ features:
 - 1.1) When updating annotation tile layers, we check if each [symbol intersects with them](https://github.com/mapbox/mapbox-gl-native/blob/master/src/mbgl/annotation/annotation_manager.cpp#L164-L167). We can enforce them to be unique by e.g. letting only the left-topmost tile to get the feature. That's not something trivally possible with regular symbol layers because we already [receive them duplicated from the server](https://github.com/mapbox/vector-tile-spec/tree/master/2.1):
> For example, if a tile has an extent of 4096, coordinate units within the tile refer to 1/4096th of its square dimensions. A coordinate of 0 is on the top or left edge of the tile, and a coordinate of 4096 is on the bottom or right edge. Coordinates from 1 through 4095 inclusive are fully within the extent of the tile,

 - 1.2) When inserting symbol geometries in layout step, we could [filter symbol instances based on their ID](https://github.com/mapbox/mapbox-gl-native/blob/master/src/mbgl/layout/symbol_layout.cpp#L496). That involves [forwarding the feature ID](https://github.com/mapbox/mapbox-gl-native/blob/master/src/mbgl/layout/symbol_layout.cpp#L229) when populating the symbol instances container.
- 1.3) When querying, we can easily filter out duplicated features based on their ID.

[...]

PR #6832, opened 2016-10-26
[core] Avoid clipping symbols in continuous mode

### LINKED ISSUE #1673: add padding to annotation tiles

brunoabinader:
~~Re-enable symbol clipping based on layout property settings, cleaning up workarounds and adds padding to symbol annotation tiles.~~

Avoid clipping symbols in continuous mode - preserving clipping behavior in still mode. To avoid overdrawing symbols on tile edges, we avoid inserting symbol features with anchors located among the tile extent edges.

Fixes #1673 and #6670.

/cc @jfirebaugh @ansis

### DIFF
--- a/src/mbgl/layout/symbol_layout.cpp
+++ b/src/mbgl/layout/symbol_layout.cpp
@@ -262,8 +262,6 @@ void SymbolLayout::addFeature(const GeometryCollection &lines,
     const SymbolPlacementType iconPlacement = layout.get<IconRotationAlignment>() != AlignmentType::Map
                                                   ? SymbolPlacementType::Point
                                                   : layout.get<SymbolPlacement>();
-    const bool mayOverlap = layout.get<TextAllowOverlap>() || layout.get<IconAllowOverlap>() ||
-        layout.get<TextIgnorePlacement>() || layout.get<IconIgnorePlacement>();
     const bool isLine = layout.get<SymbolPlacement>() == SymbolPlacementType::Line;
     const float textRepeatDistance = symbolSpacing / 2;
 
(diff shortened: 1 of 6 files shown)


jfirebaugh:
Can you comment on https://github.com/mapbox/mapbox-gl-native/issues/6670#issuecomment-254331003 and help fill in the gaps in my understanding?

brunoabinader:
> Can you comment on #6670 (comment) and help fill in the gaps in my understanding?

I've added an extensive introspection as a [comment in #6670](https://github.com/mapbox/mapbox-gl-native/issues/6670#issuecomment-256524251).

src/mbgl/layout/symbol_layout.cpp
```diff
@@ -298,8 +297,7 @@ void SymbolLayout::addFeature(const GeometryCollection &lines,
             // be drawn across tile boundaries. Instead they need to be included in
             // the buffers for both tiles and clipped to tile boundaries at draw time.
             //
-            // TODO remove the `&& false` when is #1673 implemented
-            const bool addToBuffers = (mode == MapMode::Still) || inside || (mayOverlap && false);
+            const bool addToBuffers = inside || mayOverlap;
```
jfirebaugh: I don't see how removing `mode == MapMode::Still` here is safe. In still mode, rendering a single tile, we _cannot_ throw away features in the tile buffer under any circumstances, for the reasons detailed in https://github.com/mapbox/mapbox-gl-native/issues/6670#issuecomment-254331003.

brunoabinader: You are right - for Still mode we need to render _all_ features.

brunoabinader:
Per https://github.com/mapbox/mapbox-gl-native/issues/6670#issuecomment-257227118 - it seems we no longer need padding for symbol annotations - we can sort symbol render tiles instead prior to rendering.

We still clip in still mode, but just because we need to update our render test results. I'm deferring this work to a following PR.

brunoabinader:
Based on @jfirebaugh's [comments](https://github.com/mapbox/mapbox-gl-native/issues/6670#issuecomment-259825324), I'm going to revert a few comments about clipping symbols for Still mode I've had removed on this PR. Will flag it reviewable again once it is finished.

brunoabinader:
Ready for review.

brunoabinader:
@jfirebaugh @ansis I have cleaned up the patches from this PR, removing stuff that was not related to the proposed changes:

- Avoid clipping symbols in continuous mode - preserving clipping behavior in still mode. We alleviate the rendering artifacts mentioned in https://github.com/mapbox/mapbox-gl-native/issues/1673 by sorting render tiles in reverse Y prior to rendering.
- To avoid overdrawing symbols on tile edges, we avoid inserting symbol features with anchors located among the tile extent edges - fixes https://github.com/mapbox/mapbox-gl-native/issues/6670.

PR #6833, opened 2016-10-26
document custom API URL & add Info.plist guide

incanus:
Updated nullability and documented `MGLMapboxAPIBaseURL` key in `Info.plist`. 

/cc @1ec5 @friedbunny @jakepruitt


1ec5:
This documentation won't be visible to the public. While you're at it, can you document our Info.plist keys in a jazzy guide? See realm/jazzy#523.

incanus:
> This documentation won't be visible to the public.

Right, that's fine. This is under-the-radar enough that folks can pick it up from the changelog, if at all. The docs are more for us. 

> can you document our Info.plist keys in a jazzy guide? 

Not sure I follow. Where would this show up?

1ec5:
A guide would be a separate document, listed in the jazzy docs' table of contents. Like the readme, it would be written in Markdown.

incanus:
Ok, got the guides in 09649646d4b2152803a6cc638fe54aaf1918c836. Tried a combined `darwin` file at first, but macOS doesn't have the opt-out key so had to do separate ones. The section name is tied to the filename, hence the space. This looks pretty good. I don't think we need to mention the Core Location keys around location permissions. 

screen shot 2016-10-26 at 11 30 15 am

platform/ios/Info.plist Keys.md
```diff
@@ -0,0 +1,15 @@
+# Info.plist Keys
+
+The SDK supports several custom `Info.plist` keys in your application in order to configure various settings. 
+
+## MGLMapboxAccessToken
+
+Use this key to set your Mapbox API [access token](https://www.mapbox.com/help/define-access-token/). This key is required for all map use.
```
1ec5: This would be a good place to consolidate the explanation that’s currently in [`+[MGLAccessToken setAccessToken:]`](https://github.com/mapbox/mapbox-gl-native/blob/e7bfa7849e4ed7a80d27b63ac2f504660ecfb14b/platform/darwin/src/MGLAccountManager.h#L14-L23) and [MGLMapView](https://github.com/mapbox/mapbox-gl-native/blob/e7bfa7849e4ed7a80d27b63ac2f504660ecfb14b/platform/ios/src/MGLMapView.h#L69-L74). Actually, since a link back to this document won’t work in Quick Help, duplicating that explanation would be preferable.

platform/ios/Info.plist Keys.md
```diff
@@ -0,0 +1,15 @@
...
+
+The SDK supports several custom `Info.plist` keys in your application in order to configure various settings. 
+
+## MGLMapboxAccessToken
+
+Use this key to set your Mapbox API [access token](https://www.mapbox.com/help/define-access-token/). This key is required for all map use. 
+
+## MGLMapboxAPIBaseURL
+
+Use this key if you need to customize the API base URL used throughout the SDK. If unset, the default Mapbox API is used.
```
1ec5: An example would be nice (even the default value). Otherwise, the developer has no idea whether to include the scheme, a trailing slash, etc.

platform/ios/Info.plist Keys.md
1ec5: Also link to `+[MGLAccountManager setAccessToken:]` as an alternative for developers who prefer to set their access token in code, as well as [this guide](https://www.mapbox.com/help/ios-private-access-token/) for those who want to keep their access token super private.

incanus: 1650526

incanus: 1650526

platform/ios/CHANGELOG.md
```diff
@@ -62,6 +62,7 @@ Mapbox welcomes participation and contributions from everyone. Please read [CONT
 * Fixed a crash that occurred when encountering a rate-limit error in response to a network request. ([#6223](https://github.com/mapbox/mapbox-gl-native/pull/6223))
 * Fixed an issue causing an MGLOfflinePack’s progress to continue to update after calling `-suspend`. ([#6186](https://github.com/mapbox/mapbox-gl-native/pull/6186))
 * Fixed an issue preventing cached annotation images from displaying while the device is offline. ([#6358](https://github.com/mapbox/mapbox-gl-native/pull/6358))
+* Added support for an `MGLMapboxAPIBaseURL` key  in an app's `Info.plist` in order to customize the base URL used for retrieving map data, styles, and other resources. ([#6709](https://github.com/mapbox/mapbox-gl-native/pull/6709))
```
friedbunny: Extra space: `key  in`

platform/ios/Info.plist Keys.md
```diff
@@ -0,0 +1,21 @@
+# Info.plist Keys
+
+The Mapbox iOS SDK supports custom `Info.plist` keys in your application in order to configure various settings. 
+
+## MGLMapboxAccessToken
+
+Set the <a href="https://www.mapbox.com/help/define-access-token/">Mapbox access token</a> to be used by all instances of `MGLMapView` in the current application.
```
1ec5: Nit: use Markdown link syntax.

platform/ios/scripts/document.sh
```diff
@@ -38,6 +38,7 @@ jazzy \
     --github-file-prefix https://github.com/mapbox/mapbox-gl-native/tree/${BRANCH} \
     --module-version ${SHORT_VERSION} \
     --readme ${README} \
+    --documentation="platform/ios/Info.plist Keys.md" \
```
1ec5: Please move to platform/ios/docs/ to live alongside doc-README.md.

friedbunny:
Currently looks like:
screen shot 2016-10-26 at 3 25 26 pm

1ec5:
FYI, I'm OK with how this guide is listed at the end of the table of contents, but it might be possible to bring it to the top by specifying it in the custom table of contents in jazzy.yml.

incanus:
It should be, but I was also ok with where it is now.

platform/ios/Info.plist Keys.md
incanus: 1403123942d6ed8e5cee13ce0295f8e469c5f5b7

platform/ios/scripts/document.sh
incanus: 8aeac22de96ec8849ef9fc186b7dac0143616d43

platform/ios/CHANGELOG.md
incanus: 1c72e4ab1355a0f2fe3497a9354aff78d3a03b2b

incanus:
Pending tests, this should be all good to go again.

incanus:
Cherry-picked to 3.4.0 release branch as 8ffd1f69252d28bf501d5464ac9e5f6ff8942c89.

PR #6674, opened 2016-10-26
Introduce Size type

kkaefer:
We frequently specify sizes that have a width and a height and should introduce a size type that encodes this instead of using `std::array<uint16_t, 2>` in those places.

/cc @jfirebaugh

### DIFF
--- a/include/mbgl/map/map.hpp
+++ b/include/mbgl/map/map.hpp
@@ -6,6 +6,7 @@
 #include <mbgl/util/geo.hpp>
 #include <mbgl/util/feature.hpp>
 #include <mbgl/util/noncopyable.hpp>
+#include <mbgl/util/size.hpp>
 #include <mbgl/annotation/annotation.hpp>
 #include <mbgl/style/transition_options.hpp>
 
@@ -33,7 +34,7 @@ class Layer;
 class Map : private util::noncopyable {
 public:
     explicit Map(Backend&,
-                 std::array<uint16_t, 2> size,
+                 Size size,
                  float pixelRatio,
                  FileSource&,
                  Scheduler&,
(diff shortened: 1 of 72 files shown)


include/mbgl/map/map.hpp
```diff
@@ -136,9 +137,9 @@ class Map : private util::noncopyable {
     ViewportMode getViewportMode() const;
 
     // Size
-    void setSize(const std::array<uint16_t, 2>&);
-    uint16_t getWidth() const;
-    uint16_t getHeight() const;
+    void setSize(Size);
+    uint32_t getWidth() const;
```
jfirebaugh: We should combine `getWidth()` and `getHeight()` into a single `getSize()`.

include/mbgl/sprite/sprite_image.hpp
```diff
@@ -22,8 +22,8 @@ class SpriteImage : private util::noncopyable {
     // Whether this image should be interpreted as a signed distance field icon.
     const bool sdf;
 
-    float getWidth() const { return image.width / pixelRatio; }
-    float getHeight() const { return image.height / pixelRatio; }
+    float getWidth() const { return image.size.width / pixelRatio; }
```
jfirebaugh: Same here.

platform/default/offscreen_view.cpp
```diff
@@ -21,17 +21,17 @@ void OffscreenView::bind() {
         context.bindFramebuffer = framebuffer->framebuffer;
     }
 
-    context.viewport = { 0, 0, size[0], size[1] };
+    context.viewport = { 0, 0, size.width, size.height };
```
jfirebaugh: Should be able to do `context.viewport = { 0, 0, size }`.

include/mbgl/sprite/sprite_image.hpp
kkaefer: In this particular case, we're using `float`s to represent the width/height, but our `Size` type uses integers, so we should keep this.

PR #6836, opened 2016-10-26
Refactor Atlases+FrameHistory to use wrapped texture objects

kkaefer:
Currently, the `*Atlas` objects, as well as the `FrameHistory` object are directly using OpenGL commands to manage their texture. We should change this to use the default-provided texture management logic in `gl::Context`.

A small roadblock is the fact that we currently only support RGBA textures, but the `GlyphAtlas`, `LineAtlas`, and `FrameHistory` use mono-channel (`GL_ALPHA`) textures.

Note that all of these are currently using `glTexSubImage2D` for subsequent uploads. However, internally most drivers will create a copy of the existing data and patch up that data, so there is very little advantage of subimage patching over full texture uploads.

### DIFF
--- a/include/mbgl/util/image.hpp
+++ b/include/mbgl/util/image.hpp
@@ -9,9 +9,10 @@
 
 namespace mbgl {
 
-enum ImageAlphaMode {
+enum class ImageAlphaMode {
     Unassociated,
-    Premultiplied
+    Premultiplied,
+    Exclusive, // Alpha-channel only
 };
 
 template <ImageAlphaMode Mode>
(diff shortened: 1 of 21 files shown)


src/mbgl/sprite/sprite_atlas.cpp
```diff
@@ -317,73 +317,34 @@ void SpriteAtlas::updateDirty() {
...
         dirtyFlag = false;
+    }
+
+    context.bindTexture(*texture, unit,
+                        linear ? gl::TextureFilter::Linear : gl::TextureFilter::Nearest);
 
 #if not MBGL_USE_GLES2
-        // platform::showColorDebugImage("Sprite Atlas", reinterpret_cast<const char*>(data.get()),
-        //                               pixelWidth, pixelHeight, pixelWidth, pixelHeight);
+//    if (wasDirty) {
```
kkaefer: Yeah; that code is there for debugging purposes with GLFW. If you need to visualize the glyph atlas, you can comment it out. That code used to be there before as well, I just rewrote it to match the new texture management.

include/mbgl/util/image.hpp
```diff
@@ -49,6 +49,10 @@ class Image : private util::noncopyable {
     size_t stride() const { return static_cast<size_t>(width) * 4; }
     size_t size() const { return static_cast<size_t>(width) * height * 4; }
 
+    operator bool() const {
```
kkaefer: Looks like the code isn't actually used.

kkaefer:
Ready for review

Issue #6173, opened 2016-08-26
MGLMapView calls MGLMapViewDelegate 'viewForAnnotation:' for annotations outside visible area


julaak:
**Platform: iOS**
**Mapbox SDK version: 3.3.4**
### Steps to trigger behavior
1. Build & run Mapbox iOS example app located at 'platform/ios/ios.xcodeproj'
2. Click 'gear' icon on top left corner to open Map Settings menu
3. Click 'Add 100 Views' to add annotations to map
4. Pan map so annotations are outside the visible map area.
5. Keep panning into areas with no annotations
### Expected behavior

No annotations are visible on the map or nowhere near the borders of visible area, MGLMapView should not(?) call method:
**'- (MGLAnnotationView *)mapView:(MGLMapView *)mapView viewForAnnotation:(id<MGLAnnotation>)annotation'**

Why? Because there's no annotations visible.
### Actual behavior

**'- (MGLAnnotationView *)mapView:(MGLMapView *)mapView viewForAnnotation:(id<MGLAnnotation>)annotation'** is called constantly while panning on the map while there are no annotations nowhere near the visible area.

Below is the callstack.

![screen shot 2016-08-26 at 16 59 07](https://cloud.githubusercontent.com/assets/18900605/18007887/7fdb04d0-6bae-11e6-8f0e-7e2ee55c0fd5.png)

In a nutshell 'MGLMapView notifyMapChange:' method is called with param MapChangeDidFinishRenderingFrameFullyRendered which triggers a update for all annotations(?)
### Why is this a problem?

The map becomes very unresponsive with large amount of MGLAnnotationView objects placed on the map. I feel like the annotation view reuse queue is not working like it is intended to work.

1ec5:
Related: https://github.com/mapbox/mapbox-gl-native/issues/5787#issuecomment-235052750

/cc @boundsj @incanus

boundsj:
This is fixed in https://github.com/mapbox/mapbox-gl-native/pull/5987. #5987 will be in the next release (beta 2).

PR #6840, opened 2016-10-27
[iOS][WIP] Improve annotationsView performances

naluhh:
This PR adresses the following issue:
- #6173 
# What changed ?

updateAnnotationViews was not working properly. The delegate method was called on each frame even if it wasn't in the view work. As a result, performance were worse when an annotation was outside the screen (!).
The reuse queue was also an NSArray instead of a NSSet resulting in poor performance when picking inside.

This PR improve the performance of the updateAnnotationViews especially when you are using complicated views.

One possible improvement to this would be to use queryPointAnnotations, but it's too slow when there is a lot of annotations on the screen.
# Testing done

I tested with thousand views and CAAnimation spread across the world with much better performances.

cc @1ec5

### DIFF
--- a/platform/ios/src/MGLAnnotationContainerView.m
+++ b/platform/ios/src/MGLAnnotationContainerView.m
@@ -31,8 +31,8 @@
     for (MGLAnnotationView *view in subviews)
     {
         [self addSubview:view];
-        [self.annotationViews addObject:view];
     }
+    [self.annotationViews addObjectsFromArray:subviews];
 }
 
 #pragma mark UIAccessibility methods
--- a/platform/ios/src/MGLAnnotationView.h
+++ b/platform/ios/src/MGLAnnotationView.h
@@ -83,6 +83,9 @@ typedef NS_ENUM(NSUInteger, MGLAnnotationViewDragState) {
 - (void)prepareForReuse;
 
 /**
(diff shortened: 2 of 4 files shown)


boundsj:
Thanks for the contribution @naluhh! I will review as part of https://github.com/mapbox/mapbox-gl-native/pull/5987 that I'm currently working on again now that #6055 is fixed.

boundsj:
Thanks again for your PR @naluhh. I went with the implementation in https://github.com/mapbox/mapbox-gl-native/pull/5987 because we wanted a solution that relies on the more efficient query mechanism in the mbgl (C++) core. Also, I wanted to stay away from implementations that rely on changing the visibility of the UIView objects. I was also concerned about using a set for the queue since we would not want to store only distinct views based on a hash -- we want to be able to reuse any view that has been created previously.

I understand this PR also introduced a new public API on `MGLAnnotationView` called `willBeEnqueued`. I'm not sure I see the value of that compared to the existing `prepareForReuse` API that is also more familiar to users that have experience with similar APIs (like MapKit's and UITableView). It may also leak more than necessary about the underlying reuse mechanism. Please do feel free to submit another PR with the changes for `willBeEnqueued` if you feel you can make the case for it.

Thanks again!

PR #6843, opened 2016-10-27
updated directions example in testapp to v5

### LINKED ISSUE #6811: Update directions example in Testapp
Just noticed we are still using v4 of our directions API in MAS, needs to be updated.

cc: @zugaldia

cammace:
Closes #6811 

cc: @tobrun for review


cammace:
Thanks for the approval @tobrun! merging

PR #6844, opened 2016-10-27
[core] Band-aid for crash triggered by unbound element buffer

jfirebaugh:
Refs #5731

cc @pveugen @kkaefer @friedbunny


kkaefer:
How are you verifying that this solves the issue?

jfirebaugh:
I'm not -- it's a purely speculative workaround.

PR #6848, opened 2016-10-28
[Qt] Fix viewport size for Linux on Qt5

### LINKED ISSUE #6827: Incorrect behavior of mbgl-qt example
**Platform:**
Environment:
Ubuntu 16.04
uname -a
Linux pc22 4.4.0-31-generic #50-Ubuntu SMP Wed Jul 13 00:07:12 UTC 2016 x86_64 x86_64 x86_64 GNU/Linux

**Mapbox SDK version:**
Qt SDK
Build according Qt instruction from master branch (sha a4d259c33f9bb890bba97fd89552720e3e0ec09b)

### Steps to trigger behavior

 1. Execute qt example application mbgl-qt

### Expected behavior
Window is displayed with map sized entire window.
### Actual behavior
Window is displayed with small map in left bottom corner.
screenshot from 2016-10-26 13-00-06

It was working fine last week (before backend refactoring).
QML example(mbgl-qt-qml) works fine.

This is how it works with  2 commits before version (sha 45f4dc0166f2d609d014d2174209fdbe1994c943)
screenshot from 2016-10-26 13-22-58

brunoabinader:
Qt 5.x for Linux platform implements a different behavior for `QGLWidget` when compared to the macOS - the viewport size is not initially set. Thus, we need to account for this by using the assignment operator that eventually calls for the viewport OpenGL setter function.

Fixes #6827.

/cc @kkaefer @tmpsantos

### DIFF
--- a/platform/qt/src/qmapboxgl.cpp
+++ b/platform/qt/src/qmapboxgl.cpp
@@ -850,16 +850,12 @@ void QMapboxGLPrivate::updateFramebufferBinding(QOpenGLFramebufferObject *fbo_)
     fbo = fbo_;
     if (fbo) {
         getContext().bindFramebuffer.setDirty();
-        getContext().viewport.setCurrentValue(
-            { 0,
-              0,
-              { static_cast<uint32_t>(fbo->width()), static_cast<uint32_t>(fbo->height()) } });
+        getContext().viewport = {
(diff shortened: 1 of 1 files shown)


brunoabinader:
Qt demo app on macOS is fine with the proposed change. I did find, though, that the latest Qt 5.7.0  Homebrew package makes our QML demo crash - tracking this in https://github.com/mapbox/mapbox-gl-native/issues/6849 - but this is unrelated to this change.

PR #6851, opened 2016-10-28
Improvements to Android Wear compatability

cammace:
First issue fixed is to make sure the tint color of attribution button isn't transparent. If it is we just use the `mapbox_blue` color. This resolves the issue with the button not showing on wear.

cc: @mapbox/android


cammace:
Positioning of compass and attribution are now adjusted for watches

device-2016-10-28-160200

cammace:
Closing this after discussing with @tobrun, more thought needs to be put into implementing Mapbox Android wear.

PR #6850, opened 2016-10-28
Xcode 8.1 recommended build settings for mbgl

kkaefer:
Xcode 8.1 recommends several changes to build settings in mbgl:

upgrade

Some of these settings throw errors in existing code:

```
/path/to/mapbox-gl-native/src/mbgl/renderer/painter_symbol.cpp:125:87: error: code will never be executed [-Werror,-Wunreachable-code]
    const bool drawAcrossEdges = (frame.mapMode == MapMode::Continuous) && (true || !(layout.textAllowOverlap || layout.iconAllowOverlap ||
                                                                                      ^~~~~~
/path/to/mapbox-gl-native/src/mbgl/renderer/painter_symbol.cpp:125:77: note: silence by adding parentheses to mark code as explicitly dead
    const bool drawAcrossEdges = (frame.mapMode == MapMode::Continuous) && (true || !(layout.textAllowOverlap || layout.iconAllowOverlap ||
                                                                            ^
                                                                            /* DISABLES CODE */ ( )
1 error generated.
```

```
/path/to/mapbox-gl-native/test/sprite/sprite_atlas.test.cpp:329:9: error: code will never be executed [-Werror,-Wunreachable-code]
        test.end();
        ^~~~
1 error generated.
```

These settings affect mbgl.xcodeproj, so they wouldn’t be checked in directly. Where are they managed?

/cc @kkaefer

### DIFF
--- a/platform/ios/ios.xcodeproj/xcshareddata/xcschemes/CI.xcscheme
+++ b/platform/ios/ios.xcodeproj/xcshareddata/xcschemes/CI.xcscheme
@@ -1,6 +1,6 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <Scheme
-   LastUpgradeVersion = "0800"
+   LastUpgradeVersion = "0810"
    version = "1.3">
    <BuildAction
       parallelizeBuildables = "YES"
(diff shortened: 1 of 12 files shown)


kkaefer:
Note that this PR does not actually add those flags (although we have most of them already in there, except for unreachable code).

kkaefer:
We don't really have unreachable code in droves; the few bits that are there are very intentional.

1ec5:
Er, I can’t actually approve this PR because GitHub thinks I _made_ it (since it was morphed from an issue I reported).

PR #6853, opened 2016-10-28
[core] operator bool() must always be explicit

jfirebaugh:
Otherwise, it can participate in unexpected conversions. Case in point: GlyphSet::insert was comparing the result of GlyphMetrics::operator bool() where it wanted to use operator==.


kkaefer:
Good catch. Did this affect any test?

jfirebaugh:
No, no test affected.

Issue #6799, opened 2016-10-23
[linux] Overdraw debug is broken


brunoabinader:
Overdraw debug mode no longer works on Linux;

```
terminate called after throwing an instance of 'mbgl::gl::Error'
  what():  glEnableVertexAttribArray(binding.location): Error GL_INVALID_VALUE at ../../../src/mbgl/gl/context.cpp:113
```

Bisection points out this issue started @ e4310aa87489c2db52d7ff65f71e51cc6c9700b6.

From my comments on the patch [that initially fixed overdraw on Linux](a17b0bcd82ebe8628b69588d19773d26596b86da):

> ```
> - Create a separate shader for each shader that supports overdraw.
>   Needed because each uniform location must be known for every program.
> - Create a separate VAO for each shader inside buckets.
>   Needed because we can only bind a VAO to a specific shader.
> ```

Unlike macOS, it seems that Linux mesa GL drivers are more strict into their usage policies. We should probably add an overdraw unit test to make sure both macOS and Linux supports these.

/cc @jfirebaugh

jfirebaugh:
This is what I'm seeing on macOS at a4c82b8a3b5e48f3bbccf32be80d45ca78d51515 and 39d5cb325012a5675676c81153c5e95d29fbc357 (both pre- and post-#6468). It looks like a bug in #6674 or #6596. cc @kkaefer 

![image](https://cloud.githubusercontent.com/assets/98601/19838164/b1e9af5c-9e86-11e6-8ac0-876955b00bcc.png)

kkaefer:
I cannot reproduce the issue you're seeing on macOS with most recent master (or any of the previous commits)

jfirebaugh:
@kkaefer I can reproduce it on a retina screen, but not an external 1x monitor.

kkaefer:
Thanks for the hint; I can reproduce with a 2x screen. @jfirebaugh the screenshot you posted displays the depth buffer; were you also able to reproduce the overdraw bug described by @brunoabinader in the initial comment?

jfirebaugh:
Ah, no, I was mistakenly conflating the overdraw view and the depth buffer view. I'm not able to reproduce the overdraw issue -- but I'm not sure it was ever reproducible on macOS. @brunoabinader can you check if #6468 fixed the issue on Linux?

brunoabinader:
@jfirebaugh just tested this after #6468 - same crash:
```
[...]

PR #6856, opened 2016-10-30
Shaders refactor

jfirebaugh:
Followup to #6468, specifically https://github.com/mapbox/mapbox-gl-native/pull/6468#discussion_r85246468.
- [x] Fix overdraw debug mode


brunoabinader:
Potentially fixes the issues mentioned in https://github.com/mapbox/mapbox-gl-native/issues/6799#issue-184702365 (needs verification).

jfirebaugh:
Ready for review!

jfirebaugh:
@kkaefer The test sources now need the path for generated shader .hpp's in their include paths. Where in the cmake files do I change that?

kkaefer:
You can add `${MBGL_GENERATED}/include` to the include dir list. The actual code that generates the shaders is in `cmake/shaders.cmake`.

kkaefer:
I'd propose adding this to enforce initialization of the primitive types with a value:

```diff
diff --git a/src/mbgl/gl/draw_mode.hpp b/src/mbgl/gl/draw_mode.hpp
index 8f7db0a..b099a5f 100644
--- a/src/mbgl/gl/draw_mode.hpp
+++ b/src/mbgl/gl/draw_mode.hpp
@@ -11,6 +11,7 @@ public:
     using Primitive = Point;
     static constexpr std::size_t bufferGroupSize = 1;
 
+    Points(float pointSize_) : pointSize(pointSize_) {}
     float pointSize;
 };
 
@@ -19,6 +20,9 @@ public:
     using Primitive = Line;
     static constexpr std::size_t bufferGroupSize = 2;
 
+    Lines(float lineWidth_) : lineWidth(lineWidth_) {
+        assert(lineWidth > 0);
+    }
     float lineWidth;
 };
 
@@ -29,6 +33,9 @@ public:
     using Primitive = Line;
     static constexpr std::size_t bufferGroupSize = 1;
 
+    LineStrip(float lineWidth_) : lineWidth(lineWidth_) {
+        assert(lineWidth > 0);
+    }
     float lineWidth;
 };
 
```

kkaefer:
One more thing that I'd like to see (but that is out of scope for this PR) is to somehow move shaders to the context. Background is that for Node.js, we'd like to use a shared context, and ideally we can share the shaders as well to avoid having all of these Programs that are just duplicates.

src/mbgl/util/indexed_tuple.hpp
```diff
@@ -0,0 +1,47 @@
...
+        return std::get<Index<I>>(*this);
+    }
+
+    template <class I>
+    const auto& get() const {
+        return std::get<Index<I>>(*this);
+    }
+};
+
+} // namespace mgbl
```
jfirebaugh: * Less verbose -- `std::get<typename As::SomeInnerType>(tuple)` becomes `tuple.template get<As>()`.
* Eliminates the need to use a separate helper and `std::index_sequence` in `Attributes::binder`.
* Will be consistent with use in #6868.

src/mbgl/gl/segment.hpp
```diff
@@ -8,19 +8,19 @@ namespace gl {
...
-          primitiveLength(primitiveLength_) {}
+          indexLength(indexLength_) {}
 
     const std::size_t vertexOffset;
-    const std::size_t primitiveOffset;
+    const std::size_t indexOffset;
 
     std::size_t vertexLength;
-    std::size_t primitiveLength;
+    std::size_t indexLength;
```
jfirebaugh: Fundamentally, a vertex/index buffer is a buffer of vertices/indexes, not a buffer of triangles, or lines, or whatever. Conceptually, that's how OpenGL works, and I found it easier to reason about things when I stopped trying to think in terms of buffers of primitives.

Implementation wise, it simplifies calculations like [this](https://github.com/mapbox/mapbox-gl-native/pull/6856/files#diff-617d0e0d315b4076d9fb82555ca6df0cL464).

It matches the internal implementation of the new `{Vertex,Index}Vector` -- I think if we were to have `std::vector<Triangle<Vertex>>`, the packing might come out wrong. You don't want to get in a situation where there might be extra padding every third vertex. `{Vertex,Index}Vector` still enforces filling the vector with the right sized groups though.

src/mbgl/programs/raster_program.hpp
```diff
@@ -0,0 +1,66 @@
...
+                p.x,
+                p.y
+            },
+            {
+                t.x,
+                t.y
+            }
+        };
+    }
+};
```
kkaefer: Can we factor out the attributes stuff into its own objects to be in line with the others?

jfirebaugh:
Rebasing this is getting to be a real drag. Can I get this approved and we can follow up on nits in other PRs?

brunoabinader:
Thank you @jfirebaugh - confirming that overdraw debug mode is operational again on Linux:
overdraw

Issue #5930, opened 2016-08-10
Remove sdk-manager-plugin


tobrun:
From the github repository of [sdk-manager-plugin](https://github.com/JakeWharton/sdk-manager-plugin):

> This plugin is deprecated and is no longer being developed. The Android tools team has committed to automatically downloading tools and dependencies in v2.2.0 and will be improving it in subsequent releases.

When AS 2.2.0 stable lands we should remove it.

PR #6865, opened 2016-10-31
removes sdk manager plugin

cammace:
Closes https://github.com/mapbox/mapbox-gl-native/issues/5930

cc: @tobrun for review


cammace:
Good point, removed rest of plugin

PR #6868, opened 2016-11-01
[core] Properties refactor

jfirebaugh:
First item from #4860.

The strategy here is outlined in https://github.com/mapbox/cpp/blob/master/C%2B%2B%20Structural%20Metaprogramming.md.

The interesting parts to look at are [`paint_property.hpp`](https://github.com/mapbox/mapbox-gl-native/pull/6868/files#diff-714291ad55774f40ffada5ed91e619de) and [`layout_property.hpp`](https://github.com/mapbox/mapbox-gl-native/pull/6868/files#diff-4560570d557c5a21eb9cac0b3586a13f) and one of the generated layer properties, like [`circle_layer_properties.hpp`](https://github.com/mapbox/mapbox-gl-native/pull/6868/files#diff-c9e293cd0728c34818f10960ea44a299).


src/mbgl/style/layers/symbol_layer_properties.hpp
```diff
@@ -9,69 +9,251 @@
...
-    PaintProperty<float> textHaloWidth { 0 };
-    PaintProperty<float> textHaloBlur { 0 };
-    PaintProperty<std::array<float, 2>> textTranslate { {{ 0, 0 }} };
-    PaintProperty<TranslateAnchorType> textTranslateAnchor { TranslateAnchorType::Map };
+struct SymbolPlacement : LayoutProperty<SymbolPlacementType> {
+    static SymbolPlacementType defaultValue() { return SymbolPlacementType::Point; }
 };
 
+struct SymbolSpacing : LayoutProperty<float> {
+    static float defaultValue() { return 250; }
```
kkaefer: Why are we using static functions instead of a `constexpr` member var?

jfirebaugh: Some property types are not `constexpr` (`std::string`, `std:array<float, 2>`). It's possible we could use `constexpr` type variants (`const char *`, `float[2]`) for `defaultValue` but for now I just went with this simple solution. It'll probably be inlined anyway.

jfirebaugh:
This is ready for review. Like #6856, it takes a fair amount of work to keep up to date against `master`, so I'd like to merge it soon.

src/mbgl/style/paint_property.hpp
```diff
@@ -57,83 +87,120 @@ class PaintProperty {
...
+    }
+
+    Cascading cascading;
+    Unevaluated unevaluated;
+    Evaluated evaluated;
 };
 
 } // namespace style
-} // namespace mbgl
+} // namespace mgbl
```
kkaefer: No.

src/mbgl/style/layout_property.hpp
```diff
@@ -1,44 +1,56 @@
...
-
-private:
-    T defaultValue;
-    PropertyValue<T> currentValue;
+    Unevaluated unevaluated;
 };
 
 } // namespace style
-} // namespace mbgl
+} // namespace mgbl
```
kkaefer: No.

include/mbgl/style/transition_options.hpp
```diff
@@ -6,9 +6,21 @@
...
+public:
     optional<Duration> duration = {};
     optional<Duration> delay = {};
+
+    TransitionOptions defaults(const TransitionOptions& defaults) const {
+        return {
+            duration ? duration : defaults.duration,
+            delay ? delay : defaults.delay
+        };
+    }
```
jfirebaugh: I was modeling on [`_.defaults`](https://lodash.com/docs/4.17.1#defaults) but would be fine with renaming it to `reverseMerge` or something like that.

PR #6871, opened 2016-11-01
Do not assign clip IDs to tiles that aren't rendered

kkaefer:
We are currently assigning clip IDs to all "renderable" tiles, regardless of the whether they are indeed rendered or not. A tile can be renderable but not rendered when it doesn't have any data/buckets that need rendering. This frequently happens for annotation tiles: We are "rendering" all annotations tiles (covering the entire viewport) even if there are no annotations on the map, or the annotations are elsewhere, or only a few tiles in the viewport contain annotations.


kkaefer:
After #6899 went in, we no longer see failing render tests on most runs, except for the Qt one, which still uses the old Xvfb-based rendering system. The test in question fails with an image like this:

actual

brunoabinader:
The rendering artifacts mentioned above should be fixed in #6933, where we upgrade our mesa GLX-based driver to version 13.0.0 and make it use llvmpipe instead of softpipe as software rendering backend.

brunoabinader:
Don't know why this hasn't been catch by our Linux build bots, but I just got this on my Linux desktop:
```
../../../src/mbgl/style/style.cpp: In member function ‘mbgl::RenderData mbgl::style::Style::getRenderData(mbgl::MapDebugOptions, float) const’:
../../../src/mbgl/style/style.cpp:431:27: error: assignment of member ‘mbgl::RenderTile::used’ in read-only object
                 tile.used = true;
                           ^
```
I'm using GCC (Ubuntu 5.4.0-6ubuntu1~16.04.2) 5.4.0 20160609.

brunoabinader:
Scratch that ^ this is due to my rebase of `fix-symbol-clipping` on top of master. All good :)

PR #6876, opened 2016-11-02
[core] Fix regression in Bucket::hasData

brunoabinader:
As per @jfirebaugh review in https://github.com/mapbox/mapbox-gl-native/commit/384f1a1960c9a21039642afe9bd1df58a93fddfc#commitcomment-19656901, commit 384f1a1960c9a21039642afe9bd1df58a93fddfc introduced a regression in `{Circle,Fill,Line,Symbol}Bucket::hasData()` in a sense that an empty bucket would never return `false`.

### DIFF
--- a/cmake/test-files.cmake
+++ b/cmake/test-files.cmake
@@ -23,6 +23,7 @@ set(MBGL_TEST_FILES
     test/geometry/binpack.test.cpp
 
     # gl
+    test/gl/bucket.test.cpp
     test/gl/object.test.cpp
 
     # include/mbgl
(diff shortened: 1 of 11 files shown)


kkaefer:
Why did you remove the `segments.empty()` check in the first place?

brunoabinader:
> Why did you remove the segments.empty() check in the first place?

Rationale is that we frequently update bucket contents - so we can save one less check on each pass by initializing the segments container with one empty segment.

kkaefer:
Okay, but it also feels weird to start out empty buckets with an actual segment in the list.

brunoabinader:
> Okay, but it also feels weird to start out empty buckets with an actual segment in the list.

Right, I can extract that change out of https://github.com/mapbox/mapbox-gl-native/commit/384f1a1960c9a21039642afe9bd1df58a93fddfc then.

PR #6880, opened 2016-11-02
Upgraded okhttp dependency to latest version

cammace:
We were still using `3.3.0` (latest is `3.4.1`). [Changelog](https://github.com/square/okhttp/blob/master/CHANGELOG.md). 

cc: @tobrun for review


tobrun:
👍

bradjonesca:
FYI: https://github.com/square/okhttp/issues/2756#issuecomment-258271983

bradjonesca:
@cammace FIX is in: https://github.com/square/okhttp/issues/2756#issuecomment-258337659

PR #6885, opened 2016-11-02
More verbose errors on invalid sprite metrics

mikemorris:
This breaks the `Can't create sprite with invalid metrics` catchall into individual verbose error cases. I'm pretty sure this should be tweaked slightly to avoid allocating the `std::stringstream` in non-error cases.

### DIFF
--- a/src/mbgl/sprite/sprite_parser.cpp
+++ b/src/mbgl/sprite/sprite_parser.cpp
@@ -5,6 +5,7 @@
 
 #include <mbgl/util/image.hpp>
 #include <mbgl/util/rapidjson.hpp>
+#include <mbgl/util/string.hpp>
 
 #include <cmath>
 #include <limits>
@@ -24,7 +25,10 @@ SpriteImagePtr createSpriteImage(const PremultipliedImage& image,
         ratio <= 0 || ratio > 10 ||
         srcX >= image.size.width || srcY >= image.size.height ||
         srcX + width > image.size.width || srcY + height > image.size.height) {
-        Log::Error(Event::Sprite, "Can't create sprite with invalid metrics");
+        Log::Error(Event::Sprite, "Can't create sprite with invalid metrics: %ux%u@%u,%u in %ux%u@%sx sprite",
+            width, height, srcX, srcY,
+            image.size.width, image.size.height,
+            util::toString(ratio).c_str());
         return nullptr;
     }
 
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "mapbox-gl-native",
-  "version": "3.3.3",
+  "version": "3.4.0",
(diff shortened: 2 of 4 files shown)


src/mbgl/sprite/sprite_parser.cpp
```diff
@@ -24,7 +25,11 @@ SpriteImagePtr createSpriteImage(const PremultipliedImage& image,
         ratio <= 0 || ratio > 10 ||
         srcX >= image.size.width || srcY >= image.size.height ||
         srcX + width > image.size.width || srcY + height > image.size.height) {
-        Log::Error(Event::Sprite, "Can't create sprite with invalid metrics");
+        Log::Error(Event::Sprite, "Can't create sprite with invalid metrics: " +
```
jfirebaugh: Use `Log` varargs support rather than string concatenation.

```
Log::Error(Event::Sprite, "Can't create sprite with invalid metrics: %dx%d@...",
           width, height, ...);
```

mikemorris: `%s` with `std::to_string(width).c_str()` works but feels a little silly.

mikemorris: Okay, got this working with `"%" PRIu32`... https://stackoverflow.com/questions/12120426/how-to-print-uint32-t-and-uint16-t-variables-value

src/mbgl/sprite/sprite_parser.cpp
```diff
@@ -24,7 +25,10 @@ SpriteImagePtr createSpriteImage(const PremultipliedImage& image,
         ratio <= 0 || ratio > 10 ||
         srcX >= image.size.width || srcY >= image.size.height ||
         srcX + width > image.size.width || srcY + height > image.size.height) {
-        Log::Error(Event::Sprite, "Can't create sprite with invalid metrics");
+        Log::Error(Event::Sprite, "Can't create sprite with invalid metrics: %" PRIu32 "x%" PRIu32 "@%" PRIu32 ",%" PRIu32 " in %" PRIu32 "x%" PRIu32 "@%sx sprite",
```
mikemorris: Sooo, the GLFW `g++5` builds and `android-ndk/arm-9-r12b/bin/arm-linux-androideabi-clang++` don't like this, although the QT `g++5` builds and other clang++ builds are fine... Is there a better way to handle this?

jfirebaugh: Damn, I didn't realize formatting fixed width integer types was so arcane. Try adding `#include <inttypes.h>`.

mikemorris: Got this working by reading `inttypes.h` and realizing that `PRIu32` just expands to `u`

PR #6888, opened 2016-11-03
[ios] Store annotation tags for MGLMultiPoint annotations in the map

boundsj:
This fixes an issue where, if a multipoint annotation was added, native annotation view update logic would crash. Adding the annotation tag for multipoints, too, avoids the out of range exception.

cc @1ec5 @incanus


boundsj:
Sorry @1ec5, I just realized I missed adding the check for multipoints. Does not really change anything but I added it in f0039fd for consistency.

Issue #5971, opened 2016-08-12
Enable runtime styling tests on macOS


1ec5:
220bd57fae06084584fa269fbd5367d4aa52196d disabled the runtime styling API tests on macOS. We should fix and reenable them.

/ref #5727
/cc @frederoni

1ec5:
Fixed in #6889.

PR #6889, opened 2016-11-03
Enable runtime styling tests on macOS

1ec5:
This PR adds a few missing header references to the macOS project and shares the runtime styling unit tests between the iOS and macOS test target. Despite the headers being missing from the project, the SDK still compiled without errors because the headers were already in the header search path. If they had been public headers, the compiler might’ve issued a warning about an incomplete umbrella header or non-module include statement. Unless we decide to go ahead with #5942, we’ll just have to remain vigilant about private headers that fly under the radar.

The runtime styling API can now round-trip colors without losing accuracy. Previously, we were using one color profile when converting NSColor to `mbgl::Color` and another when converting back to NSColor.

On macOS, we have to set the access token to a bogus access token before running any tests. The macOS implementation of MGLMapView doesn’t even attempt to set a default style if the access token is missing, because in macosapp, the access token may be set asynchronously by the user when the preferences window appears.

I’m targeting this PR against the release-ios-v3.4.0 branch because I intend to issue a macOS SDK release that’s as close to iOS SDK v3.4.0 as possible, and I want to ensure as much test coverage as possible.

/cc @frederoni @incanus


1ec5:
This fixed #5971.

Issue #6580, opened 2016-10-04
Use std::vector for MGLMultiPoint coordinate storage


1ec5:
MGLMultiPoint represents its coordinates internally as a C array of `CLLocationCoordinate2D` instances. This was fine at first, because the coordinates were immutable once MGLMultiPoint was created. Making the coordinates mutable (#6565) is forcing us to reckon with manual C memory management. Any errors there could potentially turn into security vulnerabilities.

We should convert the internal `_coordinates` array into an `std::vector<CLLocationCoordinate2D>` and let that object handle heap allocations for us. The public `coordinates` property would be implemented as `_coordinates.data()`, and `pointCount` as `_coordinates.size()`. The upshot is that the `memcpy()` grunt work in #6565 could be replaced with nothing more than `std::vector::reserve()` and [`std::copy_backward()`](http://www.cplusplus.com/reference/algorithm/copy_backward/).

/cc @incanus @boundsj @alulsh

1ec5:
Putting this on the v3.4.0 milestone. #6565 and #6583 only increase the risk of making an off-by-one error that can’t be detected by unit tests yet could lead to security vulnerabilities.

incanus:
Re: being on the milestone, the original feature in https://github.com/mapbox/mapbox-gl-native/pull/6565 was driven by a customer request and is "gravy" of a sort (mmmmm) beyond the focus on 3.4.0, which is runtime styling. 

Maybe consider this a loose milestone issue that can be punted on? We can always tighten things up in a point release. 

/cc @boundsj

1ec5:
Even if we punt on #6583, I think we should keep this refactoring on the milestone. It’s small and necessary given that we have no tests of this functionality. (Or we can add tests!) https://github.com/mapbox/mapbox-gl-native/pull/6565#discussion_r82091206 is illustrative of the murky waters we’re in right now.

boundsj:
Adding `release blocker` tag due to the safety concerns raised here

boundsj:
This was fixed in https://github.com/mapbox/mapbox-gl-native/pull/6890

PR #6890, opened 2016-11-03
[ios, macos] Use std::vector for MGLMultiPoint coordinate storage

boundsj:
Fixes https://github.com/mapbox/mapbox-gl-native/issues/6580

cc @incanus @1ec5

### DIFF
--- a/platform/darwin/src/MGLMultiPoint.mm
+++ b/platform/darwin/src/MGLMultiPoint.mm
@@ -68,7 +68,7 @@ mbgl::Color MGLColorObjectFromCGColorRef(CGColorRef cgColor)
 - (void)appendCoordinates:(CLLocationCoordinate2D *)coords count:(NSUInteger)count
 {
     [self willChangeValueForKey:@"coordinates"];
-    _coordinates.insert(_coordinates.end(), *coords);
+    _coordinates.insert(_coordinates.end(), count, *coords);
     [self computeBounds];
     [self didChangeValueForKey:@"coordinates"];
 }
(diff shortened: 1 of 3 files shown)


boundsj:
This won't actually work in practice without https://github.com/mapbox/mapbox-gl-native/pull/6888 😬

1ec5:
Even though MGLPointCollection is immutable, it would be great if we could migrate it over to `std::vector` too for consistency.

boundsj:
Still a WIP here @1ec5 so no need to re-review yet. Will come back and finish up the rest after 3.3.6 is released.

platform/darwin/src/MGLMultiPoint.mm
```diff
@@ -24,105 +24,84 @@ - (instancetype)initWithCoordinates:(CLLocationCoordinate2D *)coords count:(NSUI
...
 
-    for (NSUInteger i = range.location; i < range.location + range.length; i++)
-    {
-        coords[index] = _coordinates[i];
-        index++;
-    }
+- (void)appendCoordinates:(CLLocationCoordinate2D *)coords count:(NSUInteger)count
+{
+    [self willChangeValueForKey:@"coordinates"];
+    _coordinates.insert(_coordinates.end(), *coords);
```
1ec5: This appends only the first coordinate in the passed-in array. To append the entire array:

```objc
_coordinates.insert(_coordinates.end(), coords, coords + count);
```

platform/darwin/src/MGLMultiPoint.mm
```diff
@@ -24,105 +24,84 @@ - (instancetype)initWithCoordinates:(CLLocationCoordinate2D *)coords count:(NSUI
...
+    std::copy(coords, coords + range.length, _coordinates.begin() + range.location);
+    [self computeBounds];
+    [self didChangeValueForKey:@"coordinates"];
+}
+
+- (void)computeBounds
+{
+    CLLocationCoordinate2D *coords = self.coordinates;
+    mbgl::LatLngBounds bounds = mbgl::LatLngBounds::empty();
+    for (NSUInteger i = 0; i < [self pointCount]; i++)
```
1ec5: This will call `-pointCount` on every iteration. Use C++ fast enumeration instead:

```cpp
for (auto coordinate : _coordinates) {
    …
}
```

PR #6893, opened 2016-11-03
Force explicit uniform initialization

kkaefer:
We currently initialize the uniform tuples by just listing all of the values. However, there are some shaders that use the same types, so it's easy to mix up values. This PR changes the initialization so that users must explicitly specify the uniform they are setting the value for.


kkaefer:
@jfirebaugh I found that the opposite is the case: simply adding `explicit` to the `Value` constructor didn't work for and Clang happily compiled the code. Here's a reduced test case:

```cpp
template <typename Tag, typename T>
struct uniform {
    struct value {
        value(T t_) : t(std::move(t_)) {}
        T t;
    };
};

struct foo : uniform<foo, int> {};
struct bar : uniform<bar, int> {};

using values = std::tuple<foo::value, bar::value>;

values v { 1, 2 };
```

Adding `explicit` to the value constructor doesn't change anything; only using aggregate initialization requires explicit construction.

Unfortunately, just leaving out the name also works. Back to the drawing board.

jfirebaugh:
Interesting. I guess the issue is that `std::tuple` has a converting constructor that does explicit construction internally. What if we change:

```
    using Values = std::tuple<typename Us::Value...>;
```

to

```
    class Values : public std::tuple<typename Us::Value...> {
    public:
        explicit Values(const typename Us::Value&... values)
            : std::tuple<typename Us::Value...>(values...) {}
    };
```

kkaefer:
Nope, this has the same issue of not requiring the explicit naming

kkaefer:
Moving the Value class outside in 59f6487 allows me to add `explicit`, and actually requires the explicit name for constructing the value object.

kkaefer:
This is ready for review

PR #6895, opened 2016-11-03
[android] update custom layer example with invalidation example

ivovandongen:
Little more clarity about how to use CustomLayer::update to trigger a repaint.

ezgif com-video-to-gif 2


tobrun:
👍

PR #6904, opened 2016-11-03
Layer ownership refactor

fabian-guerra:
Fixes https://github.com/mapbox/mapbox-gl-native/issues/6254

cc @boundsj @1ec5

### DIFF
--- a/platform/darwin/src/MGLFillStyleLayer.h
+++ b/platform/darwin/src/MGLFillStyleLayer.h
@@ -96,6 +96,9 @@ typedef NS_ENUM(NSUInteger, MGLFillTranslateAnchor) {
  */
 @property (nonatomic, null_resettable) MGLStyleValue<NSString *> *fillPattern;
 
+
+- (void)addToMapView:(MGLMapView *)mapView;
+
 @end
 
 NS_ASSUME_NONNULL_END
--- a/platform/darwin/src/MGLFillStyleLayer.h
+++ b/platform/darwin/src/MGLFillStyleLayer.h
@@ -96,9 +96,6 @@ typedef NS_ENUM(NSUInteger, MGLFillTranslateAnchor) {
  */
 @property (nonatomic, null_resettable) MGLStyleValue<NSString *> *fillPattern;
 
-
-- (void)addToMapView:(MGLMapView *)mapView;
-
 @end
 
 NS_ASSUME_NONNULL_END
--- a/platform/darwin/src/MGLFillStyleLayer.mm
+++ b/platform/darwin/src/MGLFillStyleLayer.mm
@@ -17,23 +17,13 @@
 @end
 
 @implementation MGLFillStyleLayer
-{
-    std::unique_ptr<mbgl::style::FillLayer> _pendingLayer;
-}
 
 - (instancetype)initWithIdentifier:(NSString *)identifier source:(MGLSource *)source
 {
     if (self = [super initWithIdentifier:identifier source:source]) {
-        [self commonInit:identifier source:source];
+        _layer = new mbgl::style::FillLayer(identifier.UTF8String, source.identifier.UTF8String);
     }
     return self;
-}
-
-- (void)commonInit:(NSString *)identifier source:(MGLSource *)source
-{
-    auto layer = std::make_unique<mbgl::style::FillLayer>(identifier.UTF8String, source.identifier.UTF8String);
-    _pendingLayer = std::move(layer);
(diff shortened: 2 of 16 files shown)


platform/ios/app/Main.storyboard
```diff
@@ -1,5 +1,5 @@
-<?xml version="1.0" encoding="UTF-8"?>
-<document type="com.apple.InterfaceBuilder3.CocoaTouch.Storyboard.XIB" version="3.0" toolsVersion="11201" systemVersion="16A323" targetRuntime="iOS.CocoaTouch" propertyAccessControl="none" useAutolayout="YES" useTraitCollections="YES" colorMatched="YES" initialViewController="PSe-Ot-7Ff">
+<?xml version="1.0" encoding="UTF-8" standalone="no"?>
```
fabian-guerra: I didn't meant to merge this storyboard.

platform/darwin/src/MGLFillStyleLayer.h
```diff
@@ -96,6 +96,9 @@ typedef NS_ENUM(NSUInteger, MGLFillTranslateAnchor) {
  */
 @property (nonatomic, null_resettable) MGLStyleValue<NSString *> *fillPattern;
 
+
+- (void)addToMapView:(MGLMapView *)mapView;
```
1ec5: This method needs to be private. We want developers to continue to call `-[MGLStyle addLayer:]` or `-[MGLStyle insertLayer:belowLayer:]` as appropriate. To avoid having to create a separate private header for each layer subclass, declare it in the private header for MGLStyle.

platform/darwin/src/MGLFillStyleLayer.mm
```diff
@@ -2,6 +2,7 @@
 // Edit platform/darwin/scripts/generate-style-code.js, then run `make style-code-darwin`.
 
 #import "MGLSource.h"
+#import "MGLMapView_Private.h"
```
1ec5: These changes will eventually need to happen in all the style layer subclasses. We have a script (mentioned above) that automatically generates these headers. So when you're ready, update the templates that the script uses (the .ejs files in this directory) and run `make style-code-darwin`. Don't do that until you've updated the templates with your changes, though!

fabian-guerra: I have updated the script to import MGLMapView_Private.h

fabian-guerra:
Yesterday when I was profiling I noticed a small memory leak, with the changes we have implemented we no longer have leaks in StyleLayer classes.
 
nomemoryleaks

platform/darwin/src/MGLStyleLayer.mm.ejs
```diff
@@ -21,23 +22,39 @@
...
+    auto layer = std::make_unique<mbgl::style::<%- camelize(type) %>Layer>(identifier.UTF8String);
+    _pendingLayer = std::move(layer);
+    self.layer = _pendingLayer.get();
+}
 <% } else { -%>
 - (instancetype)initWithIdentifier:(NSString *)identifier source:(MGLSource *)source
 {
     if (self = [super initWithIdentifier:identifier source:source]) {
-        _layer = new mbgl::style::<%- camelize(type) %>Layer(identifier.UTF8String, source.identifier.UTF8String);
+        [self commonInit:identifier source:source];
```
boundsj: Not a big deal but the `commonInit` pattern is typically used when there is some shared logic that needs to be called by several different initializers and/or something that happens in init also happens latter on. In this case it might be clearer to just do the work in `initWithIdentifier:source:` itself.

platform/darwin/src/MGLStyleLayer_Private.h
```diff
@@ -5,9 +5,20 @@
 
 #include <mbgl/style/layer.hpp>
 
+@class MGLMapView;
+
 @interface MGLStyleLayer (Private)
 
 @property (nonatomic, readwrite, copy) NSString *identifier;
 @property (nonatomic) mbgl::style::Layer *layer;
```
boundsj: For future reference, I think we should add documentation for this property like we did in `MGLSource_Private.h` for the `rawSource` property. When coming back to this code (or looking at it for the first time) it is not immediately obvious that this is a raw pointer that is useable even after layer ownership is transferred in `addToMapView:`. Also, for consistency, it might be worth considering renaming this to `rawLayer` (or naming `[MGLSource rawSource]` to `[MGLSource source]` (again)).

Issue #6897, opened 2016-11-03
Camera bearing - wrong direction


tomlocks:
**Platform: Nexus 5 Android 6.0.1**
**Mapbox SDK version: 4.2.0beta4**

### Steps to trigger behavior

 1. Update camera position using this code:
   ```
 final CameraPosition.Builder builder = new CameraPosition.Builder();
    builder.target(new LatLng(52.39115,16.96389)).bearing(90).zoom(12);
    final CameraUpdate cameraUpdate = CameraUpdateFactory.newCameraPosition(builder.build());
    mapboxMap.animateCamera(cameraUpdate);
```

### Expected behavior
Map should face east.
### Actual behavior
Map is facing west.

Subtracting 360 from bearing resolves this issue, but I think that it isn't correct solution.

tobrun:
Thanks for the report, able to reproduce, fixing for the `4.2.0` release.

tobrun:
Will be resolved when #6917 is merged in release branch, thank you for reaching out closing.

johannesanderwald:
Hello,
When will be there a new beta with this issue fixed?
Best regards

PR #6917, opened 2016-11-04
[android]  - using bearing clockwise versus counterclockwise

tobrun:
closes https://github.com/mapbox/mapbox-gl-native/issues/6897.
Review @ivovandongen.


platform/android/src/jni.cpp
```diff
@@ -498,7 +498,7 @@ jdoubleArray nativeGetCameraValues(JNIEnv *env, jni::jobject* obj, jlong nativeM
     jdouble buf[5];
     buf[0] = latLng.latitude;
     buf[1] = latLng.longitude;
-    buf[2] = -(nativeMapView->getMap().getBearing()-360);
```
ivovandongen: Nit; spaces

tobrun: Removed the brackets

Issue #4746, opened 2016-04-19
onCameraChange calls during camera animation (by touch)


programmerr47:
This is copy of my comment: https://github.com/mapbox/mapbox-gl-native/issues/3745#issuecomment-211450083

Hello. Sorry for inconvenience but could you please help me with my porblem.
I'm writing android app with help of Android Studio. I'm using gradle and targetSdk 23. Also I'm using mapbox sdk 4.0.0-beta.2.

In my fragment I initialize mapbox maps something like this:

``` Java
    @Override
    public void onActivityCreated(@Nullable Bundle savedInstanceState) {
        ...
        mapView.onCreate(null);
        setupMapView(getActivity());
        ...
    }

    private void setupMapView(Context context) {
        ...
        mapView.getMapAsync(this);
        ...
    }

    @Override
    public void onMapReady(@NonNull MapboxMap mapboxMap) {
        this.mapboxMap = mapboxMap;
        ...
        this.mapboxMap.setOnCameraChangeListener(this);
        ...
    }

    @Override
    public void onCameraChange(CameraPosition cameraPosition) {
        Log.v("TestLog", "OnCameraChange: " + cameraPosition);
    }
```

When I run my project and try to use scroll or fling gestures on the map (without stopping it), I see in logs that onCameraChange calls every 400 ms (average). But for me expected result that it called only one time: when camera will be stopped. 

Maybe described situation is correct, but as I understood MapBox's OnCameraChangeListener is analog of GoogleMaps's OnCameraChangeListener and last works in the way that I described (onCameraChange calls only when camera will be in "idle" state). At least in mapbox documentation said same thing as in google maps documentation: 

> Called after the camera position has changed. During an animation, this listener may not be notified of intermediate camera positions. It is always called for the final position in the animation.

I came to this problem, because earlier I used google maps and used onCameraChange callback as indicator that map is in idle state (because I hide controls when work with map and want to show them, when camera is idle and I don't touch to map), and now I need to migrate to mapbox (because you have more and more better detalization of maps for some countries, like Russia), and I stunned with this problem. 

So if it is not hard to you, could you please help me? Is calling onCameraChange during camera animation a bug or not? If not, does mapbox provide callbacks to know when camera is in the idle state? 

Thank you.

BTW: 
I'm testing on Nexus 5, v.6.0.1. And the previous example worked on Google maps.

tobrun:
@programmerr47  
Thank you for reaching out and reporting your issue.

[...]

PR #6803, opened 2016-10-24
Scale Instrumentation tests

tobrun:
WIP to cover a big part of our public API with instrumentation tests. This will help us start refactor work from https://github.com/mapbox/mapbox-gl-native/projects/4 without breaking the whole API. 

Initial focus:
- [x] moveCamera
- [x] easeCamera
- [x] AnimateCamera
- [x] setBearing
- [x] setPitch
- [x] setLatLng
- [x] CompassView
- [x] MyLocationView
- [x] MyLocationTracking/MyLocationBearing
- [x] Attribution
- [x] addMarker
- [x] addMarkerView
- [x] addPolygon
- [x] addPolyline
- [x] selectMarker (show infowindow)
- [x] CustomInfoWindowAdapter
- [x] ...

PR #6919, opened 2016-11-04
[android] - move camera logic to dedicated transform class

### LINKED ISSUE #6532: Transform.java refactor
Introduce the concept of the core equivalent of transform in the Android binding. Will contain all our camera related operations by removing these from MapboxMap. MapboxMap will remain the public api entrypoint (facade) but all code should be contained within `Transform.java`. This refactor should result in removing the equivalent methods in MapView.

tobrun:
WIP - closes #6532 - This PR moves camera logic to a dedicated java class. 

Todo
 - [x] remove/decouple Mapbox/MapView dependencies from class  
 - [x] call into NativeMapView directly
 - [x] fixup camera invalidation logic from https://github.com/mapbox/mapbox-gl-native/issues/4746

After this we can look into integrating https://github.com/mapbox/mapbox-gl-native/issues/4746

Review @ivovandongen.

### DIFF
--- a/platform/android/MapboxGLAndroidSDK/src/main/java/com/mapbox/mapboxsdk/maps/AnnotationManager.java
+++ b/platform/android/MapboxGLAndroidSDK/src/main/java/com/mapbox/mapboxsdk/maps/AnnotationManager.java
@@ -30,7 +30,7 @@ import java.util.List;
  * Exposes convenience methods to add/remove/update all subtypes of annotations found in com.mapbox.mapboxsdk.annotations.
  * </p>
  */
-class AnnotationManager {
+class AnnotationManager implements MapView.OnMapChangedListener {
 
     private NativeMapView nativeMapView;
     private IconManager iconManager;
@@ -41,6 +41,7 @@ class AnnotationManager {
     private List<Marker> selectedMarkers;
 
     private MapboxMap.OnMarkerClickListener onMarkerClickListener;
+    private boolean isWaitingForRenderInvoke;
 
     AnnotationManager(NativeMapView view, IconManager iconManager, InfoWindowManager manager) {
         this.nativeMapView = view;
@@ -48,6 +49,19 @@ class AnnotationManager {
         this.infoWindowManager = manager;
         this.selectedMarkers = new ArrayList<>();
         this.annotations = new LongSparseArray<>();
+
+        if (view != null) {
+            // null checking needed for unit tests
+            view.addOnMapChangedListener(this);
+        }
+    }
+
+    @Override
+    public void onMapChanged(@MapView.MapChange int change) {
+        if (isWaitingForRenderInvoke && change == MapView.DID_FINISH_RENDERING_FRAME_FULLY_RENDERED) {
+            isWaitingForRenderInvoke = false;
+            markerViewManager.invalidateViewMarkersInVisibleRegion();
+        }
     }
 
     //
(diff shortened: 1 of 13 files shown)


tobrun:
Been looking into `remove/decouple Mapbox/MapView dependencies from class` while the class members were removed, I will not be able to exclude MapboxMap in total. Primary reason behind this is the usage of CameraUpdateFactory. This class takes in a MapboxMap and uses the following items from that class:

 - CameraPosition
 - UiSettings
 - TrackingSettings
 - Projection
 - minZoom
 - maxZoom

 In a normal case, we would refactor those items into constructor and replace references. For this case, I'm thinking about just leaving MapboxMap and have these dependencies resolved from that object instead. The relation between CameraUpdateFactoy and MapboxMap is a bit too complex to extract and simplify.

tobrun:
Side note: I had to remove the unit tests related to camera. This is because we aren't able to mock out `NativeMapView` in the newly introduced `Tranform` class. I recently wrote instrumentation tests on our camera API #6803, these will cover both the old unit tested implementation as the core interaction behind it.

tobrun:
2 more changes have been introduced:
 -  wrap camera update around a runnable, this is needed to be able to call an new CameraUpdate from a `onFinish` callback from another CameraUpdate. What currently occurs is that the second animation is started but is automatically canceled because the other animation wasn't fully finished.
 - simplify the internal workings on transfrom:  using 1 callback instead of 2, only one method for each nativecall.

platform/android/MapboxGLAndroidSDK/src/main/java/com/mapbox/mapboxsdk/maps/AnnotationManager.java
```diff
@@ -194,18 +194,40 @@ private Marker prepareMarker(BaseMarkerOptions markerOptions) {
     }
 
     MarkerView addMarker(@NonNull BaseMarkerViewOptions markerOptions, @NonNull MapboxMap mapboxMap) {
+        nativeMapView.addOnMapChangedListener(new MapView.OnMapChangedListener() {
+            @Override
+            public void onMapChanged(@MapView.MapChange int change) {
+                if (change == MapView.DID_FINISH_RENDERING_FRAME_FULLY_RENDERED) {
```
ivovandongen: We probably should batch this so that we do only one invalidate per render instead of one per marker added

tobrun: Done with 07ae553

platform/android/MapboxGLAndroidSDK/src/main/java/com/mapbox/mapboxsdk/maps/MapboxMap.java
```diff
@@ -424,17 +432,13 @@ public final void moveCamera(CameraUpdate update) {
...
-        mapView.jumpTo(cameraPosition.bearing, cameraPosition.target, cameraPosition.tilt, cameraPosition.zoom);
-        if (callback != null) {
-            callback.onFinish();
-        }
-
-        if (onCameraChangeListener != null) {
-            onCameraChangeListener.onCameraChange(this.cameraPosition);
-        }
+    public final void moveCamera(final CameraUpdate update, final MapboxMap.CancelableCallback callback) {
+        mapView.post(new Runnable() {
```
sddamico: @tobrun this breaks the existing contract that the new camera position will be query-able immediately and might have weird consequences in regards to "when this event occurs relative to activity lifecycle" because of the way the main thread works (lifecycle events are posted as items in the main thread handler).

Do you have a recommended strategy then for monitoring when this camera move event has been completed?

PR #6920, opened 2016-11-04
Remove remaining uses of gl.hpp

kkaefer:
We should only use `gl.hpp` in the OpenGL bindings, and not in any other code.

### DIFF
--- a/src/mbgl/gl/value.hpp
+++ b/src/mbgl/gl/value.hpp
@@ -282,6 +282,34 @@ struct PixelStoreUnpack {
     static Type Get();
 };
 
+struct PixelTransferDepth {
+    struct Type {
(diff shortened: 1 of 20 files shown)


src/mbgl/gl/value.hpp
```diff
@@ -259,7 +259,7 @@ struct RasterPos {
         double z;
         double w;
     };
-    static const constexpr Type Default = { 0, 0, 0, 0 };
+    static const constexpr Type Default = { 0, 0, 0, 1 };
```
kkaefer: According to the [documentation](https://www.opengl.org/sdk/docs/man2/xhtml/glRasterPos.xml), this is the actual default.

Issue #6901, opened 2016-11-03
Crash adding annotation — -[MGLMapView visibleAnnotationsInRect:]


bsudekum:
**Platform: iOS**
**Mapbox SDK version: 3.4.0-beta.2**

### Steps to trigger behavior

 1. Add annotation

### Expected behavior

The annotation is added

### Actual behavior

The app crashes with

```
Mapbox`::-[MGLMapView visibleAnnotationsInRect:](CGRect):
    0x1041d487b <+0>:   pushq  %rbp
    0x1041d487c <+1>:   movq   %rsp, %rbp
    0x1041d487f <+4>:   pushq  %r15
    0x1041d4881 <+6>:   pushq  %r14
    0x1041d4883 <+8>:   pushq  %r13
    0x1041d4885 <+10>:  pushq  %r12
    0x1041d4887 <+12>:  pushq  %rbx
    0x1041d4888 <+13>:  subq   $0x68, %rsp
    0x1041d488c <+17>:  movq   %rdi, %rbx
    0x1041d488f <+20>:  movq   0x6b860a(%rip), %r14      ; MGLMapView._annotationContextsByAnnotationTag
    0x1041d4896 <+27>:  xorl   %r15d, %r15d
    0x1041d4899 <+30>:  cmpq   $0x0, 0x18(%rbx,%r14)
    0x1041d489f <+36>:  je     0x1041d4a19               ; <+414> at MGLMapView.mm:2847
    0x1041d48a5 <+42>:  leaq   0x10(%rbp), %rax
    0x1041d48a9 <+46>:  movq   0x6b6c40(%rip), %rdx      ; "annotationTagsInRect:"
    0x1041d48b0 <+53>:  movq   0x18(%rax), %rcx
    0x1041d48b4 <+57>:  movq   %rcx, 0x18(%rsp)
    0x1041d48b9 <+62>:  movq   0x10(%rax), %rcx
    0x1041d48bd <+66>:  movq   %rcx, 0x10(%rsp)
    0x1041d48c2 <+71>:  movq   (%rax), %rcx
    0x1041d48c5 <+74>:  movq   0x8(%rax), %rax
    0x1041d48c9 <+78>:  movq   %rax, 0x8(%rsp)
    0x1041d48ce <+83>:  movq   %rcx, (%rsp)
    0x1041d48d2 <+87>:  leaq   -0x40(%rbp), %rdi
    0x1041d48d6 <+91>:  movq   %rbx, %rsi
[...]

PR #6924, opened 2016-11-04
[ios] Work around annotation race condition

boundsj:
Fixes https://github.com/mapbox/mapbox-gl-native/issues/6901

When annotations are removed concurrently with being updated because of a map redraw, it is possible that mbgl will still consider an annotation to exist depending on what part of the map lifecycle the call update annotations happens. However, the annotation context by
tag map in MGLMapView (on the main thread) will always be up to date.
This adds a guard that uses that map to avoid considering annotations that have actually been removed as visible.

Another issue was that adding the same annotation object twice has always been allowed and has worked, visually. However, a recent refactor to make the reuse queue work correctly is not tolerant of the a single view representing two annotation instances that are the same.
The effect is that a new UIView will be generated for each (clone) annotation that is added and only the last view will move with the map. The other views become detached and float around in the view as the map is panned. This commit adds a guard to make adding an annotation instance a no-op if it has already been added before.

cc @1ec5 @incanus @bsudekum

### DIFF
--- a/platform/ios/src/MGLMapView.mm
+++ b/platform/ios/src/MGLMapView.mm
@@ -2836,7 +2836,11 @@ public:
         
         for (auto const& annotationTag: annotationTags)
         {
-            MGLAnnotationContext annotationContext = _annotationContextsByAnnotationTag[annotationTag];
+            if (!_annotationContextsByAnnotationTag.count(annotationTag))
+            {
+                continue;
+            }
+            MGLAnnotationContext annotationContext = _annotationContextsByAnnotationTag.at(annotationTag);
             [annotations addObject:annotationContext.annotation];
         }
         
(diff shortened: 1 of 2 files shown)


platform/ios/src/MGLMapView.mm
```diff
@@ -2836,7 +2836,11 @@ - (void)removeStyleClass:(NSString *)styleClass
         
         for (auto const& annotationTag: annotationTags)
         {
-            MGLAnnotationContext annotationContext = _annotationContextsByAnnotationTag[annotationTag];
+            if (!_annotationContextsByAnnotationTag.count(annotationTag))
```
1ec5: Can you make these changes in the macOS implementation too?

boundsj: Oops. Done in 02e5978

platform/ios/src/MGLMapView.mm
```diff
@@ -2903,6 +2907,12 @@ - (void)addAnnotations:(NS_ARRAY_OF(id <MGLAnnotation>) *)annotations
     {
         NSAssert([annotation conformsToProtocol:@protocol(MGLAnnotation)], @"annotation should conform to MGLAnnotation");
 
+        // adding the same annotation object twice is a no-op
+        if ([self.annotations containsObject:annotation])
```
1ec5: Also make this change in the macOS implementation.

boundsj: 😬  done in ecf14e3

PR #6928, opened 2016-11-04
[core] Separate pathway for notifying workers of new images/glyphs

### LINKED ISSUE #6820: Runtime changes to text-field are not immediate
On the current iOS release branch (dbfa8dcc89ff55fa1f8366433d24b1a270e53a02), changing the value of `MGLStyleValue.textField` (in this case, the language labels via #6819) is not reflected immediately on the map.

Instead, the labels go blank.
- Zooming several levels (and triggering a [re]load?), then zooming back to the original zoom level will cause the labels to appear correctly at the original zoom level.
- Panning outside of the viewport at the original zoom level (to an area that was not shown) _will_ show the labels in those new areas.

asdf

/cc @boundsj @1ec5 @ivovandongen

jfirebaugh:
Fixes #6820. Corresponding test-suite PR: https://github.com/mapbox/mapbox-gl-test-suite/pull/158.

This will need to be cherrypicked to the iOS and Android release branches.

### DIFF
--- a/package.json
+++ b/package.json
@@ -23,7 +23,7 @@
     "lodash": "^4.16.4",
     "mapbox-gl-shaders": "mapbox/mapbox-gl-shaders#98a56d538b11fb331aa67a6d632d6ecd6821b007",
     "mapbox-gl-style-spec": "mapbox/mapbox-gl-style-spec#7f62a4fc9f21e619824d68abbc4b03cbc1685572",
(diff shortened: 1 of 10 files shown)


cammace:
cc: @mapbox/android

boundsj:
I tested this on the iOS release branch it looks 👍 . I can PR the cherry-pick for the iOS release branch if this lands on master and does not have additional changes.

cc @ivovandongen @tobrun

ivovandongen:
I've added a test case for Android in https://github.com/mapbox/mapbox-gl-native/pull/6936. Looks good. Some flickering when it reloads sometimes though.

PR #6932, opened 2016-11-06
Reflect changes in icon size immediately

### LINKED ISSUE #6903: SymbolLayer iconSize property not updating immediately
In the GIF when I click the map, the icon's suppose to change it's size but doesn't immediately change until I zoom in or out. 

ezgif com-video-to-gif 18

cc: @mapbox/android

ivovandongen:
Fixes #6903

### DIFF
--- a/src/mbgl/style/layers/symbol_layer.cpp
+++ b/src/mbgl/style/layers/symbol_layer.cpp
@@ -71,7 +71,7 @@ void SymbolLayer::setSymbolPlacement(PropertyValue<SymbolPlacementType> value) {
     if (value == getSymbolPlacement())
         return;
     impl->layout.symbolPlacement.set(value);
-    impl->observer->onLayerLayoutPropertyChanged(*this);
+    impl->observer->onLayerLayoutPropertyChanged(*this, "symbol-placement");
 }
 PropertyValue<float> SymbolLayer::getDefaultSymbolSpacing() {
     return { 250 };
@@ -85,7 +85,7 @@ void SymbolLayer::setSymbolSpacing(PropertyValue<float> value) {
     if (value == getSymbolSpacing())
         return;
     impl->layout.symbolSpacing.set(value);
-    impl->observer->onLayerLayoutPropertyChanged(*this);
+    impl->observer->onLayerLayoutPropertyChanged(*this, "symbol-spacing");
 }
 PropertyValue<bool> SymbolLayer::getDefaultSymbolAvoidEdges() {
     return { false };
@@ -99,7 +99,7 @@ void SymbolLayer::setSymbolAvoidEdges(PropertyValue<bool> value) {
     if (value == getSymbolAvoidEdges())
         return;
     impl->layout.symbolAvoidEdges.set(value);
-    impl->observer->onLayerLayoutPropertyChanged(*this);
+    impl->observer->onLayerLayoutPropertyChanged(*this, "symbol-avoid-edges");
 }
 PropertyValue<bool> SymbolLayer::getDefaultIconAllowOverlap() {
     return { false };
@@ -113,7 +113,7 @@ void SymbolLayer::setIconAllowOverlap(PropertyValue<bool> value) {
     if (value == getIconAllowOverlap())
         return;
     impl->layout.iconAllowOverlap.set(value);
-    impl->observer->onLayerLayoutPropertyChanged(*this);
+    impl->observer->onLayerLayoutPropertyChanged(*this, "icon-allow-overlap");
 }
 PropertyValue<bool> SymbolLayer::getDefaultIconIgnorePlacement() {
     return { false };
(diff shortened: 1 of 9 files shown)


ivovandongen:
@jfirebaugh The problem seems to be that the style needs to be recalculated for the change to appear. Recalculating on all layout property changes is not very efficient, we can limit this to `SymbolLayer`s or add another signal to the observer specifically for attributes that need style recalculation.

ivovandongen:
Added test case in: https://github.com/mapbox/mapbox-gl-test-suite/pull/161

jfirebaugh:
The regression test passes for me even without the changes here. Can you see what's needed to get that to fail, so we know it's a good regression test?

ivovandongen:
@jfirebaugh I've been trying to get the test to fail without success. Whereas the [test activity](https://github.com/mapbox/mapbox-gl-native/pull/6936) I made clearly reproduces the error, the render tests just wont. Could this be a difference in render modes (still/continuous)?

jfirebaugh:
That's possible. It'll probably require C++ debugging to track down what's going on here. I'll try setting up the equivalent scenario in macosapp.

jfirebaugh:
@ivovandongen -- I figured out why the render test was passing. There were a variety of places in which the test harness called map.setZoom, triggering a recalculate that wouldn't normally happen. They are all unnecessary and could have obscured other issues, so I'm eliminating them in #6989, https://github.com/mapbox/mapbox-gl-test-suite/pull/164, and #6990. After those PRs merge, the test case should fail as expected.

As to why recalculation is necessary, it comes down to [this special logic](https://github.com/mapbox/mapbox-gl-native/blob/0e57d10d77e555a9229f7d522d83f87df7d5180d/src/mbgl/style/layers/symbol_layer_impl.cpp#L17-L20) for `text-size` / `icon-size`. Your fix does resolve the issue, but it would be good to find a fix that triggers recalculation only for these two properties. I'm not sure what the best way to do that is yet. Do you have any ideas?

ivovandongen:
@jfirebaugh Thanks for looking into this!

>Your fix does resolve the issue, but it would be good to find a fix that triggers recalculation only for these two properties. I'm not sure what the best way to do that is yet. Do you have any ideas?

No clean ideas yet. I see two options:
- Only trigger recalculation for SymbolLayers (in `Style::onLayerLayoutPropertyChanged`). Least impact, but not very efficient.
- Add an optional parameter to `LayerObserver::onLayerLayoutPropertyChanged()` to signal recalculation is needed. The symbol layer would set this when icon or text size is changed. Not very nicely separated though.

I think I would prefer the latter. I'll update the pr accordingly.

src/mbgl/style/layers/symbol_layer.cpp
```diff
@@ -169,7 +169,7 @@ void SymbolLayer::setIconSize(PropertyValue<float> value) {
     if (value == getIconSize())
         return;
     impl->layout.iconSize.set(value);
-    impl->observer->onLayerLayoutPropertyChanged(*this);
+    impl->observer->onLayerLayoutPropertyChanged(*this, true);
```
jfirebaugh: `symbol_layer.cpp` is a generated file, so this change will get overwritten.

Instead of a boolean, how about a `const char * property`, with name from style spec, e.g. `"icon-size"`. We can update `layer.cpp.ejs` to add this parameter for every setter, and do exact matches in `Style::onLayerLayoutPropertyChanged`.

ivovandongen: > symbol_layer.cpp is a generated file, so this change will get overwritten.

Of course.. I was a bit too hasty there.

> Instead of a boolean, how about a const char * property, with name from style spec

Makes sense, I'll start on that now.

ivovandongen:
@jfirebaugh I've updated the PR according to https://github.com/mapbox/mapbox-gl-native/pull/6932#discussion_r87482348

src/mbgl/style/style.cpp
```diff
@@ -572,9 +572,17 @@ void Style::onLayerPaintPropertyChanged(Layer&) {
...
 
-void Style::onLayerLayoutPropertyChanged(Layer& layer) {
+void Style::onLayerLayoutPropertyChanged(Layer& layer, const char * property) {
     layer.accept(QueueSourceReloadVisitor { updateBatch });
-    observer->onUpdate(Update::Layout);
+
+    auto update = Update::Layout;
+
+    //Recalculate the style for certain properties
+    bool needsRecalculation = strcmp(property, "icon-size") || strcmp(property, "text-size");
```
jfirebaugh: strcmp returns 0 if they are equal.

ivovandongen: Thanks again! Updated the PR and rebased on master.

jfirebaugh:
Last step is to merge https://github.com/mapbox/mapbox-gl-test-suite/pull/161 and then update the SHA in package.json.

PR #6933, opened 2016-11-06
Refactor Headless{Backend,Display}

brunoabinader:
In preparation for adding EGL backend support, I've made some changes in this PR:
- Remove all #ifdef's in both `HeadLess{Backend,Display}`, moving platform-specific code to their respective definition files e.g. `platform/linux/src/headless_display_glx.cpp` and adopting [pimpl](http://en.cppreference.com/w/cpp/language/pimpl) idiom for the platform-specific logic.
- Platform-specific code in `HeadlessBackend` now uses template-based member function `HeadlessDisplay::attribute()` to obtain platform-specific details from `HeadlessDisplay`.


platform/ios/uitest/ios-tests.xcodeproj/project.pbxproj
```diff
@@ -284,6 +284,9 @@
 				LastUpgradeCheck = 0700;
 				ORGANIZATIONNAME = Mapbox;
 				TargetAttributes = {
+					DD043322196DB9BC00E6F39D = {
+						DevelopmentTeam = GJZR2MEM28;
+					};
```
kkaefer: Please remove

platform/default/headless_backend_osmesa.cpp
```diff
@@ -1,46 +1,64 @@
...
 
+struct HeadlessBackend::Impl {
+    OSMesaContext glContext = nullptr;
+    GLubyte fakeBuffer = 0;
+};
+
+void HeadlessBackend::ImplDeleter::operator()(Impl* impl_) const {
+    OSMesaDestroyContext(impl_->glContext);
+    delete impl_;
+}
```
kkaefer: Instead of using custom deleters, we could also use subclasses and put them into the `unique_ptr`.

platform/darwin/src/headless_backend_cgl.cpp
```diff
@@ -21,36 +31,52 @@ gl::glProc HeadlessBackend::initializeExtension(const char* name) {
...
+    impl.reset();
 }
 
 void HeadlessBackend::activateContext() {
-    CGLError error = CGLSetCurrentContext(glContext);
+    assert(hasContext());
+    CGLError error = CGLSetCurrentContext(impl->glContext);
     if (error != kCGLNoError) {
-        throw std::runtime_error(std::string("Switching OpenGL context failed:") +
+        throw std::runtime_error(std::string("Switching OpenGL impl failed:") +
```
kkaefer: Why the wording change?

brunoabinader: Sorry, this went through on a batch string replace.

platform/default/headless_backend_osmesa.cpp
brunoabinader: I avoided using subclasses to reduce indirections, but I agree the overall result is more readable.

platform/ios/uitest/ios-tests.xcodeproj/project.pbxproj
brunoabinader: Argh, isn't that supposed to be in .gitignore?

brunoabinader:
Added https://github.com/mapbox/pixelmatch-cpp/issues/3 and https://github.com/mapbox/mapbox-gl-native/issues/6943 to track down the newly added Valgrind suppressions.

platform/ios/uitest/ios-tests.xcodeproj/project.pbxproj
incanus: You need to select a code sign identity to run on device, and then it gets saved to the `pbxproj` like this. I'd recommend either `git add -p` avoiding it (what I do) or a commit script that removes it manually somehow maybe?

brunoabinader: Sounds good @incanus - yeah, if those are the only options we have atm then we can come up with some [pre-commit script hook](http://githooks.com/) in git that automatically removes these changes.

brunoabinader:
Requested changes are now in - please review @kkaefer @mikemorris 👍

platform/linux/config.cmake
```diff
@@ -1,5 +1,7 @@
 mason_use(glfw VERSION 3.2.1)
-mason_use(mesa VERSION 13.0.0${MASON_CXXABI_SUFFIX})
+if (WITH_OSMESA OR IS_CI_BUILD)
+    mason_use(mesa VERSION 13.0.0${MASON_MESA_SUFFIX}${MASON_CXXABI_SUFFIX})
```
kkaefer: Can we just always set `WITH_OSMESA` for all CI builds?

Issue #6861, opened 2016-10-31
Adding a new MGLSource subclass that implements datasource pattern for fetching tiles on iOS


JesseCrocker:
Im considering implementing a new subclass of MGLSource(and mbgl::style::Source) that uses the datasource pattern for getting tiles into the map. The datasource would have a reference to an object that implements an interface with a single method, something like `- (void)getTile:(NSUInteger)z x:(NSUInteger)x y:(NSUInteger)y callback(returnType (^)(NSData *tileData))callback`

Why i think i need this: my application needs to display data that is not in vector tile or geojson format, but it's too much data to just add it all to the map as annotations. My solution to this is to generate vector tiles from the data on the fly in the application, but currently the only way I can see to get the tiles onto the map is to run an http server in my app, and add a source pointing at the local web server. Running a local server works, but it seems like a weird hack, and likely to have odd bugs or performance issues.

Im also interested in implementing the same approach for raster tiles, to enable displaying dynamically generated rasters, but that would come after vectors.

Does this seem like a reasonable approach?
Would a pull request that implements this be considered for merging?

incanus:
Some prior art here in https://github.com/mapbox/mapbox-gl-native/issues/507.

1ec5:
Mapbox GL JS has an analogous feature in progress called “custom sources”: mapbox/mapbox-gl-js#2920.

incanus:
I do like this idea. I say go for it @JesseCrocker 👍

kkaefer:
We already have the concept of different source types in our core code. Right now, there are three basic ways to get vector data from sources: via regular tiles, via annotations, and via GeoJSON. All of these fundamentally return `GeometryTile`-derived objects that hold a representation of the geometry in a [`geometry.hpp`-compatible](https://github.com/mapbox/geometry.hpp) format.

However, before you implement a completely custom source, you should consider using a GeoJSON source. Either convert your existing geometry to GeoJSON data, or if that is not possible, a shortcut may be to use the `GeoJSONSource`'s `setGeoJSON()` member function. It accepts a parameter of type `mapbox::geojson::geojson`, which is `mapbox::util::variant<geometry, feature, feature_collection>`, essentially what you need anyway to be compatible with the internal geometry represenation that `GeometryTile`'s use. This is exposed on iOS in the `MGLGeoJSONSource` interface: You can initialize a class of that type with `MGLFeature` objects that hold your geometry information.

JesseCrocker:
[...]

Issue #119, opened 2014-03-26
non-solid matte


incanus:
We should consider a non-solid matte for better context. Consider Apple Maps's grid: 

https://dl.dropboxusercontent.com/u/575564/matte.mov

![matte](https://f.cloud.github.com/assets/17722/2520156/93393a78-b482-11e3-9109-be581d8a0efc.png)

soupman99:
+1

1ec5:
This feature is important because, especially in a full-screen-map application on iOS, the user would perceive a hang between the time the style is set and the time content starts to appear. If they’re in the midst of a gesture, the effect can be especially jarring because the user may end up panning or zooming farther than they expected to.

In order for this feature to match the user experience provided by MapKit, the grid pattern would have to be independent of any style. With respect to any given tile coordinate, it would have to appear from the moment the style is set (before the style is downloaded or parsed) to the moment any non-background layer appears at that tile coordinate. Either this pattern would appear above any background layer, or the background layer would automatically remain hidden (exposing the pattern beneath it) until a non-background layer can appear.

riastrad:
Sorry @stale, but this is still a sometimes requested feature - so I think it's worth revisiting whether or not this would be possible to include now that some time has passed.

ulusoyca:
This issue was referenced in https://github.com/mapbox/mapbox-gl-native/issues/11814 
I agree with @riastrad that we need this feature to distinguish background layer and unloaded tile visually

tobrun:
GL-native supports the same UX as GL-JS shown in https://github.com/mapbox/mapbox-gl-native/issues/119#issuecomment-44676773. This is a better UX as showing a grid or nothing at all. @ulusoyca could you elaborate why this is needed for your use-case?

ulusoyca:
@tobrun Correct me if I am wrong but the comment you pointed assumes that there are preloaded tiles. In my use case, there is no guarantee of preloaded tiles so it makes difficult to distinguish unloaded tile from background layer

tobrun:
that is correct, you will just see the background color. One way to achieve your effect, haven't verified this but should work, would be to updating your style with:
 - adding a fill layer above background layer with current background color and linking a geometry spanning the world.
 - updating background layer to use [background-pattern](https://docs.mapbox.com/mapbox-gl-js/style-spec/#paint-background-background-pattern) with something as a grid shown in OP

ulusoyca:
@tobrun 
[...]

PR #6940, opened 2016-11-07
Computed (on-demand) shape source

JesseCrocker:
Implements new API proposed in #6861 for a custom vector source, that queries a class supplied by the client application for the features to display on a tile.

This defines a new class in core,  `mbgl::style::CustomVectorSource`, and a corresponding new source class for iOS/MacOS `MGLCustomVectorSourceDataSource`. There is also an new interface defined i obj-c, `MGLCustomVectorSourceDataSource` that defines a single method `- (void)getTileForZoom:(NSInteger)zoom x:(NSInteger)x y:(NSInteger)y callback:( void (^)(NSArray<id <MGLFeature>>*) )callback`

I see 4 uses for this:
1. Data that is generated dynamically. A trivial example of this would be a lat/lon grid, that data is very simple, but a grid with a resolution of 0.01 degrees contains 216000 lines.
2. Data that is too large to fit in memory. If data has a bounded size, it can be used as a GeoJSON sources, but if the data can be arbitrarily large converting it to GeoJSON may not be possible, for example shapefiles.
3. Data that is fetched from the network, but not as Mapbox Vector Tiles. With this API you could implement an OSM editor, or a WFS client.
4. Data that already exists in a format with some sort of spatial index, where querying data for a single tile is faster than fetching all of the data. For example, a set of POI data stored in an SQLite database.

My use case for this is displaying GPS traces, which is #2 and #4 from the above list. I develop an application where users frequently have hundreds of GPS traces, but occasionally have tens of thousands, and some of them have hundreds of thousands of points per trace. The data can also be modified, either by the user, or by a background sync process, in either case the data will need to be redisplayed.


platform/darwin/src/MGLCustomVectorSource.h
```diff
@@ -0,0 +1,60 @@
...
+@protocol MGLCustomVectorSourceDataSource <NSObject>
+
+/** 
+ Fetch data for a tile
+ @param zoom
+ @param y
+ @param x
+ @param callback A block to call with the data that has been fetched for the tile.
+ */
+- (void)getTileForZoom:(NSInteger)zoom x:(NSInteger)x y:(NSInteger)y callback:( void (^)(NSArray<id <MGLFeature>>*) )callback;
```
1ec5: Rename "zoom" to "zoomLevel" for consistency with other SDK APIs. Should x and y be unsigned? Rename "callback" to "completionHandler" for consistency.

platform/darwin/src/MGLCustomVectorSource.h
```diff
@@ -0,0 +1,60 @@
...
+ @param identifier A string that uniquely identifies the source.
+ @param dataSource An object that implements the `MGLCustomVectorSourceDataSource` protocol that will be queried for tile data.
+ @param options An `NSDictionary` of options for this source.
+ */
+- (instancetype)initWithIdentifier:(NSString *)identifier dataSource:(NSObject<MGLCustomVectorSourceDataSource>*)dataSource options:(NS_DICTIONARY_OF(NSString *, id) *)options NS_DESIGNATED_INITIALIZER;
+
+/**
+ Request that the source reloads a tile. Tile will only be reloaded if it is currently on screen.
+ */
+- (void)updateTile:(NSInteger)z x:(NSInteger)x y:(NSInteger)y;
```
1ec5: Rename to `-setNeedsUpdateAtZoomLevel:x:y:`.

platform/darwin/src/MGLGeoJSONSourceBase.h
```diff
@@ -0,0 +1,63 @@
...
+
+/**
+ An `NSNumber` object containing a double; specifies the Douglas-Peucker
+ simplification tolerance. A greater value produces simpler geometries and
+ improves performance. The default value is 0.375.
+ */
+extern const MGLGeoJSONSourceOption MGLGeoJSONSourceOptionSimplificationTolerance;
+
+
+@interface MGLGeoJSONSourceBase : MGLSource
```
1ec5: This is an unfortunate name for a class. Maybe someone else has an idea for what to call it. Also, it seems like it should declare some initializers and properties related to the options declared above.

src/mbgl/style/sources/custom_vector_source.cpp
```diff
@@ -0,0 +1,19 @@
...
+
+namespace mbgl {
+  namespace style {
+    
+    CustomVectorSource::CustomVectorSource(std::string id, GeoJSONOptions options, std::function<void(uint8_t, uint32_t, uint32_t)> fetchTile)
+    : Source(SourceType::Vector, std::make_unique<CustomVectorSource::Impl>(std::move(id), *this,  options, fetchTile)),
+    impl(static_cast<Impl*>(baseImpl.get())) { }
+    
+    void CustomVectorSource::setTileData(uint8_t z, uint32_t x, uint32_t y, const mapbox::geojson::geojson& geoJSON) {
+      impl->setTileData(z, x, y, geoJSON);
```
1ec5: Indent by four spaces.

platform/darwin/src/MGLGeoJSONSourceBase.mm
```diff
@@ -0,0 +1,74 @@
...
+const MGLGeoJSONSourceOption MGLGeoJSONSourceOptionMinimumZoomLevel = @"MGLGeoJSONSourceOptionMinimumZoomLevel";
+const MGLGeoJSONSourceOption MGLGeoJSONSourceOptionBuffer = @"MGLGeoJSONSourceOptionBuffer";
+const MGLGeoJSONSourceOption MGLGeoJSONSourceOptionSimplificationTolerance = @"MGLGeoJSONSourceOptionSimplificationTolerance";
+
+@implementation MGLGeoJSONSourceBase
+
+
+- (instancetype)initWithIdentifier:(NSString *)identifier options:(NS_DICTIONARY_OF(NSString *, id) *)options
+{
+  if (self = [super initWithIdentifier:identifier]) {
```
1ec5: Use four spaces per tab stop.

platform/darwin/src/MGLCustomVectorSource.h
```diff
@@ -0,0 +1,65 @@
...
+ Data source for `MGLCustomVectorSource`.
+ */
+@protocol MGLCustomVectorSourceDataSource <NSObject>
+
+/** 
+ Fetch data for a tile
+ @param zoom
+ @param y
+ @param x
+ @param callback A block to call with the data that has been fetched for the tile.
```
1ec5: Fill in this documentation and rename `callback`. Point out that `completionHandler` must be called. Incidentally, is it absolutely necessary to make this API asynchronous? A synchronous data source method would be far less error-prone.

platform/darwin/src/MGLCustomVectorSource.h
```diff
@@ -0,0 +1,65 @@
...
+ */
+- (void)getTileForZoomLevel:(NSUInteger)zoomLevel x:(NSUInteger)x y:(NSUInteger)y completionHandler:( void (^)(NSArray<id <MGLFeature>>*) )completionHandler;
+
+@end
+
+/**
+ A source for vector data that is fetched 1 tile at a time. Usefull for sources that are
```
1ec5: s/Usefull/Useful/

platform/darwin/src/MGLCustomVectorSource.h
```diff
@@ -0,0 +1,65 @@
...
+
+/**
+ Request that the source reloads a tile. Tile will only be reloaded if it is currently on screen.
+ */
+- (void)setNeedsUpdateAtZoomLevel:(NSUInteger)z x:(NSUInteger)x y:(NSUInteger)y;
+
+/**
+ Reload all tiles.
+ */
+- (void)reload;
```
1ec5: Rename to `-reloadData` for consistency with Cocoa classes like UITableView.

platform/darwin/src/MGLCustomVectorSource.h
```diff
@@ -0,0 +1,65 @@
...
+ @param identifier A string that uniquely identifies the source.
+ @param dataSource An object that implements the `MGLCustomVectorSourceDataSource` protocol that will be queried for tile data.
+ @param options An `NSDictionary` of options for this source.
+ */
+- (instancetype)initWithIdentifier:(NSString *)identifier dataSource:(NSObject<MGLCustomVectorSourceDataSource>*)dataSource options:(NS_DICTIONARY_OF(NSString *, id) *)options NS_DESIGNATED_INITIALIZER;
+
+/**
+ Request that the source reloads a tile. Tile will only be reloaded if it is currently on screen.
+ */
+- (void)setNeedsUpdateAtZoomLevel:(NSUInteger)z x:(NSUInteger)x y:(NSUInteger)y;
```
1ec5: It isn’t usually obvious to the developer what to pass in as `x` and `y`. Would it be possible to rework this method to take an `MGLCoordinateBounds`, à la `-[MKOverlayRenderer setNeedsDisplayInMapRect:zoomScale:]`?

platform/darwin/src/MGLGeoJSONSourceBase_Private.h
```diff
@@ -0,0 +1,13 @@
+#import "MGLGeoJSONSource.h"
+
+#include <mbgl/style/sources/geojson_source.hpp>
+
+@interface MGLGeoJSONSourceBase (Private)
+
+@property (nonatomic, readwrite) NSDictionary *options;
+
+- (instancetype)initWithIdentifier:(NSString *)identifier options:(NS_DICTIONARY_OF(NSString *, id) *)options;
```
1ec5: Should this initializer be public (with a warning not to call it directly), as with `-[MGLSource initWithIdentifier:]`?

1ec5: The options dictionary’s keys should be of type `MGLGeoJSONSourceOption`.

platform/ios/app/MBXViewController.m
```diff
@@ -332,6 +334,7 @@ - (void)dismissSettings:(__unused id)sender
                 @"Style Vector Source",
                 @"Style Raster Source",
                 [NSString stringWithFormat:@"Label Countries in %@", (_usingLocaleBasedCountryLabels ? @"Local Language" : [[NSLocale currentLocale] displayNameForKey:NSLocaleIdentifier value:[self bestLanguageForUser]])],
+                @"Add Custom Lat/Lon Grid"
```
1ec5: Ooh, nice, a graticule!

platform/ios/ios.xcodeproj/project.pbxproj
```diff
@@ -409,13 +417,6 @@
 /* End PBXBuildFile section */
 
 /* Begin PBXContainerItemProxy section */
-		DA25D5C71CCDA0C100607828 /* PBXContainerItemProxy */ = {
-			isa = PBXContainerItemProxy;
-			containerPortal = DA1DC9421CB6C1C2006E619F /* Project object */;
-			proxyType = 1;
-			remoteGlobalIDString = DA25D5B81CCD9EDE00607828;
-			remoteInfo = settings;
-		};
```
1ec5: Revert this change. The code signing issue was fixed in #6948.

platform/ios/ios.xcodeproj/project.pbxproj
```diff
@@ -2061,11 +2078,6 @@
 /* End PBXSourcesBuildPhase section */
 
 /* Begin PBXTargetDependency section */
-		DA25D5C81CCDA0C100607828 /* PBXTargetDependency */ = {
```
1ec5: Revert this change too.

platform/ios/src/Mapbox.h
```diff
@@ -53,3 +53,4 @@ FOUNDATION_EXPORT const unsigned char MapboxVersionString[];
 #import "NSValue+MGLAdditions.h"
 #import "MGLStyleValue.h"
 #import "MGLTileSet.h"
+#import "MGLCustomVectorSource.h"
```
1ec5: Also import MGLGeoJSONSourceBase.h.

platform/darwin/src/MGLCustomVectorSource.h
1ec5: On second thought, maybe this should be called `-reloadTileInCoordinateBounds:zoomLevel:]`, for consistency with `-reloadData`.

platform/darwin/src/MGLCustomVectorSource.h
```diff
@@ -0,0 +1,65 @@
...
+
+/**
+ Reload all tiles.
+ */
+- (void)reload;
+
+/**
+ An object that implements the `MGLCustomVectorSourceDataSource` protocol that will be queried for tile data.
+ */
+@property (nonatomic, readonly, copy) NSObject<MGLCustomVectorSourceDataSource> *dataSource;
```
1ec5: Properties that refer to data sources should be weak to avoid a retain cycle, should not copy, and should be `id` rather than NSObject. Is there a reason it’s read-only? It’s unusual for a data source to be used on a custom queue.

platform/darwin/src/MGLCustomVectorSource.h
JesseCrocker: This doesn't need to be async, since it's called on it's own NSOperationQueue. But it definitely would not work to have be it be synchronous on the main queue. Should i change this to be synchronous?

platform/darwin/src/MGLCustomVectorSource.h
```diff
@@ -0,0 +1,65 @@
...
+ */
+@protocol MGLCustomVectorSourceDataSource <NSObject>
+
+/** 
+ Fetch data for a tile
+ @param zoom
+ @param y
+ @param x
+ */
+- (NSArray<id <MGLFeature>>*)getTileForZoomLevel:(NSUInteger)zoomLevel x:(NSUInteger)x y:(NSUInteger)y;
```
1ec5: Now that this is a simple getter, it should be called `-featuresInTileAtX:y:zoomLevel:`, without the “get”.

I think it would be just as generally useful to have a `-featuresInCoordinateBounds:zoomLevel:`, per https://github.com/mapbox/mapbox-gl-native/pull/6940#discussion_r87365181, to accommodate those migrating from MKOverlayRenderer. If you think it’s worth keeping a tile coordinate–based API around, perhaps both methods could be optional methods on this protocol. MGLCustomVectorSource would end up calling one or the other, depending on which one is implemented.

platform/darwin/src/MGLCustomVectorSource.h
```diff
@@ -0,0 +1,65 @@
...
+ Returns a custom vector datasource initialized with an identifier, datasource,  and a
+ dictionary of options for the source according to the
+ <a href="https://www.mapbox.com/mapbox-gl-style-spec/#sources-geojson">style
+ specification</a>.
+
+ @param identifier A string that uniquely identifies the source.
+ @param dataSource An object that implements the `MGLCustomVectorSourceDataSource` protocol that will be queried for tile data.
+ @param options An `NSDictionary` of options for this source.
+ */
+- (instancetype)initWithIdentifier:(NSString *)identifier dataSource:(NSObject<MGLCustomVectorSourceDataSource>*)dataSource options:(NS_DICTIONARY_OF(MGLGeoJSONSourceOption, id) *)options NS_DESIGNATED_INITIALIZER;
```
1ec5: In Cocoa, data sources aren’t normally required; when absent, the class that normally uses the data source assumes no data is available. By following that model here, it’ll be possible to set up a custom vector source before setting up its data source, which may be handy if the data source takes some time to set up. This would entail removing the `dataSource:` parameter from this method and making the `dataSource` property nullable.

platform/darwin/src/MGLCustomVectorSource.h
```diff
@@ -0,0 +1,65 @@
...
+/**
+ Data source for `MGLCustomVectorSource`.
+ */
+@protocol MGLCustomVectorSourceDataSource <NSObject>
+
+/** 
+ Fetch data for a tile
+ @param zoom
+ @param y
+ @param x
```
1ec5: When filling in the documentation, make sure to mention that this method is always called on the operation queue in the `requestQueue` property.

JesseCrocker:
@1ec5 Thanks for working with me on this, I know it's added work for you that wasn't on the roadmap.
So far I'm still at less than half the time I estimated this was going to take to implement. For a codebase of this size, it's been really easy to get started on.

One issue i'm running into is when I run `make iproj` the mbgl target in the generated workspace doesn't include the new files i've added to core. Do I need to put those somewhere?

JesseCrocker:
@1ec5 to Implement featuresInCoordinateBounds:zoomLevel: It needs to find the bounds for a tile, it looks like mbgl::LatLngBounds already has a constructor to create a LatLngBounds from a CanonicalTileID, but mbgl/tile/tile_id.hpp can't be included in any of the obj-c++ classes, because it includes boost/functional/hash.hpp, and boost isn't in the search path for the iOS targets. Is it OK to add boost to the header search path for the iOS targets?

1ec5:
> One issue i'm running into is when I run make iproj the mbgl target in the generated workspace doesn't include the new files i've added to core. Do I need to put those somewhere?

Yes, after adding the headers to core, run `./scripts/generate-core-files.sh`.

1ec5:
> Is it OK to add boost to the header search path for the iOS targets?

Yes, I think it would be OK, although we’ll want to watch the SDK for any significant size increase as a result. To add Boost, [add it as a Mason package](https://github.com/mapbox/mapbox-gl-native/blob/dbd22357441ec66c46cc7e46a01ed114a18f685c/platform/ios/config.cmake#L49) like we do [in core](https://github.com/mapbox/mapbox-gl-native/blob/dbd22357441ec66c46cc7e46a01ed114a18f685c/cmake/core.cmake#L41). Be sure to [update the macOS configuration](https://github.com/mapbox/mapbox-gl-native/blob/dbd22357441ec66c46cc7e46a01ed114a18f685c/platform/macos/config.cmake#L49) too.

1ec5:
The parts of tile_id.hpp that depend on Boost are the three template specializations of `hash` at the bottom. Maybe they could be split into a separate header; that would enable the SDK to include tile_id.hpp without including Boost.

1ec5:
As I’ve mentioned before, we need better names for MGLCustomVectorSource and MGLGeoJSONSourceBase, especially the latter. Meanwhile, #7160 proposes renaming MGLGeoJSONSource to MGLFeatureSource, since most use cases for MGLGeoJSONSource don’t even involve actual GeoJSON source code.

So, two proposals:

* MGLGeoJSONSourceBase → MGLFeatureSource<br>MGLGeoJSONSource stays the same<br>MGLCustomVectorSource → MGLComputedSource<br>MGLCustomVectorSourceDataSource → MGLComputedSourceDataSource
* MGLGeoJSONSourceBase → MGLFeatureSource<br>MGLGeoJSONSource → MGLCollectionSource (given #7160 and #7145)<br>MGLCustomVectorSource → MGLComputedSource<br>MGLCustomVectorSourceDataSource → MGLComputedSourceDataSource

/cc @incanus

1ec5:
@JesseCrocker, I just wanted to give an update around the runtime styling taxonomy:

* #7334 (on the release branch) renamed MGLGeoJSONSource to MGLShapeSource.
* #7360 will give MGLRasterSource and MGLVectorSource a new abstract superclass, MGLTileSource.
* #7250 (on the release branch) added a new MGLOpenGLStyleLayer that MGLStyle holds onto. Developers can subclass MGLOpenGLStyleLayer to implement their own custom OpenGL drawing code. This is our (unsupported) solution for custom style layers.

If you aren’t tied to the data source pattern, #7250 points to the possibility of a source class that developers can subclass directly, without having to define a data source elsewhere. Unlike the other sources, custom vector sources always come from application code, never the style JSON. So that means MGLStyle can own the source and know that no other custom vector source is lurking in the style. That neatly avoids naming a protocol with “Source…Source”. 😉 Please correct me if I’m missing a consideration that makes the data source pattern more desirable for this purpose.

We’d like to avoid a backwards incompatible change in v3.5.0, if possible, so this would be a good opportunity to plan ahead and make sure we’re really happy with the renaming in #7334. If MGLShapeSource stays as it is in v3.4.0, I’d suggest renaming the following classes the next time we merge the release-ios-v3.4.0 branch back to master:

* MGLCustomVectorSource → MGLComputedShapeSource
* MGLGeoJSONSourceBase → MGLAbstractShapeSource

This would reserve the “vector” terminology for vector tiles, in case we support programmatically generated vector tile sources in the future, and avoid the word “custom”, which means different things in different contexts. I’m not a huge fan of the word “abstract” in a class name, but I think it reads better than “base”.

You’re probably sick of hearing me write about naming, but there are a lot of moving parts, and I wanted to give you a chance to carve out some space for your feature in the API that’s forming around runtime styling.

/cc @boundsj @frederoni @ericrwolfe

platform/darwin/src/MGLAbstractShapeSource.h
```diff
@@ -0,0 +1,87 @@
...
+ 
+ This attribute corresponds to the
+ <a href="https://www.mapbox.com/mapbox-gl-style-spec/#sources-geojson-tolerance"><code>tolerance</code></a>
+ source property in the Mapbox Style Specification.
+ */
+extern MGL_EXPORT const MGLShapeSourceOption MGLShapeSourceOptionSimplificationTolerance;
+
+
+
+@interface MGLAbstractShapeSource : MGLSource
```
1ec5: Due to a change in the framework's visibility settings, aimed at reducing the framework's size, all publicly visible classes must now be declared with the MGL_EXPORT macro. See CONTRIBUTING.md for details. (There's a make rule you can run to scan for missing exports.)

platform/ios/ios.xcodeproj/project.pbxproj
```diff
@@ -1503,6 +1521,7 @@
 				DD0902AB1DB192A800C5BDCE /* MGLNetworkConfiguration.h in Headers */,
 				DA8848571CBAFB9800AB86E3 /* MGLMapboxEvents.h in Headers */,
 				DA8848311CBAFA6200AB86E3 /* NSString+MGLAdditions.h in Headers */,
+				88B079A61E363A7200834FAB /* MGLAbstractShapeSource.h in Headers */,
```
1ec5: Corresponding changes are also needed in macos.xcodeproj

platform/ios/src/Mapbox.h
```diff
@@ -48,6 +48,8 @@ FOUNDATION_EXPORT MGL_EXPORT const unsigned char MapboxVersionString[];
 #import "MGLTileSource.h"
 #import "MGLVectorSource.h"
 #import "MGLShapeSource.h"
+#import "MGLAbstractShapeSource.h"
+#import "MGLComputedShapeSource.h"
```
1ec5: Corresponding changes are also needed in platform/macos/src/Mapbox.h.

platform/darwin/src/MGLAbstractShapeSource.h
1ec5: This class needs a documentation comment. As with the other abstract classes (such as MGLSource and MGLVectorStyleLayer), be sure to point out that this class shouldn't be initialized directly.

platform/macos/macos.xcodeproj/project.pbxproj
```diff
@@ -316,6 +319,9 @@
 		558F18211D0B13B000123F46 /* libmbgl-loop.a */ = {isa = PBXFileReference; lastKnownFileType = archive.ar; name = "libmbgl-loop.a"; path = "../../build/osx/Debug/libmbgl-loop.a"; sourceTree = "<group>"; };
 		55D9B4B01D005D3900C1CCE2 /* libz.tbd */ = {isa = PBXFileReference; lastKnownFileType = "sourcecode.text-based-dylib-definition"; name = libz.tbd; path = usr/lib/libz.tbd; sourceTree = SDKROOT; };
 		55FE0E8D1D100A0900FD240B /* config.xcconfig */ = {isa = PBXFileReference; lastKnownFileType = text.xcconfig; name = config.xcconfig; path = ../../build/macos/config.xcconfig; sourceTree = "<group>"; };
+		88B079A81E3793E000834FAB /* MGLAbstractShapeSource_Private.h */ = {isa = PBXFileReference; fileEncoding = 4; lastKnownFileType = sourcecode.c.h; path = MGLAbstractShapeSource_Private.h; sourceTree = "<group>"; };
+		88B079A91E3793E000834FAB /* MGLAbstractShapeSource.h */ = {isa = PBXFileReference; fileEncoding = 4; lastKnownFileType = sourcecode.c.h; path = MGLAbstractShapeSource.h; sourceTree = "<group>"; };
```
frederoni: This one should also be marked as public

1ec5:
The change to the .mason submodule needs to be reverted to resolve the merge conflict.

1ec5:
Have you looked into adding automated tests of this feature? Here are some examples of mbgl unit tests related to sources:

https://github.com/mapbox/mapbox-gl-native/blob/master/test/style/tile_source.test.cpp
https://github.com/mapbox/mapbox-gl-native/blob/master/test/style/source.test.cpp

jonkan:
First of all, great PR! I'm really looking forward seeing this feature in mapbox!
However, I'm seeing a problem that when reloading a `MGLComputedShapeSource`, the map doesn't reload all zoom levels, i.e. the source data doesn't get fully invalidated and re-rendered. 

Steps:
- Render a feature using the `MGLComputedShapeSource`.
- Zoom the map in and out a bit (to get it rendered on multiple zoom levels). 
- Modify the feature geometry and call `reloadData` on the `MGLComputedShapeSource`.
(The map gets updated, showing the updated feature geometry in the current view)
- Zoom the map in or out. The map frequently shows the previous feature geometry on previously rendered zoom levels. 

If Instead of calling `reloadData`, re-creating the source (removing the old and adding the new to the style) the map reloads as expected across all zoom levels.

Also, I'm not seeing this behaviour with the `MGLShapeSource` when updating features by re-setting the `shape` property.

(Testing this I had merged this PR 517f44783 into master b35ca2099)

src/mbgl/style/sources/custom_vector_source_impl.cpp
```diff
@@ -0,0 +1,99 @@
...
+}
+ 
+void CustomVectorSource::Impl::reloadRegion(mbgl::LatLngBounds bounds, uint8_t z) {
+    for (const auto& tile : mbgl::util::tileCover(bounds, z)) {
+        updateTile(tile.canonical.z, tile.canonical.x, tile.canonical.z);
+    }
+}
+
+void CustomVectorSource::Impl::reload() {
+    for (auto const &item : tiles) {
```
jonkan: Should there be a `cache.clear();` here? Seems to solve my issue with the map showing previously rendered features after reloading the source.

JesseCrocker: Thanks for catching that, there should be a cache.clear there

JesseCrocker:
I've added a couple of tests for this, both on the obj-c side and the c++ side, but i dont think they are really testing anything useful. Any guidance as to how to test this, especially the core part of it, would be greatly appreciated.

JesseCrocker:
After doing some more testing with this, I turned up memory corruption when a source was removed, and a tile fetch was in progress. My last commit should fix that.

platform/darwin/test/MGLComputedShapeSourceTests.m
```diff
@@ -0,0 +1,25 @@
...
+#import <Mapbox/Mapbox.h>
+
+
+@interface MGLComputedShapeSourceTests : XCTestCase
+@end
+
+@implementation MGLComputedShapeSourceTests
+
+- (void)testInitializer {
+  MGLComputedShapeSource *source = [[MGLComputedShapeSource alloc] initWithIdentifier:@"id"
```
1ec5: Nit: use four-column-wide tabs for consistency.

platform/darwin/src/MGLAbstractShapeSource.h
```diff
@@ -0,0 +1,99 @@
...
+ `MGLAbstractShapeSource` is an abstract base class for map content sources that 
+ supply vector shapes to be shown on the map. A shape source is added to an
+ `MGLStyle` object along with an `MGLVectorStyleLayer` object. The vector style
+ layer defines the appearance of any content supplied by the shape source.
+
+ 
+ Do not create instances of this class directly, and do not create your own
+ subclasses of this class. Instead, create instances of `MGLShapeSource` or 
+ `MGLComputedShapeSource`.
+ */
```
1ec5: By the way, this is totally optional, but we have a mechanism for [inserting a Swift code example](https://github.com/mapbox/mapbox-gl-native/blob/8742dc852c82ef40003a22a8e0a764616d8beed1/platform/ios/DEVELOPING.md#adding-a-code-example) into a documentation comment via unit tests. Feel free to use MGLShapeSource’s documentation as a model.

JesseCrocker:
Im still working on fixing a use-after-free issue with this.
How to trigger the problem:
Add an MGLComputedShapeSource to the style, with a dataSource that takes some time to return data.
Call `[mapView setStyleUrl:]` to change the map style

If any calls to the data source were in progress when the style was changed, they will end up calling `setTileData` on a `mbgl::style::CustomVectorSource` that has been deallocated.

Address sanitizer picks up the use after free, but without it turned on this manifests as all sort of random crashes from memory corruption.

I've got a good test case for this at https://github.com/trailbehind/mapbox-gl-native/commit/1f5f51e9999199c5ce6ed6589aa9670a29c2e6f3 , It calls `cycleStyles:`on a timer, and adds a MGLComputedShapeSource to the style in `mapView:didFinishLoadingStyle:`

I see 2 different ways to look at this problem:
1. `dealloc` doesn't get called on MGLComputedShapeSource if there are any fetch operations in progress, because the operation has a reference to the MGLComputedShapeSource. If dealloc did get called, the MGLComputedShapeSourceFetchOperation would get canceled, and `self.source.rawSource->setTileData` would not get called.
2. MGLComputedShapeSource has no way to know when its `mbgl::style::CustomVectorSource` has been dealloced, because `removeFromMapView:` does not get called, and `MGLComputedShapeSource` only has a raw pointer to the `mbgl::style::CustomVectorSource`

Fixing based on 2 seems like the better fix, but it seems like a big change that would be beyond the scope of this PR. 
2 ways i would think it could be fixed based on 2:
* Call removeFromMapView: when a style is dealloced.
* Switch to use shared_ptr instead of unique_ptr for sources

So im guessing fixing based on 1 is the way to go, but im having trouble getting it to work right.

There's also a third option, let this crash, and add a note to the documentation that the source must be removed from the style before the style is removed from the map.

@1ec5 Do you have any suggestion for how to approach this?

JesseCrocker:
I've got the crash mentioned above fixed by removing the reference to the MGLComputedShapeSource from the MGLComputedShapeSourceFetchOperation

test/style/custom_vector_source.test.cpp
```diff
@@ -0,0 +1,53 @@
...
+
+using namespace mbgl;
+
+
+TEST(CustomVectorSource, EmptyData) {
+    mbgl::style::GeoJSONOptions options;
+
+#pragma clang diagnostic push
+#pragma clang diagnostic ignored "-Wunused-parameter"
+    const auto callback = ^void(uint8_t z, uint32_t x, uint32_t y) {};
```
kkaefer: Can we instead use C++11 lambdas rather than Apple-Clang specific syntax? You can also remove the z/x/y names to avoid the pragmas. An even better approach would be to pass a null pointer as the callback and check whether `callback` is invokable before calling it.

src/mbgl/tile/tile_id_hash.hpp
```diff
@@ -0,0 +1,37 @@
...
+    size_t operator()(const mbgl::OverscaledTileID &id) const {
+        std::size_t seed = 0;
+        boost::hash_combine(seed, std::hash<mbgl::CanonicalTileID>{}(id.canonical));
+        boost::hash_combine(seed, id.overscaledZ);
+        return seed;
+    }
+};
+
+
+} // namespace std
```
kkaefer: Why was this moved to a separate file? While not wrong, it doesn't look like it's necessary to include in this PR.

platform/macos/app/MapDocument.m
```diff
@@ -1095,6 +1139,53 @@ - (CGFloat)mapView:(MGLMapView *)mapView alphaForShapeAnnotation:(MGLShape *)ann
...
+        [features addObject:feature];
+    }
+    
+    for (double x = floor(bounds.sw.longitude / gridSpacing) * gridSpacing; x <= ceil(bounds.ne.longitude / gridSpacing) * gridSpacing; x += gridSpacing) {
+        coords[0] = CLLocationCoordinate2DMake(bounds.sw.latitude, x);
+        coords[1] = CLLocationCoordinate2DMake(bounds.ne.latitude, x);
+        MGLPolylineFeature *feature = [MGLPolylineFeature polylineWithCoordinates:coords count:2];
+        feature.attributes = @{@"value": @(x)};
+        [features addObject:feature];
+    }
```
kkaefer: Is this entire function something that should be in the core code rather than in a binding-specific code?

include/mbgl/style/sources/geojson_source.hpp
```diff
@@ -26,7 +27,10 @@ using SuperclusterPointer = std::unique_ptr<mapbox::supercluster::Supercluster>;
 
 struct GeoJSONOptions {
     // GeoJSON-VT options
+    uint8_t minzoom = 0;
+
```
kkaefer: space

include/mbgl/style/sources/custom_vector_source.hpp
```diff
@@ -0,0 +1,31 @@
...
+
+namespace mbgl {
+namespace style {
+    
+class CustomVectorSource : public Source {
+public:
+    CustomVectorSource(std::string id, GeoJSONOptions options, std::function<void(uint8_t, uint32_t, uint32_t)> fetchTile);
+  
+    void setTileData(uint8_t, uint32_t, uint32_t, const mapbox::geojson::geojson&);
+    void updateTile(uint8_t, uint32_t, uint32_t);
```
kkaefer: What is the difference between `setTileData` and `updateTile`? If they're distinct, can we add some documentation?

src/mbgl/tile/tile_id_hash.hpp
JesseCrocker: It was split into a separate file so that tile_id.hpp could be included in objective-c++ files in the iOS/macOS side of things without having to include boost.

1ec5: https://github.com/mapbox/mapbox-gl-native/pull/6940#issuecomment-262400266 👍

platform/macos/app/MapDocument.m
1ec5: You mean as an implementation of #119? I don’t think it satisfies the requirements for that feature, namely that the graticule would have to be visible before the style even starts to load. But it does make for an interesting demo.

JesseCrocker:
@kkaefer Thanks for reviewing this. I think i've addressed all of your comments. Sorry about the large number of commits, i had to do some iteration to make this work, and then there were a bunch of commits renaming things based on review feedback.

Im well versed in Android development and capable of doing the Android bindings for this, but probably wont have time to get to it for a couple months.

1ec5:
#8473 is a squashed, reorganized version of this PR.

ychescale9:
Wondering if there's any progress (or interests from other users) on this?
We are currently looking at using MBTiles in **Android** (offline) and evaluating different options.
Hence we would like to know if adding support for MBTiles source is on the roadmap. Thanks!

jfirebaugh:
→ #8473.

PR #6950, opened 2016-11-07
Added attribution tint color check

cammace:
Checks that the attribution tint color isn't set to fully transparent.

cc: @tobrun


cammace:
Thanks for the review @tobrun, merging

PR #6965, opened 2016-11-08
add WITH_CLIPPER option to CMake builds

### LINKED ISSUE #6929: make v1 vector tile cleanup with ClipperLib configurable
We've seen significant performance implications in downstream implementations of the Node.js bindings when cleaning up complex v1 vector tiles at runtime. Given that the ClipperLib pathway is primarily intended for maintaining backward compatible rendering for offline databases containing v1 tiles, we should make this code path configurable for implementations where this isn't a concern.

mikemorris:
Fixes #6929

Enabled by default, but allows building without ClipperLib by passing `WITH_CLIPPER=0`.

I'm not sure about the duplicate code the in platform `config.cmake` files, is there a shared place I could add this logic instead?

Is completely dropping ClipperLib desirable, or should we only drop it for `VectorTileFeature::getGeometries` and keep `fixupPolygons` in place for `ShapeAnnotationImpl::updateTileData` and `GeoJSONTileFeature::GeometryTileFeature` per https://github.com/mapbox/geojson-vt-cpp/issues/44?

/cc @jfirebaugh @springmeyer


CMakeLists.txt
```diff
@@ -29,7 +30,11 @@ else()
 endif()
 
 if(IS_CI_BUILD)
-    add_compile_options(-DCI_BUILD=1)
+    target_compile_options(-DCI_BUILD=1)
+endif()
+
+if(WITH_CLIPPER)
+    target_compile_options(-D__CLIPPER__)
```
kkaefer: Can we instead use something like `MBGL_USE_CLIPPER` to be in line with other defines?

mikemorris:
CI tests are passing and this is successfully building Node.js binaries without ClipperLib now, think this is ready for review. Would particularly appreciate any suggestions on the CMake syntax I'm using (`target_sources` in most CMake files vs `LIST_APPEND` for QT).

kkaefer:
I don't think we should add the clipper files per-platform, since it's a global option, and not a platform-specific option.

1ec5:
@flippmoke, is this PR still needed in light of #8191, which would replace Clipper with Wagyu?

flippmoke:
@1ec5 I do not think we should ever not build wagyu/clipper as part of the process unless we are not supporting v1 tiles anymore.

jfirebaugh:
Agreed; let's pursue other approaches here.

Issue #6959, opened 2016-11-08
iOS - Removing and re-adding a MGL*Source causes a crash


JesseCrocker:
**Platform:**
iOS

**Mapbox SDK version:**
ios-v3.4.0-beta.2

### Steps to trigger behavior

 1. Add a MGLRasterSource or MGLVectorSource to the current style
 2. Remove the source
 3. Add the source again

### Expected behavior
Source is displayed, removed, displayed again

### Actual behavior
Source is displayed, removed, then the app crashes when source is re-added.

### Test Case
https://github.com/JesseCrocker/Mapbox-gl-native-testcases/blob/source-re-add-crash/MapboxTest/ViewController.m

### Stack Trace
```
#0	0x000000010fabf823 in mbgl::style::Style::addSource(std::__1::unique_ptr<mbgl::style::Source, std::__1::default_delete<mbgl::style::Source> >) ()
#1	0x000000010f84c906 in mbgl::Map::addSource(std::__1::unique_ptr<mbgl::style::Source, std::__1::default_delete<mbgl::style::Source> >) ()
#2	0x000000010f5f2e1d in ::-[MGLRasterSource addToMapView:](MGLMapView *) at /Users/boundsj/mapbox-gl-native/platform/darwin/src/MGLRasterSource.mm:65
#3	0x000000010f62f19e in ::-[MGLStyle addSource:](MGLSource *) at /Users/boundsj/mapbox-gl-native/platform/darwin/src/MGLStyle.mm:209
#4	0x000000010f544a70 in __47-[ViewController removeAndReAddSource:toStyle:]_block_invoke_2 at /Users/jesse/projects/Mapbox-gl-native-testcases/MapboxTest/ViewController.m:67
#5	0x00000001141860cd in _dispatch_client_callout ()
#6	0x00000001141621c3 in _dispatch_continuation_pop ()
#7	0x00000001141771e2 in _dispatch_source_latch_and_call ()
#8	0x000000011416fd7d in _dispatch_source_invoke ()
#9	0x0000000114166b2e in _dispatch_main_queue_callback_4CF ()
#10	0x0000000110cd94f9 in __CFRUNLOOP_IS_SERVICING_THE_MAIN_DISPATCH_QUEUE__ ()
#11	0x0000000110c9ef8d in __CFRunLoopRun ()
#12	0x0000000110c9e494 in CFRunLoopRunSpecific ()
#13	0x0000000118521a6f in GSEventRunModal ()
[...]

PR #6966, opened 2016-11-08
[ios, macos] Use the rawLayer pointer to access the mbgl layer

boundsj:
https://github.com/mapbox/mapbox-gl-native/pull/6904 renamed the `layer` property to `rawLayer`. The subclasses of `MGLStyleLayer` were all updated to use `rawLayer` but the
`MGLStyleLayer` itself was not. So, common functionality like setting visibility caused a crash.
This removes the `layer` property in `MGLStyleLayer` itself and uses the `rawLayer` property that actually points to the mbgl layer.

cc @1ec5

### DIFF
--- a/platform/darwin/src/MGLStyleLayer.mm
+++ b/platform/darwin/src/MGLStyleLayer.mm
@@ -5,8 +5,6 @@
 
 @interface MGLStyleLayer ()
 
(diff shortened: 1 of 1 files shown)


boundsj:
I noticed this while responding to https://github.com/mapbox/mapbox-gl-native/issues/6959

PR #6964, opened 2016-11-08
Fast Android rebuilds

kkaefer:
Android builds currently regenerate the style code on every build, which triggers a recompile of many dependent files (~60 files), so running `make android-lib-arm-v7` twice without modifying any files in between isn't fast.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -273,8 +273,10 @@ ideploy:
 idocument:
 	OUTPUT=$(OUTPUT) ./platform/ios/scripts/document.sh
 
-style-code-darwin:
+.PHONY: darwin-style-code
+darwin-style-code:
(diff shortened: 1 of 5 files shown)


ivovandongen:
@kkaefer With the changes you made to the code generation script to only write the files on changes, can't we keep the style generation on each build so we catch changes to the style spec early?

kkaefer:
@ivovandongen we could, but Core + Darwin doesn't auto-generate them either. Apart from that, rebuilding the files takes some time, and style spec changes happen infrequently enough that I don't think this is an issue. Since the generated files are actually checked in (unlike generates shaders or version.hpp), generating the new style files is an explicit developer action anyway.

PR #6968, opened 2016-11-08
[core] Add debug logging for glDrawElements crasher

jfirebaugh:
cc @1ec5 @boundsj

### DIFF
--- a/src/mbgl/renderer/fill_bucket.cpp
+++ b/src/mbgl/renderer/fill_bucket.cpp
@@ -108,9 +108,10 @@ void FillBucket::upload(gl::Context& context) {
 
 void FillBucket::render(Painter& painter,
                         PaintParameters& parameters,
-                        const Layer& layer,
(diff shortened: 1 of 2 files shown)


src/mbgl/renderer/fill_bucket.cpp
```diff
@@ -129,6 +130,14 @@ void FillBucket::drawElements(FillShader& shader,
     for (auto& group : triangleGroups) {
         group.getVAO(shader, paintMode).bind(
             shader, *vertexBuffer, *triangleIndexBuffer, vertex_index, context);
+        if (context.elementBuffer.getCurrentValue() == 0) {
```
kkaefer: A note that this call doesn't actually query the state; it just returns the internally tracked state, which might be out of sync with the OpenGL state (although it shouldn't be, ideally). To query the actual OpenGL state, use `gl::value::BindElementBuffer::Get()`

Issue #6616, opened 2016-10-06
Clearly document that any runtime styling API risks breaking changes


1ec5:
Documentation for the Android, iOS, macOS, and Qt SDKs should take every opportunity to warn the developer that using a particular layer, source, or sprite name risks a breaking change on any SDK release, even a patch release. The iOS and macOS SDKs’ documentation for `MGLStyleDefaultVersion` is [very clear](https://github.com/mapbox/mapbox-gl-native/blob/a6aa786f35824922aee0d2baad94e4ce9f71e79c/platform/darwin/src/MGLStyle.h#L23-L28) that referencing this style constant comes with risks, but the same risks apply if the developer doesn’t explicitly set the style at all, instead relying on the SDK to use the latest version of Streets. Meanwhile, the Android SDK continues to have unversioned style constants like `style_outdoors` that make it all too easy to use the latest Outdoors version with the runtime styling API, which is a bad idea.

APIs like `-[MGLStyle layerWithIdentifier:]`, `-[MGLStyle sourceWithIdentifier:]`, and `-[MGLStyle insertLayer:belowLayer:]` should clearly state that the developer must set the style URL to an explicitly versioned style, whether using `+[MGLStyle outdoorsStyleURLWithVersion:]`, the “Style URL” inspectable, or a manually constructed NSURL. Similarly, the corresponding Android SDK runtime styling methods should warn that the style must be set first and should not be set to one of the style string constants.

We can also take the opportunity to plug Mapbox Studio as a tool for creating custom styles with better stability guarantees as a jumping-off point for runtime styling.

/ref mapbox/mapbox-gl-styles#267 #4759
/cc @mapbox/mobile @lucaswoj @ajashton

lucaswoj:
> Documentation for the Android, iOS, macOS, and Qt SDKs should take every opportunity to warn the developer that using a particular layer, source, or sprite name risks a breaking change on any SDK release, even a patch release.

That's an unfortunate constraint after we spent so much time developing a semver system for our styles. Is there an oppurtunity to remove, deprecate, or demphasize `MGLStyleDefaultVersion`?

1ec5:
[...]

PR #6969, opened 2016-11-08
javadoc fixes

cammace:
We'll also need to cherry-pick this into master, closing other PR. I'm still getting 28 warnings when generating the javadoc but these seem to all be related to the excluded files. 

cc: @tobrun


tobrun:
🚢

cammace:
I've added some more missing javadoc along with documenting notes on the source/layer id being depended on the currently map style (more info in #6616). Can you review again @tobrun 😄

PR #6979, opened 2016-11-09
Don't initialize timestamps to TimePoint::min()

kkaefer:
An alternative fix to #6747/#6667: In `FrameHistory::needsAnimation`, we are computing `time - previousTime`, but since `previousTime` was initialized to `TimePoint::min()`, we got a integer overflow, which produced invalid values and made the map render continuously.

/cc @tobrun


tobrun:
Works great, both CPU monitor as tests confirm.

screen shot 2016-11-09 at 17 30 20

Thank you for picking this up!

Issue #1682, opened 2015-06-03
Vertical CJK labels


1ec5:
CJK labels running along north-south roads (or roads that go from top to bottom due to map rotation) should orient glyphs vertically so that one can read the labels without tilting one’s head. Since the map can be rotated arbitrarily, it’s easy to wind up in a situation where all the road labels are vertical. CJK makes it possible to maintain legibility in such configurations.

/cc @ansis @nickidlugash @peterqliu @mayagao @friedbunny @mikemorris

friedbunny:
An example from Yahoo JP, near Tokyo Station. Notice that even roman characters can be drawn vertically.

![screen shot 2015-06-03 at 1 23 11 pm](https://cloud.githubusercontent.com/assets/1198851/7970668/7cfba458-09f4-11e5-95de-e875074ccebd.png)

http://yahoo.jp/2RqUBc

nickidlugash:
@mikemorris is planning to add vertical advance information [here](https://github.com/mapbox/node-fontnik/blob/harfbuzz/proto/glyphs.proto#L24) to allow this.

mikemorris:
> Note that only a few font formats provide vertical metrics. You can test whether a given face object contains them by using the macro FT_HAS_VERTICAL, which returns true if appropriate.

https://www.freetype.org/freetype2/docs/tutorial/step2.html is interesting and has more details on vertical glyph metrics and what we would need to include in the glyph spec to support vertical layout

1ec5:
An implementation is nearly ready to port in mapbox/mapbox-gl-js#3438.

PR #6984, opened 2016-11-09
[core] Use ICU for bidirectional text layout and Arabic text shaping

ChrisLoer:


src/mbgl/util/utf.hpp
```diff
@@ -15,7 +15,28 @@ class utf8_to_utf32 {
         boost::u8_to_u32_iterator<std::string::const_iterator> end(utf8.end());
         return std::u32string(begin,end);
     }
+
```
jfirebaugh: We should rewrite this whole file in terms of [`std::codecvt`](http://en.cppreference.com/w/cpp/locale/codecvt), dropping the boost dependency.

src/mbgl/layout/symbol_layout.cpp
```diff
@@ -90,7 +92,7 @@ SymbolLayout::SymbolLayout(std::string bucketName_,
                 u8string = platform::lowercase(u8string);
             }
 
-            ft.text = util::utf8_to_utf32::convert(u8string);
+            ft.text = bidi.bidiTransform( util::utf8_to_utf32::convert( u8string ) );
```
jfirebaugh: Since `getGlyphRange` doesn't support anything beyond the Basic Multilingual Plane, are we better off using utf16 throughout this code, rather than going utf8 ⇢ utf32 ⇢ utf16 ⇢ utf32?

src/mbgl/text/bidi.cpp
```diff
@@ -0,0 +1,42 @@
...
+    if ( transform )
+        ubiditransform_close(transform);
+}
+
+std::u32string BiDi::bidiTransform( const std::u32string& input32 )
+{
+    UErrorCode errorCode = U_ZERO_ERROR;
+    
+    std::u16string input16 = util::utf32_to_utf16::convert( input32 );
+    std::unique_ptr<UChar*> outputText = std::make_unique<UChar*>( new UChar[input16.size() * 2] ); // Maximum output of ubidi_transform is twice the size of input according to ubidi_transform.h
```
jfirebaugh: `auto outputText = std::make_unique<UChar[]>(input16.size() * 2);`

brunoabinader: This is what [valgrind is complaining about](https://travis-ci.org/mapbox/mapbox-gl-native/jobs/174564772#L2323) on the Qt5 Travis bot.

I suppose you mean `std::make_unique<UChar[]>(input16.size() * 2);` instead?

brunoabinader:
Awesome stuff @ChrisLoer! Btw, we usually add tags e.g. `[core]` in commit messages in GL native.

jfirebaugh:
We'll want some [rendering tests](https://github.com/mapbox/mapbox-gl-test-suite) to go along with these changes. I suggest keeping them very simple, modeled on test cases such as [this one](https://github.com/mapbox/mapbox-gl-test-suite/blob/master/render-tests/text-rotation-alignment/auto-symbol-placement-line/style.json) which uses a single symbol layer plus a GeoJSON source to provide simple data.

src/mbgl/layout/symbol_layout.cpp
```diff
@@ -90,10 +92,10 @@ SymbolLayout::SymbolLayout(std::string bucketName_,
                 u8string = platform::lowercase(u8string);
             }
 
-            ft.text = util::utf8_to_utf32::convert(u8string);
+            ft.text = bidi.bidiTransform( util::utf8_to_utf16::convert( u8string ) );
```
1ec5: I’m curious about how visual mirroring at this stage interacts with line wrapping. Does this PR address the concern raised in https://github.com/mapbox/mapbox-gl-native/pull/6057#issuecomment-240545901?

1ec5: Bidi POI labels could get pretty long, for instance “مكتبة الإسكندرية‎‎ Maktabat al-Iskandarīyah” for the Library of Alexandria. In this case, the LTR text is a third longer than the RTL text. I don’t know how many cases there are where the two are roughly even in length.

ChrisLoer: I pushed my changes so you can take a look at how I did the line breaking, but the results for the Library of Alexandria are pretty disheartening.

src/mbgl/layout/symbol_layout.cpp
```diff
@@ -90,10 +92,12 @@ SymbolLayout::SymbolLayout(std::string bucketName_,
                 u8string = platform::lowercase(u8string);
             }
 
-            ft.text = util::utf8_to_utf32::convert(u8string);
+            std::u16string u16string = util::utf8_to_utf16::convert( u8string );
```
jfirebaugh: No spaces after `(` or before `)` in our house style. You can use clang-format to fix formatting up.

@kkaefer Is there a way to run clang-format on all modified files in a PR?

brunoabinader: @jfirebaugh uncomment [this line](https://github.com/mapbox/mapbox-gl-native/blob/master/scripts/clang-tools.sh#L52) and run `make check`. This will run both `clang-tidy` and `clang-format` on modified files.

ChrisLoer: I ran clang-tidy and clang-format on all the files, and it looks like they hadn't been tidied in a while, so it introduced a lot of diffs unrelated to my changes. Should I still go ahead with the changes?

jfirebaugh: Can you subset it further, to the code that's being modified for other reasons?

ChrisLoer: OK, just did a quick manual review and tried to only clang-format "new" code.

src/mbgl/text/bidi.cpp
```diff
@@ -0,0 +1,55 @@
...
+        UBIDI_DEFAULT_LTR,     // Assume input is LTR unless strong RTL characters are found
+        UBIDI_LOGICAL,         // Input is in logical order
+        UBIDI_LTR,             // Output is in "visual LTR" order
+        UBIDI_VISUAL,          //  ''
+        UBIDI_MIRRORING_ON,    // Use mirroring lookups for things like parentheses that need mirroring
+                               // in RTL text
+        U_SHAPE_LETTERS_SHAPE, // Add options here for handling numbers in bidirectional text
+        &errorCode);
+
+    if (U_FAILURE(
```
jfirebaugh: Put the comment above the if statement so this doesn't line-break so awkwardly.

src/mbgl/text/bidi.cpp
```diff
@@ -0,0 +1,55 @@
...
+        return input;
+
+    return std::u16string(outputText.get(), outputLength);
+}
+
+bool BiDi::baseDirectionRightToLeft(const std::u16string& input) {
+    // This just looks for the first character with a strong direction property, it does not perform
+    // the BiDi algorithm
+    return ubidi_getBaseDirection(input.c_str(), static_cast<int32_t>(input.size())) == UBIDI_RTL
+               ? true
```
jfirebaugh: No need to write `<boolean expression> ? true : false`, just write `<boolean expression>`.

ChrisLoer:
@1ec5 you might want to double-check the changes I made to i18n.cpp while I was rebasing -- it was just moving from uint32_t characters to uint16_t characters, which I think should be fine since we're already limited to the BMP. I talked with @jfirebaugh about the line-breaking problem and we decided to start with the simplified rule of reversing the line breaks when the base direction of the label is RTL. Your heuristic for line breaking in #6057 gets better results, but it's not as easy to apply when we're doing the transformation ahead of time with the input string. Our plan is to start with this PR as a strict improvement on "totally broken", and then follow up with a PR that uses ICU's support for bidirectional line breaking.

One other issue that we haven't discussed at all is handling of numbers in Arabic text. I just used the default behavior, and we end up with the same behavior OpenStreetMap uses (numbers are written LTR using Arabic-Indic digits: for example, see the "24" at https://www.openstreetmap.org/way/26177467). Google Maps uses European digits in the same case (see https://goo.gl/maps/E2yvjkVTp8R2). ICU shaping includes options that make it easy for us to choose either format regardless of the underlying data.

1ec5:
> Your heuristic for line breaking in #6057 gets better results, but it's not as easy to apply when we're doing the transformation ahead of time with the input string.

#6057 was strictly about mirroring rather than complex text shaping. The visual length of a line of text remains constant after mirroring, so it was possible to mirror after applying line breaking.

Similarly, I think it would be possible to individually text-shape each individual line after line breaking, since Arabic words are still delimited by spaces (unlike, say, Thai). The caveat is that some lines would shorten during the transformation, so it would appear as though we were breaking some lines prematurely. So the result for purely RTL text would probably be severely width-constrained compared to the approach you’ve implemented; on the other hand, bidirectional text would finally read correctly.

> Our plan is to start with this PR as a strict improvement on "totally broken", and then follow up with a PR that uses ICU's support for bidirectional line breaking.

This PR is certainly an improvement over master. I’m looking forward to the glorious ICU future. 👍

kkaefer:
I ran a manual build for Android/arm-v7 and got a size change of **+7%**, from 4,903,788 bytes to 5,239,412 bytes

src/mbgl/text/bidi.cpp
```diff
@@ -1,42 +1,53 @@
...
+    return std::u16string(outputText.get(), outputLength);
 }
-    
+
+bool BiDi::baseDirectionRightToLeft(const std::u16string& input) {
+    // This just looks for the first character with a strong direction property, it does not perform
+    // the BiDi algorithm
+    return ubidi_getBaseDirection(input.c_str(), static_cast<int32_t>(input.size())) == UBIDI_RTL;
+}
+
```
kkaefer: Can you please merge the code formatting changes into the previous commit? That makes the code changes a lot easier to follow.

src/mbgl/text/glyph_set.hpp
```diff
@@ -9,12 +9,12 @@ class GlyphSet {
...
     void insert(uint32_t id, SDFGlyph&&);
     const std::map<uint32_t, SDFGlyph> &getSDFs() const;
-    const Shaping getShaping(const std::u16string &string, float maxWidth, float lineHeight,
+    const Shaping getShaping(const std::u16string &string, const bool rightToLeft, float maxWidth, float lineHeight,
                              float horizontalAlign, float verticalAlign, float justify,
                              float spacing, const Point<float> &translate) const;
     void lineWrap(Shaping &shaping, float lineHeight, float maxWidth, float horizontalAlign,
                   float verticalAlign, float justify, const Point<float> &translate,
-                  bool useBalancedIdeographicBreaking) const;
+                  bool useBalancedIdeographicBreaking, const bool rightToLeft) const;
```
kkaefer: We've been moving towards using enums like `enum class WritingDirection : bool { LeftToRight, RightToLeft };` to avoid ambiguity in functions like this.

src/mbgl/util/utf.hpp
```diff
@@ -2,20 +2,18 @@
 
 #include <memory>
 
-#include <locale>
 #include <codecvt>
+#include <locale>
```
kkaefer: null change

src/mbgl/text/bidi.cpp
```diff
@@ -0,0 +1,42 @@
...
+#include <mbgl/text/bidi.hpp>
+#include <unicode/ushape.h>
+
+namespace mbgl {
+
+BiDi::BiDi()
+{
+    UErrorCode errorCode = U_ZERO_ERROR;
+    transform = ubiditransform_open(&errorCode); // Only error is failure to allocate memory, in that case ubidi_transform would fall back to creating transform object on the fly
```
kkaefer: We're recording the error code, but don't do anything with it. Can we either fail hard here, or add checks in `BiDi::bidiTransform` that don't use the `transform` object if it's not initialized?

src/mbgl/text/bidi.hpp
```diff
@@ -0,0 +1,21 @@
...
+    
+    class BiDi : private util::noncopyable {
+public:
+    BiDi();
+    ~BiDi();
+    
+    std::u16string bidiTransform( const std::u16string& );
+    
+private:
+    UBiDiTransform* transform;
```
kkaefer: Initialize to `nullptr`?

src/mbgl/layout/symbol_feature.hpp
```diff
@@ -11,6 +11,7 @@ class SymbolFeature {
 public:
     GeometryCollection geometry;
     optional<std::u16string> text;
+    optional<bool> rightToLeft;
```
kkaefer: Use an enum `WritingDirection` here for clarity

src/mbgl/text/glyph_set.hpp
1ec5: This would jive nicely with the WritingMode bitfield that I'm implementing for #1682.

ChrisLoer: OK, moved this into a `WritingDirection` enum defined in `bidi.hpp`. @1ec5, I didn't see a PR that already had a `WritingMode` bitfield, so hopefully this still jives nicely.

src/mbgl/util/utf.hpp
ChrisLoer: Removed. So `clang-format` will put includes in alphabetical order... Is our style "add includes in alphabetical order, but don't reorder existing ones"?

jfirebaugh:
#7098 should fix the test-suite errors.

PR #6989, opened 2016-11-09
[core] Default camera position to that specified by the style

### LINKED ISSUE #6129: Jump to style’s default viewport on initialization

jfirebaugh:
Fixes #6129. Alternate take on #6608.

Needed for getting the regression test for #6932 to work correctly.

### DIFF
--- a/src/mbgl/map/map.cpp
+++ b/src/mbgl/map/map.cpp
@@ -47,7 +47,8 @@ struct StillImageRequest {
 
 class Map::Impl : public style::Observer {
 public:
-    Impl(Backend&,
+    Impl(Map&,
+         Backend&,
          float pixelRatio,
          FileSource&,
          Scheduler&,
@@ -67,6 +68,7 @@ public:
 
     void loadStyleJSON(const std::string&);
 
+    Map& map;
     Backend& backend;
     FileSource& fileSource;
     Scheduler& scheduler;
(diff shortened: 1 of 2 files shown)


1ec5:
> Fixes #6291

You mean #6129?

jfirebaugh:
Thanks, fixed.

Issue #309, opened 2014-06-04
Consider miniutf over boost::locale


springmeyer:
[miniutf](https://github.com/dropbox/miniutf) looks like it got its start for the same reason we fell back to using `boost::locale` on linux (#270): because libstdc++ is still missing full `codecvt_utf8 / codecvt_utf16` support.

TODO: 

  - [ ]  - benchmark to make sure miniutf is as fast as boost::locale/icu/std:: - https://github.com/mapnik/mapnik/commit/0859a74f81fc1f3fa5c3ebbbd3388df721d55e4f

kkaefer:
This is merged and we're using nunicode.

PR #6991, opened 2016-11-10
[core] Bump standard library to libstdc++5-dev and clang version to 3.8.

ChrisLoer:


jfirebaugh:
This proved to be a prerequisite for portions of #6984, namely using `<codecvt>` for UTF transcoding.

@kkaefer @brunoabinader @springmeyer See any downsides to upgrading? This basically brings mapbox-gl-js in line with the recommendations of https://github.com/mapbox/cpp.

springmeyer:
:+1 This feels like the right way forward. https://github.com/mapbox/mapbox-gl-native/issues/309 tracks when we hit this before and decided to work around without upgrading libstdc++. I think it is a good time to require at least libstdc++-5-dev.

> See any downsides to upgrading?

I think the upsides outweight the downsides. And the downside is not a blocker, but I'll detail it for the record. It is that requiring libstdc++-5 instead of 4.9 will trigger a GLIBCXX symbol dependency will jump to GLIBCXX_3.4.21. This means that binaries built of mbgl-native will work out of the box on systems >= ubuntu xenial, but will not work on ubuntu trusty or ubuntu precise without upgrading libstdc++. Because we were targeting libstdc++-4.9 before, mbgl binaries (including node-mbgl) would run on ubuntu trusty by default. However, for Mapbox this has no impact, since we've already upgraded our systems to support GLIBCXX_3.4.21 as per the upgrade methods described in https://github.com/mapbox/cpp.

kkaefer:
As an alternative to using `codecvt`, we could also make this a system-specific API and backfill with Nunicode on platforms where no platform-native alternative is available.

jfirebaugh:
`codecvt` is a standard C++11 feature -- it makes more sense to me to use it rather than platform-specific equivalents.

PR #7005, opened 2016-11-10
Track binary size over time

kkaefer:
Add logging to CloudWatch that records binary size for Linux, Android, iOS and macOS. In the case of fat binaries, the binary size is split up to individual architectures (iOS/Android) in addition to the default archive.

The idea is that we'll have nightly builds (running in EU early morning) that create a stripped release build to approximate the size of the library in a deployed setting, and record those. Using the data stored in CloudWatch, we'll be able to generate graphs of how the binary size changes over time.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -49,6 +49,8 @@ script:
 after_script:
   - ccache --show-stats
   - ./platform/linux/scripts/after_script.sh ${TRAVIS_JOB_NUMBER}
+after_success:
+  - ./platform/linux/scripts/after_success.sh
 
 matrix:
   include:
--- a/cloudformation/travis.template
+++ b/cloudformation/travis.template
@@ -8,42 +8,6 @@
             "Properties": {
                 "Policies": [
                     {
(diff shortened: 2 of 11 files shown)


kkaefer:
Can we get this approved so that we can start collecting sizes?

jfirebaugh:
@kkaefer Can you record somewhere how to view the CloudWatch history?

kkaefer:
Follow up in #7092

kkaefer:
You can view the binary size log at https://mapbox.github.io/mapbox-gl-native/metrics/binary-size/

tmpsantos:
@kkaefer this is really nice, could the same idea be used for tracking metrics like memory consumption?

kkaefer:
@tmpsantos yeah; have a look at `scripts/log_binary_size.sh`. This records an arbitrary number with arbitrary dimensions to CloudWatch, which are then pulled down by `scripts/publish_binary_size.sh` and are published as a public JSON feed, which are then loaded by the graph page.

PR #7007, opened 2016-11-10
[linux] Added EGL headless backend support

brunoabinader:
Follow-up to the work started by @tiagovignatti in #5843 - thanks Tiago!

This PR implements EGL headless backend, using `EGL_DEFAULT_DISPLAY` as primary display - thus no need for handling render nodes directly via [gbm](https://en.wikipedia.org/wiki/Mesa_(computer_graphics)#Generic_Buffer_Management).

Though we use software rendering via LLVMpipe, we still need a valid display for EGL to be properly initialized. On Travis CI, _like the Qt builds_, we still need to start Xvfb to obtain a valid display.

I've put some effort attempting to find a working setup for Mesa to have its EGL platform working with a native display that doesn't require a window system, but so far no success. The most promising option, though, comes from this recent [Mesa extension](https://www.khronos.org/registry/egl/extensions/MESA/EGL_MESA_platform_surfaceless.txt) - __MESA_platform_surfaceless__ - that promises to provide a native display independent of any native window system. Given this solution already works and is testable via CI, we can pursuit this as a follow-up work.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,27 +2,48 @@ git:
   submodules: false
 
 # Save common build configurations as shortcuts, so we can reference them later.
+addons:
+  apt:
+    sources:
+      - &common_sources [ 'ubuntu-toolchain-r-test', 'george-edison55-precise-backports' ]
+    packages:
+      - &common_packages [ 'libllvm3.8v4', 'cmake', 'cmake-data' ]
+      - &clang38_packages [ 'clang-3.8', 'libstdc++-5-dev', 'libstdc++6' ]
+      - &gcc5_packages [ 'gcc-5', 'g++-5' ]
+      - &egl_packages [ 'libgles2-mesa-dev', 'libgbm-dev' ]
+      - &glfw_packages [ 'libxrandr-dev', 'libxcursor-dev', 'libxinerama-dev' ]
+
 addons_shortcuts:
   addons_clang38: &clang38
     apt:
-      sources:  [ 'ubuntu-toolchain-r-test', 'george-edison55-precise-backports' ]
-      packages: [ 'clang-3.8', 'libstdc++-5-dev', 'libstdc++6', 'libllvm3.8v4', 'cmake', 'cmake-data',
-                  'libxrandr-dev', 'libxcursor-dev', 'libxinerama-dev' ]
+      sources: *common_sources
+      packages:
+        - *common_packages
+        - *clang38_packages
+        - *egl_packages
+        - *glfw_packages
   addons_gcc5: &gcc5
     apt:
-      sources:  [ 'ubuntu-toolchain-r-test', 'george-edison55-precise-backports' ]
-      packages: [ 'g++-5', 'gcc-5', 'libllvm3.8v4', 'cmake', 'cmake-data',
-                  'libxrandr-dev', 'libxcursor-dev', 'libxinerama-dev' ]
+      sources: *common_sources
+      packages:
+        - *common_packages
+        - *gcc5_packages
+        - *egl_packages
+        - *glfw_packages
   addons_qt4: &qt4
     apt:
-      sources:  [ 'ubuntu-toolchain-r-test', 'george-edison55-precise-backports' ]
-      packages: [ 'g++-5', 'gcc-5', 'libllvm3.8v4', 'libjemalloc-dev', 'cmake', 'cmake-data',
-                  'mesa-utils', 'qt4-default' ]
+      sources: *common_sources
+      packages:
+        - *common_packages
+        - *gcc5_packages
+        - [ 'libjemalloc-dev', 'mesa-utils', 'qt4-default' ]
   addons_qt5: &qt5
     apt:
-      sources:  [ 'ubuntu-toolchain-r-test', 'george-edison55-precise-backports' ]
-      packages: [ 'g++-5', 'gcc-5', 'libllvm3.8v4', 'cmake', 'cmake-data',
-                  'mesa-utils', 'libc6-dbg', 'qt5-default', 'libqt5opengl5-dev', 'qtdeclarative5-dev', 'qtpositioning5-dev', 'qtlocation5-dev' ]
+      sources: *common_sources
+      packages:
+        - *common_packages
+        - *gcc5_packages
+        - [ 'mesa-utils', 'libc6-dbg', 'qt5-default', 'libqt5opengl5-dev', 'qtdeclarative5-dev', 'qtpositioning5-dev', 'qtlocation5-dev' ]
 
 env:
   global:
@@ -38,9 +59,7 @@ env:
 install:
   - source ./scripts/travis_helper.sh
   - source ./scripts/travis_setup.sh
-before_script:
   - ccache --zero-stats
-  - cmake --version
 script:
   - make linux
   - make benchmark
@@ -54,6 +73,17 @@ after_success:
 
 matrix:
   include:
+    # LLVM 3.8.0 - clang-{format,tidy}
+    - os: linux
+      sudo: false
+      dist: trusty
+      language: cpp
+      env: _CXX=c++ _CC=cc
+      compiler: "check"
+      script:
+        - git fetch origin master:refs/remotes/origin/master
+        - make check
+
     # OSMesa - Node - Clang 3.8 - Debug
     - os: linux
       sudo: required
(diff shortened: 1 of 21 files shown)


.travis.yml
```diff
@@ -2,27 +2,48 @@ git:
...
+addons:
+  apt:
+    sources:
+      - &common_sources [ 'ubuntu-toolchain-r-test', 'george-edison55-precise-backports' ]
+    packages:
+      - &common_packages [ 'libllvm3.8v4', 'cmake', 'cmake-data' ]
+      - &clang35_packages [ 'clang-3.5', 'libstdc++-4.9-dev', 'libstdc++6' ]
+      - &gcc5_packages [ 'gcc-5', 'g++-5' ]
+      - &egl_packages [ 'libgles2-mesa-dev', 'libgbm-dev' ]
+      - &glfw_packages [ 'libxrandr-dev', 'libxcursor-dev', 'libxinerama-dev' ]
```
kkaefer: 😍

platform/linux/config.cmake
```diff
@@ -15,19 +15,31 @@ mason_use(benchmark VERSION 1.0.0)
...
             PRIVATE platform/default/headless_backend_osmesa.cpp
             PRIVATE platform/default/headless_display.cpp
         )
         target_add_mason_package(mbgl-core PUBLIC mesa)
+    elseif(WITH_EGL)
+        target_sources(mbgl-core
+            PRIVATE platform/linux/src/headless_backend_egl.cpp
+            PRIVATE platform/linux/src/headless_display_egl.cpp
+        )
+        # XXX: Provide surface-less EGL mesa for CI builds.
```
kkaefer: remaining todo?

platform/linux/src/headless_backend_egl.cpp
```diff
@@ -0,0 +1,73 @@
...
+};
+
+void HeadlessBackend::createContext() {
+    assert(!hasContext());
+    assert(hasDisplay());
+
+    EGLDisplay display_ = display->attribute<EGLDisplay>();
+    EGLConfig& config = display->attribute<EGLConfig&>();
+
+    eglBindAPI(EGL_OPENGL_API);
```
kkaefer: missing error handling

platform/linux/src/headless_display_egl.cpp
```diff
@@ -0,0 +1,64 @@
...
+
+    const EGLint attribs[] = {
+        EGL_SURFACE_TYPE, EGL_DONT_CARE,
+        EGL_RED_SIZE, 1,
+        EGL_GREEN_SIZE, 1,
+        EGL_BLUE_SIZE, 1,
+        EGL_ALPHA_SIZE, 0,
+        EGL_DEPTH_SIZE, 1,
+        EGL_RENDERABLE_TYPE, EGL_OPENGL_BIT,
+        EGL_NONE
```
kkaefer: Maybe add a note that the configuration here doesn't matter since we're rendering to a framebuffer anyway.

platform/linux/config.cmake
brunoabinader: Yeah - https://github.com/mapbox/mapbox-gl-native/issues/7020
I'll add that link to the comment 👍

platform/linux/src/headless_display_egl.cpp
brunoabinader: Good idea 👍

platform/linux/src/headless_backend_egl.cpp
```diff
@@ -0,0 +1,75 @@
...
+};
+
+void HeadlessBackend::createContext() {
+    assert(!hasContext());
+    assert(hasDisplay());
+
+    EGLDisplay display_ = display->attribute<EGLDisplay>();
+    EGLConfig& config = display->attribute<EGLConfig&>();
+
+    if (!eglBindAPI(EGL_OPENGL_API)) {
```
kkaefer: You are setting `MBGL_USE_GLES2` above, but requesting a desktop OpenGL API here. Did you try requesting `EGL_OPENGL_ES_API`?

brunoabinader: Yes - I initially used `EGL_OPENGL_ES_API`, but then I obtained a crash right after when querying for the shading version string.

I believe this happens because when I attempt to use that enum, querying for the version string of my local EGL implementation returns `Version: OpenGL ES-CM 1.1 Mesa 10.1.3`. From the [eglBindAPI](https://www.khronos.org/registry/egl/sdk/docs/man/html/eglBindAPI.xhtml) documentation, it says `eglBindAPI and the corresponding EGL_OPENGL_ES_API and EGL_OPENVG_API api parameters are supported only if the EGL version is 1.2 or greater.`

I think we can, however, check for the current OpenGL version prior calling `eglBindAPI` to make sure to set it for `EGL_OPENGL_ES_API` only when version is >= 1.2.

platform/linux/src/headless_display_egl.cpp
```diff
@@ -25,6 +26,15 @@ HeadlessDisplay::Impl::Impl() {
         throw std::runtime_error("eglInitialize() failed.\n");
     }
 
+    if (major >= 1 && minor >= 2) {
```
kkaefer: that fails for EGL `2.0`

platform/linux/src/headless_display_egl.cpp
```diff
@@ -25,6 +26,15 @@ HeadlessDisplay::Impl::Impl() {
...
+        EGLenum api = minor >= 4 ? EGL_OPENGL_API : EGL_OPENGL_ES_API;
```
kkaefer: also problematic works for `1.4`, but not for `2.0`? If it works, we /always/ want the OpenGL ES API, since that's what we're mainly targeting. We just happen to be largely compatible with Desktop OpenGL due to the limited featureset we're using.

brunoabinader:
@kkaefer I found the reason why we were unable to compile the shader.

Even though we bind to `EGL_OPENGL_ES_API` by default, the client version is set to `1` by default when creating the context - which doesn't have support for OpenGL ES Shading Language. Support for shaders only comes with OpenGL ES 2.0.

To fix that, we need to pass `EGL_CONTEXT_CLIENT_VERSION, 2` as an attribute inside `attrib_list`  in which we pass to `eglCreateContext`.

brunoabinader:
Heh, this worked better than I've expected: Because we really use OpenGL ES 2.0 client API, I had to append the code snippet below to each of the test shaders from `OffscreenTexture.RenderToTexture` to account for it:
```
#ifdef GL_ES
precision mediump float;
#endif
```

If my assumptions about Android using EGL as backend are correct, then this is probably one step further [running core tests on Android](https://github.com/mapbox/mapbox-gl-native/issues/5456).

brunoabinader:
@kkaefer 🚢 ?

brunoabinader:
> Can we add on a render test to those as well?

This is going to be provided in https://github.com/mapbox/mapbox-gl-native/pull/7053 👍

brunoabinader:
@kkaefer I've rebased this after #7053 got merged. There are now 4 new Travis CI builds added:

- EGL:
 - EGL/Node/Clang38/Debug: __runs Node render/query tests__
 - EGL/GCC5/Release: __runs core unit tests__
- GLX:
 - GLX/Node/Clang38/Debug: __runs Node render/query tests__
 - GLX/GCC5/Release: __runs core unit tests__

@mikemorris the added Node builds won't affect the `release` node build. These are just for ensuring our Node render/query test passes using these OpenGL APIs.

.travis.yml
```diff
@@ -91,78 +159,94 @@ matrix:
...
+    # Qt 4 - GCC 5 - Release
     - os: linux
       sudo: required
       dist: trusty
       language: cpp
       compiler: "qt4-gcc5-release"
-      env: BUILDTYPE=Release _CXX=g++-5 _CC=gcc-5 RUN_XVFB=1 WITH_QT_4=1
+      env: BUILDTYPE=Release _CXX=g++-5 _CC=gcc-5 WITH_QT_4=1
       addons: *qt4
+      before_script:
```
mikemorris: Should the `GLX - Node - Clang 3.8 - Release` build currently using `RUN_XVFB` be switched to these setup scripts instead?

scripts/travis_setup.sh
```diff
@@ -37,15 +39,23 @@ git submodule update --init .mason
...
     sh -e /etc/init.d/xvfb start
     sleep 2 # sometimes, xvfb takes some time to start up
 
     # Make sure we're connecting to xvfb
     export DISPLAY=:99.0
+}
 
+export -f mapbox_start_xvfb
+
+function mapbox_install_mesa_glx {
```
mikemorris: This is needed as a separate script for only the QT builds because Mesa is installed in `platform/linux/config.cmake` for Linux builds, right? Is there any way we could consolidate these into a single script that accepts args to install the necessary version of Mesa?

platform/linux/config.cmake
```diff
@@ -1,5 +1,5 @@
 mason_use(glfw VERSION 3.2.1)
-if (WITH_OSMESA OR IS_CI_BUILD)
+if(IS_CI_BUILD AND NOT WITH_EGL)
```
mikemorris: I'm a little confused by this - are we installing Mesa somewhere else for EGL CI builds? It appears that we're still setting `MASON_MESA_SUFFIX` for `WITH_EGL` builds in `CMakeLists.txt`.

.travis.yml
brunoabinader: Yes, forgot about that :)

platform/linux/config.cmake
brunoabinader: > I'm a little confused by this - are we installing Mesa somewhere else for EGL CI builds? 

Yes - we are using the system's mesa install via `libgles2-mesa-dev`.

> It appears that we're still setting MASON_MESA_SUFFIX for WITH_EGL builds in CMakeLists.txt.

That should cause no harm - it's going to be used when we get the EGL surface-less mesa build :)

brunoabinader:
> Can we think about ways to make the matrix smaller?

@jfirebaugh we've reduced the build matrix to 10 items.

PR #7014, opened 2016-11-10
[core] Return source and layer ownership

boundsj:
When a source or layer is removed transfer ownership back to the caller so it can (optionally) take it.

ref https://github.com/mapbox/mapbox-gl-native/issues/6959

cc @jfirebaugh

### DIFF
--- a/src/mbgl/map/map.cpp
+++ b/src/mbgl/map/map.cpp
@@ -855,7 +855,7 @@ void Map::addSource(std::unique_ptr<style::Source> source) {
 std::unique_ptr<Source> Map::removeSource(const std::string& sourceID) {
     if (impl->style) {
         impl->styleMutated = true;
-        return impl->style->removeSource(sourceID);        
+        return impl->style->removeSource(sourceID);
     }
     return nullptr;
 }
--- a/src/mbgl/map/map.cpp
+++ b/src/mbgl/map/map.cpp
@@ -852,11 +852,12 @@ void Map::addSource(std::unique_ptr<style::Source> source) {
     }
 }
 
-void Map::removeSource(const std::string& sourceID) {
+std::unique_ptr<Source> Map::removeSource(const std::string& sourceID) {
     if (impl->style) {
         impl->styleMutated = true;
-        impl->style->removeSource(sourceID);
+        return impl->style->removeSource(sourceID);        
     }
+    return nullptr;
 }
 
 style::Layer* Map::getLayer(const std::string& layerID) {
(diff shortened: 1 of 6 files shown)


boundsj:
Once this is ready, I'd like to merge it to master and then cherry-pick to the iOS (and maybe Android) release branches where the returned objects can actually be used in the platform source and layer wrappers.

src/mbgl/map/map.cpp
```diff
@@ -852,11 +852,12 @@ void Map::addSource(std::unique_ptr<style::Source> source) {
     }
 }
 
-void Map::removeSource(const std::string& sourceID) {
+std::unique_ptr<Source> Map::removeSource(const std::string& sourceID) {
     if (impl->style) {
         impl->styleMutated = true;
-        impl->style->removeSource(sourceID);
+        return impl->style->removeSource(sourceID);
```
jfirebaugh: Trailing whitespace.

src/mbgl/map/map.cpp
```diff
@@ -881,18 +882,20 @@ void Map::addLayer(std::unique_ptr<Layer> layer, const optional<std::string>& be
...
     }
 
     impl->styleMutated = true;
     impl->backend.activate();
 
-    impl->style->removeLayer(id);
+    auto removedLayer = impl->style->removeLayer(id);
     impl->onUpdate(Update::Classes);
 
     impl->backend.deactivate();
```
jfirebaugh: We will need to add special logic for `CustomLayer`.  Currently, `~CustomLayer` calls the `CustomLayerDeinitializeFunction`, which may make OpenGL calls. This is the reason that we call `impl->backend.activate()/deactivate()` -- the context needs to be active when initialization/deinitialization happens.

I think we need to preserve the behavior that removing a CustomLayer triggers deinitialization, even when returning ownership. Let's do this in `Style::removeLayer`, making it parallel `Style::addLayer`, which does initialization.

boundsj: 00c5210

src/mbgl/style/layers/custom_layer_impl.cpp
```diff
@@ -43,6 +43,11 @@ void CustomLayer::Impl::initialize() {
     initializeFn(context);
 }
 
+void CustomLayer::Impl::deinitialize() {
+    assert(deinitializeFn);
+    deinitializeFn(context);
```
jfirebaugh: Also need to remove deinitialization from `CustomLayer::Impl::~Impl`. And from the looks of that method, it's possible that `deinitializeFn` may be optional? I don't remember but we should probably keep the guard.

boundsj:
@jfirebaugh can you please double check that https://github.com/mapbox/mapbox-gl-native/pull/7014/commits/554abe9eab82251cf5e0316040cb1fc7d1d3f26b looks ok to you?

jfirebaugh:
👍

PR #7015, opened 2016-11-11
[android] - getMarkerViews in rect should return in bound views

### LINKED ISSUE #7008: getMarkerViewsInRect returns list of all the markers
When using `getMarkerViewsInRect` it seems to not be properly returning only marker views inside the `RectF` and instead gives a list of all the markers on map.

ezgif com-video-to-gif 32

https://github.com/mapbox/mapbox-gl-native/blob/master/platform/android/MapboxGLAndroidSDK/src/main/java/com/mapbox/mapboxsdk/maps/AnnotationManager.java#L353-L372

cc: @mapbox/android

tobrun:
Closes #7008 
Review @cammace


tobrun:
@cammace thank you for adding the example. I fixed up the code based on `getMarkersInRect`. I now replaced the MarkerView with a Marker and called `getMarkersInRect` and this shows the same result. There seems to be a bigger issue here.

tobrun:
The issue was that we were passing an pixel independent rectangle, taking in account screen density resolves the issue.

PR #6948, opened 2016-11-07
[ios] specify plist-file for settings bundle

frederoni:
Fixes #6947 

Did this start with Xcode 8.1 or later?
Root.plist is usually picked up automatically from the settings bundle. Not sure why it doesn't all of a sudden.

@kkaefer 👀

PR #7032, opened 2016-11-11
Cherry pick #6948

boundsj:
cc @1ec5


boundsj:
This is happening in https://github.com/mapbox/mapbox-gl-native/pull/7034

PR #7021, opened 2016-11-11
Cherry pick fixes from master to android-4.2.0

ivovandongen:
Fixes #6945
Fixes #6975
Fixes #6977
Fixes #6998

PR #6899, opened 2016-11-03
Add support of OSMesa (Gallium) headless backend

brunoabinader:
As per https://github.com/mapbox/mapbox-gl-native/issues/4498#issuecomment-258109635 - using OSMesa as headless backend fixes the rendering artifacts that were causing us to ignore 42 render tests. This matches 1:1 with macOS results, so we can safely unignore them.

Follow-up to work started by @mikemorris in #4638.
Fixes #1555 and #4498.

Caveats:
- We'll need some ❤️ in our `.travis.yml` after this, as e.g. the node build won't need X library dependencies anymore.
- EGL support to be added in a following PR.

Default backends on Linux:
- GLX: `mbgl-glfw` and `mbgl-render`
- OSMesa: `mbgl-node`, `mbgl-test` and `mbgl-benchmark`

PR #7034, opened 2016-11-11
Cherry pick #6928 #6932 #6948 #7014

boundsj:
cc @1ec5


boundsj:
Along with several Qt/Linux related commits that get the build passing in CI again, this brings in:

- [core] Separate pathway for notifying workers of new images/glyphs aa087967675cd337961ebea78bd04836600d8aed
- [core] symbol layer - recalculate style on icon/text size changes 39b0bc0a402248e29791f93aeab5a6234bf80862

This does **not** include:

- [core] separate Backend from View for headless rendering 5cc390d694fc7510d445310d8eb9e32429a5e67b
- [core] move gl::Context to Backend and refactor View a4d259c33f9bb890bba97fd89552720e3e0ec09b
- [core] don't load tiles from sources that aren't used 91272992f3a7ca7baa124c3378f1c1e2b3e5bb76
- [core] change std::array<uint16_t, 2> to mbgl::Size 62b56b799a7d4fcd1a8f151eed878054b862da5b
- [core] correctly set the framebuffer size on retina screens f7801e5d82d272d53db6925a325391e5da5c842e

I think this is the correct of action but I wanted to check since these commits that I did not include were brought into the Android release in https://github.com/mapbox/mapbox-gl-native/pull/7021. 

cc @1ec5 @jfirebaugh @ivovandongen

jfirebaugh:
@boundsj Those look like appropriate changes to cherry pick. I would add #7025 to the list as well.

#7021 pulled a broader set of commits than I would have recommended.

1ec5:
Note that this PR also pulled in 3f5972cba000f602df7504d9214e6069fb3fb874, which was part of #6899.

PR #7039, opened 2016-11-14
[android] return layer/source ownership on remove

### LINKED ISSUE #7023: Return source and layer ownership on remove
When a layer or source is removed from the style, it should return ownership to the wrapping object so it can be re-added later / treated as new. See #6959 for the original issue

Depends on 
- Core implementation: #7014
- Ownership refactoring: #6995 

cc @jfirebaugh @boundsj

ivovandongen:
Fixes #7023


ivovandongen:
@tobrun care to review?

PR #7048, opened 2016-11-14
[ios, macos] Take back source and layer ownership on removal

boundsj:
If a source or layer is removed from the style, recapture the unique pointer ownership. This makes it safe to add back sources and layers after they have been removed.

Fixes iOS and macOS portion of https://github.com/mapbox/mapbox-gl-native/issues/6959

cc @1ec5 @frederoni @incanus

### DIFF
--- a/platform/darwin/src/MGLBackgroundStyleLayer.mm
+++ b/platform/darwin/src/MGLBackgroundStyleLayer.mm
@@ -32,6 +32,31 @@
 }
 
 
+
+#pragma mark -  Adding to and removing from a map view
+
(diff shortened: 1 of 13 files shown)


boundsj:
Not that it is advisable, but it is now safe to write layer/source toggling logic that looks something like:

```Objective-C

// Assume you have pointers to "layer" and "source" instances that were previously added

if (toggleOff) {
    [self.mapView.style removeLayer:layer];
    [self.mapView.style removeSource:source];
} else {
}
```

PR #7054, opened 2016-11-14
[core] Use indexed rendering for collision boxes

jfirebaugh:
Follows https://github.com/mapbox/mapbox-gl-js/commit/b9d95095c63a8e0193b6ee391ab308e75e1ecea3.

### DIFF
--- a/src/mbgl/layout/symbol_layout.cpp
+++ b/src/mbgl/layout/symbol_layout.cpp
@@ -507,14 +507,28 @@ void SymbolLayout::addToDebugBuffers(CollisionTile& collisionTile, SymbolBucket&
                 const float maxZoom = util::clamp(zoom + util::log2(box.maxScale), util::MIN_ZOOM_F, util::MAX_ZOOM_F);
                 const float placementZoom = util::clamp(zoom + util::log2(box.placementScale), util::MIN_ZOOM_F, util::MAX_ZOOM_F);
 
-                collisionBox.vertices.emplace_back(CollisionBoxProgram::vertex(anchor, tl, maxZoom, placementZoom),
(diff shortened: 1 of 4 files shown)


src/mbgl/layout/symbol_layout.cpp
```diff
@@ -507,14 +507,25 @@ void SymbolLayout::addToDebugBuffers(CollisionTile& collisionTile, SymbolBucket&
...
 
-                collisionBox.vertices.emplace_back(CollisionBoxProgram::vertex(anchor, tl, maxZoom, placementZoom),
-                                                   CollisionBoxProgram::vertex(anchor, tr, maxZoom, placementZoom));
-                collisionBox.vertices.emplace_back(CollisionBoxProgram::vertex(anchor, tr, maxZoom, placementZoom),
-                                                   CollisionBoxProgram::vertex(anchor, br, maxZoom, placementZoom));
-                collisionBox.vertices.emplace_back(CollisionBoxProgram::vertex(anchor, br, maxZoom, placementZoom),
-                                                   CollisionBoxProgram::vertex(anchor, bl, maxZoom, placementZoom));
-                collisionBox.vertices.emplace_back(CollisionBoxProgram::vertex(anchor, bl, maxZoom, placementZoom),
-                                                   CollisionBoxProgram::vertex(anchor, tl, maxZoom, placementZoom));
+                if (collisionBox.segments.empty() || collisionBox.segments.back().vertexLength + 4 > std::numeric_limits<uint16_t>::max()) {
```
brunoabinader: Can we add a `constexpr vertexLength` instead of the raw `4` to match other bucket implementations e.g. [here](https://github.com/mapbox/mapbox-gl-native/blob/master/src/mbgl/renderer/circle_bucket.cpp#L34)?

PR #7056, opened 2016-11-14
Return to static, per-segment approach to VAOs

jfirebaugh:
Per https://github.com/mapbox/mapbox-gl-native/pull/6468#discussion_r85182731

This is safer now -- it can be an implementation detail of Segment/Context. And the typing of SegmentVector now ensures that the attributes and program match.

Also, we've determined the cause of #5731.

cc @kkaefer

### DIFF
--- a/src/mbgl/gl/segment.hpp
+++ b/src/mbgl/gl/segment.hpp
@@ -23,5 +23,18 @@ public:
     std::size_t indexLength;
 };
 
+template <class Attributes>
+class SegmentVector : public std::vector<Segment> {
+public:
+    SegmentVector() = default;
+
+    // This constructor is for unindexed rendering. It creates a SegmentVector with a
+    // single segment having 0 indexes.
+    template <class DrawMode>
+    SegmentVector(const VertexBuffer<typename Attributes::Vertex, DrawMode>& buffer) {
+        emplace_back(0, 0, buffer.vertexCount, 0);
+    }
+};
+
 } // namespace gl
 } // namespace mbgl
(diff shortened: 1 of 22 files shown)


src/mbgl/programs/raster_program.hpp
```diff
@@ -23,12 +23,14 @@ MBGL_DEFINE_UNIFORM_VECTOR(float, 3, u_spin_weights);
 MBGL_DEFINE_UNIFORM_VECTOR(float, 2, u_tl_parent);
 } // namespace uniforms
 
+using RasterAttributes = gl::Attributes<
+    attributes::a_pos,
+    attributes::a_texture_pos>;
+
```
kkaefer: 👍

src/mbgl/gl/segment.hpp
```diff
@@ -21,6 +23,21 @@ class Segment {
...
+
+template <class Attributes>
+class SegmentVector : public std::vector<Segment> {
+public:
+    SegmentVector() = default;
+
+    template <class DrawMode>
+    SegmentVector(const VertexBuffer<typename Attributes::Vertex, DrawMode>& buffer) {
+        emplace_back(0, 0, buffer.vertexCount, 0);
+    }
```
kkaefer: Did we revert to inserting empty elements from the beginning?

jfirebaugh: No, this constructor is only for unindexed rendering. I'll add a comment to clarify.

PR #7060, opened 2016-11-14
fixes #7059: raise exception for abstract sources on add

### LINKED ISSUE #7059: Raise exception when adding generic MGLSource to style
#6793 removed [an exception](https://github.com/mapbox/mapbox-gl-native/pull/6793/files#diff-67773c92ddfc9da0c5e67c35d6a988d8L21) that presented the developer with a helpful message if they tried to add a source to the style that was created as an MGLSource rather than a concrete subclass. Now the developer gets an unrecognized selector exception instead. The exception should be added to either `-[MGLStyle addSource:]` (for parallelism with `-addLayer:` and `-insertLayer:belowLayer:`) or `-[MGLSource addToMapView:]` (which is currently declared but unimplemented).

/cc @incanus @boundsj

incanus:


boundsj:
@incanus can you please cherry-pick this into the `release-ios-v3.4.0` branch?

1ec5:
And when you do, note that the removeSource method needs similar treatment.

incanus:
Ah, crud. I'll get both.

Issue #6760, opened 2016-10-19
SDK runtime styling enumerations don’t match mbgl equivalents


1ec5:
`mbgl::LineJoinType` has three values intended for public use, in order:
- Miter
- Bevel
- Round

By contrast, the iOS and macOS SDK’s `MGLLineJoin` type has:
- MGLLineJoinBevel
- MGLLineJoinRound
- MGLLineJoinMiter

So if you try to get the `lineJoin` of an MGLLineStyleLayer that has `line-join` set to `round` in the style JSON, you get an MGLStyleValue containing `MGLLineJoinMiter`, because MGLStyleValueTransformer assumes that the mbgl type was defined in the same order as the style specification. But the style specification defines these values in an object, so it’s unsafe to assume any order in the first place.

MGLStyleValueTransformer needs to explicitly map style JSON string values to Objective-C enumeration values. In this case, we’d use `mbgl::LineJoinType_names`. We should also add assertions or tests to catch these issues earlier: #6464.

The Android SDK doesn’t appear to suffer from this issue because it uses string constants throughout.

/cc @frederoni @jfirebaugh @bsudekum

1ec5:
> The Android SDK doesn’t appear to suffer from this issue because it uses string constants throughout.

Actually, I’m not sure if I follow the Android SDK source code well enough to draw this conclusion. @ivovandongen, can you verify that the Android SDK doesn’t assume a certain order for enumeration types?

ivovandongen:
@1ec5 The Android API indeed uses strings throughout. The [constants](https://github.com/mapbox/mapbox-gl-native/blob/dd7713f6ee55a15b4ec5228a46a13c184dfdc01f/platform/android/MapboxGLAndroidSDK/src/main/java/com/mapbox/mapboxsdk/style/layers/Property.java#L76-L76
) are defined in Java (generated). This is passed through [jni](https://github.com/mapbox/mapbox-gl-native/blob/fe2a26225f3746381b36ad8b6c6a3ce7727bf655/platform/android/src/style/layers/layer.cpp#L48-L48) to the [conversion templates](https://github.com/mapbox/mapbox-gl-native/blob/fe2a26225f3746381b36ad8b6c6a3ce7727bf655/include/mbgl/style/conversion/layer.hpp#L20). So there is no redefinition of an enum in that sense.

1ec5:
> In this case, we’d use `mbgl::LineJoinType_names`.

[...]

PR #7061, opened 2016-11-15
Map Darwin style enumerations to mbgl equivalents

boundsj:
Fixes https://github.com/mapbox/mapbox-gl-native/issues/6760

### DIFF
--- a/platform/darwin/scripts/generate-style-code.js
+++ b/platform/darwin/scripts/generate-style-code.js
@@ -301,6 +301,8 @@ global.mbglType = function(property) {
 const layerH = ejs.compile(fs.readFileSync('platform/darwin/src/MGLStyleLayer.h.ejs', 'utf8'), { strict: true });
 const layerM = ejs.compile(fs.readFileSync('platform/darwin/src/MGLStyleLayer.mm.ejs', 'utf8'), { strict: true});
 const testLayers = ejs.compile(fs.readFileSync('platform/darwin/src/MGLRuntimeStylingTests.m.ejs', 'utf8'), { strict: true});
+const categoryH = ejs.compile(fs.readFileSync('platform/darwin/src/NSValue+MGLStyleEnumAttributeAdditions.h.ejs', 'utf8'), { strict: true});
+const categoryM = ejs.compile(fs.readFileSync('platform/darwin/src/NSValue+MGLStyleEnumAttributeAdditions.mm.ejs', 'utf8'), { strict: true});
 
 const layers = Object.keys(spec.layer.type.values).map((type) => {
     const layoutProperties = Object.keys(spec[`layout_${type}`]).reduce((memo, name) => {
@@ -345,8 +347,28 @@ ${macosComment}${decl}
     });
 }
 
+var allLayoutProperties = [];
+var allPaintProperties = [];
+var allTypes = [];
+
 for (var layer of layers) {
+    allLayoutProperties.push(layer.layoutProperties);
+    allPaintProperties.push(layer.paintProperties);
+    allTypes.push(layer.type);
+    const containsEnumerationProperties =  _.filter(layer.layoutProperties, function(property){ return property["type"] === "enum"; }).length  || _.filter(layer.paintProperties, function(property){ return property["type"] === "enum"; }).length;
+    layer.containsEnumerationProperties = containsEnumerationProperties;
+    
     fs.writeFileSync(`platform/darwin/src/${prefix}${camelize(layer.type)}${suffix}.h`, duplicatePlatformDecls(layerH(layer)));
     fs.writeFileSync(`platform/darwin/src/${prefix}${camelize(layer.type)}${suffix}.mm`, layerM(layer));
     fs.writeFileSync(`platform/darwin/test/${prefix}${camelize(layer.type)}${suffix}Tests.m`, testLayers(layer));
 }
+
+fs.writeFileSync(`platform/darwin/src/NSValue+MGLStyleEnumAttributeAdditions.h`, categoryH({
+    layoutProperties: _.flatten(allLayoutProperties),
+    paintProperties: _.flatten(allPaintProperties),
+    types: allTypes
+}));
+fs.writeFileSync(`platform/darwin/src/NSValue+MGLStyleEnumAttributeAdditions.mm`, categoryM({
+    layoutProperties: _.flatten(allLayoutProperties),
+    paintProperties: _.flatten(allPaintProperties)
+}));
--- a/platform/darwin/src/MGLBackgroundStyleLayer.mm
+++ b/platform/darwin/src/MGLBackgroundStyleLayer.mm
@@ -10,6 +10,7 @@
 
 #include <mbgl/style/layers/background_layer.hpp>
 
+
 @interface MGLBackgroundStyleLayer ()
 
 @property (nonatomic) mbgl::style::BackgroundLayer *rawLayer;
@@ -26,12 +27,10 @@
     if (self = [super initWithIdentifier:identifier]) {
         auto layer = std::make_unique<mbgl::style::BackgroundLayer>(identifier.UTF8String);
         _pendingLayer = std::move(layer);
-        self.rawLayer = _pendingLayer.get();
+        _rawLayer = _pendingLayer.get();
     }
     return self;
 }
-
-
 
 #pragma mark -  Adding to and removing from a map view
 
@@ -54,38 +53,38 @@
 {
     auto removedLayer = mapView.mbglMap->removeLayer(self.identifier.UTF8String);
     _pendingLayer = std::move(reinterpret_cast<std::unique_ptr<mbgl::style::BackgroundLayer> &>(removedLayer));
-    self.rawLayer = _pendingLayer.get();
+    _rawLayer = _pendingLayer.get();
 }
 
 #pragma mark - Accessing the Paint Attributes
 
 - (void)setBackgroundColor:(MGLStyleValue<MGLColor *> *)backgroundColor {
     auto mbglValue = MGLStyleValueTransformer<mbgl::Color, MGLColor *>().toPropertyValue(backgroundColor);
-    self.rawLayer->setBackgroundColor(mbglValue);
+    _rawLayer->setBackgroundColor(mbglValue);
 }
 
 - (MGLStyleValue<MGLColor *> *)backgroundColor {
-    auto propertyValue = self.rawLayer->getBackgroundColor() ?: self.rawLayer->getDefaultBackgroundColor();
+    auto propertyValue = _rawLayer->getBackgroundColor() ?: _rawLayer->getDefaultBackgroundColor();
     return MGLStyleValueTransformer<mbgl::Color, MGLColor *>().toStyleValue(propertyValue);
 }
 
 - (void)setBackgroundPattern:(MGLStyleValue<NSString *> *)backgroundPattern {
     auto mbglValue = MGLStyleValueTransformer<std::string, NSString *>().toPropertyValue(backgroundPattern);
-    self.rawLayer->setBackgroundPattern(mbglValue);
+    _rawLayer->setBackgroundPattern(mbglValue);
 }
 
 - (MGLStyleValue<NSString *> *)backgroundPattern {
-    auto propertyValue = self.rawLayer->getBackgroundPattern() ?: self.rawLayer->getDefaultBackgroundPattern();
+    auto propertyValue = _rawLayer->getBackgroundPattern() ?: _rawLayer->getDefaultBackgroundPattern();
     return MGLStyleValueTransformer<std::string, NSString *>().toStyleValue(propertyValue);
 }
 
 - (void)setBackgroundOpacity:(MGLStyleValue<NSNumber *> *)backgroundOpacity {
     auto mbglValue = MGLStyleValueTransformer<float, NSNumber *>().toPropertyValue(backgroundOpacity);
-    self.rawLayer->setBackgroundOpacity(mbglValue);
+    _rawLayer->setBackgroundOpacity(mbglValue);
 }
 
 - (MGLStyleValue<NSNumber *> *)backgroundOpacity {
-    auto propertyValue = self.rawLayer->getBackgroundOpacity() ?: self.rawLayer->getDefaultBackgroundOpacity();
+    auto propertyValue = _rawLayer->getBackgroundOpacity() ?: _rawLayer->getDefaultBackgroundOpacity();
     return MGLStyleValueTransformer<float, NSNumber *>().toStyleValue(propertyValue);
 }
 
--- a/platform/darwin/src/MGLBackgroundStyleLayer.mm
+++ b/platform/darwin/src/MGLBackgroundStyleLayer.mm
@@ -10,7 +10,6 @@
 
 #include <mbgl/style/layers/background_layer.hpp>
 
-
 @interface MGLBackgroundStyleLayer ()
 
 @property (nonatomic) mbgl::style::BackgroundLayer *rawLayer;
--- a/platform/darwin/src/MGLCircleStyleLayer.mm
+++ b/platform/darwin/src/MGLCircleStyleLayer.mm
@@ -10,6 +10,20 @@
 
 #include <mbgl/style/layers/circle_layer.hpp>
 
+namespace mbgl {
+
+    MBGL_DEFINE_ENUM(MGLCircleTranslateAnchor, {
+        { MGLCircleTranslateAnchorMap, "map" },
+        { MGLCircleTranslateAnchorViewport, "viewport" },
+    });
+
+    MBGL_DEFINE_ENUM(MGLCirclePitchScale, {
+        { MGLCirclePitchScaleMap, "map" },
+        { MGLCirclePitchScaleViewport, "viewport" },
+    });
+
+}
+
 @interface MGLCircleStyleLayer ()
 
 @property (nonatomic) mbgl::style::CircleLayer *rawLayer;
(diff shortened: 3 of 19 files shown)


platform/darwin/src/MGLLineStyleLayer.mm
```diff
@@ -71,8 +71,8 @@ - (void)setLineJoin:(MGLStyleValue<NSValue *> *)lineJoin {
 }
 
 - (MGLStyleValue<NSValue *> *)lineJoin {
-    auto propertyValue = self.rawLayer->getLineJoin() ?: self.rawLayer->getDefaultLineJoin();
-    return MGLStyleValueTransformer<mbgl::style::LineJoinType, NSValue *>().toStyleValue(propertyValue);
+    auto propertyValue = self.rawLayer->getLineJoin() ?: self.rawLayer->getDefaultLineJoin();
```
1ec5: Nit: trailing whitespace.

platform/darwin/src/MGLStyleValue_Private.h
```diff
@@ -26,7 +28,24 @@ class MGLStyleValueTransformer {
             return nil;
         }
     }
-    
+
+    MGLStyleValue<ObjCType> *toLineJoinStyleValue(const mbgl::style::PropertyValue<MBGLType> &mbglValue) {
+        auto str = mbgl::Enum<MBGLType>::toString(mbglValue.asConstant());
+        NSString *enumString = [NSString stringWithCString:str encoding:NSUTF8StringEncoding];
```
1ec5: Nit: this can be expressed as `str.UTF8String`.

platform/darwin/src/MGLStyleValue_Private.h
```diff
@@ -26,7 +28,24 @@ class MGLStyleValueTransformer {
...
+        NSUInteger enumValue;
+        if ([enumString isEqualToString:@"miter"]) {
```
1ec5: Nice! Using `mbgl::Enum` is a lot cleaner than what I thought we’d have to resort to (the `_names` macros).

I think we could take further advantage of `mbgl::Enum`, using `MBGL_DEFINE_ENUM()` to map e.g. `{ MGLLineJoinMiter, "miter" }` and calling `toEnum()` on the string. That way, we can avoid going the NSString conversion and these if statements, and we can generate everything using the style specification (in a separate file).

boundsj:
@1ec5 thanks for the advice in https://github.com/mapbox/mapbox-gl-native/pull/7061#discussion_r87955273. Most of that is implemented now although I did not DRY it up yet with anything like `toEnumStyleValue` and templates. Other than that though, I think this is pretty close now.

The lookup table in https://github.com/mapbox/mapbox-gl-native/pull/7061/files#diff-01213d8ba02156a3771961755cf10994R12 is sort of unfortunate. But, since the MBGL enums are not generated and because several of them are coalesced into things like `AlignmentType` (with `auto`) it is challenging to generate code with the raw hooks we have at hand.

Anyway, this is working!

platform/darwin/scripts/generate-style-code.js
```diff
@@ -9,6 +9,30 @@ const spec = _.merge(require('mapbox-gl-style-spec').latest, require('./style-sp
 const prefix = 'MGL';
 const suffix = 'StyleLayer';
 
+global.mbglTypeName = function(str) {
```
boundsj: Looks like this can be replaced with `global.mbglType()`

boundsj:
This will also fix https://github.com/mapbox/mapbox-gl-native/issues/5970 (again)

boundsj:
@1ec5 @frederoni @incanus please review

platform/darwin/src/MGLBackgroundStyleLayer.mm
```diff
@@ -10,6 +10,13 @@
 
 #include <mbgl/style/layers/background_layer.hpp>
 
+namespace mbgl {
```
1ec5: Can we get rid of this block if there’s are no symbols to define in this translation unit?

platform/darwin/src/MGLCircleStyleLayer.mm
```diff
@@ -10,6 +10,21 @@
 
 #include <mbgl/style/layers/circle_layer.hpp>
 
+namespace mbgl {
+
+    using namespace style;
+
+
```
1ec5: Nit: extra blank line.

platform/darwin/src/MGLCircleStyleLayer.mm
```diff
@@ -10,6 +10,21 @@
 
 #include <mbgl/style/layers/circle_layer.hpp>
 
+namespace mbgl {
+
+    using namespace style;
```
1ec5: Nothing in this block is using the `mbgl::style` namespace, so this statement can go away.

platform/darwin/src/MGLCircleStyleLayer.mm
```diff
@@ -75,79 +90,121 @@ - (void)removeFromMapView:(MGLMapView *)mapView
...
+            auto mbglStopValue = mbgl::style::PropertyValue<mbgl::style::TranslateAnchorType>(mbglValue);
+            mbglStops.emplace_back(zoomKey.floatValue, mbglStopValue.asConstant());
+        }];
+        auto func = mbgl::style::Function<mbgl::style::TranslateAnchorType>({{mbglStops}}, function.base);
+        _rawLayer->setCircleTranslateAnchor(func);
+        return;
+    }
+    id value = [(MGLStyleConstantValue<NSValue *> *)circleTranslateAnchor rawValue];
+    MGLCircleTranslateAnchor circleTranslateAnchorValue;
+    [value getValue:&circleTranslateAnchorValue];
```
1ec5: In the previous code, if `circleTranslateAnchor` was `nil`, we’d [reset](https://github.com/mapbox/mapbox-gl-native/blob/307f8e4f41a440fdfcd9c7bb1833110ab9278ccf/platform/darwin/src/MGLStyleValue_Private.h#L52) the value in mbgl to an implicit default value. In this code, if `circleTranslateAnchor` is `nil`, `circleTranslateAnchorValue` will be `MGLCircleTranslateAnchorMap`, even if the implicit default would’ve been `mbgl::style::TranslateAnchorType::Viewport`.

platform/darwin/src/MGLCircleStyleLayer.mm
```diff
@@ -75,79 +90,121 @@ - (void)removeFromMapView:(MGLMapView *)mapView
...
 - (MGLStyleValue<NSValue *> *)circleTranslate {
-    auto propertyValue = self.rawLayer->getCircleTranslate() ?: self.rawLayer->getDefaultCircleTranslate();
+    auto propertyValue = _rawLayer->getCircleTranslate() ?: _rawLayer->getDefaultCircleTranslate();
     return MGLStyleValueTransformer<std::array<float, 2>, NSValue *>().toStyleValue(propertyValue);
 }
 
 - (void)setCircleTranslateAnchor:(MGLStyleValue<NSValue *> *)circleTranslateAnchor {
-    auto mbglValue = MGLStyleValueTransformer<mbgl::style::TranslateAnchorType, NSValue *>().toPropertyValue(circleTranslateAnchor);
-    self.rawLayer->setCircleTranslateAnchor(mbglValue);
+    if ([circleTranslateAnchor isKindOfClass:[MGLStyleFunction class]]) {
```
1ec5: This code should be brought back into `MGLStyleValueTransformer` and made generic. Bring back the enumeration specialization of `toMGLRawStyleValue()`, but add an additional, optional `ObjCEnum` template parameter to `MGLStyleValueTransformer` that you can plug into `mbgl::Enum`. For additional safety, I think you can also make the specialization conditional on `ObjCEnum` satisfying `std::is_enum` just like it required `MBGLEnum` to satisfy `std::is_enum`.

platform/darwin/src/NSValue+MGLStyleEnumAttributeAdditions.h
```diff
@@ -0,0 +1,150 @@
...
+#import "MGLFillStyleLayer.h"
+#import "MGLLineStyleLayer.h"
+#import "MGLSymbolStyleLayer.h"
+#import "MGLCircleStyleLayer.h"
+#import "MGLRasterStyleLayer.h"
+#import "MGLBackgroundStyleLayer.h"
+
+NS_ASSUME_NONNULL_BEGIN
+
+@interface NSValue (MGLStyleEnumAttributeAdditions)
```
1ec5: realm/jazzy#539 means that either all of NSValue (MGLAdditions) will wind up in this category, or all of this category will wind up in NSValue (MGLAdditions). Hopefully the latter.

platform/darwin/src/NSValue+MGLStyleEnumAttributeAdditions.h
```diff
@@ -0,0 +1,150 @@
...
+
+#pragma mark Runtime Styling enumerations
+
+/**
+Creates a new value object containing the given `MGLLineCap` structure.
```
1ec5: `MGLLineCap` is an enumeration, not a structure.

platform/darwin/src/NSValue+MGLStyleEnumAttributeAdditions.h
```diff
@@ -0,0 +1,150 @@
...
+
+@param type The value for the new object.
+@return A new value object that contains the style enumeration type.
+*/
++ (instancetype)valueWithMGLLineCap:(MGLLineCap)type;
```
1ec5: Rename `type` to `lineCap`.

platform/darwin/src/NSValue+MGLStyleEnumAttributeAdditions.h
```diff
@@ -0,0 +1,150 @@
...
+/**
+Creates a new value object containing the given `MGLCirclePitchScale` structure.
+
+@param type The value for the new object.
+@return A new value object that contains the style enumeration type.
+*/
++ (instancetype)valueWithMGLCirclePitchScale:(MGLCirclePitchScale)type;
+
+
+NS_ASSUME_NONNULL_END
```
1ec5: This keyword should go after the `@end`.

platform/darwin/src/NSValue+MGLStyleEnumAttributeAdditions.h
```diff
@@ -0,0 +1,150 @@
...
+
+/**
+Creates a new value object containing the given `MGLCirclePitchScale` structure.
+
+@param type The value for the new object.
+@return A new value object that contains the style enumeration type.
+*/
++ (instancetype)valueWithMGLCirclePitchScale:(MGLCirclePitchScale)type;
+
+
```
1ec5: Each factory method should come with a readonly property that converts back into the raw type ([example](https://github.com/mapbox/mapbox-gl-native/blob/307f8e4f41a440fdfcd9c7bb1833110ab9278ccf/platform/darwin/src/NSValue%2BMGLAdditions.h#L27)). This way you (and the developer) won’t ever have to use `-getValue:`.

platform/darwin/src/NSValue+MGLStyleEnumAttributeAdditions.h
```diff
@@ -0,0 +1,150 @@
...
+#import "MGLSymbolStyleLayer.h"
+#import "MGLCircleStyleLayer.h"
+#import "MGLRasterStyleLayer.h"
+#import "MGLBackgroundStyleLayer.h"
+
+NS_ASSUME_NONNULL_BEGIN
+
+@interface NSValue (MGLStyleEnumAttributeAdditions)
+
+#pragma mark Runtime Styling enumerations
```
1ec5: A documentation section name (that is,  a`mark` in a public header) should describe a task, in title case for consistency:

```objc
#pragma mark Working with Style Layer Attribute Values
```

platform/ios/app/MBXViewController.m
```diff
@@ -816,8 +817,61 @@ - (void)styleFilteredFill
...
+{
+    return [MGLStyleFunction<NSValue *> valueWithStops:@{
+                                                         @18: [self testEnum:value type:type],
+                                                         }];
+}
+
 - (void)styleFilteredLines
 {
+
+#warning remove this test stub
```
1ec5: Optional: We currently lack a demo in iosapp that draws a route line. We could add one that sets the line join and line cap to something that looks better than the default.

platform/ios/ios.xcodeproj/project.pbxproj
```diff
@@ -127,6 +127,10 @@
 		4018B1C91CDC288A00F666AF /* MGLAnnotationView_Private.h in Headers */ = {isa = PBXBuildFile; fileRef = 4018B1C31CDC277F00F666AF /* MGLAnnotationView_Private.h */; };
 		4018B1CA1CDC288E00F666AF /* MGLAnnotationView.h in Headers */ = {isa = PBXBuildFile; fileRef = 4018B1C51CDC277F00F666AF /* MGLAnnotationView.h */; settings = {ATTRIBUTES = (Public, ); }; };
 		4018B1CB1CDC288E00F666AF /* MGLAnnotationView.h in Headers */ = {isa = PBXBuildFile; fileRef = 4018B1C51CDC277F00F666AF /* MGLAnnotationView.h */; settings = {ATTRIBUTES = (Public, ); }; };
+		4032C5BF1DE1FC780062E8BD /* NSValue+MGLStyleEnumAttributeAdditions.h in Headers */ = {isa = PBXBuildFile; fileRef = 4032C5B81DE1EE7D0062E8BD /* NSValue+MGLStyleEnumAttributeAdditions.h */; };
```
1ec5: These headers should be public.

platform/ios/app/MBXViewController.m
```diff
@@ -5,6 +5,7 @@
 #import "MBXOfflinePacksTableViewController.h"
 #import "MBXAnnotationView.h"
 #import "MBXUserLocationAnnotationView.h"
+#import "NSValue+MGLStyleEnumAttributeAdditions.h"
```
1ec5: Import this header in both SDKs’ umbrella headers instead of here, since only iosapp and macosapp have this file in their header search paths.

platform/darwin/src/MGLStyleValue_Private.h
```diff
@@ -13,6 +15,35 @@
...
+    {
+        if (propertyValue.isFunction())
+        {
+            const auto &mbglStops = propertyValue.asFunction().getStops();
+            NSMutableDictionary *stops = [NSMutableDictionary dictionaryWithCapacity:mbglStops.size()];
+            for (const auto &mbglStop : mbglStops) {
+                auto str = mbgl::Enum<MBGLType>::toString(mbglStop.second);
+                MGLType mglType;
+                mglType = mbgl::Enum<MGLType>::toEnum(str).value_or(mglType);
+                stops[@(mbglStop.first)] = [MGLStyleValue valueWithRawValue:[NSValue value:&mglType withObjCType:@encode(NSUInteger)]];
```
1ec5: Encode `MGLType` instead of `NSUInteger`, if possible.

platform/darwin/src/MGLStyleValue_Private.h
```diff
@@ -13,6 +15,35 @@
 
 #include <array>
 
+template <typename MBGLType, typename MGLType>
+class MGLStyleEnumerationValueTransformer {
```
1ec5: Can this method live inside `MGLStyleValueTransformer`? Use optional template parameters on the class to allow both `NSValue *` and `MGLLineCap` to be specific, and use conditional template parameters on the enumeration-specific methods to enable the methods only for enumerations.

platform/darwin/src/NSValue+MGLStyleEnumAttributeAdditions.h
boundsj: Actually I only see (even in our current release version) one "other" category https://www.mapbox.com/ios-sdk/api/3.3.7/Other%20Categories.html

1ec5: Run `make idocument` and verify that the sole documented NSValue category isn’t called `MGLStyleEnumAttributeAdditions`.

The existing documented category should be called `MGLAdditions`, not `MGLGeometryAdditions`. While you’re at it, can you remove the rump `MGLGeometryAdditions` category in MGLGeometry.h?

1ec5: → #7153

boundsj:
This will also fix #7153

platform/darwin/src/NSValue+MGLStyleEnumAttributeAdditions.h
boundsj: screenshot 2016-11-22 11 26 39

platform/darwin/src/MGLCircleStyleLayer.mm
```diff
@@ -75,79 +85,85 @@ - (void)removeFromMapView:(MGLMapView *)mapView
...
-    auto propertyValue = self.rawLayer->getCircleTranslate() ?: self.rawLayer->getDefaultCircleTranslate();
+    auto propertyValue = _rawLayer->getCircleTranslate() ?: _rawLayer->getDefaultCircleTranslate();
     return MGLStyleValueTransformer<std::array<float, 2>, NSValue *>().toStyleValue(propertyValue);
 }
 
 - (void)setCircleTranslateAnchor:(MGLStyleValue<NSValue *> *)circleTranslateAnchor {
-    auto mbglValue = MGLStyleValueTransformer<mbgl::style::TranslateAnchorType, NSValue *>().toPropertyValue(circleTranslateAnchor);
-    self.rawLayer->setCircleTranslateAnchor(mbglValue);
+    auto mbglValue = MGLStyleValueTransformer<mbgl::style::TranslateAnchorType,
+    NSValue *,
```
1ec5: Nit: funky indentation. Can this all go on one line?

PR #7083, opened 2016-11-15
[ios, macos] Use NSNumber in exception strings

boundsj:
Avoid precision loss / cast warnings.

cc @incanus @1ec5

### DIFF
--- a/platform/darwin/src/MGLMultiPoint.mm
+++ b/platform/darwin/src/MGLMultiPoint.mm
@@ -58,8 +58,8 @@ mbgl::Color MGLColorObjectFromCGColorRef(CGColorRef cgColor)
     if (range.location + range.length > [self pointCount])
     {
         [NSException raise:NSRangeException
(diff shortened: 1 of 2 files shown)


platform/darwin/src/MGLMultiPoint.mm
```diff
@@ -58,8 +58,8 @@ - (void)getCoordinates:(CLLocationCoordinate2D *)coords range:(NSRange)range
     if (range.location + range.length > [self pointCount])
     {
         [NSException raise:NSRangeException
-                    format:@"Invalid coordinate range %@ extends beyond current coordinate count of %zu",
-                        NSStringFromRange(range), [self pointCount]];
+                    format:@"Invalid coordinate range %@ extends beyond current coordinate count of %@",
+                        NSStringFromRange(range), @([self pointCount])];
```
1ec5: This isn’t necessary. There should be a fix-it that casts `self.pointCount` to `unsigned long`.

Issue #6954, opened 2016-11-08
Compass underlaps top bar


1ec5:
In a standard view controller with a top bar and `automaticallyAdjustsScrollViewInsets` set to `YES`, the compass underlaps the top bar:

<img src="https://cloud.githubusercontent.com/assets/1231218/20088005/30cf22c8-a532-11e6-9937-7088d3259fae.png" alt="underlap" width="270">

This is almost certainly a regression from iOS SDK v3.3.x. Relevant PRs from this release cycle: #6216 #6313 #6566.

/cc @boundsj @incanus

1ec5:
This is causing KIF tests to fail:

```
Test: testBottomLayoutGuide in MapViewTests failed on iPhone 7
Assertions: ((CGRectIntersectsRect(logoBugFrame, toolbarFrame)) is false) failed - logo bug should not be under toolbar
File: MapViewTests.m:386

Test: testTopLayoutGuide in MapViewTests failed on iPhone 7
Assertions: ((CGRectIntersectsRect(compassFrame, navigationBarFrame)) is false) failed - compass should not be under navigation bar
File: MapViewTests.m:359
```

frederoni:
I think this happens because we don't handle device rotation properly or there's a bug in the topLayoutGuide.

The initial height of the topLayoutGuide is 64, rotate to landscape and it's 44, rotate back to portrait and we have 52. Additional rotations from here will switch between 44 and 52.

Here's two workarounds:

**1. Respond to device rotation in the containing view controller**
```objc
// MBXViewController.m
- (void)viewDidLayoutSubviews
{
    [super viewDidLayoutSubviews];
    [self.mapView setNeedsLayout];
}
```

**2. Generate device orientation notifications in MGLMapView**
```objc
// MGLMapView.m
[[UIDevice currentDevice] beginGeneratingDeviceOrientationNotifications];
[[NSNotificationCenter defaultCenter] addObserver:self selector:@selector(deviceOrientationDidChange:) name:UIDeviceOrientationDidChangeNotification object:nil];

- (void)deviceOrientationDidChange:(NSNotification)notification
{
    [self setNeedsLayout];
}
```

I'd prefer none of the above so I'll dig a bit more.  🐰 🕳

frederoni:
Decided to go with the latter because it works well and it's a fix in the SDK unlike the first workaround. Implemented in https://github.com/mapbox/mapbox-gl-native/pull/7004

[...]

PR #7084, opened 2016-11-15
[ios] simplified constraints and fixed regression 6954

frederoni:
Fixes regression https://github.com/mapbox/mapbox-gl-native/issues/6954#issuecomment-260768209

There's no need to attach the constraints to the superview as mentioned in the comment. This fixes the compass being out of bounds when not in full screen.

@1ec5 👀

### DIFF
--- a/platform/ios/src/MGLMapView.mm
+++ b/platform/ios/src/MGLMapView.mm
@@ -763,26 +763,16 @@ public:
 
 - (void)updateConstraints
 {
-    // If we have a view controller reference and its automaticallyAdjustsScrollViewInsets
-    // is set to YES, use its view as the parent for constraints. -[MGLMapView adjustContentInset]
-    // already take top and bottom layout guides into account. If we don't have a reference, apply
-    // constraints against ourself to maintain placement of the subviews.
-    //
-    UIViewController *viewController = self.viewControllerForLayoutGuides;
-    BOOL useLayoutGuides = viewController.view && viewController.automaticallyAdjustsScrollViewInsets;
-    UIView *view = useLayoutGuides ? viewController.view : self;
-    
     // compass
     //
-    UIView *compassContainer = self.compassView.superview;
-    [view removeConstraints:self.compassViewConstraints];
+    [self removeConstraints:self.compassViewConstraints];
     [self.compassViewConstraints removeAllObjects];
 
     [self.compassViewConstraints addObject:
-     [NSLayoutConstraint constraintWithItem:compassContainer
+     [NSLayoutConstraint constraintWithItem:self.compassView
                                   attribute:NSLayoutAttributeTop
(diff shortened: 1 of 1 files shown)


platform/ios/src/MGLMapView.mm
```diff
@@ -763,26 +763,16 @@ - (UIViewController *)viewControllerForLayoutGuides
...
     // compass
     //
-    UIView *compassContainer = self.compassView.superview;
-    [view removeConstraints:self.compassViewConstraints];
+    [self removeConstraints:self.compassViewConstraints];
     [self.compassViewConstraints removeAllObjects];
 
     [self.compassViewConstraints addObject:
-     [NSLayoutConstraint constraintWithItem:compassContainer
+     [NSLayoutConstraint constraintWithItem:self.compassView
```
1ec5: So the compass container view isn’t constrained to anything anymore? Do we still need it?

@incanus, do you remember why we stuck the compass view inside a container view in the first place?

incanus: > @incanus, do you remember why we stuck the compass view inside a container view in the first place?

Indeed! When you don't have it in a container, but constrain it to the corner using margins, it jiggles as the constraints attempt to keep up with the actual margins as they change. You'd think this would not be the case for a round image, but it is. So constraining the container keeps it solidly in place, changing only for device rotation and superview frame changes, while letting the round part rotate about its center, which is firmly constrained to the (from its perspective) unchanging container. 

So, if you get rid of the container, make sure the compass doesn't jiggle. It may have been a major iOS version-specific bug, so we should check all supported versions. 

https://github.com/mapbox/mi-sdk-legacy/commit/f7fdc9decaa6ad560969c5d2fb7ea7e3642a3159#diff-cc99abd5af5db8f48b362bcac3e8c307R514

frederoni: compassContainer was just a shorthand to the compass view's superview. There's no point in doing that since it's always self.

frederoni: > When you don't have it in a container, but constrain it to the corner using margins, it jiggles as the constraints attempt to keep up with the actual margins as they change.

Indeed that might've been the case in the early retina days. I've seen the same problems when using half pixels (compass is 34.5 w/h).

I removed the container view and the w/h constraints because UIImageView has intrinsic content size when initialized with an image or sizeToFit has been called and I can't see any regression.

1ec5: This simplification looks great, but before merging, can we get testing on iOS 8 (and ideally iOS 7, if possible, since it's still supported)?

incanus:
The original compass jiggle issue came about on iOS 7.

frederoni:
Unable to test on iOS7 so far but 8, 9 and 10 looks 👍

PR #7091, opened 2016-11-16
Raster tiles are not cached

kkaefer:
When abandoning geometry (vector) tiles, we're putting them into a [cache](https://github.com/mapbox/mapbox-gl-native/blob/5479b75c9fa971332aadfede6b380267eebbd566/src/mbgl/style/source_impl.cpp#L163-L180), but [for raster sources, that cache is of size 0](https://github.com/mapbox/mapbox-gl-native/blob/5479b75c9fa971332aadfede6b380267eebbd566/src/mbgl/style/source_impl.cpp#L142), so raster tiles get immediately deleted when they are moving outside of the viewport.

We should hold on to raster tiles as well to avoid flickering when they are getting moved out and back into the viewport.


1ec5:
> Would you do the honors and update the iOS and macOS changelogs?

Happening in #7100.

PR #7095, opened 2016-11-16
Don't use <regex> to reduce binary size

kkaefer:
We're only using `<regex>` in one instance, for removing `access_token=*` from URLs when putting them into the cache. However, `<regex>` adds quite a bit to the binary size; I'm measuring these size reductions:

- Android/arm-v7: 4,903,78 => 4,842,348 bytes, ~60KB reduction
- macOS/x86_64: 3,955,184 => 3,860,600 bytes, ~95KB reduction 
- iOS/arm64: 3,012,356 => 2,961,728 bytes, ~50KB reduction

### DIFF
--- a/src/mbgl/util/mapbox.cpp
+++ b/src/mbgl/util/mapbox.cpp
@@ -5,7 +5,6 @@
 #include <stdexcept>
 #include <vector>
 #include <iostream>
-#include <regex>
 
 namespace {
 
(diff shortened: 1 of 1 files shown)


mapsam:
This was very specifically added to ensure other parameters are preserved, and _only_ access token is removed. This is important for style-optimized vector tile requests that may have a `&style=` parameter in the tile URL. If the tests continue to pass, 👍 from me.

kkaefer:
@mapsam yup, the code just looks for access_token fields and removes them, while preserving all others.

mapsam:
Thanks @kkaefer!

PR #7098, opened 2016-11-16
[node] Revert semantics of Map#render

jfirebaugh:
Revert node `Map#render` API to "stateless" camera semantics: the center, zoom, bearing, and pitch passed are what is used to render, with defaults for unspecified values.

For the test-suite implementation, which wants stateful semantics, track the camera options manually. Also, change the implementation of `Map#render` so that it avoids calling `setCenter`, etc. when the value is not changing. (Internally, `setCenter` does not short-circuit, so it could cause undesired updates.)

cc @mikemorris

### DIFF
--- a/platform/node/test/suite_implementation.js
+++ b/platform/node/test/suite_implementation.js
@@ -36,6 +36,11 @@ module.exports = function (style, options, callback) {
         overdraw: options.showOverdrawInspector,
     };
 
+    options.center = style.center || [0, 0];
+    options.zoom = style.zoom || 0;
+    options.bearing = style.bearing || 0;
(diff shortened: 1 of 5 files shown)


platform/node/test/suite_implementation.js
```diff
@@ -66,6 +71,15 @@ module.exports = function (style, options, callback) {
             wait();
 
         } else {
+            // Ensure that the next `map.render(options)` does not overwrite this change.
```
mikemorris: Would a check for `setPitch` be necessary here too?

jfirebaugh: It's not required by test-suite yet, but might as well. ✅

PR #7105, opened 2016-11-17
[android] fixed horizontal panning going the opposite direction

### LINKED ISSUE #7104: Swiping horizontally causes the map to go the opposite direction
Swiping the map view horizontally quickly (triggering `onFling`) causes the map to animate in the opposite direction.

horizontal_master

frederoni:
Fixes #7104 

The root cause was that the nativeMapView's `gestureInProgress` was false during the fling.
This resulted in mbgl map unwrapping the longitude to find the shortest path > https://github.com/mapbox/mapbox-gl-native/blob/master/src/mbgl/map/transform.cpp#L111

Without the unwrapping, the deceleration rate was way too slow as well as the duration causing a simple fling to go around the world 50 times before stopping. The new deceleration rate mimics a scrollView with fast scrolling disabled.

horizontal_fix

@tobrun  👀


tobrun:
Work great @frederoni, thank you for looking into this!

PR #7114, opened 2016-11-18
Upright CJK characters in vertically-oriented labels

### LINKED ISSUE #1682: Vertical CJK labels

1ec5:
CJK characters and adjacent punctuation now remain upright in vertically oriented labels that have line placement.

This is a port of mapbox/mapbox-gl-js#3438 that fixes #1682. I did most of the porting before #6984 landed, so I have yet to reconcile the `WritingMode` enumeration with the new `WritingDirection` enumeration. Also, vertical punctuation is correctly laid out but doesn’t appear for some reason. I suspect the issue has to do with the glyph range not being loaded, but I need to do further investigation.

chinese

punctuation

shenzhen

/cc @kkaefer @lucaswoj @ChrisLoer @LornTao


kkaefer:
@1ec5 what needs to happen until this can be merged?

1ec5:
* [ ] Reconcile these changes with #6984
* [ ] Figure out why vertical punctuation glyphs don’t appear (don’t load?)

jfirebaugh:
@1ec5 I rebased this and debugged the test failures -- can you take a look?

src/mbgl/util/i18n.cpp
```diff
@@ -352,6 +374,188 @@ bool allowsIdeographicBreaking(uint16_t chr) {
...
+            return true;
+    }
+    if (isInUnifiedCanadianAboriginalSyllabics(chr) ||
+        isInUnifiedCanadianAboriginalSyllabicsExtended(chr) || isInVerticalForms(chr) ||
+        isInYijingHexagramSymbols(chr) || isInYiSyllables(chr) || isInYiRadicals(chr))
+        return true;
+
+    // https://github.com/mapbox/mapbox-gl/issues/29
+
+    // if (isInMeroiticHieroglyphs(chr)) return true;
```
1ec5: https://github.com/mapbox/mapbox-gl-js/pull/3608#discussion_r87851149 requested that I put the list of vertical supplementary blocks in mapbox/mapbox-gl#29, now mapbox/mapbox-gl-js#4001, instead of leaving them here as commented out lines.

PR #7118, opened 2016-11-18
Android core tests - rendering

ivovandongen:
First step in getting the core tests to run on Android.

TODO:
- [x] Generate assets zips so that they're up-to-date
- [x] Generate fixture results zip
- [x] Implement the log stub methods
- [x] Fixup commits
- [x] If possible, use some mechanism similar to `--quiet` or `> /dev/null 2>&1` in adb push to avoid it polluting the build log with lots of file upload percentage lines

Does not include:
- Http tests
- CI integration


brunoabinader:
@ivovandongen I've fixed up commits and made the zipped assets be generated on-the-fly. I've added some more TODO's. Questions:

- Can we make Bitrise test run faster? Last attempt took ~30min, and that didn't count time running the unit tests (/cc @tobrun)
- Can we obtain more available disk space on the emulator? Android test run is currently failing due to no space on disk left.

tobrun:
> Can we make Bitrise test run faster? Last attempt took ~30min, and that didn't count time running the unit tests (/cc @tobrun)

@brunoabinader capturing from conversations with @kkaefer, as first step I'm thinking of decoupling the monolith build into 2 separate builds. Initially they will both build core and execute their respective tests. As second step I want to look into building core once in one workflow and as results this will trigger two different workflows (one for core unit tests and one for firebase instrumentation tests). 

> Can we obtain more available disk space on the emulator? Android test run is currently failing due to no space on disk left.

This should be possible as an additional argument in the `create android emulator`-step. I will pick this up. Thank you for flagging!

tobrun:
In d654202 the emulator SD-card size was increased, the build is now failling with: 
> 
cd test/fixtures/annotations && zip -r assets.zip assets && cd -
	zip warning: name not matched: assets

brunoabinader:
> In d654202 the emulator SD-card size was increased, the build is now failling with:
That line is actually not needed, we're only reading zipped assets for `api` and `storage`. I'll update the patch 👍

tobrun:
Their seemed to be still issues with disksize, b693175 fixed this, build is currently failing with:
```
#Kick off the tests
adb shell "export LD_LIBRARY_PATH=/system/lib:/data/local/tmp/core-tests && cd /data/local/tmp/core-tests && dalvikvm32 -cp /data/local/tmp/core-tests/test.jar Main"
dex2oat W 2243 2244 dex2oat did not finish after 60 seconds
Aborted 
#Gather the results
adb shell "cd /data/local/tmp/core-tests && tar -cvzf results.tgz test/fixtures/*"
/system/bin/sh: tar: not found
adb pull /data/local/tmp/core-tests/results.tgz build/android-arm-v7/Debug/
adb: error: remote object '/data/local/tmp/core-tests/results.tgz' does not exist
Makefile:545: recipe for target 'run-android-core-test-arm-v7' failed
make: *** [run-android-core-test-arm-v7] Error 1
```

platform/android/config.cmake
```diff
@@ -161,6 +168,57 @@ target_link_libraries(mapbox-gl
...
+    PRIVATE platform/default/offscreen_view.cpp
+
+    PRIVATE platform/linux/src/headless_backend_egl.cpp
+    PRIVATE platform/linux/src/headless_display_egl.cpp
+)
+
+target_compile_options(mbgl-test
+    PRIVATE -fvisibility=hidden
+    PRIVATE -Os
+    PRIVATE -DMBGL_ASSET_ZIP=1
```
kkaefer: Can we move `-D` flags to `target_compile_definitions`?

platform/android/src/jni.cpp
```diff
@@ -1703,9 +1703,12 @@ void updateOfflineRegionMetadata(JNIEnv *env, jni::jobject* offlineRegion_, jni:
 
 // Offline calls end
 
-}
+} //anonymous
```
kkaefer: Style is to have a comment that says `// namespace`

platform/android/src/jni.cpp
```diff
@@ -1988,6 +1991,7 @@ extern "C" JNIEXPORT jint JNI_OnLoad(JavaVM *vm, void *reserved) {
     char release[PROP_VALUE_MAX] = "";
     __system_property_get("ro.build.version.release", release);
     androidRelease = std::string(release);
-
-    return JNI_VERSION_1_6;
 }
+    
+} //android
+} //mbgl
```
kkaefer: `// namespace mbgl`. Please add a space after all `//`.

platform/linux/src/headless_backend_egl.cpp
```diff
@@ -10,9 +10,24 @@
...
+        //Create a PBuffer surface (in conjunction with EGL_SURFACE_TYPE, EGL_PBUFFER_BIT)
+        const EGLint surfAttribs[] = {
+            EGL_WIDTH, 512,
+            EGL_HEIGHT, 512,
+            EGL_LARGEST_PBUFFER, EGL_TRUE,
+            EGL_NONE
+        };
+
+        glSurface = eglCreatePbufferSurface(display, config, surfAttribs);
+        if(glSurface == EGL_NO_SURFACE) {
```
kkaefer: Please add spaces after control statements

kkaefer: You can use `clang-format` with the formatting file in the root directory to format C++ code automatically.

platform/linux/src/headless_backend_egl.cpp
```diff
@@ -22,10 +37,16 @@ struct EGLImpl : public HeadlessBackend::Impl {
...
+            if (!eglDestroySurface(display, glSurface)) {
+                throw std::runtime_error("Failed to destroy EGL context.\n");
+            }
+            glSurface = EGL_NO_SURFACE;
+        }
     }
 
     void activateContext() final {
-        if (!eglMakeCurrent(display, EGL_NO_SURFACE, EGL_NO_SURFACE, glContext)) {
+        if (!eglMakeCurrent(display, glSurface, glSurface, glContext)) {
```
kkaefer: Will this work when no `glSurface` object is available?

platform/linux/src/headless_display_egl.cpp
```diff
@@ -31,8 +31,23 @@ HeadlessDisplay::Impl::Impl() {
         throw std::runtime_error("eglBindAPI() failed");
     }
 
+#if !defined(__ANDROID__)
     // This shouldn't matter as we're rendering to a framebuffer.
     const EGLint attribs[] = { EGL_NONE };
+#else
+    const EGLint attribs[] = {
```
kkaefer: Maybe add a comment here that Android actually needs a pbuffer? (Does it?)

test/map/transform.test.cpp
```diff
@@ -109,7 +109,13 @@ TEST(Transform, IntegerZoom) {
     auto checkIntegerZoom = [&transform](uint8_t zoomInt, double zoom) {
         double scale = transform.getState().zoomScale(zoom);
         transform.setScale(scale);
+#if defined(__ANDROID__)
+        // Android uses log(x) / M_LN2 instead of log2(x) because the latter
+        // is broken in ARMv5 - that approach being less precise than log2(x).
+        ASSERT_FLOAT_EQ(transform.getScale(), scale);
+#else
         ASSERT_DOUBLE_EQ(transform.getScale(), scale);
+#endif
```
kkaefer: You could use [`EXPECT_NEAR` instead](https://github.com/google/googletest/blob/master/googletest/docs/AdvancedGuide.md#floating-point-macros) and supply a custom epsilon value.

test/map/transform.test.cpp
```diff
@@ -109,7 +109,13 @@ TEST(Transform, IntegerZoom) {
     auto checkIntegerZoom = [&transform](uint8_t zoomInt, double zoom) {
         double scale = transform.getState().zoomScale(zoom);
         transform.setScale(scale);
+#if defined(__ANDROID__)
```
kkaefer: Can we instead use `#if __ANDROID__`? This define would also trigger in the case of `__ANDROID__ = 0`

test/storage/online_file_source.test.cpp
```diff
@@ -149,7 +149,7 @@ TEST(OnlineFileSource, TEST_REQUIRES_SERVER(RetryDelayOnExpiredTile)) {
     std::unique_ptr<AsyncRequest> req = fs.request(resource, [&](Response res) {
         counter++;
         EXPECT_EQ(nullptr, res.error);
-        EXPECT_GT(util::now(), res.expires);
+        EXPECT_GT(util::now(), *res.expires);
```
kkaefer: Why?

test/map/transform.test.cpp
brunoabinader: Nice catch - will change that.

test/map/transform.test.cpp
brunoabinader: Thanks - using `ASSERT_NEAR` as we usually want the assertions to bail if wrong.

platform/linux/src/headless_backend_egl.cpp
brunoabinader: Yes - `glSurface` is initialized with `EGL_NO_SURFACE`.

platform/linux/src/headless_display_egl.cpp
brunoabinader: From my local test runs, it seems that the EGL implementation used by the Android emulator does not support FBOs, unfortunately.

test/storage/online_file_source.test.cpp
brunoabinader: Otherwise we catch this compilation error using the Android toolchain:
```
../../../mason_packages/android-x86-9/gtest/1.7.0/include/gtest/gtest.h:1585:28: error: invalid operands to binary expression ('const std::__ndk1::chrono::time_point<std::__ndk1::chrono::system_clock, std::__ndk1::chrono::duration<long long, std::__ndk1::ratio<1, 1> > >' and 'const std::experimental::__library_fundamentals_v1::optional<std::__ndk1::chrono::time_point<std::__ndk1::chrono::system_clock, std::__ndk1::chrono::duration<long long, std::__ndk1::ratio<1, 1> > > >')
GTEST_IMPL_CMP_HELPER_(GT, >);
~~~~~~~~~~~~~~~~~~~~~~~~~~~^~
../../../mason_packages/android-x86-9/gtest/1.7.0/include/gtest/gtest.h:1562:12: note: expanded from macro 'GTEST_IMPL_CMP_HELPER_'
  if (val1 op val2) {\
      ~~~~ ^  ~~~~
```
It seems the compiler from Android toolchain is unable to properly handle `std::experimental::optional` dereferencing.

platform/linux/src/headless_backend_egl.cpp
brunoabinader: Fixed - thanks!

ivovandongen:
Very disappointingly, bitrise just [wont run an x86 emulator](https://bitrise-io.github.io/devcenter/faq/android-x86-emulator/)

I've switched back the bitrise configuration to arm and stripped the binaries so things are a little smoother. But there are still some challenges to make this work:

First off, one test fails on the arm emulator:

```
[ RUN      ] Timer.Repeat
/Users/ivo/git/mapbox-gl-native/test/util/timer.test.cpp:59: Failure
Expected: (totalTime) <= (expectedTotalTime * 1.2), actual: 8-byte object <79-02 00-00 00-00 00-00> vs 8-byte object <00-00 00-00 00-C0 82-40>
```

Secondly, besides every test running slower as expected, a couple of tests really slow down more than they should. We could either disable these for now or we should look into why these are so slow.

```
[ RUN      ] SpriteAtlas.JSONLoadingFail
[       OK ] SpriteAtlas.JSONLoadingFail (150943 ms)

[ RUN      ] OfflineDownload.Activate
[       OK ] OfflineDownload.Activate (16268 ms)

[ RUN      ] Source.LoadingFail
[       OK ] Source.LoadingFail (294792 ms)

[ RUN      ] Source.RasterTileFail
[       OK ] Source.RasterTileFail (260694 ms)

[ RUN      ] Source.VectorTileFail
[       OK ] Source.VectorTileFail (129674 ms)

[ RUN      ] GlyphAtlas.LoadingFail
[       OK ] GlyphAtlas.LoadingFail (111375 ms)

[ RUN      ] Memory.Vector
[       OK ] Memory.Vector (15582 ms)

```

cc @brunoabinader @kkaefer

ivovandongen:
For now, we're going to skip CI integration as we've exhausted our options:
- It looks like running an arm emulator on any kind of cloud platform is just too slow to actually complete (now stuck on dex2auth time cap).
- Running an x86 emulator is impossible on the popular ci’s as no cloud platform providers offer the required virtualisation extensions
- Running on Genymotion aborts with no explanation at all
- Firebase / AWS device farm don’t support adb access, only “regular” test frameworks

Another option is to make a runner that is packaged in an APK that uses the [google test listener framework](https://github.com/google/googletest/blob/master/googletest/docs/V1_6_AdvancedGuide.md#using-event-listeners) in combination with one of the supported test frameworks on Firebase / AWS device farm. This will be a follow-up PR so that we can use these tests for development in the meanwhile.

PR #7120, opened 2016-11-18
Upload binary size metrics to S3

kkaefer:
We should upload the binary size metrics to S3 so that we can have a public chart of binary size over time.


kkaefer:
Added a chart at https://mapbox.github.io/mapbox-gl-native/metrics/binary-size/ that should update over time.

Issue #7112, opened 2016-11-17
Better line-breaking support for bidirectional text


ChrisLoer:
Our current line-breaking algorithm has a naive approach to handling text that isn't fully Left-to-Right: if the "base" direction of a label is Right-to-Left instead of LTR, it breaks lines upwards instead of downwards. This is broken two ways:

- If a label starts with RTL text but contains an embedded LTR string with a line break, the two LTR lines will appear in the opposite of the expected order.
- If a label with multiple lines of RTL text also has an icon, the RTL text can overlay the icon.

The fix is to break our application of the bidirectional algorithm and Arabic shaping into two passes. In the first pass, we perform bidirectional layout and shaping on the entire label as one line. After performing shaping, we go through the line to choose line break points. On the second pass, we reapply the bidirectional algorithm with the line break points pre-set in order to get the correct behavior for line breaks of embedded LTR/RTL text.

ChrisLoer:
Closed with https://github.com/mapbox/mapbox-gl-native/pull/7123

PR #7123, opened 2016-11-19
Full support for bidirectional line breaking

ChrisLoer:
Fix for https://github.com/mapbox/mapbox-gl-native/issues/7112

Also makes small changes to handling of leading/trailing whitespace in labels:

- Trim whitespace before calculating line max-width for alignment/justification
- Don't crash if line-breaking algorithm creates a line with just a single whitespace character on it

### DIFF
--- a/package.json
+++ b/package.json
@@ -24,7 +24,7 @@
     "lodash": "^4.16.4",
     "mapbox-gl-shaders": "mapbox/mapbox-gl-shaders#597115a1e1bd982944b068f8accde34eada74fc2",
     "mapbox-gl-style-spec": "mapbox/mapbox-gl-style-spec#7f62a4fc9f21e619824d68abbc4b03cbc1685572",
-    "mapbox-gl-test-suite": "mapbox/mapbox-gl-test-suite#c32d0c5ac80e3b7393bc17b8944e64fa5cffd90a",
+    "mapbox-gl-test-suite": "mapbox/mapbox-gl-test-suite#d85534f77e1b06fbfe7aa610c98a363be86fceb4",
     "mkdirp": "^0.5.1",
     "node-cmake": "^1.2.1",
     "request": "^2.72.0",
--- a/src/mbgl/text/bidi.cpp
+++ b/src/mbgl/text/bidi.cpp
@@ -2,41 +2,30 @@
 
 #include <mbgl/text/bidi.hpp>
 #include <unicode/ubidi.h>
-#include <unicode/ubiditransform.h>
 #include <unicode/ushape.h>
 
 namespace mbgl {
 
-BiDi::BiDi() {
-    UErrorCode errorCode = U_ZERO_ERROR;
-    transform = ubiditransform_open(&errorCode); // Only error is failure to allocate memory, in
-                                                 // that case ubidi_transform would fall back to
-                                                 // creating transform object on the fly
-}
-
-BiDi::~BiDi() {
-    if (transform)
-        ubiditransform_close(transform);
-}
-
-std::u16string BiDi::bidiTransform(const std::u16string& input) {
+// Takes UTF16 input in logical order and applies Arabic shaping to the input while maintaining
+// logical order
+//  Output won't be intelligible until the bidirectional algorithm is applied
+std::u16string applyArabicShaping(const std::u16string& input) {
     UErrorCode errorCode = U_ZERO_ERROR;
 
-    std::unique_ptr<UChar[]> outputText =
-        std::make_unique<UChar[]>(input.size() * 2); // Maximum output of ubidi_transform is twice
-                                                     // the size of input according to
-                                                     // ubidi_transform.h
-    uint32_t outputLength = ubiditransform_transform(
-        transform, input.c_str(), static_cast<int32_t>(input.size()), outputText.get(),
-        static_cast<int32_t>(input.size()) * 2,
-        UBIDI_DEFAULT_LTR,     // Assume input is LTR unless strong RTL characters are found
-        UBIDI_LOGICAL,         // Input is in logical order
-        UBIDI_LTR,             // Output is in "visual LTR" order
-        UBIDI_VISUAL,          //  ''
-        UBIDI_MIRRORING_ON,    // Use mirroring lookups for things like parentheses that need mirroring
-                               // in RTL text
-        U_SHAPE_LETTERS_SHAPE, // Add options here for handling numbers in bidirectional text
-        &errorCode);
+    int32_t outputLength =
+        u_shapeArabic(input.c_str(), static_cast<int32_t>(input.size()), NULL, 0,
+                      (U_SHAPE_LETTERS_SHAPE & U_SHAPE_LETTERS_MASK) |
+                          (U_SHAPE_TEXT_DIRECTION_LOGICAL & U_SHAPE_TEXT_DIRECTION_MASK),
+                      &errorCode);
+
+    // Pre-flighting will always set U_BUFFER_OVERFLOW_ERROR
+    errorCode = U_ZERO_ERROR;
+
+    std::unique_ptr<UChar[]> outputText = std::make_unique<UChar[]>(outputLength);
+    u_shapeArabic(input.c_str(), static_cast<int32_t>(input.size()), outputText.get(), outputLength,
+                  (U_SHAPE_LETTERS_SHAPE & U_SHAPE_LETTERS_MASK) |
+                      (U_SHAPE_TEXT_DIRECTION_LOGICAL & U_SHAPE_TEXT_DIRECTION_MASK),
+                  &errorCode);
 
     // If the algorithm fails for any reason, fall back to non-transformed text
     if (U_FAILURE(errorCode))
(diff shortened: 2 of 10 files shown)


src/mbgl/text/bidi.cpp
```diff
@@ -49,8 +58,86 @@ WritingDirection BiDi::baseWritingDirection(const std::u16string& input) {
...
+    
+    return std::u16string(outputText.get(), outputLength);
+}
+    
+std::u16string BiDi::getLine(int32_t start, int32_t end)
+{
+    UErrorCode errorCode = U_ZERO_ERROR;
+    ubidi_setLine(bidiParagraph, start, end, bidiLine, &errorCode);
+    
+    // TODO: At this point we're committed to the line breaking working. As a fallback we could try to just return the equivalent substring of the original reordered paragraph, but maybe we should just throw an exception?
```
1ec5: What kinds of errors could we expect to occur here? If it’s possible for line breaking to fail due to something in the vector tile data, then we might need a graceful fallback. Otherwise, if this is more about allocation or other catastrophic failures, throwing an exception would make more sense.

src/mbgl/text/glyph_set.cpp
```diff
@@ -88,82 +85,100 @@ void justifyLine(std::vector<PositionedGlyph> &positionedGlyphs, const std::map<
...
+        for (char16_t chr : line)
+        {
+            const SDFGlyph& glyph = sdfs.find(chr)->second;
+            shaping.positionedGlyphs.emplace_back(chr, x, y);
+            x += glyph.metrics.advance + spacing;
+        }
+        
+        // Collapse invisible characters.
+        uint32_t lineEndIndex = shaping.positionedGlyphs.size() - 1;
+        // TODO refactor: Seems like isVisible should be called isInvisible? Also there's an edge case here with one character invisible lines
```
1ec5: Oh, you’re totally right, the logic should be inverted so it actually returns whether the character is visible.

src/mbgl/text/bidi.cpp
ChrisLoer: I believe the assumptions are:

- Input text is UTF-16
- Line breaks are correctly determined (as I look more closely, I think my logic is broken right now for input that already has line breaks in it)
- `applyBidiToParagraph` is called before `applyBidiLineBreaking` (I think I'll try to change the interface to make this a requirement)
- No allocation errors, etc.

In other words, only programmer mistakes should lead to an error. So I'll go with an exception.

src/mbgl/text/glyph_set.cpp
1ec5: Essentially: when the line is centered, we don’t want the trailing newline, space, or zero-width space to shift the centered line off to one side. ([Here’s the GL JS equivalent](https://github.com/mapbox/mapbox-gl-js/blob/c4bb35213b24bb2ad70715bac43db58025fb489c/js/symbol/shaping.js#L132-L136).)

ChrisLoer: I had to torture the map a little to get it to happen, but there is a crash bug here: if you have labels that include trailing line breaks (you can copy and paste them into Studio even though you can't enter them directly), you can end up with "lines" that just have a single line break on them. In that case lineEnd-- wraps around to 2^32.

1ec5: Good catch! Trailing line breaks can come from other sources, such as runtime styling and hand-crafted GeoJSON too.

ChrisLoer:
@1ec5, would you have a chance to review again? I tweaked the line breaking algorithm to bring it into conformance with the current ideographic breaking behavior, but there are still some subtle differences in the handling of whitespace (we can trim more than one character of whitespace from a line, we trim from the front as well as the back, and we calculate alignment after the whitespace trimming instead of before). I also added support for multi-paragaph text runs, although in practice we're unlikely to see those in labels. Most ICU errors now raise exceptions.

I see that one of the bitrise checks is failing, but I don't have a login with bitrise to see what the check is or why it's failing, so I'm not sure how to address that.

I'm also not sure how to merge the tests along with this PR. The tests are passing now pointing against the cloer_icu_multiline branch of mapbox-gl-test-suite, but the master branch contains tests that don't seem to pass on the master branch of mapbox-gl-native yet.

1ec5:
> I see that one of the bitrise checks is failing, but I don't have a login with bitrise to see what the check is or why it's failing, so I'm not sure how to address that.

I’ve added you to all the gl-native apps on Bitrise.

> The tests are passing now pointing against the cloer_icu_multiline branch of mapbox-gl-test-suite, but the master branch contains tests that don't seem to pass on the master branch of mapbox-gl-native yet.

If test-suite has gained new tests since gl-native’s master branch last pinned to it, you can mark those tests as ignored with links to issues in this repository that track those bugs or features, then commit those changes directly to test-suite’s master branch. Then you can rebase test-suite’s cloer_icu_multiline branch atop that change and point this gl-native branch’s package.json to that commit.

When you’re ready to land this PR, rebase test-suite’s cloer_icu_multiline branch again and update the version pin here. Quickly merge your test-suite branch to master before any intervening changes make their way in, then merge this PR.

src/mbgl/text/glyph_set.cpp
```diff
@@ -87,83 +86,113 @@ void justifyLine(std::vector<PositionedGlyph> &positionedGlyphs, const std::map<
...
 }
+  
+float GlyphSet::determineIdeographicLineWidth(const std::u16string& logicalInput, const float spacing, float maxWidth) const
+{
+    float totalWidth = 0;
+    
+    // totalWidth doesn't include the last character for magical tuning reasons. This makes the algorithm a little
+    //  more agressive about trying to fit the text into fewer lines, taking advantage of the tolerance for going a little
+    //  over maxWidth
+    for (uint32_t i = 0; i < logicalInput.size() - 1; i++) {
```
1ec5: Can this be an iterator-style `for` loop?

src/mbgl/text/glyph_set.cpp
```diff
@@ -87,83 +86,113 @@ void justifyLine(std::vector<PositionedGlyph> &positionedGlyphs, const std::map<
...
+    
+    float x = 0;
+    float y = yOffset;
+    
+    float maxLineLength = 0;
+    
+    for (std::u16string line : lines)
+    {
+        // Collapse whitespace so it doesn't throw off justification
+        boost::algorithm::trim_if(line, boost::algorithm::is_any_of(u" \t\n\v\f\r"));
```
1ec5: This doesn’t include the ideographic space (U+3000) which is used in CJK, or some of the [more exotic Western spaces](https://en.wikipedia.org/wiki/Whitespace_character#Unicode) such as the thin space and mutton. Since we’re bringing in ICU, can we take advantage of any character class functionality in there? Or are we concerned about it being too slow?

ChrisLoer: I experimented with that idea, but it pulled in a significant chunk of new code with the ICU `UnicodeString` class (which isn't used by the BiDi code), and also required an extra string copy just to do the trimming. Both of those could be non-issues performance-wise, but this didn't feel like an important enough use case to pull in the dependency. If our STL specialized `ctype<char16_t>`, `boost::algorithm::trim` would work without us having to define whitespace, but alas it doesn't seem to. Maybe for now I'll just add U+3000 to our whitespace list? I can also move the list back into 18n.cpp and just pass the predicate to `trim_if`.

src/mbgl/text/glyph_set.cpp
ChrisLoer: Could be, but then we have to add extra logic to remove the last character, and I don't think the code gets much simpler. Or am I missing something?

1ec5: Ah, I missed that. 👍

src/mbgl/text/bidi.cpp
```diff
@@ -2,55 +2,123 @@
...
+        return input;
+
+    return std::u16string(outputText.get(), outputLength);
+}
+
+ProcessedBiDiText::ProcessedBiDiText(BiDi& p_bidi) : bidi(p_bidi) {
+}
+
+void ProcessedBiDiText::mergeParagraphLineBreaks(std::set<int32_t>& lineBreakPoints) {
+    for (int32_t i = 0; i < ubidi_countParagraphs(bidi.bidiText); i++) {
```
jfirebaugh: Can/should `ubidi_countParagraphs(bidi.bidiText)` be extracted from the loop condition, so it's executed only once?

src/mbgl/text/bidi.cpp
```diff
@@ -2,55 +2,123 @@
...
-    // the BiDi algorithm
-    return ubidi_getBaseDirection(input.c_str(), static_cast<int32_t>(input.size())) == UBIDI_RTL
-        ? WritingDirection::RightToLeft
-        : WritingDirection::LeftToRight;
+std::u16string BiDi::getLine(int32_t start, int32_t end) {
+    UErrorCode errorCode = U_ZERO_ERROR;
+    ubidi_setLine(bidiText, start, end, bidiLine, &errorCode);
+
+    if (U_FAILURE(errorCode))
+        throw std::runtime_error("msg");
```
jfirebaugh: `"msg"` looks like a placeholder string.

kkaefer:
argh sorry; I had a few review comments, but apparently I never clicked on on submit.

src/mbgl/util/i18n.hpp
```diff
@@ -6,9 +6,6 @@ namespace mbgl {
 namespace util {
 namespace i18n {
 
-/** Returns whether a character is a visible character. */
-bool isVisible(uint16_t chr);
```
ChrisLoer: This was only used to remove trailing characters from a line, and only newline/space/zero-width space. I replaced it with trim because I think that's the semantics we're going for -- we don't want whitespace trailing/leading whitespace to affect layout. Zero-width space will get "rendered" now, although it won't affect layout unless the font we're using assigns it non-zero width.

"Invisible" sounds to me like a font property (if there's no glyph, the code point is effectively invisible, also the font may have a glyph that doesn't draw any pixels). As you note below Unicode also has lots of extra "whitespace" characters. This change doesn't affect our handling of them.

src/mbgl/text/glyph_set.cpp
```diff
@@ -88,81 +99,122 @@ void justifyLine(std::vector<PositionedGlyph> &positionedGlyphs, const std::map<
...
+float GlyphSet::determineIdeographicLineWidth(const std::u16string& logicalInput,
+                                              const float spacing,
+                                              float maxWidth) const {
+    float totalWidth = 0;
+
+    // totalWidth doesn't include the last character for magical tuning reasons. This makes the
+    // algorithm a little
+    //  more agressive about trying to fit the text into fewer lines, taking advantage of the
+    //  tolerance for going a little
+    //  over maxWidth
```
ChrisLoer: Ack, I should have noticed this after running clang-format.

src/mbgl/text/glyph_set.cpp
```diff
@@ -88,81 +99,122 @@ void justifyLine(std::vector<PositionedGlyph> &positionedGlyphs, const std::map<
...
+
+    float x = 0;
+    float y = yOffset;
+
+    float maxLineLength = 0;
 
-    const uint32_t height = (line + 1) * lineHeight;
+    for (std::u16string line : lines) {
+        // Collapse whitespace so it doesn't throw off justification
+        boost::algorithm::trim_if(line, boost::algorithm::is_any_of(u" \t\n\v\f\r"));
```
ChrisLoer: They won't be trimmed, so they may affect layout slightly. Whether they allow word breaking is controlled by `i18n::allowsWordBreaking`. It looks like we don't currently break on thin/em/en spaces, either.

src/mbgl/text/bidi.cpp
```diff
@@ -2,55 +2,123 @@
...
+        transformedLines.push_back(bidi.getLine(start, lineBreakPoint));
+        start = lineBreakPoint;
+    }
+
+    return transformedLines;
+}
+
+BiDi::BiDi() {
+    bidiText = ubidi_open();
+    bidiLine = ubidi_open();
```
ChrisLoer: It's just the way the ICU interface works. `ubidi_setLine` wants two `UBiDi*` objects: one "input" representing the whole run of text, and one "output" representing bidirectional information for a single line.

src/mbgl/text/bidi.cpp
```diff
@@ -2,55 +2,123 @@
...
+    ubidi_setLine(bidiText, start, end, bidiLine, &errorCode);
+
+    if (U_FAILURE(errorCode))
+        throw std::runtime_error("msg");
+
+    // Because we set UBIDI_REMOVE_BIDI_CONTROLS, the output may be smaller than what we reserve
+    //  Setting UBIDI_INSERT_LRM_FOR_NUMERIC would require
+    //  ubidi_getLength(pBiDi)+2*ubidi_countRuns(pBiDi)
+    int32_t outputLength = ubidi_getProcessedLength(bidiLine);
+    std::unique_ptr<UChar[]> outputText = std::make_unique<UChar[]>(outputLength);
```
ChrisLoer: Oh, right! I left C++ world at C++98, still getting used to all the great additions...

src/mbgl/text/bidi.cpp
ChrisLoer: OK, is that our standard style? I'll start doing that.

ChrisLoer:
@kkaefer Porting to gl-js is exposing some further changes I'll want to make (like https://github.com/mapbox/mapbox-gl-native/issues/7253). When I make a PR for those changes I'll incorporate your suggestions from here. I don't see anything you flagged that's broken and needs immediate fixing -- let me know if you think otherwise.

kkaefer:
Yeah, just minor nitpicks and suggestions.

package.json
```diff
@@ -24,7 +24,7 @@
     "lodash": "^4.16.4",
     "mapbox-gl-shaders": "mapbox/mapbox-gl-shaders#597115a1e1bd982944b068f8accde34eada74fc2",
     "mapbox-gl-style-spec": "mapbox/mapbox-gl-style-spec#7f62a4fc9f21e619824d68abbc4b03cbc1685572",
-    "mapbox-gl-test-suite": "mapbox/mapbox-gl-test-suite#c32d0c5ac80e3b7393bc17b8944e64fa5cffd90a",
+    "mapbox-gl-test-suite": "mapbox/mapbox-gl-test-suite#d85534f77e1b06fbfe7aa610c98a363be86fceb4",
```
1ec5: mapbox/mapbox-gl-test-suite#186 was rebased onto that repository’s master branch, so mapbox/mapbox-gl-test-suite@d85534f77e1b06fbfe7aa610c98a363be86fceb4 has been orphaned. This line should have been changed to point to mapbox/mapbox-gl-test-suite@0c6f3e00193248474349cd730e232eb5ffcd9bf4 before this PR was merged.

ChrisLoer: Oops, sorry about that, and thanks for fixing!

PR #7125, opened 2016-11-20
Short-circuit redundant camera changes

### LINKED ISSUE #5983: Avoid animating to the same camera options

1ec5:
Avoid canceling transitions (and triggering preexisting completion handlers) until we get a chance to ensure that a new transition really does have to begin. Consistently avoid mbgl transitions for redundant camera changes. Upon bailing, schedule the completion handler to run asynchronously on a delay equivalent to the requested animation duration.

~~MGLMapCamera now performs a float-precision equality check to avoid trivial differences. Added an equality operator to CameraOptions that similarly uses float-precision.~~ Added a “functional” equality method to MGLMapCamera that avoids trivial differences.

Having made these changes, I very much want to move these redundancy checks into `mbgl::Transform`. However, that depends on eliminating the numerous SDK-side `cancelTransitions()` calls: https://github.com/mapbox/mapbox-gl-native/issues/5833#issuecomment-241533404. This PR is less risky but should address the performance issues the iOS SDK is experiencing.

This change also fixes invocations of `XCTAssertEqualWithAccuracy()` that incorrectly expressed the accuracy as a number of digits rather than a scalar tolerance.

Fixes #5983.

/cc @incanus @jfirebaugh

### DIFF
--- a/platform/ios/src/MGLMapView.mm
+++ b/platform/ios/src/MGLMapView.mm
@@ -2269,8 +2269,6 @@ public:
 
 - (void)_setCenterCoordinate:(CLLocationCoordinate2D)centerCoordinate edgePadding:(UIEdgeInsets)insets zoomLevel:(double)zoomLevel direction:(CLLocationDirection)direction duration:(NSTimeInterval)duration animationTimingFunction:(nullable CAMediaTimingFunction *)function completionHandler:(nullable void (^)(void))completion
 {
-    _mbglMap->cancelTransitions();
-
     mbgl::CameraOptions cameraOptions;
     cameraOptions.center = MGLLatLngFromLocationCoordinate2D(centerCoordinate);
     cameraOptions.padding = MGLEdgeInsetsFromNSEdgeInsets(insets);
@@ -2297,6 +2295,20 @@ public:
             });
         };
     }
+    
+    MGLMapCamera *camera = [self cameraForCameraOptions:cameraOptions];
+    if ([self.camera isEqualToMapCamera:camera])
+    {
+        if (completion)
+        {
+            [self animateWithDelay:duration animations:^{
+                completion();
+            }];
+        }
+        return;
+    }
+    
+    _mbglMap->cancelTransitions();
     _mbglMap->easeTo(cameraOptions, animationOptions);
 }
 
@@ -2415,9 +2427,6 @@ public:
 
 - (void)_setVisibleCoordinates:(const CLLocationCoordinate2D *)coordinates count:(NSUInteger)count edgePadding:(UIEdgeInsets)insets direction:(CLLocationDirection)direction duration:(NSTimeInterval)duration animationTimingFunction:(nullable CAMediaTimingFunction *)function completionHandler:(nullable void (^)(void))completion
 {
-    _mbglMap->cancelTransitions();
-
-    [self willChangeValueForKey:@"visibleCoordinateBounds"];
     mbgl::EdgeInsets padding = MGLEdgeInsetsFromNSEdgeInsets(insets);
     padding += MGLEdgeInsetsFromNSEdgeInsets(self.contentInset);
     std::vector<mbgl::LatLng> latLngs;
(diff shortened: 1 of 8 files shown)


platform/ios/uitest/MapViewTests.m
```diff
@@ -305,6 +305,23 @@ - (void)testZoomSet {
...
+- (void)testCameraDebouncing {
+    MGLMapCamera *camera = [MGLMapCamera cameraLookingAtCenterCoordinate:CLLocationCoordinate2DMake(45, -122)
+                                                            fromDistance:100
+                                                                   pitch:30
+                                                                 heading:45];
+    tester.mapView.camera = camera;
+    XCTAssertEqualObjects(tester.mapView.camera, camera);
+    
+    [tester.mapView setCamera:camera withDuration:10 animationTimingFunction:nil completionHandler:^{
+        XCTAssert(NO, @"Camera animation should not be canceled by redundantly setting the camera to the current camera.");
```
1ec5: This is the best I could come up with for a test of the redundancy check. It essentially only verifies that `cancelTransitions()` isn’t called in the case of a redundant camera change. The test in #6060 is infeasible because the completion handler is eventually called even for redundant camera changes.

I verified via breakpoints that we do short-circuit mbgl for redundant camera changes and that significant camera changes still work.

include/mbgl/map/camera.hpp
```diff
@@ -34,6 +34,31 @@ struct CameraOptions {
     /** Pitch toward the horizon measured in radians, with 0 rad resulting in a
         two-dimensional map. */
     optional<double> pitch;
+
+    bool operator==(const CameraOptions& o) const {
+        if ((!center && o.center) || (center && !o.center) ||
```
jfirebaugh: Why isn't this defined simply as memberwise `==`?

1ec5: Due to the float casts, which are needed due to slight imprecision – comparing doubles never returns true in practice. Or should a fuzzy equals be a separate method for clarity?

1ec5: I should probably bring over the six-decimal-place rounding from #6060.

1ec5:
I’ve pared back the change to touch only SDK code and also leave the normal `-[MGLMapCamera isEqual:]` and `-[MGLMapCamera hash]` methods unaffected.

1ec5:
We’ll need to cherry-pick #8073 in order to get the macOS CI build to pass again.

platform/ios/src/MGLMapView.mm
```diff
@@ -2415,9 +2427,7 @@ - (void)setVisibleCoordinates:(const CLLocationCoordinate2D *)coordinates count:
 
 - (void)_setVisibleCoordinates:(const CLLocationCoordinate2D *)coordinates count:(NSUInteger)count edgePadding:(UIEdgeInsets)insets direction:(CLLocationDirection)direction duration:(NSTimeInterval)duration animationTimingFunction:(nullable CAMediaTimingFunction *)function completionHandler:(nullable void (^)(void))completion
 {
-    _mbglMap->cancelTransitions();
```
frederoni: Nit: Inconsistent line break.

1ec5: Good catch – fixed.

PR #7128, opened 2016-11-21
[ios, macos] rename style spec properties

### LINKED ISSUE #6098: Rename some layer and source properties

frederoni:
Fixes #6098 

Started with the most obvious properties to avoid confusion and sloppiness.

- `icon-image` => `icon-image-name`
- `raster-brightness-min` => `raster-brightness-minimum`
- `raster-brightness-max` => `raster-brightness-maximum`

line-dasharray is another one that doesn't feel particularly idiomatic but I'm ok with it and the remaining circlePitchScale & rasterHueRotate that @1ec5 pointed out https://github.com/mapbox/mapbox-gl-native/issues/6098#issue-172229079.

@1ec5 @boundsj 👀


platform/darwin/scripts/style-spec-cocoa-conventions-v8.json
```diff
@@ -0,0 +1,9 @@
+{
+  "layout_symbol": {
+    "icon-image": "icon-image-name"
+  },
+  "paint_raster": {
+    "raster-brightness-min": "raster-brightness-minimum",
```
1ec5: minimumRasterBrightness reads better. We might as well make that change as long as we're renaming this property.

platform/darwin/scripts/generate-style-code.js
```diff
@@ -4,11 +4,26 @@ const fs = require('fs');
...
+const cocoaConventions = require('./style-spec-cocoa-conventions-v8.json');
+let spec = _.merge(require('mapbox-gl-style-spec').latest, require('./style-spec-overrides-v8.json'));
 const prefix = 'MGL';
 const suffix = 'StyleLayer';
 
+function renameProperties(obj, overrides, stack='') {
+    _.forOwn(overrides, function (value, key) {
+        const keyPath = stack + '.';
+        if (_.isObject(value)) {
+            renameProperties(obj, overrides[key], stack.length ? keyPath + key : key);
```
1ec5: Does this function need to be recursive? Seems like we could assume a two-level-deep override object and use subscript notation instead of _.set and _.unset.

platform/darwin/scripts/generate-style-code.js
```diff
@@ -4,11 +4,20 @@ const fs = require('fs');
...
+const cocoaConventions = require('./style-spec-cocoa-conventions-v8.json');
+let spec = _.merge(require('mapbox-gl-style-spec').latest, require('./style-spec-overrides-v8.json'));
 const prefix = 'MGL';
 const suffix = 'StyleLayer';
 
+// Rename properties and keep _original for use with setters and getters
+_.forOwn(cocoaConventions, function (properties, kind) {
+    _.forOwn(properties, function (newName, oldName) {
+        spec[kind][newName] = spec[kind][oldName];
+        spec[kind][newName]["_original"] = oldName;
```
1ec5: Nit: normally, in JavaScript, you’d say `.original` instead of `["_original"]`.

platform/darwin/scripts/generate-style-code.js
```diff
@@ -212,6 +221,10 @@ global.propertyDefault = function (property, layerType) {
     return 'an `MGLStyleValue` object containing ' + describeValue(property.default, property, layerType);
 };
 
+global.originalPropertyName = function (property) {
+    return property._original ? property._original : property.name;
```
1ec5: Nit: this can be simplified as `property._original || property.name`.

platform/darwin/src/MGLStyleLayer.mm.ejs
```diff
@@ -21,9 +21,9 @@ namespace mbgl {
 <% if (layoutProperties.length) { -%>
 <% for (const property of layoutProperties) { -%>
```
1ec5: We’ve been relying on an implementation detail of v8 whereby an object’s properties are enumerated in the order that they’re added to the object. Unfortunately, all the renamed properties appear last in the generated source code, which is a little jarring.

To fix this issue, we’d need to sort `layoutProperties`’ keys before enumerating the object:

```js
let sortedLayoutProperties = _.pick(layoutProperties, _.keysIn(layoutProperties).sort());
```

There are probably more efficient ways of doing this, but this way involves less typing. 😉

frederoni: After the style spec has been processed, the key is removed and inserted into the object as name so I sorted with `_.sortBy(layoutProperties, ['name']`. Unfortunately, it wasn't very well sorted before so the diff increased quite a lot. Divided into separate commits to make it easier to review.

platform/darwin/scripts/generate-style-code.js
1ec5: Ack, sorry, I meant `._original`. You were right about trying to avoid a theoretical conflict with a property named `original`.

platform/darwin/src/MGLStyleLayer.mm.ejs
1ec5: I mistakenly thought the properties were alphabetically sorted to begin with, but you’re right that the properties are far from sorted in the style specification’s JSON objects. The [official documentation](https://www.mapbox.com/mapbox-gl-style-spec/) is generated using v8, just like our generated runtime styling files, so for example `icon-rotate` comes before `icon-padding`. @lucaswoj doesn’t know of any reason we need to keep the properties in that order, so alphabetizing the properties makes sense because it’ll make it easier to find them in our jazzy documentation.

PR #7133, opened 2016-11-21
[android] - allow deselection of out of viewport MarkerView

tobrun:
This PR allows to deselect markers that aren't found in the current viewport.

Review @zugaldia.


zugaldia:
@tobrun 👍

Issue #7029, opened 2016-11-11
Investigate dSYM size decrease between v3.3.6 and v3.4.0


1ec5:
Building the iOS SDK with Xcode 8 would decrease the framework size significantly. Unfortunately, #6842 is keeping us on Xcode 7.3.

Since it’s only the dynamic framework that’s giving Xcode 7.3 users problems, I’m wondering whether we could ship an _unstripped_ dynamic framework and add code in strip-frameworks.sh to strip that framework (in addition to stripping out irrelevant architecture slices as it does now). As it is, you can’t submit to the App Store without having strip-frameworks.sh in a Run Script build phase, so this should cover anyone who needs to use the dynamic framework in a simulator that comes with Xcode 7.3.

Putting on the 3.4.0 milestone for consideration for a beta.

/cc @boundsj

boundsj:
Here are some notes I took while investigating this a bit more yesterday

- The 3.3.x series branch, when built with Xcode 7.3.1 produces a 50 MB symbol stripped dynamic framework and a 190 MB dSYM
- The 3.3.x series branch when built with Xcode 8.0 produces a 38 MB symbol stripped dynamic framework and a 190 MB dSYM
- The latest 3.4.0 beta, built with Xcode 7.3.1 has a 49 MB symbol stripped dynamic framework and a 34.5 MB dSYM
- The latest 3.4.0 beta, built with Xcode 8.0 has a 37.5 MB symbol stripped dynamic framework and a 36.7 MB dSYM

I think we should use Xcode 8.0+ for future builds. Although we are left with the issue for Xcode 7.3.1 users.

> @1ec5 I’m wondering whether we could ship an unstripped dynamic framework and add code in strip-frameworks.sh to strip that framework

I'm not sure if I understand how to pull this off. We currently ship an unstripped  (i.e. does not have `strip -Sx` run on it) `mapbox-ios-sdk-#..3#-symbols-dynamic.zip` output that works fine in an Xcode 7.3.1 swift environment. But it would fail if we suggested stripping it in a build phase, I think. But I'm sure I'm missing something.

I'd also like to investigate the large **decrease** in dSYM size when Xcode 8.0 is used to make the build. We've gone from ~100 MB to close to 200 MB (in more recent 3.3.x version built with Xcode 7.3.1) to ~37 MB in the 3.4.0 branch using Xcode 8.0

1ec5:
[...]

Issue #6194, opened 2016-08-29
-[MGLGeoJSONSource mbgl_source] mapbox::geojson symbol(s) not found


friedbunny:
**Platform:** iOS
**Mapbox SDK version:** v3.4.0-alpha.2 (static)

Attempting to use the static framework with v3.4.0-alpha.2 results in several undefined symbols errors.

Both `arm64` and `x86_64` architectures are reported as missing.

```
Undefined symbols for architecture arm64:
  "mapbox::geojson::convert(rapidjson::GenericValue<rapidjson::UTF8<char>, rapidjson::CrtAllocator> const&)", referenced from:
      mbgl::style::conversion::Result<mapbox::util::variant<mapbox::geometry::geometry<double>, mapbox::geometry::feature<double>, mapbox::geometry::feature_collection<double, std::__1::vector> > > mbgl::style::conversion::convertGeoJSON<rapidjson::GenericValue<rapidjson::UTF8<char>, rapidjson::CrtAllocator> >(rapidjson::GenericValue<rapidjson::UTF8<char>, rapidjson::CrtAllocator> const&) in Mapbox(geojson_source_impl.o)
  "mapbox::geojson::parse(std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char> > const&)", referenced from:
      -[MGLGeoJSONSource mbgl_source] in Mapbox(MGLGeoJSONSource.o)
ld: symbol(s) not found for architecture arm64
```

/cc @1ec5 @boundsj @kkaefer @mourner

1ec5:
Perhaps something changed during the CMake migration that keeps us from linking the static library to geojson-vt-cpp.

friedbunny:
This appears to no longer be a problem — the static framework for v3.4.0-alpha.5 works as expected, as does one freshly compiled from master/alpha.6. Weird.

PR #7131, opened 2016-11-21
Update compile flags for iOS

kkaefer:
We should do these changes to iOS compilation settings:

- [x] [Only build active architecture when building Debug builds to speed up compile time](https://github.com/mapbox/mapbox-gl-native/pull/7388)
- [x] Remove bitcode from Debug builds
- [x] [Add `-fvisibility=hidden`](https://github.com/mapbox/mapbox-gl-native/pull/7604)
- [x] Strip non-global symbols for Release static/dynamic library
- [x] Enable monolithic LTO

/cc @1ec5


kkaefer:
> This means we won't catch any differences between 32-bit and 64-bit. We have unit tests for regressions around the size of an NSInteger, for example.

Okay; can we force `ONLY_ACTIVE_ARCH=NO` on CI, but use `YES` (first commit in this PR) for regular builds that you do when developing?

kkaefer:
> We already conditionally do this in the packaging script (just not in the Xcode project)

Any reason why we wouldn't do this as part of the Xcode build and only do it in an external packaging script?

kkaefer:
With all of the above optimizations, arm64 shrinks from 3,351 KB to 2,925 KB.

1ec5:
I’m all for performing the stripping in the Xcode project instead of the packaging script. The more of the packaging script we can put in proper Xcode build steps, the better.

>> This means we won't catch any differences between 32-bit and 64-bit. We have unit tests for regressions around the size of an NSInteger, for example.

> Okay; can we force ONLY_ACTIVE_ARCH=NO on CI, but use YES (first commit in this PR) for regular builds that you do when developing?

I’d prefer not to trim the CI build matrix at the expense of local incremental build times. We already make this compromise when it comes to static builds, which is why #6194 went undetected for awhile.

kkaefer:
> I’d prefer not to trim the CI build matrix at the expense of local incremental build times.

It's the other way around: Local build times would be faster, since we are only building the active architecture for Debug builds (typically arm64 with most modern devices). On CI, we'd explicitly set `ONLY_ACTIVE_ARCH=NO` on the Debug build to have it build both architectures. Release builds will still build both architectures local and on CI.

friedbunny:
> Any reason why we wouldn't do this as part of the Xcode build and only do it in an external packaging script?

We can do it in the Xcode project, but the implementation of conditionally stripped symbols will have to be updated in the packaging script and all of the possible packages will need to be re-validated.

kkaefer:
> all of the possible packages will need to be re-validated.
 
What does that mean? Are you referring to code signing?

friedbunny:
I mean that we’ll need to confirm that every built package still works as expected.

kkaefer:
@friedbunny what's the process here? Do we have a list of all the packages that need to build, and how to validate them?

friedbunny:
[Here’s the list of packages we publish](https://github.com/mapbox/mapbox-gl-native/blob/4577c56745d90cfa00837d00e5e5605260a54879/platform/ios/scripts/deploy-packages.sh#L111-L115) — static and dynamic w/symbols, static and dynamic w/o symbols, and one Fabric build that’s static w/o symbols.

The process for verifying that a build is usable is mostly manual. You’ll need at least two test iOS projects that integrate Mapbox.framework — one for static and one for dynamic.

Steps to confirm that a build works:
- Test it in an iOS project.
  - With and without storyboards.
  - ObjC and Swift.
- Submit that app to TestFlight/App Store.
- Confirm that that dSYM and framework UUIDs match. ([automatic](https://github.com/mapbox/mapbox-gl-native/blob/4577c56745d90cfa00837d00e5e5605260a54879/platform/ios/scripts/package.sh#L194-L203))

These steps should be repeated for every packaged build.

I’m also unsure where we stand on Xcode 8 vs 7.3.1 (https://github.com/mapbox/mapbox-gl-native/issues/7029) — it may be necessary to test builds from both. /cc @boundsj

kkaefer:
> The process for verifying that a build is usable is mostly manual.

Is there a way to automate it? It seems that I can't run `deploy-packages.sh` when that version has already been published.

What is the advantage of building with symbols? I thought we have external symbols in the .dSYM file anyway?

> You’ll need at least two test iOS projects that integrate Mapbox.framework — one for static and one for dynamic.

Where do I get those from? Just create a new sample app that embeds those frameworks? Can we automate this?

friedbunny:
I looked at implementing LTO today and wasn’t able to see any noticeable change in the bench app FPS on my iPhone 5 — plain master and various LTO combinations averaged ~40.5 FPS. After watching the WWDC session on it, this was underwhelming.

### Notes
- ThinLTO seems like the way to go (rather than Monolithic) — compile times are supposed to be much better, while performance is supposed to be more or less the same.
- It seemed compatible with ccache (surprisingly).
- Problematic: missing core symbol warnings from dsymutil when building the dynamic framework.

### Links
- [ThinLTO: Scalable and Incremental LTO - LLVM Blog](http://blog.llvm.org/2016/06/thinlto-scalable-and-incremental-lto.html)
- [ThinLTO - Clang docs](https://clang.llvm.org/docs/ThinLTO.html)
- [WWDC 2016 - What’s New in LLVM](https://developer.apple.com/videos/play/wwdc2016/405/) ([transcript](http://asciiwwdc.com/2016/sessions/405))

kkaefer:
LTO optimizes the CPU code, but it looks like rendering is GPU bound on your phone. This would explain the lack of FPS increase. Did LTO have any effect on binary size?

friedbunny:
#12502 takes on LTO, the final unmerged piece of this PR — closing.

PR #7139, opened 2016-11-22
Call correct onfinish callback

tobrun:
Targets release branch, some follow up work on the cancelable callback system.

While doing some testing I noticed that we weren't correctly calling the onFinish callback. This could led to situations that we recevied callbacks for the previously set callback while the current one should be invoked. 

Review @zugaldia


zugaldia:
@tobrun 👍

Issue #6986, opened 2016-11-09
[MGLGeoJSONSource initWithIdentifier:geoJSONData] crashes if data isn't a feature collection


jfirebaugh:
The following code crashes with the unhandled exception `mapbox::util::bad_variant_access: in get<T>()` on [this line](https://github.com/mapbox/mapbox-gl-native/blob/master/platform/darwin/src/MGLGeoJSONSource.mm#L77).

```
    MGLGeoJSONSource *geojsonSource = [[MGLGeoJSONSource alloc]
        initWithIdentifier:@"geojson"
               geoJSONData:[@"{\"type\": \"Point\", \"coordinates\": [0, 0]}" dataUsingEncoding:NSUTF8StringEncoding]
                   options:nil];
```

nitrag:
Same error `mapbox::util::bad_variant_access: in get<T>()` so I figured I'd comment here.

I'm also getting this bug when adding MGLCircleStyleLayer after a MGLGeoJSONSource without any markers (it is only a single line).

```
let jsonPath = Bundle.main.url(forResource: "example", withExtension: "geojson")!
let geoJsonSource = MGLGeoJSONSource(identifier: "testasdf", url: jsonPath, options: options)
self.mapView.style().add(geoJsonSource)
NSLog("Source has been created and added")

//comment the rest of this out and it will work
let circleLayer = MGLCircleStyleLayer(identifier: "circle-geojson", source: geoJsonSource)
circleLayer.circleColor = MGLStyleConstantValue(rawValue: UIColor.orange)
circleLayer.circleOpacity = MGLStyleConstantValue(rawValue: 0.8)
circleLayer.circleRadius = MGLStyleConstantValue(rawValue: 13)
self.mapView.style().add(circleLayer);
```

nitrag:
I take that back...

It seems to crash when I reuse `options`.  If I set:
`let geoJsonSource1 = MGLGeoJSONSource(identifier: "asdfasadfas", url: path, options: [String : Any]())`

it works fine. However as like below, it will crash

```
func addGeoJSONSource(){
    
    var options = [String : Any]()
    options[MGLGeoJSONClusterOption] = true
    options[MGLGeoJSONClusterRadiusOption] = 42
    options[MGLGeoJSONClusterMaximumZoomLevelOption] = 14
    options[MGLGeoJSONMaximumZoomLevelOption] = 20
    options[MGLGeoJSONToleranceOption] = 1
    
[...]

PR #7145, opened 2016-11-22
Allow initializing MGLGeoJSONSource from a shape

frederoni:
First stab at #6986 

I'm slightly hesitant about replacing `-[MGLGeoJSONSource initWithIdentifier:features:options:]` with `feature` because of the additional initialization of MGLShapeCollectionFeature devs has to do now compared to a concise array literal prior to this change.

@1ec5 👀

### DIFF
--- a/platform/darwin/src/MGLGeoJSONSource.h
+++ b/platform/darwin/src/MGLGeoJSONSource.h
@@ -1,6 +1,7 @@
 #import "MGLSource.h"
 
 #import "MGLTypes.h"
+#import "MGLShape.h"
 
 NS_ASSUME_NONNULL_BEGIN
 
@@ -93,28 +94,29 @@ extern const MGLGeoJSONSourceOption MGLGeoJSONSourceOptionSimplificationToleranc
 - (instancetype)initWithIdentifier:(NSString *)identifier URL:(NSURL *)url options:(nullable NS_DICTIONARY_OF(MGLGeoJSONSourceOption, id) *)options NS_DESIGNATED_INITIALIZER;
 
 /**
- Returns a GeoJSON source with an identifier, features dictionary, and dictionary 
+ Returns a GeoJSON source with an identifier, features dictionary, and dictionary
  of options for the source according to the
  <a href="https://www.mapbox.com/mapbox-gl-style-spec/#sources-geojson">style
  specification</a>.
  
  @param identifier A string that uniquely identifies the source.
- @param features An array of features that conform to the `MGLFeature` protocol.
+ @param shape A concrete subclass of `MGLShape`
  @param options An `NSDictionary` of options for this source.
  @return An initialized GeoJSON source.
  */
-- (instancetype)initWithIdentifier:(NSString *)identifier features:(NSArray<id<MGLFeature>> *)features options:(nullable NS_DICTIONARY_OF(MGLGeoJSONSourceOption, id) *)options NS_DESIGNATED_INITIALIZER;
+- (instancetype)initWithIdentifier:(NSString *)identifier shape:(nullable MGLShape *)shape options:(nullable NS_DICTIONARY_OF(MGLGeoJSONSourceOption, id) *)options NS_DESIGNATED_INITIALIZER;
 
 #pragma mark Accessing a Source’s Content
 
 /**
- The contents of the source.
+ The contents of the source. A shape can represent a GeoJSON geometry, a feature, 
+ or a collection of features.
 
  If the receiver was initialized using `-initWithIdentifier:URL:options:`, this property
  is set to `nil`. This property is unavailable until the receiver is passed into
  `-[MGLStyle addSource]`.
  */
-@property (nonatomic, nullable) NS_ARRAY_OF(id <MGLFeature>) *features;
+@property (nonatomic, nullable) MGLShape *shape;
 
 /**
  A GeoJSON representation of the contents of the source.
@@ -137,7 +139,6 @@ extern const MGLGeoJSONSourceOption MGLGeoJSONSourceOptionSimplificationToleranc
  property is set to `nil`.
  */
 @property (nonatomic, nullable) NSURL *URL;
-
 
 @end
 
--- a/platform/darwin/src/MGLGeoJSONSource.mm
+++ b/platform/darwin/src/MGLGeoJSONSource.mm
@@ -3,6 +3,7 @@
 #import "MGLMapView_Private.h"
 #import "MGLSource_Private.h"
 #import "MGLFeature_Private.h"
+#import "MGLShape_Private.h"
 
 #import "NSURL+MGLAdditions.h"
 
@@ -49,13 +50,13 @@ const MGLGeoJSONSourceOption MGLGeoJSONSourceOptionSimplificationTolerance = @"M
     return self;
 }
 
-- (instancetype)initWithIdentifier:(NSString *)identifier features:(NSArray<id<MGLFeature>> *)features options:(NS_DICTIONARY_OF(NSString *,id) *)options {
+- (instancetype)initWithIdentifier:(NSString *)identifier shape:(nullable MGLShape *)shape options:(NSDictionary<MGLGeoJSONSourceOption,id> *)options
+{
     if (self = [super initWithIdentifier:identifier]) {
-        _features = features;
+        _shape = shape;
         _options = options;
         [self commonInit];
     }
-    
     return self;
 }
 
@@ -85,21 +86,15 @@ const MGLGeoJSONSourceOption MGLGeoJSONSourceOptionSimplificationTolerance = @"M
     if (self.URL) {
         NSURL *url = self.URL.mgl_URLByStandardizingScheme;
         source->setURL(url.absoluteString.UTF8String);
-        _features = nil;
+        _shape = nil;
     } else if (self.geoJSONData) {
         NSString *string = [[NSString alloc] initWithData:self.geoJSONData encoding:NSUTF8StringEncoding];
-        const auto geojson = mapbox::geojson::parse(string.UTF8String).get<mapbox::geojson::feature_collection>();
+        const auto geojson = mapbox::geojson::parse(string.UTF8String);
         source->setGeoJSON(geojson);
-        _features = MGLFeaturesFromMBGLFeatures(geojson);
+        _shape = MGLShapeFromGeoJSON(geojson);
     } else {
-        mbgl::FeatureCollection featureCollection;
-        featureCollection.reserve(self.features.count);
-        for (id <MGLFeaturePrivate> feature in self.features) {
-            featureCollection.push_back([feature mbglFeature]);
-        }
-        const auto geojson = mbgl::GeoJSON{featureCollection};
+        const auto geojson = mbgl::GeoJSON{self.shape.geometryObject};
         source->setGeoJSON(geojson);
-        _features = MGLFeaturesFromMBGLFeatures(featureCollection);
     }
     
     _pendingSource = std::move(source);
@@ -161,10 +156,10 @@ const MGLGeoJSONSourceOption MGLGeoJSONSourceOptionSimplificationTolerance = @"M
     }
     
     NSString *string = [[NSString alloc] initWithData:_geoJSONData encoding:NSUTF8StringEncoding];
-    const auto geojson = mapbox::geojson::parse(string.UTF8String).get<mapbox::geojson::feature_collection>();
+    const auto geojson = mapbox::geojson::parse(string.UTF8String);
     self.rawSource->setGeoJSON(geojson);
     
-    _features = MGLFeaturesFromMBGLFeatures(geojson);
+    _shape = MGLShapeFromGeoJSON(geojson);
 }
 
 - (void)setURL:(NSURL *)URL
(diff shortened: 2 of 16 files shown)


1ec5:
I share you hesitation about removing `-[MGLGeoJSONSource initWithIdentifier:features:options:]`. `-[MGLShapeCollectionFeature shapeCollectionWithShapes:]` is kind of unwieldy. Should we ease the pain by keeping `-[MGLGeoJSONSource initWithIdentifier:features:options:]` as yet another initializer on MGLGeoJSONSource?

platform/darwin/src/MGLGeoJSONSource.mm
```diff
@@ -79,21 +80,43 @@ - (void)commonInit
...
         NSString *string = [[NSString alloc] initWithData:self.geoJSONData encoding:NSUTF8StringEncoding];
-        const auto geojson = mapbox::geojson::parse(string.UTF8String).get<mapbox::geojson::feature_collection>();
-        source->setGeoJSON(geojson);
-        _features = MGLFeaturesFromMBGLFeatures(geojson);
+        // TODO: Figure out a more performant way than serializing the data just to do a type check
+#warning - DEBUG CODE
+        id serialized = [NSJSONSerialization JSONObjectWithData:self.geoJSONData options:0 error:nil];
+        BOOL isFeatureCollection = [serialized isKindOfClass:NSArray.class] || ([serialized isKindOfClass:NSDictionary.class] && [[(NSDictionary *)serialized objectForKey:@"type"] isEqualToString:@"FeatureCollection"]);
+        
+        if (isFeatureCollection) {
```
1ec5: Why does this conditional exist? `mbgl::style::Source::setGeoJSON()` takes a `mapbox::geojson::geojson`, which is the type returned by `mapbox::geojson::parse()`, so no `get()` is needed for either case. `_feature` is of type `id <MGLFeature>`, so `MGLFeatureFromMBGLFeature()` works for both cases too.

platform/darwin/src/MGLGeoJSONSource.mm
```diff
@@ -79,21 +80,43 @@ - (void)commonInit
...
+            _feature = MGLFeatureFromMBGLFeature(geojson);
+        }
+        
     } else {
-        mbgl::FeatureCollection featureCollection;
-        featureCollection.reserve(self.features.count);
-        for (id <MGLFeaturePrivate> feature in self.features) {
-            featureCollection.push_back([feature mbglFeature]);
+        // Feature collection
+        if ([self.feature isMemberOfClass:MGLShapeCollectionFeature.class]) {
```
1ec5: MGLShapeCollectionFeature implements `-mbglFeature`, so this case is also unnecessary.

frederoni:
Removed debug code and cleaned up.
Left todo:
 - [ ] Support initializing MGLGeoJSONSource with a geometry?
 - [ ] `-[MGLGeoJSONSourceTests testMGLGeoJSONSourceWithPolygonFeatures]` is failing unexpectedly.
 - [ ] features are not set if `MGLGeoJSONSource` is initialized with data.

platform/darwin/src/MGLGeoJSONSource.mm
```diff
@@ -82,9 +93,8 @@ - (void)commonInit
         _features = nil;
     } else if (self.geoJSONData) {
         NSString *string = [[NSString alloc] initWithData:self.geoJSONData encoding:NSUTF8StringEncoding];
-        const auto geojson = mapbox::geojson::parse(string.UTF8String).get<mapbox::geojson::feature_collection>();
+        const auto geojson = mapbox::geojson::parse(string.UTF8String);
         source->setGeoJSON(geojson);
-        _features = MGLFeaturesFromMBGLFeatures(geojson);
```
frederoni: If we're gonna support plain geometries, we can't assign features here.
Wouldn't it make sense to rename features to geometries and write a GeoJSON evaluator which overloads  featureCollection, feature and geometry?

frederoni:
Depending on what we want, I think this is good for another review and possibly also merge.
#6986 doesn't occur anymore but there's a drawback. Using `-[MGLGeoJSONSource initWithIdentifier:geoJSONData:options:]` will leave the features property unset because it can contain plain geometries.

1ec5:
Maybe that means it needs to be a `shape` property (normally holding a shape collection feature), with `features` being a convenience property based on it?

platform/darwin/src/MGLGeoJSONSource.h
```diff
@@ -99,6 +99,19 @@ extern const MGLGeoJSONSourceOption MGLGeoJSONSourceOptionSimplificationToleranc
  specification</a>.
  
  @param identifier A string that uniquely identifies the source.
+ @param feature A feature that conform to the `MGLFeature` protocol.
+ @param options An `NSDictionary` of options for this source.
+ @return An initialized GeoJSON source.
+ */
+- (instancetype)initWithIdentifier:(NSString *)identifier feature:(nullable id<MGLFeature>)feature options:(nullable NS_DICTIONARY_OF(MGLGeoJSONSourceOption, id) *)options NS_DESIGNATED_INITIALIZER;
```
1ec5: Since `mbgl::style::GeoJSONSource::setGeoJSON()` takes an `mbgl::GeoJSON` (a variant of `geometry`, `feature`, or `feature_collection`), it would be better for this initializer to take any MGLShape. All MGLShape subclasses in turn have subclasses that conform to MGLFeature, so we wouldn’t be losing any expressiveness that way.

platform/darwin/src/MGLGeoJSONSource.mm
1ec5: We need a `shape` property that can be any MGLShape (or any MGLFeature). We can have a `features` property as a convenience for when `shape` happens to contain an MGLShapeCollectionFeature.

platform/darwin/src/MGLGeoJSONSource.h
```diff
@@ -114,7 +128,7 @@ extern const MGLGeoJSONSourceOption MGLGeoJSONSourceOptionSimplificationToleranc
  is set to `nil`. This property is unavailable until the receiver is passed into
  `-[MGLStyle addSource]`.
  */
-@property (nonatomic, nullable) NS_ARRAY_OF(id <MGLFeature>) *features;
+@property (nonatomic, nullable) MGLShape *shape;
```
1ec5: Update the documentation for this property.

platform/darwin/src/MGLGeoJSONSource.h
```diff
@@ -138,6 +152,10 @@ extern const MGLGeoJSONSourceOption MGLGeoJSONSourceOptionSimplificationToleranc
  */
 @property (nonatomic, nullable) NSURL *URL;
 
+/**
+ Returns all objects from the shape that conforms to `MGLFeature`.
```
1ec5: This makes it sound like it does a deep search through the shape collection feature for any shapes, whereas it's only set to the top-level features, and only if a shape collection feature or array of features was passed in on initialization.

platform/darwin/src/MGLGeoJSONSource.h
```diff
@@ -138,6 +152,10 @@ extern const MGLGeoJSONSourceOption MGLGeoJSONSourceOptionSimplificationToleranc
...
+ */
+- (NSArray<MGLFeature> *)features;
```
1ec5: Use `NS_ARRAY_OF()` in public APIs for the benefit of developers compiling against the iOS 8 or OS X 10.10 SDKs.

Also, this method should be a property like before.

Finally, the return type should be an array of `id <MGLFeature>`, not an NSArray conforming to the MGLFeature protocol. There's a big difference.

platform/darwin/src/MGLPolygon.mm
```diff
@@ -38,23 +38,19 @@ - (instancetype)initWithCoordinates:(CLLocationCoordinate2D *)coords count:(NSUI
...
 
 - (mbgl::Annotation)annotationObjectWithDelegate:(id <MGLMultiPointDelegate>)delegate {
-    mbgl::Polygon<double> geometry;
-    geometry.push_back(self.ring);
-    for (MGLPolygon *polygon in self.interiorPolygons) {
-        geometry.push_back(polygon.ring);
-    }
 
-    mbgl::FillAnnotation annotation { geometry };
+    auto polygon = mbgl::Polygon<double> { [self ring] };
```
1ec5: What about interior rings?

platform/darwin/src/MGLGeoJSONSource.mm
```diff
@@ -155,41 +171,47 @@ - (void)setGeoJSONData:(NSData *)geoJSONData
...
+    _shape = shape;
+    [self didChangeValueForKey:@"shape"];
+    
+    NS_MUTABLE_ARRAY_OF(id <MGLFeature>) *features = [NSMutableArray array];
+    
+    if ([_shape isMemberOfClass:[MGLShapeCollection class]]) {
+        MGLShapeCollection *collection = (MGLShapeCollection *)_shape;
+        for (MGLShape *shape in collection.shapes) {
+            if ([shape conformsToProtocol:@protocol(MGLFeature)]) {
+                [features addObject:(id <MGLFeature>)shape];
```
1ec5: Would it be more efficient to compute the `features` property as soon as `shape` is set, or on demand when calling `-features`? Or perhaps a lazy-loading strategy is called for, similar to how I treated `overlayBounds` in #7251.

platform/darwin/src/MGLGeoJSONSource.mm
```diff
@@ -49,9 +50,20 @@ - (instancetype)initWithIdentifier:(NSString *)identifier URL:(NSURL *)url optio
     return self;
 }
 
+- (instancetype)initWithIdentifier:(NSString *)identifier feature:(id<MGLFeature>)feature options:(NSDictionary<MGLGeoJSONSourceOption,id> *)options
```
1ec5: Any reason to require a feature instead of any ordinary shape? Looks like we’re only using this parameter for an MGLShape-typed property anyways.

frederoni: Mostly not to break the API. Should we rename it to `-[MGLGeoJSONSource initWithIdentifier:shape:options:]` or keep `feature:` and `features:` while adding `shape:` ?

platform/darwin/src/MGLGeoJSONSource.h
```diff
@@ -93,28 +94,29 @@ extern const MGLGeoJSONSourceOption MGLGeoJSONSourceOptionSimplificationToleranc
...
  specification</a>.
  
  @param identifier A string that uniquely identifies the source.
- @param features An array of features that conform to the `MGLFeature` protocol.
+ @param shape A concrete subclass of `MGLShape`
  @param options An `NSDictionary` of options for this source.
  @return An initialized GeoJSON source.
  */
-- (instancetype)initWithIdentifier:(NSString *)identifier features:(NSArray<id<MGLFeature>> *)features options:(nullable NS_DICTIONARY_OF(MGLGeoJSONSourceOption, id) *)options NS_DESIGNATED_INITIALIZER;
+- (instancetype)initWithIdentifier:(NSString *)identifier shape:(nullable MGLShape *)shape options:(nullable NS_DICTIONARY_OF(MGLGeoJSONSourceOption, id) *)options NS_DESIGNATED_INITIALIZER;
```
1ec5: What does it mean to create a source from no shapes at all? It does work in the current release branch, using an empty array of features, but we should make sure it works when shape is nil.

platform/darwin/src/MGLGeoJSONSource.mm
```diff
@@ -174,22 +168,20 @@ - (void)setURL:(NSURL *)URL
...
-    for (id <MGLFeaturePrivate> feature in features) {
-        featureCollection.push_back([feature mbglFeature]);
-    }
-    const auto geojson = mbgl::GeoJSON{featureCollection};
+
+    const auto geojson = mbgl::GeoJSON{shape.geometryObject};
     self.rawSource->setGeoJSON(geojson);
     
-    _features = MGLFeaturesFromMBGLFeatures(featureCollection);
+    [self willChangeValueForKey:@"shape"];
```
1ec5: These change notifications are unnecessary. Since setShape: is a standard KVC selector, these change notifications will be generated for you implicitly.

1ec5:
The Coveralls failure is caused by an unrelated change in test coverage in mbgl. You can get around it by pushing another change, so that coverage will be even.

platform/darwin/test/MGLGeoJSONSourceTests.mm
```diff
@@ -236,7 +242,8 @@ - (void)testMGLGeoJSONSourceWithShapeCollectionFeatures {
 
     MGLShapeCollectionFeature *shapeCollectionFeature_1 = [MGLShapeCollectionFeature shapeCollectionWithShapes:@[polygonFeature, polylineFeature, multiPolygonFeature, multiPolylineFeature, pointCollectionFeature, pointFeature, shapeCollectionFeature]];
 
-    XCTAssertThrowsSpecificNamed([[MGLGeoJSONSource alloc] initWithIdentifier:@"source-id" features:@[shapeCollectionFeature_1] options:nil], NSException, @"Method unavailable");
+#warning DEBUG
+    //XCTAssertThrowsSpecificNamed([[MGLGeoJSONSource alloc] initWithIdentifier:@"source-id" shape:shapeCollectionFeature_1 options:nil], NSException, @"Method unavailable");
```
frederoni: @1ec5 what was the purpose of this test?

boundsj:
The iOS release branch now has https://github.com/mapbox/mapbox-gl-native/commit/c8e34d581b80beea247229051795a8ab3920c378 so if this branch is rebased on top of that, the node tests will pass.

platform/darwin/src/MGLGeoJSONSource.h
frederoni: nil shape works but I can't think of any case where you'd want to do that.

PR #7173, opened 2016-11-23
Fix integer formatting in formatted exception strings

boundsj:
Fixes issue noted in https://github.com/mapbox/mapbox-gl-native/pull/7083#pullrequestreview-8705064

cc @1ec5

### DIFF
--- a/platform/darwin/src/MGLMultiPoint.mm
+++ b/platform/darwin/src/MGLMultiPoint.mm
@@ -58,8 +58,8 @@ mbgl::Color MGLColorObjectFromCGColorRef(CGColorRef cgColor)
     if (range.location + range.length > [self pointCount])
     {
         [NSException raise:NSRangeException
-                    format:@"Invalid coordinate range %@ extends beyond current coordinate count of %@",
-                        NSStringFromRange(range), @([self pointCount])];
+                    format:@"Invalid coordinate range %@ extends beyond current coordinate count of %ld",
+                        NSStringFromRange(range), (unsigned long)[self pointCount]];
     }
 
     std::copy(_coordinates.begin() + range.location, _coordinates.begin() + NSMaxRange(range), coords);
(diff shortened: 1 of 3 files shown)


platform/darwin/src/MGLMultiPoint.mm
```diff
@@ -58,8 +58,8 @@ - (void)getCoordinates:(CLLocationCoordinate2D *)coords range:(NSRange)range
     if (range.location + range.length > [self pointCount])
     {
         [NSException raise:NSRangeException
-                    format:@"Invalid coordinate range %@ extends beyond current coordinate count of %@",
-                        NSStringFromRange(range), @([self pointCount])];
+                    format:@"Invalid coordinate range %@ extends beyond current coordinate count of %ld",
+                        NSStringFromRange(range), (long)[self pointCount]];
```
1ec5: Shouldn't this be unsigned? What I usually do is use `%@` with an uncasted pointCount and wait for an Xcode fix-it to tell me which format specifier and cast to use.

boundsj: oops! df1d8fd fixes this and corrects one other place that should be unsigned.

platform/ios/app/MBXViewController.m
```diff
@@ -1596,7 +1596,7 @@ - (void)mapViewRegionIsChanging:(MGLMapView *)mapView
         {
             queuedAnnotations += queue.count;
         }
-        self.hudLabel.text = [NSString stringWithFormat:@"Visible: %ld  Queued: %ld", (long)mapView.visibleAnnotations.count, (long)queuedAnnotations];
+        self.hudLabel.text = [NSString stringWithFormat:@"Visible: %ld  Queued: %ld", (unsigned long)mapView.visibleAnnotations.count, (long)queuedAnnotations];
```
1ec5: `queuedAnnotations` is `unsigned long` too.

Issue #7174, opened 2016-11-23
MapboxMapOptions don't include a default CameraPosition


Guardiola31337:
I run into a NPE when launching `MyLocationDrawableActivity` because it uses `setInitialState` method from `MapView` and it tries to access to the `CameraPosition` from `MapboxMapOptions`.

I could add the `CameraPosition` in the example using `camera` method to solve the example.
But maybe the SDK should have that into account having a default implementation of `CameraPosition` so users don't run into this issue.

tobrun:
Thank you for flagging

PR #7177, opened 2016-11-23
Add location related tests

### LINKED ISSUE #7159: Dynamically grant permissions in UI tests
There are some tests excluded from being generated and/or ignored because a way to grant permissions dynamically in tests wasn't implemented.

There are some workarounds to grant permissions dynamically in tests.
Including:
* `adb install -g`
* [PermissionGranter](https://gist.github.com/rocboronat/65b1187a9fca9eabfebb5121d818a3c4)
* [Grant permissions before tests](https://github.com/googlesamples/android-testing/blob/ed62c450e43f859333b3113d44dd59f75971b529/ui/espresso/IntentsBasicSample/app/src/androidTest/java/com/example/android/testing/espresso/BasicSample/DialerActivityTest.java#L94)

I think the latter would be the best approach.

Below all the activities that request Location permissions:
`MyLocationTrackingModeActivity`, `MyLocationToggleActivity`, `MyLocationTintActivity`, `MyLocationDrawableActivity`, `LocationPickerActivity` and `MyLocationViewTest`.

Cc/ @tobrun @zugaldia

Guardiola31337:
Fixes #7159.

* Added location permission granter.
* Removed location related activities (except `MyLocationDrawableActivity` because of #7174) from being excluded.
* Removed @Ignore annotation of `testEnabled()` (`MyLocationViewTest`).

Note: Still ignoring `testTracking()` (`MyLocationViewTest`).

Cc/ @tobrun @zugaldia

### DIFF
--- a/platform/android/MapboxGLAndroidSDKTestApp/src/androidTest/java/com/mapbox/mapboxsdk/PermissionGranter.java
+++ b/platform/android/MapboxGLAndroidSDKTestApp/src/androidTest/java/com/mapbox/mapboxsdk/PermissionGranter.java
@@ -1,0 +1,21 @@
+package com.mapbox.mapboxsdk;
+
+import android.Manifest;
+import android.os.Build;
+
+import static android.support.test.InstrumentationRegistry.getInstrumentation;
+import static android.support.test.InstrumentationRegistry.getTargetContext;
+
+public class PermissionGranter {
+
+  private static final String PM_GRANT = "pm grant";
+  private static final String SEPARATOR = " ";
+
+  public static void grantLocation() {
+    if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.M) {
+      getInstrumentation().getUiAutomation().executeShellCommand(
+          PM_GRANT + SEPARATOR + getTargetContext().getPackageName() +
+          SEPARATOR + Manifest.permission.ACCESS_FINE_LOCATION);
+    }
+  }
+}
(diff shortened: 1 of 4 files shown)


Guardiola31337:
There is an issue with the permission dialog popup. It is still appearing even granting permissions (`PermissionGranter. grantLocation()`) beforehand. Maybe, the second option defined here #7177 is better. The problem is that [UIAutomator](https://developer.android.com/topic/libraries/testing-support-library/index.html#UIAutomator) requires API level 18 or higher and changing the `minSdkVersion` of testapp should be discussed.

We should discuss as well how to test activities which require permissions because if you get rid of the permission dialog (e.g. denying it forever) tests pass (because we aren't testing anything that involves permissions) and `PermissionGranter. grantLocation()` isn't doing anything, which means tests aren't giving any value in the permissions aspect.

What do you think @tobrun @zugaldia @cammace?

tobrun:
I will look into resolving conflicts and the remarks mentioned in OP.

Guardiola31337:
I managed to solve the problem with the popup dialog. The problem was that @Before is executed after the activity is launched and obviously before the test. Which means the activity is launched first and shows the dialog.
So, in this particular case @BeforeClass solves the issue.

Guardiola31337:
`testEnabled()` from `MyLocationViewTest` is not working. It's related with the order between asking for `LocationServices.getLocationServices(view.getContext()).getLastLocation()` (called in `ToggleLocationAction`) and when permissions are actually granted.

Guardiola31337:
@tobrun
I run `MyLocationTintActivityTest` locally and it's working. What error did you get?
I don't know what happened with the last Bitrise check. Could you tell me?

Guardiola31337:
Remember to remove from being excluded when #7181 is merged.

tobrun:
I saw the same thing here @Guardiola31337, but also noticed that CI was failing over a strange native crash of updateMarker (which isn't anywhere found in the code executed). Now disabling that test shows that another location related test is failing on the same updateMarker function. Need to make some time to dig into this..

zugaldia:
@Guardiola31337 @tobrun Any progress here? Do you need my 👀 ?

1ec5:
Since this PR is scheduled for Android SDK v5.0.0, it needs to be rebased and retargeted at the release-ios-v3.5.0-android-v5.0.0 branch.

tobrun:
Since this PR is related to testing already existing features, moving it from the 5.0.0 milestone.

tobrun:
Initial issue this PR was closing was made obsolete, closing this PR as well.

Issue #6912, opened 2016-11-04
Convert annotations API to use peer model


tobrun:
Our current annotations API relies heavily on MapboxMap/Mapview as a gateway to use jni to talk to core. There is actually no need to route through these classes if we implement our annotations as peer models, the classes themselves will be able to call into jni.

This type of construction is heavily used in runtime style implementation and separates the code more cleanly. An example of such a peer model class in details below.

cc @ivovandongen 

<details>

```java
/**
 * A GeoJson source. Exposes a {@link FeatureCollection} from Json.
 *
 * @see <a href="https://www.mapbox.com/mapbox-gl-style-spec/#sources-geojson">the style specification</a>
 */
public class GeoJsonSource extends Source {

    /**
     * Internal use
     *
     * @param nativePtr - pointer to native peer
     */
    public GeoJsonSource(long nativePtr) {
        super(nativePtr);
    }

    /**
     * Create an empty GeoJsonSource
     *
     * @param id the source id
     */
    public GeoJsonSource(String id) {
        initialize(id, null);
        setGeoJson(FeatureCollection.fromFeatures(new ArrayList<Feature>()));
    }

    /**
     * Create an empty GeoJsonSource with non-default {@link GeoJsonOptions}
     *
     * @param id      the source id
     * @param options options
     */
    public GeoJsonSource(String id, GeoJsonOptions options) {
        initialize(id, options);
        setGeoJson(FeatureCollection.fromFeatures(new ArrayList<Feature>()));
    }

    /**
     * Create a GeoJsonSource from a raw json string
     *
     * @param id      the source id
     * @param geoJson raw Json FeatureCollection
     */
    public GeoJsonSource(String id, String geoJson) {
        if (geoJson == null || geoJson.startsWith("http")) {
            throw new IllegalArgumentException("Expected a raw json body");
        }
        initialize(id, null);
        setGeoJson(geoJson);
    }

    /**
     * Create a GeoJsonSource from a raw json string and non-default {@link GeoJsonOptions}
     *
     * @param id      the source id
     * @param geoJson raw Json body
     * @param options options
     */
    public GeoJsonSource(String id, String geoJson, GeoJsonOptions options) {
        if (geoJson == null || geoJson.startsWith("http")) {
            throw new IllegalArgumentException("Expected a raw json body");
        }
        initialize(id, options);
        setGeoJson(geoJson);
    }

    /**
     * Create a GeoJsonSource from a remote geo json file
     *
[...]

Issue #6067, opened 2016-08-18
AnnotationManager.java refactor


tobrun:
The idea behind this refactor is to extract logic from MapboxMap and use it as facade to route the public api internally to the concept of Annotation Manager. This mimics the behaviour we have in core. We should back up this with unit tests for adding/removing and managing selection states.

PR #7192, opened 2016-11-25
7087 remove annotation base

### LINKED ISSUE #7087: Remove Annotation.java
Currently all our classes found in the `annotations` package extend from `Annotation.java`. The annotation class is responsible for managing Id, Mapview and MapboxMap. With converting all the subclasses to use a Peer model in #6912, we will be able to remove MapView and MapboxMap. The only thing that annotation class will manage at that point is id and I don't think we should enforce an abstraction for only id. (note G-maps doesn't have this abstraction either).

The change of removing the annotation class will cascade in removing the SpareArray that currently manages all the different supported annotations (markers, polygons and polylines). This will result in introducing collections for each type of annotation (we don't need to iterate each type of annotation every time we want to update an annotation). 

While implementing this change we also need to revisit our exposed API on MapboxMap.

tobrun:
WIP, closes #7087, 
second refactor on Annotations after #6067, 
makes way to implement peer model from #6912.


tobrun:
Closing as this PR has diverged too much from master with ongoing refactors.

Issue #7143, opened 2016-11-22
App crashes after shader compilation failure 


ChristopherUNIS:
**Platform: Linux**
**Mapbox SDK version: f92439d1807d743bddf48675df2414c7d0cd71d2**

### Steps to trigger behavior

 1. Compile qt platform with command "make  run-qt-qml-app" 
 
### Expected behavior

App runs

### Actual behavior

App crashes with following error : 

"[ERROR] {mbgl-qt}[Shader]: Shader failed to compile: 0:94(20): error: could not implicitly convert operands to arithmetic operator"

### Findings

Git bisect shows that this bug starts with commit 3380f0803a62525a6884ee4d05e103e072c36117

eek001c:
I've tested fix from issue #7197.
mbgl-qt works fine.

tmpsantos:
I get the same, the issue is that depending on your computer local, you get a

`#define DEVICE_PIXEL_RATIO 1,000000` instead of `#define DEVICE_PIXEL_RATIO 1.000000`

This is only reproducible if your locale uses `,` for decimal points instead of `.`, which is not the case for the US and the UK but it is for Brazil, France, Russia and pretty much the rest of the planet.

tmpsantos:
Fixed by https://github.com/mapbox/mapbox-gl-native/issues/7226

PR #7197, opened 2016-11-27
Always use a dot as a decimal point in string, independent of locale

akikoskinen:
On my environment the std::to_string() call inserted a comma (,) as a decimal point into the string. Obviously this caused compilation errors in shaders. The generated string should always use a dot (.) as the delimiter. 

Now the code proposed in this PR is probably not the best implementation (in terms of effectiveness), as it uses stringstreams and plays with locales. If anybody has a better idea how to do this, I'm very willing to change this.

### DIFF
--- a/src/mbgl/programs/program.hpp
+++ b/src/mbgl/programs/program.hpp
@@ -3,6 +3,7 @@
 #include <mbgl/gl/program.hpp>
 #include <mbgl/programs/program_parameters.hpp>
 
+#include <sstream>
 #include <cassert>
 
 namespace mbgl {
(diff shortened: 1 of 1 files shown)


akikoskinen:
Only after submitting this I found the a `mbgl::util::toString(float)` method which would probably be faster. Unfortunately that method is wise enough to leave out the fractional part if it's zero. So in this case if `pixelRatio` is 1.0, it outputs "1", which also generates a shader compilation error.

mollymerp:
Thanks for this PR @akikoskinen!! We've gotten a couple reports of compiler errors for shaders but they've been hard to diagnose without being able to reproduce. I will take a look at this approach and see if it will work for us!

akikoskinen:
Issue #7143 seems familiar

tmpsantos:
@akikoskinen thanks for your contribution, I merged it.

PR #7203, opened 2016-11-28
Check for duplicate source IDs

### LINKED ISSUE #7011: Check for duplicate source IDs
ATM there is no check on source ID when adding it to the style. This leads to issues like https://github.com/mapbox/mapbox-gl-native/issues/6612#issuecomment-259686219

ivovandongen:
Fixes #7011

### DIFF
--- a/platform/darwin/src/MGLStyle.mm
+++ b/platform/darwin/src/MGLStyle.mm
@@ -223,7 +223,11 @@ static NSURL *MGLStyleURL_emerald;
          source];
     }
 
-    [source addToMapView:self.mapView];
+    try {
+        [source addToMapView:self.mapView];
+    } catch (std::runtime_error & err) {
+        [NSException raise:@"Could not add source" format:@"%s", err.what()];
+    }
 }
 
 - (void)removeSource:(MGLSource *)source
--- a/platform/ios/test/MGLSourceTests.m
+++ b/platform/ios/test/MGLSourceTests.m
@@ -1,0 +1,24 @@
+#import "MGLMapViewTests.h"
+
+@interface MGLSourceTests : MGLMapViewTests
+
+@end
+
+@implementation MGLSourceTests
+
+- (void)testDuplicateSources {
+    MGLVectorSource *source1 = [[MGLVectorSource alloc] initWithIdentifier:@"my-source" URL:[NSURL URLWithString:@"mapbox://mapbox.mapbox-terrain-v2"]];
+    MGLVectorSource *source2 = [[MGLVectorSource alloc] initWithIdentifier:@"my-source" URL:[NSURL URLWithString:@"mapbox://mapbox.mapbox-terrain-v2"]];
+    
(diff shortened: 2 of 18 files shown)


test/style/style.test.cpp
```diff
@@ -113,3 +116,21 @@ TEST(Style, Properties) {
...
+
+TEST(Style, DuplicateSource) {
+    util::RunLoop loop;
+
+    StubFileSource fileSource;
+    Style style { fileSource, 1.0 };
+
+    style.setJSON(util::read_file("test/fixtures/resources/style-unused-sources.json"));
+
+	style.addSource(std::make_unique<VectorSource>("sourceId", "mapbox://mapbox.mapbox-terrain-v2"));
```
jfirebaugh: Mixed spaces and tabs.

platform/darwin/src/MGLStyle.mm
```diff
@@ -223,7 +223,11 @@ - (void)addSource:(MGLSource *)source
          source];
     }
 
-    [source addToMapView:self.mapView];
+    try {
+        [source addToMapView:self.mapView];
+    } catch (std::runtime_error & err) {
+        [NSException raise:@"Could not add source" format:@"%s", err.what()];
```
jfirebaugh: Darwin-side test for this error handling?

ivovandongen:
@jfirebaugh With @frederoni's help I've added unit tests for osx/ios as well.

platform/ios/test/MGLSourceTests.m
```diff
@@ -0,0 +1,24 @@
...
+
+@implementation MGLSourceTests
+
+- (void)testDuplicateSources {
+    MGLVectorSource *source1 = [[MGLVectorSource alloc] initWithIdentifier:@"my-source" URL:[NSURL URLWithString:@"mapbox://mapbox.mapbox-terrain-v2"]];
+    MGLVectorSource *source2 = [[MGLVectorSource alloc] initWithIdentifier:@"my-source" URL:[NSURL URLWithString:@"mapbox://mapbox.mapbox-terrain-v2"]];
+    
+    [self.mapView.style addSource: source1];
+    
+    @try {
```
1ec5: FYI, `@try`-`@catch` blocks are generally to be avoided in Objective-C code due to their poor performance. It’s technically necessary here because we’re trying to detect an exception being raised. However, the more idiomatic way to assert for the exception is to use the `XCTAssertThrows()`, `XCTAssertThrowsSpecific()`, or `XCTAssertThrowsSpecificNamed()` macro.

boundsj: I think that once this change is merged into the iOS release branch it might make more sense to do these tests in `MGLStyleTests` along with what has been added in https://github.com/mapbox/mapbox-gl-native/pull/7272/files#diff-ce3b7fe93109fd552f9627d0dcc0ceb4R107 for checking for duplicate instances.

PR #7204, opened 2016-11-28
[android] - disable CI failing test

tobrun:
Locally this tests works but on CI it fails.
Short term solution is to ignore the test to unblock others.

Review @ivovandongen


tobrun:
@ivovandongen confirmed that this is no longer an issue, closing

tobrun:
We are still seeing occurrences of this on CI, reopening

Issue #7163, opened 2016-11-23
Backend::activate/deactivate need to be reentrant


jfirebaugh:
In some contexts, `Backend::activate` can be called when the backend is already active. One example is if `Map::addLayer` is called during notifications such as `MapChangeWillStartRenderingFrame` and `MapChangeRegionDidChangeAnimated`.

Therefore `Backend::activate` / `Backend::deactivate` need to count the number of calls in a reference counting manner. Deactivation should happen only when the count reaches 0.

I feel like we did this at one point. Did it regress as part of the `Backend` changes?

cc @kkaefer @boundsj

jfirebaugh:
#5874, #5836, #5721

jfirebaugh:
Ok, looks like we refcount on Android but not iOS.

boundsj:
This was fixed by https://github.com/mapbox/mapbox-gl-native/pull/7208

PR #7208, opened 2016-11-28
[ios, macos] Refcount view activations

boundsj:
This adds a ref count to guard against calling activate after it has already been called. This can happen when a layer is added during certain map lifecycle methods like MapChangeWillStartRenderingFrame that also triggers activation. This also protects against issues related to a deactivation of the context happening in an inner scope while an outer scope is still expecting to be able to use the context.

Fixes https://github.com/mapbox/mapbox-gl-native/issues/7163

cc @1ec5 @incanus @jfirebaugh

### DIFF
--- a/platform/ios/src/MGLMapView.mm
+++ b/platform/ios/src/MGLMapView.mm
@@ -5068,16 +5068,28 @@ public:
 
     void activate() override
     {
+        if (activationCount++)
+        {
+            return;
+        }
+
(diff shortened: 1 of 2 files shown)


kkaefer:
Was this merged into master?

1ec5:
Not yet. We can merge the release branch back to master if you need this fix immediately; otherwise, we'll do it right after the next beta, which should be out soon.

PR #7212, opened 2016-11-29
[ios] Use new pinning cert for staging

boundsj:
This updates the cert for the staging environment. It also adjusts the telem implementation to send along the horizontal accuracy for each location event and removes the previous battery level gathering logic.


platform/ios/src/MGLMapboxEvents.m
```diff
@@ -665,11 +656,13 @@ - (void)locationManager:(MGLLocationManager *)locationManager didUpdateLocations
         double accuracy = 10000000;
         double lat = floor(loc.coordinate.latitude * accuracy) / accuracy;
         double lng = floor(loc.coordinate.longitude * accuracy) / accuracy;
+        double horizontalAccuracy = loc.horizontalAccuracy;
```
1ec5: By the way, technically this should be a CLLocationAccuracy, which happens to be a typedef for double.

boundsj:
@tobrun @ivovandongen @cammace 38cabe3 moves the SSL cert pinning files to an iOS specific location and removes any reference to them in the `make` configuration for android

platform/ios/src/MGLMapboxEvents.m
1ec5: According to the [`horizontalAccuracy`](https://developer.apple.com/reference/corelocation/cllocation/1423599-horizontalaccuracy) documentation, a value of −1 indicates an invalid `coordinate`. We should check for this case (and also `coordinate` being invalid) and avoid sending the event.

PR #7213, opened 2016-11-29
Expand documentation on MGLTileSource URL templates

incanus:
Pithy example, external reference link, and clarification. 

As best I can tell, only the first template string is ever used (cf. [`offline_download.cpp`](https://github.com/mapbox/mapbox-gl-native/blob/b57b3b4bf6841be5604c54c94db6b9d822d985e1/platform/default/mbgl/storage/offline_download.cpp#L246) & [`tile_loader_impl.hpp`](https://github.com/mapbox/mapbox-gl-native/blob/b57b3b4bf6841be5604c54c94db6b9d822d985e1/src/mbgl/tile/tile_loader_impl.hpp#L20))

/cc @1ec5 @boundsj @jfirebaugh

### DIFF
--- a/platform/darwin/src/MGLTileSource.h
+++ b/platform/darwin/src/MGLTileSource.h
@@ -148,6 +148,13 @@ typedef NS_ENUM(NSUInteger, MGLTileCoordinateSystem) {
 /**
  Returns a tile source initialized an identifier, tile URL templates, and
  options.
+ 
+ Tile URL template strings are a loose open standard for referring to custom
+ locations from which to download tiles, whether vector or raster. An example
+ template string is `http://myserver.com/tiles/{z}/{x}/{y}.pbf`.
+ 
+ In the current implementation, only the first tile URL template string
+ specified is used.
 
  After initializing and configuring the source, add it to a map view’s style
  using the `-[MGLStyle addSource:]` method.
@@ -159,6 +166,9 @@ typedef NS_ENUM(NSUInteger, MGLTileCoordinateSystem) {
     `MGLTileSourceOption` for available keys and values. Pass in `nil` to use
     the default values.
  @return An initialized tile source.
+ 
+ @see <a href="https://wiki.openstreetmap.org/wiki/Slippy_map_tilenames">OpenStreetMap's 
+    wiki page</a> for more information on tile URL template strings.
  */
 - (instancetype)initWithIdentifier:(NSString *)identifier tileURLTemplates:(NS_ARRAY_OF(NSString *) *)tileURLTemplates options:(nullable NS_DICTIONARY_OF(MGLTileSourceOption, id) *)options;
 
--- a/platform/darwin/src/MGLTileSource.h
+++ b/platform/darwin/src/MGLTileSource.h
@@ -150,11 +150,20 @@ typedef NS_ENUM(NSUInteger, MGLTileCoordinateSystem) {
  options.
  
  Tile URL template strings are a loose open standard for referring to custom
- locations from which to download tiles, whether vector or raster. An example
- template string is `http://myserver.com/tiles/{z}/{x}/{y}.pbf`.
+ locations from which to download tiles, whether vector or raster.
  
- In the current implementation, only the first tile URL template string
- specified is used.
+ Currently the tokens `x`, `y`, and `z` are supported for tile locations in the
+ world grid, as well as `ratio` for use in raster tiles, which allows
+ pre-rendered imagery to adapt to the current `UIScreen.scale`.
+ 
+ Examples:
+ 
+ <ul>
+ <li><code>http://abc.com/tiles/{z}/{x}/{y}.pbf</code> might become <code>http://abc.com/tiles/14/6/9.pbf</code></li>
+ <li><code>http://abc.com/tiles/{z}/{x}/{y}{ratio}.png</code> might become <code>http://abc.com/tiles/14/6/9@2x.png</code></li>
+ </ul>
+ 
+ Only the first tile URL template string specified is used.
 
  After initializing and configuring the source, add it to a map view’s style
  using the `-[MGLStyle addSource:]` method.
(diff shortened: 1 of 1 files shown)


platform/darwin/src/MGLTileSet.h
```diff
@@ -14,6 +14,16 @@ typedef NS_ENUM(NSUInteger, MGLTileSetScheme) {
  configuration for those strings. It can be passed to an `MGLVectorSource` or
  `MGLRasterSource` instead of an `NSURL` representing a TileJSON URL to create a 
  source.
+ 
+ Tile URL template strings are a loose open standard for referring to custom 
+ locations from which to download tiles, whether vector or raster. An example 
+ template string is `http://myserver.com/tiles/{z}/{x}/{y}.pbf`.
+ 
+ In the current implementation, only the first tile URL template string specified
```
1ec5: If this fact is subject to change, as an implementation detail, we should leave it out of documentation. If it won't change, we should allow only the first template to be set and ignore any other templates that happen to be set in the style. That would simplify the API a little.

platform/darwin/src/MGLTileSet.h
```diff
@@ -14,6 +14,16 @@ typedef NS_ENUM(NSUInteger, MGLTileSetScheme) {
...
+ is used.
+ 
+ @see <a href="https://wiki.openstreetmap.org/wiki/Slippy_map_tilenames">OpenStreetMap's
```
1ec5: It's fine to link to this wiki page, but we need to enumerate the tokens we support, because it's different than the list of tokens used in OSM editors. (Each editor has a different set of supported tokens, in fact.)

platform/darwin/src/MGLTileSet.h
```diff
@@ -14,6 +14,16 @@ typedef NS_ENUM(NSUInteger, MGLTileSetScheme) {
  configuration for those strings. It can be passed to an `MGLVectorSource` or
  `MGLRasterSource` instead of an `NSURL` representing a TileJSON URL to create a 
  source.
+ 
+ Tile URL template strings are a loose open standard for referring to custom
```
1ec5: If the developer is getting Quick Help for one of the initializers, they're going to miss this explanation. Better to put it in the property's documentation comment and refer to it in the comments for the initializers.

platform/darwin/src/MGLTileSet.h
incanus: I think it's probably easiest to just enumerate `x`, `y`, and `z`, and `ratio` (for raster) at the platform level, since these are the ones used nearly all the time by self-hosted setups.

platform/darwin/src/MGLTileSet.h
incanus: I don't think we know if it will change. Why should we leave it out? While I can't find an example right now, I've seen a number of reserved, unused arguments or options enumerations with no indication as to future plans. I think this is fine.

platform/darwin/src/MGLTileSet.h
incanus: ac1bfafe2894e15c6b24c75b59060a11c8733884 tidies this up a bit and includes better examples.

platform/darwin/src/MGLTileSet.h
incanus: Enumerated in https://github.com/mapbox/mapbox-gl-native/commit/ac1bfafe2894e15c6b24c75b59060a11c8733884's examples.

platform/darwin/src/MGLTileSet.h
```diff
@@ -19,6 +19,27 @@ typedef NS_ENUM(NSUInteger, MGLTileSetScheme) {
 
 /**
  An `NSArray` of `NSString` objects that represent the tile templates.
+ 
+ Tile URL template strings are a loose open standard for referring to custom
+ locations from which to download tiles, whether vector or raster.
+
+ Currently the tokens `x`, `y`, and `z` are supported for tile locations
```
1ec5: Also [`quadkey`, `bbox-epsg-3857`, and `prefix`](https://github.com/mapbox/mapbox-gl-native/blob/dbb520662d0145a0219e319a67bbe81326c144e3/src/mbgl/storage/resource.cpp#L105-L113). (`bbox-epsg-3857` is used for our [WMS support](https://www.mapbox.com/mapbox-gl-js/example/wms/).) Each of these tokens requires some explanation, so a bulleted list would be justified here.

platform/darwin/src/MGLTileSet.h
```diff
@@ -19,6 +19,27 @@ typedef NS_ENUM(NSUInteger, MGLTileSetScheme) {
...
+ in the world grid, as well as `ratio` for use in raster tiles, which allows
+ pre-rendered imagery to adapt to the current `UIScreen.scale`.
```
1ec5: macOS uses `NSScreen.backingScaleFactor` instead of `UIScreen.scale`.

platform/darwin/src/MGLTileSet.h
```diff
@@ -19,6 +19,27 @@ typedef NS_ENUM(NSUInteger, MGLTileSetScheme) {
...
+ <ul>
+ <li><code>http://abc.com/tiles/{z}/{x}/{y}.pbf</code> might become 
+    <code>http://abc.com/tiles/14/6/9.pbf</code></li>
+ <li><code>http://abc.com/tiles/{z}/{x}/{y}{ratio}.png</code> might become 
+    <code>http://abc.com/tiles/14/6/9@2x.png</code></li>
+ </ul>
+
+ Only the first tile URL template string specified is used.
+
+ @see <a href="https://wiki.openstreetmap.org/wiki/Slippy_map_tilenames">OpenStreetMap's
```
1ec5: While this is an incredibly useful wiki page, we need to treat it with caution because it assumes 256×256-pixel raster tiles as opposed to the 512×512-pixel tiles we use. In other words, `{z}` is one off.

platform/darwin/src/MGLTileSet.h
1ec5: Ah, I see you chose option 1 already. 👌

1ec5:
First of all, this PR needs to be retargeted at the v3.4.0 release branch if it’s going to make it into v3.4.0. There are enough conflicting changes over there that you might want to manually redo your changes on that branch. In particular, MGLTileSet has been renamed MGLTileSource, and its methods are a bit different than MGLTileSet’s.

1ec5:
tileurltemplates

PR #7210, opened 2016-11-29
Updating mapbox-gl-shaders doesn't trigger build-shaders.js

kkaefer:
If you update the commit hash of mapbox-gl-shaders in package.json, run `make xproj`, and build, it should run `build-shaders.js` for each of the modified shaders. But it doesn't -- the old output is used.

cc @kkaefer

### DIFF
--- a/cmake/shaders.cmake
+++ b/cmake/shaders.cmake
@@ -6,6 +6,7 @@ function(add_shader VAR name)
     add_custom_command(
        OUTPUT ${shader_source_prefix}/${name}.hpp
        COMMAND ${shader_build_cmd} ${name} ${shader_file_prefix} ${shader_source_prefix}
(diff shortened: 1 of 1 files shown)


cmake/shaders.cmake
```diff
@@ -6,6 +6,7 @@ function(add_shader VAR name)
     add_custom_command(
        OUTPUT ${shader_source_prefix}/${name}.hpp
        COMMAND ${shader_build_cmd} ${name} ${shader_file_prefix} ${shader_source_prefix}
+       DEPENDS ${CMAKE_SOURCE_DIR}/scripts/build-shaders.js
        DEPENDS ${shader_file_prefix}/${name}.vertex.glsl
        DEPENDS ${shader_file_prefix}/${name}.fragment.glsl
```
kkaefer: Changes to the actual GLSL files are tracked here and should trigger a rebuild if modified.

PR #7226, opened 2016-11-29
Always use a dot as a decimal point in string, independent of locale

tmpsantos:
std::to_string() potentially uses a decimal point character different
than a dot (.), depending on the environment's locale. But since this
generated string will be part of GLSL code, there needs to be a dot and
nothing else.

https://github.com/mapbox/mapbox-gl-native/pull/7197


tmpsantos:
/cc @brunoabinader @akikoskinen

I re-pushed the PR to run on the bots. Commit authorship was preserved.

kkaefer:
great catch!

Issue #7211, opened 2016-11-28
-[MGLStyle removeLayer:] relies on undefined behavior


1ec5:
Per https://github.com/mapbox/mapbox-gl-native/pull/6097#discussion_r89903599, the `-removeFromMapView:` method in each of the concrete MGLStyleLayer subclasses relies on undefined behavior by `reinterpret_cast`ing a `unique_ptr` to the `mbgl::style::Layer`. This code was added to the iOS and macOS SDKs in #7048. The Android SDK is doing something a bit different in #7039: `move`-ing a `unique_ptr` but not `reinterpret_cast`ing it.

/cc @jfirebaugh @boundsj

boundsj:
This was fixed by https://github.com/mapbox/mapbox-gl-native/pull/7227

PR #7227, opened 2016-11-29
Avoid unsafe cast when removing style layers

boundsj:
This updates the code generation for style layers to add a test to check that the style object returned from mbgl can be safely cast into the expected type so ownership can be returned to the caller when removing layers.

Previously a crash occurred if a layer of type T was added before a layer of type U but both layers shared the same identifier. In that case, if the pointer to the second layer of type U
was used to remove that layer, a runtime exception would occur since mbgl returns the layer with type T (the first one added) and the unsafe cast was triggered.

Now, even though the unsafe cast is avoided, an exception will still occur if layers with the same identifier are manipulated as described above but this is done as part of a validation check for all public style layer methods at the Darwin platform level. The raised `NSException` attempts to warn developers about reusing style identifiers to avoid this issue in the first place.

Fixes https://github.com/mapbox/mapbox-gl-native/issues/7211

cc @1ec5 @jfirebaugh

### DIFF
--- a/platform/darwin/src/MGLStyleLayer_Private.h
+++ b/platform/darwin/src/MGLStyleLayer_Private.h
@@ -5,6 +5,25 @@
 
 #include <mbgl/style/layer.hpp>
 
+/**
+ Assert that the style layer is valid.
+
+ This macro should be used at the beginning of any public-facing instance method
+ of `MGLStyleLayer` and its subclasses. For private methods, an assertion is more appropriate.
+ */
+#define MGLAssertStyleLayerIsValid() \
+    do { \
+        if (!self.rawLayer) { \
+            [NSException raise:@"Invalid style layer" \
+                        format: \
+            @"-[MGLStyle removeLayer:] has been called " \
+            @"with this instance but another style layer instance was added with the same identifer. It is an " \
+            @"error to send any message to this layer since it cannot be recovered after removal due to the " \
+            @"identifer collision. Use unique identifiers for all layer instances including layers of " \
+            @"different types."]; \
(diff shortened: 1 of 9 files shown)


platform/darwin/src/MGLStyleLayer_Private.h
```diff
@@ -5,6 +5,25 @@
...
 #include <mbgl/style/layer.hpp>
 
+/**
+ Assert that the style layer is valid.
+
+ This macro should be used at the beginning of any public-facing instance method
+ of `MGLStyleLayer` and its subclasses. For private methods, an assertion is more appropriate.
+ */
+#define MGLAssertStyleLayerIsValid() \
+do { \
```
1ec5: Please indent this macro so it’s more readable.

platform/darwin/src/MGLStyleLayer_Private.h
```diff
@@ -5,6 +5,25 @@
...
+if (!self.rawLayer) { \
+[NSException raise:@"Invalid style layer" \
+format: \
+@"-[%@ removeLayer:] has been called " \
```
1ec5: The method is named `-[MGLStyle removeLayer:]`, not `-[MGL*StyleLayer removeLayer:]`.

boundsj:
Squashed and rebased on top of https://github.com/mapbox/mapbox-gl-native/pull/7128/commits/0c518addd08e272ee9489bfdb4fddd1a52105a08 (https://github.com/mapbox/mapbox-gl-native/pull/7128)

Issue #7194, opened 2016-11-26
Tap gesture recognizer should fail unless annotation is nearby


1ec5:
MGLMapView’s built-in UITapGestureRecognizer should set its state to `UIGestureRecognizerStateFailed` as soon as it discovers that no annotations are nearby. Combined with #2278, the developer would be able to install their own “fallback” tap gesture recognizer that handles taps when the built-in gesture recognizer, meant for annotation selection, has nothing to select.

/cc @danpat @incanus

danpat:
I hit this.  I have annotations on the map, but I would also like to allow the user to tap elsewhere on the map and perform a feature lookup using `visibleFeaturesAtPoint:`.

1ec5:
Note that the recognizer should _not_ fail if an annotation is already selected and tapping closes the annotation’s callout view.

1ec5:
Fixed in #7246.

Issue #2278, opened 2015-09-08
Expose MGLMapView’s internal gesture recognizers publicly


maciekish:
Hi,
In the old MapBox there were two delegate methods that are missing in MapBox-gl. Namely `- (void)singleTapOnMap:(RMMapView *)map at:(CGPoint)point` and `- (void)longPressOnMap:(RMMapView *)map at:(CGPoint)point`. We used the first to cancel overlays and the second to drop a pin for geocoding.

Is there any plan for adding these to MapBox-gl please? I can see there are already single and long tap recognizers in `MGLMapView.mm` but there is no delegate. If you don't have time, will you perhaps take a PR for this?

We can't do this in our app code because for example we don't want the singleTapOnMap method to fire when the user taps an annotation, only when touching the map itself.

maciekish:
Here goes nothing: #2280

1ec5:
> we don't want the singleTapOnMap method to fire when the user taps an annotation, only when touching the map itself.

That’s a reasonable consideration. Perhaps built-in gesture recognizers should be exposed publicly, so you can implement your own gesture recognizer and have it depend on the built-in one failing.

f2m2rd:
+1

kristophM:
+1

1ec5:
As explained in https://github.com/mapbox/mapbox-gl-native/pull/2280#issuecomment-150007468, we should expose the gesture recognizers that MGLMapView uses internally – or at least the tap gesture recognizer – as read-only public properties. That would allow client code to implement `-gestureRecognizer:shouldRecognizeSimultaneouslyWithGestureRecognizer:` more intelligently.

To be clear, it is already possible to install your own tap and long press gesture recognizers, but by default they always block the tap gesture recognizer built into MGLMapView. It’s even possible already to pass through certain gestures by implementing `-gestureRecognizer:shouldRecognizeSimultaneouslyWithGestureRecognizer:`. But to have one gesture recognizer handle annotation selection and the other handle all other taps, you probably would need the built-in recognizer to be exposed publicly.

1ec5:
By the way, here’s a workaround:

```swift
mapView.gestureRecognizers.filter { $0 is UITapGestureRecognizer }.first
```

1ec5:
#7194 would make the built-in tap gesture recognizer fail whenever it ends up doing nothing, so that another gesture recognizer in application space can take over.

1ec5:
[...]

PR #7229, opened 2016-11-29
[ios, macos] Add new method to MGLMapViewDelegate for tapOnMap

JesseCrocker:
This adds a new method to `MGLMapViewDelegate` `mapView:tapAtCoordinate:` that is called when the map is tapped, but an annotation was not selected.


1ec5:
Would this new delete method still be necessary once we implement #7194? See #2278 for more discussion.

JesseCrocker:
This would not be necessary with #7194 and #2278

PR #7237, opened 2016-11-30
[android] telemetry - horizontal accuracy

### LINKED ISSUE #7235: Add horizontal accuracy to telemetry events
Add horizontal accuracy to location events.

ivovandongen:
Fixes #7235

Review @tobrun

### DIFF
--- a/platform/android/MapboxGLAndroidSDK/src/main/java/com/mapbox/mapboxsdk/telemetry/MapboxEventManager.java
+++ b/platform/android/MapboxGLAndroidSDK/src/main/java/com/mapbox/mapboxsdk/telemetry/MapboxEventManager.java
@@ -330,16 +330,17 @@ public class MapboxEventManager {
 
     /**
      * Adds a Location Event to the system for processing
+     *
      * @param location Location event
      */
     public void addLocationEvent(Location location) {
 
         // NaN and Infinite checks to prevent JSON errors at send to server time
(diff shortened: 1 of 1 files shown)


ivovandongen:
@aaronlidman @l-r Should be sufficient right? I haven't put in any rounding for now.

l-r:
Cool @ivovandongen. Are we going to have some logic to filter out certain values as per https://github.com/mapbox/telemetry/issues/92#issuecomment-263802942? I'm on the fence for this one.

ivovandongen:
> Are we going to have some logic to filter out certain values as per mapbox/telemetry#92 (comment)? I'm on the fence for this one.

Up to you. We can add a 0 check if you don't want those location updates.

aaronlidman:
Yeah
> If this location does not have an accuracy, then 0.0 is returned

Can we only include > 0 for Android?

l-r:
As per https://github.com/mapbox/telemetry/issues/92#issuecomment-263802942, if we can add rounding in there I think this is good to go.

aaronlidman:
> Can we only include > 0 for Android?

Ignore my previous, I've been convinced otherwise.

Sounds like rounding is the only hold up.

ivovandongen:
> If we can add rounding in there I think this is good to go.

https://github.com/mapbox/mapbox-gl-native/blob/fa96c422168c0d5006f81bba5f0c25ed4da2438c/platform/android/MapboxGLAndroidSDK/src/main/java/com/mapbox/mapboxsdk/telemetry/MapboxEventManager.java#L356

PR #7242, opened 2016-11-30
Relayout after recalculating to ensure visibility changes are handled correctly

### LINKED ISSUE #6758: Dynamic visibility layer property not applied until after moving / scaling map
**Platform:** Android
**Mapbox SDK version:** 4.2.0-beta.3

### Steps to trigger behavior

```java
Layer layer = map.getLayer("layerId");
layer.setProperties(PropertyFactory.visibility(Property.VISIBLE));
```

### Notes

- Tried calling invalidate() on MapboxMap and MapView, no effect.
- Layer changes are being applied on ApplicationContext main loop via a handler.post().
- A Java Time Thread periodically calls handler.post(), so no map interaction is required for the property change.
- There are many layers being hidden and shown. We're flipping fill layers to do a radar animation loop.
- Problem does not occur if I just change opacity using 1f and 0f values instead of VISIBLE or NONE.

### Expected behavior

Layers are visible after changing the visibility property to VISIBLE.

### Actual behavior

Layers only become visible after changing the visibility property to VISIBLE **AND** moving the map (drag / zoom).

### LINKED ISSUE #7225: Circle layer visibility not showing on all tiles
Similar to the issue previously seen when changing map languages and symbol layers, the visibility of a layer sometimes doesn't immediately update the map. This was produced on both master and the release branch.

ezgif com-video-to-gif
_in the gif clicking the map is suppose to toggle the visibility of the layer_

cc: @ivovandongen

### LINKED ISSUE #7241: Source tiles not reloaded when making associated layer visible
In the case where a source has no visible layers and the visibility of one of the layers is set to `VISIBLE`, the source tiles are not reloaded because the source is disabled when [Style::relayout()](https://github.com/mapbox/mapbox-gl-native/blob/master/src/mbgl/style/style.cpp#L252) is called:

```
        if (source && source->baseImpl->enabled) {
            source->baseImpl->reloadTiles();
        }
```

The source is only thereafter enabled in [Style::recalculate()](https://github.com/mapbox/mapbox-gl-native/blob/master/src/mbgl/style/style.cpp#L311). This was added to prevent reloading sources that aren't used: https://github.com/mapbox/mapbox-gl-native/commit/91272992f3a7ca7baa124c3378f1c1e2b3e5bb76

This results in layers not appearing visibility is set https://github.com/mapbox/mapbox-gl-native/issues/7225.

ivovandongen:
Fixes #7241
Fixes #7225
Fixes #6758

Re-ordering `relayout()` after `recalculate()` ensures that sources are enabled before checking if tiles need to be reloaded on them.

### DIFF
--- a/package.json
+++ b/package.json
@@ -24,7 +24,7 @@
     "lodash": "^4.16.4",
     "mapbox-gl-shaders": "mapbox/mapbox-gl-shaders#b51b85ffb8c512e228c36c5405293ce51d123519",
     "mapbox-gl-style-spec": "mapbox/mapbox-gl-style-spec#49e8b407bdbbe6f7c92dbcb56d3d51f425fc2653",
-    "mapbox-gl-test-suite": "mapbox/mapbox-gl-test-suite#9252ffc5108131704b5acf52d78258ac05687871",
+    "mapbox-gl-test-suite": "mapbox/mapbox-gl-test-suite#66397c8ee0426537d25e8d4cfcac134ca21a021b",
     "mkdirp": "^0.5.1",
     "node-cmake": "^1.2.1",
     "request": "^2.72.0",
--- a/platform/node/test/suite_implementation.js
+++ b/platform/node/test/suite_implementation.js
@@ -61,14 +61,9 @@ module.exports = function (style, options, callback) {
             callback();
 
         } else if (operation[0] === 'wait') {
-            var wait = function () {
-                if (map.loaded()) {
-                    applyOperations(operations.slice(1), callback);
-                } else {
-                    map.render(options, wait);
-                }
-            };
-            wait();
+            map.render(options, function () {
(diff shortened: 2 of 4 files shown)


ivovandongen:
@kkaefer @brunoabinader this pr fixes the visibility issue without revering the optimisation in https://github.com/mapbox/mapbox-gl-native/commit/91272992f3a7ca7baa124c3378f1c1e2b3e5bb76

A review would be welcome so we can cherry pick it into the android 4.2.0 release

1ec5:
This also needs to be cherry-picked to the iOS release branch.

/cc @boundsj

ivovandongen:
@jfirebaugh I fixed the unit test. Not sure how to implement the same thing in a regression test: https://github.com/mapbox/mapbox-gl-test-suite/pull/188

PR #7246, opened 2016-11-30
Make tap gesture recognizer fail if it doesn’t do anything

JesseCrocker:
I couldn't tell from the discussion on #2278 and #7194 if implementation of this was already in progress or not, but i needed this functionality to be able to continue with a progress.  If this is already is progress, please ignore this PR.

### DIFF
--- a/platform/ios/src/MGLMapView.h
+++ b/platform/ios/src/MGLMapView.h
@@ -409,7 +409,7 @@ IB_DESIGNABLE
 @property(nonatomic) CGFloat decelerationRate;
 
 /**
- The festure recognizer used to detect single taps on the map.
+ The gesture recognizer used to detect single taps on the map.
  
  Your application can use this to modify the behavior of your own gesture recognizers.
  */
--- a/platform/ios/src/MGLMapView.h
+++ b/platform/ios/src/MGLMapView.h
@@ -408,6 +408,13 @@ IB_DESIGNABLE
  */
 @property(nonatomic) CGFloat decelerationRate;
 
+/**
+ The festure recognizer used to detect single taps on the map.
+ 
+ Your application can use this to modify the behavior of your own gesture recognizers.
+ */
+@property (nonatomic, readonly) UITapGestureRecognizer *singleTapGestureRecognizer;
+
 #pragma mark Manipulating the Viewpoint
 
 /**
--- a/platform/ios/src/MGLMapView.h
+++ b/platform/ios/src/MGLMapView.h
@@ -408,13 +408,6 @@ IB_DESIGNABLE
  */
 @property(nonatomic) CGFloat decelerationRate;
 
-/**
- The gesture recognizer used to detect single taps on the map.
- 
- Your application can use this to modify the behavior of your own gesture recognizers.
- */
-@property (nonatomic, readonly) UITapGestureRecognizer *singleTapGestureRecognizer;
-
 #pragma mark Manipulating the Viewpoint
 
 /**
--- a/platform/ios/src/MGLMapView.mm
+++ b/platform/ios/src/MGLMapView.mm
@@ -1418,7 +1418,7 @@ public:
  @param singleTap An in progress tap gesture recognizer.
  @param persist True to remember the cycleable set of annotations. @see annotationTagAtPoint:persistingResults
  */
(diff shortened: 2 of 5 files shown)


platform/ios/src/MGLMapView.mm
```diff
@@ -1425,6 +1426,27 @@ - (void)handleSingleTapGesture:(UITapGestureRecognizer *)singleTap
...
+
+/**
+ Returns the annotation that would be selected by a tap gesture recognizer.
+ 
+ This is used when a gesture is recognized, and to check if the gesture should be recognized.
+ 
+ @param singleTap An in progress tap gesture recognizer.
+ @param persist True to remember the cycleable set of annotations. @see annotationTagAtPoint:persistingResults
+ */
+- (id <MGLAnnotation> _Nullable)annotationForGestureRecognizer:(UITapGestureRecognizer*)singleTap persistingResults:(BOOL)persist
```
1ec5: Nit: use the `nullable` keyword.

platform/ios/src/MGLMapView.mm
```diff
@@ -1635,6 +1655,17 @@ - (BOOL)gestureRecognizerShouldBegin:(UIGestureRecognizer *)gestureRecognizer
             }
         }
     }
+    else if (gestureRecognizer == _singleTapGestureRecognizer)
+    {
+      //Gesture will be recognized if it could deselect an annotation
+      if(!self.selectedAnnotation)
```
1ec5: What if an annotation is selected and tapping would ordinarily deselect it? Wouldn't this prevent the single tap gesture recognizer from firing to deselect the annotation?

platform/ios/src/MGLMapView.h
```diff
@@ -390,6 +390,13 @@ IB_DESIGNABLE
  */
 @property(nonatomic) CGFloat decelerationRate;
 
+/**
+ The gesture recognizer used to detect single taps on the map.
+ 
+ Your application can use this to modify the behavior of your own gesture recognizers.
+ */
+@property (nonatomic, readonly) UITapGestureRecognizer *singleTapGestureRecognizer;
```
1ec5: Is it enough for the application's gesture recognizers to avoid conflicts with this one gesture recognizer, or is it also necessary to avoid conflicts with, say, the double-tap gesture recognizer? Would it make sense for the application's gesture recognizer to require all the map view gesture recognizers (via the `gestureRecognizers` property) to fail? Or would those other gesture recognizers need to defer to the application's single tap gesture recognizer?

platform/ios/src/MGLMapView.mm
JesseCrocker: If this is false, it returns YES from the last line of the method

platform/macos/CHANGELOG.md
```diff
@@ -57,6 +57,7 @@
 * Fixed an issue where the map view’s center would always be calculated as if the view occupied the entire window. ([#6102](https://github.com/mapbox/mapbox-gl-native/pull/6102))
 * Notification names and user info keys are now string enumeration values for ease of use in Swift. ([#6794](https://github.com/mapbox/mapbox-gl-native/pull/6794))
 * Fixed a typo in the documentation for the MGLCompassDirectionFormatter class. ([#5879](https://github.com/mapbox/mapbox-gl-native/pull/5879))
+* The NSClickGestureRecognizer on MGLMapView that is used for selecting annotations now fails if a tap does not select an annotation. ([#7246](https://github.com/mapbox/mapbox-gl-native/pull/7246))
```
1ec5: Nit: it’s a click, not a tap. 🖱

platform/ios/src/MGLMapView.h
1ec5: Yeah, let’s leave out the property for now. If someone really needs it, they can filter the array for UITapGestureRecognizers as a workaround. That leaves MGLMapView free to install additional gesture recognizers in the future for other purposes.

1ec5:
Thanks again for this PR – it’s looking solid. By the way, it’d be nice to get this feature into the v3.4.1 release. We’re developing the v3.4.x series on the release-ios-v3.4.0 branch. Would you mind retargeting this PR and rebasing this branch on release-ios-v3.4.0?

JesseCrocker:
Rebased, and re-targeted

PR #7222, opened 2016-11-30
Use system NDK for building library

kkaefer:
We're currently downloading packaging and downloading the NDK via mason, but that adds download/unpacking overhead to every CI build, and burdens us with having to update the NDK.

Instead, we should use the NDK and CMake files provided with the NDK installed on the system.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -9,12 +9,12 @@ endif
 ifeq ($(shell uname -s), Darwin)
   HOST_PLATFORM = macos
   HOST_PLATFORM_VERSION = $(shell uname -m)
-  NINJA ?= platform/macos/ninja
+  export NINJA = platform/macos/ninja
   export JOBS ?= $(shell sysctl -n hw.ncpu)
 else ifeq ($(shell uname -s), Linux)
   HOST_PLATFORM = linux
   HOST_PLATFORM_VERSION = $(shell uname -m)
-  NINJA ?= platform/linux/ninja
+  export NINJA = platform/linux/ninja
   export JOBS ?= $(shell grep --count processor /proc/cpuinfo)
 else
   $(error Cannot determine host platform)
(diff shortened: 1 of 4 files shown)


jfirebaugh:
Well, this burdens every user with manually installing the NDK and ensuring that it's the right version when we require an upgrade. I think I prefer installing it via mason.

kkaefer:
@jfirebaugh well... one we integrate CMake into Gradle/Android Studio, it'll only be able to use the build-in version of the NDK, so this gets us a step into that direction (full CMake integration is still pending because Gradle doesn't really have support for building individual architectures/buildtypes). Also, our full CI builds are currently timing out because they take longer than 45 minutes, and the idea is that using the built-in version will reduce that time by removing the need to download + unpack the individual NDKs.

That being said, I can see a need for automated NDK downloads as well. Given that users who build Android need the Android SDK installed anyway, we could script the SDK installer to also download/update the NDK. Alternatively, we could try to load the NDK and check the installed version, and if we can't find it, or it isn't sufficient, fall back to using Mason to install that part of the NDK.

kkaefer:
Maybe we can get someone from team @mapbox/android weigh in on the feasibility/desirability of this?

tobrun:
On top of what is mentioned above I think this could help out resolving the timeouts on our nightly builds in #7733. With a nightly we currently download the following packages:

```
linux-x86_64/android-ndk/arm-9-r13b.tar.gz...
linux-x86_64/android-ndk/arm64-21-r13b.tar.gz...
linux-x86_64/android-ndk/x86-9-r13b.tar.gz...
linux-x86_64/android-ndk/x86_64-21-r13b.tar.gz...
linux-x86_64/android-ndk/mips-9-r13b.tar.gz...
```

I haven't benchmarked the time it takes to download them but since their size is significant I'm confident it will resolve #7733. For our CI integration we don't need to download these as Bitrise has a [Android NDK docker image](https://github.com/bitrise-docker/android-ndk) with preinstalled packages.

zugaldia:
In my opinion 1) bringing us closer to full CMake integration and 2) streamlining our builds to help with the CI timeout issues are two big 👍 to move forward with the approach proposed in this PR.

kkaefer:
This how supports both methods: It uses the system-provided NDK when it's available, but falls back to installing the arch-specific toolchain via Mason. @jfirebaugh can you please test whether this works in your NDK-less setup?

jfirebaugh:
There's a bold red warning at the beginning of the build:

```
Can't find the Android NDK.
Please use the Android SDK Manager to reinstall the NDK bundle.
```

But ignoring that, it works great. 👍

tobrun:
This is great @kkaefer:

without installing it, it falls back to mason:
screen shot 2017-02-22 at 13 54 08

After adding the ndk to my path, it picks up the local one:
screen shot 2017-02-22 at 14 20 42

kkaefer:
Made the fallback message less threatening:

untitled 2017-02-22 16-50-35

PR #7248, opened 2016-11-30
[core] Functions must have at least one stop

jfirebaugh:
This is already enforced by the style validator in mapbox-gl-style-spec. Enforce it here too.

@1ec5 @frederoni Do we need to add an equivalent validation in the SDK bindings for `Function`?


frederoni:
Yes, the assertion in core would be quite hard to guess from the SDK side so I think we should fail fast and loud here. I'll make a ticket.

PR #7250, opened 2016-11-30
Migrate MGLCustomStyleLayerAdditions to style layer API

### LINKED ISSUE #5972: Convert MGLCustomStyleLayerAdditions into a class
Now that we have a formal concept of style layers in the iOS and macOS SDKs, we should convert the MGLCustomStyleLayerAdditions category methods on MGLMapView into a proper subclass of MGLStyleLayer. We can continue to leave this class undocumented, signaling that it’s subject to change and not intended for casual use. Depending on how things shake out, it might end up being easier to extend support for macOS than to exclude macOS from the new implementation.

/cc @frederoni @friedbunny @incanus

1ec5:
Replaced the custom style layer API on MGLMapView with an equally unsupported MGLOpenGLStyleLayer API that nonetheless is consistent with the broader runtime styling API and compatible with macOS. Reprised the demo removed from iosapp in #5091 inside macosapp.

Before this can land:

* [x] Replace the callback blocks with either overrideable instance methods or delegate methods
* [x] Return the MGLStyleLayer object owned by MGLStyle on demand

A reminder: undocumented APIs like MGLOpenGLStyleLayer are unsupported and may change at any time without notice.

Fixes #5972.

/cc @friedbunny @boundsj @jfirebaugh


platform/darwin/src/MGLOpenGLStyleLayer.mm
```diff
@@ -0,0 +1,200 @@
...
+ @param mapView The map view to which the layer draws.
+ @param context A context structure with information defining the frame to draw.
+ */
+- (void)drawInMapView:(MGLMapView *)mapView withContext:(MGLStyleLayerDrawingContext)context {
+    
+}
+
+/**
+ Forces the map view associated with this style to redraw the receiving layer,
+ causing its drawing handler to be run.
```
1ec5: > causing its drawing handler to be run

This is a holdover from the former block-based API. I reworded this sentence in 178130267e3d79e53506549fb43c4fb2ef18e6e0.

PR #7251, opened 2016-11-30
More ways to reshape an MGLMultiPoint

### LINKED ISSUE #6583: MGLMultiPoint (MGLPolyline, MGLPolygon) should allow removing coordinates
Now that #6565 has landed, it’s possible to replace a segment (series of coordinates) of an MGLMultiPoint, MGLPolyline, or MGLPolygon with a segment of equal length, append coordinates to the end of the shape, or replace a segment at the end while appending. However, it isn’t possible to truncate (shrink) the shape, splice out (delete) a segment in the beginning or middle of the shape, or replace all the coordinates without replacing the shape annotation outright (since we try to avoid `memcpy()` of overlapping memory).

To complete the feature we added in #6565, we need to add the following methods to MGLMultiPoint:
- `-setCoordinates:count:`
- `-replaceCoordinatesInRange:withCoordinates:count:` (and remove or deprecate `-replaceCoordinatesInRange:withCoordinates:`)
- `-insertCoordinates:count:atIndex:`
- `-removeCoordinatesInRange:`
- optionally `-appendCoordinates:count:`

These operations have tricky edge cases that make our manual use of C memory management particularly perilous: an off-by-one error can go undetected even with unit tests. So I think we should convert the internal representation to use a `std::vector` instead of a heap-allocated C array: #6580. Once we fix #5201, we’ll need to factor out the implementations of these methods into freestanding C functions to avoid repetition.

/ref https://github.com/mapbox/mapbox-gl-native/pull/6565#discussion_r81871217
/cc @incanus @friedbunny @boundsj @ryanbaumann

1ec5:
Added the complete set of methods for mutating the vertices of an MGLMultiPoint. Invalidate `overlayBounds` whenever the coordinates change, but don’t recompute the bounds until they’re requested. Previously, `-[MGLMultiPoint replaceCoordinatesInRange:withCoordinates:]` failed to recompute the bounds.

Rewrote MGLMultiPoint documentation to refer to vertices instead of points. Removed a paragraph in MGLOverlay documentation that was full of references to features that exist in MKOverlay but not MGLOverlay.

Added a utility function for testing whether two `MGLCoordinateBounds`es intersect, based on `mbgl::LatLngBounds::intersects()`. Removed unused color conversion code.

Fixes #6583.

/cc @boundsj @incanus

### DIFF
--- a/platform/darwin/src/MGLPointCollection.mm
+++ b/platform/darwin/src/MGLPointCollection.mm
@@ -8,9 +8,11 @@ NS_ASSUME_NONNULL_BEGIN
 
 @implementation MGLPointCollection
 {
-    MGLCoordinateBounds _bounds;
+    MGLCoordinateBounds _overlayBounds;
     std::vector<CLLocationCoordinate2D> _coordinates;
 }
+
+@synthesize overlayBounds = _overlayBounds;
 
 + (instancetype)pointCollectionWithCoordinates:(const CLLocationCoordinate2D *)coords count:(NSUInteger)count
 {
@@ -28,7 +30,7 @@ NS_ASSUME_NONNULL_BEGIN
         {
             bounds.extend(mbgl::LatLng(coordinate.latitude, coordinate.longitude));
         }
-        _bounds = MGLCoordinateBoundsFromLatLngBounds(bounds);
+        _overlayBounds = MGLCoordinateBoundsFromLatLngBounds(bounds);
     }
     return self;
 }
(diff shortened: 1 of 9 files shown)


platform/darwin/src/MGLPointCollection.mm
```diff
@@ -68,7 +68,7 @@ - (MGLCoordinateBounds)overlayBounds
 
 - (BOOL)intersectsOverlayBounds:(MGLCoordinateBounds)overlayBounds
 {
-    return MGLLatLngBoundsFromCoordinateBounds(_bounds).intersects(MGLLatLngBoundsFromCoordinateBounds(overlayBounds));
+    return MGLCoordinateBoundsIntersectsCoordinateBounds(_bounds, overlayBounds);
```
boundsj: This approach could also be used in `MGLMultiPolyline(gon) intersectsOverlayBounds:`. In general, I wonder if it'd ever be useful to add an `intersectsOverlay: <MGLOverlay>` API for all of these objects that conform to `MGLOverlay`?

1ec5: > In general, I wonder if it'd ever be useful to add an `intersectsOverlay: <MGLOverlay>` API for all of these objects that conform to `MGLOverlay`?

This SDK isn’t a general-purpose geometry library, so I think we should only add geometric methods as we need them or if we need them for parity with MapKit. In the future, we may want to spin out these geometry classes as part of a separate library that would serve as a more natural home for such geometry functions.

1ec5: I’ve updated MGLMultiPolygon and MGLMultiPolyline to use the simplified expression too.

PR #7257, opened 2016-12-01
Check for duplicate layer IDs

### LINKED ISSUE #7254: Throw exception when adding layer with duplicate ID
Similar to #7011, we should throw an exception when adding a layer to the style whose ID is already in use by another layer in the style. The style specification describes the [`id`](https://www.mapbox.com/mapbox-gl-style-spec/#layer-id) property on a layer as “unique”, and `mbgl::Map::getLayer()` relies on that fact.

Alternatively, it would be possible to silently replace the existing layer that has a matching ID. But I’d prefer that we make the developer explicitly replace the layer, since reusing an ID is most likely an unintentional error.

/cc @ivovandongen @jfirebaugh

ivovandongen:
Fixes #7254

### DIFF
--- a/platform/darwin/src/MGLStyle.mm
+++ b/platform/darwin/src/MGLStyle.mm
@@ -271,10 +271,18 @@ static NSURL *MGLStyleURL_emerald;
         [NSException raise:NSRangeException
                     format:@"Cannot insert style layer at out-of-bounds index %lu.", (unsigned long)index];
     } else if (index == 0) {
-        [styleLayer addToMapView:self.mapView];
+        try {
+            [styleLayer addToMapView:self.mapView];
+        } catch (const std::runtime_error & err) {
+            [NSException raise:@"MGLRedundantLayerIdentifierException" format:@"%s", err.what()];
+        }
     } else {
-        MGLStyleLayer *sibling = [self layerFromMBGLLayer:layers.at(layers.size() - index)];
-        [styleLayer addToMapView:self.mapView belowLayer:sibling];
+        try {
+            MGLStyleLayer *sibling = [self layerFromMBGLLayer:layers.at(layers.size() - index)];
+            [styleLayer addToMapView:self.mapView belowLayer:sibling];
+        } catch (std::runtime_error & err) {
+            [NSException raise:@"MGLRedundantLayerIdentifierException" format:@"%s", err.what()];
+        }
     }
 }
 
@@ -350,7 +358,11 @@ static NSURL *MGLStyleURL_emerald;
          layer];
     }
     [self willChangeValueForKey:@"layers"];
-    [layer addToMapView:self.mapView];
+    try {
+        [layer addToMapView:self.mapView];
+    } catch (std::runtime_error & err) {
+        [NSException raise:@"MGLRedundantLayerIdentifierException" format:@"%s", err.what()];
+    }
     [self didChangeValueForKey:@"layers"];
 }
 
@@ -375,7 +387,11 @@ static NSURL *MGLStyleURL_emerald;
          sibling];
     }
     [self willChangeValueForKey:@"layers"];
-    [layer addToMapView:self.mapView belowLayer:sibling];
+    try {
+        [layer addToMapView:self.mapView belowLayer:sibling];
+    } catch (std::runtime_error & err) {
+        [NSException raise:@"MGLRedundantLayerIdentifierException" format:@"%s", err.what()];
+    }
     [self didChangeValueForKey:@"layers"];
 }
 
@@ -413,10 +429,18 @@ static NSURL *MGLStyleURL_emerald;
          @"Make sure sibling was obtained using -[MGLStyle layerWithIdentifier:].",
          sibling];
     } else if (index + 1 == layers.size()) {
-        [layer addToMapView:self.mapView];
+        try {
+            [layer addToMapView:self.mapView];
+        } catch (std::runtime_error & err) {
+            [NSException raise:@"MGLRedundantLayerIdentifierException" format:@"%s", err.what()];
+        }
     } else {
         MGLStyleLayer *sibling = [self layerFromMBGLLayer:layers.at(index + 1)];
-        [layer addToMapView:self.mapView belowLayer:sibling];
+        try {
+            [layer addToMapView:self.mapView belowLayer:sibling];
+        } catch (std::runtime_error & err) {
+            [NSException raise:@"MGLRedundantLayerIdentifierException" format:@"%s", err.what()];
+        }
     }
     [self didChangeValueForKey:@"layers"];
 }
--- a/platform/darwin/test/MGLStyleTests.mm
+++ b/platform/darwin/test/MGLStyleTests.mm
@@ -1,5 +1,8 @@
 #import "MGLMapView.h"
 #import "MGLStyle_Private.h"
+#if TARGET_OS_IPHONE
+#import "MGLMapView+MGLCustomStyleLayerAdditions.h"
+#endif
 
 #import "MGLGeoJSONSource.h"
 #import "MGLRasterSource.h"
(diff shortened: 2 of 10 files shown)


ivovandongen:
cc @1ec5 @jfirebaugh

platform/darwin/src/MGLStyle.mm
```diff
@@ -190,7 +190,11 @@ - (void)addLayer:(MGLStyleLayer *)layer
          layer];
     }
 
-    self.mapView.mbglMap->addLayer(std::unique_ptr<mbgl::style::Layer>(layer.layer));
+    try {
+        self.mapView.mbglMap->addLayer(std::unique_ptr<mbgl::style::Layer>(layer.layer));
```
ivovandongen: Added (tests) for `MGLStyle#insertLayer` and and `MGLMapView#insertCustomStyleLayerWithIdentifier`.

platform/ios/test/MGLLayerTests.m
```diff
@@ -0,0 +1,25 @@
+#import "MGLMapViewTests.h"
```
ivovandongen: Moved the files and added to macos as well. Excluded the custom layer tests for macos as the header is ios specific. Also changed the source test according to https://github.com/mapbox/mapbox-gl-native/pull/7203#discussion_r90384300

1ec5: That’s fine; the custom style layer API is getting folded into the normal style layer API in #7250, so soon there won’t be anything extra to test.

platform/darwin/test/MGLLayerTests.m
```diff
@@ -0,0 +1,43 @@
...
+    
+    //Try to insert the duplicate
+    XCTAssertThrowsSpecific([self.mapView.style insertLayer:layer3 belowLayer:layer1], NSException);
+    
+    #if TARGET_OS_IPHONE
+    //Try to insert a duplicate custom layer
+    MGLCustomStyleLayerPreparationHandler preparationHandler = ^{};
+    MGLCustomStyleLayerDrawingHandler drawingHandler = ^(CGSize size, CLLocationCoordinate2D centerCoordinate, double zoomLevel, CLLocationDirection direction, CGFloat pitch, CGFloat perspectiveSkew) {
+    };
+    MGLCustomStyleLayerCompletionHandler completionHander = ^{};
```
1ec5: FYI, feel free to inline these blocks into the method call. Long lines are par for the course in Objective-C. 😉

1ec5:
The iOS test failure appears to be what I fixed in #7228 on the release-ios-v3.4.0 branch.

ivovandongen:
> The iOS test failure appears to be what I fixed in #7228 on the release-ios-v3.4.0 branch.

@1ec5 Could you merge this into master so I can rebase this pr on it?

platform/darwin/src/MGLStyle.mm
```diff
@@ -211,7 +215,11 @@ - (void)insertLayer:(MGLStyleLayer *)layer belowLayer:(MGLStyleLayer *)otherLaye
     }
 
     const mbgl::optional<std::string> belowLayerId{otherLayer.identifier.UTF8String};
-    self.mapView.mbglMap->addLayer(std::unique_ptr<mbgl::style::Layer>(layer.layer), belowLayerId);
+    try {
+        self.mapView.mbglMap->addLayer(std::unique_ptr<mbgl::style::Layer>(layer.layer), belowLayerId);
```
1ec5: Now that #7282 has landed, including the changes in #6097, you’ll need to wrap any call to `-[MGLStyleLayer addToMapView:]` in this try-catch block or alternatively move this try-catch block to every implementation of `-addToMapView:` in the concrete subclasses of MGLStyleLayer.

ivovandongen:
@1ec5 I've rebased and updated the ios handling/tests: https://github.com/mapbox/mapbox-gl-native/pull/7257/commits/c005962117b3dda2ed663b482625c5ae62fa6333

platform/ios/src/MGLMapView.mm
```diff
@@ -5353,9 +5353,13 @@ - (void)insertCustomStyleLayerWithIdentifier:(NSString *)identifier preparationH
...
     MGLCustomStyleLayerHandlers *context = new MGLCustomStyleLayerHandlers(preparation, drawing, completion);
-    _mbglMap->addLayer(std::make_unique<mbgl::style::CustomLayer>(identifier.UTF8String, MGLPrepareCustomStyleLayer,
-                                                                  MGLDrawCustomStyleLayer, MGLFinishCustomStyleLayer, context),
-                       otherIdentifier ? mbgl::optional<std::string>(otherIdentifier.UTF8String) : mbgl::optional<std::string>());
+    try {
+        _mbglMap->addLayer(std::make_unique<mbgl::style::CustomLayer>(identifier.UTF8String, MGLPrepareCustomStyleLayer,
+                                                                      MGLDrawCustomStyleLayer, MGLFinishCustomStyleLayer, context),
+                           otherIdentifier ? mbgl::optional<std::string>(otherIdentifier.UTF8String) : mbgl::optional<std::string>());
+    } catch (std::runtime_error & err) {
+        [NSException raise:@"MGLRedundantLayerIdentiferException" format:@"%s", err.what()];
```
1ec5: I just realized we’ve been saying “identifer” instead of “identifier”, both for layers and sources. I can fix that typo in a separate PR, since we also need to add a note to the documentation for the affected methods about `MGLRedundantLayerIdentifierException` ([prior art involving sources](https://github.com/mapbox/mapbox-gl-native/blob/93b5ea060237de22b347527f3dc95e6f7cdb7251/platform/darwin/src/MGLStyle.h#L204-L206)).

ivovandongen:
@kkaefer I've added the backend guard in https://github.com/mapbox/mapbox-gl-native/pull/7257/commits/3f0c6c829443cec27764f025bcd87c6c3121a01b as you suggested and made `activate`/`deactivate` protected.

platform/ios/src/MGLMapView.mm
1ec5: If you’re in a good spot to do it, can you rename `MGLRedundantLayerIdentiferException` to `MGLRedundantLayerIdentiferException`? I don’t think we have any instances of it in the release branch, so this is a good branch to do it in. I can take care of `MGLRedundantSourceIdentiferException` separately: #7314.

platform/darwin/src/MGLStyle.mm
```diff
@@ -350,7 +358,11 @@ - (void)addLayer:(MGLStyleLayer *)layer
          layer];
     }
     [self willChangeValueForKey:@"layers"];
-    [layer addToMapView:self.mapView];
+    try {
+        [layer addToMapView:self.mapView];
+    } catch (std::runtime_error & err) {
```
kkaefer: We're catching a generic exception here, but always forward a `MGLRedundantLayerIdentifierException`. What we should do long term is to convert all of the generic `std::runtime_error` into individual exception types, and to exception-specific catches that we can then forward.

src/mbgl/style/style.cpp
```diff
@@ -198,6 +198,16 @@ Layer* Style::getLayer(const std::string& id) const {
 Layer* Style::addLayer(std::unique_ptr<Layer> layer, optional<std::string> before) {
     // TODO: verify source
 
+    //Guard against duplicate layer ids
```
kkaefer: Code style is to have a space after `//`.

src/mbgl/style/style.cpp
```diff
@@ -198,6 +198,16 @@ Layer* Style::getLayer(const std::string& id) const {
...
+    auto it = std::find_if(layers.begin(), layers.end(), [&](const auto& existing) {
+        return existing->getID() == layer->getID();
+    });
+
+    if (it != layers.end()) {
+        std::string msg = "Layer " + layer->getID() + " already exists";
+        throw std::runtime_error(msg.c_str());
```
kkaefer: no need to to `c_str()` here; `std::runtime_error` also accepts `std::string`. You could change this to something like this;

```c++
throw std::runtime_error(std::string{"Layer "} + layer->getID() + " already exists");
```

test/style/style_layer.test.cpp
```diff
@@ -268,3 +274,24 @@ TEST(Layer, Observer) {
...
+
+    //Add initial layer
+    style.addLayer(std::make_unique<LineLayer>("line", "unusedsource"));
+
+    //Try to add duplicate
+    try {
+        style.addLayer(std::make_unique<LineLayer>("line", "unusedsource"));
+        FAIL() << "Should not have been allowed to add a duplicate layer id";
+    } catch (std::runtime_error) {
+        //Expected
```
kkaefer: Check message here

test/style/style_layer.test.cpp
```diff
@@ -268,3 +274,24 @@ TEST(Layer, Observer) {
...
+    style.setJSON(util::read_file("test/fixtures/resources/style-unused-sources.json"));
+
+    //Add initial layer
+    style.addLayer(std::make_unique<LineLayer>("line", "unusedsource"));
+
+    //Try to add duplicate
+    try {
+        style.addLayer(std::make_unique<LineLayer>("line", "unusedsource"));
+        FAIL() << "Should not have been allowed to add a duplicate layer id";
+    } catch (std::runtime_error) {
```
kkaefer: The rule for exceptions in C++ is to throw by value and catch a `const` reference

platform/darwin/src/MGLStyle.mm
```diff
@@ -271,10 +271,18 @@ - (void)insertObject:(MGLStyleLayer *)styleLayer inLayersAtIndex:(NSUInteger)ind
         [NSException raise:NSRangeException
                     format:@"Cannot insert style layer at out-of-bounds index %lu.", (unsigned long)index];
     } else if (index == 0) {
-        [styleLayer addToMapView:self.mapView];
+        try {
+            [styleLayer addToMapView:self.mapView];
+        } catch (std::runtime_error & err) {
```
kkaefer: `const`

platform/darwin/test/MGLStyleTests.mm
```diff
@@ -165,6 +168,30 @@ - (void)testAddingLayersTwice {
...
+    MGLFillStyleLayer *initial = [[MGLFillStyleLayer alloc] initWithIdentifier:@"my-layer" source:source];
+    [self.mapView.style addLayer:initial];
+    
+    //Try to add the duplicate
+    XCTAssertThrowsSpecificNamed([self.mapView.style addLayer:[[MGLFillStyleLayer alloc] initWithIdentifier:@"my-layer" source:source]], NSException, @"MGLRedundantLayerIdentifierException");
+    XCTAssertThrowsSpecificNamed([self.mapView.style insertLayer:[[MGLFillStyleLayer alloc] initWithIdentifier:@"my-layer" source:source] belowLayer:initial],NSException, @"MGLRedundantLayerIdentifierException");
+    XCTAssertThrowsSpecificNamed([self.mapView.style insertLayer:[[MGLFillStyleLayer alloc] initWithIdentifier:@"my-layer" source:source] aboveLayer:initial], NSException, @"MGLRedundantLayerIdentifierException");
+    XCTAssertThrowsSpecificNamed([self.mapView.style insertLayer:[[MGLFillStyleLayer alloc] initWithIdentifier:@"my-layer" source:source] atIndex:0], NSException, @"MGLRedundantLayerIdentifierException");
+
+    #if TARGET_OS_IPHONE
```
kkaefer: Macros are not indented

ivovandongen:
@kkaefer Thanks for the review, I've addressed your notes.

PR #7261, opened 2016-12-01
Upgrade Android NDK to r13b

kkaefer:
Reverts arm64 back to use the gold linker. We've previously using another linker due to [bugs in gold on arm64](https://github.com/mapbox/mason/pull/135).


ivovandongen:
Ran the core tests as well. All good.

PR #7262, opened 2016-12-01
Refactor Android CMake targets

kkaefer:
We're currently running strip operations for every invocation, and build the test cases always, which makes the Android nightly build [go over Bitrise's 45 minute limit](https://www.bitrise.io/build/1a579128dd97a6a4). We should refactor `make apackage` so that it doesn't build the `mbgl-test` library.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -493,17 +493,17 @@ build/android-$1/$(BUILDTYPE)/Makefile: build/android-$1/$(BUILDTYPE)/toolchain.
 
 .PHONY: android-test-lib-$1
 android-test-lib-$1: build/android-$1/$(BUILDTYPE)/Makefile
-	$(NINJA) $(NINJA_ARGS) -j$(JOBS) -C build/android-$1/$(BUILDTYPE) mbgl-test-stripped
+	$(NINJA) $(NINJA_ARGS) -j$(JOBS) -C build/android-$1/$(BUILDTYPE) mbgl-test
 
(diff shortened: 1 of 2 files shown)


kkaefer:
@ivovandongen Refactored the targets so that stripping + copying into the jni folder is done as a `POST_BUILD` action, rather than a separate target. That means ninja now just has the three targets `mapbox-gl`, `example-custom-layer`, and `mbgl-test`.

In addition, I changed `example-custom-layer` so that it dynamically links against `mapbox-gl`, reducing the .so size from ~600KB to ~35KB.

Makefile
```diff
@@ -493,17 +493,17 @@ build/android-$1/$(BUILDTYPE)/Makefile: build/android-$1/$(BUILDTYPE)/toolchain.
 
 .PHONY: android-test-lib-$1
 android-test-lib-$1: build/android-$1/$(BUILDTYPE)/Makefile
-	$(NINJA) $(NINJA_ARGS) -j$(JOBS) -C build/android-$1/$(BUILDTYPE) mbgl-test-stripped
+	$(NINJA) $(NINJA_ARGS) -j$(JOBS) -C build/android-$1/$(BUILDTYPE) mbgl-test example-custom-layer
```
ivovandongen: We don't need to build the `example-custom-layer` here right? The testapp is not used for the core tests

kkaefer: Good call. I'll remove it.

Issue #7271, opened 2016-12-02
MGLSource and MGLLayer should raise an NSException if they are added to the style more than once


boundsj:
When `MGLSource` and `MGLLayer` instances are added to the style, they lose ownership of their `std::unique_ptr<T>` after it is moved to the mbgl level. Subsequent attempts to add such instances will result in a C++ exception.

An `NSException` should be raised if the source and layer `addToMapView` methods are called more than once and the pointer is invalid. In addition, the documentation in `MGLStyle` for `addSource:` and `addLayer:` should warn developers to avoid adding the same instance twice and, for that matter, instances with the same identifier.

boundsj:
ref https://github.com/mapbox/mapbox-gl-native/issues/7258

boundsj:
Fixed in https://github.com/mapbox/mapbox-gl-native/pull/7272

Issue #7258, opened 2016-12-01
Standardize NSException names


1ec5:
Most of the NSExceptions we raise in the iOS and macOS SDKs are named with ad-hoc names such as “Value not handled”. Instead, we should either use one of the [predefined exception names](https://developer.apple.com/reference/foundation/1613022-foundation_constants/1660462-general_exception_names?language=objc) or define a string constant such as “MGLIdentifierCollisionException” in MGLTypes.h for use with these exceptions. This way, test code can more easily check for the exception with `XCTAssertThrowsSpecificNamed()`, and we could start adding notes about exception conditions in our method documentation.

/cc @incanus @frederoni

fabian-guerra:
[Predefined exception names](https://developer.apple.com/library/content/documentation/Cocoa/Conceptual/Exceptions/Concepts/PredefinedExceptions.html#//apple_ref/doc/uid/20000057-BCIGHECA) link update.

PR #7272, opened 2016-12-02
[ios, macos] Raise NSException if layer or source added more than once

boundsj:
Fixes https://github.com/mapbox/mapbox-gl-native/issues/7271

cc @1ec5 @frederoni


platform/darwin/src/MGLBackgroundStyleLayer.mm
```diff
@@ -35,6 +35,10 @@ - (instancetype)initWithIdentifier:(NSString *)identifier
 
 - (void)addToMapView:(MGLMapView *)mapView
 {
+    if (_pendingLayer == nullptr) {
+        [NSException raise:NSInternalInconsistencyException format:@"Adding the same instance to the style more than once is invalid."];
```
1ec5: `NSInternalInconsistencyException` is the exception name used for assertions. Either define a custom exception name like `MGLRedundantLayerException` or use `NSGenericException`. #7258 isn’t critical, so it’s fine to specify `MGLRedundantLayerException` as a string literal for now without defining a constant.

platform/darwin/src/MGLStyle.h
```diff
@@ -200,6 +200,9 @@ static const NSInteger MGLStyleDefaultVersion = 9;
 /**
  Adds a new source to the current style.
  
+ @note Adding the same source instance more than once or reusing source
```
1ec5: We could say which exception name gets raised under which circumstance, so the developer knows what to look for. For now, spelling out the exception name would only make sense for the exceptions we’ve given CamelCase names to.

platform/darwin/src/MGLBackgroundStyleLayer.mm
1ec5: If the developer is accidentally readding an instance to a style, they’ll want to know which one. We can describe the layer and style or map view here.

PR #7278, opened 2016-12-02
[android] - remove overlay SurfaceView examples

tobrun:
VideoViewActivity was added do demonstrate showing a Video on top of a Map. We recently noticed that this doesn't work for an Android emulator. Testing of showing a SurfaceView on a SurfaceView is already been done in DoubleMapActivity so I'm removing SurfaceViewMediaControlActivity as being obsolete.

Review @cammace


platform/android/MapboxGLAndroidSDKTestApp/src/main/AndroidManifest.xml
```diff
@@ -318,14 +318,6 @@
                 android:value="@string/category_fragment" />
         </activity>
         <activity
-            android:name=".activity.maplayout.VideoViewActivity"
-            android:description="@string/description_video_view"
```
cammace: These string can be removed correct?

tobrun: great catch

PR #7287, opened 2016-12-05
Update changelogs

1ec5:
Updated the iOS and macOS SDK changelogs with changes in master that haven’t been cherry-picked into the v3.4.x release branch.

/cc @boundsj


friedbunny:
Looks like this needs https://github.com/mapbox/mapbox-gl-native/pull/7290 for the node build. 🤔

PR #7292, opened 2016-12-05
[ios] Round horizontal accuracy

boundsj:


1ec5:
Pull in #7290 to fix the Node.js build failure.

Issue #6863, opened 2016-10-31
Investigate pixelated symbols in labels


mikemorris:
**Platform:** Ubuntu 16.04 with NVIDIA closed-source GPU driver version 367.27
**Mapbox SDK version:** master

### Steps to trigger behavior

When rendering tiles from `mapbox/streets-v8` in `MapMode::Still`, some label icons appear pixelated.

- London `mapbox/streets-v8/tiles/17/65502/43585@2x`
  - `center: [ -0.092010498046875, 51.50617819737356 ]`
- New York City `mapbox/streets-v8/tiles/17/38592/49273@2x`
  - `center: [ -74.00253295898438, 40.72748642299778 ]`
- Changing the decimal precision of the lat/lng centerpoint can affect the artifacts, sometimes fixing them if precision is trimmed enough. However, trimming centerpoint precision does not fix all instances of this artifact, and in some cases makes it worse. **Is there a potential float precision issue here and if so how should it be handled?**

**master**

![43585 2x 3](https://cloud.githubusercontent.com/assets/1149913/19866793/dd73f082-9f77-11e6-9371-1fca3bf04ebb.png)

**trimmed to 8 decimal places**

![43585 2x 2](https://cloud.githubusercontent.com/assets/1149913/19866763/c4091460-9f77-11e6-88ba-0a5650ab79c1.png)

- [Removing `GL_LINEAR` mode rendering](https://github.com/mapbox/mapbox-gl-native/pull/6790) changes the style of artifact, fixing some but introduces others (fuzziness visible in `RBBTS` icon).

![1f7d29b4-9f71-11e6-8fd7-717bf4c70bec](https://cloud.githubusercontent.com/assets/1149913/19866541/e33fea44-9f76-11e6-92b9-fd96d3c94c39.png)

### Expected behavior

Label icons are rendered without pixelation or fuzziness artifacts.

![bec71b9c-9f71-11e6-868b-69ffb17f86e5](https://cloud.githubusercontent.com/assets/1149913/19866517/d0e460fa-9f76-11e6-8707-8b592080c767.png)

### Actual behavior

Label icons are pixelated (`Villa Mosconi`, `Emmet's`, `The Dutch`).

![49273 2x](https://cloud.githubusercontent.com/assets/1149913/19866612/32fd3942-9f77-11e6-94fe-5ab759d82938.png)

/cc @jfirebaugh @brunoabinader @kkaefer @samanpwbb

[...]

PR #7302, opened 2016-12-06
Fix pixelated icons

kkaefer:
This is a fix for #6863, which seems very elusive. It's highly dependent on the actual coordinates, the GPU in use (I could only replicate it with my discrete nVidia GPU, but not with the Intel integrated GPU), but even then repeated renders produce different results. My assumption is that the floating point implementation in the nVidia GPU is behaves in an undefined way when doing non-interpolated texture lookups at fringe locations (.5). I tried creating a test case, but given it's highly random behavior, I don't think adding a test case makes much sense.

The fix itself is close to what @jfirebaugh described in #6863 in that we are rounding the texture lookup coordinates when we aren't using linear interpolation.

### DIFF
--- a/package.json
+++ b/package.json
@@ -22,7 +22,7 @@
     "ejs": "^2.4.1",
     "express": "^4.11.1",
     "lodash": "^4.16.4",
-    "mapbox-gl-shaders": "mapbox/mapbox-gl-shaders#b51b85ffb8c512e228c36c5405293ce51d123519",
+    "mapbox-gl-shaders": "mapbox/mapbox-gl-shaders#cfb36126be74a303c6a5fe640a17c5e0d79b2035",
     "mapbox-gl-style-spec": "mapbox/mapbox-gl-style-spec#49e8b407bdbbe6f7c92dbcb56d3d51f425fc2653",
     "mapbox-gl-test-suite": "mapbox/mapbox-gl-test-suite#9252ffc5108131704b5acf52d78258ac05687871",
     "mkdirp": "^0.5.1",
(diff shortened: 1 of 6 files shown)


kkaefer:
Shader diff is in https://github.com/mapbox/mapbox-gl-shaders/compare/line2...pixelated-icons

kkaefer:
Bad news; it doesn't seem to resolve this issue either.

This is what I get **before** the fix:

nVidia GeFore GT 650M | Intel Graphics HD 5000
------------ | -------------
out-nvidia-gt650m-noscreen | out-intel-hd4000

...and this is **after** the "fix":

nVidia GeFore GT 650M | Intel Graphics HD 5000
------------ | -------------
fix-nvidia-gt650m | fix-intel-hd4000

kkaefer:
[New fix](https://github.com/mapbox/mapbox-gl-shaders/commit/4d3f2d4867b22c861e9088dbbd4dff748a630381) works on both GPUs I could test and seems easier.

ansis:
> New fix works on both GPUs I could test and seems easier.

Previously values close whole pixel values (`_.0`) were problematic and this just shifts the problem case to (`_.5`). Why is this better?

kkaefer:
@ansis should've explained this in the original posting. Here's why this is different from just shifting the problem:

- Previously, our quads for rendering icons could start at any fractional coordinate. It seems like that some GPUs have trouble interpolating the resulting coordinates in a uniform manner and we are seeing them sometimes being rounded up, and sometimes rounded down. 
- Now, the vertex coordinates are rounded so that they are snapping to integer pixels, leading to a more uniform interpolation of the `v_pos` value. So overall, rounding now happens in the vertex shader rather than in the fragment shader at texture lookup time.

kkaefer:
Upon further inspection, I can still observe incorrectly rendered icons at some scales with some renderers.

jfirebaugh:
Fixed in https://github.com/mapbox/mapbox-gl-native/pull/9038.

PR #7303, opened 2016-12-06
Prevent older g++ versions from segfaulting

akikoskinen:
My system has g++ version 5.3.1 and it started segfaulting recently when I pulled a new version of the code. I found the culprit, the g++ bugreports and hopefully the correct versions of g++ where this has been fixed.

### DIFF
--- a/include/mbgl/util/color.hpp
+++ b/include/mbgl/util/color.hpp
@@ -13,6 +13,11 @@ public:
     constexpr Color() = default;
     constexpr Color(float r_, float g_, float b_, float a_)
         : r(r_), g(g_), b(b_), a(a_) {
+// Some g++ versions can't deal with assertions within a constexpr constructor: https://gcc.gnu.org/bugzilla/show_bug.cgi?id=65973
+#if !defined(__GNUC__) || \
+    (__GNUC__ == 5 && __GNUC_MINOR__ >= 4) || \
+    (__GNUC__ == 6 && __GNUC_MINOR__ >= 1) || \
+     __GNUC__ >= 7
         assert(r_ >= 0.0f);
         assert(r_ <= 1.0f);
         assert(g_ >= 0.0f);
@@ -21,6 +26,7 @@ public:
         assert(b_ <= 1.0f);
         assert(a_ >= 0.0f);
         assert(a_ <= 1.0f);
+#endif
     }
 
     float r = 0.0f;


jfirebaugh:
Thanks for the heads up about the g++ issue. I would rather increase our documented minimum version of GCC than introduce selective compilation workarounds (especially workarounds that involve disabling debug assertions).

Do you have a link to the relevant GCC bug report?

akikoskinen:
> I would rather increase our documented minimum version of GCC than introduce selective compilation workarounds (especially workarounds that involve disabling debug assertions).

Makes sense.

> Do you have a link to the relevant GCC bug report?

In the code comments as well as in the commit message.

akikoskinen:
> I would rather increase our documented minimum version of GCC than introduce selective compilation workarounds (especially workarounds that involve disabling debug assertions).

Do note however that the working GCC versions (5.4 and 6.1) are quite recent: 5.4 was released on the 3rd of June 2016 and 6.1 was released on the 27th of April 2016.

jfirebaugh:
As an alternative workaround, does it work to define a `constexpr` auxiliary function that performs the assertions, and call that from the constructor?

akikoskinen:
> As an alternative workaround, does it work to define a constexpr auxiliary function that performs the assertions, and call that from the constructor?

Nope, same problem. Compiler segfaults. I also tried with a `#defined` macro but still segfault.

akikoskinen:
Is there some ideas what version(s) of GCC shall be supported? I just saw e69e93d2ad6881b024d4f3a0aeaf8e45e29c1261 and according to the commit message it is for supporting GCC 4.9 (which AFAIK doesn't even support C++14 fully). So, what's going on? Does @tmpsantos have any comments?

tmpsantos:
@akikoskinen this is a working in progress branch I'm developing together with @brunoabinader to get Mapbox GL Native building on GCC 4.9. It currently builds and works just fine with that branch on GCC 4.9, as we rewrote part of the code that was not compiling to be GCC 4.9 compatible.

The reason why we started this work was because Qt 5.9 that will hopefully ship with Mapbox GL does not yet support GCC 5: https://codereview.qt-project.org/#/c/177175/

jfirebaugh:
Going to go with an alternative fix in #7588, but thank you for bringing this to our attention.

PR #7309, opened 2016-12-06
Prevent stopless functions

### LINKED ISSUE #7249: MGLStyleFunction must have at least one stop
Raise an exception if MGLStyleFunction is initialized without any stops.

ref #7248

frederoni:
Fixes #7249 
Raises an exception if an `MGLStyleFunction` is initialized with no stops.


platform/darwin/src/MGLStyleValue.mm
```diff
@@ -63,6 +63,7 @@ - (instancetype)init {
 
 - (instancetype)initWithBase:(CGFloat)base stops:(NSDictionary *)stops {
     if (self = [super init]) {
+        NSAssert(stops.allKeys.count, @"%@ need at least one stop.", [self class]);
```
1ec5: `stops.count` would be more efficient.

1ec5: `[self class]` resolves to `MGLConstantStyleValue` or `MGLStyleFunction`, but I think `self` would be more informative since it calls `-description`.

1ec5: For internal errors, use `NSAssert()`. For developer errors, raise an NSException and give it a name like `NSInvalidArgumentException` that the developer can identify the exception by.

platform/darwin/test/MGLFunctionTests.m
```diff
@@ -0,0 +1,13 @@
+#import <XCTest/XCTest.h>
+#import <Mapbox/Mapbox.h>
+
+@interface MGLFunctionTests : XCTestCase
+@end
+
+@implementation MGLFunctionTests
+
+- (void)testStoplessFunction {
+    XCTAssertThrows([MGLStyleValue<NSNumber *> valueWithStops:@{}], @"Stopless function should throw an exception");
```
1ec5: If you give the exception a name, you can use `XCTAssertThrowsSpecificNamed()` here.

platform/darwin/test/MGLFunctionTests.m
```diff
@@ -0,0 +1,13 @@
+#import <XCTest/XCTest.h>
+#import <Mapbox/Mapbox.h>
+
+@interface MGLFunctionTests : XCTestCase
```
1ec5: Let’s call this MGLStyleValueTests for consistency with the file that contains the implementation.

frederoni: MGLStyleValueTests is taken by a Swift class with the same name but I can't use that class to catch a specific exception name if the method itself doesn't throw the exception. Renamed it to MGLStyleValueObjCTests for now.

1ec5: Ah, I see. Alternatively, maybe we can call this class MGLStyleValueTests and turn the Swift class into an extension on MGLStyleValueTests. Not sure if XCTest will be happy with that or not.

platform/darwin/src/MGLStyleValue.mm
```diff
@@ -63,6 +63,10 @@ - (instancetype)init {
 
 - (instancetype)initWithBase:(CGFloat)base stops:(NSDictionary *)stops {
     if (self = [super init]) {
+        if (!stops.count)
+        {
+            [NSException raise:@"NSInvalidArgumentException" format:@"%@ requires at least one stop.", self];
```
1ec5: Use the constant, `NSInvalidArgumentException`.

platform/darwin/test/MGLStyleValueObjCTests.m
```diff
@@ -0,0 +1,13 @@
+#import <XCTest/XCTest.h>
+#import <Mapbox/Mapbox.h>
+
+@interface MGLStyleValueObjCTests : XCTestCase
+@end
+
+@implementation MGLStyleValueObjCTests
+
+- (void)testStoplessFunction {
+    XCTAssertThrowsSpecificNamed([MGLStyleValue<NSNumber *> valueWithStops:@{}], NSException, @"NSInvalidArgumentException", @"Stopless function should raise an exception");
```
1ec5: The constant `NSInvalidArgumentException` works here too.

PR #7310, opened 2016-12-06
Rename style spec properties

### LINKED ISSUE #6098: Rename some layer and source properties

frederoni:
Fixes #6098 
Renamed a couple of more style spec properties.

- line-dasharray => line-dash-pattern
- icon-size => icon-scale


1ec5:
9f9422deb808ff5f77aa640c8222207fa6456e04 fixes a compiler error in iosapp caused by a reference to `iconSize`. Unfortunately we don’t run iosapp on Bitrise to catch errors like this.

PR #7324, opened 2016-12-07
Bump mason so that Android uses gold linker for arm64

kkaefer:
Follow up to #7109 that bumps mason to include https://github.com/mapbox/mason/pull/291, which changes the linker for arm64 back to gold after https://code.google.com/p/android/issues/detail?id=204151 was resolved.


kkaefer:
(Background is that https://www.bitrise.io/build/a84753e6d71856f7 failed)

Issue #7281, opened 2016-12-02
Many visibility warnings building macOS SDK


1ec5:
As of 84038832e45c2445b43b39ccb2f9e29c8646dac0 on master, a clean build of the macOS SDK in Xcode 8.2 results in 136 warnings of the following form:

```
ld: warning: direct access in function 'mbgl::Tileset const& mapbox::util::variant<std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char> >, mbgl::Tileset>::get<mbgl::Tileset, (void*)0>() const' from file '/Users/mxn/hub/mapbox-gl-native/build/macos/Debug/libmbgl-core.a(offline_download.o)' to global weak symbol 'typeinfo for mapbox::util::bad_variant_access' from file '/Users/mxn/hub/mapbox-gl-native/build/macos/macos.build/Debug/dynamic.build/Objects-normal/x86_64/NSArray+MGLAdditions.o' means the weak symbol cannot be overridden at runtime. This was likely caused by different translation units being compiled with different visibility settings.
```

`GCC_SYMBOLS_PRIVATE_EXTERN` is `YES` for the dynamic target in macos.xcodeproj (whereas it’s `NO` for the dynamic and static targets in ios.xcodeproj), but setting it to `NO` doesn’t address the issue.

This issue doesn’t reproduce on the release-ios-v3.4.0 branch.

/cc @kkaefer

1ec5:
Actually, setting `GCC_SYMBOLS_PRIVATE_EXTERN` to `NO` does address the issue. It appears to have been set to `YES` in #7188.

kkaefer:
`GCC_SYMBOLS_PRIVATE_EXTERN` adds `-fvisibility=hidden` for static libraries/executables, but apparently doesn't do this for dynamic libraries.

1ec5:
What about the iOS dynamic framework? I haven’t noticed these warnings there.

kkaefer:
That's because we're not yet compiling iOS with `-fvisibility=hidden` for the core library. PR for iOS is at https://github.com/mapbox/mapbox-gl-native/pull/7131

PR #7326, opened 2016-12-07
Add -fvisibility=hidden to macOS dynamic framework and export required symbols

kkaefer:
Fix for https://github.com/mapbox/mapbox-gl-native/issues/7281

### DIFF
--- a/platform/darwin/src/MGLAccountManager.h
+++ b/platform/darwin/src/MGLAccountManager.h
@@ -1,4 +1,4 @@
-#import <Foundation/Foundation.h>
+#import "MGLTypes.h"
 
 NS_ASSUME_NONNULL_BEGIN
 
@@ -6,6 +6,7 @@ NS_ASSUME_NONNULL_BEGIN
  The MGLAccountManager object provides a global way to set a Mapbox API access
  token.
  */
+MGL_EXPORT
 @interface MGLAccountManager : NSObject
 
 #pragma mark Authorizing Access
--- a/platform/darwin/src/MGLGeometry_Private.h
+++ b/platform/darwin/src/MGLGeometry_Private.h
@@ -51,6 +51,7 @@ NS_INLINE mbgl::EdgeInsets MGLEdgeInsetsFromNSEdgeInsets(NSEdgeInsets insets) {
     @param latitude The latitude of the point at the center of the viewport.
     @param size The size of the viewport.
     @return An altitude measured in meters. */
+MGL_EXPORT
 CLLocationDistance MGLAltitudeForZoomLevel(double zoomLevel, CGFloat pitch, CLLocationDegrees latitude, CGSize size);
 
 /** Converts a camera altitude to a map zoom level.
@@ -60,4 +61,5 @@ CLLocationDistance MGLAltitudeForZoomLevel(double zoomLevel, CGFloat pitch, CLLo
     @param latitude The latitude of the point at the center of the viewport.
     @param size The size of the viewport.
     @return A zero-based zoom level. */
+MGL_EXPORT
 double MGLZoomLevelForAltitude(CLLocationDistance altitude, CGFloat pitch, CLLocationDegrees latitude, CGSize size);
--- a/platform/macos/DEVELOPING.md
+++ b/platform/macos/DEVELOPING.md
@@ -54,6 +54,7 @@ To add any Objective-C type, constant, or member to the iOS SDK’s public inter
 
 To add an Objective-C class, protocol, category, typedef, enumeration, or global constant to the macOS SDK’s public interface:
 
+1. Add the macro `MGL_EXPORT` prior to the declaration. To use this macro, include `MGLTypes.h`.
 1. _(Optional.)_ Add the type or constant’s name to the relevant category in the `custom_categories` section of [the jazzy configuration file](./jazzy.yml). This is required for classes and protocols and also recommended for any other type that is strongly associated with a particular class or protocol. If you leave out this step, the symbol will appear in an “Other” section in the generated HTML documentation’s table of contents.
 1. _(Optional.)_ If the symbol would also be publicly exposed in the iOS SDK, consult [the companion iOS document](../ios/DEVELOPING.md#making-a-type-or-constant-public) for further instructions.
 
--- a/platform/macos/src/MGLMapView.h
+++ b/platform/macos/src/MGLMapView.h
@@ -47,7 +47,7 @@ NS_ASSUME_NONNULL_BEGIN
  @note You are responsible for getting permission to use the map data and for
     ensuring that your use adheres to the relevant terms of use.
  */
-IB_DESIGNABLE
+MGL_EXPORT IB_DESIGNABLE
 @interface MGLMapView : NSView
 
 #pragma mark Creating Instances
--- a/platform/darwin/src/MGLAttributionInfo.h
+++ b/platform/darwin/src/MGLAttributionInfo.h
@@ -1,4 +1,3 @@
-#import <Foundation/Foundation.h>
 #import <CoreGraphics/CoreGraphics.h>
 #import <CoreLocation/CoreLocation.h>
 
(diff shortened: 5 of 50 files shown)


platform/macos/src/MGLMapView.h
```diff
@@ -48,6 +48,7 @@ NS_ASSUME_NONNULL_BEGIN
     ensuring that your use adheres to the relevant terms of use.
  */
 IB_DESIGNABLE
+MGL_EXTERN
```
1ec5: Does Interface Builder still recognize MGLMapView as a designable? IB’s Objective-C scanner unfortunately doesn’t know anything about macros.

platform/darwin/src/MGLAccountManager.h
```diff
@@ -1,11 +1,12 @@
-#import <Foundation/Foundation.h>
+#import "MGLTypes.h"
 
 NS_ASSUME_NONNULL_BEGIN
 
 /**
  The MGLAccountManager object provides a global way to set a Mapbox API access
  token.
  */
+MGL_EXTERN
```
1ec5: If I’m not mistaken, placing this macro on a separate line between the comment block and the signature breaks Quick Help, which only looks one line up to find a comment block. Would this macro work if we stick it above the comment block? We should also make sure jazzy – which like Quick Help also relies on Clang to parse out comment blocks – is OK with whatever we end up with.

platform/darwin/src/MGLGeometry_Private.h
```diff
@@ -51,6 +51,7 @@ NS_INLINE mbgl::EdgeInsets MGLEdgeInsetsFromNSEdgeInsets(NSEdgeInsets insets) {
     @param latitude The latitude of the point at the center of the viewport.
     @param size The size of the viewport.
     @return An altitude measured in meters. */
+MGL_EXTERN
```
1ec5: Do we really need to `extern` symbols in headers with private or project visibility? What about symbols declared in .m files, then?

platform/macos/src/MGLMapView.h
kkaefer: How do I find out whether IB does that? I'm not familiar with it.

platform/darwin/src/MGLGeometry_Private.h
kkaefer: Theoretically we shouldn't, since it should be all internal to the binary we're building, however the test target also includes this private header, and linking the lib to the test fails without exporting this symbol. Another way would be to not export this, and change the test so that it doesn't require this symbol, or to move this symbol to the public header.

platform/darwin/src/MGLAccountManager.h
frederoni: <kbd>⌥</kbd> + click on class, property or method.
w/ macro
screenshot 2016-12-08 12 14 51
w/o macro
screenshot 2016-12-08 12 16 23

kkaefer: @frederoni I cannot confirm that behavior; it works for me with the macro as well.

frederoni: nvm @kkaefer I was using an invalid macro. 👍

platform/macos/src/MGLMapView.h
1ec5: Open MapDocument.xib. Xcode will automatically start building. After awhile, this:

blank

will turn into this:

designable

It’s a bit underwhelming on macOS, but the iOS SDK version of this displays some important setup instructions, so we should make sure designables in general still work.

platform/darwin/src/MGLGeometry_Private.h
1ec5: Ah, OK, so we need this on anything with Private visibility but not Project visibility (using Xcode terminology here).

platform/macos/src/MGLMapView.h
kkaefer: I tried this, but couldn't get to a conclusive answer since Xcode's rebuilding is very erratic and it doesn't seem to detect changed files. From what I could tell, it worked both in the `MGL_EXTERN IB_DESIGNABLE` order and the other way around. I changed the order though so that `IB_DESIGNABLE` now is directly before `@interface`.

platform/macos/DEVELOPING.md
```diff
@@ -48,6 +48,7 @@ To add any Objective-C type, constant, or member to the iOS SDK’s public inter
 
 1. Ensure that the symbol is pure Objective-C and does not rely on any language features specific to Objective-C++ or the C11 dialect of C – so no namespaced types or classes named with emoji! 🙃 Most projects that depend on this SDK are either written in pure Objective-C (GNU99 dialect) or Swift, which cannot yet bridge C++ types.
 1. Name the symbol according to [Cocoa naming conventions](https://developer.apple.com/library/prerelease/content/documentation/Cocoa/Conceptual/CodingGuidelines/CodingGuidelines.html#//apple_ref/doc/uid/10000146i). Use the `MGL` class prefix to avoid conflicts with client code. If the symbol has an analogue in MapKit, name the symbol according to MapKit.
+1. Add the macro `MGL_EXTERN` prior to the declaration. To get this macro, include `MGLTypes.h`.
```
1ec5: My mistake, this belongs in the next section (which could mention standalone functions as well). This section is for adding anything to a public header, including a property or method.

platform/macos/src/MGLMapView.h
```diff
@@ -47,6 +47,7 @@ NS_ASSUME_NONNULL_BEGIN
  @note You are responsible for getting permission to use the map data and for
     ensuring that your use adheres to the relevant terms of use.
  */
+MGL_EXTERN
```
1ec5: Putting `MGL_EXTERN` keyword on a separate line before the `IB_DESIGNABLE` keyword seems to break Quick Help and the index in general in Xcode 8.2. If you ⌥-click a member of a class or open the code completion popup, you get the documentation about the class itself rather than the member. If you ⌘-click to jump to the definition of a member, you get the line above it instead.

release-ios-v3.4.0 [on this line](https://github.com/mapbox/mapbox-gl-native/blob/76925d25788f163726f5897e2255633fe3fcdc90/platform/macos/app/MapDocument.m#L660):
release-quick-help<br>
release-code-completion

macos-sdk-symbol-visibility [on the same line](https://github.com/mapbox/mapbox-gl-native/blob/76925d25788f163726f5897e2255633fe3fcdc90/platform/macos/app/MapDocument.m#L660):
pr-quick-help<br>
pr-code-completion

Putting the two keywords on the same line seems to fix the problem, but I’ve only tried this so far inside macos.xcodeproj so far.

frederoni:
>  Ideally, we'd have tests that cover every class, but we generally don't have tests for UI classes yet.

If we would have tests that would cover every MGL class the test scheme would link and we wouldn't be able to catch incorrect symbol visibility in time? If that's not the case, it would be fairly easy to use the Objective-C Runtime to get a list of all MGL classes and verify visibility on them.

1ec5:
> If we would have tests that would cover every MGL class, the test scheme would link and we wouldn't be able to catch incorrect symbol visibility in time?

I'm not worried about classes we already have. I'm worried about a PR that adds a few public classes, one of which is significant enough that the contributor remembers to make it visible, while they overlook the fact that the other helper classes are also public (or have Private target membership!) and need to be visible. If we're vigilant enough to remind the contributor to add a pro forma "test", we'd already be vigilant enough to remind them to add the macro in the first place. But I estimate that we'd miss a class until after a prerelease about 30% of the time without an automated system. The only way "test everything" would work is if we gather code coverage and already had 100% code coverage, so that every dip in coverage would be detectable.

> If that's not the case, it would be fairly easy to use the Objective-C runtime to get a list of all loaded classes and validate visibility on them in a unit test.

This PR is about linker visibility, not runtime visibility. How would we distinguish between a class that's invisible because it should be versus one that's invisible because we forgot to add the macro? NSClassFromString() has to be able to return all classes, including private, invisible classes. Otherwise the class code could never run.

Maybe there's a way to enumerate SourceKitten's generated public symbol list, mangle the symbols, and ensure that they have the right entry in otool output for the framework? Sounds really complicated.

kkaefer:
I removed the `extern/extern "C"` keyword from the macro, and renamed it to `MGL_EXPORT`.

PR #7331, opened 2016-12-07
[iOS, macOS] Add support for dash patterns for Polyline Annotations

JesseCrocker:
This adds a new method to MGLMapViewDelegate -mapView:lineDasharrayForPolylineAnnotation: to enable proving dash patterns for Polyline annotations.

### DIFF
--- a/include/mbgl/annotation/annotation.hpp
+++ b/include/mbgl/annotation/annotation.hpp
@@ -32,6 +32,7 @@ public:
     style::PropertyValue<float> opacity { 1.0f };
     style::PropertyValue<float> width { 1.0f };
     style::PropertyValue<Color> color { Color::black() };
+    style::PropertyValue<std::vector<float>> dasharray { { 1 } };
 };
 
 class FillAnnotation {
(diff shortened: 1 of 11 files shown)


1ec5:
Thanks for the PR. We’ve wanted to add this functionality to the annotation API ever since we made it, and now it’s possible thanks to the runtime styling API. However, MGLMapViewDelegate has grown unwieldy with so many styling-related methods, and delegate methods have proven to be rather inconvenient for this purpose versus directly styling an object. As a result, we’re planning to steer developers towards the runtime styling API whenever more advanced styling is required.

That said, there have been proposals to bridge the two APIs, such as https://github.com/mapbox/mapbox-gl-native/issues/1734#issuecomment-242802281, which proposes a delegate method that provides a “prototype style layer” for a particular annotation. If implementing that approach minimizes the amount of maintenance overhead for annotations going forward, that might be a viable path forward for the functionality you’re after. Otherwise, reimplementing polyline annotations as MGLLineStyleLayers would be your best bet.

JesseCrocker:
Thanks for the explanation.

PR #7335, opened 2016-12-07
[ios] Document when it's safe to call runtime styling functions

ericrwolfe:
It's currently unclear when it's safe to modify an `MGLMapView`'s style. Most developers are going to try to update the map style in `viewDidLoad`.


platform/darwin/src/MGLStyle.h
```diff
@@ -200,7 +200,11 @@ static const NSInteger MGLStyleDefaultVersion = 9;
 /**
  Adds a new source to the current style.
 
- @note Adding the same source instance more than once will result in a 
+ @note Wait until the map style has finished loading before adding a new source.
+    Implement `MGLMapViewDelegate` methods `-mapViewDidFinishLoadingMap:`
+    or `-mapView:didFinishLoadingStyle:`
```
1ec5: Nit: end the sentence with a period. Also, two `@note`s in a row is a bit much; I don’t think this paragraph needs that treatment.

PR #7334, opened 2016-12-07
[ios, macos] Rename MGLGeoJSONSource to MGLShapeSource

boundsj:
ref https://github.com/mapbox/mapbox-gl-native/issues/7160

PR #7337, opened 2016-12-07
[ios] Add inline examples in documentation

ericrwolfe:
Examples for:
* Sources
  * `MGLGeoJSONSource`
  * `MGLVectorSource`
  * `MGLRasterSource`
* Layers
  * `MGLCircleStyleLayer`
  * `MGLFillStyleLayer`
  * `MGLLineStyleLayer`
  * `MGLSymbolStyleLayer`
* `MGLVectorStyleLayer.predicate`

Swift 3 for now. We can add Objective-C inline or a link to the respective examples on http://mapbox.com/ios-sdk/examples later.

### DIFF
--- a/platform/darwin/src/MGLCircleStyleLayer.h
+++ b/platform/darwin/src/MGLCircleStyleLayer.h
@@ -71,8 +71,8 @@ typedef NS_ENUM(NSUInteger, MGLCircleTranslateAnchor) {
     22: MGLStyleValue(rawValue: 180)
  ])
  layer.circleOpacity = MGLStyleValue(rawValue: 0.7)
- layer.predicate = NSPredicate(format: "%K == %@", "ethnicity", "ewok")
- mapView.style().add(layer)
+ layer.predicate = NSPredicate(format: "%K == %@", "marital-status", "married")
+ mapView.style.addLayer(layer)
  ```
  */
 @interface MGLCircleStyleLayer : MGLVectorStyleLayer
--- a/platform/darwin/src/MGLCircleStyleLayer.h
+++ b/platform/darwin/src/MGLCircleStyleLayer.h
@@ -59,6 +59,21 @@ typedef NS_ENUM(NSUInteger, MGLCircleTranslateAnchor) {
  otherwise, find it using the `MGLStyle.layers` property. You can also create a
  new circle style layer and add it to the style using a method such as
  `-[MGLStyle addLayer:]`.
+
+ ### Example ###
+
+ ```swift
+ let layer = MGLCircleStyleLayer(identifier: "circles", source: population)
+ layer.sourceLayerIdentifier = "population"
+ layer.circleColor = MGLStyleValue(rawValue: UIColor.green)
+ layer.circleRadius = MGLStyleValue(base: 1.75, stops: [
+    12: MGLStyleValue(rawValue: 2),
+    22: MGLStyleValue(rawValue: 180)
+ ])
+ layer.circleOpacity = MGLStyleValue(rawValue: 0.7)
+ layer.predicate = NSPredicate(format: "%K == %@", "ethnicity", "ewok")
+ mapView.style().add(layer)
+ ```
  */
 @interface MGLCircleStyleLayer : MGLVectorStyleLayer
 
--- a/platform/darwin/src/MGLCircleStyleLayer.h
+++ b/platform/darwin/src/MGLCircleStyleLayer.h
@@ -66,7 +66,7 @@ typedef NS_ENUM(NSUInteger, MGLCircleTranslateAnchor) {
  let layer = MGLCircleStyleLayer(identifier: "circles", source: population)
  layer.sourceLayerIdentifier = "population"
  layer.circleColor = MGLStyleValue(rawValue: UIColor.green)
- layer.circleRadius = MGLStyleValue(base: 1.75, stops: [
+ layer.circleRadius = MGLStyleValue(interpolationBase: 1.75, stops: [
     12: MGLStyleValue(rawValue: 2),
     22: MGLStyleValue(rawValue: 180)
  ])
--- a/platform/darwin/src/MGLCircleStyleLayer.h
+++ b/platform/darwin/src/MGLCircleStyleLayer.h
@@ -60,20 +60,7 @@ typedef NS_ENUM(NSUInteger, MGLCircleTranslateAnchor) {
  new circle style layer and add it to the style using a method such as
  `-[MGLStyle addLayer:]`.
 
- ### Example ###
-
- ```swift
- let layer = MGLCircleStyleLayer(identifier: "circles", source: population)
- layer.sourceLayerIdentifier = "population"
- layer.circleColor = MGLStyleValue(rawValue: UIColor.green)
- layer.circleRadius = MGLStyleValue(interpolationBase: 1.75, stops: [
-    12: MGLStyleValue(rawValue: 2),
-    22: MGLStyleValue(rawValue: 180)
- ])
- layer.circleOpacity = MGLStyleValue(rawValue: 0.7)
- layer.predicate = NSPredicate(format: "%K == %@", "marital-status", "married")
- mapView.style.addLayer(layer)
- ```
+ <!--EXAMPLE: MGLCircleStyleLayer-->
  */
 @interface MGLCircleStyleLayer : MGLVectorStyleLayer
 
--- a/platform/darwin/src/MGLLineStyleLayer.h
+++ b/platform/darwin/src/MGLLineStyleLayer.h
@@ -94,13 +94,13 @@ typedef NS_ENUM(NSUInteger, MGLLineTranslateAnchor) {
  let layer = MGLLineStyleLayer(identifier: "trails-path", source: trails)
  layer.sourceLayerIdentifier = "trails"
  layer.lineWidth = MGLStyleValue(base: 1.5, stops: [
-    14: MGLStyleValue(rawValue: 2),
-    18: MGLStyleValue(rawValue: 20),
+ 14: MGLStyleValue(rawValue: 2),
+ 18: MGLStyleValue(rawValue: 20),
  ])
  layer.lineColor = MGLStyleValue(rawValue: UIColor.brown)
  layer.lineCap = MGLStyleValue(rawValue: NSValue(mglLineCap: .round))
- layer.predicate = NSPredicate(format: "%K == %@", "trail-type" "mountain-biking")
- mapView.style().add(layer)
+ layer.predicate = NSPredicate(format: "%K == %@", "trail-type", "mountain-biking")
+ mapView.style.addLayer(layer)
  ```
  */
 @interface MGLLineStyleLayer : MGLVectorStyleLayer
--- a/platform/darwin/src/MGLLineStyleLayer.h
+++ b/platform/darwin/src/MGLLineStyleLayer.h
@@ -87,6 +87,21 @@ typedef NS_ENUM(NSUInteger, MGLLineTranslateAnchor) {
  otherwise, find it using the `MGLStyle.layers` property. You can also create a
  new line style layer and add it to the style using a method such as
  `-[MGLStyle addLayer:]`.
+
+ ### Example ###
+
+ ```swift
+ let layer = MGLLineStyleLayer(identifier: "trails-path", source: trails)
+ layer.sourceLayerIdentifier = "trails"
+ layer.lineWidth = MGLStyleValue(base: 1.5, stops: [
+    14: MGLStyleValue(rawValue: 2),
+    18: MGLStyleValue(rawValue: 20),
+ ])
+ layer.lineColor = MGLStyleValue(rawValue: UIColor.brown)
+ layer.lineCap = MGLStyleValue(rawValue: NSValue(mglLineCap: .round))
+ layer.predicate = NSPredicate(format: "%K == %@", "trail-type" "mountain-biking")
+ mapView.style().add(layer)
+ ```
  */
 @interface MGLLineStyleLayer : MGLVectorStyleLayer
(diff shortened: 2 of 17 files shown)


platform/darwin/src/MGLCircleStyleLayer.h
```diff
@@ -43,6 +43,21 @@ typedef NS_ENUM(NSUInteger, MGLCircleTranslateAnchor) {
...
+ let layer = MGLCircleStyleLayer(identifier: "circles", source: population)
+ layer.sourceLayerIdentifier = "population"
+ layer.circleColor = MGLStyleValue(rawValue: UIColor.green)
+ layer.circleRadius = MGLStyleValue(base: 1.75, stops: [
+    12: MGLStyleValue(rawValue: 2),
+    22: MGLStyleValue(rawValue: 180)
+ ])
+ layer.circleOpacity = MGLStyleValue(rawValue: 0.7)
+ layer.predicate = NSPredicate(format: "%K == %@", "ethnicity", "ewok")
+ mapView.style().add(layer)
```
1ec5: #6097 replaced `MGLMapView.style()` with an `MGLMapView.style` property.

platform/darwin/src/MGLCircleStyleLayer.h
```diff
@@ -43,6 +43,21 @@ typedef NS_ENUM(NSUInteger, MGLCircleTranslateAnchor) {
...
+ ```swift
+ let layer = MGLCircleStyleLayer(identifier: "circles", source: population)
+ layer.sourceLayerIdentifier = "population"
+ layer.circleColor = MGLStyleValue(rawValue: UIColor.green)
+ layer.circleRadius = MGLStyleValue(base: 1.75, stops: [
+    12: MGLStyleValue(rawValue: 2),
+    22: MGLStyleValue(rawValue: 180)
+ ])
+ layer.circleOpacity = MGLStyleValue(rawValue: 0.7)
+ layer.predicate = NSPredicate(format: "%K == %@", "ethnicity", "ewok")
```
1ec5: Nit: `ethnicity == 'ewok'` would have clearer intent. And lol on the choice of key-value pair.

platform/darwin/src/MGLGeoJSONSource.h
```diff
@@ -59,6 +59,15 @@ extern const MGLGeoJSONSourceOption MGLGeoJSONSourceOptionSimplificationToleranc
 /**
  A GeoJSON source.
 
+ ### Example ###
+ 
+ ```swift
+ var coordinates: [CLLocationCoordinate2D] = []
+ let polyline = MGLPolylineFeature(coordinates: &coordinates, count: UInt(coordinates.count))
+ let source = MGLGeoJSONSource(identifier: "lines", features: [polyline], options: nil)
+ self.mapView.style().add(source)
```
1ec5: Nit: other examples omit `self.`, presumably because `mapView` could be a local variable.

platform/darwin/src/MGLGeoJSONSource.h
```diff
@@ -59,6 +59,15 @@ extern const MGLGeoJSONSourceOption MGLGeoJSONSourceOptionSimplificationToleranc
 /**
  A GeoJSON source.
 
+ ### Example ###
+ 
+ ```swift
+ var coordinates: [CLLocationCoordinate2D] = []
+ let polyline = MGLPolylineFeature(coordinates: &coordinates, count: UInt(coordinates.count))
+ let source = MGLGeoJSONSource(identifier: "lines", features: [polyline], options: nil)
```
1ec5: #7145 renamed this initializer to `MGLGeoJSONSource(identifier:shape:options:)`, and #7334 would go a step further, renaming the class to MGLShapeSource. It still makes sense to demonstrate creating the source with a feature rather than an unadorned shape.

platform/darwin/src/MGLVectorStyleLayer.h
```diff
@@ -128,6 +128,22 @@ NS_ASSUME_NONNULL_BEGIN
...
  `@"%K == 'LineString'", @"$type"`.
+ 
+ ### Example ###
+ To create a filter with the logic `(index == 10 || index == 5) && ele >= 200`, 
+ you could create a predicate using `NSCompoundPredicate` along these lines:
+ 
+ ```swift
+ let meters = 1500.0
+ let layer = MGLLineStyleLayer(identifier: "contour", source: contours)
+ layer.predicate = NSCompoundPredicate(andPredicateWithSubpredicates: [
```
1ec5: This is a lot of typing. Is there a point you want to make about the individual classes within the NSPredicate class cluster, or would a single predicate format string such as `(index == 10 || index == 5) && ele >= 200` be sufficient?

friedbunny:
> Most of the comments below relate to changes that have taken place since beta 4 that break these examples. Have you considered other approaches to ensuring maintainability, like keeping these examples in a Swift playground and pulling them into the header in a build step?

Maintainability and testability are important considerations that have kept us from expanding our examples repertoire in the past, so I’m agreed that we should tackle those problems in tandem with the actual example-writing process.

Using a Swift playground is an interesting idea that opens the possibility of users being able to fiddle with live examples themselves, which would be an obvious win. I still think that we give Obj-C examples the same prominence, though, which means finding a solution that works for both (or a second Obj-C solution).

boundsj:
`MGLGeoJSONSource` was renamed to `MGLShapeSource` in https://github.com/mapbox/mapbox-gl-native/pull/7334

ericrwolfe:
Added unit tests for each of the inline examples. Should help catch any breaking changes to the examples.

platform/darwin/src/MGLLineStyleLayer.h
```diff
@@ -67,6 +67,21 @@ typedef NS_ENUM(NSUInteger, MGLLineTranslateAnchor) {
...
  `MGLMapView` for its `style` and obtain existing layers using the 
  `-[MGLStyle layerWithIdentifier:]` method. 
+ 
+ ### Example ###
+ 
+ ```swift
+ let layer = MGLLineStyleLayer(identifier: "trails-path", source: trails)
+ layer.sourceLayerIdentifier = "trails"
+ layer.lineWidth = MGLStyleValue(base: 1.5, stops: [
+ 14: MGLStyleValue(rawValue: 2),
```
friedbunny: Missed the indentation here.

friedbunny:
@mayagao is reworking jazzy theme elsewhere, but here’s how these examples look now:

screen shot 2016-12-13 at 1 54 37 pm

👍

platform/darwin/src/MGLRasterSource.h
```diff
@@ -14,6 +14,22 @@ NS_ASSUME_NONNULL_BEGIN
  raster source is added to an `MGLStyle` object along with an
  `MGLRasterStyleLayer` object. Use a raster style layer to control the
  appearance of content supplied by the raster source.
+ 
+ ### Example ###
+ 
+ ```swift
+ // Define tileset
+ let tileset = MGLTileSet(tileURLTemplates: ["https://example.com/raster-tiles/{z}/{x}/{y}.png"])
```
1ec5: #7377 removed MGLTileSet in favor of direct source initializers with option dictionaries. Sorry to keep shifting the goalposts on you. There’s still an attribution option that takes an HTML string, but now there’s also an option that takes structured MGLAttributionInfo objects.

platform/darwin/src/MGLShapeSource.h
```diff
@@ -66,6 +66,22 @@ extern const MGLShapeSourceOption MGLShapeSourceOptionSimplificationTolerance;
...
+    CLLocationCoordinate2D(latitude: 37.77, longitude: -122.42),
+    CLLocationCoordinate2D(latitude: 38.91, longitude: -77.04),
+ ]
+ let polyline = MGLPolylineFeature(coordinates: &coordinates, count: UInt(coordinates.count))
+ let shape = MGLShapeCollectionFeature(shapes: [polyline])
+ let source = MGLShapeSource(identifier: "lines", shape: shape, options: nil)
+ mapView.style.addSource(source)
+ ```
+ 
+ @see <a href="https://www.mapbox.com/mapbox-gl-style-spec/#sources-geojson">The
```
1ec5: These links become redundant once you merge in (or rebase atop) #7377.

platform/darwin/src/MGLCircleStyleLayer.h
```diff
@@ -55,8 +55,8 @@ typedef NS_ENUM(NSUInteger, MGLCircleTranslateAnchor) {
     22: MGLStyleValue(rawValue: 180)
  ])
  layer.circleOpacity = MGLStyleValue(rawValue: 0.7)
- layer.predicate = NSPredicate(format: "%K == %@", "ethnicity", "ewok")
- mapView.style().add(layer)
+ layer.predicate = NSPredicate(format: "%K == %@", "marital-status", "married")
+ mapView.style.addLayer(layer)
```
boundsj: @1ec5 it's interesting that making the style a property instead of returning it in a method had this effect. Is this a case for `NS_SWIFT_NAME`?

1ec5: You’re referring to the change from `add(_:)` to `addLayer(_:)`? I was curious about that too. It’s unlikely to be due to the change from `MGLMapView.style()` to `MGLMapView.style`. The more likely cause is that #6097 added `layers` and `sources` properties, or that we started using Xcode 8.2 instead 8.1, but I’m not really sure exactly what introduced the change.

As for whether we should use `NS_SWIFT_NAME`, I’m ambivalent. If we do, we should make sure to refine all the layer and source methods on this class.

ericrwolfe:
Last commit should cover all of the outstanding changes needed. Lmk if you see anything else that could use an inline example.

/cc @1ec5 @friedbunny

platform/darwin/src/MGLCircleStyleLayer.h
```diff
@@ -43,6 +43,21 @@ typedef NS_ENUM(NSUInteger, MGLCircleTranslateAnchor) {
  instantiate a new circle layer to add to a map style or you may query an 
  `MGLMapView` for its `style` and obtain existing layers using the 
  `-[MGLStyle layerWithIdentifier:]` method. 
+
+ ### Example ###
+
+ ```swift
```
1ec5: These style layer headers are generated by platform/darwin/scripts/generate-style-code.js, so the next time someone runs `make style-code-darwin`, these changes will get blown away. The approach I took in #7475 was to override a property on each layer’s entry in the style specification JSON to provide documentation. Unfortunately, it’d be quite inconvenient to embed a multiline code example inside a JSON string.

The approach I’d recommend would be to have the overridden style specification JSON name the relevant function within MGLDocumentationExampleTests; generate-style-code.js can then extract that function using a regular expression and insert it into the generated header.

ericrwolfe:
Last set of commits after rebasing on release-3.4.0 adds a run script that will extract the examples from `MGLDocumentationExampleTests` and replaces the corresponding comment token in the header docs.

platform/ios/scripts/document.sh
```diff
@@ -35,11 +35,16 @@ cp platform/ios/screenshot.png "${OUTPUT}"
...
+DEFAULT_FRAMEWORK_PATH="build/ios/pkg/dynamic/Mapbox.framework"
+FRAMEWORK_PATH=${FRAMEWORK_PATH:-$DEFAULT_FRAMEWORK_PATH}
+
 jazzy \
     --config platform/ios/jazzy.yml \
     --sdk iphonesimulator \
     --github-file-prefix https://github.com/mapbox/mapbox-gl-native/tree/${BRANCH} \
     --module-version ${SHORT_VERSION} \
+    --framework-root ${FRAMEWORK_PATH} \
+    --umbrella-header "${FRAMEWORK_PATH}/Headers/Mapbox.h" \
```
ericrwolfe: By default, Jazzy generates its docs from the project source, which excludes any changes by the run script, so I've added an optional `FRAMEWORK_PATH` env variable that by default searches for the dynamic framework output by `make iframework`.

platform/darwin/src/MGLVectorSource.h
```diff
@@ -24,6 +24,9 @@ NS_ASSUME_NONNULL_BEGIN
  (<var>extent</var>&nbsp;×&nbsp;2)&nbsp;−&nbsp;1, inclusive. Any vector style
  layer initialized with a vector source must have a non-`nil` value in its
  `sourceLayerIdentifier` property.
+
+ <!--EXAMPLE: MGLVectorSource-->
+ ```
```
1ec5: Stray backticks.

platform/ios/scripts/document.sh
1ec5: If someone runs `make idocument` after building only the static framework, this will probably error out. It’s a workflow I don’t personally use, but I suppose someone might end up doing that after reading DEVELOPING.md. I wonder if we could check for the existence of the dynamic framework and, if it doesn’t exist, fall back to the static framework’s Headers folder. But that’s all fodder for other PRs.

platform/ios/ios.xcodeproj/project.pbxproj
```diff
@@ -1913,6 +1920,37 @@
...
+			files = (
+			);
+			inputPaths = (
+			);
+			name = "Add Examples to Documentation";
+			outputPaths = (
+			);
+			runOnlyForDeploymentPostprocessing = 0;
+			shellPath = /bin/sh;
+			shellScript = "node \"${SRCROOT}/scripts/add-examples-to-docs.js\"";
```
1ec5: If you declare add-examples-to-docs.js and MGLDocumentationExampleTests.swift as input files to this build phase and the modified headers as output files, Xcode will avoid running this script when none of those files have changed. But since that’d be one more place that would have to get updated every time we add an example, it’s fine to leave the input and output files out.

PR #7338, opened 2016-12-07
[ios] Reorder classes in jazzy docs sidebar

ericrwolfe:
* Move guides front-and-center (to be expanded out to include more guides)
* Organize shape classes heirarchically for clarity
* Clarify features, sources, and layers as related to map styling (as opposed to annotations)

### DIFF
--- a/platform/ios/jazzy.yml
+++ b/platform/ios/jazzy.yml
@@ -26,10 +26,6 @@ custom_categories:
       - MGLMapViewDelegate
       - MGLStyle
       - MGLUserTrackingMode
-  - name: User Interaction
-    children:
-      - MGLCalloutView
-      - MGLCalloutViewDelegate
   - name: Protocols and Abstract Classes
     children:
       - MGLAnnotation
@@ -52,6 +48,10 @@ custom_categories:
       - MGLAnnotationView
       - MGLUserLocation
       - MGLUserLocationAnnotationView
+  - name: User Interaction
+    children:
+      - MGLCalloutView
+      - MGLCalloutViewDelegate
   - name: Style Features
     children:
       - MGLPointFeature
--- a/platform/ios/jazzy.yml
+++ b/platform/ios/jazzy.yml
@@ -81,6 +81,10 @@ custom_categories:
       - MGLFillStyleLayer
       - MGLLineStyleLayer
       - MGLSymbolStyleLayer
+  - name: Style Values
+    children:
+      - MGLStyleValue
+      - NSValue(MGLAdditions)
   - name: Offline Maps
     children:
       - MGLOfflineRegion
--- a/platform/ios/jazzy.yml
+++ b/platform/ios/jazzy.yml
@@ -94,6 +94,7 @@ custom_categories:
       - MGLCoordinateBounds
       - MGLCoordinateBoundsEqualToCoordinateBounds
       - MGLCoordinateBoundsGetCoordinateSpan
+      - MGLCoordinateBoundsIntersectsCoordinateBounds
       - MGLCoordinateBoundsIsEmpty
       - MGLCoordinateBoundsMake
       - MGLCoordinateBoundsOffset
--- a/platform/ios/jazzy.yml
+++ b/platform/ios/jazzy.yml
@@ -15,6 +15,9 @@ umbrella_header: src/Mapbox.h
 framework_root: ../darwin/src
 
 custom_categories:
+  - name: Guides
+    children:
+      - Info.plist Keys
   - name: Maps
     children:
       - MGLAccountManager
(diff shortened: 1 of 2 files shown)


platform/ios/jazzy.yml
```diff
@@ -80,15 +91,12 @@ custom_categories:
       - MGLTilePyramidOfflineRegion
   - name: Geometry
     children:
```
friedbunny: Let’s also add `MGLCoordinateBoundsIntersectsCoordinateBounds` to Geometry. (It’s in Other Functions, currently.)

platform/ios/jazzy.yml
```diff
@@ -23,53 +26,61 @@ custom_categories:
       - MGLMapViewDelegate
       - MGLStyle
       - MGLUserTrackingMode
-  - name: Annotations
+  - name: User Interaction
```
friedbunny: Now that this section only contains callout views, I think we should move it farther down the list — perhaps after the new Annotations section?

friedbunny:
Any ideas of what, if anything, to do with `MGLStyleConstantValue`, `MGLStyleFunction`, and `MGLStyleValue`? They’re falling under Other Classes, currently.

platform/ios/jazzy.yml
```diff
@@ -23,53 +26,61 @@ custom_categories:
...
+      - MGLPointCollectionFeature      
       - MGLShapeCollectionFeature
+  - name: Style Sources
+    children:
+      - MGLSource
+      - MGLTileSet
+      - MGLGeoJSONSource
+      - MGLRasterSource
+      - MGLVectorSource
+  - name: Style Layers (Abstract)
```
1ec5: I wonder whether dividing classes by abstractness is really a good idea for the style layer classes, which were designed to be a nested inheritance tree, with each level sort of peeling a layer off the onion. On the one hand, categorizing them this way helps developers find the classes they can safely initialize. On the other hand, it also obscures the relationships between each concrete class and the abstract superclasses that contribute to its properties and initializers.

To be sure, one alphabetical list wasn’t especially enlightening, either, but what about a flattened inheritance hierarchy like this:

* MGLStyleLayer
* MGLBackgroundStyleLayer
* MGLForegroundStyleLayer
* MGLRasterStyleLayer
* MGLVectorStyleLayer
* MGLCircleStyleLayer
* MGLFillStyleLayer
* _MGLFillExtrusionStyleLayer would go here_
* MGLLineStyleLayer
* MGLSymbolStyleLayer

platform/ios/jazzy.yml
```diff
@@ -23,53 +26,61 @@ custom_categories:
...
+  - name: User Interaction
     children:
-      - MGLAnnotation
-      - MGLAnnotationImage
-      - MGLAnnotationView
       - MGLCalloutView
       - MGLCalloutViewDelegate
+  - name: Protocols and Abstract Classes
+    children:
+      - MGLAnnotation
```
1ec5: This protocol is unlike the others in this category in that the developer can actually make any class they want conform to it, and we’ll treat that class just like an MGLPointAnnotation. I think it should go in the “Annotations” category alongside MGLAnnotationImage and MGLAnnotationView.

platform/ios/jazzy.yml
```diff
@@ -23,53 +26,61 @@ custom_categories:
...
-      - MGLMultiPolylineFeature
-      - MGLPointCollectionFeature      
       - MGLPointFeature
       - MGLPolygonFeature
       - MGLPolylineFeature
+      - MGLMultiPolygonFeature
+      - MGLMultiPolylineFeature
+      - MGLPointCollectionFeature      
       - MGLShapeCollectionFeature
+  - name: Style Sources
```
1ec5: I think we should avoid saying “style source”, even if a source is part of a style, because it seems to imply that it defines an appearance rather than unstyled content. Since “data source” has a specific meaning on iOS, how about “Map Sources” or “Content Sources”?

ericrwolfe: Main reasoning here was to communicate that the source classes are associated with runtime map styling in the same way as the feature and style layer classes. Trying to avoid confusion by grouping and differentiating all of these classes from the more primitive shape classes.

friedbunny:
@ericrwolfe I pushed some small changes, mainly addressing the review comments I made — feel free to revert if I overstepped.

platform/ios/jazzy.yml
```diff
@@ -23,53 +26,65 @@ custom_categories:
...
+      - MGLAnnotationImage
+      - MGLAnnotationView
       - MGLUserLocation
       - MGLUserLocationAnnotationView
-  - name: Map Data
+  - name: User Interaction
+    children:
+      - MGLCalloutView
+      - MGLCalloutViewDelegate
+  - name: Style Features
```
1ec5: By the way, these classes are also used as feature querying results. Whether a developer uses them more for runtime styling or feature querying probably depends on the use case.

platform/ios/jazzy.yml
1ec5: I agree with your reasoning, though I can’t shake the feeling that developers will overlook these classes when they want to add content to the map, especially in the case of MGLRasterSource, which isn’t something that people normally think they need to style.

platform/ios/jazzy.yml
1ec5: Similarly, a flattened inheritance hierarchy for content sources would look like:

* MGLSource
* MGLShapeSource
* MGLTileSource
* MGLRasterSource
* MGLVectorSource

Once #6940 lands in a future release, the list would look something like this:

* MGLSource
* MGLAbstractShapeSource
* MGLComputedShapeSource
* MGLShapeSource
* MGLTileSource
* MGLRasterSource
* MGLVectorSource

platform/ios/jazzy.yml
friedbunny: Did this in 4057ac8.

platform/ios/jazzy.yml
```diff
@@ -23,53 +26,65 @@ custom_categories:
...
-  - name: Data Sources
+  - name: Style Values
     children:
-      - MGLSource
-      - MGLShapeSource
-      - MGLRasterSource
-      - MGLTileSet
-      - MGLVectorSource
+      - MGLStyleValue
+      - NSValue(MGLAdditions)
```
1ec5: Style value–related methods are only part of this category; there are also methods related to offline maps.

ericrwolfe: Hmm, true, guess I'll just exclude then and add a note to MGLStyleValue

friedbunny:
@1ec5 @ericrwolfe This seems ready (enough) to merge, yes? With all of the other jazzy PRs still moving, there will be plenty of opportunity to address other concerns as we break-in these changes.

platform/ios/jazzy.yml
```diff
@@ -23,55 +26,62 @@ custom_categories:
...
+  - name: Style Values
     children:
-      - MGLAttributionInfo
-      - MGLRasterSource
-      - MGLShapeSource
-      - MGLSource
-      - MGLTileCoordinateSystem
-      - MGLTileSource
-      - MGLVectorSource
+      - MGLStyleValue
```
1ec5: What about MGLConstantStyleValue or MGLStyleFunction? MGLStyleValue is a class cluster that includes those two classes, so the developer could get all the information they need with just MGLStyleValue, but it seems odd to have a category with only one item. If the idea is to deemphasize MGLConstantStyleValue and MGLStyleFunction in favor of MGLStyleValue, perhaps we could move this class up to “Maps” alongside MGLStyle.

platform/ios/jazzy.yml
```diff
@@ -23,55 +26,62 @@ custom_categories:
...
+    children:
+      - MGLSource
+      - MGLTileSource
+      - MGLShapeSource
+      - MGLRasterSource
+      - MGLVectorSource
   - name: Style Layers
     children:
+      - MGLStyleLayer
+      - MGLForegroundStyleLayer
```
1ec5: MGLForegroundStyleLayer and MGLVectorStyleLayer are also abstract classes.

platform/ios/jazzy.yml
```diff
@@ -23,55 +26,62 @@ custom_categories:
       - MGLMapViewDelegate
       - MGLStyle
       - MGLUserTrackingMode
-  - name: Annotations
+  - name: Protocols and Abstract Classes
```
1ec5: I wonder if this category’s title still makes sense given that some abstract classes are in more specific categories below.

1ec5: OK, let’s move them back then. I’ll also work on a way for a way to provide more info about each class in the table of contents within jazzy.yml so we don’t have to hard-code that list inside SCSS. 😬

1ec5:
I brought the macOS table of contents more in line with the iOS table of contents in 63a85c4fb02a733204f74d74ddbf723495438b17. I had to elide a couple categories because they had too few classes once iOS-only classes were removed.

Issue #7336, opened 2016-12-07
GeoJSON (Shape) sources initialized with MGLShape objects should work with predicate filters


boundsj:
In https://github.com/mapbox/mapbox-gl-native/pull/7145 we modified `MGLGeoJSONSource` so that it can be created with an `MGLShape`. However, since `MGLShape` does not convey a feature's `identifier` and `attributes` properties, we lost the ability to apply a NSPredicate filter to a layer for the source created with a shape.

For instance, this used to work:

```Objective-C
    MGLPolygonFeature *feature = [MGLPolygonFeature polygonWithCoordinates:leafCoords count:coordsCount];
    feature.identifier = @"leaf-feature";
    feature.attributes = @{@"color": @"red"};
    
    MGLGeoJSONSource *source = [[MGLGeoJSONSource alloc] initWithIdentifier:@"leaf-source" shape:feature options:nil];
    [self.mapView.style addSource:source];
    
    MGLFillStyleLayer *layer = [[MGLFillStyleLayer alloc] initWithIdentifier:@"leaf-fill-layer" source:source];
    layer.predicate = [NSPredicate predicateWithFormat:@"color = %@", @"red"];
    MGLStyleValue *fillColor = [MGLStyleValue<UIColor *> valueWithRawValue:[UIColor redColor]];
    layer.fillColor = fillColor;
    [self.mapView.style addLayer:layer];
```

Now, If the `layer.predicate` is set, the feature is not visible since the attributes cannot be found on the object (shape).

cc @frederoni @1ec5

boundsj:
This was fixed by https://github.com/mapbox/mapbox-gl-native/pull/7339

PR #7339, opened 2016-12-08
[ios, macos] Load features into shape source if possible

boundsj:
fixes https://github.com/mapbox/mapbox-gl-native/issues/7336

### DIFF
--- a/platform/darwin/src/MGLFeature.mm
+++ b/platform/darwin/src/MGLFeature.mm
@@ -1,4 +1,5 @@
 #import "MGLFeature_Private.h"
+#import "MGLShapeCollectionFeature_Private.h"
 
 #import "MGLPointAnnotation.h"
 #import "MGLPolyline.h"
@@ -177,6 +178,15 @@
     return mbgl::Feature{geometry};
 }
 
+- (mbgl::FeatureCollection)mbglFeatureCollection {
+    mbgl::FeatureCollection featureCollection;
+    featureCollection.reserve(self.shapes.count);
+    for (id <MGLFeaturePrivate> feature in self.shapes) {
+        featureCollection.push_back([feature mbglFeature]);
+    }
+    return featureCollection;
+}
+
 @end
 
 /**
@@ -277,8 +287,7 @@ public:
     }
     
     MGLShape <MGLFeaturePrivate> * operator()(const mbgl::Feature &feature) const {
-        GeometryEvaluator<T> evaluator;
-        MGLShape <MGLFeaturePrivate> *shape = mapbox::geometry::geometry<T>::visit(feature.geometry, evaluator);
+        MGLShape <MGLFeaturePrivate> *shape = (MGLShape <MGLFeaturePrivate> *)MGLFeatureFromMBGLFeature(feature);
         return shape;
     }
     
--- a/platform/darwin/src/MGLShapeSource.mm
+++ b/platform/darwin/src/MGLShapeSource.mm
@@ -3,6 +3,7 @@
 #import "MGLMapView_Private.h"
 #import "MGLSource_Private.h"
 #import "MGLFeature_Private.h"
+#import "MGLShapeCollectionFeature_Private.h"
 #import "MGLShape_Private.h"
 
 #import "NSURL+MGLAdditions.h"
@@ -99,8 +100,15 @@ const MGLShapeSourceOption MGLShapeSourceOptionSimplificationTolerance = @"MGLSh
         source->setGeoJSON(geojson);
         _shape = MGLShapeFromGeoJSON(geojson);
     } else {
-        const auto geojson = mbgl::GeoJSON{self.shape.geometryObject};
-        source->setGeoJSON(geojson);
+        if ([self.shape isKindOfClass:[MGLShapeCollectionFeature class]]) {
+            MGLShapeCollectionFeature *feature = (MGLShapeCollectionFeature *)self.shape;
+            source->setGeoJSON(mbgl::GeoJSON{[feature mbglFeatureCollection]});
+        } else if ([self.shape conformsToProtocol:@protocol(MGLFeature)]) {
+            id<MGLFeaturePrivate> feature = (id<MGLFeaturePrivate>)self.shape;
+            source->setGeoJSON(mbgl::GeoJSON{[feature mbglFeature]});
+        } else {
+            source->setGeoJSON(mbgl::GeoJSON{self.shape.geometryObject});
+        }
     }
     
     _pendingSource = std::move(source);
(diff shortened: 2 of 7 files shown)


boundsj:
This attempts to keep the features object intact if that is what we actually get in a "shape". Features have `identifer` and `attributes` (aka `properties` in GeoJSON) properties that are required for some styling and filtering operations to work.

The WIP commit so far breaks a GeoJSON source test where a shape collection feature is nested in a shape collection feature.

platform/darwin/src/MGLFeature.mm
```diff
@@ -177,6 +177,18 @@ - (NSDictionary *)geoJSONDictionary {
...
 }
 
+- (mbgl::FeatureCollection)mbglFeatureCollection {
+    mbgl::FeatureCollection featureCollection;
+    featureCollection.reserve(self.shapes.count);
+    for (id <MGLFeaturePrivate> feature in self.shapes) {
+        if ([feature isMemberOfClass:[MGLShapeCollectionFeature class]]) {
+            featureCollection.push_back([self mbglFeatureCollection]);
+        }
+        featureCollection.push_back([feature mbglFeature]);
```
frederoni: MGLShapeCollectionFeature raises an exception upon `-mbglFeature`. I guess this should be in an else statement?

platform/darwin/src/MGLFeature.mm
```diff
@@ -177,6 +177,18 @@ - (NSDictionary *)geoJSONDictionary {
     return mbgl::Feature{geometry};
 }
 
+- (mbgl::FeatureCollection)mbglFeatureCollection {
+    mbgl::FeatureCollection featureCollection;
+    featureCollection.reserve(self.shapes.count);
+    for (id <MGLFeaturePrivate> feature in self.shapes) {
+        if ([feature isMemberOfClass:[MGLShapeCollectionFeature class]]) {
```
1ec5: `-isKindOfClass:` is more forward-compatible than `-isMemberOfClass:`. Use the latter only if you want to avoid subclasses, for instance if you want to check for a shape collection that isn’t a shape collection feature.

platform/darwin/src/MGLFeature_Private.h
```diff
@@ -42,4 +42,10 @@ NS_DICTIONARY_OF(NSString *, id) *NSDictionaryFeatureForGeometry(NSDictionary *g
 
 @end
 
+@protocol MGLFeatureCollectionPrivate <NSObject>
```
1ec5: Hidden protocols are dangerous because they make it easy to write code that crashes on unrecognized selectors. MGLFeaturePrivate is necessary because MGLFeature is a protocol, but if MGLShapeCollectionFeature is the only class that would conform to this hidden protocol, it’d be safer to declare a private extension on MGLShapeCollection specifically.

platform/darwin/src/MGLFeature.mm
boundsj: Good point. We won't ever get here if the super class, `MGLShapeCollection`, is used since that is a geometry that'll be handled on a different track than what we are doing here for features.

I don't think we actually need to support nested feature collections since it [does not look like they are supported in geometry.hpp](https://github.com/mapbox/geometry.hpp/blob/master/include/mapbox/geometry/feature.hpp#L72)

platform/darwin/src/MGLFeature.mm
boundsj: That was the intent of the if statement just above here. However I don't think we need to support this case (see [comment](https://github.com/mapbox/mapbox-gl-native/pull/7339/files#r91633052))

platform/darwin/src/MGLFeature_Private.h
boundsj: Sounds good. 

`MGLShapeCollectionFeature` declares conformance to `MGLFeaturePrivate` (and raises an exception) if you call the only method in that protocol. I just realized / remembered that this allows for the transformation in `GeometryEvaluator`.

boundsj:
Adjusted the test in 24f8abb since [feature collection objects should not be nested]( http://geojson.org/geojson-spec.html#feature-collection-objects) (ref https://github.com/mapbox/geometry.hpp/blob/master/include/mapbox/geometry/feature.hpp#L72)

boundsj:
It looks like plain old shapes (i.e. MGLPolygon instead of MGLPolygonFeature) still don't get drawn. I'm going to try and fix that, if possible, before this lands.

boundsj:
> It looks like plain old shapes (i.e. MGLPolygon instead of MGLPolygonFeature) still don't get drawn. 

Actually this works fine. Not sure what I was seeing before, probably crossed lat/lng in my test data.

@1ec5 @frederoni

boundsj:
Squashed and rebased on top of https://github.com/mapbox/mapbox-gl-native/pull/7334

platform/darwin/src/MGLShapeSource.mm
```diff
@@ -99,8 +100,15 @@ - (void)commonInit
         source->setGeoJSON(geojson);
         _shape = MGLShapeFromGeoJSON(geojson);
     } else {
-        const auto geojson = mbgl::GeoJSON{self.shape.geometryObject};
-        source->setGeoJSON(geojson);
+        if ([self.shape isMemberOfClass:[MGLShapeCollectionFeature class]]) {
```
1ec5: Using `-isKindOfClass:` will also avoid latent bugs if we ever end up needing to subclass MGLShapeCollectionFeature ourselves.

platform/ios/ios.xcodeproj/project.pbxproj
```diff
@@ -592,6 +592,7 @@
 		4085AF081D933DEA00F11B22 /* MGLTileSetTests.mm */ = {isa = PBXFileReference; fileEncoding = 4; lastKnownFileType = sourcecode.cpp.objcpp; name = MGLTileSetTests.mm; path = ../../darwin/test/MGLTileSetTests.mm; sourceTree = "<group>"; };
 		408AA8551DAEDA0800022900 /* NSDictionary+MGLAdditions.h */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.c.h; path = "NSDictionary+MGLAdditions.h"; sourceTree = "<group>"; };
 		408AA8561DAEDA0800022900 /* NSDictionary+MGLAdditions.mm */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.cpp.objcpp; path = "NSDictionary+MGLAdditions.mm"; sourceTree = "<group>"; };
+		40986FE51DFA2B5C004A7E6E /* MGLShapeCollectionFeature_Private.h */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.c.h; path = MGLShapeCollectionFeature_Private.h; sourceTree = "<group>"; };
```
1ec5: Add this file to macos.xcodeproj too.

1ec5:
This PR blocks #7377, where the conversion from an MGLFeature to an mbgl feature will be used as part of a general-purpose GeoJSON (de)serialization utility.

platform/ios/ios.xcodeproj/project.pbxproj
boundsj: That's strange... it is added locally. I'll try to figure out why it is not in this changeset

boundsj: Never mind! Fixed https://github.com/mapbox/mapbox-gl-native/pull/7339/files#diff-9276451d661540f77943125827376c97R303

PR #7296, opened 2016-12-08
Support dynamically-changing route data in QML

tmpsantos:
**Platform:** Ubuntu 14.04 with Qt 5.7
**Mapbox SDK version:** 73fa7ea6664fd91b4ccfe42a6b4cc061d615f65a

### Steps to trigger behavior

 1. Create a QML `MapboxMap` with a route JSON:

    ```qml
    MapboxMap {
      // ...
    parameters: [
        // ...
        MapParameter {
            id: myRoute
            property string type: "source"
            property string name: "routeSource"
            property string sourceType: "geojson"
            property string data: '{"type":"FeatureCollection", "features":...}'
        }
    ]
    ```
    * _Requires patching `qquickmapboxgl.cpp` to accept string data._
 2. Use QML/JS to dynamically change the data:

    ```js
    myRoute.data = someNewValidJSON;
    ```
### Expected behavior
Route re-draws with new route.

### Actual behavior
Error about `routeSource` already being defined.

### Workaround
Add the `MapParameter` initially using JS, so that it can later be destroyed and re-added with new data. _(Untested; unsure if related parameters using the `routeSource` need to be removed as well or not.)_

_Tagging @tmpsantos as requested._


tmpsantos:
:eyes: @brunoabinader

PR #7368, opened 2016-12-09
[core] Recalculate the style after adding a source

tmpsantos:
The source may have orphan layers, in case the layers were added before the source or the source was removed and added again with the same name.


tmpsantos:
:eyes: @ivovandongen

jfirebaugh:
> the source was removed and added again with the same name

This use case isn't supported -- see https://github.com/mapbox/mapbox-gl-native/issues/6612#issuecomment-252299676. Removing a source that's in use by a layer is likely to cause issues that go beyond misrendering if you add the source again with the same name. Rather than trying to fix that case, we should prevent removing active sources.

tmpsantos:
@jfirebaugh instead of removing/re-adding, would the approach of trying to update the source data field be acceptable? Something like a `Map::updateSource(...)`. This could be done atomically.

jfirebaugh:
@tmpsantos What do you mean by "update the source data field"? For GeoJSON sources, we already have these accessors (`GeoJSONSource::setURL` and `GeoJSONSource::setGeoJSON`). Are you looking for `VectorSource::setURL`?

tmpsantos:
@jfirebaugh thanks for the feedback. The mutable properties of the GeoJSON source are enough for the feature I need on Qt, closing this PR.

PR #7373, opened 2016-12-10
Fix tile debug rendering

jfirebaugh:
* Off-by-one in DebugBucket indexing
* Must unbind VAO before binding index buffer in Context::createIndexBuffer
* Get test-suite test passing so we catch this sort of regression


jfirebaugh:
*sigh* Rendering still fails in CI. `GL_LINE` just varies too much between implementations / graphics cards.

jfirebaugh:
Abandoned the test changes.

Issue #7375, opened 2016-12-11
MGLSource should not maintain parallel state


1ec5:
MGLSource and its subclasses, particularly MGLShapeSource, currently hang onto objects passed into their initializers in order to return them in their properties’ getters. This is a workaround for #6584, but it’s problematic because MGLSource objects don’t necessarily know the canonical value of these properties. For example, if you get a source via `-[MGLStyle sourceWithIdentifier:]`, the properties are all unset because the source was initialized via `-[MGLSource initWithRawSource:]`. You can then proceed to modify the source’s contents behind the backs of any other MGLShapeSource instances that wrap that source:

```swift
let one = MGLShapeSource(identifier: "one", shape: MGLPointFeature(), options: nil)
mapView.style.addSource(one)
// a short while later…
let uno = mapView.style.source(withIdentifier: "one")
uno.shape = someOtherFeature
// finally…
assert(one.shape is MGLPointFeature)
// 💣💥
```

/cc @boundsj

1ec5:
This is essentially blocked by #7376, because otherwise we can continue to expose an `-[MGLShapeSource setShape:]` but we’d have to eliminate `-[MGLShapeSource shape]`. A write-only property is theoretically possible but unheard of in Objective-C and Swift. (The closest thing I can think of is `-[NSView setNeedsDisplay:]`, which isn’t a property and which no one thinks is a good design.)

1ec5:
#7377 will eliminate `MGLShapeSource.geoJSONData` and make `MGLShapeSource.URL`’s getter computed, but `MGLShapeSource.shape` remains.

boundsj:
I updated the example in the original description above:
 s/`let uno = mapView.style.layer(withIdentifier: "one")`/`let uno = mapView.style.source(withIdentifier: "one")`

boundsj:
This gets into https://github.com/mapbox/mapbox-gl-native/issues/7376 and the memory concerns noted there, but would it make sense for `MGLStyle` to keep references to the source and layer objects in use? Then, asking the style for a source would/could always return the same source instance.

1ec5:
It's definitely possible and it'd solve several problems associated with these wrappers. We already do that for MGLOpenGLStyleLayers, which are entirely managed by the style. I wanted to avoid doing that for sources and layers in general, since there may be ways for the style to change behind MGLStyle's back, whereas a custom layer can only come from application code.

jfirebaugh:
[...]

PR #7377, opened 2016-12-12
Simplify MGLSource and subclasses

### LINKED ISSUE #3630: Provide a GeoJSON serialization and deserialization utility

### LINKED ISSUE #6987: MGLGeoJSONSource does not handle exceptions from GeoJSON parsing
The following code crashes the application with

> libc++abi.dylib: terminating with uncaught exception of type std::runtime_error: Feature must have a geometry property

It should instead raise the error using appropriate cocoa error reporting mechanism.

```
    MGLGeoJSONSource *geojsonSource = [[MGLGeoJSONSource alloc]
        initWithIdentifier:@"geojson"
               geoJSONData:[@"{\"type\": \"Feature\", \"feature\": {\"type\": \"Point\", \"coordinates\": [0, 0]}}" dataUsingEncoding:NSUTF8StringEncoding]
                   options:nil];
```

### LINKED ISSUE #7360: Fold MGLTileSet into MGLRasterSource and MGLVectorSource
MGLRasterSource and MGLVectorSource should have a new abstract superclass, MGLTileSource. MGLTileSet should be removed and its initializers and properties moved to MGLTileSource. (We could perhaps rename MGLRasterSource and MGLVectorSource to MGLRasterTileSource and MGLVectorTileSource; I don’t have a strong opinion on that.)

Per https://github.com/mapbox/mapbox-gl-native/issues/6584#issuecomment-251744922, the days of a formal “tile set” API are numbered. As it is, the fact that MGLRasterSource and MGLVectorSource use MGLTileSet as a value class for their metadata makes it impossible to elegantly roundtrip source properties (#6584). A case in point is the `MGLTileSet.attribution` property: it should be possible to get a source’s attribution string, because `mbgl::style::Source` has a `getAttribution()` method. But note how it lives on `mbgl::style::Source` rather than `mbgl::style::TileSet`: in order for `MGLTileSet.attribution` to return the attribution string of a source defined in the style, our only two options are:

* To have MGLTileSet maintain a backpointer to MGLSource, which prevents MGLTileSet from being used as a value class; that is, no two MGLSources can share the same MGLTileSet.
* To have `MGLRasterSource.tileSet` and `MGLVectorSource.tileSet` synthesize a dummy MGLTileSet that contains just the attribution string but no other information like URL templates.

Both these options are inelegant, so we should instead follow the lead of mbgl and make its initializers and properties part of every MGLRasterSource and MGLVectorSource instance. Putting them in an MGLTileSource superclass avoids proliferating initializers and duplicating documentation.

There are parallels to the hierarchy of abstract classes under MGLStyleLayer (see MGLForegroundStyleLayer and MGLVectorStyleLayer). If there’s a strong possibility that we might want to organize the concrete source classes under a criteria other than tiling, for example by vector versus raster models, we could have MGLRasterSource and MGLVectorSource share a protocol instead, but I think I like an abstract superclass more than a protocol for this purpose, given that designated initializers are involved.

/ref #5999
/cc @mapbox/ios @ericrwolfe @jfirebaugh

### LINKED ISSUE #7361: Source attribution should be object-oriented
`MGLTileSet`’s `attribution` property – which may soon live on `MGLTileSource`; see #7360 – should be an array of objects containing structured attribution information. We can use the MGLAttributionInfo class being added in #5999 to represent a single attribution item.

The current `MGLTileSet.attribution` property is a single string containing HTML code. This encourages developers to manually craft HTML strings, which can be error-prone. It also means any developer who wants to make sense of a preexisting attribution string would have to parse it as HTML. Done incorrectly, both operations can be avenues for HTML injection. Since #5999 already parses the HTML source code into structured attribution, I don’t think there’s really a use case for working with the raw source code as opposed to, say, an attributed string (which can always be converted back to HTML if needed).

/cc @boundsj @friedbunny

1ec5:
This PR simplifies MGLSource and subclasses, removing parallel code paths and parallel state and making it possible to avoid relying on backpointers as we add getters for more properties in the future.

MGLRasterSource and MGLVectorSource now share a common abstract superclass, MGLTileSource. MGLTileSet has been removed. MGLTileSource is modeled after the common parts of `mbgl::style::RasterSource` and `mbgl::style::VectorSource`. It has initializers that incorporate the parameters of MGLTileSet’s initializers, but it lacks getters for everything but the attribution string, reflecting what’s available in mbgl. MGLTileSet’s properties have been converted into options that can be passed into MGLTileSource’s initializers in a dictionary, similar to the options for MGLShapeSource. MGLTileSource retains an `attributionInfos` property that is now computed based on the underlying mbgl object and represented as an array of MGLAttributionInfo objects. MGLAttributionInfo is now public for this reason.

MGLShape has a new class initializer and instance method that deserialize GeoJSON data into a shape and serialize the shape into GeoJSON data, respectively. The new initializer handles parsing errors gracefully via an NSError out parameter, which causes the bridged method in Swift to be marked with the `throws` keyword.

In an effort to reduce parallel state and parallel code paths in MGLShape, this class no longer has an initializer that accepts GeoJSON data or a property that returns GeoJSON data. The `URL` property’s getter, meanwhile, is computed based on the underlying mbgl object. Developers working with GeoJSON data are now expected to go through the new MGLShape initializer to get an MGLShape representation. Alternatively, if performance is a concern, a local file URL can be passed into the other MGLShapeSource initializer, which has been left intact.

This refactoring exposed a latent bug in the source and style layer classes. The covariant `rawSource` property was incorrectly implemented by both MGLSource and its concrete subclasses. The idea was that the subclasses would merely constrain the property’s type, but they ended up causing shadow ivars to get autosynthesized, so that a subclass could set `_rawSource` directly (rather than going through the `rawSource` setter) but could then get a different value out of the `rawSource` getter.

These changes mean some existing example code that uses the runtime styling API will need to be modified. We should therefore land these changes before the next beta, which we expect to be the final beta. I don’t think the changes will be a showstopper, but we do need to keep an eye out for any leaps of logic that may not be obvious to developers with fresh eyes. In particular, we may want to consider adding a convenience initializer (but not a convenience property) to MGLShapeSource that takes an array of features (and turns it into an MGLShapeCollectionFeature under the hood), since that’s one of the most common use cases for MGLShapeSource. I attempted to bridge this gap in documentation, but I’m concerned that this initializer’s absence may reinforce a misconception that each shape needs its own source or style layer.

A summary of the significant changes:

* [x] Push down MGLRasterSource and MGLVectorSource into MGLTileSource (fixes #7360)
* [x] Replace MGLTileSet with an option dictionary on MGLTileSource initializers (fixes #7360)
* [x] Remove GeoJSON data members from MGLShapeSource to reduce parallel state (working towards #7375)
* [x] Add GeoJSON data (de)serialization methods to MGLShape (fixes #3630)
  * [x] …that handle parsing errors gracefully (fixes #6987)
* [x] Make MGLAttributionInfo public (fixes #7361)

Other changes:

* [x] Audit public and private MGL*Source headers for nullability
* [x] Mark `-[MGLSource init]` as unavailable and audit designated initializers on concrete MGLSource subclasses
* [x] Make MGLTileSet public in the macOS SDK
* [x] Fix covariant `MGL*Source.rawSource` properties to prevent shadowing
* [x] Fix covariant `MGL*StyleLayer.rawLayer` properties to prevent shadowing
* [x] Reduce the number of MGLShape conversion methods
* [x] Add missing headers to ios.xcodeproj and macos.xcodeproj

Before this PR can land:

* [x] MGLFeatures should be serialized as GeoJSON features, not geometries (requires #7339)

/cc @frederoni @incanus @ericrwolfe


1ec5:
The test failures are due to a test I’ve added of the new serialization routine that ensures MGLFeatures are serialized as GeoJSON features, not geometries. This test will fail until #7339 lands. A few merge conflicts will arise from #7339 landing, but they shouldn’t be a problem.

platform/darwin/src/MGLShape.mm
```diff
@@ -1,14 +1,42 @@
...
+        return nil;
+    }
+    
+    try {
+        const auto geojson = mapbox::geojson::parse(string.UTF8String);
+        return MGLShapeFromGeoJSON(geojson);
+    } catch (std::runtime_error &err) {
+        if (outError) {
+            *outError = [NSError errorWithDomain:MGLErrorDomain code:MGLErrorCodeUnknown userInfo:@{
+                NSLocalizedFailureReasonErrorKey: [NSString stringWithFormat:@"%s", err.what()],
```
frederoni: Nit: Could be simplified as `@(err.what())` because it's UTF-8 encoded.

platform/darwin/src/MGLSource.h
```diff
@@ -13,13 +15,15 @@
...
- subclasses of this class. Instead, create instances of `MGLRasterSource`,
- `MGLShapeSource`, and `MGLVectorSource`.
+ subclasses of this class. Instead, create instances of `MGLShapeSource` and the
+ concrete subclasses of `MGLTileSource`.
  */
 @interface MGLSource : NSObject
 
 #pragma mark Initializing a Source
 
+- (instancetype)init __attribute__((unavailable("Use -initWithIdentifier: instead.")));
```
frederoni: Is it not possible use explanation with NS_UNAVAILABLE?

1ec5: Unfortunately it isn't. 😕

boundsj:
https://github.com/mapbox/mapbox-gl-native/pull/7339 (with MGL -> mbgl conversion) has been merged

boundsj:
For clarity, I updated the original description as follows: 

> In an effort to reduce parallel state and parallel code paths in ~~MGLShape~~ MGLShapeSource

platform/darwin/src/MGLRasterSource.mm
```diff
@@ -15,88 +17,77 @@ - (instancetype)initWithRawSource:(mbgl::style::RasterSource *)rawSource NS_DESI
...
-- (instancetype)initWithIdentifier:(NSString *)identifier URL:(NSURL *)url tileSize:(CGFloat)tileSize
-{
-    if (self = [super initWithIdentifier:identifier]) {
-        _URL = url;
-        _tileSize = tileSize;
-        [self commonInit];
-    }
-    return self;
+- (instancetype)initWithIdentifier:(NSString *)identifier configurationURL:(NSURL *)configurationURL {
+    return [self initWithIdentifier:identifier configurationURL:configurationURL tileSize:0];
```
1ec5: Thanks, I meant to follow up on this. The style specification [says](https://www.mapbox.com/mapbox-gl-style-spec/#sources-raster-tileSize) `tileSize` is optional, defaulting to 512. However, via this code path:

https://github.com/mapbox/mapbox-gl-native/blob/27db3bbdb2b0db3c4b0d8c3bd47c9c53634aca13/src/mbgl/style/tile_source_impl.hpp#L43
https://github.com/mapbox/mapbox-gl-native/blob/27db3bbdb2b0db3c4b0d8c3bd47c9c53634aca13/src/mbgl/style/tile_source_impl.cpp#L35
https://github.com/mapbox/mapbox-gl-native/blob/27db3bbdb2b0db3c4b0d8c3bd47c9c53634aca13/src/mbgl/util/mapbox.cpp#L191

a `tileSize` of 512 should cause `@2x` to be hardcoded into the tile URL template regardless of the screen resolution, whereas any other value (including 0) would cause `{ratio}` to be used. I want the `{ratio}` behavior by default, as that optimizes for the screen resolution.

1ec5:
6a1abf5905dffdee1cb62af482fcb44b3b54aaa4 further streamlines MGLShape and subclasses by replacing `-mbglFeature` and `-mbglFeatureCollection` with `-geoJSONObject`. #7339 is no longer recognizable as such, but the fix is still in there thanks to class inheritance and variants.

platform/darwin/src/MGLRasterSource.mm
1ec5: > the raster layer is not visible because of the value 0 here

This demo explicitly applies an opacity function to the layer so that it’s invisible at low zoom levels but becomes visible as you zoom in. The initializer appears to work fine for me.

@jfirebaugh, is it intentional that, by default, we load Retina tiles on non-Retina displays like iPad 2 (due to the default tile size of 512 pixels)?

boundsj:
> 6a1abf5 further streamlines MGLShape and subclasses by replacing -mbglFeature and -mbglFeatureCollection with -geoJSONObject

👍

platform/darwin/src/MGLRasterSource.h
```diff
@@ -1,93 +1,84 @@
...
+ 
+ The only supported values for this option are 256 and 512. When the value is
+ 256, the source requests 1× images on a standard-resolution display and 2×
+ images on a Retina display, and each image is scaled to fit in a square
+ measuring 256 points on each side. When the value is 512, the source requests
+ 2× images regardless of the screen resolution, and each image is displayed in a
+ square measuring 512 <em>pixels</em> on each side. If this option is set to a
+ value other than 256 or 512, the behavior is undefined.
+ 
+ The default value for this option is 256. This option is only applicable to
```
1ec5: This was a difficult option to document. I arrived at this verbose description of the `tileSize` option based on discussions with @jfirebaugh and @tmcw. For simplicity’s sake, I described only two valid values, 256 and 512, leaving all other values undefined.

MGLRasterTileSource applies a default value of 256 for mapbox: tile URLs and 512 for any other tile URL. This differs from the style specification and [GL JS](https://github.com/mapbox/mapbox-gl-js/blob/b5fce3485d8ed21a95bb77622b02fa1a874e3d2f/js/source/raster_tile_source.js#L20), which default to 512 regardless of origin. I believe MGLRasterTileSource’s behavior is justified, because the MGLRasterTileSource default and the underlying option in mbgl are applied only to mapbox: tile URLs, and mapbox: tile URLs are always normalized into Maps API v4 URLs.

Apparently we intend to revise the Maps API in the future to serve 512-point tiles (measuring 1 square kibipixel on a Retina display). A default value of 512 would make sense for requests to that API. However, I’d imagine that we’d map something other than `mapbox://<mapid>` to Maps API v5, in which case MGLRasterTileSource would automatically apply the standard 512 default, or we’d drop support for v4 tiles, which would be a backwards incompatible change anyways.

Meanwhile, at the moment, I believe a default other than 256 for mapbox: URLs would only lead to bugs in application code.

platform/darwin/src/MGLRasterSource.mm
1ec5: Continued in https://github.com/mapbox/mapbox-gl-native/pull/7377#discussion_r92281280.

platform/darwin/src/MGLRasterSource.h
```diff
@@ -1,93 +1,84 @@
...
+ 
+ This number also corresponds to the width and height (measured in points) at
+ which the map displays each tile when the map’s zoom level is an integer. The
+ source scales its images up or down when the map’s zoom level falls between two
+ integers.
+ 
+ This option applies only to tiles hosted by Mapbox and accessed through version
+ 4 of the
+ <a href="https://www.mapbox.com/api-documentation/#maps">Mapbox Maps API</a>,
+ such as through `mapbox://<mapid>` URLs. This option has no effect on tiles
```
jfirebaugh: On the contrary, this option is relevant and does have an effect for third-party tile servers. In fact, most sources which use a third-party tile service will want to specify a value of `256`; if a tile service provides 256 pixel tiles and the source uses a value of `512`, tiles will appear at twice their intended resolution and appear blurry or pixelated.

platform/darwin/src/MGLRasterSource.h
```diff
@@ -1,93 +1,84 @@
...
+ accessed through the
+ <a href="https://www.mapbox.com/api-documentation/#static">Mapbox Static API</a>
+ or hosted on third-party tile servers.
+ 
+ The only supported values for this option are 256 and 512. When the value is
+ 256, the source requests 1× images on a standard-resolution display and 2×
+ images on a Retina display, and each image is scaled to fit in a square
+ measuring 256 points on each side. When the value is 512, the source requests
+ 2× images regardless of the screen resolution, and each image is displayed in a
```
jfirebaugh: This particular behavior is specific to sources using a `mapbox://` URL. For third party tile sources, there is no connection between this option and the behavior of the `{ratio}` placeholder in template URLs.

platform/darwin/src/MGLRasterSource.h
jfirebaugh: Other than the two corrections noted above, I agree with what you say here and with implementing a default of 256 for mapbox: URLs.

platform/darwin/src/MGLRasterSource.h
1ec5: I suppose this means 256 and 512 aren’t the only valid values, either, although 512 is still a magic value with respect to mapbox: URLs.

platform/darwin/src/MGLRasterSource.h
1ec5: The special 512 behavior for mapbox: URLs doesn’t fit into the coherent (and erroneous) model I originally thought it did. Explaining it will only make the option more confusing. I think we can omit this paragraph. mapbox: URLs will have a sensible default behavior, so it’s less likely that someone will run into this edge case.

platform/darwin/src/MGLShape.mm
1ec5: I forgot to make this change before merging. Followed up in d34356ce390bc5f995e2c8e29deaa2a6bec17a10.

PR #7386, opened 2016-12-12
[Qt] Do not install/link with ICU if WITH_QT_I18N is set

brunoabinader:
As part of the arrangements for getting Mapbox GL native as a 3rd party submodule in Qt framework, we are optionally removing ICU support via `WITH_QT_I18N` compile flag.

This PR:
- Moves bidi.cpp default implementation to `platform/default/bidi.cpp` - all platforms explicitly compiles it by default as part of `mbgl-core` target.
- Added `BiDiImpl` to hide ICU data types from BiDi interface.
- Provide a Qt-only line breaking implementation via `platform/qt/src/bidi.cpp` - arabic shaping is not supported atm.


platform/qt/src/bidi.cpp
```diff
@@ -0,0 +1,63 @@
...
+std::u16string applyArabicShaping(const std::u16string& input) {
+    QString utf16string = QString::fromStdU16String(input);
+    return utf16string.toStdU16String();
+}
+
+ProcessedBiDiText::ProcessedBiDiText(BiDi& p_bidi) : bidi(p_bidi) {
+}
+
+void ProcessedBiDiText::mergeParagraphLineBreaks(std::set<int32_t>& lineBreakPoints) {
+    QTextBoundaryFinder finder(QTextBoundaryFinder::Line, bidi.impl->string);
```
ChrisLoer: Looking at the documentation for QTextBoundaryFinder::Line ("Finds possible positions for breaking the text into multiple lines."), it seems like this will insert a line break everywhere a line break is possible? Although maybe I'm misreading, because I expect that would have broken some tests...

The intention of this method is only to insert line breaks where the ICU BiDi algorithm absolutely requires them (for "block separators"), in case the calling code for whatever reason decided not to explicitly break at one of them. The common block separator is "\n", but there are more exotic ones like U+001C "INFORMATION SEPARATOR FOUR". For stubbing this code out on Qt, I don't think we need to implement any of this behavior (beyond maybe what's necessary to pass tests).

brunoabinader: Thanks - we can maybe revisit it once we get unit tests to verify.

PR #7388, opened 2016-12-12
Only build active architecture in Xcode

kkaefer:
Trying to land #7131 as separate PRs to make the changes easier to reason about.

This changes the *default* setting to only build the active architecture (i.e. they for the device you connected to your Mac). This cuts build time in half since we don't need to build the other architecture when iterating.

On CI, and for `make ipackage/iframework`, we are using the `package.sh` script which already forces `ONLY_ACTIVE_ARCH=NO` and ensures that both architectures get built, so we can discover build-time issues for different architectures. Similarly, running other iOS-related make targets that build for Simluator already force `ONLY_ACTIVE_ARCH=YES`. This means that this change *only* affects what is being built when you use Xcode to manually compile apps; automated builds are unaffected by this change.


kkaefer:
Also note that `ONLY_ACTIVE_ARCH=YES` *is* already the current setting for Debug builds, so this PR brings the setting for `mbgl-core` in line with the settings of the targets in our checked-in projects.

Issue #2354, opened 2015-09-17
Android PNG/JPEG reader


ljbade:
From #2037 we should implement `PNGReader` and `JPEGReader` for Android using Java framework.

Allows us to drop `libpng` and `libjpeg` dependency.

ljbade:
We want to use [BitmapFactory](http://developer.android.com/reference/android/graphics/BitmapFactory.html) and [BitmapRegionDecoder](http://developer.android.com/reference/android/graphics/BitmapRegionDecoder.html)

ljbade:
Working on it in https://github.com/mapbox/mapbox-gl-native/tree/2354-android-image

incanus:
We need to ship 0.1.0 first before working on new functionality / 0.2.0 :point_right: https://github.com/mapbox/mapbox-gl-native/milestones/android-v0.1.0

jfirebaugh:
Fixed in #7828.

PR #7329, opened 2016-12-12
Android: Replace libpng/libjpeg-turbo/libwebp with BitmapFactory

kkaefer:
In https://github.com/mapbox/mapbox-gl-native/issues/2354, we captured that we want to replace these libs with a Java-based version. Here's how this will work:

* [x] Refactor `util::Image` to hide its implementation, in particular the `data` member
* [ ] When compiling for Android, add an additional constructor that takes `JNIEnv *env, jobject jbitmap`, and then uses [`libjnigraphics`](https://developer.android.com/ndk/reference/group___bitmap.html) to expose the bitmap information, and handle JNI object lifetime
* [ ] Create an Android-specific implementation of `util::decodeImage()` that uses the JNI to call [BitmapFactory::decodeByteArray](https://developer.android.com/reference/android/graphics/BitmapFactory.html#decodeByteArray(byte[],%20int,%20int,%20android.graphics.BitmapFactory.Options)) for decompressing + premultiplying the image
* [ ] Create an Android-specific version of `util::encodePNG()` that uses the JNI/Bitmap/BitmapFactory to create a Java Bitmap + compress it to PNG

/cc @mapbox/android

### DIFF
--- a/include/mbgl/util/image.hpp
+++ b/include/mbgl/util/image.hpp
@@ -10,8 +10,7 @@
 namespace mbgl {
 
 enum class ImageAlphaMode {
-    Unassociated,
-    Premultiplied,
+    Premultiplied, // Premultiplied RGBA
     Exclusive, // Alpha-channel only
 };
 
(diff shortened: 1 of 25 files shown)


kkaefer:
superseded by https://github.com/mapbox/mapbox-gl-native/issues/7828

PR #7392, opened 2016-12-12
[docs]: Add segment.io (and thus GA) to iOS doc theme

tmcw:


1ec5:
This template is also used for the docset that's bundled with the SDK for offline viewing. Can we find a way to inject this into the online version only? We have a deployment script for the /ios-sdk site that could do this on the fly.

friedbunny:
@1ec5 I’m not sure it’s worth the additional complexity — chances are that people are going to be online during development (and the failure case is not catastrophic here, if offline). If offline use wouldn’t provide meaningful metrics, then I think we should go ahead and exclude.

1ec5:
I would prefer if we exclude this markup from the generated local documentation. The most straightforward way to do that would be to allow an include file to be passed in on the command line (or in a script). #7396 plumbs that include file through to jazzy’s existing `<head>` injection option.

friedbunny:
#7396 looks like a workable improvement over hardcoding.

tmcw:
Closing in favor of #7396

Issue #7116, opened 2016-11-18
myLocationViewSettings.setBackgroundTintColor cannot be #00000000


ghengeveld:
**Platform:** Android 5.0
**Mapbox SDK version:** 4.2.0.beta.3

### Steps to trigger behavior

 1. Use `MyLocationViewSettings.setForegroundDrawable` to show a custom drawable at the user's location
 2. Use `MyLocationViewSettings.setBackgroundTintColor` with #00000000 (fully transparent black)

In `OnMapReadyCallback.onMapReady`:
```java
Drawable walkingUserMarker = ContextCompat.getDrawable(context, R.drawable.marker_user_walk);
MyLocationViewSettings myLocationViewSettings = mapboxMap.getMyLocationViewSettings();
myLocationViewSettings.setAccuracyAlpha(0);
myLocationViewSettings.setForegroundDrawable(walkingUserMarker, null);
myLocationViewSettings.setBackgroundTintColor(ContextCompat.getColor(context, R.color.transparent));
```

In colors.xml:
```xml
<color name="transparent">#00000000</color>
```

Setting the color to #00ffffff avoids this issue and yields the expected result.

### Expected behavior

Only the marker drawable should be rendered at the user's location, nothing else.

### Actual behavior

The drawable appears with a light gray dot below it. This appears to be the default background tint.

It would be nice to be able to use `setBackgroundTintColor(null)` to disable the background altogether.

cammace:
This is due to use [checking if it is transparent](https://github.com/mapbox/mapbox-gl-native/blob/master/platform/android/MapboxGLAndroidSDK/src/main/java/com/mapbox/mapboxsdk/maps/widgets/MyLocationView.java#L176) or not. We should rethink this check or expose a setVisibility method for the foreground/background images. @tobrun what do you think?

tobrun:
Makes sense @cammace, feel free to run with this

PR #7393, opened 2016-12-12
Allows for setting the user location icon to transparent

cammace:
Closes https://github.com/mapbox/mapbox-gl-native/issues/7116

Set the default user location color to be `mapbox_blue` and not transparent and allow for any color int value below 1 to tint user location foreground and background. We might be able to get rid of the color check `if (color < 1)`?


tobrun:
>  We might be able to get rid of the color check if (color < 1)?

:+1:

cammace:
Removed if check.

PR #7396, opened 2016-12-13
Custom head markup in generated documentation

1ec5:
The `idocument` and `xdocument` make rules now accept an `HTMLHEAD` variable that specifies a file to be included after the favicon in the `<head>` of every page in the generated documentation.

This is an alternative to #7392 that avoids hard-coding specific markup in the templates. Instead, the site script that pulls in this generated docset can specify its own markup.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -165,7 +165,7 @@ xpackage: $(MACOS_PROJ_PATH)
 
 .PHONY: xdocument
 xdocument:
-	OUTPUT=$(OUTPUT) ./platform/macos/scripts/document.sh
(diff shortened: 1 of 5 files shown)


1ec5:
After discussing this PR with @tmcw, we decided it would be better for the /ios-sdk site to maintain its own theme entirely separately from what’s checked into this repository, especially as we add site navigation and other things that don’t necessarily make sense locally or in Dash. So the make rule would take a path to a theme directory instead of a head include file.

1ec5:
Closing in favor of #7420.

PR #5170, opened 2016-05-27
[android] Add hole handling to PolygonOptions

nealsanche:
- Added holes to the Java SDK, JNI, and renderer
- Multiple holes are supported
- Seems stable even if no holes are specified
- Added code to add holes in the Polygon demo app which adds 3 holes to the star over portland.

PR #7397, opened 2016-12-13
Polygon holes

### LINKED ISSUE #1729: add support for multipolygons and polygon interior holes

1ec5:
A slapdash attempt at reviving #5170 to add support for holes in `Polygon` and `PolygonOptions`. I resolved the conflicts without the aid of an IDE, so this PR needs close scrutiny – there’s probably quite a bit of copy 🍝 in here.

Fixes #1729.

### DIFF
--- a/platform/android/MapboxGLAndroidSDK/src/main/java/com/mapbox/mapboxsdk/annotations/Polygon.java
+++ b/platform/android/MapboxGLAndroidSDK/src/main/java/com/mapbox/mapboxsdk/annotations/Polygon.java
@@ -2,18 +2,48 @@ package com.mapbox.mapboxsdk.annotations;
 
 import android.graphics.Color;
 
+import com.mapbox.mapboxsdk.geometry.LatLng;
 import com.mapbox.mapboxsdk.maps.MapboxMap;
+
+import java.util.ArrayList;
+import java.util.List;
 
 /**
  * Polygon is a geometry annotation that's a closed loop of coordinates.
  */
 public final class Polygon extends MultiPoint {
 
+  private List<List<LatLng>> holes;
   private int fillColor = Color.BLACK; // default fillColor is black
   private int strokeColor = Color.BLACK; // default strokeColor is black
 
   Polygon() {
     super();
+    holes = new ArrayList<>();
+  }
+
+  /**
+   * Returns a copy of the holes.
+   *
+   * @return A {@link List} of holes.
+   */
+  public List<List<LatLng>> getHoles() {
+    return new ArrayList<>(holes);
+  }
+
+  /**
+   * Sets the holes of this polygon. This method will take a copy of the holes, so further
+   * mutations to holes will have no effect on this polygon.
+   *
+   * @param points A {@link List} {@link List}s of {@link LatLng} points making up the holes.
+   */
+  public void setHoles(List<? extends List<LatLng>> holes) {
+    this.holes = new ArrayList<>(holes);
+    update();
+  }
+
+  void addHole(List<LatLng> hole) {
+    holes.add(hole);
   }
 
   /**
@@ -26,7 +56,7 @@ public final class Polygon extends MultiPoint {
   }
 
   /**
-   * Get the color fo the stroke of the polygon.
+   * Get the color for the stroke of the polygon.
    *
    * @return The color of the stroke.
    */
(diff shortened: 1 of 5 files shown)


platform/android/MapboxGLAndroidSDK/src/main/java/com/mapbox/mapboxsdk/annotations/Polygon.java
```diff
@@ -2,18 +2,48 @@
...
 
+import java.util.ArrayList;
+import java.util.List;
+
 /**
  * Polygon is a geometry annotation that's a closed loop of coordinates.
  */
 public final class Polygon extends MultiPoint {
 
+    private List<List<LatLng>> holes;
```
ivovandongen: Could you make this a `final` list and initialize in declaration?

platform/android/MapboxGLAndroidSDK/src/main/java/com/mapbox/mapboxsdk/annotations/Polygon.java
```diff
@@ -2,18 +2,48 @@
...
+    }
+
+    /**
+     * Sets the holes of this polygon. This method will take a copy of the holes, so further
+     * mutations to holes will have no effect on this polygon.
+     *
+     * @param points A {@link List} {@link List}s of {@link LatLng} points making up the holes.
+     */
+    public void setHoles(List<? extends List<LatLng>> holes) {
+        this.holes = new ArrayList<>(holes);
```
ivovandongen: Can just update the member list with the collection, so the reference is not exposed.

platform/android/src/jni.cpp
```diff
@@ -757,7 +758,26 @@ jni::jarray<jlong>* nativeAddPolygons(JNIEnv *env, jni::jobject* obj, jlong nati
...
+
+        jni::jobject* holes = jni::GetField<jni::jobject*>(*env, polygon, *polygonHolesId);
+        NullCheck(*env, holes);
+        jni::jarray<jni::jobject>* jarrayHoles =
+            reinterpret_cast<jni::jarray<jni::jobject>*>(jni::CallMethod<jni::jobject*>(*env, holes, *listToArrayId));
+        NullCheck(*env, jarrayHoles);
+
+        std::size_t size = jni::GetArrayLength(*env, *jarrayHoles);
+        for (std::size_t j = 0; j < size; j++) {
+            jni::jobject* hole = reinterpret_cast<jni::jobject*>(jni::GetObjectArrayElement(*env, *jarrayHoles, j));
```
ivovandongen: This cast is redundant, `jni::GetObjectArrayElement` returns `jni::jobject*` already

1ec5:
Thanks for taking a look!

> I don't see any holes in the example.

I think the example originally worked, but now the star's outer ring has been moved to a static member of the class instead of a local array. For the purposes of this example, I think we'll be able to define the hole similarly to the outer ring instead of trying to define the hole dynamically.

platform/android/MapboxGLAndroidSDK/src/main/java/com/mapbox/mapboxsdk/annotations/Polygon.java
1ec5: You mean `this.holes = holes`? I think the idea was to make a copy of the list, similar to `MultiPoint.setPoints()`.

platform/android/src/jni.cpp
1ec5: Fixed in 2ec060543899367c3b1f9026117488d8c133ec10.

1ec5:
I modernized the example activity, though I haven’t tried it yet.

zugaldia:
@1ec5 Polygon looks different, but I still see no holes:

screenshot_20170202-150032

1ec5:
Thanks @zugaldia. I need to actually get around to setting up a development environment. GitHub comments don't make for a good debugger. 😅

zugaldia:
@1ec5 totally :) wanna work on that today to benefit from the small delta in our coordinates?

caHarkness:
I see that there's a pull request open:
`add support for multipolygons and polygon interior holes #1729`

Does this mean that in a future version, it is guaranteed that we will have interior holes on Android?  I've been following this issue for a while and have seen iOS get a lot of attention on the subject, but I am more interested in Android.

1ec5:
#1729 is the bug report; this is the PR that would fix it for Android. The issue was fixed for iOS and macOS in #5110, which also involved adding support in the core C++ code for polygon holes. This PR plumbs that support through JNI to the Android SDK and adds an activity to demonstrate the feature. Unfortunately, the activity currently demonstrates that the JNI implementation is broken. @Guardiola31337 is currently assigned to take a look. We’re still scoping out what the next release will contain, but this PR is on our radar.

1ec5:
#8557 landed instead.

PR #7408, opened 2016-12-13
Cherry pick 7242 into iOS 3.4.0 release branch

boundsj:


boundsj:
Noting that this cherry picks: 

- [core] Swap order of Style::recalculate and Style::relayout https://github.com/mapbox/mapbox-gl-native/pull/7242/commits/904f1a319b9dc39d1f42670b3af5b1a37c58b1d4

but not:

- [tests] Implement "wait" command with Map#render https://github.com/mapbox/mapbox-gl-native/pull/7242/commits/768837a9c1ac2bf1f75cc3439408cf87ae199113
- [core] Ensure a non-zero source tile cache size https://github.com/mapbox/mapbox-gl-native/pull/7242/commits/2d323211af54499d5c822b8e45d7415bf92112f0

from: https://github.com/mapbox/mapbox-gl-native/pull/7242

I don't think 2d323211af54499d5c822b8e45d7415bf92112f0 is critical for iOS (Darwin) so we can just pick it up in the eventual merge back to master.

cc @1ec5 @jfirebaugh fyi @ivovandongen

1ec5:
> I don't think 2d323211af54499d5c822b8e45d7415bf92112f0 is critical for iOS (Darwin) so we can just pick it up in the eventual merge back to master.

FYI, I think that change would affect small map views in both iOS and macOS. (The calculation appears to depend on the viewport size, not the screen size.)

PR #7410, opened 2016-12-13
[Qt] Implement updateSource

tmpsantos:
It will create or update a source. Right now only GeoJSON sources are mutable and can be updated using this method.


tmpsantos:
:eyes: @brunoabinader 

This fixes the long standing issue of associating of binding a GeoJSON to the `data` field of a GeoJSON source `MapParameter` not updating the map.

PR #7420, opened 2016-12-14
Make theme configurable

tmcw:
This allows runners of `document.sh` to use their own custom, local theme. It uses the existing theme by default.

### DIFF
--- a/platform/ios/scripts/document.sh
+++ b/platform/ios/scripts/document.sh
@@ -32,7 +32,8 @@ mkdir -p ${OUTPUT}
 
 cp platform/ios/screenshot.png "${OUTPUT}"
 
-DEFAULT_THEME='platform/darwin/docs/theme'
+DEFAULT_THEME="platform/darwin/docs/theme"
+THEME=${JAZZY_THEME:-$DEFAULT_THEME}
 
 jazzy \
     --config platform/ios/jazzy.yml \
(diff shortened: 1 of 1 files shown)


platform/ios/scripts/document.sh
```diff
@@ -32,6 +32,9 @@ mkdir -p ${OUTPUT}
 
 cp platform/ios/screenshot.png "${OUTPUT}"
 
+DEFAULT_THEME="platform/darwin/docs/theme"
+THEME=${JAZZY_THEME:-$DEFAULT_THEME}
```
1ec5: This script is usually run by a make rule, not by itself, so the `THEME` variable needs to be plumbed through to here via the Makefile. This PR will end up looking very similar to #7396. (Preferably, the macOS script and make rule are kept in sync with this script for predictability’s sake.)

tmcw: This already works in full integration, from the `ios-sdk` side, since the variable that's plumbed through is `JAZZY_THEME` rather than theme - I'm renaming it to `THEME` here when I prevent default, just to make it easier to debug in the future.

1ec5: Interesting, I figured that the Makefile would still have to pass through this variable, just like it explicitly passes through `OUTPUT`. However, you’re right that the variable works regardless with both `make idocument JAZZY_THEME=…` and `JAZZY_THEME=… make idocument`.

Issue #7342, opened 2016-12-08
[iOS]Missing markers after removeAnnotations/addAnnotations is called


andrewstay:
**Platform:**iOS
**Mapbox SDK version:** ios-v3.4.0-beta.4

### Steps to trigger behavior

1. Create an instance of MGLMapView
2. Add annotations with random data
3. Provide view for the created annotations via
```-[MGLMapViewDelegate mapView:viewForAnnotation:]```
4. Observe views are correctly instantiated and displayed in the map.
5. Remove existing annotations as following:
```[self.mapView removeAnnotations:self.mapView.annotations];```
6. Instantiate new random annotations and set them up into the instance of the map view
```[self.mapView addAnnotations:annotations];```

### Expected behavior
All newly added annotations which fit the visible coordinates should be displayed.
### Actual behavior
The annotations which are located close the the edge of the visible bounds are not displayed. Only few annotations are visible in the map.

Please review the video  and the test application attached.
Please read "readme.txt" for details on how to run the test application.
[TestMapbox.zip](https://github.com/mapbox/mapbox-gl-native/files/638726/TestMapbox.zip)

https://www.dropbox.com/s/gnka3kf89bw0rup/AnnotationsIssue.mov?dl=0

1ec5:
Fixed in #7423 on the v3.4.0 release branch.

boundsj:
Also, thanks so much @andrewstay for the detailed description and attaching the repro app. That was very helpful! [Beta 5](https://github.com/mapbox/mapbox-gl-native/releases/tag/ios-v3.4.0-beta.5) is available so please test again with that version.

PR #7423, opened 2016-12-14
[ios] Fix dequeue view variable scope

boundsj:
Fixes https://github.com/mapbox/mapbox-gl-native/issues/7342

cc @1ec5 @frederoni

### DIFF
--- a/platform/ios/src/MGLMapView.mm
+++ b/platform/ios/src/MGLMapView.mm
@@ -4766,15 +4766,7 @@ public:
         }
     }
 
-    CGFloat largestHeight = _largestAnnotationViewSize.height;
-    CGFloat largestWidth = _largestAnnotationViewSize.width;
-    CGFloat viewHeight = CGRectGetHeight(self.frame);
-    CGFloat viewWidth = CGRectGetWidth(self.frame);
-    CLLocationCoordinate2D southwestCoordinate = [self convertPoint:{-largestWidth, viewHeight + largestHeight}
-                                               toCoordinateFromView:self];
-    CLLocationCoordinate2D northeastCoordinate = [self convertPoint:{viewWidth + largestWidth, -largestHeight}
-                                               toCoordinateFromView:self];
-    MGLCoordinateBounds coordinateBounds = {southwestCoordinate, northeastCoordinate};
+    MGLCoordinateBounds coordinateBounds = [self convertRect:viewPort toCoordinateBoundsFromView:self];
 
     // Enqueue (and move if required) offscreen annotation views
     for (id<MGLAnnotation> annotation in offscreenAnnotations)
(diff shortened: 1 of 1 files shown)


platform/ios/src/MGLMapView.mm
```diff
@@ -4799,7 +4799,7 @@ - (void)updateAnnotationViews
                 (coordinate.longitude < upperLeftCoordinate.longitude || coordinate.longitude > lowerRightCoordinate.longitude))
             {
                 CGRect adjustedFrame = annotationView.frame;
-                adjustedFrame.origin.x = -CGRectGetWidth(adjustedFrame) * 2.0;
+                adjustedFrame.origin.x = -CGRectGetWidth(adjustedFrame) * 10.0;
```
boundsj: Unrelated but taking this opportunity to increase the size of this buffer.

platform/ios/src/MGLMapView.mm
```diff
@@ -4765,14 +4765,17 @@ - (void)updateAnnotationViews
...
+
+    CGFloat largestHeight = _largestAnnotationViewSize.height;
+    CGFloat largestWidth = _largestAnnotationViewSize.width;
+    CGFloat viewHeight = CGRectGetHeight(self.frame);
+    CGFloat viewWidth = CGRectGetWidth(self.frame);
+    CLLocationCoordinate2D southwestCoordinate = [self convertPoint:{-largestWidth, viewHeight + largestHeight}
+                                               toCoordinateFromView:self];
+    CLLocationCoordinate2D northeastCoordinate = [self convertPoint:{viewWidth + largestWidth, -largestHeight}
+                                               toCoordinateFromView:self];
+    MGLCoordinateBounds coordinateBounds = {southwestCoordinate, northeastCoordinate};
```
1ec5: This code could use `UIEdgeInsetsInsetRect()` to apply a slop area around the map view’s bounds, then convert it using `-convertRect:toCoordinateBoundsFromView:`.

platform/ios/src/MGLMapView.mm
```diff
@@ -4795,17 +4798,16 @@ - (void)updateAnnotationViews
...
-                adjustedFrame.origin.x = -CGRectGetWidth(adjustedFrame) * 2.0;
-                annotationView.frame = adjustedFrame;
-                [self enqueueAnnotationViewForAnnotationContext:annotationContext];
+                annotationView.center = [self convertCoordinate:annotationContext.annotation.coordinate toPointToView:self];
             }
             else
             {
-                annotationView.center = [self convertCoordinate:annotationContext.annotation.coordinate toPointToView:self];
+                CGRect adjustedFrame = annotationView.frame;
+                adjustedFrame.origin.x = -CGRectGetWidth(adjustedFrame) * 10.0;
```
1ec5: Can you explain this formula?

1ec5: If the concern is that a view might have a layer that overshoots the official bounds of the view and would thus peek out from the top edge of the map view, you could perhaps move the view so that its `layer.presentationLayer.frame` lies outside the map view.

boundsj:
Thanks @1ec5

PR #7427, opened 2016-12-14
`make run-android-core-test-arm-v8` doesn't work

kkaefer:
Running this command doesn't work since it's using `dalvikvm32` when trying to build a JAR file that contains 64 bit libraries. When switching to `dalvikvm64`, I'm getting stuck at this error:

```
CANNOT LINK EXECUTABLE "dalvikvm64": "/system/lib/liblog.so" is 32-bit instead of 64-bit
```


kkaefer:
Changes `lib` => `lib64`, and `dalvikvm32` => `dalvikvm64` on 64-bit platforms (`arm-v8` and `x86-64`)

PR #7431, opened 2016-12-14
Add -Wl,-bind_at_load to node.js bindings on macOS

### LINKED ISSUE #7332: Restore -Wl,-bind_at_load to node OS X build
We used to set `-Wl,-bind_at_load` for the node OS X build, so that missing symbols would be caught at launch rather than when they are first needed at runtime. Without this flag, if unit tests fail to cover a particular code path, missing symbols can be overlooked.

It looks like this flag [got lost](https://github.com/mapbox/mapbox-gl-native/commit/692fe1f3ffc8f4364b39c14aa7d90cec2ff5c0a6#diff-21e418851ae48f9b11907837add9e39aL91) in the switch to CMake. We should restore it.

cc @kkaefer

kkaefer:
Fixes #7332

### DIFF
--- a/platform/macos/config.cmake
+++ b/platform/macos/config.cmake
@@ -168,5 +168,6 @@ macro(mbgl_platform_node)
     target_link_libraries(mbgl-node
         PRIVATE "-framework Foundation"
         PRIVATE "-framework OpenGL"
+        PRIVATE "-Wl,-bind_at_load"
     )
 endmacro()


kkaefer:
Not really a good way to test this on CI. I enabled the call to `platform::showColorDebugImage` as described in https://github.com/mapbox/mapbox-gl-native/issues/7332#issuecomment-267030215 and saw the node bindings fail to load with this message (which is what we want):

```
module.js:597
  return process.dlopen(module, path._makeLong(filename));
                 ^

Error: dlopen(.../lib/mapbox-gl-native.node, 1): Symbol not found: __ZN4mbgl8platform19showColorDebugImageENSt3__112basic_stringIcNS1_11char_traitsIcEENS1_9allocatorIcEEEEPKcmmmm
  Referenced from: .../lib/mapbox-gl-native.node
  Expected in: flat namespace
 in .../lib/mapbox-gl-native.node
    at Error (native)
    at Object.Module._extensions..node (module.js:597:18)
    at Module.load (module.js:487:32)
    at tryModuleLoad (module.js:446:12)
    at Function.Module._load (module.js:438:3)
    at Module.require (module.js:497:17)
    at require (internal/module.js:20:19)
    at Object.<anonymous> (.../platform/node/index.js:5:12)
    at Module._compile (module.js:570:32)
    at Object.Module._extensions..js (module.js:579:10)
Program ended with exit code: 1
```

Issue #1278, opened 2015-04-15
Scale control


1ec5:
Apple Maps (but not `MKMapView` itself) shows a scale (in both miles and kilometers) in the top-left corner of the map view while you pinch to zoom in or out. A scale would be a useful feature to bake into `MGLMapView`, either off by default or off except while you pinch. It would help to give context, especially since GL doesn’t snap the zoom level to increments and allows very high zoom levels.

1ec5:
Correction: `MKMapView.showsScale = YES` turns on a persistent scale at the top-left corner.

Bluezen:
Hi @1ec5 ! Do you know if there is a plan to add this feature in an upcoming release?

If not, how would you recommend me to proceed with the current iOS SDK 3.2.3 if I wanted to show a simple scale ? 
Adding a custom UIView on top of the MKMapView which, on every Map position changes (`-mapViewRegionIsChanging:` `-mapView:regionDidChangeAnimated:` ), uses `visibleCoordinateBounds` (or something else ?) to extrapolate and update its display?

1ec5:
We haven't scheduled this feature into an upcoming release yet, but we aren't opposed to getting the feature in if someone steps up with an implementation.

In the meantime, you have the right idea with the delegate methods (although you might find `-mapViewDidFinishRenderingFrame:fullyRendered:` to be more predictable). In terms of calculating the width of the scale's indicators, I'd use `-[MGLMapView metersPerPointAtLatitude:]`, passing in the current center latitude.

If you come up with an implementation that you feel would be appropriate for the SDK, we'd be happy to consider your PR.

Bluezen:
Thanks for pointing out `-[MGLMapView metersPerPointAtLatitude:]` and `mapViewDidFinishRenderingFrame:fullyRendered` !

Okay, if I am satisfied with my implementation I will propose a PR.

friedbunny:
Somewhat relatedly, here’s a quick way to get the geographic distance spanned by the map view:

```swift
func mapView(mapView: MGLMapView, regionDidChangeAnimated animated: Bool) {
     let screenWidthInPoints = Double(mapView.bounds.size.width)
     let metersPerPoint = mapView.metersPerPointAtLatitude(mapView.centerCoordinate.latitude)
     let screenDistanceInMeters = Int(screenWidthInPoints * metersPerPoint)

[...]

Issue #3331, opened 2015-12-17
MGLDistanceFormatter


1ec5:
MapKit provides a handy subclass of NSFormatter, MKDistanceFormatter, for turning `CLLocationDistance`s into display-appropriate strings with units. The default unit is selected based on the system locale. Developers shouldn’t have to import MapKit for this functionality.

This is a `starter-task` because it has no dependencies on MGLMapView or mbgl.

/ref #949
/cc @friedbunny

friedbunny:
Good idea. :bulb:

cph2117:
Hi guys,
I spent a little time tonight and came up with something along the lines of a Mapbox version of MKDistanceFormatter.  Unfortunately I did it in Swift before realizing that your entire iOS SDK is in Obj-C.  

I'm sure you guys have your own code standards and way of writing things, but if you think that what I wrote could serve as a jumping off point for the real MGLDistanceFormatter, then I'd be happy to take what I wrote and convert it into Obj-C.  Here is the Swift code :)

Cheers,
Connor

```swift
import Foundation
import CoreLocation

enum MGLDistanceFormatterUnits : UInt {
    case Metric
    case Imperial
    case ImperialWithYards
}
enum MGLDistanceFormatterUnitStyle : UInt {
    case Abbreviated
    case Full
}

//Using ISO 3166-1 --> https://www.iso.org/obp/ui/
enum MGLImperialCountries: String {
    case US //USA
    case GB //United Kingdom
    case MM //Myanmar
    case BS //Bahamas
    case BZ //Belize
}

public class MGLDistanceFormatter {
    
    private let METERS_PER_KM = 1000.0
    private let METERS_PER_MILE = 1609.34
    private let FEET_PER_METER = 3.28084
    private let FEET_PER_MILE = 5280.0
    private let FEET_PER_YD = 3.0
    private let YARDS_PER_METER = 1.09361
    
    private let ABB_MILES = "mi"
    private let ABB_METERS = "m"
    private let ABB_YARDS = "yds"
    private let ABB_KILOMETERS = "km"
    private let ABB_FEET = "ft"
    private let MILES = "miles"
    private let METERS = "meters"
    private let KILOMETERS = "kilometers"
    private let FEET = "feet"
    private let FOOT = "foot"
    private let YARDS = "yards"
    
    private var locale: NSLocale!
    private var localeDeterminedUnits: MGLDistanceFormatterUnits!

    var units: MGLDistanceFormatterUnits
[...]

PR #7432, opened 2016-12-14
Show scale control

frederoni:
WIP #1278 

- [ ] Expose a showsScale property on MGLMapView
- [ ] Show scale while zooming by default
- [ ] Switch units based on regional settings

screenshot 2016-12-14 16 13 12

### DIFF
--- a/platform/ios/src/MGLScaleControlView.m
+++ b/platform/ios/src/MGLScaleControlView.m
@@ -1,0 +1,48 @@
+#import "MGLScaleControlView.h"
+
+@interface MGLScaleControlView()
+@property (nonatomic, assign) double zoomLevel;
+@property (nonatomic, assign) CLLocationDistance metersPerPoint;
+@property (nonatomic) NSArray *labels;
+@property (nonatomic) NSArray *bars;
+@property (nonatomic) UILabel *debugLabel;
+@end
+
+@implementation MGLScaleControlView
+
+- (void)updateWithZoomLevel:(double)zoomLevel metersPerPoint:(CLLocationDistance)metersPerPoint {
+    self.zoomLevel = zoomLevel;
+    self.metersPerPoint = metersPerPoint;
+}
+
+- (void)setZoomLevel:(double)zoomLevel {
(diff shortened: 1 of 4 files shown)


platform/ios/src/MGLScaleControlView.m
```diff
@@ -0,0 +1,48 @@
...
+
+- (CGSize)intrinsicContentSize {
+    return CGSizeMake(100, 10);
+}
+
+- (UILabel *)debugLabel {
+    if (!_debugLabel) {
+        _debugLabel = [[UILabel alloc] init];
+        _debugLabel.font = [UIFont systemFontOfSize:8];
+        [self addSubview:_debugLabel];
```
1ec5: A label means you'd have to manage constraints internal to this view. That may be a good idea, but consider whether it might be easier to override -drawInRect: and use Core Graphics and attributed string drawing methods to accomplish the same effect, since you don't need to worry about line wrapping.

frederoni: I prefer to use -drawInRect: when I want instant rendering instead of default animation between layouts otherwise I prefer -layoutSubviews. Working directly with the views also improves readability but I haven't really decided yet. The logic should be the same so we can switch back if we don't get the expected behavior without much fuss.

platform/ios/src/MGLScaleControlView.m
```diff
@@ -3,13 +3,25 @@
...
 @end
 
 @implementation MGLScaleControlView
 
+static const NSUInteger TEMP_NUMBER_OF_BARS = 2;
+static const CGFloat BASE_WIDTH = 100;
+
+- (instancetype)init {
+    if (self == [super init]) {
+        _formatter = [[NSLengthFormatter alloc] init];
```
1ec5: We should use MGLDistanceFormatter once #3331 is implemented. NSLengthFormatter is suboptimal because, when the system is set to U.S. units, it tends to use yards instead of feet or fractional miles, whereas Americans find yards difficult to relate to (except in football).

frederoni:
Closing in favor of #7631

Issue #6232, opened 2016-09-01
Clean Android code to comply with checkstyle rules


cammace:
Now that checkstyle's been added, the code needs to be cleaned up before integration with bitrise and failing builds. Using this ticket to track progress. Follow this [new wiki guide](https://github.com/mapbox/mapbox-gl-native/wiki/Setting-up-Mapbox-checkstyle) to setup Android Studio for easy formatting. 

Related ticket: #6109 

cc: @mapbox/android

tobrun:
Merged with #7442, closing

PR #7442, opened 2016-12-14
Adds checkstyle to CI

cammace:
Picking https://github.com/mapbox/mapbox-gl-native/issues/6232 back up, this adds gradle task and integrates with Bitrise. Just produces warning for now since there's still many violations in the code. I placed the task up pretty high in the script so that later when we start failing builds, it happens before we run our test.


tobrun:
> Just produces warning

Can we just make it fail instead and fix up the code as part of this commit?

PR #7444, opened 2016-12-15
[core] port matrix and antialiasing cleanup

### LINKED ISSUE #7439: port matrix creation cleanup from -js
port https://github.com/mapbox/mapbox-gl-js/pull/3790.

> This cleans up a bunch of matrix and antialiasing code.
>
> - removes the confusing altitude and replaces it with fov
> - fixes text blurriness at non-default FOVs
> - removes the hack added in #3740 and adds a different fix
> - makes pitched line antialiasing clearer and removes lineStretch and lineAntialiasingMatrix

ansis:
fix #7439

ports https://github.com/mapbox/mapbox-gl-js/pull/3790

> This cleans up a bunch of matrix and antialiasing code.
> 
> - removes the confusing altitude and replaces it with fov
> - fixes text blurriness at non-default FOVs
> - removes the hack added in #3740 and adds a different fix
> - makes pitched line antialiasing clearer and removes lineStretch and lineAntialiasingMatrix

TODO:
I removed `altitude` from `CustomLayerRenderParameters` and added `fieldOfView`. Is this considered an external interface / breaking change? Should I remove it from the platform bindings as well? Or should I calculate it and add it? The old altitude value is just `0.5 / tan(fov / 2) * height`

@jfirebaugh

### DIFF
--- a/src/mbgl/map/transform_state.hpp
+++ b/src/mbgl/map/transform_state.hpp
@@ -56,7 +56,8 @@ public:
 
     // Rotation
     float getAngle() const;
-    float getAltitude() const;
+    float getFieldOfView() const;
+    float getCameraToCenterDistance() const;
     float getPitch() const;
 
     // State
@@ -109,7 +110,11 @@ private:
     double x = 0, y = 0;
     double angle = 0;
     double scale = 1;
-    double altitude = 1.5;
+    // This fov value is somewhat arbitrary. The altitude of the camera used
+    // to be defined as 1.5 screen heights above the ground, which was an
+    // arbitrary choice. This is the fov equivalent to that value calculated with:
+    // `fov = 2 * arctan((height / 2) / (height * 1.5))`
+    double fov = 0.6435011087932844;
     double pitch = 0.0;
 
     // cache values for spherical mercator math
--- a/include/mbgl/style/layers/custom_layer.hpp
+++ b/include/mbgl/style/layers/custom_layer.hpp
@@ -25,7 +25,7 @@ struct CustomLayerRenderParameters {
     double zoom;
     double bearing;
     double pitch;
-    double altitude;
+    double fieldOfView;
 };
 
 /**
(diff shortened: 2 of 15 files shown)


jfirebaugh:
> I removed `altitude` from `CustomLayerRenderParameters` and added `fieldOfView`. Is this considered an external interface / breaking change?

It's a breaking change to an undocumented API, so kind of a grey area. We're already breaking this API for iOS 3.2.0, and could conceivably squeeze this change into that release. However, Android 3.4.0 shipped today, and it's a breaking change there as well.

cc @1ec5 @cammace

1ec5:
We're going to merge release-ios-3.4.0 back to master tomorrow following today's release of iOS SDK v3.4.0-beta.5. The release branch contains some significant changes to custom layers (#7250) that would conflict with this branch if you were to update the existing custom style layer API.

Of note, the drawing parameters are now contained in a struct, which means we could simply add fieldOfView alongside the (meaningless, unused) perspectiveSkew field (which maps to altitude in mbgl). But I'd actually prefer that we go ahead and replace perspectiveSkew, since we haven't officially released our other backwards-incompatible changes yet.

1ec5:
Meanwhile, I can remove the perspectiveSkew field on the release branch.

1ec5:
> The old altitude value is just 0.5 / tan(fov / 2) * height

Wasn't it hardcoded to 1.5? This is why I figured it would be safe to remove without much fuss.

ansis:
I replaced `altitude` with `fieldOfView` in this ios stuff. I think this is safe to break considering it's an undocumented api and an argument probably no one was using.

All this needs is review

src/mbgl/map/transform_state.hpp
```diff
@@ -109,7 +110,7 @@ class TransformState {
     double x = 0, y = 0;
     double angle = 0;
     double scale = 1;
-    double altitude = 1.5;
+    double fov = 0.6435011087932844;
```
1ec5: [This comment](https://github.com/mapbox/mapbox-gl-native/blob/a6b1773a44d4ef086eb30967cb041cb64312650a/platform/darwin/src/MGLGeometry.mm#L8-L10) should also be updated to point to this field.

Issue #7253, opened 2016-12-01
Make gl-native handling of newlines conform to gl-js behavior


ChrisLoer:
When I ported my recent gl-native line breaking changes to gl-js, I found that gl-js has tests for labels with newlines in them (https://github.com/mapbox/mapbox-gl-js/blob/master/test/js/symbol/shaping.test.js). The gl-js tests expect each newline to cause a new line to be created, while gl-native will just treat them like whitespace (so two newlines in the middle of your label will be collapsed down to just one breaking point instead of inserting a blank line).

I'm not sure when we would expect labels to have newlines in them, but we should probably standardize on one behavior or the other.

cc @ansis @1ec5

1ec5:
The GL JS behavior sounds more correct.

ChrisLoer:
Ah, I just had to look through the file history to see it was recently added as an enhancement: https://github.com/mapbox/mapbox-gl-js/issues/955

ChrisLoer:
I'll address this with a port of https://github.com/mapbox/mapbox-gl-js/pull/3743 to gl-native. I have the code ready, but waiting for the first PR to go in before making a new PR to keep the test changes from conflicting.

1ec5:
This is happening in mapbox/mapbox-gl-native#7446.

ChrisLoer:
Closed with mapbox/mapbox-gl-native#7446.

PR #7446, opened 2016-12-15
Fix issue #7253: port gl-js line breaking behavior

ChrisLoer:
Porting newline-handling changes from gl-js -- biggest change is the "raggedness minimizing" line breaking algorithm from https://github.com/mapbox/mapbox-gl-js/pull/3743.

### DIFF
--- a/platform/default/bidi.cpp
+++ b/platform/default/bidi.cpp
@@ -19,12 +19,8 @@ public:
     UBiDi* bidiLine = nullptr;
 };
     
-BiDi::BiDi() : impl(std::make_unique<BiDiImpl>())
-{
-}
-
-BiDi::~BiDi() {
-}
+BiDi::BiDi() : impl(std::make_unique<BiDiImpl>()) {}
+BiDi::~BiDi() = default;
 
 // Takes UTF16 input in logical order and applies Arabic shaping to the input while maintaining
 // logical order. Output won't be intelligible until the bidirectional algorithm is applied
@@ -53,7 +49,7 @@ std::u16string applyArabicShaping(const std::u16string& input) {
     return std::u16string(outputText.get(), outputLength);
 }
 
-void BiDi::mergeParagraphLineBreaks(std::set<int32_t>& lineBreakPoints) {
+void BiDi::mergeParagraphLineBreaks(std::set<size_t>& lineBreakPoints) {
     int32_t paragraphCount = ubidi_countParagraphs(impl->bidiText);
     for (int32_t i = 0; i < paragraphCount; i++) {
         UErrorCode errorCode = U_ZERO_ERROR;
@@ -65,11 +61,11 @@ void BiDi::mergeParagraphLineBreaks(std::set<int32_t>& lineBreakPoints) {
                                      u_errorName(errorCode));
         }
 
-        lineBreakPoints.insert(paragraphEndIndex);
+        lineBreakPoints.insert(static_cast<std::size_t>(paragraphEndIndex));
     }
 }
 
-std::vector<std::u16string> BiDi::applyLineBreaking(std::set<int32_t> lineBreakPoints) {
+std::vector<std::u16string> BiDi::applyLineBreaking(std::set<std::size_t> lineBreakPoints) {
     // BiDi::getLine will error if called across a paragraph boundary, so we need to ensure that all
     // paragraph boundaries are included in the set of line break points. The calling code might not
     // include the line break because it didn't need to wrap at that point, or because the text was
@@ -77,8 +73,8 @@ std::vector<std::u16string> BiDi::applyLineBreaking(std::set<int32_t> lineBreakP
     mergeParagraphLineBreaks(lineBreakPoints);
 
     std::vector<std::u16string> transformedLines;
-    int32_t start = 0;
-    for (int32_t lineBreakPoint : lineBreakPoints) {
+    std::size_t start = 0;
+    for (std::size_t lineBreakPoint : lineBreakPoints) {
         transformedLines.push_back(getLine(start, lineBreakPoint));
         start = lineBreakPoint;
     }
@@ -87,7 +83,7 @@ std::vector<std::u16string> BiDi::applyLineBreaking(std::set<int32_t> lineBreakP
 }
 
 std::vector<std::u16string> BiDi::processText(const std::u16string& input,
-                                              std::set<int32_t> lineBreakPoints) {
+                                              std::set<std::size_t> lineBreakPoints) {
     UErrorCode errorCode = U_ZERO_ERROR;
 
     ubidi_setPara(impl->bidiText, input.c_str(), static_cast<int32_t>(input.size()),
@@ -100,9 +96,9 @@ std::vector<std::u16string> BiDi::processText(const std::u16string& input,
     return applyLineBreaking(lineBreakPoints);
 }
 
-std::u16string BiDi::getLine(int32_t start, int32_t end) {
+std::u16string BiDi::getLine(std::size_t start, std::size_t end) {
     UErrorCode errorCode = U_ZERO_ERROR;
-    ubidi_setLine(impl->bidiText, start, end, impl->bidiLine, &errorCode);
+    ubidi_setLine(impl->bidiText, static_cast<int32_t>(start), static_cast<int32_t>(end), impl->bidiLine, &errorCode);
 
     if (U_FAILURE(errorCode)) {
         throw std::runtime_error(std::string("BiDi::getLine (setLine): ") + u_errorName(errorCode));
--- a/platform/default/bidi.cpp
+++ b/platform/default/bidi.cpp
@@ -8,20 +8,30 @@ namespace mbgl {
 
 class BiDiImpl {
 public:
-    BiDiImpl() : bidiText(ubidi_open()), bidiLine(ubidi_open()) {}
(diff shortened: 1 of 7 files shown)


src/mbgl/text/glyph_set.cpp
```diff
@@ -99,70 +99,145 @@ void justifyLine(std::vector<PositionedGlyph>& positionedGlyphs,
...
+    // We could skip evaluating breaks where the line length (breakX - priorBreak.x) > maxWidth
+    //  ...but in fact we allow lines longer than maxWidth (if there's no break points)
+    //  ...and when targetWidth and maxWidth are close, strictly enforcing maxWidth can give
+    //     more lopsided results.
+    
+    std::list<LineBreakEdge> edges;
+    // We're iterating from last potential break towards the first
+    for (const PotentialBreak& potentialBreak : potentialBreaks) {
+        const float lineWidth = breakX - potentialBreak.x;
+        edges.push_front(LineBreakEdge{ .priorBreak = &potentialBreak,
```
jfirebaugh: Huh. Is this using C99 designated initializers? I was under the impression that these [don't work in C++](http://stackoverflow.com/questions/18731707/why-does-c11-not-support-designated-initializer-list-as-c99).

ChrisLoer: Apparently not part of C++ but still [supported with a warning by clang and gcc](http://stackoverflow.com/a/19272351/99757). I'll switch to C++-style struct initialization -- I only picked up this style from Xcode suggesting fixes to the gl-js code I had pasted into the editor.

ChrisLoer:
@brunoabinader this touches the Qt-specific bidi.cpp, although it shouldn't change behavior at all.

platform/default/bidi.cpp
```diff
@@ -8,20 +8,30 @@ namespace mbgl {
...
 
     UBiDi* bidiText = nullptr;
     UBiDi* bidiLine = nullptr;
 };
+    
+BiDi::BiDi() : impl(std::make_unique<BiDiImpl>())
+{
+}
+
+BiDi::~BiDi() {
```
jfirebaugh: `BiDi::~BiDi() = default;`

src/mbgl/text/glyph_set.cpp
```diff
@@ -99,70 +99,140 @@ void justifyLine(std::vector<PositionedGlyph>& positionedGlyphs,
...
+    //     more lopsided results.
+    
+    std::list<LineBreakEdge> edges;
+    // We're iterating from last potential break towards the first
+    for (const PotentialBreak& potentialBreak : potentialBreaks) {
+        const float lineWidth = breakX - potentialBreak.x;
+        edges.push_front(LineBreakEdge{ &potentialBreak,
+                                        calculateBadness(lineWidth, targetWidth, penalty, isLastBreak) + potentialBreak.badness });
+    }
+    edges.push_front(LineBreakEdge{ .priorBreak = nullptr, .badness = calculateBadness(breakX, targetWidth, penalty, isLastBreak)});
```
jfirebaugh: Replace the designated initializers here.

src/mbgl/text/glyph_set.cpp
```diff
@@ -99,70 +99,140 @@ void justifyLine(std::vector<PositionedGlyph>& positionedGlyphs,
...
+    const float badness;
+};
+    
+PotentialBreak evaluateBreak(const int32_t breakIndex, const float breakX, const float targetWidth, const std::list<PotentialBreak>& potentialBreaks, const float penalty, const bool isLastBreak) {
+    // We could skip evaluating breaks where the line length (breakX - priorBreak.x) > maxWidth
+    //  ...but in fact we allow lines longer than maxWidth (if there's no break points)
+    //  ...and when targetWidth and maxWidth are close, strictly enforcing maxWidth can give
+    //     more lopsided results.
+    
+    std::list<LineBreakEdge> edges;
```
jfirebaugh: Can this use `std::vector` instead, with pushes to the back and a reversed sort order?

src/mbgl/text/glyph_set.cpp
```diff
@@ -99,70 +99,140 @@ void justifyLine(std::vector<PositionedGlyph>& positionedGlyphs,
...
-            lastSafeBreakX = 0;
+            potentialBreaks.push_front(evaluateBreak(i, currentX, targetWidth, potentialBreaks,
+                                                     calculatePenalty(codePoint, previousCodePoint),
+                                                     false));
         }
 
-        currentX += glyph.metrics.advance + spacing;
+        currentX += it->second.metrics.advance + spacing;
     }
+    potentialBreaks.push_front(evaluateBreak(static_cast<int32_t>(logicalInput.size()), currentX, targetWidth, potentialBreaks, 0, true));
```
jfirebaugh: Is it possible to reduce the number of `static_cast`s in this code by using `std::size_t` in more places?

ChrisLoer: Pushed the static casts into `bidi.cpp` implementations.

src/mbgl/text/glyph_set.cpp
ChrisLoer: I switched to using a `std::vector`, which for better or worse required wrapping the `PotentialBreak` in a `shared_ptr` so that I could maintain the singly-linked `priorBreak` list.

src/mbgl/text/glyph_set.cpp
```diff
@@ -81,88 +80,168 @@ void align(Shaping& shaping,
...
+};
+    
+std::shared_ptr<PotentialBreak> evaluateBreak(const std::size_t breakIndex, const float breakX, const float targetWidth, const std::vector<std::shared_ptr<PotentialBreak>>& potentialBreaks, const float penalty, const bool isLastBreak) {
+    // We could skip evaluating breaks where the line length (breakX - priorBreak.x) > maxWidth
+    //  ...but in fact we allow lines longer than maxWidth (if there's no break points)
+    //  ...and when targetWidth and maxWidth are close, strictly enforcing maxWidth can give
+    //     more lopsided results.
+    
+    std::vector<LineBreakEdge> edges;
+    // We're iterating from last potential break towards the first
```
jfirebaugh: Is this iteration order necessary or could it use `for (const auto& potentialBreak : potentialBreaks)`?

src/mbgl/text/glyph_set.cpp
```diff
@@ -81,88 +80,168 @@ void align(Shaping& shaping,
...
+    for (auto rit = potentialBreaks.rbegin(); rit != potentialBreaks.rend(); ++rit ) {
+        const std::shared_ptr<PotentialBreak> potentialBreak = *rit;
+        const float lineWidth = breakX - potentialBreak->x;
+        edges.push_back(LineBreakEdge(potentialBreak,calculateBadness(lineWidth, targetWidth, penalty, isLastBreak) + potentialBreak->badness));
+    }
+    edges.push_back(LineBreakEdge(nullptr, calculateBadness(breakX, targetWidth, penalty, isLastBreak)));
+
+    std::sort(edges.begin(), edges.end(), [](const LineBreakEdge& a, const LineBreakEdge& b) {
```
jfirebaugh: Rather than keeping the entire `edges` vector, wouldn't it suffice to keep a running minimum badness and prior break?

src/mbgl/text/glyph_set.cpp
```diff
@@ -81,88 +80,168 @@ void align(Shaping& shaping,
...
+        if (a.badness == b.badness && a.priorBreak && b.priorBreak) {
+            // On a badness tie, break to make top line fuller than bottom line
+            return a.priorBreak->index > b.priorBreak->index;
+        }
+        return a.badness < b.badness;
+    });
+    
+    return std::make_shared<PotentialBreak>(PotentialBreak(breakIndex, breakX, edges.front().priorBreak, edges.front().badness));
```
jfirebaugh: `std::make_shared<PotentialBreak>(breakIndex, breakX, ...)`

src/mbgl/text/glyph_set.cpp
```diff
@@ -81,88 +80,168 @@ void align(Shaping& shaping,
...
-            lastSafeBreakX = 0;
+            potentialBreaks.push_back(evaluateBreak(i, currentX, targetWidth, potentialBreaks,
+                                                     calculatePenalty(codePoint, previousCodePoint),
+                                                     false));
         }
 
-        currentX += glyph.metrics.advance + spacing;
+        currentX += it->second.metrics.advance + spacing;
     }
+    potentialBreaks.push_back(evaluateBreak(logicalInput.size(), currentX, targetWidth, potentialBreaks, 0, true));
```
jfirebaugh: Return the result directly rather than doing `push_back` + `back`: `return leastBadBreaks(*evaluateBreak(...));`

src/mbgl/text/glyph_set.cpp
jfirebaugh: Ah ok. I think it's better to avoid `shared_ptr`, either by using a container that doesn't invalidate references on insert or by precalculating the number of potential breakpoints and preallocating the vector.

ChrisLoer: `std::list` is a container that doesn't invalidate references on insert, right?

ChrisLoer:
@jfirebaugh @nickidlugash @1ec5 I've updated this PR with a commit that addresses some of the small issues that came up testing with Chinese map data (although the issues weren't specific to China).

PR #7457, opened 2016-12-15
Use appropriate part of speech for properties

1ec5:
Fixed overridden property references in requirements lists. Boolean-typed properties can now have getters beginning with “is”.

Renamed a number of layout properties according to the following rules, synopsized from the [Cocoa naming guidelines for properties](https://developer.apple.com/library/content/documentation/Cocoa/Conceptual/CodingGuidelines/Articles/NamingIvarsAndTypes.html#//apple_ref/doc/uid/20001284-BAJGIIJE):

1. Boolean-typed properties should read like predicates (and include a verb).
2. Other properties must be noun phrases.
3. All properties must be grammatical.

In most cases, you can get the right getter or setter in Xcode’s code completion popup by typing part of the style specification name for the property. For example, typing `allowover` brings up both `iconAllowsOverlap` and `setIconAllowsOverlap:`.

The main impetus for these changes is that property getters and action methods can look identical in Objective-C: `[layer fillAntialias]` looks like a method that performs an action but returns no value. The new syntax, `[layer isFillAntialiased]`, is unambiguous. Meanwhile, dot syntax is unaffected: `layer.fillAntialiased`.

Here are the changes:

Old getter | New getter | Old setter | New setter
----|----|----|----
`-fillAntialias` | `-isFillAntialiased` | `-setFillAntialias:` | `-setFillAntialiased:`
`-iconAllowOverlap` | `-iconAllowsOverlap` | `-setIconAllowOverlap:` | `-setIconAllowsOverlap:`
`-iconIgnorePlacement` | `-iconIgnoresPlacement` | `-setIconIgnorePlacement:` | `-setIconIgnoresPlacement:`
`-iconKeepUpright` | `-keepsIconUpright` | `-setIconKeepUpright:` | `-setKeepsIconUpright:`
`-iconOptional` | `-isIconOptional` | `-setIconOptional:` | `-setIconOptional:`
`-iconRotate` | `-iconRotation` | `-setIconRotate:` | `-setIconRotation:`
`-symbolAvoidEdges` | `-symbolAvoidsEdges` | `-setSymbolAvoidEdges:` | `-setSymbolAvoidsEdges:`
`-textAllowOverlap` | `-textAllowsOverlap` | `-setTextAllowOverlap:` | `-setTextAllowsOverlap:`
`-textIgnorePlacement` | `-textIgnoresPlacement` | `-setTextIgnorePlacement:` | `-setTextIgnoresPlacement:`
`-textJustify` | `-textJustification` | `-setTextJustify:` | `-setTextJustification:`
`-textKeepUpright` | `-keepsTextUpright` | `-setTextKeepUpright:` | `-setKeepsTextUpright:`
`-textMaxAngle` | `-maximumTextAngle` | `-setTextMaxAngle:` | `-setMaximumTextAngle:`
`-textMaxWidth` | `-maximumTextWidth` | `-setTextMaxWidth:` | `-setMaximumTextWidth:`
`-textOptional` | `-isTextOptional` | `-setTextOptional:` | `-setTextOptional:`
`-textRotate` | `-textRotation` | `-setTextRotate:` | `-setTextRotation:`

Before this PR can land:

* [x] Add the style specification names as aliases but mark them as unavailable
* [x] Implement getters for unavailable properties to avoid bloating the class with autosynthesized ivars that never get used
* [x] Add unit tests that enforce rules 1 and 2 using `NSLinguisticTagSchemeLexicalClass`

Working towards #6098.


platform/darwin/test/MGLStyleLayerTests.h
```diff
@@ -6,5 +6,15 @@
 
 @property (nonatomic) IBOutlet MGLMapView *mapView;
 @property (nonatomic) XCTestExpectation *expectation;
+@property (nonatomic, copy, readonly, class) NSString *layerType;
```
frederoni: I suppose we dropped support for Xcode 7? Great use for the new class property. 👍  Does this mean it's safe to use not only in tests?

platform/darwin/test/MGLStyleLayerTests.m
```diff
@@ -19,4 +25,65 @@ - (void)setUp {
...
+#endif
+    }
+}
+
+@end
+
+@implementation NSString (MGLStyleLayerTestAdditions)
+
+- (NS_ARRAY_OF(NSString *) *)lexicalClasses {
+    NSOrthography *orthography = [NSOrthography orthographyWithDominantScript:@"Latn"
```
frederoni: This is fantastic. A nitpicker's dream come true ;)

platform/darwin/test/MGLStyleLayerTests.h
1ec5: I figured we only run the tests on Xcode 8, so there's no harm in using it here. If we decide to drop support for Xcode 7.3, that needs to be a more deliberate decision, and it's probably too close to the release to make that change. We just recently dropped support for Xcode 7.2 in v3.3.7.

platform/darwin/test/MGLStyleLayerTests.m
1ec5: Note that this method doesn't get used for much by default. I put most calls to this method behind a compile-time flag that's off by default, because NSLinguisticTagging is actually pretty bad at tagging parts of speech for individual words. We could do better by moving these tests into JavaScript as part of the codegen script, but unfortunately the most popular npm tagging modules either require calls to a Web API or are licensed under the GPL.

1ec5: That said, when I turned on the stricter checks locally, they did catch a property I'd missed. So it's a useful tool to run once in awhile, even if there are many false positives.

1ec5:
mapbox/mapbox-gl-style-spec#201 would upstream most of these changes.

Issue #6183, opened 2016-08-26
Depend on URL parsing library


1ec5:
We’ve had some hairy issues with the code in `mbgl::util::mapbox` that relies on artisanal string parsing to manipulate URL components. For example: #5554 #5723. At one point, it was manageable because we were only manipulating schemes and paths. But increasingly we need to work with query parameters as well (#6182). I guess I’m spoiled by the robustness of `URL` on the Web and NSURL and NSURLComponents on Apple platforms, which provide object-oriented, RFC-compliant URL parsing. Could we make `mbgl::util::mapbox` depend on a similarly robust, well-tested URL parsing library for C++?

/cc @kkaefer @jfirebaugh

jfirebaugh:
I'd love to use a well-tested URL parsing library for C++, providing it meets other quality metrics. The `uri` class from [cpp-netlib](http://cpp-netlib.org/) seems to be the leading contender. I can't find any official API reference for it, but it was the basis for a [standards proposal](http://www.open-std.org/jtc1/sc22/wg21/docs/papers/2012/n3420.html) and there is [standalone code](https://github.com/cpp-netlib/uri) floating about.

It looks promising, but I also see some warning signs. We'd probably need to devote some time to trying it out and seeing if it meets our quality bar, or whether we'd need to write something from scratch.

1ec5:
I wonder how much trouble it would be to extract [`WebCore::URLParser`](https://github.com/WebKit/webkit/blob/3711f967d0e7942ed92bf3b82203765706d9c507/Source/WebCore/platform/URLParser.h) from the WebKit project. It depends on WTF for a StringBuilder class, but that doesn’t seem like a showstopper.

springmeyer:
Does not appear maintained, so not a viable contender (https://github.com/bnoordhuis/uriparser2), but dropping here nonetheless for reference.

jingsam:
How about [http-parser](https://github.com/nodejs/http-parser). It has a function called `http_parser_parse_url`. Since it is used by Nodejs, so we can keep consistency with mapbox-gl-js in URL handling

kkaefer:
Node.js' `url` module parsing functionality is done in https://github.com/nodejs/node/blob/master/lib/url.js, and mapbox-gl-js doesn't use Node.js at all since it's typically run in the browser. That being said, we could look at porting the JavaScript code in Node.js to C++.

[...]

PR #7464, opened 2016-12-16
Refactor URL parsing code

kkaefer:
An attempt at https://github.com/mapbox/mapbox-gl-native/issues/6183.

### DIFF
--- a/src/mbgl/util/mapbox.cpp
+++ b/src/mbgl/util/mapbox.cpp
@@ -1,10 +1,12 @@
 #include <mbgl/util/mapbox.hpp>
 #include <mbgl/util/constants.hpp>
 #include <mbgl/util/logging.hpp>
+#include <mbgl/util/url.hpp>
 
 #include <stdexcept>
 #include <vector>
 #include <iostream>
+#include <cstring>
 
 namespace {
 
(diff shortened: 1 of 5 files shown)


1ec5:
It would be great if we could eventually upstream and reconcile this code with https://github.com/mapbox/uri-hpp/.

/cc @mapsam

src/mbgl/util/mapbox.cpp
```diff
@@ -21,188 +23,148 @@ bool isMapboxURL(const std::string& url) {
...
-    }
+    const auto tpl = baseURL + "/v4/{path}?access_token=" + accessToken;
+    return transformURL(tpl, str, url);
+}
 
-    const auto extensionIdx = url.find(".", basenameIdx);
-    if (extensionIdx == std::string::npos || extensionIdx == queryIdx - 1) {
-        return url;
+std::string
+canonicalizeTileURL(const std::string& str, const SourceType type, const uint16_t tileSize) {
```
jfirebaugh: The prior version of `canonicalizeTileURL` wasn't restricted to `*.mapbox.com`, right? It definitely needs to continue to support `*.mapbox.cn` (we should add a test for this) and should likely continue to support arbitrary/staging hostnames as well. It's only used in situations where the source URL is `mapbox://...`, so I don't think enforcing a particular hostname is necessary.

jfirebaugh:
👍  on `URL`/`Path` -- seems like a good middle ground between ad hoc parsing and a full blown RFC compliant library.

test/util/url.test.cpp
```diff
@@ -23,3 +23,184 @@ TEST(URL, isURL) {
...
+    EXPECT_EQ(Path::Segment({ 19, 0 }), URLPath("http://example.com/foo?query=foo.bar").directory);
+    EXPECT_EQ(Path::Segment({ 19, 0 }), URLPath("http://example.com/foo.png").directory);
+    EXPECT_EQ(Path::Segment({ 19, 0 }), URLPath("http://example.com/foo.png?query=foo.bar").directory);
+    EXPECT_EQ(Path::Segment({ 19, 0 }), URLPath("http://example.com/").directory);
+    EXPECT_EQ(Path::Segment({ 19, 0 }), URLPath("http://example.com/?query=foo.bar").directory);
+    EXPECT_EQ(Path::Segment({ 18, 0 }), URLPath("http://example.com").directory);
+    EXPECT_EQ(Path::Segment({ 18, 0 }), URLPath("http://example.com?query=foo.bar").directory);
+}
+
+TEST(Path, Extension) {
```
jfirebaugh: Should have a test for double extensions (e.g. `.vector.pbf`).

kkaefer: I guess it's debatable on what constitutes an extension and what is part of the filename. For a (hypothetical) URL of `mapbox://sources/mapbox.mapbox-terrain-v2,mapbox.mapbox-streets-v7`, would the extension be nothing? or `.mapbox-terrain-v2,mapbox.mapbox-streets-v7`?

jfirebaugh: Looks like you went with "the final period and everything following". Seems good to me!

kkaefer:
For reference, here's an insightful article by the author of cURL on why RFC compliance for URL parsing doesn't work: https://daniel.haxx.se/blog/2016/05/11/my-url-isnt-your-url/

PR #7465, opened 2016-12-16
[core] Polylabel-based "pole of inaccessibility" symbol placement

### LINKED ISSUE #6115: Add support for pole-of-inaccessibility symbol placement

jfirebaugh:
Fixes #6115

### DIFF
--- a/src/mbgl/layout/symbol_layout.cpp
+++ b/src/mbgl/layout/symbol_layout.cpp
@@ -313,8 +313,8 @@ void SymbolLayout::addFeature(const SymbolFeature& feature,
                 poly.push_back(r);
             }
 
-            // 16 here represents 2 pixels
-            auto poi = mapbox::polylabel(poly, 16.0);
+            // 1 pixel worth of precision, in tile coordinates
+            auto poi = mapbox::polylabel(poly, double(util::EXTENT / util::tileSize));
             Anchor anchor(poi.x, poi.y, 0, minScale);
             addSymbolInstance(polygon[0], anchor);
         }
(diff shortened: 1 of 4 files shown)


src/mbgl/layout/symbol_layout.cpp
```diff
@@ -301,12 +303,20 @@ void SymbolLayout::addFeature(const SymbolFeature& feature,
...
+            Polygon<double> poly;
+            for (const auto& ring : polygon) {
+                LinearRing<double> r;
+                for (const auto& p : ring) {
+                    r.push_back(convertPoint<double>(p));
+                }
+                poly.push_back(r);
             }
+
+            // 16 here represents 2 pixels
```
kkaefer: why?

mourner: Yes, exactly.

kkaefer: Ok; can we replace this with the symbolic constants that we're using throughout GL native?

jfirebaugh: Doesn't 16 equal 1 pixel though? 8192/512=16

blanchg: That parameter is the accuracy threshold, I.e. stop searching if the target size reaches 16 or less.  @mourner helped me to identify that. Is it possible at the time we were working off of 256 tiles?

PR #7468, opened 2016-12-17
Highlight current topic in documentation

1ec5:
On category pages, the current category is highlighted in the navigation list:

category

On topic pages, the current topic is highlighted:

topic

This effect assumes that categories and topics are named uniquely. The only thing that would prevent that from happening would be a guide named identically to a category or topic, which we can easily avoid.

@mayagao, I just reused the navigation bar’s background color for the highlight. Feel free to improve the style.

/cc @friedbunny

### DIFF
--- a/platform/darwin/docs/theme/assets/css/jazzy.css.scss
+++ b/platform/darwin/docs/theme/assets/css/jazzy.css.scss
@@ -367,6 +367,11 @@ pre code {
   color: $navigation_task_color;
 }
 
+.nav-group-name > .small-heading,
+.nav-group-task-link {
+  display: block;
+}
+
 .nav-group-task-link {
    color: $color_darkblue_lighten;
 }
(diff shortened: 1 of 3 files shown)


platform/darwin/docs/theme/assets/css/jazzy.css.scss
```diff
@@ -367,6 +367,11 @@ pre code {
   color: $navigation_task_color;
 }
 
+.nav-group-name > .small-heading,
+.nav-group-task-link {
+  display: block;
```
1ec5: Oh, I forgot to mention: this also extends the hit target for the links in the navigation list to the right, so that you can click the link without having the mouse cursor directly over the link text.

mayagao:
looks great @1ec5 pushed minor style tweaks. I would not highlight the entire section list when clicking on section names. I changed it to just highlight the section name.

PR #7475, opened 2016-12-18
Update and reformat documentation for runtime styling

### LINKED ISSUE #6144: Runtime styling documentation should word-wrap
Documentation comments generated from the style specification for the runtime styling API currently rely on soft wrapping. For readability and consistency with the rest of the iOS and macOS SDK documentation, we should try to hard wrap at column 80, except where an HTML tag or URL would force a line to run long ([example](https://github.com/mapbox/mapbox-gl-native/blob/7fb7bc0fbc519e1ac934eb1218c4a3032def9317/platform/ios/src/MGLMapView.h#L48-L77)).

/cc @frederoni @friedbunny

1ec5:
Rewrote documentation about MGLShape, its subclasses, and MGLFeature to emphasize runtime styling uses over annotation uses. The documentation comment for each class answers the following questions:

* What kind of shape does it represent?
* What real-world concept would I represent using this kind of shape?
* How would I add this shape to a map using the runtime styling API? Which MGLStyleLayer subclasses apply?
* How would I add this shape to the map as an annotation or overlay? Which MGLMapViewDelegate methods apply?
* How does this class handle edge cases?
* What other classes implement similar shapes or are easily confused with this class?
* What does GeoJSON call this kind of shape, if it differs from the class name?

Rewrote documentation about MGLStyleLayer and its subclasses. The documentation comments are nominally based on the documentation in the style specification. However, all the existing layer types’ documentation has been overridden to answer the following questions:

* What does the output look like?
* What kinds of vector tile features would I style with this class?
* What kinds of MGLShapes and MGLFeatures would I style with this class?
* How do I get an existing layer of this type? How do I add a new one to the style?

Generated documentation comments for MGLStyleLayer subclasses now wrap at column 80. References to properties in other classes have gone from “`MGLClass`’s `property` property” to “the `MGLClass.property` property”, because jazzy can apparently autolink the latter syntax.

Updated links to the GeoJSON specification to point to RFC 7946. Removed references to unimplemented features like circle overlays. Corrected the description of MGLBackgroundStyleLayer, which erroneously stated that the identifier must be “background”, whereas that only happens to be true by default for Studio template styles.

Fixes #6144.

/cc @ericrwolfe


platform/darwin/scripts/style-spec-overrides-v8.json
```diff
@@ -1,4 +1,28 @@
...
+        },
+        "background": {
+          "doc": "An `MGLBackgroundStyleLayer` is a style layer that covers the entire map. Use a background style layer to configure a color or pattern to show below all other map content. If the style’s other layers use the Mapbox Streets source, the background style layer is responsible for drawing land, whereas the oceans and other bodies of water are drawn by `MGLFillStyleLayer` objects.\n\nA background style layer is typically the bottommost layer in a style, because it covers the entire map and can occlude any layers below it. You can therefore access it by getting the last item in the `MGLStyle.layers` array.\n\nIf the background style layer is transparent or omitted from the style, any portion of the map view that does not show another style layer is transparent."
```
1ec5: Note that “getting the last item in the `MGLStyle.layers` array” becomes “getting the first item in the `MGLStyle.layers` array” once #7484 lands.

platform/darwin/scripts/style-spec-overrides-v8.json
```diff
@@ -1,4 +1,28 @@
 {
+  "layer": {
+    "type": {
+      "values": {
+        "fill": {
+          "doc": "An `MGLFillStyleLayer` is a style layer that renders one or more filled (and optionally stroked) polygons on the map.\n\nUse a fill style layer to configure the visual appearance of polygon or multipolygon features in vector tiles loaded by an `MGLVectorSource` object or `MGLPolygon`, `MGLPolygonFeature`, `MGLMultiPolygon`, or `MGLMultiPolygonFeature` instances in an `MGLShapeSource` object."
```
1ec5: Once fill extrusion layers land on the master branch, we’ll need to provide a similar documentation string for the `fill-extrusion` type.

/cc @boundsj

platform/darwin/scripts/style-spec-overrides-v8.json
```diff
@@ -1,4 +1,28 @@
...
+        },
+        "symbol": {
+          "doc": "An `MGLSymbolStyleLayer` is a style layer that renders icon and text labels at points or along lines on the map.\n\nUse a symbol style layer to configure the visual appearance of labels for features in vector tiles loaded by an `MGLVectorSource` object or `MGLShape` or `MGLFeature` instances in an `MGLShapeSource` object."
+        },
+        "circle": {
+          "doc": "An `MGLCircleStyleLayer` is a style layer that renders one or more filled circles on the map.\n\nUse a circle style layer to configure the visual appearance of point or point collection features in vector tiles loaded by an `MGLVectorSource` object or `MGLPointAnnotation`, `MGLPointFeature`, `MGLPointCollection`, or `MGLPointCollectionFeature` instances in an `MGLShapeSource` object.\n\nA circle style layer renders circles whose radii are measured in screen units. To display circles on the map whose radii correspond to real-world distances, use many-sided regular polygons and configure their appearance using an `MGLFillStyleLayer` object."
```
1ec5: Once this change makes it into the master branch, we’ll need to update this documentation string to note that circle style layers can be filled _and_ optionally stroked.

PR #7484, opened 2016-12-19
Reversed MGLStyle.layers

### LINKED ISSUE #7474: Reverse MGLStyle.layers
MGLStyle’s `layers` property should be ordered from back to front, not from front to back.

In #6097, I said:

> The `layers` array is ordered from front to back, for consistency with Cocoa, as opposed to mbgl’s back-to-front ordering.

I was mistaken: similar arrays in Cocoa are ordered from back to front (`NSView.subviews`, `UIView.subviews`, `UIApplication.windows`) or specify no explicit order (`NSApplication.windows`). (Apple Events and AppleScript do treat “front” as a synonym for “first” and “back” as a synonym for “last”, but that shouldn’t inform the API we expose in this SDK.)

I think a last-minute breaking change before the release is probably the right decision in the long run. Reversing the array will simplify MGLStyle somewhat, as it removes an impedance mismatch with mbgl. It’ll also make the iOS and macOS SDKs more consistent with the Android SDK and GL JS. Moreover, the `layers` property was only published a few days ago in beta 5, along with more convenient methods like `-insertLayer:aboveLayer:`, so direct indexing into `layers` should be relatively rare.

/cc @boundsj @frederoni @ericrwolfe @bsudekum

frederoni:
Fixed #7474 Reverse MGLStyle.layers.

### DIFF
--- a/platform/darwin/src/MGLStyle.mm
+++ b/platform/darwin/src/MGLStyle.mm
@@ -226,18 +226,18 @@ static NSURL *MGLStyleURL_emerald;
 {
     auto layers = self.mapView.mbglMap->getLayers();
     NSMutableArray *styleLayers = [NSMutableArray arrayWithCapacity:layers.size()];
-    for (auto layer = layers.rbegin(); layer != layers.rend(); ++layer) {
-        MGLStyleLayer *styleLayer = [self layerFromMBGLLayer:*layer];
+    for (auto layer : layers) {
+        MGLStyleLayer *styleLayer = [self layerFromMBGLLayer:layer];
         [styleLayers addObject:styleLayer];
     }
     return styleLayers;
 }
 
 - (void)setLayers:(NS_MUTABLE_ARRAY_OF(MGLStyleLayer *) *)layers {
-    for (MGLStyleLayer *layer in self.layers.reverseObjectEnumerator) {
+    for (MGLStyleLayer *layer in self.layers) {
         [self removeLayer:layer];
     }
-    for (MGLStyleLayer *layer in layers.reverseObjectEnumerator) {
+    for (MGLStyleLayer *layer in layers) {
         [self addLayer:layer];
     }
 }
@@ -250,12 +250,12 @@ static NSURL *MGLStyleURL_emerald;
 - (MGLStyleLayer *)objectInLayersAtIndex:(NSUInteger)index
 {
     auto layers = self.mapView.mbglMap->getLayers();
-    if (index > layers.size() - 1) {
+    if (index >= layers.size()) {
         [NSException raise:NSRangeException
                     format:@"No style layer at index %lu.", (unsigned long)index];
         return nil;
     }
-    auto layer = layers.at(layers.size() - 1 - index);
+    auto layer = layers.at(index);
     return [self layerFromMBGLLayer:layer];
 }
 
@@ -287,13 +287,14 @@ static NSURL *MGLStyleURL_emerald;
                     format:@"Cannot insert style layer at out-of-bounds index %lu.", (unsigned long)index];
     } else if (index == 0) {
         try {
-            [styleLayer addToMapView:self.mapView belowLayer:nil];
+            MGLStyleLayer *sibling = layers.size() ? [self layerFromMBGLLayer:layers.at(0)] : nil;
+            [styleLayer addToMapView:self.mapView belowLayer:sibling];
         } catch (const std::runtime_error & err) {
             [NSException raise:@"MGLRedundantLayerIdentifierException" format:@"%s", err.what()];
         }
     } else {
         try {
-            MGLStyleLayer *sibling = [self layerFromMBGLLayer:layers.at(layers.size() - index)];
+            MGLStyleLayer *sibling = [self layerFromMBGLLayer:layers.at(index)];
             [styleLayer addToMapView:self.mapView belowLayer:sibling];
         } catch (std::runtime_error & err) {
             [NSException raise:@"MGLRedundantLayerIdentifierException" format:@"%s", err.what()];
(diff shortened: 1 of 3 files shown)


platform/darwin/src/MGLStyle.mm
```diff
@@ -293,7 +293,7 @@ - (void)insertObject:(MGLStyleLayer *)styleLayer inLayersAtIndex:(NSUInteger)ind
         }
```
1ec5: The cases above need to be adjusted. If you pass in an index equal to the number of layers, an exception should be raised. If you pass in an index of 0, the layer should be inserted below the bottommost layer, not above the topmost layer.

1ec5: The cases above need to be adjusted. If you pass in an index equal to the number of layers, an exception should be raised. If you pass in an index of 0, the layer should be inserted below the bottommost layer, not above the topmost layer.

platform/darwin/src/MGLStyle.mm
```diff
@@ -308,7 +308,7 @@ - (void)removeObjectFromLayersAtIndex:(NSUInteger)index
         [NSException raise:NSRangeException
```
1ec5: It's valid to remove the last (topmost) layer in the style.

frederoni: These off-by-one errors are confusing but it's still possible to remove the topmost layer.
Perhaps a greater than or equal to operator would make it easier to reason about?

1ec5: If you pass in the `index` of the last layer in the array, wouldn’t `index` be equal to `layers.size() - 1`, triggering this exception?

frederoni: It would be equal but it wouldn't trigger because the operator is `>`

1ec5: Doh, I’m blind! 😆 OK, I remember intentionally using `> … - 1` here to parallel the subtraction below. Now that that line below has been simplified, let’s use `>=` instead.

1ec5:
Some exception messages still need to be adjusted: for example, [this message](https://github.com/mapbox/mapbox-gl-native/blob/81a728ee0ba8990da30551641624b9908542110f/platform/darwin/src/MGLStyle.mm#L451) should say “cannot be placed below”, not above.

frederoni:
> macosapp's Layers sidebar should continue to list the layers top to bottom, but I can take care of that in a separate PR.

Feel free to implement that in this PR if it makes it easier.

platform/darwin/src/MGLStyle.mm
```diff
@@ -282,18 +282,19 @@ - (void)insertObject:(MGLStyleLayer *)styleLayer inLayersAtIndex:(NSUInteger)ind
...
-            [styleLayer addToMapView:self.mapView belowLayer:nil];
+            MGLStyleLayer *sibling = layers.size() ? [self layerFromMBGLLayer:layers.at(0)] : nil;
+            [styleLayer addToMapView:self.mapView belowLayer:sibling];
         } catch (const std::runtime_error & err) {
             [NSException raise:@"MGLRedundantLayerIdentifierException" format:@"%s", err.what()];
         }
     } else {
         try {
-            MGLStyleLayer *sibling = [self layerFromMBGLLayer:layers.at(layers.size() - index)];
+            MGLStyleLayer *sibling = [self layerFromMBGLLayer:layers.at(index + 1)];
```
1ec5: Inserting a layer at an index should put it below (before) the layer currently at that index. Therefore, this should be `layers.at(index)`.

platform/darwin/src/MGLStyle.mm
```diff
@@ -448,7 +449,7 @@ - (void)insertLayer:(MGLStyleLayer *)layer aboveLayer:(MGLStyleLayer *)sibling {
     if (index + 1 > layers.size()) {
         [NSException raise:NSInvalidArgumentException
                     format:
-         @"A style layer cannot be placed above %@ in the style. "
+         @"A style layer cannot be placed below %@ in the style. "
```
1ec5: Oops, I misread the diff when I suggested rewording this to say “below” instead of “above” in https://github.com/mapbox/mapbox-gl-native/pull/7484#issuecomment-268092473. It should indeed say “above”, because that’s what this method is all about.

platform/darwin/src/MGLStyle.mm
```diff
@@ -282,18 +282,19 @@ - (void)insertObject:(MGLStyleLayer *)styleLayer inLayersAtIndex:(NSUInteger)ind
          styleLayer];
     }
     auto layers = self.mapView.mbglMap->getLayers();
-    if (index > layers.size()) {
+    if (index >= layers.size()) {
         [NSException raise:NSRangeException
                     format:@"Cannot insert style layer at out-of-bounds index %lu.", (unsigned long)index];
     } else if (index == 0) {
```
1ec5: `index == layers.size()` gets caught by the case above, which I think is incorrect: you can append a single object to NSMutableArray by passing its `count` into `-insertObject:atIndex:`; an exception only gets raised if the index is `count + 1` or higher.

platform/darwin/src/MGLStyle.mm
frederoni: What if you pass in an index of 0 and the number of layers is 0? The layer should be inserted and no exception should be raised. `-[NSMutableArray insertObject:atIndex:]` allows index=size but not + 1. The second argument still holds but the first should be `>`.

1ec5: I agree with you: https://github.com/mapbox/mapbox-gl-native/pull/7484#discussion_r93326377. Unfortunately it's this comment that stuck around. 😁

Issue #7471, opened 2016-12-17
Programmatically generated raster sources


1ec5:
https://github.com/mapbox/mapbox-gl/issues/27#issuecomment-267752292 demonstrates a use case for programmatically generated raster tile sources similar to what MapKit enables through [`-[MKTileOverlay loadTileAtPath:result:]`](https://developer.apple.com/reference/mapkit/mktileoverlay/1452445-loadtileatpath) and the Google Maps SDK enables through [`-[GMSTileLayer requestTileForX:y:zoom:receiver:]`](https://developers.google.com/maps/documentation/ios-sdk/reference/interface_g_m_s_tile_layer.html#a60f5c33f63f8f6067822dbb9d8171116).

In contrast to #6940, which would expose a separate source type for vector content, the SDK would make it possible for developers subclass MGLRasterSource and override a `-loadTileAtPath:result:` method that returns an image for a particular tile coordinate. `mbgl::style::RasterSource`’s initializer would accept an optional lambda that passes premultiplied image data into an asynchronous callback. `mbgl::style::RasterSource` would call this lambda for each tile, falling back on the current code path when none is provided. MGLRasterSource would pass in a lambda that calls `-loadTileAtPath:result:`.

This would require MGLStyle to own all MGLSource objects, as proposed in https://github.com/mapbox/mapbox-gl-native/issues/7375#issuecomment-266528289. Allowing an SDK source object to control the tile loading process would allow us to avoid a proliferation of source types and make the source customization hooks more intuitive for developers.

/cc @kkaefer @johndpope @JesseCrocker

1ec5:
We’d have to circumvent ambient and offline tile caching for programmatically defined raster sources.

1ec5:
At the same time as we add an API for providing images, we should also add a parallel API for providing tile URLs on a per-tile or per-request basis. Callback or block-based APIs are very common for this feature:

* MapKit allows the developer to subclass MKTileOverlay (akin to MGLRasterSource) and override the [`-URLForTilePath:`](https://developer.apple.com/reference/mapkit/mktileoverlay/1452606-urlfortilepath) method, which accepts an [MKTileOverlayPath](https://developer.apple.com/reference/mapkit/mktileoverlaypath) struct (containing `x`, `y`, etc.) and returns a NSURL.
[...]

Issue #7455, opened 2016-12-15
Tiles are cached by url template in the database


wibge:
Tilesets are cached by url template, if the url to a tile source changes previously cached and downloaded maps are not seamlessly loaded. 

We have tile sources that have frequently changing API keys as part of the URL.

**Platform:** ios 
**Mapbox SDK version:** 3.4

### Steps to trigger behavior

 1. Create a tileset with a URL containing an API key
 2. Cache tiles from the tileset
 3. Change the API key in the URL

### Expected behavior
Cached tiles are displayed when the tileset is on screen
### Actual behavior
Cached tiles are not displayed

I'd propose adding an optional tilecachekey parameter to use to override caching by url template.

jfirebaugh:
#8084 adds a feature that provides the ability to modify the URL independently of the value used for a cache key. Per https://github.com/mapbox/mapbox-gl-native/pull/8041#issuecomment-281738359, I believe this addresses your use case.

PR #7485, opened 2016-12-19
Allow setting URL scheme templates

kkaefer:
We are currently using `mapbox://` URLs, but they're hardcoded into the system. This PR adds a generic `setURLSchemeTemplate` to `OnlineFileSource` and `DefaultFileSource` that allows you to specify templates for your own URL schemes.

### DIFF
--- a/platform/darwin/src/MGLOfflineStorage+MGLURLSchemeTemplateAdditions.mm
+++ b/platform/darwin/src/MGLOfflineStorage+MGLURLSchemeTemplateAdditions.mm
@@ -1,0 +1,33 @@
+#import "MGLOfflineStorage+MGLURLSchemeTemplateAdditions.h"
+#import "MGLOfflineStorage_Private.h"
+
+@implementation MGLOfflineStorage (MGLURLSchemeTemplateAdditions)
+
+/**
+ Sets a URL template for the given scheme.
+
+ Any URLs encountered in styles and subsequently loaded sources will be fed through
+ a URL replacement mechanism that e.g. transforms `mapbox://` URLs into regular
+ `https://` URLs. You can provide URL replacement patterns for your own scheme,
+ e.g. `example://`.
+
+ Any query string will be preserved from the original URL and merged with a
+ query string of the template URL, if any.
+
+ @param tpl A templated URL. it may contain these brace-enclosed replacement
+    tokens, e.g. for `"http://example.com/foo/bar/baz.png"`, valid tokens are:
+    * `{scheme}`: `"http"`
+    * `{domain}`: `"example.com"`
+    * `{path}`: `"foo/bar/baz.png"` (without leading `/`)
+    * `{directory}`: `"foo/bar/"` (with trailing `/`)
+    * `{filename}`: `"baz"` (without file extension)
+    * `{extension}`: `".png"` (optionally with a preceding `@2x` ratio specifier)
+ @param scheme The scheme that this template should be used for. Specify the
+    scheme without trailing `://`, e.g. just `example`. You may use arbitrary
+    alphanumeric schemes, with the exception of the protected scheme `mapbox`.
+ */
+- (void)setURLTemplate:(nullable NSString *)tpl forScheme:(NSString *)scheme {
+    self.mbglFileSource->setURLSchemeTemplate(scheme.UTF8String, tpl.UTF8String);
+}
+
+@end
--- a/include/mbgl/storage/default_file_source.hpp
+++ b/include/mbgl/storage/default_file_source.hpp
@@ -36,6 +36,9 @@ public:
     void setAccessToken(const std::string&);
     std::string getAccessToken() const;
 
+    void setURLSchemeTemplate(const std::string& scheme, const std::string& tpl);
+    std::string getURLSchemeTemplate(const std::string& scheme) const;
+
     std::unique_ptr<AsyncRequest> request(const Resource&, Callback) override;
 
     /*
--- a/include/mbgl/storage/online_file_source.hpp
+++ b/include/mbgl/storage/online_file_source.hpp
@@ -3,6 +3,8 @@
 #include <mbgl/storage/file_source.hpp>
 #include <mbgl/util/constants.hpp>
 
+#include <map>
+
 namespace mbgl {
 
 class OnlineFileSource : public FileSource {
@@ -16,6 +18,9 @@ public:
     void setAccessToken(const std::string& t) { accessToken = t; }
     std::string getAccessToken() const { return accessToken; }
 
+    void setURLSchemeTemplate(const std::string& scheme, const std::string& tpl);
+    std::string getURLSchemeTemplate(const std::string& scheme) const;
+
     std::unique_ptr<AsyncRequest> request(const Resource&, Callback) override;
 
 private:
@@ -25,6 +30,7 @@ private:
     const std::unique_ptr<Impl> impl;
     std::string accessToken;
     std::string apiBaseURL = mbgl::util::API_BASE_URL;
+    std::map<std::string, std::string> urlSchemeTemplates;
 };
 
 } // namespace mbgl
--- a/platform/android/MapboxGLAndroidSDK/src/main/java/com/mapbox/mapboxsdk/maps/MapView.java
+++ b/platform/android/MapboxGLAndroidSDK/src/main/java/com/mapbox/mapboxsdk/maps/MapView.java
@@ -373,6 +373,34 @@ public class MapView extends FrameLayout {
   }
 
   /**
+   * <p>Sets a URL template for the given scheme.</p>
+   * <p>Any URLs encountered in styles and subsequently loaded sources will be fed through a URL
+   *   replacement mechanism that e.g. transforms {@code mapbox://} URLs into regular
+   *   {@code https://} URLs. You can provide URL replacement patterns for your own scheme, e.g.
+   *   {@code example://}.</p>
+   * <p>Any query string will be preserved from the original URL and merged with a query string of
+   *   the template URL, if any.</p>
+   *
+   * @param scheme scheme The scheme that this template should be used for. Specify the scheme
+   * without trailing {@code ://}, e.g. just {@code example}. You may use arbitrary alphanumeric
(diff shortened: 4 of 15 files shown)


kkaefer:
(based on https://github.com/mapbox/mapbox-gl-native/pull/7464)

platform/darwin/src/MGLOfflineStorage.h
```diff
@@ -238,6 +238,31 @@ typedef void (^MGLOfflinePackRemovalCompletionHandler)(NSError * _Nullable error
...
+    * `{domain}`: `"example.com"`
+    * `{path}`: `"foo/bar/baz.png"` (without leading `/`)
+    * `{directory}`: `"foo/bar/"` (with trailing `/`)
+    * `{filename}`: `"baz"` (without file extension)
+    * `{extension}`: `".png"` (optionally with a preceding `@2x` ratio specifier)
+ @param scheme The scheme that this template should be used for. Specify the
+    scheme without trailing `://`, e.g. just `example`. You may use arbitrary
+    alphanumeric schemes, with the exception of the protected scheme `mapbox`.
+ */
+- (void)setURLSchemeTemplate:(nullable NSString *)tpl forScheme:(NSString *)scheme;
```
1ec5: “Scheme” is redundant here. A better name would be `-setURLTemplate:forScheme:` or `-setTemplate:forURLScheme:`.

platform/darwin/src/MGLOfflineStorage.h
```diff
@@ -238,6 +238,31 @@ typedef void (^MGLOfflinePackRemovalCompletionHandler)(NSError * _Nullable error
...
+ query string of the template URL, if any.
+
+ @param tpl A templated URL. it may contain these brace-enclosed replacement
+    tokens, e.g. for `"http://example.com/foo/bar/baz.png"`, valid tokens are:
+    * `{scheme}`: `"http"`
+    * `{domain}`: `"example.com"`
+    * `{path}`: `"foo/bar/baz.png"` (without leading `/`)
+    * `{directory}`: `"foo/bar/"` (with trailing `/`)
+    * `{filename}`: `"baz"` (without file extension)
+    * `{extension}`: `".png"` (optionally with a preceding `@2x` ratio specifier)
```
1ec5: I wonder if clang and jazzy are able to handle this nested list.

1ec5: Mini-languages in strings are un-Cocoa. Instead of exposing yet another mini-language that resembles but works differently than `textField` tokens in the runtime styling API, have you considered a lambda-based (block-based) API? The developer could specify a block (or perhaps an NSValueTransformer) that accepts an NSURL and returns a transformed NSURL. Within the block, they could easily parse and manipulate the URL via NSURLComponents.

Even more natural would be a delegate pattern: the developer could have their application delegate conform to a new MGLOfflineStorageDelegate protocol by implementing an `-offlineStorage:transformedURLForTileURL:` method.

platform/darwin/src/MGLOfflineStorage.h
```diff
@@ -238,6 +238,31 @@ typedef void (^MGLOfflinePackRemovalCompletionHandler)(NSError * _Nullable error
  */
 @property (nonatomic, readonly) unsigned long long countOfBytesCompleted;
 
+/**
+ Sets a URL template for the given scheme.
```
1ec5: It isn’t clear whether this API affects only tile URLs or also style URLs, sprite sheet URLs, etc.

platform/darwin/src/MGLOfflineStorage.h
kkaefer: While that may be true, using platform-specific URL parsing code also introduces a source for potential platform deviation. Therefore, I'd prefer using the same code for all platforms.

platform/darwin/src/MGLOfflineStorage.h
kkaefer: you're right; this was a victim of automated text replacement (the previous name was `-setURLTemplate:forScheme:`). I'll change this to `-setTemplate:forURLScheme:`

platform/darwin/src/MGLOfflineStorage.h
1ec5: Platform deviation would certainly be a problem when it comes to parsing URLs in JSON files, but a block would consist of platform-specific (or at least language-specific) code. Any deviation from how URLs are handled on the platform or in the language’s standard library would be more of a surprise, wouldn’t it?

In the unlikely event that incompatibilities between standard URL parsers on different platforms poses a problem to the developer, they could do minimal parsing inline. For most scenarios, however, a callback would enable developers to bring more context into the transformation logic without requiring us to add a special case for every innovation in tile URL technology.

Even if you don’t want to leave URL parsing up to the developer, why not have the block accept a dictionary mapping e.g. `directory` to `foo/bar/` and allow the developer to perform concatenation and any other logic themselves?

1ec5: Google and Apple both provide callback or block-based APIs on their analogues to our raster source concept: https://github.com/mapbox/mapbox-gl-native/issues/7471#issuecomment-268722312.

platform/darwin/src/MGLOfflineStorage.h
kkaefer: Not sure how `Any URLs encountered in styles and subsequently loaded sources` could be any clearer

1ec5: Gah, don’t know how I missed that. 😆 I do think it would be helpful, for those who aren’t familiar with the resources that make up a style, to know exactly what “any URLs” entails. But that points to the need for a conceptual “what’s a style” guide in the documentation.

kkaefer:
An alternative implementation (which is still lacking Android support) is in https://github.com/mapbox/mapbox-gl-native/compare/url-transform. It uses a function callback that allows manipulation of the Resource/URL.

Open questions are:
- the delegate's method is called from the FileSource thread. Is that idiomatic in Cocoa? Using a synchronous dispatch to the main thread is impossible due to deadlocks.
- Is the setter [implemented correctly?](https://github.com/mapbox/mapbox-gl-native/compare/url-transform#diff-3f0caf5e71e8443a455628a8f6824fa7R50)

Usage is (e.g. in `AppDelegate.m`):

```diff
diff --git a/platform/macos/app/AppDelegate.h b/platform/macos/app/AppDelegate.h
index a1d9297..8d3e284 100644
--- a/platform/macos/app/AppDelegate.h
+++ b/platform/macos/app/AppDelegate.h
@@ -2,7 +2,7 @@
 
 extern NSString * const MGLMapboxAccessTokenDefaultsKey;
 
-@interface AppDelegate : NSObject <NSApplicationDelegate>
+@interface AppDelegate : NSObject <NSApplicationDelegate, MGLOfflineStorageDelegate>
 
 @property (weak) IBOutlet NSWindow *preferencesWindow;
 
diff --git a/platform/macos/app/AppDelegate.m b/platform/macos/app/AppDelegate.m
index d3fe2d2..cef3c1e 100644
--- a/platform/macos/app/AppDelegate.m
+++ b/platform/macos/app/AppDelegate.m
@@ -130,10 +130,19 @@ NSString * const MGLLastMapDebugMaskDefaultsKey = @"MGLLastMapDebugMask";
         [alert runModal];
         [self showPreferences:nil];
     }
+
+    [MGLOfflineStorage sharedOfflineStorage].delegate = self;
     
     [self.offlinePacksArrayController bind:@"content" toObject:[MGLOfflineStorage sharedOfflineStorage] withKeyPath:@"packs" options:nil];
 }
 
+- (NSURLComponents *)offlineStorage:(MGLOfflineStorage *)storage transformURL:(NSURLComponents *)url ofKind:(MGLResourceKind)kind {
+    if ([url.host isEqual: @"api.mapbox.com"]) {
+        url.host = @"a.tiles.mapbox.com";
+    }
+    return url;
+}
+
 - (void)applicationWillTerminate:(NSNotification *)notification {
     if (![[NSUserDefaults standardUserDefaults] boolForKey:@"NSQuitAlwaysKeepsWindows"]) {
         NSDocument *currentDocument = [NSDocumentController sharedDocumentController].currentDocument;
```

1ec5:
It’s an antipattern in Objective-C (and arguably in general) for different methods on the same class to be called on different threads. That forces the class to explicitly synchronize any shared state such as ivars and avoid implicitly capturing self in any blocks. I’ve never seen it done successfully but, conversely, have had to tear such code out many times due to concurrency bugs.

If there’s no way to call a method on the main thread, the delegate cannot be AppDelegate. We should strongly recommend that the developer instead create a helper object – a subclass of NSObject that conforms to MGLOfflineStorageDelegate. Either AppDelegate (or some other application class) would own this object, or MGLOfflineStorage would own it – in which case we’d call it something else like MGLURLProtocol or even make use of the NSURLProtocol system.

1ec5:
Related: #7455.

platform/darwin/src/MGLOfflineStorage+MGLURLSchemeTemplateAdditions.mm
```diff
@@ -0,0 +1,33 @@
...
+ e.g. `example://`.
+
+ Any query string will be preserved from the original URL and merged with a
+ query string of the template URL, if any.
+
+ @param tpl A templated URL. it may contain these brace-enclosed replacement
+    tokens, e.g. for `"http://example.com/foo/bar/baz.png"`, valid tokens are:
+    * `{scheme}`: `"http"`
+    * `{domain}`: `"example.com"`
+    * `{path}`: `"foo/bar/baz.png"` (without leading `/`)
```
jfirebaugh: Why is the path without leading `/`? That is inconsistent with the definitions of path elsewhere, e.g. [`window.Location`](https://developer.mozilla.org/en-US/docs/Web/API/Location), [node](https://nodejs.org/api/url.html#url_urlobject_pathname), [RFC 3986](https://tools.ietf.org/html/rfc3986#section-3.3).

kkaefer: Added pull request to change this in #7750

jfirebaugh:
Merged #8084 instead.

Issue #6534, opened 2016-09-30
MapboxAccountManager refactor


tobrun:
We are looking into changing the way MapboxMapManager works. Recently this object was only responsible for managing accesstokens but recently it became responsible for managing connection changes as well. With an upcoming release I suggest is completely rework currently exposed API. I would suggest updating the name to something else eg. `Mapbox`.

cc @mapbox/android

zugaldia:
I would like to move this object to `mapbox-java` so that it can easily handle tokens for API requests too. The challenge is that as you note `MapboxAccountManager`:

> recently it became responsible for managing connection changes as well.

I'm thinking of having a generic `NetworkInfo` interface without any Android deps, with the SDK providing a default implementation using Android's.

I like `Mapbox` instead of `MapboxAccountManager` too.

ivovandongen:
> I'm thinking of having a generic NetworkInfo interface without any Android deps, with the SDK providing a default implementation using Android's.

👍 This would indeed give some more flexibility here

PR #7497, opened 2016-12-20
Regression wear module

tobrun:
#6534 has resulted in regressing the wear test app. 

cc @cammace @kkaefer


kkaefer:
Can we add a test for this as well?

tobrun:
Yes, :100:, ticketed in https://github.com/mapbox/mapbox-gl-native/issues/7500. 
Going to merge this as the issue makes master not compile in certain use cases.

PR #7509, opened 2016-12-21
Add -fvisibility=hidden to macOS dynamic framework and export required symbols

### LINKED ISSUE #7281: Many visibility warnings building macOS SDK

kkaefer:
Reroll of #7326, which apparently has broken CI statuses.

Fixes #7326
Fixes #7281

### DIFF
--- a/platform/darwin/src/MGLAccountManager.h
+++ b/platform/darwin/src/MGLAccountManager.h
@@ -1,11 +1,14 @@
 #import <Foundation/Foundation.h>
 
+#import "MGLFoundation.h"
+
 NS_ASSUME_NONNULL_BEGIN
 
 /**
  The MGLAccountManager object provides a global way to set a Mapbox API access
  token.
  */
+MGL_EXPORT
 @interface MGLAccountManager : NSObject
 
 #pragma mark Authorizing Access
--- a/platform/darwin/src/MGLBackgroundStyleLayer.h
+++ b/platform/darwin/src/MGLBackgroundStyleLayer.h
@@ -1,6 +1,7 @@
 // This file is generated. 
 // Edit platform/darwin/scripts/generate-style-code.js, then run `make style-code-darwin`.
 
+#import "MGLFoundation.h"
 #import "MGLStyleValue.h"
 #import "MGLStyleLayer.h"
 
@@ -12,6 +13,7 @@ NS_ASSUME_NONNULL_BEGIN
  `style` and obtain the background layer using the `-[MGLStyle layerWithIdentifier:]` 
  method and passing `background` for the identifier. 
  */
+MGL_EXPORT
 @interface MGLBackgroundStyleLayer : MGLStyleLayer
 
 - (instancetype)initWithIdentifier:(NSString *)identifier NS_DESIGNATED_INITIALIZER;
--- a/platform/darwin/src/MGLFoundation.h
+++ b/platform/darwin/src/MGLFoundation.h
@@ -1,0 +1,5 @@
+#pragma once
+
+#import <Foundation/Foundation.h>
+
+#define MGL_EXPORT __attribute__((visibility ("default")))
--- a/platform/darwin/test/MGLVersionNumber.m
+++ b/platform/darwin/test/MGLVersionNumber.m
@@ -1,0 +1,15 @@
+#import <Mapbox/Mapbox.h>
+
+#import <XCTest/XCTest.h>
+
+@interface MGLVersionTests : XCTestCase
+
+@end
+
+@implementation MGLVersionTests
+
+- (void)testVersionNumber {
+    XCTAssertEqual(1, MapboxVersionNumber);
+}
+
+@end
--- a/platform/macos/DEVELOPING.md
+++ b/platform/macos/DEVELOPING.md
@@ -54,6 +54,7 @@ To add any Objective-C type, constant, or member to the iOS SDK’s public inter
 
 To add an Objective-C class, protocol, category, typedef, enumeration, or global constant to the macOS SDK’s public interface:
 
+1. _(Optional.)_ Add the macro `MGL_EXPORT` prior to the declaration for classes and global constants. To use this macro, include `MGLFoundation.h`. You can check whether all public symbols are exported correctly by running `make check-public-symbols`.
 1. _(Optional.)_ Add the type or constant’s name to the relevant category in the `custom_categories` section of [the jazzy configuration file](./jazzy.yml). This is required for classes and protocols and also recommended for any other type that is strongly associated with a particular class or protocol. If you leave out this step, the symbol will appear in an “Other” section in the generated HTML documentation’s table of contents.
 1. _(Optional.)_ If the symbol would also be publicly exposed in the iOS SDK, consult [the companion iOS document](../ios/DEVELOPING.md#making-a-type-or-constant-public) for further instructions.
 
--- a/platform/macos/macos.xcodeproj/project.pbxproj
+++ b/platform/macos/macos.xcodeproj/project.pbxproj
@@ -58,6 +58,8 @@
 		40E1601D1DF217D6005EA6D9 /* MGLStyleLayerTests.m in Sources */ = {isa = PBXBuildFile; fileRef = 40E1601B1DF216E6005EA6D9 /* MGLStyleLayerTests.m */; };
 		52BECB0A1CC5A26F009CD791 /* SystemConfiguration.framework in Frameworks */ = {isa = PBXBuildFile; fileRef = 52BECB091CC5A26F009CD791 /* SystemConfiguration.framework */; };
 		5548BE781D09E718005DDE81 /* libmbgl-core.a in Frameworks */ = {isa = PBXBuildFile; fileRef = DAE6C3451CC31D1200DB3429 /* libmbgl-core.a */; };
+		556660C61E1BEA0100E2C41B /* MGLFoundation.h in Headers */ = {isa = PBXBuildFile; fileRef = 556660C51E1BEA0100E2C41B /* MGLFoundation.h */; settings = {ATTRIBUTES = (Public, ); }; };
+		556660D61E1D07E400E2C41B /* MGLVersionNumber.m in Sources */ = {isa = PBXBuildFile; fileRef = 556660D51E1D07E400E2C41B /* MGLVersionNumber.m */; };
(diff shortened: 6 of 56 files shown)


kkaefer:
Grr, this seems to be broken as well...

kkaefer:
Let's try approving this to see if we can merge it.

platform/darwin/src/MGLAccountManager.h
```diff
@@ -1,11 +1,12 @@
-#import <Foundation/Foundation.h>
```
1ec5: Apparently we need these import statements or else Swift declarations in the generated documentation lack specific types: realm/jazzy#609. Importing required system frameworks is good practice anyways, unless you use a precompiled header that imports them.

(Sorry, I started this review a couple weeks ago but forgot to publish it.)

kkaefer: `MGLTypes.h` imports `Foundation.h`

kkaefer:
> What about protocols and categories? Do they need to be exported as well?

No, we only need to export compiled code, but protocols/categories are a header-only feature.

platform/darwin/src/MGLAccountManager.h
1ec5: As mentioned in that jazzy ticket, this header itself must import Foundation.h. It's unfortunate, but that shouldn't pose a problem for the visibility changes you're making, should it?

platform/darwin/src/MGLBackgroundStyleLayer.h
```diff
@@ -12,6 +12,7 @@ NS_ASSUME_NONNULL_BEGIN
  `style` and obtain the background layer using the `-[MGLStyle layerWithIdentifier:]` 
  method and passing `background` for the identifier. 
  */
+MGL_EXPORT
```
1ec5: This change will get blown away as soon as anyone invokes `make style-code`. Make this change in platform/darwin/src/MGLStyleLayer.h.ejs and run `make style-code` to propagate the change to all the affected headers.

platform/darwin/src/MGLAccountManager.h
1ec5: I’m not sure why, but we don’t seem to be affected by realm/jazzy#609 even with these changes. Running `make xdocument` turns up some broken Swift declarations, but that happens on master too. (@incanus may have a stronger opinion than I about explicitly importing frameworks in headers. My concern was limited to jazzy output.)

1ec5:
>> What about protocols and categories? Do they need to be exported as well?
> 
> No, we only need to export compiled code, but protocols/categories are a header-only feature.

Protocols are header-only in Objective-C, but categories may or may not have an implementation in a translation unit separate from the categorized class. (Otherwise, it’s a class extension.) See platform/darwin/src/NSValue+MGLAdditions.m for example.

kkaefer:
I've rebuilt this PR based on SourceKitten's output.

Here are the changes:
- Moved the `MGL_EXPORT` macro to a `MGLFoundation.h` to mirror other Cocoa frameworks usage of the `*_EXPORT` macro.
- Left includes for `Foundation.h` in place
- Added a script that checks all exported symbols for presence of a `MGL_EXPORT` macro
- Added the automated script to `make xpackage`

These symbols are not exposed in the public headers, but the Xcode CI unit tests include the private headers and need them in the library. These types of symbols are not checked by the above script, but we'll catch those easily because the test will fail to run on CI.

- `MGLAttributionButton`
- `MGLAltitudeForZoomLevel`
- `MGLZoomLevelForAltitude`
- `MGLFeaturesFromMBGLFeatures`
- `MGLTileSetFromTileURLTemplates`
- `MGLGeoJSONOptionsFromDictionary`

platform/darwin/src/MGLBackgroundStyleLayer.h
kkaefer: This is fixed

kkaefer:
> Protocols are header-only in Objective-C, but a category may or may not have an implementation in a translation unit separate from the categorized class.

I checked this and it seems they don't need the `MGL_EXPORT` statement and are automatically exported. The CI tests + macOS app uses some of those symbols and can be successfully linked.

platform/macos/src/Mapbox.h
```diff
@@ -1,10 +1,12 @@
 #import <Cocoa/Cocoa.h>
 
+#import "MGLFoundation.h"
+
 /// Project version number for Mapbox.
-FOUNDATION_EXPORT double MapboxVersionNumber;
+FOUNDATION_EXPORT MGL_EXPORT double MapboxVersionNumber;
```
1ec5: Make sure `MapboxVersionNumber` and `MapboxVersionString` are correctly set in the output build. I would hope that Xcode isn’t dependent on the particular syntax used to declare these constants, but you never know.

1ec5: I get this linker error referencing `MapboxVersonNumber` within Swift code:

```
  "_MapboxVersionNumber", referenced from:
      Invincibility.ViewController.viewDidLoad () -> () in ViewController.o
```

platform/darwin/src/MGLFoundation.h
```diff
@@ -0,0 +1,5 @@
+#pragma once
+
+#import <Foundation/Foundation.h>
+
+#define MGL_EXPORT __attribute__((visibility ("default")))
```
kkaefer: As per convention in other system frameworks, it has things that are required in virtually every header file. Most other *Foundation.h files I checked only have the export (or extern) macro.

platform/macos/DEVELOPING.md
```diff
@@ -54,6 +54,7 @@ To add any Objective-C type, constant, or member to the iOS SDK’s public inter
 
 To add an Objective-C class, protocol, category, typedef, enumeration, or global constant to the macOS SDK’s public interface:
 
+1. _(Optional.)_ Add the macro `MGL_EXPORT` prior to the declaration for classes and global constants. To use this macro, include `MGLFoundation.h`. You can check whether all public symbols are exported correctly by running `platform/darwin/scripts/check-public-symbols.js macOS`.
```
kkaefer: Added `make check-public-symbols`

platform/macos/src/Mapbox.h
kkaefer: From a C build perspective, this is very weird. `FOUNDATION_EXPORT` is only macro for `extern`, meaning it should be defined in some translation unit (but it's not). Should we manually define this version number?

kkaefer: Looks like Xcode automatically generates a file called `Mapbox_vers.c`

kkaefer:
> can you add a blurb to the macOS changelog if a significant size reduction results from these changes?

There is no size reduction with this PR. This already happened in https://github.com/mapbox/mapbox-gl-native/pull/7130. This one is just a followup that also compiles the Objective-C code with `-fvisibility=hidden` to fix https://github.com/mapbox/mapbox-gl-native/issues/7281.

platform/macos/macos.xcodeproj/project.pbxproj
```diff
@@ -1558,7 +1568,7 @@
 				PRODUCT_NAME = Mapbox;
 				SKIP_INSTALL = YES;
 				VERSIONING_SYSTEM = "apple-generic";
-				VERSION_INFO_PREFIX = "";
+				VERSION_INFO_PREFIX = "__attribute__((visibility (\"default\"))) ";
```
kkaefer: We're now also adding the visibility change to the auto-generated file.

platform/darwin/test/MGLVersionNumber.m
```diff
@@ -0,0 +1,15 @@
...
+
+@interface MGLVersionTests : XCTestCase
+
+@end
+
+@implementation MGLVersionTests
+
+- (void)testVersionNumber {
+    XCTAssertEqual(1, MapboxVersionNumber);
+}
```
kkaefer: Added a test here to check linkage. This is not a Swift-specific issue; it also happened with Objective-C.

platform/macos/macos.xcodeproj/project.pbxproj
1ec5: Clever hack. (I think this build setting is usually used for Apple to prepend a class prefix like “NS” or “UI” to the variable name.)

platform/darwin/test/MGLVersionNumber.m
1ec5: Note that `MapboxVersionString` is also available to Objective-C (but not to Swift, owing to its type). I assume it’s also covered by `VERSION_INFO_PREFIX`.

1ec5:
iOS builds are failing because the iOS version of Mapbox.h doesn’t import MGLFoundation.h:

```
❌  /Users/vagrant/git/platform/darwin/src/MGLAccountManager.h:3:9: include of non-modular header inside framework module 'Mapbox.MGLAccountManager'

#import "MGLFoundation.h"
        ^
```

platform/macos/DEVELOPING.md
1ec5: Great. Can you refer to `make check-public-symbols` in both DEVELOPING.md documents? Thanks!

platform/darwin/test/MGLVersionNumber.m
kkaefer: Yeah, it is.

Issue #5249, opened 2016-06-06
Use Git submodules for shaders/test suite


kkaefer:
@jfirebaugh, I'm frequently running into issues with npm not updating those two repositories. What are the reasons we use npm to manage these rather than just using submodules?

jfirebaugh:
In my experience, updating dependencies are more of a hassle with submodules than with node packages. All the make commands have dependencies that should cause `npm update` to be run whenever `package.json` is more recent than `node_modules`. I don't know of a similar way to get that behavior with submodules.

What situation are you seeing where node_modules doesn't get updated?

tmpsantos:
> What situation are you seeing where node_modules doesn't get updated?

For instance depending on `npm` will make the Yocto and Buildroot packages extra complicated, because you will have `npm` as buildtime dependencies for embedded systems. I second using `git` submodules.

kkaefer:
I'm using linked npm packages, and `npm update` doesn't work with that. Is there another way of updating the code of a package except for linking it?

jfirebaugh:
A submodule might work for shaders, if we can find a way to auto-update it like we do now via `npm update`. I don't think a submodule is workable for the test-suite, because it has sub-dependencies that need to be initialized by npm.

tmpsantos:
> I don't think a submodule is workable for the test-suite, because it has sub-dependencies that need to be initialized by npm.

From embedded packaging perspective that would be fine. We don't build the tests when making the packages.

kkaefer:
I've found that when the submodule rev changed on a certain parent module commit, it automatically gets checked out to that version when you change the HEAD. Apart from that, we can use this combination:

`.gitmodules`:
```
[submodule "deps/shaders"]
	path = deps/shaders
	url = https://github.com/mapbox/mapbox-gl-shaders
	ignore = dirty
```
and calling `git submodule update --init -- deps/shaders` to force the submodule to the correct pinned commit.

jfirebaugh:
> I've found that when the submodule rev changed on a certain parent module commit, it automatically gets checked out to that version when you change the HEAD.

Are you testing that in GitUp? It's a feature of that app, but not the `git checkout` command.

kkaefer:
I used the CLI, but I had GitUp open with that repo, so maybe it did that in the background.

tmpsantos:
Any decision on this? Can we move forward on not depending on `npm` for building on Linux?

jfirebaugh:
[...]

PR #7513, opened 2016-12-21
[build] Use a submodule for mapbox-gl-js rather than an npm dependency

jfirebaugh:
Refs #5249

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -10,6 +10,12 @@ if(NOT EXISTS ".mason/mason.cmake")
         WORKING_DIRECTORY ${CMAKE_SOURCE_DIR})
 endif()
 
+if(NOT EXISTS "mapbox-gl-js/package.json")
+   execute_process(
+        COMMAND git submodule update --init mapbox-gl-js
+        WORKING_DIRECTORY ${CMAKE_SOURCE_DIR})
+endif()
+
 include(.mason/mason.cmake)
 
 option(WITH_CXX11ABI "Use cxx11abi mason packages" OFF)
--- a/.gitmodules
+++ b/.gitmodules
@@ -12,3 +12,6 @@
 [submodule "platform/ios/uitest/OHHTTPStubs"]
 	path = platform/ios/uitest/OHHTTPStubs
 	url = https://github.com/AliSoftware/OHHTTPStubs.git
+[submodule "mapbox-gl-js"]
+	path = mapbox-gl-js
+	url = git://github.com/mapbox/mapbox-gl-js.git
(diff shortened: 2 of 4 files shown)


jfirebaugh:
See in particular https://github.com/mapbox/mapbox-gl-native/issues/5249#issuecomment-268642496 for the rational here.

@kkaefer @tmpsantos @1ec5 Any reservations about testing this out for a few days?

1ec5:
I think it should be OK. We’re still using npm on the release-ios-v3.4.0 branch, which is going to be active for the next few weeks, with periodic merges back to master. It just means we have to be careful not to accidentally bring a submodule change along for the ride as we perform merges.

tmpsantos:
> Any reservations about testing this out for a few days?

No problem at all. Regarding the caching issue we could have a checkout of the repo on the travis cache directory and copy it to the submodule location on the travis shell script.

CMakeLists.txt
```diff
@@ -10,6 +10,12 @@ if(NOT EXISTS ".mason/mason.cmake")
         WORKING_DIRECTORY ${CMAKE_SOURCE_DIR})
 endif()
 
+if(NOT EXISTS "mapbox-gl-js/package.json")
+   execute_process(
+        COMMAND git submodule update --init mapbox-gl-js
+        WORKING_DIRECTORY ${CMAKE_SOURCE_DIR})
```
kkaefer: Can we try using `externalproject_add` with the `GIT_SUBMODULES` setting instead?

jfirebaugh: I wasn't able to get this to work. It looks to me like the `GIT_SUBMODULES` option is for initializing submodules of the external project, rather than using a submodule as the method to obtain the external project itself.

tmpsantos: I've playing with `externalproject_add` for a while on the CMake build I'm maintaining for building Mapbox GL Native 100% offline and for Qt. On the build file for predownloading all of our dependencies I'm doing something like:

```
include(ExternalProject)

# Macro for cloning and downloading our header-only dependencies
macro(add_dep package tag git)
    externalproject_add(${package}
        URL https://github.com/${git}/archive/${tag}.zip
        CONFIGURE_COMMAND ""
        BUILD_COMMAND ""
        INSTALL_COMMAND ""
    )

    add_dependencies(mbgl-core ${package})

    externalproject_get_property(${package} SOURCE_DIR)

    target_include_directories(mbgl-core PRIVATE ${SOURCE_DIR}/include)
    target_include_directories(qmapboxgl PRIVATE ${SOURCE_DIR}/include)

    set(${package}_SOURCE_DIR ${SOURCE_DIR})
endmacro()
```

Two things: It is a lot faster to get the tarball rather then cloning the repo and CMake will try to configure + build + install, but you can override it.

PR #7515, opened 2016-12-21
Correct formatting and typos in MGLStyleValue documentation

friedbunny:
Corrected several issues with `MGLStyleValue`’s inline docs:

- Moved the enumerations responsibility to `NSValue` (from `NSNumber`).
- Corrected the names of `NSEdgeInsets` and `UIEdgeInsets`.
- ~~Changed list formatting to Markdown, as jazzy wasn’t parsing code ticks inside HTML tags.~~
- Changed the backticks to `<code>` blocks, which parse correctly in jazzy and are the same in appearance as the Markdown version.

Now in jazzy:

screen shot 2016-12-21 at 6 30 04 pm

Note that Xcode’s Quick Help now fails to add linebreaks to this list:

screen shot 2016-12-21 at 6 30 19 pm

### DIFF
--- a/platform/darwin/src/MGLStyleValue.h
+++ b/platform/darwin/src/MGLStyleValue.h
@@ -16,14 +16,12 @@ NS_ASSUME_NONNULL_BEGIN
  
  The `MGLStyleValue` class takes a generic parameter `T` that indicates the
  Foundation class being wrapped by this class. Common values for `T` include:
- 
- <ul>
- <li>`NSNumber` (for Boolean values, floating-point numbers, and enumerations)</li>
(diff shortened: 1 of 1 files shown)


1ec5:
> Note that Xcode’s Quick Help now fails to add linebreaks to this list

Quick Help is a mess in Xcode 8.x. Everyone, please file dupes of [rdar://problem/28628571](http://openradar.appspot.com/radar?id=5758406015057920).

friedbunny:
Went back to HTML, as that’s what we’re using elsewhere and it also works with jazzy. I’d be in favor of optimizing our header comments for readability generally, but some other time.

Issue #7521, opened 2016-12-22
Offlined region is not completely removed from the database


andrewstay:
**Platform:** iOS
**Mapbox SDK version:** ios-v3.4.0-beta.5

### Steps to trigger behavior

 1. Offline any region using API from MGLOfflineStorage
 2. Delete the offlined region using
```-[MGLOfflineStorage removePack:withCompletionHandler:]```
 3. Observer the offlined database size

### Expected behavior
Tiles associated with the region should be deleted

### Actual behavior
It is only the region metadata gets removed from the database

friedbunny:
From https://www.mapbox.com/help/mobile-offline/:

> When you remove an offline region, any resources that are required by that region, but not other regions, become eligible for removal from the offline database. Because the offline database is also used as a general purpose cache for map resources, such resources may not be immediately removed if the implementation determines that they remain useful for general performance of the map.

There is currently no way to force cached resources to be removed, though we have plans to implement a method that will clear the ambient cache — see https://github.com/mapbox/mapbox-gl-native/issues/4376.

1ec5:
#3437 tracks adding an API to limit the cache size, which is currently unbounded (except for tile download limits enforced by MGLOfflineStorage).

jfirebaugh:
> the cache size, which is currently unbounded

This isn't entirely accurate. The space consumed by offline regions is unbounded. However, the implementation endeavors to restrict the space consumed by ambient caching to 50 MB. This includes resources that were originally added to the database as part of an offline region, but where that offline region has subsequently been deleted. In that case, the implementation will attempt to shrink the database down to the remaining 50 MB of most recently used resources, but this depends on SQLite being able to effectively recapture free space within the database. It's possible that SQLite is not always able to do so due to database fragmentation.

PR #7522, opened 2016-12-22
Document that offline pack resource removal is not guaranteed

friedbunny:
Attempts to address a common question about why removing an offline pack does not result in the immediate deletion of its associated resources — see #7521, etc.

Text adapted from: https://www.mapbox.com/help/mobile-offline/#removing-offline-regions

screen shot 2016-12-22 at 1 56 34 pm

/cc @jfirebaugh

### DIFF
--- a/platform/darwin/src/MGLOfflineStorage.h
+++ b/platform/darwin/src/MGLOfflineStorage.h
@@ -195,12 +195,12 @@ typedef void (^MGLOfflinePackRemovalCompletionHandler)(NSError * _Nullable error
  KVO change notifications on the shared offline storage object’s `packs` key
  path. Removals from that array result in an `NSKeyValueChangeRemoval` change.
  
- @note When you remove an offline pack, any resources that are required by that
-    pack, but not other packs, become eligible for removal from the offline
-    database. Because the offline database is also used as a general purpose
-    cache for map resources, such resources may not be immediately removed if
-    the implementation determines that they remain useful for general
-    performance of the map.
+ When you remove an offline pack, any resources that are required by that pack,
+ but not other packs, become eligible for deletion from offline storage. Because
+ the backing store used for offline storage is also used as a general purpose
+ cache for map resources, such resources may not be immediately removed if the
+ implementation determines that they remain useful for general performance of
+ the map.
 
  @param pack The offline pack to remove.
  @param completion The completion handler to call once the pack has been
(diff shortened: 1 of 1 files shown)


platform/darwin/src/MGLOfflineStorage.h
```diff
@@ -195,6 +195,13 @@ typedef void (^MGLOfflinePackRemovalCompletionHandler)(NSError * _Nullable error
  KVO change notifications on the shared offline storage object’s `packs` key
  path. Removals from that array result in an `NSKeyValueChangeRemoval` change.
  
+ @note When you remove an offline pack, any resources that are required by that
```
1ec5: No need for `@note` here.

platform/darwin/src/MGLOfflineStorage.h
```diff
@@ -195,6 +195,13 @@ typedef void (^MGLOfflinePackRemovalCompletionHandler)(NSError * _Nullable error
...
+    pack, but not other packs, become eligible for removal from the offline
+    database. Because the offline database is also used as a general purpose
```
1ec5: Let’s talk about “offline storage” instead of “the offline database”. I’ve tried to avoid the word “database” to head off any suggestions that we expose more sophisticated querying functionality.

friedbunny: Using “offline storage” here adds confusion, because a database is a storage medium, whereas “storage” is vague and could be interpreted to refer to more abstract elements of MGLOfflineStorage.

>  When you remove an offline pack, any resources that are required by that pack, but not other packs, become eligible for _deletion from offline storage_. Because _offline storage_ is also used as a general purpose cache for map resources, such resources may not be immediately removed if the implementation determines that they remain useful for general performance of the map.

It strikes me as strange to say, "Because offline storage is also used as a general purpose cache for map resources”...

1ec5: You’re right that it can be awkward to talk in more general terms about what, at the end of the day, is an implementation detail. I suppose as long as we don’t say anything about SQL, SQLite, or Core Data, my concerns are a bit overblown. Merge away.

friedbunny:
Made some changes in an attempt to avoid using the word “database” — https://github.com/mapbox/mapbox-gl-native/pull/7522/commits/67179fe4c4bb2c92a6969f40b02d2269541e79e0. 😉

PR #7533, opened 2016-12-23
[core] Update gl-js for shader change

### LINKED ISSUE #7532: port fix for tinting rasters with alpha
port https://github.com/mapbox/mapbox-gl-js/pull/3846 from -js

jfirebaugh:
Fixes #7532.


src/mbgl/renderer/raster_bucket.cpp
```diff
@@ -12,8 +13,7 @@ RasterBucket::RasterBucket(PremultipliedImage&& image_) : image(std::move(image_
 }
 
 void RasterBucket::upload(gl::Context& context) {
-    texture = context.createTexture(image);
-    image = {};
+    texture = context.createTexture(util::unpremultiply(std::move(image)));
```
jfirebaugh: This passes tests, but is it safe? Isn't unpremultiplication a lossy operation?

ansis: Yeah, it's lossy ad might result it tiny differences in some cases with non-default style properties. They probably wouldn't be human-noticeable.

It would be better to not do this, but your call.

jfirebaugh: I'll leave it using `unpremultiply` for now. We can revisit if it proves to be an issue. I did move the computation to the worker thread though.

PR #7539, opened 2016-12-23
Immutable source and style layer collections

### LINKED ISSUE #7459: MGLStyle.layers and .sources bridge to Swift as untyped collections
`MGLStyle.layers` is typed as a read-write NSMutableArray of MGLStyleLayer objects, while `MGLStyle.sources` is typed as a read-write NSMutableSet of MGLSource objects. Ironically, Swift is unable to distinguish between a read-write NSArray property and a read-write NSMutableArray property (likewise for NSMutableSet), so the properties end up getting bridged as NSMutableArray and NSMutableSet, respectively, instead of [MGLStyleLayer] and [MGLSource].

We should find a way to ensure that Swift developers get the same type safety that Objective-C developers do. One option is to make these collections non-mutable but still implement the mutable KVC methods under the hood. Developers who want to work with the mutable property would then call `-mutableArrayValueForKey:` or `-mutableSetValueForKey:`. While workable, that’s hardly discoverable. On the other hand, we explicitly declare a number of non-KVC methods for layers that should reduce the need to work with the mutable array directly in code. Another option would be to define separate `layers` and `mutableLayers` properties, but that would double the number of code paths. Any other ideas?

Putting on the v3.4.0 milestone for now to consider our options. Depending on what we think the best option would be, we might be able to postpone this work until a patch release.

/cc @frederoni @incanus @boundsj @bsudekum

1ec5:
MGLStyle’s `sources` and `layers` properties are now immutable collections but remain writable. Now Swift developers will see `layers`’ type as `[MGLStyleLayer]`, not `NSMutableArray`. If you need a mutable array of layers, use `-mutableArrayValueForKey:`.

Fixes #7459.

/cc @bsudekum

### DIFF
--- a/platform/darwin/src/MGLStyle.h
+++ b/platform/darwin/src/MGLStyle.h
@@ -189,7 +189,7 @@ static const NSInteger MGLStyleDefaultVersion = 9;
 /**
  A set containing the style’s sources.
  */
-@property (nonatomic, strong) NS_MUTABLE_SET_OF(MGLSource *) *sources;
+@property (nonatomic, strong) NS_SET_OF(MGLSource *) *sources;
 
 /**
  Returns a source with the given identifier in the current style.
@@ -240,7 +240,7 @@ static const NSInteger MGLStyleDefaultVersion = 9;
  The layers included in the style, arranged according to their back-to-front
  ordering on the screen.
  */
-@property (nonatomic, strong) NS_MUTABLE_ARRAY_OF(MGLStyleLayer *) *layers;
+@property (nonatomic, strong) NS_ARRAY_OF(MGLStyleLayer *) *layers;
 
 /**
  Returns a style layer with the given identifier in the current style.
--- a/platform/darwin/src/MGLStyle.h
+++ b/platform/darwin/src/MGLStyle.h
@@ -189,7 +189,7 @@ static const NSInteger MGLStyleDefaultVersion = 9;
 /**
  A set containing the style’s sources.
  */
-@property (nonatomic, strong) NS_SET_OF(MGLSource *) *sources;
+@property (nonatomic, strong) NS_SET_OF(__kindof MGLSource *) *sources;
 
 /**
  Returns a source with the given identifier in the current style.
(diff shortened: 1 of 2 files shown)


platform/darwin/src/MGLStyle.mm
```diff
@@ -127,17 +127,17 @@ - (NSString *)name {
...
-- (NS_MUTABLE_SET_OF(MGLSource *) *)sources {
+- (NS_SET_OF(__kindof MGLSource *) *)sources {
     auto rawSources = self.mapView.mbglMap->getSources();
-    NSMutableSet *sources = [NSMutableSet setWithCapacity:rawSources.size()];
+    NS_MUTABLE_SET_OF(__kindof MGLSource *) *sources = [NSMutableSet setWithCapacity:rawSources.size()];
     for (auto rawSource = rawSources.begin(); rawSource != rawSources.end(); ++rawSource) {
         MGLSource *source = [self sourceFromMBGLSource:*rawSource];
         [sources addObject:source];
     }
     return sources;
```
boundsj: Should these mutable collections in the implementation be sent `copy` when returned?

platform/darwin/src/MGLStyle.h
```diff
@@ -189,7 +189,7 @@ static const NSInteger MGLStyleDefaultVersion = 9;
 /**
  A set containing the style’s sources.
  */
-@property (nonatomic, strong) NS_MUTABLE_SET_OF(MGLSource *) *sources;
+@property (nonatomic, strong) NS_SET_OF(__kindof MGLSource *) *sources;
```
boundsj: Should we mention that it is possible to get a mutable copy as noted in https://github.com/mapbox/mapbox-gl-native/issues/7459#issuecomment-267810124?

1ec5: `-mutableValueForKey:` is available to all KVC-compliant properties. The fact that it happens to be optimized for these two properties is an implementation detail. But if we can think of a good reason to use it in an example, I'd be all for that.

platform/darwin/src/MGLStyle.mm
1ec5: I’d prefer to return the mutable array here instead of incurring the potential cost of a copy, because the variable that holds the array has local scope and the declared return type is immutable. [It would be a programming error](http://stackoverflow.com/a/27392471/4585461) for the caller to cast the object to an NSMutableArray, even conditionally.

PR #7488, opened 2016-12-20
[ios] Adds guides to documentation sidebar

ericrwolfe:
3 guides to start, focused around runtime styling:

* Clarity on the different ways to add points to a map
* Overview of runtime styling
* Prototyping runtime styling with Mapbox Studio

Also improves the way we organize guides and documentation images in the repo.

We should think about a good way to share these between the SDK docs and the /help section of the website.

TODO:
- [x] Add example gifs in Runtime Styling Overview for annotations, interactivity, and custom shapes
- [x] Update links to examples on the website

/cc @friedbunny @captainbarbosa @1ec5

PR #7542, opened 2016-12-24
Guides for style and GeoJSON authors

### LINKED ISSUE #7530: Documentation should include GeoJSON, style specification concordances
Wherever the iOS and macOS SDKs use dramatically distinct names for shapes compared to the GeoJSON standard or rename properties from what the Mapbox Style Specification defines, we mention the original name in the symbol’s documentation. That helps developers who need to migrate something from the iOS/macOS SDK to style JSON or the Android SDK, but not the other way around. We should also add a guides to the generated HTML documentation that include tables mapping the original names to the platform-appropriate names. The style specification concordance would be generated from the same files that are used to generate the MGLStyleLayer headers.

/cc @frederoni @ericrwolfe @tmcw

1ec5:
Added two guides to the jazzy documentation:

* “Working with GeoJSON Data” maps GeoJSON concepts and types to the runtime styling API.
* “Information for Style Authors” points out a number of tips for designing a style appropriate for an iOS or macOS application and also maps style JSON types to the runtime styling API to help Mapbox Studio users. This document is generated by `make style-code-darwin` just like the rest of the runtime styling API. I avoided talking in great detail about Studio or the runtime styling API to avoid overlapping with #7488.

~~These two guides are shared between the iOS and macOS docsets. This means the iOS SDK documentation will include some language specific to macOS and vice versa. I think this is fine, given that Apple’s own guides do the same thing.~~ The two guides are now separate but generated from the same template.

Fixes #7530.

/cc @friedbunny @ericrwolfe @mapbox/copycop

### DIFF
--- a/platform/macos/jazzy.yml
+++ b/platform/macos/jazzy.yml
@@ -17,6 +17,7 @@ custom_categories:
   - name: Guides
     children:
       - Info.plist Keys
+      - Working with GeoJSON Data
   - name: Maps
     children:
       - MGLAccountManager
--- a/platform/macos/jazzy.yml
+++ b/platform/macos/jazzy.yml
@@ -16,9 +16,9 @@ framework_root: ../darwin/src
 custom_categories:
   - name: Guides
     children:
-      - Info.plist Keys
       - Working with GeoJSON Data
       - For Style Authors
+      - Info.plist Keys
   - name: Maps
     children:
       - MGLAccountManager
--- a/platform/macos/jazzy.yml
+++ b/platform/macos/jazzy.yml
@@ -18,6 +18,7 @@ custom_categories:
     children:
       - Info.plist Keys
       - Working with GeoJSON Data
+      - For Style Authors
   - name: Maps
     children:
       - MGLAccountManager
--- a/platform/darwin/docs/guides/For Style Authors.md
+++ b/platform/darwin/docs/guides/For Style Authors.md
@@ -1,0 +1,234 @@
+<!--
+  This file is generated.
+  Edit platform/darwin/scripts/generate-style-code.js, then run `make style-code-darwin`.
+-->
+# Information for Style Authors
+
+A _style_ defines a map view’s content and appearance. If you’ve authored a
+style using
+[Mapbox Studio’s Styles editor](https://www.mapbox.com/studio/styles/) or as
+JSON in a text editor, you can use that style in this SDK and manipulate it
+afterwards in code. This document provides information you can use to ensure a
+seamless transition from Mapbox Studio to your application.
+
+## Designing for the platform
+
+When designing your style, consider the context in which your application shows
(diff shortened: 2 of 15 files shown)


1ec5:
These guides highlight the need for autolinking in guides and the readme: realm/jazzy#715.

1ec5:
“Working with GeoJSON Data”

geojson

“Information for Style Authors”

for-style-authors

1ec5:
The “Information for Style Authors” guide now comes in two flavors, one for iOS and one for macOS. Both are compiled from the same template. I’ve updated the screenshots above to reflect the split.

1ec5:
> We ought to consider moving some of the content from For Style Authors.md out into the respective source documentation headers and link out from the guide, at least until auto linking in jazzy improves.

realm/jazzy#715 is high on [the 👊 list](https://github.com/mapbox/mapbox-gl-native/projects/9). We do mention most (if not all) of the information contained in the tables in the far-flung parts of the API reference, at least where there’s a difference. If you see anything that should be duplicated in the reference pages, please let me know.

In compiling these documents, I wanted to take into consideration @tmcw’s concern that someone familiar with Studio or GL JS would have no good starting point to apply their existing knowledge to the SDK or infer some of the more systematic differences in terminology or typing. The table of contents could be bewildering for these developers, despite your much-needed rework of it, especially when it comes to the source options.

platform/macos/jazzy.yml
```diff
@@ -16,6 +16,9 @@ framework_root: ../darwin/src
 custom_categories:
   - name: Guides
     children:
+      - Adding Points to a Map
+      - Runtime Styling
+      - Working with Mapbox Studio
```
1ec5: These guides are located in platform/ios/docs/guides/, so they aren’t available to the macOS SDK. While most of the Runtime Styling and Working with Mapbox Studio guides apply equally to the macOS SDK, some of the screenshots would need to be replaced, and the “Adding Points to a Map” guide discusses view-backed annotations, which have yet to be implemented on macOS: #5035.

Let’s leave these guides out for now. We can figure out how to bring them over to the macOS SDK in a separate PR. Maybe we should think at that point about whether any of this great content could be reused in Android SDK documentation as well.

1ec5:
I edited the filter-to-predicate-format table to show how multiple values are handled in the format string (especially `none` to `NOT (… OR …)`):

predicates

nitrag:
@friedbunny The URL for beta 7 docs is broken. Thanks for the prerelease! 
https://www.mapbox.com/ios-sdk/api/3.4.0-beta.7/

1ec5:
It’s up now, @nitrag.

friedbunny:
> It’s up now

Well, as soon as it finishes publishing — give it a minute or two. 😉

The delay came from the fact that we finished building late last night and the jazzy generation process is slightly more involved than usual, as we make improvements in the betas.

Issue #6900, opened 2016-11-03
Get rid of user-specified refs


jfirebaugh:
Parity issue with gl-js:

* https://github.com/mapbox/mapbox-gl-js/issues/1393
* https://github.com/mapbox/mapbox-gl-js/pull/3486

jfirebaugh:
[`groupByLayout`](https://github.com/mapbox/mapbox-gl-style-spec/blob/mb-pages/lib/group_by_layout.js) turns out to be extraordinarily difficult to port to C++. I've considered three different approaches:

* Grouping using an `unordered_map` or `unordered_set`. This requires implementing a hash function for `Layer`, `Filter`, `LayoutProperties`, `PropertyValue`, `Function`, and all types used by those types. I got deep into the weeds of trying to make `std::hash` and `boost::hash` cooperate before abandoning this approach.
* Grouping using a `map` or `set`. This requires implementing a less than operator for `Layer`, `Filter`, `LayoutProperties`, `PropertyValue`, `Function`, and all types used by those types. This seems easier until you discover that `unordered_map` (used by mapbox::geometry::value, which is used by `Filter`) doesn't provide a less than operator.
* Following the JS implementation exactly, and implementing (stable) JSON stringify for all these types. This seems equally if not more daunting than either of the above.

PR #7548, opened 2016-12-26
Rewrite predicate/filter conversion

### LINKED ISSUE #7543: None filter converted to NAND predicate
A `none` filter with multiple subfilters is converted into a `NOT` predicate containing an `AND` subpredicate for “NAND” semantics, whereas it should be converted into a `NOT` predicate containing an `OR` subpredicate for “NOR” semantics.

/cc @frederoni

### LINKED ISSUE #7544: Crash when predicate compares value to attribute
Setting `MGLVectorStyleLayer.predicate` to a predicate like `4 == foo` leads to an assertion failure:

```
caught "NSInvalidArgumentException", "*** -constantValue only defined for abstract class.  Define -[NSKeyPathExpression constantValue]!"
(
	0   CoreFoundation                      0x00007fff83b56452 __exceptionPreprocess + 178
	1   libobjc.A.dylib                     0x00007fff87bb4f7e objc_exception_throw + 48
	2   CoreFoundation                      0x00007fff83bbd41d +[NSException raise:format:] + 205
	3   Foundation                          0x00007fff851cddbd NSRequestConcreteImplementation + 229
	4   Foundation                          0x00007fff851798ce -[NSExpression constantValue] + 39
	5   Mapbox                              0x00000001050f7cb8 -[NSComparisonPredicate(MGLAdditions) mgl_filter] + 216
…
```

The problem is that `-[NSComparisonPredicate(MGLAdditions) mgl_filter]` assumes that a comparison must have a key path on the left and a constant (or `nil`) on the right. But NSComparisonPredicate is symmetric.

/cc @frederoni

### LINKED ISSUE #7545: Uncaught exception converting predicate containing aggregate expressions
If you set `MGLVectorStyleLayer.predicate` to an NSPredicate with a format like `a IN {1, 2}`, `{1, 2} CONTAINS a`, or `b BETWEEN {1, 2}`, [this exception](https://github.com/mapbox/mapbox-gl-native/blob/f03cfd8bd50765556e76dcf11e9ae19c94a066cf/platform/darwin/src/NSExpression%2BMGLAdditions.mm#L15) is raised. `-[NSExpression(MGLAdditions) mgl_filterValues]` assumes that the aggregate array must contain values like strings and numbers directly, but clients of NSPredicate normally expect the elements to be NSExpressions.

/cc @frederoni

### LINKED ISSUE #7547: Uncaught exception converting predicate containing CONTAINS operator
If you set `MGLVectorStyleLayer.predicate` to an NSPredicate with a format like `{1, 2} CONTAINS a`, [this line](https://github.com/mapbox/mapbox-gl-native/blob/f03cfd8bd50765556e76dcf11e9ae19c94a066cf/platform/darwin/src/NSComparisonPredicate%2BMGLAdditions.mm#L73) raises an exception. We should be looking at the right expression’s key path, not its constant value.

/ref #6244 #7232
/cc @frederoni @boundsj

1ec5:
This PR rewrites the predicate-to-filter conversion code on iOS and macOS to support more of the predicate language without crashing and also adds comprehensive unit tests for this code.

When converting predicates to filters, symmetric comparison predicates can now compare a value to a key in addition to the usual key-to-value order. Added error checking for unhandled combinations like key-to-key. Fixed a crash converting a `CONTAINS` predicate into a filter. Added support for constant value expressions inside aggregate expressions. Allow sets as aggregate expressions just like arrays, except in `BETWEEN` predicates where order matters. Flatten `NOT` predicates into more specialized filters.

When converting filters to predicates, use constant value expressions inside aggregate expressions. Convert to a `BETWEEN` predicate when possible.

Removed the predicate integration tests, which tested round-tripping but didn’t ensure correct conversion and omitted many edge cases. Replaced the integration tests with systematic unit tests for converting in either direction, plus unit tests for round-tripping and symmetry.

The unit tests required the addition of `==` operators for all the classes associated with `mbgl::style::Filter`.

Fixes #7543, fixes #7544, fixes #7545, fixes #7547.

/cc @boundsj @jfirebaugh


1ec5:
With the new unit tests, we have much better code coverage within the predicate-related categories:

platform/darwin/test/MGLPredicateTests.mm
```diff
@@ -0,0 +1,416 @@
...
+    NSPredicate *forwardPredicate = [NSPredicate predicateWithFormat:forwardFormat];
+    NSPredicate *reversePredicate = reverseFormat ? [NSPredicate predicateWithFormat:reverseFormat] : nil;
+    [self testSymmetryWithPredicate:forwardPredicate reversePredicate:reversePredicate mustRoundTrip:mustRoundTrip];
+}
+
+- (void)testSymmetryWithPredicate:(NSPredicate *)forwardPredicate reversePredicate:(NSPredicate *)reversePredicate mustRoundTrip:(BOOL)mustRoundTrip {
+    auto forwardFilter = forwardPredicate.mgl_filter;
+    NSPredicate *forwardPredicateAfter = [NSPredicate mgl_predicateWithFilter:forwardFilter];
+    if (mustRoundTrip) {
+        // Aggregates should round-trip, but for some reason only their format strings do.
```
frederoni: The `forwardPredicate` rhs expression is a `NSConstantValueExpression(NSArray)` containing plain `NSNumber`s initialized with ints but the `forwardPredicateAfter` is a `NSConstantValueExpression(NSArray)` containing `NSConstantValueExpression(NSNumber)`s initialized with longs. The wrong data type might be another bug in `-[NSExpression(MGLAdditions) mgl_constantMBGLValue]` ? Also, shouldn't `FilterEvaluator::getValues()` make use of `+[NSExpression expressionForAggregate:]` if necessary?

jfirebaugh:
Could you cherry pick the filter operators from https://github.com/mapbox/mapbox-gl-native/compare/deref? I need both `==` and `<` for #6900.

platform/darwin/test/MGLPredicateTests.mm
1ec5: For the `BETWEEN` case, `forwardPredicateAfter` is already an aggregate expression under the hood, due to the use of braces [here](https://github.com/mapbox/mapbox-gl-native/blob/ea4b63cb511eea62bbb2454ce59837f009313aca/platform/darwin/src/NSPredicate%2BMGLAdditions.mm#L97). For the `IN` case, I’ll add `+[NSExpression expressionForAggregate:]` for consistency with `BETWEEN`.

I’m comfortable with the SDK “upgrading” `int`s to `long`s and literal collections to aggregates when round-tripping. While it’s unfortunate that Foundation makes distinctions here, it’s basically Postel’s Law: accepting either collections or aggregates in input, but only outputting aggregates; accepting either `int`s or `long`s in input, but only outputting `long`s.

1ec5: All set.

PR #7553, opened 2016-12-27
Add layer predicate editor to macosapp

1ec5:
This PR adds a Layer Filter panel to the bottom of the macosapp document window that edits the selected layer’s predicate. A standard NSPredicateEditor bound to the style layer array controller provides the draggable, nestable rule editing UI. A [custom subclass of NSPredicateEditorRowTemplate](http://funwithobjc.tumblr.com/post/1646098126/creating-a-simple-nspredicateeditorrowtemplate) allows for arbitrary key paths in a single template and also represents an aggregate value (for the “is one of” operator) as a token field. (NSPredicateEditorRowTemplate only supports simple values by default. An editable table view would’ve been better, but NSPredicateEditor assumes fixed-height rows.) The predicate editor has a standard row template for compound predicates and a custom row template for each major value type (string, integer, floating-point number, etc.). Edits to this UI take effect immediately.

filter

As of writing, this editor is able to display predicates for the vast majority of layers in the Mapbox Streets style, and basic criteria can be edited without problems. Adding a criterion doesn’t quite work yet, and editing an “is one of” criterion often doesn’t round-trip correctly.

* [x] Add predicate editor in split view
* [x] Hook up menu items and split view delegate
* [x] Bind predicate editor to style layer array controller
* [x] Add common operators to row templates
* [x] Subclass row templates for arbitrary key paths
* [x] Replace text field with token field for aggregate values
* [ ] Prefer number templates over string template based on existing value type
* [ ] Add template for Boolean and null values
* [ ] Add “is defined” and “is not defined” operators
* [ ] Make key text field editable when adding a new criterion
* [ ] Round-trip “is one of” criteria accurately
* [ ] Fix UI after changing “is one of” to another operator
* [ ] Add “is between” operator or transform to compound operator
* [ ] Add template for `$type` key that uses a segmented control
* [ ] Track predicate changes in undo manager
* [ ] Add text fields for layer source, source layer

(For consistency with macOS and Apple applications, “filter” is the user-facing term, while “predicate” remains the API term to avoid confusion with other “filter” concepts.)

Depends on #7548.


1ec5:
This PR depends on some of the fixes in #7548.

jfirebaugh:
Stale.

PR #7560, opened 2016-12-30
Fixing some compilation errors

AndersBillLinden:
Hello mapbox team/tmpsantos!

This commit fixes two compilation errors coming from template depth and using a class after it has been declared. I got errors when I compiled the code with clang.

The maximum template depth problem should actually be fixed by not using tuples as I have understood from the #C++ channel on freenode.


tmpsantos:
Thanks @AndersBillLinden, I will incorporate your changes to my GCC4.9 branch.

AndersBillLinden:
Observe that this pull request is broken, https://github.com/mapbox/mapbox-gl-native/pull/7561 however seems to have conflicts I have not seen before... I am confused... why does it seem like 8 of your commit are into PR? I am not used to git on linux yet, apparently.

tmpsantos:
@AndersBillLinden this is maybe because I squashed your changes into other commits.

PR #7564, opened 2017-01-01
Arrange layers top to bottom in sidebar

1ec5:
#7484 reversed the `MGLStyle.layers` property to arrange the layers from back to front. In a vertical list, however, the layers should always be arranged from front to back, as if the user is viewing a cross section of the style. This change implements a `reversedLayers` property on MGLStyle via a category specific to macosapp. (Neither SDK is affected.)

Depends on #7539.


1ec5:
#7539 has landed, so this is ready for review.

PR #7570, opened 2017-01-02
Implement MGLForegroundStyleLayer.sourceIdentifier

### LINKED ISSUE #7557: MGLForegroundStyleLayer.sourceIdentifier is unimplemented
MGLForegroundStyleLayer declares a `sourceIdentifier` property, but its concrete subclasses (such as MGLLineStyleLayer) don’t implement a getter or setter for the property, so attempting to use the property will trigger a crash. The underlying mbgl classes do expose getters and setters for this property, so it should just be a matter of hooking them up in the generated source code.

/cc @frederoni

frederoni:
Fixes #7557 

This PR adds a sourceIdentifier getter for all concrete subclasses of MGLForegroundStyleLayer.

### DIFF
--- a/platform/darwin/src/MGLCircleStyleLayer.mm
+++ b/platform/darwin/src/MGLCircleStyleLayer.mm
@@ -54,6 +54,13 @@ namespace mbgl {
     super.rawLayer = rawLayer;
 }
 
+- (NSString *)sourceIdentifier
+{
+    MGLAssertStyleLayerIsValid();
+    
+    return @(self.rawLayer->getSourceID().c_str());
+}
+
 - (NSString *)sourceLayerIdentifier
 {
     MGLAssertStyleLayerIsValid();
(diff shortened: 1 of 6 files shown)


jfirebaugh:
#7557 claims that the underlying core classes contain getters and setters for source ID, but in fact setters did not exist and are being added in this PR. Changing the source ID is an operation that's more complex than simply updating the data member in the impl. Beyond that, the source tiles will need to be reparsed, and there are places in core where layers are grouped by source that will need to be updated. We'll need to add test cases to the test-suite, and port this feature to GL JS. Overall, adding a setter may turn out to be a rather complex change. Are we sure it's necessary, especially at such a late date in the release cycle?

frederoni:
Let's drop the changes in core and just keep the getter in order to fix the crash in #7557

1ec5:
My mistake: we should make the SDK property read-only. Sorry for the confusion.

Issue #7026, opened 2016-11-11
Flickering labels when panning


kkaefer:
Seeing this at a panning operation. What's happening is this:
- Step 1: tiles are loaded and placed
- Step 2: an adjacent tile becomes visible in the viewport, some (but not all!) of the labels in the previously loaded tiles vanish
- Step 3: some of the new tiles complete loading, and some of the labels that disappeared are showing again (see lower left)
- Step 4: all new tiles completed loading, and all labels are back to normal

![untitled1](https://cloud.githubusercontent.com/assets/52399/20222855/38123158-a837-11e6-80fa-c7c8ca2c5bb6.jpg)
![untitled2](https://cloud.githubusercontent.com/assets/52399/20222858/383cf640-a837-11e6-800b-b8c53b7908a8.jpg)
![untitled3](https://cloud.githubusercontent.com/assets/52399/20222856/3837141e-a837-11e6-85ca-1e95d030737f.jpg)
![untitled4](https://cloud.githubusercontent.com/assets/52399/20222857/383ce45c-a837-11e6-8b6a-bcc72b064964.jpg)

boundsj:
This seems like it might be related https://github.com/mapbox/mapbox-gl-native/issues/7003

kkaefer:
given https://github.com/mapbox/mapbox-gl-native/issues/7003#issuecomment-260009584, I don't think this is related

1ec5:
I think this is the behavior @amyleew and I were seeing in #5946.

jfirebaugh:
This is reproducible on the iOS 3.4.0 release branch but not the 3.3.7 tag.

jfirebaugh:
From code comment linked above:

```
            // We're not clipping symbol layers, so when we have both parents and children of symbol
            // layers, we drop all children in favor of their parent to avoid duplicate labels.
            // See https://github.com/mapbox/mapbox-gl-native/issues/2482
```

[...]

PR #7574, opened 2017-01-02
Unify update and render into a single step

### LINKED ISSUE #7040: Flickering when crossing international date line
When panning the map across the international date line that map tiles seem to flicker:

ezgif com-video-to-gif

cc: @kkaefer

kkaefer:
Update only when, and just prior to, rendering, giving no opportunity to interleave unexpected state changes. Abandon the use of `AsyncTask` in continuous mode; instead call `backend.invalidate()` whenever something occurs that needs to trigger a update/render, relying on the platform implementation of invalidate to perform coalescing.

This fixes #7040.

/cc @jfirebaugh @pveugen

### DIFF
--- a/src/mbgl/map/map.cpp
+++ b/src/mbgl/map/map.cpp
@@ -63,8 +63,8 @@ public:
     void onStyleError() override;
     void onResourceError(std::exception_ptr) override;
 
-    void update();
     void render(View&);
+    void renderStill();
 
     void loadStyleJSON(const std::string&);
 
@@ -83,7 +83,6 @@ public:
     MapDebugOptions debugOptions { MapDebugOptions::NoDebug };
 
     Update updateFlags = Update::Nothing;
-    util::AsyncTask asyncUpdate;
 
     std::unique_ptr<AnnotationManager> annotationManager;
     std::unique_ptr<Painter> painter;
@@ -96,10 +95,11 @@ public:
 
     std::unique_ptr<AsyncRequest> styleRequest;
 
-    std::unique_ptr<StillImageRequest> stillImageRequest;
     size_t sourceCacheSize;
-    TimePoint timePoint;
     bool loading = false;
+
+    util::AsyncTask asyncInvalidate;
+    std::unique_ptr<StillImageRequest> stillImageRequest;
 };
 
 Map::Map(Backend& backend,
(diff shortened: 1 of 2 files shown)


src/mbgl/map/map.cpp
```diff
@@ -276,46 +249,70 @@ void Map::Impl::update() {
...
+            renderState = RenderState::Fully;
+            backend.notifyMapChange(MapChangeDidFinishRenderingMapFullyRendered);
+            if (loading) {
+                loading = false;
+                backend.notifyMapChange(MapChangeDidFinishLoadingMap);
+            }
+        }
+
+        if (style->hasTransitions()) {
+            onUpdate(Update::RecalculateStyle);
```
jfirebaugh: In this and the `painter->needsAnimation()` case, should we be `|`'ing `Update::RecalculateStyle` / `Update::Repaint` with `flags` (from `updateTransitions`)?

src/mbgl/map/map.cpp
```diff
@@ -276,46 +249,70 @@ void Map::Impl::update() {
...
+                              mode,
+                              contextMode,
+                              debugOptions };
+
+        painter->render(*style,
+                        frameData,
+                        view,
+                        annotationManager->getSpriteAtlas());
+
+        painter->cleanup();
```
jfirebaugh: Can `Painter::cleanup` be called internally at the end of `Painter::render`?

kkaefer: In the static rendering case, we're calling `cleanup` after the callback (meaning that we don't have to wait for the cleanup to happen before the invoking implementation can do something with the result)

jfirebaugh:
Labels are failing to fade in/out all the way:

image

kkaefer:
Looks like this is caused by calling `invalidate` and triggering a new frame while the current frame is still in progress. In GLFW, we are setting the `dirty` flag to `false` after we rendered (which means it's always set to false), but we could just revert that. However, other platforms like macOS and iOS also suffer from the similar issue. I'll move the update call back to using a `util::AsyncTask`

src/mbgl/map/map.cpp
```diff
@@ -142,8 +142,14 @@ Map::Impl::Impl(Map& map_,
...
       pixelRatio(pixelRatio_),
-      asyncUpdate([this] { update(); }),
-      annotationManager(std::make_unique<AnnotationManager>(pixelRatio)) {
+      annotationManager(std::make_unique<AnnotationManager>(pixelRatio)),
+      asyncInvalidate([this] {
+          if (mode == MapMode::Continuous) {
+              backend.invalidate();
+          } else {
+              renderStill();
+          }
```
kkaefer: Moved the invalidation call back to happen after the async. The change to master is now that before, we had a `render() => asyncUpdate() => update() => invalidate() => render() ` flow. It's now shortened to `render() => asyncInvalidate() => invalidate() => update+render()`

jfirebaugh:
I'm noticing lots of unpleasant label flickering when testing this, but I suspect it's a pre-existing issue such as #7026.

kkaefer:
@jfirebaugh yes; I re-checked with master and saw the same types of flickering as documented in #7026

PR #7576, opened 2017-01-02
Split NSValue+MGLStyleEnumAttributeAdditions category per style layer type

1ec5:
Split the NSValue+MGLStyleEnumAttributeAdditions category into one category for each style layer type and renamed the category files to have a slightly more wieldy name.


1ec5:
In 3fb2489c0933a668f0522680cf572c04fe5d8bc6, I decided to move the categories into their respective style layer headers, so that they’d be closer to the enumeration declarations. (It also simplifies the code a bit.) Unfortunately, that change triggers realm/jazzy#539: now all the declarations are under “NSValue (MGLCircleStyleLayerAdditions)”. Unless I find a fix for the upstream issue, I’ll revert the change and leave it for another day.

1ec5:
A fix for the upstream issue is in realm/jazzy#719.

1ec5:
Merging now that realm/jazzy#719 has landed. We’ll have to make sure to generate the documentation using jazzy on master.

Issue #7575, opened 2017-01-02
Remove all deprecated methods


zugaldia:
The next Android release cycle (5.x) will break SEMVER compatibility with previous versions. This is a good moment to remove all `@Deprecated` members from the codebase. 

I'm still seeing 7 matches across 5 files:

![image](https://cloud.githubusercontent.com/assets/6964/21594509/5b14c91c-d0f1-11e6-9dff-7d07db66d406.png)

@mapbox/android: Any good reasons to keep any of these methods around one more version?

tobrun:
- [ ] Deprecation Emerald-v8:
    - good to remove, it was replaced with outdoors in v9
- [ ] Deprecation setStyleUrl:
  - there is some confusion related to this deprecation after discussing this with @cammace . This deprecation was valid if we didn't revert versioned styling methods in https://github.com/mapbox/mapbox-gl-native/issues/5193. If we decide to keep as-is implementations we should remove deprecation, if we would move to versioned style methods in Style.java we can remove the method fully.
- [ ] Deprecation textureMode:
   - I wouldn't mind keeping this around until it starts blocking implementation of other features (eg. GLSurfaceView). We currently advice using the texture mode to workaround the issue of showing a map in a dialog #6101 or if ViewMarker synchronisation should be more performant. 
- [ ] Deprecation dismiss tracking on gesture
   - need to double check but I think these are good to remove. They were replaced with methods specific to certain gestures for more granular control. 
- [ ]  setAccessToken/getAccessToken in OfflineManager
   - Other deprecations related to accesstoken have been rerouted to use `Mapbox.java`. It seems that this occurrence was missed. Will create a PR and link to this issue and the other accesstoken deprecation issue in #5621 for MapView and MapboxMap.

olliSP:
Before deprecated methods are removed, please take this issue into account:

https://github.com/mapbox/mapbox-gl-native/issues/7584

Issue #5621, opened 2016-07-08
Remove setAccessToken Methods from MapboxMap and MapView


bleege:
`MapboxAccountManager` needs to become the only way to set Access Tokens for apps.  To do this `MapView.setAccessToken()`, `MapboxMap.setAccessToken()`, and related methods need to be removed.  `MapView.setAccessToken()` and `MapboxMap.setAccessToken()` were officially deprecated and documented for eventual removal when `MapboxAccountManager` was introduced in `Mapbox Android SDK 4.1.0` ( #5004 ).

/cc @mapbox/mobile

PR #7582, opened 2017-01-03
[android] - cleanup accesstoken in OfflineManager

tobrun:
Refs #7575 and #5621 + moves validation of the access token inside Mapbox#getAccessToken.
Review @zugaldia

### DIFF
--- a/platform/android/MapboxGLAndroidSDK/src/main/java/com/mapbox/mapboxsdk/Mapbox.java
+++ b/platform/android/MapboxGLAndroidSDK/src/main/java/com/mapbox/mapboxsdk/Mapbox.java
@@ -39,7 +39,21 @@ public final class Mapbox {
    * @return Mapbox Access Token
    */
   public static String getAccessToken() {
+    validateAccessToken();
     return INSTANCE.accessToken;
+  }
+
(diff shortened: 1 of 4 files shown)


platform/android/MapboxGLAndroidSDK/src/main/java/com/mapbox/mapboxsdk/Mapbox.java
```diff
@@ -39,7 +39,13 @@ private Mapbox(@NonNull Context context, @NonNull String accessToken) {
    * @return Mapbox Access Token
    */
   public static String getAccessToken() {
-    return INSTANCE.accessToken;
+    String accessToken = INSTANCE.accessToken;
+    // validate if accessToken is correct
```
zugaldia: Just a small suggestion for clarity, could we extract the validation code into a `private void validateAccessToken()` method inside `Mapbox` so that the `getAccessToken()` is simply `validateAccessToken(INSTANCE.accessToken); return INSTANCE.accessToken;`?

tobrun:
@zugaldia 0d659ff addresses the requested changes.

PR #7585, opened 2017-01-03
Distance formatter fixes #3331

### LINKED ISSUE #3331: MGLDistanceFormatter

frederoni:
Fixes #3331

This PR adds a formatter that is meant to be used for geographic distances. (i.e. #7432)

It supports localization and pluralization. By default, it uses the user‘s current locale and opts for feet over yards unlike NSLengthFormatter.

### DIFF
--- a/platform/darwin/test/MGLDistanceFormatterTests.m
+++ b/platform/darwin/test/MGLDistanceFormatterTests.m
@@ -1,0 +1,100 @@
+#import <XCTest/XCTest.h>
+#import <Mapbox/Mapbox.h>
+
+@interface MGLDistanceFormatter(Private)
+@property (nonatomic) NSNumberFormatter *numberFormatter;
+@end
+
+@interface MGLDistanceFormatterTests : XCTestCase
+@end
+
+@implementation MGLDistanceFormatterTests
+
+- (void)testMetricAbbreviated {
+    MGLDistanceFormatter *formatter = [[MGLDistanceFormatter alloc] init];
+    formatter.units = MGLDistanceFormatterUnitsMetric;
+    formatter.unitStyle = MGLDistanceFormatterUnitStyleAbbreviated;
+    
+    XCTAssertEqualObjects([formatter stringFromDistance:0], [self testStringWithDistance:0 formatter:formatter unit:@"m"]);
+    XCTAssertEqualObjects([formatter stringFromDistance:10.5], [self testStringWithDistance:11 formatter:formatter unit:@"m"]);
+    XCTAssertEqualObjects([formatter stringFromDistance:50], [self testStringWithDistance:50 formatter:formatter unit:@"m"]);
+    XCTAssertEqualObjects([formatter stringFromDistance:500], [self testStringWithDistance:500 formatter:formatter unit:@"m"]);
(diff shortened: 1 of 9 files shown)


platform/darwin/test/MGLDistanceFormatterTests.m
```diff
@@ -0,0 +1,100 @@
...
+    XCTAssertEqualObjects([formatter stringFromDistance:0], [self testStringWithDistance:0 formatter:formatter unit:@"yards"]);
+    XCTAssertEqualObjects([formatter stringFromDistance:0.9144], [self testStringWithDistance:1 formatter:formatter unit:@"yard"]);
+    XCTAssertEqualObjects([formatter stringFromDistance:457.2], [self testStringWithDistance:500 formatter:formatter unit:@"yards"]);
+    XCTAssertEqualObjects([formatter stringFromDistance:913.486], [self testStringWithDistance:999 formatter:formatter unit:@"yards"]);
+    XCTAssertEqualObjects([formatter stringFromDistance:1609.34], [self testStringWithDistance:1 formatter:formatter unit:@"mile"]);
+    XCTAssertEqualObjects([formatter stringFromDistance:2414.02], [self testStringWithDistance:1.5 formatter:formatter unit:@"miles"]);
+    XCTAssertEqualObjects([formatter stringFromDistance:3218.68], [self testStringWithDistance:2 formatter:formatter unit:@"miles"]);
+}
+
+- (NSString *)testStringWithDistance:(CLLocationDistance)distance formatter:(MGLDistanceFormatter *)formatter unit:(NSString *)unit {
```
boundsj: In XCTest, I think we should avoid writing utility methods that start with `test`.

frederoni: It only recognizes methods prefixed with `test` if the return type is void but I agree.

1ec5:
It looks like you’ve completely reverse engineered NSLengthFormatter with some adjustments for distance-based measurements. That’s an admirable accomplishment. 👏 However, I think we should try to make this implementation more maintainable by leveraging an NSLengthFormatter internally and tweaking its output as needed. That would remove the need for hard-coded conversions and most localizable strings.

NSLengthFormatter’s behavior is almost exactly what we want. The only two differences are that we want distance-based rounding behavior, and a fractional mileage should be displayed in miles (ideally as a vulgar fraction or mixed number of miles by default) instead of yards. Besides maintainability, another important benefit of NSLengthFormatter is that it always matches the user’s system language and region settings, even if the application doesn’t. As written, MGLDistanceFormatter could potentially output something inconsistent like “૨૬ meters” based on the user’s system language.

For comparison, the Swift implementation mentioned in https://github.com/mapbox/mapbox-gl-native/issues/3331#issuecomment-214121278 weighs in at only [110 lines](https://gist.github.com/1ec5/791ccda066f7255b39d134f9c18c3506), including an “elaborate hack” that uses vulgar fractions for U.S. mileages while tiptoeing around any potential localization issues. Without that hack, it would be a mere 40 lines.

frederoni:
Closing in favor of #7592

PR #7586, opened 2017-01-03
[core] Get rid of user-specified refs

### LINKED ISSUE #6900: Get rid of user-specified refs

### LINKED ISSUE #7572: Changing ref properties must unlink existing refs
"Ref properties" are 'type', 'source', 'source-layer', 'minzoom', 'maxzoom', 'filter', and 'layout'. Of these, the first three are read-only (except see #7570 for a proposal to make 'source' read-write).

Change any of these properties must at minimum unlink layers that were originally linked via the "ref" mechanism, such that they act independently. Ideally, new possible refs are autodetected -- that's #6900. But short of that, this issue tracks the fact that the existing runtime style API will not behave as expected if you modify, say, filter or layout properties of a layer that was originally a `ref`.

jfirebaugh:
Fixes #6900
Fixes #7572

### DIFF
--- a/src/mbgl/style/group_by_layout.cpp
+++ b/src/mbgl/style/group_by_layout.cpp
@@ -1,0 +1,51 @@
+#include <mbgl/style/group_by_layout.hpp>
+#include <mbgl/style/layer.hpp>
+#include <mbgl/style/layer_impl.hpp>
+#include <mbgl/style/conversion/stringify.hpp>
+#include <mbgl/util/rapidjson.hpp>
+
+#include <rapidjson/writer.h>
+#include <rapidjson/stringbuffer.h>
+
+#include <unordered_map>
+
+namespace mbgl {
+namespace style {
+
+std::string layoutKey(const Layer& layer) {
+    using namespace conversion;
+
+    rapidjson::StringBuffer s;
+    rapidjson::Writer<rapidjson::StringBuffer> writer(s);
+
+    writer.StartArray();
+    writer.Uint(static_cast<uint32_t>(layer.type));
+    writer.String(layer.baseImpl->source);
+    writer.String(layer.baseImpl->sourceLayer);
+    writer.Double(layer.baseImpl->minZoom);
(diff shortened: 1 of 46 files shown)


src/mbgl/style/group_by_layout.cpp
```diff
@@ -0,0 +1,51 @@
...
+    writer.Uint(static_cast<uint32_t>(layer.baseImpl->visibility));
+    stringify(writer, layer.baseImpl->filter);
+    layer.baseImpl->stringifyLayout(writer);
+    writer.EndArray();
+
+    return s.GetString();
+}
+
+std::vector<std::vector<std::unique_ptr<Layer>>> groupByLayout(std::vector<std::unique_ptr<Layer>> layers) {
+    std::unordered_map<std::string, std::vector<std::unique_ptr<Layer>>> map;
```
1ec5: `std::vector<std::unique_ptr<Layer>>` is getting repeated quite a bit. Do you think the code would benefit from a `Layout` typedef or somesuch?

src/mbgl/util/color.cpp
```diff
@@ -17,4 +18,12 @@ optional<Color> Color::parse(const std::string& s) {
     }};
 }
 
+std::string Color::stringify() const {
+    return "rgba(" +
```
1ec5: Does it matter that a `Color` created via an `hsla()` string would get converted back to JSON as an `rgba()` string?

jfirebaugh: Named colors will also come back as `rgba()` strings; there are no guarantees that values are round-tripped exactly. This is consistent with GL JS and the CSSOM.

src/mbgl/style/group_by_layout.cpp
jfirebaugh: This method is short enough that I think it's ok.

PR #7594, opened 2017-01-04
[Qt] Add more docs

tmpsantos:
Add even more docs to QMapboxGL.

### DIFF
--- a/platform/qt/src/qmapboxgl.cpp
+++ b/platform/qt/src/qmapboxgl.cpp
@@ -827,21 +827,6 @@ void QMapboxGL::setPaintProperty(const QString& layer_, const QString& property,
     }
 }
 
-bool QMapboxGL::isRotating() const
-{
-    return d_ptr->mapObj->isRotating();
-}
-
-bool QMapboxGL::isScaling() const
-{
-    return d_ptr->mapObj->isScaling();
-}
-
-bool QMapboxGL::isPanning() const
-{
-    return d_ptr->mapObj->isPanning();
-}
-
 bool QMapboxGL::isFullyLoaded() const
 {
     return d_ptr->mapObj->isFullyLoaded();
--- a/platform/qt/src/qmapboxgl.cpp
+++ b/platform/qt/src/qmapboxgl.cpp
@@ -141,14 +141,14 @@ std::unique_ptr<const mbgl::SpriteImage> toSpriteImage(const QImage &sprite) {
 /*!
     \enum QMapboxGLSettings::GLContextMode
 
-    This enum sets the expectations for the GL state.
+    This enum sets the expectations for the OpenGL state.
 
-    \value UniqueGLContext  The GL context is only used by QMapboxGL, so it is not
+    \value UniqueGLContext  The OpenGL context is only used by QMapboxGL, so it is not
     reset before each rendering. Use this mode if the intention is to only draw a
     fullscreen map.
 
-    \value SharedGLContext  The GL context is shared and the state will be restored
-    before rendering. This mode is safer when GL calls are performed prior of after
+    \value SharedGLContext  The OpenGL context is shared and the state will be restored
+    before rendering. This mode is safer when OpenGL calls are performed prior of after
     we call QMapboxGL::render for rendering a map.
 
     \sa contextMode()
@@ -202,7 +202,7 @@ QMapboxGLSettings::QMapboxGLSettings()
 
 /*!
     Returns the OpenGL context mode. This is specially important when mixing
-    with other GL draw calls.
+    with other OpenGL draw calls.
 
     By default, it is set to QMapboxGLSettings::SharedGLContext.
 */
@@ -285,7 +285,7 @@ void QMapboxGLSettings::setCacheDatabaseMaximumSize(unsigned size)
     \l {https://github.com/mapbox/mapbox-gl-native/blob/master/bin/offline.sh}
     {Offline Tool}.
 
-    By default, it is set to \b :memory: meaning it will create an in-memory
+    By default, it is set to \c :memory: meaning it will create an in-memory
     cache instead of a file on disk.
 */
 QString QMapboxGLSettings::cacheDatabasePath() const
@@ -296,7 +296,7 @@ QString QMapboxGLSettings::cacheDatabasePath() const
 /*!
     Sets the cache database \a path.
 
-    Setting the \a path to \b :memory: will create an in-memory cache.
+    Setting the \a path to \c :memory: will create an in-memory cache.
 */
 void QMapboxGLSettings::setCacheDatabasePath(const QString &path)
 {
@@ -305,7 +305,7 @@ void QMapboxGLSettings::setCacheDatabasePath(const QString &path)
 
 /*!
     Returns the asset path, which is the root directory from where
-    the \b asset:// scheme gets resolved in a style. \b asset:// can be used
+    the \c asset:// scheme gets resolved in a style. \c asset:// can be used
     for loading a resource from the disk in a style rather than fetching
     it from the network.
 
@@ -327,7 +327,7 @@ void QMapboxGLSettings::setAssetPath(const QString &path)
 /*!
     Returns the access token.
 
-    By default, it is taken from the environment variable \b MAPBOX_ACCESS_TOKEN
+    By default, it is taken from the environment variable \c MAPBOX_ACCESS_TOKEN
     or empty if the variable is not set.
 */
 QString QMapboxGLSettings::accessToken() const {
@@ -381,9 +381,59 @@ void QMapboxGLSettings::setApiBaseUrl(const QString& url)
     QNetworkAccessManager for HTTP requests and QString for UTF-8 manipulation.
 
     QMapboxGL is not thread-safe and it is assumed that it will be accessed from
-    the same thread as the thread where the GL context lives.
+    the same thread as the thread where the OpenGL context lives.
 
     \since 4.7
+*/
+
+/*!
+    \enum QMapboxGL::MapChange
+
+    This enum represents the last changed occurred to the map state.
+
+    \value MapChangeRegionWillChange                      A region of the map will change, like
+    when resizing the map.
+
+    \value MapChangeRegionWillChangeAnimated              Not in use by QMapboxGL.
+
+    \value MapChangeRegionIsChanging                      A region of the map is changing.
+
+    \value MapChangeRegionDidChange                       A region of the map finished changing.
+
+    \value MapChangeRegionDidChangeAnimated               Not in use by QMapboxGL.
+
+    \value MapChangeWillStartLoadingMap                   The map is getting loaded. This state
+    is set only once right after QMapboxGL is created and a style is set.
+
+    \value MapChangeDidFinishLoadingMap                   All the resources were loaded and parsed
+    and the map is fully rendered. After this state the mapChanged() signal won't fire again unless
+    the is some client side interaction with the map or a tile expires, causing a new resource
+    to be requested from the network.
+
+    \value MapChangeDidFailLoadingMap                     An error occurred when loading the map.
+
+    \value MapChangeWillStartRenderingFrame               Just before rendering the frame. This
+    is the state of the map just after calling render() and might happened many times before
+    the map is fully loaded.
+
+    \value MapChangeDidFinishRenderingFrame               The current frame was rendered but was
+    left in a partial state. Some parts of the map might be missing because have not arrived
+    from the network or are being parsed.
+
+    \value MapChangeDidFinishRenderingFrameFullyRendered  The current frame was fully rendered.
+
+    \value MapChangeWillStartRenderingMap                 Set once when the map is about to get
+    rendered for the first time.
+
+    \value MapChangeDidFinishRenderingMap                 Not in use by QMapboxGL.
+
+    \value MapChangeDidFinishRenderingMapFullyRendered    Map is fully loaded and rendered.
+
+    \value MapChangeDidFinishLoadingStyle                 The style was loaded.
+
+    \value MapChangeSourceDidChange                       A source has changed.
+
+    \sa mapChanged()
 */
 
 /*!
@@ -445,7 +495,7 @@ void QMapboxGL::cycleDebugOptions()
 
     Sets a new \a style from a JSON that must conform to the
     \l {https://www.mapbox.com/mapbox-gl-style-spec/}
-    {Mapbox Style Specification}.
+    {Mapbox style specification}.
 
     \note In case of a invalid style it will trigger a mapChanged
     signal with QMapboxGL::MapChangeDidFailLoadingMap as argument.
@@ -506,8 +556,9 @@ void QMapboxGL::setLatitude(double latitude_)
     \property QMapboxGL::longitude
     \brief the map current longitude in degrees.
 
-    Setting a longitude doesn't necessarily mean it will be accepted since QMapboxGL
-    might constrain it within the limits of the Web Mercator projection.
+    Setting a longitude beyond the limits of the Web Mercator projection will make
+    the map wrap. As an example, setting the longitude to 360 is effectively the same
+    as setting it to 0.
 */
 double QMapboxGL::longitude() const
 {
@@ -524,8 +575,11 @@ void QMapboxGL::setLongitude(double longitude_)
     \brief the map scale factor.
 
     This property is used to zoom the map. When \a center is defined, the map will
-    scale in the direction of the center pixel coordinates. This function could be used for
-    implementing a pinch gesture or zooming by using the mouse scroll wheel.
+    scale in the direction of the center pixel coordinates. The \a center will remain
+    at the same pixel coordinate after scaling as before calling this method.
+
+    \note This function could be used for implementing a pinch gesture or zooming
+    by using the mouse scroll wheel.
 
     \sa zoom()
 */
@@ -559,11 +613,21 @@ void QMapboxGL::setZoom(double zoom_)
     d_ptr->mapObj->setZoom(zoom_, d_ptr->margins);
 }
 
+/*!
+    Returns the minimum zoom level allowed for the map.
+
+    \sa maximumZoom()
+*/
 double QMapboxGL::minimumZoom() const
 {
     return d_ptr->mapObj->getMinZoom();
 }
 
+/*!
+    Returns the maximum zoom level allowed for the map.
+
+    \sa minimumZoom()
+*/
 double QMapboxGL::maximumZoom() const
 {
     return d_ptr->mapObj->getMaxZoom();
@@ -583,12 +647,14 @@ Coordinate QMapboxGL::coordinate() const
     return Coordinate(latLng.latitude, latLng.longitude);
 }
 
-void QMapboxGL::setCoordinate(const Coordinate &coordinate_)
+void QMapboxGL::setCoordinate(const QMapbox::Coordinate &coordinate_)
 {
     d_ptr->mapObj->setLatLng(mbgl::LatLng { coordinate_.first, coordinate_.second }, d_ptr->margins);
 }
 
 /*!
+    \fn QMapboxGL::setCoordinateZoom(const QMapbox::Coordinate &coordinate, double zoom)
+
     Convenience method for setting the \a coordinate and \a zoom simultaneously.
 
     \note Setting \a coordinate and \a zoom at once is more efficient than doing
@@ -597,7 +663,7 @@ void QMapboxGL::setCoordinate(const Coordinate &coordinate_)
     \sa zoom()
     \sa coordinate()
 */
-void QMapboxGL::setCoordinateZoom(const Coordinate &coordinate_, double zoom_)
+void QMapboxGL::setCoordinateZoom(const QMapbox::Coordinate &coordinate_, double zoom_)
 {
     d_ptr->mapObj->setLatLngZoom(
             mbgl::LatLng { coordinate_.first, coordinate_.second }, zoom_, d_ptr->margins);
@@ -678,27 +744,38 @@ void QMapboxGL::setPitch(double pitch_)
     d_ptr->mapObj->setPitch(pitch_);
 }
 
+/*!
+    Returns the north orientation mode.
+*/
 QMapboxGL::NorthOrientation QMapboxGL::northOrientation() const
 {
     return static_cast<QMapboxGL::NorthOrientation>(d_ptr->mapObj->getNorthOrientation());
 }
 
+/*!
+    Sets the north orientation mode to \a orientation.
+*/
 void QMapboxGL::setNorthOrientation(NorthOrientation orientation)
 {
     d_ptr->mapObj->setNorthOrientation(static_cast<mbgl::NorthOrientation>(orientation));
 }
 
-void QMapboxGL::setGestureInProgress(bool inProgress)
+/*!
+    Tells the map rendering engine that there is currently a gesture in \a progress. This
+    affects how the map renders labels, as it will use different texture filters if a gesture
+    is ongoing.
+*/
+void QMapboxGL::setGestureInProgress(bool progress)
 {
-    d_ptr->mapObj->setGestureInProgress(inProgress);
+    d_ptr->mapObj->setGestureInProgress(progress);
 }
 
 /*!
     Adds an \a className to the list of active classes. Layers tagged with a certain class
     will only be active when the class is added.
 
-    This was removed from \l {https://www.mapbox.com/mapbox-gl-style-spec/#layer-paint.*}
-    {the vector tiles specification} and should no longer be used.
+    This was removed from the \l {https://www.mapbox.com/mapbox-gl-style-spec/#layer-paint.*}
+    {Mapbox style specification} and should no longer be used.
 
     \deprecated
     \sa removeClass()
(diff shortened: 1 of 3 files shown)


platform/qt/src/qmapboxgl.cpp
```diff
@@ -366,6 +366,54 @@ void QMapboxGLSettings::setAccessToken(const QString &token)
...
+    \value MapChangeRegionWillChangeAnimated              Not in use by QMapboxGL.
+
+    \value MapChangeRegionIsChanging                      A region of the map is changing.
+
+    \value MapChangeRegionDidChange                       A region of the map finished changing.
+
+    \value MapChangeRegionDidChangeAnimated               Not in use by QMapboxGL.
+
+    \value MapChangeWillStartLoadingMap                   The map is getting loaded. This state
+    is set only once right after QMapboxGL is created and a style is set.
```
1ec5: Link `QMapboxGL`?

platform/qt/src/qmapboxgl.cpp
```diff
@@ -657,19 +716,30 @@ void QMapboxGL::setPitch(double pitch_)
     d_ptr->mapObj->setPitch(pitch_);
 }
 
+/*!
+    Returns the North Orientation mode.
```
1ec5: Lowercase “north orientation”.

platform/qt/src/qmapboxgl.cpp
```diff
@@ -657,19 +716,30 @@ void QMapboxGL::setPitch(double pitch_)
...
+*/
 void QMapboxGL::setNorthOrientation(NorthOrientation orientation)
 {
     d_ptr->mapObj->setNorthOrientation(static_cast<mbgl::NorthOrientation>(orientation));
 }
 
-void QMapboxGL::setGestureInProgress(bool inProgress)
+/*!
+    Tells the map rendering engine that there is currently a gesture in \a progress. This
+    affects how the map renders labels, as it will use different texture filters if a gesture
```
1ec5: Is “texture filter” an appropriate concept to mention in public documentation? Where is it defined? Or is this an implementation detail that we should handwave about?

platform/qt/src/qmapboxgl.cpp
```diff
@@ -749,26 +827,57 @@ mbgl::Annotation fromPointAnnotation(const PointAnnotation &pointAnnotation) {
...
+*/
+void QMapboxGL::removeAnnotation(QMapbox::AnnotationID id)
 {
-    d_ptr->mapObj->removeAnnotation(annotationID);
+    d_ptr->mapObj->removeAnnotation(id);
 }
 
+/*!
+    Sets a layout \a property \a value to an existing \a layer_. The \a property string can be any
+    as defined by \l {https://www.mapbox.com/mapbox-gl-style-spec/} {the vector tiles specification}
```
1ec5: The link points to the style specification, not the vector tile specification (×5).

platform/qt/src/qmapboxgl.cpp
```diff
@@ -785,6 +894,21 @@ void QMapboxGL::setLayoutProperty(const QString& layer_, const QString& property
     }
 }
 
+/*!
+    Sets a paint \a property \a value to an existing \a layer_. The \a property string can be any
+    as defined by \l {https://www.mapbox.com/mapbox-gl-style-spec/} {the vector tiles specification}
+    for paint properties.
+
+    The argument \a klass_ is deprecated and is used for defining the style class for the paint
```
1ec5: Does the compiler used with Qt support a deprecation attribute? We could remove the `klass_` parameter from this method and overload the method with a deprecated one that takes `klass_`.

platform/qt/src/qmapboxgl.cpp
```diff
@@ -785,6 +894,21 @@ void QMapboxGL::setLayoutProperty(const QString& layer_, const QString& property
...
+    property.
+
+    For paint properties that take a color as \a value, such as \b "fill-color", a string such as
+    \b "blue" can be passed or a QColor.
```
1ec5: Are there any other special cases? Perhaps a mapping from style specification types to Q-types would be appropriate here. This guide doesn’t have to be as longwinded as #7542, but something simple could head off newbie mistakes. I’m curious, for example, about the type `line-dasharray` takes.

platform/qt/src/qmapboxgl.cpp
```diff
@@ -806,27 +930,47 @@ void QMapboxGL::setPaintProperty(const QString& layer_, const QString& property,
...
+    Returns true when the map is completely rendered, false otherwise. A partially
+    rendered map ranges from nothing rendered at all to only labels missing.
+*/
 bool QMapboxGL::isFullyLoaded() const
 {
     return d_ptr->mapObj->isFullyLoaded();
 }
 
+/*!
+    Pan the map by \a offset in pixels.
```
1ec5: Does `QPointF` or a related type specify an origin, or is it up to us to say whether a positive `y` is interpreted as an offset up or down? I see that we had to specify an origin in the `pixelForCoordinate()` documentation.

platform/qt/src/qmapboxgl.cpp
```diff
@@ -806,27 +930,47 @@ void QMapboxGL::setPaintProperty(const QString& layer_, const QString& property,
...
+*/
 void QMapboxGL::moveBy(const QPointF &offset)
 {
     d_ptr->mapObj->moveBy(mbgl::ScreenCoordinate { offset.x(), offset.y() });
 }
 
+/*!
+    Scale the map by \a scale_ in the direction of the \a center. This function
```
1ec5: “In the direction of the center” may not be clear to someone unfamiliar with implementing map gestures. Perhaps note that `center` corresponds to the same coordinate before and after calling this method.

platform/qt/src/qmapboxgl.cpp
```diff
@@ -806,27 +930,47 @@ void QMapboxGL::setPaintProperty(const QString& layer_, const QString& property,
...
+*/
 void QMapboxGL::rotateBy(const QPointF &first, const QPointF &second)
 {
     d_ptr->mapObj->rotateBy(
             mbgl::ScreenCoordinate { first.x(), first.y() },
             mbgl::ScreenCoordinate { second.x(), second.y() });
 }
 
+/*!
+    Resize the map to \a size and scale to fit at \a framebufferSize.
```
1ec5: What’s the relationship between `size` and `framebufferSize`?

platform/qt/src/qmapboxgl.cpp
```diff
@@ -854,32 +998,51 @@ void QMapboxGL::addAnnotationIcon(const QString &name, const QImage &icon)
     d_ptr->mapObj->addAnnotationIcon(name.toStdString(), toSpriteImage(icon));
 }
 
-QPointF QMapboxGL::pixelForCoordinate(const Coordinate &coordinate_) const
+/*!
+    Returns the offset in pixels for \a coordinate_. The origin pixel coordinate is
+    located at the top left corner of the map view.
+
+    It will work for any coordinate, even if the coordinate is not currently visible
```
1ec5: s/It will work/This method returns the correct value/

But also note that the return value depends on the zoom level, rotation, and pitch.

platform/qt/src/qmapboxgl.cpp
```diff
@@ -925,12 +1088,31 @@ QMargins QMapboxGL::margins() const
     );
 }
 
-void QMapboxGL::addSource(const QString &sourceID, const QVariantMap &params)
+/*!
+    Adds a data source \a id to the map as specified by \l
+    {https://www.mapbox.com/mapbox-gl-style-spec/#root-sources}{the vector tiles specification} with
+    \a params.
+
+    A data source starting with \b ":" stands for a file tracked by the Qt resource system. In this
```
1ec5: Does the data source start with “:”, or does the file path?

platform/qt/src/qmapboxgl.cpp
```diff
@@ -964,15 +1152,30 @@ void QMapboxGL::updateSource(const QString &sourceID, const QVariantMap &params)
     }
 }
 
-void QMapboxGL::removeSource(const QString& sourceID)
+/*!
+    Removes the source \a id.
+
+    Has no effect if the source does not exist.
```
1ec5: s/Has/This method has/

platform/qt/src/qmapboxgl.cpp
```diff
@@ -964,15 +1152,30 @@ void QMapboxGL::updateSource(const QString &sourceID, const QVariantMap &params)
...
 
+/*!
+    Adds a custom layer \a id with the initialization function \a initFn, the
+    render function \a renderFn and the deinitialization function \a deinitFn with
+    the user data \a context_ before the existing layer \a before.
+
+    This is used for delegate the rendering to the user of this API and is not
+    officially supported. Use at your own risk.
+
+    \deprecated
```
1ec5: Deprecating this method implies that it could be removed in a future release, not that it could change at any time. Changing or removing a deprecated method is usually regarded as a backwards-incompatible change to the library. (In the case of semver, that means a major revision bump.)

In the iOS and macOS SDKs, we’ve signaled that the MGLOpenGLStyleLayer API is not officially supported by leaving it undocumented (at least undocumented in public headers). This is consistent with a longstanding Apple convention, but I wonder if there’s a similar convention in Qt-land that calling an undocumented method is discouraged.

platform/qt/src/qmapboxgl.cpp
```diff
@@ -991,6 +1194,23 @@ void QMapboxGL::addCustomLayer(const QString &id,
             before ? mbgl::optional<std::string>(before) : mbgl::optional<std::string>());
 }
 
+/*!
+    Adds a style layer to the map as specified by \l
+    {https://www.mapbox.com/mapbox-gl-style-spec/#root-layers}{the vector tiles specification} with
+    \a params.
+
+    This example shows how to add a layer that will be used to show a route line on the map. Note
+    that nothing will be drawn until we set paint properties using setPaintProperty().
```
1ec5: Note that the source must be defined before the layer is added. As I understand it, changing a source after adding a layer that depends on the source results in undefined behavior.

platform/qt/src/qmapboxgl.cpp
```diff
@@ -1005,23 +1225,55 @@ void QMapboxGL::addLayer(const QVariantMap &params)
...
+    Removes the layer \a id.
+*/
 void QMapboxGL::removeLayer(const QString& id)
 {
     d_ptr->mapObj->removeLayer(id.toStdString());
 }
 
-void QMapboxGL::addImage(const QString &name, const QImage &sprite)
+/*!
+    Adds the \a image with the identifier \a id that can be used
```
1ec5: Is it possible to set an existing ID to a different image using this method?

platform/qt/src/qmapboxgl.cpp
```diff
@@ -1005,23 +1225,55 @@ void QMapboxGL::addLayer(const QVariantMap &params)
...
+    Removes the image \a id.
+*/
+void QMapboxGL::removeImage(const QString &id)
 {
-    d_ptr->mapObj->removeImage(name.toStdString());
+    d_ptr->mapObj->removeImage(id.toStdString());
 }
 
+/*!
+    Adds a \a filter_ to a style \a layer_ accordingly to the \l
```
1ec5: `accordingly to` → `according to` → `using the format described in`

platform/qt/src/qmapboxgl.cpp
```diff
@@ -1005,23 +1225,55 @@ void QMapboxGL::addLayer(const QVariantMap &params)
...
+    {https://www.mapbox.com/mapbox-gl-style-spec/#types-filter}{the vector tiles specification}.
+
+    Given a layer \b "marker" from an arbitrary GeoJSON source containing features of type \b
```
1ec5: There are two instances each of “GeoJSON source” and “GeoJSON data source”. Best to be consistent, so that developers don’t think there may be a distinction between the two.

platform/qt/src/qmapboxgl.cpp
```diff
@@ -1005,23 +1225,55 @@ void QMapboxGL::addLayer(const QVariantMap &params)
...
+    "Point" and \b "LineString", this example shows how to make sure the marker will only tag
```
1ec5: The example shows setting the filter of a style layer, not the filter of a marker.

platform/qt/src/qmapboxgl.cpp
```diff
@@ -1005,23 +1225,55 @@ void QMapboxGL::addLayer(const QVariantMap &params)
...
+    \code
+        QVariantList filterExpression;
+        filterExpression.append("==");
+        filterExpression.append("$type");
+        filterExpression.append("Point");
+
+        QVariantList filter;
+        filter.append(filterExpression);
+
+        map->setFilter("maker", filter);
```
1ec5: s/maker/marker/

platform/qt/src/qmapboxgl.cpp
1ec5: Or is this method actually deprecated because `render()` can be used instead?

platform/qt/src/qmapboxgl.cpp
```diff
@@ -1062,6 +1314,15 @@ void QMapboxGL::setFilter(const QString& layer_, const QVariant& filter_)
     qWarning() << "Layer doesn't support filters";
 }
 
+/*!
+    Renders the map using OpenGL draw calls. If \a fbo is passed, it will
+    make sure to bind the framebuffer object before drawing, otherwise a
```
1ec5: s/drawing, otherwise/drawing; otherwise/

platform/qt/src/qmapboxgl.cpp
```diff
@@ -1062,6 +1314,15 @@ void QMapboxGL::setFilter(const QString& layer_, const QVariant& filter_)
...
+    valid GL context is expected with an appropriated GL viewport state set
```
1ec5: s/appropriated/appropriate/

1ec5: In general, I don’t think “GL” is an appropriate abbreviation for “OpenGL” in public-facing documentation. Better to spell out “OpenGL”.

platform/qt/src/qmapboxgl.cpp
```diff
@@ -1062,6 +1314,15 @@ void QMapboxGL::setFilter(const QString& layer_, const QVariant& filter_)
...
+    for the size of the canvas.
+
+    This function should be called only after the signal needsRendering() is
```
1ec5: Is this documentation being generated by Doxygen or something along those lines? This sounds like a good case for Doxygen’s `\pre` command.

platform/qt/src/qmapboxgl.cpp
```diff
@@ -1085,11 +1346,42 @@ void QMapboxGL::render()
 }
 #endif
 
+/*!
+    Informs the map that the network connection was established, causing
+    all timeouted network requests to retry immediately.
```
1ec5: s/was established/has been established/
s/timeouted network requests/network requests that previously timed out/
s/to retry/to be retried/

platform/qt/src/qmapboxgl.cpp
```diff
@@ -1085,11 +1346,42 @@ void QMapboxGL::render()
...
+    This signal is emitted when the state of the map has changed. This signal
+    may be used for detecting errors when loading a style or detecting when
+    a map is fully loaded by analyzing the parameter \a change.
+*/
+
+/*!
+    \fn void QMapboxGL::copyrightsChanged(const QString &copyrightsHtml);
+
+    This signal is emitted when the copyrights of the current content of the map
+    has change. This can be caused by a style change or adding a new data source.
```
1ec5: s/has change/have changed/

platform/qt/src/qmapboxgl.cpp
```diff
@@ -366,6 +366,54 @@ void QMapboxGLSettings::setAccessToken(const QString &token)
 */
 
 /*!
+    \enum QMapboxGL::MapChange
+
+    This enum represents the last changed occurred on the map state.
```
1ec5: s/last change occurred on/last change that occurred to/

platform/qt/src/qmapboxgl.cpp
```diff
@@ -366,6 +366,54 @@ void QMapboxGLSettings::setAccessToken(const QString &token)
...
+
+    \value MapChangeRegionDidChange                       A region of the map finished changing.
+
+    \value MapChangeRegionDidChangeAnimated               Not in use by QMapboxGL.
+
+    \value MapChangeWillStartLoadingMap                   The map is getting loaded. This state
+    is set only once right after QMapboxGL is created and a style is set.
+
+    \value MapChangeDidFinishLoadingMap                   All the resources were loaded and parsed
+    and the map is fully rendered. This is the final state of the map.
```
1ec5: “Final” implies that the state will never change again, doesn’t it? This enum value sort of represents a quiescent state, but the map state can continue changing if for instance the user starts panning.

platform/qt/src/qmapboxgl.cpp
tmpsantos: `qdoc` links automatically in this case.

https://tmpsantos.com.br/qmapboxgl/qmapboxgl.html#MapChange-enum

platform/qt/src/qmapboxgl.cpp
tmpsantos: Correct, I will reword this. What I was trying to say here is this is the most complete state of the map in terms of things rendered and resources loaded.

platform/qt/src/qmapboxgl.cpp
```diff
@@ -562,21 +621,21 @@ Coordinate QMapboxGL::coordinate() const
...
 
-void QMapboxGL::setCoordinate(const Coordinate &coordinate_)
+void QMapboxGL::setCoordinate(const QMapbox::Coordinate &coordinate_)
 {
     d_ptr->mapObj->setLatLng(mbgl::LatLng { coordinate_.first, coordinate_.second }, d_ptr->margins);
 }
 
 /*!
-    Convenience method for setting the \a coordinate and \a zoom simultaneously.
+    Convenience method for setting the \a coordinate_ and \a zoom_ simultaneously.
```
tmpsantos: That is true. I've been trying when possible and being creative with the names. 

In this case, the Qt standard is to use `coordinate()` as a getter instead of `getCoordinate()` and this often overlaps with parameters names. For Qt itself it is not a problem because it doesn't care about parameter shadowing.

One solution would be do a post processing of the generated documentation or disable this warning for the Qt build.

platform/qt/src/qmapboxgl.cpp
tmpsantos: Ugh, nice catch.

platform/qt/src/qmapboxgl.cpp
```diff
@@ -749,26 +827,57 @@ mbgl::Annotation fromPointAnnotation(const PointAnnotation &pointAnnotation) {
...
 
+/*!
+    Sets a layout \a property \a value to an existing \a layer_. The \a property string can be any
+    as defined by \l {https://www.mapbox.com/mapbox-gl-style-spec/} {the vector tiles specification}
+    for layout properties.
+
+    This example hides the layer \b "route":
+
+    \code
+        map->setLayoutProperty("makerBackground", "visibility", "none");
```
tmpsantos: It should be "route" actually.

platform/qt/src/qmapboxgl.cpp
tmpsantos: I think in this case it is fine, because this is the "Qt low-level API" that could be used for implementing higher level bindings like the QML component we are pushing to Qt 5.9.

platform/qt/src/qmapboxgl.cpp
tmpsantos: needsRendering() gets detected and linked by `qtdoc` automatically because it is a signal of the same class.

platform/qt/src/qmapboxgl.cpp
tmpsantos: I replaced it with `styleClass`.

platform/qt/src/qmapboxgl.cpp
1ec5: Doh, I keep forgetting that frameworks other than jazzy can autolink without any markup. 👍

platform/qt/src/qmapboxgl.cpp
1ec5: Ah, that’s unfortunate. In Objective-C and Swift, it’s also possible for a getter and a variable to share the same namespace, and `_` always denotes an internal implementation detail that shouldn’t be touched. So one old convention is to say e.g. `theCoordinate` or `aZoom`, but I don’t know if that looks reasonable in Qt code.

Is it possible to document this method on its declaration rather than its implementation? Then you can give the parameter two different names, `coordinate` in the header (which shouldn’t conflict with `coordinate()`) and `coordinate_` here in the implementation.

platform/qt/src/qmapboxgl.cpp
1ec5: Sorry, was referring to the paragraph as a whole, not the reference to `needsRendering()`. In Doxygen syntax, [`\pre`](https://www.stack.nl/~dimitri/doxygen/manual/commands.html#cmdpre) means “precondition” and given special prominence.

platform/qt/src/qmapboxgl.cpp
```diff
@@ -735,6 +805,14 @@ QStringList QMapboxGL::getClasses() const
     return classNames;
 }
 
+/*!
+    Sets the \a duration and \a delay of style class transitions. An animation
+    morphing a style property value to another value is performed when a new
+    class is set.
+
+    \deprecated
```
1ec5: As far as I know, [global transition options](https://www.mapbox.com/mapbox-gl-style-spec/#root-transition) are not deprecated and not under consideration for removal.

platform/qt/src/qmapboxgl.cpp
1ec5: My point was that you’d have two methods:

```cpp
void setPaintProperty(const QString& layer, const QString& property, const QVariant& value);
//! \deprecated
void setPaintProperty(const QString& layer, const QString& property, const QVariant& value, const QString& klass) __attribute__((deprecated));
```

This works in C++03; is there a Qt-specific reason why we can’t overload the method like this?

platform/qt/src/qmapboxgl.cpp
tmpsantos: Hmmm, I'm revising this. This is actually the behavior on the QML API, not the low level. My bad.

platform/qt/src/qmapboxgl.cpp
tmpsantos: I tried to find something like `\unsupported` tag in the same lines as `\deprecated` on `qdoc` but could not find. Making it as `\deprecated` at least moves it to a almost hidden section of the generated HTML.

I do agree with you, it is misleading. I'm going to add a `\warning` tag.

There is no such thing as unsupported public API on Qt.

platform/qt/src/qmapboxgl.cpp
tmpsantos: If the name is the same, the image get replaced only if the dimensions are the same (I just looked it up). I'm going to document this.

platform/qt/src/qmapboxgl.cpp
tmpsantos: Ah, now I got it.

The problem is that `klass` has a default value, so the overload is ambiguous:
```
void setPaintProperty(const QString &layer, const QString &property, const QVariant &value, const QString &klass = QString());
```

platform/qt/src/qmapboxgl.cpp
tmpsantos: Excellent idea, it works! We can get rid of the `_`.

platform/qt/src/qmapboxgl.cpp
1ec5: I’d consider that a bug: #7618.

platform/qt/src/qmapboxgl.cpp
1ec5: The second version wouldn’t need a default value anymore, given the first version.

tmpsantos:
@1ec5 thanks a lot for pushing this documentation to the next level: https://tmpsantos.com.br/qmapboxgl/qmapboxgl.html

platform/qt/src/qmapboxgl.cpp
```diff
@@ -749,99 +833,227 @@ mbgl::Annotation fromPointAnnotation(const PointAnnotation &pointAnnotation) {
...
+    \row
+        \li Number
+        \li \c int, \c double or \c float
+    \row
+        \li Array
+        \li QVariantList
+    \endtable
+
+    For a giver layer \c route, setting the property \c line-dasharray that requires
+    a Mabpox Style type \b Array can be done as follows using a QVariantList:
```
1ec5: > If the style specification defines the property’s type as `Array`, use a QVariantList. For example, the following code sets a `route` layer’s `line-dasharray` property:

platform/qt/src/qmapboxgl.cpp
```diff
@@ -749,99 +833,227 @@ mbgl::Annotation fromPointAnnotation(const PointAnnotation &pointAnnotation) {
...
+    for layout properties.
+
+    This example hides the layer \c route:
+
+    \code
+        map->setLayoutProperty("route", "visibility", "none");
+    \endcode
+
+    This table describes the mapping between \l {https://www.mapbox.com/mapbox-gl-style-spec/#types}
+    {style types} and Qt types accepted by setLayoutProperty():
```
1ec5: This table also applies to `setPaintProperty()`, so copy this table to that method’s documentation. (In particular, color-typed properties are paint properties.)

tmpsantos: I did that ;)

But I'm going to remove Color.

PR #7595, opened 2017-01-04
Happy new year! 🎉

1ec5:
Here’s to an odder, more prime year than 2016.


1ec5:
Cherry-picked into the release-ios-v3.4.0 branch as e162d4f4c80989bf369e3471ee7f55c72f364b03.

PR #7372, opened 2016-12-10
[core] Add support for data-driven styling

jfirebaugh:
Core implementation of data-driven styling. Now needs SDK bindings.

Fixes #4860.
Fixes #6518.

Issue #6721, opened 2016-10-17
MGLStyleFunction needs a colorSpaceModel property


1ec5:
The iOS and macOS SDKs need to support the new color interpolation option being added in ~~#6442~~ #8720 via a new `colorSpaceModel` property on MGLStyleFunction. No need to extend the initializers, since we can assume a default color space model of [RGB](https://developer.apple.com/reference/coregraphics/cgcolorspacemodel/1408867-rgb).

Ideally, the property would be of type [CGColorSpaceModel](https://developer.apple.com/reference/coregraphics/cgcolorspacemodel) or [NSColorSpaceModel](https://developer.apple.com/reference/appkit/nscolorspacemodel). However, we’d have to provide our own HCL color space model and document the fact that some of the common color space models provided by Core Graphics or AppKit, like grayscale, are unavailable. So it might be easier, if less cool, to define our own enumeration.

/cc @friedbunny @tmcw @zugaldia

1ec5:
I think we probably will have to provide our own enumeration for this type, because CGColorSpaceModel is new as of iOS 9 and macOS 10.11, and NSColorSpaceModel is new as of macOS 10.10. But we can be consistent with naming at least.

jfirebaugh:
We removed functions. If we bring back color spaces, it'll be as an expression.

1ec5:
#8720 has been repurposed to track the addition of an expression operator. There would still need to be some platform-specific work, so this issue could be reopened to track that when the time comes.

PR #7596, opened 2017-01-04
[ios, macos] Add support for data driven styling

boundsj:
iOS portion of https://github.com/mapbox/mapbox-gl-native/issues/4860
depends on: https://github.com/mapbox/mapbox-gl-native/pull/7372
related to: https://github.com/mapbox/mapbox-gl-native/pull/7752

#### Implement DDS getters and setters

- [x] Setters
  - [x] Data Driven Property Value
    - [x] Constant value
    - [x] Camera Function with
      - [x] Exponential stops
      - [x] Interval stops
    - [x] Source Function with
      - [x] categorical stops
        - [x] string attribute key
        - [x] bool attribute key
        - [x] int attribute key
        - [x] handle default value
      - [x] exponential stops (compiles and runs but nothing appears)
      - [x] interval stops (compiles and runs but nothing appears)
      - [x] identity stops
    - [x] Composite function
      - [x] Stops with categorical functions
      - [x] Stops with interval functions
      - [x] Stops with exponential functions
  - [x] Property Value
    - [x] Non-enum
      - [x] Constant value
      - [x] Camera function
    - [x] Enum
      - [x] Constant value
      - [x] Camera function
- [x] Getters
  - [x] Get Data Driven Property Values
    - [x] Constant
    - [x] Camera Function with
      - [x] Exponential stops
      - [x] Interval stops
    - [x] Source Function with
      - [x] categorical stops
        - [x] string attribute key
        - [x] bool attribute key
        - [x] int attribute key
        - [x] handle default value
      - [x] exponential stops
      - [x] interval stops
      - [x] identity stops
    - [x] Composite function
      - [x] Stops with categorical functions
      - [x] Stops with interval functions
      - [x] Stops with exponential functions
  - [x] Get Property Values
    - [x] Non-enum
      - [x] Constant value
      - [x] Camera function
    - [x] Enum
      - [x] Constant value
      - [x] Camera function


boundsj:
@jfirebaugh 

- The below classes were modified to support round tripping values. Do the changes ([example](https://github.com/mapbox/mapbox-gl-native/pull/7596/files#diff-1228c5ab48d531c703b3ad37fa5b7182)) make sense? 
  - data_driven_property_value.hpp
  - camera_function.hpp
  - categorical_stops.hpp
  - composite_function.hpp
  - source_function.hpp

- What is the expected behavior of source functions with exponential (or interval) stops? This statement compiles and runs with no errors but causes the layer it is applied to to disappear completely: 
```
mbgl::style::SourceFunction<MBGLType> sourceFunction = {function.attributeName.UTF8String, exponentialStops}
``` 


- Noting here that the work described in the second to last checkbox in https://github.com/mapbox/mapbox-gl-native/issues/4860#issue-151269718 is probably required to eliminate an issue where DDS property changes don't take effect immediately in some cases. I'm also seeing the property function styling flicker (i.e. a color is applied to a circle / the circle is colored black) from time to time when the map is tilted. I'm not sure if that is related to the yet to be done worker round trip implementation.

jfirebaugh:
@boundsj I'm not seeing where those additions are used elsewhere in the diff. Can you point me to where they were needed?

The omission of methods such as `isConstant`/`isCameraFunction`/`asCameraFunction` was by design, and I'd like to remove the existing such methods from `PropertyValue` as well. Using the visitor functionality of `variant` is a safer way to implement type-specific dispatching.

jfirebaugh:
Ah, GitHub was hiding parts of the diff so they didn't show up in text search.

Yes, [code such as this](https://github.com/mapbox/mapbox-gl-native/blob/9cfb48088dc6b7c808c3d55b2290c73cf3e446f9/platform/darwin/src/MGLStyleValue_Private.h#L38-L46) should instead be written using a visitor ([example](https://github.com/mapbox/mapbox-gl-native/blob/2d113c7f4ff87a45728277dee58cfd1f77571895/src/mbgl/style/conversion/stringify.hpp#L222-L343)). One advantage of doing so is that future additions to the variant types will produce a compile-time error until the code is updated to handle the new type.

1ec5:
This PR has the potential to introduce a lot of classes, more than we’d want the developer to have to juggle. The point of implement MGLStyleValue as a class cluster in the first place was to reduce the number of classes the developer would have to know about in order to use style functions in a basic way. I’d prefer if we could somehow draw class-level distinctions based on only the most important axis (in terms of mapbox/mapbox-gl-style-spec#613) and represent any other axes with enum-typed properties or optional properties/parameters.

> A composite (aka "zoom and property") function that can have exponential,interval, categorial zoom portion stops can be one of MGLStyleCompositeExponentialFunction, MGLStyleCompositeIntervalFunction, MGLStyleCompositeCategoricalFunction (the interval and categorical composite functions are not actually implemented at this time).

Can we implement composite functions through composition? That is, could we represent an exponential composite function as an MGLStyleExponentialFunction with functions as stop values? This does mean shifting some validation of the stop values from design time to runtime, but I think this is a reasonable concession.

While we’re going about renaming classes, the class names proposed here are a bit ungrammatical. A subclass usually inserts an adjective right after the class prefix. (I’ve always felt weird about MGLStyleConstantValue; it should be MGLConstantStyleValue.) For example:

* MGLStyleExponentialFunction → MGLExponentialStyleFunction
* MGLStyleIntervalFunction → MGLIntervalStyleFunction
* etc.

platform/darwin/src/MGLStyleValue.h
```diff
@@ -57,6 +60,29 @@ NS_ASSUME_NONNULL_BEGIN
...
+#pragma mark Source function values
+
+// TODO: API docs
++ (instancetype)valueWithAttributeName:(NSString *)attributeName categoricalStops:(NSDictionary<id, MGLStyleValue<T> *> *)categoricalStops defaultValue:(MGLStyleValue<T> * _Nullable)defaultValue;
+
+// TODO: API docs
++ (instancetype)valueWithAttributeName:(NSString *)attributeName exponentialStops:(NSDictionary<id, MGLStyleValue<T> *> *)exponentialStops;
+
+// TODO: API docs
++ (instancetype)valueWithAttributeName:(NSString *)attributeName base:(CGFloat)base exponentialStops:(NSDictionary<id, MGLStyleValue<T> *> *)exponentialStops;
```
boundsj: oops. thanks for the reminder

boundsj:
@jfirebaugh thanks for pointing me towards the variants and visitor pattern. The previous implementation has been replaced with a set of [evaluators](https://github.com/mapbox/mapbox-gl-native/blob/boundsj-dds-darwin-wip/platform/darwin/src/MGLStyleValue_Private.h#L445-L470) and [visitors](https://github.com/mapbox/mapbox-gl-native/blob/boundsj-dds-darwin-wip/platform/darwin/src/MGLStyleValue_Private.h#L413-L422). In addition to the benefit you mentioned, I think this approach makes it easier to follow the partitions of the conversion of functions and stops in the implementation.

> The omission of methods such as isConstant/isCameraFunction/asCameraFunction was by design, and I'd like to remove the existing such methods from PropertyValue as well.

I've been able to eliminate all references to PropertyValue's `is(as)Constant` in the getter converters (the refactor I just completed). I should be able to eliminate them in the setter converters next. This should help make it possible to refactor PropertyValue as you said.

boundsj:
@1ec5 thanks for notes in https://github.com/mapbox/mapbox-gl-native/pull/7596#issuecomment-270592798. Here is an updated proposal for this API:

The principle axis of function categories are:

-  __Camera__      -- zoom-driven, possible stops: exponential, interval stops
- __Source__       -- data-driven: exponential, interval, categorical, identity stops
- __Composite__ -- both zoom and data driven (previously known as "zoom and property”): exponential, interval, categorical primary  stops and each stop can hold a `camera` or `source` function with exponential, interval, categorical stops

All of these mbgl types could be represented by the single existing class:

* `MGLStyleFunction`

Internally, `MGLStyleFunction` could gain a new enumeration ivar that allows it to keep track of where if falls on the primary axis. The ivar would be set based on the `MGLStyleFunction` factory method used to create the function. 

The stops axis could be represented by a publicly exposed enumeration with the values:

 - `MGLPropertyFunctionTypeExponential`
- `MGLPropertyFunctionTypeInterval`
 - `MGLPropertyFunctionTypeCategorical`
- `MGLPropertyFunctionTypeIdentity`

With these classes,  `MGLStyleValue` could expose the entire DDS API as follows:

### Camera functions

- `cameraFunctionValueWithType:Stops:InterpolationBase:`

param __type__: `MGLPropertyFunctionTypeExponential` or `MGLPropertyFunctionTypeInterval`
param __stops__: `NSDictionary<NSNumber *, MGLStyleValue<T> *> *` 
param __interpolationBase__:  `CGFloat` value used when the type is `exponential`

- `cameraFunctionValueWithType:Stops:`

When the type is `exponential` this is a convenience for `cameraFunctionValueWithType:Stops:InterpolationBase:` where `interpolationBase` is always `1.0`. When the type is `interval` this convenience method creates a camera
function with interval stops.

These methods would create an `MGLStyleFunction`  with an internal enum that specifies that it is a camera function that would be used to create a `mbgl::style::CameraFunction` with `mbgl::style::ExponentialStops` or `mbgl::style::IntervalStops` and a `base`.

### Source functions

- `sourceFunctionValueWithType:attributeName:stops:interpolationBase`

param __type__: `MGLPropertyFunctionTypeExponential`, `MGLPropertyFunctionTypeInterval`, `MGLPropertyFunctionTypeCategorical`, `MGLPropertyFunctionTypeIdentity`
param __attributeName__: String value representing the attribute key to connect the function to
param __stops__: Optional `NSDictionary<id, MGLStyleValue<T> *> *` value used when the type is `exponential`, `interval`, or `categorical`
param __interpolationBase__:  `CGFloat` value used when the `type` is `exponential`

- `sourceFunctionValueWithType:attributeName:stops`

When the type is `exponential` this is a convenience for `sourceFunctionValueWithType:attributeName:stops:interpolationBase` where `interpolationBase` is always `1.0`. When the type is `interval`, `categorial`, or `identity` this convenience method creates a source function with the appropriate type of stops.

- `sourceFunctionIdentityValueWithAttributeName:`

Convenience method that creates an identity function. This could be included since an identity function does not need any input except for the attribute key name.

These methods would create an `MGLStyleFunction` with an internal enum that specifies that it is a source function. This would be used to create a `mbgl::style::SourceFunction` with `mbgl::style::ExponentialStops`, `mbgl::style::IntervalStops`, `mbgl::style::CategoricalStops`, or `mbgl::style::IdentityStops`, an `attributeName`, and a `base`. 

### Composite functions

- `compositeFunctionValueWithType:Stops:InterpolationBase:`

param __type__:  `MGLPropertyFunctionTypeExponential`, `MGLPropertyFunctionTypeInterval`, `MGLPropertyFunctionTypeCategorical`
param __stops__: A `NSDictionary<id, MGLStyleValue<T> *> *` with `MGLStyleFunction` values that are `exponential`, `interval`, or `categorical`  `camera` or `source` functions
param __interpolationBase__: `CGFloat` value used when the type is `exponential`

The composite function is similar to `cameraFunctionValueWithType:Stops:InterpolationBase:` but take dictionaries with `MGLStyleFunction` values for the stops instead of constants.

This method would create an `MGLStyleFunction` with an internal enum that specifies that it is a composite function. This would be used to create a `mbgl::style::CompositeFunction` with `mbgl::style::ExponentialStops`, `mbgl::style::IntervalStops`, `mbgl::style::CategoricalStops` that hold `mbgl::style::SourceFunction` or `mbgl::style::CameraFunction` as values.

### Legacy function support

The following legacy methods could continue to be exposed and will create functions that are now known as `camera` functions on the primary axis.

`valueWithStops:`
`valueWithInterpolationBase:stops:`

boundsj:
> @1ec5 Can we implement composite functions through composition? That is, could we represent an exponential composite function as an MGLStyleExponentialFunction with functions as stop values? This does mean shifting some validation of the stop values from design time to runtime, but I think this is a reasonable concession.

https://github.com/mapbox/mapbox-gl-native/pull/7596#issuecomment-272938731 proposes essentially that. Note that mbgl requires the attribute name to be specified up front (i.e. `mbgl::style::CompositeFunction<MBGLType> compositeFunction = {function.attributeName.UTF8String, compositeStops};` so we could extract that string out of the function based stops passed in. Stops with functions with a mixture of attribute names don't make sense so we may want to assert that they are the same.

boundsj:
@1ec5 when thinking about the proposal in https://github.com/mapbox/mapbox-gl-native/pull/7596#issuecomment-272938731, one alternative that comes to mind is to create a set of classes to represent the stop types. We could still classify that secondary axis with an internal enum value  in `MGLStyleFunction` but, if the stops were actual classes (like they are in mbgl) I think that would make the public facing API less error prone.

For example, instead of creating a kind of source function that has interval stops like this:

```
[MGLStyleValue<UIColor *> sourceFunctionValueWithType:MGLPropertyFunctionTypeInterval :attributeName:@"foo" stops:stops]
```

we could offer something like this:

```
MGLIntervalStops *intervalStops = [[MGLIntervalStops alloc] initWithStops:stopsDictionary];
[MGLStyleValue<UIColor *> sourceFunctionValueWithAttributeName:@"foo" intervalStops:intervalStops]
```

I think the advantage of this is that it would help reduce mistakes like specifying that the function is 
exponential put passing stops with strings as keys (interval stops).

jfirebaugh:
When designing this API, you should expect that `interpolationBase` is not the only property that will exist that's specific to exponential stops. Exponential stops will also need a property to specify the color space (#6721).

boundsj:
> When designing this API, you should expect that interpolationBase is not the only property that will exist that's specific to exponential stops. Exponential stops will also need a property to specify the color space (#6721).

That reminds me, `interpolationBase` is currently taken in the the initializer for the MGL functions themselves. If we created classes to represent stops (as proposed in https://github.com/mapbox/mapbox-gl-native/pull/7596#issuecomment-272943821) then base and color space could be part of the stops object they are intended to be directly associated with. And, both of these values may make more sense as properties rather than extending the initializers.

Parts of https://github.com/mapbox/mapbox-gl-native/pull/7596#issuecomment-272938731 seem strange since base is a float that is only sometimes required depending on the enumerated type of the function. But base and color space properties would always make sense for a proper exponential stops object.

1ec5:
Thanks for the carefully thought-out API proposal. Class clusters come with a tradeoff: in exchange for keeping the number of public classes manageable, the one public class gains many initializers and a number of properties that may or may not be applicable, depending on which initializer was used to create the object. Therefore, misuse is usually detected at runtime rather than at build time.

So far, I’ve been drawing parallels to two class systems: the NSExpression class cluster and the NSPredicate class hierarchy (which I probably shouldn’t be calling a class cluster per se). In the NSExpression class cluster, the kind of expression is identified by an enumeration-typed argument and property, but internally each kind has its own private subclass. The reason NSExpression exposes a “flat” public API is that most kinds of NSExpression share roughly the same parameters, so the risk of accidentally using a non-applicable property is relatively low.

By contrast, the NSPredicate class hierarchy is essentially all public, at least two levels deep. In v3.4.0, MGLStyleValue was modeled on NSPredicate because MGLStyleConstantValue and MGLStyleFunction have such different parameters, the same way NSCompoundPredicate and NSComparisonPredicate share almost nothing in common.

If two kinds of functions are expected to have drastically different sets of properties, then they should be represented by two different public classes. At the same time, I don’t think it’s reasonable to have a separate function class for each _combination_ of traits. (On a practical level, that would lead to much more casting.) If the principle axis is indeed _camera_ versus _source_ versus _composite_, then we need to find an elegant way to expose properties related to the secondary axis without introducing too many properties that need magical optional values.

> When designing this API, you should expect that `interpolationBase` is not the only property that will exist that's specific to exponential stops. Exponential stops will also need a property to specify the color space (#6721).

To accommodate these options and also for forward compatibility, each function initializer could have an `options` parameter, and perhaps each class would have an `options` property as well. Each function initializer would continue to have an explicit parameter for anything that isn’t optional. This way we can avoid having to define magic “unset” values (easy for strings, but not as elegant for numbers).

While I appreciate that https://github.com/mapbox/mapbox-gl-native/pull/7596#issuecomment-272943821 would provide more build-time type safety, it would exacerbate the already awkwardly deep composition for certain types like enumerations. Imagine describing the value of a property thus:

> The property is set to a source function containing interval stops containing style constant values containing NSValue objects containing `MyEnumerationMiddle`.

I don’t think we could get away with another level of composition without introducing a format string syntax. 😬

ericrwolfe:
I think the term `cameraFunction` could be misleading and `zoomFunction` might be more appropriate and in line with the [style spec](https://www.mapbox.com/mapbox-gl-style-spec/#types-function). 

`MGLMapCamera` has the properties `centerCoordinate`, `heading`, `pitch`, and `altitude`, but none of these can currently be used by style functions, nor do they seem particularly useful to support.

jfirebaugh:
`CameraFunction` is a forward-looking name for when those other properties are supported. More discussion in https://github.com/mapbox/mapbox-gl-style-spec/issues/613.

boundsj:
@1ec5 thanks for https://github.com/mapbox/mapbox-gl-native/pull/7596#issuecomment-273036971 and highlighting the distinction between the class clusters and hierarchies we've used as a reference. I've compiled all the feedback so far into the proposal just below. I expect that we will need to continue to tweak it but it feels good enough (for now) for me to move forward on it again.

## API proposal (take 3)

### Function types

The primary function axis could be represented by three new classes:

  * `MGLCameraStyleFunction`
  * `MGLSourceStyleFunction`
  * `MGLCompositeStyleFunction`

The initializers for these classes would take an NSDictionary of options to control optional variables where applicable (i.e. `interpolationBase`). For 3.5.0, we expect the full set of valid option keys across all function types to be:
  
  - `MGLStyleFunctionOptionInterpolationBase` an NSNumber representing a float (default is 1.0)
  - `MGLStyleFunctionOptionColorSpace` an NSValue representing a TBD enumeration value (default will be RGB) (see https://github.com/mapbox/mapbox-gl-native/issues/6721)

### Stops
 
The stops axis could be represented by a publicly exposed enumeration with the values:

  * `MGLPropertyFunctionTypeExponential`
  * `MGLPropertyFunctionTypeInterval`
  * `MGLPropertyFunctionTypeCategorical`
  * `MGLPropertyFunctionTypeIdentity`

### Creating functions

With these classes and the new enumeration, the entire DDS API could be exposed as follows:

#### Camera functions

- `cameraFunctionValueWithType:stops:options:`

param __type__: `MGLPropertyFunctionTypeExponential` or `MGLPropertyFunctionTypeInterval`
param __stops__: `NSDictionary<id, MGLStyleValue<T> *> *` 
param __options__: optional `NSDictionary` containing `MGLStyleFunctionOptionInterpolationBase` and/or `MGLStyleFunctionOptionColorSpace` options for exponential stops

#### Source functions

- `sourceFunctionValueWithType:attributeName:stops:options`

param __type__: `MGLPropertyFunctionTypeExponential`, `MGLPropertyFunctionTypeInterval`, `MGLPropertyFunctionTypeCategorical`, `MGLPropertyFunctionTypeIdentity`
param __attributeName__: String value representing the attribute key to connect the function to
param __stops__: Optional `NSDictionary<id, MGLStyleValue<T> *> *` value used when the type is `MGLPropertyFunctionTypeExponential`, `MGLPropertyFunctionTypeInterval`, or `MGLPropertyFunctionTypeCategorical`
param __options__: optional `NSDictionary` containing `MGLStyleFunctionOptionInterpolationBase` and/or `MGLStyleFunctionOptionColorSpace` options when the type is `MGLPropertyFunctionTypeExponential`

#### Composite functions

Composite function objects represent an "outer" function with stops that can be exponential, interval, or categorial where each value is an "inner" `MGLSourceStyleFunction` with stops that can be exponential, interval, or categorial. 

- `compositeFunctionValueWithType:attributeName:stops:options`

param __type__:  `MGLPropertyFunctionTypeExponential`, `MGLPropertyFunctionTypeInterval`, `MGLPropertyFunctionTypeCategorical`
param __attributeName__: String value representing the attribute key to connect the function to. This value should be the same for each attribute key as the one used to create each function object in the stops
param __stops__: Primary stops `NSDictionary<id, MGLStyleValue<T> *> *` with inner `MGLSourceStyleFunction` values with `MGLPropertyFunctionTypeExponential`, `MGLPropertyFunctionTypeInterval`, or `MGLPropertyFunctionTypeCategorical` stops
param __options__: optional `NSDictionary` containing `MGLStyleFunctionOptionInterpolationBase` and/or `MGLStyleFunctionOptionColorSpace` options for exponential stops for the outer function

#### Legacy function support

`MGLStyleFunction` and the methods it introduced in the 3.4.0 release (listed below) could continue to be exposed (but deprecated) and would create mbgl camera functions under the hood.

`valueWithStops:`
`valueWithInterpolationBase:stops:`

1ec5:
> param **type**: `MGLPropertyFunctionTypeExponential` or `MGLPropertyFunctionTypeInterval`

Functions vary along two axes. In https://github.com/mapbox/mapbox-gl-style-spec/issues/613#issuecomment-265029389, there are camera, source, and composite types, and in https://github.com/mapbox/mapbox-gl-style-spec/issues/613#issue-192990141, there are exponential, interval, categorical, and identity… types? I think mbgl tries to avoid ever naming either axis, but we’d be forced to name both as soon as we document them or name an enumeration around one.

In https://github.com/mapbox/mapbox-gl-native/pull/7596#issuecomment-273654168, there are camera, source, and composite _function types_, and there are exponential, interval, categorical, and identity _property types_. But “property type” already exists as a concept: the `fillColor` property is of type UIColor (or technically MGLStyleValue).

Channeling math class, what do you think about describing camera, source, and composite _types_, but exponential, interval, categorical, and identity _domains_? Any other ideas?

boundsj:
@jfirebaugh tests are passing, but please hold off on merging this branch to `dds-wip` until there is final agreement on the platform level API and the API documentation is complete (this week). In that time, I also want to add more Swift tests of runtime and data-driven styling, update our generated examples in Darwin header files, and add DDS examples to iosapp.

Note: to support assertions in the Darwin tests, I backfilled an inequality operator in: https://github.com/mapbox/mapbox-gl-native/blob/boundsj-dds-darwin-wip/include/mbgl/style/data_driven_property_value.hpp#L29-L32

boundsj:
Here is an updated status report, with examples, on the state of the Darwin platform API:

### API

#### data-driven, camera function with exponential color stop values
```Swift
let stops = [
  0: MGLStyleValue<UIColor>(rawValue: .red),
  10: MGLStyleValue<UIColor>(rawValue: .red),
  15: MGLStyleValue<UIColor>(rawValue: .green)
]
circleStyleLayer.circleColor = MGLStyleValue<UIColor>.cameraFunctionValue(
  stopType: .exponential, 
  stops: stops, 
  options: [.interpolationBase: 10.0]
)
```

#### data-driven, source function with categorical color stop values with integer attribute keys
```Swift
let stops = [
  0: MGLStyleValue<UIColor>(rawValue: .green),
  100: MGLStyleValue<UIColor>(rawValue: .orange)
]
circleStyleLayer.circleColor = MGLStyleValue<UIColor>.sourceFunctionValue(
  stopType: .categorical,
  stops: stops,
  attributeName: "temp",
  options: [.defaultValue: MGLStyleValue<UIColor>(rawValue: .red)]
)
```

### Naming

> there are camera, source, and composite function types, and there are exponential, interval, categorical, and identity property types. But “property type” already exists as a concept: the fillColor property is of type UIColor (or technically MGLStyleValue).

As you can see above, I went with `stopType` as the parameter name and it gets set to a `MGLStyleFunctionStopType`. I chose that name because the "type of stop" is exactly what we want the user to tell us here and I could not come up with a better name for the enum.

### Composite functions

> Can we implement composite functions through composition? That is, could we represent an exponential composite function as an MGLStyleExponentialFunction with functions as stop values? This does mean shifting some validation of the stop values from design time to runtime, but I think this is a reasonable concession.

I ended up not implementing that approach because a) the core implementation truly expects a composite of stops with constant values and b) the API was much harder to use (and made less sense) if the developer had to make full on source functions to then stuff in stops (that we would then just throw away and take the stops from). It also did not make sense for the outer function to specify an `attributeName` and then have inner functions that _should_ each specify the same `attributeName` but may not.

The API we have now looks more like a Swift version of the js examples and Android implementation noted in https://github.com/mapbox/mapbox-gl-native/pull/7752#issuecomment-273818391

#### data-driven, composite function with inner exponential color stop values nested in outer camera stops
```Swift
let stops: [NSNumber: [NSNumber: MGLStyleValue<NSNumber>]] = [
  0: [0: smallRadiusValue],        // zoom level 0, temp value of 0 use small radius
  10: [200: smallRadiusValue],  // zoom level 10, temp value of 200 use small radius
  20: [200: largeRadiusValue]   // zoom level 20, temp value of 200 use small radius
]
circleStyleLayer.circleRadius = MGLStyleValue<NSNumber>.compositeFunctionValue(
  stopType: .exponential,
  stops: stops,
  attributeName: "temp",
  options: nil
)
```

As was the case before though, getting values back out requires some casting gymnastics:

```Swift
// get a value back
if let returnedCircleRadius = circleStyleLayer.circleRadius as? MGLCompositeStyleFunction<NSNumber> {
  if let returnedStops = returnedCircleRadius.stops as NSDictionary? as? [NSNumber: [NSNumber:   MGLStyleValue<NSNumber>]] {
  let lhs: MGLStyleValue<NSNumber> = returnedStops[0]!.values.first!
  let rhs: MGLStyleValue<NSNumber> = radiusCompositeExponentialOrIntervalStops[0]!.values.first!
                XCTAssertEqual(lhs, rhs)
  }
}
```

### Next steps

All of the examples above (and many more) [are available in MGLStyleValueTests.swift](https://github.com/mapbox/mapbox-gl-native/blob/boundsj-dds-darwin-wip/platform/darwin/test/MGLStyleValueTests.swift#L97). To finish up I need to:

- [ ] Address code review feedback already given that I've not taken care of so far
- [ ] Add even more coverage for the various source and composite function types in the generated Objective-C tests
- [ ] Update iOS app with a few examples and remove some of the temporary changes I've made to it for testing
- [ ] Initial API documentation

cc @1ec5 @incanus @jmkiley @jfirebaugh

boundsj:
Noting that node tests started failing here (and on `dds-wip`), ignoring that for now. cc @jfirebaugh

platform/darwin/src/MGLStyleValue.h
```diff
@@ -58,7 +78,16 @@ MGL_EXPORT
  @param stops A dictionary associating zoom levels with style values.
  @return An `MGLStyleFunction` object with the given interpolation base and stops.
  */
-+ (instancetype)valueWithInterpolationBase:(CGFloat)interpolationBase stops:(NSDictionary<NSNumber *, MGLStyleValue<T> *> *)stops;
++ (instancetype)valueWithInterpolationBase:(CGFloat)interpolationBase stops:(NSDictionary<NSNumber *, MGLStyleValue<T> *> *)stops __attribute__((deprecated("Use +cameraFunctionValueWithStopType:stops:options:")));
+
+// TODO: API docs
++ (instancetype)cameraFunctionValueWithStopType:(MGLStyleFunctionStopType)stopType stops:(NS_DICTIONARY_OF(id, MGLStyleValue<T> *) *)stops options:(nullable NS_DICTIONARY_OF(MGLStyleFunctionOption, id) *)options NS_SWIFT_NAME(cameraFunctionValue(stopType:stops:options:));
```
1ec5: The overridden Swift name notwithstanding, it feels awkward to name an initializer such that it bridges into Swift as `MGLStyleValue.cameraFunctionValue(…`, using the class method syntax, instead of `MGLStyleValue.init(…`, using the initializer syntax. For one thing, it means repetition at the call site, which Swift strives to avoid.

Naming the factory methods `+valueWith…` would have the unfortunate effect of obscuring the type (camera, source, etc.), which is important. To disambiguate the various factory methods at the call site, how about naming the selectors along the lines of `+valueWithStopType:cameraStops:options:`?

platform/darwin/src/MGLStyleValue.h
```diff
@@ -106,75 +135,74 @@ MGL_EXPORT
...
- as the zoom level changes. The layout and paint attribute properties of an
- `MGLStyleLayer` object can be set to `MGLStyleFunction` objects. Use a zoom
- level function to create the illusion of depth and control data density.
-
- The `MGLStyleFunction` class takes a generic parameter `T` that indicates the
- Foundation class being wrapped by this class.
- */
 MGL_EXPORT
-@interface MGLStyleFunction<T> : MGLStyleValue<T>
+@interface MGLCameraStyleFunction<T> : MGLStyleValue<T>
```
1ec5: A `@compatibility_alias` is needed from MGLStyleFunction to MGLCameraStyleFunction to maintain backwards compatibility with iOS SDK v3.4.0. Alternatively, we could keep MGLStyleFunction as the superclass of MGLCameraStyleFunction et al.

platform/darwin/src/MGLStyleValue.h
```diff
@@ -2,9 +2,27 @@
...
+typedef NSString *MGLStyleFunctionOption NS_STRING_ENUM;
+
+// TODO: API docs
+extern MGL_EXPORT const MGLStyleFunctionOption MGLStyleFunctionOptionInterpolationBase;
+
+// TODO: API docs
+extern MGL_EXPORT const MGLStyleFunctionOption MGLStyleFunctionOptionDefaultValue;
+
+// TODO: API docs
+typedef NS_ENUM(NSUInteger, MGLStyleFunctionStopType) {
```
1ec5: What term would refer to UIColor in `MGLCameraStyleValue<UIColor>`? No matter what we call it, I think developers would gravitate toward “stop type” or “type of stop” for that concept as well.

Instead of referring to exponential, interval, etc. as “stop types”, how about referring to them as “[interpolation modes](https://developer.apple.com/reference/spritekit/skinterpolationmode)”? `MGLInterpolationModeExponential` wouldn’t be _that_ much typing. 😛

jfirebaugh:
I agree with your choices on composite function design. I was even considering changing the core API so that it uses less composition, in order to reflect the style spec reality that an exponential stop interpolation base is provided once, function-wide, not on a per-inner-function basis.

platform/darwin/src/MGLStyleValue.h
boundsj: I like that and thanks for referencing the helpful SpriteKit example. My only concern is that, with DDS, the identify function actually uses the notion of stops only as a placeholder and interpolation is not a factor. I suppose that, no matter what we call this enum, some clarification about what an identity function is will be required.

platform/darwin/src/MGLStyleValue.h
boundsj: Done

boundsj:
@jfirebaugh I think this can be merged to `dds-wip`. 

I created https://github.com/mapbox/mapbox-gl-native/issues/7924 to track documentation work for DDS on this platform and that can be done on master or the eventual release branch.

nitrag:
Really exciting! Is color conversion supported yet? ([spec](https://www.mapbox.com/mapbox-gl-js/style-spec/#types-color)) My colors are in rgb hex and are not displaying, not sure if I should open a new issue or if I'm missing something.

```
style.lineColor = MGLStyleFunction<UIColor>(
    interpolationMode: .identity,
    sourceStops: nil,
    attributeName: "color",
    options: [.defaultValue: MGLStyleValue<UIColor>(rawValue: .gray)]
)
```

```
...
properties: {
  ...
  color: '#D50000',
  ...
}
```

It's worth noting the lines are completely invisible, not defaulting to gray.

1ec5:
RGB color interpolation should be supported. Can you open a new ticket so we can track this issue specifically? (Alternative color spaces for interpolation are tracked in #6442, by the way.)

boundsj:
@nitrag @1ec5 here is an example of a feature with rgb hex color in attributes that gets converted correctly with the current 3.5.0 beta 1

```
let feature = MGLPointFeature()
feature.coordinate = centerCoordinate
feature.attributes = ["hexColor": "#00ff00"] // green

let shapeSource = MGLShapeSource(identifier: "shape-source", features: [feature], options: nil)
style.addSource(shapeSource)

let circleStyleLayer = MGLCircleStyleLayer(identifier: "circle-layer", source: shapeSource)

// Use the value in the attributes dictionary with the key `hexColor`, otherwise use red
circleStyleLayer.circleColor = MGLStyleValue(
    interpolationMode: .identity,
    sourceStops: nil,
    attributeName: "hexColor",
    options: [.defaultValue: MGLStyleValue<UIColor>(rawValue: .red)]
)

style.addLayer(circleStyleLayer)
```

1ec5:
These examples inspired me to whip up #8025. 😄

nitrag:
Nevermind, it works when you remember to add the style layer to the mapView. Much noob!

Issue #7593, opened 2017-01-04
Remove Mapbox API examples from testapp


cammace:
We should remove the geocoding and directions examples from the test app since these are tested in the mapbox java test app.

cc: @zugaldia @tobrun

zugaldia:
Related, we'll need to remove some of the changes included by https://github.com/mapbox/mapbox-gl-native/pull/7642.

PR #7598, opened 2017-01-04
Testapp refactor

cammace:
Closes https://github.com/mapbox/mapbox-gl-native/issues/7593

- [x] Changes app colors
- [x] Removes the following examples:
  - [x] Directions
  - [x] Geocoding
  - [ ] Press for marker activity
  - [x] Marker View Scale activity
  - [x] Location picker activity
  - [x] Car driving activity?

- [x] Adds the toolbar by default to the style so we no longer need to include:
```java
Toolbar toolbar = (Toolbar) findViewById(R.id.toolbar);
    setSupportActionBar(toolbar);

    final ActionBar actionBar = getSupportActionBar();
    if (actionBar != null) {
      actionBar.setDisplayShowTitleEnabled(false);
      actionBar.setDisplayHomeAsUpEnabled(true);
      actionBar.setDisplayShowHomeEnabled(true);
    }

...

@Override
  public boolean onOptionsItemSelected(MenuItem item) {
    boolean state;
    switch (item.getItemId()) {
      case android.R.id.home:
        onBackPressed();
        return true;
    }
  }
```
- [ ] Animate marker title change and crash fix
- [x] When going back to Feature activity restore correct position
- [x] Make categories bolder
- [x] Fix up the navigation draw activity
- [x] Polyline activity -> delete all lines -> change property -> crash
- [x] resolve location memory leaks

cc: @mapbox/android


cammace:
I've made the categories more obvious so it's easier to find what you are looking for.

tobrun:
@cammace what needs to happen to merge this? would love to get this in.

kkaefer:
Can we get a super simple map activity at the top of the screen that just displays a map and remembers the location?

tobrun:
@kkaefer yes, @zugaldia has tackled that in #7817

PR #7603, opened 2017-01-05
Rename textField, textFont, textSize, circlePitchScale, *Translate, *TranslateAnchor

### LINKED ISSUE #6098: Rename some layer and source properties

1ec5:
This PR renames the following properties:

* `MGLSymbolStyleLayer.textField` to `text`, in anticipation of mapbox/mapbox-gl-style-spec#646
    The developer would expect a `textField` property to hold an NSTextField or UITextField, or at least some kind of editable control, but not an NSString. There is such a thing as NSText on macOS, but no one ever uses it directly. On the other hand, `UILabel.text`, analogous to the `text-field` layout property, is one of the most commonly used properties on iOS.
* `MGLSymbolStyleLayer.textFont` to `textFontNames`
    The developer would expect to set this property to an NSFont or UIFont, or perhaps an NSFontDescriptor or UIFontDescriptor, but definitely not an array of strings. `textFontNames` is also consistent with `NSFont.fontName`, `UIFont.fontName`, and `MGLSymbolStyleLayer.iconImageName`.
* `MGLSymbolStyleLayer.textSize` to `textFontSize` in anticipation of mapbox/mapbox-gl-style-spec#650
    Contrary to the property’s current name, it does not specify the dimensions of the overall label the way `text-max-width` does. Instead, it specifies the font size in pixels. This inconsistency means a developer would be tempted to set the property to a `CGSize` or `NSSize` value instead of an `NSNumber`.
* `MGLCircleStyleLayer.circlePitchScale` to `circleScaleAlignment`, in anticipation of mapbox/mapbox-gl-style-spec#645
    The developer would expect to set this property to a `CGFloat`, especially considering that we renamed `MGLSymbolStyleLayer.iconScale` (renamed from `iconSize`) is numeric.
* `MGLCircleStyleLayer.circleTranslate` to `circleTranslation`, `circleTranslateAnchor` to `circleTranslationAnchor`, `MGLCircleTranslate` to `MGLCircleTranslation`, and the corresponding fill, line, and symbol properties and enumerations, to complete the work done in #7457

Of these changes, `MGLSymbolStyleLayer.text` is perhaps the most disruptive, but I also see the upstream property inevitably getting renamed to something or other.

Fixes #6098 once and for all. Past this point, if we ever rename any properties unilaterally in the iOS and macOS SDKs, we should not only declare the old name but make it available (albeit deprecated), to minimize disruption. But I’m pretty confident that this will be the last batch of renames for existing properties.

/cc @boundsj @friedbunny @ericrwolfe


1ec5:
On @jfirebaugh’s suggestion in https://github.com/mapbox/mapbox-gl-native/issues/6098#issuecomment-270702105, I also replaced `translate` with `translation` in several property names in anticipation of mapbox/mapbox-gl-style-spec#201.

Finally – and I mean finally – `textSize` is now `textFontSize`. This change is also tracked upstream in mapbox/mapbox-gl-style-spec#650, but I don’t think we should wait for a new version of the style specification to make it.

platform/darwin/src/MGLCircleStyleLayer.h
```diff
@@ -153,20 +153,32 @@ typedef NS_ENUM(NSUInteger, MGLCircleTranslateAnchor) {
  `NSValue` object containing a `CGVector` struct set to 0 points from the left
  and 0 points from the top. Set this property to `nil` to reset it to the
  default value.
+ 
+ This attribute corresponds to the <a
+ href="https://www.mapbox.com/mapbox-gl-style-spec/#layout-circle-circle-translate"><code>circle-translate</code></a>
```
frederoni: This URL leads to the root of the style spec. Isn't the correct ref `#paint-circle-translate`?

1ec5: Good catch – all the paint property links were broken. Fixed in d3a2ffe747550216b19d678c40980231baeeef11.

PR #7605, opened 2017-01-05
[android] - cleanup telemetry

tobrun:
This PR cherry picks some telemetry improvements from [WIP-LocationSource](https://github.com/mapbox/mapbox-gl-native/tree/4331-location-source) branch.

- exponential backoff counter when waiting for permissions to be accepted
- test for exponential backoff
- disabled logging (normally I would remove these logs but due to lack of tests keeping them around).
- usage of a local broadcast receiver 

Review @zugaldia


zugaldia:
@tobrun If my suggestion above could delay making progress with this PR, happy to address the proposed refactor in a follow up ticket.

tobrun:
Happy to expose ExponentialBackoff as a common feature in the SDK but not sure where we could reuse it atm? Wouldn't it make sense making this class part of a commons package (with the new mas structure?).

zugaldia:
@tobrun Thanks, let's leave things as they are and I'll tackle it as part of the larger telem refactor.

PR #7604, opened 2017-01-05
Add -fvisibility=hidden to iOS

kkaefer:
Same as https://github.com/mapbox/mapbox-gl-native/pull/7509, but this time for iOS.


1ec5:
The build is failing where MGLStyleTests refers to some constants defined in mbgl-core:

```
▸ Linking test

❌  Undefined symbols for architecture x86_64
> Symbol: mbgl::util::default_styles::satelliteStreets
> Referenced from: -[MGLStyleTests testVersionedStyleURLs] in MGLStyleTests.o



❌  ld: symbol(s) not found for architecture x86_64



❌  clang: error: linker command failed with exit code 1 (use -v to see invocation)
```

GIven https://github.com/mapbox/mapbox-gl-native/pull/6301#issuecomment-252031292, we might need to rework these tests by hardcoding the version numbers in platform-specific code instead of centralizing them in mbgl.

kkaefer:
Alternatively, we could move all of those information to the header so that it doesn't need to be exported as symbols in the core library itself.

kkaefer:
Pushed an update to this

friedbunny:
Thanks @kkaefer, I’ll give this a try with the different iOS build flavors.

friedbunny:
@kkaefer I’ve pushed up a rebase. I also fixed an issue with the static build not having header search paths completely set (819824f).

It looks like setting `-fvisibility=hidden` gets us a 12% reduction in binary size for dynamic builds, bringing release, symbol-stripped, no-bitcode slices to 3.45MB (`armv7`) and 3.9MB (`arm64`); ~500KB smaller.

The static builds see no changes in binary size, so it appears that `GCC_GENERATE_DEBUGGING_SYMBOLS` portion of #9022 will need more investigation.

After testing the dynamic and static builds in iOS apps, I’m reasonably confident this PR is 👍.

kkaefer:
Why do we need the polylabel headers specifically?

friedbunny:
@kkaefer The direct polylabel dependency was added by @fabian-guerra in https://github.com/mapbox/mapbox-gl-native/pull/8713 for getting the pole of inaccessibility of a polygon annotation. If there’s a way to streamline/remove the polylabel header inclusion, that’d also be great.

friedbunny:
Proposed changing how polylabel is included in the iOS/macOS projects in https://github.com/mapbox/mapbox-gl-native/pull/9469.

friedbunny:
Rebased again — was able to get rid of 819824fb4148a969afb2905ab56dba24705f8e6b after #9469. I think this should be ready to go.

PR #7610, opened 2017-01-05
Add addImage, removeImage API

bsudekum:
This exposes the `addImage` and `removeImage` functions to platform: node. Example usage:

```js
map.addImage('myImageName', {
  imageBuffer: <Premultiplied image buffer>,
  height: 40,
  width: 40
});
```

The developer then would add a corresponding layer, `myImageName` to style the image.

@jfirebaugh I'm not seeing tests for other similar APIs like `addLayer`. How should this be tested?

### DIFF
--- a/platform/node/src/node_map.cpp
+++ b/platform/node/src/node_map.cpp
@@ -9,6 +9,7 @@
 #include <mbgl/style/conversion/source.hpp>
 #include <mbgl/style/conversion/layer.hpp>
 #include <mbgl/style/conversion/filter.hpp>
+#include <mbgl/sprite/sprite_image.cpp>
 
 #include <unistd.h>
 
@@ -58,6 +59,8 @@ void NodeMap::Init(v8::Local<v8::Object> target) {
     Nan::SetPrototypeMethod(tpl, "addSource", AddSource);
     Nan::SetPrototypeMethod(tpl, "addLayer", AddLayer);
     Nan::SetPrototypeMethod(tpl, "removeLayer", RemoveLayer);
+    Nan::SetPrototypeMethod(tpl, "addImage", AddImage);
+    Nan::SetPrototypeMethod(tpl, "removeImage", RemoveLayer);
     Nan::SetPrototypeMethod(tpl, "setLayoutProperty", SetLayoutProperty);
     Nan::SetPrototypeMethod(tpl, "setPaintProperty", SetPaintProperty);
     Nan::SetPrototypeMethod(tpl, "setFilter", SetFilter);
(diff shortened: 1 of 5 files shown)


jfirebaugh:
Can you make the following API adjustments:

* Add a `pixelRatio` property to the options, and use it instead of hard-coding `1`
* Pass the image data as a standalone second argument, before options

So, you'd call it like:

```
map.addImage("id", data, {width: w, height: h, pixelRatio: r});
```

platform/node/src/node_map.cpp
```diff
@@ -579,6 +582,68 @@ void NodeMap::RemoveLayer(const Nan::FunctionCallbackInfo<v8::Value>& info) {
...
+    
+    if (!Nan::Get(imageObject, Nan::New("width").ToLocalChecked()).ToLocalChecked()->IsNumber()) {
+        return Nan::ThrowTypeError("width parameter required");
+    }
+    
+    if (!Nan::Get(imageObject, Nan::New("imageBuffer").ToLocalChecked()).ToLocalChecked()->IsObject()) {
+        return Nan::ThrowTypeError("imageBuffer parameter required");
+    }
+    
+    double imageHeight = Nan::Get(imageObject, Nan::New("height").ToLocalChecked()).ToLocalChecked()->NumberValue();
```
jfirebaugh: `imageHeight` and `imageWidth` should be integers -- validate the type using `IsUint32` rather than `IsNumber` and convert using `Uint32Value` rather than `NumberValue`, assigning to a `uint32_t` rather than a `double`.

platform/node/src/node_map.cpp
```diff
@@ -579,6 +582,68 @@ void NodeMap::RemoveLayer(const Nan::FunctionCallbackInfo<v8::Value>& info) {
...
+    double imageWidth = Nan::Get(imageObject, Nan::New("width").ToLocalChecked()).ToLocalChecked()->NumberValue();
+    auto imageBuffer = Nan::Get(imageObject, Nan::New("imageBuffer").ToLocalChecked()).ToLocalChecked()->ToObject();
+
```
jfirebaugh: Add a check here that the expected size is correct:

```
if (node::Buffer::Length(imageBuffer) != imageHeight * imageWidth * 4) {
    return Nan::ThrowTypeError("image size does not match buffer size");
}
```

tmpsantos:
> How should this be tested?

One option is to write an unit test like `t.test('sets zoom before center', function(t) { ... }` that uses `pixelmatch` for checking the rendering results.

`platform/node/test/js/map.test.js:377`

jfirebaugh:
It would be better to write a test-suite render test that can be used to test the implementation of [the equivalent GL JS feature](https://github.com/mapbox/mapbox-gl-js/issues/2059).

Similar tests of the runtime styling API are in [this directory](https://github.com/mapbox/mapbox-gl-test-suite/tree/master/render-tests/runtime-styling). Take a look at the `"operations"` key in the style.jsons in that tree to get an idea of how this works.

Since we can't embed literal image data into the style.json, the format I'd propose for the `addImage` operation is to pass the name of a file in the [image directory](https://github.com/mapbox/mapbox-gl-test-suite/tree/master/image), and leave it up to [suite_implementation.js](https://github.com/mapbox/mapbox-gl-native/blob/master/platform/node/test/suite_implementation.js#L58-L83) to load the image data at runtime.

bsudekum:
@jfirebaugh started to go down the route of adding a custom `addImage` handler to [suite_implementation.js](https://github.com/mapbox/mapbox-gl-native/pull/7610/commits/52df9fcf32a20fd9ab9cf6a3eea111044f63173f). However, I'm not getting an added image [to the map](https://github.com/mapbox/mapbox-gl-test-suite/compare/add-image). Notice any issues?

jfirebaugh:
`addImage` doesn't work retroactively on layers that tried to use a previously non-existing image ID. Try waiting to add the `symbol` layer (using an `addLayer` operation) until after the image is added.

Also, let's require specifying an ID for the `addImage` operation, instead of reusing the path. (Using the intrinsic dimensions is fine though.)

bsudekum:
@jfirebaugh this is ready for review

platform/node/test/suite_implementation.js
```diff
@@ -65,6 +68,26 @@ module.exports = function (style, options, callback) {
...
+            console.log(path.join(__dirname, '../../../node_modules', 'mapbox-gl-test-suite', operation[2]));
+            var img = PNG.sync.read(fs.readFileSync(path.join(__dirname, '../../../node_modules', 'mapbox-gl-test-suite', operation[2])));
+
+            map.addImage(operation[1], img.data, {
+                height: img.height,
+                width: img.width,
+                pixelRatio: 1
+            });
+
+            map.addLayer({
```
jfirebaugh: This should be test-specific; that is the [add-image test](https://github.com/mapbox/mapbox-gl-test-suite/compare/add-image) should use:

```
      "operations": [
        [
          "addImage",
          "marker",
          "./image/marker.png"
        ],
        [
          "addLayer",
          ...
        ],
        [
          "wait"
        ]
      ]
```

platform/node/src/node_map.cpp
```diff
@@ -579,6 +582,82 @@ void NodeMap::RemoveLayer(const Nan::FunctionCallbackInfo<v8::Value>& info) {
...
+    uint32_t pixelRatio = Nan::Get(optionObject, Nan::New("pixelRatio").ToLocalChecked()).ToLocalChecked()->Uint32Value();
+    auto imageBuffer = Nan::To<v8::Object>(info[1]).ToLocalChecked()->ToObject();
+    
+    if (node::Buffer::Length(imageBuffer) != imageHeight * imageWidth * 4) {
+        return Nan::ThrowTypeError("Image size does not match buffer size");
+    }
+    
+    std::unique_ptr<uint8_t[]> data = std::make_unique<uint8_t[]>(node::Buffer::Length(imageBuffer));
+    std::copy(node::Buffer::Data(imageBuffer), node::Buffer::Data(imageBuffer) + node::Buffer::Length(imageBuffer), data.get());
+    mbgl::PremultipliedImage cPremultipliedImage({ static_cast<uint32_t>(imageWidth), static_cast<uint32_t>(imageHeight) }, std::move(data));
```
jfirebaugh: Don't need `static_cast`s here.

platform/node/src/node_map.cpp
```diff
@@ -579,6 +582,82 @@ void NodeMap::RemoveLayer(const Nan::FunctionCallbackInfo<v8::Value>& info) {
...
+    }
+    
+    if (!Nan::Get(optionObject, Nan::New("pixelRatio").ToLocalChecked()).ToLocalChecked()->IsUint32()) {
+        return Nan::ThrowTypeError("pixelRatio parameter required");
+    }
+    
+    uint32_t imageHeight = Nan::Get(optionObject, Nan::New("height").ToLocalChecked()).ToLocalChecked()->Uint32Value();
+    uint32_t imageWidth = Nan::Get(optionObject, Nan::New("width").ToLocalChecked()).ToLocalChecked()->Uint32Value();
+    
+    if (imageWidth > 1024 && imageHeight > 1024) {
```
jfirebaugh: Should be `||` rather than `&&`.

platform/node/test/suite_implementation.js
```diff
@@ -65,6 +68,19 @@ module.exports = function (style, options, callback) {
...
+            var img = PNG.sync.read(fs.readFileSync(path.join(__dirname, '../../../node_modules', 'mapbox-gl-test-suite', operation[2])));
+
+            map.addImage(operation[1], img.data, {
+                height: img.height,
+                width: img.width,
+                pixelRatio: 1
+            });
+
+            applyOperations(operations.slice(1), callback);
+        } else if (operation[0] === 'addLayer') {
```
jfirebaugh: `addLayer` is already handled by the subsequent `else` block.

jfirebaugh:
Can you open a PR for the mapbox-gl-test-suite additions too?

bsudekum:
@jfirebaugh done https://github.com/mapbox/mapbox-gl-test-suite/pull/201

platform/node/src/node_map.cpp
```diff
@@ -579,6 +582,82 @@ void NodeMap::RemoveLayer(const Nan::FunctionCallbackInfo<v8::Value>& info) {
...
+    }
+    
+    if (!Nan::Get(optionObject, Nan::New("pixelRatio").ToLocalChecked()).ToLocalChecked()->IsUint32()) {
+        return Nan::ThrowTypeError("pixelRatio parameter required");
+    }
+    
+    uint32_t imageHeight = Nan::Get(optionObject, Nan::New("height").ToLocalChecked()).ToLocalChecked()->Uint32Value();
+    uint32_t imageWidth = Nan::Get(optionObject, Nan::New("width").ToLocalChecked()).ToLocalChecked()->Uint32Value();
+    
+    if (imageWidth > 1024 || imageHeight > 1024) {
```
1ec5: FYI, it’s possible for the image-adding operation to have no effect (and log to stderr) if the size of all the images combined exceeds this 1024-by-1024 square.

PR #7613, opened 2017-01-06
Safe guard destroying NativeMapview from async invocation

### LINKED ISSUE #7449: assertion "nativeMapViewPtr != 0" failed
With every jni function execution we assert if the the MapView pointer is still valid. 
If not, the assertion will fail and result in following crash:

```
assertion "nativeMapViewPtr != 0" failed
A/libc: Fatal signal 6 (SIGABRT), code -6 in tid 11659 (pboxsdk.testapp)
```

With some of the recent refactors, some boilerplate state handling was removed. 
We initially added a lot of:

```
if(destroyed){
  return;
}
```
At that time this was a quick fix to be able to ship, we should now look into handling these situations more gracefully. The crash mostly occurs in situations where we still call methods on MapboxMap while the underlying map object is being destroyed.

tobrun:
Closes #7449  

This PR adds safe guards to all our core invocations to make sure the native code doesn't crash while we destroy the mapobject. This behaviour can be produced if we define a asynchronous callback that get invoked during map destruction (eg. Animation callbacks/Handlers/etc.).

I have been experimenting with different approaches but these didn't always cover all the different use-cases. For now I'm settling with these safeguards to be able to ship in near future. Will revisit this once we have made more progress on refactor project that will move ownership/responsibility and thus making this more manageable.  

 @cammace this resolves the crash you have been seeing on AnimatedMarkerActivity.

Review @zugaldia

### DIFF
--- a/platform/android/MapboxGLAndroidSDK/src/main/java/com/mapbox/mapboxsdk/maps/NativeMapView.java
+++ b/platform/android/MapboxGLAndroidSDK/src/main/java/com/mapbox/mapboxsdk/maps/NativeMapView.java
@@ -9,6 +9,7 @@ import android.graphics.RectF;
 import android.os.Build;
 import android.support.annotation.NonNull;
 import android.support.annotation.Nullable;
+import android.text.TextUtils;
 import android.util.DisplayMetrics;
 import android.view.Surface;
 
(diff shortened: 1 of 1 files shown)


tobrun:
@zugaldia thank you for the review! Requested changes have been addressed.

PR #7616, opened 2017-01-06
[core] Fix symbol flickering during tile reload

### LINKED ISSUE #7615: Flickering symbol layers when tiles are refreshed
Tile layout is a two step process: first layout is performed for all non-symbol layers, then symbol layers are placed. Upon completion of the first step, [`GeometryTile` sets its buckets](https://github.com/mapbox/mapbox-gl-native/blob/b6894dd4fc24f93c77b1ad39f466997f1525f2bd/src/mbgl/tile/geometry_tile.cpp#L108) to the result of that step, which will not contain any symbol buckets. Upon completion of the second step, [it merges in symbol buckets](https://github.com/mapbox/mapbox-gl-native/blob/b6894dd4fc24f93c77b1ad39f466997f1525f2bd/src/mbgl/tile/geometry_tile.cpp#L118-L120).

When cached tiles are revalidated and new data is available, or when the tile layout needs to be refreshed due to runtime styling changes, the result of tile layout is intended to seamlessly replace the previously existing layout. However, due to the way that `GeometryTile` is currently implemented, any delay between the two steps can lead to flickering symbol layers, because previously existing symbol buckets are effectively discarded upon completion of the first step, and not replaced until the second step completes.

`GeometryTile` could avoid this effect if it kept symbol and non-symbol buckets segregated, discarding prior symbol buckets only when new symbol buckets became available.

jfirebaugh:
Discard prior symbol buckets only when new symbol buckets became available, in order to eliminate flickering when tiles are refreshed. Plus followup refactor.

Fixes #7615

### DIFF
--- a/src/mbgl/layout/symbol_layout.cpp
+++ b/src/mbgl/layout/symbol_layout.cpp
@@ -27,7 +27,7 @@ namespace mbgl {
 
 using namespace style;
 
-SymbolLayout::SymbolLayout(std::vector<std::unique_ptr<Layer>> layers_,
+SymbolLayout::SymbolLayout(std::vector<std::string> layerIDs_,
                            std::string sourceLayerName_,
                            uint32_t overscaling_,
                            float zoom_,
@@ -37,7 +37,7 @@ SymbolLayout::SymbolLayout(std::vector<std::unique_ptr<Layer>> layers_,
                            style::SymbolLayoutProperties::Evaluated layout_,
                            float textMaxSize_,
                            SpriteAtlas& spriteAtlas_)
-    : layers(std::move(layers_)),
+    : layerIDs(std::move(layerIDs_)),
       sourceLayerName(std::move(sourceLayerName_)),
       overscaling(overscaling_),
       zoom(zoom_),
@@ -254,7 +254,7 @@ void SymbolLayout::addFeature(const SymbolFeature& feature,
                                                   ? SymbolPlacementType::Point
                                                   : layout.get<SymbolPlacement>();
     const float textRepeatDistance = symbolSpacing / 2;
-    IndexedSubfeature indexedFeature = {feature.index, sourceLayerName, layers.at(0)->getID(), symbolInstances.size()};
+    IndexedSubfeature indexedFeature = {feature.index, sourceLayerName, layerIDs.at(0), symbolInstances.size()};
 
     auto addSymbolInstance = [&] (const GeometryCoordinates& line, Anchor& anchor) {
         // https://github.com/mapbox/vector-tile-spec/tree/master/2.1#41-layers
--- a/src/mbgl/tile/geometry_tile.cpp
+++ b/src/mbgl/tile/geometry_tile.cpp
@@ -117,7 +117,7 @@ void GeometryTile::onPlacement(PlacementResult result) {
         availableData = DataAvailability::All;
     }
     symbolBuckets = std::move(result.symbolBuckets);
-    featureIndex->setCollisionTile(std::move(result.collisionTile));
+    collisionTile = std::move(result.collisionTile);
     observer->onTileChanged(*this);
 }
 
@@ -153,7 +153,8 @@ void GeometryTile::queryRenderedFeatures(
                         layerIDs,
                         *data,
                         id.canonical,
-                        style);
+                        style,
+                        collisionTile.get());
 }
 
 } // namespace mbgl
--- a/src/mbgl/tile/geometry_tile.cpp
+++ b/src/mbgl/tile/geometry_tile.cpp
@@ -6,6 +6,7 @@
 #include <mbgl/style/layer_impl.hpp>
 #include <mbgl/style/layers/background_layer.hpp>
 #include <mbgl/style/layers/custom_layer.hpp>
+#include <mbgl/style/layers/symbol_layer.hpp>
 #include <mbgl/style/style.hpp>
 #include <mbgl/storage/file_source.hpp>
 #include <mbgl/geometry/feature_index.hpp>
@@ -105,7 +106,7 @@ void GeometryTile::redoLayout() {
 
 void GeometryTile::onLayout(LayoutResult result) {
     availableData = DataAvailability::Some;
-    buckets = std::move(result.buckets);
+    nonSymbolBuckets = std::move(result.nonSymbolBuckets);
     featureIndex = std::move(result.featureIndex);
     data = std::move(result.tileData);
     observer->onTileChanged(*this);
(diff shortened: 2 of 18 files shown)


src/mbgl/tile/geometry_tile.cpp
```diff
@@ -115,9 +116,7 @@ void GeometryTile::onPlacement(PlacementResult result) {
     if (result.correlationID == correlationID) {
         availableData = DataAvailability::All;
     }
-    for (auto& bucket : result.buckets) {
-        buckets[bucket.first] = std::move(bucket.second);
-    }
+    symbolBuckets = std::move(result.symbolBuckets);
```
kkaefer: This also fixes a potential issue that could occur when after a relayout, a layer doesn't have any data anymore. In this case, it wouldn't produce a new bucket, so the old bucket would not be overwritten.

kkaefer:
Could you please use branch names that are unambiguous? Using a digit-only branch name makes it harder to check out the branch locally.

kkaefer:
I tested this with a database that has prior cache data in my testing area (Berlin), and ran the following SQL command to simulate expired + non-304 responses from the server: `UPDATE tiles SET expires = 1483675317, etag = "xxx";`. When loading the map, I'm still seeing entire loaded, but expired tiles vanish and gradually reappear. It looks like this patch solves the particular case of relayouting, but doesn't solve the flicker for expired tiles replacing the previous tile.

Here's a frame sequence I recorded with this patch: http://bl.ocks.org/kkaefer/raw/a422cc89a78917aaf718fe022c61bfe8/

kkaefer:
This pull request fixes a part of the issue; the other part of the issue was introduced just a few days ago in https://github.com/mapbox/mapbox-gl-native/pull/7586/commits/99c1a0ecbdadb54df32baddb1b0b88df71533c4b#diff-7c17b75ae445adc15c24e3f167193e4eR217:

When a tile in the cache is expired, we start a refresh request. In many cases, we're getting a 304 response (not modified) back, in which case we don't do anything. However, when the tile changed, we're calling `GeometryTile::setData()`, which triggers `GeometryTileWorker::setData()` on the worker thread. However, this is a reused worker, that has processed the previous (expired) tile already. Back when it processed the first tile, it eventually ran `GeometryTileWorker::redoLayout` and this line:

```cpp
std::vector<std::vector<std::unique_ptr<Layer>>> groups = groupByLayout(std::move(*layers));
```

This means that it *moved* the layer information objects out of itself, and through some hoops [into the `SymbolLayout` object](https://github.com/mapbox/mapbox-gl-native/blob/99c1a0ecbdadb54df32baddb1b0b88df71533c4b/src/mbgl/layout/symbol_layout.hpp#L59). When reparsing the tile with new data, `setLayers` isn't called (since nothing about them changed), but the `layers` vector optional is engaged, the the actual vector is empty, meaning that the layout process will not produce any buckets. This empty bucket list [now overwrites](https://github.com/mapbox/mapbox-gl-native/blob/99c1a0ecbdadb54df32baddb1b0b88df71533c4b/src/mbgl/tile/geometry_tile_worker.cpp#L260-L265) the previous (stale, but parsed) buckets. They reappear because immediately after invoking `GeometryTileWorker::setData()`, we're also calling [`redoLayout()`](https://github.com/mapbox/mapbox-gl-native/blob/59803d3025683a255737b5ae29bd7fc531a975a3/src/mbgl/tile/geometry_tile.cpp#L56), which gets a fresh set of cloned layer object over to the worker, and starts processing again.

I implemented a fix in https://github.com/mapbox/mapbox-gl-native/compare/7615...7615-reset-layers-after-redolayout, which works by disengaging the `optional` that holds the (now empty) layer objects.

jfirebaugh:
Thanks for tracking down #7586 as a cause of flickering. Indeed, `GeometryTileWorker` shouldn't have been consuming the layers like that. The intent is that `GeometryTileWorker::set{Layers,Data,PlacementConfig}` can be invoked any number of times, in any order, and `GeometryTileWorker` will do the right thing. In particular `GeometryTileWorker` shouldn't assume that `GeometryTile` is going to call `redoLayout` directly after invoking `setData`. I want to replace `GeometryTile::redoLayout` with `GeometryTile::setLayers`, so that the public `GeometryTile` interface is oriented around _what data has changed_ rather than _what needs to happen in response to something changing_, which should be entirely the responsibility of the implementation.

src/mbgl/tile/geometry_tile.hpp
```diff
@@ -59,7 +59,7 @@ class GeometryTile : public Tile {
 
     class PlacementResult {
     public:
-        std::unordered_map<std::string, std::shared_ptr<Bucket>> buckets;
+        std::unordered_map<std::string, std::shared_ptr<Bucket>> symbolBuckets;
```
jfirebaugh: We could, however it would make `GeometryTile::getBucket` longer because it couldn't use `layer.is<SymbolLayer>() ? symbolBuckets : nonSymbolBuckets`. Not sure if it's worth it.

src/mbgl/layout/symbol_layout.cpp
```diff
@@ -254,7 +254,7 @@ void SymbolLayout::addFeature(const SymbolFeature& feature,
                                                   ? SymbolPlacementType::Point
                                                   : layout.get<SymbolPlacement>();
     const float textRepeatDistance = symbolSpacing / 2;
-    IndexedSubfeature indexedFeature = {feature.index, sourceLayerName, layers.at(0)->getID(), symbolInstances.size()};
+    IndexedSubfeature indexedFeature = {feature.index, sourceLayerName, layerIDs.at(0), symbolInstances.size()};
```
kkaefer: Is it guaranteed that `layerIDs` always has at least one element?

kkaefer:
> what data has changed rather than what needs to happen in response to something changing

Yeah, agreed. There's a call to `redoLayout();` in `GeometryTile::setData(...)`. Instead of sending over the layers every time we call `setData()`, we should call `redoLayout()` when the object is constructed. With an expired tile, we right now call `redoLayout()` on the worker side 3×: once as part of the initial (expired) `setData`, once when calling `setData` with the fresh data, and then immediately after that again when calling `redoLayout` after the `setData` call.

src/mbgl/layout/symbol_layout.cpp
jfirebaugh: That invariant should hold: the layer IDs are constructed from the layer group, and if there's no group, no `SymbolLayout` will be created. However, just in case it doesn't, I used `at(0)` here rather than `[0]` to get an exception rather than invalid memory access.

PR #7621, opened 2017-01-07
Compare absolute value of polygon ring area

springmeyer:
This PR moves native to match JS which compares unsigned areas when limiting polygon rings (https://github.com/mapbox/mapbox-gl-js/blob/fddba1b48e7e65b821ef7c7c3813f03baa8abf9d/js/util/classify_rings.js#L19)

This is important. Currently we compare signed area which means we incorrectly sort polygons before limiting.

Fixes various rendering issues like the coast of sweden:

Before:

screen shot 2017-01-06 at 6 54 25 pm

After:

screen shot 2017-01-06 at 6 54 54 pm

### DIFF
--- a/src/mbgl/tile/geometry_tile_data.cpp
+++ b/src/mbgl/tile/geometry_tile_data.cpp
@@ -127,7 +127,7 @@ void limitHoles(GeometryCollection& polygon, uint32_t maxHoles) {
                          polygon.begin() + 1 + maxHoles,
                          polygon.end(),
                          [] (const auto& a, const auto& b) {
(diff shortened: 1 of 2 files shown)


springmeyer:
Yes, will plan to work on a testcase. Juggling a few other things right now, so anyone feel free to jump in here to help as well with help on testing.

springmeyer:
Testcase now in place. I think this is ready to merge /cc @jfirebaugh.

1ec5:
Adding this change to the iOS and macOS changelogs in #8011.

Issue #7558, opened 2016-12-29
MGLAnnnotation attach to screen when move map


iBinh:
**Platform:** iOS
**Mapbox SDK version:** ios-v3.4.0-beta.6

### Steps to trigger behavior

 1. Add annotation to MGLMapView
 2. Move the map
 

### Actual behavior
Annotation attached to screen
Here's video
https://www.youtube.com/watch?v=6oiXv-xk03A&feature=youtu.be
@1ec5

andrewstay:
Reproducible in **ios-v3.4.0-beta.5** too.

1ec5:
If this regression was introduced in beta 5, that would implicate #7423, which was intended to fix a similar issue.

/cc @boundsj

boundsj:
I can indeed reproduce a similar issue because of the changes in #7423. I've proposed a fix for that in https://github.com/mapbox/mapbox-gl-native/pull/7626. The issue is that our attempt in #7423 to clear up a magic / undocumented value with `presentationLayer` introduced a new bug that manifested as shown in the video above when an annotation view was added outside of the viewport.

boundsj:
This should be fixed by #7626

Gerz-inc:
@boundsj this merge did not fix the problem. But in my case, it start from beta 3. Beta 1 work correct. Beta 2 - I can't test.

boundsj:
@Gerz-inc I don't think I can reproduce this and the repro steps you've provided above are not detailed enough. Can you please share a code sample that reproduces the problem?

l2comdev:
From what I've tested, when the annotation pans off the mapview it appears to be placed to the left outside the mapview frame multiplied by the annotation width. If the width of the annotation is 1px then the annotation sits on the edge of the screen as shown in @Gerz-inc gif, if the annotation width is say 20px then it is placed much further to the left of the map and is not visible.

Gerz-inc:
@l2comdev, yes it's true! When I set the frame size up to 2x2px everything was fine

boundsj:
@Gerz-inc @l2comdev I can reproduce this if I create an annotation view with a frame that is much smaller than some content it holds (like, a 2x2 annotation view frame that holds a UIImage of a car that is 28x28). However, I don't think there is a great way to solve for this edge case in the SDK. Especially for the case where views are added offscreen and the presentation layer is not available (see https://github.com/mapbox/mapbox-gl-native/pull/7626).

[...]

PR #7626, opened 2017-01-08
[ios] Only use non nil presentationLayers to adjust annotation views

boundsj:
Views that are added outside of the viewport do not always have an instantiated presentationLayer. Attempting to use a nil presentationLayer to move a view offscreen resulted in the view being pinned to the left side of the view port (at x == 0).

Fixes https://github.com/mapbox/mapbox-gl-native/issues/7558

### DIFF
--- a/platform/ios/src/MGLMapView.mm
+++ b/platform/ios/src/MGLMapView.mm
@@ -4797,7 +4797,12 @@ public:
             else
             {
                 CGRect adjustedFrame = annotationView.frame;
-                adjustedFrame.origin.x = CGRectGetWidth(annotationView.layer.presentationLayer.frame) * -2.0;
+                if (annotationView.layer.presentationLayer) {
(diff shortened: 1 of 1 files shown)


platform/ios/src/MGLMapView.mm
```diff
@@ -4797,7 +4797,12 @@ - (void)updateAnnotationViews
             else
             {
                 CGRect adjustedFrame = annotationView.frame;
-                adjustedFrame.origin.x = CGRectGetWidth(annotationView.layer.presentationLayer.frame) * -2.0;
+                if (annotationView.layer.presentationLayer) {
+                    adjustedFrame.origin.x = -CGRectGetWidth(annotationView.layer.presentationLayer.frame) * 10.0;
+                } else {
+                    // views that are added off screen do not have a presentationLayer
+                    adjustedFrame.origin.x = -CGRectGetWidth(adjustedFrame) * 10.0;
```
1ec5: FYI, an off-screen view may not have a presentation layer, but it still has a layer.

Issue #7622, opened 2017-01-07
Segfault clustering an empty feature collection — KDBush::select<>()


nitrag:
**Platform:** iOS
**Mapbox SDK version:** 3.4.0-beta6

### Steps to trigger behavior

```swift
var campOptions = [MGLShapeSourceOption : Any]()
campOptions[MGLShapeSourceOption.clustered] = true  //no problem with 'false'

//create empty data source
let campsiteData = MGLShapeSource(identifier: "campsite-data", shape: nil, options: campOptions)
//crashes here
```

```assembly
Mapbox`::-[MGLShapeSource setShape:](MGLShape *):
    0x1012e43a4 <+0>:   stp    x24, x23, [sp, #-64]!
    0x1012e43a8 <+4>:   stp    x22, x21, [sp, #16]
    0x1012e43ac <+8>:   stp    x20, x19, [sp, #32]
    0x1012e43b0 <+12>:  stp    x29, x30, [sp, #48]
    0x1012e43b4 <+16>:  add    x29, sp, #48              ; =48 
    0x1012e43b8 <+20>:  sub    sp, sp, #128              ; =128 
    0x1012e43bc <+24>:  mov    x20, x0
    0x1012e43c0 <+28>:  adrp   x22, 580
    0x1012e43c4 <+32>:  ldr    x22, [x22, #2048]
    0x1012e43c8 <+36>:  ldr    x22, [x22]
    0x1012e43cc <+40>:  stur   x22, [x29, #-56]
    0x1012e43d0 <+44>:  mov    x0, x2
    0x1012e43d4 <+48>:  bl     0x1014a5770               ; symbol stub for: objc_retain
    0x1012e43d8 <+52>:  mov    x19, x0
    0x1012e43dc <+56>:  adrp   x8, 611
    0x1012e43e0 <+60>:  ldr    x1, [x8, #664]
    0x1012e43e4 <+64>:  mov    x0, x20
    0x1012e43e8 <+68>:  bl     0x1014a5740               ; symbol stub for: objc_msgSend
    0x1012e43ec <+72>:  mov    x21, x0
    0x1012e43f0 <+76>:  cbz    x19, 0x1012e4414          ; <+112> at MGLShapeSource.mm:102
    0x1012e43f4 <+80>:  adrp   x8, 611
    0x1012e43f8 <+84>:  ldr    x1, [x8, #184]
    0x1012e43fc <+88>:  mov    x8, sp
[...]

PR #7627, opened 2017-01-08
Add MGLShape from GeoJSON example

1ec5:
Added an example of deserializing an MGLShape from a GeoJSON file.

/ref https://github.com/mapbox/mapbox-gl-native/issues/7622#issuecomment-271118434
/cc @nitrag

### DIFF
--- a/platform/darwin/scripts/update-examples.js
+++ b/platform/darwin/scripts/update-examples.js
@@ -100,7 +100,7 @@ function completeExamples(os) {
         }
         
         // Get the contents of the test method whose name matches the symbol path.
-        let testMethodName = symbolPath.join('$').replace(/$[+-]/, '').replace(/:/g, '_');
+        let testMethodName = symbolPath.join('$').replace(/\$[+-]/, '$').replace(/:/g, '_');
         let example = examples[testMethodName];
         if (!example) {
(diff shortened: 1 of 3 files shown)


1ec5:
> Is there an advantage of using MGLShape's convenience initializer over `-[MGLShapeSource initWithIdentifier:URL:options:]` or vise versa (other than perhaps that one can be used for MGLOverlays while the other is runtime-styling specific)?

`MGLShapeSource.URL` is primarily useful for create a shape source with the contents of a GeoJSON file stored remotely. To be sure, you can specify a local file URL, but if you populate a shape source with a URL instead of a shape, the shape source is a black box: the only way you can access it is via feature querying, which returns tiled data.

`+[MGLShape shapeWithData:encoding:error:]` is more versatile. You can use it to deserialize GeoJSON data, avoiding the mess of typecasts that results from NSJSONSerialization, even without using MGLMapView at all. Additionally, when you create an MGLShapeSource using a shape, such as the result of `+[MGLShape shapeWithData:encoding:error:]`, you retain the ability to access the shape. Access to the shape can be important if, for example, you intend to read a LineString from GeoJSON at launch but continually to append to the polyline at runtime.

PR #7631, opened 2017-01-08
[ios] Scale bar

### LINKED ISSUE #1278: Scale control

### LINKED ISSUE #3331: MGLDistanceFormatter

frederoni:
~~~Fixes #3331~~~ (already implemented in https://github.com/mapbox/mapbox-gl-native/commit/2ab3af4b1b5dabf7087e6fabfb84617f4ff4b55d)
Fixes #1278 

This PR adds a ~~~distance formatter and~~~ scale bar.

![cropped](https://cloud.githubusercontent.com/assets/764476/24864982/2fd3739a-1e06-11e7-80b1-31193da263e1.gif)


Todo:
- [x] Expose a `scaleBar` property on the map view so user can hide the scale bar.
- [x] Test on all devices.
- [x] Mirror layout in case of right-to-left.
- [x] Rename to `MGLScaleBar`


platform/ios/src/MGLScaleBarView.m
```diff
@@ -0,0 +1,289 @@
...
+    [self updateLabels];
+    [self layoutBars];
+    [self layoutLabels];
+}
+
+- (void)layoutBars {
+    
+    CGFloat barWidth = floorf(self.bounds.size.width / self.bars.count);
+    
+    [self.bars enumerateObjectsUsingBlock:^(UIView * _Nonnull bar, NSUInteger idx, BOOL * _Nonnull stop) {
```
1ec5: Since you aren’t making use of functional programming, an ordinary for-in loop should be fine here.

platform/ios/src/MGLScaleBarView.m
```diff
@@ -0,0 +1,289 @@
...
+@property (nonatomic) UIView *containerView;
+@property (nonatomic) MGLDistanceFormatter *formatter;
+@property (nonatomic) NSArray<NSArray<NSNumber*>*> *metricTable;
+@property (nonatomic) NSArray<NSNumber*> *row;
+@property (nonatomic) UIColor *primaryColor;
+@property (nonatomic) UIColor *secondaryColor;
+@property (nonatomic) CALayer *borderLayer;
+@end
+
+static const CGFloat BAR_HEIGHT = 4;
```
1ec5: Use SCREAMING_SNAKE_CASE for macros and MGLUpperCamelCase for constants.

platform/ios/src/MGLScaleBarView.m
```diff
@@ -0,0 +1,289 @@
...
+    self.row = [self preferredRowInRows:[self validRows]];
+    
+    [self invalidateIntrinsicContentSize];
+    [self setNeedsLayout];
+}
+
+- (void)setRow:(NSArray<NSNumber *> *)row {
+    if (![row isEqualToArray:_row]) {
+        [_bars makeObjectsPerformSelector:@selector(removeFromSuperview)];
+        [_labels makeObjectsPerformSelector:@selector(removeFromSuperview)];
```
1ec5: Would setting `self.subviews` to the empty array also work here, or would that mess up constraints?

platform/ios/src/MGLScaleBarView.m
```diff
@@ -0,0 +1,289 @@
...
+    }
+    return _bars;
+}
+
+- (NSArray<UILabel *> *)labels {
+    if (!_labels) {
+        NSMutableArray *labels = [NSMutableArray array];
+        for (NSUInteger i = 0; i <= self.row.lastObject.integerValue; i++) {
+            UILabel *label = [[UILabel alloc] init];
+            label.text = @"0";
```
1ec5: Is this a placeholder string, or does it need to be localized?

platform/ios/src/MGLScaleBarView.m
```diff
@@ -0,0 +1,289 @@
...
+    return self;
+}
+
+- (void)commonInit {
+    self.clipsToBounds = NO;
+    self.alpha = 0;
+    
+    _visible = NO;
+    _primaryColor = [UIColor colorWithRed:62.0/255.0 green:62.0/255.0 blue:62.0/255.0 alpha:1];
+    _secondaryColor = [UIColor colorWithRed:247.0/255.0 green:247.0/255.0 blue:247.0/255.0 alpha:1];
```
friedbunny: The text and border colors should switch between white-on-black and black-on-white, depending on the background color of the current map style — e.g., the satellite and dark styles would need colors that are inverted from the current black-on-white.

platform/ios/src/MGLScaleBarView.h
```diff
@@ -0,0 +1,5 @@
+#import <UIKit/UIKit.h>
+
+@interface MGLScaleBarView : UIView
+
```
friedbunny: Should we expose any sort of styling properties publicly? Colors, typeface, position?

platform/ios/src/MGLScaleBarView.m
```diff
@@ -0,0 +1,289 @@
...
+        return;
+    }
+    
+    [self updateLabels];
+    [self layoutBars];
+    [self layoutLabels];
+}
+
+- (void)layoutBars {
+
```
friedbunny: Nit: there’s an inconsistent line break here.

platform/ios/src/MGLMapView.mm
```diff
@@ -230,6 +231,8 @@ @interface MGLMapView () <UIGestureRecognizerDelegate,
 @property (nonatomic) EAGLContext *context;
 @property (nonatomic) GLKView *glView;
 @property (nonatomic) UIImageView *glSnapshotView;
+@property (nonatomic, readwrite) MGLScaleBarView *scaleBarView;
```
friedbunny: We expose the other views that sit on top of the map (compass, logo, attribution) as public MGLMapView properties — I think it would be consistent and useful to do the same for this scale view. At minimum, it would allow a developer to hide the scale bar.

platform/ios/src/MGLScaleBarView_Private.h
```diff
@@ -0,0 +1,12 @@
+#import <UIKit/UIKit.h>
+#import <CoreLocation/CoreLocation.h>
+
+#import "MGLScaleBarView.h"
+
+@interface MGLScaleBarView (Private)
```
frederoni: I like @friedbunny's suggestion of exposing some style properties. Perhaps alternating colors, border color etc and then we'll need this private header.

platform/ios/src/MGLMapView.mm
```diff
@@ -1223,6 +1258,12 @@ - (void)handlePanGesture:(UIPanGestureRecognizer *)pan
         {
             CGPoint offset = CGPointMake(velocity.x * self.decelerationRate / 4, velocity.y * self.decelerationRate / 4);
             _mbglMap->moveBy({ offset.x, offset.y }, MGLDurationInSeconds(self.decelerationRate));
+            __weak MGLMapView *weakSelf = self;
```
frederoni: I tried that first but then it'll fade out while you zoom or pan if you don't move consistently. :/

platform/ios/src/MGLScaleBarView.m
frederoni: Yep, this needs to be localized.

platform/ios/src/MGLScaleBarView.m
```diff
@@ -0,0 +1,289 @@
+#import <Mapbox/Mapbox.h>
+
+#import "MGLScaleBarView_Private.h"
+
+@interface MGLScaleBarView()
+@property (nonatomic, assign, getter=isVisible) BOOL visible;
```
frederoni: `hidden` would hide the view instantly so I use this property to flag the view as visible/invisible when fadeIn/fadeOut begins in order to prevent unwanted calls to those methods.

platform/ios/src/MGLMapView.mm
1ec5: If we expose the view as an ordinary UIView, the developer can still hide the scale bar and we retain more flexibility to iterate on the view’s design and structure.

platform/ios/src/MGLScaleBarView.m
1ec5: This will never be perfect, even if the developer gets full control over the scale bar’s colors. Consider Mapbox Satellite: a scale bar that’s readable over a lush forest may be unreadable over tundra and vice versa.

Apple Maps and Google Maps solve this problem by treating the scale bar’s labels just like any other labels on the base map: in Standard mode, the labels are set in black with a subtle, white stroke for contrast:

ios-standard
_Apple Maps on iOS, Standard_

macos-standard<br>
_Apple Maps on macOS, Standard_

google-map
_Google Maps on iOS, main map_

google-terrain
_Google Maps on iOS, Terrain_

In Satellite mode, the labels are inverted to white with a subtle, black stroke. Another nice touch in Apple Maps is that the scale bar has slight translucency in Satellite mode to avoid becoming distracting.

ios-satellite
_Apple Maps on iOS, Satellite_

macos-satellite<br>
_Apple Maps on macOS, Satellite_

google-satellite
_Google Maps on iOS, Satellite_

Our SDK has to cope with a wider variety of color schemes and challenges, since the style could contain anything. Here are a few possibilities:

* Keep the appearance as-is. The scale bar isn’t the most important ornament on the screen; a little squinting may not be a showstopper.
* Place a blurred background beneath the scale bar, as the macOS SDK does for attribution. Note that this effect would be similar but not identical to a visual effect view, whose effect is a bit too vibrant for a control that shouldn’t be conspicuous.
* Choose between a light and dark appearance based on the style’s background color, as @friedbunny suggests.
* Find the most similar symbol style layer and mimic its appearance. This could yield the best results – the highest fidelity to the style – but could also be somewhat unpredictable.

platform/ios/src/MGLScaleBarView.h
1ec5: I think knobs along these lines could be tail work. That way we can address customizability of the ornaments at the same time in a consistent manner.

platform/ios/src/MGLScaleBarView.m
1ec5: In that case, `CALayer.hidden` is animatable.

platform/ios/src/MGLMapView.mm
1ec5: Good point. In that case, perhaps `-notifyGestureDidEndWithDrift:` is what you want. If not, try factoring out the animation into a similar method.

platform/ios/src/MGLScaleBarView.m
1ec5: > Find the most similar symbol style layer and mimic its appearance.

Or even a way for the developer to specify which symbol style layer they want the scale bar and logo view to match?

platform/ios/src/MGLMapView.mm
```diff
@@ -780,6 +790,31 @@ - (UIViewController *)viewControllerForLayoutGuides
...
+    [self.scaleControlViewConstraints removeAllObjects];
+    
+    [self.scaleControlViewConstraints addObject:
+     [NSLayoutConstraint constraintWithItem:self.scaleBarView
+                                  attribute:NSLayoutAttributeTop
+                                  relatedBy:NSLayoutRelationEqual
+                                     toItem:self
+                                  attribute:NSLayoutAttributeTop
+                                 multiplier:1
+                                   constant:5+self.contentInset.top]];
```
frederoni: I don't see any constraints between the logo and the attribution button.
The maximum width of the scale bar is `screenWidth/2-padding` which might be a bit too much, especially on iPads but I don't think the answer is to constrain it to the compass. It makes positioning the components with `contentInsets` more prone to misplacement.

1ec5:

1ec5:
I could see adding customization options causing this PR to drag on a bit longer, whereas the feature is already quite polished. How about we make MGLScaleBar private and change the public type of the `scaleBar` property to just UIView (so it can be hidden like the other ornaments)? Adding customization options can be a separate PR, once there's more clarity about the options people want.

frederoni:
MGLScaleBarView is now private and supports imperial units. I had to change the roundingIncrement of MGLDistanceFormatter to avoid rounding less than 12.5 feet down to 0 and you can zoom into 2 feet per bar near the poles.

This PR is ready for another round 👀

platform/ios/src/MGLScaleBarView.m
frederoni: subviews is a readonly property

platform/ios/src/MGLMapView.mm
```diff
@@ -774,6 +784,31 @@ - (UIViewController *)viewControllerForLayoutGuides
...
```
1ec5: Should we stick to a more standard constant like 4 points?

platform/ios/src/MGLMapView.mm
```diff
@@ -1566,7 +1604,11 @@ - (void)handleDoubleTapGesture:(UITapGestureRecognizer *)doubleTap
             [self animateWithDelay:MGLAnimationDuration animations:^
              {
                  [weakSelf unrotateIfNeededForGesture];
+                 [weakSelf notifyScaleBarGestureDidEnd];
              }];
+        } else {
+            [self unrotateIfNeededForGesture];
+            [self notifyScaleBarGestureDidEnd];
```
1ec5: In the past, atomic gestures like double-tapping avoided calling `-notifyGestureDidBegin` and `-notifyGestureDidEndWithDrift:`. We should have these gestures call `-notifyGestureDidBegin` and `-notifyGestureDidEndWithDrift:` one right after the other, and merge `-notifyScaleBarGestureDidEnd` into `-notifyGestureDidEndWithDrift:`. That way there are fewer things to keep track of every time we finish a gesture.

platform/ios/src/MGLMapView.mm
```diff
@@ -4759,7 +4804,11 @@ - (void)cameraIsChanging {
     }
 
     [self updateCompass];
-
+    
+    if (!self.scaleBarView.hidden) {
+        [(MGLScaleBarView *)self.scaleBarView setMetersPerPoint:[self metersPerPointAtLatitude:self.centerCoordinate.latitude]];
```
1ec5: What if the scale bar is hidden while the user zooms the map, but then the developer unhides the scale bar afterwards? Would the scale bar remain accurate? Can we make it safe to call `-[MGLScaleBarView setMetersPerPoint:metersPerPointAtLatitude:]` even when MGLScaleBarView is hidden?

platform/ios/src/MGLScaleBarView.h
```diff
@@ -0,0 +1,10 @@
+#import <UIKit/UIKit.h>
+#import <CoreLocation/CoreLocation.h>
+
+@interface MGLScaleBarView : UIView
+
+@property (nonatomic, assign) CLLocationDistance metersPerPoint;
+
+- (void)fadeOut;
```
1ec5: Even though this class is private, it would be good to give it some documentation.

platform/ios/src/MGLMapView.h
```diff
@@ -226,6 +226,15 @@ IB_DESIGNABLE
 - (IBAction)reloadStyle:(id)sender;
 
 /**
+ A control indicating the scale of the map.
+ 
+ The scale bar becomes visible when a user interacts with the
+ map and fades out when the interaction stops. It’s positioned
+ in the upper-left corner.
+ */
+@property (nonatomic, readonly) UIView *scaleBarView;
```
1ec5: “Scale bar view” is redundant. Let’s call it either a “scale bar” (`scaleBar`, `MGLScaleBar`) or a “scale view” (`scaleView`, `MGLScaleView`).

platform/ios/src/MGLMapView.h
```diff
@@ -226,6 +226,15 @@ IB_DESIGNABLE
 - (IBAction)reloadStyle:(id)sender;
 
 /**
+ A control indicating the scale of the map.
+ 
+ The scale bar becomes visible when a user interacts with the
+ map and fades out when the interaction stops. It’s positioned
```
1ec5: Nit: wrap to 80 columns.

1ec5: Nit: avoid contractions, which make the documentation sound too informal.

platform/ios/src/MGLScaleBarView.m
```diff
@@ -0,0 +1,387 @@
...
+    _borderLayer.borderColor = [self.primaryColor CGColor];
+    _borderLayer.borderWidth = 1.0f / [[UIScreen mainScreen] scale];
+    
+    [_containerView.layer addSublayer:_borderLayer];
+    
+    _formatter = [[MGLDistanceFormatter alloc] init];
+    _row = @[@0, @0];
+    
+    // @[meters, numberOfBars]
+    _metricTable = @[@[@1, @2],
```
1ec5: There seems to be quite a bit of Objective-C overhead in managing this tabular data (in terms of syntax, not necessarily performance). These tables should be statically initialized as C arrays of anonymous structs. For example, you could declare an inline type:

```objc
static const struct {
    CLLocationDistance distance;
    NSUInteger numberOfBars;
} MGLNumberOfBarsByDistance[] = {
    …
};
```

platform/ios/src/MGLMapView.h
frederoni: Let's rename to MGLScaleView, it aligns better with Cocoa naming conventions.

platform/ios/src/MGLMapView.mm
frederoni: We've been using 5 on the compassView for quite a while. Let's revisit when we replace these constraints in #7716

platform/ios/src/MGLMapView.mm
frederoni: The scale bar remains faded out until you set `metersPerPoint` making it safe. I'd prefer not to set `metersPerPoint` if the view is hidden to avoid excessive layout code from being executed.

1ec5: Sorry, something isn’t clicking for me: if the developer hides the scale bar using `mapView.scaleView.hidden = YES`, then the user zooms way in, then the developer sets `mapView.scaleView.hidden = NO`, wouldn’t the scale bar appear with the wrong scale until the user starts zooming again?

platform/ios/src/MGLMapView.h
1ec5: For what it’s worth, “bar” is primarily used in UITabBar.

platform/ios/src/MGLScaleBarView.m
```diff
@@ -0,0 +1,387 @@
...
+        bar.frame = CGRectMake(xPosition, self.borderWidth, barWidth, MGLBarHeight);
+        i++;
+    }
+    
+    self.containerView.frame = CGRectMake(CGRectGetMinX(self.bars.firstObject.frame),
+                                          CGRectGetMaxY(self.bounds)-MGLBarHeight,
+                                          self.actualWidth,
+                                          MGLBarHeight+self.borderWidth*2);
+    
+    self.containerView.layer.cornerRadius = 2.0f;
```
friedbunny: Any objections to getting rid of the corner radius here? (The white border around the outside.)

screen shot 2017-04-05 at 12 11 52 am

It gives me a vague sense of unevenness, probably because the corners of the black outline are not rounded.

platform/ios/src/MGLScaleBarView.m
```diff
@@ -0,0 +1,387 @@
...
+@end
+
+@implementation MGLScaleBarLabel
+
+- (void)drawTextInRect:(CGRect)rect {
+    CGSize shadowOffset = self.shadowOffset;
+    UIColor *textColor = self.textColor;
+    
+    CGContextRef context = UIGraphicsGetCurrentContext();
+    CGContextSetLineWidth(context, 1);
```
friedbunny: The stroke width here (around the outside of the label’s numbers/text) feels a bit too thin and diminishes legibility.

stroke-light
dark-stroke
_Top: `1` unit; bottom: `2` units_

It does look like widening the stroke width does cause the sides to get cropped a bit, though... especially the zero character — its stroke is very uneven and sharp when `2`.

friedbunny:
The scale bar, as of 7de5c272b80ae683a58648073f4d67ec35f7f4ed:


_Apologies for the choppy gif — bar changes are smooooth._

friedbunny:
One thing I noticed about the current implementation: the scale bar unexpectectly appears due to non-zoom-related changes in the scale, such as vertical panning at low zooms (which causes the scale to change due to map projection). I think we should only make the scale bar appear on zoom-initiated scale changes.

Apple and Google both only have the scale bar first appear when zooming, though they take slightly different approaches to updates/hiding:
- **Apple:** Scale stops updating when zooming stops. Scale bar hides very quickly (~500ms) when the zoom gesture is finished.
- **Google:** Scale continues updating when zooming stops. Scale bar does not hide until the map has been idle for ~2000ms — this includes waiting until new gestures have finished.

platform/ios/src/MGLScaleBarView.m
```diff
@@ -0,0 +1,387 @@
...
+    return _bars;
+}
+
+- (NSArray<UILabel *> *)labels {
+    if (!_labels) {
+        NSMutableArray *labels = [NSMutableArray array];
+        for (NSUInteger i = 0; i <= self.row.lastObject.integerValue; i++) {
+            UILabel *label = [[MGLScaleBarLabel alloc] init];
+            label.font = [UIFont systemFontOfSize:8 weight:UIFontWeightMedium];
+            NSCharacterSet *characterSet = [[NSCharacterSet decimalDigitCharacterSet] invertedSet];
```
1ec5: Take this out of the `for` loop. Although `+[NSCharacterSet decimalDigitCharacterSet]` is probably static, the inverted copy of it is not.

platform/ios/src/MGLScaleBarView.m
```diff
@@ -0,0 +1,387 @@
...
+}
+
+- (void)updateLabels {
+    NSArray *labels = [self.labels subarrayWithRange:NSMakeRange(1, self.labels.count-1)];
+    BOOL useMetric = [self usesMetricSystem];
+    NSUInteger i = 0;
+    
+    for (MGLScaleBarLabel *label in labels) {
+        CLLocationDistance barDistance = (self.row.firstObject.integerValue / self.row.lastObject.integerValue) * (i + 1);
+        if (!useMetric) { barDistance /= MGLFeetPerMeter; }
```
1ec5: Nit: put the body of the `if` block on a separate line.

platform/ios/src/MGLScaleBarView.m
```diff
@@ -0,0 +1,387 @@
...
+        
+        label.text = [self.formatter stringFromDistance:barDistance];
```
1ec5: As seen in https://github.com/mapbox/mapbox-gl-native/pull/7631#discussion_r95090015, Apple only labels the rightmost distance with the unit, which saves space. (The rightmost label is centered on the number, which suggests that the unit is actually a separate label.)

To get the formatted number, call `-[NSLengthFormatter unitStringFromMeters:usedUnit:]` to get the appropriate unit, then multiply by the unit’s factor, then call the length formatter’s `numberFormatter`’s `-stringFromNumber:`.

platform/ios/src/MGLMapView.mm
1ec5: I wonder if this will also help with an issue I’m seeing in which pinching multiple times in quick succession causes the scale bar to stay visible; only pinching again gets the scale bar unstuck.

platform/ios/src/MGLScaleBarView.m
```diff
@@ -0,0 +1,387 @@
...
+    [self layoutBars];
+    [self layoutLabels];
+}
+
+- (void)layoutBars {
+    CGFloat barWidth = (CGRectGetWidth(self.bounds) - self.borderWidth * 2.0f) / self.bars.count;
+    
+    NSUInteger i = 0;
+    for (UIView *bar in self.bars) {
+        CGFloat xPosition = barWidth * i + self.borderWidth;
```
1ec5: If the application is localized into a right-to-left language like Arabic, the scale bar never appears. We should [check the interface direction](https://developer.apple.com/library/content/documentation/MacOSX/Conceptual/BPInternational/SupportingRight-To-LeftLanguages/SupportingRight-To-LeftLanguages.html#//apple_ref/doc/uid/10000171i-CH17-SW2) and subtract this position from the superview’s width if the interface is laid out right-to-left.

platform/ios/src/MGLScaleBarView.m
frederoni: I tend to agree. I wanted to round the background two pixels and the border one pixel but rounding one pixel is handled differently in CG compared to Sketch.

platform/ios/src/MGLScaleBarView.m
frederoni: Changing the fixed x position of the bar would make alignments look weird as long as the bar is on the left side. Would it make sense to only mirror the labels? Preferably I guess the compass and the bar should switch places in case of right-to-left?

platform/ios/src/MGLScaleBarView.m
frederoni: It seems that I'm adding more overhead when using static C arrays of anonymous structs, either in terms of manual memory management when allocating new C arrays or using a C++ list to return a subset of the rows needed for `-[MGLScaleBarView validRows]` and `-[MGLScaleBarView preferredRowInRows:]`. I'm leaning towards rewriting those methods in objc++.

platform/ios/src/MGLScaleBarView.m
frederoni: Then we need to modify the rows and MGLDistanceFormatter to avoid cases like this;
mixed

platform/ios/src/MGLMapView.mm
frederoni: If the zoom interaction occurred less than one second ago, it would show the incorrect scale for the remaining milliseconds up to 1 sec but since it was hidden, we never call `-fadeIn` leaving the view.alpha at 0 making the incorrect scale invisible. You have to set `metersPerPoint` again to make it visible.

Though, it is possible to trigger the case you think of but it's slightly harder in practice;
Using `scaleView.hidden = NO`, zoom to fade in the scale bar, hide the scale bar, change the zoom, and unhide the scale bar all within 1 second in order for it to show an incorrect scale.

However, we can remove the if-statement if the overhead is exaggerated.

platform/ios/src/MGLScaleBarView.m
```diff
@@ -0,0 +1,387 @@
...
+
+// Find the row closest to -[MGLScaleBarView preferredWidth]
+- (NSArray<NSNumber *> *)preferredRowInRows:(NSArray<NSArray<NSNumber *> *> *)rows {
+    CLLocationDistance diff = CGFLOAT_MAX;
+    CLLocationDistance preferredDistance = [self preferredWidth] * [self unitsPerPoint];
+    NSArray<NSNumber *> *preferredRow = rows.firstObject;
+    for (NSArray<NSNumber *> *row in rows) {
+        CLLocationDistance currentDiff = ABS(preferredDistance - row.firstObject.floatValue);
+        if (currentDiff < diff) {
+            preferredRow = row;
```
frederoni: Update diff is missing here so it always prefers the last row which looks better and doesn't make the labels overlap. This method can be removed then always pick the last row from -validRows.

platform/ios/src/MGLMapView.mm
frederoni: `-[MGLMapView handlePanGesture:]` calls `-notifyGestureDidBegin`.
We have to keep them separated if non-zoom-related gestures like panning shouldn't trigger the scale bar.
Should we extract into `-notifyGestureDidBegin` and `-notifyZoomChangingGestureDidBegin`?
 /cc @friedbunny

platform/ios/src/MGLMapView.h
frederoni: _waiting with renaming of the class to keep the reviewing process intact_

platform/ios/src/MGLScaleBarView.m
1ec5: The compass already appears in the top-left corner due to Auto Layout mirroring. Ideally everything about the scale bar would be mirrored, so that it reads from right to left.

platform/ios/src/MGLScaleBarView.m
1ec5: There are two important points here:

* The tables never change, so they should be allocated statically. A static C structure would be best; if an NSArray is really necessary, it should be created inside a dispatch_once block.
* The table is already sorted by distance, so the valid rows are always contiguous. Therefore, we can represent the valid rows as a range of indices into the static table rather than allocating a new subarray each time we relayout.

This code doesn't benefit from NSArray's mutability or its ability to hold objects.

platform/ios/src/MGLScaleBarView.m
1ec5: True, I guess we’d have a special case for when `-[NSLengthFormatter unitStringFromMeters:usedUnit:]` reports a different unit for the two labels. Avoiding repeated units can be tail work; I only suggested this approach as a space-saving measure, since the map view may not necessarily occupy the entire window.

platform/ios/src/MGLMapView.h
1ec5: I see that you've renamed the class to MGLScaleBar. Just to clarify, my comment above was meant to point out that UITabBar is a "bar" but almost nothing else is. Regardless, I'm comfortable with either name. 👍

platform/ios/src/MGLScaleBar.mm
```diff
@@ -0,0 +1,384 @@
...
+
+static const CGFloat MGLFeetPerMile = 5280;
+
+struct MGLRow {
+    CLLocationDistance distance;
+    NSUInteger numberOfBars;
+};
+
+static const struct {
+    struct MGLRow row;
```
1ec5: This is one extra level of indirection. All we need is an array of MGLRows; this creates an array of anonymous structs containing MGLRows. With a simple array of MGLRows, we can eliminate some of the braces below.

My earlier suggestion was that, as long as you don't need to refer to the specific type of these rows anywhere else, it's possible to inline MGLRow here, without naming it, and use `auto` everywhere else you'd refer to the row's type.

1ec5: Anyhow, since the Imperial table needs access to the same type as this table, it does make sense to name MGLRow, just not to then stuff it into an anonymous struct as this code does.

platform/ios/src/MGLMapView.mm
friedbunny: @frederoni I think we have to keep the scale bar unhiding in `-cameraIsChanging`, otherwise programmatic zooming won’t trigger it. If that’s true, then bifurcating `-notifyGestureDidBegin` wouldn’t do us any good. (Have we considered programmatic zooming?)

If we don’t care about programmatic zoom triggering the scale bar, then perhaps we can keep a single method but have it take an enum, e.g., `-notifyGestureDidBeginWithType:` and `MGLGestureType`, that would define whether or not it was initiated by a zoom-related gesture. (Or just a `-notifyGestureDidBeginWithZoom:` boolean.)

1ec5: > Have we considered programmatic zooming?

MapKit only shows the scale in response to user gestures, not in response to programmatic camera changes. However, there is an `-[MKMapCamera showsScale]` property that determines whether the scale bar is always visible: if the developer wants the scale bar to be visible during programmatic zooming, they can set and unset `showsScale` around that change to the map’s region or camera.

1ec5:
FYI, a couple differences remain between the implementation here and MapKit’s:

* MapKit hides the scale bar as soon as it needs to display a distance longer than 375 miles or 750 kilometers. (This distance is dependent on both the altitude and latitude, so not just the zoom level.)
* When the `showsScale` property is set to `YES`, the scale bar is always visible. When it’s set to the default of `NO`, the scale is never visible. The auto-fading functionality appears to be specific to the Maps application on iOS. I like the auto-fading functionality, but I suppose it would simplify the code somewhat to unconditionally show or hide the scale bar. 🤷‍♂️

frederoni:
Leave show/hide up to the developer simplifies a lot indeed. 👍 
I added a maximum scale of 400 miles or 800 kilometers. I also removed the fade in and out animation.

platform/ios/src/MGLScaleBar.mm
```diff
@@ -0,0 +1,359 @@
...
+    NSUInteger count = metric
+    ? sizeof(MGLMetricTable) / sizeof(MGLMetricTable[0])
+    : sizeof(MGLImperialTable) / sizeof(MGLImperialTable[0]);
+    
+    CLLocationDistance maximumDistance = [self maximumWidth] * [self unitsPerPoint];
+    CLLocationDistance allowedDistance = metric
+    ? MGLMetricTable[count-1].distance
+    : MGLImperialTable[count-1].distance;
+    
+    self.alpha = maximumDistance > allowedDistance ? 0 : 1;
```
1ec5: MapKit fades the scale bar in and out as it appears and disappears due to the maximum distance threshold. We could use a UIView animation block here perhaps.

platform/ios/src/MGLMapView.h
```diff
@@ -226,6 +226,12 @@ IB_DESIGNABLE
 - (IBAction)reloadStyle:(id)sender;
 
 /**
+ A control indicating the scale of the map. The scale bar is positioned in the
+ upper-left corner.
```
1ec5: Note that the view is hidden by default.

platform/ios/src/MGLScaleBar.mm
```diff
@@ -235,7 +235,13 @@ - (void)updateVisibility {
     ? MGLMetricTable[count-1].distance
     : MGLImperialTable[count-1].distance;
     
-    self.alpha = maximumDistance > allowedDistance ? 0 : 1;
+    CGFloat alpha = maximumDistance > allowedDistance ? .0f : 1.0f;
+    
+    if(self.alpha != alpha) {
+        [UIView animateWithDuration:.2f delay:0 options:UIViewAnimationOptionBeginFromCurrentState animations:^{
```
1ec5: FYI, most of MGLMapView’s animations occur within 300 milliseconds, via `MGLAnimationDuration`. (200 milliseconds is fine, though.)

PR #7634, opened 2017-01-08
Document requirement about adding sources and layers

boundsj:
Sources and layers added outside of mapView:didFinishLoadingStyle: and mapViewDidFinishLoadingMap: are not guaranteed to appear since the map may not be fully loaded and ready to accept new sources and layers.

This also fixes a compiler warning in iosapp.

ref https://github.com/mapbox/mapbox-gl-native/issues/7492

### DIFF
--- a/platform/darwin/src/MGLStyle.h
+++ b/platform/darwin/src/MGLStyle.h
@@ -40,10 +40,10 @@ static const NSInteger MGLStyleDefaultVersion = 9;
  via `-[MGLMapView style]` by updating the style's data sources or layers.
  
  @note Wait until the map style has finished loading before modifying a map's
-    style via any of the `MGLStyle` instance methods below.
-    You can use the `MGLMapViewDelegate` methods `-mapViewDidFinishLoadingMap:`
-    or `-mapView:didFinishLoadingStyle:` as indicators that it's safe
-    to modify the map's style.
+    style via any of the `MGLStyle` instance methods below. You can use the 
+    `-[MGLMapViewDelegate mapView:didFinishLoadingStyle:]` or
+    `-[MGLMapViewDelegate mapViewDidFinishLoadingMap:]` methods as indicators
+    that it's safe to modify the map's style.
  */
 @interface MGLStyle : NSObject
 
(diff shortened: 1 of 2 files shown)


platform/darwin/src/MGLStyle.h
```diff
@@ -213,8 +213,11 @@ static const NSInteger MGLStyleDefaultVersion = 9;
  @note Adding the same source instance more than once will result in a
     `MGLRedundantSourceException`. Reusing the same source identifier, even with
     different source instances, will result in a 
-    `MGLRedundantSourceIdentifierException`.
- 
+    `MGLRedundantSourceIdentifierException`. Sources should be added in
```
1ec5: Does jazzy let us split this into two `@note`s?

platform/darwin/src/MGLStyle.h
```diff
@@ -213,8 +213,11 @@ static const NSInteger MGLStyleDefaultVersion = 9;
  @note Adding the same source instance more than once will result in a
     `MGLRedundantSourceException`. Reusing the same source identifier, even with
     different source instances, will result in a 
-    `MGLRedundantSourceIdentifierException`.
- 
+    `MGLRedundantSourceIdentifierException`. Sources should be added in 
+    `MGLMapViewDelegates`'s `mapView:didFinishLoadingStyle:` or
```
1ec5: s/MGLMapViewDelegates/MGLMapViewDelegate/

Also, nowadays we say `-[MGLMapViewDelegate mapView:didFinishLoadingStyle:]` so that jazzy can autolink it: realm/jazzy#362.

platform/darwin/src/MGLStyle.h
```diff
@@ -213,8 +213,11 @@ static const NSInteger MGLStyleDefaultVersion = 9;
...
+    `mapViewDidFinishLoadingMap:` methods to ensure that the map has loaded
+    the style and is ready to accept a new source or layer.
```
1ec5: s/source or layer/source/

platform/ios/app/MBXViewController.m
```diff
@@ -1615,6 +1615,7 @@ - (void)mapView:(MGLMapView *)mapView tapOnCalloutForAnnotation:(id <MGLAnnotati
     }
     
     MGLPointAnnotation *point = annotation;
```
1ec5: This line is redundant.

platform/darwin/src/MGLStyle.h
boundsj: Yes

boundsj:
@1ec5 I think that two `@note`s look good:

screenshot 2017-01-08 16 04 47

I guess the linking will start working after your change to jazzy is released.

I also went ahead and updated the note for the class itself but the mentions of other class's methods look like they are handled differently (i.e. Not bolded. And, probably won't be linked to?):

screenshot 2017-01-08 16 06 05

1ec5:
> I guess the linking will start working after your change to jazzy is released.

Correct. In the meantime, you can clone and install jazzy (or point document.sh to bin/jazzy in your checkout) to see what it looks like.

> Not bolded. And, probably won't be linked to?

It looks like it’s in a `<code>` tag, so it should get a link. Apparently we’re applying the bolding and border somewhat selectively in the stylesheet.

Issue #7483, opened 2016-12-19
MGLMapView moves custom MGLCalloutView, ignoring custom frame


andrewstay:
**Platform:** iOS
**Mapbox SDK version:** ios-v3.4.0-beta.5

### Steps to trigger behavior

 1. Implement your custom Callout view which conforms to ```MGLCalloutView``` protocol.
 2. Implement ```-[MGLCalloutView presentWithContentSize:representedFromRect:inView:constrainedToView:animated:]``` and make your custom layout logic in there.

### Expected behavior
Custom frame set to the callout should remain persistent within the superview the callout was added to.

### Actual behavior
SDK moves the custom callout based on its own undefined rules.

### Regression
Issue is not reproducible in **ios-v3.4.0-beta.4**

Please review the test application attached.
Read "readme.txt" for details on how to run the test application.
[TestMapbox.zip](https://github.com/mapbox/mapbox-gl-native/files/660658/TestMapbox.zip)

frederoni:
Thank you for your report.
What changed between beta 4 and 5 is that the callout view stays open when you pan the map.
If you want the same behavior as in beta 4, you could call `-[MGLMapView deselectAnnotation:animated:]` from `-[MGLMapViewDelegate mapView:regionDidChangeAnimated:]` if you have a visible callout view.

If you want the behavior of beta 5 but with `representedFromRect` taken into account, then we need to figure out another solution.

andrewstay:
@frederoni the solution you have suggested does not work, as callout gets misplaced by the SDK before it gets dismissed by the client request. I implemented  the following code:
```
- (void)mapViewRegionIsChanging:(MGLMapView *)aView
{
    [self.mapView deselectAnnotation:self.mapView.selectedAnnotations.lastObject animated:YES];
}
```

Please review the video of the result.
https://www.dropbox.com/s/jwc28dofqoxqjn4/CallotMisplaced.mov?dl=0

1ec5:
(Noting that this regression was introduced in #6676.)

1ec5:
We’re repositioning the callout view by [setting its `center` property](https://github.com/mapbox/mapbox-gl-native/blob/f0d54947b21095223cd1719df459b6308fbfd097/platform/ios/src/MGLMapView.mm#L4830-L4832). One workaround might be to override `-center` and `-setCenter:` to be fixed at the desired location. It’s a bit unfortunate if a developer tries to use constraints to affix the callout view to a particular location, though.

[...]

PR #7646, opened 2017-01-09
Optional positioning and stickiness for callout view

frederoni:
This proposal makes the `MGLCalloutView` a bit more versatile by adding an `anchoredToAnnotation` and a `dismissesAutomatically` property. It also fixes issues like #7483 by falling back to the same behavior as in 3.4.0 beta 4 by default.

animation

/cc @friedbunny @1ec5 @boundsj

### DIFF
--- a/platform/ios/src/MGLCalloutView.h
+++ b/platform/ios/src/MGLCalloutView.h
@@ -46,6 +46,24 @@ NS_ASSUME_NONNULL_BEGIN
  */
 - (void)dismissCalloutAnimated:(BOOL)animated;
 
+@optional
+
+/**
+ A Boolean value indicating whether the callout view should be anchored to
+ the corresponding annotation. You can adjust the callout view’s precise location by
+ overriding -[UIView setCenter:]. The callout view will not be anchored to the
+ annotation if this optional property is unimplemented.
+ */
+@property (nonatomic, readonly, assign, getter=isAnchoredToAnnotation) BOOL anchoredToAnnotation;
+
+/**
+ A Boolean value indicating whether the callout view should be dismissed automatically
+ when the map view’s viewport changes. Note that a single tap on the map view
+ still dismisses the callout view regardless of the value of this property.
+ The callout view will be dismissed if this optional property is unimplemented.
+ */
+@property (nonatomic, readonly, assign) BOOL dismissesAutomatically;
+
 @end
 
 /**
--- a/platform/ios/src/MGLCompactCalloutView.m
+++ b/platform/ios/src/MGLCompactCalloutView.m
@@ -14,6 +14,14 @@
     return [[self alloc] init];
 }
 
+- (BOOL)isAnchoredToAnnotation {
+    return YES;
+}
+
+- (BOOL)dismissesAutomatically {
+    return NO;
+}
+
 - (void)setRepresentedObject:(id <MGLAnnotation>)representedObject
 {
     _representedObject = representedObject;
--- a/platform/ios/app/MBXCustomCalloutView.h
+++ b/platform/ios/app/MBXCustomCalloutView.h
@@ -7,4 +7,8 @@
  * callout title for demonstration purpose.
  */
 @interface MBXCustomCalloutView : UIView <MGLCalloutView>
+
+@property (nonatomic, assign, getter=isAnchoredToAnnotation) BOOL anchoredToAnnotation;
+@property (nonatomic, assign) BOOL dismissesAutomatically;
+
 @end
--- a/platform/ios/app/MBXCustomCalloutView.m
+++ b/platform/ios/app/MBXCustomCalloutView.m
@@ -17,6 +17,8 @@ static CGFloat const tipWidth = 10.0;
 }
 
 @synthesize representedObject = _representedObject;
+@synthesize anchoredToAnnotation = _anchoredToAnnotation;
+@synthesize dismissesAutomatically = _dismissesAutomatically;
 @synthesize leftAccessoryView = _leftAccessoryView;
 @synthesize rightAccessoryView = _rightAccessoryView;
 @synthesize delegate = _delegate;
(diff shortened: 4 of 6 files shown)


platform/ios/src/MGLCalloutView.h
```diff
@@ -36,6 +36,18 @@ NS_ASSUME_NONNULL_BEGIN
 @property (nonatomic, weak) id<MGLCalloutViewDelegate> delegate;
 
 /**
+ A Boolean value indicating whether the callout view should be positioned automatically.
+ The exact position can be adjusted by overriding -[UIView setCenter:]
+ */
+@property (nonatomic, assign) BOOL automaticPositioning;
```
1ec5: A Boolean-typed property shouldn’t be named with a noun phrase. How about `anchoredToAnnotation`, with a getter `isAnchoredToAnnotation`?

platform/ios/src/MGLCalloutView.h
```diff
@@ -36,6 +36,18 @@ NS_ASSUME_NONNULL_BEGIN
...
+
+/**
+ A Boolean value indicating whether the callout view should be dismissed or not
+ when the map view changes viewport.
+ */
+@property (nonatomic, assign) BOOL staysOpen;
```
1ec5: Can we reverse the logic and call it `dismissesAutomatically`? I think that would make the behavior clearer, as evidenced by the description you wrote for this property.

platform/ios/src/MGLCalloutView.h
1ec5: Positioning automatically can mean many things. I think the fundamental difference between MGLCompactCalloutView and the custom callout view that ran into this bug is that MGLCompactCalloutView is relatively positioned, while that custom callout view is absolutely positioned – kind of like `MGLRotationAlignmentMap` versus `MGLRotationAlignmentViewport`. So another possibility would be to reverse the logic and call this property `relativelyPositioned`.

1ec5: Mark these two properties as optional; adding a required member to a public protocol is a backwards-incompatible change. We’ll have to figure out what to do if the properties aren’t implemented.

1ec5: Mark these properties as `readonly`, because MGLMapView doesn’t care if the class implements `-setAutomaticPositioning:` – it’ll never call that setter anyways.

platform/ios/src/MGLCalloutView.h
```diff
@@ -46,6 +46,22 @@ NS_ASSUME_NONNULL_BEGIN
...
+/**
+ A Boolean value indicating whether the callout view should be anchored to
+ the corresponding annotation. The exact position can be adjusted by 
+ overriding -[UIView setCenter:]
+ */
+@property (nonatomic, readonly, assign, getter=isAnchoredToAnnotation) BOOL anchoredToAnnotation;
+
+/**
+ A Boolean value indicating whether the callout view should be dismissed automatically
+ when the map view changes viewport. Note that a single tap on the map view
```
1ec5: s/when the map view changes viewport/when the map view’s viewport changes/

platform/ios/src/MGLCalloutView.h
```diff
@@ -46,6 +46,22 @@ NS_ASSUME_NONNULL_BEGIN
...
+ still dismisses the callout view.
```
1ec5: …regardless of the value of this property.

platform/ios/src/MGLCalloutView.h
```diff
@@ -46,6 +46,22 @@ NS_ASSUME_NONNULL_BEGIN
  */
 - (void)dismissCalloutAnimated:(BOOL)animated;
 
+@optional
+
+/**
+ A Boolean value indicating whether the callout view should be anchored to
+ the corresponding annotation. The exact position can be adjusted by 
+ overriding -[UIView setCenter:]
```
1ec5: What happens if this property (and the other one) is omitted?

1ec5: Nit: avoid passive voice:

> You can adjust the callout view’s precise location by overriding `-[UIView setCenter:]`.

platform/ios/src/MGLCompactCalloutView.m
```diff
@@ -8,6 +8,16 @@ @implementation MGLCompactCalloutView
 }
 
 @synthesize representedObject = _representedObject;
+@synthesize anchoredToAnnotation = _anchoredToAnnotation;
+@synthesize dismissesAutomatically = _dismissesAutomatically;
+
+- (instancetype)init {
```
1ec5: `-init` is not the designated initializer for UIView instances; that’d be either `-initWithFrame:` or `-initWithCoder:`. This class’s `+platformCalloutView` implementation happens to call `-init`, but this is a bit of an antipattern.

Instead of maintaining ivars for read-only variables, implement `-anchoredToAnnotation` and `-dismissesAutomatically` to return `YES` and `NO`, respectively.

platform/ios/src/MGLCalloutView.h
frederoni: Good question. As of https://github.com/mapbox/mapbox-gl-native/pull/7646/commits/46dd0a6c402df4959fb6a20de4f874d6cd48bd47 , it defaults to `dismissesAutomatically`.

1ec5: Can you point out the default behavior in the documentation for both properties (so developers know whether they need to implement them or not)?

friedbunny:
Let’s also update the changelog entry for this feature/fix. Currently, it’s:

> Fixed an issue causing the callout view to be dismissed when panning around. (#6676)

1ec5:
The changelog entry is correct: in v3.3.7, panning incorrectly dismissed the callout view; in v3.4.0, the callout view will stay open. We can update the beta release notes to note this fix, however.

PR #7648, opened 2017-01-09
Clear out GeometryTileWorker::layers during redoLayout

kkaefer:
This avoids a corner case where redoLayout was called with a new set of data, but the layer objects were moved out during the previous call to redoLayout. When setting new data, we call setData() first, then `setLayout()`. The `setData()` call doesn't immediately process the tiles, since it will still be lacking layer data.

### DIFF
--- a/src/mbgl/tile/geometry_tile_worker.cpp
+++ b/src/mbgl/tile/geometry_tile_worker.cpp
@@ -215,6 +215,8 @@ void GeometryTileWorker::redoLayout() {
     BucketParameters parameters { id, obsolete, *featureIndex, mode };
 
     std::vector<std::vector<std::unique_ptr<Layer>>> groups = groupByLayout(std::move(*layers));
+    layers = nullopt;
+
     for (auto& group : groups) {
         if (obsolete) {
             return;
@@ -280,7 +282,7 @@ bool GeometryTileWorker::hasPendingSymbolDependencies() const {
 }
 
 void GeometryTileWorker::attemptPlacement() {
-    if (!data || !layers || !placementConfig) {
+    if (!placementConfig) {
         return;
     }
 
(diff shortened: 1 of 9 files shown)


src/mbgl/tile/geometry_tile_worker.cpp
```diff
@@ -215,6 +215,8 @@ void GeometryTileWorker::redoLayout() {
     BucketParameters parameters { id, obsolete, *featureIndex, mode };
 
     std::vector<std::vector<std::unique_ptr<Layer>>> groups = groupByLayout(std::move(*layers));
+    layers = nullopt;
```
kkaefer: Reset the `optional`

src/mbgl/tile/geometry_tile_worker.cpp
```diff
@@ -280,7 +282,7 @@ bool GeometryTileWorker::hasPendingSymbolDependencies() const {
 }
 
 void GeometryTileWorker::attemptPlacement() {
-    if (!data || !layers || !placementConfig) {
+    if (!placementConfig) {
```
kkaefer: We don't need access to `data` and `layers` anymore, since all of that is encoded in the `SymbolLayout` objects now. This change is necessary, since disengaging the `optional` will make this check fail when it is part of `redoLayout()` (called at the very end), meaning that there won't be any symbols until placement is triggered again.

jfirebaugh: I have a feeling at least one of these guards was needed to ensure that the body of `attemptPlacement` does not run if `setPlacementConfig` is called before `setData` and `setLayers`.

jfirebaugh:
Alternate fix added to #7616.

jfirebaugh:
Merged alternate fix.

Issue #7506, opened 2016-12-21
Publish macOS SDK documentation


1ec5:
The macOS SDK has a script for generating an HTML docset using jazzy, just like the iOS SDK does. We should publish the macOS SDK docset to mapbox.github.io, which already hosts documentation for auxiliary libraries such as Mapbox Java Services.

/cc @friedbunny @kkaefer

1ec5:
#7649 updated the documentation script to produce a standalone docset for mapbox.github.io and updated the deployment script to link to it. Leaving this issue open until the docset is published.

1ec5:
Documentation is now available at https://mapbox.github.io/mapbox-gl-native/macos/0.3.0-rc.1/. #7730 tracks a broken link and other errata.

PR #5653, opened 2016-07-12
[ios] fixes #4536 include podspec for static framework

frederoni:
Fixes #4536

PR #7649, opened 2017-01-09
Upgrade macOS deployment and distribution

1ec5:
Ported iOS deployment script improvements in #6581 and #6740 to macOS. (The main difference is that the macOS SDK is relying on GitHub for distribution rather than s3.)

Added CocoaPods podspecs to parallel the iOS SDK’s podspecs. The podspecs assume that the macOS SDK API reference will live on mapbox.github.io.

To recap:

* [x] Port deployment script improvements in #6581 and #6740
* [x] Add CocoaPods podspecs
* [x] Port local podspec generation from #5653
* [x] Link to future Dash docset
* [x] Update documentation to refer to future online API reference
* [x] Update screenshot

Prerequisite for #7506.

@friedbunny, would you mind checking whether I goofed anything up in the deployment script? Thanks!


frederoni:
I think we need to port parts of `package.sh` as well if we want to support local podspec installation and less manual copy-pasting during deploy.

1ec5:
Uploaded a new screenshot:

screenshot

1ec5:
The published docset will live at `https://mapbox.github.io/mapbox-gl-native/macos/{version}/`, with https://mapbox.github.io/mapbox-gl-native/macos/ serving as a redirect to the latest version. To build the published docset, which includes installation instructions, run `make xdocument STANDALONE=1`.

PR #7652, opened 2017-01-10
Rename npm package to "@mapbox/gl-native"

lucaswoj:


jfirebaugh:
I'm not sure we want to do this. If we do this, it implies we should rename `mapbox-gl-js` to `@mapbox/gl`, which would conflict if we ever forked https://www.npmjs.com/package/gl (and could be confusing even if we didn't).

Also, for comparison, this repo isn't at `https://github.com/mapbox/gl-native`.

lucaswoj:
All good points @jfirebaugh. I'm echoing the points made at https://github.com/mapbox/mapbox-gl-style-spec/pull/631#issuecomment-269002154. We should be sure to agree on our strategy going forward. 

> Also, for comparison, this repo isn't at https://github.com/mapbox/gl-native

Nor is GL JS published at "https://github.com/mapbox/mapbox-gl". We have already set the precedent that github repo names != npm package names. 

---

I can see two potential paths forward here that allow us to end up with a consistent naming scheme across GL projects

# Option 1: drop the `mapbox-` prefix, restore the `-js` suffix

 - `mapbox-gl` -> `@mapbox/gl-js`
 - `mapbox-gl-native` -> `@mapbox/gl-native`
 - `mapbox-gl-style-spec` -> `@mapbox/gl-style-spec`

# Option 2: copy the package names as-is

 - `mapbox-gl` -> `@mapbox/mapbox-gl`
 - `mapbox-gl-native` -> `@mapbox/mapbox-gl-native`
 - `mapbox-gl-style-spec` -> `@mapbox/mapbox-gl-style-spec`

jfirebaugh:
My preference:

* `@mapbox/mapbox-gl` (GitHub repo name will change to match this eventually)
* `@mapbox/mapbox-gl-native`
* `@mapbox/mapbox-style-spec`

mourner:
I prefer @lucaswoj's option 1 — it seems the least confusing and it fully matches current repository names (which I don't think we should rename just yet).

PR #7656, opened 2017-01-10
Add Proguard integration to CI

### LINKED ISSUE #7608: CI build for ProGuard
When preparing for a new release, we tend to run into a few ProGuard issues that require fixing beforehand. If we add a task in our CI to test while ProGuard is enabled we can catch any issues quicker. Thoughts @mapbox/android?

tobrun:
Closes #7608:
 - enables proguard for release builds
 - removes obsolete debug config
 - changes buildtype to release for `make android-ui-test`
 - update CI configuration to upload release version apk to Firebase.

Review @zugaldia


tobrun:
It seems that we aren't able to use the debug keystore while targeting release. 
It builds fine but when we try launching the test on Firebase:

` The application failed to install. ` 

I'm guessing that providing a keystore for signing will help resolve this.

platform/android/MapboxGLAndroidSDKTestApp/build.gradle
```diff
@@ -45,11 +45,9 @@ android {
     buildTypes {
         debug {
             testCoverageEnabled = true
-            minifyEnabled false
```
zugaldia: @tobrun Why does this line need to be removed? Isn't it better to leave the explicit setting?

zugaldia:
> I'm guessing that providing a keystore for signing will help resolve this.

Cool. Let me know when you're ready for another 👀 pass.

platform/android/MapboxGLAndroidSDKTestApp/build.gradle
tobrun: default behaviour is false so it's enforced by default, just trying to limit amount of code here

zugaldia: "Explicit is better than implicit." (That's my Pythonic component I guess ;)

tobrun:
Looked into this again and using a release build with instrumentation tests doesn't seem to be possible:

```
Permission Denial: starting instrumentation ComponentInfo{com.mapbox.mapboxsdk.testapp.test/android.support.test.runner.AndroidJUnitRunner} from pid=14384, uid=14384 not allowed because package com.mapbox.mapboxsdk.testapp.test does not have a signature matching the target com.mapbox.mapboxsdk.testapp
```

1ec5:
Since this PR is scheduled for Android SDK v5.0.0, it needs to be rebased and retargeted at the release-ios-v3.5.0-android-v5.0.0 branch.

tobrun:
I currently don't see a way of getting this working without introducing proguard during development (which is something we want to avoid at all costs). Closing.

PR #7664, opened 2017-01-10
Make MGLMapView.style property nullable

### LINKED ISSUE #7512: Warn when accessing MGLStyle before style is loaded

1ec5:
MGLMapView’s `style` property is now nullable (optional in Swift) and is set to `nil` while the style loads and in the event that the style has failed to load. The `styleURL` property remains null-resettable (implicitly-unwrapped-optional in Swift) as before.

Essentially, the API as it exists has a trap: the developer is tempted to have runtime styling code run synchronously after initializing the map view, for example in `-viewDidLoad`. Unfortunately, any runtime styling API operations they perform while the style loads [gets wiped out](https://github.com/mapbox/mapbox-gl-native/blob/5da85bafd8266ebea0c1e48e70f6e93fbbfbeecd/src/mbgl/style/style.cpp#L100-L101). Instead, they must wait until the style has finished loading, in `-[MGLMapViewDelegate mapView:didFinishLoadingStyle:]`.

We want developers to have a way of knowing whether the style has finished loading so they can easily find out why their runtime styling code isn’t working as expected. The problem with adding a warning in `-style`, as #7639 does, is that we’d assume the developer intends to use the MGLStyle object they obtain from `MGLMapView.style`. So if the developer calls the `style` getter to check whether the style has loaded, they get console spew. It turns out that, if the developer adds a key-value observer (or binding) to the `style` key, the KVO system calls the `style` getter even just before the style is set (as a side effect of `-willChangeValueForKey:`).

This PR avoids that whole mess by preventing the developer from accessing MGLStyle until MGLStyle can be used fruitfully. Swift developers must acknowledge the possibility that `style` isn’t set yet. Unfortunately, Objective-C developers will have little indication as to the problem without stepping in the debugger.

Fixes #7512.

/ref https://github.com/mapbox/mapbox-gl-native/pull/7639#issuecomment-271427589
/cc @boundsj @ericrwolfe @incanus @frederoni @friedbunny @jfirebaugh

### DIFF
--- a/platform/darwin/test/MGLStyleLayerTests.m
+++ b/platform/darwin/test/MGLStyleLayerTests.m
@@ -4,9 +4,7 @@
 
 #define TEST_STRICT_NAMING_CONVENTIONS 0
 
-@implementation MGLStyleLayerTests {
-    XCTestExpectation *_styleLoadingExpectation;
-}
+@implementation MGLStyleLayerTests
 
 @dynamic layerType;
 
@@ -21,20 +19,19 @@
     [vc view]; // Force load xib
     _mapView = [[MGLMapView alloc] initWithFrame:CGRectMake(0, 0, 256, 256) styleURL:styleURL];
     [vc.view addSubview:_mapView];
-    _mapView.delegate = self;
 #else
     NSWindowController *windowController = [[NSWindowController alloc] initWithWindowNibName:@"MGLStyleLayerTests" owner:self];
-    self.mapView.styleURL = styleURL;
+    NSView *contentView = windowController.window.contentView;
+    _mapView = [[MGLMapView alloc] initWithFrame:contentView.bounds styleURL:styleURL];
+    [contentView addSubview:_mapView];
     [windowController showWindow:nil];
 #endif
-    XCTAssertNil(self.mapView.style);
-    _styleLoadingExpectation = [self expectationWithDescription:@"Map view should finish loading style."];
+    _mapView.delegate = self;
+    XCTAssertNil(_mapView.style);
+    [self keyValueObservingExpectationForObject:self.mapView keyPath:@"style" handler:^BOOL(MGLMapView * _Nonnull observedMapView, NSDictionary * _Nonnull change) {
+        return observedMapView.style != nil;
+    }];
     [self waitForExpectationsWithTimeout:1 handler:nil];
-}
-
-- (void)mapView:(MGLMapView *)mapView didFinishLoadingStyle:(MGLStyle *)style {
-    XCTAssertNotNil(mapView.style);
-    [_styleLoadingExpectation fulfill];
 }
 
 - (void)tearDown {
--- a/platform/darwin/test/MGLStyleLayerTests.m
+++ b/platform/darwin/test/MGLStyleLayerTests.m
@@ -11,18 +11,25 @@
 - (void)setUp {
     [super setUp];
     [MGLAccountManager setAccessToken:@"pk.feedcafedeadbeefbadebede"];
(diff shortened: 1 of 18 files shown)


1ec5:
Missed a couple files due to a merge conflict. Rebased and force-pushed the proper fix.

1ec5:
@boundsj pointed out in chat that #6386 will provide a convenient alternative to implementing `-mapView:didFinishLoadingStyle:`, in the event that the developer would rather perform runtime styling operations inside `-viewDidLoad`.

boundsj:
I think this looks good and makes a lot of sense 👍 

I'm not sure about the macOS test failure (especially after efcf574)

platform/darwin/test/MGLStyleLayerTests.m
```diff
@@ -21,22 +19,21 @@ - (void)setUp {
     [vc view]; // Force load xib
     _mapView = [[MGLMapView alloc] initWithFrame:CGRectMake(0, 0, 256, 256) styleURL:styleURL];
     [vc.view addSubview:_mapView];
-    _mapView.delegate = self;
 #else
     NSWindowController *windowController = [[NSWindowController alloc] initWithWindowNibName:@"MGLStyleLayerTests" owner:self];
-    self.mapView.styleURL = styleURL;
+    NSView *contentView = windowController.window.contentView;
+    _mapView = [[MGLMapView alloc] initWithFrame:contentView.bounds styleURL:styleURL];
```
boundsj: 👍

1ec5:
2b743bc3011e22f285bef1c2b3a47c79935c21a8 fixes the macOS test failure by creating MGLMapView programmatically for both platforms so that the initializer sets the intended style URL instead of the default Streets URL.

1ec5:
One downside is that this change, in combination with #6386, would lead to a weird scenario where you could set the `style` property but getting it right after would return a different value (`nil`). Cocoa APIs tend to address this problem using a `loading` or `loaded` property (which we’d put on MGLStyle itself).

incanus:
I have mixed feelings about making `style` optional due to the verbosity of `?` and `!` that would be introduced in runtime styling code. 

> So if the developer calls the style getter to check whether the style has loaded, they get console spew.

Is there a way that we can check whether `style` is being messaged or not to avoid this? For example, a gate on its methods?

1ec5:
> I have mixed feelings about making `style` optional due to the verbosity of `?` and `!` that would be introduced in runtime styling code.

Agreed, it can get repetitive. Fortunately, any code inside `MGLMapViewDelegate.mapView(_:didFinishLoading:)` can take advantage of the `style` argument, which isn’t optional. Code located elsewhere, such as in an action method, can mitigate the repetitiveness by unwrapping the optional upfront:

```swift
guard let style = mapView.style else { return }
```

or:

```swift
let style = mapView.style!
```

> Is there a way that we can check whether `style` is being messaged or not to avoid this? For example, a gate on its methods?

Yes, it would be possible to warn or assert if any method of MGLStyle is called before it loads. That would be closer to the `loading` pattern described in https://github.com/mapbox/mapbox-gl-native/pull/7664#issuecomment-271700216. However, in that case, aren’t we really using MGLStyle as a placeholder for `nil`?

1ec5:
Some integration tests started failing as soon as this PR landed, even though it’s fine on my machine and was fine on the branch. A fix is in #7684.

Issue #7453, opened 2016-12-15
MGLShapeSource should have a shape array convenience initializer


1ec5:
Despite `-[MGLShapeSource initWithIdentifier:shape:options:]` documentation mentioning the possibility of using an MGLShapeCollectionFeature to manage multiple features, developers unfamiliar with the Simple Features specification are unlikely to discover this capability and will assume that a shape source can only be used with one shape. We should add the following convenience initializers to MGLShapeSource so developers don’t have to go through MGLShapeCollection explicitly:

* `-initWithIdentifier:shapes:options:` would wrap the shapes in an MGLShapeCollection.
* `-initWithIdentifier:features:options:` would wrap the features in an MGLShapeCollectionFeature.

If we encounter a `shapes` array that contains an MGLFeature, should we treat it as an ordinary shape, or should we upgrade to an MGLShapeCollectionFeature and wrap the other shapes in features?

I’d prefer not to add a `shapes` or `features` property, since that blurs the definition of the kind of data this source actually holds.

/cc @boundsj @frederoni @bsudekum

boundsj:
> If we encounter a shapes array that contains an MGLFeature, should we treat it as an ordinary shape, or should we upgrade to an MGLShapeCollectionFeature and wrap the other shapes in features?

This is a good question. I think that developers who are unfamiliar with the Simple Features specification will also be unfamiliar with the distinction between geometry objects (shapes) and feature objects. 

Since creating an `MGLFeature` is as easy as creating an `MGLShape` (even if the meta data properties of `MGLFeature` are not used) and because most developers will eventually realize they want to create feature objects to unlock the full power of runtime and data driven styling, I think that offering only `-initWithIdentifier:features:options:` and directing developers to that API in documentation and examples could help simplify one of the most initially confusing parts of the runtime styling API and reduce the support workload a bit.

The existing APIs (and a new `-initWithIdentifier:shapes:options:` that treats any `MGLFeature`as an ordinary shape) should suffice for developers who really want or need to work with geometries for some reason.

boundsj:
Fixed in #7665 on the release-ios-v3.4.0 branch.

Issue #7454, opened 2016-12-15
MGLFeature should contain MGLShape, not inherit from it


1ec5:
MGLFeature should be a class, not a protocol, and it should contain an MGLShape in its `shape` property. The various MGL*Feature classes that conform to the MGLFeature protocol should go away. There should also be an MGLFeatureCollection protocol. MGLShape, MGLFeature, and MGLFeatureCollection should conform to a new MGLGeoJSONRepresentable protocol.

In #5110, I needed to add classes for all the possible feature types that could be returned from the feature querying API. I made a series of MGL*Feature classes: each class conforms to the MGLFeature protocol and inherits from an MGLShape class (which corresponds to a GeoJSON geometry type). For example, MGLPolygonFeature inherits from MGLPolygon. By contrast, in the GeoJSON specification, a feature _contains_ a geometry; it isn’t a kind of geometry.

I don’t remember if the taxonomy I implemented was a misreading of the GeoJSON specification on my part or a deliberate shortcut, but it proved to be very convenient for working with feature querying results. You can take the feature querying results and turn right around and add those results to the map as annotations (since some MGLShape types are annotation types). Unfortunately, this taxonomy works less well with the runtime styling API: for #7453, what happens when an array of MGLShapes is really a mixture of shapes and features?

The proposed refactoring mirrors the GeoJSON specification exactly and removes the rough edges around the runtime styling API.

/ref #5201 https://github.com/mapbox/mapbox-gl-native/issues/7160#issuecomment-264083136
/cc @boundsj @frederoni @jfirebaugh

1ec5:
Since `+[MGLShape shapeWithData:encoding:error:]` could return any instance of a class conforming to MGLGeoJSONRepresentable, not necessarily an MGLShape, we’ll need to pull out this method and `-[MGLShape geoJSONDataUsingEncoding:]` into a new class, MGLGeoJSONSerialization.

fabian-guerra:
# This is our current Annotation's class hierarchy
<img width="1019" alt="annotation_api" src="https://user-images.githubusercontent.com/2745166/29984404-520d8dea-8f28-11e7-8a69-c0de4be95e8a.png">

1ec5:
[...]

PR #7665, opened 2017-01-10
Add convenience initializers to shape source

boundsj:
Fixes https://github.com/mapbox/mapbox-gl-native/issues/7453

This adds two new convenience initializers to MGLShapeSource:

- `-initWithIdentifier:features:options:` takes an array of shape objects that conform to MGLFeature, inserts them in a shape collection feature, and creates a source with that shape. 

- `-initWithIdentifier:shapes:options:` does the same but with concrete MGLShape objects that get added to a shape collection.

Although both are technically convenience initializers, developers will most often want to use -`initWithIdentifier:features:options:` since features objects unlock key runtime styling functionality. So, the inline example has been updated to demonstrate `MGLShapeSource` usage with `-initWithIdentifier:features:options:`

The type system will help ensure that Swift developers use the correct convenience initializer for objects that are features (i.e. it is not possible to send plain shapes to the features initializer without an unsafe cast). There is currently no warning system for this though so Objective-C developers will not be protected from bugs where layer properties or predicate filters don't work as expected because plain shapes were previously used with the features initializer.

cc @frederoni @bsudekum @ericrwolfe

### DIFF
--- a/platform/darwin/docs/guides/Working with GeoJSON Data.md
+++ b/platform/darwin/docs/guides/Working with GeoJSON Data.md
@@ -36,6 +36,16 @@ resulting shape or feature object into the
 the map, or you can use the object and its properties to power non-map-related
 functionality in your application.
 
+To include multiple shapes in the source, create and pass an `MGLShapeCollection` or
+ `MGLShapeCollectionFeature` object to 
+ `-[MGLShapeSource initWithIdentifier:shape:options:]`. Alternatively, use the
+ `-[MGLShapeSource initWithIdentifier:features:options:]` or 
+ `-[MGLShapeSource initWithIdentifier:shapes:options:]` method to create a shape source 
+ with an array. `-[MGLShapeSource initWithIdentifier:features:options:]` accepts only `MGLFeature`
+ instances, such as `MGLPointFeature` objects, whose attributes you can use when
+ applying a predicate to `MGLVectorStyleLayer` or configuring a style layer’s
+ appearance.
+
 ## Extracting GeoJSON data from the map
 
 Any `MGLShape`, `MGLFeature`, or `MGLShapeCollectionFeature` object has an
--- a/platform/darwin/src/MGLShapeSource.h
+++ b/platform/darwin/src/MGLShapeSource.h
@@ -103,8 +103,7 @@ extern const MGLShapeSourceOption MGLShapeSourceOptionSimplificationTolerance;
      CLLocationCoordinate2D(latitude: 38.91, longitude: -77.04),
  ]
  let polyline = MGLPolylineFeature(coordinates: &coordinates, count: UInt(coordinates.count))
- let shape = MGLShapeCollectionFeature(shapes: [polyline])
- let source = MGLShapeSource(identifier: "lines", shape: shape, options: nil)
+ let source = MGLShapeSource(identifier: "lines", features: [polyline], options: nil)
  mapView.style?.addSource(source)
  ```
  */
@@ -131,7 +130,9 @@ extern const MGLShapeSourceOption MGLShapeSourceOptionSimplificationTolerance;
  To specify attributes about the shape, use an instance of an `MGLShape`
  subclass that conforms to the `MGLFeature` protocol, such as `MGLPointFeature`.
  To include multiple shapes in the source, use an `MGLShapeCollection` or
- `MGLShapeCollectionFeature` object.
+ `MGLShapeCollectionFeature` object, or use the
+ `-initWithIdentifier:features:options:` or 
+ `-initWithIdentifier:shapes:options:` methods.
  
  To create a shape from GeoJSON source code, use the
  `+[MGLShape shapeWithData:encoding:error:]` method.
@@ -142,6 +143,45 @@ extern const MGLShapeSourceOption MGLShapeSourceOptionSimplificationTolerance;
  @return An initialized shape source.
  */
 - (instancetype)initWithIdentifier:(NSString *)identifier shape:(nullable MGLShape *)shape options:(nullable NS_DICTIONARY_OF(MGLShapeSourceOption, id) *)options NS_DESIGNATED_INITIALIZER;
+
+/**
+ Returns a shape source with an identifier, an array of features, and a dictionary
+ of options for the source.
+ 
+ Unlike `-initWithIdentifier:shapes:options:`, this method accepts `MGLFeature`
+ instances, such as `MGLPointFeature` objects, whose attributes you can use when
+ applying a predicate to `MGLVectorStyleLayer` or configuring a style layer’s
+ appearance.
+
+ To create a shape from GeoJSON source code, use the
+ `+[MGLShape shapeWithData:encoding:error:]` method.
+
+ @param identifier A string that uniquely identifies the source.
+ @param features An array of objects that conform to the MGLFeature protocol.
+ @param options An `NSDictionary` of options for this source.
+ @return An initialized shape source.
+ */
+- (instancetype)initWithIdentifier:(NSString *)identifier features:(NS_ARRAY_OF(MGLShape<MGLFeature> *) *)features options:(nullable NS_DICTIONARY_OF(MGLShapeSourceOption, id) *)options;
+
+/**
+ Returns a shape source with an identifier, an array of shapes, and a dictionary of
+ options for the source.
+ 
+ Any `MGLFeature` instance passed into this initializer is treated as an ordinary
+ shape, causing any attributes to be inaccessible to an `MGLVectorStyleLayer` when
+ evaluating a predicate or configuring certain layout or paint attributes. To 
+ preserve the attributes associated with each feature, use the 
+ `-initWithIdentifier:features:options:` method instead.
+
+ To create a shape from GeoJSON source code, use the
+ `+[MGLShape shapeWithData:encoding:error:]` method.
+
+ @param identifier A string that uniquely identifies the source.
+ @param shapes An array of shapes; each shape is a member of a concrete subclass of MGLShape.
+ @param options An `NSDictionary` of options for this source.
+ @return An initialized shape source.
+ */
+- (instancetype)initWithIdentifier:(NSString *)identifier shapes:(NS_ARRAY_OF(MGLShape *) *)shapes options:(nullable NS_DICTIONARY_OF(MGLShapeSourceOption, id) *)options;
 
 #pragma mark Accessing a Source’s Content
 
--- a/platform/darwin/src/MGLShapeSource.mm
+++ b/platform/darwin/src/MGLShapeSource.mm
@@ -56,6 +56,21 @@ const MGLShapeSourceOption MGLShapeSourceOptionSimplificationTolerance = @"MGLSh
     return self;
 }
 
+- (instancetype)initWithIdentifier:(NSString *)identifier features:(NS_ARRAY_OF(MGLShape<MGLFeature> *) *)features options:(nullable NS_DICTIONARY_OF(MGLShapeSourceOption, id) *)options {
+    for (id <MGLFeature> feature in features) {
+        if (![feature conformsToProtocol:@protocol(MGLFeature)]) {
+            [NSException raise:NSInvalidArgumentException format:@"The object %@ included in the features argument does not conform to the MGLFeature protocol.", feature];
+        }
+    }
+    MGLShapeCollectionFeature *shapeCollectionFeature = [MGLShapeCollectionFeature shapeCollectionWithShapes:features];
+    return [self initWithIdentifier:identifier shape:shapeCollectionFeature options:options];
+}
+
+- (instancetype)initWithIdentifier:(NSString *)identifier shapes:(NS_ARRAY_OF(MGLShape *) *)shapes options:(nullable NS_DICTIONARY_OF(MGLShapeSourceOption, id) *)options {
+    MGLShapeCollection *shapeCollection = [MGLShapeCollection shapeCollectionWithShapes:shapes];
+    return [self initWithIdentifier:identifier shape:shapeCollection options:options];
+}
+
 - (instancetype)initWithRawSource:(mbgl::style::GeoJSONSource *)rawSource {
     return [super initWithRawSource:rawSource];
 }
--- a/platform/darwin/test/MGLShapeSourceTests.mm
+++ b/platform/darwin/test/MGLShapeSourceTests.mm
@@ -278,4 +278,43 @@
     XCTAssert(shape.shapes.count == 6, @"Shape collection should contain 6 shapes");
 }
 
+- (void)testMGLShapeSourceWithFeaturesConvenienceInitializer {
+    CLLocationCoordinate2D coordinates[] = {
+        CLLocationCoordinate2DMake(100.0, 0.0),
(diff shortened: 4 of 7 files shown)


boundsj:
Do we actually need a `-initWithIdentifier:shapes:options:`? It's nice that we support geometries when GeoJSON is loaded via a URL (or by the existing `shape` API) but I think, to avoid confusion, that we could get away with just `-initWithIdentifier:features:options:`.

platform/darwin/src/MGLShapeSource.mm
```diff
@@ -56,6 +56,16 @@ - (instancetype)initWithIdentifier:(NSString *)identifier shape:(nullable MGLSha
     return self;
 }
 
+- (instancetype)initWithIdentifier:(NSString *)identifier features:(NS_ARRAY_OF(MGLFeature) *)features options:(nullable NS_DICTIONARY_OF(MGLShapeSourceOption, id) *)options {
+    MGLShapeCollectionFeature *shapeCollectionFeature = [MGLShapeCollectionFeature shapeCollectionWithShapes:features];
```
boundsj: Should we check the type of each object in `features` and log a one time warning if a non-feature makes it through?

platform/darwin/src/MGLShapeSource.h
```diff
@@ -143,6 +142,41 @@ extern const MGLShapeSourceOption MGLShapeSourceOptionSimplificationTolerance;
  */
 - (instancetype)initWithIdentifier:(NSString *)identifier shape:(nullable MGLShape *)shape options:(nullable NS_DICTIONARY_OF(MGLShapeSourceOption, id) *)options NS_DESIGNATED_INITIALIZER;
 
+/**
+ Returns a shape source with an identifier, a shape that is a 
+ `MGLShapeCollectionFeature` holding the features passed in, and dictionary of
```
1ec5: This method takes an array of features, not a single shape collection feature object.

platform/darwin/src/MGLShapeSource.h
```diff
@@ -143,6 +142,41 @@ extern const MGLShapeSourceOption MGLShapeSourceOptionSimplificationTolerance;
  */
 - (instancetype)initWithIdentifier:(NSString *)identifier shape:(nullable MGLShape *)shape options:(nullable NS_DICTIONARY_OF(MGLShapeSourceOption, id) *)options NS_DESIGNATED_INITIALIZER;
 
+/**
+ Returns a shape source with an identifier, a shape that is a 
+ `MGLShapeCollectionFeature` holding the features passed in, and dictionary of 
+ options for the source.
+ 
+ Passing an array of `MGLShape` subclass instances that conform to the
```
1ec5: This is a given.

platform/darwin/src/MGLShapeSource.h
```diff
@@ -143,6 +142,41 @@ extern const MGLShapeSourceOption MGLShapeSourceOptionSimplificationTolerance;
...
+ @param identifier A string that uniquely identifies the source.
+ @param features An array of `MGLShape` subclass instances that conform to the `MGLFeature` protocol
+ @param options An `NSDictionary` of options for this source.
+ @return An initialized shape source.
+ */
+- (instancetype)initWithIdentifier:(NSString *)identifier features:(NS_ARRAY_OF(id <MGLFeature>) *)features options:(nullable NS_DICTIONARY_OF(MGLShapeSourceOption, id) *)options;
+
+/**
+ Returns a shape source with an identifier, a shape that is a
+ `MGLShapeCollection` holding the shapes passed in, and dictionary of
```
1ec5: This method takes an array of shapes, not a single shape collection object.

platform/darwin/src/MGLShapeSource.h
```diff
@@ -143,6 +142,41 @@ extern const MGLShapeSourceOption MGLShapeSourceOptionSimplificationTolerance;
...
+ Passing an array of `MGLShape` subclass instances that conform to the 
+ `MGLFeature` protocol, such as `MGLPointFeature`, will allow for specifying
+ attributes about the shapes that can be used to help control how they appear
+ on the map using `MGLStyleLayer` properties and filters.
+
+ To create a shape from GeoJSON source code, use the
+ `+[MGLShape shapeWithData:encoding:error:]` method.
+
+ @param identifier A string that uniquely identifies the source.
+ @param features An array of `MGLShape` subclass instances that conform to the `MGLFeature` protocol
```
1ec5: > An array of objects that conform to the `MGLFeature` protocol.

platform/darwin/src/MGLShapeSource.h
```diff
@@ -143,6 +142,41 @@ extern const MGLShapeSourceOption MGLShapeSourceOptionSimplificationTolerance;
...
+/**
+ Returns a shape source with an identifier, a shape that is a
+ `MGLShapeCollection` holding the shapes passed in, and dictionary of
+ options for the source.
+
+ To create a shape from GeoJSON source code, use the
+ `+[MGLShape shapeWithData:encoding:error:]` method.
+
+ @param identifier A string that uniquely identifies the source.
+ @param shapes An array of `MGLShape` subclass instances
```
1ec5: Missing a period, and “subclass instances” is a bit confusing:

> An array of shapes; each shape is a member of a concrete subclass of `MGLShape`.

platform/darwin/src/MGLShapeSource.mm
```diff
@@ -56,6 +56,16 @@ - (instancetype)initWithIdentifier:(NSString *)identifier shape:(nullable MGLSha
     return self;
 }
 
+- (instancetype)initWithIdentifier:(NSString *)identifier features:(NS_ARRAY_OF(MGLFeature) *)features options:(nullable NS_DICTIONARY_OF(MGLShapeSourceOption, id) *)options {
```
1ec5: `NS_ARRAY_OF(MGLFeature) *` resolves to `NSArray<MGLFeature> *`, meaning the passed-in value must be an NSArray object that conforms to the MGLFeature protocol. I think what you meant was `NS_ARRAY_OF(id <MGLFeature>) *`.

platform/darwin/src/MGLShapeSource.mm
1ec5: Since lightweight generic qualifiers are completely optional in Objective-C, I think it’s fairly likely that a developer would incorrectly attempt to pass in an array of MGLShapes, especially if we get rid of `-initWithIdentifier:shapes:options:`. I think we should raise an exception in that case, since the value would contradict the type expected by this method.

platform/darwin/test/MGLShapeSourceTests.mm
```diff
@@ -278,4 +278,38 @@ - (void)testMGLShapeSourceWithShapeCollectionFeatures {
...
+        CLLocationCoordinate2DMake(100.0, 1.0),
+        CLLocationCoordinate2DMake(100.0, 0.0)};
+
+    MGLPolygonFeature *polygonFeature = [MGLPolygonFeature polygonWithCoordinates:coordinates count:sizeof(coordinates)/sizeof(coordinates[0]) interiorPolygons:nil];
+
+    MGLShapeSource *source = [[MGLShapeSource alloc] initWithIdentifier:@"source-id" features:@[polygonFeature] options:nil];
+    MGLShapeCollectionFeature *shape = (MGLShapeCollectionFeature *)source.shape;
+
+    XCTAssertTrue([shape isKindOfClass:[MGLShapeCollectionFeature class]]);
+    XCTAssert(shape.shapes.count == 1, @"Shape collection should contain 1 shape");
```
1ec5: Use `XCTAssertEqual()` (×2).

platform/darwin/src/MGLShapeSource.h
boundsj: It does _take_ an array but that is not what it returns in the "shape" property of the instantiated object. Isn't that what this sentence is trying to explain?

1ec5: This sentence explains what you pass into the method.

1ec5:
Also, can you see if the “Working with GeoJSON Data” guide needs any updates?

platform/darwin/src/MGLShapeSource.h
boundsj: Maybe this could be a `@note` instead? My thinking is that we should take every chance we have to point out the subtle difference between shapes and features. If we do get rid of `-initWithIdentifier:shapes:options:` then this note could probably go away completely.

1ec5: > Unlike `-initWithIdentifier:shapes:options:`, this method accepts `MGLFeature` instances, such as `MGLPointFeature` objects, which you can filter using `MGLVectorStyleLayer.predicate` and whose appearance you can configure using the properties of `MGLVectorStyleLayer`’s concrete subclasses.

platform/darwin/src/MGLShapeSource.h
```diff
@@ -143,6 +142,41 @@ extern const MGLShapeSourceOption MGLShapeSourceOptionSimplificationTolerance;
...
+ @param options An `NSDictionary` of options for this source.
+ @return An initialized shape source.
+ */
+- (instancetype)initWithIdentifier:(NSString *)identifier features:(NS_ARRAY_OF(id <MGLFeature>) *)features options:(nullable NS_DICTIONARY_OF(MGLShapeSourceOption, id) *)options;
+
+/**
+ Returns a shape source with an identifier, a shape that is a
+ `MGLShapeCollection` holding the shapes passed in, and dictionary of
+ options for the source.
+
```
1ec5: > Any `MGLFeature` instance passed into this initializer is treated as an ordinary shape, causing any attributes to be inaccessible to an `MGLVectorStyleLayer` for the purposes of filtering or configuring appearance. To preserve the attributes associated with each feature, use the `-initWithIdentifier:features:options:` or `-initWithIdentifier:shape:options:` method instead.

platform/darwin/src/MGLShapeSource.h
```diff
@@ -143,6 +142,41 @@ extern const MGLShapeSourceOption MGLShapeSourceOptionSimplificationTolerance;
  */
```
1ec5: The documentation for this method currently says:

> To include multiple shapes in the source, use an `MGLShapeCollection` or `MGLShapeCollectionFeature` object.

Change it to:

> To include multiple shapes in the source, use an `MGLShapeCollection` or `MGLShapeCollectionFeature` object, or use the `-initWithIdentifier:features:options:` or `-initWithIdentifierr:shapes:options:` method.

platform/darwin/src/MGLShapeSource.h
boundsj: Ah of course! I think not seeing "initialized with" threw me off for some reason today :D

boundsj:
Updated the guide in https://github.com/mapbox/mapbox-gl-native/pull/7665/commits/6ba0034aea6ab709763d83bd138531ac31ddbe5a

platform/darwin/src/MGLShapeSource.h
boundsj: This was done in https://github.com/mapbox/mapbox-gl-native/pull/7665/commits/6ba0034aea6ab709763d83bd138531ac31ddbe5a

platform/darwin/docs/guides/Working with GeoJSON Data.md
```diff
@@ -36,6 +36,17 @@ resulting shape or feature object into the
 the map, or you can use the object and its properties to power non-map-related
 functionality in your application.
 
+To include multiple shapes in the source, create and pass an `MGLShapeCollection` or
+ `MGLShapeCollectionFeature` object to 
+ `-[MGLShapeSource initWithIdentifier:shape:options:]`. Or, use the
```
1ec5: s/Or/Alternatively/

platform/darwin/docs/guides/Working with GeoJSON Data.md
```diff
@@ -36,6 +36,17 @@ resulting shape or feature object into the
...
+ `-[MGLShapeSource initWithIdentifier:features:options:]` or 
+ `-[MGLShapeSource initWithIdentifier:shapes:options:]` methods to create shape
```
1ec5: > method to create a shape source with an array

platform/darwin/src/MGLShapeSource.mm
```diff
@@ -56,6 +56,21 @@ - (instancetype)initWithIdentifier:(NSString *)identifier shape:(nullable MGLSha
     return self;
 }
 
+- (instancetype)initWithIdentifier:(NSString *)identifier features:(NS_ARRAY_OF(id <MGLFeature>) *)features options:(nullable NS_DICTIONARY_OF(MGLShapeSourceOption, id) *)options {
+    for (id <MGLFeature> feature in features) {
+        if (![feature conformsToProtocol:@protocol(MGLFeature)]) {
+            [NSException raise:NSInvalidArgumentException format:@"The object %@ included in the features argument does not conform to the MGLFeature protocol.", feature];
+        }
+    }
+    MGLShapeCollectionFeature *shapeCollectionFeature = [MGLShapeCollectionFeature shapeCollectionWithShapes:(NSArray *)features];
```
1ec5: You likely had to cast here because `+[MGLShapeCollectionFeature shapeCollectionWithShapes:]` accepts an array of `MGLShape <MGLFeature> *` instead of an array of `id <MGLFeature>`. Let’s make this method and `+[MGLShapeCollectionFeature shapeCollectionWithShapes:]` consistent with each other. (I have no strong preference one way or the other, but `id <MGLFeature>` might leave more options open for fixing #7454.)

boundsj: Since `MGLShapeCollectionFeature` inherits from `MGLShapeCollection` and `MGLShapeCollection` uses an `NS_ARRAY_OF(MGLShape *)` I don't think I can change `MGLShapeCollectionFeature` since it would require the change to bubble up to `MGLShapeCollection` 😬 

The implementation of `MGLShapeCollectionFeature` does not use lightweight generics (just a plain old NSArray). I made the same change to `MGLShapeSource` in https://github.com/mapbox/mapbox-gl-native/pull/7665/commits/6287c6bc3bbf6c7df7653c84318b01f026e910f5. Are you ok with this solution for this case, for now?

1ec5: The implementation of `+[MGLShapeCollectionFeature shapeCollectionWithShapes:]` uses an unqualified `NSArray` in its signature, but the declaration has a qualified `NSArray<MGLShape<MGLFeature> *> *`. Omitting the qualifier in the implementation only takes away opportunities for the compiler to catch mistakes. We should qualify as much as possible.

Since `+[MGLShapeCollectionFeature shapeCollectionWithShapes:]` is declared to take an `NSArray<MGLShape<MGLFeature> *> *` rather than the broader type `NSArray<id<MGLFeature>> *`, `-[MGLShapeSource initWithIdentifier:features:options:]` should take an `NSArray<MGLShape<MGLFeature> *> *` as well.

boundsj: Got it. This is done in https://github.com/mapbox/mapbox-gl-native/pull/7665/commits/e85da7f9998774a300d933089269bd651d2a7f44. I'll merge after the build passes.

PR #7670, opened 2017-01-11
Fix miscellaneous static analyzer warnings

1ec5:
This PR fixes the following static analyzer warnings, which were introduced in #6709, #7377, #6559, and #7300:

```
/path/to/mapbox-gl-native/platform/darwin/src/MGLNetworkConfiguration.m:16:9: Null is returned from a method that is expected to return a non-null value
/path/to/mapbox-gl-native/platform/darwin/src/MGLRasterSource.mm:63:23: Function call argument is an uninitialized value (within a call to 'make_unique')
/path/to/mapbox-gl-native/platform/darwin/src/MGLShape.mm:98:10: The left expression of the compound assignment is an uninitialized value. The computed value will also be garbage
/path/to/mapbox-gl-native/platform/ios/src/MGLAnnotationImage.m:60:54: Null passed to a callee that requires a non-null 1st parameter
/path/to/mapbox-gl-native/platform/ios/src/MGLAnnotationImage.m:65:10: The left expression of the compound assignment is an uninitialized value. The computed value will also be garbage
/path/to/mapbox-gl-native/platform/macos/src/MGLAnnotationImage.m:61:10: The left expression of the compound assignment is an uninitialized value. The computed value will also be garbage
/path/to/mapbox-gl-native/platform/macos/src/NSImage+MGLAdditions.mm:13:5: Returning 'self' while it is not set to the result of '[(super or self) init...]'
```

Working towards #7668.

/cc @frederoni @incanus

### DIFF
--- a/platform/darwin/src/MGLNetworkConfiguration.m
+++ b/platform/darwin/src/MGLNetworkConfiguration.m
@@ -1,5 +1,4 @@
 #import "MGLNetworkConfiguration.h"
-#import "NSProcessInfo+MGLAdditions.h"
 
 @implementation MGLNetworkConfiguration
 
@@ -12,9 +11,6 @@
 }
 
 + (instancetype)sharedManager {
-    if (NSProcessInfo.processInfo.mgl_isInterfaceBuilderDesignablesAgent) {
-        return nil;
-    }
     static dispatch_once_t onceToken;
     static MGLNetworkConfiguration *_sharedManager;
     void (^setupBlock)() = ^{
--- a/platform/darwin/src/MGLRasterSource.mm
+++ b/platform/darwin/src/MGLRasterSource.mm
@@ -51,7 +51,7 @@ static const CGFloat MGLRasterSourceRetinaTileSize = 512;
     if (self = [super initWithIdentifier:identifier tileURLTemplates:tileURLTemplates options:options]) {
         mbgl::Tileset tileSet = MGLTileSetFromTileURLTemplates(tileURLTemplates, options);
         
-        uint16_t tileSize;
+        uint16_t tileSize = MGLRasterSourceRetinaTileSize;
         if (NSNumber *tileSizeNumber = options[MGLTileSourceOptionTileSize]) {
             if (![tileSizeNumber isKindOfClass:[NSNumber class]]) {
                 [NSException raise:NSInvalidArgumentException
--- a/platform/darwin/src/MGLShape.mm
+++ b/platform/darwin/src/MGLShape.mm
@@ -94,9 +94,7 @@ bool operator==(const CLLocationCoordinate2D lhs, const CLLocationCoordinate2D r
 
 - (NSUInteger)hash
 {
-    NSUInteger hash;
-    hash += _title.hash;
-    hash += _subtitle.hash;
+    NSUInteger hash = _title.hash + _subtitle.hash;
 #if !TARGET_OS_IPHONE
     hash += _toolTip.hash;
 #endif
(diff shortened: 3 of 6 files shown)


platform/darwin/src/MGLNetworkConfiguration.m
```diff
@@ -12,9 +11,6 @@ + (void)load {
 }
 
 + (instancetype)sharedManager {
-    if (NSProcessInfo.processInfo.mgl_isInterfaceBuilderDesignablesAgent) {
```
1ec5: `+[MGLAccountManager sharedManager]` bails in IB because MGLAccountManager spins up lots of other machinery, including MGLMapboxEvents.

platform/ios/src/MGLAnnotationImage.m
```diff
@@ -57,15 +57,11 @@ - (BOOL)isEqual:(id)other {
     
     return ((!_reuseIdentifier && !otherAnnotationImage.reuseIdentifier) || [_reuseIdentifier isEqualToString:otherAnnotationImage.reuseIdentifier])
     && _enabled == otherAnnotationImage.enabled
-    && ((!_image && !otherAnnotationImage.image) || [UIImagePNGRepresentation(_image) isEqualToData:UIImagePNGRepresentation(otherAnnotationImage.image)]);
```
1ec5: @frederoni, any reason why comparing two UIImages for object equality isn’t good enough?

platform/darwin/src/MGLRasterSource.mm
```diff
@@ -51,7 +51,7 @@ - (instancetype)initWithIdentifier:(NSString *)identifier tileURLTemplates:(NS_A
     if (self = [super initWithIdentifier:identifier tileURLTemplates:tileURLTemplates options:options]) {
         mbgl::Tileset tileSet = MGLTileSetFromTileURLTemplates(tileURLTemplates, options);
         
-        uint16_t tileSize;
```
1ec5: I had somehow forgotten that primitive-typed stack variables are uninitialized in Objective-C just as they are in C. ARC nils out only object pointers on the stack.

platform/ios/src/MGLAnnotationImage.m
1ec5: I see that `testAnnotationImage()` is failing as a result of my change. It’s true that drawing the image to a new context could make the two images unequal, but there are a number of reasons that would be the case anyways (such as different color profiles). In general, I don’t think a class’s `-isEqual:` implementation should be more thorough than the `-isEqual:` implementations of its ivars.

platform/darwin/src/MGLShape.mm
```diff
@@ -94,9 +94,7 @@ - (BOOL)isEqual:(id)other
 
 - (NSUInteger)hash
 {
-    NSUInteger hash;
-    hash += _title.hash;
-    hash += _subtitle.hash;
+    NSUInteger hash = _title.hash + _subtitle.hash;
```
kkaefer: Is this a safe way to compute the hash?

frederoni: A XOR over the values could improve it but I'm ok with it because this is an abstract class and can't be used directly. All concrete subclasses include the hash of the GeoJSON dictionary.

1ec5: his is a fairly conventional implementation, as far as I know. The only hard requirement for a `-hash` implementation is that two equal objects (as in `-isEqual:`) must have the same hash. But feel free to improve on the implementation in a separate PR.

PR #7671, opened 2017-01-11
Fix runtime styling static analyzer warnings

boundsj:
This fixes warnings introduced over the course of the runtime styling implementation.

cc @1ec5 @frederoni

### DIFF
--- a/platform/darwin/src/MGLBackgroundStyleLayer.mm
+++ b/platform/darwin/src/MGLBackgroundStyleLayer.mm
@@ -92,7 +92,7 @@
 - (MGLStyleValue<MGLColor *> *)backgroundColor {
     MGLAssertStyleLayerIsValid();
 
-    auto propertyValue = self.rawLayer->getBackgroundColor() ?: self.rawLayer->getDefaultBackgroundColor();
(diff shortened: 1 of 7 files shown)


platform/darwin/src/MGLBackgroundStyleLayer.mm
```diff
@@ -92,7 +92,7 @@ - (void)setBackgroundColor:(MGLStyleValue<MGLColor *> *)backgroundColor {
 - (MGLStyleValue<MGLColor *> *)backgroundColor {
     MGLAssertStyleLayerIsValid();
 
-    auto propertyValue = self.rawLayer->getBackgroundColor() ?: self.rawLayer->getDefaultBackgroundColor();
+    static const auto propertyValue = self.rawLayer->getBackgroundColor() ?: self.rawLayer->getDefaultBackgroundColor();
```
1ec5: Can you explain why `static` is needed here?

boundsj: Well that doesn't make much sense does it? It stops the warning but causes other more serious issues, obviously. I'll keep looking.

boundsj:
I don't think this PR is the answer 😄  (https://github.com/mapbox/mapbox-gl-native/issues/7668#issuecomment-271780465). Closing

Issue #7578, opened 2017-01-03
3.3.7 Crash in MGLMapboxEvents.m locationEventWithAttributes:


erichoracek:
Hi,

We've received this crash report via HockeyApp for Mapbox 3.3.7:

**Platform: iOS 10.1.1/10.2**
**Mapbox SDK version: 3.3.7**

```
Date/Time:       2016-12-27T20:49:48Z
Launch Time:     2016-12-27T17:53:55Z
OS Version:      iPhone OS 10.2 (14C92)
Report Version:  104

Exception Type:  SIGABRT
Exception Codes: #0 at 0x187a77014
Crashed Thread:  0

Application Specific Information:
*** Terminating app due to uncaught exception 'NSInvalidArgumentException', reason: '*** -[__NSPlaceholderDictionary initWithObjects:forKeys:count:]: attempt to insert nil object from objects[3]'

Last Exception Backtrace:
0   CoreFoundation                       0x0000000188aa91b8 __exceptionPreprocess + 124
1   libobjc.A.dylib                      0x00000001874e055c objc_exception_throw + 52
2   CoreFoundation                       0x000000018898ceac -[__NSPlaceholderDictionary initWithObjects:forKeys:count:] + 360
3   CoreFoundation                       0x000000018898cd1c +[NSDictionary dictionaryWithObjects:forKeys:count:] + 60
4   Mapbox                               0x000000010133e04c -[MGLMapboxEvents locationEventWithAttributes:] (MGLMapboxEvents.m:440)
5   Mapbox                               0x000000010133df0c -[MGLMapboxEvents fullyFormedEventForEvent:withAttributes:] (MGLMapboxEvents.m:434)
6   Mapbox                               0x000000010133dc50 -[MGLMapboxEvents pushEvent:withAttributes:] (MGLMapboxEvents.m:406)
7   Mapbox                               0x000000010133db80 +[MGLMapboxEvents pushEvent:withAttributes:] (MGLMapboxEvents.m:390)
8   Mapbox                               0x000000010133fa44 -[MGLMapboxEvents locationManager:didUpdateLocations:] (MGLMapboxEvents.m:669)
9   Mapbox                               0x00000001013419dc -[MGLLocationManager locationManager:didUpdateLocations:] (MGLLocationManager.m:151)
10  CoreLocation                         0x0000000190aaca38 CLClientGetCapabilities + 35644
[...]

PR #7672, opened 2017-01-11
Flush telemetry cache when data collection is paused

boundsj:
This flushes the cache just before pausing so that data collected before the pause is not thrown away. It also refactors event creation to avoid possible nil value inserts into dictionaries
and adds a small optimization to avoid creating immutable dictionaries to help create mutable ones.

Fixes https://github.com/mapbox/mapbox-gl-native/issues/7578

### DIFF
--- a/platform/ios/src/MGLMapboxEvents.m
+++ b/platform/ios/src/MGLMapboxEvents.m
@@ -434,21 +434,15 @@ const NSTimeInterval MGLFlushInterval = 180;
     return nil;
 }
 
-- (void)updateDictionary:(NSMutableDictionary *)dictionary withKey:(NSString *)key value:(id)value {
-    if (key && value) {
-        [dictionary setValue:value forKey:key];
-    }
-}
-
 - (MGLMapboxEventAttributes *)locationEventWithAttributes:(MGLMapboxEventAttributes *)attributeDictionary {
     MGLMutableMapboxEventAttributes *attributes = [NSMutableDictionary dictionary];
-    [self updateDictionary:attributes withKey:MGLEventKeyEvent value:MGLEventTypeLocation];
-    [self updateDictionary:attributes withKey:MGLEventKeySource value:MGLEventSource];
-    [self updateDictionary:attributes withKey:MGLEventKeySessionId value:self.instanceID];
-    [self updateDictionary:attributes withKey:MGLEventKeyOperatingSystem value:self.data.iOSVersion];
+    attributes[MGLEventKeyEvent] = MGLEventTypeLocation;
+    attributes[MGLEventKeySource] = MGLEventSource;
+    attributes[MGLEventKeySessionId] = self.instanceID;
+    attributes[MGLEventKeyOperatingSystem] = self.data.iOSVersion;
     NSString *currentApplicationState = [self applicationState];
     if (![currentApplicationState isEqualToString:MGLApplicationStateUnknown]) {
-        [self updateDictionary:attributes withKey:MGLEventKeyApplicationState value:currentApplicationState];
+        attributes[MGLEventKeyApplicationState] = currentApplicationState;
     }
 
     return [self eventForAttributes:attributes attributeDictionary:attributeDictionary];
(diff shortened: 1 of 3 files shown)


platform/ios/src/MGLMapboxEvents.m
```diff
@@ -433,51 +434,63 @@ - (MGLMapboxEventAttributes *)fullyFormedEventForEvent:(NSString *)event withAtt
...
+
 - (MGLMapboxEventAttributes *)locationEventWithAttributes:(MGLMapboxEventAttributes *)attributeDictionary {
-    MGLMutableMapboxEventAttributes *attributes = [@{MGLEventKeyEvent: MGLEventTypeLocation,
-                                                     MGLEventKeySource: MGLEventSource,
-                                                     MGLEventKeySessionId: self.instanceID,
-                                                     MGLEventKeyOperatingSystem: self.data.iOSVersion} mutableCopy];
+    MGLMutableMapboxEventAttributes *attributes = [NSMutableDictionary dictionary];
+    [self updateDictionary:attributes withKey:MGLEventKeyEvent value:MGLEventTypeLocation];
+    [self updateDictionary:attributes withKey:MGLEventKeySource value:MGLEventSource];
+    [self updateDictionary:attributes withKey:MGLEventKeySessionId value:self.instanceID];
```
1ec5: Since this is already a mutable dictionary and the keys are constant, it would be just as easy to say e.g. `attributes[MGLEventKeySessionId] = self.instanceID`.

boundsj: @1ec5 it would be, but I'd like to run everything through the `updateDictionary:...` method (even if it is not always necessary) for consistency. Some of the values could be nil (either as a matter of course or because of bugs). The utility method helps ensure that we don't crash because of such a mistake. Always using the utility method instead of a mix of the utility method and `dictionary[key] = value` is easier to read, I think.

boundsj:
https://github.com/mapbox/mapbox-gl-native/pull/7672/commits/2ca223129ca4242862f58b932433d76fe009cf88 is intended to address https://github.com/mapbox/mapbox-gl-native/issues/7679. Maintaining the array of data tasks has proven to be too problematic especially given that we don't use the API it was written for.

cc @friedbunny

platform/ios/src/MGLMapboxEvents.m
```diff
@@ -433,51 +434,63 @@ - (MGLMapboxEventAttributes *)fullyFormedEventForEvent:(NSString *)event withAtt
...
+    [self updateDictionary:attributes withKey:MGLEventKeyEvent value:MGLEventTypeMapDragEnd];
+
     return [self eventForAttributes:attributes attributeDictionary:attributeDictionary];
 }
 
 - (MGLMutableMapboxEventAttributes *)interactionEvent {
-    MGLMutableMapboxEventAttributes *attributes = [@{MGLEventKeyCreated: [self.rfc3339DateFormatter stringFromDate:[NSDate date]],
-                                                     MGLEventKeyOrientation: [self deviceOrientation],
-                                                     MGLEventKeyWifi: @([[MGLReachability reachabilityForLocalWiFi] isReachableViaWiFi])} mutableCopy];
-    [self addBatteryStateToAttributes:attributes];
```
friedbunny: It looks like battery state is no longer added to the attributes and `addBatteryStateToAttributes:` hasn’t been updated to use `updateDictionary:withKey:value:`.

platform/ios/src/MGLMapboxEvents.m
```diff
@@ -433,51 +434,63 @@ - (MGLMapboxEventAttributes *)fullyFormedEventForEvent:(NSString *)event withAtt
...
-    MGLMutableMapboxEventAttributes *attributes = [@{MGLEventKeyEvent: MGLEventTypeLocation,
-                                                     MGLEventKeySource: MGLEventSource,
-                                                     MGLEventKeySessionId: self.instanceID,
-                                                     MGLEventKeyOperatingSystem: self.data.iOSVersion} mutableCopy];
+    MGLMutableMapboxEventAttributes *attributes = [NSMutableDictionary dictionary];
+    [self updateDictionary:attributes withKey:MGLEventKeyEvent value:MGLEventTypeLocation];
+    [self updateDictionary:attributes withKey:MGLEventKeySource value:MGLEventSource];
+    [self updateDictionary:attributes withKey:MGLEventKeySessionId value:self.instanceID];
+    [self updateDictionary:attributes withKey:MGLEventKeyOperatingSystem value:self.data.iOSVersion];
     [self addApplicationStateToAttributes:attributes];
```
friedbunny: If we intend to consistently use `updateDictionary:...` everywhere, `addApplicationStateToAttributes:` hasn’t been updated.

platform/ios/src/MGLMapboxEvents.m
boundsj: Battery monitoring was effectively removed in https://github.com/mapbox/mapbox-gl-native/pull/7212/commits/da6e48fd98c2d2e3a6c263ade1da46f7dc622b1d#diff-6e8e31cd0ffe52aeda9c8b2671bcea19L195. This PR now removes the last of it in https://github.com/mapbox/mapbox-gl-native/pull/7672/commits/c705f17ac1d151a902da740f856d846b475b0360

platform/ios/src/MGLMapboxEvents.m
1ec5: `attributes[key] = nil` is perfectly fine. There’s no need for this utility method.

boundsj: Ah right. I was conflating `-[NSMutableDictionary setObject:forKey:]` (which does not accept nil) and `attributes[key] = nil` which removes the key (which is what we want). I'll update this.

Issue #6426, opened 2016-09-22
[core] Remove util::Thread from DefaultFileSource


tmpsantos:
Part of #6425 effort to migrate to the Actor model.

tmpsantos:
Fixed by #6072

PR #7678, opened 2017-01-11
Convert DefaultFileSource and dependents to actor model

kkaefer:
Getting started on https://github.com/mapbox/mapbox-gl-native/issues/6426. This is still a work in progress.

A few large changes:
- the `DefaultFileSource` constructor now takes a `Scheduler`
- the `AssetFileSource` does not use its own thread anymore; instead it uses the shared worker threads

Given that our file requests need a way to "cancel" them (i.e. have the callback not invoked, I introduced `AssetFileSource::Impl::Request`, which holds a `shared_ptr` to the callback. If that object gets destroyed, the `weak_ptr` we're passing to the worker thread will be empty, and back on the main thread, the callback obviously can't be fired anymore. I'm not sure if using this violates @jfirebaugh's plea about not using `std::shared_ptr` with actors (since technically, you could try to lock the `weak_ptr` in the worker.

I also added a `MGLThreadPool` object for Darwin, but implemented a global singleton for Android . I believe we should go the global singleton instance for Darwin as well (and remove `MGLThreadPool` again), since Darwin never deallocates the `DefaultFileSource` as part of the regular application workflow anyway (it's a singleton in `MGLOfflineStorage`).

### DIFF
--- a/platform/darwin/src/MGLThreadPool.mm
+++ b/platform/darwin/src/MGLThreadPool.mm
@@ -1,0 +1,39 @@
+#import "MGLThreadPool_Private.h"
+
+#import "NSProcessInfo+MGLAdditions.h"
+
+@interface MGLThreadPool ()
+
+@property (nonatomic) mbgl::ThreadPool *mbglThreadPool;
+
+@end
+
+@implementation MGLThreadPool
+
+#pragma mark - Internal
+
++ (instancetype)sharedPool {
+    if (NSProcessInfo.processInfo.mgl_isInterfaceBuilderDesignablesAgent) {
+        return nil;
+    }
+    static dispatch_once_t onceToken;
+    static MGLThreadPool *sharedThreadPool;
+    dispatch_once(&onceToken, ^{
+        sharedThreadPool = [[self alloc] init];
+    });
+    return sharedThreadPool;
+}
+
+- (instancetype)init {
+    if (self = [super init]) {
+        _mbglThreadPool = new mbgl::ThreadPool(4, { "Worker" });
+    }
+    return self;
+}
+
+- (void)dealloc {
+    delete _mbglThreadPool;
+    _mbglThreadPool = nullptr;
+}
+
+@end
--- a/platform/darwin/src/MGLThreadPool_Private.h
+++ b/platform/darwin/src/MGLThreadPool_Private.h
@@ -1,0 +1,18 @@
+#import <Foundation/Foundation.h>
+
+#include <mbgl/util/default_thread_pool.hpp>
+
+@interface MGLThreadPool : NSObject
+@end
+
+@interface MGLThreadPool (Private)
+
+/// Returns the shared instance of the `MGLThreadPool` class.
++ (instancetype)sharedPool;
+
+/**
+ The shared thread pool object owned by the shared thread pool object.
+ */
+@property (nonatomic) mbgl::ThreadPool *mbglThreadPool;
+
+@end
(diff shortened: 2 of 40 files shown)


kkaefer:
The long term goal is to move the other FileSources onto the actor model as well, which enables us to remove `util::Thread`.

tmpsantos:
I gave this a shot once and I got stuck. The main problem I faced when removing `Thread` from the `DefaultFileSource` was that we need a `RunLoop` for some platforms on the worker thread for pooling for events. Create a new one for each new request is probably no viable. Also the `HTTPFileSource::Impl` is usually not thread safe and lives in a fixed thread.

We probably need a new scheduler that is not `ThreadPool` for the `DefaultFileSource` at least in some platforms we support.

platform/darwin/src/MGLThreadPool.mm
```diff
@@ -0,0 +1,39 @@
...
+@property (nonatomic) mbgl::ThreadPool *mbglThreadPool;
+
+@end
+
+@implementation MGLThreadPool
+
+#pragma mark - Internal
+
++ (instancetype)sharedPool {
+    if (NSProcessInfo.processInfo.mgl_isInterfaceBuilderDesignablesAgent) {
```
1ec5: This is unnecessary, since there won't be an offline storage object if we're rendering for IB, and even if we were, spinning up this thread pool should be cheap. The reason we do this in MGLOfflineStorage and MGLAccountManager is to avoid potentially expensive operations that may block MGLMapView from rendering within the allotted time.

platform/darwin/src/MGLThreadPool_Private.h
```diff
@@ -0,0 +1,18 @@
+#import <Foundation/Foundation.h>
+
+#include <mbgl/util/default_thread_pool.hpp>
+
+@interface MGLThreadPool : NSObject
+@end
+
+@interface MGLThreadPool (Private)
```
1ec5: No need for a class extension, since the whole header is already internal to the project. A class extension would need to go in a separate header from the public header anyways.

kkaefer: Thanks; as per the comment above, I'm planning on removing `MGLThreadPool` altogether, since it's a singleton anyway.

kkaefer:
@jfirebaugh given that you're most familiar with this system, could you please do a quick review of whether I'm moving into the right direction with this?

jfirebaugh:
The way I would model this is to make the request and response callback both first-class actors:

```c++
class RequestCallback {
public:
    // This class exists only to accept and ignore the "self" argument. Maybe it should be optional?
    RequestCallback(ActorRef<RequestCallback>, std::function<void (Response)>);
    void operator() (Response response) { callback(response); }

private:
    std::function<void (Response)> callback;
};

class AssetRequest {
public:
    AssetRequest(ActorRef<AssetRequest>);
    ~AssetRequest(); // cancels

    void send(const Resource& resource, ActorRef<RequestCallback> callback) {
        // do whatever, then eventually:
        callback.invoke(&RequestCallback::operator(), response);
    }
};

class SomeClass {
public:
    SomeClass(Scheduler& requestScheduler)
      : request(requestScheduler) {}

private:
    void requestingMethod() {
        request.invoke(&AssetRequest::send, resource, callback);
    }

    void receivingMethod(Response);

    Actor<AssetRequest> request;
    Actor<RequestCallback> callback {
        *util::RunLoop::Get(),
        std::bind(&SomeClass::receivingMethod, this)
    };
};
```

jfirebaugh:
Stale.

PR #7685, opened 2017-01-11
[Qt] Implement mapbox::sqlite::{Database,Statement} using QtSql

brunoabinader:
Updates mapbox::sqlite API (from `platform/default/sqlite3.hpp`) to be platform-agnostic, providing means of having a Qt-only implementation that doesn't directly depend on sqlite3 headers/libraries.

### DIFF
--- a/cmake/test-files.cmake
+++ b/cmake/test-files.cmake
@@ -68,6 +68,7 @@ set(MBGL_TEST_FILES
     test/storage/offline_download.test.cpp
     test/storage/online_file_source.test.cpp
     test/storage/resource.test.cpp
+    test/storage/sqlite.test.cpp
 
     # style/conversion
     test/style/conversion/function.test.cpp
--- a/platform/default/sqlite3.hpp
+++ b/platform/default/sqlite3.hpp
@@ -46,9 +46,6 @@ public:
     void exec(const std::string &sql);
     Statement prepare(const char *query);
 
-    int64_t lastInsertRowid() const;
-    uint64_t changes() const;
-
 private:
     std::unique_ptr<DatabaseImpl> impl;
 
@@ -85,6 +82,9 @@ public:
     bool run();
     void reset();
     void clearBindings();
+
+    int64_t lastInsertRowId() const;
+    uint64_t changes() const;
 
 private:
     std::unique_ptr<StatementImpl> impl;
--- a/platform/default/sqlite3.hpp
+++ b/platform/default/sqlite3.hpp
@@ -20,9 +20,15 @@ enum OpenFlag : int {
 };
 
 struct Exception : std::runtime_error {
+    enum Code : int {
+        OK = 0,
+        CANTOPEN = 14,
+        NOTADB = 26
+    };
+
     Exception(int err, const char *msg) : std::runtime_error(msg), code(err) {}
     Exception(int err, const std::string& msg) : std::runtime_error(msg), code(err) {}
-    const int code = 0;
+    const int code = OK;
 };
 
 class DatabaseImpl;
--- a/platform/default/sqlite3.hpp
+++ b/platform/default/sqlite3.hpp
@@ -57,8 +57,6 @@ private:
     Statement(const Statement &) = delete;
     Statement &operator=(const Statement &) = delete;
 
-    void check(int err);
-
 public:
     Statement(Database *db, const char *sql);
     Statement(Statement &&);
--- a/platform/default/sqlite3.hpp
+++ b/platform/default/sqlite3.hpp
@@ -4,9 +4,7 @@
 #include <vector>
 #include <stdexcept>
 #include <chrono>
-
-typedef struct sqlite3 sqlite3;
-typedef struct sqlite3_stmt sqlite3_stmt;
+#include <memory>
 
 namespace mapbox {
 namespace sqlite {
@@ -27,7 +25,9 @@ struct Exception : std::runtime_error {
     const int code = 0;
 };
 
+class DatabaseImpl;
 class Statement;
+class StatementImpl;
 
 class Database {
 private:
@@ -42,7 +42,6 @@ public:
 
     explicit operator bool() const;
 
-    static void errorLogCallback(void *arg, const int err, const char *msg);
     void setBusyTimeout(std::chrono::milliseconds);
     void exec(const std::string &sql);
     Statement prepare(const char *query);
@@ -51,7 +50,9 @@ public:
     uint64_t changes() const;
 
 private:
-    sqlite3 *db = nullptr;
+    std::unique_ptr<DatabaseImpl> impl;
+
+    friend class Statement;
 };
 
 class Statement {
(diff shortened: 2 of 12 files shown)


platform/default/sqlite3.hpp
```diff
@@ -42,27 +48,23 @@ class Database {
 
     explicit operator bool() const;
 
-    static void errorLogCallback(void *arg, const int err, const char *msg);
     void setBusyTimeout(std::chrono::milliseconds);
     void exec(const std::string &sql);
     Statement prepare(const char *query);
 
-    int64_t lastInsertRowid() const;
```
jfirebaugh: As methods on `Statement`, `lastInsertRowid` and `changes` invite misuse in this implementation:

```
Statement stmt1(...);
Statement stmt2(...);
stmt1.execute();
stmt2.execute();
stmt1.lastInsertRowid(); // Would actually return row ID from stmt2
```

We need to either keep them as `Database` methods, or cache the values in `Statement` after every execution so that code like the above behaves as expected. Also, we should validate that the Qt implementation do so as well.

brunoabinader: Good point - we can add a unit test for verifying that. It's unfortunate that `QSqlDatabase` does not provide direct access for these - so for keeping them in `Database` we'll need a temporary `QSqlQuery` object that makes the calls based on that db.

brunoabinader: > for keeping them in Database we'll need a temporary QSqlQuery object that makes the calls based on that db.

This idea won't work as expected - `QSqlQuery::lastInsertId` expects an active QSqlQuery and it wouldn't change the fact that the result would still be potentially overwritten by another statement.

> cache the values in Statement after every execution

This sounds like the way to go - we can cache last insert row ID and the number of rows affected on every `Statement::run` - which can also be done for the default sqlite3 implementation.

jfirebaugh: > This sounds like the way to go

Agreed. My hunch is that this is what Qt does internally for its own SQLite implementation. Can you verify that? If so, we only need to do it ourselves for the non-Qt implementation.

brunoabinader: Qt SQLite driver does not cache that - [it calls for `sqlite3_changes` and `sqlite3_last_insert_rowid` directly](https://github.com/qt/qtbase/blob/1dcd61d553e25c8986e7e6e65efacc51c3cf3bee/src/plugins/sqldrivers/sqlite/qsql_sqlite.cpp#L519-L534).

> we only need to do it ourselves for the non-Qt implementation.

Yep 👍 Already have a working solution for the non-Qt implementation - will update this PR once I get everything in order.

brunoabinader: This is now done in 793947bd5adc99a349059f8d812cad2c2e9a5fe9.

cmake/test-files.cmake
```diff
@@ -43,9 +43,6 @@ set(MBGL_TEST_FILES
     test/sprite/sprite_image.test.cpp
     test/sprite/sprite_parser.test.cpp
 
-    # src
-    test/src/main.cpp
```
tmpsantos: This file is generated, you have to change the script too to make sure it won't be added again.

brunoabinader: Fixed in #7769.

platform/default/sqlite3.hpp
jfirebaugh: > Qt SQLite driver does not cache that - [it calls for `sqlite3_changes` and `sqlite3_last_insert_rowid` directly](https://github.com/qt/qtbase/blob/1dcd61d553e25c8986e7e6e65efacc51c3cf3bee/src/plugins/sqldrivers/sqlite/qsql_sqlite.cpp#L519-L534).

Interesting -- so Qt will have the unexpected behavior if you have multiple statements and don't call `numRowsAffected()` / `lastInsertId()` immediately after executing each one?

platform/default/sqlite3.hpp
```diff
@@ -85,8 +87,11 @@ class Statement {
     void reset();
     void clearBindings();
 
+    int64_t lastInsertRowId() const;
+    int64_t changes() const;
```
jfirebaugh: Can we keep the `changes` return value unsigned?

test/storage/offline_database.test.cpp
```diff
@@ -80,6 +80,25 @@ class FileLock {
 
 } // namespace
 
+TEST(OfflineDatabase, Statement) {
```
jfirebaugh: Let's make a new test file for generic `mapbox::sqlite` tests.

platform/default/sqlite3.hpp
```diff
@@ -22,12 +20,20 @@ enum OpenFlag : int {
 };
 
 struct Exception : std::runtime_error {
+    enum Code : int {
+        OK = 0,
```
tmpsantos: `static_assert` to make sure we track sqlite.h?

platform/qt/src/sqlite3.cpp
```diff
@@ -0,0 +1,395 @@
...
+}
+
+template <>
+void Statement::bind(int offset, mbgl::Timestamp value) {
+    bind(offset, std::chrono::system_clock::to_time_t(value));
+}
+
+template <>
+void Statement::bind(int offset, optional<std::string> value) {
+    if (value)
```
tmpsantos: Style nit, use brackets.

platform/qt/src/sqlite3.cpp
```diff
@@ -0,0 +1,395 @@
...
+        bind(offset, *value);
+    else
+        bind(offset, nullptr);
+}
+
+template <>
+void Statement::bind(int offset, optional<mbgl::Timestamp> value) {
+    if (value)
```
tmpsantos: Ditto.

platform/default/sqlite3.hpp
brunoabinader: > so Qt will have the unexpected behavior if you have multiple statements and don't call numRowsAffected() / lastInsertId() immediately after executing each one?

Qt checks the state of each statement prior to calling these - a statement there is considered _active_ after it has been exec()'d. From my tests, though, two different statements, after being executed, do not deactivate each other, so I guess the behavior is unexpected indeed.

platform/default/sqlite3.hpp
brunoabinader: We can, though my rationale for moving it back to a signed value is because the signature for [sqlite3_changes](https://www.sqlite.org/c3ref/changes.html) returns an `int`, which is also mirrored in [QSqlQuery::numRowsChanged](http://doc.qt.io/qt-5/qsqlquery.html#numRowsAffected) - the latter even uses `-1` to report an error state e.g. when the statement is not _active_. My concern is that mixing signed and unsigned return values is prone to integer overflow bugs.

brunoabinader: We can alleviate that by checking if the return value is less than 0, and returning 0 on that case - this way we can keep using unsigned 👍

brunoabinader:
@jfirebaugh can you please review the core-related changes?

PR #7689, opened 2017-01-11
Restore GL_KHR_debug messages in release builds

### LINKED ISSUE #6689: Restore gl::debugging::enable() in release mode
8c4da72d09e585bd10d030129c99e1e609a2d2f9 disabled OpenGL `GL_KHR_debug`-type messages in release mode. But we want to keep these -- they may help diagnose issues like #5731. We should revert this change.

cc @kkaefer

jfirebaugh:
Fixes #6689.

### DIFF
--- a/src/mbgl/gl/debugging.cpp
+++ b/src/mbgl/gl/debugging.cpp
@@ -1,4 +1,3 @@
-#ifndef NDEBUG
 #include <mbgl/gl/debugging.hpp>
 #include <mbgl/gl/gl.hpp>
 #include <mbgl/gl/extension.hpp>
@@ -190,5 +189,3 @@ group::~group() {
 } // namespace debugging
 } // namespace gl
 } // namespace mbgl
-
-#endif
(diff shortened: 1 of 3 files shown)


src/mbgl/gl/debugging.cpp
```diff
@@ -190,5 +189,3 @@ group::~group() {
 } // namespace debugging
 } // namespace gl
 } // namespace mbgl
-
-#endif
```
kkaefer: Removing those include guards means that the symbol `mbgl::gl::debugging::group` is defined (but not declared), while we shouldn't even declare it when `NDEBUG` is set.

It also means that we attempt to load the extensions `GL_KHR_debug` and `GL_EXT_debug_marker` even though we are not using them.

jfirebaugh: Fixed! (We do still try to load the `GL_KHR_debug` extension, but that's the intent of this PR.)

PR #7691, opened 2017-01-12
[linux] No need for CXXABI suffix for mesa

jfirebaugh:


jfirebaugh:
> Clang is buggy

Can you say more? What's the bug? Is it filed in the llvm issue tracker?

brunoabinader:
@jfirebaugh I won't have access to my Linux desktop this week, I'll post here more information on that whenever possible.

PR #7693, opened 2017-01-12
Make style-related tests more robust

1ec5:
This PR is an attempt to fix macOS test failures that occur only in CI while making iOS tests potentially more reliable as well.

Removed UIViewController from MGLStyleLayerTests on iOS, because we aren’t really testing integration at that level. Switched to delegate-based expectation fulfillment for both MGLStyleLayerTests and MGLStyleTests, conditionalized on whether the style has already loaded, to avoid race conditions that bleed into other tests.

/cc @friedbunny


1ec5:
I reran the macOS build multiple times over the course of the night and it passed consistently. Merging.

Issue #6947, opened 2016-11-07
iOS CLI build fails during code signing


kkaefer:
I'm trying to run `make ipackage` or `make iframework`, but getting this:

```
mkdir -p build/ios
(cd build/ios && cmake -G Xcode ../.. \
        -DCMAKE_TOOLCHAIN_FILE=../../platform/ios/toolchain.cmake \
        -DMBGL_PLATFORM=ios)
-- Configuring done
-- Generating done
-- Build files have been written to: /Users/kkaefer/Code/gl/native/build/ios
FORMAT= BUILD_DEVICE= SYMBOLS= \
    ./platform/ios/scripts/package.sh
Configuring dynamic and static Release framework for iphoneos; symbols: YES; self-contained static framework: NO
* Recording library version…
mapbox-gl-native b32ba799e0
* Building targets (build 9259, version 3.4.0-beta.2)…
▸ Building ios/settings [Release]
▸ Check Dependencies
▸ Copying Root.strings
▸ Copying Root.plist
▸ Touching Settings.bundle
▸ Signing /Users/kkaefer/Code/gl/native/build/ios/Release-iphonesimulator/Settings.bundle
▸ Building ios/bundle [Release]
▸ Check Dependencies
▸ Processing Info-static.plist
▸ Copying Localizable.strings
▸ Check Dependencies
▸ Running script 'CMake Rules'
** BUILD FAILED **


The following build commands failed:
    CodeSign /Users/kkaefer/Code/gl/native/build/ios/Release-iphonesimulator/Settings.bundle
(1 failure)
make: *** [ipackage] Error 65
```

/cc @1ec5 @brunoabinader

JesseCrocker:
I ran into this same issue, and as a temporary workaround i ran `cp -r build/ios/Debug-iphoneos/Settings.bundle build/ios/Release-iphoneos/Settings.bundle` and was able to get a working framework.

1ec5:
#6948 looks like it’ll be a proper fix for the issue.

PR #7698, opened 2017-01-12
Silence dependency analysis warning

frederoni:
#6948 fixed a compile-time error that started with Xcode 8.1 but introduced a new warning in 8.2

screenshot 2017-01-12 14 42 28

This PR fixes that by specifying the correct Info.plist instead of the Root.plist that's bundled in settings. 

https://developer.apple.com/library/content/qa/qa1649/_index.html

Verified that the Settings.bundle is still working:
screenshot 2017-01-12 15 16 25

### DIFF
--- a/platform/ios/ios.xcodeproj/project.pbxproj
+++ b/platform/ios/ios.xcodeproj/project.pbxproj
@@ -115,6 +115,7 @@
 		357579871D502AFE000B822E /* MGLLineStyleLayerTests.mm in Sources */ = {isa = PBXBuildFile; fileRef = 357579861D502AFE000B822E /* MGLLineStyleLayerTests.mm */; };
 		357579891D502B06000B822E /* MGLCircleStyleLayerTests.mm in Sources */ = {isa = PBXBuildFile; fileRef = 357579881D502B06000B822E /* MGLCircleStyleLayerTests.mm */; };
 		3575798B1D502B0C000B822E /* MGLBackgroundStyleLayerTests.mm in Sources */ = {isa = PBXBuildFile; fileRef = 3575798A1D502B0C000B822E /* MGLBackgroundStyleLayerTests.mm */; };
+		357FB1FC1F7A564700E9A22E /* Root.plist in Resources */ = {isa = PBXBuildFile; fileRef = DA25D5BF1CCD9F8400607828 /* Root.plist */; };
 		357FE2DD1E02D2B20068B753 /* NSCoder+MGLAdditions.h in Headers */ = {isa = PBXBuildFile; fileRef = 357FE2DB1E02D2B20068B753 /* NSCoder+MGLAdditions.h */; };
 		357FE2DE1E02D2B20068B753 /* NSCoder+MGLAdditions.h in Headers */ = {isa = PBXBuildFile; fileRef = 357FE2DB1E02D2B20068B753 /* NSCoder+MGLAdditions.h */; };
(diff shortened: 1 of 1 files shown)


platform/ios/ios.xcodeproj/project.pbxproj
```diff
@@ -2349,7 +2349,7 @@
 		DA25D5BC1CCD9EDE00607828 /* Debug */ = {
 			isa = XCBuildConfiguration;
 			buildSettings = {
-				INFOPLIST_FILE = framework/Settings.bundle/Root.plist;
+				INFOPLIST_FILE = "framework/Info-static.plist";
```
1ec5: This build setting is part of the `settings` target, which produces Settings.bundle, not Mapbox.framework. Doesn’t this mean Settings.bundle will contain an Info.plist that describes the entire framework?

frederoni: It would. Same goes for the Bundle.bundle which the static target also depends on. (I copied this from the bundle target)
Is the Settings.bundle only meant to be used in iosapp?

1ec5: Settings.bundle is an example meant to be used in any iOS application. A Settings bundle isn’t supposed to have an Info.plist, though.

1ec5:
Moving to v3.4.1 because this issue doesn’t impact shipping code and may well be a regression in Xcode. (A settings bundle isn’t supposed to have an Info.plist to begin with.)

1ec5:

friedbunny:
What’s the status of this, now that Xcode 8.3 is available?

1ec5:
This whole saga started with a code-signing error: #6947. Do we even need to code-sign the Settings.bundle if it’s going to be plopped into an application bundle at the end of the day?

frederoni:
This issue seems to be fixed in Xcode 8.3 so we can remove plist workaround.

frederoni:
Migrating the project to Xcode 9 also removes the plist from build phases which confirms that this is the right fix.

platform/ios/ios.xcodeproj/project.pbxproj
```diff
@@ -2110,7 +2109,6 @@
 			buildActionMask = 2147483647;
 			files = (
 				DA25D5C61CCDA06800607828 /* Root.strings in Resources */,
-				DA25D5C01CCD9F8400607828 /* Root.plist in Resources */,
```
1ec5: I’m confused by this change. We’re removing Root.plist from the Copy Files build phase, but it [remains the bundle’s Info plist](https://github.com/mapbox/mapbox-gl-native/blob/2031c2fb7175061005352f7ab66b1c2a93bac66e/platform/ios/ios.xcodeproj/project.pbxproj#L2720)?

frederoni: Which seems to be correct, no? It's what you get when you create a new framework from scratch. The plist is bundled in the framework but omitted from the Copy Files build phase.
 (https://developer.apple.com/library/content/documentation/MacOSX/Conceptual/BPFrameworks/Concepts/FrameworkAnatomy.html)

1ec5: But the file in the resulting bundle is called Info.plist, not Root.plist, right? Does the Settings screen show the correct content after installing this bundle?

frederoni: 🙈

friedbunny:
There appears to be no good way to get Xcode to accept-without-warning our non-standard, example-only Settings.bundle.

PR #7701, opened 2017-01-12
Document simpler types for style values in Swift

1ec5:
For numeric style attributes, the generic argument to MGLStyleValue is technically an NSNumber in Swift, but you don’t need to use NSNumber `*Value` properties with it:

```swift
layer.circleOpacity = MGLStyleValue(rawValue: 0.7)
let value = layer.circleOpacity as! MGLStyleConstantValue
assert(Float(value) == 0.7)
```

This PR replaces NSNumber properties with standard Swift types in the Swift column of type correspondence tables. I think we need to do this, because I see a lot of instances where developers unnecessarily write, for instance:

```swift
layer.circleOpacity = MGLStyleValue(rawValue: NSNumber(value: 0.7))
```

There’s a risk that the developer would read this and think they can write the following:

```swift
let opacity = layer.circleOpacity as! MGLStyleConstantValue
let percentage: Float = opacity
```

In which case the compiler will provide a fix-it that inserts a full-width conversion:

```swift
let opacity = layer.circleOpacity as! MGLStyleConstantValue
let percentage: Float = Float(opacity)
```

However, there is no fix-it for code like this:

```swift
let opacity = layer.circleOpacity as! MGLStyleConstantValue
let percentage = opacity * 100
```

I think getting into language-specific details is already slightly out of scope for these guides, but we do need to find a concise way to avoid misleading developers.

### DIFF
--- a/platform/darwin/docs/guides/For Style Authors.md.ejs
+++ b/platform/darwin/docs/guides/For Style Authors.md.ejs
@@ -269,9 +269,9 @@ In style JSON | In Objective-C        | In Swift
 Color         | `<%- cocoaPrefix %>Color` | `<%- cocoaPrefix %>Color`
 Enum          | `NSValue` (see `NSValue(MGLAdditions)`) | `NSValue` (see `NSValue(MGLAdditions)`)
 String        | `NSString`            | `String`
-Boolean       | `NSNumber.boolValue`  | `NSNumber.boolValue`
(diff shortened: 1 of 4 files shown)


friedbunny:
These changes look good and we should do them. 👍

Differences like these, though, make me that much more worried about the lack of objc snippets in our API docs. Runtime styling types are difficult enough, without having to think about language differences.

Issue #7704, opened 2017-01-12
MGLVectorStyleLayer.predicate cannot be set to nil


1ec5:
If you set `MGLVectorStyleLayer.predicate` to `nil`, then get the predicate, you get an NSComparisonPredicate with the bogus format string ` == nil`, whereas you should get `nil`.

/cc @boundsj

1ec5:
Fixed in #7705 on the release-ios-v3.4.0 branch.

PR #7708, opened 2017-01-13
Rewrite style layer tests

1ec5:
Replaced the style layer integration tests with unit tests. Until #7664, we weren’t really testing integration in these style layer test classes, anyways, since the added layers were getting blown away as soon as the style finished loading.

Test conversion of style values to property values and vice versa, not just round-tripping. Test the initial state and null-resetting of each null-resettable property. Test NSValue additions for style attribute enumerations. Test properties common to all style layer classes. Test MGLStyle’s source and layer collections.

Eviscerated implementations of unavailable style layer properties corresponding to style specification properties that were renamed. (This improves test coverage; this code can never run anyways.) Implemented corresponding getters to prevent ivars from being autosynthesized for these unavailable properties.

/cc @boundsj @friedbunny @frederoni


1ec5:
I put this issue on the v3.4.0 milestone because it doesn’t touch any shipping code (other than some dead code the compiler makes us put in) but makes us more confident about the shipping code (for example, leading me to find #7704). These tests can just as easily go into v3.4.1, however.

1ec5:
macOS SDK test coverage is at 29.16% overall, up from 27%:

macos

Relevant files compared to https://github.com/mapbox/mapbox-gl-native/pull/7548#issuecomment-269248581:

* MGLForegroundStyleLayer.m: 100%
* NSValue+MGLStyleAttributeAdditions.m: 100%
* NSPredicate+MGLAdditions.mm: 99.24%
* MGLStyleLayer.mm: 97.78% 📈
* NSCompoundPredicate+MGLAdditions.mm: 96.72%
* NSComparisonPredicate+MGLAdditions.mm: 96.72%
* MGLAttributionInfo.mm: 95.50%
* MGLFillStyleLayer.mm: 89.44% 📈
* MGLLineStyleLayer.mm: 87.14% 📈
* MGLShapeSource.mm: 85.82% 📈
* MGLSymbolStyleLayer.mm: 84.89% 📈
* MGLCircleStyleLayer.mm: 81.91% 📈
* MGLStyleValue_Private.h: 80.72% 📈
* MGLStyleValue.mm: 76.19% 📈
* MGLRasterStyleLayer.mm: 76.16% 📈
* MGLBackgroundStyleLayer.mm: 74.70% 📈
* MGLVectorSource: 72.55% 📈
* MGLTileSource.mm: 62.89%
* MGLRasterSource.mm: 61.97% 📈
* MGLSource.mm: 58.62%
* MGLStyle.mm: 53.54% 📈
* MGLVectorStyleLayer.mm: 47.06% 📉
* MGLOpenGLStyleLayer.mm: 39.47%
* NSExpression+MGLAdditions.mm: 27.13%
* MGLMapView.mm: 17.40% 📉

1ec5:
MGLRuntimeStylingHelper did make the tests more readable. However, I decided to replace it with more codegen so that `objCTestValue()` and `mbglTestValue()` always chooses a value other than the default. (I punted on doing that for some types, like `number` and `enum`.) That way we can be sure the property was actually set as opposed to being kept at the default.

platform/darwin/test/MGLStyleLayerTests.h
```diff
@@ -17,3 +15,11 @@
 @property (nonatomic, readonly, copy) NSString *lemma;
 
 @end
+
+@interface NSValue (MGLStyleLayerTestAdditions)
+
++ (instancetype)valueWithMGLVector:(CGVector)vector;
```
1ec5: macOS lacks `+[NSValue valueWithCGVector:]` and `NSValue.cgVectorValue` for some reason. It’s possible to call `+[NSValue value:withObjCType:]` and `-[NSValue getValue:]` instead, but that approach requires more than one statement at a time, so it doesn’t fit neatly into the assertion statements.

I’m considering upstreaming this category to the SDK proper, using `+[NSValue valueWithMGLCoordinate:]` and `NSValue.mglCoordinateValue` as a precedent.

Issue #6724, opened 2016-10-17
Add support for color interpolation


zugaldia:
https://github.com/mapbox/mapbox-gl-native/pull/6442 is introducing support for color interpolation in parity with GL JS' https://github.com/mapbox/mapbox-gl-js/pull/3245. We should support this on Android as part of the Runtime Style API.

This is the Android equivalent of https://github.com/mapbox/mapbox-gl-native/issues/6721 for iOS.

/cc: @ivovandongen @tmcw  @1ec5

tobrun:
The PR in OP was replaced with https://github.com/mapbox/mapbox-gl-native/issues/8720.

jfirebaugh:
We removed functions. If we bring back color spaces, it'll be as an expression.

PR #7711, opened 2017-01-13
refs #6779: mobile SDK style transition options

### LINKED ISSUE #6779: expose property transitions to platform SDKs
As in GL JS, we support paint property transitions (i.e animations over time) in core (see [here](https://github.com/mapbox/mapbox-gl-native/blob/7bbd49c72c0daa7036a575a49f6cf13bf8a7020e/src/mbgl/style/paint_property.hpp#L32)) but not yet in the platform SDKs. 

This is an enhancement to runtime styling, but not needed to ship that feature. 

/cc @1ec5 @mollymerp @peterqliu

incanus:
Starting to plumb the core transition options for the style out to iOS and eventually Android. 

Duration works fine, but delay causes a crash within `Interpolator<Color>` due to `assert` in the `Color` constructor failing, in my case due to negative RGBA values (going from gray to black buildings). Delay probably isn't factored in when continuing to subtract below zero. This may be a core bug. 

Fixes #6779. 

fade


platform/darwin/src/MGLStyle.h
```diff
@@ -427,6 +427,13 @@ MGL_EXPORT
  */
 - (void)removeImageForName:(NSString *)name;
 
+#pragma mark Managing a Style’s Transition Options
+
+- (void)setTransitionDuration:(NSTimeInterval)duration;
+- (NSTimeInterval)transitionDuration;
+- (void)setTransitionDelay:(NSTimeInterval)delay;
+- (NSTimeInterval)transitionDelay;
```
1ec5: These options should be exposed as properties.

platform/darwin/src/NSDate+MGLAdditions.h
```diff
@@ -7,4 +7,7 @@
 /// Converts from a duration in seconds to a duration object usable in mbgl.
 mbgl::Duration MGLDurationInSeconds(NSTimeInterval duration);
 
+/// Converts from an mbgl duration object to a duration in seconds.
+NSTimeInterval MGLSecondsFromDuration(mbgl::Duration duration);
```
1ec5: `MGLTimeIntervalFromDuration()` would make the return type clearer, presuming that an `NSTimeInterval` is always measured in seconds (which is the case).

platform/darwin/src/MGLStyle.mm
```diff
@@ -568,6 +568,34 @@ - (MGLImage *)imageForName:(NSString *)name
     return spriteImage ? [[MGLImage alloc] initWithMGLSpriteImage:spriteImage] : nil;
 }
 
+#pragma mark Style transitions
+
+- (void)setTransitionDuration:(NSTimeInterval)duration
+{
+    mbgl::style::TransitionOptions transitionOptions = self.mapView.mbglMap->getTransitionOptions();
```
1ec5: `auto` is typically used in situations like this to avoid repetition.

platform/darwin/src/NSDate+MGLAdditions.h
incanus: Then should we not also rename `MGLDurationInSeconds()` to `MGLDurationFromTimeInterval()`?

platform/darwin/src/MGLStyle.h
incanus: b2b0198e2b0f88b3c83e298329cb253650f38c98

platform/darwin/src/MGLStyle.mm
incanus: 705cb6b85a42895cb9e051e297362ebd9eab0aaa

incanus:
I'm going to bring this to Android too, for parity.

platform/darwin/src/NSDate+MGLAdditions.h
1ec5: `MGLDurationInSecondsFromTimeInterval()`, actually. `mbgl::Duration` can be expressed in many units; seconds is the one we’ve chosen to use for consistency with `NSTimeInterval`, but other parts of mbgl use milliseconds instead.

1ec5:
> delay causes a crash within `Interpolator<Color>` due to `assert` in the `Color` constructor failing, in my case due to negative RGBA values (going from gray to black buildings). Delay probably isn't factored in when continuing to subtract below zero. This may be a core bug. 

Can you file a separate issue about that? Depending on how systematic that issue is, we may want to hold off on a delay option until it’s fixed.

There’s an outstanding PR to support [realistic color interpolation](https://www.mapbox.com/mapbox-gl-style-spec/#function-colorSpace) in style properties: #6442. (#6724 and #6721 track the SDK frontend to this support.) Does the style specification need something similar for global transitions?

platform/darwin/src/NSDate+MGLAdditions.h
incanus: Right, took me a while to figure that out, but true. Do you think it's worth adding an assertion or other check here that it remains in seconds in this part of the code?

1ec5: No, I don’t think that’s an issue with `std::chrono::duration` or `mbgl::Duration` (which is actually in nanoseconds, my bad). `MGLDurationInSeconds()` performs a [`duration_cast`](http://en.cppreference.com/w/cpp/chrono/duration/duration_cast) to ensure that the output has the right quantity (if not the same unit as the input).

incanus:
> Can you file a separate issue about that? Depending on how systematic that issue is, we may want to hold off on a delay option until it’s fixed.

👉 https://github.com/mapbox/mapbox-gl-native/issues/7720

platform/darwin/src/NSDate+MGLAdditions.h
jfirebaugh: `mbgl::Duration` is similar to `NSTimeInterval` in the sense that the unit is fixed and built into the type. You can't have one `mbgl::Duration` representing a duration "in seconds" and another representing a duration "in milliseconds". The actual unit is defined by `std::chrono::steady_clock` and I'm not sure that it's invariant across platforms or standard library implementations. But it is fixed at compile time.

incanus:
Delay now works fine when rolling in the fix for #7720 from https://github.com/mapbox/mapbox-gl-native/issues/7756. 

Working on Android next.

incanus:
@cammace Any chance you could review the Android part of this? I'm also all 👂 when it comes to possible testing ideas. I checked out some other tests for animated components like camera changes, but from what I could see, none of them actually try a non-zero duration?

platform/darwin/src/NSDate+MGLAdditions.h
incanus: 6ddc2162adadbb302d4e79ed110f1b72e99d3f37

platform/ios/src/MGLMapView.mm
```diff
@@ -1226,7 +1226,7 @@ - (void)handlePanGesture:(UIPanGestureRecognizer *)pan
         if (drift)
         {
             CGPoint offset = CGPointMake(velocity.x * self.decelerationRate / 4, velocity.y * self.decelerationRate / 4);
-            _mbglMap->moveBy({ offset.x, offset.y }, MGLDurationInSeconds(self.decelerationRate));
+            _mbglMap->moveBy({ offset.x, offset.y }, MGLDurationInSecondsFromTimeInterval(self.decelerationRate));
```
1ec5: You’ll need to make corresponding changes to the macOS implementation of MGLMapView.

incanus: https://github.com/mapbox/mapbox-gl-native/pull/7711/commits/1fa75dd1d79fda380b6b912774ec2c57f6ac26b0

platform/darwin/src/MGLStyle.h
```diff
@@ -442,6 +442,18 @@ MGL_EXPORT
  */
 - (void)removeImageForName:(NSString *)name;
 
+#pragma mark Managing a Style’s Transition Options
+
+/**
+ Animation duration for style changes. Defaults to zero (instant change).
```
1ec5: What kind of changes can be animated? Just style layer attributes?

> The duration (in seconds) to animate any changes to layout or paint attributes in the style.
> 
> By default, this property is set to 0 seconds, so any changes take effect without animation.

platform/darwin/src/MGLStyle.h
```diff
@@ -442,6 +442,18 @@ MGL_EXPORT
...
+ */
+@property (nonatomic) NSTimeInterval transitionDuration;
+
+/**
+ Animation delay for style changes. Defaults to zero (instant change).
```
1ec5: > The amount of time (in seconds) to wait before animating any changes to layout or paint attributes in the style.
> 
> By default, this property is set to 0 seconds, so any changes begin to animate immediately.

platform/darwin/src/MGLStyle.h
```diff
@@ -442,6 +442,22 @@ MGL_EXPORT
  */
 - (void)removeImageForName:(NSString *)name;
 
+#pragma mark Managing a Style’s Transition Options
+
+/**
+ The duration in seconds to animate any changes to the style URL or to layout and paint attributes.
```
1ec5: Huh, I overlooked the fact that changes to the style URL can also be animated. That’s kind of awkward, since changing the style URL means swapping one MGLStyle object for another.

1ec5: Actually, #5665 makes it sound like we don’t have “diff and patch” functionality in mbgl yet. Has the situation changed since July?

PR #7716, opened 2017-01-13
Observe layout guides

### LINKED ISSUE #6954: Compass underlaps top bar

frederoni:
Fixes #6954

This PR adds observers to the top and bottom layout guides if necessary in order to re-layout after changes have occurred. (i.e due to showing/hiding a navigation bar)

It also replaces constraints with manual layout to avoid manipulating the view controller's view which is the developer's responsibility.

animation

### DIFF
--- a/platform/ios/src/MGLMapView.mm
+++ b/platform/ios/src/MGLMapView.mm
@@ -131,6 +131,9 @@ const CGFloat MGLAnnotationImagePaddingForCallout = 1;
 
 const CGSize MGLAnnotationAccessibilityElementMinimumSize = CGSizeMake(10, 10);
 
+// Context for KVO observing UILayoutGuides.
+static void * MGLLayoutGuidesUpdatedContext = &MGLLayoutGuidesUpdatedContext;
+
 /// Unique identifier representing a single annotation in mbgl.
 typedef uint32_t MGLAnnotationTag;
 
@@ -233,13 +236,9 @@ public:
 @property (nonatomic) GLKView *glView;
 @property (nonatomic) UIImageView *glSnapshotView;
 @property (nonatomic, readwrite) MGLScaleBar *scaleBar;
-@property (nonatomic) NS_MUTABLE_ARRAY_OF(NSLayoutConstraint *) *scaleBarConstraints;
 @property (nonatomic, readwrite) UIImageView *compassView;
-@property (nonatomic) NS_MUTABLE_ARRAY_OF(NSLayoutConstraint *) *compassViewConstraints;
 @property (nonatomic, readwrite) UIImageView *logoView;
-@property (nonatomic) NS_MUTABLE_ARRAY_OF(NSLayoutConstraint *) *logoViewConstraints;
 @property (nonatomic, readwrite) UIButton *attributionButton;
-@property (nonatomic) NS_MUTABLE_ARRAY_OF(NSLayoutConstraint *) *attributionButtonConstraints;
 @property (nonatomic, readwrite) MGLStyle *style;
 @property (nonatomic) UITapGestureRecognizer *singleTapGestureRecognizer;
 @property (nonatomic) UITapGestureRecognizer *doubleTap;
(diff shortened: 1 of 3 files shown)


platform/ios/src/MGLMapView.mm
```diff
@@ -1866,6 +1890,10 @@ - (void)observeValueForKeyPath:(NSString *)keyPath ofObject:(id)object change:(N
             }
         }
     }
+    else if ([keyPath isEqualToString:@"bounds"] && [object isKindOfClass:NSClassFromString(@"_UILayoutGuide")])
```
frederoni: Comparing against the private class _UILayoutGuide is not particularly future proof.

platform/ios/src/MGLMapView.mm
```diff
@@ -122,6 +122,9 @@ typedef NS_ENUM(NSUInteger, MGLUserTrackingState) {
 
 const CGSize MGLAnnotationAccessibilityElementMinimumSize = CGSizeMake(10, 10);
 
+// Context for KVO observing UILayoutGuides.
+static char MGLLayoutGuidesUpdatedContext = 1;
```
1ec5: This works, but the more common practice is to pass in a selector such as `_cmd`.

platform/ios/src/MGLMapView.mm
```diff
@@ -655,6 +658,11 @@ - (void)dealloc
     [[NSNotificationCenter defaultCenter] removeObserver:self];
     [_attributionButton removeObserver:self forKeyPath:@"hidden"];
 
+    if (_isObservingLayoutGuides) {
+        [(UIView *)self.viewControllerForLayoutGuides.topLayoutGuide removeObserver:self forKeyPath:@"bounds" context:(void*)&MGLLayoutGuidesUpdatedContext];
```
1ec5: Is topLayoutGuide really a UIView? I'd expect it to be a UILayoutGuide.

platform/ios/src/MGLMapView.mm
1ec5: Er, I may be mistaken about `_cmd`. But in any case, the important thing is to ensure that the context’s _value_ is unique. If we later add a key-value observer with a different constant set to 1, suddenly there’s an opportunity for crosstalk.

A simple fix is to [set the context equal to its own address](http://nshipster.com/key-value-observing/#correct-context-declarations).

1ec5:
Postponing to v3.4.1, per https://github.com/mapbox/mapbox-gl-native/issues/6954#issuecomment-273283732.

1ec5:

frederoni:
- Rebased and retargeted against the 3.6 release branch.
- Fixed layout for the recently added scale bar.

observe

@boundsj @fabian-guerra 👀

platform/ios/src/MGLMapView.mm
```diff
@@ -931,6 +856,38 @@ - (void)layoutSubviews
     [self updateUserLocationAnnotationView];
 }
 
+- (void)layoutOrnaments
+{
+    CGFloat x, y;
```
1ec5: This is OK, but just a stylistic tip: `x` and `y` variables that get repurposed multiple times carry the risk that we might fail to update one of these variables and its previous value may “bleed” into a subsequent use of it.

Instead, I sometimes take advantage of the struct initializer syntax to keep the values scoped to where they’re used. For example:

```objc
self.compassView.center = {
    CGRectGetWidth(self.bounds) - CGRectGetMidX(self.compassView.bounds) - self.contentInset.right - 5,
    CGRectGetMidY(self.compassView.bounds) + self.contentInset.top + 5,
};
```

If you’re concerned about accidentally swapping `x` and `y`, you can also use the C99 designated initializer syntax:

```objc
self.compassView.center = {
    .x = CGRectGetWidth(self.bounds) - CGRectGetMidX(self.compassView.bounds) - self.contentInset.right - 5,
    .y = CGRectGetMidY(self.compassView.bounds) + self.contentInset.top + 5,
};
```

PR #7724, opened 2017-01-14
Wrap camera heading

### LINKED ISSUE #7723: MGLMapView.camera.heading goes negative when rotating to face northwest
When you rotate an MGLMapView so that it faces anywhere between due north and due west, exclusive, its `direction` property is wrapped to between 0° and 360°, but its `camera`’s `heading` goes negative, indicating an invalid direction according to `CLLocationDirection` semantics. Therefore, if you then set the `camera` property of any map view to that MGLMapCamera, you wind up facing the wrong direction.

This regression was introduced in #4790 but was not addressed by df0d1187da91537ef55fefc154159f1a6fd747ac.

/cc @incanus @friedbunny

1ec5:
Fixed an issue causing MGLMapView’s `camera`’s `heading` to be set to a negative value, indicating an undefined heading, when the map view faces northwest. The heading is now wrapped to between zero and 360 degrees, for consistency with MGLMapView’s `direction` property.

Fixes #7723.

/cc @friedbunny

### DIFF
--- a/platform/ios/uitest/MapViewTests.m
+++ b/platform/ios/uitest/MapViewTests.m
@@ -260,7 +260,7 @@
                                @"disabling pan gesture should disallow horizontal panning");
 }
 
-- (void)testRotate {
+- (void)testRotateClockwise {
     CLLocationDirection startAngle = tester.mapView.direction;
 
     XCTAssertNotEqual(startAngle,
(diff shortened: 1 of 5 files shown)


1ec5:
I put this PR on the v3.4.1 milestone, because it’s so close to the v3.4.0 release and we shipped this regression in v3.3.0. However, I’m open to moving it up to v3.4.0 if there’s a need for the fix sooner.

platform/ios/uitest/MapViewTests.m
```diff
@@ -274,6 +274,21 @@ - (void)testRotate {
...
+    CLLocationDirection startAngle = tester.mapView.direction;
+    
+    XCTAssertNotEqual(startAngle,
+                      -45,
+                      @"start angle must not be destination angle");
+    
+    [tester.mapView twoFingerRotateAtPoint:tester.mapView.center angle:-45];
+    
+    XCTAssertGreaterThanOrEqual(fabs(tester.mapView.direction - startAngle),
+                                -20,
```
incanus: The absolute value of one number will always be greater than or equal to another, negative number.

1ec5: Fixed in 1fa4c076c41913756297a4cba4e240a0732438f3.

PR #7725, opened 2017-01-14
Default to system font in attribution titles

### LINKED ISSUE #7722: Attribution is set in Times
In applications that link against the macOS SDK, the attribution view’s text is set in 9pt Times:

times

It should be set in the system font, as shown here in macosapp:

system

The font differs even inside `+[MGLAttributionInfo(Private) attributionInfosFromHTMLString:fontSize:linkColor:]`, where the attribution HTML string is converted to an attributed string.

This issue reproduces whether the application is set up in a XIB or a storyboard.

1ec5:
The inline stylesheet used when converting an attribution HTML string to an attributed string now specifies the system font as the default font. It still isn’t clear to me why macosapp has been defaulting to the system font, but it makes sense to me that the default default font would be Times without this change. This change makes use of WebKit’s proprietary `-apple-system` keyword, which [Apple recommends](https://webkit.org/blog/3709/using-the-system-font-in-web-content/) for this purpose.

There is still a slight difference between macosapp and an application outside macos.xcworkspace, but I think the two results are close enough:

fonts

Fixes #7722.

/cc @frederoni @friedbunny

### DIFF
--- a/platform/darwin/src/MGLAttributionInfo.mm
+++ b/platform/darwin/src/MGLAttributionInfo.mm
@@ -26,9 +26,20 @@
     // Apply a bogus, easily detectable style rule to any feedback link, since
     // NSAttributedString doesn’t preserve the class attribute.
     NSMutableString *css = [NSMutableString stringWithString:
+                            @"html { font-family: -apple-system, -apple-system-font, sans-serif; }"
                             @".mapbox-improve-map { -webkit-text-stroke-width: 1000px; }"];
     if (fontSize) {
-        [css appendFormat:@"html { font-size: %.1fpx; }", fontSize];
+        NSString *sizeRule = [NSString stringWithFormat:@"font-size: %.1fpx;", fontSize];
+#if !TARGET_OS_IPHONE
+        if (fontSize == [NSFont systemFontSizeForControlSize:NSMiniControlSize]) {
+            sizeRule = @"font: -webkit-mini-control";
+        } else if (fontSize == [NSFont systemFontSizeForControlSize:NSSmallControlSize]) {
+            sizeRule = @"font: -webkit-small-control";
+        } else if (fontSize == [NSFont systemFontSizeForControlSize:NSRegularControlSize]) {
+            sizeRule = @"font: -webkit-control";
+        }
+#endif
+        [css appendFormat:@"html { %@ }", sizeRule];
     }
     if (linkColor) {
         CGFloat red;


1ec5:
This issue is tagged as a release blocker because it’s a highly visible regression on macOS. It is not a release blocker on iOS, as it affects iOS only to the extent that a developer would try to display an MGLAttributionInfo title as rich text. The attribution sheet is unaffected because it displays each title as plain text. Even so, I’d like to land this change in time for macOS SDK v0.3.0, which means it’ll be in iOS SDK v3.4.0 as well.

friedbunny:
The [WebKit blog says](https://webkit.org/blog/3709/using-the-system-font-in-web-content/) `-apple-system` is supported from OS X 10.11 (and iOS 9, incidentally). We support back to 10.10, so I think it’d make sense to also add `"Helvetica Neue"` to this rule, the previous system font.

The only question I would then have is if OS X 10.10 still used Helvetica Neue at small sizes...

1ec5:
> The only question I would then have is if OS X 10.10 still used Helvetica Neue at small sizes...

Yes, Lucida Grande was supplanted by Helvetica Neue at all three control sizes.

1ec5:
Support for the `-apple-system` keyword was introduced in WebKit/webkit@a03afe285974b0ee0d3dcd0fc017d7ac21359e7c for [bug 144707](https://bugs.webkit.org/show_bug.cgi?id=144707), which made it into iOS 9 and OS X 10.10. But the same functionality was actually available earlier via the identical `-apple-system-font` keyword, upstreamed from iOS in WebKit/webkit@b8ade587ad740cb2837293d8bc77cd22fe82bb7b for [bug 126654](https://bugs.webkit.org/show_bug.cgi?id=126654) (which would’ve made it into [iOS 8](https://trac.webkit.org/browser/tags/Safari-538.35/Source/WebCore/platform/graphics/ios/FontCacheIOS.mm#L570)) and introduced to OS X in WebKit/webkit@989ba141ffea9573061d337e73bea17467399a99 for [bug 133382](https://bugs.webkit.org/show_bug.cgi?id=133382) (which made it into [Safari 538.1](https://trac.webkit.org/browser/tags/Safari-538.1/Source/WebCore/platform/mac/WebFontCache.mm) and thus [OS X 10.10](https://en.wikipedia.org/wiki/Safari_version_history#Mac)). Therefore I’ve added `-apple-system-font` as a fallback. I also added `sans-serif` to at least ensure something reasonably close to Helvetica Neue on iOS 7.

Additionally, on macOS, I set `font: -webkit-mini-control` to apply a font identical to what macosapp displays:

font

friedbunny:
@1ec5 You’re absolutely correct — the blog piece I cited doesn’t actually say when `-apple-system` became available, just that you could use San Francisco via that keyword in OS X 10.11 and iOS 9. :bow:

PR #7730, opened 2017-01-16
Correct podspecs, readmes

1ec5:
Synchronized bundled README installation instructions with those in INSTALL.md. Bumped screenshot up a bit. Corrected link to published documentation.

/cc @frederoni @boundsj


1ec5:
Three of the four podspecs failed `pod spec lint`, so I corrected those issues as well.

1ec5:
Three of the four podspecs failed `pod spec lint`, so I corrected those issues as well.

PR #7731, opened 2017-01-16
Update iOS, macOS changelogs

1ec5:
Updated the iOS SDK and macOS SDK changelogs to mention #7446, #7356, #7465, #7616, #7445, #7444, #7526, #7586, #7574, and #7770. Also corrected the blurb about #7711. The changelogs are now current as of bf82b07ac8679b30b5f98cb2393c531d2a7d4e12.

### DIFF
--- a/platform/ios/CHANGELOG.md
+++ b/platform/ios/CHANGELOG.md
@@ -4,11 +4,29 @@ Mapbox welcomes participation and contributions from everyone. Please read [CONT
 
 ## master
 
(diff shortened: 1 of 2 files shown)


1ec5:
https://github.com/mapbox/mapbox-gl-native/pull/7711#pullrequestreview-17299141 made me aware of the fact that the Android changelog is also being compiled incrementally.

@cammace, do any of these mbgl changes rise to the level of a blurb in the Android changelog?

PR #7734, opened 2017-01-16
iOS: Add custom layer example to debugging app

kkaefer:


1ec5:
@friedbunny removed the original custom layer example in #5091. There were concerns about making this functionality look officially supported. Putting the example into macosapp was a compromise because the macOS SDK isn’t officially supported anyways.

friedbunny:
Not that I’m the final arbiter of these things, but I think this would be alright to re-add, presuming we add more warnings about it being unsupported and at-your-own-risk. ([As we do in MGLOpenGLStyleLayer.](https://github.com/mapbox/mapbox-gl-native/blob/df2b6cfec1d3acb41520d1277809ba3d17a7ceb7/platform/darwin/src/MGLOpenGLStyleLayer.mm#L70-L71))

platform/ios/app/LimeGreenStyleLayer.h
```diff
@@ -0,0 +1,5 @@
+#import <Mapbox/Mapbox.h>
+
+@interface LimeGreenStyleLayer : MGLOpenGLStyleLayer
```
1ec5: As far as I know, LimeGreenStyleLayer.h and LimeGreenStyleLayer.m would be the first files that are identical between iosapp and macosapp. Do you think we should keep two copies around or share the code in platform/darwin/app/?

friedbunny:
@jmkiley I saw that you pushed an updated branch for this — would you like to carry this PR over the finish line? 🤔

jmkiley:
👍  Rebased this PR on master.

friedbunny:
@jmkiley actually, how about retargeting this against `release-boba` — that should make the eventual merge back for that branch somewhat easier, if `master` hasn’t diverged as much.

platform/darwin/app/LimeGreenStyleLayer.m
```diff
@@ -1,7 +1,12 @@
 #import "LimeGreenStyleLayer.h"
 
-#include <OpenGLES/ES2/gl.h>
-#include <OpenGLES/ES2/glext.h>
+#if TARGET_OS_IPHONE
```
friedbunny: If we were to instead `@import GLKit;` here, that would allow us to get rid of these conditionals and remove `OpenGLES.framework` as an explicitly linked framework in iosapp.

PR #7736, opened 2017-01-16
WIP GLSurfaceView

### LINKED ISSUE #5766: Implement GLSurfaceView
This ticket is where work on refactoring Mapbox Android to use `GLSurfaceView` will take place.

In #5000 it was originally proposed to bring back `SurfaceView` and `GLSurfaceView` in favor of `TextureView` to increase performance of Map rendering and interactivity.  While working on #5000 it was suggested to the team to implement `SurfaceView` first as it's a more straightforward implementation and doesn't require as much Core GL / C++ work while still providing a performance improvement so that's the direction the ticket took.  To keep things more clear, compartmentalized, and easier to track we decided to break the two implementations up into separate tickets.

/cc @mapbox/mobile

tobrun:
WIP, Closes #5766, this PR migrates from using SurfaceView as a drawing surface to a GLSurfaceView implementation. More information on this topic in #5766. 

Current state:
 - [x] replace SurfaceView with GLSurfaceView
 - [x] remove (deprecated) support for TextureView 
 - [x] allow GLSurfaceView to create and maintain GL components (eglContext, eglDisplay, etc).
 - [x] post callbacks coming from the render thread to the main thread (invalidate, mapchange, etc.)
 - [x] cleanup callback code 
 - [ ] move rendering to the render thread*  
 - [ ] ...

*= it's currently unclear what next actions are to resolve this, currently queuing a runnable to be executed on the render thread is resulting in a native crash. I have tried instantiating the map object on the render thread itself but our runloop implementation doesn't agree with that.

### DIFF
--- a/platform/android/MapboxGLAndroidSDK/src/main/java/com/mapbox/mapboxsdk/maps/MapView.java
+++ b/platform/android/MapboxGLAndroidSDK/src/main/java/com/mapbox/mapboxsdk/maps/MapView.java
@@ -76,8 +76,13 @@ public class MapView extends FrameLayout implements GLSurfaceView.Renderer {
   private boolean hasSurface = false;
 
   private GLSurfaceView glSurfaceView;
+  private CompassView compassView;
+  private ImageView attrView;
+  private View logoView;
+  private MyLocationView myLocationView;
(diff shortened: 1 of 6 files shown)


tobrun:
continous in #8031

PR #7739, opened 2017-01-16
[core] Improve tile cache sizing calculation

brunoabinader:
`TileCacheStrategy` can now be passed on `Map` ctor as a hint for the style sources about their cache strategy:
- `NoCache`: Do not retain tiles.
- `ConservativeCache`: Retain visible tiles from all zoom levels.
- `DoubleCache`: Twice as conservative.

This is also combined with updating the cache size whenever the transform dimensions or zoom range changes.

### DIFF
--- a/src/mbgl/style/source_impl.cpp
+++ b/src/mbgl/style/source_impl.cpp
@@ -139,7 +139,7 @@ void Source::Impl::updateTiles(const UpdateParameters& parameters) {
     algorithm::updateRenderables(getTileFn, createTileFn, retainTileFn, renderTileFn,
                                  idealTiles, zoomRange, tileZoom);
 
-    if (type != SourceType::Annotations && cache.getSize() == 0) {
+    if (type != SourceType::Annotations) {
         size_t conservativeCacheSize =
             std::max((float)parameters.transformState.getSize().width / tileSize, 1.0f) *
             std::max((float)parameters.transformState.getSize().height / tileSize, 1.0f) *
--- a/src/mbgl/style/source_impl.cpp
+++ b/src/mbgl/style/source_impl.cpp
@@ -141,8 +141,8 @@ void Source::Impl::updateTiles(const UpdateParameters& parameters) {
 
     if (type != SourceType::Annotations && cache.getSize() == 0) {
         size_t conservativeCacheSize =
-            std::max((float)parameters.transformState.getSize().width / util::tileSize, 1.0f) *
-            std::max((float)parameters.transformState.getSize().height / util::tileSize, 1.0f) *
+            std::max((float)parameters.transformState.getSize().width / tileSize, 1.0f) *
+            std::max((float)parameters.transformState.getSize().height / tileSize, 1.0f) *
             (parameters.transformState.getMaxZoom() - parameters.transformState.getMinZoom() + 1) *
             0.5;
         cache.setSize(conservativeCacheSize);


include/mbgl/map/mode.hpp
```diff
@@ -37,6 +37,16 @@ enum class ViewportMode : EnumType {
...
 };
 
+// Selects the tile cache strategy, as follows:
+// - NoCache: Do not retain tiles.
+// - ConservativeCache: Retain visible tiles from all zoom levels.
+// - DoubleCache: Twice as conservative.
+enum class TileCacheStrategy : EnumType {
+    NoCache = 0,
+    ConservativeCache,
+    DoubleCache
```
tmpsantos: Agressive?

mb12:
Does 'NoCache' also disable disk cache (sqlite)?

tmpsantos:
> Does 'NoCache' also disable disk cache (sqlite)?

Semantics need to be clarified. But this is not the disk cache, it is the memory cache of parsed tiles.

include/mbgl/map/mode.hpp
```diff
@@ -37,6 +37,17 @@ enum class ViewportMode : EnumType {
...
+// Selects the memory cache of parsed tiles strategy, as follows:
+// - NoCache: Do not retain tiles.
+// - ConservativeCache: Retain visible tiles from all zoom levels.
+// - AggressiveCache: Twice as conservative.
+// Notice that it doesn't affect the disk cache (SQLite) operation.
+enum class TileCacheStrategy : EnumType {
+    NoCache = 0,
+    ConservativeCache,
+    AggressiveCache
+};
```
kkaefer: Let's add `Memory` to the name of this enum

src/mbgl/style/source_impl.cpp
```diff
@@ -139,12 +139,16 @@ void Source::Impl::updateTiles(const UpdateParameters& parameters) {
     algorithm::updateRenderables(getTileFn, createTileFn, retainTileFn, renderTileFn,
                                  idealTiles, zoomRange, tileZoom);
 
-    if (type != SourceType::Annotations && cache.getSize() == 0) {
+    bool isSameDimension = (lastDimension == parameters.transformState.getSize());
+    bool isSameZoomRange = (lastZoomRange == parameters.transformState.getZoomRange());
+    if (type != SourceType::Annotations && (cache.getSize() == 0 || !isSameDimension || !isSameZoomRange)) {
```
kkaefer: I don't think this optimization is worth it

src/mbgl/style/source_impl.cpp
```diff
@@ -139,17 +139,23 @@ void Source::Impl::updateTiles(const UpdateParameters& parameters) {
     algorithm::updateRenderables(getTileFn, createTileFn, retainTileFn, renderTileFn,
                                  idealTiles, zoomRange, tileZoom);
 
+    bool isSameStrategy = (cacheStrategy == parameters.tileCacheStrategy);
     bool isSameDimension = (lastDimension == parameters.transformState.getSize());
     bool isSameZoomRange = (lastZoomRange == parameters.transformState.getZoomRange());
-    if (type != SourceType::Annotations && (cache.getSize() == 0 || !isSameDimension || !isSameZoomRange)) {
+    if (type != SourceType::Annotations && (!isSameStrategy || !isSameDimension || !isSameZoomRange)) {
+        cacheStrategy = parameters.tileCacheStrategy;
```
kkaefer: The caching strategy can't change, since it's only ever set in the constructor. It shouldn't be part of the `UpdateParameters`.

jfirebaugh:
Does this actually need to be configurable? I think 48dac2e6504efb2cac0a81143084cc8e31ebdfb1 may be sufficient.

kkaefer:
Like @jfirebaugh I doubt that this is something we'll have to configure. Instead, we could also make the cache size dependent on currently available RAM: if there's space free, we can cache as much as we want, but if there isn't, evicting items and the short reload delay are preferable to the App hitting OOM.

brunoabinader:
> Does this actually need to be configurable? I think 48dac2e and fc4f56d may be sufficient.

Thanks for the feedback @jfirebaugh and @kkaefer - the only downside I see is that with fc4f56d the cache size gets recalculated on every `Source::Impl::updateTiles` for non-annotation sources - however I don't see any immediate perf impact by doing that.

I'll cherry-pick these and update this PR accordingly.

jfirebaugh:
@kkaefer Can you re-review? 🙇

PR #7741, opened 2017-01-16
[core] Prefetch low resolution tiles

### LINKED ISSUE #1626: Never show blank spots
When zooming out, or panning aggressively, we're often showing blank screens because tiles haven't loaded or parsed yet. We should lazily keep tiles in cache that are a couple of zoom levels lower so we can show it immediately when the user zooms or pans.

tmpsantos:
The idea is to try to cover more map area and show the user something ASAP while the other tiles are still loading.

Exposes 2 new APIs:

- Dynamic: Will request tiles from current zoom level - N. Better resolution on the prefetched tiles.
- Fixed: Will always request tiles from zoom level N regardless of the current zoom level. Allows optimize for bandwidth.

Fixes #1626


include/mbgl/map/map.hpp
```diff
@@ -185,6 +185,10 @@ class Map : private util::noncopyable {
     std::vector<Feature> queryRenderedFeatures(const ScreenBox&,        const optional<std::vector<std::string>>& layerIDs = {});
     AnnotationIDs queryPointAnnotations(const ScreenBox&);
 
+    // Tile prefatching
```
kkaefer: typo

include/mbgl/map/map.hpp
```diff
@@ -185,6 +185,10 @@ class Map : private util::noncopyable {
...
+    void setFixedPrefetchZoom(uint32_t);
+    void setDynamicPrefetchZoom(uint32_t);
```
kkaefer: Missing documentation

src/mbgl/style/source_impl.cpp
```diff
@@ -89,19 +89,34 @@ void Source::Impl::updateTiles(const UpdateParameters& parameters) {
...
         // Make sure we're not reparsing overzoomed raster tiles.
         if (type == SourceType::Raster) {
             tileZoom = idealZoom;
         }
 
         idealTiles = util::tileCover(parameters.transformState, idealZoom);
+
+        // Request lower zoom level tiles (if configure to do so) in an attempt
+        // to show something on the screen faster at the cost of a little of bandwidth.
+        lowResolutionZoom = parameters.fixedPrefetchZoom ?
```
kkaefer: This breaks if `fixedPrefetchZoom` is set to `0`, which is a valid use case.

src/mbgl/style/source_impl.cpp
```diff
@@ -89,19 +89,34 @@ void Source::Impl::updateTiles(const UpdateParameters& parameters) {
...
+            parameters.fixedPrefetchZoom :
+            idealZoom - parameters.dynamicPrefetchZoom;
+
+        lowResolutionZoom = lowResolutionZoom < zoomRange.min ? zoomRange.min : lowResolutionZoom;
+
+        if (lowResolutionZoom < idealZoom) {
+            lowResolutionTiles = util::tileCover(parameters.transformState, lowResolutionZoom);
```
kkaefer: In previous implementations (JS), we've called this `"panZoom"`. Not sure if that's a better name though.

src/mbgl/style/source_impl.cpp
```diff
@@ -111,8 +126,10 @@ void Source::Impl::updateTiles(const UpdateParameters& parameters) {
     std::set<OverscaledTileID> retain;
 
     auto retainTileFn = [&retain](Tile& tile, Resource::Necessity necessity) -> void {
-        retain.emplace(tile.id);
-        tile.setNecessity(necessity);
+        if (retain.find(tile.id) == retain.end()) {
+            retain.emplace(tile.id);
+            tile.setNecessity(necessity);
+        }
```
kkaefer:
```cpp
if (retain.emplace(tile.id).second) {
    tile.setNecessity(necessity);
}
```

 `.second` is `true` when something was inserted, `false` otherwise.

src/mbgl/style/source_impl.cpp
```diff
@@ -89,19 +89,34 @@ void Source::Impl::updateTiles(const UpdateParameters& parameters) {
...
             tileZoom = idealZoom;
         }
 
         idealTiles = util::tileCover(parameters.transformState, idealZoom);
+
+        // Request lower zoom level tiles (if configure to do so) in an attempt
+        // to show something on the screen faster at the cost of a little of bandwidth.
+        lowResolutionZoom = parameters.fixedPrefetchZoom ?
+            parameters.fixedPrefetchZoom :
+            idealZoom - parameters.dynamicPrefetchZoom;
```
kkaefer: possible integer overflow here if `idealZoom < parameters.dynamicPrefetchZoom`?

src/mbgl/style/source_impl.cpp
```diff
@@ -89,19 +89,34 @@ void Source::Impl::updateTiles(const UpdateParameters& parameters) {
...
+
+        lowResolutionZoom = lowResolutionZoom < zoomRange.min ? zoomRange.min : lowResolutionZoom;
```
kkaefer: Can we use `std::min` to make this more readable?

src/mbgl/style/update_parameters.hpp
```diff
@@ -42,6 +46,9 @@ class UpdateParameters {
 
     // TODO: remove
     Style& style;
+
+    const uint32_t fixedPrefetchZoom;
+    const uint32_t dynamicPrefetchZoom;
```
kkaefer: What is the use case for having both fixed and dynamic prefetch? Which one takes precedence?

src/mbgl/util/tile_cover.cpp
```diff
@@ -127,8 +127,8 @@ std::vector<UnwrappedTileID> tileCover(const Point<double>& tl,
 
 } // namespace
 
-int32_t coveringZoomLevel(double zoom, SourceType type, uint16_t size) {
-    zoom += std::log(util::tileSize / size) / std::log(2);
+uint32_t coveringZoomLevel(double zoom, SourceType type, uint16_t size) {
+    zoom = std::max(0., zoom) + std::log(util::tileSize / size) / std::log(2);
```
kkaefer: Should this `max` call include the addition (which accounts for satellite tile size differences)? Otherwise, it could be impossible to use z0 for satellite tiles.

src/mbgl/style/source_impl.cpp
tmpsantos: I'm OK with that. Let's use `panZoom` for the sake of consistency.

src/mbgl/style/update_parameters.hpp
tmpsantos: Documented now. Fixed takes precedence.

tmpsantos:
:eyes: @kkaefer

tmpsantos:
Should I close this as deferred?

src/mbgl/map/map.cpp
```diff
@@ -95,6 +95,9 @@ class Map::Impl : public style::Observer {
 
     std::unique_ptr<AsyncRequest> styleRequest;
 
+    int32_t fixedPrefetchZoom = -1;
```
tmpsantos: Done.

tmpsantos:
@kkaefer is disabling partial tile rendering enough for landing this patch?

I did a cheap test by simply tweaking `isRenderable()`, and it does improve flickering, but pretty much makes prefetching a requirement for areas with a lot of symbols.

```--- a/src/mbgl/tile/tile.hpp
+++ b/src/mbgl/tile/tile.hpp
@@ -74,7 +74,7 @@ public:
     // partial state is still waiting for network resources but can also
     // be rendered, although layers will be missing.
     bool isRenderable() const {
-        return availableData != DataAvailability::None;
+        return isComplete();
     }```

That said, nothing stops us for landing this today, since it is still very useful for a satellite map and would spare me from rebase this branch and it is optional anyway.

**Comment by tmpsantos** [comment]
ezgif-3-27c4b7c630

@kkaefer rebased and ready for reviewing again. This new version enables prefetching only for raster until we fix the label flickering. On this gif you see the difference (the tiles are offline cached on both scenarios, it is even more evident when you get tiles from the network).

File: include/mbgl/map/map.hpp [hunk 11]
```diff
@@ -156,6 +156,21 @@ class Map : private util::noncopyable {
...
+    // tile source range and if it is lower than the current zoom level, in a attempt to display
+    // a map as quick as possible. `DynamicPrefetch` works similarly, but will prefetch a tile
+    // at `zoom = (getZoom() - delta)`, constrained at the lowest zoom level for the source.
+    //
+    // If both are set, only a tile for the fixed zoom is requested.
+    void setFixedPrefetchZoom(optional<uint8_t> zoom);
+    optional<uint8_t> getFixedPrefetchZoom() const;
+
+    void setDynamicPrefetchZoomDelta(optional<uint8_t> delta);
+    optional<uint8_t> getDynamicPrefetchZoomDelta() const;
```
tmpsantos: The fixed is if you want to be cheap: you can set the prefetch zoom level to say, zoom level 5 and you will never have a blank spot on the map, but it will look very low res and overscaled. Soon you will get the whole world at zoom 5 in your cache.

The dynamic IMO is more interesting, but it will use more resources. You could set it to say, 2, and it will always prefetch "CurrentZoom - 2", it will pretty much make you never see blank spots with better looking overscaled tile but it will fill the cache faster and also use more bandwidth.

I find both approaches useful. We could drop the "dynamic" API and leave it to be implemented client side if you think we don't need 2 APIs.

src/mbgl/renderer/tile_pyramid.cpp
```diff
@@ -138,6 +158,12 @@ void TilePyramid::update(const std::vector<Immutable<style::Layer::Impl>>& layer
     };
 
     renderTiles.clear();
+
+    if (!panTiles.empty()) {
+        algorithm::updateRenderables(getTileFn, createTileFn, retainTileFn,
+                [](const UnwrappedTileID&, Tile&) {}, panTiles, zoomRange, panZoom);
+    }
+
```
tmpsantos: Capturing from the chat: we have a separated `algorithm::updateRenderables` because we don't create render tiles for pan tiles.

1ec5:
Would you like to implement the SDK side of this change, or should we track that in separate tickets?

tmpsantos:
> Would you like to implement the SDK side of this change, or should we track that in separate tickets?

I would like to track it in a separated ticket.

1ec5:
#9437 for Android; #9438 for iOS and macOS.

PR #7743, opened 2017-01-16
Expose DefaultFileSource via jni.hpp

kkaefer:
This pull request wraps `DefaultFileSource` with the high level jni.hpp bindings. It also does these things:

- Remove access token API from the `MapView`/`NativeMapView` objects
- Remove API base URL API from the `MapboxMap` and `MapboxMapOptions` objects
- Remove `mapbox_apiBaseUrl` from `public.xml` and `attrs.xml`
- Added optional `mapbox_api_base_url` to `developer-config.xml`
- The `NativeMapView` constructor now takes a `DefaultFileSource` object and reuses the object instead of creating its own instance.


kkaefer:
What's missing is a conversion of the `DefaultFileSource` in `OfflineManager`. It still uses its own instance.

platform/android/src/storage/default_file_source_peer.hpp
```diff
@@ -0,0 +1,42 @@
...
+
+#include <memory>
+
+namespace mbgl {
+
+class DefaultFileSource;
+
+namespace android {
+
+class DefaultFileSourcePeer {
```
ivovandongen: We haven't been using the *Peer suffix up till now, but makes sense.

kkaefer: I decided to use it because we'll get conflicts with the regular `DefaultFileSource` otherwise. In the style code, we're using absolutely qualified identifiers `mbgl::style::SymbolLayer` in those places, which works, but is just another source of potential conflicts.

kkaefer:
@jfirebaugh I pushed an optional second commit that adds a `Peer` object, which is a catchall template, allowing referring to wrapped types without defining an explicit header file. Instead, you declare the wrapped object as a subclass in the implementation file entirely, which removes duplication and makes the whole thing more concise.

platform/android/src/native_map_view.cpp
```diff
@@ -65,13 +66,9 @@ NativeMapView::NativeMapView(JNIEnv* env_,
...
     }
 
-    jni::Field<DefaultFileSourcePeer, jlong> peerField{ *env, DefaultFileSourcePeer::javaClass,
-                                                        "peer" };
-    auto fileSourcePtr = reinterpret_cast<DefaultFileSourcePeer*>(fileSource->Get(*env, peerField));
-
     map = std::make_unique<mbgl::Map>(
         *this, mbgl::Size{ static_cast<uint32_t>(width), static_cast<uint32_t>(height) },
-        pixelRatio, fileSourcePtr->getFileSource(), threadPool, MapMode::Continuous);
+        pixelRatio, Peer<DefaultFileSource>::Get(*env, *fileSource), threadPool, MapMode::Continuous);
```
kkaefer: moved the whole peer object accessor to `Peer::Get`

platform/android/src/jni/peer.hpp
```diff
@@ -0,0 +1,40 @@
...
+private:
+    static const char* peer;
+    static jni::Class<Peer<T>> javaClass;
+};
+
+template <typename T>
+jni::Class<Peer<T>> Peer<T>::javaClass;
+
+template <typename T>
+const char* Peer<T>::peer = "peer";
```
kkaefer: defaults to `"peer"`, but can be customized per wrapped class if the Java implementation uses a different name

kkaefer:
@jfirebaugh I know you expressed concern about exposing `DefaultFileSource`. What we're doing here is mirror iOS, which exposes the `DefaultFileSource` singleton as a `MGLOfflineStorage` object used by all instances of maps created in the application.

Android has a `OfflineManager` class as well, which is the current Android counterpart to `OfflineManager`, but it uses its own instance of `DefaultFileSource` (even after this patch). This is something I'd like to remove, so that we only use one instance of `DefaultFileSource`. Ultimately, what I'd like to move to is a model where we automatically wrap all C++ objects with some type of auto-generated binding, and then implement the actual SDK logic on top of that in pure Java (similar to how the iOS bindings work), rather than having a mix of C++ and Java code that is very hard to follow.

zugaldia:
> Ultimately, what I'd like to move to is a model where we automatically wrap all C++ objects with some type of auto-generated binding, and then implement the actual SDK logic on top of that in pure Java (similar to how the iOS bindings work), rather than having a mix of C++ and Java code that is very hard to follow.

I'd love to move in this direction too 👍 .

The current approach could make more sense when our Android API surface area was basically meant to provide compatibility with third-party offerings. Currently, however, we're past that and we offer a growing API set (i.e. offline, runtime, DDS, extrusions) that have no immediate translation externally. I'd like to move into having a lower-level consistent (possibly auto-generated on top of https://github.com/mapbox/jni.hpp) layer and build everything else on top of that.

jfirebaugh:
Why do we need to expose `DefaultFileSource` via JNI in order for `OfflineManager` to share its instance with maps? Wouldn't an alternate approach be for the map constructor to obtain the `DefaultFileSource` instance from the `OfflineManager` singleton?

jfirebaugh:
Stale.

PR #7759, opened 2017-01-17
Typo fixes, added line to MGLStyleLayer initializer description

jmkiley:
- Corrected typos in MGLMapView.h and MGLMapCamera.h
- Removed line from MGLShapeCollection.h
- Inserted a reminder in MGLStyleLayer.h initializer description to remind user not to initialize MGLStyleLayer directly

### DIFF
--- a/platform/darwin/src/MGLShapeCollection.h
+++ b/platform/darwin/src/MGLShapeCollection.h
@@ -28,9 +28,6 @@ NS_ASSUME_NONNULL_BEGIN
  convenient to use an `MGLPointCollection`, `MGLMultiPolyline`, or
  `MGLMultiPolygon` object, respectively.
  
- A multipolyline is known as a
- <a href="https://tools.ietf.org/html/rfc7946#section-3.1.8">GeometryCollection</a>
- geometry in GeoJSON.
  */
 @interface MGLShapeCollection : MGLShape
 
--- a/platform/darwin/src/MGLShapeCollection.h
+++ b/platform/darwin/src/MGLShapeCollection.h
@@ -28,6 +28,9 @@ NS_ASSUME_NONNULL_BEGIN
  convenient to use an `MGLPointCollection`, `MGLMultiPolyline`, or
  `MGLMultiPolygon` object, respectively.
  
+ A shape collection is known as a 
+ <a href="https://tools.ietf.org/html/rfc7946#section-3.1.8">GeometryCollection</a> 
+ geometry in GeoJSON.
  */
 @interface MGLShapeCollection : MGLShape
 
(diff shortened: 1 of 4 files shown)


platform/darwin/src/MGLShapeCollection.h
```diff
@@ -28,9 +28,6 @@ NS_ASSUME_NONNULL_BEGIN
  convenient to use an `MGLPointCollection`, `MGLMultiPolyline`, or
  `MGLMultiPolygon` object, respectively.
  
- A multipolyline is known as a
- <a href="https://tools.ietf.org/html/rfc7946#section-3.1.8">GeometryCollection</a>
- geometry in GeoJSON.
```
1ec5: This sentence was correct except for the inclusion of “multipolyline” instead of “shape collection”.

platform/darwin/src/MGLStyleLayer.h
```diff
@@ -29,7 +29,10 @@ NS_ASSUME_NONNULL_BEGIN
  
  After initializing and configuring the style layer, add it to a map view’s
  style using the `-[MGLStyle addLayer:]` or
- `-[MGLStyle insertLayer:belowLayer:]` method.
+ `-[MGLStyle insertLayer:belowLayer:]` method. Should not be used to create an 
+ instance of MGLStyleLayer directly. Instead, create instances of
+ `MGLBackgroundStyleLayer` and the concrete subclasses of
+ `MGLForegroundStyleLayer`.
```
1ec5: This change will get overwritten the next time someone runs `make style-code-darwin`. This documentation is defined in [style-spec-overrides-v8.json](https://github.com/mapbox/mapbox-gl-native/blob/fc9f0ad9427fd61397f4d6eede4a6e5b746d088c/platform/darwin/scripts/style-spec-overrides-v8.json).

jmkiley:
Removed the change to MGLStyleLayer.h, edited the line in MGLShapeCollection.h. cc @1ec5

1ec5:
Did you still want to point this out?

> Should not be used to create an instance of MGLStyleLayer directly. Instead, create instances of `MGLBackgroundStyleLayer` and the concrete subclasses of `MGLForegroundStyleLayer`.

You can do that in style-spec-overrides-v8.json, then run `make style-code-darwin`. You should see the line magically reappear in the header as a result.

Regarding this note, I’d say something like the following, so developers know they _can_ use this initializer, just not on MGLStyleLayer itself:

> The default implementation of this initializer in MGLStyleLayer creates an invalid style layer. Call this initializer on `MGLBackgroundStyleLayer` or one of the concrete subclasses of `MGLForegroundStyleLayer` to create a valid style layer.

platform/darwin/src/MGLStyleLayer.h
1ec5: Actually, this is the one layer file that isn’t autogenerated, so you can put the additional text directly in here. Sorry for the confusion.

jmkiley:
Inserted the line @1ec5 wrote into the MGLStyleLayer initializer method.

PR #7766, opened 2017-01-18
git + npm is run on every CMake invocation

kkaefer:
There are [checks](https://github.com/mapbox/mapbox-gl-native/blob/master/CMakeLists.txt#L7-L25) for both of these, but they don't work since the current working directory is not the source directory (it's the build directory). Instead, we'll have to supply the absolute path for these checks to work correctly.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -3,27 +3,6 @@ project(mbgl LANGUAGES CXX C)
 set(CMAKE_CXX_STANDARD 14)
 
 include(cmake/mbgl.cmake)
-
-if(NOT EXISTS ".mason/mason.cmake")
-   execute_process(
-        COMMAND git submodule update --init .mason
-        WORKING_DIRECTORY ${CMAKE_SOURCE_DIR})
-endif()
-
-if(NOT EXISTS "mapbox-gl-js/package.json")
-   execute_process(
-        COMMAND git submodule update --init mapbox-gl-js
-        WORKING_DIRECTORY ${CMAKE_SOURCE_DIR})
-   # Symlink mapbox-gl-js/node_modules so that the modules that are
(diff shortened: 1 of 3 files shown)


kkaefer:
Changed this so that we are always running a submodule update when the project is generated *and* as part of every build (`npm-install` is a dependency of `mbgl-header` now).

The `npm-install` target depends on the `package.json` files of both the main and the `mapbox-gl-js` project, which means we are running `npm update` whenever the package.json was changed. We are using the `node_modules` folder and touch it to store the update state (the Makefile does something similar). We'll have to see how well using the folder as freshness source works; if it doesn't, we can move to individual stamp files.

This approach has the advantage that it doesn't run `npm install` most of the time; only when the `package.json` has changed. The submodule is always run, but I've found that it is very fast in the no-op case (<0.2s).

Long term outlook: I'm planning to migrate node-cmake to the new version ([blocker](https://github.com/cjntaylor/node-cmake/issues/17)); the new workflow of version 2 is to include the CMake file within your project. That means we can remove any calls to `npm` from our Makefile and have this code in CMake handle npm installation entirely. Planning to do the same thing with Mason (we don't actually need the .mason submodule, except for the .cmake file, since we're now using [mason.ini](https://github.com/mapbox/mason/pull/176) for all packages.

kkaefer:
Let's try this for a few days. It adds a network requirement to the build process, but only if you have out-of-date dependencies: It'll be hard to build without internet access after you've updated the repo, but not the dependencies. Once you've done the `npm install` of the dependencies, no network access is required until you modify the `package.json`. It think this is an acceptable compromise that removes errors occurring when failing to update the npm dependencies or submodules.

Note that when you are making changes to a submodule, you'll have to stage these changes prior to building (otherwise the submodule will be reset to the pinned version).

PR #7767, opened 2017-01-18
deprecation cleanup for 5.0.0

### LINKED ISSUE #7575: Remove all deprecated methods

tobrun:
Closes #7575, finalising public api for 5.0.0, cleanup deprecation
- remove deprecated emerald
- remove deprecation annotation for setStyle
- remove deprecated dismiss tracking on gesture

This PR doesn't include the removal of TextureView. 
As proposed [here](https://github.com/mapbox/mapbox-gl-native/issues/7575#issuecomment-270070568), I'm going to pick this up with https://github.com/mapbox/mapbox-gl-native/issues/5766.

Review @zugaldia

### DIFF
--- a/platform/android/MapboxGLAndroidSDK/src/main/java/com/mapbox/mapboxsdk/maps/MapboxMap.java
+++ b/platform/android/MapboxGLAndroidSDK/src/main/java/com/mapbox/mapboxsdk/maps/MapboxMap.java
@@ -824,7 +824,6 @@ public final class MapboxMap {
    * @deprecated use {@link #setStyleUrl(String)} instead with versioned url methods from {@link Style}
    */
   @UiThread
-  @Deprecated
   public void setStyle(@Style.StyleUrl String style) {
     setStyleUrl(style);
   }
--- a/platform/android/MapboxGLAndroidSDK/src/main/java/com/mapbox/mapboxsdk/constants/Style.java
+++ b/platform/android/MapboxGLAndroidSDK/src/main/java/com/mapbox/mapboxsdk/constants/Style.java
@@ -23,7 +23,7 @@ public class Style {
    * constants means your map style will always use the latest version and may change as we
    * improve the style
    */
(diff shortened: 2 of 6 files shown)


platform/android/MapboxGLAndroidSDK/src/main/java/com/mapbox/mapboxsdk/maps/MapboxMap.java
```diff
@@ -824,7 +824,6 @@ public void setStyleUrl(@NonNull String url) {
    * @deprecated use {@link #setStyleUrl(String)} instead with versioned url methods from {@link Style}
    */
   @UiThread
-  @Deprecated
```
zugaldia: This isn't removing the method, only the annotation.

tobrun: That is correct. There is actually some back story behind this deprecation.. At some point in time we introduced [versioned style methods](https://github.com/mapbox/mapbox-gl-native/issues/4702). This introduced the `@Deprecated` annotation to `setStyleUrl`. Not much after that, this was reverted in https://github.com/mapbox/mapbox-gl-native/issues/5193 but the deprecated annotation was not removed. I'm definitely open to revisit the decisions made in these issues, I actually don't have a strong opinion on this,  but I think removing the annotation in this PR reflects the current `Style.java` implementation.

zugaldia: >  I think removing the annotation in this PR reflects the current Style.java implementation

In that case, let's leave like you put it.

PR #7770, opened 2017-01-18
[core] Evict cached resources and tiles equally by access time

### LINKED ISSUE #7755: [ios, core] Cached resources evicted prematurely
**Platform:** iOS
**Mapbox SDK version:** 3.4.0-beta.7

### Expected behavior
Maps should immediately and consistently load cached styles + tiles when a device is in airplane mode, even if the cached data might be beyond the `Expires` header timestamp.  

### Actual behavior
App calls `-[MGLMapView setStyleURL:]`, and despite the app having loaded a map style + tiles recently (multiple times in the past 24 hours), the map disappears and the user is presented with a blank black or grey screen.

The only resolution has been to turn off airplane mode and reconnect to the internet (note: the style does not automatically reload once reachability is gained, the app must call `-[MGLMapView setStyleURL:]` to trigger a refresh).

This has happened multiple times in the wilderness where having a reliable map is absolutely essential. In these cases where no internet connection is available to reload the style, any navigation apps powered by Mapbox become useless.

### Steps to trigger behavior
This is difficult to replicate in isolation and it usually takes a day or so to appear since it seems related to expired cache headers. Might be able to fake some requests with short cache headers, and then restart a test app in airplane mode. Let me know if you have any ideas :/.

/cc @1ec5, @jfirebaugh

ericrwolfe:
Fixes #7755.

### DIFF
--- a/platform/default/mbgl/storage/offline_database.cpp
+++ b/platform/default/mbgl/storage/offline_database.cpp
@@ -793,6 +793,31 @@ bool OfflineDatabase::evict(uint64_t neededFreeSize) {
     // size, and because pages can get fragmented on the database.
     while (usedSize() + neededFreeSize + pageSize > maximumCacheSize) {
         // clang-format off
+        Statement accessedStmt = getStatement(
+            "SELECT max(accessed) "
+            "FROM ( "
+            "    SELECT accessed "
+            "    FROM resources "
+            "    LEFT JOIN region_resources "
+            "    ON resource_id = resources.id "
+            "    WHERE resource_id IS NULL "
+            "  UNION ALL "
+            "    SELECT accessed "
+            "    FROM tiles "
+            "    LEFT JOIN region_tiles "
+            "    ON tile_id = tiles.id "
+            "    WHERE tile_id IS NULL "
+            "  ORDER BY accessed ASC LIMIT ?1 "
+            ") "
+        );
+        accessedStmt->bind(1, 50);
+        // clang-format on
+        if (!accessedStmt->run()) {
+            return false;
+        }
+        Timestamp accessed = accessedStmt->get<Timestamp>(0);
+        
+        // clang-format off
         Statement stmt1 = getStatement(
             "DELETE FROM resources "
             "WHERE id IN ( "
@@ -800,10 +825,10 @@ bool OfflineDatabase::evict(uint64_t neededFreeSize) {
             "  LEFT JOIN region_resources "
             "  ON resource_id = resources.id "
             "  WHERE resource_id IS NULL "
-            "  ORDER BY accessed ASC LIMIT ?1 "
+            "  AND accessed <= ?1 "
             ") ");
         // clang-format on
-        stmt1->bind(1, 50);
+        stmt1->bind(1, accessed);
         stmt1->run();
         uint64_t changes1 = db->changes();
 
@@ -815,10 +840,10 @@ bool OfflineDatabase::evict(uint64_t neededFreeSize) {
             "  LEFT JOIN region_tiles "
             "  ON tile_id = tiles.id "
             "  WHERE tile_id IS NULL "
-            "  ORDER BY accessed ASC LIMIT ?1 "
+            "  AND accessed <= ?1 "
             ") ");
         // clang-format on
-        stmt2->bind(1, 50);
+        stmt2->bind(1, accessed);
         stmt2->run();
         uint64_t changes2 = db->changes();
 


platform/default/mbgl/storage/offline_database.cpp
```diff
@@ -793,17 +793,42 @@ bool OfflineDatabase::evict(uint64_t neededFreeSize) {
     // size, and because pages can get fragmented on the database.
     while (usedSize() + neededFreeSize + pageSize > maximumCacheSize) {
         // clang-format off
+        Statement accessedStmt = getStatement(
+            "SELECT max(accessed) "
+            "FROM ( "
+            "    SELECT accessed "
```
jfirebaugh: Are you sure the inner query is parsed as suggested by the indentation, and not "(select from resources [unordered and unlimited]) UNION ALL (select from tiles ... ordery by ... limit 50)"?

platform/default/mbgl/storage/offline_database.cpp
```diff
@@ -793,17 +793,42 @@ bool OfflineDatabase::evict(uint64_t neededFreeSize) {
     // size, and because pages can get fragmented on the database.
     while (usedSize() + neededFreeSize + pageSize > maximumCacheSize) {
         // clang-format off
+        Statement accessedStmt = getStatement(
+            "SELECT max(accessed) "
```
jfirebaugh: Query plan for this:

```
2|0|0|SCAN TABLE resources USING COVERING INDEX resources_accessed
2|1|1|SEARCH TABLE region_resources USING COVERING INDEX region_resources_resource_id (resource_id=?)
3|0|0|SCAN TABLE tiles USING COVERING INDEX tiles_accessed
3|1|1|SEARCH TABLE region_tiles USING COVERING INDEX region_tiles_tile_id (tile_id=?)
1|0|0|COMPOUND SUBQUERIES 2 AND 3 (UNION ALL)
0|0|0|SEARCH SUBQUERY 1
```

This concerns me a bit: it's table scanning both `resources` and `tiles`. The scan is done on indexes that are ordered appropriately for the `ORDER BY ... LIMIT` clause, and the hope is that these scans will be run in parallel, and both aborted when the limit is reached. But the query plan doesn't really make clear whether that's the case or not.

Can you do some benchmarking over databases of a few different sizes (empty, just ambient use, large offline region), and see how query time scales empirically?

ericrwolfe: I had a similar concern, since as part of the union it seem it would need to load the entirety of each table before sorting.

The below query adds complexity but leverages the indexes within subqueries before the union:
```sql
SELECT max(accessed) 
FROM ( 
  SELECT *
  FROM (
    SELECT accessed 
    FROM resources 
    LEFT JOIN region_resources 
    ON resource_id = resources.id 
    WHERE resource_id IS NULL 
    ORDER BY accessed ASC LIMIT 50 
  )
  UNION ALL
  SELECT * 
  FROM ( 
    SELECT accessed 
    FROM tiles 
    LEFT JOIN region_tiles 
    ON tile_id = tiles.id 
    WHERE tile_id IS NULL 
    ORDER BY accessed ASC LIMIT 50 
  )
  ORDER BY accessed ASC LIMIT 50
);
```

And has the following query plan:
```
3|0|0|SCAN TABLE resources USING COVERING INDEX resources_accessed
3|1|1|SEARCH TABLE region_resources USING COVERING INDEX region_resources_resource_id (resource_id=?)
2|0|0|SCAN SUBQUERY 3
2|0|0|USE TEMP B-TREE FOR ORDER BY
5|0|0|SCAN TABLE tiles USING COVERING INDEX tiles_accessed
5|1|1|SEARCH TABLE region_tiles USING COVERING INDEX region_tiles_tile_id (tile_id=?)
4|0|0|SCAN SUBQUERY 5
4|0|0|USE TEMP B-TREE FOR ORDER BY
1|0|0|COMPOUND SUBQUERIES 2 AND 4 (UNION ALL)
0|0|0|SEARCH SUBQUERY 1
```

I've used sqlite's built in `.timer ON` to attempt to profile the queries, and the differences seem negligible for a full cache (50mb database, 1010 tiles, 9 resources) but no offline regions stored.
Query in commit: `Run Time: real 0.000 user 0.000155 sys 0.000076`
Query in comment: `Run Time: real 0.000 user 0.000207 sys 0.000120`

I'm not sure of a better way to profile sqlite here, and not sure if those query results are cached. Let me know if you have any ideas, couldn't find much.

platform/default/mbgl/storage/offline_database.cpp
ericrwolfe: See https://github.com/mapbox/mapbox-gl-native/issues/7755#issuecomment-273561347

platform/default/mbgl/storage/offline_database.cpp
jfirebaugh: That seems promising. It's pretty important that we don't regress eviction performance for larger databases too, so can you test both approaches with a large offline region as well?

ericrwolfe: Testing on a database with SF downloaded for offline: 153mb, 3300 tiles, 1284 resources

Query in commit, run 3x:
```
Run Time: real 0.014 user 0.002650 sys 0.002637
Run Time: real 0.002 user 0.001782 sys 0.000058
Run Time: real 0.002 user 0.001967 sys 0.000073
```

Query in comment:
```
Run Time: real 0.028 user 0.003179 sys 0.003182
Run Time: real 0.002 user 0.001882 sys 0.000116
Run Time: real 0.002 user 0.001802 sys 0.000082
```

jfirebaugh: 👍  Let's go with the simpler query then.

1ec5:
@ericrwolfe, can you update the Android, iOS, and macOS changelogs to reflect this change?

PR #7773, opened 2017-01-18
Pass ICU to libtool in package script

boundsj:
In https://github.com/mapbox/mapbox-gl-native/pull/6984 we started using [ICU](https://github.com/mapbox/mason/pull/275) in core. This causes linker errors in an iOS application that uses a static build of the Mapbox iOS SDK.

This passes ICU (via Mason) to libtool when the iOS static build is created so that the dependency is available when the Mapbox iOS SDK is used in a host iOS application. 

The ICU lib actually replaces libgeojson. On our build machines libgeojson was not actually getting passed to libtool since the path used was provided by {IOS_SDK_VERSION} (created with `xcrun --sdk iphonesimulator|iphoneos --show-sdk-version` was greater than the `8.0` value that is [currently hard coded in the mason.sh script](https://github.com/mapbox/mason/blob/master/mason.sh#L70). I'm still unclear about why `libgeojson` was added but it does seem safe to remove it.

~~To attempt to start to sync up with the mason.sh file, this also adds a `MASON_PLATFORM_VERSION` variable to the iOS package.sh script that is now hardcoded to 8.0 just like the mason.sh script. This is required to get things to link but is fragile and I'm unclear about how this dependency should actually be handled in the various scripts. I'm not sure that hard coding this value in mason itself is a good idea.~~

** edit **

As noted below, we ended up using the cmake cache to find the ICU archive and to avoid having the package script know about the mason and ICU versions that are already specified in the cmake config.

cc @1ec5 @ChrisLoer @jfirebaugh

### DIFF
--- a/platform/ios/scripts/package.sh
+++ b/platform/ios/scripts/package.sh
@@ -117,7 +117,7 @@ if [[ ${BUILD_FOR_DEVICE} == true ]]; then
             -o ${OUTPUT}/static/${NAME}.framework/${NAME} \
             ${LIBS[@]/#/${PRODUCTS}/${BUILDTYPE}-iphoneos/lib} \
             ${LIBS[@]/#/${PRODUCTS}/${BUILDTYPE}-iphonesimulator/lib} \
-            `find mason_packages/ios-${IOS_SDK_VERSION} -type f -name libgeojson.a`
+            `cmake -LA -N ${DERIVED_DATA} | grep MASON_PACKAGE_icu_LIBRARIES | cut -d= -f2`
 
         cp -rv ${PRODUCTS}/${BUILDTYPE}-iphoneos/${NAME}.bundle ${STATIC_BUNDLE_DIR}
     fi
@@ -157,7 +157,7 @@ else
         libtool -static -no_warning_for_no_symbols \
             -o ${OUTPUT}/static/${NAME}.framework/${NAME} \
             ${LIBS[@]/#/${PRODUCTS}/${BUILDTYPE}-iphonesimulator/lib} \
-            `find mason_packages/ios-${IOS_SDK_VERSION} -type f -name libgeojson.a`
+            `cmake -LA -N ${DERIVED_DATA} | grep MASON_PACKAGE_icu_LIBRARIES | cut -d= -f2`
 
         cp -rv ${PRODUCTS}/${BUILDTYPE}-iphonesimulator/${NAME}.bundle ${STATIC_BUNDLE_DIR}
     fi


1ec5:
A minimum deployment target of 8.0 is also defined [in CMake configuration](https://github.com/mapbox/mapbox-gl-native/blob/f6715694f918a884eda77151aa4fec05521b9536/platform/ios/config.cmake#L6). ios.xcodeproj itself specifies a [minimum deployment target of 7.0](https://github.com/mapbox/mapbox-gl-native/blob/531a1c3070a098b5b31082c0500f0a3ce6a500bb/platform/ios/ios.xcodeproj/project.pbxproj#L2283) that applies to the static target. (The dynamic target overrides the minimum deployment target to 8.0 because iOS 7.0 doesn’t support dynamic frameworks.) This all makes me wonder whether the SDK really does support deploying back to iOS 7.0 anymore. Unfortunately, I don’t have a good way to check.

Also related: mapbox/mason#290.

/cc @kkaefer

kkaefer:
Background on `libgeojson` is that it used to have a compiled component, but that changed in https://github.com/mapbox/mapbox-gl-native/pull/6494

kkaefer:
How is mapbox/mason#290 related to this?

@boundsj can we add a CI task that checks the static library for linking errors?

1ec5:
> How is mapbox/mason#290 related to this?

Not strictly related. I just figured it was worth mentioning in the context of the binaries we’re getting from Mason.

> can we add a CI task that checks the static library for linking errors?

We used to build the dynamic+static scheme on CI, but it got trimmed down as part of a push to speed up CI builds. I’d appreciate building dynamic+static again, in order to catch issues like #6194 sooner.

boundsj:
> I think we should move the framework bundle creation to the Xcode stage. http://stackoverflow.com/a/16492468 has a guide as to how to do it.

@kkaefer thanks for the suggestion. That SO post walks through creating a Bundle configured as a relocatable object file. Although this seems to allow for more flexibility with regards to symbol hiding, I'd prefer to defer that change for now and stick to something along the lines of our existing solution.

In e9d1138, I cleaned the up `package.sh` so that libtool is no longer tasked with combining the SDK and libicuuc. Instead, I configured the `static` target to link with libicuuc with a `LIBRARY_SEARCH_PATHS` that points to the mason location. The next step could be to update iOS's `config.cmake` to make these changes. I think this would be more future proof since the cmake system is aware of the exact path to libicuuc and the version in use. For example, it writes something like `mbgl_core_LINK_LIBRARIES = "/Users/boundsj/mapbox-gl-native/mason_packages/ios-8.0/icu/58.1/lib/libicuuc.a" "-lz"` to the generated `config.xcconfig` file.

boundsj:
Actually, @kkaefer @1ec5 what do you think of 1c72fc2? It reverts e9d1138 and uses libtool again to combine libicuuc in the fat binary. However, it uses cmake's cache to get the correct version of ICU. 

This fixes the problem and seems pretty reliable. Maybe we can use something like this approach for now so we can close this release blocker? We can ticket additional static build system changes (if required) and a future upgrade to CI to catch this sort of issue.

platform/ios/ios.xcodeproj/project.pbxproj
```diff
@@ -616,6 +616,7 @@
 		40CFA6501D787579008103BD /* MGLShapeSourceTests.mm */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.cpp.objcpp; name = MGLShapeSourceTests.mm; path = ../../darwin/test/MGLShapeSourceTests.mm; sourceTree = "<group>"; };
 		40EDA1BD1CFE0D4A00D9EA68 /* MGLAnnotationContainerView.h */ = {isa = PBXFileReference; fileEncoding = 4; lastKnownFileType = sourcecode.c.h; path = MGLAnnotationContainerView.h; sourceTree = "<group>"; };
 		40EDA1BE1CFE0D4A00D9EA68 /* MGLAnnotationContainerView.m */ = {isa = PBXFileReference; fileEncoding = 4; lastKnownFileType = sourcecode.c.objc; path = MGLAnnotationContainerView.m; sourceTree = "<group>"; };
+		40F066C41E52E89D002B1DB5 /* libicuuc.a */ = {isa = PBXFileReference; lastKnownFileType = archive.ar; name = libicuuc.a; path = "../../mason_packages/ios-8.0/icu/58.1/lib/libicuuc.a"; sourceTree = "<group>"; };
```
1ec5: Is this reference still needed?

boundsj: Thanks. I squashed all the previous commits into 2c9bcb78f8f7c1e1786ee641c4ff2703b8d05453 that only has the path change for libtool.

platform/ios/scripts/package.sh
```diff
@@ -117,7 +117,7 @@ if [[ ${BUILD_FOR_DEVICE} == true ]]; then
             -o ${OUTPUT}/static/${NAME}.framework/${NAME} \
             ${LIBS[@]/#/${PRODUCTS}/${BUILDTYPE}-iphoneos/lib} \
             ${LIBS[@]/#/${PRODUCTS}/${BUILDTYPE}-iphonesimulator/lib} \
-            `find mason_packages/ios-${IOS_SDK_VERSION} -type f -name libgeojson.a`
+            `cmake -LA -N ${DERIVED_DATA} | grep MASON_PACKAGE_icu_LIBRARIES | cut -d= -f2`
```
kkaefer: Instead of using this, we could also use `MASON_PLATFORM=ios MASON_PLATFORM_VERSION=8.0 ./scripts/mason.sh LIBRARIES icu VERSION 58.1`

boundsj: I like looking up the value in the cmake cache because it automatically gets the correct path that (I think) is a result of `mason_use(icu VERSION 58.1)` and `target_add_mason_package(mbgl-core PUBLIC icu)` in the `config.cmake` file in the iOS platform directory. This way, this package.sh script does not need to know anything about the platform (although that would not be bad), the mason platform version, or the version of ICU. That can all come from the result of cmake and its config.cmake.

Issue #7577, opened 2017-01-03
Deprecate style class methods


1ec5:
Per https://github.com/mapbox/mapbox-gl-native/pull/7571#pullrequestreview-14875673, we should mark any style class methods in MGLStyle as deprecated with no replacement. The older methods on MGLMapView have already been deprecated as of iOS SDK v3.4.0 and macOS SDK v0.3.0.

/cc @frederoni

1ec5:
Fixed in #7785 on the release-ios-v3.4.0 branch.

Issue #2875, opened 2015-10-30
Remove style class support


ljbade:
It is time to mark the style class APIs as deprecated in iOS and Android.

Mapbox Studio does not support creating styles with classes in them. None of the bundled styles uses them. Neither the iOS nor Android test apps/Sirius touch these APIs.

As far as I know they could be completely broken and we would not even notice.

We should first update the documentation to mark them as deprecated.

Eventually we can remove them entirely.

This does not require any changes to core GL at this stage. Basically will only affect the higher level OS specific wrappers.

This will allow us to close:
https://github.com/mapbox/mapbox-gl-native/issues/794
https://github.com/mapbox/mapbox-gl-native/issues/1744
https://github.com/mapbox/mapbox-gl-native/issues/2077

@mikemorris Does node still need to support style classes?

@1ec5 Are there any open iOS tickets this will close?

mikemorris:
Nope, Node doesn't use these APIs.

tobrun:
This will be picked up transitioning API from MapView to MapboxMap with #3145

jfirebaugh:
Let's hold on removing any features around this. The final status of style classes has not been decided yet.

tobrun:
I removed the public java API since it was crashing, the jni code is still there.

cammace:
@tobrun can we remove the Android tag from this ticket?

tobrun:
Don't think this is relevant anymore. Closing

PR #7776, opened 2017-01-18
Deprecate style methods

fabian-guerra:
According to #7577 and #2875 we should mark style methods and properties as deprecated because we will not support them on future releases.

The Android SDK has already deprecated same API.

This is a temporarily PR to branch release-ios-v3.4.0, this should be included in release-ios-v3.4.1

### DIFF
--- a/platform/darwin/src/MGLStyle.h
+++ b/platform/darwin/src/MGLStyle.h
@@ -361,7 +361,7 @@ static const NSInteger MGLStyleDefaultVersion = 9;
 /**
  Currently active style classes, represented as an array of string identifiers.
  */
-@property (nonatomic) NS_ARRAY_OF(NSString *) *styleClasses;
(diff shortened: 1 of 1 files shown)


platform/darwin/src/MGLStyle.h
```diff
@@ -370,14 +370,14 @@ static const NSInteger MGLStyleDefaultVersion = 9;
  @param styleClass The style class to query for.
  @return Whether the style class is currently active.
  */
-- (BOOL)hasStyleClass:(NSString *)styleClass;
+- (BOOL)hasStyleClass:(NSString *)styleClass __attribute__((deprecated("This method will be removed in a future release. Runtime styling is :sunglasses:")));
```
boundsj: lol. we should remove the reference to runtime styling and 😎  here

boundsj:
> This is a temporarily PR to branch release-ios-v3.4.0, this should be included in release-ios-v3.4.1

👍  we can point this to the release-ios-v3.4.1 branch when it is created

PR #7779, opened 2017-01-18
[Qt] Use system defaults CA

tmpsantos:
The CA certificate we are bundling will expire in Aug/2018.

### DIFF
--- a/platform/qt/resources/common.qrc
+++ b/platform/qt/resources/common.qrc
@@ -1,5 +1,6 @@
 <RCC>
     <qresource prefix="/">
+        <file alias="icon.png">../../../common/icon.png</file>
         <file>source1.geojson</file>
         <file>source2.geojson</file>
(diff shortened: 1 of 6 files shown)


platform/qt/resources/common.qrc
```diff
@@ -1,5 +1,6 @@
 <RCC>
     <qresource prefix="/">
+        <file alias="icon.png">../../../common/icon.png</file>
```
kkaefer: unrelated change

tmpsantos: Not really unrelated, this was on the same `qmapbox.qrc` file the ca-bundle used to be. I removed this `qmapbox.qrc` altogether and moved the icon to the `commom.qrc` shared by the examples, where the icon is used.

tmpsantos:
For the record I tested this with `https://badssl.com/` and it works fine.

PR #7781, opened 2017-01-19
[Qt] Fix build when using Qt namespaces

tmpsantos:
Fix when building with namespaces enabled.

http://wiki.qt.io/Qt_In_Namespace


tmpsantos:
It could have done by using `QT_BEGIN_NAMESPACE`, but I personally find that ugly. Since we don't export headers on QML it won't matter much not doing forward declarations.

PR #7785, opened 2017-01-19
Deprecate style api methods

fabian-guerra:
According to #7577 and #2875 we should mark style methods and properties as deprecated because we will not support them on future releases.

The Android SDK has already deprecated same API.

This is a temporarily PR to branch release-ios-v3.4.0, this should be included in release-ios-v3.4.1


boundsj:
Looks 👍  to me. Added `do not merge` until we have a release branch to put this.

1ec5:
To clarify, this is about the style _class_ API, not the style API (aka the runtime styling API).

boundsj:
@fabian-guerra I changed the milestone to be 3.5.0. I think you can go ahead and target this against the master branch and merge it.

Issue #7778, opened 2017-01-18
Port 180º + very acute line join fixes 


lbud:
https://github.com/mapbox/mapbox-gl-js/pull/4008 includes fixes for 180º and very acute line joins in miter and bevel joins, and updates line-join render tests to add lines with these angles. The new line-join tests currently fail on native as follows:

| test | actual | new expected | diff |
| ---- | ---- | ---- | ---- |
| bevel | ![actual](https://cloud.githubusercontent.com/assets/3170312/22086833/96d68102-dd8f-11e6-831b-b24d8123d348.png) | ![expected](https://cloud.githubusercontent.com/assets/3170312/22086987/6824c200-dd90-11e6-9b16-9cee79acc982.png) | ![diff](https://cloud.githubusercontent.com/assets/3170312/22086834/99e4e82a-dd8f-11e6-8766-46bc2ac383e5.png) |
| bevel-transparent | ![actual](https://cloud.githubusercontent.com/assets/3170312/22086845/a8ec3198-dd8f-11e6-84ea-3224e07b71df.png) | ![expected 1](https://cloud.githubusercontent.com/assets/3170312/22086986/681f5630-dd90-11e6-9ca2-10feb1c6e76c.png) | ![diff](https://cloud.githubusercontent.com/assets/3170312/22086844/a8ea53b4-dd8f-11e6-9f3e-f8fb6742690f.png) |
| default | ![actual](https://cloud.githubusercontent.com/assets/3170312/22086865/bf9a3ae8-dd8f-11e6-9f98-95424e15bee7.png) | ![expected 2](https://cloud.githubusercontent.com/assets/3170312/22086983/6816fd28-dd90-11e6-8dfe-e20cbddcbd19.png) | ![diff](https://cloud.githubusercontent.com/assets/3170312/22086866/bfa140fe-dd8f-11e6-9550-03198d4254b2.png) |
| miter | ![actual](https://cloud.githubusercontent.com/assets/3170312/22086875/cef511f2-dd8f-11e6-937c-fbbf375facb6.png) | ![expected 3](https://cloud.githubusercontent.com/assets/3170312/22086984/681838c8-dd90-11e6-967e-a1003e0bf2e9.png) | ![diff](https://cloud.githubusercontent.com/assets/3170312/22086876/cef83832-dd8f-11e6-8550-44407cc995af.png) |
[...]

PR #7786, opened 2017-01-19
Fix rendering 180° line joins

kkaefer:
Fixes https://github.com/mapbox/mapbox-gl-native/issues/7778, a alternative fix for https://github.com/mapbox/mapbox-gl-js/pull/4008

@lbud  I am using a different fix: I found that computing the unit vector of (0, 0) on native [returns a vector of (0, 0)](https://github.com/mapbox/mapbox-gl-native/commit/768c259#commitcomment-20538139), while JS uses NaN. I investigated a little further and ultimately found that this makes `miterLength` NaN as well on JS. Instead of using NaN all the way down, this patch uses an explicit value of `0`, and sets the `miterLimit` to Infinity in that case. This change also fixes the missing line caps we saw on native (which was also a result of the different way of computing `unit(0, 0)`).

I think we should port this change back to JS as well, since it removes division by zero.

### DIFF
--- a/src/mbgl/programs/line_program.hpp
+++ b/src/mbgl/programs/line_program.hpp
@@ -49,7 +49,7 @@ struct LineAttributes : gl::Attributes<
                 static_cast<int16_t>((p.y * 2) | t.y)
             },
             {
-                // add 128 to store an byte in an unsigned byte
+                // add 128 to store a byte in an unsigned byte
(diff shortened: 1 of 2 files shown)


kkaefer:
Enabled test suite in https://github.com/mapbox/mapbox-gl-js/commit/2663093c9d7f0cba6235131fb9b9f30328a76603

PR #7793, opened 2017-01-20
Upgrade Android build tools to 25.0.2

### LINKED ISSUE #7763: Android CI builds failing on release-ios-v3.4.0 branch
Android builds on Bitrise began failing on the release-ios-v3.4.0 branch (but not master), due to [a Bitrise VM update](https://discuss.bitrise.io/t/weekly-virtual-machine-stack-updates-2017-01-14/139), specifically bitrise-docker/android#55. For example, [this build](https://www.bitrise.io/build/6e0df94dbb89dd82) from #7759 failed with:

```
AWS Device Farm configuration is NOT VALID

FAILURE: Build failed with an exception.

* What went wrong:
A problem occurred configuring project ':MapboxGLAndroidSDKWearTestApp'.
> failed to find Build Tools revision 24.0.2
```

As a stopgap, since we’re very close to an iOS SDK release, I put in some triggers that diverted the Android build to a no-op workflow that always passes for branches pushed by @mapbox/ios. However, the triggers are based on branch name patterns, so for example I couldn’t find a good way to exempt @incanus from Android builds. Not that that would be a great idea: we’re starting to develop on master again, which means we’ll need Android builds to run; meanwhile, we still need to keep the release-ios-v3.4.0 branch alive for v3.4.1. We need to find a way to make these builds run for real but continue to pass.

/cc @mapbox/ios @mapbox/android

1ec5:
Cherry-picked 9a635672a6251f05b864efc71214046f9ea4acff (#7729) into the release-ios-v3.4.0 branch to staunch Android CI failures.

Fixes #7763.

/cc @friedbunny @zugaldia


1ec5:
Note that I removed the shim in https://github.com/mapbox/mapbox-gl-native/issues/7763#issuecomment-273389355 before running the Android build, which passed thanks to this change.

PR #7800, opened 2017-01-20
[core][Qt] Use Qt for UTF16, since `codecvt` is not always available

tmpsantos:
Instead, use Qt's codec.

### DIFF
--- a/platform/qt/src/string_stdlib.cpp
+++ b/platform/qt/src/string_stdlib.cpp
@@ -9,13 +9,13 @@ namespace mbgl {
 namespace platform {
 
 std::string uppercase(const std::string& str) {
-    auto upper = QString::fromUtf8(str.c_str()).toUpper().toUtf8();
+    auto upper = QString::fromUtf8(str.data(), str.length()).toUpper().toUtf8();
 
     return std::string(upper.constData(), upper.size());
 }
 
 std::string lowercase(const std::string& str) {
-    auto lower = QString::fromUtf8(str.c_str()).toLower().toUtf8();
+    auto lower = QString::fromUtf8(str.data(), str.length()).toLower().toUtf8();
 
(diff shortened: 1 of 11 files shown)


platform/qt/config.cmake
```diff
@@ -50,6 +50,12 @@ macro(mbgl_platform_core)
         target_sources(mbgl-core PRIVATE platform/qt/src/bidi.cpp)
     endif()
 
+    if(WITH_QT_4)
+        target_sources(mbgl-core PRIVATE platform/default/utf.cpp)
+    else()
+        target_sources(mbgl-core PRIVATE platform/qt/src/utf.cpp)
```
tmpsantos: `QString::fromStdString(...).toStdU16String();` is only available for Qt5+

platform/qt/src/string_stdlib.cpp
```diff
@@ -9,13 +9,13 @@ namespace mbgl {
 namespace platform {
 
 std::string uppercase(const std::string& str) {
```
tmpsantos: @brunoabinader I added this change after you reviewed. Do you mind having a look again?

tmpsantos: The context is we should never use c_str() for multibyte.

brunoabinader: > The context is we should never use c_str() for multibyte.

Interesting - I wasn't aware of that! 👍

ChrisLoer: > The context is we should never use c_str() for multibyte.

That is interesting -- I thought the only difference was that `data()` could leave off null termination (and probably for most of our compiler targets it null terminates anyway, which makes it easy to introduce sneaky bugs depending on the null termination).

PR #7808, opened 2017-01-21
Note that canvas sources are unsupported

1ec5:
mapbox/mapbox-gl-style-spec#656 added a `canvas` source type that only makes sense in a webpage context (but see mapbox/mapbox-gl-native#7471). This PR points out that canvas sources are unsupported in the iOS and macOS SDKs’ “Information for Style Authors” guide. Whereas most of the runtime styling API is generated automatically, the source API is maintained by hand, since each source type requires a unique set of initializers and methods.

Also fixed a typo in the same guide.

/cc @lbud @lucaswoj

### DIFF
--- a/platform/darwin/docs/guides/For Style Authors.md.ejs
+++ b/platform/darwin/docs/guides/For Style Authors.md.ejs
@@ -282,9 +282,15 @@ Array (`-padding`) | `NSValue.edgeInsetsValue` | `NSValue.edgeInsetsValue`
 <% } -%>
 
 For padding attributes, note that the arguments to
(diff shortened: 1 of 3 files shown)


lucaswoj:
1e97815320ce3f7a56b555712733c246e0c8a5fe looks good 👍 

Not so sure what to make of b2c27650d0dd6b699f7e891fd22a11d8c9688426

1ec5:
> Not so sure what to make of b2c27650d0dd6b699f7e891fd22a11d8c9688426

Just something I noticed along the way. Pay no mind. 😉

Issue #6238, opened 2016-09-01
Snapshot blocks UI thread for 800ms


mkv123:
**Platform:** Android
**Mapbox SDK version:** Latest 4.2 snapshot
### Steps to trigger behavior
1. mapboxMap.snapshot(new SnaphshotReadyCallback() { public void onSnapshotReady(Bitmap snapshot){} } );
### Expected behavior

Snapshot is taken while application usage remains silky smooth
### Actual behavior

All animations etc stop for a significant amount of time (800ms on a high end phone, Nexus 5X).

Alternatively if I try to excute the snapshot() from a non-ui thread I get a crash.

Test code.

``` java
final long start = SystemClock.currentThreadTimeMillis();

mapboxMap().snapshot(new SnapshotReadyCallback() {
    @Override
    public void onSnapshotReady(Bitmap snapshot) {
        Log.d(TAG, "onSnapshotReady: " + (SystemClock.currentThreadTimeMillis() - start));
    }
}); 
```

This makes using the snapshotting pretty useless as there are very few cases where blocking the UI thread for such a long time would be ok.

jfirebaugh:
Fixed in #7828.

PR #7318, opened 2016-12-07
added text changing to symbol example

cammace:
Added a text changing example to the symbol activity.

ezgif com-video-to-gif 1

cc: @ivovandongen

PR #7828, opened 2017-01-24
Android: use system-provided image decoding + encoding

### LINKED ISSUE #7329: Android: Replace libpng/libjpeg-turbo/libwebp with BitmapFactory
In https://github.com/mapbox/mapbox-gl-native/issues/2354, we captured that we want to replace these libs with a Java-based version. Here's how this will work:

* [ ] Refactor `util::Image` to hide its implementation, in particular the `data` member
* [ ] When compiling for Android, add an additional constructor that takes `JNIEnv *env, jobject jbitmap`, and then uses [`libjnigraphics`](https://developer.android.com/ndk/reference/group___bitmap.html) to expose the bitmap information, and handle JNI object lifetime
* [ ] Create an Android-specific implementation of `util::decodeImage()` that uses the JNI to call [BitmapFactory::decodeByteArray](https://developer.android.com/reference/android/graphics/BitmapFactory.html#decodeByteArray(byte[],%20int,%20int,%20android.graphics.BitmapFactory.Options)) for decompressing + premultiplying the image
* [ ] Create an Android-specific version of `util::encodePNG()` that uses the JNI/Bitmap/BitmapFactory to create a Java Bitmap + compress it to PNG

/cc @mapbox/android

kkaefer:
Another take at #7329.

- Removes `libjpeg-turbo` and `libpng` in favor of calls to [`BitmapFactory.decodeByteArray`](https://developer.android.com/reference/android/graphics/BitmapFactory.html#decodeByteArray(byte%5B%5D,%20int,%20int,%20android.graphics.BitmapFactory.Options)). We're doing a bunch of copies, but I don't think we can avoid any of them. I implemented a version that used `java.nio.ByteBuffer` and supplying the pointer address directly, but Java can't generate a `byte[]` required for `decodeByteArray`. We could alternatively implement a `NativeInputStream` (similar to the `NativeOutputStream` contained in this PR) and use [`BitmapFactory.decodeStream`](https://developer.android.com/reference/android/graphics/BitmapFactory.html#decodeStream(java.io.InputStream)) to avoid one copy. We are doing a second copy from the decoded Bitmap object to memory we manage ourselves (using [`libjnigraphics`](https://developer.android.com/ndk/reference/group___bitmap.html). Instead, we could completely replace `util::Image` for android builds with a version that keeps around the `Bitmap` object and keeps the pixels locked so they don't go away.
- Removes indirection for the snapshot functionality: Previously, we were compressing to a PNG, only to decompress to a `Bitmap` object right away. Instead, we're now directly converting the raw C++ `PremultipliedImage` to a Java `Bitmap` object. This reduces the snapshotting time on my test device from ~550ms to ~80ms.

Build size for Release/arm-v8 drops from 5620 KB to 5259 KB.

Fixes #7329, #6238.

### DIFF
--- a/platform/android/src/test/main.jni.cpp
+++ b/platform/android/src/test/main.jni.cpp
@@ -1,17 +1,11 @@
 #include "../jni.hpp"
 
-#include <android/log.h>
-#include <jni/jni.hpp>
 #include <jni/jni.hpp>
 
 #include <mbgl/util/logging.hpp>
 #include <mbgl/test.hpp>
 
 #include <vector>
-
-#pragma clang diagnostic ignored "-Wunused-parameter"
-
-#define MAKE_NATIVE_METHOD(name, sig) jni::MakeNativeMethod<decltype(name), name>( #name, sig )
 
 namespace {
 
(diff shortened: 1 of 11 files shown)


kkaefer:
Implementing a native InputStream also requires a copy.

jfirebaugh:
Decoding part looks good. For the encoding part, I like the solution in #7813 better, given that it is only used in tests.

kkaefer:
@jfirebaugh yeah, plan is to get #7318 in first, then remove encoding from this PR.

kkaefer:
Read for review

platform/android/src/test/main.jni.cpp
```diff
@@ -48,54 +42,35 @@ struct Main {
...
-}
-
-static jint logger_entry_max_payload_native(JNIEnv* env, jni::jobject* clazz) {
-    return static_cast<jint>(4068);
-}
-
+// We're declaring the function, which is libandroid_runtime.so.
+// It is defined in AndroidRuntime.cpp:
+// https://github.com/android/platform_frameworks_base/blob/master/core/jni/AndroidRuntime.cpp
+// Once this symbol goes away, we'll have to revisit.
```
jfirebaugh: If this symbol goes away, the Mapbox GL SDK will start failing, right? It would be worse than a build time failure -- existing installations might stop working after an OS or software upgrade.

If this is not a public API, let's find a different way.

kkaefer: This file is not part of the SDK; it is only used to bootstrap the core unit tests (with Google Test). If this symbol went away, the only thing that would break is launching the unit tests, but it does not impact in any way the function of a deployed version of Mapbox GL.

I've checked the past couple of Android versions and this symbol remained stable from (at least) 4.4 to 7.0.

PR #7829, opened 2017-01-24
[Qt] Refactor QMapbox Annotations API

brunoabinader:
Add new types in `QMapbox` namespace to support the new `geometry.hpp`-based geometries.

### DIFF
--- a/platform/qt/src/qmapboxgl.cpp
+++ b/platform/qt/src/qmapboxgl.cpp
@@ -17,8 +17,10 @@
 #include <mbgl/style/transition_options.hpp>
 #include <mbgl/sprite/sprite_image.hpp>
 #include <mbgl/storage/network_status.hpp>
+#include <mbgl/util/color.hpp>
 #include <mbgl/util/constants.hpp>
 #include <mbgl/util/geo.hpp>
+#include <mbgl/util/geometry.hpp>
 #include <mbgl/util/run_loop.hpp>
 #include <mbgl/util/traits.hpp>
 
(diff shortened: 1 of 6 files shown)


platform/qt/src/qmapboxgl.cpp
```diff
@@ -848,42 +836,114 @@ void QMapboxGL::setTransitionOptions(qint64 duration, qint64 delay) {
     d_ptr->mapObj->setTransitionOptions(mbgl::style::TransitionOptions{ convert(duration), convert(delay) });
 }
 
-mbgl::Annotation fromPointAnnotation(const PointAnnotation &pointAnnotation) {
-    const Coordinate &coordinate = pointAnnotation.first;
-    const QString &icon = pointAnnotation.second;
-    return mbgl::SymbolAnnotation { mbgl::Point<double> { coordinate.second, coordinate.first }, icon.toStdString() };
-}
+mbgl::ShapeAnnotationGeometry asMapboxGLGeometry(const QMapbox::ShapeAnnotationGeometry &geometry) {
+    auto asMapboxGLLineString = [&](const QMapbox::Coordinates &lineString) {
```
tmpsantos: My suggestion here is to move these conversion functions to `platform/qt/src/qt_conversion.hpp` and rename them to `toSomething` which is the pattern we use there.

brunoabinader: @tmpsantos let's do that on a separate PR, which also involves moving e.g. `fromStringList` and `toSpriteImage` which do not belong to this change.

PR #7834, opened 2017-01-24
macOS: Allow use of integrated GPU

kkaefer:
On computers that have more than one GPU, macOS automatically switches to the discrete one if the app doesn't opt in to also using the integrated GPU: https://developer.apple.com/library/content/qa/qa1734/_index.html


1ec5:
Cherry-picking this change into the release-ios-v3.4.0 branch (from which v0.3.1 will be released) in #7861.

PR #7816, opened 2017-01-23
Document how to avoid gesture recognizer conflicts

nitrag:
Adding Gesture info as discussed: https://github.com/mapbox/mapbox-gl-native/issues/2278#issuecomment-273685832

Not sure if using codeblocks is kosher.

PR #7835, opened 2017-01-24
Test MGLMapView gesture recognizer example; refer to Night Shift on macOS

1ec5:
Added the examples from #7816 to the unit tests to fix `make darwin-update-examples` and `make style-code-darwin`.

Also include a mention of Night Shift in the macOS version of “Information for Style Authors”, now that macOS Sierra 10.12.4 beta includes Night Shift.


1ec5:
Yeah, a test would be pretty meaningless without XCUITest or KIF.

Issue #7845, opened 2017-01-25
Printing is disabled


1ec5:
@kkaefer pointed out that File ‣ Print is always disabled in macosapp. I’m also unable to get anything hooked up to the `print:` action to light up in a standalone application linked against macOS SDK v0.3.0. This feature obviously worked in #4962, and I’m pretty sure it worked just a few months ago.

1ec5:
This only seems to be an issue in terms of menu item or toolbar validation. If you call `-[MGLMapView print:]` manually, the print dialog does appear.

kkaefer:
Short term fix merged in #7848, but as @1ec5 notes, we should resolve the issues in the responder chain.

PR #7848, opened 2017-01-25
Enable printing on macOS

kkaefer:
Short term fix for #7845.

Longer term, we should have a look at the [correct way of printing documents](https://developer.apple.com/library/content/documentation/Cocoa/Conceptual/Printing/osxp_printapps/osxp_printapps.html#//apple_ref/doc/uid/20000861-BAJBFGED)

### DIFF
--- a/platform/macos/app/Base.lproj/MainMenu.xib
+++ b/platform/macos/app/Base.lproj/MainMenu.xib
@@ -135,7 +135,7 @@
                             </menuItem>
                             <menuItem title="Print…" keyEquivalent="p" id="aTl-1u-JFS">
                                 <connections>
-                                    <action selector="print:" target="-1" id="qaZ-4w-aoO"/>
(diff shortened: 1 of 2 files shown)


platform/macos/app/Base.lproj/MainMenu.xib
```diff
@@ -135,7 +135,7 @@
                             </menuItem>
                             <menuItem title="Print…" keyEquivalent="p" id="aTl-1u-JFS">
                                 <connections>
-                                    <action selector="print:" target="-1" id="qaZ-4w-aoO"/>
+                                    <action selector="print:" target="Voe-Tx-rLC" id="rXg-Pd-AG8"/>
```
1ec5: This suggests that the map view isn't in the responder chain somehow, which is unexpected. Locally, I hooked the map view up to the window's initialFirstResponder, but that didn't seem to be sufficient. I wonder if the recently added Layers table view is stealing keyboard focus.

kkaefer:
Merging this as a short term fix so that I can test printing.

Issue #1490, opened 2015-05-09
Annotation callout disables user location tracking


1ec5:
1. Turn on user location tracking.
2. While driving down a highway, select an annotation near your current position.
   - User tracking turns off.
3. Turn user tracking back on.
   - The annotation is deselected as the map pans to the new user location.
   - Then user tracking quickly turns back off as the annotation reselects itself.

incanus:
Is this still valid @1ec5?

1ec5:
Yes.

RomainQuidet:
I don't see this behavior anymore, the tracking back on keep user tracking.

friedbunny:
In `ios-app`, longpressing to add an annotation will disable the tracking mode. Tapping on an annotation will also disable tracking — e.g., tapping on the user dot.

1ec5:
What I'm seeing now is that selecting an ordinary annotation turns off tracking, but selecting the user dot does not.

1ec5:
Fixed in #7856.

PR #7856, opened 2017-01-25
Persist user tracking mode despite programmatic camera changes

### LINKED ISSUE #7789: Manipulate camera pitch without leaving .followWithCourse
**Platform: iOS**
**Mapbox SDK version: 3.4.0-rc.1**

### Steps to trigger behavior

 1. Enter tracking mode `.followWithCouse`
 2. Change camera pitch 

### Expected behavior

Tracking mode remains in `.followWithCourse`

### Actual behavior

Tracking mode enters `.none`

--------------
I'm open to extending this beyond manipulating `pitch`. I can think of use cases for `heading` as well.

/cc @frederoni @boundsj @1ec5 @pveugen

frederoni:
Fixes #7789 

The stuttering effect is not as noticeable as I first thought. Concurrent animations (#3625) might still be a blocker though.

animation

The transition is more or less smooth depending on the timing of the subsequent location update.

@bsudekum 👀

### DIFF
--- a/platform/ios/CHANGELOG.md
+++ b/platform/ios/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 Mapbox welcomes participation and contributions from everyone. Please read [CONTRIBUTING.md](../../CONTRIBUTING.md) to get started.
 
+## 3.4.2
+
+* Programmatic camera changes no longer resets user tracking mode. ([#7856](https://github.com/mapbox/mapbox-gl-native/pull/7856))
+
 ## 3.4.1 
 
 * Fixed a build error in the static framework flavor of this SDK caused by a missing header. ([#7844](https://github.com/mapbox/mapbox-gl-native/pull/7844))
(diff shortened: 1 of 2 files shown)


bsudekum:
@frederoni nice! Are there any other camera changing function that this should also be removed from?

I like @1ec5 summed it up:

> In other words, we should avoid kicking the user out of user tracking mode due to a programmatic camera change, but still kick the user out of user tracking mode in response to a user gesture. This would fix #1490 as well.

frederoni:
All camera updating functions are chained so this change will impact all of them.

> In other words, we should avoid kicking the user out of user tracking mode due to a programmatic camera change, but still kick the user out of user tracking mode in response to a user gesture. This would fix #1490 as well.

This PR would fix that.

1ec5:
What about [`-flyToCamera:withDuration:peakAltitude:completionHandler:`](https://github.com/mapbox/mapbox-gl-native/blob/f5bcf186f63f0f10642dcee14a680667c36cbfdb/platform/ios/src/MGLMapView.mm#L2571)?

frederoni:
@1ec5 Good catch. Fixed in https://github.com/mapbox/mapbox-gl-native/pull/7856/commits/8f79fea6191f4c549dcb34e5224fab367118b7cf

PR #7812, opened 2017-01-25
macOS: Don't round-trip still images through encoding

kkaefer:
When calling `encodePNG`, on the bytes of the textures, only to decompress it in the very next line: https://github.com/mapbox/mapbox-gl-native/blob/f1c3cfb7e2d46a4563d680bcab888fa01b2ca160/platform/macos/src/MGLMapView.mm#L790-L792

Instead, we can just use the data we already have to initialize the `NSImage` without going through the compression/decompression routine.

### DIFF
--- a/include/mbgl/util/image.hpp
+++ b/include/mbgl/util/image.hpp
@@ -48,6 +48,13 @@ public:
         return size && data.get() != nullptr;
     }
 
+    template <typename T = Image>
+    T clone() const {
+        T copy(size);
+        std::copy(data.get(), data.get() + bytes(), copy.data.get());
+        return copy;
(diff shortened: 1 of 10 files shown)


1ec5:
Looks like some non-SDK targets need to be updated, though:

```
▸ Linking mbgl-test

❌  ld: symbol(s) not found for architecture x86_64



❌  clang: error: linker command failed with exit code 1 (use -v to see invocation)


** BUILD FAILED **


The following build commands failed:
	Ld build/macos/Debug/mbgl-test normal x86_64
(1 failure)
make: *** [test] Error 65
```

Issue #7256, opened 2016-12-01
MGLMapView should be non-opaque by default


1ec5:
In https://github.com/mapbox/mapbox-gl-native/issues/3101#issuecomment-263752036, @mxcl argues that MGLMapView’s `isOpaque` property should be set to `NO` by default on iOS. In #3096, I originally defaulted `isOpaque` to `NO` but switched it back to `YES` out of concern that performance might be impacted. Assuming the flash of black is still a problem, I’m in favor of making the map view non-opaque by default. On macOS, we render the map in an NSOpenGLLayer, which like all CALayers is transparent by default, and I haven’t noticed any problems stemming from that choice.

Alternatively, we could mitigate the flash without impacting performance by defaulting `isOpaque` to `NO` at initialization, then switching it to `YES` as soon as the style finishes loading for the first time (unless the developer has manually set `isOpaque` in the meantime).

/cc @incanus @frederoni

friedbunny:
I’m also in favor of moving ahead with the switch to `MGLMapView.opaque = NO`. Being that this  change that would be trivial to revert in a patch release, would there be any issues with putting this on the v3.4.0 or v3.4.1 milestones?

1ec5:
I’d be fine with either v3.4.0 or v3.4.1. Any PR that changes the default should come with profiles comparing performance before and after (just due diligence, nothing extremely detailed).

friedbunny:
Certainly we should run iosbench, but I suspect the results will be inconclusive as to the absolute effect — last time I ran it, we were maxing out at 60 FPS consistently. Perhaps an impetus to expand iosbench.

incanus:
> last time I ran it, we were maxing out at 60 FPS consistently

Hmm, this is in conflict with #5489. Are we sure about this, or am I misunderstanding what we're talking about?

friedbunny:
@incanus [iosbench](https://github.com/mapbox/mapbox-gl-native/blob/2cf238f74eef021b5dd097822ceafc6da83cf4d3/platform/ios/benchmark/MBXBenchViewController.mm) doesn’t stress the map terribly these days — it doesn’t add any sort of annotation and focuses almost solely on core rendering performance. The goal has always been to expand the range of tests in iosbench, and perhaps this is a good opportunity.

incanus:
Right, but my point is that even bare, we have not been hitting 60fps, hence the annotation sync problem root. Are we now reliably doing that?

[...]

PR #7859, opened 2017-01-25
Make MGLMapView non opaque by default

fabian-guerra:
According to #7256 to solve "flash of black" one alternative is set `MGLMapView.opaque = NO` but one concern was how this may affect the performance. Changing `opaque = NO` resulted in a reduction in FPS from an average of 50.63 to 49.61. This is unnoticeable but we can keep tracking this in case something comes up.

(Benchmark tests were done using bench app)

### DIFF
--- a/platform/ios/CHANGELOG.md
+++ b/platform/ios/CHANGELOG.md
@@ -6,6 +6,7 @@ Mapbox welcomes participation and contributions from everyone. Please read [CONT
 
 * Fixed a build error in the static framework flavor of this SDK caused by a missing header. ([#7844](https://github.com/mapbox/mapbox-gl-native/pull/7844))
 * Fixed an issue causing MGLMapView’s `camera`’s `heading` to be set to a negative value, indicating an undefined heading, when the map view faces northwest. The heading is now wrapped to between zero and 360 degrees, for consistency with MGLMapView’s `direction` property. ([#7724](https://github.com/mapbox/mapbox-gl-native/pull/7724))
+* Fixed an issue causing MGLMapView flash of black. ([#7859](https://github.com/mapbox/mapbox-gl-native/pull/7859))
 
 ## 3.4.0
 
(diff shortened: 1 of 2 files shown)


friedbunny:
When you open a PR to fix an issue, please mention that issue number: #7256 in this case.

GitHub also allows you to [automatically close](https://help.github.com/articles/closing-issues-via-commit-messages/) an issue simply by including “Fixes #7256” (or similar). GitHub suggests doing this in the commit message, but that leads to a lot of noise if you end up rebasing — better to do it in the PR body.

boundsj:
We should add notes to the ios and macos changelogs.

1ec5:
The Android build is failing due to #7763. It looks like you branched off d829570cca4e2df98902236cb6d3e62dabddf487. Please rebase against a current release-ios-v3.4.0, which will also give you a proper place to put in a changelog blurb.

platform/macos/CHANGELOG.md
```diff
@@ -5,6 +5,7 @@
 This version of the Mapbox macOS SDK corresponds to version 3.4.1 of the Mapbox iOS SDK. The two SDKs have very similar feature sets. The main differences are the lack of user location tracking and annotation views. Some APIs have been adapted to macOS conventions, particularly the use of NSPopover for callout views.
 
 * Fixed an issue causing MGLMapView’s `camera`’s `heading` to be set to a negative value, indicating an undefined heading, when the map view faces northwest. The heading is now wrapped to between zero and 360 degrees, for consistency with MGLMapView’s `direction` property. ([#7724](https://github.com/mapbox/mapbox-gl-native/pull/7724))
+* Fixed an issue causing MGLMapView flash of black. ([#7859](https://github.com/mapbox/mapbox-gl-native/pull/7859))
```
1ec5: This PR doesn’t change anything about the macOS SDK, since MGLMapView has separate implementations on iOS and macOS. The behavior is already correct on macOS, so you can remove this blurb.

PR #7875, opened 2017-01-26
Check whitespace on CI

kkaefer:
We have a lot of trailing whitespace + missing newlines that make diffing hard.

This adds a check on the CI for these things.


kkaefer:
@mapbox/gl This change mean that CI will fail if we have trailing whitespace or missing trailing newlines. To avoid this, you can run `make codestyle` to automatically fix this.

1ec5:
Odd that the iOS and macOS SDKs builds are now nitpicky about whitespace, yet this part of the codebase remains terribly inconsistent about braces, line wrapping, and spaces before angle brackets, which to me are much more visible inconsistencies.

kkaefer:
Whitespace changes are much more subtle and less noticeable. Many systems also automatically trim trailing whitespace (e.g. git in some configurations, and many editors), while the same isn't true for non-whitespace symbols.

1ec5:
Xcode adds indentation to blank lines by default (for example, when pressing Return twice), which now fails the build.

jfirebaugh:
Yeah, this seems like an unnecessary hassle to me. Trailing whitespace is harmless.

kkaefer:
This was removed in #7891

PR #7884, opened 2017-01-27
Update and cleanup ca-bundle.crt

kkaefer:
Our current version of the certificate bundles is pretty outdated. It's only used on Linux for non-node.js builds, but we should still update it.


kkaefer:
Added a script `scripts/update_ca_bundle.sh` that automatically downloads + updates the CA bundle. We should run this occasionally.

kkaefer:
Also removes it from the Android build, since we're no longer using it since switching to OkHTTP.

PR #7886, opened 2017-01-27
Don't add trailing spaces in generate-style-code.js

kkaefer:
Our `indent` function adds trailing spaces for multi line strings when generating the code. Follow up to #7875


1ec5:
I actually disagreed with #7875. In my opinion, enforcing a code style that differs from Xcode's default only increases friction for new contributors with no material gain. The "no trailing spaces" rule is now the only code style rule we enforce for the iOS and macOS SDKs, which seems a bit strange. A good compromise would have been to enforce the rule only for non-blank lines, or only in components where we actually care.

kkaefer:
Superseded by https://github.com/mapbox/mapbox-gl-native/pull/7372

PR #7895, opened 2017-01-29
[ios] Stop including framework stripping script in built product

friedbunny:
This self-deletes `strip-frameworks.sh` during the build script phase, after it’s done doing its deed. Without this trap, the script is distributed to users in released apps in Mapbox.framework.

Does not address the CocoaPods use-case, where this script is unnecessary and not executed, but still ultimately included in the built app.

/cc @boundsj


friedbunny:
Brought in https://github.com/realm/realm-cocoa/commit/cb8e3b643b7b0fe54bef5399d05a7bbbf8c287ef and https://github.com/realm/realm-cocoa/commit/453e0c1bd71f629748bf89afb6c0af28403b4492, per @1ec5’s request.

PR #7316, opened 2016-12-07
Simplified Chinese localization

YunjieLi:
Per chat w/ @1ec5 adding the Chinese localization.

PR #7899, opened 2017-01-30
Traditional Chinese localization

### LINKED ISSUE #7370: Traditional Chinese localization
Now that #7316 has landed, it should be pretty straightforward to implement a Traditional Chinese localization based on the Simplified one. As a starting point, we can run each .strings file through [OpenCC](https://github.com/BYVoid/OpenCC/):

```bash
$ brew install opencc
$ echo 汉语 | opencc -c s2t
漢語
```

/cc @YunjieLi @friedbunny

1ec5:
Added Traditional Chinese localizations to the iOS and macOS SDKs, using [OpenCC](https://github.com/BYVoid/OpenCC/) to convert the Simplified Chinese localizations added in #7316 and #7503. ([Instructions](https://github.com/mapbox/mapbox-gl-native/blob/master/platform/ios/DEVELOPING.md#adding-a-localization))

I spot-checked the results but would appreciate a review by @lily-chai, @virginiayung, or @jenyuhk. Also, if any of these strings could have more idiomatic translations, we could add Taiwan- or Hong Kong–specific localizations, [again using OpenCC](https://github.com/BYVoid/OpenCC/#預設配置文件).

Fixes #7370.

/cc @YunJieLi @friedbunny

### DIFF
--- a/platform/ios/resources/zh-Hant.lproj/Localizable.strings
+++ b/platform/ios/resources/zh-Hant.lproj/Localizable.strings
@@ -1,0 +1,84 @@
+/* Accessibility hint */
+"ANNOTATION_A11Y_HINT" = "顯示信息";
+
+/* No comment provided by engineer. */
+"API_CLIENT_400_DESC" = "The session data task failed. Original request was: %@";
+
+/* No comment provided by engineer. */
+"API_CLIENT_400_REASON" = "The status code was %ld";
+
+/* No comment provided by engineer. */
+"CANCEL" = "取消";
+
+/* Accessibility hint */
+"COMPASS_A11Y_HINT" = "旋轉地圖使正北朝上";
+
+/* Accessibility label */
+"COMPASS_A11Y_LABEL" = "指南針";
+
+/* Compass abbreviation for north */
+"COMPASS_NORTH" = "北";
+
+/* Copyright notice in attribution sheet */
+"COPY_MAPBOX" = "© Mapbox";
+
+/* Copyright notice in attribution sheet */
+"COPY_OSM" = "© OpenStreetMap";
+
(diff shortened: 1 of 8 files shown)


1ec5:
The Node macOS build failure is #7681.

1ec5:
I’ve uploaded this translation to Transifex, where corrections are welcome. I’d prefer to merge this PR before #7923 so that this repository reflects the fact that zh-Hant is complete in Transifex.

1ec5:
Merging as a prerequisite to #7923; if you spot any mistranslations, please correct them in Transifex.

platform/ios/resources/zh-Hant.lproj/Localizable.strings
```diff
@@ -0,0 +1,84 @@
...
+"COMPASS_NORTH" = "北";
+
+/* Copyright notice in attribution sheet */
+"COPY_MAPBOX" = "© Mapbox";
+
+/* Copyright notice in attribution sheet */
+"COPY_OSM" = "© OpenStreetMap";
+
+/* Instructions in Interface Builder designable; {key}, {plist file name} */
+"DESIGNABLE" = "在%2$@中將你的access token設爲%1$@可在這裏顯示Mapbox上的地圖\n\n更多說明請見";
```
YunjieLi: Looking great!
The only thing I'm not sure about is 这裡 vs 这裏(L29 and L83). I know both are correct but it seems to me that 这裡 is more common in digital media? 

cc @lily-chai @virginiayung ?

1ec5: You can correct any issues you find inside Transifex. There's also a Comments tab you can use to flag mistranslations for others to review. We'll periodically pull translations in all languages from Transifex before release.

PR #7904, opened 2017-01-31
[ios] [macos] handle 403s same as 404s when downloading tiles

wibge:
Without this change, offline region downloads will stall after hitting a few tiles that return 403. 

We encounter this a lot with tile servers coming from s3 buckets. The buckets are configured to return 403 when the tile isn't available.

### DIFF
--- a/platform/darwin/src/http_file_source.mm
+++ b/platform/darwin/src/http_file_source.mm
@@ -287,7 +287,8 @@ std::unique_ptr<AsyncRequest> HTTPFileSource::request(const Resource& resource, 
 
                     if (responseCode == 200) {
                         response.data = std::make_shared<std::string>((const char *)[data bytes], [data length]);
-                    } else if (responseCode == 204 || (responseCode == 404 && resource.kind == Resource::Kind::Tile)) {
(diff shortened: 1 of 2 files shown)


1ec5:
For what it’s worth, Mapbox’s own APIs sometimes return HTTP 403 to indicate an invalid access token. If anything, it sounds like this PR would improve the behavior for Mapbox-hosted tiles as well, correct?

/cc @jfirebaugh @bsudekum

bsudekum:
> Mapbox’s own APIs sometimes return HTTP 403 to indicate an invalid access token.

@1ec5 I'm not certain this is true. A missing or malformed access token should always return a 401.

jfirebaugh:
Thanks for the contribution @wibge, but I think we need to consider a different approach here. Treating 404 as "no content" is a historical accommodation that we want to move away from (in favor of 204), rather than extend to other error codes. Furthermore, 403 is more commonly and correctly used to indicate a type of access control error, and I think it would be confusing to the majority of developers to silently discard such errors and instead behave as as if the tile was empty.

PR #7906, opened 2017-01-31
[android] - enforce usage bitmap for icon, cleanup test app resources

### LINKED ISSUE #7897: IconFactory should only support bitmaps
If you use the Google Maps SDK for Android and pass in a drawable resource that isn't a BitmapDrawable it will throw an exception indicating only Bitmaps are supported (thus BitmapDrawable and no ShapeDrawable or VectorDrawable). Supporting these other types of drawables is something we currently do but as a side effect we are drawing bitmaps ourselves, The code used for this is often the source of memory leaks and memory churn.  I think we are good to remove support for these and go for a Bitmap only approach with a dedicated exception. 

cc @mapbox/android

tobrun:
Closes #7897 + general cleanup of unused resources + fixes the memory leaks on AnimatedMarkerActivity.


tobrun:
Yes, I introduced [IconUtils](https://github.com/mapbox/mapbox-gl-native/blob/b229bfbf0c03fbe746fbff19d4bffba36ec3f7ce/platform/android/MapboxGLAndroidSDKTestApp/src/main/java/com/mapbox/mapboxsdk/testapp/utils/IconUtils.java). This utility is used in multiple places in the test app code.

PR #7907, opened 2017-01-31
Allow ObjectAnimator to animate rotation of MarkerView

### LINKED ISSUE #7798: Cannot animate MarkerView rotation
Observed in mapbox-android-sdk 4.2.1

I'm trying to animate a MarkerView rotation.
Using the example code from https://www.mapbox.com/android-sdk/examples/animated-marker/
with "position" animation replaced by "rotation" animation in onMapClick:

```
   ValueAnimator markerAnimator = ObjectAnimator.ofObject(marker, "rotation", new FloatEvaluator(), -90, 90);
   markerAnimator.setDuration(5000);
   markerAnimator.start();
```

### Expected behavior
Marker rotates slowly from -90 to 90
### Actual behavior
Nothing happens for 5 sec, then marker rotates quickly to the end angle (90)

My suggestion is that internal animation implemented in MarkerView.setRotation does not work correctly when rotation is set every frame by the external animation.
There is a workaround: get a View and apply animation directly to that View.

tobrun:
This PR allows to use an `ObjectAnimator` to animate the rotation value instead of doing the animation ourselves. Closes #7798

### DIFF
--- a/platform/android/MapboxGLAndroidSDK/src/main/java/com/mapbox/mapboxsdk/annotations/MarkerView.java
+++ b/platform/android/MapboxGLAndroidSDK/src/main/java/com/mapbox/mapboxsdk/annotations/MarkerView.java
@@ -266,18 +266,9 @@ public class MarkerView extends Marker {
    * @param rotation the rotation value to animate to.
    */
   public void setRotation(float rotation) {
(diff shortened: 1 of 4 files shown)


tobrun:
That code shouldn't be there at all, is fixed up now

PR #7909, opened 2017-01-31
[Qt] Use GCC 4.9 compatible initialization

brunoabinader:
This causes build errors when compiling on older compilers e.g. GCC 4.9.


brunoabinader:
Cancelling this PR - all GCC 4.9-related changes are eventually going to be collected and sent on a single PR.

PR #7916, opened 2017-02-01
7910: cancel tracking if ongoing animation is stopped manually

### LINKED ISSUE #7910: [iOS] desynchro on user annotation
**Platform: iOS 9, 10**
**Mapbox SDK version: 3.4.1 **

### Steps to trigger behavior

 1. Swipe far away from your location 
 2. Set the user tracking mode to follow
 3. Immediately tap on screen to stop the animation

### Expected behavior
The map should starts to fly to your location. The tap on screen stops the map and the user location is not visible (as it is far from the bounding box)

### Actual behavior
The map starts to fly to the user location. The tap on screen stops the map but the user location is shown on the middle of the map where it should not be. It stays there as long a new location is received.

The bug comes from the user location considered as centered from the start of the action but it is not yet centered before the end of animation.

I don't know yet how to correct that.

RomainQuidet:
Fix #7910

### DIFF
--- a/platform/ios/CHANGELOG.md
+++ b/platform/ios/CHANGELOG.md
@@ -51,6 +51,7 @@ Mapbox welcomes participation and contributions from everyone. Please read [CONT
 * The error passed into `-[MGLMapViewDelegate mapViewDidFailLoadingMap:withError:]` now includes a more specific description and failure reason. ([#8418](https://github.com/mapbox/mapbox-gl-native/pull/8418))
 * Improved CPU and battery performance while animating a tilted map’s camera in an area with many labels. ([#9031](https://github.com/mapbox/mapbox-gl-native/pull/9031))
 * Fixed an issue rendering polylines that contain duplicate vertices. ([#8808](https://github.com/mapbox/mapbox-gl-native/pull/8808))
+* Fixed an issue when user starts tracking position and stops manually the animation ([#7916](https://github.com/mapbox/mapbox-gl-native/pull/7916))
 
 ## 3.5.4 - May 9, 2017
 
--- a/platform/ios/CHANGELOG.md
+++ b/platform/ios/CHANGELOG.md
@@ -42,6 +42,7 @@ Mapbox welcomes participation and contributions from everyone. Please read [CONT
 * Fixed an issue preventing the Mapbox Telemetry confirmation dialog from appearing when opened from within a map view in a modal view controller. ([#9027](https://github.com/mapbox/mapbox-gl-native/pull/9027))
 * Corrected the size of MGLMapView’s compass. ([#9060](https://github.com/mapbox/mapbox-gl-native/pull/9060))
 * The Improve This Map button in the attribution action sheet now leads to a feedback tool that matches MGLMapView’s rotation and pitch. `-[MGLAttributionInfo feedbackURLAtCenterCoordinate:zoomLevel:]` no longer respects the feedback URL specified in TileJSON. ([#9078](https://github.com/mapbox/mapbox-gl-native/pull/9078))
+* Fixed an issue when user starts tracking position and stops manually the animation. ([#7916](https://github.com/mapbox/mapbox-gl-native/pull/7916))
 
 ### Other changes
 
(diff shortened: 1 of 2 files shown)


platform/ios/src/MGLMapView.mm
```diff
@@ -1169,6 +1169,10 @@ - (void)touchesBegan:(__unused NS_SET_OF(UITouch *) *)touches withEvent:(__unuse
 {
     _changeDelimiterSuppressionDepth = 0;
     _mbglMap->setGestureInProgress(false);
+    if (self.userTrackingState == MGLUserTrackingStateBegan)
+    {
+        [self setUserTrackingMode:MGLUserTrackingModeNone animated:NO];
```
1ec5: This code runs for every gesture, unconditionally. However, we intentionally stay in user tracking mode for gestures other than panning. For example, it should be possible to remain in location tracking mode after pinching or rotating and remain in heading tracking mode after tilting.

RomainQuidet: I agree, it's not incompatible. If you are pinching or rotating, the user tracking state should be possible or changed. In this case the tracking mode will not be changed. This code still works in all cases.
When the user changes the tracking mode, if an animation is needed, the tracking state goes to began as long as the flyTo completion is called. In any cases, if the user touch the screen while animating the map (to go to user location), it means he wants to cancel the move (and you to so by calling cancelTransistions()). Stopping the map where it is without returning to no tracking mode is a bug.
If the user is in tracking mode, pinching or rotating the map will not be affected by this code as the tracking state is not began.

1ec5: That’s a good point when it comes to the [initial transition](https://github.com/mapbox/mapbox-gl-native/blob/13f11635921b660cd4817a1615a003ea183b3005/platform/ios/src/MGLMapView.mm#L4262) for the first user location update. This is definitely an improvement in that case. On the other hand, we also [set `MGLUserTrackingStateBegan`](https://github.com/mapbox/mapbox-gl-native/blob/13f11635921b660cd4817a1615a003ea183b3005/platform/ios/src/MGLMapView.mm#L4227) before animating in response to a significant user location update. Consider the following scenario:

1. The user is in user location tracking mode while a passenger in a moving car.
1. The user enters a tunnel.
1. Upon emerging from the tunnel half a mile away, `-didUpdateLocationSignificantlyAnimated:` is called, setting `MGLUserTrackingStateBegan` before flying to the new location.
1. While flying to the new location, the user double-taps to zoom in.

Double-tapping ideally wouldn’t stop the transition or boot the user out of user tracking mode. However, as you’ve noted, we do call `cancelTransitions()` because we lack the ability to perform animations concurrently (#3625). Still, the current code automatically resumes following the user’s location upon the next location update, whereas this change means the user will have to somehow reenter user location tracking mode.

I wonder if there’s some way to persist the user tracking mode if the current transition is due to a significant user location update, but not if the current transition is the _first_ user location update.

/cc @frederoni

platform/ios/CHANGELOG.md
```diff
@@ -51,6 +51,7 @@ Mapbox welcomes participation and contributions from everyone. Please read [CONT
 * The error passed into `-[MGLMapViewDelegate mapViewDidFailLoadingMap:withError:]` now includes a more specific description and failure reason. ([#8418](https://github.com/mapbox/mapbox-gl-native/pull/8418))
 * Improved CPU and battery performance while animating a tilted map’s camera in an area with many labels. ([#9031](https://github.com/mapbox/mapbox-gl-native/pull/9031))
 * Fixed an issue rendering polylines that contain duplicate vertices. ([#8808](https://github.com/mapbox/mapbox-gl-native/pull/8808))
+* Fixed an issue when user starts tracking position and stops manually the animation ([#7916](https://github.com/mapbox/mapbox-gl-native/pull/7916))
```
1ec5: This should go in the “User interaction” section above. (Also, the sentence is missing a period.)

platform/ios/src/MGLMapView.mm
```diff
@@ -84,6 +84,8 @@ typedef NS_ENUM(NSUInteger, MGLUserTrackingState) {
     MGLUserTrackingStatePossible = 0,
     /// The map view has begun to move to the first reported user location.
     MGLUserTrackingStateBegan,
+    /// The map view has begun to move to
```
1ec5: …to a significant new location?

PR #7917, opened 2017-02-01
[Qt] Use ICU whenever available

brunoabinader:
Take advantage of `qtConfig(icu)` provided via QMake to let Mapbox GL native know when to use the default ICU-based bidi pimpl over the stub Qt one.

### DIFF
--- a/mapbox-gl-native.pro
+++ b/mapbox-gl-native.pro
@@ -460,7 +460,6 @@ SOURCES += \
     platform/default/mbgl/util/default_thread_pool.cpp \
     platform/default/online_file_source.cpp \
     platform/qt/src/async_task.cpp \
-    platform/qt/src/bidi.cpp \
     platform/qt/src/http_file_source.cpp \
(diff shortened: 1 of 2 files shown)


mapbox-gl-native.pro
```diff
@@ -575,3 +574,11 @@ qtConfig(system-zlib) {
 } else {
     QT_PRIVATE += zlib-private
 }
+
+qtConfig(icu) {
```
tmpsantos: Don't you need to link with ICU when making the shared library? Something like...

`qtbase/src/3rdparty/icu_dependency.pri`

tmpsantos: If you need, this is not the repo for this. This PR is correct.

tmpsantos: Remember to make sure we are linking with ICU when creating the plugin at `mapboxgl.pro`

brunoabinader: Fixed - thanks!

PR #7922, opened 2017-02-01
Do not draw foregroundBearingDrawable when rotation sensor not available

fauger:


tobrun:
this was merged in #8065. Closing.

PR #7923, opened 2017-02-01
Manage translations in Transifex

1ec5:
This PR configures and documents a workflow for pulling translations of the iOS and macOS SDKs from the [Mapbox GL Native project on Transifex](https://www.transifex.com/mapbox/mapbox-gl-native/). This setup can easily be generalized to manage any localizable strings in the Android SDK.

Although the `tx` tool is considerably slower than the REST API usage in iD and osrm-text-instructions, I like the various conveniences built the tool. One area for future improvement would be automating adding a new language to the Xcode projects using [node-xcode](https://github.com/alunny/node-xcode/tree/master/lib) or the [xcodeproj](https://github.com/CocoaPods/Xcodeproj) gem.

/cc @bhousel @freenerd

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,4 +1,5 @@
 .DS_Store
+.transifexrc
 *.pyc
 *.gcno
 *.gcda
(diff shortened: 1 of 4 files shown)


1ec5:
I’d prefer to merge #7899 before this PR so that this repository reflects the fact that zh-Hant is complete in Transifex.

PR #7926, opened 2017-02-02
[qt] Added support for features as input for GeoJSON source

brunoabinader:
Using a mix of runtime styles and annotations API, we can e.g. create a circle layer using `AnnotationManage::SourceID` as source, and then later adding a `StyleSourcedAnnotation` passing that newly-created layer ID.

The technique above already works for most annotation geometry types - except `Point`. This PR:
1. Adds support for `Point<double>` in `ShapeAnnotationGeometry` variant, used by `StyleSourcedAnnotation` to store the annotation's geometry
2. Adds support for `CircleLayer` in `StyleSourcedAnnotation`
3. Forwards API changes into QMapbox{GL} API

This makes it possible for implementing [MapCircle](http://doc.qt.io/qt-5/qml-qtlocation-mapcircle.html) QML item natively for our QtLocation Mapbox GL plugin.

### DIFF
--- a/platform/qt/app/mapwindow.cpp
+++ b/platform/qt/app/mapwindow.cpp
@@ -258,6 +258,30 @@ void MapWindow::keyPressEvent(QKeyEvent *ev)
             }
         }
         break;
+    case Qt::Key_5: {
+            if (m_map->layerExists("circleLayer")) {
+                m_map->removeLayer("circleLayer");
+                m_map->removeSource("circleSource");
+            } else {
+                QMapbox::CoordinatesCollections geometry { { { m_map->coordinate() } } };
+                QMapbox::Feature feature { QMapbox::Feature::PointType, geometry, {}, {} };
+
+                QVariantMap circleSource;
+                circleSource["type"] = "geojson";
+                circleSource["data"] = QVariant::fromValue<QMapbox::Feature>(feature);
+                m_map->addSource("circleSource", circleSource);
+
+                QVariantMap circle;
+                circle["id"] = "circleLayer";
+                circle["type"] = "circle";
+                circle["source"] = "circleSource";
+                m_map->addLayer(circle);
+
+                m_map->setPaintProperty("circleLayer", "circle-radius", 10.0);
+                m_map->setPaintProperty("circleLayer", "circle-color", QColor("black"));
+            }
+        }
+        break;
     case Qt::Key_Tab:
         m_map->cycleDebugOptions();
         break;
--- a/platform/qt/app/mapwindow.cpp
+++ b/platform/qt/app/mapwindow.cpp
@@ -220,7 +220,7 @@ void MapWindow::keyPressEvent(QKeyEvent *ev)
                 QMapbox::Coordinate topLeft     = m_map->coordinateForPixel({ 0, 0 });
                 QMapbox::Coordinate bottomRight = m_map->coordinateForPixel({ qreal(size().width()), qreal(size().height()) });
                 QMapbox::CoordinatesCollections geometry { { { topLeft, bottomRight } } };
-                QMapbox::LineAnnotation line { { QMapbox::ShapeAnnotationGeometry::Type::LineStringType, geometry }, 0.5f, 1.0f, Qt::red };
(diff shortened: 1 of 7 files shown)


jfirebaugh:
Why not use the runtime styling API to implement this? `StyleSourcedAnnotation` was a stopgap from before we had that.

brunoabinader:
> Why not use the runtime styling API to implement this?

Ideally we wanted to use runtime styling for everything. However, currently there is no straightforward way for encoding raw coordinate data into GeoJSON format - required by e.g. `GeoJSONSource`. In that sense, the annotation source population methods fits that necessity just fine.

jfirebaugh:
> currently there is no straightforward way for encoding raw coordinate data into GeoJSON format - required by e.g. `GeoJSONSource`

Not sure what you mean by that. Can you elaborate?

brunoabinader:
> Not sure what you mean by that. Can you elaborate?

From every platform that supports the annotations API, we obtain a coordinate (or a list of) contained in a platform-specific object e.g. [QGeoCoordinate](http://doc.qt.io/qt-5/qgeocoordinate.html). Because `GeoJSONSource` accepts either an url or a raw `GeoJSON` object, we then need to convert all these platform-specific coordinates into GeoJSON.

Qt currently only converts [from raw QByteArray into GeoJSON](https://github.com/mapbox/mapbox-gl-native/blob/master/platform/qt/src/qt_geojson.hpp), and since there is no ready-made way for converting Qt geographical coordinates into GeoJSON, I thought that it'd be simpler to just use annotation sources as this is exactly what they were meant for - storing coordinates data independently from the style sources.

1ec5:
Instead of implementing circles as a new type of shape annotation backed by a point geometry, would it be possible to reuse the existing point annotation functionality but add options for displaying point annotations using circle layers instead of symbol layers?

brunoabinader:
> Instead of implementing circles as a new type of shape annotation backed by a point geometry, would it be possible to reuse the existing point annotation functionality but add options for displaying point annotations using circle layers instead of symbol layers?

By existing _point annotation functionality_ you mean `SymbolAnnotation`? From what I see, `SymbolAnnotation` is an exception as-is because it cannot be _styled_ - the closest we can get to it is updating the annotation's icon image - but that's not part of the style.

My point in having it using `StyleSourcedAnnotations` is that it is then much easier to separate data from style changes - currently, for every annotation change, say a color change, we need to update the whole `mbgl::Annotation` object via `Map::updateAnnotation`. When using a corresponding style layer, instead, we can then apply runtime style changes without the need for storing (or recreating) a whole `mbgl::Annotation` object.

However, as you previously mentioned in https://github.com/mapbox/mapbox-gl-native/issues/1734#issuecomment-242802281 - if we expose the style layer from each annotation, I'd be fine then with having a `PointAnnotation` in a similar fashion as we have `SymbolAnnotation` today.

brunoabinader:
> `SymbolAnnotation` is an exception as-is because it cannot be _styled_

Just making sure I was referring to the mechanism how we add, update and remove symbol annotations and its icons, and not the symbol layer used underneath it, as you pointed out earlier.

1ec5:
> By existing _point annotation functionality_ you mean `SymbolAnnotation`?

Yes, that’s what I meant.

> `SymbolAnnotation` is an exception as-is because it cannot be _styled_

OK, that makes sense. Symbol annotations are special in that all symbols are part of a single symbol layer, whereas each shape annotation gets its own layer. The current implementation for symbol annotations works because `icon-image` is set to `{some-feature-property}`, whereas for other properties we’d need to ensure support for property functions.

brunoabinader:
Coverage decreased due to recent unit tests changes made upstream.

@jfirebaugh @tmpsantos can you please review?

jfirebaugh:
I would like to see this implemented using runtime styling, making whatever enhancements are necessary to support conversion to `GeoJSONSource`, rather than extending the annotation API.

brunoabinader:
Given the feedback on core circle annotations support, I've added `QMapbox::Feature` and its conversion helper to be accepted as input when creating a GeoJSON source.

These changes now only touches Qt-related code.

brunoabinader:
Documentation handled in https://github.com/mapbox/mapbox-gl-native/issues/7980.

PR #7930, opened 2017-02-02
Remove unneeded <regex> includes

springmeyer:
Removes a few unneeded includes from tests, adds one that is needed.


jfirebaugh:
Thanks @springmeyer!

PR #7938, opened 2017-02-03
[build] Don't require running npm install within mapbox-gl-js submodule

jfirebaugh:


kkaefer:
Running `make style-code` fails with `Error: Cannot find module 'sort-object'`, which is only installed by mapbox-gl-js

kkaefer:
Also requires `lodash.isequal`, and `jsonlint-lines-primitives`.

jfirebaugh:
What are the stack traces for those errors? As far as I can see, those dependencies are needed only used if the `style-spec` root index.js is required, which this PR bypasses by requiring `style-spec/reference/latest` instead.

kkaefer:
For some reason, I can no longer replicate the issue with `make style-code`. However, I'm still getting errors when running the node render tests:

```
module.js:327
    throw err;
    ^

Error: Cannot find module 'd3-queue'
    at Function.Module._resolveFilename (module.js:325:15)
    at Function.Module._load (module.js:276:25)
    at Module.require (module.js:353:17)
    at require (internal/module.js:12:17)
    at Object.<anonymous> (/Users/kkaefer/Code/gl/native/mapbox-gl-js/test/integration/lib/harness.js:7:15)
    at Module._compile (module.js:409:26)
    at Object.Module._extensions..js (module.js:416:10)
    at Module.load (module.js:343:32)
    at Function.Module._load (module.js:300:12)
    at Module.require (module.js:353:17)
Program ended with exit code: 1
```

jfirebaugh:
Do you still get that if you update to npm 3 or later?

kkaefer:
@jfirebaugh I'm not sure we should rely on users upgrading their npm from the version that shipped with the stock installation.

jfirebaugh:
OK, I'll try different approaches.

kkaefer:
Don't get me wrong, I think there's a lot of value in removing the npm install for `mapbox-gl-js`, if we can find a way to just install the parts that are required for the test suite. Alternatively, we could also up the minimum required node version to v6.

jfirebaugh:
https://github.com/mapbox/mapbox-gl-native/commit/94a3c272ffd6676edbe59342d70bca49aaf676a5#diff-e43a15f1b424c5893699e4225835cc9d should do for now -- next steps are to see if we can remove `npm install` from the Makefile, and update node-cmake to remove that as an npm dependency.

kkaefer:
there's a new version of node-cmake that explicitly doesn't use npm and instead advises you to copy the cmake file into your project. I've encountered 2 issues switching to the new version: https://github.com/cjntaylor/node-cmake/issues/17 https://github.com/cjntaylor/node-cmake/issues/18

PR #7939, opened 2017-02-03
[core] Implement data-driven styling for {text,icon}-{color,opacity,halo-color,halo-blur,halo-width}

anandthakker:


anandthakker:
Currently running into the following problem:

 - `SymbolIconProgram` and `SymbolSDFProgram` both use `SymbolPaintProperties` when instantiating `Program<>`.
 - `SymbolPaintProperties`, being generated from the style spec, includes properties that bind to GL attributes like `a_halo_color` that are not present in the (non-sdf-)icon shader.

I _think_ this is the reason that I'm getting a `glBindAttribLocation(id, location, name): Error GL_INVALID_VALUE` error: according to the debugger, this is happening during `mbgl::gl::bindAttributeLocation(id, location, "a_halo_color_max")`, during the instantiation of the `symbolIcon` program.

jfirebaugh:
I don't think that's the reason -- that pattern occurs elsewhere, like with `a_outline_color` and fill programs. I suspect the reason is rather that we've now introduced more than `GL_MAX_VERTEX_ATTRIBS` attributes. This is a problem. We're either going to have to find creative ways of packing multiple attributes into one, or move to a system more like mapbox-gl-js, where shaders are dynamically generated that use only as many attributes as required by data-driven properties, and use uniforms otherwise -- this moves the failure condition to the point when someone tries to use too many data-driven properties at the same time.

anandthakker:
Yeah - so it looks like we're ending up with attribute ids >= 16 because _each_ data-driven `icon-` / `text-` property is mapping to a separate attribute, so we're ending up with many more than we need.  E.g., `text-opacity` and `icon-opacity` both become `a_opacity`, but they take up two separate positions in the parameter list:

```c++
mbgl::gl::Attributes<
mbgl::attributes::a_pos,
mbgl::attributes::a_offset<2>,
mbgl::attributes::a_texture_pos,
mbgl::attributes::a_data<4>,
mbgl::attributes::Min<mbgl::attributes::a_opacity>,
mbgl::attributes::Min<mbgl::attributes::a_fill_color>,
mbgl::attributes::Min<mbgl::attributes::a_halo_color>,
mbgl::attributes::Min<mbgl::attributes::a_halo_width>,
mbgl::attributes::Min<mbgl::attributes::a_halo_blur>,
mbgl::attributes::Min<mbgl::attributes::a_opacity>,
mbgl::attributes::Min<mbgl::attributes::a_fill_color>,
mbgl::attributes::Min<mbgl::attributes::a_halo_color>,
mbgl::attributes::Min<mbgl::attributes::a_halo_width>,
mbgl::attributes::Min<mbgl::attributes::a_halo_blur>,
mbgl::attributes::Max<mbgl::attributes::a_opacity>,
mbgl::attributes::Max<mbgl::attributes::a_fill_color>,
mbgl::attributes::Max<mbgl::attributes::a_halo_color>,
mbgl::attributes::Max<mbgl::attributes::a_halo_width>,
mbgl::attributes::Max<mbgl::attributes::a_halo_blur>,
mbgl::attributes::Max<mbgl::attributes::a_opacity>,
mbgl::attributes::Max<mbgl::attributes::a_fill_color>,
mbgl::attributes::Max<mbgl::attributes::a_halo_color>,
mbgl::attributes::Max<mbgl::attributes::a_halo_width>,
mbgl::attributes::Max<mbgl::attributes::a_halo_blur> >
```

src/mbgl/programs/symbol_program.hpp
```diff
@@ -94,7 +94,7 @@ class SymbolIconProgram : public Program<
                                        const TransformState&);
 };
 
-class SymbolSDFProgram : public Program<
+class SymbolSDFIconProgram : public Program<
```
anandthakker: I initially tried to just parameterize `SymbolSDFProgram` on the paint property types, but it was proving sort of unweildy.   Opted to duplicate the declarations and minimize code duplication in implementation via the shared  `makeSDFValues` function

anandthakker:
@jfirebaugh Okay, I think this is ready for a 👀 .

 - History is hopefully reasonable to read commit-by-commit (with the generated code in its own commit).
 - There are three render tests failing for what seem like maybe platform-specific detail reasons?  In two of the failures the difference between expected/actual isn't visible to the naked eye... The `text-halo-blur` one _could_ actually be related to the blur property, but it seems more like something weird w/ the background.

src/mbgl/layout/symbol_instance.hpp
```diff
@@ -28,6 +28,7 @@ class SymbolInstance {
     CollisionFeature textCollisionFeature;
     CollisionFeature iconCollisionFeature;
     WritingModeType writingModes;
+    std::size_t featureIndex;
```
anandthakker: In GL JS, we did this with a direct reference to the associated feature.  Would it be better to do a similar thing here, making this a `SymbolFeature *`?

anandthakker: ^ note: the reason I didn't was because, since `SymbolInstance`:`SymbolFeature` is `many:1`, it seemed like there would have to be some weird acrobatics using `std::shared_ptr`.  Rather than introduce that machinery, I figured since both of these objects are really internal to SymbolLayout, this index-based coupling seemed okay.

anandthakker:
_Almost_ there, I think:
 - Working on trying to run the android tests locally (as @ivovandongen indicated that most are skipped in CI bc of timing issues)
 - One of the Qt travis builds failed the Memory.Footprint test.  Re-running it now just in case it was a fluke... but no real clue where to start on solving that one.

jfirebaugh:
So we're going ~660kb over the Memory.Footprint  test limit. How much under the limit are we before this change, i.e. how much is the total increase?

src/mbgl/layout/symbol_instance.hpp
jfirebaugh: I think this is okay for now. I started learning more about the various classes used during symbol placement and I'm thinking maybe we can move to a hierarchy where `SymbolFeature` owns `SymbolInstance`s, and can provide a direct `SymbolFeature&` reference at the appropriate time.

anandthakker:
> So we're going ~660kb over the Memory.Footprint test limit. How much under the limit are we before this change, i.e. how much is the total increase?

Force-logged it off of master (by setting the threshold to 0 in the unit test, because apparently googletest doesn't let you just arbitrary informational stuff 🙄 ):

> Expected: (vectorFootprint) < (0), actual: 6.71485e+07 vs 0

( https://travis-ci.org/mapbox/mapbox-gl-native/jobs/204010132#L1013 )

On this branch, we get:

> Expected: (vectorFootprint) < (65 * 1024 * 1024), actual: 6.88349e+07 vs 68157440

So that's a 1.6MB increase.  Seems pretty significant.

anandthakker:
It occurs to me that this test is measuring [RSS](https://en.wikipedia.org/wiki/Resident_set_size), and so I'm not sure whether these literal numbers are meaningfully interpretable as "the increase in memory usage".

tmpsantos:
> It occurs to me that this test is measuring RSS, and so I'm not sure whether these literal numbers are meaningfully interpretable as "the increase in memory usage".

What this test is doing is:

1. "Warming up" the memory by rendering 10x and deleting the map object after rendering.
2. Measure the RSS, which should now include all the static data, thread locals and 3rd party libraries initialization data.
3.  Render N times (currently 15), but keep the map objects.
4. Measure the current RSS compared to the first measurement and divide by N (15). This should be roughly how much each Map object is using. I set the bar at 65 MB per Map for vector and 25 MB per Map for raster. These numbers are arbitrary.
5. Fail if it goes over this limit.

Few notes:

- We use jmalloc because it is very aggressive at freeing memory, making the measurement more precise.
- This test is surprisingly stable, and has protected us from memory growth in the past. Last one I remember was a PR that would pre-allocate a buffer for the annotations layer even if you don't have any annotation on the map. The regression was prevented by making this allocation lazy.
- We could have used more sophisticated ways of tracking memory allocations, but RSS snapshots look reliable and fast enough for running on Travis. That said, any suggestion is welcome here.

tmpsantos:
One last thing that is worth mentioning. This test is focused on Map objects rather than tiles. If your code increases tile memory usage by 250 Kb per tile it is unlikely that this test will catch it, but the effects will be seen in an app when rendering 10 tiles at once in a tilted map (~2.5 MB memory increase).

Additional manual testing with larger maps is advised, I personally like this tool:
https://github.com/pixelb/ps_mem

anandthakker:
Capturing some notes from chat:

@anandthakker:
> Agreed that this test seems like a good/reliable way to detect when a  regression has occurred in memory usage.
> My question is because, now that I know that the regression exists in my branch,  I am interested in assessing more quantitatively what the memory impact of these changes are

@tmpsantos:
> the other thing is, as we add more features, eventually we will have to use more memory
> so it is just a matter of raise the memory bar there and life goes on

@anandthakker:
> yeah, this branch is one such case, for sure: it includes a necessary change wherein `SymbolLayout` (one per symbol style layer) now holds on to a vector of features, rather than clearing it
> we might, in the end, have to raise the memory bar.  But if we’re going to do that, then ideally I’d like to:
1. Determine how much more memory usage we’re incurring for this change.
2. Actually _account_ for (roughly) that amount based on the known increases in the change (e.g. the symbol features)
> I’m working on 2 by doing some napkin math on the number of symbol layers / symbols that show up in this test, but I’m not sure how to do 1 with any confidence/accuracy (edited)

tmpsantos:
Did some tests to measure the memory impact of this change. On my Linux box, for rendering this map (1920x1200) using GLFW:

screenshot from 2017-02-22 16-14-46

```
$ cat /tmp/mbgl-native.cfg 
-73.9976
40.7333
14.128
-57.7533
60
2
```

I started the map with the same configuration 10x with this PR applied and without (the hash just before the first commit of this branch.

In average `ps_mem` was giving me 139.17 MB with this PR and 128.01 MB without. An increase of 11.16 MB.

Repeating the method with a 1024x768 map, the observed increase was 8 MB.

anandthakker:
Looks like in the Memory.Footprint test, we're dealing with 980 symbol features total... that's 1.7KB per symbol feature, which would be plausible if we're keeping the geometries around but seems too big for just the text, placement, properties.

@jfirebaugh two questions:
1. Do you see any reason why clearing each feature's geometry [here](https://github.com/mapbox/mapbox-gl-native/pull/7939/commits/dcacc170d5ce69b863f4674cb190e530cc2d771a#diff-f27a3a157c549a13f8fe1ee8c8157ea8R299)  is failing to release the memory?
2. Could the paint vertex buffers be part of the problem here?  I've been assuming not thus far, because I figured that, like in GL JS, they'd be destroyed after uploading to GL.

anandthakker:
stats from profiling in Instruments:

`master`
screen shot 2017-02-22 at 3 57 02 pm

`dds-symbol`
screen shot 2017-02-22 at 3 54 05 pm

Obtained by profiling the Memory.Footprint test, and the filtering the Instruments time range to go from the beginning to point of peak heap usage.

The +43MB delta in heap usage includes:

+8MB - `Malloc 4.00 MiB`
+8MB - `Malloc 1.00 MiB`
+9.5MB - `Malloc 80 Bytes`
+3.91MB - `SymbolFeature`

Other than SymbolFeature, hard to say what these others are... probably allocations by various container classes?

jfirebaugh:
Here are a couple things I would try to reduce the memory growth:

* Have `SymbolFeature` assume and hold ownership of the `std::unique_ptr<GeometryTileFeature>`, rather than copying its constituent parts. This may help reduce fragmentation.
* Remove `SymbolFeature::iconOffset` and `iconRotation` in favor of calculating them as needed, like with the other properties in this PR.

src/mbgl/layout/symbol_layout.cpp
```diff
@@ -229,7 +232,8 @@ void SymbolLayout::prepare(uintptr_t tileUID,
     const bool textAlongLine = layout.get<TextRotationAlignment>() == AlignmentType::Map &&
         layout.get<SymbolPlacement>() == SymbolPlacementType::Line;
 
-    for (const auto& feature : features) {
+    for (auto it = features.begin(); it != features.end(); ++it) {
+        auto feature = *it;
```
jfirebaugh: This is creating a copy; it should be `auto& feature = *it;`. Then `feature.geometry.clear();` will have a meaningful effect.

src/mbgl/layout/symbol_layout.cpp
```diff
@@ -496,6 +502,12 @@ std::unique_ptr<SymbolBucket> SymbolLayout::place(CollisionTile& collisionTile)
                     layout.get<IconKeepUpright>(), iconPlacement, collisionTile.config.angle, symbolInstance.writingModes);
             }
         }
+        
+        const auto feature = features.at(symbolInstance.featureIndex);
```
jfirebaugh: Same here, should be `const auto& feature = ...`.

src/mbgl/layout/symbol_layout.cpp
anandthakker: 🙈

anandthakker: Alas, this change only seems to take us down by 51K in the Memory.Footprint test: 6.88349e+07 => 6.87836e+07.

anandthakker:
> Have SymbolFeature assume and hold ownership of the std::unique_ptr<GeometryTileFeature>, rather than copying its constituent parts. This may help reduce fragmentation.

I think that one consequence of this is that it would be problematic to then clear the geometries after we didn't need them anymore (unless we added something like a `clearGeometries()` mutator to the `GeometryTileFeature` interface).  I guess since clearing them [didn't do much](https://github.com/mapbox/mapbox-gl-native/pull/7939#discussion_r102739499), maybe that's okay.

src/mbgl/layout/symbol_feature.hpp
```diff
@@ -10,42 +10,19 @@ namespace mbgl {
...
 class SymbolFeature : public GeometryTileFeature {
 public:
-    SymbolFeature(const std::unique_ptr<GeometryTileFeature>& feature) :
-        id(feature->getID()),
-        type(feature->getType()),
-        geometry(feature->getGeometries()),
-        properties(feature->getProperties())
+    SymbolFeature(std::unique_ptr<GeometryTileFeature> feature_) :
+        feature(std::move(feature_)),
+        geometry(feature->getGeometries()) // we need a mutable copy of the geometry for mergeLines()
```
anandthakker: ☝️ @jfirebaugh note - mergeLines relies on mutating the SymbolFeatures' geometries, but getGeometries() returns a copy, so this bit of copying still needed to happen.

anandthakker:
@jfirebaugh implemented your two suggestions from https://github.com/mapbox/mapbox-gl-native/pull/7939#issuecomment-281825520.  Moving `iconOffset` and `iconRotate` off only helped marginally, but transferring ownership of the `GeometryTileFeature`s made a pretty big difference: brought us down to 6.81918e+07 -- so, in the Memory.Footprint test, that's a `+1MB` delta from master (instead of  `+1.7MB`).

src/mbgl/programs/symbol_program.hpp
jfirebaugh: How about:

```C++
template <class PaintProperties>
class SymbolSDFProgram : public Program<
    ...
    PaintProperties>
{
    ...
};

using SymbolSDFIconProgram = SymbolSDFProgram<style::IconPaintProperties>;
using SymbolSDFTextProgram = SymbolSDFProgram<style::TextPaintProperties>;
```

anandthakker: Yeah, that's something like what I was trying before and having trouble with -- but it's very possible that it was just other mid-refactor issues that were making it hard to see clearly.  Lemme try again.

src/mbgl/layout/symbol_layout.cpp
```diff
@@ -87,21 +87,23 @@ SymbolLayout::SymbolLayout(const BucketParameters& parameters,
...
+            layer->as<SymbolLayer>()->impl->textPaintProperties()
+        ));
     }
 
     // Determine and load glyph ranges
     const size_t featureCount = sourceLayer.featureCount();
     for (size_t i = 0; i < featureCount; ++i) {
-        auto feature = sourceLayer.getFeature(i);
-        if (!leader.filter(feature->getType(), feature->getID(), [&] (const auto& key) { return feature->getValue(key); }))
+        SymbolFeature ft(sourceLayer.getFeature(i));
```
jfirebaugh: It would be better to construct `SymbolFeature` only for features that pass the filter, to avoid unnecessarily building and copying the geometry.

src/mbgl/programs/symbol_program.cpp
```diff
@@ -1,14 +1,21 @@
+#include <iostream>
```
jfirebaugh: Leftover from debugging?

src/mbgl/programs/symbol_program.cpp
```diff
@@ -1,14 +1,21 @@
...
 #include <mbgl/style/layers/symbol_layer_impl.hpp>
+#include <mbgl/util/enum.hpp>
 
 namespace mbgl {
 
 using namespace style;
 
 static_assert(sizeof(SymbolLayoutVertex) == 16, "expected SymbolLayoutVertex size");
 
+MBGL_DEFINE_ENUM(SymbolSDFPart, {
```
jfirebaugh: `MBGL_DEFINE_ENUM` is needed only where enum ↔ string conversion is required, which isn't the case with `SymbolSDFPart`.

src/mbgl/programs/symbol_program.hpp
```diff
@@ -8,6 +8,7 @@
 #include <mbgl/util/geometry.hpp>
 #include <mbgl/util/size.hpp>
 #include <mbgl/style/layers/symbol_layer_properties.hpp>
+#include <mbgl/style/layers/symbol_layer_impl.hpp>
```
jfirebaugh: Shouldn't be necessary.

src/mbgl/programs/symbol_program.hpp
```diff
@@ -94,47 +95,73 @@ class SymbolIconProgram : public Program<
...
+                                       const SymbolSDFPart);
 };
 
+using SymbolSDFIconProgram = SymbolSDFProgram<style::IconPaintProperties>;
+using SymbolSDFTextProgram = SymbolSDFProgram<style::TextPaintProperties>;
+
 using SymbolLayoutVertex = SymbolLayoutAttributes::Vertex;
-using SymbolAttributes = SymbolIconProgram::Attributes;
+using SymbolIconAttributes = SymbolIconProgram::Attributes;
+using SymbolGlyphAttributes = SymbolSDFTextProgram::Attributes;
```
jfirebaugh: Rename this to `SymbolTextAttributes`.

src/mbgl/programs/symbol_program.hpp
anandthakker: This is needed here because it's where `style::{Icon,Text}PaintProperties` are declared.

src/mbgl/layout/symbol_layout.cpp
jfirebaugh: Sorry, should have been more explicit:

```
auto feature = sourceLayer.getFeature(i);
if (!filter)
    continue;
SymbolFeature ft(std::move(feature));
```

We don't want to call `sourceLayer.getFeature(i)` twice.

anandthakker: Oh whoops, yeah that's what I'd intended -- trying to go too fast heh

anandthakker: @jfirebaugh amended.

anandthakker:
A little more on the memory usage increase.  Running a single iteration from the Memory.Footprint test and estimating each feature's footprint as `4 * # coordinates + sum(length(each string property)) + 8 * # numeric properties` yields a total of 1,481,348 bytes (213,664 from geometry, the rest from properties).

src/mbgl/tile/vector_tile.cpp
```diff
@@ -15,9 +15,17 @@ class VectorTileLayer;
 
 using packed_iter_type = protozero::iterator_range<protozero::pbf_reader::const_uint32_iterator>;
 
+struct VectorTileLayerData {
```
jfirebaugh: This should also have a `shared_ptr<const std::string> data` member, to ensure that the data backing the `tags_iter` and `geometry_iter` is retained.

src/mbgl/tile/vector_tile.cpp
```diff
@@ -117,8 +121,8 @@ Value parseValue(protozero::pbf_reader data) {
     return false;
 }
 
-VectorTileFeature::VectorTileFeature(protozero::pbf_reader feature_pbf, const VectorTileLayer& layer_)
-    : layer(layer_) {
+VectorTileFeature::VectorTileFeature(protozero::pbf_reader feature_pbf, std::shared_ptr<VectorTileLayerData> layerData_)
+    : layerData(layerData_) {
```
jfirebaugh: `layerData(std::move(layerData_))`

src/mbgl/tile/vector_tile.cpp
anandthakker: @jfirebaugh 👍 done.

anandthakker:
Latest changes bring us down to `68096819`, which is under the Memory.Footprint ceiling.  This is great!  (Except: I really am not sure _why_ this helped.  If anything, I thought that using `shared_ptr`s to a separate `VectorTileLayerData`, rather than reference to `VectorTileLayer`, should have cause a small _increase_ in memory footprint...)

src/mbgl/tile/vector_tile.cpp
springmeyer: @anandthakker @jfirebaugh - Landed here in a revised effort to integrate https://github.com/mapbox/vector-tile back into mbgl. I have some questions.

1) Without `VectorTileLayerData` and is `shared_ptr<const std::string> data` member what exactly would go wrong? I figure there must be somewhere in the code that expects to create a `VectorTileLayer` and use it/cache it in such a way beyond when the underlying data would be kept alive. So keeping a shared_ptr to the data keeps it alive as long as all `VectorTileLayer` instances are alive? Can you point to the code that needs this? (I could not find it)

2) Why does the `VectorTileFeature` need a `std::shared_ptr<VectorTileLayerData> layerData` rather than just `const VectorTileLayer&`. This indicates to me that somewhere in the code expects to keep a  `VectorTileFeature` instance alive longer than the parent `VectorTileLayer`. Is this happening and can you point to the place in the code that requires is? Or is this perhaps unneeded - in that the `VectorTileLayer` will always be alive as long as the `VectorTileFeature` created from it?

 I noticed that `shared_ptr<const std::string> data` gets passed to not only `VectorTileLayerData` but also `VectorTileFeature` and

anandthakker: > Why does the VectorTileFeature need a std::shared_ptr<VectorTileLayerData> layerData rather than just const VectorTileLayer&. This indicates to me that somewhere in the code expects to keep a VectorTileFeature instance alive longer than the parent VectorTileLayer

@springmeyer Yeah -- this is a possibility that (I think) did not exist before, but was introduced in this PR.  (And apologies: I should have left a comment justifying the `shared_ptr`s introduced here.) 

The reason it can now happen is over in `SymbolLayout`, and specifically `SymbolFeature` ([here](https://github.com/mapbox/mapbox-gl-native/pull/7939/files#diff-9a0baf47fcfbcbe492c348bfe28cacefL11)).  `SymbolLayout` is constructed with a vector tile layer, and then later on--when there's no guarantee that the vector tile data is still around--expected to produce a "bucket" containing final rendering data.  Before this PR, we copied the necessary data into `SymbolFeature` structs which `SymbolLayout` could hang on to; but because we need feature properties for data-driven styling, `SymbolFeature` now holds a reference to the `VectorTileFeature` because it turns out that copying those over is a large memory burden.

Because of this, `VectorTileFeature`s may need to outlive their parent `VectorTileLayer`s, so it needs shared ownership of `VectorTileLayerData`.

As for the `shared_ptr<const std::string> data` that's held by `VectorTileLayerData` -- I think we need this basically as a consequence of the above, since `VectorTileFeature`s may parse PBF data lazily.  But even if we didn't, it's making explicit a dependency that already existed, since there's no explicit guarantee that the parent `VectorTileData` would outlive a particular `VectorTileLayer`.  The lifecycles _happened_ to be okay before this, but only because of implicit constraints on how we used these objects.

@jfirebaugh should probably check my math on this, as I haven't spent time in the parts of our vector tile parsing/caching stuff outside of `SymbolLayout`.

jfirebaugh: Yes, that's correct. One minor note is that `shared_ptr<const std::string>` predated the addition of `VectorTileLayerData` and comes from the definition of [`Response`](https://github.com/mapbox/mapbox-gl-native/blob/b404457a74d7a63103285712552a8564823db6d9/include/mbgl/storage/response.hpp#L31). I believe this was originally done to avoid string copies. Might be better to use move semantics now.

springmeyer: @anandthakker and @jfirebaugh - awesome thank you for the confirmation. I'm grateful for the confirmation and will now proceed with my PR revising this code and keep all the explicit ownership relationships fully intact. The context I should have mentioned before is that I hope my PR will help performance so I was looking closely at the nearby code for any optimization potential, which is what lead me to the question of whether the shared pointers were absolutely needed. Great to know they are.

springmeyer: Ticketed @jfirebaugh's idea: https://github.com/mapbox/mapbox-gl-native/issues/9167

PR #7942, opened 2017-02-03
[core] Check in generated shader code

jfirebaugh:
One step toward eliminating the node/npm dependency for platforms other than node.


scripts/generate-shaders.js
```diff
@@ -0,0 +1,82 @@
...
+    "public:\n" +
+    "    static constexpr const char* name = \"" + shaderName + "\";\n" +
+    "    static constexpr const char* vertexSource = R\"MBGL_SHADER(\n" + vertexSource() + ")MBGL_SHADER\";\n" +
+    "    static constexpr const char* fragmentSource = R\"MBGL_SHADER(\n" + fragmentSource() + ")MBGL_SHADER\";\n" +
+    "};\n" +
+    "\n" +
+    "} // namespace shaders\n" +
+    "} // namespace mbgl\n";
+
+    fs.writeFileSync(path.join(outputPath, shaderName + '.hpp'), content);
```
kkaefer: Let's use a pattern that conditionally writes the shader code if changes occurred, like in [style-code.js](https://github.com/mapbox/mapbox-gl-native/blob/master/scripts/style-code.js#L25). Previously, we were running this script for every shader file, while we are now running it once for every shader file. Previously, the build system tracked changes to shader files and only rebuild the files when that particular file changed. Rewriting all files (even if they have the same file content) will trigger a rebuild that takes time. Changes to shaders are necessary when debugging rendering issues or implementing new features or adding new shaders.

scripts/generate-shaders.js
```diff
@@ -0,0 +1,82 @@
...
+'fill',
+'fill_outline',
+'fill_outline_pattern',
+'fill_pattern',
+'line',
+'line_pattern',
+'line_sdf',
+'raster',
+'symbol_icon',
+'symbol_sdf'
```
kkaefer: indent these lines?

scripts/generate-shaders.js
kkaefer: Once this PR gets merged, we should either move the actual shader code to implementation files, or change the include so that the shader code only gets included in implementation files so as to avoid recompiles if all the code when modifying a shader (this used to be the case prior to the OpenGL refactor).

jfirebaugh: 👍  Did both these things.

Issue #7934, opened 2017-02-03
Restore MGLStyleFunction


1ec5:
MGLCameraStyleFunction et al. should be subclasses of an MGLStyleFunction abstract class, to make it easier to talk about style functions as a concept in documentation and special case functions of all types in code.

/cc @boundsj

Issue #7924, opened 2017-02-02
Add documentation for data-driven styling


boundsj:
Support for data-driven styling for iOS and macOS is being added in https://github.com/mapbox/mapbox-gl-native/pull/7596. After we have used the API proposed there and before the final Mapbox iOS SDK 3.5.0 beta, we will fully document the new APIs added for DDS.

#### This will include (at least):

- [x] MGLStyleFunctionOption string enum
  - [x] MGLStyleFunctionOptionInterpolationBase
  - [x] MGLStyleFunctionOptionDefaultValue
- [x] MGLInterpolationMode enum
  - [x] MGLInterpolationModeExponential
  - [x] MGLInterpolationModeInterval
  - [x] MGLInterpolationModeCategorical
  - [x] MGLInterpolationModeIdentity
- [x] Update class documentation for `MGLStyleValue` (as a class cluster it represents something different now)
- [x] +[MGLStyleValue valueWithInterpolationMode:cameraStops:options:]
- [x] +[MGLStyleValue valueWithInterpolationMode:sourceStops:attributeName:options:]
- [x] +[MGLStyleValue valueWithInterpolationMode:compositeStops:attributeName:options:]
- [x] MGLCameraStyleFunction (and all methods and properties)
- [x] MGLSourceStyleFunction (and all methods and properties)
- [x] MGLCompositeStyleFunction (and all methods and properties)
- [x] Update generated code examples in `MGLDocumentationExampleTests.swift` to use camera functions and maybe add examples for `source` and `composite` functions
- [ ] Update `guides/For Style Authors.md` with DDS specific terminology differences
- [ ] Update `guides/Runtime Styling.md` with information about DDS (or add a DDS specific guide)

cc @friedbunny @jmkiley @1ec5 @incanus @ericrwolfe @pveugen

1ec5:
We should fill out the documentation before the _first_ beta, because that's when we tell developers to start coding against new features in nontrivial projects. We can ship an alpha without the documentation if necessary to meet time commitments, however.

1ec5:
* [ ] Update the iOS and macOS changelogs to discuss all that’s changed

boundsj:
#7931, #7943, and #7960 added API documentation that should be suitable for the beta release. We still need a PR to update the guides but I don't consider that a requirement for the beta.

Before I create the beta, I will wait for:

[...]

PR #7943, opened 2017-02-04
Add back MGLStyleFunction and re-document MGLStyleValue

boundsj:
Fixes: https://github.com/mapbox/mapbox-gl-native/issues/7934

Addresses `MGLStyleValue` documentation todos noted in https://github.com/mapbox/mapbox-gl-native/issues/7924:

- [x] Update class documentation for `MGLStyleValue` (as a class cluster it represents something different now)
- [x] +[MGLStyleValue valueWithInterpolationMode:cameraStops:options:]
- [x] +[MGLStyleValue valueWithInterpolationMode:sourceStops:attributeName:options:]
- [x] +[MGLStyleValue valueWithInterpolationMode:compositeStops:attributeName:options:]
- [x] MGLCameraStyleFunction (and all methods and properties)
- [x] MGLSourceStyleFunction (and all methods and properties)
- [x] MGLCompositeStyleFunction (and all methods and properties)

### DIFF
--- a/platform/darwin/src/MGLStyleValue.h
+++ b/platform/darwin/src/MGLStyleValue.h
@@ -189,6 +189,59 @@ MGL_EXPORT
 MGL_EXPORT
 @interface MGLStyleFunction<T> : MGLStyleValue<T>
 
+#pragma mark Creating a Style Function
+
+/**
+ Creates and returns an `MGLStyleFunction` object representing a camera function. 
+ If the function is set as a style layer property value, it will be interpreted 
+ as a camera function with a linear interpolation curve.
+
+ @note Do not create instances of `MGLStyleFunction` unless it is required for
+    backwards compatiblity with your application code. You should create and use
+    instances of `MGLCameraStyleFunction` to specify how properties will
+    be visualized at different zoom levels.
+
+ @param stops A dictionary associating zoom levels with style values.
+ @return An `MGLStyleFunction` object with the given stops.
+ */
++ (instancetype)functionWithStops:(NSDictionary<id, MGLStyleValue<T> *> *)stops __attribute__((deprecated("Use +[MGLStyleValue valueWithInterpolationMode:cameraStops:options:]")));
+
+/**
+ Creates and returns an `MGLStyleFunction` object representing a camera function.
+ If the function is set as a style layer property value, it will be interpreted 
+ as a camera function with an interpolation curve controlled by the provided 
+ interpolation base.
+
+ @note Do not create instances of `MGLStyleFunction` unless it is required for
+    backwards compatiblity with your application code. You should create and use
+    instances of `MGLCameraStyleFunction` to specify how properties will
+    be visualized at different zoom levels.
+
+ @param interpolationBase The exponential base of the interpolation curve.
+ @param stops A dictionary associating zoom levels with style values.
+ @return An `MGLStyleFunction` object with the given interpolation base and stops.
+ */
++ (instancetype)functionWithInterpolationBase:(CGFloat)interpolationBase stops:(NSDictionary<id, MGLStyleValue<T> *> *)stops __attribute__((deprecated("Use +[MGLStyleValue valueWithInterpolationMode:cameraStops:options:]")));
+
+#pragma mark Initializing a Style Function
+
+/**
+ Returns an `MGLStyleFunction` object representing a camera function. If the 
+ function is set as a style layer property value, it will be interpreted
+ as a camera function with an interpolation curve controlled by the provided
+ interpolation base.
+ 
+ @note Do not create instances of `MGLStyleFunction` unless it is required for
+    backwards compatiblity with your application code. You should create and use
+    instances of `MGLCameraStyleFunction` to specify how properties will
+    be visualized at different zoom levels.
+
+ @param interpolationBase The exponential base of the interpolation curve.
+ @param stops A dictionary associating zoom levels with style values.
+ @return An `MGLStyleFunction` object with the given interpolation base and stops.
+ */
+- (instancetype)initWithInterpolationBase:(CGFloat)interpolationBase stops:(NSDictionary<id, MGLStyleValue<T> *> *)stops NS_DESIGNATED_INITIALIZER;
+
 #pragma mark Accessing the Parameters of a Function
 
 /**
@@ -207,7 +260,8 @@ MGL_EXPORT
  
  @note This property specifies the exponential base of the interpolation curve 
     of `MGLCameraStyleFunction` and `MGLSourceStyleFunction` functions that use 
-    a `MGLInterpolationModeExponential` `interpolationMode`.
+    a `MGLInterpolationModeExponential` `interpolationMode`. Otherwise, it is
+    ignored.
  */
 @property (nonatomic) CGFloat interpolationBase;
 
--- a/platform/darwin/src/MGLStyleValue.h
+++ b/platform/darwin/src/MGLStyleValue.h
@@ -200,14 +200,11 @@ MGL_EXPORT
 /**
  A dictionary associating zoom levels with style values.
  */
-@property (nonatomic, copy) NSDictionary *stops;
+@property (nonatomic, copy, nullable) NSDictionary *stops;
 
 /**
  The exponential interpolation base of the function’s interpolation curve.
  
- @note This property only applies to `MGLCameraStyleFunction` and 
-    `MGLSourceStyleFunction` functions that use a `MGLInterpolationModeExponential`
-    `interpolationMode`.
  @note This property specifies the exponential base of the interpolation curve 
     of `MGLCameraStyleFunction` and `MGLSourceStyleFunction` functions that use 
     a `MGLInterpolationModeExponential` `interpolationMode`.
--- a/platform/darwin/src/MGLStyleValue.h
+++ b/platform/darwin/src/MGLStyleValue.h
@@ -239,6 +239,12 @@ MGL_EXPORT
 #pragma mark Accessing the Parameters of a Function
 
 /**
+ The modes used to interpolate property values between map zoom level changes or
+ over a range of feature attribute values.
+ */
+@property (nonatomic) MGLInterpolationMode interpolationMode;
+
+/**
  A dictionary associating zoom levels with style values.
  */
 @property (nonatomic, copy, nullable) NSDictionary *stops;
@@ -286,12 +292,6 @@ MGL_EXPORT
 #pragma mark Accessing the Parameters of a Camera Function
 
 /**
- The modes used to interpolate property values between map zoom level changes or
- over a range of feature attribute values.
- */
-@property (nonatomic) MGLInterpolationMode interpolationMode;
-
-/**
  A dictionary associating zoom levels with style values.
 
  Each of the function’s stops is represented by one key-value pair in the
@@ -337,12 +337,6 @@ MGL_EXPORT
 + (instancetype)functionWithInterpolationMode:(MGLInterpolationMode)interpolationMode stops:(nullable NS_DICTIONARY_OF(id, MGLStyleValue<T> *) *)stops attributeName:(NSString *)attributeName options:(nullable NS_DICTIONARY_OF(MGLStyleFunctionOption, id) *)options;
 
 #pragma mark Accessing the Parameters of a Source Function
-
-/**
- The modes used to interpolate property values between map zoom level changes or
- over a range of feature attribute values.
- */
-@property (nonatomic) MGLInterpolationMode interpolationMode;
 
 /**
  A string that specifies the feature attribute key whose value be used as the function
@@ -403,12 +397,6 @@ MGL_EXPORT
 + (instancetype)functionWithInterpolationMode:(MGLInterpolationMode)interpolationMode stops:(NS_DICTIONARY_OF(id, NS_DICTIONARY_OF(id, MGLStyleValue<T> *) *) *)stops attributeName:(NSString *)attributeName options:(nullable NS_DICTIONARY_OF(MGLStyleFunctionOption, id) *)options;
 
 #pragma mark Accessing the Parameters of a Composite Function
-
-/**
- The modes used to interpolate property values between map zoom level changes or
- over a range of feature attribute values.
- */
-@property (nonatomic) MGLInterpolationMode interpolationMode;
 
 /**
  A string that specifies the feature attribute key whose value be used as the function
--- a/platform/darwin/src/MGLStyleValue.h
+++ b/platform/darwin/src/MGLStyleValue.h
@@ -240,7 +240,7 @@ MGL_EXPORT
  @param stops A dictionary associating zoom levels with style values.
  @return An `MGLStyleFunction` object with the given interpolation base and stops.
  */
-- (instancetype)initWithInterpolationBase:(CGFloat)interpolationBase stops:(NS_DICTIONARY_OF(NSNumber *, MGLStyleValue<T> *) *)stops NS_DESIGNATED_INITIALIZER;
+- (instancetype)initWithInterpolationBase:(CGFloat)interpolationBase stops:(NS_DICTIONARY_OF(NSNumber *, MGLStyleValue<T> *) *)stops __attribute__((deprecated("Use +[MGLStyleValue valueWithInterpolationMode:cameraStops:options:]")));
 
 #pragma mark Accessing the Parameters of a Function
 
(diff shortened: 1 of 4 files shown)


boundsj:
This PR adds back `MGLStyleFunction` so that style functions can be considered more abstractly in code and documentation. 

A constraint here is that we don't want to break backwards compatibility with 3.4.x's `MGLStyleFunction` and code that may have expected not only to reference the class but also its `interpolationBase` and `stops` properties. Concurrently, we want `MGLStyleFunction` to be a base class for all of the concrete function classes and not a kind of wrapper for `MGLCameraStyleFunction` specifically.

To do this I:

- Removed `-@compatibility_alias MGLStyleFunction MGLCameraStyleFunction`
- Added back a `MGLStyleFunction` but documented it as an abstract base class that is not to be instantiated directly or subclassed
- Added `interpolationBase` and `stops` properties to `MGLStyleFunction` so it is fully backwards compatible

The main tension point was the `stops` property combined with lightwieght generics:  Since `MGLStyleFunction` needs `stops` to be backwards compatible but not all of the concrete function subclasses share the same generic annotation types, I intentionally added a plain `NSDictionary` `stops` property to the abstract `MGLStyleFunction` with covariants in each of the subclasses with the fully expressed, lightweight generic annotations. I think this allows for full backwards compatibility, stays out of the way when dealing with the subclasses, and still allows for (in Swift) iterating over functions, cast in the abstract, and casting stops even for the most complex (composite) case:

```
// circleStyleLayer.circleRadius  previously set to a `MGLCompositeStyleFunction`
if let returnedCircleRadius = circleStyleLayer.circleRadius as? MGLStyleFunction<NSNumber> {
  if let returnedStops = returnedCircleRadius.stops as NSDictionary? as? [NSNumber: [NSNumber: MGLStyleValue<NSNumber>]] {
    let lhs: MGLStyleValue<NSNumber> = returnedStops[0]!.values.first!
    let rhs: MGLStyleValue<NSNumber> = radiusCompositeExponentialOrIntervalStops[0]!.values.first!
                XCTAssertEqual(lhs, rhs)
   }
}
```

platform/darwin/src/MGLStyleValue.h
```diff
@@ -62,32 +63,70 @@ MGL_EXPORT
...
+    range of feature attribute values for each outer zoom level.
+ @param sourceStops A dictionary associating feature attributes with style values.
+ @param attributeName Specifies the feature attribute to take as the function
+    input.
+ @param options A dictionary containing `MGLStyleFunctionOption` values that
+    specify how a function is applied.
+ @return An `MGLCompositeStyleFunction` object with the given interpolation mode,
+    composite stops, attribute name, and options.
+ */
++ (instancetype)valueWithInterpolationMode:(MGLInterpolationMode)interpolationMode compositeStops:(NS_DICTIONARY_OF(id, NS_DICTIONARY_OF(id, MGLStyleValue<T> *) *) *)compositeStops attributeName:(NSString *)attributeName options:(nullable NS_DICTIONARY_OF(MGLStyleFunctionOption, id) *)options;
```
1ec5: If each stop is “interpreted as a source function”, should the type of `compositeStops` be `NSDictionary<id, NSDictionary<id, MGLSourceStyleFunction<T> *> *> *`?

platform/darwin/src/MGLStyleValue.h
```diff
@@ -135,77 +174,261 @@ MGL_EXPORT
...
-@property (nonatomic, copy) NS_DICTIONARY_OF(id, MGLStyleValue<T> *) *stops;
+/**
+ A dictionary associating zoom levels with style values.
+ */
+@property (nonatomic, copy) NSDictionary *stops;
 
-// TODO: API docs
+/**
+ The exponential interpolation base of the function’s interpolation curve.
+ */
```
1ec5: Document the fact that `interpolationBase` may not apply to all style functions.

platform/darwin/src/MGLStyleValue.h
```diff
@@ -135,77 +174,261 @@ MGL_EXPORT
...
 @property (nonatomic) CGFloat interpolationBase;
 
 @end
```
1ec5: For backwards compatibility, MGLStyleFunction also needs the initializers that it had in v3.4.0. These initializers can be marked as deprecated.

platform/darwin/src/MGLStyleValue.h
```diff
@@ -135,77 +174,261 @@ MGL_EXPORT
...
 
-@compatibility_alias MGLStyleFunction MGLCameraStyleFunction;
+/**
+ An `MGLCameraStyleFunction` is a value function defining a style value that changes
+ as the zoom level changes. The layout and paint attribute properties of an
+ `MGLStyleLayer` object can be set to `MGLCameraStyleFunction` objects. Use a zoom
+ level function to create the illusion of depth and control data density.
```
1ec5: “Zoom level functions” are now known as “camera functions”.

1ec5:
> I intentionally added a plain NSDictionary stops property to the abstract MGLStyleFunction with covariants in each of the subclasses with the fully expressed, lightweight generic annotations. I think this allows for full backwards compatibility, stays out of the way when dealing with the subclasses, and still allows for (in Swift) iterating over functions, cast in the abstract, and casting stops even for the most complex (composite) case

Note that this approach technically doesn’t guarantee full backwards compatibility with Swift code:

```swift
let circleRadius = circleStyleLayer.circleRadius as! MGLStyleFunction<NSNumber>
circleStyleLayer.circleRadius = circleRadius.stops[14]
```

must now be converted to at least something like this:

```swift
let circleRadius = circleStyleLayer.circleRadius as! MGLStyleFunction<NSNumber>
circleStyleLayer.circleRadius = (circleRadius.stops as! [NSNumber: MGLStyleFunction<NSNumber>])[14]
```

This isn’t a problem in Objective-C, where you can use `NSDictionary *` as a more specific type without casting. We’ve [carved out an exception to semantic versioning](https://github.com/mapbox/mapbox-gl-native/wiki/Versions-and-tagging#mapbox-ios-sdk) for Swift bridging changes, although it’s a last resort.

platform/darwin/src/MGLStyleValue.h
boundsj: This was me doing a poor job of trying to clarify the opposite: "composite" in this case means a composite of stops dictionaries not functions (see discussion in https://github.com/mapbox/mapbox-gl-native/pull/7596#issuecomment-276530479). In https://github.com/mapbox/mapbox-gl-native/pull/7943/commits/d46559a5ff8791af9fb0b39cce20d31c13a57bb8, I punted on trying to explain this in the class factory method in favor of the explanation in the `MGLCompositeStyleFunction` `stops` property

platform/darwin/src/MGLStyleValue.h
boundsj: https://github.com/mapbox/mapbox-gl-native/pull/7943/commits/1fa2bb7498f786d0056e908a5971ff910da3c6f2

platform/darwin/src/MGLStyleValue.h
boundsj: This will probably look a bit strange no matter what we do. I think https://github.com/mapbox/mapbox-gl-native/pull/7943/commits/406e647705665e96ef8ba9e0575056ac0ff53289 is a workable solution until a semver bump when we can more officially make this class abstract.

The initializers have been brought back and an actual instance of `MGLStyleFunction` is created and returned if you use them. The conversion methods have been updated to coerce `MGLStyleFunction` instances into `MGLCameraStyleFunction` instances with the appropriate interpolation mode. However, documentation attempts to point developers to the MGLStyleValue class factory methods to avoid all of this.

platform/darwin/src/MGLStyleValue.h
```diff
@@ -62,32 +63,68 @@ MGL_EXPORT
...
+
+ @param interpolationMode The mode used to interpolate property values between 
+    map zoom level changes.
+ @param cameraStops A dictionary associating zoom levels with style values.
+ @param options A dictionary containing `MGLStyleFunctionOption` values that 
+    specify how a function is applied.
+ @return An `MGLCameraStyleFunction` object with the given interpolation mode,
+    camera stops, and options.
+ */
++ (instancetype)valueWithInterpolationMode:(MGLInterpolationMode)interpolationMode cameraStops:(NS_DICTIONARY_OF(id, MGLStyleValue<T> *) *)cameraStops options:(nullable NS_DICTIONARY_OF(MGLStyleFunctionOption, id) *)options;
```
1ec5: Remember to use the `NS_…_OF` macros until we drop iOS 8 support.

boundsj: It looks like we forgot to do this in the `3.4.0` release. In fbb8068 and 3b90d85 I updated the deprecated methods and everything in the resurrected `MGLStyleFunction` to use the macro. The only use of `NSDictionary` (with no lightweight generic attributes) in `MGLStyleValue` now is for the `stops` property of `MGLStyleFunction` for reasons already discussed in this PR.

platform/darwin/src/MGLStyleValue.h
```diff
@@ -135,77 +172,315 @@ MGL_EXPORT
...
+ */
+- (instancetype)initWithInterpolationBase:(CGFloat)interpolationBase stops:(NS_DICTIONARY_OF(NSNumber *, MGLStyleValue<T> *) *)stops NS_DESIGNATED_INITIALIZER;
+
+#pragma mark Accessing the Parameters of a Function
+
+/**
+ The modes used to interpolate property values between map zoom level changes or
+ over a range of feature attribute values.
+ */
 @property (nonatomic) MGLInterpolationMode interpolationMode;
```
1ec5: Could we avoid effectively adding an `interpolationMode` property to MGLStyleFunction that would only sometimes be applicable?

platform/darwin/src/MGLStyleValue.h
```diff
@@ -135,77 +172,315 @@ MGL_EXPORT
...
+ @note Do not create instances of `MGLStyleFunction` unless it is required for
+    backwards compatiblity with your application code. You should create and use
+    instances of `MGLCameraStyleFunction` to specify how properties will
+    be visualized at different zoom levels.
+
+ @param interpolationBase The exponential base of the interpolation curve.
+ @param stops A dictionary associating zoom levels with style values.
+ @return An `MGLStyleFunction` object with the given interpolation base and stops.
+ */
+- (instancetype)initWithInterpolationBase:(CGFloat)interpolationBase stops:(NS_DICTIONARY_OF(NSNumber *, MGLStyleValue<T> *) *)stops NS_DESIGNATED_INITIALIZER;
```
1ec5: If you mark this as a designated initializer, does that mean all the subclass initializers have to call this method? Or would removing `NS_DESIGNATED_INITIALIZER` be considered a breaking change?

platform/darwin/src/MGLStyleValue.mm
```diff
@@ -62,16 +62,59 @@ - (NSUInteger)hash {
...
 @end
 
+@implementation MGLStyleFunction
+
++ (instancetype)functionWithStops:(NSDictionary *)stops {
+    return [[self alloc] initWithInterpolationBase:1.0 stops:stops];
+}
+
++ (instancetype)functionWithInterpolationBase:(CGFloat)interpolationBase stops:(NSDictionary *)stops {
+    return [[self alloc] initWithInterpolationBase:interpolationBase stops:stops];
```
1ec5: Can we initialize an MGLCameraStyleFunction here instead (and in `+functionWithStops:`, but not `-init` or `-initWithInterpolationBase:stops:`)? If there are warnings associated with the designated initializer, we can suppress the warnings.

platform/darwin/src/MGLStyleValue.h
1ec5: Ideally, in a future release, we’d end up making all `-init…` methods private (as would be expected of class clusters). So we should make all the new `-init…` methods (like in `MGLCameraStyleFunction`) private and leave this one public but deprecated, for backwards compatibility.

platform/darwin/src/MGLStyleValue.mm
```diff
@@ -62,16 +62,59 @@ - (NSUInteger)hash {
...
+}
+
+- (BOOL)isEqual:(MGLStyleFunction *)other {
+    return ([other isKindOfClass:[self class]]
+            && [other.stops isEqualToDictionary:self.stops]
+            && other.interpolationBase == self.interpolationBase);
+}
+
+- (NSUInteger)hash {
+    return  self.stops.hash + self.interpolationBase;
```
1ec5: Use `@(self.interpolationBase).hash`, not `interpolationBase` itself, to further minimize collisions. There are similar problems with `interpolationMode` in the subclasses. Feel free to take care of this in a separate PR.

platform/darwin/src/MGLStyleValue.h
boundsj: > If you mark this as a designated initializer, does that mean all the subclass initializers have to call this method? 

Yes

Or would removing NS_DESIGNATED_INITIALIZER be considered a breaking change?

I don't think so. We never suggested that developers should subclass MGLStyleFunction.

In 7cbeba7, I exposed only MGLStyleFunction's initializer publicly but
deprecated it and removed its designated initializer status. Internally this allows us to ignore the deprecated initializer in subclasses.

platform/darwin/src/MGLStyleValue.h
boundsj: In ee89e94 I removed it in `MGLStyleFunction` and added it back to the concrete subclasses. This sort of makes sense because an `MGLStyleFunction` technically does not (did not) have an interpolation mode. In the future, we should add the property back to `MGLStyleFunction` because every actual function class does have an interpolation mode.

platform/darwin/src/MGLStyleValue.mm
boundsj: Done 2f414f2

platform/darwin/src/MGLStyleValue.mm
boundsj: Done 1b67069

platform/darwin/src/MGLStyleValue.h
1ec5: > every actual function class does have an interpolation mode

Ah, I misunderstood. In that case, it’s fine (and preferable) to have it in MGLStyleFunction to avoid duplication.

boundsj: Yeah. For now, it'll be a bit of a lie, but still reasonable. 

`MGLStyleFunction`'s interpolation mode will be exponential even though it did not have such a concept before in 3.4.x. However, any `MGLStyleFunction` instance that is put to use in 3.5.0 is treated like a camera function with an exponential interpolation mode so I feel ok about that. I'll put it back on `MGLStyleFunction`.

nitrag:
I'm super lost.

```swift
let carLayer = MGLSymbolStyleLayer(identifier: "cars-layer", source: carSource)
style.setImage(UIImage(named: "car-red")!, forName: "car-red")
style.setImage(UIImage(named: "car-blue")!, forName: "car-blue")
style.setImage(UIImage(named: "car-silver")!, forName: "car-silver")
style.setImage(UIImage(named: "car-green")!, forName: "car-green")
style.setImage(UIImage(named: "car-orange")!, forName: "car-orange")
carLayer.iconImageName = MGLSourceStyleFunction(
    interpolationMode: .identity,
    sourceStops: nil,
    attributeName: "car-color",
    options: [.defaultValue: MGLStyleValue<NSString>(rawValue: "car-silver")]
)
carLayer.iconImageName = MGLCameraStyleFunction(
    interpolationMode: .identity,
    sourceStops: nil,
    attributeName: "car-color",
    options: [.defaultValue: MGLStyleValue<NSString>(rawValue: "car-silver")]
)
```

regardless of which one I choose:
`'NSInvalidArgumentException', reason: 'This property can only be set to camera functions. Use +[MGLStyleValue valueWithInterpolationMode:cameraStops:options:] instead.'` 

but I need `attributeName` to tell it what field? No??

jfirebaugh:
Data-driven styling for iconImageName is not in any released iOS SDK version yet. It'll be in 3.6.0.

If you're ever in doubt about SDK compatibility, https://www.mapbox.com/mapbox-gl-js/style-spec/ has support matrixes for every property (albeit you'll need to translate from SDK property name to style spec property name).

nitrag:
Ahhhhhh. I'll remember this for the future. Thanks!

1ec5:
The documentation for [`MGLSymbolStyleLayer.iconImageName`](https://www.mapbox.com/ios-sdk/api/3.5.2/Classes/MGLSymbolStyleLayer.html#/c:objc%28cs%29MGLSymbolStyleLayer%28py%29iconImageName) in v3.5.2 says it supports camera functions but not source functions. We use the same automated process to generate the property’s implementation, documentation, and the exception message you ran into, so you can be sure of its accuracy.

> albeit you'll need to translate from SDK property name to style spec property name

The iOS SDK’s documentation comes with [a full correspondence table](https://www.mapbox.com/ios-sdk/api/3.5.2/for-style-authors.html#configuring-the-map-content-s-appearance).

wr6524798:
please, when I use Data-driven styling for iconImageName,the Application crash,  I want use Data-driven styling for iconImageName,when the 3.6.0. release?

1ec5:
@wr6524798, this closed pull request isn’t the best place for new bug reports. Please open a new issue about the crash you’re seeing, and be sure to include the crash trace so we can properly diagnose and prioritize the problem. Thank you!

PR #7944, opened 2017-02-04
[core] Enable property functions for text-{field,transform}

anandthakker:
So, I got to what I thought should at least compile here so that I could try it / iterate, but ran into this semi-mysterious compiler error:

```
▸ Compiling parser.cpp

❌  /Users/anand/c/mapbox/mapbox-gl-native/include/mbgl/style/conversion/function.hpp:333:20: no matching constructor for initialization of 'CompositeFunction<std::__1::basic_string<char> >'

            return CompositeFunction<T>(*propertyString, convertedStops, *defaultValue);
                   ^                    ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~



❌  /Users/anand/c/mapbox/mapbox-gl-native/include/mbgl/style/conversion/function.hpp:333:20: no matching constructor for initialization of 'CompositeFunction<mbgl::style::TextTransformType>'

            return CompositeFunction<T>(*propertyString, convertedStops, *defaultValue);
      ^
```

Telling that this has something to do with style functions, and the types here are string and `TextTransformType` -- precisely the ones I'm expecting to get via `layout.evaluate<TextField>(zoom, *feature)` and `layout.evaluate<TextTransform>(zoom, *feature)` in [symbol_layout](https://github.com/mapbox/mapbox-gl-native/commit/095cdab4fd8055b142ce12b5b6fbbe778f48ddc0#diff-f27a3a157c549a13f8fe1ee8c8157ea8).    (This seems to work fine for `IconOffset` and `IconRotate`.)

I'll keep messing w/ it, just flagging here in case it's something I'm unlikely to find on my own.  

cc @jfirebaugh

### DIFF
--- a/platform/darwin/src/MGLStyleValue_Private.h
+++ b/platform/darwin/src/MGLStyleValue_Private.h
@@ -365,41 +365,6 @@ private: // Private utilities for converting from mgl to mbgl values
         return sourceFunction;
     }
 
-    mbgl::style::SourceFunction<MBGLType> toMBGLExponentialSourceFunction(MGLSourceStyleFunction<ObjCType> *sourceStyleFunction) {
-        __block std::map<float, MBGLType> stops = {};
-        [sourceStyleFunction.stops enumerateKeysAndObjectsUsingBlock:^(NSNumber * _Nonnull zoomKey, MGLStyleValue<ObjCType> * _Nonnull stopValue, BOOL * _Nonnull stop) {
-            NSCAssert([stopValue isKindOfClass:[MGLStyleValue class]], @"Stops should be MGLStyleValues");
-            auto mbglStopValue = toPropertyValue(stopValue);
-            NSCAssert(mbglStopValue.isConstant(), @"Stops must be constant");
-            stops[zoomKey.floatValue] = mbglStopValue.asConstant();
-        }];
-        mbgl::style::ExponentialStops<MBGLType> exponentialStops = {stops, (float)sourceStyleFunction.interpolationBase};
-        mbgl::style::SourceFunction<MBGLType> sourceFunction = {sourceStyleFunction.attributeName.UTF8String, exponentialStops};
-        setDefaultMBGLValue(sourceStyleFunction, sourceFunction);
-        return sourceFunction;
-    }
-
-    mbgl::style::SourceFunction<MBGLType> toMBGLIntervalSourceFunction(MGLSourceStyleFunction<ObjCType> *sourceStyleFunction) {
-        __block std::map<float, MBGLType> stops = {};
-        [sourceStyleFunction.stops enumerateKeysAndObjectsUsingBlock:^(NSNumber * _Nonnull zoomKey, MGLStyleValue<ObjCType> * _Nonnull stopValue, BOOL * _Nonnull stop) {
-            NSCAssert([stopValue isKindOfClass:[MGLStyleValue class]], @"Stops should be MGLStyleValues");
-            auto mbglStopValue = toPropertyValue(stopValue);
-            NSCAssert(mbglStopValue.isConstant(), @"Stops must be constant");
-            stops[zoomKey.floatValue] = mbglStopValue.asConstant();
-        }];
-        mbgl::style::IntervalStops<MBGLType> intervalStops = {stops};
-        mbgl::style::SourceFunction<MBGLType> sourceFunction = {sourceStyleFunction.attributeName.UTF8String, intervalStops};
-        setDefaultMBGLValue(sourceStyleFunction, sourceFunction);
-        return sourceFunction;
-    }
-
-    mbgl::style::SourceFunction<MBGLType> toMBGLIdentitySourceFunction(MGLSourceStyleFunction<ObjCType> *sourceStyleFunction) {
-        mbgl::style::IdentityStops<MBGLType> identityStops;
-        mbgl::style::SourceFunction<MBGLType> sourceFunction = {sourceStyleFunction.attributeName.UTF8String, identityStops};
-        setDefaultMBGLValue(sourceStyleFunction, sourceFunction);
-        return sourceFunction;
-    }
-
     void setDefaultMBGLValue(MGLSourceStyleFunction<ObjCType> *sourceStyleFunction, mbgl::style::SourceFunction<MBGLType> &sourceFunction) {
         if (sourceStyleFunction.defaultValue) {
             NSCAssert([sourceStyleFunction.defaultValue isKindOfClass:[MGLStyleConstantValue class]], @"Default value must be constant");
--- a/platform/darwin/src/MGLStyleValue_Private.h
+++ b/platform/darwin/src/MGLStyleValue_Private.h
@@ -185,7 +185,11 @@ private: // Private utilities for converting from mgl to mbgl values
         return @(color.mgl_color.stringify().c_str());
     }
     
-    NSObject* toRawStyleSpecValue(NSObject *rawMGLValue) {
+    template <typename MBGLEnum = MBGLType,
+    class = typename std::enable_if<!std::is_enum<MBGLEnum>::value>::type,
+    typename MGLEnum = ObjCEnum,
+    class = typename std::enable_if<!std::is_enum<MGLEnum>::value>::type>
+    NSObject* toRawStyleSpecValue(NSObject *rawMGLValue, MBGLEnum &mbglValue) {
         if ([rawMGLValue isKindOfClass:[NSValue class]]) {
             const auto rawNSValue = (NSValue *)rawMGLValue;
             if (strcmp([rawNSValue objCType], @encode(CGVector)) == 0) {
@@ -197,6 +201,19 @@ private: // Private utilities for converting from mgl to mbgl values
         // noop pass-through plain NSObject-based items
         return rawMGLValue;
     }
+    
+    // hack to convert enum => string values, and just pass-through other plain NSObject values.
+    template <typename MBGLEnum = MBGLType,
+    class = typename std::enable_if<std::is_enum<MBGLEnum>::value>::type,
+    typename MGLEnum = ObjCEnum,
+    class = typename std::enable_if<std::is_enum<MGLEnum>::value>::type>
+    NSString* toRawStyleSpecValue(ObjCType rawValue, MBGLEnum &mbglValue) {
+        MGLEnum mglEnum;
+        [rawValue getValue:&mglEnum];
+        return @(mbgl::Enum<MGLEnum>::toString(mglEnum));
+    }
+    
+
     
     NSObject* toRawStyleSpecValue(MGLStyleFunction<ObjCType>* styleFunction) {
         NSMutableDictionary * rawFunction = [NSMutableDictionary new];
@@ -226,7 +243,8 @@ private: // Private utilities for converting from mgl to mbgl values
             // zoom-only function (no default value)
             __block NSMutableArray *stops = [[NSMutableArray alloc] init];
             [styleFunction.stops enumerateKeysAndObjectsUsingBlock:^(NSNumber * _Nonnull zoomKey, MGLStyleConstantValue<ObjCType> * _Nonnull outputValue, BOOL * _Nonnull stop) {
-                NSArray *rawStop = @[zoomKey, toRawStyleSpecValue([outputValue rawValue])];
+                MBGLType dummyMbglValue;
+                NSArray *rawStop = @[zoomKey, toRawStyleSpecValue([outputValue rawValue], dummyMbglValue)];
                 [stops addObject:rawStop];
             }];
             rawFunction[@"stops"] = stops;
@@ -237,7 +255,8 @@ private: // Private utilities for converting from mgl to mbgl values
             // property-only function
             __block NSMutableArray *stops = [[NSMutableArray alloc] init];
             [styleFunction.stops enumerateKeysAndObjectsUsingBlock:^(NSObject * _Nonnull propertyKey, MGLStyleConstantValue<ObjCType> * _Nonnull outputValue, BOOL * _Nonnull stop) {
-                NSArray *rawStop = @[propertyKey, toRawStyleSpecValue([outputValue rawValue])];
+                MBGLType dummyMbglValue;
+                NSArray *rawStop = @[propertyKey, toRawStyleSpecValue([outputValue rawValue], dummyMbglValue)];
                 [stops addObject:rawStop];
             }];
             rawFunction[@"stops"] = stops;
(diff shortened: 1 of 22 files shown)


anandthakker:
Ah, hmm:  the generated `MGLSymbolStyleLayerTests.mm` attempts to construct `CompositeFunction<std::string>` and `CompositeFunction<mbgl::style::TextTransformType>`, which are specifically what the two errors are complaining about.  Seems promising.

anandthakker:
Okay, maybe not: I didn't trust the "Comiling parser.cpp" in the `make` output, but building in Xcode where I can trace the error, it does indeed originate there.

anandthakker:
Alright, new hypothesis: the new data driven properties I'm adding here, `text-field` and `text-transform`, appear to the the first non-interpolatable properties to be implemented.  I think perhaps that the problem is:
 - We [enforce at compile-time](https://github.com/mapbox/mapbox-gl-native/blob/master/include/mbgl/style/function/composite_function.hpp#L26-L44) that if `T` is not interpolatable, then `CompositeFunction<T>::Stops` must be a map from `float` to `IntervalStops` or `CategoricalStops` (_not_ `ExponentialStops`).
 - However, [here in `conversion/function.hpp`](https://github.com/mapbox/mapbox-gl-native/blob/master/include/mbgl/style/conversion/function.hpp#L326-L333), we construct it with ExponentialStops.  (I think this couldn't actually happen at runtime?)

anandthakker:
Took a crack at solving the issue w/ converting non-interpolatable in 941c7a8, using `std::enable_if` to differing results based on the predicate `Interpolatable<T>`.

@jfirebaugh is there a cleaner or more idiomatic way to do this?

platform/darwin/src/MGLSymbolStyleLayer.h
```diff
@@ -619,7 +619,8 @@ MGL_EXPORT
 
 /**
  Value to use for a text label. Feature properties are specified using tokens
- like {field_name}.
+ like {field_name}.  (Token replacement is only supported for literal text-field
+ values--not for data-driven property functions.)
```
1ec5: Did you add this note manually after running `make style-code`? If so, it’ll be wiped away the next time someone runs `make style-code`. Make sure to put any documentation overrides in platform/darwin/scripts/style-spec-overrides-v8.json.

1ec5: Surround “text-field” in backticks. This allows generate-style-code.js to replace it with `text` (the overridden name of this property).

1ec5: What is a “data-driven property function”? An `MGLSourceStyleFunction`?

anandthakker: Nope!  The note comes directly from the description of this property in v8.json

1ec5: Ah, I was confused because [this page](https://www.mapbox.com/mapbox-gl-js/style-spec/) doesn’t include the new wording.

anandthakker: Hm, yeah good point, I should have been more precise with the language here.  I'll ticket this in -gl-js, along with the backtick thing.

1ec5: An override in style-spec-overrides-v8.json is probably inevitable, because the style specification doesn’t yet refer to “source functions”: mapbox/mapbox-gl-js#4154.

anandthakker: Ah, okay.  I suppose, in that case, that the least confusion-causing thing to do might be to just _drop_ the parenthetical in the `style-spec-overrides-v8.json`, since there's no established vocabulary for these function types yet?

@1ec5 what do you think?

1ec5: It’s a good point to document. There’s no harm in using one term in the style specification and overriding it here to refer to the specific class; that’s what style-spec-overrides-v8.json was designed for.

anandthakker: Ah okay, I misunderstood: there _is_ a publicly documented way to refer to it.  It'll need to refer to both `MGLSourceStyleFunction` and `MGLCompositeStyleFunction`.

jfirebaugh:
`CameraFunction` and `SourceFunction` avoid attempting to construct `ExponentialStops` for a non-interpolatable type via [`StopsConverter`](https://github.com/mapbox/mapbox-gl-native/blob/3212e543b79c0a0d2836e32326a66ca830101777/include/mbgl/style/conversion/function.hpp#L147-L182), which is templated on the possible variant types (e.g. from [here](https://github.com/mapbox/mapbox-gl-native/blob/3212e543b79c0a0d2836e32326a66ca830101777/include/mbgl/style/function/source_function.hpp#L19-L29)). The implementation of `StopsConverter` loops through the possible stop types statically using parameter pack expansion, testing each in turn. (This code is far from obvious -- sorry I neglected to comment it.)

We should follow a similar approach for `CompositeFunction`. And likewise default to `IntervalStops` if the type is non-interpolatable.

anandthakker:
@boundsj @1ec5 I'm seeing a compiler error in `MGLStyleValue_Private.h` that I think is analogous to the one discussed above ([here](https://github.com/mapbox/mapbox-gl-native/pull/7944#issuecomment-277482968) and [here](https://github.com/mapbox/mapbox-gl-native/pull/7944#issuecomment-277583200)): namely, there's a [branch of code](https://github.com/mapbox/mapbox-gl-native/blob/master/platform/darwin/src/MGLStyleValue_Private.h#L152-L174) that is only reachable at runtime if the underlying type for the MGLCompositeStyleFunction is interpolatable, but it fails at compile-time because we are trying to construct a composite function with a string output type.

Could you point me in the right direction for the ideal way to resolve this?

boundsj:
@anandthakker [these conditionals in MGLStyleLayer.mm.ejs](https://github.com/mapbox/mapbox-gl-native/blob/dds-text-field-transform/platform/darwin/src/MGLStyleLayer.mm.ejs#L184) generate an implementation that is compatible with the static type checking by calling either the `toPropertyValue` or `toInterpolatablePropertyValue` methods in `MGLStyleValue_Private.h` depending on whether the property is interpolatable or not.

Following that convention, the existing `toDataDrivenPropertyValue` in  `MGLStyleValue_Private.h` could be renamed `toInterpolatableDataDrivenPropertyValue` and a `toDataDrivenPropertyValue` conversion method could be added (for the non-interpolatable case) that does not attempt to create invalid functions like the one noted above. `MGLStyleLayer.mm.ejs` would then need to be updated with another conditional to call that method for these new non-interpolatable data driven property values.

Note that we don't have this problem with the getter implementation (going from the C++ objects back to Objective-C) since we use the `PropertyValueEvaluator` in `MGLStyleValue_Private.h` with stops visitors that works with all the possible variant types. Since the setters (Objective-C to C++) can't take advantage of generics in this way, I'm not sure if there is a better option than what I described above (adding a new specific method and calling it based on the property capabilities).

cc @1ec5 @jfirebaugh

1ec5:
> Following that convention, the existing toDataDrivenPropertyValue in MGLStyleValue_Private.h could be renamed toInterpolatableDataDrivenPropertyValue and a toDataDrivenPropertyValue conversion method could be added (for the non-interpolatable case) that does not attempt to create invalid functions like the one noted above.

To avoid code duplication, we could split off the MGLCompositeStyleFunction case from `toDataDrivenPropertyValue()` into two `toCompositeDataDrivenPropertyValue()` methods, one of them a stub that’s conditioned (via `std::enable_if`) on whether `MBGLType` is `TextTransformType` etc.

anandthakker:
Followup on https://github.com/mapbox/mapbox-gl-native/pull/7944#issuecomment-278390927: As I started digging into resolving the compilation issues in the MGLCompositeStyleFunction => mbgl::style::CompositeFunction conversion today, it felt like I was solving the same problems that are being solved by `mbgl::style::conversion::*`.

I'm now thinking that we should accomplish the MGL => mbgl conversion — at least for property values — by having the darwin platform bindings implement [the Value conversions required](https://github.com/mapbox/mapbox-gl-native/blob/master/include/mbgl/style/conversion.hpp#L30-L52) to leverage the existing `mbgl::style::conversion` system.  This will hopefully mean less code/logic duplication and better maintainability.

@1ec5 notes:

> The only catch I can think of is that you can't overload a method based on ObjC class types and expect it to respect inheritance.

I'll dig into this tomorrow AM -- hopefully the ObjC overloading thing is not a dealbreaker

src/mbgl/style/function/identity_stops.cpp
```diff
@@ -12,6 +14,11 @@ optional<float> IdentityStops<float>::evaluate(const Value& value) const {
 }
 
 template <>
+optional<std::string> IdentityStops<std::string>::evaluate(const Value& value) const {
+    return value.get<std::string>();
```
jfirebaugh: Needs an `if (!value.is<std::string>()) return {}` guard.

anandthakker: 👍  Looks like the `<float>` one above also needs one of these?

jfirebaugh: No, `numericValue` handles that.

anandthakker: Oh, I see -- that's built into `numericValue`.

anandthakker:
Couple pieces of unfinished business addressed since @jfirebaugh 's 🍏 :

1. The [conversion system refactor](https://github.com/mapbox/mapbox-gl-native/pull/8026) failed to account for Enums properly.  Handled that here: https://github.com/mapbox/mapbox-gl-native/pull/7944/commits/462cd0d40ff8bc26abb009f90901fc19039ca976 (with a small fixup for colors in 3b83f01)
2. The MGLStyleLayerTests template needed a couple of fixes.  Easiest to read those changes [here](https://github.com/mapbox/mapbox-gl-native/compare/602b755a88c29a5e65ab0cde4f4349207b2c9c4e...dds-text-field-transform#diff-6b846658359efdccf5cd8807eec95f56)

@boundsj or @1ec5 could I get 👀 from one of you on these changes?

1ec5:
Can you add tests to exercise enums, colors, and other types, which should’ve failed in #8026?

platform/darwin/test/MGLFillStyleLayerTests.mm
```diff
@@ -75,12 +75,10 @@ - (void)testProperties {
...
         XCTAssertEqualObjects(layer.fillAntialiased, defaultStyleValue,
                               @"fillAntialiased should return the default value after being unset.");
-
-        functionStyleValue = [MGLStyleValue<NSNumber *> valueWithInterpolationMode:MGLInterpolationModeIdentity sourceStops:nil attributeName:@"" options:nil];
-        XCTAssertThrowsSpecificNamed(layer.fillAntialiased = functionStyleValue, NSException, NSInvalidArgumentException, @"MGLStyleValue should raise an exception if it is applied to a property that cannot support it");
-
-        functionStyleValue = [MGLStyleValue<NSNumber *> valueWithInterpolationMode:MGLInterpolationModeInterval compositeStops:@{@18: constantStyleValue} attributeName:@"" options:nil];
-        XCTAssertThrowsSpecificNamed(layer.fillAntialiased = functionStyleValue, NSException, NSInvalidArgumentException, @"MGLStyleValue should raise an exception if it is applied to a property that cannot support it");        
+functionStyleValue = [MGLStyleValue<NSNumber *> valueWithInterpolationMode:MGLInterpolationModeIdentity sourceStops:nil attributeName:@"" options:nil];
```
1ec5: Missing indentation.

platform/darwin/src/MGLStyleValue_Private.h
```diff
@@ -181,11 +181,16 @@ class MGLStyleValueTransformer {
...
+     This allows us to use `std::enable_if` to test (at compile time) whether or not MBGLType is an Enum.
+     */
+    template <typename MBGLEnum = MBGLType,
+    class = typename std::enable_if<!std::is_enum<MBGLEnum>::value>::type,
+    typename MGLEnum = ObjCEnum,
+    class = typename std::enable_if<!std::is_enum<MGLEnum>::value>::type>
+    NSObject* toRawStyleSpecValue(NSObject *rawMGLValue, MBGLEnum &mbglValue) {
         if ([rawMGLValue isKindOfClass:[NSValue class]]) {
             const auto rawNSValue = (NSValue *)rawMGLValue;
             if (strcmp([rawNSValue objCType], @encode(CGVector)) == 0) {
```
1ec5: By the way, I see special handling for CGVector values, but not NSEdgeInsets/UIEdgeInsets values. Do we need it too? Here’s the [full correspondence](https://www.mapbox.com/ios-sdk/api/3.5.0-beta.1/for-style-authors.html#setting-attribute-values) between JSON and Foundation types.

anandthakker:
> Can you add tests to exercise enums, colors, and other types, which should’ve failed in #8026?

 - Colors are covered in `MGLStyleValueTests::testFunctionsWithDataDrivenProperties()`
 - Enums and strings are covered in `MGLSymbolLayerTests::testProperties()` (`text-transform` and `text-field`, respectively)
 - Numbers and strings are covered covered in `MGLCircleLayerTests::testProperties()` (`circle-radius`)

I think what's missing here are the array-based types -- which, you're right, include CGVectorValue and UIEdgeInsetValues that we'll need special-case handling for.

Currently, there is no unit test specific to `MGLStyleValueTransformer` -- instead, the code therein gets exercised via the style layer tests.  One option is to create some unit tests specific to `MGLStyleValueTransformer`.  Another is to wait for DDS support to be implemented for style properties with these array-based types, which will result in their conversions being exercised in the style layer tests.  My gut feeling is that MGLStyleValueTransformer is not an especially meaningful abstraction, and thus encoding unit tests against it would create more maintenance burden than testing value.

platform/darwin/src/MGLStyleValue_Private.h
anandthakker: This is a good catch.  I've ticketed it out [here](https://github.com/mapbox/mapbox-gl-native/issues/8088) as part of the completionist quest to propagate this approach to all of the MGL=>mbgl style property value conversions (or, at least the runtime-styling ones...)

anandthakker:
@1ec5 fixed the indentation issue and responded above on the two other issues you raised -- I think they're both covered by #8088 and not urgent to resolve before that... what do you think?

anandthakker:
@mapbox/android after fixing a small typo, the android bitrise build passed, but could I get quick  👀 on this before it goes in?  The android-related changes are conveniently in three consecutive files in the diff, starting here: https://github.com/mapbox/mapbox-gl-native/pull/7944/files#diff-ffc041b38fb81bccaa2c261dbc6b938e

PR #7952, opened 2017-02-06
Post offline pack notifications from MGLOfflinePack

### LINKED ISSUE #7908: [iOS] Mapbox Offline storage does not follow the notification usage pattern
In documentation the `MGLOfflinePackProgressChangedNotification` notification is described as being posted by `MGLOfflineStorage` instance, however when notification arrives the `object` attribute of notification is the actual pack who's progress has changed. However in Apple's documentation on `NotificationCenter`, it states that the `object` parameter must contain the object performing the post.

I think one of the following two implementations would be better and more consistent with rest of the codebase posting/observing notifications:

1. The notification should be posted by the `MGLOfflinePack` instance that has its progress updated.
2. The `object` attribute of the notification should be the storage and the `MGLOfflinePack` should be stored in the `userInfo` dictionary (thus removing the need for redundant `progress` and `state` attribute storage under same `userInfo` instance).

I understand that this is a very breaking change and should (if at all released) follow the semantic versioning.

eimantas:
Fixes #7908.

### DIFF
--- a/platform/darwin/src/MGLOfflinePack.mm
+++ b/platform/darwin/src/MGLOfflinePack.mm
@@ -4,6 +4,8 @@
 #import "MGLOfflineRegion_Private.h"
 #import "MGLTilePyramidOfflineRegion.h"
 
+#import "NSValue+MGLAdditions.h"
+
 #include <mbgl/storage/default_file_source.hpp>
 
 /**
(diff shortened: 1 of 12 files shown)


eimantas:
Done. Please check the changelog entries.

1ec5:
Thanks again!

PR #7954, opened 2017-02-06
[build] Remove mkdirp from node deps

brunoabinader:
Simple replacement for `mkdirp`, reducing the amount of npm development dependencies.

### DIFF
--- a/package.json
+++ b/package.json
@@ -22,7 +22,6 @@
     "ejs": "^2.4.1",
     "express": "^4.11.1",
     "lodash": "^4.16.4",
-    "mkdirp": "^0.5.1",
     "node-cmake": "^1.2.1",
     "pixelmatch": "^4.0.2",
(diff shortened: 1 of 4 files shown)


mourner:
It's a really tiny dependency and it's battle-tested across many mapbox repos. Should we really replace it with a custom script? That increases maintenance overhead.

tmpsantos:
This is a good thing I think. It removes the dependency on NPM and makes "offline building" easier, specially for embedded devices.

Rationale is on Linux embedded distributions like Yocto you can have nodejs pre-installed via apt-get, but not `mkdirp`.

jfirebaugh:
I'm aiming to eliminate the need for a mkdirp operation entirely in #7942 and a followup PR for build-version.js.

brunoabinader:
Deferring in favor of #7942.

Issue #7955, opened 2017-02-06
MGLMapView leaks MGLReachability objects


danielmatzke:
**Platform:**
macOS 10.12.3 

**Mapbox SDK version:**
0.3.0-rc.1

### Steps to trigger behavior

 1. create and destroy MGLMapView
 2. create and destroy MGLMapView
 3. see that number of persistent MGLReachability objects in Instruments is 2

### Expected behavior

MGLReachability should be released when MGLMapView object is destroyed. 

### Actual behavior

MGLReachability is not released when MGLMapView object is destroyed. Note that the MGLMapView.commonInit creates a MGLReachability object which is not shared and never released.

1ec5:
Fixed in #7956.

PR #7956, opened 2017-02-06
7955 fix MGLMapView leaks MGLReachability objects (and threads)

danielmatzke:
see https://github.com/mapbox/mapbox-gl-native/issues/7955


danielmatzke:
Thanks for the quick response. I just uploaded an similar fix for the iOS MGLMapView. I verified that the fix compiles but did not test it in an app.

PR #7966, opened 2017-02-07
Simplify version portion of user agent string and version generation

jfirebaugh:
Removes node dependency from another part of the build.


CMakeLists.txt
```diff
@@ -27,6 +27,17 @@ if(IS_CI_BUILD)
     add_compile_options(-DCI_BUILD=1)
 endif()
 
+if(EXISTS ${CMAKE_SOURCE_DIR}/.git/HEAD)
+    exec_program(
+        "git"
+        ${CMAKE_SOURCE_DIR}
+        ARGS "rev-parse --short=8 HEAD"
+        OUTPUT_VARIABLE MBGL_VERSION_REV )
+    add_definitions(-DMBGL_VERSION_REV="${MBGL_VERSION_REV}")
```
kkaefer: Can we add the definition only to the one file that actually uses it? This avoids recompiles after every commit.

kkaefer: This would be something like `set_source_files_properties(filename PROPERTIES COMPILE_DEFINITIONS MBGL_VERSION_REV=${MBGL_VERSION_REV}`

kkaefer: Seeing that we are removing the global version symbol, so we'll either have to reintroduce that, or list all the files that use the define. Also noticing that Android is not using this define yet.

platform/darwin/src/http_file_source.mm
```diff
@@ -133,8 +132,8 @@ void cancel() {
 
     // Avoid %s here because it inserts hidden bidirectional markers on macOS when the system
     // language is set to a right-to-left language.
-    [userAgentComponents addObject:[NSString stringWithFormat:@"MapboxGL/%@ (%@)",
-                                    CFSTR(MBGL_VERSION_STRING), CFSTR(MBGL_VERSION_REV)]];
+    [userAgentComponents addObject:[NSString stringWithFormat:@"MapboxGL/0.0.0 (%@)",
```
1ec5: This is a reasonable change: other than requiring a version to be here, we don’t check its value anywhere as far as I know. (Another option would be to hard-code it to the last version that it was set to, even if that version isn’t particularly meaningful. That would preserve sort order for this part of the string, but the SDK version appears before this component anyways.)

PR #7969, opened 2017-02-07
[core, ios, macos] Refactor composite stop types and conversions

jfirebaugh:
Adjust the types to reflect the style spec reality that an exponential composite function has a single base, not a separate base per inner stops. This also unblocks #7944 by avoiding attempted construction of exponential composite stops for non-interpolatable property types.

I was able to update the iOS/macOS bindings easily, since they already reflected this setup, but it looks as if the Android bindings may have a per inner stops base. @ivovandongen can you help with updating the Android bindings?

cc @anandthakker


platform/darwin/test/MGLStyleLayerTests.mm.ejs
```diff
@@ -99,8 +99,8 @@
         functionStyleValue = [MGLStyleValue<<%- propertyType(property) %>> valueWithInterpolationMode:MGLInterpolationModeExponential compositeStops:@{@10: @{@18: constantStyleValue}} attributeName:@"keyName" options:nil];
         layer.<%- objCName(property) %> = functionStyleValue;
 
-        mbgl::style::ExponentialStops<<%- mbglType(property) %>> innerStops = { { {18, <%- mbglTestValue(property, type) %>}}, 1.0 };
-        std::map<float, mbgl::style::ExponentialStops<<%- mbglType(property) %>>> compositeStops = { {10.0, innerStops} };
+        std::map<float, <%- mbglType(property) %>> innerStops { {18, <%- mbglTestValue(property, type) %>} };
+        mbgl::style::CompositeExponentialStops<<%- mbglType(property) %>>> compositeStops { {10.0, innerStops}, 1.0 };
```
boundsj: Missing a few `>`'s and a `}`'s, I think. I think this works:

```
        std::map<float, std::map<float, <%- mbglType(property) %>>> stops = { {10, { {18, <%- mbglTestValue(property, type) %>} }} };
        mbgl::style::CompositeExponentialStops<<%- mbglType(property) %>> compositeStops { stops };
```

Issue #7552, opened 2016-12-27
Layer::getFilter() returns filter with $type compared to unsigned integer


1ec5:
A filter in style JSON that compares `$type` to a string such as `Polygon` gets parsed as an `mbgl::style::EqualsFilter` with an integer value such as 3. That seems like a reasonable in-memory optimization, but developers using the runtime styling API would expect a string representation such as `Polygon`.

For example, on iOS and macOS, the following Objective-C code:

```objc
mapView.styleURL = [MGLStyle streetsStyleURLWithVersion:10];
…
MGLFillStyleLayer *fillLayer = (MGLFillStyleLayer *)[mapView.style layerWithIdentifier:@"barrier_line-land-polygon"];
NSLog("%@", fillLayer.predicate);
```

produces this predicate:

```
$type == 3 AND class == "land"
```

whereas it should produce:

```
$type == "Polygon" AND class == "land"
```

at least until #7298 is fixed.

@jfirebaugh, should the SDK be responsible for converting these integers back into strings, or should mbgl handle that?

/cc @boundsj

jfirebaugh:
We should probably have enumerated constants for the feature types, and explicit filter types `TypeEqualsFilter`, `TypeNotEqualsFilter`, `TypeInFilter`, `TypeNotInFilter`.

1ec5:
The way the style specification is written, something like `[">=", "$type", "LineString"]` would be allowed. We might want to redefine that property as being an enumeration and specify which operators are allowed for enumerations.

jfirebaugh:
Fortunately we have a validation that prohibits that: https://github.com/mapbox/mapbox-gl-style-spec/blob/b87a7bb2b073cace5fb42cb6b515af4fbd1a11ca/lib/validate/validate_filter.js#L38-L38

ericrwolfe:
Looks like as it stands in 3.4.0, setting a layer's predicate to `"$type == 'LineString'"` etc will not work as described in the documentation. 

The only workaround right now is to use the integer values for type e.g. `"$type == 2"` (1 for Point, 2 for LineString, 3 is for Polygon).

1ec5:
[...]

PR #7971, opened 2017-02-07
[core] Introduce dedicated filter types for $type and $id special cases

jfirebaugh:
Refs #7552

### DIFF
--- a/include/mbgl/style/conversion/filter.hpp
+++ b/include/mbgl/style/conversion/filter.hpp
@@ -27,9 +27,9 @@ public:
         }
 
         if (*op == "==") {
-            return convertBinaryFilter<EqualsFilter>(value);
+            return convertEqualityFilter<EqualsFilter, TypeEqualsFilter, IdentifierEqualsFilter>(value);
         } else if (*op == "!=") {
-            return convertBinaryFilter<NotEqualsFilter>(value);
+            return convertEqualityFilter<NotEqualsFilter, TypeNotEqualsFilter, IdentifierNotEqualsFilter>(value);
         } else if (*op == ">") {
             return convertBinaryFilter<GreaterThanFilter>(value);
         } else if (*op == ">=") {
@@ -39,9 +39,9 @@ public:
         } else if (*op == "<=") {
             return convertBinaryFilter<LessThanEqualsFilter>(value);
         } else if (*op == "in") {
-            return convertSetFilter<InFilter>(value);
+            return convertSetFilter<InFilter, TypeInFilter, IdentifierInFilter>(value);
         } else if (*op == "!in") {
-            return convertSetFilter<NotInFilter>(value);
+            return convertSetFilter<NotInFilter, TypeNotInFilter, IdentifierNotInFilter>(value);
         } else if (*op == "all") {
             return convertCompoundFilter<AllFilter>(value);
         } else if (*op == "any") {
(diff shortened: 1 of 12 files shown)


1ec5:
iOS and macOS tests are failing because `FilterEvaluator` in NSPredicate+MGLAdditions.mm has no operator `()` for the new type filter types:

```
▸ Compiling NSPredicate+MGLAdditions.mm

❌  /Users/vagrant/git/mason_packages/headers/variant/1.1.4/include/mapbox/variant.hpp:312:20: no matching function for call to object of type 'FilterEvaluator'

            return f(unwrapper<T>::apply(v.template get_unchecked<T>()));
                   ^



❌  /Users/vagrant/git/mason_packages/headers/variant/1.1.4/include/mapbox/variant.hpp:312:20: no matching function for call to object of type 'FilterEvaluator'

            return f(unwrapper<T>::apply(v.template get_unchecked<T>()));
                 ^


make: *** [run-test-*] Error 65
```

For now, let’s have the conversion match [the existing documentation](https://www.mapbox.com/ios-sdk/api/3.4.1/Classes/MGLVectorStyleLayer.html#/c:objc%28cs%29MGLVectorStyleLayer%28py%29predicate) to avoid backwards compatibility concerns. For example:

```objc
[NSPredicate predicateWithFormatString:@"%@ = 'LineString'", @"$type"]
[NSPredicate predicateWithFormatString:@"%@ IN { 'LineString', 'Polygon' }", @"$type"]
```

If we’re considering cherry-picking this change into iOS SDK v3.4.2, then I’d recommend aliasing 1, 2, 3 to `Point`, `LineString`, and `Polyline`, respectively, when setting a layer’s predicate (but always producing the strings when getting a predicate). Otherwise, I think we should chalk the current behavior up to buggy behavior that isn’t worth preserving.

jfirebaugh:
Feel free to push SDK changes to this PR.

@ivovandongen, can you comment on whether the Android SDK has a similar issue to #7552 and how this change would affect it?

1ec5:
The style validator allows the `has` and `!has` operators with `$type`, right?

ivovandongen:
@jfirebaugh The `$type` filter works correctly with string values on this branch, thanks. There is no accessor yet on Android, added https://github.com/mapbox/mapbox-gl-native/issues/7978 for this.

1ec5:
@ivovandongen, so if you set `eq("$type", "Point")` on Android, will you get a `TypeEqualsFilter` under the hood? Or will mbgl be smart enough to convert it from an `EqualsFilter` to a `TypeEqualsFilter` even when specified via runtime styling?

jfirebaugh:
> The style validator allows the `has` and `!has` operators with `$id`, right?

The validator allows them but they don't work as expected: https://github.com/mapbox/mapbox-gl-js/issues/4235.

1ec5:
b7fa2c0cf54f25249c1d24eb0df59c3a265536ec takes care of the changes needed for iOS and macOS.

jfirebaugh:
Updated with `["has"/"!has", "$id"]` support.

1ec5:
8efe3b97102d753ca4c6015c4071c99ab88193b7 also takes care of `$id = nil` and `$id != nil` on iOS and macOS and adds more unit tests.

1ec5:
```
failed regressions mapbox-gl-js#4235
```

boundsj:
The iOS and macOS changes look good to me.

1ec5:
Thanks! I think all that’s left is:

* [ ] Core regression test failure: https://github.com/mapbox/mapbox-gl-native/pull/7971#issuecomment-278760879 @jfirebaugh
* [ ] Does the `eq()` syntax produce the expected results on Android? https://github.com/mapbox/mapbox-gl-native/pull/7971#issuecomment-278397228 @ivovandongen

ivovandongen:
> Does the eq() syntax produce the expected results on Android

@1ec5 Yes (https://github.com/mapbox/mapbox-gl-native/pull/7971#issuecomment-278327256)

> ... Android, will you get a TypeEqualsFilter under the hood

Yes, we use the conversion templates for filter conversion (and the rest of the style conversions) which will produce the `TypeEqualsFilter` in this case [here](https://github.com/mapbox/mapbox-gl-native/blob/master/include/mbgl/style/conversion/filter.hpp#L137).

Thanks for the heads up.

Issue #7733, opened 2017-01-16
Nightly builds are timing out / Reenable MIPS ABI


tobrun:
Last week we were unable to publish nightly builds due to #7640. Last Saturday, Bitrise has resolved this with upgrading their Android docker images. Now with re enabling these integrations, I'm noticing that some of these release builds are timing out (going over the 45 min limit). Before this issue, we were already hitting 40 <-> 43 minutes per build. We need to look into optimizing these build times for the `scheduled` workflow. 

cc @mapbox/android

tobrun:
## Measuring
In [this](https://github.com/mapbox/mapbox-gl-native/tree/tvn-limit-nighly-build-time) branch I'm measuring execution time of the different steps in our workflow: 

### Git clone: 
37 secs

### Configure GL build env: 
24 sec 

### Configure AWS CLI:
27 sec

### Download Maven credentials 
1.27 sec

### Building core: 
39.9 minutes
 - arm-v5: 483s
 - arm-v7: 371s
 - arm-v8: 374s
 - x86: 376s
 - x86-64: 384s
 - mips: 407s

### Publish to maven:
Did not fully run, was aborted  

## Optimizing
At this moment, I'm seeing 3 small optimizations:
 - do not build any android modules as part of make apackage
 - use bitrise build in support for storing secrets instead of AWS CLI (~ 2 minutes gain)
 - look into optimizing the uploadArchives gradle command (it currently also builds Phone + Wear module)

Not sure if above will make the difference, if not we will have to look into requesting more job execution from Bitrise or start splitting up the scheduled build into smaller builds.

zugaldia:
Thanks much @tobrun for looking into this. I suggest that we temporarily drop `mips` from the build as it's an architecture with [minimal](https://www.reddit.com/r/Android/comments/4ofdkv/state_of_mips_plus_some_statistics/) [impact](https://plus.google.com/+EricRichardson/posts/DPjStJmcFs4) on regular users.

At the same time we might consider to revisit the approach that you suggested in https://github.com/mapbox/mapbox-gl-native/issues/6734.

tobrun:
To unblock us from publishing new SNAPSHOTS I applied the approach in https://github.com/mapbox/mapbox-gl-native/issues/7733#issuecomment-273502978. 
This is making the [build](https://www.bitrise.io/build/32aeb927ec28c1ca) pass in 39 minutes! :tada:

zugaldia:
[...]

PR #7986, opened 2017-02-08
Removed arch from bitrise script so snapshots build

cammace:
Currently our SNAPSHOTs are timing out, removing this should fix the issue.

cc: @zugaldia


zugaldia:
Note this is a temporary measure and we're still tracking this issue on https://github.com/mapbox/mapbox-gl-native/issues/7733.

PR #7993, opened 2017-02-08
Update changelogs for DDS

boundsj:
This also adds dates to two previous releases that did not have them.

cc @1ec5 @friedbunny

### DIFF
--- a/platform/ios/CHANGELOG.md
+++ b/platform/ios/CHANGELOG.md
@@ -12,6 +12,7 @@ Mapbox welcomes participation and contributions from everyone. Please read [CONT
 
 ### Styles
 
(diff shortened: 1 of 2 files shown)


boundsj:
I also added an update for macOS. I did not see a tag for that platform's 0.3.1 release (noted in the changelog) so I did not add a date. I did add a date for 0.3.0

platform/ios/CHANGELOG.md
```diff
@@ -12,6 +12,7 @@ Mapbox welcomes participation and contributions from everyone. Please read [CONT
 
 ### Styles
 
+* A new data-driven styling API allows you to style base map features based on their properties. ([#7596](https://github.com/mapbox/mapbox-gl-native/pull/7596))
```
1ec5: This isn’t a new API, just an extension of the existing runtime styling API:

> Added support for data-driven styling in the form of source and composite style functions. `MGLStyleFunction` is now an abstract class, with `MGLCameraStyleFunction` providing the behavior of `MGLStyleFunction` in previous releases. New `MGLStyleFunction` subclasses allow you to vary a style attribute by the values of attributes of features in the source.

PR #7997, opened 2017-02-09
[qt-android] added build target for android devices

jackopoz:
includes
[QQuickMapboxGL] 
fix cache database path
added access_token MapParameter

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -552,6 +552,34 @@ run-android-$1: android-$1
 	cd platform/android && ./gradlew :MapboxGLAndroidSDKTestApp:install$(BUILDTYPE) && adb shell am start -n com.mapbox.mapboxsdk.testapp/.activity.FeatureOverviewActivity
 
 apackage: android-lib-$1
+
+#### Qt Android Targets
+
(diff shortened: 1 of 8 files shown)


tmpsantos:
Hi @jackopoz, thanks for this PR! We are going to support Qt + Android via the official Qt SDK from version 5.9. Good thing there is if you are already using the QML Map API you just need to change a string. All the Map items, route models, etc will just work.

The plan is to deprecated this QML item in favor of what we are contributing to Qt:
https://github.com/mapbox/mapbox-gl-native/projects/7#card-1241792

You can build and run it already, 5.9 was just branched and the API is stable. We are only adding new map parameters.


Mapbox GL Native was merged into Qt here:
https://codereview.qt-project.org/#/c/183479/

And documentation for now is here:
https://doc-snapshots.qt.io/qt5-5.9/location-plugin-mapboxgl.html

Let us know if you need help building it.

tmpsantos:
@jackopoz we also would like to invite you to contribute to the plugin on the Qt 5.9 tree. We are on #qt-mobility @ Freenode.

PR #7999, opened 2017-02-09
French, German, Japanese, Portuguese, Swedish, Vietnamese

1ec5:
Pulled the latest translations from Transifex and added new languages to the iOS and macOS projects. Added .stringsdict files according to issues posted to Transifex by translators. The `tx` configuration included languages with at least one file translated 80% or more. Brazilian Portuguese, French, and Japanese are partial translations at this time.

/cc @friedbunny @frederoni @kkaefer @brunoabinader


frederoni:
Swedish is now at 💯

PR #8002, opened 2017-02-09
Update changelogs for iOS SDK v3.5.0-beta.1

1ec5:
Added mention of #7786, #7989, #7902. Moved #7956 to the correct section.


boundsj:
🙇

PR #8004, opened 2017-02-09
macOS: reuse OpenGL context when switching screens

### LINKED ISSUE #6654: macOS SDK crashes when moving between screens
When moving the macOS SDK window between screens (one with a Retina pixel ratio, and the other with a regular pixel ratio in my case), I'm getting a crash when trying to update the `FrameHistory` texture. It doesn't appear to be specific to that texture since it crashes at other locations as well if that upload is commented out. It seems like doing any sort of OpenGL API call triggers this crash.

Using apitrace, I found that just prior to the crash, `CGLUpdateContext` is called, and returns `kCGLBadDrawable`.

The crash does **not** appear when moving GLFW windows between screens, making me believe that there's something fishy in our use of `NSOpenGLLayer` or the macOS implementation in general.

/cc @1ec5 @incanus

kkaefer:
Fixes #6654

### DIFF
--- a/platform/macos/src/MGLOpenGLLayer.mm
+++ b/platform/macos/src/MGLOpenGLLayer.mm
@@ -2,7 +2,9 @@
 
 #import "MGLMapView_Private.h"
 
-@implementation MGLOpenGLLayer
+@implementation MGLOpenGLLayer {
+    NSOpenGLContext *_context;
+}
 
 - (MGLMapView *)mapView {
     return (MGLMapView *)super.view;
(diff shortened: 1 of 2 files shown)


platform/macos/src/MGLOpenGLLayer.mm
```diff
@@ -4,6 +4,8 @@
 
 @implementation MGLOpenGLLayer
 
+NSOpenGLContext* context = nil;
```
1ec5: This is a global. Did you mean to make it an ivar?

kkaefer: Yes. How do I do that?

1ec5:
```objc
@implementation MGLOpenGLLayer {
    NSOpenGLContext *_context; // automatically initialized to nil
}
```

PR #8008, opened 2017-02-09
[core] Restore support for *-transition properties

### LINKED ISSUE #7863: Per-property transitions are not parsed from style
16c435b1517b15a5ea8654987979ef58800b838b inadvertently removed the call to `parseTransitionOptions()` and thus support for per-property transitions in style JSON. This would represent a regression from iOS SDK v3.3.x and Android SDK v4.1.0.

Per mapbox/mapbox-gl-js#2788, mapbox/mapbox-gl-style-spec#171, and mapbox/mapbox-gl-test-suite#116, the current plan is to give per-property transitions all the trappings of a formal style specification feature, so here’s one more: the developer might expect to be able to configure per-property transitions using the runtime styling API, now that they can configure style-wide transitions (#7711).

/cc @jfirebaugh @incanus

jfirebaugh:
Fixes #7863.


1ec5:
Tracking iOS/macOS runtime styling support in #8010. @tobrun, can you ticket out the corresponding work on the Android side?

PR #8009, opened 2017-02-09
[ios] Add partial Japanese localization

friedbunny:
Adds an 84% complete Japanese translation of our iOS strings.

(Xcode also automatically removed a duplicate[?] entry for `MGLDistanceFormatter.h` in the project.)

/cc @boundsj

### DIFF
--- a/platform/ios/resources/ja.lproj/Localizable.strings
+++ b/platform/ios/resources/ja.lproj/Localizable.strings
@@ -1,0 +1,75 @@
+﻿/* Accessibility hint */
+"ANNOTATION_A11Y_HINT" = "詳細を伝える";
+
+/* No comment provided by engineer. */
+"API_CLIENT_400_DESC" = "The session data task failed. Original request was: %@";
+
+/* No comment provided by engineer. */
+"API_CLIENT_400_REASON" = "The status code was %ld";
+
+/* No comment provided by engineer. */
+"CANCEL" = "キャンセル";
+
+/* Accessibility hint */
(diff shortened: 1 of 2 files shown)


friedbunny:
After discussion with @1ec5, we’re not sure why Xcode tried to remove MGLDistanceFormatter.h’s membership from the static target (https://github.com/mapbox/mapbox-gl-native/pull/8001). I’ve tried to re-add it in 076dcee, but I can’t be sure it’ll take.

platform/ios/resources/ja.lproj/Localizable.strings
```diff
@@ -0,0 +1,75 @@
...
+"INFO_A11Y_LABEL" = "この地図について";
+
+/* Accessibility label */
+"LOGO_A11Y_LABEL" = "Mapbox";
+
+/* Accessibility label */
+"MAP_A11Y_LABEL" = "地図";
+
+/* Map accessibility value */
+"MAP_A11Y_VALUE" = "ズーム %1$dx\n%2$ld ピン現れる";
```
1ec5: Are you sure you want “x” to be read aloud after the zoom level number? In English, it would sound like “ten ex”. Not sure about Japanese.

friedbunny: Good point, I didn’t realize this wasn’t just some format string named `$dx`. ;) The suffix 倍 is what gets used for camera zoom — it means “times” or “-fold”.

friedbunny: Updated in Transifex and amended in 2caaa46.

PR #8012, opened 2017-02-10
[ios] Added an optional tile list parameter to MGLTilePyramidOfflineRegion

wibge:
This allows downloading non rectangular map packs, by specifying a specific list of tiles to download. This is useful for downloading along a track or within a polygon area.


wibge:
I agree too. I had tried to it originally with proper subclasses, and got a little hung up with the variants. I'll give it a try again now that I'm more familiar with the code.

asheemmamoowala:
Addressed by #11447.

PR #8022, opened 2017-02-10
Don't deselect annotations on pan/zoom, even if the annotation moves out of the visible bounds

### LINKED ISSUE #8021: Callout-less annotation is deselected when user pans or zooms the map
**Platform:** iOS
**Mapbox SDK version:** 3.4.1

I thought #4389 would be fixed in 3.4.0, but unfortunately the problem still exists. `- (void)mapView:(MGLMapView *)mapView didDeselectAnnotation:(id<MGLAnnotation>)annotation` is still called when the user pans or zooms the map. We don't use custom callout views but another info view that is placed below the map. This view is shown when an annotation is selected and hidden, when an annotation is deselected. Furtermore, we display tracks on the map when certain annotations are selected. If the user now zooms in or out to see the complete track, the annotation is deselected again and our logic removes the track from the map.

As a solution, I'd like to add a property called `deselectOnRegionChange` to `MGLAnnotationView` which defaults to true so that the current behavior is unchanged. If this property is set to false, the annotation is not deselected any longer when the user pans or zooms the map.

### Steps to trigger behavior

 1. Add annotation to map (also provide an annotation view)
 2. Keep track of the annotations selection state (possible by implement `setSelected:animated` in the annotation view)
 3. Select annotation
 4. Change region by panning the map a little bit

### Expected behavior

Annotation should still be selected

### Actual behavior

Annotation is immediately deselected

Ben305:
This fixes #8021 

Don't deselect annotations on pan/zoom, even if the annotation moves out of the visible bounds. Now MapBox behaves like Google Maps and MapKit.

### DIFF
--- a/platform/ios/src/MGLMapView.mm
+++ b/platform/ios/src/MGLMapView.mm
@@ -3259,6 +3259,11 @@ public:
 
     if (annotationView)
     {
+        // Make sure that the annotation views are selected/deselected correctly because
+        // annotations are not dismissed when they move out of the visible bounds
+        BOOL isViewForSelectedAnnotation = self.selectedAnnotation == annotation;
+        [annotationView setSelected:isViewForSelectedAnnotation];
+
         annotationView.annotation = annotation;
         annotationView.mapView = self;
         CGRect bounds = UIEdgeInsetsInsetRect({ CGPointZero, annotationView.frame.size }, annotationView.alignmentRectInsets);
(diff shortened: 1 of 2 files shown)


1ec5:
Thanks for this PR. Based on https://github.com/mapbox/mapbox-gl-native/issues/8021#issuecomment-279073188, I think we could get away with a fix that doesn’t add any additional options to the API, but I’m curious whether that would address your use case.

platform/ios/src/MGLMapView.mm
```diff
@@ -4729,7 +4736,7 @@ - (void)notifyMapChange:(mbgl::MapChange)change
                             rect = annotationView.frame;
                         }
 
-                        if ( ! CGRectIntersectsRect(rect, self.frame)) {
+                        if ( ! CGRectIntersectsRect(rect, self.frame) && self.deselectWhenOutOfBounds) {
```
frederoni: I'm not convinced this property is needed. I would prefer to just remove these lines unless I can see a use case where deselect when out of bounds should be the default behavior.

Ben305: I'm happy to remove this, I just added it because I thought that @1ec5 was not fine with removing the current way how deselection is handled completely.

1ec5: Given https://github.com/mapbox/mapbox-gl-native/issues/8021#issuecomment-279137067, I’ve changed my mind; if MapKit and Google Maps avoid deselecting the annotation, then that’s what developers expect, and an extra option is unnecessary.

Are there any implications for callout views? Does any of the callout code assume that a callout view (if it exists) is always visible as long as its annotation is selected?

Ben305: Cool, I have updated my PR to that annotations are not deselected any longer. If the annotation has a callout view which responds to `dismissAutomatically`, the annotation is deselected if that function returns `true`.

Ben305:
Happy to do that. Done ;-)

PR #8025, opened 2017-02-11
Convert NSColor/UIColor expressions to CSS color string values

1ec5:
When converting a constant NSExpression to an `mbgl::Value`, turn an NSColor or UIColor into an `std::string` containing a CSS color string. This allows developers to set an attribute of an MGLFeature to an NSColor or UIColor (rather than a CSS color string, which would be foreign), then use it in an MGLStyleFunction with identity interpolation.

Along the way, I added language to `MGLFeature.attributes`’ documentation to talk about the use of feature attributes in identity-interpolated style functions. I also simplified the NSExpression unit tests to more specifically test the category methods that convert constant NSExpressions to `mbgl::Value`s, without going through NSPredicate. Normally I’d be hesitant to rely so heavily on macros, which makes debugging more difficult, but the existing tests rely on `for` loops, which make test failures just as difficult to diagnose.

/cc @boundsj @jfirebaugh

### DIFF
--- a/platform/darwin/src/MGLFeature.h
+++ b/platform/darwin/src/MGLFeature.h
@@ -83,12 +83,21 @@ NS_ASSUME_NONNULL_BEGIN
  `MGLVectorStyleLayer.predicate` to an `NSPredicate` with the format
  `importance > 50`.
 
- You can also configure some attributes of an `MGLSymbolStyleLayer` object to
- include the value of an attribute in this dictionary whenever it renders this
- feature. For example, to label features in an `MGLShapeSource` object by their
- names, you can assign a `name` attribute to each of the source’s features, then
- set `MGLSymbolStyleLayer.textField` to an `MGLStyleValue` object containing the
- string `{name}`.
+ You can also configure many layout and paint attributes of an `MGLStyleLayer`
+ object to match the value of an attribute in this dictionary whenever it
+ renders this feature. For example, if you display features in an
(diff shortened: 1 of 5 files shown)


platform/darwin/src/MGLFeature.h
```diff
@@ -83,12 +83,21 @@ NS_ASSUME_NONNULL_BEGIN
...
- set `MGLSymbolStyleLayer.textField` to an `MGLStyleValue` object containing the
- string `{name}`.
+ You can also configure many layout and paint attributes of an `MGLStyleLayer`
+ object to match the value of an attribute in this dictionary whenever it
+ renders this feature. For example, if you display features in an
+ `MGLShapeSource` using an `MGLCircleStyleLayer`, you can assign a `halfway`
+ attribute to each of the source’s features, then set
+ `MGLCircleStyleLayer.circleRadius` to an `MGLStyleValue` object with an
+ interpolation mode of `MGLInterpolationModeIdentity` and an attribute name of
+ `halfway`.
```
anandthakker: Ack, this didn't occur to me earlier, but would this be a good place to suggest referring to the DDS support tables to for those properties where ^ this approach is actually available?

1ec5: `MGLCircleStyleLayer.circleRadius` will link automatically to the documentation for that property, which like all layout and paint properties will list the interpolation modes and function types that are supported. I would like to refer the developer to a guide on data-driven styling, once it’s written as part of #7924.

PR #8026, opened 2017-02-11
Refactor: use mbgl::style::conversion system to convert MGLStyleFunction to mbgl core types

anandthakker:
Quoting from https://github.com/mapbox/mapbox-gl-native/pull/7944#issuecomment-278809848 (see the thread there for a bit more context):

> Followup on https://github.com/mapbox/mapbox-gl-native/pull/7944#issuecomment-278390927: As I started digging into resolving the compilation issues in the MGLCompositeStyleFunction => mbgl::style::CompositeFunction conversion today, it felt like I was solving the same problems that are being solved by `mbgl::style::conversion::*`.

> I'm now thinking that we should accomplish the MGL => mbgl conversion — at least for property values — by having the darwin platform bindings implement [the Value conversions required](https://github.com/mapbox/mapbox-gl-native/blob/master/include/mbgl/style/conversion.hpp#L30-L52) to leverage the existing `mbgl::style::conversion` system.  This will hopefully mean less code/logic duplication and better maintainability.

---

The approach I'm going for here is for `MGLStyleValueTransformer::toDataDrivenPropertyValue()` to:
1. Use the new `NSObject * MGLStyleValueTransformer::toRawStyleSpecValue(MGLStyleFunction<T> *value)` method to convert an `MGLStyleFunction` to an plain NSObject-based hierarchy isomorphic to the style spec JSON for the given style function value.
2. Stuff that into a `MGLConversionValue`, whose job is simply to be a convenience wrapper for the `mbgl::style::conversion` overloads provided in `darwin_conversion.hpp`.
3. Having implemented these base overloads, we can now just call `convert<mbgl::style::DataDrivenPropertyValue<MBGLType>>(value)` and it "just works".  (More on the conversion system [here](https://github.com/mapbox/mapbox-gl-native/blob/master/include/mbgl/style/conversion.hpp#L30-L52))

cc @1ec5 @boundsj  - (note this is still WIP; happy to have early course corrections if you want to peek, but feel free to just ignore for now and I'll ping when it's ready)


anandthakker:
Oh, another note: my intent was to keep this PR/refactor scoped to handling StyleFunction types (because those are the ones presenting difficulty due to the stop type variants based on interpolatability).  But once it's working, propagating this approach to the other MGL => mbgl conversions should be pretty much as simple as implementing `toRawStyleSpecValue` for each MGL type.

platform/darwin/src/darwin_conversion.hpp
```diff
@@ -0,0 +1,93 @@
...
+        return member;
+    } else {
+        return {};
+    }
+}
+
+template <class Fn>
+optional<Error> eachMember(const MGLConversionValue&, Fn&&) {
+    // TODO
+    // mbgl::Log::Warning(mbgl::Event::Android, "eachMember not implemented");
```
anandthakker: Here, and in some other places, need to raise/log an error in whatever way is "right" for this SDK/project.

platform/darwin/src/MGLConversionValue.h
```diff
@@ -0,0 +1,73 @@
...
+    bool toBool() const {
+        return (bool) ((NSNumber *)value).boolValue;
+    };
+    
+    Value get(const char* key) const {
+        NSDictionary *dict = (NSDictionary *)value;
+        return Value(value[key]);
+    };
+    
+    int getLength() const {
```
1ec5: `arrayLength()` returns a `size_t`, while `-[NSArray count]` returns an NSUInteger (which is `unsigned int` on 32-bit systems and `unsigned long` on 64-bit systems). Either way, specifying `int` here risks overflow.

platform/darwin/src/MGLStyleValue_Private.h
```diff
@@ -115,162 +117,14 @@ class MGLStyleValueTransformer {
...
-             @"Make sure the style value was created as a member of a concrete subclass of MGLStyleValue.",
-             NSStringFromClass([value class])];
-            return {};
-        } else {
-            return {};
+        // TODO: check that it's an MGLStyleFunction before casting.
+        NSObject *rawValue = toRawStyleSpecValue((MGLStyleFunction<ObjCType> *) value);
+        mbgl::style::conversion::Result<mbgl::style::DataDrivenPropertyValue<MBGLType>> result = mbgl::style::conversion::convert<mbgl::style::DataDrivenPropertyValue<MBGLType>>(rawValue);
+        if (!result) {
+            // raise an exception?
```
1ec5: There are three ways to properly raise an exception in Objective-C:

* Use the `NSAssert()` macro to indicate some error we’ve made in our code.
* Use the `NSCAssert()` macro for the same thing, but inside a C or C++ function where `self` is unavailable.
* Call the `+[NSException raise:format:]` method to indicate an error that’s due to programming error on the part of the developer or a situation that’s out of our control. The first argument is usually a constant like `NSInvalidArgumentException`.

Additionally, if you just want to log a message to the console, the `NSLog()` macro works throughout Objective-C code.

platform/darwin/src/MGLStyleValue_Private.h
```diff
@@ -322,6 +176,75 @@ class MGLStyleValueTransformer {
...
+        // defaultValue => default
+        if (styleFunction.defaultValue) {
+            NSCAssert([styleFunction.defaultValue isKindOfClass:[MGLStyleConstantValue class]], @"Default value must be constant");
+            rawFunction[@"default"] = [(MGLStyleConstantValue<ObjCType> *)styleFunction.defaultValue rawValue];
+        }
+        
+        // interpolationBase => base
+        if (styleFunction.interpolationBase) {
+            // TODO: does this direct cast work?
+            rawFunction[@"base"] = (NSNumber *) styleFunction.interpolationBase;
```
1ec5: This cast merely treats a `CGFloat` (`float` or `double`, depending on the architecture) as a pointer to an NSNumber object. To convert a primitive number to an NSNumber object, `@(box)` it or call a method like `+[NSNumber numberWithDouble:]`.

platform/darwin/src/MGLStyleValue_Private.h
```diff
@@ -322,6 +176,75 @@ class MGLStyleValueTransformer {
...
+            [compositeStyleFunction.stops enumerateKeysAndObjectsUsingBlock:^(NSNumber * _Nonnull zoomKey, NSDictionary * _Nonnull stopValue, BOOL * _Nonnull stop) {
+                for (NSObject *valueKey in stopValue.allKeys) {
+                    NSDictionary *stopKey = @{
+                         @"zoom": zoomKey,
+                         @"value": valueKey
+                     };
+                    MGLStyleConstantValue<ObjCType> *outputValue = stopValue[valueKey];
+                    NSCAssert([outputValue isKindOfClass:[MGLStyleConstantValue<ObjCType> class]], @"Stop outputs should be MGLStyleConstantValues");
+                    NSArray *rawStop = [NSArray arrayWithObjects:stopKey, [outputValue rawValue]];
+                    [stops addObject:rawStop]; // TODO weird: why doesn't stops.addObject(rawStop) work?
```
1ec5: In Objective-C, the [syntax for calling a method](https://developer.apple.com/library/content/documentation/Cocoa/Conceptual/ProgrammingWithObjectiveC/WorkingwithObjects/WorkingwithObjects.html) is `[object sendMessage:message withArgument:argument options:options]`. The syntactic sugar for an instance method taking no parameters is `dot.syntax`, which is meant for getting a `@property` (since the getter of a property is an instance method taking no parameters).

Incidentally, since a property’s getter is a method like any other, you should be aware that getting a property may sometimes be an expensive operation with side effects. [Some developers](https://www.bignerdranch.com/blog/dot-notation-syntax/) advise against using dot notation to emphasize this point. In fact, there’s no downside to using bracket notation for methods and properties alike, other than forcing your poor code reviewer to read Lisp.

platform/darwin/src/darwin_conversion.hpp
```diff
@@ -0,0 +1,93 @@
...
+inline bool isArray(const MGLConversionValue& value) {
+    return value.isArray();
+}
+
+inline bool isObject(const MGLConversionValue& value) {
+    return value.isObject();
+}
+
+inline std::size_t arrayLength(const MGLConversionValue& value) {
+    return value.getLength();;
```
1ec5: Nit: stray semicolon.

platform/ios/ios.xcodeproj/project.pbxproj
```diff
@@ -518,6 +518,8 @@
 /* End PBXCopyFilesBuildPhase section */
 
 /* Begin PBXFileReference section */
+		17C4564F1E4E19BB0029C9B1 /* darwin_conversion.hpp */ = {isa = PBXFileReference; fileEncoding = 4; lastKnownFileType = sourcecode.cpp.h; path = darwin_conversion.hpp; sourceTree = "<group>"; };
```
1ec5: Make sure this file is a member of both the dynamic and static targets.

1ec5: Remember to add this file to macos.xcodeproj as well, since all the runtime styling code is shared between both platforms.

platform/darwin/src/darwin_conversion.hpp
```diff
@@ -0,0 +1,93 @@
...
+}
+
+inline mbgl::android::Value arrayMember(const MGLConversionValue& value, std::size_t i) {
```
1ec5: Did you mean `mbgl::Value`?

platform/darwin/src/MGLConversionValue.h
```diff
@@ -0,0 +1,73 @@
...
+    float toNumber() const {
+        return (float) ((NSNumber *)value).floatValue;
+    };
+    
+    bool toBool() const {
+        return (bool) ((NSNumber *)value).boolValue;
+    };
+    
+    Value get(const char* key) const {
+        NSDictionary *dict = (NSDictionary *)value;
```
1ec5: Is it guaranteed that this method will only be called if `isObject()` returns true? Even if so, it might be worth asserting that we’re dealing with an NSDictionary.

platform/darwin/src/darwin_conversion.hpp
```diff
@@ -0,0 +1,93 @@
...
+inline optional<Value> toValue(const MGLConversionValue& value) {
+    if (value.isNull()) {
+        return {};
+    } else if (value.isBool()) {
+        return { value.toBool() };
+    } else if (value.isString()) {
+        return { value.toString() };
+    } else if (value.isNumber()) {
+        // Need to cast to a double here as the float is otherwise considered a bool...
+       return { (double) value.toNumber() };
```
1ec5: In C++ and Objective-C++, the `static_cast` syntax allows the compiler to warn you when you try to cast to an unrelated type. A C-style cast, by contrast, allows you to cast to any type with abandon.

1ec5: Could `MGLConversionValue::toNumber()` return a `double` in the first place, avoiding the need for this cast? Converting `float` to `double` often leaves [unwanted](https://github.com/mapbox/mapbox-gl-native/blob/e6c15d0f6f285bc604c0d93381b9b1d3957cb5c9/platform/darwin/src/NSExpression%2BMGLAdditions.mm#L43-L46) artifacts at the smallest decimal places.

platform/darwin/src/MGLStyleValue_Private.h
1ec5: In Objective-C, it’s extremely rare (read: kludge) to `@try`-`@catch` an NSException. (`NSAssert()` and `NSCAssert()` both raise NSExceptions under the hood.) In fact, on iOS, an uncaught exception will crash the application. On macOS, an uncaught exception will spew to the console and skip to the next iteration of the run loop, which may either be graceful or lead to maddeningly undebuggable downstream misbehavior.

If you don’t want the application to crash or misbehave, warn with `NSLog()` or whatever C++ warning facility you have at your disposal.

platform/darwin/src/MGLConversionValue.h
```diff
@@ -0,0 +1,73 @@
...
+
+#ifndef MGLConversionValue_h
+#define MGLConversionValue_h
+
+class MGLConversionValue {
+public:
+    Value(NSObject * _value) : value(_value) {}
+    
+    bool isNull() const {
+        return value;
```
anandthakker: 🙈

platform/darwin/src/MGLConversionValue.h
```diff
@@ -0,0 +1,73 @@
...
+    
+    bool isNumber() const {
+        return [value isKindOfClass:[NSNumber class]] && !isBool();
+    }
+    
+    std::string toString() const {
+        return std::string([(NSString *)value UTF8String]);
+    };
+    
+    float toNumber() const {
```
anandthakker: @1ec5 hmm, yeah -- so the reason I went with this here is that (I believe) the mbgl conversion system expects `mbgl::style::conversion::toNumber` to return `optional<float>`. I suppose we could have `MGLConversionValue::toNumber()` return an NSNumber, and leave it to `conversion::toNumber()` to retrieve the floatValue.  (Though this makes me second-guess whether the `MGLConversionValue` wrapper is even necessary -- couldn't we just overload `conversion::toNumber()` for `NSObject` directly?)

By the way, is replacing `mgl_constantMBGLValue` something I should consider as a requirement/use case for MGLConversionValue?   If so, I may need a little help understanding the function/purpose of `mgl_constantMBGLValue` and friends.

platform/darwin/src/MGLConversionValue.h
anandthakker: It's guaranteed in the spec, but agreed on asserting anyway.

platform/darwin/src/MGLConversionValue.h
```diff
@@ -0,0 +1,73 @@
...
+        return [size count];
+    };
+    
+    Value get(const int index ) const {
+        NSArray * array = (NSArray *)value;
+        return array[index];
+    };
+    
+private:
+    NSObject * value;
```
anandthakker: It seems like we _do_ want to constrain the object's class here, though, since we're making assumptions about `value` being an instance of the NSObject class hierarchy -- no?

platform/darwin/src/MGLStyleValue_Private.h
anandthakker: Huh -- does this mean that there's no reliable way to fail hard-but-descriptively?  Or are you saying that whenever we _need_ to raise an exception, we should also `NSLog()` so that it isn't maddening to debug?

platform/darwin/src/MGLStyleValue_Private.h
anandthakker: Ahh, I see.  I missed the fact that the dot notation was specifically for property access!

> forcing your poor code reviewer to read Lisp.

😂

platform/darwin/src/darwin_conversion.hpp
anandthakker: Heh, actually it should have been MGLConversionValue.

anandthakker:
@1ec5 Thank you for this extremely helpful review-slash-Objective-C-lesson!

platform/darwin/src/MGLConversionValue.h
```diff
@@ -0,0 +1,68 @@
+#include <string>
+
+class MGLConversionValue {
```
jfirebaugh: What's the purpose of `MGLConversionValue` in this PR, as opposed to inlining each of its methods in the corresponding method in darwin_conversion.hpp? E.g.:

```objc
inline bool isUndefined(NSObject *value) {
    return !value || value == [NSNull null];
}
```

anandthakker: @jfirebaugh hah -- yeah, this is exactly the question I started asking myself this morning: https://github.com/mapbox/mapbox-gl-native/pull/8026#discussion_r100799109.  I started out with this wrapper because I didn't know at the time that a fully/cleanly `NSObject`-based hierarchy would be possible.  If you agree that the wrapper's unnecessary, I'd love to remove it.

platform/ios/ios.xcodeproj/project.pbxproj
anandthakker: I believe this should now be done.

anandthakker: Done.

anandthakker:
@1ec5 @jfirebaugh okay, I think this should now be ready for a (re-)review.  Two outstanding issues:

1. One failing unit test--`MGLStyleValueTests::testFunctionsWithDataDrivenProperties`--due to precision loss in color values -- this happens because, in order to use the conversion subsystem, we end up roundtripping the color value through an `"rgb(...)"` string -- so `[..., 0.5, ...]` becomes `"rgb(..., 127.5, ...)"` becomes `[..., 0.498039,...]`.
2. Per https://github.com/mapbox/mapbox-gl-native/pull/8026#pullrequestreview-21591841, remove MGLConversionValue, as it is an unnecessary level of indirection.  (I just tried this locally and am getting some mysterious error `'__val__' is unavailable in ARC`.   Storing that on this [temporary branch](https://github.com/mapbox/mapbox-gl-native/tree/drop-mgl-conversion-value) till tomorrow AM when I'll take another look.)

jfirebaugh:
`"rgb(...)"` strings don't limit the precision of their components, do they? Seems like we should be able to fix the precision loss.

platform/darwin/src/MGLConversionValue.h
1ec5: > couldn't we just overload the conversion functions for NSObject directly?

That would be desirable; however, bool and float are both represented in object form as NSNumber objects.

1ec5: > By the way, is replacing mgl_constantMBGLValue something I should consider as a requirement/use case for MGLConversionValue?

Not necessarily, although we should make sure they're somewhat consistent. mgl_constantMBGLValue is primarily intended as a way to convert constant NSExpressions representing vector tile feature properties into the equivalent mbgl types. This is why it makes a distinction between integers and floats, for instance, and doesn't handle enumeration values. This method is used when converting an NSPredicate (which is made up of NSExpressions) into an mbgl::Filter (which generally contains mbgl::Values).

jfirebaugh:
Oh, I see, the issue isn't the stringification, it's that the conversion back goes through `CSSColorParser::Color` which uses `unsigned char` components. We don't guarantee more precision than that in other parts of the rendering, so I think this is something that the test will have to adjust for.

platform/darwin/src/MGLConversionValue.h
1ec5: For all practical purposes, NSObject and NSProxy are the only two root classes in Objective-C. If you don't care about the difference between NSObject and NSProxy – you may never encounter any reason to care about this distinction – then `id` is tantamount to `NSObject *` except with the convenience of not having to cast to a more specific type.

platform/darwin/src/MGLStyleValue_Private.h
1ec5: Neither. Use NSAssert() or its variants, NSCAssert() or its variants, or NSException directly when you want to fail hard with a descriptive message in the crash log.

NSLog() is just a glorified printf() that happens to know how to print an Objective-C object descriptively; you can use it to warn on stdout.

My point about try-catch is that you shouldn't rely on exceptions for flow control or recoverable errors (not that that's what you're trying to do here). For recoverable errors, an Objective-C method should accept an out parameter of type `NSError **`.

platform/darwin/src/MGLConversionValue.h
1ec5: Good point – we aren't actually using method overloading here, so the fact that both bool and float are encapsulated in NSNumber just means the implementation of isBool() and isNumber() would involve an ||.

1ec5:
> some mysterious error '__val__' is unavailable in ARC

ARC is unable to reason about references like `NSObject *&`. With an Objective-C object, you should only ever need to use `*` or more rarely `**`, never `&` or `*&`.

platform/darwin/src/MGLConversionValue.h
anandthakker: Ah, I see 👍

platform/darwin/src/MGLStyleValue_Private.h
anandthakker: > My point about try-catch is that you shouldn't rely on exceptions for flow control or recoverable errors

Gotcha--thanks for flagging the more general point!

anandthakker:
> ARC is unable to reason about references like NSObject *&. With an Objective-C object, you should only ever need to use * or more rarely **, never & or *&.

Ah.  Yeah, the *& did seem weird and unnecessary even when I was typing it.  But, alas, changing the types of the function parameters to the more sensical `NSObject *` doesn't solve the problem -- I still get the ARC error, and I _think_ it's because we need an `optional<NSObject *>` -- the error log references a line from std::experimental::optional that looks like:

```
    _LIBCPP_INLINE_VISIBILITY constexpr optional(const value_type& __v)
```

So it seems like trying to make an  `optional<NSObject *>`, we're still stuck with `NSObject *&` under the hood.

@jfirebaugh @1ec5 does this mean we basically need to stick with having a wrapper class?  Or is there another way out?

anandthakker:
Added an accuracy threshold to the color equality checks per https://github.com/mapbox/mapbox-gl-native/pull/8026#issuecomment-279571804.  @1ec5 as those were my first lines of Swift, I would imagine that I've violated any number of conventions/idioms, heh.

1ec5:
> So it seems like trying to make an  `optional<NSObject *>`, we're still stuck with `NSObject *&` under the hood.

It’s already a pointer, so you can indicate the absence of a value with `nil`.

platform/darwin/test/MGLStyleValueTests.swift
```diff
@@ -1,8 +1,53 @@
 import XCTest
 import Mapbox
 
-
+#if os(iOS) || os(watchOS) || os(tvOS)
+typealias ColorType = UIColor
```
1ec5: In Objective-C code, we’ve been using `MGLColor` as the name of the platform-agnostic color type.

platform/darwin/test/MGLStyleValueTests.swift
```diff
@@ -1,8 +1,53 @@
...
+#elseif os(macOS)
+typealias ColorType = NSColor
+#endif
+    
 extension MGLStyleValueTests {
+    func assertColorsEqualWithAccuracy(_ actual: ColorType, _ expected: ColorType, accuracy: Float = 0.1) {
```
1ec5: Just an observation, not a request for change: Swift lets you overload methods like in C++. So you could overload `XCTAssertEqualWithAccuracy(_:_:accuracy:)` with two `MGLColor` arguments. Copying the existing `XCTAssertEqualWithAccuracy(_:_:accuracy:)` signature, you’d wind up with:

```swift
func XCTAssertEqualWithAccuracy<T : FloatingPoint>(_ color1: MGLColor, _ color2: MGLColor, accuracy: T, _ message: @autoclosure () -> String = "", file: StaticString = #file, line: UInt = #line)
```

With this fancy signature, you could pass `file` and `line` into the calls to the normal `XCTAssertEqualWithAccuracy(_:_:accuracy:)` and get a failure message that refers to the original line number.

anandthakker:
> It’s already a pointer, so you can indicate the absence of a value with nil.

In general, yes, but the problem is that the contract for the mbgl conversion system expects  `optional<>` return values for some of the conversion functions: https://github.com/mapbox/mapbox-gl-native/blob/master/include/mbgl/style/conversion.hpp#L47-L52

platform/darwin/test/MGLStyleValueTests.swift
1ec5: With this typealias in place, it might be possible to unconditionalize the giant `#if` block in `testFunctionsWithDataDrivenProperties()`.

platform/darwin/test/MGLStyleValueTests.swift
1ec5: Is 0.1 all the accuracy we can guarantee?

platform/darwin/test/MGLStyleValueTests.swift
```diff
@@ -1,8 +1,53 @@
...
+        for (ac, ec) in zip(actualComponents, expectedComponents) {
+            XCTAssertEqualWithAccuracy(Float(ac), Float(ec), accuracy: accuracy)
+        }
+    }
+    
+    func assertColorValuesEqual(_ actual: MGLStyleValue<ColorType>, _ expected: MGLStyleValue<ColorType>) {
+        if let actualConstant = actual as? MGLStyleConstantValue<ColorType> {
+            XCTAssertTrue(expected is MGLStyleConstantValue<ColorType>)
+            assertColorsEqualWithAccuracy(actualConstant.rawValue, (expected as! MGLStyleConstantValue<ColorType>).rawValue)
+            return
```
1ec5: It looks like no code would get executed beyond this `if` statement, so no `return` statement is needed.

platform/darwin/test/MGLStyleValueTests.swift
```diff
@@ -1,8 +1,53 @@
...
+        } else if let actualFunction = actual as? MGLStyleFunction<ColorType> {
+            XCTAssertTrue(expected is MGLStyleFunction<ColorType>)
+            let expectedFunction = expected as! MGLStyleFunction<ColorType>
+            XCTAssertEqual(actualFunction.interpolationBase, expectedFunction.interpolationBase)
+            XCTAssertEqual(actualFunction.interpolationMode, expectedFunction.interpolationMode)
+            // TODO: assert default values are equal
+            if actualFunction.stops == nil {
+                XCTAssertNil(expectedFunction.stops)
+                return
```
1ec5: Use a `guard … else` statement for an early return. Better yet, use a `guard let … else` statement for an early return conditioned on the existence of a variable you want to make use of below.

platform/darwin/test/MGLStyleValueTests.swift
anandthakker: Oh, whoops, I meant for that to be 0.01, although I guess 0.005 is maybe a better, as a roundish number close to 1/255.

platform/darwin/test/MGLStyleValueTests.swift
anandthakker: Ooh, `guard` is so nice!

platform/darwin/test/MGLStyleValueTests.swift
1ec5: `= 1/255` also works.

platform/darwin/src/MGLConversionValue.h
```diff
@@ -0,0 +1,68 @@
...
+        return { dict[@(key)] };
+    };
+    
+    size_t getLength() const {
+        NSCAssert([value isKindOfClass:[NSArray class]], @"Value must be an NSArray for getLength().");
+        NSArray * array = (NSArray *)value;
+        return [array count];
+    };
+    
+    MGLConversionValue get(const std::size_t index ) const {
```
1ec5: `size_t` isn’t explicitly defined to be the same size as `NSUInteger` (the type used to subscript NSArray below), even though I think that happens to be the case. To avoid even the potential of an out-of-bounds array access, you could assert that `index` is at most `NSUIntegerMax`.

platform/darwin/src/MGLStyleValue_Private.h
```diff
@@ -5,6 +5,10 @@
 #import "NSValue+MGLStyleAttributeAdditions.h"
 #import "MGLTypes.h"
 
+#include "darwin_conversion.hpp"
```
1ec5: The convention for our predominantly Objective-C targets is to use .h for Objective-C++ headers. We also use MGLCamelCase for these headers.

platform/darwin/src/darwin_conversion.hpp
```diff
@@ -0,0 +1,132 @@
+#pragma once
```
1ec5: As long as this header is `#import`ed, no `#pragma once` guard is needed.

platform/darwin/src/darwin_conversion.hpp
```diff
@@ -0,0 +1,132 @@
...
+
+#include <mbgl/util/logging.hpp>
+#include <mbgl/style/conversion.hpp>
+#include <mbgl/util/feature.hpp>
+#include <mbgl/util/optional.hpp>
```
1ec5: Also import the Foundation/Foundation.h umbrella header.

platform/darwin/src/darwin_conversion.hpp
```diff
@@ -0,0 +1,132 @@
...
+public:
+    OptionalNSObjectValue(NSObject *_value) : value(_value) {}
+    OptionalNSObjectValue() {}
+    
+    explicit operator bool() const {
+        return value;
+    }
+    
+    NSObject* operator*() {
+        assert(this);
```
1ec5: Use `NSCAssert()` instead of `assert()` for files that are part of the Objective-C targets. `NSCAssert()` provides better diagnostics via NSException and works with Xcode’s built-in Objective-C exception breakpoint.

platform/darwin/src/darwin_conversion.hpp
```diff
@@ -0,0 +1,132 @@
...
+namespace mbgl {
+namespace style {
+namespace conversion {
+
+// A minimal wrapper class conforming to the requirements for `objectMember(v, name)` (see mbgl/style/conversion.hpp)
+// This is necessary because using `NSObject*` as the value type in `optional<NSObject*>` causes problems for the ARC,
+// due to things like `optional(const value_type& __v)`
+class OptionalNSObjectValue {
+public:
+    OptionalNSObjectValue(NSObject *_value) : value(_value) {}
```
1ec5: Annotate `value`’s type as `_Nullable` and surround the header’s contents in `NS_ASSUME_NONNULL_BEGIN` and `NS_ASSUME_NONNULL_END` to enable better compiler/analyzer diagnostics.

platform/darwin/src/darwin_conversion.hpp
```diff
@@ -0,0 +1,132 @@
...
+    OptionalNSObjectValue() {}
+    
+    explicit operator bool() const {
+        return value;
+    }
+    
+    NSObject* operator*() {
```
1ec5: Nit: move the space to before the `*`.

platform/darwin/src/MGLConversionValue.h
jfirebaugh: So this file should be removed now, right?

platform/darwin/src/darwin_conversion.hpp
```diff
@@ -0,0 +1,132 @@
...
+    return [value isKindOfClass:[NSArray class]];
+}
+
+inline bool isObject(const NSObject* value) {
+    return [value isKindOfClass:[NSDictionary class]];
+}
+
+inline std::size_t arrayLength(const NSObject* value) {
+    NSCAssert([value isKindOfClass:[NSArray class]], @"Value must be an NSArray for getLength().");
+    NSArray * array = (NSArray *)value;
```
1ec5: I’d still encourage you to explore the use of `id` instead of `NSObject *`. Saying “NSObject” here tells the compiler (and the code reviewer) that the function wants an NSObject specifically, not some subclass, which is why you end up having to cast all the time.

platform/darwin/src/darwin_conversion.hpp
```diff
@@ -0,0 +1,132 @@
...
+namespace style {
+namespace conversion {
+
+// A minimal wrapper class conforming to the requirements for `objectMember(v, name)` (see mbgl/style/conversion.hpp)
+// This is necessary because using `NSObject*` as the value type in `optional<NSObject*>` causes problems for the ARC,
+// due to things like `optional(const value_type& __v)`
+class OptionalNSObjectValue {
+public:
+    OptionalNSObjectValue(NSObject *_value) : value(_value) {}
+    OptionalNSObjectValue() {}
```
jfirebaugh: Omit this constructor...

platform/darwin/src/darwin_conversion.hpp
```diff
@@ -0,0 +1,132 @@
...
+inline optional<mbgl::Value> toValue(const NSObject* value) {
+    if (isUndefined(value)) {
+        return {};
+    } else if (_isBool(value)) {
+        return { *toBool(value) };
+    } else if ( _isString(value)) {
+        return { *toString(value) };
+    } else if (_isNumber(value)) {
+        // Need to cast to a double here as the float is otherwise considered a bool...
+       return { (double) *toNumber(value) };
```
1ec5: Use `static_cast` instead of a C-style cast.

platform/darwin/src/darwin_conversion.hpp
```diff
@@ -0,0 +1,132 @@
...
+
+inline OptionalNSObjectValue objectMember(const NSObject* value, const char* key) {
+    NSCAssert([value isKindOfClass:[NSDictionary class]], @"Value must be an NSDictionary for get(string).");
+    auto dict = (NSDictionary *)value;
+    NSObject *member = dict[@(key)];
+
+    if (value && value != [NSNull null]) {
+        return { member };
+    } else {
+        return {};
```
jfirebaugh: ...and `return { nullptr };` here.

platform/darwin/src/darwin_conversion.hpp
```diff
@@ -0,0 +1,132 @@
...
+    }
+}
+
+template <class Fn>
+optional<Error> eachMember(const NSObject*, Fn&&) {
```
jfirebaugh: Don't define this method at all. Then you'll get a compiler error if something tries to use it.

platform/darwin/src/MGLConversion.h
```diff
@@ -0,0 +1,136 @@
...
+    if (_isNumber(value)) {
+        return ((NSNumber *)value).floatValue;
+    } else {
+        return {};
+    }
+}
+
+inline optional<std::string> toString(const id value) {
+    if (_isString(value)) {
+        return std::string([((NSString *)value) UTF8String]);
```
1ec5: Nit: extra parentheses around `value` make the line harder to read. Incidentally, you can also write:

```objc
std::string((const char *)[value UTF8String])
```

platform/darwin/test/MGLStyleValueTests.swift
```diff
@@ -1,8 +1,61 @@
...
+        var expectedComponents : [CGFloat] = [0, 0, 0, 0]
+        actual.getRed(&(actualComponents[0]), green: &(actualComponents[1]), blue: &(actualComponents[2]), alpha: &(actualComponents[3]))
+        expected.getRed(&(expectedComponents[0]), green: &(expectedComponents[1]), blue: &(expectedComponents[2]), alpha: &(expectedComponents[3]))
+        for (ac, ec) in zip(actualComponents, expectedComponents) {
+            XCTAssertEqualWithAccuracy(Float(ac), Float(ec), accuracy: accuracy)
+        }
+    }
+    
+    func assertColorValuesEqual(_ actual: MGLStyleValue<MGLColor>, _ expected: MGLStyleValue<MGLColor>) {
+        if let actualConstant = actual as? MGLStyleConstantValue<MGLColor> {
```
1ec5: This is fine, but FYI you can avoid having to come up with a unique name for the coerced value by shadowing the original variable’s name:

```swift
if let actual = actual as? MGLStyleConstantValue<MGLColor> …
```

platform/darwin/test/MGLStyleValueTests.swift
```diff
@@ -1,8 +1,61 @@
...
+            XCTAssertEqual(actualFunction.interpolationMode, expectedFunction.interpolationMode)
+            // TODO: assert default values are equal
+            guard let actualStops = actualFunction.stops else {
+                XCTAssertNil(expectedFunction.stops)
+                return
+            }
+            let expectedStops = expectedFunction.stops!
+            XCTAssertEqual(actualStops.keys.count, expectedStops.keys.count)
+            let actualKeys = actualStops.keys.sorted(by: { String(describing: $0) < String(describing: $1) })
+            let expectedKeys = expectedStops.keys.sorted(by: { String(describing: $0) < String(describing: $1) })
```
1ec5: You can assert that `actualStops.keys == expectedStops.keys` to check both the count and the existence of identical pairs at the same time.

platform/darwin/test/MGLStyleValueTests.swift
```diff
@@ -1,8 +1,61 @@
...
+            XCTAssertEqual(actualFunction.interpolationBase, expectedFunction.interpolationBase)
+            XCTAssertEqual(actualFunction.interpolationMode, expectedFunction.interpolationMode)
+            // TODO: assert default values are equal
+            guard let actualStops = actualFunction.stops else {
+                XCTAssertNil(expectedFunction.stops)
+                return
+            }
+            let expectedStops = expectedFunction.stops!
+            XCTAssertEqual(actualStops.keys.count, expectedStops.keys.count)
+            let actualKeys = actualStops.keys.sorted(by: { String(describing: $0) < String(describing: $1) })
```
1ec5: Normally, you can call the operator directly instead of writing a new closure:

```swift
let actualKeys = actualStops.keys.sorted(by: <)
```

This syntax relies on type inference to choose the right operator. I’m not sure if it’s a problem that the compiler thinks the keys are of type `Any` rather than `Comparable`.

The original keys are `String`s in some cases and NSNumbers (or perhaps `Float`s?) in other cases. With NSNumber, you’d probably need `.floatValue` or at least `as Float`. With the current code, you’re going to wind up with some weird sorting in the case of NSNumber. For example, `"100" < "2"`. Fortunately, there shouldn’t be a problem as long as both arrays are sorted consistently.

platform/darwin/test/MGLStyleValueTests.swift
```diff
@@ -1,8 +1,61 @@
...
+            let expectedKeys = expectedStops.keys.sorted(by: { String(describing: $0) < String(describing: $1) })
+            
+            for (ak, ek) in zip(actualKeys, expectedKeys) {
```
1ec5: The documentation for `Array.values` says:

> When iterated over, values appear in this collection in the same order as they occur in the dictionary’s key-value pairs.

So it might be possible to say `zip(actual.values, expected.values)` to avoid the relative inefficiency of subscripting both dictionaries for each key.

platform/darwin/test/MGLStyleValueTests.swift
```diff
@@ -1,8 +1,61 @@
...
+    func assertColorValuesEqual(_ actual: MGLStyleValue<MGLColor>, _ expected: MGLStyleValue<MGLColor>) {
+        if let actualConstant = actual as? MGLStyleConstantValue<MGLColor> {
+            XCTAssertTrue(expected is MGLStyleConstantValue<MGLColor>)
+            assertColorsEqualWithAccuracy(actualConstant.rawValue, (expected as! MGLStyleConstantValue<MGLColor>).rawValue)
+        } else if let actualFunction = actual as? MGLStyleFunction<MGLColor> {
+            XCTAssertTrue(expected is MGLStyleFunction<MGLColor>)
+            let expectedFunction = expected as! MGLStyleFunction<MGLColor>
+            XCTAssertEqual(actualFunction.interpolationBase, expectedFunction.interpolationBase)
+            XCTAssertEqual(actualFunction.interpolationMode, expectedFunction.interpolationMode)
+            // TODO: assert default values are equal
```
1ec5: Does this still need to happen?

include/mbgl/style/conversion/function.hpp
```diff
@@ -200,7 +200,7 @@ struct Converter<CameraFunction<T>> {
 
 template <class T, class V>
 Result<optional<T>> convertDefaultValue(const V& value) {
-    auto defaultValueValue = objectMember(value, "defaultValue");
+    auto defaultValueValue = objectMember(value, "default");
```
jfirebaugh: Can you investigate why this mistake didn't trigger test failures? Contrary to my earlier statement, we do have tests which I would have expected to fail due to this, added in https://github.com/mapbox/mapbox-gl-js/commit/e328b807d9ad39a8194e9c13ab28ee1b16a8ce56.

anandthakker: 👍 yep, will do

platform/darwin/src/MGLConversion.h
```diff
@@ -0,0 +1,136 @@
...
+    auto length = [array count];
+    NSCAssert(length < std::numeric_limits<size_t>::max(), @"Array length out of bounds.");
+    return length;
+}
+
+inline NSObject *arrayMember(const id value, std::size_t i) {
+    NSCAssert([value isKindOfClass:[NSArray class]], @"Value must be an NSArray for get(int).");
+    NSCAssert(i < NSUIntegerMax, @"Index must be less than NSUIntegerMax");
+    NSArray *array = value;
+    return array[i];
```
anandthakker: 👍 @1ec5 I take it that the purpose of doing this would be for code conciseness, right?

platform/darwin/src/MGLConversion.h
anandthakker: In the form you suggested, would `static_cast` be preferable?

platform/darwin/test/MGLStyleValueTests.swift
anandthakker: Yeah -- I originally tried just using `<` as the sort function, but it seemed to think that the keys were of type `Any`.  I landed at the same conclusion as you: that the weird alphanumeric sort wouldn't matter here... _but_, alas, I think we should probably figure this key type thing out in more detail -- if the conversion stuff is turning numeric stop values into string values, that's a bug, and comparing stringified values of the keys like we're doing here would occlude it.

platform/darwin/test/MGLStyleValueTests.swift
anandthakker: It's not clear to me from that documentation that two _different_ dictionaries' `.values` would be in a consistent order... indeed, why would they?

platform/darwin/src/MGLConversion.h
1ec5: Yes. It matters more for more complex functions where you’d otherwise potentially have to keep both variables in sync.

platform/darwin/test/MGLStyleValueTests.swift
1ec5: What if?

```swift
var actualKeys: [Any]
if let keys = actualStops.keys as? [String] {
    actualKeys = keys.sorted(by: <)
} else if let keys = actualStops.keys as? [Float] {
    actualKeys = keys.sorted(by: <)
}
```

anandthakker:
@1ec5 updated to address the issues you raised, including an attempt to clean up the `assertColorValuesEqual`.  Note that in the end, I found using `.keys` to require much more weird acrobatics than just switching on the type of the dict itself -- but lemme know what you think.

include/mbgl/style/conversion/function.hpp
anandthakker: @jfirebaugh The reason it didn't trigger failures is that the tests are marked `ignored` on native.  I'll submit a PR to gl-js to un-ignore them.

platform/darwin/test/MGLStyleValueTests.swift
```diff
@@ -19,41 +19,62 @@ extension MGLStyleValueTests {
...
                     XCTAssertEqual(actualValue, expectedValue)
                 } else {
-                    XCTFail("Unsupported stop value type \(type(of: actualStops[ak])).")
+                    XCTFail("Unsupported stop value type \(type(of: actualValue)).")
                 }
             }
+            
+            XCTAssertEqual(actualStops.count, expectedStops.count)
+            if let actualStops = actualStops as? [String:Any], let expectedStops = expectedStops as? [String: Any] {
+                actualStops.forEach({ assertStopEqual($1, expectedStops[$0]) })
```
1ec5: Nit: when a function call ends with a closure, you can move it to after the argument list (after the parentheses), and you can also remove the parentheses if there are no other arguments. This is called [trailing closure syntax](https://developer.apple.com/library/prerelease/content/documentation/Swift/Conceptual/Swift_Programming_Language/Closures.html#//apple_ref/doc/uid/TP40014097-CH11-ID102):

```swift
actualStops.forEach { assertStopEqual($1, expectedStops[$0]) }
```

One way to think about trailing closure syntax is that it effectively allows you to define your own custom control statement as a function that takes a closure.

In this case, you aren’t really making use of any functional programming, so it’s clearer to read (and easier to debug) a standard `for` loop:

```swift
for (actual, expected) in actualStops {
    assertStopEqual(actual, expectedStops[expected])
}
```

platform/darwin/test/MGLStyleValueTests.swift
```diff
@@ -99,9 +120,9 @@ extension MGLStyleValueTests {
         var circleTranslationTwo = CGVector(dx: 0, dy: 0)
         let circleTranslationValueTwo = NSValue(bytes: &circleTranslationTwo, objCType: "{CGVector=dd}")
 
-        let circleTranslationStops = [
-            0: MGLStyleValue<NSValue>(rawValue: circleTranslationValueOne),
-            10: MGLStyleValue<NSValue>(rawValue: circleTranslationValueTwo)
+        let circleTranslationStops : [Float:MGLStyleValue<NSValue>] = [
+            0.0: MGLStyleValue<NSValue>(rawValue: circleTranslationValueOne),
```
1ec5: Are the `.0`s needed, now that you’ve specified the type above?

1ec5:
Asserting on the types of the dictionaries is a good idea, by the way!

PR #7630, opened 2017-01-08
[android] - rounded zoom when zooming with zoom button controller

tobrun:
This PR enables the zoom button controller to zoom to rounded values. 
eg. map with zoom 1.435 will zoom in to 2 and 3 instead of current 2.435 and 3.435. 

This PR also cleans up some duplicate code in MapGestureDetector related to zoom button controller.

cc  @lorntao @1ec5
Review @zugaldia

PR #8027, opened 2017-02-11
Round tap-zoom gestures to nearest integer

### LINKED ISSUE #2276: Zoom to rounded zoom levels for raster tiles
Hi,
Is there any way you could perhaps add an option to force the map to zoom to the closest integer zoom level when double-tapping or two-finger-tapping please? As it is right now if you pinch to 12.342 and then double tap, it will zoom to 13.342 instead of 13.0. This would be most useful for raster tiles with labels, otherwise text is too hard to read.

friedbunny:
### iOS changes
Rounds double-tap and two-finger tap zoom gestures to the nearest integer zoom level. This has benefits for raster tiles, as well as styles with zoom-based functions that transition on integers.

This results in a wider possible zoom range per gesticulation — 1.0 → ~0.5-1.5 levels. Here’s a comparison of double-tapping to zoom in with the changes in this PR:

| Starting Zoom | Old Result | New Result |
|:----------:|:-------------:|:------:|
| 4.0 |    5.0 (+1.0)   |   5.0 (+1.0) |
| 4.4 |    5.4 (+1.0)   |   **5.0** (+0.6) |
| 4.6 |    5.6 (+1.0)   |   **6.0** (+1.4) |

Fixes #2276. 

### Core changes
Exposes an already-existing `setZoom` variation that takes an `anchor` parameter — I did this to avoid calculating scale levels in the iOS zoom code. This makes `Map::setZoom(zoom, {}, duration)` ambiguous,  so it’s now necessary to specify a type for the middle parameter.

### DIFF
--- a/platform/ios/src/MGLMapView.mm
+++ b/platform/ios/src/MGLMapView.mm
@@ -1556,10 +1556,12 @@ public:
     if (doubleTap.state == UIGestureRecognizerStateEnded)
     {
         MGLMapCamera *oldCamera = self.camera;
-        
+
+        double newZoom = round(self.zoomLevel) + 1.0;
+
         CGPoint gesturePoint = [self anchorPointForGesture:doubleTap];
-        
-        MGLMapCamera *toCamera = [self cameraByZoomingToZoomLevel:self.zoomLevel + 1.0 aroundAnchorPoint:gesturePoint];
+
+        MGLMapCamera *toCamera = [self cameraByZoomingToZoomLevel:newZoom aroundAnchorPoint:gesturePoint];
         
         if (![self.delegate respondsToSelector:@selector(mapView:shouldChangeFromCamera:toCamera:)] ||
             [self.delegate mapView:self shouldChangeFromCamera:oldCamera toCamera:toCamera])
@@ -1567,7 +1569,7 @@ public:
             [self trackGestureEvent:MGLEventGestureDoubleTap forRecognizer:doubleTap];
             
             mbgl::ScreenCoordinate center(gesturePoint.x, gesturePoint.y);
-            _mbglMap->scaleBy(2, center, MGLDurationInSecondsFromTimeInterval(MGLAnimationDuration));
+            _mbglMap->setZoom(newZoom, center, MGLDurationInSecondsFromTimeInterval(MGLAnimationDuration));
             
             __weak MGLMapView *weakSelf = self;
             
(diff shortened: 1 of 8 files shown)


platform/ios/src/MGLMapView.mm
```diff
@@ -1556,18 +1556,20 @@ - (void)handleDoubleTapGesture:(UITapGestureRecognizer *)doubleTap
     if (doubleTap.state == UIGestureRecognizerStateEnded)
     {
         MGLMapCamera *oldCamera = self.camera;
-        
+
+        double newZoom = round(self.zoomLevel) + 1.0;
```
1ec5: Can you make the corresponding changes to the Zoom In and Zoom Out buttons in the Android and macOS SDKs?

https://github.com/mapbox/mapbox-gl-native/blob/e6c15d0f6f285bc604c0d93381b9b1d3957cb5c9/platform/android/MapboxGLAndroidSDK/src/main/java/com/mapbox/mapboxsdk/maps/Transform.java#L180-L184
https://github.com/mapbox/mapbox-gl-native/blob/e6c15d0f6f285bc604c0d93381b9b1d3957cb5c9/platform/macos/src/MGLMapView.mm#L1045

/cc @tobrun

friedbunny:
As an extreme example, here’s Mapbox Streets at z10.99 vs z11:

simulator screen shot feb 11 2017 2 59 11 pm copy

platform/ios/src/MGLMapView.mm
friedbunny: Went a bit further on macOS in 4ed80f9 and rounded all of the non-freeform zoom methods.

platform/macos/CHANGELOG.md
```diff
@@ -39,6 +39,7 @@
 * Fixed flickering that occurred when panning past the antimeridian. ([#7574](https://github.com/mapbox/mapbox-gl-native/pull/7574))
 * Added a method to MGLMapViewDelegate, `-mapView:shouldChangeFromCamera:toCamera:`, that you can implement to restrict which parts the user can navigate to using gestures. ([#5584](https://github.com/mapbox/mapbox-gl-native/pull/5584))
 * Added a `MGLDistanceFormatter` class for formatting geographic distances. ([#7888](https://github.com/mapbox/mapbox-gl-native/pull/7888))
+* Double-tap, two-finger tap, zoom buttons, and demo app menu items and shortcut keys now zoom to the nearest integer zoom level. ([#8027](https://github.com/mapbox/mapbox-gl-native/pull/8027))
```
1ec5: FYI, the ⌥↑ and ⌥↓ shortcuts are built into MGLMapView. The menu items (which have their own shortcuts) are defined in MapDocument but could conceivably be moved into MGLMapView as well.

friedbunny: Ah, did not know that — I’ll update the text here.

1ec5:
It looks like #7630 would’ve implemented a similar fix on the Android side. Not sure why @tobrun closed it.

platform/ios/src/MGLMapView.mm
friedbunny: I’ll leave Android to @tobrun, in light of #7630.

tobrun:
@1ec5 @friedbunny I initially created that PR to showcase the possibility of such a feature but this didn't make it to master at that time as I didn't want to change the "default" behaviour. We could definitely look into reopening it for consistency between the bindings. 

cc @mapbox/android

zugaldia:
> We could definitely look into reopening it for consistency between the bindings.

Yeah, let's do that for 5.0. It doesn't look like a big lift.

1ec5:
The Android side is happening once again in #7630. 🚢

PR #8033, opened 2017-02-13
[macos] Make + and - keys zoom the map

### LINKED ISSUE #8029: + and − keys should zoom in and out
⌘+ and ⌘− zoom in and out, respectively, but + and − do not. When MGLMapView is the first responder, it should respond to + and −, as well as = (as an alias for +).

/cc @friedbunny

eimantas:
My take on improving parity with MapKit. Fixes #8029.

### DIFF
--- a/platform/macos/src/MGLMapView.mm
+++ b/platform/macos/src/MGLMapView.mm
@@ -1633,18 +1633,11 @@ public:
 #pragma mark Keyboard events
 
 - (void)keyDown:(NSEvent *)event {
-    NSEventModifierFlags modFlags = event.modifierFlags;
-    BOOL noModFlags = modFlags == 256;
-    BOOL isZoomEvent = ([@[@"-", @"="] containsObject:[event characters]] && noModFlags) ||
-                        ([@"+" isEqual:[event characters]] && (modFlags & NSShiftKeyMask));
-
-    if (modFlags & NSNumericPadKeyMask) {
+    if (event.modifierFlags & NSNumericPadKeyMask) {
         // This is the recommended way to handle arrow key presses, causing
         // methods like -moveUp: and -moveToBeginningOfParagraph: to be called
         // for various standard keybindings.
         [self interpretKeyEvents:@[event]];
-    } else if (isZoomEvent) {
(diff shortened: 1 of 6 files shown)


platform/macos/src/MGLMapView.mm
```diff
@@ -1618,11 +1618,18 @@ - (BOOL)gestureRecognizer:(NSGestureRecognizer *)gestureRecognizer shouldAttempt
 #pragma mark Keyboard events
 
 - (void)keyDown:(NSEvent *)event {
-    if (event.modifierFlags & NSNumericPadKeyMask) {
+    NSEventModifierFlags modFlags = event.modifierFlags;
+    BOOL noModFlags = modFlags == 256;
+    BOOL isZoomEvent = ([@[@"-", @"="] containsObject:[event characters]] && noModFlags) ||
+                        ([@"+" isEqual:[event characters]] && (modFlags & NSShiftKeyMask));
```
1ec5: Does + require the Shift key on all keyboard layouts?

platform/macos/src/MGLMapView.mm
```diff
@@ -1618,11 +1618,18 @@ - (BOOL)gestureRecognizer:(NSGestureRecognizer *)gestureRecognizer shouldAttempt
...
+
+    if (modFlags & NSNumericPadKeyMask) {
         // This is the recommended way to handle arrow key presses, causing
         // methods like -moveUp: and -moveToBeginningOfParagraph: to be called
         // for various standard keybindings.
         [self interpretKeyEvents:@[event]];
+    } else if (isZoomEvent) {
+        [self adjustZoomLevelForKey:[event characters]];
```
1ec5: If I'm not mistaken, a more common pattern is to call [`-interpretKeyEvents:`](https://developer.apple.com/reference/appkit/nsresponder/1531599-interpretkeyevents?language=objc), as above, then override `-insertText:`. That approach would keep us from having to worry about the vagaries of input methods and keyboard layouts.

eimantas: I am not very much familiar with macOS programming (let alone keyboard event handling). I'll update the code to reflect your recommendation.

PR #8041, opened 2017-02-13
[ios] Feature/cachebykey

wibge:
Added an optional cachekey for tilesets. If present, it is used instead of the url template for caching in the database. This is useful when tile servers urls changed, but you still want the offline cache to work. 

This happens often for tile servers that have the access key as part of the url. But we've also used it before when changing USGS topographic servers or aerial imagery servers to different providers. 

You specify the cacheKey in the tilejson and via a MGLTileSourceOption

### DIFF
--- a/include/mbgl/storage/resource.hpp
+++ b/include/mbgl/storage/resource.hpp
@@ -23,6 +23,7 @@ public:
 
     struct TileData {
         std::string urlTemplate;
+        std::string cacheKey;
         uint8_t pixelRatio;
         int32_t x;
         int32_t y;
@@ -44,6 +45,7 @@ public:
     static Resource style(const std::string& url);
     static Resource source(const std::string& url);
     static Resource tile(const std::string& urlTemplate,
+                         const std::string& cacheKey,
                          float pixelRatio,
                          int32_t x,
                          int32_t y,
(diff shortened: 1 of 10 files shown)


1ec5:
> This happens often for tile servers that have the access key as part of the url.

Part of the motivation of #7485 was to generalize the handling of the Mapbox-specific `access_token` query parameter to work with other tile servers, so that you could keep an access key parameter out of the cache key – or indeed any parameter that should change without busting caches, such as a timestamp or campaign ID.

> You specify the cacheKey in the tilejson and via a MGLTileSourceOption

The TileJSON format is defined in [this specification](https://github.com/mapbox/tilejson-spec/). It might be more convenient to define this option as an inline option in the style JSON format, similar to `tileSize`. It would still require a change to the [Mapbox Style Specification](https://www.mapbox.com/mapbox-gl-js/style-spec/), but that wouldn’t be quite as big of a deal.

jfirebaugh:
Hi @wibge, thanks for the contribution. #8084 adds the ability to transform URLs prior to the request but after the URL is recorded for cache purposes. It was motivated by similar needs; will it address your use cases? I'd like to minimize the feature set around this if possible.

wibge:
It looks like that will meet our needs for this. Thanks. I'll close this now.

PR #8042, opened 2017-02-13
Safer image

jfirebaugh:
Provide `Image::copy`, which handles copying rectangles from a source to a destination, with thorough bounds checking.

Also fixes an indexing error in SpriteAtlas, where the top row of pixels in a wrapped image was copied from the wrong source row (https://github.com/mapbox/mapbox-gl-js/pull/4264).


include/mbgl/util/image.hpp
```diff
@@ -58,6 +71,50 @@ class Image : private util::noncopyable {
...
     size_t bytes() const { return stride() * size.height; }
 
+    void fill(uint8_t value) {
+        std::fill(data.get(), data.get() + bytes(), value);
+    }
+
+    // Copy image data within `rect` from `src` to the rectangle of the same size at `pt`
+    // in `dst`. If the specified bounds exceed the bounds of the source or destination,
+    // throw `std::out_of_range`. Must not be used to move data within a single Image.
+    static void copy(const Image& srcImg, Image& dstImg, const Point<uint32_t>& srcPt, const Point<uint32_t>& dstPt, const Size& size) {
```
jfirebaugh: This could alternatively be a member function on the source or a member function on the destination. I like it as a static method because the arguments are given equal prominence, and it's easier to see the parallelism in argument ordering: `srcImg`, `dstImg`, `srcPt`, `dstPt`.

kkaefer: Agreed, I find member functions for this very confusing, since various APIs use different copy directions (e.g. Canvas uses `drawImage` to draw the passed argument into itself, but Node's `Buffer.copy` copies *into* the passed argument...)

include/mbgl/util/image.hpp
```diff
@@ -58,6 +71,50 @@ class Image : private util::noncopyable {
...
+            srcPt.y > srcImg.size.height - size.height) {
+            throw std::out_of_range("out of range source coordinates for image copy");
+        }
+
+        if (size.width > dstImg.size.width ||
+            size.height > dstImg.size.height ||
+            dstPt.x > dstImg.size.width - size.width ||
+            dstPt.y > dstImg.size.height - size.height) {
+            throw std::out_of_range("out of range destination coordinates for image copy");
+        }
```
kkaefer: Instead of failing hard in the out of range case, we could also clip to the common intersection, which is what Canvas does.

src/mbgl/text/glyph_pbf.cpp
```diff
@@ -64,26 +65,35 @@ void parseGlyphPBF(GlyphSet& glyphSet, const GlyphRange& glyphRange, const std::
...
-                glyph.metrics.advance < 256 && glyph.bitmap.size() == expectedBitmapSize &&
-                glyph.id >= glyphRange.first && glyph.id <= glyphRange.second) {
-                glyphSet.insert(glyph.id, std::move(glyph));
+            if (glyph.metrics.width && glyph.metrics.height) {
+                const Size size {
+                    glyph.metrics.width + 2 * SDFGlyph::borderSize,
+                    glyph.metrics.height + 2 * SDFGlyph::borderSize
+                };
+
+                if (size.area() != glyphData.size()) {
```
kkaefer: Why are we comparing area here instead of width/height?

jfirebaugh: The width/height comparison would be `size.width * size.height != glyphData.size()`. Is that what you mean?

include/mbgl/util/image.hpp
jfirebaugh: The out of range case is a logic error in all current uses of `copy`, so I think it makes sense to fail hard.

src/mbgl/text/glyph_pbf.cpp
kkaefer: Looking at this again; the question is moot.

PR #8046, opened 2017-02-14
[macos] Make +, - and = keys zoom in/out 

eimantas:
I messed up previous commit with some rebase. This one contains all the changes from review of #8033.

### DIFF
--- a/platform/macos/src/MGLMapView.mm
+++ b/platform/macos/src/MGLMapView.mm
@@ -1633,13 +1633,16 @@ public:
 #pragma mark Keyboard events
 
 - (void)keyDown:(NSEvent *)event {
-    if (event.modifierFlags & NSNumericPadKeyMask) {
-        // This is the recommended way to handle arrow key presses, causing
-        // methods like -moveUp: and -moveToBeginningOfParagraph: to be called
-        // for various standard keybindings.
-        [self interpretKeyEvents:@[event]];
+    [self interpretKeyEvents:@[event]];
+}
+
+- (void)insertText:(NSString *)text {
+    BOOL textIsZoomCharacter = [@[@"-", @"=", @"+"] containsObject:text];
+
+    if (textIsZoomCharacter) {
+        [self adjustZoomLevelForKey:text];
     } else {
-        [super keyDown:event];
+        [super insertText:text];
     }
 }
 
@@ -1693,6 +1696,23 @@ public:
     _rotateEnabled = rotateEnabled;
     _compass.enabled = rotateEnabled;
     _compass.hidden = !rotateEnabled;
+}
+
+- (void)adjustZoomLevelForKey:(NSString *)key {
+    double newZoomLevel = self.zoomLevel;
+    unichar keyUnichar = [key characterAtIndex:0];
+    switch(keyUnichar) {
+        case '-': {
+            newZoomLevel -= 1;
+            break;
+        }
+        case '=':
+        case '+': {
+            newZoomLevel += 1;
+        }
+    }
+
+    [self setZoomLevel:newZoomLevel animated:YES];
 }
 
 #pragma mark Ornaments
--- a/platform/macos/src/MGLMapView.mm
+++ b/platform/macos/src/MGLMapView.mm
@@ -1633,16 +1633,40 @@ public:
 #pragma mark Keyboard events
 
 - (void)keyDown:(NSEvent *)event {
+    // This is the recommended way to handle arrow key presses, causing
+    // methods like -moveUp: and -moveToBeginningOfParagraph: to be called
+    // for various standard keybindings.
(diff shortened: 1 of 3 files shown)


platform/macos/src/MGLMapView.mm
```diff
@@ -1695,6 +1696,21 @@ - (void)setRotateEnabled:(BOOL)rotateEnabled {
     _compass.hidden = !rotateEnabled;
 }
 
+- (void)adjustZoomLevelForKey:(NSString *)key {
+    static NSDictionary<NSString *, void(^)(void)> *zoomOpDict = nil;
+
+    if (zoomOpDict == nil) {
+        void (^zoomOut)(void) = ^ { [self setZoomLevel:(self.zoomLevel - 1) animated:YES]; };
+        void (^zoomIn)(void) = ^ { [self setZoomLevel:(self.zoomLevel + 1) animated:YES]; };
+        zoomOpDict = @{@"-": zoomOut, @"+": zoomIn, @"=": zoomIn};
```
1ec5: Perhaps you could simplify this code with a series of if/else statements that sets a local `newZoomLevel` variable, followed by a single call to `-setZoomLevel:`. If calling `-[NSString isEqualToString:]` in multiple places is undesirable, then you could also switch on the string’s first character.

platform/macos/src/MGLMapView.mm
```diff
@@ -1633,13 +1633,14 @@ - (BOOL)gestureRecognizer:(NSGestureRecognizer *)gestureRecognizer shouldAttempt
 #pragma mark Keyboard events
 
 - (void)keyDown:(NSEvent *)event {
-    if (event.modifierFlags & NSNumericPadKeyMask) {
-        // This is the recommended way to handle arrow key presses, causing
-        // methods like -moveUp: and -moveToBeginningOfParagraph: to be called
-        // for various standard keybindings.
-        [self interpretKeyEvents:@[event]];
-    } else {
-        [super keyDown:event];
```
1ec5: `-keyDown:` and `-insertText:` should call through to the superclass’s implementation for any unhandled keys/text so that, for example, pressing <kbd>M</kbd> beeps and pressing ⇥ or ⇧⇥ cycles to another responder.

platform/macos/src/MGLMapView.mm
eimantas: AFAIR Objective-C `switch` switches over integers only. No?

1ec5: True, so you could say `[key characterAtIndex:0]` and switch over that `unichar`.

platform/macos/src/MGLMapView.mm
eimantas: I'm not quite clear how should I go about calling `[super keyDown:event]` in this method. Should I check if the event contains known keystrokes or call `super` in any case?

platform/macos/src/MGLMapView.mm
```diff
@@ -1695,6 +1698,23 @@ - (void)setRotateEnabled:(BOOL)rotateEnabled {
...
+            newZoomLevel -= 1;
+            break;
+        }
+        case '=':
+        case '+': {
+            newZoomLevel += 1;
+        }
+    }
+
+    [self setZoomLevel:newZoomLevel animated:YES];
```
1ec5: It just occurred to me that we have existing methods that zoom in and out (for ⌘+ and ⌘−, respectively): `-moveToBeginningOfParagraph:` and `-moveToEndOfParagraph:`. By calling those methods (so named because of the ⌥↑ and ⌥↓ key bindings), we’ll also respect the `zoomEnabled` property and the rounding added in #8027.

platform/macos/src/MGLMapView.mm
1ec5: [This document](https://developer.apple.com/library/content/documentation/Cocoa/Conceptual/EventOverview/HandlingKeyEvents/HandlingKeyEvents.html#//apple_ref/doc/uid/10000060i-CH7-SW13) has a useful code snippet. (Why oh why isn’t this the default on macOS?) The following code respects ⇥ and ⇧⇥ while beeping for unrecognized keys:

```objc
- (void)keyDown:(NSEvent *)event {
    // This is the recommended way to handle arrow key presses, causing
    // methods like -moveUp: and -moveToBeginningOfParagraph: to be called
    // for various standard keybindings.
    [self interpretKeyEvents:@[event]];
}

// The following action methods are declared in NSResponder.h.

- (void)insertTab:(id)sender {
    if (self.window.firstResponder == self) {
        [self.window selectNextKeyView:self];
    }
}

- (void)insertBacktab:(id)sender {
    if (self.window.firstResponder == self) {
        [self.window selectPreviousKeyView:self];
    }
}

- (void)insertText:(id)insertString {
    …
}
```

Issue #4411, opened 2016-03-21
Ambient caching disabled when the database contains 50 MB+ of offline resources


jfirebaugh:
Once an offline database contains 50 MB or more of offline resources, it effectively no longer permits any ambient caching. Any attempt to add an ambiently-cached resource to the database will evict any existing ambiently-cached resources, fail to get the database below 50 MB, and log the error "Unable to make space for entry".

The 50 MB limit should be treated as the size limit for ambiently-cached resources only, not as a size limit on the complete database.

RomainQuidet:
Confirmed, seeing a lot of logs
2016-04-04 11:45:17.778 Mapbox GL[6813:244371] [WARNING] {DefaultFileSource}[Database]: Unable to make space for entry

When the 50MB limit is reached, I have the feeling that the map lags a lot, showing new tiles is really slow.

halset:
Is this to expensive to run instead of the pragma trick?
```
sqlite> select coalesce(sum(length(data)),0) from tiles where id not in (select tile_id from region_tiles);
0
Run Time: real 0.013 user 0.005565 sys 0.004869
```

tmpsantos:
> The 50 MB limit should be treated as the size limit for ambiently-cached resources only, not as a size limit on the complete database.

Unless we split this into two different databases, I see this as a feature not as a bug. Otherwise we would have to calculate "offline size + ambient size" to make sure it fits into the space you want to reserve for caching.

It is convenient to download a 600 MB offline data and set the total size to 1 GB and forget about it. The current implementation will make sure we stay within limits.

jfirebaugh:
I think the behavior most end user / developers would expect is:

* Use up to X MB of space for ambiently cached data.
* Store all the requested offline regions, without limit.

That's not possible right now.

halset:
@tmpsantos you wrote "It is convenient to download a 600 MB offline data and set the total size to 1 GB and forget about it". Is there an API to set a custom max cache size from the iOS API?

friedbunny:
@halset Nope, maximum cache size is current defined as [a constant in mbgl](https://github.com/mapbox/mapbox-gl-native/blob/eb1b385e50f783e33bb500693977f7bf45f4126e/include/mbgl/util/constants.hpp#L39).

halset:
@friedbunny thanks.

Trying to figure out the best way to fix this for mye users with more than 50MB of offline data. 

[...]

PR #8054, opened 2017-02-14
[ios] new eviction strategy to fix ambient caching when offline resources are downloaded

wibge:
This is a new eviction policy to address: https://github.com/mapbox/mapbox-gl-native/issues/4411 

With this strategy, the entire maxCacheSize is available for ambiently cached items. Instead of checking for file space on every insert of an ambient item, it keeps a running tally of the size of inserted tiles, and checks if it needs to purge every maxCacheSize/10.

On an iPhone 5S, with a 500MB offline database, running the purge operation takes 0.2 seconds. Checking if it needs to purge takes 0.02 seconds. On the iphone 7, about 0.1 seconds to check and 0.3 to purge. Maybe the thread is more likely to be interrupted there?

The database is still vacuumed only after offline packs are deleted.

### DIFF
--- a/platform/default/mbgl/storage/offline_database.cpp
+++ b/platform/default/mbgl/storage/offline_database.cpp
@@ -19,6 +19,7 @@ OfflineDatabase::OfflineDatabase(std::string path_, uint64_t maximumCacheSize_)
     : path(std::move(path_)),
       maximumCacheSize(maximumCacheSize_) {
     ensureSchema();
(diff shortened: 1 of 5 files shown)


ChrisLoer:
Looks like maybe this fell through the cracks and we never got a chance to review/comment? @kkaefer  do you think this approach makes sense? I haven't wrapped my head around the problem at all, but those purge times do sound pretty high.

@wibge are you still at Gaia GPS? Can you :wave: hi to @lucaswoj for us?

PR #8055, opened 2017-02-14
Recreate pending and raw layer ref when layer is removed

### LINKED ISSUE #8053: Removing a style layer before adding the same layer should not cause a crash
From https://github.com/mapbox/mapbox-gl-native/pull/7962#issuecomment-279551127:

> On iOS/macOS, -[MGL*StyleLayer removeFromMapView:] clears its internal state before telling mbgl to remove the layer from the style. If the layer isn’t part of the style, -removeFromMapView: will bail, resulting in an unusable MGLStyleLayer object. The following code would crash:

> MGLBackgroundStyleLayer *fillLayer = [[MGLBackgroundStyleLayer alloc] initWithIdentifier:@"bg"];
[style removeLayer:fillLayer]; // no-op
// …
[style addLayer:fillLayer]; // MGLRedundantLayerException
I think the correct behavior would be to ensure that _pendingLayer ends up with whatever either _pendingLayer or rawLayer had at the beginning of this method.

I think we can address this by moving the pointer / property assign statements in between the early returns of the methods and setting them to values that they would have had when the Objective-C layer object was created in the first place (https://github.com/mapbox/mapbox-gl-native/pull/7962#issuecomment-279791747)

cc @1ec5

boundsj:
Fixes #8053 

cc @1ec5

### DIFF
--- a/platform/darwin/src/MGLBackgroundStyleLayer.mm
+++ b/platform/darwin/src/MGLBackgroundStyleLayer.mm
@@ -62,13 +62,13 @@
 
 - (void)removeFromMapView:(MGLMapView *)mapView
 {
-    _pendingLayer = nullptr;
-    self.rawLayer = nullptr;
-
     auto removedLayer = mapView.mbglMap->removeLayer(self.identifier.UTF8String);
     if (!removedLayer) {
         return;
     }
+    auto originalLayer = std::make_unique<mbgl::style::BackgroundLayer>(self.identifier.UTF8String);
+    _pendingLayer = std::move(originalLayer);
+    self.rawLayer = _pendingLayer.get();
 
     mbgl::style::BackgroundLayer *layer = dynamic_cast<mbgl::style::BackgroundLayer *>(removedLayer.get());
     if (!layer) {
--- a/platform/darwin/src/MGLCircleStyleLayer.mm
+++ b/platform/darwin/src/MGLCircleStyleLayer.mm
@@ -112,13 +112,14 @@ namespace mbgl {
 
 - (void)removeFromMapView:(MGLMapView *)mapView
 {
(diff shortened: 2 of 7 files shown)


boundsj:
When #7962 lands, this PR could be updated to add tests for the scenario described in https://github.com/mapbox/mapbox-gl-native/pull/7962#issuecomment-279551127

platform/darwin/src/MGLBackgroundStyleLayer.mm
```diff
@@ -62,13 +62,13 @@ - (void)addToMapView:(MGLMapView *)mapView belowLayer:(MGLStyleLayer *)otherLaye
...
-    _pendingLayer = nullptr;
-    self.rawLayer = nullptr;
-
     auto removedLayer = mapView.mbglMap->removeLayer(self.identifier.UTF8String);
     if (!removedLayer) {
         return;
     }
+    auto originalLayer = std::make_unique<mbgl::style::BackgroundLayer>(self.identifier.UTF8String);
+    _pendingLayer = std::move(originalLayer);
+    self.rawLayer = _pendingLayer.get();
```
1ec5: If I understand correctly, these new lines are only relevant to the case where we bail below due to removing a layer of a different type than this class, correct? Otherwise, we’ll end up making the removed layer the pending layer.

boundsj: That is correct @1ec5. It is only to protect against that edge. I guess that could happen if you add a layer of type foo with identity bar and then (before adding) try to remove a layer of type baz with identity bar.

boundsj:
This is now happening in https://github.com/mapbox/mapbox-gl-native/pull/7962

PR #8060, opened 2017-02-15
prefix constants with mapbox_

### LINKED ISSUE #8016: Bundle constants aren't prefixed in MapboxConstants
There are situations that a user could save a key inside his Activity bundle with the same value as one of ours keys in `MapboxConstants`. This could result in overwriting our saved state and thus we would be unable to restore our state correctly. To avoid such issues we should prefix the constant values ,as we have been doing for our resources, with a unique mapbox prefix. 

cc @mapbox/android

tobrun:
closes #8016


tobrun:
@zugaldia correct

PR #8066, opened 2017-02-15
[android] - use project wide configurations for all modules

### LINKED ISSUE #7732: Use project wide configurations for modules
When looking in to changing the build tools version in #7729, I noticed that I had to change this in each module in this project. It's possible to configure this type of configuration once in a the top level gradle file and let those values change the module configurations. Example is shown [here](https://developer.android.com/studio/build/gradle-tips.html#configure-project-wide-properties).

cc @mapbox/android

tobrun:
This PR introduces a `dependencies.gradle` file where we can define project wide configurations.
Review @zugaldia 
Closes #7732


platform/android/MapboxGLAndroidSDKWearTestApp/src/main/java/com/mapbox/weartestapp/utils/OffsettingHelper.java
```diff
@@ -34,7 +34,7 @@ public void updateChild(View child, WearableRecyclerView parent) {
   }
 
   @Override
-  protected void adjustAnchorOffsetXY(View child, float[] anchorOffsetXY) {
+  public void adjustAnchorOffsetXY(View child, float[] anchorOffsetXY) {
```
zugaldia: @tobrun This looks like coming from a different PR.

platform/android/MapboxGLAndroidSDK/build.gradle
```diff
@@ -24,29 +20,23 @@ dependencies {
 }
 
 android {
-    compileSdkVersion Integer.parseInt(project.ANDROID_BUILD_SDK_VERSION)
-    buildToolsVersion project.ANDROID_BUILD_TOOLS_VERSION
+    compileSdkVersion rootProject.ext.compileSdkVersion
```
zugaldia: @tobrun Why not doing this too with all compile dependencies like MAS or `compile 'com.squareup.okhttp3:okhttp:3.5.0'`, `compile 'com.mapzen.android:lost:2.0.0'`, `compile 'com.jakewharton.timber:timber:4.3.1'`?

platform/android/MapboxGLAndroidSDKWearTestApp/src/main/java/com/mapbox/weartestapp/utils/OffsettingHelper.java
tobrun: to double check if the wear module was working correctly I ran it on an emulator but couldn't compile as AS showed this as a violation of visibility of the base class method. (I added that as second part of the commit message: `... + fixup wear compile error`

platform/android/MapboxGLAndroidSDK/build.gradle
tobrun: let me try that, now I just did the duplicate ones.

tobrun:
comments have been fixed up in https://github.com/mapbox/mapbox-gl-native/pull/8066/commits/8abcfe5fa9cee067ac24a896d7ea7e010e3d9a04

platform/android/MapboxGLAndroidSDKWearTestApp/src/main/java/com/mapbox/weartestapp/utils/OffsettingHelper.java
zugaldia: Sorry, missed that commit.

PR #7833, opened 2017-01-24
[ios] Disable region monitoring when stopping location monitoring

JesseCrocker:
See #7831 for description

PR #8067, opened 2017-02-15
[iOS] Remove geofences when location monitoring is stopped

erichoracek:
I’ve tested this change in an app that was opted out of MBGL location collection and have verified that the iOS location services preferences no longer reported the app as using geofences.

As far as I can tell, without the changes in this commit, the `MGLLocationManager` would errantly attempt to start location updates when the user exited a previously monitored geofence _even if telemetry collection was disabled by the user_. This would occur when the user exited a region that was previously monitored, according to the current logic in `-[MGLLocationManager locationManager:didExitRegion:]`, since they were not removed when `-[MGLLocationManager  stopUpdatingLocation]` was invoked.

Of course, please let me know if I'm missing something here. Thanks!


boundsj:
Thanks @erichoracek. This looks like https://github.com/mapbox/mapbox-gl-native/pull/7833. I would prefer to close this in favor of #7833 since we are already planning this change there for our next release.

erichoracek:
Sure, thanks! I looked at issues but not at open pull requests.

Issue #5364, opened 2016-06-15
[test] Timer.Basic is flaky


brunoabinader:
`Timer.Basic` fails sporadically on Travis - unfortunately I don't have a build log link as Travis cleans up when we retry the build.

:eyes: @tmpsantos

jfirebaugh:
Haven't seen this recently.

PR #8073, opened 2017-02-15
[tests] Disable timing-based Timer tests in CI

jfirebaugh:
These tests have always been flaky (https://github.com/mapbox/mapbox-gl-native/issues/5364), and started failing more consistently on macOS CI after upgrading the Bitrise stack.


1ec5:
Cherry-picked into the release-ios-v3.4.0 branch in #8100.

Issue #8077, opened 2017-02-15
MGLCircleStyleLayer with transparent fill & non-transparent stroke not rendering


jmkiley:
**Platform:** iOS
**Mapbox SDK version:** v3.5.0-beta.1

### Steps to trigger behavior

 1. Add a `MGLCircleStyleLayer`
 2. Set `circleOpacity` to 0
 3. Set `circleStrokeColor` to 1

[The code I used](https://gist.github.com/jmkiley/67ed51a3c04423173d10e2285a0c5d9c)

### Expected behavior

For the outline of a circle to appear on the map.

### Actual behavior

The circles are not visible.

One possible work around is to set `circleOpacity` to a low value (like 0.000000001).

PR #8078, opened 2017-02-15
Render a circle if either the color or stroke color are visible

boundsj:
This adds a line to pass the circle layer render check if the circle stroke color is visible. Previously, a transparent circle color would cause the layer to be passed up even if the circle stroke color had
a non-zero alpha.

cc @jfirebaugh @jmkiley

Fixes https://github.com/mapbox/mapbox-gl-native/issues/8077

### DIFF
--- a/src/mbgl/style/layers/circle_layer_impl.cpp
+++ b/src/mbgl/style/layers/circle_layer_impl.cpp
@@ -14,9 +14,12 @@ void CircleLayer::Impl::cascade(const CascadeParameters& parameters) {
 bool CircleLayer::Impl::evaluate(const PropertyEvaluationParameters& parameters) {
     paint.evaluate(parameters);
 
-    passes = (paint.evaluated.get<CircleRadius>().constantOr(1) > 0
-           && paint.evaluated.get<CircleColor>().constantOr(Color::black()).a > 0
-           && paint.evaluated.get<CircleOpacity>().constantOr(1) > 0)
+    passes = ((paint.evaluated.get<CircleRadius>().constantOr(1) > 0 ||
+               paint.evaluated.get<CircleStrokeWidth>().constantOr(1) > 0)
+           && (paint.evaluated.get<CircleColor>().constantOr(Color::black()).a > 0 ||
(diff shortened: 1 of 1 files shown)


src/mbgl/style/layers/circle_layer_impl.cpp
```diff
@@ -15,7 +15,8 @@ bool CircleLayer::Impl::evaluate(const PropertyEvaluationParameters& parameters)
     paint.evaluate(parameters);
 
     passes = (paint.evaluated.get<CircleRadius>().constantOr(1) > 0
-           && paint.evaluated.get<CircleColor>().constantOr(Color::black()).a > 0
+           && (paint.evaluated.get<CircleColor>().constantOr(Color::black()).a > 0 ||
+               paint.evaluated.get<CircleStrokeColor>().constantOr(Color::black()).a > 0)
            && paint.evaluated.get<CircleOpacity>().constantOr(1) > 0)
```
jfirebaugh: This needs to check `CircleStrokeOpacity` as well.

src/mbgl/style/layers/circle_layer_impl.cpp
```diff
@@ -15,7 +15,8 @@ bool CircleLayer::Impl::evaluate(const PropertyEvaluationParameters& parameters)
     paint.evaluate(parameters);
 
     passes = (paint.evaluated.get<CircleRadius>().constantOr(1) > 0
```
jfirebaugh: ...and this needs to check `CircleStrokeWidth`.

boundsj:
I'll work on adding the test @jfirebaugh. So far, I'm stuck in gl-js land. When I run `npm run test-suite` I get:

```
/Users/boundsj/mapbox-gl-js/src/util/window.js:30
    window.HTMLElement.prototype.clientLeft = 0;
TypeError: Cannot set property clientLeft of [object Object] which has only a getter
```

Maybe that has something to do with jsdom?

boundsj:
I think this will require https://github.com/mapbox/mapbox-gl-native/pull/7944 since the `mapbox-gl-js` submodule change also pulls in some unskipped tests that won't pass until #7944 lands.

Issue #4386, opened 2016-03-18
OfflineManager and MapView should share a singleton DefaultFileSource


1ec5:
As of #4383, all instances of `mbgl::DefaultFileSource` in the Android SDK are backed by the same file on disk. Per #4362, the singleton `OfflineManager` and all `mbgl::View` instances managed by `MapView`s should ideally share the same instance of `mbgl::DefaultFileSource`, so that the 20-connection throttle can be applied to the whole process instead of on a map-by-map basis.

In the iOS and OS X SDKs, in #4377, we made the singleton MGLOfflineStorage object responsible for the singleton `mbgl::DefaultFileSource` instance. Thus MGLMapView depends on MGLOfflineStorage (as ambient caching is just a special case of offline caching), but not the other way around.

/cc @zugaldia @jfirebaugh

tobrun:
This landed as part of https://github.com/mapbox/mapbox-gl-native/pull/8083

Issue #4250, opened 2016-03-08
Offline terminology


jfirebaugh:
For offline, we are inventing a lot of new terminology. For prior SDK features, we've taken cues from existing mapping SDKs, which in many cases leads to divergence of terminology between our iOS and Android SDKs. But offline is a feature other SDKs don't have, and I think it would be to the long term benefit of our customers and support staff to use consistent terminology for offline concepts -- consistent both between the iOS and Android offline APIs, and the words we use when talking about offline in prose.

A proposal follows. The terms are different that what we've been using, but don't panic: I don't think renaming will be a huge undertaking, provided we do it _now_, before these APIs have been formally released and we've written more than one piece of documentation about them.
## **offline resource collection**

Type names: `{mbgl::,MGL,com.com.mapbox.mapboxsdk.offline.}OfflineResourceCollection`.

An **offline resource collection** is the basic unit of offline. You can create them, list them, start and stop their downloads, and delete them. Each resource collection has a _region_, and may have some additional application-supplied data. (Former core/Android name: `OfflineRegion`.)
## **offline region**

Type names: `{mbgl::,MGL,com.com.mapbox.mapboxsdk.offline.}OfflineRegion`, `{mbgl::,MGL,com.com.mapbox.mapboxsdk.offline.}OfflineTilePyramidRegion`

The geographic extent of a resource collection -- which part of the world it covers -- is defined by its **region**. We plan eventually to support region types, but for now there is only a single type: a tile pyramid. (Former core/Android name: `OfflineRegionDefinition` / `OfflineRegionTilePyramidDefinition`.)
## application-supplied data

Here the terminology follows established platform-specific conventions: on Android, application-supplied data is called "metadata". On iOS, it's called "context". Both use preexisting platform types.
## **offline download status** / **offline download progress**

The current revision of the iOS bindings separates these into two concepts. The core API and Android bindings have a single concept combining both. TBD.
## **resource database**

Type names: `{mbgl::,MGL,com.com.mapbox.mapboxsdk.offline.}ResourceDatabase`.

[...]

PR #8084, opened 2017-02-16
Allow transforming URLs before requesting

kkaefer:
An alternative to https://github.com/mapbox/mapbox-gl-native/pull/7485. This implements a way of modifying the URLs that get requested from the internet. Potential use cases are altering URLs or hostnames and supporting use cases as described in https://github.com/mapbox/mapbox-gl-native/issues/7455

This implementation invokes the supplied callback on the main thread (where the `DefaultFileSource:: setResourceTransform` happens) to avoid threading issues. I benchmarked the time it takes to transform the URL, and typically got times <1ms roundtrip (for scheduling the callback on the main thread and returning the result to the `OnlineFileSource` thread), but it can sometimes be higher when the main thread is busy rendering the map.

To do:
- [x] Android implementation
- [x] Unit tests

Caveats:
- Does not allow transforming `asset://` and `file://` URLs; the transform is only called on requests that go out to the internet
- When a request is scheduled, the internet connection goes offline for a long time, and then goes back online, it'll use the URL that was transformed during initialization; it does *not* attempt to transform the URL again when the request is attempted again

When using this feature, note that requests will still be cached by their canonical path, and not the transformed path. This means that the transformed URL should point to the same resources (e.g. on a different server, or with an time-limited access token) to avoid a caching mismatch. This is up to the server operator and application developer to ensure.

### DIFF
--- a/platform/darwin/src/MGLFoundation_Private.h
+++ b/platform/darwin/src/MGLFoundation_Private.h
@@ -1,0 +1,5 @@
+#import "MGLFoundation.h"
+
+#include <mbgl/util/run_loop.hpp>
+
+extern mbgl::util::RunLoop mgl_runLoop;
--- a/platform/darwin/src/MGLOfflineStorage.h
+++ b/platform/darwin/src/MGLOfflineStorage.h
@@ -7,6 +7,7 @@ NS_ASSUME_NONNULL_BEGIN
 
 @class MGLOfflinePack;
 @protocol MGLOfflineRegion;
+@protocol MGLOfflineStorageDelegate;
 
 /**
  Posted by the shared `MGLOfflineStorage` object when an `MGLOfflinePack`
@@ -132,6 +133,30 @@ typedef void (^MGLOfflinePackAdditionCompletionHandler)(MGLOfflinePack * _Nullab
 typedef void (^MGLOfflinePackRemovalCompletionHandler)(NSError * _Nullable error);
 
 /**
+ The type of resource that is requested.
+ */
+typedef NS_ENUM(NSUInteger, MGLResourceKind) {
+    /** Unknown type */
+    MGLResourceKindUnknown,
+    /** Style sheet JSON file */
+    MGLResourceKindStyle,
+    /** TileJSON file as specified in https://www.mapbox.com/mapbox-gl-js/style-spec/#root-sources */
+    MGLResourceKindSource,
+    /** A vector or raster tile as described in the style sheet at
+        https://www.mapbox.com/mapbox-gl-js/style-spec/#sources */
+    MGLResourceKindTile,
+    /** Signed distance field glyphs for text rendering. These are the URLs specified in the style
+        in https://www.mapbox.com/mapbox-gl-js/style-spec/#root-glyphs */
+    MGLResourceKindGlyphs,
+    /** Image part of a sprite sheet. It is constructed of the prefix in
+        https://www.mapbox.com/mapbox-gl-js/style-spec/#root-sprite and a PNG file extension. */
+    MGLResourceKindSpriteImage,
+    /** JSON part of a sprite sheet. It is constructed of the prefix in
+        https://www.mapbox.com/mapbox-gl-js/style-spec/#root-sprite and a JSON file extension. */
+    MGLResourceKindSpriteJSON,
+};
+
+/**
  MGLOfflineStorage implements a singleton (shared object) that manages offline
  packs. All of this class’s instance methods are asynchronous, reflecting the
  fact that offline resources are stored in a database. The shared object
@@ -144,6 +169,20 @@ MGL_EXPORT
  Returns the shared offline storage object.
  */
 + (instancetype)sharedOfflineStorage;
+
+#pragma mark - Accessing the Delegate
+
+/**
+ The receiver’s delegate.
+
+ An offline storage object sends messages to its delegate to allow it to
+ transform URLs before they are requested from the internet. This can be used
+ add or remove custom parameters, or reroute certain requests to other servers
+ or endpoints.
+ */
+@property(nonatomic, weak, nullable) IBOutlet id<MGLOfflineStorageDelegate> delegate;
+
+#pragma mark - Managing Offline Packs
 
 /**
  An array of all known offline packs, in the order in which they were created.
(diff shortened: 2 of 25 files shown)


platform/darwin/src/MGLFoundation_Private.h
```diff
@@ -0,0 +1,3 @@
+#import "MGLFoundation.h"
+
+void MGLInitializeRunLoop();
```
kkaefer: Removed this method

kkaefer:
Here's how to transform URLs on Android:

```diff
diff --git a/platform/android/MapboxGLAndroidSDKTestApp/src/main/java/com/mapbox/mapboxsdk/testapp/MapboxApplication.java b/platform/android/MapboxGLAndroidSDKTestApp/src/main/java/com/mapbox/mapboxsdk/testapp/MapboxApplication.java
index a10c6ea..80877e0 100644
--- a/platform/android/MapboxGLAndroidSDKTestApp/src/main/java/com/mapbox/mapboxsdk/testapp/MapboxApplication.java
+++ b/platform/android/MapboxGLAndroidSDKTestApp/src/main/java/com/mapbox/mapboxsdk/testapp/MapboxApplication.java
@@ -4,13 +4,16 @@ import android.app.Application;
 import android.os.StrictMode;
 
 import com.mapbox.mapboxsdk.Mapbox;
+import com.mapbox.mapboxsdk.storage.Resource;
+import com.mapbox.mapboxsdk.offline.OfflineManager;
+import com.mapbox.mapboxsdk.offline.OfflineManager.ResourceTransformCallback;
 import com.squareup.leakcanary.LeakCanary;
 
 import timber.log.Timber;
 
 import static timber.log.Timber.DebugTree;
 
-public class MapboxApplication extends Application {
+public class MapboxApplication extends Application implements ResourceTransformCallback {
 
   @Override
   public void onCreate() {
@@ -38,6 +41,11 @@ public class MapboxApplication extends Application {
       .build());
 
     Mapbox.getInstance(getApplicationContext(), getString(R.string.mapbox_access_token));
+    OfflineManager.getInstance(getApplicationContext()).setResourceTransform(this);
+  }
+
+  public String onURL(@Resource.Kind int kind, String url) {
+    return url;
   }
 
   private void initializeLogger() {
```

and here's how to do it on iOS:

```diff
diff --git a/platform/ios/app/MBXAppDelegate.h b/platform/ios/app/MBXAppDelegate.h
index 7ff321b..f70973e 100644
--- a/platform/ios/app/MBXAppDelegate.h
+++ b/platform/ios/app/MBXAppDelegate.h
@@ -1,8 +1,8 @@
-#import <UIKit/UIKit.h>
+#import <Mapbox/Mapbox.h>
 
 extern NSString * const MBXMapboxAccessTokenDefaultsKey;
 
-@interface MBXAppDelegate : UIResponder <UIApplicationDelegate>
+@interface MBXAppDelegate : UIResponder <UIApplicationDelegate, MGLOfflineStorageDelegate>
 
 @property (strong, nonatomic) UIWindow *window;
 
diff --git a/platform/ios/app/MBXAppDelegate.m b/platform/ios/app/MBXAppDelegate.m
index c2834bf..b8cfd56 100644
--- a/platform/ios/app/MBXAppDelegate.m
+++ b/platform/ios/app/MBXAppDelegate.m
@@ -23,7 +23,16 @@ - (BOOL)application:(UIApplication *)application didFinishLaunchingWithOptions:(
         [MGLAccountManager setAccessToken:accessToken];
     }
 
+    [[MGLOfflineStorage sharedOfflineStorage] setDelegate:self];
+
     return YES;
 }
 
+- (NSURL *)offlineStorage:(MGLOfflineStorage *)storage
+    URLForResourceOfKind:(MGLResourceKind)kind
+                 withURL:(NSURL *)url {
+    // Change the URL here if required.
+    return url;
+}
+
 @end
```

The approach is very similar in both SDKs: implement the delegate/interface and provide a method that gets called whenever a URL needs transforming. The URL is called on the main thread.

kkaefer:
This PR also includes a bare-bones version of https://github.com/mapbox/mapbox-gl-native/issues/4386 for Android. Long term, we should align the iOS and Android implementations and move the ownership of `DefaultFileSource` to `OfflineManager`.

kkaefer:
Instead of passing a `Resource&&`, we could also pass a `const Resource&`, and expect a `std::string` as a return value. This would reflect the fact that the `url` is the only thing that can be changed (changes to all other values are ignored, since their values have been used prior to this call to construct this URL). I.e. it's not meaningful to change `kind`, `necessity`, or `tileData`, and the user probably shouldn't change `priorModified`, `priorExpires` and `priorEtag` either.

platform/darwin/src/MGLOfflineStorage.h
```diff
@@ -132,6 +133,30 @@ typedef void (^MGLOfflinePackAdditionCompletionHandler)(MGLOfflinePack * _Nullab
...
 
 /**
+ The type of resource that is requested.
+ */
+typedef NS_ENUM(NSUInteger, MGLResourceKind) {
+    /** Unknown type */
+    MGLResourceKindUnknown,
+    /** Style sheet JSON file */
+    MGLResourceKindStyle,
+    /** TileJSON file as specified in https://www.mapbox.com/mapbox-gl-js/style-spec/#root-sources */
```
1ec5: Use an `<a>` tag with the text "Mapbox Style Specification", so that the documentation looks cleaner in the jazzy documentation.

1ec5: I suppose we could alternatively refer to the TileJSON file as the file specified by the `configurationURL` parameter in MGLTileSource's initializers.

platform/darwin/src/MGLOfflineStorage.h
```diff
@@ -250,4 +289,25 @@ MGL_EXPORT
...
+/**
+ Sent whenever a URL needs to be transformed.
+
+ @param storage The storage object processing the download.
+ @param kind The kind of URL to be transformed.
+ @param url The original URL to be transformed.
+ @return A URL that will now be downloaded.
+ */
+- (NSURL *)offlineStorage:(MGLOfflineStorage *)storage
+    URLForResourceOfKind:(MGLResourceKind)kind
```
1ec5: Nit: adding the space before the asterisk caused these colons to become misaligned. 😅

kkaefer:
Longer term, we should follow the iOS model, which has a wrapper class for `DefaultFileSource` (which is essentially what `OfflineManager` is). On iOS, it's called `MGLOfflineStorage`, but it's kind of a misnomer since it's also used in online environments. A better name would've been `MGLStorageManager`. On Android, we could e.g. rename `OfflineManager` to `StorageManager`, and have a Map object take a reference to a `StorageManager` on construction.

1ec5:
> On iOS, it's called `MGLOfflineStorage`, but it's kind of a misnomer since it's also used in online environments. A better name would've been `MGLStorageManager`. On Android, we could e.g. rename `OfflineManager` to `StorageManager`, and have a Map object take a reference to a `StorageManager` on construction.

Reopened https://github.com/mapbox/mapbox-gl-native/issues/4250#issuecomment-281186462 to track the renaming discussion.

PR #8087, opened 2017-02-17
add make android-help target

incanus:
Per chat, this helps navigate the myriad Android architecture options so that we don't have to `grep` the `Makefile` all the time (assuming it's kept up to date). 

- Does this cover our bases? 
- Is it likely to get updated as things move along? 
- Anything else? 

@mapbox/android


incanus:
I realize we could iterate the ABI list, but that doesn't get the backwards-compat options, nor the generic `make android`, so I think a manual entry is the way to go.

zugaldia:
@incanus love it, thanks for doing this. How about we also add the meta-targets `make android` and `make apackage` information?

incanus:
> How about we also add the meta-targets `make android` and `make apackage` information?

The `make android` is there (see the aliases under `android-arm-v7-9`) but I'll add the other.

incanus:
How about now, with indentation for the aliases, as well as `make apackage`?

PR #8092, opened 2017-02-17
Transform::moveBy with AnimationOptions

### LINKED ISSUE #8081: mbgl::AnimationOptions with mbgl::Transform::moveBy
I'm proposing altering `Transfrom::moveBy` to allow using `AnimationOptions`:

```cpp
//before
void moveBy(const ScreenCoordinate& offset, const Duration& = Duration::zero());

//after
void moveBy(const ScreenCoordinate& offset, const AnimationOptions& options);
```

This allows for bindings to provide a `mbgl::util::UnitBezier` as interpolator to achieve more natural `moveBy` transformations. On Android this would benefit the user experience around the fling gesture.

Current linear interpolated moveBy:

ezgif com-video-to-gif 18

With a circular ease out interpolation: 

ezgif com-video-to-gif 16

cc @jfirebaugh

tobrun:
Closes #8081, this PR enables to use AnimationOptions when performing a moveBy transformation.
For Android I'm adding `mbgl::util::UnitBezier { 0, 0.3, 0.6, 1.0 }` as interpolator to get a better UX for current fling gesture:

Before:
ezgif com-video-to-gif 19

After:
ezgif com-video-to-gif 20

### DIFF
--- a/platform/android/src/jni.cpp
+++ b/platform/android/src/jni.cpp
@@ -484,7 +484,14 @@ void nativeMoveBy(JNIEnv *env, jni::jobject* obj, jlong nativeMapViewPtr, jdoubl
                           jlong duration) {
     assert(nativeMapViewPtr != 0);
     NativeMapView *nativeMapView = reinterpret_cast<NativeMapView *>(nativeMapViewPtr);
-    nativeMapView->getMap().moveBy({dx, dy}, mbgl::Milliseconds(duration));
+
+    mbgl::AnimationOptions animationOptions;
+    if (duration > 0) {
+       animationOptions.duration.emplace(mbgl::Milliseconds(duration));
+       animationOptions.easing.emplace(mbgl::util::UnitBezier { 0, 0.3, 0.6, 1.0 });
+    }
+
+    nativeMapView->getMap().moveBy({dx, dy}, animationOptions);
 }
 
 void nativeSetLatLng(JNIEnv *env, jni::jobject* obj, jlong nativeMapViewPtr, jdouble latitude, jdouble longitude, jlong duration) {
--- a/src/mbgl/map/transform.cpp
+++ b/src/mbgl/map/transform.cpp
@@ -322,7 +322,7 @@ void Transform::flyTo(const CameraOptions &camera, const AnimationOptions &anima
 
 #pragma mark - Position
 
-void Transform::moveBy(const ScreenCoordinate& offset, const Duration& duration) {
+void Transform::moveBy(const ScreenCoordinate& offset, const AnimationOptions& animation) {
     ScreenCoordinate centerOffset = {
         offset.x,
         -offset.y,
@@ -331,7 +331,7 @@ void Transform::moveBy(const ScreenCoordinate& offset, const Duration& duration)
 
     CameraOptions camera;
     camera.center = state.screenCoordinateToLatLng(centerPoint);
-    easeTo(camera, duration);
+    easeTo(camera, animation);
 }
 
 void Transform::setLatLng(const LatLng& latLng, const Duration& duration) {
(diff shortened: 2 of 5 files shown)


src/mbgl/map/transform.cpp
```diff
@@ -322,7 +322,7 @@ void Transform::flyTo(const CameraOptions &camera, const AnimationOptions &anima
 
 #pragma mark - Position
 
-void Transform::moveBy(const ScreenCoordinate& offset, const Duration& duration) {
+void Transform::moveBy(const ScreenCoordinate& offset, const AnimationOptions &animation) {
```
jfirebaugh: Nit: move the space after the `&`.

platform/android/src/jni.cpp
```diff
@@ -484,6 +484,13 @@ void nativeMoveBy(JNIEnv *env, jni::jobject* obj, jlong nativeMapViewPtr, jdoubl
...
     assert(nativeMapViewPtr != 0);
     NativeMapView *nativeMapView = reinterpret_cast<NativeMapView *>(nativeMapViewPtr);
+
+    mbgl::AnimationOptions animationOptions;
+    if (duration > 0) {
+       animationOptions.duration.emplace(mbgl::Milliseconds(duration));
+       animationOptions.easing.emplace(mbgl::util::UnitBezier { 0, 0.3, 0.6, 1.0 });
+    }
+
     nativeMapView->getMap().moveBy({dx, dy}, mbgl::Milliseconds(duration));
```
jfirebaugh: `animationOptions` isn't being used here. Looks like you'll need to change `Map::moveBy` before that's possible.

tobrun: 🤦‍♂️

tobrun:
Review comments have been fixed up

tobrun:
Updated PR for iOS and QT

platform/ios/src/MGLMapView.mm
```diff
@@ -1224,7 +1224,7 @@ - (void)handlePanGesture:(UIPanGestureRecognizer *)pan
         if (![self.delegate respondsToSelector:@selector(mapView:shouldChangeFromCamera:toCamera:)] ||
             [self.delegate mapView:self shouldChangeFromCamera:oldCamera toCamera:toCamera])
         {
-            _mbglMap->moveBy({ delta.x, delta.y });
+            _mbglMap->moveBy({ delta.x, delta.y }, mbgl::AnimationOptions{});
```
1ec5: You can probably omit `mbgl::AnimationOptions` and leave just `{}`, since there’s only one `moveBy()` method and no ambiguity as to the second parameter’s type.

As an alternative to making this change, you could also make `{}` the default for the second parameter to `mbgl::Map::moveBy()`, as you did in `mbgl::Transform::moveBy()`.

tobrun: nice, thank you for 👀

platform/ios/src/MGLMapView.mm
```diff
@@ -1224,7 +1224,7 @@ - (void)handlePanGesture:(UIPanGestureRecognizer *)pan
         if (![self.delegate respondsToSelector:@selector(mapView:shouldChangeFromCamera:toCamera:)] ||
             [self.delegate mapView:self shouldChangeFromCamera:oldCamera toCamera:toCamera])
         {
-            _mbglMap->moveBy({ delta.x, delta.y });
+            _mbglMap->moveBy({ delta.x, delta.y }, {});
```
1ec5: With the default parameter value defined, you don’t even need the `{}` anymore.

tobrun:
That was exactly what I was trying to achieve, didn't want to impact other bindings. 
Thank you for walking me through this @1ec5.

PR #8095, opened 2017-02-17
Android: Refactor Mapbox singleton initialization

kkaefer:


kkaefer:
Long term goal is to remove parameters from `.getInstance()` calls, since this allows passing different parameters in different invocations, but we only have the one singleton from the first invocation.

jfirebaugh:
Stale.

Issue #7295, opened 2016-12-05
Camera updates get called even when the maps camera's already there.


cammace:
While looking into whether the same behavior described in https://github.com/mapbox/mapbox-gl-native/issues/7119 exist on Android also, I noticed we aren't checking whether the camera is already at the position requested. For example, when I call `animateCamera` to move the camera to a new position, and once the animation ends, call it again using the same position, we are still making a call to `nativeFlyTo`. 

cc: @tobrun

tobrun:
This should be feasible to add, check if the current camera position in `Transform.java` isn't the same as the newly set one. Note that LatLng values rely on double and you can have some floating point issues. 

eg. moving a camera object to Latitude 37 can result in getting 36.99999999999. 

I see 2 solutions for this:
 - we compare 2 LatLngs with a delta
 - we store previous set camera position from a `CameraUpdateFactory` method and decouple it from the current cameraPosition found in `Transform` (this one is being recreated from core).

PR #8097, opened 2017-02-17
fix the animation with two positions too close

TranThuong:
Hello,

We have an issue in animating with the method MapboxMap.animateCamera(...), the zoom is from 16 to 7, the center isn't changed. That make the mapbox in the unstable state, the center and the bounding box are NaN. We must restart the application.

After further investigation, we identify the root cause of the problem : 
In the class transform.cpp, the method Transform::flyTo cause the issue. If endPoint and startPoint are too close (<1) the following piece of code return "inf" due to "std::sqrt(b * b + 1) - b" equals absolute 0 then "std::log(std::sqrt(b * b + 1) - b);" becomes std::log(0) = -inf

` auto r = [=](double i) {
        /// bᵢ
        double b = (w1 * w1 - w0 * w0 + (i ? -1 : 1) * rho2 * rho2 * u1 * u1) / (2 * (i ? w1 : w0) * rho2 * u1);
        return std::log(std::sqrt(b * b + 1) - b);
    };` 

My proposition to fix it is to make a protection if the delta of the  endPoint and startPoint is smaller than 0.5 we set the delta is 0.5. That fix our issue.

### DIFF
--- a/src/mbgl/map/transform.cpp
+++ b/src/mbgl/map/transform.cpp
@@ -211,6 +211,11 @@ void Transform::flyTo(const CameraOptions &camera, const AnimationOptions &anima
     /// Length of the flight path as projected onto the ground plane, measured
     /// in pixels from the world image origin at the initial scale.
     double u1 = ::hypot((endPoint - startPoint).x, (endPoint - startPoint).y);
+    //fix: if startpoint and endpoint are too close, we force the distance at 1 px
+    //      otherwise r(x) could return inf
+    if (u1<0.5) {
+        u1 = 0.5;
+    }
 
     /** ρ: The relative amount of zooming that takes place along the flight
         path. A high value maximizes zooming for an exaggerated animation, while


tobrun:
this would close https://github.com/mapbox/mapbox-gl-native/issues/7295 on Android

1ec5:
> this would close #7295 on Android

I don’t think that’s the case. It merely loosens the threshold for turning a `flyTo()` into an `easeTo()`; an animation takes place regardless.

1ec5:
@mourner, do you recall where [this threshold](https://github.com/mapbox/mapbox-gl-native/pull/8097/files#diff-ab28faca52ff082a59393dbf97ee6fe2R253) came from? I can’t find anything similar in [the original paper](https://www.win.tue.nl/~vanwijk/zoompan.pdf) describing flight animation.

mourner:
@1ec5 it came originally from @ansis's implementation here https://github.com/cutting-room-floor/easey-DEAD/pull/20.

tobrun:
@TranThuong been looking into the same issue myself in light of #8300. It seems that the value of u1=0.5 is not sufficient. In #8302 I'm trying to catch the root cause that could avoid this behaviour for all use-cases. Thank you for reaching out and contributing. Closing this in favor of #8302.

PR #8099, opened 2017-02-17
[core] Fix HeadlessBackend context management

### LINKED ISSUE #7681: Flapping render tests -- "shader failed to compile"
Examples:

https://travis-ci.org/mapbox/mapbox-gl-native/jobs/191016628
https://travis-ci.org/mapbox/mapbox-gl-native/jobs/191016630

I've also seen `GL_INVALID_OPERATION` errors, probably related.

### LINKED ISSUE #8076: Wrong destruction order for backend/context
`HeadlessBackend` destroys the OpenGL context, and then tries to destroy resources (shaders, programs, buffers, etc.) used by that context. I'm surprised this works at all. It could be contributing to #7681.

[`HeadlessBackend` derives from `Backend`](https://github.com/mapbox/mapbox-gl-native/blob/117863f1114551407c481abc752f5fcfd139c878/platform/default/mbgl/gl/headless_backend.hpp#L14-L14) and [`Backend` holds a `std::unique_ptr<gl::Context>`](https://github.com/mapbox/mapbox-gl-native/blob/c76a933514e4e1514a58ac0e668b13eebab37794/include/mbgl/map/backend.hpp#L45-L45). Therefore, the destruction order is:

1. [`~HeadlessBackend` calls `destroyContext()`](https://github.com/mapbox/mapbox-gl-native/blob/117863f1114551407c481abc752f5fcfd139c878/platform/default/mbgl/gl/headless_backend.cpp#L21-L21)
4. [`destroyContext()` releases `HeadlessBackend::Impl`](https://github.com/mapbox/mapbox-gl-native/blob/117863f1114551407c481abc752f5fcfd139c878/platform/default/mbgl/gl/headless_backend.cpp#L53-L53)
4. [`~EGLImpl` (or equivalent) calls `eglDestroyContext()` (or equivalent)](https://github.com/mapbox/mapbox-gl-native/blob/e987e406c7864440d02dc634b9d29777f353f21b/platform/linux/src/headless_backend_egl.cpp#L34-L45)
4. [The default `~Backend` calls `~std::unique_ptr<gl::Context>`](https://github.com/mapbox/mapbox-gl-native/blob/a4d259c33f9bb890bba97fd89552720e3e0ec09b/src/mbgl/map/backend.cpp#L15-L15)
5. [`~Context` calls `reset()`](https://github.com/mapbox/mapbox-gl-native/blob/0bbc6b814cbec44be7026a0bac83d56e4d71a287/src/mbgl/gl/context.cpp#L38-L38)
6. [`reset()` calls `performCleanup()`](https://github.com/mapbox/mapbox-gl-native/blob/0bbc6b814cbec44be7026a0bac83d56e4d71a287/src/mbgl/gl/context.cpp#L355-L355)
7. [`performCleanup()` deletes a bunch of context resources](https://github.com/mapbox/mapbox-gl-native/blob/0bbc6b814cbec44be7026a0bac83d56e4d71a287/src/mbgl/gl/context.cpp#L496-L557)

This needs to be reversed, so that `~Context` happens before `eglDestroyContext()`.

cc @kkaefer

jfirebaugh:
Eliminate manual activate/deactivate, and always use `BackendScope`. `BackendScope` now works as a stack, pushing the new context when coming into scope and restoring the prior context when going out of scope.

Fixes #7681
Fixes #8076

### DIFF
--- a/include/mbgl/map/backend_scope.hpp
+++ b/include/mbgl/map/backend_scope.hpp
@@ -1,0 +1,17 @@
+#pragma once
+
+namespace mbgl {
+
+class Backend;
+
+class BackendScope {
+public:
+    BackendScope(Backend&);
+    ~BackendScope();
+
+private:
+    Backend* priorBackend;
+    Backend& backend;
+};
+
+} // namespace mbgl
--- a/include/mbgl/map/backend_scope.hpp
+++ b/include/mbgl/map/backend_scope.hpp
@@ -10,7 +10,8 @@ public:
     ~BackendScope();
 
 private:
-    Backend* priorBackend;
+    BackendScope* priorScope;
+    BackendScope* nextScope;
     Backend& backend;
 };
 
--- a/src/mbgl/map/backend_scope.cpp
+++ b/src/mbgl/map/backend_scope.cpp
@@ -1,0 +1,26 @@
+#include <mbgl/map/backend_scope.hpp>
+#include <mbgl/map/backend.hpp>
+#include <mbgl/util/thread_local.hpp>
+
+namespace mbgl {
+
+static util::ThreadLocal<Backend> currentBackend;
+
+BackendScope::BackendScope(Backend& backend_)
+    : priorBackend(currentBackend.get()),
+      backend(backend_) {
+    backend.activate();
+    currentBackend.set(&backend);
+}
+
+BackendScope::~BackendScope() {
+    if (priorBackend) {
+        priorBackend->activate();
+        currentBackend.set(priorBackend);
+    } else {
+        backend.deactivate();
+        currentBackend.set(nullptr);
+    }
+}
+
+} // namespace mbgl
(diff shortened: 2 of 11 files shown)


include/mbgl/map/backend_scope.hpp
```diff
@@ -0,0 +1,17 @@
...
+
+class Backend;
+
+class BackendScope {
+public:
+    BackendScope(Backend&);
+    ~BackendScope();
+
+private:
+    Backend* priorBackend;
```
kkaefer: As an additional safeguard to aid in debugging, we could make this a doubly linked list and `assert` that there is not a `nextBackend` in the destructor.

src/mbgl/map/backend_scope.cpp
```diff
@@ -0,0 +1,26 @@
...
+
+namespace mbgl {
+
+static util::ThreadLocal<Backend> currentBackend;
+
+BackendScope::BackendScope(Backend& backend_)
+    : priorBackend(currentBackend.get()),
+      backend(backend_) {
+    backend.activate();
+    currentBackend.set(&backend);
```
kkaefer: Can we check whether the backend is identical to the previously enabled backend and make activation/deactivation a no-op in this case?

kkaefer: Looks like this actually never happens in our code, so probably not worth the increased complexity.

PR #8112, opened 2017-02-18
Reuse OpenGL context when switching screens

1ec5:
Cherry-picked from #8004.


kkaefer:
Verified that moving windows between screens works.

PR #8117, opened 2017-02-20
[android] fix unit tests for #7944

ivovandongen:
https://github.com/mapbox/mapbox-gl-native/pull/7944 left some unit tests broken, the generation script generated invalid test assertions.

cc @anandthakker


anandthakker:
Thanks @ivovandongen -- I was having trouble getting tests to even run locally, and I mistakenly took the CI success to mean that the tests were okay, forgetting that it's not currently reliable to do so.

PR #8121, opened 2017-02-20
[android] Fix a NPE that occurs when MapView is initialized with null options

benoitletondor:
This commit fix a NPE that occurs when calling `MapView` constructor with a null `MapboxMapOptions` object.

It crashes when reaching:

```
initalizeDrawingSurface(context, options);
```

on the first line of the method:

```
if (options.getTextureMode()) {
```


zugaldia:
Thank you @benoitletondor for the report and your contribution! Tagging @tobrun, @Guardiola31337 for review.

tobrun:
Continuing this PR in #8131

PR #8122, opened 2017-02-20
add setting to toggle two maps for thread debugging

incanus:
Per chat with @kkaefer, a quick settings toggle in the iOS demo app for toggling a second map view for use in debugging a shared thread pool in https://github.com/mapbox/mapbox-gl-native/pull/8094. 

The bulk of the work is updating constraints to keep the two map views in the right place & size during device rotation. 

second


1ec5:
Filed #8128 to replace this demo with a more robust split view.

Issue #5236, opened 2016-06-04
Add pause/resume API to DefaultFileSource


boundsj:
Currently, instances of DefaultFileSource allow revalidation requests regardless of if the host process is running in the background and the map that would benefit from the requests is visible. Mobile applications that are extremely concerned about limiting network requests that is not strictly necessary   may see battery performance gains if the revalidation requests were not made in the background. 

In general, this behavior probably makes sense as the default on both iOS and Android. In addition, as usage of the SDK grows, the backend will benefit from not having to respond to as many requests as it does today.

We should consider adding and exposing an API to DefaultFileSource (or similar approach) so that each platform can inform it if it makes sense to carryout network requests based on background state (or other factors).

cc @jfirebaugh @1ec5 @mick

1ec5:
In the iOS SDK, we have a natural mechanism to make use of a request-pausing API, in `-[MGLMapView sleepGL:]` (which happens when the application recedes to the background or enters background mode) or a similar method that would be added to MGLOfflineStorage.

We could also more aggressively pause at the point where we stop a map view's CADisplayLink (which would also happen when the map view is removed from the view hierarchy or hidden). However, the catch is that the application could have multiple views, so we'd need to refcount somehow to avoid interfering with still-visible views.

tmpsantos:
This could maybe implemented more generically at our class abstraction like `util::Thread::pause()`.

boundsj:
This was fixed by https://github.com/mapbox/mapbox-gl-native/pull/8125 and https://github.com/mapbox/mapbox-gl-native/pull/8194

Issue #4841, opened 2016-04-26
Fatal Exception: mapbox::sqlite::Exception: disk I/O error


MikeKasperlik:
We see a crash that happens frequently. I don't know how to trigger this. We also do nothing fancy with mapbox currently, only setting the camera to the users location + creating some annotations.

According to crashlytics this only happens when the app is not in foreground - maybe an issue with file protection?

**Platform:** iOS 9.2.1 - iOS 9.3.1
**Mapbox SDK version:** 3.2.1

```
Crashed: DefaultFileSource
0  PACEApp                        0x1005075ac CLSProcessRecordAllThreads + 4300240300
1  PACEApp                        0x1005075ac CLSProcessRecordAllThreads + 4300240300
2  PACEApp                        0x1005079cc CLSProcessRecordAllThreads + 4300241356
3  PACEApp                        0x1004f8680 CLSHandler + 4300179072
4  PACEApp                        0x100505cc4 __CLSExceptionRecord_block_invoke + 4300233924
5  libdispatch.dylib              0x1811fd47c _dispatch_client_callout + 16
6  libdispatch.dylib              0x181208728 _dispatch_barrier_sync_f_invoke + 100
7  PACEApp                        0x100505770 CLSExceptionRecord + 4300232560
8  PACEApp                        0x1005052f4 CLSTerminateHandler() + 4300231412
9  libc++abi.dylib                0x180e0af44 std::__terminate(void (*)()) + 16
10 libc++abi.dylib                0x180e0a85c __cxxabiv1::exception_cleanup_func(_Unwind_Reason_Code, _Unwind_Exception*) + 134
11 Mapbox                         0x10149fc04 mapbox::sqlite::Statement::run() + 156
12 Mapbox                         0x10149a50c long long mbgl::OfflineDatabase::getPragma<long long>(char const*) + 36
13 Mapbox                         0x101498f80 mbgl::OfflineDatabase::evict(unsigned long long) + 72
14 Mapbox                         0x101498e74 mbgl::OfflineDatabase::putInternal(mbgl::Resource const&, mbgl::Response const&, bool) + 280
15 Mapbox                         0x101489f40 mbgl::DefaultFileSource::Impl::Task::Task(mbgl::Resource, std::__1::function<void (mbgl::Response)>, mbgl::DefaultFileSource::Impl*)::'lambda'(mbgl::Response)::operator()(mbgl::Response) const + 40
[...]

Issue #7443, opened 2016-12-15
Updating offline maps on cellular


jmkiley:
Currently, offline maps are updated whenever device detects a network connection. The user would need to disable this using their platform's connectivity monitoring API if they would like to prevent users from downloading offline maps over a cellular connection. 

Would it be possible to let the user to choose whether they want to allow offline maps to update while the end user is connected to cellular? 

cc @friedbunny

jfirebaugh:
Note that "updated" here means the following: when connectivity is available, resources (principally tiles) which are needed to render visible regions of the map are checked for freshness according to the caching headers on the most recently stored response. There's no "en masse" update of whole offline regions -- only the resources that are actually used are updated.

barnaclebarnes:
+1 for this. I want to be able to 'lock' the offline map so that it doesn't create any network traffic when viewing the offline map. Even if the user scrolls outside the viewable area it shouldn't try and download more data. That defeats the purpose of an offline map. This is especially critical with tourism apps where users may be on a __very__ limited/expensive data plan while roaming.

boundsj:
Noting that https://github.com/mapbox/mapbox-gl-native/pull/8125 proposes a way to turn off file / network activity in the SDK. That PR intends to use the (private to the Mapbox SDK) feature to automatically disable network and file system activity when the host app is running in the background. However, something like this could be used by the application developer to disable activity as a response to connectivity changes (or any other reason) if it were exposed publicly.

jfirebaugh:
As written, the `pause`/`resume` mechanism is not a good fit for this use case though, because it disables loading any resources, including those that can be satisfied by the offline database.

robmaceachern:
@jfirebaugh is there any solution for the use case mentioned by @barnaclebarnes?

We also need a way to allow users to download maps offline and then disable all (or virtually all) mapping network traffic to avoid unnecessary data charges for users.

It'd be great if a delegate method was exposed to enable control over when network requests were attempted.

barnaclebarnes:
@robmaceachern We haven't found one yet. This is still stopping us moving 100% too Mapbox maps unfortunately.

jfirebaugh:
The use case described by @barnaclebarnes is what this feature request is tracking. I've updated the title to make that clearer.

barnaclebarnes:
This issue is still live. AFAIK there is still no way to have offline maps that don't try and access the network. /c @jfirebaugh

bwandersOA:
[...]

Issue #6205, opened 2016-08-30
Crash mapbox::sqlite::Exception authorisation denied on iOS with Data Protection enabled


macteo:
**Platform: iOS**
**Mapbox SDK version: 3.3.4**
### Steps to trigger behavior
1. Integrate MapBox (through CocoaPods).
2. Enable background location updates with `CoreLocation`.
3. Enable [Data Protection](https://developer.apple.com/library/ios/documentation/IDEs/Conceptual/AppDistributionGuide/AddingCapabilities/AddingCapabilities.html).
4. Create a controller with a `MGLMapView` instance visible that appears as soon as you launch the app.
### Expected behaviour

The app will be relaunched in background by the system as soon as a location update is available. The device can be locked and the data will be encrypted.
The app should be able to run anyway in background.
### Actual behaviour

The app crashes as with a `mapbox::sqlite::Exception` _authorization denied_.
I suppose that it is caused by MapBox trying to access a sqlite database that is present but not accessible as Data Protection kicks in.

If you really need to be always able to access the database, you should consider to create the file with no file protection specifying `FileProtectionType.none`:

``` swift
FileManager.default.createFile(atPath: path, contents: data, attributes: [FileAttributeKey.protectionKey.rawValue: FileProtectionType.none])
```

It would even better if you check if the file is available and use it only if the app is in foreground, so its content stays encrypted.

[mapbox-crashlog.txt](https://github.com/mapbox/mapbox-gl-native/files/444169/mapbox-crashlog.txt)

1ec5:
The SQLite database is created in C++ code that the Objective-C code has no control over, but maybe we can set the attribute on the directory containing it like we do for [excluding the database from backups](https://github.com/mapbox/mapbox-gl-native/blob/5588935412fbf6c62f88d92f4685e517a7bb84c1/platform/darwin/src/MGLOfflineStorage.mm#L80).

jfirebaugh:
Previously: #4841.

macteo:
I've got another crash, with different exception, but I think the cause is the same.

[mapbox-crashlog-2.txt](https://github.com/mapbox/mapbox-gl-native/files/450231/mapbox-crashlog-2.txt)

macteo:
I'm not interacting with the map or other MapBox APIs while the app is in background. 
Do you think that removing the map from it's superview before entering background can solve the issue?
Or there still can be some other orphaned object (a singleton maybe) that will cause the issue?

[...]

PR #8125, opened 2017-02-20
Introduce revalidation state API on default file source

boundsj:
Implementation proposal for https://github.com/mapbox/mapbox-gl-native/issues/5236 to address https://github.com/mapbox/mapbox-gl-native/issues/4841
 
This adds an API to default file source so that it can be put in a state where requests are not checked for invalidation in sqlite (or sent over the network). The iOS map view has also been updated to use this new API when it knows that the host application is going into the background or will otherwise be in a state where it is not visible by the user.

This optimization means that the SDK will not make requests for data that the user cannot see or hit edge cases on some OSs where resources (i.e. sqlite) are not available when the host app is not visible to the user.

I did not go as far down as the suggestion in https://github.com/mapbox/mapbox-gl-native/issues/5236#issuecomment-224014020. I think some notion of this API on the file source will still make sense even if we do add a similar API at the thread level.

To test this on an iOS device, I ran a primary test app with these changes in the background and then a secondary app so that I could simulate location changes in Xcode. Before the changes, if the primary app was launched in the background due to location changes, there were sqliite revalidation hits. This does not happen after these changes.

This proposed implementation keeps a file source "active" by default so no platform changes are required and it is completely backwards compatible. However, the Android platform might be able to take advantage of the new API in the MapView activity lifecycle methods.

cc @jfirebaugh @1ec5 @tmpsantos


platform/ios/src/MGLMapView.mm
```diff
@@ -1070,6 +1070,8 @@ - (void)sleepGL:(__unused NSNotification *)notification
     {
         self.dormant = YES;
 
+        [MGLOfflineStorage sharedOfflineStorage].mbglFileSource->setRevalidationState(mbgl::DefaultFileSourceRevalidationState::Inactive);
```
1ec5: Since there may be multiple MGLMapViews but only one MGLOfflineStorage, can we make MGLOfflineStorage observe these notifications to set its own revalidation state?

1ec5: That would also allow iOS and macOS to share the same implementation. For example, the iOS implementation of MGLMapView calls `-sleepGL:` in response to `UIApplicationDidEnterBackgroundNotification`, which probably corresponds to `NSApplicationDidHideNotification` and/or `NSApplicationDidChangeOcclusionStateNotification` on the Mac.

jfirebaugh:
Adding `DefaultFileSource::setRevalidationState` is a start, but this implementation doesn't go deep enough into the `FileSource` stack to pause revalidation and thus fix what's likely the underlying cause of crashes of the type seen in #4841 and #6205. Revalidation per cache-control headers is driven by [timers in `OnlineFileRequest::schedule`](https://github.com/mapbox/mapbox-gl-native/blob/657c9f92cb697b9ab927a14cde6d5cb77c04f1c1/platform/default/online_file_source.cpp#L284-L305), not by calls to `FileSource::request`.

tmpsantos:
@jfirebaugh is right. The tiles we already see loaded/rendered on the screen will re-request when they expire.

I keep my suggestion on making a low level `Thread::pause` so we could pause on a thread synchronization primitive instead of sleeping on the main loop. This has the nice side effect of when we do `Thread::unpause` and go back to the main loop, the expired timers will be processed automatically.

boundsj:
Thanks @jfirebaugh @tmpsantos. I was focused on the case I was seeing while testing where the initial style request that happens at initialization when a host iOS app is launched in the background triggers the interaction with sqlite in DefaultFileSource. 

I do think it is possible that some/many of the `SqlLite disk I/O error` errors that have been reported are because of the db calls in that scenario. The scheduled network calls in the timer may happen in iOS applications that are running for several seconds in the background from time to time (in response to push notifications, location updates, etc.), but I'm seeing the DefaultFileSource sqlite interactions happen _every_ time an app with a Mapbox map is launched.

I do agree we should solve for the schedule case too. I will look again at the `Thread::pause` approach that has been proposed.

jfirebaugh:
I'm not sure that a `Thread::pause` approach will work. That would me that if you pause the FileSource thread, and then call methods like `DefaultFileSource::setAccessToken` that use `invokeSync`, the application will deadlock.

tmpsantos:
> DefaultFileSource::setAccessToken

`invokeSync` could unpause the Thread if this use case is supported. `DefaultFileSource::setAccessToken` in fact could by async since the messages to the thread are ordered.

platform/ios/src/MGLMapView.mm
boundsj: 👍  I'll come back to this once the core implementation is done.

boundsj:
@jfirebaugh @tmpsantos as a checkpoint, I committed https://github.com/mapbox/mapbox-gl-native/pull/8125/commits/60bc871331150d538bf9a316f336c8f83de1dffb that removes references to tasks when the revalidation state is set to inactive. This removes tasks for any loaded tiles that were scheduled and avoids the revalidation noted in https://github.com/mapbox/mapbox-gl-native/pull/8125#issuecomment-281200764. 

I also committed https://github.com/mapbox/mapbox-gl-native/pull/8125/commits/a311ebe796d2aaedbc69c7845ce75d55786c4eff that saves the objects required to make the tasks that https://github.com/mapbox/mapbox-gl-native/pull/8125/commits/60bc871331150d538bf9a316f336c8f83de1dffb removes. Those objects are re-requested if the revalidation state is set to active so that tile expiry still works as before.

Removing all of this in favor of simply pausing the thread would be nice. I'll keep investigating that since I'm still not very familiar with the thread / run loop / scheduler implementation. If it turns out there are issues with pausing the thread, maybe we can fall back to something like https://github.com/mapbox/mapbox-gl-native/pull/8125/commits/60bc871331150d538bf9a316f336c8f83de1dffb and https://github.com/mapbox/mapbox-gl-native/pull/8125/commits/a311ebe796d2aaedbc69c7845ce75d55786c4eff?

tmpsantos:
I wrote a prototype of what I had in mind:
https://github.com/mapbox/mapbox-gl-native/tree/tmpsantos-pause_resume

To test with the GLFW app, just press `P` for pause/resume the FileSource thread.

boundsj:
Thanks so much @tmpsantos for the example https://github.com/mapbox/mapbox-gl-native/pull/8125#issuecomment-281718300 with C++ promises and futures. It seems to work well.

@jfirebaugh is correct that this approach does lock the calling (main) thread when other paths that use the file source thread's `invokeSync` are called after the thread is stopped to wait for the `resume` promise to be set. 

However, since we expect that pause would only be called by our iOS (and maybe Android) platform implementations when an application is in the background and not visible to the user then this may not actually matter most of the time. Also the file source APIs that would cause a deadlock are probably very unlikely to ever be called when the file source thread is paused in the background. 

Still though, in our iOS SDK, there is nothing stopping a developer from calling platform methods that, in turn, call `DefaultFileSource::setOfflineMapboxTileCountLimit`, `DefaultFileSource::getAccessToken`, `DefaultFileSource::setAccessToken`, `DefaultFileSource::getAPIBaseURL`, and `DefaultFileSource::setAPIBaseURL` when the application is running in the background. An iOS app developer who expects their application to be able to set or read one of those values in the background and then continue to do work would be disappointed if we change the thread behavior as proposed here.

I think the most common scenario (and it probably would not be _that_ common) would be for the app developer to set / change the base URL and access token in the background after the file source thread is paused. In https://github.com/mapbox/mapbox-gl-native/pull/8125/commits/3a682d8874762bd46ad9ed1f6415cbfe86dafab2, I went ahead and made these operations async (as discussed in https://github.com/mapbox/mapbox-gl-native/pull/8125#issuecomment-281459533) to mitigate this part of the deadlock concern. For the setters, I don't think that temporarily resuming the thread in the background would be viable since it may lead to processing of any expired tasks and revalidation (and the potential sqlite crash), right?

Do others think that this approach with the modifications to `Thread` is the way we should go? Because the implementation is so much cleaner (and I can't think of a way other than something like a queue that I described in https://github.com/mapbox/mapbox-gl-native/pull/8125#issuecomment-281530135), I think we should go with the `Thread::pause` approach. I'm not sure if this means we would need to document that calling certain APIs (I think it would probably just end up being `getAccessToken` and `getAPIBaseURL`) in the background can lock the main thread.

cc @jfirebaugh @1ec5

tmpsantos:
> I think it would probably just end up being getAccessToken and getAPIBaseURL) in the background can lock the main thread.

You can still make this work. You could cache in the DefaultFileSource the values of the AccessToken and BaseURL when you set them (an initialize with the same default as you initialize them in the main thread). So when you get, you don't need to fetch it from the DefaultFileSource thread.

platform/ios/src/MGLMapView.mm
```diff
@@ -1070,6 +1070,8 @@ - (void)sleepGL:(__unused NSNotification *)notification
     {
         self.dormant = YES;
 
+        [MGLOfflineStorage sharedOfflineStorage].mbglFileSource->pause();
```
1ec5: Just a reminder to move this code into MGLOfflineStorage itself if possible: https://github.com/mapbox/mapbox-gl-native/pull/8125#discussion_r102102611.

tmpsantos:
Please, do not forget to add the utests for `Thread::pause/resume`. I can help if needed.

platform/ios/src/MGLMapView.mm
boundsj: This is done in c11da3a

boundsj: Or rather, https://github.com/mapbox/mapbox-gl-native/pull/8125/commits/57a6b0f922c96989354f4e93774965412fe25db9

test/util/thread.test.cpp
```diff
@@ -129,6 +129,34 @@ TEST(Thread, context) {
...
+        requests.push_back(thread.invokeWithCallback(&TestObject::fn2, [&] (int result) {
+            EXPECT_EQ(tid, std::this_thread::get_id());
+            EXPECT_EQ(result, 1);
+            loop.stop();
+        }));
+
+        requests.push_back(unpausedThread.invokeWithCallback(&TestObject::transferOut, [&] (std::unique_ptr<int> result) {
+            EXPECT_EQ(tid, std::this_thread::get_id());
+            EXPECT_EQ(*result, 1);
+            thread.resume();
```
tmpsantos: I wrote a few more unit tests on my branch, turned out I found some bugs. The thread will hang if it gets destroyed when paused. I fixed that and made a PR. What about we land #8194 and you rebase your branch on top?

boundsj: Sounds good @tmpsantos

boundsj: @tmpsantos I just pushed a wip commit here that duplicates some of what you did in https://github.com/mapbox/mapbox-gl-native/pull/8194 so that I could make sure I understood what you meant about calling resume on a different thread than the one that owns the thread object. I see now that my test still passes because the work task impl `packageArgumentsAndCallback` calls the lambdas such that, in my test, the callback is called on the calling (main in this case) thread so the `MBGL_VERIFY_THREAD` check passes.

Anyway, https://github.com/mapbox/mapbox-gl-native/pull/8194 looks good. I will rebase this on top of that once it is merged. Thanks again.

tmpsantos: My apologies, I was thinking of `Thread::invoke`. You are right, the callback in this case will fire on the thread that invoked the method.

boundsj:
@tmpsantos I've rebased the changes in this PR to include your work in https://github.com/mapbox/mapbox-gl-native/pull/8194. Can you please take another look at this when you are able?

include/mbgl/storage/default_file_source.hpp
```diff
@@ -127,6 +143,8 @@ class DefaultFileSource : public FileSource {
     const std::unique_ptr<util::Thread<Impl>> thread;
     const std::unique_ptr<FileSource> assetFileSource;
     const std::unique_ptr<FileSource> localFileSource;
+    std::string cachedBaseURL;
```
tmpsantos: You probably need to initialize this one with `mbgl::util::API_BASE_URL`, otherwise if you call the getter before setting, the result will be inconsistent.

boundsj: Thanks. I rewrote e277b6e to add this initialization.

include/mbgl/storage/default_file_source.hpp
```diff
@@ -115,7 +115,23 @@ class DefaultFileSource : public FileSource {
      */
     void setOfflineMapboxTileCountLimit(uint64_t) const;
 
+    /*
+     * Pause file request activity.
+     *
+     * If pause is called then no revalidation or network request activity
+     * will occur.
+     *
+     * Note: Calling pause and then calling getAPIBaseURL or getAccessToken
```
tmpsantos: This is no longer valid.

boundsj:
Noting that this solution ended up being sort of specific to the iOS platform because it completely shuts down the Mapbox iOS SDK's map related file system and network activity in core when the host application goes into the background. This will need to be revisited when https://github.com/mapbox/mapbox-gl-native/issues/4291 is addressed.

This solution is not yet granular enough for issues like https://github.com/mapbox/mapbox-gl-native/issues/7443 or to be used on Android where things like background transfers of tiles for offline use are already supported. For now, this is probably ok since the sqlite crashes this solution intends to address are specific to the iOS platform.

include/mbgl/storage/default_file_source.hpp
boundsj: fyi this was fixed in https://github.com/mapbox/mapbox-gl-native/pull/8465 @tmpsantos

robertodias180:
any news on this?

PR #8129, opened 2017-02-20
[ios] Set project deployment target to iOS 8

friedbunny:
With Xcode 8 having dropped support for iOS 7 and Apple [declaring it dangerously insecure](https://www.washingtonpost.com/opinions/apple-vp-the-fbi-wants-to-roll-back-safeguards-that-keep-us-a-step-ahead-of-criminals/2016/03/06/cceb0622-e3d1-11e5-a6f3-21ccdbc5f74e_story.html?utm_term=.094143f73ef0), it’s come time to move our minimum deployment version to [iOS 8](https://developer.apple.com/library/content/releasenotes/General/WhatsNewIniOS/Articles/iOS8.html).

### Notes
- The static framework build infrastructure still exists, but for how much longer is unclear.
- The dynamic framework has been our recommended path since Mapbox iOS SDK v3.1.0 (released February 2016) and developers should migrate to it as soon as possible, if they haven’t already.
- UIActionSheet and UIAlertView remain in use for the attribution button, despite being deprecated in favor of UIAlertController. UIAlertController is a view controller (as opposed to views, previously), which presents a variety of issues to us as framework vendors that we’ll need to consider before adoption.
- There is no change to the hardware we support — we already explicitly did not support iPhone 4, fourth-generation iPod Touch, or the first-generation iPad.

/cc @mapbox/ios


1ec5:
> The static framework build infrastructure still exists, but for how much longer is unclear.

To be clear, we currently have no plans to drop the static framework and will provide advance notice before doing so.

platform/ios/INSTALL.md
```diff
@@ -52,8 +52,6 @@ bash "${BUILT_PRODUCTS_DIR}/${FRAMEWORKS_FOLDER_PATH}/Mapbox.framework/strip-fra
 
 #### Static framework
 
-If your application targets iOS 7.x, you’ll need to install the static framework instead:
```
1ec5: I'd leave in a sentence just to make it clear that these are alternative instructions, not additive. "If you need a static framework…"

friedbunny: I struggled to come up with anything here — “if you can’t say anything nice...” 😉

Will revisit this tomorrow. Very open to detailed suggestions!

1ec5: Here’s the “You know who you are” version:

> You can alternatively install the SDK as a static framework:

friedbunny: Did this in a1c477b.

PR #8132, opened 2017-02-21
[android] - rename simple example map view 

tobrun:
We currently have two "simple" mapview examples in our project. One for the Phone module and one for the wear module. The issue I'm constantly hitting is trying to do some tests on the phone module but accidentally changing the code for the wear module instead. Renaming the wear example to explicitly state Wear in it's name will resolve this.


tobrun:
in 3a6f6e9 I'm also sneaking in ignoring a test that I forgot in #8063.

PR #8135, opened 2017-02-21
Add .idea, .mason and cmake-build-debug folders to gitignore

Guardiola31337:
`.idea`, `.mason` and `cmake-build-debug` folders weren't included in `.gitignore`.

Cc/ @kkaefer

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -33,3 +33,5 @@ xcuserdata
 /documentation
 test/fixtures/api/assets.zip
 test/fixtures/storage/assets.zip
+/.idea
+/cmake-build-debug


.gitignore
```diff
@@ -33,3 +33,6 @@ xcuserdata
 /documentation
 test/fixtures/api/assets.zip
 test/fixtures/storage/assets.zip
+/.idea
+/cmake-build-debug
```
kkaefer: This is not a folder that is created by default. Where is it from?

.gitignore
```diff
@@ -33,3 +33,6 @@ xcuserdata
...
+/.mason
```
kkaefer: You can just delete that folder locally. It's not needed anymore.

.gitignore
Guardiola31337: Created by CLion when I loaded Mapbox GL Native from it.

kkaefer: If it's something that you only see locally due to the use of your toolchain, you can also add these folders in a system-wide .gitignore file:

`~/.gitconfig`
```
[core]
	excludesfile = /Users/USERNAME/.gitignore
```

Guardiola31337: Potential devs who use same toolchain might not know this and will find same problem.

Guardiola31337:
As @kkaefer suggested we can't possibly exclude all temp files created by all setups and those folders should be managed locally. Closing it.

PR #8139, opened 2017-02-21
[ios] MGLAnnotationView annotation property made writable

### LINKED ISSUE #7321: Why is annotation a read-only property on annotation view?
**Platform:** iOS

I don't understand the motivation behind making the `annotation` property a read-only in `MGLAnnotationView` class. Apple's MapKit provides an annotation view's initialiser that accepts both the reuse identifier and annotation. This moves the view setup into the view class and doesn't pollute the delegate method with view setup code. Do mapbox developers imply that the view setup should be done in `mapView:viewForAnnotation:` delegate's method?

fabian-guerra:
Fixes #7321

### DIFF
--- a/platform/ios/src/MGLAnnotationView.h
+++ b/platform/ios/src/MGLAnnotationView.h
@@ -89,7 +89,7 @@ typedef NS_ENUM(NSUInteger, MGLAnnotationViewDragState) {
  contains a non-`nil` value while the annotation view is visible on the map. If
  the view is queued, waiting to be reused, the value is `nil`.
  */
-@property (nonatomic, readonly, nullable) id <MGLAnnotation> annotation;
+@property (nonatomic, nullable) id <MGLAnnotation> annotation;
 
 /**
  The string that identifies that this annotation view is reusable.
--- a/platform/ios/src/MGLAnnotationView.mm
+++ b/platform/ios/src/MGLAnnotationView.mm
@@ -10,7 +10,6 @@
 @interface MGLAnnotationView () <UIGestureRecognizerDelegate>
 
 @property (nonatomic, readwrite, nullable) NSString *reuseIdentifier;
-@property (nonatomic, readwrite, nullable) id <MGLAnnotation> annotation;
 @property (nonatomic, readwrite) CATransform3D lastAppliedScaleTransform;
 @property (nonatomic, weak) UIPanGestureRecognizer *panGestureRecognizer;
 @property (nonatomic, weak) UILongPressGestureRecognizer *longPressRecognizer;
@@ -240,6 +239,7 @@
     }
     else if (dragState == MGLAnnotationViewDragStateCanceling)
     {
+        NSAssert(self.annotation, @"Annotation property should not be nil.");
         self.panGestureRecognizer.enabled = NO;
         self.longPressRecognizer.enabled = NO;
         self.center = [self.mapView convertCoordinate:self.annotation.coordinate toPointToView:self.mapView];
--- a/platform/ios/src/MGLAnnotationView_Private.h
+++ b/platform/ios/src/MGLAnnotationView_Private.h
@@ -8,7 +8,6 @@ NS_ASSUME_NONNULL_BEGIN
 @interface MGLAnnotationView (Private)
 
 @property (nonatomic, readwrite, nullable) NSString *reuseIdentifier;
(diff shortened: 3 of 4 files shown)


platform/ios/src/MGLAnnotationView.h
```diff
@@ -89,7 +89,7 @@ typedef NS_ENUM(NSUInteger, MGLAnnotationViewDragState) {
  contains a non-`nil` value while the annotation view is visible on the map. If
  the view is queued, waiting to be reused, the value is `nil`.
  */
-@property (nonatomic, readonly, nullable) id <MGLAnnotation> annotation;
+@property (nonatomic, nullable) id <MGLAnnotation> annotation;
```
1ec5: What is the effect of changing this property? Does it accomplish the use case envisioned in #7321, given that MGLAnnotationView’s initializer doesn’t take an MGLAnnotation parameter?

1ec5: Even though `MKAnnotationView.annotation` is read-write, it also comes with this note in the documentation:

> You should not change the value of this property directly.

which we also state above, redundantly. What is our reason for making the property read-write but advising developers not to write to it?

fabian-guerra: If `setAnnotation` exists publicly, then they can override it in a subclass and have the annotation model change the behavior of the view.

fabian-guerra: Changing the annotation in `mapView:viewForAnnotation:` will not affect `annotationViewForAnnotation:` (the caller) it's mean to be a convenient way to set up any UI update code. This is also how Mapkit behaves.
From a conversation with @boundsj 
> The developer setting it in the delegate first is redundant but facilitates them writing code to update the view when the annotation model property is set.

fabian-guerra:
@incanus https://github.com/mapbox/mapbox-gl-native/pull/8139#discussion_r102307572 does this makes sense?

incanus:
@fabian-guerra Yes, but I think we either need to guard against, or at very least warn against, setting the annotation to `nil` since it's technically now possible. Especially if that can result in a crash.

fabian-guerra:
@incanus I'm thinking a `NSAssert` clause would be better since it won't hide the failing reason nor creating unexpected behavior, thoughts?

eimantas:
One more approach would be to create a designated initialiser that accepts the annotation as an argument (just like in MapKit). This would also increase the MapKit parity for the mapbox' darwin SDKs. With this it'd be possible to mark the current initialiser as deprecated and leave it like that for a couple of releases until people who update the SDK, would know to update the initialisers too.

incanus:
Made a minor grammar tweak in https://github.com/mapbox/mapbox-gl-native/pull/8139/commits/590c6cecd7ed4e0462965cea3be96e9d64fd9e40 @fabian-guerra, but otherwise this feels ok to me. I'm unsure how sustainable it will be in the future, but I think that just comes with the territory of this sort of change.

boundsj:
@fabian-guerra please cherry pick this to the new release branch `release-ios-v3.5.0-android-v5.0.0` thanks!

platform/ios/src/MGLAnnotationView.mm
```diff
@@ -240,6 +239,7 @@ - (void)setDragState:(MGLAnnotationViewDragState)dragState animated:(BOOL)animat
     }
     else if (dragState == MGLAnnotationViewDragStateCanceling)
     {
+        NSAssert(self.annotation, @"Annotation property should not be nil.");
```
1ec5: When you want to indicate an error on the developer’s part, raise a named NSException instead of calling `NSAssert()`. See https://github.com/mapbox/mapbox-gl-native/pull/8026#discussion_r100661990 for more details. @fabian-guerra, can you change these assertions into named exceptions?

dfelber:
After upgrading from 3.4.2 to 3.5.0 I'm experiencing crashes due to the newly introduced assertion.

When trying to select an image annotation (provided by `func mapView(_ mapView: MGLMapView, imageFor annotation: MGLAnnotation) -> MGLAnnotationImage?`) the assertion gets triggered.
I assume it is due `func mapView(_ mapView: MGLMapView, viewFor annotation: MGLAnnotation) -> MGLAnnotationView?` returning `nil` for that annotation.

PR #8140, opened 2017-02-21
Use node-cmake 2.x

kkaefer:
- Removes need to run `npm install` before invoking CMake

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -47,12 +47,7 @@ endif
 .PHONY: default
 default: test
 
-.NOTPARALLEL: node_modules
-node_modules: package.json
-	npm install --ignore-scripts # Install dependencies but don't run our own install script.
-
 BUILD_DEPS += Makefile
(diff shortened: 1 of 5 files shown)


kkaefer:
Note that in 2.x, the CMake file should be committed to the project rather than pulled in via node-cmake, according to the documentation.

jfirebaugh:
Does [this line](https://github.com/mapbox/mapbox-gl-native/blob/b49fc44498be14d0a5f989fb811b4603fd80de1a/cmake/test.cmake#L9) need updating?

jfirebaugh:
Travis build failure was #8142. We're down to this issue on Node macOS:

```
▸ Compiling node_feature.cpp

❌  /Users/vagrant/git/platform/node/src/node_feature.hpp:8:10: 'nan.h' file not found

#include <nan.h>
         ^
```

PR #8141, opened 2017-02-21
Improve MGLUserTrackingMode docs and exclude from macOS

captainbarbosa:
This pull request clarifies the documentation around `MGLUserTrackingModeFollowWithHeading` to include more detail. It also excludes the `MGLUserTrackingMode` enum from macOS.

 cc @friedbunny

### DIFF
--- a/platform/darwin/src/MGLTypes.h
+++ b/platform/darwin/src/MGLTypes.h
@@ -44,21 +44,6 @@ typedef NS_ENUM(NSInteger, MGLErrorCode) {
     MGLErrorCodeConnectionFailed = 3,
 };
 
-/**
- The mode used to track the user location on the map. Used with
- `MGLMapView.userTrackingMode`.
- */
-typedef NS_ENUM(NSUInteger, MGLUserTrackingMode) {
-    /** The map does not follow the user location. */
-    MGLUserTrackingModeNone              = 0,
-    /** The map follows the user location. */
-    MGLUserTrackingModeFollow,
-    /** The map follows the user location and rotates when the heading changes. */
-    MGLUserTrackingModeFollowWithHeading,
-    /** The map follows the user location and rotates when the course changes. */
-    MGLUserTrackingModeFollowWithCourse,
-};
-
 /** Options for enabling debugging features in an `MGLMapView` instance. */
 typedef NS_OPTIONS(NSUInteger, MGLMapDebugMaskOptions) {
     /** Edges of tile boundaries are shown as thick, red lines to help diagnose
--- a/platform/ios/src/MGLMapView.h
+++ b/platform/ios/src/MGLMapView.h
@@ -51,11 +51,30 @@ typedef NS_ENUM(NSUInteger, MGLAnnotationVerticalAlignment) {
 typedef NS_ENUM(NSUInteger, MGLUserTrackingMode) {
     /** The map does not follow the user location. */
     MGLUserTrackingModeNone              = 0,
-    /** The map follows the user location. */
+    /** The map follows the user location. This tracking mode falls back
+        to `MGLUserTrackingModeNone` if the user pans the map view. */
     MGLUserTrackingModeFollow,
-    /** The map follows the user location and rotates when the heading changes. */
+    /**
+        The map follows the user location and rotates when the heading changes.
+        The default user location annotation displays a fan-shaped indicator with
+        the current heading. The heading indicator represents the direction the
+        device is facing, which is sized according to the reported accuracy.
+     
+        This tracking mode is disabled if the user pans the map view, but
+        remains enabled if the user zooms in. If the user rotates the map
+        view, this tracking mode will fall back to `MGLUserTrackingModeFollow`.
+     */
     MGLUserTrackingModeFollowWithHeading,
-    /** The map follows the user location and rotates when the course changes. */
+    /**
+        The map follows the user location and rotates when the course changes.
+        Course represents the direction in which the device is traveling.
+        The default user location annotation shows a puck-shaped indicator
+        that rotates as the course changes.
+     
+        This tracking mode is disabled if the user pans the map view, but
+        remains enabled if the user zooms in. If the user rotates the map view,
+        this tracking mode will fall back to `MGLUserTrackingModeFollow`.
+     */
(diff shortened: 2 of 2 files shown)


platform/darwin/src/MGLTypes.h
```diff
@@ -53,11 +54,14 @@ typedef NS_ENUM(NSUInteger, MGLUserTrackingMode) {
     MGLUserTrackingModeNone              = 0,
     /** The map follows the user location. */
     MGLUserTrackingModeFollow,
-    /** The map follows the user location and rotates when the heading changes. */
+    /** The map follows the user location and rotates when the heading changes. 
+        The default user annotation shows an indicator  with the current heading,
```
friedbunny: There’s an extra space here after “indicator”.

platform/darwin/src/MGLTypes.h
```diff
@@ -53,11 +54,14 @@ typedef NS_ENUM(NSUInteger, MGLUserTrackingMode) {
     MGLUserTrackingModeNone              = 0,
     /** The map follows the user location. */
     MGLUserTrackingModeFollow,
-    /** The map follows the user location and rotates when the heading changes. */
+    /** The map follows the user location and rotates when the heading changes. 
+        The default user annotation shows an indicator  with the current heading, 
+        sized according to the reported accuracy.  */
```
friedbunny: Nit: there only needs to be a single space between the period and the end-comment stuff.

platform/darwin/src/MGLTypes.h
```diff
@@ -44,6 +44,7 @@ typedef NS_ENUM(NSInteger, MGLErrorCode) {
     MGLErrorCodeConnectionFailed = 3,
 };
 
+#if TARGET_OS_IPHONE
```
1ec5: Let’s move this declaration to platform/ios/src/MGLMapView.h, where it’s actually used.

platform/darwin/src/MGLTypes.h
```diff
@@ -53,11 +54,14 @@ typedef NS_ENUM(NSUInteger, MGLUserTrackingMode) {
     MGLUserTrackingModeNone              = 0,
     /** The map follows the user location. */
     MGLUserTrackingModeFollow,
-    /** The map follows the user location and rotates when the heading changes. */
+    /** The map follows the user location and rotates when the heading changes. 
+        The default user annotation shows an indicator  with the current heading, 
+        sized according to the reported accuracy.  */
     MGLUserTrackingModeFollowWithHeading,
     /** The map follows the user location and rotates when the course changes. */
```
1ec5: Can you describe the user puck here?

platform/darwin/src/MGLTypes.h
1ec5: s/user annotation/user location annotation/

friedbunny:
@1ec5 @captainbarbosa I think we should also document the criteria for automatic disengagement of these modes. That’s been a fairly consistent point of confusion — “why does the heading indicator go away when I pan the map?” and such.

platform/ios/src/MGLMapView.h
```diff
@@ -57,7 +57,9 @@ typedef NS_ENUM(NSUInteger, MGLUserTrackingMode) {
         The default user location annotation shows a fan-shaped indicator with 
         the current heading, sized according to the reported accuracy. */
     MGLUserTrackingModeFollowWithHeading,
-    /** The map follows the user location and rotates when the course changes. */
+    /** The map follows the user location and rotates when the course changes. 
+        The default user location annotation shows a puck-shaped indicator
+        that rotates as the user direction changes */
```
friedbunny: > rotates as the user _course_ changes

(It also needs a period at the end.)

platform/ios/src/MGLMapView.h
```diff
@@ -45,6 +45,31 @@ typedef NS_ENUM(NSUInteger, MGLAnnotationVerticalAlignment) {
 };
 
 /**
+ The mode used to track the user location on the map. Used with
+ `MGLMapView.userTrackingMode`.
+ */
+typedef NS_ENUM(NSUInteger, MGLUserTrackingMode) {
+    /** The map does not follow the user location. */
+    MGLUserTrackingModeNone              = 0,
+    /** The map follows the user location. */
```
1ec5: Panning also disables `MGLUserTrackingModeFollow`, reverting to `MGLUserTrackingModeNone`.

platform/ios/src/MGLMapView.h
```diff
@@ -45,6 +45,31 @@ typedef NS_ENUM(NSUInteger, MGLAnnotationVerticalAlignment) {
...
+        tracking mode is disabled if the user pans the map view, but remains
+        unchanged if the user zooms in. If the user rotates the map view, this
+        tracking mode will fall back to `MGLUserTrackingModeFollow`. */
+    MGLUserTrackingModeFollowWithHeading,
+    /** The map follows the user location and rotates when the course changes.
+        The default user location annotation shows a puck-shaped indicator
+        that rotates as the user course changes. This tracking mode is
+        disabled if the user pans the map view, but remains if the user zooms
+        in. If the user rotates the map view, this tracking mode will fall back
+        to `MGLUserTrackingModeFollow`. */
```
1ec5: Maybe we should refer the developer to `CLLocation.course`, since the distinction between a heading and a course is often lost on developers.

friedbunny:
@captainbarbosa We talked about adding some paragraph breaks, but hit a jazzy parsing issue. Here’s how to avoid that:

```
    /** 
        The map follows the user location and rotates when the heading changes.
        The default user location annotation shows a fan-shaped indicator with
        the current heading, sized according to the reported accuracy.
     
        This tracking mode is disabled if the user pans the map view, but remains
        unchanged if the user zooms in. If the user rotates the map view, this
        tracking mode will fall back to `MGLUserTrackingModeFollow`.
     */
```

... move the body of the text onto its own lines, away from the comment block opening/closing.

friedbunny:
I put this on the v3.5.0 milestone — I think we can get it in. We should retarget this PR against the release branch and probably rebase to make sure there aren’t any huge conflicts.

captainbarbosa:
@friedbunny Cool, here's a screenshot of the docs:

trkg

PR #8145, opened 2017-02-21
[core] Clean up symbol-related code

jfirebaugh:
Cosmetic stuff: fixing `&` placement, converting long parameter lists to one-parameter-per-line, etc.

### DIFF
--- a/src/mbgl/geometry/anchor.hpp
+++ b/src/mbgl/geometry/anchor.hpp
@@ -6,17 +6,15 @@
 
 namespace mbgl {
 
-struct Anchor {
+class Anchor {
+public:
     Point<float> point;
     float angle = 0.0f;
     float scale = 0.0f;
     int segment = -1;
 
-    explicit Anchor(float x_, float y_, float angle_, float scale_)
-        : point(x_, y_), angle(angle_), scale(scale_) {}
-    explicit Anchor(float x_, float y_, float angle_, float scale_, int segment_)
+    Anchor(float x_, float y_, float angle_, float scale_, int segment_ = -1)
         : point(x_, y_), angle(angle_), scale(scale_), segment(segment_) {}
-
 };
 
 typedef std::vector<Anchor> Anchors;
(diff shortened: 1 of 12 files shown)


src/mbgl/geometry/anchor.hpp
```diff
@@ -6,17 +6,15 @@
...
 
-struct Anchor {
+class Anchor {
+public:
     Point<float> point;
     float angle = 0.0f;
     float scale = 0.0f;
     int segment = -1;
 
-    explicit Anchor(float x_, float y_, float angle_, float scale_)
```
anandthakker: @jfirebaugh what purpose was `explicit` serving here?

src/mbgl/layout/symbol_instance.hpp
```diff
@@ -7,17 +7,27 @@
 
 namespace mbgl {
 
-struct Anchor;
```
anandthakker: Is there a story behind the previous use of `struct` and the preference now for `class`?

src/mbgl/geometry/anchor.hpp
jfirebaugh: Here, no purpose. `explicit` is used to prevent [unintended implicit conversions via a single-argument constructor](http://stackoverflow.com/questions/121162/what-does-the-explicit-keyword-in-c-mean). This isn't a single-argument constructor, so that doesn't apply.

Using `explicit` on a multi-argument constructor does prohibit the type name `Anchor` from being omitted with initializer lists such as:

```
Anchor getAnchor() {
    return { x, y, angle, scale };
}
```

But I'm not aware of any cases where we actually want that behavior.

src/mbgl/layout/symbol_instance.hpp
jfirebaugh: `struct` and `class` are interchangeable, modulo default visibility, but annoyingly, the keyword used for forward declaration must match the keyword used for the definition. Our convention is to always use `class`, so that forward declarations are consistent.

PR #8146, opened 2017-02-21
[core] Move interpolate.hpp to public include directory

jfirebaugh:


jfirebaugh:
Turns out there are a bunch more. Widened the scope and made a new branch: #8147.

PR #8147, opened 2017-02-21
Fix private header includes

### LINKED ISSUE #4647: Non-core libraries should never need -Isrc
There are a few [comments like this](https://github.com/mapbox/mapbox-gl-native/blob/f964e40e7e9220d08751d8607af61ac5a7c0794c/platform/ios/platform.gyp#L51) in the gyp files where non-core helper libraries currently have to be able to `#include` headers that live in `src`. We should figure out how to remove these. Only files in `src` should need to include headers in `src`.

jfirebaugh:
Move some includes to public include directory, avoid others.

Fixes #4647.


cmake/core-files.cmake
```diff
@@ -444,13 +444,16 @@ set(MBGL_CORE_FILES
...
+    include/mbgl/util/ignore.hpp
     include/mbgl/util/image.hpp
+    include/mbgl/util/interpolate.hpp
     include/mbgl/util/logging.hpp
     include/mbgl/util/noncopyable.hpp
     include/mbgl/util/optional.hpp
     include/mbgl/util/platform.hpp
     include/mbgl/util/projection.hpp
     include/mbgl/util/range.hpp
+    include/mbgl/util/rapidjson.hpp
```
kkaefer: We've kept this private because it requires having the RapidJSON headers present. What's the reason for making this public?

platform/node/src/node_map.cpp
```diff
@@ -9,7 +9,7 @@
 #include <mbgl/style/conversion/source.hpp>
 #include <mbgl/style/conversion/layer.hpp>
 #include <mbgl/style/conversion/filter.hpp>
-#include <mbgl/sprite/sprite_image.cpp>
+#include <mbgl/sprite/sprite_image.hpp>
```
kkaefer: How did this ever work without duplicate symbol errors?

cmake/core-files.cmake
```diff
@@ -53,43 +53,43 @@ set(MBGL_CORE_FILES
...
-    src/mbgl/gl/stencil_mode.hpp
-    src/mbgl/gl/texture.hpp
-    src/mbgl/gl/types.hpp
     src/mbgl/gl/uniform.cpp
-    src/mbgl/gl/uniform.hpp
     src/mbgl/gl/value.cpp
-    src/mbgl/gl/value.hpp
     src/mbgl/gl/vertex_array.cpp
-    src/mbgl/gl/vertex_array.hpp
-    src/mbgl/gl/vertex_buffer.hpp
```
kkaefer: Why do we make them public? They only seem to be used within the rendering subsystem, which isn't exposed.

platform/node/src/node_map.cpp
kkaefer: Ah, https://github.com/mapbox/mapbox-gl-native/issues/8148

cmake/core-files.cmake
kkaefer: Ah, https://github.com/mapbox/mapbox-gl-native/issues/8148

platform/node/src/node_map.cpp
tmpsantos: Indeed.

jfirebaugh: Yeah, right? I don't know.

cmake/core-files.cmake
jfirebaugh: It's used by Qt SDK bindings, specifically https://github.com/mapbox/mapbox-gl-native/blob/d3ca6c67f1b8fbf5d71c307e71ba89a483ec75bc/platform/qt/src/qt_geojson.hpp#L5-L5. @tmpsantos @brunoabinader do you see a way to keep it private?

brunoabinader: We could make `Result<GeoJSON> convertGeoJSON(const QVariant& value)` to use an overloaded version of `convertGeoJSON` internally that receives a `std::string` as input. This overloaded version could then use _rapidjson_ to parse this and reside outside of Qt scope.

The same could be made in e.g. https://github.com/mapbox/mapbox-gl-native/blob/master/platform/android/src/style/conversion/geojson.hpp.

brunoabinader: Done in f6f826d0b5d3948422b764272f388d6487559c87.

jfirebaugh:
Nice, thanks @brunoabinader.

Ok, ready for a final review here.

jfirebaugh:
@kkaefer #8148 seems to be stalled. Could we merge this now to avoid future merge conflicts (which are already starting to accumulate)?

kkaefer:
@jfirebaugh please see https://github.com/mapbox/mapbox-gl-native/pull/8216 for an alternative that includes all of the commits in this PR, except for the one making the GL headers public.

jfirebaugh:
⇢ #8216

PR #8166, opened 2017-02-22
Fix runtime style template

### LINKED ISSUE #7659: RuntimeStyle template files do not conform to checkstyle
When updating classes from our Runtime style implementation in #7658 in the related ejs template files, I noticed that our CI job is failing on checkstyle because the ejs template files weren't updated as part of #7442.

cc @cammace

Guardiola31337:
Fixes #7659 

Had to increase file length max allowed to 2118 to conform to checkstyle.

Review @ivovandongen


Guardiola31337:
As @tobrun pointed out [here](https://github.com/mapbox/mapbox-gl-native/issues/7659#issuecomment-281745560) it was fixed when DDS landed.
Closing PR.

PR #8170, opened 2017-02-22
[core] Log Memory.Footprint test results

### LINKED ISSUE #8149: Memory safeguard unit test should be a benchmark
The memory usage test (which is run by the travis Qt jobs) would be even more useful as a benchmark that logged the actual usage even if it's "passing" below the threshold.  Ideally, this could be run in the nightly build and produce a metric similar to https://mapbox.github.io/mapbox-gl-native/metrics/binary-size/

anandthakker:
Even when we're passing this test, it can be very useful to have a log of the memory usage value, so that we can get a sense of the memory impact of new features and also to make sure that it's reasonable when we to increase the thresholds for these checks, when necessary.

Closes #8149  (leaving integration w/ the fancy nightly metrics graph for another day)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -173,7 +173,8 @@ matrix:
         - mapbox_export_mesa_library_path
       script:
         - make qt-app
-        - LD_PRELOAD=/usr/lib/x86_64-linux-gnu/libjemalloc.so make run-qt-test-Memory.*:*.Load
+        - GTEST_OUTPUT=xml LD_PRELOAD=/usr/lib/x86_64-linux-gnu/libjemalloc.so make run-qt-test-Memory.*:*.Load
+        - scripts/log_memory_benchmarks.sh test_detail.xml "Platform=Linux,Compiler=${_CC},Arch=$(uname -m)"
 
     # Qt 5 - GCC 5 - Release
     - os: linux
--- a/test/util/memory.test.cpp
+++ b/test/util/memory.test.cpp
@@ -1,4 +1,5 @@
 #include <mbgl/test/stub_file_source.hpp>
+#include <mbgl/test/getrss.hpp>
 #include <mbgl/test/util.hpp>
 
 #include <mbgl/map/map.hpp>
@@ -20,29 +21,6 @@
 
 using namespace mbgl;
 using namespace std::literals::string_literals;
-
-long getRSS() {
-    auto statm = util::read_file("/proc/self/statm");
-
-    std::vector<std::string> stats;
-    std::istringstream stream(statm);
-
-    std::copy(std::istream_iterator<std::string>(stream),
-        std::istream_iterator<std::string>(),
-        std::back_inserter(stats));
-
-    return std::stol(stats[1]) * getpagesize();
-}
-
-bool isUsingJemalloc() {
-    const char* preload = getenv("LD_PRELOAD");
-
-    if (preload) {
-        return std::string(preload).find("libjemalloc.so") != std::string::npos;
-    } else {
-        return false;
-    }
-}
 
 class MemoryTest {
 public:
@@ -115,10 +93,6 @@ TEST(Memory, Raster) {
 // reasonable limits, so this test acts more like a
 // safeguard.
 TEST(Memory, Footprint) {
-    if (!isUsingJemalloc()) {
-        return;
-    }
-
     MemoryTest test;
 
     auto renderMap = [&](Map& map, const char* style){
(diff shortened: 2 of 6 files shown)


tmpsantos:
Can we hold on merging this? I would like to review it tomorrow with more time.

benchmark/include/mbgl/benchmark.hpp
```diff
@@ -4,4 +4,5 @@ namespace mbgl {
 
 int runBenchmark(int argc, char* argv[]);
 
+
```
tmpsantos: Not needed.

benchmark/memory/memory.benchmark.cpp
```diff
@@ -0,0 +1,255 @@
...
+#include <string>
+#include <unordered_map>
+#include <utility>
+#include <sstream>
+
+#include <stdlib.h>
+#include <unistd.h>
+
+/*
+ * Begin getRSS.c
```
tmpsantos: Can we move this to `benchmark/src/mbgl/benchmark/getrss.c`? Maybe add the function prototypes to `utils.hpp`.

benchmark/include/mbgl/memory.benchmark.hpp
```diff
@@ -0,0 +1,7 @@
+#pragma once
+
+namespace mbgl {
+namespace benchmark {
+    bool memoryFootprintBenchmark();
```
tmpsantos: nit, we don't indent inside the namespace

.travis.yml
```diff
@@ -173,7 +173,8 @@ matrix:
         - mapbox_export_mesa_library_path
       script:
         - make qt-app
-        - LD_PRELOAD=/usr/lib/x86_64-linux-gnu/libjemalloc.so make run-qt-test-Memory.*:*.Load
+        - LD_PRELOAD=/usr/lib/x86_64-linux-gnu/libjemalloc.so make run-qt-test-*.Load
```
tmpsantos: Let's keep the other memory tests.

benchmark/memory/memory.benchmark.cpp
```diff
@@ -0,0 +1,255 @@
...
+private:
+    Response response(const std::string& path) {
+        Response result;
+
+        auto it = cache.find(path);
+        if (it != cache.end()) {
+            result.data = it->second;
+        } else {
+            auto data = std::make_shared<std::string>(
+                util::read_file(std::string("test/fixtures/resources/") + path));
```
tmpsantos: Copy the test resources used by this test to `benchmark/fixtures/`.

benchmark/memory/memory.benchmark.cpp
```diff
@@ -0,0 +1,255 @@
...
+#else
+	/* AIX, BSD, Solaris, and Unknown OS ------------------------ */
+	return (size_t)0L;			/* Unsupported. */
+#endif
+}
+
+/** End getRSS.c */
+
+
+using namespace mbgl;
```
tmpsantos: No needed, you are already using `namespace mbgl {`

test/util/memory.test.cpp
```diff
@@ -1,169 +0,0 @@
...
-            return "raster";
-        } else {
-            return "vector";
-        }
-    };
-
-    std::unordered_map<std::string, std::shared_ptr<std::string>> cache;
-};
-
-TEST(Memory, Vector) {
```
tmpsantos: Do not remove this. These (Memory.Vector and Memory.Raster) are the only tests we have that do a full pass rendering inside valgrind looking for memory leaks.

anandthakker:
Thanks for the review @tmpsantos 

> Why not keep it on gtest? I think for this particular case it is even more prepared for having a XML generated that we could use to plot a graph:

> https://github.com/google/googletest/blob/master/googletest/docs/AdvancedGuide.md#logging-additional-information

Somehow in my attempts to get gtest to simply log additional info to the screen, I missed this.  You're right: if we can use gtest's existing functionality, then I should rework this PR to:
1. Keep the more portable version of getRSS (it's convenient to be able to at least run the memory test on non-linux box, even if it's not as reliable without jemalloc)
2. Use gtest's `RecordProperty` feature to record the memory metrics to XML output
3. Add a simple travis step that greps (or whatever) for these values from the XML results, so that we can see them in the travis logs.  Later, when we have a chance, we can add this to the nightly metrics charts.

tmpsantos:
> Keep the more portable version of getRSS (it's convenient to be able to at least run the memory test on non-linux box, even if it's not as reliable without jemalloc)

:+1:

> Use gtest's RecordProperty feature to record the memory metrics to XML output
> Add a simple travis step that greps (or whatever) for these values from the XML results, so that we can see them in the travis logs. Later, when we have a chance, we can add this to the nightly metrics charts.

This would be an amazing follow-up. There was prior work done here for binary size:
https://github.com/mapbox/mapbox-gl-native/pull/7005

Metrics can be seen here:
https://mapbox.github.io/mapbox-gl-native/metrics/binary-size/

anandthakker:
@tmpsantos updated to output results from gtest as we discussed, and laid the groundwork for reporting / visualizing the benchmark over time in `log_memory_benchmarks.cpp`

It appears there isn't currently a nightly build for `qt`, and I think posting for _every_ build would be overkill.  Should we add a `nightly-release` workflow to the qt bitrise setup?

Issue #8150, opened 2017-02-21
Android Wear support 5.0.0


cjkiller:
Hi, I'm trying to create an Android Wear map app. Using mapbox 5.0.0 All working fine on Nixon Mission, Moto 360 but is crashing out on the Sony smartwatch 3. Renders fine with 4.2.2. I've noticed the Sony smartwatch 3 has a lot less opengl es extensions than the others, maybe thats important.

**Platform:** Android (wear)
**Mapbox SDK version:** 5.0.0 beta and snapshot

### Steps to trigger behavior

 1. Show a map

### Expected behavior
Map renders
### Actual behavior
Crashes out with 
E/mbgl: {.trackwear.maps}[Shader]: Program failed to link: ERROR:LINK-4 (line -1) Too many attribute values
E/AndroidRuntime: FATAL EXCEPTION: main
                                                 Process: io.trackwear.maps, PID: 3416
                                                 java.lang.Error: program failed to link
                                                     at com.mapbox.mapboxsdk.maps.NativeMapView.nativeRender(Native Method)
                                                     at com.mapbox.mapboxsdk.maps.NativeMapView.render(NativeMapView.java:175)
                                                     at com.mapbox.mapboxsdk.maps.MapView.onDraw(MapView.java:444)
                                                     at android.view.View.draw(View.java:16184)

marksev1:
+1, please fix this, there are a lot of SW3 users that would like to use your mapping solution in the Wear apps.

marksev1:
Should this issue be fixed in the beta2 already?

niedzielski:
One of our users is reporting a similar issue on his phone, a Samsung GT-S7582 with CyanogenMod 4.4.4 (API 19). This is on Mapbox v5.0.0-beta.2:

```
03-07 22:33:06.582 11630-11630/org.wikipedia.alpha D/HAWAII_EGL: eglCreateContext() config: 13 context: 0x52a21c98, VC context 3, Thread 11630
03-07 22:33:06.642 11630-11630/org.wikipedia.alpha D/HAWAII_EGL: eglMakeCurrent(NULL) Thread: 11630
03-07 22:33:06.642 11630-11630/org.wikipedia.alpha D/HAWAII_EGL: eglDestroySurface() surface: 0x40063f00, android window 0x506a22a0, Thread: 11630
[...]

PR #8177, opened 2017-02-22
Log GL_MAX_VERTEX_ATTRIBS value before making programs

anandthakker:
Temporary diagnostic logging for https://github.com/mapbox/mapbox-gl-native/issues/8150

### DIFF
--- a/src/mbgl/gl/context.cpp
+++ b/src/mbgl/gl/context.cpp
@@ -41,8 +41,8 @@ Context::~Context() {
     
 void Context::logGlMaxVertexAttribsValue() {
     int glMaxVertexAttribs;
-    glGetIntegerv(GL_MAX_VERTEX_ATTRIBS, &glMaxVertexAttribs);
-    std::cout << "GL_MAX_VERTEX_ATTRIBS = " << glMaxVertexAttribs << std::endl;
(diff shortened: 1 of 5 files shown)


anandthakker:
@zugaldia still working on getting my Android build working, but I confirmed that it logs `GL_MAX_VERTEX_ATTRIBS = 16` to the console when I run the glfw test app.

src/mbgl/gl/context.cpp
```diff
@@ -37,10 +38,16 @@ static_assert(underlying_type(TextureFormat::Alpha) == GL_ALPHA, "OpenGL type mi
 Context::~Context() {
     reset();
 }
+    
+void Context::logGlMaxVertexAttribsValue() {
+    int glMaxVertexAttribs;
+    glGetIntegerv(GL_MAX_VERTEX_ATTRIBS, &glMaxVertexAttribs);
```
mb12: Is it possible to wrap this call under MBGL_CHECK_ERROR macro. Its helpful to catch GL bugs. Its stripped out in release builds.

zugaldia:
Just confirming that I can see the message in the Android Studio logs:

image

src/mbgl/gl/context.cpp
anandthakker: 👍 thanks @mb12

jfirebaugh:
I think we're good here; we got the information we need, which is that some implementations have a max of 8.

Issue #6533, opened 2016-09-30
JNI cleanup/refactor


tobrun:
We need to find a way to make our current `jni.cpp` implementation a bit more manageable. 
Using [jni.hpp](https://github.com/mapbox/jni.hpp) has given us a modern type safe wrapper c++14 approach to the jni but it doesn't help us manage the amount of configuration we currently need to do. We should look into structuring the content differently or approach the idea in a different way.

cc @mapbox/android

jfirebaugh:
* Most of jni.cpp is using the [low-level wrappers](https://github.com/mapbox/jni.hpp#low-level-wrappers). Switching to the [high-level wrappers](https://github.com/mapbox/jni.hpp#high-level-wrappers) will cut down on a lot of boilerplate.
* Code in jni.cpp should be split out into individual files for each native type.

PR #8181, opened 2017-02-23
[core] - allow zooming/scaling to use AnimationOptions

tobrun:
This PR enables to use AnimationOptions when performing a zoom/scale transformation (equivalent to #8092 for moveBy). This allows bindings to use an interpolator while doing a zoom/scale transformation.

To showcase the possibilities with this API, here is an example of the double tap zoom gesture using the `easeInOutBack` interpolator with an increased duration:

ezgif com-video-to-gif 22

Note that atm I'm not changing the default behaviour for Android.

### DIFF
--- a/include/mbgl/map/map.hpp
+++ b/include/mbgl/map/map.hpp
@@ -88,22 +88,22 @@ public:
 
     // Position
     void moveBy(const ScreenCoordinate&, const AnimationOptions& = {});
-    void setLatLng(const LatLng&, optional<ScreenCoordinate>, const Duration& = Duration::zero());
-    void setLatLng(const LatLng&, optional<EdgeInsets>, const Duration& = Duration::zero());
-    void setLatLng(const LatLng&, const Duration& = Duration::zero());
+    void setLatLng(const LatLng&, optional<ScreenCoordinate>, const AnimationOptions& = {});
+    void setLatLng(const LatLng&, optional<EdgeInsets>, const AnimationOptions& = {});
+    void setLatLng(const LatLng&, const AnimationOptions& = {});
     LatLng getLatLng(optional<EdgeInsets> = {}) const;
     void resetPosition(optional<EdgeInsets> = {});
 
     // Scale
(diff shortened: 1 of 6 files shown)


include/mbgl/map/map.hpp
```diff
@@ -95,12 +95,12 @@ class Map : private util::noncopyable {
...
     double getScale() const;
-    void setZoom(double zoom, const Duration& = Duration::zero());
-    void setZoom(double zoom, optional<ScreenCoordinate>, const Duration& = Duration::zero());
-    void setZoom(double zoom, optional<EdgeInsets>, const Duration& = Duration::zero());
+    void setZoom(double zoom, const AnimationOptions& = {});
+    void setZoom(double zoom, optional<ScreenCoordinate>, const AnimationOptions& = {});
+    void setZoom(double zoom, optional<EdgeInsets>, const AnimationOptions& = {});
     double getZoom() const;
     void setLatLngZoom(const LatLng&, double zoom, const Duration& = Duration::zero());
     void setLatLngZoom(const LatLng&, double zoom, optional<EdgeInsets>, const Duration& = Duration::zero());
```
1ec5: How about changing these other zoom-related methods too, for consistency?

tobrun: great catch, let me add those as well

jfirebaugh:
How about making this change universally in map.hpp? `s/const Duration& = Duration::zero()/const AnimationOptions& = {}/g`

tobrun:
Requested changes have been implemented (resulted in changing some GLFW code) and were rebased on @ivovandongen JNI refactor PR for #6533. Can I get another round of 👀?

src/mbgl/map/transform_state.cpp
```diff
@@ -1,4 +1,4 @@
-#include <mbgl/map/transform_state.hpp>
+    #include <mbgl/map/transform_state.hpp>
```
jfirebaugh: Accidental added whitespace.

tobrun:
nit was fixed, merging.

Issue #6835, opened 2016-10-26
Remove duplicate mutexes in SpriteAtlas


kkaefer:
We currently have two mutexes in `SpriteAtlas` and should refactor it so only one is used. I assume this is from the `Atlas`/`Store` merge. /cc @jfirebaugh

jfirebaugh:
Yep, looks like it.

PR #8199, opened 2017-02-24
[core] Replace three maps/mutexes in GlyphAtlas with a single map and mutex

jfirebaugh:
Fixes the `GlyphAtlas` equivalent of #6835.

### DIFF
--- a/src/mbgl/text/glyph_atlas.cpp
+++ b/src/mbgl/text/glyph_atlas.cpp
@@ -23,16 +23,17 @@ GlyphAtlas::GlyphAtlas(const Size size, FileSource& fileSource_)
 GlyphAtlas::~GlyphAtlas() = default;
 
 void GlyphAtlas::requestGlyphRange(const FontStack& fontStack, const GlyphRange& range) {
-    std::lock_guard<std::mutex> lock(rangesMutex);
-    auto& rangeSets = ranges[fontStack];
+    std::lock_guard<std::mutex> lock(mutex);
+    auto& rangeSets = entries[fontStack].ranges;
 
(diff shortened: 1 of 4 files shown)


src/mbgl/text/glyph_atlas.cpp
```diff
@@ -64,16 +65,8 @@ bool GlyphAtlas::hasGlyphRanges(const FontStack& fontStack, const GlyphRangeSet&
...
 util::exclusive<GlyphSet> GlyphAtlas::getGlyphSet(const FontStack& fontStack) {
-    auto lock = std::make_unique<std::lock_guard<std::mutex>>(glyphSetsMutex);
-
-    auto it = glyphSets.find(fontStack);
-    if (it == glyphSets.end()) {
-        it = glyphSets.emplace(fontStack, std::make_unique<GlyphSet>()).first;
-    }
-
-    // FIXME: We lock all GlyphSets, but what we should
-    // really do is lock only the one we are returning.
```
kkaefer: This Fixme isn't resolved, is it?

jfirebaugh: No, but I don't think it'll ever be fixed in the way described here. Instead, we'll do #667.

tmpsantos:
This causes `platform/node/test/memory.test.js` to deadlock on `src/mbgl/text/glyph_pbf.cpp:120`.

tmpsantos:
> This causes platform/node/test/memory.test.js to deadlock on src/mbgl/text/glyph_pbf.cpp:120.

After some investigation, the cause seems to be related to the mock file source used in that test returning the reply synchronously.

Issue #8148, opened 2017-02-21
Eliminate mbgl/gl/context.hpp include from public includes


jfirebaugh:
`mbgl/gl/context.hpp` and all of its dependencies are currently forced to be public headers because it is used by SDK files such as `native_map_view.cpp` in Android and `MGLMapView.mm` in iOS/macOS.

This needs to be refactored so that `mbgl/gl/context.hpp` is only exposed internally.

kkaefer:
Fixed in #8214.

PR #8210, opened 2017-02-27
Refactor OpenGL extension loading mechanism

kkaefer:
We are currently loading extensions globally and use the retrieved pointer in all contexts. While this has worked for us so far, the OpenGL spec doesn't guarantee that pointers are identical across contexts. The [OpenGL page has this](https://www.opengl.org/archives/resources/features/OGLextensions/):

> Be careful because the function returned by wglGetProcAddress is only be guaranteed to work for the OpenGL rendering context that was current when wglGetProcAddress was called. If you have multiple contexts that return different extension function addresses, keeping the function addresses in global variables as shown above may create problems. You may need to maintain distinct function addresses on a per-context basis.


src/mbgl/gl/context.hpp
```diff
@@ -155,11 +166,14 @@ class Context : private util::noncopyable {
...
 
+    std::unique_ptr<extension::Debugging> debugging;
+    std::unique_ptr<extension::VertexArray> vertexArray;
+
     State<value::ActiveTexture> activeTexture;
     State<value::BindFramebuffer> bindFramebuffer;
     State<value::Viewport> viewport;
     std::array<State<value::BindTexture>, 2> texture;
-    State<value::BindVertexArray> vertexArrayObject;
+    State<value::BindVertexArray, const Context&> vertexArrayObject { *this };
```
kkaefer: We capture the `Context` into the state object and have it pass it to the individual functions.

src/mbgl/gl/state.hpp
```diff
@@ -50,22 +55,43 @@ class State {
     }
 
 private:
+    template <std::size_t... I>
+    void set(std::index_sequence<I...>) {
+        T::Set(currentValue, std::get<I>(params)...);
+    }
```
kkaefer: This is for passing additional arguments to some setters/getters, in particular to pass `Context` to `BindVertexArray::Set`

include/mbgl/map/backend.hpp
```diff
@@ -20,6 +20,9 @@ class Backend {
     // Returns the backend's context which manages OpenGL state.
     gl::Context& getContext();
 
+    using glProc = void (*)();
+    void initializeExtensions(glProc (*getProcAddress)(const char*));
```
jfirebaugh: I suggest inverting the control here: add an abstract method `glProc getExtensionProc(const char * name) = 0;`, to be called by `Context` at an appropriate time. That will clean up the `Backend` implementations and ensure that they can't forget to initialize the extensions.

platform/qt/include/qmapbox.hpp
```diff
@@ -97,8 +97,6 @@ typedef void (*CustomLayerInitializeFunction)(void* context) ;
 typedef void (*CustomLayerRenderFunction)(void* context, const CustomLayerRenderParameters&);
 typedef void (*CustomLayerDeinitializeFunction)(void* context);
 
-Q_DECL_EXPORT void initializeGLExtensions();
```
jfirebaugh: @tmpsantos @brunoabinader Should we keep this export for backward compatibility, but make the implementation a no-op?

include/mbgl/map/backend.hpp
kkaefer: So far, that "ideal time" is when the backend implementors call `initializeExtensions`. We'd still have to retain that call, but we could remove the `getProcAddress` parameter. Not sure that it's worth it, tbh

kkaefer:
This also brings us closer to #8148

src/mbgl/renderer/painter.cpp
```diff
@@ -89,7 +89,7 @@ Painter::Painter(gl::Context& context_, const TransformState& state_, float pixe
     tileBorderSegments.emplace_back(0, 0, 4, 5);
     rasterSegments.emplace_back(0, 0, 4, 6);
 
-    gl::debugging::enable();
+    context.enableDebugging();
```
kkaefer: We always need it, but currently, there's no guarantee that the OpenGL context will be active when the `Context` object is created (the same goes for extension initialization). We could refactor this code to wait with initialization of the `Backend` object until we know that the OpenGL context is active, e.g. by lazily generating it in the `Backend::getContext()` call.

platform/qt/include/qmapbox.hpp
brunoabinader: @jfirebaugh no need - we can safely remove these. We'll then update the Qt plugin accordingly 👍

include/mbgl/map/backend.hpp
kkaefer: Implemented this.

src/mbgl/renderer/painter.cpp
kkaefer: Refactored this so it always gets enabled as part of the `Backend`'s `Context` object construction.

kkaefer:
@jfirebaugh this is ready for re-review

PR #8216, opened 2017-02-28
Make Context private

kkaefer:
As noted in https://github.com/mapbox/mapbox-gl-native/issues/8148, some SDKs use the GL Context object. We should expose those routines and keep `gl::Context` private.


kkaefer:
Contains https://github.com/mapbox/mapbox-gl-native/pull/8210 and https://github.com/mapbox/mapbox-gl-native/pull/8147

platform/android/config.cmake
```diff
@@ -226,77 +103,182 @@ endmacro()
...
+target_compile_options(mbgl-android
     PRIVATE -fvisibility=hidden
     PRIVATE -ffunction-sections
     PRIVATE -fdata-sections
 )
 
-target_link_libraries(mapbox-gl
+target_link_libraries(mbgl-android
     PUBLIC mbgl-core
-    PUBLIC -Wl,--gc-sections
```
jfirebaugh: Do we not need the `--gc-sections` option anymore?

jfirebaugh:
Build failure should be fixed by rebasing after 83cef9eb686991d5ba31f89e3fe11cafcb4fd5bf.

platform/android/config.cmake
kkaefer: This is the static archive, it's still added to the .so file.

PR #8225, opened 2017-02-28
Per-attribute transition properties on MGLStyleLayer

### LINKED ISSUE #8010: Per-attribute transition properties on MGLStyleLayer
MGLStyleLayer’s concrete subclasses should allow the developer to set transitions that affect individual layout or paint attributes.

We could pair every layout or paint attribute with a `transitionFor…` property. Or we could fight property proliferation by adding `-transitionForKey:` and `-setTransition:forKey:` methods to MGLStyleLayer itself (along the lines of `-[CALayer addAnimation:forKey:]`). (The downside would be pushing validation of the key until runtime.)

Either way, we should add an MGLTransition class or struct to encapsulate the details of a transition. We’ll want to refine the MGLStyle methods added in #7711 to make use of MGLTransition instead of working with durations and delays directly.

/ref #7863
/cc @incanus @boundsj @jfirebaugh @tobrun

fabian-guerra:
Fixes #8010

### DIFF
--- a/platform/darwin/src/MGLBackgroundStyleLayer.h
+++ b/platform/darwin/src/MGLBackgroundStyleLayer.h
@@ -69,6 +69,11 @@ MGL_EXPORT
 #endif
 
 /**
+ `backgroundColor` transition attributes.
+*/
+@property (nonatomic) MGLTransition backgroundColorTransition;
+
+/**
  The opacity at which the background will be drawn.
  
  The default value of this property is an `MGLStyleValue` object containing an
@@ -85,6 +90,11 @@ MGL_EXPORT
 @property (nonatomic, null_resettable) MGLStyleValue<NSNumber *> *backgroundOpacity;
 
 /**
+ `backgroundOpacity` transition attributes.
+*/
+@property (nonatomic) MGLTransition backgroundOpacityTransition;
+
+/**
  Name of image in style images to use for drawing an image background. For
  seamless patterns, image width and height must be a factor of two (2, 4, 8,
  ..., 512).
@@ -96,6 +106,11 @@ MGL_EXPORT
  `MGLInterpolationModeInterval`
  */
 @property (nonatomic, null_resettable) MGLStyleValue<NSString *> *backgroundPattern;
+
+/**
+ `backgroundPattern` transition attributes.
+*/
+@property (nonatomic) MGLTransition backgroundPatternTransition;
 
 @end
 
--- a/platform/darwin/src/MGLBackgroundStyleLayer.h
+++ b/platform/darwin/src/MGLBackgroundStyleLayer.h
@@ -69,7 +69,9 @@ MGL_EXPORT
 #endif
 
 /**
- `backgroundColor` transition attributes.
+ The transition affecting any changes to this layer’s `backgroundColor` property.
+
+ This property corresponds to the background-color-transition property in the style JSON file format.
 */
 @property (nonatomic) MGLTransition backgroundColorTransition;
 
@@ -90,7 +92,9 @@ MGL_EXPORT
 @property (nonatomic, null_resettable) MGLStyleValue<NSNumber *> *backgroundOpacity;
 
 /**
- `backgroundOpacity` transition attributes.
+ The transition affecting any changes to this layer’s `backgroundOpacity` property.
+
+ This property corresponds to the background-color-transition property in the style JSON file format.
 */
 @property (nonatomic) MGLTransition backgroundOpacityTransition;
 
@@ -108,7 +112,9 @@ MGL_EXPORT
 @property (nonatomic, null_resettable) MGLStyleValue<NSString *> *backgroundPattern;
 
 /**
- `backgroundPattern` transition attributes.
+ The transition affecting any changes to this layer’s `backgroundPattern` property.
+
+ This property corresponds to the background-color-transition property in the style JSON file format.
 */
 @property (nonatomic) MGLTransition backgroundPatternTransition;
 
--- a/platform/darwin/src/MGLBackgroundStyleLayer.h
+++ b/platform/darwin/src/MGLBackgroundStyleLayer.h
@@ -71,7 +71,7 @@ MGL_EXPORT
 /**
  The transition affecting any changes to this layer’s `backgroundColor` property.
 
- This property corresponds to the background-color-transition property in the style JSON file format.
+ This property corresponds to the `background-color-transition` property in the style JSON file format.
 */
 @property (nonatomic) MGLTransition backgroundColorTransition;
 
@@ -94,7 +94,7 @@ MGL_EXPORT
 /**
  The transition affecting any changes to this layer’s `backgroundOpacity` property.
 
- This property corresponds to the background-color-transition property in the style JSON file format.
+ This property corresponds to the `background-opacity-transition` property in the style JSON file format.
 */
 @property (nonatomic) MGLTransition backgroundOpacityTransition;
 
@@ -114,7 +114,7 @@ MGL_EXPORT
 /**
  The transition affecting any changes to this layer’s `backgroundPattern` property.
 
- This property corresponds to the background-color-transition property in the style JSON file format.
+ This property corresponds to the `background-pattern-transition` property in the style JSON file format.
 */
 @property (nonatomic) MGLTransition backgroundPatternTransition;
 
--- a/platform/darwin/src/MGLBackgroundStyleLayer.mm
+++ b/platform/darwin/src/MGLBackgroundStyleLayer.mm
@@ -119,6 +119,7 @@
 
     return transition;
 }
+
 - (void)setBackgroundOpacity:(MGLStyleValue<NSNumber *> *)backgroundOpacity {
     MGLAssertStyleLayerIsValid();
 
@@ -153,6 +154,7 @@
 
     return transition;
 }
+
 - (void)setBackgroundPattern:(MGLStyleValue<NSString *> *)backgroundPattern {
     MGLAssertStyleLayerIsValid();
 
--- a/platform/darwin/src/MGLBackgroundStyleLayer.mm
+++ b/platform/darwin/src/MGLBackgroundStyleLayer.mm
@@ -85,7 +85,7 @@
 
 #pragma mark - Accessing the Paint Attributes
 
-- (NSArray *)transitionKeys
++ (NSArray *)transitionKeys
 {
     NSMutableArray *transitionKeys = [NSMutableArray array];
     [transitionKeys addObject:@"backgroundColor"];
--- a/platform/darwin/src/MGLBackgroundStyleLayer.mm
+++ b/platform/darwin/src/MGLBackgroundStyleLayer.mm
@@ -85,30 +85,11 @@
 
 #pragma mark - Accessing the Paint Attributes
 
-+ (NSArray *)transitionKeys
-{
-    return @[
-        @"backgroundColor",
-        @"backgroundOpacity",
-        @"backgroundPattern",
-    ];
-}
-
 - (void)setBackgroundColor:(MGLStyleValue<MGLColor *> *)backgroundColor {
     MGLAssertStyleLayerIsValid();
 
     auto mbglValue = MGLStyleValueTransformer<mbgl::Color, MGLColor *>().toInterpolatablePropertyValue(backgroundColor);
     self.rawLayer->setBackgroundColor(mbglValue);
-}
-
-- (void)mbx_setBackgroundColorTransition:(NSValue *)transitionValue {
-    MGLAssertStyleLayerIsValid();
-
-    MGLTransition transition;
-    [transitionValue getValue:&transition];
-
-    mbgl::style::TransitionOptions options { { MGLDurationFromTimeInterval(transition.duration) }, { MGLDurationFromTimeInterval(transition.delay) } };
-    self.rawLayer->setBackgroundColorTransition(options);
 }
 
 - (MGLStyleValue<MGLColor *> *)backgroundColor {
@@ -121,7 +102,14 @@
     return MGLStyleValueTransformer<mbgl::Color, MGLColor *>().toStyleValue(propertyValue);
 }
 
-- (NSValue *)mbx_getBackgroundColorTransition {
+- (void)setBackgroundColorTransition:(MGLTransition )transition {
+    MGLAssertStyleLayerIsValid();
+
+    mbgl::style::TransitionOptions options { { MGLDurationFromTimeInterval(transition.duration) }, { MGLDurationFromTimeInterval(transition.delay) } };
+    self.rawLayer->setBackgroundColorTransition(options);
+}
+
+- (MGLTransition)backgroundColorTransition {
     MGLAssertStyleLayerIsValid();
 
     mbgl::style::TransitionOptions transitionOptions = self.rawLayer->getBackgroundColorTransition();
@@ -129,25 +117,13 @@
     transition.duration = MGLTimeIntervalFromDuration(transitionOptions.duration.value_or(mbgl::Duration::zero()));
     transition.delay = MGLTimeIntervalFromDuration(transitionOptions.delay.value_or(mbgl::Duration::zero()));
 
-    NSValue *transitionValue = [NSValue value:&transition withObjCType:@encode(MGLTransition)];
-    return transitionValue;
+    return transition;
 }
-
 - (void)setBackgroundOpacity:(MGLStyleValue<NSNumber *> *)backgroundOpacity {
     MGLAssertStyleLayerIsValid();
 
     auto mbglValue = MGLStyleValueTransformer<float, NSNumber *>().toInterpolatablePropertyValue(backgroundOpacity);
     self.rawLayer->setBackgroundOpacity(mbglValue);
-}
-
-- (void)mbx_setBackgroundOpacityTransition:(NSValue *)transitionValue {
-    MGLAssertStyleLayerIsValid();
-
-    MGLTransition transition;
-    [transitionValue getValue:&transition];
-
-    mbgl::style::TransitionOptions options { { MGLDurationFromTimeInterval(transition.duration) }, { MGLDurationFromTimeInterval(transition.delay) } };
-    self.rawLayer->setBackgroundOpacityTransition(options);
 }
 
 - (MGLStyleValue<NSNumber *> *)backgroundOpacity {
@@ -160,7 +136,14 @@
     return MGLStyleValueTransformer<float, NSNumber *>().toStyleValue(propertyValue);
 }
 
-- (NSValue *)mbx_getBackgroundOpacityTransition {
+- (void)setBackgroundOpacityTransition:(MGLTransition )transition {
+    MGLAssertStyleLayerIsValid();
+
+    mbgl::style::TransitionOptions options { { MGLDurationFromTimeInterval(transition.duration) }, { MGLDurationFromTimeInterval(transition.delay) } };
+    self.rawLayer->setBackgroundOpacityTransition(options);
+}
+
+- (MGLTransition)backgroundOpacityTransition {
     MGLAssertStyleLayerIsValid();
 
     mbgl::style::TransitionOptions transitionOptions = self.rawLayer->getBackgroundOpacityTransition();
@@ -168,25 +151,13 @@
     transition.duration = MGLTimeIntervalFromDuration(transitionOptions.duration.value_or(mbgl::Duration::zero()));
     transition.delay = MGLTimeIntervalFromDuration(transitionOptions.delay.value_or(mbgl::Duration::zero()));
 
-    NSValue *transitionValue = [NSValue value:&transition withObjCType:@encode(MGLTransition)];
-    return transitionValue;
+    return transition;
 }
-
 - (void)setBackgroundPattern:(MGLStyleValue<NSString *> *)backgroundPattern {
     MGLAssertStyleLayerIsValid();
 
     auto mbglValue = MGLStyleValueTransformer<std::string, NSString *>().toPropertyValue(backgroundPattern);
     self.rawLayer->setBackgroundPattern(mbglValue);
-}
-
-- (void)mbx_setBackgroundPatternTransition:(NSValue *)transitionValue {
-    MGLAssertStyleLayerIsValid();
-
-    MGLTransition transition;
-    [transitionValue getValue:&transition];
-
-    mbgl::style::TransitionOptions options { { MGLDurationFromTimeInterval(transition.duration) }, { MGLDurationFromTimeInterval(transition.delay) } };
-    self.rawLayer->setBackgroundPatternTransition(options);
 }
 
 - (MGLStyleValue<NSString *> *)backgroundPattern {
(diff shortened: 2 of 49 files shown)


platform/darwin/src/MGLTypes.h
```diff
@@ -85,6 +85,21 @@ typedef NS_OPTIONS(NSUInteger, MGLMapDebugMaskOptions) {
 #endif
 };
 
+/**
+ A structure containing information about a transition values.
```
incanus: Remove `values`.

platform/darwin/src/MGLTypes.h
```diff
@@ -85,6 +85,21 @@ typedef NS_OPTIONS(NSUInteger, MGLMapDebugMaskOptions) {
...
+ */
+typedef struct MGLTransition {
+    /**
+     The duration in seconds to animate any changes to the style URL or to layout and paint attributes.
+     */
+    NSTimeInterval duration;
+    
+    /**
+     The delay in seconds to before applying any changes to the style URL or to layout and paint attributes.
```
incanus: Remove `to` in `to before`.

platform/darwin/src/MGLSymbolStyleLayer.mm
```diff
@@ -887,13 +889,45 @@ - (void)setTextTransform:(MGLStyleValue<NSValue *> *)textTransform {
 
 #pragma mark - Accessing the Paint Attributes
 
+- (void)setTransition:(MGLTransition)transition forKey:(NSString *)key
+{
+    NSString *camelCaseKey;
+    if ([key length] > 1) {
+        camelCaseKey = [NSString stringWithFormat:@"%@%@",[[key substringToIndex:1] uppercaseString],[key substringFromIndex:1]];
```
incanus: Code style: spaces after `,`.

platform/darwin/src/MGLStyleLayer.mm.ejs
```diff
@@ -14,6 +14,8 @@
 #import "MGLStyleValue_Private.h"
 #import "MGL<%- camelize(type) %>StyleLayer.h"
 
+#import "NSDate+MGLAdditions.h"
+
```
incanus: Put this further up with the others, next to `#import "NSPredicate+MGLAdditions.h"`, a similar category use.

platform/darwin/src/MGLStyle.h
```diff
@@ -30,6 +30,7 @@ NS_ASSUME_NONNULL_BEGIN
  */
 static MGL_EXPORT const NSInteger MGLStyleDefaultVersion = 9;
 
+
```
incanus: Extraneous whitespace.

platform/darwin/src/MGLStyle.h
```diff
@@ -194,6 +195,11 @@ MGL_EXPORT
 @property (nonatomic, strong) NS_SET_OF(__kindof MGLSource *) *sources;
 
 /**
+ Transition values.
```
incanus: Should describe it a bit more here, maybe: 

> Values describing animated transitions to styling changes, either to the style URL or to individual properties.

platform/darwin/src/MGLStyle.mm
```diff
@@ -572,6 +572,20 @@ - (MGLImage *)imageForName:(NSString *)name
 
 #pragma mark Style transitions
 
+- (void)setTransition:(MGLTransition)transition
+{
+    [self setTransitionDuration:transition.duration];
+    [self setTransitionDelay:transition.delay];
```
incanus: We should probably get rid of the separate internal `-setTransitionDuration:` and `-setTransitionDelay:` since externally we only set full transition structures containing both values.

platform/darwin/src/MGLStyleLayer.mm.ejs
```diff
@@ -229,6 +231,31 @@ namespace mbgl {
...
+        NSInvocation *setPropertyInvocation = [NSInvocation
+                                                invocationWithMethodSignature:
+                                                [MGL<%- camelize(type) %>StyleLayer instanceMethodSignatureForSelector:setPropertyTransitionSelector]];
+
+        [setPropertyInvocation setSelector:setPropertyTransitionSelector];
+        [setPropertyInvocation setTarget:self];
+
+        [setPropertyInvocation setArgument:&transition atIndex:2];
+
+        [setPropertyInvocation performSelector:@selector(invoke)];
```
incanus: Couldn't this be simply `-[NSObject performSelector:withObject:]` since it's only one argument, passing the `transition` wrapped in `NSValue`?

platform/darwin/src/MGLStyleLayer.mm.ejs
```diff
@@ -229,6 +231,31 @@ namespace mbgl {
...
+    NSString *camelCaseKey;
+    if ([key length] > 1) {
+        camelCaseKey = [NSString stringWithFormat:@"%@%@",[[key substringToIndex:1] uppercaseString],[key substringFromIndex:1]];
+    } else {
+        camelCaseKey = [key uppercaseString];
+    }
+    NSString *setPropertyTransitionString = [NSString stringWithFormat:@"mbx_set%@Transition:", camelCaseKey];
+    SEL setPropertyTransitionSelector = NSSelectorFromString(setPropertyTransitionString);
+
+    if ([self respondsToSelector:setPropertyTransitionSelector]) {
```
boundsj: What is the main motivation to introduce dynamic method invocation for transitions? 

Most of the complexity here is about parsing the string and forming the private method name to match the actual implementation and that introduces new types of edge cases that we did not previously need to handle. For example, what if the client app passes in a nonsensical string at runtime? So far, I think this implementation would fail silently. Should it log a warning, raise an exception, etc.? Do we expect to offer a string parameter based getter for each transition, too?

Since the style layer implementation is generated and in the spirit of consistency with the existing API and its per property getters and setters, I suggest that we consider making the `mbx_` methods in this implementation public and avoid the dynamic invocation altogether.

platform/darwin/src/MGLSymbolStyleLayer.h
```diff
@@ -1204,6 +1206,15 @@ MGL_EXPORT
  * `MGLCameraStyleFunction` with an interpolation mode of:
    * `MGLInterpolationModeExponential`
    * `MGLInterpolationModeInterval`
+ * `MGLSourceStyleFunction` with an interpolation mode of:
```
boundsj: This is a byproduct of the style spec / gl native mismatch. All of these comments should be manually removed in this PR (or work against a style spec gitsha that does not cause this to happen to avoid the inconvenience)

boundsj:
> @incanus Should this PR be based against the 3.5.0 release branch since it builds upon data-driven styling?

The `3.5.0` is actually currently `master`. We expect to make (and share with Android) a 3.5.0 release branch later today or tomorrow. We can point this PR to that branch once it exists.

platform/darwin/src/MGLSymbolStyleLayer.h
boundsj: Also, it is likely that https://github.com/mapbox/mapbox-gl-native/pull/7939 will land before this PR and, I think, this problem will go away then once this branch is rebased on top of that change set.

boundsj: Yes, if you rebase this PR on top of master as of now, f901e77 will fix this problem 👍

platform/darwin/src/MGLStyleLayer.mm.ejs
boundsj: I was pointed to the original post in https://github.com/mapbox/mapbox-gl-native/issues/8010#issue-206634577 so this makes sense now. Thanks!

I still vote for explicit getters and setters. However, if we do go with the CALayer influenced API, it would be nice to try a refactor so that some or all of this string parsing and method invocation logic goes into the base MGLStyleLayer class.

platform/darwin/src/MGLStyleLayer.mm.ejs
fabian-guerra: @incanus using an `NSValue` I will have to wrap it then unwrap in `set***Transition:`, also I will have to add pragma marks to avoid the "PerformSelect may cause a leak because...".

I think this approach is more straightforward, what are your thoughts?

platform/darwin/src/MGLStyleLayer.mm
```diff
@@ -65,6 +65,56 @@ - (float)minimumZoomLevel
...
+    NSString *setPropertyTransitionString = [NSString stringWithFormat:@"mbx_%@%@Transition%@", type, camelCaseKey, parameters];
+    SEL propertyTransitionSelector = NSSelectorFromString(setPropertyTransitionString);
+    
+    return propertyTransitionSelector;
+}
+
+- (NSArray *)transitionKeys
+{
+    // This is overridden by subclasses
+    return nil;
```
1ec5: Raise an MGLAbstractClassException. Search for that string for an example to follow.

platform/darwin/src/MGLBackgroundStyleLayer.mm
```diff
@@ -84,13 +85,33 @@ - (void)removeFromMapView:(MGLMapView *)mapView
...
 #pragma mark - Accessing the Paint Attributes
 
+- (NSArray *)transitionKeys
+{
+    NSMutableArray *transitionKeys = [NSMutableArray array];
+    [transitionKeys addObject:@"backgroundColor"];
+    [transitionKeys addObject:@"backgroundOpacity"];
+    [transitionKeys addObject:@"backgroundPattern"];
+
+    return transitionKeys;
```
1ec5: It would be more efficient to return an array literal.

platform/darwin/src/MGLBackgroundStyleLayer.mm
```diff
@@ -84,13 +85,33 @@ - (void)removeFromMapView:(MGLMapView *)mapView
 
 #pragma mark - Accessing the Paint Attributes
 
+- (NSArray *)transitionKeys
```
1ec5: Since this method doesn’t have any dependencies on other instance methods, you can make this a class method. But I’m unclear on this method’s purpose. It appears to be unused.

1ec5:
Since this PR is scheduled for iOS SDK v3.5.0, please rebase and retarget it at the release-ios-v3.5.0-android-v5.0.0 branch. Thanks!

platform/darwin/src/MGLBackgroundStyleLayer.mm
fabian-guerra: This method returns the keys which you can add a transition

platform/darwin/src/MGLStyleLayer.mm.ejs
```diff
@@ -229,6 +230,19 @@ namespace mbgl {
...
+<% for (const property of paintProperties) { -%>
+    [transitionKeys addObject:@"<%- camelizeWithLeadingLowercase(property.name) %>"];
+<% if (property.original) { -%>
+    [transitionKeys addObject:@"<%- camelizeWithLeadingLowercase(originalPropertyName(property)) %>"];
+<% } -%>
+<% } -%>
+
+    return transitionKeys;
+}
+
```
incanus: Yeah, per https://github.com/mapbox/mapbox-gl-native/pull/8225/files#r104082076, you could do like this: 

```objc
+ (NSArray *)transitionKeys
{    
    return @[
<% for ...
       @"<%- camelizeWithLeadingLowercase(property.name) %>",
<% } -%>
    ];
}
```

fabian-guerra: @incanus I was using master but retarget to rebase release-ios-v3.5.0-android-v5.0.0

platform/android/MapboxGLAndroidSDK/src/main/java/com/mapbox/mapboxsdk/Mapbox.java
```diff
@@ -43,7 +43,7 @@ public static synchronized Mapbox getInstance(@NonNull Context context, @NonNull
     return INSTANCE;
   }
 
-  private Mapbox(@NonNull Context context, @NonNull String accessToken) {
+  Mapbox(@NonNull Context context, @NonNull String accessToken) {
```
boundsj: This PR seems to have picked up additional, unrelated commits along the way when it was rebased like this one from https://github.com/mapbox/mapbox-gl-native/pull/8228

incanus:
Fixed https://github.com/mapbox/mapbox-gl-native/pull/8225#pullrequestreview-25405304 manually with a rebase dropping these oddball non-@fabian-guerra commits and force-pushed to fix 👍

platform/darwin/src/MGLStyleLayer.mm.ejs
1ec5: In the future, do `git rebase -i newBranchName` and remove things you don't want as part of the PR. Then force-push and finally retarget the PR.

platform/darwin/src/MGLStyleLayer.h
```diff
@@ -69,6 +69,17 @@ MGL_EXPORT
...
+#pragma mark Layer Transitions
+
+- (void)setTransition:(MGLTransition)transition forKey:(NSString *)key;
+
+- (MGLTransition)transitionForKey:(NSString *)key;
+
+/**
+ Returns an array of strings that identify transition keys attached to the layer.
+ */
++ (NSArray *)transitionKeys;
```
1ec5: Should we expose this method publicly? If so, let's call it `transitionableAttributes` and also make the return type a set of strings instead of an unqualified array.

platform/darwin/src/MGLStyleLayer.h
```diff
@@ -69,6 +69,17 @@ MGL_EXPORT
  */
 @property (nonatomic, assign) float minimumZoomLevel;
 
+#pragma mark Layer Transitions
```
1ec5: jazzy uses these marks to produce section headings in the generated documentation. For consistency with Apple's documentation, we label each mark according to a task. Here, we can say "Timing Changes to Attributes".

platform/darwin/src/MGLStyleLayer.h
```diff
@@ -69,6 +69,17 @@ MGL_EXPORT
...
+
+- (void)setTransition:(MGLTransition)transition forKey:(NSString *)key;
```
1ec5: Public methods need documentation comments.

1ec5: I'm thinking we might want to call it `-setTransition:forAttribute:` to be more specific about the kind of key that's being transitioned.

platform/darwin/src/MGLStyleLayer.mm
```diff
@@ -65,6 +65,58 @@ - (float)minimumZoomLevel
...
+#pragma clang diagnostic ignored "-Warc-performSelector-leaks"
+        NSValue *transitionValue = [self performSelector:getPropertyTransitionSelector];
+        [transitionValue getValue:&transition];
+#pragma clang diagnostic pop
+    }
+    
+    return transition;
+}
+
+- (SEL)mbx_selectorForKey:(NSString *)key type:(NSString *)type hasParams:(BOOL)params
```
1ec5: There's no need to prefix private methods with a class prefix the way we prefix category methods.

1ec5: This method would be a bit simpler if we had two methods: one for getters and one for setters.

1ec5: Since all the valid setter method names form a closed set, we should generate a lookup table mapping attributes to the corresponding setter name, then look up the passed-in attribute in that table to decide what to call. That way, we can raise an exception when the developer passes in an untransitionable attribute or a bogus attribute. Currently, doing that leads to a nasty crash with no useful information.

platform/darwin/src/MGLBackgroundStyleLayer.mm
1ec5: Ah, looking more closely, this method appears to be used only in tests (but is exposed publicly). As an alternative to this method, could we move the array of keys into the tests themselves?

platform/darwin/src/MGLStyleLayer.h
fabian-guerra: This Method is intended to explicitly return which layer attributes are transitionable, we can as well add a note on each property, but I think this is a better way to let know devs about what they can change, also this can be used as helper to set the same transition property to all transitionable attributes

fabian-guerra:
After talking to @1ec5 and @boundsj I realized that the trade-off for having a single method to set transition VS a straightforward approach was not enough to justify the added complexity.

Each transitionable style attribute has it's corresponding method.

platform/darwin/src/MGLBackgroundStyleLayer.h
```diff
@@ -69,6 +69,11 @@ MGL_EXPORT
 #endif
 
 /**
+ `backgroundColor` transition attributes.
```
1ec5: > The transition affecting any changes to this layer’s `backgroundColor` property.
>
> This property corresponds to the `background-color-transition` property in the style JSON file format.

(We can't quite say "in the Mapbox Style Specification", as we do elsewhere, because of mapbox/mapbox-gl-js#4081.)

platform/darwin/test/MGLBackgroundStyleLayerTests.mm
```diff
@@ -59,6 +62,11 @@ - (void)testProperties {
         XCTAssertThrowsSpecificNamed(layer.backgroundColor = functionStyleValue, NSException, NSInvalidArgumentException, @"MGLStyleValue should raise an exception if it is applied to a property that cannot support it");
         functionStyleValue = [MGLStyleValue<MGLColor *> valueWithInterpolationMode:MGLInterpolationModeInterval compositeStops:@{@18: constantStyleValue} attributeName:@"" options:nil];
         XCTAssertThrowsSpecificNamed(layer.backgroundColor = functionStyleValue, NSException, NSInvalidArgumentException, @"MGLStyleValue should raise an exception if it is applied to a property that cannot support it");
+        // Transition property test
+        layer.backgroundColorTransition = transitionTest;
+        MGLTransition backgroundColorTransition = layer.backgroundColorTransition;
```
1ec5: By setting and getting before asserting, we're verifying that the transition value round-trips. That's good, but it doesn't rule out the possibility that both the getter and setter are buggy. To rule that out, we'd need to assert after setting that the underlying `mbgl::style::BackgroundLayer::getBackgroundColorTransition()` is correct.

platform/darwin/src/MGLSymbolStyleLayer.h
```diff
@@ -1753,6 +1814,7 @@ MGL_EXPORT
  */
 @property (nonatomic, null_resettable) MGLStyleValue<NSValue *> *textTranslationAnchor;
 
+
```
1ec5: Nit: stray line.

platform/darwin/src/MGLTypes.h
```diff
@@ -70,6 +70,21 @@ typedef NS_OPTIONS(NSUInteger, MGLMapDebugMaskOptions) {
...
+     The duration in seconds to animate any changes to the style URL or to layout and paint attributes.
+     */
+    NSTimeInterval duration;
+    
+    /**
+     The delay in seconds before applying any changes to the style URL or to layout and paint attributes.
+     */
+    NSTimeInterval delay;
+} MGLTransition;
+
```
1ec5: We should provide a factory method MGLTransitionMake(). We might also want to add transition-related methods to NSValue(MGLAdditions) to make it easier to work with MGLTransition instances when an object is needed.

platform/darwin/src/MGLTypes.h
```diff
@@ -70,6 +70,21 @@ typedef NS_OPTIONS(NSUInteger, MGLMapDebugMaskOptions) {
 #endif
 };
 
+/**
+ A structure containing information about a transition.
+ */
+typedef struct MGLTransition {
+    /**
+     The duration in seconds to animate any changes to the style URL or to layout and paint attributes.
```
1ec5: It's a bit of a nonsequitur to talk about style URLs and attributes here. This structure is generic enough that its documentation should be more self-contained. For example:

> The amount of time the animation should take, not including the `delay`.

(It doesn't include the delay, does it?)

platform/darwin/src/MGLTypes.h
```diff
@@ -70,6 +70,21 @@ typedef NS_OPTIONS(NSUInteger, MGLMapDebugMaskOptions) {
...
+     */
+    NSTimeInterval duration;
+    
+    /**
+     The delay in seconds before applying any changes to the style URL or to layout and paint attributes.
```
1ec5: > The amount of time in seconds to wait before beginning the animation.

platform/darwin/src/MGLBackgroundStyleLayer.h
```diff
@@ -69,6 +69,13 @@ MGL_EXPORT
 #endif
 
 /**
+ The transition affecting any changes to this layer’s `backgroundColor` property.
+
+ This property corresponds to the background-color-transition property in the style JSON file format.
```
1ec5: Surround `background-color-transition` in backticks.

platform/darwin/test/MGLFillStyleLayerTests.mm
```diff
@@ -137,6 +142,18 @@ - (void)testProperties {
...
         XCTAssertEqualObjects(layer.fillColor, defaultStyleValue,
                               @"fillColor should return the default value after being unset.");
+        // Transition property test
+        layer.fillColorTransition = transitionTest;
+        MGLTransition options;
+        auto toptions = rawLayer->getFillColorTransition();
+        options.delay = MGLTimeIntervalFromDuration(toptions.delay.value_or(mbgl::Duration::zero()));
+        options.duration = MGLTimeIntervalFromDuration(toptions.duration.value_or(mbgl::Duration::zero()));
+        XCTAssertEqual(options.delay, transitionTest.delay);
+        XCTAssertEqual(options.duration, transitionTest.duration);
```
1ec5: I would structure this a little differently:

* Set `fillColorTransition` to `transitionTest`.
* Call `getFillColorTransition()`.
* Assert `toptions.delay && *toptions.delay == transitionTest.delay`.

platform/darwin/src/MGLTypes.h
```diff
@@ -70,6 +70,32 @@ typedef NS_OPTIONS(NSUInteger, MGLMapDebugMaskOptions) {
...
+    
+    /**
+     The amount of time in seconds to wait before beginning the animation.
+     */
+    NSTimeInterval delay;
+} MGLTransition;
+
+/**
+ Creates a new `MGLTransition` from the given duration and delay.
+ */
```
1ec5: Document each `@param` and the `@return` value.

platform/darwin/src/NSValue+MGLAdditions.h
```diff
@@ -70,6 +71,22 @@ NS_ASSUME_NONNULL_BEGIN
  */
 @property (readonly) MGLOfflinePackProgress MGLOfflinePackProgressValue;
 
+#pragma mark Working with Transition values
```
1ec5: Typo: s/values/Values/

platform/darwin/src/MGLTypes.h
```diff
@@ -70,6 +70,39 @@ typedef NS_OPTIONS(NSUInteger, MGLMapDebugMaskOptions) {
...
+ Creates a new `MGLTransition` from the given duration and delay.
+ 
+ @param duration The amount of time the animation should take, not including 
+ the delay.
+ @param delay The amount of time in seconds to wait before beginning the 
+ animation.
+ 
+ @return Returns a `MGLTransition` struct containing the transition attributes.
+ */
+NS_INLINE MGLTransition MGLTransitionMake(NSTimeInterval duration, NSTimeInterval delay) {
```
1ec5: I'm not entirely sure, but you may also need to MGL_EXPORT this function. Generally we have to export public, top-level symbols, but this one is inlined. @kkaefer would know for sure.

platform/darwin/src/NSValue+MGLAdditions.m
```diff
@@ -46,4 +46,16 @@ - (MGLOfflinePackProgress)MGLOfflinePackProgressValue {
...
++ (NSValue *)valueWithMGLTransition:(MGLTransition)transition; {
+    return [NSValue value:&transition withObjCType:@encode(MGLTransition)];
+}
+
+- (MGLTransition)MGLTransitionValue {
+    MGLTransition transition;
+    [self getValue:&transition];
+    return transition;
+}
+
```
incanus: Why have these utilities anymore? We don't use them.

incanus: I've been pointed at https://github.com/mapbox/mapbox-gl-native/pull/8225#discussion_r104857840. 👍🏻

incanus:
🎉

PR #8242, opened 2017-03-01
Use default thread pool instead of node thread pool

bsudekum:
We started to see some adverse behavior after https://github.com/mapbox/mapbox-gl-native/pull/6687, most notably workers becoming unresponsive.

Going to test reverting this change and see how individual threads pools per worker looks.

/cc @jfirebaugh @tmpsantos @brunoabinader


tmpsantos:
Rebased, let me know if we should abandon this or not.

jfirebaugh:
Stale.

PR #8246, opened 2017-03-01
[core] Add support for queryRenderedFeatures filter

### LINKED ISSUE #7319: queryRenderedFeatures should accept filters
`queryRenderedFeatures()` should accept a filter (just like a layer filter) in lieu of or in addition to a bounding box. The developer would be able to search for all POIs with a specific name within the current viewport and get more information about them. To a developer just getting started with these SDKs, this is exactly what `queryRenderedFeatures()` appears to be for. (By contrast, the spatial querying this method currently performs might best be described as _picking_ or _getting_ rather than querying, even if querying happens under the hood.)

Currently, the following workaround exists for querying rendered features by property: create a throwaway layer with minimal opacity, filter it to just the features you want to get back, and pass a bounding box and that layer into `queryRenderedFeatures()`. Alternatively, depending on the number of visible features, it may be reasonably performant to pass a bounding box and an existing layer into `queryRenderedFeatures()`, then manually filter the results to taste.

Note that this wouldn’t be a complete solution for local geocoding: POI and address features in the Mapbox Streets source lack context about associated streets and containing administrative areas, and it `queryRenderedFeatures()` isn’t as comprehensive as `querySourceFeatures()`.

/ref mapbox/mapbox-gl-js#3753

### LINKED ISSUE #8193: Add support for "layers" option for queryRenderedFeatures to node bindings

jfirebaugh:
Fixes #7319
Fixes #8193


platform/ios/src/MGLMapView.mm
```diff
@@ -4605,7 +4606,7 @@ - (void)updateHeadingForDeviceOrientation
     return [self visibleFeaturesAtPoint:point inStyleLayersWithIdentifiers:nil];
 }
 
-- (NS_ARRAY_OF(id <MGLFeature>) *)visibleFeaturesAtPoint:(CGPoint)point inStyleLayersWithIdentifiers:(NS_SET_OF(NSString *) *)styleLayerIdentifiers
+- (NS_ARRAY_OF(id <MGLFeature>) *)visibleFeaturesAtPoint:(CGPoint)point inStyleLayersWithIdentifiers:(NS_SET_OF(NSString *) *)styleLayerIdentifiers withPredicate:(NSPredicate *)predicate
```
1ec5: This change needs to be reflected in the header. Otherwise, someone will call `-visibleFeaturesAtPoint:inStyleLayersWithIdentifiers:` and crash because dynamic method dispatch will fail.

1ec5: Renaming `-visibleFeaturesAtPoint:inStyleLayersWithIdentifiers:` to `-visibleFeaturesAtPoint:inStyleLayersWithIdentifiers:withPredicate:` breaks backwards compatibility. Checking for nil below isn’t enough, because this signature now requires the developer to _specify_ nil rather than omitting that part of the selector. (In Swift, this is the difference between an optional argument and an argument with a default value.) You’ll need to bring back `-visibleFeaturesAtPoint:inStyleLayersWithIdentifiers:`, which can call into this method.

1ec5: Replace `withPredicate:` with `predicate:`. While it can be helpful to make a selector read like an English sentence, at a certain point that falls apart and it’s better to name the rest of the parameters without prepositions.

ivovandongen:
@1ec5 Addressed your comments. Thanks.

platform/ios/src/MGLMapView.h
```diff
@@ -1183,10 +1202,11 @@ IB_DESIGNABLE
  @param styleLayerIdentifiers A set of strings that correspond to the names of
     layers defined in the current style. Only the features contained in these
     layers are included in the returned array.
+ @param predicate A predicate to filter the returned features.
  @return An array of objects conforming to the `MGLFeature` protocol that
     represent features in the sources used by the current style.
  */
-- (NS_ARRAY_OF(id <MGLFeature>) *)visibleFeaturesAtPoint:(CGPoint)point inStyleLayersWithIdentifiers:(nullable NS_SET_OF(NSString *) *)styleLayerIdentifiers NS_SWIFT_NAME(visibleFeatures(at:styleLayerIdentifiers:));
+- (NS_ARRAY_OF(id <MGLFeature>) *)visibleFeaturesAtPoint:(CGPoint)point inStyleLayersWithIdentifiers:(nullable NS_SET_OF(NSString *) *)styleLayerIdentifiers predicate:(nullable NSPredicate*)predicate NS_SWIFT_NAME(visibleFeatures(at:styleLayerIdentifiers:predicate:));
```
1ec5: Nit: space between `NSPredicate` and `*`. The predominent Objective-C coding style is `Type *name` and `Type * const name`. There’s less certainty about the proper style for Objective-C++ code, which is one reason we’ve yet to lint for that sort of stylistic detail.

platform/ios/src/MGLMapView.h
```diff
@@ -1136,6 +1136,25 @@ IB_DESIGNABLE
...
+ 
+ This method may return all features from the specified layers. To filter
+ the returned features, use the
+ `-visibleFeaturesAtPoint:inStyleLayersWithIdentifiers:predicate` method. For more
+ information about searching for map features, see that method’s documentation.
+ 
+ @param point A point expressed in the map view’s coordinate system.
+ @param styleLayerIdentifiers A set of strings that correspond to the names of
+ layers defined in the current style. Only the features contained in these
+ layers are included in the returned array.
```
1ec5: Tip: in Xcode, ⇧⌥⌘V pastes preserving indentation. (The same shortcut preserves formatting in any native macOS application.)

platform/macos/src/MGLMapView.h
```diff
@@ -824,21 +851,22 @@ MGL_EXPORT IB_DESIGNABLE
...
-    layers are included in the returned array.
+ layers defined in the current style. Only the features contained in these
+ layers are included in the returned array.
+ @param predicate A predicate to filter the returned features.
  @return An array of objects conforming to the `MGLFeature` protocol that
-    represent features in the sources used by the current style.
+ represent features in the sources used by the current style.
  */
-- (NS_ARRAY_OF(id <MGLFeature>) *)visibleFeaturesAtPoint:(NSPoint)point inStyleLayersWithIdentifiers:(nullable NS_SET_OF(NSString *) *)styleLayerIdentifiers NS_SWIFT_NAME(visibleFeatures(_:styleLayerIdentifiers:));
+- (NS_ARRAY_OF(id <MGLFeature>) *)visibleFeaturesAtPoint:(NSPoint)point inStyleLayersWithIdentifiers:(nullable NS_SET_OF(NSString *) *)styleLayerIdentifiers predicate:(nullable NSPredicate*)predicate NS_SWIFT_NAME(visibleFeatures(at:styleLayerIdentifiers:predicate:));
```
1ec5: Oh wow, good catch for replacing `_:` with `at:` here!

platform/macos/src/MGLMapView.h
```diff
@@ -858,7 +886,28 @@ MGL_EXPORT IB_DESIGNABLE
...
+ documentation.
+
+ @param rect A rectangle expressed in the map view’s coordinate system.
+ @param styleLayerIdentifiers A set of strings that correspond to the names of
+    layers defined in the current style. Only the features contained in these
+    layers are included in the returned array.
+ @return An array of objects conforming to the `MGLFeature` protocol that
+    represent features in the sources used by the current style.
+ */
+- (NS_ARRAY_OF(id <MGLFeature>) *)visibleFeaturesInRect:(NSRect)rect inStyleLayersWithIdentifiers:(nullable NS_SET_OF(NSString *) *)styleLayerIdentifiers NS_SWIFT_NAME(visibleFeatures(_:styleLayerIdentifiers:));
```
1ec5: Can you use `in:` instead of `_:` here, like you did above?

platform/macos/src/MGLMapView.h
```diff
@@ -789,6 +789,33 @@ MGL_EXPORT IB_DESIGNABLE
  Returns an array of rendered map features that intersect with a given point,
  restricted to the given style layers.
 
+ This method may return all features from the specified layers. To filter
+ the returned features, use the
+ `-visibleFeaturesAtPoint:inStyleLayersWithIdentifiers:predicate` method. For more
```
1ec5: There’s a missing `:` here too. (×2)

platform/macos/src/MGLMapView.h
1ec5: `NSPredicate *` instead of `NSPredicate*`

platform/macos/src/MGLMapView.h
```diff
@@ -789,6 +789,33 @@ MGL_EXPORT IB_DESIGNABLE
...
+    style’s layers over time.
+
+ @param point A point expressed in the map view’s coordinate system.
+ @param styleLayerIdentifiers A set of strings that correspond to the names of
+    layers defined in the current style. Only the features contained in these
+    layers are included in the returned array.
+ @return An array of objects conforming to the `MGLFeature` protocol that
+    represent features in the sources used by the current style.
+ */
+- (NS_ARRAY_OF(id <MGLFeature>) *)visibleFeaturesAtPoint:(NSPoint)point inStyleLayersWithIdentifiers:(nullable NS_SET_OF(NSString *) *)styleLayerIdentifiers NS_SWIFT_NAME(visibleFeatures(_:styleLayerIdentifiers:));
```
1ec5: Can you replace `_:` with `at:` for consistency? Thanks!

PR #8249, opened 2017-03-01
[node] Add request failure tests

tmpsantos:
Test for timeout and when the request fails.

### DIFF
--- a/platform/node/src/node_map.cpp
+++ b/platform/node/src/node_map.cpp
@@ -57,6 +57,7 @@ void NodeMap::Init(v8::Local<v8::Object> target) {
     Nan::SetPrototypeMethod(tpl, "loaded", Loaded);
     Nan::SetPrototypeMethod(tpl, "render", Render);
     Nan::SetPrototypeMethod(tpl, "release", Release);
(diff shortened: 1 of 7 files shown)


tmpsantos:
@bsudekum don't know why the test is crashing on Linux. I can't reproduce it locally. :-/

platform/node/src/node_map.cpp
```diff
@@ -508,6 +509,42 @@ void NodeMap::release() {
...
+void NodeMap::cancel() {
+    auto style = map->getStyleJSON();
+
+    map = std::make_unique<mbgl::Map>(backend, mbgl::Size{ 256, 256 },
+            pixelRatio, *this, threadpool, mbgl::MapMode::Still);
+
+    // FIXME: Reload the style after recreating the map. We need to find
+    // a better way of canceling an ongoing rendering on the core level
+    // without resetting the map, which is way too expensive.
+    map->setStyleJSON(style);
```
bsudekum: I know at one point you said this might not work, is this still the case?

tmpsantos: What was not working is `map->setStyleJSON("{}")`.

PR #8251, opened 2017-03-02
[core] Support ICU cast when UCHAR_TYPE != char16_t

brunoabinader:
/cc @tmpsantos

### DIFF
--- a/platform/default/bidi.cpp
+++ b/platform/default/bidi.cpp
@@ -1,8 +1,10 @@
-#include <memory>
-
 #include <mbgl/text/bidi.hpp>
+#include <mbgl/util/traits.hpp>
+
 #include <unicode/ubidi.h>
 #include <unicode/ushape.h>
+
+#include <memory>
 
 namespace mbgl {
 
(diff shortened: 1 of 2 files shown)


platform/default/bidi.cpp
```diff
@@ -1,9 +1,25 @@
...
+
+namespace {
+
+template <typename To, typename From>
+typename std::enable_if_t<std::is_convertible<From, To>::value, To> convertCharPtr(From data) {
+    return data;
+}
+
+template <typename To, typename From>
+typename std::enable_if_t<!std::is_convertible<From, To>::value && sizeof(To) == sizeof(From), To> convertCharPtr(From data) {
```
jfirebaugh: `To` and `From` are both pointer types here, so `sizeof(To) == sizeof(From)` is always true.

jfirebaugh: Some code you can borrow for this: https://github.com/jfirebaugh/PlusPlus/blob/master/Code/PlusPlus/Standardish/charlike.h

Issue #7777, opened 2017-01-18
Enable bidirectional algorithm and Arabic text shaping when using "WITH_QT_I18N" flag


ChrisLoer:
When we added the ICU library to gl-native to support bidirectional text and Arabic text shaping, we stubbed out the ICU support for Qt when it was built with the "WITH_QT_I18N" flag (https://github.com/mapbox/mapbox-gl-native/pull/7386). This was so that when we include Mapbox GL as a 3rd party submodule within the Qt framework we don't duplicate the ICU dependency Qt already has.

There's no fundamental reason we can't use the same ICU as the rest of the Qt framework, we just need to make sure we're compatible with the version the rest of Qt is using and figure out how to link to it.

cc @tmpsantos @brunoabinader

tmpsantos:
Turns out that not all the platforms supported by Qt ship with ICU. We will have to keep it disabled by now for these platforms.

ChrisLoer:
If the platform doesn't include ICU, is there any problem with linking in our version from mason? Especially since it's a small slice of the library we're pulling in - what makes that different from any other code we'd write ourselves?

tmpsantos:
@ChrisLoer all the other libraries we use in Qt are header-only. ICU we would have to build inside the Qt CI, which is not so simple (but achievable, and we will aim for that).

PR #8252, opened 2017-03-02
[core] Safeguard ICU UChar usage

brunoabinader:
In environments where `Uchar*` cannot be implicitly convertible to `char16_t*` (e.g. when using a system-provided ICU in which `UCHAR_TYPE` was not defined in compile time), we need to safeguard parameter passing from `std::u16string` types.


brunoabinader:
Related: #7777

PR #8255, opened 2017-03-02
MyLocationView activate/deactivate LocationSource

### LINKED ISSUE #6084: Accuracy circle scaling incorrectly
The accuracy circle diameter seems to be at most the size of the screen width - when I zoom in close enough, I can notice that the circle should be bigger than it is shown. See the screenshots for details:
a
Accuracy ring almost touches the road going south on the left when zoomed out
b
When zoomed in, the ring fits the screen and ends up barely after the road going north on the left, which is the first northern road to the left of current location on the first screen.

Another thing is, that the radius is too big - it looked to big for me on wifi in the building, so I've used a mocked location with 40m accuracy. The first screenshot shows a 200+ radius based on google's "measure distance", the second one is closer to 40. Maybe I'm doing something wrong with the location ring? DPI setting, or something like that?

**Platform:** Android
**Mapbox SDK version:** 4.20-SNAPSHOT
### Steps to trigger behavior
1. Show a map with user location with not that great accuracy (couple dozen meters).
2. Zoom a bit to see accuracy ring and take notice where it reaches.
3. Zoom in further to see that the ring is not growing since certain zoom and effectively represents a different accuracy.
### Expected behavior

The distance represented by accuracy radius should be consistent regardless of the zoom level.
### Actual behavior

Zooming in increases accuracy ;)

tobrun:
This PR fixes two issues: 
 - not activating/deactivating the LocationSource when toggling MyLocationView state, we aren't receiving any location updates atm.
 - incorrect accuracy circle width (closes #6084).

### DIFF
--- a/platform/android/MapboxGLAndroidSDK/src/main/java/com/mapbox/mapboxsdk/maps/widgets/MyLocationView.java
+++ b/platform/android/MapboxGLAndroidSDK/src/main/java/com/mapbox/mapboxsdk/maps/widgets/MyLocationView.java
@@ -257,9 +257,7 @@ public class MyLocationView extends View {
 
     final PointF pointF = screenLocation;
     float metersPerPixel = (float) projection.getMetersPerPixelAtLatitude(location.getLatitude());
-    float accuracyPixels = (Float) accuracyAnimator.getAnimatedValue() / metersPerPixel / 2;
-    float maxRadius = getWidth() / 2;
-    accuracyPixels = accuracyPixels <= maxRadius ? accuracyPixels : maxRadius;
+    float accuracyPixels = (Float) accuracyAnimator.getAnimatedValue() / metersPerPixel;
 
     // reset
     matrix.reset();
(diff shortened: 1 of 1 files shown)


platform/android/MapboxGLAndroidSDK/src/main/java/com/mapbox/mapboxsdk/maps/widgets/MyLocationView.java
```diff
@@ -423,22 +421,18 @@ public void onRestoreInstanceState(Parcelable state) {
   private void toggleGps(boolean enableGps) {
     LocationEngine locationEngine = LocationSource.getLocationEngine(getContext());
     if (enableGps) {
-      // Set an initial location if one available
-      Location lastLocation = locationEngine.getLastLocation();
```
zugaldia: @tobrun why not request a manual update? This way we could have a faster location reading before the `LocationEngineListener` receives the first update?

tobrun: can you clarify manual update? Expose a setter for end-users? or saving the last location ourselves? some background on why these lines were removed, this line will always return null as the LocationEngine hasn't been connected (this is due the change in LOST 1.0 -> 2.0).

PR #8259, opened 2017-03-02
[ios] Added a timing note for two delegate methods

jmkiley:
Added a sentence to `regionIsChanging` and `regionDidChangeAnimated` to state that these delegate methods can be called before the map view has finished loading.

### DIFF
--- a/platform/ios/src/MGLMapViewDelegate.h
+++ b/platform/ios/src/MGLMapViewDelegate.h
@@ -22,12 +22,13 @@ NS_ASSUME_NONNULL_BEGIN
 #pragma mark Responding to Map Position Changes
 
 /**
- Tells the delegate that the viewpoint depicted by the map view is about to
- change.
-
- This method is called whenever the currently displayed map camera will start
- changing for any reason.
-
+ Tells the delegate that the viewpoint depicted by the map view is changing.
+ 
+ This method is called whenever the currently displayed map camera has finished
+ changing, after any calls to `-mapViewRegionIsChanging:` due to animation.
+ 
+ This method can be called before the map view has finished loading.
+ 
  @param mapView The map view whose viewpoint will change.
  @param animated Whether the change will cause an animated effect on the map.
  */
@@ -38,10 +39,12 @@ NS_ASSUME_NONNULL_BEGIN
 
  This method is called as the currently displayed map camera changes as part of
  an animation, whether due to a user gesture or due to a call to a method such
- as `-[MGLMapView setCamera:animated:]`. During the animation, this method may
- be called many times to report updates to the viewpoint. Therefore, your
- implementation of this method should be as lightweight as possible to avoid
- affecting performance.
+ as `-[MGLMapView setCamera:animated:]`. This method can be called before the map 
+ view has finished loading.
+ 
+ During the animation, this method may be called many times to report updates to 
+ the viewpoint. Therefore, your implementation of this method should be as lightweight 
+ as possible to avoid affecting performance.
 
  @param mapView The map view whose viewpoint is changing.
  */
@@ -52,7 +55,8 @@ NS_ASSUME_NONNULL_BEGIN
  changing.
 
  This method is called whenever the currently displayed map camera has finished
- changing, after any calls to `-mapViewRegionIsChanging:` due to animation.
+ changing, after any calls to `-mapViewRegionIsChanging:` due to animation. Therefore, 
+ this method can be called before the map view has finished loading.
 
  @param mapView The map view whose viewpoint has changed.
  @param animated Whether the change caused an animated effect on the map.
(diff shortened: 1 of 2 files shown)


jmkiley:
It may also be better to insert the clarification [here](https://github.com/mapbox/mapbox-gl-native/blob/master/platform/ios/src/MGLMapViewDelegate.h#L11-L13) than in individual delegate methods.

platform/ios/src/MGLMapViewDelegate.h
```diff
@@ -38,10 +38,10 @@ NS_ASSUME_NONNULL_BEGIN
 
  This method is called as the currently displayed map camera changes as part of
  an animation, whether due to a user gesture or due to a call to a method such
- as `-[MGLMapView setCamera:animated:]`. During the animation, this method may
- be called many times to report updates to the viewpoint. Therefore, your
- implementation of this method should be as lightweight as possible to avoid
- affecting performance.
+ as `-[MGLMapView setCamera:animated:]`. It can be called before the map view
```
friedbunny: Even though it’s a bit repetitive, I’d continue to use “this method” instead of a less specific article like “it”.

I’d also suggest that we move this note into a new paragraph and possibly expand it to mention `-mapViewDidFinishLoadingMap:` and/or `-mapView:didFinishLoadingStyle:`.

friedbunny: In general, I think we should expand the delegate method docs to clarify their relative timing.

jmkiley: Agreed, it would better in the long run to expand on that.

1ec5:
This PR pulled in several unrelated commits. Please rebase this branch against master to clean up the commit history. (Then, if you retarget the PR to the release branch, also rebase against the release branch.)

jmkiley:
Thanks @1ec5! Rebased and added the changes to macOS. I am getting a lot of file changes when I rebase on `release-ios-v3.5.0-android-v5.0.0` and switch the target, so I'm going to keep working on fixing that.

jmkiley:
Reset and rebased.

platform/ios/src/MGLMapViewDelegate.h
```diff
@@ -22,12 +22,13 @@ NS_ASSUME_NONNULL_BEGIN
...
-
- This method is called whenever the currently displayed map camera will start
- changing for any reason.
-
+ Tells the delegate that the viewpoint depicted by the map view is changing.
+ 
+ This method is called whenever the currently displayed map camera has finished
+ changing, after any calls to `-mapViewRegionIsChanging:` due to animation.
+ 
+ This method can be called before the map view has finished loading.
```
1ec5: “Has finished loading” is a bit ambiguous, since MGLMapViewDelegate contains both `-mapViewDidFinishLoadingMap:` and `-mapView:didFinishLoadingStyle:`, and the developer might understand this to mean `-mapViewDidFinishRenderingMap:fullyRendered:`. How about “before `-mapViewDidFinishLoadingMap:` is called”.

platform/ios/src/MGLMapViewDelegate.h
```diff
@@ -22,12 +22,11 @@ NS_ASSUME_NONNULL_BEGIN
 #pragma mark Responding to Map Position Changes
 
 /**
- Tells the delegate that the viewpoint depicted by the map view is about to
- change.
-
- This method is called whenever the currently displayed map camera will start
- changing for any reason.
-
+ Tells the delegate that the viewpoint depicted by the map view is changing.
```
1ec5: Saying "is changing" here blurs the distinction between this method and `-mapViewRegionIsChanging:`.

platform/ios/src/MGLMapViewDelegate.h
```diff
@@ -22,12 +22,11 @@ NS_ASSUME_NONNULL_BEGIN
...
+ 
+ This method is called whenever the currently displayed map camera has finished
+ changing, after any calls to `-mapViewRegionIsChanging:` due to animation.
```
1ec5: This is incorrect: this method is called _before_ any calls to `-mapViewRegionIsChanging:`.

jmkiley: Eep, thanks! Accidentally copy/pasted into the wrong method after the reset.

platform/macos/src/MGLMapViewDelegate.h
```diff
@@ -55,7 +57,9 @@ NS_ASSUME_NONNULL_BEGIN
 
  This method is called whenever the currently displayed map camera has finished
  changing, after any calls to `-mapViewRegionIsChanging:` due to animation.
-
+ Therefore, this method can be called before `-mapViewDidFinishLoadingMap:` is
```
1ec5: "Therefore" seems out of place; I think it can be removed.

PR #8263, opened 2017-03-02
[core] query source features

### LINKED ISSUE #5792: Query all features in sources (querySourceFeatures)

ivovandongen:
Fixes #5792


include/mbgl/map/map.hpp
```diff
@@ -3,6 +3,7 @@
 #include <mbgl/util/optional.hpp>
 #include <mbgl/util/chrono.hpp>
 #include <mbgl/map/mode.hpp>
+#include <mbgl/map/query.hpp>
```
jfirebaugh: This is already included below.

include/mbgl/map/query.hpp
```diff
@@ -6,13 +6,24 @@
...
+    optional<style::Filter> filter;
+};
+
+/**
+ * Options for query source features
+ */
+class SourceQueryOptions {
+public:
+    // Required for VectorSource, ignored for GeoJSONSource
+    optional<std::string> sourceLayer;
```
jfirebaugh: Let's make this a `std::vector<std::string>` like `RenderedQueryOptions`. I don't think there's a good reason that the GL JS API is inconsistent (https://github.com/mapbox/mapbox-gl-js/issues/4372).

include/mbgl/map/map.hpp
```diff
@@ -182,10 +183,11 @@ class Map : private util::noncopyable {
...
 
     // Feature queries
-    std::vector<Feature> queryRenderedFeatures(const ScreenCoordinate&, const QueryOptions& options = {});
-    std::vector<Feature> queryRenderedFeatures(const ScreenBox&,        const QueryOptions& options = {});
-    
+    std::vector<Feature> queryRenderedFeatures(const ScreenCoordinate&, const RenderedQueryOptions& options = {});
+    std::vector<Feature> queryRenderedFeatures(const ScreenBox&,        const RenderedQueryOptions& options = {});
+
     AnnotationIDs queryPointAnnotations(const ScreenBox&);
+    std::vector<Feature> querySourceFeatures(const std::string&,        const SourceQueryOptions& options = {});
```
jfirebaugh: GL native exposes an object-oriented API for sources -- move this method to `mbgl::Style::Source`, rename it to `queryFeatures`, and drop the first parameter.

ivovandongen: @jfirebaugh This does mean that you have to retrieve the source in the SDKs before querying. Small overhead, but still. Is this what we want?

ivovandongen: @jfirebaugh In this case; shall I move [SourceQueryOptions](https://github.com/mapbox/mapbox-gl-native/blob/03969287248730779ed65287f28c45f7cb566f04/include/mbgl/map/query.hpp#L22) to a separate header in the `mbgl::style` namespace? or keep it here?

1ec5:
> This does mean that you have to retrieve the source in the SDKs before querying. Small overhead, but still. Is this what we want?

I've been thinking about making the same change to the iOS/macOS SDKs' public API for queryRenderedFeatures: accept an array of MGLStyleLayers instead of an array of identifiers in an awkwardly named parameter. The overhead is negligible for moderately complex styles. macosapp, for instance, retrieves and wraps every style layer on load, and any performance impact seems reasonable.

platform/ios/src/MGLMapView.h
```diff
@@ -1300,6 +1300,34 @@ IB_DESIGNABLE
...
+ 
+ @param sourceIdentifier The source to query.
+ @param sourceLayerIdentifier The source layer to include in the query. Only the 
+    features contained in this source layer are included in the returned array. 
+    The source layer is ignored for GeoJSON sources and required for vector sources.
+ @param predicate A predicate to filter the returned features.
+ @return An array of objects conforming to the `MGLFeature` protocol that
+    represent features in the sources used by the current style.
+ */
+- (NS_ARRAY_OF(id <MGLFeature>) *)featuresWithSourceIdentifier:(NSString *)sourceIdentifier sourceLayerIdentifier:(nullable NSString *)sourceLayerIdentifier predicate:(nullable NSPredicate *)predicate NS_SWIFT_NAME(features(sourceIdentifier:sourceLayerIdentifier:predicate:));
```
1ec5: How about `-featuresLoadedInSource:sourceLayerIdentifier:predicate:`?

Especially if this method only works with a single source at a time, I think it would be desirable to require the developer to pass in an MGLSource object instead of a source identifier. That encourages the developer to obtain an existing source and helps to clarify the distinction between a source identifier and a source layer identifier.

Additionally, we probably want to emphasize that this method returns the features that have already loaded, not all the possible features in the source (which for a vector source could be quite unwieldy). This isn't quite a reverse geocoding API, after all.

platform/ios/src/MGLMapView.h
```diff
@@ -1300,6 +1300,34 @@ IB_DESIGNABLE
...
+ Each object in the returned array represents a feature for the
+ current style and provides access to attributes specified by the relevant
+ <a href="https://www.mapbox.com/mapbox-gl-style-spec/#sources">tile sources</a>.
+ The returned array includes features specified in vector and GeoJSON tile
+ sources but does not include anything from raster, image, or video sources.
+
+ Features come from tiled vector data or GeoJSON data that is converted to tiles
+ internally, so feature geometries are clipped at tile boundaries and features
+ may appear duplicated across tiles. For example, suppose the specified
+ rectangle intersects with a road that spans the screen. The resulting array
```
1ec5: Some of this explanation is specific to -visibleFeaturesInRect:. This method by contrast accepts no rectangle parameter.

platform/ios/src/MGLMapView.h
1ec5: `-featuresLoadedBySource:sourceLayerIdentifier:predicate:` would avoid overusing "in", especially if we later add a bounding box option.

Swift can probably come up with a decent name for this method without the `NS_SWIFT_NAME` attribute. (Try it out locally by trying to call the method in one of our Swift unit test files.) If not, try providing a Swift name of `features(loadedBy:sourceLayerIdentifier:predicate:)`.

include/mbgl/map/map.hpp
jfirebaugh: > This does mean that you have to retrieve the source in the SDKs before querying. Small overhead, but still. Is this what we want?

Yes, I think so -- by eliminating the `map` middleman in `map.querySourceFeatures(source)`, `source.queryFeatures()` becomes conceptually simpler and possibly actually _more_ efficient in the case of repeated queries (though I would indeed expect the benefit to be negligible).

> shall I move SourceQueryOptions to a separate header in the `mbgl::style` namespace?

Yeah, I think that makes sense.

platform/ios/src/MGLMapView.h
jfirebaugh: We'll want to push the change suggested above (moving the query method to the `Source` object itself) out to the SDK level, making this `-[MGLSource featuresLoaded:inLayerWithIdentifier:predicate:]` or similar.

platform/darwin/src/MGLShapeSource.h
```diff
@@ -210,6 +210,31 @@ MGL_EXPORT
...
+ Each object in the returned array represents a feature for the
+ current style and provides access to attributes specified by the relevant
+ <a href="https://www.mapbox.com/mapbox-gl-style-spec/#sources">tile sources</a>.
+ The returned array includes features specified in vector and GeoJSON tile
+ sources but does not include anything from raster, image, or video sources.
+ 
+ Features come from tiled vector data or GeoJSON data that is converted to tiles
+ internally, so feature geometries are clipped at tile boundaries and features
+ may appear duplicated across tiles. For example, suppose the specified
+ rectangle intersects with a road that spans the screen. The resulting array
```
1ec5: No rectangle is specified as part of this API.

platform/darwin/src/MGLVectorSource.h
```diff
@@ -42,10 +43,41 @@ NS_ASSUME_NONNULL_BEGIN
 MGL_EXPORT
 @interface MGLVectorSource : MGLTileSource
 
+#pragma mark Initializing a Source
+
 - (instancetype)initWithIdentifier:(NSString *)identifier configurationURL:(NSURL *)configurationURL NS_DESIGNATED_INITIALIZER;
 
 - (instancetype)initWithIdentifier:(NSString *)identifier tileURLTemplates:(NS_ARRAY_OF(NSString *) *)tileURLTemplates options:(nullable NS_DICTIONARY_OF(MGLTileSourceOption, id) *)options NS_DESIGNATED_INITIALIZER;
 
+#pragma mark Querying the source's features
```
1ec5: Since jazzy uses `mark`s to generate section headings, it’s important to keep them consistent (title case, describes a task, uses the most commonplace terms possible). How about “Accessing a Source’s Content”, which is the same `mark` we use in MGLShapeSource?

platform/darwin/src/MGLShapeSource.h
```diff
@@ -210,6 +210,31 @@ MGL_EXPORT
  */
 @property (nonatomic, copy, nullable) NSURL *URL;
 
+#pragma mark Querying the source's features
```
1ec5: I think `-featuresMatchingPredicate:` falls neatly under the existing “Accessing a Source’s Content” section heading.

platform/darwin/src/MGLShapeSource.h
```diff
@@ -210,6 +210,31 @@ MGL_EXPORT
...
+ rectangle intersects with a road that spans the screen. The resulting array
+ includes those parts of the road that lie within the map tiles covering the
+ specified rectangle, even if the road extends into other tiles. The portion of
+ the road within each map tile is included individually.
+ 
+ @param predicate A predicate to filter the returned features.
+ @return An array of objects conforming to the `MGLFeature` protocol that
+ represent features in the sources used by the current style.
+ */
+- (NS_ARRAY_OF(id <MGLFeature>) *)featuresWithPredicate:(nullable NSPredicate *)predicate NS_SWIFT_NAME(features(predicate:));
```
1ec5: Call this method `-featuresMatchingPredicate:` and remove the `NS_SWIFT_NAME` attribute. Swift will automatically bridge it as `features(matching:)`.

The convention in Swift 3 is for argument labels to be prepositions or participles like “matching” instead of prepositional objects like “predicate”. `matching:` would be more intuitive argument label than `with:` in this case. (It’d also be consistent with methods like `-[CNContactStore groupsMatchingPredicate:error:]`, `-[EKEventStore eventsMatchingPredicate:]`, and `-[XCUIElementQuery matchingPredicate:]` in standard Apple libraries.)

platform/darwin/src/MGLShapeSource.h
```diff
@@ -210,6 +210,31 @@ MGL_EXPORT
...
+#pragma mark Querying the source's features
+
+/**
+ Returns an array of map features for this source, filtered by the given predicate.
+ 
+ Each object in the returned array represents a feature for the
+ current style and provides access to attributes specified by the relevant
+ <a href="https://www.mapbox.com/mapbox-gl-style-spec/#sources">tile sources</a>.
+ The returned array includes features specified in vector and GeoJSON tile
+ sources but does not include anything from raster, image, or video sources.
```
1ec5: Now that this method lives on MGLShapeSource, it’s unnecessary to point out that it won’t return features from non-shape sources. It’s also important to note that this method may return features that exist in the source but are not currently drawn by a style layer.

platform/darwin/src/MGLVectorSource.h
```diff
@@ -42,10 +43,41 @@ NS_ASSUME_NONNULL_BEGIN
...
+
+/**
+ Returns an array of map features for this source, restricted to the given
+ source layers and filtered by the given predicate.
+ 
+ Each object in the returned array represents a feature for the
+ current style and provides access to attributes specified by the relevant
+ <a href="https://www.mapbox.com/mapbox-gl-style-spec/#sources">tile sources</a>.
+ The returned array includes features specified in vector and GeoJSON tile
+ sources but does not include anything from raster, image, or video sources.
```
1ec5: Similarly, there’s no need here to talk about other sources.

platform/darwin/src/MGLVectorSource.h
```diff
@@ -42,10 +43,41 @@ NS_ASSUME_NONNULL_BEGIN
...
+ Features come from tiled vector data or GeoJSON data that is converted to tiles
+ internally, so feature geometries are clipped at tile boundaries and features
+ may appear duplicated across tiles. For example, suppose the specified
+ rectangle intersects with a road that spans the screen. The resulting array
+ includes those parts of the road that lie within the map tiles covering the
+ specified rectangle, even if the road extends into other tiles. The portion of
+ the road within each map tile is included individually.
+ 
+ @param sourceLayerIdentifiers The source layers to include in the query. Only the
+    features contained in this source layers are included in the returned array. At
```
1ec5: s/this source layers/these source layers/

platform/darwin/src/MGLVectorSource.h
```diff
@@ -42,10 +43,41 @@ NS_ASSUME_NONNULL_BEGIN
...
+    least one source layer is required.
+ @param predicate A predicate to filter the returned features.
+ @return An array of objects conforming to the `MGLFeature` protocol that 
+    represent features in the sources used by the current style.
```
1ec5: Is there any significance to the order in which the features are returned? (It’s fine if there isn’t; we may even want to leave the order unmentioned as wiggle room for future improvements.)

platform/darwin/src/MGLVectorSource.h
```diff
@@ -42,10 +43,41 @@ NS_ASSUME_NONNULL_BEGIN
...
+#pragma mark Initializing a Source
+
 - (instancetype)initWithIdentifier:(NSString *)identifier configurationURL:(NSURL *)configurationURL NS_DESIGNATED_INITIALIZER;
 
 - (instancetype)initWithIdentifier:(NSString *)identifier tileURLTemplates:(NS_ARRAY_OF(NSString *) *)tileURLTemplates options:(nullable NS_DICTIONARY_OF(MGLTileSourceOption, id) *)options NS_DESIGNATED_INITIALIZER;
 
+#pragma mark Querying the source's features
+
+/**
+ Returns an array of map features for this source, restricted to the given
```
1ec5: s/for this source/loaded by this source/

platform/darwin/test/MGLSourceQueryTests.m
```diff
@@ -0,0 +1,73 @@
...
+        _styleLoadingExpectation = [self expectationWithDescription:@"Map view should finish loading style."];
+        [self waitForExpectationsWithTimeout:1 handler:nil];
+    }
+}
+
+- (void)mapView:(MGLMapView *)mapView didFinishLoadingStyle:(MGLStyle *)style {
+    XCTAssertNotNil(mapView.style);
+    XCTAssertEqual(mapView.style, style);
+    
+    [_styleLoadingExpectation fulfill];
```
1ec5: In case you’re wondering why this works for shape sources but not vector sources, `-mapView:didFinishLoadingStyle:` gets called as soon as we’re done loading and parsing the style itself. It isn’t guaranteed that all sources load by this point. If you wanted to query a vector source too, you’d need to wait until `-mapViewDidFinishLoadingMap:` is called. But this is perfectly fine for the inline GeoJSON sources in query-style.json, since they load as part of the style.

ivovandongen: > If you wanted to query a vector source too, you’d need to wait until -mapViewDidFinishLoadingMap: is called

I've tried this, but this callback is never fired in this setup.

> this is perfectly fine for the inline GeoJSON sources in query-style.json, since they load as part of the style.

I expected the same, but the inline geojson sources also don't seem to load this way.

platform/darwin/src/MGLVectorSource.h
ivovandongen: No, not any order in particular and I don't think we should make any guarantees for now. Right now the order is sourceLayer, feature index.

ivovandongen:
@jfirebaugh Made the changes you requested and adapted the SDK bindings. Adding support for multiple sourceLayers resulted in some code duplication in: [GeoJSONTile:: querySourceFeatures](https://github.com/mapbox/mapbox-gl-native/blob/4219e261a1d4024a26bb614609be41c46a974c2e/src/mbgl/tile/geojson_tile.cpp#L95). Could extract the common functionality, not sure if it's worth it.

platform/darwin/src/MGLShapeSource.h
```diff
@@ -210,6 +210,23 @@ MGL_EXPORT
  */
 @property (nonatomic, copy, nullable) NSURL *URL;
 
+/**
+ Returns an array of map features for this source, filtered by the given predicate.
+ 
+ Each object in the returned array represents a feature for the
+ current style and provides access to attributes specified by the
+ <a href="https://www.mapbox.com/mapbox-gl-style-spec/#sources">source</a>.
```
1ec5: There’s also no need to link “source”, since this method lives inside an MGLSource subclass.

platform/darwin/src/MGLVectorSource.h
```diff
@@ -42,10 +43,35 @@ NS_ASSUME_NONNULL_BEGIN
...
+#pragma mark Initializing a Source
+
 - (instancetype)initWithIdentifier:(NSString *)identifier configurationURL:(NSURL *)configurationURL NS_DESIGNATED_INITIALIZER;
 
 - (instancetype)initWithIdentifier:(NSString *)identifier tileURLTemplates:(NS_ARRAY_OF(NSString *) *)tileURLTemplates options:(nullable NS_DICTIONARY_OF(MGLTileSourceOption, id) *)options NS_DESIGNATED_INITIALIZER;
 
+#pragma mark Accessing a Source’s Content
+
+/**
+ Returns an array of map features for this source, restricted to the given
```
1ec5: Reminder: s/for this source/loaded by this source/

platform/darwin/src/MGLVectorSource.h
```diff
@@ -42,10 +43,35 @@ NS_ASSUME_NONNULL_BEGIN
...
+ Each object in the returned array represents a feature for the
+ current style and provides access to attributes specified by the
+ <a href="https://www.mapbox.com/mapbox-gl-style-spec/#sources">source</a>.
+ 
+ Features come from tiled vector data that is converted to tiles internally, 
+ so feature geometries are clipped at tile boundaries and features
+ may appear duplicated across tiles.
+ 
+ @param sourceLayerIdentifiers The source layers to include in the query. Only the
+    features contained in this source layers are included in the returned array. At
```
1ec5: Reminder: s/this source layers/these source layers/

PR #8265, opened 2017-03-03
[ios] read only annotation

fabian-guerra:
This fix was merged into master but should be merged into release branch.
\cc @boundsj


1ec5:
This went into the release branch as part of #8324. Nothing left to do here.

fabian-guerra:
@1ec5 this branch raises an `NSException` instead of using an `NSAssert`

1ec5:
Ah, I missed that. Can you open a new PR that does just that, now that all the other changes are already included in the release branch?

PR #8266, opened 2017-03-03
Add Glyph Atlas debug window back

kkaefer:
At some point we lost the commented out code that shows the glyph window in GLFW.


jfirebaugh:
Pretty sure this code is going to continue to be a tempting target for removal if it's commented and ifdeffed out -- not to mention getting broken by changes since it's not compiled by default. Can we figure out a more durable solution?

kkaefer:
This needs significant refactoring once #9213 lands.

jfirebaugh:
Stale.

PR #8267, opened 2017-03-03
[core] Reduce number of vertex attributes passed to shaders

### LINKED ISSUE #8183: Reduce the number of vertex attributes used for data-driven styling.
@ivovandongen discovered, on his (older?) Nexus device w/ Android 4.3 (Dalvik):

```
01-01 03:37:46.234 2208-2208/com.mapbox.mapboxsdk.testapp W/dalvikvm: threadid=1: thread exiting with uncaught exception (group=0x41a32700)
01-01 03:37:46.257 2208-2208/com.mapbox.mapboxsdk.testapp E/AndroidRuntime: FATAL EXCEPTION: main
                                                                            java.lang.Error: glBindAttribLocation(id, location, name): Error GL_INVALID_VALUE
```

Using https://github.com/mapbox/mapbox-gl-native/pull/8177, it looks like his device has `GL_MAX_VERTEX_ATTRIBS = 8`.  Sounds like this is a common enough device that we need to reduce the number of attributes we're binding for each program. 

Couple ways to do this that @jfirebaugh and I discussed for a different issue (which was solved another way):

> we can pack the base attributes (a_pos, a_offset, a_texture_pos, a_data) more

> make the min/max for `float` attributes into vectors

mollymerp:
fixes #8183 
related gl-js PR: https://github.com/mapbox/mapbox-gl-js/pull/4363

@anandthakker and I worked on this for the past couple days 🤖

cc @jfirebaugh

### DIFF
--- a/scripts/generate-shaders.js
+++ b/scripts/generate-shaders.js
@@ -60,9 +60,11 @@ ${fragmentPrelude}
 ].forEach(function (shaderName) {
     function applyPragmas(source, pragmas) {
         return source.replace(/#pragma mapbox: ([\w]+) ([\w]+) ([\w]+) ([\w]+)/g, (match, operation, precision, type, name) => {
+            const a_type = type === "float" ? "vec2" : "vec4";
             return pragmas[operation]
                 .join("\n")
                 .replace(/\{type\}/g, type)
+                .replace(/\{a_type}/g, a_type)
                 .replace(/\{precision\}/g, precision)
                 .replace(/\{name\}/g, name);
         });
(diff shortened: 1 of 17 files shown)


anandthakker:
Couple outstanding TODOs:

- [ ] Add some code comments, particularly (a) providing explanation for what "MinMax" is all about, and (b) documenting the rationale and scheme for encoding colors into a pair of floats.
- [ ] Retarget to the release branch.
- [ ] (minor, possibly punt for now) Use c++ template trickery to concatenate [these arrays](https://github.com/mapbox/mapbox-gl-native/pull/8267/files#diff-f33795e130a92381a0896a3792f68cb1R46) statically rather than with a loop.

jfirebaugh:
When source functions are used, this implementation will create vertex buffers that contain both min and max values, when only the min value is used. In other words, the buffers are twice as big as they need to be. We discussed an implementation where the vertex buffer contains only the min value, even though the shader continues to use a `vec2`.

anandthakker:
Ah, yep @jfirebaugh meant to put that on the todo list -- updated now.

anandthakker:
Rebased back to master.  Because of https://github.com/mapbox/mapbox-gl-native/pull/8316 -- and the fact that we had multiple commits that modified the shaders, it was easier to first squash to a single commit before rebasing.  (The history wasn't especially useful in this case anyway.)

PR #8269, opened 2017-03-03
Update callout/popover when relocating annotation

### LINKED ISSUE #6284: Callout dismisses when annotation coordinate changes slightly
We should avoid deselecting an annotation (closing its callout) when its `coordinate` property changes, as long as the annotation remains within the viewport. We did this for the user dot (#2618), but not for annotations in general. #3154 is related in the sense that changing the viewport also shouldn’t close the callout.

/cc @boundsj @friedbunny

1ec5:
When relocating a selected point annotation, update the associated callout (iOS) or popover (macOS) to reflect the annotation’s new coordinates rather than deselecting the annotation.

Fixes #6284.

/cc @frederoni

### DIFF
--- a/platform/ios/CHANGELOG.md
+++ b/platform/ios/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 Mapbox welcomes participation and contributions from everyone. Please read [CONTRIBUTING.md](../../CONTRIBUTING.md) to get started.
 
-## master
+## 3.5.0
 
 ### Internationalization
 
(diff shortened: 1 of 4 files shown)


incanus:
FYI the historical reason here was because SMCalloutView (when we used it) required a re-layout since it weights left or right when near the screen edges.

1ec5:
We still use SMCalloutView, and it’s true that the anchor will remain weighted based on the original location on screen. But this affects all callouts, including the user dot’s callout. Due to how SMCalloutView encapsulates its layout code but doesn’t support dynamic layout, a fix would essentially entail replacing SMCalloutView with something better (#4392), since at that point we’d be unable to use almost anything that SMCalloutView provides as-is.

PR #8276, opened 2017-03-03
[ios] duration to time interval fix

### LINKED ISSUE #8264: NSDate(MGLAdditions) turns seconds into nanoseconds
`MGLDurationInSecondsFromTimeInterval()` converts an `NSTimeInterval` (measured in seconds) into an `mbgl::Duration` (measured in nanoseconds), but `MGLTimeIntervalFromDurationInSeconds()` reinterprets an `mbgl::Duration` (measured in nanoseconds) as an `NSTimeInterval` without changing its value. As a result, these functions don’t round-trip and `-[MGLStyle transitionDuration]` and `-[MGLStyle transitionDelay]` return the wrong values. We should fix `MGLTimeIntervalFromDurationInSeconds()` to convert back into seconds. [This page](http://en.cppreference.com/w/cpp/chrono/duration) has examples for working with `duration`. We should also add a unit test to ensure round-tripping.

Additionally, the functions’ names incorrectly imply that `mbgl::Duration` is measured in seconds. They should be renamed `MGLDurationFromTimeInterval()` and `MGLTimeIntervalFromDuration()`, respectively.

Finally, #5980 declared and defined these functions within `@interface` and `@implementation` blocks, as though they were category methods, but they aren’t even methods in the first place. The `NSDate(MGLAdditions)` category should be removed, because these functions have nothing to do with NSDate. (These functions should remain in NSDate+MGLAdditions.h, because they relate to `NSTimeInterval`, which is declared in NSDate.h.)

/cc @fabian-guerra @incanus

fabian-guerra:
Fixes #8264

### DIFF
--- a/platform/darwin/src/NSDate+MGLAdditions.mm
+++ b/platform/darwin/src/NSDate+MGLAdditions.mm
@@ -1,15 +1,14 @@
 #import "NSDate+MGLAdditions.h"
 
-@implementation NSDate (MGLAdditions)
-
-mbgl::Duration MGLDurationInSecondsFromTimeInterval(NSTimeInterval duration)
+mbgl::Duration MGLDurationFromTimeInterval(NSTimeInterval duration)
 {
     return std::chrono::duration_cast<mbgl::Duration>(std::chrono::duration<NSTimeInterval>(duration));
 }
 
-NSTimeInterval MGLTimeIntervalFromDurationInSeconds(mbgl::Duration duration)
+NSTimeInterval MGLTimeIntervalFromDuration(mbgl::Duration duration)
 {
-    return duration.count();
+    std::chrono::nanoseconds nano(duration.count());
+    std::chrono::seconds sec = std::chrono::duration_cast<std::chrono::seconds>(nano);
+    
+    return sec.count();
 }
-
-@end
--- a/platform/ios/test/MGLNSDateAdditionsTests.mm
+++ b/platform/ios/test/MGLNSDateAdditionsTests.mm
@@ -1,0 +1,26 @@
+#import <XCTest/XCTest.h>
+
+#include <mbgl/util/chrono.hpp>
+#import "../../darwin/src/NSDate+MGLAdditions.h"
+
+using namespace std::chrono_literals;
+
+@interface MGLNSDateAdditionsTests : XCTestCase
+@end
+
+@implementation MGLNSDateAdditionsTests
+
+- (void)testDurationToNSTimeInterval {
+    
+    NSTimeInterval timeInterval = 5;
+    mbgl::Duration duration = MGLDurationFromTimeInterval(timeInterval);
+    NSTimeInterval durationTimeInterval = MGLTimeIntervalFromDuration(duration);
+    
+    mbgl::Duration expectedDuration = 5s;
+    
+    XCTAssertEqual(timeInterval, durationTimeInterval);
+    XCTAssertEqual(timeInterval, MGLTimeIntervalFromDuration(expectedDuration));
+    
+}
+
+@end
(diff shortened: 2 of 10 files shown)


platform/darwin/src/NSDate+MGLAdditions.mm
```diff
@@ -1,15 +1,14 @@
...
 {
     return std::chrono::duration_cast<mbgl::Duration>(std::chrono::duration<NSTimeInterval>(duration));
 }
 
-NSTimeInterval MGLTimeIntervalFromDurationInSeconds(mbgl::Duration duration)
+NSTimeInterval MGLTimeIntervalFromDuration(mbgl::Duration duration)
 {
-    return duration.count();
+    std::chrono::nanoseconds nano(duration.count());
+    std::chrono::seconds sec = std::chrono::duration_cast<std::chrono::seconds>(nano);
```
1ec5: FYI, since this is Objective-C++ code, you could condense these lines somewhat by passing `duration.count()` directly into the `duration_cast`. You can also use `auto` instead of declaring the local variable to be of type `std::chrono::seconds`. What you’ve got is totally reasonable, though.

platform/ios/test/MGLNSDateAdditionsTests.mm
```diff
@@ -0,0 +1,24 @@
+#import <XCTest/XCTest.h>
+
+#include <mbgl/util/chrono.hpp>
+#import "../../darwin/src/NSDate+MGLAdditions.h"
```
1ec5: Since NSDate+MGLAdditions.h is shared between iOS and macOS, the tests should be shared as well. Can you move this file to platform/darwin/test/ and add it to macos.xcodeproj?

platform/ios/test/MGLNSDateAdditionsTests.mm
```diff
@@ -0,0 +1,24 @@
...
+
+@implementation MGLNSDateAdditionsTests
+
+- (void)testDurationToNSTimeInterval {
+    
+    NSTimeInterval timeInterval = 5;
+    mbgl::Duration duration = MGLDurationFromTimeInterval(timeInterval);
+    NSTimeInterval durationTimeInterval = MGLTimeIntervalFromDuration(duration);
+    
+    mbgl::Duration durationInNanoseconds(5000000000);
```
1ec5: Since chrono.hpp imports [`<chrono>`](http://en.cppreference.com/w/cpp/chrono/operator%22%22s), you can write `using namespace std::chrono_literals;` at the top, then use `auto expectedDuration = 5000000000ns;` or even `mbgl::Duration expectedDuration = 5s;`.

1ec5:
Actually, sharing the tests between iOS and macOS is pretty important, so if you could take care of that, that’d be 👌.

platform/darwin/src/NSDate+MGLAdditions.mm
```diff
@@ -1,15 +1,14 @@
...
+mbgl::Duration MGLDurationFromTimeInterval(NSTimeInterval duration)
 {
     return std::chrono::duration_cast<mbgl::Duration>(std::chrono::duration<NSTimeInterval>(duration));
 }
 
-NSTimeInterval MGLTimeIntervalFromDurationInSeconds(mbgl::Duration duration)
+NSTimeInterval MGLTimeIntervalFromDuration(mbgl::Duration duration)
 {
-    return duration.count();
+    std::chrono::nanoseconds nano(duration.count());
```
jfirebaugh: This implementation makes the unsafe assumption that `mbgl::Duration`'s count is in nanoseconds. `mbgl::Duration` is a typedef to [`std::chrono::steady_clock::duration`](http://en.cppreference.com/w/cpp/chrono/steady_clock), and `std::chrono::steady_clock` does not specify a particular tick period.

In addition, this implementation truncates any fractional seconds in the duration. `NSTimeInterval` is a floating-point type, and so the conversion should be expected to preserve fractional seconds.

I believe the correct conversion is as follows (but don't trust me -- write some more tests 😄):

```
NSTimeInterval MGLTimeIntervalFromDuration(mbgl::Duration duration)
{
    std::chrono::duration<NSTimeInterval, std::ratio<1>>(duration).count();
}
```

Issue #7311, opened 2016-12-06
consider dev update notification in IDE


incanus:
Noticed something in the Realm Swift SDK the other day which I thought was pretty cool and could be useful for our mobile SDKs. 

When you fire up your app (at least in Xcode) it lets you know if there is a new version available each and every launch: 

```
2016-12-05 14:20:17.662 AppName[10934:484937] Version 2.1.1 of Realm is now available: https://github.com/realm/realm-cocoa/blob/v2.1.1/CHANGELOG.md
```

This is a nice feature and is both a notification as well as a subtle reminder to keep up to date. We could probably do it on Android, too (though IIRC the console is usually more verbose there and it might be missed). 

@mapbox/ios @mapbox/android

1ec5:
Realm runs a version-checking Web service. Here’s the [client-side code](https://github.com/realm/realm-cocoa/blob/216333cf82e1afcba5e6c4547e6fb983165d510f/Realm/RLMUpdateChecker.mm#L41), which runs only in the Simulator.

incanus:
> runs only in the Simulator

Ooh, neat. 👍

tobrun:
That is pretty cool, for android we could hook into the `BuildConfig.DEBUG`.

Side note: below is the most straightforward way of notifying users directly in the IDE:

<img width="700" alt="screen shot 2016-12-06 at 18 45 08" src="https://cloud.githubusercontent.com/assets/2151639/20936783/3c5b37d0-bbe4-11e6-90aa-00b76fe944db.png">

It seems that this configuration isn't enabled by default for non-Google dependencies and requires the developer to enable this  😞.

friedbunny:
I really like this idea, especially as Realm has implemented it (e.g., simulator-only). Google Maps iOS SDK prints a similar message:

> New version of Google Maps SDK for iOS available: 2.1.28321.0

tobrun:
Noticed today that Realm is showing an update notice it in the gradle console:

<img width="813" alt="screen shot 2017-03-03 at 10 18 09" src="https://cloud.githubusercontent.com/assets/2151639/23556489/d2d97816-fffa-11e6-9a3f-003a27874531.png">

friedbunny:
This happened for iOS in #8282.

tobrun:
[...]

PR #8282, opened 2017-03-05
[ios] Add SDK version update checking/notification

friedbunny:
Implements #7311 for iOS. This adds a check that the developer is using the latest stable release of our iOS SDK and, if not, prints a note to console with a link to the new release.

Inspired by [Realm’s update check](https://github.com/realm/realm-cocoa/blob/216333cf82e1afcba5e6c4547e6fb983165d510f/Realm/RLMUpdateChecker.mm). 🙇 

- Only runs if using the simulator — method is a no-op anywhere else, including Swift Playgrounds.
- Version comparison is extremely simple and ignores semver — if the version strings don’t match, it assumes the developer needs to upgrade.
- Deploy script uploads a text file to http://mapbox.s3.amazonaws.com/mapbox-gl-native/ios/latest_version with the version in format: `x.x.x`
  - @mick Is s3 the best location for this link? Would it be better on https://mapbox.com/ios-sdk?
- Adds a `SDK_UPDATE_AVAILABLE` string that can be localized.

/cc @incanus @tobrun

### DIFF
--- a/platform/ios/resources/ja.lproj/Localizable.strings
+++ b/platform/ios/resources/ja.lproj/Localizable.strings
@@ -43,6 +43,9 @@
 /* Action sheet title */
 "SDK_NAME" = "Mapbox iOS SDK";
 
+/* Developer-only SDK update notification; {latest version, in format x.x.x} */
+"SDK_UPDATE_AVAILABLE" = "現在Mapbox iOS SDK %1$@が入手できる：";
+
 /* Telemetry prompt message */
 "TELEMETRY_DISABLED_MSG" = "You can help make OpenStreetMap and Mapbox maps better by contributing anonymous usage data.";
 
--- a/platform/ios/src/MGLMapView.mm
+++ b/platform/ios/src/MGLMapView.mm
@@ -55,6 +55,7 @@
 #import "MGLStyle_Private.h"
 #import "MGLStyleLayer_Private.h"
 #import "MGLMapboxEvents.h"
+#import "MGLSDKUpdateChecker.h"
 #import "MGLCompactCalloutView.h"
 #import "MGLAnnotationContainerView.h"
 #import "MGLAnnotationContainerView_Private.h"
@@ -345,6 +346,14 @@ public:
         self.styleURL = nil;
     }
     return self;
+}
+
++ (void)initialize
+{
+    if (self == [MGLMapView self])
+    {
+        [MGLSDKUpdateChecker checkForUpdates];
+    }
 }
 
 + (NS_SET_OF(NSString *) *)keyPathsForValuesAffectingStyle
--- a/platform/ios/src/MGLSDKUpdateChecker.mm
+++ b/platform/ios/src/MGLSDKUpdateChecker.mm
@@ -1,0 +1,39 @@
+#import "MGLSDKUpdateChecker.h"
+#import "NSBundle+MGLAdditions.h"
+#import "NSProcessInfo+MGLAdditions.h"
+
+@implementation MGLSDKUpdateChecker
+
++ (void)checkForUpdates {
+#if TARGET_IPHONE_SIMULATOR
+    // Abort if running in a playground.
+    if ([[NSBundle mainBundle].bundleIdentifier hasPrefix:@"com.apple.dt.playground."] ||
+        NSProcessInfo.processInfo.mgl_isInterfaceBuilderDesignablesAgent) {
+        return;
+    }
+
+    NSString *currentVersion = [NSBundle mgl_frameworkInfoDictionary][@"MGLSemanticVersionString"];
+
+    // Skip version check if weʼre doing gl-native development, as the framework
+    // version is `1` until built for packaging.
+    if ([currentVersion isEqualToString:@"1.0.0"]) {
+        return;
+    }
+
+    NSURL *url = [NSURL URLWithString:@"https://mapbox.s3.amazonaws.com/mapbox-gl-native/ios/latest_version"];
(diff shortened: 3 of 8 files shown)


platform/ios/resources/ja.lproj/Localizable.strings
```diff
@@ -43,6 +43,9 @@
 /* Action sheet title */
 "SDK_NAME" = "Mapbox iOS SDK";
 
+/* Developer-only SDK update notification; {latest version, in format x.x.x} */
+"SDK_UPDATE_AVAILABLE" = "現在Mapbox iOS SDK %1$@が入手できる：";
```
1ec5: In general, you should wait until Transifex notices the new base string to translate it over there, rather than checking the translation in ahead of time.

platform/ios/src/MGLMapView.mm
```diff
@@ -564,6 +565,8 @@ - (void)commonInit
     if ([UIApplication sharedApplication].applicationState != UIApplicationStateBackground) {
         [MGLMapboxEvents pushEvent:MGLEventTypeMapLoad withAttributes:@{}];
     }
+
+    [MGLSDKUpdateChecker checkForUpdates];
```
1ec5: Have you considered implementing this check in a `+load` method?

platform/ios/src/MGLSDKUpdateChecker.mm
```diff
@@ -0,0 +1,37 @@
+#import "MGLSDKUpdateChecker.h"
+#import "NSBundle+MGLAdditions.h"
+
+@implementation MGLSDKUpdateChecker
+
++ (void)checkForUpdates {
+#if TARGET_IPHONE_SIMULATOR
+    // Abort if running in a playground.
+    if ([[NSBundle mainBundle].bundleIdentifier hasPrefix:@"com.apple.dt.playground."]) {
```
1ec5: Does this run as part of the Interface Builder designable agent process? You can use `-[NSProcessInfo(MGLAdditions) mgl_isInterfaceBuilderDesignablesAgent]` to check for that condition.

platform/ios/src/MGLMapView.mm
boundsj: I also like the idea of calling this from the `load` or `initialize` class methods. If you do move it, calling from `initialize` [like Realm do](https://github.com/realm/realm-cocoa/blob/8d48a1dd6430f8ed5e30f7fe5125e999c1b24465/Realm/RLMRealm.mm#L136) may be slightly safer since this ends up making a network call but probably fine either way.

platform/ios/resources/ja.lproj/Localizable.strings
friedbunny: I added this as a way to validate my localization implementation — should we take it out?

platform/ios/src/MGLMapView.mm
friedbunny: `+load` in MGLMapView or MGLSDKUpdateChecker?

platform/ios/src/MGLSDKUpdateChecker.mm
friedbunny: Will investigate this — likely no harm in adding such a guard, anyway.

mick:
> @mick Is s3 the best location for this link? Would it be better on https://mapbox.com/ios-sdk?

👍  to the link being to s3

platform/ios/src/MGLMapView.mm
friedbunny: (Responded initially without refreshing and seeing @boundsj comments.)

Sure, we can move this to a new `+load` or `+initialize` in MGLMapView. This update check is clearly not essential, so that’s why I dropped it at the end of `-commonInit`.

platform/ios/src/MGLSDKUpdateChecker.mm
friedbunny: Did not see evidence of the update check being called in IB, but IB is also failing to load our designable anyway... so I added the process exclusion in 470c6c0.

platform/ios/src/MGLMapView.mm
friedbunny: Moved this call to `+[MGLMapView initialize]` in 4d3f390.

platform/ios/resources/ja.lproj/Localizable.strings
1ec5: I'm not sure. I guess we can leave it in and take it out if there's any fallout.

platform/ios/src/MGLMapView.mm
1ec5: `+[MGLOfflineStorage initialize]` or `-[MGLOfflineStorage init]` would also cover the case where the developer uses offline packs early on but buries the map view in a secondary screen. But that's an edge case, so I'm fine with putting it here. Basically we want this check to happen early on while the developer is still paying attention.

1ec5: Fine with putting it in `+[MGLMapView initialize]`, that is.

platform/ios/src/MGLSDKUpdateChecker.mm
1ec5: I later realized we bail in IB at the beginning of `-commonInit`. But this is a good check to add now that we're running code in the initialize method.

friedbunny:
I think we’re ready to go here. Should this have a changelog entry?

1ec5:
Yes, I think it should have a changelog entry.

incanus:
This looks great.

PR #8294, opened 2017-03-06
WIP Avoid using double-sized buffers for source functions

anandthakker:
- Rename `MinMax` attribute to `ZoomInterpolatedAttribute`, signaling
  intent to use it only for composite functions. The idea is that
  source functions will bind the base `Attribute` instead, so as not to
  use double-sized vertex buffers.
- Change the attribute name string in ^ from `a_{name}_minmax` to just
  `a_{name}`, to match the base attributes, and update the shaders to
  match.  (We're only ever binding the base one or the ZoomInterpolated
  one, never both.)
- Change the specific `*PaintPropertyBinder` types and the general `PaintPropertyBinder` type to use the base attribute for non-composite-function values and the `ZoomInterpolatedAttribute<>` for composite-function values.  **🚧 I think this is the part that I don't have quite right.**

cc @jfirebaugh @mollymerp  (note that this PR is against the branch for https://github.com/mapbox/mapbox-gl-native/pull/8267, not master )

### DIFF
--- a/scripts/generate-shaders.js
+++ b/scripts/generate-shaders.js
@@ -41,11 +41,11 @@ require('./style-code');
         return prelude + applyPragmas(source, {
                 define: [
                     "uniform lowp float a_{name}_t;",
-                    "attribute {precision} {a_type} a_{name}_minmax;",
+                    "attribute {precision} {a_type} a_{name};",
                     "varying {precision} {type} {name};"
                 ],
                 initialize: [
-                    "{name} = unpack_mix_{a_type}(a_{name}_minmax, a_{name}_t);"
+                    "{name} = unpack_mix_{a_type}(a_{name}, a_{name}_t);"
                 ]
             });
     }
--- a/src/mbgl/gl/attribute.cpp
+++ b/src/mbgl/gl/attribute.cpp
@@ -38,7 +38,7 @@ void VariableAttributeBinding<T, N>::bind(Context& context,
     MBGL_CHECK_ERROR(glEnableVertexAttribArray(location));
     MBGL_CHECK_ERROR(glVertexAttribPointer(
         location,
-        static_cast<GLint>(N),
+        static_cast<GLint>(attributeSize),
         static_cast<GLenum>(DataTypeOf<T>),
         static_cast<GLboolean>(IsNormalized<T>),
         static_cast<GLsizei>(vertexSize),
--- a/src/mbgl/gl/attribute.hpp
+++ b/src/mbgl/gl/attribute.hpp
@@ -17,10 +17,12 @@ class VariableAttributeBinding {
 public:
     VariableAttributeBinding(BufferID vertexBuffer_,
                              std::size_t vertexSize_,
-                             std::size_t attributeOffset_)
+                             std::size_t attributeOffset_,
+                             std::size_t attributeSize_ = N)
         : vertexBuffer(vertexBuffer_),
           vertexSize(vertexSize_),
-          attributeOffset(attributeOffset_)
+          attributeOffset(attributeOffset_),
+          attributeSize(attributeSize_)
         {}
 
     void bind(Context&, AttributeLocation, optional<VariableAttributeBinding<T, N>>&, std::size_t vertexOffset) const;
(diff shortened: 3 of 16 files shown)


jfirebaugh:
The issue here is that `PaintPropertyBinders::attributeBindings` needs to return a tuple where each element is a `gl::Attribute<...>::Binding`, but `PaintPropertyBinder<...>::AttributeBinding` is now `variant<Attribute::Binding, ZoomInterpolatedAttribute::Binding>`.

I think it will be easier to introduce the needed flexibility at a slightly different spot. I suggest reverting the changes to `{Constant,SourceFunction,CompositeFunction}PaintPropertyBinder` (so that they always return a `ZoomInterpolatedAttribute::Binding`), and instead:

* Permit `VariableAttributeBinding` to be constructed with an attribute size different (less than) the template parameter `N`.
* Expose a helper method `static VariableBinding Attribute<T, N>::variableBinding(...)`, with an `attributeSize` argument that defaults to `N` but can be overridden.
* Use this method, rather than `Attributes::allVariableBindings(...).template get<Attribute>()`, in `SourceFunctionPaintPropertyBinder::attributeBinding`, passing the base attribute dimensions for `attributeSize` rather than the 2x size.

Here's a diff showing the first two of those points:

```diff
diff --git a/src/mbgl/gl/attribute.cpp b/src/mbgl/gl/attribute.cpp
index 2c16dac3f..ff313fdcd 100644
--- a/src/mbgl/gl/attribute.cpp
+++ b/src/mbgl/gl/attribute.cpp
@@ -38,7 +38,7 @@ void VariableAttributeBinding<T, N>::bind(Context& context,
     MBGL_CHECK_ERROR(glEnableVertexAttribArray(location));
     MBGL_CHECK_ERROR(glVertexAttribPointer(
         location,
-        static_cast<GLint>(N),
+        static_cast<GLint>(attributeSize),
         static_cast<GLenum>(DataTypeOf<T>),
         static_cast<GLboolean>(IsNormalized<T>),
         static_cast<GLsizei>(vertexSize),
diff --git a/src/mbgl/gl/attribute.hpp b/src/mbgl/gl/attribute.hpp
index ee418c579..85c1c8aa3 100644
--- a/src/mbgl/gl/attribute.hpp
+++ b/src/mbgl/gl/attribute.hpp
@@ -17,10 +17,12 @@ class VariableAttributeBinding {
 public:
     VariableAttributeBinding(BufferID vertexBuffer_,
                              std::size_t vertexSize_,
-                             std::size_t attributeOffset_)
+                             std::size_t attributeOffset_,
+                             std::size_t attributeSize_ = N)
         : vertexBuffer(vertexBuffer_),
           vertexSize(vertexSize_),
-          attributeOffset(attributeOffset_)
+          attributeOffset(attributeOffset_),
+          attributeSize(attributeSize_)
         {}
 
     void bind(Context&, AttributeLocation, optional<VariableAttributeBinding<T, N>>&, std::size_t vertexOffset) const;
@@ -29,13 +31,15 @@ public:
                            const VariableAttributeBinding& rhs) {
         return lhs.vertexBuffer == rhs.vertexBuffer
             && lhs.vertexSize == rhs.vertexSize
-            && lhs.attributeOffset == rhs.attributeOffset;
+            && lhs.attributeOffset == rhs.attributeOffset
+            && lhs.attributeSize == rhs.attributeSize;
     }
 
 private:
     BufferID vertexBuffer;
     std::size_t vertexSize;
     std::size_t attributeOffset;
+    std::size_t attributeSize;
 };
 
 template <class T, std::size_t N>
@@ -61,6 +65,8 @@ private:
 template <class T, std::size_t N>
 class Attribute {
 public:
+    using ValueType = T;
+    static constexpr size_t Dimensions = N;
     using Value = std::array<T, N>;
 
     using VariableBinding = VariableAttributeBinding<T, N>;
@@ -71,10 +77,19 @@ public:
     using Binding = variant<
         ConstantBinding,
         VariableBinding>;
-    
-    using ValueType = T;
-    
-    static constexpr size_t Dimensions = N;
+
+    template <class Vertex, class DrawMode>
+    static VariableBinding variableBinding(const VertexBuffer<Vertex, DrawMode>& buffer,
+                                           std::size_t attributeIndex,
+                                           std::size_t attributeSize = N) {
+        static_assert(std::is_standard_layout<Vertex>::value, "vertex type must use standard layout");
+        return VariableBinding {
+            buffer.buffer,
+            sizeof(Vertex),
+            Vertex::attributeOffsets[attributeIndex],
+            attributeSize
+        };
+    }
 
     static void bind(Context& context,
                      const Location& location,
@@ -227,15 +242,7 @@ public:
 
     template <class DrawMode>
     static Bindings allVariableBindings(const VertexBuffer<Vertex, DrawMode>& buffer) {
-        static_assert(std::is_standard_layout<Vertex>::value, "vertex type must use standard layout");
-
-        return Bindings {
-            typename As::VariableBinding {
-                buffer.buffer,
-                sizeof(Vertex),
-                Vertex::attributeOffsets[Index<As>]
-            }...
-        };
+        return Bindings { As::variableBinding(buffer, Index<As>)... };
     }
 
     static void bind(Context& context,
```

jfirebaugh:
Pushed a complete implementation of the above ^

anandthakker:
Ahh, yes, this is simpler.

🙇  @jfirebaugh

PR #8296, opened 2017-03-06
Batch update

DBoo:
This branch queues the screen updates for the QML platform if one of two events happen:
1) a certain amount of time elapses (user defined. Example 100ms)
2) number of events reach a critical amount (user defined.  Example 10)

By default batchMode is turned off, only by setting the batch timer to something > 1ms and the cached size also greater then 1 does it do anything.

While you do not see that much savings on a desktop computer, I think it will help out a lot for an embedded computer.

### DIFF
--- a/cmake/core.cmake
+++ b/cmake/core.cmake
@@ -5,6 +5,9 @@ add_library(mbgl-core STATIC
 target_compile_options(mbgl-core
     PRIVATE -fPIC
     PRIVATE -fvisibility-inlines-hidden
+    -Wno-misleading-indentation
+    -Wno-terminate
+    -Wno-error=terminate
 )
 
 target_include_directories(mbgl-core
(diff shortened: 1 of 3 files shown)


tmpsantos:
Hi @DBoo, thanks for you contribution. We moved the QQuick plugin to the Qt 5.9 tree.

https://codereview.qt-project.org/gitweb?p=qt/qtlocation.git;a=tree;f=src/plugins/geoservices/mapboxgl;h=f744a08b4502abc55414a5e6524d6c018d6eecec;hb=refs/heads/5.9

The idea there is to let Qt update using QSGRenderNode and it will be up to Qt to throttle the frame rate.

tmpsantos:
@DBoo I invite you to discuss this change on #qt-mobility @ Freenode and discuss if it could be backported to Qt 5.9.

PR #8297, opened 2017-03-06
[android] - build aar only once for release builds

### LINKED ISSUE #8293: Don't build/publish aar if c++ code fails to compile
It seems that `make apackage` builds an aar file even if building of the c++ code failed. [example](https://www.bitrise.io/build/3314ded43a051536).

screen shot 2017-03-06 at 11 56 51

tobrun:
Closes #8293

### DIFF
--- a/platform/android/bitrise.yml
+++ b/platform/android/bitrise.yml
@@ -181,10 +181,8 @@ workflows:
         inputs:
         - content: |-
             #!/bin/bash
-            echo "Compile libmapbox-gl.so for all supportd abi's:"
-            export BUILDTYPE=Release
-            make apackage
-            cd platform/android && ./gradlew :MapboxGLAndroidSDK:assembleRelease
+            echo "Compile libmapbox-gl.so for all supported abi's and create an aar file:"
+            BUILDTYPE=Release make apackage
     - script:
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -636,7 +636,7 @@ run-android-ui-test-spoon:
 
 .PHONY: apackage
 apackage:
-	cd platform/android && ./gradlew --parallel --max-workers=$(JOBS) assemble$(BUILDTYPE)
+	if [ -d $(“platform/android/MapboxGLAndroidSDK/src/main/jniLibs”) ]; then cd platform/android && ./gradlew --parallel --max-workers=$(JOBS) assemble$(BUILDTYPE); fi
```
kkaefer: Wrong quotes

kkaefer: The existence of this directory is not an indication of whether the build was successful or not. The test would succeed on subsequent builds, even if they failed.

tobrun: This fix was mostly for CI integration where we always start with a clean docker image. Any ideas how to achieve this more cleanly?

tobrun: I will try getting `shell commands to note the error and exit` to work

PR #8299, opened 2017-03-07
[android] - zoom to rounded levels

tobrun:
Follow up from #7630 (but now covers all interactions)
Android equivalent of iOS PR in https://github.com/mapbox/mapbox-gl-native/pull/8027

ezgif com-video-to-gif 12

Besides zooming to rounded zoom levels, this PR also fixes the issue of not getting the final value during OnCameraChange event. Without this change I was receiving zoom levels that just not hit the actual value ( eg. 0.9999988 instead of 1.0).

### DIFF
--- a/platform/android/MapboxGLAndroidSDK/src/main/java/com/mapbox/mapboxsdk/maps/MapView.java
+++ b/platform/android/MapboxGLAndroidSDK/src/main/java/com/mapbox/mapboxsdk/maps/MapView.java
@@ -508,7 +508,7 @@ public class MapView extends FrameLayout {
       if (destroyed) {
         return;
       }
-      mapboxMap.onUpdate();
+      mapboxMap.onUpdateRegionChange();
     }
   }
 
@@ -969,8 +969,10 @@ public class MapView extends FrameLayout {
             mapboxMap.onPostMapReady();
           }
         });
+      } else if (change == DID_FINISH_RENDERING_FRAME || change == DID_FINISH_RENDERING_FRAME_FULLY_RENDERED) {
+        mapboxMap.onUpdateFullyRendered();
       } else if (change == REGION_IS_CHANGING || change == REGION_DID_CHANGE || change == DID_FINISH_LOADING_MAP) {
-        mapboxMap.onUpdate();
+        mapboxMap.onUpdateRegionChange();
       }
     }
 
--- a/platform/android/MapboxGLAndroidSDK/src/main/java/com/mapbox/mapboxsdk/maps/MapboxMap.java
+++ b/platform/android/MapboxGLAndroidSDK/src/main/java/com/mapbox/mapboxsdk/maps/MapboxMap.java
@@ -155,12 +155,16 @@ public final class MapboxMap {
   /**
    * Called when the user
    */
-  void onUpdate() {
-    CameraPosition cameraPosition = transform.invalidateCameraPosition();
-    uiSettings.update(cameraPosition);
-    // FIXME introduce update method with camera position
+  void onUpdateRegionChange() {
     trackingSettings.update();
     annotationManager.update();
+  }
+
+  void onUpdateFullyRendered() {
+    CameraPosition cameraPosition = transform.invalidateCameraPosition();
+    if (cameraPosition != null) {
+      uiSettings.update(cameraPosition);
+    }
   }
 
   // Style
--- a/platform/android/MapboxGLAndroidSDK/src/main/java/com/mapbox/mapboxsdk/maps/NativeMapView.java
+++ b/platform/android/MapboxGLAndroidSDK/src/main/java/com/mapbox/mapboxsdk/maps/NativeMapView.java
@@ -61,6 +61,8 @@ final class NativeMapView {
   // Listener invoked to return a bitmap of the map
   private MapboxMap.SnapshotReadyCallback snapshotReadyCallback;
 
+  private static final int CENTER_XY_VALUE = -1;
+
   //
   // Static methods
   //
@@ -382,14 +384,23 @@ final class NativeMapView {
     if (isDestroyedOn("setZoom")) {
       return;
     }
-    setZoom(zoom, 0);
+    setZoom(zoom, CENTER_XY_VALUE, CENTER_XY_VALUE, 0);
   }
 
-  public void setZoom(double zoom, long duration) {
+  public void setZoom(double zoom, double cx, double cy, long duration) {
     if (isDestroyedOn("setZoom")) {
       return;
     }
-    nativeSetZoom(zoom, duration);
+
+    if (cx != CENTER_XY_VALUE) {
+      cx = cx / pixelRatio;
+    }
+
+    if (cy != CENTER_XY_VALUE) {
+      cy = cy / pixelRatio;
+    }
+
+    nativeSetZoom(zoom, cx, cy, duration);
   }
 
   public double getZoom() {
@@ -1019,7 +1030,7 @@ final class NativeMapView {
 
   private native double nativeGetScale();
 
-  private native void nativeSetZoom(double zoom, long duration);
+  private native void nativeSetZoom(double zoom, double cx, double cy, long duration);
 
   private native double nativeGetZoom();
 
--- a/platform/android/MapboxGLAndroidSDK/src/main/java/com/mapbox/mapboxsdk/maps/Transform.java
+++ b/platform/android/MapboxGLAndroidSDK/src/main/java/com/mapbox/mapboxsdk/maps/Transform.java
@@ -182,11 +182,15 @@ final class Transform implements MapView.OnMapChangedListener {
     // Cancel any animation
     cancelTransitions();
 
-    if (zoomIn) {
-      mapView.scaleBy(2.0, x, y, MapboxConstants.ANIMATION_DURATION);
-    } else {
-      mapView.scaleBy(0.5, x, y, MapboxConstants.ANIMATION_DURATION);
+    CameraPosition cameraPosition = invalidateCameraPosition();
+    if (cameraPosition != null) {
+      zoom(cameraPosition, zoomIn, x, y);
     }
+  }
+
+  private void zoom(@NonNull CameraPosition cameraPosition, boolean zoomIn, float x, float y) {
+    int newZoom = (int) Math.round(cameraPosition.zoom + (zoomIn ? 1 : -1));
+    mapView.setZoom(newZoom, x, y, MapboxConstants.ANIMATION_DURATION);
   }
 
   void setZoom(double zoom) {
--- a/platform/android/MapboxGLAndroidSDKTestApp/src/main/java/com/mapbox/mapboxsdk/testapp/activity/maplayout/DebugModeActivity.java
+++ b/platform/android/MapboxGLAndroidSDKTestApp/src/main/java/com/mapbox/mapboxsdk/testapp/activity/maplayout/DebugModeActivity.java
@@ -5,7 +5,9 @@ import android.support.annotation.NonNull;
 import android.support.design.widget.FloatingActionButton;
 import android.support.v7.app.AppCompatActivity;
 import android.view.View;
+import android.widget.TextView;
 
+import com.mapbox.mapboxsdk.camera.CameraPosition;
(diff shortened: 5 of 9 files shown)


platform/android/MapboxGLAndroidSDK/src/main/java/com/mapbox/mapboxsdk/maps/MapView.java
```diff
@@ -966,8 +966,10 @@ public void run() {
             mapboxMap.onPostMapReady();
           }
         });
+      } else if (change == DID_FINISH_RENDERING_FRAME || change == DID_FINISH_RENDERING_FRAME_FULLY_RENDERED) {
```
Guardiola31337: What do you think about decomposing the conditional to make the intention clearer?
We normally end up with long methods and the length of a method is in itself a factor that makes it harder to read (using conditions increase the difficulty even more).
The problem usually lies in the fact that the code tells you what happens but can easily obscure why it happens.
Decomposing the conditional you highlight the condition and make it clearly what you are branching on. You also highlight the reason for the branching.

platform/android/MapboxGLAndroidSDK/src/main/java/com/mapbox/mapboxsdk/maps/MapboxMap.java
```diff
@@ -154,14 +154,18 @@ void onPostMapReady() {
...
-    CameraPosition cameraPosition = transform.invalidateCameraPosition();
-    uiSettings.update(cameraPosition);
-    // FIXME introduce update method with camera position
+  void onUpdateRegionChange() {
     trackingSettings.update();
     annotationManager.update();
   }
 
+  void onUpdateFullyRendered() {
+    CameraPosition cameraPosition = transform.invalidateCameraPosition();
```
Guardiola31337: What do you think of moving these lines into a private method?

- It increases the chances that other methods can use a method when the method is finely grained.
- It allows the higher-level methods to read more like a series of comments (public methods should look like pseudo code, without language statements, etc.).

platform/android/MapboxGLAndroidSDK/src/main/java/com/mapbox/mapboxsdk/maps/NativeMapView.java
```diff
@@ -381,14 +381,14 @@ public void setZoom(double zoom) {
     if (isDestroyedOn("setZoom")) {
       return;
     }
-    setZoom(zoom, 0);
+    setZoom(zoom, -1, -1, 0);
```
Guardiola31337: What do you think of replacing those magic numbers with a symbolic constant in order to increase readability?
Magic numbers are numbers with special values that usually are not obvious.
If the numbers might ever change, making the change is a nightmare.

platform/android/MapboxGLAndroidSDK/src/main/java/com/mapbox/mapboxsdk/maps/NativeMapView.java
```diff
@@ -381,14 +381,14 @@ public void setZoom(double zoom) {
...
   }
 
-  public void setZoom(double zoom, long duration) {
+  public void setZoom(double zoom, double cx, double cy, long duration) {
     if (isDestroyedOn("setZoom")) {
       return;
     }
-    nativeSetZoom(zoom, duration);
+    nativeSetZoom(zoom, cx == -1 ? cx : cx / pixelRatio, cy == -1 ? cy : cy / pixelRatio, duration);
```
Guardiola31337: What about extracting `cx == -1 ? cx : cx / pixelRatio, cy == -1 ? cy : cy / pixelRatio` into a query or a explaining variable?
It's really hard to read it and therefore to understand what's going on.

platform/android/MapboxGLAndroidSDK/src/main/java/com/mapbox/mapboxsdk/maps/Transform.java
```diff
@@ -182,10 +182,10 @@ void zoom(boolean zoomIn, float x, float y) {
     // Cancel any animation
     cancelTransitions();
 
-    if (zoomIn) {
-      mapView.scaleBy(2.0, x, y, MapboxConstants.ANIMATION_DURATION);
-    } else {
-      mapView.scaleBy(0.5, x, y, MapboxConstants.ANIMATION_DURATION);
+    CameraPosition cameraPosition = invalidateCameraPosition();
```
Guardiola31337: What do you think of extracting this code into a `private` method?

platform/android/MapboxGLAndroidSDKTestApp/src/main/java/com/mapbox/mapboxsdk/testapp/activity/maplayout/DebugModeActivity.java
```diff
@@ -47,9 +49,21 @@ protected void onCreate(Bundle savedInstanceState) {
...
         mapboxMap = map;
+
+        mapboxMap.getUiSettings().setZoomControlsEnabled(true);
+
+        // show current zoom level on screen
+        final TextView textView = (TextView) findViewById(R.id.textZoom);
+        mapboxMap.setOnCameraChangeListener(new MapboxMap.OnCameraChangeListener() {
+          @Override
+          public void onCameraChange(CameraPosition position) {
+            textView.setText(String.format("Zoom : %s", position.zoom));
```
Guardiola31337: What do you think of replacing `"Zoom : %s"` magic number?

tobrun: Strings aren't numbers AFAIK

platform/android/MapboxGLAndroidSDK/src/main/java/com/mapbox/mapboxsdk/maps/MapView.java
tobrun: this is high traffic code, every small performance gain we can get, we should leverage

platform/android/MapboxGLAndroidSDK/src/main/java/com/mapbox/mapboxsdk/maps/NativeMapView.java
tobrun: let me patch this up

platform/android/MapboxGLAndroidSDK/src/main/java/com/mapbox/mapboxsdk/maps/NativeMapView.java
tobrun: let me patch this up

platform/android/MapboxGLAndroidSDK/src/main/java/com/mapbox/mapboxsdk/maps/Transform.java
tobrun: let me patch this up

platform/android/MapboxGLAndroidSDKTestApp/src/main/java/com/mapbox/mapboxsdk/testapp/activity/maplayout/DebugModeActivity.java
Guardiola31337: Better said, [magic numbers](http://wiki.c2.com/?MagicNumber) are literals, so strings are included ;)

platform/android/MapboxGLAndroidSDK/src/main/java/com/mapbox/mapboxsdk/maps/MapboxMap.java
tobrun: not sure what you mean, this is as simplistic as a method can be:

```java
  void onUpdateFullyRendered() {
    CameraPosition cameraPosition = transform.invalidateCameraPosition();
    if (cameraPosition != null) {
      uiSettings.update(cameraPosition);
    }
 
```

moving  one line of code to another method is overthinking it + this is high traffic code, being able to not have additional method invocations is a win.

platform/android/MapboxGLAndroidSDK/src/main/java/com/mapbox/mapboxsdk/maps/MapView.java
Guardiola31337: What do you mean? Do you think that adding new methods could affect the performance?

platform/android/MapboxGLAndroidSDK/src/main/java/com/mapbox/mapboxsdk/maps/MapboxMap.java
Guardiola31337: I strive to leave `public` and `protected` methods as clean as I can. BTW, in this particular case as you pointed out is not that bad so 👍

platform/android/MapboxGLAndroidSDK/src/main/java/com/mapbox/mapboxsdk/maps/MapView.java
tobrun: yes, very much, would love to explore this by profiling and measuring the difference. For now, to get the release out the door, would love to get this PR merged. Feel free to ticket this separately so we don't lose track of it.

platform/android/MapboxGLAndroidSDKTestApp/src/main/java/com/mapbox/mapboxsdk/testapp/activity/maplayout/DebugModeActivity.java
tobrun: when it comes to readability, it think showing where %s is located for String.format is much clearer else you need to jump to the constant field and check what the value is. Putting it in the constant is making readability worse. The correct solution would be to put this in strings.xml as android studio shows you the value directly in the IDE.

platform/android/MapboxGLAndroidSDK/src/main/java/com/mapbox/mapboxsdk/maps/MapboxMap.java
tobrun: not sure if I follow this is neither a public or protected method..

Guardiola31337: Or `default`. Anyway, looks good to me.

platform/android/MapboxGLAndroidSDKTestApp/src/main/java/com/mapbox/mapboxsdk/testapp/activity/maplayout/DebugModeActivity.java
Guardiola31337: Agreed, but you could always split it into more literals which opens the possibility of explaining your intention clearer.
BTW, as you mentioned putting it in `strings.xml` is also a great solution.

johanlunds:
In Android 5.0.0-beta.3 `onCameraChange(CameraPosition position)` is continously called. Could it be related to this change?

It does not happen on beta2. It does happen in today's SNAPSHOT.

johanlunds:
I think it has something to do with rendering the map.

When I comment out `Layer.setProperties()` that we continously call to emulate an animation there are far less `onCameraChange` calls.

Still, the problem does not go away completely. Since in our `onCameraChange` callback, we load new data over the network and then call `GeoJsonSource.setGeoJson()`, the map seems to render again thereby triggering `onCameraChange` again (and the endless loop continues).

johanlunds:
This temporary workaround seems to work for now:

```java
    private CameraPosition mLastRefreshCameraPosition;

    @Override
    public void onCameraChange(CameraPosition position) {
        if (!position.equals(mLastRefreshCameraPosition)) {
            mLastRefreshCameraPosition = position;
            refreshData();
        }
    }
```

zugaldia:
@johanlunds Thanks for the report. I want to make sure this doesn't fall through the cracks, could you check if this issue is still happening on 5.0 released today https://github.com/mapbox/mapbox-gl-native/issues/8452 and if so, open a new ticket about it? Thanks!

PR #8302, opened 2017-03-07
[core] - avoid animating with infinite zoom out factor for Transform::FlyTo

### LINKED ISSUE #8300: Map goes blank when animating dezoom
**Platform:** Android
**Mapbox SDK version:** 4.2.2

### Steps to trigger behavior

 1. Create a minimal app (largely inspired by the sample app):

```java
public class MainActivity extends Activity {

    MapView mapView;
    MapboxMap map;

    @Override
    protected void onCreate(Bundle savedInstanceState) {
        super.onCreate(savedInstanceState);
        MapboxAccountManager.start(getApplicationContext(), "mapbox token goes here");

        setContentView(R.layout.activity_main);

        mapView = (MapView) findViewById(R.id.mapview);
        mapView.getMapAsync(new OnMapReadyCallback() {
            @Override
            public void onMapReady(MapboxMap map) {
                MainActivity.this.map = map;

                double distance = 0.001;
                LatLngBounds bounds = new LatLngBounds.Builder()
                    .include(new LatLng(48.866800199999986 - distance, 2.3536729199999993 - distance))
                    .include(new LatLng(48.866800199999986 + distance, 2.3536729199999993 + distance))
                    .build();
                map.moveCamera(CameraUpdateFactory.newLatLngBounds(bounds, 10));
                map.setOnMapClickListener(new MapboxMap.OnMapClickListener() {
                    @Override
                    public void onMapClick(@NonNull LatLng point) {
                        // works:
                        // MainActivity.this.map.moveCamera(CameraUpdateFactory.newLatLngZoom(new LatLng(48.866800199999986, 2.3536729199999993), 0));

                        // doesn't work:
                        // (coordinates could be something else)
                        MainActivity.this.map.animateCamera(CameraUpdateFactory.newLatLngZoom(new LatLng(48.866800199999986, 2.3536729199999993), 0), 100);
                    }
                });
            }
        });
    }

    // Add the mapView lifecycle to the activity's lifecycle methods
    @Override
    public void onResume() {
        super.onResume();
        mapView.onResume();
    }

    @Override
    public void onPause() {
        super.onPause();
        mapView.onPause();
    }

    @Override
    public void onLowMemory() {
        super.onLowMemory();
        mapView.onLowMemory();
    }

    @Override
    protected void onDestroy() {
        super.onDestroy();
        mapView.onDestroy();
    }

    @Override
    protected void onSaveInstanceState(Bundle outState) {
        super.onSaveInstanceState(outState);
        mapView.onSaveInstanceState(outState);
    }
}
```

 2. Tap the map

### Expected behavior

Tapping the map should make us dezoom to the earth view

### Actual behavior

Tapping the map causes the map to disappear

### Notes

Upon investigation:
- debugging shows that camera position target LatLng has NaN longitude
- using moveCamera instead of animateCamera works

Based on these observations, my guess would be that something goes wrong near https://github.com/mapbox/mapbox-gl-native/blob/master/src/mbgl/map/transform.cpp#L169

Thank's in advance for your help!

tobrun:
This PR fixes the issue if you try to flyTo a camera position that is zoomed out but has the (allmost) the same target LatLng.

closes #8300

### DIFF
--- a/src/mbgl/map/transform.cpp
+++ b/src/mbgl/map/transform.cpp
@@ -241,16 +241,19 @@ void Transform::flyTo(const CameraOptions &camera, const AnimationOptions &anima
         return std::log(std::sqrt(b * b + 1) - b);
     };
 
+    /// r₀: Zoom-out factor during ascent.
+    double r0 = r(0);
+    /// r1: Zoom-in factor during descent
+    double r1 = r(1);
+
     // When u₀ = u₁, the optimal path doesn’t require both ascent and descent.
     bool isClose = std::abs(u1) < 0.000001;
-    // Perform a more or less instantaneous transition if the path is too short.
-    if (isClose && std::abs(w0 - w1) < 0.000001) {
+    // Perform a more or less instantaneous transition if the path is too short or zoom in/out factor is infinte
+    if ((isClose && std::abs(w0 - w1) < 0.000001) || std::isinf(r0) || std::isinf(r1)) {
         easeTo(camera, animation);
         return;
     }
-
-    /// r₀: Zoom-out factor during ascent.
-    double r0 = r(0);
+    
     /** w(s): Returns the visible span on the ground, measured in pixels with
         respect to the initial scale.
 
@@ -268,7 +271,7 @@ void Transform::flyTo(const CameraOptions &camera, const AnimationOptions &anima
     };
     /// S: Total length of the flight path, measured in ρ-screenfuls.
     double S = (isClose ? (std::abs(std::log(w1 / w0)) / rho)
-                : ((r(1) - r0) / rho));
+                : ((r1 - r0) / rho));
 
     Duration duration;
     if (animation.duration) {


src/mbgl/map/transform.cpp
```diff
@@ -241,8 +241,8 @@ void Transform::flyTo(const CameraOptions &camera, const AnimationOptions &anima
         return std::log(std::sqrt(b * b + 1) - b);
     };
 
-    // When u₀ = u₁, the optimal path doesn’t require both ascent and descent.
-    bool isClose = std::abs(u1) < 0.000001;
+    // When u₀ = u₁ or zoom out factor = inf, the optimal path doesn’t require both ascent and descent.
+    bool isClose = std::abs(u1) < 0.000001 || isinf(r(0));
```
tobrun: Actually now thinking that this should not be added as part of `bool isClose` but should be added to the if check below..

src/mbgl/map/transform.cpp
```diff
@@ -243,8 +243,8 @@ void Transform::flyTo(const CameraOptions &camera, const AnimationOptions &anima
 
     // When u₀ = u₁, the optimal path doesn’t require both ascent and descent.
     bool isClose = std::abs(u1) < 0.000001;
-    // Perform a more or less instantaneous transition if the path is too short.
-    if (isClose && std::abs(w0 - w1) < 0.000001) {
+    // Perform a more or less instantaneous transition if the path is too short or zoom out factor is inf.
+    if ((isClose && std::abs(w0 - w1) < 0.000001) || isinf(r(0))) {
```
1ec5: What causes the ascent factor to be infinite? Do we need to check the descent factor too?

mrlem: Pardon me if miss something obvious, but for me, this code doesn't compile (it does if I use explicitely std::isinf, as is done for abs)

1ec5: Good catch @mrlem – we should use `std::` explicitly. 👍

tobrun: @1ec5 good call on descent factor, let me try reproducing the reverse interaction. 

`What causes the ascent factor to be infinite`: a division by 0 in the following code

```
    auto r = [=](double i) {
        /// bᵢ
        double b = (w1 * w1 - w0 * w0 + (i ? -1 : 1) * rho2 * rho2 * u1 * u1) / (2 * (i ? w1 : w0) * rho2 * u1);
        return std::log(std::sqrt(b * b + 1) - b);
    };
```

tobrun: Wasn't able to reproducing a infinite descent factor but I don't see an issue adding a check for this either. I added this in eeaefae.

1ec5: > a division by 0 in the following code

So that means either `w1` or `rho2` or `u1` is 0, right? Sorry to keep going here; just want to make sure we aren’t running into something more fundamental.

src/mbgl/map/transform.cpp
```diff
@@ -243,8 +243,8 @@ void Transform::flyTo(const CameraOptions &camera, const AnimationOptions &anima
 
     // When u₀ = u₁, the optimal path doesn’t require both ascent and descent.
     bool isClose = std::abs(u1) < 0.000001;
-    // Perform a more or less instantaneous transition if the path is too short.
-    if (isClose && std::abs(w0 - w1) < 0.000001) {
+    // Perform a more or less instantaneous transition if the path is too short or zoom in/out factor is infinte
+    if ((isClose && std::abs(w0 - w1) < 0.000001) || std::isinf(r(0)) || std::isinf(r(1))) {
```
kkaefer: Can you please explain why this is a correct solution? It feels like putting band-aid on the problem without fixing the underlying problem.

tobrun: Been looking back and forth between the [original paper](https://www.win.tue.nl/~vanwijk/zoompan.pdf) and our implementation. One difference that I noticed is that we added a check for small changes, as these aren't ideal for the algorithm. Instead we are currently performing an easeTo instead. I don't feel that I'm adding a band-aid to the problem, I'm hardening the existing one.

screen shot 2017-03-13 at 14 31 36

Above is a normal situation of a flyTo implementation but the use-case described in
https://github.com/mapbox/mapbox-gl-native/issues/8300 is having a origin and target camera position that is almost the same. This isn't fully captured by current bandaid solution. The algorithm seem to generate -infinite zoom out values as a result of log(o) in following code with i=1:

```cpp
/** rᵢ: Returns the zoom-out factor at one end of the animation.
  @param i 0 for the ascent or 1 for the descent. */
auto r = [=](double i) {
            /// bᵢ
      double b = (w1 * w1 - w0 * w0 + -1 * rho2 * rho2 * u1 * u1) / (2 * (i ? w1 : w0) * rho2 * u1);
      return std::log(std::sqrt(b * b + 1) - b);
};
```

My proposal is to optimise the bandaid solution as this interaction is not ideal for a flyTo interaction.

kkaefer: 👍

kkaefer: Micro-optimization: Instead of computing `r(0)` and `r(1)` here, can we split up this conditional return and store the results of `r(0)` and `r(1)` into a constant; they are recomputed further down.

tobrun: :+1:

1ec5: There’s already an `r0` variable just below.

1ec5:
I’m able to reproduce the issue in macosapp using the parameters in https://github.com/mapbox/mapbox-gl-native/issues/8300#issuecomment-286068097, but I’m noticing that `r(0)` is -12.395270530594363, not infinity (and the transition isn’t “close” either). It may be relevant that a map view frame 683 points wide by 722 points high results in an `mbgl::TransformState::minZoom` of 0.49585502688717098 rather than 0, due to constraining the map at the world bounds.

tobrun:
@1ec5 the device I was using had a `mgbl::TransfrormState::minZooom =  0.2360141919000848`.

1ec5:
My point is that the minZoom depends on the aspect ratio of the map view. For a non-full-screen map view, the aspect ratio could be just about anything.

tobrun:
Closing as suggested band aid solution didn't capture all use-cases

PR #8304, opened 2017-03-07
[ios] Disable/enable related gesture recognizer when disabling/enabling user interactivity.

### LINKED ISSUE #7217: MapBox in UIScrollView, scrolling behavior
**Platform: iOS 10 
**Mapbox SDK version: 3.3.7

We have an IOS project where we have migrated to the MapBox iOS SDK, currently on version 3.3.7

We have identified one behavior related to the scrolling that differs from the MKMapView. We want to present the map without letting the user pan (scroll) the map. For this we set the property self.mapView.allowsScrolling to NO. This works as intended, the user is not able to pan the map.

However, the problem we have is that the MGLMapView is presented in a UIScrollView. When we have disabled the scrolling as above, the map seems to ignore the swipe gesture that would normally be detected by the UIScrollView.

To scroll the view presented in the scrollview you need to touch the screen either above or below the MGLMapView. Swiping on the map do not cause the UIScrollView to move.

This works with the regular MKMapView, setting the scrolling to disabled there, still let it send the received gestures to the underlying UIScrollView.

davidchiles:
Fixes #7217

Added an example behavior of MGLMapView as a subview of UIScrollView in
iOS app.

### DIFF
--- a/platform/ios/app/MBXEmbeddedMapViewController.m
+++ b/platform/ios/app/MBXEmbeddedMapViewController.m
@@ -1,0 +1,89 @@
+#import "MBXEmbeddedMapViewController.h"
+#import <Mapbox/Mapbox.h>
+
+typedef NS_ENUM(NSInteger, MBXEmbeddedControl) {
+    MBXEmbeddedControlZoom = 0,
+    MBXEmbeddedControlScroll,
+    MBXEmbeddedControlRotation,
+    MBXEmbeddedControlPitch
+};
+
+@interface MBXEmbeddedMapViewController () <UIScrollViewDelegate>
+
+@property (weak, nonatomic) IBOutlet UIScrollView *scrollView;
+@property (weak, nonatomic) IBOutlet MGLMapView *mapView;
+
+@end
+
+@implementation MBXEmbeddedMapViewController
+
+- (void)viewDidLoad {
+    [super viewDidLoad];
+
+    self.scrollView.delegate = self;
+    self.scrollView.contentSize = self.view.bounds.size;
+}
+
+- (IBAction)didSwitch:(UISwitch *)controlSwitch {
+    [self switchControl:controlSwitch.tag];
+}
+
+- (IBAction)rotation:(UIRotationGestureRecognizer *)rotationGesture {
+    self.mapView.transform = CGAffineTransformRotate(rotationGesture.view.transform, rotationGesture.rotation);
+}
+
+- (void)switchControl:(MBXEmbeddedControl) control {
+    switch (control) {
+        case MBXEmbeddedControlZoom:
+            self.mapView.zoomEnabled = !self.mapView.zoomEnabled;
+            break;
+        case MBXEmbeddedControlScroll:
+            self.mapView.scrollEnabled = !self.mapView.scrollEnabled;
+            break;
+        case MBXEmbeddedControlRotation:
+            self.mapView.rotateEnabled = !self.mapView.rotateEnabled;
+            break;
+        case MBXEmbeddedControlPitch:
+            self.mapView.pitchEnabled = !self.mapView.pitchEnabled;
+            break;
+    }
+}
+
+- (BOOL)statusForControl:(MBXEmbeddedControl) control {
+    switch (control) {
+        case MBXEmbeddedControlZoom:
+            return self.mapView.zoomEnabled;
+        case MBXEmbeddedControlScroll:
+            return self.mapView.scrollEnabled;
+        case MBXEmbeddedControlRotation:
+            return self.mapView.rotateEnabled;
+        case MBXEmbeddedControlPitch:
+            return self.mapView.pitchEnabled;
+    }
+}
+
+#pragma mark UIScrollViewDelegate methods
+
+- (UIView *)viewForZoomingInScrollView:(UIScrollView *)scrollView {
+    return self.mapView;
+}
+
+#pragma mark Class method
+
++ (NSString *)titleForControl:(MBXEmbeddedControl) control {
+    switch (control) {
+        case MBXEmbeddedControlZoom:
+            return @"Zoom Enabled";
+        case MBXEmbeddedControlScroll:
+            return @"Scroll Enabled";
+            break;
+        case MBXEmbeddedControlRotation:
+            return @"Rotation Enabled";
+            break;
+        case MBXEmbeddedControlPitch:
+            return @"Pitch Enabled";
+            break;
+    }
+}
+
+@end
--- a/platform/ios/app/MBXViewController.m
+++ b/platform/ios/app/MBXViewController.m
@@ -5,6 +5,7 @@
 #import "MBXOfflinePacksTableViewController.h"
 #import "MBXAnnotationView.h"
 #import "MBXUserLocationAnnotationView.h"
+#import "MBXEmbeddedMapViewController.h"
 
 #import <Mapbox/Mapbox.h>
 
@@ -79,6 +80,7 @@ typedef NS_ENUM(NSInteger, MBXSettingsMiscellaneousRows) {
     MBXSettingsMiscellaneousWorldTour,
     MBXSettingsMiscellaneousCustomUserDot,
     MBXSettingsMiscellaneousShowZoomLevel,
+    MBXSettingsMiscellaneousScrollView,
     MBXSettingsMiscellaneousPrintLogFile,
     MBXSettingsMiscellaneousDeleteLogFile,
 };
@@ -350,6 +352,7 @@ typedef NS_ENUM(NSInteger, MBXSettingsMiscellaneousRows) {
                 @"Start World Tour",
                 [NSString stringWithFormat:@"%@ Custom User Dot", (_customUserLocationAnnnotationEnabled ? @"Disable" : @"Enable")],
                 [NSString stringWithFormat:@"%@ Zoom Level", (_showZoomLevelEnabled ? @"Hide" :@"Show")],
+                @"Embedded Map View",
             ]];
 
             if (self.debugLoggingEnabled)
@@ -613,6 +616,13 @@ typedef NS_ENUM(NSInteger, MBXSettingsMiscellaneousRows) {
                     self.showZoomLevelEnabled = !self.showZoomLevelEnabled;
                     self.hudLabel.hidden = !self.showZoomLevelEnabled;
                     self.reuseQueueStatsEnabled = NO;
+                    break;
+                }
+                case MBXSettingsMiscellaneousScrollView:
+                {
+                    UIStoryboard *storyboard = [UIStoryboard storyboardWithName:@"Main" bundle:nil];
+                    MBXEmbeddedMapViewController *embeddedMapViewController = (MBXEmbeddedMapViewController *)[storyboard instantiateViewControllerWithIdentifier:@"MBXEmbeddedMapViewController"];
+                    [self.navigationController pushViewController:embeddedMapViewController animated:YES];
                     break;
                 }
                 default:
--- a/platform/ios/src/MGLMapView.mm
+++ b/platform/ios/src/MGLMapView.mm
@@ -240,6 +240,8 @@ public:
 @property (nonatomic) UIActionSheet *attributionSheet;
 @property (nonatomic, readwrite) MGLStyle *style;
 @property (nonatomic) UITapGestureRecognizer *singleTapGestureRecognizer;
+@property (nonatomic) UITapGestureRecognizer *doubleTap;
+@property (nonatomic) UITapGestureRecognizer *twoFingerTap;
 @property (nonatomic) UIPanGestureRecognizer *pan;
 @property (nonatomic) UIPinchGestureRecognizer *pinch;
 @property (nonatomic) UIRotationGestureRecognizer *rotate;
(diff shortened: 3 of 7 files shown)


boundsj:
Thanks! I think this is a good implementation of the suggestion in https://github.com/mapbox/mapbox-gl-native/issues/7217#issuecomment-271988753 to fully disable gesture recognizers instead of short circuiting their handlers. When they are disabled, then containing view recognizers (like a UIScrollView) work as expected (as illustrated by the new demo in the iosapp in this PR).

The full list is of gesture recognizers affected with this proposal are:

* UIPanGestureRecognizer `pan` handled by handlePanGesture  and guarded by isScrollEnabled
* UIPanGestureRecognizer `twoFingerDrag` handled by handleTwoFingerDragGesture and guarded by isZoomEnabled
* UIPinchGestureRecognizer `pinch` handled by handlePinchGesture  and guarded by isZoomEnabled
* UITapGestureRecognizer `doubleTap` handled by handleDoubleTapGesture and guarded by isZoomEnabled
* UILongPressGestureRecognizer `quickZoom` handled by handleQuickZoomGesture and guarded by isZoomEnabled
* UITapGestureRecognizer `twoFingerTap` handled by handleTwoFingerTapGesture and guarded by isPitchEnabled
* UIRotationGestureRecognizer `rotate` handled by handleRotateGesture and guarded by isRotateEnabled

All of the handlers noted in this list have guards like `if ( ! self.isScrollEnabled) return;`. I think we can probably remove those now since the handler methods will never get called since the gesture recognizers are fully disabled.

cc @incanus @friedbunny @1ec5

platform/ios/app/MBXViewController.m
```diff
@@ -79,8 +80,10 @@ typedef NS_ENUM(NSInteger, MBXSettingsMiscellaneousRows) {
     MBXSettingsMiscellaneousWorldTour,
     MBXSettingsMiscellaneousCustomUserDot,
     MBXSettingsMiscellaneousShowZoomLevel,
+    MBXSettingsMiscellaneousScrollView,
     MBXSettingsMiscellaneousPrintLogFile,
-    MBXSettingsMiscellaneousDeleteLogFile,
+    MBXSettingsMiscellaneousDeleteLogFile
+
```
incanus: Nix this space.

platform/ios/app/MBXViewController.m
```diff
@@ -79,8 +80,10 @@ typedef NS_ENUM(NSInteger, MBXSettingsMiscellaneousRows) {
     MBXSettingsMiscellaneousWorldTour,
     MBXSettingsMiscellaneousCustomUserDot,
     MBXSettingsMiscellaneousShowZoomLevel,
+    MBXSettingsMiscellaneousScrollView,
     MBXSettingsMiscellaneousPrintLogFile,
-    MBXSettingsMiscellaneousDeleteLogFile,
+    MBXSettingsMiscellaneousDeleteLogFile
```
incanus: Our internal style is to retain commas in these situations to make it easier to rearrange and add new entries.

platform/ios/app/MBXViewController.m
```diff
@@ -350,6 +353,7 @@ - (void)dismissSettings:(__unused id)sender
                 @"Start World Tour",
                 [NSString stringWithFormat:@"%@ Custom User Dot", (_customUserLocationAnnnotationEnabled ? @"Disable" : @"Enable")],
                 [NSString stringWithFormat:@"%@ Zoom Level", (_showZoomLevelEnabled ? @"Hide" :@"Show")],
+                @"Embeded Map View"
```
incanus: Another trailing comma to add.

platform/ios/app/MBXViewController.m
```diff
@@ -615,6 +619,14 @@ - (void)performActionForSettingAtIndexPath:(NSIndexPath *)indexPath
                     self.reuseQueueStatsEnabled = NO;
                     break;
                 }
+                case MBXSettingsMiscellaneousScrollView:
+                {
+                    UIStoryboard *storyboard = [UIStoryboard storyboardWithName:@"Main" bundle:nil];
+                    MBXEmbededMapViewController *embeddedMapViewController = (MBXEmbededMapViewController *)[storyboard instantiateViewControllerWithIdentifier:@"MBXEmbededMapViewController"];
+                    [self.navigationController pushViewController:embeddedMapViewController animated:YES];
+
```
incanus: Remove blank line.

platform/ios/src/MGLMapView.mm
```diff
@@ -238,9 +238,11 @@ @interface MGLMapView () <UIGestureRecognizerDelegate,
 @property (nonatomic) UIActionSheet *attributionSheet;
 @property (nonatomic, readwrite) MGLStyle *style;
 @property (nonatomic) UITapGestureRecognizer *singleTapGestureRecognizer;
+@property (nonatomic) UITapGestureRecognizer *twoFingerTap;
 @property (nonatomic) UIPanGestureRecognizer *pan;
 @property (nonatomic) UIPinchGestureRecognizer *pinch;
 @property (nonatomic) UIRotationGestureRecognizer *rotate;
+@property (nonatomic) UITapGestureRecognizer *doubleTap;
```
incanus: Keep the taps all together?

platform/ios/src/MGLMapView.mm
```diff
@@ -536,7 +538,7 @@ - (void)commonInit
         _quickZoom = [[UILongPressGestureRecognizer alloc] initWithTarget:self action:@selector(handleQuickZoomGesture:)];
         _quickZoom.numberOfTapsRequired = 1;
         _quickZoom.minimumPressDuration = 0;
-        [_quickZoom requireGestureRecognizerToFail:doubleTap];
+        [_quickZoom requireGestureRecognizerToFail:self.doubleTap];
```
incanus: Maintain style with above and use `_doubleTap`?

platform/ios/app/MBXEmbededMapViewController.m
```diff
@@ -0,0 +1,153 @@
...
+#pragma mark UIScrollViewDelegate methods
+
+- (UIView *)viewForZoomingInScrollView:(UIScrollView *)scrollView
+{
+    return self.mapView;
+}
+
+#pragma mark Class method
+
+
```
incanus: Extraneous blank lines.

platform/ios/app/MBXEmbededMapViewController.m
```diff
@@ -0,0 +1,153 @@
...
++ (NSString *)titleForControl:(MBXEmbeddedControl) control {
```
incanus: Need consistent bracket style, which is a new line as per rest of SDK and other parts of this file.

platform/ios/app/MBXEmbededMapViewController.m
```diff
@@ -0,0 +1,153 @@
...
+        
+    }
+}
+
+- (void)rotation:(UIRotationGestureRecognizer *)rotationGesture
+{
+    self.mapView.transform = CGAffineTransformRotate(rotationGesture.view.transform, rotationGesture.rotation);
+}
+
+- (void)switchContorl:(MBXEmbeddedControl) control {
```
incanus: Typo.

platform/ios/app/MBXEmbededMapViewController.m
```diff
@@ -0,0 +1,153 @@
...
+    
+    UIStackView *stackView = [[UIStackView alloc] initWithArrangedSubviews:items];
+    stackView.alignment = UIStackViewAlignmentFill;
+    stackView.distribution = UIStackViewDistributionFillEqually;
+    stackView.frame = CGRectMake(0, CGRectGetMaxY(self.navigationController.navigationBar.frame), CGRectGetWidth(self.view.frame), 65);
+    
+    [self.view addSubview:stackView];
+}
+
+- (void)didSwitch:(UISwitch *)sw {
```
incanus: We always try to write out variable names.

platform/ios/app/MBXEmbededMapViewController.m
```diff
@@ -0,0 +1,153 @@
...
+    CGFloat widthScale = self.view.bounds.size.width / CGRectGetWidth(mapRect);
+    CGFloat heightScale = self.view.bounds.size.height / CGRectGetHeight(mapRect);
+    CGFloat minScale = MIN(widthScale, heightScale);
+    
+    scrollView.maximumZoomScale = minScale*10;
+    scrollView.minimumZoomScale = minScale;
+    
+    //Create list of all the possible control titles
+    NSMutableArray <UIView*>*items = [[NSMutableArray alloc] init];
+    for(NSInteger index =0; index < 4; index++) {
```
incanus: Consistent spacing.

platform/ios/app/MBXEmbededMapViewController.m
```diff
@@ -0,0 +1,153 @@
...
+
+typedef NS_ENUM(NSInteger, MBXEmbeddedControl) {
+    MBXEmbeddedControlZoom = 0,
+    MBXEmbeddedControlScroll,
+    MBXEmbeddedControlRotation,
+    MBXEmbeddedControlPitch
+};
+
+
+@interface MBXEmbededMapViewController () <UIScrollViewDelegate>
```
incanus: Class name is misspelled: `Embeded`.

platform/ios/app/MBXEmbededMapViewController.h
```diff
@@ -0,0 +1,13 @@
+//
+//  MBXEmbededMapViewController.h
+//  ios
+//
+//  Created by David Chiles on 3/7/17.
+//  Copyright © 2017 Mapbox. All rights reserved.
+//
```
incanus: Remove comment header & personal identification.

platform/ios/app/MBXEmbeddedMapViewController.m
```diff
@@ -0,0 +1,146 @@
...
+        view.backgroundColor = [UIColor colorWithWhite:1 alpha:0.5];
+        UILabel *label = [[UILabel alloc] initWithFrame:CGRectZero];
+        label.font = [UIFont systemFontOfSize:12];
+        label.translatesAutoresizingMaskIntoConstraints = NO;
+        label.text = [[self class] titleForControl:index];
+        UISwitch *switchControl = [[UISwitch alloc] initWithFrame:CGRectZero];
+        switchControl.tag = index;
+        [switchControl addTarget:self action:@selector(didSwitch:) forControlEvents:UIControlEventValueChanged];
+        switchControl.translatesAutoresizingMaskIntoConstraints = NO;
+        switchControl.on = [self statusForControl:index];
```
1ec5: The body of this `for` loop seems like a good candidate to break out into a separate method. Considering that there are only four switch controls in all, perhaps it makes sense to set each switch’s title explicitly, track them with separate properties, and hook them up to separate action methods. Someone theoretically adding a “Roll Enabled” switch (!) would have fewer methods to update.

platform/ios/app/MBXEmbeddedMapViewController.m
```diff
@@ -0,0 +1,146 @@
...
+    
+    UIRotationGestureRecognizer *gestureRecognizer = [[UIRotationGestureRecognizer alloc] initWithTarget:self action:@selector(rotation:)];
+    
+    UIScrollView *scrollView = [[UIScrollView alloc] initWithFrame:self.view.frame];
+    scrollView.delegate = self;
+    [self.view addSubview:scrollView];
+    [scrollView addGestureRecognizer:gestureRecognizer];
+    
+    CGRect mapRect = CGRectMake(0, 0, CGRectGetWidth(scrollView.frame) * 2, CGRectGetHeight(scrollView.frame) * 2);
+    self.mapView = [[MGLMapView alloc] initWithFrame:mapRect];
```
1ec5: Since the view controller itself is already defined in the storyboard, I wonder if it would make the demo clearer if we moved some of this setup to the storyboard as well. Then most of this class’s code would relate to the switches’ behavior instead of the layout code.

friedbunny: Aye, this is a bit awkward — let’s go one way or the other.

platform/ios/app/MBXEmbeddedMapViewController.m
```diff
@@ -0,0 +1,188 @@
...
+    [super viewDidLoad];
+    
+    UIRotationGestureRecognizer *gestureRecognizer = [[UIRotationGestureRecognizer alloc] initWithTarget:self action:@selector(rotation:)];
+    
+    UIScrollView *scrollView = [[UIScrollView alloc] initWithFrame:self.view.frame];
+    scrollView.delegate = self;
+    [self.view addSubview:scrollView];
+    [scrollView addGestureRecognizer:gestureRecognizer];
+    
+    CGRect mapRect = CGRectMake(0, 0, CGRectGetWidth(scrollView.frame) * 2, CGRectGetHeight(scrollView.frame) * 2);
```
friedbunny: I’m not following the purpose of a gigantic scrollview — is this a common use case that we’re demoing? It seems like multiple smaller maps within a regular viewport frame (with only a vertical scroll axis) is a potential use case that would demonstrate these conditionally disabled gestures in a more straightforward way.

boundsj: It's true that https://github.com/mapbox/mapbox-gl-native/issues/7217 demonstrated the need for only vertical (and perhaps horizontal) gesture overrides. However, the solution implemented in this PR technically allows for all possible UIScrollView gestures to function instead of the built in map view gestures. This demo may not be realistic but it is comprehensive and I don't think it would make sense to throw away the pinch and rotate parts in iosapp.

That said, a more realistic example in mapbox/ios-sdk-examples would be helpful, I think.

platform/ios/app/MBXEmbeddedMapViewController.m
davidchiles: I'm not very fluent in storyboard so I can just remove the view controller from there.

platform/ios/app/MBXEmbeddedMapViewController.m
```diff
@@ -0,0 +1,188 @@
...
+    MBXEmbeddedControlZoom = 0,
+    MBXEmbeddedControlScroll,
+    MBXEmbeddedControlRotation,
+    MBXEmbeddedControlPitch
+};
+
+
+@interface MBXEmbeddedMapViewController () <UIScrollViewDelegate>
+
+@property (nonatomic, strong) MGLMapView *mapView;
```
incanus: Again, super minor, but `strong` is the default so can be left out.

platform/ios/app/MBXEmbeddedMapViewController.m
```diff
@@ -0,0 +1,188 @@
...
+                                                      constant:0.0]];
+    
+    return view;
+}
+
++ (NSString *)titleForControl:(MBXEmbeddedControl) control
+{
+    switch (control) {
+        case MBXEmbeddedControlZoom:
+            return @"Zoom Enabled";
```
incanus: Missing `break`?

platform/ios/app/MBXViewController.m
```diff
@@ -350,6 +352,7 @@ - (void)dismissSettings:(__unused id)sender
                 @"Start World Tour",
                 [NSString stringWithFormat:@"%@ Custom User Dot", (_customUserLocationAnnnotationEnabled ? @"Disable" : @"Enable")],
                 [NSString stringWithFormat:@"%@ Zoom Level", (_showZoomLevelEnabled ? @"Hide" :@"Show")],
+                @"Embeded Map View",
```
incanus: Typo.

platform/ios/src/MGLMapView.mm
```diff
@@ -2135,6 +2137,33 @@ - (void)emptyMemoryCache
...
+    _rotateEnabled = rotateEnabled;
+    self.rotate.enabled = rotateEnabled;
+}
+
+- (void)setPitchEnabled:(BOOL)pitchEnabled
+{
+    _pitchEnabled = pitchEnabled;
+    self.twoFingerDrag.enabled = pitchEnabled;
+}
+
```
incanus: Love these routines BTW. Great way to see what the setters impact across the view.

platform/ios/app/MBXEmbeddedMapViewController.m
1ec5: A `break` isn’t needed after a `return`.

boundsj:
Adding this to the 3.6.0 milestone.

boundsj:
https://github.com/mapbox/mapbox-gl-native/pull/8304/commits/2cefd09426e522147aaebdecb908ae49938ee00d makes the changes suggested above. This includes moving all view setup boilerplate to the storyboard so the salient bits of the example are clearer.

cc @incanus @1ec5

platform/ios/app/MBXEmbeddedMapViewController.m
```diff
@@ -0,0 +1,90 @@
...
+@end
+
+@implementation MBXEmbeddedMapViewController
+
+- (void)viewDidLoad {
+    [super viewDidLoad];
+
+    self.scrollView.delegate = self;
+    self.scrollView.contentSize = CGSizeMake(CGRectGetWidth(self.view.bounds),
+                                             CGRectGetHeight(self.view.bounds));
```
1ec5: self.view.bounds.size would have the same effect.

PR #8306, opened 2017-03-08
For data-driven paint setters, transition immediately to target value

### LINKED ISSUE #8237: Changing symbol layer to SourceFunction with runtime styling fails to update rendered layer.
**Platform: macos**
**Mapbox SDK version: master**

### Steps to trigger behavior

1. Update the demo app by adding the following lines to `manipulateStyle`:
```objc
    MGLSymbolStyleLayer *labelLayer = (MGLSymbolStyleLayer *)[self.mapView.style layerWithIdentifier:@"country-label-lg"];
    MGLStyleValue *labelColorFunction = [MGLStyleValue<NSColor *> valueWithInterpolationMode:MGLInterpolationModeExponential sourceStops:@{
        @0.0: [MGLStyleValue<NSColor *> valueWithRawValue:[NSColor redColor]],
        @3.0: [MGLStyleValue<NSColor *> valueWithRawValue:[NSColor yellowColor]],
        @6.0: [MGLStyleValue<NSColor *> valueWithRawValue:[NSColor blackColor]],
    } attributeName: @"scalerank" options:nil];
    labelLayer.textColor = labelColorFunction;
```

2. Run the app and select Debug > Manipulate Style

### Expected behavior

The country labels change from black to (data-driven) color with no further user action.

### Actual behavior

Even after the paint transition ends, the country labels remain black.  However, if you use the layer list in the left-hand panel to hide and then re-show the `country-label-lg` layer, then it has the correct colors when it comes back.

anandthakker:
Closes #8237

### DIFF
--- a/src/mbgl/style/paint_property.hpp
+++ b/src/mbgl/style/paint_property.hpp
@@ -44,11 +44,18 @@ public:
     template <class Evaluator>
     auto evaluate(const Evaluator& evaluator, TimePoint now) {
         auto finalValue = value.evaluate(evaluator);
+        
         if (!prior) {
             // No prior value.
             return finalValue;
         } else if (now >= end) {
             // Transition from prior value is now complete.
+            prior = {};
+            return finalValue;
+        } else if (value.isDataDriven()) {
+            // Transitions to data-driven properties are not supported.
+            // We snap immediately to the data-driven value so that, when we perform layout,
+            // we see the data-driven function and can use it to populate vertex buffers.
             prior = {};
             return finalValue;
         } else if (now < begin) {
@@ -60,7 +67,7 @@ public:
             return util::interpolate(prior->get().evaluate(evaluator, now), finalValue, util::DEFAULT_TRANSITION_EASE.solve(t, 0.001));
         }
     }
-
+    
     bool hasTransition() const {
         return bool(prior);
     }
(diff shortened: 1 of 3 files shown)


anandthakker:
The code block in #8237 can be used for manual testing/debugging.

@jfirebaugh surprisingly, using ☝️ , I am seeing the data-driven value show up at the _end_ of the transition... not sure why exactly -- I'll look into it tomorrow, but in the meantime I think this is ready for a first 👓

jfirebaugh:
I imagined this being a one character change, `a` ⇢ `b` [here](https://github.com/mapbox/mapbox-gl-native/blob/f591421aaaf698c23ccff37a03d03c87eaa0d988/src/mbgl/style/possibly_evaluated_property_value.hpp#L60). Is that something you tried? Does it not work?

anandthakker:
I considered that, but the problem was transitions w a delay -- for those,
the interpolation wouldn't start immediately, so we'd be back to the
problem of getting the prior value during layout.

src/mbgl/style/paint_property.hpp
```diff
@@ -44,13 +44,20 @@ class UnevaluatedPaintProperty {
     template <class Evaluator>
     auto evaluate(const Evaluator& evaluator, TimePoint now) {
         auto finalValue = value.evaluate(evaluator);
+
```
jfirebaugh: Nit: revert the whitespace changes here and below.

PR #8308, opened 2017-03-08
Make annotation view rotation alignment configurable

eimantas:
This PR adds `rotatesWithMap` property to `MGLAnnotationView`. This
property, when set to `YES` fixes the annotation to a map such that view
follows map's rotation angle. This is useful when user wants to display
rotation-dependent annotations (e.g. sector lights).

Defaults to `NO`. Encoded.

### DIFF
--- a/platform/ios/src/MGLAnnotationView.h
+++ b/platform/ios/src/MGLAnnotationView.h
@@ -152,7 +152,7 @@ typedef NS_ENUM(NSUInteger, MGLAnnotationViewDragState) {
  The default value of this property is `NO`. Set this property to `YES` if the
  view’s rotation is important.
  */
-@property (nonatomic, assign) BOOL rotatesWithMap;
+@property (nonatomic, assign) BOOL rotatesToMatchCamera;
 
 #pragma mark Managing the Selection State
 
--- a/platform/ios/src/MGLAnnotationView.h
+++ b/platform/ios/src/MGLAnnotationView.h
@@ -143,7 +143,7 @@ typedef NS_ENUM(NSUInteger, MGLAnnotationViewDragState) {
 
 /**
  A Boolean value that determines whether the annotation view rotates together
- with the mapview.
+ with the map.
 
  When the value of this property is `YES` and the map is rotated, the annotation
  view rotates. This is also the behavior of `MGLAnnotationImage` objects. When the
--- a/platform/ios/src/MGLAnnotationView.h
+++ b/platform/ios/src/MGLAnnotationView.h
@@ -141,6 +141,19 @@ typedef NS_ENUM(NSUInteger, MGLAnnotationViewDragState) {
  */
 @property (nonatomic, assign) BOOL scalesWithViewingDistance;
 
+/**
+ A Boolean value that determines whether the annotation view rotates together
+ with the mapview.
+
+ When the value of this property is `YES` and the map is rotated, the annotation
+ view rotates. This is also the behavior of `MGLAnnotationImage` objects. When the
+ value of this property is `NO` the annotation has its rotation angle fixed.
+
+ The default value of this property is `NO`. Set this property to `YES` if the
+ view’s rotation is important.
+ */
+@property (nonatomic, assign) BOOL rotatesWithMap;
+
 #pragma mark Managing the Selection State
 
 /**
--- a/platform/ios/src/MGLAnnotationView.mm
+++ b/platform/ios/src/MGLAnnotationView.mm
@@ -103,6 +103,7 @@
 
     super.center = center;
     [self updateScaleTransformForViewingDistance];
+    [self updateRotateTransform];
 }
 
 - (void)setScalesWithViewingDistance:(BOOL)scalesWithViewingDistance
--- a/platform/ios/src/MGLAnnotationView.mm
+++ b/platform/ios/src/MGLAnnotationView.mm
@@ -11,6 +11,7 @@
 
 @property (nonatomic, readwrite, nullable) NSString *reuseIdentifier;
 @property (nonatomic, readwrite) CATransform3D lastAppliedScaleTransform;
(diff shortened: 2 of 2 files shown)


platform/ios/src/MGLAnnotationView.h
```diff
@@ -141,6 +141,19 @@ typedef NS_ENUM(NSUInteger, MGLAnnotationViewDragState) {
...
+ with the mapview.
+
+ When the value of this property is `YES` and the map is rotated, the annotation
+ view rotates. This is also the behavior of `MGLAnnotationImage` objects. When the
+ value of this property is `NO` the annotation has its rotation angle fixed.
+
+ The default value of this property is `NO`. Set this property to `YES` if the
+ view’s rotation is important.
+ */
+@property (nonatomic, assign) BOOL rotatesWithMap;
```
1ec5: #5245 at one point called this property `rotatesToMatchCamera`, which I think would do a better job of describing exactly what it does. (After all, technically you could rotate the MGLMapView itself in 3D space! :joy:)

eimantas:
I actually spent some time trying to match the tilting too, but didn't manage to get the transform matrix calculations right so just came with this simple implementation since that was enough for the project I was working on (no tilting was enabled on the map).

The exposed transform from core to sdk would definitely make things easier.

platform/ios/src/MGLAnnotationView.h
```diff
@@ -141,6 +141,19 @@ typedef NS_ENUM(NSUInteger, MGLAnnotationViewDragState) {
  */
 @property (nonatomic, assign) BOOL scalesWithViewingDistance;
 
+/**
+ A Boolean value that determines whether the annotation view rotates together
+ with the mapview.
```
1ec5: Nit: replace “mapview” with “map”, since the `direction` property reflects the rotation of the content within the map view, not the map view itself.

platform/ios/src/MGLAnnotationView.mm
```diff
@@ -146,6 +151,29 @@ - (void)updateScaleTransformForViewingDistance
...
+- (void)setRotatesToMatchCamera:(BOOL)rotatesToMatchCamera
+{
+    if (_rotatesToMatchCamera != rotatesToMatchCamera)
+    {
+        _rotatesToMatchCamera = rotatesToMatchCamera;
+        [self updateRotateTransform];
+    }
+}
+
+- (void)updateRotateTransform
```
1ec5: As far as I can tell, this method only gets called when `rotatesToMatchCamera` is set explicitly by application code. But the annotation view’s transform matrix is always overridden whenever the map viewport changes. `-updateScaleTransformForViewingDistance` gets called by `-setCenter:`, which coincidentally happens to be the method that `-[MGLMapView update AnnotationViews]` calls to ~~clobber~~ reposition the annotation view to match the map. We could merge this method and `-updateScaleTransformForViewingDistance` into an `-updateTransform` method that comes up with the canonical transform to apply.

platform/ios/src/MGLAnnotationView.h
```diff
@@ -141,6 +141,19 @@ typedef NS_ENUM(NSUInteger, MGLAnnotationViewDragState) {
...
+
+ When the value of this property is `YES` and the map is rotated, the annotation
+ view rotates. This is also the behavior of `MGLAnnotationImage` objects. When the
+ value of this property is `NO` the annotation has its rotation angle fixed.
+
+ The default value of this property is `NO`. Set this property to `YES` if the
+ view’s rotation is important.
+ */
+@property (nonatomic, assign) BOOL rotatesToMatchCamera;
```
1ec5: Thanks for making this change. FYI, I was torn between suggesting `rotatesToMatchCamera` and suggesting a `rotationAlignment` property of type `MGLIconRotationAlignment`, for consistency with MGLSymbolStyleLayer. But I think a Boolean `rotatesToMatchCamera` will end up being more intuitive for developers, since we have no need to support an “auto” value here.

platform/ios/src/MGLAnnotationView.mm
eimantas: Right. I was copy-pasting the code from my private repository that was used with the project. Forgot to add the call to `[self updateRotateTransform];` from `setCenter`.

eimantas:
I guess this won't make it to 3.5.0?

1ec5:
Unfortunately, we’re unable to take additional features for v3.5.0 due to the condensed schedule for that release, but I put this PR on the v3.6.0 milestone to make sure it doesn’t fall off our radar.

boundsj:
@eimantas @1ec5 

This PR is in a strange state where GitHub seems to think the fork has been deleted (see `from unknown repository` above) [but I don't think that is actually the case](https://github.com/eimantas/mapbox-gl-native/tree/eima-rotatesWithMap).

I just rebased on top of the `release-ios-v3.6.0-android-v5.1.0` branch and added https://github.com/mapbox/mapbox-gl-native/commit/10c4acb3ff1e479c129fb50f0ae6fea75bbc113e to fix an issue I was seeing where, after the transform was applied, the view frame became `<MBXAnnotationView: 0x7fe22cc6e380; frame = (-8.98847e+307 -8.98847e+307; 1.79769e+308 1.79769e+308);` -- this caused a crash when the map view tried to move the annotation view when a user panned the map.

After my commit (https://github.com/mapbox/mapbox-gl-native/commit/10c4acb3ff1e479c129fb50f0ae6fea75bbc113e), the behavior looks like the gif below which I think is what we want.

rotating_annotations

I'd like to get another round of review on this and merge it next week. However, with the PR in the state it is in now, I think I may need to redo all of these changes in a new PR or @eimantas might need to resubmit the PR from a new fork.

boundsj:
Continuing in https://github.com/mapbox/mapbox-gl-native/pull/9147

Issue #7702, opened 2017-01-12
Support for compiling with clang++ >= 3.9.1


springmeyer:
We currently hit errors due to warnings with clang++ 3.9.1. This is because the `enum.hpp` header emits a warning with `-Wundefined-var-template` which is newly added to clang++ and enabled by default: http://llvm.org/viewvc/llvm-project?view=revision&revision=266719.

@kkaefer @jfirebaugh - any ideas to work around? Or should we suppress?

```
../../../include/mbgl/util/enum.hpp:17:32: error: instantiation of variable 'mbgl::Enum<mbgl::style::SymbolPlacementType>::begin' required here, but no definition is available [-Werror,-Wundefined-var-template]
        auto it = std::find_if(begin, end, [&] (const auto& v) { return t == v.first; });
                               ^
../../../src/mbgl/style/conversion/stringify.hpp:52:28: note: in instantiation of member function 'mbgl::Enum<mbgl::style::SymbolPlacementType>::toString' requested here
    writer.String(Enum<T>::toString(v));
                           ^
../../../src/mbgl/style/conversion/stringify.hpp:236:39: note: in instantiation of function template specialization 'mbgl::style::conversion::stringify<rapidjson::Writer<rapidjson::GenericStringBuffer<rapidjson::UTF8<char>, rapidjson::CrtAllocator>, rapidjson::UTF8<char>, rapidjson::UTF8<char>, rapidjson::CrtAllocator, 0>, mbgl::style::SymbolPlacementType, void>' requested here
    v.evaluate([&] (const auto& v_) { stringify(writer, v_); });
                                      ^
../../../mason_packages/headers/variant/1.1.4/include/mapbox/variant.hpp:304:51: note: in instantiation of member function 'mapbox::util::detail::dispatcher<const (lambda at ../../../src/mbgl/style/conversion/stringify.hpp:236:16) &, mapbox::util::variant<mbgl::style::Undefined, mbgl::style::SymbolPlacementType, mbgl::style::Function<mbgl::style::SymbolPlacementType> >, void, mbgl::style::SymbolPlacementType, mbgl::style::Function<mbgl::style::SymbolPlacementType> >::apply_const' requested here
            return dispatcher<F, V, R, Types...>::apply_const(v, std::forward<F>(f));
                                                  ^
[...]

PR #8313, opened 2017-03-08
[build] Add -Wno-undefined-var-template to fix builds with clang 3.9.1

jfirebaugh:
Stopgap for #7702.


jfirebaugh:
It occurs to me that this won't prevent the warning from occurring when downstream projects compile files that `#include` Mapbox GL headers with clang 3.9.1. So I'm going to try to actually fix the warning instead.

PR #8316, opened 2017-03-08
[core] De-duplicate shader prelude source

jfirebaugh:
```
  +2.1% +3.38Ki [None]                  +3.18Ki  +1.9%
  +0.1% +2.41Ki __TEXT,__text           +2.41Ki  +0.1%
  +0.1%    +392 __TEXT,__gcc_except_tab    +392  +0.1%
  +0.1%     +16 __DATA,__data               +16  +0.1%

 -20.4% -22.2Ki __TEXT,__cstring        -22.2Ki -20.4%

  -0.4% -16.0Ki TOTAL                   -16.2Ki  -0.4%
```


src/mbgl/programs/program.hpp
```diff
@@ -54,7 +55,7 @@ class Program {
     }
 
     static std::string vertexSource(const ProgramParameters& parameters) {
-        return pixelRatioDefine(parameters) + Shaders::vertexSource;
+        return pixelRatioDefine(parameters) + shaders::vertexPrelude + Shaders::vertexSource;
```
kkaefer: OpenGL supports taking multiple strings and compile them to a single shader. This could be a micro optimization (so we don't have to do string concatenation)

kkaefer:
Another optimization we could do in this area is to run our shader source code to a pre-compiler that normalizes and minifies shaders, akin to CSS/JS minification. We've previously used https://github.com/aras-p/glsl-optimizer

PR #8321, opened 2017-03-08
Darwin static libs: don't rely on static variable initialization

kkaefer:
We should explicitly initialize rather than use a global static variables for the main thread's `RunLoop`.

When linking Mapbox GL as a static library, the object file (MGLFoundation.o) containing the global static variable initialization was not linked into the resulting binary since it was not referenced. This resulted in the RunLoop not being initialized in the main thread.

Reverts 03a14ff0003e976a4ded70d284bc80adf54bc6c9 in  https://github.com/mapbox/mapbox-gl-native/pull/8084

### DIFF
--- a/platform/darwin/src/MGLFoundation.mm
+++ b/platform/darwin/src/MGLFoundation.mm
@@ -1,4 +1,6 @@
 #import "MGLFoundation_Private.h"
 
 /// Initializes the run loop shim that lives on the main thread.
-mbgl::util::RunLoop mgl_runLoop;
+void MGLInitializeRunLoop() {
+    static mbgl::util::RunLoop runLoop;
+}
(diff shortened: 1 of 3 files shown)


kkaefer:
This patch has the added benefit that we don't initialize anything during library load if Mapbox GL isn't used during the application workflow.

platform/ios/src/MGLMapView.mm
```diff
@@ -392,6 +392,8 @@ - (IBAction)reloadStyle:(__unused id)sender {
 
 - (void)commonInit
 {
+    MGLInitializeRunLoop();
```
1ec5: Since `-commonInit` forces the shared MGLOfflineStorage object to initialize, is it necessary to also initialize it here?

kkaefer: Good call, we can remove these.

PR #8325, opened 2017-03-08
Pin Android NDK version

kkaefer:
The [Android NDK r14 was released a few days ago](https://developer.android.com/ndk/downloads/revision_history.html) and ships with Clang 3.9, which [seems to have a regression around undefined variable template error message](https://github.com/mapbox/mapbox-gl-native/issues/7702).


kkaefer:
hardcoded local path in the previous commit.

PR #8332, opened 2017-03-09
[android] - NPX mapZoomButtonController

### LINKED ISSUE #8326: MapBoxView does crash on visibility="gone"
**Platform:** Android
**Mapbox SDK version:** 5.0.0-beta.1

### Steps to trigger behavior

 1. Do something like
```
   <com.mapbox.mapboxsdk.maps.MapView
        android:id="@+id/mapView"
        android:layout_width="fill_parent"
        android:layout_height="fill_parent"
        android:background="@color/transparent"
        android:visibility="gone" <<<<<<<<<
      />
```

 2. Try to open that view
 3. You get a crash, with a NPE in

MapView:
```
  // Called when view is hidden and shown
  @Override
  protected void onVisibilityChanged(@NonNull View changedView, int visibility) {
    if (isInEditMode()) {
      return;
    }
    >>>>>the mapZoomButtonController is NULL
    mapZoomButtonController.setVisible(visibility == View.VISIBLE);
  }
```

### Expected behavior
View is hidden
### Actual behavior
NPE crashing

tobrun:
closes #8326

### DIFF
--- a/platform/android/MapboxGLAndroidSDK/src/main/java/com/mapbox/mapboxsdk/maps/MapView.java
+++ b/platform/android/MapboxGLAndroidSDK/src/main/java/com/mapbox/mapboxsdk/maps/MapView.java
@@ -521,13 +521,15 @@ public class MapView extends FrameLayout {
   @CallSuper
   protected void onDetachedFromWindow() {
     super.onDetachedFromWindow();
-    mapZoomButtonController.setVisible(false);
+    if (mapZoomButtonController != null) {
+      mapZoomButtonController.setVisible(false);
+    }
   }
 
   // Called when view is hidden and shown
   @Override
   protected void onVisibilityChanged(@NonNull View changedView, int visibility) {
-    if (isInEditMode()) {
+    if (isInEditMode() || mapZoomButtonController == null) {
       return;
     }
     mapZoomButtonController.setVisible(visibility == View.VISIBLE);


platform/android/MapboxGLAndroidSDK/src/main/java/com/mapbox/mapboxsdk/maps/MapView.java
```diff
@@ -521,13 +521,15 @@ public void onSurfaceTextureUpdated(SurfaceTexture surface) {
   @CallSuper
   protected void onDetachedFromWindow() {
     super.onDetachedFromWindow();
-    mapZoomButtonController.setVisible(false);
+    if (mapZoomButtonController != null) {
```
Guardiola31337: What do you think of moving the check into a `private` method?
- It increases the chances that other methods can use a method when the method is finely grained.
- It allows the higher-level methods to read more like a series of comments (`public` methods should look like pseudo code, without language statements, etc.).

platform/android/MapboxGLAndroidSDK/src/main/java/com/mapbox/mapboxsdk/maps/MapView.java
```diff
@@ -521,13 +521,15 @@ public void onSurfaceTextureUpdated(SurfaceTexture surface) {
...
+      mapZoomButtonController.setVisible(false);
+    }
   }
 
   // Called when view is hidden and shown
   @Override
   protected void onVisibilityChanged(@NonNull View changedView, int visibility) {
-    if (isInEditMode()) {
+    if (isInEditMode() || mapZoomButtonController == null) {
```
Guardiola31337: What do you think of introducing a _explaining variable_ here?
_Explaining variables_ are particularly valuable with conditional logic in which it is useful to take each clause of a condition and explain what the condition means with a well-named temp.
What about putting the result of the expression in a temp variable with a name that explains the purpose?

platform/android/MapboxGLAndroidSDK/src/main/java/com/mapbox/mapboxsdk/maps/MapView.java
tobrun: Guidelines indicate avoiding internal getters and setters: https://developer.android.com/training/articles/perf-tips.html#GettersSetters

platform/android/MapboxGLAndroidSDK/src/main/java/com/mapbox/mapboxsdk/maps/MapView.java
tobrun: I'm not seeing the added benefit here for null checking as:

```
boolean mapboxZoomButtonControllerIsNotNull = mapbooxZoomButtonController == null;
if(isInEditMode() || mapboxZoomButtonControllerIsNotNull){
   ///
}
```

imo this is harder to read as a simple null check, happy to hear the thoughts of @mapbox/android on this.

zugaldia: Agree with the value of explaining variables but at the same time I also think the code is well readable as it is.

platform/android/MapboxGLAndroidSDK/src/main/java/com/mapbox/mapboxsdk/maps/MapView.java
tobrun: Also happy to hear @mapbox/android on this as well

platform/android/MapboxGLAndroidSDK/src/main/java/com/mapbox/mapboxsdk/maps/MapView.java
tobrun: just to clarify, I also completely agree if there is more to it than just null checking but this just feels weird to write/read:
```
boolean something = somehtinToTest == null;
```

platform/android/MapboxGLAndroidSDK/src/main/java/com/mapbox/mapboxsdk/maps/MapView.java
tobrun: If we would go this route, we should  move MapboxZoomButtonController to a seperate component.

platform/android/MapboxGLAndroidSDK/src/main/java/com/mapbox/mapboxsdk/maps/MapView.java
Guardiola31337: Expressions can become very complex and hard to read. You quickly end up with a pretty long condition. What about extracting the condition into a query?

tobrun: agree that they can become complex. The way how I see it, one condition ok, two condition ok, three condition -> time to apply suggestion from https://github.com/mapbox/mapbox-gl-native/pull/8332#discussion_r105248115.

Guardiola31337: 👍  You follow the [rule of three](https://en.wikipedia.org/wiki/Rule_of_three_(computer_programming)).

tobrun: lol, wasn't aware of that 😅 this just felt natural to me

Guardiola31337: > Never go to bed without learning something new.

👍

PR #8339, opened 2017-03-10
Raise exception when MGLAnnotationView.annotation property is nil

fabian-guerra:
Raises an exception outlining the problem.

### DIFF
--- a/platform/ios/src/MGLAnnotationView.mm
+++ b/platform/ios/src/MGLAnnotationView.mm
@@ -239,7 +239,10 @@
     }
     else if (dragState == MGLAnnotationViewDragStateCanceling)
     {
-        NSAssert(self.annotation, @"Annotation property should not be nil.");
+        if (!self.annotation) {
+            [NSException raise:@"MGLInvalidAnnotationException"
+                        format:@"Annotation property should not be nil."];
+        }
         self.panGestureRecognizer.enabled = NO;
         self.longPressRecognizer.enabled = NO;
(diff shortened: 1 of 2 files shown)


platform/ios/src/MGLAnnotationView.mm
```diff
@@ -239,7 +239,10 @@ - (void)setDragState:(MGLAnnotationViewDragState)dragState animated:(BOOL)animat
     }
     else if (dragState == MGLAnnotationViewDragStateCanceling)
     {
-        NSAssert(self.annotation, @"Annotation property should not be nil.");
+        if (!self.annotation) {
+            [NSException raise:@"MGLInvalidAnnotationException"
```
1ec5: Let’s stick to standard exception names, except for similar exceptions strewn throughout the codebase, such as `MGLAbstractClassException`. In this case, `NSInvalidArgumentException` seems reasonable. (We are using ad-hoc names in other places, but we should clean them up at some point.)

platform/ios/src/MGLAnnotationView.mm
```diff
@@ -239,7 +239,10 @@ - (void)setDragState:(MGLAnnotationViewDragState)dragState animated:(BOOL)animat
     }
     else if (dragState == MGLAnnotationViewDragStateCanceling)
     {
-        NSAssert(self.annotation, @"Annotation property should not be nil.");
+        if (!self.annotation) {
+            [NSException raise:@"NSInvalidArgumentException"
```
1ec5: [`NSInvalidArgumentException`](https://developer.apple.com/reference/foundation/nsinvalidargumentexception) is a constant.

PR #8340, opened 2017-03-10
Enable Android NDK r14

kkaefer:
Now that https://github.com/mapbox/mapbox-gl-native/pull/8336 has landed, we can upgrade the Android NDK to the recently released version r14, which uses Clang 3.9.


kkaefer:
Mason side is happening at https://github.com/mapbox/mason/pull/384

Issue #8192, opened 2017-02-23
MGLInterpolationMode docs should clarify relative differences


friedbunny:
v3.5.0 deprecates [`+[MGLStyleValue valueWithStops:]`](https://www.mapbox.com/ios-sdk/api/3.5.0-beta.1/Classes/MGLStyleValue.html#/c:objc(cs)MGLStyleValue(cm)valueWithStops:), replacing it with [`+[MGLStyleValue valueWithInterpolationMode:cameraStops:options:]`](https://www.mapbox.com/ios-sdk/api/3.5.0-beta.1/Classes/MGLStyleValue.html#/c:objc(cs)MGLStyleValue(cm)valueWithInterpolationMode:cameraStops:options:).

Besides being a much more verbose initializer, this change adds ambiguity: what [`MGLInterpolationMode`](https://www.mapbox.com/ios-sdk/api/3.5.0-beta.1/Enums/MGLInterpolationMode.html) was being used in v3.4 and which of the four options should I use now?

We should clarify:

- What the default interpolation mode is, if any can be said to be the default.
- The relative advantages of each interpolation mode for typical use cases.

/cc @boundsj

boundsj:
The iOS `MGLInterpolationMode` corresponds to the GL-JS notion of [function type](https://www.mapbox.com/mapbox-gl-js/style-spec/#function-type).

The iOS SDK 3.4.x versions offer a single function type (`MGLStyleFunction` now known in the 3.5.0 beta 1 as `MGLCameraStyleFunction`) that uses exponential or interval interpolation depending on the property data type. Although we've talked about exponential (linear) interpolation in the 3.4.x API docs, I don't think we documented clearly that `MGLStyleFunction`'s exponential stops would be treated as interval stops if the data type of the property it was applied to could not be interpolated (i.e. color vs an enumeration or boolean).

The Mapbox iOS 3.5.0 beta 1 release introduces `source` and `composite` functions along with porting the legacy function to `camera` function. Not all interpolation modes apply to all function types (i.e. only exponential and interval make sense for camera) and interpolation modes can mean different things depending on the function type (i.e. source vs. composite functions). 

[...]

PR #7569, opened 2017-01-02
Insert example code in original headers

1ec5:
The script for inserting example code into documentation comments in headers (added in #7337) is no longer run as a build phase. Instead, after you add the Swift code to the test method and add

> &#x60;&#x60;&#x60;swift
> &#x60;&#x60;&#x60;

and any front matter to the original, checked-in headers, you invoke `make darwin-update-examples` manually. The script inserts the examples into the original checked-in headers instead of the headers in build products. You no longer have to invoke `make iframework` before invoking `make idocument`.

You can invoke the make rule as many times as you like. The headers will always be updated with the latest code from the test methods, even if the code block has already been filled in. This makes `darwin-update-examples` a lot like `style-code-darwin`, except that we still manage the rest of the header (including parts of the comment adjacent to the example) in the header itself. Also, there’s no reminder at the top of every header to invoke `make darwin-update-examples`. If folks tend to forget to update the examples via the test methods, we could make the script’s dry run output more parsable and run it from within a unit test.

The script updates code blocks in all public headers for both the iOS SDK and the macOS SDK. In the event that a header is used by both SDKs and the test method has a conditional compilation block, the iOS version takes precedence over the macOS version. Thanks to type inference in Swift, there’s only one place where conditional compilation is required for passing tests.

In order to reliably associate documentation comments with their symbols, the script calls out to [`sourcekitten`](https://github.com/jpsim/SourceKitten/).

Fixes #7565.

PR #8354, opened 2017-03-10
[WIP] update iOS SDK guides

jmkiley:
- [x] Update the `For Style Authors` guide to include DDS related terminology
- [ ] Add a data-driven styling guide. 
~- [ ] Update Runtime Styling guide as necessary~

The data-driven styling guide currently includes more code than the other previous guides, and may be more tutorial. It is intended to illustrate the differences between the different interpolation modes and how to use them, but it can be trimmed down.

Intended to partially address #8192  and the last two items on #7924.


platform/ios/docs/guides/For Style Authors.md
```diff
@@ -253,9 +271,9 @@ In style JSON | In Objective-C | In Swift
 
 Each property representing a layout or paint attribute is set to an
 `MGLStyleValue` object, which is either an `MGLConstantStyleValue` object (for
-constant values) or an `MGLStyleFunction` object (for zoom level functions). The
-style value object is a container for the raw value or function parameters that
-you want the attribute to be set to.
+constant values) or an `MGLStyleFunction` subclass. The style value object is a
+container for the raw value or function parameters that you want the attribute
+to be set to.
```
1ec5: Setting style functions is a subtask of configuring the map content’s appearance. I think the table mapping function types to MGLStyleFunction subclasses and MGLInterpolationMode values should be moved down here, to help progress the reader from simpler concepts to more advanced concepts.

platform/ios/docs/guides/For Style Authors.md
```diff
@@ -169,6 +169,24 @@ To create a shape source from local GeoJSON data, first
...
+In style JSON                | In the SDK
+-----------------------------|-----------
+`zoom function`              | `MGLCameraStyleFunction`
+`property function`          | `MGLSourceStyleFunction`
+`zoom-and-property functions`| `MGLCompositeStyleFunction`
+`type`                       | `MGLInterpolationMode`
+`exponential`                | `MGLInterpolationModeExponential`
+`interval`                   | `MGLInterpolationModeInterval`
+`categorical`                | `MGLInterpolationModeCategorical`
+`identity`                   | `MGLInterpolationModeIdentity`
```
1ec5: Let’s split up this table into two: one for function types and the other for interpolation modes.

platform/ios/docs/guides/For Style Authors.md
```diff
@@ -169,6 +169,24 @@ To create a shape source from local GeoJSON data, first
...
+[Data-driven styling](data-driven-styling.html) expands `MGLStyleFunction` and
+introduces `MGLInterpolationMode`. Individual style property documentation
+includes which subclasses of `MGLStyleFunction` are enabled for that property.
+
+In style JSON                | In the SDK
+-----------------------------|-----------
+`zoom function`              | `MGLCameraStyleFunction`
+`property function`          | `MGLSourceStyleFunction`
+`zoom-and-property functions`| `MGLCompositeStyleFunction`
+`type`                       | `MGLInterpolationMode`
```
1ec5: This is also a terminological difference. It’s worth mentioning that a “function type” is an “interpolation mode” up in the first table.

platform/ios/docs/guides/For Style Authors.md
```diff
@@ -169,6 +169,24 @@ To create a shape source from local GeoJSON data, first
...
 then use the `-[MGLShapeSource initWithIdentifier:shape:options:]` method.
 
+# Style functions and interpolation mode
+
+The runtime styling API introduced `MGLStyleFunction` to the iOS SDK.
+[Data-driven styling](data-driven-styling.html) expands `MGLStyleFunction` and
+introduces `MGLInterpolationMode`. Individual style property documentation
+includes which subclasses of `MGLStyleFunction` are enabled for that property.
+
+In style JSON                | In the SDK
```
1ec5: Function types (“zoom”, “property”, “zoom-and-property”) don’t appear in the style JSON, only in the style specification.

platform/ios/docs/guides/Data-Driven Styling.md
```diff
@@ -0,0 +1,127 @@
...
+
+```
+  if let url = URL(string: "https://earthquake.usgs.gov/earthquakes/feed/v1.0/summary/all_week.geojson") {
+            let symbolSource = MGLSource(identifier: "source")
+            let symbolLayer = MGLSymbolStyleLayer(identifier: "place-city-sm", source: symbolSource)
+
+            let source = MGLShapeSource(identifier: "earthquakes", url: url, options: nil)
+            style.addSource(source)
+
+            let stops = [0 : MGLStyleValue(rawValue: UIColor.yellow),
```
1ec5: We could make this guide platform-agnostic, either by replacing UIColor with some other type like NSNumber that also exists on macOS, or by relying on type inference in Swift and the `MGLColor` macro in Objective-C. Then we could move this guide to platform/darwin/docs/guides/.

platform/ios/docs/guides/For Style Authors.md
```diff
@@ -28,7 +28,7 @@ make sure the contents of these elements remain legible with the map view
 underneath.
 The user location annotation view, the attribution button, any buttons in
 callout views, and any items in the navigation bar are influenced by your
-application’s tint color, so choose a tint color that constrasts well with your
+application’s tint color, so choose a tint color that contrasts well with your
```
1ec5: Good catch.

platform/ios/docs/guides/For Style Authors.md
```diff
@@ -169,6 +169,24 @@ To create a shape source from local GeoJSON data, first
 [convert the GeoJSON data into a shape](working-with-geojson-data.html#converting-geojson-data-into-shape-objects),
 then use the `-[MGLShapeSource initWithIdentifier:shape:options:]` method.
 
+# Style functions and interpolation mode
+
+The runtime styling API introduced `MGLStyleFunction` to the iOS SDK.
+[Data-driven styling](data-driven-styling.html) expands `MGLStyleFunction` and
```
1ec5: Since the DDS guide currently lives in platform/ios/docs/guides/, this link is broken in the macOS generated documentation. https://github.com/mapbox/mapbox-gl-native/pull/8354/files#r105503718 would be preferable, but if we can’t do that, then we should conditionalize this sentence on the platform. There are some examples above of only including a sentence on iOS or only on macOS.

platform/ios/docs/guides/Data-Driven Styling.md
```diff
@@ -0,0 +1,127 @@
...
+_Insert cool overview of things you can do!_
+
+### How to use Data-Driven Styling
+This guide uses earthquake data from the [U.S. Geological Survey](https://earthquake.usgs.gov/earthquakes/feed/v1.0/geojson.php) to illustrate to style a map based on attributes. For more information about how to work with GeoJSON data in our iOS SDK, please see our [working with GeoJSON data](https://www.mapbox.com/ios-sdk/api/3.5.0-beta.1/working-with-geojson-data.html) guide.
+
+`MGLStyleFunction`
+
+There are three subclasses of `MGLStyleFunction`:
+
+* `MGLCameraStyleFunction` - For a style value that changes with zoom level. For example, you can make the radius of a circle to increase based on zoom level. In iOS SDK v3.4, a `MGLStyleFunction` referred to what is now a `MGLCameraStyleFunction`.
```
friedbunny: > referred to what is now a `MGLCameraStyleFunction`.

This seems like unnecessary repetition. Could we say something like: “In iOS SDK v3.4, this function was called `MGLStyleFunction`.”?

platform/ios/docs/guides/Data-Driven Styling.md
```diff
@@ -0,0 +1,127 @@
...
+
+####`MGLInterpolationMode`
+
+The effect a key has on the style value is determined by the interpolation mode. There are four interpolation modes that can be used with a source style function: exponential, interval, categorical, and identity. You can also use exponential and interval interpolation modes with a camera style function.
+
+#####Exponential:
+
+`MGLInterpolationModelExponential` creates a linear effect based on the values. The key value is the base for interpolation, and the style value is based on where an attribute value falls between two keys.
+
+The stops dictionary below, for example, shows colors that continuously shift from yellow -> orange -> red -> blue -> white based on the attribute value.
```
friedbunny: Let’s convert `->` to `→` (or some other arrow symbol).

platform/ios/docs/guides/Data-Driven Styling.md
friedbunny: Definitely onboard with consistently using type inference in Swift, though.

platform/ios/docs/guides/Data-Driven Styling.md
```diff
@@ -0,0 +1,127 @@
...
+            style.insertLayer(layer, below: symbolLayer)
+        }
+```
+![exponential mode](img/data-driven-styling/exponential.png)
+
+#####Interval
+
+`MGLInterpolationModeInterval` creates a range using the keys stops dictionary. The range is from the given key to just less than the next key. The attribute values that fall into that range are then styled using the style value assigned to that key.
+
+When we use the stops dictionary given above with an interval interpolation mode, we create ranges where earthquakes with a magnitude of 0 to just less than 2.5 be yellow, 2.5 to just less than 5 would be orange, and so on.
```
friedbunny: > we create ranges where earthquakes with a magnitude of 0 to just less than 2.5 _would_ be yellow,

platform/ios/docs/guides/Data-Driven Styling.md
```diff
@@ -0,0 +1,127 @@
...
+
+```
+layer.circleColor = MGLStyleValue<UIColor>(interpolationMode: .interval,
+                                                          sourceStops: stops,
+                                                          attributeName: "mag",
+                                                          options: [.defaultValue: MGLStyleValue<UIColor>(rawValue: .green)])
+```
+<img width="320" alt="screenshot 2017-02-28 19 14 40" src="https://cloud.githubusercontent.com/assets/12474734/23444472/30d4e8b4-fdea-11e6-8184-ad9631529c74.png">
```
friedbunny: Unfortunately we can’t rely on GitHub to host these assets — be sure to commit and use the files required for this guide.

platform/ios/docs/guides/Data-Driven Styling.md
```diff
@@ -0,0 +1,127 @@
...
+
+#####Categorical:
```
friedbunny: Remove the colon in this heading.

platform/ios/docs/guides/Data-Driven Styling.md
```diff
@@ -0,0 +1,127 @@
...
+                                                          attributeName: "type",
+                                                          options: [.defaultValue: MGLStyleValue(rawValue: UIColor.blue)])
+
+```
+
+<img width="320" alt="screenshot 2017-03-01 13 05 28" src="https://cloud.githubusercontent.com/assets/12474734/23481308/d7c4d644-fe7f-11e6-9e71-11a6ecb89775.png"> <img width="320" alt="screenshot 2017-03-01 13 06 00" src="https://cloud.githubusercontent.com/assets/12474734/23481311/d90ca586-fe7f-11e6-87b8-e63cdeb1db84.png">
+
+
+
+#####Identity:
```
friedbunny: Colon.

platform/ios/docs/guides/Data-Driven Styling.md
1ec5: Point taken about MGLColor being pedagogically problematic. In the Mapbox API Documentation examples in Objective-C, I took to using conditional compilation macros, which are better documented but even more poorly understood by less experienced developers.

What about bringing this guide under the same codegen mechanism that the “Information for Style Authors” guide uses?

1. Move this guide to platform/darwin/docs/guides/Data-Driven Styling.md.ejs.
1. Define a [`cocoaPrefix`](https://github.com/mapbox/mapbox-gl-native/blob/a0a69ad6a7d4428961d99564b9b09b23a0acfcf6/platform/darwin/docs/guides/For%20Style%20Authors.md.ejs#L5) variable at the top.
1. Put a `<% cocoaPrefix %>` variable anywhere we’re currently saying `UI`.
1. Extend generate-style-code.js to also [write out](https://github.com/mapbox/mapbox-gl-native/blob/a0a69ad6a7d4428961d99564b9b09b23a0acfcf6/platform/darwin/scripts/generate-style-code.js#L607-L616) iOS- and macOS-specific versions of the guide.
1. Run `make style-code`.

With this approach, the style-related guides would be more consistent with each other even at the code level. Additionally, it’ll give us the ability to put some of these (rather substantial) code examples in unit tests and insert them into this document automatically, ensuring that the examples stay current. We already use that approach for examples embedded in API documentation.

platform/ios/docs/guides/Data-Driven Styling.md
1ec5: Or “to”, since this is prose.

platform/ios/docs/guides/Data-Driven Styling.md
```diff
@@ -0,0 +1,127 @@
...
+
+The effect a key has on the style value is determined by the interpolation mode. There are four interpolation modes that can be used with a source style function: exponential, interval, categorical, and identity. You can also use exponential and interval interpolation modes with a camera style function.
+
+#####Exponential:
+
+`MGLInterpolationModelExponential` creates a linear effect based on the values. The key value is the base for interpolation, and the style value is based on where an attribute value falls between two keys.
+
+The stops dictionary below, for example, shows colors that continuously shift from yellow -> orange -> red -> blue -> white based on the attribute value.
+
+```
```
1ec5: Specifying `swift` here would get us syntax coloring.

platform/ios/docs/guides/Data-Driven Styling.md
```diff
@@ -0,0 +1,127 @@
...
+  if let url = URL(string: "https://earthquake.usgs.gov/earthquakes/feed/v1.0/summary/all_week.geojson") {
```
1ec5: `URL(string:)` only returns nil if the string is a malformed URL. Since the URL is formed from a string literal, we know it’s well-formed, so we can `!` unwrap the optional without indenting everything below.

platform/ios/docs/guides/Data-Driven Styling.md
```diff
@@ -0,0 +1,127 @@
...
+            let symbolSource = MGLSource(identifier: "source")
```
1ec5: But if we were to indent, we should use four spaces for each level of indentation.

platform/ios/docs/guides/Data-Driven Styling.md
```diff
@@ -0,0 +1,127 @@
...
+                         2.5 : MGLStyleValue(rawValue: UIColor.orange),
+                         5: MGLStyleValue(rawValue: UIColor.red),
+                         7.5 : MGLStyleValue(rawValue: UIColor.blue),
+                         10 : MGLStyleValue(rawValue: UIColor.white)]
+
+            let layer = MGLCircleStyleLayer(identifier: "circles", source: source)
+            layer.circleColor = MGLStyleValue<UIColor>(interpolationMode: .exponential,
+                                                          sourceStops: stops,
+                                                          attributeName: "mag",
+                                                          options: [.defaultValue: MGLStyleValue<UIColor>(rawValue: .green)])
```
1ec5: This reminds me: we should find a good place to mention the ability to use #8025, even if it isn’t important enough to show in a code example.

platform/ios/docs/guides/For Style Authors.md
1ec5: Per https://github.com/mapbox/mapbox-gl-native/pull/8025#discussion_r101864980, let’s take this opportunity to point out that not every interpolation mode is available for every attribute. The developer should consult the documentation for an individual layout or paint attribute to see which function types and interpolation modes it supports.

platform/ios/docs/guides/Data-Driven Styling.md
jmkiley: Oh, good point. I will see if I can adjust the example to show that. Alternatively, we can use it in an upcoming example

platform/ios/docs/guides/Data-Driven Styling.md
jmkiley: Definitely! In the process of doing that. I copied this from a gist and am still working on optimizing the screenshots

platform/ios/docs/guides/Data-Driven Styling.md
```diff
@@ -0,0 +1,124 @@
...
+
+There are three subclasses of `MGLStyleFunction`:
+
+* `MGLCameraStyleFunction` - For a style value that changes with zoom level. For example, you can make the radius of a circle to increase based on zoom level. In iOS SDK v3.4, this was a `MGLStyleFunction`.
+* `MGLSourceStyleFunction` - For a style value that changes with the attributes of a feature. You can adjust the radius of a circle based on the magnitude of an earthquake, for example.
+* `MGLCompositeStyleFunction` - For a style value that changes with both zoom level and attribute values. For example, you can add a circle layer where each circle has a radius based on both zoom level and the magnitude of an earthquake.
+
+The documentation for individual style properties will note which style functions are enabled for that property.
+
+####Stops
```
boundsj: I noticed that 4 `#`s cause `Stops` to become `STOPS` in the generated documentation and that looks strange. Also, super nit, but we should probably format the headers consistently (add a space between the `#` and the first character of the text.

platform/ios/docs/guides/Data-Driven Styling.md
```diff
@@ -0,0 +1,124 @@
...
+
+Stops are key-value pairs that that determine a style value. With a `MGLCameraSourceFunction` stop, you can use a dictionary with a zoom level for a key and a `MGLStyleValue` for the value. For example, you can use a stops dictionary with zoom levels 0, 10, and 20 as keys, and yellow, orange, and red as the values. A `MGLSourceStyleFunction` uses the relevant attribute value as the key.
+
+####`MGLInterpolationMode`
```
boundsj: In the generated documentation, the four `#`s capitalization issue combined with the formatting caused by the back ticks makes this looke like `MGLINTERPOLATIONMODE` with a really thin font weight.

platform/ios/docs/guides/Data-Driven Styling.md
```diff
@@ -0,0 +1,124 @@
...
+
+The effect a key has on the style value is determined by the interpolation mode. There are four interpolation modes that can be used with a source style function: exponential, interval, categorical, and identity. You can also use exponential and interval interpolation modes with a camera style function.
```
boundsj: Instead of calling out that "exponential and interval interpolation modes" can be used on a camera function, this might be a good place to add another table that shows which functions support which modes. That is:

camera: exponential & interval
source: all of them
composite: exponential, interval, categorial (but not identity)

platform/ios/docs/guides/For Style Authors.md
boundsj: I agree. Especially since the tables currently repeat for each property.

Data-Driven Styling.md.ejs
```diff
@@ -0,0 +1 @@
+wat
```
friedbunny: 🤔

jmkiley: that's from debugging /o\

platform/darwin/docs/guides/Data-Driven Styling.md.ejs
```diff
@@ -0,0 +1,138 @@
...
+- [ ] Add cocoaprefix so this can be moved to darwin
+---->
+
+<%
+  const os = locals.os;
+  const iOS = os === 'iOS';
+  const macOS = os === 'macOS';
+  const cocoaPrefix = iOS ? 'UI' : 'NS';
+  const layers = locals.layers;
+  const renamedProperties = locals.renamedProperties;
```
1ec5: A couple of these variables appear to be unused. You can safely remove them.

platform/darwin/docs/guides/Data-Driven Styling.md.ejs
```diff
@@ -0,0 +1,138 @@
+<!---
+To do - JK:
+- [ ] Fill it out with more detailed explanations
+- [ ] Add examples for camera or composite style functions
+- [ ] Cluster earthquakes as an example for camera function?
+- [ ] Clean up & optimize screenshots - replace with screenshots showing circles below symbol layer
+- [ ] Link API docs
+- [ ] Add cocoaprefix so this can be moved to darwin
```
1ec5: Looks like this item is complete.

platform/darwin/docs/guides/Data-Driven Styling.md.ejs
```diff
@@ -0,0 +1,138 @@
...
+![exponential mode](img/data-driven-styling/exponential.png)
+
+#### Interval
+
+`MGLInterpolationModeInterval` creates a range using the keys stops dictionary. The range is from the given key to just less than the next key. The attribute values that fall into that range are then styled using the style value assigned to that key.
+
+When we use the stops dictionary given above with an interval interpolation mode, we create ranges where earthquakes with a magnitude of 0 to just less than 2.5 would be yellow, 2.5 to just less than 5 would be orange, and so on.
+
+``` swift
+layer.circleColor = MGLStyleValue<<%- cocoaPrefix %>Color>(interpolationMode: .interval,
```
1ec5: It isn't necessary to specify NSColor/UIColor here, because Swift can infer the type from the type of circleColor.

platform/darwin/docs/guides/For Style Authors.md.ejs
```diff
@@ -166,6 +166,7 @@ layer                      | style layer
 property                   | attribute
 SDF icon                   | template image
 source                     | content source
+function type              | interpolation mode
```
1ec5: Move this entry up to keep the table alphabetized by the first column.

platform/darwin/docs/guides/For Style Authors.md.ejs
```diff
@@ -239,6 +240,28 @@ whose names differ from the style specification are listed below:
...
+introduces `MGLInterpolationMode`. Individual style property documentation
+includes which subclasses of `MGLStyleFunction` are enabled for that property.
+
+In style specification       | In the SDK
+-----------------------------|-----------
+`zoom function`              | `MGLCameraStyleFunction`
+`property function`          | `MGLSourceStyleFunction`
+`zoom-and-property functions`| `MGLCompositeStyleFunction`
+
+### <%- camelize(type) %> interpolation mode
```
1ec5: This table doesn't need its own subheading, but a complete sentence introducing the concept of an interpolation mode is would help to make this document flow better.

platform/darwin/scripts/generate-style-code.js
```diff
@@ -614,3 +615,13 @@ fs.writeFileSync(`platform/macos/docs/guides/For Style Authors.md`, guideMD({
     renamedProperties: renamedPropertiesByLayerType,
     layers: layers,
 }));
+fs.writeFileSync(`platform/ios/docs/guides/Data-Driven Styling.md`, ddsGuideMD({
+    os: 'iOS',
+    renamedProperties: renamedPropertiesByLayerType,
+    layers: layers,
```
1ec5: Of these variables, it looks like only os is in use.

platform/darwin/docs/guides/For Style Authors.md.ejs
```diff
@@ -239,6 +240,28 @@ whose names differ from the style specification are listed below:
 <% for (const type in renamedProperties) { -%>
 <% if (renamedProperties.hasOwnProperty(type)) { -%>
 
+### <%- camelize(type) %> style functions
+
+The runtime styling API introduced `MGLStyleFunction` to the <%- os %> SDK.
```
1ec5: This reads sort of like a release note. Let's cast this sentence in the present tense and talk about how the API includes style functions rather than introducing them.

Additionally, we have a tendency to use MGLStyleValue as an umbrella for both constants and functions. So even if we talk about the various style function subclasses, I think we should discuss the MGLStyleValue factory method that produces each kind of style function.

platform/darwin/docs/guides/Data-Driven Styling.md.ejs
```diff
@@ -0,0 +1,133 @@
+<!---
+To do - JK:
```
friedbunny: Are these to-does done?

platform/macos/docs/guides/Data-Driven Styling.md
```diff
@@ -0,0 +1,125 @@
+<!---
+To do - JK:
```
friedbunny: Todoooooooooo.

platform/ios/docs/guides/Data-Driven Styling.md
```diff
@@ -0,0 +1,125 @@
+<!---
+To do - JK:
```
friedbunny: Toooooooooodoooooooo.

friedbunny:
General note: these guides would benefit greatly from linking up the method/class names to their jazzy pages.

platform/darwin/docs/guides/Data-Driven Styling.md.ejs
```diff
@@ -0,0 +1,122 @@
...
+    let symbolLayer = MGLSymbolStyleLayer(identifier: "place-city-sm", source: symbolSource)
+
+    let source = MGLShapeSource(identifier: "earthquakes", url: url, options: nil)
+    style.addSource(source)
+
+    let stops = [0 : MGLStyleValue(rawValue: <%- cocoaPrefix %>Color.yellow),
+                 2.5 : MGLStyleValue(rawValue: <%- cocoaPrefix %>Color.orange),
+                 5: MGLStyleValue(rawValue: <%- cocoaPrefix %>Color.red),
+                 7.5 : MGLStyleValue(rawValue: <%- cocoaPrefix %>Color.blue),
+                 10 : MGLStyleValue(rawValue: <%- cocoaPrefix %>Color.white)]
```
friedbunny: Remove the spaces before the colons.

screen shot 2017-03-17 at 4 59 55 pm

friedbunny: You can also remove the initial tab indentation, as it’s indenting the entire block by one.

platform/darwin/docs/guides/Data-Driven Styling.md.ejs
```diff
@@ -0,0 +1,122 @@
...
+                                      options: [.defaultValue: MGLStyleValue<<%- cocoaPrefix %>Color>(rawValue: .green)])
+    layer.circleRadius = MGLStyleValue(rawValue: 10)
+    style.insertLayer(layer, below: symbolLayer)
+```
+
+![exponential mode](img/data-driven-styling/exponential.png)
+
+#### Interval
+
+`MGLInterpolationModeInterval` creates a range using the keys stops dictionary. The range is from the given key to just less than the next key. The attribute values that fall into that range are then styled using the style value assigned to that key.
```
friedbunny: >keys stops dictionary

Is this a dictionary of stops for multiple keys? Pluralization is a little awkward, but if the meaning is correct then it’s probably OK.

platform/darwin/docs/guides/Data-Driven Styling.md.ejs
```diff
@@ -0,0 +1,122 @@
...
+
+#### Categorical
+
+Returns the output value that is equal to the stop for the function input. We’re going to use a different stops dictionary than we did for the previous two modes.
+
+There are three main types of events in the dataset: earthquakes, explosions, and quarry blasts. In this case, the color of the circle layer will be determined by the type of event, with a default value of green to catch any events that do not fall into any of those categories.
+
+``` swift
+let categoricalStops = ["earthquake" : MGLStyleValue(rawValue: <%- cocoaPrefix %>Color.orange),
+                                    "explosion" : MGLStyleValue(rawValue: <%- cocoaPrefix %>Color.red),
```
friedbunny: Remove the spaces before the colons and realign these lines.

platform/darwin/docs/guides/Data-Driven Styling.md.ejs
```diff
@@ -0,0 +1,122 @@
...
+layer.circleRadius = MGLStyleValue(interpolationMode: .identity,
+                                   sourceStops: nil,
+                                   attributeName: "mag",
+                                   options: [.defaultValue: MGLStyleValue<NSNumber>(rawValue: 0)])
+
+```
+
+![identity mode](img/data-driven-styling/identity.png)
+
+Resources:
```
friedbunny: Promote this to a bigger header, as in the bottom of the Runtime Styling guide.

friedbunny:
Are these code snippets testable? We’re dynamically inserting tested snippets into header documentation, but I’m not sure we’ve done that yet for guides.

/cc @ericrwolfe

jmkiley:
data-driven styling reference 20170317

ericrwolfe:
Re: testability

I'd originally shied away from adding code snippets to the runtime styling guide out of concern for maintainability. But now that we have testable code examples we could certainly take a similar approach for the markdown guides. 

We might be able to rework the original script from this PR https://github.com/mapbox/mapbox-gl-native/pull/7337 to scan through the guides directory and replace placeholder tokens in the markdown files. 

Note that the current examples script in `platform/darwin/scripts/update-examples.js` from https://github.com/mapbox/mapbox-gl-native/pull/7569 is a bit more specific to work around the fact that the style layer classes are generated dynamically.

cc @1ec5

1ec5:
> these guides would benefit greatly from linking up the method/class names to their jazzy pages.
> ... which is something we should look into contributing upstream to jazzy.

realm/jazzy#715

> Note that the current examples script in `platform/darwin/scripts/update-examples.js` from https://github.com/mapbox/mapbox-gl-native/pull/7569 is a bit more specific to work around the fact that the style layer classes are generated dynamically.

We also took that approach for the examples in the headers because jazzy documentation links to specific lines in the headers based on the source code it consumes – we didn’t want the line numbers to go out of sync because of inserted code. But that problem doesn’t affect the Markdown guides.

jmkiley:
I opened a ticket so we can discuss how to make code snippets testable. #8461

PR #8361, opened 2017-03-11
Verbose commenting on getSegmentGlyphs

ChrisLoer:
@anandthakker and I paired on trying to understand how getSegmentGlyphs works, since we couldn't pick @ansis 's brain. Our goal was (1) make sure we understand how the function is supposed to work, and then (2) use the comments as guidelines to refactor the function so that some of the magic parts are less mysterious.

### DIFF
--- a/src/mbgl/text/quads.cpp
+++ b/src/mbgl/text/quads.cpp
@@ -138,20 +138,15 @@ inline void insertSegmentGlyph(std::back_insert_iterator<GlyphInstances> glyphs,
  Given the distance along the line from the label anchor to the beginning of the current segment,
  project a "virtual anchor" point at the same distance along the line extending out from this segment.
  
+                B <-- beginning of current segment
+* . . . . . . . *--------* E <-- end of current segment
+VA              |      
+                /        VA = "virtual segment anchor"
+               /
+     ---*-----`
+        A = label anchor
  
-                     B <-- beginning of current segment
-                    .*-----
-                  .  |
-                .    /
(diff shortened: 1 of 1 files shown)


ChrisLoer:
OK @anandthakker , over to you. :)

I took a stab at refactoring based on the understanding we developed during the commenting. If you did a round of improvements on variable names, function inlining, control flow, and commenting, I bet it would start to look pretty good.

anandthakker:
@ChrisLoer your first pass was a huge improvement / most of the work.  Just took a pass to add some more comments, over to @jfirebaugh to 👀

ChrisLoer:
Awesome ASCII art! But unless I'm misunderstanding the diagram, I don't think it represents the "virtual anchor" concept correctly. Instead of:

                    B <-- beginning of current segment
                    .*-----
                  .  |
                .    /
              .     /
            .      /
    ------.*------`
        .  A
      .    anchor
     *
    VA <- "virtual" anchor

I think it's more like:


                      B <-- beginning of current segment
    * . . . . . . . *--------* E <-- end of current segment
    VA              |      
                    /        VA = "virtual segment anchor"
                   /
         ---*-----`
            A = label anchor

anandthakker:
@ChrisLoer oh yeah, you are totally right.  I misread the vector math!

anandthakker:
@jfirebaugh okay, this should be ready for 👀 now

ChrisLoer:
🎉

PR #8363, opened 2017-03-11
Integrate CMake build into Gradle + Android Studio

kkaefer:


tobrun:
I have taken this PR for a spin and is looking great:

Compiling c++ code as part of the build:

screen shot 2017-03-13 at 10 06 00

Using LLDB debugger directly in IDE with breakpoint support:

screen shot 2017-03-13 at 10 21 24

platform/android/config.cmake
```diff
@@ -250,11 +250,6 @@ target_link_libraries(mapbox-gl
     PUBLIC -Wl,--gc-sections
 )
 
-# Create a stripped version of the library and copy it to the JNIDIR.
```
jfirebaugh: Does this still happen somewhere else? Do we need to update [metrics.sh](https://github.com/mapbox/mapbox-gl-native/blob/36a04e49c834f52828dd26871641517c000c45f5/platform/android/scripts/metrics.sh#L7-L12)?

kkaefer:
```
before:
200,816,848 bytes: build/android-arm-v8-21/Release/libmapbox-gl.so
  5,510,144 bytes: platform/android/MapboxGLAndroidSDK/src/main/jniLibs/arm64-v8a/libmapbox-gl.so

after:
201,035,760 bytes: platform/android/MapboxGLAndroidSDK/build/intermediates/cmake/release/obj/arm64-v8a/libmapbox-gl.so
  5,510,144 bytes: platform/android/MapboxGLAndroidSDK/build/intermediates/bundles/default/jni/arm64-v8a/libmapbox-gl.so
```

The Gradle build automatically strips the symbols before putting them into the APK.

Given that the output path is now different, I also updated `metrics.sh`.

PR #8369, opened 2017-03-13
Note that quick zoom is unavailable on iPad

1ec5:
Added a little note to the gesture recognizers guide to indicate that the “quick zoom” (“one-finger zoom”) gesture is only available on iPhone and iPod touch, not iPad.

/cc @incanus


friedbunny:
Hrm, I had forgotten that we disable quick zoom on non-phones and was surprised that we do so. I’ve put in #8379 to remove this limitation.

1ec5:
Closing in favor of #8379.

PR #8377, opened 2017-03-13
[core] Replace mbgl::MapChange with mbgl::MapObserver

### LINKED ISSUE #6383: Replace MapChange enum and `Backend::notifyMapChange` with dedicated observer interface

brunoabinader:
Implements `mbgl::MapObserver`, replacing `mbgl::MapChange` enumerator and `mbgl::Backend::notifyMapChange`. Now `mbgl::Backend` inherits from `mbgl::MapObserver`, which provides non-pure virtual equivalents for the enumerator items.

Caveats:
- Qt preserves the `QMapboxGL::MapChange` enumerator from the public API.
- Android refactor still requires `mbgl::MapChange` - this is not necessary but finishing up this platform refactor can happen on a separate PR. Once we finish cleaning up the Android platform, then we can safely remove `mbgl::MapChange` source code entirely.

Added suggested reviewers for platform-specific patches as well as core changes - please replace yourself with someone else from the mobile team otherwise.

Fixes #6383.

### DIFF
--- a/include/mbgl/map/map_observer.hpp
+++ b/include/mbgl/map/map_observer.hpp
@@ -1,0 +1,38 @@
+#pragma once
+
+#include <cstdint>
+
+namespace mbgl {
+
+class MapObserver {
+public:
+    static MapObserver& nullObserver() {
+        static MapObserver mapObserver;
+        return mapObserver;
+    }
+
+    enum class CameraChangeMode : uint32_t {
+        Immediate,
+        Animated
+    };
+
+    enum class RenderMode : uint32_t {
+        Partial,
+        Full
+    };
+
+    virtual void onCameraWillChange(CameraChangeMode) {}
+    virtual void onCameraIsChanging() {}
+    virtual void onCameraDidChange(CameraChangeMode) {}
+    virtual void onWillStartLoadingMap() {}
+    virtual void onDidFinishLoadingMap() {}
+    virtual void onDidFailLoadingMap() {}
+    virtual void onWillStartRenderingFrame() {}
+    virtual void onDidFinishRenderingFrame(RenderMode) {}
+    virtual void onWillStartRenderingMap() {}
+    virtual void onDidFinishRenderingMap(RenderMode) {}
+    virtual void onDidFinishLoadingStyle() {}
+    virtual void onSourceDidChange() {}
+};
+
+} // namespace mbgl
--- a/platform/ios/src/MGLMapView.mm
+++ b/platform/ios/src/MGLMapView.mm
@@ -1167,7 +1167,8 @@ public:
 }
 
 - (void)notifyGestureDidBegin {
-    [self notifyMapChange:mbgl::MapChangeRegionWillChange];
+    BOOL animated = NO;
+    [self cameraWillChangeAnimated:animated];
     _mbglMap->setGestureInProgress(true);
     _changeDelimiterSuppressionDepth++;
 }
@@ -1181,7 +1182,8 @@ public:
     }
     if ( ! drift)
     {
-        [self notifyMapChange:mbgl::MapChangeRegionDidChange];
+        BOOL animated = NO;
+        [self cameraDidChangeAnimated:animated];
     }
 }
 
@@ -1218,7 +1220,7 @@ public:
             [pan setTranslation:CGPointZero inView:pan.view];
         }
 
-        [self notifyMapChange:mbgl::MapChangeRegionIsChanging];
+        [self cameraIsChanging];
     }
     else if (pan.state == UIGestureRecognizerStateEnded || pan.state == UIGestureRecognizerStateCancelled)
     {
@@ -1301,7 +1303,7 @@ public:
             _mbglMap->setLatLng(MGLLatLngFromLocationCoordinate2D(centerCoordinate),
                                 mbgl::ScreenCoordinate { centerPoint.x, centerPoint.y });
         }
-        [self notifyMapChange:mbgl::MapChangeRegionIsChanging];
+        [self cameraIsChanging];
     }
     else if (pinch.state == UIGestureRecognizerStateEnded || pinch.state == UIGestureRecognizerStateCancelled)
     {
@@ -1400,9 +1402,8 @@ public:
         {
            _mbglMap->setBearing(newDegrees, mbgl::ScreenCoordinate { centerPoint.x, centerPoint.y });
         }
-        
-        [self notifyMapChange:mbgl::MapChangeRegionIsChanging];
-        
+
+        [self cameraIsChanging];
     }
     else if (rotate.state == UIGestureRecognizerStateEnded || rotate.state == UIGestureRecognizerStateCancelled)
     {
@@ -1651,7 +1652,7 @@ public:
             _mbglMap->scaleBy(scale, mbgl::ScreenCoordinate { centerPoint.x, centerPoint.y });
         }
 
-        [self notifyMapChange:mbgl::MapChangeRegionIsChanging];
+        [self cameraIsChanging];
     }
     else if (quickZoom.state == UIGestureRecognizerStateEnded || quickZoom.state == UIGestureRecognizerStateCancelled)
     {
@@ -1690,7 +1691,7 @@ public:
             _mbglMap->setPitch(pitchNew, mbgl::ScreenCoordinate { centerPoint.x, centerPoint.y });
         }
 
-        [self notifyMapChange:mbgl::MapChangeRegionIsChanging];
+        [self cameraIsChanging];
     }
     else if (twoFingerDrag.state == UIGestureRecognizerStateEnded || twoFingerDrag.state == UIGestureRecognizerStateCancelled)
     {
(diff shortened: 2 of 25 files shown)


include/mbgl/map/map_observer.hpp
```diff
@@ -0,0 +1,30 @@
...
+namespace mbgl {
+
+class MapObserver {
+public:
+    static MapObserver& nullObserver() {
+        static MapObserver mapObserver;
+        return mapObserver;
+    }
+
+    virtual void onRegionWillChange() {}
```
jfirebaugh: Suggest `s/Region/Camera/g`.

include/mbgl/map/map_observer.hpp
```diff
@@ -0,0 +1,30 @@
...
+    virtual void onRegionWillChangeAnimated() {}
```
jfirebaugh: Would it make sense to provide a boolean (or two-valued-enum) parameter to `on*Change` , rather than having separate `on*Change` and `on*ChangeAnimated` methods? Similarly with `on*Rendering{Frame,Map}`/`on*Rendering{Frame,Map}FullyRendered`.

include/mbgl/map/map_observer.hpp
```diff
@@ -0,0 +1,30 @@
...
+    virtual void onRegionIsChanging() {}
+    virtual void onRegionDidChange() {}
+    virtual void onRegionDidChangeAnimated() {}
+    virtual void onWillStartLoadingMap() {}
+    virtual void onDidFinishLoadingMap() {}
+    virtual void onDidFailLoadingMap() {}
+    virtual void onWillStartRenderingFrame() {}
+    virtual void onDidFinishRenderingFrame() {}
+    virtual void onDidFinishRenderingFrameFullyRendered() {}
+    virtual void onWillStartRenderingMap() {}
```
jfirebaugh: Is there a useful difference between `on*RenderingFrame` and `on*RenderingMap`?

1ec5: `onWillStartRenderingMap()` and `onDidFinishRenderingMap()` are based on MapKit’s [`-[MKMapViewDelegate mapViewWillStartRenderingMap:]`](https://developer.apple.com/reference/mapkit/mkmapviewdelegate/1451970-mapviewwillstartrenderingmap?language=objc) (“when one or more tiles are revealed and require rendering”) and [`-[MKMapViewDelegate mapViewDidFinishRenderingMap:fullyRendered:]`](https://developer.apple.com/reference/mapkit/mkmapviewdelegate/1451897-mapviewdidfinishrenderingmap?language=objc) (“finishes rendering all of the currently visible tiles to the best of its ability”), which have analogues in our iOS and macOS SDKs.

If our iOS documentation is to be believed, `on*RenderingFrame()` would be called on each frame of a camera animation or property transition, even where no additional tiles need to be fetched and displayed. However, I don’t know for sure if that’s how we’ve implemented these notifications, given #2775.

/cc @friedbunny

1ec5: In terms of usefulness, we’ve found the “rendering frame” notifications to be useful in situations where the developer wants to synchronize a view atop the map or override some annotation view behavior that we apply on every frame. But the “rendering map” notifications remain more generally useful for knowing when it’s safe to snapshot a map view or have the UI change once the map is quiescent.

platform/ios/src/MGLMapView.mm
```diff
@@ -4736,153 +4737,167 @@ - (void)unrotateIfNeededAnimated:(BOOL)animated
...
+    }
+
+    self.style = [[MGLStyle alloc] initWithMapView:self];
+    if ([self.delegate respondsToSelector:@selector(mapView:didFinishLoadingStyle:)])
+    {
+        [self.delegate mapView:self didFinishLoadingStyle:self.style];
+    }
+}
+
+- (void)onSourceDidChange {
```
1ec5: iOS doesn’t need to do anything in response to source changes, since attribution is displayed on-demand. You can safely remove this method and make its caller a no-op.

platform/macos/src/MGLMapView.mm
```diff
@@ -820,133 +820,151 @@ - (void)setNeedsGLDisplay {
...
+    }
+
+    self.style = [[MGLStyle alloc] initWithMapView:self];
+    if ([self.delegate respondsToSelector:@selector(mapView:didFinishLoadingStyle:)])
+    {
+        [self.delegate mapView:self didFinishLoadingStyle:self.style];
+    }
+}
+
+- (void)onSourceDidChange {
```
1ec5: Pass in the `mbgl::style::Source` and rename this method to `-sourceDidChange:`.

src/mbgl/map/map.cpp
```diff
@@ -1095,11 +1099,11 @@ void Map::Impl::onUpdate(Update flags) {
 }
 
 void Map::Impl::onStyleLoaded() {
-    backend.notifyMapChange(MapChangeDidFinishLoadingStyle);
+    observer.onDidFinishLoadingStyle();
 }
 
 void Map::Impl::onStyleError() {
-    backend.notifyMapChange(MapChangeDidFailLoadingMap);
+    observer.onDidFailLoadingMap();
```
1ec5: We should plumb [`res.error`](https://github.com/mapbox/mapbox-gl-native/blob/936cc33d5291a77bc4e0dda9c7956860bef5f872/src/mbgl/map/map.cpp#L371) through to here or implement `onResourceError()` instead to allow the SDKs to provide more information about the failure. Currently, in order to fudge parity with MapKit, the iOS and macOS SDKs have to [synthesize an uninformative NSError](https://github.com/mapbox/mapbox-gl-native/blob/de9c0748a216d05b15509f1a73086f70067643b2/platform/ios/src/MGLMapView.mm#L4830), but `res.error`’s `message` field would be perfect for this purpose.

platform/ios/src/MGLMapView.mm
```diff
@@ -4736,153 +4737,167 @@ - (void)unrotateIfNeededAnimated:(BOOL)animated
     }
 }
 
-- (void)notifyMapChange:(mbgl::MapChange)change
-{
-    // Ignore map updates when the Map object isn't set.
+- (void)onRegionWillChange:(bool)animated {
```
1ec5: `-cameraWillChangeAnimated:` would also be acceptable.

The macOS SDK implementation of MGLMapViewDelegate already uses the term “camera” instead of “region”. The iOS SDK implementation blindly copied MapKit’s use of “region” without bringing over MapKit’s `MKRegion` struct.

We’ll want to fix the public interface when we’re ready for backwards-incompatible changes, but for now, either name works for this internal method.

include/mbgl/map/map_observer.hpp
brunoabinader: Ack. For now I'm abstaining from renaming the public APIs for Qt & Android. For iOS & macOS I'm adding entries to `CHANGELOG.md` about the renaming.

include/mbgl/map/map_observer.hpp
brunoabinader: I think so - I'm adding two enums to `MapObserver`: `CameraChangeMode` and `RenderMode`.

include/mbgl/map/map_observer.hpp
1ec5: Renaming the public API for iOS and macOS would be a backwards-incompatible change that would require a version bump. Fortunately, this PR doesn’t touch the public API (MGLMapView.h) at all, only the internal implementation of MGLMapView (MGLMapView.mm), so no changelog entry is required.

platform/macos/src/MGLMapView.mm
brunoabinader: I'm 👍  with the renaming, but I'd prefer to defer the source argument change as a platform-specific change meant to a separate PR in the same way @tobrun did in #8389 for Android.

src/mbgl/map/map.cpp
brunoabinader: Same as above, I'd prefer if we could move this to a platform-specific refactor.

platform/macos/src/MGLMapView.mm
1ec5: → #8397

src/mbgl/map/map.cpp
1ec5: This suggestion isn’t platform-specific but rather an improvement on the code being added in this PR. Tracking in #8398.

platform/ios/src/MGLMapView.mm
```diff
@@ -1167,7 +1167,8 @@ - (void)touchesBegan:(__unused NS_SET_OF(UITouch *) *)touches withEvent:(__unuse
 }
 
 - (void)notifyGestureDidBegin {
-    [self notifyMapChange:mbgl::MapChangeRegionWillChange];
+    bool animated = false;
```
1ec5: s/bool/BOOL/ and s/false/NO/ (×2)

brunoabinader: I'm a bit confused - I've seen `bool` usage when the code is a glue between C++ and Objective-C, and `BOOL` on Objective-C code along this source file - should we open an issue then to fix this for all remaining `bool`s?

src/mbgl/map/map.cpp
brunoabinader: I see - thanks for the clarification. Let's handle this in #8398 👍

platform/ios/src/MGLMapView.mm
brunoabinader: Ok, I think I got it now - this function is not part of the C++ glue and thus should use the Objective-C notation - got it :) Fixing now.

brunoabinader:
Qt public API refactor handled in #8419.

platform/ios/src/MGLMapView.mm
1ec5: 👍 As a rule of thumb, in Objective-C++ code, if you’re inside an Objective-C class or category (look for `@implementation` and `@end`), use Objective-C types; if you’re inside a C++ `class` or `struct`, use C++ types. There are some exceptions: for example, I find it useful to use `static_cast` over C-style casts throughout Objective-C++ code. And `std::vector` can often be faster than NSMutableArray. [This article](https://pspdfkit.com/blog/2016/swifty-objective-c/) suggests some more ways C++ can improve Objective-C code, but a lot of it is a matter of preference.

PR #8379, opened 2017-03-13
[ios] Enable one-finger quick-zoom gesture on iPad

friedbunny:
Enables the one-finger quick-zoom gesture on iPad. We had been specifically limiting this gesture to the iPhone, but it makes sense to consistently offer this gesture on all iOS devices (as Google does).

/cc @jmkiley


friedbunny:
> though I’m not sure the gesture is actually that practical for one-handed iPad use.

This is probably true — one-handed iPad’ing can’t be a common use case (right?). I occasionally find myself using quick zoom in other situations, though, such as in smaller map views or when I don’t want my pinch-gesturing hand to obscure the screen/view.

incanus:
Agree with adding this with same caveat in https://github.com/mapbox/mapbox-gl-native/pull/8379#pullrequestreview-26625423. But if the Goog does it, 👍.

PR #8394, opened 2017-03-14
[node] Fix memory test hanging after GlyphAtlas refactoring

tmpsantos:
Replies are assumed to be asynchronous after e3500c1f791be82.


bsudekum:
Local testing looks good on my end.

PR #8396, opened 2017-03-14
[ios, macos] API doc revisions

jmkiley:
- Changed `make style-code-darwin` to `make style-code` in relevant files
- Added some doc clarifications:
   - `MGLInterpolationModeExponential` and `MGLInterpolationModeInterval` can be used with `MGLCameraStyleFunction` (I am not sure how to accurately reflect this in the documentation for `MGLCompositeStyleFunction`)
    - Do not add a style layer to the style property on a map view before the style has finished loading
    - Create individual style layers from a `MGLShapeCollectionFeature`

cc @captainbarbosa @friedbunny

### DIFF
--- a/platform/darwin/src/MGLFeature.h
+++ b/platform/darwin/src/MGLFeature.h
@@ -229,6 +229,12 @@ MGL_EXPORT
 /**
  An `MGLShapeCollectionFeature` object associates a shape collection with an
  optional identifier and attributes.
+ 
+ `MGLShapeCollectionFeature` is most commonly used to add multiple shapes to a single
+ `MGLShapeSource`. Configure the appearance of an `MGLSource`’s shape collection
+ collectively using an `MGLSymbolStyleLayer` object, or use multiple instances of
+ `MGLCircleStyleLayer`, `MGLFillStyleLayer`, and `MGLLineStyleLayer` to
+ configure the appearance of each kind of shape inside the collection.
 
  A shape collection feature is known as a
  <a href="https://tools.ietf.org/html/rfc7946#section-3.3">feature collection</a>
--- a/platform/darwin/src/MGLStyleLayer.h
+++ b/platform/darwin/src/MGLStyleLayer.h
@@ -18,6 +18,9 @@ NS_ASSUME_NONNULL_BEGIN
  subclasses of this class. Instead, create instances of
  `MGLBackgroundStyleLayer` and the concrete subclasses of
  `MGLForegroundStyleLayer`.
+ 
+ Do not add `MGLStyleLayer` objects to the style property of a `MGLMapView` before
+ `-mapView:didFinishLoadingStyle:` is called.
  */
 MGL_EXPORT
 @interface MGLStyleLayer : NSObject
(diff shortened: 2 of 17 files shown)


platform/darwin/src/MGLStyleLayer.mm.ejs
```diff
@@ -5,7 +5,7 @@
   const enumProperties = locals.enumProperties;
 -%>
 // This file is generated.
-// Edit platform/darwin/scripts/generate-style-code.js, then run `make darwin-style-code`.
+// Edit platform/darwin/scripts/generate-style-code.js, then run `make style-code`.
```
boundsj: I think that, since the generation comes from the Darwin specific script, we should not tell developers to use `style-code` since that runs the scripts for core and all platforms. [darwin-style-code](https://github.com/mapbox/mapbox-gl-native/blob/release-ios-v3.5.0-android-v5.0.0/Makefile#L268) (most of the time) probably makes the expected changes for a developer wanting to modify the Darwin platforms implementation.

platform/darwin/src/MGLStyleLayer.h.ejs
```diff
@@ -6,7 +6,7 @@
   const enumProperties = locals.enumProperties;
 -%>
 // This file is generated.
-// Edit platform/darwin/scripts/generate-style-code.js, then run `make style-code-darwin`.
+// Edit platform/darwin/scripts/generate-style-code.js, then run `make style-code`.
```
boundsj: Good catch! The make file invocation was changed previously but this was not updated to match.

However, I think we should update this comment to use the renamed [darwin-style-code](https://github.com/mapbox/mapbox-gl-native/blob/release-ios-v3.5.0-android-v5.0.0/Makefile#L268) instead of `style-code` since the latter runs codegen for core and all platforms.

jmkiley:
Changed `make style-code` to `make darwin-style-code`! 

cc @boundsj

platform/darwin/src/MGLStyleLayer.h
```diff
@@ -18,6 +18,9 @@ NS_ASSUME_NONNULL_BEGIN
  subclasses of this class. Instead, create instances of
  `MGLBackgroundStyleLayer` and the concrete subclasses of
  `MGLForegroundStyleLayer`.
+ 
+ Do not add `MGLStyleLayer` objects to the style property of a `MGLMapView` before
```
friedbunny: `style` should have code ticks around it, too, to make clear that we’re referring to a literal property.

platform/darwin/src/MGLFeature.h
```diff
@@ -229,6 +229,12 @@ MGL_EXPORT
 /**
  An `MGLShapeCollectionFeature` object associates a shape collection with an
  optional identifier and attributes.
+ 
+ `MGLShapeCollectionFeature` is most commonly used to add multiple shapes to a single
+ `MGLShapeSource`. Configure the appearance of an `MGLSource`’s shape collection
+ collectively using an `MGLSymbolStyleLayer` object, or use multiple instances of
+ `MGLCircleStyleLayer`, `MGLFillStyleLayer`, and `MGLLineStyleLayer` to
+ configure the appearance of each kind of shape inside the collection.
```
1ec5: Thanks for tying these loose threads together!

PR #8416, opened 2017-03-15
[android] - use FocalPoint for zoom

### LINKED ISSUE #8413: ZoomButtonController not zooming to center
ezgif com-video-to-gif 18

tobrun:
This PR fixes #8413 by taking in account focal point configuration or use the center of the screen when performing zoom transformations. This removes the `-1` default value from the codebase and replaces it with a concrete value as soon as possible (inside the interfaces that get called as a result of an action). This simplifies the code in `Transfrom.java` and `NativeMapView.java`.


platform/android/MapboxGLAndroidSDK/src/main/java/com/mapbox/mapboxsdk/maps/MapView.java
```diff
@@ -939,12 +942,57 @@ public void onRegisterFlingListener(MapboxMap.OnFlingListener listener) {
...
+
+    MapZoomControllerListener(MapGestureDetector detector, UiSettings uiSettings, Transform transform) {
+      this.mapGestureDetector = detector;
+      this.uiSettings = uiSettings;
+      this.transform = transform;
+    }
+
+    // Not used
+    @Override
+    public void onVisibilityChanged(boolean visible) {
```
Guardiola31337: If it isn't used, should we apply the [interface-segregation principle (ISP)](https://en.wikipedia.org/wiki/Interface_segregation_principle)?
This is only a comment (no blocker at all), but something to take into account and fix in the future.

platform/android/MapboxGLAndroidSDK/src/main/java/com/mapbox/mapboxsdk/maps/Transform.java
```diff
@@ -174,27 +174,23 @@ private long getDurationNano(long durationMs) {
     return cameraPosition.zoom;
   }
 
-  void zoom(boolean zoomIn) {
-    zoom(zoomIn, -1.0f, -1.0f);
-  }
-
-  void zoom(boolean zoomIn, float x, float y) {
+  void zoom(boolean zoomIn, float focalX, float focalY) {
```
Guardiola31337: We should start considering refactoring [data clumps](https://martinfowler.com/bliki/DataClump.html) like `focalX`/ `focalY`. Here we could [introduce a parameter object](https://refactoring.com/catalog/introduceParameterObject.html).

platform/android/MapboxGLAndroidSDK/src/main/java/com/mapbox/mapboxsdk/maps/MapView.java
tobrun: :+1:

platform/android/MapboxGLAndroidSDK/src/main/java/com/mapbox/mapboxsdk/maps/Transform.java
tobrun: :+1:

Issue #8086, opened 2017-02-16
Map change errors/failures should be specified and passed through to platform SDKs


ericrwolfe:
Currently [map.cpp](https://github.com/mapbox/mapbox-gl-native/blob/d4c569a0187d98bca6b71671fa5daf6ee09747da/src/mbgl/map/map.cpp#L1100) only returns single `mbgl::MapChangeDidFailLoadingMap` event for any and all map failures, making it impossible for a developer to distinguish between map failures. 

For example, on iOS the same error is returned whether a map load failed due to a bad access token, a bad style, or a bad network connection. 

/cc @1ec5

1ec5:
The error was implemented in #3063 to use a generic code and `nil` message because of this issue. As such, it’s debatable whether we’ve actually fixed #2762, and certainly this blocks #2775. mbgl notifications could be structs with richer information, but at that point we might as well start throwing real C++ exceptions and catching them at the SDK level, along the lines of #3063.

1ec5:
This is happening in #8418.

boundsj:
#8418 landed in the master branch.

Issue #7890, opened 2017-01-27
Replace std::runtime_error usage with more specialized exceptions


1ec5:
Per https://github.com/mapbox/mapbox-gl-native/pull/7257#discussion_r91258551, generic `std::runtime_error`s should be replaced with more specific exception types that would allow SDK code to rethrow exceptions with platform-consistent error codes. This would also enable us to provide richer diagnostic detail in exceptions.

/cc @kkaefer @jfirebaugh

PR #8418, opened 2017-03-15
Pass std::exception_ptr in MapObserver::onDidFailLoadingMap

### LINKED ISSUE #8398: -[MGLMapViewDelegate mapViewDidFailLoadingMap:withError:] error is uninformative
The error passed into `-[MGLMapViewDelegate mapViewDidFailLoadingMap:withError:]` is an uninformative error. We synthesize this generic error for compatibility with MapKit because mbgl doesn’t actually tell the SDK what went wrong. Per https://github.com/mapbox/mapbox-gl-native/pull/8377#discussion_r105758312, `mbgl::MapObserver::onDidFailLoadingMap()` should accept the error or error message as a parameter. This would allow all the SDKs – not only the iOS and macOS SDKs – to provide more information when the map fails to load for some reason.

/cc @brunoabinader @fabian-guerra

brunoabinader:
Forwards the ~error type and message~ `std::exception_ptr` from style::Observer::onStyleError as well as non-style related errors.

As per @1ec5 comments in #8398:
> This would allow all the SDKs – not only the iOS and macOS SDKs – to provide more information when the map fails to load for some reason.

Fixes #8398.

### DIFF
--- a/platform/ios/src/MGLMapView.mm
+++ b/platform/ios/src/MGLMapView.mm
@@ -22,6 +22,7 @@
 #include <mbgl/style/layers/custom_layer.hpp>
 #include <mbgl/map/backend.hpp>
 #include <mbgl/math/wrap.hpp>
+#include <mbgl/util/exception.hpp>
 #include <mbgl/util/geo.hpp>
 #include <mbgl/util/constants.hpp>
 #include <mbgl/util/image.hpp>
@@ -30,6 +31,7 @@
 #include <mbgl/util/chrono.hpp>
 #include <mbgl/util/run_loop.hpp>
 #include <mbgl/util/shared_thread_pool.hpp>
+#include <mbgl/util/string.hpp>
 
 #import "Mapbox.h"
 #import "MGLFeature_Private.h"
@@ -4794,7 +4796,7 @@ public:
     }
 }
 
-- (void)willStartLoadingMap {
+- (void)mapViewWillStartLoadingMap {
     if (!_mbglMap) {
         return;
     }
@@ -4805,7 +4807,7 @@ public:
     }
 }
 
-- (void)didFinishLoadingMap {
+- (void)mapViewDidFinishLoadingMap {
     if (!_mbglMap) {
         return;
     }
@@ -4820,19 +4822,18 @@ public:
     }
 }
 
-- (void)didFailLoadingMap {
+- (void)mapViewDidFailLoadingMapWithError:(NSError *)error {
     if (!_mbglMap) {
         return;
     }
 
     if ([self.delegate respondsToSelector:@selector(mapViewDidFailLoadingMap:withError:)])
     {
-        NSError *error = [NSError errorWithDomain:MGLErrorDomain code:0 userInfo:nil];
         [self.delegate mapViewDidFailLoadingMap:self withError:error];
     }
 }
 
-- (void)willStartRenderingFrame {
+- (void)mapViewWillStartRenderingFrame {
     if (!_mbglMap) {
         return;
     }
@@ -4843,7 +4844,7 @@ public:
     }
 }
 
-- (void)didFinishRenderingFrameFullyRendered:(BOOL)fullyRendered {
+- (void)mapViewDidFinishRenderingFrameFullyRendered:(BOOL)fullyRendered {
     if (!_mbglMap) {
         return;
     }
@@ -4861,7 +4862,7 @@ public:
     }
 }
 
-- (void)willStartRenderingMap {
+- (void)mapViewWillStartRenderingMap {
     if (!_mbglMap) {
         return;
     }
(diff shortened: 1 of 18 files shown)


brunoabinader:
@1ec5 does the new entries in `MGLErrorCode` requires updating `CHANGELOG.md`?

jfirebaugh:
What do you think about using `std::exception_ptr` as the error type rather than `ErrorType` + `std::string`? This would match our convention in other places, e.g. `style::Observer::on{Glyphs,Sprite,Source,Tile,Resource}Error`.

platform/ios/src/MGLMapView.mm
```diff
@@ -5375,33 +5374,50 @@ void onCameraDidChange(mbgl::MapObserver::CameraChangeMode mode) override {
...
+        MGLErrorCode errorCode;
+        switch (type) {
+        case mbgl::MapObserver::ErrorType::ParseStyle:
+            errorCode = MGLErrorCodeParseStyleFailed;
+            break;
+        case mbgl::MapObserver::ErrorType::NotFound:
+            errorCode = MGLErrorCodeNotFound;
+            break;
+        case mbgl::MapObserver::ErrorType::LoadingStyle:
+            errorCode = MGLErrorCodeLoadStyleFailed;
```
1ec5: Would either `MGLErrorCodeBadServerResponse` or `MGLErrorCodeConnectionFailed` describe this failure, or is `MGLErrorCodeLoadStyleFailed` a distinct case?

platform/ios/src/MGLMapView.mm
```diff
@@ -5375,33 +5374,50 @@ void onCameraDidChange(mbgl::MapObserver::CameraChangeMode mode) override {
...
+            break;
+        }
+        error = [NSError errorWithDomain:MGLErrorDomain code:errorCode userInfo:errorDescription ? @{
+            NSLocalizedDescriptionKey: errorDescription,
```
1ec5: Thank you so much for this! 👍 It turns out the [error user info dictionary](https://developer.apple.com/library/content/documentation/Cocoa/Conceptual/ErrorHandlingCocoa/ErrorObjectsDomains/ErrorObjectsDomains.html#//apple_ref/doc/uid/TP40001806-CH202-SW3) is called that because its contents are often displayed to the user with a localized description and sometimes a recovery suggestion. Here are some suggestions for improvements, taking advantage of the error type and assuming `message` comes directly from an `std::exception_ptr`, as suggested in https://github.com/mapbox/mapbox-gl-native/pull/8418#issuecomment-286820676:

* `ParseStyle`:
  * `NSLocalizedDescriptionKey`: “The map failed to load because the style is corrupted.”
  * `NSLocalizedFailureReasonErrorKey`: `@(message.c_str())`
* `NotFound`:
  * `NSLocalizedDescriptionKey`: “The map failed to load because the style can’t be found or is incompatible.”
  * `NSLocalizedFailureReasonErrorKey`: `@(message.c_str())`
* `LoadingStyle`:
  * `NSLocalizedDescriptionKey`: `@(message.c_str())`

For the localized descriptions, we can use the `NSLocalizedStringWithDefaultValue()` macro (as described [here](https://github.com/mapbox/mapbox-gl-native/blob/master/platform/ios/DEVELOPING.md#adding-user-facing-text)) to potentially localize the error message.

Let me know if you’d prefer that I make these changes myself.

1ec5:
This fixes #8086, which I previously missed before filing #8398 as a duplicate.

platform/ios/src/MGLMapView.mm
brunoabinader: I'd prefer for map loading failure to have its own error type. However, we're no longer distinctively separating these in three different error types because `std::runtime_error` only provides a string.

platform/ios/src/MGLMapView.mm
brunoabinader: > Would that means we’d have different subclasses of std::exception_ptr for each failure mode, or would everything be an std::runtime_error?

Good question - it seems we're using only plain `std::runtime_error` everywhere in GL native code, so I'd say let's keep the same pattern.

As explained above, we'll no longer be able to distinct between the three error types - however I don't think that would make much difference anyways - the error message is the most important debugging information we need.

1ec5: We do have a couple subclasses of `mbgl::util::Exception` meant to provide more distinguishable errors. Currently, the only subclasses appear to be `SpriteImageException` and `MisuseException`; could we add more subclasses for these error types?

brunoabinader: > could we add more subclasses for these error types?

I don't have a strong opinion on this, though for this approach to work we'd need to deduct the exception type via `catch` - see the exception juggling in https://github.com/mapbox/mapbox-gl-native/blob/master/include/mbgl/util/string.hpp#L44-L51 - for all platform-specific codes that needed to handle this.

1ec5: For the iOS and macOS implementations, the catch statements would replace the existing switch statement. For other implementations that don't need to distinguish individual failure reasons, catching a runtime_error would be sufficient.

kkaefer:
related: https://github.com/mapbox/mapbox-gl-native/issues/7890

PR #8425, opened 2017-03-15
[ios, macos] Add more information to interpolation mode docs

jmkiley:
Addresses #8192 until #8354 lands.

cc @boundsj @friedbunny @1ec5


platform/darwin/src/MGLStyleValue.h
```diff
@@ -54,23 +54,29 @@ extern MGL_EXPORT const MGLStyleFunctionOption MGLStyleFunctionOptionDefaultValu
...
      Values between two stops are not interpolated. Instead, properties are set
-     to the value of the stop equal to the function input's key value.
+     to the value of the stop equal to the function input's key value. Use
+     categorical interpolation mode to show values that fit into categories.
      */
     MGLInterpolationModeCategorical,
     /**
      Values between two stops are not interpolated. Instead, values are set 
-     to their input value.
+     to their input value. Use to show values that can be used as style values.
```
boundsj: nit: I suggest changing this to "Use **identity interpolation mode** to show ..." just for consistency with the other values.

Issue #8035, opened 2017-02-13
Reduce binary size


kkaefer:
Looking at our [binary size tracker](https://mapbox.github.io/mapbox-gl-native/metrics/binary-size/), it looks like we got a massive increase on Feb 2/3, which seems to coincide with the merge of https://github.com/mapbox/mapbox-gl-native/pull/7372. In some cases like Android arm-v7/8, the increase is as much as 1 MB. We should investigate ways of making the API leaner, and check whether there are certain templates that cause a huge amount of code to be generated.

/cc @jfirebaugh

jfirebaugh:
Do you know of any tools for profiling binary size?

kkaefer:
There's https://github.com/google/bloaty, but I've found it to be of limited use since most (~80-90%) of the pie is typically attributed to `[None]`. Looking at the `*.o` files, any code that uses RapidJSON or the conversion headers seems to be quite heavy.

jfirebaugh:
Looked into this on macOS and I can reproduce the ~500kb increase in Mapbox.framework between 6a6bddb and 22b3834. I have build directories for both of those commits, but I'm not sure how to identify why exactly the size increase occurs or how to reduce it. It looks like I'd have to start writing some custom tooling that does Mach-O binary reversing combined with analysis of the corresponding dSYM. (Apple's own [Mach-O format documentation](https://developer.apple.com/library/content/documentation/DeveloperTools/Conceptual/MachORuntime/index.html) is now a 404.)

1ec5:
Does [this article](https://developer.apple.com/library/content/documentation/DeveloperTools/Conceptual/MachOTopics/0-Introduction/introduction.html) have the Mach-O documentation you need?

jfirebaugh:
Tried switching to [mpark/variant](https://github.com/mpark/variant), it was a pessimization:

```
     VM SIZE                               FILE SIZE
 ++++++++++++++ GROWING                 ++++++++++++++
  +481%  +163Ki __TEXT,__objc_methtype   +163Ki  +481%
  +1.7% +45.0Ki __TEXT,__text           +45.0Ki  +1.7%
  +157% +37.0Ki __DATA,__const          +37.0Ki  +157%
  +5.7% +16.4Ki __TEXT,__gcc_except_tab +16.4Ki  +5.7%
  +6.4% +10.3Ki [None]                  +11.6Ki  +6.8%
[...]

PR #8430, opened 2017-03-15
[core] Replace Result<T> with optional<T> plus out Error parameter

jfirebaugh:
😞 , but `variant` is just too heavy in code-size to be instantiated with so many different template parameters.

```

  -1.3% -32.5Ki __TEXT,__text           -32.5Ki  -1.3%
  -5.3% -15.4Ki __TEXT,__gcc_except_tab -15.4Ki  -5.3%
 -10.2% -6.64Ki __TEXT,__unwind_info    -6.64Ki -10.2%
  -3.1% -5.10Ki [None]                  -3.11Ki  -1.8%
  -0.1%    -208 __TEXT,__const             -208  -0.1%
  -0.1%    -131 __TEXT,__cstring           -131  -0.1%

  -1.7% -60.0Ki TOTAL                   -58.0Ki  -1.7%
```

Refs #8035.


brunoabinader:
Sad to see this Rust-like paradigm go 😢

benchmark/parse/filter.benchmark.cpp
```diff
@@ -14,7 +14,8 @@ using namespace mbgl;
 style::Filter parse(const char* expression) {
     rapidjson::GenericDocument<rapidjson::UTF8<>, rapidjson::CrtAllocator> doc;
     doc.Parse<0>(expression);
-    return *style::conversion::convert<style::Filter, JSValue>(doc);
+    style::conversion::Error error;
+    return *style::conversion::convert<style::Filter, JSValue>(doc, error);
```
kkaefer: This produces a 💥 if conversion fails. Given that this is a test case, can we just add a `ASSERT_TRUE`, which fails hard when there is no conversion result?

test/style/filter.test.cpp
```diff
@@ -16,7 +16,8 @@ using namespace mbgl::style;
 Filter parse(const char * expression) {
     rapidjson::GenericDocument<rapidjson::UTF8<>, rapidjson::CrtAllocator> doc;
     doc.Parse<0>(expression);
-    return *conversion::convert<Filter, JSValue>(doc);
+    conversion::Error error;
+    return *conversion::convert<Filter, JSValue>(doc, error);
```
kkaefer: Same as above

jfirebaugh:
> Sad to see this Rust-like paradigm go 😢

Me too. Looking at the code that is produced with `variant`, I find it hard not to come to the conclusion that a discriminated union type really needs to be something built into the language, rather than a library-level feature.

I also tried https://github.com/mpark/variant/, which is the implementation libc++'s `std::variant` uses, and it generated even larger code sizes.

benchmark/parse/filter.benchmark.cpp
jfirebaugh: This is used in the inner loop of one of the two benchmarks, so don't we want to avoid any error checking overhead?

PR #8431, opened 2017-03-15
Render fill-extrusion layers

lbud:
Opening this for tracking purposes as it's getting close enough to have a public living to-do list:

- [x] `[functionality]` implement transitions in light options
- [x] `[bug]` In https://github.com/mapbox/mapbox-gl-native/commit/9514b62f1e50e75ff806f2397b37730ba92ce72a I rewrote `RenderItem` to hold a vector of `RenderTile`s, rather than a single `RenderTile`, so that we can render a whole layer at a time (necessary for the offscreen texture approach used herein). In doing so I wrote a logic bug — need to comb over that rewrite 🔍 
- [x] `[bug]` ~I suspect there may be a bug in the way we're handling zoom-and-property functions — investigate~ punted to https://github.com/mapbox/mapbox-gl-native/issues/8460 as it is not actually related
- [x] `[bug]` there are weird striping/zigzag artifacts on glfw and macOS, but not on ios. Need to investigate what's going on here 
  * (update: was caused by using a 16-bit depth component resolution `GL_DEPTH_COMPONENT16`; changed to `GL_DEPTH_COMPONENT` which leaves resolution up to the implementation) 
  * (update 2: this fixed the issue on MacOS/glfw but broke on iOS as `GL_DEPTH_COMPONENT16` is the only valid depth internalformat for renderbufferStorage :thinking: — onward)
  * (resolution: GL version-specific)
- [x] `[tests]` un-ignore render tests + ensure all pass https://github.com/mapbox/mapbox-gl-js/pull/4519 🚀 
- [x] ☝️ merge gl-js PR in sync with this
- [x] `[tests]` unit tests
- [x] `[cleanup]` this depends on the [`extrusions-native-tweaks`](https://github.com/mapbox/mapbox-gl-js/compare/extrusions-native-tweaks) branch of mapbox-gl-js, which makes a few changes to shaders — merge that + update dep here
- [x] `[bug]` many render tests are broken (opacity issue?); investigate
- [x] ~`[cleanup]` squash/clean up commits (apologies to reviewers 🙈)~ lost cause, will squash upon merging
- [x] `[cleanup]` revert committed changes in `./platform/glfw/main.cpp`
- [x] `[platform]` address @1ec5's comments below + corresponding concerns across other generated platform code


platform/darwin/src/MGLFillExtrusionStyleLayer.h
```diff
@@ -0,0 +1,281 @@
...
+   * `MGLInterpolationModeExponential`
+   * `MGLInterpolationModeInterval`
+   * `MGLInterpolationModeCategorical`
+ */
+@property (nonatomic, null_resettable) MGLStyleValue<NSNumber *> *fillExtrusionBase;
+
+#if TARGET_OS_IPHONE
+/**
+ The base color of the extruded fill. The extrusion's surfaces will be shaded
+ differently based on this color in combination with the root `light` settings.
```
1ec5: `light` will in all likelihood be a property of MGLStyle; the word "root" is meaningless in the context of these SDKs.

platform/darwin/src/MGLFillExtrusionStyleLayer.h
```diff
@@ -0,0 +1,281 @@
...
+ You can access an existing fill-extrusion style layer using the
+ `-[MGLStyle layerWithIdentifier:]` method if you know its identifier;
+ otherwise, find it using the `MGLStyle.layers` property. You can also create a
+ new fill-extrusion style layer and add it to the style using a method such as
+ `-[MGLStyle addLayer:]`.
+
+ ### Example
+
+ ```swift
+ ```
```
1ec5: Eventually we'll want to fill in some example code here, via MGLDocumentationTests.swift, but that can be tail work.

platform/darwin/src/MGLFillExtrusionStyleLayer.h
```diff
@@ -0,0 +1,281 @@
...
+    /**
+     The fill extrusion is translated relative to the viewport.
+     */
+    MGLFillExtrusionTranslateAnchorViewport,
+};
+
+/**
+ An extruded (3D) polygon.
+
+ You can access an existing fill-extrusion style layer using the
```
1ec5: s/fill-extrusion/fill extrusion/

anandthakker:
> I suspect there may be a bug in the way we're handling zoom-and-property functions — investigate

FYI, our handling of these underwent some recent changes (by @mollymerp and I) - happy to help 💭 / 🕵️

lbud:
Thanks for the 👀 @1ec5! All platform-specific code committed thus far has been autogenerated with scripts/make targets so your attention is especially appreciated 🙇‍♀️ — adding to the list.

platform/darwin/src/MGLFillExtrusionStyleLayer.h
lbud: @1ec5 I am of the opinion (though not strongly) that this should remain hyphenated, I think, for clarity — do you have a strong feeling that it should be `fill extrusion`?

1ec5: Yes, as a noun phrase, it should be unhyphenated in prose. ("Fill-extrusion style layer" would be more natural as shorthand for "fill and extrusion style layer", like "zoom-property function" as shorthand for "zoom-and-property function".) Unlike in JavaScript, the hyphenated form `fill-extrusion` will never appear in the developer's code.

platform/darwin/scripts/style-spec-overrides-v8.json
```diff
@@ -50,6 +53,17 @@
...
   },
+  "paint_fill-extrusion": {
+    "fill-extrusion-pattern": {
+      "doc": "Name of image in style images to use for drawing image fill-extrusions. For seamless patterns, image width and height must be a factor of two (2, 4, 8, ..., 512)."
+    },
+    "fill-extrusion-translate": {
+      "doc": "The geometry's offset."
+    },
+    "fill-extrusion-color": {
+      "doc": "The base color of this layer. The extrusion's surfaces will be shaded differently based on this color in combination with the `light` settings. If this color is specified with an alpha component, the alpha component will be ignored; use `fill-extrusion-opacity` to set layer opacityco."
```
1ec5: Typo: "opacityco".

platform/darwin/scripts/style-spec-overrides-v8.json
```diff
@@ -50,6 +53,17 @@
       "doc": "Name of image in style images to use for drawing image fills. For seamless patterns, image width and height must be a factor of two (2, 4, 8, ..., 512)."
     }
   },
+  "paint_fill-extrusion": {
+    "fill-extrusion-pattern": {
+      "doc": "Name of image in style images to use for drawing image fill-extrusions. For seamless patterns, image width and height must be a factor of two (2, 4, 8, ..., 512)."
```
1ec5: "Image fills" sort of made sense for `fill-pattern`, but I'm not sure about "image fill-extrusions". How about "for drawing the surfaces of the extrusion"?

lbud:
@jfirebaugh @kkaefer I'm trying to figure out one last thing here: there's a failing TileCover test on a few Bitrise platforms (though passing on Travis) — the returned tiles the same but are slightly out of order; this is the result of the [changes in `transform_state.cpp`](https://github.com/mapbox/mapbox-gl-native/pull/8431/files#diff-928311be4c7a1a2c0abb7b1b023aad53) to scale the projection matrix vertically to meters per pixel (corresponding gl-js change: https://github.com/mapbox/mapbox-gl-js/pull/3509).

Aside from that however this is ready for review (so please start reviewing whenever you feel ready) — the only thing that _may_ change is that small part of `transform_state.cpp`. I'm not sure if this would be easier to review if I squashed or rearranged some commits, or if this is reviewable as-is — if I can make it easier for you let me know 🙏

lbud:
Also tagging in @ivovandongen @1ec5 to do another pass over generated Darwin/Android code 🙏

include/mbgl/style/conversion/constant.hpp
```diff
@@ -113,6 +134,27 @@ struct Converter<std::array<float, 2>> {
 };
 
 template <>
+struct Converter<std::array<float, 3>> {
```
jfirebaugh: I've been meaning to generalize this to a single templated `Converter<std::array<float, N>>` that can handle any size, and now's a good time. If you rebase on top of https://github.com/mapbox/mapbox-gl-native/commit/7c90db9241b47cb2b0a3ecd6a4d1df2fa1fd3f54 you can remove this addition.

include/mbgl/style/light.hpp
```diff
@@ -0,0 +1,294 @@
...
+#include <mbgl/util/indexed_tuple.hpp>
+#include <mbgl/util/ignore.hpp>
+
+#include <memory>
+
+namespace mbgl {
+namespace style {
+
+template <class Value>
+class UnevaluatedLightProperty {
```
jfirebaugh: Rather than introduce a new class, let's generalize the existing `UnevaluatedPaintProperty` so that it can be used here. The functionality is solid, it just needs to be extracted from `paint_property.hpp` into its own header, and given a name that does not tie it to paint properties specifically. I suggest `TransitioningProperty`, since its main responsibility is to interpolate between a current and prior value based on the current clock and transition options.

include/mbgl/style/conversion/constant.hpp
jfirebaugh: Oops, missed a test update. 49f4d7b779341bfa0146384d099f063a25ae13fb

include/mbgl/style/light.hpp
```diff
@@ -0,0 +1,294 @@
...
+
+private:
+    optional<mapbox::util::recursive_wrapper<UnevaluatedLightProperty<Value>>> prior;
+    TimePoint begin;
+    TimePoint end;
+    Value value;
+};
+
+template <class Value>
+class CascadingLightProperty {
```
jfirebaugh: Per chat, light properties support transitions but not style classes, so this class can be simplified, and "Cascading" removed from the name. It just serves to tie the `Value` and `TransitionOptions` together.

include/mbgl/util/position.hpp
```diff
@@ -0,0 +1,69 @@
+#pragma once
```
jfirebaugh: This file defines a class in the `mbg::style` namespace, so place it in the `mbgl/style` directory.

include/mbgl/util/position.hpp
```diff
@@ -0,0 +1,69 @@
...
+
+#include <mbgl/style/types.hpp>
```
jfirebaugh: Trim the includes here to just what's needed -- looks like `constants.hpp` and `array`.

include/mbgl/util/position.hpp
```diff
@@ -0,0 +1,69 @@
...
+
+#include <cassert>
+#include <string>
+
+namespace mbgl {
+namespace style {
+class Position {
+public:
+    Position() = default;
+    Position(std::array<float, 3> position_)
```
jfirebaugh: Take `std::array<float, 3>` inputs by const reference (also in `set`).

include/mbgl/util/position.hpp
```diff
@@ -0,0 +1,69 @@
...
+    friend bool operator==(const Position& lhs, const Position& rhs) {
+        return lhs.radial == rhs.radial && lhs.azimuthal == rhs.azimuthal && lhs.polar == rhs.polar;
+        // TODO should this address wrapping? or should somewhere else in Position address wrapping?
+    }
+
+    friend bool operator!=(const Position& lhs, const Position& rhs) {
+        return !(lhs == rhs);
+    }
+
+    const std::array<float, 3> get() const {
```
jfirebaugh: `getCartesian`?

include/mbgl/util/position.hpp
```diff
@@ -0,0 +1,69 @@
...
+private:
+    float radial;
+    float azimuthal;
+    float polar;
+    float x;
+    float y;
+    float z;
+
+    void calculateCartesian() {
+        auto _a = azimuthal;
```
jfirebaugh: Write the calculations as a one liner, e.g. `const float _a = (azimuthal + 90) * util::DEG2RAD`.

platform/macos/docs/guides/For Style Authors.md
```diff
@@ -243,6 +244,37 @@ In style JSON | In Objective-C | In Swift
 `fill-translate` | `MGLFillStyleLayer.fillTranslation` | `MGLFillStyleLayer.fillTranslation`
 `fill-translate-anchor` | `MGLFillStyleLayer.fillTranslationAnchor` | `MGLFillStyleLayer.fillTranslationAnchor`
 
+### FillExtrusion style functions
```
1ec5: This section will go away once #8526 lands in master as part of #8672. If that PR lands first, you can rerun `make darwin-style-code` to eliminate it; otherwise, @boundsj can take care of that in #8672.

platform/glfw/main.cpp
```diff
@@ -167,9 +165,10 @@ int main(int argc, char *argv[]) {
...
     if (style.empty()) {
-        const char *url = getenv("MAPBOX_STYLE_URL");
+        const char* url = getenv("MAPBOX_STYLE_URL");
         if (url == nullptr) {
-            mbgl::util::default_styles::DefaultStyle newStyle = mbgl::util::default_styles::orderedStyles[0];
+            mbgl::util::default_styles::DefaultStyle newStyle =
+                mbgl::util::default_styles::orderedStyles[0];
             style = newStyle.url;
             view->setWindowTitle(newStyle.name);
         } else {
```
kkaefer: While all of the formatting changes in main.cpp are good, they aren't related to the pull request. If it's not too complicated, removing them might increase the reviewability.

src/mbgl/gl/context.hpp
```diff
@@ -81,7 +81,10 @@ class Context : private util::noncopyable {
 
     template <RenderbufferType type>
     Renderbuffer<type> createRenderbuffer(const Size size) {
-        static_assert(type == RenderbufferType::RGBA || type == RenderbufferType::DepthStencil,
+        static_assert(type == RenderbufferType::RGBA ||
+                      type == RenderbufferType::DepthStencil ||
+                      type == RenderbufferType::DepthComponent ||
+                      type == RenderbufferType::RGBA4,  // TODO I added RGBA4 because in JS RGBA8 (here 'RGBA') emits a "renderbufferStorage: invalid internalformat" warning (though still renders fine) -- investigate
```
kkaefer: For some reason, `glRenderbufferStorage` doesn't support a RGBA8 format by default. For OpenGL ES 2.0, there [is an extension](https://www.khronos.org/registry/OpenGL/extensions/OES/OES_rgb8_rgba8.txt) that seems widely supported (and probably the reason why we haven't run into this issue on mobile devices before). For WebGL, I haven't found an extension.

The issue with RGB4 is that it uses less precision (only 4 bits per channel), so we'll end up with less precise colors. An alternative might be to render to a texture rather than a Renderbuffer.

src/mbgl/gl/context.cpp
```diff
@@ -384,6 +384,22 @@ Framebuffer Context::createFramebuffer(const Texture& color) {
...
+        throw new std::runtime_error("Renderbuffer size mismatch");
+    }
+    auto fbo = createFramebuffer();
+    bindFramebuffer = fbo;
+    MBGL_CHECK_ERROR(glFramebufferRenderbuffer(GL_FRAMEBUFFER, GL_COLOR_ATTACHMENT0, GL_RENDERBUFFER, colorTarget.renderbuffer));
+    MBGL_CHECK_ERROR(glFramebufferRenderbuffer(GL_FRAMEBUFFER, GL_DEPTH_ATTACHMENT, GL_RENDERBUFFER, depthTarget.renderbuffer));
+    MBGL_CHECK_ERROR(glFramebufferTexture2D(GL_FRAMEBUFFER, GL_COLOR_ATTACHMENT0, GL_TEXTURE_2D, fboTexture.texture, 0));
+    checkFramebuffer();
+    return { colorTarget.size, std::move(fbo) };
+}
```
kkaefer: Rather than adding a separate function for every enum, we should change the code to have a private function that accepts an RenderbufferID, and move the validation to the header file.

src/mbgl/gl/context.cpp
```diff
@@ -384,6 +384,22 @@ Framebuffer Context::createFramebuffer(const Texture& color) {
...
+                           const Renderbuffer<RenderbufferType::DepthComponent>& depthTarget,
+                           const Texture& fboTexture) {
+    if (colorTarget.size != depthTarget.size) {
+        throw new std::runtime_error("Renderbuffer size mismatch");
+    }
+    auto fbo = createFramebuffer();
+    bindFramebuffer = fbo;
+    MBGL_CHECK_ERROR(glFramebufferRenderbuffer(GL_FRAMEBUFFER, GL_COLOR_ATTACHMENT0, GL_RENDERBUFFER, colorTarget.renderbuffer));
+    MBGL_CHECK_ERROR(glFramebufferRenderbuffer(GL_FRAMEBUFFER, GL_DEPTH_ATTACHMENT, GL_RENDERBUFFER, depthTarget.renderbuffer));
+    MBGL_CHECK_ERROR(glFramebufferTexture2D(GL_FRAMEBUFFER, GL_COLOR_ATTACHMENT0, GL_TEXTURE_2D, fboTexture.texture, 0));
```
kkaefer: Here, `GL_COLOR_ATTACHMENT0` is first bound to the renderbuffer, then to the texture. This means the renderbuffer isn't used at all, and we can directly call `createFramebuffer(texture, depth)`

src/mbgl/util/offscreen_texture.cpp
```diff
@@ -36,15 +39,31 @@ class OffscreenTexture::Impl {
...
+        context.viewport = { 0, 0, size };
+    }
+
 private:
     gl::Context& context;
     const Size size;
+    gl::TextureUnit unit;
     optional<gl::Framebuffer> framebuffer;
     optional<gl::Texture> texture;
+    optional<gl::Renderbuffer<gl::RenderbufferType::RGBA4>> colorTarget;
```
kkaefer: Why do we add an additional color target? The point of this class is to draw to the texture.

src/mbgl/util/offscreen_texture.cpp
```diff
@@ -36,15 +39,31 @@ class OffscreenTexture::Impl {
...
+    optional<gl::Renderbuffer<gl::RenderbufferType::DepthComponent>> depthTarget;
 };
 
-OffscreenTexture::OffscreenTexture(gl::Context& context, const Size size)
-    : impl(std::make_unique<Impl>(context, std::move(size))) {
+OffscreenTexture::OffscreenTexture(gl::Context& context, const Size size, gl::TextureUnit unit)
+    : impl(std::make_unique<Impl>(context, std::move(size), unit)) {
```
kkaefer: Instead of adding the `unit` parameter to the constructor (where it isn't used), we should add it to the `bind` call to be in line with other uses of this parameter. The texture unit is a render-time property, and not a property of the texture object.

src/mbgl/gl/context.hpp
kkaefer: Given that we don't seem to need a color renderbuffer (since we're only rendering to a texture), we can remove this and not worry about initialization OpenGL extensions.

platform/darwin/src/MGLStyle.mm
```diff
@@ -327,6 +328,9 @@ - (MGLStyleLayer *)layerFromMBGLLayer:(mbgl::style::Layer *)mbglLayer
     if (auto fillLayer = mbglLayer->as<mbgl::style::FillLayer>()) {
         MGLSource *source = [self sourceWithIdentifier:@(fillLayer->getSourceID().c_str())];
         styleLayer = [[MGLFillStyleLayer alloc] initWithIdentifier:identifier source:source];
+    } else if (auto fillExtrusionLayer = mbglLayer->as<mbgl::style::FillExtrusionLayer>()) {
+        MGLSource *source = [self sourceWithIdentifier:@(fillExtrusionLayer->getSourceID().c_str())];
+        styleLayer = [[MGLLineStyleLayer alloc] initWithIdentifier:identifier source:source];
```
1ec5: This should be MGLFillExtrusionStyleLayer, not MGLLineStyleLayer.

lbud: oops 🤖 🙏 ty

test/util/position.test.cpp
```diff
@@ -0,0 +1,49 @@
+#include <mbgl/test/util.hpp>
+
+#include <mbgl/style/types.hpp>
+#include <mbgl/style/position.hpp>
+#include <mbgl/util/constants.hpp>
+
+using namespace mbgl;
+using namespace style;
+
+auto ARR_EQ = [](std::array<float, 3> got, std::array<float, 3> expected) {
```
jfirebaugh: Write these as regular functions, e.g. `void expectElementsNear(...)`.

src/mbgl/programs/extrusion_texture_program.hpp
```diff
@@ -0,0 +1,48 @@
...
+#include <mbgl/shaders/extrusion_texture.hpp>
+#include <mbgl/style/layers/fill_extrusion_layer_properties.hpp>
+#include <mbgl/util/geometry.hpp>
+
+#include <string>
+
+namespace mbgl {
+
+struct ExtrusionTextureLayoutAttributes : gl::Attributes<
+attributes::a_pos>
```
jfirebaugh: Nit: indent the template parameter lists by one level (⨉3 here, ⨉5 in `fill_extrusion_program.hpp`).

src/mbgl/programs/extrusion_texture_program.hpp
```diff
@@ -0,0 +1,48 @@
...
+uniforms::u_matrix,
+uniforms::u_world,
+uniforms::u_image,
+uniforms::u_opacity>
+{};
+
+class ExtrusionTextureProgram : public Program<
+shaders::extrusion_texture,
+gl::Triangle,
+ExtrusionTextureLayoutAttributes,
```
jfirebaugh: If the type names aren't needed anywhere else, go ahead and inline `ExtrusionTextureLayoutAttributes` and `ExtrusionTextureUniforms`:

```C++
class ExtrusionTextureProgram : public Program<
    shaders::extrusion_texture
    gl::Triangle,
    gl::Attributes<
        attributes::a_pos>,
    gl::Uniforms<
        uniforms::u_matrix,
        uniforms::u_world,
        uniforms::u_image,
        uniforms::u_opacity>,
    style::PaintProperties<>>
```

src/mbgl/programs/extrusion_texture_program.hpp
```diff
@@ -0,0 +1,48 @@
+#pragma once
+
+#include <mbgl/programs/program.hpp>
+#include <mbgl/programs/attributes.hpp>
+#include <mbgl/programs/uniforms.hpp>
+#include <mbgl/shaders/extrusion_texture.hpp>
+#include <mbgl/style/layers/fill_extrusion_layer_properties.hpp>
```
jfirebaugh: This include is probably unused here.

src/mbgl/programs/extrusion_texture_program.hpp
```diff
@@ -0,0 +1,48 @@
...
+#include <mbgl/util/geometry.hpp>
+
+#include <string>
```
jfirebaugh: Unused.

src/mbgl/programs/fill_extrusion_program.cpp
```diff
@@ -0,0 +1,49 @@
...
+                                     const std::array<float, 3>& lightPos,
+                                     const float lightIntensity) {
+    int32_t tileSizeAtNearestZoom =
+        util::tileSize * state.zoomScale(state.getIntegerZoom() - tileID.canonical.z);
+    int32_t pixelX = tileSizeAtNearestZoom *
+                     (tileID.canonical.x + tileID.wrap * state.zoomScale(tileID.canonical.z));
+    int32_t pixelY = tileSizeAtNearestZoom * tileID.canonical.y;
+
+    return FillExtrusionPatternUniforms::Values{
+        uniforms::u_matrix::Value{ matrix }, uniforms::u_pattern_tl_a::Value{ a.tl },
```
jfirebaugh: Put each parameter on a separate line. That way parallel things line up, e.g.:

```
        uniforms::u_pattern_tl_a::Value{ a.tl }
        uniforms::u_pattern_br_a::Value{ a.br }
```

src/mbgl/programs/fill_extrusion_program.cpp
```diff
@@ -0,0 +1,49 @@
...
+
+FillExtrusionPatternUniforms::Values
+FillExtrusionPatternUniforms::values(mat4 matrix,
+                                     const SpriteAtlasElement& a,
+                                     const SpriteAtlasElement& b,
+                                     const Faded<std::string>& fading,
+                                     const UnwrappedTileID& tileID,
+                                     const TransformState& state,
+                                     const float heightFactor,
+                                     const std::array<float, 3>& lightColor,
```
jfirebaugh: Can this be a `Color`, `lightPos` be a `Position`? Or could all three of `lightColor`, `lightPos`, and `lightIntensity` be passed as one `Light::Evaluated` value?

src/mbgl/renderer/painter.cpp
```diff
@@ -87,12 +99,15 @@ Painter::Painter(gl::Context& context_,
       state(state_),
       tileVertexBuffer(context.createVertexBuffer(tileVertices())),
       rasterVertexBuffer(context.createVertexBuffer(rasterVertices())),
+      extrusionTextureVertexBuffer(context.createVertexBuffer(extrusionTextureVertices())),
       tileTriangleIndexBuffer(context.createIndexBuffer(tileTriangleIndices())),
-      tileBorderIndexBuffer(context.createIndexBuffer(tileLineStripIndices())) {
+      tileBorderIndexBuffer(context.createIndexBuffer(tileLineStripIndices())),
+      extrusionTextureTriangleIndexBuffer(context.createIndexBuffer(tileTriangleIndices())) {
```
jfirebaugh: Since this contains the same data as `tileTriangleIndexBuffer`, let's reuse that buffer. We can rename to `quadTriangleIndexBuffer` / `quadTriangleIndices` for generality.

src/mbgl/gl/context.hpp
lbud: 😳 you're right @kkaefer — removing 🙇‍♀️

src/mbgl/gl/context.cpp
lbud: @kkaefer I've removed the color renderbuffer entirely but am not entirely sure what you mean here — are you talking about reusing the existing 
``` cpp
Framebuffer
Context::createFramebuffer(const Texture& color,
                           const Renderbuffer<RenderbufferType::DepthStencil>& depthStencil) {}
```
function? I think this won't work because that's using depth/stencil and this is just using depth, right?

lbud:
@ivovandongen I wasn't able to reproduce the [assertion failure](https://github.com/mapbox/mapbox-gl-native/blob/extrusions/src/mbgl/renderer/fill_extrusion_bucket.cpp#L77-L78) you mentioned on Android the other day (I tested the same polygon on iOS). If you can reproduce it, would you mind setting a breakpoint to see what `triangleIndex` and `totalVertices` are when it fails? (When testing that single polygon on iOS they were 0 and 5, respectively, which is expected…)

lbud:
@jfirebaugh I reworked uniform initialization to take just a single `Light::Evaluated` argument in https://github.com/mapbox/mapbox-gl-native/pull/8431/commits/63ec39c9302154aebeb2cd6ae709f1061c81c464 — wanted to flag for visibility to make sure the utility functions I added in fill_extrusion_program.cpp are okay, or if there is a different place I should be doing that.

lbud:
> Would you like help implementing an API for light settings in the platform SDKs? Or should we handle that in separate PRs?

@1ec5 o 🦌 

I will need help, and I'd be inclined to say those APIs may be best as follow-up PRs (pre-releases), though I would defer to better judgment…

src/mbgl/style/style.hpp
```diff
@@ -113,6 +114,8 @@ class Style : public GlyphAtlasObserver,
     std::unique_ptr<SpriteAtlas> spriteAtlas;
     std::unique_ptr<LineAtlas> lineAtlas;
 
+    Light light;
```
1ec5: We’ll need to expose this field to SDK code via methods on `mbgl::Map`.

ivovandongen:
@lbud 
> would you mind setting a breakpoint to see what triangleIndex and totalVertices are when it fails? (When testing that single polygon on iOS they were 0 and 5, respectively, which is expected…)

`triangleIndex == 0` and `totalVertices == 0`, which means the assertion is `-4` which is not a correct value for a unsigned integer. Guarding that assertion (without knowing what it's for) like: 
`assert(triangleIndex + (5 * (totalVertices > 0 ? totalVertices - 1 : 0) + 1) <= std::numeric_limits<uint16_t>::max());` works.

The example in Android uses a camera animation, that might cause these unexpected values? Somehow a feature is passed without any vertices?

Let me know if I can debug something else for you.

lbud:
@jfirebaugh also wanted to flag that I feel really uncertain about my rewrite of `CascadingLightProperty` https://github.com/mapbox/mapbox-gl-native/pull/8431/commits/04eb22baaddf860a4a5c5d2efdeda0ace0dcb228 — about the general architectural changes, whether there is still some redundancy, and also how the TransitioningProperty setters no longer have access to a `CascadeParameters` argument so they use `Clock::now()` to come up with a TimePoint, which doesn't feel right… 🙏

lbud:
@ivovandongen odd 🤔 thanks for digging! https://github.com/mapbox/mapbox-gl-native/pull/8431/commits/543b5497722b98fe45611d20dedf4c3394a547d3 should sufficiently guard for those cases then 🙏

jfirebaugh:
@lbud Re `CascadingLightProperty`, I think we want to keep the access to `CascadeParameters`, and the call within `Style::cascade`. Even though there aren't multiple classes at play, `Style::cascade` is the right point to set up the transition from prior to new value when a property is updated via runtime styling.

What I suggest is to revert to your previous implementation, and then make `LightProperty` itself hold the value and transition:

```c++
template <class T>
class LightProperty {
public:
    using ValueType = PropertyValue<T>;
    using UnevaluatedType = TransitioningProperty<ValueType>;
    using EvaluatorType = PropertyEvaluator<T>;
    using EvaluatedType = T;

    ValueType value;
    TransitionOptions transition;

    template <class TransitioningProperty>
    TransitioningProperty cascade(const CascadeParameters& params,
                                  TransitioningProperty prior) const {
        return TransitioningProperty(value, std::move(prior),
                                     transition.reverseMerge(params.transition), params.now);
    }
};

template <class... Ps>
class LightProperties {
public:
    using Properties = TypeList<Ps...>;

    ...

    class Cascading : public Tuple<Properties> {
    public:
        using Tuple<Properties>::Tuple;
    };

    template <class P>
    auto get() const {
        return cascading.template get<P>().value;
    }

    template <class P>
    void set(const typename P::ValueType& value) {
        cascading.template get<P>().value = value;
    }

    ...
};
```

We probably should also come up with some term other than "cascading" here, but I'm not sure what. "Raw" maybe?

platform/darwin/src/MGLFillExtrusionStyleLayer.h
```diff
@@ -0,0 +1,281 @@
...
+// Edit platform/darwin/scripts/generate-style-code.js, then run `make style-code-darwin`.
+
+#import "MGLFoundation.h"
+#import "MGLStyleValue.h"
+#import "MGLVectorStyleLayer.h"
+
+NS_ASSUME_NONNULL_BEGIN
+
+/**
+ Controls the translation reference point.
```
lbud: @1ec5 it might be — this language is copied from `*-translate-anchor` in every other layer type. I'll ticket this in that repo as a follow-up.

src/mbgl/style/style.hpp
lbud: ✅

lbud:
Ping @jfirebaugh @kkaefer — I think this is done. (I do want to flag the additions of [`Map#getLight` / `Map#setLight`](https://github.com/mapbox/mapbox-gl-native/pull/8431/files#diff-fd1625168ba48a6fb963e63343c0ac59) because they were recent and may have slipped past review 👀, and because I don't know enough about pointer types to know if those argument/return types are right.)

@1ec5 @ivovandongen would love it if you could take a last pass over generated platform-specific code. Also, I have a good degree of confidence that this works on _at least the iOS version that I run on my phone_ but @ivovandongen I'd really appreciate it if you could check again to make sure this runs on Android. Like @1ec5 brought up, platform-specific `getLight`/`setLight` APIs can be added in follow-up PRs 🙏 😄

jfirebaugh:
`Map::setLight` should either take `Light` (by value), or if it takes it by `unique_ptr`, we should also change `Map::light` to be a `unique_ptr`.

* Accepting `Light` by value shows that the setter makes a copy, and subsequent mutations by the caller to its copy won't affect the copy used by the style.
* Accepting a `unique_ptr` implies that the style is taking ownership of the pointer, but leaves open the possibility that the caller can mutate the `Light` via a non-owning pointer. This is consistent with `Layer`s and `Source`s, and I think the right choice for that reason.

1ec5:
> This is consistent with Layers and Sources, and I think the right choice for that reason.

Light is more lightweight than Layer or Source, isn't it? I've always thought of Light as being something we'd represent in Objective-C as a value object or C struct that gets passed in by value. Pass-by-reference semantics means we'd have to be careful about, for instance, the same Light object being used in two different map view instances.

platform/ios/docs/guides/For Style Authors.md
```diff
@@ -206,6 +207,13 @@ In style JSON | In Objective-C | In Swift
 `fill-translate` | `MGLFillStyleLayer.fillTranslation` | `MGLFillStyleLayer.fillTranslation`
 `fill-translate-anchor` | `MGLFillStyleLayer.fillTranslationAnchor` | `MGLFillStyleLayer.fillTranslationAnchor`
 
+### FillExtrusion style layers
```
1ec5: This should be “Fill extrusion style layers”.

lbud:
🤔 @jfirebaugh @1ec5 
I took the `unique_ptr` approach in https://github.com/mapbox/mapbox-gl-native/pull/8431/commits/ede001af4946c3a4b8ba32983a521f97495e9144. Happy to revert if there is consensus on a different option.

include/mbgl/map/map.hpp
```diff
@@ -182,7 +182,7 @@ class Map : private util::noncopyable {
 
     // Light
     void setLight(std::unique_ptr<style::Light>);
-    style::Light* getLight();
+    std::unique_ptr<style::Light> getLight();
```
jfirebaugh: This should continue to return `style::Light*`; returning a `std::unique_ptr` gives up ownership, which we don't want to do.

src/mbgl/style/style.cpp
```diff
@@ -308,7 +309,7 @@ void Style::cascade(const TimePoint& timePoint, MapMode mode) {
         layer->baseImpl->cascade(parameters);
     }
 
-    transitioningLight = TransitioningLight(light, std::move(transitioningLight), parameters);
+    transitioningLight = TransitioningLight(light.get(), std::move(transitioningLight), parameters);
```
jfirebaugh: `light.get()` ⇢ `*light`, and revert the changes in `light_impl.hpp`

src/mbgl/programs/fill_extrusion_program.hpp
```diff
@@ -0,0 +1,129 @@
...
+#include <string>
+
+namespace mbgl {
+
+class SpriteAtlasElement;
+class UnwrappedTileID;
+class TransformState;
+
+namespace style {
+    template <class> class Faded;
```
kkaefer: Nitpick: No indentation in namespaces

src/mbgl/renderer/painter_fill_extrusion.cpp
```diff
@@ -0,0 +1,88 @@
...
+
+namespace mbgl {
+
+using namespace style;
+
+void Painter::renderFillExtrusion(PaintParameters& parameters,
+                                  FillExtrusionBucket& bucket,
+                                  const FillExtrusionLayer& layer,
+                                  const RenderTile& tile,
+                                  const Style& style) {
```
kkaefer: Rather than passing the `Style` object through the whole bucket dispatch system, can we retrieve the global light property as part of `Painter::render()` and store it as a property on the `Painter` object?

src/mbgl/util/mat3.cpp
```diff
@@ -94,4 +95,11 @@ void matrix::scale(mat3& out, const mat3& a, double x, double y) {
     out[8] = a[8];
 }
 
+void transformMat3f(vec3f& out, const vec3f& a, const mat3& m) {
```
kkaefer: Our matrix functions don't have the type of the argument in their name. C++ can automatically select the correct overload when passing a `vec3f`/`mat3`. However, given that we added the 4-component equivalents back in https://github.com/mapbox/mapbox-gl-native/commit/cef1b254445bdcfb20ada157ecc7f72c27aeacf9 with a similar name, this is probably okay.

src/mbgl/map/transform_state.cpp
```diff
@@ -64,6 +64,14 @@ void TransformState::getProjMatrix(mat4& projMatrix) const {
 
     matrix::translate(projMatrix, projMatrix, pixel_x() - size.width / 2.0f,
                       pixel_y() - size.height / 2.0f, 0);
+
+    const auto circumferenceOfEarth = 2 * M_PI * 6378137;
```
1ec5: [`EARTH_RADIUS_M`](https://github.com/mapbox/mapbox-gl-native/blob/6f708ac5458fe332e25ca398431928b7ff5ba404/include/mbgl/util/constants.hpp#L31) is available.

src/mbgl/gl/context.cpp
```diff
@@ -250,6 +250,7 @@ UniqueRenderbuffer Context::createRenderbuffer(const RenderbufferType type, cons
     bindRenderbuffer = renderbuffer;
     MBGL_CHECK_ERROR(
         glRenderbufferStorage(GL_RENDERBUFFER, static_cast<GLenum>(type), size.width, size.height));
+    bindRenderbuffer = 0;
```
lbud: @kkaefer after I removed the color renderbuffer in https://github.com/mapbox/mapbox-gl-native/pull/8431/commits/051481a78456792e4aed756a2674b28fdc1db583, the line above this (`glRenderbufferStorage`) aborts with a `GL_INVALID_OPERATION` error. I have no idea why it worked fine before I removed the color buffer (with two `glRenderbufferStorage` calls) and I don't really understand why this fixes it, but…it does… o_0 
And I figured since the binding is necessary here just to initialize the renderbuffer's storage, it wouldn't be incorrect to rebind the default renderbuffer / leaving the new renderbuffer bound wasn't necessary. ??

lbud:
@tobrun I'm hoping you can help me test the effects of https://github.com/mapbox/mapbox-gl-native/commit/34dc5bf , which checks for the presence of the [OES_depth24 extension](https://www.khronos.org/registry/OpenGL/extensions/OES/OES_depth24.txt) to use a higher resolution depth buffer when available. We do this because using too low a resolution causes artifacts that look like triangles on the sides of extrusions, e.g. 
image
or a more extreme case (on desktop)
image


This extension is supported in the few newer devices that I've tested and therefore eliminates these artifacts (it's also supported on my iPhone 7, although for some reason the 16 bit depth buffer was high enough resolution there anyway). But I'm trying to determine whether there may still be artifacts present when the extension was not supported — @ivovandongen suggested in Slack that you have an older Android device and also may be able to help do an AWS device farm run to see if there are still devices that suffer from these kinds of artifacts (possibly old, high resolution Android devices)?

tobrun:
I was able to reproduce the following render artefact on an ASUS Nexus 7 - Android 4.2.1:

ezgif com-video-to-gif 18

For reference this is the same test on a Nexus 6 - Android 5.0.1:

ezgif com-video-to-gif 16

lbud:
🙏 @tobrun thanks for testing that! I think I've solved it 🤞 in https://github.com/mapbox/mapbox-gl-native/pull/8431/commits/6b756af1ba17a4a7ec33a99bf0de18ce4ad6dd8f — would you mind testing that again?

src/mbgl/gl/context.cpp
kkaefer: 🤕 I think the underlying issue is that Framebuffers are "binding objects", meaning that you first bind a framebuffer, then bind other objects *to* the framebuffer (like textures and renderbuffer), similar to how vertex arrays work. What's happening is that once we bind/unbind a framebuffer, we don't restore the values to the renderbuffer/texture state values. This means that our implementation *thinks* that something is bound, while in reality it's not, because the overarching state object (framebuffer) has been unbound.

lbud:
Recording the solution to the depth buffer issue for posterity's sake:
In an ideal world we'd be able to use 24-bit depth buffers everywhere to solve the artifacts issue, but in reality we can't count on anything better than GL ES 2.0 without extensions. The approach I'm going with here is compatible with core ES2 (16-bit depth buffers — removes all references to 24-bit depth buffers, even where supported): for the projection matrix used for fill-extrusions, [set the near clip plane to 100](https://github.com/mapbox/mapbox-gl-native/pull/8431/commits/72278bfe94ce0d19b515bb92e9d0be88eb53aa89) (a sort of arbitrary number that seems to suit our needs well). This way we don't waste a ton of unnecessary depth precision on very near space where we aren't actually rendering anything, and we have enough precision for the rest of the scene.

ivovandongen:
@lbud I get a failed UI test on `FillExtrusionLayerTest#testSourceLayer`:

```
org.junit.ComparisonFailure: expected:<[]> but was:<[composite]>
at org.junit.Assert.assertEquals(Assert.java:115)
at org.junit.Assert.assertEquals(Assert.java:144)
at com.mapbox.mapboxsdk.testapp.style.FillExtrusionLayerTest.testSourceLayer(FillExtrusionLayerTest.java:90)
...
```

tobrun:
@lbud https://github.com/mapbox/mapbox-gl-native/pull/8431/commits/695da93c1de27a715015c16317942ca5a7debbaf resolves the failing test.

1ec5:
> platform-specific `getLight`/`setLight` APIs can be added in follow-up PRs 🙏 😄

Android: #8841
iOS and macOS: #8840

1ec5:
Looks like this PR [rolled back](https://github.com/mapbox/mapbox-gl-native/pull/8431/files#diff-a5247ef1b2f19cbe453449e7b27f8131) #6619 (nfarina/calloutview#107).

PR #8436, opened 2017-03-16
[build] Post nightly iOS build status to Slack

friedbunny:
Posts a notification to #gl-bots on Slack that reports the success/failure of the nightly iOS Bitrise workflow.

(Note that I already added this through Bitrise’s web interface as a test run for tonight — will remove that once this is merged.)

/cc @boundsj @kkaefer


friedbunny:
Removed the temporary Slack notification step from the `nightly-release` workflow.

PR #778, opened 2015-01-27
Remove program binary caching

ljbade:
The program binary caching that was added for https://github.com/mapbox/mapbox-gl-native/issues/588 keep creating headaches with out of date shaders and other issues where drivers return garbage values.

Considering most sane GL implementations (such as Android, iOS, and Linux with NVIDIA/AMD) already have internal shader caching I think that it is a waste of effort doing it ourselves. Our shaders are small too so I doubt we are saving much time with caching them.

If a GL implementation takes a long time to compile our shaders, it is their problem not ours.

Issue #588, opened 2014-11-11
Cache shaders


kkaefer:
We should use [GL_ARB_get_program_binary](https://www.opengl.org/registry/specs/ARB/get_program_binary.txt) and [GL_OES_get_program_binary](https://www.khronos.org/registry/gles/extensions/OES/OES_get_program_binary.txt) if available to cache the compiled shaders so a subsequent launch is a lot faster.

ljbade:
We decided to remove this due to many issues with it. https://github.com/mapbox/mapbox-gl-native/pull/778

PR #8393, opened 2017-03-16
Reconsider compiled shader programs

kkaefer:
On some Android devices, compiling shaders contributes significantly to slow renderer initialization time. We've [previously attempted to use compiled shaders](https://github.com/mapbox/mapbox-gl-native/issues/588), but removed it in https://github.com/mapbox/mapbox-gl-native/pull/778. It's time to reconsider.

Most Android devices seem to support [`GL_OES_get_program_binary`](https://www.khronos.org/registry/OpenGL/extensions/OES/OES_get_program_binary.txt), while macOS and iOS devices do not. However, we haven't identified slow shader compiles as a problem on these platforms.

### DIFF
--- a/src/mbgl/programs/programs.hpp
+++ b/src/mbgl/programs/programs.hpp
@@ -26,8 +26,8 @@ public:
           symbolIcon(context, programParameters),
           symbolIconSDF(context, programParameters),
           symbolGlyph(context, programParameters),
-          debug(context, ProgramParameters(programParameters.pixelRatio, false)),
-          collisionBox(context, ProgramParameters(programParameters.pixelRatio, false)) {
+          debug(context, ProgramParameters(programParameters.pixelRatio, false, programParameters.cacheDir)),
+          collisionBox(context, ProgramParameters(programParameters.pixelRatio, false, programParameters.cacheDir)) {
     }
 
     CircleProgram circle;
--- a/cmake/core-files.cmake
+++ b/cmake/core-files.cmake
@@ -63,6 +63,7 @@ set(MBGL_CORE_FILES
     src/mbgl/gl/draw_mode.hpp
     src/mbgl/gl/extension.cpp
     src/mbgl/gl/extension.hpp
+    src/mbgl/gl/features.hpp
     src/mbgl/gl/framebuffer.hpp
     src/mbgl/gl/gl.cpp
     src/mbgl/gl/index_buffer.hpp
@@ -70,6 +71,8 @@ set(MBGL_CORE_FILES
     src/mbgl/gl/object.hpp
     src/mbgl/gl/primitives.hpp
     src/mbgl/gl/program.hpp
+    src/mbgl/gl/program_binary.cpp
+    src/mbgl/gl/program_binary.hpp
     src/mbgl/gl/renderbuffer.hpp
     src/mbgl/gl/segment.cpp
     src/mbgl/gl/segment.hpp
@@ -130,6 +133,8 @@ set(MBGL_CORE_FILES
 
     # programs
     src/mbgl/programs/attributes.hpp
+    src/mbgl/programs/binary_program.cpp
+    src/mbgl/programs/binary_program.hpp
     src/mbgl/programs/circle_program.cpp
     src/mbgl/programs/circle_program.hpp
     src/mbgl/programs/collision_box_program.cpp
--- a/cmake/test-files.cmake
+++ b/cmake/test-files.cmake
@@ -38,6 +38,9 @@ set(MBGL_TEST_FILES
     test/math/minmax.test.cpp
     test/math/wrap.test.cpp
 
+    # programs
+    test/programs/binary_program.test.cpp
+
     # sprite
     test/sprite/sprite_atlas.test.cpp
     test/sprite/sprite_image.test.cpp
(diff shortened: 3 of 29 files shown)


kkaefer:
Instead of using anything asynchronous, this implementation uses synchronous file I/O for now, and writes the compiled program objects with their associated attribute and uniform bindings to disk, using PBF. It uses hashes of the shader content to avoid loading old compiled program objects (i.e. it'll recompile if the shader code changed).

With this, I was able to obtain an initialization speedup for subsequent app launches/initializations. On a Pixel C tablet, full compilation took ~350ms while loading precompiled programs only took ~40ms. Similarly, on a Moto G 2, full compilation took 320ms, and loading precompiled programs only took ~80ms.

src/mbgl/programs/programs.hpp
```diff
@@ -13,21 +13,23 @@ namespace mbgl {
...
-          lineSDF(context, programParameters),
-          linePattern(context, programParameters),
-          raster(context, programParameters),
-          symbolIcon(context, programParameters),
-          symbolIconSDF(context, programParameters),
-          symbolGlyph(context, programParameters),
-          debug(context, ProgramParameters(programParameters.pixelRatio, false)),
-          collisionBox(context, ProgramParameters(programParameters.pixelRatio, false)) {
+    Programs(gl::Context& context,
+             const ProgramParameters& programParameters,
```
kkaefer: yeah; but for the last two shaders we are constructing new ProgramParameter objects as well...

scripts/generate-shaders.js
```diff
@@ -82,18 +81,21 @@ ${fragmentPrelude}
...
+    const fragmentSource = applyPragmas(fs.readFileSync(path.join(inputPath, shaderName + '.fragment.glsl'), 'utf8'), {
                 define: [
                     "varying {precision} {type} {name};"
                 ],
                 initialize: [
                 ]
             });
-    }
+
+    const shasum = crypto.createHash('sha1');
```
kkaefer: I considered that as well, but it means that we'd have to expose some hashing function to C++ land via platform bindings, and incur the run-time cost of hashing. If you think it's safer, I'll go back and add it.

kkaefer: Another approach would be to use a unique version per build, e.g. the commit hash, along with the name of the shader.

kkaefer:
Some more measurements with the library compiled in Release mode:

**Pixel C (NVIDIA Tegra X1)**

- first launch: 349.984009ms
- relaunch: 19.046459ms

**Moto G 2 (Adreno 305)**

- first launch: 259.492188ms
- relaunch: 76.219635ms

**Nexus 4 (Adreno 320)**

- first launch: 707.584290ms
- relaunch: 90.004578ms

scripts/generate-shaders.js
kkaefer: We are using `std::hash` now to hash both shaders, convert the result to a hexadecimal string, and concatenate them. This results in 64 bits of entropy on 32 bit systems, and 128 bits of entropy on 64 bit systems. @jfirebaugh do you think `std::hash` is enough to protect us from identical strings, or should we use a different hash function?

src/mbgl/programs/programs.hpp
kkaefer: changed this

kkaefer:
Here's some more testing from Firebase:

Device | GPU | first | second
--- | --- | --- | ---
Samsung Galaxy S3 | Mali 400 MP4 | 194.1ms | 42.6ms
Samsung Galaxy S6 | Mali T760 MP8 | 313.6ms | 18.7ms
Samsung Galaxy Note S4 | Mali T760 MP6 | 319.6ms | 29.7ms
Moto G3 | Adreno 306 | 318.8ms | 22.1ms
Nexus 6 | Adreno 420 | 582.7ms | 44.0ms
Sony Xperia Z5 Compact | Adreno 430 | 360.7ms | 12.6ms
Asus Google Nexus 7 | Nvidia Tegra 3 | 💥 too many varyings |
Samsung Galaxy Tab 3 | Vivante GC1000 | 859.6ms | :warning: fails to cache
Samsung Galaxy J5 | Adreno 510 | 564.0ms | 60.8ms

kkaefer:
Added tests for serializing the shaders + bindings, and `#ifdef`ed binary shaders to not run on Apple platforms, since they don't support the required OpenGL extension anyway.

kkaefer:
> [...] binary shaders to not run on Apple platforms, since they don't support the required OpenGL extension anyway.

The good news here is that on iOS, this either happens automatically, or shader compilation is a lot faster to begin with. On an iPhone 6s, I'm observing an overall program compilation time that is ~20ms on the first run.

jfirebaugh:
> Asus Google Nexus 7 | Nvidia Tegra 3 | 💥 too many varyings |

Should this be filed as a separate issue?

> Samsung Galaxy Tab 3 | Vivante GC1000 | 859.6ms | :warning: fails to cache

Any idea why not?

scripts/generate-shaders.js
jfirebaugh: What's the behavior if there were to be a hash collision?

jfirebaugh:
@kkaefer Any answers on the last three questions I posted?

kkaefer:
> Should this be filed as a separate issue?

@incanus verified this on an actual device, and it works there. I'm assuming this is an issue with the SwiftShader implementation that Google uses on the virtualized Firebase devices.

> Any idea why not?

I went back to the log and it doesn't actually fail to cache. However, loading binary programs only seems marginally faster than a full compilation on this GPU (~700ms), but it doesn't actually fail, or is *slower* than not using precompiled programs.

I'm only counting two questions. What is the third one you are referring to?

PR #8426, opened 2017-03-15
[core] Prefer vector<pair<A, B>> to unordered_map<A, B> for small collections

jfirebaugh:
```
     VM SIZE                               FILE SIZE
 ++++++++++++++ GROWING                 ++++++++++++++
  +0.0%      +3 __TEXT,__cstring             +3  +0.0%

 -------------- SHRINKING               --------------
  -1.2% -29.5Ki __TEXT,__text           -29.5Ki  -1.2%
  -3.2% -5.32Ki [None]                  -1.75Ki  -1.0%
  -0.3% -1.02Ki __TEXT,__gcc_except_tab -1.02Ki  -0.3%
  -0.2%    -164 __TEXT,__unwind_info       -164  -0.2%
  -0.0%     -32 __TEXT,__const              -32  -0.0%

  -1.0% -36.0Ki TOTAL                   -32.4Ki  -0.9%
```

Refs #8035

PR #8439, opened 2017-03-16
[core] Prefer std::map to std::unordered_map for smaller binary size

jfirebaugh:
Replacement for #8426.

```
  +0.1%     +92 __TEXT,__unwind_info        +92  +0.1%

  -1.0% -25.7Ki __TEXT,__text           -25.7Ki  -1.0%
  -2.9% -4.80Ki [None]                  -1.12Ki  -0.7%
  -0.5% -1.55Ki __TEXT,__gcc_except_tab -1.55Ki  -0.5%
  -0.0%     -48 __TEXT,__const              -48  -0.0%

  -0.9% -32.0Ki TOTAL                   -28.3Ki  -0.8
```

Refs #8035

### DIFF
--- a/src/mbgl/annotation/annotation_manager.hpp
+++ b/src/mbgl/annotation/annotation_manager.hpp
@@ -9,7 +9,6 @@
 #include <string>
 #include <vector>
 #include <unordered_set>
-#include <unordered_map>
 
 namespace mbgl {
 
--- a/src/mbgl/gl/context.hpp
+++ b/src/mbgl/gl/context.hpp
@@ -21,7 +21,6 @@
 #include <vector>
 #include <array>
 #include <string>
-#include <unordered_map>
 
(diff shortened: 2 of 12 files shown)


tmpsantos:
Why is `std::map` smaller than `std::unordered_map`? Wouldn't this affect performance since we keep ordering when we don't need it?

jfirebaugh:
> Why is `std::map` smaller than `std::unordered_map`?

This is kind of tautological, but: it's just simpler. I'm not really familiar enough with the implementation details to give a comprehensive answer, but see [this microbenchmark](https://gist.github.com/jfirebaugh/02679a67784427dbac2caee507b6687a).

> Wouldn't this affect performance since we keep ordering when we don't need it?

Performance isn't a concern for these collections. They typically contain only one element, and very rarely more than a few.

PR #8443, opened 2017-03-17
Pass source object in MapObserver::onSourceChanged

### LINKED ISSUE #8397: mbgl::MapObserver::onSourceDidChange() should accept a source parameter
#8377 adds `mbgl::MapObserver::onSourceDidChange()` for informing SDK code that a source has changed (loaded). Per https://github.com/mapbox/mapbox-gl-native/pull/8377#discussion_r105757158, this method should accept a source parameter, which among other things would enable SDK code to be more selective about the source-related UI that it refreshes.

/cc @brunoabinader

brunoabinader:
Rename `MapObserver::onSourceDidChange` to `MapObserver::onSourceAttributionChanged`, passing both `sourceID` and `attribution` as strings.

Fixes #8397.

### DIFF
--- a/include/mbgl/map/map_observer.hpp
+++ b/include/mbgl/map/map_observer.hpp
@@ -1,7 +1,10 @@
 #pragma once
 
+#include <mbgl/style/source.hpp>
+
 #include <cstdint>
 #include <exception>
+#include <string>
 
 namespace mbgl {
 
@@ -33,7 +36,7 @@ public:
     virtual void onWillStartRenderingMap() {}
     virtual void onDidFinishRenderingMap(RenderMode) {}
     virtual void onDidFinishLoadingStyle() {}
-    virtual void onSourceDidChange() {}
+    virtual void onSourceChanged(style::Source&) {}
 };
 
 } // namespace mbgl
--- a/platform/android/src/native_map_view.cpp
+++ b/platform/android/src/native_map_view.cpp
@@ -241,7 +241,7 @@ void NativeMapView::onDidFinishLoadingStyle() {
     notifyMapChange(MapChange::MapChangeDidFinishLoadingStyle);
 }
 
-void NativeMapView::onSourceDidChange() {
+void NativeMapView::onSourceChanged(mbgl::style::Source&) {
     notifyMapChange(MapChange::MapChangeSourceDidChange);
 }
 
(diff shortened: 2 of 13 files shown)


jfirebaugh:
Agree with @1ec5, and we can generalize `SourceObserver::onSourceAttributionChanged` to `SourceObserver::onSourceChanged()` as well. Come to think of it, should we be exposing `SourceObserver`/`StyleObserver` directly, rather than partially exposing their APIs indirectly through `MapObserver`?

brunoabinader:
> we can generalize SourceObserver::onSourceAttributionChanged to SourceObserver::onSourceChanged() as well.

This might sound confusing, as `SourceObserver` contains other notifications that can be also considered _changes_. Would then this new `onSourceChanged` be related to not only attribution, but also data changes?

> Come to think of it, should we be exposing SourceObserver/StyleObserver directly, rather than partially exposing their APIs indirectly through MapObserver?

I'm okay with that - we could then be also exposing `StyleObserver` in case of `onDidFinishLoadingStyle`.

brunoabinader:
> I think it would be better for the method to accept an mbgl::style::Source directly

I have concerns regarding data consistency if we forward the `Source` object to its listeners via _pass-by-ref_ e.g. we're allowing listeners to make changes to the `Source` object while being notified of some other change. Thus why I decided to proceed with passing only the ID, which could then be used to obtain a pointer to the source via Map API.

If we proceed with passing the `Source` object, then I suggest we switch the paradigm to _pass-by-const-ref_.

brunoabinader:
> I'm okay with that - we could then be also exposing StyleObserver in case of onDidFinishLoadingStyle.

Currently `mbgl::styleObserver` inherits from `SourceObserver` itself - so we could just end up exposing it as a combo.

brunoabinader:
> Come to think of it, should we be exposing SourceObserver/StyleObserver directly, rather than partially exposing their APIs indirectly through MapObserver?

On a second thought, exposing `StyleObserver` involves exposing at least 4 other non-public headers, unless we reduce `StyleObserver` scope by removing `GlyphAtlasObserver` and `SpriteAtlasObserver` inheritances.

jfirebaugh:
> I have concerns regarding data consistency if we forward the `Source` object to its listeners via _pass-by-ref_ e.g. we're allowing listeners to make changes to the `Source` object while being notified of some other change.

This is a possibility even if not passed as a reference or passed by `const` reference, right? The receiver could be holding a reference already, or could look up the source by ID.

PR #8447, opened 2017-03-17
[android] - only dispatch events if not handled by MarkerView

### LINKED ISSUE #8444: MarkerView clicked
Using Mapbox Android SDK
Mapbox SDK version: 5.0.0-beta3

Steps to trigger behavior

when I click a markerview, the onMarkerClick and onMapClick method both are called。

Expected behavior

Only the onMarkerClick method is called；

i don't konw it is a bug or not.I just want distinguish the map or markerview is clicked，Is there any good way to do it?

tobrun:
Closes #8444

### DIFF
--- a/platform/android/MapboxGLAndroidSDK/src/main/java/com/mapbox/mapboxsdk/annotations/MarkerViewManager.java
+++ b/platform/android/MapboxGLAndroidSDK/src/main/java/com/mapbox/mapboxsdk/annotations/MarkerViewManager.java
@@ -8,6 +8,7 @@ import android.support.annotation.NonNull;
 import android.support.annotation.Nullable;
 import android.support.v4.util.LongSparseArray;
 import android.view.LayoutInflater;
+import android.view.MotionEvent;
 import android.view.View;
 import android.view.ViewGroup;
 import android.widget.ImageView;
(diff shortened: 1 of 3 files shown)


tobrun:
Retesting this and we actually need to be able to handle touches that are handled by the child views.. this makes starting a pan on a MarkerView impossible.

tobrun:
Will need to re implement the OnClick logic with OnTouch to be able to control the flag returned to MarkerViewContainer.

tobrun:
8f38a77 fixes the issue, want to review @Guardiola31337?

platform/android/MapboxGLAndroidSDK/src/main/java/com/mapbox/mapboxsdk/annotations/MarkerViewManager.java
```diff
@@ -503,18 +504,24 @@ public void invalidateViewMarkersInVisibleRegion() {
...
-                    clickHandled = onMarkerViewClickListener.onMarkerClick(marker, v, adapter);
-                  }
+              adaptedView.setOnTouchListener(new View.OnTouchListener() {
 
-                  if (!clickHandled) {
-                    ensureInfoWindowOffset(marker);
-                    select(marker, v, adapter);
+                @Override
+                public boolean onTouch(View v, MotionEvent event) {
+                  switch (event.getAction()) {
```
Guardiola31337: Why using `switch` instead `if`? There is only one case (`MotionEvent.ACTION_UP`).

tobrun: Comment fixed in #439f9f6

Issue #6584, opened 2016-10-05
Cannot introspect source or layer obtained from style


1ec5:
A source or style layer obtained from a map’s current style has its identifier set but has no other properties set. Therefore, there’s no way to obtain the source associated with a particular layer that comes with the Mapbox Streets style, nor is there a way to get the tile sets associated with that source.

I’m filing this issue in the course of fixing #6406, based on a close reading of the iOS/macOS and Android code base. To reproduce this issue on iOS or macOS, try the following code with the Mapbox Streets v9 style:

``` objc
NSURL *url = [NSURL URLWithString:@"mapbox://mapbox.mapbox-terrain-v2,mapbox.mapbox-streets-v7"];
NSAssert([[mapView.style layerWithIdentifier:@"country-label-md"].source.URL isEqual:url]);
```

To fix this issue, we’ll have to expose more members of `mbgl::style::Layer` and `mbgl::style::Source` that up to now have been stashed away in private `Impl` classes. For instance, we’ll need an `mbgl::style::VectorSource::getTileset()` that returns the existing `mbgl::style::TileSourceImpl::tileset`. I’m sure these members were hidden for good reason. To mitigate any questions about the expected lifetime of the returned object, it may be necessary to fix #6254 in the course of fixing this bug. (#6254 probably applies to the Android SDK as well.)

/cc @jfirebaugh @ivovandongen @incanus @frederoni @boundsj

jfirebaugh:
> I’m sure these members were hidden for good reason.

Not really, there's just never been a requirement to expose accessors for them before.

API design considerations:

* I'd prefer _not_ to expose a `Tileset` / TileJSON concept in public APIs. I think these concepts are on their way out, and sources will eventually just have inline properties for their URL templates, zoom range, etc. So those properties are best exposed directly on the source object, without an intermediate "Tileset" domain object.
* In the current implementation, the tileset properties might not be available during the initial load, while the TileJSON is being requested. This wrinkle will go away if we inline TileJSON properties, but for now the API needs to take it into account.

boundsj:
> I'd prefer not to expose a Tileset / TileJSON concept in public APIs.

Noting that this concept was recently added to the iOS SDK in https://github.com/mapbox/mapbox-gl-native/pull/6316 to support the initialization of sources with tile templates.

ivovandongen:
> Noting that this concept was recently added to the iOS SDK in #6316 to support the initialization of sources with tile templates.

[...]

PR #8472, opened 2017-03-20
[core] Expose Source::getZoomRange

brunoabinader:
Expose `Source::getZoomRange` as part of `Source` public API, also marking it `const`.

Part of #6584.

### DIFF
--- a/src/mbgl/style/sources/geojson_source_impl.cpp
+++ b/src/mbgl/style/sources/geojson_source_impl.cpp
@@ -153,7 +153,7 @@ void GeoJSONSource::Impl::loadDescription(FileSource& fileSource) {
     });
 }
 
-Range<uint8_t> GeoJSONSource::Impl::getZoomRange() {
+Range<uint8_t> GeoJSONSource::Impl::getZoomRange() const {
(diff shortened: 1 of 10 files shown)


src/mbgl/style/sources/geojson_source_impl.cpp
```diff
@@ -153,7 +153,7 @@ void GeoJSONSource::Impl::loadDescription(FileSource& fileSource) {
     });
 }
 
-Range<uint8_t> GeoJSONSource::Impl::getZoomRange() {
+Range<uint8_t> GeoJSONSource::Impl::getZoomRange() const {
     assert(loaded);
```
jfirebaugh: As hinted at more generally in the second point of [this comment](https://github.com/mapbox/mapbox-gl-native/issues/6584#issuecomment-251744922), this assert is likely to get triggered if we start using `getZoomRange` more widely. It may need to return `optional<Range<uint8_t>>`, so that callers are aware of the possibility of absence.

brunoabinader: 👉  #8486

PR #8475, opened 2017-03-20
[ios][macos] Document that transitions using setStyleURL are not supported

ericrwolfe:
Unless I'm mistaken, transitions via `-[MGLMapView setStyleURL:]` are not currently supported.

/cc @1ec5 @incanus

### DIFF
--- a/platform/darwin/src/MGLStyle.h
+++ b/platform/darwin/src/MGLStyle.h
@@ -194,8 +194,11 @@ MGL_EXPORT
 @property (nonatomic, strong) NS_SET_OF(__kindof MGLSource *) *sources;
 
 /**
- Values describing animated transitions to styling changes, either to the style URL 
- or to individual properties.
+ Values describing animated transitions to styling changes on individual 
+ paint properties. 
+ 
+ @note Transitions between styles changed using `-[MGLMapView setStyleURL:]`
+    are not currently supported.
  */
 @property (nonatomic) MGLTransition transition;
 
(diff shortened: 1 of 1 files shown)


platform/darwin/src/MGLStyle.h
```diff
@@ -194,8 +194,11 @@ MGL_EXPORT
 @property (nonatomic, strong) NS_SET_OF(__kindof MGLSource *) *sources;
 
 /**
- Values describing animated transitions to styling changes, either to the style URL 
- or to individual properties.
+ Values describing animated transitions to styling changes on individual 
+ paint properties. 
+ 
+ @note Transitions between styles changed using `-[MGLMapView setStyleURL:]`
+    are not currently supported.
```
1ec5: In general, we use `@note` for preconditions and postconditions – gotchas of that nature – since the `@pre` and `@post` Doxygen commands aren’t as well-supported. Often it’s sufficient to just have a separate paragraph with no command.

Also, I’d consider the ability to transition between two styles to be a feature in its own right, or at least the lack of this behavior is a bug. Perhaps we don’t need to call out this fact at all in the documentation, unless you think this is a compatibility issue.

ericrwolfe:
Sounds good, I'm fine with removing the note for now.

platform/darwin/src/MGLStyle.h
```diff
@@ -194,8 +194,8 @@ MGL_EXPORT
 @property (nonatomic, strong) NS_SET_OF(__kindof MGLSource *) *sources;
 
 /**
- Values describing animated transitions to styling changes, either to the style URL 
- or to individual properties.
+ Values describing animated transitions to changes on a style's individual
+ paint properties.
```
1ec5: This sounds convoluted but more consistent with the individual transition properties’ documentation comments:

> The transition affecting any changes to the paint attributes of the style’s layers.

> This property corresponds to the [`transition`](https://www.mapbox.com/mapbox-gl-js/style-spec/#root-transition) property in the style JSON file format.

PR #8492, opened 2017-03-22
[build] Autodetect Qt{4,5}

brunoabinader:
Detects the Qt version based on the `qmake` found in the user path - we'll no longer need to specify `WITH_QT_4=1` when e.g. building using Qt 4.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -166,7 +166,7 @@ matrix:
       dist: trusty
       language: cpp
       compiler: "qt4-gcc5-release"
-      env: BUILDTYPE=Release _CXX=g++-5 _CC=gcc-5 WITH_QT_4=1
+      env: BUILDTYPE=Release _CXX=g++-5 _CC=gcc-5
       addons: *qt4
       before_script:
         - mapbox_start_xvfb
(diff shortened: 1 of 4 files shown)


brunoabinader:
It seems there is no easy way of issuing a dry run of `find_package` e.g. attempting to find Qt4 package, even if not required, makes us link with those libraries. I'm gonna change the strategy to have `WITH_QT_4` CMake variable automatically set by our Makefile.

brunoabinader:
> While we're at it, can we make sure that running any Qt target in the Makefile displays a proper error message when qmake can't be found?

Done - when `qmake` is not found, the following output is produced:
```
$ make qt-app
Makefile:365: *** 'No qmake found: missing Qt development files'.  Stop.
```

tmpsantos:
I have Qt4 and Qt5 installed on my machine and I used to build both for testing. How can I do it now?

brunoabinader:
> I have Qt4 and Qt5 installed on my machine and I used to build both for testing. How can I do it now?

The build system will now detect the current qmake found in the user path. Switching from Qt4 and Qt5 is now a matter of using a different path for each e.g. in macOS you can `unlink` the qt and `link` the qt5 brew recipes.

PR #8493, opened 2017-03-22
[android] geojson conversion optimisation

### LINKED ISSUE #8483: Speed up GeoJSON conversion in GeoJsonSource
Currently, when setting data as a FeatureCollection on a [GeoJsonSource](https://github.com/mapbox/mapbox-gl-native/blob/master/platform/android/MapboxGLAndroidSDK/src/main/java/com/mapbox/mapboxsdk/style/sources/GeoJsonSource.java#L135), the data is serialised to a string, transferred over JNI to the c++ peer object where it is deserialised with rapidJSON. This comes at a high performance penalty.

Instead we should convert straight from the [MAS GeoJson](https://github.com/mapbox/mapbox-java/tree/master/mapbox/libjava-geojson) to [mapbox::geojson::geojson](https://github.com/mapbox/geojson-cpp).

ivovandongen:
Fixes #8483


platform/android/src/geojson/geometry.cpp
```diff
@@ -0,0 +1,52 @@
...
+
+#include <string>
+
+namespace mbgl {
+namespace android {
+namespace geojson {
+
+mapbox::geojson::geometry Geometry::convert(jni::JNIEnv &env, jni::Object<Geometry> jGeometry) {
+    auto type = Geometry::getType(env, jGeometry);
+    if (type == Point::Type()) {
```
brunoabinader: A switch case sounds more convenient here.

ivovandongen: It's a string type unfortunately. I wanted to use `IsInstanceOf`, but that would mean 6 calls into the JVM to determine the type in the worst case scenario.

platform/android/MapboxGLAndroidSDK/src/main/java/com/mapbox/mapboxsdk/style/sources/GeoJsonSource.java
```diff
@@ -133,8 +132,7 @@ public GeoJsonSource(String id, FeatureCollection features, GeoJsonOptions optio
    * @param features the GeoJSON {@link FeatureCollection}
    */
   public void setGeoJson(FeatureCollection features) {
```
jfirebaugh: Is it the convention on Android to always have a Java wrapper for a native method, even when it's possible to expose the native method directly (e.g. `public native void setGeoJson(FeatureCollection features);`)?

platform/android/src/geojson/feature_collection.cpp
```diff
@@ -0,0 +1,40 @@
...
+
+#include "feature.hpp"
+
+namespace mbgl {
+namespace android {
+namespace geojson {
+
+mbgl::FeatureCollection FeatureCollection::convert(jni::JNIEnv& env, jni::Object<FeatureCollection> jCollection) {
+    auto jFeatureList = FeatureCollection::getFeatures(env, jCollection);
+    auto jFeatures = java::util::List::toArray<Feature>(env, jFeatureList);
```
jfirebaugh: Have we ever benchmarked:

1. Using `List::toArray()` and then iterating over the result, extracting individual elements via `GetArrayRegion`?
2. Using `List::toArray()`, then converting the entire array to `std::vector` (or raw array) via `GetArrayRegion`, then iterating over that?
3. Using `List::toArray()`, iterating over reasonably-sized chunks copied to `std::vector` (or raw array) via `GetArrayRegion`?
4. Iterating over the `List` directly via JNI access to `List::iterator`?

[This article](https://www.ibm.com/developerworks/library/j-jni/) suggests that (3) is the best approach. We could package that up as a reusable template in jni.hpp that handles both chunking and deleting the temporary local refs.

platform/android/src/geojson/position.cpp
```diff
@@ -0,0 +1,25 @@
...
+
+namespace mbgl {
+namespace android {
+namespace geojson {
+
+mapbox::geojson::point Position::convert(jni::JNIEnv &env, jni::Object<Position> jPosition) {
+    static auto method = Position::javaClass.GetMethod<jni::Array<jdouble> ()>(env, "getCoordinates");
+    // Array with 0: longitude, 1: latitude (and optionally 2: altitude)
+    auto coordinates = jPosition.Call(env, method);
+    mapbox::geojson::point point (coordinates.Get(env, 0), coordinates.Get(env, 1));
```
jfirebaugh:
```
jdouble lngLat[2];
coordinates.GetRegion(env, 0, lngLat);
mapbox::geojson::point point(lngLat[0], lngLat[1]);
```

platform/android/MapboxGLAndroidSDK/src/main/java/com/mapbox/mapboxsdk/style/sources/GeoJsonSource.java
ivovandongen: Not necessarily, no. This is a left-over from a method which had some additional checking on the jvm side before passing it on to the native method.

platform/android/src/geojson/feature_collection.cpp
ivovandongen: No, haven't benchmarked this. Curious if a method call (though collection class methods) has more overhead than a memory access through array access.

Options 2 and 3 are not applicable here though, region access is not available for objects, only primitives as far as I know.

platform/android/src/geojson/position.cpp
ivovandongen: 👍

platform/android/src/geojson/feature_collection.cpp
jfirebaugh: Ah, right, dang.

ivovandongen:
Some cursory tests show conversion times are cut in half for the android conversion part.

PR #8517, opened 2017-03-24
[android] - harden offline region creation / update LatLngBounds

### LINKED ISSUE #4321: Add LatLngBounds.including
We currently don't allow dynamically updating the LatLngBounds. 
Including takes in `LatLng` and recalculates the current `LatLngBounds`

### LINKED ISSUE #8514: [Android] Invalid LatLng bounds in OfflineRegion causes JNI failure
**Platform: Android**
**Mapbox SDK version: 4.2.2**
**Device: Nexus 4, Android 5.1.1**

Hey there guys. This is a big one that I've just come across in 4.2.2. Unfortunately, all I have is a stack trace for you, and I can't test (or really reproduce this) on to 5.0.0 to try out because of #8497 .

From what I can tell, this was the flow:

 1. Fired up an offline region to download with a specific style (one of our own)
 2. Started bouncing around our app to test something map related (just display a map in this case)
 3. When it came time to show the map, the error: "failed to load style (timeout)" came up, and the callback for 'onMapReady' never hit.
 4. Without the mapboxMap from onMapReady, one of my features crashed.
 5. On loading up the app again and trying to figure out why I didn't have the mapboxMap instance, the app crashed a few seconds later (while trying to download an offline region) in the JNI stack.

The stack trace is here: https://gist.github.com/tikimcfee/c2da09cc0ffa7d7866a56bd28887a5d1

A little worried about this one, as we use offline regions pretty extensively and need to understand how this might affect our launch of our new feature set. Any help is greatly appreciated!

tobrun:
Closes #8514
Closes  #4321

### DIFF
--- a/platform/android/MapboxGLAndroidSDK/src/main/java/com/mapbox/mapboxsdk/offline/OfflineManager.java
+++ b/platform/android/MapboxGLAndroidSDK/src/main/java/com/mapbox/mapboxsdk/offline/OfflineManager.java
@@ -5,6 +5,8 @@ import android.os.Handler;
 import android.os.Looper;
 import android.support.annotation.NonNull;
 
+import com.mapbox.mapboxsdk.R;
+import com.mapbox.mapboxsdk.geometry.LatLngBounds;
 import com.mapbox.mapboxsdk.net.ConnectivityReceiver;
 import com.mapbox.mapboxsdk.storage.FileSource;
 
@@ -25,7 +27,6 @@ public class OfflineManager {
   static {
     System.loadLibrary("mapbox-gl");
   }
-
 
   // Native peer pointer
   private long nativePtr;
@@ -139,7 +140,8 @@ public class OfflineManager {
    *
    * @param callback the callback to be invoked
    */
-  public void listOfflineRegions(@NonNull final ListOfflineRegionsCallback callback) {
+  public void listOfflineRegions(@NonNull
+                                 final ListOfflineRegionsCallback callback) {
     listOfflineRegions(fileSource, new ListOfflineRegionsCallback() {
 
       @Override
@@ -180,10 +182,15 @@ public class OfflineManager {
    * @param metadata   the metadata in bytes
    * @param callback   the callback to be invoked
    */
-  public void createOfflineRegion(
-    @NonNull OfflineRegionDefinition definition,
-    @NonNull byte[] metadata,
-    @NonNull final CreateOfflineRegionCallback callback) {
+  public void createOfflineRegion(@NonNull OfflineRegionDefinition definition, @NonNull byte[] metadata,
+                                  final CreateOfflineRegionCallback callback) {
+    if (!isValidOfflineRegionDefinition(definition)) {
+      callback.onError(
+        String.format(context.getString(R.string.mapbox_offline_error_region_definition_invalid),
+          definition.getBounds())
+      );
+      return;
+    }
 
     ConnectivityReceiver.instance(context).activate();
     createOfflineRegion(fileSource, definition, metadata, new CreateOfflineRegionCallback() {
@@ -210,6 +217,16 @@ public class OfflineManager {
         });
       }
     });
+  }
+
+  /**
+   * Validates if the offline region definition bounds is valid for an offline region download.
+   *
+   * @param definition the offline region definition
+   * @return true if the region fits the world bounds.
+   */
+  private boolean isValidOfflineRegionDefinition(OfflineRegionDefinition definition) {
+    return LatLngBounds.world().contains(definition.getBounds());
   }
 
   /*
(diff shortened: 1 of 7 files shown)


platform/android/MapboxGLAndroidSDK/src/main/java/com/mapbox/mapboxsdk/offline/OfflineManager.java
```diff
@@ -114,6 +116,7 @@ public void run() {
     }).start();
   }
 
+  @UiThread
```
zugaldia: @tobrun Why are we requiring `@UiThread` for methods here? I'm thinking of issues if the developer wants to download regions from a background thread, say from a `Service`.

platform/android/MapboxGLAndroidSDKTestApp/src/test/java/com/mapbox/mapboxsdk/geometry/LatLngBoundsTest.java
```diff
@@ -34,47 +35,40 @@ public void beforeTest() {
       .build();
   }
 
-  @Test
-  public void testSanity() {
```
zugaldia: @tobrun Why is necessary to remove this test?

platform/android/MapboxGLAndroidSDKTestApp/src/test/java/com/mapbox/mapboxsdk/geometry/LatLngBoundsTest.java
```diff
@@ -194,6 +211,66 @@ public void testOuterUnion() {
   }
 
   @Test
+  public void northWest() {
```
zugaldia: @tobrun For all these new methods, could we add specific tests for cases where latitude and/or longitude are crossing their limits (i.e International Date Line)?

platform/android/MapboxGLAndroidSDK/src/main/java/com/mapbox/mapboxsdk/offline/OfflineManager.java
tobrun: I thought this was a requirement as the manager itself must be loaded from main thread afaik, let me fix this up.

platform/android/MapboxGLAndroidSDKTestApp/src/test/java/com/mapbox/mapboxsdk/geometry/LatLngBoundsTest.java
tobrun: A sanity test becomes obsolete when there other tests that indirectly test the same thing (this is a common test driven development practice).

platform/android/MapboxGLAndroidSDKTestApp/src/test/java/com/mapbox/mapboxsdk/geometry/LatLngBoundsTest.java
tobrun: Not sure if there is a reason for it as it's valid to construct a bounds that is not found inside of the world bounds (eg. no wrapping should be enforced by LatLngBounds). The current LatLngBounds logic now matches what [core](https://github.com/mapbox/mapbox-gl-native/blob/80915992ab4e970ca005009ca1175bae24cb448b/include/mbgl/util/geo.hpp#L85) exposes. FWIW this is also the reason why this PR only validates creation of an offline region instead of enforcing this with LatLngBounds.

platform/android/MapboxGLAndroidSDKTestApp/src/test/java/com/mapbox/mapboxsdk/geometry/LatLngBoundsTest.java
zugaldia: I always keep those, well, to make sure I keep my sanity ;)

platform/android/MapboxGLAndroidSDKTestApp/src/test/java/com/mapbox/mapboxsdk/geometry/LatLngBoundsTest.java
zugaldia: I'm worried about developers using this class outside offline cases. If wrapping is not tested at the core level either, then it's fine to leave it there but I'd make a note in Javadoc.

platform/android/MapboxGLAndroidSDK/src/main/java/com/mapbox/mapboxsdk/offline/OfflineManager.java
zugaldia: @tobrun Is it currently possible to get the manager instance from the main thread but then call the other methods in a separate thread?

platform/android/MapboxGLAndroidSDKTestApp/src/test/java/com/mapbox/mapboxsdk/geometry/LatLngBoundsTest.java
tobrun: I'm all down for keeping you sane! Will keep them.

platform/android/MapboxGLAndroidSDKTestApp/src/test/java/com/mapbox/mapboxsdk/geometry/LatLngBoundsTest.java
tobrun: added this in class javadoc

PR #8520, opened 2017-03-24
[Qt] Refactor QMapboxGL MapObserver signals

### LINKED ISSUE #8419: [Qt] Refactor QMapboxGL API to reflect MapObserver semantics
After replacing `mbgl::MapChange` enum with `mbgl::MapObserver` in #8377, we might want to update/refactor `QMapboxGL` public API to reflect this.

/cc @tmpsantos

brunoabinader:
Fixes #8419.

### DIFF
--- a/platform/qt/include/qmapboxgl.hpp
+++ b/platform/qt/include/qmapboxgl.hpp
@@ -94,31 +94,24 @@ class Q_DECL_EXPORT QMapboxGL : public QObject
     Q_PROPERTY(QMargins margins READ margins WRITE setMargins)
 
 public:
-    enum MapChange {
-        MapChangeRegionWillChange = 0,
-        MapChangeRegionWillChangeAnimated,
-        MapChangeRegionIsChanging,
-        MapChangeRegionDidChange,
-        MapChangeRegionDidChangeAnimated,
-        MapChangeWillStartLoadingMap,
-        MapChangeDidFinishLoadingMap,
-        MapChangeDidFailLoadingMap,
-        MapChangeWillStartRenderingFrame,
-        MapChangeDidFinishRenderingFrame,
-        MapChangeDidFinishRenderingFrameFullyRendered,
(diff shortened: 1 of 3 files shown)


platform/qt/include/qmapboxgl.hpp
```diff
@@ -232,15 +225,29 @@ public slots:
 
 signals:
     void needsRendering();
-    void mapChanged(QMapboxGL::MapChange);
     void copyrightsChanged(const QString &copyrightsHtml);
 
+signals:
+    void cameraWillChange(QMapboxGL::CameraChangeMode);
```
tmpsantos: I'm not sure if I like this. For the high level API we don't need these signals anyway...

brunoabinader: Rationale is that instead of having a switch case to catch the specific notifications from `cameraWillChange`, you connect to the specific signals you need.

Also, the current implementation does not provide the extra arguments passed e.g. via [onDidFailLoadingMap](https://github.com/mapbox/mapbox-gl-native/blob/master/include/mbgl/map/map_observer.hpp#L33) or [onSourceChanged](https://github.com/mapbox/mapbox-gl-native/blob/master/include/mbgl/map/map_observer.hpp#L39).

brunoabinader: @tmpsantos based on the comments above, can we revisit this? I'm also :+1: in finding other ways for passing the arguments for the signals mentioned.

tmpsantos: Alright, but we need to add documentation for all these signals :smile:

tmpsantos: I think this PR was before we add docs for everything...

brunoabinader:
Superseded by https://github.com/mapbox/mapbox-gl-native/pull/11039, which also implements `mbgl::MapObserver`.

PR #8535, opened 2017-03-27
[core] Do not throw on ctor and dtor

tmpsantos:
Fix build on GCC6. C+11 defaults ctors to nothrow.

### DIFF
--- a/platform/default/sqlite3.cpp
+++ b/platform/default/sqlite3.cpp
@@ -30,7 +30,7 @@ public:
 
         const int error = sqlite3_close(db);
         if (error != SQLITE_OK) {
-            throw Exception { error, sqlite3_errmsg(db) };
+            mbgl::Log::Error(mbgl::Event::Database, "%s (Code %i)", sqlite3_errmsg(db), error);
         }
     }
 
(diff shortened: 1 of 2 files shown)


tmpsantos:
@jfirebaugh not sure about what is the best way to handle this on sqlite3 bindings. Should we have something like `::isValid()`?

jfirebaugh:
Constructors should still throw. It's just destructors that need either `nothrow(false)` or some other strategy.

tmpsantos:
@jfirebaugh fixed, but in the case of the `~ThreadLocal()` the issue is we don't catch the exception.

```
In file included from ../../../src/mbgl/map/backend_scope.cpp:3:0:
../../../src/mbgl/util/thread_local.hpp: In instantiation of ‘mbgl::util::ThreadLocal<T>::~ThreadLocal() [with T = mbgl::BackendScope]’:
../../../src/mbgl/map/backend_scope.cpp:9:40:   required from here
../../../src/mbgl/util/thread_local.hpp:32:75: error: throw will always call terminate() [-Werror=terminate]
             throw std::runtime_error("Failed to delete local storage key.");
                                                                           ^
../../../src/mbgl/util/thread_local.hpp:32:75: note: in C++11 destructors default to noexcept
```

Issue #8528, opened 2017-03-26
Android builds failing on Bitrise on release-ios-v3.5.0-android-v5.0.0 branch


1ec5:
Android builds on Bitrise on the release-ios-v3.5.0-android-v5.0.0 branch started failing over the weekend with an error that seems to be related to the generated code for runtime styling:

```
[...]

PR #8536, opened 2017-03-27
Use prebuilt Docker image for Android/Bitrise

kkaefer:
This PR intents to lower the build time we spend on the Android/Bitrise app. We are past the 25 minute mark, but most of that time is spent downloading Gradle dependencies and installing other dependencies.

Instead, we can use a pre-built Docker image that contains most of these things. While we're at it, we can also add `ccache` to increase compilation times. Our builds are already switched to using `mbgl/android:latest` instead of `bitriseio/android-ndk:latest` and build time has decreased from 26 minutes to just 10 minutes, including Firebase tests.

Bitrise updates their base image every Saturday, so we'll have to add a scheduled build that updates our image to avoid pulling the old Bitrise image. This conenviently lets us refresh our primed `ccache`.

### DIFF
--- a/docker/bitrise/android/Dockerfile
+++ b/docker/bitrise/android/Dockerfile
@@ -1,0 +1,21 @@
+FROM bitriseio/android-ndk:latest
+
+# Install Google Cloud SDK for Firebase
+RUN set -eu && \
+    (echo "deb http://packages.cloud.google.com/apt cloud-sdk-$(lsb_release -c -s) main" | sudo tee /etc/apt/sources.list.d/google-cloud-sdk.list) && \
+    (curl https://packages.cloud.google.com/apt/doc/apt-key.gpg | sudo apt-key add -) && \
+    sudo apt-get update && \
+    sudo apt-get install -y google-cloud-sdk python-dev python-setuptools ccache && \
+    sudo apt-get clean && \
+    sudo easy_install -U pip && \
+    pip install --no-cache-dir awscli && \
+    mkdir -p "${ANDROID_HOME}/licenses" && \
+    (echo "8933bad161af4178b1185d1a37fbf41ea5269c55" > "${ANDROID_HOME}/licenses/android-sdk-license")
+
+RUN set -eu && \
+    git clone --depth=1 "https://github.com/mapbox/mapbox-gl-native.git" . && \
+    ccache -z && \
+    BUILDTYPE=Debug make android-test-lib-arm-v7 && \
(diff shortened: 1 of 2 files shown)


docker/bitrise/android/Dockerfile
```diff
@@ -0,0 +1,21 @@
...
+    sudo apt-get update && \
+    sudo apt-get install -y google-cloud-sdk python-dev python-setuptools ccache && \
+    sudo apt-get clean && \
+    sudo easy_install -U pip && \
+    pip install --no-cache-dir awscli && \
+    mkdir -p "${ANDROID_HOME}/licenses" && \
+    (echo "8933bad161af4178b1185d1a37fbf41ea5269c55" > "${ANDROID_HOME}/licenses/android-sdk-license")
+
+RUN set -eu && \
+    git clone --branch=bitrise-docker --depth=1 "https://github.com/kkaefer/mapbox-gl-native.git" . && \
```
kkaefer: This line needs to be changed to 

```
git clone --depth=1 "https://github.com/mapbox/mapbox-gl-native.git"
```

jfirebaugh:
> Our builds are already switched to using `mbgl/android:latest` instead of `bitriseio/android-ndk:latest`

Is this the cause of #8528?

kkaefer:
No, because I didn't change the image until this afternoon.

jfirebaugh:
What was the image setting prior to that? Have you made changes to it recently?

jfirebaugh:
Found the issue: https://github.com/mapbox/mapbox-gl-native/issues/8528#issuecomment-289620365. This suggests that we do _not_ want to use `:latest`, and instead should control when build environment updates happen.

kkaefer:
> This suggests that we do _not_ want to use `:latest`, and instead should control when build environment updates happen.

Unfortunately this isn't how Bitrise works :( They rebuild their image every Saturday, and *only* cache the latest images on their VMs. Using an older image means that every build machine has to redownload the 25 GB image. In addition, their images are [needlessly large](https://github.com/bitrise-docker/android/issues/71). Instead of using anything based on their image, we could also attempt to create a minimal build environment that we maintain ourselves and only includes the required tools. However, this would likely also be several gigabytes in size, which we'd have to pull from Docker Hub on every build.

kkaefer:
Merged, since #8528 happened without this change too (since the new Docker image we're using is merely augmenting the Bitrise default image.

PR #8540, opened 2017-03-27
Small edit to improve code resiliency

anandthakker:
h/t @ChrisLoer catching this place where we were relying on member
initialization order.  While it _happened_ to be okay here, it's
unnecessarily fragile.


ChrisLoer:
This is good... but should we go all the way and define/use a constructor for `VirtualSegment`? Otherwise changing the order of member declaration could still change behavior unexpectedly. On the other hand, if that broke it would probably be in a pretty clear "this is totally broken" kind of way.

anandthakker:
My inclination is to leave it as-is, but that's mostly because it seems aesthetically more lightweight this way (and because this is an internal type used during processing, so exposure's very limited anyway)

Issue #8420, opened 2017-03-15
signature for callOutViewFor:annotation should have MGLCalloutView return type


idrougge:
**Platform:** Swift
**Mapbox SDK version:** 3.4.1

`func mapView(_ mapView: MGLMapView, calloutViewFor annotation: MGLAnnotation) -> UIView?` as a signature is misleading since the function expects the return `UIView` to conform to `MGLCalloutView` and will cause a crash because of sending incorrect selectors. The return type should be changed to `MGLCalloutView` in order to reflect this.

1ec5:
Retitling, because macOS is unaffected: `-[MGLMapViewDelegate mapView:calloutViewControllerForAnnotation:]` returns an NSViewController, and the return value is always treated as a plain NSViewController.

boundsj:
This was fixed by https://github.com/mapbox/mapbox-gl-native/pull/8541 that landed in the `release-ios-v3.5.0-android-v5.0.0` branch and will be in the [ios v3.5.1 patch release](https://github.com/mapbox/mapbox-gl-native/milestone/57)

boundsj:
Reopening.

Unfortunately, I missed that this change causes a warning in `MGLMapView`:

```Objective-C
calloutView = [self.delegate mapView:self calloutViewForAnnotation:annotation];
```

triggers:

```
/mapbox-gl-native/platform/ios/src/MGLMapView.mm:3764:27: Incompatible pointer types assigning to 'UIView<MGLCalloutView> *' from 'id<MGLCalloutView> _Nullable'
```

The naming of `MGLCalloutView` protocol and its documentation make the correct choice for what base type to return fairly obvious. However, I think we should guard against using objects that are not a kind of `UIView` and, if they are views, cast them as such to silence the warning.

cc @1ec5

PR #8541, opened 2017-03-27
[ios] Modify return type so callout bridges with protocol in Swift

boundsj:
Fixes https://github.com/mapbox/mapbox-gl-native/issues/8420

`- (nullable UIView <MGLCalloutView> *)mapView:(MGLMapView *)mapView calloutViewForAnnotation:(id <MGLAnnotation>)annotation;` bridges with a `UIView?` return type in Swift. This changes replaces the `UIView` type in the method signature so that, when bridged to Swift, the returned object is defined as `MGLCalloutView?`.

### DIFF
--- a/platform/ios/CHANGELOG.md
+++ b/platform/ios/CHANGELOG.md
@@ -2,6 +2,12 @@
 
 Mapbox welcomes participation and contributions from everyone. Please read [CONTRIBUTING.md](../../CONTRIBUTING.md) to get started.
 
+## 3.5.1
+
+### Other changes
+
+* Fixed an issue that caused the return type of a map view delegate method to bridge incorrectly to applications written in Swift. ([#8541](https://github.com/mapbox/mapbox-gl-native/pull/8541))
+
 ## 3.5.0
 
 ### Packaging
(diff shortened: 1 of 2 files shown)


boundsj:
Blocked by https://github.com/mapbox/mapbox-gl-native/issues/8528

platform/ios/src/MGLMapViewDelegate.h
```diff
@@ -456,7 +456,7 @@ NS_ASSUME_NONNULL_BEGIN
  @return A view conforming to the `MGLCalloutView` protocol, or `nil` to use the
     default callout view.
  */
-- (nullable UIView <MGLCalloutView> *)mapView:(MGLMapView *)mapView calloutViewForAnnotation:(id <MGLAnnotation>)annotation;
+- (nullable id <MGLCalloutView>)mapView:(MGLMapView *)mapView calloutViewForAnnotation:(id <MGLAnnotation>)annotation;
```
1ec5: This is worth a mention in the iOS changelog.

platform/ios/CHANGELOG.md
```diff
@@ -2,6 +2,12 @@
 
 Mapbox welcomes participation and contributions from everyone. Please read [CONTRIBUTING.md](../../CONTRIBUTING.md) to get started.
 
+## 3.5.1
+
+### Other changes
```
1ec5: Nit: normally we don’t break out patch releases by topic, since they’re so small. This will be pretty easy to address right before release when we copy-edit release notes anyhow.

PR #8550, opened 2017-03-28
GLFW window resizing fails

kkaefer:
Resizing the window, or moving it to a different screen fails after https://github.com/mapbox/mapbox-gl-native/pull/8216

### DIFF
--- a/platform/glfw/glfw_view.cpp
+++ b/platform/glfw/glfw_view.cpp
@@ -377,6 +377,8 @@ void GLFWView::onFramebufferResize(GLFWwindow *window, int width, int height) {
     auto *view = reinterpret_cast<GLFWView *>(glfwGetWindowUserPointer(window));
     view->fbWidth = width;
     view->fbHeight = height;
+
+    mbgl::BackendScope scope { *view, mbgl::BackendScope::ScopeType::Implicit };
     view->bind();
 
     // This is only triggered when the framebuffer is resized, but not the window. It can


platform/glfw/glfw_view.cpp
```diff
@@ -381,7 +381,6 @@ void GLFWView::onFramebufferResize(GLFWwindow *window, int width, int height) {
     GLFWView *view = reinterpret_cast<GLFWView *>(glfwGetWindowUserPointer(window));
     view->fbWidth = width;
     view->fbHeight = height;
-    view->bind();
```
tmpsantos: This might fix for OSX, but breaks Linux.

tmpsantos: screenshot from 2017-03-28 18-40-07

This is how it looks when resizing on Linux with this patch. I'm not sure why it behaves differently on both platforms, but if you don't feel like debugging glfw a `#ifdef` would do it.

brunoabinader:
Attempting to resize the GLFW window with the proposed patch throws the following:

> mbgl-glfw: ../../../src/mbgl/map/backend.cpp:37: void mbgl::Backend::assumeViewportSize(const mbgl::Size&): Assertion `gl::value::Viewport::Get() == getContext().viewport.getCurrentValue()' failed.

brunoabinader:
@kkaefer it turns out adding an implicit backend scope fixes the issue on Linux - I'm almost sure it is safe to assume we have a valid backend by the time the callback is called - if you agree, then we can squash that into the original patch.

brunoabinader:
@tmpsantos can you please re-review?

PR #8552, opened 2017-03-28
Xcode: lexical or preprocessor issue

kkaefer:
When rebuilding an Xcode project while Xcode is open, compilation sometimes fails because Xcode doesn't properly read the `.xcconfig` file generated during the CMake run. This leads to errors like this:

screen shot 2017-03-28 at 1 31 20 pm

The temporary solution is to close and reopen Xcode.


kkaefer:
We previously used incremental `file(WRITE ...)` commands that gradually recreated the file on every CMake invocation. This sometimes lead to Xcode parsing a partially written file, which in turn breaks building dependend targets. Instead, we're now using a templated configuration file, which ensure that the file is created in one go and hopefully reduces the race condition between CMake and Xcode's automatic project updating.

PR #8556, opened 2017-03-29
[android] - Transifex intregration and Dutch translations.

tobrun:
This PR adds [mapbox-gl-native Transifex project](https://www.transifex.com/mapbox/mapbox-gl-native) integration for the Android Mapbox SDK. 
To test integration, Dutch translations were added.

@1ec5 I moved the lang_map of the previous integration to the specific tasks (Android follows
the ISO 639-1 standard). I feel this was the best solution as applying a lang_map to a specific task will not override the project lang_map but will extend it. 

cc @mapbox/android

### DIFF
--- a/.tx/config
+++ b/.tx/config
@@ -1,25 +1,35 @@
 [main]
 host = https://www.transifex.com
-lang_map = pt_BR: pt-BR, pt_PT: pt-PT, zh_CN: zh-Hans, zh_TW: zh-Hant
 type = STRINGS
 minimum_perc = 80
 
 [mapbox-gl-native.foundationstrings-darwin]
+lang_map = pt_BR: pt-BR, pt_PT: pt-PT, zh_CN: zh-Hans, zh_TW: zh-Hant
 source_file = platform/darwin/resources/Base.lproj/Foundation.strings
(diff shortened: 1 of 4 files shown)


zugaldia:
@tobrun I believe this is worth including in the CHANGELOG.

tobrun:
> A run of tx pull -a shows that this PR works as advertised. I did notice that tx changed several existing .strings files back to UTF-16 encoding (which git treats as binary), so we’ll have to keep an eye out for that and restore UTF-8 encoding to those files.

I saw this as well, wasn't sure if this was expected behavior. Thank you for adding this. 

> Incidentally, you may want to run tx pull -a to get the Catalan translation, which has been updated in Transifex to cover the new Android strings! :tada:

Wow, that was fast! or should I say: `això va ser ràpid`.

> If you haven’t already, don’t forget to configure Transifex to fetch the raw version of strings.xml daily.

I haven't, will look into it.

Guardiola31337:
@tobrun 😂 
> Wow, that was fast! or should I say: `això va ser ràpid`.

^ `què ràpid!` 👍

PR #8557, opened 2017-03-29
[android] Polygon holes

Guardiola31337:
New attempt of fixing #1729 

- [x] Support for 1 hole
  - [x] Support in `Polygon`
  - [x] Support in `PolygonOptions`
- [x] Support for multiple holes

@ivovandongen Could you review/add more comments on C++ part?
At the moment, still remains the following:

- [ ] Tweak C++ code
  - [ ] Naming
  - [ ] Use of aliases?
  - [ ] Separate the implementation in an implementation file
  - [ ] Separate the concerns in multiple methods

Anything else?

BTW, does the overall PR make sense? Would you tackle it using a different approach?


1ec5:
For reference, #7397 attempted to support multiple holes at the Java level, and mbgl does support multiple holes (as seen on iOS and macOS); it’s the JNI part that apparently wasn’t working.

PR #8562, opened 2017-03-29
Tighten size check

brunoabinader:
Some platforms provide their own size type defaulting to -1 in both width and height e.g. Qt's [QSize](http://doc-snapshots.qt.io/qt5-5.9/qsize.html). We should special case this and make this part of the assertion check.

### DIFF
--- a/platform/qt/src/qmapboxgl.cpp
+++ b/platform/qt/src/qmapboxgl.cpp
@@ -8,6 +8,7 @@
 #include <mbgl/map/camera.hpp>
 #include <mbgl/map/map.hpp>
 #include <mbgl/map/backend_scope.hpp>
+#include <mbgl/math/minmax.hpp>
 #include <mbgl/style/conversion.hpp>
 #include <mbgl/style/conversion/layer.hpp>
 #include <mbgl/style/conversion/source.hpp>
@@ -78,6 +79,13 @@ auto fromQStringList(const QStringList &list)
     }
     return strings;
 }
+
+mbgl::Size sanitizedSize(const QSize& size) {
+    return mbgl::Size {
+        mbgl::util::max(0u, static_cast<uint32_t>(size.width())),
+        mbgl::util::max(0u, static_cast<uint32_t>(size.height())),
+    };
+};
 
 std::unique_ptr<const mbgl::SpriteImage> toSpriteImage(const QImage &sprite) {
     const QImage swapped = sprite
@@ -433,6 +441,8 @@ void QMapboxGLSettings::setApiBaseUrl(const QString& url)
 QMapboxGL::QMapboxGL(QObject *parent, const QMapboxGLSettings &settings, const QSize& size, qreal pixelRatio)
     : QObject(parent)
 {
+    assert(!size.isEmpty());
+
     // Multiple QMapboxGL running on the same thread
     // will share the same mbgl::util::RunLoop
     if (!loop.hasLocalData()) {
@@ -1079,8 +1089,7 @@ void QMapboxGL::resize(const QSize& size, const QSize& framebufferSize)
     d_ptr->size = size;
     d_ptr->fbSize = framebufferSize;
 
-    d_ptr->mapObj->setSize(
-        { static_cast<uint32_t>(size.width()), static_cast<uint32_t>(size.height()) });
+    d_ptr->mapObj->setSize(sanitizedSize(size));
 }
 
 /*!
@@ -1547,14 +1556,15 @@ QMapboxGLPrivate::QMapboxGLPrivate(QMapboxGL *q, const QMapboxGLSettings &settin
         settings.assetPath().toStdString(),
         settings.cacheDatabaseMaximumSize()))
     , threadPool(mbgl::sharedThreadPool())
-    , mapObj(std::make_unique<mbgl::Map>(
-        *this, mbgl::Size{ static_cast<uint32_t>(size.width()), static_cast<uint32_t>(size.height()) },
-        pixelRatio, *fileSourceObj, *threadPool,
-        mbgl::MapMode::Continuous,
-        static_cast<mbgl::GLContextMode>(settings.contextMode()),
-        static_cast<mbgl::ConstrainMode>(settings.constrainMode()),
-        static_cast<mbgl::ViewportMode>(settings.viewportMode())))
 {
+    mapObj = std::make_unique<mbgl::Map>(
+            *this, sanitizedSize(size),
+            pixelRatio, *fileSourceObj, *threadPool,
+            mbgl::MapMode::Continuous,
+            static_cast<mbgl::GLContextMode>(settings.contextMode()),
+            static_cast<mbgl::ConstrainMode>(settings.constrainMode()),
+            static_cast<mbgl::ViewportMode>(settings.viewportMode()));
+
     qRegisterMetaType<QMapboxGL::MapChange>("QMapboxGL::MapChange");
 
     fileSourceObj->setAccessToken(settings.accessToken().toStdString());
@@ -1570,7 +1580,7 @@ QMapboxGLPrivate::~QMapboxGLPrivate()
 }
 
 mbgl::Size QMapboxGLPrivate::framebufferSize() const {
-    return { static_cast<uint32_t>(fbSize.width()), static_cast<uint32_t>(fbSize.height()) };
+    return sanitizedSize(fbSize);
 }
 
 void QMapboxGLPrivate::updateAssumedState() {
--- a/src/mbgl/map/transform.cpp
+++ b/src/mbgl/map/transform.cpp
@@ -45,9 +45,13 @@ Transform::Transform(MapObserver& observer_,
 
 #pragma mark - Map View
 
-bool Transform::resize(const Size size) {
+void Transform::resize(const Size size) {
+    if (size.isEmpty()) {
+        throw std::runtime_error("failed to resize: size is empty");
+    }
+
     if (state.size == size) {
-        return false;
+        return;
     }
 
     observer.onCameraWillChange(MapObserver::CameraChangeMode::Immediate);
@@ -56,8 +60,6 @@ bool Transform::resize(const Size size) {
     state.constrain(state.scale, state.x, state.y);
 
     observer.onCameraDidChange(MapObserver::CameraChangeMode::Immediate);
-
-    return true;
 }
 
 #pragma mark - Camera
--- a/src/mbgl/map/transform_state.hpp
+++ b/src/mbgl/map/transform_state.hpp
@@ -74,6 +74,10 @@ public:
     double zoomScale(double zoom) const;
     double scaleZoom(double scale) const;
 
+    bool valid() const {
+        return !size.isEmpty() && (scale >= min_scale && scale <= max_scale);
+    }
+
 private:
     bool rotatedNorth() const;
     void constrain(double& scale, double& x, double& y) const;
--- a/include/mbgl/util/image.hpp
+++ b/include/mbgl/util/image.hpp
@@ -58,7 +58,7 @@ public:
     }
 
     bool valid() const {
-        return size && data.get() != nullptr;
+        return !size.isEmpty() && data.get() != nullptr;
     }
 
     template <typename T = Image>
--- a/include/mbgl/util/size.hpp
+++ b/include/mbgl/util/size.hpp
@@ -7,8 +7,7 @@ namespace mbgl {
 
 class Size {
 public:
-    constexpr Size() : width(0), height(0) {
-    }
+    constexpr Size() = default;
 
     constexpr Size(const uint32_t width_, const uint32_t height_) : width(width_), height(height_) {
     }
@@ -17,12 +16,12 @@ public:
         return width * height;
     }
 
-    constexpr explicit operator bool() const {
-        return width > 0 && height > 0;
+    constexpr bool isEmpty() const {
+        return width == 0 || height == 0;
     }
 
-    uint32_t width;
-    uint32_t height;
+    uint32_t width = 0;
+    uint32_t height = 0;
 };
 
 constexpr inline bool operator==(const Size& a, const Size& b) {
--- a/platform/android/build.gradle
+++ b/platform/android/build.gradle
@@ -3,7 +3,7 @@ buildscript {
         jcenter()
     }
     dependencies {
-        classpath 'com.android.tools.build:gradle:2.3.0'
+        classpath 'com.android.tools.build:gradle:2.3.1'
         classpath 'com.amazonaws:aws-devicefarm-gradle-plugin:1.2'
         classpath 'com.stanfy.spoon:spoon-gradle-plugin:1.2.1'
     }
--- a/platform/android/src/native_map_view.cpp
+++ b/platform/android/src/native_map_view.cpp
@@ -73,6 +73,10 @@ NativeMapView::NativeMapView(jni::JNIEnv& _env,
         MapMode::Continuous, GLContextMode::Unique, ConstrainMode::HeightOnly,
         ViewportMode::Default, jni::Make<std::string>(_env, _programCacheDir));
 
+    recalculateSourceTileCacheSize();
+}
+
+void NativeMapView::recalculateSourceTileCacheSize() {
     //Calculate a fitting cache size based on device parameters
     float zoomFactor   = map->getMaxZoom() - map->getMinZoom() + 1;
     float cpuFactor    = availableProcessors;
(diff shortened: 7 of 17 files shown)


brunoabinader:
@kkaefer I've added a sanitizer for the Qt wrapper when converting from `QSize` to `mbgl::Size`. I've also modified the approach to not touch `mbgl::Size` check, but instead failing in `Transform::resize` for zero-ed sizes and asserting a valid size in `mbgl::util::tileCover`.

include/mbgl/util/projection.hpp
```diff
@@ -13,7 +13,7 @@ namespace mbgl {
 class Projection {
 public:
     // Map pixel width at given scale.
-    static double worldSize(double scale) {
+    static constexpr double worldSize(double scale) {
```
jfirebaugh: What's the rationale for this change?

src/mbgl/map/transform.cpp
```diff
@@ -46,6 +46,10 @@ Transform::Transform(MapObserver& observer_,
 #pragma mark - Map View
 
 bool Transform::resize(const Size size) {
+    if (!size) {
+        return false;
```
jfirebaugh: If (0, 0) is an invalid size, shouldn't this throw rather than silently doing nothing?

src/mbgl/map/transform_state.hpp
```diff
@@ -20,7 +20,10 @@ class TransformState {
     friend class Transform;
 
 public:
-    TransformState(ConstrainMode = ConstrainMode::HeightOnly, ViewportMode = ViewportMode::Default);
+    constexpr TransformState() = default;
+
+    constexpr TransformState(ConstrainMode constrain_, ViewportMode viewport_)
```
jfirebaugh: This removes the ability to construct with a single argument. (Could you please limit your changes to fixing the main issue, and propose cosmetic changes in a separate PR?)

src/mbgl/map/transform_state.hpp
```diff
@@ -74,13 +77,17 @@ class TransformState {
...
+
 private:
     bool rotatedNorth() const;
     void constrain(double& scale, double& x, double& y) const;
 
     // Limit the amount of zooming possible on the map.
-    double min_scale = std::pow(2, 0);
-    double max_scale = std::pow(2, 20);
+    double min_scale = 1.0; // 2^0
+    double max_scale = 1048576.0; // 2^20
```
jfirebaugh: What's the rationale for this change?

src/mbgl/map/transform_state.hpp
```diff
@@ -74,13 +77,17 @@ class TransformState {
     double zoomScale(double zoom) const;
     double scaleZoom(double scale) const;
 
+    constexpr explicit operator bool() const {
```
jfirebaugh: Let's not use `operator bool()` for validity checks. If we need a validity check, name it `bool valid() const`. But first, try to eliminate the possibility of an invalid state.

platform/qt/src/qmapboxgl.cpp
```diff
@@ -1547,14 +1555,15 @@ QMapboxGLPrivate::QMapboxGLPrivate(QMapboxGL *q, const QMapboxGLSettings &settin
...
-    , mapObj(std::make_unique<mbgl::Map>(
-        *this, mbgl::Size{ static_cast<uint32_t>(size.width()), static_cast<uint32_t>(size.height()) },
-        pixelRatio, *fileSourceObj, *threadPool,
-        mbgl::MapMode::Continuous,
-        static_cast<mbgl::GLContextMode>(settings.contextMode()),
-        static_cast<mbgl::ConstrainMode>(settings.constrainMode()),
-        static_cast<mbgl::ViewportMode>(settings.viewportMode())))
 {
+    mapObj = std::make_unique<mbgl::Map>(
+            *this, sanitizedSize(size),
```
brunoabinader: Because `QSize` default constructor initializes witdh and height with -1. This causes an overflow when converting from `int` to `uint32_t`- using `std::numeric_limits<uint32_t>::max()` for width and height has proven to cause undefined behavior in places like [tileCover](https://github.com/mapbox/mapbox-gl-native/blob/master/src/mbgl/util/tile_cover.cpp#L159-L160).

Qt sometimes initializes a dummy map plugin as means of discovering which map plugins it has available - when it does, it uses a default `QSize` for initializing the item size.

src/mbgl/map/transform.cpp
```diff
@@ -46,6 +46,10 @@ Transform::Transform(MapObserver& observer_,
...
+    }
```
brunoabinader: Because if we allow `Transform` to have a zero-ed size, it can result in division-by-zero issues e.g. [here](https://github.com/mapbox/mapbox-gl-native/blob/master/src/mbgl/map/transform_state.cpp#L45).

src/mbgl/map/transform.cpp
brunoabinader: I supposed that was the meaning of the return `bool` - should we make it `void` instead and throw on such cases?

src/mbgl/map/transform_state.hpp
brunoabinader: These values are known and `std::pow` is not `constexpr`, causing the `TransformState` ctor to not be `constexpr`-able too.

src/mbgl/map/transform_state.hpp
brunoabinader: > Let's not use operator bool() for validity checks. 

We've discussed this before, yet we still use `explicit operator bool` in other parts of the code e.g. [Size::operator bool](https://github.com/mapbox/mapbox-gl-native/blame/94f011895c8e1bde36ee2ec235dbbcf2c994ac4c/include/mbgl/util/size.hpp#L20).

I don't have a strong opining about replacing it with something like `valid()`, however I'd prefer to have a separate PR that replaces every `explicit operator bool` usage for validation in our code 👍

src/mbgl/map/transform_state.hpp
brunoabinader: > This removes the ability to construct with a single argument.

Do we have a need for this? If so, we could add an extra ctor that receives a single `ConstrainMode`.

> Could you please limit your changes to fixing the main issue, and propose cosmetic changes in a separate PR?

Fair enough - though I've seen this as a common behavior in some of our PRs.

include/mbgl/util/projection.hpp
brunoabinader: Nothing special besides the fact `util::tileSize` is already known in compile time - as requested below I'll subtract this cosmetic change from this PR 👍

src/mbgl/map/transform_state.hpp
brunoabinader: 👉 https://github.com/mapbox/mapbox-gl-native/pull/8601.

include/mbgl/util/projection.hpp
jfirebaugh: Is there a need for `TransformState` to be `constexpr`? Based on the history with GCC 4.9 it sounded like something we should avoid using unless it's truly necessary.

src/mbgl/map/transform.cpp
jfirebaugh: I think at one point the return value was used to check if the size was _changing_, but it looks like this may be used only in tests. If so, yes, `void` and throw on invalid input.

brunoabinader: We can use try/catch in the tests 👍

include/mbgl/util/projection.hpp
brunoabinader: @jfirebaugh my understanding is that using `constexpr` allows clearer understanding of what we want to do, and what the compiler is possible of calculating in compile time - I personally find class/struct ctors using `constexpr` to be a more elegant pattern too.

There are specific cases where `constexpr` usage is bogus on GCC 4.9 - though this is not the case for simple cases like the ones above i.e. we still use plain `constexpr` in lots of places in our https://github.com/mapbox/mapbox-gl-native/tree/qt-staging branch.

As example, there are cases where `constexpr` works fine in GCC 4.9 - like this [example](https://godbolt.org/g/PQ4fRk), and cases [where it doesn't](https://godbolt.org/g/zmmCg8). 

Probably the best way to make sure we're compliant with GCC 4.9 is to have a GCC 4.9-based build bot running as part of our CI, but I guess that's out of scope for this PR.

brunoabinader:
@jfirebaugh I've moved the `constexpr` changes to a specific patch and made `Transform::resize` throw in case it receives an invalid size.

include/mbgl/util/projection.hpp
jfirebaugh: > using `constexpr` allows clearer understanding of what we want to do

How so? We don't inherently intend that these calculations or constructors happen at compile time, and the fact that they (currently) _can_ is not an essential feature of the code.

I don't think we should go around adding `constexpr` everywhere it's possible to do so. If we're talking about introducing coding guidelines for C++ features, then I would say my guideline for `constexpr` is "use only in cases where compile-time evaluation is an intended or essential feature of the code, or for general-purpose library functions which may need to be used in `constexpr` contexts."

platform/qt/src/qmapboxgl.cpp
```diff
@@ -79,6 +79,13 @@ auto fromQStringList(const QStringList &list)
     return strings;
 }
 
+mbgl::Size sanitizedSize(const QSize& size) {
+    return mbgl::Size {
+        size.width() <= 0 ? 0 : static_cast<uint32_t>(size.width()),
+        size.height() <= 0 ? 0 : static_cast<uint32_t>(size.height())
```
tmpsantos: You could use `std::max` here, but I'm OK with ternary.

brunoabinader: Good point, I'll make the change :+1:

brunoabinader:
@kkaefer @jfirebaugh can you please re-review?

jfirebaugh:
Can you replace `valid` with `empty` or `isEmpty` (and flip the meaning)? There's nothing inherently invalid about an empty size.

1ec5:
On iOS and macOS, it’s possible to create a view with zero size then give it size at some point before the view appears on screen, such as after the view is added to the view hierarchy. This pattern is especially common in iOS applications that build the interface programmatically using manual layout such as `-[UIView sizeToFit]`, which causes the view to resize to fit its subviews.

As reported in #1572, the current implementation on iOS and macOS accommodates this pattern only partially. On master, the following code:

```objc
MGLMapView *mapView = [[MGLMapView alloc] initWithFrame:CGRectZero];
mapView.frame = CGRectMake(0, 0, 300, 300);
[self.view addSubview:mapView];
```

results in the following console spew:

```
2017-04-06 13:32:11.888 Mapbox GL[72997:2170914] Failed to bind EAGLDrawable: <CAEAGLLayer: 0x60800042e7e0> to GL_RENDERBUFFER 1
2017-04-06 13:32:11.889 Mapbox GL[72997:2170914] Failed to make complete framebuffer object 8cd6
```

but otherwise everything is perfectly fine.

However, swap two lines so that the view is added to the view hierarchy before it gets some size, and the application crashes with an OpenGL error:

```
2017-04-06 13:32:56.752 Mapbox GL[73050:2172273] Failed to bind EAGLDrawable: <CAEAGLLayer: 0x600000234e00> to GL_RENDERBUFFER 1
2017-04-06 13:32:56.753 Mapbox GL[73050:2172273] Failed to make complete framebuffer object 8cd6
2017-04-06 13:32:56.756 Mapbox GL[73050:2172273] Failed to bind EAGLDrawable: <CAEAGLLayer: 0x600000234e00> to GL_RENDERBUFFER 1
2017-04-06 13:32:56.756 Mapbox GL[73050:2172273] Failed to make complete framebuffer object 8cd6
libc++abi.dylib: terminating with uncaught exception of type mbgl::gl::Error: glGetIntegerv(GL_VIEWPORT, viewport): Error GL_INVALID_ENUM at /Users/mxn/hub/mapbox-gl-native/src/mbgl/gl/value.cpp:264
```

As I understand it, this PR would ensure that the crash occurs earlier with a more informative error message. Unfortunately, it also causes the first case (setting the size right after initialization) to crash in the same way:

```
2017-04-06 13:40:43.244 Mapbox GL[73386:2182614] Failed to bind EAGLDrawable: <CAEAGLLayer: 0x60000043bbc0> to GL_RENDERBUFFER 1
2017-04-06 13:40:43.244 Mapbox GL[73386:2182614] Failed to make complete framebuffer object 8cd6
libc++abi.dylib: terminating with uncaught exception of type std::runtime_error: failed to resize: size is empty
```

I think this would be a surprising change for developers who build their interface programmatically. Is there any way to modify `mbgl::Map` so that it only trips this new size assertion when the map is visible?

If not, I suppose we could implement https://github.com/mapbox/mapbox-gl-native/issues/1572#issuecomment-291597674 by making 64×64 points the minimum size in [`-[MGLMapView size]`](https://github.com/mapbox/mapbox-gl-native/blob/7bcb6b7ed73ca7a45d502007f8187ba71e103954/platform/ios/src/MGLMapView.mm#L579), which is used when [creating `mbgl::Map`](https://github.com/mapbox/mapbox-gl-native/blob/7bcb6b7ed73ca7a45d502007f8187ba71e103954/platform/ios/src/MGLMapView.mm#L437) and when the view itself is resized:

```diff
diff --git a/platform/ios/src/MGLMapView.mm b/platform/ios/src/MGLMapView.mm
index d80e741c0..67b5ac09f 100644
--- a/platform/ios/src/MGLMapView.mm
+++ b/platform/ios/src/MGLMapView.mm
@@ -576,8 +576,12 @@ public:
 
 - (mbgl::Size)size
 {
-    return { static_cast<uint32_t>(self.bounds.size.width),
-             static_cast<uint32_t>(self.bounds.size.height) };
+    CGSize size = CGSizeMake(MIN(self.bounds.size.width, 64),
+                             MIN(self.bounds.size.height, 64));
+    return {
+        static_cast<uint32_t>(size.width),
+        static_cast<uint32_t>(size.height),
+    };
 }
 
 - (mbgl::Size)framebufferSize
diff --git a/platform/macos/src/MGLMapView.mm b/platform/macos/src/MGLMapView.mm
index e19755044..bccc7cff9 100644
--- a/platform/macos/src/MGLMapView.mm
+++ b/platform/macos/src/MGLMapView.mm
@@ -307,8 +307,12 @@ public:
 }
 
 - (mbgl::Size)size {
-    return { static_cast<uint32_t>(self.bounds.size.width),
-             static_cast<uint32_t>(self.bounds.size.height) };
+    CGSize size = CGSizeMake(MIN(self.bounds.size.width, 64),
+                             MIN(self.bounds.size.height, 64));
+    return {
+        static_cast<uint32_t>(size.width),
+        static_cast<uint32_t>(size.height),
+    };
 }
 
 - (mbgl::Size)framebufferSize {
```

Methods like `-[MGLMapView convertCoordinate:toPointToView:]` will return incorrect values until the map is resized beyond 64×64 points, but that’s still better than crashing. Besides `-size`, would `-framebufferSize` also need to be limited like this?

/cc @boundsj

1ec5:
Note that, even with this change, the console spew about failing to bind the drawable persists but appears to be harmless:

```
2017-04-06 15:13:01.489 Mapbox GL[75226:2237986] Failed to bind EAGLDrawable: <CAEAGLLayer: 0x608000225780> to GL_RENDERBUFFER 1
2017-04-06 15:13:01.490 Mapbox GL[75226:2237986] Failed to make complete framebuffer object 8cd6
```

brunoabinader:
> On iOS and macOS, it’s possible to create a view with zero size then give it size at some point before the view appears on screen

We are not going to modify this pattern - as you mentioned the size can be silently updated to 64x64 inside `size` getter (or `initWithFrame`).

The suggested patch for iOS is no different from the approach took by [MGLStyleTests](https://github.com/mapbox/mapbox-gl-native/blob/master/platform/darwin/test/MGLStyleTests.mm#L31) and [MGLSourceQueryTests](https://github.com/mapbox/mapbox-gl-native/blob/master/platform/darwin/test/MGLSourceQueryTests.m#L28).

>  Is there any way to modify mbgl::Map so that it only trips this new size assertion when the map is visible?

Initializing a map with a _valid_ (by _valid_ I mean at least the minimum texture size as specified by the OpenGL ES 2.0 which is 64x64) size is currently mandatory in our code - because we are immediately supposed to be able to call `render()`. One alternative would be to have a check in core to ignore `render()` calls until we have a valid texture size set.

Complexity aside, I believe the less painful change would be to just set the size to 64x64 silently when passing it to the `Map` object - this way developers won't notice the change and we can be safeguarded. This approach is i.e. the proposed one for Qt and Android already.

1ec5:
Alright, let’s go with silently upping the size to 64×64 points. Does `-framebufferSize` need to have a minimum of 64×64 points as well?

brunoabinader:
> Does -framebufferSize need to have a minimum of 64×64 points as well?

The only restriction I'm aware of for framebuffer sizes is that their width and height values must be a power of 2 for OpenGL versions prior to 2.0.

Issue #8489, opened 2017-03-22
Annotation view reuse causes annotation views to disappear when the backing annotation coordinate changes to be located out of the viewport


bulbuliator:
**Platform:** iOS
**Mapbox SDK version:** 3.5.0-rc.2

### Steps to trigger behavior

 1. Add view-based annotation as described in [example](https://www.mapbox.com/ios-sdk/examples/annotation-views/)
 2. Change coordinate randomly on 200-300 meters by Core Animation UIView.animate method
 3. Zoom to level >= 13

### Expected behavior

When new coordinate is assigned out of the screen bounds move annotation animated to the screen edge 

### Actual behavior

Annotation disappears instantly without animation

boundsj:
Hi @bulbuliator. I don't think I understand from the repro steps above what the exact issue is or what you are trying to accomplish. Can you please provide a code sample that clearly illustrates what you are trying to do? You may also find the [demo application's implicit animation implementation](https://github.com/mapbox/mapbox-gl-native/blob/ios-v3.4.2/platform/ios/app/MBXAnnotationView.m#L52-L63) useful. Thanks!

bulbuliator:
Hi, @boundsj. 

<details> 
  <summary>My code sample </summary>
   
```swift
class ViewController: UIViewController, MGLMapViewDelegate {

    var point = MGLPointAnnotation()
    
    override func viewDidLoad() {
        super.viewDidLoad()
        
        let mapView = MGLMapView(frame: view.bounds)
        mapView.autoresizingMask = [.flexibleWidth, .flexibleHeight]
        mapView.styleURL = MGLStyle.darkStyleURL(withVersion: 9)
        mapView.tintColor = .lightGray
        mapView.centerCoordinate = CLLocationCoordinate2D(latitude: 0, longitude: 66)
        mapView.zoomLevel = 3
        mapView.delegate = self
        view.addSubview(mapView)
        
        let coordinate = CLLocationCoordinate2D(latitude: 0, longitude: 66)
        point.coordinate = coordinate
        point.title = "\(coordinate.latitude), \(coordinate.longitude)"
        
        mapView.addAnnotation(point)
        
        startLocationSimulatorTimer()
        
    }

    override func viewWillDisappear(_ animated: Bool) {
        super.viewWillDisappear(animated)
        stopLocationSimulatorTimer()
    }
    
    // MARK: - Random annotation moving methods
    
    // Timer to simulate point moving
[...]

PR #8565, opened 2017-03-29
[ios] added check for annotation view animation

jmkiley:
Checks whether an annotation view has any `animationKeys` in `updateAnnotations` before adding the annotation view to reuse queue. Should prevent annotations that are being moved offscreen from disappearing. 

 

Bug and Fix

Addresses #8489, uses @boundsj's suggested fix.

### DIFF
--- a/platform/ios/app/MBXAnnotationView.m
+++ b/platform/ios/app/MBXAnnotationView.m
@@ -49,17 +49,4 @@
 
 }
 
-- (nullable id<CAAction>)actionForLayer:(CALayer *)layer forKey:(NSString *)event
-{
-    if (([event isEqualToString:@"transform"] || [event isEqualToString:@"position"])
-        && self.dragState == MGLAnnotationViewDragStateNone)
-    {
-        CABasicAnimation *animation = [CABasicAnimation animationWithKeyPath:event];
-        animation.timingFunction = [CAMediaTimingFunction functionWithName:kCAMediaTimingFunctionEaseInEaseOut];
-        animation.speed = 0.1;
-        return animation;
-    }
-    return [super actionForLayer:layer forKey:event];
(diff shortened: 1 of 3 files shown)


platform/ios/app/MBXAnnotationView.m
```diff
@@ -49,17 +49,4 @@ - (void)setDragState:(MGLAnnotationViewDragState)dragState animated:(BOOL)animat
 
 }
 
-- (nullable id<CAAction>)actionForLayer:(CALayer *)layer forKey:(NSString *)event
```
boundsj: Just noting that I think removing this makes sense only because I also think that we should feature UIView animation blocks as the recommended way of performing annotation view animations. Animation blocks have the convenient side effect of animating any open callout view along with the annotation view. cc @1ec5

platform/ios/app/MBXViewController.m
```diff
@@ -703,6 +708,22 @@ - (void)parseFeaturesAddingCount:(NSUInteger)featuresCount usingViews:(BOOL)useV
     });
 }
 
+- (void)animateAnnotationView
+    {
+        MGLPointAnnotation *annot = [[MGLPointAnnotation alloc] init];
+        annot.coordinate = self.mapView.centerCoordinate;
+        [self.mapView addAnnotation:annot];
+        
+        CGPoint point = CGPointMake(-200, CGRectGetMidY(self.view.frame));
```
boundsj: The magic value (-200) happens to move annotation views into the view reuse area outside of the viewport. If we change the size of annotation views in the future it may not. Would it be worth it to somehow special case this annotation to guarantee the size of the view? If not, it may at least be worth adding a comment here explaining why this value works today.

PR #8568, opened 2017-03-29
Create .topissuesrc

lucaswoj:
There are many reasons why different projects may want to prioritize issues differently. I just added the ability for projects to specify a custom `.topissuesrc` file which determines how labels and reactions are scored. (We may want to add more knobs and dials to this file in the future.)

Some light docs about `.topissuesrc`: https://github.com/mapbox/top-issues/blob/master/README.md#topissuesrc

Note to self: Once this merges it might be nice to reformulate the default top-issues settings to play nicely with the default Github issue labels

### DIFF
--- a/.topissuesrc
+++ b/.topissuesrc
@@ -2,8 +2,6 @@
     "labels": {
         "release blocker": 25,
         "beta blocker": 20,
-        "high priority": 7,
-        "medium priority": 5,
         "crash": 5,
         "build": 5,
         "bug": 2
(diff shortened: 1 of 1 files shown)


.topissuesrc
```diff
@@ -0,0 +1,19 @@
+{
+    "labels": {
+        "release blocker": 25,
+        "beta blocker": 20,
+        "high priority": 7,
+        "medium priority": 5,
```
1ec5: This repository doesn’t use high/medium/low priority labels.

lucaswoj: Rodger that. Fixed in the next commit.

PR #8576, opened 2017-03-30
[android] - Style wide transition options in milliseconds

### LINKED ISSUE #8575: TransitionOptions conversion when set
When looking into adding transition properties for layer attributes in #8015. I noticed that the style wide configuration of TransitionOptions in https://github.com/mapbox/mapbox-gl-native/pull/7711 is not returning the same values as the value set. Eg. following test is failing:

```java
long transitionDuration = 1;
mapboxMap.setTransitionDuration(transitionDuration);
assertEquals("TransitionDuration should match", transitionDuration, mapboxMap.getTransitionDuration(), 0);
```
with 
```
AssertionFailedWithCauseError: TransitionDuration should match expected:<1.0> but was:<1.0E9>
```
Looking at the [style specification](https://www.mapbox.com/mapbox-gl-js/style-spec/#transition-duration), it seems that the unit should be in milliseconds while it nows seems to be set in seconds and returned in nano seconds.

tobrun:
Closes #8575 

cc @incanus

### DIFF
--- a/platform/android/MapboxGLAndroidSDK/src/main/java/com/mapbox/mapboxsdk/maps/MapboxMap.java
+++ b/platform/android/MapboxGLAndroidSDK/src/main/java/com/mapbox/mapboxsdk/maps/MapboxMap.java
@@ -194,7 +194,7 @@ public final class MapboxMap {
    * </p>
    * The default value is zero, so any changes take effect without animation.
    *
-   * @return Duration in seconds
+   * @return Duration in milliseconds
    */
   @UiThread
   public long getTransitionDuration() {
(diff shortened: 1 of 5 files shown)


platform/android/MapboxGLAndroidSDK/src/main/java/com/mapbox/mapboxsdk/maps/MapboxMap.java
```diff
@@ -204,11 +204,11 @@ public long getTransitionDuration() {
   /**
    * Set the animation duration for style changes.
    *
-   * @param duration Duration in seconds
+   * @param durationMs Duration in milliseconds
    */
   @UiThread
-  public void setTransitionDuration(long duration) {
```
ivovandongen: Do we consider this a breaking change?

jfirebaugh: Why not use `java.time.Duration` for durations and avoid unit bugs through type safety?

tobrun: Ivo mentioned the same on chat regards using Duration from joda.time, will ticket it out.

PR #8581, opened 2017-03-30
Fix compilation with Android NDK r13b

kkaefer:
Clang 3.8 and below seems to be unable to implicitly convert derived objects when they inherit from the same base class.

### DIFF
--- a/circle.yml
+++ b/circle.yml
@@ -1,0 +1,84 @@
+version: 2
+
+jobs:
+
+# ------------------------------------------------------------------------------
+  build:
+    docker:
+      - image: mbgl/ci:trigger_job
+    working_directory: /
+    steps:
+      - deploy:
+          name: Trigger 'android-debug-arm-v7'
+          command: trigger_job android-debug-arm-v7
+      - deploy:
+          name: Trigger 'android-release-all'
+          command: trigger_job android-release-all
+
+# ------------------------------------------------------------------------------
+  android-debug-arm-v7:
+    docker:
+      - image: mbgl/android-ci:ndk-r13b
+    working_directory: /src
+    environment:
+      LD_PRELOAD: /usr/lib/libsysconfcpus.so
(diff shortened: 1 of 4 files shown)


kkaefer:
@jfirebaugh I'm not sure what it is about the older Clang that makes it no convert `multi_point` to `line_string`. Maybe you have more insight into this? It feels odd that the constructor doesn't accept the derived object, since both are ultimately derived from the same base class.

daniel-j-h:
I think you may be hitting the inheriting-constructor language defect, see the following and its refs.
http://daniel-j-h.github.io/post/strong-typedefs-inheriting-constructors/

The defect report should be implemented in gcc7 and clang 3.9, see this thread
https://www.reddit.com/r/cpp/comments/5hvf92/c17_stdvariant_lands_in_llvm/db42a7h/

---

Upstream, instead of inheriting from a std collection and inheriting its constructors here

- https://github.com/mapbox/geometry.hpp/blob/b0d64e6e8d55026781325df1068e8043ff880aff/include/mapbox/geometry/multi_point.hpp#L17
- https://github.com/mapbox/geometry.hpp/blob/b0d64e6e8d55026781325df1068e8043ff880aff/include/mapbox/geometry/line_string.hpp#L17

a probably more robust design would not inherit from the std collection but store it as a private member and delegate to its functions through a public interface.

```c++
template <typename T, template <typename...> class Cont = std::vector>
struct multi_point
{
    using coordinate_type = T;
    using point_type = point<T>;
    using container_type = Cont<point_type>;

    std::size_t size() { return cont.size(); } // expose only what you need

    private:
      Cont<point<T>> cont;
};
```

jfirebaugh:
@daniel-j-h We considered that design; the issue is that we really do want the entire base interface exposed. (Well, what we really want is a "strong" typedef, but C++ doesn't offer that.)

Issue #3602, opened 2016-01-19
Limit Viewport


tobrun:
I have been hearing about the feature that developers want to limit the viewport for the end-user.
##### Sample use-case:

Create a tourist application for NYC, users are restricted to:
- move the map in NYC area 
- only allowed to zoom between 8 - 14 level 
- tilt is limited to only 20 degrees.  
##### API:

This will result in exposing a similar API as the old raster API:
- setScrollableAreaLimit() // definition ready
- setMinZoom() // implemented
- setMaxZoom() // implemented
- setMinTilt()
- setMaxTilt()

iOS is currently thinking about implementing https://github.com/mapbox/mapbox-gl-native/issues/2457, which will work similar to the rotation snap-back animation we currently have at very low zoom levels.

1ec5:
With the map tilted, does it matter that the user will be able to see beyond the scrollable area limit?

1ec5:
In other words, should these properties constrain the center point or the viewable bounds? If the latter, there could potentially be a complex interaction between the scrollable area limit, zoom, and tilt constraints.

friedbunny:
#2341 is one suggested approach, on iOS.

tobrun:
> With the map tilted, does it matter that the user will be able to see beyond the scrollable area limit?

For this Google Maps SDK exposes [VisibleRegion]((https://developers.google.com/android/reference/com/google/android/gms/maps/model/VisibleRegion) ):

> Contains the four points defining the four-sided polygon that is visible in a map's camera. This polygon can be a trapezoid instead of a rectangle, because a camera can have tilt. If the camera is directly over the center of the camera, the shape is rectangular, **but if the camera is tilted, the shape will appear to be a trapezoid whose smallest side is closest to the point of view**.

Which seems a bit more complex than #2341

Zakay:
+1 on setScrollableAreaLimit
At least something like #2341

daklock:
+1 on limiting scrollview

s-a-y:
+1 limiting scrollview

bleege:
@tobrun Let's get this into `4.0.0`.  Cool?

tobrun:
@bleege :+1: 

@zugaldia 
any preference on how we are going to implement this? 
This seems to be an overlapping requirement with offline?

zugaldia:
> @zugaldia any preference on how we are going to implement this? This seems to be an overlapping requirement with offline?

[...]

PR #6088, opened 2016-08-19
[build] geojsonvt @ v6.1.3

brunoabinader:
Fixes annotations antimeridian wrapping.

Fixes #5600 and #6043.

PR #8583, opened 2017-03-30
[core] AddedMap::{set,get}LatLngBounds

brunoabinader:
Implements a setter and getter for coordinate bounds in `Map` API - which can be used e.g. to limit the boundaries in which the user can set geographical coordinates.

Related: #3602.

### DIFF
--- a/include/mbgl/util/geo.hpp
+++ b/include/mbgl/util/geo.hpp
@@ -1,5 +1,6 @@
 #pragma once
 
+#include <mbgl/math/clamp.hpp>
 #include <mbgl/math/wrap.hpp>
 #include <mbgl/util/constants.hpp>
 
@@ -120,6 +121,10 @@ public:
     // Constructs a LatLngBounds object with the tile's exact boundaries.
     LatLngBounds(const CanonicalTileID&);
 
+    bool valid() const {
+        return (sw.latitude() <= ne.latitude()) && (sw.longitude() <= ne.longitude());
+    }
+
     double south() const { return sw.latitude(); }
     double west()  const { return sw.longitude(); }
     double north() const { return ne.latitude(); }
@@ -133,6 +138,16 @@ public:
     LatLng center() const {
         return LatLng((sw.latitude() + ne.latitude()) / 2,
                       (sw.longitude() + ne.longitude()) / 2);
+    }
+
+    LatLng constrain(const LatLng& p) const {
+        if (contains(p)) {
+            return p;
+        }
+        return LatLng {
+            util::clamp(p.latitude(), sw.latitude(), ne.latitude()),
+            util::clamp(p.longitude(), sw.longitude(), ne.longitude())
+        };
     }
 
     void extend(const LatLng& point) {
--- a/src/mbgl/map/transform.cpp
+++ b/src/mbgl/map/transform.cpp
@@ -435,6 +435,12 @@ void Transform::setScale(double scale, optional<EdgeInsets> padding, const Anima
 
 #pragma mark - Bounds
 
+void Transform::setLatLngBounds(const LatLngBounds& bounds) {
+    if (bounds.valid()) {
+        state.setLatLngBounds(bounds);
+    }
+}
+
 void Transform::setMinZoom(const double minZoom) {
     if (std::isnan(minZoom)) return;
     state.setMinZoom(minZoom);
--- a/src/mbgl/map/transform.cpp
+++ b/src/mbgl/map/transform.cpp
@@ -120,7 +120,7 @@ void Transform::easeTo(const CameraOptions& camera, const AnimationOptions& anim
     // Constrain camera options.
     zoom = util::clamp(zoom, state.getMinZoom(), state.getMaxZoom());
     const double scale = state.zoomScale(zoom);
-    pitch = util::clamp(pitch, 0., util::PITCH_MAX);
+    pitch = util::clamp(pitch, state.min_pitch, state.max_pitch);
 
     Update update = state.getZoom() == zoom ? Update::Repaint : Update::RecalculateStyle;
 
@@ -188,7 +188,7 @@ void Transform::flyTo(const CameraOptions &camera, const AnimationOptions &anima
 
     // Constrain camera options.
     zoom = util::clamp(zoom, state.getMinZoom(), state.getMaxZoom());
-    pitch = util::clamp(pitch, 0., util::PITCH_MAX);
+    pitch = util::clamp(pitch, state.min_pitch, state.max_pitch);
 
     // Minimize rotation by taking the shorter path around the circle.
     angle = _normalizeAngle(angle, state.angle);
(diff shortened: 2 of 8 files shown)


include/mbgl/util/geo.hpp
```diff
@@ -103,14 +107,18 @@ class LatLngBounds {
...
+        LatLngBounds bounds = LatLngBounds::world();
         std::swap(bounds.sw, bounds.ne);
         return bounds;
     }
 
     // Constructs a LatLngBounds object with the tile's exact boundaries.
     LatLngBounds(const CanonicalTileID&);
 
+    explicit operator bool() const {
+        return (sw.latitude <= ne.latitude) && (sw.longitude <= ne.longitude);
```
1ec5: This would be consistent with the GeoJSON specification, but inconsistent with GeoJSONVT’s behavior of allowing a single feature or annotation to span the antimeridian (#6088), correct? If one were to [calculate the bounding box of an annotation](https://github.com/mapbox/mapbox-gl-native/blob/62ea1f21858c69f6921c775ba7a3de201f0514d8/platform/darwin/src/MGLMultiPoint.mm#L164-L167), would this operator end up evaluating to `false`? Would it still be possible to [fit the bounds to the annotation](https://github.com/mapbox/mapbox-gl-native/blob/6c92199a16a9fc6648244547a183bc80da7a56d7/platform/ios/src/MGLMapView.mm#L3960-L3972)?

brunoabinader:
>  I have some concerns about the antimeridian edge case (the gift that keeps on giving)

Do you have one in particular? From my tests, this shouldn't be an issue as every coordinate passed to the map gets unwrapped e.g. [here](https://github.com/mapbox/mapbox-gl-native/blob/master/src/mbgl/map/transform.cpp#L91).

brunoabinader:
> #3602 (comment) called for a setMinTilt() and setMaxTilt(), which would imply a setMinPitch() and setMaxPitch() in mbgl::Map. Would that fall within the scope of this PR?

Yep, we can certainly add it too 😃

include/mbgl/util/geo.hpp
brunoabinader: My understanding of [LatLngBounds::extend](https://github.com/mapbox/mapbox-gl-native/blob/master/include/mbgl/util/geo.hpp#L129) is that it preserves the assertion above. Also, this should be fine as long as the coordinates used are unwrapped.

Nonetheless, we can add a unit test to verify 👍

1ec5:
>> I have some concerns about the antimeridian edge case (the gift that keeps on giving)

> Do you have one in particular? From my tests, this shouldn't be an issue as every coordinate passed to the map gets unwrapped e.g. here.

I was referring to https://github.com/mapbox/mapbox-gl-native/pull/8583#discussion_r108976652. But as long as you’ve considered this possibility, 👍.

include/mbgl/util/geo.hpp
```diff
@@ -23,17 +23,19 @@ class LatLng {
...
 
     enum WrapMode : bool { Unwrapped, Wrapped };
 
-    LatLng(null) : latitude(std::numeric_limits<double>::quiet_NaN()), longitude(latitude) {}
+    constexpr LatLng() = default;
 
-    LatLng(double lat = 0, double lon = 0, WrapMode mode = Unwrapped)
+    constexpr LatLng(null) : latitude(std::numeric_limits<double>::quiet_NaN()), longitude(latitude) {}
+
+    constexpr LatLng(double lat, double lon, WrapMode mode = Unwrapped)
```
brunoabinader: By making `LatLngBounds::world()` constexpr [here](https://github.com/mapbox/mapbox-gl-native/pull/8583/commits/6fce4a65b242fc2d185c787104493b3c0770b511#diff-cf54d0ce1dec44ab05dcd5daf776c7f8R92), `LatLngBounds` needed a constexpr ctor, which required its members (`LatLng` for southwest and northeast bounds) to also have a constexpr ctor. We use`LatLngBounds::world()` to initialize the default `TransformState` bounds [here](https://github.com/mapbox/mapbox-gl-native/pull/8583/commits/4b8849b596fedaaeca4c7437eba9fbb45aa6f776#diff-14cb746fd86755f6100a64ac3c8ee6a9R85).

src/mbgl/map/transform.cpp
```diff
@@ -438,12 +438,14 @@ void Transform::setScale(double scale, optional<EdgeInsets> padding, const Anima
...
-void Transform::setMinZoom(const double minZoom) {
+#pragma mark - Bounds
+
+void Transform::setMinZoom(double minZoom) {
     if (std::isnan(minZoom)) return;
     state.setMinZoom(minZoom);
 }
 
-void Transform::setMaxZoom(const double maxZoom) {
+void Transform::setMaxZoom(double maxZoom) {
```
brunoabinader: As you mentioned, there is no point in adding this to headers. Also, _const-pass-by-value_ is not a common behavior in our codebase.

For me, it is clear that a non-const _pass-by-value_ is a different object from the one it has been copied from i.e. declaring it `const` does not guarantee that the original object remains immutable.

src/mbgl/map/transform_state.cpp
```diff
@@ -321,16 +330,18 @@ void TransformState::moveLatLng(const LatLng& latLng, const ScreenCoordinate& an
...
     Bc = newWorldSize / util::DEGREES_MAX;
     Cc = newWorldSize / util::M2PI;
 
     const double m = 1 - 1e-15;
-    const double f = util::clamp(std::sin(util::DEG2RAD * latLng.latitude), -m, m);
+    const double f = util::clamp(std::sin(util::DEG2RAD * constrained.latitude), -m, m);
 
     ScreenCoordinate point = {
-        -latLng.longitude * Bc,
+        -constrained.longitude * Bc,
```
brunoabinader: That's where we constrain the received `LatLng` to the specified bounds. I preferred this to be done when inserting because I suppose we set `LatLng` values much less than we request them from `TransformState`.

tobrun:
Been testing out and adding binding integration in #8622, 
thank you for picking this up @brunoabinader.

brunoabinader:
Removed the `constexpr`-related changes and added `{set,get}{Min,Max}Pitch`.

src/mbgl/map/transform.cpp
kkaefer: Sorry, what I meant to say is that I'm doing this as a convention to avoid accidental changes to the parameter that might get read later on in the function.

include/mbgl/util/geo.hpp
```diff
@@ -126,6 +130,16 @@ class LatLngBounds {
                       (sw.longitude + ne.longitude) / 2);
     }
 
+    LatLng constrain(const LatLng& p) const {
+        if (contains(p)) {
+            return p;
+        }
+        return LatLng {
+            p.latitude < sw.latitude ? sw.latitude : p.latitude > ne.latitude ? ne.latitude : p.latitude,
+            p.longitude < sw.longitude ? sw.longitude : p.longitude > ne.longitude ? ne.longitude : p.longitude,
```
kkaefer: Parentheses to clarify operator precedence would make the code mode readable

src/mbgl/map/transform.cpp
brunoabinader: Ack - let's add them back on the definition file 👍

brunoabinader:
@kkaefer can you please re-review?

Issue #8163, opened 2017-02-22
Crash on render with view-backed annotations present — -[MGLMapView visibleAnnotationsInRect:]


ollie-eman:
**Platform:** iOS
**Mapbox SDK version:** 3.4.1

### Steps to trigger behaviour

Updated to the latest SDK version and get this error, it has effected over 1000 users in the past 7 days. Could you help me understand it so I can get a fix deployed?

```
Incident Identifier: D3B3C119-B99F-45D9-96DD-68BEADF88543
CrashReporter Key:   44069ad0adad7f6fe68b477dddc9285b987cc622
Hardware Model:      iPhone9,3
Process:             Snatch Production [7629]
Path:                /private/var/containers/Bundle/Application/9E753DDA-4C28-48DD-A30D-12A2C702B907/Snatch Production.app/Snatch Production
Identifier:          com.iPhone.Snatch
Version:             2 (1.2.4)
Code Type:           ARM-64 (Native)
Role:                Foreground
Parent Process:      launchd [1]
Coalition:           com.iPhone.Snatch [3379]


Date/Time:           2017-02-20 21:07:17.2052 +0000
Launch Time:         2017-02-20 19:36:04.1710 +0000
OS Version:          iPhone OS 10.3 (14E5239e)
Report Version:      104

Exception Type:  EXC_CRASH (SIGABRT)
Exception Codes: 0x0000000000000000, 0x0000000000000000
Exception Note:  EXC_CORPSE_NOTIFY
Triggered by Thread:  0

Last Exception Backtrace:
0   CoreFoundation                	0x18c5a2fb8 __exceptionPreprocess + 124 (NSException.m:165)
1   libobjc.A.dylib               	0x18b008538 objc_exception_throw + 56 (objc-exception.mm:521)
2   CoreFoundation                	0x18c484868 -[__NSArrayM insertObject:atIndex:] + 1372 (NSArray.m:654)
3   Mapbox                        	0x100dbf8c8 -[MGLMapView visibleAnnotationsInRect:] + 260 (MGLMapView.mm:2866)
4   Mapbox                        	0x100dc94bc -[MGLMapView updateAnnotationViews] + 408 (MGLMapView.mm:4728)
5   Mapbox                        	0x100dc8aa8 -[MGLMapView notifyMapChange:] + 796 (MGLMapView.mm:4681)
[...]

PR #8513, opened 2017-03-23
[ios] Crash fix on render with view-backed annotations

fabian-guerra:
Fixes #8163 while this is only a nil check, we need to continue investigating why `annotationContext.annotation` is nil when calling `visibleAnnotationsInRect`

PR #8588, opened 2017-03-30
[ios] assertion to check if an annotationTag > 0

fabian-guerra:
Addresses https://github.com/mapbox/mapbox-gl-native/issues/8163#issuecomment-290246849

### DIFF
--- a/platform/ios/src/MGLMapView.mm
+++ b/platform/ios/src/MGLMapView.mm
@@ -3044,10 +3044,9 @@ public:
                 continue;
             }
             MGLAnnotationContext annotationContext = _annotationContextsByAnnotationTag.at(annotationTag);
-            if (annotationContext.annotation)
-            {
-                [annotations addObject:annotationContext.annotation];
-            }
+            NSAssert(annotationContext.annotation, @"Missing annotation for tag %u.", annotationTag);
+            [annotations addObject:annotationContext.annotation];
+
         }
 
         return [annotations copy];
--- a/platform/ios/src/MGLMapView.mm
+++ b/platform/ios/src/MGLMapView.mm
@@ -3045,7 +3045,10 @@ public:
             }
             MGLAnnotationContext annotationContext = _annotationContextsByAnnotationTag.at(annotationTag);
             NSAssert(annotationContext.annotation, @"Missing annotation for tag %u.", annotationTag);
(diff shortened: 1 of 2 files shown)


platform/ios/src/MGLMapView.mm
```diff
@@ -3044,6 +3044,7 @@ - (void)removeStyleClass:(NSString *)styleClass
                 continue;
             }
             MGLAnnotationContext annotationContext = _annotationContextsByAnnotationTag.at(annotationTag);
+            NSAssert(_annotationContextsByAnnotationTag.count(annotationTag), @"Missing annotation for tag %u.", annotationTag);
```
1ec5: This assertion checks whether there’s an entry for `annotationTag` in `_annotationContextsByAnnotationTag`. There’s no way this assertion can fail, because we already have a guard a few lines above that `continue`s in that case. Instead, here, we need to assert that `annotationContext.annotation` is non-nil.

platform/ios/src/MGLMapView.mm
```diff
@@ -3044,10 +3044,9 @@ - (void)removeStyleClass:(NSString *)styleClass
                 continue;
             }
             MGLAnnotationContext annotationContext = _annotationContextsByAnnotationTag.at(annotationTag);
-            if (annotationContext.annotation)
-            {
-                [annotations addObject:annotationContext.annotation];
-            }
+            NSAssert(annotationContext.annotation, @"Missing annotation for tag %u.", annotationTag);
+            [annotations addObject:annotationContext.annotation];
```
boundsj: https://github.com/mapbox/mapbox-gl-native/issues/8163#issuecomment-290246849 mentions supplementing the fix with an assertion. This change fully removes the guard added in https://github.com/mapbox/mapbox-gl-native/pull/8513. We can keep the assertion although I'm now confident we will hit it (or other similar assertions) from time to time when developing because of what I described in https://github.com/mapbox/mapbox-gl-native/issues/8163#issuecomment-291695262

1ec5: While you’re at it, please add the same assertion and guard to the macOS implementation of `-[MGLMapView visibleAnnotationsInRect:]`. We should keep the two implementations consistent. Thanks!

PR #8593, opened 2017-03-31
[core] Add DDS support for {text,icon}-size

anandthakker:
- [x] Work through render test errors and segfaults ( 😱 )
 - [x] Review `SymbolSizeData` / `SymbolSizeAttributes` design ( @jfirebaugh could I get your take on this?   Main place to look/start is probably [top of symbol_program.cpp](https://github.com/mapbox/mapbox-gl-native/pull/8593/commits/32f95a9368ea58652d429f95ca0ebba6597ea438#diff-5907166344b0d46b118326c98eb9915cR91), [SymbolLayout::addSymbol](https://github.com/mapbox/mapbox-gl-native/pull/8593/commits/32f95a9368ea58652d429f95ca0ebba6597ea438#diff-f27a3a157c549a13f8fe1ee8c8157ea8R605), and maybe [makeValues()](https://github.com/mapbox/mapbox-gl-native/pull/8593/commits/32f95a9368ea58652d429f95ca0ebba6597ea438#diff-b45e5b497062a1fdda23428c5936cfaeR36))
 - [x] Refactor `SymbolProgram`, currently just copy-pasted from `Program` to make for easier initial hacking.
 - [x] Correct naive linear interpolation to make use of exponential base.
 - [x] Smaller TODOs
 - [x] Request full review


src/mbgl/programs/symbol_program.hpp
```diff
@@ -57,19 +73,207 @@ struct SymbolLayoutAttributes : gl::Attributes<
...
+                )
             }}
         };
     }
 };
+    
+struct SymbolSizeAttributes : gl::Attributes<attributes::a_size> {
+    using SourceFunctionVertex = gl::detail::Vertex<gl::Attribute<uint16_t, 1>>;
+    struct monostate {};
+    using VertexVector = variant<
```
anandthakker: Ah, yeah!  I started out thinking that duplicating a PaintPropertyBinder hierarchy like that would be overkill, but actually it could be pretty minimal/scoped and, like you say, cleaner than having all these different `match`es.  🙇

anandthakker:
@jfirebaugh refactored to a paintpropertybinders-like polymorphic approach https://github.com/mapbox/mapbox-gl-native/pull/8593/commits/f219d8c67b3ceb9b360713782104bc15d6e3bf14, as you suggested.  It's a few more LOC, but it does feel like a cleaner design.  The repetition of the PaintPropertyBinders hierarchy makes me itch to refactor the redundancy, but it would probably be premature abstraction to do it just yet; I think we should [wait till we need a third copy](https://en.wikipedia.org/wiki/Rule_of_three_(computer_programming)).

anandthakker:
> Refactor SymbolProgram, currently just copy-pasted from Program to make for easier initial hacking

I think the main issue here is to dedupe the identical constructors of `SymbolProgram` and `Program` (e.g.: https://github.com/mapbox/mapbox-gl-native/pull/8593/files#diff-5907166344b0d46b118326c98eb9915cR312).  They're really just initializing a `gl::Program`, so I think that logic can be extracted into a static factor function or maybe even just a constructor for `gl::Program`.

src/mbgl/gl/program.hpp
```diff
@@ -37,6 +41,57 @@ class Program {
           attributeLocations(Attributes::loadNamedLocations(binaryProgram)),
           uniformsState(Uniforms::loadNamedLocations(binaryProgram)) {
     }
+    
+    template <class Shaders>
```
jfirebaugh: To reduce binary size, pass `vertexSource`, `fragmentSource`, and `name` as parameters rather than templating on `Shaders`.

src/mbgl/programs/attributes.hpp
```diff
@@ -8,17 +8,32 @@
...
+
 // Layout attributes
 
 MBGL_DEFINE_ATTRIBUTE(int16_t, 2, a_pos);
 MBGL_DEFINE_ATTRIBUTE(int16_t, 2, a_extrude);
 MBGL_DEFINE_ATTRIBUTE(int16_t, 4, a_pos_offset);
 MBGL_DEFINE_ATTRIBUTE(uint16_t, 2, a_texture_pos);
 
-template <std::size_t N>
+template <std::size_t N, typename T>
```
jfirebaugh: Flip the order of `T` and `N`, so that it matches `gl::Attribute`.

anandthakker:
@jfirebaugh updated to address your comments and also rebased + cleaned up the history a bit.  Should be ready for a full review now.

PR #8600, opened 2017-03-31
Analysis of mapbox-gl-native source code by PVS-Studio

DieselMachine84:
To demonstrate the abilities of PVS-Studio analyzer, we decided to find several bugs.
I would like to suggest a variant of the way to fix the error, detected with the help of V603 diagnostic: The object was created but it is not being used. If you wish to call constructor, 'this->ThreadLocal::ThreadLocal(....)' should be used.

PVS-Studio is a tool for bug detection in the source code of programs, written in C, C++ and C#. It works in Windows and Linux environment. https://www.viva64.com/en/pvs-studio/

We suggests having a look at the emails, sent from @pvs-studio.com.

### DIFF
--- a/src/mbgl/util/thread_local.hpp
+++ b/src/mbgl/util/thread_local.hpp
@@ -12,8 +12,7 @@ namespace util {
 template <class T>
 class ThreadLocal : public noncopyable {
 public:
-    ThreadLocal(T* val) {
-        ThreadLocal();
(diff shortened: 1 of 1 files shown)


jfirebaugh:
Thanks for pointing this out! This constructor appears to be unused, so I'm just going to remove it entirely. https://github.com/mapbox/mapbox-gl-native/pull/8607

PR #8601, opened 2017-03-31
Refactor explicit operator bool usage for some cases

brunoabinader:


include/mbgl/util/geo.hpp
```diff
@@ -48,7 +48,7 @@ class LatLng {
         else if (longitude < 0 && end.longitude > 0) longitude += util::DEGREES_MAX;
     }
 
-    explicit operator bool() const {
+    bool valid() const {
```
brunoabinader: Sure, I'm going to defer these to #3802.

include/mbgl/util/size.hpp
```diff
@@ -17,7 +17,7 @@ class Size {
         return width * height;
     }
 
-    constexpr explicit operator bool() const {
+    constexpr bool valid() const {
```
brunoabinader: Deferring this to #8562.

brunoabinader:
@jfirebaugh let's wait for #3802 and #8562 to land, so I can rebase and address the review comments in one step.

include/mbgl/style/property_value.hpp
```diff
@@ -34,7 +34,7 @@ class PropertyValue {
     const                T & asConstant()       const { return value.template get<               T >(); }
     const CameraFunction<T>& asCameraFunction() const { return value.template get<CameraFunction<T>>(); }
 
-    explicit operator bool() const { return !isUndefined(); };
+    bool valid() const { return !isUndefined(); };
```
brunoabinader: This is actually not used, it seems `isUndefined()` suits the purpose already.

brunoabinader:
@jfirebaugh I've applied the requested changes based on your comments - thank you for the detailed explaining on these!

The only exception is `Shaping`, which starts failing some render tests once I replace the explicit operator bool usage with optional. I'm going to skip it for now and get back to it once I figure out exactly why in some cases the symbols are not being inserted into the buckets when the text shaping is not valid.

src/mbgl/text/glyph_pbf.hpp
```diff
@@ -21,7 +21,7 @@ class SDFGlyph {
     AlphaImage bitmap;
 
     // Glyph metrics
-    GlyphMetrics metrics;
+    optional<GlyphMetrics> metrics;
```
jfirebaugh: Once `Glyph` is optional, `GlyphMetrics::operator bool()` is unused, so we don't need to make this optional.

jfirebaugh: I have an alternate fix for glyph-related `operator bool()` prepared as part of #8747, so I'll omit the last commit here and merge.

brunoabinader: Thank you @jfirebaugh :+1:

PR #8608, opened 2017-03-31
[ios] Silence incompatible type warning for callout view

boundsj:
Continuation of https://github.com/mapbox/mapbox-gl-native/issues/8420.
Fixes warning introduced in https://github.com/mapbox/mapbox-gl-native/pull/8541

This silences a compatibility warning that was introduced in a previous refactor that changed the return type of the callout view for annotation delegate method. This also adds checks to ensure that the callout view provided by the delegate to the map view is of the correct type and conforms to the correct protocol.

### DIFF
--- a/platform/ios/CHANGELOG.md
+++ b/platform/ios/CHANGELOG.md
@@ -4,8 +4,6 @@ Mapbox welcomes participation and contributions from everyone. Please read [CONT
 
 ## 3.5.1
 
-### Other changes
-
 * Fixed an issue that caused the return type of a map view delegate method to bridge incorrectly to applications written in Swift. ([#8541](https://github.com/mapbox/mapbox-gl-native/pull/8541))
 
 ## 3.5.0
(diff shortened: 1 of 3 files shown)


platform/ios/src/MGLMapView.mm
```diff
@@ -3761,7 +3761,12 @@ - (void)selectAnnotation:(id <MGLAnnotation>)annotation animated:(BOOL)animated
         UIView <MGLCalloutView> *calloutView;
         if ([self.delegate respondsToSelector:@selector(mapView:calloutViewForAnnotation:)])
         {
-            calloutView = [self.delegate mapView:self calloutViewForAnnotation:annotation];
+            id providedCalloutView = [self.delegate mapView:self calloutViewForAnnotation:annotation];
+            if (providedCalloutView) {
+                NSAssert([providedCalloutView isKindOfClass:[UIView class]], @"callout view should be a kind of UIView");
```
1ec5: This should be an NSException, not an NSAssert. Now that the return value of `-mapView:calloutViewForAnnotation:` is no longer required to be a kind of UIView, returning something other than a UIView is a developer error, not an SDK error.

platform/ios/CHANGELOG.md
```diff
@@ -4,8 +4,6 @@ Mapbox welcomes participation and contributions from everyone. Please read [CONT
 
 ## 3.5.1
 
-### Other changes
```
1ec5: Thanks!

PR #8613, opened 2017-04-01
Fix edge case in composite function interpolation

anandthakker:
This fixes a bug where, for a zoom value greater than that of the highest zoom
stop, composite function interpolation would return `nan`. (Blocking a render
test over in https://github.com/mapbox/mapbox-gl-native/pull/8593)

### DIFF
--- a/test/style/function/composite_function.test.cpp
+++ b/test/style/function/composite_function.test.cpp
@@ -1,0 +1,46 @@
+#include <mbgl/test/util.hpp>
+#include <mbgl/test/stub_geometry_tile_feature.hpp>
+
+#include <mbgl/style/function/composite_function.hpp>
+
+using namespace mbgl;
+using namespace mbgl::style;
+
+using namespace std::string_literals;
+
+static StubGeometryTileFeature oneInteger {
+    PropertyMap {{ "property", uint64_t(1) }}
+};
+
+TEST(CompositeFunction, ZoomInterpolation) {
+    EXPECT_EQ(40.0f, CompositeFunction<float>("property", CompositeExponentialStops<float>({
+        {0.0f, {{uint64_t(1), 24.0f}}},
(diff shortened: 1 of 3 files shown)


test/style/function/composite_function.test.cpp
```diff
@@ -0,0 +1,46 @@
...
+
+TEST(CompositeFunction, ZoomInterpolation) {
+    EXPECT_EQ(40.0f, CompositeFunction<float>("property", CompositeExponentialStops<float>({
+        {0.0f, {{uint64_t(1), 24.0f}}},
+        {1.5f, {{uint64_t(1), 36.0f}}},
+        {3.0f, {{uint64_t(1), 48.0f}}}
+    }), 0.0f)
+    .evaluate(2.0f, oneInteger, -1.0f)) << "Should interpolate between stops";
+    
+    EXPECT_EQ(0.0, CompositeFunction<float>("property", CompositeExponentialStops<float>({
```
jfirebaugh: Shouldn't the expected result be `33.0`? Functions of all types are supposed to be flat before their first stop.

test/style/function/composite_function.test.cpp
```diff
@@ -0,0 +1,46 @@
...
+        {5.0f, {{uint64_t(1), 33.0f}}},
+        {10.0f, {{uint64_t(1), 66.0f}}}
+    }), 0.0f)
+    .evaluate(0.0f, oneInteger, -1.0f)) << "Should interpolate before the first stop";
+    
+    EXPECT_EQ(99.0, CompositeFunction<float>("property", CompositeExponentialStops<float>({
```
jfirebaugh: Similarly, I'd expect `66.0`.

anandthakker:
@jfirebaugh thanks for the review.  Updated to hold function constant outside the min,max stop range.

PR #8615, opened 2017-04-02
Fix broken images in macOS DDS guide

1ec5:
Shared images used by the data-driven styling guide between the iOS SDK and macOS SDK docsets to fix broken images in the macOS version.

Also removed `MGLUserTrackingMode` from the macOS jazzy table of contents to resolve a warning. `MGLUserTrackingMode` was removed from the macOS SDK in #8141.

/cc @friedbunny @captainbarbosa


1ec5:
If in the future we come up with macOS screenshots to replace the iOS SDK screenshots in this guide, we can unshare them by moving them back to the old location.

PR #8617, opened 2017-04-02
[ios, macos] Add `overlays` property to `MGLMapView.mm`

### LINKED ISSUE #4281: Add `overlays` property to `MGLMapView`

eimantas:
Implemented by filtering the `annotations` property for
annotations that conform to `MGLOverlay` protocol.

Fixes #4281.

### DIFF
--- a/platform/ios/src/MGLMapView.h
+++ b/platform/ios/src/MGLMapView.h
@@ -1135,9 +1135,9 @@ IB_DESIGNABLE
 
  The objects in this array must adopt the `MGLOverlay` protocol. If no
  overlays are associated with the map view, the value of this property is
- `nil`.
+ empty array.
  */
-@property (nonatomic, readonly, nullable) NS_ARRAY_OF(id <MGLOverlay>) *overlays;
+@property (nonatomic, readonly, nonnull) NS_ARRAY_OF(id <MGLOverlay>) *overlays;
 
 /**
  Adds a single overlay object to the map.
--- a/platform/ios/src/MGLMapView.h
+++ b/platform/ios/src/MGLMapView.h
@@ -1131,6 +1131,15 @@ IB_DESIGNABLE
 #pragma mark Overlaying the Map
 
 /**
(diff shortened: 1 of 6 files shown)


platform/ios/src/MGLMapView.h
```diff
@@ -1113,6 +1113,15 @@ IB_DESIGNABLE
 #pragma mark Overlaying the Map
 
 /**
+ The complete list of overlays associated with the receiver. (read-only)
+
+ The objects in this array must adopt the `MGLOverlay` protocol. If no
+ overlays are associated with the map view, the value of this property is
+ `nil`.
+ */
+@property (nonatomic, readonly, nullable) NS_ARRAY_OF(id <MGLOverlay>) *overlays;
```
1ec5: As of iOS 9, `MKMapView.annotations` and `MKMapView.overlays` have changed to being non-nullable, returning an empty array when the map contains no annotation or no overlay. This would be a good opportunity to follow suit. (Since making the property non-nullable would also change it from optional to non-optional in Swift, we whould need to add a blurb to both changelogs about this change.)

platform/ios/src/MGLMapView.mm
```diff
@@ -3433,6 +3433,22 @@ - (void)removeAnnotations:(NS_ARRAY_OF(id <MGLAnnotation>) *)annotations
     }
 }
 
+- (nullable NS_ARRAY_OF(id <MGLOverlay>) *)overlays
+{
+    if (self.annotations == nil) { return nil; }
+
+    NS_MUTABLE_ARRAY_OF(id <MGLOverlay>) *mutableOverlays = [NSMutableArray new];
```
1ec5: Nit: use `+[NSMutableArray array]` instead of `+[NSMutableArray new]`. In general, we avoid `+new` because it’s inconsistent with the normal init pattern.

eimantas:
Should I match [non]nullability to macOS SDK too?

PR #8630, opened 2017-04-04
[android] - derived source attribution

### LINKED ISSUE #2723: Derive attribution from sources in current style

tobrun:
Closes #2723


tobrun:
It seems we need some additional logic here, currently attribution for digital globe when using a satellite map shows up under `Improve this map` while on iOS it's placed above.

1ec5:
> It seems we need some additional logic here, currently attribution for digital globe when using a satellite map shows up under `Improve this map` while on iOS it's placed above. 

“Improve This Map” should appear before DigitalGlobe. ([This screenshot](https://github.com/mapbox/mapbox-gl-native/pull/5999#issuecomment-239640411) is outdated.)

ios

platform/android/MapboxGLAndroidSDK/src/main/res/values/arrays.xml
```diff
@@ -1,16 +0,0 @@
...
-        <item>&#169; Mapbox</item>
-        <item>&#169; OpenStreetMap</item>
-        <item>Improve this map</item>
-        <item>Telemetry Settings</item>
-    </array>
-    <!-- If editing this array update MapView.ATTRIBUTION_INDEX_IMPROVE_THIS_MAP -->
-    <array name="mapbox_attribution_links" formatted="false" translatable="false">
-        <item>https://www.mapbox.com/about/maps/</item>
-        <item>http://www.openstreetmap.org/about/</item>
-        <item>https://www.mapbox.com/map-feedback/#/%1$f/%2$f/%3$d</item>
```
1ec5: The iOS, macOS, and JS implementations automatically [append](https://github.com/mapbox/mapbox-gl-native/blob/62ea1f21858c69f6921c775ba7a3de201f0514d8/platform/darwin/src/MGLAttributionInfo.mm#L129) this hash to the feedback URL. That way, the Map Feedback tool starts out at the right location, which makes their feedback much more actionable. The feedback link has a `class` attribute of `mapbox-improve-map`.

platform/android/MapboxGLAndroidSDK/src/main/java/com/mapbox/mapboxsdk/maps/AttributionDialogManager.java
```diff
@@ -0,0 +1,130 @@
...
+  @Override
+  public void onClick(View view) {
+    rebuildAttributionMap();
+    showAttributionDialog();
+  }
+
+  private void rebuildAttributionMap() {
+    attributionMap = new LinkedHashMap<>();
+    for (Source source : mapboxMap.getSources()) {
+      parseAttribution(source.getAttribution());
```
tobrun: A map collection cannot contain duplicate keys so this solves the issues of duplicate entries. Is there a use-case that I'm a not supporting with not implementing `// remove any entries that are substrings of another entry`.?

platform/android/MapboxGLAndroidSDK/src/main/java/com/mapbox/mapboxsdk/maps/AttributionDialogManager.java
```diff
@@ -0,0 +1,130 @@
...
+    }
+    addTelemetryEntryToAttributionMap();
+  }
+
+  private void parseAttribution(String attributionSource) {
+    if (!TextUtils.isEmpty(attributionSource)) {
+      SpannableStringBuilder htmlBuilder = (SpannableStringBuilder) Html.fromHtml(attributionSource);
```
tobrun: They use [TagSoup](https://github.com/android/platform_external_tagsoup/blob/master/src/org/ccil/cowan/tagsoup/Parser.java) underneath. With current design there isn't any possibility of icon placement though the API support it with an ImageGetter https://developer.android.com/reference/android/text/Html.html#fromHtml(java.lang.String,%20int,%20android.text.Html.ImageGetter,%20android.text.Html.TagHandler).

platform/android/MapboxGLAndroidSDK/src/main/res/values/arrays.xml
tobrun: That got lost, will readd

platform/android/MapboxGLAndroidSDK/src/main/java/com/mapbox/mapboxsdk/maps/AttributionDialogManager.java
1ec5: That logic was originally implemented in GL JS in mapbox/mapbox-gl-js#2453. I guess it’s meant to avoid “© OpenStreetMap contributors © OpenStreetMap” or somesuch, not sure.

platform/android/MapboxGLAndroidSDK/src/main/java/com/mapbox/mapboxsdk/maps/AttributionDialogManager.java
1ec5: No external icon support is perfectly fine. I just wanted to double-check because the use of HTML in attribution has been problematic for GL JS and mapbox.js, mostly concerning XSS risk.

platform/android/MapboxGLAndroidSDK/src/main/java/com/mapbox/mapboxsdk/maps/AttributionDialogManager.java
1ec5: I think what you’ve got should be fine when it comes to the Mapbox default styles at least. None of the sources ultimately loaded by Mapbox Streets or Mapbox Satellite Streets have any links whose texts differ but overlap.

tobrun:
Fixed up @1ec5 comment about improving this map, did a cleanup and is ready for review.

platform/android/src/style/sources/source.cpp
```diff
@@ -43,6 +43,10 @@ namespace android {
         return jni::Make<jni::String>(env, source.getID());
     }
 
+    jni::String Source::getAttribution(jni::JNIEnv& env) {
+        return source.getAttribution() ? jni::Make<jni::String>(env, source.getAttribution().value()) : jni::Make<jni::String>(env,"");
```
ivovandongen: Minor thing, but maybe invoke `source.getAttribution()` only once. Now it doesn't really do much, but if the implementation gets more complex it's a bit wasteful.

platform/android/MapboxGLAndroidSDK/src/main/java/com/mapbox/mapboxsdk/maps/AttributionDialogManager.java
```diff
@@ -0,0 +1,161 @@
...
+import com.mapbox.mapboxsdk.R;
+import com.mapbox.mapboxsdk.camera.CameraPosition;
+import com.mapbox.mapboxsdk.style.sources.Source;
+import com.mapbox.services.android.telemetry.MapboxTelemetry;
+
+import java.util.HashMap;
+import java.util.LinkedHashMap;
+import java.util.Locale;
+
+class AttributionDialogManager implements View.OnClickListener, DialogInterface.OnClickListener {
```
ivovandongen: Nit; some class level javadoc?

tobrun: done in https://github.com/mapbox/mapbox-gl-native/pull/8630/commits/527ffbe733979f393dafd23a062d895e524ea294#diff-8a5300c6fbc51b88f2e971461adcb513R33

platform/android/src/style/sources/source.cpp
tobrun: done in https://github.com/mapbox/mapbox-gl-native/pull/8630/commits/527ffbe733979f393dafd23a062d895e524ea294#diff-8a5300c6fbc51b88f2e971461adcb513R33. re-review that part of the code?

platform/android/MapboxGLAndroidSDK/src/main/java/com/mapbox/mapboxsdk/maps/AttributionDialogManager.java
```diff
@@ -0,0 +1,169 @@
...
+ * Responsible for managing attribution interactions on the map.
+ * <p>
+ * When the user clicks the attribution icon, {@link AttributionDialogManager#onClick(View)} will be invoked.
+ * An attribution dialog will be shown to the user with contents based on the attributions found in the map style.
+ * Additionally an telemetry option item is shown to configure telemetry settings.
+ * </p>
+ */
+class AttributionDialogManager implements View.OnClickListener, DialogInterface.OnClickListener {
+
+  private static final String MAP_FEEDBACK_URL = "https://www.mapbox.com/map-feedback";
```
1ec5: We shouldn’t hard-code this URL because sources other than Mapbox Streets are free to provide their own map feedback URLs. Look for `<a>` tags with `class` set to `mapbox-improve-map`.

1ec5: I ticketed this out as tail work in #8730, since custom feedback links are admittedly quite an edge case.

tobrun: Thank you for ticketing that out, will fix it separately from this PR.

PR #8633, opened 2017-04-04
[darwin] Remove unnecessary MGLTileSource initializers

jfirebaugh:
Both `-initWithIdentifier:configurationURL:` and `-initWithIdentifier:tileURLTemplates:options:` threw away all but the first argument and called the superclass `-initWithIdentifier`. Working versions of these initializers are provided by the subclasses `MGLRasterSource` and `MGLVectorSource`, so there's no need to define stubs in the base class.


platform/darwin/src/MGLRasterSource.h
```diff
@@ -104,6 +104,109 @@ MGL_EXPORT
...
 - (instancetype)initWithIdentifier:(NSString *)identifier configurationURL:(NSURL *)configurationURL tileSize:(CGFloat)tileSize NS_DESIGNATED_INITIALIZER;
 
+/**
+ Returns a raster source initialized an identifier, tile URL templates, and
+ options.
+
+ After initializing and configuring the source, add it to a map view’s style
+ using the `-[MGLStyle addSource:]` method.
+
+ #### Tile URL templates
```
1ec5: Redeclaring `-initWithIdentifier:tileURLTemplates:options:` in the abstract superclass gave us a common place to put this substantial amount of documentation. It’s important for the developer to know that the identically or similarly named initializers in the subclasses behave similarly – or even to think of MGLRasterSource and MGLVectorSource as sharing the same initializers – even if they don’t share the same implementation in mbgl under the hood (but the developer doesn’t need to know that).

If we need to eliminate the superclass method, we should at least factor this documentation out into a jazzy guide and link to it from each identical method.

jfirebaugh: There's currently [no indication](https://www.mapbox.com/ios-sdk/api/3.5.0/Classes/MGLVectorSource.html) in the documentation of how to initialize `MGLVectorSource`. I think this is an improvement, despite the duplication.

1ec5: There’s the line `@interface MGLVectorSource : MGLTileSource` at the top, which links to the documentation for [MGLTileSource](https://www.mapbox.com/ios-sdk/api/3.5.0/Classes/MGLTileSource.html). However, that page says “Do not create instances of this class directly”, which the developer may misinterpret to mean “don’t use these methods at all”. This is a shortcoming of our documentation for abstract classes in general (and of some of Apple’s own abstract class documentation).

1ec5: Tracking as tail work in #8634 and more broadly in #8635.

platform/darwin/src/MGLRasterSource.mm
```diff
@@ -37,76 +33,32 @@ - (instancetype)initWithIdentifier:(NSString *)identifier configurationURL:(NSUR
...
-                                                                  configurationURL.mgl_URLByStandardizingScheme.absoluteString.UTF8String,
-                                                                  uint16_t(round(tileSize)));
-        _pendingSource = std::move(source);
-        self.rawSource = _pendingSource.get();
-    }
-    return self;
+    auto source = std::make_unique<mbgl::style::RasterSource>(identifier.UTF8String,
+                                                              configurationURL.mgl_URLByStandardizingScheme.absoluteString.UTF8String,
+                                                              uint16_t(round(tileSize)));
+    return [super initWithPendingSource:std::move(source)];
```
1ec5: Initializers should always follow [the pattern](https://developer.apple.com/library/content/documentation/General/Conceptual/CocoaEncyclopedia/Initialization/Initialization.html#//apple_ref/doc/uid/TP40010810-CH6-SW58) `if (self = [super …]) { … } return self;`, even if it looks redundant.

platform/darwin/src/MGLSource_Private.h
```diff
@@ -13,11 +15,18 @@ namespace mbgl {
...
+ Initializes and returns a source with a raw pointer to the backing store,
+ associated with a style.
  */
 - (instancetype)initWithRawSource:(mbgl::style::Source *)rawSource;
 
 /**
+ Initializes and returns a source with an owning pointer to the backing store,
+ unassociated from a style.
+ */
+- (instancetype)initWithPendingSource:(std::unique_ptr<mbgl::style::Source>)pendingSource;
```
1ec5: This method doesn’t appear to be implemented in MGLSource.mm. Calling it will result in a crash.

platform/darwin/src/MGLRasterSource.mm
jfirebaugh: Even if the body of the `if` is empty? What's the reason for that?

1ec5: If you wanted to, you could say `return self = [super …]`, but the `if` is a convention that makes it easier to add and initialize ivars later on. The important part is assigning `self` and returning it; otherwise, ARC is free to assume that `self` is no longer in use and can release the object. (Along these lines, if you’re partway through the `if` block and discover that the object can’t be initialized for some reason, setting `self` to `nil` has the same effect of releasing the object.)

jfirebaugh:
Rolling this into #8626.

PR #8637, opened 2017-04-05
Harden std::map usage in MGLMapVIew

1ec5:
Replaced `std::map::operator[]` usage with `std::map::at()`. Backed up an assertion on iOS with a guard when `-annotationTagsInRect:` returns the tag of a nonexistent annotation, for consistency with macOS. Removed an unnecessary and risky subscript into `_annotationContextsByAnnotationTag` in `-positioningRectForCalloutForAnnotationWithTag:` on iOS.

/ref https://github.com/mapbox/mapbox-gl-native/issues/8163#issuecomment-291736586

### DIFF
--- a/platform/ios/src/MGLMapView.mm
+++ b/platform/ios/src/MGLMapView.mm
@@ -1452,7 +1452,7 @@ public:
         }
         else
         {
-            nextElement = _annotationContextsByAnnotationTag[_selectedAnnotationTag].accessibilityElement;
+            nextElement = _annotationContextsByAnnotationTag.at(_selectedAnnotationTag).accessibilityElement;
         }
         [self deselectAnnotation:self.selectedAnnotation animated:YES];
         UIAccessibilityPostNotification(UIAccessibilityScreenChangedNotification, nextElement);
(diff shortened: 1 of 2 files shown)


platform/ios/src/MGLMapView.mm
```diff
@@ -3550,8 +3550,12 @@ - (MGLAnnotationTag)annotationTagAtPoint:(CGPoint)point persistingResults:(BOOL)
         {
             id <MGLAnnotation> annotation = [self annotationWithTag:annotationTag];
             NSAssert(annotation, @"Unknown annotation found nearby tap");
```
boundsj: Just noting that this assertion will fail in a debug build if you perform the reproduction steps in https://github.com/mapbox/mapbox-gl-native/issues/8163#issuecomment-291695262. The issue is that mbgl notion of visible annotations will always lag a bit behind the platforms' notion of the same when annotations are removed.

1ec5: Should we consider that a bug, or is it an edge case we want to fail gracefully around from now on?

boundsj: I think it is a bug that is an edge case 😄  
But, the fix would require a callback feature for at least some of the mbgl APIs right? The platforms should not actually update their state until _after_ core has completed updating its state (i.e. removing annotations so that queries of visible annotations won't include them)

1ec5: I agree, and we should make sure the map remains usable even in debug builds. That could be tail work, though; it seems like #8513 would have the same considerations.

PR #8640, opened 2017-04-05
Preserve symlinks when zipping framework

### LINKED ISSUE #8636: Framework contains duplicate files instead of symlinks
As of macOS SDK v0.4.0, Mapbox.framework contains duplicate copies of files such as the Mapbox executable. Instead of symlinks to files and folders under Versions/A/, the root of the framework and the Versions/Current/ folder are polluted with copies of these files. As a result, the framework is bloated, fails to code-sign, and probably fails in other ways.

This is a regression of #5818, which was fixed in #5850 for v0.2.0.

/cc @kkaefer @frederoni

1ec5:
When zipping the framework for deployment, preserve symlinks instead of resolving them. Prior to this change, the published framework bundle contained multiple copies of the Mapbox executable and of every resource file, because the standard layout for a framework on macOS requires symlinks from the root folder and Versions/Current/ to Versions/A/.

Fixes #8636.

/cc @kkaefer @frederoni

### DIFF
--- a/platform/ios/scripts/publish.sh
+++ b/platform/ios/scripts/publish.sh
@@ -27,7 +27,7 @@ cd build/ios/pkg
 ZIP=mapbox-ios-sdk-${PUBLISH_VERSION}${PUBLISH_STYLE}.zip
 step "Compressing ${ZIP}…"
 rm -f ../${ZIP}
-zip -r ../${ZIP} *
+zip -yr ../${ZIP} *
 
 #
 # upload
--- a/platform/macos/CHANGELOG.md
+++ b/platform/macos/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Changelog for Mapbox macOS SDK
 
+## 0.4.1
+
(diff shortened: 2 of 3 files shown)


platform/ios/scripts/publish.sh
```diff
@@ -27,7 +27,7 @@ cd build/ios/pkg
 ZIP=mapbox-ios-sdk-${PUBLISH_VERSION}${PUBLISH_STYLE}.zip
 step "Compressing ${ZIP}…"
 rm -f ../${ZIP}
-zip -r ../${ZIP} *
+zip -yr ../${ZIP} *
```
kkaefer: Why is there no changelog entry for iOS?

1ec5: There are no symlinks in the iOS SDK, so there’s no developer impact in practice. I made this change to the iOS SDK regardless in case we ever do need a symlink in there.

PR #8651, opened 2017-04-05
[ios] Replaced link to select a feature example

jmkiley:
Replaces link with a different example of how to add a tap gesture.


boundsj:
For PRs that target the release branch, please set a release Milestone so we can properly schedule this work for a 3.5.x patch release. I just added this PR to v3.5.1

cc @friedbunny

PR #8652, opened 2017-04-05
[glfw] Include the GL header before GLFW header

brunoabinader:
Fixes the GLFW build for Linux in newer distros e.g. Ubuntu 16.04.


brunoabinader:
Closed as per conversation with @kkaefer we'll be using #pragma to tell GCC to ignore the macro redefinition error.

PR #8657, opened 2017-04-06
[ios, macos] edited identity interpolation mode documentation

jmkiley:
Edited the documentation for identity interpolation mode to clarify its use.

### DIFF
--- a/platform/darwin/src/MGLStyleValue.h
+++ b/platform/darwin/src/MGLStyleValue.h
@@ -75,9 +75,9 @@ typedef NS_ENUM(NSUInteger, MGLInterpolationMode) {
      */
     MGLInterpolationModeCategorical,
     /**
-     Values between two stops are not interpolated. Instead, values are set to their
-     input value. Use identity interpolation mode to show attribute values that can be
-     used as style values.
+     Values between two stops are not interpolated. Instead, style values are set
+     to the attribute value for a feature. Use identity interpolation mode to show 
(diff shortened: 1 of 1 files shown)


platform/darwin/src/MGLStyleValue.h
```diff
@@ -75,9 +75,9 @@ typedef NS_ENUM(NSUInteger, MGLInterpolationMode) {
      */
     MGLInterpolationModeCategorical,
     /**
-     Values between two stops are not interpolated. Instead, values are set to their
-     input value. Use identity interpolation mode to show attribute values that can be
-     used as style values.
+     Values between two stops are not interpolated. Instead, style values are set
+     to the attribute value for a feature. Use identity interpolation mode to show
```
boundsj: maybe "... to **an** attribute value ..."

boundsj:
Don't forget to retarget to `release-ios-v3.6.0-android-v5.1.0` or cherry pick to that branch if this lands in `master` first.

platform/darwin/src/MGLStyleValue.h
1ec5: Maybe we should say “match”, “correspond to”, “determined by” instead of “set to”, so that the developer doesn’t misinterpret this sentence to mean that they have to set a value explicitly?

> Instead, for any given feature, the style value matches a value in that feature’s `attributes` dictionary.

PR #8669, opened 2017-04-06
[core] Inline GlyphPBF into GlyphAtlas

jfirebaugh:
First of several followup refactors to #8341: the split between `GlyphAtlas` and `GlyphPBF` was awkward. Inline `GlyphPBF`, while leaving the actual PBF parsing code independent and tested in isolation.

### DIFF
--- a/src/mbgl/text/glyph_atlas.cpp
+++ b/src/mbgl/text/glyph_atlas.cpp
@@ -4,6 +4,9 @@
 #include <mbgl/gl/context.hpp>
 #include <mbgl/util/logging.hpp>
 #include <mbgl/util/platform.hpp>
+#include <mbgl/storage/file_source.hpp>
+#include <mbgl/storage/resource.hpp>
+#include <mbgl/storage/response.hpp>
 
 #include <cassert>
 #include <algorithm>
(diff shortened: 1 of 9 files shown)


src/mbgl/text/glyph_atlas.cpp
```diff
@@ -87,34 +79,56 @@ void GlyphAtlas::getGlyphs(GlyphRequestor& requestor, GlyphDependencies glyphDep
...
 }
 
-void GlyphAtlas::setObserver(GlyphAtlasObserver* observer_) {
-    observer = observer_;
-}
-    
-void GlyphAtlas::onGlyphsError(const FontStack& fontStack, const GlyphRange& range, std::exception_ptr err) {
-    if (observer) {
-        observer->onGlyphsError(fontStack, range, err);
+GlyphAtlas::GlyphRequest& GlyphAtlas::requestRange(Entry& entry, const FontStack& fontStack, const GlyphRange& range) {
```
ChrisLoer: Why does `requestRange` return the `GlyphAtlas::GlyphRequest&`? Doesn't seem like we're using it anywhere...

src/mbgl/text/glyph_atlas.hpp
```diff
@@ -96,16 +78,33 @@ class GlyphAtlas : public util::noncopyable, public GlyphAtlasObserver {
...
         std::map<uint32_t, GlyphValue> glyphValues;
     };
 
     std::unordered_map<FontStack, Entry, FontStackHash> entries;
-    
-    void removeGlyphValues(GlyphRequestor& requestor, std::map<uint32_t, GlyphValue>& face);
-    void removePendingRanges(GlyphRequestor& requestor, std::map<GlyphRange, GlyphPBF>& ranges);
+
+    // Only used by GlyphAtlasTest
+    friend class ::GlyphAtlasTest;
```
ChrisLoer: While we're moving these, we should probably group `getGlyphSet` here too, it's only used by `GlyphAtlasTest`

src/mbgl/text/glyph_pbf.cpp
```diff
@@ -94,12 +92,12 @@ void parseGlyphPBF(GlyphSet& glyphSet, const GlyphRange& glyphRange, const std::
                 glyph.bitmap = AlphaImage(size, reinterpret_cast<const uint8_t*>(glyphData.data()), glyphData.size());
             }
 
-            glyphSet.insert(glyph.id, std::move(glyph));
+            result.push_back(std::move(glyph));
```
ChrisLoer: I doubt that this is a very hot path, but in the rest of the code we try to avoid unnecessarily moving around the glyph bitmap, and growing the vector will cause a few copies. Is it worth trying to avoid that with a `reserve` (we can guess that 256 is a good size for the vector)?

src/mbgl/text/glyph_atlas.cpp
jfirebaugh: It was used, but in a non-obvious way. I changed it to be more obvious.

src/mbgl/text/glyph_atlas.hpp
jfirebaugh: This doesn't quite work, as it's used not by `GlyphAtlasTest` itself but by individual tests. We should clean it up, but I'll leave it for another followup.

PR #8684, opened 2017-04-07
Update changelog, podspecs for macOS SDK v0.4.1

1ec5:
Updated the changelog and podspecs for macOS SDK v0.4.1 and cherry-picked #8692 to keep the Qt5 macOS build passing.

/cc @frederoni @boundsj


1ec5:
The Qt5 macOS build failure is being tracked in #8691.

PR #8685, opened 2017-04-07
WIP - Log out shader variable precision

### LINKED ISSUE #8533: Add check for 16-bit `highp` value
Since #8183 (and the followup in #8496), we are packing pairs of colors into a single `highp vec4`. This strategy means that we have a hard requirement for `highp` providing at least 16 bits of precision.

It's possible that [this color bug](https://github.com/mapbox/mapbox-gl-native/issues/8385#issuecomment-289418069) _could_ be caused by a device that doesn't provide 16-bit `highp`s.

We can add a check at startup, using [`glGetShaderPrecisionFormat`](https://www.khronos.org/registry/OpenGL-Refpages/es2.0/xhtml/glGetShaderPrecisionFormat.xml), so that if a device doesn't have enough precision, we fail with an error message instead of trying to proceed.

anandthakker:
Closes #8533

### DIFF
--- a/src/mbgl/gl/context.cpp
+++ b/src/mbgl/gl/context.cpp
@@ -90,6 +90,28 @@ void Context::enableDebugging() {
     MBGL_CHECK_ERROR(debugging->debugMessageCallback(extension::Debugging::DebugCallback, nullptr));
 }
 
+void Context::logGLDiagnostics() {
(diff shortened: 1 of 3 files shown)


jfirebaugh:
Stale; probably not needed after all.

Issue #8674, opened 2017-04-07
User location annotation callout crash in v3.5.1


lopemac:
I'm experiencing an immediate crash when displaying the current location circle label. If the map moves automatically or is moved or tapped in any way while the "You Are Here" default label is showing, the app immediately crashes. My app is using the latest Xcode and Mapbox SDK. I did not experience this behavior with SDK v3.5. When I downgrade back to v3.5, no crash occurs and the map can be moved as normal. I'm displaying the map in a rounded square view and am using these settings in the code under viewDidLoad:

```
mapinView.tintColor = .darkGray
mapinView.delegate = self
view.addSubview(mapinView)
self.view.addSubview(resetShadow) //used to show location button on top of map
mapinView.layer.cornerRadius = 20
```

I'm using "mapbox://styles/mapbox/light-v9" in the Style URL settings box. I'm also using the clustering code from the mapbox example page which shows labels without any issues or crashing.

Platform: iOS Swift 3.1 Xcode 8.3.1 running on iPhone 6 10.3.1 and iPhone 5 10.3
Mapbox SDK version: v3.5.1

Error Code from Xcode: 
```
libc++abi.dylib: terminating with uncaught exception of type std::out_of_range: unordered_map::at: key not found
(lldb)
```

### Steps to trigger behavior

 1. Zoom into current location or scroll to show current location
 2. Press on my location circle to get default pop up label that says "You Are Here"
 3. Move map around or tap anywhere while label is still showing

### Expected behavior
It should let you scroll map or tap anywhere without any abnormal behavior

### Actual behavior
It immediately crashes following the steps above.

![simulator screen shot apr 6 2017 6 50 01 pm](https://cloud.githubusercontent.com/assets/18113616/24786203/41438f38-1afb-11e7-8c42-542e4c3fb12e.png)

friedbunny:
Thanks for the report and clear reproduction steps, @lopemac. Here’s the error message associated with this crash:

```
libc++abi.dylib: terminating with uncaught exception of type std::out_of_range: unordered_map::at: key not found
```

This would seem to implicate https://github.com/mapbox/mapbox-gl-native/pull/8637, but there were a few other annotation-related changes in v3.5.1.

/cc @1ec5 @boundsj @fabian-guerra

friedbunny:
@1ec5 This crash did bisect to https://github.com/mapbox/mapbox-gl-native/pull/8637 for me.

1ec5:
[...]

PR #8686, opened 2017-04-07
[ios] Guard against looking up annotation contexts MGLAnnotationTagNotFound

boundsj:
Fixes https://github.com/mapbox/mapbox-gl-native/issues/8674

In https://github.com/mapbox/mapbox-gl-native/pull/8637 we stopped indexing into std::map to avoid silently creating bogus entries in collections like `_annotationContextsByAnnotationTag`. However, since the user annotation is effectively always represented with `MGLAnnotationTagNotFound` (that equals `UINT32_MAX`) and the user annotation is processed similarly to regular annotations in some cases, we opened the door to a crash.

Specifically, with #8637, attempts to lookup an annotation context for the user annotation's tag (`MGLAnnotationTagNotFound`) in order to update the callout view triggered the map's ` std::out_of_range` exception.

This PR guards against this by first checking to see if the annotation tag involved in the context lookup for the callout view update is `MGLAnnotationTagNotFound` or not. It also adds several more similar defensive checks in other areas that attempt to obtain a context.

Methods that are now guarded:

* `handleSingleTapGesture:`
* `updateCalloutView:`
* `observeValueForKeyPath:` (for coordinate)
* `visibleAnnotationsInRect:`
* `selectedAnnotation:`

Methods that already had this check in place:

* `updateCalloutView:`
* `viewForAnnotation:`
* `removeAnnotations:`
* `imageOfAnnotationWithTag:`

cc @friedbunny @1ec5 @fabian-guerra

### DIFF
--- a/platform/ios/CHANGELOG.md
+++ b/platform/ios/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 Mapbox welcomes participation and contributions from everyone. Please read [CONTRIBUTING.md](../../CONTRIBUTING.md) to get started.
 
+## 3.5.2
+
+* Fixed an issue that caused a crash when the user annotation was presenting a callout view and the map was moved. ([#8686](https://github.com/mapbox/mapbox-gl-native/pull/8686))
+
 ## 3.5.1 - April 5, 2017
 
 * Fixed an issue that caused the return type of a map view delegate method to bridge incorrectly to applications written in Swift. ([#8541](https://github.com/mapbox/mapbox-gl-native/pull/8541))
--- a/platform/ios/CHANGELOG.md
+++ b/platform/ios/CHANGELOG.md
@@ -4,7 +4,7 @@ Mapbox welcomes participation and contributions from everyone. Please read [CONT
 
 ## 3.5.2
 
-* Fixed an issue that caused a crash when the user annotation was presenting a callout view and the map was moved. ([#8686](https://github.com/mapbox/mapbox-gl-native/pull/8686))
+* Fixed an issue that caused a crash when the user location annotation was presenting a callout view and the map was moved. ([#8686](https://github.com/mapbox/mapbox-gl-native/pull/8686))
 
 ## 3.5.1 - April 5, 2017
 
--- a/platform/ios/src/MGLMapView.mm
+++ b/platform/ios/src/MGLMapView.mm
@@ -3555,10 +3555,6 @@ public:
         auto end = std::remove_if(nearbyAnnotations.begin(), nearbyAnnotations.end(),
                                   [&](const MGLAnnotationTag annotationTag)
         {
-            if (annotationTag == MGLAnnotationTagNotFound) {
-                return true;
-            }
-
             id <MGLAnnotation> annotation = [self annotationWithTag:annotationTag];
             NSAssert(annotation, @"Unknown annotation found nearby tap");
             if ( ! annotation)
--- a/platform/ios/src/MGLMapView.mm
+++ b/platform/ios/src/MGLMapView.mm
@@ -1452,7 +1452,9 @@ public:
         }
         else
         {
-            nextElement = _annotationContextsByAnnotationTag.at(_selectedAnnotationTag).accessibilityElement;
+            if (_selectedAnnotationTag != MGLAnnotationTagNotFound) {
+                nextElement = _annotationContextsByAnnotationTag.at(_selectedAnnotationTag).accessibilityElement;
+            }
         }
         [self deselectAnnotation:self.selectedAnnotation animated:YES];
         UIAccessibilityPostNotification(UIAccessibilityScreenChangedNotification, nextElement);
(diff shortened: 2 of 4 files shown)


boundsj:
Will update this to include macOS and changelog entries next cc @1ec5

boundsj:
For macOS, I added the guard to:

* `visibleAnnotationsInRect:`
* `annotationWithTag:`

These methods already had the guard:

* `selectedAnnotation:`
* `imageOfAnnotationWithTag:`

platform/ios/CHANGELOG.md
```diff
@@ -2,6 +2,10 @@
 
 Mapbox welcomes participation and contributions from everyone. Please read [CONTRIBUTING.md](../../CONTRIBUTING.md) to get started.
 
+## 3.5.2
+
+* Fixed an issue that caused a crash when the user annotation was presenting a callout view and the map was moved. ([#8686](https://github.com/mapbox/mapbox-gl-native/pull/8686))
```
friedbunny: I think we tend to call it the “user _location_ annotation”, to make clear that it’s not some annotation otherwise belonging to the user.

boundsj: I made that change for the iOS changelog in https://github.com/mapbox/mapbox-gl-native/pull/8686/commits/99eee0af084c5172c6c751e742e540b394061fb4. In that same commit, I also made the macOS changelog less specific since the possibility of a crash there is more theoretical since there is currently no user location annotation.

platform/ios/src/MGLMapView.mm
```diff
@@ -3077,8 +3082,8 @@ - (void)removeStyleClass:(NSString *)styleClass
 /// Returns the annotation assigned the given tag. Cheap.
 - (id <MGLAnnotation>)annotationWithTag:(MGLAnnotationTag)tag
 {
-    if ( ! _annotationContextsByAnnotationTag.count(tag))
-    {
+    if ( ! _annotationContextsByAnnotationTag.count(tag) ||
+        tag == MGLAnnotationTagNotFound) {
```
1ec5: Should this return `userLocation`?

1ec5: Never mind. The code used to overload MGLAnnotationTagNotFound to mean the user location annotation tag. But in retrospect that's a pretty risky practice. This change moves us further away from that practice, forcing us to explicitly account for the user location annotation explicitly, which is a good thing.

platform/ios/src/MGLMapView.mm
```diff
@@ -3550,6 +3555,10 @@ - (MGLAnnotationTag)annotationTagAtPoint:(CGPoint)point persistingResults:(BOOL)
         auto end = std::remove_if(nearbyAnnotations.begin(), nearbyAnnotations.end(),
                                   [&](const MGLAnnotationTag annotationTag)
         {
+            if (annotationTag == MGLAnnotationTagNotFound) {
```
1ec5: This is redundant to the check in -annotationWithTag:. Should we put an assertion in right above for this case? Is it legitimate for the user location annotation to be a nearby annotation?

1ec5: Going to answer my own question here: `-annotationTagsInRect:` wouldn’t know about the user location annotation, so it should never return `MGLAnnotationTagNotFound`. I still think we should either rely on the assertion below or assert here about `MGLAnnotationTagNotFound`, in case some sort of internal consistency leads it to show up in `nearbyAnnotations`.

platform/ios/src/MGLMapView.mm
```diff
@@ -2002,19 +2004,21 @@ - (void)observeValueForKeyPath:(NSString *)keyPath ofObject:(id)object change:(N
...
+                NSString *symbolName = annotationImage.styleIconIdentifier;
 
-            // Update the annotation’s backing geometry to match the annotation model object. Any associated annotation view is also moved by side effect. However, -updateAnnotationViews disables the view’s animation actions, because it can’t distinguish between moves due to the viewport changing and moves due to the annotation’s coordinate changing.
-            _mbglMap->updateAnnotation(annotationTag, mbgl::SymbolAnnotation { point, symbolName.UTF8String });
-            [self updateCalloutView];
+                // Update the annotation’s backing geometry to match the annotation model object. Any associated annotation view is also moved by side effect. However, -updateAnnotationViews disables the view’s animation actions, because it can’t distinguish between moves due to the viewport changing and moves due to the annotation’s coordinate changing.
+                _mbglMap->updateAnnotation(annotationTag, mbgl::SymbolAnnotation { point, symbolName.UTF8String });
+                [self updateCalloutView];
```
1ec5: This change is correct with respect to the user location annotation, because we rely on Core Location, not KVO, to tell us when the model object has changed.

platform/ios/src/MGLMapView.mm
boundsj: 👍

platform/ios/src/MGLMapView.mm
boundsj: That makes sense. I'll remove this check to rely on the assertion below.

boundsj: Also, this is the assertion we hit in debug builds if annotations are removed at almost the same time and just barely before the user tries to select them (part of the repro from https://github.com/mapbox/mapbox-gl-native/issues/8163#issuecomment-291695262). 

The check below recently added in https://github.com/mapbox/mapbox-gl-native/pull/8637 to make sure the annotation is not nil protects release builds from crashing at the call to `std::map::at` also just below.

PR #8696, opened 2017-04-10
Run Firebase on Circle CI + publish binary metrics on Circle CI

kkaefer:
This allows us to remove Android from the Bitrise builds.


kkaefer:
Note that we can't remove/disable the Bitrise Android tests until we've switched them over, otherwise Github will prevent merging ;)

PR #8692, opened 2017-04-08
Rename qt5 formula to qt

1ec5:
Updated the Qt5 macOS Bitrise bot’s configuration to reflect the `qt5` formula’s rename to `qt` in Homebrew/homebrew-core@b92c519bbd3cb869b4861017210e07fc42a0cc5f.

Fixes #8691.

PR #8698, opened 2017-04-10
Avoid hardcoding Qt cellar path

1ec5:
This change avoids hard-coding the path to the Homebrew cellar for the `qt` formula, per https://github.com/mapbox/mapbox-gl-native/pull/8692#issuecomment-292762518.

/cc @friedbunny @brunoabinader

### DIFF
--- a/platform/qt/bitrise-qt5.yml
+++ b/platform/qt/bitrise-qt5.yml
@@ -19,9 +19,10 @@ workflows:
             brew install qt
             brew link qt --force
             brew linkapps qt
+            export HOMEBREW_QT5_CELLAR=$(brew --cellar qt)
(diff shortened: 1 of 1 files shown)


platform/qt/bitrise-qt5.yml
```diff
@@ -19,9 +19,10 @@ workflows:
             brew install qt
             brew link qt --force
             brew linkapps qt
+            export HOMEBREW_QT5_CELLAR=$(brew --cellar qt5)
```
friedbunny: I suppose we may as well use `qt` in the command here, too, for consistency.

1ec5: Thanks, fixed in 5a467692cfe1ac9462a0b508881048640c5ec4c8.

PR #8709, opened 2017-04-11
[android] - expose source layer identifier

### LINKED ISSUE #8663: Expose sourceLayerIdentifier on Layer
A useful addition to our current Layer API is exposing `sourceLayerIdentfier`, while we use it internally we don't expose a getter for it. Such a getter can be useful to filter layers. For example take traffic, you want to add a traffic layer just above all other road layers. For this you would need to get all the style layers and filter out the ones with `sourceLayerIdentifier="road"`.

tobrun:
Closes #8663, need to fixup some whitespacing in the `.ejs` template file before reviewing.
This PR add a getter for source layer identifier on the layer classes that support it. Exposing this allows to iterate all layers and check for a specfic source layer identifiers, eg. `road`.

### DIFF
--- a/platform/android/MapboxGLAndroidSDK/src/main/java/com/mapbox/mapboxsdk/style/layers/layer.java.ejs
+++ b/platform/android/MapboxGLAndroidSDK/src/main/java/com/mapbox/mapboxsdk/style/layers/layer.java.ejs
@@ -80,6 +80,15 @@ public class <%- camelize(type) %>Layer extends Layer {
 <% } -%>
 <% if (type !== 'background' && type !== 'raster') { -%>
   /**
+   * Get the source layer.
+   *
+   * @return sourceLayer the source layer to get
+   */
+  public String getSourceLayer() {
+    return nativeGetSourceLayer();
+  }
+
+  /**
    * Set a single filter.
    *
    * @param filter the filter to set
--- a/platform/android/src/style/layers/layer.cpp
+++ b/platform/android/src/style/layers/layer.cpp
@@ -144,6 +144,25 @@ namespace android {
         }
     }
 
+    jni::String Layer::getSourceLayer(jni::JNIEnv& env) {
+        using namespace mbgl::style;
+
+        std::string sourceLayerId;
+        if (layer.is<FillLayer>()) {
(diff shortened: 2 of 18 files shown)


tobrun:
Whitespacing has been adressed + fixed a couple of nit's in the existing code. 
@ivovandongen want tor review?

device-2017-04-13-143701

platform/android/MapboxGLAndroidSDK/src/main/java/com/mapbox/mapboxsdk/style/layers/layer.java.ejs
```diff
@@ -99,7 +108,7 @@ public class <%- camelize(type) %>Layer extends Layer {
     return this;
   }
 
-<% } -%>
+<% }  -%>
```
ivovandongen: nit: probably accidental?

platform/android/src/style/layers/layer.cpp
```diff
@@ -144,6 +144,25 @@ namespace android {
         }
     }
 
+    jni::String Layer::getSourceLayer(jni::JNIEnv& env) {
```
ivovandongen: I set a bad example here (my apologies), but a more extensible way of doing this would be something like:

```
struct Evaluator {
    std::string operator()(mbgl::style::BackgroundLayer&) { return {}; }
    std::string operator()(mbgl::style::RasterLayer&) { return {}; }
    std::string operator()(mbgl::style::CustomLayer&) { return {}; }

    template <class VectorLayer>
    std::string operator()(VectorLayer& layer) {
        return layer.getSourceLayer();
    }
};

...

        std::string sourceLayerId = layer.accept(Evaluator());
```

maybe it's better though to clean this up separately in a pr and do the others in this class at the same time (might be quite some overlap).

tobrun: Follow up ticket in #8758

tobrun:
- Review [comment](https://github.com/mapbox/mapbox-gl-native/pull/8709#discussion_r111434281) addressed 
- Follow up for https://github.com/mapbox/mapbox-gl-native/pull/8709#discussion_r111436935  in #8758. 
- Changelog updated 
- Rebased with head of master

Once CI approves, going to merge

Issue #8639, opened 2017-04-05
Android: crash when using library that also uses Lost for location


chenguo:
I'm writing an application that uses MapView. I'm using a library that also uses the Lost API under the hood, and the location updates being requested of that library is crashing MyLocationView:

```
android.view.ViewRootImpl$CalledFromWrongThreadException: Only the original thread that created a view hierarchy can touch its views.
    at android.view.ViewRootImpl.checkThread(ViewRootImpl.java:6902)
    at android.view.ViewRootImpl.invalidateChildInParent(ViewRootImpl.java:1087)
    at android.view.ViewGroup.invalidateChild(ViewGroup.java:5246)
    at android.view.View.invalidateInternal(View.java:13573)
    at android.view.View.invalidate(View.java:13537)
    at android.view.View.invalidate(View.java:13521)
    at com.mapbox.mapboxsdk.maps.widgets.MyLocationView$MyLocationShowBehavior.invalidate(MyLocationView.java:851)
    at com.mapbox.mapboxsdk.maps.widgets.MyLocationView.update(MyLocationView.java:379)
    at com.mapbox.mapboxsdk.maps.widgets.MyLocationView$MarkerCoordinateAnimatorListener.onAnimationUpdate(MyLocationView.java:700)
    at android.animation.ValueAnimator.animateValue(ValueAnimator.java:1346)
    at android.animation.ValueAnimator.end(ValueAnimator.java:1055)
    at com.mapbox.mapboxsdk.maps.widgets.MyLocationView$MyLocationShowBehavior.updateLatLng(MyLocationView.java:833)
    at com.mapbox.mapboxsdk.maps.widgets.MyLocationView.setLocation(MyLocationView.java:458)
    at com.mapbox.mapboxsdk.maps.widgets.MyLocationView$GpsLocationListener.onLocationChanged(MyLocationView.java:592)
    at com.mapbox.mapboxsdk.location.LocationSource.onLocationChanged(LocationSource.java:136)
    at com.mapzen.android.lost.internal.LostClientManager.reportLocationChanged(LostClientManager.java:173)
    at com.mapzen.android.lost.internal.FusedLocationProviderServiceImpl.reportLocation(FusedLocationProviderServiceImpl.java:126)
    at com.mapzen.android.lost.internal.FusionEngine.onLocationChanged(FusionEngine.java:191)
[...]

PR #8710, opened 2017-04-11
[android] Add a way to use a custom location source

### LINKED ISSUE #8668: Allow setting of custom Location Engine
Currently, it doesn't appear to be a way to use a custom location engine rather then the Lost location provider. 

cc: @mapbox/android

Guardiola31337:
Fixes #8668 

👀  @tobrun

### DIFF
--- a/platform/android/MapboxGLAndroidSDK/src/main/java/com/mapbox/mapboxsdk/maps/MapboxMap.java
+++ b/platform/android/MapboxGLAndroidSDK/src/main/java/com/mapbox/mapboxsdk/maps/MapboxMap.java
@@ -37,10 +37,12 @@ import com.mapbox.mapboxsdk.constants.MyLocationTracking;
 import com.mapbox.mapboxsdk.constants.Style;
 import com.mapbox.mapboxsdk.geometry.LatLng;
 import com.mapbox.mapboxsdk.geometry.LatLngBounds;
+import com.mapbox.mapboxsdk.location.LocationSource;
 import com.mapbox.mapboxsdk.maps.widgets.MyLocationViewSettings;
 import com.mapbox.mapboxsdk.style.layers.Filter;
 import com.mapbox.mapboxsdk.style.layers.Layer;
 import com.mapbox.mapboxsdk.style.sources.Source;
+import com.mapbox.services.android.telemetry.location.LocationEngine;
 import com.mapbox.services.commons.geojson.Feature;
 
 import java.lang.reflect.ParameterizedType;
@@ -1745,6 +1747,18 @@ public final class MapboxMap {
   public void setOnMyLocationChangeListener(@Nullable MapboxMap.OnMyLocationChangeListener
                                               listener) {
     trackingSettings.setOnMyLocationChangeListener(listener);
+  }
+
+  /**
+   * Replaces the location source of the my-location layer.
+   *
+   * @param locationSource A {@link LocationEngine} location source to use in the my-location layer.
+   *                       Set to null to use the default {@link LocationSource}
+   *                       location source.
+   */
+  @UiThread
+  public void setLocationSource(@Nullable LocationEngine locationSource) {
+    trackingSettings.setLocationSource(locationSource);
   }
 
   /**
--- a/platform/android/MapboxGLAndroidSDK/src/main/java/com/mapbox/mapboxsdk/maps/TrackingSettings.java
+++ b/platform/android/MapboxGLAndroidSDK/src/main/java/com/mapbox/mapboxsdk/maps/TrackingSettings.java
@@ -12,6 +12,7 @@ import com.mapbox.mapboxsdk.constants.MyBearingTracking;
 import com.mapbox.mapboxsdk.constants.MyLocationTracking;
 import com.mapbox.mapboxsdk.location.LocationSource;
 import com.mapbox.mapboxsdk.maps.widgets.MyLocationView;
+import com.mapbox.services.android.telemetry.location.LocationEngine;
 import com.mapbox.services.android.telemetry.location.LocationEngineListener;
 import com.mapbox.services.android.telemetry.permissions.PermissionsManager;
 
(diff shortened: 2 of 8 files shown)


platform/android/MapboxGLAndroidSDK/src/main/java/com/mapbox/mapboxsdk/maps/MapboxMap.java
```diff
@@ -1731,6 +1733,18 @@ public void setOnMyLocationChangeListener(@Nullable MapboxMap.OnMyLocationChange
...
 
   /**
+   * Replaces the location source of the my-location layer.
+   *
+   * @param locationSource A {@link LocationEngine} location source to use in the my-location layer.
+   *                       Set to null to use the default {@link LocationSource}
+   *                       location source.
+   */
+  @UiThread
+  public void setLocationSource(@Nullable LocationEngine locationSource) {
```
zugaldia: Let's call this `setLocationEngine`.

platform/android/MapboxGLAndroidSDK/src/main/java/com/mapbox/mapboxsdk/maps/TrackingSettings.java
```diff
@@ -362,6 +365,11 @@ void setMyLocationEnabled(boolean locationEnabled) {
     myLocationView.setEnabled(locationEnabled);
   }
 
+  void setLocationSource(LocationEngine locationSource) {
```
zugaldia: Similarly, let's call it setLocationEngine().

platform/android/MapboxGLAndroidSDK/src/main/java/com/mapbox/mapboxsdk/maps/widgets/MyLocationView.java
```diff
@@ -58,6 +58,7 @@
 
   private LatLng latLng;
   private Location location;
+  private LocationEngine locationSource;
```
zugaldia: I'd rather call it `locationEngine` even if `LocationSource` is our default implementation.

platform/android/MapboxGLAndroidSDK/src/main/java/com/mapbox/mapboxsdk/maps/widgets/MyLocationView.java
```diff
@@ -561,22 +563,29 @@ public void setContentPadding(int[] padding) {
     contentPaddingY = (padding[1] - padding[3]) / 2;
   }
 
+  public void setLocationSource(LocationEngine locationSource) {
```
zugaldia: Same: setLocationEngine().

platform/android/MapboxGLAndroidSDK/src/main/java/com/mapbox/mapboxsdk/maps/widgets/MyLocationView.java
```diff
@@ -561,22 +563,29 @@ public void setContentPadding(int[] padding) {
...
+    this.locationSource = locationSource;
+    toggleGps(isEnabled());
+  }
+
   private static class GpsLocationListener implements LocationEngineListener {
 
     private WeakReference<MyLocationView> userLocationView;
+    private WeakReference<LocationEngine> locationSource;
```
zugaldia: Similarly, let's call the variable `locationEngine`.

platform/android/MapboxGLAndroidSDKTestApp/src/main/java/com/mapbox/mapboxsdk/testapp/activity/userlocation/CustomLocationEngineActivity.java
```diff
@@ -0,0 +1,131 @@
...
+        mapboxMap = map;
+      }
+    });
+
+    locationToggleFab = (FloatingActionButton) findViewById(R.id.fabLocationToggle);
+    locationToggleFab.setOnClickListener(new View.OnClickListener() {
+      @Override
+      public void onClick(View view) {
+        if (mapboxMap != null) {
+          mapboxMap.setLocationSource(locationServices);
```
zugaldia: Do we need to set a location source every time the FAB is clicked? Wouldn't we need to set it just once if it hasn't been set before? Also, I don't see a `getLocationSource()` method to check that, or alternatively a `isCustomLocationSourceSet()`.

Guardiola31337: @zugaldia you're right, setting it once would be enough.
I don't see the necessity of having `getLocationSource` or `isCustomLocationSourceSet` methods, because if it isn't set we use `LocationSource` (default). What would be the use case?

zugaldia: > What would be the use case?

@Guardiola31337 Check if it's been set or not. For example, this activity could've been written like:

```
if (mapboxMap.getLocationSource() == null) {
  mapboxMap.setLocationSource(locationServices);
}
```

Guardiola31337: @zugaldia there is always a location source. It can't be `null`. So it doesn't make sense.

zugaldia: @Guardiola31337 Alternatively:

```
if (mapboxMap.getLocationSource() instanceof MyCustomLocationSource) {
  // Safe cast
}
```

Ehekatl:
I manage to use this to fix issue #8639 #8848
However after I change location source to my own location engine in background service, the map view lost the ability to update user location gradually and start jumping to the new position immediately on location change.
Is there any way to solve it?

zugaldia:
@Ehekatl thanks for the report. I'd like to explore the root cause for this misbehavior. Could you open a new ticket and, if possible, share the code for your custom location engine? For reference, we provide a sample implementation of a custom location engine on [`CustomLocationEngineActivity`](https://github.com/mapbox/mapbox-gl-native/blob/master/platform/android/MapboxGLAndroidSDKTestApp/src/main/java/com/mapbox/mapboxsdk/testapp/activity/userlocation/CustomLocationEngineActivity.java)

PR #8711, opened 2017-04-11
Fix Node.js detection for Android

kkaefer:
In #8662 we added support for `nodejs` when `node` wasn't found, but the check was missing a second `|`.


brunoabinader:
Nice catch!

Issue #7070, opened 2016-11-15
MGLMultiPolygon doesn't support coordinate property


aayers-agribile:
**Platform:**
iOS
**Mapbox SDK version:**
3.3.6
### Steps to trigger behavior

 1. Create a MGLMultiPolygon object (I parse them from a geojson source)
 2. Request the MGLMultiPolygon object's 'coordinate' property
 3. Get "MGLShape is an abstract class error"

### Expected behavior
The centroid of the MultiPoly would be returned from the coordinate property

### Actual behavior
I get an 'MGLShape is an abstract class' error.

1ec5:
There are two issues here:

* `-[MGLMultiPolygon coordinate]` should be implemented in order to avoid the exception in MGLShape. For now, we can have it return the first coordinate, as MGLPolyline and MGLPolygon do.
* `-[MGLPolyline coordinate]`, `-[MGLPolygon coordinate]`, and `-[MGLMultiPolygon coordinate]` should return the centroid. This is a larger task that would require [Polylabel](https://github.com/mapbox/polylabel/).

/cc @mourner

mourner:
@1ec5 wasn't Polylabel already integrated recently by @jfirebaugh?

jfirebaugh:
I would argue that the problem here is that `-coordinate` is even in the interface of `MGLShape` and subclasses. It should be in the interface of `MGLPointAnnotation` only. The artificial definition of a "coordinate" for polygons, multipolygons, line strings -- and the lengths we will need to go through to implement it -- are signs that this is a design error.

1ec5:
`coordinate` is defined on the MGLAnnotation protocol, so any class that conforms to it must implement the property. This is consistent with [MapKit](https://developer.apple.com/reference/mapkit/mkannotation/1429524-coordinate).

1ec5:
To expand on https://github.com/mapbox/mapbox-gl-native/issues/7070#issuecomment-272346175, having `coordinate` return something meaningful for the more exotic shape types is just icing on the cake. The real bug here is that `-[MGLMultiPolygon coordinate]` is unimplemented. Even having `-[MGLMultiPolygon coordinate]` return the first coordinate of the first polygon would’ve been fine, but #8713 goes the extra mile to return something more meaningful.

jfirebaugh:
Right -- I'm suggesting that either:

* `MGLShape` should not conform to `MGLAnnotation`, or
[...]

PR #8713, opened 2017-04-11
[ios,macos] multipolygon coordinate

fabian-guerra:
Fixes https://github.com/mapbox/mapbox-gl-native/issues/7070

### DIFF
--- a/platform/darwin/src/MGLGeometry.h
+++ b/platform/darwin/src/MGLGeometry.h
@@ -6,6 +6,30 @@
 
 NS_ASSUME_NONNULL_BEGIN
 
+
+typedef double MGLLocationRadians;
+typedef double MGLRadianDistance;
+typedef double MGLRadianDirection;
+
+extern double const MGLMetersPerRadian;
+
+/** Defines the coordinate by a `MGLRadianCoordinate2D`. */
+typedef struct MGLRadianCoordinate2D {
+    MGLLocationRadians latitude;
+    MGLLocationRadians longitude;
+} MGLRadianCoordinate2D;
+
+/**
+ Creates a new `MGLRadianCoordinate2D` from the given latitudinal and longitudinal.
+ */
+NS_INLINE MGLRadianCoordinate2D MGLRadianCoordinate2DMake(MGLLocationRadians latitude, MGLLocationRadians longitude) {
+    MGLRadianCoordinate2D radianCoordinate;
+    radianCoordinate.latitude = latitude;
+    radianCoordinate.longitude = longitude;
+    return radianCoordinate;
+}
+
+
 /** Defines the area spanned by an `MGLCoordinateBounds`. */
 typedef struct MGLCoordinateSpan {
     /** Latitudes spanned by an `MGLCoordinateBounds`. */
@@ -126,5 +150,46 @@ NS_INLINE CGFloat MGLRadiansFromDegrees(CLLocationDegrees degrees) {
 NS_INLINE CLLocationDegrees MGLDegreesFromRadians(CGFloat radians) {
     return radians * 180 / M_PI;
 }
+
+/** Returns MGLRadianCoordinate2D, converted from CLLocationCoordinate2D. */
+NS_INLINE MGLRadianCoordinate2D MGLRadianCoordinateFromLocationCoordinate(CLLocationCoordinate2D locationCoordinate) {
+    return MGLRadianCoordinate2DMake(MGLRadiansFromDegrees(locationCoordinate.latitude),
+                                     MGLRadiansFromDegrees(locationCoordinate.longitude));
+}
+
+/*
+ Returns the distance in radians given two coordinates.
+ */
+NS_INLINE MGLRadianDistance MGLDistanceBetweenRadianCoordinates(MGLRadianCoordinate2D from, MGLRadianCoordinate2D to)
+{
+    double a = pow(sin((to.latitude - from.latitude) / 2), 2)
+        + pow(sin((to.longitude - from.longitude) / 2), 2) * cos(from.latitude) * cos(to.latitude);
+    
+    return 2 * atan2(sqrt(a), sqrt(1 - a));
+}
+
+/*
+ Returns direction in radians given two coordinates.
+ */
+NS_INLINE MGLRadianDirection MGLRadianCoordinatesDirection(MGLRadianCoordinate2D from, MGLRadianCoordinate2D to) {
+    double a = sin(to.longitude - from.longitude) * cos(to.latitude);
+    double b = cos(from.latitude) * sin(to.latitude)
+                - sin(from.latitude) * cos(to.latitude) * cos(to.longitude - from.longitude);
+    return atan2(a, b);
+}
+
+/*
+ Returns coordinate at a given distance and direction away from coordinate.
+ */
+NS_INLINE MGLRadianCoordinate2D MGLRadianCoordinateAtDistanceFacingDirection(MGLRadianCoordinate2D coordinate,
+                                                                               MGLRadianDistance distance,
+                                                                               MGLRadianDirection direction) {
+    double otherLatitude = asin(sin(coordinate.latitude) * cos(distance)
+                             + cos(coordinate.latitude) * sin(distance) * cos(direction));
+    double otherLongitude = coordinate.longitude + atan2(sin(direction) * sin(distance) * cos(coordinate.latitude),
+                                           cos(distance) - sin(coordinate.latitude) * sin(otherLatitude));
+    return MGLRadianCoordinate2DMake(otherLatitude, otherLongitude);
+}
+    
 
 NS_ASSUME_NONNULL_END
--- a/platform/darwin/src/MGLGeometry.h
+++ b/platform/darwin/src/MGLGeometry.h
@@ -6,30 +6,6 @@
 
 NS_ASSUME_NONNULL_BEGIN
 
-
-typedef double MGLLocationRadians;
-typedef double MGLRadianDistance;
-typedef double MGLRadianDirection;
-
-extern double const MGLMetersPerRadian;
-
-/** Defines the coordinate by a `MGLRadianCoordinate2D`. */
-typedef struct MGLRadianCoordinate2D {
-    MGLLocationRadians latitude;
-    MGLLocationRadians longitude;
-} MGLRadianCoordinate2D;
-
-/**
- Creates a new `MGLRadianCoordinate2D` from the given latitudinal and longitudinal.
- */
-NS_INLINE MGLRadianCoordinate2D MGLRadianCoordinate2DMake(MGLLocationRadians latitude, MGLLocationRadians longitude) {
-    MGLRadianCoordinate2D radianCoordinate;
-    radianCoordinate.latitude = latitude;
-    radianCoordinate.longitude = longitude;
-    return radianCoordinate;
-}
-
-
 /** Defines the area spanned by an `MGLCoordinateBounds`. */
 typedef struct MGLCoordinateSpan {
     /** Latitudes spanned by an `MGLCoordinateBounds`. */
@@ -150,46 +126,5 @@ NS_INLINE CGFloat MGLRadiansFromDegrees(CLLocationDegrees degrees) {
 NS_INLINE CLLocationDegrees MGLDegreesFromRadians(CGFloat radians) {
     return radians * 180 / M_PI;
 }
-
-/** Returns MGLRadianCoordinate2D, converted from CLLocationCoordinate2D. */
-NS_INLINE MGLRadianCoordinate2D MGLRadianCoordinateFromLocationCoordinate(CLLocationCoordinate2D locationCoordinate) {
-    return MGLRadianCoordinate2DMake(MGLRadiansFromDegrees(locationCoordinate.latitude),
-                                     MGLRadiansFromDegrees(locationCoordinate.longitude));
-}
-
-/*
- Returns the distance in radians given two coordinates.
- */
-NS_INLINE MGLRadianDistance MGLDistanceBetweenRadianCoordinates(MGLRadianCoordinate2D from, MGLRadianCoordinate2D to)
-{
-    double a = pow(sin((to.latitude - from.latitude) / 2), 2)
-        + pow(sin((to.longitude - from.longitude) / 2), 2) * cos(from.latitude) * cos(to.latitude);
-    
-    return 2 * atan2(sqrt(a), sqrt(1 - a));
-}
-
-/*
- Returns direction in radians given two coordinates.
- */
-NS_INLINE MGLRadianDirection MGLRadianCoordinatesDirection(MGLRadianCoordinate2D from, MGLRadianCoordinate2D to) {
-    double a = sin(to.longitude - from.longitude) * cos(to.latitude);
-    double b = cos(from.latitude) * sin(to.latitude)
-                - sin(from.latitude) * cos(to.latitude) * cos(to.longitude - from.longitude);
-    return atan2(a, b);
-}
-
-/*
- Returns coordinate at a given distance and direction away from coordinate.
- */
-NS_INLINE MGLRadianCoordinate2D MGLRadianCoordinateAtDistanceFacingDirection(MGLRadianCoordinate2D coordinate,
-                                                                               MGLRadianDistance distance,
-                                                                               MGLRadianDirection direction) {
-    double otherLatitude = asin(sin(coordinate.latitude) * cos(distance)
-                             + cos(coordinate.latitude) * sin(distance) * cos(direction));
-    double otherLongitude = coordinate.longitude + atan2(sin(direction) * sin(distance) * cos(coordinate.latitude),
-                                           cos(distance) - sin(coordinate.latitude) * sin(otherLatitude));
-    return MGLRadianCoordinate2DMake(otherLatitude, otherLongitude);
-}
-    
 
 NS_ASSUME_NONNULL_END
--- a/platform/darwin/src/MGLGeometry_Private.h
+++ b/platform/darwin/src/MGLGeometry_Private.h
@@ -18,6 +18,10 @@ NS_INLINE mbgl::Point<double> MGLPointFromLocationCoordinate2D(CLLocationCoordin
     return mbgl::Point<double>(coordinate.longitude, coordinate.latitude);
 }
 
+NS_INLINE CLLocationCoordinate2D MGLLocationCoordinate2DFromPoint(mbgl::Point<double> point) {
+    return CLLocationCoordinate2DMake(point.y, point.x);
+}
+
 NS_INLINE CLLocationCoordinate2D MGLLocationCoordinate2DFromLatLng(mbgl::LatLng latLng) {
     return CLLocationCoordinate2DMake(latLng.latitude(), latLng.longitude());
 }
--- a/platform/darwin/src/MGLGeometry_Private.h
+++ b/platform/darwin/src/MGLGeometry_Private.h
@@ -96,7 +96,7 @@ NS_INLINE MGLRadianCoordinate2D MGLRadianCoordinateFromLocationCoordinate(CLLoca
 NS_INLINE MGLRadianDistance MGLDistanceBetweenRadianCoordinates(MGLRadianCoordinate2D from, MGLRadianCoordinate2D to)
 {
     double a = pow(sin((to.latitude - from.latitude) / 2), 2)
-    + pow(sin((to.longitude - from.longitude) / 2), 2) * cos(from.latitude) * cos(to.latitude);
+                + pow(sin((to.longitude - from.longitude) / 2), 2) * cos(from.latitude) * cos(to.latitude);
     
     return 2 * atan2(sqrt(a), sqrt(1 - a));
 }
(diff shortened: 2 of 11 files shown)


platform/darwin/src/MGLPolygon.mm
```diff
@@ -54,6 +55,17 @@ - (NSUInteger)hash {
     return [super hash] + [[self geoJSONDictionary] hash];
 }
 
+- (CLLocationCoordinate2D)coordinate {
+    CLLocationCoordinate2D centroid;
+    
+    // pole of inaccessibility
+    auto poi = mapbox::polylabel([self polygon]);
+    centroid.latitude = poi.y;
+    centroid.longitude = poi.x;
```
1ec5: In general, we should initialize simple structs like this using factory methods. This is especially important for geometry structs where it’s easy to confuse latitude and longitude or x and y. Create an `MGLLocationCoordinate2DFromPoint()` method similar to [`MGLPointFromLocationCoordinate2D()`](https://github.com/mapbox/mapbox-gl-native/blob/4b84ae119dd9489b6fbd6a11eca2102e0a79747b/platform/darwin/src/MGLGeometry_Private.h#L17) and call it here.

platform/darwin/src/MGLPolygon.mm
```diff
@@ -155,6 +167,22 @@ - (NSUInteger)hash {
     return hash;
 }
 
+- (CLLocationCoordinate2D)coordinate {
+    CLLocationCoordinate2D centroid;
+    
+    mbgl::Polygon<double> geometry;
+    for (MGLPolygon *polygon in self.polygons) {
+        geometry.push_back(polygon.ring);
```
1ec5: This doesn’t account for holes in the polygon, which can affect the shape’s centroid. (We try to avoid labeling a polygon inside one of its holes.)

This class already has a `-geometryObject` method that returns a `mbgl::Geometry<double>`, but it actually creates an `mbgl::MultiPolygon<double>`. Consider factoring out a `-multiPolygon` method, similar to how `-[MGLPolygon geometryObject]` calls [`-[MGLPolygon polygon]`](https://github.com/mapbox/mapbox-gl-native/blob/4b84ae119dd9489b6fbd6a11eca2102e0a79747b/platform/darwin/src/MGLPolygon.mm#L69).

platform/darwin/src/MGLPolyline.mm
```diff
@@ -52,6 +53,20 @@ - (BOOL)isEqual:(id)other {
     return self == other || ([other isKindOfClass:[MGLPolyline class]] && [super isEqual:other]);
 }
 
+- (CLLocationCoordinate2D)coordinate {
+    CLLocationCoordinate2D centroid;
+    
+    mbgl::Polygon<double> multiLineString;
```
1ec5: A polygon isn’t a multilinestring.

platform/darwin/test/MGLCodingTests.m
```diff
@@ -66,12 +66,27 @@ - (void)testPolyline {
     [unarchivedPolyline replaceCoordinatesInRange:NSMakeRange(0, 1) withCoordinates:otherCoordinates];
 
     XCTAssertNotEqualObjects(polyline, unarchivedPolyline);
+    
+    CLLocationCoordinate2D lineCoordinates[] = {
+        CLLocationCoordinate2DMake(100.0, 0.0),
+        CLLocationCoordinate2DMake(101.0, 0.0),
+    };
+    
+    NSUInteger lnc = sizeof(lineCoordinates) / sizeof(CLLocationCoordinate2D);
```
1ec5: Avoid excessive abbreviation: `lineCoordinatesCount` would be clearer.

platform/darwin/test/MGLCodingTests.m
```diff
@@ -248,6 +279,31 @@ - (void)testMultiPolygon {
         CLLocationCoordinate2DMake(20, 21),
         CLLocationCoordinate2DMake(30, 31),
     };
+    
+    CLLocationCoordinate2D outterSquare[] = {
```
1ec5: Typo: “outter”.

platform/darwin/test/MGLCodingTests.m
```diff
@@ -84,8 +99,24 @@ - (void)testPolygon {
...
+        CLLocationCoordinate2DMake(101.0, 1.0),
+        CLLocationCoordinate2DMake(100.0, 1.0),
+    };
+    
+    NSUInteger snc = sizeof(squareCoordinates) / sizeof(CLLocationCoordinate2D);
+    MGLPolygon *squarePolygon = [MGLPolygon polygonWithCoordinates:squareCoordinates count:snc];
+    CLLocationCoordinate2D squareCenter = CLLocationCoordinate2DMake(100.5, 0.5);
+    
+    XCTAssert([squarePolygon coordinate].latitude == squareCenter.latitude &&
+              [squarePolygon coordinate].longitude == squareCenter.longitude);
```
1ec5: Split this into two calls to `XCTAssertEqual()`, which provides a more informative message than `XCTAssert()` when it fails.

platform/darwin/src/MGLPolyline.mm
1ec5: This appears to be the cause of the build failures on iOS and macOS:

```
▸ Compiling MGLPolyline.mm

❌  /Users/vagrant/git/platform/darwin/src/MGLPolyline.mm:60:21: no matching member function for call to 'push_back'

    multiLineString.push_back([self lineString]);
    ~~~~~~~~~~~~~~~~^~~~~~~~~



❌  /Users/vagrant/git/platform/darwin/src/MGLPolyline.mm:138:25: no matching member function for call to 'push_back'

        multiLineString.push_back([polyline lineString]);
```

platform/darwin/src/MGLPolygon.mm
1ec5: A multipolygon can represent multiple disjoint polygons, but this code gloms them together into a single polygon geometry. I believe that would be invalid according to the GeoJSON specification.

I belatedly realized that `polylabel()` only accepts an `mbgl::geometry::polygon`, not an `mbgl::geometry::multipolygon`. As I pointed out in https://github.com/mapbox/mapbox-gl-native/issues/7070#issuecomment-294022436, returning the centroid of an exotic shape like a multipolygon would be a low priority compared to some of the other functionality implemented here. For now, let’s just calculate the centroid of the first polygon in the shape and return that.

platform/darwin/src/MGLPolyline.mm
1ec5: Polylabel probably isn’t the right tool for this job. In JavaScript, we’d use [turf-line-distance](https://github.com/Turfjs/turf/tree/master/packages/turf-line-distance) to determine the length of the polyline, then use [turf-along](https://github.com/Turfjs/turf/tree/master/packages/turf-along) to get the midway point. It might be interesting to port this functionality to C or C++ for use in the SDKs, but that’s enough work that we should probably tackle it in a separate PR.

platform/darwin/src/MGLPolygon.mm
```diff
@@ -54,6 +55,14 @@ - (NSUInteger)hash {
     return [super hash] + [[self geoJSONDictionary] hash];
 }
 
+- (CLLocationCoordinate2D)coordinate {
+    // pole of inaccessibility
+    auto poi = mapbox::polylabel([self polygon]);
+    CLLocationCoordinate2D centroid = MGLLocationCoordinate2DFromPoint(poi);
+    
+    return centroid;
```
1ec5: FYI, it's perfectly acceptable to return the result of MGLLocationCoordinate2DFromPoint() directly without creating a local variable for it. If you need to verify the return value here, Xcode's debugger displays a "Return Value" entry in the Variables view when stepping out of a function.

platform/darwin/src/MGLPolyline.mm
```diff
@@ -52,6 +53,27 @@ - (BOOL)isEqual:(id)other {
     return self == other || ([other isKindOfClass:[MGLPolyline class]] && [super isEqual:other]);
 }
 
+- (CLLocationCoordinate2D)coordinate {
+    mbgl::Polygon<double> polyline;
```
1ec5: This code is incorrectly treating a polyline as if it's a polygon. Pole of inaccessibility isn't the right tool to use for calculating the midpoint of a polyline. Instead, you would need to manually [determine the length of the polyline](https://github.com/Turfjs/turf/tree/master/packages/turf-line-distance) and [find the point half of that distance](https://github.com/Turfjs/turf/tree/master/packages/turf-along) from the start. We've already ported this functionality to Swift as part of [Geometry.swift](https://github.com/mapbox/mapbox-navigation-ios/blob/master/MapboxCoreNavigation/Geometry.swift). Feel free to take care of this in a separate PR.

1ec5: Specifically, [`turf-along`](https://github.com/Turfjs/turf/blob/142e137ce0c758e2825a260ab32b24db0aa19439/packages/turf-along/index.js) was ported to Swift as [`coordinate(at:fromStartOf:)`](https://github.com/mapbox/mapbox-navigation-ios/blob/0acfa7f3719418af138b49d970be586092fea5ea/MapboxCoreNavigation/Geometry.swift#L228). (Note that Geometry.swift relies on [this overloaded operator](https://github.com/mapbox/mapbox-navigation-ios/blob/0acfa7f3719418af138b49d970be586092fea5ea/MapboxCoreNavigation/Geometry.swift#L63-L67), which can be difficult to spot sometimes.)

fabian-guerra: @1ec5 could you please see the imp for this. I changed it according to your comment.

platform/darwin/src/MGLGeometry.h
```diff
@@ -6,6 +6,30 @@
...
 
+
+typedef double MGLLocationRadians;
+typedef double MGLRadianDistance;
+typedef double MGLRadianDirection;
+
+extern double const MGLMetersPerRadian;
+
+/** Defines the coordinate by a `MGLRadianCoordinate2D`. */
+typedef struct MGLRadianCoordinate2D {
```
1ec5: Let’s move all this new code to MGLGeometry_Private.h. MGLRadianCoordinate2D is only useful for an internal calculation, but it isn’t used in any public API. I think its presence would confuse developers who are looking for normal coordinates to use with various public APIs.

platform/darwin/src/MGLPolyline.mm
```diff
@@ -54,23 +54,62 @@ - (BOOL)isEqual:(id)other {
...
+            MGLRadianCoordinate2D from = MGLRadianCoordinateFromLocationCoordinate(coordinates[i]);
+            MGLRadianCoordinate2D to = MGLRadianCoordinateFromLocationCoordinate(coordinates[i + 1]);
+            
+            if (traveled >= middle) {
+                double overshoot = middle - traveled;
+                if (overshoot == 0) {
+                    return coordinates[i];
+                }
+                to = MGLRadianCoordinateFromLocationCoordinate(coordinates[i - 1]);
+                MGLRadianDirection direction = [self direction:from to:to] - 180;
```
1ec5: `-direction:to:` returns a `CLLocationDirection`, not an `MGLRadianDirection`.

platform/darwin/src/MGLPolyline.mm
```diff
@@ -54,23 +54,62 @@ - (BOOL)isEqual:(id)other {
...
+    double longitude = 0.0;
     
-    polyline.push_back(geometry);
+    NSUInteger count = self.pointCount;
+    CLLocationCoordinate2D *coordinates = self.coordinates;
     
-    // pole of inaccessibility
-    auto poi = mapbox::polylabel(polyline);
+    for (NSUInteger i = 0; i < count; i++) {
+        if (i + 1 >= count) {
```
fabian-guerra: Actually I can restrict to `count - 1` since we don't need to check the last element.

fabian-guerra:
@1ec5 I moved the radian functions to MGLGeometry_Private.h and fixed minor issues, could you please take a look again.

platform/darwin/src/MGLGeometry_Private.h
```diff
@@ -63,3 +85,43 @@ CLLocationDistance MGLAltitudeForZoomLevel(double zoomLevel, CGFloat pitch, CLLo
...
+                                     MGLRadiansFromDegrees(locationCoordinate.longitude));
+}
+
+/*
+ Returns the distance in radians given two coordinates.
+ */
+NS_INLINE MGLRadianDistance MGLDistanceBetweenRadianCoordinates(MGLRadianCoordinate2D from, MGLRadianCoordinate2D to)
+{
+    double a = pow(sin((to.latitude - from.latitude) / 2), 2)
+    + pow(sin((to.longitude - from.longitude) / 2), 2) * cos(from.latitude) * cos(to.latitude);
```
1ec5: Nit: this follow-on line should be indented (also with `b` in `MGLRadianCoordinatesDirection()` below).

platform/darwin/src/MGLPolyline.mm
```diff
@@ -73,38 +73,34 @@ - (CLLocationCoordinate2D)coordinate {
                     return coordinates[i];
                 }
                 to = MGLRadianCoordinateFromLocationCoordinate(coordinates[i - 1]);
-                MGLRadianDirection direction = [self direction:from to:to] - 180;
+                CLLocationDirection direction = [self direction:from to:to] - 180;
                 MGLRadianCoordinate2D otherCoordinate = MGLRadianCoordinateAtDistanceFacingDirection(from,
-                                                                                                     overshoot/MGLMetersPerRadian,
+                                                                                                     overshoot/mbgl::util::M2PI * mbgl::util::EARTH_RADIUS_M,
```
1ec5: 2π/6,378,137 is many orders of magnitude smaller than the value that `MGLMetersPerRadian` was originanlly set to. Can you explain why we need to divide the overshoot by 2π and multiply it by the Earth’s radius? [The equivalent code in Geometry.swift](https://github.com/mapbox/mapbox-navigation-ios/blob/5be6ada250e43804d3d74c5724687a761abb6271/MapboxCoreNavigation/Geometry.swift#L243) doesn’t seem to do that.

platform/darwin/test/MGLCodingTests.m
```diff
@@ -84,8 +100,24 @@ - (void)testPolygon {
...
+    CLLocationCoordinate2D squareCoordinates[] = {
+        CLLocationCoordinate2DMake(100.0, 0.0),
+        CLLocationCoordinate2DMake(101.0, 0.0),
+        CLLocationCoordinate2DMake(101.0, 1.0),
+        CLLocationCoordinate2DMake(100.0, 1.0),
+    };
+    
+    NSUInteger squareCoordinatesCount = sizeof(squareCoordinates) / sizeof(CLLocationCoordinate2D);
+    MGLPolygon *squarePolygon = [MGLPolygon polygonWithCoordinates:squareCoordinates count:squareCoordinatesCount];
+    CLLocationCoordinate2D squareCenter = CLLocationCoordinate2DMake(100.5, 0.5);
```
1ec5: Do you think this is an adequate test of the added functionality? Do we need more unit tests specifically about the `coordinate` properties that this PR adds? Perhaps we could port some tests from Turf or from [GeometryTests.swift](https://github.com/mapbox/mapbox-navigation-ios/blob/5be6ada250e43804d3d74c5724687a761abb6271/MapboxCoreNavigationTests/GeometryTests.swift).

platform/darwin/src/MGLPolyline.mm
fabian-guerra: This was a paste mistake, also Gometry.swift implements it this [way](https://github.com/mapbox/mapbox-navigation-ios/blob/5be6ada250e43804d3d74c5724687a761abb6271/MapboxCoreNavigation/Geometry.swift#L90).

1ec5: `coordinate(at:facing:)` divides `distance` by `metersPerRadian`. That’s very different than dividing `overshoot` by 2π and multiplying the result by `mbgl::util::EARTH_RADIUS_M`.

Regardless, this line is a _call_ to `MGLRadianCoordinateAtDistanceFacingDirection()`, not the implementation of `MGLRadianCoordinateAtDistanceFacingDirection()`. Shouldn’t it match [this line](https://github.com/mapbox/mapbox-navigation-ios/blob/5be6ada250e43804d3d74c5724687a761abb6271/MapboxCoreNavigation/Geometry.swift#L243) instead?

fabian-guerra: > coordinate(at:facing:) divides distance by metersPerRadian. That’s very different than dividing overshoot by 2π and multiplying the result by mbgl::util::EARTH_RADIUS_M.

Right that's why I said it was a paste mistake, it should be this way:
```
MGLRadianCoordinate2D otherCoordinate = MGLRadianCoordinateAtDistanceFacingDirection(from,
                                                                                                    overshoot/MGLMetersPerRadian,
                                                                                                     MGLRadiansFromDegrees(direction));
```

I mistakenly replaced a constant I had before called:
`const double MGLMetersPerRadian = 6373000.0;`

This is the underlaying impl I didn't consider I needed another function since this is the only place we are using this code.
```
/// Returns a coordinate a certain Haversine distance away in the given direction.
    func coordinate(at distance: CLLocationDistance, facing direction: CLLocationDirection) -> CLLocationCoordinate2D {
        let radianCoordinate = RadianCoordinate2D(self).coordinate(at: distance / metersPerRadian, facing: direction.toRadians())
        return CLLocationCoordinate2D(radianCoordinate)
    }
```

platform/darwin/test/MGLCodingTests.m
fabian-guerra: > Do you think this is an adequate test of the added functionality? 
Yes... if you think I'm missing test cases I will port the tests you mention above.

platform/darwin/src/MGLPolyline.mm
1ec5: It was probably preferable to use `mbgl::util::EARTH_RADIUS_M` instead of `MGLMetersPerRadian`, as I mentioned in https://github.com/mapbox/mapbox-gl-native/pull/8713#discussion_r115836563. The two values are within the same order of magnitude, which is a good sign. 😉

platform/darwin/test/MGLCodingTests.m
fabian-guerra: Looking at the tests are less related to poi but I added two test scenarios to `testPolyline` that cover a single and multiple segments.

platform/darwin/src/MGLPolyline.mm
fabian-guerra: You are right. Changed.

platform/darwin/test/MGLCodingTests.m
```diff
@@ -80,6 +80,32 @@ - (void)testPolyline {
...
+    CLLocationCoordinate2D segmentCoordinates[] = {
+        CLLocationCoordinate2DMake(35.040390, -85.311477),
+        CLLocationCoordinate2DMake(35.040390, -85.209510),
+    };
+    
+    NSUInteger segmentCoordinatesCount = sizeof(segmentCoordinates) / sizeof(CLLocationCoordinate2D);
+    MGLPolyline *segmentLine = [MGLPolyline polylineWithCoordinates:segmentCoordinates count:segmentCoordinatesCount];
+    CLLocationCoordinate2D segmentCenter = CLLocationCoordinate2DMake(35.0404006631, -85.2604935);
+    
+    XCTAssertEqualWithAccuracy([segmentLine coordinate].latitude, segmentCenter.latitude, 1);
```
1ec5: The third parameter to this macro is a float with the same magnitude as the maximum delta, not the number of decimal places. So here we’re only testing that the center between 35.040390 and 35.040390 rounded to the ones place is 35. One degree of latitude or longitude is quite a big difference; we should assert with much higher accuracy.

platform/darwin/test/MGLCodingTests.m
```diff
@@ -80,6 +80,32 @@ - (void)testPolyline {
     XCTAssertEqual([multiLine coordinate].latitude, multiLineCenter.latitude);
     XCTAssertEqual([multiLine coordinate].longitude, multiLineCenter.longitude);
     
+    CLLocationCoordinate2D segmentCoordinates[] = {
+        CLLocationCoordinate2DMake(35.040390, -85.311477),
+        CLLocationCoordinate2DMake(35.040390, -85.209510),
```
1ec5: We should vary both the latitude and the longitude to ensure that the logic works in both directions.

fabian-guerra: This test case is for a straight line. I addressed testing a non linear polyline in a test case below.

platform/ios/CHANGELOG.md
```diff
@@ -26,6 +26,7 @@ Mapbox welcomes participation and contributions from everyone. Please read [CONT
 ### Other changes
 
 * Xcode 8.0 or higher is now recommended for using this SDK. ([#8775](https://github.com/mapbox/mapbox-gl-native/pull/8775))
+* Fixed a crash when calling `MGLMultiPolygon.coordinate` [#8713](https://github.com/mapbox/mapbox-gl-native/pull/8713)
```
1ec5: Also note (perhaps in the “Annotations” section) that the `coordinate` property on MGLPolyline and MGLPolygon now returns a sensible value.

1ec5: We’ve already branched for iOS SDK v3.6.0: either retarget and rebase this PR on the release-ios-v3.6.0-android-v5.1.0 branch or add it to [this project](https://github.com/mapbox/mapbox-gl-native/projects/18) to be cherry-picked into that branch.

1ec5: > Also note (perhaps in the “Annotations” section) that the `coordinate` property on MGLPolyline and MGLPolygon now returns a sensible value.

→ #9088

PR #8712, opened 2017-04-11
Use AAssetManager instead of bundling libzip

kkaefer:
The Android NDK exposes [`AAssetManager`](https://developer.android.com/ndk/reference/group___asset.html), which is a way to access the [`AssetManager`](https://developer.android.com/reference/android/content/res/AssetManager.html), which is a way to access resources from the zipfile (among other things).

Implementing this would allow us to get rid of bundling libzip, and at the same time change the meaning of "asset" to be the Android definition of "asset".


kkaefer:
New:

```
* Measured  4,158,380 bytes for 'Platform=Android,Arch=arm-v5'
* Measured  3,818,420 bytes for 'Platform=Android,Arch=arm-v7'
* Measured  5,755,936 bytes for 'Platform=Android,Arch=arm-v8'
* Measured  6,280,252 bytes for 'Platform=Android,Arch=x86'
* Measured  5,936,208 bytes for 'Platform=Android,Arch=x86_64'
* Measured  8,085,868 bytes for 'Platform=Android,Arch=mips'
* Measured 14,047,814 bytes for 'Platform=Android,Arch=Archive'
```

Old:
```
* Reporting  4,199,356 bytes for 'Platform=Android,Arch=arm-v5'
* Reporting  3,851,204 bytes for 'Platform=Android,Arch=arm-v7'
* Reporting  5,796,896 bytes for 'Platform=Android,Arch=arm-v8'
* Reporting  6,337,612 bytes for 'Platform=Android,Arch=x86'
* Reporting  5,977,168 bytes for 'Platform=Android,Arch=x86_64'
* Reporting  8,151,916 bytes for 'Platform=Android,Arch=mips'
* Reporting 14,203,155 bytes for 'Platform=Android,Arch=Archive'
```

so we got ~30-50 KB of savings for standard archs.

platform/android/src/asset_manager_file_source.hpp
```diff
@@ -0,0 +1,28 @@
...
+public:
+    AssetManagerFileSource(jni::JNIEnv&, jni::Object<android::AssetManager>);
+    ~AssetManagerFileSource() override;
+
+    std::unique_ptr<AsyncRequest> request(const Resource&, Callback) override;
+
+private:
+    jni::UniqueObject<android::AssetManager> assetManager;
+    class Impl;
+    std::unique_ptr<util::Thread<Impl>> thread;
```
ivovandongen: We don't want to use the thread pool?

kkaefer: We do, but I've modeled this FileSource after all the other FileSources, and they are still using `util::Thread`. I attempted to port them to the actor system in https://github.com/mapbox/mapbox-gl-native/pull/7678

PR #8722, opened 2017-04-12
[android] Polygon holes signature refactor

Guardiola31337:
Refactored signature from `Hole` to `List<LatLng>`.

👀  @tobrun @ivovandongen


platform/android/MapboxGLAndroidSDK/src/main/java/com/mapbox/mapboxsdk/annotations/Hole.java
```diff
@@ -12,7 +12,7 @@
 /**
  * Encapsulates a {@link List} of {@link LatLng} points defining a hole
  */
-public class Hole extends ArrayList<LatLng> implements Parcelable {
+class Hole extends ArrayList<LatLng> implements Parcelable {
```
tobrun: since we are removing Hole.java from public API, is there a benefit of keeping this class around?

Guardiola31337: Didn't find a way to create a `Parcelable` of `List<List<LatLng>>`.
Do you know how?

ivovandongen: If the outermost class is parelable it can take care of all it's members. No need for a separate class for the holes in that case. Just move the logic over to Polygon.

Guardiola31337: It seems
```
List<List<LatLng>> holes = new ArrayList<>();
in.readList(holes, LatLng.class.getClassLoader());
```
does the trick. Fixed it 😉

Issue #8666, opened 2017-04-06
Exercise public APIs in Swift integration tests


1ec5:
Sometimes we make changes like #8541 that, per our [versioning guidelines](https://github.com/mapbox/mapbox-gl-native/wiki/Versions-and-tagging#mapbox-ios-sdk), don’t warrant a major version bump but may break backwards compatibility in Swift code due to the vagaries of Objective-C-to-Swift bridging. Avoiding these changes would be a major burden on development, but we should have a system for tracking them, to ensure that we communicate those changes in release notes.

Most of our unit tests are written in Objective-C, which is unable to catch changes to Swift bridging. We should supplement these unit tests with integration tests written in Swift that exercise the public APIs. At a minimum, they should call methods, implement delegate methods, etc., but they wouldn’t necessarily verify the results – that would remain the responsibility of the Objective-C++ unit tests, which have access to the underlying mbgl types.

/cc @mapbox/ios

boundsj:
https://github.com/mapbox/mapbox-gl-native/pull/8729 added integration tests for `MGLMapViewDelegate`. It also introduced a utility to help ensure test coverage. That utility and pattern introduced in #8729 can be used to help test our other protocols (i.e. `MGLOfflineStorageDelegate`, `MGLOverlay`, `MGLCalloutViewDelegate`)

friedbunny:
With recent Swift interface troubles around implementing optional location manager properties, this remains a valid issue.

/cc @riastrad @1ec5 @julianrex @frederoni

PR #8729, opened 2017-04-12
[ios, macos] Add Swift integration tests for map view delegate

boundsj:
Here is an idea to add the kind of tests described in https://github.com/mapbox/mapbox-gl-native/issues/8666 for `MGLMapViewDelegate`. 

As noted in #8666, the simple presence of the delegate method implementations will trigger warnings/errors in the tests if the delegate methods change -- no assertions required. There is a single test that reflects on the methods in the subject and the test class to ensure that there is full coverage. If new methods are added to the delegate in the future, this test will fail until the developer adds coverage.

cc @1ec5 @jmkiley @friedbunny

### DIFF
--- a/platform/ios/ios.xcodeproj/project.pbxproj
+++ b/platform/ios/ios.xcodeproj/project.pbxproj
@@ -138,6 +138,7 @@
 		4018B1C91CDC288A00F666AF /* MGLAnnotationView_Private.h in Headers */ = {isa = PBXBuildFile; fileRef = 4018B1C31CDC277F00F666AF /* MGLAnnotationView_Private.h */; };
 		4018B1CA1CDC288E00F666AF /* MGLAnnotationView.h in Headers */ = {isa = PBXBuildFile; fileRef = 4018B1C51CDC277F00F666AF /* MGLAnnotationView.h */; settings = {ATTRIBUTES = (Public, ); }; };
 		4018B1CB1CDC288E00F666AF /* MGLAnnotationView.h in Headers */ = {isa = PBXBuildFile; fileRef = 4018B1C51CDC277F00F666AF /* MGLAnnotationView.h */; settings = {ATTRIBUTES = (Public, ); }; };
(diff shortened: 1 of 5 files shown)


boundsj:
799a1a6 adds a flag to treat Swift warnings as errors to force the developer to reconcile small API changes that otherwise might fly under the radar if the tests target is not built locally and build output details are not observed.

platform/ios/CHANGELOG.md
```diff
@@ -15,6 +15,7 @@ Mapbox welcomes participation and contributions from everyone. Please read [CONT
 * Fixed a crash or console spew when MGLMapView is initialized with a frame smaller than 64 points wide by 64 points tall. ([#8562](https://github.com/mapbox/mapbox-gl-native/pull/8562))
 * The error passed into `-[MGLMapViewDelegate mapViewDidFailLoadingMap:withError:]` now includes a more specific description and failure reason. ([#8418](https://github.com/mapbox/mapbox-gl-native/pull/8418))
 * Fixed an issue where gesture recognizers associated with map view interactivity were not disabled when their related interactions were disabled. ([#8304](https://github.com/mapbox/mapbox-gl-native/pull/8304))
+* Added more Swift integration test coverage to help ensure that the effects of subtle SDK changes to Swift bridging are understood and captured in future release notes. ([#8729](https://github.com/mapbox/mapbox-gl-native/pull/8729))
```
1ec5: This is an internal change to our testing infrastructure; not sure we need a changelog entry about it.

platform/ios/ios.xcodeproj/project.pbxproj
```diff
@@ -2609,6 +2621,7 @@
 					"$(geometry_cflags)",
 					"$(geojson_cflags)",
 				);
+				OTHER_SWIFT_FLAGS = "-warnings-as-errors";
```
1ec5: Clever idea. I'm a bit concerned that this could be problematic in the future, since Swift doesn't have the pragmas that Objective-C does for intentionally suppressing warnings. On the other hand, if we rely on per-file flags, we risk missing the flag on a test file. So let's go with this and consider pulling it out if we ever need to live with a warning without porting a test to Objective-C.

boundsj:
d511335 consolidates the coverage check and turns it into something that'll be easier to reuse with tests of other protocols.

PR #8731, opened 2017-04-12
[all] Remove redundant scale-related camera methods

jfirebaugh:
We don't need to have two different measurement systems for map zoom.

Tested:

* Scroll zoom, pinch zoom, double click zoom, +/-, menu items in macosapp.
* Pinch zoom, tap-and-drag zoom, double tap zoom in iosapp.
* Pinch zoom, tap-and-drag zoom, double tap zoom in Android test app.
* Scroll zoom, double click zoom in glfw-app.

### DIFF
--- a/platform/android/MapboxGLAndroidSDK/src/main/java/com/mapbox/mapboxsdk/maps/MapGestureDetector.java
+++ b/platform/android/MapboxGLAndroidSDK/src/main/java/com/mapbox/mapboxsdk/maps/MapGestureDetector.java
@@ -230,7 +230,7 @@ final class MapGestureDetector {
           float scrollDist = event.getAxisValue(MotionEvent.AXIS_VSCROLL);
 
           // Scale the map by the appropriate power of two factor
-          transform.zoomBy(Math.pow(2.0, scrollDist), event.getX(), event.getY());
+          transform.zoomBy(scrollDist, event.getX(), event.getY());
 
           return true;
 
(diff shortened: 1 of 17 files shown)


platform/android/MapboxGLAndroidSDK/src/main/java/com/mapbox/mapboxsdk/maps/MapGestureDetector.java
```diff
@@ -230,7 +230,7 @@ boolean onGenericMotionEvent(MotionEvent event) {
           float scrollDist = event.getAxisValue(MotionEvent.AXIS_VSCROLL);
 
           // Scale the map by the appropriate power of two factor
-          transform.zoomBy(Math.pow(2.0, scrollDist), event.getX(), event.getY());
+          transform.zoomBy(scrollDist, event.getX(), event.getY());
```
jfirebaugh: Note that `Transform::zoomBy` used to be misleadingly named: it actually was a `scaleBy` operation. Now it's really `zoomBy`, and consequently we take the log2 of the input.

platform/macos/src/MGLMapView.mm
```diff
@@ -1402,10 +1383,10 @@ - (void)handlePanGesture:(NSPanGestureRecognizer *)gestureRecognizer {
         _mbglMap->cancelTransitions();
 
         if (gestureRecognizer.state == NSGestureRecognizerStateBegan) {
-            _scaleAtBeginningOfGesture = _mbglMap->getScale();
+            _zoomAtBeginningOfGesture = _mbglMap->getZoom();
         } else if (gestureRecognizer.state == NSGestureRecognizerStateChanged) {
-            CGFloat newZoomLevel = log2f(_scaleAtBeginningOfGesture) - delta.y / 75;
-            [self scaleBy:powf(2, newZoomLevel) / _mbglMap->getScale() atPoint:startPoint animated:NO];
+            CGFloat newZoomLevel = _zoomAtBeginningOfGesture - delta.y / 75;
```
1ec5: Did you test Shift-dragging to zoom in and out? Does the behavior match the previous behavior? Subtracting this delta value from a zoom level would presumably have a different effect than subtracting the same value from a scale.

1ec5: The double-tap-drag gesture behaves similarly on iOS.

In case it helps, the iOS gestures are documented [in jazzy docs](https://www.mapbox.com/ios-sdk/api/3.5.2/gesture-recognizers.html), and the macOS gestures are listed in macosapp under Help > Mapbox GL Help.

jfirebaugh: I just tested this and it's working as expected. Subtracting from a zoom level is equivalent to the previous behavior of subtracting from the log₂ of a scale.

1ec5: Ah, I missed the `log2f()` call there. 👍

PR #8732, opened 2017-04-12
[core] Inline GlyphSet into GlyphAtlas

jfirebaugh:
Continuation of #8669.


src/mbgl/text/glyph_atlas.cpp
```diff
@@ -87,45 +87,70 @@ GlyphAtlas::GlyphRequest& GlyphAtlas::requestRange(Entry& entry, const FontStack
...
+            auto it = entry.sdfs.find(glyph.id);
+            if (it == entry.sdfs.end()) {
+                // Glyph doesn't exist yet.
+                entry.sdfs.emplace(glyph.id, std::move(glyph));
+            } else if (it->second.metrics == glyph.metrics) {
+                if (it->second.bitmap != glyph.bitmap) {
+                    // The actual bitmap was updated; this is unsupported.
+                    Log::Warning(Event::Glyph, "Modified glyph changed bitmap represenation");
+                }
+                // At least try to update it in case it's currently unsused.
```
ChrisLoer: You didn't change this, but this logic is weird. We only request each glyph range once, so this logic seems to be supporting... the server responding to a range request with glyphs outside of that range?

I think we could just remove the "glyph update" logic, but if it's serving a purpose:

- "unsused" typo
- Why `it->second.bitmap = std::move(glyph.bitmap)` for the case that the two bitmaps are already equal?

jfirebaugh: This logic is ostensibly for the case where we revalidate a cached glyph PBF and receive an updated version. I'm not sure it ever happens in practice.

The two bitmaps are not necessarily equal; we might have logged the warning above and then continued through.

Issue #8708, opened 2017-04-11
-[MGLMapView showAnnotations:] should span the antimeridian


ab-rdarts:
**Platform:**
iOS
**Mapbox SDK version:**
3.5.2

### Steps to trigger behavior

 1. Add annotations like (latitude:0, longitude:179), (latitude:0, longitude:0), (latitude:0, longitude:-180)
 2. use showAnnotations method

### Expected behavior
MapView shows three annotations

### Actual behavior
MapView shows only two annotations and only after pan map a little bit (doesn't work at all if not pan)

[TestMap.zip](https://github.com/mapbox/mapbox-gl-native/files/913075/TestMap.zip)

1ec5:
It isn’t possible to zoom all the way out to the point where the Prime Meridian, antimeridian, and 179° east are all visible at the same time. The map is constrained to avoid showing multiple “copies” of the world simultaneously and to avoid showing anything beyond 85.051129° north or south (due to [Web Mercator](https://en.wikipedia.org/wiki/Web_Mercator) projection). In the sample application provided above, the map view is 375 points wide and 563 points tall. At this aspect ratio, attempting to show all three annotations would require breaking both of those constraints.

jfirebaugh:
@1ec5 Should we close this as working as designed then?

ab-rdarts:
@1ec5 I understand that in this particular case it's not possible to fit all annotations, but the problem in another side. you can move annotation (0, 0) close to other annotations so that it can fit. And in this case it will not work also, it will show only two of them which close to 180 longitude. You should take into account span of longitude and latitude instead of calculating min/max of longitudes

1ec5:
That’s a good point, @ab-rdarts! I’ve retitled this issue to reflect the fact that `-showAnnotations:` should be able to straddle the antimeridian if necessary. This would require `mbgl::Map::cameraForLatLngs()` to be able to straddle the antimeridian as well.

1ec5:
Fixed in #8744.

PR #8744, opened 2017-04-13
[core] Ignore shortest path in Map::cameraForLatLngs

### LINKED ISSUE #8737: Map::cameraForLatLngBounds returns incorrect CameraOptions
Been looking into an user reported bug that has the camera positioned above New York City  and is trying to move the camera to a `LatLngBounds` in China.

For the following China bounds:

screenshot at 2017-04-13 10 10 42

The following center coordinate is calculated with Map::cameraForLatLngBounds:

screenshot at 2017-04-13 10 11 15

This results in centering the camera on the east coast

image

While we were expecting: [34.6212005,104.1363335]

screenshot at 2017-04-13 10 22 10

#### Data to reproduce this with
New York City = [40.712730, -74.005953], z=12
China bounds =  [15.68169,73.499857], [53.560711, 134.77281]

GL-JS is handling the use-case described above correctly, see [here](http://jsfiddle.net/7jr0nbcv/).

brunoabinader:
When calculating the screen coordinates for the given geographical coordinates boundaries in `Map::cameraForLatLngs`, we use `Map::pixelForLatLng`, which is implemented by calling `Transform::latLngToScreenCoordinate`.

The latter assumed that all conversions were required to be using the longitude shortest path. This is not true for `Map::cameraForLatLngs`, in which we require the longest path to properly calculate the virtual camera center.

Fixes #8737.

/cc @tobrun @jfirebaugh

### DIFF
--- a/src/mbgl/map/map.cpp
+++ b/src/mbgl/map/map.cpp
@@ -586,7 +586,7 @@ CameraOptions Map::cameraForLatLngs(const std::vector<LatLng>& latLngs, const Ed
     ScreenCoordinate swPixel = {INFINITY, INFINITY};
     double viewportHeight = getSize().height;
     for (LatLng latLng : latLngs) {
-        ScreenCoordinate pixel = pixelForLatLng(latLng);
+        ScreenCoordinate pixel = impl->transform.latLngToScreenCoordinate(latLng);
         swPixel.x = std::min(swPixel.x, pixel.x);
         nePixel.x = std::max(nePixel.x, pixel.x);
         swPixel.y = std::min(swPixel.y, viewportHeight - pixel.y);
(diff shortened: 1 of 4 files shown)


src/mbgl/map/transform.hpp
```diff
@@ -144,7 +144,12 @@ class Transform : private util::noncopyable {
     bool isPanning() const { return state.isPanning(); }
 
     // Conversion and projection
-    ScreenCoordinate latLngToScreenCoordinate(const LatLng&) const;
+    enum class ConversionMode : bool {
+        FindShortestPath = 0,
+        IgnoreShortestPath
```
1ec5: 👍 `cameraForLatLngs()` itself may end up taking a `ConversionMode` option as part of the fix for #8708.

jfirebaugh:
Ah see, I disagree with fc0a0ef. It's much easier to avoid these sorts of bugs if the project (latLngToScreenCoordinate) and unproject (screenCoordinateToLatLng) operations are exact inverses, which can't be true if unproject automatically wraps its return value.

brunoabinader:
> It's much easier to avoid these sorts of bugs if the project (latLngToScreenCoordinate) and unproject (screenCoordinateToLatLng) operations are exact inverses, which can't be true if unproject automatically wraps its return value.

I agree. I've then moved the shortest path logic into `Map::pixelForLatLng`, which uses `ShortestPathMode::Use` by default. Inside `Map::cameraForLatLngs` we call `Map::pixelForLatLng` with `ShortestPathMode::Ignore`.

I'll be out until Tuesday so please feel free to add a patch for having `Map::cameraForLatLngs` to also use that enum (if needed) - with the changes made I don't think we need it though.

PR #8747, opened 2017-04-14
More GlyphAtlas refactoring

jfirebaugh:
* Replace `GlyphRangeSet` in `onGlyphsAvailable` with optionals in the map. `GlyphRangeSet` isn't keyed by `FontStack`, so using it to indicate that a particular range was loaded could have produced false positives.
* Combine the two `GlyphID`-keyed maps in `GlyphAtlas::Entry` into one.

### DIFF
--- a/src/mbgl/text/glyph_atlas.cpp
+++ b/src/mbgl/text/glyph_atlas.cpp
@@ -96,10 +96,14 @@ void GlyphAtlas::processResponse(const Response& res, const FontStack& fontStack
         }
 
         for (auto& glyph : glyphs) {
-            auto it = entry.sdfs.find(glyph.id);
-            if (it == entry.sdfs.end()) {
+            auto it = entry.glyphs.find(glyph.id);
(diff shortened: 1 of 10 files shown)


src/mbgl/text/glyph_atlas.cpp
```diff
@@ -147,27 +146,23 @@ void GlyphAtlas::addGlyphs(GlyphRequestor& requestor, const GlyphDependencies& g
         Entry& entry = entries[fontStack];
 
         for (const auto& glyphID : glyphIDs) {
-            loadedRanges.insert(getGlyphRange(glyphID));
+            optional<Glyph>& glyph = positions[glyphID];
```
ChrisLoer: This line is really important to the logic in the tile worker, and it's effect is a little bit subtle. Maybe put a comment like:

// Make a glyph position entry even if we didn't get an SDF for the glyph
// This distinguishes between "not loaded yet" and "loaded but nothing to show"

PR #8749, opened 2017-04-14
[ios] Fix desynchronization on user annotation

fabian-guerra:
Fixes https://github.com/mapbox/mapbox-gl-native/issues/7910

### DIFF
--- a/platform/ios/src/MGLMapView.mm
+++ b/platform/ios/src/MGLMapView.mm
@@ -1159,6 +1159,10 @@ public:
 {
     _changeDelimiterSuppressionDepth = 0;
     _mbglMap->setGestureInProgress(false);
+    if (self.userTrackingState == MGLUserTrackingStateBegan)
+    {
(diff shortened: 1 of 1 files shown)


platform/ios/src/MGLMapView.mm
```diff
@@ -1159,6 +1159,10 @@ - (void)touchesBegan:(__unused NS_SET_OF(UITouch *) *)touches withEvent:(__unuse
 {
     _changeDelimiterSuppressionDepth = 0;
     _mbglMap->setGestureInProgress(false);
+    if (self.userTrackingState == MGLUserTrackingStateBegan)
+    {
+        [self setUserTrackingMode:MGLUserTrackingModeNone];
```
1ec5: The only difference between this change and the one in #7916 is the `animated` parameter. Is it important to keep animation on in this case?

fabian-guerra:
This fix is already ongoing in https://github.com/mapbox/mapbox-gl-native/pull/7916

